PR #2531, opened 2020-08-06
Send immutable params on ext:configure

joehan:
### Description
Start sending immutable params on ext:configure. Previously, we were excluding theses, and since we use config.params as an update mask, this was wiping out immutable params and causing unexpected behavior.

### Scenarios Tested
Previously, we could not reconfigure instances with a location other than the default. With this change, I was able to successfully reconfigure an instance in us-east1.


src/commands/ext-configure.ts
```diff
@@ -65,9 +65,9 @@ export default new Command("ext:configure <extensionInstanceId>")
         const plural = immutableParams.length > 1;
         logger.info(`The following param${plural ? "s are" : " is"} immutable:`);
         for (const { param } of immutableParams) {
-          logger.info(
-            `param: ${param}, value: ${_.get(existingInstance, `config.params.${param}`)}`
-          );
+          const value = _.get(existingInstance, `config.params.${param}`);
+          logger.info(`param: ${param}, value: ${value}`);
+          params[param] = value;
```
tinaliang, line 70: Was not sending the immutable params on purpose? Or just an oversight?

joehan: This was a misunderstanding of how updateMasks work - my original intent was to not change those params by not sending them. However, since the updateMask is config.params, this was replacing the entire params object with what we sent, and getting rid of the immutable params. We never noticed because location is the only immutable param at the moment, and in testing, we've mostly used the default location us-central1

PR #2557, opened 2020-08-18
Wait for functions to exit before exiting functions emulator

### LINKED ISSUE #1813: emulators:exec should wait until there are no more pending function executions

**firebase-tools:** 7.8.0

**Platform:** Linux

### [REQUIRED] Test case

Two functions:

  * FunctionA has an HTTP trigger
  * FunctionB has a PubSub trigger on topic 'foo'

Let's say FunctionA looks like this:
```
for (let i = 0; i < 10; i++) {
  pubsub.send('foo', { data: i })
}
```

I would like to be able to wait until all 10 of the PubSub functions run but this is not captured by:

```
$ firebase emulators:exec "curl http://localhost:3001/my-project/us-central1/projectA
```

It would be nice if `emulators:exec` could ask the Functions emulator (and maybe all the others) if they're done.

Or this could also be implemented as having the Functions emulator's `stop()` function have a waiting option.

See above

See above

See above

samtstern:
### Description

Fixes #1813

Changes made:
  * Defines a consistent emulator stopping order
  * Waits up to 60s for the functions work queue to "flush" before killing them

### Scenarios Tested

I wrote a long-running function:
```js
exports.slowFunction = functions.https.onRequest(async (request, response) => {
  functions.logger.info("Hello logs!");
  await new Promise(res => setTimeout(res, 10 * 1000));
  functions.logger.info("Finished running");
});
```

And this is what it looks like if you kill while it's running:
```
i  functions: Beginning execution of "slowFunction"
>  {"severity":"INFO","message":"Hello logs!"}
^C 
i  emulators: Received SIGINT (Ctrl-C) for the first time. Starting a clean shutdown.
i  emulators: Please wait for a clean shutdown or send the SIGINT (Ctrl-C) signal again to stop right now.
i  ui: Stopping Emulator UI
⚠  Emulator UI has exited upon receiving signal: SIGINT
i  functions: Stopping Functions Emulator
i  functions: Waiting for all functions to finish...
>  {"severity":"INFO","message":"Finished running"}
i  functions: Finished "slowFunction" in ~9s
i  firestore: Stopping Firestore Emulator
i  hub: Stopping emulator hub
i  logging: Stopping Logging Emulator
```

### Sample Commands

N/A

### DIFF
--- a/src/emulator/functionsEmulator.ts
+++ b/src/emulator/functionsEmulator.ts
@@ -368,10 +368,20 @@ export class FunctionsEmulator implements EmulatorInstance {
     return loadTriggers();
   }
 
-  stop(): Promise<void> {
+  async stop(): Promise<void> {
+    try {
+      await this.workQueue.flush();
+    } catch (e) {
+      this.logger.logLabeled(
+        "WARN",
+        "functions",
+        "Functions emulator work queue did not empty before stopping"
+      );
+    }
+
     this.workQueue.stop();
     this.workerPool.exit();
-    return this.destroyServer ? this.destroyServer() : Promise.resolve();
+    this.destroyServer && this.destroyServer();
   }
 
   addRealtimeDatabaseTrigger(
@@ -641,6 +651,10 @@ export class FunctionsEmulator implements EmulatorInstance {
     const childProcess = spawn(opts.nodeBinary, args, {
       env: { node: opts.nodeBinary, ...opts.env, ...process.env },
       cwd: frb.cwd,
+      // By running as detached the child proccesses don't pick up
+      // kill signals to the host. This allows functions to finish
+      // running in workqueue.flush()
+      detached: true,
       stdio: ["pipe", "pipe", "pipe", "ipc"],
     });
 
--- a/src/emulator/functionsEmulator.ts
+++ b/src/emulator/functionsEmulator.ts
@@ -651,10 +651,6 @@ export class FunctionsEmulator implements EmulatorInstance {
     const childProcess = spawn(opts.nodeBinary, args, {
       env: { node: opts.nodeBinary, ...opts.env, ...process.env },
       cwd: frb.cwd,
-      // By running as detached the child proccesses don't pick up
-      // kill signals to the host. This allows functions to finish
-      // running in workqueue.flush()
-      detached: true,
       stdio: ["pipe", "pipe", "pipe", "ipc"],
     });
 
--- a/src/emulator/functionsEmulator.ts
+++ b/src/emulator/functionsEmulator.ts
@@ -381,7 +381,9 @@ export class FunctionsEmulator implements EmulatorInstance {
 
     this.workQueue.stop();
     this.workerPool.exit();
-    this.destroyServer && this.destroyServer();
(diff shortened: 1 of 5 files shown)


src/emulator/functionsEmulator.ts
```diff
@@ -641,6 +651,10 @@ export class FunctionsEmulator implements EmulatorInstance {
     const childProcess = spawn(opts.nodeBinary, args, {
       env: { node: opts.nodeBinary, ...opts.env, ...process.env },
       cwd: frb.cwd,
+      // By running as detached the child proccesses don't pick up
+      // kill signals to the host. This allows functions to finish
+      // running in workqueue.flush()
+      detached: true,
```
samtstern, line 657: This is the part that scares me ... is this tradeoff worth it?  The workqueue should still kill all the child processes during `stop()` but there's always a chance it explodes and leaves zombies.

yuchenshi: I haven't really reviewed the rest of the PR, but wild idea:

Instead of spinning these up detached, can we inject SIGINT handlers into the child processes instead, so they gracefully shutdown (and maybe force shutdown when pressed twice)?

We may have to do [`node -r`](https://nodejs.org/api/cli.html#cli_r_require_module) or put more logic into the testing module or require hooks, but I think that scares me less than having detached processes run loose.

samtstern: @yuchenshi great idea!  I installed a signal handler in the runtime and it works perfectly.

src/emulator/functionsEmulator.ts
```diff
@@ -368,10 +368,20 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+        "WARN",
+        "functions",
+        "Functions emulator work queue did not empty before stopping"
+      );
+    }
+
     this.workQueue.stop();
     this.workerPool.exit();
-    return this.destroyServer ? this.destroyServer() : Promise.resolve();
+    this.destroyServer && this.destroyServer();
```
yuchenshi, line 384: Should still return / await this so that server is properly killed before exiting

src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -1088,6 +1088,25 @@ async function handleMessage(message: string) {
 }
 
 async function main(): Promise<void> {
+  // Since the functions run as attached processes they naturally inherit SIGINT
+  // sent to the functions emulator. We want them to ignore the first signal
+  // to allow for a clean shutdown.
+  let lastSignal = new Date().getTime();
```
yuchenshi, line 1094: Optional: extract a helper function for this double signal logic (and reuse for `emulators:start`) as well

src/emulator/functionsEmulator.ts
samtstern: Oops good catch!

src/emulator/functionsEmulatorRuntime.ts
samtstern: Yeah it's not quite similar enough due to the different logging/messaging and also we try not to share too much code with the runtime, so non-DRY it is :-/

PR #2563, opened 2020-08-20
Fix failure policy deploys

samtstern:
### Description

Spotted by @armordog ... the failure policy PR makes **ALL** deploys go through the warning step.

### Scenarios Tested

1):
```js
const admin = require('firebase-admin');
admin.initializeApp();

exports.helloworld = functions.https.onRequest(async (req, res) => {
  res.json({ ok: true });
});
```

2)
```js
const admin = require('firebase-admin');
admin.initializeApp();

exports.helloworld = functions.https.onRequest(async (req, res) => {
  res.json({ ok: true });
});

exports.policyEmptyObject = functions
  .runWith({
    failurePolicy: { retry: {} }
  })  
  .pubsub.topic("retry-topic")
  .onPublish((msg, ctx) => {
    return true;
  });
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Fixes a bug where all deploys asked about failure policies.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,1 @@
-- Fixes a bug where all deploys asked about failure policies.
+- Fixes a bug where all Cloud Functions for Firebase deploys asked about failure policies.
--- a/src/deploy/functions/release.js
+++ b/src/deploy/functions/release.js
@@ -158,7 +158,7 @@ module.exports = function(context, options, payload) {
   });
 
   let proceedPrompt = Promise.resolve(true);
-  if (failurePolicyFunctions.length > 0) {
+  if (failurePolicyFunctions.length) {
     var failurePolicyFunctionLabels = failurePolicyFunctions.map((fn) => {
       return helper.getFunctionLabel(_.get(fn, "name"));
     });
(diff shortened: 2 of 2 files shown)


src/deploy/functions/release.js
```diff
@@ -157,30 +157,32 @@ module.exports = function(context, options, payload) {
...
-    throw new FirebaseError("Pass the --force option to deploy functions with a failure policy", {
-      exit: 1,
-    });
-  } else if (!options.nonInteractive) {
-    proceedPrompt = promptOnce({
-      type: "confirm",
-      name: "confirm",
-      default: false,
-      message: "Would you like to proceed with deployment?",
+  if (failurePolicyFunctions.length > 0) {
```
joehan, line 161: This can just be `if (failurePolicyFunction.length)`

CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- Fixes a bug where all deploys asked about failure policies.
```
bkendall, line 1: Are these specific deploys? 
```suggestion
- Fixes a bug where all Cloud Functions for Firebase deploys asked about failure policies.
```

rbruggem:
Could you please clarify why the `--force` flag is necessary when setting a failure policy and the function has no failure policy set?
I do understand the need of having a `--force` flag if the current state has a failure policy and the function is being deployed without an explicit failure policy.

joehan:
@rbruggem IIRC, it requires the `--force` flag to add a failure policy because function retries can be dangerous in certain cases - specifically, functions that are not written to be idempotent could have unexpected effects, and functions that perform potentially expensive operations (like reading large amounts of data from a database) could run up someone's bill.

samtstern:
@joehan @rbruggem that's correct.  Failure policies can be quite dangerous and are a frequent source of surprise bills for customers who use them improperly.  That's why we took so long to support them and make the developer clearly opt-in to using them by accepting their risks.

PR #2580, opened 2020-08-28
Add support for service account in functions

egor-miasnikov:
### Description

Supports firebase-functions SDK feature in [firebase/firebase-function#770](https://github.com/firebase/firebase-functions/pull/770)

```js
functions
      .runWith({
        serviceAccountEmail: 'test-service-account@test.iam.gserviceaccount.com'
      })
      .auth.user()
      .onCreate((user) => user);
```

It's an important feature for people who want to decrease and/or granulate roles and permission. For instance, read-only access for their cloud functions to a `CloudSQL` database. Currently, a developer needs to use the second deploy process for functions to change the service account.

### Scenarios Tested

I've tested creating a new cloud function with specific service account and it works. I also tried updating a function and works fine too. I also tested without setting the service account property and it worked too (will be set default service account)

### Sample Commands

No actual change in CLI options.


samtstern:
@egor-miasnikov thanks for this contribution!  Let's first discuss on `firebase-functions` and then come back here if/when that is approved.

lmakarov:
> Currently, a developer needs to use the second deploy process for functions to change the service account.

@egor-miasnikov would you mind sharing more details about how you are currently changing the service account for a firebase function?

egor-miasnikov:
hey, @lmakarov we did double deploy for functions, but currently, after this PR we're using patched firebase libraries with our changes ;) and waiting for some news from @samtstern

lmakarov:
> we did double deploy for functions

@egor-miasnikov do you mean something like?

```
firebase deploy --only functions:<function-name>
gcloud functions deploy <function-name> --service-account=<custom-service-account>
```

egor-miasnikov:
@egor-miasnikov exactly

joehan:
Hey all, sorry for the delay on this, the API review slipped past us for a bit, but it is approved now! There were a few other changes  to the API requested, so I'm going to make those this week and do some testing. I'm hoping to get this released sometime later this week if all goes well. Thanks for the contribution and patience!

joehan (approved):

PR #2582, opened 2020-08-28
Redirect user to Console for uninstallation of pubsub-stream-bigquery

tinaliang:
### Description

Regular uninstall flow:
Screen Shot 2020-08-28 at 12 28 45 PM

Uninstall flow for a pubsub-stream-bigquery extension
Screen Shot 2020-09-01 at 10 19 09 AM

### DIFF
--- a/src/commands/ext-uninstall.ts
+++ b/src/commands/ext-uninstall.ts
@@ -34,7 +34,8 @@ function consoleUninstallOnly(projectId: string, instanceId: string): Promise<vo
   const consoleUninstall =
     "This extension can only be uninstalled through the Firebase Console. " +
     `Please visit **[${instanceURL}](${instanceURL})** to uninstall this extension.`;
-  return utils.reject(marked(consoleUninstall));
+  utils.logLabeledWarning(logPrefix, marked(consoleUninstall));
+  return Promise.resolve();
 }
 
 export default new Command("ext:uninstall <extensionInstanceId>")
--- a/src/commands/ext-uninstall.ts
+++ b/src/commands/ext-uninstall.ts
@@ -24,10 +24,9 @@ marked.setOptions({
 
 /**
  * We do not currently support uninstalling extensions that require additional uninstall steps to be taken in the CLI. Direct them to the Console to uninstall the extension.
- * - pubsub-stream-bigquery: requires calling a public HTTP uninstall function
  *
- * @param projectId id of the user's project
- * @param instanceId id of the extension instance
+ * @param projectId ID of the user's project
+ * @param instanceId ID of the extension instance
  * @return a void Promise
  */
 function consoleUninstallOnly(projectId: string, instanceId: string): Promise<void> {
--- a/src/commands/ext-uninstall.ts
+++ b/src/commands/ext-uninstall.ts
@@ -32,8 +32,9 @@ marked.setOptions({
 function consoleUninstallOnly(projectId: string, instanceId: string): Promise<void> {
   const instanceURL = `https://console.firebase.google.com/project/${projectId}/extensions/instances/${instanceId}`;
   const consoleUninstall =
-    "This extension can only be uninstalled through the Firebase Console. " +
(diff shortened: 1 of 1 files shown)


src/commands/ext-uninstall.ts
```diff
@@ -15,6 +18,26 @@ import { requirePermissions } from "../requirePermissions";
 import * as utils from "../utils";
 import * as logger from "../logger";
 
+marked.setOptions({
+  renderer: new TerminalRenderer(),
+});
+
+/**
+ * We do not currently support uninstalling extensions that require additional uninstall steps to be taken in the CLI. Direct them to the Console to uninstall the extension.
+ * - pubsub-stream-bigquery: requires calling a public HTTP uninstall function
```
laurenzlong, line 27: Please delete this comment line about "pubsub-stream-bigquery", don't need to go into implementation details here.

src/commands/ext-uninstall.ts
```diff
@@ -15,6 +18,26 @@ import { requirePermissions } from "../requirePermissions";
...
+ *
+ * @param projectId id of the user's project
```
laurenzlong, line 29: capitalize ID in this line and below.

src/commands/ext-uninstall.ts
```diff
@@ -15,6 +18,26 @@ import { requirePermissions } from "../requirePermissions";
...
+ * @param instanceId id of the extension instance
+ * @return a void Promise
+ */
+function consoleUninstallOnly(projectId: string, instanceId: string): Promise<void> {
+  const instanceURL = `https://console.firebase.google.com/project/${projectId}/extensions/instances/${instanceId}`;
+  const consoleUninstall =
```
samtstern, line 35: Is there any reason to use Markdown here?  I assume the renderer will only underline the link, or does it somehow make it more clickable?

Elsewhere in the CLI codebase I'd expect:
```ts
import * as clc from "cli-color";
// ...
const consoleUninstall = `This extension can only be uninstalled through the Firebase Console. Please visit ${clc.bold(underline(instanceUrl))} to uninstall this extension`
```

src/commands/ext-uninstall.ts
```diff
@@ -15,6 +18,26 @@ import { requirePermissions } from "../requirePermissions";
 import * as utils from "../utils";
 import * as logger from "../logger";
 
+marked.setOptions({
```
samtstern, line 21: Does this modify a global state that will affect other uses of `marked` in the codebase?  If so is there a way to get an "instance"?

src/commands/ext-uninstall.ts
tinaliang: On certain terminals, this turns into a clickable link

src/commands/ext-uninstall.ts
tinaliang: This creates a reference instance: see https://marked.js.org/#/USING_ADVANCED.md

annie-ryan:
> Option 1: Warning
> Screen Shot 2020-08-31 at 9 36 21 AM
> 

I agree that it should be presented more as info, rather than as if the user did something wrong.

I think having a space between "Required API ___ is enabled" and "This extension can only be installed through the Firebase console..." would help make it stand out.

Is there any other info we could add to explain to users why they must go to the console?

tinaliang:
> > Option 1: Warning
> > Screen Shot 2020-08-31 at 9 36 21 AM
> 
> I agree that it should be presented more as info, rather than as if the user did something wrong.
> 
> I think having a space between "Required API ___ is enabled" and "This extension can only be installed through the Firebase console..." would help make it stand out.
> 
> Is there any other info we could add to explain to users why they must go to the console?

Updated the wording (see Description for screenshot), let me know what you think!

tinaliang:
> > Updated the wording (see Description for screenshot), let me know what you think!
> 
> Maybe we can say something along the lines of "This extension has additional uninstall **checks** that are not currently supported by the CLI..."
> 
> I don't know if "checks" is the right word, but I don't want people going to the console and looking for extra steps to complete. Since the uninstall dialog is the same(ish).

I like that wording! Updated.

PR #2595, opened 2020-09-02
Change update warning to account for other ways of installing the CLI

joehan:
### Description
Change update warning to account for other ways of installing the CLI.
Copy is from b/157902668
### Scenarios Tested
Screen Shot 2020-09-02 at 1 29 03 PM

### DIFF
--- a/src/bin/firebase.js
+++ b/src/bin/firebase.js
@@ -18,7 +18,16 @@ if (!semver.satisfies(nodeVersion, pkg.engines.node)) {
 }
 
 const updateNotifier = require("update-notifier")({ pkg: pkg });
-updateNotifier.notify({ defer: true, isGlobal: true });
+const clc = require("cli-color");
+const TerminalRenderer = require("marked-terminal");
+const marked = require("marked");
+marked.setOptions({
+  renderer: new TerminalRenderer(),
+});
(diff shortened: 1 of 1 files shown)


src/bin/firebase.js
```diff
@@ -18,7 +18,19 @@ if (!semver.satisfies(nodeVersion, pkg.engines.node)) {
...
 const updateNotifier = require("update-notifier")({ pkg: pkg });
-updateNotifier.notify({ defer: true, isGlobal: true });
+const clc = require("cli-color");
+const TerminalRenderer = require("marked-terminal");
+const marked = require("marked");
+marked.setOptions({
+  renderer: new TerminalRenderer(),
+});
+const updateMessage =
+  `Update available ${clc.xterm(240)("{currentVersion}")} → ${clc.green("{latestVersion}")}\n` +
```
joehan, line 28: I tried to copy the style from update-notifier's default message as closely as possible. Unfortunately, they use `chalk` instead of `cli-color` for styling, and there isn't an equivalent method in `cli-color to `chalk.cim()`. `xterm(240)` is pretty close, however.

bkendall (approved):
Add an entry to the changelog:

```
- Updates the update notifier messaging.
```

Or some such?

PR #2596, opened 2020-09-02
Adds checks for mininmum required version for ext and ext-dev commands

joehan:
### Description
Adds a generic function to check for minimum required version for a command, and call it before the `ext` and `ext-dev` commands. The minimum versions are set in the configuration fetched by [fetchMOTD](https://github.com/firebase/firebase-tools/blob/master/src/fetchMOTD.ts) - we'll need to go set those separately.


src/checkMinRequiredVersion.ts
```diff
@@ -0,0 +1,20 @@
+import * as semver from "semver";
+
+import { configstore } from "./configstore";
```
huangjeff5, line 3: What's the workflow for updating configstore?

huangjeff5 (approved):

src/checkMinRequiredVersion.ts
joehan: @mbleigh knows more about this than I do - it involves updating an RTDB node on a project we own. When we want to set this min version, he should be able to help us get it set up.

PR #2599, opened 2020-09-03
Auth for Hosting Channels

dahliasalem:
### Description

Implementing Auth for Channels as described in https://goto.google.com/fh-preview-auth-domains


src/hosting/api.ts
```diff
@@ -275,3 +275,41 @@ export async function createRelease(site: string, channel: string, version: stri
   );
   return res.body;
 }
+
+/**
+ * Adds channel domain to auth.
+ * @param project the project ID.
+ * @param url the url of the cjannel.
```
jthegedus, line 282: typo
> cjannel

src/hosting/api.ts
```diff
@@ -275,3 +275,82 @@ export async function createRelease(site: string, channel: string, version: stri
   );
   return res.body;
 }
+
+async function getAuthDomains(project: string): Promise<string[]> {
```
bkendall, line 279: None of these functions belong to the Hosting API. Could we move them to their own file? We have a catch-all `src/gcp/` folder that contains other GCP related things, maybe that would be a good candidate.

src/hosting/api.ts
```diff
@@ -275,3 +275,82 @@ export async function createRelease(site: string, channel: string, version: stri
...
+  const res = await api.request("GET", `/admin/v2/projects/${project}/config`, {
+    auth: true,
+    origin: "https://identitytoolkit.googleapis.com",
+  });
+  return res?.body?.authorizedDomains;
+}
+
+async function updateAuthDomains(project: string, authDomains: string[]): Promise<any> {
```
bkendall, line 287: Can a type be defined for these `any`s that are returned? It'll make it much easier for other readers to understand what's being returned.

src/hosting/api.ts
```diff
@@ -275,3 +275,82 @@ export async function createRelease(site: string, channel: string, version: stri
...
+  authDomains.push(domain);
+  return await updateAuthDomains(project, authDomains);
+}
+
+/**
+ * Syncs the state of authorized domains with existing channels.
+ * @param project the project ID.
+ * @param site the site for the channel.
+ */
+export async function syncAuthState(project: string, site: string): Promise<any> {
```
bkendall, line 322: Can this function be tested? It looks like some non-trivial logic.

bkendall (changes requested):

src/api.js
```diff
@@ -141,6 +141,7 @@ var api = {
     utils.envOverride("FIREBASE_GOOGLE_URL", "https://www.googleapis.com")
   ),
   hostingOrigin: utils.envOverride("FIREBASE_HOSTING_URL", "https://web.app"),
+  identityOrigin: "https://identitytoolkit.googleapis.com",
```
bkendall, line 144: wrap this in `utils.envOverride`, maybe with `FIREBASE_IDENTITY_URL`?

```suggestion
  identityOrigin: utils.envOverride("FIREBASE_IDENTITY_URL", "https://identitytoolkit.googleapis.com"),
```

src/commands/hosting-channel-create.ts
```diff
@@ -69,6 +70,20 @@ export default new Command("hosting:channel:create [channelId]")
         throw e;
       }
 
+      try {
+        await addAuthDomain(projectId, channel.url);
+      } catch (e) {
+        logLabeledWarning(
```
bkendall, line 76: maybe `logger.debug` this error as well so we can debug when users have issues

src/commands/hosting-channel-create.ts
```diff
@@ -69,6 +70,20 @@ export default new Command("hosting:channel:create [channelId]")
...
+          LOG_TAG,
+          marked(
+            `Unable to add channel domain to Firebase Auth. Visit the Firebase Console at ${consoleUrl(
+              projectId,
+              "/overview"
+            )}`
```
bkendall, line 82: I'm curious about this call to action - what am I supposed to be doing on the Console? Is there something more specific we can say? Should we drop the call to action?

src/hosting/api.ts
```diff
@@ -275,3 +276,64 @@ export async function createRelease(site: string, channel: string, version: stri
...
+  // Create a map of channel domain names
+  const channelMap = channels
+    .map((channel: Channel) => channel.url.replace("https://", ""))
+    .reduce((acc: { [key: string]: boolean }, current: string) => {
+      acc[current] = true;
+      return acc;
+    }, {});
+
+  // match any string that has ${site}--*
+  const siteMatch = new RegExp(site, "i");
```
bkendall, line 307: Should this be `${site}--` in the regular expression, to match the comment?

src/hosting/api.ts
```diff
@@ -275,3 +276,64 @@ export async function createRelease(site: string, channel: string, version: stri
...
+    const domainWithNoChannel = siteMatch.test(domain) && !channelMap[domain];
+    if (domainWithNoChannel) {
+      return;
+    }
+    // add all other domains (ex: "localhost", etc)
+    authDomains.push(domain);
+  });
+  return authDomains;
+}
+/**
```
bkendall, line 331: keep a line break between functions, please

src/commands/hosting-channel-deploy.ts
```diff
@@ -102,9 +109,28 @@ export default new Command("hosting:channel:deploy [channelId]")
                 logger.debug("[hosting] updated TTL for existing channel for site", site, chan);
               }
             }
+            try {
+              await syncAuthState(projectId, site);
+            } catch (e) {
+              // not sure if we actually want to print a warning when we cant sync?
+              logLabeledWarning(LOG_TAG, "Unable to sync Firebase Auth state.");
+            }
```
bkendall, line 117: Could this syncAuthState be called after both the "channel exists" and "channel doesn't exist" logic is done? Sure it means that we might list the channels more often than necessary, but it would reduce the number of locations where we deal with auth state - I think that's a worthy tradeoff.

src/hosting/api.ts
```diff
@@ -275,3 +276,64 @@ export async function createRelease(site: string, channel: string, version: stri
   );
   return res.body;
 }
+
+/**
+ * Adds channel domain to Firebase Auth list.
+ * @param project the project ID.
+ * @param url the url of the channel.
+ */
+export async function addAuthDomain(project: string, url: string): Promise<Domains> {
```
bkendall, line 285: It's a little hard to keep track of the return values here when one is Domains and the other is string[] - could this and getApprovedDomains be unified to the same return type, maybe `string[]`?

src/commands/hosting-channel-deploy.ts
```diff
@@ -102,9 +109,28 @@ export default new Command("hosting:channel:deploy [channelId]")
                 logger.debug("[hosting] updated TTL for existing channel for site", site, chan);
               }
             }
+            try {
+              await syncAuthState(projectId, site);
+            } catch (e) {
```
bkendall, line 114: log.debugger this error too?

src/test/hosting/api.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+describe("hosting", () => {
+  beforeEach(() => {
+    helpers.mockAuth(sinon);
+  });
+
+  afterEach(() => {
+    sinon.restore();
+  });
+  describe("getApprovedDomains", () => {
```
bkendall, line 44: line break between blocks, please :)

src/test/hosting/api.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+      nock.cleanAll();
+    });
+
+    it("it should return the list of expetecd auth domains after syncing", async () => {
+      // mock listChannels response
+      nock(api.hostingApiOrigin)
+        .get(`/v1beta1/projects/${PROJECT_ID}/sites/${SITE}/channels`)
+        .query((queryParams: any) => {
+          return queryParams.pageSize === "100";
+        })
```
bkendall, line 55: I love me some typescript code-golf :)

```typescript
.query((p: any) => p.pageSize === "100");
```

src/test/hosting/api.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  describe("getApprovedDomains", () => {
+    afterEach(() => {
+      nock.cleanAll();
+    });
+
+    it("it should return the list of expetecd auth domains after syncing", async () => {
+      // mock listChannels response
+      nock(api.hostingApiOrigin)
+        .get(`/v1beta1/projects/${PROJECT_ID}/sites/${SITE}/channels`)
+        .query((queryParams: any) => {
```
bkendall, line 53: I don't recall why you need this `.query` - does `nock` filter on query parameters (and this wouldn't work without specifying?)?

src/test/hosting/api.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+import * as helpers from "../helpers";
+import * as api from "../../api";
+import * as hostingApi from "../../hosting/api";
+
+const TEST_CHANNELS_RESPONSE = {
+  channels: [
+    // domain exists in TEST_GET_DOMAINS_RESPONSE
+    { url: "https://my-site--ch1-4iyrl1uo.web.app" },
+    // domain does not exist in TEST_GET_DOMAINS_RESPONSE
+    { url: "https://my-site--ch2-ygd8582v.web.app" },
```
bkendall, line 13: If this is an existing channel, and it's not in GET_DOMAINS, shouldn't it be added (and be in EXPECTED_DOMAINS)?

bkendall (changes requested):

src/commands/hosting-channel-create.ts
dahliasalem: I am not sure.. but the design docs specify to give them a link to the console? I was hoping you would know go/fh-preview-auth-domains

src/commands/hosting-channel-deploy.ts
dahliasalem: I am fine with that just thought that the design specifies to syncAuthState only after a deploy to an existing channel

src/test/hosting/api.spec.ts
dahliasalem: Not necessarily because this domain could have been added manually via the identity api I just added to show that no new domains get be added during sync, domains are only removed in sync

src/test/hosting/api.spec.ts
dahliasalem: yes because the for the mock to work it need to match the real query 100%

src/commands/hosting-channel-deploy.ts
```diff
@@ -105,8 +112,26 @@ export default new Command("hosting:channel:deploy [channelId]")
           } else {
             chan = await createChannel(projectId, site, channelId, expireTTL);
             logger.debug("[hosting] created new channnel for site", site, chan);
+            try {
+              await addAuthDomain(projectId, chan.url);
+            } catch (e) {
```
bkendall, line 117: logger.debug here too

src/test/hosting/api.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  describe("getApprovedDomains", () => {
+    afterEach(() => {
+      nock.cleanAll();
+    });
+
+    it("it should return the list of expetecd auth domains after syncing", async () => {
```
bkendall, line 50:
```suggestion
    it("should return the list of expected auth domains after syncing", async () => {
```

src/test/hosting/api.spec.ts
bkendall: I don't understand, the domain here (`ch2`) is coming from the Channels response, not the identity API. The case I think you described (not removing other domains from the identity API that Hosting doesn't control) would be a domain in GET_DOMAINS _not_ being remove in EXPECTED_DOMAINS, right? (a la `randomurl.com`).

Since `ch2` is a channel that exists (coming in the CHANNELS_RESPONSE) and it is NOT in GET_DOMAINS, I would expect that it would need to be added to the list of authorized domains. (The case here may be that a previous sync failed, or the domain was accidentally removed - the CLI should add it back)

bkendall (changes requested):

src/test/hosting/api.spec.ts
mbleigh: So this was an intentional part of the design - if someone manually removes a channel that isn't the current channel being deployed, we should not assume that it was done so accidentally and "correct" it.

If the channel being deployed / created doesn't already have an authorizedDomain, we should add it (because we offer a flag to skip adding it if users desire), but if an unrelated channel is missing its authorizedDomain we should not make assumptions about adding it back. Removing channels domains that no longer exist is safe because we know for a fact the domain doesn't exist and therefore can't break anything the user intended to do.

dahliasalem: As Micheal said, we do not add any domains during sync we only remove them and that is to respect if the user had manually removed that domain themselves via the console or the identity api

src/test/hosting/api.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  describe("getApprovedDomains", () => {
+    afterEach(() => {
+      nock.cleanAll();
+    });
+
+    it("it should return the list of expected auth domains after syncing", async () => {
```
bkendall, line 52: the "it" doesn't need to be repeated

```suggestion
    it("should return the list of expected auth domains after syncing", async () => {
```

src/hosting/api.ts
```diff
@@ -275,3 +276,65 @@ export async function createRelease(site: string, channel: string, version: stri
...
+      return;
+    }
+    // add all other domains (ex: "localhost", etc)
+    authDomains.push(domain);
+  });
+  return authDomains;
+}
+
+/**
+ * Syncs the state of authorized domains with existing channels.
```
bkendall, line 333: Could this doc string (maybe getApprovedDomains also/instead?) be updated to reflect how this sync works? When I read "sync" I expect the two lists to come together completely. In this case, any domain listed in the identity API that is a Firebase Channel that no longer exists is removed, and no additional Channels are added to the list (this last part I wouldn't have expected).

Maybe "sync" is the wrong verb here. What about "prune"? If I've followed correctly now, this only will ever remove expired channels, right? That might help clarify what's going on

bkendall (changes requested):
Looking good to me - I have some comments on testing and docs and this point, but generally things look good to go! Thanks!

src/test/hosting/api.spec.ts
bkendall: The _manually removing a channel from the identity API_ was the part that wasn't coming across. "Sync" to me means that two lists become unified, not that one list is pruned - added a comment to that effect as well. Thanks for all the clarification.

src/commands/hosting-clone.ts
```diff
@@ -74,7 +77,27 @@ export default new Command("hosting:clone <source> <targetChannel>")
         )}, creating it...`
       );
       tChannel = await createChannel("-", targetSiteId, targetChannelId);
+      if (!tChannel) {
```
bkendall, line 80: If it cannot create the channel, does `createChannel` return a falsey value or does it throw an error? (my guess in the second, so this would need to be a `catch`)

src/hosting/api.ts
```diff
@@ -355,3 +356,65 @@ export async function createRelease(
...
+
+/**
+ * Adds channel domain to Firebase Auth list.
+ * @param project the project ID.
+ * @param url the url of the channel.
+ */
+export async function addAuthDomain(project: string, url: string): Promise<string[]> {
+  const domains = await getAuthDomains(project);
+  const domain = url.replace("https://", "");
+  let authDomains = domains || [];
```
bkendall, line 368:
```suggestion
  const authDomains = domains || [];
```

src/hosting/api.ts
bkendall: Can the docs be updated here to describe what is done in cleanAuthState? "Sync" is the word that really traps me here - can the description be made more explicit?

src/utils.ts
```diff
@@ -478,3 +478,14 @@ export function datetimeString(d: Date): string {
 export function isCloudEnvironment() {
   return !!process.env.CODESPACES;
 }
+
+/**
+ * Returns the projectId from a channel name string.
```
bkendall, line 483: This is a little too specific of a description for the global `utils` folder, and I'm not sure it warrants being a util given the simplicity of it. Consider refactoring this as a local function in the clone command where it's used.

bkendall (changes requested):
looks good. couple tiny bits of clean up and let's ship it!

src/hosting/api.ts
bkendall: This wasn't fixed

bkendall (approved):
One tiny catch that didn't make it, but otherwise LGTM.

Please be sure to "Squash and Merge"

Issue #1835, opened 2019-11-26
Deploy with pubsub.schedule says success when failed to create Scheduler Job


stari4ek:
### Related issues

https://github.com/firebase/firebase-functions/issues/578

### [REQUIRED] Version info

**node:** 

```
$ node --version
v12.4.0
```

**firebase-functions:**
```
$ npm list --depth 0 | grep firebase
├── firebase-admin@8.8.0
├── firebase-functions@3.3.0
├── firebase-functions-test@0.1.7
```

**firebase-admin:**
```
$ firebase --version
7.8.1
```

### [REQUIRED] Steps to reproduce

deploy function with `pubsub.schedule`

### [REQUIRED] Expected behavior

* pubsub topic and subscription are created
* scheduled job is created or error reported.
* firebase function is created

### [REQUIRED] Actual behavior

```
$ firebase deploy --only functions:auto_update_tick

=== Deploying to 'SOMEPROJ'...

i  deploying functions
Running command: npm --prefix "$RESOURCE_DIR" run lint

> functions@ lint /home/astar/dev/proj/develop/fb/functions
> tslint --project tsconfig.json

Running command: npm --prefix "$RESOURCE_DIR" run build

> functions@ build /home/astar/dev/proj/develop/fb/functions
> tsc

✔  functions: Finished running predeploy script.
i  functions: ensuring necessary APIs are enabled...
✔  functions: all necessary APIs are enabled
i  functions: preparing functions directory for uploading...
i  functions: packaged functions (133.62 KB) for uploading
✔  functions: functions folder uploaded successfully
i  functions: current functions in project: ...
i  functions: uploading functions in project: auto_update_tick(us-central1)
i  functions: creating Node.js 10 (Beta) function auto_update_tick(us-central1)...
i  scheduler: ensuring necessary APIs are enabled...
i  pubsub: ensuring necessary APIs are enabled...
✔  scheduler: all necessary APIs are enabled
✔  pubsub: all necessary APIs are enabled
✔  functions: created scheduler job firebase-schedule-auto_update_tick-us-central1
✔  functions[auto_update_tick(us-central1)]: Successful create operation. 

✔  Deploy complete!
```

While in `firebase-debug.log`:
```
[...]

PR #2609, opened 2020-09-11
Fix bug where scheduled CF3 deploys unexpectedly succeeded with invalid Cloud Scheduler configurations.

taeold:
Today, deploying scheduled CF3 with invalid cloud scheduler config unexpectedly succeeds:

```js
// functions/index.js
exports.scheduledHello = functions.pubsub
    .schedule('* * * * *')
    .timeZone('INVALID TIMEZONE')
    .onRun(async context => {
        functions.logger.info("hello world");
        return null;
    });
```
```shell
$ firebase deploy --only functions:scheduledHello
...
✔  scheduler: required API cloudscheduler.googleapis.com is enabled
✔  pubsub: required API pubsub.googleapis.com is enabled
✔  functions: created scheduler job firebase-schedule-scheduledHello-us-central1
✔  functions[scheduledHello(us-central1)]: Successful update operation.

✔  Deploy complete!
```

We end up with a broken setup where Cloud Function and Cloud Pubsub topics are deployed without a Cloud Scheduler job.

With the fix, the deployment will fail instead:

```shell
$ firebase deploy --only functions:scheduledHello
...
✔  scheduler: required API cloudscheduler.googleapis.com is enabled
✔  pubsub: required API pubsub.googleapis.com is enabled

Error: HTTP Error: 400, Schedule or time zone is invalid.
```

Neither Cloud Function or Pubsub topic are created since creating Cloud Scheduler job is done before Functions of Pubsub creation.

Fixes https://github.com/firebase/firebase-tools/issues/1835.


CHANGELOG.md
```diff
@@ -1 +1,2 @@
 - Replaces `tslint` with `eslint` as the default linter in new Cloud Functions for Firebase directories. (#1663)
+- Fix bug where scheduled CF3 deploys unexpectedly succeeded with invalid Cloud Scheduler configurations. (#2609)
```
bkendall, line 2:
```suggestion
- Fixes bug where scheduled Cloud Functions for Firebase deploys unexpectedly succeeded with invalid Cloud Scheduler configurations. (#2609)
```

src/gcp/cloudscheduler.ts
```diff
@@ -108,6 +108,7 @@ export async function createOrReplaceJob(job: Job): Promise<any> {
           )
         );
       }
+      return Promise.reject(err);
```
bkendall, line 111: Can you `throw` this error (and the one above it, for that matter) rather than returning a rejected Promise? They should function more or less the same way but we'd be not adding one more Promise to the mix.

```suggestion
      throw err;
```

src/test/gcp/cloudscheduler.spec.ts
```diff
@@ -115,5 +115,21 @@ describe("cloudscheduler", () => {
...
+      nock(api.cloudschedulerOrigin)
+        .get(`/${VERSION}/${TEST_JOB.name}`)
+        .reply(404, { context: { response: { statusCode: 404 } } });
+      nock(api.cloudschedulerOrigin)
+        .post(`/${VERSION}/projects/test-project/locations/us-east1/jobs`)
+        .reply(400, { context: { response: { statusCode: 400 } } });
+
+      await expect(cloudscheduler.createOrReplaceJob(TEST_JOB)).to.be.rejectedWith(
+        FirebaseError,
+        ""
```
bkendall, line 129: Why is this testing that the error message contains an empty string? Can this be more specific?

bkendall (changes requested):
Is it possible (useful?) to include the function name that has a bad schedule in the error message? For instance:

```
Error... scheduled function `helloWorld` failed to create schedule: ${error message returned}
```

src/gcp/cloudscheduler.ts
taeold: Done!

src/test/gcp/cloudscheduler.spec.ts
taeold: Done!

taeold:
@bkendall I think it's simple to add scheduler name (where careful reading if the error message will let users know which scheduled function is impacted) but less simple to add function name:

I've opted for adding scheduler name to the error message, e.g:
```
Error: Failed to create scheduler job projects/~locations/us-central1/jobs/firebase-schedule-scheduledHello1-us-central1: HTTP Error: 400, Schedule or time zone is invalid.
```

I like it, but let me know if you feel strongly about exposing function name. It'll just be less than few lines of changes I had started here.

joehan:
IMO, the schedule name should be sufficient, since it contains that function name.

joehan (approved):
LGTM

bkendall (approved):

PR #2613, opened 2020-09-14
hosting normalizing and typing

bkendall:
### Description

* normalized the name for the channels when they are provided in the channels commands and in the clone commands.
* added a couple types and use them in clone

### DIFF
--- a/src/hosting/api.ts
+++ b/src/hosting/api.ts
@@ -85,6 +85,85 @@ export interface Channel {
   labels: { [key: string]: string };
 }
 
+enum VersionStatus {
+  // The default status; should not be intentionally used.
+  VERSION_STATUS_UNSPECIFIED = "VERSION_STATUS_UNSPECIFIED",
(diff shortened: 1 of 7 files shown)


src/hosting/api.ts
```diff
@@ -85,6 +85,85 @@ export interface Channel {
...
+}
+
+/**
+ * normalizeName normalizes a name given to it. Most useful for normalizing
+ * user provided names. This removes any `/` characters and replaces them with
+ * dashes (`-`).
+ * @param s the name to normalize.
+ * @return the normalized name.
+ */
+export function normalizeName(s: string): string {
```
mbleigh, line 159: I think we should also normalize `:`, and maybe `_` as well (since underscores aren't technically valid subdomain names).

dahliasalem (approved):

src/hosting/api.ts
bkendall: done.

PR #2619, opened 2020-09-16
Update google-auth-library to v6.0.6

### LINKED ISSUE #2617: Security Vulnerability in generated functions/package-lock.json

 

**firebase-tools:** 8.10.0

**Platform:** Windows 10 2004

### [REQUIRED] Test case

I created a new project and used the firebase-tools CLI to configure some cloud functions that I intend to use; then, I push it to a github repository and activated the dependabot alerts wich created one alert saying that one of the dependencies (node-forge version 0.7.6) in the generatged package-lock.json has a known vulnerability that can be consulted in https://github.com/advisories/GHSA-92xj-mqp7-vmcj 

Link a local project with a firebase project using the firebase CLI and activate the use of functions in the setup process.

This generates a series of various files and directories that includes the package-lock.json in a new functions directory.

The dependency node-forge is forced to be 0.7.6 and cannot be automatically upgraded by dependabot after it found the vulnerability because it might conflict with other dependencies.

rohansingh:
### Description

Old versions of `google-auth-library` have a transitive
dependency on a vulnerable version of `node-forge`.

### Scenarios Tested

`npm test`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1324,6 +1324,22 @@
             "@types/node": "*"
           }
         },
+        "google-auth-library": {
+          "version": "5.10.1",
+          "resolved": "https://registry.npmjs.org/google-auth-library/-/google-auth-library-5.10.1.tgz",
+          "integrity": "sha512-rOlaok5vlpV9rSiUu5EpR0vVpc+PhN62oF4RyX/6++DG1VsaulAFEMlDYBLjJDDPI6OcNOCGAKy9UVB/3NIDXg==",
+          "requires": {
+            "arrify": "^2.0.0",
+            "base64-js": "^1.3.0",
+            "ecdsa-sig-formatter": "^1.0.11",
+            "fast-text-encoding": "^1.0.0",
+            "gaxios": "^2.1.0",
+            "gcp-metadata": "^3.4.0",
+            "gtoken": "^4.1.0",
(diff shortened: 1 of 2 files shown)


bkendall:
Unfortunately, we cannot accept this quite yet because we still support Node 8 environments. `google-auth-library >= 6` specifies node `>= 10` in the engines field.

rohansingh:
Would you consider `shrinkwrap`? Upgrading to node-forge v0.10.0 would fix the issue without breaking Node 8 compatibility.

bkendall:
We've gone back and forth on `shrinkwrap`. Using it makes a lot of sense for a tool like ours (lock the dependencies down), but we'd lose out on automatically getting patch/minor updates that may fix vulnerabilities on the user's end too. The catch is that we have to rely on other developers using semver correctly, which doesn't always happen.

Also, fwiw, I don't think this fixes #2617 - the package.json that is provided there comes from a template in this repo, not the root `package.json` of the CLI. Until `firebase-functions` updates their repo, I don't think we'll be able to directly solve #2617

rohansingh:
Ah good point, I misread the report in #2617.

bkendall:
Thanks for raising this. We've updated `google-auth-library` as part of a big dependencies fix going to `v9` of this library (#2942). Should be all set!

PR #2624, opened 2020-09-18
Adds changelog for #2621

joehan:
Forgot to add this on the original PR, whoops 😬


mbleigh:
This is internal-facing for testing only, I don't think we need it in the release notes. So you didn't forget anything!

PR #2648, opened 2020-09-28
Always export any imported namespaces

### LINKED ISSUE #2634: Empty namespaces are not exported on exit
If I click delete on the root node the data is removed in the browser.

image

Im using  `RUNNING_LOCAL=true firebase emulators:start --only database,functions  --inspect-functions   --import=./seed --export-on-exit`

however if I close the terminal and restart all data is still there. The only way I can remove the data on restart is to delete the seed file.

samtstern:
### Description

Fixes #2634 

### Scenarios Tested

1. Import non-null data
2. Delete it all
3. Confirm export is `null` after `--export-on-exit`

### Sample Commands

N/A


yuchenshi (changes requested):
This is a great change, but I still have questions about non-default namespaces.

I think given the scenario where developer import from an Export 1 with namespace A, B, C and then remove B before exporting, the CLI should always produce an Export 2 with namespace B being empty (or not present). That is to say, importing Export 2 to a fresh session should not introduce any data to namespace B.

I'd like to see tests that ensures the above, even if B is not a default namespace, and even if Export 2 happens to be the same directory as Export 1. Let's add those test cases and we can figure out what we need to change on the implementation side.  I don't think this change alone will be enough.

samtstern:
@yuchenshi thanks for all the thoughts.  Those cases were intentionally omitted because I wasn't sure how to keep track of them but I just realized I was way overthinking it, will add coverage now.

src/emulator/controller.ts
```diff
@@ -528,42 +527,7 @@ export async function startAll(options: any, noUi: boolean = false): Promise<voi
...
-              }
-            }
-          );
-
-          req.on("error", reject);
-          readStream.pipe(req, { end: true });
-        }).catch((e) => {
-          throw new FirebaseError("Error during database import.", { original: e, exit: 1 });
-        });
+        await databaseEmulator.importData(ns, fPath);
```
samtstern, line 530: This logic is moved mostly  unchanged to be inside the emulator.  Only change is to add successful imports to the internal state.

src/emulator/databaseEmulator.ts
```diff
@@ -106,6 +109,50 @@ export class DatabaseEmulator implements EmulatorInstance {
...
+          port,
+          path: `/.json?ns=${ns}&disableTriggers=true&writeSizeLimit=unlimited`,
+          headers: {
+            Authorization: "Bearer owner",
+            "Content-Type": "application/json",
+          },
+        },
+        (response) => {
+          if (response.statusCode === 200) {
+            this.importedNamespaces.push(ns);
```
samtstern, line 136: This line is the main change.

yuchenshi (approved):

scripts/triggers-end-to-end-tests/tests.ts
```diff
@@ -362,6 +362,28 @@ describe("import/export end to end", () => {
     expect(aSnap.val()).to.eql("namespace-a");
     expect(bSnap.val()).to.eql("namespace-b");
 
+    // Delete all of the import files
+    for (const f of fs.readdirSync(dbExportPath)) {
+      const fullPath = path.join(dbExportPath, f);
+      await fs.unlinkSync(fullPath);
```
yuchenshi, line 368: nit: Do we still need this in the test? Won't the files be overwritten by the export later anyway?

samtstern: Yeah the idea of unlinking them is to be sure that the export produces them again and that we're not just seeing "leftover" files.

PR #2649, opened 2020-09-28
Fix NaN when using Emulator UI defaut port

### LINKED ISSUE #2627: Emulator UI port shows as NaN on firebase init emulators

Latest `firebase-tools` GitHub master

**firebase-tools:** 8.10.0

**Platform:** ALL

### [REQUIRED] Test case

Use a new project that does not have 

1. `firebase init emulators`
2. Enable Emulator UI (default)
3. Do not select a port (default)

It displays "port" as empty or "(automatic)" after confirming a port selection

It shows "NaN" as the port number, which looks broken and tricks me into thinking something went wrong ("should I have selected a port?")

samtstern:
### Description

Fixes #2627 

### Scenarios Tested

  * Init with empty port
  * Init with custom port
  * Init with invalid port

### Sample Commands

N/A

### DIFF
--- a/src/init/features/emulators.ts
+++ b/src/init/features/emulators.ts
@@ -90,11 +90,6 @@ export async function doSetup(setup: any, config: any) {
         // Parse the input as a number
         const portNum = Number.parseInt(ui.port);
         ui.port = isNaN(portNum) ? undefined : portNum;
-
-        if (!ui.port) {
-          // Don't write `port: ""` into the config file.
-          delete ui.port;
-        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/init/features/emulators.ts
```diff
@@ -79,13 +79,18 @@ export async function doSetup(setup: any, config: any) {
...
               uiDesc
             )} (leave empty to use any available port)?`,
           },
         ]);
+
+        // Parse the input as a number
+        const portNum = Number.parseInt(ui.port);
+        ui.port = isNaN(portNum) ? undefined : portNum;
+
         if (!ui.port) {
```
yuchenshi, line 94: We can drop this now

yuchenshi (approved):

PR #2654, opened 2020-09-29
Fix crash in channels workflow setup

jhuleatt:
### Description

- Fixes bug where answering "no" to "Set up automatic deployment to your site's live channel when a PR is merged?" resulted in a crash
- Fixes bug where workflow yaml files weren't written to the repository root if the command wasn't run from repository root

### Scenarios Tested

Screen Shot 2020-09-29 at 12 22 25 PM


src/init/features/hosting/github.ts
```diff
@@ -184,9 +186,27 @@ export async function initGitHub(setup: Setup, config: any, options: any): Promi
   logLabeledBullet("Action required", `Push any new workflow file(s) to your repo`);
 }
 
+function getGitFolderPath(): string {
+  const commandDir = process.cwd();
+  let projectRootDir = commandDir;
+
+  while (!fs.existsSync(path.resolve(projectRootDir, ".git"))) {
+    const parentDir = path.dirname(projectRootDir);
```
jamesdaniels, line 194: I'm assuming this lobs off the current dirname `blah/blah` => `blah`

src/init/features/hosting/github.ts
```diff
@@ -184,9 +186,27 @@ export async function initGitHub(setup: Setup, config: any, options: any): Promi
...
+    if (parentDir === projectRootDir) {
```
jamesdaniels, line 195: And stop case when it can't go up higher? This is a little unclear, might want to add a comment.

jamesdaniels (approved):
LGTM, fix lint & maybe add comments on the while loop

src/init/features/hosting/github.ts
jamesdaniels: FYI testing Windows paths show that stopcase is `.`, e.g, `path.dirname('C://')` => `"."`, this logic might still work but I don't know if what `.` will equate too in powershell. You might get undefined behavior.

src/init/features/hosting/github.ts
jhuleatt: yeah. I modeled this after https://github.com/firebase/firebase-tools/blob/24916de2be9740968798216bebb4daf30119d3da/src/detectProjectRoot.ts

bkendall (approved):

src/init/features/hosting/github.ts
jhuleatt: added a comment to hopefully make this clearer

src/init/features/hosting/github.ts
jhuleatt: Added a comment. Since this is really similar to the code that finds `firebase.json`, I think we're ok here

PR #2663, opened 2020-10-01
Implement IDP login POC for widget+iframe for web.

yuchenshi:
### Description

This adds a proof of concept for IDP sign-ins for the web. The UI and account picker is just a stub for now, but the communication between iframe / popup / redirect should be considered fully implemented and final. Please review assuming `iframe` code will stay, and most of `handler` code will not be changed (except HTML and DOM obviously, and mobile support soon).

Mobile flows will be implemented later.

### Scenarios Tested

ONLY tested with the existing Auth JS SDK (auth-emulator branch).
Please help testing auth-exp if you have time.

Flows tested:

* Sign-in with redirect on the web
* Sign-in with popup on the web

### Sample Commands

N/A

### DIFF
--- a/src/emulator/auth/handlers.ts
+++ b/src/emulator/auth/handlers.ts
@@ -130,28 +130,203 @@ export function registerHandlers(
 
   app.get(`/emulator/auth/handler`, (req, res) => {
     res.set("Content-Type", "text/html; charset=utf-8");
+    // Note: For browser compatibility, please avoid ES6 and newer browser
+    // features as much as possible in the page below.
     res.end(
       `<!DOCTYPE html>
 <meta charset="utf-8">
-<title>Not Implemented</title>
-<h1>Not Implemented in Auth Emulator</h1>
-<p>
-  Sign-in with popup / redirect / web-views with the Auth Emulator
-  is not yet supported.
-</p>
-<p>We're working on it. Please stay tuned.</p>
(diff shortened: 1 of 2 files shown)


src/emulator/auth/handlers.ts
```diff
@@ -130,28 +130,203 @@ export function registerHandlers(
...
+      return cb('No matching frame');
+    }
+    return cb();
+  }
+
+  // DOM logic
+
+  document.querySelector('.js-provider-id').textContent = providerId;
+  var reuseAccountEls = document.querySelectorAll('.js-reuse-account');
+  [].forEach.call(reuseAccountEls, function (el) {
```
samhorlbeck, line 219: Isn't this a noop?

samhorlbeck (approved):

src/emulator/auth/handlers.ts
yuchenshi: This is a crazy hack from the old days of JavaScript, and what it really does is call `Array.prototype.forEach` but with `this` set to the NodeList. Note that `NodeList` is not an Array and doesn't have `.forEach`, thus the hack. See https://developer.mozilla.org/en-US/docs/Web/API/NodeList#Example

This hack was obsoleted by `for...of`. Modern browsers also added `.forEach` (but not universal, notably not in IE).

PR #2666, opened 2020-10-02
Allow reusing existing IDP accounts for real.

yuchenshi:
### Description

Now you can reuse existing IDP accounts with one click.

I promise I won't do the templating thing again. There's no need to do that again anyway -- I believe this is the only information needed from the backend.

### Scenarios Tested

Sign-in with credentials first and then sign-in with popup. Verify old accounts shows up in the popup window.

### Sample Commands

N/A

### DIFF
--- a/src/emulator/auth/handlers.ts
+++ b/src/emulator/auth/handlers.ts
@@ -1,7 +1,7 @@
 import { URL } from "url";
 import * as express from "express";
-import { resetPassword, setAccountInfoImpl } from "./operations";
-import { ProjectState } from "./state";
+import { IdpJwtPayload, resetPassword, setAccountInfoImpl } from "./operations";
+import { ProjectState, ProviderUserInfo } from "./state";
 import { BadRequestError, NotImplementedError } from "./errors";
 
 /**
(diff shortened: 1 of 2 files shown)


src/emulator/auth/handlers.ts
```diff
@@ -331,3 +345,21 @@ export function registerHandlers(
...
+    iss: "",
+    aud: "",
+    exp: 0,
+    iat: 0,
+    name: info.displayName,
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    screen_name: info.screenName,
+    email: info.email,
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    email_verified: true, // TODO: Shall we allow changing this?
```
samhorlbeck, line 361: Probably will want to allow people to change this; folks get confused when facebook accounts with gmail don't count as account_verified even though facebook does verify emails

samhorlbeck (approved):

src/emulator/auth/handlers.ts
yuchenshi: I'll bring with up with UX

PR #2674, opened 2020-10-06
Add command `firestore:rules:compile`

sgr-ksmt:
### Description

I introduced a new command `firebase firestore:rules:compile` so that developers can check `firestore.rules` before deploying to Firebase.
Also, they can use it for lint. For example, they embed this command into CI, and then they can check for problems with `firestore.rules` in a pull request.

#### Review note

I was torn between naming the command `firestore:rules:compile` or `firestore:rules:lint`.
If you have any good ideas, please let me know.

### Scenarios Tested

- Manually execute `firebase firestore:rules:compile`

### Sample Commands

- Usage:

```bash
$ firebase firestore:rules:compile
```

- Example results:

```bash
$ firebase firestore:rules:lint
i  cloud.firestore: checking firestore.rules for compilation errors...
✔  cloud.firestore: rules file firestore.rules compiled successfully
```

```bash
$ firebase firestore:rules:lint
i  cloud.firestore: checking firestore.rules for compilation errors...

Error: Compilation error in firestore.rules:
[E] 22:11 - Unexpected '}'.
```


sgr-ksmt:
@googlebot I signed it!

samtstern:
@sgr-ksmt thank you for submitting this!  Seems like a cool feature.

As you may know, new commands have to go through an API review which means:

1. I'll need to find a Firebaser who wants to "sponsor" your change through the internal process
2. It will take ~2 weeks once we have a sponsor to get the API approved for implementation

I'll get started on (1) now.

yuchenshi:
Thanks for the proposal! While this may be useful, we're more interested in supporting a unified `firebase deploy --dry-run` or `--validate` API that covers not only rules but also index definitions, etc. It should also work with `--only`, e.g. `firebase deploy --only firestore:rules --dry-run`. As you can see, this scales better than one dedicated command per item and allows you to validate multiple things (and the whole app) at once.

With that being said, in the short term, we think the best way for developers to 'compile' their rules is to [write a unit test](https://firebase.google.com/docs/firestore/security/test-rules-emulator). With that, you can make sure your rules compile, AND on top of that you can optionally validate that your rules do the right thing. We'd love to hear your thoughts about this!

sgr-ksmt:
@yuchenshi 

Thank you for your reply.

> With that being said, in the short term, we think the best way for developers to 'compile' their rules is to write a unit test. With that, you can make sure your rules compile, AND on top of that you can optionally validate that your rules do the right thing.

I think so, too.
But more easily than that, I wanted to know if the syntax of the rule itself is correct, so I added a new command this time.

But if you're (Firebase is) thinking of adding a `-dry-run` option in the future, I think it's best than adding a `compile` command.
If so, I think I'll close this PR and wait for the `-dry-run` option. 👍🏻

samtstern:
@sgr-ksmt thanks for understanding!  And this PR definitely helped kick off the discussion about `--dry-run` again, and now @mbleigh is going to look into it.  So even though we won't be adding `rules:compile` for now your contribution is going to lead to improvements!

kafkas:
Hey @samtstern is there any update on this? I would've expected Firebase to expose a tool to validate the rules. If the rules are managed in code there needs to be a way to syntax-check them in CI without having to deploy to Firestore.

PR #2683, opened 2020-10-08
Styling the providers list widget

mjchristy:
### Description

Adding styling to the auth account widget. 

image
image

### Scenarios Tested

No existing accounts
Some existing accounts
New account creation
Auto generate user

### DIFF
--- a/src/emulator/auth/widget_ui.ts
+++ b/src/emulator/auth/widget_ui.ts
@@ -327,6 +327,14 @@ button {
   color: #858585;
 }
 
+.subtitle {
+ color: var(--mdc-theme-text-secondary-on-background);
+ font-size: 14px;
+ line-height: 20px;
+ margin-block-end: 0em;
+ margin-block-start: 0em;
+}
+
 #content {
   box-sizing: border-box;
   width: 515px;
@@ -335,10 +343,6 @@ button {
 
 .content-wrapper {
   padding: 0 24px;
-}
-
-#add-account-button {
-  height: 56px !important;
 }
 
 .mdc-list--avatar-list .mdc-list-item {
@@ -382,6 +386,10 @@ button {
   background-size: contain;
 }
 
+#add-account-button {
+  height: 56px !important;
+}
+
 .callout {
   align-items: center;
   background: #e5eaf0;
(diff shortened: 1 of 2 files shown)


src/emulator/auth/widget_ui.ts
```diff
@@ -294,21 +327,67 @@ const STYLE = `
...
-  width: 500px;
+  width: 515px;
+  margin: 16px auto;
 }
 
-button {
-  text-transform: none !important;
+.content-wrapper {
+  padding: 0 24px;
+}
```
NothingEverHappens, line 346: nit: combine with below?

src/emulator/auth/widget_ui.ts
```diff
@@ -294,21 +327,67 @@ const STYLE = `
...
+.mdc-list--avatar-list .mdc-list-item__graphic {
+  border-radius: 50%;
+  height: 36px;
+  margin-left: 0;
+  margin-right: 16px;
+  width: 36px;
+}
+
+.mdc-list .mdc-list-item__graphic {
+  display: -ms-inline-flexbox;
```
NothingEverHappens, line 361: I don't know if we really need all this -ms prefixed stuff, do we even support ie? (I don't think we do in the console)

src/emulator/auth/widget_ui.ts
```diff
@@ -294,21 +327,67 @@ const STYLE = `
...
+}
+
+.mdc-list--avatar-list .mdc-list-item__graphic {
+  background-color: rgba(0,0,0,.3);
+  color: #fff;
+}
+
+.mdc-list-item__graphic {
+  align-items: center;
+  color: background-color: rgba(0,0,0,.3);
```
NothingEverHappens, line 372: nit/optional: consider using a var?

NothingEverHappens (approved):

src/emulator/auth/widget_ui.ts
mjchristy: Done.

src/emulator/auth/widget_ui.ts
mjchristy: Ah, copied some of this from material website. I'll drop it.

src/emulator/auth/widget_ui.ts
mjchristy: We're not really using variables, but I'm gonna go with an opaque gray.

yuchenshi:
Great work!

PR #2686, opened 2020-10-09
Fix auth handler styling for mobile devices

samhorlbeck:
Turns out it was as simple as going from `width` -> `max-width`

### DIFF
--- a/src/emulator/auth/widget_ui.ts
+++ b/src/emulator/auth/widget_ui.ts
@@ -341,6 +341,7 @@ button {
   box-sizing: border-box;
   margin: 16px auto;
   max-width: 515px;
+  min-width: 300px;
 }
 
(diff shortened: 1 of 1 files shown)


src/emulator/auth/widget_ui.ts
```diff
@@ -339,8 +339,8 @@ button {
 
 #content {
   box-sizing: border-box;
-  width: 515px;
   margin: 16px auto;
+  max-width: 515px;
```
yuchenshi, line 343: Please also add a min-width to keep it usable for extreme cases (with scrolling, but at least still usable).

```suggestion
  max-width: 515px;
  min-width: 300px;
```

I think `300px` is a good cutoff since almost all mobile phones are bigger than that. See https://viewportsizes.mattstow.com/?filter=

yuchenshi (approved):
Good stuff!

PR #2700, opened 2020-10-12
Add warning for certain generic IDP flows.

yuchenshi:
This shows a warning banner for some flows supported in Auth Emulator only but not in production. We will still let the developer proceed for testing though.

cc'd a few folks for print and UX.

Screen Shot 2020-10-12 at 12 53 18 PM

### DIFF
--- a/src/emulator/auth/widget_ui.ts
+++ b/src/emulator/auth/widget_ui.ts
@@ -38,6 +38,17 @@ assert(
   'Missing one of appName / clientId / appId / apn / ibi query params.'
 );
 
+// Warn the developer of a few flows only available in Auth Emulator.
+if ((providerId === 'facebook.com' && appIdentifier) || (providerId === 'apple.com' && ibi)) {
+  var providerName = (providerId === 'facebook.com') ? 'Facebook' : 'Apple';
+  var productionMethod = providerName + (apn ? ' Android SDK' : ' iOS SDK');
+  var warningEl = document.querySelector('.js-signin-warning');
+  warningEl.querySelector('.content').textContent =
+    'Sign-in with ' + providerName + ' via generic IDP is only supported in the Auth Emulator; ' +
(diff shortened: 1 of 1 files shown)


samhorlbeck:
Hmm when it's multiple lines, the (i) icon should be aligned with the top. Unfortunately, removing "align-items: center" makes it look bad with a single line.

Maybe put the (i) icon in a div with a height: 100%? That will keep the text center-aligned vertically but should keep the (i) at the top

yuchenshi:
> Hmm when it's multiple lines, the (i) icon should be aligned with the top. Unfortunately, removing "align-items: center" makes it look bad with a single line.
> 
> Maybe put the (i) icon in a div with a height: 100%? That will keep the text center-aligned vertically but should keep the (i) at the top

Updated code and screenshot. I think I've found a much simpler solution.

mjchristy (approved):
LGTM

yuchenshi:
FYI: Callouts with one line aren't affected.

Screen Shot 2020-10-12 at 1 39 11 PM

src/emulator/auth/widget_ui.ts
```diff
@@ -385,16 +396,21 @@ button {
...
 }
 
+.callout-warning {
+  background: #fff3e0;
+  color: #bf360c;
+}
+
 .callout .content {
   flex: 1;
+  align-self: center;
```
samhorlbeck, line 413: nit: consider alphabetizing these. I started doing this style-- we don't need to continue but worth considering

samhorlbeck (approved):

henrypbriffel (approved):

henrypbriffel (approved):

src/emulator/auth/widget_ui.ts
yuchenshi: Next time I touch this file I'll sort them all

PR #2708, opened 2020-10-14
Add error message when default database doesn't exist

sarajmunjal:
### Description
Add a helpful error message for when the default RTDB instance is not present. This can happen in any `database:*` command, in the case that the default RTDB instance was not created during project creation.

### Scenarios Tested

Commands that error out
```
firebase database:get /
firebase database:set /
firebase database:push /
firebase database:remove /
firebase database:settings:get
firebase database:settings:set / x

```

Output: https://screenshot.googleplex.com/67kdgQ2MGst27yH

Commands that have unchanged UX
```
firebase hosting:channel:list
firebase hosting:channel:deploy
firebase hosting:channel:open
firebase hosting:channel:create
firebase hosting:channel:delete

firebase deploy --only hosting
```

### DIFF
--- a/src/getDefaultHostingSite.ts
+++ b/src/getDefaultHostingSite.ts
@@ -1,0 +1,19 @@
+import * as logger from "./logger";
+import { getFirebaseProject } from "./management/projects";
+import * as _ from "lodash";
+
+/**
+ * Tries to determine the default hosting site for a project, else falls back to projectId.
+ * @param {Object} options The command-line options object
+ * @returns {Promise<String>} The hosting site ID
+ */
+export async function getDefaultHostingSite(options: any): Promise<string> {
+  const project = await getFirebaseProject(options.project);
+  if (!_.has(project, "resources.hostingSite")) {
+    logger.debug(
+      `No default hostingSite found for project: ${options.project}. Using projectId as hosting site name.`
+    );
+    return options.project;
+  }
+  return _.get(project, "resources.realtimeDatabaseInstance");
+}
--- a/src/getDefaultHostingSite.ts
+++ b/src/getDefaultHostingSite.ts
@@ -1,19 +1,19 @@
 import * as logger from "./logger";
 import { getFirebaseProject } from "./management/projects";
-import * as _ from "lodash";
 
 /**
  * Tries to determine the default hosting site for a project, else falls back to projectId.
- * @param {Object} options The command-line options object
- * @returns {Promise<String>} The hosting site ID
+ * @param options The command-line options object
+ * @return The hosting site ID
  */
 export async function getDefaultHostingSite(options: any): Promise<string> {
   const project = await getFirebaseProject(options.project);
-  if (!_.has(project, "resources.hostingSite")) {
+  const site = project.resources?.hostingSite;
+  if (!site) {
     logger.debug(
-      `No default hostingSite found for project: ${options.project}. Using projectId as hosting site name.`
+      `No default hosting site found for project: ${options.project}. Using projectId as hosting site name.`
     );
     return options.project;
   }
-  return _.get(project, "resources.realtimeDatabaseInstance");
+  return site;
 }
--- a/src/getInstanceId.js
+++ b/src/getInstanceId.js
@@ -2,7 +2,8 @@
 
 var _ = require("lodash");
 var { getFirebaseProject } = require("./management/projects");
-var logger = require("./logger");
+var { FirebaseError } = require("./error");
+var clc = require("cli-color");
 
 /**
  * Tries to determine the instance ID for the provided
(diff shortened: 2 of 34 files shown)


fredzqm (commented):
Please have someone from the CLI to confirm it is OK.
I saw that `requireAccess` uses it. `requireAccess` is used broadly in many commands.
I am afraid this might affect a broader range of commands than we intended.

src/getInstanceId.js
```diff
@@ -13,14 +14,13 @@ var logger = require("./logger");
 module.exports = function(options) {
   return getFirebaseProject(options.project).then(function(project) {
     if (!_.has(project, "resources.realtimeDatabaseInstance")) {
-      logger.debug(
-        "[WARNING] Unable to fetch default resources. Falling back to project id (" +
-          options.project +
```
fredzqm, line 18: Might be a good idea to fallback to `<project>-default-rtdb` now.

sarajmunjal: If `<project>-default-rtdb` exists, it will be present in the response for `getFirebaseProject`. I think we should remove this fallback now.

bkendall (changes requested):
This is troublesome because getInstanceId is used (via requireInstanceId) in several Hosting commands which will make it very confusing when a Hosting command fails and the error message says "you didn't create a RTDB instance". This discovery mechanism on Hosting's side is trying to find the "default site", which Hosting does not currently have it's own API for. 

@mbleigh - what do you think of this? I'm scratching my head trying to think of a way this might be separated a bit (so Hosting can discover the default site without relying on the creation of the RTDB instance). Should Hosting be searching for hosting.gapis.com/projects/project_id/sites/project_id instead?

mbleigh:
@bkendall is correct -- this is not safe as-is, since both RTDB and Hosting currently rely on this "instance" logic since they were historically tied. We need to make sure that any codepath that depends on this is separated appropriately. We can't remove the fallback because for Hosting the fallback is still a reasonable assumption, and we can't change the fallback to `-default-rtdb` because that's not correct for Hosting.

We may want to change this to something like `getDefaultResources` that does an in-memory cache of the response and separate `requireHosting` or `requireDatabase` or something in respective commands. Those names may be wrong but in general this is going to need more work to be safe.

sarajmunjal:
I like the idea of splitting this into `requireHosting` and `requireDatabase`. I imagine that the hosting `channel:*` commands will fallback to the `hostingSite` field in `defaultResources`: https://firebase.google.com/docs/projects/api/reference/rest/v1beta1/projects#defaultresources.

We could do an in-memory cache for projectDetails. Wondering why it hasn't been done before. Is it because we didn't feel the need, or because cache invalidation introduces weird cases?

sarajmunjal:
Alright I've made hosting commands fetch `getProjectDetails().defaultResources.hostingSite` and fall back to `projectId` instead of using `getProjectDetails().defaultResources.realtimeDatabaseInstance`. I tested with `hosting:channels:create` and `hosting:channels:list` and it seems to work.

Haven't implemented caching yet. Seems to be the same performance because we were anyway hitting `getProjectDetails()`, we're just relying on a different field now.

sarajmunjal:
@bkendall @mbleigh PTAL when you get the chance. Thanks!

src/getDefaultHostingSite.ts
```diff
@@ -0,0 +1,19 @@
+import * as logger from "./logger";
+import { getFirebaseProject } from "./management/projects";
+import * as _ from "lodash";
+
+/**
+ * Tries to determine the default hosting site for a project, else falls back to projectId.
+ * @param {Object} options The command-line options object
```
bkendall, line 7: In typescript, types are not necessary in jsdoc comments.

```suggestion
 * @param options The command-line options object
```

src/getDefaultHostingSite.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @returns {Promise<String>} The hosting site ID
```
bkendall, line 8:
```suggestion
 * @return The hosting site ID
```

src/getDefaultHostingSite.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export async function getDefaultHostingSite(options: any): Promise<string> {
+  const project = await getFirebaseProject(options.project);
+  if (!_.has(project, "resources.hostingSite")) {
```
bkendall, line 12: I would like to discourage the use of lodash here. The returned value from `getFirebaseProject` is well defined (`FirebaseProjectMetadata`). Could use more typescripty ways of checking for these that allow the types to be relied upon?

For `has` (untested):

```suggestion
  if (project.resources?.hostingSite) {
```

src/getDefaultHostingSite.ts
```diff
@@ -0,0 +1,19 @@
...
+    logger.debug(
+      `No default hostingSite found for project: ${options.project}. Using projectId as hosting site name.`
+    );
+    return options.project;
+  }
+  return _.get(project, "resources.realtimeDatabaseInstance");
```
bkendall, line 18: 1. Should this be returning `hostingSite`?
2. Maybe this function can drop lodash completely and save off the hosting site once:

```typescript
  const site = project.resources?.hostingSite;
  if (!site) {
    logger.debug(
      `No default hostingSite found for project: ${options.project}. Using projectId as hosting site name.`
    );
    return options.project;
  }
  return site;
```

bkendall (changes requested):

bkendall (approved):

PR #2706, opened 2020-10-13
Implement endpoint to enable/disable functions triggers

samtstern:
### Description

Googlers: See API proposal [here](https://docs.google.com/document/d/1mBOYZKJv6-S1Rfzb25FVkv3yJSjrXeFykeCclQZwGiM/edit?usp=sharing)

SDK Implementation: https://github.com/firebase/firebase-js-sdk/pull/3928

### Scenarios Tested

Manually tested with/without background triggers enabled

### Sample Commands

N/A

PR #2711, opened 2020-10-14
Make sure to return after sending error

samtstern:
### Description

Follow up to #2706 which adds some missing `return`

### Scenarios Tested

Discovered when e2e testing

### Sample Commands

N/A


yuchenshi (approved):

henrypbriffel (commented):
Appreciate the work

PR #2713, opened 2020-10-14
Update function deploys to default to node 12

joehan:
### Description
Updates `firebase init` to use node 12 as the default for functions, and updates the other firebase dependencies  in to their latest versions. Also adds more in depth warning message for node 8 deploys laying out the full timeline of when deploys/executions will stop working.

### Scenarios Tested
Deploying a node 8 function:
Screen Shot 2020-10-14 at 12 46 10 PM

I also deployed the default helloWorld function from 'firebase init' with the new dependencies, and it is working correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 - Updates the Google Cloud Run proxy API calls to `v1` (from `v1alpha1`) (#2695).
 - Release RTDB emulator v4.6.0: Get wire protocol with optional query.
-- Updates Cloud Functions for Firebase templates to better support function development.
+- Updates Cloud Functions for Firebase templates to use Node 12 and better support function development.
 - Release Firestore emulator v1.11.9: Fixes != and not-in operators.
 - Add endpoints to enable/disable background triggers in the Cloud Functions emulator.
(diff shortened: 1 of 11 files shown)


joehan:
@mbleigh Yep, I was just able to repro that error on a new project without a billing account: 
Screen Shot 2020-10-15 at 10 23 43 AM

The errors happening before we ever get to IsBillingError - its failing when we try to enable CloudBuild. I think the best way to handle this is to check for billing errors in EnsureApi, and print out a generic error containing a console link to go upgrade. WDYT about this:
```
This command requires the following APIs to be enabled: {list of apis from original error}, which require the pay-as-you-go (Blaze) billing plan. To upgrade your project, visit the following URL:
      
https://console.firebase.google.com/project/${projectId}/usage/details
```
This would cover this case, and any similar errors for other APIs in the future.

mbleigh:
I like it, but let's change up the order a bit:

```
Error: Your project (${projectId}) must be on the Blaze (pay-as-you-go) plan to complete this command. Required API ${apiName} can't be enabled until the upgrade is complete. To upgrade, visit the following URL:

https://console.firebase.google.com/project/${projectId}/usage/details
```

joehan:
Addressed the billing error in https://github.com/firebase/firebase-tools/pull/2715

mbleigh (approved):

PR #2714, opened 2020-10-15
Update copy text for no accounts and styling providerId.

yuchenshi:
### Description

Two changes: `google.com` => `Google.com`, and update help text when empty.

Screen Shot 2020-10-14 at 5 20 24 PM

### Scenarios Tested

Manually started and verified (see screenshot above).

### DIFF
--- a/src/emulator/auth/widget_ui.ts
+++ b/src/emulator/auth/widget_ui.ts
@@ -142,15 +142,23 @@ function redirectToIos(authEvent) {
 
 // DOM logic
 
-document.querySelectorAll('.js-provider-id').forEach(function(e) {e.textContent = providerId});
-var reuseAccountEls = document.querySelectorAll('.js-reuse-account');
-[].forEach.call(reuseAccountEls, function (el) {
-  var urlEncodedIdToken = el.dataset.idToken;
-  el.addEventListener('click', function (e) {
-    e.preventDefault();
-    finishWithUser(urlEncodedIdToken);
-  });
+var formattedProviderId = providerId[0].toUpperCase() + providerId.substring(1);
+document.querySelectorAll('.js-provider-id').forEach(function(e) {
+  e.textContent = formattedProviderId;
 });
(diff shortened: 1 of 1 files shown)


mjchristy:
Thanks Yuchen. Looks great.

mjchristy (approved):
Looks great, thanks.

src/emulator/auth/widget_ui.ts
```diff
@@ -142,15 +142,23 @@ function redirectToIos(authEvent) {
...
+});
+
 var reuseAccountEls = document.querySelectorAll('.js-reuse-account');
-[].forEach.call(reuseAccountEls, function (el) {
-  var urlEncodedIdToken = el.dataset.idToken;
-  el.addEventListener('click', function (e) {
-    e.preventDefault();
-    finishWithUser(urlEncodedIdToken);
+if (reuseAccountEls.length) {
+  [].forEach.call(reuseAccountEls, function (el) {
```
samhorlbeck, line 152: Out of curiosity, why are you doing
```js
[].forEach.call(reuseAccountEls, ...
```
instead of
```js
reuseAccountEls.forEach(...
```

samhorlbeck (approved):

src/emulator/auth/widget_ui.ts
samhorlbeck: Apparently I have the memory of a goldfish

NothingEverHappens: consider using `Array.from(reuseAccountEls)`, (or even `[...reuseAccountEls]`) for better readability

NothingEverHappens (approved):

src/emulator/auth/widget_ui.ts
yuchenshi: Sorry but no ES6 for now. In the long term we may even get a transpiler, but not now.

PR #2715, opened 2020-10-15
Add more actionable error message for billing errors when enabling apis

joehan:
### Description
Adds a more actionable error message when a user tries to enable an API that requires billing, without billing enabled on that project.

### Scenarios Tested
Trying to deploy a Node 12 Cloud Function that requires billing when billing is not enabled.
Screen Shot 2020-10-15 at 12 37 03 PM

### DIFF
--- a/src/ensureApiEnabled.ts
+++ b/src/ensureApiEnabled.ts
@@ -4,7 +4,7 @@ import { bold } from "cli-color";
 import * as track from "./track";
 import * as api from "./api";
 import * as utils from "./utils";
-import { FirebaseError } from "./error";
+import { FirebaseError, isBillingError } from "./error";
 
 export const POLL_SETTINGS = {
   pollInterval: 10000,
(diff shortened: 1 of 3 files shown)


src/error.ts
```diff
@@ -32,3 +32,28 @@ export class FirebaseError extends Error {
...
+          violations?: { type: string }[];
+        }[];
+      };
+    };
+  };
+}): boolean {
+  return !!e.context?.body?.error?.details?.find((d) => {
+    return (
+      d.violations?.find((v) => v.type === "serviceusage/billing-enabled") ||
+      d.reason == "UREQ_PROJECT_BILLING_NOT_FOUND"
```
joehan, line 56: Added this check because the violation returned when failing to enable an API is not specific to billing issues - it looks like:
```
[ { type: 'googleapis.com',
    subject:
     '?error_code=390001&project=1747788017&services=cloudbuild.googleapis.com&services=containerregistry.googleapis.com' } ]
```

src/ensureApiEnabled.ts
```diff
@@ -43,10 +43,23 @@ export async function check(
  * @param apiName The name of the API e.g. `someapi.googleapis.com`.
  */
 export async function enable(projectId: string, apiName: string): Promise<void> {
-  return api.request("POST", `/v1/projects/${projectId}/services/${apiName}:enable`, {
-    auth: true,
-    origin: api.serviceUsageOrigin,
-  });
+  return api
```
mbleigh, line 46: Please use `try/catch` with `await` here to be consistent with current practices.

mbleigh (approved):
One thing to fix, sorry I forgot this was waiting on me!

PR #2723, opened 2020-10-20
Fix default arguments and option when using as a module

### LINKED ISSUE #322: expand the documentation on how to use firebase-tools as a module
A lot of the commands you can do via `CLI` are very hard to figure out how to do programmatically when loading it as a module. Updated documentation on syntax, available commands, etc, would be helpful.

Questions I have (or had but soluved):
- how to print a list of all keys using `tools.functions.config.get`
- how to deploy functions
- how to get a list of all functions deployed
    > note on this one: had to really dig into cloudfunctions.js, api.js and auth.js. turns out this functionality is available inside the package, just not surfaced in firebase-tools. this is really helpful.

### LINKED ISSUE #2672: Using "apps.list" as a Module returns error "cannot read property 'project' of undefined"

**firebase-tools:**

firebase-tools : 8.11.2
firebase version : 8.10.0

**Platform:**
macOS

### [REQUIRED] Test case

```
var client = require("firebase-tools");
client
        .use("projectId", { add: true })
        .then(function (data) {
            client.apps.list()
                .then(function (data) {
                    console.log(data);
                })
                .catch(function (err) {
                    console.log(err)
                    // handle error
                });
        })
        .catch(function (err) {
            console.log(err)
        });

```

Call above code from CLI.

I expect that when calling use() with projectId, and then call apps.list it would print apps list in current active project.

TypeError: Cannot read property 'project' of undefined
    at module.exports (/Volumes/Data/Folder test/node_modules/firebase-tools/lib/getProjectId.js:7:18)
    at /Volumes/Data/Folder test/node_modules/firebase-tools/lib/commands/apps-list.js:45:23
    at Generator.next (<anonymous>)
    at /Volumes/Data/Folder test/node_modules/firebase-tools/lib/commands/apps-list.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/Volumes/Data/Folder test/node_modules/firebase-tools/lib/commands/apps-list.js:4:12)
    at Command.actionFn (/Volumes/Data/Folder test/node_modules/firebase-tools/lib/commands/apps-list.js:44:41)
    at Command.<anonymous> (/Volumes/Data/Folder test/node_modules/firebase-tools/lib/command.js:166:25)
    at Generator.next (<anonymous>)
    at fulfilled (/Volumes/Data/Folder test/node_modules/firebase-tools/lib/command.js:5:58)

samtstern:
### Description

Fixes #2672
Fixes #322

### Scenarios Tested

```ts
const client = require("firebase-tools");

async function main() {
    await client.use("fir-dumpster", { add: true });

    // This did not previously work!
    const apps = await client.apps.list();
    console.log(apps);
}

main().catch(console.warn);
```

### Sample Commands

N/A

### DIFF
--- a/scripts/client-integration-tests/tests.ts
+++ b/scripts/client-integration-tests/tests.ts
@@ -58,6 +58,11 @@ describe("apps:list", () => {
       project: process.env.FBTOOLS_TARGET_PROJECT,
     });
     expect(undefinedArgsApps).to.have.length.greaterThan(0);
+
+    const nullArgsApps = await client.apps.list(null, {
+      project: process.env.FBTOOLS_TARGET_PROJECT,
+    });
+    expect(nullArgsApps).to.have.length.greaterThan(0);
   });
 
   it("should list apps configuration", async () => {
--- a/scripts/client-integration-tests/tests.ts
+++ b/scripts/client-integration-tests/tests.ts
@@ -56,6 +56,9 @@ describe("apps:list", () => {
 
     const undefinedArgsApps = await client.apps.list(undefined, { project: process.env.FBTOOLS_TARGET_PROJECT });
     expect(undefinedArgsApps).to.have.length.greaterThan(0);
+
+    const nullArgsApps = await client.apps.list(null, { project: process.env.FBTOOLS_TARGET_PROJECT });
+    expect(nullArgsApps).to.have.length.greaterThan(0);
   });
 
   it("should list apps configuration", async () => {
--- a/scripts/client-integration-tests/tests.ts
+++ b/scripts/client-integration-tests/tests.ts
@@ -58,11 +58,6 @@ describe("apps:list", () => {
       project: process.env.FBTOOLS_TARGET_PROJECT,
     });
     expect(undefinedArgsApps).to.have.length.greaterThan(0);
-
-    const nullArgsApps = await client.apps.list(null, {
-      project: process.env.FBTOOLS_TARGET_PROJECT,
-    });
-    expect(nullArgsApps).to.have.length.greaterThan(0);
   });
 
   it("should list apps configuration", async () => {
--- a/scripts/client-integration-tests/tests.ts
+++ b/scripts/client-integration-tests/tests.ts
@@ -50,6 +50,14 @@ describe("deployHosting", () => {
 });
 
 describe("apps:list", () => {
+  it("should be able to list apps with missing or undefined optional arguments", async () => {
+    const noArgsApps = await client.apps.list({ project: process.env.FBTOOLS_TARGET_PROJECT });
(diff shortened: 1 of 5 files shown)


src/command.ts
```diff
@@ -280,10 +284,18 @@ export class Command {
   runner(): (...a: any[]) => Promise<void> {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     return async (...args: any[]) => {
-      // always provide at least an empty object for options
-      if (args.length === 0) {
+      // Make sure the last argument is an object for options, add {} if none
+      if (typeof last(args) !== "object" || last(args) === null) {
```
bkendall, line 288: In what case will `last(args)` be (exactly) `null`? `last` returns `undefined` if there's nothing: https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L7609

Should the second condition be `last(args) === undefined`? Or maybe just omitted, since `undefined` (or `null` for that matter) won't be an object?

bkendall (changes requested):
This change seems like a really good candidate for a couple new tests in the client-integration folder. Let's add some tests there to validate and ensure this is working as expected? https://github.com/firebase/firebase-tools/blob/master/scripts/client-integration-tests/tests.ts

src/command.ts
samtstern: `typeof null === 'object'` and technically someone could do `firebase.apps.list(null)` so I wanted to defend against thinking that `null` was `{}`

bkendall: Oh you're right. Stupid javascript...

samtstern:
@bkendall added some tests

scripts/client-integration-tests/tests.ts
```diff
@@ -50,6 +50,21 @@ describe("deployHosting", () => {
...
+    const undefinedArgsApps = await client.apps.list(undefined, {
+      project: process.env.FBTOOLS_TARGET_PROJECT,
+    });
+    expect(undefinedArgsApps).to.have.length.greaterThan(0);
+
+    const nullArgsApps = await client.apps.list(null, {
+      project: process.env.FBTOOLS_TARGET_PROJECT,
+    });
+    expect(nullArgsApps).to.have.length.greaterThan(0);
+  });
```
bkendall, line 66: These are awesome! Thanks!

bkendall (approved):

bkendall:
It looks like you might have an issue in the tests, though:

```
  1) apps:list
       should be able to list apps with missing or undefined optional arguments:
     TypeError: Cannot read property 'toUpperCase' of null
      at Object.getAppPlatform (src/management/apps.ts:73:20)
      at /home/runner/work/firebase-tools/firebase-tools/src/commands/apps-list.ts:44:27
```

PR #2730, opened 2020-10-21
Fix the issue that users can still sign-in when disabled.

yuchenshi:
This time for real.

### DIFF
--- a/src/test/emulators/auth.spec.ts
+++ b/src/test/emulators/auth.spec.ts
@@ -451,6 +451,20 @@ describe("Auth emulator", () => {
         });
     });
 
+    it("should error if account to be linked is disabled", async () => {
+      const { idToken, localId } = await registerAnonUser(authApp);
+      await updateAccountByLocalId(authApp, localId, { disableUser: true });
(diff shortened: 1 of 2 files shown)


src/test/emulators/auth.spec.ts
```diff
@@ -1240,6 +1269,31 @@ describe("Auth emulator", () => {
...
+        .send({ idToken, displayName: "Foo" })
+        .then((res) => {
+          expectStatusCode(400, res);
+          expect(res.body.error)
+            .to.have.property("message")
+            .equals("USER_DISABLED");
+        });
+    });
+
+    it("should still update user dispite user is disabled when authenticated", async () => {
```
avolkovi, line 1288: typo: despite

avolkovi (approved):

src/test/emulators/auth.spec.ts
yuchenshi:
```suggestion
    it("should still update user despite user is disabled when authenticated", async () => {
```

PR #2733, opened 2020-10-22
allow --only hosting:site with bare hosting configs without targets

### LINKED ISSUE #2731: Error: Hosting target (project-name) not detected in firebase.json

**firebase-tools:**v8.13.1

**Platform:**macOS

### [REQUIRED] Test case

`firebase deploy --only hosting:(project-name)`

I'm using the following Github action: https://github.com/w9jds/firebase-action/ which uses the latest release of firebase-tools.
Since the last update, it's not working anymore. I had to manually set to an older version of firebase-tools (I went for v8.12.1, I haven't tried with v8.13.0)

Successful deploy

Deploy error
Error: Hosting target (project-name) not detected in firebase.json

bkendall:
### Description

Fixes #2731 

In `v8.12.1`, we did not do any validation of `target` in `--only hosting:<target>`, but _did_ allow the ability to specify `--only hosting:<site-name>` because of how the filtering logic played out when a user had a single Hosting config (think immediately post `init hosting`).

In #2704, we started validating that `target` was a valid Hosting target, as our documentation states that `target` must match a `target` in `firebase.json`. This is well meaning, but it broke the case where the Hosting site could be specified by name.

Specifying a Hosting `target` is intended for [multi-site setups, described in the documentation](https://firebase.google.com/docs/hosting/multisites#set_up_deploy_targets). However, we also want to maintain support for specifying the `site` in a multisite configuration (rather than `target`).

### Scenarios Tested

`firebase deploy` with
- nothing
  - on a basic config
    - with a `target` specified in firebase.json
  - on a multi-site config
   - with `site` instead of `target`
- `--only hosting`
  - on a basic config
    - with a `target` specified in firebase.json
  - on a multi-site config
   - with `site` instead of `target`
- `--only hosting:<sitename>`
  - on a basic config (works)
    - with a `target` specified in firebase.json (fails)
  - on a multi-site config (fails)
   - with `site` instead of `target` (works, since it's the site)
- `only hosting:<target>`
  - on a basic config (fails, no target)
    - with a `target` specified in firebase.json (works)
  - on a multi-site config (works)
   - with `site` instead of `target` (fails, since it's not a target)

I've also added new tests for `normalizedHostingConfigs` to help solidify this behavior (yay 100% test coverage).

cc @thomasmburke

### DIFF
--- a/src/hosting/normalizedHostingConfigs.ts
+++ b/src/hosting/normalizedHostingConfigs.ts
@@ -1,18 +1,13 @@
 import { bold } from "cli-color";
 
 import { FirebaseError } from "../error";
-import logger = require("../logger");
 
 interface HostingConfig {
   site: string;
   target: string;
 }
 
-function filterOnly(
-  configs: HostingConfig[],
-  onlyString: string,
-  defaultSite: string
-): HostingConfig[] {
+function filterOnly(configs: HostingConfig[], onlyString: string): HostingConfig[] {
   if (!onlyString) {
     return configs;
   }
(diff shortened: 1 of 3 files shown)


src/hosting/normalizedHostingConfigs.ts
```diff
@@ -24,16 +24,35 @@ function filterOnly(configs: HostingConfig[], onlyString: string): HostingConfig
...
+      configsByTarget.set(c.target, c);
+    }
+  }
+
+  const filteredConfigs: HostingConfig[] = [];
+  // Check to see that all the hosting deploy targets exist in the hosting
+  // config as either `site`s or `target`s.
+  for (const onlyTarget of onlyTargets) {
+    if (configsBySite.has(onlyTarget)) {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
mbleigh, line 43: Why is this needed?

mbleigh (approved):

src/hosting/normalizedHostingConfigs.ts
bkendall: Oh! Yeah, so Typescript doesn't know that if you do `if (Map.has(thing)) Map.get(thing)`, the `get` is guaranteed to exist (and have the type as defined by the Map): https://github.com/microsoft/TypeScript/issues/9619

PR #2740, opened 2020-10-23
Swap emulator Node runtime discovery to favor local cache

abeisgoat:
Right now we always assume the healthiest node install you have is the global one, this isn't the case when running in firepit, so I've swapped the preference to prefer any version of node installed to `node_modules` (if it exists and matches the requested runtime version).


abeisgoat:
This merge looks super ugly, I literally just moved one `if` above the other

src/emulator/functionsEmulator.ts
```diff
@@ -672,6 +672,18 @@ export class FunctionsEmulator implements EmulatorInstance {
       disabled_features: this.args.disabledRuntimeFeatures,
     };
   }
+  /**
+   * Returns a node major version ("10", "8") or null
+   * @param frb the current Functions Runtime Bundle
+   */
+  getRequestedNodeRuntimeVersion(frb: FunctionsRuntimeBundle): string | null {
```
samtstern, line 679: Is there a reason this has to be `string | null` not `string | undefined`?  If you return `undefined` in the case of nothing then it's just:
```js
return frb.nodeMajorVersion || (pkg.engines && pkg.engines.node)
```

If you do need `null` then:
```js
const pkg = require(path.join(frb.cwd, "package.json"));
return frb.nodeMajorVersion || (pkg.engines && pkg.engines.node) || null
```

But this is all optional, method is clearly correct as-is just maybe overly complex?

samtstern (approved):
Small nits but LGTM

samtstern:
@abeisgoat don't forget a `CHANGELOG` entry

PR #2741, opened 2020-10-24
feat(serve): add --inspect flag

aleclarson:
### Description

Add an `--inspect` flag to `firebase serve`.

When defined, the `debugPort` of `FunctionsEmulator` becomes 9229 or the specified port (eg: `--inspect=9230`).

### Scenarios Tested

N/A

### Sample Commands

```sh
firebase serve --inspect
firebase serve --inspect=9230
```

### DIFF
--- a/src/commands/serve.js
+++ b/src/commands/serve.js
@@ -41,6 +41,7 @@ module.exports = new Command("serve")
     "--except <targets>",
     "serve all except specified targets (valid targets are: " + VALID_TARGETS.join(", ") + ")"
   )
+  .option("--inspect [port]", "attach the node.js debugger")
   .before((options) => {
     if (
       options.only &&
--- a/src/serve/functions.ts
+++ b/src/serve/functions.ts
@@ -23,6 +23,12 @@ module.exports = {
       projectId,
       functionsDir,
       nodeMajorVersion: parseRuntimeVersion(options.config.get("functions.runtime")),
+      debugPort:
+        typeof options.inspect === "string"
+          ? Number(options.inspect)
+          : options.inspect
+          ? 9229
+          : undefined,
 
       ...(args as object),
     };


samtstern:
@bkendall thoughts on this?  I know we're not supposed to be improving or investing in `serve` anymore but this seems harmless, and there are a few Hosting edge cases where `serve` is still superior to `emulators:start`

@aleclarson thanks for this PR, a few comments:

1. Please make this flag `--inspect-functions` to match what we have with `emulators:start`
2. Is there a reason you still prefer to use `serve` and not the `emulators:` commands?

bkendall:
My preference would be to keep this in `emulators:start` and not add it to `serve`. I too would like to know about (2) in Sam's comment. 😄

aleclarson:
2. Is there documentation on the differences between them? I'm using `serve` because I don't understand why switching is necessary.

samtstern:
@aleclarson the difference is that `serve` starts up individual emulators which are mostly not connected to each other. It was designed for Hosting, then we added on Functions, and then we added on the two database emulators.  We consider `serve` to be pretty much deprecated, we just don't want to break it because so many Hosting developers rely on it.

`emulators:start` does everything `serve` does and more, such as:

- Connects all the emulators together so you can, for example, trigger local Cloud Functions from the Realtime Database emulator
- Centralizes emulator configuration in `firebase.json` through `firebase init emulators`
- Starts up a local UI which allows you to visually interact with the emulators
- Supports importing/exporting data and other complex flows
- .... more!

So basically in  99/100 cases if you just replace `firebase serve` with `firebase emulators:start` everything will keep working and you'll get some great new features as well.

aleclarson:
Why's there no `--port` flag for `emulators:start`?

samtstern:
@aleclarson because there are many emulators a single `--port` flag no longer makes sense.  That's all handled through `firebase.json` now.  So if you want to change the Hosting emulator port:

**firebase.json**
```js
{
  "emulators": {
    "hosting": {
      "port": 3003 // Example
    }
}
```

PR #2753, opened 2020-10-28
Automatically call useEmulator in Hosting emulator

samtstern:
### Description

This PR changes the dynamic `/__/firebase/init.js` file in the hosting emulator to automatically call `useEmulator()` for each running emulator when the `?useEmulator=true` query param is passed.  This makes that parameter the default going forward.

This is *NOT* a breaking change because it only affects people who run `firebase init hosting` in the future.  Those who have already initialized hosting will not see any change in behavior.  It's also easy to undo by removing `useEmulator=true` from `index.html`.

#### Alternative proposal

We could also allow the user to configure this through `firebase.json` through something like:
```json
{
  "emulators": {
    "hosting": {
      "connectSdk": true
    }
}
```

My preference is to not add this config as it would be inconsistent with our approach in the Functions emulator.  In the Functions emulator the Admin SDK automatically connects to any running emulators.  To opt out of this, your only choice is to not run those emulators.  We have not received complaints about this.

I think hosting should behave the same way.  If you don't want to connect to the RTDB emulator from your website, don't run it.  The `useEmulator=false` option is a small concession to users who are for some reason used to the current behavior. 

The only reason we never did this before is that before the `useEmulator` APIs were available there was no simple way to do this.  RTDB had to be pointed at the emulator during `initializeApp()` and all the other SDKs had to be pointed at the emulator after `initializeApp()`,  making it very hard to undo.

### Scenarios Tested

Tested the following URLs:

- http://localhost:5000/__/firebase/init.js?useEmulator=true
- http://localhost:5000/__/firebase/init.js?useEmulator=false
- http://localhost:5000/__/firebase/init.js?useEmulator=foo
- http://localhost:5000/__/firebase/init.js

I also tested `__firebase/init.js?useEmulator=true` in production to ensure that it does not break.

When all emulators are running here's how it looks:
Screen Shot 2020-10-29 at 12 18 47 PM

### Sample Commands

N/A


templates/hosting/initEmulators.js
```diff
@@ -0,0 +1,31 @@
...
+  if (firebaseEmulators.firestore && typeof firebase.firestore === 'function') {
+    firebase.firestore().useEmulator(firebaseEmulators.firestore.host, firebaseEmulators.firestore.port);
+  }
+
+  if (firebaseEmulators.functions && typeof firebase.functions === 'function') {
+    firebase.functions().useEmulator(firebaseEmulators.functions.host, firebaseEmulators.functions.port);
+  }
+
+  if (firebaseEmulators.auth && typeof firebase.auth === 'function') {
+    firebase.auth().useEmulator('http://' + firebaseEmulators.auth.host + ':' + firebaseEmulators.auth.port);
```
yuchenshi, line 29: Please quote IPv6 hosts like `::1` => `[::1]`

templates/init/hosting/index.html
```diff
@@ -10,13 +10,17 @@
...
+    <script defer src="/__/firebase/{{VERSION}}/firebase-functions.js"></script>
     <script defer src="/__/firebase/{{VERSION}}/firebase-messaging.js"></script>
     <script defer src="/__/firebase/{{VERSION}}/firebase-storage.js"></script>
     <script defer src="/__/firebase/{{VERSION}}/firebase-analytics.js"></script>
     <script defer src="/__/firebase/{{VERSION}}/firebase-remote-config.js"></script>
     <script defer src="/__/firebase/{{VERSION}}/firebase-performance.js"></script>
     <!-- initialize the SDK after all desired features are loaded -->
     <script defer src="/__/firebase/init.js"></script>
+    <!-- connect to any running emulators -->
+    <script defer src="/__/firebase/initEmulators.js"></script>
```
yuchenshi, line 23: But this will cause an error in production when developers deploy it. Remembering to remove before deploy and add it back after deploy is really confusing.

Can we do something like `/__/firebase/init.js?useEmulatorsIfAvailable=true` instead and then remove that flag in the next CLI major release?

yuchenshi (changes requested):

templates/hosting/initEmulators.js
samtstern: Can you explain this one a little more?  Not sure how IPv6 works

templates/init/hosting/index.html
samtstern: Oh wow great point!  Ok I will do that.

templates/hosting/initEmulators.js
yuchenshi: Basically when constructing URLs, if the hostname contains `:`, the entire hostname must be quoted around `[]` to disambiguate from `:port`. In other words, host `::1` and port `9009` should be encoded as `http://[::1]:9099/`.

This is one of the other reasons why I don't like the API taking URLs, but we gotta do what we have to do.

templates/init/hosting/index.html
samtstern: Changed and expanded PR description to include alternate proposal.

samtstern:
Tests are failing on deployed hosting:
```
Expected Thu Oct 29 12:35:29 UTC 2020 to equal Thu Oct 29 12:35:12 UTC 2020.
```
Not sure if this could be my fault?

templates/init/hosting/index.html
```diff
@@ -10,13 +10,18 @@
...
     <script defer src="/__/firebase/{{VERSION}}/firebase-analytics.js"></script>
     <script defer src="/__/firebase/{{VERSION}}/firebase-remote-config.js"></script>
     <script defer src="/__/firebase/{{VERSION}}/firebase-performance.js"></script>
-    <!-- initialize the SDK after all desired features are loaded -->
-    <script defer src="/__/firebase/init.js"></script>
+    <!-- 
+      initialize the SDK after all desired features are loaded, set useEmulator to false
+      to avoid connecting the SDK to running emulators.
+    -->
+    <script defer src="/__/firebase/init.js?useEmulator=true"></script>
```
bkendall, line 24: I'm not sure `useEmulator=true` here will be okay. What happens when this file is deployed to a production (on real Hosting) environment? Well, I guess nothing, right? That `init.js` knows nothing about this... is this functionally a noop?

src/hosting/implicitInit.ts
```diff
@@ -54,8 +62,31 @@ export async function implicitInit(options: any): Promise<TemplateServerResponse
...
+      const host = info.host.includes(":") ? `[${info.host}]` : info.host;
+      emulators[e] = {
+        host,
+        port: info.port,
+      };
+    }
+  }
+  const emulatorsJson = JSON.stringify(emulators, null, 2);
+
+  const js = INIT_TEMPLATE
```
bkendall, line 81: Please run the formatter.

templates/hosting/init.js
```diff
@@ -1,5 +1,31 @@
 if (typeof firebase === 'undefined') throw new Error('hosting/init-error: Firebase SDK not detected. You must include it before /__/firebase/init.js');
 /*--CONFIG--*/
+/*--EMULATORS--*/
```
bkendall, line 3: Should this go with the `if (firebaseEmulators)` line?

bkendall (changes requested):

bkendall:
I was having all sorts of trouble with that Hosting test the other day, but mine was that it was constantly off by a single second. It was very weird and I'm not sure how to improve it.

templates/init/hosting/index.html
samtstern: Yeah in production to my eyes it looks like Hosting just ignores it ... but I was hoping you could confirm that by looking into the implementation (or tell me this is a bad idea and I am relying on something I shouldn't)

yuchenshi (approved):

src/hosting/implicitInit.ts
samtstern: Wow the formatter makes this awful haha but I gotta stan prettier

templates/hosting/init.js
samtstern: Yep

templates/init/hosting/index.html
samtstern: @bkendall I added some tests to the hosting integration test to verify this, so at least if production breaks this we'd have a canary.

samtstern:
@bkendall I realized why the hosting test is broken!  It's because this integration test is now running twice in parallel (Node 8 and Node 12) and they are clobbering each other!

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
 - Fixes Cloud Function inspection when using standalone binary release (#2740)
 - Release RTDB emulator v4.6.1: Fix emulator crash on invalid `.validate` rules (#2734)
 - Fixes lookup empty results using Admin SDK in Auth Emulator.
+- Automatically connect to running emulators from the Hosting emulator.
```
bkendall, line 4:
```suggestion
- Automatically connects to other running emulators from the Firebase Hosting emulator.
```

bkendall (approved):

PR #2754, opened 2020-10-28
Add e2e tests for the Auth emulator

samtstern:
### Description

Follow up to https://github.com/firebase/firebase-tools/pull/2727

To properly test this we need Admin SDK 9+ which means Node 10+ which means I had to do some funky business.

### Scenarios Tested

The whole PR is tests

### Sample Commands

N/A


yuchenshi (approved):
Thanks for adding test coverage!

scripts/npm-link.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -e
+
+if [ "$CI" = "true" ]; then
```
bkendall, line 4: If you don't care what CI is set to (just that it isn't empty), you could use `-z`:

```bash
if [ ! -z "$CI" ]; then
```

bkendall (approved):

PR #2762, opened 2020-10-29
apiv2

bkendall:
### Description

Creating a new `apiv2` package that's in typescript and will allow us to change out the current `api` package (based on `request`). This is built using `node-fetch` which is fairly light-weight.

The exported `Client` module has a generic `request` method for... generic requesting, but also has 'verb' methods matching various HTTP verbs to make most requests very simple to write.

### Scenarios Tested

I've kept this to the `hosting:channel:*` commands and have been using them as my tests. Seems to be working quite well!

- `hosting:channel:list` works
- `hosting:channel:create` works (and with `--expires`)
- `hosting:channel:deploy` works
- `hosting:channel:delete` works


src/apiv2.ts
```diff
@@ -0,0 +1,270 @@
...
+import { FirebaseError } from "./error";
+import * as logger from "./logger";
+import * as responseToError from "./responseToError";
+import * as scopes from "./scopes";
+
+const CLI_VERSION = require("../package.json").version;
+
+const VALID_METHODS = new Set(["GET", "PUT", "POST", "DELETE", "PATCH"]);
+
+type FirebaseRequestOptions = {
```
bkendall, line 12: This _could_ be a generic to type `json`, but that _might_ be overkill.

src/apiv2.ts
```diff
@@ -0,0 +1,272 @@
...
+  responseType?: "json";
+  searchParams?: { [key: string]: string | number };
+};
+
+interface FirebaseRequestHandlingOptions {
+  auth?: boolean;
+  log?: {
+    searchParams?: boolean;
+    body?: boolean;
+    resBody?: boolean;
```
mbleigh, line 27: Don't these all default to `true`? Are you proposing to default them to `false`? Otherwise we should avoid having `true` by omission options.

src/apiv2.ts
```diff
@@ -0,0 +1,272 @@
...
+const VALID_METHODS = new Set(["GET", "PUT", "POST", "DELETE", "PATCH"]);
+
+type FirebaseRequestOptions = {
+  method?: string;
+  baseURL: string;
+  path: string;
+  headers?: { [key: string]: string };
+  json?: any; // eslint-disable-line @typescript-eslint/no-explicit-any
+  responseType?: "json";
+  searchParams?: { [key: string]: string | number };
```
mbleigh, line 19: Prefer `queryParams` or just `query` (even if that doesn't match `fetch` semantics, I think).

src/apiv2.ts
```diff
@@ -0,0 +1,272 @@
...
+const CLI_VERSION = require("../package.json").version;
+
+const VALID_METHODS = new Set(["GET", "PUT", "POST", "DELETE", "PATCH"]);
+
+type FirebaseRequestOptions = {
+  method?: string;
+  baseURL: string;
+  path: string;
+  headers?: { [key: string]: string };
+  json?: any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
mbleigh, line 17: Should this be a generic as well? Should we end up with something like:

    request<ReqType = any, ResType = any>(...)

that would allow us to have strong typing on both requests and responses?

src/apiv2.ts
```diff
@@ -0,0 +1,272 @@
...
+import * as responseToError from "./responseToError";
+import * as scopes from "./scopes";
+
+const CLI_VERSION = require("../package.json").version;
+
+const VALID_METHODS = new Set(["GET", "PUT", "POST", "DELETE", "PATCH"]);
+
+type FirebaseRequestOptions = {
+  method?: string;
+  baseURL: string;
```
mbleigh, line 14: How about just `urlPrefix` (assuming you don't want to keep `origin` because *technically* it's not always set to an origin?

mbleigh (commented):
I think creating a parallel package is a great idea (allows us to migrate gradually without worrying about getting it 100% right all at once). But if we're gonna do that, feels like we can kinda just go nuts on the new implementation and make the interface exactly what we want without worrying about backwards compat. For instance, maybe we should take an instantiated client approach, something like:

```ts
import { Client, hostingOrigin } from "../apiv2";

const api = new Client({
  origin: hostingOrigin,
  apiVersion: "v1beta1",
  auth: true
});

const channel = await api.post<CreateChannelRequest, Channel>({
  ttl: `${ttlMillis / 1000}s`
});
```

that's just an example, but we could try to build a more opinionated client experience and maybe pull in some OP-specific utilities (for instance, imagine a `listPaginated()` method that auto-paginates using OP parlance).

I'm not opposed to a "the existing thing, but TypeScript", but it looks like this approach is changing a fair amount of the surface but not really as a ground-up reimagining.

Another example here is that we no longer really need to vary requests by scope -- that's not a thing we do anymore since we rely on IAM to be our permissions gate.

src/apiv2.ts
bkendall: Maybe I misunderstood what these were for. My interpretation of these options were "these are good for debugging while coding", and could be turned on as necessary to print out more (or sensitive) information about a request.

Thinking back, is it intended to be the opposite? Is it desirable to print out as much information as possible by default rather than prevent it from being logged? (I generally agree with your thoughts though)

mbleigh: By default we debug log lots of information about requests/responses so that if there are errors we have a detailed log to dig into. But in some cases the request body is very large (e.g. hosting uploads), the response body is very large, or the query params contain sensitive information (auth-based requests).

It's not meant for use while developing, it's meant for "what do we want to show up in debug logs our customers paste into GitHub"

bkendall: Got it. I've swapped this around to `skipLog`

src/apiv2.ts
```diff
@@ -0,0 +1,283 @@
...
+  auth?: boolean;
+};
+
+export class Client {
+  constructor(private opts: ClientOptions) {
+    if (this.opts.auth === undefined) {
+      this.opts.auth = true;
+    }
+    if (this.opts.urlPrefix.endsWith("/")) {
+      this.opts.urlPrefix = this.opts.urlPrefix.substring(0, this.opts.urlPrefix.length - 1);
```
mbleigh, line 74: Isn't `substr` preferred now?

src/apiv2.ts
```diff
@@ -0,0 +1,283 @@
...
+    if (options.responseType === "json") {
+      body = await res.json();
+    } else {
+      // This is how the linter wants the casting to T to work.
+      body = ((await res.text()) as unknown) as ResT;
+    }
+
+    this.logResponse(res, body, options);
+
+    if (res.status >= 400) {
```
mbleigh, line 244: This shouldn't throw if `resolveOnHttpError` is in the options.

src/apiv2.ts
```diff
@@ -0,0 +1,283 @@
...
+    if (fetchOptions.json) {
+      if (!fetchOptions.skipLog?.body) {
+        logger.debug("[apiv2] HTTP REQUEST BODY:", JSON.stringify(fetchOptions.json));
+      } else {
+        logger.debug("[apiv2] HTTP REQUEST BODY OMITTED");
+      }
+    }
+  }
+
+  private logResponse(res: Response, body: unknown, options: ClientRequestOptions<unknown>): void {
```
mbleigh, line 276: I think we should tag each log line about a request with either the full URL so it's easy to correlate with requests. I also liked the `>>>` and `<<<` personally (helps me visually scan for reqs/responses). How about:

    >>> [v2][headers] POST https://the.url.goes.here/foo/bar {headers: 'go here'}
    >>> [v2][body] POST https://the.url.goes.here/foo/bar {headers: 'go here'}
    <<< [v2][status] POST https://the.url.goes.here/foo/bar 200
    <<< [v2][body] POST https://the.url.goes.here/foo/bar {response: 'body here'}

...or something

src/apiv2.ts
bkendall: Not that I recall. They're two different functions (which can be basically substituted here): the second argument for substr is the number is the _length_ to return, the second argument to substring is the index to stop at. Using 0 for the first argument, either is technically fine.

src/apiv2.ts
bkendall: Done. Was going to say "nothing uses this yet, I'll add a todo", but it's so simple it's not worth that effort haha

src/apiv2.ts
bkendall: I can get behind this idea. I did a bit of cleaning up of the log logic so hopefully it's a little easier to come back to in the future.

src/apiv2.ts
```diff
@@ -0,0 +1,293 @@
...
+
+  private logRequest(options: ClientRequestOptions<unknown>): void {
+    let searchParamLog = "[omitted]";
+    if (options.queryParams) {
+      if (!options.skipLog?.queryParams) {
+        searchParamLog = JSON.stringify(options.queryParams);
+      }
+    }
+    const logURL = this.requestURL(options);
+    logger.debug(`>>> [apiv2][searchParams] ${options.method} ${logURL} ${searchParamLog}`);
```
mbleigh, line 269:
```suggestion
    logger.debug(`>>> [apiv2][query] ${options.method} ${logURL} ${queryParamLog}`);
```

src/apiv2.ts
```diff
@@ -0,0 +1,293 @@
...
+    };
+  }
+
+  private logRequest(options: ClientRequestOptions<unknown>): void {
+    let searchParamLog = "[omitted]";
+    if (options.queryParams) {
+      if (!options.skipLog?.queryParams) {
+        searchParamLog = JSON.stringify(options.queryParams);
+      }
+    }
```
mbleigh, line 267:
```suggestion
    let queryParamLog = "[omitted]";
    if (options.queryParams) {
      if (!options.skipLog?.queryParams) {
        queryParamLog = JSON.stringify(options.queryParams);
      }
    }
```

mbleigh (approved):
Please manually test codepaths for all swapped implementations before merging.

PR #2774, opened 2020-11-02
Fix lookup empty results in Auth Emulator.

yuchenshi:
### Description

Fixes https://github.com/firebase/firebase-admin-node/issues/1078

Similar things can potentially happen to other fields that are `repeated` in proto definition, but I'd like to quick fix this one and we can look into solutions if this comes up again.

### Scenarios Tested

See tests added.

### Sample Commands

N/A

### DIFF
--- a/src/emulator/auth/operations.ts
+++ b/src/emulator/auth/operations.ts
@@ -215,10 +215,6 @@ function lookup(
         }
       }
     }
-    return {
-      kind: "identitytoolkit#GetAccountInfoResponse",
-      users,
-    };
   } else {
     assert(reqBody.idToken, "MISSING_ID_TOKEN");
     const { user } = parseIdToken(state, reqBody.idToken);
@@ -226,7 +222,10 @@ function lookup(
   }
   return {
     kind: "identitytoolkit#GetAccountInfoResponse",
-    users,
+
+    // Drop users property if no users are found. This is needed for Node.js
+    // Admin SDK: https://github.com/firebase/firebase-admin-node/issues/1078
+    users: users.length ? users : undefined,
   };
 }
 
--- a/src/test/emulators/auth.spec.ts
+++ b/src/test/emulators/auth.spec.ts
@@ -3334,6 +3334,33 @@ describe("Auth emulator", () => {
     });
   });
 
+  describe("accounts:lookup", () => {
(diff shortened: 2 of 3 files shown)


src/emulator/auth/operations.ts
```diff
@@ -215,18 +215,17 @@ function lookup(
...
   } else {
     assert(reqBody.idToken, "MISSING_ID_TOKEN");
     const { user } = parseIdToken(state, reqBody.idToken);
     users.push(redactPasswordHash(user));
   }
   return {
     kind: "identitytoolkit#GetAccountInfoResponse",
-    users,
+
+    // Drop users property if no users are found. This is needed for Node.js
```
samhorlbeck, line 226: nit: this isn't technically dropping the property; the response will still have a "users" property with value `undefined`. That means it'll still show up in `Object.keys(response)`.

Change the wording of this comment to be more explicit that the field is being set to undefined if the user array is empty

src/test/emulators/auth.spec.ts
```diff
@@ -3334,6 +3334,33 @@ describe("Auth emulator", () => {
...
+    });
+
+    it("should return empty result when localId is not found", async () => {
+      await supertest(authApp)
+        .post(`/identitytoolkit.googleapis.com/v1/projects/${PROJECT_ID}/accounts:lookup`)
+        .set("Authorization", "Bearer owner")
+        .send({ localId: ["noSuchId"] })
+        .then((res) => {
+          expectStatusCode(200, res);
+          expect(res.body).not.to.have.property("users");
```
samhorlbeck, line 3359: I have no idea how this works

samhorlbeck (approved):

src/emulator/auth/operations.ts
yuchenshi: JSON doesn't have an undefined type and `JSON.stringify` will just drop that field.

src/test/emulators/auth.spec.ts
yuchenshi: See above, this test is correct IMO

src/emulator/auth/operations.ts
samhorlbeck: I forgot the responses are `JSON.stringify`'d

PR #2778, opened 2020-11-04
ts-ify database:get

bkendall:
### Description

high-level switch of `database:get` from js to ts. I'm _not_ trying to fix or optimize code - keeping it simple to bring it to typescript first. Then I'll replace `request` with something else and clean up what I can.

The gnarliest change is that a new function `writeOut` needs to handle writing to the output streams because the callbacks of each options (file or stdout) don't match in signature.

### Scenarios Tested

`database:get` to a file and to stdout

### DIFF
--- a/src/commands/database-get.ts
+++ b/src/commands/database-get.ts
@@ -1,32 +1,31 @@
-"use strict";
+import * as _ from "lodash";
+import * as fs from "fs";
+import * as url from "url";
 
-var { Command } = require("../command");
-var requireInstance = require("../requireInstance");
-var { requirePermissions } = require("../requirePermissions");
-var request = require("request");
-var api = require("../api");
-var responseToError = require("../responseToError");
-var logger = require("../logger");
-var { FirebaseError } = require("../error");
-var { Emulators } = require("../emulator/types");
-var { printNoticeIfEmulated } = require("../emulator/commandUtils");
-var { populateInstanceDetails } = require("../management/database");
-const { realtimeOriginOrEmulatorOrCustomUrl } = require("../database/api");
-var utils = require("../utils");
-var _ = require("lodash");
-var fs = require("fs");
-var url = require("url");
+import { Command } from "../command";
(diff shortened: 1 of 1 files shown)


src/commands/database-get.ts
```diff
@@ -99,40 +98,51 @@ module.exports = new Command("database:get <path>")
...
-        var fileOut = !!options.output;
-        var outStream = fileOut ? fs.createWriteStream(options.output) : process.stdout;
-        var erroring;
-        var errorResponse = "";
-        var response;
+    return api.addRequestHeaders(reqOptions).then((reqOptionsWithToken) => {
+      return new Promise((resolve, reject) => {
+        const fileOut = !!options.output;
+        const outStream = fileOut ? fs.createWriteStream(options.output) : process.stdout;
+        const writeOut = (s: Buffer | string, cb?: Function): void => {
```
yuchenshi, line 109: The branches below are rather involved... Have we considered this?

```suggestion
        const writeOut = (s: Buffer | string, cb?: Function = () => {}): void => {
```

(and unify the branches below)

yuchenshi (approved):

src/commands/database-get.ts
bkendall: I went and tried, but unfortunately I still get a very annoying error:

> Each member of the union type '{ (chunk: any, cb?: ((error: Error | null | undefined) => void) | undefined): boolean; (chunk: any, encoding?: string | undefined, cb?: ((error: Error | null | undefined) => void) | undefined): boolean; } | { ...; }' has signatures, but none of those signatures are compatible with each other.

It's just a subtle difference between the two that stinks.

Fortunately - I'm working on the next version of this that doesn't use `request` and this logic _does_ go away (kinda), so it'll be nicer in the very near future.

yuchenshi: In that case, I don't think it's worth our time to dig into this any more. Resolving.

sarajmunjal (approved):

PR #2780, opened 2020-11-04
Work around Java emulators + WSL connectivity issues.

yuchenshi:
### Description

Fixes https://github.com/firebase/firebase-tools-ui/issues/332

This affects Firebase CLI running on WSL only. It has an undesirable side effect of breaking IPv6 use cases, but this is arguably better than not being able to use it at all from the host. I don't think we have a way to serve both IPv4 and IPv6 from host until https://github.com/microsoft/WSL/issues/4851 is fixed. See https://github.com/microsoft/WSL/issues/4851#issuecomment-722003091 for the root cause of the issue.

### Scenarios Tested

Manually tested by crwilcox@ on WSL2. Verified that Firestore Viewer in host can successfully list and edit data in Firestore emulator in subsystem.

### Sample Commands

* `firebase emulators:start` in WSL2.
* `curl http://127.0.0.1:8080/` from host to verify (it fails with connection refused without this patch).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Work around Java emulators + WSL connectivity issues.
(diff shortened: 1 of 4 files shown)


samtstern (approved):

samtstern:
@yuchenshi wow I'm so glad this is (mostly) solved now!

PR #2781, opened 2020-11-04
migrate database:get to new apiv2 client (with stream support!)

bkendall:
### Description

1. Adds streaming response support to the apiv2 client.
2. Migrates the `database:get` command to the new apiv2 `Client`.

### Scenarios Tested

`database:get`
- with a valid path
- with no auth (code change - tests the `4xx` response path)
- with an invalid path (returns `null`, same as head).
- to a file

### DIFF
--- a/src/apiv2.ts
+++ b/src/apiv2.ts
@@ -1,4 +1,5 @@
 import fetch, { Response, RequestInit } from "node-fetch";
+import { Readable } from "stream";
 
 import { FirebaseError } from "./error";
 import * as logger from "./logger";
@@ -12,7 +13,7 @@ interface RequestOptions<T> extends VerbOptions<T> {
   method: HttpMethod;
   path: string;
   json?: T;
-  responseType?: "json";
+  responseType?: "json" | "stream";
 }
 
 interface VerbOptions<T> {
@@ -238,6 +239,8 @@ export class Client {
     let body: ResT;
     if (options.responseType === "json") {
       body = await res.json();
+    } else if (options.responseType === "stream") {
+      body = (res.body as unknown) as ResT;
     } else {
       // This is how the linter wants the casting to T to work.
       body = ((await res.text()) as unknown) as ResT;
@@ -283,10 +286,15 @@ export class Client {
 
     let logBody = "[omitted]";
     if (!options.skipLog?.resBody) {
-      try {
-        logBody = JSON.stringify(body);
-      } catch (_) {
-        logBody = `${body}`;
+      if (body instanceof Readable) {
+        // Don't attempt to read any stream type, in case the caller needs it.
+        logBody = "[stream]";
+      } else {
+        try {
+          logBody = JSON.stringify(body);
+        } catch (_) {
+          logBody = `${body}`;
+        }
       }
     }
     logger.debug(`<<< [apiv2][body] ${options.method} ${logURL} ${logBody}`);
--- a/src/apiv2.ts
+++ b/src/apiv2.ts
@@ -144,6 +144,15 @@ export class Client {
       reqOptions.responseType = "json";
     }
 
+    // TODO(bkendall): stream + !resolveOnHTTPError makes for difficult handling.
+    //   Figure out if there's a better way to handle streamed >=400 responses.
+    if (reqOptions.responseType === "stream" && !reqOptions.resolveOnHTTPError) {
+      throw new FirebaseError(
+        "apiv2 will not handle HTTP errors while streaming and you must set `resolveOnHTTPError` and check for res.status >= 400 on your own",
+        { exit: 2 }
+      );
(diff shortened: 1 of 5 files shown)


src/apiv2.ts
```diff
@@ -238,6 +239,8 @@ export class Client {
     let body: ResT;
     if (options.responseType === "json") {
       body = await res.json();
+    } else if (options.responseType === "stream") {
+      body = (res.body as unknown) as ResT;
```
yuchenshi, line 243: I don't think `src/responseToError.js` will properly handle stream bodies. Mind fixing that (with a test)?

src/commands/database-get.ts
```diff
@@ -91,70 +100,45 @@ export default new Command("database:get <path>")
...
-                cb(err);
-              }
-            });
-          }
-        };
-        let erroring = false;
-        let errorResponse = "";
-        let response: any;
+    if (res.status >= 400) {
+      const r = await utils.streamToString(res.body);
```
yuchenshi, line 119: Can we move this to `src/responseToError.js`? See comment above.

src/commands/database-get.ts
```diff
@@ -91,70 +100,45 @@ export default new Command("database:get <path>")
...
+      if (outStream === process.stdout) {
+        // `stdout` can simply be written to.
+        outStream.write("\n");
+      } else if (outStream instanceof fs.WriteStream) {
+        // .pipe closes the output file stream, so we need to re-open the file.
+        const s = fs.createWriteStream(options.output, { flags: "a" });
+        s.write("\n");
+        s.close();
+      } else {
+        throw new FirebaseError("Could not write line break", { status: 2 });
```
yuchenshi, line 141: Is this severe enough that we want to throw? What if we just let it go?

src/test/apiv2.spec.ts
```diff
@@ -71,6 +72,54 @@ describe("apiv2", () => {
...
+        method: "GET",
+        path: "/path/to/foo",
+        responseType: "stream",
+        resolveOnHTTPError: true,
+      });
+      const data = await streamToString(r.body);
+      expect(data).to.deep.equal("does not exist");
+      expect(nock.isDone()).to.be.true;
+    });
+
```
yuchenshi, line 122: Please add a test for non-404 responses.

yuchenshi (changes requested):

src/test/apiv2.spec.ts
bkendall: Line 75 is a streamed GET request with a 200 response.

src/commands/database-get.ts
bkendall: Ehhh - probably not. I'll logger.debug it for posterity, but yeah, it wouldn't be the end of the world

src/commands/database-get.ts
bkendall: Yes, I think that's possible. But if I'm going to do that, responseToError is going to be converted to TS for my sanity. I can take this as a TODO to follow up on for sure.

Some of me is considering that I shouldn't though, if only because it's not necessarily obvious what we _should_ be doing with the streamed error content. In this case, we know that the 4xx response should be well-formed JSON, but other (more custom - maybe downloading emulators) may not return JSON. It can still _try_ I suppose, but that's my thought process - we definitely can always manually handle errors!

src/apiv2.ts
bkendall: It... would JSON encode the stream object, which would just be weird. Until responseToError is TS'd, I would keep it as an exercise for the developer to not misuse the library (which we use in a correct way in this PR).

yuchenshi: I'd push back on this one. Since `resolveOnHTTPError` is false by default, it's really easy to write and test for the happy cases but hit the error cases sometimes, leaving a cryptic error message. Yes I see your comments below -- but would it make sense to throw if `resolveOnHTTPError` is not set to true with streaming for now and recommend manually handling all errors? We can figure out a story if we find more common usages.

src/commands/database-get.ts
yuchenshi: Ack. See comment above.

src/test/apiv2.spec.ts
yuchenshi: Ah, can we have a 400 error too? I see those errors hit a different branch that deals with response body.

src/apiv2.ts
bkendall: Alright - I'm throwing an error if (streaming && !resolvingOnHttpError), forcing the developer to have to set that and handle the error for now.

I'm open to future ideas on how to make responseToError better - especially when it comes to streaming!

yuchenshi (approved):

PR #2788, opened 2020-11-06
use apiv2 in operation-poller

bkendall:
### Description

Swap out `api` for `apiv2` in the operation poller. AFAIK, since these clients are incredibly similar, this should work pretty well.

While updating the tests, I removed "stubbing the API client" for "stubbing the http responses" with nock. This means that the APIv2 client is being fully utilized and not just stubbed away. This gives me more confidence that the tests are actually validating the interaction between the two.

### Scenarios Tested

- `hosting:clone`
- `ext:uninstall`

### DIFF
--- a/src/operation-poller.ts
+++ b/src/operation-poller.ts
@@ -1,6 +1,6 @@
-import { Queue } from "./throttler/queue";
-import * as api from "./api";
+import { Client, ClientResponse } from "./apiv2";
 import { FirebaseError } from "./error";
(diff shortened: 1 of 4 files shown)


yuchenshi (approved):

yuchenshi:
FWIW please assign me in "assignees" for PRs -- that helps me get to it faster. I get cc'd on way too many PRs and often tagged as reviewer on non-actionable PRs (due to being a codeowner)

PR #2795, opened 2020-11-06
use apiv2 in fetchmotd

bkendall:
### Description

Replace `request` with `apiv2` in fetching the message of the day.

Did you know that the logger commands in this file never did anything? Turns out that winston doesn't have a console output at this point. So, swapped to console.logs/errors.

### Scenarios Tested

- `hosting:channels:list` works as normal
- removing the motd item from the config store triggers the logic which repopulates it (and then gets printed the next invocation)


src/fetchMOTD.ts
```diff
@@ -21,7 +19,7 @@ export function fetchMOTD(): void {
 
   if (motd && motdFetched > Date.now() - ONE_DAY_MS) {
     if (motd.minVersion && semver.gt(motd.minVersion, pkg.version)) {
-      logger.error(
```
samtstern, line 24: Why use `console.` here?  I thought we don't use that in our codebase.

samtstern (approved):
Fine with the API changes but confused by the logging.

src/fetchMOTD.ts
samtstern: Oh wait I see your description now ... who knew!

PR #2803, opened 2020-11-09
update database:update to use apiv2 (with streaming request support)

bkendall:
### Description

Update `database:update` to use the new `apiv2` module.

Also updates the `apiv2` module to accept streaming inputs for requests.

### Scenarios Tested

- `database:update` with
  - entering json in stdin
  - a file
  - `--data` value


src/apiv2.ts
```diff
@@ -89,25 +89,25 @@ export class Client {
     json?: ReqT,
     options: ClientVerbOptions<ReqT> = {}
   ): Promise<ClientResponse<ResT>> {
-    const reqOptions: ClientRequestOptions<ReqT> = Object.assign(options, {
+    const reqOptions: ClientRequestOptions<string> = Object.assign(options, {
       method: "POST",
       path,
-      json,
+      body: JSON.stringify(json),
```
mbleigh, line 95: Should we just switch this to pass `body` straight through to request? Why allow `body` to take an arbitrary object and serialize it but then do it manually here?

Also, it might be "too much magic" but we should consider setting `content-type: application/json` when a JSON body type is passed (so long as the `content-type` header is not otherwise explicitly set). Some APIs will reject JSON content without the header.

bkendall: Ah - this was a strange artifact of fixing the typing when I was working on it. I'm passing `json` straight through now.

By default, `.request` assumes the response type will be `json` and sets the `content-type` header appropriately. This is likely the vast majority of cases. If the response is going to be streamed (`responseType: "stream"`), the `content-type` header isn't set automatically and it would be up to the caller to set the header and/or parse the response as appropriate. This seems reasonable to me since `.request` is not going to be used often, but I'm open to revisiting as we find more use cases...

sarajmunjal (approved):

PR #2804, opened 2020-11-10
ts database:profile and use apiv2

bkendall:
### Description

Rewrite `profiler.js` into Typescript, reworking the logic for better handling with a cancelable `apiv2` client.

Now that `apiv2` can be canceled with a signal, we can simplify the logic around when the profiling should stop. If a duration is provided, or when enter is pressed, we can cancel the ongoing request and write the report.

### Scenarios Tested

`database:profile`, doing a `database:get` in another tab
- with a duration
- open-ended pressing enter after a couple `get`s


IanWyszynski:
I tested this the old and new versions of the profiler while running some load against my prod namespace, results were consistent. Also tested with the `--duration` flag, and cancelling the request with `[enter]` a few times. 

Thanks!

IanWyszynski (approved):

PR #2806, opened 2020-11-11
[cleanup] remove unexpected `await`s

bkendall:
### Description

Removing the `await-thenable` warning and fixing all the errors.

### Scenarios Tested

- [ ] give a few more commands some sanity testing


src/emulator/functionsEmulator.ts
```diff
@@ -424,7 +424,7 @@ export class FunctionsEmulator implements EmulatorInstance {
               added = await this.addPubsubTrigger(this.args.projectId, definition);
               break;
             case Constants.SERVICE_AUTH:
-              added = await this.addAuthTrigger(this.args.projectId, definition);
+              added = this.addAuthTrigger(this.args.projectId, definition);
```
samtstern, line 427: This one is a bit suspicious to me.  I guess because auth is in-process it can be added synchronously?

src/hosting/functionsProxy.ts
```diff
@@ -49,6 +49,6 @@ export default function(
       }
     }
 
-    return await proxyRequestHandler(url, `${destLabel} Function ${rewrite.function}`);
+    return Promise.resolve(proxyRequestHandler(url, `${destLabel} Function ${rewrite.function}`));
```
samtstern, line 52: Why this change?

samtstern (approved):
Two small comments

src/emulator/functionsEmulator.ts
bkendall: yup - it's a sync function, not async, so no need to await

src/hosting/functionsProxy.ts
bkendall: Two fold:
1. proxyRequestHandler is _not_ async, and does _not_ return a promise (async and await are removed)
2. Since the signature of the function required a promise (and the JSDocs too), the returned handler must be wrapped in a promise.

src/hosting/functionsProxy.ts
```diff
@@ -25,7 +25,7 @@ export interface FunctionProxyRewrite {
 export default function(
   options: FunctionsProxyOptions
 ): (r: FunctionProxyRewrite) => Promise<RequestHandler> {
-  return async (rewrite: FunctionProxyRewrite) => {
+  return (rewrite: FunctionProxyRewrite) => {
```
yuchenshi, line 28: nit: You may want to wrap the whole function body around `new Promise` to keep the behavior the same. Before, any errors in the function body will be returned as a rejected Promise. Now it just synchronously throws (which can be really bad), unless you wrap everything.

yuchenshi (approved):
LGTM (with nit above)

src/hosting/functionsProxy.ts
bkendall: Makes sense. I'll wrap it. Thanks.

PR #2810, opened 2020-11-11
Adding support for ${param:FOO} syntax

### LINKED ISSUE #2777: Cloud function not picking up RTDB emulator when using handler namespace from firebase-functions sdk

firebase CLI 8.14.1

**firebase-tools:**

macOS Catalina 10.15.4

**Platform:**

### [REQUIRED] Test case

I've created a repository for the bug with the steps in the `README.md` file to reproduce: https://github.com/russellwheatley/emulator-extension-issue

I expected the cloud function emulator to run with the RTDB emulator when using the handler namespace of the `firebase-functions` sdk (apologies if this is in the wrong repo, not sure whose responsibility this falls under). 

It doesn't recognise the RTDB emulator. Here is a log from the console:

```bash
functions[rtdblimit]: function ignored because the unknown emulator does not exist or is not running.
```

Let me know if you need more information.

joehan:
### Description
Fixes an issue where the extension emulator did not support ${param:FOO} syntax. Should unblock https://github.com/firebase/extensions/pull/482 and fix #2777 

I also did some deduping here - there were 2 substituteParams functions in different files, so I removed one and migrated the usage.

### Scenarios Tested
I tested this by running both `ext:dev:emulators` commands on an extension that used a mix of ${FOO} and ${param:FOO} syntax in its triggers. Both syntaxes worked correctly.


tinaliang (approved):

bkendall:
(drive-by merging; fixed your changelog at the same time)

PR #2816, opened 2020-11-12
typescript the Hosting uploader and replace api.request and request with apiv2

bkendall:
### Description

Typescripting Hosting's uploader.js and replacing the old usages of the API request method and `request` module with the new apiv2 module.

### Scenarios Tested

Lots of hosting deploys. And redeploys with no changes!


yuchenshi:
FYI reviewers: You can find a[ side-by-side comparison here](https://github.com/firebase/firebase-tools/commit/e94d534012caf2513bf3844b643f6d04796d46e5#diff-ba3db83e9e7bbc92bd71735a62444986b9c8e3febd4d06314487669289fdd0a9) for the `uploader.ts` file.

Side note: git did not track the move as a move since the files are not similar enough, so GitHub won't show a side-by-side. To make the life of reviewers easier, you can instead create a single commit with just `git mv` (no changes) and then change things in follow-up commits, like this: https://github.com/firebase/firebase-tools/compare/ys/bk-uploader-diff. This allows reviewers to change the diffbase to the first commit instead of master to compare.

src/deploy/hosting/uploader.ts
```diff
@@ -0,0 +1,268 @@
...
+  const percent = Math.floor(((current * 1.0) / total) * 100).toString();
+  return `${message} [${current}/${total}] (${clc.bold.green(`${percent}%`)})`;
+}
+
+export class Uploader {
+  private version: string;
+  private cwd: string;
+  private projectRoot: string;
+  private gzipLevel: number;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
yuchenshi, line 30: nit: Can we consider `/* eslint-disable @typescript-eslint/no-explicit-any */` and then re-enable a few lines after? A disable comment every two lines seems really noisy.

src/deploy/hosting/uploader.ts
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    const fstream = this.zipStream(filePath);
+    const hash = crypto.createHash("sha256");
+
+    fstream.pipe(hash);
+
+    return new Promise((resolve, reject) => {
+      fstream.on("end", () => {
+        const hashVal = hash.read().toString("hex");
```
yuchenshi, line 180: nit: Let's just `fstream.on("end", resolve);` and move these logic after awaiting the promise.

src/deploy/hosting/uploader.ts
```diff
@@ -0,0 +1,268 @@
...
+
+  async uploadHandler(toUpload: string): Promise<void> {
+    if (!this.uploadClient) {
+      throw new FirebaseError("No upload client available.", { exit: 2 });
+    }
+    const controller = new AbortController();
+    const timeout = setTimeout(() => {
+      controller.abort();
+    }, this.uploadTimeout(this.hashMap[toUpload]));
+    const res = await this.uploadClient.request({
```
yuchenshi, line 236: Did you forget to pass in `controller`?

Also, side topic: can we just add `timeout` as an option to `apiv2`? This seems to be a common enough case and while having the flexibility of `controller` is nice, it seems boilerplate-y and prone to errors like forgetting to pass it in, not clearing timeouts, etc. etc.

yuchenshi (approved):

src/deploy/hosting/uploader.ts
bkendall: seems reasonable enough of an adjustment. done.

src/deploy/hosting/uploader.ts
bkendall: I did, thanks. I didn't know how many use-cases we had for timeouts yet, so I haven't added it to `apiv2`. I'm still back and forth on it a bit, but the case is a little stronger. Not sure how many other spots in the codebase deal with timeouts (I've only seen 2-3, and it's pretty non-trivial logic handling it, usually, which may benefit from more custom solutions like this rather than a single setting)

src/deploy/hosting/uploader.ts
bkendall: ack. I've found using disable blocks far easier to set and then forget to unset, and I'd rather have the pain of excluding any `any` individually because they _really_ should have a specific type (but in converting js->ts, I'm not usually bothered to fix all of them)

PR #2827, opened 2020-11-16
Update extension template with new Cloud Function locations and up to date dependencies

joehan:
### Description
Adds missing Cloud Functions locations to the default extension created by ext:dev:init.


rosariopfernandes:
Most of those regions were added in [firebase-functions v3.8.0](https://github.com/firebase/firebase-functions/releases/tag/v3.8.0) and [v3.9.0](https://github.com/firebase/firebase-functions/releases/tag/v3.9.0). Maybe we should also update the template's `package.json` files?

joehan:
Great catch @rosariopfernandes  - just updated the firebase-admin and firebase-functions dependencies.

tinaliang (approved):

PR #2828, opened 2020-11-16
remove request module from database helpers

bkendall:
### Description

`request` is deprecated. replacing usages of it with the new `apiv2` client.

I also found some interesting cases where `databse:get` wouldn't write to the output file correctly, so I re-vamped the logic to make sure that the command only exits when the streams fully complete.

I also added some client-integration tests for the `database:*` commands, which helped debug & fix.

### Scenarios Tested

- `database:get`
- `database:set`
- `database:remove`

### DIFF
--- a/scripts/client-integration-tests/tests.ts
+++ b/scripts/client-integration-tests/tests.ts
@@ -110,6 +110,6 @@ describe("database:set|get|remove", () => {
     await client.database.remove(path, Object.assign({ confirm: true }, opts));
 
     await client.database.get(path, Object.assign({ output: file.name }, opts));
-    expect(JSON.parse(readFileSync(file.name, 'utf-8'))).to.equal(null);
+    expect(JSON.parse(readFileSync(file.name, "utf-8"))).to.equal(null);
   }).timeout(10 * 1e3);
 });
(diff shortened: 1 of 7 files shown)


scripts/client-integration-tests/tests.ts
```diff
@@ -85,3 +89,27 @@ describe("apps:sdkconfig", () => {
...
+    // Have to read to a file in order to get data.
+    const file = tmp.fileSync();
+
+    await client.database.get(path, Object.assign({ output: file.name }, opts));
+    expect(JSON.parse(readFileSync(file.name).toString())).to.deep.equal(data);
+
+    await client.database.remove(path, Object.assign({ confirm: true }, opts));
+
+    await client.database.get(path, Object.assign({ output: file.name }, opts));
+    expect(JSON.parse(readFileSync(file.name).toString())).to.equal(null);
```
yuchenshi, line 113:
```suggestion
    expect(JSON.parse(readFileSync(file.name, 'utf-8'))).to.equal(null);
```

More explicit, less conversion

yuchenshi (approved):

PR #2836, opened 2020-11-17
remove `request` from `src/hosting` files (proxy middleware)

bkendall:
### Description

Replaces usage of the `request` library in the `hosting` directory with our new `apiv2`. `apiv2` needed to be extended to support query parameters and not follow redirects - tests have been added for that.

### Scenarios Tested

Tested against the emulator and a real function. Tested with the function taking too long and the abort logic taking effect. Tested both GET and POST requests.

### DIFF
--- a/src/apiv2.ts
+++ b/src/apiv2.ts
@@ -8,7 +8,7 @@ import * as responseToError from "./responseToError";
 
 const CLI_VERSION = require("../package.json").version;
 
-type HttpMethod = "GET" | "PUT" | "POST" | "DELETE" | "PATCH";
+export type HttpMethod = "GET" | "PUT" | "POST" | "DELETE" | "PATCH";
 
 interface RequestOptions<T> extends VerbOptions<T> {
   method: HttpMethod;
--- a/src/apiv2.ts
+++ b/src/apiv2.ts
@@ -1,6 +1,7 @@
 import fetch, { Response, RequestInit } from "node-fetch";
 import { AbortSignal } from "abort-controller";
 import { Readable } from "stream";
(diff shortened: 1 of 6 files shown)


src/apiv2.ts
```diff
@@ -270,7 +275,9 @@ export class Client {
       body = (res.body as unknown) as ResT;
     } else {
       // This is how the linter wants the casting to T to work.
```
yuchenshi, line 277: Drop this comment

src/hosting/proxy.ts
```diff
@@ -35,67 +41,94 @@ function makeVary(vary?: string): string {
...
-        Cookie: sessionCookie,
-      },
-      followRedirect: false,
-      timeout: 60000,
-    });
+    // req.url is just the full path (e.g. /foo?key=value; no origin).
+    const u = new URL(url + req.url);
+    const c = new Client({ urlPrefix: u.origin, auth: false });
+    const controller = new AbortController();
+    const timer: NodeJS.Timeout = setTimeout(() => controller.abort(), 60000);
```
yuchenshi, line 57: nit: Consider using `Promise.race` with a Promise that resolves in 60s to make it less scary w.r.t race conditions

yuchenshi (approved):

src/hosting/proxy.ts
yuchenshi: Ah nevermind, I see how abort errors are handled before pipes now

src/apiv2.ts
bkendall: Done.

PR #2842, opened 2020-11-19
remove request module from database commands

bkendall:
### Description

`request` is deprecated. Long live `apiv2`.

### Scenarios Tested

- `database:push` with `--data` and with a file
- `database:set` success, and not confirming (which errored)


sarajmunjal (approved):

sarajmunjal:
LGTM. Could we try this with a European instance too, @bkendall ?

bkendall:
Yup! Done. Works like a charm!

PR #2873, opened 2020-11-29
Fix emulated auth popup widget UI cross-origin frame issue

SploxFox:
### Description

I ran into an issue with the emulated auth popup widget ui [(stackoverflow issue)](https://stackoverflow.com/questions/65063526/emulated-google-idp-authentication-throwing-cross-origin-frame-error) and did some investigating and found that the issue was that it was trying to access window.location.search on a cross-origin frame. I simply wrapped the code in a try/catch so that way if the frame is cross origin, it will just move onto the next frame.


SploxFox:
@googlebot I signed it!

samtstern:
@samhorlbeck or @yuchenshi PTAL

samhorlbeck (approved):

samtstern:
@SploxFox thanks for the PR!  Since it's approved we're almost ready to merge.  Could you:

1. Add an entry to `CHANGELOG.md` describing the fix?
2. Update your branch (you can do this directly in the GitHub UI or at the command line)

yuchenshi (approved):
This is a great change and matches production behavior too

samtstern:
@SploxFox small lint issue but I fixed it

PR #2879, opened 2020-12-01
Split Auth Emulator tests into multiple files.

yuchenshi:
### Description

This is a no-op refactor to break tests into multiple files, because people don't love 3700-line files. Tests aren't changed (although their names may be prefixed by `Auth Emulator: ` now instead of lumped together).

Interesting things to look at is `setup.ts` and (maybe) `helpers.ts`. Other changes are pretty mechanical.

### Scenarios Tested

`npm test`

### Sample Commands

N/A


src/test/emulators/auth/setup.ts
```diff
@@ -0,0 +1,57 @@
...
+ */
+export function describeAuthEmulator(
+  title: string,
+  fn: (this: Suite, utils: AuthTestUtils) => void
+): Suite {
+  // Avoid arrow functions in order to pass Mocha `this` properly in:
+  return describe(`Auth Emulator: ${title}`, function() {
+    let authApp: Express.Application;
+    beforeEach("setup or reuse auth server", async function() {
+      // eslint-disable-next-line no-invalid-this
```
samhorlbeck, line 23: Try to explicitly type `this` if you have a type (Probably from something in mocha)
https://www.typescriptlang.org/docs/handbook/functions.html#this-parameters

samhorlbeck (approved):

src/test/emulators/auth/setup.ts
yuchenshi: `this` is already typed in the `beforeEach` callback. The warning won't go away either even if I repeat it in my code like `async function(this: Mocha.Context)`. I think the `eslint-disable` is really needed.

yuchenshi: Ahh turns out we need [`@typescript-eslint/no-invalid-this`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-invalid-this.md,) instead of the plain eslint rule to achieve what you've described. I can follow up with another PR.

yuchenshi: https://github.com/firebase/firebase-tools/pull/2882

PR #2881, opened 2020-12-01
Extensions Registry Alpha

huangjeff5:
### Description

This branch contains new CLI commands and updates to existing CLI commands to support the Registry launch, which allows EAP developers to create, publish, and share extensions to other EAP developers.

Equivalent PR on firebasePrivate (with link to UX walkthrough): https://github.com/FirebasePrivate/firebase-tools/pull/670


tinaliang (approved):
Approving on the basis that Joe/myself have reviewed each individual PR

samtstern (approved):

src/test/extensions/resolveSource.spec.ts
```diff
@@ -66,25 +64,24 @@ describe("checkForUpdateWarnings", () => {
         "After updating, you must switch your Cloud Scheduler jobs to <b>PubSub</b>, otherwise your extension will stop running.",
     };
     expect(confirmUpdateWarningSpy).to.have.been.calledWith(expectedUpdateWarning);
-  });
+  }).timeout(5000);
```
bkendall, line 67: Why is this timeout needed? I've found that if timeouts are necessary, it's because some network call isn't being stubbed/mocked/nock'd correctly. I've recently forced all tests to have auth (for API calls) mocked out, so timeouts like this shouldn't be necessary. Is it possible to remove this timeout? (you can run this specific test by using `it.only` or passing the file name directly to `mocha`)

src/test/extensions/resolveSource.spec.ts
```diff
@@ -66,25 +64,24 @@ describe("checkForUpdateWarnings", () => {
...
   it("should display no warnings if none are applicable", async () => {
     await resolveSource.promptForUpdateWarnings(testRegistryEntry, "0.1.1", "0.1.2");
 
     expect(confirmUpdateWarningSpy).not.to.have.been.called;
   });
 });
-
-describe("isOfficialSource", () => {
-  it("should return true for an official source", () => {
+describe("isPublishedSource", () => {
```
bkendall, line 75: Please keep a line break between blocks.

src/test/extensions/updateHelper.spec.ts
```diff
@@ -31,177 +33,465 @@ const SPEC = {
...
+        "test-project",
+        "test-instance",
+        ".",
+        SPEC,
+        SPEC.name
+      );
+      expect(name).to.equal(SOURCE.name);
     });
-  });
+    it("should throw an error for an invalid source", async () => {
```
bkendall, line 165: Please keep a line break between blocks. (applies throughout - makes it easier to parse)

src/test/extensions/updateHelper.spec.ts
```diff
@@ -31,177 +33,465 @@ const SPEC = {
...
+      promptStub.resolves(true);
+      createSourceStub.throwsException("Invalid source");
+      await expect(
+        updateHelper.updateFromLocalSource("test-project", "test-instance", ".", SPEC, SPEC.name)
+      ).to.to.be.rejectedWith(FirebaseError, "Unable to update from the source");
```
bkendall, line 170: one `.to` is probably enough

src/test/extensions/updateHelper.spec.ts
```diff
@@ -31,177 +33,465 @@ const SPEC = {
...
+    });
+    it("should not update if the update warning is not confirmed", async () => {
+      promptStub.resolves(false);
+      createSourceStub.resolves(SOURCE);
+      await expect(
+        updateHelper.updateFromLocalSource("test-project", "test-instance", ".", SPEC, SPEC.name)
+      ).to.be.rejectedWith(FirebaseError, "Update cancelled.");
+    });
+  }).timeout(2500);
```
bkendall, line 179: Why is this timeout needed? I've found that if timeouts are necessary, it's because some network call isn't being stubbed/mocked/nock'd correctly. I've recently forced all tests to have auth (for API calls) mocked out, so timeouts like this shouldn't be necessary. Is it possible to remove this timeout? (you can run this specific test by using `it.only` or passing the file name directly to `mocha`)

src/test/extensions/updateHelper.spec.ts
```diff
@@ -31,177 +33,465 @@ const SPEC = {
...
+        updateHelper.updateToVersionFromPublisherSource(
+          "test-project",
+          "test-instance",
+          "test-publisher/test@0.2.0",
+          SPEC,
+          SPEC.name
+        )
+      ).to.be.rejectedWith(FirebaseError, "Update cancelled.");
+    });
+  }).timeout(2500);
```
bkendall, line 303: Another suspicious timeout

bkendall (changes requested):

src/test/extensions/updateHelper.spec.ts
huangjeff5: Good catch. I believe it was probably for those auth API calls, since I removed the timeouts and the tests still passed.

src/test/extensions/updateHelper.spec.ts
huangjeff5: Removed it and tests still pass, thanksfully!

src/commands/ext-dev-list.ts
```diff
@@ -0,0 +1,67 @@
...
+import * as utils from "../utils";
+import * as extensionsUtils from "../extensions/utils";
+import { listExtensions } from "../extensions/extensionsApi";
+import * as logger from "../logger";
+import { requireAuth } from "../requireAuth";
+
+/**
+ * List all published extensions associated with this publisher ID
+ */
+export default new Command("ext:dev:list [publisherId]")
```
bkendall, line 17: if `publisherId` is required, it should be in `<>` angle brackets

src/commands/ext-dev-list.ts
```diff
@@ -0,0 +1,67 @@
...
+  .description("list all published extensions associated with this publisher ID")
+  .before(requireAuth)
+  .action(async (publisherId: string, options: any) => {
+    if (!publisherId) {
+      throw new FirebaseError(
+        `You need to enter a publisher ID to view their published extensions. Expected Usage: ${clc.bold(
+          "ext:dev:list <publisher ID>"
+        )}`
+      );
+    }
```
bkendall, line 27: If you make publisherId required (with angle brackets), you _might_ not need this check.

If you _do_ need this check, make sure the parameters match and please print out the full command:

`firebase ext:dev:list <publisherId>`

src/commands/ext-dev-publish.ts
```diff
@@ -0,0 +1,36 @@
...
+  .description(`publish a new version of an extension`)
+  .help(
+    "if you have not previously published a version of this extension, this will " +
+      "create the extension. If you have previously published a version of this extension, this version must " +
+      "be greater than previous versions."
+  )
+  .before(requireAuth)
+  .action(async (extensionRef: string, options: any) => {
+    const [publisherId, extensionId] = extensionRef.split("/");
+    if (!publisherId || !extensionId) {
```
bkendall, line 21: My understanding of Commander's behavior (that powers our CLI options) is that since you specify `<publisherId>/<extensionId>`, you'll need to have two parameters in the `action` function before `options` to accept both of those parameters. Can you verify this behavior for me?

Thinking it through a bit more, the CLI may actually be hiding such an error from you, in a way. I'd recommend you go in and check this out. (My gut tells me that `<publisherId>/<extensionId>` really needs to be `<extensionRef>` and it needs to be documented to say "in the form of 'publisherId/extensionId'" based on my past experience, but I may be wrong. I _think_ this is the only place we've had compound parameters like this. The fact that you're trying to split them makes me... curious.

src/commands/ext-dev-publish.ts
```diff
@@ -0,0 +1,36 @@
...
+      throw new FirebaseError(
+        `Error parsing publisher ID and extension ID from extension reference '${clc.bold(
+          extensionRef
+        )}'. Please use the format '${clc.bold("<publisher-id>/<extension-id>")}'.`
```
bkendall, line 25: Please make these parameters match the definition of the command (`publisherId`, `extensionId`)

src/commands/ext-dev-unpublish.ts
```diff
@@ -0,0 +1,47 @@
+import { Command } from "../command";
+import { logPrefix } from "../extensions/extensionsHelper";
+import { unpublishExtension, parseRef, getExtension } from "../extensions/extensionsApi";
+import * as utils from "../utils";
+import { promptOnce } from "../prompt";
+import * as clc from "cli-color";
+import { requireAuth } from "../requireAuth";
+
+module.exports = new Command("ext:dev:unpublish [ref]")
```
bkendall, line 9: This `ref` doesn't match the `publisherId` and `extensionId` pair of parameters as set out in `:publish`. Is there a reason for this?

src/commands/ext-dev-unpublish.ts
```diff
@@ -0,0 +1,47 @@
...
+  .description("unpublish an extension")
+  .help(
+    "use this command to unpublish an extension, and make it unavailable for developers to install or reconfigure. " +
+      "Specify the extension you want to unpublish using the format '<publisherId/extensionId>'."
+  )
+  .before(requireAuth)
+  .action(async (ref: string, options: any) => {
+    const { publisherId, extensionId, version } = parseRef(ref);
```
bkendall, line 17: `parseRef` seems like something that could be used in `:publish` too, if the input parameter needed to change.

src/commands/ext-dev-unpublish.ts
```diff
@@ -0,0 +1,47 @@
...
+    const message =
+      "If you unpublish this extension, developers won't be able to install it. For developers who currently have this extension installed, it will continue to run and will appear as unpublished when listed in the Firebase console or Firebase CLI.";
+    utils.logLabeledWarning(logPrefix, message);
+    if (version) {
+      return utils.reject(
+        `Unpublishing a single version is not currently supported. You can only unpublish ${clc.bold(
+          "ALL versions"
+        )} of an extension. To unpublish all versions, please remove the version from the reference.`,
+        { exit: 1 }
```
bkendall, line 26: I haven't been putting this everywhere, but FWIW: `exit: 1` is the default behavior so this can be omitted.

src/commands/ext-dev-unpublish.ts
```diff
@@ -0,0 +1,47 @@
...
+    await getExtension(ref);
+    const consent = await comfirmUnpublish(publisherId, extensionId);
+    if (!consent) {
+      return utils.reject("Unpublishing cancelled.", { exit: 1 });
+    }
+    await unpublishExtension(ref);
+    utils.logLabeledSuccess(logPrefix, "successfully unpublished all versions of this extension.");
+  });
+
+export async function comfirmUnpublish(publisherId: string, extensionId: string): Promise<string> {
```
bkendall, line 38: Does this need to be exported?

src/commands/ext-dev-unpublish.ts
```diff
@@ -0,0 +1,47 @@
...
+        `Unpublishing a single version is not currently supported. You can only unpublish ${clc.bold(
+          "ALL versions"
+        )} of an extension. To unpublish all versions, please remove the version from the reference.`,
+        { exit: 1 }
+      );
+    }
+    await getExtension(ref);
+    const consent = await comfirmUnpublish(publisherId, extensionId);
+    if (!consent) {
+      return utils.reject("Unpublishing cancelled.", { exit: 1 });
```
bkendall, line 32: Also, fwiw, since you are in an async function, you can `throw new FirebaseError` instead of `return utils.reject` (they have the same call signature)

```suggestion
      throw new FirebaseError("Unpublishing cancelled.");
```

src/commands/ext-install.ts
```diff
@@ -126,7 +151,7 @@ export default new Command("ext:install [extensionName]")
   .description(
     "install an official extension if [extensionName] or [extensionName@version] is provided; " +
       (previews.extdev
-        ? "install a local extension if [localPathOrUrl] or [url#root] is provided; "
+        ? "install a local extension if [localPathOrUrl] or [url#root] is provided; install a published extension (not authored by Firebase) if [publisherId/extensionName] is provided "
```
bkendall, line 154: Should this be:

```suggestion
        ? "install a local extension if [localPathOrUrl] or [url#root] is provided; install a published extension (not authored by Firebase) if [publisherId/extensionId] is provided "
```

src/commands/ext-install.ts
```diff
@@ -156,47 +181,84 @@ export default new Command("ext:install [extensionName]")
...
     try {
       if (learnMore) {
         utils.logLabeledBullet(
           logPrefix,
-          `You selected: ${clc.bold(source.spec.displayName)}.\n` +
-            `${source.spec.description}\n` +
-            `View details: https://firebase.google.com/products/extensions/${name}\n`
+          `You selected: ${clc.bold(spec.displayName)}.\n` +
+            `${spec.description}\n` +
+            `View details: https:// firebase.google.com/products/extensions/${name}\n`
```
bkendall, line 261: I don't think you want a space here

```suggestion
            `View details: https://firebase.google.com/products/extensions/${name}\n`
```

src/commands/ext-update.ts
```diff
@@ -60,148 +68,219 @@ export default new Command("ext:update <extensionInstanceId> [localDirectoryOrUr
...
+      // We only allow the following types of updates.
+      let validUpdate = false;
+      if (existingSourceOrigin === SourceOrigin.OFFICIAL_EXTENSION) {
+        if (
+          [
+            SourceOrigin.LOCAL,
+            SourceOrigin.URL,
+            SourceOrigin.OFFICIAL_EXTENSION,
+            SourceOrigin.OFFICIAL_EXTENSION_VERSION,
+          ].indexOf(newSourceOrigin) > -1
```
bkendall, line 122: This seems to be the same list (and check) as later in this file. Consider moving the list to a constant above; and consider doing something like this (rather than .indexOf)

```typescript
allowedSourceOrigins.includes(newSourceOrigin)
```

src/commands/ext-update.ts
```diff
@@ -60,148 +68,219 @@ export default new Command("ext:update <extensionInstanceId> [localDirectoryOrUr
...
-        try {
-          const registryEntry = await resolveSource.resolveRegistryEntry(currentSpec.name);
-          updatingFromOfficial = resolveSource.isOfficialSource(registryEntry, existingSource);
-        } catch {
-          // If registry entry does not exist, assume local directory or URL extension source.
+      } else if (
+        existingSourceOrigin === SourceOrigin.LOCAL ||
+        existingSourceOrigin === SourceOrigin.URL
+      ) {
+        if ([SourceOrigin.LOCAL, SourceOrigin.URL].indexOf(newSourceOrigin) > -1) {
```
bkendall, line 141: Prefer `Array.prototype.includes` to `indexOf() > -1` throughout

src/commands/ext-update.ts
```diff
@@ -60,148 +68,219 @@ export default new Command("ext:update <extensionInstanceId> [localDirectoryOrUr
...
+              instanceId,
+              updateSource,
+              existingSpec,
+              existingSource
+            );
+            break;
+          }
+          // falls through
+        }
+        // eslint-disable-next-line no-fallthrough
```
bkendall, line 175: I _think_ this can be omitted since you say `// falls through` above?

If that's not the case, it's because you're including `{}` after the `case` statement which is unnecessary - they should probably be removed

src/extensions/askUserForConsent.ts
```diff
@@ -0,0 +1,100 @@
...
+import { promptOnce, Question } from "../prompt";
+import * as utils from "../utils";
+
+marked.setOptions({
+  renderer: new TerminalRenderer(),
+});
+
+/**
+ * Returns a string that will be displayed in the prompt to user.
+ * @param {string} role
```
bkendall, line 18: In TS files, the type can be omitted in jsdocs

```suggestion
 * @param role
```

src/extensions/askUserForConsent.ts
```diff
@@ -0,0 +1,100 @@
...
+ * @return {!Promise<string>}
+ */
+export async function formatDescription(extensionName: string, projectId: string, roles: string[]) {
+  const question = `${clc.bold(
+    extensionName
+  )} will be granted the following access to project ${clc.bold(projectId)}`;
+  const results: string[] = await Promise.all(
+    roles.map(async (role: string) => {
+      return await retrieveRoleInfo(role);
+    })
```
bkendall, line 28: It's a really subtle thing, but you can omit `async` and `await` here - it'll still map to a list of promises which can be Promise.all'd

src/extensions/askUserForConsent.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/**
+ * Displays roles and corresponding descriptions and asks user for consent
+ * @param {Array<string>} roles
+ * @return {Promise<?>}
+ */
+export async function prompt(extensionName: string, projectId: string, roles: string[]) {
+  if (!roles || !roles.length) {
+    return Promise.resolve();
```
bkendall, line 53: Returning a resolved promise is unnecessary in an `async` function.

```suggestion
    return;
```

src/extensions/updateHelper.ts
```diff
@@ -244,6 +167,255 @@ export interface UpdateOptions {
...
+}
+
+/**
+ * Preparatory work for updating an published extension instance to the latest version.
+ *
+ * @param instanceId Id of the instance to update
+ * @param existingSpec ExtensionSpec of the existing instance source
+ * @param existingSource name of existing instance source
+ */
+export async function updateFromRegistry(
```
bkendall, line 414: `async` is not necessary here since you're not `await`-ing anything

bkendall (changes requested):

src/extensions/updateHelper.ts
huangjeff5: Hmmm... Not sure I follow on this one. 

updateFromRegistry returns the output of calling updateToVersionFromRegistry(), which returns a promise. 

updateToVersionFromRegistry() is an async function that needs to await an API call. So as far as I understand, this is still needed?

src/commands/ext-update.ts
huangjeff5: It's actually a slightly different list. I would prefer to continue being explicit here so that someone can analyze the behavior more easily. There's more context shift if they have to look elsewhere, and in this particular case, I think having all the context in one place is nice because this part of the code is quite complicated (and should be refactored/simplified in Q1 when we migrate official extensions on the Registry).

src/commands/ext-install.ts
huangjeff5: Good catch!

src/commands/ext-dev-unpublish.ts
huangjeff5: Sounds good. Updated t/o this branch.

src/commands/ext-dev-unpublish.ts
huangjeff5: Good catch. Removed

src/commands/ext-dev-unpublish.ts
huangjeff5: Good point. This was initially supposed to support two ref formats:
1) extension ref: publisherId/extensionId
2) version ref: publisherId/extensionId@versionId

However we de-scoped un-publishing a specific version because it introduced several edge cases. That's why the code looks like this.

But you're right - I refactored code in `ext:dev:publish` to use parseRef, and pass in extensionRef instead of two values.

src/commands/ext-dev-publish.ts
huangjeff5: Good catch on this, you're right. the `/<extensionId>` part of the was never being parsed or used at all. 

Only one value was ever being passed in to the function, if I tried to make the signature
`.action(async (publisherId: string, extensionId: string, options: any) => {`
it would just pass in `options` value for `extensionId` field, and `options` field itself would just be null.

src/commands/ext-dev-list.ts
huangjeff5: Done!

src/commands/ext-update.ts
huangjeff5: I removed the { } - good call on that! Now I only need to disable eslint once using // eslint-disable-next-line no-fallthrough (rather than copy/paste it several times).

For the // falls through comment, though, I still think I need to disable eslint at least once, because the linter says that you're NOT allowed to write case statements that fall through to the default after falling into a case statement. We want exactly that behavior in this case.

src/extensions/askUserForConsent.ts
huangjeff5: Good catch!

src/extensions/askUserForConsent.ts
```diff
@@ -0,0 +1,100 @@
...
+import { promptOnce, Question } from "../prompt";
+import * as utils from "../utils";
+
+marked.setOptions({
+  renderer: new TerminalRenderer(),
+});
+
+/**
+ * Returns a string that will be displayed in the prompt to user.
+ * @param role
```
joehan, line 18: Let's fix these jsdocs lint issues before we merge - its always easy to forget about these once they're actually merged

src/extensions/displayExtensionInfo.ts
```diff
@@ -1,37 +1,233 @@
...
+  if (spec.billingRequired && !newSpec.billingRequired) {
+    lines.push("", "**Billing is no longer required for this extension.**");
+  }
+  logger.info(marked(lines.join("\n")));
+  return lines;
+}
+
+/**
+ * Checks for spec changes that require explicit user consent,
+ * and individually prompts the user for each changed field
```
joehan, line 127: Missing a period here

src/extensions/displayExtensionInfo.ts
```diff
@@ -1,37 +1,233 @@
...
+          spec: spec,
           extensionName: extensionName,
         },
       }
     );
   }
 }
+
+/**
+ * Prints out all changes to the spec that don't require explicit approval or input
```
joehan, line 68: Missing as period here

src/extensions/displayExtensionInfo.ts
```diff
@@ -1,37 +1,233 @@
...
+  }
+  logger.info(marked(lines.join("\n")));
+  return lines;
+}
+
+/**
+ * Checks for spec changes that require explicit user consent,
+ * and individually prompts the user for each changed field
+ *
+ * @param spec The current spec of a ExtensionInstance
```
joehan, line 129: Missing a period here and on some other jsdocs descriptions

joehan (approved):
Some nits that you could fix up, and I'd like the jsdocs to be completed, but overall I think this is good to go!

src/commands/ext-dev-list.ts
```diff
@@ -0,0 +1,59 @@
...
+    } catch (err) {
+      throw new FirebaseError(err);
+    }
+
+    if (extensions.length < 1) {
+      throw new FirebaseError(
+        `There are no published extensions associated with publisher ID ${clc.bold(
+          publisherId
+        )}. This could happen for two reasons:\n` +
+          "  -The publisher ID doesn't exist or could be misspelled\n" +
```
bkendall, line 33:
```suggestion
          "  - The publisher ID doesn't exist or could be misspelled\n" +
```

src/commands/ext-dev-list.ts
```diff
@@ -0,0 +1,59 @@
...
+          "  -This publisher has not published any extensions\n\n" +
```
bkendall, line 34:
```suggestion
          "  - This publisher has not published any extensions\n\n" +
```

bkendall (approved):

src/extensions/askUserForConsent.ts
huangjeff5: Sounds good

src/extensions/displayExtensionInfo.ts
huangjeff5: I went thru the code and added all periods that I personally noticed were missing.

PR #2882, opened 2020-12-01
Change ESLint no-invalid-this to error.

yuchenshi:
### Description

This PR changes ESLint `no-invalid-this` to error (instead of warning) and fixes the errors.

All changes are noop except one real error caught by the rule, as commented below.

### How to fix the error

If you get `no-invalid-this` error in TypeScript, first see if your use of `this` is safe. If you meant to use `this` from the enclosing scope, change the inner function to an arrow function instead. In rare occasions where you do want to use the `this` passed into the current function (e.g. mocha context), DO NOT `eslint-disable` this rule. Instead, explicitly declare that you're using `this`:

```typescript
    beforeEach("setup", async function(this) {  // <---- add this to enclosing function
      this.timeout(10000); // <---- now safe to use, no eslint errors :)
      // ...
    });
```

In this case, the library already has the correct typing on `beforeEach` callback so you don't need to declare the type of `this`. In case the library typing is funky, use explicit typing like `function(this: MyContextType)`. Obviously, `this` won't work for arrow functions and TypeScript won't let you write `(this) => foo`.

If you get `no-invalid-this` error in JavaScript (plain `.js`), first check to see if you're missing `var self = this;`. If you really meant to use `this` passed into the current function, your only real choice is `// eslint-disable-line no-invalid-this` (and friends).

### Scenarios Tested

`npm test`.

### Sample Commands

N/A

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -39,7 +39,7 @@ module.exports = {
         "@typescript-eslint/no-inferrable-types": "warn", // TODO(bkendall): remove, allow to error.
         "@typescript-eslint/no-misused-promises": "warn", // TODO(bkendall): remove, allow to error.
         "@typescript-eslint/no-unnecessary-type-assertion": "warn", // TODO(bkendall): remove, allow to error.
-        "@typescript-eslint/no-use-before-define": ["warn", { "functions": false, "typedefs": false }], // TODO(bkendall): change to error.
+        "@typescript-eslint/no-use-before-define": ["warn", { functions: false, typedefs: false }], // TODO(bkendall): change to error.
         "@typescript-eslint/no-var-requires": "warn", // TODO(bkendall): remove, allow to error.
         "@typescript-eslint/prefer-includes": "warn", // TODO(bkendall): remove, allow to error.
         "@typescript-eslint/prefer-regexp-exec": "warn", // TODO(bkendall): remove, allow to error.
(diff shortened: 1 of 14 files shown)


src/firestore/delete.js
```diff
@@ -396,7 +396,7 @@ FirestoreDelete.prototype._deletePath = function() {
       }
 
       // For a shallow delete, this error is fatal.
-      return utils.reject("Unable to delete " + clc.cyan(this.path));
+      return utils.reject("Unable to delete " + clc.cyan(self.path));
```
yuchenshi, line 399: This was a real error, fixed in this PR.

.eslintrc.js
```diff
@@ -48,6 +48,11 @@ module.exports = {
         "no-case-declarations": "warn", // TODO(bkendall): remove, allow to error.
         "no-constant-condition": "warn", // TODO(bkendall): remove, allow to error.
         "no-fallthrough": "warn", // TODO(bkendall): remove, allow to error.
+
+        // Use the TS version of the rule which allows `function foo(this) ...`.
+        // https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-invalid-this.md
+        "no-invalid-this": "off",
+        "@typescript-eslint/no-invalid-this": ["error"],
```
bkendall, line 55: Can you move these up to the grouping around L34-35? The first block here are "rules that will stay", the second block is "rules that need to be changed/removed". Adding a third block is not necessary.

I'd also suggest formatting as the following (searching for the documentation is fast - pasting the link seems redundant - and keeping the comment on the line keeps the keys sortable):

```js
        "no-invalid-this": "off", // Turned off in favor of @typescript-eslint/no-invalid-this.
        "@typescript-eslint/no-invalid-this": ["error"],
```

bkendall (changes requested):
Please fix the new lint errors that are surfacing.

.eslintrc.js
yuchenshi: Done.

yuchenshi:
Fixed lint errors and grouping. PTAL!

bkendall (approved):

PR #2885, opened 2020-12-02
Implement batchGet in Auth Emulator.

### LINKED ISSUE #2751: Listing users is not implemented in the Auth Emulator
I just found a feature which isn't implemented and I really need this. I thought I'll submit an issue for this.

**firebase-tools:**
v8.14.1

**Platform:**

MacOS 10.15.7

### [REQUIRED] Test case

```
import * as admin from "firebase-admin"
admin.initializeApp()s
admin.auth().listUsers(10)
```

Try to call `admin.auth().listUsers(10)` 

I'd like to be able to query all users at once (or rather, in batches of 1000) from Firebase Auth.

It crashes with

```
identitytoolkit.projects.accounts.batchGet is not implemented in the Auth Emulator:
```

```
An internal error has occurred. Raw server response: "{"error":{"code":501,"message":"identitytoolkit.projects.accounts.batchGet is not implemented in the Auth Emulator.","errors":[{"message":"identitytoolkit.projects.accounts.batchGet is not implemented in the Auth Emulator.","reason":"unimplemented"}],"status":"NOT_IMPLEMENTED"}}"
```

yuchenshi:
### Description

Now you can use `batchGet`.

### Scenarios Tested

See tests added.

### Sample Commands

N/A

Future consideration: It may be useful to allow the CLI to export accounts from the Auth Emulator too, like `FIREBASE_AUTH_EMULATOR_HOST=localhost:9099 firebase auth:export`. That should export from the emulator instead of production.

Fixes #2751.


src/emulator/auth/operations.ts
```diff
@@ -229,6 +230,33 @@ function lookup(
   };
 }
 
+function batchGet(
+  state: ProjectState,
+  reqBody: unknown,
+  ctx: ExegesisContext
+): Schemas["GoogleCloudIdentitytoolkitV1DownloadAccountResponse"] {
+  const limit = Math.min(parseInt(ctx.params.query.maxResults) || 20, 1000);
```
samhorlbeck, line 238: Don't use `parseInt()` for base-10 parsing. Use `Number()` instead

go/tsstyle#type-coercion

samhorlbeck (approved):

src/emulator/auth/operations.ts
yuchenshi: Like this? But how can we coerce it to an integer then? The code below assumes it's an integer (e.g. `array.length`)
```suggestion
  const limit = Math.min(Number(ctx.params.query.maxResults) || 20, 1000);
```

samhorlbeck: Yes like this. Do `Math.floor(Number(...))`

yuchenshi: Did `Math.floor` only since the param is already converted to a number by exegeisis.

PR #2886, opened 2020-12-02
Make other commands know that `init` can now create default DBs

sarajmunjal:
### Description
This PR
1. Edits the default message printed out when any database command. Instead of pointing the user to the console, we tell them to use firebase init.
2. Adds this check to `database:instances:create` which is used to create secondary instances. If a user tries to create a secondary instance without having a default instance, they will be pointed to `firebase init`.
### Scenarios Tested
`firebase database:get`
`firebase database:instances:create`
[Screenshot](https://screenshot.googleplex.com/Ajxn54hB9gLzD42)


src/commands/database-instances-create.ts
```diff
@@ -22,6 +25,16 @@ export default new Command("database:instances:create <instanceName>")
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   .action(async (instanceName: string, options: any) => {
     const projectId = getProjectId(options);
+    const defaultDatabaseInstance = await getDefaultDatabaseInstance({ project: projectId });
+    if (defaultDatabaseInstance === "") {
+      throw new FirebaseError(
+        `It looks like you haven't created a Realtime Database instance in this project before. Please run ${clc.cyan.bold(
+          "firebase init"
```
rachelsaunders, line 32: Any reason that you're not just telling them to run `firebase init database`?

sarajmunjal: Not specifically. I could do this.

bkendall (approved):

PR #2897, opened 2020-12-04
Always return access tokens in Auth Emulator.

yuchenshi:
You can haz access tokens!


malcolmdeck (approved):
Looks like standard providers don't pass back idTokens, so LGTM

samhorlbeck (approved):

PR #2900, opened 2020-12-04
Release RTDB emulator v4.7.2

IanWyszynski:
Ran npm install and node lib/bin/firebase.js emulators:start --only database


CHANGELOG.md
```diff
@@ -1 +1,2 @@
 - Adds [i18n rewrites](https://firebase.google.com/docs/hosting/i18n-rewrites) to the Firebase Hosting emulator.
+- Release RTDB emulator v4.7.2: Emulator namespace created on the metered plan by default
```
bkendall, line 2: "metered plan" seems a little scary, fwiw, but I don't quite understand what that means either in regard to the emulator.

```suggestion
- Releases RTDB emulator v4.7.2: Emulator namespace created on the metered plan by default.
```

IanWyszynski: Sorry I missed your comment on my first pass. It just means that the limit on the number of connection people can make is now 200k instead of 100. We received a customer case about this last week, and generally it makes sense to not have artificial limits in local environments (for load testing use-cases etc.).

samtstern (approved):

PR #2903, opened 2020-12-04
Functions support for Node.js 14 runtime.

joehan:
From @mbleigh's original PR:

### Description
This change does not add support for ES Modules based functions packages. That is going to be a tricky change because we'll need an ESM version of the trigger parser and TypeScript currently compiles imports into require under the hood. It's not unsolvable but I don't think we need to block basic support on it.

### Scenarios Tested
Create new function with nodejs14 runtime.
Update nodejs12 function to nodejs14.


mbleigh (approved):

mbleigh:
Looks like this will be ready to land on Thursday or Friday this week (barring delays)

laurentpayot:
What version number exactly is the node 14 runtime ? This info is needed the to install the same node version on the local dev machine.

PR #2909, opened 2020-12-07
Make default Firestore rules in the CLI expiring

samhorlbeck:
### Description

Bringing the Firestore default rules into line with the Console (rules now expire after 30 days instead of being `read, write: true`)

### Scenarios Tested

Tested init flow with a new project, an existing project (so that the existing rules _don't_ get overwritten), and with no project.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -18,6 +18,7 @@ import { Socket } from "net";
 const IS_WINDOWS = process.platform === "win32";
 const SUCCESS_CHAR = IS_WINDOWS ? "+" : "✔";
 const WARNING_CHAR = IS_WINDOWS ? "!" : "⚠";
+const THIRTY_DAYS_IN_MILLISECONDS = 30 * 24 * 60 * 60 * 1000;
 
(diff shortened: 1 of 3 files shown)


samtstern (approved):

yuchenshi (approved):

src/utils.ts
```diff
@@ -506,3 +507,10 @@ export function isCloudEnvironment() {
 export function isRunningInWSL(): boolean {
   return !!process.env.WSL_DISTRO_NAME;
 }
+
+/**
+ * Generates a date that is 30 days from Date.now()
```
yuchenshi, line 512:
```suggestion
 * Generates a date that is 30 days from Date.now() in UTC
```

samhorlbeck: I don't think this is necessary-- JS dates don't have a timezone, they're just milliseconds since 1970/01/01 UTC. As I understand it, most methods on Date objects operate in the local timezone, while there are some methods that will convert / use UTC

yuchenshi: Good point

PR #2919, opened 2020-12-10
update Functions template files

### LINKED ISSUE #2913: Should functions contain a .eslintignore file?

**firebase-tools:** 8.17.0

**Platform:** macOS Big Sur 11.0.1

### [REQUIRED] Test case

See repository at https://github.com/hithomasmorelli/vscode-eslint-parseroptions-project-repro

Follow the steps in the README.md file within the repository linked above

As `npm run lint` is set to only run eslint on `"src/**/*"`, I think it would be expected for (or, at least, I expected) the VS Code ESLint extension to lint those same files.

However, the VS Code ESLint extension seems to work by calling `eslint .` on the specified working directory. This leads to the following problem to display within VS Code when opening `functions/.eslintrc.js` (as, if I understand correctly, only the `src` directory - which does not include `.eslintrc.js` - is `include`d within `tsconfig.json`):

101280324-6e608580-37c0-11eb-8223-3c229d51ee40

I raised issue microsoft/vscode-eslint#1135 in relation to this, but it was closed as the extension working as expected - as the same result was obtained by running `eslint .` from the terminal.

I then [enquired](https://github.com/microsoft/vscode-eslint/issues/1135#issuecomment-740721766) as to whether or not there was a way to tell the VS Code ESLint extension to only match a specified pattern (`"src/**/*"`) within the `functions` working directory - to which the answer was ["`.eslintignore`"](https://github.com/microsoft/vscode-eslint/issues/1135#issuecomment-741650798).

So, here is the dilemma. Should the default `functions` setup include an `.eslintignore` file, in order to prevent VS Code (and any other editors/extensions that may use ESLint in this way) from flagging files that we don't actually intend to be linted? And, if an `.eslintignore` file *is* added, and set up so that it excluded everything except for `"src/**/*"`, should the command that is run for `npm run lint` be changed to just `eslint .`?

Interested to hear people's thoughts on this.

bkendall:
### Description

1. Updates both JS and TS eslint setups to be the same (same template name) and _greatly_ simplify their configurations.
2. Adds a `tsconfig.dev.json` to the TS+eslint path to appropriately lint `.eslintrc.js`.

Fixes #2913

As an aside: I wouldn't be opposed to simplifying this further and not offering the no-lint path. Also, we don't have any script for `npm test`, so there's nothing _blocking_ about always including lint (except your IDE might start flagging issues sooner). Food for thought...

### Scenarios Tested

- JS with and without lint
- TS with and without lint

### Sample Commands

`firebase init functions`


samtstern:
@bkendall FWIW I always choose the no-lint path.  I find JS linters to be mostly annoying, but I am maybe in the minority.

samtstern (approved):

templates/init/functions/typescript/package.lint.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "functions",
   "scripts": {
-    "lint": "eslint \"src/**/*\"",
+    "lint": "eslint --ext .js,.ts .",
```
joehan, line 4: Will this cause eslint to lint compiled js files? I think we don't want that.

joehan (commented):
One issue, but otherwise this LGTM

templates/init/functions/typescript/package.lint.json
hithomasmorelli: @joehan I don't think it would, as JS files (apart from `.eslintrc.js`) are listed in the [`.gitignore`](https://github.com/firebase/firebase-tools/tree/bk-2913/templates/init/functions/typescript/_gitignore)

bkendall: (I've been working on coming back to this but @hithomasmorelli is correct)

joehan (approved):

PR #2932, opened 2020-12-14
Fixes #2931 - Support multiple Set-Cookie headers

### LINKED ISSUE #2931: Emulated Hosting redirects to Cloud Functions erroneously combines Set-Cookie headers

**firebase-tools:** 8.18.1

**Platform:** OSX

### [REQUIRED] Test case

Write a function that sets multiple cookies:

```js
exports.myFunc = require('firebase-functions').https.onRequest((req, res) => {
    res.setHeader('Set-Cookie', [
        'some=cookie',
        'other=one'
    ]);
    res.statusCode = 200;
    res.end('');
});
```

Use Hosting as a proxy to the function. `firebase.json`:

```json
{
    "hosting": {
        "rewrites": [
            { "source": "/wherever", "function": "myFunc" }
        ]
    }
}
```

Hit the proxied function URL: `http://myhostingsubdomain.blah.com/wherever`. Check the response headers in Chrome DevTools and check the Application Cookies - only the first will be registered.

The response should include multiple Set-Cookie headers - one for each cookie. Browsers (including Chrome) don't handle a single Set-Cookie header with multiple cookies in it and will ignore all but the first cookie (especially with cookie fields like Expires which can include commas in them). This prevents usage like https://dev.to/johncarroll/how-to-share-firebase-authentication-across-subdomains-1ka8 where your function has to be hosted at the same origin as your Hosting sites (so cloudfunctions.net/blah is no good).

I expect to see response headers like:

```http
Set-Cookie: some=cookie
Set-Cookie: other=one
```

The response has a single header like:

```http
set-cookie: some=cookie, other=one
```

### Cause

`node-fetch` is being used incorrectly, indirectly from `lib/hosting/proxy.js` via `lib/apiv2.js`'s `Client::request`. See https://github.com/node-fetch/node-fetch/issues/582 for a similar case. The headers are being extracted from `proxyRes.response.headers` in `lib/hosting/proxy.js`:

```js
for (const [key, value] of proxyRes.response.headers) {
    res.setHeader(key, value);
}
```

But because `node-fetch` stringifies the incoming response headers (which for an Array, effectively concatenates them with commas), we lose our array of headers in the proxied response. The recommended approach is to use `response.headers.raw()` to grab those raw Array `value`s. When I make that change in `lib/hosting/proxy.js`, it works:

```js
for (const [key, value] of Object.entries(proxyRes.response.headers.raw())) {
    res.setHeader(key, value);
}
```

I have no yet validated whether the same problem happens in production or not. Really hoping it doesn't 🤞
UPDATE: Production is fine! This is an emulator-only problem.

adam-remotesocial:
apiv2.js uses node-fetch under the hood, which will concatenate duplicate headers with ', '. This is invalid for Set-Cookie headers.
The solution is to grab the raw headers (which will be an array) and call `res.setHeader()` with that value.

### Description

Fixing #2931. I've not run npm test yet.

### Scenarios Tested

I've tested that this solves the Set-Cookie problem locally. I haven't tested any potential edge cases (what might they be? Is there a case where the raw value is dangerous?)

### Sample Commands

No changes to commands.

### DIFF
--- a/src/hosting/proxy.ts
+++ b/src/hosting/proxy.ts
@@ -139,7 +139,7 @@ export function proxyRequestHandler(url: string, rewriteIdentifier: string): Req
 
     proxyRes.response.headers.set("vary", makeVary(proxyRes.response.headers.get("vary")));
 
-    for (const [key, value] of proxyRes.response.headers) {
+    for (const [key, value] of Object.entries(proxyRes.response.headers.raw())) {
       res.setHeader(key, value);
     }
     res.statusCode = proxyRes.status;


adam-remotesocial:
@googlebot I signed it!

bkendall:
Thanks for catching this! Migrating the underlying request-ing library away from `request` has been such the adventure in discovering all the little things that `request` used to "just handle".

I've made a test for this, added a changelog, and have pushed it as another branch `adam-remotesocial-patch-1` on this repo. I'm planning on merging that branch unless you want to (relatively quickly - other wheels are turning) pull in those extra commits to your fork and update this PR. Otherwise, you'll still get the credit, but the commit will come from a slightly different branch.

bkendall:
Merged via #2939

adam-remotesocial:
Hey thanks a ton! All I needed was a fix, but I appreciate your sensitivity to people who need the credit! Thanks for the super quick turnaround!

PR #2935, opened 2020-12-15
Removing code to handle legacy functions

joehan:
### Description
Removing code that supported legacy functions triggers and functions deployed prior to version 3.17.6 of firebase-tools.

 The impact of removing this is that functions with legacy trigger types will show an uglier error, and functions deployed prior to v3.17.6 won't be recognized as Firebase functions (and therefore won't be deleted).

I tested normal function deploys, but I didn't have any legacy triggers or old functions to test this on.


mbleigh (approved):
Changed base to `next` since I think that's where we're staging 9.0.0 things.

PR #2938, opened 2020-12-15
Improve Firestore delete

samtstern:
### Description

Started with an investigation into large document deletes, made a few changes:

- Move the "Transaction too big" error handling code into the operation so that it can be handled visibly
- Before doing n parallel delete batches wait until one succeeds. For cases where all documents are large this means we'll get the right batch size sooner.
- Fix some logging.

TODO:

- [x] Merge #2940
- [x] Address https://github.com/firebase/firebase-tools/pull/2940#discussion_r543714192

### Scenarios Tested

Wrote a small script to add 1000 big docs and 1000 small docs to collections:

```
$ time node index.js 
Adding 1000 small documents
Done
Adding 1000 big documents
Done

real    0m49.459s
user    0m6.877s
sys     0m0.842s
```

Deleting the small docs works as always:

```
$ firebase --project=fir-dumpster firestore:delete --recursive --yes small-data
Deleted 250 docs (250000 docs/s)
Deleted 500 docs (810 docs/s)
Deleted 750 docs (934 docs/s)
Deleted 1000 docs (1037 docs/s)
```

Deleting the big docs hits the new error handling and succeeds (slowly):

```
$ firebase --project=fir-dumpster firestore:delete --recursive --yes big-data 
⚠  firestore: delete transaction too large, reducing batch size from 250 to 25
Deleted 25 docs (Infinity docs/s)
Deleted 50 docs (8 docs/s)
Deleted 75 docs (11 docs/s)
Deleted 100 docs (15 docs/s)
Deleted 125 docs (18 docs/s)
Deleted 150 docs (21 docs/s)
Deleted 175 docs (24 docs/s)
Deleted 200 docs (27 docs/s)
Deleted 225 docs (29 docs/s)
Deleted 250 docs (32 docs/s)
Deleted 275 docs (34 docs/s)
Deleted 300 docs (37 docs/s)
Deleted 325 docs (39 docs/s)
Deleted 350 docs (42 docs/s)
Deleted 375 docs (39 docs/s)
Deleted 400 docs (39 docs/s)
Deleted 425 docs (39 docs/s)
Deleted 450 docs (35 docs/s)
Deleted 475 docs (37 docs/s)
Deleted 500 docs (35 docs/s)
Deleted 525 docs (37 docs/s)
Deleted 550 docs (37 docs/s)
Deleted 575 docs (39 docs/s)
Deleted 600 docs (39 docs/s)
Deleted 625 docs (40 docs/s)
Deleted 675 docs (42 docs/s)
[...]

PR #2940, opened 2020-12-15
Convert firestore/delete to TypeScript

samtstern:
### Description

See: https://github.com/firebase/firebase-tools/pull/2938/files

I need to fix a bug in this file but want to TS-ify it first for confidence.  This commit does the bare minimum to get it into `.ts` and then #2938 cleans it up more and fixes the bug,

### Scenarios Tested

N/A

### Sample Commands

N/A


src/firestore/delete.ts
```diff
@@ -0,0 +1,486 @@
+"use strict";
```
bkendall, line 1: you can drop use strict in ts files.... wonder if there's a lint rule for that...

src/firestore/delete.ts
```diff
@@ -0,0 +1,486 @@
...
+// Datastore allowed numeric IDs where Firestore only allows strings. Numeric IDs are
+// exposed to Firestore as __idNUM__, so this is the lowest possible negative numeric
+// value expressed in that format.
+const MIN_ID = "__id-9223372036854775808__";
+
+export class FirestoreDelete {
+  /**
+   * Progress bar shared by the class.
+   */
+  static progressBar = new ProgressBar("Deleted :current docs (:rate docs/s)\n", {
```
bkendall, line 21: I'm a little confused why this is a static on the class, to be honest. Consider when improving the logic to remove it or make it a local variable where it's used...

src/firestore/delete.ts
```diff
@@ -0,0 +1,486 @@
...
+
+  /**
+   * Construct a new Firestore delete operation.
+   *
+   * @constructor
+   * @param {string} project the Firestore project ID.
+   * @param {string | undefined} path path to a document or collection.
+   * @param {boolean} options.recursive true if the delete should be recursive.
+   * @param {boolean} options.shallow true if the delete should be shallow (non-recursive).
+   * @param {boolean} options.allCollections true if the delete should universally remove all collections and docs.
```
bkendall, line 46: All the types can be dropped since you now type them in the function signature! Applies throughout.

src/firestore/delete.ts
```diff
@@ -0,0 +1,486 @@
...
+  private allCollections: boolean;
+
+  private allDescendants: boolean;
+  private root: string;
+  private parent: string;
+
+  /**
+   * Construct a new Firestore delete operation.
+   *
+   * @constructor
```
bkendall, line 41: Drop constructor - it's understood in a Class

src/firestore/delete.ts
```diff
@@ -0,0 +1,486 @@
...
+    // When --all-collections is passed any other flags or arguments are ignored
+    if (!options.allCollections) {
+      this._validateOptions();
+    }
+  }
+
+  /**
+   * Validate all options, throwing an exception for any fatal errors.
+   */
+  private _validateOptions() {
```
bkendall, line 90: Private functions don't need to be prefixed with underscores. Applies throughout.

src/firestore/delete.ts
```diff
@@ -0,0 +1,486 @@
...
+    const url = this.parent + ":runQuery";
+    let body;
+    if (this.isDocumentPath) {
+      body = this._docDescendantsQuery(allDescendants, batchSize, startAfter);
+    } else {
+      body = this._collectionDescendantsQuery(allDescendants, batchSize, startAfter);
+    }
+
+    return api
+      .request("POST", "/v1beta1/" + url, {
```
bkendall, line 251: When you improve this code, consider using `apiv2.Client`'s `post` (or `request`, if a lot of custom handling is required) method!

bkendall (approved):
LGTM from a typescript point of view - a few comments, but if you clean some of them up now or later I'm not too concerned as long as they're eventually addressed :)

src/firestore/delete.ts
samtstern: I'll do this one in the other PR

PR #2947, opened 2020-12-17
Support batchCreate.

yuchenshi:
### Description

This adds support for batchCreate, with a few corner cases intentionally omitted for sanity. One step closer to import/export!

### Scenarios Tested

Comprehensive spec test added.


src/emulator/auth/operations.ts
```diff
@@ -230,6 +231,157 @@ function lookup(
...
+          // These providers are handled automatically by create / update.
+          if (providerId !== PROVIDER_PASSWORD && providerId !== PROVIDER_PHONE) {
+            assert(
+              providerId && rawId,
+              "((Auth Emulator only supports providerUserInfo with providerId AND rawId))"
+            );
+            const existingUserWithProviderRawId = state.getUserByProviderRawId(providerId, rawId);
+            assert(
+              !existingUserWithProviderRawId,
+              `((Auth Emulator does not support importing duplicate Provider id(${providerId}), Raw id(${rawId})))`
```
yuchenshi, line 319: This is not the production error message. I'll add a test and fix this before I merge.

src/emulator/auth/operations.ts
```diff
@@ -230,6 +231,157 @@ function lookup(
...
+      if (fields.createdAt && !isNaN(Number(userInfo.createdAt))) {
+        fields.createdAt = userInfo.createdAt;
+      }
+      if (userInfo.email) {
+        const email = userInfo.email;
+        assert(isValidEmailAddress(email), "email is invalid");
+        const existingUserWithEmail = state.getUserByEmail(email);
+        assert(
+          !existingUserWithEmail || existingUserWithEmail.localId === userInfo.localId,
+          `((Auth Emulator does not support importing duplicate email: ${email}))`
```
yuchenshi, line 345: TODO: Production catches this case only if `sanityCheck && emailRequired`. We should use production error message in that case, otherwise throw this emulator specific error message (since we do not want to support this edge case)

samtstern (approved):
LGTM assuming you resolve your own TODOs

PR #2952, opened 2020-12-17
Remove node version checks in tests

samtstern:
### Description

Follow up to #2620, undoes the testing hack from #2754

Note that the Hosting integration test actually **did not run** as part of the v9 release because the env var check was looking for Node 8 only.

### Scenarios Tested

N/A

### Sample Commands

N/A


scripts/triggers-end-to-end-tests/tests.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable prefer-arrow-callback */
```
samtstern, line 1: This lint rule did not play nice with tests that need the mocha callbacks with `(this)`

bkendall: It looks like this can be removed and the test "should have have triggered cloud functions" can be changed to an arrow function (it specifies `this` as an argument to `function` but doesn't use `this`, so that's why it's failing).

bkendall (changes requested):

samtstern:
@bkendall good catch on the lint thing, fixed.  I'm gonna step away from the computer now (and start the long weekend).  If there's anything else to do on this PR feel free to just fork my branch and abandon this one since this PR is (imo) relatively important and I don't want to block it.

bkendall (approved):

Issue #2782, opened 2020-11-05
Firebase database emulator crashes when using with bolt rules


SamyPesse:
Running `firebase start:emulator` doesn't seem to support bolt file anymore.

### Environment info

firebase-tools v8.15.0 on Mac (it was working with 8.6.0)

### Steps to reproduce

1. Write some bolt rules
2. Configure it in your `firebase.json` file:
    ```json
    {
      "database": {
        "rules": "./packages/app-database/rules/realtime.bolt"
      }
    }
    ```
3. Run `firebase emulators:start --only database --debug`

It crashes with:

```
[2020-11-05T14:34:23.171Z] <<< HTTP RESPONSE 400 {"content-length":"38","content-type":"application/json; charset=utf-8","access-control-allow-origin":"*","cache-control":"no-cache","x-firebase-project-id":"projectId","x-firebase-project-number":"","x-firebase-uuid":""}
[2020-11-05T14:34:23.171Z] <<< HTTP RESPONSE BODY {
  "error" : "5:1: Parse error.\n"
```

### Expected behavior

Firebase should parse the bolt file before updating the rule in the emulator.

### Actual behavior

It crashes.

samtstern:
@SamyPesse thanks for reporting this!  You're right, we had not considered this situation properly.  We'll need to make the rules loader compile bolt rules.

Andradeerik:
I thought I had the same problem but in my (database.rules.json) rules it was false.
:)

Issue #767, opened 2018-05-13
Firebase-tools should support a local firebase-bolt


mwaldstein:
### Version info

3.18.4

### Platform Information

Linux

### Steps to reproduce

1. Create a firebase project
2. Create firebase/database.rules.bolt
3. Configure firebase-db to use the bolt rules:
`{
  "database": {
    "rules": "firebase/database.rules.bolt"
  }
}`
4. Install firebase-tools and firebase-bolt as project dependencies
`npm i firebase-tools firebase-bold`
5. Attempt to deploy

### Expected behavior

The bolt rules are compiled and loaded to the project

### Actual behavior

An error is thrown that firebase-bolt must be installed globally

### Notes:

* The spawn of firebase-bolt only looks in the path: https://github.com/firebase/firebase-tools/blob/02f5e2eec9c558253e193e577fa45eeba1d9329a/lib/parseBoltRules.js

samtstern:
I wonder if we could fix this by calling `npx firebase-bolt` which would look either locally or globally?

PR #2962, opened 2020-12-21
Fix bolt issues

samtstern:
### Description

- Fixes https://github.com/firebase/firebase-tools/issues/2782
- Fixes https://github.com/firebase/firebase-tools/issues/767
 
### Scenarios Tested

- `firebase deploy` with Bolt rules
- `firebase deploy` with JSON rules
- `firebase emulators:start` with Bolt rules
  - and hot reload
- `firebase emulators:start` with JSON rules
  - and hot reload

All of the above tested with `firebase-bolt` only installed in my local `devDependencies`, not globally installed.

### Sample Commands

N/A


src/emulator/databaseEmulator.ts
```diff
@@ -153,7 +153,13 @@ export class DatabaseEmulator implements EmulatorInstance {
     });
   }
 
-  private async updateRules(instance: string, content: string): Promise<any> {
+  private async updateRules(instance: string, rulesPath: string): Promise<any> {
+    const rulesExt = path.extname(rulesPath);
+    const content =
+      rulesExt === ".bolt"
+        ? parseBoltRules(rulesPath).toString()
+        : fs.readFileSync(rulesPath, "utf8").toString();
```
bkendall, line 161: IIRC, `readFileSync(path, "utf8")` returns a string, so `.toString()`-ing it is redundant

```suggestion
        : fs.readFileSync(rulesPath, "utf8");
```

bkendall (approved):

PR #2972, opened 2020-12-27
Auth Emulator: Google Sign In auto-generated user information email change

eligeske:
The `auto-generate user information` is awesomely simple and helpful. Thanks!

One small issue is the domain chosen for the email address. It is a valid domain and not recommended for testing purposes. For open-source content the IANA-managed Reserved Example Domains are preferred. https://www.iana.org/domains/reserved
(Should use IANA-managed like `example.com` instead of actual registered domains, such as `test.com`)

image

### Description

- changes auto-generated user information email domain `test.com` to IANA-managed Reserved Domains `example.com`

### Scenarios Tested

- Auto-generated user information via the auth provider popup from local auth emulator.

### Sample Commands

NA


samtstern (approved):

samtstern:
@eligeske thank you for this clear description and the change!  I agree, we should be using `example.{com,org}` domains.

PR #2979, opened 2020-12-28
Fix incorrect databaseUrl in Functions emulator

### LINKED ISSUE #2965: Hosting emulator uses a different RTDB instance from Functions and Database emulators.

**firebase-tools:**
9.1.0

**Platform:** Raspbian
Linux raspberrypi 5.4.79-v7l+

### [REQUIRED] Test case

functions/index.js

```js
const admin = require("firebase-admin");

admin.initializeApp();

exports.addMessage = functions.https.onRequest(async (req, res) => {
    const original = req.query.text;
    const messageRef = await admin.database().ref('/messages').push();
    await messageRef.set({original: original});
    res.json({result: `Message with ID: ${messageRef.key} added.`});
  });
```

public/index.html
[database-debug.log](https://github.com/firebase/firebase-tools/files/5740506/database-debug.log)
[firebase-debug.log](https://github.com/firebase/firebase-tools/files/5740507/firebase-debug.log)
[pubsub-debug.log](https://github.com/firebase/firebase-tools/files/5740508/pubsub-debug.log)
[ui-debug.log](https://github.com/firebase/firebase-tools/files/5740509/ui-debug.log)

```html
<html>
<head>
  <meta charset="utf-8">
  <title>DB Instance Name Bug</title>
  <script defer src="/__/firebase/8.2.1/firebase-app.js"></script>
  <script defer src="/__/firebase/8.2.1/firebase-database.js"></script>
  <script defer src="/__/firebase/init.js?useEmulator=true"></script>
</head>

<body>
  <script>
    document.addEventListener('DOMContentLoaded', function () {
      firebase.database().ref('/messages').once('value').then(
        (snapshot) => console.log(snapshot.val()),
        (error) => console.log(error)
      );
    });
  </script>
</body>
</html>
```

1. Create a new project in the Firebase console.
2. Make a new local directory for the project with `firebase init`.
   A. Select the project from (1). 
   B. Select database, hosting, emulators.
   C. Accept all defaults.
3. Create public/index.html and functions/index.js as above
4. Run `firebase emulators:start`
5. Invoke the addMessage function: curl http://localhost:5001/*project-id*/us-central1/addMessage?text=test
6. Load the hosted website: `http://localhost:5000/`

All three emulators, Functions, Database, and Hosting should all refer to the same instance.

When the page is loaded in (6), an object with the message from (5) should be logged to the console. 

I have verified that this project works as intended when deployed to production.

The Functions emulator and the Database emulator both use http://localhost:9000?ns=*project-id*, but the Hosting emulator connects to ws://localhost:9000/.ws?v=5&ns=*project-id*-default-rtdb 

The Hosting emulator uses a database with the "-default-rtdb" suffix, while the other two emulators do not use that suffix.

When the page is loaded in (6), a "null" is logged to the console.

After loading the page, the database emulator shows two databases, one with and one without the "-default-rtdb" suffix.

samtstern:
### Description

Fixes #2965

Due to the new location support in Realtime Database it is no longer safe to assume that `https://<PROJECT-ID>.firebaseio.com` is the default RTDB instance location.  This PR uses the `getAdminSdkConfig` REST endpoint to fetch the real database URL and pass it down to the Functions emulator.

In order to make this behavior work offline there are two fallbacks:

  - Successful fetches populate a local cache which can be used when offline.
  - If the cache is empty we can try and guess the Admin SDK config using the old (and not so good) method.

This fallback behavior is based on what we do with web config in the Hosting emulator. 

### Scenarios Tested

See the setup in https://github.com/firebase/firebase-tools/issues/2965.  I created a brand new project and created the defalt RTDB through `firebase init`.

First add a message using the Cloud Function:
```
$ http http://localhost:5001/fir-tools-2965/us-central1/addMessage?text="hello"
HTTP/1.1 200 OK
connection: keep-alive
content-length: 57
content-type: application/json; charset=utf-8
date: Mon, 28 Dec 2020 18:14:23 GMT
etag: W/"39-bGCJNRIDTzsIri/GyVcXPXLVrbs"
x-powered-by: Express

{
    "result": "Message with ID: -MPeYivLKiIxoGNp2Eil added."
}
```

I can see that the data goes to the right place in the Emulator UI:
Screen Shot 2020-12-28 at 6 15 19 PM

Finally if I load the hosting emulator I can see the correct logs:
Screen Shot 2020-12-28 at 6 16 21 PM

### Sample Commands

N/A

### DIFF
--- a/src/emulator/functionsEmulatorRuntime.ts
+++ b/src/emulator/functionsEmulatorRuntime.ts
@@ -663,7 +663,7 @@ async function initializeEnvironmentalVariables(frb: FunctionsRuntimeBundle): Pr
       ns = asUrl.hostname.split(".")[0];
     }
 
-    emulatedDatabaseURL = `http://${formatHost(frb.emulators.database)}?ns=${ns}`;
+    emulatedDatabaseURL = `http://${formatHost(frb.emulators.database)}/?ns=${ns}`;
   }
 
   process.env.FIREBASE_CONFIG = JSON.stringify({
--- a/src/emulator/functionsEmulatorRuntime.ts
+++ b/src/emulator/functionsEmulatorRuntime.ts
@@ -654,20 +654,22 @@ async function initializeEnvironmentalVariables(frb: FunctionsRuntimeBundle): Pr
   const functionsGt380 = compareVersionStrings(functionsResolution.version, "3.8.0") >= 0;
   let emulatedDatabaseURL = undefined;
   if (frb.emulators.database && functionsGt380) {
-    emulatedDatabaseURL = `http://${formatHost(frb.emulators.database)}?ns=${
-      process.env.GCLOUD_PROJECT
-    }`;
+    // Database URL will look like one of:
+    //  - https://${namespace}.firebaseio.com
+    //  - https://${namespace}.${location}.firebasedatabase.app
+    let ns = frb.projectId;
+    if (frb.adminSdkConfig.databaseURL) {
+      const asUrl = new URL(frb.adminSdkConfig.databaseURL);
+      ns = asUrl.hostname.split(".")[0];
+    }
+
+    emulatedDatabaseURL = `http://${formatHost(frb.emulators.database)}?ns=${ns}`;
   }
 
-  // Do our best to provide reasonable FIREBASE_CONFIG, based on firebase-functions implementation
-  // https://github.com/firebase/firebase-functions/blob/59d6a7e056a7244e700dc7b6a180e25b38b647fd/src/setup.ts#L45
   process.env.FIREBASE_CONFIG = JSON.stringify({
-    databaseURL:
-      process.env.DATABASE_URL ||
-      emulatedDatabaseURL ||
-      `https://${process.env.GCLOUD_PROJECT}.firebaseio.com`,
-    storageBucket: process.env.STORAGE_BUCKET_URL || `${process.env.GCLOUD_PROJECT}.appspot.com`,
-    projectId: process.env.GCLOUD_PROJECT,
+    storageBucket: frb.adminSdkConfig.storageBucket,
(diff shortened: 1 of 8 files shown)


src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -654,20 +654,22 @@ async function initializeEnvironmentalVariables(frb: FunctionsRuntimeBundle): Pr
...
+    // Database URL will look like one of:
+    //  - https://${namespace}.firebaseio.com
+    //  - https://${namespace}.${location}.firebasedatabase.app
+    let ns = frb.projectId;
+    if (frb.adminSdkConfig.databaseURL) {
+      const asUrl = new URL(frb.adminSdkConfig.databaseURL);
+      ns = asUrl.hostname.split(".")[0];
+    }
+
+    emulatedDatabaseURL = `http://${formatHost(frb.emulators.database)}?ns=${ns}`;
```
yuchenshi, line 666:
```suggestion
    emulatedDatabaseURL = `http://${formatHost(frb.emulators.database)}/?ns=${ns}`;
```

yuchenshi (approved):

cjam:
@samtstern I recently updated the tools locally and found that I cant get my local database rules to apply to the correct instance anymore.  I have my `database.rules.json` which automatically gets applied to `projectId` but now the actual database instance id is pulled in from the cloud.  How do I configure the local database rules / firebase.json so that the rules apply to this database instance?

samtstern:
@cjam can you file a new issue about this and explain in a little more detail?  Definitely sounds like a bug.

cjam:
@samtstern I can.  I'm a bit busy at the moment to create all of the required details for the bug report but I'll try and get to it in the next few days.

PR #3010, opened 2021-01-08
upgrade prettier

bkendall:
### Description

*Note, this will follow the `eslint` upgrade PR*

upgrade `prettier` and apply the formatter to the codebase. cleaned up the config as well to remove values that are now defaults.

I think the thing that's the biggest change is that `function()` is now `function ()`. This _only_ is true for anonymous functions using the `function` keyword - arrow functions don't require this 'cause arrow functions are better anyway.


yuchenshi (approved):
Rubber stamp

bkendall:
(hmmm... chose to rebase against master and force-push instead of merging (since I couldn't do it automatically in the UI). not sure how I feel about that, but there it is)

PR #3019, opened 2021-01-12
update GH action versions

bkendall:
### Description

Updating the GH action versions of:
- checkout
- setup-node
- cache


joehan (approved):
LGTM if all the CI passes

PR #3024, opened 2021-01-12
Catch auth channel update on channel deletion

bkendall:
### Description

When we delete channels, if we are unable to update the authorized domains list, we should catch the error and continue, like we do on create.

While I'm here, migrating the api file to apiv2.

### Scenarios Tested

- `hosting:channel:delete` after both a `create` and `deploy` workflow.

### DIFF
--- a/src/gcp/auth.ts
+++ b/src/gcp/auth.ts
@@ -1,24 +1,34 @@
-import * as api from "../api";
+import { Client } from "../apiv2";
+import { identityOrigin } from "../api";
 
+const apiClient = new Client({ urlPrefix: identityOrigin, auth: true });
+
+/**
+ * Returns the list of authorized domains.
+ * @param project project identifier.
+ * @return authorized domains.
+ */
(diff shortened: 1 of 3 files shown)


src/gcp/auth.ts
```diff
@@ -1,24 +1,34 @@
...
 export async function getAuthDomains(project: string): Promise<string[]> {
-  const res = await api.request("GET", `/admin/v2/projects/${project}/config`, {
-    auth: true,
-    origin: api.identityOrigin,
-  });
-  return res?.body?.authorizedDomains;
+  const res = await apiClient.get<{ authorizedDomains: string[] }>(
+    `/admin/v2/projects/${project}/config`
+  );
+  return res.body.authorizedDomains;
```
dahliasalem, line 15: Nice, I forgot we can just catch the error instead of using optional chains

dahliasalem (approved):

PR #3030, opened 2021-01-13
Fix auth deploy

dahliasalem:
Fixed issue where auth domain was not working properly because it was inside an asynchronous block. Made it synchronous.

Added debug log output to confirm which domains get added to the authorized domains list


src/commands/hosting-channel-deploy.ts
```diff
@@ -151,7 +131,28 @@ export default new Command("hosting:channel:deploy [channelId]")
 
       logger.info();
       const deploys: { [key: string]: ChannelInfo } = {};
-      sites.forEach((d) => {
+      for (const d of sites) {
+        try {
+          await addAuthDomain(projectId, d.url);
```
bkendall, line 136: Would it behoove us to write an `addAuthDomains` function that allowed us to add more than one at once? Then we would only have to make one update of the list.

src/commands/hosting-channel-deploy.ts
```diff
@@ -151,7 +131,28 @@ export default new Command("hosting:channel:deploy [channelId]")
...
+            )
+          );
+          logger.debug("[hosting] unable to add auth domain", e);
+        }
+        try {
+          await cleanAuthState(projectId, d.site);
+        } catch (e) {
+          logLabeledWarning(LOG_TAG, "Unable to sync Firebase Auth state.");
+          logger.debug("[hosting] unable to sync auth domain", e);
+        }
```
bkendall, line 155: I'd consider moving this into a helper function just to clean up this massive function body we've built...

CHANGELOG.md
```diff
@@ -1,2 +1,3 @@
 - Catches errors while updating authorized domains when deleting channels, printing a warning instead of failing.
 - Fixes issue where `host` header was being incorrectly set when proxying to Cloud Run or Cloud Functions for Firebase from the Hosting emulator. (#3012)
+- Fixed domain authorization to be synchronous in hosting channel deploy.
```
bkendall, line 3:
```suggestion
- Fixes issue where authorized domains were not being correctly updated when deploying to Hosting channels. (#3002)
```

bkendall (changes requested):

src/commands/hosting-channel-deploy.ts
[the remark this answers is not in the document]
dahliasalem: I dont see the problem with doing it after the deploy after all there is no point in adding domains to the auth list if the deploy was not successful

src/hosting/api.ts
```diff
@@ -432,9 +434,17 @@ export async function getCleanDomains(project: string, site: string): Promise<st
...
+ * @param sites the list of sites for the channel.
  */
-export async function cleanAuthState(project: string, site: string): Promise<string[]> {
-  const authDomains = await getCleanDomains(project, site);
-  return await updateAuthDomains(project, authDomains);
+export async function cleanAuthState(
+  project: string,
+  sites: string[]
+): Promise<Map<string, Array<string>>> {
+  let siteDomainMap = new Map();
```
bkendall, line 443:
```suggestion
  const siteDomainMap = new Map();
```

Good use of Map too :)

src/hosting/api.ts
```diff
@@ -353,18 +353,20 @@ export async function createRelease(
...
+ * @param urls the list of urls of the channel.
  */
-export async function addAuthDomain(project: string, url: string): Promise<string[]> {
+export async function addAuthDomains(project: string, urls: string[]): Promise<string[]> {
   const domains = await getAuthDomains(project);
-  const domain = url.replace("https://", "");
   const authDomains = domains || [];
-  if (authDomains.includes(domain)) {
-    return authDomains;
+  for (let url of urls) {
```
bkendall, line 363:
```suggestion
  for (const url of urls) {
```

bkendall (approved):
Approval with a couple nits. Looks good! :)

PR #3035, opened 2021-01-14
move emulator suite tests into scripts folder

bkendall:
### Description

These tests aren't really "unit" tests as much as they are "integration" or "functional" tests. They take ~500ms each to run, so taking them out of our "unit" test path would speed them up greatly (down to 9s, in fact).

This does a little bit of fancy footwork. The process for these particular tests is this:
- build the scripts and src directories into a `dev` directory
- run the tests in `dev/scripts` using mocha
- clean up `dev`

It's a little odd, but it allows the fact that it spawns child processes using node and the generated code, making it a bit easier.

The reason I _don't_ want to do this with all the unit tests (even though these tests still take ~10s, which doubles the unit tests runtime) is that I lose the source mappings and coverage information. For some reason, these tests fail in strange ways when I generate the source maps alongside the code (which is why they're disabled).


bkendall:
Yes, the coverage went down, but there's not much I can do about that at the moment. New, more unity tests would have to be written for some of the emulator parts.

scripts/emulator-tests/run.sh
```diff
@@ -0,0 +1,19 @@
...
+# Clean the destination of the upcoming build.
+rm -rf dev
+# Run a special build for these tests and the source code.
+tsc --build scripts/emulator-tests/tsconfig.dev.json
+# Need to copy `package.json` to the directory so it can be referenced in code.
+cp package.json dev/package.json
+
+# Run the tests from the built dev directory.
+mocha \
+  --bail \
```
samtstern, line 13: Generally we don't use `--bail` (although I don't mind), did you intend to do this?

scripts/emulator-tests/run.sh
```diff
@@ -0,0 +1,19 @@
...
+  --require ts-node/register \
+  --require src/test/helpers/mocha-bootstrap.ts \
+  dev/scripts/emulator-tests/*.spec.*
+
+# Remove the built artifacts.
+rm -rf dev
```
samtstern, line 19: Will this run if mocha fails?

samtstern (approved):

scripts/emulator-tests/run.sh
bkendall: Removed.

scripts/emulator-tests/run.sh
bkendall: I replaced it with a trap so it will always be run on exit :)

PR #3037, opened 2021-01-14
nock out API call to extensions registry

bkendall:
### Description

In the tests, the logic eventually would call to the extensions registry which would slow them down. I've nocked out those calls so that they are not made any more, bringing the tests down from 500ms to next to nothing.


joehan (approved):
Good catch!

huangjeff5 (approved):

PR #3046, opened 2021-01-16
Adding onPoll option to operation-poller

joehan:
### Description
Adds `onPoll` to operation-poller. When this option is set, each time an operation is polled, the onPoll function will be called with the result of the operation.

This code isn't used yet, but will be part of a large refactor of the Cloud Functions deploy codepath.

### DIFF
--- a/src/operation-poller.ts
+++ b/src/operation-poller.ts
@@ -9,6 +9,7 @@ export interface OperationPollerOptions {
   operationResourceName: string;
   backoff?: number;
   masterTimeout?: number;
+  onPoll?: (operation: OperationResult<any>) => any;
 }
 
 const DEFAULT_INITIAL_BACKOFF_DELAY_MILLIS = 250;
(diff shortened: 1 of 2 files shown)


samtstern (approved):

bkendall (approved):

src/operation-poller.ts
```diff
@@ -9,6 +9,7 @@ export interface OperationPollerOptions {
   operationResourceName: string;
   backoff?: number;
   masterTimeout?: number;
+  onPoll?: (operation: OperationResult<any>) => any;
```
bkendall, line 12: I think the "really right" way to do this in JS would be to implement/extend an event emitter, but it's functionally just this anyway. :) LGTM

mbleigh: My 2c: EventEmitter is generally for when you might need to attach many/arbitrary listeners. If you don't have a need to support multiple listeners, you don't really need an EventEmitter.

PR #3051, opened 2021-01-17
Fix functions emulator url on console

KYUUUW:
### Description

When I emulated organized function, console shows wrong local emulated functions url.

Ex) code
``` javascript
// index.js

import * as admin from 'firebase-admin';
import * as functions from 'firebase-functions';

const aFunction = functions.https.onRequest((req, res) => { ... some codes... });

export const group  = {
  function1: aFunction
}
```

When emulator shows emulated functions local url in console
```
✔  functions[group-function1]: http function initialized (http://localhost:5001/{projectId}/{region}/group.function1).
```

But it should be
```
✔  functions[group-function1]: http function initialized (http://localhost:5001/{projectId}/{region}/group-function1).
```

The functions should be grouped by symbol `-` not `.` !!

### Scenarios Tested

Set my firebase cli to this developing code and tested on my project. It worked very well.

### Sample Commands

```
```

### DIFF
--- a/src/emulator/functionsEmulator.ts
+++ b/src/emulator/functionsEmulator.ts
@@ -121,7 +121,7 @@ export class FunctionsEmulator implements EmulatorInstance {
     name: string,
     region: string
   ): string {
-    return `http://${host}:${port}/${projectId}/${region}/${name}`;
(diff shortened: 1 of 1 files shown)


KYUUUW:
@googlebot I fixed it.

KYUUUW:
It looks updated already

PR #3061, opened 2021-01-20
Typescriptifying deploymentTool

joehan:
### Description

Refactoring deploymentTool to use Typescript.

This is part of a large refactor of the Cloud Functions deploy codepath. This PR should not have any functional changes - in order to keep things understandable, I'm doing a series of js to ts refactors on this flow before I start sending out PRs for functional changes.


src/deploymentTool.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export function labels() {
+  return {
+    "deployment-tool": value(),
+  };
+}
+
+export function check(labels: { [key: string]: any }) {
+  return labels && labels["deployment-tool"] && labels["deployment-tool"].indexOf(BASE) === 0;
```
huangjeff5, line 18: nit: This might be slightly more readable as `labels["deployment-tool"].startsWith(BASE)` over `labels["deployment-tool"].indexOf(BASE) === 0`

huangjeff5 (approved):

src/deploy/hosting/prepare.js
```diff
@@ -71,7 +71,7 @@ module.exports = function (context, options) {
 
     const data = {
       config: convertConfig(cfg),
-      labels: deploymentTool.labels,
+      labels: deploymentTool.labels(),
```
huangjeff5, line 74: Why was this turned into a function?

joehan: Previously, deploymentTool was setting module.exports to an object, which allowed us to use a getter function for labels & value. I wanted to move it over to use `export function` syntax to be more consistent with the rest of the codebase, but in order to do so, I had to turn this from a getter function into a normal one

src/deploymentTool.ts
mbleigh: Could also use optional chaining:

```suggestion
  return labels?.["deployment-tool"]?.startsWith(BASE);
```

PR #3065, opened 2021-01-20
pass through compressed content in hosting proxy

### LINKED ISSUE #3052: Firebase SDK proxy error: net::ERR_CONTENT_DECODING_FAILED 200 (OK)

**firebase-tools:**
9.2.1

**Platform:**
macOS
### [REQUIRED] Test case

1. Unzip, and run `firebase emulators:start --only hosting --project your-project-id` from within the directory
2. Load `http://localhost:5000/`
3. `GET http://localhost:5000/__/firebase/8.2.3/firebase-app.js net::ERR_CONTENT_DECODING_FAILED 200 (OK)`

[firebase-issue-3041.zip](https://github.com/firebase/firebase-tools/files/5817143/firebase-issue-3041.zip) - same repro files as #3041 

1. Make index.html which contains:
```
<script src="/__/firebase/8.2.3/firebase-app.js"></script>
<script src="/__/firebase/8.2.3/firebase-analytics.js"></script>
<script src="/__/firebase/init.js"></script>
```

2. Start the hosting emulator:
```
firebase emulators:start --only hosting
```

3. Load `http://localhost:5000/` and inspect javascript errors

`firebase-app.js` loads correctly

> GET http://localhost:5000/__/firebase/8.2.3/firebase-app.js net::ERR_CONTENT_DECODING_FAILED 200 (OK)

This does not seem to happen in production and only affects the emulator's response of the firebase libraries.

### LINKED ISSUE #3055: NextJS emulators:start --only functions,hosting bug
Hi There,
The following is my branch and I'm trying to create a simple firebase cloud "hosting, functions" for my NextJS application.
_Repo_
**https://github.com/jasdevsidhu12/test-next**
_Branch_ 
**firebase-issue**

For the whole of last week I kept trying and trying but I couldn't figure it out why `firebase emulators:start --only functions,hosting` **wouldn't work** but  `firebase deploy --only functions,hosting` **worked.**

**Result of  `firebase emulators:start --only functions,hosting`**
image
firebase-debug.log(**No errors shown**)
image

**Result of `firebase deploy --only functions,hosting`**
image

To replicate bug

- clone the Repo

- `npm run serve`

bkendall:
### Description

When using the Hosting emulator, we insert middleware for two cases:
1. fetching javascript SDK files from gstatic and providing initialization helpers, and
2. proxying to CF3, either locally emulated functions or live functions.

In both of these cases, we send along most of the original request through `node-fetch` and receive a response. Now, `node-fetch` (in apiv2) "helpfully" will automatically decompress the response if it is a recognized compression (via `content-encoding` headers) and return to our proxy logic the decompressed content. When we take that response and pass it (along with the original headers) along to the originating response, we sometimes _lose_ content because the compressed response is not re-compressed automatically and the `content-length` header can cause the response to look truncated.

This PR adds support for the `compress` flag, which originates from `node-fetch`. The `compress` flag indicates to `apiv2` (and `node-fetch`) that it (a) should not add an `accept-encoding` header to the response if one does not exist and (b) should not decompress the response if it recognizes the encoding. With that in place, our proxies correctly return compressed content as appropriate and the requests don't throw errors in browsers any longer.

Fixes #3052
Fixes #3055

### Scenarios Tested

I was able to replicate this behavior fairly reliably for (1) by initializing a new Hosting directory and loading it up in the emulator. Looking at the network tab of the inspector, I see no more "failed" responses.

#3055 provided a repro case that I was able to use for (2). Their `/test` endpoint would fail to load, but with this fix, it is successfully loading again.

I also added a test that verifies that the middleware powering (1) doesn't decompress the request when it's being proxied. This was slightly tricky to write because I have to use more basic Node tools, but I believe the test to be valid (and it breaks if `compress` is removed).

I guess this goes into the pile of "things `request` used to do automatically". This was a bugger to track down, but I feel more confident in it now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Improves handling of compressed content when proxying through the Hosting emulator. (#3052, #3055)
--- a/src/apiv2.ts
+++ b/src/apiv2.ts
@@ -19,6 +19,7 @@ interface BaseRequestOptions<T> extends VerbOptions<T> {
   body?: T | string | NodeJS.ReadableStream;
   responseType?: "json" | "stream";
   redirect?: "error" | "follow" | "manual";
+  compress?: boolean;
 }
 
 interface RequestOptionsWithSignal<T> extends BaseRequestOptions<T> {
@@ -291,6 +292,7 @@ export class Client {
       headers: options.headers,
       method: options.method,
       redirect: options.redirect,
+      compress: options.compress,
     };
 
     if (this.opts.proxy) {
--- a/src/hosting/initMiddleware.ts
+++ b/src/hosting/initMiddleware.ts
@@ -29,6 +29,7 @@ export function initMiddleware(init: TemplateServerResponse): RequestHandler {
         path: u.pathname,
         responseType: "stream",
         resolveOnHTTPError: true,
+        compress: false,
       })
         .then((sdkRes) => {
           if (sdkRes.status === 404) {
--- a/src/hosting/initMiddleware.ts
+++ b/src/hosting/initMiddleware.ts
@@ -24,9 +24,15 @@ export function initMiddleware(init: TemplateServerResponse): RequestHandler {
       const sdkName = match[2];
       const u = new url.URL(`https://www.gstatic.com/firebasejs/${version}/${sdkName}`);
       const c = new Client({ urlPrefix: u.origin, auth: false });
+      const headers: { [key: string]: string } = {};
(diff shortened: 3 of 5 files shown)


yuchenshi (approved):

samtstern (approved):

clintecker:
Just want to say thank you for this, I was running down this bug myself in hosting rewrites to callable functions that began affecting us a couple days ago. :+1:

Issue #3022, opened 2021-01-12
Emulators:exec Auth doesn't import data


wtpalexander:
### [REQUIRED] Environment info

**firebase-tools:** 9.1.0

**Platform:** macOS

### [REQUIRED] Test case

```
firebase emulators:start
# Create some test accounts

# On separate terminal
firebase emulators:export "./emulator-data/

# Running in CI
firebase emulators:exec "./XCodeTest.sh" --import=./emulator-data/
```

### [REQUIRED] Steps to reproduce

Attempt to use the accounts listed in the `auth_export/accounts.json` when running in CI but notice that these accounts don't exist and there are no logs stated the auth emulator imported its data, unlike when the emulators are started locally with `firebase emulators:start --import=./emulator-data/`.

### [REQUIRED] Expected behavior

Expect to have the accounts within `auth_export/accounts.json` accessible when CI runs using `emulators:exec` rather than `emulators:start`.

### [REQUIRED] Actual behavior

```
[2021-01-12T17:03:54.893Z] > command requires scopes: ["email","openid","https://www.googleapis.com/auth/cloudplatformprojects.readonly","https://www.googleapis.com/auth/firebase","https://www.googleapis.com/auth/cloud-platform"]
[2021-01-12T17:03:54.914Z] FirebaseError: Failed to authenticate, have you run firebase login?
⚠  emulators: You are not currently authenticated so some features may not work correctly. Please run firebase login to authenticate the CLI. 
i  emulators: Starting emulators: auth, functions, firestore {"metadata":{"emulator":{"name":"hub"},"message":"Starting emulators: auth, functions, firestore"}}
[2021-01-12T17:03:54.929Z] [hub] writing locator at /var/folders/6q/wgy6jtp12w5gzgm9lzcglpqw0000gn/T/hub-progression-2fa20.json
⚠  functions: The following emulators are not running, calls to these services from the Functions emulator will affect production: database, hosting, pubsub {"metadata":{"emulator":{"name":"functions"},"message":"The following emulators are not running, calls to these services from the Functions emulator will affect production: \u001b[1mdatabase, hosting, pubsub\u001b[22m"}}
✔  functions: Using node@12 from host. {"metadata":{"emulator":{"name":"functions"},"message":"Using node@12 from host."}}
[2021-01-12T17:03:54.954Z] defaultcredentials: no credential available.
[...]

PR #3066, opened 2021-01-20
Add logging to auth emulator when import directory is invalid

ssbushi:
### Description

Addresses Issue #3022 by adding more logs when auth import files cannot be loaded.

### Scenarios Tested
- `firebase emulators:exec` with `import` flag and valid auth import directory
- `firebase emulators:exec` with `import` flag and invalid auth import directory
- `firebase emulators:exec` without `import` flag

### Sample Commands
`firebase emulators:exec "./test.sh" --import=./emulator-data/ --only auth --project=my-test-project`

### DIFF
--- a/src/emulator/auth/index.ts
+++ b/src/emulator/auth/index.ts
@@ -77,7 +77,7 @@ export class AuthEmulator implements EmulatorInstance {
       logger.logLabeled(
         "WARN",
         "auth",
-        `Error importing config from ${configPath}. Skipping config import`
+        `Skipped importing config because ${configPath} does not exist.`
       );
     }
 
@@ -105,7 +105,7 @@ export class AuthEmulator implements EmulatorInstance {
       logger.logLabeled(
         "WARN",
         "auth",
-        `Error importing accounts from ${accountsPath}. Skipping accounts import`
+        `Skipped importing accounts because ${accountsPath} does not exist.`
       );
     }
   }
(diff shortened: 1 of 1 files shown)


src/emulator/auth/index.ts
```diff
@@ -73,6 +73,12 @@ export class AuthEmulator implements EmulatorInstance {
         },
         configPath
       );
+    } else {
+      logger.logLabeled(
+        "WARN",
+        "auth",
+        `Error importing config from ${configPath}. Skipping config import`
+      );
```
yuchenshi, line 81: nit: wording: `Skipped importing config because ${configPath} does not exist.`

src/emulator/auth/index.ts
```diff
@@ -95,6 +101,12 @@ export class AuthEmulator implements EmulatorInstance {
         // Ignore the error when there are no users. No action needed.
         { ignoreErrors: ["MISSING_USER_ACCOUNT"] }
       );
+    } else {
+      logger.logLabeled(
+        "WARN",
+        "auth",
+        `Error importing accounts from ${accountsPath}. Skipping accounts import`
```
yuchenshi, line 108: Ditto

yuchenshi (approved):

src/emulator/auth/index.ts
ssbushi: Done.

src/emulator/auth/index.ts
ssbushi: Done.

PR #3067, opened 2021-01-21
Refactoring prepare stage of functions deploy

joehan:
### Description
Refactoring the prepare step of functions deployments to use Typescript. Additionally, moving the parsing of function triggers and the check for failure policies to this step, per go/cf3-deployment refactor . 

### Scenarios Tested
Ran a series of function deploys to test the following situations:

- Create 1 function
- Create multiple function
- Update function
- Delete, create, update functions all in one deploy
- Create, update functions, N to prompt when asked to delete functions
- Create scheduled function
- Update scheduler function to different schedule
- Delete scheduled function


src/deploy/functions/prepare.ts
```diff
@@ -0,0 +1,65 @@
...
+  payload.functions = {};
+  payload.functions.triggers = getFunctionsInfo(
+    options.config.get("functions.triggers"),
+    projectId
+  );
+
+  // Validate the function code that is being deployed.
+  validate.functionsDirectoryExists(options, sourceDirName);
+  // validate.functionNamesAreValid(payload.functionNames);
+  // TODO: This doesn't do anything meaningful right now because payload.functions is not defined
```
joehan, line 60: Noticed while I was working on this that payload.functions is not defined - looks like we lost this at some point in the past and never noticed because validate.functionsNamesAreValid doesn't throw an error if called with undefined. 

I'm going to do some refactoring to get this working later, but I left it out for now to keep the size of this PR reasonable & to focus on implementing more important beahvior.

src/deploy/functions/release.js
```diff
@@ -154,49 +154,9 @@ module.exports = function (context, options, payload) {
...
 
-  return proceedPrompt
-    .then((proceed) => {
-      if (!proceed) {
-        throw new FirebaseError("Deployment canceled.", { exit: 1 });
-      }
-
-      return Promise.resolve(context.existingFunctions);
-    })
+  return Promise.resolve(context.existingFunctions)
```
joehan, line 159: This is pretty ugly right now - i left it like this because I've got  full refactor of this file coming and didn't want to clean up something that was going away.

bkendall (approved):
Cursory look looks good, though maybe someone on the functions side can comment on any logic stuff (there's not much, it seems)

PR #3069, opened 2021-01-21
Add AAB support to App Distribution

jeremydurham:
### Description

Add support for AABs to App Distribution. Allow uploading .aab files and verify that apps are in the appropriate state to upload AABs.

### Scenarios Tested

1. Manually verified that APK and IPA uploads continue to work
1. Manually verified that uploading an AAB works
1. Added a test to verify that uploading an AAB requests full app details
1. Added a test to verify that uploading an IPA and APK does not request full app details


src/appdistribution/client.ts
```diff
@@ -24,6 +25,16 @@ export enum UploadStatus {
...
 }
 
+/** Enumn representing the App Bundles state for the App */
+export enum AabState {
+  AAB_STATE_UNSPECIFIED = 1,
+  ACTIVE,
+  PLAY_ACCOUNT_NOT_LINKED,
+  NO_APP_WITH_GIVEN_BUNDLE_ID_IN_PLAY_ACCOUNT,
+  APP_NOT_PUBLISHED,
+  AAB_STATE_UNAVAILABLE,
```
tagboola, line 35: We should also add `PLAY_IAS_TERMS_NOT_ACCEPTED`

src/appdistribution/client.ts
```diff
@@ -42,10 +53,12 @@ export class AppDistributionClient {
...
   constructor(private readonly appId: string) {}
 
-  async getApp(): Promise<AppDistributionApp> {
-    utils.logBullet("getting app details...");
-
-    const apiResponse = await api.request("GET", `/v1alpha/apps/${this.appId}`, {
+  async getApp(
+    distributionFileType: DistributionFileType = DistributionFileType.APK
+  ): Promise<AppDistributionApp> {
+    utils.logBullet(`getting app details (Distribution type: ${distributionFileType})...`);
```
tagboola, line 59: Do we need to specify the Distribution type here? Will the developer know what it means in the context of getting the app details?

src/appdistribution/distribution.ts
```diff
@@ -23,7 +26,8 @@ export class Distribution {
     const distributionType = path.split(".").pop();
     if (
       distributionType !== DistributionFileType.IPA &&
-      distributionType !== DistributionFileType.APK
+      distributionType !== DistributionFileType.APK &&
+      distributionType !== DistributionFileType.AAB
     ) {
       throw new FirebaseError("unsupported distribution file format, should be .ipa or .apk");
```
tagboola, line 32: Update to "unsupported distribution file format, should be .ipa, .apk or .aab"

src/appdistribution/distribution.ts
```diff
@@ -49,13 +58,18 @@ export class Distribution {
     switch (this.fileType) {
       case DistributionFileType.IPA:
         return "ios";
+      case DistributionFileType.AAB:
       case DistributionFileType.APK:
         return "android";
       default:
         throw new FirebaseError("Unsupported distribution file format, should be .ipa or .apk");
```
tagboola, line 65: Update to "unsupported distribution file format, should be .ipa, .apk or .aab"

src/commands/appdistribution-distribute.ts
```diff
@@ -98,11 +103,39 @@ module.exports = new Command("appdistribution:distribute <distribution-file>")
...
     }
 
+    if (
+      distribution.distributionFileType() === DistributionFileType.AAB &&
+      app.aabState !== AabState.ACTIVE &&
+      app.aabState !== AabState.AAB_STATE_UNAVAILABLE
+    ) {
+      switch (app.aabState) {
+        case AabState.PLAY_ACCOUNT_NOT_LINKED: {
+          throw new FirebaseError("This project is not linked to a GooglePlay account.", {
```
tagboola, line 113: I know the copy isn't final but there should be a space between Google and Play

src/commands/appdistribution-distribute.ts
```diff
@@ -98,11 +103,39 @@ module.exports = new Command("appdistribution:distribute <distribution-file>")
...
+          throw new FirebaseError('"This app is not published in the Google Play console.', {
+            exit: 1,
+          });
+        }
+        case AabState.NO_APP_WITH_GIVEN_BUNDLE_ID_IN_PLAY_ACCOUNT: {
+          throw new FirebaseError("App with matching package name does not exist in Google Play.", {
+            exit: 1,
+          });
+        }
+        default: {
```
tagboola, line 127: If we do add `PLAY_IAS_TERMS_NOT_ACCEPTED` add a case here for it.

src/commands/appdistribution-distribute.ts
```diff
@@ -98,11 +103,39 @@ module.exports = new Command("appdistribution:distribute <distribution-file>")
...
 
+    if (
+      distribution.distributionFileType() === DistributionFileType.AAB &&
+      app.aabState !== AabState.ACTIVE &&
+      app.aabState !== AabState.AAB_STATE_UNAVAILABLE
+    ) {
+      switch (app.aabState) {
+        case AabState.PLAY_ACCOUNT_NOT_LINKED: {
+          throw new FirebaseError("This project is not linked to a GooglePlay account.", {
+            exit: 1,
```
bkendall, line 114: `exit: 1` is the default and does not need to be specified

src/test/appdistro/client.spec.ts
```diff
@@ -39,17 +41,44 @@ describe("distribution", () => {
 
   describe("getApp", () => {
     it("should throw error when app does not exist", () => {
-      nock(api.appDistributionOrigin).get(`/v1alpha/apps/${appId}`).reply(404, {});
+      nock(api.appDistributionOrigin)
+        .get(`/v1alpha/apps/${appId}?appView=${appViewBasic}`)
+        .reply(404, {});
       return expect(appDistributionClient.getApp()).to.be.rejected;
```
bkendall, line 47: Something that's important with `nock` is to make sure the calls you set up were in fact called. In most cases, the HTTP call would just fail if it didn't match the `nock` setup, but it's a little better to assert that. I'd recommend this pattern throughout:

```typescript
it("should...", async () => {
  nock(things).get(200, "stuff)

  await expect(theThing).to.do.something;

  // This is the only serious line: asserts that there are no more `nock`'d requests.
  expect(nock.isDone()).to.be.true;
})
```

So, if your `nock` didn't get called, the test would fail on the last `expect` (which is good, because it wouldn't have done what you expected).

Could you apply this pattern throughout?

src/test/appdistro/client.spec.ts
```diff
@@ -39,17 +41,44 @@ describe("distribution", () => {
...
     });
 
     it("should resolve when request succeeds", () => {
-      nock(api.appDistributionOrigin).get(`/v1alpha/apps/${appId}`).reply(200, {});
+      nock(api.appDistributionOrigin)
+        .get(`/v1alpha/apps/${appId}?appView=${appViewBasic}`)
+        .reply(200, {});
       return expect(appDistributionClient.getApp()).to.be.fulfilled;
```
bkendall, line 54: What are these supposed to be fulfilled with? The returned value isn't being checked, which seems like an oversight.

src/test/appdistro/client.spec.ts
```diff
@@ -39,17 +41,44 @@ describe("distribution", () => {
...
+      nock(api.appDistributionOrigin)
+        .get(`/v1alpha/apps/${appId}?appView=${appViewBasic}`)
+        .reply(404, {});
       return expect(appDistributionClient.getApp()).to.be.rejected;
     });
 
     it("should resolve when request succeeds", () => {
-      nock(api.appDistributionOrigin).get(`/v1alpha/apps/${appId}`).reply(200, {});
+      nock(api.appDistributionOrigin)
+        .get(`/v1alpha/apps/${appId}?appView=${appViewBasic}`)
```
bkendall, line 52: Optionally, if it's easier to understand/manage, you can use `.query({})` in nock to specify the query to match: https://www.npmjs.com/package/nock#specifying-request-query-string

bkendall (changes requested):

src/appdistribution/client.ts
jeremydurham: That's a good question. I could see this as a useful debugging tool for us, but it's unclear if it would be useful for the customer

src/appdistribution/client.ts
jeremydurham: 👍

src/appdistribution/client.ts
pranavrajgopal: we can get rid of this. I added it mainly for debugging. Not sure if we want to expose this to the customer (yet)

src/commands/appdistribution-distribute.ts
pranavrajgopal: Ahh my bad. I forgot to add the space there. Nice catch!

src/commands/appdistribution-distribute.ts
pranavrajgopal: +1

src/commands/appdistribution-distribute.ts
jeremydurham: 👍  - I may do a follow up change to clean more of these up. We seem to be about 90-10 exit: 1 vs not specifying

src/test/appdistro/client.spec.ts
jeremydurham: Definitely!

src/test/appdistro/client.spec.ts
jeremydurham: good question, this was just an update of an existing test. I'll take a look at this.

src/appdistribution/client.ts
jeremydurham: Actually, thinking about this more, I think this is useful. Going to leave it.

src/appdistribution/client.ts
jeremydurham: Done!

src/appdistribution/distribution.ts
jeremydurham: Done!

src/appdistribution/distribution.ts
jeremydurham: Done!

src/commands/appdistribution-distribute.ts
jeremydurham: All good! Updated!

src/test/appdistro/client.spec.ts
jeremydurham: @bkendall I assume we only care about checking nock.isDone() when we're "nocking". PTAL again at your convenience and see if I missed anything

src/test/appdistro/client.spec.ts
jeremydurham: Good idea, I'll update

tagboola (approved):

src/test/appdistro/client.spec.ts
jeremydurham: @bkendall I may just call appDistributionClient.getApp() without an expectation. We're mocking the response, so ultimately the response will be whatever we mock. We're really testing this to ensure that we send the correct request to the backend, and less to verify the response. What do you think?

src/appdistribution/client.ts
```diff
@@ -42,10 +54,12 @@ export class AppDistributionClient {
...
   constructor(private readonly appId: string) {}
 
-  async getApp(): Promise<AppDistributionApp> {
-    utils.logBullet("getting app details...");
-
-    const apiResponse = await api.request("GET", `/v1alpha/apps/${this.appId}`, {
+  async getApp(
+    distributionFileType: DistributionFileType = DistributionFileType.APK
+  ): Promise<AppDistributionApp> {
+    utils.logBullet(`Getting app details (Distribution type: ${distributionFileType})...`);
```
bkendall, line 60: typically log lines are lowercased

```suggestion
    utils.logBullet(`getting app details (Distribution type: ${distributionFileType})...`);
```

bkendall (changes requested):
LGTM. Please add a line in the CHANGELOG as well:

```markdown
- Adds support for...
```

src/appdistribution/client.ts
jeremydurham: Fixed!

CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- Adds support for uploading AABs to App Distribution
```
bkendall, line 1: What is an AAB? It's not the first thing that shows up when I search it, so please spell it out here (optionally keeping the abbreviation in parenthesis).

Also end with a period please.

```suggestion
- Adds support for uploading All Aardvark Backs (AABs) to App Distribution.
```

bkendall (approved):
One final touch, but LGTM

CHANGELOG.md
jeremydurham: Done!

PR #3071, opened 2021-01-21
Typescriptifying release step of functions deployment

joehan:
### Description
Refactoring release step of functions deploy to use Typescript. This PR doesn't have any significant functional changes - I wanted to first change this over to Typescript so that  it would be easier to understand what is actually being changed in future PRs.

### Scenarios Tested

Ran a series of function deploys to test the following situations:
- Create 1 function
- Create multiple function
- Update function
- Delete, create, update functions all in one deploy
- Create, update functions, N to prompt when asked to delete functions
- Create scheduled function
- Update scheduler function to different schedule
- Delete scheduled function


src/deploy/functions/release.ts
```diff
@@ -0,0 +1,477 @@
...
+  logger.info(
+    "In a few minutes, you can check status at " + utils.consoleUrl(projectId, "/functions/logs")
+  );
+  logger.info(
+    "You can use the --only flag to deploy only a portion of your functions in the future."
+  );
+  deployments = []; // prevents analytics tracking of deployments
+}
+
+export async function release(context: any, options: any, payload: any): Promise<void> {
```
inlined, line 133: This is a 340LOC method that has nested conditionals since we can't early exit anywhere. It would probably be cleaner to actually use helper methods. Just a drive-by comment since this code was just translated, not written in the PR.

src/deploy/functions/release.ts
```diff
@@ -0,0 +1,477 @@
...
+    if (!options.force) {
+      proceed = await promptOnce({
+        type: "confirm",
+        name: "confirm",
+        default: false,
+        message:
+          "Would you like to proceed with deletion? Selecting no will continue the rest of the deployments.",
+      });
+    }
+    if (proceed) {
```
inlined, line 354: nit: If we tested for `!proceed` (which I'd probably call `skip` to avoid a double negative) we would have the smaller block sooner. Then we wouldn't forget what the `else` block nearly 100 lines lower is referring to.

src/deploy/functions/release.ts
```diff
@@ -0,0 +1,477 @@
...
+      if (deployments.length !== 0) {
+        utils.logBullet(clc.bold.cyan("functions: ") + "continuing with other deployments.");
+      }
+    }
+  }
+  // filter out functions that are excluded via --only and --except flags
+  const functionsInDeploy = functionsInfo.filter((trigger: helper.CloudFunctionTrigger) => {
+    return functionFilterGroups.length > 0 ? _.includes(deleteReleaseNames, trigger.name) : true;
+  });
+  await createOrUpdateSchedulesAndTopics(
```
inlined, line 433: nit: the name for "createOrUpdate" is typically "upsert"

src/deploy/functions/release.ts
```diff
@@ -0,0 +1,477 @@
...
+    context.projectId,
+    functionsInDeploy,
+    existingScheduledFunctions,
+    appEngineLocation
+  );
+  const allOps = await utils.promiseAllSettled(
+    _.map(deployments, async (op) => {
+      const res = await op.retryFunction();
```
inlined, line 441: If we only ever call this once, why is it called `retryFunction`?

inlined (approved):
Reviewed this one first to get into the groove of reviewing this codebase. I suspect this was largely just a rename as TS and changing a few syntax styles. I know you didn't write any of this whole cloth and don't expect any of my comments to actually be taken as action items.

For future reviews like this though, I like to do the fork as one commit and the edits as another. It offers the reviewer a way to scan through the changes much more easily.

src/deploy/functions/release.ts
joehan: Agree 100 percent - this is getting torn to shreds in a coming PR

PR #3076, opened 2021-01-24
Adding logic to build regional deployments

joehan:
### Description
Implements the logic to build regional deployments, per the design at go/cf3-deployment-refactor. This will be followed up by another PR to execute these deployments - I separated this chunk so it would be easier to verify that the new logic for what gets deployed matches the old logic.

### Scenarios Tested
I ran some deploys to confirm that deploys weren't broken, but I relied on the unit tests to test createRegionalDeployment, since it's not yet used in the deployment.

### DIFF
--- a/src/deploy/functions/checkIam.ts
+++ b/src/deploy/functions/checkIam.ts
@@ -3,7 +3,11 @@ import { bold } from "cli-color";
 
 import { debug } from "../../logger";
 import * as track from "../../track";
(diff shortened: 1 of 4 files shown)


joehan:
As I've been working on the follow up PR, I realized that I need to change some of this logic slightly to correctly handle existing, non-Firebase functions that are being included in the deployment. Gonna close this for now, and reopen once I get that implemented as to not waste any reviewer time.

PR #3081, opened 2021-01-26
Implementing createDeploymentPlan

joehan:
### Description
Implementing createDeploymentPlan, per the design at go/cf3-deployment-refactor. Part of my goal for this refactor is to separates out the logic for what to deploy from how to deploy it, to make the code more testable and easier to change in the future. This PR handles the first part of that, and will followed up by another PR to execute a deployment plan .

### Scenarios Tested
In addition to the cases covered by the unit tests, I locally modified the release step to call createDeploymentPlan and print out the results, and manually tested out a bunch of different permutations of function creates, updates, and deletes, with and without schedules and --only flags.


src/functionsDeployHelper.ts
```diff
@@ -161,10 +184,124 @@ export function getFunctionsInfo(parsedTriggers: CloudFunctionTrigger[], project
...
+ * Helper method to turn a RegionMap into a flat list of all functions in a deployment.
+ * @param regionMap A RegionMap for the deployment.
+ */
+export function flattenRegionMap(regionMap: RegionMap): CloudFunctionTrigger[] {
+  return _.chain(regionMap)
+    .map((value: CloudFunctionTrigger[]) => {
+      return value;
+    })
+    .flatten()
+    .value();
```
bkendall, line 206: What is your feeling about `_.chain`? Personally, I'm not a fan.

Is this equivalent?

```typescript
const triggers: CloudFunctionTrigger[] = [];
for (const [k, v] of Object.entries(regionMap)) {
  triggers.push(...v);
}
return triggers;
```

If you're really wanting to keep `_.chain` though, here's a simplification you can make:

```suggestion
  return _.chain(regionMap)
    .map((v: CloudFunctionTrigger[]) => v)
    .flatten()
    .value();
```

src/functionsDeployHelper.ts
```diff
@@ -161,10 +184,124 @@ export function getFunctionsInfo(parsedTriggers: CloudFunctionTrigger[], project
...
+  functionsToCreate: CloudFunctionTrigger[];
+  functionsToUpdate: CloudFunctionTrigger[];
+  schedulesToCreateOrUpdate: CloudFunctionTrigger[];
+}
+
+export interface DeploymentPlan {
+  regionalDeployments: RegionalDeployment[];
+  functionsToDelete: string[];
+  schedulesToDelete: string[];
+}
```
bkendall, line 222: Can all the interfaces in this file be collected at the top? It'll make them easier to find later

src/functionsDeployHelper.ts
```diff
@@ -161,10 +184,124 @@ export function getFunctionsInfo(parsedTriggers: CloudFunctionTrigger[], project
...
+      schedulesToCreateOrUpdate: [],
+    };
+    const localFunctionsInRegion = functionsInLocalSource[region];
+    for (const fn of localFunctionsInRegion) {
+      // Check if this function matches the --only filters
+      if (functionMatchesAnyGroup(fn.name, filters)) {
+        // Check if this local function has the same name as an exisiting one.
+        const matchingExistingFunction = _.find(existingFunctions, (exFn) => {
+          return exFn.name === fn.name;
+        });
```
bkendall, line 256:
```suggestion
        const matchingExistingFunction = existingFunctions.find((exFn) => {
          return exFn.name === fn.name;
        });
```

(Can you tell my feelings on lodash in general now? haha)

src/functionsDeployHelper.ts
```diff
@@ -161,10 +184,124 @@ export function getFunctionsInfo(parsedTriggers: CloudFunctionTrigger[], project
...
+    .filter((fn) => {
+      return filters.length ? functionMatchesAnyGroup(fn.name, filters) : true;
+    })
+    .value();
+  deployment.functionsToDelete = _.map(functionsToDelete, (fn) => {
+    return fn.name;
+  });
+  // Also delete any schedules for functions that we are deleting.
+  _.forEach(functionsToDelete, (fn) => {
+    if (_.get(fn, "labels.deployment-scheduled") === "true") {
```
bkendall, line 300: I _think_ you could do `fn.labels?.["deployment-scheduled"]  === "true"`

(the last thing is `["thing"]` because of the kabab-casing)

bkendall (approved):
Honestly just some nitty things, but looks fine overall. Maybe someone on Functions should definitely check it out before merging though :)

src/functionsDeployHelper.ts
joehan: I believe you're right - plus, its only written this way because I copied this over from old functions code that was written before optional chaining existed. I definitely prefer using the native option, so I'll switch this over.

src/functionsDeployHelper.ts
joehan: I'm not the biggest fan of `_.chain` either - its tough to understand what going on with it if you're not familiar with lodash, and its always been unintuitive to me that it doesn't execute any of the chained functions until you call '.value()'. 

 However, as I'm working on the follow up PR, I've realized that I can get rid of this method entirely with some minor refactoring of checkIam.ts. For now, I think I'll go with the simplified _.chain approach - in a follow up PR, this should be gone entirely.

bkendall: heh - seriously though, my goal is to get rid of lodash entirely. So much of it can be done with plain JS (or a simple helper function), so it's not unattainable...

src/functionsDeployHelper.ts
```diff
@@ -140,14 +150,27 @@ export function logFilters(
...
+ * to that region.
+ * @param projectId The project in use.
+ * @param parsedTriggers A list of all CloudFunctions in the deployment.
+ */
+export function createFunctionRegionMap(
+  projectId: string,
+  parsedTriggers: CloudFunctionTrigger[]
+): RegionMap {
+  const regionMap: RegionMap = {};
   _.forEach(parsedTriggers, (trigger) => {
```
mbleigh, line 168: Please move to either `for...of` or `.forEach` from lodash. Don't have to do it everywhere, but I'd like us to do it in the places we're touching code.

src/functionsDeployHelper.ts
mbleigh: +💯  to getting rid of lodash (especially the more esoteric bits) everywhere we're rewriting code.

src/functionsDeployHelper.ts
```diff
@@ -161,10 +184,124 @@ export function getFunctionsInfo(parsedTriggers: CloudFunctionTrigger[], project
         "functions",
         trigger.name,
       ].join("/");
-      functionsInfo.push(triggerDeepCopy);
+      if (!_.get(regionMap, region)) {
+        regionMap[region] = [];
+      }
```
mbleigh, line 189:
```suggestion
      regionMap[region] = regionMap[region] || [];
```

You might be able to do `||=` but not sure if that's built-in to the version of Typescript we're using.

src/functionsDeployHelper.ts
mbleigh: +1

src/functionsDeployHelper.ts
```diff
@@ -161,10 +184,124 @@ export function getFunctionsInfo(parsedTriggers: CloudFunctionTrigger[], project
...
+/**
+ * Create a plan for deploying all functions in one region.
+ * @param region The region of this deployment
+ * @param functionsInLocalSource The functions present in the code currently being deployed.
+ * @param existingFunctionNames The names of all functions that already exist.
+ * @param existingScheduledFunctionNames The names of all schedules functions that already exist.
+ * @param filters The filters, passed in by the user via  `--only functions:`
+ */
+export function createDeploymentPlan(
+  functionsInLocalSource: RegionMap,
```
mbleigh, line 233: Rename `functionsInSourceByRegion` or something? Again the naming is confusing me a bit.

src/functionsDeployHelper.ts
```diff
@@ -161,10 +184,124 @@ export function getFunctionsInfo(parsedTriggers: CloudFunctionTrigger[], project
...
+  existingFunctions: CloudFunctionTrigger[],
+  filters: string[][]
+): DeploymentPlan {
+  const deployment: DeploymentPlan = {
+    regionalDeployments: [],
+    functionsToDelete: [],
+    schedulesToDelete: [],
+  };
+  for (const region of Object.keys(functionsInLocalSource)) {
```
mbleigh, line 242:
```suggestion
  for (const region in functionsInLocalSource) {
```

`for...in` is equivalent to `for...of Object.keys`

src/functionsDeployHelper.ts
```diff
@@ -161,10 +184,124 @@ export function getFunctionsInfo(parsedTriggers: CloudFunctionTrigger[], project
...
+    const regionalDeployment: RegionalDeployment = {
+      region,
+      functionsToCreate: [],
+      functionsToUpdate: [],
+      schedulesToCreateOrUpdate: [],
+    };
+    const localFunctionsInRegion = functionsInLocalSource[region];
+    for (const fn of localFunctionsInRegion) {
+      // Check if this function matches the --only filters
+      if (functionMatchesAnyGroup(fn.name, filters)) {
```
mbleigh, line 252: It's kinda weird to me that this filter is happening at this point. Why are we not filtering down the set of functions we're actually worrying about way at the beginning? Is it because delete detection still hasn't happened? Should we do delete detection earlier?

src/functionsDeployHelper.ts
```diff
@@ -161,10 +184,124 @@ export function getFunctionsInfo(parsedTriggers: CloudFunctionTrigger[], project
...
+        // Check if this local function has the same name as an exisiting one.
+        const matchingExistingFunction = _.find(existingFunctions, (exFn) => {
+          return exFn.name === fn.name;
+        });
+        // Check if the matching exisitng function is scheduled
+        const isMatchingExisitingFnScheduled =
+          _.get(matchingExistingFunction, "labels.deployment-scheduled") === "true";
```
mbleigh, line 259: Use optional chaining instead.

```suggestion
          matchingExistingFunction?.labels?.["deployment-scheduled"] === "true";
```

src/functionsDeployHelper.ts
```diff
@@ -161,10 +184,124 @@ export function getFunctionsInfo(parsedTriggers: CloudFunctionTrigger[], project
...
+        // Check if the local function is a scheduled function
+        const isScheduled = _.has(fn, "schedule");
```
mbleigh, line 261: Why not move this down below where the rest of the schedule handling logic is?

src/functionsDeployHelper.ts
```diff
@@ -161,10 +184,124 @@ export function getFunctionsInfo(parsedTriggers: CloudFunctionTrigger[], project
...
+        }
+      }
+    }
+    deployment.regionalDeployments.push(regionalDeployment);
+  }
+
+  // Delete any remaining existing functions that:
+  // 1 - Have the deployment-tool: 'firebase-cli' label and
+  // 2 - Match the --only filters, if any are provided.
+  const functionsToDelete = _.chain(existingFunctions)
```
mbleigh, line 287: 👴   *bernie voice* I am once again asking you to factor out lodash.

src/functionsDeployHelper.ts
```diff
@@ -161,10 +184,124 @@ export function getFunctionsInfo(parsedTriggers: CloudFunctionTrigger[], project
...
+        const matchingExistingFunction = _.find(existingFunctions, (exFn) => {
+          return exFn.name === fn.name;
+        });
+        // Check if the matching exisitng function is scheduled
+        const isMatchingExisitingFnScheduled =
+          _.get(matchingExistingFunction, "labels.deployment-scheduled") === "true";
+        // Check if the local function is a scheduled function
+        const isScheduled = _.has(fn, "schedule");
+
+        if (!matchingExistingFunction) {
```
mbleigh, line 263: In general can you try to move the `const` declarations as close to their first use as possible? I think it'd read easier.

src/test/functionsDeployHelper.spec.ts
```diff
@@ -284,4 +295,453 @@ describe("functionsDeployHelper", () => {
       expect(promptStub).not.to.have.been.called;
     });
   });
+
+  describe("createDeploymentPlan", () => {
+    it("should put new functions into functionsToCreate", () => {
+      const regionMap: helper.RegionMap = {
+        "us-east1": [
+          {
```
mbleigh, line 303: Might be easier to read these tests if you had a helper to generate the basic schema here.

src/test/functionsDeployHelper.spec.ts
```diff
@@ -284,4 +295,453 @@ describe("functionsDeployHelper", () => {
...
+            schedulesToCreateOrUpdate: [],
+          },
+        ],
+        functionsToDelete: [],
+        schedulesToDelete: [],
+      };
+      expect(deploymentPlan).to.deep.equal(expected);
+    });
+
+    it("should delete existing functions not in local code, only if they were deployed via CLI", () => {
```
mbleigh, line 461: Just checking that you didn't do anything to alter the logic that prompts before doing any deletes.

mbleigh (commented):
I *think* this all mostly makes sense. I like that we are parsing the triggers into a concrete set of operations in the prepare phase, and I didn't see anything obviously wrong with any of the logic. Some high-level comments:

1. Please remove lodash wherever possible. We should aim to never use lodash in code we're writing today.
2. Don't feel beholden to the existing file structure. You're cramming a lot of stuff into `functionsDeployHelper.ts` which is a pretty poorly named file. Feel free to put things in `deploy/functions/...` in a more modular fashion if you think it will help clarity.

src/deploy/functions/prepare.ts
```diff
@@ -49,17 +56,24 @@ export async function prepare(context: any, options: any, payload: any): Promise
...
   // Validate the function code that is being deployed.
   validate.functionsDirectoryExists(options, sourceDirName);
   // validate.functionNamesAreValid(payload.functionNames);
   // TODO: This doesn't do anything meaningful right now because payload.functions is not defined
   validate.packageJsonIsValid(sourceDirName, sourceDir, projectDir, !!runtimeFromConfig);
 
-  // Display a warning and prompt if any functions have failurePolicies.
-  await promptForFailurePolicies(options, payload.functions.triggers);
+  // Check what --only filters have been passed in.
+  context.filters = getFilterGroups(options);
```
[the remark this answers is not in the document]
joehan: Honestly, I don't. From reading through the existing code, I kind of figured out that payload = 'stuff that is being deployed' and context = 'things we want to remember for later in the process', but a lot of the things in the code today blur the line between those. As I go thru and address comments on this, I'll probably follow your advice of using context whenever possible.

src/functionsDeployHelper.ts
```diff
@@ -47,6 +48,15 @@ export interface CloudFunctionTrigger {
   regions?: string[];
 }
 
+export function functionMatchesAnyGroup(fnName: string, filterGroups: string[][]) {
+  if (!filterGroups.length) {
+    return true;
+  }
+  return _.some(filterGroups, (groupChunks) => {
```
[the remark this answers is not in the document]
joehan: Almost certainly can get rid of lodash here, this was just some old code that I moved from its original home in release.ts and didn't think to rewrite.

src/test/functionsDeployHelper.spec.ts
joehan: Nope (or at least I didn't intend to). This may have just been a confusing title for the test - basically, this is checking that when there are no --only filters, we delete all existing functions that a - have the deployment-tool:cli-firebase label and b- don't have a matching local function.

src/functionsDeployHelper.ts
joehan: Will do - I checked, and ||=, &&=, and ??= were added in 4.0 https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html, but we're still on 3.9.5, so no luck there.

src/functionsDeployHelper.ts
joehan: Didn't realize there was a native version of this - much nicer.

src/functionsDeployHelper.ts
joehan: Originally used for.. of here because of the following eslint error: ``` 112:3  error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in``` - we maybe want to get rid of that rule?

Ended up switching to for... in and adding a comment to disable the rule for this line.

src/deploy/functions/prompts.ts
```diff
@@ -0,0 +1,53 @@
+import * as clc from "cli-color";
+
+import { getFunctionLabel } from "../../functionsDeployHelper";
+import { CloudFunctionTrigger } from "./deploymentPlanner";
+import { FirebaseError } from "../../error";
+import { promptOnce } from "../../prompt";
+import * as utils from "../../utils";
+
```
joehan, line 8: There another prompt for deleting functions that is currently in release.ts - I'll move that here in next PR which refactors release.ts to execute these deploymentPlans.

src/deploy/functions/deploymentPlanner.ts
```diff
@@ -0,0 +1,175 @@
...
+  schedulesToDelete: string[];
+}
+
+/**
+ * Creates a map of regions to all the CloudFunctions being deployed
+ * to that region.
+ * @param projectId The project in use.
+ * @param parsedTriggers A list of all CloudFunctions in the deployment.
+ */
+export function createFunctionsByRegionMap(
```
mbleigh, line 51: nit: `createFunctionsByRegionMap` sounds like you're creating functions and passing a region map...how about `groupTriggersByRegion`?

(sorry for the naming nits I think there's a lot of subtlety to these)

src/deploy/functions/deploymentPlanner.ts
```diff
@@ -0,0 +1,175 @@
...
+  functionsInSourceByRegion: RegionMap,
+  existingFunctions: CloudFunctionTrigger[],
+  filters: string[][]
+): DeploymentPlan {
+  const deployment: DeploymentPlan = {
+    regionalDeployments: [],
+    functionsToDelete: [],
+    schedulesToDelete: [],
+  };
+  // eslint-disable-next-line guard-for-in
```
mbleigh, line 112: I'm curious how @bkendall feels about dropping this ESLint rule. I can't think of a time when I use `for...in` against something that isn't a plain object, so I don't feel like I'd ever hit the scenario it's guarding against.

src/deploy/functions/deploymentPlanner.ts
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * Create a plan for deploying all functions in one region.
+ * @param region The region of this deployment
+ * @param functionsInSourceByRegion The functions present in the code currently being deployed.
+ * @param existingFunctionNames The names of all functions that already exist.
+ * @param existingScheduledFunctionNames The names of all schedules functions that already exist.
+ * @param filters The filters, passed in by the user via  `--only functions:`
+ */
+export function createDeploymentPlan(
+  functionsInSourceByRegion: RegionMap,
```
mbleigh, line 103: Is there a reason to have this take a `RegionMap` instead of having the region grouped functions be an implementation detail of creating a deployment plan? Should the signature be instead

```ts
export function createDeploymentPlan(
  projectId: string,
  parsedTriggers: CloudFunctionTrigger[],
  existingFunctions: CloudFunctionTrigger[],
  filters: string[][]
): DeploymentPlan {
```

mbleigh (approved):
I think this is looking good. Since this is a big risky change we'll need to make sure we do lots of manual testing, but I'm good to go.

src/deploy/functions/deploymentPlanner.ts
joehan: I like that - and don't worry about the naming nits, its actually much appreciated. I've always found that this code path is difficult to choose good names for, partly because of semantic satiation of the word function, and partly because of the historical lack of clear typing making everything cloudy.

src/deploy/functions/prompts.ts
```diff
@@ -0,0 +1,53 @@
...
+export async function promptForFailurePolicies(
+  options: any,
+  functions: CloudFunctionTrigger[]
+): Promise<void> {
+  // Collect all the functions that have a retry policy
+  const failurePolicyFunctions = functions.filter((fn: CloudFunctionTrigger) => {
+    return !!fn.failurePolicy;
+  });
+
+  if (failurePolicyFunctions.length) {
```
inlined, line 24:
```
if (failurePolicyFunctions.length == 0) {
  return;
}
```

src/deploy/functions/prompts.ts
```diff
@@ -0,0 +1,53 @@
...
+      "Retried executions are billed as any other execution, and functions are retried repeatedly until they either successfully execute or the maximum retry period has elapsed, which can be up to 7 days. " +
+      "For safety, you might want to ensure that your functions are idempotent; see https://firebase.google.com/docs/functions/retries to learn more.";
+
+    utils.logLabeledWarning("functions", retryMessage);
+
+    if (options.nonInteractive && !options.force) {
+      throw new FirebaseError("Pass the --force option to deploy functions with a failure policy", {
+        exit: 1,
+      });
+    } else if (!options.nonInteractive) {
```
inlined, line 41: ooof.. I hate else if (! something)

I wonder if it's clearer to have

```
if (options.force) {
 return;
} 
if (options.nonInteractive) { // I hate that this is a name that includes a built-in negative
  throw
}
prompt
```

As a side-rant, I really really hate "nonInteractive" because you end up with code that reads "not non interactive" in your head. I might add "options.interactive" just to fix this... 😅

inlined (approved):
Lots of style nits. There's some things I don't really like about the codebase like complex nesting, non-descriptive variable/parameter names (e.g. `params` or `payload`) and double negatives. I'm going to try to nudge us away from that if possible.

src/deploy/functions/deploymentPlanner.ts
```diff
@@ -0,0 +1,175 @@
...
+  serviceAccountEmail?: string;
+  httpsTrigger?: any;
+  eventTrigger?: any;
+  failurePolicy?: {};
+  schedule?: object;
+  timeZone?: string;
+  regions?: string[];
+}
+
+export interface RegionMap {
```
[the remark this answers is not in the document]
joehan: Seems reasonable to change up this name - Michael also thought it was unclear. If we end up using this type in the final code, I'll switch it to FunctionTiggersByRegion

src/deploy/functions/deploymentPlanner.ts
```diff
@@ -0,0 +1,175 @@
...
+  [region: string]: CloudFunctionTrigger[];
+}
+
+export interface RegionalDeployment {
+  region: string;
+  sourceToken?: string;
+  firstFunctionDeployment?: CloudFunctionTrigger;
```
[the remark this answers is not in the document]
joehan: Will explain more over chat on Monday - this is here to take advantage of the sourceToken field in operation metadata, which will let us reuse builds across multiple fns & speed up deploys of many functions.

src/deploy/functions/deploymentPlanner.ts
```diff
@@ -0,0 +1,175 @@
...
+  functionsToCreate: CloudFunctionTrigger[];
+  functionsToUpdate: CloudFunctionTrigger[];
+  schedulesToCreateOrUpdate: CloudFunctionTrigger[];
+}
+
+export interface DeploymentPlan {
+  regionalDeployments: RegionalDeployment[];
+  functionsToDelete: string[];
```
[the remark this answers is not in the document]
joehan: Scheduler jobs are technically regional - however, they can only live in the default resource location (aka the region where app engine exists on that project). Because of this, they are effectively global.

FunctionsToDelete is also regional - however, we don't need to wait for the first function in that region to be deployed to  fire off the delete calls, so I figured it would be simpler to put them all in one array.

src/deploy/functions/deploymentPlanner.ts
```diff
@@ -0,0 +1,175 @@
...
+      region,
+      functionsToCreate: [],
+      functionsToUpdate: [],
+      schedulesToCreateOrUpdate: [],
+    };
+    const localFunctionsInRegion = functionsInSourceByRegion[region];
+    for (const fn of localFunctionsInRegion) {
+      // Check if this function matches the --only filters
+      if (functionMatchesAnyGroup(fn.name, filters)) {
+        // Check if this local function has the same name as an exisiting one.
```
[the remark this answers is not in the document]
joehan: They are - I vote for local functions, since 'parsedTriggers' clashes with the idea that these are Cloud Functions, not just CloudFunctionTriggers, and IMO 'inSource' is more ambiguous than 'local'.

src/deploy/functions/prompts.ts
joehan: Hmm, that does slightly change the behavior (in the suggested code, --force applies even if we're not in nonInteractive mode), but i think thats actually more correct, so I'm gonna make this change.

huangjeff5 (approved):

src/deploy/functions/deploymentPlanner.ts
```diff
@@ -0,0 +1,175 @@
+import * as deploymentTool from "../../deploymentTool";
+import { functionMatchesAnyGroup, getTopicName } from "../../functionsDeployHelper";
+
+export interface CloudFunctionTrigger {
```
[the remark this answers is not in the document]
joehan: Makes sense to me to not call this a trigger - I originally had this as CloudFunction and then I noticed the name clash with the type exported by firebase-functions. Leaving myself a note to come back at the end of implementation and choose a better name for this - maybe CloudFunctionInfo?

joehan:
Wanted to get this merged so that the next PR would be a bit simpler to look at, so I added some TODOs for comments that I thought were valuable but I didn't want to address at this moment.

PR #3091, opened 2021-01-28
Implement batchDelete in Auth Emulator.

yuchenshi:
### Description

This implements batchDelete endpoint, used in Admin SDKs and so on.

### Scenarios Tested

See tests added.

### DIFF
--- a/src/test/emulators/auth/batch.spec.ts
+++ b/src/test/emulators/auth/batch.spec.ts
@@ -14,6 +14,7 @@ import {
   signInWithPassword,
   signInWithPhoneNumber,
   TEST_PHONE_NUMBER,
+  updateAccountByLocalId,
 } from "./helpers";
 import { UserInfo } from "../../../emulator/auth/state";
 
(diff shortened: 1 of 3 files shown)


src/test/emulators/auth/batch.spec.ts
```diff
@@ -502,3 +503,128 @@ describeAuthEmulator("accounts:batchCreate", ({ authApi }) => {
...
+  });
+
+  it("should error if localIds array is empty", async () => {
+    await authApi()
+      .post(`/identitytoolkit.googleapis.com/v1/projects/${PROJECT_ID}/accounts:batchDelete`)
+      .set("Authorization", "Bearer owner")
+      .send({ localIds: [], force: true })
+      .then((res) => {
+        expectStatusCode(400, res);
+        expect(res.body.error.message).to.equal("LOCAL_ID_LIST_EXCEEDS_LIMIT");
```
ssbushi, line 599: Optional: Do we want to make this more explicit? `LOCAL_ID_LIST_EMPTY`? Or is this what we expect from Prod too?

yuchenshi: Prod behavior, also error codes are parsed by SDKs so we cannot change or introduce freely

PR #3094, opened 2021-01-29
Implement session cookies in Auth Emulator.

yuchenshi:
### Description

This PR adds support for the `createSessionCookie` API.

### Scenarios Tested

See tests added

### DIFF
--- a/src/emulator/auth/operations.ts
+++ b/src/emulator/auth/operations.ts
@@ -534,15 +534,15 @@ function createSessionCookie(
   ctx: ExegesisContext
 ): Schemas["GoogleCloudIdentitytoolkitV1CreateSessionCookieResponse"] {
   assert(reqBody.idToken, "MISSING_ID_TOKEN");
-  const validSince = Number(reqBody.validDuration) || SESSION_COOKIE_MAX_VALID_DURATION;
+  const validDuration = Number(reqBody.validDuration) || SESSION_COOKIE_MAX_VALID_DURATION;
   assert(
-    validSince >= SESSION_COOKIE_MIN_VALID_DURATION &&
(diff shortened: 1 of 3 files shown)


src/emulator/auth/operations.ts
```diff
@@ -524,6 +525,42 @@ function createAuthUri(
...
+const SESSION_COOKIE_MIN_VALID_DURATION = 5 * 60; /* 5 minutes in seconds */
+export const SESSION_COOKIE_MAX_VALID_DURATION = 14 * 24 * 60 * 60; /* 14 days in seconds */
+
+function createSessionCookie(
+  state: ProjectState,
+  reqBody: Schemas["GoogleCloudIdentitytoolkitV1CreateSessionCookieRequest"],
+  ctx: ExegesisContext
+): Schemas["GoogleCloudIdentitytoolkitV1CreateSessionCookieResponse"] {
+  assert(reqBody.idToken, "MISSING_ID_TOKEN");
+  const validSince = Number(reqBody.validDuration) || SESSION_COOKIE_MAX_VALID_DURATION;
```
samtstern, line 537: nit: the name `validSince` doesn't match my understanding of what this variable is.  `validDuration` or `validFor` would make more sense.

The reason I am commenting at all is because of this line later:
 ```
 const expiresAt = issuedAt + validSince;
 ```
 
 This reads like we're adding two dates to get a date (which doesn't make sense) but we're adding a date `issuedAt` to a duration `validSince` which does makes sense.

samtstern (approved):

src/emulator/auth/operations.ts
yuchenshi: Thanks for catching that! I think I got too tired and autocompleted the wrong thing.

PR #3096, opened 2021-01-29
Add OOB flow to reset email when updateEmail is used.

### LINKED ISSUE #3070: oobCode not appearing after updateEmail is called while using emulators

**firebase-tools: 9.2.1**

**Platform: macOS**

### [REQUIRED] Test case

```js
await testFirebaseApp.auth().createUserWithEmailAndPassword('test@email.com', 'testing');
await testFirebaseApp.auth().currentUser?.updateEmail('test1@email.com');

const oobCodes = await axios.get(`http://localhost:9099/emulator/v1/projects/project-test/oobCodes`);
```

Running the code above

oobCodes contains an oobCode for test@email.com to revert the email

The response does not contain an oobCode to revert the email. But does contain oobCodes for other actions such as resetPassword and verifyEmail

ssbushi:
### Description

Fixes #3070. Tests are pending. Drafting a PR to clarify a few things.

### Scenarios Tested

Manually tested creating an account, updating email and using the oobLink to reset email.

### Sample Commands

await firebase.auth().createUserWithEmailAndPassword('test@email.com', 'testing');
await firebase.auth().currentUser?.updateEmail('test1@email.com');

HTTP GET http://localhost:9099/emulator/v1/projects/project-test/oobCodes


src/emulator/auth/handlers.ts
```diff
@@ -36,6 +36,49 @@ export function registerHandlers(
     const state = getProjectStateByApiKey(apiKey);
 
     switch (mode) {
+      case "recoverEmail": {
+        const EXPIRED_LINK_ERROR = {
+          error: `Your request to revert your email has expired or the link has already been used.`,
+          instructions:
+            "Try reverting using this link again. If you're still unsuccessful. You can edit the email in the Emulator UI.",
```
ssbushi, line 43: Not a big fan of this message. Any ideas are welcome.

src/emulator/auth/operations.ts
```diff
@@ -881,18 +886,35 @@ export function setAccountInfoImpl(
...
+        updates.emailVerified = true;
+        if (oob.email !== user.email) {
+          updates.email = oob.email;
+        }
+        break;
+      }
+      case "RECOVER_EMAIL": {
+        state.deleteOobCode(reqBody.oobCode);
+        // TODO: Ask if this is needed for RECOVER_EMAIL.
+        // signInProvider = PROVIDER_PASSWORD;
```
ssbushi, line 905: AFAIU, setting the signInProvider is needed to issue tokens in the response. But it is not clear how that's used or when it is necessary.

Thanks!

src/emulator/auth/operations.ts
```diff
@@ -1011,6 +1041,31 @@ export function setAccountInfoImpl(
   });
 }
 
+function sendOobForEmailReset(state: ProjectState, oldEmail: string, newEmail: string, url: URL) {
+  const mode: string = "recoverEmail";
+
+  // Encode the new email with the OOB code. The old email is received as a URL parameter.
```
ssbushi, line 1047: Ideally both the old state and the new state should be stored in `OobRecord`. Since we are not doing that here, I opted to expose the reset email as a query parameter. We could have set the new email as a query parameter too. While both these options are insecure (which is okay for the auth emulator), I've opted to go with `newEmail` in `OobRecord` and oldEmail` as a parameter to be consistent with other use-cases of `OobRecord`.

src/emulator/auth/operations.ts
```diff
@@ -1011,6 +1041,31 @@ export function setAccountInfoImpl(
...
+  const { oobCode, oobLink } = state.createOob(newEmail, "RECOVER_EMAIL", (oobCode) => {
```
yuchenshi, line 1048: These seem copy pasted from `sendOobCode`. Can we extract a reusable function?

src/emulator/auth/operations.ts
yuchenshi: This needs to be kept on the server side to be aligned with production (and it MUST not be an input parameter on `setAccountInfo`. I think production stores `oldEmail` in `OobRecord` (since the code sent to the `oldEmail`). Why do we need to store `newEmail`?

src/emulator/auth/operations.ts
yuchenshi: Please test this against production to see if a new token is indeed returned. If so, you can implement that by setting `signInProvider`.

yuchenshi (changes requested):

src/emulator/auth/operations.ts
ssbushi: > Why do we need to store newEmail?

Hmm. You're right. I'm using the wrong lookUp method.

src/emulator/auth/operations.ts
ssbushi: Will do. Thanks!

src/emulator/auth/operations.ts
ssbushi: Done.

src/emulator/auth/operations.ts
ssbushi: Done. Added `initialEmail` support.

src/emulator/auth/operations.ts
ssbushi: I tested against prod. Tokens are not issued.

src/emulator/auth/handlers.ts
yuchenshi:
```suggestion
            "If you're trying to test the reverting email flow, try changing the email again to generate a new link.",
```

src/emulator/auth/handlers.ts
```diff
@@ -36,6 +36,39 @@ export function registerHandlers(
...
+        };
+        const oob = state.validateOobCode(oobCode);
+        if (oob?.requestType !== "RECOVER_EMAIL") {
+          return res.status(400).json({
+            authEmulator: EXPIRED_LINK_ERROR,
```
yuchenshi, line 48: Let's give a more direct error message (wrong type of action code) here.

src/emulator/auth/operations.ts
```diff
@@ -918,18 +899,34 @@ export function setAccountInfoImpl(
...
+          updates.email = oob.email;
+        }
+        break;
+      }
+      case "RECOVER_EMAIL": {
+        state.deleteOobCode(reqBody.oobCode);
+        const maybeUser = state.getUserByInitialEmail(oob.email);
+        assert(maybeUser, "INVALID_OOB_CODE");
+        user = maybeUser;
+        // TODO: Make the check that the new email and init email are diff when sending out the OOB in first place.
```
yuchenshi, line 920:
```suggestion
```
No need for this comment. Also, considering the email may be changed again after sending out the link, we need to guard against this anyway. I think you're doing the right thing here.

src/emulator/auth/operations.ts
```diff
@@ -944,12 +941,23 @@ export function setAccountInfoImpl(
 
     if (reqBody.email) {
       assert(isValidEmailAddress(reqBody.email), "INVALID_EMAIL");
+      if (!emulatorUrl) {
+        throw new Error("Internal assertion error: Emulator URL invalid");
```
yuchenshi, line 945:
```suggestion
        throw new Error("Internal assertion error: missing emulatorUrl param");
```

src/emulator/auth/operations.ts
```diff
@@ -944,12 +941,23 @@ export function setAccountInfoImpl(
...
+      }
+
       const newEmail = canonicalizeEmailAddress(reqBody.email);
       if (newEmail !== user.email) {
         assert(!state.getUserByEmail(newEmail), "EMAIL_EXISTS");
```
yuchenshi, line 950: BLOCKING: Is it an error to set your email to someone else's `initialEmail`? Please double check production behavior and add a test case. If you can, then what happens if that person reverts their email using an oobCode? Would that violate the uniqueness constraint of `user.email`?

src/emulator/auth/operations.ts
```diff
@@ -1048,6 +1056,72 @@ export function setAccountInfoImpl(
...
+  email: string,
+  url: URL,
+  params: {
+    requestType: OobRequestType;
+    mode: string;
+    continueUrl?: string;
+  }
+): {
+  oobRecord: OobRecord;
+  maybeMessage?: string;
```
yuchenshi, line 1082: Let's split logging into another function, maybe `logOobRecord` or something that just logs an OOB as a side effect. It's much cleaner that way.

src/emulator/auth/state.ts
```diff
@@ -153,6 +155,10 @@ export class ProjectState {
       deleteProviders.push(PROVIDER_PASSWORD);
     }
 
+    if (!oldInitialEmail && user.initialEmail) {
```
yuchenshi, line 158: Let's just get rid of the `oldInitialEmail` variable. It really does't hurt to set the key in the map over and over again.

src/emulator/auth/operations.ts
```diff
@@ -944,12 +941,23 @@ export function setAccountInfoImpl(
...
+      }
+
       const newEmail = canonicalizeEmailAddress(reqBody.email);
       if (newEmail !== user.email) {
         assert(!state.getUserByEmail(newEmail), "EMAIL_EXISTS");
         updates.email = newEmail;
         // TODO: Set verified if email is verified by IDP linked to account.
         updates.emailVerified = false;
+        if (user.email && !user.initialEmail) {
+          updates.initialEmail = user.email;
```
yuchenshi, line 955: BLOCKING: This is not the only place where an email can be added to an account. `signUp` and `signInWithEmailLink` and a few other places can too. It's probably better to do it in `state.ts` in `updateUserByLocalId` so that we can capture all cases.

yuchenshi (changes requested):
General approach LGTM. See blocking comments below and tests please.

src/emulator/auth/operations.ts
ssbushi: Agreed. But the prod behaviour is to set `initialEmail` only in RECOVER_EMAIL flow.

src/emulator/auth/state.ts
ssbushi: Done.

src/emulator/auth/operations.ts
yuchenshi: Makes sense.

src/emulator/auth/operations.ts
```diff
@@ -944,12 +941,23 @@ export function setAccountInfoImpl(
...
+        }
+        const initialEmail = user.initialEmail ?? updates.initialEmail;
+        if (initialEmail && newEmail !== initialEmail) {
+          sendOobForEmailReset(state, initialEmail, emulatorUrl);
```
yuchenshi, line 959: You may want to move this side effect to the very end, after `updateUserByLocalId` to minimize the chance of out-of-order issues with developers' test code.

src/emulator/auth/handlers.ts
ssbushi: Done.

src/emulator/auth/operations.ts
ssbushi: Done.

src/emulator/auth/operations.ts
ssbushi: Done

src/emulator/auth/operations.ts
ssbushi: Good call. You can set your email to someone else's `initialEmail`, but it will fail with `EMAIL_EXISTS` if the latter user tries to revert their email. Updated the code here to do the same.

Looks like the behaviour is the same even if `allowDuplicateEmails` is set to true.

src/emulator/auth/handlers.ts
ssbushi: Thanks! Done.

src/emulator/auth/operations.ts
```diff
@@ -206,6 +208,7 @@ function lookup(
 
   if (ctx.security?.Oauth2) {
     if (reqBody.initialEmail) {
+      // TODO: This is now possible. See ProjectState.getUserByInitialEmail.
```
yuchenshi, line 211: This can be done with a follow up PR with test cases.

src/emulator/auth/operations.ts
```diff
@@ -1048,6 +1063,74 @@ export function setAccountInfoImpl(
   });
 }
 
+function sendOobForEmailReset(state: ProjectState, initialEmail: string, url: URL) {
+  const MODE: string = "recoverEmail";
+  const RECOVER_EMAIL_REQUEST_TYPE: OobRequestType = "RECOVER_EMAIL";
```
yuchenshi, line 1068: Just inline these variables below

src/emulator/auth/operations.ts
```diff
@@ -1048,6 +1063,74 @@ export function setAccountInfoImpl(
...
+      url.searchParams.set("continueUrl", params.continueUrl);
+    }
+
+    return url.toString();
+  });
+
+  return oobRecord;
+}
+
+function logOobMessage(requestType: OobRequestType, oobLink: string, email: string) {
```
yuchenshi, line 1110: This can just take an `OobRecord`.

src/emulator/auth/operations.ts
```diff
@@ -1048,6 +1063,74 @@ export function setAccountInfoImpl(
...
+  state: ProjectState,
+  email: string,
+  url: URL,
+  params: {
+    requestType: OobRequestType;
+    mode: string;
+    continueUrl?: string;
+  }
+): OobRecord {
+  // Encode the old email with the OOB code. The new email is stored in the UserInfo object.
```
yuchenshi, line 1088:
```suggestion
```

yuchenshi (approved):
LGTM with nits

src/emulator/auth/operations.ts
ssbushi: Yup. That's why I left it as it is. Open for external contributions too.

src/emulator/auth/operations.ts
ssbushi: Done.

src/emulator/auth/operations.ts
ssbushi: Done

yuchenshi (approved):

PR #3099, opened 2021-01-29
fixes support for calls against local emulators as admin

### LINKED ISSUE #3084: firestore#delete not working with the emulator after more or less recent change

**firebase-tools:9.2.2**

**Platform:macOS**

### [REQUIRED] Test case

Check the steps to reproduce. It does not require anything else. Downgrading firebase-tools to, say, 8.16.2, works with that same code.

Have emulators running, then

```
process.env.FIREBASE_AUTH_EMULATOR_HOST = 'localhost:9099'
process.env.FIRESTORE_EMULATOR_HOST = 'localhost:8080'
process.env.NODE_ENV = 'development'
process.env.FIREBASE_TOKEN = 'FAKE_TOKEN'

firebaseTools.firestore.delete(userRef.path, {
    project:   admin.instanceId().app.options.projectId,
    recursive: true,
    yes:       true
 })
```

Before this commit (https://github.com/firebase/firebase-tools/commit/9d90a94ae17f8bf48ccf66a0be1f55c2de3f2bcc), which added apiv2 to requireAuth.ts, having `process.env.FIREBASE_TOKEN = 'FAKE_TOKEN'` was enough to work with the local emulator. After this change, it's not possible.

There was a fix for api.js here (https://github.com/firebase/firebase-tools/pull/2211/files/3a7515e2339c571d1f389ecd42dfc597703ea06e#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed) which uses `var getTokenPromise = secureRequest ? api.getAccessToken() : Promise.resolve({ access_token: "owner" });`. apiv2 on the other hand has just `const token = await this.getAccessToken()`

Error is returned: "Failed to fetch page after undefined FirebaseError: HTTP Error: 400, invalid jwt" from apiv2.js

bkendall:
### Description

Fixes #3084

Local emulators support a bearer token of `owner` to fake an admin token. This is done so that real admin tokens are not accidentally forwarded somewhere they're not supposed to go. The previous API package did this shimming, and we're replicating it here.

### DIFF
--- a/src/apiv2.ts
+++ b/src/apiv2.ts
@@ -416,9 +416,7 @@ export class Client {
 
 function isLocalInsecureRequest(urlPrefix: string): boolean {
   const u = parse(urlPrefix);
-  const isLocal = u.hostname === "localhost" || u.hostname === "127.0.0.1";
-  const isInsecure = u.protocol === "http:";
-  return isInsecure && isLocal;
+  return u.protocol === "http:";
 }
 
 function bodyToString(body: unknown): string {
--- a/src/apiv2.ts
+++ b/src/apiv2.ts
@@ -1,6 +1,6 @@
 import { AbortSignal } from "abort-controller";
 import { Readable } from "stream";
-import { URLSearchParams } from "url";
+import { parse, URLSearchParams } from "url";
 import * as ProxyAgent from "proxy-agent";
 import AbortController from "abort-controller";
 import fetch, { HeadersInit, Response, RequestInit, Headers } from "node-fetch";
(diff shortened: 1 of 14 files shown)


src/apiv2.ts
```diff
@@ -408,6 +414,13 @@ export class Client {
   }
 }
 
+function isLocalInsecureRequest(urlPrefix: string): boolean {
+  const u = parse(urlPrefix);
+  const isLocal = u.hostname === "localhost" || u.hostname === "127.0.0.1";
```
yuchenshi, line 419: You can use emulators on a different host too, like `FIRESTORE_EMULATOR_HOST=192.168.1.0 firebase ... `. How shall we handle that? I'd propose dropping `isLocal` check and just don't send real tokens over any insecure connection, like Firestore Admin SDK.

yuchenshi (approved):

src/apiv2.ts
samtstern: +1, `0.0.0.0` is very common as are the examples that Yuchen mentioned.  Sending secure tokens over `http` is always a bad idea and we should drop them.

bkendall: Alright. It has been simplified.

PR #3105, opened 2021-02-03
Fixes manipulated location header in Hosting emulator proxy logic

### LINKED ISSUE #3097: Dynamic content functions redirect in emulators to :5001 port instead of :5000

**firebase-tools:** 9.1.0

**Platform:** macOS

### [REQUIRED] Test case

- Setup serving dynamic content with firebase hosting as per firebase documentation (using Express).
- Trigger a function by navigating in browser to localhost:5000/test
- Return res.redirect("/destination")  at in the express function
- User navigated to localhost:5001/destination
- User gets 404 error (expected for localhost:5001/destination as nothing is there)

See above

Expect the redirect to use the :5000 port to navigate the user to: localhost:5000/destination
NOTE - when deployed to live, this redirect works fine. Only when operating in the emulator do the ports seem to get muddled up.

Redirects to localhost:5001/destination.

I understand this is because the functions emulator is running at :5001, however production redirects to the hosting path without any issues.

Is there either a fix to be done to the emulator, or a way I can tell express to use :5000 port instead of :5001 ?

bkendall:
### Description

Fixes #3097

Node fetch attempts to make life easier, but in fact changes behavior that's defined in the spec and doesn't currently document it: https://github.com/node-fetch/node-fetch/blob/4abbfd231f4bce7dbe65e060a6323fc6917fd6d9/src/index.js#L117-L120

I've filed a bug with them: https://github.com/node-fetch/node-fetch/issues/1086

For now, the best we can do is attempt to fix the `location` header that comes back if it's a resolved URL and that URL points us to the same origin as our original proxy request. This allows valid URLs to come back from the proxy target and only strips the origin if it's a redirect to the same origin.

### Scenarios Tested

I have a manual setup where a redirect is returned. I see now that the `location` header is what I expect:

before:

```
» curl -I http://localhost:5000/start
HTTP/1.1 302 Found
x-powered-by: Express
location: http://localhost:5001/somewhereelse
```

after:

```
» curl -I http://localhost:5000/start
HTTP/1.1 302 Found
x-powered-by: Express
location: /somewhereelse
```

### DIFF
--- a/src/hosting/proxy.ts
+++ b/src/hosting/proxy.ts
@@ -38,6 +38,8 @@ function makeVary(vary: string | null = ""): string {
  * when writing out logs or errors.  This makes some minor changes to headers,
  * cookies, and caching similar to the behavior of the production version of
  * the Firebase Hosting origin.
+ * @param url
+ * @param rewriteIdentifier
  */
 export function proxyRequestHandler(url: string, rewriteIdentifier: string): RequestHandler {
   return async (req: IncomingMessage, res: ServerResponse, next: () => void): Promise<void> => {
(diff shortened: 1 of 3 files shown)


src/hosting/proxy.ts
```diff
@@ -141,6 +141,29 @@ export function proxyRequestHandler(url: string, rewriteIdentifier: string): Req
 
     proxyRes.response.headers.set("vary", makeVary(proxyRes.response.headers.get("vary")));
 
+    // Fix the location header that `node-fetch` attempts to helpfully fix:
+    // https://github.com/node-fetch/node-fetch/blob/4abbfd231f4bce7dbe65e060a6323fc6917fd6d9/src/index.js#L117-L120
```
kmcnellis, line 145: Maybe also leave a link to the issue you filed on their repo?

kmcnellis (approved):

PR #3107, opened 2021-02-04
Implements logic to execute deployment plan

joehan:
### Description
Refactors release.ts to build and execute deploymentPlans when deploying functions.

This also changes the release code to use the same operation polling and throttler code as the rest of the codebase.  In a follow up PR, I'll get rid of the last few usages of the old polling code (in the functions:delete command), and I'll clean up that code as well.

### Scenarios Tested
I've been running a ton of test deploys with this code and I'm going to continue to do so while this PR is under review. Some situations I've tested so far:
- Deploying 10+ functions across multiple regions, and verifying that sourceToken is used all deploys after the first in a region.
- Deployments that include deleting functions
- Updating scheduled pubsub functions into unscheduled ones, and vice versa
- Deleting scheduled functions.


src/deploy/functions/release.ts
```diff
@@ -2,477 +2,120 @@
...
-          sortedFailedDeployments.map((name) => `functions:${name.replace(/-/g, ".")}`).join(",")
-        ) +
-        clc.bold('"')
-    );
-    logger.info("\n\nTo continue deploying other features (such as database), run:");
-    logger.info("    " + clc.bold("firebase deploy --except functions"));
-    throw new FirebaseError("Functions did not deploy properly.");
-  }
+  // Once everything has been added to queues, starting processing.
+  // Note: We need to set up these wait before calling process and close.
```
joehan, line 103: Figuring this out drove me up a wall for a day and a half :(. It turns out that our implementation of throttler.wait() returns a promise that never resolves if you call it after .process() and .close(). I'd like to fix this at some point, but that code is used in a bunch of other places that I don't understand well enough to test effectively.

src/deploy/functions/release.ts
```diff
@@ -2,477 +2,120 @@
...
+  const shouldDeleteFunctions = await promptForFunctionDeletion(
+    fullDeployment.functionsToDelete,
+    options.force,
+    options.nonInteractive
+  );
+  if (shouldDeleteFunctions) {
+    for (const fnName of fullDeployment.functionsToDelete) {
+      cloudFunctionsQueue
+        .run(tasks.deleteFunctionTask(taskParams, fnName))
+        .then(() => {
```
joehan, line 54: Figured out a better way to do this error handling - another commit coming to refactor this slightly.

src/deploy/functions/deploymentPlanner.ts
```diff
@@ -34,7 +34,6 @@ export interface RegionMap {
 export interface RegionalDeployment {
   region: string;
   sourceToken?: string;
-  firstFunctionDeployment?: () => any;
```
joehan, line 37: Realized that I could turn this into an implementation detail of runRegionalFunctionDeployment, so I removed this from the type.

src/functionsDeployHelper.ts
```diff
@@ -6,10 +6,12 @@ import * as logger from "./logger";
 import * as track from "./track";
 import * as utils from "./utils";
 import * as cloudfunctions from "./gcp/cloudfunctions";
+import { Job } from "./gcp/cloudscheduler";
 import * as pollOperations from "./pollOperations";
 import { CloudFunctionTrigger } from "./deploy/functions/deploymentPlanner";
+import Queue from "./throttler/queue";
 
-// TODO: Get rid of this when switching to use operation-poller.
+// TODO(joehan): Get rid of this once we refactor functions-delete.js
```
joehan, line 14: This file has a mix of random helper functions, and chaff from the old deployment code that I can't quite delete yet because its used on functions-delete and other code paths. I'm gonna make another PR later to clean up the old functions, and move the still relevant functions into a less generically named file inside of the src/deploy/functions directory.

src/deploy/functions/tasks.ts
```diff
@@ -0,0 +1,279 @@
...
+      const pollerOptions: OperationPollerOptions = Object.assign(
+        {
+          pollerName: `delete-${fnName}`,
+          operationResourceName: deleteRes.name,
+        },
+        defaultPollerOptions
+      );
+      const operationResult = await pollOperation<void>(pollerOptions);
+      params.timer.endTimer(fnName);
+      helper.printSuccess(fnName, "delete");
```
taeold, line 176: If I'm reading this correctly, I think this will print fully qualified name (including project, region) vs other tasks print function name only. Is that okay?

src/deploy/functions/tasks.ts
```diff
@@ -0,0 +1,279 @@
...
+import { getHumanFriendlyRuntimeName } from "../../parseRuntimeAndValidateSDK";
+import { deleteTopic } from "../../gcp/pubsub";
+import { DeploymentTimer } from "./deploymentTimer";
+import { ErrorHandler } from "./errorHandler";
+
+// TODO: Tune this for better performance.
+const defaultPollerOptions = {
+  apiOrigin: functionsOrigin,
+  apiVersion: cloudfunctions.API_VERSION,
+  masterTimeout: 150000,
```
taeold, line 22: nit* can we write this as

```
40 * 60 * 60 // 40 hours
```

Also - did I get this right? 40 hours looks like a very generous timeout.

src/deploy/functions/tasks.ts
```diff
@@ -0,0 +1,279 @@
...
+      }
+      logger.debug(`Scheduler job ${jobName} not found.`);
+    }
+    try {
+      await deleteTopic(topicName);
+      helper.printSuccess(fnName, "delete schedule");
+    } catch (err) {
+      // If the topic has already been deleted, don't throw an error.
+      if (err.status !== 404) {
+        params.errorHandler.record("error", fnName, "delete schedule", err.message || "");
```
taeold, line 223: do we need to add `return;` here? Otherwise, the debug message below will always be printed.

src/deploy/functions/tasks.ts
```diff
@@ -0,0 +1,279 @@
...
+): () => Promise<void> {
+  return async () => {
+    const jobName = helper.getScheduleName(fnName, appEngineLocation);
+    const topicName = helper.getTopicName(fnName);
+    try {
+      await cloudscheduler.deleteJob(jobName);
+    } catch (err) {
+      // If the job has already been deleted, don't throw an error.
+      if (err.status !== 404) {
+        params.errorHandler.record("error", fnName, "delete schedule", err.message || "");
```
taeold, line 213: do we need to add `return;` here? Otherwise, the debug message below will always be printed.

src/functionsDelete.js
```diff
@@ -84,9 +85,7 @@ module.exports = function (functionsToDelete, projectId, appEngineLocation) {
           })
           .then(() => {
             return cloudfunctions.deleteFunction({
-              projectId: projectId,
-              region: helper.getRegion(name),
-              functionName,
+              functionName: name,
```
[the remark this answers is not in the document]
inlined: The Google API style guide defines `myfunc` as the "id" and `projects/my-projc/locations/us-central1/functions/myfunc` as the "name". IMO the only solution here is to be pedantically consistent.

inlined (commented):
LGTM with the logging fixes Daniel pointed out.

src/deploy/functions/errorHandler.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (this.warnings.length === 0) {
+      return;
+    }
+    const failedIamCalls = this.warnings.filter((e) => e.operationType === "make public");
+    if (failedIamCalls.length) {
+      logger.info("\nUnable to set publicly accessible IAM policy on the following functions:");
+      for (const failedDep of failedIamCalls) {
+        logger.info(`\t${failedDep.functionName}`);
+      }
+      logger.info("\nUnauthorized users will not be able access this function.");
```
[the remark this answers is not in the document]
joehan: You are correct! I'm also gonna get @egilmorez to take a look at this to shop the language further at some point.

src/deploy/functions/errorHandler.ts
```diff
@@ -0,0 +1,93 @@
...
+      logger.info(
+        "\nThis may be caused by an organization policy that restricts Network Access on your project."
```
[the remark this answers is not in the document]
joehan: Ooof, I didn't realize that was the case. I'll reword this to have a bulleted list of common reasons, and include that as the first one.

src/deploy/functions/errorHandler.ts
```diff
@@ -0,0 +1,93 @@
...
+          this.errors
+            .map((failedDep) => `functions:${failedDep.functionName.replace(/-/g, ".")}`)
+            .join(",")
+        ) +
+        clc.bold('"')
+    );
+    logger.info("\nTo continue deploying other features (such as database), run:");
+    logger.info("    " + clc.bold("firebase deploy --except functions"));
+    // Print all the original messages at debug level.
+    for (const failedDep of this.errors) {
```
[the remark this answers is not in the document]
joehan: Deployment - I'll replace this with the full word to make this clearer

src/deploy/functions/tasks.ts
```diff
@@ -0,0 +1,279 @@
...
+    utils.logBullet(
+      clc.bold.cyan("functions: ") +
+        "creating " +
+        getHumanFriendlyRuntimeName(params.runtime) +
+        " function " +
+        clc.bold(helper.getFunctionLabel(fn.name)) +
+        "..."
+    );
+    params.timer.startTimer(fn.name, "create");
+    const eventType = fn.eventTrigger ? fn.eventTrigger.eventType : "https";
```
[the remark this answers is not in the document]
joehan: Yep, it could have been written either way

src/deploy/functions/tasks.ts
```diff
@@ -0,0 +1,279 @@
...
+  sourceUrl: string;
+  sourceToken?: string;
+  timer: DeploymentTimer;
+  errorHandler: ErrorHandler;
+}
+
+export function createFunctionTask(
+  params: TaskParams,
+  fn: CloudFunctionTrigger,
+  onPoll?: (op: any) => any
```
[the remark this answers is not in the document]
joehan: Ahh yes, this was an oversight - will fix

src/deploy/functions/checkIam.ts
```diff
@@ -53,7 +53,7 @@ export async function checkServiceAccountIam(projectId: string): Promise<void> {
  */
 export async function checkHttpIam(context: any, options: any, payload: any): Promise<void> {
   const functionsInfo = payload.functions.triggers;
-  const filterGroups = getFilterGroups(options);
+  const filterGroups = context.filters;
```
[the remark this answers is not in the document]
joehan: Good catch on the import! Its not the prettiest code flow, but in prepare.ts, we have a line that already does this work: ``` context.filters = getFilterGroups()```. 

I figured it would be nice to avoid redoing this work - however, it does sacrifice some readability, and it could lead to subtle bugs if this code ever gets used elsewhere. I'll make this fall back to getFilterGroups() if context.filters is undefined.

src/deploy/functions/tasks.ts
joehan: This is in ms, so its 150 seconds. I'll add a comment clarifying this for now. I'd like to make some improvements to throttler at some point in the future since it wasn't the clearest to work with here - when I do so, I'll rename this option to masterTimeoutMs for clarity.

src/deploy/functions/tasks.ts
joehan: This is working as intended - printSuccess takes a fully qualified name. CloudFunctionTrigger.name is the fully qualified name, so the other tasks behave the same way.

src/functionsDelete.js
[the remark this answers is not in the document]
joehan: This definitely is confusing and ive introduced a few bugs during development because of it. I'll go through and switch everything to id and name to follow the style guide

src/deploy/functions/tasks.ts
```diff
@@ -52,7 +52,7 @@ export function createFunctionTask(
         projectId: params.projectId,
         region: helper.getRegion(fn.name),
         eventType: eventType,
-        functionName: helper.getFunctionName(fn.name),
+        functionName: helper.getFunctionId(fn.name),
```
taeold, line 55: Not for this PR for we should change functionName argument to functionId soon? I volunteer to do that later.

src/functionsDelete.js
```diff
@@ -54,7 +54,7 @@ module.exports = function (functionsToDelete, projectId, appEngineLocation) {
   deletes = _.map(functionsToDelete, function (name) {
     const scheduleName = helper.getScheduleName(name, appEngineLocation);
     const topicName = helper.getTopicName(name);
-    const functionName = helper.getFunctionName(name);
+    const functionName = helper.getFunctionId(name);
```
taeold, line 57: Same here - functionName vs Id is getting tricky.

taeold (approved):

src/deploy/functions/tasks.ts
joehan: Ahhh, this one slipped past me. I like the idea of doing this in a separate PR later - it should be easier to track down any places we missed it then

src/functionsDelete.js
joehan: +100 on these - i just found another bug caused by this. I was accidentally using name instead of ID for a call to setIamPolicy, and didn't notice until now because that call was failing for me anyway due to the org policy.

PR #3112, opened 2021-02-06
docs(emulator): remove extra `-` in `---project` from missing project ID warning

MichaelSolati:
### Description

If you run the emulator you get a warning that you need to set a project ID and can use the `---project` flag, but it had an extra `-` so if you use it as documented you get this...

```
error: unknown option '---project'
```

### Scenarios Tested

I just created a simple project, added firebase with `firebase init` and never assigned a project. Then I went to run the emulator, etc, etc...


samtstern:
@MichaelSolati thanks for catching this!

samtstern (approved):

PR #3118, opened 2021-02-08
Fix check annotations that show up in every PR

inlined:
### Description

Fix check errors that show up in every pull request

### Scenarios Tested

`npm run test`

### Sample Commands

n/a

### DIFF
--- a/src/apiv2.ts
+++ b/src/apiv2.ts
@@ -4,16 +4,18 @@ import { parse, URLSearchParams } from "url";
 import * as ProxyAgent from "proxy-agent";
 import AbortController from "abort-controller";
 import fetch, { HeadersInit, Response, RequestInit, Headers } from "node-fetch";
+import util from "util";
 
+import * as auth from "./auth";
 import { FirebaseError } from "./error";
 import * as logger from "./logger";
 import * as responseToError from "./responseToError";
 
-const CLI_VERSION = require("../package.json").version;
+import { version as CLI_VERSION } from "../package.json";
 
 export type HttpMethod = "GET" | "PUT" | "POST" | "DELETE" | "PATCH";
 
-interface BaseRequestOptions<T> extends VerbOptions<T> {
+interface BaseRequestOptions<T> extends VerbOptions {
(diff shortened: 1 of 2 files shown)


src/apiv2.ts
```diff
@@ -4,16 +4,17 @@ import { parse, URLSearchParams } from "url";
...
 import AbortController from "abort-controller";
 import fetch, { HeadersInit, Response, RequestInit, Headers } from "node-fetch";
+import * as auth from "./auth";
 
 import { FirebaseError } from "./error";
 import * as logger from "./logger";
 import * as responseToError from "./responseToError";
 
-const CLI_VERSION = require("../package.json").version;
+import CLI_VERSION from "../package.json";
```
samtstern, line 13: `.version` is dropped

bkendall, line 13: I find it unlikely that this variable would remain the same. Would the location using this need to change to do something like `CLI_VERSION.version`? in which case, should this have a better name?

bkendall (approved):
Thanks for doing this - there were a few spots that just didn't get cleaned up correctly as I was iterating through this file...

src/apiv2.ts
inlined: Wow. you guys are quick on the draw. I didn't intend this to be reviewed yet. The correct version is `import {version as CLI_VERSION} from '../package.json'`

inlined: OIC. The git workflow automatically added people as reviewers? I meant this to be a draft until now.

samtstern: Nope I just saw it in my inbox and was curious!  I was not added as a reviewer.  Since the CLI has no real staffing or maintainers for common bits like this you'll get a lot of drive-by gawkers.

Carry on!

bkendall: It doesn't automatically recommend folks - but I think the button to create a PR defaults to whatever you did last? Either way, we happened to see it :P Sorry we're super eager!

bkendall: Hi Sam! bahaha

PR #3131, opened 2021-02-12
Implement validation for changing trigger types, and fixes from bug bash

joehan:
### Description
Adds validation to throw a clearly worded error before deployment when trying to change a functions trigger type in a way that would cause the deployment to fail. Also cleaning up the error messages printed when a deployment fails, and increasing the timeout on our pollers.

### Scenarios Tested

Trying to change a function from https to Firestore (also tested the other direction, as well as Firestore -> Storage, they look the same):
Screen Shot 2021-02-12 at 11 59 00 AM

The updated error messages printed when deployment fails - I also tested & confirmed that the redeploy command works as intended.
Screen Shot 2021-02-12 at 12 00 47 PM

### DIFF
--- a/src/deploy/functions/validate.ts
+++ b/src/deploy/functions/validate.ts
@@ -86,18 +86,25 @@ export function checkForInvalidChangeOfTrigger(
   fn: CloudFunctionTrigger,
   exFn: CloudFunctionTrigger
 ) {
-  let shouldThrow = false;
   if (fn.httpsTrigger && !exFn.httpsTrigger) {
-    shouldThrow = true;
-  }
-  if (fn.eventTrigger?.service != exFn.eventTrigger?.service) {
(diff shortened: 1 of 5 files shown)


src/deploy/functions/validate.ts
```diff
@@ -80,6 +82,25 @@ export function packageJsonIsValid(
   }
 }
 
+export function checkForInvalidChangeOfTrigger(
+  fn: CloudFunctionTrigger,
+  exFn: CloudFunctionTrigger
+) {
```
inlined, line 88: I'd prefer throwing in place with a different error message for each case. That will help when customers come to support asking for questions (or can better explain an issue to a customer before they can reach out to support)

inlined (approved):

PR #3138, opened 2021-02-15
Upgrade logger to TypeScript

inlined:
### Description

Convert logger to TypeScirpt.

I had to subvert the type system here. I can't overwrite the definition of `LogMethod` or `LeveledLogMethod` so I had to subclass it instead. Then to define log.* as using the new expanded signature I had to define a new `Logger` class. These definitions are safe to use in practice because `Logger` is a pure interface, but we had to cast to `unknown` to override the obvious compiler issues.

Fields in `logger.ts` need to be exported, for e.g. `LogLevel` which `util.ts` uses to call `logger[level]()`. To avoid wonkiness I made `logger` a named export.

I highly recommend reviewing this code by commit rather than by files:

1. "Rename logger.js to logger.ts" was created with `git mv`
2. "Rewrite logger.ts to be real typescript" was manually written and should be reviewed.
3. "UpdateJavascript files to import logger by named export" was created with `git grep | sed`
4. "Migrate all TypeScript files to import logger by named import" was created with `git grep | sed`
5. "Update import * as logger to import { logger }" was created with `git grep | sed`
6. "Manually fix outstanding compiler issues" was written manually and should be reviewed.

### Scenarios Tested

firebase deploy # invalid template file to cause deploy errors
firebase deploy # valid case
firebase deploy --debug


src/logger.ts
```diff
@@ -1,29 +1,94 @@
...
+
+function expandErrors(logger: winston.Logger): winston.Logger {
+  const oldLogFunc: winston.LogMethod = logger.log.bind(logger);
+  const newLogFunc: winston.LogMethod = function (
+    levelOrEntry: string | winston.LogEntry,
+    message?: string | Error,
+    ...meta: any[]
+  ): winston.Logger {
+    if (message && message instanceof Error) {
+      message = message.stack || message.name;
```
joehan, line 57: Should this fall back to Error.message instead of Error.name?

src/logger.ts
```diff
@@ -1,29 +1,94 @@
...
   };
+  logger.log = newLogFunc;
   return logger;
 }
 
-const logger = expandErrors(winston.createLogger());
+function annotateDebugLines(logger: winston.Logger): winston.Logger {
+  const debug: winston.LeveledLogMethod = logger.debug.bind(logger);
+  const newDebug: winston.LeveledLogMethod = function (
+    message: string | any | object,
```
joehan, line 71: Why do we need `any | object` here? It seems a bit redundant to me & I don't see any special casing for Object here.

joehan (commented):
Reviewed `Rewrite logger.ts to be real typescript`, I think it looks good with some small questions

joehan (commented):
'manually fix outstanding compiler issues' LGTM as well

joehan (approved):
This LGTM - however, I think Bryan should take a look too since it affects such a large swath of the codebase

bkendall (approved):
LGTM, though I would suggest running the emulators and making sure expected logs show up there as well.

May consider adding a log entry for our sakes (and updating the title of the PR or the commit message) to make it a bit clearer ("typescript the logger" vs "inlined.ts logger")

inlined:
Whoops. Forgive the bad title. I'm still getting into the GitHub groove again. I thought it used to take the title from the first line of the commit message...

Issue #3137, opened 2021-02-15
firebase init remoteconfig will always overwrite existing template files


inlined:
Found through code analysis. It looks like we [overwrite an existing template file regardless of user choice](https://github.com/firebase/firebase-tools/blob/master/src/init/features/remoteconfig.ts#L47).

inlined:
Verified:

```
$ echo '{"key": "value"}' > remoteconfig.template.json
$ firebase init remoteconfig

     ######## #### ########  ######## ########     ###     ######  ########
     ##        ##  ##     ## ##       ##     ##  ##   ##  ##       ##
     ######    ##  ########  ######   ########  #########  ######  ######
     ##        ##  ##    ##  ##       ##     ## ##     ##       ## ##
     ##       #### ##     ## ######## ########  ##     ##  ######  ########

You're about to initialize a Firebase project in this directory:

  /Users/inlined/git/junkdrawer

Before we get started, keep in mind:

  * You are initializing in an existing Firebase project directory


=== Project Setup

First, let's associate this project directory with a Firebase project.
You can create multiple project aliases by running firebase use --add,
but for now we'll just set up a default project.

i  .firebaserc already has a default project, using inlined-junkdrawer.

=== Remoteconfig Setup
? What file should be used for your Remote Config template? remoteconfig.template.json
? File remoteconfig.template.json already exists. Do you want to overwrite the existing Remote Config template? No

i  Writing configuration info to firebase.json...
i  Writing project information to .firebaserc...

✔  Firebase initialization complete!
$ cat remoteconfig.template.json
$ # nothing printed
```

inlined:
Second error: the default is an empty file, which fails deployment.

PR #3148, opened 2021-02-18
Increasing operation poller timeout to prevent unexpected behavior for slow function deployments (#3137)

joehan:
### Description
Increasing operation poller timeout to prevent unexpected behavior for slow function deployments (see #3137).

Additionally, in release.ts, I moved the call to printTriggerUrls to before the call to printErrors, to ensure that it gets called even if some functions experience errors.


samtstern (approved):

samtstern:
Huge +1 for this. Over here in London with my slow internet my functions deploys take ~3m on average and ~5m+ on a bad day, this will save me a lot of headaches.

PR #3150, opened 2021-02-19
Increases timeout for function operations.

### LINKED ISSUE #3147: firebase deploy --only functions:oneFunction fails but function is deployed successfully

**firebase-tools:** 9.4.0

**Platform:** Arch Linux on WSL2 with Node 12.20.1

### [REQUIRED] Test case

I don't have a minimal test case for this.

- Run `firebase deploy --only functions:oneFunction`

- Output in the console that the function was successfully deployed.

- Output says there were errors deploying the function:
image
- If I run the command with the `--debug` flag, the last lines of the logs say this:
image

- But in the Google Cloud Console for Cloud Functions dashboard, in the "last deployed" column it displays the time as if it was correctly deployed, I had to manually test my app and the change was deployed correctly.
image

mbleigh:
A customer reported issues with deploys timing out. When I asked them to investigate closer, it appears that the CLI is timing out but the build does eventually succeed. Increasing the operation timeout to the actual backend limit to avoid false failures.

Fixes #3147

### DIFF
--- a/src/deploy/functions/tasks.ts
+++ b/src/deploy/functions/tasks.ts
@@ -19,7 +19,7 @@ import { ErrorHandler } from "./errorHandler";
 const defaultPollerOptions = {
   apiOrigin: functionsOrigin,
   apiVersion: cloudfunctions.API_VERSION,
-  masterTimeout: 300000, // 300000ms = 5 minutes
+  masterTimeout: 25 * 60000, // 25 minutes is the maximum build time for a function
 };
 
 export interface TaskParams {


dgobaud:
yes please merge and release a new version - thank you! it times out almost every time :(

taeold (approved):

joehan:
LGTM - sorta duplicates https://github.com/firebase/firebase-tools/pull/3148, but i think we want to merge that one as well since it has another small bugfix.

joehan (approved):

mbleigh:
Ah, I didn't see that you already had a pending PR for this. How about you edit your PR to just be about the other change (and I'll keep this one since 25m is the backend timeout)?

PR #3157, opened 2021-02-22
Return version  in `hosting:channel:deploy`

slam:
### Description

`firebase hosting:clone` accepts two forms of the source argument - `site:channel` and `site@version`. The second form allows cloning a specific version and is great for many companies that have a robust QA process to validate a preview build before promoting it to live.

However, there is currently no way to extract the version that was deployed. This PR adds the version info to the `hosting:channel:deploy` output.

### Scenarios Tested

Run `firebase hosting:channel:deploy <test-channel>` with and without the `--json` option. Make sure the version info appears in both text and json outputs.

### Sample Commands

Example output:

Text:

```
✔  hosting:channel: Channel URL (<site-name>): https://<site-name>--<channel>-bf71pekc.web.app [expires 2021-02-28 16:26:30] [version 5b47f66ce57331b4]
```

JSON:

```
{
  "status": "success",
  "result": {
    "<site-name>": {
      "site": "<site-name>",
      "url": "https://<site-name>--<channel>-bf71pekc.web.app",
      "version": "a44e17f03a7e3c2c",
      "expireTime": "2021-03-01T00:25:13.973338740Z"
    }
  }
}
```


src/commands/hosting-channel-deploy.ts
```diff
@@ -28,9 +28,18 @@ interface ChannelInfo {
...
   url: string;
+  version: string;
   expireTime: string;
 }
 
+interface DeployResult {
+  hosting: {
+    [site: string]: {
+      [key: string]: string;
+    };
```
bkendall, line 39: Might the value of this map be the Release interface in `hosting/api.ts`? (I don't know off hand, but it would make for stronger typing)

src/commands/hosting-channel-deploy.ts
```diff
@@ -127,20 +142,31 @@ export default new Command("hosting:channel:deploy [channelId]")
         })
       );
 
-      await deploy(["hosting"], options, { hostingChannel: channelId });
+      const results = (await deploy(["hosting"], options, {
+        hostingChannel: channelId,
+      })) as DeployResult;
```
bkendall, line 147: My TS-brain is rusty: could this be `const results: DeployResult = await deploy()` rather than using `as`?

src/commands/hosting-channel-deploy.ts
```diff
@@ -127,20 +142,31 @@ export default new Command("hosting:channel:deploy [channelId]")
...
       logger.info();
       await syncAuthState(projectId, sites);
       const deploys: { [key: string]: ChannelInfo } = {};
+      const siteDetails = results.hosting;
       sites.forEach((d) => {
-        deploys[d.target || d.site] = d;
+        const siteKey = d.target || d.site;
+        if (siteDetails && siteKey in siteDetails) {
+          d.version = siteDetails[siteKey]["version"];
+        }
```
bkendall, line 157: What is the case where `siteDetails` is not defined?

src/deploy/index.js
```diff
@@ -107,8 +107,12 @@ var deploy = function (targetNames, options, customContext = {}) {
         _.each(context.hosting.deploys, function (deploy) {
           logger.info(clc.bold("Hosting URL:"), utils.addSubdomain(api.hostingOrigin, deploy.site));
         });
-        const versionNames = context.hosting.deploys.map((deploy) => deploy.version);
-        return { hosting: versionNames.length === 1 ? versionNames[0] : versionNames };
+        var siteDetails = {};
```
bkendall, line 110:
```suggestion
        const siteDetails = {};
```

src/deploy/index.js
```diff
@@ -107,8 +107,12 @@ var deploy = function (targetNames, options, customContext = {}) {
...
+        context.hosting.deploys.forEach((deploy) => {
+          var version = deploy.version.replace(`sites/${deploy.site}/versions/`, "");
```
bkendall, line 112: I'm not actually sure we want to truncate the full name to just the version ID. If we wanted to only display that later, we can manipulate the name at that location.

src/deploy/index.js
```diff
@@ -107,8 +107,12 @@ var deploy = function (targetNames, options, customContext = {}) {
...
+          siteDetails[deploy.site] = { version };
+        });
+        return { hosting: siteDetails };
```
bkendall, line 115: This change may also have an effect on the output of `deploy --json`. Could you check the before/after of the behavior there?

bkendall (changes requested):
I like the idea here, but I'm not certain that this is safe as-is because it changes the returned value of deploy. If you check the output of `deploy --json` before and after, it must not be a breaking change (well, it can be, but that would require a different effort).

src/deploy/index.js
slam: I had `const` initially, but then noticed that this `js` file is using `var` everywhere else. I'd like to leave it as `var` if it is okay with you.

src/deploy/index.js
slam: That's a good point. Will move that logic into `host-channel-deploy.ts`.

src/deploy/index.js
slam: It looks compatible. The only new field is `version`.

Before:

```
{
  "status": "success",
  "result": {
    "<site>": {
      "site": "<site>",
      "url": "https://<site>--slam-test-6161efgc.web.app",
      "expireTime": "2021-03-02T19:40:07.615661137Z"
    }
  }
}
```

After:

```
{
  "status": "success",
  "result": {
    "<site>": {
      "site": "<site>",
      "url": "https://<site>--slam-test-6161efgc.web.app",
      "version": "09ef5462c91e2bba",
      "expireTime": "2021-03-02T19:39:05.793378023Z"
    }
  }
}
```

slam: This makes sense to me because before this PR the return value of `deploy()` was not used anywhere:

https://github.com/firebase/firebase-tools/pull/3157/files#diff-90fa65dc2e075506e4e87cc1c6bee0c8a1e769440908e956179ad0fa4a79d999L130

src/commands/hosting-channel-deploy.ts
slam: The method `deploy()` returns `any` since it is implemented in js. Without the `as` cast, eslint issues this warning: `Unsafe assignment of an any value.`.

I'm fine with either construct. Please advise.

src/commands/hosting-channel-deploy.ts
slam: This is a great idea. It would require a bit more code changes to capture the payload returned by the hosting release api, but it makes more sense to me. Let me give it a try.

src/commands/hosting-channel-deploy.ts
slam: Good point. Removed.

slam:
@bkendall could you pls review again when you get a chance?

slam:
> @bkendall could you pls review again when you get a chance?

It's been a month since I responded to the feedback. Is there anything else I could do to speed this up?

bkendall:
Thanks for your patience. Unfortunately there's been other more pressing things on my plate recently, but I hope to come back around to this soon.

bkendall:
When I do a normal deploy (`firebase deploy --only hosting`) on `master`, I get this output (with `--json`):

```
{
  "status": "success",
  "result": {
    "hosting": "sites/bkend-test-stuff/versions/db18ce3e1f438cb2"
  }
}
```

Now I'm getting an object at that `hosting` key rather than a string (which is a breaking change): 

```
{
  "status": "success",
  "result": {
    "hosting": {
      "bkend-test-stuff": {
        "release": {
          "name": "sites/bkend-test-stuff/releases/1617216425557023",
          "version": {
            "name": "sites/bkend-test-stuff/versions/e23837c2b1183abb",
            // Other fields trimmed for sanity
        }
      }
    }
  }
}
```

Checking on `hosting:channel:deploy`, it does look like only `version` is added, which is safe.

Looking at the code, I think the version may be able to be pulled from the bit of code that gets/creates the channels before calling `deploy`. That would allow the version to be added into `hosting:channel:deploy`'s output while not changing `firebase deploy`'s output. What do you think?

bkendall (changes requested):
(See previous comment regarding `firebase deploy`)

slam:
> (See previous comment regarding `firebase deploy`)

Thank you for the explanation and the example. I finally got it.

Updated the code. Decided not to touch `src/deploy/index.js` at all, and limited all the changes to `hosting:channel:deploy`.

bkendall (changes requested):
This looks great. Could you add an entry to the changelog?

```markdown
- Adds the deployed version to the output when deploying to Firebase Hosting.
```

LGTM otherwise! Thanks for being incredibly patient with me

src/commands/hosting-channel-deploy.ts
```diff
@@ -137,9 +160,18 @@ export default new Command("hosting:channel:deploy [channelId]")
         if (d.expireTime) {
           expires = `[expires ${bold(datetimeString(new Date(d.expireTime)))}]`;
         }
+        const versionPrefix = `sites/${d.target || d.site}/versions/`;
```
bkendall, line 163: I tested this out with a test project that has two sites and deploy targets set up for them. This line ends up breaking because `d.target` isn't what `d.site` is (`target` is the local alias for it). I think `.site` is the only thing that's needed here.

```suggestion
        const versionPrefix = `sites/${d.site}/versions/`;
```

Could you double check this fix? I think it should be fine, but it'd be nice for you to test as well :)

bkendall (changes requested):

src/commands/hosting-channel-deploy.ts
slam: Good catch. `version` is empty:

```
$ firebase --debug hosting:channel:deploy t1-test --project slam-home-api --json --only t1   
{
  "status": "success",
  "result": {
    "t1": {
      "site": "thing1",
      "target": "t1",
      "url": "https://thing1--t1-test-ltf4oe3y.web.app",
      "version": "",
      "expireTime": "2021-05-21T00:12:33.800426131Z"
    }
  }
}
```

Will fix.

slam: Fixed:

```
$ firebase --debug hosting:channel:deploy t1-test --project slam-home-api --json --only t1
{
  "status": "success",
  "result": {
    "t1": {
      "site": "thing1",
      "target": "t1",
      "url": "https://thing1--t1-test-ltf4oe3y.web.app",
      "version": "9af1bb9abd9d5d01",
      "expireTime": "2021-05-21T00:16:09.860986599Z"
    }
  }
}
```

bkendall (approved):
LGTM. I may not merge it right this moment, but I'll do so soon! Thanks for the PR!

PR #3160, opened 2021-02-22
Fixes error tracking for function deployments

joehan:
### Description
Correcting an issue where all function deployments were being tracked as successful.

### DIFF
--- a/src/functionsDeployHelper.ts
+++ b/src/functionsDeployHelper.ts
@@ -184,8 +184,8 @@ export function logAndTrackDeployStats(queue: Queue<any, any>, errorHandler: Err
   logger.debug(`${errorHandler.errors.length} Functions Errored`);
   logger.debug(`Average Function Deployment time: ${stats.avg}`);
   if (stats.total > 0) {
-    track("Functions Deploy (Result)", "failure", stats.errored);
-    track("Functions Deploy (Result)", "success", stats.success);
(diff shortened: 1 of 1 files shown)


taeold (approved):

src/functionsDeployHelper.ts
```diff
@@ -184,8 +184,8 @@ export function logAndTrackDeployStats(queue: Queue<any, any>, errorHandler: Err
   logger.debug(`${errorHandler.errors.length} Functions Errored`);
   logger.debug(`Average Function Deployment time: ${stats.avg}`);
   if (stats.total > 0) {
-    track("Functions Deploy (Result)", "failure", stats.errored);
-    track("Functions Deploy (Result)", "success", stats.success);
+    track("Functions Deploy (Result)", "failure", errorHandler.errors.length);
+    track("Functions Deploy (Result)", "success", stats.total - errorHandler.errors.length);
```
mbleigh, line 188: Since this data is completely broken, I wonder if we should consider tracking it differently. What if we did something like:

```suggestion
    if (errorHandler.errors.length === 0) {
      track("functions_deploy_result", "success", stats.total);
    } else if (errorHandler.errors.length < stats.total) {
      track("functions_deploy_result", "partial", stats.total);
    } else {
      track("functions_deploy_result", "failure", stats.total)
    }
```

This way we end up with discrete numbers for successful deploys, partial successes, and full failures.

PR #3162, opened 2021-02-22
Rethinking metrics for functions deployment

joehan:
### Description
Rethinking functions deploy metrics. This strategy lets us understand:
- How many individual functions have succeeded/errored? What is the overall success rate?
- How many fully successful, partially successful, and completely failed deployments?
- When a deployment partially fails, how big is it on average? What percent of its functions didn't successfully deploy, on average?

### DIFF
--- a/src/functionsDeployHelper.ts
+++ b/src/functionsDeployHelper.ts
@@ -184,8 +184,19 @@ export function logAndTrackDeployStats(queue: Queue<any, any>, errorHandler: Err
   logger.debug(`${errorHandler.errors.length} Functions Errored`);
   logger.debug(`Average Function Deployment time: ${stats.avg}`);
   if (stats.total > 0) {
-    track("Functions Deploy (Result)", "failure", errorHandler.errors.length);
-    track("Functions Deploy (Result)", "success", stats.total - errorHandler.errors.length);
+    if (errorHandler.errors.length === 0) {
+      track("functions_deploy_result", "success", stats.total);
+    } else if (errorHandler.errors.length < stats.total) {
+      track("functions_deploy_result", "partial_success", stats.total - errorHandler.errors.length);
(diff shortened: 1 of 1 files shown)


src/functionsDeployHelper.ts
```diff
@@ -184,8 +184,19 @@ export function logAndTrackDeployStats(queue: Queue<any, any>, errorHandler: Err
   logger.debug(`${errorHandler.errors.length} Functions Errored`);
   logger.debug(`Average Function Deployment time: ${stats.avg}`);
   if (stats.total > 0) {
-    track("Functions Deploy (Result)", "failure", errorHandler.errors.length);
-    track("Functions Deploy (Result)", "success", stats.total - errorHandler.errors.length);
+    if (errorHandler.errors.length === 0) {
+      track("functions_deploy_result", "success", stats.total);
+    } else if (errorHandler.errors.length < stats.total) {
+      track("functions_deploy_result", "partial_success", stats.total - errorHandler.errors.length);
+      track("functions_deploy_result", "partial_failure", errorHandler.errors.length);
```
mbleigh, line 191: I'd still weakly argue that "average number of functions attempted to deploy on partial failure" is more useful than "average number of successes/failures on a partial failure". So this suggestion would be my change, but I don't think I care enough to argue with you if you just want to say "nah I prefer this."

```suggestion
      track("functions_deploy_result", "partial", stats.total);
```

mbleigh (approved):

src/functionsDeployHelper.ts
joehan: Agreed with you on that - however, my implementation lets you calculate that as well (its just the value of partial_success + value of  partial_failure). The reason to split these up is that it lets us calculate the overall rate of individual function successes vs failures. Thats a metric we wished we had while trying to calculate # of builds saved, and AFAIK, theres no way to get it for only firebase functions from API side metrics.

PR #3164, opened 2021-02-23
Only prompt for failurePolicies if they are new.

inlined:
### Description

Fixes https://github.com/firebase/firebase-functions/issues/798

We can now conditionally list functions earlier in the deploy pipeline
if there are any functions with a failure policy in the current
set of functions to deploy. Silence the prompt if all functions with
failure policies already had failure policies.

Also modifies prepare.ts to avoid redundantly calling listAllFunctions
later.

### Scenarios Tested

1. Write a function with a failure policy
2. Deploy (expect a prompt)
3. Deploy (do not expect a prompt)

### Sample Commands

`firebase deploy [--only functions]`

### Notes

Tests previously checked for throwing instead of rejecting, which worked if and only if the throw happened before yielding the thread. This broke all tests that ran into the new check for existing functions.


src/deploy/functions/prompts.ts
```diff
@@ -25,21 +27,46 @@ export async function promptForFailurePolicies(
   if (failurePolicyFunctions.length === 0) {
     return;
   }
-  const failurePolicyFunctionLabels = failurePolicyFunctions.map((fn: CloudFunctionTrigger) => {
-    return getFunctionLabel(fn.name);
+
+  // N.B. Because context is an any, we need to store it in a temporary. This
+  // lets us do a 'for of' loop later with typeinfo.
+  const existingFunctions = await gcp.cloudfunctions.listAllFunctions(context.projectId);
```
joehan, line 33: We already make this call once during the deployment process, at the start of 'deploy.ts' - WDYT about reordering things such that we only need to do this once?

joehan (commented):
Mostly LGTM, but I'd prefer to not repeat API calls unnecessarily

joehan (approved):

PR #3165, opened 2021-02-23
Fix problem with typescript initial eslint config for functions project

MateoKruk:
### Description
When setting up a brand new functions project with eslint. I was getting the following linting errors:

image
image

After some research I found we where missing extending some `@typescript-eslint` options in the eslint config. [link](https://stackoverflow.com/questions/57802057/eslint-configuring-no-unused-vars-for-typescript)

### DIFF
--- a/templates/init/functions/typescript/_eslintrc
+++ b/templates/init/functions/typescript/_eslintrc
@@ -10,6 +10,8 @@ module.exports = {
     "plugin:import/warnings",
     "plugin:import/typescript",
     "google",
+    "plugin:@typescript-eslint/eslint-recommended",
+    "plugin:@typescript-eslint/recommended",
   ],
   parser: "@typescript-eslint/parser",
   parserOptions: {


MateoKruk:
@googlebot I signed it!

templates/init/functions/typescript/_eslintrc
```diff
@@ -10,6 +10,8 @@ module.exports = {
     "plugin:import/warnings",
     "plugin:import/typescript",
     "google",
+    "plugin:@typescript-eslint/eslint-recommended",
```
bkendall, line 13: What is `@typescript-eslint/eslint-recommended`?  I'm having a hard time finding documentation for it.

templates/init/functions/typescript/_eslintrc
```diff
@@ -10,6 +10,8 @@ module.exports = {
...
+    "plugin:@typescript-eslint/recommended",
```
bkendall, line 14: If we were to add these, I would rather put them immediately below the `eslint:recommended` line so that the following entries can override as appropriate.

bkendall (changes requested):

templates/init/functions/typescript/_eslintrc
MateoKruk: It's a bit buried in the documentation but you can see it [here](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/README.md)

templates/init/functions/typescript/_eslintrc
MateoKruk: Yes this makes sense. I think `"plugin:@typescript-eslint/eslint-recommended"` it's not really needed. I'll change that

bkendall (commented):
This LGTM. I'm going to ask one of the Functions folks to look quick too

bkendall (approved):

templates/init/functions/typescript/_eslintrc
```diff
@@ -6,6 +6,7 @@ module.exports = {
   },
   extends: [
     "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
```
joehan, line 9: Can you move this to after 'google'? This ruleset [turns off some common rules](https://github.com/typescript-eslint/typescript-eslint/blob/498f397ff3898dde631f37311615b555f38a414e/packages/eslint-plugin/src/configs/recommended.ts#L27) in favor of the plugin's version, but the 'google' ruleset is [turning those back on](https://github.com/google/eslint-config-google/blob/master/index.js#L167). As is, this ends throwing repetitive errors and warnings for the same issue in some cases.

joehan (changes requested):
LGTM if we move this to the end of the extends block

templates/init/functions/typescript/_eslintrc
MateoKruk: Done!

joehan (approved):

PR #3166, opened 2021-02-23
fixes issue with hosting emulator targets

### LINKED ISSUE #2820: firebase emulators:start --only hosting:TARGET_NAME command does not work
First, I did add a custom target by running `firebase target:apply demo my-project`. However, when running `firebase emulators:start --only hosting:demo` I receive the following error message:

> Error: hosting:demo is not a valid emulator name, valid options are: ["auth","functions","firestore","database","hosting","pubsub"]

As per documentation (https://firebase.google.com/docs/hosting/multisites?authuser=0#cli-commands-with-deploy-targets) `firebase emulators:start --only hosting:TARGET_NAME` should emulate the Hosting content and configuration of the specified Hosting site at a locally hosted URL.

bkendall:
### Description

Fixes #2820

### Scenarios Tested

- `emulators:start --only hosting:one`
- `emulators:start --only hosting:two`
- `emulators:start --only hosting` (to see both start)
- `emulators:start --only functions:echo` (to make sure my function still works)

### DIFF
--- a/src/emulator/controller.ts
+++ b/src/emulator/controller.ts
@@ -174,8 +174,12 @@ export function filterEmulatorTargets(options: any): Emulators[] {
     return options.config.has(e) || options.config.has(`emulators.${e}`);
   });
 
-  if (options.only) {
-    targets = _.intersection(targets, options.only.split(","));
+  const onlyOptions: string = options.only;
+  if (onlyOptions) {
+    const only = onlyOptions.split(",").map((o) => {
+      return o.split(":")[0];
+    });
+    targets = _.intersection(targets, only as Emulators[]);
   }
 
   return targets;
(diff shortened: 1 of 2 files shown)


src/emulator/controller.ts
```diff
@@ -321,8 +325,11 @@ export async function startAll(options: any, showUI: boolean = true): Promise<vo
     "emulators",
     `Starting emulators: ${targets.join(", ")}`
   );
-  if (options.only) {
-    const requested: string[] = options.only.split(",");
+  const onlyOptions: string = options.only;
```
samtstern, line 328: Do we have to repeat this code block twice?

samtstern (approved):

src/emulator/controller.ts
bkendall: Yes. It likely can be refactored in the future, but since it's more lines to refactor than it would be to duplicate, I'm leaving it as-written.

PR #3173, opened 2021-02-24
Authentication Emulator Supports MFA for Simple Read/Write User Operations (Fixes #3170)

### LINKED ISSUE #3170: Authentication Emulator returns `undefined` Multi-Factor Info for Users
<!-- DO NOT DELETE
validate_template=true
template_path=.github/ISSUE_TEMPLATE/bug_report.md
-->

When running the **node admin SDK** locally with an Authentication Emulator, user operations allow multifactor data on write, but do not reflect multifactor data on read.

**firebase-tools:**
```
% ./node_modules/.bin/firebase --version
9.5.0
```

**Platform:**

```
% node -v
v12.16.3
% npm -v
6.14.4
% ./node_modules/.bin/tsc -v 
Version 4.1.3
```

```
% uname -mrsv
Darwin 20.2.0 Darwin Kernel Version 20.2.0: Wed Dec  2 20:39:59 PST 2020; root:xnu-7195.60.75~1/RELEASE_X86_64 x86_64
```
### [REQUIRED] Test case

Assumes a typescript project with `firebase-admin` and `jest` installed, and `firebase-tools` installed as a dev dependency.

```
import firebaseAdmin from 'firebase-admin';

let app: firebaseAdmin.app.App;

beforeAll(() => {
    app = firebaseAdmin.initializeApp({
        projectId: 'run-me-in-an-emulator',
    });
});

afterAll(async () => {
    await app.delete();
});

test.only('Firebase Authentication Emulator Returns MultiFactorSettings Data', async () => {
    const user = await app.auth().createUser({
        email: `test@test.com`,
        emailVerified: true,
        password: undefined,
        multiFactor: {
            enrolledFactors: [
                {
                    phoneNumber: '+12813308004',
                    factorId: 'phone',
                    displayName: 'Mobile',
                } as firebaseAdmin.auth.CreatePhoneMultiFactorInfoRequest,
            ],
        },
    });

    expect(user.multiFactor).toBeTruthy();
});
```


Run the following test with the `node.js` admin SDK using a local firebase authentication emulator, something like:

```
./node_modules/.bin/firebase emulators:exec 'npm run test firebaseRepro.test.ts'
```

The test does the following:

1. Create a user with multi-factor info using the node admin sdk.

UserRecord object is returned with a non-null `multiFactor` MultiFactorSettings object.

UserRecord object is returned with an undefined `multiFactor` MultiFactorSettings object.

```
 % ./node_modules/.bin/firebase emulators:exec 'npm run test firebaseRepro.test.ts'
i  emulators: Starting emulators: auth, pubsub
i  pubsub: Pub/Sub Emulator logging to pubsub-debug.log
i  Running script: npm run test firebaseRepro.test.ts

> jest "firebaseRepro.test.ts"

 FAIL  src/firebaseRepro.test.ts (5.781 s)
  ✕ Firebase Authentication Emulator Returns MultiFactorSettings Data (74 ms)

  ● Firebase Authentication Emulator Returns MultiFactorSettings Data

    expect(received).toBeTruthy()

    Received: undefined

      29 |     });                                                                                                                                                                                                                                                                               
      30 |
    > 31 |     expect(user.multiFactor).toBeTruthy();                                                                                                                                                                                                                                            
         |                              ^
      32 | });                                                                                                                                                                                                                                                                                   
      33 |

      at src/firebaseRepro.test.ts:31:30
      at fulfilled (src/firebaseRepro.test.ts:5:58)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        5.818 s, estimated 7 s
Ran all test suites matching /firebaseRepro.test.ts/i.
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! test: `jest "firebaseRepro.test.ts"`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the test script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/wokkaflokka/.npm/_logs/2021-02-24T00_05_17_664Z-debug.log
⚠  Script exited unsuccessfully (code 1)
i  pubsub: Stopping Pub/Sub Emulator
i  auth: Stopping Authentication Emulator
i  hub: Stopping emulator hub

Error: Script "npm run test firebaseRepro.test.ts" exited with code 1

```

wokkaflokka:
### Description

This PR proposes an initial resolution for #3170 by implementing rudimentary support for "mutli-factor" information during admin SDK operations (_e.g._, `admin.auth().createUser(...)`).

Note, deletes, client operations, and advanced scenarios are intentionally excluded from this change at this time. It seems like there is large potential scope of support for MFA in general, and possibly some changes going on under the hood, and I wanted to keep this as focused as possible to start.

### Scenarios Tested

This change is focused on supporting read/write operations from the admin SDK while using the Authentication Emulator. It has been tested using the test case provided on #3170 as well as a modified variant of the test case from #3170, provided here.

```
import firebaseAdmin from 'firebase-admin';

let app: firebaseAdmin.app.App;

beforeAll(() => {
    app = firebaseAdmin.initializeApp({
        projectId: 'run-me-in-an-emulator',
    });
});

afterAll(async () => {
    await app.delete();
});

test('Firebase Authentication Emulator Returns MultiFactorSettings Data', async () => {
    expect(!!process.env.FIREBASE_AUTH_EMULATOR_HOST).toEqual(true);

    const user = await app.auth().createUser({
        email: 'test@test.com',
        emailVerified: true,
        password: undefined,
        multiFactor: {
            enrolledFactors: [
                {
                    phoneNumber: '+12813308004',
                    factorId: 'phone',
                    displayName: 'Mobile',
                } as firebaseAdmin.auth.CreatePhoneMultiFactorInfoRequest,
            ],
        },
    });

    expect(user.multiFactor).toEqual({
        enrolledFactors: [{
            phoneNumber: '+12813308004',
            factorId: 'phone',
            displayName: 'Mobile',
            enrollmentTime: null,
            uid: expect.stringMatching(/^.+$/),
        }],
    });
});
```

### Sample Commands

To run the test case above, configure a minimal typescript project with the implied dependencies and execute something like:

`firebase emulators:exec --debug 'npm run test firebaseRepro.test.ts'`

Fixes #3170.


samtstern:
@wokkaflokka thanks for the super clear issue and the PR! Assigning to @yuchenshi who does most of the Auth emulator things around here.

src/emulator/auth/state.ts
```diff
@@ -171,10 +173,30 @@ export class ProjectState {
...
     return this.updateUserProviderInfo(user, upsertProviders, deleteProviders);
   }
 
+  private updateMfaEnrollments(user: UserInfo): UserInfo {
+    for (const factor of user.mfaInfo ?? []) {
+      const allowUpdate = !!factor.mfaEnrollmentId;
+      const mfaEnrollmentId = factor.mfaEnrollmentId ?? randomId(28);
+      const enrollments = this.mfaEnrollmentIdsForLocalId.get(user.localId) ?? new Set();
+      if (!(enrollments.add(mfaEnrollmentId) || allowUpdate)) {
+        throw new Error("New MfaEnrollmentId wasn't unique.");
```
yuchenshi, line 191: This check also needs to happen in `operations.ts` with the right error code. This check here should only serve as an internal sanity check.

src/test/emulators/auth/signUp.spec.ts
```diff
@@ -395,4 +395,29 @@ describeAuthEmulator("accounts:signUp", ({ authApi }) => {
         expect(res.body.error.message).to.equal("INVALID_PHONE_NUMBER : Invalid format.");
       });
   });
+
+  it("should create new account with multi factor info when authenticated", async () => {
```
yuchenshi, line 399: need test cases for invalid payloads and duplicate IDs

yuchenshi (changes requested):
Thanks for the PR. Unfortunately, this creates a confusing situation where a user with `mfaInfo` can still sign in without MFA. I understand that you won't need to full implementation for you use case, but would you mind making it so that users with `mfaInfo` cannot sign in _at all_ with a `NotImplementedError`?

wokkaflokka:
@yuchenshi thanks for the review. I'll get to work on those changes.

src/emulator/auth/state.ts
wokkaflokka: @yuchenshi I am working on improving this validation. In doing so, it unearthed a few things worth considering.

There are three ways a user can be opted into MFA via the Firebase SDK. 

1. Using the Admin SDK `createUser` API 
2. Using the Admin SDK `updateUser` API. 
3.. Using the client SDK.

As far as I can tell, there is no support for the client SDK MFA enrollment flow in the Firebase Emulator (I could be wrong, here). So, we only need to focus on the Admin SDK entrypoints in the scope of this PR.

The schema for MFA Info on the sign up request doesn't specify an `mfaEnrollmentId` field, and moreoever, the Admin SDK [validates and rejects any attempt](https://github.com/firebase/firebase-admin-node/blob/master/src/auth/auth-api-request.ts#L1499-L1515) to specify an enrollment ID in the "createUser" flow. So, for the Admin SDK creation case, the MFA enrollment ID will always be [absent in the signUp flow](https://github.com/firebase/firebase-admin-node/blob/master/src/auth/user-import-builder.ts#L96-L113), and we will always need to [create a new enrollmentId](https://github.com/firebase/firebase-admin-node/blob/master/src/auth/auth-api-request.ts#L278).


Here's the underlying type for an instance of MFA Info on the create request, as well (no ID-like field present):

```
    GoogleCloudIdentitytoolkitV1MfaFactor: {
      /**
       * Display name for this mfa option e.g. "corp cell phone".
       */
      displayName?: string;
      /**
       * Phone number to receive OTP for MFA.
       */
      phoneInfo?: string;
    };
  ```
  
As I worked on this change, it seemed like this PR would be incomplete without extending support (and validation) to the API's exercised by the `updateUser` API as well. So, I am taking the spirit of this comment, and working it into the `signUp` and `setAccountInfo` flows, while also accounting for the fact that `mfaEnrollmentId` must not be specified on a `signUp` request. I am keeping these commits separate, so if there's a desire to limit the scope, we can evaluate rolling back some or all of the subsequent changes related to this.

Comment (yuchenshi) [reply src/emulator/auth/state.ts:191]: Great work so far. There's indeed no support for client-side MFA enrollment flow and yes, let's leave it like that.

As you noticed, `MfaFactor` only contains a subset fields (`displayName`, `phoneInfo`) compared to `MfaEnrollment`, making it impossible to specify an ID through `signUp`. Please kindly write a test to try to specify `mfaEnrollmentId` and see what happens. Best case, our routers will automatically block it. Otherwise, you may want to code defensively (e.g. picking only two fields manually from `reqBody.mfaInfo`). And I'd like to see that happen within this PR.

On the other hand,`setAccountInfo` seems to require `mfaEnrollmentId`, and that validation should be done in `operations.ts`. Can one reuse an existing second factor (but say, add one more)? What is the exact error code when adding new factor but `mfaEnrollmentId` isn't unique? And is uniqueness enforced for each account, or across all accounts? This can raise a lot of edge cases here. If you want to extend support, feel free to test against your real project and document the finding as tests. If you want to limit the scope, you can also choose to only support typical payloads issued by the admin SDK and throw `NotImplementedError` on anything else.

Comment (wokkaflokka) [reply src/emulator/auth/state.ts]: I went ahead and extended support to the update case as well (along with a _hopefully_ spanning set of test cases). I really appreciate the thoughtful comment you left previously, here are some responses to the framing questions you posed.

> Can one reuse an existing second factor (but say, add one more)? 

Using the real backend, yes, a user can re-use an existing second factor, update an existing second factor, add another second factor, remove a second factor, remove all second factors. I attempted to support all of this behavior in the scope of this change.

> What is the exact error code when adding new factor but mfaEnrollmentId isn't unique?

  I created a distinct error for this scenario in the emulator (`INVALID_MFA_ID : enrollment ID was not unique.`). I’m not sure offhand how the admin API works against the real backend when trying to re-use an existing ID that is owned by another user, and if it throws an error, what the specific error is. I will make a note to test this locally.  
  
  > ... is uniqueness enforced for each account, or across all accounts?  
  
  My assumption — currently untested — is that uniqueness for the ID is enforced across all accounts. It’s worth noting, uniqueness is not enforced for the MFA phone number using the real API — any number of users can share the same underlying phone number for MFA.

There are a few behaviors resulting from how update is implemented as it pertains to “state” as managed by `src/emulators/auth/state.ts`. Certain updates may effectively “delete” a device corresponding to an “enrollment ID”, and in those scenarios the identifier for the deleted device will remain in the `localIdForMfaEnrollmentId` map, and similarly, the device is not removed from the set of ID’s maintained at `mfaEnrollmentIdsForLocalId`. These stateful maps are currently only cleared when `deleteAllAccounts` is called. It would be possible to extend this change to clean up the stateful maps storing MFA enrollment data, but it also seems like the current implementation may suffice for now. I think this only becomes an issue when a request to create a device has an ID collision with one of the deleted devices (which should be mitigated by randomness). Do you have an opinion here?

**Comment by wokkaflokka** [comment]
@yuchenshi appreciate all of your help to date. I've completed a second pass, addressing your initial review comments and extending support to the update case. My personal schedule should be less chaotic for the foreseeable future, so I should also be able to respond to any and all feedback on a quicker cycle at this time.

**Summary of Changes**

1. PR updated to throw `NotImplementedError` for all "sign in" methods whenever a user with `mfaInfo` attempts to sign-in ([9e6da88](https://github.com/firebase/firebase-tools/pull/3173/commits/9e6da8816ad62430648e10f091827a829114da76)).
2. Additional test-cases added around validation and formatting for "sign-up" related MFA flows ([a41f116](https://github.com/firebase/firebase-tools/pull/3173/commits/a41f1160730b1585a2a1d2f513f0f366a3190180).
3. Validation brought into `operations.ts` for MFA-related operations ([a41f116](https://github.com/firebase/firebase-tools/pull/3173/commits/a41f1160730b1585a2a1d2f513f0f366a3190180).
4. Extended MFA support to update flows and added additional test cases for the update case ([a41f116](https://github.com/firebase/firebase-tools/pull/3173/commits/a41f1160730b1585a2a1d2f513f0f366a3190180).

File: src/emulator/auth/state.ts [hunk 3]
```diff
@@ -21,6 +21,8 @@ export class ProjectState {
   private userIdForProviderRawId: Map<string, Map<string, string>> = new Map();
   private refreshTokens: Map<string, RefreshTokenRecord> = new Map();
   private refreshTokensForLocalId: Map<string, Set<string>> = new Map();
+  private mfaEnrollmentIdsForLocalId: Map<string, Set<string>> = new Map();
```
yuchenshi, line 24: How do you feel about just dropping `mfaEnrollmentIdsForLocalId` altogether for this PR? I don't think it's read anywhere, and the same information can be easily retrieved from `user.mfaInfo` when we need it. This will make the code much much easier.

src/emulator/auth/state.ts
```diff
@@ -171,10 +173,58 @@ export class ProjectState {
     } else {
       deleteProviders.push(PROVIDER_PHONE);
     }
+    if (user.mfaInfo) {
```
yuchenshi, line 176: To keep it consistent, we also need to go through the existing mfaEnrollmentIds and remove the ones that will no longer be present after the update. A simple yet powerful implementation to cover all cases: 1. remove all from `user.mfaInfo` (and also from index) 2. for each enrollments in `fields.mfaInfo`, add to `user` and also add to index. All of these should be done in `updateMfaEnrollments` function and we don't need this if statement here.

src/test/emulators/auth/helpers.ts
```diff
@@ -113,6 +114,29 @@ export function registerAnonUser(
     });
 }
 
+export function registerMfaUser(
```
yuchenshi, line 117: Combine this with `registerUser` above. afacit this is just `registerUser` but with an additional request field.

src/test/emulators/auth/customToken.spec.ts
```diff
@@ -231,4 +233,25 @@ describeAuthEmulator("sign-in with custom token", ({ authApi }) => {
...
   });
+
+  it("should error if user has MFA", async () => {
+    const user = {
+      email: "alice@example.com",
+      password: "notasecret",
+      mfaInfo: [{ displayName: "Cell Phone", phoneInfo: TEST_PHONE_NUMBER }],
+    };
+    const { localId } = await registerMfaUser(authApi(), user);
+    expect(localId).to.be.a("string").and.not.empty;
```
yuchenshi, line 244:
```suggestion
```

Don't need to test the test helpers

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -415,6 +416,416 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
+        localId,
+        mfa: {
+          enrollments: [mfaInfo],
+        },
+      })
+      .then((res) => expectStatusCode(200, res));
+
+    const info = await getAccountInfoByIdToken(authApi(), idToken);
+    expect(info.mfaInfo).to.have.length(1);
+    const updated = info.mfaInfo?.pop();
```
yuchenshi, line 442:
```suggestion
    const updated = info.mfaInfo![0];
```

Don't mutate the array

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -415,6 +416,416 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
+      displayName: "Cell Phone",
+      phoneInfo: TEST_PHONE_NUMBER,
+    };
+    const user = {
+      email: "bob@example.com",
+      password: "notasecret",
+      mfaInfo: [mfaInfo, mfaInfo, mfaInfo],
+    };
+    const { localId, idToken } = await registerMfaUser(authApi(), user);
+    let info = await getAccountInfoByIdToken(authApi(), idToken);
```
yuchenshi, line 567: Don't reuse the variable -- use two distinct ones for readability. `oldInfo` and `newInfo` may be a bit better here.

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -415,6 +416,416 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
+        localId,
+        mfa: {
+          enrollments: [newMfaInfo],
+        },
+      })
+      .then((res) => expectStatusCode(200, res));
+
+    info = await getAccountInfoByIdToken(authApi(), idToken);
+    expect(info.mfaInfo).to.have.length(1);
+    const updated = info.mfaInfo?.pop();
```
yuchenshi, line 588: Don't mutate the array here, or your for-loop below won't work as you expect.
```suggestion
    const updated = info.mfaInfo[0];
```

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -415,6 +416,416 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
+    expect(updated?.phoneInfo).to.eq(newMfaInfo.phoneInfo);
+    expect(updated?.displayName).to.eq(newMfaInfo.displayName);
+    expect(updated?.mfaEnrollmentId).to.eq(newMfaInfo.mfaEnrollmentId);
+    for (const { mfaEnrollmentId } of info.mfaInfo ?? []) {
```
yuchenshi, line 592: What are you trying to test here? If you want to see if the updated mfaEnrollmentId is different than the previous three, then I think you'd want `expect(oldEnrollmentIds).not.to.include(updated.mfaEnrollmentId)`, where `oldEnrollmentIds` should be saved earlier.

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -415,6 +416,416 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
+
+  it("should error if mfaEnrollmentId is absent", async () => {
+    const mfaInfo: any = {
+      displayName: "Cell Phone",
+      phoneInfo: TEST_PHONE_NUMBER,
+    };
+    const user = { email: "bob@example.com", password: "notasecret", mfaInfo: [mfaInfo] };
+    const { localId, idToken } = await registerMfaUser(authApi(), user);
+    const info = await getAccountInfoByIdToken(authApi(), idToken);
+    expect(info.mfaInfo).to.have.length(1);
```
yuchenshi, line 663:
```suggestion
```

Probably don't need this since this is all setup code

src/emulator/auth/operations.ts
```diff
@@ -166,6 +166,31 @@ function signUp(
...
+        "UNEXPECTED_PARAMETER : mfaEnrollmentId"
+      );
+      assert(
+        factor.phoneInfo && isValidPhoneNumber(factor.phoneInfo),
+        "INVALID_MFA_PHONE_NUMBER : Invalid format."
+      );
+      const mfaEnrollmentId = randomId(28);
+      assert(
+        state.validateMfaEnrollmentIdForCreate(mfaEnrollmentId),
+        "INVALID_MFA_ID : enrollment ID was not unique."
```
yuchenshi, line 185: This should just throw an internal error since we got unlucky here and got a duplicate. There's nothing the API caller can do to fix this (apart from retrying). If you feel especially charitable, you may want to generate multiple times and pick the first one available (like what we do for localIds).

src/emulator/auth/operations.ts
```diff
@@ -974,6 +999,31 @@ export function setAccountInfoImpl(
...
+            enrolledFactor.phoneInfo && isValidPhoneNumber(enrolledFactor.phoneInfo),
+            "INVALID_MFA_PHONE_NUMBER : Invalid format."
+          );
+          assert(
+            enrolledFactor.mfaEnrollmentId,
+            "INVALID_MFA_ID : Must specify non-null mfaEnrollmentId on update."
+          );
+          assert(
+            state.validateMfaEnrollmentIdForUpdate(enrolledFactor.mfaEnrollmentId, user),
+            "INVALID_MFA_ID : enrollment ID was not unique."
```
yuchenshi, line 1017:
```suggestion
            "DUPLICATE_MFA_ENROLLMENT_ID"
```

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -415,6 +416,416 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
+      })
+      .then((res) => {
+        expectStatusCode(400, res);
+        expect(res.body.error.message).to.eq(
+          "INVALID_MFA_ID : Must specify non-null mfaEnrollmentId on update."
+        );
+      });
+  });
+
+  it("should error if mfaEnrollmentId is in use by another user", async () => {
```
yuchenshi, line 682: see discussion

src/test/emulators/auth/signUp.spec.ts
```diff
@@ -395,4 +396,94 @@ describeAuthEmulator("accounts:signUp", ({ authApi }) => {
...
+    const mfaInfo = { displayName: "Cell Phone", phoneInfo: TEST_PHONE_NUMBER };
+    const user = { email: "alice@example.com", password: "notasecret", mfaInfo: [mfaInfo] };
+    const localId = await registerMfaUser(authApi(), user).then(({ localId }) => {
+      expect(localId).to.be.a("string").and.not.empty;
+      return localId;
+    });
+
+    const info = await getAccountInfoByLocalId(authApi(), localId);
+    expect(info.mfaInfo).to.have.length(1);
+    const savedMfaInfo = info.mfaInfo?.pop();
```
yuchenshi, line 410:
```suggestion
    const savedMfaInfo = info.mfaInfo![0];
```

src/test/emulators/auth/signUp.spec.ts
```diff
@@ -395,4 +396,94 @@ describeAuthEmulator("accounts:signUp", ({ authApi }) => {
...
+    };
+    const localId = await registerMfaUser(authApi(), user).then(({ localId }) => {
+      expect(localId).to.be.a("string").and.not.empty;
+      return localId;
+    });
+
+    const info = await getAccountInfoByLocalId(authApi(), localId);
+    expect(info.mfaInfo).to.have.length(3);
+    for (const savedMfaInfo of info.mfaInfo ?? []) {
+      expect(savedMfaInfo).to.include(mfaInfo);
```
yuchenshi, line 432: This already covers `phoneInfo` and `displayName` -- no need for the two expects below

src/test/emulators/auth/signUp.spec.ts
```diff
@@ -395,4 +396,94 @@ describeAuthEmulator("accounts:signUp", ({ authApi }) => {
...
+      expect(savedMfaInfo?.phoneInfo).to.equal(mfaInfo.phoneInfo);
+      expect(savedMfaInfo?.displayName).to.equal(mfaInfo.displayName);
+      expect(savedMfaInfo?.mfaEnrollmentId).to.be.a("string").and.not.empty;
+    }
+    // each factor should have a unique id...
+    expect(info.mfaInfo![0].mfaEnrollmentId).not.to.eq(info.mfaInfo![1].mfaEnrollmentId);
```
yuchenshi, line 438: I think we can just go with `expect(new Set(info.mfaInfo!.map(info => info.mfaEnrollmentId))).to.have.length(3)`.

src/emulator/auth/operations.ts
```diff
@@ -166,6 +166,31 @@ function signUp(
     updates.passwordUpdatedAt = Date.now();
     updates.validSince = toUnixTimestamp(new Date()).toString();
   }
+  if (reqBody.mfaInfo) {
+    const mfaInfo = [];
+    for (const factor of reqBody.mfaInfo) {
+      // the node SDK validation asserts the absence of a uid on create requests prior to submitting any API requests.
```
yuchenshi, line 172: I think production just simply ignores the `mfaEnrollmentId` field and let's just follow that. We can simply just write `mfaInfo.push({displayName: factor.displayName, phoneInfo: factor.phoneInfo, mfaEnrollmentId})` below.

src/emulator/auth/state.ts
yuchenshi: > What is the exact error code when adding new factor but mfaEnrollmentId isn't unique?
tested this for you :) see comment inline

>   My assumption — currently untested — is that uniqueness for the ID is enforced across all accounts. 
This doesn't seem aligned with my quick glance at the code. Please make sure to double check this. It's very counterintuitive that one user will need to avoid duplicates with all other users and the error code may leak information.

src/emulator/auth/state.ts
wokkaflokka: I agree with the recommendation -- the thought had previously crossed my mind, and frankly, this is around primarily because I had based my initial implementation off of related code for managing provider id's.

yuchenshi (commented):
This is looking great and I've left a few comments inline. The big blockers are the uniqueness checks and consistency for the maps in state.

And as of this PR, `mfaInfo` will be lost when you export and re-import users. The scope of the PR is already quite big and I think we should leave a TODO and address it in a follow-up PR.

wokkaflokka:
@yuchenshi once again, thanks for the thorough review. I'll have another pass this evening or tomorrow morning addressing the points you raise. To summarize what I gleaned from review/what I intend to change in my follow-up:

1. Fix the uniqueness behavior to require uniqueness per-user rather than across all users.
2. General updates to validation / internal procedures (as directed by comments).
3. Improvements to style and implementation of some test code (as directed by comments).
4. Add a TODO for mfa behavior related import/export flows. I may file a github issue for the behavior, as well.

wokkaflokka:
@yuchenshi -- well, overpromised and underdelivered, but I've completed another pass of this PR. Changes are in two commits.

In the first [commit](https://github.com/firebase/firebase-tools/pull/3173/commits/0c592cfa78f2e248877e737f5ada5f248bbbbb86), I implemented the points 1-4 I mentioned I would previously fix. One of the benefits of this change is that I think we achieved a vastly simplified state management for MFA data.

In a final testing pass, I spent a little bit of time testing against my real project to see how create/update work when duplicate enrollment ID's and duplicate phone numbers are specified. Based on the behaviors I observed, I implemented some changes in a [second commit](https://github.com/firebase/firebase-tools/pull/3173/commits/d77f6f5e8878b3dd21d26c6aaafecb339d0f574b). These behaviors match the SDK for the following:

1. If duplicate phone numbers are specified on create, only a single factor is created per duplicated phone number.
2. If duplicate phone numbers are specified on update, only a single factor is created per duplicated phone number.
3. MFA Enrollment ID is completely ignored on create (previously implemented).
4. On update, a duplicate MFA Enrollment ID is (a) ignored if it is provided on a duplicated phone number (ignored because the duplicated phone number is ignored), and (b) raises an error if two different phone numbers specify the same MFA enrollment ID.

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -419,45 +420,41 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
-    const savedMfaInfo = info.mfaInfo![0];
+    const user = { email: "bob@example.com", password: "notasecret", mfaInfo: [TEST_MFA_INFO] };
+    const { localId, idToken } = await registerUser(authApi(), user);
+    const savedUserInfo = await getAccountInfoByIdToken(authApi(), idToken);
+    expect(savedUserInfo.mfaInfo).to.have.length(1);
+    const savedMfaInfo = savedUserInfo.mfaInfo![0];
     const secondMfaFactor = {
       displayName: "Second MFA Factor",
-      phoneInfo: TEST_PHONE_NUMBER,
+      phoneInfo: "+12813308004",
```
yuchenshi, line 457: I appreciate the reference, but let's not put Mike's number in the source code. If you need a different phone number, please create a constant like `TEST_PHONE_NUMBER_2` and stick to the fictional blocks (e.g. the next number after `TEST_PHONE_NUMBER`). I think we'd need three or four of these.

src/emulator/auth/operations.ts
```diff
@@ -1838,6 +1811,47 @@ function validateCustomClaims(claims: unknown): asserts claims is Record<string,
...
+  throw new InternalError(
+    "INTERNAL_ERROR : Failed to generate a random ID after 10 attempts",
+    "INTERNAL"
+  );
+}
+
+function getMfaEnrollmentsFromRequest(
+  state: ProjectState,
+  request: MfaEnrollments,
+  getEnrollmentId: (enrollment: MfaEnrollment, enrollmentIds: Set<string>) => string
```
yuchenshi, line 1834: nit: I'd actually suggest just `options: {generateEnrollmentIds: boolean}` instead. Less code, less abstractions and more readable.

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -472,65 +469,53 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
-    mfaInfo.displayName = "New Display Name";
-    mfaInfo.phoneInfo = "+15555550101";
+    const user = { email: "bob@example.com", password: "notasecret", mfaInfo: [TEST_MFA_INFO] };
+    const { localId, idToken } = await registerUser(authApi(), user);
+    const savedUserInfo = await getAccountInfoByIdToken(authApi(), idToken);
+    expect(savedUserInfo.mfaInfo).to.have.length(1);
+    const savedMfaInfo = savedUserInfo.mfaInfo![0];
+    expect(savedMfaInfo?.mfaEnrollmentId).to.be.a("string").and.not.empty;
+    savedMfaInfo.displayName = "New Display Name";
+    savedMfaInfo.phoneInfo = "+15555550101";
```
yuchenshi, line 491: Similarly this could use a constant like `TEST_PHONE_NUMBER_2`

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -539,37 +524,34 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
   it("should overwrite existing MFA info", async () => {
-    const mfaInfo: any = {
-      displayName: "Cell Phone",
-      phoneInfo: TEST_PHONE_NUMBER,
-    };
     const user = {
       email: "bob@example.com",
       password: "notasecret",
-      mfaInfo: [mfaInfo, mfaInfo, mfaInfo],
+      mfaInfo: [TEST_MFA_INFO, { ...TEST_MFA_INFO, phoneInfo: "+12813308004" }],
```
yuchenshi, line 545: Same

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -539,37 +524,34 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
-    expect(info.mfaInfo).to.have.length(3);
+    const { localId, idToken } = await registerUser(authApi(), user);
+    const savedUserInfo = await getAccountInfoByIdToken(authApi(), idToken);
+    expect(savedUserInfo.mfaInfo).to.have.length(2);
+    const oldEnrollmentIds = savedUserInfo.mfaInfo!.map((_) => _.mfaEnrollmentId);
 
     const newMfaInfo = {
       displayName: "New New",
-      phoneInfo: "+12813308004",
+      phoneInfo: "+12813308005",
```
yuchenshi, line 554: use a test phone number please

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -648,104 +620,226 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
+            {
+              ...TEST_MFA_INFO,
+              mfaEnrollmentId: randomId(28),
+            },
+            {
+              ...TEST_MFA_INFO,
+              mfaEnrollmentId: randomId(28),
+            },
+            {
+              phoneInfo: "+12813308004",
```
yuchenshi, line 743: use a test phone number

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -648,104 +620,226 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
+              mfaEnrollmentId: randomId(28),
+            },
+            {
+              phoneInfo: "+12813308004",
```
yuchenshi, line 747: use a test phone number

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -648,104 +620,226 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
+      })
+      .then((res) => expectStatusCode(200, res));
+
+    const fourthUpdate = await getAccountInfoByIdToken(authApi(), idToken);
+    expect(fourthUpdate.mfaInfo).to.have.length(2);
+    for (const mfaInfo of fourthUpdate.mfaInfo!) {
+      if (mfaInfo.phoneInfo === TEST_MFA_INFO.phoneInfo) {
+        expect(mfaInfo).to.include(TEST_MFA_INFO);
+      } else {
+        expect(mfaInfo.phoneInfo).to.eq("+12813308004");
```
yuchenshi, line 761: test ... phone ... number ... Sorry, but I'm _calling_ it out. :wink:

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -648,104 +620,226 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
+      .send({
+        localId,
+        mfa: {
+          enrollments: [
+            {
+              ...TEST_MFA_INFO,
+              mfaEnrollmentId,
+            },
+            {
+              phoneInfo: "+12813308004",
```
yuchenshi, line 787: Like it's 2005

src/test/emulators/auth/setAccountInfo.spec.ts
```diff
@@ -648,104 +620,226 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
...
+        expect(mfaInfo).to.include(TEST_MFA_INFO);
+      } else {
+        expect(mfaInfo.phoneInfo).to.eq("+12813308004");
+      }
+      expect(mfaInfo.mfaEnrollmentId).to.be.a("string").and.not.empty;
+      expect(mfaInfo.mfaEnrollmentId).not.to.eq(mfaEnrollmentId);
+    }
+  });
+
+  it("should error if MFA Enrollment ID is Duplicated For Different Phone Numbers", async () => {
```
yuchenshi, line 768:
```suggestion
  it("should error if MFA Enrollment ID is duplicated For different phone numbers", async () => {
```

src/test/emulators/auth/signUp.spec.ts
```diff
@@ -398,67 +400,84 @@ describeAuthEmulator("accounts:signUp", ({ authApi }) => {
...
   });
 
-  it("should create new account with multiple second factors", async () => {
-    const mfaInfo = { displayName: "Cell Phone", phoneInfo: TEST_PHONE_NUMBER };
+  it("should create new account with two MFA factors", async () => {
     const user = {
       email: "alice@example.com",
       password: "notasecret",
-      mfaInfo: [mfaInfo, mfaInfo, mfaInfo],
+      mfaInfo: [TEST_MFA_INFO, { ...TEST_MFA_INFO, phoneInfo: "+12813308004" }],
```
yuchenshi, line 416: I wish we had more test phone numbers _Back Then_.

src/test/emulators/auth/signUp.spec.ts
```diff
@@ -398,67 +400,84 @@ describeAuthEmulator("accounts:signUp", ({ authApi }) => {
...
-    expect(info.mfaInfo).to.have.length(3);
-    for (const savedMfaInfo of info.mfaInfo ?? []) {
-      expect(savedMfaInfo).to.include(mfaInfo);
-      expect(savedMfaInfo?.phoneInfo).to.equal(mfaInfo.phoneInfo);
-      expect(savedMfaInfo?.displayName).to.equal(mfaInfo.displayName);
-      expect(savedMfaInfo?.mfaEnrollmentId).to.be.a("string").and.not.empty;
+    expect(info.mfaInfo).to.have.length(2);
+    for (const savedMfaInfo of info.mfaInfo!) {
+      if (savedMfaInfo.phoneInfo !== TEST_MFA_INFO.phoneInfo) {
+        expect(savedMfaInfo.phoneInfo).to.eq("+12813308004");
```
yuchenshi, line 423: If I were you I'd probably do a search and replace

src/test/emulators/auth/signUp.spec.ts
```diff
@@ -398,67 +400,84 @@ describeAuthEmulator("accounts:signUp", ({ authApi }) => {
...
+    expect(aliceInfo.mfaInfo![0].mfaEnrollmentId).to.be.a("string").and.not.empty;
+
+    const bob = {
+      email: "bob@example.com",
+      password: "notasecret",
+      mfaInfo: [
+        TEST_MFA_INFO,
+        TEST_MFA_INFO,
+        TEST_MFA_INFO,
+        { ...TEST_MFA_INFO, phoneInfo: "+12813308004" },
```
yuchenshi, line 450: 25,000 calls per day

src/test/emulators/auth/signUp.spec.ts
```diff
@@ -398,67 +400,84 @@ describeAuthEmulator("accounts:signUp", ({ authApi }) => {
...
+      ],
+    };
+    const { localId: bobLocalId } = await registerUser(authApi(), bob);
+    const bobInfo = await getAccountInfoByLocalId(authApi(), bobLocalId);
+    expect(bobInfo.mfaInfo).to.have.length(2);
+    for (const savedMfaInfo of bobInfo.mfaInfo!) {
+      if (savedMfaInfo.phoneInfo !== TEST_MFA_INFO.phoneInfo) {
+        expect(savedMfaInfo.phoneInfo).to.eq("+12813308004");
```
yuchenshi, line 458: Are we stress testing the phone carrier?

src/test/emulators/auth/signUp.spec.ts
```diff
@@ -470,20 +489,29 @@ describeAuthEmulator("accounts:signUp", ({ authApi }) => {
       });
   });
 
-  it("should error if a multi factor enrollment ID is provided on create", async () => {
-    const mfaInfo = {
-      displayName: "Cell Phone",
-      phoneInfo: TEST_PHONE_NUMBER,
-      mfaEnrollmentId: randomId(28),
+  it("should ignore if multi factor enrollment ID is specified on create", async () => {
+    const mfaEnrollmentId1 = randomId(28);
```
yuchenshi, line 493: Let's hard code some IDs like `thisShouldBeIgnored1` -- more readable, and the test won't randomly fail if we had really really bad (good?) luck with the RNG.

yuchenshi (commented):
I think we're pretty close here. The only real blocker is using some constants for phone numbers instead of real phone numbers.

wokkaflokka:
@yuchenshi all very reasonable. Can't blame ya!

1. Test numbers moved to constants and changed.
2. Replace function parameter with options (great recommendation).
3. Remove explicit test dependency on RNG (another great rec).

Thanks very much for all your time and help so far. Much appreciated.

yuchenshi:
@wokkaflokka  One last thing -- I meant that we should use `"thisShouldBeIgnored1` for MFA IDs that are ignored (during sign up). For the other MFA enrollment IDs, let's use some other name such as `enrollmentId1` etc. Once that is done and the branch is up to date, I'll do the changelogs and merge.

wokkaflokka:
@yuchenshi roger that, pushed another one.

yuchenshi (approved):

yuchenshi:
@wokkaflokka Merged! Thanks a ton for the PR and I've filed #3221 to track the remaining work.

samtstern:
@wokkaflokka thank you so much for this contribution and thanks @yuchenshi for the super thorough reviews!

PR #3181, opened 2021-03-02
Added validation for protected function names on package.json

rcopstein:
### Description

Whenever the `package.json` file that accompanies the Cloud Functions project contains a script named "prepare", a deploy will fail with no proper error message. I believe this happens because the `npm ci` is run, it attempt to run a script named "prepare" that conflicts with the user-defined one.

To avoid this I have added an extra step on the `package.json` verification prior to deploy to ensure that any scripts with _protected_ names (for now, just "prepare") are identified and a proper error is thrown.

### Scenarios Tested

As far as I can see, the only test that really made sense here was to check for a failure if the `package.json` file contained a script named "prepare". Which passes properly. This will have to be extended for other protected (conflicting?) names.

### Sample Commands

N/A

### DIFF
--- a/src/deploy/functions/validate.ts
+++ b/src/deploy/functions/validate.ts
@@ -134,9 +134,9 @@ function assertFunctionsSourcePresent(data: any, sourceDir: string, projectDir: 
  */
 function checkForProtectedNpmScriptNames(data: any): void {
   const protectedScriptNames = ["prepare"];
-  if (Object.prototype.hasOwnProperty.call(data, "scripts")) {
+  if (data?.scripts) {
     for (const scriptName of protectedScriptNames) {
-      if (Object.prototype.hasOwnProperty.call(data.scripts, scriptName)) {
+      if (data.scripts.[scriptName]) {
         const msg = `package.json contains script with protected name ${scriptName}`;
         throw new FirebaseError(msg);
       }
--- a/src/deploy/functions/validate.ts
+++ b/src/deploy/functions/validate.ts
@@ -72,6 +72,7 @@ export function packageJsonIsValid(
     data = cjson.load(packageJsonFile);
     logger.debug("> [functions] package.json contents:", JSON.stringify(data, null, 2));
     assertFunctionsSourcePresent(data, sourceDir, projectDir);
+    checkForProtectedNpmScriptNames(data);
   } catch (e) {
     const msg = `There was an error reading ${sourceDirName}${path.sep}package.json:\n\n ${e.message}`;
     throw new FirebaseError(msg);
@@ -123,6 +124,23 @@ function assertFunctionsSourcePresent(data: any, sourceDir: string, projectDir: 
       indexJsFile
     )} does not exist, can't deploy Cloud Functions`;
     throw new FirebaseError(msg);
+  }
+}
+
+/**
+ * Asserts that package.json does not include any scripts with invalid names
+ * @param data Object representing package.json file.
+ * @throws { FirebaseError } package.json must not contain scripts with protected names
+ */
+function checkForProtectedNpmScriptNames(data: any): void {
+  const protectedScriptNames = ["prepare"];
+  if (Object.prototype.hasOwnProperty.call(data, "scripts")) {
+    for (const scriptName of protectedScriptNames) {
+      if (Object.prototype.hasOwnProperty.call(data.scripts, scriptName)) {
+        const msg = `package.json contains script with protected name ${scriptName}`;
+        throw new FirebaseError(msg);
+      }
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


rcopstein:
@googlebot I signed it!

joehan:
Interesting find @rcopstein, and thanks for the PR. We have some code in our deploy process that defines a 'prepare', 'release', and 'deploy' stage that may be causing the behavior you saw. I'm a bit confused as to why an npm script would collide with these, tho. Could you test to see if you get the same behavior when you add a script named 'release' or 'deploy', and add those to the list of protected names?

src/deploy/functions/validate.ts
```diff
@@ -126,6 +127,23 @@ function assertFunctionsSourcePresent(data: any, sourceDir: string, projectDir:
...
 }
 
+/**
+ * Asserts that package.json does not include any scripts with invalid names
+ * @param data Object representing package.json file.
+ * @throws { FirebaseError } package.json must not contain scripts with protected names
+ */
+function checkForProtectedNpmScriptNames(data: any): void {
+  const protectedScriptNames = ["prepare"];
+  if (Object.prototype.hasOwnProperty.call(data, "scripts")) {
```
joehan, line 137: While this is technically safer, I prefer this as the simpler to read:
```suggestion
  if (data?.scripts)) {
```

joehan: Furthermore, if we do this, we can simplify this whole chunk even further, to something like:
```suggestion
    for (const scriptName of protectedScriptNames) {
      if (data?.scripts?.[scriptName])) {
        throw new FirebaseError(`package.json contains script with protected name ${scriptName}`);
      }
    }
```

joehan (approved):
Pretty much LGTM with some style suggestions. I'd like to run this by @inlined as well first though - he might have some more insight on the problem you found.

src/deploy/functions/validate.ts
rcopstein: @joehan Thanks for the suggestions. I agree with you that stylistically you approach is better. It frames the fact that _data_ can be null/undefined better than my approach. 

The only downside I see is the fact that we are testing for the existence of _data_ and _data.scripts_ with every iteration of the loop. Right now this makes no difference since we only have one "protected word" (I'm still unsure about the term haha). If we do find other script names that conflict with the deploy we would be needlessly running checks for all of them even if _data_ or _data.scripts_ is not available.

I guess it boils down to project preference...

src/deploy/functions/validate.ts
```diff
@@ -126,6 +127,23 @@ function assertFunctionsSourcePresent(data: any, sourceDir: string, projectDir:
...
+    for (const scriptName of protectedScriptNames) {
+      if (Object.prototype.hasOwnProperty.call(data.scripts, scriptName)) {
```
rcopstein, line 139:
```suggestion
  if (data?.scripts) {
    for (const scriptName of protectedScriptNames) {
      if (data.scripts.[scriptName]) {
```

@joehan Just getting your first suggestion but keeping the "fail fast" approach of that first conditional so it doesn't start the for loop unnecessarily. What do you think?

joehan: Seems reasonable to me

inlined:
`prepare` is a [well known script](https://docs.npmjs.com/cli/v7/using-npm/scripts), which is called many times, including during `npm install` or `npm pack` (I think we're running into an issue with the former).  The error you're seeing is coming from Cloud Build, which I believe is running `npm install --production <archive created by firebase-tools>`.

I was able to successfully run simple shell scripts from the `prepare` script. I was also able to run `tsc` from the script _as long as I changed typescript from a devDependency to a normal dependency_ . Is there another command that's not working? As far as I can tell this is working as expected, even if it's easy to forget that the production environment doesn't include our dev dependencies.

rcopstein:
@inlined, so you're saying that the issue is not that the `prepare` script is conflicting with the one in `npm install`, but rather it is failing because it uses devDependencies that are not there to begin with? Interesting...

It might be out of the scope of this project, but I would say that a warning would go a long way in helping people that got stuck because of this problem (myself included haha). Do you think we could add a warning line to the `firebase deploy` output if it finds well know scripts in `package.json`?

inlined:
To be clear, there is no "conflicting" script that I can see here. There's simply the well-documented behavior that any "prepare" script will be called at certain points of the node lifecycle. I'm not sure what we should do to "fix" this since it's (reasonably) well known and documented behavior. A warning that someone has chosen to use an advanced feature seems like a very slippery slope.

I'm honestly not sure what the right solution is other than search engine optimization for the error message.

rcopstein:
You have a very compelling point, indeed. This is looking like an issue that would have to be solved in the back end with the message that they decide to return on error.

I think there's nothing more to discuss here. Thanks for looking into this :)

PR #3186, opened 2021-03-04
Fix database:remove bug

fredzqm:
In `listRemove.ts`, `const paths = Object.keys(res.body);` fails if `res.body` is null.
This is causing errors when listing a path already deleted.

This bug was introduced in https://github.com/firebase/firebase-tools/pull/2828.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -89,15 +89,11 @@ export function getDatabaseViewDataUrl(
   pathname: string
 ): string {
   const urlObj = new url.URL(origin);
-  if (
-    urlObj.hostname.includes("firebaseio.com") ||
-    urlObj.hostname.includes("firebasedatabase.app")
-  ) {
+  if (urlObj.hostname.includes("firebase")) {
     return consoleUrl(project, `/database/${namespace}/data${pathname}`);
-  } else {
-    // TODO(samstern): View in Emulator UI
-    return getDatabaseUrl(origin, namespace, pathname + ".json");
   }
+  // TODO(samstern): View in Emulator UI
+  return getDatabaseUrl(origin, namespace, pathname + ".json");
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -89,15 +89,11 @@ export function getDatabaseViewDataUrl(
   pathname: string
 ): string {
   const urlObj = new url.URL(origin);
-  if (
-    urlObj.hostname.includes("firebaseio.com") ||
-    urlObj.hostname.includes("firebasedatabase.app")
-  ) {
+  if (urlObj.hostname.includes("firebase")) {
```
fredzqm, line 92: So that `FIREBASE_REALTIME_URL=https://firebaseio-staging.com` will work.

yuchenshi: Let's just list all three full domain names to be on the safe side. `myfirebaseapp.example.com` should be filtered out.

src/utils.ts
```diff
@@ -110,15 +106,11 @@ export function addDatabaseNamespace(origin: string, namespace: string): string
   if (urlObj.hostname.includes(namespace)) {
     return urlObj.href;
   }
-  if (
-    urlObj.hostname.includes("firebaseio.com") ||
-    urlObj.hostname.includes("firebasedatabase.app")
-  ) {
+  if (urlObj.hostname.includes("firebase")) {
```
yuchenshi, line 109: same

yuchenshi (approved):

src/utils.ts
fredzqm: Is there a way to detect emulator specific url instead? It might be cleaner.
Are they always "localhost"?

yuchenshi: No, in theory the emulator can listen on all sorts of hosts and ports. Common configurations include `127.0.0.1`, `localhost`, but it can be changed by the user to LAN or even public IP addresses and domains, and I've seen those in the wild. We had a similar discussion re: RTDB emulator and the conclusion so far is not to try and detect emulator URLs

fredzqm: SG. I will just check for `firebaseio` or `firebasedatabase` if u think `firebase` is too general.

Issue #2887, opened 2020-12-02
photoUrl vs photoURL


chrismwendt:
### [REQUIRED] Step 2: Describe your environment

  * Operating System version: macOS
  * Firebase SDK version: CLI 8.16.2, firebase-admin 9.4.1
  * Firebase Product: auth
  * Node.js version: 12
  * NPM version: 6

### [REQUIRED] Step 3: Describe the problem

The type definition says `photoURL`:

https://github.com/firebase/firebase-admin-node/blob/a98a3cc275f9fee0474ff1905d2217a600d5f57e/src/auth/index.ts#L101

But when running in the emulator, each `user.providerData` has `photoUrl` (different capitalization).

#### Steps to reproduce:

- Register an `onCreate` function that logs `user.
- Create a user
- Notice it prints `photoUrl` but the type definition says `photoURL`

#### Relevant Code:

```js
export const onUserCreated = functions.auth.user().onCreate(user => {
  console.log(user.providerData[0])
})
```

Relevant lines from the CLI:

https://github.com/firebase/firebase-tools/blob/08d57ae172297532e7728615e7ab281a08cae2bf/src/emulator/auth/state.ts#L135

https://github.com/firebase/firebase-tools/blob/08d57ae172297532e7728615e7ab281a08cae2bf/src/emulator/auth/cloudFunctions.ts#L70

yuchenshi:
Great catch! I looked a bit into this, and the functions payload should indeed be `photoURL` (caps) as you said. Feel free to open a PR to correct it in `src/emulator/auth/cloudFunctions.ts` -- bonus points for tests and a proper TypeScript interface (instead of `object[]` right now) too!

For background, the reason we had this bug is that the REST API uses `photoUrl` (snake case) while the function trigger is different (`photoURL`) and we've overlooked this when copying `providerData`.

haythamhakim:
any updates :")?

haythamhakim:
@yuchenshi any updates :")?

swseverance:
I believe this issue can be closed.  Consider the following `onCreate()` function:
```typescript
import * as logger from "firebase-functions/logger";
import * as v1Auth from "firebase-functions/v1/auth";

export const onCreate = v1Auth.user().onCreate((user) => {
  logger.info(JSON.stringify(user));
});
```

If you drill into the type definition for user you will see:
```typescript
export declare class UserRecord {
    ...

   readonly photoURL?: string;

   ...

[...]

PR #3194, opened 2021-03-06
Change Uppercase to lowercase ready #2887

ucusita:


ucusita:
@googlebot I signed it!

yuchenshi:
`photoURL` is supposed to be spelled and capitalized like that, with `URL` all uppercase. See: https://firebase.google.com/docs/reference/admin/node/admin.auth.UserRecord#photourl

The Auth Emulator needs to stay consistent with production behavior here.

PR #3203, opened 2021-03-11
Some UX polish for enabling/disabling functions

samtstern:
### Description

Two changes:

- Prevents log spam about "disabling" functions that are already disabled (old trigger generations) when functions are repeatedly enabled/disabled.
- Makes disabling wait for the work queue to flush so that no functions appear to run after the disabled message returns success.

### Scenarios Tested

Tested both of these with the emulators codelab. 

### Sample Commands

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Fixes extra logging and race conditions when disabling emulated background functions.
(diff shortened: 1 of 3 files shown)


yuchenshi (approved):

yuchenshi:
CHANGELOG please

PR #3220, opened 2021-03-18
Adding tsconfig.dev.json to extensions templates

joehan:
### Description
Adding tsconfig.dev.json to the extensions/init templates folder.

We might want to make a 'shared' folder for some of the files that are common between extensions and functions, but I'd like to just get this fixed for now.

### Scenarios Tested
`firebase ext:dev:init`, choose Typescript:
Screen Shot 2021-03-18 at 1 01 41 PM


bkendall (approved):
+1 for consolidating templates into one directory, especially if they're identical (or adding a placeholder for changes - you know, like a template hah)

PR #3223, opened 2021-03-19
Fix port conflicts with functions:shell

### LINKED ISSUE #3210: Can't connect to functions shell when hosting emulator is running

**firebase-tools:** 9.6.1

**Platform:** macOS

### [REQUIRED] Test case
A project with all the emulators enabled (hosting, pubsub, functions, database, ui) - a good example is https://github.com/Hacksore/firebase-shell-bug 

See: https://github.com/Hacksore/firebase-shell-bug/blob/master/README.md

I should be able to get the functions shell REPL while the hosting emulator is running.

When all emulators are running via `firebase emulators:start` you will get a connection error when attempting to `firebase functions:shell`

```log
Error: Port 5000 is not open on localhost, could not start functions emulator.
```

samtstern:
### Description

Fixes #3210

Also I converted the file to `.ts` while I was in there

### Scenarios Tested

- Picking up the port from `firebase.json`
- Detecting another running emulator and printing a warning
- Using the `--port` flag

### Sample Commands

N/A


src/functionsShellCommandAction.ts
```diff
@@ -61,11 +83,11 @@ module.exports = async function (options) {
         process.exit();
       }
 
-      var initializeContext = function (context) {
-        _.forEach(emulator.triggers, function (trigger) {
+      const initializeContext = (context: any) => {
+        _.forEach(emulator.triggers, (trigger) => {
```
joehan, line 87: Optional - it would be pretty easy to refactor this chunk to use plain JS instead of lodash, and then we're one step closer to removing the dependency for good

joehan (approved):
LGTM after a few tiny cleanups + lodash removal.

src/functionsShellCommandAction.ts
samtstern: I got rid of `_.forEach` and `_.includes` but `_.set` with dot-separated path notation is a bit tricky so I left that. We'll probably need to create our own `utils.set` to replace that throughout the codebase and that seems out of scope for this PR.

PR #3225, opened 2021-03-22
Configurable project id for the authentication emulator

mvgijssel:
### Description

When I'm developing on a Firebase project I have two separate projects for development and testing, for example "project-dev" and "project-test". The reason for this is test isolation, I don't want work from the development environment affecting the outcome of a test.

All the Firebase emulators seem to use the `projectId` that's passed into the `firebase.initializeApp(...)` method except for the authentication emulator. I noticed this when I was writing tests using Cypress and trying to create accounts for testing the login flow, because all users created would show up in the "project-dev" accounts instead of the "project-test" accounts.

Looking at the implementation I found that this POST request is made when signing up a new user containing the **key** parameter

```
http://localhost:9099/www.googleapis.com/identitytoolkit/v3/relyingparty/signupNewUser?key=fake-api-key
```

That key is passed into the authentication emulator and ignored (as far as I can see), so my proposal is to use the `key` parameter as a way to determine the project id. This is also in line how Firebase works in production according to this article https://firebase.google.com/docs/projects/api-keys.

### Scenarios Tested

I've modified my local version of the `firebase-tools` project and it seems to behave as expected, creating users into different projects.

### DIFF
--- a/src/emulator/auth/server.ts
+++ b/src/emulator/auth/server.ts
@@ -321,9 +321,7 @@ export async function createApp(
   return app;
 
   function getProjectIdByApiKey(apiKey: string): string {
-    /* unused */ apiKey;
-    // We treat any non-empty string as a valid key for the default projectId.
-    return defaultProjectId;
+    return apiKey;
   }
 
   function getProjectStateById(projectId: string): ProjectState {


mvgijssel:
@googlebot I signed it!

yuchenshi:
I agree with the idea overall, but I think this will break existing developers who expects a dummy API key to "just work" and act as the default project. I'd suggest using a prefix like `emulator-project:foo-bar` (where `foo-bar` is the desired project ID). Thoughts? (@mvgijssel @samtstern )

samtstern:
@yuchenshi just to confirm, in production is the API key used to look up the project ID?  If so then I agree adding this functionality to the emulators makes sense and I like your idea of some sort of `<prefix>:<projectId>` system.

mvgijssel:
For backwards compatibility reasons I think the prefix solution makes sense!

@samtstern seems the API key is used for routing projects. From the [docs](https://firebase.google.com/docs/projects/api-keys)

> An API key is a unique string that's used to route` requests to your Firebase project when interacting with Firebase and Google services.

yuchenshi:
As correctly observed by @mvgijssel, the API key is indeed used for identifying the project in production APIs. Let's add a prefix and a few tests and we can make this feature happen 🚢

yuchenshi:
Turned out that `:` is probably not the best way for this, see https://github.com/firebase/firebase-js-sdk/pull/4706. I'm open to alternative proposals.

PR #3243, opened 2021-03-30
Add support for maxBackoff to throttler

joehan:
### Description
Adds support for setting a max value for backoff in throttler.ts, and set the default maxBackoff to 60 seconds.

While working on retrying quota errors in Cloud Functions deployment, I noticed that the throttler will exponentially backoff to an arbitrary backoff time. For tasks that took many retries, this caused huge backoffs and severe delays

### Scenarios Tested
I manually tested this by adding a log statement that printed out the timeToWait at each attempt of a task, and  confirming that it  never exceeded maxBackoff. I also added some unit tests to ensure that backoff is waiting the appropriate amount of time.

### DIFF
--- a/src/test/throttler/throttler.spec.ts
+++ b/src/test/throttler/throttler.spec.ts
@@ -3,7 +3,7 @@ import { expect } from "chai";
 
 import Queue from "../../throttler/queue";
 import Stack from "../../throttler/stack";
-import { Throttler, ThrottlerOptions } from "../../throttler/throttler";
+import { Throttler, ThrottlerOptions, timeToWait } from "../../throttler/throttler";
 import TaskError from "../../throttler/errors/task-error";
 import TimeoutError from "../../throttler/errors/timeout-error";
 import RetriesExhaustedError from "../../throttler/errors/retries-exhausted-error";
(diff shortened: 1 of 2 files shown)


src/throttler/throttler.ts
```diff
@@ -258,12 +261,12 @@ export abstract class Throttler<T, R> {
     const t0 = Date.now();
     let result;
     try {
-      result = await this.handler(taskData.task);
+      if ((this.name = "cloudFunctionsDeployment")) result = await this.handler(taskData.task);
```
bkendall, line 264: Why does there need to be a special case based on the name of a task?

bkendall (changes requested):

bkendall (approved):

mbleigh (approved):

src/test/throttler/throttler.spec.ts
```diff
@@ -402,6 +402,28 @@ describe("Throttler", () => {
...
 });
 
+describe("timeToWait", () => {
+  it("Should wait the base delay on the first attempt", () => {
+    const retryCount = 0;
+    const delay = 100;
+    const maxDelay = 1000;
+    expect(timeToWait(retryCount, delay, maxDelay)).to.equal(delay);
+  });
+  it("Should back off exponentially", () => {
```
bkendall, line 412: please have spaces between the cases (`it` blocks) and don't capitalize "Should")

Good thought to break out these tests though

bkendall (approved):

Issue #1372, opened 2019-06-05
Limit throughput of function deployments


merlinnot:
Google Cloud functions has a [quota](https://cloud.google.com/functions/quotas#rate_limits) on API calls (WRITE), which makes deployments of large number of functions impossible. This can be solved by limiting the number of requests `firebase-tools` makes per 100 seconds by queueing and performing operations within this limitation.

~~I'd be willing to work on adding support for this scenario, please let me know if such a PR would be accepted.~~ Update: don't have capacity for this at the moment.

ryanpbrewster:
I have encountered the same issue and considered the same fix.

thechenky:
Hah, we have an internal bug for specifically this! Would love your help in implementing :) thanks for offering!

Internal issue reference: 126680733

merlinnot:
I'm still going to implement it, I started with migrating some of the files I'll need to touch from JavaScript to TypeScript.

Today I received some more information from Google Cloud Support which I might end up incorporating in the PR for this issue:

> The overall problem is that the deployments are very bursty (like 100+ in a single second and then nothing for 30 minutes). To make it scalable and sustainable we'd suggest changing the deployment automation so that it does one function deployment every few seconds or so. It will allow for O(10000) deployments per day. Once it's not enough there's a possibility to use more granular deployment unit by deploying several functions as separate endpoints in a single App Engine or Cloud Run service.

Request limit for one of my projects was actually reduced to 50 (the default is 80):

> We've reduced GCF Write (i.e. deployment) quotas from 100 to 50 per 100s for project ****. The customer recently asked for the quota increase from 80 to 100 per 100s in sfstory/****. The increase has been granted but since then we observed some problems caused by bursts of deployments traffic. E.g. on Jul 21st 04:17 PDT there were 115 requests in a single second, most of which failed due to overload. Hope that reducing the quota will allow traffic to level out and be eventually more successful. 

My initial intention was to fire 80 requests and then fire another batch after 100 seconds and so on, with a capped exponential backoff for contention errors. Given the above, I'd go for firing 1 request every 1.25 seconds so it complies with the default limit and the suggestion quoted above.

In the future we might want to think of exposing this value (time between deployments) as a CLI option for occasions like this, but I'll make a separate issue for it once the PR with 1.25 lands in master.

Let me know what you think, if this approach would be ok with you. Maybe you have some other ideas?

merlinnot:
[...]

PR #3246, opened 2021-03-31
Retry 'Quota Exceeded' errors during function deployment

### LINKED ISSUE #2606: Request: Automatically retry failed function deploys that fail due to "You have exceeded your deployment quota"
## Problem & Background

We have a deployment of ~70 functions and deploy them as part of CI at once. Whenever we deploy a random set of functions fails to deploy due to "You have exceeded your deployment quota". This used to block us completely a few weeks ago when there still was a quota on build time. Since this has been moved to Cloud Build as far as I can tell there are no more limits and you simply pay for your build time. Which is awesome, no more blocking the build due to quotas!

Unfortunately there still seems to be some other quotas/rate limits that are causing the deploy to fail. From what I can tell these might be simply too many write requests against the Cloud Functions API (or perhaps cloud build?).   

The error message suggests to deploy with --only - Unfortunately it is not easy for us to split these up into separately deployed functions. It is also impossible to do when a dependency is updated or we change a data model or utility library that is used by different functions. Analyzing which function changed and has to be redeployed is not possible (for us) automatically and has to be done manually. This then brings new pains, where automated deploys after code review become impossible.

Right now this means our pipelines all fail regularily - and we manually retry them until every functions deployed successfully once. 
image

## Suggestion

Since retrying the failed functions right after always works, my suggestion would be to retry failed deploys that hit code 8 here:

https://github.com/firebase/firebase-tools/blob/3994bf7209a71ed5dde100c69ca96dc2d9c3634e/src/deploy/functions/release.js#L97-L105

I suppose this could be implemented here:

https://github.com/firebase/firebase-tools/blob/3994bf7209a71ed5dde100c69ca96dc2d9c3634e/src/deploy/functions/release.js#L526

If this would be a way forward I'm happy to create a PR for this.

joehan:
### Description
Adds logic to retry 'Quota Exceeded' errors during functions deployment. Fixes #2606 and #1372.

I also took this opportunity to deduplicate + test some error handling logic that was repeated across the different tasks by moving it into handler functions for the throttler.

### Scenarios Tested
I tested this by deploying 240 functions across 3 regions, and was able to do so successfully (with quite a few quota error retries).
I also tested out `functions:delete` by successfully deleting a large group of functions.


src/deploy/functions/tasks.ts
```diff
@@ -45,63 +62,60 @@ export function createFunctionTask(
...
   };
+  task.functionName = fn.name;
+  task.operationType = "create";
+  return task;
 }
 
 export function updateFunctionTask(
   params: TaskParams,
   fn: CloudFunctionTrigger,
+  sourceToken?: string,
```
joehan, line 115: I removed this from TaskParams after realizing I had accidentally introduced a bug where a sourceToken from a different  region would be used. I could have fixed this by creating a different copy of TaskParams for every task - however, I think the code is easier to understand overall if we just separate this out.

src/deploy/functions/release.ts
```diff
@@ -94,7 +105,20 @@ export async function release(context: any, options: any, payload: any) {
   // Wait for the first function in each region to be deployed, and all the other calls to be queued,
   // then close the queue.
   // Wait for all of the deployments to complete.
-  await Promise.all(queuePromises);
+  try {
+    await Promise.all(queuePromises);
+  } catch (err) {
+    utils.reject(
```
mbleigh, line 111: Is this error always going to be encountered because of large function batches? Can we detect whether quota errors specifically were the cause of the rejection?

Also wonder about literally forking the error message based on the number of functions deployed...

src/deploy/functions/tasks.ts
```diff
@@ -110,122 +124,92 @@ export function updateFunctionTask(
...
+    const updateRes = await cloudfunctions.updateFunction({
+      projectId: params.projectId,
+      region: helper.getRegion(fn.name),
+      eventType: eventType,
+      functionName: helper.getFunctionId(fn.name),
+      entryPoint: fn.entryPoint,
+      trigger: helper.getFunctionTrigger(fn),
+      labels: Object.assign({}, deploymentTool.labels(), fn.labels),
+      sourceUploadUrl: params.sourceUrl,
+      sourceToken: sourceToken,
```
mbleigh, line 137: So where does this come from now if not the params above?

mbleigh:
Oh also can you add a `CHANGELOG` for this?

src/deploy/functions/tasks.ts
```diff
@@ -22,21 +22,38 @@ const defaultPollerOptions = {
...
+export type OperationType =
+  | "create"
+  | "update"
+  | "delete"
+  | "upsert schedule"
+  | "delete schedule"
+  | "make public";
+
+export interface DeploymentTask {
+  (): Promise<any>;
```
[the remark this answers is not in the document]
joehan: Yeah, this is probably a little too cute, and would be easier to work with if the function was a named member. I'll make the switch.

src/deploy/functions/tasks.ts
joehan: Same place as before (the onPollFn in runRegionalDeployment), I just moved it out of the TaskParams object and passed it into this function separately. 

I found a bug while developing where, because every task was sharing the same TaskParam object, we'd overwrite the sourceToken, and cause some deployments to fail because we sent a sourceToken from a different region.

I could have fixed it by making a new TaskParams for every region, but I figured it was cleaner and less bug prone to separate it - having a mutating object passed around and shared by a bunch of simultaneously running async functions seemed poor.

src/deploy/functions/tasks.ts
```diff
@@ -110,122 +124,92 @@ export function updateFunctionTask(
...
-      }
-      logger.debug(`Scheduler topic ${topicName} not found.`);
+      errorHandler.record(
+        "error",
+        task.functionName,
+        task.operationType,
+        err.original?.message || ""
+      );
     }
+    timer.endTimer(task.functionName);
```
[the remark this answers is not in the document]
joehan: I don't think it matters because the only time we don't hit this line is if we throw a quota error, and the Queue will retry on errors, so endTimer will eventually be called.

src/deploy/functions/tasks.ts
```diff
@@ -270,13 +254,66 @@ export function runRegionalFunctionDeployment(
...
+  task.functionName = fnName;
+  task.operationType = "delete schedule";
+  return task;
+}
+
+export function schedulerDeploymentHandler(
+  errorHandler: ErrorHandler
+): (task: DeploymentTask) => Promise<any | undefined> {
+  return async (task: DeploymentTask) => {
+    let result;
```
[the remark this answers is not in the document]
joehan: Good catch!

src/deploy/functions/release.ts
joehan: This will only ever throw if we exceed the max retries on a task, and we only retry quota errors. 

 I don't see any quotas that we're likely to exceed that aren't related to large function deployments, but it might still be worth forking the error message at 50? functions or so.

src/deploy/functions/release.ts
```diff
@@ -30,15 +30,26 @@ export async function release(context: any, options: any, payload: any) {
...
+
+  // This queue needs to retry quota errors.
+  // The main quotas that can be exceeded are per 1 minute quotas,
+  // so we start with a larger backoff to reduce the liklihood of extra retries.
+  const cloudFunctionsQueue = new Queue<tasks.DeploymentTask, void>({
+    retries: 30,
+    backoff: 20000,
+    concurrency: 40,
+    maxBackoff: 40000,
+    handler: tasks.functionsDeploymentHandler(timer, errorHandler),
```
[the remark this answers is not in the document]
joehan: SchedulerQueue doesn't time its operations because we don't do anything with the results. We could record those times in GA like we do for function deployments, but scheduler deployments are an order of magnitude quicker than function deployments so it's not very interesting. 

I like the idea of moving some of this functionality into Queue/Throttler though. I've been hesitant to change that code more than necessary because its used in a bunch of other places that i don't understand well enough to properly test - however, I've definitely noticed a bunch of things I'd like to improve in there while working with it.

I don't really want to hold up this PR on moving that functionality over to queue. At some point though (maybe our next fixit week?) I'd like to give throttler.ts some love and:
- add timers for individual tasks
- fix up some confusing cases where  promises returned by the throttler never resolve
- add explicit support for quotas (instead of tuning the throttler params, you'd provide your quota number and time period, and the throttler would set the concurrency appropriately, automatically retry quota errors, and slow down all tasks if it got a quota error on one)

src/test/deploy/functions/tasks.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+import * as sinon from "sinon";
+
+import * as tasks from "../../../deploy/functions/tasks";
+import { DeploymentTimer } from "../../../deploy/functions/deploymentTimer";
+import { ErrorHandler } from "../../../deploy/functions/errorHandler";
+import { FirebaseError } from "../../../error";
+
+describe("Function Deployment tasks", () => {
+  describe("functionsDeploymentHandler", () => {
+    const sandbox = sinon.createSandbox();
```
inlined, line 11: It feels weird that this is initialized in global scope instead of beforeEach but reset in afterEach

inlined (approved):

PR #3250, opened 2021-03-31
modify SpecParamType.MULTISELECT to be case-sensitive

huangjeff5:
### Description

Some of our official extensions use  the `multiSelect` spec param type. Our validation logic expects `multiselect` (all lowercase), causing publish validation logic to fail. This PR fixes that.


joehan (approved):
LGTM, as long as you tested that you can successfully publish an extension with a multiSelect param.

PR #3262, opened 2021-04-06
Add better typing for CloudFunctions

inlined:
Trying to fork off a few smaller changes from my mondo refactor. We currently use the type of TriggerAnnotation as our type for CloudFunctions' API even though they don't line up 1:1. For example, there's the difference between failurePolicy vs eventType.failurePolicy and serviceAccount vs serviceAccountEmail.

This change introduces a new API type to protect us against type mismatches between trigger discovery and API calls.


src/gcp/cloudfunctions.ts
```diff
@@ -20,17 +33,126 @@ interface Operation {
...
+export type CloudFunctionStatus =
+  | "ACTIVE"
+  | "OFFLINE"
+  | "DEPLOY_IN_PROGRESS"
+  | "DELETE_IN_PROGRESS"
+  | "UNKNOWN";
+export type SecurityLevel = "SECURE_ALWAYS" | "SECURE_OPTIONAL";
+
+export interface FailurePolicy {
+  // oneof action
```
joehan, line 83: I see that you're still holding out hope for more options here 😄

src/gcp/proto.ts
```diff
@@ -0,0 +1,62 @@
...
+        ","
+      )}`
+    );
+  }
+}
+
+// Utility function to help copy fields from type A to B.
+// As a safety net, catches typos or fields that aren't named the same
+// in A an B, but cannot verify that both Src and Dest have the same type for the same field.
+export function copyIfPresent<Src, Dest>(
```
joehan, line 35: Optional: Would be reasonable to add some quick unit tests for these helper functions. They're simple enough that its not blocking tho.

src/gcp/proto.ts
```diff
@@ -0,0 +1,62 @@
...
+    throw new FirebaseError(
+      `Invalid ${typename} definition. ${oneof} can only have one field defined, but found ${defined.join(
+        ","
+      )}`
+    );
+  }
+}
+
+// Utility function to help copy fields from type A to B.
+// As a safety net, catches typos or fields that aren't named the same
```
joehan, line 33: > catches typos or fields that aren't named the same

Does it? it seems like this just silently does nothing in that case

src/gcp/cloudfunctions.ts
```diff
@@ -20,17 +33,126 @@ interface Operation {
...
+  corsPolicy?: CorsPolicy;
+  vpcConnector?: string;
+  vpcConnectorEgressSettings?: "PRIVATE_RANGES_ONLY" | "ALL_TRAFFIC";
+  ingressSettings?: "ALLOW_ALL" | "ALLOW_INTERNAL_ONLY" | "ALLOW_INTERNAL_AND_GCLB";
+
+  kmsKeyName?: string;
+  buildWorkerPool?: string;
+  secretEnvironmentVariables?: SecretEnvVar[];
+  secretVolumes?: SecretVolume[];
+  sourceToken?: string;
```
joehan, line 134: Should we add a comment that `sourceToken` is input only?

joehan (approved):
A few comments, but this seems pretty reasonable to me

src/gcp/cloudfunctions.ts
```diff
@@ -5,7 +5,20 @@ import * as api from "../api";
 import { FirebaseError } from "../error";
 import { logger } from "../logger";
 import * as utils from "../utils";
-import { CloudFunctionTrigger } from "../deploy/functions/deploymentPlanner";
+import { Runtime } from "inspector";
```
mbleigh, line 8: Is this intentional or was it an accidental auto-import? It seems to be shadowed below.

src/gcp/cloudfunctions.ts
```diff
@@ -20,17 +33,126 @@ interface Operation {
   error?: { code: number; message: string };
 }
 
-export const API_VERSION = "v1";
+export interface HttpsTrigger {
+  // output only
+  url?: string;
```
mbleigh, line 38: Could use `readonly` for this if we want to get super nitty ([docs](https://www.typescriptlang.org/docs/handbook/classes.html#readonly-modifier)).

src/gcp/cloudfunctions.ts
mbleigh: It's not though, is it? Couldn't you read this out of an existing function to pass to another?

src/gcp/proto.ts
mbleigh: +1 if we're refactoring to improve code quality may as well improve test coverage.

src/test/deploy/functions/prompts.spec.ts
```diff
@@ -87,11 +98,12 @@ describe("promptForFailurePolicies", () => {
   });
 
   it("should propmt if an existing function adds a failure policy", async () => {
-    const func: CloudFunctionTrigger = {
+    const func = {
       name: "projects/a/locations/b/functions/c",
       entryPoint: "",
       labels: {},
       environmentVariables: {},
+      runtime: "nodejs14" as gcf.Runtime,
```
mbleigh, line 106: nit: Is this cast required? If so, I think I might prefer a plain string union type to a separately declared type. Having to cast string literals is kinda annoying.

src/gcp/cloudfunctions.ts
joehan: Strangely enough, its not returned from the API. I just double checked by adding some console logs to the functions deployment code, and sourceToken is never returned on a function by the API. For our deployment logic, we get it off of the operation metadata.

src/gcp/cloudfunctions.ts
inlined: I later defined a `FunctionSpec` type and just simplify it to `retry` but I want to document the proto accurately in the GCP folder.

src/gcp/proto.ts
inlined: Yup. It's turned typos into compiler errors for me already. `keyof` is magical.

src/gcp/cloudfunctions.ts
inlined: Accidental auto-import

src/gcp/cloudfunctions.ts
inlined: SGTM

src/test/deploy/functions/prompts.spec.ts
inlined: This is a case of type forgetting. If we called this a gcf.CloudFunction then the literal string would compile completely OK, but we're currently using one literal to be both an `Omit<gcf.CloudFunction, gcf.OutputOnlyFields>` and as a `CloudFunctionTrigger` and the literal type just rembers "nodejs14" a string.

The final version of this code uses a `backend.FunctionSpec` for both input and output functions and the literal "nodejs14" is validated and remembered to be a `backend.Runtime` type.

mbleigh (approved):

PR #3265, opened 2021-04-07
Adds handling for when regions are unreachable

joehan:
### Description
Added handling for when regions are unreachable during functions deploy and functions:delete.

### Scenarios Tested
Tested out that a deploy still works if no regions are unreachable. Haven't been able to test the error cases though.

### DIFF
--- a/src/deploy/functions/deploy.ts
+++ b/src/deploy/functions/deploy.ts
@@ -26,9 +26,6 @@ async function uploadSource(context: any): Promise<void> {
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export async function deploy(context: any, options: any, payload: any): Promise<void> {
   if (options.config.get("functions")) {
-    context.existingFunctions =
-      context.existingFunctions || (await gcp.cloudfunctions.listAllFunctions(context.projectId));
-
     await checkHttpIam(context, options, payload);
 
     if (!context.functionsSource) {
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -98,8 +98,8 @@ export async function prepare(context: any, options: any, payload: any): Promise
     );
     if (unreachableRegionsInDeployment) {
       throw new FirebaseError(
-        "The following Cloud Functions regions are currently unreachable:\n" +
-          unreachableRegionsInDeployment.join("\n") +
+        "The following Cloud Functions regions are currently unreachable:\n\t" +
+          unreachableRegionsInDeployment.join("\n\t") +
           "\nThis deployment contains functions in those regions. Please try again in a few minutes, or exclude these regions from your deployment."
       );
     } else {
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -112,5 +112,5 @@ export async function prepare(context: any, options: any, payload: any): Promise
     }
   }
   context.existingFunctions = res.functions;
-  await promptForFailurePolicies(context, options, localFnsInRelease);
+  await promptForFailurePolicies(options, localFnsInRelease, context.existingFunctions);
 }
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -100,14 +100,14 @@ export async function prepare(context: any, options: any, payload: any): Promise
       throw new FirebaseError(
         "The following Cloud Functions regions are currently unreachable:\n" +
           unreachableRegionsInDeployment.join("\n") +
-          "This deployment contains functions in those regions. Please try again in a few minutes, or exclude these regions from your deployment."
+          "\nThis deployment contains functions in those regions. Please try again in a few minutes, or exclude these regions from your deployment."
       );
     } else {
       logLabeledWarning(
         "functions",
         "The following Cloud Functions regions are currently unreachable:\n" +
           res.unreachable.join("\n") +
-          "Cloud Functions in these regions won't be deleted."
+          "\nCloud Functions in these regions won't be deleted."
       );
     }
   }
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -3,15 +3,17 @@ import * as clc from "cli-color";
 import * as ensureApiEnabled from "../../ensureApiEnabled";
 import * as functionsConfig from "../../functionsConfig";
 import * as getProjectId from "../../getProjectId";
-import { logBullet } from "../../utils";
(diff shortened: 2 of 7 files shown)


src/deploy/functions/deploy.ts
```diff
@@ -26,9 +26,6 @@ async function uploadSource(context: any): Promise<void> {
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export async function deploy(context: any, options: any, payload: any): Promise<void> {
   if (options.config.get("functions")) {
-    context.existingFunctions =
```
joehan, line 29: Moved this earlier in the process.

src/deploy/functions/prepare.ts
```diff
@@ -87,5 +89,28 @@ export async function prepare(context: any, options: any, payload: any): Promise
...
+  if (res.unreachable) {
+    const regionsInDeployment = Object.keys(payload.functions.byRegion);
+    const unreachableRegionsInDeployment = regionsInDeployment.filter((region) =>
+      res.unreachable.includes(region)
+    );
+    if (unreachableRegionsInDeployment) {
+      throw new FirebaseError(
+        "The following Cloud Functions regions are currently unreachable:\n" +
+          unreachableRegionsInDeployment.join("\n") +
+          "This deployment contains functions in those regions. Please try again in a few minutes, or exclude these regions from your deployment."
```
taeold, line 103: nit* I think we might want to add a space or a new line before the last sentence for readability. Similar changes to other log lines.

taeold (approved):

src/deploy/functions/prepare.ts
joehan: Ahhh, good point. Will do

src/deploy/functions/prompts.ts
```diff
@@ -30,7 +30,7 @@ export async function promptForFailurePolicies(
 
   context.existingFunctions =
     (context.existingFunctions as CloudFunctionTrigger[]) ||
-    (await gcp.cloudfunctions.listAllFunctions(context.projectId));
+    (await gcp.cloudfunctions.listAllFunctions(context.projectId)).functions;
```
inlined, line 33: In prepare.ts you only stored the functions.

src/functionsDeployHelper.ts
```diff
@@ -211,8 +211,8 @@ export function printSuccess(funcName: string, type: string) {
 }
 
 export async function printTriggerUrls(projectId: string, sourceUrl: string) {
-  const functions = await cloudfunctions.listAllFunctions(projectId);
-  const httpsFunctions = functions.filter((fn) => {
+  const res = await cloudfunctions.listAllFunctions(projectId);
```
inlined, line 214: Separately, why is this needed? We already have the results from create and update.

src/gcp/cloudfunctions.ts
```diff
@@ -435,17 +441,17 @@ export async function listFunctions(projectId: string, region: string): Promise<
       origin: api.functionsOrigin,
     });
     if (res.body.unreachable && res.body.unreachable.length > 0) {
-      throw new FirebaseError(
-        "Some Cloud Functions regions were unreachable, please try again later.",
-        { exit: 2 }
-      );
+      logger.debug(`[functions] unable to reach the following regions: ${res.body.unreachable}`);
```
inlined, line 444: `unreachable` is a list. Let's join it for debug printing.

src/gcp/cloudfunctions.ts
```diff
@@ -426,7 +426,13 @@ export async function deleteFunction(options: any): Promise<Operation> {
  * @param projectId the Id of the project to check.
  * @param region the region to check in.
  */
-export async function listFunctions(projectId: string, region: string): Promise<CloudFunction[]> {
+export async function listFunctions(
```
inlined, line 429: I don't see how it's possible to get unreachable regions when region isn't `-`. Should we use a common helper method and then make `listFunctions` return just a `CloudFunction[]`?

src/deploy/functions/prepare.ts
```diff
@@ -87,5 +89,28 @@ export async function prepare(context: any, options: any, payload: any): Promise
...
+  const res = await gcp.cloudfunctions.listAllFunctions(context.projectId);
+  if (res.unreachable) {
+    const regionsInDeployment = Object.keys(payload.functions.byRegion);
+    const unreachableRegionsInDeployment = regionsInDeployment.filter((region) =>
+      res.unreachable.includes(region)
+    );
+    if (unreachableRegionsInDeployment) {
+      throw new FirebaseError(
+        "The following Cloud Functions regions are currently unreachable:\n" +
+          unreachableRegionsInDeployment.join("\n") +
```
inlined, line 102: nit: Wouldn't it be more consistent in the style to put a tab before elements in a list?

src/gcp/cloudfunctions.ts
```diff
@@ -457,7 +463,12 @@ export async function listFunctions(projectId: string, region: string): Promise<
  * List all existing Cloud Functions in a project.
  * @param projectId the Id of the project to check.
  */
-export async function listAllFunctions(projectId: string): Promise<CloudFunction[]> {
+export async function listAllFunctions(
+  projectId: string
+): Promise<{
```
inlined, line 468: Nit: consider making this a named type so tests can declare things to use that type.

src/functionsDeployHelper.ts
joehan: Good point, this is probably an unnecessary call.

src/deploy/functions/prompts.ts
joehan: Looking closer, this is probably an unnecessary level of safety - we only use this in one place, and we could easily just pass in a list of existing functions instead of taking them from context.

I'm gonna leave this for now in this PR, and then open another PR to do some of these cleanups

src/gcp/cloudfunctions.ts
joehan: Good point, I moved this into a helper method and changed the signature of the exported method. We currently don't use listFunctions anywhere, but it seems like a usual building block to have around, so i'm leaving it for now

src/deploy/functions/prompts.ts
joehan: Ended up being simpler than expected to just pull this call out entirely, so I did so in this pr

inlined: ::sigh:: If I had a dollar for every time I've had to re-edit prompts.ts while doing this refactor. I'm not sure if it's a code smell that this particular file/test breaks so much.

PR #3276, opened 2021-04-14
Add backend namespace before rebasing codebase onto it.

inlined:
The Backend namespace defines types that are neither the __trigger
annotations nor the GCP API types. This follows Google best practices
for implementing services because it buffers the system from changes
in storage or API layers.

In this codebase, it will allow us to support new trigger discovery
techniques (e.g. a container contract HTTP API) and new versions
of APIs if they should ever be needed.

In this change I try out a new method for using args.Context; it's
based on the Go style of Context where a namespace can provide
a value using Context as a private implementation detail.
I'm personally a fan. WDYT?


src/deploy/functions/backend.ts
```diff
@@ -0,0 +1,356 @@
...
+}
+
+export function isEventTrigger(trigger: HttpsTrigger | EventTrigger): trigger is EventTrigger {
+  return "eventType" in trigger;
+}
+
+export type VpcEgressSettings = "PRIVATE_RANGES_ONLY" | "ALL_TRAFFIC";
+export type IngressSettings = "ALLOW_ALL" | "ALLOW_INTERNAL_ONLY" | "ALLOW_INTERNAL_AND_GCLB";
+export type MemoryOptions = 128 | 256 | 512 | 1024 | 2048 | 4096;
+export type Runtime = "nodejs6" | "nodejs8" | "nodejs10" | "nodejs12" | "nodejs14";
```
mbleigh, line 67: `nodejs6` and `nodejs8` are not supported runtimes anymore...can we detect this at the point we would otherwise cast to it instead of supporting it here?

src/deploy/functions/backend.ts
```diff
@@ -0,0 +1,356 @@
...
+  // What we're actually planning to invoke with this topic
+  targetService: FunctionSpec;
+}
+
+export interface ScheduleSpec {
+  id: string;
+  project: string;
+  // Note: schedule is missing in the existingBackend because we
+  // don't actually spend the API call looking up the schedule;
+  // we just infer it from function labels.
```
mbleigh, line 29:
```suggestion
  // we just infer identifiers from function labels.
```

Tiny nit, but when I first read this I thought "it" meant the actual schedule itself.

src/deploy/functions/backend.ts
```diff
@@ -0,0 +1,356 @@
...
+  maxBackoffDuration?: proto.Duration;
+  maxDoublings?: number;
+}
+
+export interface PubSubSpec {
+  id: string;
+  project: string;
+
+  // What we're actually planning to invoke with this topic
+  targetService: FunctionSpec;
```
mbleigh, line 21: It strikes me as odd to embed the entire `FunctionSpec` here instead of just an identifier or reference.

src/deploy/functions/backend.ts
```diff
@@ -0,0 +1,356 @@
...
+}
+
+export type EventFilterKey = "resource";
+export interface EventTrigger {
+  eventType: string;
+
+  // In v1 API, only "resource" is set.
+  // In v2 API, "resource" will be hoisted up as
+  // a top-level "pubsubTopic" resource for the
+  // google.cloud.pubsub.topic.v1.messagePublished event.
```
mbleigh, line 50: I don't understand what this is getting at.

src/deploy/functions/backend.ts
```diff
@@ -0,0 +1,356 @@
...
+}
+
+export const EMPTY: Backend = Object.freeze({
+  requiredAPIs: {},
+  cloudFunctions: [],
+  schedules: [],
+  topics: [],
+});
+
+export function isEmptyBackend(backend: Backend) {
```
mbleigh, line 107: We care about this so we can detect no-op deploys? Or...?

src/deploy/functions/backend.ts
```diff
@@ -0,0 +1,356 @@
...
+  return (
+    Object.keys(backend.requiredAPIs).length == 0 &&
+    backend.cloudFunctions.length === 0 &&
+    backend.schedules.length === 0 &&
+    backend.topics.length === 0
+  );
+}
+
+export type RuntimeConfigValues = Record<string, any>;
```
mbleigh, line 116: This isn't otherwise used in this PR, do we want RTC to be part of this world?

src/deploy/functions/backend.ts
```diff
@@ -0,0 +1,356 @@
...
+  retryCount?: number;
+  maxRetryDuration?: proto.Duration;
+  minBackoffDuration?: proto.Duration;
+  maxBackoffDuration?: proto.Duration;
+  maxDoublings?: number;
+}
+
+export interface PubSubSpec {
+  id: string;
+  project: string;
```
mbleigh, line 18: Not sure how I feel about having `project` show up in each of these specs. AFAIK we aren't contemplating cross-project resource creation, and a given deploy only ever affects a single project. Do we need to embed it everywhere?

src/deploy/functions/backend.ts
```diff
@@ -0,0 +1,356 @@
...
+
+export function functionName(cloudFunction: FunctionSpec) {
+  return `projects/${cloudFunction.project}/locations/${cloudFunction.region}/functions/${cloudFunction.id}`;
+}
+
+// Curried function that's useful in filters. Compares fields in decreasing entropy order
+// to short circuit early (not like there's much point in optimizing JS...)
+export const sameFunctionName = (func: FunctionSpec) => (test: FunctionSpec) => {
+  return func.id === test.id && func.region === test.region && func.project == test.project;
+};
```
mbleigh, line 126: This feels a little overclever maybe? This just take two function specs instead of being a curried function.

src/deploy/functions/backend.ts
```diff
@@ -0,0 +1,356 @@
...
+ ** The naming pattern used to create a Pub/Sub Topic or Scheduler Job ID for a given scheduled function.
+ ** This pattern is hard-coded and assumed throughout tooling, both in the Firebase Console and in the CLI.
+ ** For e.g., we automatically assume a schedule and topic with this name exists when we list funcitons and
+ ** see a label that it has an attached schedule. This saves us from making extra API calls.
+ ** DANGER: We use the pattern defined here to deploy and delete schedules,
+ ** and to display scheduled functions in the Firebase console
+ ** If you change this pattern, Firebase console will stop displaying schedule descriptions
+ ** and schedules created under the old pattern will no longer be cleaned up correctly
+ */
+export function scheduleIdForScheduledFunction(cloudFunction: FunctionSpec) {
```
mbleigh, line 148: `scheduleIdForFunction` instead maybe?

src/deploy/functions/backend.ts
```diff
@@ -0,0 +1,356 @@
...
+
+  proto.renameIfPresent(
+    gcfFunction,
+    cloudFunction,
+    "timeout",
+    "timeoutSeconds",
+    proto.durationFromSeconds
+  );
+  proto.renameIfPresent(gcfFunction, cloudFunction, "serviceAccountEmail", "serviceAccount");
+  proto.copyIfPresent(gcfFunction, cloudFunction, "availableMemoryMb");
```
mbleigh, line 192: Can `copyIfPresent` take a variadic final arg so it doesn't have to be repeated for each field?

src/deploy/functions/backend.ts
```diff
@@ -0,0 +1,356 @@
...
+  timeZone?: string;
+  retryConfig?: ScheduleRetryConfig;
+  transport: "pubsub" | "https";
+
+  // What we're actually planning to invoke with this schedule
+  targetService: FunctionSpec;
+}
+
+export interface HttpsTrigger {
+  httpsOnly: boolean;
```
mbleigh, line 40: If we want to default to `https` should the default false be HTTPS and this becomes `allowInsecure`?

mbleigh (commented):
Looking over the entire thing I'm starting to feel like this might be making the codebase more complicated by being overly functional. I think the API might read a lot cleaner if we had classes for these various types with instance methods to do the various utility work.

We could keep the "spec" types but also have `BackendX` wrappers (so `Backend`, `BackendFunction`, `BackendSchedule`, `BackendTopic` etc) that would each have their own behavior:

```ts
const function = new BackendFunction(spec);
function.name // equivalent to functionName(spec);
function.isSameResource(otherFunction) // equivalent to sameFunctionName(spec)(spec2)
function.toGCFv1()

BackendFunction.fromGCFv1(cloudFunction);
```

We could then also put each of those classes in its own file and re-export them from a top-level `backend/index.ts` (I find the number of exports and things in one file a bit unwieldy here).

FWIW I recognize that the CLI codebase overall is a big mishmash of coding styles. When we started this thing we were on Node 0.10 which IIRC didn't even support `class`. So I understand if you're just taking cues from surrounding code, but I think we're free to chart the course that we think will result in the most maintainable codebase in the long term here.

inlined (commented):
I'm currently leaving comments without changes because I'm editing the final codebase and will just copy back those files once I'm in a good state again.

RE functional coding styles, this was very intentional with complete disregard to the surrounding codebase. I've learned the hard way (mostly at Parse, where I had the freedom to make mistakes) that API types lend themselves very well to POJOs. They're less code and dramatically easier to mock in tests, which obviously leads to better testing. There's a good reason why Google uses Proto even for in-memory types like this; it's just a great best practice. After working hard on some more complex codebases, I really like a dichotomy between stupid POJO data/state types + complex and (seemingly) stateless objects (e.g. a JSON record and a Mongo DB handle).

The fact that this leads to functional programming is just a bonus IMO. JavaScript is a functional language and functional code can be very fluent. Most of what we do on these types is list comprehensions, and those read easier with free-floating functions. `const functionNames = functions.map(functionName);` reads better than `const functionNames = functions.map((fn) => { return fn.name; });`

Similarly, I've replaced cases like:

```
const matchingFucntions = [];
for (const fn of functions) {
  for (const existingFn of existingFunctions) {
    if (fn.name === existingFn.name) {
      matchingFunctions.push(fn);
    }
  }
}
```

With

```
const matchingFunctions = functions.filter(fn => existingFunctions.some(sameFunctionName(fn)));
```

src/deploy/functions/backend.ts
inlined: Now I remember why I didn't do this. I had an optional transform function. Since it was only used once I made that one case an explicit copy.

mbleigh:
After talking it over more offline, I'm not going to block one way or the other on OOP vs. functional other than to say I want the code to be well-organized and maintainable. @bkendall I'd love for you to take a look more from a "patterns and practices for CLI" perspective than an in-depth code review.

inlined:
So I dove back into the code to remember why pubsub and schedule specs have a `targetService`. This is used in:
* `tasks.ts`, where (eventually) scheudle/topic creation/deletion is printed as an operation on the function label
* `release.ts`, where if a customer chooses not to delete their functions we filter out the functions' topics and schedules from the set to delete (we still delete a schedule if it was a scheduled function and no longer is)
* `deploymentPlanner.ts` where we check whether the function for a schedule matches a target group.

I can maybe simplify this a bit and say:

```
// unique name for a service that can be targeted.
// NOTE: May need information about what kind of target this is,
// for e.g. a v1 cloud function, v2 cloud function, or run service.
interface TargetIds {
  id: string;
  region: string;
  project: string;
};
  
interface FunctionSpec extends TargetIds {
  // other fields
}
    
interface PubSubSpec {
  id: string;
  project: string;
  target: TargetIds;
}
     
interface ScheduleSpec {
  // other fields
  target: TargetIds;
}

// Formerly functionMatchesAnyGroup
function targetMatchesAnyGroup(target: TargetIds, groups: string[][]):

// Formerly functionLabels
function serviceLabels(target: TargetIds);

// deploymentHelper.printSuccess takes a TargetIds
```

joehan (commented):
Waiting to see the next round of changes before approving, but I wanted to weigh in on the conversation re: functional vs OOP. I think the changes so far (plus some other snippets shared OOB) look good overall. I think it makes the codebase less approachable overall, even tho its much more elegant in places. 

However, I think that the code in deploymentPlanner.ts is particularly well suited for a functional approach - it involves a ton of list comprehensions to find which functions to update vs create , filters to identify which functions to make schedules for, etc.  Deployment planner contains most of the tricky, subtle, error prone logic here, so anything that makes it clearer and more understandable SGTM

src/deploy/functions/backend.ts
```diff
@@ -0,0 +1,368 @@
...
+  return `firebase-schedule-${cloudFunction.id}-${cloudFunction.region}`;
+}
+
+export function toGCFv1Function(
+  cloudFunction: FunctionSpec,
+  sourceUploadUrl: string
+): Omit<gcf.CloudFunction, gcf.OutputOnlyFields> {
+  if (cloudFunction.apiVersion != 1) {
+    throw new FirebaseError(
+      "Trying to create CloudFunction with wrong API. This should never happen"
```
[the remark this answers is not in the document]
inlined: Hard coding "v1' and "v2". In the codebase I'm going to have FunctionSpec.apiVersion = 2 mean cloudfunctionsv2.ts's API_VERSION regardless of whether that's v2 or v2alpha1 etc.

src/deploy/functions/backend.ts
```diff
@@ -0,0 +1,368 @@
...
+        id,
+        project: specFunction.project,
+        targetService: specFunction,
+      });
+    }
+  }
+  ctx.unreachableRegions.gcfv1 = [...unreachable];
+}
+
+// TODO(inilned): This is just copying existing functionality. Should we complciate
```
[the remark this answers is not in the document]
inlined: Done

src/test/deploy/functions/backend.spec.ts
```diff
@@ -0,0 +1,512 @@
+import { expect } from "chai";
+import * as backend from "../../../deploy/functions/backend";
+import * as gcf from "../../../gcp/cloudfunctions";
+import * as sinon from "sinon";
```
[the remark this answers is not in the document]
inlined: I like that. I also went through the codebase in my master PR to sort everything by default import, named imports, import as namespace.

inlined:
I've been trying to play devil's advocate and get in the mindspace of backend.Foo being a class instead of an interface. I have a few hangups:

1. The constructor is complex enough to require an interface, so I'll basically need today's FunctionSpec and a new FunctionSpecClass.
2. In the full PR, parseTriggers.ts has `addTriggerToBackend(project: string, runtime: backend.Runtime, trigger: TriggerAnnotation, result: backend.Backend)`. This works well because a single TriggerAnnotation can add a FunctionSpec, PubSubSpec, and ScheduleSpec into a Backend. I don't know that it would be as clean if I have to use separate constructors. Though this may not be as horrific as I worry.

src/deploy/functions/backend.ts
inlined: I wrote the code but am not sure I like it. The problem is that GCF can return a node6 or node8 function so `fromGCFv1Function` has to force-cast the Runtime field. I currently warn, but I'm not sure I like having to manually subvert the type system.

bkendall (approved):
This LGTM. I'm happy with providing interfaces and helper functions. Be sure to document the exported functions where appropriate too so that folks can find them and know what they do!

src/deploy/functions/backend.ts
```diff
@@ -0,0 +1,483 @@
+import * as proto from "../../gcp/proto";
+import * as gcf from "../../gcp/cloudfunctions";
+import * as cloudscheduler from "../../gcp/cloudscheduler";
+import * as utils from "../../utils";
+import { FirebaseError } from "../../error";
+import { Context } from "./args";
+import { cloudfunctions } from "../../gcp";
```
joehan, line 7: Looks like this might have been an accidental import

joehan (approved):

PR #3277, opened 2021-04-14
Add field mask support to proto.ts

inlined:
I found a utility for generating field masks useful in the new version of the GCF APIs. Full disclosure, I have mixed feelings about the empty values flag. I'm considering a design pattern like

```
cloudFunction.eventTrigger.failurePolicy = undefined;
// now includes "eventTrigger.failurePolicy"
const fieldMasks = proto.fieldMasks(cloudFunction, /*includeEmptyValues=*/true);
```

It would serialize correctly (`JSON.stringify` drops keys for undefined values), but I'm reticent to make the codebase have semantic differences between missing and undefined.

### DIFF
--- a/src/gcp/proto.ts
+++ b/src/gcp/proto.ts
@@ -60,3 +60,44 @@ export function renameIfPresent<Src, Dest>(
   }
   dest[destField] = converter(src[srcField]);
 }
+
+export function fieldMasks(
+  object: Record<string, any>,
+  includeEmptyValues: boolean = false
+): string[] {
+  const masks: string[] = [];
+  for (const key of Object.keys(object)) {
+    fieldMasksHelper(key, object[key], includeEmptyValues, masks);
+  }
+  return masks;
+}
+
+function fieldMasksHelper(
+  prefix: string,
+  cursor: any,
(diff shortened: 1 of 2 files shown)


src/gcp/proto.ts
```diff
@@ -60,3 +60,44 @@ export function renameIfPresent<Src, Dest>(
...
+  for (const key of Object.keys(object)) {
+    fieldMasksHelper(key, object[key], includeEmptyValues, masks);
+  }
+  return masks;
+}
+
+function fieldMasksHelper(
+  prefix: string,
+  cursor: any,
+  includeEmptyValues: boolean,
```
joehan, line 78: I share your reticence on this - it feels like this would be very confusing behavior to someone who isn't familiar with the codebase. However, I do like much simpler this makes our API methods, and I can't think of a better way to 'hide' the complexity of defining field masks when unsetting fields when updating.

This param name is a bit misleading right now, tho - `includeEmptyValues` has no effect on whether empty arrays or protobuf.Empty values are included. Something like "includeNullOrUndefinedValues" would be more verbose but clearer, imo.

joehan (approved):

PR #3279, opened 2021-04-14
Fixes over-reporting of emulators:start

abeisgoat:
1. The "emulators:start" event is now only reported for "hub". This allows us to use the historical "hub" data to continue to see growth.
2. We now report "emulator:boot" events with emulator IDs for all other emulators.


src/commands/emulators-start.ts
```diff
@@ -106,6 +107,12 @@ Issues? Report them at ${stylizeLink(
     )} and attach the *-debug.log files.
  `);
 
+    // Log the command for analytics, we report the emulator as "hub"
+    // since we originally mistakenly reported emulators:start events
+    // for each emulator, by reporting the "hub" we ensure that our
+    // historical data can still be viewed.
+    track("emulators:start", "hub");
```
samtstern, line 114: Aren't there some situations where hub does not run? Maybe we should just track this in the very top level of the command action.

abeisgoat: Is there? I thought it was a pre-req for any emulator?

samtstern: Pretty sure the hub only runs if we detect a Project ID. @yuchenshi

abeisgoat: So hub runs...

```js
shouldStart() {
  if (name === Emulators.HUB) {
    // The hub only starts if we know the project ID.
    return !!options.project;
  }
  ...
}
  ```
  
 So I guess that could mean the hub doesn't always start?

Comment (samtstern) [reply src/commands/emulators-start.ts:114]: Yep `!!options.project` is checking for a project ID.

File: src/emulator/controller.ts [hunk 2]
```diff
@@ -129,7 +129,7 @@ export async function startEmulator(instance: EmulatorInstance): Promise<void> {
   const name = instance.getName();
 
   // Log the command for analytics
-  track("emulators:start", name);
+  track("emulator:boot", name);
```
yuchenshi, line 132:
```suggestion
  track("emulators:boot", name);
```

yuchenshi (approved):

src/commands/emulators-start.ts
abeisgoat: If that's the case, we could switch to just reporting `emulators:start` with no other args, but that'll make it hard to correlate to historical data. At least if we report "hub" we can filter that in GA and get close.

src/emulator/controller.ts
yuchenshi: nvm I thought it was a typo, now I see it is intentional

abeisgoat: So I thought about that, but this event is now only a single emulator starting hence `emulator` whereas `emulators:start` is several so it's plural? idk, I wasn't sure whats right. If you feel remotely strongly we can go plural :)

yuchenshi: I'd still suggest something unique that cannot be mistaken as a command -- something like `Emulator Run` (following the format of `Hosting Deploy`)

src/commands/emulators-start.ts
samtstern: Let's just take the `track()` out of `startEmulator` and put it in this block:

```ts
  if (shouldStart(options, Emulators.HUB)) {
    const hubAddr = await getAndCheckAddress(Emulators.HUB, options);
    const hub = new EmulatorHub({ projectId, ...hubAddr });
    
    track("emulators:start", "hub"); // ADD THIS LINE
    await startEmulator(hub);
  }
  ```
  
  This way the `emulators:start, hub` line in Analytics means the same thing tomorrow as it does today.

PR #3298, opened 2021-04-21
Typescript-ifying Extensions billing code

joehan:
### Description
Updating some of the billing related code for Extensions to use Typescript & modern best practices. This should have no functional changes.


src/extensions/checkProjectBilling.ts
```diff
@@ -0,0 +1,130 @@
...
+    type: "list",
+    message: `The extension ${clc.underline(
+      extensionName
+    )} requires your project to be upgraded to the Blaze plan. You have access to the following billing accounts.
+Please select the one that you would like to associate with this project:`,
+    choices: choices,
+  });
+
+  let billingEnabled: boolean;
+  if (answer === "Add new billing account") {
```
huangjeff5, line 69: Should this be a constant?

src/extensions/checkProjectBilling.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/**
+ * Directs user to set up billing account over the web and stalls until
+ * user responds.
+ */
+async function setUpBillingAccount(projectId: string, extensionName: string) {
+  const billingURL = `https://console.cloud.google.com/billing/linkedaccount?project=${projectId}`;
+
+  logger.info();
```
huangjeff5, line 87: Is this intentional to add a newline?

src/extensions/checkProjectBilling.ts
```diff
@@ -0,0 +1,130 @@
...
+  return logBillingStatus(billingEnabled, projectId);
+}
+
+/**
+ * Checks whether billing is enabled on the given project.
+ * @param {string} projectId
+ * @returns {Promise<boolean>} True if billing is enabled
+ */
+export async function isBillingEnabled(projectId: string): Promise<boolean> {
+  return cloudbilling.checkBillingEnabled(projectId);
```
huangjeff5, line 113: why do we need a helper for this?

joehan: Honestly, not sure of the original intent. I suspect it was to simplify the dependency graph (you'd only need to import this instead of cloudbilling directly). Having said that, I'm all for getting rid of this.

src/extensions/checkProjectBilling.ts
joehan: Yep, this was in the old JS code. We use this pattern in a bunch of places throughout the CLI to space out logs/prompts to be more readable. Personally, I like it since it makes the code resemble what is going to be printed

src/extensions/checkProjectBilling.ts
joehan: I don't think so? We setto something different in both the if and else block, and then use it after. Do you see a cleaner way to write this chunk that keeps it as a const?

huangjeff5: Oh, I meant the string "Add new billing account" could be a constant defined at the top of the file. Since this is an option that the user selects, not a string they enter. Also if the string option changes, we have to remember to update it in two places.

joehan: Oh, i totally misread which line this comment was referring to 🤦 , good point.

huangjeff5 (approved):

PR #3301, opened 2021-04-22
Fixes checkMinRequiredVersion so that it can be called from before

joehan:
### Description
Adds options as a first argument to checkMinRequiredVersion, so that it can be called from before().

### Scenarios Tested
Manually set extMinVersion in ~/config/configstore/firebase-tools.json, confirmed that the command errors:
Screen Shot 2021-04-22 at 11 04 49 AM

### DIFF
--- a/src/checkMinRequiredVersion.ts
+++ b/src/checkMinRequiredVersion.ts
@@ -8,9 +8,10 @@ const pkg = require("../package.json"); // eslint-disable-line @typescript-eslin
 /**
  * Checks if the CLI is on a recent enough version to use a command.
  * Errors if a min version is found and the CLI is below the minimum required version.
+ * @param options
  * @param key the motd key to that contains semver for the min version for a command.
  */
-export function checkMinRequiredVersion(key: string) {
+export function checkMinRequiredVersion(options: any, key: string) {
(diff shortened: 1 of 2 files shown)


src/checkMinRequiredVersion.ts
```diff
@@ -8,9 +8,10 @@ const pkg = require("../package.json"); // eslint-disable-line @typescript-eslin
 /**
  * Checks if the CLI is on a recent enough version to use a command.
  * Errors if a min version is found and the CLI is below the minimum required version.
+ * @param options
  * @param key the motd key to that contains semver for the min version for a command.
  */
-export function checkMinRequiredVersion(key: string) {
+export function checkMinRequiredVersion(options: any, key: string) {
```
huangjeff5, line 14: Should options be the second parameter? (That's usually the pattern I see but not sure if there's a clear semantic reason)

huangjeff5 (approved):

src/checkMinRequiredVersion.ts
joehan: In this case, no. command.before(func, args...) calls  func(options, args..), so we need options to be the first param.

PR #3305, opened 2021-04-23
Add linter exceptions

inlined:
Per offline conversation, we're allowing JSDocs to omit @param and @return since the Google style guide allows us to omit them when trivial. Reviewers should still ask for documentation on non-obvious parameters or return types.


inlined:
Sorry I let that slip through the cracks. This no longer includes any backend changes.

bkendall (approved):

PR #3310, opened 2021-04-26
Fixes for "Can not convert variables for Cloud Storage rules runtime"

abeisgoat:
Adds boolean type to ExpressionValue translation along with other iOS fixes


samtstern (approved):

src/emulator/storage/rules/runtime.ts
```diff
@@ -313,6 +313,10 @@ function toExpressionValue(obj: any): ExpressionValue {
     return { string_value: obj };
   }
 
+  if (typeof obj == "boolean") {
```
yuchenshi, line 316: nit: these should use triple equals `===`.

src/emulator/storage/rules/runtime.ts
```diff
@@ -313,6 +313,10 @@ function toExpressionValue(obj: any): ExpressionValue {
...
+    return { bool_value: obj };
+  }
+
   if (typeof obj == "number") {
```
yuchenshi, line 320: Ideally we should have some tests for these branches and corner cases (e.g. `undefined`)

yuchenshi (approved):

PR #3311, opened 2021-04-26
Fix Auth emulator export only exporting 20 users.

yuchenshi:
### Description

Fixes b/186445689.

### Scenarios Tested

Not tested 🙃 

### Sample Commands

`firebase emulators:export`


ssbushi (approved):
Perhaps change the wording from `limit` to `maxResults` in src/emulator/auth/operations.ts (Line 451)

PR #3318, opened 2021-04-27
Make promptOnce strongly typed

inlined:
Sorry this took so long. I originally tried to make promptOnce the new prompt and move all cases over. That turned out to be boiling the ocean, so I restarted with much narrower scope. I did expand a bit to make some additional changes along the way:

1. Added the option to pass Options in promptOnce.  This allowed some cases to be simplified by removing a now-redundant options check. I made the options parameter an optional second parameter so I could minimize code changed. This does make it possibly confusing that prompt takes an options as a first parameter and promptOnce takes options as a second parameter. This shouldn't be a major problem since both APIs are typed now.
 2. Changed many prompts with one question to promptOnces


bkendall (approved):
LGTM. Looks like there's a test to fix, but otherwise looks good.

PR #3317, opened 2021-04-27
Storage emulator import/export

samtstern:
### Description

Add import/export support to the Storage emulator.

Googlers: [see design here](https://docs.google.com/document/d/1Dsims-jaOVXj3fGtntFch3pnRU475UIt7OYl1OXGucg/edit?usp=sharing).

### Scenarios Tested

- [x] Manual testing
- [x] e2e integration tests
- [x] Unit tests

### Sample Commands

N/A

PR #3320, opened 2021-04-29
Actually run storage integration tests

samtstern:
### Description

- Actually run the tests (they were not run in CI before)
- Add some tests to the storage emulator to verify some behavior exercised in import/export PR (#3317)

### Scenarios Tested

Uh, the PR is tests.

### Sample Commands

N/A


scripts/storage-emulator-integration/tests.ts
```diff
@@ -244,6 +244,20 @@ describe("Storage emulator", () => {
         });
       });
 
+      describe("#get()", () => {
```
samtstern, line 247: This test is failing with an admin credential error? I don't understand why/how

scripts/storage-emulator-integration/tests.ts
```diff
@@ -244,6 +246,20 @@ describe("Storage emulator", () => {
...
```
samtstern, line 249: There is a failure in our metadata, this errors out due to a `crc32c` mismatch.

abeisgoat:
The tests not running was intentional since our Cloud Build pipeline doesn't currently have a `java` install, however the crc32c mismatch is not intentional.

abeisgoat:
@googlebot I consent.

src/emulator/storage/server.ts
```diff
@@ -63,6 +63,11 @@ export function createApp(
     })
   );
 
+  app.get("/internal/export", (req, res) => {
```
samtstern, line 66: This seems like a mistake?

abeisgoat: yes

abeisgoat (approved):

PR #3324, opened 2021-04-29
Fixes bug where functions:delete didn't respect confirmation prompt (#3314)

### LINKED ISSUE #3314: Firebase CLI deletes function when "No" is replied to the "Are you sure?" question
Hey, Firebase CLI deletes a function when users reply "No" to the "Are you sure?" question.

image

### Environment info

**firebase-tools:** 9.8.0

**Platform:** macOS 11.2.3

### Steps to reproduce

1. Deploy a function to Cloud Functions
2. Delete the function using the Firebase CLI `firebase functions:delete <FN_NAME> --region <REGION>`
3. When asked if you're sure or not, reply with "No"

### Expected behavior

```
When deleting a function using Firebase CLI
And I reply "No" to "Are you sure?" question
Then Firebase *does not* delete my function
```

### Actual behavior

```
When deleting a function using Firebase CLI
And I reply "No" to "Are you sure?" question
Then Firebase *deletes* my function
```

joehan:
### Description
Fixes #3314.

### Scenarios Tested
Screen Shot 2021-04-29 at 3 18 27 PM

### DIFF
--- a/src/commands/functions-delete.ts
+++ b/src/commands/functions-delete.ts
@@ -68,17 +68,15 @@ export default new Command("functions:delete [filters...]")
           return "\t" + helper.getFunctionLabel(func);
         })
         .join("\n");
-      confirmDeletion = await promptOnce(
-        {
-          type: "confirm",
-          name: "confirm",
-          default: false,
-          message:
-            "You are about to delete the following Cloud Functions:\n" +
-            deleteList +
-            "\n  Are you sure?",
-        }
-      );
+      confirmDeletion = await promptOnce({
+        type: "confirm",
+        name: "confirm",
+        default: false,
+        message:
+          "You are about to delete the following Cloud Functions:\n" +
+          deleteList +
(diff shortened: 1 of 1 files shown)


src/commands/functions-delete.ts
```diff
@@ -68,17 +68,15 @@ export default new Command("functions:delete [filters...]")
...
-          type: "confirm",
-          name: "confirm",
-          default: false,
-          message:
-            "You are about to delete the following Cloud Functions:\n" +
-            deleteList +
-            "\n  Are you sure?",
-        },
-      ]);
+      confirmDeletion = await promptOnce({
```
joehan, line 71: The bug here was that prompt returns a map of {question.name: response}, so confirmDeletion was an object that looked like {"confirm": false}. This evaluates truthy, so the command would never abort.

inlined:
Already fixed in #3318 which is getting pushed now.

PR #3326, opened 2021-04-30
Update action-hosting-deploy workflow template

jhuleatt:
### Description

Update the default workflow template for the [Deploy to Firebase Hosting](https://github.com/marketplace/actions/deploy-to-firebase-hosting) action so that it doesn't run on forks.

Github secrets aren't available to forks, so the action will always fail on a PR from a fork.

### Scenarios Tested

`firebase init hosting:github`

Here's the diff between the old output and the new one:

```diff
# This file was auto-generated by the Firebase CLI
# https://github.com/firebase/firebase-tools

name: Deploy to Firebase Hosting on PR
'on': pull_request
jobs:
  build_and_preview:
+    if: github.repository == "jhuleatt/hosting-deploy-sample"
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: FirebaseExtended/action-hosting-deploy@v0
        with:
          repoToken: '${{ secrets.GITHUB_TOKEN }}'
          firebaseServiceAccount: '${{ secrets.FIREBASE_SERVICE_ACCOUNT_MY_PROJECT }}'
          projectId: jeff-test-699d3
-        env:
-          FIREBASE_CLI_PREVIEWS: hostingchannels
```

### DIFF
--- a/src/init/features/hosting/github.ts
+++ b/src/init/features/hosting/github.ts
@@ -148,6 +148,7 @@ export async function initGitHub(setup: Setup, config: any, options: any): Promi
       YML_FULL_PATH_PULL_REQUEST,
       githubSecretName,
       setup.projectId,
+      repo,
       script
     );
 
@@ -269,6 +270,7 @@ type GitHubWorkflowConfig = {
   on: string | { [key: string]: { [key: string]: string[] } };
   jobs: {
     [key: string]: {
+      if: string;
       "runs-on": string;
       steps: {
         uses?: string;
@@ -284,6 +286,7 @@ function writeChannelActionYMLFile(
   ymlPath: string,
   secretName: string,
   projectId: string,
+  repo: string,
   script?: string
 ): void {
   const workflowConfig: GitHubWorkflowConfig = {
@@ -291,6 +294,7 @@ function writeChannelActionYMLFile(
     on: "pull_request",
     jobs: {
       ["build_and_preview"]: {
+        if: `github.repository == "${repo}"`, // secrets aren't accessible on PRs from forks
         "runs-on": "ubuntu-latest",
         steps: [{ uses: CHECKOUT_GITHUB_ACTION_NAME }],
       },


samtstern (approved):

src/init/features/hosting/github.ts
```diff
@@ -284,13 +286,15 @@ function writeChannelActionYMLFile(
...
   projectId: string,
+  repo: string,
   script?: string
 ): void {
   const workflowConfig: GitHubWorkflowConfig = {
     name: "Deploy to Firebase Hosting on PR",
     on: "pull_request",
     jobs: {
       ["build_and_preview"]: {
+        if: `github.repository == "${repo}"`, // secrets aren't accessible on PRs from forks
```
samtstern, line 297: I think this needs to be inside ${{ }}

samtstern: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif

jhuleatt: That's JS string template syntax. The output looks like:

```yaml
if: github.repository == "jhuleatt/hosting-deploy-sample"
```

Since we're outputting the exact repo name, we don't need any brackets, do we?

samtstern: Also I think it needs to be `'` single quotes around repo.

So:
```yaml
if: ${{ github.repository == 'foo/bar' }}
```

jhuleatt: ohh, the whole thing needs to be in brackets?

jhuleatt: oh yeah. [docs](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#example-expression-in-an-if-conditional)

jhuleatt: per our chat, I've updated to `${{ github.event.pull_request.head.repo.full_name == github.repository }}`

bkendall (approved):

PR #3332, opened 2021-05-04
setRules endpoint for Storage Emulator

samtstern:
### Description

Needed to implement some version of:
https://github.com/firebase/firebase-js-sdk/pull/4863

#### Good Rules

Screen Shot 2021-05-04 at 6 25 10 PM

#### Bad Rules

Screen Shot 2021-05-04 at 6 23 57 PM

#### Bad Requests

Screen Shot 2021-05-04 at 6 11 58 PM
Screen Shot 2021-05-04 at 6 11 26 PM

### Sample Commands

N/A

### DIFF
--- a/src/emulator/storage/index.ts
+++ b/src/emulator/storage/index.ts
@@ -93,6 +93,10 @@ export class StorageEmulator implements EmulatorInstance {
     this.destroyServer = utils.createDestroyer(server);
   }
 
+  public updateRules(source: Source): void {
+    this._rulesetSource = source;
+  }
+
   private updateRulesSource(rulesFile: string): void {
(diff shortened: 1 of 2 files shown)


abeisgoat:
Do the other emulators also return a non-json response? I feel like making that machine readable would be ideal

samtstern:
> Do the other emulators also return a non-json response? I feel like making that machine readable would be ideal

I can change this to a simple JSON object like `{ "message": ... }` but I hesitate to include the whole `issues` payload since right now there is no code which will actually consume it and therefore it's just making an API promise earlier than we need to.

abeisgoat (approved):

PR #3334, opened 2021-05-05
Wire the new flags into the backend refactor

inlined:
Finally a stopping point! Poke me for access to the demo app. Some methods are missing, so things aren't as smooth as they could be.

### DIFF
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -447,7 +447,6 @@ export function toGCFv2Function(cloudFunction: FunctionSpec, source: gcfV2.Stora
   }
   proto.copyIfPresent(gcfFunction, cloudFunction, "labels");
 
-  console.error("GCFv2 Function is:", JSON.stringify(gcfFunction, null, 2));
   return gcfFunction;
 }
 
--- a/src/deploy/functions/deploy.ts
+++ b/src/deploy/functions/deploy.ts
@@ -1,21 +1,27 @@
 import * as clc from "cli-color";
 import { setGracefulCleanup } from "tmp";
 
+import { checkHttpIam } from "./checkIam";
 import { functionsUploadRegion } from "../../api";
 import { logSuccess, logWarning } from "../../utils";
-import { checkHttpIam } from "./checkIam";
 import * as args from "./args";
-import * as gcp from "../../gcp";
+import * as backend from "./backend";
+import * as gcs from "../../gcp/storage";
+import * as gcf from "../../gcp/cloudfunctions";
 
 const GCP_REGION = functionsUploadRegion;
 
 setGracefulCleanup();
 
-async function uploadSource(context: args.Context): Promise<void> {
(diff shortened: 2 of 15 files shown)


inlined:
This PR is dependent on #3333. After it is merged into `launch.backend-refactor`, this will be rebased against that branch.

src/deploy/functions/backend.ts
```diff
@@ -335,7 +335,7 @@ export function fromGCFv1Function(gcfFunction: gcf.CloudFunction): FunctionSpec
   }
 
   if (!isValidRuntime(gcfFunction.runtime)) {
-    logger.debug("GCFv1 function has a deprecated runtime:", JSON.stringify(gcfFunction, null, 2));
+    logger.debug("GCFv2 function has a deprecated runtime:", JSON.stringify(gcfFunction, null, 2));
```
taeold, line 338: This change seems accidental - do we really want to do this on `fromGCFv1Function`?

src/deploy/functions/deploy.ts
```diff
@@ -1,21 +1,38 @@
...
 import { logSuccess, logWarning } from "../../utils";
-import { checkHttpIam } from "./checkIam";
 import * as args from "./args";
-import * as gcp from "../../gcp";
+import * as backend from "./backend";
+import * as fs from "fs";
+import * as gcs from "../../gcp/storage";
+import * as gcf from "../../gcp/cloudfunctions";
+import { previews } from "../../previews";
+import { promise } from "ora";
```
taeold, line 13: nit* I don't really get how we order our imports, but the `ora` import looks like it should be part of the first group of imports instead.

src/deploy/functions/tasks.ts
```diff
@@ -291,28 +326,25 @@ export function deleteTopicTask(params: TaskParams, topic: backend.PubSubSpec):
...
-        // Ignore 404 errors from scheduler calls since they may be deleted out of band.
-        errorHandler.record(
-          "error",
-          backend.functionName(task.fn),
-          task.operationType,
-          err.message || ""
-        );
-      }
-    }
+export const schedulerDeploymentHandler = (errorHandler: ErrorHandler) => async (
```
taeold, line 329: q - why suddenly arrow fn style instead of `export async function schedulerDeploymentHandler...`?

src/gcp/run.ts
```diff
@@ -0,0 +1,54 @@
...
+  ],
+};
+
+/**
+ * Sets the IAM policy of a Service
+ * @param name Fully qualified name of the Service.
+ * @param policy The [policy](https://cloud.google.com/run/docs/reference/rest/v1/projects.locations.services/setIamPolicy) to set.
+ */
+export async function setIamPolicy(name: string, policy: IamPolicy): Promise<void> {
+  // Cloud Run has an atypical REST binding for SetIamPolicy. Instead of making the bod a policy and
```
taeold, line 37: nit ` Instead of making the bod` -> `Instead of making the body`

taeold (approved):
lgtm minor nits.

Issue #2126, opened 2020-04-11
409 error while deploying to Firebase Hosting


khaled6mmd:
whenever I try to deploy my project to new firebase project with `firebase deploy` I get this error 

```
⠇  hosting: uploading new files [0/7] (0%)
Error: Task 028c5d7ed641a4a3b636f112e22c192200a3b5932144f1b5d765f73e7c02a6d6 
failed: retries exhausted after 6 attempts
```

old firebase projects work fine

mikejlaurie:
Me too!!

mikejlaurie:
See https://stackoverflow.com/questions/60699826/firebase-hosting-deploy-error-task-5fc-failed-retries-exhausted-after-6-att 
I changed the "public": "www"," to public": "WWW", in firebase.json and it started working..

bkendall:
@khaled6mmd please follow the template. There's a lot of information missing. What OS are you on? What version of firebase-tools are you using?

dalepo:
same is happening to me, tried deploying several times today and nothing.
Debian 9
firebase --version: 7.16.1

bkendall:
@dalepo can you delete the cache that Hosting deploys use and try again? From the root of your project (likely where `firebase.json` is located), the command will look like `rm .firebase/hosting.*.cache` (note the beginning period - it's a hidden file)

dalepo:
thanks for responding @bkendall 
I already tried deleting the entire .firebase folder
```
deploying hosting
hosting[my-project]: beginning deploy...
hosting[my-project]: found 1451 files in www
hosting: uploading new files [28/947] (2%)
Error: Task <long-task-string> failed: retries exhausted after 6 attempts
```
It seems that it gets stuck on certain files, these are mostly minified files. Earlier it got stuck on the 64th

bkendall:
That's super curious. I'm working with some other people to try to figure this out - I'll follow up when I have news.

khaled6mmd:
> @khaled6mmd please follow the template. There's a lot of information missing. What OS are you on? What version of firebase-tools are you using?


windows 10

firebase 8.0.3  

problem got solved after changing path to public then dist/project-name
it deployed successfully

dalepo:
Tried what @khaled6mmd did but still have the same error today.

bkendall:
That's super strange that changing the path fixes it.

[...]

PR #3346, opened 2021-05-07
Delete hash cache after uploads fail with content-hash error

kmcnellis:
### Description

If the content-hash cache gets corrupted somehow, the files uploaded will not have the same sha256 hash as we are expecting, and the upload will fail.  To prevent this, if an upload fails with an error containing "content hash" (the current error is `content hash doesn't match content`), delete all the cached gziped file hashes.  Then the user will be able to re-gzip-hash and deploy successfully next time.

This fixes one of the reported issues in #2126

### Scenarios Tested

- Deploy a hosting site
- Go to .firebase/hosting.<sitehash>.cache
- Modify the hash for any file
- Re-deploy the site (`firebase deploy --only hosting`).  Observe it fails.
- Re-deploy the site again (`firebase deploy --only hosting`).  The deploy should now succeed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Fixes error where Hosting gets in an undeployable state due to the hash cache having incorrect entries.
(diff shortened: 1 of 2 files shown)


bkendall (changes requested):
Please add an entry to the changelog e.g. `- Fixes issue where...`

PR #3358, opened 2021-05-11
Bug bash fixes:

inlined:
1. No longer prints "No HTTPS functions" when there are no HTTPS
   functions
2. No longer prints HTTPS functions installed by extensions


taeold (approved):

inlined:
Accidentally pushed this commit into another PR. Since they're both approved I'll continue there.

PR #3361, opened 2021-05-12
Refactor Functions deploy code to use types that aren't literal API structs

inlined:
Previous versions of the Functions deploy code used a mangled type that wasn't a pure API CloudFunction and wasn't purely the type that we get from the SDK. This introduces a third "conceptual" type for FunctionSpec that we translate to and from so we can be much more rigorous about which format we're using at a given time. While we're at it, I:

* Moved code from src/ to src/deploy/functions
* Refactored the trigger parsing logic in preparation for a more "container contract" discovery mechanism
* Updated some of the API calls we use
* Cleaned up the GCP API calls with a new "proto" library that makes tedious tasks easier and safer


inlined:
Note: This is a "launch" branch of previously reviewed commits.

joehan (approved):
Reviewed all the previous PRs that went into this launch branch, and took a quick pass over this. LGTM!

PR #3363, opened 2021-05-13
Fix missing return when sending errors

IchordeDionysos:
### Description

In both cases there was no fall through leading to an `Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client`.

### Scenarios Tested

Run a script that throws these errors and led to the "headers already sent"-error now don't through it anymore.

### Sample Commands

```
firebase emulators:exec --only storage 'jest --runInBand'
```


samtstern (approved):

samtstern:
@IchordeDionysos great catch, thanks!

PR #3374, opened 2021-05-16
Fix storage emulator for file.download() API

jpreynat:
### Description

The current storage emulator doesn't support using `@google-cloud/storage`'s API to download a file directly.
Instead, it returns the metadata of the file.

This PR fixes this by supporting the `alt=media` query in the `gcloud` emulator API, like this is already done for the `firebase` API.

### Scenarios Tested

```ts
import { Storage } from '@google-cloud/storage';

const storage = new Storage({
    credentials: // ...
});

const bucket = storage.bucket('my-bucket');

const file = bucket.file('path/to/my/file');

file.download().then(([content]) => {
    // content is the metadata of the file instead of the actual content of the file
}));
```

I added a test that failed before updating the `gcloud` API with the error:
```
Error: The downloaded data did not match the data from the server. To be sure the content is the same, you should download the file again.
```

After the update, it passes properly 👌

### DIFF
--- a/src/emulator/storage/apis/gcloud.ts
+++ b/src/emulator/storage/apis/gcloud.ts
@@ -42,7 +42,7 @@ export function createCloudEndpoints(emulator: StorageEmulator): Router {
         res.sendStatus(404);
         return;
       }
-      
+
       const isGZipped = md.contentEncoding == "gzip";
       if (isGZipped) {
         data = gunzipSync(data);
@@ -155,7 +155,7 @@ export function createCloudEndpoints(emulator: StorageEmulator): Router {
       bufs.push(data);
     });
 
-    await new Promise((resolve) => {
+    await new Promise<void>((resolve) => {
       req.on("end", () => {
         req.body = Buffer.concat(bufs);
         resolve();
--- a/src/emulator/storage/apis/gcloud.ts
+++ b/src/emulator/storage/apis/gcloud.ts
@@ -1,4 +1,5 @@
 import { Router } from "express";
+import { gunzipSync } from "zlib";
 import { EmulatorLogger } from "../../emulatorLogger";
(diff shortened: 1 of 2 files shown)


abeisgoat:
Thanks for this, looks good at first glance, will take a closer look in the next few minutes and hopefully merge in 🙏

src/emulator/storage/apis/gcloud.ts
```diff
@@ -35,6 +36,40 @@ export function createCloudEndpoints(emulator: StorageEmulator): Router {
...
+      res.setHeader("Content-Disposition", md.contentDisposition);
+      res.setHeader("Content-Encoding", "identity");
+
+      const byteRange = [...(req.header("range") || "").split("bytes="), "", ""];
+
+      const [rangeStart, rangeEnd] = byteRange[1].split("-");
+
+      if (rangeStart) {
+        const range = {
+          start: parseInt(rangeStart),
```
yuchenshi, line 62: nit: we may want to handle errors like NaN

yuchenshi (approved):

src/emulator/storage/apis/gcloud.ts
jpreynat: This is not handled on the [firebase API either](https://github.com/firebase/firebase-tools/blob/master/src/emulator/storage/apis/firebase.ts#L189-L195).

I'm not sure how you'd like to handle this. Would you mind opening a specific issue for this so it's treated separately that in this PR?
It feels like for the emulator these values should always be correct anyway, since the API calls are supposed to be made by the  `@google-cloud/storage` and `firebase` modules.

jpreynat:
Thanks for merging it @abeisgoat @yuchenshi  🚀

PR #3375, opened 2021-05-16
Fix storage emulator for file.delete() API

jpreynat:
### Description

The current storage emulator doesn't support using `@google-cloud/storage`'s API to delete a file directly.
Instead, it fails with a `Not Found` error.

This PR fixes this by updating the `gcloud` emulator API `DELETE` endpoint to not decode the decode the `objectId` and fix the parameter name in the router, like for other routes.

### Scenarios Tested

```ts
import { Storage } from '@google-cloud/storage';

const storage = new Storage({
    credentials: // ...
});

const bucket = storage.bucket('my-bucket');

const file = bucket.file('path/to/my/file');

file.delete().then(() => {
    // Never reaches here as the method throws even when the file exists
}));
```

I added a test that failed before updating the `gcloud` API with the error:
```
Not Found
```

After the update, it passes properly 👌


scripts/storage-emulator-integration/tests.ts
```diff
@@ -244,6 +246,19 @@ describe("Storage emulator", () => {
...
       });
 
+      describe("#delete()", () => {
+        it.only("should properly delete a file from the bucket", async () => {
+          await testBucket.upload(toDeleteFilePath);
+          await testBucket.file(toDeleteFilePath.split("/").slice(-1)[0]).delete();
+
+          try {
+            fs.statSync(toDeleteFilePath);
+          } catch (err) {
```
yuchenshi, line 256: This also passes if file exists. To mitigate this just assert file doesn't exist (!fs.existsSync).

FYI, in general testing for errors should use the formula `
expect(func).to.throw().with.property('code', 'ENOENT');`

src/emulator/storage/apis/gcloud.ts
```diff
@@ -90,9 +90,8 @@ export function createCloudEndpoints(emulator: StorageEmulator): Router {
     res.json(listResult);
   });
 
-  gcloudStorageAPI.delete("/b/:bucketId/o/:object", (req, res) => {
-    const decodedObjectId = decodeURIComponent(req.params.objectId);
```
yuchenshi, line 94: Do we need to decode? I think the best way to find out is to add a test with a file under a nested path (e.g. `a/b/c.txt`) and delete that

abeisgoat: +1 I'm pretty sure decoding is needed

jpreynat: Decoding was one of the reason why it didn't work.
If you look at the other route handlers (get and patch for instance), they don't apply decoding.

scripts/storage-emulator-integration/tests.ts
jpreynat: Sure, I'll use this check instead 👍

jpreynat: I also see that I forgot to remove the `it.only` on the test.
I'll remove that at the same time.

src/emulator/storage/apis/gcloud.ts
abeisgoat: Can you add a test of deleting at a nested path so we know the behavior is verified?

jpreynat: I actually rewrote the whole test to make sure to use a nested path and that the file exists and is then properly deleted.
I was also foolishly testing if the file was deleted on `fs` instead of on the bucket. This is now checked properly.
Let me know what you think.

abeisgoat: looks great! I think we'll be good to merge unless Yuchen has any issues

jpreynat: @abeisgoat @yuchenshi
Just updated the branch to resolve the conflicts after my other PR was merged.

yuchenshi: Good to go.

yuchenshi (approved):

yuchenshi:
Thanks for the PR and making the changes! This is merged and will be included in the next release.

jpreynat:
Perfect, thanks for the quick review 👌

PR #3378, opened 2021-05-17
Fix storage emulator not handling mediaLink

SamyPesse:
### Description

This PR fixes an issue with the storage emulator where the `mediaLink` in the file metadata being returned when creating a file cannot be used to access the file content.

The `mediaLink` is defined at https://github.com/firebase/firebase-tools/blob/1e4d2e6e7af6b4512edaebdc68263a95c9d88464/src/emulator/storage/metadata.ts#L356

But no route handles this prefix `/download/storage/v1/b/`.

This PR fixes it by adding an alternative route `"/download/storage/v1/b/:bucketId/o/:objectId"` to `"/b/:bucketId/o/:objectId"`

### Scenarios Tested

This PR added an integration test to ensure the mediaLink can be fetched using an http client after the upload.

### DIFF
--- a/src/emulator/storage/apis/gcloud.ts
+++ b/src/emulator/storage/apis/gcloud.ts
@@ -28,58 +28,64 @@ export function createCloudEndpoints(emulator: StorageEmulator): Router {
     next();
   });
 
-  gcloudStorageAPI.get("/b/:bucketId/o/:objectId", (req, res) => {
-    const md = storageLayer.getMetadata(req.params.bucketId, req.params.objectId);
+  gcloudStorageAPI.get(
+    ["/b/:bucketId/o/:objectId", "/download/storage/v1/b/:bucketId/o/:objectId"],
+    (req, res) => {
+      const md = storageLayer.getMetadata(req.params.bucketId, req.params.objectId);
 
-    if (!md) {
-      res.sendStatus(404);
-      return;
-    }
-
-    if (req.query.alt == "media") {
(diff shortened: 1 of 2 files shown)


src/emulator/storage/apis/gcloud.ts
```diff
@@ -28,58 +28,64 @@ export function createCloudEndpoints(emulator: StorageEmulator): Router {
...
 
-  gcloudStorageAPI.get("/b/:bucketId/o/:objectId", (req, res) => {
-    const md = storageLayer.getMetadata(req.params.bucketId, req.params.objectId);
-
-    if (!md) {
-      res.sendStatus(404);
-      return;
-    }
+  gcloudStorageAPI.get(
+    ["/b/:bucketId/o/:objectId", "/download/storage/v1/b/:bucketId/o/:objectId"],
```
SamyPesse, line 32: The only change in this file is this line. The rest is caused by the prettier formatting.

SamyPesse:
An alternative solution is to change this line https://github.com/firebase/firebase-tools/blob/1e4d2e6e7af6b4512edaebdc68263a95c9d88464/src/emulator/storage/metadata.ts#L356 to use the `/b/:bucketId/o/:objectId`  route instead. Let me know if you prefer this alternative!

abeisgoat:
Thanks for this, since the merge is messy I think I might just take your fix and apply it in #3370 if that's alright

abeisgoat:
Ahh dang, it's just as ugly in my branch, I thought it would help the gross diff. My bad, should have just merged your PR. Lesson learned.

PR #3384, opened 2021-05-18
Revert "Hold back Functions refactor until after I/O"

joehan:
Opening this up so I have something to develop new functions features off of.


inlined:
I merged, tried to run tests, and got into linter errors. So I ran the formatter and noticed that one of its changes showed a bug in the promptOnce code for `firebase open` and fixed it.

inlined (approved):

PR #3388, opened 2021-05-19
Allow specifying the region for Hosting functions rewrites, defaulting to us-central1 if not provided

kmcnellis:
### Description

In the firebase.json > hosting > rewrites, add a field `region` to specify that a function is in a non-`us-central1` region.  Does nothing if `function` is not also specified.  If the region is not specified, default to `us-central1`.

### Scenarios Tested

Set the hosting config to the following
```
  "hosting": {
    "public": "public",
    "rewrites": [
      {
        "source": "/testing",
        "function": "testing",
        "region":"europe-west1"
      }
    ]
  }
``` 
and create a function:
```
exports.testing = functions
.region('europe-west1', 'europe-west2')
.https.onRequest(
  async (request, res) => {

    res.send(`target: ${process.env.FUNCTION_TARGET}`)
  },
);
```

- Run the emulator, and the function will properly get rewritten to. 
- Change `europe-west1` to `us-central1`, and the rewrite will no longer work.

Also, deploy hosting with the `--debug` flag, and notice the uploaded config includes the region:
```
 {"config":{"rewrites":[{"glob":"/testing","function":"testing","function_region":"europe-west1"}]},"labels":{"deployment-tool":"cli-firebase"}}
 ```

### CODE REVIEW

**Review by bkendall — APPROVED** [review]

File: src/deploy/hosting/convertConfig.ts [hunk 1]
```diff
@@ -58,8 +58,8 @@ export function convertConfig(config?: HostingConfig): { [k: string]: any } {
         vRewrite.path = rewrite.destination;
       } else if ("function" in rewrite) {
         vRewrite.function = rewrite.function;
-        if (rewrite.function_region) {
-          vRewrite.function_region = rewrite.function_region;
+        if (rewrite.region) {
+          vRewrite.function_region = rewrite.region;
         } else {
           vRewrite.function_region = "us-central1";
```
kmcnellis, line 64: If DynamicLinks is cammelcase, I think function_region should be here, too

bkendall: Ah yes, this is true. I do think this should work too

inlined:
Sorry that i thought this was still in the works, not already done. I thought we were going to make the CLI automatically detect which region if none was specified?

kmcnellis:
That's a future improvement - this is just the first iteration to get the basic behaviour added

PR #3389, opened 2021-05-19
Make functions deploys a bit safer.

inlined:
1. Function deletes come after function creates, which will help make function renaming safer
2. We now explicitly do not support making a scheduled pub/sub function a generic pub/sub function (which was probably broken anyway)


joehan (changes requested):
It looks like you didn't remove the old code to delete functions, just added it to runRegionalDeployment. Am I missing something?

inlined:
Total oversight; should be fixed now.

src/deploy/functions/release.ts
```diff
@@ -60,26 +60,21 @@ export async function release(context: args.Context, options: args.Options, payl
...
+  const allFnsToDelete = Object.values(fullDeployment.regionalDeployments).reduce(
+    (accum, region) => [...accum, ...region.functionsToDelete],
+    [] as backend.FunctionSpec[]
+  );
   const shouldDeleteFunctions = await promptForFunctionDeletion(
     allFnsToDelete,
     options.force,
     options.nonInteractive
   );
-  if (shouldDeleteFunctions) {
```
joehan, line 71: It looks like this new code will always delete functions, even shouldDeleteFunctons = false.

joehan (changes requested):
I think this still may be broken - I don't see where you check whether the user wants to delet functions.

src/deploy/functions/release.ts
inlined: @#$! it removes the topics and schedules from the delete list and still deletes the function. Will fix.

I guess the most dangerous PR is the one that's "too trivial to mess up"

joehan (approved):
LGTM now, as long as you've manually tested both cases

inlined:
Tested both cases

PR #3392, opened 2021-05-20
Create Pub/Sub topic where necessary in Function deploys

inlined:
V2 doesn't currently create Pub/Sub topics. There's an outstanding bug to decide if it should. In the meantime, this papers over the issue by creating one on demand.

While we're at it, I defined a fuller Pub/Sub library in `gcp/` and made `labels` an explicit thing in PubSubSpec so we aren't stuck creating topics w/ the `"deployment": "firebase-scheduled"` label annotation.

### DIFF
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -114,6 +114,8 @@ export function memoryOptionDisplayName(option: MemoryOptions): string {
   }[option];
 }
 
+export const SCHEDULED_FUNCTION_LABEL = Object.freeze({ deployment: "firebase-schedule" });
+
 /** Supported runtimes for new Cloud Functions. */
 export type Runtime = "nodejs10" | "nodejs12" | "nodejs14";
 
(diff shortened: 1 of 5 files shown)


src/gcp/pubsub.ts
```diff
@@ -1,18 +1,67 @@
...
+  return {
+    name: backend.topicName(spec),
+    labels: { ...spec.labels },
+  };
+}
+
+// NOTE: We currently don't need or have specFromTopic.
+// backend.ExistingBackend infers actual topics by the fact that it sees a function
+// with a scheduled annotation. This may not be good enough when we're doing the actual
+// using Run, because we'll have to create topics there. Topic specs have a "target service"
```
joehan, line 66: I think you maybe forgot a word here? `doing the actual` what `using Run`?

src/deploy/functions/backend.ts
```diff
@@ -622,6 +623,7 @@ async function loadExistingBackend(ctx: Context & PrivateContextFields): Promise
       ctx.existingBackend.topics.push({
         id,
         project: specFunction.project,
+        labels: { deployment: "firebase-schedule" },
```
joehan, line 626: Maybe turn "firebase-schedule" into an exported constant somewhere, since we use it in a few places across a few files.

joehan (approved):
Some nits, but LGTM

PR #3408, opened 2021-05-26
Add support for setting user labels on scheduled functions

joehan:
### Description
Adds support for user labels on scheduled functions. 

Turns out we already supported user labels in this code path - except that they were being overwritten on scheduled functions.

### Scenarios Tested
Confirmed that scheduled functions and non-scheduled functions are deployed with labels when firebase-functions passes them through.

### DIFF
--- a/src/deploy/functions/discovery/jsexports/parseTriggers.ts
+++ b/src/deploy/functions/discovery/jsexports/parseTriggers.ts
@@ -227,9 +227,10 @@ export function addResourcesToBackend(
         cloudFunction.trigger.eventFilters.resource = `${cloudFunction.trigger.eventFilters.resource}/${id}`;
       }
 
-      cloudFunction.labels = Object.assign(cloudFunction.labels, {
+      cloudFunction.labels = {
+        ...cloudFunction.labels,
         "deployment-scheduled": "true",
-      });
+      };
     }
 
     want.cloudFunctions.push(cloudFunction);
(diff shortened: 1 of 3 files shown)


src/deploy/functions/discovery/jsexports/parseTriggers.ts
```diff
@@ -227,7 +227,9 @@ export function addResourcesToBackend(
         cloudFunction.trigger.eventFilters.resource = `${cloudFunction.trigger.eventFilters.resource}/${id}`;
       }
 
-      cloudFunction.labels = { "deployment-scheduled": "true" };
+      cloudFunction.labels = Object.assign(cloudFunction.labels, {
```
inlined, line 230: Nit: now that we have object destructuring I personally find it cleaner to use that syntax:

```js
cloudfunctions.labels = {
   ...cloudfunctions.lables,
   "deployment-scheduled": true,
}
```
   
Or at least I would if we wanted to create a copy of the dictionary. In this case, your code is equivalent to:

```js
cloudFunction.labels["deployment-scheduled"] = true;
```

inlined (approved):
LGTM with nit

src/deploy/functions/discovery/jsexports/parseTriggers.ts
joehan: I prefer that too, so I'll switch this over. As is, this is not quite equivalent to cloudFunction.labels["deployment-scheduled"], because labels is optional & can be undefined. Object.assign safely handles that case, wheres the above does not.

inlined:
```
> obj = {}
{}
> obj.labels = Object.assign(obj.lables, {"foo": "bar"});
Uncaught TypeError: Cannot convert undefined or null to object
    at Function.assign (<anonymous>)
```

Object splat supports undefined but `Object.assign` does not.

Issue #1409, opened 2019-06-17
Document complete schema for `firebase.json`


samtstern:
Originally suggested by @matheo here:
https://github.com/firebase/firebase-tools/issues/1303#issuecomment-502407506

We should document the complete schema of `firebase.json` in both human-readable and machine-readable formats.  This file has expanded a lot and it's not always clear what's there and what the defaults are.

bkendall:
Related: https://firebase.google.com/docs/cli/#the_firebasejson_file

long1eu:
Any news on this?

johnnyasantoss:
Any updates?

The provided link has a section called "The firebase.json file" that is not helpful. I still think that a json schema (like [these ones](http://schemastore.org/json/)) would be a better way to add documentation to this mysterious file.

george43g:
so this repo: https://github.com/SirWindfield/vscode-firebase-configuration-schema-plugin

Is a plugin for Vscode that supplies it with a schema for firebase.json

I agree that a schema is needed, at least for autocomplete benefits. That repo is outdated and archived, I do not believe it's maintained. However, it contains a valid and working schema for firebase.json.

PS: I'm tempted to fork the repo and update it... does anyone else want to help out?

samtstern:
@george43g if you do decide to contribute to that repo I would be happy to review your pull requests (I can't approve them, but I can at least give a Firebase stamp of approval).

We actually have a few different needs around documenting the firebase.json schema:

1. It would be great to have a TypeScript definition for use within this repository, right now it's all a lot of `any` when dealing with `firebase.json`
2. We need some way to generate reference documentation for this file to host on firebase.google.com ... we use `externs.js` files in a lot of places now because our tooling supports it but I don't want to double down on that.
3. We need to give developers like all of you autocomplete support in your editors.  I assume JSON Schema is the most standardized way to do this?  But maybe not ...

jthegedus:
>We need to give developers like all of you autocomplete support in your editors. I assume JSON Schema is the most standardized way to do this?

@samtstern Yes, JSON Schema is the most standard way to support autocomplete in editors. IntelliJ uses this by default and VSCode can have support added by individual extensions as @george43g shows or by using popular plugins like the [RedHat YAML](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) plugin which pulls from JSON Schema store and provides both YAML and JSON autocomplete for all available schemas.

sshquack:
[...]

PR #3420, opened 2021-05-27
Begin typing config and options

samtstern:
### Description

Begin to address #1409 and make the use of `options` and `options.config` around the codebase **much** more typesafe.

If we're happy with this approach I will send another PR to slowly introduce the use of these types throughout the codebase. As a proof of concept I used them in exactly one place: `src/deploy/firestore/release.ts`.

Future work could include turning the TS into a JSON Schema and then using it to actually validate the JSON at load time.

### Scenarios Tested

Unit tests should cover this.

### Sample Commands

N/A


src/config.ts
```diff
@@ -0,0 +1,245 @@
+"use strict";
```
samtstern, line 1: The logic in this file is 100% unmodified besides the addition of a new `get src()` function which does a cast.

It seems GitHub couldn't recognize the TS-ification as a clean diff.

src/deploy/firestore/release.ts
```diff
@@ -1,19 +1,26 @@
...
+import { Options } from "../../options";
+import { FirebaseError } from "../../error";
 
 /**
  * Releases Firestore rules.
  * @param context The deploy context.
  * @param options The CLI options object.
  */
-export default async function (context: any, options: any): Promise<void> {
+export default async function (context: any, options: Options): Promise<void> {
```
samtstern, line 12: This is the only file where I am actually using `Options`, `Config`, and `FirebaseConfig` types. It's a proof of concept.

src/config.ts
joehan: Certainly not required, but it would be great to modernize this file a bit as well. Theres a lot of old conventions in here that could be switched easily - private methods that start with __, lots of lodash that can be switched for native methods, etc.

src/options.ts
```diff
@@ -0,0 +1,8 @@
+import { Config } from "./config";
+
+export interface Options {
```
joehan, line 3: @inlined has been working toward a similar goal for the functions deploy codepath: https://github.com/firebase/firebase-tools/blob/master/src/deploy/functions/args.ts#L17. I agree with the approach of trying to standardize these types across the whole codebase, but I think there are some conflicts between these types that should be resolved.

Added him as a reviewer too, since he spent a good bit of time thinking about this recently & may have some insights.

src/deploy/firestore/release.ts
```diff
@@ -1,19 +1,23 @@
...
     return;
   }
-  await rulesDeploy.release(
-    options.config.get("firestore.rules"),
-    RulesetServiceType.CLOUD_FIRESTORE
-  );
+
+  const rulesFile = options.config.src.firestore?.rules;
+  if (!rulesFile) {
+    throw new FirebaseError(`Invalid firestore config: ${options.config.src.firestore}`);
```
joehan, line 20: I _think_ this may print as "Invalid firestore config [Object object]" - can you double check and stringify if needed?

joehan (commented):
This is great!

src/options.ts
samtstern: Oh nice, I will definitely unify this.

src/config.ts
samtstern: I was planning on doing this in a follow-up so that the diff is clear. This is a super critical file and I don't want to mess with it any more than I need to all at once.

bkendall (approved):
I like the idea of this. Joe/Thomas have been in this codebase more than I have recently, so it's good for them to take a look and help join efforts. Thanks for taking on the challenge!

src/config.ts
```diff
@@ -0,0 +1,246 @@
...
+      if (_.get(this._src, target)) {
+        _.set(this.data, target, this._materialize(target));
+      }
+    });
+
+    // auto-detect functions from package.json in directory
+    if (
+      this.projectDir &&
+      !this.get("functions.source") &&
+      fsutils.fileExistsSync(this.path("functions/package.json"))
```
inlined, line 69: Can we just fall back to `functions` and not `functions/package.json`? I want to try adding new languages in our upcoming hackathon.

src/config.ts
```diff
@@ -0,0 +1,246 @@
...
+    ) {
+      this.set("functions.source", "functions");
+    }
+  }
+
+  _hasDeepKey(obj: any, key: string) {
```
inlined, line 75: `obj: Record<string, unknown>` is the right type for what we think of as "an object"

src/config.ts
```diff
@@ -0,0 +1,246 @@
...
+    if (_.has(obj, key)) {
+      return true;
+    }
+
+    for (const k in obj) {
```
inlined, line 80: Please add a `hasOwnProperty` guard or switch to `for (const k of Object.keys(obj))` (I'm not sure how this wasn't a lint blocker, but it certainly goes against the Google style guide)

src/config.ts
inlined, line 75: Why are we still using underscore methods instead of `private` annotations?

src/config.ts
```diff
@@ -0,0 +1,246 @@
...
+    const val = _.get(this._src, target);
+    if (_.isString(val)) {
+      let out = this._parseFile(target, val);
+      // if e.g. rules.json has {"rules": {}} use that
+      const lastSegment = _.last(target.split("."));
+      if (_.size(out) === 1 && _.has(out, lastSegment)) {
+        out = out[lastSegment];
+      }
+      return out;
+    } else if (_.isPlainObject(val) || _.isArray(val)) {
```
inlined, line 98: super nit, but isn't this just `typeof val === "object"`?

src/config.ts
```diff
@@ -0,0 +1,246 @@
...
+    }
+  }
+
+  get src(): FirebaseConfig {
+    // TODO(samstern): We should do JSON Schema validation on this at load time
+    // and then make the _src type stronger.
+    return this._src as FirebaseConfig;
+  }
+
+  get(key: string, fallback?: any) {
```
inlined, line 152: Totally not the right PR for this level of change, but I'd really love to get rid of all these untyped methods in the future.

inlined (approved):
LGTM. Thanks for the initiative here. I'd especially love getting rid of set/get.

For Context, wdyt about the coding style [I used in `backend.ts`](https://github.com/firebase/firebase-tools/blob/master/src/deploy/functions/backend.ts#L587) where individual libraries will extend the object internally w/ a private interface? I took the inspiration from Golang's context.Context where each library uses some key in context to be a private implementation detail.

src/config.ts
samtstern: Just because my intent was to change this file as little as possible in this PR. I found at least one case in the codebase where another file was accessing what should be a "private" method/property of Config.

src/config.ts
samtstern: Same as above, I am not touching the logic of this file as I only want to improve types while minimizing the chance that I introduce a regression.

src/config.ts
samtstern: Agreed!

src/config.ts
samtstern: Yes, in a future PR this would be a good improvement (see other comments for justification)

src/config.ts
samtstern: 👍

src/config.ts
samtstern: Done

PR #3422, opened 2021-05-27
Make environment variables on function instances be accessible during trigger parsing

taeold:
We move the logic for setting up environment variables for function instances from post-trigger parsing to pre-trigger parsing. This makes environment variables accessible during trigger parsing, e.g.

```
const myFunc = functions.runWith({minInstances: process.env.TIER == "PROD" ? 20 : 0})...
```

(Of course, there isn't a way for users to specify environment variables for now. That may change soon.)

This change simply refactors existing logic - nothing new is introduced and function deploys should work exactly as it is done today.

*Warning* There is technically a breaking change where we no longer pass caller's `process.env` when forking a process to parse function triggers in user's source code. We think this is removing what was previously an undefined behavior but recognize we may be breaking someone's workflow. We will soon propose a support for setting project-level environment variables more easily.


src/deploy/functions/discovery/jsexports/parseTriggers.ts
```diff
@@ -69,18 +69,14 @@ function removeInspectOptions(options: string[]): string[] {
 function parseTriggers(
   projectId: string,
   sourceDir: string,
-  configValues: backend.RuntimeConfigValues
+  configValues: backend.RuntimeConfigValues,
+  envs: backend.EnvironmentVariables
 ): Promise<TriggerAnnotation[]> {
   return new Promise((resolve, reject) => {
-    const env = _.cloneDeep(process.env);
+    const env = { ..._.cloneDeep(process.env), ...envs } as NodeJS.ProcessEnv;
```
inlined, line 76: This may be a breaking change, but I actually think we _shouldn't_ ever have been carrying forward `process.env`. It was always weird undefined behavior before and I'd be OK dropping it for now. IMO, we should be populating exactly the env they'll be getting in production.

src/deploy/functions/discovery/jsexports/parseTriggers.ts
```diff
@@ -179,6 +177,7 @@ export function addResourcesToBackend(
       entryPoint: annotation.entryPoint,
       runtime: runtime,
       trigger: trigger,
+      environmentVariables: envs,
```
inlined, line 180: Hmm.. this gives me a bit of pause. If we're requiring project-level env, should env be a backend attribute and not a FunctionSpec attribute?

taeold: Woo that's an excellent point! Thanks - will address.

src/deploy/functions/discovery/jsexports/parseTriggers.ts
inlined: oh, as a nit, you don't need `_.cloneDeep` anymore since you're already cloning with the splat operator and this is just a simple map so there's no recursive cloning to do.

inlined (approved):

PR #3434, opened 2021-06-01
Respect TimeZone in scheduled functions

inlined:
In scheduler spec to job we weren't copying timezones which caused the CLI to fall back to the default timezone (America/Los_Angeles)


joehan (approved):
LGTM, but lets add a test case to backend.spec.ts to cover this case as well

joehan (approved):

PR #3439, opened 2021-06-01
Delete Container Registry images left after Functions deployment

inlined:
Woohoo! Functions customers will no longer have a $0.02 bill that is both annoying and can cause major problems when their credit card refuses the bill and GCP locks their billing profile for failed payment.


inlined:
So I was reviewing some notes on ArtifactRegistry. ArtifactRegistry also implements the Docker API as well as serving its own One Platform API. I'm wondering if `gcp/containerregistry.ts` should be smething more like `gcp/docker.ts` and it can take a URL. Then I can move the domain calculations into `containerCleaner.ts` where it will pivot wether to use GCR or AR based on a preview flag `artifactregistry`.

joehan (commented):
I generally like the idea of switching `gcp/containerregistry.ts` to a more generic `gcp.docker.ts` - however, I have two concerns:
1 - Will the AR still use the weird extra fields on tag that GCR has? If not, I could see the code getting pretty sharded & ugly.
2 - Since the Docker API is so complex, would it be simpler to just use the OnePlatform API for AR, and then eventually get rid of the Docker/GCR code?

src/deploy/functions/containerCleaner.ts
```diff
@@ -0,0 +1,160 @@
...
+          const path = `${func.project}/gcf/${func.region}/${uuid}`;
+          const tags = (await helper.ls(path)).tags;
+          uuidTags[path] = tags;
+        })()
+      );
+    }
+    await Promise.all(loadUuidTags);
+
+    const extractFunction = /^(.*)_version-\d+$/;
+    const entry = Object.entries(uuidTags).find(([, tags]) => {
```
joehan, line 101: Since you throw away the keys in the .`find()` predicate function, should you use `Object.values()` instead of `Object.entries()` here?

src/deploy/functions/containerCleaner.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  async rm(path: string): Promise<void> {
+    const stat = await this.ls(path);
+    const deleteChildren: Promise<void>[] = [];
+    const recursive = stat.children.map((child) => this.rm(`${path}/${child}`));
+    // Let children ("directories") be cleaned up in parallel while we clean
+    // up the "files" in this location.
+
+    const deleteTags = stat.tags.map((tag) => this.client.deleteTag(path, tag));
```
joehan, line 149: Do we need to delete all tags before starting to delete the images? If not, we could combine these await Promise.all()'s and do them in parallel

joehan: I'm also a little concerned about how this behaves if just one call fails - for example, if the first Promise.all rejects, this errors and we never try to deleteImages... but we already made our recursive call, so that has started but will get 'cut off' at some arbitrary point.

src/test/deploy/functions/containerCleaner.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+    trigger: {
+      allowInsecure: false,
+    },
+  };
+
+  // The first function in a region has subdirectories "cache/" and "worker/" in it.
+  it("Handles cleanup of first function in the region", async () => {
+    const cleaner = new containerCleaner.ContainerRegistryCleaner();
+
+    // Any cast because the stub apparently isn't stubbing getNode as a priavte member.
```
joehan, line 130: Typo in private

src/test/deploy/functions/containerCleaner.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+    await cleaner.cleanupFunction(FUNCTION);
+
+    expect(stub.rm).to.have.been.calledOnceWith("project/gcf/us-central1/uuid");
+  });
+
+  // The second function of the region doesn't have subdirectories
+  it("Handles cleanup of second function in the region", async () => {
+    const cleaner = new containerCleaner.ContainerRegistryCleaner();
+
+    // Any cast because the stub apparently isn't stubbing getNode as a priavte member.
```
joehan, line 159: Typo in private

src/test/deploy/functions/containerCleaner.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+
+    await cleaner.cleanupFunction(FUNCTION);
+
+    expect(stub.rm).to.have.been.calledOnceWith("project/gcf/us-central1/uuid");
+  });
+
+  it("Leaves other directories alone", async () => {
+    const cleaner = new containerCleaner.ContainerRegistryCleaner();
+
+    // Any cast because the stub apparently isn't stubbing getNode as a priavte member.
```
joehan, line 187: Typo in private

src/deploy/functions/containerCleaner.ts
joehan: Sounds good to me - this is definitely a tricky one to handle and it seems likely that if one call fails, the rest are high likelihood to as well. I feel better about this strategy tho.

joehan (approved):

web2033:
Do I need to redeploy functions for this to take effect?

inlined:
Yes, run `npm install -i firebase-tools` to get 9.14.0 or greater and re-deploy. The CLI will only clean up images for functions that are still in your deployment; we don't just `rm -rf` everything.

c0dification:
@inlined - I upgraded to 9.14 last week and did a few deployments since then. But since yesterday morning, when I deploy a function, the script successfully deploys the function and then just sits on the last line, `i  functions: cleaning up build files...`, and never terminates. I have to cntr-c to get my terminal back. I've let it set for up to an hour in the terminal without any change. Any thoughts on this? 

Here's an example. 

```
firebase deploy --only functions:scheduledFirestoreExport

=== Deploying to 'vizzn-poc'...

i  deploying functions
Running command: npm --prefix "$RESOURCE_DIR" run lint

> functions@3.22.0 lint
> tslint --project tsconfig.json

Running command: npm --prefix "$RESOURCE_DIR" run build

> functions@3.22.0 build
> rimraf ./lib && ./node_modules/.bin/tsc && cpx "./src/**/*.hbs" ./lib --clean

✔  functions: Finished running predeploy script.
⚠  functions: package.json indicates an outdated version of firebase-functions.
Please upgrade using npm install --save firebase-functions@latest in your functions directory.
i  functions: packaged functions (2.02 MB) for uploading
i  pubsub: ensuring required API pubsub.googleapis.com is enabled...
i  scheduler: ensuring required API cloudscheduler.googleapis.com is enabled...
✔  scheduler: required API cloudscheduler.googleapis.com is enabled
✔  pubsub: required API pubsub.googleapis.com is enabled
i  functions: updating Node.js 10 function scheduledFirestoreExport(us-central1)...
✔  functions[scheduledFirestoreExport(us-central1)]: Successful upsert schedule operation. 
✔  functions[scheduledFirestoreExport(us-central1)]: Successful update operation. 
i  functions: cleaning up build files...
```

inlined:
That seems quite peculiar. Have you had many different functions by many different names in this project in the past? Unfortunately the GCR layout is to have each function in a different UUID named "directory" so we have to peek into every directory before we even find the ones we want to clean up. Could you check out the "gcf" directory in https://console.cloud.google.com/gcr and see how many images are there? You might need to manually clean them up to get unstuck. Unfortunately you cannot delete by directory, only by image, because the API is _very_ chatty and the console team decided not to implement a feature with unbounded complexity.

PR #3444, opened 2021-06-03
Import/export download tokens

### LINKED ISSUE #3414: Emulator does not restore firebaseStorageDownloadTokens on import

**firebase-tools:** 9.12.0

**Platform:** Windows 10

### [REQUIRED] Test case

```javascript
import { getStorage, useStorageEmulator, ref, getDownloadURL } from '@firebase/storage'

const storage = getStorage()
useStorageEmulator(storage, '127.0.0.1', 9199)

async function getMyFile () {
  const reference = ref(storage, 'myfile')
  return await getDownloadURL(reference)
}
```

1. Run the emulators (including storage): `firebase emulators:start --import data --export-on-exit`
2. Upload a test file `myfile` to Storage using the Emulator UI
3. Call the `getMyFile()` function from Javascript - works as expected, can fetch the file
4. `Ctrl+C` to shut down the emulator. `myfile` is exported to `data/`
5. Note that `data/storage_export/metadata/<bucket>/myfile.json` doesn't contain anything in `customMetadata` or `downloadTokens` properties? (This seems unexpected)
6. Start the emulators again: `firebase emulators:start --import data --export-on-exit`
7. Call the `getMyFile()` function from Javascript - receive error:
`FirebaseError: Firebase Storage: The given file does not have any download URLs. (storage/no-download-url)`

Download tokens are exported and imported between emulator runs.

Download tokens are not exported/imported by the storage emulator.

samtstern:
### Description

Fixes #3414

### Scenarios Tested

Expanded unit test

### Sample Commands

N/A


src/emulator/storage/metadata.ts
```diff
@@ -132,10 +132,15 @@ export class StoredFileMetadata {
...
 
   private setDownloadTokensFromCustomMetadata() {
-    if (!this.customMetadata) return;
+    if (!this.customMetadata) {
+      return;
+    }
 
     if (this.customMetadata.firebaseStorageDownloadTokens) {
-      this.downloadTokens = this.customMetadata.firebaseStorageDownloadTokens.split(",");
+      this.downloadTokens = [
```
samtstern, line 140: This is the part I am unsure about ... but it feels right to me.

abeisgoat: seens very reasonable to me, we must have just missed this when I made the tweaks to download token logic last week

abeisgoat (approved):

PR #3452, opened 2021-06-04
Add functions:env:* family of commands

taeold:
We add following commands that allows users to interact with the EnvStore service to store environment variables for CF3:

* `functions:env:add [FOO=bar BAR=foo...]` - adds given key-value pairs as env vars in EnvStore
* `functions:env:remove [FOO,BAR,...]` - removes env vars associated with given keys from the EnvStore
* `functions:env:set [FOO=bar BAR=foo...]` - sets given key-value pairs as env vars, completely replacing the existing set 
* `functions:env:clear` - removes all env vars from the EnvStore
* `functions:env:clone --from=another-project` - copies env vars from another project

Few missing pieces (will follow up soon):

- `functions:env:migrate` command to help users migrate from `functions:config`.
- Prompts for opting users in to the CF3-managed environment variables.


src/commands/functions-env-clone.ts
```diff
@@ -0,0 +1,46 @@
...
+  .option("--from <projectId>", "the project from which to clone environment variables")
+  .option("--only <keys>", "a comma-separated list of keys to clone")
+  .option("--except <keys>", "a comma-separated list of keys to not clone")
+  .before(requirePermissions, [
+    "firebase.envstores.create",
+    "firebase.envstores.get",
+    "firebase.envstores.list",
+    "firebase.envstores.update",
+  ])
+  .action(async (_: string[], options: any) => {
```
joehan, line 20: Why do we need this? I checked a few other commands that don't take args, and none of them do this.

src/commands/functions-env-clone.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { Command } from "../command";
+import { requirePermissions } from "../requirePermissions";
+import * as fenv from "../functions/env";
+import * as getProjectId from "../getProjectId";
+import * as utils from "../utils";
+
+export default new Command("functions:env:clone")
+  .description("clone environment variables from another project")
+  .option("--from <projectId>", "the project from which to clone environment variables")
```
joehan, line 11: I'm sure this was discussed in the API proposal, but this isn't really optional - why is it an option?

src/functions/env.ts
```diff
@@ -0,0 +1,203 @@
...
+  if (!/^[A-Z_][A-Z0-9_]+$/.test(key)) {
+    throw new FirebaseError(
+      "Invalid environment variable name " +
+        clc.bold(key) +
+        ", must start with an uppercase ASCII letter or underscore" +
+        ", and then consist of uppercase ASCII letters, digits, and underscores."
+    );
+  }
+  // Keys cannot contain the prefix X_GOOGLE_.
+  if (/^X_GOOGLE_.*$/.test(key)) {
```
joehan, line 64: Nit: Prefer key.startsWith() since it makes it more obvious what this is doing.

src/functions/env.ts
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Parse list of strings in key=value format as an object.
+ *
+ * @param {string[]} args List of strings in key=value format.
+ * @return {Record<string, string>} An object with key, value pairs.
+ */
+export function parseKvArgs(args: string[]): Record<string, string> {
+  const envs: Record<string, string> = {};
+  args.forEach((arg) => {
```
joehan, line 81: Can't link it here, but the internal TS style guide suggests avoiding forEach and instead using for/of (or just plain for loops)

src/functions/env.ts
```diff
@@ -0,0 +1,203 @@
...
+    const parts = arg.split("=");
+    if (parts.length < 2) {
+      throw new FirebaseError(`Invalid argument ${clc.bold(arg)}, must be in key=val format`);
+    }
+    const key = parts[0];
+    validateKey(key);
+    const val = parts.slice(1).join("="); // Val may have contained '='.
```
joehan, line 88: You can avoid having to do this by adding a limit to the split: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split

src/functions/env.ts
```diff
@@ -0,0 +1,203 @@
...
+}
+
+/**
+ * Get environment variables from the EnvStore Service.
+ *
+ * @return {Promise<Record<string, string>>} An object that contains environment variables.
+ */
+export async function getEnvs(projectId: string): Promise<Record<string, string>> {
+  const envStore = await envstore.getStore(projectId, ENVSTORE_ID);
+  return envStore.vars || {};
```
joehan, line 101: Why do we need to include this `|| {}` everywhere? Based on your types, envStore.vars always exists

src/functions/env.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  let filterFn: (k: string) => boolean;
+  if (only.length) {
+    filterFn = (k) => only.includes(k);
+  } else {
+    filterFn = (k) => !except.includes(k);
+  }
+  const envs: Record<string, string> = {};
+  const fromEnvs = await getEnvs(fromProjectId);
+  Object.entries(fromEnvs).forEach(([k, v]) => {
```
joehan, line 197: Same as above - prefer using for... of... instead of forEach

src/functions/env.ts
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * Clone environment variables in fromProjectId to toProjectId.
+ *
+ * @param {string} fromProjectId Project to clone environment variables from.
+ * @param {string} toProjectId Project to clone environment variables to.
+ * @param {string[]} only List of keys to clone.
+ * @param {string[]} except List of keys to exclude when cloning.
+ * @return {Promise<Record<string, string>>} An object with environment variables from the EnvStore.
+ */
+export async function clone(
```
joehan, line 179: Optional: Given that this has 2 pairs of identically typed args that are _really_ bad to mix up, this is probably a good candidate to use a parameter type

joehan (commented):
This looks pretty good to me! Some minor style suggestions

src/commands/functions-env-clone.ts
taeold: Great to know. I didn't know not having args was an option.

src/commands/functions-env-clone.ts
taeold: Don't see anyone calling this out - I think the intent here to mirror `functions:config:clone` command. I tried looking up design doc by @inlined from 2016(?) to see if anyone called this out, but no one did!

I do like that --from makes it explicit what the argument is about. At the end of the day `--from` flag isn't optional here either.

src/functions/env.ts
taeold: Good to know! Fixed.

src/functions/env.ts
taeold: Hmm this didn't work as expected:

```js
const w = "ABC=EF=G"
console.log(w.split("="))
=> ["ABC", "EF", "G"] 
console.log(w.split("=", 2))
=> ["ABC", "EF"] 
```
[link](https://www.typescriptlang.org/play?#code/MYewdgzgLgBA7jAvDARAQQEIGFEFEBiiA4igFCiQgA2ApgHRUgDmAFHHRAA5UCWULKRCgCUw8uAjV6jVuy69+glABoYAJlFA)

From docs:

> limit Optional
A non-negative integer specifying a limit on the number of substrings to be included in the array. If provided, splits the string at each occurrence of the specified separator, but stops when limit entries have been placed in the array. **Any leftover text is not included in the array at all**.

Am I holding this correctly?

src/functions/env.ts
taeold: Good catch!

src/functions/env.ts
taeold: Can you expand on what you mean by parameter type? Searching for parameter type on typescript handbook isn't helping me too much.

joehan: Sorry - I misspoke a bit, the real term is parameter object pattern. Chatted you a link that explains it well, but basically its refactoring this to take a CloneArgs object so it has named parameters

src/functions/env.ts
joehan: Ahhh, i dint realize it tossed away the rest. Your original way seems best then!

src/commands/functions-env-clone.ts
joehan: Seems reasonable to me - I agree that its good to mirror the old command

joehan (commented):
Looks like the diff for this got a bit mangled :/

taeold:
@joehan Whoops. Didn't realize Github doesn't pick up changes when rebasing. Fixed now!

joehan (approved):

src/functions/env.ts
```diff
@@ -0,0 +1,208 @@
...
+ * Validates string for use as an env var key.
+ *
+ * We restrict key names to ones that conform to POSIX standards.
+ * This is more restrictive than what is allowed in Cloud Functions or Cloud Run.
+ *
+ * @param key {string} Key to validate
+ */
+export function validateKey(key: string): void {
+  // key must not be one of the reserved key names.
+  if (RESERVED_KEY_NAMES.includes(key)) {
```
inlined, line 47: I would probably reserve every key that starts with "FIREBASE"

src/commands/functions-env-set.ts
```diff
@@ -0,0 +1,61 @@
...
+    "firebase.envstores.update",
+  ])
+  .action(async (args: string[], options: { file?: string; force: boolean }) => {
+    if (!options.file && !args.length) {
+      return utils.reject("Must supply at least one key/value pair, e.g. " + clc.bold("FOO=bar"));
+    }
+    let setEnvs: Record<string, string>;
+    if (options.file) {
+      const buf = fs.readFileSync(path.resolve(options.file), "utf8");
+      setEnvs = dotenv.parse(buf.toString().trim());
```
inlined, line 30: Don't we need to validate keys here?

src/functions/envstore.ts
```diff
@@ -0,0 +1,96 @@
+import { firebaseApiOrigin } from "../api";
+import { FirebaseError } from "../error";
+import { Client } from "../apiv2";
+
+interface EnvStoreEntry {
+  name?: string;
+  vars: Record<string, string>;
+}
+
+const apiClient = new Client({ urlPrefix: firebaseApiOrigin, auth: true });
```
inlined, line 10: initialize with `apiVersion: v1alpha` and remove the prefix in your fetch requests.

src/functions/envstore.ts
```diff
@@ -0,0 +1,96 @@
...
+  envStoreId: string,
+  envs: Record<string, string>
+): Promise<EnvStoreEntry> {
+  const body = {
+    name: envStoreId,
+    vars: envs,
+  };
+  let response;
+  try {
+    response = await apiClient.post<EnvStoreEntry, EnvStoreEntry>(
```
inlined, line 45: If this is why you created EnvStoreEntry you should probably either type `body` or just call this template parameter `typeof body`

src/functions/envstore.ts
```diff
@@ -0,0 +1,96 @@
...
+ * @return {EnvStoreEntry} EnvStore entry for given envstore id.
+ */
+export async function getStore(projectId: string, envStoreId: string): Promise<EnvStoreEntry> {
+  let response;
+  try {
+    response = await apiClient.get<EnvStoreEntry>(
+      `/v1alpha/projects/${projectId}/envStores/${envStoreId}`
+    );
+  } catch (err) {
+    throw new FirebaseError(`Failed to make request: ${err.message}`, { original: err });
```
inlined, line 24: Here and everywhere else, you'll probably thank yourself in the future if your debug log lines describe the request that failed.

src/test/functions/env.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+      const keys = ["FOO", "ABC_EFG", "A1_B2"];
+      keys.forEach((key) => {
+        expect(() => {
+          fenv.validateKey(key);
+        }).not.to.throw;
+      });
+    });
+
+    it("should throw error given invalid keys", () => {
+      const keys = ["", "A", "1F", "B=C"];
```
inlined, line 34: Nit: Isn't it impossible for you to have the key `B=C` because the parser will first split at the `=` and consider only `B`?

src/test/functions/env.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+      keys.forEach((key) => {
+        expect(() => {
+          fenv.validateKey(key);
+        }).not.to.throw;
```
inlined, line 38: Didn't you want these to throw?

src/test/functions/env.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    it("should clone all environment variables from the source project", async () => {
+      const envs = {
+        FOO: "bar",
+        AB1: "ab1",
+      };
+      const fromP = "project-1";
+      const toP = "project-2";
+
+      createStore.onFirstCall().resolves({ vars: envs });
```
inlined, line 101: Getting very nitty here, but:

1. It might be clearer if these API calls are in sequence (delete before create)
2. It might be clearer to replace `onFirstCall` with `withArgs`
3. Since the order of these API calls matters (e.g. delete before create) you might want to add
`expect(deleteStore.calledBefore(createStore))`

inlined (commented):
I was absolutely derelict of my duty to prioritize getting yesterday's demo finished over unblocking a teammate. I feel especially bad that I did this because of the CL size when the CL was refreshingly easy to read.

Thank you @joehan for covering in my absence.

taeold:
@inlined Thanks for the feedback. I'll create a new PR to address your comments.

src/commands/functions-env-set.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { Command } from "../command";
+import { logger } from "../logger";
+import { promptOnce } from "../prompt";
+import { requirePermissions } from "../requirePermissions";
+import * as fenv from "../functions/env";
+import * as getProjectId from "../getProjectId";
+import * as utils from "../utils";
+
+export default new Command("functions:env:set [values...]")
```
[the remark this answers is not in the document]
taeold: Hmm adding `<KEY>=<value>`  seems to mess up the argument parsing in unexpected ways. I'm adding more details on the description to help users through.

src/functions/env.ts
```diff
@@ -0,0 +1,208 @@
+import * as clc from "cli-color";
+
+import { FirebaseError } from "../error";
+import * as envstore from "./envstore";
+
+// Exported for testing-only
+export const ENVSTORE_ID = "firebase-functions";
+const RESERVED_KEY_NAMES = [
```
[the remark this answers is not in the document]
taeold: Added bunch more, including GCLOUD_PROJECT, from https://cloud.google.com/functions/docs/env-var#nodejs_8_python_37_and_go_111.

src/functions/env.ts
```diff
@@ -0,0 +1,208 @@
...
+  "K_CONFIGURATION",
+];
+
+/**
+ * Format environment variables into console-friendly strings.
+ *
+ * @param {Record<string, string>} envs Environment variables to format.
+ * @return {string} Formatted string suitable for printing.
+ */
+export function formatEnv(envs: Record<string, string>): string {
```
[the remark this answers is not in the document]
taeold: After reading Google TS style guide advising against `.forEach`, I think I sided with using a loop instead of using `maps` and it's friends. Seems more readable to use `map `here.

src/commands/functions-env-set.ts
taeold: Great catch!!

PR #3460, opened 2021-06-07
Add asian-southeast1 to RTDB CLI

fredzqm:
### Description

Fixes #3455

### Scenarios Tested

Create, read, edit, update rules for RTDB database in Singapore.

### Sample Commands

`firebase database:instances:create fredzqm-staging-apse1-2 --location asia-southeast1`
`firebase --instance fredzqm-staging-apse1 /`
`firebase deploy --only database`

PR #3463, opened 2021-06-08
Unbreak build

inlined:
A recent refactor left the emulator with an unresolved `require`. 

This change should unblock #3460 3460


bkendall (approved):
Here's your LGTM for when all the tests pass :)

PR #3482, opened 2021-06-10
Default timeout is 60s + user define custom timeout per emulator

### LINKED ISSUE #2379: Error: TIMEOUT: Port 8080 on localhost was not active within 30000ms
[firebase-debug.log](https://github.com/firebase/quickstart-nodejs/files/4795530/firebase-debug.log)
Cloud Firestore emulator quickstart
I am getting the Timeout error running the ````emulators:exec```` command on an unchanged project.

I tried other ports ````8081, 8082```` and 

````
$ sudo firebase emulators:exec --only firestore 'npm test'
$ firebase emulators:exec 'npm test'
$ firebase emulators:start
```` 
produced the same Timeout error.

````

./firebase/quickstart-nodejs/firestore-emulator/javascript-quickstart

macOS v 10.15.6

$ firebase --version
8.4.2
$ node --version
v14.4.0
$ 

````

````
$ firebase emulators:exec --only firestore 'npm test'
i  emulators: Starting emulators: firestore
i  firestore: Firestore Emulator logging to firestore-debug.log

Error: TIMEOUT: Port 8080 on localhost was not active within 30000ms
````

### PR DESCRIPTION
**Pull request by mrenrich84**
### Description

Fix #2379 
Firestore is failing to start in the hardcoded timeout of 30 secs.
This PR:
- increase the default timeout to 60 secs
- leave open to the user the configuration of the timeout for each emulator in the same pattern of host and port

### Scenarios Tested

GIVEN a user does not provide a timeout
WHEN `firebase emulators:start` is called
THEN firestore emulator will fail to run with `Error: TIMEOUT: Port 8081 on localhost was not active within 60000ms`

GIVEN a user provides a timeout in `firebase.json` for firestore of 10ms
WHEN `firebase emulators:start --only firestore` is called
THEN firestore emulator will fail to run with `Error: TIMEOUT: Port 8081 on localhost was not active within 10ms`

GIVEN a user provides a timeout in `firebase.json` for hosting of 10ms
WHEN `firebase emulators:start --only firestore` is called
THEN firestore emulator will fail to run  with `Error: TIMEOUT: Port 8081 on localhost was not active within 60000ms`*

*I understand these is a bit flacky - couldn't think any better at midnight 😛 
I actually tested them with the old value of 30 secs

### Sample Commands

In `firebase.json`:
```
{
  //...
  "emulators": {
    "auth": {
      "port": 9099
    },
    "firestore": {
      "timeout": 60000,
      "port": 8081
    },
    "hosting": {
      "port": 5000
    },
    "ui": {
      "enabled": true
    }
  }
}
```

### DIFF
--- a/src/emulator/controller.ts
+++ b/src/emulator/controller.ts
@@ -9,7 +9,7 @@ import * as track from "../track";
 import * as utils from "../utils";
 import { EmulatorRegistry } from "./registry";
 import {
-  Address,
+  UserConfig,
   ALL_SERVICE_EMULATORS,
   EmulatorInstance,
   Emulators,
@@ -43,7 +43,7 @@ import { StorageEmulator } from "./storage";
 import { getDefaultDatabaseInstance } from "../getDefaultDatabaseInstance";
 import { getProjectDefaultAccount } from "../auth";
 
-async function getAndCheckAddress(emulator: Emulators, options: any): Promise<Address> {
+async function getAndCheckConfig(emulator: Emulators, options: any): Promise<UserConfig> {
   let host = Constants.normalizeHost(
     options.config.get(Constants.getHostKey(emulator), Constants.getDefaultHost(emulator))
   );
@@ -118,7 +118,10 @@ async function getAndCheckAddress(emulator: Emulators, options: any): Promise<Ad
     );
   }
 
-  return { host, port };
+  const timeoutVal = options.config.get(Constants.getTimeoutKey(emulator), undefined);
+  const timeout = timeoutVal ? parseInt(timeoutVal, 10) : Constants.getDefaultTimeout(emulator);
+
+  return { host, port, timeout };
 }
 
 /**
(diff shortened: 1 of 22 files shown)

### CODE REVIEW

File: src/emulator/controller.ts [hunk 1]
```diff
@@ -311,7 +314,10 @@ function findExportMetadata(importPath: string): ExportMetadata | undefined {
   }
 }
 
-export async function startAll(options: any, showUI: boolean = true): Promise<void> {
+/**
```
mrenrich84, line 317: linting non sense...

samtstern:
@mrenrich84 thanks for this! I am in favor of increasing the timeout to 60s but not yet convinced that adding a `timeout` parameter to the `json` is the right move. Plus even if it is, we have an internal review process for all API / configuration changes and we'd have to go through that.

Would you mind reducing this PR to just the 30s --> 60s change and then we can have a separate discussion about the `timeout` configuration?

mrenrich84:
😆 😆 😆 
Was thinking about that all the time...
What about an env variable?

samtstern:
@mrenrich84 I think an env var brings up the same discussion. I'd like to think about a few things before moving forward with more configuration:

- Is there a way we can change the behavior of Firestore imports so that they don't cause timeouts at all?
- If we need timeout configuration, do we need per-emulator timeout configs or should it be global?

mrenrich84:
There are few conflicts with the main branch and I'm too lazy to get lost in merge/rebase now 😛 
Opened a new PR for just hardcoding the timeout to 60 secs.

Issue #2994, opened 2021-01-05
Support ES Modules for Node 14 functions


mbleigh:
Currently, it is not possible to use native ES Modules with Cloud Functions for Firebase in Node 14 due to how triggers are parsed from the code. Unfortunately, adding support is not trivial (or I would have done it when I added Node 14 support generally), but we'd like to do it in the future.

Please upvote this issue if it's important to you!

taeold:
In addition to problems with how triggers are parsed in firebase-tools repo, the functions framework (used to load and run user-provided code in google cloud functions) doesn't seem to support ESM yet: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/233.

mbleigh:
@taeold good catch. Seems like even if we finished the work here we'd be blocked on that. Perhaps we should put a pin in this until work has been done in the functions framework.

mesqueeb:
I just want to mention to all who come here to upvote:

Don't forget to **click "subscribe" on the right side to get updates!**!

larssn:
Does this affect typescript projects?

We use this tsconfig for node 14: https://github.com/tsconfig/bases/blob/master/bases/node14.json

mbleigh:
@larssn correct, since TypeScript with default configuration compiles down into CommonJS, it is not affected -- you can continue to use ESM in TypeScript just like you can in Node < 14.

spicemix:
Although even in a Typescript project, the emulators refuse to use `p-queue@7` for instance, leaving me using an old non-ESM version. Thanks in advance!

Spencer-Easton:
@mbleigh Will top level await work it compiles from TS?

taeold:
@Spencer-Easton Yes. As long the compiled JS code that's deployed to Firebase Functions doesn't contain top level await, things will worl.

jthegedus:
@taeold 

> @Spencer-Easton Yes. As long the compiled JS code that's deployed to Firebase Functions doesn't contain top level await, things will work.

Do the ESLint rules generated by `firebase init functions` capture top-level await? If not, can we configure this rule in the template?

taeold:
@jthegedus Good idea. We will probably want to update the template for both JS and TS to reflect ES module support in GCF.

DibyodyutiMondal:
So if we manually add in the latest version of functions framework as well as `"type":"modules"` right now, will it work?

taeold:
[...]

PR #3485, opened 2021-06-11
Add support ES modules for Firebase Functions

taeold:
To import modules packaged as ES module, we have to use `import` instead of `require`. Both the emulator and function deploy code hardcodes `require` to load user's function code and ends up throwing `ERR_REQUIRE_ESM` when given ES modules.

We will now try using `import` when `require` fails with an `ERR_REQUIRE_ESM` error. This is a bit lazy - we could've detected whether given module is ES module vs CommonJS and dispatch `require` vs `import` appropriately - but I think it simplifies the script while still being correct.

~~Note that this PR doesn't itself provide ES module support for Firebase Functions (https://github.com/firebase/firebase-tools/issues/2994); we need Google Cloud Function to provide ES module support first (https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/233).~~

Functions Framework now supports ES modules in 1.9.0! https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/301

**NOTE** ES Module support is available only for Node.js v13 and beyond. This leads to tricky situations where node version on dev's machine might support ES modules while the selected GCF runtime doesn't, and vice versa. Maybe a follow up item for early detection to reduce number of support tickets we might get because of this confusion?

Will fix once released: https://github.com/firebase/firebase-tools/issues/2994


src/deploy/functions/runtimes/node/triggerParser.js
```diff
@@ -7,7 +7,28 @@ var EXIT = function () {
...
 
-(function () {
+/**
+ * Dynamically load import function to prevent TypeScript from
+ * transpiling into a require.
+ *
+ * See https://github.com/microsoft/TypeScript/issues/43329.
+ */
+// eslint-disable-next-line @typescript-eslint/no-implied-eval
+var dynamicImport = new Function("modulePath", "return import(modulePath)");
```
joehan, line 17: Prefer const

src/deploy/functions/runtimes/node/triggerParser.js
```diff
@@ -7,7 +7,28 @@ var EXIT = function () {
...
+
+function loadModule(packageDir) {
+  try {
+    return Promise.resolve(require(packageDir));
+  } catch (e) {
+    if (e.code === "ERR_REQUIRE_ESM") {
+      var pkg = require.resolve(packageDir);
```
joehan, line 24: Prefer const

src/deploy/functions/runtimes/node/triggerParser.js
```diff
@@ -7,7 +7,28 @@ var EXIT = function () {
...
+/**
+ * Dynamically load import function to prevent TypeScript from
+ * transpiling into a require.
+ *
+ * See https://github.com/microsoft/TypeScript/issues/43329.
+ */
+// eslint-disable-next-line @typescript-eslint/no-implied-eval
+var dynamicImport = new Function("modulePath", "return import(modulePath)");
+
+function loadModule(packageDir) {
```
joehan, line 19: Should this be an async function?

joehan (approved):
Some nits, but this LGTM

src/deploy/functions/runtimes/node/triggerParser.js
taeold: Q: Does it matter that this script, for some reason, uses "old" JS style? (e.g. uses `var` pervasively instead of `let` or `const`). I thought it had to do with `const` not being supported on old NodeJS version, but it looks like `const` is supported from NodeJS 6 and up.

src/deploy/functions/runtimes/node/triggerParser.js
taeold: It can be? Does making the function `async` matter in this case?

inlined:
LGTM to me too. This will go into the firebase runtime code before it hits production, but the discovery is still quite useful.

src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -567,6 +576,7 @@ async function initializeFirebaseAdminStubs(frb: FunctionsRuntimeBundle): Promis
   // Stub the admin module in the require cache
   require.cache[adminResolution.resolution] = {
     exports: proxiedAdminModule,
+    path: path.dirname(adminResolution.resolution),
```
taeold, line 579: This needed to be done - otherwise, the dynamic import called failed with errors like this:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:363:5)
    at validateString (node:internal/validators:119:11)
    at Object.resolve (node:path:1098:7)
    at Function.Module._nodeModulePaths (node:internal/modules/cjs/loader:625:17)
    at cjsPreparseModuleExports (node:internal/modules/esm/translators:258:30)
    at Loader.commonjsStrategy (node:internal/modules/esm/translators:188:35)
```

@samtstern Think this is safe modification, but I'm 100% unfamiliar with this code path. Let me know if this is bad (I tested things out, and emulator continues to work normally).

samtstern: This seems safe to me.

samtstern (approved):

Issue #3492, opened 2021-06-15
Firebase emulators should impose HTTP/2 request/response size limits


remie:
### [REQUIRED] Environment info

**firebase-tools:** 8.14.1

**Platform:** macOS

### [REQUIRED] Test case

Write a function that proxies a large file (10MB+) by piping the data to the firebase cloud function response stream.

### [REQUIRED] Steps to reproduce

Call the test function and verify that your 10MB+ file is downloaded successfully.

### [REQUIRED] Expected behavior

The download should be cut-off at 10MB with a "response error" as this is the request/response size limit imposed by firebase quotas (see https://firebase.google.com/docs/functions/quotas)

### [REQUIRED] Actual behavior

This file will be proxied by the emulator in full without any request/response size limitation. This means the app will behave differently in local development (emulator) compared to the final product deployed to Firebase.

abeisgoat:
Thanks for this, we will take a look and either implement this or document the incongruity with production

weixifan:
I filed an internal b/211321188 to track this.

PR #3494, opened 2021-06-16
Enforce quota limit on response data (#3492)

remie:
### Description

Proposed fix for "Firebase emulators should impose HTTP/2 request/response size limits" (#3492)

This PR currently does not include any tests, as I'm not entirely confident that it is a good idea to emulate / test processing of a 10MB+ response stream during test execution. Please advice if you prefer me to include such test.

### Scenarios Tested

- Normal execution of https functions (non-binary / JSON responses)
- Execution of an https function while downloading a payload of <10MB
- Execution of an https function while downloading a payload of >10MB

### Sample Commands

N/A

### DIFF
--- a/src/emulator/functionsEmulatorRuntime.ts
+++ b/src/emulator/functionsEmulatorRuntime.ts
@@ -24,6 +24,12 @@ import * as fs from "fs";
 import { URL } from "url";
 import * as _ from "lodash";
 
+/*
+ * Firebase imposes quota limits on response data
+ * See https://firebase.google.com/docs/functions/quotas
+ */
+const MAX_DATA_LIMIT = 10000000;
+
 let triggers: EmulatedTriggerMap | undefined;
 let developerPkgJSON: PackageJSON | undefined;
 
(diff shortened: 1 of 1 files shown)


samtstern:
Since this came out of a Twitter conversation between you and @mbleigh I've assigned him as a reviewer.

src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -803,6 +809,56 @@ function rawBodySaver(req: express.Request, res: express.Response, buf: Buffer):
...
+      return false;
+    } else {
+      logDebug(`Total response size received: ${responseSize}`);
+      return true;
+    }
+  };
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  res.end = function (chunk: any, encoding?: any, cb?: () => void) {
+    if (chunk) {
```
remie, line 849: Another thing to consider, apart from the missing unit tests, is that the data limit is not imposed on a direct call to `res.end()` but only on a stream (using `res.write`). This is somewhat deliberate, as I would not expect anyone to directly flush 10MB+ on the response stream, but if the Google Cloud Functions production code does check the single chunk it might be useful to include it here as well.

samtstern: I am not sure about the prod behavior with `res.end()` but this is definitely an improvement either way!

samtstern (commented):
@remie I can see why you were hesitant to add unit tests but I think one test would be pretty helpful. We can add one where the function just generates random bytes until it gets to > 10MB and ensure that the middleware works.

You don't need to worry much about the test speed, the functions runtime tests are in a separate "integration" suite which won't slow down our normal unit tests.

Let me know if you have any trouble writing the test.

samtstern:
@remie let me know if you don't have time to add a unit test, I could helpl!

remie:
@samtstern I've been busy with some of our product releases, I was planning to circle back to this early next week. However, feel free to start / take over, I won't feel offended 😄

samtstern:
@remie no rush at all just checking in! Next week sounds great.

PR #3499, opened 2021-06-16
Fix issue where local extension is incorrectly inferred as published

huangjeff5:
Thank you for contributing to the Firebase community! Please fill out the form below.

Run the linter and test suite
==============================
Run `npm test` to make sure your changes compile properly and the tests all pass on your local machine. We've hooked up this repo with continuous integration to double check those things for you.

-->

### Description

Addressing bug where instances of local extensions are improperly treated as published extensions in the update scenario.

### DIFF
--- a/src/test/extensions/updateHelper.spec.ts
+++ b/src/test/extensions/updateHelper.spec.ts
@@ -109,7 +109,7 @@ const INSTANCE = {
   state: "ACTIVE",
   config: {
     name:
-      "projects/invader-zim/instances/image-resizer/configurations/95355951-397f-4821-a5c2-9c9788b2cc63",
+      "projects/invader-zim/instances/instance-of-official-ext/configurations/95355951-397f-4821-a5c2-9c9788b2cc63",
     createTime: "2019-05-19T00:20:10.416947Z",
     sourceId: "fake-official-source",
     sourceName: "projects/firebasemods/sources/fake-official-source",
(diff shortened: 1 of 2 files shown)


src/test/extensions/updateHelper.spec.ts
```diff
@@ -563,3 +581,71 @@ describe("inferUpdateSource", () => {
...
+    getInstanceStub.restore();
+  });
+
+  it("should return official extension as source origin", async () => {
+    registryEntryStub = sinon.stub(resolveSource, "resolveRegistryEntry");
+    registryEntryStub.resolves(REGISTRY_ENTRY);
+    isOfficialStub = sinon.stub(resolveSource, "isOfficialSource");
+    isOfficialStub.returns(true);
+    getInstanceStub = sinon.stub(extensionsApi, "getInstance").resolves(INSTANCE);
+    const result = await updateHelper.getExistingSourceOrigin(
```
joehan, line 602: Nit: Repo convention is to split unit tests into 3 chunks separated by newlines- setup, call functions under test, and expects. For example, in this test, you should add a newline between l601 and l602, and one between l607 and l608.

joehan (approved):
LGTM, with one small style nit

src/test/extensions/updateHelper.spec.ts
huangjeff5: Sounds good. Thanks for the heads up!

PR #3505, opened 2021-06-17
Generate JSON Schema for firebase.json

### LINKED ISSUE #1409: Document complete schema for `firebase.json`

samtstern:
### Description

Fixes #1409

TODO:
- [x] Actually validate the JSON files at runtime
- [x] Add test cases

Future work:
- [ ] Figure out how to host these JSON files somewhere
- [ ] Add `$schema` key to newly generated `firebase.json` files

### Scenarios Tested

See unit tests

### Sample Commands

N/A


jhuleatt:
One hosting option could be to directly fetch from github: `https://raw.githubusercontent.com/firebase/firebase-tools/{BRANCH}/schema/firebase-config.json`

For example:
```bash
curl https://raw.githubusercontent.com/firebase/firebase-tools/ss-fix-1409/schema/firebase-config.json
```

Not sure if Github provides any SLOs on speed or uptime for serving this way, though

src/firebaseConfigValidate.ts
```diff
@@ -0,0 +1,43 @@
...
+ * function. By doing this lazily we don't impose this I/O cost on those using
+ * the CLI as a Node module.
+ */
+export function getValidator(): ValidateFunction {
+  if (!_VALIDATOR) {
+    const schemaStr = fs.readFileSync(
+      path.resolve(__dirname, "../schema/firebase-config.json"),
+      "UTF-8"
+    );
+    const schema = JSON.parse(schemaStr);
```
jhuleatt, line 23: Node also supports:

```js
const schema = require(path.resolve(__dirname, "../schema/firebase-config.json"));
```

bkendall (approved):

src/firebaseConfigValidate.ts
samtstern: Yeah I think that will also put the JSON into the require cache though which seems a bit strange to me.

PR #3525, opened 2021-06-23
Ensure active EnvStore API before running functions:env:* commands

taeold:
Wire up `.before` action on `functions:env:*` family of commands that checks where the user's project has opted-in to have CF3 manage environment variables on its functions.

### Opt-in Experience
Running `functions:env:*` family of commands on a Firebase project that's yet to opt-in to have CF3 manage its environment variable will go through an opt-in flow:

```
$ firebase functions:env:get      
i  functions:env family of commands manages environment variables for your
Cloud Functions for Firebase. Learn more about this feature at
https://firebase.google.com/docs/functions/env.

⚠  By opting in to have Firebase manage your function's environment variables,
you may lose existing environment variables you've manually set up on your
functions. Also, any environment variables you subsequently set with the
gcloud command line tool or in the Google Cloud console will be overwritten
or removed.

? Would you like to have Firebase manage your functions' environment variables? (y/N) y 
```

If we detect that functions already has custom environment variables, we'll show additional messages that warn that those environment variables will be removed on subsequent deploys once the project has opted in:

```
...
⚠  If you opt in, the following environment variables will be deleted on next deploy:
	helloAsia(asia-northeast1): @@@@@2=111
	helloEu(europe-west1): FOO=bar
	helloUs(us-central1): FOO=foo, BAR=bar

To preserve these environment variable, run the following command after opt-in:
	firebase functions:env:add @@@@@2=111 FOO=foo BAR=bar

⚠  You have multiple values for keys FOO. Make sure you choose one value for each key before proceeding.

? Would you like to have Firebase manage your functions' environment variables? (y/N) y 
```

### Opt-in Checks
For opt-in check, we implement a bespoke method for activating the CF3-managed env var feature by keeping an extra state variable in the EnvStore Service. Specifically, we do this by setting up an internal EnvStore instance; if an EnvStore instance with the internal ID exist, we'll consider that the project has opted-in to the feature.


src/functions/ensureEnv.ts
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * Lookup existing cloud functions and collect user-defined env variables.
+ *
+ * If user's project hasn't activated the EnvStore API, we consider
+ * all non-default environment variable as "user-defined". The user-defined
+ * env is significant because they will be removed on the next deploy following
+ * EnvStore API activation.
+ */
+async function getUserEnvs(projectId: string): Promise<UserEnv[]> {
+  const have = await backend.existingBackend({ projectId, filters: [] }, /* forceRefresh= */ true);
```
inlined, line 80: 1. Will this be used during a CLI command that also has a context object? The context object is also used to store a cache.
2. If not, you don't need forceRefresh because this context object doesn't have the cache fields.

src/functions/ensureEnv.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  const fnEnvs: UserEnv[] = [];
+  for (const fn of have.cloudFunctions) {
+    // Filter out non CF3 function instances.
+    if (!deploymentTool.isFirebaseManaged(fn.labels || {})) {
```
inlined, line 85: I wonder if we should consider always setting labels.

src/functions/ensureEnv.ts
```diff
@@ -0,0 +1,161 @@
...
+      continue;
+    }
+    const uenvs: Record<string, string> = {};
+    const envs = fn.environmentVariables;
+    if (envs && Object.keys(envs).length > 1) {
```
inlined, line 90: Nit: I hate superflous blocks. You don't actually need the if block here:

```
const envs = fn.environmentVariables || {}
for (const [k, v] of Object.entries(envs)) {
  ...
}
```

src/functions/ensureEnv.ts
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * Lookup existing cloud functions and collect user-defined env variables.
+ *
+ * If user's project hasn't activated the EnvStore API, we consider
+ * all non-default environment variable as "user-defined". The user-defined
+ * env is significant because they will be removed on the next deploy following
+ * EnvStore API activation.
+ */
+async function getUserEnvs(projectId: string): Promise<UserEnv[]> {
```
inlined, line 79: Silly Q, but why did we make up a type for `UserEnv[]` instead of just `Record<string, backend.EnvironmentVariables>`

src/functions/ensureEnv.ts
```diff
@@ -0,0 +1,161 @@
...
+      "or in the Google Cloud console will be overwritten or removed.\n"
+  );
+
+  const userEnvs = await getUserEnvs(projectId);
+  if (userEnvs.length > 0) {
+    let msg =
+      "If you opt in, the following environment variables will be deleted on next deploy:\n";
+
+    const allKvs: Record<string, string> = {};
+    for (const { fnLabel, envs } of userEnvs) {
```
inlined, line 132: Not asking you to change anything, but I thought this was a fun time to try list comprehensions. If userEnvs was a Record<string, backend.EnvironmentVariables>. It's a fun programming challenge!

```
const functionEnvBreakdown = Object.entries(userEnvs).map(([label, envs]) => {
  const envList = Object.entries(envs).map(([k, v]) => clc.bold(`${k}=${v}`).join(", ");
  return `\t${label}: ${envList}`;
}).join("\n");

const allEnvs = Object.values(userEnvs).reduce((accum, next) => { return {...accum, ...next}}, {});
const allEnvsLabels = Object.entries(allEnvs).map(([k, v]) => `${k}=${v}`).join(" ");
```

This won't work though because of the much bigger issue: We don't handle the edge case where different functions have different values for the same environment variable.

src/functions/ensureEnv.ts
```diff
@@ -0,0 +1,161 @@
...
+import { logBullet, logWarning } from "../utils";
+import * as backend from "../deploy/functions/backend";
+import * as deploymentTool from "../deploymentTool";
+import * as envstore from "./envstore";
+import * as getProjectId from "../getProjectId";
+import * as helper from "../deploy/functions/functionsDeployHelper";
+
+const ENVSTORE_INTERNAL_ID = "firebase-functions-internal";
+const CONFIGSTORE_KEY = "envstore";
+const CONFIGSTORE_TTL = 1000 * 60 * 60 * 24; // 1 day
```
inlined, line 15: AFAICT, opting into envstore is a one-way process. Why do we TTL out the cache that it has been enabled?

src/functions/ensureEnv.ts
```diff
@@ -0,0 +1,161 @@
+import * as clc from "cli-color";
```
inlined, line 1: Why can't this just be in src/functions/env?

src/commands/functions-env-add.ts
```diff
@@ -1,6 +1,7 @@
 import * as clc from "cli-color";
 
 import { Command } from "../command";
+import { ensure as ensureEnvStore } from "../functions/ensureEnv";
```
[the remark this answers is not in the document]
joehan: +1 - we also already have another widely used function call ensure.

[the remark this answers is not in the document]
taeold: Renamed!

src/functions/ensureEnv.ts
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * Check if the EnvStore API is active.
+ *
+ * This is an bespoke method of checking whether user has opted in
+ * to activate the EnvStore API.
+ *
+ * We check for existance of non-emtpy ENV_ID=${ENVSTORE_INTERNAL_ID}
+ * which is created when user "enables" the EnvStore API.
+ */
+async function _check(projectId: string): Promise<boolean> {
```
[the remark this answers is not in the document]
taeold: Great suggestion. Refactored a bit.

src/functions/ensureEnv.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+/**
+ * Enable the EnvStore API.
+ *
+ * This is an bespoke method of "enabling" EnvStore API. We "enable" the
+ * EnvStore API setting up a non-empty EnvStore ENV_ID=${ENVSTORE_INTERNAL_ID}.
+ */
+export async function enable(projectId: string): Promise<void> {
+  await envstore.patchStore(projectId, ENVSTORE_INTERNAL_ID, { ENABLED: "1" });
```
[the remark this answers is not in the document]
taeold: Great catch.

src/functions/ensureEnv.ts
taeold: (2) - changing the passed value to `false`.

src/functions/ensureEnv.ts
taeold: Yes. `fn.labels || {}` is repeated in several places 😢.

src/functions/ensureEnv.ts
taeold: This code is a bit too subtle in that we are actually checking to see if there are non-default environment variables on the given function.

I believe there's only 1 environment variables that we consider default `FIREBASE_CONFIG`, so the code assumes that if there is more than 1 environment variable, the others must be user-defined.

I modified the code to be less verbose and hopefully more clear.

src/functions/ensureEnv.ts
taeold: Oh I didn't consider using `backend.EnvironmentVariables` - without it, the type definition looked verbose enough for me to just define it for 2-time uses. Inlined your shorter type definition instead.

src/functions/ensureEnv.ts
taeold: Woo these are fun code - going to shamelessly steal them!

As for the last bit - yes it won't detect conflicts, and we don't resolve it, but I'm hoping that it's obvious to the customer that they will have to manually resolve it themselves.

src/functions/ensureEnv.ts
taeold: I was expecting opt-out to be a thing sooner or later and wanted to design for it. Don't think hitting the server to check isn't very costly at the end of the day - do you consider this a YAGNI?

src/functions/ensureEnv.ts
taeold: Could be. I just preferred to keep the check&enable logic on a separate file. Is there a reason to prefer keeping the logic together?

src/functions/ensureEnv.ts
inlined: TTLing the cache is a reversible decision, whereas a permanent cache isn't. I think this is a YAGNI (I don't know that we want to support opting out), but I'll stand by your defensive coding.

src/functions/ensureEnv.ts
inlined: Largely because I'm trying to add some general structure to the project layout so that an engineer ramping up is guided to see and understand the codebase in units. I'd probably request this either be the same `env` file or that we have a separate `env/index.ts` and `env/ensure.ts`.

src/functions/ensureEnv.ts
```diff
@@ -124,27 +139,33 @@ export async function ensure(options: any): Promise<void> {
...
+    msg += Object.entries(userEnvs)
+      .map(([label, envs]) => {
+        const envList = Object.entries(envs)
+          .map(([k, v]) => clc.bold(`${k}=${v}`))
+          .join(", ");
+        return `\t${label}: ${envList}`;
+      })
+      .join("\n");
+
+    // Transform userEnvs into a string of form:
```
inlined, line 158: We're still not handling the case where two functions have different values for the same env. What do we even want to do here? Error out until they fix it? Just warn them in big bold letters?

src/functions/ensureEnv.ts
taeold: Thanks for clarifying. Will put things together in `env.ts`.

src/functions/ensureEnv.ts
taeold: Added a little warning!

inlined (approved):
Sorry, I've been barely working after giving myself a pretty bad concussion. LGTM

taeold:
This PR is obsolete.

PR #3529, opened 2021-06-23
Fixes a number of issues where ext:dev:emulator:* did not allow valid extension configurations

### LINKED ISSUE #2928: default PARAM isn't being injected during extension emulation

**firebase-tools:**
8.17.0

**Platform:**
macOS Catalina 10.15.4

### [REQUIRED] Test case

See steps to reproduce.

1. git clone `git@github.com:firebase/extensions.git`
2. run `npm i` from root of extension's directory.
3. cd `delete-user-data` directory
4. create `test-params.env` file. Add environment variables & insert project id as appropriate:
```env
LOCATION=europe-west2
STORAGE_PATHS={DEFAULT}/test/{UID}-test.png
SELECTED_DATABASE_INSTANCE=[INSERT YOUR OWN PROJECT ID]
FIRESTORE_DELETE_MODE=shallow
```
5. run CLI command with your project id: `firebase ext:dev:emulators:start --test-params=test-params.env --project=[INSERT YOUR OWN PROJECT ID]` 

The auth emulator and the cloud function emulator running.

The emulator command is erroring because the `${DATABASE_INSTANCE}'` param which is auto populated in the `extensions.yaml` during build time, is not being recognised by the emulator. 
```
Error: The extension.yaml has the following errors: 
  - Param SELECTED_DATABASE_INSTANCE has default value '${DATABASE_INSTANCE}', which does not pass the validationRegex ^([0-9a-z_.-]*)$
```

I did also try `${param:DATABASE_INSTANCE}` in the `extension.yaml` but I got the same error:

```
Error: The extension.yaml has the following errors: 
  - Param SELECTED_DATABASE_INSTANCE has default value '${param:DATABASE_INSTANCE}', which does not pass the validationRegex ^([0-9a-z_.-]*)$
```

joehan:
### Description
Fixes a bunch of issues where the param and spec validation for the extension emulator behaved badly:
- False-y default values were broken in multiple places. I went through and fixed these.
- Param validation was happening before variable substitution. I refactored buildOptions to make sure this happened first.
- Spec validation for params with defaults did not substitute autopopulated params into those defaults. We were also checking defaults against validationRegex. This meant that these regex checks would fail for some defaults that referenced autopopulated params. I removed this.

I also did some minor modernization and cleanup of these files.
Fixes #2928 
- ### Scenarios Tested
I tested a number of extension that Invertase reported were broken on the emulator. I also found some cases where the Storage and auth emulators weren't working correctly with extension - I'm going to try to fix that in a followup.


src/extensions/askUserForParam.ts
```diff
@@ -13,7 +13,7 @@ export function checkResponse(response: string, spec: Param): boolean {
   let valid = true;
   let responses: string[];
 
-  if (spec.required && !response) {
+  if (spec.required && (response == "" || response == undefined)) {
```
joehan, line 16: This is a little bit ugly, but necessary to cover both the case where a param is completely omitted from the .env file, and the case where a param is set to nothing in the .env file.

src/extensions/extensionsHelper.ts
```diff
@@ -138,21 +144,21 @@ export function substituteParams(original: object[], params: { [key: string]: st
...
-export function populateDefaultParams(paramVars: any, paramSpec: any): any {
+export function populateDefaultParams(paramVars: Record<string, string>, paramSpecs: Param[]): any {
   const newParams = paramVars;
 
-  _.forEach(paramSpec, (env) => {
-    if (!paramVars[env.param]) {
-      if (env.default) {
-        newParams[env.param] = env.default;
-      } else if (env.required) {
+  for (const param of paramSpecs) {
```
taeold, line 150: 💯

src/extensions/emulator/optionsHelper.ts
```diff
@@ -46,6 +43,27 @@ export async function buildOptions(options: any): Promise<any> {
...
 }
 
+// Exported for testing
+export async function getParams(options: any, extensionSpec: ExtensionSpec) {
+  const projectId = getProjectId(options, false);
+  const userParams = await paramHelper.readParamsFile(options.testParams);
+  const autoParams = {
+    PROJECT_ID: projectId,
+    EXT_INSTANCE_ID: extensionSpec.name,
+    DATABASE_INSTANCE: projectId,
```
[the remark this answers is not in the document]
taeold: Also, is `getFirebaseProjectParams` not appropriate for emulator?

src/extensions/emulator/optionsHelper.ts
```diff
@@ -46,6 +43,27 @@ export async function buildOptions(options: any): Promise<any> {
...
+    STORAGE_BUCKET: `${projectId}.appspot.com`,
+  };
+  const unsubbedParamsWithoutDefaults = Object.assign(autoParams, userParams);
+
+  const unsubbedParams = extensionsHelper.populateDefaultParams(
+    unsubbedParamsWithoutDefaults,
+    extensionSpec.params
+  );
+  // Run a substitution to support  params that reference auto populated params.
+  return extensionsHelper.substituteParams<Record<string, string>>(unsubbedParams, unsubbedParams);
```
taeold, line 64: nit*: I think this it would be slightly more clear if the arguments were `(unsubbedParams, autoParams)`

src/extensions/extensionsHelper.ts
```diff
@@ -61,6 +60,13 @@ export const EXTENSIONS_BUCKET_NAME = envOverride(
   "FIREBASE_EXTENSIONS_UPLOAD_BUCKET",
   "firebase-ext-eap-uploads"
 );
+const autopopulatedParamNames = [
```
taeold, line 63: Should this be in CONSTANT_CASE?

https://google.github.io/styleguide/tsguide.html#identifiers

(Looks like we go back and forth, but my reading of the style guide is that they all be in CONSTANT_CASE).

src/extensions/extensionsHelper.ts
```diff
@@ -117,7 +123,7 @@ export async function getFirebaseProjectParams(projectId: string): Promise<any>
  * @param params params to substitute the placeholders for
  * @return Resources object with substituted params
  */
-export function substituteParams(original: object[], params: { [key: string]: string }): Param[] {
+export function substituteParams<T>(original: T, params: Record<string, string>): T {
```
taeold, line 126: slick!

src/extensions/extensionsHelper.ts
```diff
@@ -163,28 +169,28 @@ export function populateDefaultParams(paramVars: any, paramSpec: any): any {
...
+  paramSpec: Param[]
 ): void {
-  if (_.size(envVars) > _.size(paramSpec)) {
-    const paramList = _.map(paramSpec, (param) => {
-      return param.param;
-    });
-    const misnamedParams = Object.keys(envVars).filter((key: any) => {
-      return !paramList.includes(key);
-    });
+  const paramNames = paramSpec.map((param) => {
```
taeold, line 175: *nit: Can be even shorter as

```
paramSpec.map(p => p.param);
```

src/extensions/extensionsHelper.ts
```diff
@@ -163,28 +169,28 @@ export function populateDefaultParams(paramVars: any, paramSpec: any): any {
...
+    return param.param;
+  });
+  const misnamedParams = Object.keys(envVars).filter((key: any) => {
```
taeold, line 178: Looks like `key: string` can replace `key: any` no?

src/extensions/extensionsHelper.ts
```diff
@@ -163,28 +169,28 @@ export function populateDefaultParams(paramVars: any, paramSpec: any): any {
...
+    return !paramNames.includes(key) && !autopopulatedParamNames.includes(key);
+  });
+  if (misnamedParams.length) {
     logger.info(
```
taeold, line 182: nit* not your change, but this sounds like a `logger.warn`!

src/extensions/extensionsHelper.ts
```diff
@@ -263,17 +269,6 @@ export function validateSpec(spec: any) {
           } cannot have options because it is type STRING`
         );
       }
-      if (
-        param.default &&
-        param.validationRegex &&
-        !RegExp(param.validationRegex).test(param.default)
-      ) {
-        errors.push(
```
taeold, line 271: Why is this gone? Do we do regex validation elsewhere?

src/test/extensions/emulator/optionsHelper.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+          param: "USER_PARAM2",
+        },
+      ];
+      readParamsFileStub.resolves({
+        USER_PARAM1: "val1",
+        USER_PARAM2: "val2",
+      });
+
+      expect(optionsHelper.getParams(testOptions, testSpec)).to.eventually.deep.eq(
+        Object.assign(
```
taeold, line 56: nit* I've grown fond of the spread operator:

```
{
  ...{
            USER_PARAM1: "val1",
            USER_PARAM2: "val2",
          },
  ...autoParams
}
```

Do you prefer `Object.assign`?

src/test/extensions/emulator/optionsHelper.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+        },
+        {
+          label: "param3",
+          param: "USER_PARAM3",
+        },
+      ];
+      readParamsFileStub.resolves({
+        USER_PARAM1: "${PROJECT_ID}-hello",
+        USER_PARAM2: "val2",
+        USER_PARAM3: "${USER_PARAM2}",
```
taeold, line 84: I don't think substitution from user param is allowed outside of the emulated environment - it seems like only the auto params are allowed to be substituted https://github.com/firebase/firebase-tools/blob/66447909ab6f2defd04b257b48193c70d778c7ea/src/extensions/askUserForParam.ts#L142.

Can we double check if this is what we want?

src/test/extensions/emulator/optionsHelper.spec.ts
joehan: I'll switch this over - i actually had a bug here for a bit caused by using Object.assign that wouldn;t have happened with a spread operator, which seems like a good enough reason to favor your option.

src/extensions/emulator/optionsHelper.ts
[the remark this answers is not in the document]
joehan: Nope, this is actually correct (see google3/firebase/mods/go/dm/instance_converter_test.go or google3/firebase/mods/go/dm/instance_converter.go on the backend). This is just the instance name, which is the first little chunk of the DATABASE_URL. This defaults to projectId.

GetFirebaseProjectParams is not emulator safe, since we need to let you emulate any projectid, real or fake, and that function calls getFirebaseConfig.

src/extensions/extensionsHelper.ts
joehan: Good point, I'll convert all of these

joehan: Ended up leaving logPrefix as is for now because its used in a dozen files, and i didn't want to scope creep quite that much

src/extensions/askUserForParam.ts
joehan: So, looking at this again, I think the real issue is that the typing here is pretty wonky. The YAML parser and .ENV parser we use cast strings that look like numbers or booleans to actual numbers or booleans, and we don't always cast those back to strings before passing them around. In this case, this was breaking if you ever tried to set a param to "false" or "0" - the parsers would cast those to  boolean false or number 0, and suddenly `!response == true`. 

If you think this is too hacky, I can try to fix these typing issues in this PR as well.

huangjeff5: It seems like we're trying to exclude the `response=false` from falling into this if block? Maybe we should explicitly specify that?

i.e. (!response && response !== false)

src/extensions/extensionsHelper.ts
huangjeff5: This is awesome - learned some new syntax today :)

src/extensions/extensionsHelper.ts
joehan: This was checking that default values pass the validation_regex. However, its really hard to reliably test this for default values that reference autopopulated params, since we don't know what those are. I considered changing this to inject AUTOPOPULATED_PARAM_PLACEHOLDERS in first, and checking based on that - however, that still could let through defaults that would fail validation on some projects.

Instead of keeping an imperfect validation for this infrequent case, I figured it was better to get rid of it entirely.

src/test/extensions/emulator/optionsHelper.spec.ts
joehan: I looked into the backend, and it seems like we do support user params that reference other user params (see l627 of instance_converter.go). This seems like a pretty big corner case, but I feel like we may as well try to match the backend behavior as closely as posisible. 

askUserForParams is a bit different, since at that point user params don't have any values (we're about to ask for them!), so we only need to substitute in the autpopulated values.

src/extensions/emulator/optionsHelper.ts
joehan: Addressed below, but this is intentional to cover cases where user params are referenced. I'll clarify this comment

src/extensions/askUserForParam.ts
taeold: YAML parsing is full of weird surprises..

src/extensions/emulator/optionsHelper.ts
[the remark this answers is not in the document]
taeold: oh ho. Good to know - thanks for clarifying.

src/test/extensions/emulator/optionsHelper.spec.ts
taeold: Thanks for checking! That's a tricky behavior - my guess is that no user is using this feature, but I agree following the backend behavior is the way to go.

taeold (approved):

huangjeff5 (approved):

PR #3532, opened 2021-06-25
feat: expand firebase config definition

jthegedus:
### Description

Expand the JSON Schema & Firebase Config type definition as discussed in https://github.com/firebase/firebase-tools/issues/1409#issuecomment-867637328

The large diff is because the final `.json` schema is generated.

I am not a daily TS user so there may be better ways of doing some of this.

**deploy hooks with hosting:**

I am unsure of `predeploy`/`postdeploy` hooks when referencing arrays of resources, ie: when `hosting` is an array: 

```js
{
	hosting: [{
		predeploy: "",
		postdeploy: "",
	}, {
		predeploy: "",
		postdeploy: "",
	}]
}
```

This seems logical as they cannot appear under the root of `hosting` as it's no longer an object, but I have never seen anyone actually use pre/post hooks in this type of a config.

Side note, I really wish the object or array of objects (`{...} | [{...}]`) stuff didn't exist. As with the `source | regex` in Hosting config. It makes validation much more difficult and the output config very large.

### Scenarios Tested

NA

### Sample Commands

NA


src/firebaseConfig.ts
```diff
@@ -45,7 +46,7 @@ export type HostingConfig = {
     function?: string;
     run?: {
       serviceId: string;
-      region?: string;
+      region?: "us-central1";
```
jthegedus, line 49: I am not sure about this, but currently Hosting rewrites can only rewrite to `us-central1` resources.

src/firebaseConfig.ts
```diff
@@ -61,18 +62,58 @@ export type HostingConfig = {
...
+export type FirestoreConfig = {
+  rules?: string;
+  indexes?: string;
+} & DeployAsset;
+
+export type FunctionsConfig = {
+  // TODO: Add types for "backend"
+  source?: string;
+  ignore?: string[];
+  runtime?: string;
```
jthegedus, line 107: Should this be an list of valid strings? Is there a central list this can be sourced from?

src/firebaseConfig.ts
```diff
@@ -5,35 +5,36 @@
 // 'npm run generate:json-schema' to regenerate the schema files.
 //
 
-export type DatabaseConfig =
+// base configs
+type DeployAsset = {
+  predeploy?: string | string[];
+  postdeploy?: string | string[];
+};
```
jthegedus, line 12: The documentation for pre/post hooks https://firebase.google.com/docs/cli#hooks only states
>You can set up predeploy and postdeploy hooks for any of the [assets that you can deploy](https://firebase.google.com/docs/cli#deployment)

src/firebaseConfig.ts
```diff
@@ -5,74 +5,121 @@
...
-  ignore?: string[];
-  predeploy?: string[];
+type HostingRewrites = ({ source: string } | { regex: string }) &
+  (
+    | { destination: string }
+    | { function: string }
+    | {
+        run: {
+          serviceId: string;
+          region?: string;
```
jthegedus, line 44: I am not sure about this value.

The [Hosting Cloud Functions docs state](https://firebase.google.com/docs/hosting/functions)
>Important: Firebase Hosting supports Cloud Functions in us-central1 only.

The [Hosting Cloud Run docs](https://firebase.google.com/docs/hosting/cloud-run) do not make a clarification, just that `us-central1` is the default

The [full-config docs](https://firebase.google.com/docs/hosting/full-config#rewrite-cloud-run-container) shows the entire list of available Cloud Run regions. Though when I tested this in February I couldn't get Hosting rewrites to work in non-`us-central1` regions.

Needless to say I am confused of the requirements here. I feel we should use string literals for this with all valid region options.

src/firebaseConfig.ts
jthegedus: I tried using:

```diff
+import * as runtimes from "./deploy/functions/runtimes";

export type FunctionsConfig = {
  // TODO: Add types for "backend"
  source?: string;
  ignore?: string[];
-  runtime?: string;
+  runtime?: runtimes.Runtime;
} & DeployAsset;
```

But the `npm run generate:json-schema` function produces errors because of the import:

<details>
<summary>errors</summary>s

```
/home/jthegedus/projects/fork-firebase-tools/node_modules/agent-base/dist/src/index.d.ts (2,8): Module '"net"' has no default export.
/home/jthegedus/projects/fork-firebase-tools/node_modules/agent-base/dist/src/index.d.ts (3,8): Module '"http"' has no default export.
/home/jthegedus/projects/fork-firebase-tools/node_modules/agent-base/dist/src/index.d.ts (4,8): Module '"https"' has no default export.
/home/jthegedus/projects/fork-firebase-tools/src/apiv2.ts (7,8): Module '"util"' has no default export.
/home/jthegedus/projects/fork-firebase-tools/src/defaultCredentials.ts (79,5): Type 'string | undefined' is not assignable to type 'undefined'.
  Type 'string' is not assignable to type 'undefined'.
/home/jthegedus/projects/fork-firebase-tools/src/defaultCredentials.ts (83,7): Type 'string' is not assignable to type 'undefined'.
/home/jthegedus/projects/fork-firebase-tools/src/gcp/proto.ts (32,20): Argument of type 'keyof T' is not assignable to parameter of type 'never'.
  Type 'string | number | symbol' is not assignable to type 'never'.
    Type 'string' is not assignable to type 'never'.
/home/jthegedus/projects/fork-firebase-tools/src/prompt.ts (37,20): Argument of type 'Question' is not assignable to parameter of type 'never'.
  Type 'ListQuestion<Answers>' is not assignable to type 'never'.
/home/jthegedus/projects/fork-firebase-tools/src/prompt.ts (53,15): No overload matches this call.
  Overload 1 of 2, '(o: {}): string[]', gave the following error.
    Argument of type 'unknown' is not assignable to parameter of type '{}'.
  Overload 2 of 2, '(o: object): string[]', gave the following error.
    Argument of type 'unknown' is not assignable to parameter of type 'object'.
/home/jthegedus/projects/fork-firebase-tools/src/prompt.ts (54,18): Object is of type 'unknown'.
/home/jthegedus/projects/fork-firebase-tools/node_modules/typescript-json-schema/dist/typescript-json-schema.js:1290
                throw new Error("No output definition. Probably caused by errors prior to this?");
```

</details>

jthegedus: Updated with the string literals of currently available runtimes

src/firebaseConfig.ts
jthegedus: Must have messed something up during my last testing because this seems to work now.

src/firebaseConfig.ts
```diff
@@ -5,74 +5,146 @@
 // 'npm run generate:json-schema' to regenerate the schema files.
 //
 
-export type DatabaseConfig =
+// Sourced from - https://firebase.google.com/docs/hosting/full-config#rewrite-cloud-run-container
+type CloudRunRegions =
```
samtstern, line 9: Keeping up with ever-expanding Cloud region launches is consistently hard for us. I know this is more typesafe but I would prefer to actually just use `"string"` for these fields and leave the validation to a runtime concern which can be validated via API calls. That way we don't have to constantly manually update this list as new regions launch.

src/firebaseConfig.ts
samtstern: That sounds reasonable to me! Nit: maybe `Deployable` is a better name?

src/firebaseConfig.ts
```diff
@@ -5,74 +5,146 @@
...
+  | "us-west1";
+
+type CloudFunctionRuntimes = "nodejs10" | "nodejs12" | "nodejs14";
+
+type DeployAsset = {
+  predeploy?: string | string[];
+  postdeploy?: string | string[];
+};
+
+type DatabaseOne = {
```
samtstern, line 39: nit: I think `DatabaseSingle` and `DatabaseMultiple` are better names.

src/firebaseConfig.ts
```diff
@@ -5,74 +5,146 @@
...
+  rules: string;
+} & DeployAsset;
+
+type DatabaseMany = ({
+  rules: string;
+} & (
   | {
-      rules?: string;
+      instance: string;
```
samtstern, line 47: This took me a while to read, maybe just inline the `rules` field? So:

```
{
      target?: string;
      instance: string;
      rules: string;
 } | {
      target: string;
      instance?: string;
      rules: string;
    }
```

src/firebaseConfig.ts
samtstern: Sames goes with all of the other `One`/`Many` types.

src/firebaseConfig.ts
```diff
@@ -5,74 +5,146 @@
...
+  DeployAsset & {
+    site?: string;
+    target?: string;
+  };
+
+type HostingMany = (HostingBase &
+  DeployAsset &
+  (
+    | {
+        site: string;
```
samtstern, line 106: Seems like in a lot of places you're using a `RequireAtLeastOne` type to make sure at least one field of a set is defined. Maybe we can use some TS wizardry for this?

https://docs.microsoft.com/en-us/javascript/api/@azure/keyvault-certificates/requireatleastone?view=azure-node-latest

It's a pretty nasty type definition but we can just define it once up top, link to the docs, and then simplify a lot of the other types.

samtstern (commented):
Thank you for taking this on! I have some comments. A lot of annoying naming nitpicks but I want to make sure this file is maintainable.

src/firebaseConfig.ts
jthegedus: I was trying to pull this from elsewhere in the code, as with runtimes list.

I know maintenance will be difficult if not shared within this codebase, I would urge this should be a goal to strive for in future.

src/firebaseConfig.ts
jthegedus: I used this type. It is only actually used in two places.

src/firebaseConfig.ts
jthegedus: I used the `RequireAtLeastOne` type mentioned below.

jthegedus:
Updated. I would urge in future we source the appropriate `runtime` & `region` values from elsewhere in this codebase as they will need to be hardcoded somewhere here surely?

samtstern (approved):

src/firebaseConfig.ts
samtstern: Thanks!

samtstern:
cc @joehan @inlined @taeold do we have a canonical list of Functions / Cloud Run regions in the codebase anywhere?

taeold:
@samtstern Found (this][https://github.com/firebase/firebase-functions/blob/2e718c74a382655a9756c0ff781278c0e8a9146b/src/function-configuration.ts#L4-L25] in `firebase-functions` repo, but don't see something in this repo.

I do think the omission is intentional though - based on [this comment](https://github.com/firebase/firebase-functions/pull/722#discussion_r451024460), I think the intention is to query the GCF backend to get the list of supported region during deploys instead of hard-coding the list in the codebase

(tangent: I learned that the CLI never actually validates whether the region is valid. It just fails in the process because some URL we create w/ the user-provided region points to a non-existent endpoint 🤷 )

samtstern:
@taeold I think that makes sense, this way we don't have to chase the regions list and update `firebase-tools` all the time. It also means that users of old CLI versions can deploy to new GCF regions, which is often useful. Thanks for confirming!

jthegedus:
It would be cool if we could fetch the external list in the schema file :thinking:

jthegedus:
>(tangent: I learned that the CLI never actually validates whether the region is valid. It just fails in the process because some URL we create w/ the user-provided region points to a non-existent endpoint shrug )

It should probably validate before say a joint resource deployment puts the environment in a partially deployed state (I haven't checked the codebase yet to see if this is already catered to)

taeold:
@jthegedus I think the deployment fails early enough to not put deployment states in a partial state.

The deployment fails late enough to have wasted time uploading user code and with somewhat hard-to-spot error message:

```
exports.helloWorld = functions
    .region("hello")
    .https.onRequest((request, response) => {
    });
```

```bash
 % firebase deploy --only functions
✔  functions: Finished running predeploy script.
i  functions: packaged functions (68.65 KB) for uploading
i  functions: creating Node.js 12 function helloWorld(hello)...
⚠  functions: failed to create function projects/chmods-info/locations/hello/functions/helloWorld
i  functions: cleaning up build files...

Functions deploy had errors with the following functions:
	helloWorld(hello)

To try redeploying those functions, run:
    firebase deploy --only "functions:helloWorld"

To continue deploying other features (such as database), run:
    firebase deploy --except functions

Error: Functions did not deploy properly.
```

Pretty late in the deployment without clear error message :(. It's only obvious if you use `--debug` flag.

PR #3538, opened 2021-06-30
Polish functions:env:* family of commands

taeold:
Follow up to https://github.com/firebase/firebase-tools/pull/3452 to address review comments made post-merge.

Highlights:

* Add messaging to warn users that functions must be deploy again for env var changes to take effect.
* Add more reserved environment vars, including the prefix `FIREBASE_`
* Add missing key validation for various env commands.
* Fix broken tests.

### DIFF
--- a/src/commands/functions-env-set.ts
+++ b/src/commands/functions-env-set.ts
@@ -60,7 +60,9 @@ export default new Command("functions:env:set [values...]")
     const envs = await fenv.setEnvs(projectId, setEnvs);
     logger.info(fenv.formatEnv(envs) + "\n");
     utils.logWarning(
-      "Please deploy your functions for the change to take effect by running firebase deploy --only functions"
+      "Please deploy your functions for the change to take effect by running " +
+        clc.bold("firebase deploy --only functions") +
+        "."
     );
     return envs;
   });
--- a/src/commands/functions-env-set.ts
+++ b/src/commands/functions-env-set.ts
@@ -59,6 +59,8 @@ export default new Command("functions:env:set [values...]")
 
     const envs = await fenv.setEnvs(projectId, setEnvs);
     logger.info(fenv.formatEnv(envs) + "\n");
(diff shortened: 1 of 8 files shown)


src/commands/functions-env-set.ts
```diff
@@ -56,6 +58,9 @@ export default new Command("functions:env:set [values...]")
     }
 
     const envs = await fenv.setEnvs(projectId, setEnvs);
-    logger.info(fenv.formatEnv(envs));
+    logger.info(fenv.formatEnv(envs) + "\n");
+    utils.logWarning(
+      "Please deploy your functions for the change to take effect by running firebase deploy --only functions"
```
joehan, line 63: Nit: We should bold the command here, like we do in https://github.com/firebase/firebase-tools/blob/66447909ab6f2defd04b257b48193c70d778c7ea/src/getProjectId.js#L34

src/functions/env.ts
```diff
@@ -111,6 +120,9 @@ export async function addEnvs(
   projectId: string,
   envs: Record<string, string>
 ): Promise<Record<string, string>> {
+  for (const [key] of Object.entries(envs)) {
```
joehan, line 123: Is there a reason to use Object.entries instead of Object.keys here? I feel like this is cleaner as:
```
for (const key of Object.keys(envs)) {
```

src/functions/env.ts
```diff
@@ -152,6 +165,9 @@ export async function setEnvs(
   // failure isn't too bad (users can simply try the command again). Regardless,
   // we should work with the EnvStore service team to develop an delete+create
   // transactionality.
+  for (const [key] of Object.entries(envs)) {
```
joehan, line 168: Same suggestion as above

joehan (approved):
Mostly LGTM, just a few small things

src/functions/env.ts
taeold: Good call. Will fix.

PR #3540, opened 2021-06-30
Add missing properties to Auth functions context

### LINKED ISSUE #3536: Incomplete event context when creating a new user

**firebase-tools:** 9.14.0

**Platform:** Mac

### [REQUIRED] Test case

Creating a new user (either using google, magic link, or anonymous) which in its turn trigger firebase functions are such:

image

Receives an incomplete auth event context as such:

image

The event is missing among other things the timestamp which I use to determine if the event is older than 10 minutes if so, stop to retry (fail-safe circuit for retriable functions).

Described above.

The event context should return eventId and timestamp.

EventId and timestamp are missing.

Debug log is attached.

Thanks
[log.txt](https://github.com/firebase/firebase-tools/files/6724251/log.txt)

samtstern:
### Description

Fixes #3536

### Scenarios Tested

I created a simple function and triggered it in the Emulators UI:

```js

exports.helloWorld = functions.auth.user().onCreate(async (user, ctx) => {
  console.log(JSON.stringify(ctx, undefined, 2));
});
```

The logs are correct:

```
>  {
>    "eventId": "d01d812f-03b5-4731-98c2-d539f4cb606f",
>    "eventType": "google.firebase.auth.user.create",
>    "resource": {
>      "service": "firebaseauth.googleapis.com",
>      "name": "projects/demo-project-12345"
>    },
>    "params": {},
>    "timestamp": "2021-06-30T13:45:49.161Z"
>  }
```

I also deployed a function to production to see the behavior and here is the `context` payload:

```
{
  "eventId":"2bcd79bb-dd0d-45b7-85f7-1a51b0b00e5a",
  "eventType":"google.firebase.auth.user.create",
  "notSupported":{},
  "resource":{
    "service":"firebaseauth.googleapis.com",
    "name":"projects/fir-dumpster"
  },
  "timestamp":"2021-07-01T16:51:17.729Z",
  "params":{}
} 
```

### Sample Commands

N/A

### DIFF
--- a/src/emulator/auth/cloudFunctions.ts
+++ b/src/emulator/auth/cloudFunctions.ts
@@ -1,3 +1,6 @@
+import * as uuid from "uuid";
+
+import { EventContext } from "firebase-functions";
 import { Client } from "../../apiv2";
 
 import { EmulatorInfo, Emulators } from "../types";
@@ -6,6 +9,10 @@ import { EmulatorRegistry } from "../registry";
 import { UserInfo } from "./state";
 
 type AuthCloudFunctionAction = "create" | "delete";
+
+type CreateEvent = EventContext & {
+  data: UserInfoPayload;
+};
 
 export class AuthCloudFunction {
   private logger = EmulatorLogger.forEmulator(Emulators.AUTH);
@@ -54,9 +61,16 @@ export class AuthCloudFunction {
   private createEventRequestBody(
     action: AuthCloudFunctionAction,
     userInfoPayload: UserInfoPayload
-  ): { eventType: string; data: UserInfoPayload } {
+  ): CreateEvent {
     return {
+      eventId: uuid.v4(),
       eventType: `providers/firebase.auth/eventTypes/user.${action}`,
+      resource: {
+        name: `projects/${this.projectId}`,
+        service: "firebaseauth.googleapis.com",
+      },
+      params: {},
+      timestamp: new Date().toISOString(),
       data: userInfoPayload,
     };
   }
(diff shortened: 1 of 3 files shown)


src/test/emulators/cloudFunctions.spec.ts
```diff
@@ -21,10 +21,17 @@ describe("cloudFunctions", () => {
       nock.cleanAll();
     });
 
-    it("should make a request to the functions emulator", async () => {
+    it.only("should make a request to the functions emulator", async () => {
```
yuchenshi, line 24:
```suggestion
    it("should make a request to the functions emulator", async () => {
```

src/emulator/auth/cloudFunctions.ts
```diff
@@ -54,9 +61,16 @@ export class AuthCloudFunction {
   private createEventRequestBody(
     action: AuthCloudFunctionAction,
     userInfoPayload: UserInfoPayload
-  ): { eventType: string; data: UserInfoPayload } {
+  ): CreateEvent {
     return {
+      eventId: uuid.v4(),
       eventType: `providers/firebase.auth/eventTypes/user.${action}`,
+      resource: {
+        name: `projects/${this.projectId}`,
```
yuchenshi, line 69: Shall this include the user id? I'm not sure what the production behavior actually is though

yuchenshi (approved):

src/emulator/auth/cloudFunctions.ts
samtstern: I was just following the docs here, but I will deploy a function quickly to confirm.

samtstern: Updated the description with a production payload.

PR #3547, opened 2021-07-03
Further improve function initialization

inlined:
The primary purpose of this change is to introduce the "discovery"
subpackage of "runtimes". This package is a common utility that all
runtimes will use to discover their backend spec based on the "container
contract" being finalized currently.

The container contract says that we must interact with customer code
using tools we'd have in a docker container: a fully-contained directory
of code, a "run" command, and environment variables. Backend specs can
be discovered as:

1. The file /backend.yaml at the root of the container
2. The response to http://localhost:${ADMIN_PORT}/backend.yaml when
   running the container's RUN command with ADMIN_PORT set to a free
   port.

The golang runtime fulfills #2 through two steps:

1. The Functions SDK includes the "codegen" package that reads the
   customer code and creates an "autogen" main package that runs
   the customer code as the Firebase runtime (skeleton implementation
   currently).
2. We then start up the Firebase runtime with ADMIN_PORT set and
   fetch /backend.yaml

After this + a minor fix to the template code, we have working Go
deploys!

I also moved gomod parsing into its own file to keep index.ts to a
minimal set of things that aren't easily unit tested. Unfortunately
the diff is largr than it needs to be now.


src/deploy/functions/runtimes/discovery/parsing.ts
```diff
@@ -0,0 +1,58 @@
...
+export function assertKeyTypes<T extends Object>(
+  prefix: string,
+  yaml: T | undefined,
+  schema: Record<keyof T, Type>
+) {
+  if (!yaml) {
+    return;
+  }
+  for (const [keyAsString, value] of Object.entries(yaml)) {
+    // I don't know why Object.entries(foo)[0] isn't type of keyof foo...
```
taeold, line 26: Interesting read at https://github.com/microsoft/TypeScript/pull/12253#issuecomment-263132208

src/test/deploy/functions/runtimes/discovery/parsing.spec.ts
```diff
@@ -0,0 +1,94 @@
+import { expect } from "chai";
+import { FirebaseError } from "../../../../../error";
+import * as parsing from "../../../../../deploy/functions/runtimes/discovery/parsing";
+
+describe("requireKeys", () => {
+  it("accepts found keys keys", () => {
```
joehan, line 6: repeated a word here?

src/test/deploy/functions/runtimes/discovery/parsing.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    undefined: undefined,
+    number: 0,
+    boolean: false,
+    string: "",
+    array: [],
+    object: {},
+  };
+  for (const type of tests) {
+    const schema = { [type]: type as parsing.Type };
+    for (const [testType, val] of Object.entries(values)) {
```
joehan, line 50: Very clever!

src/test/deploy/functions/runtimes/discovery/parsing.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+      it(`Handles a ${testType} when expecting a ${type}`, () => {
```
joehan, line 51: Nit: descriptions should start with lowercase

joehan (approved):
Some nitty things, but this mostly LGTM

src/deploy/functions/runtimes/discovery/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  yaml: any,
+  project: string,
+  region: string,
+  runtime: runtimes.Runtime
+): backend.Backend {
+  try {
+    if (!yaml.specVersion) {
+      throw new FirebaseError("Expect backend yaml to specify a version number");
+    }
+    if (yaml.specVersion === "v1alpha1") {
```
mbleigh, line 26: If we want to follow AIP [stability levels](https://google.aip.dev/181) this should be `v1alpha` not `v1alpha1`...do you intend for the spec version to match API version, or do you think they should be versioned separately? I can see argument for both.

src/deploy/functions/runtimes/discovery/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (err.code === "ENOENT") {
+      logger.debug("Could not find backend.yaml. Must use http discovery");
+    } else {
+      logger.debug("Unexpected error looking for backend.yaml file:", err);
+    }
+    return;
+  }
+
+  logger.debug("Found backend.yaml. Got spec:", text);
+  // TOODO(inlined): use a schema instead of manually checking everything or blindly trusting input.
```
mbleigh, line 55:
```suggestion
  // TODO(inlined): use a schema instead of manually checking everything or blindly trusting input.
```

src/deploy/functions/runtimes/discovery/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    }, timeout);
+  });
+
+  while (true) {
+    try {
+      res = await Promise.race([fetch(`http://localhost:${port}/backend.yaml`), timedOut]);
+      break;
+    } catch (err) {
+      // Allow us to wait until the server is listening.
+      if (/ECONNREFUSED/.exec(err?.message)) {
```
mbleigh, line 80: Probably want some amount of backoff or delay here, right?

src/deploy/functions/runtimes/discovery/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  let res: { text(): Promise<string> };
+  const timedOut = new Promise<never>((resolve, reject) => {
+    setTimeout(() => {
+      reject(new FirebaseError("User code failed to load. Cannot determine backend specification"));
+    }, timeout);
+  });
+
+  while (true) {
+    try {
+      res = await Promise.race([fetch(`http://localhost:${port}/backend.yaml`), timedOut]);
```
mbleigh, line 76: I don't think this will work if an `ECONNREFUSED` actually occurs, will it? Is it safe to use the same promise in multiple `Promise.race` calls?

I would think you'd need to race the timeout against an async function that does the looping retry logic...

src/deploy/functions/runtimes/discovery/index.ts
mbleigh: Also, is this the correct way to check for error types? Is this available instead as `err?.code` without having to regexp it? Even if not, I probably prefer `err?.message?.includes("ECONNREFUSED")` to inline regexp exec.

src/deploy/functions/runtimes/golang/index.ts
```diff
@@ -56,113 +63,119 @@ export async function tryCreateDelegate(
...
+    return Promise.resolve();
   }
 
-  build(): Promise<void> {
-    const res = spawn.sync("go", ["build"], {
+  async build(): Promise<void> {
+    try {
+      await promisify(fs.mkdir)(path.join(this.sourceDir, "autogen"));
+    } catch (err) {
+      if (!/EEXIST/.exec(err?.message)) {
```
mbleigh, line 83: Again if `err?.code` is available that feels better, and if not I think I prefer `err?.message?.includes("EEXIST")`

src/deploy/functions/runtimes/golang/index.ts
```diff
@@ -56,113 +63,119 @@ export async function tryCreateDelegate(
...
-  discoverSpec(
+  serve(
+    port: number,
+    adminPort: number,
+    envs: backend.EnvironmentVariables
+  ): Promise<() => Promise<void>> {
+    const childProcess = spawn("go", ["run", "./autogen"], {
+      env: {
+        ...envs,
+        ...process.env,
```
taeold, line 130: Is inheriting `process.env` intended? I would've thought that we won't want to do this in anticipation of a future where parsing is done on the server side.

src/deploy/functions/runtimes/golang/index.ts
```diff
@@ -56,113 +63,119 @@ export async function tryCreateDelegate(
...
+    }
+    await promisify(fs.writeFile)(
+      path.join(this.sourceDir, "autogen", "main.go"),
+      genBinary.stdout
+    );
+
+    // Now that we've created main.go, we will have new dependencies on the packages pulled in
+    // by /support/runtime. We need to run one more `go get`. We could have possibly frontloaded
+    // this work by adding a hand curated list of dependencies in project creation, but this
+    // would be more birttle.
```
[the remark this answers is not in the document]
inlined: Actually, when I tried a second time I was able to get everything working by go getting the subpackages in the firebase init code. Removed this for now. It bothers me how fragile this code can seem.

src/deploy/functions/runtimes/golang/index.ts
inlined: It turns out the magic env is `HOME`. Without `HOME` set all `go` commands fail with the error `error missing $GOPATH`...

PR #3551, opened 2021-07-04
emulator: Remove superfluous INFO log

anishkny:
### Description

Emulator does not log about the default being used for any other option. Why log about default used for `sourceDirectory`? Seems unnecessary.

### Scenarios Tested

Tested locally.

### Sample Commands

N/A


joehan (approved):
Seems reasonable to me

Issue #3573, opened 2021-07-13
Cannot start firebase emulator functions with node 14. ES Module import failure on windows.


paulvanj:
I am getting the following error:
```
i  emulators: Starting emulators: auth, functions, firestore, hosting, storage
!  functions: The following emulators are not running, calls to these services from the Functions emulator will affect production: database, pubsub
+  functions: Using node@14 from host.
i  firestore: Firestore Emulator logging to firestore-debug.log
i  hosting: Serving hosting files from: dist/apps/public-client
+  hosting: Local server: http://localhost:5005
i  ui: Emulator UI logging to ui-debug.log
i  functions: Watching "D:\Github\ros-firebase\dist\libs\cloud-functions" for Cloud Functions...
!  functions: Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'd:'
!  Your function was killed because it raised an unhandled error.
```

Is there anything I can do from my side to solve this problem, or am I forced to use commonjs as long as I am developing on windows? (I do not like the idea of adding `simport` throughout the entire dependency chain of my cloud-functions code.)

_Originally posted by @DibyodyutiMondal in https://github.com/firebase/firebase-tools/issues/2994#issuecomment-878954659_

taeold:
I have a proposed fix on branch [`dl-cf3-esm-windows`](https://github.com/firebase/firebase-tools/tree/dl-cf3-esm-windows). Can someone help me confirm the fix on windows?

```bash
> git clone https://github.com/firebase/firebase-tools.git
> cd firebase-tools
> git checkout origin/dl-cf3-esm-windows 
> npm i
> npm run build
> npm link # firebase cli should now point to locally built one
# cd to your functions project
> firebase functions:shell
```

paulvanj:
> I have a proposed fix on branch [`dl-cf3-esm-windows`](https://github.com/firebase/firebase-tools/tree/dl-cf3-esm-windows). Can someone help me confirm the fix on windows?
> 
> ```shell
> > git clone https://github.com/firebase/firebase-tools.git
> > cd firebase-tools
> > git checkout origin/dl-cf3-esm-windows 
> > npm i
> > npm run build
> > npm link # firebase cli should now point to locally built one
> # cd to your functions project
> > firebase functions:shell
> ```

[...]

PR #3574, opened 2021-07-14
Fix bug where functions in ES module failed to load on Windows

taeold:
Loading ES module using absolute path fails on Windows https://github.com/nodejs/node/issues/31710.

As suggested in mentioned issue, we use `url.pathToURL` to prefix absolute paths with `file://`.

Fixes https://github.com/firebase/firebase-tools/issues/3573.


samtstern (approved):

samtstern:
@taeold nice find!

PR #3575, opened 2021-07-15
Enable artifact registry

inlined:
GCF has moved to AR in their new API and we need to enable it on projects for any functions to successfully deploy.

### DIFF
--- a/src/deploy/functions/deploy.ts
+++ b/src/deploy/functions/deploy.ts
@@ -58,10 +58,10 @@ export async function deploy(
   try {
     const want = payload.functions!.backend;
     const uploads: Promise<void>[] = [];
-    if (want.cloudFunctions.some((fn) => fn.apiVersion === 1)) {
+    if (want.cloudFunctions.some((fn) => fn.platform === "gcfv1")) {
       uploads.push(uploadSourceV1(context));
     }
-    if (want.cloudFunctions.some((fn) => fn.apiVersion === 2)) {
+    if (want.cloudFunctions.some((fn) => fn.platform === "gcfv2")) {
       uploads.push(uploadSourceV2(context));
     }
     await Promise.all(uploads);
--- a/src/deploy/functions/deploy.ts
+++ b/src/deploy/functions/deploy.ts
@@ -4,13 +4,11 @@ import { setGracefulCleanup } from "tmp";
 import { checkHttpIam } from "./checkIam";
 import { functionsUploadRegion } from "../../api";
 import { logSuccess, logWarning } from "../../utils";
+import { Options } from "../../options";
 import * as args from "./args";
-import * as backend from "./backend";
 import * as fs from "fs";
 import * as gcs from "../../gcp/storage";
 import * as gcf from "../../gcp/cloudfunctions";
-import { Options } from "../../options";
-import { Config } from "../../config";
 import * as utils from "../../utils";
 
 const GCP_REGION = functionsUploadRegion;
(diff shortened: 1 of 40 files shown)


src/deploy/functions/deploy.ts
```diff
@@ -4,13 +4,11 @@ import { setGracefulCleanup } from "tmp";
 import { checkHttpIam } from "./checkIam";
 import { functionsUploadRegion } from "../../api";
 import { logSuccess, logWarning } from "../../utils";
+import { Options } from "../../options";
 import * as args from "./args";
-import * as backend from "./backend";
 import * as fs from "fs";
```
joehan, line 9: Not your change, but while you're here - move fs to the external imports block

src/deploy/functions/runtimes/node/parseTriggers.ts
```diff
@@ -151,9 +153,13 @@ export function addResourcesToBackend(
     }
 
     if (annotation.httpsTrigger) {
-      trigger = {
```
joehan, line 154: This change looks unrelated to AR - it looks ok to me but I just wanted to check that it should be here.

src/deploy/functions/prepare.ts
```diff
@@ -43,6 +43,12 @@ export async function prepare(
       /* silent=*/ true
     ),
     ensureCloudBuildEnabled(projectId),
+    ensureApiEnabled.ensure(
```
joehan, line 46: This will throw an error if it is unable to enable the API for some reason. We ought to catch this error so that this doesn't break users who can't use AR for some reason.

joehan (changes requested):

src/deploy/functions/deploy.ts
inlined: Done

src/deploy/functions/prepare.ts
inlined: using `then(()=>{}, ()=>{})` to catch errors. This will cause us to not see any error message if the enablement fails and the customer tries to dpeloy on GCFv2 though.

src/deploy/functions/runtimes/node/parseTriggers.ts
inlined: Sorry, I was originally writing a bunch of changes at once, but decided to be kinder. This is an intended but unrelated change. It suppresses a warning when deploying GCFv2 HTTPS functions.

src/deploy/functions/runtimes/node/parseTriggers.ts
```diff
@@ -150,9 +152,13 @@ export function addResourcesToBackend(
...
-        allowInsecure: true,
-      };
+      let allowInsecure: boolean;
+      if ("allowInsecure" in annotation.httpsTrigger) {
+        allowInsecure = !!annotation.httpsTrigger.allowInsecure;
+      } else {
+        allowInsecure = annotation.apiVersion != 2;
```
joehan, line 159: Looks like we need to change this to platform now?

src/deploy/functions/prepare.ts
joehan: Seems reasonable to me. To make our lives easier when debugging v2 issues, we could probably throw a logger.debug() in the catch. 
Alternatively, we could move this below `wantBackend` and actually throw the error if anything will be deployed on gcfv2.

inlined: Doing that. Making it not silent either. By putting it below it should throw with a clearer error that halts deploys rather than failing the functions deploys but succeeding the deploy command overall.

src/deploy/functions/runtimes/node/parseTriggers.ts
inlined: Yup. Changing.

joehan (approved):

PR #3580, opened 2021-07-16
Improve concurrency during functions deploys

inlined:
Though the method is a bit obtuse, this will enable concurrency.

We recognize concurrency as a field in the trigger annotations
and plumb it through to a ReplaceService call. Unfortuantely
GCFv2 doesn't have a direct concurrency option and Run doesn't
have a PATCH option so...

### DIFF
--- a/src/deploy/functions/tasks.ts
+++ b/src/deploy/functions/tasks.ts
@@ -116,6 +116,16 @@ export function createFunctionTask(
         params.errorHandler.record("warning", fnName, "make public", err.message);
       }
     }
+    if ("concurrency" in fn) {
+      if (fn.platform === "gcfv1") {
+        logger.debug("Cannot set concurrency for a v1 function");
+      } else {
+        await setConcurrency(
+          (cloudFunction as gcfV2.CloudFunction).serviceConfig.service!,
+          fn.concurrency || 1
+        );
+      }
+    }
   };
   return {
     run,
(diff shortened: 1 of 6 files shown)


src/deploy/functions/tasks.ts
```diff
@@ -196,6 +216,14 @@ export function deleteFunctionTask(params: TaskParams, fn: backend.FunctionSpec)
   };
 }
 
+export async function setConcurrency(name: string, concurrency: number) {
```
joehan, line 219: This probably doesn't need to be exported, unless you just want to do so for writing a unit test.

joehan (approved):
Looks fine to me once you `npm run format`
Like you mentioned in the sync, you could also add a unit test for setConcurrency, tho it simple enough that I won't block on it.

src/deploy/functions/tasks.ts
```diff
@@ -116,6 +116,16 @@ export function createFunctionTask(
         params.errorHandler.record("warning", fnName, "make public", err.message);
       }
     }
+    if ("concurrency" in fn) {
+      if (fn.platform === "gcfv1") {
+        logger.debug("Cannot set concurrency for a v1 function");
```
mbleigh, line 121: Feels like this ought to be a "this should never happen" scenario, because we should detect and error out a deployment if `concurrency` is set for a v1 function in the `prepare` stage, no?

src/deploy/functions/tasks.ts
```diff
@@ -116,6 +116,16 @@ export function createFunctionTask(
...
+      } else {
+        await setConcurrency(
+          (cloudFunction as gcfV2.CloudFunction).serviceConfig.service!,
+          fn.concurrency || 1
```
mbleigh, line 125: We should default to `n > 1` for creates on GCFv2 platform. My suggestion would be 50 or 80.

PR #3586, opened 2021-07-19
Delete GCR Images

colerogers:
### Description

An open-sesame command for purging the contents of the `gcf/` directory in every GCR subdomain. The `gcf/` directory is created by Google Cloud Functions on a function deployment. GCF places function images inside this directory, a user should not create custom images inside this directory. To limit misuse of this command, it requires the open-sesame preview flag and displays all artifacts that will be deleted during the confirmation step. The `force` flag is not enabled for the command.

### Scenarios Tested

- Invalid region
- Searching single region
- Searching multiple regions
- Searching multiple subdomains
- Delete single regions
- Delete multiple regions
- Delete multiple subdomains

### Sample Commands

`firebase --open-sesame deletegcfartifacts`

`firebase functions:deletegcfartifacts`

`firebase --close-sesame deletegcfartifacts`


src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +133,55 @@ export class ContainerRegistryCleaner {
   }
 }
 
+// top level function for purging artifacts
```
joehan, line 136: Prefer describing the behavior in a little more detail , and documenting any non-obvious behavior (like what happens if location is omitted). Something like:
```
/**
* Deletes all artifacts from GCF folder in GCR.
* @param projectId
* @param location: the subdomain to clean up. If omitted, will delete all locations.
*/
```

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +133,55 @@ export class ContainerRegistryCleaner {
...
+export async function purgeArtifacts(projectId: string, location?: string) {
+  // purge Container Registry
+  const purger = new ContainerRegistryPurger();
+  await purger.purge(projectId, location);
+
+  // TODO: purge Artifact Registry
+}
+
+export class ContainerRegistryPurger {
```
joehan, line 145: I don't think we should make this a class - theres no constructor, and there's only one real method, so this seems like it would be more readable as a plain old function.

src/previews.ts
```diff
@@ -18,6 +19,7 @@ export const previews: PreviewFlags = {
   rtdbmanagement: false,
   functionsv2: false,
   golang: false,
+  purgegcr: false,
```
joehan, line 22: This will all be subject to API council changes, but IMO, the open-sesame flag should just match whatever the name of the command is in this case.

src/test/deploy/functions/containerCleaner.spec.ts
```diff
@@ -209,3 +209,44 @@ describe("ContainerRegistryCleaner", () => {
...
+        children: ["region1", "region2", "region3", "region4"],
+        digests: [],
+        tags: [],
+      })
+    );
+    sinon.stub(purger, "getHelper").returns(stub);
+
+    await purger.purge("project");
+
+    expect(stub.rm).to.have.callCount(3); // 3 subdomain regions
```
joehan, line 250: Prefer a little more specificity here - you could check that it was specifically called with the ocrrect subdomains, for example.

joehan (commented):
Took a quick pass over the non user facing bits - please rerequest review from me once API is approved.

src/deploy/functions/containerCleaner.ts
colerogers: Added TSDoc comments to all functions

src/deploy/functions/containerCleaner.ts
colerogers: Changed it to an exportable function. Added a default parameter for DockerHelper for unit testing

src/previews.ts
colerogers: Change to deletegcfartifacts per API review

src/test/deploy/functions/containerCleaner.spec.ts
colerogers: Updated all unit tests

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +133,196 @@ export class ContainerRegistryCleaner {
...
+ * @param location: the specific region to search for artifacts. If omitted, will search all locations.
+ * @param dockerHelpers: a map of {@link SUBDOMAINS} to {@link DockerHelper}. If omitted, will use the default value and create each {@link DockerHelper} internally.
+ *
+ * @throws {@link FirebaseError}
+ * Thrown if the provided location is not a valid Google Cloud region or we fail to search subdomains.
+ */
+export async function listGCFArtifacts(
+  projectId: string,
+  locations?: string[],
+  dockerHelpers: Record<string, DockerHelper> = {}
```
colerogers, line 216: This is using a default value so we can inject a stub for unit testing.

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +133,196 @@ export class ContainerRegistryCleaner {
...
+ * @param location: the specific region to be clean up. If omitted, will delete all locations.
+ * @param dockerHelpers: a map of {@link SUBDOMAINS} to {@link DockerHelper}. If omitted, will use the default value and create each {@link DockerHelper} internally.
+ *
+ * @throws {@link FirebaseError}
+ * Thrown if the provided location is not a valid Google Cloud region or we fail to delete subdomains.
+ */
+export async function deleteGCFArtifacts(
+  projectId: string,
+  locations?: string[],
+  dockerHelpers: Record<string, DockerHelper> = {}
```
colerogers, line 293: Same here - This is using a default value so we can inject a stub for unit testing.

src/test/deploy/functions/containerCleaner.spec.ts
```diff
@@ -209,3 +209,362 @@ describe("ContainerRegistryCleaner", () => {
...
+
+    await containerCleaner.deleteGCFArtifacts(
+      "project",
+      ["us-central1", "us-west2", "europe-west1"],
+      helpers
+    );
+
+    expect(stubUS.rm).to.have.been.calledTwice;
+    expect(stubUS.rm).to.have.been.calledWith("project/gcf/us-central1");
+    expect(stubUS.rm).to.have.been.calledWith("project/gcf/us-central1");
```
inlined, line 540: extra line?

inlined (changes requested):
Lots of through testing and documentation! I think it's worth questioning whether the amount of detail we're giving customers before deletion is useful/too costly. I'd play with console.cloud.google.com/gcr to see what it looks like yourself. As mentioned in comments, "digests" are the actual images, "children" are folders for images, and "tags" are freeform references to those "digests" that pin them from being deleted.

src/commands/functions-deletegcfartifacts.ts
```diff
@@ -0,0 +1,47 @@
+import { Command } from "../command";
+import * as utils from "../utils";
+import * as getProjectId from "../getProjectId";
+import { listGCFArtifacts, deleteGCFArtifacts } from "../deploy/functions/containerCleaner";
+import { promptOnce } from "../prompt";
+import { requirePermissions } from "../requirePermissions";
+
+function getConfirmationMessage(artifacts: Set<string>): string {
```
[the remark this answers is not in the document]
colerogers: @inlined  I'm okay with listing out the full GCR path since that would allow the user to easily check for artifacts without looking up UUIDs. This requirement came from a discussion with @kaibolay on the API proposal. Kai ultimately approved the proposal, so we should probably get his opinion on this change.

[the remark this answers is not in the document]
kaibolay: I really like listing the full GCR path(s).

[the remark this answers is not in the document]
inlined: I guess it's time to disagree yet commit =D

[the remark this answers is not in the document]
kaibolay: Sorry, I should have been clearer. I don't think we disagree. 

What I meant to say is that I really like displaying the full directory path (without the UUIDs of the individual images).

Maybe we could display the number of images in those directories; e.g.:

```
You are about to delete all images in the following directories:

us.gcr.io/myproject/gcf/us-central1 (3 images)
us.gcr.io/myproject/gcf/us-west1 (1 image)

Are you sure?
```

But that's only a nice to have. The directory paths alone contain 99.9% of the important information.

src/commands/functions-deletegcfartifacts.ts
```diff
@@ -0,0 +1,47 @@
...
+        {
+          type: "confirm",
+          name: "force",
+          default: false,
+          message: getConfirmationMessage(gcfArtifacts),
+        },
+        options
+      );
+      if (!confirmDeletion) {
+        return utils.reject("Command aborted.", { exit: 1 });
```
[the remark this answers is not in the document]
colerogers: Changed

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +133,196 @@ export class ContainerRegistryCleaner {
...
+}
+
+/**
+ * Helper function to find all the subdomains for the supplied locations
+ * @param locations: an optional array of google cloud regions, if omitted will return all {@link SUBDOMAINS}
+ * @returns a set of subdomains
+ */
+function getSubdomainsFromLocations(locations?: string[]): Set<string> {
+  if (locations === undefined) {
+    return new Set(SUBDOMAINS);
```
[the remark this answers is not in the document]
colerogers: Removed this function, refactored for easier readability

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +133,196 @@ export class ContainerRegistryCleaner {
...
+  }
+  const subdomains = new Set<string>();
```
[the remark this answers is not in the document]
colerogers: Removed this function, refactored for easier readability

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +133,196 @@ export class ContainerRegistryCleaner {
...
+ * Helper function to find all relative Google Cloud Registry paths for use with searching for GCF artifacts
+ * @param projectId: the current project that contains GCF artifacts
+ * @param subdomain: the current subdomain that we're searching or deleting from
+ * @param locations: an optional array that contains specific Google Regions. If ommited, will return the root path.
+ * @returns an array of relative GCR paths.
+ */
+function findGCFPaths(projectId: string, subdomain: string, locations?: string[]): string[] {
+  const paths: string[] = [];
+  if (locations) {
+    for (const location of locations) {
```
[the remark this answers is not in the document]
colerogers: Removed this function, refactored for easier readability

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +133,196 @@ export class ContainerRegistryCleaner {
...
+ *
+ * @throws {@link FirebaseError}
+ * Thrown if the provided location is not a valid Google Cloud region or we fail to search subdomains.
+ */
+export async function listGCFArtifacts(
+  projectId: string,
+  locations?: string[],
+  dockerHelpers: Record<string, DockerHelper> = {}
+): Promise<Set<string>> {
+  if (locations && locations.find((location) => SUBDOMAIN_MAPPING[location] === undefined)) {
```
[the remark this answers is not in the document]
colerogers: Changed to `!SUBDOMAIN_MAPPING[location]`

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +133,196 @@ export class ContainerRegistryCleaner {
...
+}
+
+/**
+ * Helper function to find all relative Google Cloud Registry paths for use with searching for GCF artifacts
+ * @param projectId: the current project that contains GCF artifacts
+ * @param subdomain: the current subdomain that we're searching or deleting from
+ * @param locations: an optional array that contains specific Google Regions. If ommited, will return the root path.
+ * @returns an array of relative GCR paths.
+ */
+function findGCFPaths(projectId: string, subdomain: string, locations?: string[]): string[] {
```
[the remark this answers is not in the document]
colerogers: Fixed

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +133,196 @@ export class ContainerRegistryCleaner {
...
+  for (const subdomain of subdomains) {
+    if (dockerHelpers[subdomain] === undefined) {
+      const origin = `https://${subdomain}.${containerRegistryDomain}`;
+      dockerHelpers[subdomain] = new DockerHelper(origin);
+    }
+
+    const purgePaths = findGCFPaths(projectId, subdomain, locations);
+    for (const path of purgePaths) {
+      try {
+        await dockerHelpers[subdomain].rm(path);
```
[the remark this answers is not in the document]
colerogers: Changed

src/test/deploy/functions/containerCleaner.spec.ts
```diff
@@ -209,3 +209,362 @@ describe("ContainerRegistryCleaner", () => {
...
+  it("should throw an error on invalid location", () => {
+    expect(() => containerCleaner.listGCFArtifacts("project", ["invalid"])).to.throw;
+  });
+
+  it("should list artifacts, single location param", async () => {
+    const stubDH = sinon.createStubInstance(containerCleaner.DockerHelper);
+    stubDH.ls.withArgs("project/gcf/us-central1").returns(UUIDS);
+    stubDH.ls.withArgs("project/gcf/us-central1/uuid1").returns(CHILD_TAGS);
+    stubDH.ls.withArgs("project/gcf/us-central1/uuid2").returns(CHILD_TAGS);
+    stubDH.ls.withArgs("project/gcf/us-central1/uuid3").returns(CHILD_EMPTY);
```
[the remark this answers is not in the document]
colerogers: Fixed

src/test/deploy/functions/containerCleaner.spec.ts
```diff
@@ -209,3 +209,362 @@ describe("ContainerRegistryCleaner", () => {
...
+    const helpers: Record<string, containerCleaner.DockerHelper> = {};
+    helpers["us"] = stubDH;
```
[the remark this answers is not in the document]
colerogers: Fixed

src/test/deploy/functions/containerCleaner.spec.ts
colerogers: Yep, fixed.

src/commands/functions-deletegcfartifacts.ts
```diff
@@ -0,0 +1,47 @@
...
+      }
+      await deleteGCFArtifacts(projectId, regions);
+    } catch (err) {
+      utils.reject(err);
```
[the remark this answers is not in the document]
colerogers: fixed

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +133,196 @@ export class ContainerRegistryCleaner {
...
+    for (const path of searchPaths) {
+      try {
+        const rootChildren = (await dockerHelpers[subdomain].ls(path)).children;
+        if (rootChildren.length === 0) {
+          continue;
+        }
+        // partition children into subsets that are artifacts and location dirs
+        const locationDirs = rootChildren.filter((child) => SUBDOMAIN_MAPPING[child] !== undefined);
+        const artifactCandidates = rootChildren.filter(
+          (child) => SUBDOMAIN_MAPPING[child] === undefined
```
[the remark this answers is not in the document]
colerogers: The linter seems to place longer lambdas on their own line. I went through and changed them to blocks.

src/test/deploy/functions/containerCleaner.spec.ts
```diff
@@ -209,3 +209,362 @@ describe("ContainerRegistryCleaner", () => {
...
+    tags: [],
+  });
+
+  const LOCATIONS_EU = Promise.resolve({
+    children: ["europe-west1", "europe-central2"],
+    digests: [],
+    tags: [],
+  });
+
+  const MIXED = Promise.resolve({
```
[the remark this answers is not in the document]
colerogers: Removed

src/test/deploy/functions/containerCleaner.spec.ts
```diff
@@ -209,3 +209,362 @@ describe("ContainerRegistryCleaner", () => {
     expect(stub.rm).to.not.have.been.called;
   });
 });
+
+describe("listGCFArtifacts", () => {
+  const UUIDS = Promise.resolve({
+    children: ["uuid1", "uuid2", "uuid3"],
+    digests: [],
```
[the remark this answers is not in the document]
colerogers: Changed the tests and added in digests & tags.

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +133,196 @@ export class ContainerRegistryCleaner {
...
+  dockerHelpers: Record<string, DockerHelper> = {}
+): Promise<Set<string>> {
+  if (locations && locations.find((location) => SUBDOMAIN_MAPPING[location] === undefined)) {
+    throw new FirebaseError("Invalid region supplied.");
+  }
+
+  const artifacts = new Set<string>();
+  const failedSubdomains = new Set<string>();
+  const subdomains = getSubdomainsFromLocations(locations);
+  for (const subdomain of subdomains) {
```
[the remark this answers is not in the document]
colerogers: Thanks for writing this all out! I've changed the list function to use  what you wrote here. I modified the parallel work to go over the subdomains since we only need to run a `ls` at the root of the `gcf` directory. I also changed to helper to accept a subdomain instead of a location.

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +133,196 @@ export class ContainerRegistryCleaner {
   }
 }
 
+/**
+ * Helper function to flatten all docker tags from the given path to a comma separated string
```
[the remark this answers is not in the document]
colerogers: After refactoring the code, I ended up removing all of these helper functions. I kept the TSDocs on the two main exportable functions.

src/deploy/functions/containerCleaner.ts
```diff
@@ -10,10 +10,11 @@ import { logger } from "../../logger";
 import * as docker from "../../gcp/docker";
 import * as backend from "./backend";
 import * as utils from "../../utils";
+import { FirebaseError } from "../../error";
 
 // A flattening of container_registry_hosts and
 // region_multiregion_map from regionconfig.borg
-const SUBDOMAIN_MAPPING: Record<string, string> = {
+export const SUBDOMAIN_MAPPING: Record<string, string> = {
```
colerogers, line 17: I added an export here so I can use the actual object in unit testing for deleting every region. I didn't want to create a 2nd source of truth and have failures if this ever gets updated

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +129,112 @@ export class ContainerRegistryCleaner {
...
+  }
+  const invalidRegion = locations.find((loc) => !SUBDOMAIN_MAPPING[loc]);
+  if (invalidRegion) {
+    throw new FirebaseError(`Invalid region ${invalidRegion} supplied`);
+  }
+  const locationsSet = new Set(locations); // for quick lookup
+  const subdomains = new Set(Object.values(SUBDOMAIN_MAPPING));
+  const failedSubdomains = new Set<string>();
+
+  const listAll = [...subdomains].map((subdomain) => {
```
colerogers, line 164: Looping through the subdomains here and filtering later since that limits us to calling the docker API only 3 times.

src/commands/functions-deletegcfartifacts.ts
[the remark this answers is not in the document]
colerogers: I've changed it to display the full directory path. I explored adding in the image counts but that is a recursive operation that calls the docker API for each directory and sub directories. We would probably get rate limited and could potently run into a similar issue as [this](https://github.com/firebase/firebase-tools/issues/3639) where the command fails.

Thomas also pointed out that some customers that are blocked on this issue have a massive amount of images. If those customers were to run this command then they'd likely not make much progress waiting for the list to load or get fully blocked when running this command.

src/commands/functions-deletegcfartifacts.ts
```diff
@@ -0,0 +1,53 @@
...
+  let message = "You are about to delete all images in the following directories:\n\n";
+  for (const path of paths) {
+    message += `${path}\n`;
+  }
+  message += "\nAre you sure?\n";
+  return message;
+}
+
+export default new Command("functions:deletegcfartifacts")
+  .description("Deletes all artifacts created by Google Cloud Functions on Google Cloud Registry.")
```
inlined, line 23: on Google Container Registry

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +129,112 @@ export class ContainerRegistryCleaner {
...
+  dockerHelpers: Record<string, DockerHelper> = {}
+): Promise<void> {
+  if (!locations) {
+    locations = Object.keys(SUBDOMAIN_MAPPING);
+  }
+  const invalidRegion = locations.find((loc) => !SUBDOMAIN_MAPPING[loc]);
+  if (invalidRegion) {
+    throw new FirebaseError(`Invalid region ${invalidRegion} supplied`);
+  }
+  const subdomains = new Set(Object.values(SUBDOMAIN_MAPPING));
```
inlined, line 217: Am I crazy here or are we creating a set from an array and then only using that set to create another array below (for map)

src/deploy/functions/containerCleaner.ts
```diff
@@ -128,6 +129,112 @@ export class ContainerRegistryCleaner {
...
+  const failedSubdomains = new Set<string>();
```
inlined, line 218: arrays are idiomatic in JavaScript and you're only appending to the end. You can use an array and just use push(). Array insertion into an end is O(1) when using the doubling method.

src/test/deploy/functions/containerCleaner.spec.ts
```diff
@@ -209,3 +209,212 @@ describe("ContainerRegistryCleaner", () => {
...
+  });
+
+  const EMPTY = Promise.resolve({
+    children: [],
+    digests: [],
+    tags: [],
+  });
+
+  it("should throw an error on invalid location", () => {
+    expect(() => containerCleaner.listGcfPaths("project", ["invalid"])).to.throw;
```
inlined, line 239: Nit: checking for it throwing is fragile because it only passes if the error is thrown before any await command. expect it to reject instead.

src/test/deploy/functions/containerCleaner.spec.ts
```diff
@@ -209,3 +209,212 @@ describe("ContainerRegistryCleaner", () => {
...
+  it("should list paths, single location param", async () => {
+    const stub = sinon.createStubInstance(containerCleaner.DockerHelper);
+    stub.ls.withArgs("project/gcf").returns(LOCATIONS_US);
+    const stubEmpty = sinon.createStubInstance(containerCleaner.DockerHelper);
+    stubEmpty.ls.withArgs("project/gcf").returns(EMPTY);
+    const helpers = { us: stub, eu: stubEmpty, asia: stubEmpty };
+
+    const paths = await containerCleaner.listGcfPaths("project", ["us-central1"], helpers);
+
+    expect(paths.length).to.eq(1);
```
inlined, line 262: FYI. You can use `deep` here (though you'd have to use `sort` if the list has multiple elements.

```
expect(paths).to.deep.equal(["us.gcr.io/projects/gcf/us-central1"]);
```

src/test/deploy/functions/containerCleaner.spec.ts
```diff
@@ -209,3 +209,212 @@ describe("ContainerRegistryCleaner", () => {
...
+
+  it("should throw an error when subdomains fail search", () => {
+    const stub = sinon.createStubInstance(containerCleaner.DockerHelper);
+    const helpers = {
+      us: stub,
+      eu: stub,
+      asia: stub,
+    };
+
+    expect(async () => await containerCleaner.deleteGcfArtifacts("project", undefined, helpers)).to
```
inlined, line 346:
```
await expect(() => containerCleaner.deleteGcfArtifacts("project", undefined, helpers)).to.reject
```

(make your method async)

inlined (approved):

src/commands/functions-deletegcfartifacts.ts
colerogers: whoops, fixed

src/deploy/functions/containerCleaner.ts
colerogers: I was using a set to remove duplicate values and used an array to keep consistency with the other delete function. I just switched to the `for ... of ...` syntax and `push` each `Stat` into a `Promise` array.

src/deploy/functions/containerCleaner.ts
colerogers: Changed to an array.

src/test/deploy/functions/containerCleaner.spec.ts
colerogers: Changed to use `deep` in all occurrences.

src/test/deploy/functions/containerCleaner.spec.ts
colerogers: Fixed

src/test/deploy/functions/containerCleaner.spec.ts
colerogers: Fixed

PR #3592, opened 2021-07-21
typescript filterTargets and add tests

bkendall:
### Description

Typescript'd `filterTargets` and added some tests before realizing that it doesn't actually fix the issue I was looking at. So, just saving this work :)

### Scenarios Tested

Deploying with `--only` flags.

### DIFF
--- a/src/filterTargets.ts
+++ b/src/filterTargets.ts
@@ -20,7 +20,7 @@ export function filterTargets(options: Options, validTargets: string[]): string[
       })
     );
   } else if (options.except) {
-    targets = difference(targets, (options.except || "").split(","));
+    targets = difference(targets, options.except.split(","));
   }
   if (targets.length === 0) {
     let msg = "Cannot understand what targets to deploy/serve.";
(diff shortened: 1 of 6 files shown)


src/filterTargets.ts
```diff
@@ -1,23 +1,27 @@
...
+    targets = intersection(
       targets,
-      options.only.split(",").map(function (opt) {
+      options.only.split(",").map((opt: string) => {
         return opt.split(":")[0];
       })
     );
   } else if (options.except) {
-    targets = _.difference(targets, options.except.split(","));
+    targets = difference(targets, (options.except || "").split(","));
```
joehan, line 23: Since this is inside a `if (options.except)`, this `|| ""` seems redundant.

joehan (approved):
LGTM after one fix

src/filterTargets.ts
bkendall: I think something was yelling at me before about it maybe being undefined. Looks like I can remove it now. Thanks.

PR #3596, opened 2021-07-22
fix: print functions source directory in error message when no language was detected

andipaetzold:
### Description

When running `firebase deploy --only functions` and no language could be detected, the following message is displayed even though a `functions.source` is configured in `firebase.json`:

```
Error: Could not detect language for functions at
```

The actual location where `firebase-tools` tried to detect the language is not printed. That's because the directory is passed into `FirebaseError` as the second argument which is `options` and different to e.g. `console.log` the arguments are not concatenated.

With this PR the `firebase-tools` correctly print the source directory. In my case:

```
Error: Could not detect language for functions at dist
```

Depending on your preferred error message style, the path could be wrapped in quotes.

### Scenarios Tested

I only ran `firebase deploy --only functions` and the message above was printed.

### DIFF
--- a/src/deploy/functions/runtimes/index.ts
+++ b/src/deploy/functions/runtimes/index.ts
@@ -128,8 +128,5 @@ export async function getRuntimeDelegate(
     }
   }
 
-  throw new FirebaseError(
-    "Could not detect language for functions at",
-    options.config.get("functions.source")
-  );
+  throw new FirebaseError("Could not detect language for functions at " + sourceDirName);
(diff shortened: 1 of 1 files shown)


inlined:
It's probably better to just have a simple check if the directory exists and print an error message in that case.

inlined:
Thank you for the contribution. I'm not sure why we did the same change ourselves instead of taking your PR, but the bug is fixed. Closing this version.

PR #3610, opened 2021-07-27
fix:  auth_time emulator behavior should match production (auth_time is now set to user's last sign in time). (#3608)

### LINKED ISSUE #3608: auth_time doesn't match user's lastSignInTime (always matches iat instead)
### [REQUIRED] Describe your environment

- Operating System version: Windows 10 Home Version 20H2, Build 19042.1052
- Browser version: Chrome Version 91.0.4472.164 (64-bit)
- Firebase SDK version: firebase 9.0.0-beta.6
- Firebase Product: Firestore Emulator, Database Emulator, Storage Emulator

### [REQUIRED] Describe the problem

According to the documentation, auth_time should match the user's last sign in time, and iat should match the time this token was issued. This means auth_time should only be updated once, when the user signed in, and iat should change every hour when a new token is issued. This is not the case however. Despite what the documentation says, auth_time changes every hour at the same time as iat. This means auth_time always matches iat, and can't be used in security rules to check the time the user last signed in

https://firebase.google.com/docs/reference/admin/node/admin.auth.DecodedIdToken#auth_time

#### Steps to reproduce:
1. Create the following firestore.rules file
```
rules_version = '2';
service cloud.firestore {
    allow read, write: if request.auth != null && request.auth.token.auth_time < request.auth.token.iat;
  }
}
```
2. Use ```npm run serve``` to run the new rules in your local emulators
3. Sign in as a user on your client app.

#### Expected Result:

All reads and writes should fail for about the first hour. But after an hour has passed (and firebase has automatically issued the user a new token) the reads and writes should start succeeding

#### Actual Result:

All reads and writes fail forever. This is because auth_time always changes in lock-step with iat, despite the documentation saying that auth_time should remain stable after login and only iat should change every hour.

lovelle-cardoso:
### Description

Fixes #3608 so that emulator sets auth_time the same way production does (auth_time should match lastLoginAt in seconds)

### Scenarios Tested

Added unit test to src/test/emulators/auth/misc.spec.ts called `"should populate auth_time to match lastLoginAt (in seconds since epoch)"`

This unit test...
1. Registers a user
2. Exchanges a refresh token for a new token
3. Checks that the new token's auth_time matches the user's lastLoginAt in seconds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,3 @@
+- Makes auth_time emulator behavior match production (auth_time is now set to user's last sign in time). (#3608)
 - Fixes Auth Emulator errors when importing many users. (#3577)
 - Fixes support for `--except` flag when used for deploying Hosting. (#3397)
--- a/src/emulator/auth/operations.ts
+++ b/src/emulator/auth/operations.ts
@@ -1726,7 +1726,11 @@ function generateJwt(
     // This field is only set for anonymous sign-in but not for any other
     // provider (such as email or Google) in production. Let's match that.
     provider_id: signInProvider === "anonymous" ? signInProvider : undefined,
-    auth_time: toUnixTimestamp(new Date()),
+    auth_time: user.lastLoginAt
+      ? toUnixTimestamp(new Date(user.lastLoginAt))
+      : user.lastRefreshAt
+      ? toUnixTimestamp(new Date(user.lastRefreshAt))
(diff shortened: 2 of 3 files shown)


lovelle-cardoso:
@googlebot I signed it!

lovelle-cardoso:
My commit author wasn't correct. So I fixed it here: https://github.com/firebase/firebase-tools/pull/3611

PR #3611, opened 2021-07-27
fix: emulator auth_time (#3608)

### LINKED ISSUE #3608: auth_time doesn't match user's lastSignInTime (always matches iat instead)

lovelle-cardoso:
### Description

Fixes #3608 so that emulator sets auth_time the same way production does (auth_time should match lastLoginAt in seconds)

### Scenarios Tested

Added unit test to src/test/emulators/auth/misc.spec.ts called `"should populate auth_time to match lastLoginAt (in seconds since epoch)"`

This unit test...
1. Registers a user
2. Exchanges a refresh token for a new token
3. Checks that the new token's auth_time matches the user's lastLoginAt in seconds.

### DIFF
--- a/src/emulator/auth/operations.ts
+++ b/src/emulator/auth/operations.ts
@@ -1726,11 +1726,12 @@ function generateJwt(
     // This field is only set for anonymous sign-in but not for any other
     // provider (such as email or Google) in production. Let's match that.
     provider_id: signInProvider === "anonymous" ? signInProvider : undefined,
-    auth_time: user.lastLoginAt != null
-      ? toUnixTimestamp(new Date(user.lastLoginAt))
-      : user.lastRefreshAt != null
-      ? toUnixTimestamp(new Date(user.lastRefreshAt))
-      : toUnixTimestamp(new Date()),
+    auth_time:
+      user.lastLoginAt != null
+        ? toUnixTimestamp(new Date(user.lastLoginAt))
+        : user.lastRefreshAt != null
+        ? toUnixTimestamp(new Date(user.lastRefreshAt))
+        : toUnixTimestamp(new Date()),
     user_id: user.localId,
     firebase: {
       identities,
--- a/src/emulator/auth/operations.ts
+++ b/src/emulator/auth/operations.ts
@@ -1726,9 +1726,9 @@ function generateJwt(
     // This field is only set for anonymous sign-in but not for any other
     // provider (such as email or Google) in production. Let's match that.
     provider_id: signInProvider === "anonymous" ? signInProvider : undefined,
-    auth_time: user.lastLoginAt
+    auth_time: user.lastLoginAt != null
       ? toUnixTimestamp(new Date(user.lastLoginAt))
-      : user.lastRefreshAt
+      : user.lastRefreshAt != null
       ? toUnixTimestamp(new Date(user.lastRefreshAt))
       : toUnixTimestamp(new Date()),
     user_id: user.localId,
--- a/src/emulator/auth/operations.ts
+++ b/src/emulator/auth/operations.ts
@@ -1726,7 +1726,11 @@ function generateJwt(
     // This field is only set for anonymous sign-in but not for any other
     // provider (such as email or Google) in production. Let's match that.
     provider_id: signInProvider === "anonymous" ? signInProvider : undefined,
-    auth_time: toUnixTimestamp(new Date()),
+    auth_time: user.lastLoginAt
+      ? toUnixTimestamp(new Date(user.lastLoginAt))
+      : user.lastRefreshAt
+      ? toUnixTimestamp(new Date(user.lastRefreshAt))
+      : toUnixTimestamp(new Date()),
     user_id: user.localId,
     firebase: {
       identities,
@@ -2122,6 +2126,7 @@ export interface FirebaseJwtPayload {
   exp: number; // expiresAt (in seconds since epoch)
   iss: string; // issuer
   aud: string; // audience (=projectId)
+  auth_time: number; // lastLoginAt (in seconds since epoch)
   // ...and other fields that we don't care for now.
 
   // Firebase-specific fields:
--- a/src/test/emulators/auth/misc.spec.ts
+++ b/src/test/emulators/auth/misc.spec.ts
@@ -21,7 +21,7 @@ import {
 } from "../../../emulator/auth/operations";
 import { toUnixTimestamp } from "../../../emulator/auth/utils";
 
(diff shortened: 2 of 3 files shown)


src/emulator/auth/operations.ts
```diff
@@ -1726,7 +1726,11 @@ function generateJwt(
     // This field is only set for anonymous sign-in but not for any other
     // provider (such as email or Google) in production. Let's match that.
     provider_id: signInProvider === "anonymous" ? signInProvider : undefined,
-    auth_time: toUnixTimestamp(new Date()),
+    auth_time: user.lastLoginAt
```
yuchenshi, line 1729:
```suggestion
    auth_time: user.lastLoginAt != null
```

`user.lastLoginAt` may as well be the number zero if it happens at epoch. (This won't happen in real applications, but may happen during tests where we mock out the clock)

src/test/emulators/auth/misc.spec.ts
```diff
@@ -39,6 +44,31 @@ describeAuthEmulator("token refresh", ({ authApi }) => {
...
+    const res = await authApi()
+      .post("/securetoken.googleapis.com/v1/token")
+      .type("form")
+      // snake_case parameters also work, per OAuth 2.0 spec.
+      .send({ refresh_token: refreshToken, grantType: "refresh_token" })
+      .query({ key: "fake-api-key" });
+
+    const idToken = res.body.id_token;
+    const user = await getAccountInfoByIdToken(authApi(), idToken);
+    const lastLoginAtSeconds = user.lastLoginAt
```
yuchenshi, line 60: We should add `expect(user.lastLoginAt).not.to.be.undefined` before this -- no need to handle the case where `lastLoginAt` is not set. You can use `new Date(user.lastLoginAt!)` to tell TypeScript it is not null if it complains.

src/test/emulators/auth/misc.spec.ts
```diff
@@ -39,6 +44,31 @@ describeAuthEmulator("token refresh", ({ authApi }) => {
       });
   });
 
+  it("should populate auth_time to match lastLoginAt (in seconds since epoch)", async () => {
+    const emailUser = { email: "alice@example.com", password: "notasecret" };
+    const { refreshToken } = await registerUser(authApi(), emailUser);
+
+    const res = await authApi()
```
yuchenshi, line 51: Please advance the mocked clock by a few seconds before this so that we can verify that the auth_time is indeed the login time, not the refresh time. You may use this as an inspiration:

https://github.com/firebase/firebase-tools/blob/66447909ab6f2defd04b257b48193c70d778c7ea/src/test/emulators/auth/oob.spec.ts#L215

yuchenshi (commented):
Thanks for the contribution! Looks pretty good to me. A few comments inline for you though. @lisajian will also help with the review

lisajian (commented):
Overall, LGTM pending @yuchenshi's feedback. Thank you for the fix!

lovelle-cardoso:
@yuchenshi @lisajian Resolved all 3 comments. I think that's everything! 👍

src/test/emulators/auth/misc.spec.ts
```diff
@@ -39,6 +44,32 @@ describeAuthEmulator("token refresh", ({ authApi }) => {
       });
   });
 
+  it("should populate auth_time to match lastLoginAt (in seconds since epoch)", async () => {
+    const emailUser = { email: "alice@example.com", password: "notasecret" };
+    const { refreshToken } = await registerUser(authApi(), emailUser);
+
+    getClock().tick(2000); // Wait for idToken to be issued in the past.
```
yuchenshi, line 51:
```suggestion
    getClock().tick(2000); // Wait 2 seconds before refreshing.
```

src/test/emulators/auth/misc.spec.ts
```diff
@@ -39,6 +44,32 @@ describeAuthEmulator("token refresh", ({ authApi }) => {
...
+    const user = await getAccountInfoByIdToken(authApi(), idToken);
+    expect(user.lastLoginAt).not.to.be.undefined;
+    const lastLoginAtSeconds = toUnixTimestamp(new Date(user.lastLoginAt!));
+    const decoded = decodeJwt(idToken, { complete: true }) as {
+      header: JwtHeader;
+      payload: FirebaseJwtPayload;
+    } | null;
+    expect(decoded, "JWT returned by emulator is invalid").not.to.be.null;
+    expect(decoded!.header.alg).to.eql("none");
+    expect(decoded!.payload.auth_time).to.equal(lastLoginAtSeconds);
```
yuchenshi, line 70:
```suggestion
    // This should match login time, not token refresh time.
    expect(decoded!.payload.auth_time).to.equal(lastLoginAtSeconds);
```

yuchenshi (approved):
LGTM. I'll update the comments and merge after CI passes

yuchenshi:
Apparently linter is not happy about the code formatting, but we'll fix this on our side -- just don't delete your fork branch yet!

lisajian (approved):

yuchenshi:
This is now merged and will be included in the next release. Thanks again for the help!

PR #3619, opened 2021-07-29
Set Access Control for HTTP Functions (v1)

colerogers:
### Description

Creating a way for customers to explicitly set the access control on HTTP triggered functions. The access control is set at the function level, so different functions can have different access policies. The customer can choose to set access control to `public`, `private`, or a custom service account. Access control is set in the `runWith()` function by setting the property `invoker`.  The functions SDK will always force the `invoker` to be an array of strings.

This PR enables this feature on the CLI for functions v1.

### Scenarios Tested

- Invalid invoker
- Converting invoker members to their proper invoker role types
- Getting and Setting invoker roles on functions

### Sample Commands


src/deploy/functions/backend.ts
```diff
@@ -133,6 +133,11 @@ export interface TargetIds {
 
 export type FunctionsPlatform = "gcfv1" | "gcfv2";
 
+/**
+ * Invoker access control type for https functions.
+ */
+export type Invoker = "public" | "private" | string;
```
joehan, line 139: Same typing comments from firebase-functions - I prefer making this include the array type

src/gcp/cloudfunctions.ts
```diff
@@ -260,6 +251,74 @@ export async function setIamPolicy(options: IamOptions): Promise<void> {
...
+ * @param serviceAccount the custom service account created by the user
+ * @param projectId the ID of the current project
+ * @returns a correctly formatted service account string
+ */
+function formatServiceAccount(serviceAccount: string, projectId: string): string {
+  if (serviceAccount.length === 0) {
+    throw new Error("Service account cannot be an empty string");
+  }
+  let emailId = "";
+  if (serviceAccount.charAt(serviceAccount.length - 1) === "@") {
```
joehan, line 265: if you want to be fancy, you can use a ternary here:
```
const emailId = (serviceAccount.charAt(serviceAccount.length - 1) === "@")
  ? `${serviceAccount}${projectId}.iam.gserviceaccount.com` 
  : `${serviceAccount}@${projectId}.iam.gserviceaccount.com`;
```

src/gcp/cloudfunctions.ts
```diff
@@ -260,6 +251,74 @@ export async function setIamPolicy(options: IamOptions): Promise<void> {
...
+      });
+    } else {
+      policy.bindings.push({
+        role: invokerRole,
+        members: [formatServiceAccount(invoker, projectId)],
+      });
+    }
+  } else {
+    const serviceAccountMembers: string[] = [];
+    invoker.forEach((element) =>
```
joehan, line 310: Prefer for... of... : go/tsstyle#iterating-containers

```
Do not use Array.prototype.forEach, Set.prototype.forEach, and Map.prototype.forEach. 
They make code harder to debug and defeat some useful compiler checks (e.g. reachability).

````

Comment (joehan) [reply src/gcp/cloudfunctions.ts:310]: If you're feeling this functional style tho, maybe use map instead?
```
else {
  policy.bindings.push({
    role: invokerRole,
    members: invoker.map(e => formatServiceAccount(e, projectId)),
  })
}
```

File: src/test/gcp/cloudfunctions.spec.ts [hunk 4]
```diff
@@ -258,4 +258,50 @@ describe("cloudfunctions", () => {
...
     });
   });
+
+  describe("generateIamPolicy", () => {
+    it("should throw error on empty service account string", () => {
+      expect(() => cloudfunctions.generateIamPolicy("project", "")).to.throw();
+    });
+
+    it("should generate the public policy", () => {
+      const policy = cloudfunctions.generateIamPolicy("project", "public");
```
joehan, line 268: Its a good practice to split your tests with newlines into arrange, act, assert: go/unit-testing-practices#structure

For these, theres no arrange step, but you can throw a newline between the call to generateIamPolicy and the expects

src/test/gcp/cloudfunctions.spec.ts
```diff
@@ -258,4 +258,50 @@ describe("cloudfunctions", () => {
...
+      expect(policy.bindings[0].members[0]).to.eq("allUsers");
+    });
+
+    it("should generate the private policy", () => {
+      const policy = cloudfunctions.generateIamPolicy("project", "private");
+      expect(policy.bindings.length).to.eq(1);
+      expect(policy.bindings[0].members.length).to.eq(0);
+    });
+
+    it("should generate a policy for a single service account", () => {
```
joehan, line 280: Can you add a test for passing a full service account (service-account@proj.iam.gserviceaccount.com) as well?

joehan (commented):
This looks pretty good! Just a few style suggestions and minor comments.

src/deploy/functions/errorHandler.ts
```diff
@@ -38,6 +38,20 @@ export class ErrorHandler {
...
+    if (failedIamCalls.length) {
+      logger.info("\nUnable to set the invoker for the IAM policy on the following functions:");
+      for (const failedDep of failedIamCalls) {
+        logger.info(`\t${failedDep.functionName}`);
+      }
+      logger.info("\nSome common causes of this:");
+      logger.info(
+        "\n- You may not have the roles/functions.admin IAM role. Note that roles/functions.developer does not allow you to change IAM policies."
+      );
+      logger.info("\n- An organization policy that restricts Network Access on your project.");
```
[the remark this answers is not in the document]
colerogers: Removed all newlines

src/gcp/cloudfunctions.ts
colerogers: Switched to the ternary operator

src/gcp/cloudfunctions.ts
colerogers: Switched to the for ... of ... syntax

src/test/gcp/cloudfunctions.spec.ts
colerogers: Added a check and tests for using a full service account in the invoker

src/test/gcp/cloudfunctions.spec.ts
colerogers: Cleaned up spacing in the tests

src/deploy/functions/backend.ts
colerogers: resolving this since we agreed to leave as-is for easier future development.

src/deploy/functions/errorHandler.ts
```diff
@@ -34,11 +34,31 @@ export class ErrorHandler {
...
     }
-    logger.info("\nFunctions deploy had errors with the following functions:");
+    logger.info("");
+    logger.info("Functions deploy had errors with the following functions:");
     for (const failedDeployment of this.errors) {
       logger.info(`\t${getFunctionLabel(failedDeployment.functionName)}`);
     }
-    logger.info("\nTo try redeploying those functions, run:");
+
+    const failedIamCalls = this.errors.filter((e) => e.operationType === "set invoker");
```
colerogers, line 43: changed this to error type since we want the function deployment to return a non-zero exit code

src/gcp/cloudfunctions.ts
```diff
@@ -260,6 +253,73 @@ export async function setIamPolicy(options: IamOptions): Promise<void> {
...
+  return `serviceAccount:${emailId}`;
+}
+
+/**
+ * Generates the correct IAM Policy for changing the invoker for HTTP triggered functions.
+ * @param projectId the ID of the current project
+ * @param invoker the member(s) that will be granted the invoker role
+ * @returns correctly formatted IAM policy for use in IAM API calls
+ */
+export function generateIamPolicy(
```
colerogers, line 292: We can reuse this function for setting the IAM policy on v2 functions since cloud run uses the same policy structure. I believe that leaving it in this file makes sense, but interested to hear others thoughts.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -51,6 +51,7 @@ function tryValidate(typed: backend.Backend) {
       environmentVariables: "omit",
       uri: "omit",
       sourceUploadUrl: "omit",
+      invoker: "object",
```
inlined, line 54: Array. We'll have to make the container contract return an array regardless of whether the dev passed a scalar.

src/deploy/functions/runtimes/node/parseTriggers.ts
```diff
@@ -57,6 +57,7 @@ export interface TriggerAnnotation {
   timeZone?: string;
   regions?: string[];
   concurrency?: number;
+  invoker?: string | string[];
```
inlined, line 60: Let's make this an array always. Your code will be cleaner later.

src/deploy/functions/tasks.ts
```diff
@@ -125,14 +125,14 @@ export function createFunctionTask(
         if (fn.platform === "gcfv1") {
           await gcf.setIamPolicy({
```
inlined, line 126: This should noop if the policy is `private`

src/deploy/functions/tasks.ts
```diff
@@ -192,6 +192,20 @@ export function updateFunctionTask(
       onPoll,
     };
     const cloudFunction = await pollOperation<unknown>(pollerOptions);
+    if (!backend.isEventTrigger(fn.trigger)) {
+      try {
+        if (fn.platform === "gcfv1") {
+          await gcf.setIamPolicy({
```
inlined, line 198: This should noop if `fn.invoker` isn't set.

src/gcp/cloudfunctions.ts
```diff
@@ -260,6 +253,73 @@ export async function setIamPolicy(options: IamOptions): Promise<void> {
...
+/**
+ * Formats the service account to be used with IAM API calls, a vaild service account string is
+ * 'service-account@project.iam.gserviceaccount.com', 'service-account@', or 'service-account'.
+ * @param serviceAccount the custom service account created by the user
+ * @param projectId the ID of the current project
+ * @returns a correctly formatted service account string
+ *
+ * @throws {@link Error} if the supplied service account string is empty
+ */
+function formatServiceAccount(serviceAccount: string, projectId: string): string {
```
inlined, line 265: `gcp/proto.ts` is becoming a grab bag of types useful for handling GCP interfaces. This could probably go there so you can reuse it in GCFv1 and Run

src/gcp/cloudfunctions.ts
```diff
@@ -260,6 +253,73 @@ export async function setIamPolicy(options: IamOptions): Promise<void> {
...
+
+  const policy: iam.Policy = {
+    version: 3,
+    bindings: [
+      {
+        role: invokerRole,
+        members: roleMembers,
+      },
+    ],
+    etag: "",
```
inlined, line 318: It's not great that we're blitzing away any IAM settings by setting etag to "". We should probably merge the `generateIamPolicy` and `setIamPolicy` functions into a `setInvoker` function that would (in a loop that handles race conditions) fetch the existing IAM policy, update only the invoker binding, and then set it with the etag to catch races.

src/gcp/cloudfunctions.ts
```diff
@@ -260,6 +253,73 @@ export async function setIamPolicy(options: IamOptions): Promise<void> {
...
+ * @returns correctly formatted IAM policy for use in IAM API calls
+ */
+export function generateIamPolicy(
+  projectId: string,
+  invoker: backend.Invoker | backend.Invoker[] = "public"
+): iam.Policy {
+  const invokerRole = "roles/cloudfunctions.invoker";
+  const roleMembers: string[] = [];
+
+  if (typeof invoker === "string") {
```
inlined, line 299: If invoker is `private` you need to remove all bindings with the invoker role.

src/gcp/cloudfunctions.ts
inlined: The roles for v1 and v2 are different, so it's hard to reuse the function.

src/gcp/cloudfunctions.ts
```diff
@@ -260,6 +253,73 @@ export async function setIamPolicy(options: IamOptions): Promise<void> {
...
+/**
+ * Generates the correct IAM Policy for changing the invoker for HTTP triggered functions.
+ * @param projectId the ID of the current project
+ * @param invoker the member(s) that will be granted the invoker role
+ * @returns correctly formatted IAM policy for use in IAM API calls
+ */
+export function generateIamPolicy(
+  projectId: string,
+  invoker: backend.Invoker | backend.Invoker[] = "public"
+): iam.Policy {
```
inlined, line 295: I would probably take `invoker: backend.Invoker[]` always, special case if any of the elements is "private" first, if any of them are "public" next, and then fall back to `formatServiceAccount`

inlined (changes requested):

inlined:
FYI, this is going to cause you some rebase problems: https://github.com/firebase/firebase-functions/pull/943

src/deploy/functions/runtimes/discovery/v1alpha1.ts
colerogers: Changed

src/deploy/functions/runtimes/node/parseTriggers.ts
colerogers: Changed

src/deploy/functions/tasks.ts
colerogers: Added check

src/gcp/cloudfunctions.ts
colerogers: Moved it over there

src/gcp/cloudfunctions.ts
colerogers: Fixed

src/gcp/cloudfunctions.ts
colerogers: Updated signature with the function rewrite

src/deploy/functions/tasks.ts
colerogers: Changed the functionality a bit and now private is handled in the `setInvoker` function.

src/gcp/cloudfunctions.ts
```diff
@@ -228,33 +219,74 @@ export async function createFunction(
...
+ * Gets the current invoker IAM policy for the function and overrides the invoker with the supplied invoker members
+ * @param projectId id of the project
+ * @param fnName function name
+ * @param invoker an array of invoker strings
+ *
+ * @throws {@link FirebaseError} on an empty invoker, when the IAM Polciy fails to be grabbed or set
  */
-export async function setIamPolicy(options: IamOptions): Promise<void> {
-  const endpoint = `/${API_VERSION}/${options.name}:setIamPolicy`;
+export async function setInvoker(
```
colerogers, line 237: Right now, I perform both api calls internally, would it be a better practice to separate the get & set calls into their own functions?

src/gcp/proto.ts
```diff
@@ -123,3 +123,37 @@ function fieldMasksHelper(
...
+ * Formats a single invoker to be used with IAM API calls, a vaild invoker is
+ * 'public', 'service-account', 'service-account@', or
+ * 'service-account@project.iam.gserviceaccount.com'.
+ * @param invokerMember the single invoker that is a member of the invoker array
+ * @param projectId the ID of the current project
+ * @returns a correctly formatted service account string
+ *
+ * @throws {@link Error} if the supplied service account string is empty
+ */
+export function formatInvokerMember(invokerMember: string, projectId: string): string {
```
inlined, line 137: nit: I liked the method better when it was just for picking <sa>@<domain> vs <sa>@. Now that we special-case "public" I'm not sure if it works in other cases like the function.serviceAccount annotation. Can we keep this the way it was (formatServiceAccount), make the existing code use the helper, and then special case "public" => "allUsers" either in a helper function that uses formatServiceAccount or in setInvoker?

src/test/gcp/cloudfunctions.spec.ts
```diff
@@ -259,82 +261,112 @@ describe("cloudfunctions", () => {
...
+          { role: "roles/cloudfunctions.invoker", members: ["some-service-account"] },
+        ],
+        etag: "1234",
+        version: 3,
+      });
+      apiRequestStub.onSecondCall().callsFake((method: any, resource: any, options: any) => {
+        expect(options.data.policy).to.deep.eq({
+          bindings: [
+            { role: "random-role", members: ["user:pineapple"] },
+            { role: "roles/cloudfunctions.invoker", members: [] },
```
inlined, line 335: Interesting. It didn't even occur to me to have an empty binding instead of just removing the binding. Does this work?

src/test/gcp/cloudfunctions.spec.ts
```diff
@@ -259,82 +261,112 @@ describe("cloudfunctions", () => {
...
+      });
 
-      expect(policy.bindings.length).to.eq(1);
-      for (let i = 0; i < 10; i++) {
-        expect(policy.bindings[0].members.length).to.eq(10);
-        expect(policy.bindings[0].members[i]).to.eq(`serviceAccount:${accounts[i]}`);
-      }
+      await expect(
+        cloudfunctions.setInvoker("project", "function", [
+          "service-account1",
```
inlined, line 364: IIRC the SDK always requires a service account to end with or at least include a '@' character. It might be good to enforce this in the SDK as well, which means that when we add new shorthands like `hosting` users of an outdated CLI will get an error instead of adding a nonsense binding of `hosting@projectId.iam.gserviceaccount.com`

src/deploy/functions/tasks.ts
```diff
@@ -120,13 +120,10 @@ export function createFunctionTask(
       operationResourceName: op.name,
       onPoll,
     });
-    if (!backend.isEventTrigger(fn.trigger)) {
+    if (!backend.isEventTrigger(fn.trigger) && fn.invoker) {
```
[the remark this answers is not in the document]
inlined: Honestly, on second review, I think we should possibly have a version of setInvoker that does blind writes vs one that updates on top of an existing IAM policy. It'd be nice to cut N IAM policy changes out, but it isn't a blocker.

src/gcp/cloudfunctions.ts
```diff
@@ -253,71 +251,65 @@ export async function setIamPolicy(options: IamOptions): Promise<void> {
...
-    bindings: [
-      {
-        role: invokerRole,
-        members: roleMembers,
-      },
-    ],
-    etag: "",
+    bindings: bindings,
+    etag: currentPolicy.etag || "",
+    version: currentPolicy.version || 3,
```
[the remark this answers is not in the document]
colerogers: From the docs, valid versions are [0,1,2,3] and specify the format of the policy. I thought using the same version from the get call made sense. But I'll change it to force version 3.

src/gcp/proto.ts
colerogers: Sure i'll change it back

src/test/gcp/cloudfunctions.spec.ts
colerogers: From my testing it does, sending an empty array to the api will remove the binding on a subsequent GET request.

src/test/gcp/cloudfunctions.spec.ts
colerogers: changed the service account formats to either end with or include a '@'.

src/gcp/proto.ts
colerogers: Changed back and added a helper function to get the correct invoker members.

src/gcp/cloudfunctions.ts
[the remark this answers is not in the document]
colerogers: Changed to force version 3 in latest push

src/deploy/functions/tasks.ts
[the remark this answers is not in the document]
colerogers: I've set `fn.invoker = ["public"]` if it's undefined and added a check in the if statement `fn.invoker[0] !== "private"`. Since in the sdk we already make sure that `public` and `private` are in an array of length 1, checking the first element here makes sense to me. Thoughts?

Also, I created two setInvoker functions `setInvokerCreate` and `setInvokerUpdate`. `setInvokerCreate` does not pull the current IAM policies, just assigns the invoker role and corresponding members. `setInvokerUpdate` does the GET call first, then updates the invoker role members.

src/deploy/functions/tasks.ts
```diff
@@ -120,10 +120,13 @@ export function createFunctionTask(
       operationResourceName: op.name,
       onPoll,
     });
-    if (!backend.isEventTrigger(fn.trigger) && fn.invoker) {
+    if (!fn.invoker) {
```
inlined, line 123: 1. Only for https triggers. We will possibly start supporting invoker for other triggers in v2 (you can set EventArc's invoker), but we would never default to public
2. I'm nervous about modifying the fn itself. It might be better to just copy the list out

src/gcp/cloudfunctions.ts
```diff
@@ -307,7 +335,7 @@ export async function setInvoker(
   const policy: iam.Policy = {
     bindings: bindings,
     etag: currentPolicy.etag || "",
-    version: currentPolicy.version || 3,
+    version: 3,
   };
   await setIamPolicy({ name: fnName, policy: policy });
```
inlined, line 340: One more thing I just realized: it might be nice to not update the bindings if they're already set correctly (the 99% case). This will not only save API requests, but avoid triggering scary sounding audit logs.

inlined (approved):

PR #3622, opened 2021-07-30
Improve Pub/Sub function deployments

inlined:
Fixes some issues discovered when testing pub/sub function deployments manually.

GCFv2 has some behavior changes that we need to take into account in our CLI:

1. GCF no longer creates Pub/Sub topics automatically. We handle this in tasks.ts by manually creating any task we need. The underlying API is `put` so we are safe if the topic already exists.
2. GCF fails if `pubsubTopic` is provided in UpdateFunction irrespective of whether that topic even changed. We've moved the check for a change in topic to `checkForInvalidChangeOfTrigger` and no longer send the topic in the API call.

### DIFF
--- a/src/deploy/functions/tasks.ts
+++ b/src/deploy/functions/tasks.ts
@@ -99,9 +99,16 @@ export function createFunctionTask(
       // for Pub/Sub event handlers. This may change, at which point this code
       // could be deleted.
       if (apiFunction.eventTrigger?.pubsubTopic) {
-        await pubsub.createTopic({
-          name: apiFunction.eventTrigger.pubsubTopic,
-        });
+        try {
+          await pubsub.getTopic(apiFunction.eventTrigger.pubsubTopic);
+        } catch (err) {
+          if (err.status !== 404) {
+            throw new FirebaseError("Unexpected error looking for Pub/Sub topic", {original: err})
+          }
+          await pubsub.createTopic({
+            name: apiFunction.eventTrigger.pubsubTopic,
+          });
+        }
       }
       op = await gcfV2.createFunction(apiFunction);
     }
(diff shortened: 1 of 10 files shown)


src/deploy/functions/tasks.ts
```diff
@@ -95,6 +95,14 @@ export function createFunctionTask(
       op = await gcf.createFunction(apiFunction);
     } else {
       const apiFunction = gcfV2.functionFromSpec(fn, params.storageSource!);
+      // N.B. As of GCFv2 private preview GCF no longer creates Pub/Sub topics
+      // for Pub/Sub event handlers. This may change, at which point this code
+      // could be deleted.
+      if (apiFunction.eventTrigger?.pubsubTopic) {
+        await pubsub.createTopic({
```
mbleigh, line 102: Is this idempotent? What if the topic already exists?

Do we need to ensure the API is enabled first? Or is it going to be automatically since GCF depends on Pub/Sub?

src/deploy/functions/validate.ts
```diff
@@ -85,4 +86,32 @@ export function checkForInvalidChangeOfTrigger(
...
   }
+
+  if (changedV2PubSubTopic(exFn, fn)) {
+    throw new FirebaseError(
+      `[${getFunctionLabel(fn)}] Cannot change the Pub/Sub topic of a v2 Cloud Function`
+    );
+  }
+}
+
+function changedV2PubSubTopic(exFn: backend.FunctionSpec, fn: backend.FunctionSpec): boolean {
```
mbleigh, line 97: Should we delete and recreate with the new topic in this case? Or should we require a remove, deploy, re-add, deploy as this would (I think)?

src/deploy/functions/tasks.ts
mbleigh: Enablement checks add latency (even if small), so if we know it'll be there I think it's fine. Do fix idempotency tho - are you doing that in this PR or a followup?

src/deploy/functions/validate.ts
mbleigh: If someone is using `firebase deploy` in a CI/CD environment, this means that if they want to redeploy the exact same function they had before but with a new topic, they would need to comment out, commit/push, uncomment, commit/push.

Even though the backend doesn't allow changing topic, we could detect a topic change as similar to redeploying a different trigger to the same function name and just delete the old one and create the new one. That seems more consistent with our behavior elsewhere.

src/deploy/functions/tasks.ts
inlined: Already fixed

src/deploy/functions/validate.ts
inlined: Are you sure it's wise to risk deleting the function and then failing to recreate it? It's a bit of a PITA to have some functions that need to be deleted before created and other functions that just get deleted outright after creates are done, but I probably need that to upgrade a scheduled function from Pub/Sub to HTTPS anyway

src/deploy/functions/tasks.ts
inlined: Whoops. I never published that part of the change. Fixing.

src/deploy/functions/validate.ts
inlined: On one hand I'm not 100% sure this is a great idea, but OTOH it lays the foundation for upgrading scheduled functions, so I'm onboard I guess. Additional changes incoming

mbleigh (approved):

Issue #3613, opened 2021-07-28
Improve support for storing multiline strings in functions config


taeold:
Spun out of #371.

Trying to add firebase config value with newline leads to surprising result:

```
$ firebase functions:config:set a.b="Hello\nWorld"
$ firebase functions:config:get
{
  "a": {
    "b": "Hello\\nWorld" # LOOK
  },
}
```

Note that extra back slash.

Unfortunately, this issue isn't specific to Firebase CLI. It's tricky to pass escape characters like newline (e.g. [stackoverflow question](https://stackoverflow.com/questions/58896676/how-to-pass-multiline-text-to-a-command-in-a-parameter)).

For illustration, consider this program:

```js
// args.js
console.log(process.argv.slice(2))
```

```
$ node args.js "a\nb"
[ 'a\\nb' ]

$ node args.js "Hello\World"
[ 'Hello\\World' ]

$ node args.js "abc$def"
[ 'abc' ]
```

Trick is to use [ANSI-C Quoting](http://www.gnu.org/software/bash/manual/html_node/ANSI_002dC-Quoting.html):

```
$ node args.js $'a\nb'
[ 'a\nb' ]
$ firebase functions:config:set a.b=$'Hello\nWorld'
$ firebase functions:config:get
{
  "a": {
    "b": "Hello\nWorld"
  },
}
```

This is a common enough mistake that I would like to have a simpler way pass multiline strings as a value for functions config. For example, add a `--file` flag to allow values to be read from a file instead?

IzaakGough:
Closing this as `functions:config:*` is deprecated and will be removed before March 2027, in favor of .env files / params, which don't have this problem.

The original bug was really shell argv handling - without $'...' quoting, \n arrives as literal backslash+n before firebase-tools ever sees it. .env files avoid this structurally: values are read from file contents, not process.argv, so there's no shell-quoting step to trip over, and multiline values are supported natively.

Existing legacy config can be migrated automatically with `firebase functions:config:export`: https://firebase.google.com/docs/functions/config-env#migrate-config

PR #3625, opened 2021-08-02
Add module to parse dotenv file

taeold:
Small regex-based implementation for parsing dotenv files, e.g. turns

```
FOO=bar
MY_SERCVICE_URL=http://example.com
```

into a JS objects:

```js
{
  FOO: "bar",
  MY_SERVICE_URL: "http://example.com"
}
```

The parser is used to load environment variables for Cloud Functions for Firebase by searching for `.env` and `.env.<project or alias or local>` in the function source directory.

We add additional requirements that the dotenv file be free of errors (i.e. don't have any lines with invalid contents) and not contain keys that are reserved for internal use.

There is already popular NPM package for parsing dotenv files (https://www.npmjs.com/package/dotenv), but I decided to not use it because it lacks few capabilities like multiline string support (we want this! See https://github.com/firebase/firebase-tools/issues/3613) and inspecting invalid lines for clearer error messages.


src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+ * We restrict key names to ones that conform to POSIX standards.
+ * This is more restrictive than what is allowed in Cloud Functions or Cloud Run.
+ */
+export function validateKey(key: string): void {
+  if (RESERVED_KEYS.includes(key)) {
+    throw new Error("Key reserved for internal use.");
+  }
+  if (!/^[A-Z_][A-Z0-9_]*$/.test(key)) {
+    throw new Error(
+      "Key must start with an uppercase ASCII letter or underscore" +
```
mbleigh, line 134: nit: print out the invalid key in the error

src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+        ", and then consist of uppercase ASCII letters, digits, and underscores."
+    );
+  }
+  if (key.startsWith("X_GOOGLE_") || key.startsWith("FIREBASE_")) {
+    throw new Error("Key starts with a reserved prefix (X_GOOGLE_ or FIREBASE_)");
```
mbleigh, line 139: nit: print out the invalid key in the error

src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+// prettier-ignore
+const LINE_RE = new RegExp(
+  "^" +                    // begin line
+  "\\s*" +                 //   leading whitespaces
+  "(\\w+)" +               //   key
+  "\\s*=\\s*" +            //   separator (=)
+  "(" +                    //   begin optional value
+  `\\s*'(?:\\'|[^'])*'|` + //     single quoted or
+  `\\s*"(?:\\"|[^"])*"|` + //     double quoted or
+  "[^\\#\\r\\n]+" +        //     unquoted
```
inlined, line 46: What is `\#` in regexp? I can't find any reference to that. Do you mean a literal #?

src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * @return {Record<string, string>} Environment variables for the project.
+ */
+export function load(options: {
+  functionsSource: string;
+  projectId: string;
+  projectAlias?: string;
+  isEmulator?: boolean;
+}): Record<string, string> {
+  const targetFiles = [".env", `.env.${options.projectId}`];
```
mbleigh, line 195: Do you intend to have `.env.${options.projectId}` here? Looks like your conditional would only included it if not emulator, and it would end up being in there twice since it gets pushed in the else.

src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+  const targetPaths = targetFiles
+    .map((f) => path.join(options.functionsSource, f))
+    .filter(fs.existsSync);
+
+  // Check if both .env.<project> and .env.<alias> exists.
+  if (targetPaths.some((p) => path.basename(p) === `.env.${options.projectId}`)) {
+    if (options.projectAlias && options.projectAlias.length) {
+      for (const p of targetPaths) {
+        if (path.basename(p) === `.env.${options.projectAlias}`) {
+          throw new FirebaseError("Can't have both .env.<project> and .env.<alias> files.");
```
mbleigh, line 214: nit: Print out the specific filenames that conflict.

src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+    }
+  }
+
+  let envs: Record<string, string> = {};
+  for (const targetPath of targetPaths) {
+    try {
+      const data = fs.readFileSync(targetPath, "utf8");
+      envs = { ...envs, ...parseStrict(data) };
+    } catch (err) {
+      logger.debug(`Failed to load environment variables from ${targetPath}`, err);
```
mbleigh, line 226: The error will get printed anyway when bubbled up, so why a debug log line?

src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+const LINE_RE = new RegExp(
+  "^" +                    // begin line
+  "\\s*" +                 //   leading whitespaces
+  "(\\w+)" +               //   key
+  "\\s*=\\s*" +            //   separator (=)
+  "(" +                    //   begin optional value
+  `\\s*'(?:\\'|[^'])*'|` + //     single quoted or
+  `\\s*"(?:\\"|[^"])*"|` + //     double quoted or
+  "[^\\#\\r\\n]+" +        //     unquoted
+  ")?" +                   //   end optional value
```
inlined, line 47: Can we actually handle Env vars with no value?

src/test/functions/env.spec.ts
```diff
@@ -0,0 +1,352 @@
...
+      expect(() => {
+        env.validateKey("FIREBASE_FOOBAR");
+      }).to.throw("starts with a reserved prefix");
+    });
+  });
+
+  describe("load", () => {
+    const createEnvFiles = (sourceDir: string, envs: Record<string, string>): void => {
+      for (const [filename, data] of Object.entries(envs)) {
+        fs.writeFileSync(path.join(sourceDir, filename), data);
```
mbleigh, line 205: Idle thought, I wonder if we should bring in something like https://www.npmjs.com/package/mock-fs instead of actually writing these files for each test run.

src/test/functions/env.spec.ts
```diff
@@ -0,0 +1,352 @@
+import * as fs from "fs";
+import * as path from "path";
+import * as os from "os";
+import { sync as rimraf } from "rimraf";
+import { expect } from "chai";
+
+import * as env from "../../functions/env";
+
+describe("functions/env", () => {
+  describe("parse", () => {
```
mbleigh, line 10: You might want to consider a table-based test (e.g. programmatically drive tests from an array of data) just to cut down on repetitive test code, but honestly this is pretty readable and you've already done the heavy lifting so I think it's also fine as-is.

mbleigh (commented):
❤️  all the attention to detail in edge case testing!

src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+  const errors: string[] = [];
+
+  data = data.replace(/\r\n?/, "\n"); // For Windows support.
+  let match;
+  while ((match = LINE_RE.exec(data))) {
+    let [, k, v] = match;
+    v = (v || "").trim();
+
+    let quotesMatch;
+    if ((quotesMatch = /^(["'])(.*)\1$/ms.exec(v)) != null) {
```
inlined, line 96: Ugh.. TIL that you can actually inline regexp options after the trailing /. How is that not a parser can of worms? I wonder if `new RegExp("^([\"'])(.*)\1$", "ms")` is clearer, but I think it's all going to be horrific.

src/functions/env.ts
inlined, line 139: nit: I would consider using the same error as above so that our error space doesn't strictly map to our implementation.

src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+
+/**
+ * Validates string for use as an env var key.
+ *
+ * We restrict key names to ones that conform to POSIX standards.
+ * This is more restrictive than what is allowed in Cloud Functions or Cloud Run.
+ */
+export function validateKey(key: string): void {
+  if (RESERVED_KEYS.includes(key)) {
+    throw new Error("Key reserved for internal use.");
```
inlined, line 130: Include the key name please, and why aren't we throwing a FirebaseError?

src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+      invalidKeys[key] = err.message || "unknown error";
+    }
+  }
+  if (Object.keys(invalidKeys).length > 0) {
+    logger.debug(
+      `Found invalid keys: ${Object.entries(invalidKeys)
+        .map(([k, v]) => `${k}: ${v}`)
+        .join("\n")}`
+    );
+    throw new Error(`Found invalid keys: ${Object.keys(invalidKeys).join(",")}`);
```
inlined, line 168: Again, I would consider
1. Using Firebase Error
2. Using the key name in the error message
3. Using FirebaseError.children instead of a new message here (assuming it shows up in the console correctly)

src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+ *   2. .env.<project or alias> (or .env.local if isEmulator is true).
+ *
+ * If both .env.<project> and .env.<alias> files are found, an error is thrown.
+ *
+ * @return {Record<string, string>} Environment variables for the project.
+ */
+export function load(options: {
+  functionsSource: string;
+  projectId: string;
+  projectAlias?: string;
```
inlined, line 192: I don't like the fact that the param projectOrAlias means nothing when isEmulator is true. Why not just document that we use "local" as projectOrAilas for the emulator?

src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+  isEmulator?: boolean;
+}): Record<string, string> {
+  const targetFiles = [".env", `.env.${options.projectId}`];
+  if (options.isEmulator) {
+    targetFiles.push(".env.local");
```
inlined, line 197: IIRC we explicitly chose _not_ to include project files when running in the emulator.

src/functions/env.ts
taeold: Oh thanks for catching this!

src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+  } else {
+    targetFiles.push(`.env.${options.projectId}`);
+    if (options.projectAlias && options.projectAlias.length) {
+      targetFiles.push(`.env.${options.projectAlias}`);
+    }
+  }
+
+  const targetPaths = targetFiles
+    .map((f) => path.join(options.functionsSource, f))
+    .filter(fs.existsSync);
```
inlined, line 207: It's much more of a PITA, but I'm wondering if we should be doing this in parallel.

src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+ * .env files are searched and merged in the following order:
+ *
+ *   1. .env
+ *   2. .env.<project or alias> (or .env.local if isEmulator is true).
+ *
+ * If both .env.<project> and .env.<alias> files are found, an error is thrown.
+ *
+ * @return {Record<string, string>} Environment variables for the project.
+ */
+export function load(options: {
```
inlined, line 189: nit: if this were async you could read all the files in parallel, though I guess it probably doesn't make a huge difference to save a ms.

inlined (approved):
Some pretty dense reading there for a bit, but LGTM. I do want to discuss the local vs project ID/alias thing though. I don't think we should be supporting separate environments + "local" the same way we don't support alias + project ID. I'd argue that semantically we should just pass "local" as projectId when using the emulator.

src/functions/env.ts
taeold: Good point. Originally, debug line contained more info, but I decided to push it all on to the error as well.

src/test/functions/env.spec.ts
taeold: Works great!

src/test/functions/env.spec.ts
taeold: Tried it. Feels prretty good.

src/functions/env.ts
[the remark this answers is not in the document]
taeold: The way I'm using the error message would look a bit redundant if I do this:

```
Found invalid key:
X_GOOGLE_FOOBAR: Key X_GOOGLE_FOOBAR is reserved for internal use

VS

Found invalid key:
X_GOOGLE_FOOBAR: Key reserved for internal use
```

With that said, the code looks a bit cleaner if I include the key name in the error message in the first place. Moving in that direction.

src/functions/env.ts
```diff
@@ -0,0 +1,238 @@
...
+// Inspired by:
+//   https://github.com/bkeepers/dotenv/blob/master/lib/dotenv/parser.rb
+// prettier-ignore
+const LINE_RE = new RegExp(
+  "^" +                    // begin line
+  "\\s*" +                 //   leading whitespaces
+  "(\\w+)" +               //   key
+  "\\s*=\\s*" +            //   separator (=)
+  "(" +                    //   begin optional value
+  `\\s*'(?:\\'|[^'])*'|` + //     single quoted or
```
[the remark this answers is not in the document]
taeold: Whoops. This piece of code went through several iteration (e.g. I tried using `String.raw...` but didn't like how it looked), and this is just a fossil from those times. Updating to use `"` and `'` instead.

src/functions/env.ts
taeold: Huh I thought # had to be escaped, but the tests still pass without it. Thanks for simplifying this.

src/functions/env.ts
taeold: Oh that's a great point. I cannot!

https://cloud.google.com/functions/docs/configuring/env-var#best_practices_and_reserved_environment_variables

src/functions/env.ts
taeold: Yep updated to include keys.

As used, error thrown by `validateKey` is meant to be caught by the caller that will decorate it and throw `FirebaseError` at the end of the day. Any reason not to use `Error` here?

src/functions/env.ts
taeold: Maybeeeeeeeeeee?

I'm for keeping the "slow" implementation since the code read at most 2 files and each file would normally be small, but I also don't know how much more of a PITA it would be to convert this into "parallel" code. If there is some existing implementation I can peek at?

src/functions/env.ts
taeold: Happy to give it a try - would love if you can give me some pointers.

src/functions/env.ts
taeold: oh that's a great point. Emulator can just pass `projectId: "local"`!

src/test/functions/env.spec.ts
taeold: Reverted since `mock-fs` only supports Nodejs12 and above 😢

src/functions/env.ts
inlined: I thought we display FirebaseError messages and not Error messages. I could be wrong.

src/functions/env.ts
inlined: On further thought and consulting with the CLI chat room, don't sweat it.

src/functions/env.ts
inlined: Per above, don't sweat it.

src/functions/env.ts
taeold: Yes. Fixed!

src/functions/env.ts
taeold: Oh I read the docs wrong. I tested and it's valid to have an empty value for env vars.

joehan:
Nice job on this! The extensions emulator uses that dotenv package you mentioned (https://github.com/firebase/firebase-tools/blob/c86418676eb11dacf476ef636d394b43812ee2a4/src/extensions/paramHelper.ts#L136) - I may switch it over to use this instead at some point to take advantage of the nicer errors & dedupe code

src/functions/env.ts
```diff
@@ -0,0 +1,229 @@
+import * as fs from "fs";
+import * as path from "path";
+
+import { FirebaseError } from "../error";
+import { logger } from "../logger";
+
+const RESERVED_KEYS = [
+  // Cloud Functions for Firebase
+  "FIREBASE_CONFIG",
```
inlined, line 9: Just found out that we also support CLOUD_RUNTIME_CONFIG as a way of passing Runtime Config data. We should disallow that one as well.

src/functions/env.ts
taeold: Played around with errors, and here's what I've got now:

```
i  deploying functions
Error: Failed to load environment variables from /Users/danielylee/google/cf3-env-vars/functions/.env.:
- Error Key foo must start with an uppercase ASCII letter or underscore, and then consist of uppercase ASCII letters, digits, and underscores.
- Error Key X_GOOGLE_FOO starts with a reserved prefix (X_GOOGLE_ or FIREBASE_)
```

Pretty happy w/ the output using `children` property on `FirebaseError`.

src/functions/env.ts
taeold: Updated to use `children` prop. Like the result.

PR #3626, opened 2021-08-03
Converts RC to TypeScript

mbleigh:
Also refactors RC so that the lodash-y methods are not used at all outside of the class itself to better encapsulate the behavior.

I ended up tackling this as preparation to do some work to clean up what happens if you specify a project number instead of id when using Firebase CLI.


src/rc.ts
```diff
@@ -0,0 +1,234 @@
...
+    return this.data.targets[project]?.[type]?.[name] || [];
+  }
+
+  applyTarget(project: string, type: string, targetName: string, resources: string | string[]) {
+    if (!TARGET_TYPES[type]) {
+      throw new FirebaseError(
+        `Unrecognized target type ${clc.bold(type)}. Must be one of ${Object.keys(
+          TARGET_TYPES
+        ).join(", ")}`,
+        { exit: 1 }
```
bkendall, line 118: `{ exit: 1 }` is the default, please don't include it

src/rc.ts
```diff
@@ -0,0 +1,234 @@
...
+        "Deploy target " +
+          clc.bold(name) +
+          " not configured for project " +
+          clc.bold(project) +
+          ". Configure with:\n\n  firebase target:apply " +
+          type +
+          " " +
+          name +
+          " <resources...>",
+        { exit: 1 }
```
bkendall, line 215: Again, don't need to specify `exit: 1`

bkendall (changes requested):
You'll need to use an older version of `npm` (latest version of `v6`?) to not break the package-lock check.

mbleigh:
> You'll need to use an older version of `npm` (latest version of `v6`?) to not break the package-lock check.

Done.

bkendall (approved):

PR #3629, opened 2021-08-03
Transition to generateUploadUrl

inlined:
The GenerateUploadUrl API is now live for v2. After talking to Diana, she said that deploys from uploads to different regions might work today, but they don't like it and want to add validation to break it in the future. At minimum it would potentially cause features like GenerateDownloadUrl and cloud console source viewing to break.


src/deploy/functions/deploy.ts
```diff
@@ -22,16 +23,20 @@ async function uploadSourceV1(context: args.Context): Promise<void> {
...
+async function uploadSourceV2(context: args.Context, region: string): Promise<void> {
+  const res = await gcfv2.generateUploadUrl(context.projectId, region);
   const uploadOpts = {
     file: context.functionsSource!,
     stream: fs.createReadStream(context.functionsSource!),
   };
-  context.storageSource = await gcs.uploadObject(uploadOpts, bucket);
+  await gcs.upload(uploadOpts, res.uploadUrl);
+  context.storage = context.storage || {};
+  context.storage[region] = res.storageSource;
```
taeold, line 39: nit* Not sure if I prefer this, but a single-liner:

```js
context.storage = { ...context.storage, ...{[region]: res.storageSource}};
```

src/deploy/functions/functionsDeployHelper.ts
```diff
@@ -106,19 +106,17 @@ export function printSuccess(func: backend.TargetIds, type: string) {
...
-    // TODO: way to filter out extensions deployed on GCFv2. May have to just replace
-    // the existing backend with operation results as functions deploy rather than
-    // calling existingBackend twice.
-    if (fn.platform == "gcfv1" && fn.sourceUploadUrl !== context.uploadUrl) {
+    if (backend.isEventTrigger(fn.trigger)) {
       return false;
     }
-    return !backend.isEventTrigger(fn.trigger) && deploymentTool.isFirebaseManaged(fn.labels || {});
+
+    return want.cloudFunctions.some(backend.sameFunctionName(fn));
```
taeold, line 119: Curious - is this fixing some existing bug? How so?

taeold (approved):

src/deploy/functions/deploy.ts
inlined: There's a simpler one-liner:

```
context.storage = {...context.storge, [region]: res.storageSource};
```

It definitely feels foreign to me, but I guess it's not bad. I can change.

src/deploy/functions/functionsDeployHelper.ts
inlined: It was meant to be an alternate implementation that works for v2 when v1's solution wont (we don't have an uploadURL and the storageSource is different than what we uploaded to GCS)

PR #3630, opened 2021-08-03
Stop supporting RuntimeConfig in v2

inlined:
Stop uploading .runtimeconfig.json for new functions APIs.


src/deploy/functions/prepareFunctionsUpload.ts
```diff
@@ -90,10 +90,12 @@ async function packageSource(options: Options, sourceDir: string, configValues:
         mode: file.mode,
       });
     });
-    archive.append(JSON.stringify(configValues, null, 2), {
-      name: CONFIG_DEST_FILE,
-      mode: 420 /* 0o644 */,
-    });
+    if (typeof configValues !== "undefined") {
```
taeold, line 93: q: Is this equivalent to:

```js
if (configValues !== undefined) {
...
```

Is there reason to prefer one over another?

src/deploy/functions/prepare.ts
```diff
@@ -80,7 +80,15 @@ export async function prepare(
       clc.bold(options.config.src.functions.source) +
       " directory for uploading..."
   );
-  context.functionsSource = await prepareFunctionsUpload(runtimeConfig, options);
+  if (wantBackend.cloudFunctions.find((fn) => fn.platform === "gcfv1")) {
+    context.functionsSourceV1 = await prepareFunctionsUpload(runtimeConfig, options);
+  }
+  if (wantBackend.cloudFunctions.find((fn) => fn.platform === "gcfv2")) {
+    context.functionsSourceV2 = await prepareFunctionsUpload(
```
taeold, line 87: Is there an opportunity for us to inform the user that runtimeconfig will be ignored for GCFv2 deploy?

taeold (approved):
LGTM, 1 comment and 1 question.

src/deploy/functions/prepareFunctionsUpload.ts
inlined: Though it doesn't seem to work in new versions of Node, I was raised in a world where `undefined` was a variable and not a literal, so you were always supposed to check the type of a value and not compare it to the variable `undefined`.

src/deploy/functions/prepare.ts
inlined: I don't think so? It's perfectly valid, though wonky, to use config in your v1 functions and not in your v2 functions. And I don't recall your migrate script deleting config either. Do you have a good way to decide when to warn?

PR #3635, opened 2021-08-04
Fix broken functions deploy e2e test.

taeold:
This isn't a test target that runs as part of CI; it's a useful one to run manually. Also including a small fix in the `functions/prepare.ts` script that I found after fixing the e2e test!

I'm removing the `unknownFilter` test because we now treat any filter as a valid one (Deploy code simply deploys nothing and reports "Deploy Complete"). This is subpar experience, but I've found fix to the issue to be a little more than few lines of change, and I'm not keen on fixing this bug right now.


src/deploy/functions/prepare.ts
```diff
@@ -56,7 +57,7 @@ export async function prepare(
   const wantBackend = await runtimeDelegate.discoverSpec(runtimeConfig, env);
   payload.functions = { backend: wantBackend };
   if (backend.isEmptyBackend(wantBackend)) {
-    return;
+    throw new FirebaseError("Found nothing to deploy. Is your functions source empty?");
```
inlined, line 60: We traditionally have explicitly not errored on an empty backend. This allows a customer to say `firebase init`, initialize the kitchen sink, and then still deploy before they get to editing their functions.

scripts/test-functions-deploy.js
```diff
@@ -329,10 +312,6 @@ var main = function () {
       console.log(clc.green("\u2713 Test passed: creating functions with filters"));
       return testDeleteWithFilter();
     })
-    .then(function () {
-      console.log(clc.green("\u2713 Test passed: deleting functions with filters"));
-      return testUnknownFilter();
```
inlined, line 334: Why are we yanking this test? Also, we might want to make this use async so it's not so confusing that the console.log is for the previous function block.

inlined (approved):

src/deploy/functions/prepare.ts
taeold: This is was a band-aid patch on current bug - right now, the Firebase CLI throws and error on empty `index.js` file:

```
=== Deploying to '...'..

i  deploying functions

Having trouble? Try again or contact support with contents of firebase-debug.log
```

It sounds like this is a bad fix. I'll revert this change and make a separate PR to fix the issue for empty backends.

scripts/test-functions-deploy.js
taeold: Yeah promise chain here is yucky, but it was written in the time when maybe it was in vogue? I'm not trying to refactor this test in any way - just want it to be not broken. If I'm actually making additions to this test, I will refactor but probably not worth it at this time.

taeold: Also, copying snippet from the PR description for why I'm removing this:

> I'm removing the unknownFilter test because we now treat any filter as a valid one (Deploy code simply deploys nothing and reports "Deploy Complete"). This is subpar experience, but I've found fix to the issue to be a little more than few lines of change, and I'm not keen on fixing this bug right now.

PR #3636, opened 2021-08-04
Better support for project numbers as identifiers

mbleigh:
After some recent reports from users that function deploys weren't working when a project was set to a project *number* instead of a project *id*, I decided to take it upon myself to improve that early-command resolution process for projects.

Long story short, I ended up:

1. Converting `getProjectId.js` into `projectUtils.ts`
2. Changing `getProjectId(p: string, allowNull: boolean)` to `needProjectId()` and `getProjectId()` to allow for less null-checking.
3. Adding logic to detect a fully-numeric `project` early in command processing and resolve it to a project id and number
4. Moving `getProjectNumber` to be `needProjectNumber` and putting it in `projectUtils`

Boy that escalated quickly


src/projectUtils.ts
```diff
@@ -0,0 +1,90 @@
+import { getFirebaseProject } from "./management/projects";
+import { RC } from "./rc";
+
+const _ = require("lodash");
+const clc = require("cli-color");
+const marked = require("marked");
+
+const { FirebaseError } = require("./error");
+
+export function getProjectId({
```
bkendall, line 10: Please document exported functions - I am unsure what project or projectId _should_ be coming from

src/projectUtils.ts
```diff
@@ -0,0 +1,90 @@
...
+  projectId?: string;
+  project?: string;
+}): string | undefined {
+  return projectId || project;
+}
+
+/**
+ * Tries to determine the correct projectId given current
+ * command context.
+ * @returns {String} The projectId
```
bkendall, line 23:
```suggestion
 * @return the projectId
```

src/projectUtils.ts
```diff
@@ -0,0 +1,90 @@
...
+        clc.bold("firebase use --add") +
+        ", then rerun this command.\n" +
+        "To list all the Firebase projects to which you have access, run " +
+        clc.bold("firebase projects:list") +
+        ".\n" +
+        marked(
+          "To learn about active projects for the CLI, visit https://firebase.google.com/docs/cli#project_aliases"
+        ),
+      {
+        exit: 1,
```
bkendall, line 58: This is the default - remove

src/emulator/controller.ts
```diff
@@ -370,7 +370,7 @@ export async function startAll(options: Options, showUI: boolean = true): Promis
 
   if (shouldStart(options, Emulators.HUB)) {
     const hubAddr = await getAndCheckAddress(Emulators.HUB, options);
-    const hub = new EmulatorHub({ projectId, ...hubAddr });
+    const hub = new EmulatorHub({ projectId: projectId!, ...hubAddr });
```
bkendall, line 373: Why do you need to assert the value here? rather than doing this, I'd rather the code check that it exists first and fail out otherwise. Doesn't look like it was _required_ before...

src/test/command.spec.ts
```diff
@@ -81,6 +82,51 @@ describe("Command", () => {
       const result = run();
       await expect(result).to.be.rejectedWith("foo");
     });
+
+    it("should resolve a numeric --project flag into a project id", async () => {
+      nock.disableNetConnect();
```
bkendall, line 87: If you *must* include these, you should but them in a before/after so that if the test fails it doesn't leave everything disabled (not that that should matter, but it may)

bkendall (changes requested):

src/emulator/controller.ts
```diff
@@ -548,7 +548,7 @@ export async function startAll(options: Options, showUI: boolean = true): Promis
     }
 
     const rc = dbRulesConfig.normalizeRulesConfig(
-      dbRulesConfig.getRulesConfig(projectId, options),
+      dbRulesConfig.getRulesConfig(projectId!, options),
```
bkendall, line 551: You still need the `!` here (and below)? It looks like it's guaranteed to be a string now, so it shouldn't be needed.

bkendall (approved):
One comment about the use of `!`, but LGTM

PR #3638, opened 2021-08-05
Enforce v2 name schemes

inlined:
Enforce v2 name schemes

### DIFF
--- a/src/deploy/functions/validate.ts
+++ b/src/deploy/functions/validate.ts
@@ -6,10 +6,6 @@ import * as backend from "./backend";
 import * as fsutils from "../../fsutils";
 import * as projectPath from "../../projectPath";
 
-// have to require this because no @types/cjson available
-// tslint:disable-next-line
-const cjson = require("cjson");
-
 /**
  * Check that functions directory exists.
  * @param options options object. In prod is an Options; in tests can just be {cwd: string}
@@ -53,8 +49,8 @@ export function functionIdsAreValid(functions: { id: string; platform: string }[
   });
   if (invalidV2Ids.length !== 0) {
     const msg =
-    `${invalidV2Ids.join(", ")} v2 function name(s) can only contin lower ` +
-    `case letters, numbers, hyphens, and not exceed 62 characters in length`;
+      `${invalidV2Ids.join(", ")} v2 function name(s) can only contin lower ` +
+      `case letters, numbers, hyphens, and not exceed 62 characters in length`;
     throw new FirebaseError(msg);
   }
 }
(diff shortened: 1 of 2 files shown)


src/deploy/functions/validate.ts
```diff
@@ -35,9 +35,15 @@ export function functionsDirectoryExists(
  * @param functionNames Object containing function names as keys.
  * @throws { FirebaseError } Function names must be valid.
  */
-export function functionIdsAreValid(functions: { id: string }[]): void {
-  const validFunctionNameRegex = /^[a-zA-Z0-9_-]{1,63}$/;
-  const invalidIds = functions.filter((fn) => !validFunctionNameRegex.test(fn.id));
+export function functionIdsAreValid(functions: { id: string; platform: string }[]): void {
```
mbleigh, line 38: Extraneous nit: There is now a `@types/cjson` and I actually already added it for a recent JS->TS refactor I did. Should be able to change to an import at the top.

src/deploy/functions/validate.ts
```diff
@@ -35,9 +35,15 @@ export function functionsDirectoryExists(
...
+  const v1FunctionName = /^[a-zA-Z][a-zA-Z0-9_-]{0,62}$/;
+  const v2FunctionName = /^[a-z][a-z-]{0,62}$/;
```
mbleigh, line 40:
```suggestion
  const v2FunctionName = /^[a-z][a-z0-9-]{0,62}$/;
```
I assume you can use numbers as well?

src/deploy/functions/validate.ts
```diff
@@ -35,9 +35,15 @@ export function functionsDirectoryExists(
...
+  const invalidIds = functions.filter((fn) => {
+    return (
+      (fn.platform === "gcfv1" && !v1FunctionName.test(fn.id)) ||
+      (fn.platform === "gcfv2" && !v2FunctionName.test(fn.id))
+    );
+  });
   if (invalidIds.length !== 0) {
     const msg =
       `${invalidIds.join(", ")} function name(s) can only contain letters, ` +
```
mbleigh, line 49: I think you need to clarify the error message for the v2 case (add a specific "v2 function names can only contain lowercase letters, numbers, and hyphens") or it will be confusing.

mbleigh (approved):
Couple small nits but don't need another pass after they're fixed.

src/deploy/functions/validate.ts
inlined: Actually, cjson isn't even in this file anymore. Yanked.

src/deploy/functions/validate.ts
inlined: Fixed.

PR #3643, opened 2021-08-05
Add retries for Docker APIs

### LINKED ISSUE #3639: Frequent failures when deploying functions (Unhandled error cleaning up build images)
###  Environment info
**firebase-tools:** 9.16.0

**Platform:**macOS

### [REQUIRED] Test case

Try deploying functions
```
```
functions fail with "Unhandled error cleaning up build images."
```
 functions: cleaning up build files...
⚠  functions: Unhandled error cleaning up build images. This could result in a small monthly bill if not corrected. You can attempt to delete these images by redeploying or you can delete them manually at https://console.cloud.google.com/gcr/images/api-dev/us/gcf
```

Should not fail deploying this frequently.

For deploying all my functions(around 50) per environment, I need to redeploy nearly 6 times to finish it. It keeps failing frequently from past few days. 

 I'm currently having 4 environments and attempting nearly **30 times** to finish deploying!

inlined:
Blind attempt to fix #3639 before I get debug logs.

Normally I'd worry that we hit a rate limit, but according to [docs](https://cloud.google.com/container-registry/quotas) the rate limit is 50K requests per 10m. Even with an API this chatty it's unlikely to hit the limit...

Without knowing any more info, we can realize that an API as chatty as Docker is likely to fail just through statistics. Deleting 15 functions would probably require 60-75 API calls. If gcr.io is a 3-9s service, that's a 6% chance of failure.

### DIFF
--- a/src/deploy/functions/containerCleaner.ts
+++ b/src/deploy/functions/containerCleaner.ts
@@ -40,6 +40,25 @@ export const SUBDOMAIN_MAPPING: Record<string, string> = {
   "australia-southeast1": "asia",
 };
 
+async function retry<Return>(func: () => Promise<Return>): Promise<Return> {
+  const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+  const MAX_RETRIES = 3;
+  const INITIAL_BACKOFF = 100;
+  let retry = 0;
+  while (true) {
+    try {
+      return await func();
+    } catch (error) {
+      logger.debug("Failed docker command with error", error);
+      retry += 1;
+      if (retry >= MAX_RETRIES) {
+        throw new FirebaseError("Failed to clean up artifacts", { original: error });
+      }
+      await sleep(Math.pow(INITIAL_BACKOFF, retry - 1));
+    }
+  }
+}
+
 export async function cleanupBuildImages(functions: backend.FunctionSpec[]): Promise<void> {
   utils.logBullet(clc.bold.cyan("functions: ") + "cleaning up build files...");
   const gcrCleaner = new ContainerRegistryCleaner();
@@ -258,7 +277,7 @@ export class DockerHelper {
 
   async ls(path: string): Promise<Stat> {
     if (!this.cache[path]) {
-      const raw = await this.client.listTags(path);
+      const raw = await retry(() => this.client.listTags(path));
       this.cache[path] = {
         tags: raw.tags,
         digests: Object.keys(raw.manifest),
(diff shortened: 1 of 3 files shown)


src/deploy/functions/containerCleaner.ts
```diff
@@ -39,6 +40,25 @@ const SUBDOMAIN_MAPPING: Record<string, string> = {
   "australia-southeast1": "asia",
 };
 
+async function retry<Return>(func: () => Promise<Return>): Promise<Return> {
```
mbleigh, line 43: I'm a little surprised that we don't have a generic retry library in the CLI at this point (I thought we did), but looks like we don't so 🤷

src/deploy/functions/containerCleaner.ts
```diff
@@ -39,6 +40,25 @@ const SUBDOMAIN_MAPPING: Record<string, string> = {
...
+  const INITIAL_BACKOFF = 100;
+  let retry = 0;
+  while (true) {
+    try {
+      return await func();
+    } catch (error) {
+      logger.debug("Failed docker command with error", error);
+      retry += 1;
+      if (retry >= MAX_RETRIES) {
+        throw new FirebaseError("Failed to clean up artifacts", { original: error });
```
mbleigh, line 55: Should we actually consider it an error if we fail to clean up artifacts, or should we print a warning? I'm not sure I'd want to consider it an error, especially since until very recently we didn't clean them up at all...

mbleigh (approved):

src/deploy/functions/containerCleaner.ts
taeold: +1 for warning

src/deploy/functions/containerCleaner.ts
```diff
@@ -39,6 +40,25 @@ const SUBDOMAIN_MAPPING: Record<string, string> = {
...
+      }
+      await sleep(Math.pow(INITIAL_BACKOFF, retry - 1));
```
taeold, line 57:
```suggestion
      await sleep(Math.pow(INITIAL_BACKOFF, retry));
```

Otherwise, it will sleep for 1ms on first iteration I think.

taeold (approved):

src/deploy/functions/containerCleaner.ts
inlined: And making it a "real" retry function would involve parameterizing things, so I avoided it in my rush.

src/deploy/functions/containerCleaner.ts
inlined: SGTM. I will catch the error higher up and warn.

src/deploy/functions/containerCleaner.ts
inlined: It's intentional that the first retry does not have a backoff. Not all failures need congestion control (and if this library really was about congestion control I'd add jitter)

src/deploy/functions/containerCleaner.ts
inlined: Turns out I already did.

PR #3645, opened 2021-08-06
Deploy environment variables in dotenv files for functions

taeold:
Add support for setting up environment variables for functions using dotenv files.

On deploy, Firebase CLI looks for naive dotenv file (`.env`) file or project-specific ones (`.env.<project id>` or `.env.<project alias>`) files, loads environment variables, and injects them on deployed functions. Environment variable set in project-specific dotenv file wins, and you cannot have both `.<project id>` and `<.project alias>` files.

Environment variables are also loaded up on Firebase Emulators. Emulator will look for `.env` and `.env.local` files and ignore project-specific dotenv files.

We are hiding the feature behind `dotenv` preview flag. We want to get a feeling for the development experience from our users before an official launch + documentations.

Also including a small e2e test for env var support. It's meant to run manually by the Firebase team via command `node scripts/test-functions-env.js`.


scripts/emulator-tests/functionsEmulatorRuntime.spec.ts
```diff
@@ -457,6 +461,42 @@ describe("FunctionsEmulator-Runtime", () => {
 
       expect(res.var).to.eql("localhost:9099");
     }).timeout(TIMEOUT_MED);
+
+    it("should inject user environment variables when preview is enabled", async () => {
+      // Setup
+      const restore = previews.dotenv;
```
mbleigh, line 467: I'd recommend extracting this out into a describe block with a `beforeEach` and `afterEach` so that if the test fails it still properly restores afterwards.

src/deploy/functions/deploymentPlanner.ts
```diff
@@ -73,12 +74,13 @@ export function calculateRegionalFunctionChanges(
 
       checkForInvalidChangeOfTrigger(fn, haveFn);
 
-      // Remember old environment variables that might have been set with
-      // gcloud or the cloud console.
-      fn.environmentVariables = {
-        ...haveFn.environmentVariables,
-        ...fn.environmentVariables,
-      };
+      if (!previews.dotenv) {
```
mbleigh, line 77: I wonder about how we should handle the case where someone has set env OOB in today's world. I think this is probably reasonable behavior (so long as we call it out) for a preview flag, but once we're ready to unflag I think it'd be a breaking change to suddenly start wiping out existing env variables.

Perhaps we should only "forget" these if the preview is enabled **and** at least one `.env` file exists locally?

src/deploy/functions/prepare.ts
```diff
@@ -4,19 +4,48 @@ import { Options } from "../../options";
...
+}): Record<string, string> {
+  const { firebaseConfig, functionsSource, projectId, projectAlias } = options;
+
+  let envs = {
+    FIREBASE_CONFIG: JSON.stringify(firebaseConfig),
+    GCLOUD_PROJECT: projectId,
+  };
+  if (previews.dotenv) {
+    envs = {
+      ...envs,
```
mbleigh, line 35: Since we never want to allow a customer to override this, even though the logic of `functionsEnv.load` won't allow these to be set I'd still prefer to put `...envs` last.

src/deploy/functions/prepare.ts
```diff
@@ -51,10 +80,21 @@ export async function prepare(
   const firebaseConfig = await functionsConfig.getFirebaseConfig(options);
   context.firebaseConfig = firebaseConfig;
   const runtimeConfig = await getFunctionsConfig(context);
-  const env = await getEnvs(context);
+
+  utils.assertDefined(
+    options.config.src.functions.source,
```
mbleigh, line 85: Why did you add this assertion? Is this whole thing safe to do without optional chaining? What happens if you don't have functions set up but you're deploying e.g. hosting, won't this throw?

src/deploy/functions/prepare.ts
```diff
@@ -51,10 +80,21 @@ export async function prepare(
...
+  const envs = getEnvs({
+    firebaseConfig: runtimeConfig,
+    functionsSource: options.config.path(source),
+    projectId: projectId,
+    projectAlias: options.projectAlias,
+  });
 
   logger.debug(`Analyzing ${runtimeDelegate.name} backend spec`);
-  const wantBackend = await runtimeDelegate.discoverSpec(runtimeConfig, env);
+  const wantBackend = await runtimeDelegate.discoverSpec(runtimeConfig, envs);
```
mbleigh, line 97: Do we want the specified env values to be available during discovery? @inlined

inlined (commented):
I got partway through the review. Sharing what I have because I don't know where I was and need to start over.

src/deploy/functions/prepare.ts
taeold: This was meant to move existing assertion made couple of lines below this one (while rebasing, looks like I forgot to delete the existing assertion).

https://github.com/firebase/firebase-tools/blob/98ff7f72f9abad53dc2d675b61171f59382df90f/src/deploy/functions/prepare.ts#L79-L83

IIUC, behavior today is that if your firebase.json does not have `functions` attribute, it won't attempt to deploy any functions hence will never hit this code. Is there a different scenario you are referring to here?

taeold: On second scan, this check seems unnecessary since we always provide a default for `functions.source` when missing:

https://github.com/firebase/firebase-tools/blob/98ff7f72f9abad53dc2d675b61171f59382df90f/src/config.ts#L78-L86

I'm going to opt on removing this check all togther.

taeold:
Emulator test appears broken, but it works on my machine 😢 . I'm debugging the issue, but I'd appreciate one more round of review - I made some hefty changes to support "overwrite envs only if there are any dotenv files."

I see a lot of opportunity to refactor env-related logic. I'm not touching it her to make the PR small-ish, but I'm going to find time this week to put in the refactoring work.

src/deploy/functions/prepare.ts
inlined: Let's start with "no" and see how much pushback we get. I'm nervous that allowing Env variables will mean that we've regressed to where parameter discovery is dynamic again.

src/functions/env.ts
```diff
@@ -191,49 +219,70 @@ function parseStrict(data: string): Record<string, string> {
...
-        if (path.basename(p) === `.env.${options.projectAlias}`) {
-          throw new FirebaseError(
-            `Can't have both .env.${options.projectId} and .env.${options.projectAlias}> files.`
-          );
-        }
-      }
+  // Disallow setting both .env.<projectId> and .env.<projectAlias>
+  if (projectAlias) {
+    if (envFiles.includes(`.env.${projectId}`) && envFiles.includes(`.env.${projectAlias}`)) {
+      throw new FirebaseError(`Can't have both .env.${projectId} and .env.${projectAlias}> files.`);
```
inlined, line 236: I wonder if we should make it clear that they can't have those files because they can't load files by both project ID and project alias. I'd be confused if I was told `Can't have both .env.inlined-junkdrawer and .env.dev> files`

Also, is the ">" a typo?

src/functions/env.ts
```diff
@@ -191,49 +219,70 @@ function parseStrict(data: string): Record<string, string> {
...
         exit: 2,
         children: err.children?.length > 0 ? err.children : [err],
       });
     }
   }
-  logger.debug(
-    `Loaded environment variables ${JSON.stringify(envs)} from ${targetPaths.join(",")}.`
+  logBullet(
+    clc.cyan.bold("functions: ") +
+      `Loaded environment variables ${JSON.stringify(envs)} from ${envFiles.join(", ")}.`
```
inlined, line 254: I don't know that we need to fully dump the list of environment variables. This could make the CLi very messy very quickly.

src/functions/env.ts
```diff
@@ -191,49 +219,70 @@ function parseStrict(data: string): Record<string, string> {
...
   return envs;
 }
+
+/**
+ * Load environment variables for the given project.
+ *
+ * @return Environment varibles for functions.
+ */
+export function load(options: {
+  firebaseConfig: { [key: string]: any };
```
inlined, line 266: Record<string, any>

inlined (changes requested):

src/deploy/functions/prepare.ts
taeold: Oh I thought having env vars available during trigger discovery was something that we wanted to support. E.g.

```js
functions.runWith({ minInstances: process.env.DEPLOY_ENV === "prod" ? 5 : 0 })...
```

Are we saying we don't want this anymore?

src/functions/env.ts
taeold: Yes on typo. Agreed on the confusion - added more detailed message.

src/deploy/functions/prepare.ts
taeold: Offline - no won't support this usecase.

inlined (approved):

PR #3646, opened 2021-08-06
Track functions deploys

inlined:
Per offline conversation, this now tracks individual function deploy timing by trigger "tag".

We also track overall codebase deploys, tagged by API type.

### DIFF
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -96,6 +96,29 @@ export function isEventTrigger(trigger: HttpsTrigger | EventTrigger): trigger is
   return "eventType" in trigger;
 }
 
+/** Friendly name to label a function in stats */
+export function triggerTag(fn: FunctionSpec) {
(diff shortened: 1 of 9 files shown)


mbleigh (approved):
Please fix lint errors or lobby with @bkendall to remove any you don't like 😛

inlined:
Heh. I remember when I first dove back into this codebase and made a big deal about the linter errors. I think we should get rid of a lot of the warnings and then make the rest of them blocking. IIRC @bkendall and i agreed to get rid of @param and @return requirements because they often turn into nonsense javadocs. E.g.

```
/**
  *  Returns a human-readable name with MB or GB suffix for a MemoryOption (MB)
  * @return a human-readable name with MB or GB suffix for a MemoryOption
  */
export function memoryOptionDisplayName(option: MemoryOptions): string;
```

I don't see @param warnings anymore and we should get rid of the @return warnings too. I added the JSDoc comment for `memoryOptionDisplayName` though.

PR #3657, opened 2021-08-10
Increase timeout on detectFromPort to reduce flakes.

taeold:
detectFromPort test reliably fails in Cloud Build environment. Increasing timeout to compensate.


taeold:
Reverting in favor of https://github.com/firebase/firebase-tools/pull/3658

PR #3658, opened 2021-08-10
Replace fake server with nock

inlined:
The discoverFromPort test is causing too many flakes. While it was neat that the test actually worked with a real webserver, nock is better for every other developer.


mbleigh (approved):

bkendall (approved):

bkendall (changes requested):
Remove the imports that are no longer used, please

bkendall (approved):

PR #3666, opened 2021-08-12
Fix require issue with new firebase-functions version.

mbleigh:


inlined (approved):
We can figure out tomorrow whether we even need the versioned import, but this would stem the bleeding

PR #3672, opened 2021-08-13
Display releaseNotes during ext:update, and streamline ext:update code

joehan:
### Description
Display releaseNotes during ext:update if they exist.

I also removed most of the code supporting old official extensions. @huangjeff5 and I checked the backend, and all instances have been migrated to the new registry, so these code paths should be totally unused and safe to remove at this point. 
### Scenarios Tested
Update containing a major version:
Screen Shot 2021-08-12 at 5 02 08 PM

Update without a major version:
Screen Shot 2021-08-12 at 5 03 43 PM
 
Updating when releaseNotes is not returned by the backend:
Screen Shot 2021-08-12 at 5 54 50 PM


src/commands/ext-update.ts
```diff
@@ -229,10 +190,8 @@ export default new Command("ext:update <extensionInstanceId> [updateSource]")
           return;
         }
       }
-      const isOfficial =
-        newSourceOrigin === SourceOrigin.OFFICIAL_EXTENSION ||
-        newSourceOrigin === SourceOrigin.OFFICIAL_EXTENSION_VERSION;
-      await displayChanges(existingSpec, newSpec, isOfficial);
+
+      await displayChanges(existingSpec, newSpec, false);
```
huangjeff5, line 194: I remember this logic was supposed to display before/after sources only if the extension wasn't official, I believe. Worth checking with Sonakshi to see if it's okay to deprecate this logic.

If it is okay to deprecate that logic, then I think we can remove the last parameter altogether from displayChanges()

elvisun:
UX Question: If I see 0.2.0 and don't want to upgrade to that, but want to upgrade to 0.1.26 instead, is that a supported flow?

src/extensions/changelog.ts
```diff
@@ -0,0 +1,85 @@
...
+
+/*
+ * getReleaseNotesForUpdate fetches all version between toVersion and fromVersion and returns the relase notes
+ * for those versions if they exist.
+ * @param extensionRef
+ * @param fromVersion the version you are updating from
+ * @param toVersion the version you are upodating to
+ * @returns a Record of version number to releaseNotes for that version
+ */
+export async function getReleaseNotesForUpdate(
```
elvisun, line 23: prefer named params to reduce confusion at call site
go/tsstyle#comments-when-calling-a-function

src/extensions/changelog.ts
```diff
@@ -0,0 +1,85 @@
...
+  extensionRef: string,
+  fromVersion: string,
+  toVersion: string
+): Promise<Record<string, string>> {
+  const releaseNotes: Record<string, string> = {};
+  const filter = `id<="${toVersion}" AND id>"${fromVersion}"`;
+  const extensionVersions = await listExtensionVersions(extensionRef, filter);
+  for (const extensionVersion of extensionVersions) {
```
elvisun, line 31: This assumes `listExtensionVersions` returns in a sorted manner, which is not part of the contract from our backend
https://source.corp.google.com/piper///depot/google3/firebase/mods/go/spanner/db.go;l=1330;rcl=391185722

So it'd be safer to sort it here so users are guaranteed to see all versions in incremental order

src/extensions/changelog.ts
```diff
@@ -0,0 +1,85 @@
...
+  return releaseNotes;
+}
+
+/**
+ * displayReleaseNotes prints out a nicely formatted table containing all release notes in an update.
+ * If there is a major version change, it also prints a warning and highlights those release notes.
+ */
+export function displayReleaseNotes(releaseNotes: Record<string, string>, fromVersion: string) {
+  const versions = [fromVersion].concat(Object.keys(releaseNotes));
+  const breaks = breakingChangesInUpdate(versions);
```
elvisun, line 46: this is kinda confusing with `break`, maybe `breakingVersions`?

src/extensions/changelog.ts
```diff
@@ -0,0 +1,85 @@
...
+ */
+export function breakingChangesInUpdate(versionsInUpdate: string[]): string[] {
+  const breaks: string[] = [];
+  const semvers = versionsInUpdate.map((v) => semver.parse(v)!).sort(semver.compare);
+  for (let i = 1; i < semvers.length; i++) {
+    if (
+      semvers[i - 1].major < semvers[i].major ||
+      (semvers[i - 1].major == 0 &&
+        semvers[i].major == 0 &&
+        semvers[i - 1].minor < semvers[i].minor)
```
elvisun, line 79: This is pretty scary to read, can you extract the variables in `if` to const with some meaningful names? Something like:
hasMajorBump || hasMinorBumpInPreview

src/extensions/updateHelper.ts
```diff
@@ -279,23 +254,12 @@ export async function updateToVersionFromPublisherSource(
...
-    // If the publisher and extension ID both match, we know it's an official extension (i.e. it's specifically listed in our Registry File)
-    // Otherwise, it's simply a published extension in the Registry
     registryEntry = await resolveSource.resolveRegistryEntry(existingSpec.name);
-    sourceOrigin =
-      registryEntry.publisher === refObj.publisherId
-        ? SourceOrigin.OFFICIAL_EXTENSION
-        : SourceOrigin.PUBLISHED_EXTENSION;
   } catch (err) {
-    sourceOrigin = SourceOrigin.PUBLISHED_EXTENSION;
+    // Ignore errors if we can't fetch the registry
```
elvisun, line 260: is this registry.json?

silently failing worries me... do we have some kind of error reporting for the CLI?

src/test/extensions/changelog.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+        "0.1.1": "foo",
+        "0.1.2": "bar",
+      };
+
+      const got = await changelog.getReleaseNotesForUpdate("test/test", "0.1.0", "0.1.2");
+
+      expect(got).to.deep.equal(want);
+    });
+
+    it("should exclude versions that don't have releaseNotes", async () => {
```
elvisun, line 59: Or should we say no release note? providers forgetting to write release note might be pretty common

src/extensions/updateHelper.ts
joehan: Adding a debug log here, but I think this is necessary to ignore. We want to be able to take down the registry at some point without breaking this. In the case of an outage before then, this would just accidentally allow updates below minVersion, which is a small corner case that we're already in the process of migrating to the backend.

src/test/extensions/changelog.spec.ts
joehan: Gonna leave this as is since:
- We're gonna check for a CHANGELOG entry and enforce that it should be there on ext:publish, so this should not happen frequently
- In the case where an extension doesn't have release notes populated on the backend, it may still have release notes available elsewhere, so it could be misleading.

src/commands/ext-update.ts
joehan: Cleaned this up as well, good eye!

joehan:
@elvisun There is! firebase ext:update <instance-id> 0.1.26 -P my-proj

huangjeff5 (approved):

src/test/extensions/changelog.spec.ts
elvisun: sgtm, thanks!

src/extensions/updateHelper.ts
elvisun: sg with noop just wanted to get some logs here in case we need to figure out how often this happens

src/test/extensions/changelog.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    });
+
+    afterEach(() => {
+      listExtensionVersionStub.restore();
+    });
+
+    it("should return release notes for each version in the update", async () => {
+      const extensionVersions: extensionApi.ExtensionVersion[] = [
+        testExtensionVersion("0.1.1", "foo"),
+        testExtensionVersion("0.1.2", "bar"),
```
elvisun, line 44: flip this to test the sort

elvisun (approved):

src/test/extensions/changelog.spec.ts
joehan: Oooh, good idea!

joehan: Ahhh, on second though it won't actually cover that - the sorting happens in a different method that the one under test.

PR #3684, opened 2021-08-13
Fix an issue where emulator:exec would not work without a project ID

### LINKED ISSUE #3681: `firebase emulators:exec` failing in 9.16.1, 9.16.2, 9.16.3 and 9.16.4
### Environment info

**firebase-tools:** 9.16.4

**Platform:** Ubuntu

### Test case

I have a test suite running in CI using `firebase-tools`. Here’s the command:
`firebase emulators:exec --only firestore 'npm run jest'`.

These tests ran and passed with firebase-tools 9.16.0 (and earlier) ([most recent successful run](https://github.com/delucis/bgio-firebase/actions/runs/1122835651)).

When upgrading to 9.16.2 or 9.16.4, the CI fails with a “No currently active project” error. Here’s [the error in the PR attempting to upgrade to 9.16.4](https://github.com/delucis/bgio-firebase/runs/3324601387?check_suite_focus=true#step:7:17).

My guess is that https://github.com/firebase/firebase-tools/pull/3636 is the culprit.

### Steps to reproduce

1. Upgrade firebase-tools to a version greater than 9.16.0
2. Run a command with `firebase emulators:exec`

### Expected behavior

The command continues to run as with <=9.16.0

### Actual behavior

The command fails with the following error:

```
Error: No currently active project.
To run this command, you need to specify a project. You have two options:
- Run this command with --project <alias_or_project_id>.
- Set an active project by running firebase use --add, then rerun this command.
To list all the Firebase projects to which you have access, run firebase projects:list.
To learn about active projects for the CLI, visit firebase.google.com/docs/cli#project_aliases
```

joehan:
### Description
Fix an issue where emulator:exec would not work without a project ID. Fixes #3681 

### Scenarios Tested
I'm able to run `firebase emulators:exec test.js` without the error described in #3681, as long as the functions emulator is not running. This matches the behavior from <9.16.0


src/emulator/controller.ts
```diff
@@ -330,7 +329,7 @@ export async function startAll(options: Options, showUI: boolean = true): Promis
   const hubLogger = EmulatorLogger.forEmulator(Emulators.HUB);
   hubLogger.logLabeled("BULLET", "emulators", `Starting emulators: ${targets.join(", ")}`);
 
-  const projectId: string = needProjectId(options);
+  const projectId: string = getProjectId(options) || ""; // TODO: Next breaking change, consider making this fall back to demo project.
```
joehan, line 332: I think we should strongly consider giving this a sane default, but I'm afraid to change the behavior at all given our recent run ins with Hyrum's law.

yuchenshi (approved):

PR #3685, opened 2021-08-14
Set maximum backoff to 10s for OperationPoller

kaibolay:
Previously it used 60s (which is the default).

This change doesn't make much of a difference when using the default timeout of 30s, but for longer timeouts it still makes sense to poll frequently to minimize excessive delays for interactive (human) users.

### DIFF
--- a/src/operation-poller.ts
+++ b/src/operation-poller.ts
@@ -13,7 +13,8 @@ export interface OperationPollerOptions {
 }
 
 const DEFAULT_INITIAL_BACKOFF_DELAY_MILLIS = 250;
-const DEFAULT_MASTER_TIMEOUT_MILLIS = 30000;
+const MAX_BACKOFF_MILLIS = 10_000;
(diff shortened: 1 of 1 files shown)


lfkellogg:
LGTM. Agreed that 10 seconds seems like a more appropriate max backoff when polling in a CLI tool.

bkendall (changes requested):
1. Could you adjust this change to have an added field in `OperationPollerOptions` for maximum backoff, and set that option where a shorter max backoff would be useful? That way we don't have to audit all the places where this poller is used and make a determination about each one (it probably isn't many places, but I'm a little concerned about it). What do you think?
1. Please include a `CHANGELOG.md`

lfkellogg:
@bkendall Sounds good. I've put up a new PR with a change that doesn't actually modify the backoff of any existing calls, but will let us configure it on a new call we're adding: https://github.com/firebase/firebase-tools/pull/3687

lfkellogg:
Closing this in favor of https://github.com/firebase/firebase-tools/pull/3687

PR #3687, opened 2021-08-16
Make max backoff configurable in OperationPoller

lfkellogg:
### Description

https://github.com/firebase/firebase-tools/pull/3670 uses the OperationPoller to poll for the status of an uploaded distribution. The default 60 second max backoff is too long for this purpose, since customers could have to wait up to an additional minute while their binary is processed: https://github.com/firebase/firebase-tools/blob/037b7ea3499a2a7b4aed8ea6b248be977491027a/src/throttler/throttler.ts#L77

Making this parameter configurable is a precondition to shipping that PR.

This is a replacement for https://github.com/firebase/firebase-tools/pull/3685 which set the max backoff to 10 seconds globally. This leaves the default as 60 seconds and therefore does not change any existing behavior.

### Scenarios Tested

Tested project creation, which uses the OperationPoller:

```
firebase projects:create lee-tests-project-creation --folder 261046259366
```

### DIFF
--- a/src/operation-poller.ts
+++ b/src/operation-poller.ts
@@ -8,6 +8,7 @@ export interface OperationPollerOptions {
   apiVersion: string;
   operationResourceName: string;
   backoff?: number;
+  maxBackoff?: number;
   masterTimeout?: number;
   onPoll?: (operation: OperationResult<any>) => any;
 }
@@ -39,6 +40,7 @@ export class OperationPoller<T> {
       name: options.pollerName || "LRO Poller",
       concurrency: 1,
       retries: Number.MAX_SAFE_INTEGER,
+      maxBackoff: options.maxBackoff,
       backoff: options.backoff || DEFAULT_INITIAL_BACKOFF_DELAY_MILLIS,
     });
 


src/operation-poller.ts
```diff
@@ -39,6 +41,7 @@ export class OperationPoller<T> {
       name: options.pollerName || "LRO Poller",
       concurrency: 1,
       retries: Number.MAX_SAFE_INTEGER,
+      maxBackoff: options.maxBackoff || DEFAULT_MAX_BACKOFF_MILLIS,
```
bkendall, line 44: I don't think it's necessary to include `DEFAULT_MAX_BACKOFF_MILLIS` here. Since `maxBackoff` has a default in the Queue class, we don't need to duplicate that here (which is all that it's doing).

bkendall (changes requested):

src/operation-poller.ts
lfkellogg: Sounds good, done.

bkendall (approved):

Issue #3689, opened 2021-08-18
firebase deploy --only functions = [ERR_UNSUPPORTED_ESM_URL_SCHEME]


fouchekeagan:
### [REQUIRED] Environment info

**firebase-tools:** 9.16.5

**Platform:** Windows 10

**Node:** 14.17.3

**NPM:** 6.14.13

Package.json Dep:

**Firebase:** 8.9.1
**@google-cloud/functions-framework:** 1.9.0
**firebase-functions:** 3.15.4
**Typescript:** 4.3.5

### [REQUIRED] Steps to reproduce

1. Upgrade from node 10 to node 14 (update from commonJS to ESM).
2. Attempt to deploy.

### [REQUIRED] Expected behavior

The firebase functions to be deployed.

### [REQUIRED] Actual behavior

Deploy fails while running pre-deploy script:  
```Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'```

![image](https://user-images.githubusercontent.com/72212540/129882917-a09c0c88-1efc-4370-935d-b70956f9116d.png)

Additional info: 
The same is shown if I specify a single function.
We had previously an issue with starting the emulators when initially updating to Node14 - this was resolved by #2994 /  #3573 /  #3574 (Thanks Daniel Lee!)
I can start the emulators and use the functions locally.

**Comment by taeold** [comment]
@fouchekeagan Thanks for reporting the issue!

It looks like I forgot to apply the fix to module path issue on Windows (https://github.com/firebase/firebase-tools/pull/3574) to the code path relevant to function deploys (https://github.com/firebase/firebase-tools/blob/7f99ac5868cf2c2265ff0923c8889d90c16c51c5/src/deploy/functions/runtimes/node/triggerParser.js#L24) 🤦‍♂️

**Comment by taeold** [comment]
I will try to get the fix merged soon so that the bug is fixed by next CLI release.

In the meantime, if anyone is willing to test out the fix (I don't have a Windows machine), I'd appreciate it:

```
> git clone https://github.com/firebase/firebase-tools.git
> cd firebase-tools
> git checkout origin/dl-cf3-esm-windows-fix2 
> npm i
> npm run build
> npm link # firebase cli should now point to locally built one
# cd to your functions project
> firebase deploy --only functions
```

fouchekeagan:
[...]

PR #3692, opened 2021-08-19
Fix bug where ESM module load fails on windows.

taeold:
Patch allows Windows users to deploy Firebase Functions packaged as ES module.

We apply the same fix we made for Windows ES module support in the Functions Emulator (https://github.com/firebase/firebase-tools/pull/3574) to the triggerParser script t
Fixes https://github.com/firebase/firebase-tools/issues/3689

### DIFF
--- a/src/deploy/functions/runtimes/node/triggerParser.js
+++ b/src/deploy/functions/runtimes/node/triggerParser.js
@@ -2,6 +2,7 @@
 // from a functions package directory.
 "use strict";
 
+var url = require("url");
 var extractTriggers = require("./extractTriggers");
 var EXIT = function () {
   process.exit(0);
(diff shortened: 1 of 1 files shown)


ZYinMD:
@fouchekeagan When writing your a blog, you might want to see this: https://github.com/microsoft/vscode/issues/131837. I'd like to see how you set up your per-folder vscode settings.

web2033:
E-O_KcpXoAIsGSE

This is the minimal 'Hello World' example I came up with (Node 16 and ESM)
https://github.com/web2033/hello-esm-firebase

If I remove `@google-cloud/functions-framework` it won't deploy.

taeold:
@web2033 I was able to deploy my function packaged as an ES module without including the `@google-cloud/functions-framework` package.

Maybe GCF team is slowing rolling out new version of the buildpack that includes the latest version of the `@google-cloud/functions-framework`, and I got lucky to be included in the early release batch.

web2033:
Node 16?

taeold:
@web2033 Oh interesting. Deploying Node 16 function fails for me as well. Node 14 works.

web2033:
Do we need an issue to track it? In which package?

taeold:
@web2033 The change needs to happen on Google Cloud Functions backend team. I've let them know - will ping this thread when the change is rolled out. If you prefer following an issue, feel free to open one in firebase-functions or open a public one is GCF.

Issue #960, opened 2018-10-18
firebase init cannot read property 'projectId'


adrielwerlich:
### Platform Information

On Windows

### Steps to reproduce

firebase init

### Expected behavior

create config file firebase.json  at project root

### Actual behavior

You're about to initialize a Firebase project in this directory:

  D:\projetos\portfolio_project\jobs\templates

[info] 
=== Project Setup
[info] 
[info] First, let's associate this project directory with a Firebase project.
[info] You can create multiple project aliases by running firebase use --add, 
[info] but for now we'll just set up a default project.
[info] 
[debug] [2018-10-18T15:03:51.589Z] >>> HTTP REQUEST GET https://firebase.googleapis.com/v1beta1/projects?page_size=100  
 
[debug] [2018-10-18T15:03:53.158Z] <<< HTTP RESPONSE 200
[debug] [2018-10-18T15:03:53.795Z] **TypeError: Cannot read property 'projectId' of undefined**
    at C:\Users\adrie\AppData\Roaming\npm\node_modules\firebase-tools\lib\init\features\project.js:48:23
    at arrayMap (C:\Users\adrie\AppData\Roaming\npm\node_modules\firebase-tools\node_modules\lodash\lodash.js:639:23)
    at Function.map (C:\Users\adrie\AppData\Roaming\npm\node_modules\firebase-tools\node_modules\lodash\lodash.js:9556:14)
    at C:\Users\adrie\AppData\Roaming\npm\node_modules\firebase-tools\lib\init\features\project.js:46:21
    at process._tickCallback (internal/process/next_tick.js:68:7)
[error] 
[error] Error: An unexpected error has occurred.

mbleigh:
Can you tell me -- how many Firebase projects do you have on your account?

adrielwerlich:
I have 2 google accounts and was trying to access through the cli interface the one that had no firebase projects, that´s why no projects were showing...

mbleigh:
Still looks like a bug we should fix :)

ian:
This is still an issue.  Downgrading to 5.0.1 fixes this.  

The cause is firebase-tools/lib/init/features/project.js:49:23 fails, the array has the following:
`[ undefined ]`

Vloz:
Same problem here, as mentionned bi Ian downgrading to 5.0.1 remove the error.

PR #3696, opened 2021-08-23
OnRequest Access Policy Invoker - move inside trigger block

colerogers:
### Description

This is the CLI change that corresponds with [#960](https://github.com/firebase/firebase-functions/pull/960). This change moves the `invoker` into the `HttpsTrigger` type from the root `FunctionSpec`.

### Scenarios Tested

- Parsing trigger correctly

### Sample Commands


inlined (approved):
Thanks for doing this!

PR #3704, opened 2021-08-24
Track usage of runtime config and dotenv on function deploys.

taeold:
Also includes a small refactor to make arguments for user env functions consistent.

### DIFF
--- a/src/functions/env.ts
+++ b/src/functions/env.ts
@@ -191,20 +191,18 @@ function findEnvfiles(functionsSource: string, projectId: string, projectAlias?:
     .map((p) => path.basename(p));
 }
 
+export interface UserEnvsOpts {
+  functionsSource: string;
+  projectId: string;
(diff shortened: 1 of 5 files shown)


src/functions/env.ts
```diff
@@ -196,11 +196,15 @@ function findEnvfiles(functionsSource: string, projectId: string, projectAlias?:
  *
  * @return True if there are any user-specified environment variables
  */
-export function hasUserEnvs(
-  functionsSource: string,
-  projectId: string,
-  projectAlias?: string
-): boolean {
+export function hasUserEnvs({
```
inlined, line 199: I don't generally like inlining types. Does it happen elsewhere in the codebase?

inlined (approved):

src/functions/env.ts
taeold: Great callout. Named this type instead.

PR #3707, opened 2021-08-25
Extract environment variables initialization out of Functions Emulator Runtime.

taeold:
As a step towards standardizing Functions Runtime environment, we extract all logic for initializing environment variables for a n emulated function out of the `FunctionsEmulatorRuntime` and place them in the caller (`FunctionsEmulator`).

All behaviors should remain the same minus the following details:

1. Environment variable values will not change once the runtime has started. Env vars like `FUNCTIONS_TARGET` might look like it should change per function invocation, but since each worker is associated with a trigger,`FUNCTION_TARGET` won't actually ever change. This may be a problem though if other emulators (e.g. storage emulator) is started _after_ the functions emulator is started and a function is invoked. (I don't think this is a common use case?).

2. We no longer check user's Functions SDK version to optionally override realtime emulator path. I think it's actually better to fail the emulator IF user has configured realtime database emulator and using an unsupported version of the functions SDK rather than silently talking to the prod database. I can be convinced otherwise though.


src/emulator/functionsEmulator.ts
```diff
@@ -818,7 +832,127 @@ export class FunctionsEmulator implements EmulatorInstance {
     return process.execPath;
   }
 
-  invokeRuntime(frb: FunctionsRuntimeBundle, opts: InvokeRuntimeOpts): RuntimeWorker {
+  getUserEnvs(): Record<string, string> {
+    const projectInfo = {
+      functionsSource: this.args.functionsDir,
+      projectId: "local", // "local" is reserved for Emulator use in functions env var.
```
abeisgoat, line 838: Can you elaborate on this?

src/emulator/functionsEmulator.ts
```diff
@@ -818,7 +832,127 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+      envs.K_SERVICE = service;
+    }
+    return envs;
+  }
+
+  getEmulatorEnvs(): Record<string, string> {
+    const envs: Record<string, string> = {};
+
+    envs.FUNCTIONS_EMULATOR = "true";
+    envs.TZ = "UTC"; // Fixes https://github.com/firebase/firebase-tools/issues/2253
```
abeisgoat, line 879: nice!

src/emulator/functionsEmulator.ts
```diff
@@ -818,7 +832,127 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+    const authEmulator = this.getEmulatorInfo(Emulators.AUTH);
+    if (authEmulator) {
+      envs[Constants.FIREBASE_AUTH_EMULATOR_HOST] = formatHost(authEmulator);
+    }
+
+    // Make firebase-admin point at the Storage emulator
+    const storageEmulator = this.getEmulatorInfo(Emulators.STORAGE);
+    if (storageEmulator) {
+      envs[Constants.FIREBASE_STORAGE_EMULATOR_HOST] = formatHost(storageEmulator);
+      envs[Constants.CLOUD_STORAGE_EMULATOR_HOST] = `http://${formatHost(storageEmulator)}`;
```
abeisgoat, line 903: We should not need both of these, they were put in for backwards compat with earlier @google-cloud/storage SDKs, we can probably just do the Firebase one as long as we require a new-ish admin SDK

src/emulator/functionsEmulator.ts
```diff
@@ -818,7 +832,127 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+    }
+
+    const pubsubEmulator = this.getEmulatorInfo(Emulators.PUBSUB);
+    if (pubsubEmulator) {
+      const pubsubHost = formatHost(pubsubEmulator);
+      process.env.PUBSUB_EMULATOR_HOST = pubsubHost;
+      logger.debug(`Set PUBSUB_EMULATOR_HOST to ${pubsubHost}`);
```
abeisgoat, line 910: Why only log for this one?

src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -649,139 +644,6 @@ function warnAboutAuthProd(frb: FunctionsRuntimeBundle): void {
...
-// This is a duplicate of the helper we use elsewhere but it's important not to
-// add dependencies to this runtime.
-function formatHost(info: { host: string; port: number }) {
-  if (info.host.includes(":")) {
-    return `[${info.host}]:${info.port}`;
-  } else {
-    return `${info.host}:${info.port}`;
-  }
-}
-
```
abeisgoat, line 784: So much red, you love to see it.

abeisgoat (commented):
Looks great! Thanks for all the hard work, sorry for being a blocker on this!

abeisgoat (approved):

src/emulator/functionsEmulator.ts
taeold: We decided to reserved `.env.local` file for setting env vars for the Functions Emulator (go/cf3-env-vars-dotenv).

Now that you point this out, I feel like this logic is better kept interrnal in the functoinsEnv module. Refactored to make finding ".env.local" a concern of the functions env var module.

src/emulator/functionsEmulator.ts
taeold: That's good to know. Since this is already a "risky" refactor, I'll lean on making cleanup changes like that in a separate PR.

src/emulator/functionsEmulator.ts
taeold: 🤷  Mostly copying & pasting code here. I'm removing it.

src/emulator/functionsEmulator.ts
```diff
@@ -818,7 +832,129 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+    targetName: string;
+    triggerType: EmulatedTriggerType;
+  }): Record<string, string> {
+    const envs: Record<string, string> = {};
+
+    // Env vars guaranteed by GCF platform.
+    //   https://cloud.google.com/functions/docs/env-var
+    envs.GCLOUD_PROJECT = this.args.projectId;
+    envs.K_REVISION = "1";
+    envs.PORT = "80";
```
inlined, line 863: This seems problematic

src/emulator/functionsEmulator.ts
```diff
@@ -818,7 +832,129 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+      envs[Constants.FIREBASE_AUTH_EMULATOR_HOST] = formatHost(authEmulator);
+    }
+
+    // Make firebase-admin point at the Storage emulator
+    const storageEmulator = this.getEmulatorInfo(Emulators.STORAGE);
+    if (storageEmulator) {
+      envs[Constants.FIREBASE_STORAGE_EMULATOR_HOST] = formatHost(storageEmulator);
+      // TODO(taeold): We only need FIREBASE_STORAGE_EMULATOR_HOST, as long as the users are using new-ish SDKs.
+      //   Clean up and update documentation in a subsequent patch.
+      envs[Constants.CLOUD_STORAGE_EMULATOR_HOST] = `http://${formatHost(storageEmulator)}`;
```
inlined, line 906: Have we left a note to clean this up in our doc for v10 CLI changes?

src/emulator/functionsEmulator.ts
```diff
@@ -860,7 +996,7 @@ export class FunctionsEmulator implements EmulatorInstance {
     }
 
     const childProcess = spawn(opts.nodeBinary, args, {
-      env: { node: opts.nodeBinary, ...opts.env, ...process.env },
+      env: { node: opts.nodeBinary, ...process.env, ...(runtimeEnv ?? {}) },
```
inlined, line 999: FYI you can `...undefined` in a map comprehension safely.

src/emulator/functionsEmulatorShared.ts
```diff
@@ -282,3 +282,11 @@ export function findModuleRoot(moduleName: string, filepath: string): string {
 
   return "";
 }
+
+export function formatHost(info: { host: string; port: number }): string {
+  if (info.host.includes(":")) {
+    return `[${info.host}]:${info.port}`;
```
inlined, line 288: This admittedly isn't my corner of the codebase, but I have no idea what this is or what this means.  Does it merit a comment?

src/functions/env.ts
```diff
@@ -179,10 +181,20 @@ function parseStrict(data: string): Record<string, string> {
...
 
-function findEnvfiles(functionsSource: string, projectId: string, projectAlias?: string): string[] {
-  const files = [".env", `.env.${projectId}`];
-  if (projectAlias && projectAlias.length) {
-    files.push(`.env.${projectAlias}`);
+function findEnvfiles(
+  functionsSource: string,
+  projectId: string,
+  projectAlias?: string,
+  isEmulator?: boolean
```
inlined, line 188: Why did we go back to `isEmulator`? Did `projectId: "local"` stop working?

src/emulator/functionsEmulator.ts
taeold: We will clean this up once we refactor the functions emulator to stop using socket for comms. Is there another problem you are worried about?

src/emulator/functionsEmulator.ts
taeold: Good point. Done!

src/emulator/functionsEmulator.ts
taeold: TIL

src/emulator/functionsEmulator.ts
inlined: code may rely on PORT. At least in Run, you're responsible for listening on $PORT. When we're actually listening on a port this variable should be accurate.

PR #3712, opened 2021-08-27
Use eventType to infer service from function trigger definition in the Emulator.

taeold:
Google Cloud Function's [EventTrigger](https://cloud.google.com/functions/docs/reference/rest/v1/projects.locations.functions#eventtrigger) message does not require `service` attribute to be set. When empty, GCF infers the service by looking at the `eventType`.

Firebase Functions SDK is moving towards leaving the `service` attribute empty when defining function triggers (v2 namespace already does this).

We update the emulator to infer service from eventType when service is missing. Note that in the future, Audit Log events may use `service` to designate the service being audited, and we may need to make additional changes here in the future to support that behavior.


taeold:
@yuchenshi @abeisgoat Ping. This PR is fairly small and straightforward and required to unlock CF3v2 PubSub support!

yuchenshi (approved):

PR #3713, opened 2021-08-27
fix: correct typo in operationType

tobiaseisenschenk:
### Description

This simply fixes a typo that caught my eye. The firebase cli said 'upsert' instead of 'update' when updating a schedule.


joehan:
This is actually intentional - this bit code of creates a scheduler job if it doesn't exist, or updates it if it already does. See https://en.wiktionary.org/wiki/upsert#:~:text=upsert%20(plural%20upserts),updates%20them%20if%20they%20do.

PR #3720, opened 2021-09-01
migrate remoteconfig over to apiv2

bkendall:
### Description

1. Replaced all the instances of the old api client with the new one (apiv2).
2. Better formatted the times printed in `remoteconfig:versions:list`

image

### Scenarios Tested

Ran through using all the `remoteconfig:*` commands. Great success.


src/remoteconfig/versionslist.ts
```diff
@@ -1,8 +1,14 @@
...
+import { remoteConfigApiOrigin } from "../api";
+import { Client } from "../apiv2";
 import { FirebaseError } from "../error";
 import { ListVersionsResult } from "./interfaces";
 import { logger } from "../logger";
 
+const apiClient = new Client({
+  urlPrefix: remoteConfigApiOrigin,
+  apiVersion: "v1",
+});
```
mbleigh, line 10: Worth putting this in `remoteconfig/client.ts` like Hosting has?

src/test/remoteconfig/rollback.spec.ts
```diff
@@ -34,90 +29,56 @@ const latestTemplate: RemoteConfigTemplate = createTemplate("115", "2020-08-06T2
...
-          auth: true,
-          origin: api.remoteConfigApiOrigin,
-          timeout: 30000,
-        }
-      );
     });
 
-    it("should reject invalid rollback version number", async () => {
-      apiRequestStub.onFirstCall().resolves({ body: latestTemplate });
+    // TODO: there is no logic that this is testing. Is that intentional?
```
mbleigh, line 48: These are just going to sit around unless someone explicitly looks at them, can you loop in an appropriate person from RC to comment?

src/remoteconfig/versionslist.ts
bkendall: I go back and forth - I only thought of making a single client recently. Might be a good thing to do, but I'll leave it be unless someone from RC says otherwise.

trekforever: I think this would be a nice to have 👍 would def be a bit more cleaner

trekforever (approved):
LGTM overall, thanks for doing this migration!

mbleigh (approved):

PR #3726, opened 2021-09-01
Add GCFv2 support for function logging

colerogers:
### Description

Adds support for obtaining the cloud function logs for customers who have GCFv2 functions. To view v2 logs, you must have `previews.functionsv2` enabled.

### Scenarios Tested

- Obtaining the correct api filter for v2 functions
- Logging the correct output from cloud run

### Sample Commands

```
# logging only v1 functions
firebase functions:log

# logging v1 & v2 functions
firebase --open-sesame functionsv2
firebase functions:log
firebase --close-sesame functionsv2
```


CHANGELOG.md
```diff
@@ -0,0 +1 @@
+`functions:log` now supports v2 functions.
```
inlined, line 1: These go out in public release notes. Make this vague or omit it from the relntoes entirely.

src/gcp/cloudlogging.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+}
+
+/**
+ * The correct API filter to use when GCFv2 is enabled and/or we want specific function logs
+ * @param v2Enabled check if the user has the preview v2 enabled
+ * @param functionList list of functions seperated by comma
+ * @returns the correct filter for use when calling the list api
+ */
+export function getApiFilter(v2Enabled: boolean, functionList?: string) {
```
inlined, line 62: You've put GCF-specific code in a cloudlogging library. Let's move this back to the functions logging code.

src/gcp/cloudlogging.ts
```diff
@@ -0,0 +1,101 @@
...
+ */
+export function logEntries(entries: LogEntry[]): void {
+  if (!entries || entries.length === 0) {
+    logger.info("No log entries found.");
+    return;
+  }
+  for (let i = entries.length - 1; i >= 0; i--) {
+    const entry = entries[i];
+
+    logger.info(
```
inlined, line 91: This would probably read a lot cleaner if you broke it up into components:

```
const timestamp = entry.timestamp || "---";
const severity = (entry.severity || "?").substring(0, 1);
const name = entry.resource.labels.function_name || entry.resoruce.labels.service_name;
const message = entry.textPayload || JSON.stringify(entry.jsonPayload) || JSON.stringify(entry.protoPayload) || "";

logger.info(`${timestamp} ${severity} ${name}: ${message}`);
```

inlined (changes requested):
I had no idea it was this straightforward. That's an easy win!

colerogers:
> I had no idea it was this straightforward. That's an easy win!

Credit goes completely to @taeold, I just refactored some things.

src/gcp/cloudlogging.ts
colerogers: Makes sense to remove from the gcf library, instead of putting the functions back into the `Command` where we can't export them, I think it we should move them into a file like `src/functions/functionsLog.ts`, thoughts?

inlined: SGTM

CHANGELOG.md
colerogers: removed

src/gcp/cloudlogging.ts
colerogers: done

src/gcp/cloudlogging.ts
colerogers: changed

src/test/functions/functionsLog.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    it("should return list api filter for v1 functions", () => {
+      expect(functionsLog.getApiFilter(false, "fn1,fn2")).to.eq(
+        'resource.type="cloud_function"\n(resource.labels.function_name="fn1" OR resource.labels.function_name="fn2")'
+      );
+    });
+
+    it("should return list api filter for v1&v2 functions", () => {
+      expect(functionsLog.getApiFilter(true, "fn1,fn2")).to.eq(
+        'resource.type="cloud_function" OR (resource.type="cloud_run_revision" AND labels."goog-managed-by"="cloudfunctions")\n(resource.labels.function_name="fn1" OR resource.labels.service_name="fn1" OR resource.labels.function_name="fn2" OR resource.labels.service_name="fn2")'
```
taeold, line 26: nit* Would love to see us break this into multiline string 

e.g.

```
'resource.type="cloud_function" OR +
  '(resource.type="..)
...
```

taeold (approved):

src/test/functions/functionsLog.spec.ts
colerogers: broke out into a multiline string

src/functions/functionslog.ts
```diff
@@ -0,0 +1,51 @@
+import { logger } from "../logger";
+import { LogEntry } from "../gcp/cloudlogging";
+
+/**
+ * The correct API filter to use when GCFv2 is enabled and/or we want specific function logs
+ * @param v2Enabled check if the user has the preview v2 enabled
+ * @param functionList list of functions seperated by comma
+ * @returns the correct filter for use when calling the list api
+ */
+export function getApiFilter(v2Enabled: boolean, functionList?: string) {
```
inlined, line 10: nit: my personal preference would be for the API signature to exist as if there was no preview flag and the code itself to work around the preview (e.g. so `getApiFilter(functionList?: boolean)` which then does the right thing depending on `previews.functionsv2`. That helps keep concerns stay as locally isolated as possible.

inlined (approved):

src/functions/functionslog.ts
colerogers: Makes sense, fixed in the latest commit

PR #3727, opened 2021-09-01
Functions List Command

colerogers:
### Description

A command for listing all functions in a customers Firebase project.

### Scenarios Tested

- No functions in project
- Listing sorted functions

### Sample Commands

```
firebase functions:list
```


src/test/functions/listFunctions.spec.ts
```diff
@@ -0,0 +1,160 @@
+import { expect } from "chai";
+import * as sinon from "sinon";
+import { listFunctions } from "../../functions/listFunctions";
+import * as backend from "../../deploy/functions/backend";
+import * as args from "../../deploy/functions/args";
+import { previews } from "../../previews";
+
+describe("listFunctions", () => {
```
taeold, line 8: Is nesting `describe` under another `describe` intentional here?

src/test/functions/listFunctions.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+      backendStub = sandbox.stub(backend, "existingBackend");
+      previews.functionsv2 = false;
+    });
+
+    afterEach(() => {
+      sandbox.restore();
+    });
+
+    it("should return an empty array on empty function spec list", async () => {
+      backendStub.returns(Promise.resolve({ cloudFunctions: [] }));
```
taeold, line 24: nit* this is equivalent to:

```
backendStub.resolves({ cloudFunctions: [] });
```

src/functions/listFunctions.ts
```diff
@@ -0,0 +1,60 @@
...
+      table.push([
+        fnSpec.entryPoint,
+        trigger,
+        fnSpec.region,
+        fnSpec.availableMemoryMb || "---",
+        fnSpec.runtime,
+      ]);
+    }
+  }
+  logger.info(table.toString());
```
taeold, line 58: I feel like this entire logic could live in the `command/functions-list.ts` instead of as a its own module.

I think we are trying to keep the logic under `command/functions-list.ts` as light as possible, but most of the logic here seems to be about cosmetics of the output, which (I think) should live in the command side since comestic codes are less shareable.

At the same time, if all of the cosmetic code is pull out of here, then we are basically left with `backend.existingBackend` - maybe this function can just be responsible for returning the right set of functions given user's preview flag?

@bkendall recently shared a great thought on Firebase CLI chat room what he thinks should and should not happen in the command code - might want to ping him for opinion!

src/test/functions/listFunctions.spec.ts
colerogers: Yeah for the filename and function. I guess it's kind of confusing since they're both the same name.

src/functions/listFunctions.ts
colerogers: I struggled a bit on where to put this logic too. I read and re-read @bkendall's comment and having all the logic in this function does make the `Command` very flat. But on the other hand, if the `Command`'s job is to format output for the user, then we should move the `Table` stuff there and just leave the API call here (which kind of makes this function useless if it's just calling an existing api & sorting the results). Also, the pattern that I used here already exists in the project - `ext:list` uses this technique pretty much exactly. 

I'm curious to see what others think. CC @inlined @bkendall

src/functions/listFunctions.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @returns mapping that contains a list of functions under the 'functions' key
+ */
+export async function listFunctions(
+  context: args.Context
+): Promise<{ functions: backend.FunctionSpec[] }> {
+  const bkend = await backend.existingBackend(context, true);
+  const functionSpecs = previews.functionsv2
+    ? bkend.cloudFunctions
+    : bkend.cloudFunctions.filter((fn) => fn.platform === "gcfv1");
+  functionSpecs.sort((fn1, fn2) => {
```
inlined, line 19: Let's improve consistency by promoting [this method](https://github.com/firebase/firebase-tools/blob/master/src/deploy/functions/prompts.ts#L16) to something in backend.ts and use it here.

Totally unrelated to this change, I'd also like to consider changing the sorting method to sort by region tier before region name.

src/functions/listFunctions.ts
```diff
@@ -0,0 +1,60 @@
...
+    ? new Table({
+        head: ["Function", "Version", "Trigger", "Location", "Memory", "Runtime"],
+        style: { head: ["yellow"] },
+      })
+    : new Table({
+        head: ["Function", "Trigger", "Location", "Memory", "Runtime"],
+        style: { head: ["yellow"] },
+      });
+  for (const fnSpec of functionSpecs) {
+    const trigger = backend.isEventTrigger(fnSpec.trigger) ? fnSpec.trigger.eventType : "https";
```
inlined, line 38: I wonder if we want something more clever than just this. E.g. should we call a scheduled function "scheduled" instead of "google.pubsub.topic.publish" and should we call a callable function "callable" instead of "https".

WDYT @taeold?

src/functions/listFunctions.ts
inlined: While I would not have made the same decision as @bkendall, I want to respect his experience and authority. If I can paraphrase everyone's concerns, this seems like an MVC codebase. We don't mind mixing models and controllers (e.g. most of backend.ts and existingBackend()), but we're not sure views belong in the same place. I would love some thoughts on how to cleanly separate the view codebases from the model codebases, but this is a fine first start.

inlined: OTOH, unit tests are only verifying that the sort function works. If that's all we're going to test, then we can put the sort function in backend.ts per my other comment and move this untested code to the command.

src/functions/listFunctions.ts
taeold: that sounds like a good ux improvement, but I'd be happy with shipping what we have here and tackling this as a follow up work to pick up later if its going to take more than few hours.

src/test/functions/listFunctions.spec.ts
colerogers: I removed this set of unit tests and added test cases to compareFunction

src/test/functions/listFunctions.spec.ts
colerogers: same as above

src/functions/listFunctions.ts
colerogers: I exported the compareFunctions function and used it in backend.ts where we list all the functions and I removed this file.

src/functions/listFunctions.ts
colerogers: We can create a mapping for trigger types from full name to shorthand, like "google.pubsub.topic.publish" -> "scheduled". Is there an internal doc somewhere that has similar shorthand mappings like this?

Also, since this is a user facing change, would this need to go back through council?

src/functions/listFunctions.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Lists all functions of the Firebase project in order
+ * @param context the Context of the project
+ * @returns a mapping that contains an array of {@link FunctionSpec} in order under the 'functions' key
+ */
+export async function listFunctions(
+  context: Context
+): Promise<{ functions: backend.FunctionSpec[] }> {
+  const bkend = await backend.existingBackend(context, true);
```
bkendall, line 15: 🤨

src/functions/listFunctions.ts
bkendall: (I like your MVC codebase comparison. The idea that I usually base my thoughts off of are "functional core, imperative shell". Short version: the less the `Command` has to do with the actual core of getting the data and making decisions, the better. `Command`s shouldn't know how to make API calls, and the core (like `listFunctions` shouldn't know how to write tables to stdout.)

CHANGELOG.md
```diff
@@ -1,2 +1,3 @@
 - `ext:install` now supports `--force` and `--non-interactive` flags.
 - Fixes a crash when customers deploy an empty Cloud Functions project (#3705)
+- Adds command for listing all functions in the Firebase project `functions:list`.
```
bkendall, line 3:
```suggestion
- Adds a command (`functions:list`) for listing all functions in the Firebase project.
```

Also, is it "all Google Cloud Functions" or "all Cloud Functions for Firebase" in the project?

bkendall (approved):
This looks good to me. Maybe get a 👍  from another CF3 folk (and take a look at the changelog question), but looks ready to go

src/functions/listFunctions.ts
colerogers: Sounds good, I'll make a feature request for creating this mapping and assign myself

CHANGELOG.md
colerogers: We'd expect this to usually be all Cloud Functions for Firebase since the user is running this command in their Firebase project. However, under the hood this command calls the GCP REST API for obtaining all Google Cloud Functions. So if a user creates a Google Cloud Function in the same project as they deploy Firebase Functions, then this command will display that function.

taeold: FWIW - Firebase Console also displays all GCFs on the user project; it doesn't try to filter anything.

src/functions/listFunctions.ts
```diff
@@ -0,0 +1,21 @@
...
+  const functionSpecs = previews.functionsv2
```
taeold, line 16: nit* `backend.existingBackend` already includes logic for filtering out gcfv2 functions, so this is a redundant check.

taeold (approved):
minor nits, otherwise lgtm.

src/functions/listFunctions.ts
colerogers: Removed

src/commands/functions-list.ts
```diff
@@ -0,0 +1,51 @@
...
+              availableMemoryMb,
+              fnSpec.runtime,
+            ]
+          : [fnSpec.entryPoint, trigger, fnSpec.region, availableMemoryMb, fnSpec.runtime];
+        table.push(entry);
+      }
+      logger.info(table.toString());
+      return functionList;
+    } catch (err) {
+      throw new FirebaseError(`Failed to list functions ${err.message}`, { exit: 1 });
```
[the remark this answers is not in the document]
colerogers: Fixed

PR #3730, opened 2021-09-02
Fix adding testers from a file

lfkellogg:
Figured I'd try and fix this since it was probably something simple. Confirmed that this fixes it:

```
$ firebase appdistribution:testers:add --file testers.txt --project ...
i  Adding 3 testers to project
✔  Testers created successfully
```

This was the error: http://go/paste/6521938970148864


kaibolay (approved):

J-coder13 (commented):
Ahh thanks for catching that!

PR #3731, opened 2021-09-02
Create new endpoint type definition

inlined:
Creating the new types to use to match up with Extensions and the recent design reviews for future products. The main difference here is that Trigger is a oneof type (Triggered) and that these types can embed things like a Schedule in them. This will allow us to have complex "trigger" types that are both a Cloud Functions trigger as well as describing the resoruces that feed into those triggers (or that omit features which shouldn't be available in those triggers. For example, invoker cannot be set on the HTTPS functions that will serve blocking auth hooks)


src/deploy/functions/backend.ts
```diff
@@ -196,6 +212,7 @@ export interface Backend {
   schedules: ScheduleSpec[];
   topics: PubSubSpec[];
   environmentVariables: EnvironmentVariables;
+  endpoints: Endpoint[];
```
mbleigh, line 215: Should this be an array or a `Record<string,Endpoint>` (as per stack spec discussion)?

mbleigh (approved):
Seems reasonable, but please take a look at the test failures.

src/deploy/functions/backend.ts
inlined: Hmmm... would the key be `${project}/${region}/${id}`? The stack spec discussions inadvertently presumed that the ID is a unique key, which it isn't.

mbleigh: Hmm yeah my assumption is that Spec would be both project and region locked. In this case would there ever be multiple projects at the same time? Location maybe, but not project...

PR #3733, opened 2021-09-03
Remove HttpsTrigger.allowInsecure

inlined:
Removing the start of a feature we'll probably never use.

Cloud Run/GCFv2 always redirect HTTP traffic to HTTPS. This year, GCFv1 released a feature where customers could opt into the same behavior, so the FunctionSpec had an "allowInsecure" option that the SDK could trigger to tell the CLI to opt into `SECURE_ALWAYS`.

It's unlikely that we'll actually invest in this feature for GCFv1. It's probably better to spend more PR energy just getting people on to v2, so we can remove this wart that will have to be remembered in our future products long after it can ever be made `true`.


colerogers (approved):
LGTM

PR #3735, opened 2021-09-04
Add converters between API types and Endpoint types

inlined:
Add alternate converters functionToEndpoint and endpointToFunction based on functionToSpec and specToFunction. I highly recommend viewing this by commit. The first commit is a mechanical copy + paste + function name change and the second commit is actually interesting.


src/gcp/cloudfunctions.ts
```diff
@@ -579,44 +579,47 @@ export function functionFromSpec(
...
+  let trigger: backend.Triggered;
   let uri: string | undefined;
   if (gcfFunction.httpsTrigger) {
+    trigger = { httpsTrigger: {} };
+    uri = gcfFunction.httpsTrigger.url;
+  } else if (gcfFunction.labels?.["deployment-scheduled"]) {
     trigger = {
-      // Note: default (empty) value intentionally means true
-      allowInsecure: gcfFunction.httpsTrigger.securityLevel !== "SECURE_ALWAYS",
+      scheduleTrigger: {},
```
taeold, line 591: q: why do we not copy over attributes like `schedule`, `timeZone`, and `retryConfig` here?

src/gcp/cloudfunctionsv2.ts
```diff
@@ -510,86 +511,80 @@ export function functionFromEndpoint(cloudFunction: backend.FunctionSpec, source
...
       for (const { attribute, value } of gcfFunction.eventTrigger.eventFilters || []) {
-        trigger.eventFilters[attribute] = value;
+        trigger.eventTrigger.eventFilters[attribute] = value;
       }
     }
   } else {
-    trigger = {
-      allowInsecure: false,
-    };
+    trigger = { httpsTrigger: {} };
```
taeold, line 569: q: what about `uri`?

tangent: does GCFv2 expose url for all functions types (event, http) like Cloud Run?

taeold (approved):

src/gcp/cloudfunctions.ts
inlined: We would have to make a second API request to Cloud Scheduler to fill in these details and historically we have not done this.

src/gcp/cloudfunctionsv2.ts
inlined: URI is an attribute of Function, not httpsTrigger, and is set a few lines below (see line 584). It is present for all trigger types.

PR #3744, opened 2021-09-09
Fix X-Client-Version header

kaibolay:
Provide actual version number instead of `undefined`.


lfkellogg (approved):
Thanks! Verified this fix locally.

Before: http://screen/AjSM7tA6KmKXfPP
After: http://screen/5hiDzs2b7BZUdnt (I changed the version number locally)

PR #3745, opened 2021-09-09
Adding noninteractive support to ext:dev:publish

joehan:
### Description
ext:dev:publish now works correctly with the --force and --non-interactive flags

### Scenarios Tested
No flags:
Screen Shot 2021-09-09 at 10 21 43 AM

--force:
Screen Shot 2021-09-09 at 10 35 56 AM

--non-interactive:
Screen Shot 2021-09-09 at 10 36 32 AM

--force, --non-interactive
Screen Shot 2021-09-09 at 10 23 25 AM


src/extensions/paramHelper.ts
```diff
@@ -56,14 +56,15 @@ export function getParamsWithCurrentValuesAsDefaults(
...
-  if (noninteractive && !envFilePath) {
-    const paramsMessage = paramSpecs
+export async function getParams(args: {
+  projectId: string;
+  paramSpecs: extensionsApi.Param[];
+  nonInteractive?: boolean;
+  paramsEnvPath?: string;
+}): Promise<{ [key: string]: string }> {
+  let params: any;
+  if (args.nonInteractive && !args.paramsEnvPath) {
```
huangjeff5, line 66: nit: Should we handle the case / error out where its !args.nonInteractive and paramsEnvPath is provided?

huangjeff5 (approved):

src/extensions/paramHelper.ts
joehan: We do, in the `else if (args.paramsEnvPath) {` block.  Whether we're in interactive or noninteractive mode, we should read params from that file if provided.

PR #3732, opened 2021-09-03
Add MFA support for Auth Emulator

yuchenshi:
### Description

This adds partial MFA support for password sign-in etc. Error messages are NOT yet done (strings in double braces may not match production) and signInWithIdp + MFA doesn't work right now.

Many tests for edge cases (especially error strings) TBD.

### Scenarios Tested

See tests added

PR #3746, opened 2021-09-09
Refactor in preparation for adding multitenancy support

lisajian:
### Description

This change introduces the refactor discussed in [go/emulator-multi-tenancy](http://go/emulator-multi-tenancy). Main changes include:

- Updating how `ProjectState` is indexed in `getProjectStateById()`
- Extending `ProjectState` with `AgentProjectState` and `TenantProjectState` 

### Scenarios Tested

`npm test` passes

### Sample Commands

N/A


src/emulator/auth/apiSpec.js
```diff
@@ -2981,6 +2981,52 @@ export default {
   },
   components: {
     schemas: {
+      GoogleCloudIdentitytoolkitV1Argon2Parameters: {
```
lisajian, line 2984: I see that #3732 is also updating apiSpec.js and schema.ts, and it seems like the changes are the same across the PRs. LMK if you would prefer I leave this out so it's easier to merge @yuchenshi

src/emulator/auth/state.ts
```diff
@@ -550,6 +550,45 @@ export class ProjectState {
...
+  // TODO(lisajian): Fill in when v2.projects.tenants.delete is added
+  deleteTenant(): void {
+    return;
+  }
+}
+
+export class TenantProjectState extends ProjectState {
+  readonly tenantId: string;
+
+  constructor(projectId: string, tenantId: string) {
```
sam-gc, line 586: This can be simplified to:

```ts
export class TenantProjectState extends ProjectState {
  constructor(projectId: string, readonly tenantId: string) {
    super(projectId);
  }
}
```

(This makes use of the parameter properties in typescript constructors: https://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties)

sam-gc (commented):
LGTM, Approval pending resolution regarding the schema files

src/emulator/auth/server.ts
```diff
@@ -461,10 +476,8 @@ function toExegesisController(
         // See: https://cloud.google.com/identity-platform/docs/reference/rest/v1/accounts/signUp
         targetProjectId = ctx.user;
       }
-      if (ctx.params.path.tenantId || ctx.requestBody?.tenantId) {
-        throw new NotImplementedError("Multi-tenancy is unimplemented.");
-      }
-      return operation(getProjectStateById(targetProjectId), ctx.requestBody, ctx);
+      const targetTenantId: string = ctx.params.path.tenantId || ctx.requestBody?.tenantId;
```
yuchenshi, line 479: please check if both path and body have tenantId then they must match

src/emulator/auth/state.ts
```diff
@@ -550,6 +550,45 @@ export class ProjectState {
...
+    return;
+  }
+
+  // TODO(lisajian): Fill in when v2.projects.tenants.delete is added
+  deleteTenant(): void {
+    return;
+  }
+}
+
+export class TenantProjectState extends ProjectState {
```
yuchenshi, line 583: A few things should have only one single instance that applies to both the agent + all tenants -- e.g. `oneAccountPerEmail`, `authCloudFunction` -- how would you like to handle this?

e.g. when `oneAccountPerEmail` is updated, it should affect the behavior of the agent project and all tenants under.

src/emulator/auth/state.ts
```diff
@@ -550,6 +550,45 @@ export class ProjectState {
     }
   }
 }
+
+export class AgentProjectState extends ProjectState {
+  private tenantForTenantId: Map<string, Tenant> = new Map();
+
+  // TODO(lisajian): Fill in when v2.projects.tenants.get is added
```
yuchenshi, line 557: These should all throw `NotImplemented`

src/emulator/auth/apiSpec.js
yuchenshi: Please leave those out

yuchenshi (commented):
LGTM with comments inline

src/emulator/auth/apiSpec.js
lisajian: Cool thanks, reverted!

src/emulator/auth/state.ts
lisajian: Ah, thanks for pointing that out - I've updated `state.ts` such that the top level project settings are only set in the `AgentProjectState`s and `TenantProjectState`s simply refer to their parent state's settings. Thanks for your patience while I was debugging a gnarly [shadowing bug](https://stackoverflow.com/questions/53584705/javascript-extending-es6-class-setter-will-inheriting-getter#:~:text=the%20instance%20x%20will%20not,of%20the%20setter%2C%20as%20undefined.). LMK what you think with the new revision

src/emulator/auth/server.ts
```diff
@@ -461,10 +476,14 @@ function toExegesisController(
         // See: https://cloud.google.com/identity-platform/docs/reference/rest/v1/accounts/signUp
         targetProjectId = ctx.user;
       }
-      if (ctx.params.path.tenantId || ctx.requestBody?.tenantId) {
-        throw new NotImplementedError("Multi-tenancy is unimplemented.");
+      if (ctx.params.path.tenantId && ctx.requestBody?.tenantId) {
+        assert(
+          ctx.params.path.tenantId === ctx.requestBody?.tenantId,
```
yuchenshi, line 481: nit:
```suggestion
          ctx.params.path.tenantId === ctx.requestBody.tenantId,
```

src/emulator/auth/server.ts
```diff
@@ -461,10 +476,14 @@ function toExegesisController(
...
+          "Tenant IDs in path and request body do not match"
```
yuchenshi, line 482: Needs a test (and also I believe production has a different error code for this)

src/emulator/auth/state.ts
```diff
@@ -29,20 +29,30 @@ export class ProjectState {
   private oobs: Map<string, OobRecord> = new Map();
   private verificationCodes: Map<string, PhoneVerificationRecord> = new Map();
   private temporaryProofs: Map<string, TemporaryProofRecord> = new Map();
-  public oneAccountPerEmail = true;
-  private authCloudFunction: AuthCloudFunction;
-  public usageMode: UsageMode = UsageMode.DEFAULT;
+  protected _oneAccountPerEmail = true;
+  protected _authCloudFunction: AuthCloudFunction | undefined;
```
yuchenshi, line 33: nit: This should be `protected readonly` and should be passed in through a constructor argument. (Feel free to use the shorthand notation in constructor)

src/emulator/auth/state.ts
```diff
@@ -74,7 +84,7 @@ export class ProjectState {
     const user = this.updateUserByLocalId(localId, props, {
       upsertProviders: props.providerUserInfo,
     });
-    this.authCloudFunction.dispatch("create", user);
+    this._authCloudFunction?.dispatch("create", user);
```
yuchenshi, line 87: Should triggers work for tenant projects? Not sure about this, but my guess is yes. (When in doubt try it out in production)

src/emulator/auth/state.ts
```diff
@@ -29,20 +29,30 @@ export class ProjectState {
   private oobs: Map<string, OobRecord> = new Map();
   private verificationCodes: Map<string, PhoneVerificationRecord> = new Map();
   private temporaryProofs: Map<string, TemporaryProofRecord> = new Map();
-  public oneAccountPerEmail = true;
-  private authCloudFunction: AuthCloudFunction;
-  public usageMode: UsageMode = UsageMode.DEFAULT;
+  protected _oneAccountPerEmail = true;
```
yuchenshi, line 32: This should be abstract getter only. Underlying storage should be moved into `AgentProjectState` (since only those instances hold the value)

src/test/emulators/auth/misc.spec.ts
```diff
@@ -365,6 +365,9 @@ describeAuthEmulator("accounts:query", ({ authApi }) => {
 });
 
 describeAuthEmulator("emulator utility APIs", ({ authApi }) => {
+  // TODO(lisajian): Add test that errors when changing allowDuplicateEmails config on tenant project
```
yuchenshi, line 368: Is this test possible at all? How would one set `tenantId` in the request if it is neither in the path or the request body?

yuchenshi (approved):

src/emulator/auth/state.ts
```diff
@@ -29,20 +29,21 @@ export class ProjectState {
...
 
   get projectNumber(): string {
     // TODO: Shall we generate something different for each project?
     // Hard-coding an obviously fake number for clarity for now.
     return "12345";
   }
 
+  abstract get oneAccountPerEmail(): boolean;
+
+  abstract get authCloudFunction(): AuthCloudFunction | undefined;
```
yuchenshi, line 43: This shouldn't be nullable

src/emulator/auth/state.ts
```diff
@@ -74,7 +75,7 @@ export class ProjectState {
     const user = this.updateUserByLocalId(localId, props, {
       upsertProviders: props.providerUserInfo,
     });
-    this.authCloudFunction.dispatch("create", user);
+    this.authCloudFunction?.dispatch("create", user);
```
yuchenshi, line 78: revert these

src/emulator/auth/state.ts
```diff
@@ -551,6 +552,87 @@ export class ProjectState {
     }
   }
 }
+
+export class AgentProjectState extends ProjectState {
+  private tenantForTenantId: Map<string, Tenant> = new Map();
+
+  constructor(
+    projectId: string,
+    private _oneAccountPerEmail = true,
```
yuchenshi, line 561: Those things don't make sense in constructor since they are never injected in. Consider making them simply just properties (like `tenantForTenantId`)

yuchenshi (approved):

src/emulator/auth/state.ts
lisajian: Ah my bad, removed the `| undefined`

src/emulator/auth/state.ts
lisajian: Done

src/emulator/auth/state.ts
lisajian: Done

PR #3753, opened 2021-09-13
Start moving over cases of existingBackend

inlined:
Starting the migration process from FunctionSpec to Endpoint

In the process, I removed listFunctions.ts since it had a cylclomatic complexity of 0. Its unit tests were therefore redundant with other tests and it could be inlined with functions-list.


src/deploy/functions/backend.ts
```diff
@@ -537,10 +560,31 @@ export async function checkAvailability(context: Context, want: Backend): Promis
   }
 }
 
-// To be a bit more deterministic, print function lists in a prescribed order.
-// Future versions might want to compare regions by GCF/Run pricing tier before
-// location.
-export function compareFunctions(left: FunctionSpec, right: FunctionSpec): number {
+/** A helper utility for flattening all endpoints in a backend since typing is a bit wonky. */
+export function allEndpoints(backend: Backend): Endpoint[] {
```
joehan, line 564: Consider adding some light unit tests for these new methods. I wouldn't normally suggest doing so for helper methods, but these seem like utility methods that are used in a few places (and could easily become more widely used over time).

joehan (approved):

taeold (approved):

PR #3756, opened 2021-09-14
Add Endpoints version of DeploymentPlanner

inlined:
This is generally a rewrite of the deploymentPlanner code. The code looks vaguely the same, except there's a lot less wonky stuff to do now that Endpoints are already stored by region and now that we have Endpoints helper methods in backend.ts

I've decided that the release code is complex enough that I'll build a new version in pieces in a `release` subdirectory. I think I might actually leave the code there though since there's a lot of things that are generically named and don't obviously scream "this is for the release phase".


src/deploy/functions/release/planner.ts
```diff
@@ -0,0 +1,236 @@
...
+ * Throws an error when upgrading/downgrading GCF versions.
+ * This is a separate function that is designed to be stubbed in tests to allow
+ * upgrading to v2 in tests before production is ready
+ */
+export function checkForV2Upgrade(want: backend.Endpoint, have: backend.Endpoint): void {
+  if (want.platform == "gcfv2" && have.platform == "gcfv1") {
+    throw new FirebaseError(
+      `[${getFunctionLabel(
+        have
+      )}] Upgrading from GCFv1 to GCFv2 is not yet supported. Please delete your old function or wait for this feature to be ready.`
```
taeold, line 233: I'm wondering if we should use `CF3v{1/2}` instead of `GCFv{1,2}`? Is `CF3v2 === GCFv2` always going to hold or could `CF3v2` target platform other than `GCFv2`?

taeold (approved):

joehan (approved):

src/test/deploy/functions/release/planner.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+            eventType: gcfv2.PUBSUB_PUBLISH_EVENT,
+            eventFilters: {
+              resource: "topic",
+            },
+            retry: false,
+          },
+        }),
+        platform: "gcfv2",
+      };
+      const changed = JSON.parse(JSON.stringify(original)) as backend.Endpoint;
```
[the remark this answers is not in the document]
inlined: 😬  ...that support matrix...

src/deploy/functions/release/planner.ts
inlined: CF3v2 is an SDK, GCFv2 is a platform. It's possible we would have a different `platform` from the same SDK, just like we're going to support `platform: "gcfv1"` in CF3v1 (even if begrudgingly). So I think GCFv1/2 is the right label here.

PR #3758, opened 2021-09-15
Add Crashlytics Symbol Upload Command for Android NDK

samedson:


samedson:
The `'SymbolGenerator' is defined but never used` warnings seem to be false positives? Reference to similar issue: https://stackoverflow.com/questions/58981482/false-positives-for-no-unused-vars

src/commands/crashlytics-symbols-upload.ts
```diff
@@ -0,0 +1,195 @@
...
+const JAR_VERSION = "2.7.1";
+const JAR_URL = `https://dl.google.com/android/maven2/com/google/firebase/firebase-crashlytics-buildtools/${JAR_VERSION}/firebase-crashlytics-buildtools-${JAR_VERSION}.jar`;
+
+export default new Command("crashlytics:symbols:upload <symbol-files...>")
+  .description("Upload symbols for native code, to symbolicate stack traces.")
+  .option("--app <app_id>", "the app id of your Firebase app")
+  .option("--generator [breakpad|csym]", "the symbol generator being used, defaults to breakpad.")
+  .option("--dry-run", "generate symbols without uploading them")
+  .option("--debug", "print debug output and logging from the underlying uploader tool")
+  .option("--local-jar <path>", "override the fetched jar with one on the filesystem")
```
samedson, line 49: Is there a way to make hidden flags? It didn't seem to accept the option unless I added this

src/utils.ts
```diff
@@ -191,6 +192,29 @@ export function logLabeledWarning(
...
+ * Log an info statement with a gray bullet at the start of the line.
+ */
+export function logLabeledDebug(
+  label: string,
+  message: string,
+  type: LogLevel = "debug",
+  data: LogDataOrUndefined = undefined
+): void {
+  logger[type](clc.blue.bold(`${DEBUG_CHAR}  ${label}:`), message, data);
+}
```
bkendall, line 216: What's the utility of these two functions? AFAIK, what will happen here is that these will make it into the debug log (output only if `--debug` is provided or only into `firebase-debug.log`). That's the same as just using `logger.debug` in the end. I don't see the value add here...

src/commands/crashlytics-symbols-upload.ts
samedson: Maybe I can use an environment variable instead

src/utils.ts
samedson: Makes sense - I didn't see the `logger.debug` 👍

bkendall (approved):
one thing, but LGTM

PR #3761, opened 2021-09-17
Update gen-auth-api-spec.ts to use flatPath instead of path

lisajian:
### Description

Add preprocessing logic to use `flatPath`s instead of `path`s in discovery object. 

Some endpoints/methods have the same `path`, which led to endpoints being overwritten and not appearing in `apiSpec.js` (e.g. see `identitytoolkit.projects.inboundSamlConfigs.delete` and `identitytoolkit.projects.defaultSupportedIdpConfigs.delete` in [v2 discovery docs](https://identitytoolkit.googleapis.com/$discovery/rest?version=v2)). To work around this, use the `flatPath` which are distinct between overloaded paths. This change largely impacts paths that have params with slashes in them, which is [currently not supported by OpenAPI](https://github.com/OAI/OpenAPI-Specification/issues/892).

Corresponding internal bug: b/199768026

### Scenarios Tested

Running the following sequentially passes: 

`npm run generate:auth-api`
`npm test`

### Sample Commands

N/A

### DIFF
--- a/scripts/gen-auth-api-spec.ts
+++ b/scripts/gen-auth-api-spec.ts
@@ -28,11 +28,6 @@ import swaggerToTS from "@manifoldco/swagger-to-ts";
 // @ts-ignore
 import * as googleDiscoveryToSwagger from "google-discovery-to-swagger";
 
-// Swagger and OpenAPI 3.0 formats do not support RFC 6570 Level 2 features
-// (like {+var}) which are present in API Discovery. Setting this flag in this
-// library makes it produce Level 1-only URL templates and fully spec-compliant.
-googleDiscoveryToSwagger.setStrict(true);
-
 async function main(): Promise<void> {
   const [v1Disc, v2Disc, tokenDisc] = await Promise.all([
     fetchJson("https://identitytoolkit.googleapis.com/$discovery/rest?version=v1"),
@@ -113,6 +108,9 @@ async function toOpenapi3(discovery: any): Promise<any> {
   const apiKeyDescription = discovery.parameters.key.description;
   delete discovery.parameters.key;
 
+  // Preprocess and replace paths with flatPaths
+  replaceWithFlatPath(discovery.resources);
+
   // We first convert the discovery document to Swagger (a.k.a. OpenAPI 2.0) and
   // then to OpenAPI 3.0 because there is tool that does direct conversion. Some
   // tools offer one single API call for the entire conversion, but perform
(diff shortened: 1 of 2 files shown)


scripts/gen-auth-api-spec.ts
```diff
@@ -130,6 +128,70 @@ async function toOpenapi3(discovery: any): Promise<any> {
   return openapi3;
 }
 
+const pathParamsForFlatPathParam: Map<string, string> = new Map([
```
sam-gc, line 131: I *think* the type can be inferred here (if it can, remove the explicit typing), otherwise no worries

go/tsstyle#type-inference

scripts/gen-auth-api-spec.ts
```diff
@@ -130,6 +128,70 @@ async function toOpenapi3(discovery: any): Promise<any> {
...
+  ["{projectsId}", "{targetProjectId}"],
+  ["{tenantsId}", "{tenantId}"],
+]);
+
+const paramPattern = /{([^}]+)}/g;
+
+function replaceWithFlatPath(discovery: { [key: string]: any }): void {
```
sam-gc, line 138: Is it possible to type the discovery more narrowly than `any`? (I'm not super familiar with this code)

scripts/gen-auth-api-spec.ts
```diff
@@ -130,6 +128,70 @@ async function toOpenapi3(discovery: any): Promise<any> {
...
+    Object.entries(discovery.methods).forEach(([_, method]) => {
+      if (typeof method === "object") {
+        // Replace flat path param names with path param names
+        // e.g. for endpoint identitytoolkit.projects.defaultSupportedIdpConfigs.get:
+        // path = v2/{name}
+        // flatPath = v2/projects/{projectsId}/defaultSupportedIdpConfigs/{defaultSupportedIdpConfigsId}
+        //
+        // transform v2/projects/{projectsId}/defaultSupportedIdpConfigs/{defaultSupportedIdpConfigsId}
+        //       --> v2/projects/{targetProjectId}/defaultSupportedIdpConfigs/{defaultSupportedIdpConfigsId}
+        let flatPath = (method as any).flatPath;
```
sam-gc, line 149: Avoid `any` by casting it as:
`let flatPath = (method as {flatPath: string}).flatPath`

scripts/gen-auth-api-spec.ts
```diff
@@ -130,6 +128,70 @@ async function toOpenapi3(discovery: any): Promise<any> {
...
+        const cleanedParams: { [key: string]: any } = {};
+
+        // Get all param names in path
+        // e.g. ["projectsId", "defaultSupportedIdpConfigsId"]
+        const pathParamMatches = [...flatPath.matchAll(paramPattern)];
+        const paramsInPath = pathParamMatches.map((match) => match[1]);
+
+        // Remove method path parameters that don't appear in the path
+        // e.g. remove parameter "name" that appears in original path
+        const params = (method as any).parameters;
```
sam-gc, line 163: Same here re casting

scripts/gen-auth-api-spec.ts
```diff
@@ -130,6 +128,70 @@ async function toOpenapi3(discovery: any): Promise<any> {
...
+        // flatPath = v2/projects/{projectsId}/defaultSupportedIdpConfigs/{defaultSupportedIdpConfigsId}
+        //
+        // transform v2/projects/{projectsId}/defaultSupportedIdpConfigs/{defaultSupportedIdpConfigsId}
+        //       --> v2/projects/{targetProjectId}/defaultSupportedIdpConfigs/{defaultSupportedIdpConfigsId}
+        let flatPath = (method as any).flatPath;
+        pathParamsForFlatPathParam.forEach((pathParam, flatPathParam) => {
+          flatPath = flatPath.replace(flatPathParam, pathParam);
+        });
+
+        const cleanedParams: { [key: string]: any } = {};
```
sam-gc, line 154: Can this `any` be removed as well?

scripts/gen-auth-api-spec.ts
```diff
@@ -130,6 +128,70 @@ async function toOpenapi3(discovery: any): Promise<any> {
...
+          if (!Object.keys(cleanedParams).some((name) => name === param)) {
+            cleanedParams[param] = {
+              location: "path",
+              required: true,
+              type: "string",
+            };
+          }
+        });
+
+        (method as any).parameters = cleanedParams;
```
sam-gc, line 182: It would probably be worth creating some sort of interface to cast `method` to instead

sam-gc (commented):
Generally looks good, but if at all possible I'd prefer if we could remove as many of the `any` types as we can (go/tsstyle#any)

I'll defer to Yuchen regarding the spec changes, which I'm not at all familiar with

lisajian:
> Generally looks good, but if at all possible I'd prefer if we could remove as many of the `any` types as we can (go/tsstyle#any)
> 
> I'll defer to Yuchen regarding the spec changes, which I'm not at all familiar with

Ah, shoot you're right - sorry about that, I tunnel visioned trying to get it working that I glossed over the style guide, let me see if I can get some more typing here and send out a new revision

scripts/gen-auth-api-spec.ts
yuchenshi: Most of the objects from the schema won't have good typing since they are a Google-specific format. You can however write your own interfaces, if that's worth your time, on a case by case basis.

scripts/gen-auth-api-spec.ts
```diff
@@ -130,6 +128,70 @@ async function toOpenapi3(discovery: any): Promise<any> {
...
+        let flatPath = (method as any).flatPath;
+        pathParamsForFlatPathParam.forEach((pathParam, flatPathParam) => {
+          flatPath = flatPath.replace(flatPathParam, pathParam);
+        });
+
+        const cleanedParams: { [key: string]: any } = {};
+
+        // Get all param names in path
+        // e.g. ["projectsId", "defaultSupportedIdpConfigsId"]
+        const pathParamMatches = [...flatPath.matchAll(paramPattern)];
```
yuchenshi, line 158: nit: inline this

yuchenshi (approved):
Approach looks good, nits inline

scripts/gen-auth-api-spec.ts
lisajian: No longer applicable after adding in `Method` interface

scripts/gen-auth-api-spec.ts
lisajian: Added in `Method` interface

scripts/gen-auth-api-spec.ts
lisajian: Done

lisajian:
Added interfaces based on [discovery doc format](https://developers.google.com/discovery/v1/using)

PR #3767, opened 2021-09-21
Generate CloudEvents in the PubSub Emulator

taeold:
Update PubSub Emulator to emit events in CloudEvent format given a trigger whose function signature is `cloudevent`.

Previously, the Functions Emulator considered 2 types of functions - `https` and `background`. With the introduction of [CloudEvents](https://cloud.google.com/functions/docs/writing/cloudevents) in Google Cloud Functions,  we now have 2 types of `background` functions - `legacy` vs `cloudevent`. For example:

```js
/* Legacy Events */
exports.helloPubSub = functions.pubsub.topic('topic-name').onPublish((message) => {
  // ...
});

/* CloudEvents */
exports.helloPubSub = functions.v2.pubsub.topic('topic-name').onPublish((cloudevent) => {
  const message = cloudevent.data.message;
});
```

To accommodate, we adopt concept of "function signature" introduced in Google Cloud Function's [Functions Framework](https://github.com/GoogleCloudPlatform/functions-framework-nodejs) to categorize each function trigger as one of `http`, `event`, or`cloudevent`.

When registering a function trigger with the Pubsub emulator, the functions's signature type will be included in the registration request, and the Pubsub emulator will use the info to emit an event in the format expected by the underlying function trigger. We also make minor changes to the Functions Emulator to selectively massage the incoming events in-route to the function trigger.

As an added bonus, we are now able to remove `triggerType` field in the `FunctionsRuntimeArg` as that info is now outdated and not needed when invoking an emulated function.

### DIFF
--- a/scripts/triggers-end-to-end-tests/functions/index.js
+++ b/scripts/triggers-end-to-end-tests/functions/index.js
@@ -1,7 +1,13 @@
 const admin = require("firebase-admin");
 const functions = require("firebase-functions");
-// TODO(taeold): firebase-functions/lib path is unsupported, but this is the only way to access v2 namesapce in Node10
-const functionsV2 = require("firebase-functions/lib/v2");
+let functionsV2;
+try {
+  functionsV2 = require("firebase-functions/v2");
+} catch {
+  // TODO: firebase-functions/lib path is unsupported, but this is the only way to access the v2 namespace in Node 10.
+  // Remove this ugly hack once we cut support for Node 10.
+  functionsV2 = require("firebase-functions/lib/v2");
+}
 const { PubSub } = require("@google-cloud/pubsub");
 
 /*
--- a/scripts/triggers-end-to-end-tests/functions/index.js
+++ b/scripts/triggers-end-to-end-tests/functions/index.js
@@ -1,6 +1,6 @@
 const admin = require("firebase-admin");
 const functions = require("firebase-functions");
-// TODO(taeold): fireebase-functions/lib path is unsupported, but this is the only way to access v2 namesapce in Node10
+// TODO(taeold): firebase-functions/lib path is unsupported, but this is the only way to access v2 namesapce in Node10
 const functionsV2 = require("firebase-functions/lib/v2");
 const { PubSub } = require("@google-cloud/pubsub");
 
--- a/scripts/triggers-end-to-end-tests/functions/index.js
+++ b/scripts/triggers-end-to-end-tests/functions/index.js
@@ -1,15 +1,19 @@
 const admin = require("firebase-admin");
 const functions = require("firebase-functions");
+const functionsV2 = require("firebase-functions/v2");
 const { PubSub } = require("@google-cloud/pubsub");
 
 /*
  * Log snippets that the driver program above checks for. Be sure to update
  * ../test.js if you plan on changing these.
  */
+/* Functions V1 */
 const RTDB_FUNCTION_LOG = "========== RTDB FUNCTION ==========";
 const FIRESTORE_FUNCTION_LOG = "========== FIRESTORE FUNCTION ==========";
 const PUBSUB_FUNCTION_LOG = "========== PUBSUB FUNCTION ==========";
 const AUTH_FUNCTION_LOG = "========== AUTH FUNCTION ==========";
+/* Functions V2 */
+const PUBSUB_FUNCTION_V2_LOG = "========== PUBSUB V2 FUNCTION ==========";
 
 /*
  * We install onWrite triggers for START_DOCUMENT_NAME in both the firestore and
@@ -118,6 +122,13 @@ exports.pubsubReaction = functions.pubsub.topic(PUBSUB_TOPIC).onPublish((msg /* 
   console.log(PUBSUB_FUNCTION_LOG);
   console.log("Message", JSON.stringify(msg.json));
   console.log("Attributes", JSON.stringify(msg.attributes));
+  return true;
+});
+
+exports.pubsubv2reaction = functionsV2.pubsub.onMessagePublished(PUBSUB_TOPIC, (cloudevent) => {
+  console.log(PUBSUB_FUNCTION_V2_LOG);
+  console.log("Message", JSON.stringify(cloudevent.data.message.json));
+  console.log("Attributes", JSON.stringify(cloudevent.data.message.attributes));
   return true;
 });
 
--- a/scripts/triggers-end-to-end-tests/functions/index.js
+++ b/scripts/triggers-end-to-end-tests/functions/index.js
@@ -1,6 +1,7 @@
 const admin = require("firebase-admin");
 const functions = require("firebase-functions");
-const functionsV2 = require("firebase-functions/v2");
+// TODO(taeold): fireebase-functions/lib path is unsupported, but this is the only way to access v2 namesapce in Node10
+const functionsV2 = require("firebase-functions/lib/v2");
 const { PubSub } = require("@google-cloud/pubsub");
 
 /*
--- a/src/emulator/events/types.ts
+++ b/src/emulator/events/types.ts
@@ -9,6 +9,7 @@
 import * as _ from "lodash";
 
 import { Resource } from "firebase-functions";
+import * as express from "express";
 
 /**
  * Wire formal for v1beta1 EventFlow.
@@ -39,6 +40,49 @@ export interface Event {
 }
 
 /**
+ * A CloudEvent is a cross-platform format for encoding a serverless event.
+ * More information can be found in https://github.com/cloudevents/spec
+ */
+export interface CloudEvent<T> {
+  /** Version of the CloudEvents spec for this event. */
+  specversion: string;
+
+  /** A globally unique ID for this event. */
+  id: string;
+
+  /** The resource which published this event. */
+  source: string;
+
+  /** The resource, provided by source, that this event relates to */
+  subject?: string;
+
+  /** The type of event that this represents. */
+  type: string;
+
+  /** When this event occurred. */
+  time: string;
+
+  /** Information about this specific event. */
+  data: T;
+
+  /**
+   * A map of template parameter name to value for subject strings.
+   *
+   * This map is only available on some event types that allow templates
+   * in the subject string, such as Firestore. When listening to a document
+   * template "/users/{uid}", an event with subject "/documents/users/1234"
+   * would have a params of {"uid": "1234"}.
+   *
+   * Params are generated inside the firebase-functions SDK and are not
+   * part of the CloudEvents spec nor the payload that a Cloud Function
+   * actually receives.
+   */
+  params?: Record<string, string>;
+}
+
+export type CloudEventContext = Omit<CloudEvent<unknown>, "data" | "params">;
+
+/**
  * Legacy AuthMode format.
  */
 export interface AuthMode {
@@ -47,7 +91,7 @@ export interface AuthMode {
 }
 
 /**
- * Utilities for determining event types.
+ * Utilities for operating on event types.
  */
 export class EventUtils {
   static isEvent(proto: any): proto is Event {
@@ -56,5 +100,25 @@ export class EventUtils {
 
   static isLegacyEvent(proto: any): proto is LegacyEvent {
     return _.has(proto, "data") && _.has(proto, "resource");
+  }
+
+  static isBinaryCloudEvent(req: express.Request): boolean {
+    return !!(
+      req.header("ce-type") &&
+      req.header("ce-specversion") &&
+      req.header("ce-source") &&
+      req.header("ce-id")
+    );
+  }
+
+  static extractBinaryCloudEventContext(req: express.Request): CloudEventContext {
+    const context: Partial<CloudEventContext> = {};
+    for (const name of Object.keys(req.headers)) {
+      if (name.startsWith("ce-")) {
+        const attributeName = name.substr("ce-".length) as keyof CloudEventContext;
+        context[attributeName] = req.header(name);
+      }
+    }
+    return context as CloudEventContext;
   }
 }
--- a/src/emulator/functionsEmulator.ts
+++ b/src/emulator/functionsEmulator.ts
@@ -5,6 +5,7 @@ import * as express from "express";
 import * as clc from "cli-color";
 import * as http from "http";
 import * as jwt from "jsonwebtoken";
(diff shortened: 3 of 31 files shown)


src/emulator/functionsEmulator.ts
```diff
@@ -229,7 +231,16 @@ export class FunctionsEmulator implements EmulatorInstance {
       const projectId = req.params.project_id;
 
       const reqBody = (req as RequestWithRawBody).rawBody;
-      const proto = JSON.parse(reqBody.toString());
+      let proto = JSON.parse(reqBody.toString());
```
yuchenshi, line 234: nit: Consider renaming this (now that it is no longer a proto)

src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -887,9 +903,9 @@ async function invokeTrigger(
 
   const trigger = triggers[frb.triggerId];
   logDebug("triggerDefinition", trigger.definition);
-  const mode = trigger.definition.httpsTrigger ? "HTTPS" : "BACKGROUND";
+  const signature = getSignatureType(trigger.definition);
 
-  logDebug(`Running ${frb.triggerId} in mode ${mode}`);
+  logDebug(`Running ${frb.triggerId} in mode ${signature}`);
```
yuchenshi, line 908:
```suggestion
  logDebug(`Running ${frb.triggerId} with signature ${signature}`);
```

src/emulator/functionsEmulatorShared.ts
```diff
@@ -4,13 +4,12 @@ import * as os from "os";
...
+
 import { Constants } from "./constants";
 import { InvokeRuntimeOpts } from "./functionsEmulator";
+import { FunctionsPlatform } from "../deploy/functions/backend";
 
-export enum EmulatedTriggerType {
-  BACKGROUND = "BACKGROUND",
-  HTTPS = "HTTPS",
-}
+export type SignatureType = "http" | "event" | "cloudevent";
```
yuchenshi, line 12: nit: Why string instead of enum?

src/emulator/pubsubEmulator.ts
```diff
@@ -16,14 +19,19 @@ export interface PubsubEmulatorArgs {
...
 export class PubsubEmulator implements EmulatorInstance {
   pubsub: PubSub;
 
   // Map of topic name to a list of functions to trigger
-  triggers: Map<string, Set<string>>;
+  triggersByTopic: Map<string, { triggerKeys: Set<string>; triggers: Trigger[] }>;
 
   // Map of topic name to a PubSub subscription object
-  subscriptions: Map<string, Subscription>;
+  subscriptionsByTopic: Map<string, Subscription>;
```
yuchenshi, line 34: Consider renaming to `subscriptionByTopic` (drop `s` since there is only one subscription per topic)

src/emulator/pubsubEmulator.ts
```diff
@@ -16,14 +19,19 @@ export interface PubsubEmulatorArgs {
...
+  triggerKey: string;
+  signatureType: SignatureType;
+}
+
 export class PubsubEmulator implements EmulatorInstance {
   pubsub: PubSub;
 
   // Map of topic name to a list of functions to trigger
-  triggers: Map<string, Set<string>>;
+  triggersByTopic: Map<string, { triggerKeys: Set<string>; triggers: Trigger[] }>;
```
yuchenshi, line 31: Please change to `triggersByTopicAndKey: Map<string, Map<string, Trigger>>`. Having to keep the key set in sync with array is annoying and may lead to edge cases.

In addition, consider dropping `Trigger#triggerKey` if not needed to reduce duplication. You can just loop around the inner map `entries`.

yuchenshi (approved):

src/emulator/functionsEmulatorShared.ts
taeold: I learned to prefer string unions from @joehan and @inlined. Found some interesting discussions on enums vs union type [online](https://stackoverflow.com/questions/40275832/typescript-has-unions-so-are-enums-redundant) - do you have preference for enums?

src/emulator/functionsEmulator.ts
taeold: This is true, but `proto` name is used throughout the `functionsEmulator.ts` and ultimately used to include in `FunctionsRuntimeBundle`. I'm leaning towards cleaning this up when we start to trim `frb` more aggrssively.

src/emulator/pubsubEmulator.ts
taeold: Great suggestion. Simplified the data structure and renamed the variable.

scripts/triggers-end-to-end-tests/functions/index.js
```diff
@@ -1,15 +1,20 @@
 const admin = require("firebase-admin");
 const functions = require("firebase-functions");
+// TODO(taeold): firebase-functions/lib path is unsupported, but this is the only way to access v2 namesapce in Node10
+const functionsV2 = require("firebase-functions/lib/v2");
```
inlined, line 4: Yet this will throw in Node 12. I would at least make this a try require for both formats (prefer the correct one)

src/emulator/functionsEmulator.ts
```diff
@@ -229,7 +231,16 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+
+      if (req.headers["content-type"]?.includes("cloudevent")) {
+        // Convert request payload to CloudEvent.
+        // TODO(taeold): Converting request payload to CloudEvent object should be done by the functions runtime.
+        // However, the Functions Emulator communicates with the runtime via socket not HTTP, and CE metadata
+        // embedded in HTTP header may get lost. Once the Functions Emulator is refactored to communicate to the
+        // runtime instances via HTTP, move this logic there.
+        proto = HTTP.toEvent({ headers: req.headers, body: proto });
```
inlined, line 242: I have mixed feelings about the CloudEvents SDK (partially just because I was strongly against the committee including canonical SDKs). If we're going to put it in the prod SDK we should see how big the SDK is. AFAIK functions framework doesn't use the CloudEvents library because we always know the encoding that our clients will use.

Also, I just take issue that `toEvent` calls `Object.freeze` IIRC, which means that if we were to use this in `firebase-functions` we'd have to clone the object just so we could massage `event.data` from a raw wire protocol into an SDK type.

src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -20,6 +22,7 @@ import * as admin from "firebase-admin";
 import * as bodyParser from "body-parser";
 import { pathToFileURL, URL } from "url";
 import * as _ from "lodash";
+import { CloudEvent } from "firebase-functions/lib/v2";
```
inlined, line 25: 1. This will crash in Node 12 and above
2. This makes me very nervous since `firebase-functions` is only a dev dependency. I wouldn't mind a comment warning future devs to only import interfaces because the SDK is only a dev dependency.

src/emulator/functionsEmulatorShared.ts
```diff
@@ -290,3 +289,10 @@ export function formatHost(info: { host: string; port: number }): string {
     return `${info.host}:${info.port}`;
   }
 }
+
+export function getSignatureType(def: EmulatedTriggerDefinition): SignatureType {
+  if (def.httpsTrigger) {
+    return "http";
+  }
+  return def.platform === "gcfv2" ? "cloudevent" : "event";
```
inlined, line 297: Is this our product decision? That "compat mode" where firebase-functions/v1 is used for GCFv2 or firebase-functions/v2 is used for GCFv1 isn't supported in the emulator?

We need to think deeply about how that scenario should work, but it's too late for me to think deeply tonight.

src/emulator/pubsubEmulator.ts
```diff
@@ -66,11 +74,18 @@ export class PubsubEmulator implements EmulatorInstance {
...
+    );
 
-    const topicTriggers = this.triggers.get(topicName) || new Set();
-    if (topicTriggers.has(topicName) && this.subscriptions.has(topicName)) {
+    const triggers = this.triggersForTopic.get(topicName) || [];
+    if (
+      triggers.some((t) => t.triggerKey === triggerKey) &&
+      this.subscriptionForTopic.has(topicName)
+    ) {
       this.logger.logLabeled("DEBUG", "pubsub", "Trigger already exists");
```
inlined, line 89: Couldn't two functions listen to the same topic?

src/emulator/pubsubEmulator.ts
```diff
@@ -105,20 +120,71 @@ export class PubsubEmulator implements EmulatorInstance {
...
+          attributes: message.attributes,
+          orderingKey: message.orderingKey,
+          data: message.data.toString("base64"),
+        },
+        subscription: this.subscriptionForTopic.get(topic)!.name,
+      };
+      const ce = HTTP.structured(
+        new CloudEvent({
+          type: "google.cloud.pubsub.topic.v1.messagePublished",
+          source: `//pubsub.googleapis.com/projects/${this.args.projectId}/topics/${topic}`,
```
inlined, line 171: Really? It's all `source` and none of it is `subject`?

src/emulator/pubsubEmulator.ts
```diff
@@ -105,20 +120,71 @@ export class PubsubEmulator implements EmulatorInstance {
...
+        message: {
+          messageId: message.id,
+          publishTime: message.publishTime,
+          attributes: message.attributes,
+          orderingKey: message.orderingKey,
+          data: message.data.toString("base64"),
+        },
+        subscription: this.subscriptionForTopic.get(topic)!.name,
+      };
+      const ce = HTTP.structured(
```
inlined, line 168: Is this library giving us any more value than saying

```
{
  specVersion: 1,
  type: "google.cloud.pubsub.topic.v1.messagePublished",
  source: `//pubsub.googleapis.com/projects/${this.args.projectId}/topics/${topic}`
  data,
}
```

src/emulator/pubsubEmulator.ts
```diff
@@ -105,20 +120,71 @@ export class PubsubEmulator implements EmulatorInstance {
...
+        new CloudEvent({
+          type: "google.cloud.pubsub.topic.v1.messagePublished",
+          source: `//pubsub.googleapis.com/projects/${this.args.projectId}/topics/${topic}`,
+          data,
+        })
+      );
+      return { ...baseOpts, data: ce.body, headers: ce.headers };
```
inlined, line 175: You don't really need to worry about ce.body and ce.headers if you don't want to. You can just set the contenttype to the cloudevent content type and put everything in body (structured vs binary encoding; both are allowed in the spec)

inlined (commented):
Great work. Between it being late and my lack of familiarity I shouldn't be the final approver, but LGTM.

src/emulator/functionsEmulatorShared.ts
yuchenshi: I'm fine either way, but you know there's gonna be a nit question when you change things :)

src/emulator/functionsEmulator.ts
taeold: Thanks for flagging this. Answering couple of issues you raised here for the record:

1. cloudevents npm package is [pretty big](https://bundlephobia.com/package/cloudevents@4.0.3) at 145.9KB minified. In contrast, firebase-functions SDK is is [115.7KB](https://bundlephobia.com/package/firebase-functions@3.15.7).

2. Functions Framework seems to take a clever shortcut for converting binary encoded cloudevent to structured one. Going to copy their methods https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/9cf46edc9b1b890e2a6fd1aec8d818074585783b/src/invoker.ts#L146-L149.

src/emulator/functionsEmulatorRuntime.ts
taeold: Great points. I don't think correctly typing the function that received `CloudEvent` is that useful anyway - removing this import.

src/emulator/functionsEmulatorShared.ts
taeold: This is a practical decision - there aren't any info I can use in the trigger definition to distinguish functions expecting events in legacy or cloudevent format.  I'll add a todo for us to update this logic once we have a path forward on how declarative function signature will work w/ the Firebase Functions SDK.

src/emulator/pubsubEmulator.ts
taeold: It was a bit confusing, but I learned that in the PubSub emulator, there is a single subscription per pubsub topic, and the events sent on that subscription is routed to all triggers listening on the topic.

src/emulator/pubsubEmulator.ts
taeold: 🤷  That's what I saw:

Screen Shot 2021-09-22 at 11 35 52 PM

src/emulator/pubsubEmulator.ts
taeold: Done!

src/emulator/pubsubEmulator.ts
taeold: Yes that should work?

src/emulator/events/types.ts
```diff
@@ -38,6 +39,49 @@ export interface Event {
...
+   * This map is only available on some event types that allow templates
+   * in the subject string, such as Firestore. When listening to a document
+   * template "/users/{uid}", an event with subject "/documents/users/1234"
+   * would have a params of {"uid": "1234"}.
+   *
+   * Params are generated inside the firebase-functions SDK and are not
+   * part of the CloudEvents spec nor the payload that a Cloud Function
+   * actually receives.
+   */
+  params?: Record<string, string>;
```
inlined, line 80: Params isn't part of the wire protocol, so we should be able to keep it out of the CLI.

src/emulator/functionsEmulatorShared.ts
inlined: WDYT about the idea that we'll use the FUNCTION_SIGNATURE_TYPE env variable in the SDK if the event type doesn't match the default for the platform? So if someone does `functions.runWith({platform: "gcfv2"}).pubsub.topic('foo').onPublish(handler)` we'll add `FUNCTION_SIGNATURE_TYPE=event` in the env annotation.

So that just means that we need to check the env variable here before we check platform.

PR #3771, opened 2021-09-22
Propose some functional helpers

inlined:
While reviewing #3690 I noticed that (A) one of the functions are generic enough to go in a library and (B) I really wish we had tools like zip/flatten/etc here. While writing my own code, I wished TypeScript could either assert or detect that code was dead due to type guards. Now we have all the above!

I leaned into using generator functions because they work just as well in `for-of` loops with less syntactic goop. LMK if you feel we should be returning arrays instead.

Feel free to tell me that these should be named differently, that we're missing something obvious, or that if God wanted us to have better list comprehensions they would have joined TC39.

### DIFF
--- a/src/functional.ts
+++ b/src/functional.ts
@@ -1,0 +1,80 @@
+
+/**
+ * Flattens an object so that the return value's keys are the path
+ * to a value in the source object. E.g. flattenObject({the: {answer: 42}})
+ * returns {"the.answser": 42}
+ * @param obj An object to be flattened
+ * @returns An array where values come from obj and keys are the path in obj to that value.
+ */
+export function flattenObject(obj: Record<string, unknown>): Record<string, unknown> {
+  function* helper(path: string[], obj: Record<string, unknown>): Generator<[string, unknown]> {
+    for (const [k, v] of Object.entries(obj)) {
+      if (typeof v !== "object" || v === null) {
+        yield [[...path, k].join("."), v];
+      } else {
+        // Object.entries loses type info, so we must cast
+        yield* helper([...path, k], v as Record<string, unknown>);
+      }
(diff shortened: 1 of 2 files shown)


bkendall (approved):
LGTM. First one to venture into making generators, but I have no problem with that 👍

src/functional.ts
```diff
@@ -0,0 +1,90 @@
...
+/** Flattens an object or array. */
+export function flatten<T>(
+  objOrArr: Record<string, unknown> | unknown[]
+): Generator<[string, unknown]> | Generator<T> {
+  if (Array.isArray(objOrArr)) {
+    return flattenArray<T>(objOrArr);
+  } else {
+    return flattenObject(objOrArr);
+  }
+}
```
taeold, line 53: yessss

src/test/functional.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+
+describe("functional", () => {
+  describe("flatten", () => {
+    it("Can iterate an empty object", () => {
+      expect([...f.flatten({})]).to.deep.equal([]);
+    });
+
+    it("Can iterate an object that's already flat", () => {
+      expect([...f.flatten({ a: "b" })]).to.deep.equal([["a", "b"]]);
+    });
```
taeold, line 15:
```suggestion
    it("can iterate an empty object", () => {
      expect([...f.flatten({})]).to.deep.equal([]);
    });

    it("can iterate an object that's already flat", () => {
      expect([...f.flatten({ a: "b" })]).to.deep.equal([["a", "b"]]);
    });
```

taeold (approved):
👏 This is so much fun.

PR #3776, opened 2021-09-23
GCS Enhancement

colerogers:
### Description

Enhancement on how customers deploy GCS triggered functions. Relevant SDK changes found [here](https://github.com/firebase/firebase-functions/pull/981)

### Scenarios Tested

- Deploying GCS triggered function
- Transforming previously deployed GCS triggered function


src/test/gcp/cloudfunctionsv2.spec.ts
```diff
@@ -460,8 +462,96 @@ describe("cloudfunctionsv2", () => {
       };
 
       expect(
-        cloudfunctionsv2.functionFromSpec(complexFunction, CLOUD_FUNCTION_V2_SOURCE)
+        await cloudfunctionsv2.functionFromSpec(complexFunction, CLOUD_FUNCTION_V2_SOURCE)
+      ).to.deep.equal(complexGcfFunction);
+    });
+
+    it("should calculate non-trivial fields with a GCS trigger", async () => {
+      const storageStub = sinon
```
taeold, line 470: *nit: I'd personally setup this stub in `beforeAll` and restore in `afterAll` block even when the some test case does not require `getStorageBucket` to be stubbed. That way, I avoid any possibility of pulling my hair our wondering why my test cases are failnig only to realize I forgot to `stub.restore()` after one test case...

src/test/gcp/cloudfunctionsv2.spec.ts
```diff
@@ -460,8 +462,96 @@ describe("cloudfunctionsv2", () => {
       };
 
       expect(
-        cloudfunctionsv2.functionFromSpec(complexFunction, CLOUD_FUNCTION_V2_SOURCE)
+        await cloudfunctionsv2.functionFromSpec(complexFunction, CLOUD_FUNCTION_V2_SOURCE)
+      ).to.deep.equal(complexGcfFunction);
```
taeold, line 466: nit* This is identical to

```
await expect(asuncFn()).to.eventually.be.deep.equal(expected);
```

src/test/gcp/cloudfunctionsv2.spec.ts
```diff
@@ -460,8 +462,96 @@ describe("cloudfunctionsv2", () => {
...
+          eventType: "google.cloud.storage.object.v1.archived",
+          eventFilters: {
+            resource: "my-bucket",
+          },
+          retry: false,
+        },
+      };
+
+      await expect(cloudfunctionsv2.functionFromSpec(gcsFunction, CLOUD_FUNCTION_V2_SOURCE)).to.be
+        .rejected;
```
taeold, line 553:
```suggestion
        .rejectedWith("Can't find the storage bucket region");
```

taeold (approved):
nits, lgtm!

src/gcp/cloudfunctionsv2.ts
```diff
@@ -386,6 +394,19 @@ export function functionFromSpec(cloudFunction: backend.FunctionSpec, source: St
     };
     if (gcfFunction.eventTrigger.eventType === PUBSUB_PUBLISH_EVENT) {
       gcfFunction.eventTrigger.pubsubTopic = cloudFunction.trigger.eventFilters.resource;
+    } else if (GCS_EVENTS.has(gcfFunction.eventTrigger.eventType)) {
```
inlined, line 397: I see why you extended `else if` off of the previous `if`, but Pub/Sub _must_ be represented differently in GCFv2, GCS doesn't. If our hack is here, then the container contract _also_ needs to use `resource` instead of `bucket`, which makes this everlasting techncial debt.

Instead, please put logic in triggerParser to detect the event is a GCS event and store the TriggerAnnotation's `resource` as the `FunctionSpec` and `Endpoint`'s `bucket`. That way the GCP folder can stay simpler.

src/gcp/storage.js
```diff
@@ -91,9 +91,31 @@ function _deleteObject(location) {
   });
 }
 
+/**
+ * Gets a storage bucket from GCP.
+ * Ref: https://cloud.google.com/storage/docs/json_api/v1/buckets/get
+ * @param {string} bucketName name of the storage bucket
+ * @returns a bucket resource object
+ */
+async function _getStorageBucket(bucketName) {
```
inlined, line 100: Nit: This method is probably just called `getBucket` in the actual API, so we should probably use the same.

src/deploy/functions/tasks.ts
```diff
@@ -94,7 +95,10 @@ export function createFunctionTask(
       }
       op = await gcf.createFunction(apiFunction);
     } else {
-      const apiFunction = await gcfV2.functionFromSpec(fn, params.storage![fn.region]);
+      if (backend.isEventTrigger(fn.trigger) && !fn.trigger.region) {
+        await setTriggerRegionFromTriggerType(fn.trigger);
```
inlined, line 99: Let's do simple lookups before the release phase. I'd rather these errors be discovered before we start changing the customer's backend.

inlined (approved):

src/test/gcp/cloudfunctionsv2.spec.ts
colerogers: Fixed

src/test/gcp/cloudfunctionsv2.spec.ts
colerogers: Changed the method and removed this code change

src/gcp/cloudfunctionsv2.ts
colerogers: Moved all this logic into triggerParser

src/gcp/cloudfunctionsv2.ts
```diff
@@ -322,7 +330,7 @@ export async function deleteFunction(cloudFunction: string): Promise<Operation>
   }
 }
 
-export function functionFromSpec(cloudFunction: backend.FunctionSpec, source: StorageSource) {
+export async function functionFromSpec(cloudFunction: backend.FunctionSpec, source: StorageSource) {
```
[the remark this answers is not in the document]
colerogers: Removed the API call from this method and placed it in the prepare stage

src/gcp/storage.js
colerogers: Changed to getBucket

src/deploy/functions/tasks.ts
colerogers: I believe the best place to perform this lookup is in the prepare step. Since we are already looping through each functionSpec and checking for a cached value, performing the API lookup there seems like the best place.

src/test/gcp/cloudfunctionsv2.spec.ts
colerogers: Fixed

PR #3777, opened 2021-09-23
Add tenant management endpoints support

lisajian:
### Description

Add support for `create`, `delete`, `get`, `list`, and `patch` project tenants. This change also includes a bit of refactoring/restructuring to support these endpoints given the use of flatPaths instead of paths (#3761), including:

* Removed `AgentProject` interface from `server.ts` and move `tenantId` |-> `TenantProjectState` map to `AgentProjectState`
* Moved `Tenant` reference to `TenantProjectState`

Associated internal bug: b/192387245

### Scenarios Tested

`npm test` passes

### Sample Commands

N/A

### DIFF
--- a/src/emulator/auth/operations.ts
+++ b/src/emulator/auth/operations.ts
@@ -31,6 +31,8 @@ import {
   SecondFactorRecord,
   UsageMode,
   AgentProjectState,
+  Tenant,
+  TenantProjectState,
 } from "./state";
 import { MfaEnrollments, Schemas } from "./types";
 
@@ -82,6 +84,14 @@ export const authOperations: AuthOps = {
         batchDelete,
         batchGet,
       },
+      tenants: {
+        create: createTenant,
+        delete: deleteTenant,
+        get: getTenant,
+        list: listTenants,
+        patch: updateTenant,
+      },
+      // TODO(lisajian): Adjust all other endpoints that use tenantId
     },
   },
   securetoken: {
@@ -109,6 +119,9 @@ export const authOperations: AuthOps = {
 /* Handlers */
 
 const PASSWORD_MIN_LENGTH = 6;
+const PHONE_NUMBER_MAX_LENGTH = 250;
+const MIN_LENGTH_FOR_NSN = 2;
+const MAX_LENGTH_FOR_NSN = 17;
 
 // https://cloud.google.com/identity-platform/docs/use-rest-api#section-verify-custom-token
 export const CUSTOM_TOKEN_AUDIENCE =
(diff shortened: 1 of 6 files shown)


src/emulator/auth/operations.ts
```diff
@@ -2543,6 +2556,123 @@ function parsePendingCredential(
   return { user, signInProvider };
 }
 
+function createTenant(
+  state: ProjectState,
+  reqBody: Schemas["GoogleCloudIdentitytoolkitAdminV2Tenant"]
+): Schemas["GoogleCloudIdentitytoolkitAdminV2Tenant"] {
+  assert(state instanceof AgentProjectState, "((Can only create tenant in agent project.))");
+
+  const formattedPhoneNumbers: { [key: string]: string } = {};
```
sam-gc, line 2565: This can be a `Record<string, string>` type: https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype

src/emulator/auth/operations.ts
```diff
@@ -2543,6 +2556,123 @@ function parsePendingCredential(
...
+    mfaConfig: reqBody.mfaConfig,
+    tenantId: "", // Placeholder until one is generated
+  };
+
+  return state.createTenant(tenant);
+}
+
+function formatPhoneNumber(phoneNumber: string) {
+  assert(
+    phoneNumber.length < PHONE_NUMBER_MAX_LENGTH,
```
sam-gc, line 2588: `PHONE_NUMBER_MAX_LENGTH` implies phoneNumber can also equal that length. Should it instead be `<=` ?

src/emulator/auth/operations.ts
```diff
@@ -2543,6 +2556,123 @@ function parsePendingCredential(
...
+    "TOO_LONG ((The string supplied was too long to parse.))"
+  );
+
+  // Only check that the phone number is reasonably formatted, but does not
+  // fully verify that phone number is valid.
+  // TODO: support unicode, alpha characters and extensions?
```
sam-gc, line 2594: Consider explaining this regex a bit in the comment (high level, what is it matching)

Also, I don't think you need to do `new RegExp`; the literal `/.../g` itself is already an instance of RegExp that is created at compile time.

src/emulator/auth/operations.ts
```diff
@@ -2543,6 +2556,123 @@ function parsePendingCredential(
...
+}
+
+function listTenants(
+  state: ProjectState,
+  reqBody: unknown,
+  ctx: ExegesisContext
+): Schemas["GoogleCloudIdentitytoolkitAdminV2ListTenantsResponse"] {
+  assert(state instanceof AgentProjectState, "((Can only list tenants in agent project.))");
+  const pageSize = Math.min(Math.floor(ctx.params.query.pageSize) || 20, 1000);
+  const tenants: Tenant[] = state.listTenants(ctx.params.query.pageToken);
```
sam-gc, line 2617: I believe the type of this variable can be inferred

src/emulator/auth/operations.ts
```diff
@@ -2543,6 +2556,123 @@ function parsePendingCredential(
...
+}
+
+function updateTenant(
+  state: ProjectState,
+  reqBody: Schemas["GoogleCloudIdentitytoolkitAdminV2Tenant"],
+  ctx: ExegesisContext
+): Schemas["GoogleCloudIdentitytoolkitAdminV2Tenant"] {
+  assert(state instanceof TenantProjectState, "((Can only update tenant on tenant projects.))");
+  const update: Partial<Tenant> = { ...reqBody };
+  const formattedPhoneNumbers: { [key: string]: string } = {};
```
sam-gc, line 2665: Same here re. `Record<string, string>`

sam-gc (commented):
Nice tests

src/emulator/auth/operations.ts
lisajian: Sick, done!

src/emulator/auth/operations.ts
lisajian: Ah yes yes, thanks! Removed the explicit typing

src/emulator/auth/operations.ts
lisajian: Sick, good call, thanks!

src/emulator/auth/operations.ts
lisajian: Ah you're right, I meant for it to be <=, thanks for the catch!

src/emulator/auth/operations.ts
lisajian: Added a comment that makes the regex a bit more readable and removed the `new RegExp`, thanks!

src/emulator/auth/operations.ts
```diff
@@ -2543,6 +2556,126 @@ function parsePendingCredential(
...
 }
 
+function createTenant(
+  state: ProjectState,
+  reqBody: Schemas["GoogleCloudIdentitytoolkitAdminV2Tenant"]
+): Schemas["GoogleCloudIdentitytoolkitAdminV2Tenant"] {
+  assert(state instanceof AgentProjectState, "((Can only create tenant in agent project.))");
+
+  const formattedPhoneNumbers: Record<string, string> = {};
+  if (reqBody.testPhoneNumbers) {
```
lisajian, line 2566: Discussed w/ Yuchen offline - should remove this field

sam-gc (approved):

src/emulator/auth/state.ts
```diff
@@ -597,36 +597,57 @@ export class AgentProjectState extends ProjectState {
     this._usageMode = usageMode;
   }
 
-  // TODO(lisajian): Fill in when v2.projects.tenants.get is added
-  getTenant(): void {
-    throw new NotImplementedError("getTenant is not implemented yet.");
+  getTenantProject(tenantId: string): TenantProjectState {
+    assert(this.tenantProjectForTenantId.has(tenantId), "TENANT_NOT_FOUND");
```
lisajian, line 601: Discussed offline during emulator sync - should create tenant if not present. Will include in next revision

src/emulator/auth/state.ts
```diff
@@ -597,36 +597,66 @@ export class AgentProjectState extends ProjectState {
     this._usageMode = usageMode;
   }
 
-  // TODO(lisajian): Fill in when v2.projects.tenants.get is added
-  getTenant(): void {
-    throw new NotImplementedError("getTenant is not implemented yet.");
+  getTenantProject(tenantId: string): TenantProjectState {
```
lisajian, line 600: [Non-actionable] Highlighting this line to reference in PR #3795

yuchenshi (approved):
Will there be a different PR for enabling sign-in methods on implicit tenant creation?

src/emulator/auth/operations.ts
```diff
@@ -2543,6 +2556,123 @@ function parsePendingCredential(
...
+  state.delete();
+  return {};
+}
+
+function getTenant(
+  state: ProjectState,
+  reqBody: unknown,
+  ctx: ExegesisContext
+): Schemas["GoogleCloudIdentitytoolkitAdminV2Tenant"] {
+  // Ideally, this would be called on the top level AgentProjectState, but
```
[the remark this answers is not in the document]
lisajian: Ah, perhaps a bit confusing here - this is just highlighting that typically the `AgentProjectState` is in charge of tenant management (e.g. what we see in GCIP console) but this subverts that. I'll remove the comment here and in `deleteTenant` to avoid confusion

src/emulator/auth/state.ts
```diff
@@ -643,6 +664,69 @@ export class TenantProjectState extends ProjectState {
...
+    return this._tenantConfig;
+  }
+
+  delete(): void {
+    this.parentProject.deleteTenant(this.tenantId);
+  }
+
+  updateTenant(update: Partial<Tenant>, updateMask: string | undefined): Tenant {
+    // Empty masks indicate a full update
+    if (!updateMask) {
```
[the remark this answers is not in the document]
lisajian: Hm.. Not sure if I know what you mean in terms of moving `updateMask` logic out to `operations.ts` since this update is all closely tied to the mask. Marking as resolve, but feel free to follow up / reopen if you have further thoughts about it

PR #3779, opened 2021-09-23
Move ref parsing code to a separate file.

joehan:
### Description
While working on a larger project, decided this code was bothering me and wanted to clean it up. Moved out ref parsing code out of extensionsApi.ts, added a few more convenience methods to transform between fully qualified name & ref, and started using those methods in a few places.

 I'm sure there are more places in the codebase that could use these too, but this is a start


src/extensions/changelog.ts
```diff
@@ -5,9 +5,10 @@ import * as semver from "semver";
 import TerminalRenderer = require("marked-terminal");
 import Table = require("cli-table");
 
-import { listExtensionVersions, parseRef } from "./extensionsApi";
+import { listExtensionVersions } from "./extensionsApi";
 import { readFile } from "./localHelper";
 import { logger } from "../logger";
+import * as ref from "./ref";
```
huangjeff5, line 11: nit: I do like the usage of `refHelper` more than `ref` because it is easier to find all references to this file. Perhaps the file itself should also be named `refHelper` for the same reason, and also for consistency with our existing pattern.

huangjeff5 (approved):

src/extensions/changelog.ts
joehan: Chatted elsewhere - renaming this to refs to avoid the xHelper pattern that we've fallen into and also to avoid collision with all of the other uses of ref throughout the codebase

PR #3784, opened 2021-09-24
Functions List fix error message

colerogers:
### Description

When the a project needs to re-auth and the list command fails, the error message is `Error: Failed to list functions undefined`

This change fixes the catch block of the cloud functions api to return a FirebaseError wrapped in a rejected promise instead of a string - `Promise.reject(new FirebaseError(...))` instead of the previous `Promise.reject("...")`. Rejecting with a string instead of an error is not recommended and can hide failures.


src/commands/functions-delete.ts
```diff
@@ -33,58 +34,65 @@ export default new Command("functions:delete [filters...]")
...
+        throw new Error("Command aborted.");
+      }
+      return await deleteFunctions(
+        functionsToDelete,
+        schedulesToDelete,
+        topicsToDelete,
+        appEngineLocation
+      );
+    } catch (err) {
+      throw new FirebaseError(`Failed to delete functions ${err.message}`, {
```
inlined, line 93: I'm wondering if we need err.message or if we should just set `original: err` in the options. Can you try that out and see if it renders correctly?

src/gcp/cloudfunctions.ts
```diff
@@ -432,9 +432,13 @@ async function list(projectId: string, region: string): Promise<ListFunctionsRes
       unreachable: res.body.unreachable || [],
     };
   } catch (err) {
-    logger.debug("[functions] failed to list functions for " + projectId);
+    logger.debug(`[functions] failed to list functions for ${projectId}`);
     logger.debug(`[functions] ${err?.message}`);
-    return Promise.reject(err?.message);
+    return Promise.reject(
```
inlined, line 437: Indeed, you were right that we shouldn't reject/throw with a string. But since this is an async function you can just `throw` instead of returning a rejected promise.

inlined (approved):
LGTM with minor fixes.

src/commands/functions-delete.ts
colerogers: It renders correctly and hides the actual error message under the `--debug` flag

colerogers: Do we want to hide potentially useful error messages under the debug flag? Spitting out the `err.message` can save customers some time debugging.

src/gcp/cloudfunctions.ts
colerogers: Fixed

PR #3786, opened 2021-09-25
Correctly prefix storage emulator routes

jcmorrow:
### Description

Hello! This is my attempt at a bug-fix, I don't *think* there are related issues, but there might be. I sunk like six hours into figuring this out yesterday, and of course the fix turned out to be a change of ~10 characters. Let me give some context as to how I got here:

- I am working on an application that hits the storage API at start-up time, and I wanted to spin up local emulators so that new devs wouldn't need to have a fully-set-up cloud project just to run a local server.
- I got the local emulator running, interacted with it via the WebUI, confirmed that it was working, but when I tried to list the buckets, or list objects in a particular bucket from my Go code, it failed.
- With a bit of digging, I realized that it was failing because the server was returning 501 not implemented.
- I decided to take a different tack, since in the docs of the client library it mentions that buckets are implicitly created upon writing an object in them. I decided to try and write an object to a bucket, and that created the bucket just fine, I could see it in the webUI and everything. You can see me live debugging this a bit on [the Slack](https://firebase-community.slack.com/archives/C013HBXTTL3/p1632547066010000)
- Anyways, long story short, I dug into the code this morning and saw that *some* of the routes for the storage emulator are prefixed with `/storage/v1` and some are not.
- The Go client library explicitly sets that prefix as the root path: https://github.com/googleapis/google-cloud-go/blob/0e7ccf1cc95c84838cc3de722ff4cc2e5098f579/storage/storage.go#L143. I don't know if the other client libraries do this, but it seems reasonable, since all of the documentation includes that prefix.
- I don't see any tests that cover how routes link up to actions, so I'm not adding anything here, but I've confirmed that this works with the Go client library as expected.

### Scenarios Tested

- Creating a bucket from the Go client library without creating an object inside of it, as in this example: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/storage/buckets/create_bucket.go
- Listing buckets from the Go client, as in this example: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/storage/buckets/list_buckets.go

### DIFF
--- a/src/emulator/storage/apis/gcloud.ts
+++ b/src/emulator/storage/apis/gcloud.ts
@@ -27,7 +27,7 @@ export function createCloudEndpoints(emulator: StorageEmulator): Router {
     next();
   });
 
-  gcloudStorageAPI.get("/b", (req, res) => {
+  gcloudStorageAPI.get(["/b", "/storage/v1/b"], (req, res) => {
     res.json({
       kind: "storage#buckets",
       items: storageLayer.listBuckets(),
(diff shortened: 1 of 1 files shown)


jcmorrow:
@googlebot I signed it!

PR #3789, opened 2021-09-28
Invoker user account

zzz815340:
### Description

When setting user accounts on the invoker config, they get pre-fixed with `serviceAccount:`. This causes an error from the api. We can check for the `user:` prefix and not format as a service account.

### Scenarios Tested

The account given is a user account. E2E tested from a local firebase project as well.

### DIFF
--- a/src/gcp/proto.ts
+++ b/src/gcp/proto.ts
@@ -139,7 +139,13 @@ export function getInvokerMembers(invoker: string[], projectId: string): string[
   if (invoker[0] === "public") {
     return ["allUsers"];
   }
-  return invoker.map((inv) => formatServiceAccount(inv, projectId));
+  return invoker.map((inv) => {
+    if (inv.startsWith('user:')) {
+      return inv
+    } else {
+      return formatServiceAccount(inv, projectId)
+    }
+  });
 }
 
 /**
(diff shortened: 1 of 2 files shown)


zzz815340:
still working on getting CLA permission from work

pwmb1:
@googlebot I signed it!

inlined:
Can you explain why it would be a best practice to use human account tokens instead of service account tokens? Service accounts are such a more preferred authentication method that I'm almost surprised user tokens still work.

1. If you build your infrastructure based on individual users, your infrastructure breaks when that user leaves your company and the email is disabled. Service accounts live on.
2. You, as a person, are probably going to have many permissions available to you. Service accounts are easily restricted by removing unused permissions. Cloud will even notify you if there are permissions granted to a service account that you aren't using so you can mitigate the damage an attacker does when they compromise an account
3. Service accounts can be shared amongst multiple people, so you don't need to re-deploy your entire infrastructure to grant a new employee permissions. You just grant them the ability to mint tokens as a service account
4. Creating tokens from a service account creates audit logs you can use to verify whether you've been attacked
5. Machines run as service accounts. If you wanted to authenticate as a user on a backend you'd have to give that backend scary permissions. Enough permissions to read your email.

If you really want bob@mydomain.com to be able to invoke a function, I recommend you just grant bob@mydomain.com the `roles/functions.invoker` role across your project. If you're trying to add the sophistication of per-function ACLs, then you should be using service accounts.

PR #3790, opened 2021-09-28
Start inferring wantBackend details from existingBackend

inlined:
Creating a new point in our deploy process where we can modify the desired backend to fill in details from the backend we already have. For example, we can preserve existing environment variables when previews.dotenv is off or we can remember a trigger region from the previous deploy.

This simplifies our existing codebase to no longer need to pass overwriteExistingEnvs to the release planner. I also noticed that this is the third time I've wanted to code `backendOf` in a unit test, so I created `backend.of`

### DIFF
--- a/src/test/deploy/functions/prepare.spec.ts
+++ b/src/test/deploy/functions/prepare.spec.ts
@@ -30,7 +30,7 @@ describe("prepare", () => {
         },
       };
 
-      prepare.inferDetailsFromExisting(backend.of(newE), backend.of(oldE), /* usedDotEnv= */ false);
+      prepare.inferDetailsFromExisting(backend.of(newE), backend.of(oldE), /* usedDotenv= */ false);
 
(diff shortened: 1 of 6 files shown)


src/test/deploy/functions/prepare.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+      };
+      const newE = {
+        ...ENDPOINT,
+        environmentVariables: {
+          foo: "new value",
+          new: "value",
+        },
+      };
+
+      prepare.inferDetailsFromExisting(backend.of(newE), backend.of(oldE), /* usedDotEnv= */ false);
```
taeold, line 33:
```suggestion
      prepare.inferDetailsFromExisting(backend.of(newE), backend.of(oldE), /* usedDotenv= */ false);
```

taeold (approved):
Love it - so much cleaner.

PR #3795, opened 2021-10-02
Update all other endpoints that are impacted by multi-tenancy

lisajian:
### Description

Any endpoints that use `tenantId` explicitly in the request path or in the request body are updated to account for `Tenant` settings and any other tenant specific logic in this PR. Some things to highlight

- Providers are disabled by default on tenant projects based on observable behavior on Cloud dashboard. Agent projects have settings default enabled to be consistent with existing behavior until project `Config` management is introduced
- Supported tenant operations are listed at go/firebase-tools-3795-tenantOps; all other endpoints throw errors when called with a `TenantProjectState`. Some endpoints are not explicitly listed but are supported on tenant projects including:
  - `getRecaptchaParams`
  - `createSessionCookie`
  - `grantToken`
- Some, but not all, emulator endpoints were modified to take `tenantId` as a query param to correctly retrieve `TenantProjectState`. Not all emulator endpoints were updated but should be - a `TODO` was added in `gen-auth-api-spec.ts` that will be addressed in a subsequent PR (this one was getting bloated)

**Callouts:**
- `apiSpec.js` was not regenerated with the command `npm run generate:auth-api` so far. Regenerating `apiSpec.js` appears to break several tests and will be investigated separately (b/201833004)
- Noticed that no tests currently exist for `getProjects` and `getRecaptchaParams` - not sure if this was intentionally left out, clarification on this front would be appreciated!

### Scenarios Tested

`npm test` passes

**See above for callouts**

### Sample Commands

N/A

### DIFF
--- a/src/emulator/auth/operations.ts
+++ b/src/emulator/auth/operations.ts
@@ -1,5 +1,5 @@
 import { URLSearchParams } from "url";
-import { decode as decodeJwt, sign as signJwt, JwtHeader, decode } from "jsonwebtoken";
+import { decode as decodeJwt, sign as signJwt, JwtHeader } from "jsonwebtoken";
 import * as express from "express";
 import { ExegesisContext } from "exegesis-express";
 import {
@@ -1054,10 +1054,6 @@ export function setAccountInfoImpl(
 
       const newEmail = canonicalizeEmailAddress(reqBody.email);
       if (newEmail !== user.email) {
-        assert(
-          state.mfaConfig.state === "ENABLED" || state.mfaConfig.state === "MANDATORY",
-          "EMAIL_CHANGE_NEEDS_VERIFICATION"
-        );
         assert(!state.getUserByEmail(newEmail), "EMAIL_EXISTS");
         updates.email = newEmail;
         // TODO: Set verified if email is verified by IDP linked to account.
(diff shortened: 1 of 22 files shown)


src/emulator/auth/operations.ts
```diff
@@ -658,7 +689,10 @@ function getProjects(
   };
 }
 
-function getRecaptchaParams(): Schemas["GoogleCloudIdentitytoolkitV1GetRecaptchaParamResponse"] {
+function getRecaptchaParams(
```
lisajian, line 692: Noticed that no tests currently exist for `getProjects` and `getRecaptchaParams` - not sure if this was intentionally left out, clarification on this front would be appreciated!

cc: @yuchenshi

sam-gc (approved):
Looks good (very thorough tests!) but please also make sure Yuchen has a chance to look through---I only have a high-level understanding of this stuff

src/emulator/auth/state.ts
```diff
@@ -669,6 +705,26 @@ export class TenantProjectState extends ProjectState {
     return this._tenantConfig;
   }
 
+  get allowPasswordSignup() {
```
lisajian, line 708: Discussed offline during emulator sync - all these should be default enabled. Will include in next revision

yuchenshi (commented):
Mostly LGTM except tenant auto-creation as discussed

lisajian:
> Mostly LGTM except tenant auto-creation as discussed

Yep! Auto-creation is here: https://github.com/firebase/firebase-tools/pull/3777/files#r722731660

yuchenshi (approved):

PR #3798, opened 2021-10-04
Implements `deploy` and `release` stage of `firebase deploy --only extensions`

joehan:
### Description
Finishes implementing `firebase deploy --only extensions`

### Scenarios Tested
Screen Shot 2021-10-01 at 1 37 30 PM
Screen Shot 2021-10-01 at 3 20 21 PM
Screen Shot 2021-10-01 at 3 28 35 PM

Our error messages need some work on our backend - we need to start wrapping validation errors in public errors so that the details can be seen clientside. Planning to work on that this week.
Screen Shot 2021-10-01 at 4 05 11 PM

### Sample Commands
- Add some `extensions` to firebase.json:
```
{
  "extensions": {
    "my-ext-id": "firebase/storage-resize-images@^0.1.0",
    "storage-resize-imagesy" : "firebase/storage-resize-images@^0.1.0",
    "storage-resize-imagesz" : "firebase/storage-resize-images@^0.1.0"
  }
}
```
- Create some .env files  in the `extensions` directory to configure them - for example:
```
## In myProject/extensions/my-ext-id.env
LOCATION=us-central1
IMG_BUCKET=joehanley-public.appspot.com
IMG_SIZES=100x100
DELETE_ORIGINAL_FILE=false
IMAGE_TYPE=jpeg,png
```
- `firebase deploy --only extensions`


src/deploy/extensions/tasks.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+export function extensionsDeploymentHandler(
+  errorHandler: ErrorHandler
+): (task: ExtensionDeploymentTask) => Promise<any | undefined> {
+  return async (task: ExtensionDeploymentTask) => {
+    let result;
+    try {
+      result = await task.run();
+    } catch (err) {
+      if (err.status == 429) {
```
huangjeff5, line 24: Wondering why 429 specifically is being handled? Is there some reason this is likely to happen? May be worth documenting.

src/extensions/extensionsApi.ts
```diff
@@ -330,12 +347,14 @@ export async function configureInstance(
...
  * @param params params to configure the extension instance
+ * @param validateOnly if true, only validates the update and makes no changes
  */
 export async function updateInstance(
   projectId: string,
   instanceId: string,
   extensionSource: ExtensionSource,
-  params?: { [option: string]: string }
+  params?: { [option: string]: string },
+  validateOnly: boolean = false
```
huangjeff5, line 357: Wonder if we should make this signature an interface? The number of arguments for {create|update|delete|configure}Instance seems to be growing.

src/deploy/extensions/tasks.ts
joehan: Refactored this a bit to be more clear. 429 (and, as I'm realizing now, 409) errors are retryable errors, so we should retry then with backoff. This handler function is used with queue.ts, which retries when the handler throws an error.

For all other error codes, we don't want to retry, so we instead record and return.

joehan: In a larger sense - I'm a bit concerned about hitting Cloud Build quotas when deploying a lot of extensions at once. I'm planning on making some backend changes to make sure that we pipe through 429's that we get from DM, so that we would retry them here

PR #3800, opened 2021-10-04
Enhance function deployment

colerogers:
### Description

Fixes deployment for some projects that are not part of a GCP allowlist.

### Scenarios Tested

- Deploy from a project that is not on the allowlist

### DIFF
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -431,7 +431,7 @@ async function loadExistingBackend(ctx: Context & PrivateContextFields): Promise
   try {
     gcfV2Results = await gcfV2.listAllFunctions(ctx.projectId);
   } catch (err) {
-    if (err.status === 404 && err.message?.includes("Method not found")) {
+    if (err.status === 404 && err.message?.toLowerCase().includes("method not found")) {
       return; // customer has preview enabled without allowlist set
     }
     throw err;
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -427,7 +427,15 @@ async function loadExistingBackend(ctx: Context & PrivateContextFields): Promise
     return;
   }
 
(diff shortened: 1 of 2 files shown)


src/deploy/functions/backend.ts
```diff
@@ -427,7 +427,15 @@ async function loadExistingBackend(ctx: Context & PrivateContextFields): Promise
     return;
   }
 
-  const gcfV2Results = await gcfV2.listAllFunctions(ctx.projectId);
+  let gcfV2Results;
+  try {
+    gcfV2Results = await gcfV2.listAllFunctions(ctx.projectId);
+  } catch (err) {
+    if (err.status === 404 && err.message?.toLowerCase().includes("method not found")) {
```
colerogers, line 434: I thought this is the best place to make the check since the gcfV2 package shouldn't care what error is thrown.

taeold: Isn't there another error thrown below (in `gcfV2Results.functions`) since `gcfV2Results` will be `undefined` in this case?

taeold: (It obviously doesn't given the test cases, but I'm curious how it works 🤔 )

colerogers: I don't think I follow. If `gcfV2.listAllFunctions` throws an error we either return from the method or throw the error to the next method. Otherwise `gcfV2Results` is set to the internal `gcfV2` type of `ListFunctionsResponse` which has the `functions` member. 

I can initialize `gcfV2Results` to something like:
```
let gcfV2Results = {
    functions: [],
    unreachable: [],
  };
```
But I'm not sure that would really do much since TypeScript implicitly figures out this - `gcfV2Results: ListFunctionsResponse`

taeold: Ah my mistake - I read `return` and thought `continue` 🤦 .

taeold (approved):

PR #3806, opened 2021-10-06
Make emulator monkey-patching compatible with newer versions of firebase-functions SDK.

taeold:
We will soon land a (backward-compatible) change in the Firebase Functions SDK that allows `onCall` functions to accept optional configuration argument, e.g.

```js

export const helloWorld = functions
    .https
    .onCall({allowInvalidAppCheckToken: true}, (data, context) => {
      return `Hello ${data.name}`;
    });
```

The existing monkey-patching will not correctly patch the new signature as is. The proposed change sniffs the number of arguments expected by the version of Firebase Functions SDK being used and monkey-patches  accordingly.

### DIFF
--- a/src/emulator/functionsEmulatorRuntime.ts
+++ b/src/emulator/functionsEmulatorRuntime.ts
@@ -448,14 +448,34 @@ async function initializeFirebaseFunctionsStubs(frb: FunctionsRuntimeBundle): Pr
   const onCallInnerMethodName = "_onCallWithOptions";
   const onCallMethodOriginal = httpsProvider[onCallInnerMethodName];
 
-  httpsProvider[onCallInnerMethodName] = (handler: CallableHandler, opts: DeploymentOptions) => {
-    const wrapped = wrapCallableHandler(handler);
-    const cf = onCallMethodOriginal(wrapped, opts);
-    return cf;
-  };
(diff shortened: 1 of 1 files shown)


src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -448,14 +448,35 @@ async function initializeFirebaseFunctionsStubs(frb: FunctionsRuntimeBundle): Pr
...
+    httpsProvider[onCallInnerMethodName] = (handler: any, opts: DeploymentOptions) => {
+      const wrapped = wrapCallableHandler(handler);
+      const cf = onCallMethodOriginal(wrapped, opts);
+      return cf;
+    };
+  }
 
-  httpsProvider.onCall = (handler: CallableHandler) => {
-    return httpsProvider[onCallInnerMethodName](handler, {});
+  // Newer versions of the firebase-functions package's onCAll method can accept upto 2 arguments.
```
yuchenshi, line 470:
```suggestion
  // Older versions of the firebase-functions package's onCall method can accept upto 2 arguments.
```

yuchenshi: Make sense to me now. Sorry for the confusion earlier.

inlined:
Probably should have waited until the other PR was approved before landing this one =p

PR #3807, opened 2021-10-06
Adds Fabricator to release code

inlined:
Fabricator is the repalcement for the previous Tasks codebase, though it will not handle the portion of tracking or error reporting internally (those will be future utilities for better testing).

Some differences between tasks.ts and fabricator.ts:
* Queues are used for narrow transactions, which means we don't accidentally have an error bubble up and cause a retry on a bulk operation. There's a design pattern of executor.run.catch(rethrowAs)
* Function triggers (not EventArc triggers, which are deployed as part of the function, but external triggers like topics, and in the future intents or task queues) are only created after a successful function creation and functions are only deleted after a successful trigger deletion
* We do not delete functions if an upsert failed (this helps catch cases where a user renamed a function)
* This code is actually unit tested (hooray!)
* It handles v1 and v2 functions completely separately. IMO this dramatically improved readability by being less clever.

Personal apologies to @joehan that I'm not sure it was as respectful as I could have been to start from scratch. I tried on the from-scratch implementation for a while and ended up feeling like I was rewriting the original tasks.ts. Then I started editing tasks.ts and felt I was coming back to this design. I resumed working on the from-scratch implementation just because there were a few lower-level details I liked more about the fresh approach and I was already pretty far along.

### DIFF
--- a/src/deploy/functions/release/fabricator.ts
+++ b/src/deploy/functions/release/fabricator.ts
@@ -1,0 +1,463 @@
+import * as clc from "cli-color";
+
+import { FirebaseError } from "../../../error";
+import { functionsOrigin, functionsV2Origin } from "../../../api";
+import * as utils from "../../../utils";
+import { assertExhaustive } from "../../../functional";
+import * as planner from "./planner";
+import * as backend from "../backend";
+import * as helper from "../functionsDeployHelper";
+import * as gcf from "../../../gcp/cloudfunctions";
+import * as gcfV2 from "../../../gcp/cloudfunctionsv2";
+import * as run from "../../../gcp/run";
+import * as proto from "../../../gcp/proto";
+import * as pubsub from "../../../gcp/pubsub";
+import * as scheduler from "../../../gcp/cloudscheduler";
+import { getHumanFriendlyRuntimeName } from "../runtimes";
+import { SourceTokenScraper } from "./sourceTokenScraper";
+import * as poller from "../../../operation-poller";
+import { logger } from "../../../logger";
+import { Executor } from "./executor";
+import { Timer } from "./timer";
+import * as reporter from "./reporter";
+
+// TODO: Tune this for better performance.
+const gcfV1PollerOptions = {
+  apiOrigin: functionsOrigin,
+  apiVersion: gcf.API_VERSION,
+  masterTimeout: 25 * 60 * 1000, // 25 minutes is the maximum build time for a function
+};
+
+const gcfV2PollerOptions = {
+  apiOrigin: functionsV2Origin,
+  apiVersion: gcfV2.API_VERSION,
+  masterTimeout: 25 * 60 * 1000, // 25 minutes is the maximum build time for a function
+};
+
+export interface FabricatorArgs {
+  executor: Executor;
+  functionExecutor: Executor;
+  sourceUrl: string;
+  storage: Record<string, gcfV2.StorageSource>;
+  appEngineLocation: string;
+}
+
+const rethrowAs = (endpoint: backend.Endpoint, op: reporter.OperationType) => (
+  err: unknown
+): void => {
+  throw new reporter.DeploymentError(endpoint, op, err);
+};
+
+/** Fabricators make a customer's backend match a spec by applying a plan. */
+export class Fabricator {
+  executor!: Executor;
+  functionExecutor!: Executor;
+  sourceUrl!: string;
+  storage!: Record<string, gcfV2.StorageSource>;
+  appEngineLocation!: string;
+
+  constructor(args: FabricatorArgs) {
+    proto.copyIfPresent(
+      this,
+      args,
+      "executor",
+      "functionExecutor",
+      "sourceUrl",
+      "storage",
+      "appEngineLocation"
+    );
+  }
+
+  async applyPlan(plan: planner.DeploymentPlan): Promise<reporter.DeployResults> {
+    const deployRegions = Object.values(plan).map((regional) =>
+      this.applyRegionalChanges(regional)
+    );
+    const promiseResults = await utils.allSettled(deployRegions);
+
+    const errs = promiseResults
+      .filter((r) => r.status === "rejected")
+      .map((r) => (r as utils.PromiseRejectedResult).reason);
+    if (errs.length) {
+      logger.debug(
+        "Fabricator.applyRegionalChanges returned an unhandled exception. This should never happen",
+        JSON.stringify(errs, null, 2)
+      );
+    }
+
+    const deployResults: reporter.DeployResults = {};
+    const regionalResults = promiseResults
+      .filter((r) => r.status === "fulfilled")
+      .map<reporter.DeployResults>((r) => (r as utils.PromiseFulfilledResult).value);
+
+    for (const regionalResult of regionalResults) {
+      for (const [region, mapping] of Object.entries(regionalResult)) {
+        // Should always be empty...
+        deployResults[region] = deployResults[region] || {};
+        for (const [id, result] of Object.entries(mapping)) {
+          if (deployResults[region][id]) {
+            throw new Error(`Unexpected duplicate results for function ${id} in region ${region}`);
+          }
+          deployResults[region][id] = { ...result };
+        }
+      }
+    }
+    return deployResults;
+  }
+
+  async applyRegionalChanges(changes: planner.RegionalChanges): Promise<reporter.DeployResults> {
+    const deployResults: reporter.DeployResults = {};
+    let hasErrors = false;
+    const handle = async (
+      op: reporter.OperationType,
+      endpoint: backend.Endpoint,
+      fn: () => Promise<void>
+    ): Promise<void> => {
+      const timer = new Timer();
+      const result: reporter.DeployResult = { durationMs: 0 };
+      try {
+        await fn();
+        this.logOpSuccess(op, endpoint);
+      } catch (err) {
+        hasErrors = true;
+        result.error = err as Error;
+      }
+      result.durationMs = timer.stop();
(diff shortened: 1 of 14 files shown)


src/deploy/functions/release/fabricator.ts
```diff
@@ -0,0 +1,463 @@
...
+        .run(async () => {
+          try {
+            await pubsub.getTopic(topic);
+          } catch (err) {
+            if (err.status !== 404) {
+              throw new FirebaseError("Unexpected error looking for Pub/Sub topic", {
+                original: err as Error,
+              });
+            }
+            await pubsub.createTopic({ name: topic });
```
joehan, line 247: Optional - consider a blind create here?

src/deploy/functions/release/fabricator.ts
```diff
@@ -0,0 +1,463 @@
...
+      await this.executor
+        .run(async () => {
+          try {
+            await pubsub.getTopic(topic);
+          } catch (err) {
+            if (err.status !== 404) {
+              throw new FirebaseError("Unexpected error looking for Pub/Sub topic", {
+                original: err as Error,
+              });
+            }
```
joehan, line 246: Missing a return here - this looks like it still tries to create when we get a 404 error?

src/deploy/functions/release/fabricator.ts
```diff
@@ -0,0 +1,463 @@
...
+    if (backend.isHttpsTriggered(endpoint)) {
+      const invoker = endpoint.httpsTrigger.invoker || ["public"];
+      if (invoker[0] !== "private") {
+        await this.executor
+          .run(() => run.setInvokerCreate(endpoint.project, serviceName, invoker))
+          .catch(rethrowAs(endpoint, "set invoker"));
+      }
+    }
+
+    await this.setConcurrency(endpoint, serviceName, endpoint.concurrency || 80);
```
joehan, line 274: Move this magic number to a named constant.

src/deploy/functions/release/fabricator.ts
```diff
@@ -0,0 +1,463 @@
...
+
+        delete service.status;
+        delete (service.spec.template.metadata as any).name;
+        service.spec.template.spec.containerConcurrency = concurrency;
+        await run.replaceService(serviceName, service);
+      })
+      .catch(rethrowAs(endpoint, "set concurrency"));
+  }
+
+  async setTrigger(endpoint: backend.Endpoint): Promise<void> {
```
joehan, line 394: Is the plan here to add the work we need for future 'special' trigger types like task queues to this function? If so, I very much like the idea, but a brief comment explaining it would be useful.

joehan (commented):
Wooof, this was a big one! Mostly looks good to me , and no need to apologize for reworking the code. I'd definitely suggest doing a good amount of manual testing here - you added some very thorough unit tests, but this code path is so complicated that its really easy to miss random edge cases.

Basically approved - I'm jsut leaving it unapproved for now to make sure I come back and rereview after any changes

src/deploy/functions/release/timer.ts
```diff
@@ -0,0 +1,13 @@
+/** Measures the time taken from construction to the call to stop() */
+export class Timer {
+  private readonly start: [number, number];
+
+  constructor() {
+    this.start = process.hrtime();
```
taeold, line 6: Nit: prefer `process.hrtime().bigint()` per https://nodejs.org/api/process.html#process_process_hrtime_time.

src/gcp/cloudscheduler.ts
```diff
@@ -212,3 +213,36 @@ export function jobFromSpec(schedule: backend.ScheduleSpec, appEngineLocation: s
...
+
+/** Converts an Endpoint to a CloudScheduler v1 job */
+export function jobFromEndpoint(
+  endpoint: backend.Endpoint & backend.ScheduleTriggered,
+  appEngineLocation: string
+): Job {
+  const job: Partial<Job> = {};
+  if (endpoint.platform === "gcfv1") {
+    const id = backend.scheduleIdForFunction(endpoint);
+    const region = appEngineLocation;
```
taeold, line 225: q: Is it possible and okay for `endpoint.region` to be different than `appEngineLocation`?

src/deploy/functions/release/sourceTokenScraper.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  // Token Promise will return undefined for the first caller
+  // (because we presume it's this function's source token we'll scrape)
+  // and then returns the promise generated from the first function's onCall
+  tokenPromise(): Promise<string | undefined> {
+    if (this.firstCall) {
+      this.firstCall = false;
+      return Promise.resolve(undefined);
+    }
```
taeold, line 24: Is it possible to have race condition here?

src/deploy/functions/release/reporter.ts
```diff
@@ -0,0 +1,37 @@
...
+  | "create topic"
+  | "delete topic"
+  | "set invoker"
+  | "set concurrency";
+
+export class DeploymentError extends Error {
+  constructor(
+    readonly endpoint: backend.Endpoint,
+    readonly op: OperationType,
+    readonly original: unknown
```
taeold, line 27: Do we do anything with the original error?

src/deploy/functions/release/sourceTokenScraper.ts
taeold: nvm this is JS.

src/deploy/functions/release/fabricator.ts
```diff
@@ -0,0 +1,463 @@
...
+    }
+  }
+
+  async updateV2Function(endpoint: backend.Endpoint): Promise<void> {
+    const apiFunction = gcfV2.functionFromEndpoint(endpoint, this.storage[endpoint.region]);
+
+    // N.B. As of GCFv2 private preview the API chokes on any update call that
+    // includes the pub/sub topic even if that topic is unchanged.
+    // We know that the user hasn't changed the topic between deploys because
+    // of checkForInvalidChangeOfTrigger().
```
taeold, line 312: quriky...

src/test/deploy/functions/release/fabricator.spec.ts
```diff
@@ -0,0 +1,972 @@
...
+    run.getIamPolicy.rejects(new Error("unexpected run.getIamPolicy"));
+    run.setIamPolicy.rejects(new Error("unexpected run.setIamPolicy"));
+    run.setInvokerCreate.rejects(new Error("unexpected run.setInvokerCreate"));
+    run.setInvokerUpdate.rejects(new Error("unexpected run.setInvokerUpdate"));
+    run.replaceService.rejects(new Error("unexpected run.replaceService"));
+    poller.pollOperation.rejects(new Error("unexpected poller.pollOperation"));
+    pubsub.createTopic.rejects(new Error("unexpected pubsub.createTopic"));
+    pubsub.deleteTopic.rejects(new Error("unexpected pubsub.deleteTopic"));
+    scheduler.createOrReplaceJob.rejects(new Error("unexpected scheduler.createOrReplaceJob"));
+    scheduler.deleteJob.rejects(new Error("unexpected scheduler.deleteJob"));
```
taeold, line 56: holy moly

taeold (approved):
Very educational as always 👍

src/deploy/functions/release/sourceTokenScraper.ts
```diff
@@ -0,0 +1,39 @@
...
+    return this.promise;
+  }
+
+  get poller() {
+    return (op: any) => {
+      if (op.metadata?.sourceToken || op.done) {
```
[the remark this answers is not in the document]
inlined: Actually it's the opposite. We get a sourceToken after the Cloud Build phase, well before op.done. The op.done bit is as a safety net to make sure we don't block forever if GCF doesn't give us a sourceToken for some reason.

src/deploy/functions/release/fabricator.ts
```diff
@@ -0,0 +1,463 @@
...
+  async setTrigger(endpoint: backend.Endpoint): Promise<void> {
+    if (backend.isScheduleTriggered(endpoint)) {
+      if (endpoint.platform === "gcfv1") {
+        await this.upsertScheduleV1(endpoint);
+        return;
+      } else if (endpoint.platform === "gcfv2") {
+        await this.upsertScheduleV2(endpoint);
+        return;
+      }
+      assertExhaustive(endpoint.platform);
```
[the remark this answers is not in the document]
inlined: Because of the return statements. If platform is `gcfv1` or `gcfv2` then the LOC is unreachable. only `never` could reach that LOC.

src/deploy/functions/release/fabricator.ts
```diff
@@ -0,0 +1,463 @@
...
+      .catch(rethrowAs(endpoint, "delete topic"));
+  }
+
+  deleteScheduleV2(endpoint: backend.Endpoint & backend.ScheduleTriggered): Promise<void> {
+    return Promise.reject(
+      new reporter.DeploymentError(endpoint, "delete schedule", new Error("Not implemented"))
+    );
+  }
+
+  logOpStart(op: string, endpoint: backend.Endpoint): void {
```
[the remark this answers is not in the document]
inlined: Sadly it can't. The OperationType is "create" but the log is "creating"

src/test/deploy/functions/release/fabricator.spec.ts
inlined: Admittedly a bit of a PITA, but not as much of a PITA as you'd expect  with VIM bindings:

1. Use code complete to make a line for each export
2. Use Shift + V to enter visual mode and highlight all lines
3. Use a find & replace: `%s/ +(.*)/\1.rejects(new Error("unexpected \1"));`

PR #3809, opened 2021-10-07
storage trigger works

colerogers:
### Description

Storage Emulator Enhancement

### DIFF
--- a/src/emulator/storage/cloudFunctions.ts
+++ b/src/emulator/storage/cloudFunctions.ts
@@ -4,8 +4,15 @@ import * as request from "request";
 import { EmulatorLogger } from "../emulatorLogger";
 import { CloudStorageObjectMetadata, toSerializedDate } from "./metadata";
 import { Client } from "../../apiv2";
+import { SignatureType } from "../functionsEmulatorShared";
 
 type StorageCloudFunctionAction = "finalize" | "metadataUpdate" | "delete" | "archive";
+const STORAGE_ACTION_MAP: Record<string, string> = {
+  finalize: "finalized",
+  metadataUpdate: "metadataUpdated",
+  delete: "deleted",
+  archive: "archived",
+};
 
 export class StorageCloudFunctions {
   private logger = EmulatorLogger.forEmulator(Emulators.STORAGE);
@@ -13,6 +20,7 @@ export class StorageCloudFunctions {
   private multicastOrigin = "";
   private multicastPath = "";
   private enabled = false;
+  private client: Client = new Client({ urlPrefix: "" });
 
   constructor(private projectId: string) {
     const functionsEmulator = EmulatorRegistry.get(Emulators.FUNCTIONS);
@@ -24,6 +32,7 @@ export class StorageCloudFunctions {
         this.functionsEmulatorInfo
       )}`;
       this.multicastPath = `/functions/projects/${projectId}/trigger_multicast`;
+      this.client = new Client({ urlPrefix: this.multicastOrigin, auth: false });
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/emulator/storage/cloudFunctions.ts
```diff
@@ -69,6 +85,22 @@ export class StorageCloudFunctions {
       }, // bucket
       data: objectMetadataPayload,
     });
+    /** Cloud Event */
+    const ceAction = STORAGE_V2_ACTION_MAP[action];
```
taeold, line 89: I think we have opportunity to make this a bit safer - either have a runtime check to make sure that ceAction isn't undefined or have a type-level checking to make sure that `STORAGE_V2_ACTION_MAP` mapping contains all the mapping for possible values of `StorageCloudFunctionAction`

src/emulator/storage/cloudFunctions.ts
```diff
@@ -69,6 +85,22 @@ export class StorageCloudFunctions {
...
+    const ce = {
+      specVersion: 1,
+      type: `google.cloud.storage.object.v1.${ceAction}`,
+      source: `//storage.googleapis.com/projects/_/buckets/${objectMetadataPayload.bucket}/objects/${objectMetadataPayload.name}`,
+      data,
+    };
+    const cloudEvent: string = JSON.stringify({
+      eventType: `google.cloud.storage.object.v1.${ceAction}`, // need eventType for multicast triggerId
+      data: ce,
+    });
```
taeold, line 101: Huh I'm surprised that this works since I expected the body of the cloudevent http call to look like:

```js
{
  specVersion: 1,
  type: "google.cloud.storage.object.v1..."
  source: "..."
  data: { ... }
}
```
vs

```js
{
  data: {
    specVersion: 1,
    type: "google.cloud.storage.object.v1..."
    source: "..."
    data: { ... }
  }
}
```

Something must be unboxing the nested `data` property before sending it off to the function? Can you help me step through how this works?

src/emulator/storage/cloudFunctions.ts
```diff
@@ -69,6 +85,22 @@ export class StorageCloudFunctions {
...
+    const data = { ...objectMetadataPayload };
+    delete (data as any).acl; // remove this field from cloud events
+    const ce = {
+      specVersion: 1,
+      type: `google.cloud.storage.object.v1.${ceAction}`,
+      source: `//storage.googleapis.com/projects/_/buckets/${objectMetadataPayload.bucket}/objects/${objectMetadataPayload.name}`,
+      data,
+    };
+    const cloudEvent: string = JSON.stringify({
+      eventType: `google.cloud.storage.object.v1.${ceAction}`, // need eventType for multicast triggerId
```
taeold, line 99: Hmm this implementation feels a bit brittle since `eventType` isn't a thing in a cloudevent payload. It also makes  an implicit assumption that the name of the event will always be different between legacy and cloudevent. That is actually, probably true, but it's a subtle assumption that's worth being explicit (either in code or in implementation).

Another way to implement this change would be to teach the multicast handler to understand the difference between cloudevent and legacy event payloads (either by having different endpoint, having a wrapper request body e.g. `{ eventFormat: "cloudevent", eventBody: body }`, etc. Do you think that is going to work?

src/emulator/storage/cloudFunctions.ts
```diff
@@ -1,18 +1,24 @@
...
+  archive: "archived",
+};
 
 export class StorageCloudFunctions {
   private logger = EmulatorLogger.forEmulator(Emulators.STORAGE);
   private functionsEmulatorInfo?: EmulatorInfo;
   private multicastOrigin = "";
   private multicastPath = "";
   private enabled = false;
+  private client: Client = new Client({ urlPrefix: "" });
```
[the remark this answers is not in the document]
colerogers: Since we only initialize in the constructor when `functionsEmulator` is set, we need to have `client` be be initialized to some default value (can't be null/undefined) so this seemed like a reasonable place to set a default value

src/emulator/storage/cloudFunctions.ts
```diff
@@ -53,12 +67,14 @@ export class StorageCloudFunctions {
     }
   }
 
-  private createEventRequestBody(
+  /** Creates a legacy event and a CloudEvent body */
+  private createEventRequestBodies(
```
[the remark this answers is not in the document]
colerogers: broke it up into 2 functions

src/emulator/storage/cloudFunctions.ts
taeold: Discussed offline - we will probably use the value of the `Content-Type` header to differentiate cloudevent vs legacy payload.

Also, I think this already works w/o any modification, but I'd make sure that trigger registration where we construct the key for the `multicastTriggers` object still works:

https://github.com/firebase/firebase-tools/blob/e4b9b938716b6679e132a13a82fd7f16bdda5804/src/emulator/functionsEmulator.ts#L695-L703

taeold (approved):
Overall LGTM. You may want to think about adding some test cases in https://github.com/firebase/firebase-tools/tree/master/scripts/triggers-end-to-end-tests for the new storage triggers (old legacy style and new cloudevent style if you are willing to contribute some time).

PR #3811, opened 2021-10-08
Implementing some UX feedback for `ext:export `and `deploy --only extensions`

joehan:
### Description
Implementing some UX feedback from Will and Sonakshi, and cleaning up some ugly bits that made it through the last PR.

Specifically:
- During ext:export, I added numbered bullets when listing out the extensions to save locally, to help anchor users when there are a lot of extension instances.
- During deploy, I separated out configures and updates to match our messaging elsewhere.
- During deploy, I listed the instances to be delete in the prompt, to clarify exactly what will be deleted
- 
### Scenarios Tested
A deploy containing creates, configures, updates, and deletes
Screen Shot 2021-10-08 at 4 15 57 PM

ext:export:
Screen Shot 2021-10-08 at 2 12 43 PM
Screen Shot 2021-10-08 at 2 12 52 PM

### DIFF
--- a/src/deploy/extensions/deploymentSummary.ts
+++ b/src/deploy/extensions/deploymentSummary.ts
@@ -1,10 +1,8 @@
-import * as clc from "cli-color";
-
 import * as planner from "./planner";
 import * as refs from "../../extensions/refs";
 
 export const humanReadable = (dep: planner.InstanceSpec) =>
-  `${clc.bold(dep.instanceId)} (${
+  `${dep.instanceId} (${
     dep.ref ? `${refs.toExtensionVersionRef(dep.ref)}` : `Installed from local source`
   })`;
 
@@ -13,14 +11,12 @@ const humanReadableUpdate = (from: planner.InstanceSpec, to: planner.InstanceSpe
     from.ref?.publisherId == to.ref?.publisherId &&
     from.ref?.extensionId == to.ref?.extensionId
   ) {
-    return `\t${clc.bold(from.instanceId)} (${refs.toExtensionVersionRef(from.ref!)} => ${
-      to.ref?.version
-    })`;
+    return `\t${from.instanceId} (${refs.toExtensionVersionRef(from.ref!)} => ${to.ref?.version})`;
   } else {
     const fromRef = from.ref
       ? `${refs.toExtensionVersionRef(from.ref)}`
       : `Installed from local source`;
-    return `\t${clc.bold(from.instanceId)} (${fromRef} => ${refs.toExtensionVersionRef(to.ref!)})`;
+    return `\t${from.instanceId} (${fromRef} => ${refs.toExtensionVersionRef(to.ref!)})`;
   }
 };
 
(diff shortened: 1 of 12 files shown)


src/deploy/extensions/deploymentSummary.ts
```diff
@@ -45,6 +45,14 @@ export function updatesSummary(
   return toUpdate.length ? `The following extension instances will be updated:\n${summary}\n` : "";
 }
 
+export function configuresSummary(toConfigure: planner.InstanceSpec[]) {
+  return toConfigure.length
+    ? `The following extension instances will be configured:\n${toConfigure
+        .map((s) => `\t${humanReadable(s)}`)
+        .join("\n")}\n`
```
elvisun, line 52: a bit message, pull the inlined part into a variable?

src/deploy/extensions/prepare.ts
```diff
@@ -57,7 +65,14 @@ export async function prepare(
 
   await requirePermissions(options, permissionsNeeded);
 }
+const matchesInstanceId = (dep: planner.InstanceSpec) => (test: planner.InstanceSpec) => {
+  return dep.instanceId == test.instanceId;
```
elvisun, line 69: `===`?

src/deploy/extensions/prepare.ts
```diff
@@ -57,7 +65,14 @@ export async function prepare(
 
   await requirePermissions(options, permissionsNeeded);
 }
+const matchesInstanceId = (dep: planner.InstanceSpec) => (test: planner.InstanceSpec) => {
```
elvisun, line 68: can these be just functions?
go/tsstyle#function-declarations

src/extensions/export.ts
```diff
@@ -25,13 +24,16 @@ export function displayExportInfo(withRef: InstanceSpec[], withoutRef: InstanceS
   }
 }
 
-function displayParams(spec: InstanceSpec): void {
-  logger.info(humanReadable(spec));
-  logger.info(`Configuration will be written to 'extensions/${spec.instanceId}.env`);
-  for (const p of Object.entries(spec.params)) {
-    logger.info(`\t${p[0]}: ${p[1]}`);
+function displaySpecs(specs: InstanceSpec[]): void {
```
elvisun, line 27: Kind of hard to understand what this function is doing, can you add an example to the comment?

src/extensions/extensionsApi.ts
```diff
@@ -320,20 +320,20 @@ export async function listInstances(projectId: string): Promise<ExtensionInstanc
...
-  projectId: string,
-  instanceId: string,
-  params: { [option: string]: string },
-  validateOnly: boolean = false
-): Promise<any> {
+export async function configureInstance(args: {
+  projectId: string;
+  instanceId: string;
+  params: { [option: string]: string };
+  validateOnly?: boolean;
```
elvisun, line 327: nit: boolean > boolean|null 

consider giving this a default of `false` instead

src/extensions/refs.ts
```diff
@@ -90,3 +90,16 @@ export function toExtensionVersionName(ref: Ref): string {
   }
   return `publishers/${ref.publisherId}/extensions/${ref.extensionId}/versions/${ref.version}`;
 }
+
+/**
+ * Checks if two refs refer to the same extensionVersion.
+ */
+export function equal(a?: Ref, b?: Ref): boolean {
+  return (
+    !!a &&
```
elvisun, line 99: !! can be dropped

src/deploy/extensions/deploymentSummary.ts
joehan: Done, and also did the same for the other summary methods.

src/extensions/extensionsApi.ts
joehan: I wanted to, but unfortunately you can't use default values inside of an args object like this. I figured that it was preferable to have named params here instead of a little extra type safety

PR #3813, opened 2021-10-10
Add description to firebase.json schema

johnnyasantoss:
### Description

Adds description to firebase.json schema

### Sample Commands

Add `generate:json-schema:ordered` command, which makes the same thing as the previous one but also sorts the json keys.


johnnyasantoss:
I was trying to rebase, but it seems that all my changes have been implemented already.

PR #3816, opened 2021-10-11
Merge master; backport recent changes

inlined:
Merges master + corrections and backports.

1. Applied @colerogers's hack where we detect "region" for GCS functions and make it "bucket" to the Endpoints version of parseTriggers.ts
2. Moved the logic where we get trigger regions from previous deploys to the new `inferDetailsFromExisting`
3. Renamed the function for looking up missing trigger names to clarify that it only affects missing names.

While I was looking at the codebase for (3) again, I rewrote it quickly.
1. We need to be making these API calls in parallel. This could have massive implications on first deploy speed.
2. This common idiom for this kind of code is a [dispatch table](https://en.wikipedia.org/wiki/Dispatch_table). This makes it very easy to add cases without creating nested spaghetti code.

### DIFF
--- a/src/deploy/functions/triggerRegionHelper.ts
+++ b/src/deploy/functions/triggerRegionHelper.ts
@@ -1,0 +1,51 @@
+import * as backend from "./backend";
+import * as storage from "../../gcp/storage";
+import { FirebaseError } from "../../error";
+import { logger } from "../../logger";
+
+const noop = (): Promise<void> => Promise.resolve();
+
+const LOOKUP_BY_EVENT_TYPE: Record<string, (ep: backend.EventTriggered) => Promise<void>> = {
+  "google.cloud.pubsub.topic.v1.messagePublished": noop,
+  "google.cloud.storage.object.v1.finalized": lookupBucketRegion,
+  "google.cloud.storage.object.v1.archived": lookupBucketRegion,
+  "google.cloud.storage.object.v1.deleted": lookupBucketRegion,
+  "google.cloud.storage.object.v1.metadataUpdated": lookupBucketRegion,
+};
+
+/**
+ * Sets the trigger region to what we currently have deployed
+ * @param want the list of function specs we want to deploy
+ * @param have the list of function specs we have deployed
+ */
+export async function lookupMissingTriggerRegions(want: backend.Backend): Promise<void> {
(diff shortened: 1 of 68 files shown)


inlined:
Unfortunately I had merge conflicts and had to make my changes in the merge commit, so this is impossible to review without guidance.

For a superset of the files I changed, see
* `src/deploy/functions/prepare.ts`: I added the missing trigger check to `inferDetailsFromExisting`
* `src/deploy/functions/runtimes/node/parseTriggers.ts`: I modified the clause where you check for GCS event types so it modifies both the endpoint and the function spec we're parsing
* `src/deploy/functions/triggerRegionHelper.ts`: Largely rewrote. Renamed entrypoint to indicate that this only affects functions w/ missing trigger definitions, changed signature to be backend, ran promises in parallel, and use a dispatch table
* `src/test/deploiy/functions/backend.spec.ts`: "should read v1 functions only when user is not allowlisted for v2" now includes a v1 endpoint in its oracle
* `src/test/deploy/functions/prepare.ts`: Added test that we copy trigger regions from existing functions
* `src/test/deploy/functions/triggerRegionHelper.spec.ts`: Tests now match the new function signature

src/deploy/functions/triggerRegionHelper.ts
```diff
@@ -0,0 +1,51 @@
+import * as backend from "./backend";
+import * as storage from "../../gcp/storage";
+import { FirebaseError } from "../../error";
+import { logger } from "../../logger";
+
+const noop = (): Promise<void> => Promise.resolve();
+
+const LOOKUP_BY_EVENT_TYPE: Record<string, (ep: backend.EventTriggered) => Promise<void>> = {
+  "google.cloud.pubsub.topic.v1.messagePublished": noop,
```
colerogers, line 9: Curious, why do we prefer a noop here instead of the if statement check that we had before?

inlined: Having a function pointer for noop makes it clear to the dispatching code whether something is "handled" by doing nothing or if we don't know what we're supposed to do with it. Once this code gets more event types in the table it'll seem more obvious that it's cleaner to have a flat dispatch table instead of a file full of `if/else`

PR #3819, opened 2021-10-12
Functions release now uses the Endpoints model

inlined:
Holy crap... it's happening! 🎉

Will continue to do smoke tests, but so far I've successfully deployed V1 & V2 functions. And the purge of old code has started as well!


src/deploy/functions/release/index.ts
```diff
@@ -0,0 +1,101 @@
...
+  // subtleties are so we can take out a round trip API call to get the latest
+  // trigger URLs by calling existingBackend again.
+  printTriggerUrls(payload.functions!.backend);
+
+  await containerCleaner.cleanupBuildImages(backend.allEndpoints(payload.functions!.backend));
+
+  const allErrors = summary.results.filter((r) => r.error).map((r) => r.error) as Error[];
+  if (allErrors.length) {
+    const opts = allErrors.length == 1 ? { original: allErrors[0] } : { children: allErrors };
+    throw new FirebaseError("There was an error deploying functions", { ...opts, exit: 1 });
```
taeold, line 79: I notice that exiting with `exit: 1` means that `firebase-debug.log` isn't generated. I know that we've improved our error logs, but having debug log is pretty handy; can this be `exit: 2` instead?

src/deploy/functions/release/index.ts
```diff
@@ -0,0 +1,101 @@
...
+  if (httpsFunctions.length === 0) {
+    return;
+  }
+
+  for (const httpsFunc of httpsFunctions) {
+    if (!httpsFunc.uri) {
+      logger.debug("Missing URI for HTTPS function in printTriggerUrls. This shouldn't happen");
+      continue;
+    }
+    logger.info(clc.bold("Function URL"), `(${getFunctionLabel(httpsFunc)}):`, httpsFunc.uri);
```
taeold, line 99: Think inserting a space between the label and the uri would make the link clickable on supported terminals.

src/deploy/functions/release/index.ts
inlined: WTF. I had no idea there was any meaning besides 0 for non-error

src/deploy/functions/release/index.ts
inlined: JS loggers add a space between parameters. It becomes clickable for me when in iTerm2 when I hold ⌘

PR #3820, opened 2021-10-12
Add tenant support for all other emulator endpoints

lisajian:
### Description

Added the following endpoint to support multi-tenancy op's:

- `/emulator/v1/projects/{targetProjectId}/tenants/{tenantId}/accounts`

Regenerated `apiSpec.js` and `schema.ts` with this update

### Scenarios Tested

`npm run generate:auth-api`
`npm test` passes

### Sample Commands

N/A


src/test/emulators/auth/misc.spec.ts
```diff
@@ -527,4 +581,18 @@ describeAuthEmulator("emulator utility APIs", ({ authApi }) => {
           .equals("Users are present, unable to set passthrough mode");
       });
   });
+
+  it("should error on updating usageMode on PATCH /emulator/v1/projects/{PROJECT_ID}/tenants/{TENANT_ID}/config", async () => {
```
yuchenshi, line 585: This whole API makes no sense. As you've observed, `usageMode` cannot be set per tenant, and I don't think `allowDuplicateEmails` can be set per tenant in production either. Shouldn't tenant inherit these two settings from agent project config?

scripts/gen-auth-api-spec.ts
```diff
@@ -399,6 +438,73 @@ function addEmulatorOperations(openapi3: any): void {
       tags: ["emulator"],
     },
   };
+  openapi3.paths["/emulator/v1/projects/{targetProjectId}/tenants/{tenantId}/config"] = {
```
yuchenshi, line 441: see comment above

src/test/emulators/auth/misc.spec.ts
lisajian: Yeah, that all makes sense, thanks for the clarification - I'll go ahead and revert the APIs and if they're needed later on, we can address it then 👍👍

scripts/gen-auth-api-spec.ts
lisajian: Removed

yuchenshi (approved):

PR #3825, opened 2021-10-13
Adding integration test for `firebase deploy --only extensions`

joehan:
### Description
Adds a basic e2e test for `firebase deploy --only extensions`. This test deploys 2 extensions, then verifies that they were successfully installed via `firebase ext:list --json`.

I also made some changes to listExtensions, so that the output of `firebase ext:list --json` would be simpler to parse. Previously, this was returning _every_ field in the extension instance object, including the entire ExtensionSpec. This does not have any effect on `firebase ext:list` without the `--json` flag.

Once this command is ready to merge publicly, I'll add this test to our CI as well.

### Scenarios Tested
`npm run test:extensions-deploy`


scripts/extensions-deploy-tests/tests.ts
```diff
@@ -0,0 +1,60 @@
+import { expect } from "chai";
+import * as subprocess from "child_process";
+import { cli } from "winston/lib/winston/config";
+
+import { CLIProcess } from "../integration-helpers/cli";
+
+const FIREBASE_PROJECT = process.env.FBTOOLS_TARGET_PROJECT || "";
+
+const TEST_SETUP_TIMEOUT = 10000;
```
elvisun, line 9: what unit is this? TEST_TIMEOUT_MS?

src/extensions/listExtensions.ts
```diff
@@ -31,6 +29,7 @@ export async function listExtensions(
   });
   // Order instances newest to oldest.
   const sorted = _.sortBy(instances, "createTime", "asc").reverse();
+  const formatted: any = [];
```
elvisun, line 32: is this `any` or can it be `any[]`?

PR #3828, opened 2021-10-14
Replace projectId with ${PROJECT_ID} so cross project deploys work

joehan:
### Description
`deploy --only extensions` now substitutes auto populated params into param values.
`ext:export` now checks params for projectID, and replaces it with ${param:PROJECT_ID}.

Combined, this means (in many more cases than before) that you can `ext:export` from one project and `deploy --only extensions` onto another project, without needing to change .env files. 

### Scenarios Tested
`ext:export`ed from one project, and verified that the .env files looked correct:
Screen Shot 2021-10-14 at 1 08 40 PM

then successfully deployed onto a different project:
Screen Shot 2021-10-14 at 1 00 10 PM

### DIFF
--- a/scripts/extensions-deploy-tests/tests.ts
+++ b/scripts/extensions-deploy-tests/tests.ts
@@ -6,12 +6,13 @@ import { CLIProcess } from "../integration-helpers/cli";
 
 const FIREBASE_PROJECT = process.env.FBTOOLS_TARGET_PROJECT || "";
 
-const TEST_SETUP_TIMEOUT = 10000;
-const TEST_TIMEOUT = 600000;
+const TEST_SETUP_TIMEOUT_MS = 10000;
+const TEST_TIMEOUT_MS = 600000;
+
 describe("firebase deploy --only extensions", () => {
   let cli: CLIProcess;
   before(function (this) {
-    this.timeout(TEST_SETUP_TIMEOUT);
+    this.timeout(TEST_SETUP_TIMEOUT_MS);
     expect(FIREBASE_PROJECT).to.exist.and.not.be.empty;
     cli = new CLIProcess("default", __dirname);
   });
@@ -21,7 +22,7 @@ describe("firebase deploy --only extensions", () => {
   });
 
   it("should have deployed the expected extensions", async function (this) {
-    this.timeout(TEST_TIMEOUT);
+    this.timeout(TEST_TIMEOUT_MS);
 
     await cli.start(
       "deploy",
--- a/src/extensions/listExtensions.ts
+++ b/src/extensions/listExtensions.ts
@@ -29,7 +29,7 @@ export async function listExtensions(projectId: string): Promise<any> {
   });
   // Order instances newest to oldest.
   const sorted = _.sortBy(instances, "createTime", "asc").reverse();
-  const formatted: any = [];
+  const formatted: Record<string, string>[] = [];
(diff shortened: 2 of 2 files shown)


src/commands/ext-export.ts
```diff
@@ -19,17 +19,22 @@ module.exports = new Command("ext:export")
...
+        if (s.ref) {
+          s.ref.version = `^${s.ref.version}`;
+        }
+        return s;
+      })
+      .map((i) => stripProjectId(projectId, i));
     // If an instance spec is missing a ref, that instance must have been installed from a local source.
     const [withRef, withoutRef] = partition(have, (s) => {
       return !!s.ref;
     });
```
elvisun, line 36: (s) => !!s.ref

src/deploy/extensions/planner.ts
```diff
@@ -32,8 +33,9 @@ export async function have(projectId: string): Promise<InstanceSpec[]> {
 }
 
 export async function want(
```
elvisun, line 35: can you add some comments here too?

src/extensions/export.ts
```diff
@@ -8,6 +8,22 @@ import { humanReadable } from "../deploy/extensions/deploymentSummary";
 import { logger } from "../logger";
 import { FirebaseError } from "../error";
 
+/**
+ * stripProjectId searchs spec.params for any param that include projectId,
+ * and replaces it with a parameterized version that can be used on other projects.
+ * For example, 'my-project-id.appspot.com' becomes '${param:PROJECT_ID}.appspot.com`
+ */
+export function stripProjectId(projectId: string, spec: InstanceSpec): InstanceSpec {
```
elvisun, line 16: wdyt `parameterizeProjectId` ?

src/commands/ext-export.ts
```diff
@@ -19,17 +19,22 @@ module.exports = new Command("ext:export")
...
-    const have = (await planner.have(projectId)).map((s) => {
-      if (s.ref) {
-        s.ref.version = `^${s.ref.version}`;
-      }
-      return s;
-    });
+    // Look up the instances that already exist,
+    // add a ^ to their version,
+    // and strip project IDs from the param values.
+    const have = (await planner.have(projectId))
```
elvisun, line 25: what's `have`?... 

Somehow reminds me of: https://www.youtube.com/watch?v=HTjqz0_wsZo

src/extensions/export.ts
```diff
@@ -8,6 +8,22 @@ import { humanReadable } from "../deploy/extensions/deploymentSummary";
...
+  const newParams: Record<string, string> = {};
+  for (const [key, val] of Object.entries(spec.params)) {
+    const parameterizedVal = val.replace(projectId, "${param:PROJECT_ID}");
```
elvisun, line 19: Seems a bit hacky, what if my project ID is `appspot` or something like it? Do you have an example of a list of params so we can see how common this problem can be?

src/commands/ext-export.ts
joehan: 🤣  

This is copying the pattern we use over in function deploy - we compare what you 'have' deployed on your project now to what to 'want' deployed.

huangjeff5 (approved):

src/extensions/export.ts
elvisun: sgtm, just making sure it won't break too often :)

PR #3831, opened 2021-10-14
Launch endpoints refactor

inlined:
Aggregated changes for the Endpoints refactor. This change replaces FunctionSpec/PubSubSpec/ScheduleSpc with Endpoint. Endpoint is also loosely what we parse now for the v1alpha1 container contract, though we use a slightly different type (no `id` or `project` and 1region1 is an optional array).

Some product changes:
1. We have a platform for new types of triggers not supported by GCS (i.e. you can update `backend.Triggered` to include a new type union)
2. Deployments are fully regional
3. Deletes only happen after creates and updates succeed
4. We have more and better error messages
5. We have lots more unit testing!


taeold (approved):
🔥

PR #3832, opened 2021-10-15
Fixes from bugbashing

inlined:
Fix branch for issues found during bug bashes:

1. Issues in `firebase functions:delete` are generic => Don't catch some errors (maybe shouldn't catch any) because the generic catchall error message is unhelpful
2. `firebase functions:delete` wasn't deleting functions => Add deploymentTool labels to `functionFromEndpoint`. Add option in deployment planner to delete all functions irrespective of deployment tool (for `firebase functions:delete`)
3. URLs aren't printed at the end of a deploy => Record the URI into the endpoint after deploying a function (I did this before... I must have lost it in a git rebase?)
4. Fix the extra space before "creating" "updating" or "deleting" functions
5. Don't cache trigger region if the trigger changes (e.g. new bucket)
6. firebase functions:list should list the function ID not the entrypoint
7. Logs for enabling APIs when using functions should use the prefix "functions", not the API name
8. Start copying the trigger region from the GCF API. Delete and recreate the function on update if the region changes.

### DIFF
--- a/src/commands/functions-delete.ts
+++ b/src/commands/functions-delete.ts
@@ -46,11 +46,10 @@ export default new Command("functions:delete [filters...]")
     if (options.region) {
       existingBackend.endpoints = { [options.region]: existingBackend.endpoints[options.region] };
     }
-    const plan = planner.createDeploymentPlan(
-      /* want= */ backend.empty(),
-      existingBackend,
-      context.filters
-    );
+    const plan = planner.createDeploymentPlan(/* want= */ backend.empty(), existingBackend, {
+      filters: context.filters,
+      deleteAll: true,
+    });
     const allEpToDelete = Object.values(plan)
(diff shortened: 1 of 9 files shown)


src/commands/functions-delete.ts
```diff
@@ -36,60 +36,57 @@ export default new Command("functions:delete [filters...]")
       filters: filters.map((f) => f.split(".")),
     };
 
-    try {
-      const [config, existingBackend] = await Promise.all([
-        functionsConfig.getFirebaseConfig(options),
-        backend.existingBackend(context),
-      ]);
-      await backend.checkAvailability(context, /* want=*/ backend.empty());
-      const appEngineLocation = functionsConfig.getAppEngineLocation(config);
```
inlined, line 45: Real diff: Moved everything before the fabricator call out of the `try` block

PR #3835, opened 2021-10-15
Make environment variables available to user code during function deployment

dominicbartl:
### Description

We're using the framework NestJS which has the handy feature to validate environment variables before booting an application. This has the effect that deployments (for Cloud Functions) fail when an environment variable is missing or invalid. This is not a NestJS specific feature, this could also be useful for any application. I posted a short snippet below.

I enabled the new dotenv feature and tried to deploy our functions with environment variables. Though I encountered the following problem: When deploying Firebase Functions the source code gets loaded, but the environment variables defined in the .env aren't available in this context.

I introduced a small change that the user defined variables as well as the Firebase default environment variables will be merged before being passed on. As far as I can know this will be the same set of variables when the function gets deployed runs in the cloud.

### Scenarios Tested

Here's a small example which fails when trying to deploy a function which requires an environment variable to be present.

```javascript

if (!process.env.FOO) {
    throw new Error('Environment variable FOO not set');
}

exports.helloWorld = functions.https.onRequest((request, response) => {
    response.send("Hello from Firebase! FOO=" + process.env.FOO);
});
```

Deploying this code without the change exits with the following error:

```
Error occurred while parsing your function triggers.

Error: Environment variable FOO not set
...
```

Deploying the function trigger using this change succeeds.

### DIFF
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -87,8 +87,9 @@ export async function prepare(
   track("functions_codebase_deploy_env_method", tag);
 
   logger.debug(`Analyzing ${runtimeDelegate.name} backend spec`);
-  const wantBackend = await runtimeDelegate.discoverSpec(runtimeConfig, firebaseEnvs);
-  wantBackend.environmentVariables = { ...userEnvs, ...firebaseEnvs };
+  const mergedEnvs = { ...userEnvs, ...firebaseEnvs };
+  const wantBackend = await runtimeDelegate.discoverSpec(runtimeConfig, mergedEnvs);
+  wantBackend.environmentVariables = mergedEnvs;
   payload.functions = { backend: wantBackend };
 
   // Note: Some of these are premium APIs that require billing to be enabled.


penx:
This would be very useful. My app also errors on load if environment variables are missing so `firebase deploy` does not work.

I've been able to get around this my creating `functions.yaml` in the root of my project, which means firebase deploy will skip spinning up the app locally:

```yml
specVersion: v1alpha1
endpoints:
  myFunction:
    platform: gcfv1
    entryPoint: myFunction
    httpsTrigger: {}
```

...however this took quite a bit of debugging to figure out, as it seems to be an undocumented feature.

penx:
@bkendall any chance of a review?

architectapp:
Hey, i have a way of solving this problem, although its not very clean and i have documented it here -- [Managing development, staging and production environments in firebase](https://github.com/architectapp/architect/blob/main/firebase/Managing%20development%2C%20staging%20and%20production%20environments%20in%20firebase.md)

PR #3840, opened 2021-10-17
Support Artifact Registry as an experiment

inlined:
Adds a new `artifactregistry` experiment to switch from using Google Container Registry for Artifact Registry for builds.

Customers will not actually use artifactregistry unless they are explicitly opted in on the backend, but turning on this flag will cause the CLI to clean up AR only.

When the flag is removed and everyone is opted in, we need to make the CLI delete both GCR and AR images because the backend will use GCR or AR depending on a rollout percentage. Once the experiment is fully rolled out in the backend we'll release a 3rd CLI version that then causes only AR to be cleaned up.

Some notes on the change:
1. The [docker API](https://docs.docker.com/registry/spec/api/#catalog) specifies that `DELETE <host>/v2/<name>/manifests/<manifest>` should return 202 (Accepted). GCR also returned an empty Errors response, but AR returns no content. This crashed the `apiv2` library, so I modified it to return `undefined` safely when results are zero length + add guards to the `gcr` library.
2. The change is enough to clean up the actual images, but it doesn't seem to be enough to clean up the directories. I'll follow up w/ the TL of AR to figure out why.


src/apiv2.ts
```diff
@@ -361,11 +361,14 @@ export class Client {
 
     let body: ResT;
     if (options.responseType === "json") {
-      // 204 statuses have no content. Don't try to `json` it.
-      if (res.status === 204) {
+      const text = await res.text();
```
colerogers, line 364: It might be good to get @bkendall 's approval on the `api` changes, even though it's only a few lines

src/deploy/functions/containerCleaner.ts
```diff
@@ -96,6 +109,21 @@ export async function cleanupBuildImages(functions: backend.FunctionSpec[]): Pro
...
+export class ArtifactRegistryCleaner {
+  // GCFv1 for AR has the following directory structure
+  // Hostname: <region>-docker.pkg.dev
+  // Directory structure:
+  // gcf-artifacts/
+  //     +- <function ID>
+  //     +- <function ID>/cache
+  // We leave the cache directory alone because it only costs
+  // a few MB and improves performance
+  async cleanupFunction(func: backend.TargetIds, helper?: DockerHelper): Promise<void> {
```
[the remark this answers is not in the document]
inlined: Fair. I was just copying the existing code but that's probably unnecessary.

src/apiv2.ts
inlined: Agreed. Will wait on his approval before submitting.

inlined:
Committing now since this is a preview feature. Outstanding fixes:

1. We should add an option to stop recursion in `rm` because we want to delete `functionID` but not `functionID/cache`
2. We need to delete the "directory" on cleanup
3. We need to modify GCR and AR to fully purge functions' data on deletion. IIRC we only delete existingBackend.

PR #3853, opened 2021-10-21
Adding secrets support to `firebase deploy --only extensions`

joehan:
### Description
Adding secrets support to `firebase deploy --only extensions`. 
Unit tests for the new methods in secrets.ts to follow - I wanted to get some eyes on this before it got too late in the day/

The logic here can be a little hard to follow - to help, I made a flow chart: go/firex-deploy-secrets-diagram

### Scenarios Tested
I tested each of the cases in go/firex-deploy-secrets-diagram using a bunch of instances of pavelj/secrets-test@0.0.11 & 0.0.10
Screen Shot 2021-10-21 at 10 55 09 AM

### DIFF
--- a/src/deploy/extensions/secrets.ts
+++ b/src/deploy/extensions/secrets.ts
@@ -1,0 +1,233 @@
+import * as clc from "cli-color";
+
+import * as secretUtils from "../../extensions/secretsUtils";
+import * as secretManager from "../../gcp/secretManager";
+
+import { Payload } from "./args";
+import { getExtensionVersion, InstanceSpec } from "./planner";
+import { promptCreateSecret } from "../../extensions/askUserForParam";
+import { ExtensionSpec, Param, ParamType } from "../../extensions/extensionsApi";
+import { FirebaseError } from "../../error";
+import { logger } from "../../logger";
+import { logLabeledBullet } from "../../utils";
+
+interface SecretInfo {
+  secret?: secretManager.Secret;
+  secretVersion?: secretManager.SecretVersion;
+  labels: Record<string, string>;
+}
+
+export async function handleSecretParams(
+  payload: Payload,
+  have: InstanceSpec[],
+  nonInteractive: boolean
+) {
+  for (const i of payload.instancesToCreate ?? []) {
+    if (await checkSpecForSecrets(i)) {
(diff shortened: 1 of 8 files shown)


src/deploy/extensions/secrets.ts
```diff
@@ -0,0 +1,232 @@
...
+  secretVersion?: secretManager.SecretVersion;
+  labels: Record<string, string>;
+}> {
+  const secretInfo: any = {
+    labels: {},
+  };
+  try {
+    secretInfo.secret = await secretManager.getSecret(projectId, secretName);
+    secretInfo.secretVersion = await secretManager.getSecretVersion(projectId, secretName, version);
+    secretInfo.labels = await secretManager.getSecretLabels(projectId, secretName);
```
pavelgj, line 204: optional: getSecretLabel is identical to getSecret two lines above, it just only returns labels and the getSecret doesn't include labels... we should really combine those two....

pavelgj: I mean, it's my own fault... for implementing it this way, I just didn't need those two in the same place.... :)

src/deploy/extensions/secrets.ts
```diff
@@ -0,0 +1,232 @@
...
+    throw new FirebaseError(
+      `${clc.bold(i.instanceId)}: Found '${providedValue}' for secret param ${
+        secretParam.param
+      }. ` +
+        `projects/${projectId}/secrets/${secretName} is managed by a different extension instance (${
+          secretInfo.labels[secretUtils.SECRET_LABEL]
+        }), so reusing it here can lead to unexpected behavior. ` +
+        "Please choose a different name for this secret, and rerun this command."
+    );
+  }
```
pavelgj, line 133: If secret/version exist and don't have the label we just use it, right? I think you still need to call `secretsUtils.grantFirexServiceAgentSecretAdminRole(secretInfo.secret)` on it so that we get all the necessary P4SA permissions. Similar in the update case.

pavelgj (changes requested):

src/deploy/extensions/secrets.ts
joehan: Haha, I noticed this and made a mental note to fix it later, then forgot to fix it later. I'll dedupe this logic

PR #3854, opened 2021-10-21
GCS fixes

colerogers:
This change fixes the following:

- Cloud Storage service agent doesn't have the pubsub publisher role
- Trigger region must match function region or be in the same multi-region as bucket
- Trigger region not being set before the function deployment

We also move some key type definitions into a separate file `src/deploy/functions/types.ts` for cleaner separation of our data models.


src/deploy/functions/checkIam.ts
```diff
@@ -135,69 +137,85 @@ function reduceEventsV2(filtered: Set<EventShorthand>, option: backend.Endpoint)
...
         (b) => b.role === PUBSUB_PUBLISHER_ROLE && b.members.find((m) => m === storageServiceAgent)
       )
     ) {
-      throw new FirebaseError("IAM Policy did not update correctly");
+      throw new Error(
+        `Could not find the storage service agent under the pubsub role binding in the updated policy.`
+      );
     }
+  } catch (err) {
+    logger.error();
```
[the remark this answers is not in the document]
colerogers: Left over log line, removed

src/gcp/location.ts
```diff
@@ -0,0 +1,100 @@
+// A flattening of container_registry_hosts and
+// region_multiregion_map from regionconfig.borg
```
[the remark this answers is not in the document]
colerogers: That was actually in the old file too, but it makes sense to remove internal references, fixed in latest commit

src/gcp/location.ts
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Helper function to return a mapping from a geographical region to a subdomain.
+ * This mapping includes subdomains that can be outside the geographical multi-region mapping
+ * Some examples are 'southamerica-east1' -> 'us' or 'australia-southeast1' -> 'asia'
+ * @returns a Record<string, string> of region to subdomain
+ */
+export function getSubdomainMapping(): Record<string, string> {
+  return SUBDOMAIN_MAPPING;
+}
```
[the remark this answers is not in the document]
colerogers: I was a bit hesitant since we'd have to expose api's that basically re-implement the underlying Record features, like Object.keys() & Object.values(). But if we ever end up changing the data structure, it would be nice to have this interface. Ended up adding `regionToSubdomain()`, `getContainerRegistryRegions()`, and `getContainerRegistrySubdomains()` in the latest commit.

taeold (approved):
lgtm - let's check in with @inlined to see if there are any other fixes he'd want us to land.

src/deploy/functions/triggerRegionHelper.ts
```diff
@@ -2,50 +2,72 @@ import * as backend from "./backend";
...
+const ENSURE_BY_EVENT_TYPE: Record<
+  string,
+  (ep: backend.EventTriggered, r: string) => Promise<void>
+> = {
   "google.cloud.pubsub.topic.v1.messagePublished": noop,
-  "google.cloud.storage.object.v1.finalized": lookupBucketRegion,
-  "google.cloud.storage.object.v1.archived": lookupBucketRegion,
-  "google.cloud.storage.object.v1.deleted": lookupBucketRegion,
-  "google.cloud.storage.object.v1.metadataUpdated": lookupBucketRegion,
+  "google.cloud.storage.object.v1.finalized": ensureStorageTriggerRegion,
```
inlined, line 14: This is the second time we have a lookup map from event type -> strategy function (that varies only by service).

I wonder if a cleaner way of understanding the codebase as a whole is to have a lookup function from event type => service delegate.

So something like:

```
// deploy/functions/services/index.ts
interface Service {
   // E.g. "pubsub", "storage"
   readonly name: string;
  
   // Rough swag here
   missingProjectBindings(projectId: string, existingPolicy: iam.Policy): Promise<Array<iam.Binding>>
   ensureTriggerRegion(): Promise<void>;
}

// Up to you to decide whether to return undefined or a noop-ing service for v1/pub/sub etc
// Though I personally might choose a noop impl to avoid null checks everywhere.
function serviceForEndpoint(endpoint: backend.Endpoint): Service

// deploy/functions/service/storage.ts
class StorageService implements Service {
  ...
};
```

Take this as a "wondering aloud", not a strong nudge. I'm curious for your thoughts on future comprehensibility.

src/deploy/functions/triggerRegionHelper.ts
```diff
@@ -2,50 +2,72 @@ import * as backend from "./backend";
...
+      endpoint.eventTrigger.region = bucket.location.toLowerCase();
+      logger.debug("Setting the event trigger region to", endpoint.eventTrigger.region, ".");
+    } catch (err) {
+      throw new FirebaseError("Can't find the storage bucket region", { original: err });
+    }
+  }
+  // check for invalid cloud storage trigger region
+  if (
+    endpointRegion !== endpoint.eventTrigger.region &&
+    endpoint.eventTrigger.region !== "us-central1" &&
```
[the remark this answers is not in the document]
inlined: Your confusion is merited and it's worth noting in a comment that GCF allows any trigger to be in us-central

src/deploy/functions/triggerRegionHelper.ts
```diff
@@ -2,50 +2,72 @@ import * as backend from "./backend";
...
+    !regionInLocation(endpointRegion, endpoint.eventTrigger.region!)
+  ) {
+    throw new FirebaseError("Function cannot be deployed outside of the trigger region");
```
inlined, line 71: Our customers probably aren't thinking about "trigger regions" since we don't ask them to provide one. Can we say something like `A function in region ${endpoint.region} cannot listen to a bucket in region ${endpoint.eventTrigger.region}` ?

src/gcp/location.ts
```diff
@@ -0,0 +1,112 @@
+// A mapping from geographical region to subdomain, usefull for Container Registry
+const SUBDOMAIN_MAPPING: Record<string, string> = {
```
inlined, line 2: If this is strictly GCR related it can possibly be in `docker.ts` and be something like `GCR_SUBDOMAIN_MAPPING` (I know we've made you move this already)

src/gcp/storage.ts
```diff
@@ -0,0 +1,256 @@
+import * as path from "path";
+import * as api from "../api";
+import { logger } from "../logger";
+import { FirebaseError } from "../error";
+
+/** Bucket Interface */
+interface BucketResponse {
```
inlined, line 7: These fields are all guaranteed to be present?

src/gcp/location.ts
```diff
@@ -0,0 +1,112 @@
...
+  "europe-west4": "eur4",
+  /** nam4 */
+  "us-central1": "nam4",
+  "us-east1": "nam4",
+};
+
+/**
+ * Obtains the subdomains for all the hostname options in Container Registry
+ */
+export function getContainerRegistrySubdomains(): string[] {
```
inlined, line 76: What is the point of these trivial functions? So we can stub them in tests?

src/gcp/storage.ts
colerogers: Both the public documentation (https://cloud.google.com/storage/docs/json_api/v1/buckets#resource-representations) and the proto definition show all these fields as non-optional

src/gcp/location.ts
colerogers: In the discussion above with @taeold, we thought it would be best to expose a helper function in case we ever had to change the underlying data structure. Would you prefer that we just expose the actual constants?

inlined: I think for a public API that level of defensiveness can make sense. For an internal codebase it may be overly defensive.

src/deploy/functions/checkIam.ts
```diff
@@ -3,11 +3,25 @@ import { bold } from "cli-color";
...
+import { Policy, testIamPermissions, testResourceIamPermissions } from "../../gcp/iam";
 import * as args from "./args";
 import * as backend from "./backend";
 import * as track from "../../track";
 import { Options } from "../../options";
+import * as storage from "../../gcp/storage";
+import { getIamPolicy, setIamPolicy } from "../../gcp/resourceManager";
+import { Service, getV2ServiceMapping } from "./eventTypes";
+
+const EVENT_V2_SERVICE_MAPPING = getV2ServiceMapping();
```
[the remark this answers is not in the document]
colerogers: Changed to exporting the constants per the other thread

src/deploy/functions/checkIam.ts
```diff
@@ -104,3 +118,103 @@ export async function checkHttpIam(
   }
   logger.debug("[functions] found setIamPolicy permission, proceeding with deploy");
 }
+
+/** Callback function to find all v2 service types in from events */
+function reduceEventsToServicesV2(filtered: Set<Service>, option: backend.Endpoint) {
```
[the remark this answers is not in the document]
colerogers: It was called "option" since this is a callback for a reduce function and we have the option to take each endpoint, I've changed it to "endpoint" in the latest commit

src/deploy/functions/checkIam.ts
```diff
@@ -104,3 +118,103 @@ export async function checkHttpIam(
...
+      logger.debug(
+        "Cannot find the correct function mapping that grants roles for ",
+        service,
+        " service."
+      );
+      continue;
+    }
+    ensureCorrectRoles.push(rolesFn(projectId));
+  }
+  // TODO(colerogers): When we add another service to enable, check if we also call setIamPolicy,
```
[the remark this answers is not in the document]
colerogers: It makes sense to refactor the code now, I've moved the getIamPolicy and setIamPolicy to happen before and after we find the bindings & re-worked the parallel function to return the necessary bindings.

src/deploy/functions/containerCleaner.ts
```diff
@@ -318,20 +295,21 @@ export async function deleteGcfArtifacts(
...
+  const invalidRegion = locations.find((loc) => !regionToSubdomain(loc));
   if (invalidRegion) {
     throw new FirebaseError(`Invalid region ${invalidRegion} supplied`);
   }
-  const subdomains = new Set(Object.values(SUBDOMAIN_MAPPING));
+  const subdomains = new Set(Object.values(getContainerRegistrySubdomains()));
   const failedSubdomains: string[] = [];
 
   const deleteLocations = locations.map((loc) => {
+    const subdomain = regionToSubdomain(loc) || "us";
```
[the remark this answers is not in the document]
colerogers: This isn't an edge case we need to handle, changed back to what we had before

src/deploy/functions/eventTypes.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const EVENT_V2_SERVICE_MAPPING: Record<string, Service> = {
+  "google.cloud.pubsub.topic.v1.messagePublished": "pubsub",
+  "google.cloud.storage.object.v1.finalized": "storage",
+  "google.cloud.storage.object.v1.archived": "storage",
+  "google.cloud.storage.object.v1.deleted": "storage",
+  "google.cloud.storage.object.v1.metadataUpdated": "storage",
+};
+
+const STORAGE_V2_EVENTS_SET: Set<string> = new Set<string>([
```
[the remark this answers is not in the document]
colerogers: Changed to use an array

src/deploy/functions/eventTypes.ts
```diff
@@ -0,0 +1,32 @@
...
+  "google.cloud.storage.object.v1.finalized",
+  "google.cloud.storage.object.v1.archived",
+  "google.cloud.storage.object.v1.deleted",
+  "google.cloud.storage.object.v1.metadataUpdated",
+]);
+
+/**
+ * @returns A mapping from v2 events to the underlying service
+ */
+export function getV2ServiceMapping(): Record<string, Service> {
```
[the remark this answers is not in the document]
colerogers: Removed all getter functions & exported the constants directly

src/deploy/functions/triggerRegionHelper.ts
colerogers: That makes a lot of sense and will probably help us out in the future. In the latest commit I've implemented this pattern.

src/gcp/location.ts
colerogers: Moved

src/gcp/location.ts
colerogers: Removed all getters & exported the constants directly

src/deploy/functions/checkIam.ts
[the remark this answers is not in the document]
colerogers: Removed all getters and exported the constants directly

src/deploy/functions/checkIam.ts
```diff
@@ -104,3 +107,82 @@ export async function checkHttpIam(
...
+    if (updatedBindings.length === 0) {
+      continue;
+    }
+    updatedBindings.forEach((updatedBinding) => {
+      const ndx = policy.bindings.findIndex(
+        (policyBinding) => policyBinding.role === updatedBinding.role
+      );
+      if (ndx === -1) {
+        policy.bindings.push(updatedBinding);
+        return;
```
inlined, line 168: Is this supposed to be a continue?

src/deploy/functions/checkIam.ts
```diff
@@ -104,3 +107,82 @@ export async function checkHttpIam(
...
+  } catch (err) {
+    logger.warn(
+      "We failed to obtain the IAM policy for the project,",
+      " the function deployment might fail if service ",
+      "accounts do not have the correct assigned roles."
+    );
+    return;
+  }
+  // run in parallel all the missingProjectBindings jobs
+  const findUpdatedBindings: Array<Promise<Array<iam.Binding>>> = [];
```
inlined, line 152: nit: This IAM policy merge logic is dense. I'm surprised you don't want it to be a helper method so you can easily test it

src/deploy/functions/checkIam.ts
```diff
@@ -104,3 +107,82 @@ export async function checkHttpIam(
...
+  newServices.forEach((service) =>
+    findUpdatedBindings.push(service.missingProjectBindings(projectId, policy))
```
inlined, line 154: thought: does it get any simpler if service just returns `requiredProjectBindings` rather than `missingProjectBindings`? We already handle the case where the policy already exists.

src/deploy/functions/checkIam.ts
```diff
@@ -104,3 +107,82 @@ export async function checkHttpIam(
...
+  );
+  if (newServices.length === 0) {
+    return;
+  }
+  // get the full project iam policy
+  let policy: iam.Policy;
+  try {
+    policy = await getIamPolicy(projectId);
+  } catch (err) {
+    logger.warn(
```
inlined, line 144: This should probably be logLabeledBullet with the "functions" prefix. I'd also ditch the royal "We". I'm also wondering if we can make this warning more actionable. Maybe something like:

```
Could not verify the necessary IAM configuration for the following newly-integrated services: pubsub.googleapis.com, storage.googleapis.com. Deployment may fail.
```

This would spawn StackOverflow support Qs where people ask what they can do if storage.googleapis.com cannot be validated and deployments are failing. Those Qs would be answered with the exact steps an admin needs to do to unblock deployments.

src/gcp/storage.ts
```diff
@@ -0,0 +1,256 @@
+import * as path from "path";
```
inlined, line 1: nit: blank line between external library imports and intra-package imports

src/test/deploy/functions/checkIam.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+  runtime: "nodejs14",
+};
+
+describe("checkIam", () => {
+  let storageStub: sinon.SinonStub;
+  let getIamStub: sinon.SinonStub;
+  let setIamStub: sinon.SinonStub;
+
+  beforeEach(() => {
+    storageStub = sinon.stub(storage, "getServiceAccount").throws("Do not call");
```
inlined, line 30: Nit: I'd prefer a clearer error like "unexpected call to storage.getServiceAccount" because it helps debug test failures.

src/test/deploy/functions/triggerRegionHelper.spec.ts
```diff
@@ -104,5 +104,27 @@ describe("TriggerRegionHelper", () => {
...
+            bucket: "my-bucket",
+          },
+          retry: false,
+        },
+        ...SPEC,
+        region: "europe-west4",
+      };
+
+      await expect(triggerRegionHelper.ensureTriggerRegions(backend.of(wantFn))).to.be.rejectedWith(
+        "A function in region europe-west4 cannot listen to a bucket in region us"
```
inlined, line 126: My heart fluttered with excitement at the clarity we're adding to these complex rules.

src/deploy/functions/checkIam.ts
```diff
@@ -104,3 +107,82 @@ export async function checkHttpIam(
...
+  want: backend.Backend,
+  have: backend.Backend
+): Promise<void> {
+  // find new services
+  const wantServices = backend.allEndpoints(want).reduce(reduceEventsToServices, []);
+  const haveServices = backend.allEndpoints(have).reduce(reduceEventsToServices, []);
+  const newServices = wantServices.filter(
+    (wantS) => !haveServices.find((haveS) => wantS.name === haveS.name)
+  );
+  if (newServices.length === 0) {
```
inlined, line 136: I know I was in favor of serviceForEndpoint always returning a service (even if noop) but that means that I'll have a getIamPolicy call if I had a pub/sub function and then I add an HTTPS function. I wonder if there's a way to clean this up a bit.

src/deploy/functions/checkIam.ts
inlined: Note that we'd need to avoid listing "noop" here.

inlined (approved):
Really liking this so far. Let's just clean up the error reporting for checking IAM policies.

src/deploy/functions/checkIam.ts
colerogers: Nope, this was just returning from the callback. I ended up changing it `for(...)` for better readability in the latest commit.

src/deploy/functions/checkIam.ts
colerogers: Broke out the merge logic to a separate function and added more tests

src/deploy/functions/checkIam.ts
colerogers: It doesn't really get any simpler, just shaves off a couple lines. But since we already have the logic to merge bindings, i went ahead and changed the function to always return the required bindings.

src/gcp/storage.ts
colerogers: fixed

src/test/deploy/functions/checkIam.spec.ts
colerogers: Fixed

src/deploy/functions/checkIam.ts
colerogers: I went ahead and added a check in the reduce callback that does not all services that are `noop` or `pubsub` since those don't require us to set bindings

src/deploy/functions/checkIam.ts
colerogers: Sounds good. Since I ended up changing how the reduce function works (no longer adds `noop`s or `pubsub`) we can assume that every new service will actually require a binding update.

src/deploy/functions/checkIam.ts
colerogers: Actually having an if check didn't make sense long term, update the type of the `requiredProjectBindings` function to be `undefined` in services and simply check for that in the reduce callback

PR #3882, opened 2021-11-03
GCS Emulator Tests

colerogers:
Storage integration tests for the Firebase Emulator.


taeold (approved):
👏🏼

scripts/integration-helpers/framework.ts
```diff
@@ -10,11 +10,19 @@ const FIREBASE_PROJECT_ZONE = "us-central1";
  */
 /* Functions V2 */
 const PUBSUB_FUNCTION_V2_LOG = "========== PUBSUB V2 FUNCTION ==========";
+const STORAGE_FUNCTION_ARCHIVED_LOG = "========== STORAGE FUNCTION ARCHIVED ==========";
+const STORAGE_FUNCTION_DELETED_LOG = "========== STORAGE FUNCTION DELETED ==========";
+const STORAGE_FUNCTION_FINALIZED_LOG = "========== STORAGE FUNCTION FINALIZED ==========";
+const STORAGE_FUNCTION_METADATA_LOG = "========== STORAGE FUNCTION METADATA ==========";
```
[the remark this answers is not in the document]
colerogers: Wooooooops, fixed

package.json
```diff
@@ -195,7 +195,7 @@
     "eslint-plugin-prettier": "^3.3.1",
     "firebase": "^7.24.0",
     "firebase-admin": "^9.4.2",
-    "firebase-functions": "^3.15.0",
+    "firebase-functions": "^3.16.0",
```
colerogers, line 198: Wondering if it's okay to bump this? Is there anything else that I need to do?

taeold: Not really, but for this change, don't we need to update https://github.com/firebase/firebase-tools/blob/master/scripts/triggers-end-to-end-tests/functions/package.json instead?

colerogers: Thought it would be better to update the root package since 3.16 includes all storage enhancements.

I've reverted this bump and bumped the test package instead.

inlined (commented):
What happens if the test project doesn't have GCFv2 enabled? I'm surprised this hasn't broken CI/CD.

colerogers:
> What happens if the test project doesn't have GCFv2 enabled? I'm surprised this hasn't broken CI/CD.

Do you mean having the project be on the GCFv2 allow list or just having the preview flag enabled? 

This testing package uses the project id to control the emulator, nothing is sent to GCF so the project doesn't need to be allowlisted. 

If you mean the preview flag, then I think that change should go in a separate PR as a new feature. Right now, the entire emulator suite doesn't take into account `previews.functionsv2` when handling v2 functions.

inlined (approved):
I see. For some reason I assumed this test (also?) ran against production, but that would be the integration tests in `firebase-functions`. LGTM

PR #3884, opened 2021-11-03
Added ext:dev:deprecate and ext:dev:undeprecate commands.

apascal07:


src/commands/ext-dev-deprecate.ts
```diff
@@ -0,0 +1,77 @@
+import * as clc from "cli-color";
+
+import { Command } from "../command";
```
joehan, line 3: Super nit: I prefer to move all * imports to the top of this section.

src/commands/ext-dev-deprecate.ts
```diff
@@ -0,0 +1,77 @@
...
+  .option("-m, --message <deprecationMessage>", "deprecation message")
+  .option(
+    "-f, --force",
+    "override deprecation message for existing deprecated extension versions that match"
+  )
+  .before(requireAuth)
+  .before(ensureExtensionsApiEnabled)
+  .action(async (extensionRef: string, versionPredicate: string, options: any) => {
+    const { publisherId, extensionId, version } = refs.parse(extensionRef);
+    const baseExtensionRef = publisherId + "/" + extensionId;
```
joehan, line 27: This seems unnecessary - you immediately error out if the provided extensionRef isn't in this format, so you can probably just use that.

src/commands/ext-dev-deprecate.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+    if (!publisherId || !extensionId) {
+      throw new FirebaseError(
+        `Error parsing publisher ID and extension ID from extension reference '${clc.bold(
+          extensionRef
+        )}'. Please use the format '${clc.bold("<publisherId>/<extensionId>")}'.`
+      );
+    }
+    const { comparator, targetSemVer } = parseVersionPredicate(versionPredicate);
+    const extensionVersions = await listExtensionVersions(baseExtensionRef);
```
joehan, line 43: I'm ok with this approach - however, the backend does support > and < list filtering for ExtensionVersions. We use it for displaying releaseNotes in the CLI if you want an example

src/commands/ext-dev-deprecate.ts
```diff
@@ -0,0 +1,77 @@
...
+      }
+      const message =
+        extensionVersion.state === "DEPRECATED" ? ", will overwrite deprecation message" : "";
+      utils.logLabeledBullet(extensionVersion.ref, extensionVersion.state + message);
+      return true;
+    });
+    if (filteredExtensionVersions.length > 0) {
+      const confirmMessage =
+        "You are about to deprecate these extension version(s). Do you wish to continue?";
+      const consent = await promptOnce({
```
joehan, line 60: If --force is passed, this should not prompt, and instead should automatically continue

src/commands/ext-dev-deprecate.ts
```diff
@@ -0,0 +1,77 @@
...
+        message: confirmMessage,
+        default: false,
+      });
+      if (!consent) {
+        throw new FirebaseError("Deprecation canceled.");
+      }
+    } else {
+      throw new FirebaseError("No extension versions matched the version predicate.");
+    }
+    await utils.allSettled(
```
joehan, line 71: What happens if one of these calls errors? Do we report it correctly with this set up?

src/commands/ext-dev-undeprecate.ts
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Undeprecate all extension versions that match the version predicate.
+ */
+export default new Command("ext:dev:undeprecate <extensionRef> <versionPredicate>")
+  .description("undeprecate extension versions that match the version predicate")
+  .before(requireAuth)
+  .before(ensureExtensionsApiEnabled)
+  .action(async (extensionRef: string, versionPredicate: string, options: any) => {
+    const { publisherId, extensionId, version } = refs.parse(extensionRef);
+    const baseExtensionRef = publisherId + "/" + extensionId;
```
joehan, line 27: Same as above - probably unneeded

src/commands/ext-dev-undeprecate.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+    if (!publisherId || !extensionId) {
+      throw new FirebaseError(
+        `Error parsing publisher ID and extension ID from extension reference '${clc.bold(
+          extensionRef
+        )}'. Please use the format '${clc.bold("<publisherId>/<extensionId>")}'.`
+      );
+    }
+    const { comparator, targetSemVer } = parseVersionPredicate(versionPredicate);
+    const extensionVersions = await listExtensionVersions(baseExtensionRef);
```
joehan, line 43: Optional: Same as above - consider using the filter query param instead

joehan (approved):
Mostly LGTM - a few comments and things to consider, but nothing blocking.

src/commands/ext-dev-deprecate.ts
apascal07: Moved to top of local sources, I think that's what you meant?

src/commands/ext-dev-deprecate.ts
apascal07: I believe if one call errors out, the rest will finish successfully but the user will receive an error instead of the success message. Still not clear that some may have succeeded though. Do you think this is a problem?

src/commands/ext-dev-deprecate.ts
apascal07: Oh yeah, this is way better. Thanks for the suggestion!

src/commands/ext-dev-deprecate.ts
joehan: Ok - I think its Ok as is for now, but lets follow up and improve this error handling. Ideally, we should build a little rig to catch all the errors, and report which version were successfully deprecated/undeprecated, and which had errors.

For now, I suspect that the error rate once we get to this point is so low that it is not important to have nice error handling.

apascal07: I agree, not ideal UX right now, but the saving grace is that if let's say two of three versions fail to get deprecated and the user receives a "failed to deprecate" message, simply retrying the previous command will correctly deprecate the remaining versions whether it's with `--force` or not.

PR #3887, opened 2021-11-05
Emulator bugfixes

taeold:
1. Functions emulator now checks that the function names are valid before loading them.
2. Fixes up the incomplete CloudEvent payload for Storage and Pub/Sub


yuchenshi (approved):

src/emulator/storage/cloudFunctions.ts
```diff
@@ -77,9 +84,9 @@ export class StorageCloudFunctions {
   private createLegacyEventRequestBody(
     action: StorageCloudFunctionAction,
     objectMetadataPayload: ObjectMetadataPayload
-  ): string {
+  ) {
```
colerogers, line 87: Should we type this function? Since we're already importing `CloudEvent` from `events/types.ts`, we can also import `LegacyEvent` and use that here.

colerogers (approved):

src/emulator/storage/cloudFunctions.ts
taeold: Great call.

Issue #3407, opened 2021-05-26
firebase init functions should set package.json's engine to correct nodejs version


taeold:
When running `firebase init functions`, package.json setup on the dev machine [hardcodes the `engine.node` attribute](https://github.com/firebase/firebase-tools/blob/master/templates/init/functions/typescript/package.lint.json#L13):

```
"engines": { "node": "14"  }
```

`engine.node` attribute is used at `firebase deploy` time to figure out the runtime of the deployed cloud function instance.

In `npm`, it's okay for node version specified in the `engines.node` to differ w/ the node version on the developer's machine, but `yarn` will [throw an error if they differ](https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-ignore-scripts).

It would be nice if the `firebase init functions` can dynamically set the engine. If dev is using a node version unsupported by Google Cloud Function, we can probably error.

Relates to https://github.com/firebase/firebase-functions/issues/782

oliver-pham:
Hi @taeold! I'm interested in giving this issue a try. Do you think it's a good approach to replace the value of `engine.node` with [`process.version`](https://nodejs.org/api/process.html#processversion) before writing `package.json`?

https://github.com/firebase/firebase-tools/blob/c8e99d2d13ad5f371965e26ce4344034dab5f51c/src/init/features/functions/typescript.js#L34-L47

I can add a function to load the JSON file (`package.lint.json` or `package.nolint.json`) as an object and modify `engine.node` attribute.

taeold:
@oliver-pham Great to hear from you!

Few thoughts:

1. I'd lean on doing a simple replacement replacement, similar to what's done in the [Firestore Rules template](https://github.com/firebase/firebase-tools/blob/c8e99d2d13ad5f371965e26ce4344034dab5f51c/templates/init/firestore/firestore.rules#L13).
2. We'll need a logic to error/warn users if the current version of node being used in not supported in Google Cloud Functions. Following node versions are supported:

https://github.com/firebase/firebase-tools/blob/c2feb0836f6f64236e117f2906ef6083840e212b/src/deploy/functions/runtimes/index.ts#L10 for

oliver-pham:
@taeold as for warning users of deprecated node version, I think it's already handled by Firebase CLI. Should I still implement it?

For example, if my node version is 9, when I run `firebase init functions`, I immediately get this error:
```
[...]

PR #3894, opened 2021-11-09
Dynamically validate & set `engines.node` in `package.json` on `firebase init functions`

oliver-pham:
### Description

Fix https://github.com/firebase/firebase-tools/issues/3407

This PR should help the `engines.node` attribute in `package.json` file  dynamically set to the current Node.js version on a dev's machine. If their Node.js version is not supported by Google Cloud Functions, an error will be logged.

For example, if the Node.js version on a dev's machine is 12. After running `firebase init functions`, the generated `package.json` file should contain thess lines:

```json
"engines": { 
  "node": "12"
}
```

### Scenarios Tested

* Automatically detect and set Node engine version for `package.json` with ESLint
* Automatically detect and set Node engine version for `package.json` without ESLint

### DIFF
--- a/src/init/features/functions/typescript.js
+++ b/src/init/features/functions/typescript.js
@@ -6,6 +6,7 @@ var path = require("path");
 
 var npmDependencies = require("./npm-dependencies");
 var { prompt } = require("../../../prompt");
+var utils = require("../../../utils");
 
 var TEMPLATE_ROOT = path.resolve(__dirname, "../../../../templates/init/functions/typescript/");
 var PACKAGE_LINTING_TEMPLATE = fs.readFileSync(
@@ -38,13 +39,13 @@ module.exports = function (setup, config) {
           'npm --prefix "$RESOURCE_DIR" run build',
         ]);
         return config
-          .askWriteProjectFile("functions/package.json", PACKAGE_LINTING_TEMPLATE)
+          .askWriteProjectFile("functions/package.json", getPackage(setup.functions.lint))
           .then(function () {
             return config.askWriteProjectFile("functions/.eslintrc.js", ESLINT_TEMPLATE);
           });
       }
       _.set(setup, "config.functions.predeploy", 'npm --prefix "$RESOURCE_DIR" run build');
-      return config.askWriteProjectFile("functions/package.json", PACKAGE_NO_LINTING_TEMPLATE);
+      return config.askWriteProjectFile("functions/package.json", getPackage(setup.functions.lint));
     })
     .then(function () {
       return config.askWriteProjectFile("functions/tsconfig.json", TSCONFIG_TEMPLATE);
@@ -64,3 +65,11 @@ module.exports = function (setup, config) {
       return npmDependencies.askInstallDependencies(setup.functions, config);
     });
 };
+
+function getPackage(useLint) {
+  const nodeEngineVersion = utils.getNodeVersionString();
+  if (useLint) {
+    return PACKAGE_LINTING_TEMPLATE.replace(/{{NODE_VERSION}}/g, nodeEngineVersion);
+  }
+  return PACKAGE_NO_LINTING_TEMPLATE.replace(/{{NODE_VERSION}}/g, nodeEngineVersion);
+}
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -559,6 +559,20 @@ export function thirtyDaysFromNow(): Date {
 }
 
 /**
+ * Get the current version of Node.js engine from `process.version`.
+ * @return Node.js major release version
+ */
+export function getNodeVersionString(): string {
+  const nodeVersion: RegExpExecArray | null = /\d+/.exec(process.version);
+
+  if (nodeVersion === null) {
+    throw new Error("Node version is not found");
+  }
+
+  return nodeVersion.join("");
+}
+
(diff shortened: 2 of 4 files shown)


src/init/features/functions/typescript.js
```diff
@@ -6,6 +6,7 @@ var path = require("path");
 
 var npmDependencies = require("./npm-dependencies");
 var { prompt } = require("../../../prompt");
+var utils = require("../../../utils");
```
taeold, line 9: Since we are only using `getNodeVersionString`, I think it make sense to do something like

```
var { getNodeVersionString}  = require("../../../utils");
```

src/init/features/functions/typescript.js
```diff
@@ -38,13 +39,13 @@ module.exports = function (setup, config) {
           'npm --prefix "$RESOURCE_DIR" run build',
         ]);
         return config
-          .askWriteProjectFile("functions/package.json", PACKAGE_LINTING_TEMPLATE)
+          .askWriteProjectFile("functions/package.json", getPackage(setup.functions.lint))
```
taeold, line 42: Since we know `setup.functions.lint` is `true` at this point, let's explicitly pass it.

src/init/features/functions/typescript.js
```diff
@@ -64,3 +65,11 @@ module.exports = function (setup, config) {
       return npmDependencies.askInstallDependencies(setup.functions, config);
     });
 };
+
+function getPackage(useLint) {
+  const nodeEngineVersion = utils.getNodeVersionString();
```
taeold, line 70: When do we ever check that the user's node version is one supported by Google Cloud Functions?

src/utils.ts
```diff
@@ -558,6 +558,20 @@ export function thirtyDaysFromNow(): Date {
   return new Date(Date.now() + THIRTY_DAYS_IN_MILLISECONDS);
 }
 
+/**
+ * Get the current version of Node.js engine from `process.version`.
+ * @return Node.js major release version
+ */
+export function getNodeVersionString(): string {
+  const nodeVersion: RegExpExecArray | null = /\d+/.exec(process.version);
```
taeold, line 566: nit* I think I'd prefer non-regex solution like `process.versions.node.split(".")[0]`

src/utils.ts
```diff
@@ -558,6 +558,20 @@ export function thirtyDaysFromNow(): Date {
...
+
+  if (nodeVersion === null) {
+    throw new Error("Node version is not found");
```
taeold, line 569: Instead of throwing an error, let's return some sane default value like `14`.

taeold (commented):
This is a great start! Thanks for your contribution.

Can we also do the same trick for JS template in addition to typescript template?

oliver-pham:
> This is a great start! Thanks for your contribution.
> 
> Can we also do the same trick for JS template in addition to typescript template?

Sure! I'll do the same thing for JS template after I'm done with the Typescript one.

src/init/features/functions/typescript.js
oliver-pham: Alright, I'll also use boolean value for the other call of `getPackage()`.

src/utils.ts
oliver-pham: Sure, that'll work!

src/utils.ts
oliver-pham: I can add a constant for the default value, i.e. `const DEFAULT_NODE_VERSION = 14`, and return it if it seems good enough for you.

src/init/features/functions/typescript.js
oliver-pham: @taeold I was thinking of using [`isDeprecatedRuntime()`](https://github.com/firebase/firebase-tools/blob/f413eb9eb6940ee20dea748b18bb25ce185e7d7f/src/deploy/functions/runtimes/index.ts#L22) to check for deprecated node version and [`utils.logWarning()`](https://github.com/firebase/firebase-tools/blob/f413eb9eb6940ee20dea748b18bb25ce185e7d7f/src/utils.ts#L174) to display the warning message. Something like this:

```js
if (isDeprecatedRuntime(`nodejs${nodeEngineVersion}`)) {
    utils.logWarning("Your Node version is not supported in Google Cloud Functions.");
}
```

If it's good enough, I'll go ahead and update the code.

src/init/features/functions/typescript.js
oliver-pham: I intend to use `logWarning()` from `utils` to warn users of deprecated Node versions (like this code below):

https://github.com/firebase/firebase-tools/blob/f413eb9eb6940ee20dea748b18bb25ce185e7d7f/src/accountImporter.js#L163-L166

In that case, should I keep it that way?

src/init/features/functions/typescript.js
taeold: 👍🏼  Tiny bug I'd watch out for is that something like `nodejs13`, which is not supported by Google Cloud Functions, may pass the `isDeprecatedRuntime()` check. You might to throw in `isValidRuntime` check somewhere in the code.

oliver-pham:
Hi @taeold, I just wanna circle back to this PR and see if it's still on your radar. Let me know if minor changes need to be made.

PR #3896, opened 2021-11-10
Fix exporting users with newline in display name and custom attributes

### LINKED ISSUE #3881: Newlines in firebase auth:export are not escaped
### Environment info

**firebase-tools:** v9.21.0

**Platform:** macOS

### Test case

Firebase Auth has to be enabled for the issue to appear, but apart from that the issue can be reproduced using the 3 steps below.

### Steps to reproduce

1. Create a new user like this:
```ts
admin.auth().createUser({
  displayName: 'Foo\nBar'
});
```

2. Use the Firebase CLI to export all users to a file:
```
firebase auth:export ./accounts.csv --format CSV --project old-project
```

3. Use the Firebase CLI to import all users from the accounts.csv file:
```
firebase auth:import ./accounts.csv --project=new-project
```

### Expected behavior

The exported users from Auth simply import to a different Firebase project using the CLI.

### Actual behavior

I am getting the following error:
The import fails with an error like this:
Error: Line 2 (...) has invalid data format: Password hash should be base64 encoded.

After investigating the data a bit I noticed some interesting data regarding all name fields:
A user added via Google Login had a newline (\n) in their "Name" (column 6) and "Google Display Name" (column 10), which resulted in the CSV also including a line break!

So the accounts.csv looked something like this:
123123123,foo.bar@example.com,true,passwordHash==,salt==,Foo
Bar,https://example.com/images.png,foo.bar@example.com,Foo
Bar,https://example.com/images.png,,,,,,,,,,,,,1111,1111,,false,,

Because of this new line in the name fields, the resulting CSV is incorrect and the importer assumed that the second line was a new entry even though it still was connected to the previous user.

IchordeDionysos:
### Description

This PR fixes #3881 which is caused by newlines in the display name.
While fixing this PR I also noticed that the custom attributes are not escaped, which leads to similar issues.

Now the display name and custom attribute CSV values are escaped when a `"` or a newline is detected.

### Scenarios Tested

Created a user as described in #3881 with a newline, that is exported and then imported using `auth:export` and `auth:import`.

```bash
firebase auth:export ./accounts.csv --format CSV --project $OLD_PROJECT
```

```bash
firebase auth:import ./accounts.csv    --project=$PROJECT_ID      \  
    --hash-algo=SCRYPT                                                                       \
    --hash-key=$AUTH_HASH_KEY                                                     \
    --salt-separator=$AUTH_HASH_SALT_SEPARATOR                    \
    --rounds=$AUTH_HASH_ROUNDS                                                \
    --mem-cost=$AUTH_HASH_MEM_COSTS
```

### DIFF
--- a/src/accountExporter.js
+++ b/src/accountExporter.js
@@ -34,9 +34,9 @@ var PROVIDER_ID_INDEX_MAP = {
   "github.com": 19,
 };
 
-var _escapeComma = function (str) {
-  if (str.indexOf(",") !== -1) {
-    // Encapsulate the string with quotes if it contains a comma.
+var _escapeCSV = function(str) {
+  if (str.indexOf(",") !== -1 || str.indexOf("\n") !== -1 || str.indexOf("\r") !== -1) {
+    // Encapsulate the string with quotes if it contains a comma or a newline.
     return `"${str}"`;
   }
   return str;
@@ -49,7 +49,7 @@ var _convertToNormalBase64 = function (data) {
 var _addProviderUserInfo = function (providerInfo, arr, startPos) {
   arr[startPos] = providerInfo.rawId;
   arr[startPos + 1] = providerInfo.email || "";
-  arr[startPos + 2] = _escapeComma(providerInfo.displayName || "");
+  arr[startPos + 2] = _escapeCSV(providerInfo.displayName || "");
   arr[startPos + 3] = providerInfo.photoUrl || "";
 };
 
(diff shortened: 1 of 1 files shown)


IchordeDionysos:
Not sure on the tests side, haven't found any tests here :)

IchordeDionysos:
@yuchenshi do you think it would be possible to merge this PR?

bkendall (changes requested):
Can you add a test for this behavior? You can use this as a starting point, since it's the test for handling commas: https://github.com/firebase/firebase-tools/blob/master/src/test/accountExporter.spec.js#L139-L184

Then, add a line to the changelog:

```markdown
- Fixes issue in `auth:export` when handling newlines. (#3881)
```

bkendall:
Sorry for the delay - this slipped through the cracks. I'm certain this is something we'd be happy to include - I left a couple notes for you to address before we can merge it. Thanks!

PR #3901, opened 2021-11-15
Corrects validation of secrets in extension installation

philnash:
### Description

[Discovered while trying to install the Twilio send-messages extension](https://github.com/twilio-labs/twilio-firebase-extensions/issues/13), even with a correctly formatted secret, it never passed the validation rules (though it was added to secrets manager).

The issue was that the methods `promptCreateSecret` and `promptReconfigureSecret` were returning the result of `addNewSecretVersion` which looked like this: `projects/${version.secret.projectId}/secrets/${version.secret.name}/versions/${version.versionId}`. This string then failed to pass the secret validation.

This pull request returns the original string value entered as the secret allowing validation to be performed correctly. Except in the case of choosing not to reconfigure the secret, in which case there is no new secret and the existing secret is deemed valid.

---

This update does not change the fact that an invalid secret will be stored in Secrets Manager before the validation check is done. A bigger refactor of this code could rearrange that order to save on API requests when a secret is invalid. That refactor would also benefit from test cases for select, multiselect and secret types on top of the existing string type.


i14h:
@pavelgj since Joe is out today

philnash:
@pavelgj I've updated this so that `promptCreateSecret` and `promptReconfigureSecret` handle the validation of the response within themselves, which simplifies `askForParam`. Both `promptCreateSecret` and `promptReconfigureSecret` both respond with the secret ID that is created (which I think would have broken things before).

My only question is what is supposed to happen when you are reconfiguring and you choose to leave a secret as it is? You don't seem to have access to the existing secret ID and returning an empty string feels wrong (though is what I have implemented for now).

src/test/extensions/askUserForParam.spec.ts
```diff
@@ -198,7 +200,7 @@ describe("askUserForParam", () => {
       expect(res).to.equal("");
     });
   });
-  describe("askForParam", () => {
+  describe("askForParam with string param", () => {
```
pavelgj, line 203: undo?

src/extensions/askUserForParam.ts
```diff
@@ -157,23 +160,33 @@ async function promptReconfigureSecret(
...
+        } else {
+          return promptReconfigureSecret(projectId, instanceId, paramSpec);
+        }
+      } else {
+        return "";
       }
-      return addNewSecretVersion(projectId, instanceId, secret, paramSpec, secretValue);
     case SecretUpdateAction.LEAVE:
     default:
-      return paramSpec.default || "";
```
pavelgj, line 176: I believe this should be `paramSpec.default || ""` -- `default` contains previous value, which we want in the `LEAVE` case.

pavelgj (dismissed):

src/test/extensions/askUserForParam.spec.ts
philnash: This was to contrast with the [new test below "askForParam with secret param"](https://github.com/firebase/firebase-tools/pull/3901/files#diff-3e09552fce53dd0c656d89aaaea5adcffb1e8b33316dbfd5332afbf35fce1a9bR223).

src/extensions/askUserForParam.ts
philnash: Thanks! Updated.

src/extensions/askUserForParam.ts
```diff
@@ -113,11 +113,15 @@ export async function askForParam(
             "You may select multiple options.",
           choices: convertExtensionOptionToLabeledList(paramSpec.options as ParamOption[]),
         });
+        valid = checkResponse(response, paramSpec);
         break;
       case ParamType.SECRET:
-        response = reconfiguring
-          ? await promptReconfigureSecret(projectId, instanceId, paramSpec)
-          : await promptCreateSecret(projectId, instanceId, paramSpec);
+        if (reconfiguring) {
```
joehan, line 119: Nit: Prefer this as a ternary, like it was before.

joehan (approved):
One nite, but code LGTM. Thanks for fixing this!

joehan:
Could you also add an entry to CHANGELOG.md describing this fix? Something like:
- Fixes issue where secrets for Firebase Extensions would be created before validation.

philnash:
Updated those two bits @joehan.

philnash:
@joehan Is this good to be merged now? Anything else you need from me?

joehan:
@philnash Nope, we had a code freeze on for a while (and then this slipped past me for a few days 😬). This looks good to go - thanks for your contribution!

PR #3911, opened 2021-11-18
Turn on AR experiment for all clients.

inlined:
After some limited internal testing, we believe that we're ready for
a rollout that is only controlled by the server. The code has to
be ready to always request AR but sometimes be on GCR. This means
that we clean up GCR images and also that we handle 404s from AR.


bkendall (approved):

src/deploy/functions/containerCleaner.ts
```diff
@@ -74,42 +74,43 @@ export async function cleanupBuildImages(
...
-          await gcrCleaner.cleanupFunction(func);
-        } catch (err) {
-          const path = `${func.project}/${SUBDOMAIN_MAPPING[func.region]}/gcf`;
-          failedDomains.add(`https://console.cloud.google.com/gcr/images/${path}`);
-        }
-      })
-    );
-  }
+  const cleanup: Array<Promise<void>> = [];
+  const arCleaner = cleaners.ar || new ArtifactRegistryCleaner();
```
taeold, line 78: *meta - keep reading this as `arsCleaner` and chuckling about it...

taeold (approved):

PR #3938, opened 2021-12-09
Merge Provider Metrics feature branch into master

elvisun:
All commits already reviewed. Originally on the private repo


src/api.js
```diff
@@ -99,6 +99,10 @@ var api = {
     "FIREBASE_CLOUDLOGGING_URL",
     "https://logging.googleapis.com"
   ),
+  cloudMonitroingOrigin: utils.envOverride(
```
LanderBeeuwsaert, line 102: is this correct? => cloudMonitroingOrigin

elvisun: Thanks, updated

src/extensions/extensionsApi.ts
```diff
@@ -601,6 +601,24 @@ export async function listExtensionVersions(
...
+  projectId: string,
+  publisherId?: string
+): Promise<PublisherProfile> {
+  const res = await api.request("GET", `/${VERSION}/projects/${projectId}/publisherProfile`, {
+    auth: true,
+    origin: api.extensionsOrigin,
+    query: {
+      publisherId,
+    },
+  });
```
bkendall, line 618: Please prefer using `apiv2` rather than `api`. It would be great to not add any more usage of `api.request`. Thanks.

src/test/gcp/cloudmonitoring.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  };
+
+  it("should make a POST call to the correct endpoint", async () => {
+    nock(api.cloudMonitoringOrigin)
+      .get(`/${CLOUD_MONITORING_VERSION}/projects/${PROJECT_NUMBER}/timeSeries/`)
+      .query(true)
+      .reply(200, RESPONSE);
+
+    const res = await queryTimeSeries(query, PROJECT_NUMBER);
+    expect(res).to.deep.equal(RESPONSE.timeSeries);
```
bkendall, line 36: May want to assert that `nock` was indeed called. An empty array response resulting in an empty returned value could be coincidence (or programmer error)

src/test/gcp/cloudmonitoring.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  });
+
+  it("should throw a FirebaseError if the endpoint returns an error response", async () => {
+    nock(api.cloudMonitoringOrigin)
+      .post(`/${CLOUD_MONITORING_VERSION}/projects/${PROJECT_NUMBER}/timeSeries/`)
+      .query(true)
+      .reply(404);
+    await expect(queryTimeSeries(query, PROJECT_NUMBER)).to.be.rejectedWith(FirebaseError);
```
bkendall, line 44: I'd make sure `nock` was called here too!

src/extensions/extensionsApi.ts
elvisun: Done. Wasn't aware of this effort, might be a good idea to mark the old api as deprecated.

src/test/gcp/cloudmonitoring.spec.ts
elvisun: Done

src/test/gcp/cloudmonitoring.spec.ts
elvisun: Done, great call, actually caught a bug in the test! (had `post` rather than `get`)

huangjeff5 (approved):

PR #3940, opened 2021-12-09
Added finer grained metrics to ext:install.

apascal07:
Tested locally with live view, seems to work but will need to check the labels once GA populates them in the regular view.

### DIFF
--- a/src/commands/ext-install.ts
+++ b/src/commands/ext-install.ts
@@ -243,7 +243,8 @@ async function infoInstallBySource(
 }
 
 async function infoInstallByReference(
-  extensionName: string
+  extensionName: string,
+  interactive: boolean
 ): Promise<extensionsApi.ExtensionVersion> {
   // Infer firebase if publisher ID not provided.
   if (extensionName.split("/").length < 2) {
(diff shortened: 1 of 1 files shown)


src/commands/ext-install.ts
```diff
@@ -302,8 +305,10 @@ export default new Command("ext:install [extensionName]")
     // If the user types in URL, or a local path (prefixed with ~/, ../, or ./), install from local/URL source.
     // Otherwise, treat the input as an extension reference and proceed with reference-based installation.
     if (isLocalOrURLPath(extensionName)) {
+      track("ext:install", "Install by Source");
       source = await infoInstallBySource(projectId, extensionName);
     } else {
+      track("ext:install", "Install by Extension Ref");
```
joehan, line 311: Let's use something other than 'ext:install' for the action here - maybe 'Extension Installed'?. The default command tracking uses that same action, so we'll be double counting installs if we do this.

joehan: Also, this may be a bit hacky, but we could  combine these 2 metrics into one and track interactive vs non interactive in the duration field, by passing in 1 for interactive mode and 0 for noninteractive. This would let us see interactive vs noninteractive for each label.

joehan (changes requested):

joehan (approved):
Sorry for the delay, forgot to come back to this with the holidays

apascal07:
No worries! Thanks for the review.

PR #3943, opened 2021-12-13
Do not try to clean up AR images if AR is not enabled

inlined:
🤬 This was supposed to go in the last release.

VS code had some unsaved files and the last release has a bug in it as a result =/


bkendall (approved):
run `npm run format` on the codebase and that should fix the formatting error

PR #3927, opened 2021-12-01
Add and verify OIDC/SAML flows

lisajian:
### Description

Includes assertions thrown for generic OIDC/SAML IdP providers that parallel GCIP behavior. Verified that existing `signInWithIdp` flows work against JSSDK Auth Demo page.

Corresponding internal bug: b/192387796

### Scenarios Tested

* `npm test` passes
* See videos included in https://github.com/firebase/firebase-js-sdk/pull/5776

### Sample Commands

N/A

PR #3946, opened 2021-12-14
Fill in SAMLResponse nameId for pre-populated users

lisajian:
### Description

A bit of cleanup from https://github.com/firebase/firebase-tools/pull/3927. Widget should fill in the SAMLResponse correctly even for users that were signed in previously.

Corresponding internal bug: b/192387796

### Scenarios Tested

Manually tested against the JSSDK Auth Demo page. See instructions [here](https://github.com/firebase/firebase-js-sdk/tree/master/packages/auth/demo#running-against-auth-emulator)

### Sample Commands

N/A

### DIFF
--- a/src/emulator/auth/handlers.ts
+++ b/src/emulator/auth/handlers.ts
@@ -201,7 +201,7 @@ export function registerHandlers(
           <span class="mdc-list-item__text"><span class="mdc-list-item__primary-text">${
             info.displayName || "(No display name)"
           }</span>
-          <span class="mdc-list-item__secondary-text fallback-secondary-text">${
+          <span class="mdc-list-item__secondary-text fallback-secondary-text" id="reuse-email">${
             info.email || ""
           }</span>
       </li>`
--- a/src/emulator/auth/widget_ui.ts
+++ b/src/emulator/auth/widget_ui.ts
@@ -176,7 +176,7 @@ function finishWithUser(urlEncodedIdToken) {
 
   // Save reasonable defaults for SAML providers
   if (isSamlProvider) {
-    var email = document.getElementById('email-input').value;
+    var email = document.getElementById('email-input').value || document.getElementById('reuse-email').innerText;
     url += '&SAMLResponse=' + encodeURIComponent(JSON.stringify({
       assertion: {
         subject: {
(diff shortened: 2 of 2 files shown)


src/emulator/auth/handlers.ts
```diff
@@ -201,7 +201,7 @@ export function registerHandlers(
           <span class="mdc-list-item__text"><span class="mdc-list-item__primary-text">${
             info.displayName || "(No display name)"
           }</span>
-          <span class="mdc-list-item__secondary-text fallback-secondary-text">${
+          <span class="mdc-list-item__secondary-text fallback-secondary-text" id="reuse-email">${
```
yuchenshi, line 204: This ID is not unique because there will be one element for each existing `providerInfo`. Please find some other way of extracting the email, for example by modifying the logic that parses `data-id-token`

yuchenshi (changes requested):

src/emulator/auth/handlers.ts
yuchenshi: I think a more reliable way is to modify `finishWithUser` to take an additional email argument and modify https://github.com/firebase/firebase-tools/blob/da585894ef1e226ffed49af7b0dca6987cba9b25/src/emulator/auth/widget_ui.ts#L160 and https://github.com/firebase/firebase-tools/blob/da585894ef1e226ffed49af7b0dca6987cba9b25/src/emulator/auth/widget_ui.ts#L237 to pass in the email.

src/emulator/auth/widget_ui.ts
```diff
@@ -234,7 +234,7 @@ document.getElementById('main-form').addEventListener('submit', function(e) {
     if (screenName) claims.screenName = screenName;
     if (photoUrl) claims.photoUrl = photoUrl;
 
-    finishWithUser(createFakeClaims(claims));
+    finishWithUser(createFakeClaims(claims), email);
```
yuchenshi, line 237:
```suggestion
    finishWithUser(createFakeClaims(claims), claims.email);
```

To make it default to `undefined` (omitted from JSON output) instead of `""`

yuchenshi (approved):

src/emulator/auth/widget_ui.ts
lisajian: Sick, done!

Issue #3709, opened 2021-08-26
Can't deploy to european region with organisation policy


LukasCornille:
### [REQUIRED] Environment info

**firebase-tools:** 9.16.6

**Platform:** macOS

### [REQUIRED] Test case

Any test project works for this case I believe, as long as the organisation policies are set to only deploy to any EU location.

I have this function:
```ts
import * as functions from 'firebase-functions';
import sendToMailUp from './services/mailup';

// Empty interface because it's irrelevant for this example
interface User {}

const addUserToMailup = functions
  .region('europe-west1')
  .database.ref('/users')
  .onCreate(async ({ val }) => {
    const user = val() as User;

    return sendToMailUp(user);
  });

export default addUserToMailup;
```

which then gets import in the index.ts file as so:

```ts
import admin from 'firebase-admin';
import addUserToMailup from './addNewUserToMailup';

admin.initializeApp();

exports.addUserToMailup = addUserToMailup;
```

### [REQUIRED] Steps to reproduce

I've ran this with:

```console
firebase deploy --only functions

firebase deploy --only functions:addUserToMailup
```

### [REQUIRED] Expected behavior

I would expect this function to be deployed on the europe-west1 region

### [REQUIRED] Actual behavior

As you can see from the debug log, when trying to receive the location it will try to upload the function on the us-central1 region, which is not allowed by a policy from the company I am working for.

[firebase-debug.log](https://github.com/firebase/firebase-tools/files/7059799/firebase-debug.log)

joehan:
Hi @LukasCornille - you can control which region your function code is uploaded to by setting the `FIREBASE_FUNCTIONS_UPLOAD_REGION` environment variable. From your terminal/.bash_profile, run:
```
export FIREBASE_FUNCTIONS_UPLOAD_REGION = eu-west4 
## or any other region that works for you
```

@inlined has also been talking about changing this to always use one of the regions you're deploying your functions to, so adding him for visibility.

inlined:
This might be a good action item for CLI version 10. Will add it to our internal tracker for breaking changes.

PR #3950, opened 2021-12-15
Match GCFv1 source upload region to the function deploy region.

taeold:
Today, GCFv1 source are uploaded to `us-central1` unless overridden via `FIREBASE_FUNCTIONS_UPLOAD_REGION` environment variable.

We change the deploy logic to upload source to the region where functions our deployed. When user's function source contains functions in more than one regions, we'll upload the source in the "first" region, whatever first happens to be in the list of functions.

We also get rid of `FIREBASE_FUNCTIONS_UPLOAD_REGION` , as that env var is no longer very useful.

This is a breaking change in the behavior of function deploys and will only be released for firebase cli v10 and upward.

Fixes https://github.com/firebase/firebase-tools/issues/3709

### DIFF
--- a/src/deploy/functions/deploy.ts
+++ b/src/deploy/functions/deploy.ts
@@ -63,6 +63,7 @@ export async function deploy(
 
     const v1Endpoints = backend.allEndpoints(want).filter((e) => e.platform === "gcfv1");
     if (v1Endpoints.length > 0) {
+      // Choose one of the function region for source upload.
       uploads.push(uploadSourceV1(context, v1Endpoints[0].region));
     }
 
--- a/src/deploy/functions/deploy.ts
+++ b/src/deploy/functions/deploy.ts
@@ -18,7 +18,7 @@ setGracefulCleanup();
 
 async function uploadSourceV1(context: args.Context, region: string): Promise<void> {
   const uploadUrl = await gcf.generateUploadUrl(context.projectId, region);
-  context.uploadUrl = uploadUrl;
+  context.sourceUrls = { ...context.sourceUrls, region: uploadUrl };
   const uploadOpts = {
     file: context.functionsSourceV1!,
     stream: fs.createReadStream(context.functionsSourceV1!),
--- a/src/deploy/functions/deploy.ts
+++ b/src/deploy/functions/deploy.ts
@@ -12,20 +12,21 @@ import * as gcfv2 from "../../gcp/cloudfunctionsv2";
 import * as utils from "../../utils";
 import * as backend from "./backend";
 
-const DEFAULT_UPLOAD_REGION = process.env.FIREBASE_FUNCTIONS_UPLOAD_REGION;
-
 setGracefulCleanup();
 
 async function uploadSourceV1(context: args.Context, region: string): Promise<void> {
-  const uploadUrl = await gcf.generateUploadUrl(context.projectId, region);
-  context.sourceUrls = { ...context.sourceUrls, [region]: uploadUrl };
-  const uploadOpts = {
-    file: context.functionsSourceV1!,
-    stream: fs.createReadStream(context.functionsSourceV1!),
-  };
-  await gcs.upload(uploadOpts, uploadUrl, {
-    "x-goog-content-length-range": "0,104857600",
-  });
+  if (context.sourceUrl == undefined) {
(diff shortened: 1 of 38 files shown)


src/deploy/functions/deploy.ts
```diff
@@ -13,13 +12,13 @@ import * as gcfv2 from "../../gcp/cloudfunctionsv2";
 import * as utils from "../../utils";
 import * as backend from "./backend";
 
-const GCP_REGION = functionsUploadRegion;
+const DEFAULT_UPLOAD_REGION = process.env.FIREBASE_FUNCTIONS_UPLOAD_REGION;
```
inlined, line 15: Do we actually need to have this env anymore?

src/deploy/functions/deploy.ts
```diff
@@ -63,11 +62,12 @@ export async function deploy(
   try {
     const want = payload.functions!.backend;
     const uploads: Promise<void>[] = [];
-    if (backend.allEndpoints(want).some((endpoint) => endpoint.platform === "gcfv1")) {
-      uploads.push(uploadSourceV1(context));
-    }
 
     for (const region of Object.keys(want.endpoints)) {
+      if (backend.allEndpoints(want).some((endpoint) => endpoint.platform === "gcfv1")) {
```
inlined, line 67: This doesn't do what you think it does. This will upload to every region if any region has GCF v1.

src/deploy/functions/release/fabricator.ts
```diff
@@ -45,7 +45,7 @@ export interface FabricatorArgs {
   appEngineLocation: string;
 
   // Required if creating or updating any GCFv1 functions
-  sourceUrl?: string;
+  sourceUrls?: Record<string, string>;
```
inlined, line 48: We can probably just stick with one region. We had to use multiple regions in GCFv2 because GCF threatened to block cross-region installs. In GCFv1 I don't think they can make that change retroactively. And we're not violating data residency if we know they'll need to store functions in that region anyway.

inlined (changes requested):

src/deploy/functions/deploy.ts
taeold: Which is what I want (though I can now see that we can shortcut the uploading if we've uploaded source to a region already). e.g.

```
export const fn-us = functions.region('us-central1').https...
export const fn-eu = functoins.region('europe-west1').https... 
```

Then don't we want to deploy the source to both us-central and europe-west?

src/deploy/functions/deploy.ts
taeold: Thought the env var provided a config knobs to existing users so it is less breaking of a change.

src/deploy/functions/release/fabricator.ts
taeold: But which region given the possibility of multiple regions on a function source (ties nicely w/ the analytics PR update you made earlier)?

src/deploy/functions/deploy.ts
inlined: I'd start with the assumption that people can always upload to the region where they're deploying. I can't fathom a reason they'd need a different region.

src/deploy/functions/deploy.ts
inlined: If you have a v1 function in US-central1 and a v2 function in us-west1 this will upload v1 source to both regions. I think you want:

```
if (want.endpoints[region].some(endpoint => endpoint.platform === "gcfv1"))
```

src/deploy/functions/release/fabricator.ts
inlined: The analytics PR tracks the regions we deploy to, not the regions we upload to. We _can_ make this multi-regional if you really believe in it, but it seems like

A) We're not going to run into data residency problems since the dev clearly has access to that region
B) We're going to reduce bandwidth from the user to not upload X MB to Y regions whereas we used to only upload X MB

If you really believe in per-region uploads though I won't stand in your way.

taeold: I thought we had no choice but to have per-region uploads, but I see now that I had unnecessary strict requirements for myself. Happy to not go down that path unless we change our minds later.

joehan (approved):

inlined (approved):

Issue #3934, opened 2021-12-07
Parsing trouble for non-required extension parameters


alallema:
### [REQUIRED] Environment info

**firebase-tools:** 9.23.0

**Platform:** macOS

### [REQUIRED] Test case

I'm not sure if it's a regression, but I found this [issue](https://github.com/firebase/firebase-tools/issues/3126) that seems pretty similar.
Installs an extension installation or emulator with the `--param` file as an argument. If there are optional parameters that are empty in the file, the next line will be read.
This issue seems to be present in the `v9.22.0` but not in the `v9.21.0`

### [REQUIRED] Steps to reproduce

*This is the `extension.yml` file:*
```yml
  - param: FIELDS_TO_INDEX
    label: Fields to index in Meilisearch
    description: >-
        What are the names of the fields you want to be indexed in MeiliSearch?
        Create a comma-separated list of the fields name, or leave blank to take all fields.
        The id field is always indexed even when omitted in the list.
    example: 'title,overview,release_date'
    validationRegex: "^[^,]?[a-zA-Z-_0-9,]*[^,]$"
    validationErrorMessage:
      Fields must be given through a comma-separated list.
    default: ''
    required: false

  - param: SEARCHABLE_FIELDS
    label: Fields in which to search. - Optional
    description: >-
      What are the names of the fields you want to make searchable in MeiliSearch?
      This features is optional. See [the documentation about it](https://docs.meilisearch.com/reference/features/field_properties.html#searchable-fields)
      for more information.
      Create a comma-separated list of fields, or leave blank to include all fields.
    example: 'title,overview,release_date'
    validationRegex: "^[^,]?[a-zA-Z-_0-9,]*[^,]$"
    validationErrorMessage:
      Fields must be given through a comma-separated list.
    default: ''
    required: false
```

*This is the `param.env` file:*
```
LOCATION=us-central1
COLLECTION_PATH=movies
FIELDS_TO_INDEX=
SEARCHABLE_FIELDS=
MEILISEARCH_INDEX_NAME=movies
MEILISEARCH_HOST='http://127.0.0.1:7700'
MEILISEARCH_API_KEY='masterKey'
```

```bash
firebase ext:install <extension-name> --params=params.env --project=<project-id>
```

### [REQUIRED] Expected behavior

The optional parameter does not have to be filled in.

### [REQUIRED] Actual behavior

**This is a print of `process.env`**
```bash
[...]

PR #3951, opened 2021-12-15
Fixing a bug where empty variables in .env files were read as multi-line values (#3934)

joehan:
### Description
Fixes an issue where the .env parser would mishandled empty values. The following .env file:
```
FOO=
BAR=something
```
would be parsed into
```
{ FOO: "BAR=something" }
```
Fixes https://github.com/firebase/firebase-tools/issues/3934

### Scenarios Tested
Added a test case to cover this

### DIFF
--- a/src/functions/env.ts
+++ b/src/functions/env.ts
@@ -49,7 +49,7 @@ const LINE_RE = new RegExp(
   "(" +                      //   begin optional value
   "\\s*'(?:\\\\'|[^'])*'|" + //     single quoted or
   '\\s*"(?:\\\\"|[^"])*"|' + //     double quoted or
-  "[^#\\r\\n]+" +            //     unquoted
+  "[^#=\\r\\n]*" +           //     unquoted
   ")?" +                     //   end optional value
   "\\s*" +                   //   trailing whitespaces
   "(?:#[^\\n]*)?" +          //   optional comment
@@ -109,6 +109,7 @@ export function parse(data: string): ParseResult {
         v = v.replace(/\\([\\'"])/g, "$1");
       }
     }
+
     envs[k] = v;
   }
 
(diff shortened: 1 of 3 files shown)


src/test/functions/env.spec.ts
```diff
@@ -135,26 +135,33 @@ BAR=bar
...
+      //       {
+      //         description: "should ignore empty lines",
+      //         input: `
+      // FOO=foo
+
+      // BAR=bar
+
+      // `,
+      //         want: { FOO: "foo", BAR: "bar" },
+      //       },
```
taeold, line 164: Do we want to uncomment this?

src/functions/env.ts
```diff
@@ -49,7 +49,7 @@ const LINE_RE = new RegExp(
   "(" +                      //   begin optional value
   "\\s*'(?:\\\\'|[^'])*'|" + //     single quoted or
   '\\s*"(?:\\\\"|[^"])*"|' + //     double quoted or
-  "[^#\\r\\n]+" +            //     unquoted
+  "[^#=\\r\\n]*" +           //     unquoted
```
taeold, line 52: q: Did we need `=`?

IIUC, we are making something like

```
FOO=bar=a
```

from being a valid entry (`{FOO: "bar=a"}`) to an invalid one.

taeold (changes requested):

src/test/functions/env.spec.ts
joehan: 🤦  Forgot to save the file before commiting

src/functions/env.ts
joehan: Hmmm, its a bit of a weird case - i guess what we really want to do is allow =, but not after a line break?
So something like ```
[[^#\\r\\n][^#=\\r\\n]]
```?
There may be other regex tricks that could solve this as well - maybe replacing the `\\s` that. matches whitespace after the equals with something that doesn't match newlines?

joehan: Ahhhh, i just read your original code comments more closely, and realized we explicitly don't support multiline unquoted values. I think i know how to fix this, will play around a bit and make sure all the tests work.

taeold (approved):

PR #3957, opened 2021-12-16
upgrade typescript to 4.5

bkendall:
### Description

Upgraded typescript to 4.5 and fixed all issues.

### DIFF
--- a/src/commands/ext-dev-register.ts
+++ b/src/commands/ext-dev-register.ts
@@ -35,7 +35,7 @@ export default new Command("ext:dev:register")
     });
     try {
       await registerPublisherProfile(projectId, publisherId);
-    } catch (err) {
+    } catch (err: any) {
       if (err.status === 409) {
         const error =
           `Couldn't register the publisher ID '${clc.bold(publisherId)}' to the project '${clc.bold(
@@ -52,8 +52,7 @@ export default new Command("ext:dev:register")
       throw new FirebaseError(
         `Failed to register publisher ID ${clc.bold(publisherId)} for project ${clc.bold(
           projectId
-        )}: ${err.message}`,
-        { exit: 1 }
+        )}: ${err.message}`
       );
     }
     return utils.logLabeledSuccess(
--- a/src/emulator/functionsEmulatorRuntime.ts
+++ b/src/emulator/functionsEmulatorRuntime.ts
@@ -53,7 +53,7 @@ function requireAsync(moduleName: string, opts?: { paths: string[] }): Promise<a
   return new Promise((res, rej) => {
     try {
       res(require(require.resolve(moduleName, opts))); // eslint-disable-line @typescript-eslint/no-var-requires
-    } catch (e) {
+    } catch (e: any) {
       rej(e);
     }
   });
@@ -63,7 +63,7 @@ function requireResolveAsync(moduleName: string, opts?: { paths: string[] }): Pr
   return new Promise((res, rej) => {
     try {
       res(require.resolve(moduleName, opts));
-    } catch (e) {
+    } catch (e: any) {
       rej(e);
     }
   });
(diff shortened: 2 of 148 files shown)


src/commands/ext-dev-register.ts
```diff
@@ -52,8 +52,7 @@ export default new Command("ext:dev:register")
       throw new FirebaseError(
         `Failed to register publisher ID ${clc.bold(publisherId)} for project ${clc.bold(
           projectId
-        )}: ${err.message}`,
-        { exit: 1 }
```
yuchenshi, line 56: Is removing the exit code intentional?

src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -166,7 +166,7 @@ class Proxied<T extends ProxyTarget> {
       },
       apply: (target, thisArg, argArray) => {
         if (this.appliedValue) {
-          return this.appliedValue.apply(thisArg, argArray);
+          return this.appliedValue.apply(thisArg);
```
yuchenshi, line 169: Is this intentional? This will remove all the positional arguments

src/test/management/apps.spec.ts
```diff
@@ -469,7 +469,7 @@ describe("App management", () => {
       const appCounts = 10;
       const expectedAppList = generateIosAppList(appCounts);
       const apiResponseAppList = expectedAppList.map((app) => {
-        const iosApp = { ...app };
+        const iosApp: any = { ...app };
```
yuchenshi, line 472: Would you mind leaving a TODO to clean this up?

src/test/management/apps.spec.ts
```diff
@@ -488,7 +488,7 @@ describe("App management", () => {
       const appCounts = 10;
       const expectedAppList = generateAndroidAppList(appCounts);
       const apiResponseAppList = expectedAppList.map((app) => {
-        const androidApps = { ...app };
+        const androidApps: any = { ...app };
```
yuchenshi, line 491: Ditto

src/test/management/apps.spec.ts
```diff
@@ -507,7 +507,7 @@ describe("App management", () => {
       const appCounts = 10;
       const expectedAppList = generateWebAppList(appCounts);
       const apiResponseAppList = expectedAppList.map((app) => {
-        const webApp = { ...app };
+        const webApp: any = { ...app };
```
yuchenshi, line 510: Ditto

yuchenshi (approved):
try { approve(pr) } catch (err: any) {}

src/commands/ext-dev-register.ts
bkendall: Yes - sine I was cleaning up the block, I removed some `exit: 1`s that I saw - that's the default value and doesn't need to be provided.

src/emulator/functionsEmulatorRuntime.ts
bkendall: Yes, but only because appliedValue is defined as `() => any` - it doesn't (in theory) have any arguments. The typing got more strict around this. Let me see if I can just make it accept anything, that might be the safer route (but unsure how much work it will be)

bkendall: Ah, yeah, okay. I fixed it, I think. Thanks for flagging.

src/test/management/apps.spec.ts
bkendall: Done.

src/test/management/apps.spec.ts
bkendall: Done.

src/test/management/apps.spec.ts
bkendall: Done.

PR #3958, opened 2021-12-16
upgrade npm in node12 container for building

bkendall:
### Description

`npm` needs to be updated in our building container in order for it to work!

Just released 10.0.1 with this change, so it works great!


joehan (approved):
LGTM, tho I'm surprised that there isn't a base image with node 12 & npm 8 that we can use instead of upgrading.

Issue #3509, opened 2021-06-18
Failed to authenticate, have you run firebase login?


Liker777:
I cross reference this [stackoverflow question](https://stackoverflow.com/questions/68028795/failed-to-authenticate-have-you-run-firebase-login) also since it is not resolved yet. 

**firebase-tools:**

**Platform:**

### [REQUIRED] Test case

I am following instructions for Firebase functions. I have successful login from my console, however when I am trying to deploy functions, terminal gives an error: enter image description here

```
oleg@oleg-HP-EliteBook-8460p:~/Desktop/Oleg/projects/firebase-functions$ firebase login

Already logged in as [my user email]

oleg@oleg-HP-EliteBook-8460p:~/Desktop/Oleg/projects/firebase-functions$ sudo firebase deploy --only functions

Error: Failed to authenticate, have you run firebase login? (node:11224) MetadataLookupWarning: received unexpected error = Unsuccessful response status code. Request failed with status code 403 code = 403 Having trouble? Try firebase [command] --help
```

So the error is kind of controversial. Firstly, Firebase says that I am already successfully logged in. Right next, Firebase says that I am not logged in... I suspect there should be some setting for user name or something like that in kind of local firebase.json file or other files, but I didn't find it.. Or should I change something in the Firebase Console?

Liker777:
I filed the fields of the template by they disappear for some reason..

samtstern:
@Liker777 the difference is that you're running `firebase login` and then `sudo firebase deploy`. You should never need to use `sudo` with Firebase.

Liker777:
> @Liker777 the difference is that you're running `firebase login` and then `sudo firebase deploy`. You should never need to use `sudo` with Firebase.

Thanks a lot! Without sudo, it works differently.

PR #3959, opened 2021-12-17
CF3 Secrets Support

taeold:
Support deploying secret environment variables on a function.

Prior to deploying functions with secret configuration, the CLI will run somewhat comprehensive validation to ensure that secret config will work when deployed, e.g.

1. Secret version exists.
2. Secret version is in ENABLED state.
3. Secret version can be access by the runtime service account.

We do this since the GCF doesn't do the same level of validation and instead repeatedly fail to spin up a new instance with an invalid secret config. This often results on super long deploys (probably until some master timeout is met for function instance deploy).

I took the opportunity to refactor the code a little to group various "ensure" and "validate" used in function deploys in their own files.

Emulator support for secrets will come in a separate PR.


src/deploy/functions/runtimes/node/parseTriggers.ts
```diff
@@ -219,6 +224,41 @@ export function addResourcesToBackend(
...
+          if (secret.includes("/versions/")) {
+            const secretVersion = parseSecretVersionResourceName(secret);
+            secretEnv.version = secretVersion.version;
+          }
+          secretEnvs.push(secretEnv);
+        } else {
+          // Assume secret resource id (e.g. MY_API_KEY[@VERSION]).
+          const [key, version] = secret.split("@");
+          const secretEnv: backend.SecretEnvVar = {
+            key,
```
colerogers, line 249: It looks like we set `key` & `secret` to the same value in both if cases, why do we need both of them?

src/deploy/functions/prepare.ts
```diff
@@ -180,6 +182,126 @@ export async function prepare(
...
+  await ensureSecretAccess(matchingBackend);
+}
+
+/**
+ * Validate secret environment variables setting, if any.
+ * A bad secret configuration can lead to a significant delay in function deploys.
+ *
+ * If validation fails for any secret config, throws a FirebaseError.
+ */
+export async function validateSecrets(b: backend.Backend) {
```
colerogers, line 195: nit: return type

src/deploy/functions/prepare.ts
```diff
@@ -180,6 +182,126 @@ export async function prepare(
...
+    .allEndpoints(b)
+    .filter((e) => e.secretEnvironmentVariables && e.secretEnvironmentVariables.length > 0);
+  validatePlatformTargets(endpoints);
+  await validateSecretVersions(endpoints);
+}
+
+/**
+ * Ensures that all endpoints specifying secret environment variables target platform that supports the feature.
+ */
+function validatePlatformTargets(endpoints: backend.Endpoint[]) {
```
colerogers, line 206: nit: return type

src/deploy/functions/prepare.ts
```diff
@@ -180,6 +182,126 @@ export async function prepare(
...
+}
+
+/**
+ * Validate each secret version referenced in target endpoints.
+ *
+ * A secret version is valid if:
+ *   1) It exists.
+ *   2) It's in state "enabled".
+ */
+async function validateSecretVersions(endpoints: backend.Endpoint[]) {
```
colerogers, line 225: nit: return type

src/deploy/functions/prepare.ts
```diff
@@ -180,6 +182,126 @@ export async function prepare(
...
+  }
+}
+
+/**
+ * Ensures that runtime service account has access to the secrets.
+ *
+ * To avoid making more than one simultaneous call to setIamPolicy calls per secret, the function batches all
+ * service account that requires access to it.
+ */
+export async function ensureSecretAccess(b: backend.Backend) {
```
colerogers, line 265: nit: return type

src/deploy/functions/prepare.ts
```diff
@@ -180,6 +182,126 @@ export async function prepare(
...
+    }
+  }
+
+  const ensure = [];
+  for (const [projectId, secrets] of Object.entries(toEnsure)) {
+    for (const [secret, serviceAccounts] of Object.entries(secrets)) {
+      ensure.push(ensureAccess(projectId, secret, Array.from(serviceAccounts)));
+    }
+  }
+  await Promise.all(ensure);
```
colerogers, line 304: Personal preference but I would move all these secret helper functions into their own file

src/deploy/functions/backend.ts
```diff
@@ -164,10 +164,18 @@ export interface TargetIds {
   project: string;
 }
 
+export interface SecretEnvVar {
+  key: string;
```
colerogers, line 168: We don't seem to use `key` at all, is this necessary?

colerogers (commented):
Yay secrets 🔐

src/deploy/functions/runtimes/node/parseTriggers.ts
taeold: For context, key = name of the environment variable and secret is the name of the secret.

We happen to enforce that `key == secret` in CF3, but I don't think we necessarily need that to be case for all backend. I'm leaning into "manifest should be general, our SDK should be opinionated" the idea.

E.g. if you are hand-writing stack.yaml, I think it's nice to have the flexibility of renaming a secret name to a different env var key.

WDYT?

src/deploy/functions/prepare.ts
taeold: Agreed.

src/deploy/functions/backend.ts
taeold: Being discussed on another thread.

src/deploy/functions/prepare.ts
taeold: After reading few more TS tutorirals, I don't think typing `Promise<void>` adds to much value here, especially since TS can easily figure it out implicitly. WDYT?

src/deploy/functions/prepare.ts
taeold: Being discussed in another thread.

src/deploy/functions/prepare.ts
taeold: Being discussed in another thread.

src/deploy/functions/prepare.ts
taeold: Being discussed in another thread.

src/deploy/functions/prepare.ts
colerogers: Sure that's fine with me, but the GitHub actions linter thing still throws a warning

src/deploy/functions/runtimes/node/parseTriggers.ts
colerogers: Yeah I like that take, thanks for clarifying!

colerogers (approved):

src/deploy/functions/ensure.ts
```diff
@@ -0,0 +1,131 @@
+import * as clc from "cli-color";
+
+import { ensure } from "../../ensureApiEnabled";
+import { FirebaseError, isBillingError } from "../../error";
+import { logLabeledBullet, logLabeledSuccess } from "../../utils";
+import { ensureServiceAgentRole } from "../../gcp/secretManager";
+import { defaultServiceAccount } from "../../gcp/cloudfunctions";
```
inlined, line 7: You're going to need a v1 and v2 here. The default service account for v1 functions is the GAE service account and for v2 functions is the compute service account.

src/deploy/functions/ensure.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    throw e;
+  }
+}
+
+// We previously force-enabled AR. We want to wait on this to see if we can give
+// an upgrade warning in the future. If it already is enabled though we want to
+// remember this and still use the cleaner if necessary.
+export async function maybeEnableAR(projectId: string): Promise<boolean> {
+  if (previews.artifactregistry) {
```
inlined, line 73: Uh oh.... isn't this backwards? If the experiment is enabled we should ensure and otherwise we should check.

src/deploy/functions/ensure.ts
```diff
@@ -0,0 +1,131 @@
...
+      "functions",
+      `ensured ${clc.bold(serviceAccounts.join(", "))} access to ${clc.bold(secret)}.`
+    );
+  };
+
+  // Collect all service accounts that requires access to a secret.
+  // projectId -> secretName -> Set of service accounts
+  const toEnsure: Record<string, Record<string, Set<string>>> = {};
+  for (const e of backend.allEndpoints(b)) {
+    const sa = e.serviceAccountEmail || defaultServiceAccount(e.project);
```
inlined, line 112: Again, we're going to need to use the right service account for e.platform

src/deploy/functions/ensure.ts
```diff
@@ -0,0 +1,131 @@
...
+    );
+    logLabeledSuccess(
+      "functions",
+      `ensured ${clc.bold(serviceAccounts.join(", "))} access to ${clc.bold(secret)}.`
+    );
+  };
+
+  // Collect all service accounts that requires access to a secret.
+  // projectId -> secretName -> Set of service accounts
+  const toEnsure: Record<string, Record<string, Set<string>>> = {};
```
inlined, line 110: 🤯 This can't be any simpler? E.g. this supports cross-project secrets; do we really want this?

src/deploy/functions/ensure.ts
inlined: Also, I think we have this backwards in a dangerous way. It's very possible that I'd run a privileged function as a separate service account, but this logic grants all known service accounts to all known environment variables.

src/deploy/functions/ensure.ts
```diff
@@ -0,0 +1,131 @@
...
+
+      secrets[s.secret] = serviceAccounts;
+      toEnsure[s.projectId] = secrets;
+    }
+  }
+
+  const ensure = [];
+  for (const [projectId, secrets] of Object.entries(toEnsure)) {
+    for (const [secret, serviceAccounts] of Object.entries(secrets)) {
+      ensure.push(ensureAccess(projectId, secret, Array.from(serviceAccounts)));
```
inlined, line 127: If I could make your life even more demanding, I wonder if we should only do the ensure check for secret <-> service account pairs that are new compared to the previous deployment. Not only will this make deployments trivially faster, it'll make sure users only need access to set these secret permissions the first time. Otherwise you need to be OWNER to use CF3 with secrets ever.

src/deploy/functions/prepare.ts
```diff
@@ -1,22 +1,21 @@
...
+import { ensureCloudBuildEnabled, ensureSecretAccess, maybeEnableAR } from "./ensure";
+import { functionIdsAreValid, secretsAreValid } from "./validate";
+import { functionMatchesAnyGroup, getFilterGroups } from "./functionsDeployHelper";
+import { logBullet } from "../../utils";
+import { getFunctionsConfig, prepareFunctionsUpload } from "./prepareFunctionsUpload";
+import { promptForFailurePolicies, promptForMinInstances } from "./prompts";
 import { previews } from "../../previews";
 import { needProjectId } from "../../projectUtils";
 import { track } from "../../track";
-import * as runtimes from "./runtimes";
```
inlined, line 17: but.. but... go-style `module.function` made it easier to keep track of where all this damned code came from =/

Ah well. Guess it's time for me to be more hands off as a manager.

src/deploy/functions/prepare.ts
```diff
@@ -168,7 +151,7 @@ export async function prepare(
   );
 
   // Validate the function code that is being deployed.
-  validate.functionIdsAreValid(backend.allEndpoints(wantBackend));
+  functionIdsAreValid(backend.allEndpoints(wantBackend));
```
inlined, line 154: See... this is why I liked the package name. validate.functionIdsAreValid is a command. functionIdsArevalid sounds like it should return a bool. If we're not going to use the package name can we rename the function something like `validateFunctionIds` or `assertFunctionIdsAreValid`?

src/deploy/functions/validate.ts
```diff
@@ -49,3 +53,73 @@ export function functionIdsAreValid(functions: { id: string; platform: string }[
...
+    .allEndpoints(b)
+    .filter((e) => e.secretEnvironmentVariables && e.secretEnvironmentVariables.length > 0);
+  validatePlatformTargets(endpoints);
+  await validateSecretVersions(endpoints);
+}
+
+/**
+ * Ensures that all endpoints specifying secret environment variables target platform that supports the feature.
+ */
+function validatePlatformTargets(endpoints: backend.Endpoint[]) {
```
inlined, line 74: We shouldn't disable a feature in v2 that exists in v1. In the future, people should be able to use the v2 params syntax in v1 (even if it looks wonky) and will be able to use the old string syntax instead of a param if they prefer to hard-code things and do it TheHardWay (TM)

src/deploy/functions/validate.ts
```diff
@@ -49,3 +53,73 @@ export function functionIdsAreValid(functions: { id: string; platform: string }[
...
+/**
+ * Validate each secret version referenced in target endpoints.
+ *
+ * A secret version is valid if:
+ *   1) It exists.
+ *   2) It's in state "enabled".
+ */
+async function validateSecretVersions(endpoints: backend.Endpoint[]) {
+  const validate = async (s: backend.SecretEnvVar) => {
+    const sv = await getSecretVersion(s.projectId, s.secret, "latest");
```
inlined, line 95: What's the permission level required for this API call? Is it something we should do to assist customers when we have permissions but otherwise be silent?

src/deploy/functions/validate.ts
```diff
@@ -49,3 +53,73 @@ export function functionIdsAreValid(functions: { id: string; platform: string }[
...
+      throw new FirebaseError(
+        `Expected secret ${s.secret}@${s.version} to be in state ENABLED not ${sv.state}.`
+      );
+    }
+  };
+
+  const validations: Promise<void>[] = [];
+  for (const e of endpoints) {
+    for (const s of e.secretEnvironmentVariables! || []) {
+      validations.push(validate(s));
```
inlined, line 113: This will revalidate the same secret repeatedly if it's used in multiple endpoints.

src/deploy/functions/validate.ts
```diff
@@ -49,3 +53,73 @@ export function functionIdsAreValid(functions: { id: string; platform: string }[
...
+    }
+  }
+  const results = await utils.allSettled(validations);
+
+  const errs: { message: string }[] = results
+    .filter((r) => r.status === "rejected")
+    .map((r) => (r as utils.PromiseRejectedResult).reason as { message: string });
+  if (errs.length) {
+    const msg = errs.map((e) => e.message).join(", ");
+    throw new FirebaseError(`Failed to validate secret versions: ${msg}`);
```
inlined, line 123: FYI: FirebaseError also has a `children` option. Not sure I'm a fan of it (since it seems to only show up in the debug log) so I'm totally cool with putting the message here too.

src/gcp/secretManager.ts
```diff
@@ -1,5 +1,22 @@
...
 import { logLabeledSuccess } from "../utils";
-import * as api from "../api";
+import { FirebaseError } from "../error";
+import { Client } from "../apiv2";
+import { secretManagerOrigin } from "../api";
+
+// Matches projects/{PROJECT}/secrets/{SECRET}
+const SECRET_NAME_REGEX = new RegExp(
+  "projects\\/" +
+    "((?:[0-9]+)|(?:[A-Za-z]+[A-Za-z\\d-]*[A-Za-z\\d]?))\\/" +
```
inlined, line 11: nit: it's weird to see you combine "[0-9]" and "[\d]" in the same capture group

inlined: Also, are we sure we want to do project #3509

src/gcp/secretManager.ts
```diff
@@ -1,5 +1,22 @@
...
+    "secrets\\/" +
+    "([A-Za-z\\d\\-_]+)"
```
inlined, line 13: Can we actually end with a dash or underscore?

src/gcp/secretManager.ts
```diff
@@ -63,123 +94,131 @@ export async function secretExists(projectId: string, name: string): Promise<boo
...
 export function parseSecretResourceName(resourceName: string): Secret {
-  const nameTokens = resourceName.split("/");
+  const tokens = resourceName.match(SECRET_NAME_REGEX);
+  if (tokens == null) {
+    throw new FirebaseError(`Invalid secret resource name [${resourceName}].`);
+  }
   return {
-    projectId: nameTokens[1],
-    name: nameTokens[3],
+    projectId: tokens[1],
```
inlined, line 102: If you're going to go through the work of a bunch of non-capture groups and captures by index I wonder if it's worth having named captures instead.

src/gcp/secretManager.ts
```diff
@@ -63,123 +94,131 @@ export async function secretExists(projectId: string, name: string): Promise<boo
...
 
-export async function grantServiceAgentRole(
+export async function getIamPolicy(secret: Secret): Promise<iam.Policy> {
+  const res = await client.get<iam.Policy>(
+    `projects/${secret.projectId}/secrets/${secret.name}:getIamPolicy`
+  );
+  return res.body;
+}
+
+export async function setIamPolicyBindings(secret: Secret, bindings: iam.Binding[]): Promise<void> {
```
inlined, line 167: nit: why "Bindings"?

src/gcp/secretManager.ts
```diff
@@ -63,123 +94,131 @@ export async function secretExists(projectId: string, name: string): Promise<boo
...
-    "GET",
-    `/v1beta1/projects/${secret.projectId}/secrets/${secret.name}:getIamPolicy`,
-    {
-      auth: true,
-      origin: api.secretManagerOrigin,
+  const policy = await module.exports.getIamPolicy(secret);
+  const bindings = policy.bindings || [];
+
+  const newBindings = [];
+  for (const serviceAccountEmail of serviceAccountEmails) {
```
inlined, line 192: Rather than adding one binding per service account, find or create a binding for the required role and add all service accounts as members. You can see how Cole did this in the functions APIs for the invoker role.

src/gcp/secretManager.ts
```diff
@@ -63,123 +94,131 @@ export async function secretExists(projectId: string, name: string): Promise<boo
...
-        updateMask: {
-          paths: "bindings",
-        },
-      },
-    }
-  );
+
+  bindings.push(...newBindings);
+
+  await module.exports.setIamPolicyBindings(secret, bindings);
```
inlined, line 214: We should also probably return early if the role already exists.

inlined (changes requested):
Still need to review some of the tests; it's getting late.

src/deploy/functions/ensure.ts
taeold: This make sense, but since overall v2 support is going to be nontrivial addition to the code, I'm thinking of tackling in a separate PR (will make a separate cf3-secrets branch). WDTY?

src/deploy/functions/ensure.ts
taeold: uh oh... let me create another PR to fix this...

taeold: https://github.com/firebase/firebase-tools/pull/4034

src/deploy/functions/ensure.ts
taeold: Cut cross-project secret here and elsewhere!

src/deploy/functions/ensure.ts
taeold: very good idea. will see if i can crank this out.

src/deploy/functions/ensure.ts
taeold: hmm I don't think I understand what you are trying to say here.

This code:

1. Constructs a mapping from secret to all service account that needs access to the secret
2. Maybe SetIamPolicy to allow each SA associated w/ a secret to access the secret

IIUC, you are worried about possibly sending IAM policy update to allow all SAs identified in the backend to access all secrets referenced in the source. It doesn't do that.

src/deploy/functions/prepare.ts
taeold: Lol funny story - the reason I do this, here and onward, is because IntelliJ defaults doing per-object imports 🤷🏼 . I see your point and will revert the changes.

src/deploy/functions/validate.ts
taeold: Using another thread to discuss plans for v2 support.

src/gcp/secretManager.ts
taeold: I wasn't careful enough to check before this comment, but it turns out Secret Manager does allow secrets to end with - or _.

src/gcp/secretManager.ts
taeold: This regex is used to parse Secret Manager output, so we do need project (I'm not sure how the linked issue relates to this?)

src/gcp/secretManager.ts
taeold: I can see why we might want to do it (it's strictly more correct?), but IAM endpoint seems happily collate the bindings into one for me. Do you prefer we don't rely on this behavior?

src/gcp/secretManager.ts
taeold: I think that's what we do above. Am I mistaking something?

src/deploy/functions/ensure.ts
taeold: Cranked the code out to reduce # of iam calls!

src/deploy/functions/prepare.ts
taeold: Reverted.

src/deploy/functions/validate.ts
taeold: I don't think permission issues raised here is going to be confusing - you either have permission to list secrets in a project or you don't. 

Ensuring secret access code - that I think will benefit from some hand-holding. Will seek change there.

src/gcp/secretManager.ts
taeold: I thought there were more things than bindings in a policy (there arer, but not really mutable by the user).

src/deploy/functions/validate.ts
taeold: Good catch. Moved around the code to eliminate repeated calls.

src/deploy/functions/validate.ts
taeold: Great suggestion! Looks pretty good:

```
✔  functions: resolved secret version of MY_SECRET to 3.
Error: Failed to validate secret versions:
- FirebaseError HTTP Error: 404, Secret [projects/829294825753/secrets/MY_SECRET123] not found or has no versions.
- FirebaseError Expected secret MY_SECRET@3 to be in state ENABLED not DESTROYED.
```

colerogers (commented):
I went through it, everything makes sense and LGTM. Holding off until @inlined looks it over since he has some outstanding comments

src/deploy/functions/ensure.ts
taeold: Went ahead and implemented the fn to support both v1 and v2 platform.

src/deploy/functions/ensure.ts
taeold: this code snippet probably needs a new review - it's gone through several changes:

1) default service account logic works for both v1 and v2.
2) It looks at existing function definitions to avoid making IAM calls if secret is already bound to a SA.

src/gcp/secretManager.ts
taeold: This is now fixed.

src/gcp/secretManager.ts
taeold: This code has changed a bit, but the modified code will short circuit if no changes are necessary.

inlined (approved):

src/deploy/functions/backend.ts
```diff
@@ -470,6 +479,17 @@ export function someEndpoint(
   return false;
 }
 
+/** A helper utility for finding an endpoint that matches the predicate. */
+export function findEndpoint(
```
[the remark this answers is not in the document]
taeold: It happened few times where I thought it would be fun to have this function (though I think I use it in another PR).

src/deploy/functions/ensure.ts
```diff
@@ -0,0 +1,164 @@
...
+    );
+  };
+
+  const wantSecrets = await secretsToServiceAccounts(wantBackend);
+  const haveSecrets = await secretsToServiceAccounts(haveBackend);
+
+  // Remove secret/service account pairs that already exists to avoid unnecessary IAM calls.
+  for (const [secret, serviceAccounts] of Object.entries(haveSecrets)) {
+    for (const serviceAccount of serviceAccounts) {
+      if (wantSecrets?.[secret].has(serviceAccount)) {
```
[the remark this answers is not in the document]
taeold: Yes

src/deploy/functions/validate.ts
```diff
@@ -84,3 +87,93 @@ export function functionIdsAreValid(functions: { id: string; platform: string }[
...
+  const toResolve: Set<string> = new Set();
+  for (const e of endpoints) {
+    for (const s of e.secretEnvironmentVariables! || []) {
+      toResolve.add(s.secret);
+    }
+  }
+
+  const results = await utils.allSettled(
+    Array.from(toResolve).map(async (secret): Promise<SecretVersion> => {
+      const sv = await getSecretVersion(projectId, secret, "latest");
```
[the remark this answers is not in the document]
taeold: I think the current error message (returned by Secret Manager) is descriptive enough, e.g.

```
Error: Failed to validate secret versions:
- FirebaseError HTTP Error: 404, Secret [projects/829294825753/secrets/MY_SECRETS] not found or has no versions.
```

src/deploy/functions/validate.ts
```diff
@@ -84,3 +87,93 @@ export function functionIdsAreValid(functions: { id: string; platform: string }[
...
+      logLabeledSuccess(
```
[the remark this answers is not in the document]
taeold: Fair point. Changed.

src/deploy/functions/validate.ts
```diff
@@ -84,3 +87,93 @@ export function functionIdsAreValid(functions: { id: string; platform: string }[
...
+  if (errs.length) {
+    throw new FirebaseError("Failed to validate secret versions", { children: errs });
+  }
+
+  // Fill in versions.
+  for (const e of endpoints) {
+    for (const s of e.secretEnvironmentVariables! || []) {
+      s.version = secretVersions[s.secret].versionId;
+      if (!s.version) {
+        throw new FirebaseError("Secret version is unexpectedly undefined. Please file a ticket.");
```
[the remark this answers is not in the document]
taeold: Retracted suggestion to file a ticket and made error message more dreadful.

Issue #3926, opened 2021-12-01
`GOOGLE_APPLICATION_CREDENTIALS` doesn't accept workload identity federation


jebibot:
### [REQUIRED] Environment info

**firebase-tools: 9.23.0**

**Platform: Ubuntu**

### [REQUIRED] Test case

GitHub Actions workflow:
```yaml
jobs:
  deploy:
    runs-on: ubuntu-latest
    permissions:
      id-token: write
    steps:
      - name: Install Firebase CLI
        run: npm install -g firebase-tools

      # this automatically exports `GOOGLE_APPLICATION_CREDENTIALS`
      # https://github.com/google-github-actions/auth/blob/abba7c4d9a90b51f27710d62c17077e3e0bc9b78/src/client/workload_identity_client.ts#L196-L214
      # https://github.com/google-github-actions/auth/blob/abba7c4d9a90b51f27710d62c17077e3e0bc9b78/src/main.ts#L104-L106
      - id: auth
        name: Authenticate to GCP
        uses: google-github-actions/auth@v0.4.1
        with:
          create_credentials_file: true
          workload_identity_provider: projects/123456789/locations/global/workloadIdentityPools/my-pool/providers/my-provider
          service_account: my-service-account@my-project.iam.gserviceaccount.com

      - name: Run Firebase
        run: firebase projects:list
```

### [REQUIRED] Steps to reproduce

1. Setup Workload Identity Federation according to https://github.com/google-github-actions/auth#setting-up-workload-identity-federation
2. Setup the GitHub Actions workflow above (replace `workload_identity_provider` and `service_account`)
3. Run the GitHub Actions workflow

### [REQUIRED] Expected behavior

Successfully authenticates to Firebase projects

### [REQUIRED] Actual behavior

```
> command requires scopes: ["email","openid","https://www.googleapis.com/auth/cloudplatformprojects.readonly","https://www.googleapis.com/auth/firebase","https://www.googleapis.com/auth/cloud-platform"]
Error: The incoming JSON object does not contain a client_email field
    at JWT.fromJSON (/opt/hostedtoolcache/node/14.18.1/x64/lib/node_modules/firebase-tools/node_modules/google-auth-library/build/src/auth/jwtclient.js:195:19)
    at GoogleAuth._cacheClientFromJSON (/opt/hostedtoolcache/node/14.18.1/x64/lib/node_modules/firebase-tools/node_modules/google-auth-library/build/src/auth/googleauth.js:301:16)
[...]

PR #3972, opened 2021-12-28
Upgrade google-auth-library

rockwotj:
### Description

This should allow for authenticating via federated workload identity in Github Actions. See: https://github.com/FirebaseExtended/action-hosting-deploy/issues/174

### Scenarios Tested

```
firebase logout
firebase login
firebase projects:list
firebase login:ci
firebase --token="$TOKEN_FROM_PREV_CMD" projects:list
```

### Sample Commands

N/A


bkendall (commented):
Yay! I like this.

Could you please test the login/logout flows and make note of that in the PR description? It makes me nervous when I don't see that things have been tested.

For what it's worth: the fancier tests do work with a service account, so I think that path is relatively safe, but it'd be nice to know the normal login/out flows work correctly. Thanks!

(Also, hi! Been a while! 😄)

rockwotj:
Hi @bkendall! I didn't expect anyone to be working this week :) But I'm working so I suppose that's hypocritical. Thanks for a speedy response - I've done a smattering of basic tests of auth flows and the `projects:list` command. Those all work fine. If there are commands with fancier auth usage I'm happy to test those too. The PR description has been updated.

FWIW I've looked through the [changelog for google-auth-client](https://github.com/googleapis/google-auth-library-nodejs/blob/main/CHANGELOG.md) and it seems like the only "breaking" changes are these new features to support workload-identity.

bkendall:
Fixes #3926

bkendall (approved):

Issue #3793, opened 2021-09-30
Calling functions.config() from JS broken on 9.19.0 when running functions emulator


ljle:
### [REQUIRED] Environment info

**firebase-tools:** 9.19.0

**Platform:** Windows with WSL2 on Arch Linux

### [REQUIRED] Test case

- No repo, see comment below.

### [REQUIRED] Steps to reproduce
- Call `functions.config()` from any JS code by importing `firebase-functions` while running the functions emulator returns an empty object.
![image](https://user-images.githubusercontent.com/696474/135492620-13bea166-6968-4fed-95e4-63304a2dbb62.png)

### [REQUIRED] Expected behavior

- `functions.config()` returns the values inside `functions/.runtimeconfig.json`.

### [REQUIRED] Actual behavior

- `functions.config()` returns an empty object. However, this does not happen after downgrading to `9.18.0`.

ggirotto:
Same here

FelDev:
I have the same issue on a mac.
Reverting to 9.18.0 fixed it.

taeold:
Can folks share the version of `firebase-functions` being used? 

I tried reproducing with the latest firebase-functions@3.15.7 and firebase-tools@9.19.0 on MacOS Big Sur without an issue.

xwz:
Seems to be removed without replacement? https://github.com/firebase/firebase-tools/compare/v9.18.0...master#diff-0366293848c53849685ead9c0871e6ab4c0b2232ebdb379ba79a6def21f99fc3L678

ljle:
> Can folks share the version of `firebase-functions` being used?
> 
> I tried reproducing with the latest firebase-functions@3.15.7 and firebase-tools@9.19.0 on MacOS Big Sur but things worked as expected.

Can confirm this fixes the issue.

I just tried it with `firebase-functions@3.15.7` and `firebase-tools@9.20.0` and the emulator loaded successfully.

It was previously failing while using `firebase-functions@3.14.1`.

vlombardifiters:
I confirm that the issue exists in v9.20.0 too.

roblonj:
@taeold i tried using firebase-functions@3.15.7 and firebase-tools@9.20.0 and it didn't work. Reverting firebase-tools@9.18.0 worked for me.

ketansontakke:
[...]

PR #3974, opened 2021-12-29
Revive the runtime config loader in the emulator runtime.

taeold:
Most recent version of Firebase Functions SDK automatically picks up locally stored `.runtimeconfig.json` to populate the config entries.

However, due to a bug in some older version of the Function SDK, this process may fail. See the following issues for more detail:

* https://github.com/firebase/firebase-tools/issues/3793
* https://github.com/firebase/firebase-functions/issues/877

As a workaround, the emulator runtime will load the contents of the `.runtimeconfig.json` to the `CLOUD_RUNTIME_CONFIG` environment variable. In the future, we will bump up the minimum version of the Firebase Functions SDK required to run the functions emulator to v3.15.1 and get rid of this workaround.

Also - noticed a dead code that can be removed, so removed.


src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -521,6 +523,34 @@ function getDefaultConfig(): any {
...
+  // In the future, we will bump up the minimum version of the Firebase Functions SDK
+  // required to run the functions emulator to v3.15.1 and get rid of this workaround.
+  const configPath = `${frb.cwd}/.runtimeconfig.json`;
+  try {
+    const configContent = fs.readFileSync(configPath, "utf8");
+    if (configContent) {
+      try {
+        JSON.parse(configContent.toString());
+        logDebug(`Found local functions config: ${configPath}`);
+        process.env.CLOUD_RUNTIME_CONFIG = configContent.toString();
```
joehan, line 544: I'm a bit unfamiliar with how this all works, so this might be off base - however, should we check if `process.env.CLOUD_RUNTIME_CONFIG` is already defined before we try to load it?

joehan (approved):
Code looks good, just wondering if we should check wether this env variable is already defined before trying to read the file

bkendall:
Does this fix #3983?

src/emulator/functionsEmulatorRuntime.ts
taeold: Good catch. Updated.

taeold:
@bkendall Yep. Responded to the issue.

PR #3986, opened 2022-01-05
increase auth emulator timeout

bkendall:
### Description

10s seems to be too tight of a deadline. Increasing to 20s (and adding a tiny bit of logging so we can track for a bit)


bkendall:
the three I checked were 8.5, 10.5, and 11.2 seconds. Seems like this is a good change.

yuchenshi (approved):

PR #3988, opened 2022-01-05
update some dev dependencies

bkendall:
### Description

Updating a few dev dependencies:

- eslint and friends for parsing typescript, etc.
- prettier for formatting (and support for newer typescript features)
- ts-node for the latest runtime from them
- moved @types/archiver into dev dependencies

I don't love some of the formatting changes, but I've also seen worse from prettier. All well.


src/utils.ts
```diff
@@ -578,6 +583,9 @@ export function assertIsString(val: any, message?: string): asserts val is strin
   }
 }
 
+/**
+ *
```
yuchenshi, line 587:
```suggestion
 * Asserts val is a number, or throw an error with message otherwise.
```

src/utils.ts
```diff
@@ -586,6 +594,9 @@ export function assertIsNumber(val: any, message?: string): asserts val is numbe
...
```
yuchenshi, line 598:
```suggestion
 * Asserts val is a string or undefined, or throw an error with message otherwise.
```

yuchenshi (approved):

PR #3989, opened 2022-01-05
Hosting deploy migration to apiv2, typescript, project-scoped

### LINKED ISSUE #2558: .firebaserc and firebase cli completely ignored
If you have a .firebaserc file setup
```
{
  "projects": {
    "dev": "dev-land",
    "stage": "stage-land",
    "prod": "prod-land",
    "default": "dev-land"
  }
}
```

And have a firebase.json like this

```
{
  "hosting": {
    "public": "dist",
    "rewrites": [{
      "source": "**",
      "destination": "/index.html"
    }],
    "site": "dev-land-9008c",
    "ignore": ["firebase.json", "**/.*", "**/node_modules/**"]
  }
}
```
Some firebase command or process puts the "site": "dev-land-9008c", line in there, not sure what does it.  Doesn't really matter.  What matters is the behaviors observed when this line exists. This line overrides .firebaserc, and also overrides ALL cli commands.

if you run: ```firebase use dev``` the CLI reports back that it successfully switched to 'dev-land'

If you then run ```firebase deploy``` it will say '=== Deploying to dev-land...' but then in the status updates it will say 'Deployed successfully to 'dev-land-9008c'.

If you run ```firebase deploy --project dev-land``` same thing, it reports its deploying to 'dev-land' but deploys to 'dev-land-9008c' instead.

If you delete .firebaserc, and redo all of it using ```firebase use --add```, same behavior.
If you run ```firebase use --clear```, same behavior.
If you re-run ```firebase init```, same behavior.

Needless to say this took me over 3 hours to find, EXTREMELY frustrating.  It started in my CI/CD pipeline, I spent 2 hours there, then noticed it was happening locally.  Like I said I have no idea what adds the site line to firebase.json but that line shouldn't exist.  Especially with the aliasing, and .firebaserc.  That should be removed entirely, beyond confusing.

bkendall:
### Description

I've updated the hosting deploy path to be:

- in typescript
- using apiv2
- project scoped

This helpfully fixes the issue where a deploy can go to a site _not_ on the same project, which is definitely not intended.

Fixes #2558

### Scenarios Tested

- `firebase deploy --only hosting`
- `firebase hosting:channel:deploy staging`


src/deploy/hosting/prepare.ts
```diff
@@ -0,0 +1,66 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
```
mbleigh, line 6: I thought we had typically been fixing these, not just eslint ignoring them, when transitioning to TS?

mbleigh (commented):
LGTM but question about ES Lint exceptions before approval.

src/deploy/hosting/prepare.ts
bkendall: Removed these lines.

src/deploy/hosting/prepare.ts
```diff
@@ -0,0 +1,60 @@
...
+import { needProjectNumber } from "../../projectUtils";
+import { normalizedHostingConfigs } from "../../hosting/normalizedHostingConfigs";
+import { validateDeploy } from "./validate";
+import { convertConfig } from "./convertConfig";
+import * as deploymentTool from "../../deploymentTool";
+
+/**
+ *  Prepare creates versions for each Hosting site to be deployed.
+ */
+export async function prepare(context: any, options: any): Promise<void> {
```
joehan, line 12: Can we use the options type from options.ts so these warnings are silenced?

joehan (approved):
LGTM after typing options

PR #3994, opened 2022-01-07
Refactor Functions runtime delegation process to no longer depend on cli options and context

taeold:
Code for parsing CF3 triggers from function source relies on CLI options/context object to retrieve active project id, functions source directory, project root, etc.

The refactoring here pulls out all references to CLI options and context object from the runtime delegate caller. This is done in preparation for having another caller for the runtime delegate (the Functions Emulator) in the near future.


colerogers (approved):
Overall this looks like a clean refactor change, lgtm

src/deploy/functions/prepare.ts
```diff
@@ -53,18 +54,28 @@ export async function prepare(
   options: Options,
   payload: args.Payload
 ): Promise<void> {
-  if (!options.config.src.functions) {
-    return;
+  const projectId = needProjectId(options);
+
+  const sourceDirName = options.config.get("functions.source") as string;
+  if (!sourceDirName) {
+    throw new FirebaseError(
```
inlined, line 61: Be careful here. I think I tried this and we started throwing exceptions for projects that didn't include functions.

src/deploy/functions/runtimes/index.ts
```diff
@@ -100,36 +98,34 @@ export interface RuntimeDelegate {
...
 }
 
-type Factory = (context: args.Context, options: Options) => Promise<RuntimeDelegate | undefined>;
+export interface DelegateContext {
+  projectId: string;
+  // Absolute path of the Firebase project directory.
+  projectDir: string;
+  // Absolute path of the source directory.
+  sourceDir: string;
+  runtime: string;
```
inlined, line 107: nit: I might call this a runtime override or something. This overrules the normal sniffing we do.

inlined: Also, shouldn't it be optional since the firebase.json value is optional?

inlined (commented):
Looking good. Sorry I've been AWOL on reviews.

PR #3996, opened 2022-01-07
apiv2: add support for retry codes

bkendall:
### Description

Adds support for retry codes in apiv2. If specified, the api client will retry indefinitely on specific error codes.


src/apiv2.ts
```diff
@@ -383,6 +384,9 @@ export class Client {
     this.logResponse(res, body, options);
 
     if (res.status >= 400) {
+      if (options.retryCodes?.includes(res.status)) {
+        return this.doRequest(options);
+      }
```
mbleigh, line 389: Isn't this going to potentially cause a flood of requests? I'm wary of adding retry logic without any kind of exponential backoff controls.

bkendall: Alright, there's retry logic now. I'm a _little_ meh about it, but it does work (and I don't have to think about the delay calculations or anything).

mbleigh (approved):

PR #3997, opened 2022-01-07
migrate cloudbilling and functions to apiv2

bkendall:
### Description

Migrating some GCP libraries over to apiv2. It does require retryCodes support, which is why this is based on a different branch, but it does seem an easy enough migration.


src/gcp/cloudfunctions.ts
```diff
@@ -3,14 +3,16 @@ import * as clc from "cli-color";
...
 import { logger } from "../logger";
 import { previews } from "../previews";
-import * as api from "../api";
 import * as backend from "../deploy/functions/backend";
 import * as utils from "../utils";
 import * as proto from "./proto";
 import * as runtimes from "../deploy/functions/runtimes";
 import * as iam from "./iam";
+import { Client } from "../apiv2";
+import api, { functionsOrigin } from "../api";
```
taeold, line 12: If I'm reading this correctly, I don't think we need the `api` bit here:

```js
import { functionsOrigin } from "../api";
```

src/test/gcp/cloudfunctions.spec.ts
```diff
@@ -1,6 +1,7 @@
 import { expect } from "chai";
-import * as sinon from "sinon";
-import * as api from "../../api";
+import * as nock from "nock";
+
+import { functionsOrigin } from "../../api";
```
taeold, line 4: nit* Either we don't need this import or all occurrence of `https://cloudfunctions.googleapis.com` should be replaced with it.

taeold (approved):
Awesome stuff.

src/gcp/cloudfunctions.ts
bkendall: Thanks

src/test/gcp/cloudfunctions.spec.ts
bkendall: Thanks.

PR #4004, opened 2022-01-10
shrinkwrap CLI dependencies

bkendall:
### Description

Now that we've been bitten by npm dependencies changing under us and breaking the CLI experience, we're gonna bite the bullet and shrinkwrap our dependencies. It means more work to keep things up to date for us, but it should give our wonderful, amazing, best of the best developers a more consistent experience, less prone to random breakages.


anthonyjdella:
Sorry if this is a dumb question, but should both the `npm-shrinkwrap.json` & `package-lock.json` be present? If they are, `package-lock.json` will be ignored.

[NPM docs state](https://docs.npmjs.com/cli/v6/configuring-npm/package-lock-json): 
"If both package-lock.json and npm-shrinkwrap.json are present in the root of a package, package-lock.json will be completely ignored."

bkendall:
@anthonyjdella I'm unsure of your concern. `npm-shrinkwrap` is the only one that is present in this branch. Since it does the same function as package-lock, both don't need to be present.

anthonyjdella:
@bkendall My mistake, I thought both were present 😅

mbleigh (approved):

PR #4008, opened 2022-01-12
Align storage emulator with GCS

### LINKED ISSUE #3953: Storage Emulator does not match Google Cloud Storage library API

**firebase-tools:** 
9.23.0

**Platform:**
Windows 10

### [REQUIRED] Test case
Using the Google Cloud Storage Python library which follows the [Google Cloud Storage JSON API spec](https://cloud.google.com/storage/docs/json_api/v1/objects/insert) upload a file
```
blob = blob_from_path(path)
blob.metadata = metadata
blob.upload_from_string(content, content_type, num_retries=3)
```

Any GCS Client library which puts name inside the body of the "file" metadata of the upload and doesn't put name as a query parameter. (Again as per [Google Cloud Storage JSON API spec](https://cloud.google.com/storage/docs/json_api/v1/objects/insert) upload a file)

The emulator should get the name of the file from the metadata passed by the client library. Then the file should be uploaded and file metadata returned in a 200 response.

Returns 400 because name is not found in the query (even though name is in the metadata)

**The Problem**
The problems stems from the emulator/storage/apis/gcloud.js implementation
```
gcloudStorageAPI.post("/upload/storage/v1/b/:bucketId/o", (req, res) => {
        if (!req.query.name) {
            res.sendStatus(400);
            return;
        }
        let name = req.query.name.toString();
        if (name.startsWith("/")) {
            name = name.slice(1);
        }
```
The POST handler is checking for a name parameter but the spec says it is optional and may be instead passed via metadata which is in the metadata section of a multipart upload or in the body for a resumable upload.

I just thought it might be worthy of fixing.

kenttregenza:
### Description

Align storage emulator with GCS way of handling name. Simply this gets the name from either the query parameter or the metadata. In the process this patch tries to improve the multipart handling as the Python client puts quotes around the boundary value in the content type header. Fixes  #3953 

### Scenarios Tested

Using Google Cloud Storage client for Python. Uploading a small file (that uses multipart) and a large file that simulates a resume able upload

### Sample Commands

NA

### DIFF
--- a/src/emulator/storage/apis/gcloud.ts
+++ b/src/emulator/storage/apis/gcloud.ts
@@ -4,6 +4,7 @@ import { Emulators } from "../../types";
 import {
   CloudStorageObjectAccessControlMetadata,
   CloudStorageObjectMetadata,
+  IncomingMetadata,
   StoredFileMetadata,
 } from "../metadata";
 import { EmulatorRegistry } from "../../registry";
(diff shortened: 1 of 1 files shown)


tonyjhuang:
Hi @kenttregenza, as mentioned on https://github.com/firebase/firebase-tools/issues/3953, could you please rebase off master and split this into two separate PRs, one of the name change and one for the multipart-related changes. Thanks!

PR #4012, opened 2022-01-12
Use runtime delegate to parse function triggers in the Functions Emulator

taeold:
Today, Functions Emulator parses function trigger from source as follows:

1) Spin up an instance of Functions runtime process
2) Invoke a "function" which triggers a path that parses the trigger by calling out to `extractTriggers.js`
3) Send parsed triggers via IPC from runtime to emulator. Emulator now knows about the triggers.

This has the advantage of running the trigger parsing in the emulated runtime (which properly mocks out calls to the DB, applies network filtering, uses the same node version when possible, etc.) but has the disadvantage of complicating the runtime implementation as well as diverging from how the triggers are parsed in `firebase deploy`.

Using runtime delegate, we have:

1) Use runtime delegate to discover the delegate appropriate for function source (i.e. Node, but in the future can be some other runtime)
2) Spin up a node subprocess to parse trigger. Emulator now knows about the triggers.

i.e. the same procedure used during `firebase deploy`

By using runtime delegate, we align the function deploy to production and to the emulated environment and simplify the runtime code a bit. This also puts us into a good position in the future when we make the function deploy process a little more complex, e.g. params and secrets support.

One missing feature that could not be ported nicely here is soft-enforcement of function timeouts. I wanted to leave it out until I make further refinement in the Functions Emulator to make function invocation simpler. It will be made up in a subsequent PR.


src/emulator/functionsEmulator.ts
```diff
@@ -440,14 +444,8 @@ export class FunctionsEmulator implements EmulatorInstance {
...
    * When a user changes their code, we need to look for triggers defined in their updates sources.
-   * To do this, we spin up a "diagnostic" runtime invocation. In other words, we pretend we're
-   * going to invoke a cloud function in the emulator, but stop short of actually running a function.
-   * Instead, we set up the environment and catch a special "triggers-parsed" log from the runtime
-   * then exit out.
-   *
-   * A "diagnostic" FunctionsRuntimeBundle looks just like a normal bundle except triggerId == "".
    *
-   * TODO(abehaskins): Gracefully handle removal of deleted function definitions
+   * TODO(danielylee): Gracefully handle removal of deleted function definitions
```
inlined, line 448: small brain: leaving TODO to ex employee
big brain: changing TODO to self
galaxy brain: filing bug, putting it in our backlog, and leaving it in the TODO assignment

src/emulator/functionsEmulator.ts
```diff
@@ -459,33 +457,35 @@ export class FunctionsEmulator implements EmulatorInstance {
...
-    );
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-    const parsedDefinitions = triggerParseEvent.data
-      .triggerDefinitions as ParsedTriggerDefinition[];
-
-    const triggerDefinitions: EmulatedTriggerDefinition[] =
-      emulatedFunctionsByRegion(parsedDefinitions);
 
+    let triggerDefinitions: EmulatedTriggerDefinition[];
+    if (emulatableBackend.predefinedTriggers) {
```
inlined, line 462: Probably another TODO that this should be unified after extensions no longer use extensions.yaml for their resource definitions (e.g. we'll start using the normal path for newly written extensions and then in CLIv11 we'll break the old extension definitions)

src/emulator/functionsEmulator.ts
```diff
@@ -459,33 +457,35 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+      triggerDefinitions = emulatedFunctionsByRegion(emulatableBackend.predefinedTriggers);
+    } else {
+      const runtimeDelegate = await getRuntimeDelegate({
+        projectId: this.args.projectId,
+        projectDir: this.args.projectDir,
+        sourceDir: emulatableBackend.functionsDir,
+        runtime: "",
```
inlined, line 469: let's just make this omittable.

src/emulator/functionsEmulator.ts
```diff
@@ -459,33 +457,35 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+        // Don't include user envs when parsing triggers.
+        {
+          ...this.getSystemEnvs(),
+          ...this.getEmulatorEnvs(),
+          FIREBASE_CONFIG: this.getFirebaseConfig(),
+          ...emulatableBackend.env,
+        }
+      );
+      const endpoints = backend.allEndpoints(discoveredBackend);
+      triggerDefinitions = emulatedFunctionsFromEndpoints(endpoints);
```
inlined, line 487: is "emulated function" going to barf on scheduled, task queue, etc. functions?

src/emulator/functionsEmulator.ts
```diff
@@ -659,7 +659,7 @@ export class FunctionsEmulator implements EmulatorInstance {
         return true;
       })
       .catch((err) => {
-        this.logger.log("WARN", "Error adding trigger: " + err);
+        this.logger.log("WARN", "Error adding rtdb trigger: " + err);
```
inlined, line 662: s/rtdb/realtime database

These aren't employees and they don't know the slang.

src/emulator/functionsEmulator.ts
```diff
@@ -954,6 +960,7 @@ export class FunctionsEmulator implements EmulatorInstance {
     envs.TZ = "UTC"; // Fixes https://github.com/firebase/firebase-tools/issues/2253
     envs.FIREBASE_DEBUG_MODE = "true";
     envs.FIREBASE_DEBUG_FEATURES = JSON.stringify({ skipTokenVerification: true });
+    // TODO(danielylee): Support timeouts. Temporarily dropping the feature until we finish refactoring.
```
inlined, line 963: Does this put a freeze on CLI releases? Should this be checked into a feature branch?

src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -960,26 +941,29 @@ async function invokeTrigger(
...
+    }
+    const timeoutMs = parseInt(timeout, 10) * 1000;
     timeoutId = setTimeout(() => {
       new EmulatorLog(
         "WARN",
         "runtime-status",
-        `Your function timed out after ~${
-          trigger.definition.timeout || "60s"
-        }. To configure this timeout, see
+        `Your function timed out after ~${timeout}. To configure this timeout, see
```
inlined, line 953: nit: add back the "s" or "seconds" after timeout.

src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -935,23 +927,12 @@ function logDebug(msg: string, data?: any): void {
...
   new EmulatorLog("INFO", "runtime-status", `Beginning execution of "${frb.triggerId}"`, {
     frb,
   }).log();
 
-  const trigger = triggers[frb.triggerId];
-  logDebug("triggerDefinition", trigger.definition);
-  const signature = getSignatureType(trigger.definition);
-
-  logDebug(`Running ${frb.triggerId} in signature ${signature}`);
+  logDebug(`Running ${frb.triggerId} in signature ${FUNCTION_TARGET_NAME}`);
```
inlined, line 935: is FUNCTION_TARGET_NAME really the thing you want to print here as a "signature"?

src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -1031,27 +1037,25 @@ async function initializeRuntime(
     } catch (err: any) {
       if (err.code !== "ERR_REQUIRE_ESM") {
         await moduleResolutionDetective(frb, err);
-        return;
+        throw err;
```
inlined, line 1040: Why are we awaiting moduleResolutionDetective and then throwing still? Can we at least get a comment?

src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -1031,27 +1037,25 @@ async function initializeRuntime(
...
   }
-  if (extensionTriggers) {
-    parsedDefinitions = extensionTriggers;
-  } else {
-    require("../deploy/functions/runtimes/node/extractTriggers")(triggerModule, parsedDefinitions);
+  const maybeTrigger = functionTarget.split(".").reduce((mod, functionTargetPart) => {
+    if (typeof mod === "undefined") {
+      return undefined;
+    } else {
+      return mod[functionTargetPart];
```
inlined, line 1052: Isn't this just `mod?.[functionTargetPart]`?

src/emulator/functionsEmulatorShared.ts
```diff
@@ -135,6 +142,65 @@ export class EmulatedTrigger {
...
+    } else if (isEventTriggered(endpoint)) {
+      const eventTrigger = endpoint.eventTrigger;
+      if (endpoint.platform === "gcfv1") {
+        def.eventTrigger = {
+          eventType: eventTrigger.eventType,
+          resource: eventTrigger.eventFilters.resource,
+        };
+      } else {
+        // Only pubsub and storage events are supported for gcfv2.
+        const { resource, topic, bucket } = endpoint.eventTrigger.eventFilters;
```
inlined, line 180: Clever, but be careful. For example, topic is an id not a name now. Will that get us in trouble? Also, contrary to my original hopes, it seems like "resource" will not be used for v2 stuff.

src/emulator/functionsEmulatorShared.ts
```diff
@@ -135,6 +142,65 @@ export class EmulatedTrigger {
...
+        const eventResource = resource || topic || bucket;
+        if (!eventResource) {
+          // Unsupported event type for GCFv2
+          continue;
```
inlined, line 184: Silently continuing seems like a recipe for customer confusion. I wonder if there's a better way to make sure this transformation is exhaustive, including reporting which triggers are unsupported (e.g. task queue triggers for now)

src/emulator/functionsEmulatorShared.ts
```diff
@@ -135,6 +142,65 @@ export class EmulatedTrigger {
...
+        }
+        def.eventTrigger = {
+          eventType: eventTrigger.eventType,
+          resource: eventResource,
+        };
+      }
+    } else if (isScheduleTriggered(endpoint)) {
+      // TODO: This is an awkward transformation. Emulator does not understand scheduled triggers - maybe it should?
+      def.eventTrigger = { eventType: "pubsub", resource: "" };
```
inlined, line 193: This is going to break in v2 where scheduled triggers are over HTTPS...

inlined (approved):
I'm not sure we should be committing to master while timeouts are disabled, but everything seems sane to me. I'm worried we're using a chisel where a jackhammer is required. We want to end the quarter with a forked version of functionsEmulator.ts that execs into a binary that replaces functionsEmulatorRuntime.

I'm vaguely of the opinion that we should:
1. Create a flag "emulatorv2"
2. Fork the implementation so that "emulatorv2" goes to a new copy of the codebase
3. Hack to our heart's content
4. Replace #1 to "customer has enabled functionsv2 and has a min SDK version"
5. Bake with all our trusted testers for 2+ months ideally
6. At I/O make a major version bump where we drop the old codebase and set a min SDK version.

There's a major danger that we'll lose all the quirks that the emulator has baked in over the years, but I don't see any other way of evolving the architecture.

src/emulator/functionsEmulator.ts
taeold: Yep. I think that' something I'll leave to Extensions team to think through 🤔

src/emulator/functionsEmulator.ts
taeold: It's never barfed on scheduled function, though it will silently ignore task queue and other unsupported functions. The user will warning log like this one:

```
          `Trigger "${definition.name}" has neither "httpsTrigger" or "eventTrigger" member`
```

to let them know about these unknown triggers.

src/emulator/functionsEmulator.ts
taeold: Technically yes, though timeouts are never really enforced.

What we lose is a warning message that would've let users know that the timeout has been reached and in prod would've killed the fn.

I'm wasn't too worried about loosing this capability, but that's kinda slacking. I'll bring this up for discussion offline.

src/emulator/functionsEmulatorRuntime.ts
taeold: great catch!

src/emulator/functionsEmulatorRuntime.ts
taeold: Isn't it actually confusing that the function isn't throwing error if the function module loading fails?

`moduleResolutionDetective` just logs some useful diagnostic messages.

src/emulator/functionsEmulatorRuntime.ts
taeold: ah very nice.

src/emulator/functionsEmulatorShared.ts
taeold: I improved the code to make sure we log a warning for every function trigger that isn't converted into a endpoint.

src/emulator/functionsEmulatorShared.ts
taeold: v1 schedule function barely works in the emulator 🤷🏼‍♂️

src/emulator/functionsEmulatorRuntime.ts
taeold: moduleResolutionDetective just prints some useful err message before terminating the instance - I think throwing makes this clear. Added comments to make it a bit more obvious what's going on here.

taeold: (In theory, this would never happen anyway - if module fails to load... we won't even get to this stage!)

src/emulator/functionsEmulatorShared.ts
taeold: Fortunately, no. 

This code path will need revision as we revise the Endpoint definition in the container contract.

src/emulator/functionsEmulatorShared.ts
joehan: +1 - Extensions would love if scheduled functions had better support

joehan (approved):
I think this LGTM, and does a good job of separating the predefinedTriggers code

inlined (approved):

src/emulator/controller.ts
```diff
@@ -414,10 +414,8 @@ export async function startAll(options: EmulatorOptions, showUI: boolean = true)
     );
 
     utils.assertIsStringOrUndefined(options.extensionDir);
-    const functionsDir = path.join(
-      options.extensionDir || options.config.projectDir,
-      options.config.src.functions.source
-    );
+    const projectDir = options.extensionDir || options.config.projectDir;
+    const functionsDir = path.join(projectDir, options.config.src.functions.source);
```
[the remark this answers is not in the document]
taeold: `options.config` object has mechanism in its constructor to fallback to "functions". So yes `config.src.functions.source` is always defined.

src/emulator/functionsEmulator.ts
```diff
@@ -584,7 +583,7 @@ export class FunctionsEmulator implements EmulatorInstance {
       } else {
         this.logger.log(
           "WARN",
-          `Trigger trigger "${definition.name}" has has neither "httpsTrigger" or "eventTrigger" member`
+          `Trigger "${definition.name}" has neither "httpsTrigger" or "eventTrigger" member`
```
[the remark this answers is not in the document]
taeold: Good point - updated.

Note that we have some weird dance in parsing the trigger in the emluator that converts scheduled triggers into event triggers. This isn't great, and I want us to eventually move to a point where Functions Emulator speaks Endpoint instead of this "EmulatedTrigger" variant, but I'm pushing that out till another time.

src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -1074,38 +1075,32 @@ async function handleMessage(message: string) {
...
+        runtimeArgs.frb,
+        FUNCTION_TARGET_NAME,
+        serializedTriggers
+      );
+    } catch (e: any) {
+      logDebug(e);
+      new EmulatorLog(
+        "FATAL",
+        "runtime-status",
+        `Failed to initialize and load trigger. This shouldn't happen.`
```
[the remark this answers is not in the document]
taeold: I did a quick scan in the emulator logs to make changes when appropriate.

src/emulator/functionsEmulator.ts
taeold: TQ is such a nice project. Let's talk about it more this Friday!

PR #4016, opened 2022-01-13
remove typing from javascript... string

bkendall:
### Description

Evidently this breaks the UI.


yuchenshi (approved):

yuchenshi:
Follow-up to clean the JS in string: b/214443916

PR #4022, opened 2022-01-18
Remove preview flag, add option to disable dotenv support.

taeold:
Preparing for launching dotenv support for CF3.

At launch, CF3 environment variables support will default to picking up dotenv file if any, without need for preview flag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Add support for specifying environment variable of CF3 function using dotenv.
--- a/src/commands/index.js
+++ b/src/commands/index.js
@@ -95,9 +95,7 @@ module.exports = function (client) {
   client.functions = {};
   client.functions.config = {};
   client.functions.config.clone = loadCommand("functions-config-clone");
-  if (previews.dotenv) {
-    client.functions.config.export = loadCommand("functions-config-export");
-  }
+  client.functions.config.export = loadCommand("functions-config-export");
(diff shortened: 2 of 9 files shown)


colerogers (approved):

inlined (commented):
I'm a bit confused. When/why would people disable dotenv? We don't blitz away env vars unless we see a dotenv, so why would they need to disable it?

I want to make sure that this isn't a breaking change since people would have to opt into disabling.

taeold:
@inlined I'm thinking of cases where users already make use of `.env` file for whatever purpose they need to fulfill. Having a switch they can press to continue on seems useful.

Are you thinking that this use case is unlikely? I don't see why having this switch is harmful, but maybe I'm missing something.

inlined:
It doesn't harm anything other than adding trivial code complexity. I just hate to see us get to a codebase like GAE where there's zillions of flags that nobody knows what they do. Maybe add an analytics line to track when it's set? Then we can delete it if we don't have a single case in 3mo

taeold:
> It doesn't harm anything other than adding trivial code complexity. I just hate to see us get to a codebase like GAE where there's zillions of flags that nobody knows what they do. Maybe add an analytics line to track when it's set? Then we can delete it if we don't have a single case in 3mo

Understood. I'll actually put the new flag in my back pocket and will make it available once there is a user complaint (YAGNI!)

PR #4026, opened 2022-01-19
Add functions:secrets:{access, destroy, get} commands.

taeold:
Follow up https://github.com/firebase/firebase-tools/pull/4021 to add other management commands for CF3 secrets.

Note that `destroy` commands can be improved by making sure we don't accidentally delete secrets versions currently in use (which would immediately break the function!). I'll add these feature in a follow up PR when we finish reviewing the PR w/ `prune` command.


src/commands/functions-secrets-destroy.ts
```diff
@@ -0,0 +1,35 @@
...
+  .description("Destroy a secret. Defaults to destroying the latest version.")
+  .withForce("Destroys a secret without confirmation.")
+  .action(async (key: string, options: Options) => {
+    const projectId = needProjectId(options);
+    let [name, version] = key.split("@");
+    if (!version) {
+      version = "latest";
+    }
+
+    if (!options.force) {
```
inlined, line 18: nit: A command that always requires you to use "force" without even doing a dry run to see what the affects are seem off. Maybe something like requiring --force if the secret is currently in use? Or at least resolving what "version" is?

src/commands/functions-secrets-destroy.ts
```diff
@@ -0,0 +1,35 @@
...
+          message: `Are you sure you want to destroy ${name}@${version}`,
+        },
+        options
+      );
+      if (!confirm) {
+        return;
+      }
+    }
+
+    await destroySecretVersion(projectId, name, version);
```
inlined, line 33: Should we destroy the actual secret if there are no versions left?

src/commands/functions-secrets-destroy.ts
taeold: If you don't specify force, users would get a prompt asking if they actually want to delete secret at the resolved version. If you do have --force, you don't get the prompt.

Not sure why you think this command always requires a force flag?

taeold: I see what you mean by resolving the version though - that's a great idea.

inlined (approved):

PR #4027, opened 2022-01-20
account import update

bkendall:
### Description

- Removing the dependency on JSONStream. using `stream-json` now, along with `stream-chain`. It's not exactly intuitive, but it works.
- Updates the csv streaming dependency to be `csv-parse`, which has been updated in the past 4 years and provides types for typescript.
- Rewrites import into typescript.

### Scenarios Tested

import and export of csv and json files.


src/commands/auth-import.ts
```diff
@@ -0,0 +1,138 @@
...
+  .before(requirePermissions, ["firebaseauth.users.create", "firebaseauth.users.update"])
+  .action(async (dataFile, options) => {
+    const projectId = needProjectId(options);
+    const checkRes = validateOptions(options);
+    if (!checkRes.valid) {
+      return checkRes;
+    }
+    const hashOptions = checkRes;
+
+    if (!_.endsWith(dataFile, ".csv") && !_.endsWith(dataFile, ".json")) {
```
joehan, line 56: Prefer native `dataFile.endsWith()` + getting rid of lodash from this file

src/commands/auth-import.ts
```diff
@@ -0,0 +1,138 @@
...
+  .option("--parallelization <parallelization>", "specify the parallelization for standard scrypt.")
+  .option("--block-size <blockSize>", "specify the block size (normally is 8) for standard scrypt.")
+  .option("--dk-len <dkLen>", "specify derived key length for standard scrypt.")
+  .option(
+    "--hash-input-order <hashInputOrder>",
+    "specify the order of password and salt. Possible values are SALT_FIRST and PASSWORD_FIRST. " +
+      "MD5, SHA1, SHA256, SHA512, HMAC_MD5, HMAC_SHA1, HMAC_SHA256, HMAC_SHA512 support this flag."
+  )
+  .before(requirePermissions, ["firebaseauth.users.create", "firebaseauth.users.update"])
+  .action(async (dataFile, options) => {
```
joehan, line 48: Can we mkae this `dataFile: string, options: Options` so some of these linter warnings go away?

src/commands/auth-import.ts
```diff
@@ -0,0 +1,138 @@
...
+            let record: string[] = [];
+            while ((record = parser.read()) !== null) {
+              counter++;
+              const trimmed = record.map((s) => {
+                const str = s.trim().replace(/^["|'](.*)["|']$/, "$1");
+                return str === "" ? undefined : str;
+              });
+              const user = transArrayToUser(trimmed);
+              const err = _.get(user, "error");
+              if (err) {
```
joehan, line 82: We can get rid of the previous line and just make this `if (user.error)`

src/commands/auth-import.ts
```diff
@@ -0,0 +1,138 @@
...
+    } else {
+      userListArr = await new Promise<any[]>((resolve, reject) => {
+        const pipeline = new Chain([
+          Pick.withParser({ filter: /^users$/ }),
+          StreamArray.streamArray(),
+          ({ value }) => {
+            counter++;
+            const user = validateUserJson(value);
+            const err = _.get(user, "error");
+            if (err) {
```
joehan, line 113: Same as above - prefer getting rid of previous line and making this `if (user.error)`

joehan (approved):
Some nits, Chain is a bit hariy, and I think we can get rid of lodash while you're here, but this is wayyyy better than what we had before already

src/commands/auth-import.ts
bkendall: Done.

src/commands/auth-import.ts
bkendall: Done.

src/commands/auth-import.ts
bkendall: Sure.

PR #4034, opened 2022-01-21
Fix bug where we enabled AR only when preview flag is off

taeold:
We had the condition backward 😮


inlined (approved):
Thanks for fixing my stupid mistake.

PR #4096, opened 2022-01-25
Emulate extensions from firebase.json

joehan:
### Description
Adds support to download, build, and emulate Extensions declared in firebase.json

### Scenarios Tested
I tested out firebase/storage-resize-images, firebase/firestore-translate-text, and a few test extensions published under joehanley, and verified that their functions run & triggered as expected.

### Sample Commands
Add an extensions section to firebase.json
```
  "extensions": {
    "translate": "firebase/firestore-translate-text",
  }
```
Add a config file to the ./extensions/ directory:
```
## extensions/translate.env
LOCATION=us-west1
LANGUAGES=en,es,de,fr
COLLECTION_PATH=translations
INPUT_FIELD_NAME=input
OUTPUT_FIELD_NAME=translated
```
Then, run an emulator command:
```
firebase emualtors:start -P my-project
```

### DIFF
--- a/src/emulator/controller.ts
+++ b/src/emulator/controller.ts
@@ -447,15 +447,15 @@ export async function startAll(options: EmulatorOptions, showUI: boolean = true)
     }
 
     const account = getProjectDefaultAccount(options.projectRoot);
-    const extensionEmulator = new ExtensionsEmulator({options});
-    const extensionsBackends = await extensionEmulator.getEmulatableBackends();
+    const extensionEmulator = new ExtensionsEmulator({ options });
+    const extensionsBackends = await extensionEmulator.getExtensionBackends();
     const emulatableBackends: EmulatableBackend[] = [
       {
         functionsDir,
         env: {
           ...options.extensionEnv,
         },
-        // TODO: predefinedTriggers and nodeMajorVersion are here to support ext:dev:emulators:* commands. 
+        // TODO: predefinedTriggers and nodeMajorVersion are here to support ext:dev:emulators:* commands.
         // Ideally, we should handle that case via ExtensionEmulator.
         predefinedTriggers: options.extensionTriggers as ParsedTriggerDefinition[] | undefined,
         nodeMajorVersion: parseRuntimeVersion(
--- a/src/emulator/controller.ts
+++ b/src/emulator/controller.ts
@@ -332,6 +332,7 @@ export async function startAll(options: EmulatorOptions, showUI: boolean = true)
     );
   }
   const hubLogger = EmulatorLogger.forEmulator(Emulators.HUB);
+  // TODO (b/217189992): Support '--only extensions` & add extensions to this logging.
   hubLogger.logLabeled("BULLET", "emulators", `Starting emulators: ${targets.join(", ")}`);
 
   const projectId: string = getProjectId(options) || ""; // TODO: Next breaking change, consider making this fall back to demo project.
--- a/src/emulator/controller.ts
+++ b/src/emulator/controller.ts
@@ -24,7 +24,7 @@ import { DatabaseEmulator, DatabaseEmulatorArgs } from "./databaseEmulator";
 import { FirestoreEmulator, FirestoreEmulatorArgs } from "./firestoreEmulator";
 import { HostingEmulator } from "./hostingEmulator";
 import { FirebaseError } from "../error";
-import { getProjectId, needProjectId } from "../projectUtils";
+import { getProjectId, needProjectId, getAliases, needProjectNumber } from "../projectUtils";
 import { PubsubEmulator } from "./pubsubEmulator";
 import * as commandUtils from "./commandUtils";
 import { EmulatorHub } from "./hub";
@@ -447,8 +447,6 @@ export async function startAll(options: EmulatorOptions, showUI: boolean = true)
     }
 
     const account = getProjectDefaultAccount(options.projectRoot);
-    const extensionEmulator = new ExtensionsEmulator({ options });
-    const extensionsBackends = await extensionEmulator.getExtensionBackends();
     const emulatableBackends: EmulatableBackend[] = [
       {
         functionsDir,
@@ -462,8 +460,23 @@ export async function startAll(options: EmulatorOptions, showUI: boolean = true)
           options.extensionNodeVersion || options.config.get("functions.runtime")
         ),
       },
-      ...extensionsBackends,
     ];
+    if (options.config.has("extensions")) {
+      // TODO: This should not error out when called with a fake project.
+      const projectNumber = await needProjectNumber(options);
+      const aliases = getAliases(options, projectId);
+
+      const extensionEmulator = new ExtensionsEmulator({
+        projectId,
+        projectDir: options.config.projectDir,
+        projectNumber,
+        aliases,
+        extensions: options.config.get("extensions"),
+      });
+      const extensionsBackends = await extensionEmulator.getExtensionBackends();
+      emulatableBackends.push(...extensionsBackends);
+    }
+
     // TODO(b/213241033): Figure out how to watch for changes to extensions .env files & reload triggers when they change.
     const functionsEmulator = new FunctionsEmulator({
       projectId,
--- a/src/emulator/controller.ts
+++ b/src/emulator/controller.ts
@@ -455,7 +455,7 @@ export async function startAll(options: EmulatorOptions, showUI: boolean = true)
         env: {
           ...options.extensionEnv,
         },
-        // TODO: predefinedTriggers and nodeMajorVersion are here to support ext:dev:emulators:* commands.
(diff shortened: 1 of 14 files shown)


src/test/emulators/extensionsEmulator.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { expect } from "chai";
+
+import { ExtensionsEmulator } from "../../emulator/extensionsEmulator";
+
+describe("Extensions Emulator", () => {
+  describe("getEmulatableBackends", () => {
+    // TODO: Write some tests for this (or maybe just write functional tests?)
```
joehan, line 7: Working on these now, just a ironing out a few final details. I want to include these in this PR

src/extensions/emulator/optionsHelper.ts
```diff
@@ -43,6 +43,29 @@ export async function buildOptions(options: any): Promise<any> {
...
+  params: Record<string, string>
+): Promise<{
+  nodeMajorVersion: number;
+  extensionTriggers: ParsedTriggerDefinition[];
+}> {
+  const spec = await specHelper.readExtensionYaml(extensionDir);
+  const functionResources = specHelper.getFunctionResourcesWithParamSubstitution(
+    spec,
+    params
+  ) as Resource[];
```
taeold, line 58: nit* it feels awkward that we need a type assertion here.

src/test/emulators/extensionsEmulator.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+import { ExtensionsEmulator } from "../../emulator/extensionsEmulator";
+import { EmulatableBackend } from "../../emulator/functionsEmulator";
+import * as planner from "../../deploy/extensions/planner";
+
+describe("Extensions Emulator", () => {
+  describe("toEmulatableBackends", () => {
+    let previousCachePath: string | undefined;
+    beforeEach(() => {
+      previousCachePath = process.env.FIREBASE_EXTENSIONS_CACHE_PATH;
+      process.env.FIREBASE_EXTENSIONS_CACHE_PATH = "./extensions";
```
taeold, line 12: nit* It looks like we expect each test case to set ^ instead, so this default value is never used. That is fine - I was just a bit confused.

src/emulator/extensionsEmulator.ts
```diff
@@ -0,0 +1,141 @@
...
+import { getExtensionFunctionInfo } from "../extensions/emulator/optionsHelper";
+
+export interface ExtensionEmulatorArgs {
+  projectId: string;
+  projectNumber: string;
+  aliases?: string[];
+  extensions: Record<string, string>;
+  projectDir: string;
+}
+// TODO: Consider a different name, since this does not implement the EmulatorInstance interface
```
taeold, line 20: In theory, I wanted the emulated extensions to be just a type of EmulatableBackend and not another type of EmulatorInstance.

There are some problems with that ideal that you brought up here e.g. it's a lot nicer to have logging bulleted by "extensions" rather then "functions" when performing extensions related operations, but as you noted in the comments, we end up in a confusing state where ExtensionsEmulator is not really an emulator type because it relies heavily on the FunctionsEmulator for runtime things.

I think the alternative here is to implement "runtime delegate for extensions" (introduced to the emulator in https://github.com/firebase/firebase-tools/pull/4012) and have that be responsible for validating, building, and watching the source to load up in the emulator (https://github.com/firebase/firebase-tools/blob/d16a978df1be865a402a9713b29f75545b9384f4/src/deploy/functions/runtimes/index.ts#L53).

I'm certain there will be wrinkles in trying to implement that for the extensions emulator, but that might be an exercise that's worth doing (at least it was for the Functions side since now we use mostly same code for "deploying" functions to both the emulator and the production).

Happy to chat offline - I'm a bit cautious here since what I'm asking here is a large consideration and I don't know what other considerations you have in mind (e.g your launch schedule?)

src/emulator/download.ts
```diff
@@ -43,6 +43,34 @@ export async function downloadEmulator(name: DownloadableEmulators): Promise<voi
...
+  const sourceCodeZip = await downloadUtils.downloadToTmp(sourceDownloadUri);
+  await unzip(sourceCodeZip, targetDir);
+
+  EmulatorLogger.forExtension(extensionVersionRef).logLabeled(
+    "BULLET",
+    "extensions",
+    `Downloaded to ${targetDir}...`
+  );
+  // TODO: We should not need to do this wait
+  // However, when I remove this, unzipDir doesn't contain everything yet.
```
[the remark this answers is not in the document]
joehan: Yep - this code was actually from hackweek 2021, and i spent like 2 of the 5 days trying to figure out what was happening here. I haven't yet revisited why this is happening, but I'm planning to do so before we merge this to master.

src/emulator/extensionsEmulator.ts
joehan: We should definitely chat offline on this, but I largely agree with your points here and think it's reasonable to refactor this. However, for launch planning purposes, I'd prefer to do so in a later PR (before merging to master). This would unblock the emulator UI work & let them test with something more realistic than an API spec. WDYT?

elvisun: please add me if you were to chat :)

src/test/emulators/extensionsEmulator.spec.ts
joehan: Nope, this was sloppy from me  - I was moving this bit around while I was trying to debug this test & forgot to clean it up.

src/extensions/emulator/specHelper.ts
```diff
@@ -124,16 +105,5 @@ export function getNodeVersion(resources: Resource[]): string {
...
-  if (_.includes(versions, "nodejs6")) {
-    EmulatorLogger.forEmulator(Emulators.FUNCTIONS).logLabeled(
-      "WARN",
-      "extensions",
-      "Node 6 is deprecated. We recommend upgrading to a newer version."
-    );
-    return "6";
-  }
-  return "8";
+  return Math.max(...versions);
```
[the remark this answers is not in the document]
joehan: Hmmm - I like the idea of adding enforcement on what node versions extensions can be published with, but I think we should do it at publish time, not in the emulator. In this case, the user doesn't control what version of node their extensions target, so I think it better to err on the side of just using the version the extension asks for.

I would be in favor of adding a backend check during publishExtensionVersion that errors out if someone tries to publish with an odd Node version (or one that is not yet supported by GCF)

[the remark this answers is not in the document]
elvisun: agree with adding validations at publish time. I guess I'm just confused why we need the `Math.max(...versions);` So today do we allow functions with multiple node versions running under one extension? And can multiple emulated extensions use different node versions?

src/emulator/controller.ts
```diff
@@ -43,6 +43,7 @@ import { getDefaultDatabaseInstance } from "../getDefaultDatabaseInstance";
 import { getProjectDefaultAccount } from "../auth";
 import { Options } from "../options";
 import { ParsedTriggerDefinition } from "./functionsEmulatorShared";
+import { ExtensionsEmulator } from "./extensionsEmulator";
```
elvisun, line 47: Can you add a todo at line 335 (github don't let me comment on that line..), we need to show up on that list 
```
lihes-macbookpro2:ext-emu lihes$ firebase emulators:start
i  emulators: Starting emulators: functions, firestore
⚠  functions: The following emulators are not running, calls to these services from the Functions emulator will affect production: auth, database, hosting, pubsub, storage
i  extensions: firebase/delete-user-data@0.1.11 already downloaded...
...
```

elvisun: And also supporting the --only flag? Related note, if someone does `--only extensions`, we need to bring up the functions emulator but probably shouldn't bring up user's functions

src/extensions/emulator/specHelper.ts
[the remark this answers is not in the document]
joehan: You can technically set different node versions for each function (since the underlying resources are totally separate) - however, as a bit of a hack/simplification, we decided not to support that in the emulator because its a bad pattern and an edge case. To handle the case where a single extensions asks for multiple versions of node, we just default to the highest version they ask for. 

We _do_ support emulating different extensions in different versions of node. I think this is probably a hard requirement, too - we want users to be able to emulate any combo of extensions, and some won't work correctly (or at all) on the wrong version of node.

I'd also chose to do Math.max(versions) instead of just capping at 14 so that this doesn't become an annoyance to maintain. If we cap at 14, we need to remember to come back here when 16, 18, 20, etc are released.

src/emulator/controller.ts
joehan: Good catch - adding a TODO here and filed b/217189992 to track it

PR #4100, opened 2022-01-26
Emulator: change default-bucket to the actual default bucket

colerogers:
Fixes a regression that was caused by adding bucket filtering in the emulator.


taeold (approved):
How was this caught? Did we get a failing test or was there a user reported issue?

inlined (approved):

inlined:
Unhappy user reported issue. The customer was trying to listen to `functions.storage.object().onFinalize()` but that didn't work because the default bucket the emulator creates is `default-bucket`. That used to not be a problem because we fired events for every bucket on a write to any of them. Now that we added filtering, we broke customers. Also CC @yuchenshi as the new emulator owner.

yuchenshi (approved):

PR #4101, opened 2022-01-26
Add validation and API restrictions to concurrency

inlined:
To prevent price regressions, we are introducing fractional CPU for gen 2 and keeping the same memory -> CPU mapping. This means that functions with fewer than 2GB memory will have less than one CPU. While the Cloud Run team has agreed to support fractional CPU for our use cases, they have chosen to disable concurrency when using < 1 CPU.

This means that functions with fewer than 2GB will not support concurrency. We've removed API calls to set concurrency by default for functions created with < 2GB RAM and to added client-side validation to error out when we see these cases (which will otherwise error out with cryptic messages with the API call.


src/test/deploy/functions/validate.spec.ts
```diff
@@ -113,4 +114,77 @@ describe("validate", () => {
...
+    it("Allows endpoints with enough mem and no concurrency", () => {
+      for (const mem of [2 << 10, 4 << 10, 8 << 10] as backend.MemoryOptions[]) {
+        const ep: backend.Endpoint = {
+          ...ENDPOINT_BASE,
+          availableMemoryMb: mem,
+        };
+        expect(() => validate.endpointsAreValid(backend.of(ep))).to.not.throw;
+      }
+    });
+
```
taeold, line 168: think only missing case is "Allows endpoints with enough mem _and concurrency_"

taeold (approved):

PR #4107, opened 2022-01-27
Use relative redirect URLs for resumable storage uploads.

yuchenshi:
### Description

Fixes https://github.com/firebase/firebase-admin-node/issues/1562

### Scenarios Tested

See tests enabled.

### Sample Commands

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ node_modules
 /.nyc_output
 firebase-debug.log
 firebase-debug.*.log
+ui-debug.log
 npm-debug.log
 yarn.lock
 .npmrc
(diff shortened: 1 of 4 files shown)


yuchenshi:
Actually this breaks on CI with `Only absolute URLs are supported`. Will investigate and may drastically change my approach. Converting to draft now

bkendall:
> Actually this breaks on CI with `Only absolute URLs are supported`. Will investigate and may drastically change my approach. Converting to draft now

yay integration test? I think they're _absolutely_ necessary in cases like this

vajahath:
Please let me know if I can do something here to support. If this PR is already in pipeline, all good. :+1:

yuchenshi:
Closing in favor of #4409

yuchenshi:
Closing in favor of #4409

PR #4110, opened 2022-01-29
Remove @types/winston dependency

andipaetzold:
### Description

[`winston`](https://www.npmjs.com/package/winston) provides its own type definitions and [`@types/winston`](https://www.npmjs.com/package/@types/winston) is only a stub. The dependency is therefore not needed anymore.

See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/27002 and https://github.com/winstonjs/winston/pull/1374

This removes  the deprecation warning from the logs when installing `firebase-tools`:
```
npm WARN deprecated @types/winston@2.4.4: This is a stub types definition. winston provides its own type definitions, so you do not need this installed.
```

### Scenarios Tested

* `npm test`
* `npm run build`

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -116,7 +116,6 @@
         "@types/triple-beam": "^1.3.0",
         "@types/unzipper": "^0.10.0",
         "@types/uuid": "^8.3.1",
-        "@types/winston": "^2.4.4",
         "@types/ws": "^7.2.3",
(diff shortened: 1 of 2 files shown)


bkendall:
Seems reasonable to me. Let's make sure the tests pass and we can merge it.

bkendall (approved):

bkendall:
Thanks!

PR #4120, opened 2022-02-01
Fix bug in ext:install - only firebase extensions can be installed

elvisun:
### Description

Currently `firebase ext:install -i` only lists extensionName in the interactive flow, so any non-firebase published extensions will then be treated as firebase extensions, which then breaks the install flow.

Before:
Screen Shot 2022-02-01 at 12 11 37 PM
Screen Shot 2022-02-01 at 11 51 08 AM

After:
Screen Shot 2022-02-01 at 12 02 25 PM
Screen Shot 2022-02-01 at 12 02 36 PM

`infoInstallByReference` handles both ref and extensionName, so this doesn't impact the installation code path after line 306

### DIFF
--- a/src/extensions/utils.ts
+++ b/src/extensions/utils.ts
@@ -33,12 +33,14 @@ export function convertExtensionOptionToLabeledList(options: ParamOption[]): Lis
 export function convertOfficialExtensionsToList(officialExts: {
   [key: string]: RegistryEntry;
 }): ListItem[] {
-  return _.map(officialExts, (entry: RegistryEntry, key: string) => {
+  const l = _.map(officialExts, (entry: RegistryEntry, key: string) => {
     return {
       checked: false,
       value: `${entry.publisher}/${key}`,
     };
   });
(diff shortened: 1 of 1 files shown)


src/extensions/utils.ts
```diff
@@ -33,12 +33,14 @@ export function convertExtensionOptionToLabeledList(options: ParamOption[]): Lis
 export function convertOfficialExtensionsToList(officialExts: {
   [key: string]: RegistryEntry;
 }): ListItem[] {
-  return _.map(officialExts, (entry: RegistryEntry, key: string) => {
+  const l = _.map(officialExts, (entry: RegistryEntry, key: string) => {
```
joehan, line 36: Nit: Prefer using native functions (`officialExts.map(...).sort(...)`) instead of lodash

joehan (approved):

src/extensions/utils.ts
elvisun: officialExts is an obj so there's no native map, could do object.entries but since this code will get deleted soon I'm leaving it as is

PR #4123, opened 2022-02-02
Updates --no-localhost to use new Auth proxy service.

mbleigh:
This replaces the existing `--no-localhost` flow with an improved Firebase-branded experience with enhanced security and usability.


src/auth.ts
```diff
@@ -406,25 +419,53 @@ async function respondWithFile(
...
-  const callbackUrl = getCallbackUrl();
-  const authUrl = getLoginUrl(callbackUrl, userHint);
+function urlsafeBase64(base64string: string) {
+  return base64string.replace(/\+/g, "-").replace(/=+$/, "").replace(/\//g, "_");
+}
+
+const authProxyClient = new apiv2.Client({
+  urlPrefix: api.authProxyOrigin,
+  auth: false,
+});
```
bkendall, line 429: For some reason, this is making the emulator tests fail? I'm not sure why this isn't working in those tests (but works fine in practice)

bkendall (changes requested):
1. Update to `master`
2. Changelog entry, please, just so "the login experience changed" isn't a surprise if folks come looking.

bkendall:
Running it myself, found a couple bits:

1. The webpage says "confirm you ran `firebase login --remote` still - we should either stick with `--no-localhost` or we should reconsider breaking/deprecating `--no-localhost`.
2. It's a tiny detail, but I did run the flag with `--reauth` and _that_ is not listed on the page. I don't know if this is a common enough path to worry about, but thought I'd bring it up. (ah, but my session ID matches, which is super comforting)

Otherwise, worked great.

mbleigh:
I think I'll leave `--reauth` as a "future work if anyone complains" - I can imagine how I'd solve it but I don't think it's worth the effort.

Fixed the emu test and added a changelog.

bkendall (approved):

PR #4124, opened 2022-02-02
Enable parsing function triggers via functions.yaml

taeold:
With this change, CF3 users who've opted in for functionsv2 preview will now have their function triggers be parsed via the new container contract based mechanism. This relies on having the latest version of the Firebase Functions SDK, so we will fall back to the good ol' parseTrigger script when we detect an incompatible version.

The PR includes various other changes we've made to the Manifest - e.g. explicit callableTrigger type, minor tweaks to the vpc trigger setting,  supplying correct `requiredAPIs` in parseTrigger script, etc.)


src/deploy/functions/release/reporter.ts
```diff
@@ -233,7 +233,12 @@ export function triggerTag(endpoint: backend.Endpoint): string {
     return `${prefix}.taskQueue`;
   }
 
+  if (backend.isCallableTriggered(endpoint)) {
+    return `${prefix}.callable`;
+  }
+
   if (backend.isHttpsTriggered(endpoint)) {
+    // NOTE: Legacy trigger annotation relies on a special label to differentiate http vs callable triggers.
```
inlined, line 241: What if we instead modified triggerParser to just create a callableTriggered function so we don't have two deployment paths?

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -131,6 +133,9 @@ function parseEndpoints(
       });
       triggered = { httpsTrigger: {} };
       copyIfPresent(triggered.httpsTrigger, ep.httpsTrigger, "invoker");
+    } else if (backend.isCallableTriggered(ep)) {
+      triggered = { callableTrigger: {} };
+      ep.labels = { ...ep.labels, "deployment-callabled": "true" };
```
inlined, line 138: Rather than setting the labels here, let's set them in the gcp/cloudfunctions and gcp/cloudfunctionsv2 `endpointToFunction` methods.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -145,6 +150,7 @@ function parseEndpoints(
         maxRetryDuration: "string",
       });
       triggered = { scheduleTrigger: ep.scheduleTrigger };
+      ep.labels = { ...ep.labels, "deployment-scheduled": "true" };
```
inlined, line 153: Same as above. The label should be part of the Function, not necessarily the Endpoint (though it isn't a disaster if it is)

src/deploy/functions/runtimes/node/index.ts
```diff
@@ -1,16 +1,28 @@
...
+import { previews } from "../../../../previews";
+import { logLabeledWarning } from "../../../../utils";
 import * as backend from "../../backend";
 import * as runtimes from "..";
 import * as validate from "./validate";
 import * as versioning from "./versioning";
 import * as parseTriggers from "./parseTriggers";
+import * as discovery from "../discovery";
 
+const MIN_FUNCTIONS_CONTROL_API_VERSION = "3.17.0";
```
inlined, line 21: Be careful about dependency ordering here too. Don't release this CLI version until you _know_ 3.17.0 supports this endpoint and it's been bug bashed.

src/deploy/functions/runtimes/node/index.ts
```diff
@@ -80,10 +92,77 @@ export class Delegate {
...
+  serve(
+    port: number,
+    adminPort: number,
+    envs: backend.EnvironmentVariables
+  ): Promise<() => Promise<void>> {
+    const childProcess = spawn("./node_modules/.bin/firebase-functions", [this.sourceDir], {
+      env: {
+        ...envs,
+        PORT: port.toString(),
+        STACK_CONTROL_API_PORT: adminPort.toString(),
```
inlined, line 104: Just STACK_CONTROL_API (or FUNCTIONS_CONTROL_API according to internal discussion) = true. Early prototypes spun up two ports, but Python cannot do this, so we're using __ instead.

src/deploy/functions/runtimes/node/index.ts
```diff
@@ -80,10 +92,77 @@ export class Delegate {
     return Promise.resolve(() => Promise.resolve());
   }
 
+  serve(
+    port: number,
+    adminPort: number,
```
inlined, line 97: You can collapse port and adminPort. That's a prototype holdover that didn't scale to Python

src/gcp/cloudfunctionsv2.ts
```diff
@@ -416,6 +414,16 @@ export function functionFromEndpoint(endpoint: backend.Endpoint, source: Storage
...
+  if (endpoint.vpc) {
+    proto.renameIfPresent(gcfFunction.serviceConfig, endpoint.vpc, "vpcConnector", "connector");
+    proto.renameIfPresent(
+      gcfFunction.serviceConfig,
+      endpoint.vpc,
+      "vpcConnectorEgressSettings",
+      "egressSettings"
+    );
+  }
+
```
inlined, line 426: add `else if (backend.isCallableTriggered(endoint)) { gcfFunction.labels = { ...gcfFucntion.labels, "deployment-callable": true }; }` below

src/gcp/cloudfunctions.ts
```diff
@@ -582,11 +588,17 @@ export function functionFromEndpoint(
...
-
+  if (endpoint.vpc) {
+    proto.renameIfPresent(gcfFunction, endpoint.vpc, "vpcConnector", "connector");
+    proto.renameIfPresent(
+      gcfFunction,
+      endpoint.vpc,
+      "vpcConnectorEgressSettings",
+      "egressSettings"
+    );
+  }
```
inlined, line 602: add `else if (backend.isCallableTriggered(endoint)) { gcfFunction.labels = { ...gcfFucntion.labels, "deployment-callable": true }; }` below

src/test/deploy/functions/runtimes/discovery/v1alpha1.spec.ts
```diff
@@ -311,7 +311,11 @@ describe("backendFromV1Alpha1", () => {
           },
         },
       };
-      const expected = backend.of({ ...DEFAULTED_ENDPOINT, scheduleTrigger });
+      const expected = backend.of({
+        ...DEFAULTED_ENDPOINT,
+        scheduleTrigger,
+        labels: { "deployment-scheduled": "true" },
```
inlined, line 317: This LOC is why we add the labels in `functionFromEndpoint`, so we don't have any logic that breaks if tests are missing a label. DRY and make the one field (scheduleTrigger) mean the one thing (that this function is scheduled)

inlined (approved):
Only pushback:
1. Labels should be on the function not the endpoint
2. The "__" prefix to the control API was meant to replace using 2 ports
3. We should probably use functions.yaml and FUNCTION_CONTROL_API=true instead of STACK because we may diverge

src/deploy/functions/release/reporter.ts
taeold: I can see that it will be cleaner, but this requires quite a bit of plumbing here and in the Emulator side of things.

I can work on it as a follow up to the PR instead.

PR #4125, opened 2022-02-02
Pass environment variable of .env files during deployment

IchordeDionysos:
### Description

Make it possible to configure the function region or other function configuration using an environment variable defined in a `.env` file.

During the deployment, the Firebase CLI loads the source code to get all the configurations needed to deploy the function.
To make sure that all configuration is present, the CLI previously added some environment variables that contain the function config accessible via `functions.config()`.

So it was possible to do something like this:

```
exports.helloWorld = functions.region(functions.config().function.region).https.onRequest((request, response) => {
});
```

With the new environment variable based approach this does not work anymore:

This function fails during deployment, as the `REGION` environment variable is not present.
```
if (!process.env.REGION) throw new Error('REGION not set');

exports.helloWorld = functions.region(process.env.REGION).https.onRequest((request, response) => {
});
```

---
This PR passes the environment variable defined in the `.env` file on the parser that detects which functions are present and how they should be deployed.

This issue also occurs if an environment variable somewhere else is accessed when loading the index.js file.

### Scenarios Tested

Deployment of this function works with the `dotenv` preview enabled:
```
if (!process.env.REGION) throw new Error('REGION not set');
exports.helloWorld = functions.region(process.env.REGION).https.onRequest((request, response) => {
});
```
And the following `.env` file:
```
REGION=europe-west1
```

### Sample Commands

### DIFF
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -104,6 +104,7 @@ export async function prepare(
   };
   const userEnvs = functionsEnv.loadUserEnvs(userEnvOpt);
   const usedDotenv = hasDotenv(userEnvOpt);
+  const environmentVariables = { ...userEnvs, ...firebaseEnvs };
   const tag = hasUserConfig(runtimeConfig)
     ? usedDotenv
       ? "mixed"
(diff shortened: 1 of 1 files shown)


akhilnarayanan1:
Why do we need .env when firebase already has this [https://firebase.google.com/docs/functions/config-env](https://firebase.google.com/docs/functions/config-env)

clucher91 (approved):

PR #4127, opened 2022-02-03
Helping prevention of accidental directory exclusion when using the --export-on-exit option

Muritavo:
### Description
Adding condition to prevent the user from setting the **--export-on-exit** option to the CWD, inflicting in directory exclusion after stopping the emulators

### Scenarios Tested
The firebase emulators:start command should fail now if the **--export-on-exit** is set to the **CWD** or **parent directories**

Commands tested:

- Current dir (SHOULD FAIL): `firebase emulators:start --export-on-exit ./`
- Current dir parent (SHOULD FAIL): `firebase emulators:start --export-on-exit ../`
- Implicit --export-on-exit via setting the --import flag  (SHOULD FAIL): `firebase emulators:start --export-on-exit ./dump-db`

- A sub directory (SHOULD PASS): `firebase emulators:start --export-on-exit ./dump-db`
- Implicit --export-on-exit sub directory (SHOULD PASS): `firebase emulators:start --import ./dump-db --export-on-exit`

### Sample Commands

Should not allow:
```cmd
firebase emulators:start --export-on-exit ./
```

One that should allow:
```cmd
firebase emulators:start --export-on-exit ./dump-db
```

### DIFF
--- a/src/emulator/commandUtils.ts
+++ b/src/emulator/commandUtils.ts
@@ -45,6 +45,8 @@ export const EXPORT_ON_EXIT_USAGE_ERROR =
   `"${FLAG_EXPORT_ON_EXIT_NAME}" must be used with "${FLAG_IMPORT}"` +
   ` or provide a dir directly to "${FLAG_EXPORT_ON_EXIT}"`;
 
+export const EXPORT_ON_EXIT_CWD_DANGER = `"${FLAG_EXPORT_ON_EXIT_NAME}" must point to a folder outside the CWD path`;
+
 export const FLAG_UI = "--ui";
 export const DESC_UI = "run the Emulator UI";
 
@@ -202,6 +204,10 @@ export function setExportOnExitOptions(options: any) {
       // options.exportOnExit might be an empty string when used as:
       // firebase emulators:start --debug --import '' --export-on-exit ''
       throw new FirebaseError(EXPORT_ON_EXIT_USAGE_ERROR);
+    }
+
+    if (path.resolve(".").includes(path.resolve(options.exportOnExit))) {
+      throw new FirebaseError(EXPORT_ON_EXIT_CWD_DANGER);
     }
   }
   return;
--- a/src/test/emulators/commandUtils.spec.ts
+++ b/src/test/emulators/commandUtils.spec.ts
@@ -34,6 +34,16 @@ describe("commandUtils", () => {
     });
   });
 
+  it("Should disallow the user to set the current folder via the --import flag", () => {
+    expect(() => testSetExportOnExitOptions({ import: ".", exportOnExit: true })).to.throw(
+      EXPORT_ON_EXIT_CWD_DANGER
+    );
+    const cwdSubDir = join(".", "some-dir");
+    expect(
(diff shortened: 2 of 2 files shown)


yuchenshi:
Thanks for the PR! Getting someone to take a look

src/emulator/commandUtils.ts
```diff
@@ -45,6 +45,8 @@ export const EXPORT_ON_EXIT_USAGE_ERROR =
   `"${FLAG_EXPORT_ON_EXIT_NAME}" must be used with "${FLAG_IMPORT}"` +
   ` or provide a dir directly to "${FLAG_EXPORT_ON_EXIT}"`;
 
+export const EXPORT_ON_EXIT_CWD_DANGER = `"${FLAG_EXPORT_ON_EXIT_NAME}" must point to a folder outside the CWD path`;
```
yuchenshi, line 48:
```suggestion
export const EXPORT_ON_EXIT_CWD_DANGER = `"${FLAG_EXPORT_ON_EXIT_NAME}" must not point to the current directory or parents. Please choose a new/dedicated directory for exports.`;
```

src/emulator/commandUtils.ts
```diff
@@ -203,6 +205,10 @@ export function setExportOnExitOptions(options: any) {
       // firebase emulators:start --debug --import '' --export-on-exit ''
       throw new FirebaseError(EXPORT_ON_EXIT_USAGE_ERROR);
     }
+
+    if (path.resolve(".").includes(path.resolve(options.exportOnExit))) {
```
yuchenshi, line 209:
```suggestion
    if (path.resolve(".").startsWith(path.resolve(options.exportOnExit))) {
```

Prevents false positives like `/a/b/c/d` and `/b/c/d`

yuchenshi (changes requested):
Overall I think is a good stop-gap solution and I'm happy to merge this once the comments are addressed. (It would be even better though if we can find a way to export without nuking the target directory, but that is a longer term solution.)

src/emulator/commandUtils.ts
Muritavo: Nice observations @yuchenshi. I've also included a test for this cenario

src/emulator/commandUtils.ts
Muritavo: Thanks, it is much clearer now for the user what it needs to be done.

Muritavo:
> Overall I think is a good stop-gap solution and I'm happy to merge this once the comments are addressed. (It would be even better though if we can find a way to export without nuking the target directory, but that is a longer term solution.)

Yup, I just built this PR in like 1 hour to reduce the chances for people to accidentally exclude their projects by runnning the command without prior understanding. I almost lost 2 weeks of work because of this. If someone would try with their root user folder the destruction would be massive.

If I get the time, I will try to dive into the core of firebase persistency to fix this in a definitive manner.

Thanks for pointing out which PR introduced the issue, that will save up some time :)

yuchenshi (approved):

yuchenshi:
Thanks again for your contribution! This PR is now merged and will be included in the next release.

PR #4132, opened 2022-02-05
Fix get and delete endpoints for storage not accepting the documented endpoint: /storage/v1/b/bucket/o/object

soonick:
### Description

While I was testing getting and deleting pictures from the firestore emulator I noticed that I was getting errors about the object not being found.

I'm noticed that my client (golang client) was making requests to:

```
/storage/v1/b/my-bucket/o/ikEYroV6zfdX9q3bVWqp%2F611c9b5a-2145-4d75-8296-c9d25f09ecab.jpeg?alt=json&prettyPrint=false&projection=full
```
I wanted to see what was happening in the emulator and noticed that there is no listener for those endpoints. Suspecting it could be an issue with the golang client I checked the cloud storage documentation for the official paths:

https://cloud.google.com/storage/docs/json_api/v1/objects/get
https://cloud.google.com/storage/docs/json_api/v1/objects/delete

They seem to match what the golang client is doing. 

I made the change on this pull request and things started working.

### Scenarios Tested

This is a snippet of the code that was failing and works after my changes:

```
bucketName := "my-bucket"
bkt := client.Bucket(bucketName)

objPath := "my-folder/myobject"
attrs, err := bkt.Object(objPath).Attrs(ctx)
if err == storage.ErrObjectNotExist {
	// This is what was happening before
	log.Errorf("The object does not exist, %v", err)
	return err
}
if err != nil {
	return err
}
log.Errorf("The object exists and has attributes: %#v\n", attrs)

obj := bkt.Object(objPath)
// After fixing the endpoint above, this one was failing too
err = obj.Delete(ctx)
if err != nil {
	return err
}

return nil
```


geoah:
The file upload in go seems to be doing something similar. The endpoint in the emulator is `/upload/storage/v1/b/:bucketId/o` but the golang client seems to be trying to use `/storage/v1/upload/b/:bucketId/o`.

PR #4141, opened 2022-02-08
Adding support for --only emulators

joehan:
### Description
Adds support for `--only extensions` to emulators commands.

### Scenarios Tested
I tested `firebase emulators:start`, 
Screen Shot 2022-02-08 at 2 58 52 PM

`firebase emulators:start --only extensions`,
Screen Shot 2022-02-08 at 2 58 28 PM

`firebase emulators:start --only functions`, 

Screen Shot 2022-02-08 at 2 57 50 PM
and `firebase emulators:start --only functions,extensions,firestore,storage`.

Screen Shot 2022-02-08 at 2 55 49 PM

In all four cases, I also tested that the correct functions were running (and that functions hat should not be running were not).

### DIFF
--- a/src/commands/emulators-start.ts
+++ b/src/commands/emulators-start.ts
@@ -75,7 +75,7 @@ module.exports = new Command("emulators:start")
           const isSupportedByUi = EMULATORS_SUPPORTED_BY_UI.includes(emulator);
           // The Extensions emulator runs as part of the Functions emulator, so display the Functions emulators info instead.
           const info = EmulatorRegistry.getInfo(
-            emulator == Emulators.EXTENSIONS ? Emulators.FUNCTIONS : emulator
+            emulator === Emulators.EXTENSIONS ? Emulators.FUNCTIONS : emulator
           );
           if (!info) {
             return [emulatorName, "Failed to initialize (see above)", "", ""];
--- a/src/commands/emulators-start.ts
+++ b/src/commands/emulators-start.ts
@@ -71,10 +71,12 @@ module.exports = new Command("emulators:start")
       ...controller
         .filterEmulatorTargets(options)
         .map((emulator) => {
-          const info = EmulatorRegistry.getInfo(emulator);
           const emulatorName = Constants.description(emulator).replace(/ emulator/i, "");
           const isSupportedByUi = EMULATORS_SUPPORTED_BY_UI.includes(emulator);
-
+          // The Extensions emulator runs as part of the Functions emulator, so display the Functions emulators info instead.
+          const info = EmulatorRegistry.getInfo(
+            emulator == Emulators.EXTENSIONS ? Emulators.FUNCTIONS : emulator
+          );
           if (!info) {
             return [emulatorName, "Failed to initialize (see above)", "", ""];
           }
--- a/src/emulator/controller.ts
+++ b/src/emulator/controller.ts
@@ -46,7 +46,7 @@ import { ParsedTriggerDefinition } from "./functionsEmulatorShared";
 import { ExtensionsEmulator } from "./extensionsEmulator";
 
 async function getAndCheckAddress(emulator: Emulators, options: Options): Promise<Address> {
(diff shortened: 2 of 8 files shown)


elvisun:
https://user-images.githubusercontent.com/4635763/153089746-405815d5-9ebe-4f7d-aa97-eb0674dc5f64.png
the list of emulators not running should also include extensions

src/commands/emulators-start.ts
```diff
@@ -71,10 +71,12 @@ module.exports = new Command("emulators:start")
       ...controller
         .filterEmulatorTargets(options)
         .map((emulator) => {
-          const info = EmulatorRegistry.getInfo(emulator);
           const emulatorName = Constants.description(emulator).replace(/ emulator/i, "");
           const isSupportedByUi = EMULATORS_SUPPORTED_BY_UI.includes(emulator);
-
+          // The Extensions emulator runs as part of the Functions emulator, so display the Functions emulators info instead.
+          const info = EmulatorRegistry.getInfo(
+            emulator == Emulators.EXTENSIONS ? Emulators.FUNCTIONS : emulator
```
elvisun, line 78: === ?

src/emulator/registry.ts
```diff
@@ -51,6 +51,9 @@ export class EmulatorRegistry {
       // Functions is next since it has side effects and
       // dependencies across all the others
       functions: 1,
+      // The Extensions emulator runs on the same process as the Functions emulator
+      // so this is a no-op.
+      extensions: 1.1,
```
elvisun, line 56: should we put extensions above functions since it has dep on functions? (I supposed the reason why we want it here even it's a no-op is to be more future-proof?)

src/emulator/controller.ts
```diff
@@ -46,6 +46,10 @@ import { ParsedTriggerDefinition } from "./functionsEmulatorShared";
 import { ExtensionsEmulator } from "./extensionsEmulator";
 
 async function getAndCheckAddress(emulator: Emulators, options: Options): Promise<Address> {
+  if (emulator == Emulators.EXTENSIONS) {
```
elvisun, line 49: ===

src/emulator/controller.ts
```diff
@@ -421,6 +422,41 @@ export async function startAll(options: EmulatorOptions, showUI: boolean = true)
       options.config.src.functions.source
     );
 
+    emulatableBackends.push({
+      functionsDir,
+      env: {
+        ...options.extensionEnv,
```
elvisun, line 428: can you document why we need extensionEnv when starting functions emu?

also wondering if we can do better than this? perhaps something like 
```
...(shouldStart(options, Emulators.EXTENSIONS) ? options.extensionEnv : [])
```

elvisun (approved):

src/emulator/controller.ts
joehan: Will do. extensionsEnv is a (now very confusingly named) remnant from the local Extensions emulator commands (firebase ext:dev:emulators:*) - it is unused by the Extensions Emulator. Eventually, I want to get rid of those commands entirely and have publishers emulate the extensions they are testing by pointing an extension manifest at the local directory.

I'll go thru and rename  + comment the parts that are there only to support firebase ext:dev:emulators:*. Does extDevEnv sound like a reasonable name?

src/emulator/registry.ts
joehan: The type system actually forces us to include this here - I get compilation errors otherwise. Good point on moving it above tho, that would probably be more future proof

src/emulator/controller.ts
elvisun: I see thanks a lot for the explanation! yes that sounds like a good name :)

joehan: https://github.com/firebase/firebase-tools/pull/4150 Put these changes in a separate PR so its easy to see - i'm going to merge that into this then merge this.

Issue #4158, opened 2022-02-12
functions emulator crashes when accessing config in global space


danbraik:
### [REQUIRED] Environment info

**firebase-tools:**10.2.0

**Platform:**Debian

### [REQUIRED] Test case

initialise a project with functions emulator

functions/src/index.ts 
`
import * as functions from "firebase-functions";
const a = functions.config().stripe.stripe_secret_key;
`

add .runtimeconfig.json
`
{
"stripe": { "stripe_secret_key": "hello" }
}
`

### [REQUIRED] Steps to reproduce

$ npm run build
$ firebase emulators:start --only functions

### [REQUIRED] Expected behavior

I expect the emulator to be launched, as in previous version ( <= 10.1.5)

### [REQUIRED] Actual behavior

There is an error for accessing config during global space initialisation (it works when it is called later)

`
Error: Error occurred while parsing your function triggers.

TypeError: Cannot read property 'stripe_secret_key' of undefined
    at Object.<anonymous> (.../functions/lib/functions/src/index.js:16:36)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Module.require (internal/modules/cjs/loader.js:887:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at loadModule (.../.npm-global/lib/node_modules/firebase-tools/lib/deploy/functions/runtimes/node/triggerParser.js:10:16)
    at .../.npm-global/lib/node_modules/firebase-tools/lib/deploy/functions/runtimes/node/triggerParser.js:34:21
    at Object.<anonymous> (...npm-global/lib/node_modules/firebase-tools/lib/deploy/functions/runtimes/node/triggerParser.js:72:3)

`

spicemix:
I would like to mention that 10.2 emulator functions broke entirely for me, though they load fine, and I don't notice anything unusual in the logs, other than my functions no longer work. Some of the functions do launch and complete once, but not repeatedly, as I'm using a lot of pubsub cron tasks, also database triggers, etc. 

10.1.5 works perfectly, as does the most recent 9.x, and I'm up to date on all other dependencies. I see no errors anywhere, but functions tend to hang, and I have a long wait for them to timeout or I just force kill them along with all the emulators.

[...]

PR #4162, opened 2022-02-14
Load runtime config when loading triggers in the Functions Emulator

taeold:
Fix for https://github.com/firebase/firebase-tools/issues/4158.

I've tried to find a way to add a test for this usecase, but there isn't a framework where I can easily hook in a change like this without big investments. I'm leaning towards releasing this bugfix as is, and finding a less urgent time to add regression test.


colerogers (approved):
I would also get @inlined 's opinion on merging this as bugfix. But to me, it makes sense to merge this asap to fix the customer issue

PR #4172, opened 2022-02-16
fix issue in auth:import with JSON files

### LINKED ISSUE #4157: Auth user import doesn't seem to work
<!-- DO NOT DELETE
validate_template=true
template_path=.github/ISSUE_TEMPLATE/bug_report.md
-->

<!-- What version of the Firebase CLI (`firebase-tools`) are you using? Note that your issue may already be fixed in the latest versions. The latest version can be found at https://github.com/firebase/firebase-tools/releases -->

<!-- Output of `firebase --version` -->

**firebase-tools:** 10.1.4

<!-- e.g. macOS, Windows, Ubuntu -->

**Platform:** macOS

`firebase auth:import users.json`

Successful import.

Two issues. One is that no error details are actually displayed:

```
Processing users.json (9500 bytes)                                                                                                    
Starting importing 27 account(s).                                                                                                                             
Encountered problems while importing accounts. Details:                                                                                                       
                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                              
```

Digging into the code, it seems like this line isn't logging the error: https://github.com/firebase/firebase-tools/blob/2559abb6b8029450149caf1700f4a1f41235d376/src/accountImporter.js#L318

I don't know the details of how Firebase logging works, but when I switched it to console.log for debugging purposes I then saw these errors:

```
[                                                                                                                                                             
  { index: 0, message: 'localId is missing' },                                                                                                                
  { index: 1, message: 'localId is missing' },                                                                                                                
  { index: 2, message: 'localId is missing' },                                                                                                                
  { index: 3, message: 'localId is missing' },                                                                                                                
  { index: 4, message: 'localId is missing' },                                                                                                                
  { index: 5, message: 'localId is missing' },                                                                                                                
  { index: 6, message: 'localId is missing' },                                                                                                                
  { index: 7, message: 'localId is missing' }, 
  . . .
```

Looking at my specific file `localId` _was_ specified (I got it via the corresponding auth export command). Digging it appears that this section of code is the issue:

https://github.com/firebase/firebase-tools/blob/2559abb6b8029450149caf1700f4a1f41235d376/src/commands/auth-import.ts#L111-L117

Specifically, it calls `validateUserJson` which returns an empty object if there are no validation errors. But this value is used as the user, so each "user" is just an empty object which get rejected by the API. By switching the code to push `value` instead of the `user` value: `currentBatch.push(value)`, I was able to successfully import my auth users:

```
Starting importing 27 account(s).
✔  Imported successfully.
```

cc @bkendall It seems possibly introduced by this change https://github.com/firebase/firebase-tools/commit/b705926cf4a77a58998589581ec95a605d63122b

bkendall:
### Description

Fixes issue where the user was not being correctly put into the batch from a JSON file.

Fixes #4157

### Scenarios Tested

`auth:import users.json`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
 - Fixes an issue where ext:list was not printing out information about installed Extension instances.
 - Fixes issue where Functions Emulator crashed when parsing triggers if accessing functions config values.
 - `firebase emulators:start --export-on-exit <dir>` now rejects overwriting the current directory or parents (#4127).
+- Fixes issue where `auth:import` would fail when reading a JSON file. (#4157)
--- a/src/commands/auth-import.ts
+++ b/src/commands/auth-import.ts
@@ -114,7 +114,7 @@ module.exports = new Command("auth:import [dataFile]")
             if (err) {
               throw new FirebaseError(`Validation Error: ${err}`);
             }
-            currentBatch.push(user);
+            currentBatch.push(value);
             if (currentBatch.length === MAX_BATCH_SIZE) {
               batches.push(currentBatch);
               currentBatch = [];


yuchenshi (approved):
This is a highly `value`d patch. Thanks!

PR #4149, opened 2022-02-09
Release CF3's support for environment variables and secrets

taeold:
This is a combination of several PRs, all separately reviewed.

https://github.com/firebase/firebase-tools/pull/4012
https://github.com/firebase/firebase-tools/pull/3959
https://github.com/firebase/firebase-tools/pull/4021
https://github.com/firebase/firebase-tools/pull/4105
https://github.com/firebase/firebase-tools/pull/4026
https://github.com/firebase/firebase-tools/pull/4108
https://github.com/firebase/firebase-tools/pull/4022
https://github.com/firebase/firebase-tools/pull/4130
https://github.com/firebase/firebase-tools/pull/4148

PR #4174, opened 2022-02-16
Use trigger.entryPoint instead of trigger.name

joehan:
### Description
Fixed the errors that arose when merging in the Function emulator changes from master. PTAL at the second commit by itself here - that is the actual code that I needed to change to get this working.

The problem was that #4149 subtly changed the behavior of the function emulator. Previous, we looked for the functions code at `EmulatedTriggerDefinition.entryPoint`. After those changes, we started looking for the functions code at `EmulatedTriggerDefinition.name`. AFAICT these are always the same value for CF3, but for extensions, they are always different.

### Scenarios Tested
I can now successfuly execute functions with running into the "Failed to initialize and load trigger. This shouldn't happen: Failed to find function ext.httptest.greetTheWorld in the loaded module" error.


taeold (commented):
One suggestion since the PR diff is actually HUGE!

1. Fix the issue upstream
2. Mer master into the launch branch

Issue #4146, opened 2022-02-09
Improve error message when function deployment fails because the packaged function source is >100MB


taeold:
In https://github.com/firebase/firebase-tools/issues/4134, we had several users report function deployment failure with a cryptic error messages:

```
Error: Unable to parse JSON: SyntaxError: Unexpected token < in JSON at position 0
```

We think the root cause of this issue is because GCF limits the size of the packaged function source to be <100MB per https://cloud.google.com/functions/quotas#resource_limits.

It would be great if the deployment process can be improved to check for these resource constraints and log more useful error messages.

PR #4180, opened 2022-02-17
Preserve response body on api requests with XML response types.

taeold:
Today, apiv2 will try to parse XML responses as JSON with poor results, e.g.

```
⚠ functions: Upload Error: Unable to parse JSON: SyntaxError: Unexpected token < in JSON at position 0

Error: Unable to parse JSON: SyntaxError: Unexpected token < in JSON at position 0
```

Here, we explicitly add support for `xml` as a responseType and deal with it as a string. E.g.

```
⚠  functions: Upload Error: HTTP Error: 400, <?xml version='1.0' encoding='UTF-8'?><Error><Code>EntityTooLarge</Code><Message>Your proposed upload is larger than the maximum object size specified in your Policy Document.</Message><Details>Content-length exceeds upper bound on range</Details></Error>

Error: HTTP Error: 400, <?xml version='1.0' encoding='UTF-8'?><Error><Code>EntityTooLarge</Code><Message>Your proposed upload is larger than the maximum object size specified in your Policy Document.</Message><Details>Content-length exceeds upper bound on range</Details></Error>
```

Fixes https://github.com/firebase/firebase-tools/issues/4146


src/gcp/storage.ts
```diff
@@ -158,6 +158,7 @@ export async function upload(
     method: "PUT",
     path: url.pathname,
     queryParams: url.searchParams,
+    responseType: "xml",
```
bkendall, line 161: Forgive me a silly question: could we not just add `accept: application/json` to get JSON responses from this API? Or does it exclusively respond in XML?

It feels a little weird to add support for XML for a single place - not that it's _wrong_, but _everywhere else_ returns JSON, so it's an outlier...

CHANGELOG.md
```diff
@@ -4,3 +4,4 @@
 - Fixes broken functions:config:clone command (#4173).
 - Fixes issue where `auth:import` would fail when reading a JSON file. (#4157)
 - Fixes issue where custom claims added in Auth Emulator UI was not properly shown.
+- Improve handling api requests with XML responses (#4180).
```
bkendall, line 7:
```suggestion
- Improves handling of api requests with XML responses (#4180).
```

src/gcp/storage.ts
taeold: It's strictly an XML endpoint we are using since the URL returned by GCF when we call "generateUploadURL" works against it:

https://cloud.google.com/storage/docs/xml-api/overview

We _could_ do some work to move everything to JSON API

https://cloud.google.com/storage/docs/json_api/v1/buckets/patch

But that's gonna be a lot more work than this.

taeold: (Looks like XML is the encoding of choice over JSON in the storage API world)

taeold: whoops more correct answer - signed URL only supports XML API:

https://cloud.google.com/storage/docs/access-control/signed-urls

joehan (approved):

PR #4197, opened 2022-02-18
Set env var FUNCTION_SIGNATURE_TYPE=cloudevent when deploying gen2 functions

taeold:
By setting FUNCTION_SIGNATURE_TYPE=cloudevent, we prevent deployed gen2 functions from downcasting cloudevent messages.

### DIFF
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -31,9 +31,6 @@ function hasDotenv(opts: functionsEnv.UserEnvsOpts): boolean {
   return functionsEnv.hasUserEnvs(opts);
 }
 
-/**
- *
- */
 export async function prepare(
   context: args.Context,
   options: Options,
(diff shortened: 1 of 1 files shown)


src/deploy/functions/prepare.ts
```diff
@@ -138,6 +138,13 @@ export async function prepare(
   // Setup environment variables on each function.
   for (const endpoint of backend.allEndpoints(wantBackend)) {
     endpoint.environmentVariables = wantBackend.environmentVariables;
+    if (endpoint.platform === "gcfv2") {
```
mbleigh, line 141: nit: Should we also check whether it's actually an event-handling function, or just set it unilaterally in gen2? I'm probably fine either way.

mbleigh (approved):

src/deploy/functions/prepare.ts
inlined: We should only be doing it if `backend.isEventTriggered`, otherwise I believe we can force HTTPS functions into parsing as event functions.

PR #4206, opened 2022-02-22
Reorganize ext command code

elvisun:
Reorganize ext command code to move utility functions to the bottom of the file. This way the code flows vertically. More on this in Clean Code: https://screenshot.googleplex.com/5NykTruHkwiAYVq

### DIFF
--- a/src/commands/ext-install.ts
+++ b/src/commands/ext-install.ts
@@ -54,6 +54,104 @@ interface InstallExtensionOptions {
   force?: boolean;
 }
 
+/**
+ * Command for installing an extension
+ */
+export default new Command("ext:install [extensionName]")
+  .description(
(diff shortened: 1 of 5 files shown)


src/commands/ext-install.ts
```diff
@@ -54,6 +54,146 @@ interface InstallExtensionOptions {
   force?: boolean;
 }
 
+/**
```
joehan, line 57: Consider moving this above InstallExtensionOptions too, so that InstallExtensionOptions is next to installExtension (or just move InstallExtensionOptions to be an inline args type)

joehan (approved):
LGTM, with one extra thing we could clean up

src/commands/ext-install.ts
elvisun: Good call, done

PR #4208, opened 2022-02-22
Remove httpsTrigger from field mask to stop overwrite of securityLevel

### LINKED ISSUE #4171: HTTPS trigger securityLevel is overwritten on function re-deploy

**firebase-tools:** 10.2.0

**Platform:** macOS

### [REQUIRED] Test case

Using the [getting started with Firebase functions](https://firebase.google.com/docs/functions/get-started) article deploy and redeploy a function. The Require HTTPs flag is removed when the function is updated.

1. Create a GCP project in the Cloud Console
2. Follow steps 1-3 from the article
3. Edit `index.js` to include the `helloWorld` function
```
exports.helloWorld = functions.https.onRequest((request, response) => {
});
```
4. Deploy the function `firebase deploy --project PROJECT_ID --only functions:helloWorld`
5. Go back to the console, click edit. The Function has the `Require HTTPS` flag enabled
6. Execute the `deploy` command again
7. Go back to Cloud Console, edit the function 
8. The function no longer has the `Require HTTPS` flag

The function retains the `Require HTTPS` flag

The `Require HTTPS` flag is removed

### Additional information

@inlined enforced HTTPS in newly created functions in the [createV1Function](https://github.com/firebase/firebase-tools/blob/master/src/deploy/functions/release/fabricator.ts#L212-L214) which enables the flag (awesome)

However, when we re-deploy, the [updateV1Function](https://github.com/firebase/firebase-tools/blob/master/src/deploy/functions/release/fabricator.ts#L325) sends the `updateMask` which contains the `httpsTrigger` field but the trigger is an empty object. 

Either `httpsTrigger` should not be sent as part of the `updateMask` properties or enforce HTTPs in the `updateV1Function` as well.

image

colerogers:
Fixes #4171 by removing `httpsTrigger` from the field mask array on updates.

This is a patch bug fix that does not break currently deployed https functions. @inlined suggested changing `proto.fieldMasks` to ignore empty objects, but empty objects are needed for protobuf.Empty. When we introduce a breaking change, then we should either set the `securityLevel` in Fabricator's `updateV1Function` or directly in `functionFromEndpoint`.

### DIFF
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -257,8 +257,10 @@ export type Endpoint = TargetIds &
     // on GCFv2 always
     uri?: string;
     sourceUploadUrl?: string;
-    // GCFv1 can be http or https and
-    // GCFv2 is always https
+    // TODO(colerogers): yank this field and set securityLevel to SECURE_ALWAYS
+    // in functionFromEndpoint during a breaking change release.
+    // This is a temporary fix to address https://github.com/firebase/firebase-tools/issues/4171
+    // GCFv1 can be http or https and GCFv2 is always https
     securityLevel?: SecurityLevel;
   };
 
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -204,8 +204,6 @@ export interface ServiceConfiguration {
 
 export type FunctionsPlatform = "gcfv1" | "gcfv2";
 
-export type SecurityLevel = "SECURE_ALWAYS" | "SECURE_OPTIONAL";
-
 export type Triggered =
   | HttpsTriggered
   | CallableTriggered
@@ -261,7 +259,7 @@ export type Endpoint = TargetIds &
     // in functionFromEndpoint during a breaking change release.
     // This is a temporary fix to address https://github.com/firebase/firebase-tools/issues/4171
     // GCFv1 can be http or https and GCFv2 is always https
-    securityLevel?: SecurityLevel;
+    securityLevel?: gcf.SecurityLevel;
   };
 
 export interface RequiredAPI {
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -204,6 +204,8 @@ export interface ServiceConfiguration {
 
 export type FunctionsPlatform = "gcfv1" | "gcfv2";
 
+export type SecurityLevel = "SECURE_ALWAYS" | "SECURE_OPTIONAL";
+
 export type Triggered =
   | HttpsTriggered
   | CallableTriggered
@@ -255,6 +257,9 @@ export type Endpoint = TargetIds &
     // on GCFv2 always
     uri?: string;
     sourceUploadUrl?: string;
+    // GCFv1 can be http or https and
+    // GCFv2 is always https
+    securityLevel?: SecurityLevel;
   };
 
 export interface RequiredAPI {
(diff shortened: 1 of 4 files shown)


src/gcp/cloudfunctions.ts
```diff
@@ -356,11 +356,16 @@ export async function updateFunction(
   cloudFunction: Omit<CloudFunction, OutputOnlyFields>
 ): Promise<Operation> {
   const endpoint = `/${cloudFunction.name}`;
+  // TODO(colerogers): remove cloudFunctionWithoutHttpsTrigger and set securityLevel in functionFromEndpoint during a breaking change release
+  const cloudFunctionWithoutHttpsTrigger = { ...cloudFunction };
+  delete cloudFunctionWithoutHttpsTrigger.httpsTrigger;
```
taeold, line 361: nit* can you add a comment on why we are clearing this field (e.g. link to the issue)

taeold (approved):

src/gcp/cloudfunctions.ts
inlined, line 361: This can work for now, but it won't scale if we have additional fields added to httpsTrigger. Is `prepare.ts:inferDetailsFromExisting` a good candidate here?

inlined (commented):
LGTM with Daniel's request for a comment.

src/gcp/cloudfunctions.ts
colerogers: Changed the implementation to set `securityLevel` in `backend.Endpoint`. I removed this delete field code

colerogers: Latest commit we set securityLevel in `backend.Endpoint` and use `prepare.ts:inferDetailsFromExisting` to copy the current value.

src/deploy/functions/backend.ts
```diff
@@ -255,6 +257,9 @@ export type Endpoint = TargetIds &
     // on GCFv2 always
     uri?: string;
     sourceUploadUrl?: string;
+    // GCFv1 can be http or https and
+    // GCFv2 is always https
+    securityLevel?: SecurityLevel;
```
taeold, line 262: Hmm I honestly liked the old solution better. I'm a little concerned that `securityLevel` is only relevant to GCFv1 HTTP triggers yet we are proposing here that it's part of any Endpoint.

Should we at lesast annotate this as a temporary hack? IIUC, the way forward is to have a breaking change where we always set HTTP function to HTTPS-only.

colerogers: Would dumping it in `backend.Endpoint.httpsTrigger` fit a bit better? I'll also make a comment about this being temporary and link to the Issue.

Looking at the old solution, deleting the entire `httpsTrigger` object doesn't really match with the rest of the deployment process. This way we're at least utilizing the built-in flow, leveraging `inferDetailsFromExisting`, and actually creating the exact CloudFunction object that should be deployed.

taeold: One thing I liked about the old way is that it is egregious (and therefore obvious) that it's an hack! 

Regardless, if this is something we intend to clean up in the future, I don't think this is that harmful.

src/deploy/functions/backend.ts
```diff
@@ -204,6 +204,8 @@ export interface ServiceConfiguration {
 
 export type FunctionsPlatform = "gcfv1" | "gcfv2";
 
+export type SecurityLevel = "SECURE_ALWAYS" | "SECURE_OPTIONAL";
```
taeold, line 207: nit* you might want to use type definition define din cloudfunctions.ts instead.

taeold (approved):

PR #4214, opened 2022-02-23
Fix Storage Emulator rules resource evaluation

### LINKED ISSUE #3760: resource is null in storage rules for "delete" rule
**firebase-tools:** version 9.18.0

**Platform:** Windows

### Test case
Trying to delete a file in local storage emulator by reading metadata off of ```resource``` value in ```allow delete``` rule, but get null value error instead

#### Step 1: Add rule to storage.rules
```javascript
rules_version = '2';
service firebase.storage {
    match /b/{bucket}/o {
        match /users/{media} {
            allow read: if true;
            allow delete: if resource.metadata.public == "false";
            allow create, update: if true;
        }
    }
}
```

#### Step 3: Set custom metadata on file 1.webp
```javascript
await updateMetadata(ref(storage, 'users/1.webp'), {customMetadata: {public: "false"}});
```

#### Step 4: Try to delete file
```javascript
await deleteObject(ref(storage, 'users/1.webp'));
```

### Expected behavior

1. ```resource``` in storage rule is not null, instead contains metadata of resource that is to be deleted
2. Storage rule allows to successfully delete file

[Official Firebase doc](https://firebase.google.com/docs/storage/security/rules-conditions#resource_evaluation) about Resource Evaluation mentions that "These [metadata] properties can be inspected on read or write requests to ensure data integrity". ```delete``` rule is part of write request, hence it should work.

### Actual behavior

Console logs error
```
com.google.firebase.rules.runtime.common.EvaluationException: Error: C:\code\pr-next\storage.rules line [6], column [30]. Null value error.
```

Storage file is not deleted. ```resource``` is null.

I tested it only locally with emulator not in production.

tohhsinpei:
#### Issue
Fixes #3760.

The `resource` object was not being evaluated correctly in security rules for the Storage Emulator. An `EvaluationException` is thrown and the user receives a null value error. Moreover, the corresponding rule is not enforced.

#### Cause
The file metadata corresponding to the `resource` object was not being passed to the evaluator binary.

### Scenarios Tested
Added additional security rule evaluating `resource` and verified that existing integration tests now pass with the fix.

### DIFF
--- a/scripts/storage-emulator-integration/tests.ts
+++ b/scripts/storage-emulator-integration/tests.ts
@@ -1187,30 +1187,46 @@ describe("Storage emulator", () => {
         });
       });
 
-      it("#delete()", async () => {
-        const downloadUrl = await page.evaluate((filename) => {
-          return firebase.storage().ref(filename).getDownloadURL();
-        }, filename);
+      describe("deleteFile", () => {
+        it("should delete file", async () => {
+          await page.evaluate((filename) => {
+            return firebase.storage().ref(filename).delete();
+          }, filename);
 
-        expect(downloadUrl).to.be.not.null;
+          const error = await page.evaluate((filename) => {
+            return new Promise((resolve) => {
+              firebase
+                .storage()
+                .ref(filename)
+                .getDownloadURL()
+                .catch((err) => {
+                  resolve(err.message);
+                });
+            });
+          }, filename);
 
-        await page.evaluate((filename) => {
-          return firebase.storage().ref(filename).delete();
-        }, filename);
+          expect(error).to.contain("does not exist.");
+        });
 
-        const error = await page.evaluate((filename) => {
-          return new Promise((resolve) => {
-            firebase
-              .storage()
-              .ref(filename)
-              .getDownloadURL()
-              .catch((err) => {
-                resolve(err.message);
-              });
-          });
-        }, filename);
+        it("should not delete file when security rule on resource object disallows it", async () => {
+          await page.evaluate((filename) => {
+            firebase.storage().ref(filename).updateMetadata({ contentType: "text/plain" });
+          }, filename);
 
-        expect(error).to.contain("does not exist.");
+          const error = await page.evaluate((filename) => {
+            return new Promise((resolve) => {
+              firebase
+                .storage()
+                .ref(filename)
+                .delete()
+                .catch((err) => {
+                  resolve(err.message);
+                });
+            });
+          }, filename);
+
+          expect(error).to.contain("does not have permission to access");
+        });
       });
     });
 
--- a/src/emulator/storage/apis/firebase.ts
+++ b/src/emulator/storage/apis/firebase.ts
@@ -591,6 +591,13 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
   firebaseStorageAPI.delete("/b/:bucketId/o/:objectId", async (req, res) => {
     const decodedObjectId = decodeURIComponent(req.params.objectId);
     const operationPath = ["b", req.params.bucketId, "o", decodedObjectId].join("/");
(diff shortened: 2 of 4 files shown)


src/emulator/storage/apis/firebase.ts
```diff
@@ -591,16 +591,21 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
   firebaseStorageAPI.delete("/b/:bucketId/o/:objectId", async (req, res) => {
     const decodedObjectId = decodeURIComponent(req.params.objectId);
     const operationPath = ["b", req.params.bucketId, "o", decodedObjectId].join("/");
+    const md = storageLayer.getMetadata(req.params.bucketId, decodedObjectId);
+
+    const rulesFiles: { before?: RulesResourceMetadata } = {};
+
+    if (md) {
+      rulesFiles.before = md.asRulesResource();
```
tonyjhuang, line 599: Nice! Can we add a test?

tonyjhuang (changes requested):

src/emulator/storage/apis/firebase.ts
tohhsinpei: There is an [existing](https://github.com/firebase/firebase-tools/blob/5564ad1836ca91010e4b99ff78ef70f8f44bdc89/scripts/storage-emulator-integration/tests.ts#L1190) integration test for deletes. I thought I'd just add a condition on `resource` in the security rules for the test.

tonyjhuang: In that case can we add a test for the negative case? (ie security rules denies the delite)

tohhsinpei: Done.

scripts/storage-emulator-integration/tests.ts
```diff
@@ -1212,6 +1212,32 @@ describe("Storage emulator", () => {
 
         expect(error).to.contain("does not exist.");
       });
+
+      it("delete prevented by security rule", async () => {
+        const downloadUrl = await page.evaluate((filename) => {
+          return firebase.storage().ref(filename).getDownloadURL();
+        }, filename);
+
+        expect(downloadUrl).to.be.not.null;
```
tonyjhuang, line 1221: I know the rest of this file doesn't do a good job of this but please test one thing per test case. See [go/unit-test-practices#behavior-testing](https://goto.google.com/unit-test-practices#behavior-testing)

tonyjhuang: Same comment for `#delete()` test case above since we're already here

tonyjhuang (changes requested):
Mostly lgtm, just some test related nits

scripts/storage-emulator-integration/tests.ts
tohhsinpei: Ok yeah, I realised there's no need to test `getDownloadURL` here since there's already a test for it above. Removed from both delete tests.

scripts/storage-emulator-integration/tests.ts
```diff
@@ -1212,6 +1212,32 @@ describe("Storage emulator", () => {
 
         expect(error).to.contain("does not exist.");
       });
+
+      it("delete prevented by security rule", async () => {
```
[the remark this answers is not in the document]
tohhsinpei: The `#delete()` test suite is under `Admin SDK Endpoints`; added a new one under `Firebase Endpoints`.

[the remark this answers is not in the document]
tohhsinpei: Agree but would prefer to do in another PR since it's unrelated.

tonyjhuang (approved):
Nice fix, thanks!

tonyjhuang (approved):

yuchenshi (approved):

PR #4215, opened 2022-02-23
Add void in front of unawaited promises to remove lint errors

elvisun:
In an async context, it is normally expected that all promises are awaited, and that is the basis of the lint. However, there are times where one or more promises are intentionally not awaited. 

This in case, while 'track' returns a promise, the consumers of the commands that use track are unlikely to want to wait for that promise to complete. They only care about the command they are running.

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -469,7 +469,7 @@ async function loginRemotely(userHint?: string): Promise<UserCredentials> {
       codeVerifier
     );
 
-    track("login", "google_remote");
+    void track("login", "google_remote");
 
     return {
       user: jwt.decode(tokens.id_token!) as User,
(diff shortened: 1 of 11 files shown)


joehan (approved):
This change is OK since it isn't functionally changing anything. However, I want to double check that we are intentionally not awaiting these promises & that track will still work if we don't. @bkendall does this make sense to you, or should we await these instead?

bkendall:
IIRC so long as the process isn't forcefully exited (allowed to complete) the promise will eventually be resolved before it the process exits. With this change, is there an eslint rule that's currently a warning that can be made an error (or, just remove the downgrade to the warning)? That'd be super to include. Otherwise, this should be cool

elvisun:
@bkendall 56 more errors when I remove `no-floating-promises` 😅 Saving it for another week that's not perf week...

PR #4217, opened 2022-02-24
Extensions - Move manifest related util into manifest.ts

elvisun:
1. Move manifest related util into manifest.ts. 
2. Make some params in `readInstanceParam` optional so it can be reused more easily.


src/extensions/manifest.ts
```diff
@@ -92,3 +95,55 @@ async function writeEnvFiles(
...
+
+/**
+ * readParams gets the params for an extension instance from the `extensions` folder,
+ * checking for project specific env files, then falling back to generic env files.
+ * This checks the following locations & if a param is defined in multiple places, it prefers
+ * whichever is higher on this list:
+ *  - extensions/{instanceId}.env.{projectID}
+ *  - extensions/{instanceId}.env.{projectNumber}
+ *  - extensions/{instanceId}.env.{projectAlias}
+ *  - extensions/{instanceId}.env
```
joehan, line 107: Consider adding .local to this list with a note it only will be checked in checkLocal is true

joehan (approved):

src/extensions/manifest.ts
elvisun: sg, added to top of the list

PR #4219, opened 2022-02-24
Add --local flag to ext:configure

elvisun:
### Description

Add --local flag to ext:configure

### Scenarios Tested

```
lihes-macbookpro2:ext-emu lihes$ firebase ext:configure delete-user-data --local
i  extensions: Checking project IAM policy...
✔  extensions: Project IAM policy OK
⚠  extensions: The following param is immutable and won't be changed:


param: LOCATION, value: us-central1
To set a different value for this param, uninstall the extension, then install a new instance of this extension.
i  extensions: answer the questions below to configure your extension:

Cloud Firestore paths (Optional): Which paths in your Cloud Firestore instance contain user data? Leave empty if you don't use Cloud Firestore.
Enter the full paths, separated by commas. You can represent the User ID of the deleted user with {UID}.
For example, if you have the collections users and admins, and each collection has documents with User ID as document IDs, then you can enter users/{UID},admins/{UID}.
? Enter a value for Cloud Firestore paths: a

Cloud Firestore delete mode: (Only applicable if you use the Cloud Firestore paths parameter.) How do you want to delete Cloud Firestore documents? To also delete documents in subcollections, set this parameter to recursive.
? Which option do you want enabled for this parameter? Select an option with the arrow keys
, and use Enter to confirm your choice. You may only select one option. Recursive

Realtime Database instance (Optional): From which Realtime Database instance do you want to delete user data?
? Enter a value for Realtime Database instance: asdf

Realtime Database paths (Optional): Which paths in your Realtime Database instance contain user data? Leave empty if you don't use Realtime Database.
Enter the full paths, separated by commas. You can represent the User ID of the deleted user with {UID}.
For example: users/{UID},admins/{UID}.
? Enter a value for Realtime Database paths: zxcv

Cloud Storage paths (Optional): Where in Google Cloud Storage do you store user data? Leave empty if you don't use Cloud Storage.
Enter the full paths to files or directories in your Storage buckets, separated by commas. Use {UID} to represent the User ID of the deleted user, and use {DEFAULT} to represent your default Storage bucket.
Here's a series of examples. To delete all the files in your default bucket with the file naming scheme {UID}-pic.png, enter {DEFAULT}/{UID}-pic.png. To also delete all the files in another bucket called my-app-logs with the file naming scheme {UID}-logs.txt, enter {DEFAULT}/{UID}-pic.png,my-app-logs/{UID}-logs.txt. To also delete a User ID-labeled directory and all its files (like media/{UID}), enter {DEFAULT}/{UID}-pic.png,my-app-logs/{UID}-logs.txt,{DEFAULT}/media/{UID}.
? Enter a value for Cloud Storage paths: qwer

Adding Extensions to firebase.json...
✔  Wrote extensions/delete-user-data.env
lihes-macbookpro2:ext-emu lihes$ 
lihes-macbookpro2:ext-emu lihes$ 
```

```
lihes-macbookpro2:ext-emu lihes$ firebase ext:configure delete-user-data --local --non-interactive
i  extensions: Checking project IAM policy...
✔  extensions: Project IAM policy OK

Error: Command not supported in non-interactive mode, edit ./extensions/delete-user-data.env directly instead

lihes-macbookpro2:ext-emu lihes$ 

```


src/commands/ext-configure.ts
```diff
@@ -27,18 +30,75 @@ export default new Command("ext:configure <extensionInstanceId>")
...
+    if (options.local) {
+      if (options.nonInteractive) {
+        throw new FirebaseError(
+          `Command not supported in non-interactive mode, edit ./extensions/${instanceId}.env directly instead`
+        );
+      }
+
+      const config = manifest.loadConfig(options);
+      const targetRef = manifest.getInstanceRef(instanceId, config);
+      const extVer = await extensionsApi.getExtensionVersion(refs.toExtensionVersionRef(targetRef));
```
joehan, line 52: nit: prefer using full words (extensionVersion) for variable names

src/commands/ext-configure.ts
```diff
@@ -55,16 +115,13 @@ export default new Command("ext:configure <extensionInstanceId>")
       }
       const paramSpecWithNewDefaults =
         paramHelper.getParamsWithCurrentValuesAsDefaults(existingInstance);
-      const immutableParams = _.remove(paramSpecWithNewDefaults, (param) => {
-        return param.immutable || param.param === "LOCATION";
-        // TODO: Stop special casing "LOCATION" once all official extensions make it immutable
-      });
+      const immutableParams = _.remove(paramSpecWithNewDefaults, (param) => param.immutable);
```
joehan, line 118: Consider using partition from functional.ts here instead (https://github.com/firebase/firebase-tools/blob/master/src/functional.ts#L99)

src/test/extensions/manifest.spec.ts
```diff
@@ -34,6 +36,24 @@ describe("manifest", () => {
     });
   });
 
+  describe(`${manifest.getInstanceRef}`, () => {
```
joehan, line 39: Won't this print out the function object? I think this probably should be describe("getInstanceRef"... instead

joehan (changes requested):

src/test/extensions/manifest.spec.ts
elvisun: oh added `.name` to the end now it works. I like this pattern because it gives better IDE support. (renaming, clicking through, etc..) If you think we should follow the rest of the codebase I'm happy to change it too

src/commands/ext-configure.ts
elvisun: good suggestion, wasn't a fan of this lodash function too.. 

https://media.makeameme.org/created/reach-for-immutable.jpg

PR #4220, opened 2022-02-24
Adding extensions to emulatorRegistry

joehan:
### Description
Adds extensions emulator to emulator registry so that it is returned from http://localhost:{hubPort}/emulators. Since the Extensions emulator is part of the functions emulator under the hood, this endpoint will return the same info for both.

### Scenarios Tested
Verified that extensions is returned from /emulators when the Extensions emulator is running, and that the emulators cleanly start up and shut down.
Screen Shot 2022-02-24 at 10 22 20 AM

### DIFF
--- a/src/emulator/hub.ts
+++ b/src/emulator/hub.ts
@@ -81,9 +81,12 @@ export class EmulatorHub implements EmulatorInstance {
 
     this.hub.get(EmulatorHub.PATH_EMULATORS, (req, res) => {
       const body: GetEmulatorsResponse = {};
-      EmulatorRegistry.listRunning().forEach((name) => {
-        body[name] = EmulatorRegistry.get(name)!.getInfo();
-      });
+      for (const emulator of EmulatorRegistry.listRunning()) {
+        const info = EmulatorRegistry.getInfo(emulator);
+        if (info) {
+          body[emulator] = info;
+        }
+      }
       res.json(body);
     });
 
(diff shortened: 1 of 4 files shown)


src/emulator/registry.ts
```diff
@@ -92,6 +98,11 @@ export class EmulatorRegistry {
   }
 
   static isRunning(emulator: Emulators): boolean {
+    if (emulator === Emulators.EXTENSIONS) {
+      return (
+        this.extensionsEmulatorRegistered && this.INSTANCES.get(Emulators.FUNCTIONS) !== undefined
```
elvisun, line 103: wdyt of this to avoid relying on implementation detail? 
```
        this.extensionsEmulatorRegistered && this.isRunning(Emulators.FUNCTIONS)
```

elvisun (approved):

src/emulator/hub.ts
```diff
@@ -81,9 +81,12 @@ export class EmulatorHub implements EmulatorInstance {
 
     this.hub.get(EmulatorHub.PATH_EMULATORS, (req, res) => {
       const body: GetEmulatorsResponse = {};
-      EmulatorRegistry.listRunning().forEach((name) => {
-        body[name] = EmulatorRegistry.get(name)!.getInfo();
-      });
+      for (const emulator of EmulatorRegistry.listRunning()) {
+        const info = EmulatorRegistry.getInfo(emulator);
+        if (info) {
```
tjlav5, line 86: If this was to be undefined, should we be throwing an error? Perhaps we can wrap this with an assertion instead

src/emulator/registry.ts
joehan: Ahh, that's probably cleaner. I had automerge on, so I'll make another PR to clean this up

PR #4226, opened 2022-02-25
Add warning when emulating extensions that use unemulated APIs

joehan:
### Description
Adds a warning when emulating extensions that use unemulated APIs

### Scenarios Tested
Screen Shot 2022-02-28 at 3 19 46 PM


src/emulator/extensions/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const EMULATED_APIS = [
+  "storage-component.googleapis.com",
+  "firestore.googleapis.com",
+  "pubsub.googleapis.com",
+  "identitytoolkit.googleapis.com",
+  // TODO: Is there a RTDB API we need to add here? I couldn't find one.
+];
+
+type UnemulatedAPI = {
```
elvisun, line 12: how about `emulatableAPI` or something similar? The type doesn't have to be specific to unemulated APIs

src/emulator/extensions/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+  apiName: string;
+  instanceIds: string[];
+  enabled: boolean;
+};
+/**
+ * getNonEmulatedAPIs checks a list of InstanceSpecs for APIs that are not emulated.
```
elvisun, line 18: getUnemulatedAPIs ?

src/emulator/extensions/utils.ts
```diff
@@ -0,0 +1,44 @@
+import * as planner from "../../deploy/extensions/planner";
```
elvisun, line 1: let's give this file a more specific name so it doesn't become a dump yard in the future..

src/emulator/extensionsEmulator.ts
```diff
@@ -169,4 +174,25 @@ export class ExtensionsEmulator {
...
+
+  private async checkAndWarnAPIs(instances: planner.InstanceSpec[]): Promise<void> {
+    const apisToWarn = await getUnemulatedAPIs(this.args.projectId, instances);
+    if (apisToWarn.length) {
+      const table = new Table({
+        head: ["API Name", "Instances using this API", `Enabled on ${this.args.projectId}`],
+        style: { head: ["yellow"] },
+      });
+      for (const apiToWarn of apisToWarn) {
+        table.push([apiToWarn.apiName, apiToWarn.instanceIds, apiToWarn.enabled]);
```
elvisun, line 186: have we considered giving users a way out on the disabled APIs?

If they have an instance that uses translate API, but they don't have that enabled, they are basically stuck. Maybe we can ask if they want to enable the apis or log a url to pantheon?

elvisun (changes requested):

src/emulator/extensionsEmulator.ts
joehan: This is a good point. I like the idea of including a link to enable the API in console - I think we want to be _very_ careful that users know that they are taking an action that affects production, & offering a prompt to enable them might not make that abundantly clear.

elvisun: sgtm, that's an excellent point, I'd be one of those people that spam enter through everything lol

src/emulator/extensions/utils.ts
joehan: Swapped this to the more generic APIInfo

src/emulator/extensions/utils.ts
joehan: Renamed this to validation.ts, with the intention of using it as a place to separate out & unit test various functions we use to validate the users extensions and project before emulation

src/emulator/extensionsEmulator.ts
```diff
@@ -169,4 +184,39 @@ export class ExtensionsEmulator {
...
+
+  private async checkAndWarnAPIs(instances: planner.InstanceSpec[]): Promise<void> {
+    const apisToWarn = await getUnemulatedAPIs(this.args.projectId, instances);
+    if (apisToWarn.length) {
+      const table = new Table({
+        head: [
+          "API Name",
+          "Instances using this API",
+          `Enabled on ${this.args.projectId}`,
+          `Enable this API`,
```
elvisun, line 196: generally lgtm, something to think about with UX, there could be benefit to put this outside of the table - it makes it clear it is an optional step, maybe the user doesn't want to enable this API until they do more testing, see the logs, or they simply don't want to incur any cost etc

src/test/emulators/extensions/validation.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+      ]);
+
+      expect(result).to.deep.equal([
+        {
+          apiName: testAPI,
+          instanceIds: [instanceIdWithUnemulatedAPI],
+          enabled: true,
+        },
+      ]);
+      // expect(result[0].apiName).to.equal(testAPI);
```
elvisun, line 62: still need this?

src/test/emulators/extensions/validation.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+      state: "PUBLISHED",
+      sourceDownloadUri: "test.com",
+      hash: "abc123",
+      spec: {
+        name: "test",
+        version: "0.1.0",
+        sourceUrl: "test.com",
+        resources: [],
+        params: [],
+        apis: [{ apiName, reason: "because" }],
```
elvisun, line 24: 😂

src/test/emulators/extensions/validation.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+      const result = await utils.getUnemulatedAPIs(testProjectId, [
+        getTestInstanceSpecWithAPI(instanceIdWithEmulatedAPI, "firestore.googleapis.com"),
+        getTestInstanceSpecWithAPI(instanceIdWithUnemulatedAPI, testAPI),
+      ]);
+
+      expect(result).to.deep.equal([
+        {
+          apiName: testAPI,
+          instanceIds: [instanceIdWithUnemulatedAPI],
```
elvisun, line 58: consider adding another instance using `test.googleapis.com` for more coverage

elvisun (approved):

Issue #4189, opened 2022-02-17
Local Function Emulator Triggering wrong function


markdagraca:
Trying to run my react typescript project using Local Emulator and it is triggering the wrong function when called from the front end. Code is working perfectly in the production and test environment but not in Local

markdagraca:
Downgraded to 10.1.5 and it appears to be working properly but I'm now missing env variables features found in 10.2

Albertbol:
exactly the same issue for the node js environment, all latest libraries
```
    "engines": {
        "node": "16"
    },
    "main": "dist/index.js",
    "dependencies": {
        "firebase": "^9.6.7",
        "firebase-admin": "^10.0.2",
        "firebase-functions": "^3.18.1"
    },
```

roobd:
I have the same issue that my function triggers are not working anymore after the upgrade to 10.2.1.
There is only one onCreate function triggered all the time, even a creation on a completely different document path is triggering this one function.

A downgrade to 10.2.0. does not fix the issue.

I use firebase-functions at version 3.18.1

A downgrade to 10.1.5 is fixing the issue and all triggers are working as expected again.

taeold:
We are taking a look at the issue. Sorry for the incovenience!

ggirotto:
We were facing the same issue but it stopped happening yesterday.

gustavrannestig:
I have the same issue. From what I can see the patterin is that it seems to be stuck on the first called function.

So when I restart the emulators, I can choose what function to use for that "session" by firing that first. But then I can't switch and call another function without restarting.

So I'm guessing some sort of cache bug? :)

haroldadmin:
Can confirm, facing the same issue that @gustavrannestig described. It keeps triggering the first triggered function for any subsequent API calls.

taeold:
Been investigating the issue, and @gustavrannestig's description helps a ton. Will post an update here once I get a sense of the bug - apologies to everyone for the inconvenience.

Zebreus:
I also have the same issue. The first triggered function is called for any subsequent API calls.
The first bad firebase-tools version is 10.2.0.
The issue only happens, when the functions emulator is in debug mode.

taeold:
@Zebreus Thanks for a very helpful pointer.

Is everyone reporting the issue also starting the emulator with `--inspect-functions` flag?

markdagraca:
Yes im using the following command to launch the tool firebase
[...]

Issue #4166, opened 2022-02-15
Debuggers (Webstorm/VSCode) not attaching to emulators on versions >10.2.0


cybex-dev:
I'm filling this bug report more as an issue rather than a bug. The reason for it being a bug report is I've attempted debugging on multiple OS's on my PC (Windows, Ubuntu) and a Mac, none of which allow the debugger to attach. I thus resolved to start from scratch and work my way up to finding the problem. This isn't working out as even the simplest firebase project can't be debugged. This is after following these steps to [remove Node from my Windows install ](https://stackoverflow.com/questions/20711240/how-to-completely-remove-node-js-from-windows) to the letter, but no success.

### Environment info
**node:** v16.13.2
**npm:** 8.1.2

**IDE:** Webstorm 2021.1.1
**alt IDE:** VSCode 1.64.2 (latest atm)

**firebase-tools:** 10.2.0

**Platform:** Windows 10 x64

### Test case

Source can be found [here](https://github.com/cybex-dev/firebase-minimal). This is (at the time of writing) the simplest firebase project that can be created (see steps to reproduce).

There are 4 core files:
```
/ functions
    / index.js
    / package.json
.firebaserc
.firebase.json
```

**File: package.json**
```json
"engines": {
  "node": "16"
},
"dependencies": {
  "firebase-admin": "^9.8.0",
  "firebase-functions": "^3.14.1"
},
"devDependencies": {
  "firebase-functions-test": "^0.2.0"
},
```

the important **firebase.json**:
```json
"functions": {
  "source": "functions"
},
"emulators": {
  "functions": {
    "port": 5001
  },
  "ui": {
    "enabled": true
  }
}
```

### Steps to reproduce

1. Create folder i.e. test-project-emulator
2. Create **new** firebase project (assume logged into firebase), specify javascript & functions feature (minimalist example) & install dependencies 
i.e. `firebase init`

3. Install functions emulator, leave all as default and install emulator (see console output*)
i.e. `firebase init emulators`
4. Launch functions emulator
i.e. `firebase emulators:start`
5. launch IDE and attempt to attach

**IDE Debugging:**
1. WebStorm (use Attach to Node.JS/Chrome with localhost:9229 as described [here](https://stackoverflow.com/a/61975914/4628115)) 
2. VSCode with launch script:
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "node",
            "request": "attach",
            "name": "Debug",
            "port": 9229
        }
    ]
[...]

PR #4232, opened 2022-03-01
Dynamically invoke correct trigger when running functions emulator with --inspect-functions flag

taeold:
In https://github.com/firebase/firebase-tools/pull/4149, we made large refactor of the Functions Runtime. Namely:

1) We no longer relied on Functions Runtime to parse triggers. Instead, we use [`RuntimeDelegate`](https://github.com/firebase/firebase-tools/blob/2a56d9520241ab9897a59aac22c9fd016251a7cd/src/deploy/functions/runtimes/index.ts#L114), the same procedure used to parse trigger in `firebase deploy` command.

2) Each process running the Functions Runtime was bound to a specific trigger.

Unfortunately, this change broke support for `--inspect-function` where users can attach Node Debugger to step through their function when triggered.

The debugging experience supported w/ `--inspect-function` relies on the fact that a single process executes all function triggers. This is drastically different to how functions run in Production environment, but it is a very useful feature that makes it easy to step through all functions in a single debug session.

I wasn't aware of the debugging capabilities when making the refactor. Unfortunately, the debug feature cuts across a strong assumption I've made for the future of Functions Emulator. This is a rather hasty rollback - I'm going to have to think more deeply about how we want the evolve this feature.

Fixes https://github.com/firebase/firebase-tools/issues/4189, https://github.com/firebase/firebase-tools/issues/4166

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,3 +5,4 @@
 - Fixes Storage Emulator rules resource evaluation (#4214).
 - Fixes bug where securityLevel is overwritten on https function re-deploys (#4208).
 - Fixes bug where functions emulator ignored functions.runtime option in firebase.json (#4207).
+- Fixes bug where functions emulator triggered wrong functions when when started w/ --inspect-functions flag (#4232).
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -5,3 +5,4 @@
 - Fixes Storage Emulator rules resource evaluation (#4214).
 - Fixes bug where securityLevel is overwritten on https function re-deploys (#4208).
 - Fixes bug where functions emulator ignored functions.runtime option in firebase.json (#4207).
+- Fixes bug where functions emulator triggered wrong functions when when started w/ --inspect-functions flag (#4232).
```
joehan, line 8: Typo: "wrong functions when when" should just be "wrong functions when"

Also prefer "with" instead of "w/"

PR #4240, opened 2022-03-02
Migrate endpoint annotation eventTrigger.eventFilters to list of filters

taeold:
Previously, `eventFilter` attribute was an object:

```
eventTrigger: {
  eventType: "an.event",
  eventFilter: {
    resource: "my-storage-bucket",
    appId: "12345",
  }
}
```

We now prefer eventFilter as a list:

```
eventTrigger: {
  eventType: "an.event",
  eventFilter: [
  {
    attribute: "resource",
    value: "my-storage-bucket",
  }.
  {
    attribute: "appId",
    value: "12345",
  }.
  ]
}
```

Most of the code here are simply makes necessary changes to operate on a list instead of an object.

One new behavior worth noting:

Manifest discovery now construct full topic path (e.g. `my-topic` -> `projects/my-project/topics/topic`.

(Surprisingly, we don't need to do the same for storage buckets since bucket names are globally unique)


src/deploy/functions/services/storage.ts
```diff
@@ -38,15 +38,19 @@ export async function obtainStorageBindings(
...
+  const { eventTrigger } = endpoint;
   if (!eventTrigger.region) {
     logger.debug("Looking up bucket region for the storage event trigger");
-    try {
-      const bucket: { location: string } = await storage.getBucket(
-        eventTrigger.eventFilters.bucket!
+    const bucketFilter = backend.findEventFilter(endpoint, "bucket");
+    if (!bucketFilter) {
+      throw new FirebaseError(
+        "Storage event trigger unexpectedly missing event filter with bucket attribute."
```
colerogers, line 49: Nice! More specific errors

src/gcp/cloudfunctionsv2.ts
```diff
@@ -18,6 +18,14 @@ const client = new Client({
 });
 
 export const PUBSUB_PUBLISH_EVENT = "google.cloud.pubsub.topic.v1.messagePublished";
+export const STORAGE_EVENTS = [
```
colerogers, line 21: Should we make a distinction that these are only v2 storage events?

src/gcp/cloudfunctionsv2.ts
```diff
@@ -18,6 +18,14 @@ const client = new Client({
...
+  "google.cloud.storage.object.v1.finalized",
+  "google.cloud.storage.object.v1.archived",
+  "google.cloud.storage.object.v1.deleted",
+  "google.cloud.storage.object.v1.metadataUpdated",
+];
+
+export const PUBSUB_V2_EVENT = "google.cloud.pubsub.topic.v1.messagePublished";
```
colerogers, line 28: In `parseTriggers.ts` we're only importing `PUBSUB_PUBLISH_EVENT` from L20, should we remove this line?

src/gcp/cloudfunctionsv2.ts
colerogers: Also, I'm wondering if we should export these from `services/index.ts` instead. We already have the event/service mapping there, so I can see us also defining the constants in that file.

src/deploy/functions/services/storage.ts
```diff
@@ -38,15 +38,19 @@ export async function obtainStorageBindings(
  * @param eventTrigger the endpoints event trigger
  */
 export async function ensureStorageTriggerRegion(
-  endpoint: backend.Endpoint,
-  eventTrigger: backend.EventTrigger
+  endpoint: backend.Endpoint & backend.EventTriggered
```
colerogers, line 41: Oh nice, way cleaner 👌

src/gcp/cloudfunctionsv2.ts
taeold: Ah good catch!

src/gcp/cloudfunctionsv2.ts
taeold: Since the events are defined in `src/gcp/cloudfunctionsv2.ts`, I felt that annotating w/ `_V2_` redundant.

In hindsight, I like your original structure where we had explicit module for keeping event types. Let me see what it looks like if I follow that model.

src/gcp/cloudfunctionsv2.ts
```diff
@@ -436,12 +435,27 @@ export function functionFromEndpoint(endpoint: backend.Endpoint, source: Storage
...
+
+      for (const filter of endpoint.eventTrigger.eventFilters) {
+        if (filter.attribute === "topic") {
+          continue;
+        }
+        if (!gcfFunction.eventTrigger.eventFilters) {
+          gcfFunction.eventTrigger.eventFilters = [];
+        }
+        gcfFunction.eventTrigger.eventFilters.push(filter);
+      }
```
colerogers, line 456: I think we could just inline this whole loop to 
```
gcfFunction.eventTrigger.eventFilters = endpoint.eventTrigger.eventFilters.filter((ef) => ef.attribute !== "topic"); 
```

colerogers (commented):
Few more comments/q's

src/deploy/functions/runtimes/node/parseTriggers.ts
```diff
@@ -168,6 +168,9 @@ export function mergeRequiredAPIs(backend: backend.Backend) {
   backend.requiredAPIs = merged;
 }
 
+/**
+ *
+ */
```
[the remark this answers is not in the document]
taeold: Woos EsLIint --fix got the best of me. Removed.

src/functions/events/v2.ts
```diff
@@ -0,0 +1,8 @@
+export const PUBSUB_PUBLISH_EVENT = "google.cloud.pubsub.topic.v1.messagePublished" as const;
+export const STORAGE_EVENTS = [
+  "google.cloud.storage.object.v1.finalized",
+  "google.cloud.storage.object.v1.archived",
+  "google.cloud.storage.object.v1.deleted",
+  "google.cloud.storage.object.v1.metadataUpdated",
+] as const;
+export type Event = typeof PUBSUB_PUBLISH_EVENT | typeof STORAGE_EVENTS[number];
```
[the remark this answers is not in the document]
taeold: Put spaces between constants.

Not sure if you also preferred something like

```js
export type Event = typeof PUBSUB_PUBLISH_EVENT 
  | typeof STORAGE_EVENTS[number];
```

Sadly, ESLint doesn't like this format.

src/gcp/cloudfunctionsv2.ts
```diff
@@ -436,12 +435,27 @@ export function functionFromEndpoint(endpoint: backend.Endpoint, source: Storage
...
+          "Invalid pubsub endpoint. Expected eventFilter with 'topic' attribute but found none."
+        );
       }
+      gcfFunction.eventTrigger.pubsubTopic = pubsubFilter.value;
+
+      for (const filter of endpoint.eventTrigger.eventFilters) {
+        if (filter.attribute === "topic") {
+          continue;
+        }
+        if (!gcfFunction.eventTrigger.eventFilters) {
```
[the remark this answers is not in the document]
taeold: whoots thanks for the catch. I meant to remove initialization that's done in L438. That way, I'd have undefined `eventTrigger.eventFilters` so we don't override eventFilters updated out-of-band.

src/gcp/cloudfunctionsv2.ts
taeold: See comment above. I did start off with this but then realized that

1. `eventTrigger.eventFilters = []`
2. `eventTrigger.eventFilters = undefined`

Are different things when it comes to deployment - the former would override whatever eventFilter values that changed out-of-band while the later preserves it.

colerogers: Ahhh that makes sense, thanks

colerogers (approved):

src/deploy/functions/runtimes/node/parseTriggers.ts
```diff
@@ -211,19 +211,40 @@ export function addResourcesToBackend(
...
+            {
+              attribute: "topic",
+              value: annotation.eventTrigger!.resource,
+            },
+          ];
+        }
+
+        if (
+          v2events.STORAGE_EVENTS.find(
+            (event) => event === (annotation.eventTrigger?.eventType || "")
```
inlined, line 238: nit: I don't think you need the `|| ""` because you shouldn't crash comparing a string to `undefined`

src/deploy/functions/services/index.ts
```diff
@@ -48,12 +49,12 @@ export const EVENT_SERVICE_MAPPING: Record<string, any> = {
...
- * @returns a Service object that corresponds to the event type of the endpoint or noop
+ * @return a Service object that corresponds to the event type of the endpoint or noop
  */
 export function serviceForEndpoint(endpoint: backend.Endpoint): Service {
   if (!backend.isEventTriggered(endpoint)) {
     return NoOpService;
   }
 
-  return EVENT_SERVICE_MAPPING[endpoint.eventTrigger.eventType] || NoOpService;
+  return EVENT_SERVICE_MAPPING[endpoint.eventTrigger.eventType as v2events.Event] || NoOpService;
```
inlined, line 59: It's no less safe than previous iterations, but I kinda feel like if we're going to cast to a `v2events.Event` we should have a function that checks whether a string is a valid event type first.

inlined (commented):
Given how much of a PITA these changes have been and how much more annoying it is to do key lookup in an array, I wonder if we should have adopted something like the Mongo query syntax:

```
type EventFilterOperators = "match-path-pattern";
type EventFilterValueMatch = string | { [EventFilterOperators]: string }
type EventFilters = Record<string, EventFilterValueMatch>
```

Then all existing code would have worked and the RTDB would just have something like:

```
{
  ref: {
    "match-path-pattern": "projects/_/databases/myfirebaseio/refs/users/{uid}"
  }
}
```

src/deploy/functions/services/index.ts
taeold: It isn't safer here, but it does important typos when we add new entries to the `EVENT_SERVICE_MAPPING`.

Agreed that having that check would've felt more idiomatic in hindsight.

PR #4242, opened 2022-03-02
Track ext:install with url source usage

elvisun:
Track ext:install with url source usage

### DIFF
--- a/src/commands/ext-install.ts
+++ b/src/commands/ext-install.ts
@@ -31,6 +31,7 @@ import {
   promptForValidInstanceId,
   isLocalOrURLPath,
   diagnoseAndFixProject,
+  isUrlPath,
 } from "../extensions/extensionsHelper";
 import { update } from "../extensions/updateHelper";
 import { getRandomString } from "../extensions/utils";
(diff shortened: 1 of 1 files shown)


src/commands/ext-install.ts
```diff
@@ -31,6 +31,7 @@ import {
   promptForValidInstanceId,
   isLocalOrURLPath,
   diagnoseAndFixProject,
+  isUrlPath,
```
joehan, line 34: While you're here, get rid of the unused lodash dependency?

src/commands/ext-install.ts
```diff
@@ -83,6 +84,12 @@ export default new Command("ext:install [extensionName]")
     }
     let source;
     let extVersion;
+
+    // TODO(b/220900194): Track url install usage to determine if we can remove this feature.
+    if (isUrlPath(extensionName)) {
+      track("Extension Install", "Install by url path", options.interactive ? 1 : 0);
```
joehan, line 90: void track()?

src/commands/ext-install.ts
```diff
@@ -83,6 +84,12 @@ export default new Command("ext:install [extensionName]")
     }
     let source;
     let extVersion;
+
+    // TODO(b/220900194): Track url install usage to determine if we can remove this feature.
```
joehan, line 88: Isn't this PR completing this TODO?

joehan (approved):
LGTM with trivial fixes, so I disabled automerge

src/commands/ext-install.ts
elvisun: ah good catch, I made changes for this file in the launch branch, good idea to keep this clean

src/commands/ext-install.ts
elvisun: Updated this to: Remove tracking of url install once we remove this feature.

PR #4248, opened 2022-03-03
Refactor persistence out of StorageLayer

tonyjhuang:
### Description

Pulls `Persistence` into its own file and moves the constructor up into the `StorageEmulator`. This is to prepare for the eventual addition of an `UploadService` which will need a reference to a`Persistence` as well.


tohhsinpei (commented):
Looks reasonable, left a question

tohhsinpei (approved):

Issue #3716, opened 2021-08-24
Setting memory to more than 1 GiB still only allow 1 GiB to be used


ibobo:
### Related issues

None.

### [REQUIRED] Version info

**node:** v12.22.1

**firebase-functions:** 3.15.4

**firebase-tools:** 9.16.5

**firebase-admin:** 9.11.1

### [REQUIRED] Test case

```js
export const test = functions
  .runWith({
    memory: '8GB',
  })
  .https.onRequest(async (req, res) => {
    console.log(process.env);
  });
```

### [REQUIRED] Steps to reproduce

Just try to deploy the above mentioned function

### [REQUIRED] Expected behavior

The function gets deployed with the indicated memory and it can fully use the 8 GiB memory.

### [REQUIRED] Actual behavior

The function gets deployed with the indicated memory, but, it won't be able to use it all, since the runtime is configured to only use 1 or 2 GiB.

As stated in the Google Cloud documentation [here](https://cloud.google.com/functions/docs/concepts/nodejs-runtime#environment_variables), you need to specify the `NODE_OPTIONS` environment variable to have a `max_old_space_size` value (8192 for 8 GiB), but this is not possible with current firebase-functions deployment process.

I suggest that when a function is required to have more than 1 GiB memory size the given `NODE_OPTIONS` environment variable is set automatically. At this stage, to have it set, I need to go to Google Cloud Console, edit the function and manually add the environment variable (or at least check it was retained) after every deploy.

### Were you able to successfully deploy your functions?

Yes.

ikusuki:
So right now we can add the `NODE_OPTIONS` runtime env var with `--max_old_space_size=2048` (for 2GB for example) but it will be reset on every deploy, right?

any news on this?

taeold:
@ikusuki Good news - Google Cloud Functions team recently made a patch to the Node.js runtime that will automaticlly set the appropriate `--max_old_space_size` based on the memory you selected to deploy the function. e.g.

```js
const functions = require("firebase-functions");
const v8 = require('v8');

exports.heap = functions.runWith({ memory: "8GB" }).https.onRequest((request, response) => {
  functions.logger.log("memory", v8.getHeapStatistics());
  response.send(v8.getHeapStatistics());
});
```

returns for me

```
{
 "total_heap_size":27226112,
 "total_heap_size_executable":573440,
 "total_physical_size":17225552,
 "total_available_size":8539261544, /* LOOK - 8GB */
[...]

PR #4253, opened 2022-03-04
Split function deployments by region AND memory

taeold:
Today, Function deployments are made in batches of functions grouped by region. Combined w/ source token to re-used built containers for deployment, we get meaningful decrease in deploy time and in Cloud Build resource usage.

Unfortunately, this setup has a peculiar bug; Google Cloud Functions bakes in the flag to expand heap size for appropriate for the memory configuration of a function (`--max_old_space_size`) on the container itself. That means that if you batch two functions with differing memory configuration (e.g. 256MB vs 4 GB), it's guaranteed that one function will have wrongly configured `--max_old_space_size` flag value (Follow https://github.com/firebase/firebase-tools/issues/3716#issuecomment-1012573174 for how we discovered this issue).

This PR proposes batching function by region AND `availalbeMemoryMB` to fix this bug. We do this by refactoring `calculateRegionalChanges` to generate a collection of `Changeset` (previously known as `RegionalChanges`) that sub-divides functions in a region by their memory. In fact, we generalize the approach by allowing arbitrary subdivision by `keyFn` (e.g. `keyFn: (endpoint) => endpoint.availableMemoryMb`) as I anticipate that we will revisit this section of the code once I start working on "codebases".

Fixes https://github.com/firebase/firebase-tools/issues/3716

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Fixes bug where function's memory configurations weren't preserved in batched function deploys (#4253).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,1 @@
-- Fixes bug where function's memory configurations weren't preserved in batched function deploys (#4253).
+- Fixes bug where functions' memory configurations weren't preserved in batched function deploys (#4253).
--- a/src/deploy/functions/release/planner.ts
+++ b/src/deploy/functions/release/planner.ts
@@ -24,8 +24,8 @@ export interface Options {
   deleteAll?: boolean;
 }
 
-/** Calculate the changesets needed for a given region by grouping endpoints with keyFn. */
-export function calculateRegionalChanges(
+/** Calculate the changesets of given endpoints by grouping endpoints with keyFn. */
+export function calculateChangesets(
   want: Record<string, backend.Endpoint>,
   have: Record<string, backend.Endpoint>,
   keyFn: (e: backend.Endpoint) => string,
@@ -112,7 +112,7 @@ export function createDeploymentPlan(
 
   const regions = new Set([...Object.keys(want.endpoints), ...Object.keys(have.endpoints)]);
   for (const region of regions) {
-    const changesets = calculateRegionalChanges(
+    const changesets = calculateChangesets(
       want.endpoints[region] || {},
       have.endpoints[region] || {},
       (e) => `${e.region}-${e.availableMemoryMb || "default"}`,
--- a/src/deploy/functions/release/planner.ts
+++ b/src/deploy/functions/release/planner.ts
@@ -11,13 +11,13 @@ export interface EndpointUpdate {
   deleteAndRecreate?: backend.Endpoint;
 }
 
-export interface RegionalChanges {
+export interface Changeset {
   endpointsToCreate: backend.Endpoint[];
   endpointsToUpdate: EndpointUpdate[];
   endpointsToDelete: backend.Endpoint[];
 }
 
-export type DeploymentPlan = Record<string, RegionalChanges>;
+export type DeploymentPlan = Record<string, Changeset>;
 
 export interface Options {
   filters?: string[][];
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- Fixes bug where function's memory configurations weren't preserved in batched function deploys (#4253).
```
joehan, line 1: nit: This should be `functions' memory configurations`

src/deploy/functions/release/planner.ts
```diff
@@ -10,39 +10,63 @@ export interface EndpointUpdate {
...
   // If set to false, will delete only functions that are managed by firebase
   deleteAll?: boolean;
 }
 
-/** Calculate the changes needed for a given region. */
+/** Calculate the changesets needed for a given region by grouping endpoints with keyFn. */
 export function calculateRegionalChanges(
   want: Record<string, backend.Endpoint>,
   have: Record<string, backend.Endpoint>,
+  keyFn: (e: backend.Endpoint) => string,
```
joehan, line 31: Very clever bit of functional programming here 😄

src/deploy/functions/release/planner.ts
```diff
@@ -10,39 +10,63 @@ export interface EndpointUpdate {
...
+    Object.keys(want)
+      .filter((id) => !have[id])
+      .map((id) => want[id]),
+    keyFn
+  );
+
+  const toDelete = utils.groupBy(
+    Object.keys(have)
+      .filter((id) => !want[id])
+      .filter((id) => options.deleteAll || isFirebaseManaged(have[id].labels || {}))
```
joehan, line 44: This doesn't seem like new behavior, but it looks like options.deleteAll deletes every function on the project, not just Firebase managed functions. Is that intentional?

src/deploy/functions/release/planner.ts
```diff
@@ -10,39 +10,63 @@ export interface EndpointUpdate {
...
 
 export interface Options {
   filters?: string[][];
   // If set to false, will delete only functions that are managed by firebase
   deleteAll?: boolean;
 }
 
-/** Calculate the changes needed for a given region. */
+/** Calculate the changesets needed for a given region by grouping endpoints with keyFn. */
 export function calculateRegionalChanges(
```
joehan, line 28: We ought to rename this to something like "calculateChangesets" now that it is not specifically for regions. There isn't really any specific logic for grouping by region baked into this any more

joehan (approved):
Some nits but LGTM overall

src/deploy/functions/release/planner.ts
taeold: I think this option is only used in `functions:delete` where you can delete arbitrary function as long as you specify it's name correctly. I think that might be useful, though we might want to revisit that API.

odeniso:
Hello! Do you have an idea of when this change will be released. Otherwise, do you know which version introduced the bug that this PR addresses?

I am hoping this fixes an issue we are facing, where cloud functions get deployed with runtime options that are different than those specified via `runWith`.

taeold:
@odeniso Unfortunately, this bug excited for quite some time. I'll try to get the team to push out a release containing this change soon.

gustavopch:
I also need this to get released as soon as possible. I have a function that parses a Photoshop file and it's crashing with "JavaScript heap out of memory" frequently. Firebase Support pointed me to this workaround https://cloud.google.com/functions/docs/troubleshooting#javascript_heap_out_of_memory, but apparently, it doesn't work with `firebase-tools` and I'll need to manually update `NODE_OPTIONS` through the Cloud Console GUI after every deployment if I want the function to keep working.

src/deploy/functions/release/planner.ts
```diff
@@ -88,11 +112,13 @@ export function createDeploymentPlan(
 
   const regions = new Set([...Object.keys(want.endpoints), ...Object.keys(have.endpoints)]);
   for (const region of regions) {
-    deployment[region] = calculateRegionalChanges(
+    const changesets = calculateChangesets(
       want.endpoints[region] || {},
       have.endpoints[region] || {},
+      (e) => `${e.region}-${e.availableMemoryMb || "default"}`,
```
inlined, line 118: I don't think we actually do heap tuning for all memory sizes. It might be worth talking to the cloud folks to see if we can merge deployments of <2GB

inlined (commented):
I love how clean this is, though, per offline conversation, I'm worried about the impact this could have on deletes. We're talking about some obscure edge cases, but if I rename a function _and change its memory_ allocation in the same change, we can now delete the old function before creating the new one.

Should we fall back to just deleting after all creates succeed again? Should I sweat these edge cases less?

src/deploy/functions/release/planner.ts
taeold: Looked in and discovered that the environment variable is set per memory configuration:

https://github.com/GoogleCloudPlatform/buildpacks/blob/663180f2cf1fbfdee4a889f6c9d17ee2eb9091bf/cmd/nodejs/functions_framework/main.go#L166

Seems like they took the simplest path forward.

PR #4255, opened 2022-03-05
static verification code in emulator

cedvdb:
This PR makes the verification code for phone authentication in the emulator to be static. The reason to use a static code instead of a randomly generated one is that it's much easier to predict in testing. It's also easier to predict in manual testing which makes the process easier where it is not needed to have to look at the emulator stdout.

I do not see a good reason to have random digits here.


cedvdb:
@joehan This is a very small change, could you launch the workflow action ?

cedvdb:
@taeold @tonyjhuang

PR #4257, opened 2022-03-07
Merge ext:* --local launch branch into master

elvisun:
### Description

The extensions config for your project is saved in firebase.json and an .env file for each extension. You can read more about it [here](https://firebase.google.com/docs/extensions/reuse-project-config).

To make this experience better, we are adding a --local flag to all extensions commands. This will allow users to install and manage extensions to your configuration files. To deploy the changes to your firebase projects, run `firebase deploy (--only extensions)`.

This allow us to move towards a declarative approach on managing extensions in your project using the CLI.

In the next breaking release, we plan to make this behavior the default.


src/commands/ext-install.ts
```diff
@@ -200,6 +235,62 @@ interface InstallExtensionOptions {
...
+/**
+ * Installs the extension in user's project.
+ *
+ * 1. Checks products are provisioned.
+ * 2. Checks billings are enabled if needed.
+ * 3. Asks for permission to grant sa roles.
+ * 4. Asks for extension params
+ * 5. Install
+ * @param options
+ * @returns
```
joehan, line 292: Nit: remove this empty @returns

src/commands/ext-update.ts
```diff
@@ -62,10 +65,96 @@ export default new Command("ext:update <extensionInstanceId> [updateSource]")
   .before(diagnoseAndFixProject)
   .withForce()
   .option("--params <paramsFile>", "name of params variables file with .env format.")
-  .action(async (instanceId: string, updateSource: string, options: any) => {
+  .option(
+    "--local",
+    "save the update to firebase.json rather than directly update at a Firebase project"
```
joehan, line 70: Nit: consider rewording this to "rather than directly update an existing Extension instance on a Firebase project"

src/extensions/manifest.ts
```diff
@@ -72,3 +126,87 @@ async function writeEnvFiles(
...
+}
+
+// TODO(lihes): Add a docs link once exists.
+/**
+ * Show deprecation warning about --local flag taking over current default bahaviors.
+ */
+export function showDeprecationWarning() {
+  utils.logLabeledWarning(
+    logPrefix,
+    "The behavior of the extensions commands will change in firebase-tools@11.0.0. " +
```
joehan, line 190: Nit: Replace with "The behavior of ext:install, ext:update, ext:configure, and ext:uninstall"

joehan:
One more thing I just thought of - when using --local, there should be a way to successfully run these commands without passing in a projectId, so that we have a nice story for testing using the emulator without ever needing a real project. Probably worth doing in its own PR onto this branch?

elvisun:
> One more thing I just thought of - when using --local, there should be a way to successfully run these commands without passing in a projectId, so that we have a nice story for testing using the emulator without ever needing a real project. Probably worth doing in its own PR onto this branch?

That's a good idea! We only use project ID for creating the secrets, we'll need to wait until we have secrets figured out before we can do that. Filed https://b.corp.google.com/issues/223210586 to track this

src/commands/ext-configure.ts
```diff
@@ -27,18 +31,80 @@ export default new Command("ext:configure <extensionInstanceId>")
...
+      );
+      infoImmutableParams(immutableParams, oldParamValues);
+
+      // Ask for mutable param values from user.
+      paramHelper.setNewDefaults(tbdParams, oldParamValues);
+      const mutableParamsValues = await paramHelper.getParams({
+        projectId,
+        paramSpecs: tbdParams,
+        nonInteractive: false,
+        paramsEnvPath: options.params as string,
```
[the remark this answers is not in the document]
elvisun: good catch, had this in the other files but forgot in this one...

src/extensions/manifest.ts
elvisun: good call

elvisun:
Going to merge this to get off the feature branch, since the secrets work will still take some time to complete

PR #4258, opened 2022-03-07
Adds new alerting event provider

colerogers:
Adds Firebase Alerts as a deployable event provider.

```
```

note: we cannot deploy until the container contract changes are merged


src/deploy/functions/services/firebaseAlerts.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @param eventTrigger the endpoints event trigger
+ */
+export function ensureFirebaseAlertsTriggerRegion(
+  endpoint: backend.Endpoint & backend.EventTriggered
+): void {
+  if (!endpoint.eventTrigger.region) {
+    endpoint.eventTrigger.region = "global";
+  }
+  if (endpoint.eventTrigger.region !== "global") {
+    throw new FirebaseError("A firebase alerts function must have a 'global' trigger location");
```
taeold, line 47: I think having error message that is function-neutral is going to last longer: 
 
```suggestion
    throw new FirebaseError("A firebase alerts trigger must specify 'global' trigger location");
```

src/deploy/functions/triggerRegionHelper.ts
```diff
@@ -7,7 +7,7 @@ import { serviceForEndpoint } from "./services";
  * @param have the list of function specs we have deployed
  */
 export async function ensureTriggerRegions(want: backend.Backend): Promise<void> {
-  const regionLookups: Array<Promise<void>> = [];
+  const regionLookups: Array<Promise<void> | void> = [];
```
taeold, line 10: huh I wonder why this change was necessary 🤔

src/deploy/functions/services/firebaseAlerts.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Finds the required project level IAM bindings for the Pub/Sub service agent
+ * If the user enabled Pub/Sub on or before April 8, 2021, then we must enable the token creator role
+ * @param projectId project identifier
+ * @param existingPolicy the project level IAM policy
+ */
+export async function obtainFirebaseAlertsBindings(
+  projectId: string,
+  existingPolicy: iam.Policy
+): Promise<Array<iam.Binding>> {
+  const projectNumber = await getProjectNumber({ projectId });
```
taeold, line 18: I'd like for us to avoid making extra API call if possible; it adds small amount of delay in our deploy (though I agree that it isn't that much)

It looks like we are the only caller of `ensureServiceAgentRoles` which ultimately makes call to this function. Can we pass around projectNumber in `prepare.ts` so subsequent calls to `getProjectNumber` is properly cached instead of making one-off calls here?

src/deploy/functions/triggerRegionHelper.ts
colerogers: This was added for the typescript compiler. We don't need a lookup for the trigger region in firealerts and the corresponding dispatch function is not async, just returns void

src/deploy/functions/services/firebaseAlerts.ts
colerogers: Great point, I added this fetch call in prepare and I passed both projectId & projectNumber into requireProjectBindings in the latest commit. I combined them into an object, let me know if you prefer them to be separate params

src/deploy/functions/services/firebaseAlerts.ts
colerogers: Fixed

src/deploy/functions/prepare.ts
```diff
@@ -37,6 +38,7 @@ export async function prepare(
   payload: args.Payload
 ): Promise<void> {
   const projectId = needProjectId(options);
+  const projectNumber = await getProjectNumber({ projectId });
```
taeold, line 41: Check out definition for [`getProjectNumber`](https://github.com/firebase/firebase-tools/blob/8fc53ac15f0624aae86cca32c4f8eb438054b696/src/getProjectNumber.ts#L8).

It's designed to receive `options` here so that when you later call `getProjectNumber` with the same `options` object, you'd use the cached value instead of making yet another API call.

In fact, I'd actually suggest that we use [`needProjectNumber`](https://github.com/firebase/firebase-tools/blob/094faede699f7a9070ba641fdcbc10d1883803ba/src/projectUtils.ts#L84) instead.

src/deploy/functions/checkIam.ts
```diff
@@ -148,7 +148,7 @@ export function mergeBindings(policy: iam.Policy, allRequiredBindings: iam.Bindi
  * @param have backend that we have currently deployed
  */
 export async function ensureServiceAgentRoles(
-  projectId: string,
+  project: { projectId: string; projectNumber: string },
```
taeold, line 151: nit* In theory, having just `projectNumber` should be sufficient since all GCP API calls accept either projectId or projectNumber. Will leave it up to you to decide whether that would be appropriate.

src/deploy/functions/services/index.ts
```diff
@@ -2,17 +2,25 @@ import * as backend from "../backend";
...
 /** A service interface for the underlying GCP event services */
 export interface Service {
   readonly name: string;
   readonly api: string;
 
   // dispatch functions
-  requiredProjectBindings: ((pId: any, p: any) => Promise<Array<iam.Binding>>) | undefined;
-  ensureTriggerRegion: (ep: backend.Endpoint & backend.EventTriggered) => Promise<void>;
+  requiredProjectBindings:
+    | ((project: Project, policy: iam.Policy) => Promise<Array<iam.Binding>> | Array<iam.Binding>)
```
taeold, line 21: Since this type got really complicated, what if we instead keep the original type signature and instead have `obtainFirebaseAlertsBindings` return a promise instead? e.g.

```ts
export function obtainFirebaseAlertsBindings(
  project: { projectId: string; projectNumber: string },
  existingPolicy: iam.Policy
): Promise<Array<iam.Binding>> {
  // ...implementation
  return Promise.resolve([pubsubBinding]);
}
```

src/deploy/functions/services/index.ts
```diff
@@ -2,17 +2,25 @@ import * as backend from "../backend";
...
+    | undefined;
+  ensureTriggerRegion: (ep: backend.Endpoint & backend.EventTriggered) => Promise<void> | void;
```
taeold, line 23: Same trick as I mentioned above - what if we "clean" this type up by requiring everything to return a promise - it is easy to make non-promise returning fns to return a promise instead.

src/deploy/functions/triggerRegionHelper.ts
taeold: Repeating message above:

what if we "clean" this type up by requiring everything to return a promise - it is easy to make non-promise returning fns to return a promise instead.

src/deploy/functions/checkIam.ts
colerogers: Oh snap, I didn't realize treated them as interchangeable!!

src/deploy/functions/prepare.ts
colerogers: Ahh, thanks for pointing this out, fixed in the latest commit

src/deploy/functions/services/index.ts
colerogers: fixed in the latest commit

src/deploy/functions/services/index.ts
colerogers: fixed in the latest commit

src/deploy/functions/triggerRegionHelper.ts
colerogers: fixed in the latest commit

src/deploy/functions/checkIam.ts
```diff
@@ -148,7 +148,7 @@ export function mergeBindings(policy: iam.Policy, allRequiredBindings: iam.Bindi
  * @param have backend that we have currently deployed
  */
 export async function ensureServiceAgentRoles(
-  projectId: string,
+  projectNumber: string,
```
inlined, line 151: per lint, please update javadoc to match the new param name.

src/deploy/functions/services/firebaseAlerts.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const SERVICE_ACCOUNT_TOKEN_CREATOR_ROLE = "roles/iam.serviceAccountTokenCreator";
+
+/**
+ * Finds the required project level IAM bindings for the Pub/Sub service agent
+ * If the user enabled Pub/Sub on or before April 8, 2021, then we must enable the token creator role
+ * @param projectId project identifier
+ * @param existingPolicy the project level IAM policy
+ */
+export function obtainFirebaseAlertsBindings(
```
inlined, line 14: Is this really a Firebase Alerts feature or something that all event providers are going to need?

src/deploy/functions/services/firebaseAlerts.ts
```diff
@@ -0,0 +1,49 @@
...
+    pubsubBinding = {
+      role: SERVICE_ACCOUNT_TOKEN_CREATOR_ROLE,
+      members: [],
+    };
+  }
+  if (!pubsubBinding.members.find((m) => m === pubsubServiceAgent)) {
+    pubsubBinding.members.push(pubsubServiceAgent);
+  }
+  return Promise.resolve([pubsubBinding]);
+}
```
inlined, line 32: This logic happens over and over and over again in our codebase. Maybe we should have something in gcp/iam.ts that ensures a binding exists or is removed?

src/deploy/functions/services/index.ts
```diff
@@ -30,12 +33,19 @@ export const PubSubService: Service = {
...
+export const StorageService: Service = {
   name: "storage",
   api: "storage.googleapis.com",
   requiredProjectBindings: obtainStorageBindings,
   ensureTriggerRegion: ensureStorageTriggerRegion,
 };
+/** A firebase alerts service object */
+export const FirebaseAlertsService: Service = {
+  name: "firebasealerts",
+  api: "logging.googleapis.com",
```
inlined, line 45: This is really the API for Firebase Alerts?? How is this not a conflict with Cloud Audit Logs?

src/deploy/functions/services/firebaseAlerts.ts
inlined: NVM. See below comments. I think we could have had something like:

```
export function obtainFirebaseAlertBindings(projectNumber) {
  iam.ensureBinding(firebaseAlertRole, firebaseAlertPermission)
}
```

src/deploy/functions/checkIam.ts
colerogers: This was already updated in a subsequent PR, master has the correct tsdoc

src/deploy/functions/services/firebaseAlerts.ts
colerogers: You're correct, this is not a Firebase Alerts specific binding. It was yanked in #4324 and I added a noop instead

src/deploy/functions/services/index.ts
colerogers: This was just a placeholder, I'll update it with the correct API. Additionally, this field is not currently being used in the CLI

PR #4262, opened 2022-03-07
Move security rules evaluation to StorageLayer

tonyjhuang:
### Description

Refactors security rules evaluation out of `firebase.ts` into `StorageLayer` for the following methods:

* Update object metadata
* Delete object
* List objects


src/emulator/storage/apis/gcloud.ts
```diff
@@ -39,38 +40,54 @@ export function createCloudEndpoints(emulator: StorageEmulator): Router {
...
+            decodedObjectId: req.params.objectId,
+          },
+          /* skipAuth = */ true
+        );
+      } catch (err) {
+        if (err instanceof NotFoundError) {
+          return res.sendStatus(404);
+        }
+        if (err instanceof ForbiddenError) {
+          throw new Error("Request failed unexpectedly due to Firebase Rules.");
```
tohhsinpei, line 58: Curious why we don't send a 403 here?

src/emulator/storage/files.ts
```diff
@@ -177,11 +207,17 @@ export class StorageLayer {
...
+  public async handleGetObject(
+    request: GetObjectRequest,
+    skipAuth = false
+  ): Promise<GetObjectResponse> {
     const metadata = this.getMetadata(request.bucketId, request.decodedObjectId);
 
+    let authorized = skipAuth;
     // If a valid download token is present, skip Firebase Rules auth. Mainly used by the js sdk.
-    let authorized = (metadata?.downloadTokens || []).includes(request.downloadToken ?? "");
+    if (!authorized) {
```
tohhsinpei, line 218: Nit: I'd just do an or

src/emulator/storage/files.ts
```diff
@@ -135,6 +135,36 @@ export type GetObjectResponse = {
...
+export type ListObjectsRequest = {
+  bucketId: string;
+  prefix: string;
+  delimiter: string;
+  pageToken?: string;
+  maxResults?: number;
+  authorization?: string;
+};
+
+/**  Response object for {@link StorageLayer#handleListObjects}. */
```
tohhsinpei, line 163: I probably wouldn't bother making a new type; just return a `ListResponse`

src/emulator/storage/apis/firebase.ts
```diff
@@ -141,76 +152,57 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
-    if (
-      !(await isPermitted({
-        ruleset: emulator.rules,
-        method: RulesetOperationMethod.LIST,
-        path: operationPath,
-        file: {},
+    const maxResults = req.query.maxResults?.toString();
+    let response: ListObjectsResponse;
+    try {
+      response = await storageLayer.handleListObjects({
```
tohhsinpei, line 186: Can this ever return a 404? What happens if I put a nonsense bucket id?

src/emulator/storage/apis/firebase.ts
```diff
@@ -470,48 +464,35 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
   firebaseStorageAPI.put("/b/:bucketId/o/:objectId?", async (req, res) => {
     switch (req.header("x-http-method-override")?.toLowerCase()) {
       case "patch":
-        return handleMetadataUpdate(req, res);
+        return await handleMetadataUpdate(req, res);
```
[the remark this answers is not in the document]
tonyjhuang: good to know thanks!

src/emulator/storage/files.ts
```diff
@@ -215,10 +136,10 @@ export class StorageLayer {
...
     let authorized = skipAuth;
     // If a valid download token is present, skip Firebase Rules auth. Mainly used by the js sdk.
-    if (!authorized) {
-      authorized = (metadata?.downloadTokens || []).includes(request.downloadToken ?? "");
-    }
-    if (!authorized) {
+    const hasValidDownloadToken = (metadata?.downloadTokens || []).includes(
+      request.downloadToken ?? ""
+    );
+    if (!authorized || hasValidDownloadToken) {
```
tohhsinpei, line 142: `!(authorized || hasValidDownloadToken)`

scripts/storage-emulator-integration/tests.ts
```diff
@@ -1067,27 +1067,18 @@ describe("Storage emulator", () => {
...
-                  .on("end", () => {
-                    expect(Buffer.concat(data)).to.deep.equal(
-                      Buffer.from(IMAGE_FILE_BASE64, "base64")
-                    );
-                  })
-                  .on("close", resolve)
-                  .on("error", reject);
-              }
-            );
+            requestClient.get(downloadUrl, (response) => {
```
tohhsinpei, line 1070: Do you not need to pass an empty object as the options?

src/emulator/storage/apis/gcloud.ts
```diff
@@ -292,6 +292,7 @@ export function createCloudEndpoints(emulator: StorageEmulator): Router {
   });
 
   gcloudStorageAPI.get("/:bucketId/:objectId(**)", async (req, res) => {
+    console.log("this cant be right");
```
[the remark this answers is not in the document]
tonyjhuang: done

scripts/storage-emulator-integration/tests.ts
tonyjhuang: no, `get` has two versions, a 3 arg version that takes options and 2 arg version that takes just the path and callback

src/emulator/storage/files.ts
```diff
@@ -134,12 +134,12 @@ export class StorageLayer {
   ): Promise<GetObjectResponse> {
     const metadata = this.getMetadata(request.bucketId, request.decodedObjectId);
 
-    let authorized = skipAuth;
     // If a valid download token is present, skip Firebase Rules auth. Mainly used by the js sdk.
     const hasValidDownloadToken = (metadata?.downloadTokens || []).includes(
       request.downloadToken ?? ""
     );
-    if (!authorized || hasValidDownloadToken) {
+    let authorized = skipAuth || !!hasValidDownloadToken;
```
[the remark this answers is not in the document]
tonyjhuang: done

src/emulator/storage/apis/firebase.ts
```diff
@@ -104,7 +110,12 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
       } else if (err instanceof ForbiddenError) {
-        return res.sendStatus(403);
+        return res.status(403).json({
+          error: {
+            code: 403,
+            message: `Permission denied. No READ permission.`,
+          },
+        });
       }
       throw err;
```
abhis3, line 120: Not entirely familiar of the standard for when we're throwing an error vs returning a status code, but could we throw a `400` error as the catch all case?

src/emulator/storage/apis/firebase.ts
```diff
@@ -141,76 +152,57 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
-      !(await isPermitted({
-        ruleset: emulator.rules,
-        method: RulesetOperationMethod.UPDATE,
-        path: operationPath,
+    let metadata: StoredFileMetadata;
+    try {
+      metadata = await storageLayer.handleUpdateObjectMetadata({
+        bucketId: req.params.bucketId,
+        decodedObjectId: decodeURIComponent(req.params.objectId),
+        metadata: req.body as IncomingMetadata,
```
abhis3, line 160: We should avoid type assertions

https://g3doc.corp.google.com/javascript/typescript/g3doc/dev/readability/styleguide.md?cl=head#type-and-non-nullability-assertions

src/emulator/storage/apis/gcloud.ts
abhis3, line 58: Why are we not returning a `403` here?

src/emulator/storage/apis/firebase.ts
tonyjhuang: The intent of throwing an error here is to signal there is something fundamentally wrong with the emulator, ie an unexpected error case. Returning a 400 Bad Request can be confusing to users as they would assume something wrong with their request as opposed to something broken in the emulator. A better error code to return would be a 500 Internal, but raising an error and letting the emulator die is easier to debug than trying to understand why the "server" is failing

src/emulator/storage/apis/gcloud.ts
tonyjhuang: Same comment as above, this is to signal a critical error in the emulator, as the current assumption is this endpoint should never fail with a 403

tohhsinpei (approved):

src/emulator/storage/apis/firebase.ts
abhis3: I see makes sense, thanks for the info!

Issue #3390, opened 2021-05-20
Storage emulator does not accept multiple storage targets configuration


DennisAlund:
### [REQUIRED] Environment info

**firebase-tools:** 9.11.0

**Platform:**  macOS 11.3.1

### [REQUIRED] Test case
Follow the steps to set up multiple storage targets: https://firebase.google.com/docs/cli/targets#set-up-deploy-target-storage-database

Ensure that the `firebase.json` configuration has an array of storage targets 
```jsonc
{
 // ... your other firebase configuration
 "storage": [
    {
      "rules": "storage.rules",
      "target": "default"
    },
    {
      "rules": "our-other-target.rules",
      "target": "our-other-target"
    }
  ]
  // ... rest of configuration
}
```

### [REQUIRED] Steps to reproduce
1. Configure multiple storage targets as described https://firebase.google.com/docs/cli/targets#set-up-deploy-target-storage-database
2. Start emulators
3. **See error:** _Cannot start the Storage emulator without rules file specified in firebase.json: run 'firebase init' and set up your Storage configuration_
4. Remove the array of targets and declare single target as object, not array declaration
5. Start emulators
6. **Success**

### [REQUIRED] Expected behavior
Configuration that is acceptable for regular deployment should be acceptable for emulator.

### [REQUIRED] Actual behavior
Get error when starting the emulators with multiple storage targets

#### Output
```
$ firebase emulators:start
i  emulators: Starting emulators: auth, functions, firestore, database, hosting, pubsub, storage
✔  functions: Using node@14 from host.
i  firestore: Firestore Emulator logging to firestore-debug.log
i  database: Database Emulator logging to database-debug.log
i  pubsub: Pub/Sub Emulator logging to pubsub-debug.log
i  emulators: Shutting down emulators.
i  functions: Stopping Functions Emulator
i  database: Stopping Database Emulator
i  firestore: Stopping Firestore Emulator
i  pubsub: Stopping Pub/Sub Emulator
i  auth: Stopping Authentication Emulator
i  hub: Stopping emulator hub

Error: Cannot start the Storage emulator without rules file specified in firebase.json: run 'firebase init' and set up your Storage configuration
```

#### Output with debug flags

```
[2021-05-20T06:58:23.662Z] > command requires scopes: ["email","openid","https://www.googleapis.com/auth/cloudplatformprojects.readonly","https://www.googleapis.com/auth/firebase","https://www.googleapis.com/auth/cloud-platform"]
[...]

PR #4263, opened 2022-03-07
Add method to parse multiple targets for Storage Emulator rules

tohhsinpei:
### Description
Part of fix for #3390.

This PR adds `getStorageRulesConfig` to parse multiple rules files and targets from the Storage config. 

After this, it remains to map resource to `StorageRulesManagers` (#4245) in the Storage Emulator and pass resource in each request.

### DIFF
--- a/src/test/emulators/storage/rules/config.spec.ts
+++ b/src/test/emulators/storage/rules/config.spec.ts
@@ -1,0 +1,91 @@
+import * as path from "path";
+import { expect } from "chai";
+import { readFileSync } from "fs-extra";
+import { tmpdir } from "os";
+import { v4 as uuidv4 } from "uuid";
+
+import { Config } from "../../../../config";
+import { Options } from "../../../../options";
+import { RC } from "../../../../rc";
+import { getStorageRulesConfig } from "../../../../emulator/storage/rules/config";
+import { StorageRulesFiles } from "../../fixtures";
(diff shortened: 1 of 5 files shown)


src/test/emulators/storage/rules/config.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+import { RC } from "../../../../rc";
+import { getStorageRulesConfig } from "../../../../emulator/storage/rules/config";
+import { StorageRulesFiles } from "../../fixtures";
+import { Persistence } from "../../../../emulator/storage/persistence";
+import { FirebaseError } from "../../../../error";
+
+const PROJECT_ID = "test-project";
+
+describe("Storage Rules Config", () => {
+  const tmpDir = `${tmpdir()}/${uuidv4()}`;
```
tonyjhuang, line 18: I know I did this in another PR but I think there's a more canonical way to do this. See https://github.com/firebase/firebase-tools/blob/master/scripts%2Fstorage-emulator-integration%2Ftests.ts#L112

tonyjhuang (approved):
LGTM outside of two tiny things, thanks for the PR!

src/test/emulators/storage/rules/config.spec.ts
tohhsinpei: Done.

PR #4269, opened 2022-03-08
Allow multiple-config form for functions config in firebase.json

taeold:
This change extends `functions` config in `firebase.json` so that you can write it in multi-config form. E.g.

```json
// This has always been valid
{
  "functions": {
    "source": "source"
  }
}

// This is also valid now
{
  "functions": [
    {
      "source": "source"
    }
  ]
}
```

Since having multiple configuration for function doesn't make sense (yet), we add runtime checks to make sure that there is exactly 1 functions source if we have an array of function config. I had fun trying to add better types around configs and consolidated config validations to remove redundant config checks throughout the codebase.


src/deploy/functions/deploy.ts
```diff
@@ -77,16 +77,10 @@ export async function deploy(
...
-      "Error: 'functions.source' is not defined"
-    );
+    const source = context.config.source;
     if (uploads.length) {
       logSuccess(
-        clc.green.bold("functions:") +
-          " " +
-          clc.bold(options.config.src.functions.source) +
-          " folder uploaded successfully"
+        clc.green.bold("functions:") + " " + clc.bold(source) + " folder uploaded successfully"
```
bkendall, line 83:
```suggestion
        `${clc.green.bold("functions:")} ${clc.bold(source)} folder uploaded successfully`
```

src/deploy/functions/prepare.ts
```diff
@@ -127,12 +128,17 @@ export async function prepare(
...
+      runtimeConfig
+    );
   }
   if (backend.someEndpoint(wantBackend, (e) => e.platform === "gcfv2")) {
     context.functionsSourceV2 = await prepareFunctionsUpload(
-      /* runtimeConfig= */ undefined,
-      options
+      sourceDir,
+      context.config,
+      /* runtimeConfig= */ undefined
```
bkendall, line 141: Can be omitted if it's undefined.

src/deploy/functions/prepareFunctionsUpload.ts
```diff
@@ -116,15 +114,9 @@ async function packageSource(options: Options, sourceDir: string, configValues:
 }
 
 export async function prepareFunctionsUpload(
-  runtimeConfig: backend.RuntimeConfigValues | undefined,
-  options: Options
+  sourceDir: string,
+  config: projectConfig.ValidatedSingle,
+  runtimeConfig: backend.RuntimeConfigValues | undefined
```
bkendall, line 119: Can use an optional argument

```suggestion
  runtimeConfig?: backend.RuntimeConfigValues
```

src/firebaseConfig.ts
```diff
@@ -102,13 +102,14 @@ export type FirestoreConfig = {
...
 
-export type FunctionsConfig = {
-  // TODO: Add types for "backend"
+export type FunctionsSingle = {
   source?: string;
   ignore?: string[];
   runtime?: CloudFunctionRuntimes;
 } & Deployable;
 
+export type FunctionsConfig = FunctionsSingle | FunctionsSingle[];
```
bkendall, line 111: I'm not sure if we have other guidance on this or not, but why not:

```typescript
export type FunctionConfig = {}

export type FunctionsConfig = FunctionConfig | FunctionConfig[];
```

Maybe it's a little "too close" with being plural or not, but "FunctionsSingle" seems... I dunno, I wouldn't know what to expect from that.

src/functions/projectConfig.ts
```diff
@@ -0,0 +1,50 @@
...
+import { FirebaseError } from "../error";
+
+export type NormalizedConfig = [FunctionsSingle, ...FunctionsSingle[]];
+export type ValidatedSingle = FunctionsSingle & { source: string };
+export type ValidatedConfig = [ValidatedSingle];
+
+/**
+ * Normalize functions config to return functions config in an array form.
+ */
+export function normalize(config: FunctionsConfig | undefined): NormalizedConfig {
```
bkendall, line 11:
```suggestion
export function normalize(config?: FunctionsConfig): NormalizedConfig {
```

src/functions/projectConfig.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (config === undefined) {
```
bkendall, line 12: Does it need to specifically equal undefined?

```suggestion
  if (!config) {
```

src/functions/projectConfig.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+  return [validateSingle(config[0])];
+}
+
+/**
+ * Normalize and validate functions config.
+ *
+ * Valid functions config has exactly one config and has all required fields set.
+ */
+export function normalizeAndValidate(config: FunctionsConfig | undefined): ValidatedConfig {
```
bkendall, line 48:
```suggestion
export function normalizeAndValidate(config?: FunctionsConfig): ValidatedConfig {
```

src/functions/projectConfig.ts
```diff
@@ -0,0 +1,50 @@
+import { FunctionsConfig, FunctionsSingle } from "../firebaseConfig";
+import { FirebaseError } from "../error";
+
+export type NormalizedConfig = [FunctionsSingle, ...FunctionsSingle[]];
```
bkendall, line 4: So, it's `FunctionsSingle[]`? Why is a special config type needed?

src/serve/functions.ts
```diff
@@ -9,6 +9,7 @@ import { parseRuntimeVersion } from "../emulator/functionsEmulatorUtils";
 import { needProjectId } from "../projectUtils";
 import { getProjectDefaultAccount } from "../auth";
 import { Options } from "../options";
+import * as functionsConfig from "../functions/projectConfig";
```
bkendall, line 12: I find this rename confusing. Based on the file name, it should be `projectConfig`

```suggestion
import * as projectConfig from "../functions/projectConfig";
```

src/test/functions/projectConfig.spec.ts
```diff
@@ -0,0 +1,74 @@
+import { expect } from "chai";
+
+import * as projectConfig from "../../functions/projectConfig";
+import { FirebaseError } from "../../error";
+
+const TEST_CONFIG = { source: "foo" };
+
+describe("projectConfig", () => {
+  describe("normalize", () => {
```
bkendall, line 9: What if it's given more than one? Should that have a test?

bkendall (changes requested):

src/firebaseConfig.ts
taeold: Lol. Naming things is hard. `Config` vs `Configs` was also a thing that was messing up my brain 🤔 .

Going ahead w/ your suggestion.

src/functions/projectConfig.ts
taeold: This is probably fine!

src/functions/projectConfig.ts
taeold: This type definition enforces that there be at least one item in the array. Too hard to read?

src/test/functions/projectConfig.spec.ts
taeold: That is checked in `validate`, and we have a test case for that.

src/functions/projectConfig.ts
bkendall: It's maybe a little weird, but I guess I can live with it hah.

bkendall (approved):
Looks fine to me, but I think someone from functions should really take a look as well

src/functions/projectConfig.ts
```diff
@@ -0,0 +1,50 @@
...
+export type NormalizedConfig = [FunctionConfig, ...FunctionConfig[]];
+export type ValidatedSingle = FunctionConfig & { source: string };
+export type ValidatedConfig = [ValidatedSingle];
+
+/**
+ * Normalize functions config to return functions config in an array form.
+ */
+export function normalize(config?: FunctionsConfig): NormalizedConfig {
+  if (!config) {
+    throw new FirebaseError("No valid functions configuration detected in firebase.json");
```
colerogers, line 13: I think I'd prefer this error to be thrown before we call this function, like in `functions-config-export.ts` so we don't have an optional param. But this shouldn't block the PR

src/test/functions/projectConfig.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    it("normalizes singleton config", () => {
+      expect(projectConfig.normalize(TEST_CONFIG)).to.deep.equal([TEST_CONFIG]);
+    });
+
+    it("normalizes array config", () => {
+      expect(projectConfig.normalize([TEST_CONFIG])).to.deep.equal([TEST_CONFIG]);
+    });
+
+    it("throws error if given empty config", () => {
+      expect(() => projectConfig.normalize([])).to.throw(FirebaseError);
```
colerogers, line 19: nit - don't we usually check for the expected error message

src/test/functions/projectConfig.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+  });
+
+  describe("validate", () => {
+    it("passes validation for simple config", () => {
+      expect(projectConfig.validate([TEST_CONFIG])).to.deep.equal([TEST_CONFIG]);
+    });
+
+    it("fails validation given more than one config", () => {
+      expect(() =>
+        projectConfig.validate([TEST_CONFIG, { ...TEST_CONFIG, source: "bar" }])
```
colerogers, line 30: why don't you just write `projectConfig.validate([TEST_CONFIG, { source: "bar" }])`?

src/test/functions/projectConfig.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+      ).to.throw(FirebaseError);
```
colerogers, line 31: same with this error

src/test/functions/projectConfig.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    });
+
+    it("fails validation given config w/o source", () => {
+      expect(() => projectConfig.validate([{ runtime: "nodejs10" }])).to.throw(FirebaseError);
```
colerogers, line 35: same with this error

src/test/functions/projectConfig.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    });
+
+    it("fails validation given config w/ empty source", () => {
+      expect(() => projectConfig.validate([{ source: "" }])).to.throw(FirebaseError);
```
colerogers, line 39: same with this error

src/test/functions/projectConfig.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+      expect(projectConfig.normalizeAndValidate(TEST_CONFIG)).to.deep.equal([TEST_CONFIG]);
+    });
+
+    it("returns normalized config for multi-resource config", () => {
+      expect(projectConfig.normalizeAndValidate([TEST_CONFIG])).to.deep.equal([TEST_CONFIG]);
+    });
+
+    it("fails validation given singleton config w/o source", () => {
+      expect(() => projectConfig.normalizeAndValidate({ runtime: "nodejs10" })).to.throw(
+        FirebaseError
```
colerogers, line 54: same with this error

src/test/functions/projectConfig.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+      );
+    });
+
+    it("fails validation given singleton config w empty source", () => {
+      expect(() => projectConfig.normalizeAndValidate({ source: "" })).to.throw(FirebaseError);
```
colerogers, line 59: same with this error

src/test/functions/projectConfig.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    });
+
+    it("fails validation given multi-resource config w/o source", () => {
+      expect(() => projectConfig.normalizeAndValidate([{ runtime: "nodejs10" }])).to.throw(
+        FirebaseError
```
colerogers, line 64: same with this error

src/test/functions/projectConfig.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+      );
+    });
+
+    it("fails validation given more than one config", () => {
+      expect(() =>
+        projectConfig.normalizeAndValidate([TEST_CONFIG, { ...TEST_CONFIG, source: "bar" }])
+      ).to.throw(FirebaseError);
```
colerogers, line 71: same with this error

src/config.ts
```diff
@@ -76,13 +76,17 @@ export class Config {
...
+    if (this.projectDir && fsutils.dirExistsSync(this.path(Config.DEFAULT_FUNCTIONS_SOURCE))) {
+      if (Array.isArray(this.get("functions"))) {
+        if (!this.get("functions.[0].source")) {
+          this.set("functions.[0].source", Config.DEFAULT_FUNCTIONS_SOURCE);
+        }
+      } else {
+        if (!this.get("functions.source")) {
+          this.set("functions.source", Config.DEFAULT_FUNCTIONS_SOURCE);
+        }
+      }
```
colerogers, line 89: Can we unwind this a bit?
```
if (!this.projectDir || !fsutils.dirExistsSync(this.path(Config.DEFAULT_FUNCTIONS_SOURCE))) {
    return;
}
if (Array.isArray(this.get("functions"))) {
    if (!this.get("functions.[0].source")) {
        this.set("functions.[0].source", Config.DEFAULT_FUNCTIONS_SOURCE);
    }
} else {
    if (!this.get("functions.source")) {
        this.set("functions.source", Config.DEFAULT_FUNCTIONS_SOURCE);
    }
}
```

Actually, I'm not sure if that makes it much easier to read

colerogers (approved):

src/functions/projectConfig.ts
```diff
@@ -0,0 +1,50 @@
+import { FunctionsConfig, FunctionConfig } from "../firebaseConfig";
+import { FirebaseError } from "../error";
+
+export type NormalizedConfig = [FunctionConfig, ...FunctionConfig[]];
```
inlined, line 4: Is this the type definition of an array with at least one element?

src/firebaseConfig.ts
```diff
@@ -102,13 +102,14 @@ export type FirestoreConfig = {
...
 
-export type FunctionsConfig = {
-  // TODO: Add types for "backend"
+export type FunctionConfig = {
   source?: string;
   ignore?: string[];
   runtime?: CloudFunctionRuntimes;
 } & Deployable;
 
+export type FunctionsConfig = FunctionConfig | FunctionConfig[];
```
inlined, line 111: Function*s* vs Function config is incredibly subtle. I might prefer names as clear as FunctionsSingle | FunctionsMultiple

inlined (commented):
I feel like I missed my baby's first steps. This is looking good and I'm excited to catch up.

PR #4271, opened 2022-03-09
Cloud Functions: Fix env var escaping

### LINKED ISSUE #4270: Cloud Functions: Incorrect/incomplete transformation of escape sequences in .env files

**firebase-tools:** 10.2.2

**Platform:** macOS (11.6.2 Big Sur)

### [REQUIRED] Test case

1. Create a new Node project

   ```sh
   mkdir firebasetoolstest
   cd firebasetoolstest
   npm init # (accept default answers)
   ```

2. Add this package as a devDependency

   ```sh
   npm i -DE firebase-tools@10.2.2
   ```

3. Init a new Firebase project

   ```sh
   npx firebase init
   # Which Firebase features: enable 'Functions'
   # Project setup: Create a new project
   # Please specify a unique project id: (your choice)
   # What would you like to call your project: (your choice)
   # What language would you like to use to write Cloud Functions?: JavaScript
   # Do you want to use ESLint to catch probable bugs and enforce style? N
   # Do you want to install dependencies with npm now? Y
   ```

4. Edit `functions/index.js` to have the following content:

   ```js
   
   exports.envParseTest = functions.https.onRequest((request, response) => {
     const text = process.env.ENVTEST_SAMPLE;
     const lines = text.split("\n");
     const lineCount = lines.length;
     const bytes = Array.from(Buffer.from(text));
     response.json({ text, lines, lineCount, bytes });
   });
   ```

5. Create `functions/.env` with the following content:

   ```sh
   ENVTEST_SAMPLE="foo\nbar\nbaz"
   ```

6. At this point it's sufficient to run the Cloud Functions emulator to reproduce the issue.

   ```sh
   npx firebase emulators:start --only functions
   ```

Open the indicated URL for the `envParseTest` function in a browser, or some terminal-based HTTP client.

```sh
curl 'http://localhost:5001/$GCLOUD_PROJECT/us-central1/envParseTest'
```

A response with the content

```json
{"text":"foo\nbar\nbaz","lines":["foo","bar","baz"],"lineCount":3,"bytes":[102,111,111,10,98,97,114,10,98,97,122]}
```

A response with the content

```json
{"text":"foo\nbar\\nbaz","lines":["foo","bar\\nbaz"],"lineCount":2,"bytes":[102,111,111,10,98,97,114,92,110,98,97,122]}
```

Note in the `"text"` property - the first `\n` from the `.env` has been correctly interpreted as a newline; the second remains as a literal \ followed by a literal n.  This test also outputs a `bytes` property to show the parse result unambiguously. A newline appears in the array as decimal `10`, a literal \ followed by literal n appears as decimal `92,110`.

Debug log from `firebase emulators:start` is attached.
[firebase-debug.log](https://github.com/firebase/firebase-tools/files/8210800/firebase-debug.log)

yellcorp:
### Description

Fixes #4270 and a little bit more.

The bug results from the use of `String.prototype.replace` with a first (search) argument of type `string`. This is an unintuitive corner of JavaScript that only replaces the *first* occurrence of the search string. [See MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace). To replace all occurrences, either the search argument must be expressed as an equivalent regex with the `/g` flag, or [.replaceAll](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) must be used. .replaceAll isn't supported below Node 15, and given this package's dependency on Node >= 12, the regex approach has been used here.

Quick demonstration:
```
Welcome to Node.js v16.13.2.
Type ".help" for more information.
> const aaaa = "aaaa";
undefined
> aaaa.replace("a", "b") // string argument; replaces only the first occurrence
'baaa'
> aaaa.replace(/a/, "b") // regex argument *without* /g flag; replaces only the first occurrence
'baaa'
> aaaa.replace(/a/g, "b") // regex argument *with* /g flag; replaces all occurrences
'bbbb'
> aaaa.replaceAll("a", "b") // replaceAll method in Node >= 15; replaces all occurrences
'bbbb'
```

There's another reason to use a regex approach - running the string through repeated invocations of .replace(All) processes the escapes in an implementation-defined order, whereas the string really should be scanned from start to end in order to correctly convert (for example) `\\n` to `\` `n`. The popular npm package [dotenv](https://www.npmjs.com/package/dotenv) gets tripped up by this as well.

The changes outside `.env` files result from manually searching the repo for uses of `String.prototype.replace` having a non-regex first argument. There are more occurrences than the ones fixed here that probably should be audited, but 1. Not all instances are immediately obvious to me whether they intend to replace *all* occurrences of the search argument 2. I wanted to keep this PR focused.

#### User journeys
* User wishes to deploy or emulate Cloud Functions using an .env file in which one or more values contain a quoted value having 2 or more escape sequences.

#### Edge cases
* Sequences that resemble escaped newlines, but actually aren't. For example `\\n` should be interpreted as backslash, letter n.


yellcorp:
Apologies, looks like the entire commit history has been dragged along with this! Investigating…

yellcorp:
Other String.replace usage of potential concern:
- [src/extensions/export.ts#L18-L19](https://github.com/firebase/firebase-tools/blob/e9ce0870c6264c57cda0ba621f90b857e0c33393/src/extensions/export.ts#L18-L19) - Not clear if `projectId` or `projectNumber` can/should appear more than once. If not, it's ok.
- [defaultCredentials.ts#L105](https://github.com/firebase/firebase-tools/blob/e9ce0870c6264c57cda0ba621f90b857e0c33393/src/defaultCredentials.ts#L105) - This will only replace at most one dot with an underscore

taeold:
@yellcorp This is an incredible contribution - I learned so much from your thorough writeup. Thank you so much for your time in reporting, fixing, and educating us about the issue!

It's late in US, so I'll defer reviewing the code to first thing tomorrow morning!

src/functions/runtimeConfigExport.ts
```diff
@@ -169,14 +169,19 @@ export function hydrateEnvs(pInfos: ProjectConfigInfo[], prefix: string): string
...
+
 function escape(s: string): string {
-  // Escape newlines and tabs
-  const result = s
-    .replace("\n", "\\n")
-    .replace("\r", "\\r")
-    .replace("\t", "\\t")
-    .replace("\v", "\\v");
-  return result.replace(/(['"])/g, "\\$1");
+  // Escape newlines, tabs, backslashes and quotes
```
taeold, line 183: +1 on adding backlashes to the list of things we replace!

taeold (approved):
❤️  Incredible work!

Issue #4239, opened 2022-03-02
Project-specific .env files aren't loaded when running functions emulator


larrifax:
### [REQUIRED] Environment info

**firebase-tools:** 10.2.1

**Platform:** Ubuntu

### [REQUIRED] Test case

The documentation around environment variables using .env-files while using the local functions emulator states the following:

> Contents of .env.local take precedence over .env and the project-specific .env file
> Source: https://firebase.google.com/docs/functions/config-env#emulator_support

However, when running locally the project-specific .env file isn't even loaded, only the `.env` and `.env.local` is.
And indeed, the source code doesn't even attempt to load project-specific files when running in an emulator setting:

https://github.com/firebase/firebase-tools/blob/5a8bd4e4337536821a258fa7d14cdc11d431c478/src/functions/env.ts#L199-L206

According to my interpretation of the documentation linked above, there's a mismatch between the documentation and the implementation, which should be resolved one way or another.

### [REQUIRED] Steps to reproduce

1. Add a `.env`, `.env.development` and `.env.local` file to your project, filling them with different environment variables
2. `firebase use development`
3. `firebase emulators:start --only functions`
4. Trigger a function

### [REQUIRED] Expected behavior

Functions emulator is started, and logs should show that all relevant `.env*` files are loaded:
`i  functions: Loaded environment variables from .env, .env.development, .env.local.`

### [REQUIRED] Actual behavior

Functions emulator is started, but logs shows that the project specific .env file isn't loaded (`.env.development` is missing from output):
`i  functions: Loaded environment variables from .env, .env.local.`

jaytavares:
I'm seeing this same behavior.

taeold:
Thanks for a thorough writeup. I think the documentation is right and the implementation is wrong. We will work on a fix to be shipped in the next release.

larrifax:
Thanks @taeold. This is maybe another bug/feature request, but I also noticed that .env variables aren't available when the "startup code" is ran, so it's currently not possible to e.g. store the region in an environment variable and use it like so:

`functions/index.js`:
```
import * as functions from "firebase-functions";
import * as callable from "./callables";

export const someFunction = functions.region(process.env.REGION).https.onCall(callable.someFunction);
```
The environment variables aren't available to use at the time when `.region()` or `.runWith()` are evaluated. They're only available "within `onCall`".

Is this something you're aware of/decided not to support? Or should I create a new bug/feature request for it?

pradella:
[...]

PR #4273, opened 2022-03-09
Load project-specific environment variables in the Functions Emulator

taeold:
Today, Functions Emulator does not load project-specific environment variables. This contradicts the public documentation which states:

> When using a local Cloud Functions emulator, you can override environment variables for your project by setting up a .env.local file. Contents of .env.local take precedence over .env and the project-specific .env file.

https://firebase.google.com/docs/functions/config-env#emulator_support

The patch fixes the bug by making appropriate changes to load project-specific environment variables (either in `.env.projectId` or `.env.projectAlias`) when emulating functions.

Fixes https://github.com/firebase/firebase-tools/issues/4239

### DIFF
--- a/src/functions/env.ts
+++ b/src/functions/env.ts
@@ -161,6 +161,9 @@ export function validateKey(key: string): void {
 // Parse dotenv file, but throw errors if:
 //   1. Input has any invalid lines.
 //   2. Any env key fails validation.
+/**
+ *
+ */
 export function parseStrict(data: string): Record<string, string> {
   const { envs, errors } = parse(data);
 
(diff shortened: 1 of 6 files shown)


src/functions/env.ts
```diff
@@ -198,11 +198,10 @@ function findEnvfiles(
...
-  } else {
-    files.push(`.env.${projectId}`);
-    if (projectAlias && projectAlias.length) {
-      files.push(`.env.${projectAlias}`);
-    }
+  }
+  files.push(`.env.${projectId}`);
+  if (projectAlias) {
+    files.push(`.env.${projectAlias}`);
   }
```
aheimlich, line 205: > Contents of .env.local take precedence over .env and the project-specific .env file.

With what you're doing here, `.env.<project-id-or-alias>` would override `.env.local`.

taeold: 🤦🏼‍♂️ That was sloppy of me. Thanks for the catch!

PR #4277, opened 2022-03-09
Check for admin credentials in gcloud endpoints

tonyjhuang:
### Description
Enforces admin credentials when running gcloud endpoints now as opposed to skipping auth checks.


src/emulator/storage/apis/request.ts
```diff
@@ -0,0 +1,18 @@
+import { Request } from "express";
```
tohhsinpei, line 1: Would put this file further up (maybe in `src/emulator`?) since it's not specific to Storage

src/emulator/storage/index.ts
```diff
@@ -42,13 +48,23 @@ export class StorageEmulator implements EmulatorInstance {
       getAdminCredentialValidator(),
       this._persistence
     );
+    this._adminStorageLayer = new StorageLayer(
+      args.projectId,
+      getAdminOnlyRulesValidator(),
+      getAdminCredentialValidator(),
```
tohhsinpei, line 54: Huh. I don't know about making another `StorageLayer`. If the purpose is just to validate requests a different way, why do you need to separate all the stored objects? Is there any reason you can't do what you did for `AdminCredentialValidator`?

src/emulator/storage/apis/request.ts
tonyjhuang: done

src/emulator/storage/rules/utils.ts
```diff
@@ -50,6 +50,25 @@ export function getRulesValidator(rulesetProvider: RulesetProvider): RulesValida
   };
 }
 
+/**
+ * Returns a validator that only checks for admin credentials.
```
tohhsinpei, line 54: Would be helpful to say what the difference is between `getAdminOnlyRulesValidator` and `getAdminCredentialValidator`, I think.

src/emulator/storage/index.ts
```diff
@@ -67,7 +92,8 @@ export class StorageEmulator implements EmulatorInstance {
   }
 
   reset(): void {
-    this._storageLayer.reset();
```
tohhsinpei, line 70: Not resetting the `StorageLayer`s?

tohhsinpei (commented):
Yeah, this is fine. Since we now have two `StorageLayer`s, could you add a test to make sure that they access the same objects?

abhis3 (commented):
I don't have any additional comments beside what Hsin-pei had already said, more than happy to do another review once the fixes go in

src/emulator/storage/apis/gcloud.ts
```diff
@@ -43,19 +41,16 @@ export function createCloudEndpoints(emulator: StorageEmulator): Router {
...
+          bucketId: req.params.bucketId,
+          decodedObjectId: req.params.objectId,
+        });
       } catch (err) {
         if (err instanceof NotFoundError) {
           return res.sendStatus(404);
         }
         if (err instanceof ForbiddenError) {
-          throw new Error("Request failed unexpectedly due to Firebase Rules.");
+          return res.sendStatus(403);
```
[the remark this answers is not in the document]
tonyjhuang: Previously, the only way a `ForbiddenError` could be thrown is if Firebase Rules auth failed. This represented an error state since we didn't expect Firebase Rules to be run at all when executing admin operations.

Now, a `ForbiddenError` is an acceptable error state since the `StorageLayer` can deny a request if the admin credential validator fails, so we return a 403.

src/emulator/storage/index.ts
tonyjhuang: No need, StorageLayer no longer manages its own state. Instead, we pass in the bucket and file registry state in the constructor, which we reset here.

scripts/storage-emulator-integration/tests.ts
```diff
@@ -1989,11 +1989,135 @@ describe("Storage emulator", () => {
...
+          // Wiring the app to use either the auth emulator or production auth
+          // based on the config flag.
+          const auth = firebase.auth();
+          if (!useProductionServers) {
+            auth.useEmulator(emulatorHost);
+          }
+          (window as any).auth = auth;
+        },
+        appConfig,
+        TEST_CONFIG.useProductionServers,
```
tohhsinpei, line 2036: `emulatorSpecificDescribe` means this test suite should skip if `TEST_CONFIG.useProductionServers`. You can probably get rid of the checks here and below.

scripts/storage-emulator-integration/tests.ts
```diff
@@ -1989,11 +1989,135 @@ describe("Storage emulator", () => {
...
+      }
+      if (TEST_CONFIG.useProductionServers) {
         delete process.env.GOOGLE_APPLICATION_CREDENTIALS;
+      } else {
+        delete process.env.STORAGE_EMULATOR_HOST;
+        await test.stopEmulators();
       }
     });
+
+    it.only("gcloud API persisted files should be accessible via Firebase SDK", async () => {
```
tohhsinpei, line 2093: Delete `only`, here and below

scripts/storage-emulator-integration/tests.ts
```diff
@@ -1989,11 +1989,135 @@ describe("Storage emulator", () => {
...
-        await test.stopEmulators();
+      if (TEST_CONFIG.useProductionServers) {
+        delete process.env.GOOGLE_APPLICATION_CREDENTIALS;
       } else {
+        await test.stopEmulators();
+      }
+    });
+  });
+
+  emulatorSpecificDescribe("Emulator Integrity", function (this) {
```
tohhsinpei, line 2000: Not sure I understand the test name

scripts/storage-emulator-integration/tests.ts
```diff
@@ -1989,11 +1989,135 @@ describe("Storage emulator", () => {
...
+      page = await browser.newPage();
+      await page.goto("https://example.com", { waitUntil: "networkidle2" });
+
+      await page.addScriptTag({
+        url: "https://www.gstatic.com/firebasejs/7.24.0/firebase-app.js",
+      });
+      await page.addScriptTag({
+        url: "https://www.gstatic.com/firebasejs/7.24.0/firebase-auth.js",
+      });
+      // url: "https://storage.googleapis.com/fir-tools-builds/firebase-storage-new.js",
```
tohhsinpei, line 2017: ?

scripts/storage-emulator-integration/tests.ts
tonyjhuang: Renamed to Emulator Internals, what do you think?

scripts/storage-emulator-integration/tests.ts
tonyjhuang: good question, removed here and where I copied this from

scripts/storage-emulator-integration/tests.ts
tonyjhuang: yup good call

scripts/storage-emulator-integration/tests.ts
tonyjhuang: done

tohhsinpei (approved):

abhis3 (approved):

Issue #3398, opened 2021-05-24
iOS SDK misses some metadata fields


mikehardy:
I have a little batch of ios-specific metadata issues using the storage emulator and firebase-ios-sdk (firebase-android-sdk works in these cases)

https://github.com/invertase/react-native-firebase/blob/fe3d22f54d0d706529eb5367795b568d9e783a4e/packages/storage/e2e/StorageReference.e2e.js#L373-L387

The FIXMEs here:

```javascript
  describe('updateMetadata', function () {
    it('should return the updated metadata for a file', async function () {
      const storageReference = firebase.storage().ref(WRITE_ONLY_NAME);
      const metadata = await storageReference.updateMetadata({
        contentType: 'image/jpeg',
        cacheControl: 'true',
        contentDisposition: 'disposed',
        contentEncoding: 'encoded',
        contentLanguage: 'martian',
        customMetadata: {
          hello: 'world',
        },
      });

      metadata.customMetadata.hello.should.equal('world');
      metadata.generation.should.be.a.String();
      metadata.fullPath.should.equal(WRITE_ONLY_NAME);
      metadata.name.should.equal(WRITE_ONLY_NAME);
      metadata.size.should.be.a.Number();
      should.equal(metadata.size > 0, true);
      metadata.updated.should.be.a.String();
      metadata.timeCreated.should.be.a.String();
      metadata.contentEncoding.should.be.a.String();
      metadata.contentDisposition.should.be.a.String();
      if (device.getPlatform() === 'android') {
        // FIXME on iOS this is 'application/octet-stream'?
        metadata.contentType.should.equal('image/jpeg');
      }
      metadata.bucket.should.equal(`${firebase.app().options.projectId}.appspot.com`);
      metadata.metageneration.should.be.a.String();
      metadata.md5Hash.should.be.a.String();
      if (device.getPlatform() === 'android') {
        // FIXME on iOS this comes back null ?
        should.equal(metadata.cacheControl, 'true');
      }
      if (device.getPlatform() === 'android') {
        // FIXME on iOS this comes back null ?
        should.equal(metadata.contentLanguage, 'martian');
      }
      metadata.customMetadata.should.be.an.Object();
    });

```

Also this one

```javascript
  describe('getMetadata', function () {
    it('should return a metadata for a file', async function () {
      const storageReference = firebase.storage().ref(`${PATH}/list/file1.txt`);
      const metadata = await storageReference.getMetadata();
      metadata.generation.should.be.a.String();
[...]

PR #4278, opened 2022-03-09
Set custom metadata correctly for resumable uploads

abhis3:
### Description

- Fixes a bug where resumable uploads were not honoring and setting user provided metadata on file upload. 
  - A lot of our existing unit tests use the `multipart` path in the code, hence why this case wasn't covered.

- Fixes a bug where certain metadata fields weren't present in `CloudStorageObjectMetadata`, making them revert to a default value and hence returning incorrect metadata on the final `PUT` requests.
  - GCS supports these metadata fields, see:
    - https://cloud.google.com/storage/docs/metadata#content-disposition
    - https://cloud.google.com/storage/docs/metadata#content-encoding

- Fixes a bug where GCS metadataUpdate cloud functions were triggered in incorrect situations

Part of fixes to [#3398](https://github.com/firebase/firebase-tools/issues/3398)

### Scenarios Tested

Tested functionality on a sample iOS app, code here: [https://github.com/abhis3/Storage-Emulator-Issue/blob/main/Storage%20Emulator%20Issue/StorageManager.swift](https://github.com/abhis3/Storage-Emulator-Issue/blob/main/Storage%20Emulator%20Issue/StorageManager.swift)

Testing against prod, we see the expected behavior so it's definitely an emulator issue.

Prior to the change, we see that the set metadata looked something like this (using the above app):
```
FIRStorageMetadata 0x600000b5ca00: {
    bucket = "ios-sdk-issues.appspot.com";
    contentDisposition = inline;
    contentEncoding = identity;
    contentType = "";
    generation = 1646779748099;
    md5Hash = "i91E7z6la2nMpxdarYKjrw==";
    metageneration = 1;
    name = "Profiles/mh1MOpqFgaMI9Iq7kTB72QmsFbYH.jpg";
    size = 31769;
    timeCreated = "2022-03-08T14:49:08.099Z";
    updated = "2022-03-08T14:49:08.099Z";
}
```

After the changes, metadata looks similar to:
```
FIRStorageMetadata 0x600000b5ca00: {
    bucket = "ios-sdk-issues.appspot.com";
    contentDisposition = initialCommit;
    contentEncoding = identity;
    contentType = "image/jpg";
    generation = 1646779748099;
    md5Hash = "i91E7z6la2nMpxdarYKjrw==";
    metageneration = 1;
    name = "Profiles/mh1MOpqFgaMI9Iq7kTB72QmsFbYH.jpg";
    size = 31769;
    timeCreated = "2022-03-08T14:49:08.099Z";
    updated = "2022-03-08T14:49:08.099Z";
}
```

Note the changes in the `contentDisposition` and `contentType` fields.

### DIFF
--- a/scripts/integration-helpers/framework.ts
+++ b/scripts/integration-helpers/framework.ts
@@ -320,10 +320,6 @@ export class TriggerEndToEndTest {
     return this.invokeHttpFunction("writeToDefaultStorage");
   }
 
-  abhisunTest(): Promise<Response> {
-    return this.invokeHttpFunction("abhisunTest");
-  }
-
   writeToSpecificStorageBucket(): Promise<Response> {
     return this.invokeHttpFunction("writeToSpecificStorageBucket");
   }
--- a/scripts/integration-helpers/framework.ts
+++ b/scripts/integration-helpers/framework.ts
@@ -320,6 +320,10 @@ export class TriggerEndToEndTest {
     return this.invokeHttpFunction("writeToDefaultStorage");
   }
 
+  abhisunTest(): Promise<Response> {
+    return this.invokeHttpFunction("abhisunTest");
+  }
+
   writeToSpecificStorageBucket(): Promise<Response> {
     return this.invokeHttpFunction("writeToSpecificStorageBucket");
   }
--- a/scripts/integration-helpers/framework.ts
+++ b/scripts/integration-helpers/framework.ts
@@ -187,9 +187,6 @@ export class TriggerEndToEndTest {
         this.storageFinalizedTriggerCount++;
       }
       if (data.includes(STORAGE_FUNCTION_METADATA_LOG)) {
-        console.log("storage md trigger");
-        console.log(data);
-        console.log("---");
         this.storageMetadataTriggerCount++;
       }
       if (data.includes(STORAGE_BUCKET_FUNCTION_ARCHIVED_LOG)) {
@@ -218,9 +215,6 @@ export class TriggerEndToEndTest {
         this.storageV2FinalizedTriggerCount++;
       }
       if (data.includes(STORAGE_FUNCTION_V2_METADATA_LOG)) {
-        console.log("storage md trigger v2");
-        console.log(data);
-        console.log("---");
         this.storageV2MetadataTriggerCount++;
       }
       if (data.includes(STORAGE_BUCKET_FUNCTION_V2_ARCHIVED_LOG)) {
--- a/scripts/integration-helpers/framework.ts
+++ b/scripts/integration-helpers/framework.ts
@@ -187,6 +187,9 @@ export class TriggerEndToEndTest {
         this.storageFinalizedTriggerCount++;
       }
       if (data.includes(STORAGE_FUNCTION_METADATA_LOG)) {
+        console.log("storage md trigger");
+        console.log(data);
+        console.log("---");
         this.storageMetadataTriggerCount++;
       }
       if (data.includes(STORAGE_BUCKET_FUNCTION_ARCHIVED_LOG)) {
@@ -215,6 +218,9 @@ export class TriggerEndToEndTest {
         this.storageV2FinalizedTriggerCount++;
       }
       if (data.includes(STORAGE_FUNCTION_V2_METADATA_LOG)) {
+        console.log("storage md trigger v2");
+        console.log(data);
+        console.log("---");
         this.storageV2MetadataTriggerCount++;
       }
       if (data.includes(STORAGE_BUCKET_FUNCTION_V2_ARCHIVED_LOG)) {
--- a/scripts/storage-emulator-integration/tests.ts
+++ b/scripts/storage-emulator-integration/tests.ts
@@ -664,6 +664,7 @@ describe("Storage emulator", () => {
             bucket: "string",
             contentType: "string",
             contentDisposition: "string",
+            contentEncoding: "string",
             generation: "string",
             md5Hash: "string",
             crc32c: "string",
--- a/scripts/storage-emulator-integration/tests.ts
+++ b/scripts/storage-emulator-integration/tests.ts
@@ -884,10 +884,13 @@ describe("Storage emulator", () => {
             }
           }
 
+          expect(metadata.name).to.equal("small_file");
+          expect(metadata.contentType).to.equal("application/octet-stream");
           expect(metadataTypes).to.deep.equal({
             bucket: "string",
-            contentType: "string",
             contentDisposition: "string",
+            contentEncoding: "string",
+            contentType: "string",
             generation: "string",
             md5Hash: "string",
             crc32c: "string",
@@ -905,6 +908,34 @@ describe("Storage emulator", () => {
             selfLink: "string",
             timeStorageClassUpdated: "string",
           });
+        });
+
+        it("should return generated custom metadata for new upload", async () => {
+          const customMetadata = {
+            contentDisposition: "initialCommit",
+            contentType: "image/jpg",
+            name: "test_upload.jpg",
+          };
+
+          const uploadURL = await supertest(STORAGE_EMULATOR_HOST)
+            .post(
+              `/upload/storage/v1/b/${storageBucket}/o?name=test_upload.jpg&uploadType=resumable`
+            )
+            .send(customMetadata)
+            .set({
+              Authorization: "Bearer owner",
+            })
+            .expect(200)
+            .then((res) => new URL(res.header["location"]));
+
+          const returnedMetadata = await supertest(STORAGE_EMULATOR_HOST)
+            .put(uploadURL.pathname + uploadURL.search)
+            .expect(200)
+            .then((res) => res.body);
+
+          expect(returnedMetadata.name).to.equal(customMetadata.name);
+          expect(returnedMetadata.contentType).to.equal(customMetadata.contentType);
+          expect(returnedMetadata.contentDisposition).to.equal(customMetadata.contentDisposition);
         });
 
         it("should return a functional media link", async () => {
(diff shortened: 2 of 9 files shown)


scripts/storage-emulator-integration/tests.ts
```diff
@@ -756,6 +759,46 @@ describe("Storage emulator", () => {
       expect(uploadState).to.equal("success");
     });
 
+    it("should should set metadata correctly on multipart uploads", async () => {
```
tonyjhuang, line 762: replace "multipart" -> "resumable"

scripts/storage-emulator-integration/tests.ts
```diff
@@ -756,6 +759,46 @@ describe("Storage emulator", () => {
...
+        .expect(200)
+        .then((res) => new URL(res.header["x-goog-upload-url"]));
+
+      await supertest(STORAGE_EMULATOR_HOST)
+        .put(uploadURL.pathname + uploadURL.search)
+        .set({
+          "X-Goog-Upload-Protocol": "resumable",
+          "X-Goog-Upload-Command": "upload",
+        })
+        .expect(200);
```
tonyjhuang, line 786: do you need this request?

src/emulator/storage/files.ts
```diff
@@ -335,23 +335,29 @@ export class StorageLayer {
...
    * @returns The stored file.
    */
   public finalizeUpload(upload: ResumableUpload): StoredFile {
     upload.status = UploadStatus.FINISHED;
 
-    const metadata = this.createMetadata(upload);
+    const storedMetadata = this.createMetadata(upload);
+    const incomingMetadata = upload.metadata;
+    storedMetadata.update(incomingMetadata);
+    if (storedMetadata.downloadTokens.length === 0) {
```
tonyjhuang, line 348: `finalizeUpload` is also used by `gcloud.ts` which should not generate a new download token on upload. I would move this back to `firebase.ts`

tonyjhuang:
synced offline, do you mind rebasing off of https://github.com/firebase/firebase-tools/pull/4262 and pulling the `md` -> `metadata` rename to a separate PR?

src/emulator/storage/files.ts
abhis3: Good catch, I'll move the token generation back. I think the `storedMetadata.update(incomingMetadata);` should be fine to keep as `gcloud.ts` and `firebase.ts` both take the metadata response and return it

scripts/storage-emulator-integration/tests.ts
abhis3: Done

scripts/storage-emulator-integration/tests.ts
abhis3: Not needed, I added it when I was testing a few things but I'll remove it.

src/emulator/storage/files.ts
abhis3: Scratch what I said, didn't end up doing what I suggested it after looking at the rebased code again

src/emulator/storage/apis/firebase.ts
```diff
@@ -353,8 +354,10 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
         }
         throw err;
       }
-      metadata.addDownloadToken();
-      return res.status(200).json(new OutgoingFirebaseMetadata(metadata));
+
+      storedMetadata.update(upload.metadata);
```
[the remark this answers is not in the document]
abhis3: Fixed

scripts/triggers-end-to-end-tests/tests.ts
```diff
@@ -274,8 +274,8 @@ describe("storage emulator function triggers", () => {
     // default bucket
     expect(test.storageFinalizedTriggerCount).to.equal(1);
     expect(test.storageV2FinalizedTriggerCount).to.equal(1);
-    expect(test.storageMetadataTriggerCount).to.equal(0);
-    expect(test.storageV2MetadataTriggerCount).to.equal(0);
+    expect(test.storageMetadataTriggerCount).to.equal(1);
+    expect(test.storageV2MetadataTriggerCount).to.equal(1);
```
abhis3, line 278: Note for reviewer: Not sure how to run these locally/values seem to change between runs? Will investigate more on Monday but not sure if we even consider these tests before pushing?

scripts/triggers-end-to-end-tests/tests.ts
```diff
@@ -385,8 +385,8 @@ describe("storage emulator function triggers", () => {
     // default bucket
     expect(test.storageFinalizedTriggerCount).to.equal(1);
     expect(test.storageV2FinalizedTriggerCount).to.equal(1);
-    expect(test.storageMetadataTriggerCount).to.equal(0);
-    expect(test.storageV2MetadataTriggerCount).to.equal(0);
+    expect(test.storageMetadataTriggerCount).to.equal(1);
```
tonyjhuang, line 388: sorry why are these increasing here?

scripts/triggers-end-to-end-tests/tests.ts
```diff
@@ -274,8 +274,8 @@ describe("storage emulator function triggers", () => {
...
```
tonyjhuang, line 277: I don't think this is an intended behavior. object upload should only ever emit a finalize trigger, not a metadata trigger

tonyjhuang: same comment through this file

abhis3: Since we do a `metadata.update` in `files.ts` I believe this will now fire for all uploads.

I'm looking at [this code ](https://github.com/firebase/firebase-tools/blob/5564ad1836ca91010e4b99ff78ef70f8f44bdc89/scripts/triggers-end-to-end-tests/functions/index.js#L261) and the [GCS Object Builder docs](https://firebase.google.com/docs/reference/functions/providers_storage.objectbuilder#onmetadataupdate)

Based off of those I think it's expected that these counts should be incremented?

abhis3: Previously the metadata was never updated with whatever was incoming, but now after this change it will be. Since this now counts as a GCS metadata update, I think it makes sense to me that these counts should be incremented?

scripts/integration-helpers/framework.ts
```diff
@@ -214,6 +217,9 @@ export class TriggerEndToEndTest {
         this.storageV2FinalizedTriggerCount++;
       }
       if (data.includes(STORAGE_FUNCTION_V2_METADATA_LOG)) {
+        console.log("storage md trigger v2");
+        console.log(data);
+        console.log("---");
```
abhis3, line 222: Leaving these and other debug statements here for now while we come to a consensus on what the integ tests trigger values should be, I'll remove them all before a merge

scripts/triggers-end-to-end-tests/tests.ts
tonyjhuang: I agree that looking at the code, that's what we would expect to happen but this doesn't conform to the actual spec of GCS storage triggers. Metadata update triggers should only fire on changing the metadata of an *existing* object. 

See https://cloud.google.com/functions/docs/calling/storage#metadata

abhis3: Synced offline, for now I'll add a boolean param to fix this. Left a TODO to refactor the logic in the longer term.

src/emulator/storage/apis/firebase.ts
```diff
@@ -356,8 +357,9 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
         }
         throw err;
       }
-      metadata.addDownloadToken();
-      return res.status(200).json(new OutgoingFirebaseMetadata(metadata));
+
+      storedMetadata.addDownloadToken();
```
tonyjhuang, line 361: I'm surprised this isn't failing an integration test somewhere since `addDownloadToken` calls `update({})` internally. Can you make sure we cover this in a test case that this doesn't trigger a metadata update cloud function as well

abhis3: Synced offline. It will take some work to get tests working that also account for the number of times different cloud functions are called specifically from Storage. I've added a TODO to capture the fact we need to test for trigger counts in these tests as well.

In the meanwhile I've amended the PR to pass in the flag in the correct spots.

scripts/storage-emulator-integration/tests.ts
```diff
@@ -1109,6 +1143,9 @@ describe("Storage emulator", () => {
     });
   });
 
+  /**
+   * TODO(abhisun): Add test coverage to validate how many times various cloud functions are triggered.
```
tonyjhuang, line 1147: We probably want to do this in the triggers end to end test file

src/emulator/storage/apis/firebase.ts
```diff
@@ -352,8 +353,9 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
```
tonyjhuang, line 357: Just peeked at our server implementation; interestingly enough, we only generate the download token on metadata get, not object get (data fetch). The metadata get handler actually updates GCS with the new download token, so I have a hunch the server will actually trigger any metadata update functions that are set up.

The rest of this PR looks good to me but can you verify the behavior of the server and mimic it here? I just want to make sure we limit the amount of divergence between the emulator and the server.

abhis3: Yup that makes sense thanks for the callout. Here's all the places I was able to find that we add download tokens:

- Metadata Get ([Link](https://source.corp.google.com/piper///depot/google3/java/com/google/firebase/storage/requesthandlers/MetadataGetHandler.java;l=47;rcl=436357765))
  - We generate a new download token if one isn't present when we fetch metadata
- Object Creation ([Link](https://source.corp.google.com/piper///depot/google3/java/com/google/firebase/storage/requesthandlers/ObjectCreateHandler.java;l=200;bpv=1;bpt=1;rcl=436357765))
- If we delete a download token and it was the last one, we replace it with a fresh token ([Link](https://source.corp.google.com/piper///depot/google3/java/com/google/firebase/storage/requesthandlers/TokenDeleteHandler.java;l=52;rcl=436357765))

tonyjhuang (approved):

PR #4280, opened 2022-03-09
Use swc for ts-node

taeold:
Recent version of `ts-node` provides an option to use `swc` to compile typescript source. As expected, using `swc` improves time to compile TS code and hence the overall time it takes to run our tests written in TS.

Before:

Screen Shot 2022-03-09 at 11 50 51 AM

After:

Screen Shot 2022-03-09 at 11 49 59 AM

Not the biggest difference, but it's not hard to setup + it will hugely benefit some changes I expect to make to allow Functions Emulator integration test to run without a test harness by directly running `ts-node` in spawned worker processes (WIP: https://github.com/firebase/firebase-tools/pull/4272)

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -75,6 +75,7 @@
       "devDependencies": {
         "@google/events": "^5.1.1",
         "@manifoldco/swagger-to-ts": "^2.0.0",
+        "@swc/core": "^1.2.151",
         "@types/archiver": "^5.1.0",
         "@types/body-parser": "^1.17.0",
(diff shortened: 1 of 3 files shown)


taeold:
I didn't realize specifying `swc` as tranpiler for `ts-node` disables type-checking. This probably is useful for lots of scripts that directly invokes `ts-node` or registers it in their mocha run since I notice that many of these scripts don't have separate mechanism for type-checking (as recommended in [`ts-node` doc](https://github.com/firebase/firebase-tools/blob/master/package.json#L25)).

PR #4287, opened 2022-03-10
Adding secret support for Extensions emulator

joehan:
### Description
Adding (partial) secrets support for Extensions emulator. This work turned out to be bigger than expected, so I'm trying to split it up a bit - this PR implements the logic to discover secrets in the Extension spec, transform them into SecretEnvVar, and pipe them through to the functions emulator. However, it still only checks `secret.local` for overrides - I'll add support for `{instanceId}.secret.local` in a follow up PR.

### Scenarios Tested
I tested this by emulating the firestore-stripe-payments extension, and I was able to access a test API key & have createCustomer successfully call Stripe with that key. I validated that a new customer object was created in stripe
Screen Shot 2022-03-10 at 2 19 04 PM

### DIFF
--- a/src/emulator/functionsEmulator.ts
+++ b/src/emulator/functionsEmulator.ts
@@ -64,7 +64,6 @@ import * as backend from "../deploy/functions/backend";
 import * as functionsEnv from "../functions/env";
 import { flattenArray } from "../functional";
 import { SecretEnvVar } from "../gcp/cloudfunctions";
-import { ENV_DIRECTORY } from "../extensions/manifest";
 
 const EVENT_INVOKE = "functions:invoke";
 const LOCAL_SECRETS_FILE = ".secret.local";
@@ -123,7 +122,6 @@ export interface FunctionsEmulatorArgs {
   debugPort?: number;
   remoteEmulators?: { [key: string]: EmulatorInfo };
   adminSdkConfig?: AdminSdkConfig;
-  projectAlias?: string;
 }
 
 // FunctionsRuntimeInstance is the handler for a running function invocation
@@ -957,7 +955,6 @@ export class FunctionsEmulator implements EmulatorInstance {
     const projectInfo = {
       functionsSource: backend.functionsDir,
       projectId: this.args.projectId,
-      projectAlias: this.args.projectAlias,
       isEmulator: true,
     };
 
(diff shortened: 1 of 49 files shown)


src/emulator/functionsEmulator.ts
```diff
@@ -1082,11 +1093,15 @@ export class FunctionsEmulator implements EmulatorInstance {
     if (trigger) {
       const secrets: backend.SecretEnvVar[] = trigger.secretEnvironmentVariables || [];
       const accesses = secrets
-        .filter((s) => !secretEnvs[s.secret])
+        .filter((s) => !secretEnvs[s.key])
```
joehan, line 1096: Made these changes after discussing with Daniel. In CF3 world, secret and key are the same always, so the old code worked. However, based on what the fields of secretEnvVar represent, we  had these backwards before.

src/extensions/emulator/optionsHelper.ts
```diff
@@ -47,26 +48,75 @@ export async function buildOptions(options: any): Promise<any> {
...
+    if (paramValues[s.param]) {
+      const [, projectId, , secret, , version] = paramValues[s.param].split("/");
+      secretEnvVar.push({
+        key: s.param,
+        secret,
+        projectId,
+        version,
+      });
+    }
+    // TODO: Throw an error if a required secret is missing?
```
joehan, line 115: Holding off on this for now, will add it in the follow up PR.

src/extensions/extensionsApi.ts
```diff
@@ -149,7 +150,7 @@ export interface Param {
   label: string;
   description?: string;
   default?: string;
-  type?: ParamType;
+  type?: ParamType | SpecParamType;
```
joehan, line 153: This is a bit strange, but we use Param all over the place, both in the context of a spec that we read from an extension.yaml file and one that was returned from the backend. In the first case this is SpecParamType, in the second this is ParamType.

elvisun (approved):
Sorry about the delay!! somehow this fell through my inbox

src/extensions/extensionsApi.ts
joehan: Ahhh, the actual issue here is that SpecParamType (the input the backend expects) is CamelCase, but ParamType (the output the backend returns) is ALLCAPS. This makes enum comparisons super awkward and buggy. I filed b/224618262 to fix this, and added a TODO.

src/emulator/functionsEmulator.ts
```diff
@@ -99,7 +102,7 @@ export interface EmulatableBackend {
  */
 export interface BackendInfo {
   directory: string;
-  env: Record<string, string>;
+  env: Record<string, string>; // TODO: Consider exposing more information about where param values come from & if they are locally overwritten.
```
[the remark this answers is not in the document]
joehan: Haha, more like you solved my problems for me :)

src/emulator/functionsEmulator.ts
```diff
@@ -833,9 +839,14 @@ export class FunctionsEmulator implements EmulatorInstance {
       .filter((t) => !t.backend.extensionInstanceId)
       .map((t) => t.def);
     return this.args.emulatableBackends.map((e: EmulatableBackend) => {
+      const envWithSecrets = Object.assign({}, e.env);
```
inlined, line 842: Don't we typically use the splat operator now instead of `Object.assign`? Silly nit but could just have said `{ ...e.env }`

joehan: Ack - I'll slip this into a future PR since this already went in.

PR #4288, opened 2022-03-11
Fix broken unit test for resumable uploads

abhis3:
### Description

Fix a broken unit test for resumable uploads. These have been an issue point for a few recent bugs, patching to improve test coverage.

### DIFF
--- a/scripts/storage-emulator-integration/tests.ts
+++ b/scripts/storage-emulator-integration/tests.ts
@@ -264,11 +264,9 @@ describe("Storage emulator", () => {
         });
 
         it("should handle large (resumable) uploads", async () => {
-          const fileName = "test_upload.jpg"
+          const fileName = "test_upload.jpg";
           const uploadURL = await supertest(STORAGE_EMULATOR_HOST)
-            .post(
-              `/upload/storage/v1/b/${storageBucket}/o?name=${fileName}&uploadType=resumable`
-            )
+            .post(`/upload/storage/v1/b/${storageBucket}/o?name=${fileName}&uploadType=resumable`)
             .send({})
             .set({
               Authorization: "Bearer owner",
--- a/scripts/storage-emulator-integration/tests.ts
+++ b/scripts/storage-emulator-integration/tests.ts
@@ -268,7 +268,9 @@ describe("Storage emulator", () => {
           const uploadURL = await supertest(STORAGE_EMULATOR_HOST)
             .post(`/upload/storage/v1/b/${storageBucket}/o?name=${fileName}&uploadType=resumable`)
             .send({})
-            .set({Authorization: "Bearer owner",})
+            .set({
+              Authorization: "Bearer owner",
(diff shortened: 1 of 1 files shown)


scripts/storage-emulator-integration/tests.ts
```diff
@@ -263,9 +263,51 @@ describe("Storage emulator", () => {
           });
         });
 
-        // TODO(abehaskins): This test is temporarily disabled due to a credentials issue
-        it.skip("should handle large (resumable) uploads", async () => {
-          await testBucket.upload(largeFilePath, { resumable: true });
+        it.only("should handle large (resumable) uploads", async () => {
```
tohhsinpei, line 266: Delete `only`

scripts/storage-emulator-integration/tests.ts
```diff
@@ -263,9 +263,51 @@ describe("Storage emulator", () => {
...
+          const fileName = "test_upload.jpg";
+          const uploadURL = await supertest(STORAGE_EMULATOR_HOST)
+            .post(`/upload/storage/v1/b/${storageBucket}/o?name=${fileName}&uploadType=resumable`)
+            .send({})
```
tohhsinpei, line 270: Do you need this `send` then?

scripts/storage-emulator-integration/tests.ts
```diff
@@ -263,9 +263,51 @@ describe("Storage emulator", () => {
...
+            .set({
+              Authorization: "Bearer owner",
+            })
```
tohhsinpei, line 273: Nit: one line

scripts/storage-emulator-integration/tests.ts
tohhsinpei: By the way, what does this unit test have to do with large uploads?

scripts/storage-emulator-integration/tests.ts
abhis3: Done

scripts/storage-emulator-integration/tests.ts
abhis3: Good call, deleted only

scripts/storage-emulator-integration/tests.ts
abhis3: Yup, without it we throw a `400 "Bad Request"`

scripts/storage-emulator-integration/tests.ts
abhis3: Nothing in particular with large uploads, this is more to test for resumable functionality. I'll remove the reference to a large file to simplify this.

scripts/storage-emulator-integration/tests.ts
abhis3: Actually prettier complains if it's not 2 separate lines, reverting to what it was originally

PR #4292, opened 2022-03-14
Refactor getParam to allow multiple configurations

elvisun:
### Description
This is the first step of supporting local secrets in the manifest when using the `--local` flags in ext:* commands.
1. This PR: **Refactor getParam logic by introducing a new `ParamBindingOptions` interface. This will encapsulate different values (default, local, project-specific, etc) for a param in one obj. So rather than passing around a `Record<string, string>`, we pass all binding options in a `Record<string, ParamBindingOptions>` instead. Then in downstream logic we can choose how we want to write them to the manifest.**
2. Ask for a local value when user is configuring a secret
3. Write the local value to manifest (a `.secret.local` file).

#### Alternatives considered
- Using special-case prefix like `LOCAL-VALUE-{env-ver-key}` when collecting user inputs, then we add special handling when using the values. This is the least amount of work, but it feels pretty hacky. Considering the manifest supports `.local`, `.project-id`, `.alias` etc, it feels most natural to just refactor this to use a more general approach rather than adding more tech-debt that we'll likely need to cleanup anyways

### Scenarios Tested

Manually tested install, update, configure

### DIFF
--- a/src/extensions/paramHelper.ts
+++ b/src/extensions/paramHelper.ts
@@ -38,6 +38,17 @@ export function getDefaultParamBindings(params: { [key: string]: ParamBindingOpt
   return ret;
 }
 
+export function buildBindingOptionsWithDefault(defaultParams: { [key: string]: string }) : { [key: string]: ParamBindingOptions }{
+  let paramOptions : { [key: string]: ParamBindingOptions } = {};
+  Object.entries(defaultParams).forEach(([k,v]) => {
+    paramOptions = {
+      ...paramOptions,
+      ...{[k]: {default: v}},
+    }
+  });
+  return paramOptions;
+}
+
 /**
  * A mutator to switch the defaults for a list of params to new ones.
  * For convenience, this also returns the params
@@ -128,7 +139,7 @@ export async function getParamsForUpdate(args: {
   nonInteractive?: boolean;
   instanceId: string;
 }): Promise<{ [key: string]: ParamBindingOptions }> {
-  let params: any;
+  let params: { [key: string]: ParamBindingOptions };
   if (args.nonInteractive && !args.paramsEnvPath) {
     const paramsMessage = args.newSpec.params
       .map((p) => {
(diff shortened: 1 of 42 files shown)


src/extensions/paramHelper.ts
```diff
@@ -16,6 +16,41 @@ import * as askUserForParam from "./askUserForParam";
 import * as track from "../track";
 import * as env from "../functions/env";
 
+/**
+ * Interface for holding different param values for different environments/configs.
+ */
+export interface ParamBindingOptions {
+  default: string;
```
joehan, line 23: Very much like this ParamBindingOptions approach overall. One nit is that I'd prefer to use a different word than 'default' to represent the value provided in `instanceId.env` - default already has a different meaning in the context of params (the Param.default value). Consider something like `baseValue` instead?

src/extensions/paramHelper.ts
```diff
@@ -16,6 +16,41 @@ import * as askUserForParam from "./askUserForParam";
...
+  local?: string;
+  // Add project specific key:value here when we want to support that.
+}
+
+export function getDefaultParamBindings(params: { [key: string]: ParamBindingOptions }): {
```
joehan, line 28: Same comment as above - this sounds like it returns the default param values defined in extension.yaml

src/extensions/paramHelper.ts
```diff
@@ -16,6 +16,41 @@ import * as askUserForParam from "./askUserForParam";
...
+  [key: string]: string;
+} {
+  let ret = {};
+  Object.entries(params).forEach(([k, v]) => {
```
joehan, line 32: Prefer `for... of...` instead of `forEach()`  - [go/tsjs-practices/iteration](https://goto.google.com/tsjs-practices/iteration)

joehan (changes requested):
Overall approach LGTM, but I think we're overloading the word 'default' here in a pretty confusing way.

src/extensions/paramHelper.ts
```diff
@@ -16,6 +16,44 @@ import * as askUserForParam from "./askUserForParam";
 import * as track from "../track";
 import * as env from "../functions/env";
 
+/**
+ * Interface for holding different param values for different environments/configs.
+ *
+ * baseValue: The base value of the configurations.
```
joehan, line 22: Nit - include in this comment what file each of these is read from

PR #4297, opened 2022-03-14
Temporarily skip Storage Rules Manager unit tests

tohhsinpei:
### Description
This PR temporarily skips the Storage Rules Manager unit tests added in #4245. Since the test setup downloads the rules JAR, we should make it an integration test instead. We'll do this in another PR; skipping for now so as to unblock the new release.


bkendall (approved):
I'm not bothering with the typo since this should be looked at sooner rather than later :)

PR #4300, opened 2022-03-15
Read {instanceId}.secret.local for Extension secret overrides

joehan:
### Description
Adds support for overriding Extensions secret by including them in `extensions/{instanceId}.secret.local`

TODO: Add unit tests to this

### Scenarios Tested
I validated that I could provide secrets for the Stripe extension via `{instanceId}.secret.local`, on both a real and demo project. I also confirmed that I could still fall back to accessing secrets from Secret manager when using a real project.

### DIFF
--- a/src/emulator/functionsEmulator.ts
+++ b/src/emulator/functionsEmulator.ts
@@ -64,6 +64,7 @@ import * as backend from "../deploy/functions/backend";
 import * as functionsEnv from "../functions/env";
 import { flattenArray } from "../functional";
 import { SecretEnvVar } from "../gcp/cloudfunctions";
+import { ENV_DIRECTORY } from "../extensions/manifest";
 
 const EVENT_INVOKE = "functions:invoke";
 const LOCAL_SECRETS_FILE = ".secret.local";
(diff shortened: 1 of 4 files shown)


src/emulator/functionsEmulator.ts
```diff
@@ -1077,16 +1078,23 @@ export class FunctionsEmulator implements EmulatorInstance {
...
     let secretEnvs: Record<string, string> = {};
-
+    // Extension secret overrides live in {instanceID}.secret.local, CF3 secrets live in .secret.local
+    const secretsFile = backend.extensionInstanceId
+      ? `${backend.extensionInstanceId}${LOCAL_SECRETS_FILE}`
+      : LOCAL_SECRETS_FILE;
+    const secretDirectory = backend.extensionInstanceId
+      ? path.join(this.args.projectDir, ENV_DIRECTORY)
+      : backend.functionsDir;
+    const secretPath = path.join(secretDirectory, secretsFile);
```
elvisun, line 1088: seems like a good candidate for a function, then you can unit test it

elvisun (approved):

src/emulator/functionsEmulator.ts
joehan: Good call - separated out and unit tested

PR #4303, opened 2022-03-15
Ask for local secret param when collecting user inputs

elvisun:
### Description
Ask for local secret param, then write it to extensions manifest.

## When creating a secret, user will be asked how they want to handle the secret.
Screen Shot 2022-03-16 at 10 38 25 AM

## New prompt for entering local secret.
Screen Shot 2022-03-16 at 10 38 47 AM

## Install flow for only using secret manager
Screen Shot 2022-03-16 at 10 37 58 AM

## Install flow for using both secret manager and local
Screen Shot 2022-03-16 at 10 38 58 AM

## Reconfigure
Screen Shot 2022-03-16 at 11 18 31 AM

#### Alternatives Considered
- Asking a yes or no questions on configuring a local secret after setting a cloud secret. This doesn't work because users might want to **only** configure a local secret.

### Scenarios Tested

Manually tested install configure update


joehan:
Gonna start reviewing this now - however, it looks like the git history is funky, probably because this includes all the individual commits from #4292 that got squashed into master. If it doesn't create too much of a headache, could you rebase this onto master?

src/extensions/askUserForParam.ts
```diff
@@ -150,16 +157,42 @@ export async function askForParam(args: {
...
+              },
+              {
+                checked: false,
+                name: "Local file (Only used by Firebase Emulator)",
+                value: SecretCreationAction.LOCAL.toString(),
+              },
+            ],
+          });
+        }
+        if (secretLocations.includes(SecretCreationAction.CLOUD.toString())) {
```
joehan, line 187: Optional: This chunk from l187 to l193 could be good to separate into its own function, for readability.

src/deploy/extensions/planner.ts
```diff
@@ -6,7 +6,30 @@ import { FirebaseError } from "../../error";
...
+/**
+ * Instance spec used by manifest.
+ *
+ * Params are passed in ParamBindingOptions so we know the param bindings for
+ * all environments user has configured.
+ *
+ * So far this is only used for writing to the manifest, but in the future
+ * we want to read manifest into this interface.
+ */
+export interface ManifestInstanceSpec {
```
joehan, line 20: I think creating this type split is reasonable for now, but I'd prefer to join these into a single type in the future. They have a ton of overlap (and InstanceSpec has some useful caching functions to fetch extension & extensionVersion), and the only difference is params, so I'd rather we move the type split down to just params. Maybe something like:
```
export type ParamValue = string | ParamBindingOptions
export function resolveParamValues(p: ParamValue) {
// Either resolves the ParamBindingOption or returns the string value
}
```

src/extensions/manifest.ts
```diff
@@ -56,6 +57,44 @@ export async function writeToManifest(
...
+  config: Config,
+  force?: boolean
+): Promise<void> {
+  for (const spec of specs) {
+    if (!spec.paramSpecs) {
+      continue;
+    }
+
+    let writeBuffer: Record<string, string> = {};
+    for (const paramSpec of spec.paramSpecs.filter((p) => p.type === ParamType.SECRET)) {
```
joehan, line 75: Nit: I think this code ends up cleaner if we also filter for only params that have a local override- ie:
```
const locallyOverridenSecretParams = spec.paramsSpecs.filter((p) => p.type === ParamType.Secret && spec.params[p.param].local)
```
```

File: src/extensions/askUserForParam.ts [hunk 4]
```diff
@@ -6,15 +6,20 @@ const { marked } = require("marked");
...
-import { logPrefix, substituteParams } from "./extensionsHelper";
+import { confirm, logPrefix, substituteParams } from "./extensionsHelper";
 import { convertExtensionOptionToLabeledList, getRandomString, onceWithJoin } from "./utils";
 import { logger } from "../logger";
 import { promptOnce } from "../prompt";
 import * as utils from "../utils";
 import { ParamBindingOptions } from "./paramHelper";
 
+enum SecretCreationAction {
+  CLOUD = 1,
```
joehan, line 17: I think i might just be not understanding some nuance of TS enums, but why did you do this `CLOUD = 1, LOCAL,`? Does the `=1` set a default or something?

joehan (commented):
Took a pass through with some comments - will come back once tests are passing too

src/deploy/extensions/planner.ts
elvisun: I explored in that direction but not exactly a union type, I decided to not touch `InstanceSpec` because it's used in too many places, and it'd be hard to untangle..

The reason why I ended up introducing another type is because I thought conceptually it made sense. `ManifestInstanceSpec` is the bare minimum spec that we pull from the manifest. But after hydration with ExtensionVersion, Extension, and collapsing the params , it turns into `DeploymentInstanceSpec`, something we use for deployment/emulator. So in the long term I'm leaning towards keeping both types but maybe make them more clearly separated? wdyt?

src/extensions/manifest.ts
elvisun: good suggestion, done

src/extensions/askUserForParam.ts
elvisun: it's just always better to start enum at 1 :)

Don't think this made it into the ts style but we have a firebase style guide that recommended this:
go/fireconsole-styleguide#always-start-enums-with-1

elvisun:
> Gonna start reviewing this now - however, it looks like the git history is funky, probably because this includes all the individual commits from #4292 that got squashed into master. If it doesn't create too much of a headache, could you rebase this onto master?

I tried to rebase but didn't look like a good idea... I'll leave this as is so we don't risk missing this for the bug bash. Since we are going to squash merge I think it's ok?

Fixed the tests and addressed the comments. I'll come back after dinner to add some unit tests and do some more manual testing and polishing

joehan (approved):
This LGTM as is - we should add some unit test for writeLocalSecrets, but I think it is OK to merge this in time for the bug bash and add them afterwards. No need to work late into the night for this one :)

elvisun:
> This LGTM as is - we should add some unit test for writeLocalSecrets, but I think it is OK to merge this in time for the bug bash and add them afterwards. No need to work late into the night for this one :)

Thanks for the review! no worries, wanted to do another pass before the bugbash :) Just added some unit tests, going to merge this so I can send you some other small fixes

PR #4309, opened 2022-03-16
Moved Storage rules integration tests under scripts/

tohhsinpei:
### Description
Follow up to #4297 (and #3233). Previously, we disabled the Storage rules unit tests because they were failing on deploy. This is because our deploy pipeline doesn't support `java`, which is needed to run the rules JAR. However, unit tests should not be downloading and spinning up a JAR in the first place. We therefore move them to `scripts/`, where integration tests live.

### Sample Commands
`npm run test:storage-emulator-integration` will run these tests before `storage-emulator-integration/tests.ts`.

### DIFF
--- a/scripts/storage-emulator-integration/run.sh
+++ b/scripts/storage-emulator-integration/run.sh
@@ -7,6 +7,8 @@ set -e # Immediately exit on failure
 # Prepare the storage emulator rules runtime
 firebase setup:emulators:storage
 
+mocha scripts/storage-emulator-integration/rules/*.test.ts
+
 mocha \
   --require ts-node/register \
   --require source-map-support/register \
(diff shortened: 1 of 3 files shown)


scripts/storage-emulator-integration/run.sh
```diff
@@ -7,6 +7,8 @@ set -e # Immediately exit on failure
 # Prepare the storage emulator rules runtime
 firebase setup:emulators:storage
 
+mocha scripts/storage-emulator-integration/rules/*.test.ts
```
bkendall, line 10: I'm a little surprised you don't need the `-require` lines below, but it does look like the tests are being run (successfully, too).

bkendall (approved):
Thanks for doing this!

scripts/storage-emulator-integration/run.sh
taeold: I believe mocha automatically pickes up .mocharc in the parent dir.

bkendall: That makes total sense. Probably means I could clean up a lot of these lines in the integration tests haha

tohhsinpei: Done; PR here: #4312

PR #4312, opened 2022-03-16
Remove --require from mocha commands in integration test scripts

tohhsinpei:
### Description
2 changes:

1. What the title of this PR says. As per this [comment](https://github.com/firebase/firebase-tools/pull/4309#discussion_r828357428), we don't need to specify `--require` in the `mocha` commands in the integration test scripts because the requirements are already specified in the `.mocharc` in the parent directory: https://github.com/firebase/firebase-tools/blob/c69b43757248078e0487eb9d9b666a30b1f8e9ed/.mocharc.yml#L1-L4

2. (Just throwing this one in because it has been bothering me) Add `ui-debug.log`, which is generated when running `npm run test:storage-emulator-integration`, to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ node_modules
 firebase-debug.log
 firebase-debug.*.log
 npm-debug.log
+ui-debug.log
 yarn.lock
 .npmrc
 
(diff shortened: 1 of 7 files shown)


taeold (commented):
Neat!

taeold (approved):

bkendall (approved):
👍

Issue #4313, opened 2022-03-16
TypeError: Cannot read property 'data' of undefined since v10.3.1


jnizet:
### [REQUIRED] Environment info

**firebase-tools:** 10.3.1

**Platform:** macOS, Ubuntu

### [REQUIRED] Test case

Hard to do. Here's however the URL of a PR on a public project that makes the tests fail because a firebase function running in the emulator is killed due to an internal error: https://github.com/jnizet/decouvertes-nature/pull/110.  

The stack trace is:
```
[emulators] i  functions: Beginning execution of "listUsers"
[emulators] ⚠  functions: TypeError: Cannot read property 'data' of undefined
[emulators]     at processBackground (/home/runner/work/decouvertes-nature/decouvertes-nature/node_modules/firebase-tools/lib/emulator/functionsEmulatorRuntime.js:550:24)
[emulators]     at invokeTrigger (/home/runner/work/decouvertes-nature/decouvertes-nature/node_modules/firebase-tools/lib/emulator/functionsEmulatorRuntime.js:640:19)
[emulators]     at handleMessage (/home/runner/work/decouvertes-nature/decouvertes-nature/node_modules/firebase-tools/lib/emulator/functionsEmulatorRuntime.js:742:15)
[emulators]     at processTicksAndRejections (internal/process/task_queues.js:95:5)
[emulators] ⚠  Your function was killed because it raised an unhandled error.
```

The source code of the function is that one: https://github.com/jnizet/decouvertes-nature/blob/renovate/firebase-tools-10.x/functions/src/users/users.ts#L51-L56

This function has never had any issue until v10.3.1

taeold:
Fix is released in 10.4.0! Thanks for reporting the issue.

PR #4314, opened 2022-03-16
Fix bug where callable functions can't be emulated.

taeold:
Fixes https://github.com/firebase/firebase-tools/issues/4313

In https://github.com/firebase/firebase-tools/pull/4310, we started to support new type of triggers `callableTrigger` (distinct from `httpTrigger`).

Functions Emulator does not support `callableTrigger`. We fix this by converting `callbleTrigger` to `httpTrigger`. This is tech debt - at some point, we should use `endpoint` definitions throughout the emulator (cc @joehan).

Also adds integration tests for emulating callable triggers to prevent this kind of breakages in the future.

### DIFF
--- a/scripts/integration-helpers/framework.ts
+++ b/scripts/integration-helpers/framework.ts
@@ -268,7 +268,7 @@ export class TriggerEndToEndTest {
     body: Record<string, unknown>,
     zone = FIREBASE_PROJECT_ZONE
   ): Promise<Response> {
-    const url = `http://localhost:${[this.functionsEmulatorPort, this.project, zone, name].join(
+    const url = `http://localhost:${this.functionsEmulatorPort}/${this.project, zone, name].join(
       "/"
     )}`;
     return fetch(url, {
--- a/scripts/integration-helpers/framework.ts
+++ b/scripts/integration-helpers/framework.ts
@@ -268,7 +268,7 @@ export class TriggerEndToEndTest {
     body: Record<string, unknown>,
     zone = FIREBASE_PROJECT_ZONE
   ): Promise<Response> {
-    const url = `http://localhost:${this.functionsEmulatorPort}/${this.project, zone, name].join(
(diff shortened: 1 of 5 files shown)


scripts/integration-helpers/framework.ts
```diff
@@ -263,6 +263,21 @@ export class TriggerEndToEndTest {
     return fetch(url);
   }
 
+  invokeCallableFunction(
+    name: string,
+    body: Record<string, unknown>,
+    zone = FIREBASE_PROJECT_ZONE
+  ): Promise<Response> {
+    const url = `http://localhost:${[this.functionsEmulatorPort, this.project, zone, name].join(
```
bkendall, line 271: This is real strange, just given how URLs should be thought of. Consider pulling the port out of the path:

```suggestion
    const url = `http://localhost:${this.functionsEmulatorPort}/${this.project, zone, name].join(
```

scripts/triggers-end-to-end-tests/functions/index.js
```diff
@@ -359,3 +364,8 @@ exports.storagebucketv2metadatareaction = functionsV2.storage.onObjectMetadataUp
     return true;
   }
 );
+
+exports.oncallv2 = functionsV2.https.onCall((req) => {
```
bkendall, line 368: `onCallv2` Doesn't match `onCall` above as written

bkendall (changes requested):

scripts/triggers-end-to-end-tests/functions/index.js
taeold: v2 functions can't contain capital letters.

bkendall (approved):

colerogers (approved):

BaptisteCool:
Release in 1.4.0 google [here](https://github.com/firebase/firebase-tools/issues/4320)

PR #4317, opened 2022-03-17
typescript requireConfig

### LINKED ISSUE #4082: Convert to TS: src/requireConfig.js
The following file `src/requireConfig.js` should be converted to Typescript.

bkendall:
### Description

Typescript requireConfig, simplify the logic, and testaroo.

Fixes #4082

### Scenarios Tested

Deploy hosting in a firebase folder, and try in not a firebase folder.


src/requireConfig.ts
```diff
@@ -0,0 +1,10 @@
+import { FirebaseError } from "./error";
+import { Options } from "./options";
+
+export function requireConfig(options: Options): void {
+  if (!options.config) {
+    throw options.configError
+      ? options.configError
+      : new FirebaseError("Not in a Firebase project directory (could not locate firebase.json)");
+  }
+}
```
taeold, line 10: Looks like `requireConfig` is expected to be a promise since it's often used in commands' `before(requireConfig)` chain.

bkendall: Booo - I thought I checked in on this. There's absolutely no reason it _needs_ to be in a promise... I'll see if I can fix the underlying bit or make this actually return a promise

src/requireConfig.ts
```diff
@@ -0,0 +1,11 @@
+import { FirebaseError } from "./error";
+import { Options } from "./options";
+
+export async function requireConfig(options: Options): Promise<void> {
+  await Promise.resolve(); // Allows this function to remain `async`.
```
joehan, line 5: Lol, now I'm wondering why this was ever async in the first place.

taeold: lol amazing hack

PR #4324, opened 2022-03-17
Adds required IAM bindings for EventArc events

colerogers:
Adds the required roles to the pubsub and default compute service agents in order to correctly deploy EventArc triggers.


src/deploy/functions/checkIam.ts
```diff
@@ -181,6 +262,9 @@ export async function ensureServiceAgentRoles(
     findRequiredBindings.push(service.requiredProjectBindings!(projectNumber, policy))
   );
   const allRequiredBindings = await Promise.all(findRequiredBindings);
+  allRequiredBindings.push(obtainPubSubServiceAgentBindings(projectNumber, policy));
+  allRequiredBindings.push(obtainDefaultComputeServiceAgentBindings(projectNumber, policy));
+  allRequiredBindings.push(obtainEventarcServiceAgentBindings(projectNumber, policy));
```
taeold, line 267: IIUC, we try to setup these bindings on every deployment, regardless of whether the deployment requires it. Is that true? I'd like to avoid it, especially since I suspect CF3v1 customers don't even have eventarc enabled.

colerogers: Resolved over chat, we don't set these since we map to v2 services and v1 is ignored

src/test/deploy/functions/checkIam.spec.ts
```diff
@@ -253,4 +430,174 @@ describe("checkIam", () => {
...
+        retry: false,
+      },
+      ...SPEC,
+    };
+
+    await checkIam.ensureServiceAgentRoles(projectNumber, backend.of(wantFn), backend.of(haveFn));
+
+    expect(getIamStub).to.have.been.calledOnce;
+    expect(setIamStub).to.have.been.calledOnce;
+    expect(setIamStub).to.have.been.calledWith(projectNumber, newIamPolicy, "bindings");
```
taeold, line 601: aside - why does setIam gets called if no new binding is introduced?

src/test/deploy/functions/checkIam.spec.ts
```diff
@@ -253,4 +430,174 @@ describe("checkIam", () => {
...
+
+    await checkIam.ensureServiceAgentRoles(projectNumber, backend.of(wantFn), backend.of(haveFn));
+
+    expect(storageStub).to.have.been.calledOnce;
+    expect(getIamStub).to.have.been.calledOnce;
+    expect(setIamStub).to.have.been.calledOnce;
+    expect(setIamStub).to.have.been.calledWith(projectNumber, newIamPolicy, "bindings");
+  });
+
+  it("should add the default bindings for a new v2 alerts function without v2 deployed functions", async () => {
```
taeold, line 494: hm I seem to be missing something. Can you explain to me why PUBSUB TOKEN CREATOR binding isn't required when setting up firealerts functions?

colerogers: You mean the `roles/pubsub.publisher` role with the storage service account, right? That binding is only required for storage since the storage service account needs to be able to create events.

src/test/deploy/functions/checkIam.spec.ts
```diff
@@ -44,6 +50,157 @@ describe("checkIam", () => {
...
+  describe("obtainPubSubServiceAgentBindings", () => {
+    it("should add the binding", () => {
+      const policy = { ...iamPolicy };
+
+      const bindings = checkIam.obtainPubSubServiceAgentBindings(projectNumber, policy);
+
+      expect(bindings.length).to.equal(1);
+      expect(bindings[0]).to.deep.equal({
+        role: checkIam.SERVICE_ACCOUNT_TOKEN_CREATOR_ROLE,
+        members: ["serviceAccount:service-123456789@gcp-sa-pubsub.iam.gserviceaccount.com"],
```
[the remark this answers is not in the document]
colerogers: woops, fixed these set of strings

src/test/deploy/functions/checkIam.spec.ts
```diff
@@ -44,6 +50,157 @@ describe("checkIam", () => {
...
+
+      expect(bindings.length).to.equal(2);
+      expect(bindings[0]).to.deep.equal({
+        role: checkIam.RUN_INVOKER_ROLE,
+        members: [`serviceAccount:${projectNumber}-compute@developer.gserviceaccount.com`],
+      });
+      expect(bindings[1]).to.deep.equal({
+        role: checkIam.EVENTARC_EVENT_RECEIVER_ROLE,
+        members: [`serviceAccount:${projectNumber}-compute@developer.gserviceaccount.com`],
+      });
```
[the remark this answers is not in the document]
colerogers: Oh nice, way cleaner 🔥

src/test/deploy/functions/checkIam.spec.ts
colerogers: ahh good catch, added a check to return early if we don't have new bindings to add

taeold (approved):

PR #4328, opened 2022-03-18
Substitute params into backendInfo, and add unit tests

joehan:
### Description
During our bug bash, we noticed that param values were not being subbed into postinstall. In console, this comes from the backend, so I figure it is easiest to do it on the backend for the emulator as well.

I also separated out the logic for transforming an EmulatableBackend to a BackendInfo and unit tested it.
### Scenarios Tested
Ran this against the snapshot UI and confirmed that we see subbed postinstall.
Screen Shot 2022-03-17 at 3 11 11 PM


src/emulator/functionsEmulator.ts
```diff
@@ -63,6 +64,7 @@ import { accessSecretVersion } from "../gcp/secretManager";
 import * as runtimes from "../deploy/functions/runtimes";
 import * as backend from "../deploy/functions/backend";
 import * as functionsEnv from "../functions/env";
+import { substituteParams } from "../extensions/extensionsHelper";
```
elvisun, line 67: unused import

src/emulator/functionsEmulatorShared.ts
```diff
@@ -391,3 +390,39 @@ export function getSecretLocalPath(backend: EmulatableBackend, projectDir: strin
...
   return path.join(secretDirectory, secretsFile);
 }
+
+/**
+ * toBackendInfo trasnforms an EmulatableBackend into its correspondign API type, BackendInfo
+ * @param e the emulatableBackend to transform
+ * @param cf3Triggers a list of CF3 triggers. If e does not include predefinedTriggers, these will be used instead.
+ */
+export function toBackendInfo(
+  e: EmulatableBackend,
```
elvisun, line 400: nit: maybe backend?

elvisun (approved):

src/emulator/functionsEmulatorShared.ts
```diff
@@ -391,3 +390,39 @@ export function getSecretLocalPath(backend: EmulatableBackend, projectDir: strin
...
+  cf3Triggers: ParsedTriggerDefinition[]
+): BackendInfo {
+  const envWithSecrets = Object.assign({}, e.env);
+  for (const s of e.secretEnv) {
+    envWithSecrets[s.key] = backend.secretVersionName(s);
+  }
+  let extensionVersion = e.extensionVersion;
```
NothingEverHappens, line 407: This is a weird construct with let, could this be replaced with &&
Same below

NothingEverHappens (approved):

Issue #4329, opened 2022-03-18
Firestore Storage Emulator always evaluate overwriting files as "RulesetOperationMethod.CREATE"


SamyPesse:
### [REQUIRED] Environment info

**firebase-tools:** 10.4.1

**Platform:** macOS

### [REQUIRED] Steps to reproduce

While implementing tests to ensure a fie cannot be modified in a bucket, we are confronted to an issue with the Storage emulator: all file updates are being evaluated as "create" operations: https://github.com/firebase/firebase-tools/blob/089bea2ba739c0e03d858e76119e47144eda342c/src/emulator/storage/files.ts#L322-L323

It means that the following rules cannot be tested:

```
service firebase.storage {
    match /b/{bucket}/o {
        match /something/{fileName} {
            allow read: if true;
            allow write: if  resource == null ;
        }
    }
}
```

⚠️ I expect the production storage backend does't behave like this otherwise it's a massive security flaw.

### [REQUIRED] Expected behavior

If the file exists in the storage, it should be evaluated as a `UPDATE`, similar to how it's done for metadata: https://github.com/firebase/firebase-tools/blob/089bea2ba739c0e03d858e76119e47144eda342c/src/emulator/storage/files.ts#L273-L277

### [REQUIRED] Actual behavior

The operation is evaluated as a `CREATE`, independently from the file already existing or not.

tonyjhuang:
Hi @SamyPesse, thanks for filing this. This behavior mimics the behavior of the storage server in production today. Please see https://github.com/firebase/firebase-js-sdk/issues/5079#issuecomment-871712273 for an explanation of what's going on.

Note that since the storage emulator does not support GCS object versioning, the workaround in the linked issue will not work in the emulator environment. I understand this can be confusing and but as of now, this is WAI.

SamyPesse:
@tonyjhuang While looking deeper into it. I still can't reproduce this behaviour (which is a good thing imo, but I feel like I'm missing something).

Here is a minimal reproducible example: https://github.com/SamyPesse/firebase-storage-repro
I've just tested with a new firebase project and I'm getting the following when running it:

```
node ./test.js
Uploading first version...
Uploading second version...
[StorageError [FirebaseError]: Firebase Storage: User does not have permission to access 'test-9309.txt'. (storage/unauthorized)] {
  code: 'storage/unauthorized',
  customData: { serverResponse: '' },
[...]

PR #4330, opened 2022-03-18
Storage emulator: define "resource" when writing over an existing file

SamyPesse:
### Description

This PR is an attempts at fixing the bug https://github.com/firebase/firebase-tools/issues/4329

It ensures a `firebase.storage().ref('filePath').putString()` on an existing `filePath` will correctly evaluate the rules with `resource` being defined.


src/emulator/storage/files.ts
```diff
@@ -309,13 +310,20 @@ export class StorageLayer {
...
+
     const authorized = await this._rulesValidator.validate(
       ["b", upload.bucketId, "o", upload.objectId].join("/"),
       upload.bucketId,
+      // Uploading a file always evaluates 'create' and 'write'
+      // 'update' is only evaluated on metadata update
       RulesetOperationMethod.CREATE,
-      { after: metadata?.asRulesResource() },
+      {
+        before: storedMetadata?.asRulesResource(),
```
tonyjhuang, line 321: thanks for the fix, can you add an integration test?

SamyPesse:
Closed as it has been fixed in https://github.com/firebase/firebase-tools/pull/4828

Issue #4327, opened 2022-03-18
firebase CLI 10.3.1 does not issue calls to make callable functions public


taeold:
I'm also having this issue Using firebase CLI 10.3.1 and on deploying a _new_ function, it is always deployed in a way that causes it to return 403 when triggered using the firebase SDK `firebaseFunctions().httpsCallable()`. My fix so far has been to edit it in the console to 'Allow Unauthenticated', but it would like to avoid that if it's supposed to work differently.

I'm running firebase deploy locally, logged in as the account owner, so I doubt it is a permissions issue. The debug log has these lines early on:
```
[debug] [2022-03-18T10:20:59.704Z] > command requires scopes: ["email","openid","https://www.googleapis.com/auth/cloudplatformprojects.readonly","https://www.googleapis.com/auth/firebase","https://www.googleapis.com/auth/cloud-platform"]
[debug] [2022-03-18T10:20:59.705Z] > authorizing via signed-in user ([REMOVED])
[debug] [2022-03-18T10:20:59.705Z] [iam] checking project [REMOVED] for permissions ["cloudfunctions.functions.create","cloudfunctions.functions.delete","cloudfunctions.functions.get","cloudfunctions.functions.list","cloudfunctions.functions.update","cloudfunctions.operations.get","datastore.indexes.create","datastore.indexes.delete","datastore.indexes.list","datastore.indexes.update","firebase.projects.get","firebaserules.releases.create","firebaserules.releases.update","firebaserules.rulesets.create"]
[debug] [2022-03-18T10:20:59.708Z] >>> [apiv2][query] POST https://cloudresourcemanager.googleapis.com/v1/projects/[REMOVED]:testIamPermissions [none]
[debug] [2022-03-18T10:20:59.708Z] >>> [apiv2][body] POST https://cloudresourcemanager.googleapis.com/v1/projects/[REMOVED]:testIamPermissions {"permissions":["cloudfunctions.functions.create","cloudfunctions.functions.delete","cloudfunctions.functions.get","cloudfunctions.functions.list","cloudfunctions.functions.update","cloudfunctions.operations.get","datastore.indexes.create","datastore.indexes.delete","datastore.indexes.list","datastore.indexes.update","firebase.projects.get","firebaserules.releases.create","firebaserules.releases.update","firebaserules.rulesets.create"]}
[...]

Issue #3965, opened 2021-12-20
allUsers > Cloud Functions Invoker role randomly added to newly deployed HTTP functions


feliksmatsko:
### Environment info
**firebase-tools:** 10.0.1
**Platform:** macOS

### Test case
https://github.com/feliksmatsko/firebase-cli-issue

### Steps to reproduce
1. Create more than 40 normal http cloud functions ([example](https://github.com/feliksmatsko/firebase-cli-issue/blob/master/functions/src/index.ts))
2. run cli: _firebase deploy --only functions_

### Expected behavior
For ALL newly deployed cloud functions to have the same permissions

### Actual behavior
Some functions allow authenticated and some don't. The number of functions with and without invoker created is random.
Multiple errors in cli: "Error Failed to set invoker function f38 in region us-central1"

<img width="1266" alt="Screenshot 2021-12-20 at 15 20 16" src="https://user-images.githubusercontent.com/3799407/146773625-d94a02b9-c906-4252-8eb7-e9d75e3c586c.png">

bkendall:
Do you have a log file for when this happens? That would help greatly in trying to debug this. I'm not sure off hand of an issue where this would only occasionally work but maybe there's something that can help shed light on it.

inlined:
We can add more retries to this API call, but it tends to just be flaky. If you set `invoker: 'public'` in your functions we make sure this is set on every deploy.

PR #4335, opened 2022-03-18
Ensure callable functions have invoker correctly set to "public"

chrisbrown-io:
### Description

Implemented fix for issue introduced in `v10.3.0` where callable functions were not having their `invoker` property correctly set to `"public"` that would in turn cause access issues with the deployed functions (as detailed in #4327).

Will also fix issues such as #3965, that are presumably cause by version mismatches/upgrades introducing this issue.

[See comment for explicit info](https://github.com/firebase/firebase-tools/issues/4327#issuecomment-1072650006).

Fixes issue by adding specific handling for callable functions and always setting the `invoker` to `"public"`.

Also fixed issue with `taskQueueTrigger` endpoints calling wrong method/passing wrong args when trying to set invoker.

### Scenarios Tested

Thanks to @taeold we have explicit tests for each function type and the expected outcomes for both `createV1Function` and `createV2Function`  methods.


chrisbrown-io:
@taeold - you mentioned in your [previous comment](https://github.com/firebase/firebase-tools/issues/4327#issuecomment-1072690202) that the invoker is not updated on update, however the source code/tests for the update methods look as if they would update the invoker if set (and subsequently not update the invoker for callable functions unless additional logic was added).

Are you saying we don't need to handle this case, as users can't specify an `invoker` for callable functions, meaning this scenario of changing the invoker on update will never happen?

src/deploy/functions/release/fabricator.ts
```diff
@@ -233,8 +233,16 @@ export class Fabricator {
           })
           .catch(rethrowAs(endpoint, "set invoker"));
       }
+    } else if (backend.isCallableTriggered(endpoint)) {
+      // Callable functions should always be public
+      const invoker = ["public"];
```
taeold, line 238: nit* Personal preference, but I'd inline this value instead of creating an intermediate var.

Same comment for L310

src/deploy/functions/release/fabricator.ts
```diff
@@ -233,8 +233,16 @@ export class Fabricator {
...
+      await this.executor
+        .run(async () => {
+          await gcf.setInvokerCreate(endpoint.project, backend.functionName(endpoint), invoker);
+        })
+        .catch(rethrowAs(endpoint, "set invoker"));
     } else if (backend.isTaskQueueTriggered(endpoint)) {
-      // Like HTTPS triggers, taskQueueTriggers have an invoker, but unlike HTTPS they don't default
+      // Like other triggers, taskQueueTriggers can have an invoker, but they don't default
```
taeold, line 245: The previous comment is still true since only task queue triggers and http triggers can explicitly set invokers. I'd prefer if we revert this change.

Same comment for L315

taeold (changes requested):
> Are you saying we don't need to handle this case, as users can't specify an invoker for callable functions, meaning this scenario of changing the invoker on update will never happen?

Yes. User may have changed invoker for good reasons. I think we should respect that.

Also - can we add an entry to CHANGELOG.md file?

src/deploy/functions/release/fabricator.ts
chrisbrown-io: Sure - makes sense.

src/deploy/functions/release/fabricator.ts
chrisbrown-io: 👍

CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- Fix for issue introduced where callable functions were not having their invoker property correctly set to "public" (#4327).
```
taeold, line 1:
```suggestion
- Fixes bug where new callable functions were not publicly accessible (#4327).
```

taeold (approved):
🔥 Amazing contribution!

PR #4281, opened 2022-03-09
Add StorageRulesManagerRegistry to handle rules files for multiple targets

tohhsinpei:
### Description
Fixes #3390. Following up to #4263, this PR adds a `StorageRulesManager` that will live inside the Storage Emulator. The interface is implemented by:

- `StorageRulesManagerImplementation`, which keeps track of rules for a single resource
- `StorageRulesManagerRegistry`, which maps resource to `StorageRulesManagerImplementation`

In the server, all calls to the validator will pass along the `resource` aka bucket id in order to determine which ruleset to use.

PR #4337, opened 2022-03-18
Update /internal/setRules endpoint; fix ruleset file watcher

tohhsinpei:
### Description
2 changes:

1. Follow up to this [comment](https://github.com/firebase/firebase-tools/pull/4281#discussion_r828326627) in #4281. This PR adds support for per-resource Storage rules files on the `/internal/setRules` endpoint. Note that we only allow callers to overwrite the entire rules config, not update rules for a given resource. This means we can kill the `updateSourceFile` endpoint on the rules manager.
2. Fix the rules file watcher, which I broke in #4245. We need to reread the source file every time the file watcher detects a change at the given path; otherwise we're just reloading the same contents as before. This was not caught by the integration test because of another bug there: https://github.com/firebase/firebase-tools/blob/089bea2ba739c0e03d858e76119e47144eda342c/scripts/storage-emulator-integration/rules/manager.test.ts#L102
Note `opts` does not contain `ruleset`, so this always returns false: https://github.com/firebase/firebase-tools/blob/089bea2ba739c0e03d858e76119e47144eda342c/src/emulator/storage/rules/utils.ts#L71-L77


src/emulator/storage/rules/manager.ts
```diff
@@ -103,10 +91,15 @@ class DefaultStorageRulesManager implements StorageRulesManager {
           "storage",
           "Change detected, updating rules for Cloud Storage..."
         );
+        this.updateRulesSource(rulesFile);
```
tohhsinpei, line 94: This is the fix for 2)

src/emulator/storage/index.ts
```diff
@@ -113,6 +113,16 @@ export class StorageEmulator implements EmulatorInstance {
...
   }
 
+  async replaceRules(rules: SourceFile | RulesConfig[]): Promise<StorageRulesIssues> {
+    await this._rulesManager.stop();
+    this.createRulesManager(rules);
+    return this._rulesManager.start();
+  }
+
+  private createRulesManager(rules: SourceFile | RulesConfig[]): void {
+    this._rulesManager = createStorageRulesManager(rules, this._rulesRuntime);
```
tonyjhuang, line 123: small nit: Would prefer we minimize side effects and remove them where possible. See go/java-practices/methods#side-effects. 

e.g. compare:
```
this._rulesManager = createRulesManager(...);
// vs 
createRulesManager(...);
```

tonyjhuang: While this isn't necessarily too unexpected from a method named `create...()`, it would still be nice to make mutation explicit. Another resource: go/java-practices/immutability

src/emulator/storage/rules/manager.ts
```diff
@@ -103,10 +91,15 @@ class DefaultStorageRulesManager implements StorageRulesManager {
...
         await this.loadRuleset();
       });
   }
 
+  private updateRulesSource(rulesFile: string): void {
```
tonyjhuang, line 99: nit: inline

src/emulator/storage/rules/manager.ts
```diff
@@ -157,12 +150,6 @@ class ResourceBasedStorageRulesManager implements StorageRulesManager {
     return this._rulesManagers.get(resource)?.getRuleset();
   }
 
-  updateSourceFile(rules: SourceFile, resource: string): Promise<StorageRulesIssues> {
```
tonyjhuang, line 160: I just noticed `ResourceBasedStorageRulesManager` is missing hot reloading of storage rules. Can you add that in too?

src/emulator/storage/server.ts
```diff
@@ -119,3 +162,11 @@ export function createApp(
 
   return Promise.resolve(app);
 }
+
+type RulesFile = { name: string; content: string; resource?: string };
+
+function isRulesFile(file: unknown): file is RulesFile {
```
tonyjhuang, line 168: why can't you replace this entire function with `obj instanceof RulesFile`?

src/emulator/storage/rules/manager.ts
```diff
@@ -65,22 +59,16 @@ class DefaultStorageRulesManager implements StorageRulesManager {
...
 
   getRuleset(): StorageRulesetInstance | undefined {
     return this._ruleset;
   }
 
-  async updateSourceFile(rules: SourceFile): Promise<StorageRulesIssues> {
-    const prevRulesFile = this._rules.name;
-    this._rules = rules;
-    const issues = await this.loadRuleset();
-    this.updateWatcher(rules.name, prevRulesFile);
```
tonyjhuang, line 80: I don't see where the `prevRulesFile` arg is being used anymore

src/emulator/storage/index.ts
tohhsinpei: Done.

src/emulator/storage/rules/manager.ts
tohhsinpei: Done.

src/emulator/storage/server.ts
tohhsinpei: Because `RulesFile` is a type alias, not a class. Btw, I'm not too familiar with type predicates either, just following https://google.github.io/styleguide/tsguide.html#any-unknown.

src/emulator/storage/rules/manager.ts
tohhsinpei: Good catch, we can get rid of it

src/emulator/storage/rules/manager.ts
tonyjhuang: Sorry you're right, forgot about the relationship between the two

tonyjhuang (approved):
thanks for the fix!

PR #4339, opened 2022-03-19
revert hosting upload concurrency to 200

bkendall:
### Description

So here's the deal. 10 is real slow. Gains after ~150 get pretty minimal. I'm going to leave the default at 200 but still allow folks to adjust the concurrency with `FIREBASE_HOSTING_UPLOAD_CONCURRENCY`.

### Scenarios Tested

Concurrency | File Count | Upload Time | Seconds
-- | -- | -- | --
10 | 10000 | 61.41s user 12.11s system 10% cpu 11:52.00 total | 712
50 | 1000 | 52.94s user 10.13s system 49% cpu 2:07.57 total | 127
100 | 10000 | 48.56s user 9.07s system 65% cpu 1:27.61 total | 87
150 | 10000 | 44.92s user 7.98s system 105% cpu 49.914 total | 49
200 | 10000 | 42.33s user 7.42s system 115% cpu 43.071 total | 43


bkendall:
I do better with graphs

image

taeold (approved):

PR #4349, opened 2022-03-22
Refactor gcloud to use same list method in StorageLayer

tonyjhuang:
### Description

Refactor gcloud to use same list method in StorageLayer.

Also drops the `handle...` prefix from `StorageLayer` method names.


src/emulator/storage/apis/firebase.ts
```diff
@@ -146,9 +147,9 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
   // list object handler
   firebaseStorageAPI.get("/b/:bucketId/o", async (req, res) => {
     const maxResults = req.query.maxResults?.toString();
-    let response: ListResponse;
+    let listResponse: ListObjectsResponse;
     try {
-      response = await storageLayer.handleListObjects({
+      listResponse = await storageLayer.listObjects({
```
abhis3, line 152: Do you mind refactoring [this method](https://github.com/firebase/firebase-tools/blob/tonyjhuang/list/src/emulator/storage/files.ts#L469) to remove the `handle` prefix since you're doing it for the other methods

src/emulator/storage/apis/gcloud.ts
```diff
@@ -83,30 +83,35 @@ export function createCloudEndpoints(emulator: StorageEmulator): Router {
...
+        bucketId: req.params.bucketId,
+        prefix: req.query.prefix ? req.query.prefix.toString() : "",
+        delimiter: req.query.delimiter ? req.query.delimiter.toString() : "",
+        pageToken: req.query.pageToken ? req.query.pageToken.toString() : undefined,
+        maxResults: req.query.maxResults ? +req.query.maxResults.toString() : undefined,
+        authorization: req.header("authorization"),
+      });
+    } catch (err) {
+      if (err instanceof ForbiddenError) {
+        return res.sendStatus(403);
```
abhis3, line 100: Not specific to this PR, but do you know why we aren't sending error messages in `gcloud` [like we do in `firebase.ts`](https://github.com/firebase/firebase-tools/blob/cdca8f24230987e6d1d54d2772aefce47e9a2350/src/emulator/storage/apis/firebase.ts#L165)

Seems like it would be useful in general, but perhaps it's because we don't have any guarantees of what GCS actually returns as the message besides the 403?

abhis3: https://cloud.google.com/storage/docs/json_api/v1/status-codes#403-forbidden

Seems it can be a plethora of messages, fine with leaving the 403 as is. I'll leave the comment open in case you have any thoughts, but feel free to resolve.

src/emulator/storage/apis/firebase.ts
tonyjhuang: good catch. done.

src/emulator/storage/files.ts
```diff
@@ -89,14 +86,21 @@ export type ListObjectsRequest = {
   authorization?: string;
 };
 
-/**  Parsed request object for {@link StorageLayer#handleCreateDownloadToken}. */
+/** Response object for {@link StorageLayer#listObjects}. */
+export type ListObjectsResponse = {
```
tohhsinpei, line 90: I'd combine this and the existing [ListResponse](https://github.com/firebase/firebase-tools/blob/ba53736ade37a4bc147fc792d2403fe1e4635149/src/emulator/storage/list.ts#L10) type. If you choose to return undefined instead of empty arrays (as is the current behaviour), make sure this doesn't break anything on the SDK side.

tonyjhuang: Removed `list.ts`, added another integration test to make sure the response shape hasn't changed.

tohhsinpei: Could you make the new type have non-optional properties then? Then no need for `?? []`s

tohhsinpei: ? In `gcloud.ts` is it not just `listResponse.items.map(...)`?

tonyjhuang: We expect the `items` and `prefix` key to not be present in the response object of `gcloud.ts` if there are no elements in those arrays. I'm not sure express strips out empty arrays passed into `res.json`

src/emulator/storage/apis/gcloud.ts
abhis3: Agreed, mainly just asked out of curiosity this shouldn't be our priority now. Thanks for the clarification!

src/emulator/storage/files.ts
tohhsinpei: Talked about this offline, what you have now is fine then

tohhsinpei (approved):

abhis3 (approved):

PR #4350, opened 2022-03-22
Add support for codebase property for function deployments.

taeold:
**Warning** This feature is being actively developed. Please do not use.

## Feature description
We add preliminary support for new "codebase" property for function deployments.

We allow users to add a codebase property in their `firebase.json`:

```json
"functions": { 
  "source": "functions",
  "codebase": "my-fns"
}
```

Adding codebase annotation has following effects:

  1. Deployed function will be tagged with label `firebase-functions-codebase={codebase}`.
  2. Function deployment will be scoped down to only look at functions within the said codebase. Practically, this means that a function deployment will _not_ touch (a.k.a. delete) functions from other codebases.

This makes it easy for users to split their function definitions across different Firebase project repositories.

In addition, filters specified using `--only` flag will allow users to specify `codebase` as a resource selector. e.g.

```bash
# "Only deploy function w/ id func-id"
$ firebase deploy --only functions:func-id

#  "Only deploy function w/ id func-id in codebase foo"
$ firebase deploy --only functions:foo:func-id
```

This isn't very useful today as long as we only can have single codebase within a project directory.

(Note that this PR does not bring codebase support to other function commands like `functions:{delete,log,list}`. This will come in coming PRs.

## Implementation notes

* Every endpoint is now annotated w/ "codebase" property. When parsing triggers, we use value from `firebase.json` to fill in the property. When loading existing functions from GCP, labels are derived from label `firebase-functions-codebase` (defaults to `default` if missing).

* Minor refactoring of filtering logic (i.e. `--only`) to support filtering by codebase.


src/gcp/cloudfunctions.ts
```diff
@@ -8,12 +8,12 @@ import * as utils from "../utils";
...
 import * as runtimes from "../deploy/functions/runtimes";
 import * as iam from "./iam";
+import * as projectConfig from "../functions/projectConfig";
 import { Client } from "../apiv2";
 import { functionsOrigin } from "../api";
-import { getFirebaseProject } from "../management/projects";
-import { assertExhaustive } from "../functional";
 
 export const API_VERSION = "v1";
+export const CODEBASE_LABEL = "firebase-functions-codebase";
```
colerogers, line 16: Maybe we should be doing this with all of our labels?

src/deploy/functions/args.ts
```diff
@@ -1,13 +1,15 @@
 import * as backend from "./backend";
 import * as gcfV2 from "../../gcp/cloudfunctionsv2";
 import * as projectConfig from "../../functions/projectConfig";
+import * as deployHelper from "./functionsDeployHelper";
 
 // These types should proably be in a root deploy.ts, but we can only boil the ocean one bit at a time.
 
 // Payload holds the output types of what we're building.
```
colerogers, line 8: If we're adding what we already have, we should probably update this description a bit

src/deploy/functions/functionsDeployHelper.ts
```diff
@@ -1,42 +1,129 @@
...
+  // If id chunks is undefined, match all function in the said codebase.
+  idChunks?: string[];
+}
+
+/**
+ * Returns true if endpoint matches any of the given filter.
+ *
+ * If no filter is passed, always returns true.
+ */
+export function endpointMatchesAnyFilter(endpoint: backend.Endpoint, filters?: EndpointFilter[]) {
```
colerogers, line 17: nit: return type

src/gcp/cloudfunctions.ts
taeold: Good call. But let's do it in another PR?

colerogers (approved):
🔥 🔥 🔥

inlined (commented):
It seems this is missing a lot of code in prepare/deploy to support multiple codebases? I guess that's just coming?

I have mixed feels about naming a default codebase. That seems to be contrary to our API proposal.

Finally, I'd love some metrics for how many codebase are being deployed. Maybe a "functions_codebase_count" field with a property that is either "1", "2", "3", "4", or "5+"

PR #4352, opened 2022-03-22
Only validate function name

joehan:
### Description
While testing the extensions emulator, I noticed that this length validation on functionId was blocking some function names that would be valid in prod. In the emulator, functionId = {region}-{functionName}, but in prod, functionId=functionName. Therefore, to match prod behavior, we should only validate functionName.


CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- Fixes an issue where some valid Cloud Functions for Firebase names would not pass validation in the Functions emulator.
```
taeold, line 1:
```suggestion
- Fixes an issue where some valid Cloud Functions for Firebase names would not pass validation in the Functions emulator (#4352).
```

src/emulator/functionsEmulator.ts
```diff
@@ -551,7 +551,9 @@ export class FunctionsEmulator implements EmulatorInstance {
     for (const definition of toSetup) {
       // Skip function with invalid id.
       try {
-        functionIdsAreValid([definition]);
+        // Note - in the emulator, functionId = {region}-{functionName}, but in prod, functionId=functionName.
+        // To match prod behavior, only validate functionName
+        functionIdsAreValid([{ id: definition.name, platform: definition.platform }]);
```
taeold, line 556: personal pref, but I'd do something like:

```
functionIdsAreValid([{ ...definition, id: definition.name }]);
```

taeold (approved):

src/emulator/functionsEmulator.ts
joehan: Ahhh, i do like this a bit more - done.

PR #4354, opened 2022-03-23
Improve error messages when trying to use alpha featuers without enablement

inlined:
Trying to avoid people hitting the same error over and over again.


taeold (approved):

taeold (approved):
nit* I wouldn't bother checking in `npm-shrinkwrap.json`

PR #4362, opened 2022-03-25
When channel is defined in the CF3 trigger pass it along to GCF CP

pavelgj:
This is to support custom events.

### DIFF
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -98,6 +98,8 @@ export interface EventTrigger {
    * This field is ignored for v1 and defaults to the
    */
   serviceAccountEmail?: string;
+
+  channel?: string;
 }
 
 /** Something that has an EventTrigger */
--- a/src/deploy/functions/runtimes/discovery/v1alpha1.ts
+++ b/src/deploy/functions/runtimes/discovery/v1alpha1.ts
@@ -126,6 +126,7 @@ function parseEndpoints(
         retry: "boolean",
         region: "string",
         serviceAccountEmail: "string",
+        channel: "string",
       });
       triggered = { eventTrigger: ep.eventTrigger };
       for (const eventFilter of triggered.eventTrigger.eventFilters) {
--- a/src/deploy/functions/runtimes/node/parseTriggers.ts
+++ b/src/deploy/functions/runtimes/node/parseTriggers.ts
@@ -10,8 +10,11 @@ import * as proto from "../../../../gcp/proto";
 import * as args from "../../args";
 import * as runtimes from "../../runtimes";
 import * as v2events from "../../../../functions/events/v2";
+import { normalizeRulesConfig } from "../../../../database/rulesConfig";
 
 const TRIGGER_PARSER = path.resolve(__dirname, "./triggerParser.js");
+const DEFAULT_CHANNEL_REGION = "us-central1";
+const CHANNEL_NAME_REGEX = /^(projects\/([^/]+)\/)?locations\/([^/]+)\/channels\/([^/]+)$/;
 
 export interface ScheduleRetryConfig {
   retryCount?: number;
@@ -50,12 +53,11 @@ export interface TriggerAnnotation {
   };
   eventTrigger?: {
     eventType: string;
-    resource: string;
-    channelId?: string;
-    channelLocation?: string;
+    resource?: string;
+    channel?: string;
     filters?: Record<string, string>;
     // Deprecated
-    service: string;
+    service?: string;
   };
   taskQueueTrigger?: {
     rateLimits?: {
@@ -115,7 +117,7 @@ function parseTriggers(
 
     parser.on("message", (message) => {
       if (message.triggers) {
-        console.log(message.triggers)
+        console.log(message.triggers);
         resolve(message.triggers);
       } else if (message.error) {
         reject(new FirebaseError(message.error, { exit: 1 }));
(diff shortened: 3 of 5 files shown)


src/deploy/functions/backend.ts
```diff
@@ -80,7 +80,7 @@ export interface EventTrigger {
    * V2 will have arbitrary filters and some EventArc filters will be
    * top-level keys in the GCF API (e.g. "pubsubTopic").
    */
-  eventFilters: EventFilter[];
+  eventFilters?: EventFilter[];
```
taeold, line 83: I don't see the code where we don't initialize `eventFilters` with an empty list - is this change really necessary?

src/deploy/functions/runtimes/node/parseTriggers.ts
```diff
@@ -216,19 +221,27 @@ export function addResourcesToBackend(
...
+          annotation?.eventTrigger?.channel
+        );
+      }
+
+      if (annotation.eventTrigger!.resource) {
+        triggered.eventTrigger.eventFilters!.push({
+          attribute: "resource",
+          value: annotation.eventTrigger!.resource,
+        });
+      }
```
taeold, line 240: IIUC, custom events are only supported in Cloud Functions Gen2.

We no longer use `parseTrigger` to read trigger definitions for CF3v2 deploys - we exclusively rely on container contract to do it. See https://github.com/firebase/firebase-tools/blob/master/src/deploy/functions/runtimes/node/index.ts#L127-L141.

This code should probably be written in `src/deploy/functions/runtimes/discovery/v1alpha1.ts` instead.

I suspect this change worked and you were able to deploy your custom event fns from your local dev machine... My guess is that either

1) your locally built tarball specifies an old version of the SDK
2) you haven't enabled cloudfunctionsv2 preview flag

taeold (changes requested):
IIUC, this change depends on changes to container contract changes.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -119,7 +121,7 @@ function parseEndpoints(
   for (const region of ep.region || [defaultRegion]) {
     let triggered: backend.Triggered;
     if (backend.isEventTriggered(ep)) {
-      requireKeys(prefix + ".eventTrigger", ep.eventTrigger, "eventType", "eventFilters");
+      requireKeys(prefix + ".eventTrigger", ep.eventTrigger, "eventType");
```
taeold, line 124: I don't think we don't have to relax `eventFilters` being an optional field anymore?

src/deploy/functions/runtimes/node/parseTriggers.ts
pavelgj: Done!

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -130,6 +132,13 @@ function parseEndpoints(
         channel: "string",
       });
       triggered = { eventTrigger: ep.eventTrigger };
+      if (typeof triggered.eventTrigger.channel !== "undefined") {
+        triggered.eventTrigger.channel = resolveChannelName(
+          project,
+          triggered.eventTrigger.channel,
+          defaultRegion
+        );
+      }
```
taeold, line 141: *nit - we could use `proto.renameIfPresent` with `converter` option here instead, but I think that only saves us like 5 key strokes 🤷🏼‍♂️

src/deploy/functions/runtimes/node/parseTriggers.ts
```diff
@@ -10,8 +10,11 @@ import * as proto from "../../../../gcp/proto";
 import * as args from "../../args";
 import * as runtimes from "../../runtimes";
 import * as v2events from "../../../../functions/events/v2";
+import { normalizeRulesConfig } from "../../../../database/rulesConfig";
```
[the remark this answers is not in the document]
pavelgj: Done.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -4,6 +4,8 @@ import { copyIfPresent } from "../../../../gcp/proto";
 import { assertKeyTypes, requireKeys } from "./parsing";
 import { FirebaseError } from "../../../../error";
 
+const CHANNEL_NAME_REGEX = /^(projects\/([^/]+)\/)?locations\/([^/]+)\/channels\/([^/]+)$/;
```
taeold, line 7: nit* We found that using named capturing group to help with readability of regex parsing, e.g.

https://github.com/firebase/firebase-tools/blob/19d3e24b488b4b96bc4428284648f17404b0146d/src/gcp/secretManager.ts#L9-L14

(Although it only really helps with readability if you know about the feature lol).

It feels slightly less brittle than using `match[0]` indexing imo.

src/gcp/cloudfunctionsv2.ts
```diff
@@ -509,6 +514,9 @@ export function endpointFromFunction(gcfFunction: CloudFunction): backend.Endpoi
         trigger.eventTrigger.eventFilters[attribute] = value;
       }
     }
+    if (gcfFunction.eventTrigger.channel) {
+      trigger.eventTrigger.channel = gcfFunction.eventTrigger.channel;
+    }
```
taeold, line 519:
```suggestion
    proto.copyIfPresent(trigger.eventTrigger, gcfFunction.eventTrigger, "channel");
```

pavelgj: Done.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
pavelgj: Done.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
pavelgj: Done.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
pavelgj: Right, reverted.

taeold (approved):

Issue #4372, opened 2022-03-30
Error: incorrect header check when downloading from storage emulator


dermasmid:
### [REQUIRED] Environment info

**firebase-tools:** 10.5.0

**Platform:** Ubuntu

### [REQUIRED] Test case

download any file from storage emulator using https://github.com/googleapis/nodejs-storage

```js
import {  Storage } from '@google-cloud/storage';
const storage = new Storage({ projectId: process.env.GCLOUD_PROJECT });
storage.bucket('bucketName').file('filename').download();
```
### [REQUIRED] Steps to reproduce

1. upload a file to storage emulator
2. try to download using the nodejs stroage package

### [REQUIRED] Expected behavior

should be able to download files from emulator with the @google-cloud/storage package

### [REQUIRED] Actual behavior

downlaod fails with Error: incorrect header check

this bug was intoduced here: https://github.com/firebase/firebase-tools/commit/9795181587cc39812e1c3fbe9852aa1ef0bf6840

weilinzung:
same issue

PR #4373, opened 2022-03-30
fix: incorrect header check fixes: #4372

dermasmid:
### Description
fixing #4372


tonyjhuang:
Thanks for the PR! LGTM, can you add an integration test?

tonyjhuang (approved):

PR #4379, opened 2022-03-30
Introduce IamBinder to abstract away work required to add new IAM bindings

taeold:
Today, Firebase CLI does lots of bookeepings to ensure all necessary IAM bindings exists for user-defined CF3v1/v2 functions. Sadly, the work to get-update-set new IAM policy for a resource is scattered throughout the codebase.

Here's an attempt to consolidate all IAM policy update logic. To use `IamBinder`, you'd make calls like this:

```js
  const iamBinder = new IamBinder(projectId, projectNumber);
  await iamBinder.addEndpoints(backend.allEndpoints(matchingBackend));
  await iamBinder.apply();
```

In `iamBinder.apply()`, IAM policy for all affected policies will be looked up and issue an update to IAM policy only if the existing policy does not include the bindings we need.

Some improvements are intentionally not implemented here:

1) Maybe issue testIamPolicy in case the principal has permission to perform the specific operation because it has higher-powered IAM binding (e.g. service account is an EDITOR which effectively gives it read/write access to most resources).

2) Manage policy deletion as well as additions. We don't really have a usecase for this atm.

I haven't hooked this logic into our deployment code - would like to get a review of the approach here before I go ahead and migrate/delete bunch of code.

### DIFF
--- a/src/deploy/functions/iambinder.ts
+++ b/src/deploy/functions/iambinder.ts
@@ -1,0 +1,298 @@
+import { createHash } from "crypto";
+
+import * as iam from "../../gcp/iam";
+import * as backend from "./backend";
+import * as storage from "../../gcp/storage";
+import * as v2Events from "../../functions/events/v2";
+import { assertExhaustive } from "../../functional";
+import { Client, ClientResponse } from "../../apiv2";
+
+type ResourceManagerResource = `//cloudresourcemanager.googleapis.com/${string}`;
+type SecretManagerResource = `//secretmanager.googleapis.com/${string}`;
+type Resource = ResourceManagerResource | SecretManagerResource;
+
+function isResourceManagerResource(r: Resource): r is ResourceManagerResource {
+  return r.startsWith("//cloudresourcemanager.googleapis.com/");
+}
+function isSecreteManagerResource(r: Resource): r is SecretManagerResource {
+  return r.startsWith("//secretmanager.googleapis.com/");
+}
+
+/* @internal */
+/**
+ * Construct URL components by splitting up full resource name.
+ */
+export function splitResource(r: Resource): { service: string; resource: string } | undefined {
+  const match = /\/\/(?<service>[^/]+)\/(?<resource>.+)/.exec(r);
+  if (!match?.groups) {
+    return undefined;
+  }
+  return { service: match.groups.service, resource: match.groups.resource };
+}
+
+export type Role = `roles/${string}`;
+
+function isRole(s: string): s is Role {
+  return s.startsWith("roles/");
+}
+
+interface IamBinding extends Omit<iam.Binding, "members" | "role"> {
+  role: Role;
+  members: Set<string>;
+}
+
+/* @internal */
+export class IamBindings {
+  additions: Record<Role, Record<string, IamBinding>>; // role -> condition -> bindings
+
+  constructor() {
+    this.additions = {};
+  }
+
+  add(role: Role, members: string[], condition?: iam.Binding["condition"]) {
+    const conditions = this.additions[role] || {};
+    const conditionKey = condition ? IamBindings.getConditionKey(condition) : "";
+    const binding = conditions[conditionKey] || { role, members: new Set() };
+    if (condition) {
+      binding.condition = { ...condition };
+    }
+
+    for (const member of members) {
+      binding.members.add(member);
+    }
+
+    conditions[conditionKey] = binding;
+    this.additions[role] = conditions;
+  }
+
+  diff(base: IamBindings): IamBindings {
+    const diff = this.clone();
+    for (const { role, members, condition } of base.allBindings()) {
+      const conditionKey = condition ? IamBindings.getConditionKey(condition) : "";
+      if (diff.additions[role]?.[conditionKey]) {
+        for (const member of members) {
+          diff.additions[role][conditionKey].members.delete(member);
+        }
+        if (diff.additions[role][conditionKey].members.size === 0) {
+          delete diff.additions[role][conditionKey];
+        }
+        if (Object.keys(diff.additions[role]).length === 0) {
+          delete diff.additions[role];
+        }
+      }
+    }
+    return diff;
+  }
+
+  merge(other: IamBindings): IamBindings {
+    const merged = new IamBindings();
+    for (const bindings of [this, other]) {
+      for (const binding of bindings.allBindings()) {
+        merged.add(binding.role, Array.from(binding.members), binding.condition);
+      }
+    }
+    return merged;
+  }
+
+  clone(): IamBindings {
+    const copied = new IamBindings();
+    for (const binding of this.allBindings()) {
+      copied.add(binding.role, Array.from(binding.members), binding.condition);
+    }
+    return copied;
+  }
+
+  asIamBindings(): iam.Binding[] {
+    const iamBindings: iam.Binding[] = [];
+    for (const binding of this.allBindings()) {
+      const copy: iam.Binding = { role: binding.role, members: Array.from(binding.members) };
+      if (binding.condition) {
+        copy.condition = { ...binding.condition };
+      }
+      iamBindings.push(copy);
+    }
+    return iamBindings;
+  }
+
+  private *allBindings(): Generator<IamBinding> {
+    for (const role of Object.keys(this.additions) as Role[]) {
+      for (const conditionKey of Object.keys(this.additions[role])) {
+        yield this.additions[role][conditionKey];
+      }
+    }
+  }
+
+  private static getConditionKey(condition: iam.Binding["condition"]): string {
+    return createHash("md5").update(JSON.stringify(condition)).digest("base64");
+  }
+
+  static fromIamBindings(bindings: iam.Binding[]): IamBindings {
+    const bindingInternal = new IamBindings();
+    for (const { role, members, condition } of bindings) {
+      if (!isRole(role)) {
+        throw new Error(`Invalid role. Role ${role} must be prefixed with "roles/"`);
+      }
+      bindingInternal.add(role, members, condition);
+    }
+    return bindingInternal;
+  }
+}
+
+/**
+ * IamBinder generates necessary IAM binding given different types of function endpoints.
+ */
+export class IamBinder {
+  additions: Record<Resource, IamBindings>; // resource -> bindings
+  storageServiceAccount: string | undefined;
+
+  constructor(private readonly projectId: string, private readonly projectNumber: string) {
+    this.additions = {};
+  }
+
+  async apply(): Promise<iam.Policy[]> {
+    const updates = Object.entries(this.additions).map(async ([resource, bindings]) => {
+      return await IamBinder.updatePolicy(resource as Resource, bindings);
+    });
+    return Promise.all(updates);
+  }
+
+  async addEndpoints(endpoints: backend.Endpoint[]): Promise<void> {
+    for (const endpoint of endpoints) {
+      await this.addEndpoint(endpoint);
+    }
+  }
+
+  private async addEndpoint(endpoint: backend.Endpoint): Promise<void> {
+    let serviceAccount = this.defaultComputeServiceAccount(endpoint);
+    if (endpoint.serviceAccountEmail && endpoint.serviceAccountEmail !== "default") {
+      serviceAccount = `serviceAccount:${endpoint.serviceAccountEmail}`;
+    }
+
+    for (const secret of endpoint.secretEnvironmentVariables || []) {
+      this.addBinding(
+        `//secretmanager.googleapis.com/projects/${this.projectNumber}/secrets/${secret.secret}`,
+        "roles/secretmanager.secretAccessor",
+        [serviceAccount]
+      );
+    }
+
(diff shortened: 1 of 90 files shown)


src/deploy/functions/iambinder.ts
```diff
@@ -0,0 +1,298 @@
...
+  additions: Record<Resource, IamBindings>; // resource -> bindings
+  storageServiceAccount: string | undefined;
+
+  constructor(private readonly projectId: string, private readonly projectNumber: string) {
+    this.additions = {};
+  }
+
+  async apply(): Promise<iam.Policy[]> {
+    const updates = Object.entries(this.additions).map(async ([resource, bindings]) => {
+      return await IamBinder.updatePolicy(resource as Resource, bindings);
```
colerogers, line 154: Are we now calling `getIamPolicy` & `setIamPolicy` for each binding addition? Or did I misunderstand this?

src/deploy/functions/iambinder.ts
```diff
@@ -0,0 +1,298 @@
...
+        );
+        // Service account associated w/ eventarc trigger must have permission to invoke a Cloud Run instance.
+        // We don't allow users to specify this service account hence it will always be the default compute service
+        // account. Usually, the default service account has EDITOR role which includes permission to invoke a Cloud
+        // Run instance, but given strict org policy it may not have this permission.
+        // See https://cloud.google.com/compute/docs/access/service-accounts#default_service_account.
+        this.addBinding(
+          `//cloudresourcemanager.googleapis.com/projects/${this.projectNumber}`,
+          "roles/run.invoker",
+          [this.defaultComputeServiceAccount(endpoint)]
```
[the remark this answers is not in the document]
taeold: From https://cloud.google.com/eventarc/docs/roles-permissions#triggertypes-roles, it looked like we only need that for Cloud Audit Logs which we don't support yet. Did we discover during bugbash that this binding is needed anyway?

src/deploy/functions/iambinder.ts
```diff
@@ -0,0 +1,298 @@
...
+        );
```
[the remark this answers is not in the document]
taeold: Sigh I feel pretty sad about that - in theory, this limitation applies to any service-specific service account that's created when we enable services.

src/deploy/functions/iambinder.ts
```diff
@@ -0,0 +1,298 @@
...
+        if (v2Events.isStorageTriggered(endpoint)) {
+          // Storage account associated w/ Cloud Storage must have permission to publish pubsub messages.
+          this.addBinding(
```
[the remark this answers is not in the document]
taeold: Huh I assumed that service abstraction is roughly "grouping endpoints by event types (into "service" grouping)" which I hoped to emulate here. Do you think I'm missing something here?

src/deploy/functions/iambinder.ts
taeold: This is meant to iterate over all "resources" - e.g. secret manager secret. Adding roles to service account is bit of an outlier in that the "resource" is the GCP project itself.

src/deploy/functions/iambinder.ts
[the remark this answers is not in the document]
colerogers: I guess this is a more general question about our UX. @inlined - Should we try to get the project to the correct state by enabling role permissions manually rather than erroring and having the user re-try?

src/deploy/functions/iambinder.ts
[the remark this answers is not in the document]
colerogers: I was just thinking out loud haha. I think I'm okay with doing that we're doing here since `storage` on v2 seems like the only special permissions case that we have. The `service` abstraction was originally created to avoid a bunch of `if` checks and instead just do a `noop` per @inlined 's direction

src/deploy/functions/iambinder.ts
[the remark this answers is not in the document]
taeold: I think Cloud Audit Log used to serve as a shim to receive events from other cloud resources (like storage) until there is direct integration. E.g. Cloud Storage now has direct integration - so will other event providers we expect to support soon.

I'm not sure if there is harm adding this permission - just working off the principle of least privilege here.

src/deploy/functions/iambinder.ts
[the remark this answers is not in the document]
taeold: Ah I see what you mean. I like the explicitness of bunch of if statements, but I can see it being unwieldy when there is TOO much of that 🤔 . I hope this list isn't getting that much bigger, but if this is more confusing and you think service abstraction makes more sense, I can bring that back here.

src/deploy/functions/iambinder.ts
taeold: oh resource string _is_ different per iteration. e.g.

```js
{
  '//cloudresourcemanager.googleapis.com/projects/12345': [
    { role: "roles/someRole", members: ["a", "b"] },
    { role: "roles/someRole", members: ["a", "b"] },
  ],

  '//secretmanager.googleapis.com/projects/12345/secrets/MY_SECRET': [
    { role: "roles/secretAccessor", members: ["c"] },
  ],

  '//secretmanager.googleapis.com/projects/12345/secrets/ANOTHER_SECRET': [
    { role: "roles/secretAccessor", members: ["d"] },
  ],

  // ... in the future we may add support for other resources e.g.
 '//cloudfunctions.googleapis.com/projects/12345/locations/us-central1/functions/myFunc': [
    { role: "roles/invoker", members: ["e", "f"] },
  ]
}
```

I'm playing on the idea that there is distinct IAM policy per GCP resource (referred by its [full resource name[(https://cloud.google.com/iam/docs/full-resource-names)).

I was having fun w/ types to help identify different types of resource we support (see `ResourceManagerResource`, `SecretManagerResource`, and `Resource` type definitions) but I'm getting the feeling that this is maybe too "clever"?

src/deploy/functions/iambinder.ts
```diff
@@ -0,0 +1,298 @@
...
+        yield this.additions[role][conditionKey];
+      }
+    }
+  }
+
+  private static getConditionKey(condition: iam.Binding["condition"]): string {
+    return createHash("md5").update(JSON.stringify(condition)).digest("base64");
+  }
+
+  static fromIamBindings(bindings: iam.Binding[]): IamBindings {
```
colerogers, line 129: I think this function name should be singular - `fromIamBinding` - since we're converting from the `iam.Binding` array, right?

src/deploy/functions/iambinder.ts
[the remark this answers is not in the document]
colerogers: I'm okay with keeping this as a one-off check since it seems like storage is a special case, we can revisit if other providers pop up in the future.

src/deploy/functions/iambinder.ts
```diff
@@ -0,0 +1,298 @@
...
+    const resp = await IamBinder.clientReq<{ policy: iam.Policy }, iam.Policy>(
+      "POST",
+      resource,
+      ":setIamPolicy",
+      { policy }
+    );
+    return resp.body;
+  }
+
+  private static clientReq<ReqT, ResT>(
```
colerogers, line 277: Since we usually have all of our GCP api calls contained in the `gcp` folder, this kinda seems wrong to me. But I get that we need some special logic to figure out the correct API endpoint, maybe we could keep that here but call out to the resource manager & secret manager APIs?

colerogers (commented):
just a couple more questions

src/deploy/functions/iambinder.ts
taeold: I'm expecting bindings from `policy.bindings` which is plural. I _think_ I'd refer to an array of `iam.Binding` in plural form - is that what we do in other places 🤔 ?

src/deploy/functions/iambinder.ts
taeold: Hmm I see your point.

I was trying to abuse the pattern on that there is always going to be `setIamPolicy` and `getIamPolicy` endpoints for each GCP resources for which the URL can be fully constructed by the full resource name.

The fact that I need to sprinkle on API version and that HTTP method differs for `getIamPolicy` for secret manager and resource manager API ends up making this more complicated than I hoped :(. Going to see what I can do to cleanup here.

src/deploy/functions/iambinder.ts
```diff
@@ -0,0 +1,298 @@
...
+
+/* @internal */
+export class IamBindings {
+  additions: Record<Role, Record<string, IamBinding>>; // role -> condition -> bindings
+
+  constructor() {
+    this.additions = {};
+  }
+
+  add(role: Role, members: string[], condition?: iam.Binding["condition"]) {
```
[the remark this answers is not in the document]
taeold: I don't know why our eslinter is so strict about adding return type for `void` 🤷🏼

src/deploy/functions/iambinder.ts
```diff
@@ -0,0 +1,286 @@
...
+
+type ResourceManagerResource = `//cloudresourcemanager.googleapis.com/projects/${string}`;
+type SecretManagerResource = `//secretmanager.googleapis.com/projects/${string}/secrets/${string}`;
+type Resource = ResourceManagerResource | SecretManagerResource;
+
+function isResourceManagerResource(r: Resource): r is ResourceManagerResource {
+  return r.startsWith("//cloudresourcemanager.googleapis.com/");
+}
+function isSecreteManagerResource(r: Resource): r is SecretManagerResource {
+  return r.startsWith("//secretmanager.googleapis.com/");
```
inlined, line 19: I feel a bit iffy about this approach. What if IAM binder took the resource address as a parameter?

src/deploy/functions/iambinder.ts
```diff
@@ -0,0 +1,286 @@
...
+}
+
+/* @internal */
+/**
+ * Construct URL components by splitting up full resource name.
+ */
+export function splitResource(r: Resource): { service: string; resource: string } | undefined {
+  const match = /\/\/(?<service>[^/]+)\/(?<resource>.+)/.exec(r);
+  if (!match?.groups) {
```
inlined, line 28: Is this ever a valid case? Should we just throw here?

src/deploy/functions/iambinder.ts
inlined: An array of `foo` is a `foos`.

src/deploy/functions/iambinder.ts
inlined: I think that it might be too clever. Maybe we can just take an API and a resource name.

src/deploy/functions/iambinder.ts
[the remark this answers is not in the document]
inlined: If possible we should succeed on first deploy.

src/deploy/functions/iambinder.ts
```diff
@@ -0,0 +1,286 @@
...
+      return `serviceAccount:${this.projectId}@appspot.gserviceaccount.com`;
+    } else if (endpoint.platform === "gcfv2") {
+      return `serviceAccount:${this.projectNumber}-compute@developer.gserviceaccount.com`;
+    }
+    assertExhaustive(endpoint.platform);
+  }
+
+  static async updatePolicy(resource: Resource, bindings: IamBindings): Promise<iam.Policy> {
+    const existingPolicy = await IamBinder.getPolicy(resource);
+    const existingBindings = IamBindings.fromIamBindings(existingPolicy.bindings || []);
```
inlined, line 236: Nit. If this is the design pattern you'll always use you can probably just make `getPolicy` return an `IamBindings`

src/functions/events/v2.ts
```diff
@@ -13,3 +15,48 @@ export type Event =
   | typeof PUBSUB_PUBLISH_EVENT
   | typeof STORAGE_EVENTS[number]
   | typeof FIREBASE_ALERTS_PUBLISH_EVENT;
+
+type EventFunctionBase = backend.EventTriggered & { platform: backend.FunctionsPlatform };
+type EventFunction = EventFunctionBase & { platform: "gcfv2" };
+
+type PubsubFunction = EventFunction & {
+  eventTrigger: Omit<EventFunction["eventTrigger"], "eventType"> & {
```
inlined, line 23: IDK if you have to actually specify an omit

inlined (commented):
I have mixed feelings about this. When I said I wanted a utility i meant something like:

```
const existing = await functions.getIamPolicy(endpoint);
// optionally, have addPublicBinding and addServiceAccount to simplify
// not using "serviceAccount:" as a prefix
iam.addBinding(existing, "allUsers", functions.INVOKER_ROLE);
await functions.setIamPolicy(existing);
```

I think it may be a feature, not a bug, that IAM code is sprinkled throughout. If we have all the concerns for updating a resource in that resource, then a future contributor will see that IAM is a concern and copy from that code. If IAM is in a giant monolith, then contributors may miss this. I'm also worried by the gymnastics you're already pulling to make this function handle all resource types. Also, by moving IAM into its own utility you are possibly going to take the IAM update out of the executor loop which means we either don't retry failures or we have redundant API calls after any failure. If the failure is coming from API request limits we're screwed too.

PR #4380, opened 2022-03-30
Redirect console links in postinstalls to emulator UI

joehan:
### Description
Redirect console links in postinstalls to emulator UI. This covers storage, Firestore, database, functions, extensions and auth.

### Scenarios Tested
Verified that a link to the Firestore console was redirected to the emulator UI (you can see the link target in the lower left corner)
Screen Shot 2022-03-30 at 3 53 10 PM

### DIFF
--- a/src/emulator/extensions/postinstall.ts
+++ b/src/emulator/extensions/postinstall.ts
@@ -1,0 +1,38 @@
+import { EmulatorRegistry } from "../registry";
+import { Emulators } from "../types";
+
+export function replaceConsoleLinks(postinstall: string): string {
+  const uiInfo = EmulatorRegistry.getInfo(Emulators.UI);
+  const uiUrl = uiInfo ? `http://${EmulatorRegistry.getInfoHostString(uiInfo)}` : "unknown";
+  let subbedPostinstall = postinstall;
+  const linkReplacements = new Map<RegExp, string>([
+    [
+      /(http[s]?:\/\/)?console\.firebase\.google\.com\/project\/[A-Za-z0-9-]+\/storage[A-Za-z0-9\/-]*(?=[\)\]\s])/,
+      `${uiUrl}/storage`,
+    ], // Storage console links
+    [
+      /(http[s]?:\/\/)?console\.firebase\.google\.com\/project\/[A-Za-z0-9-]+\/firestore[A-Za-z0-9\/-]*(?=[\)\]\s])/,
+      `${uiUrl}/firestore`,
+    ], // Firestore console links
+    [
+      /(http[s]?:\/\/)?console\.firebase\.google\.com\/project\/[A-Za-z0-9-]+\/database[A-Za-z0-9\/-]*(?=[\)\]\s])/,
+      `${uiUrl}/database`,
+    ], // RTDB console links
+    [
+      /(http[s]?:\/\/)?console\.firebase\.google\.com\/project\/[A-Za-z0-9-]+\/authentication[A-Za-z0-9\/-]*(?=[\)\]\s])/,
+      `${uiUrl}/auth`,
+    ], // Auth console links
+    [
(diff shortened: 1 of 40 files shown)


src/emulator/extensions/postinstall.ts
```diff
@@ -0,0 +1,38 @@
+import { EmulatorRegistry } from "../registry";
+import { Emulators } from "../types";
+
+export function replaceConsoleLinks(postinstall: string): string {
```
elvisun, line 4: lint nit: Missing JSDoc comment

src/emulator/extensions/postinstall.ts
```diff
@@ -0,0 +1,38 @@
...
+  const uiInfo = EmulatorRegistry.getInfo(Emulators.UI);
+  const uiUrl = uiInfo ? `http://${EmulatorRegistry.getInfoHostString(uiInfo)}` : "unknown";
+  let subbedPostinstall = postinstall;
+  const linkReplacements = new Map<RegExp, string>([
+    [
+      /(http[s]?:\/\/)?console\.firebase\.google\.com\/project\/[A-Za-z0-9-]+\/storage[A-Za-z0-9\/-]*(?=[\)\]\s])/,
```
elvisun, line 10: lint nit: Unnecessary escape character: \/

src/emulator/extensions/postinstall.ts
```diff
@@ -0,0 +1,38 @@
...
+      /(http[s]?:\/\/)?console\.firebase\.google\.com\/project\/[A-Za-z0-9-]+\/database[A-Za-z0-9\/-]*(?=[\)\]\s])/,
+      `${uiUrl}/database`,
+    ], // RTDB console links
+    [
+      /(http[s]?:\/\/)?console\.firebase\.google\.com\/project\/[A-Za-z0-9-]+\/authentication[A-Za-z0-9\/-]*(?=[\)\]\s])/,
+      `${uiUrl}/auth`,
+    ], // Auth console links
+    [
+      /(http[s]?:\/\/)?console\.firebase\.google\.com\/project\/[A-Za-z0-9-]+\/functions[A-Za-z0-9\/-]*(?=[\)\]\s])/,
+      `${uiUrl}/logs`,
```
elvisun, line 27: is there a way we can keep the ui repo and this in sync? so a route change doesn't break us.

an IfThisThenThat lint could work except they aren't in the same repo... 

maybe throw a comment in the emulatorUI repo?

src/test/emulators/extensions/postinstall.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  const tests: {
+    desc: string;
+    input: string;
+    expected: string;
+  }[] = [
+    {
+      desc: "should replace Firestore links",
+      input:
+        " Go to your [Cloud Firestore dashboard](https://console.firebase.google.com/project/test-project/firestore/data) in the Firebase console.",
```
elvisun, line 28: another common pattern is 
`https://console.firebase.google.com/u/0/project/test-project/firestore/data` if there's multiple GAIA users logged in. Can you add a test case for that?

elvisun (approved):
Can you also confirm with Will to see if we need to add any UI affordance to show that we've changed the link? (If it's not too hard to do)

src/emulator/extensions/postinstall.ts
joehan: I ended up refactoring this to be more similar to the way we generate links for the table that is printed out at the end of emulators:start. Hopefully that makes them a little less likely to break, since those links are very obvious + well established.

src/emulator/extensions/postinstall.ts
joehan: / does need to be escaped. Otherwise it indicates the end of the regex.

PR #4381, opened 2022-03-30
time is now a number. Cleaned up code

inlined:
Tester reported a bug where the functions SDK is reporting "timeoutSeconds" (number) from the container contract but we're expecting "timeout" (Duration).

This bug changes the v1beta1.Manifest parsing to expect the correct timeoutSeconds. Since the container contract and GCFv2 use timeoutSeconds, I switched backend.Endpoint to use timeoutSeconds. This required me to look into the emulator code...

In the emulator code it became clear that timeout wasn't clearly typed (code was defensive for either a string or an int). I fixed this to always be a number inside the emulator code, but then had to address the extensions manifest parsing. That was using an `any` type so I added some type annotations. The type annotations require a literal type which meant I had to add annotations to a test and those annotations where getting lost with `_.cloneDeep`. Since lodash is deprecated I made my own `cloneDeep` function and moved lodash to use it. And I used that opportunity to clean up lodash from the affected files...

... Oh... and `expect` needs to be awaited if comparing against a promise or the test will start failing if the promise doesn't resolve within the same tick of the runloop...

... I really let this one snowball. Sorry.


joehan (approved):
One note about the implementation of cloneDeep, but overall LGTM. Thanks for taking the time to clean up so much code!

src/utils.ts
```diff
@@ -638,3 +638,38 @@ export function groupBy<T, K extends string | number | symbol>(
...
+    clone[k] = cloneDeep(v);
+  }
+  return clone as T;
+}
+
+/**
+ * replacement for lodash cloneDeep that preserves type.
+ */
+// TODO: replace with builtin once Node 18 becomes the min version.
+export function cloneDeep<T>(obj: T): T {
```
taeold, line 658: ah yes I'm definitely going to be using this.

taeold (approved):

Issue #4341, opened 2022-03-19
Duplicate access token on file.setMetadata


Expipiplusone:
### Environment info

Firebase CLI 10.1.2 
Using emulators for functions, storage, firestore
Ubuntu 21.10

### Test case

Context: setting new file metadata via (node.js) function.
Example code:
```
  const fileRef = bucket.file( storagePath );
  const fileMetadata = ( await fileRef.getMetadata() )[0];
  
  // do something with metadata
  
  await fileRef.setMetadata( { metadata: newMetadata } );
```

### Steps to reproduce

Invoke function multiple times

### Expected behavior

New metadata should be set, access tokens should not be duplicated (see below)

### Actual behavior

New metadata are set as expected. However, in addition to that, a new access token (identical to the one already present) is added to the file, as can be seen in the emulator suite console under File location.
![Screenshot from 2022-03-19 02-23-44](https://user-images.githubusercontent.com/33549645/159101246-59e494e0-718c-47b2-996a-c15505b63237.png)

**If the file has N access tokens, they are all duplicated.**

**After a few repetitions, the file has so many access tokens that the function goes in error for `PayloadTooLargeError: request entity too large`.**

Tried on actual firebase (not emulator), no duplication occurs.

tonyjhuang:
@abhis3 can you take a look at this?

abhis3:
Ack, I'll take a look into it

PR #4383, opened 2022-03-31
Fix download token duplication on multiple setMetadata calls

abhis3:
### Description

This fixes a bug (#4341) where download tokens are duplicated when `setMetadata` is called multiple times in a row. This happened because we always concatenated the user-provided and existing token arrays and never enforced uniqueness.

### DIFF
--- a/scripts/storage-emulator-integration/tests.ts
+++ b/scripts/storage-emulator-integration/tests.ts
@@ -1107,7 +1107,7 @@ describe("Storage emulator", () => {
 
           // Check that the tokens are saved in Cloud metadata
           const [storedMetadata] = await cloudFile.getMetadata();
-          expect(storedMetadata.metadata.firebaseStorageDownloadTokens).to.deep.equal(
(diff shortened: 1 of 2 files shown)


scripts/storage-emulator-integration/tests.ts
```diff
@@ -1075,6 +1075,43 @@ describe("Storage emulator", () => {
...
+            .then((res) => {
+              const firebaseMd = res.body;
+              expect(firebaseMd.downloadTokens).to.equal(
+                incomingMetadata.metadata.firebaseStorageDownloadTokens
+              );
+            });
+
+          // Check that the tokens are saved in Cloud metadata
+          const [storedMetadata] = await cloudFile.getMetadata();
+          expect(storedMetadata.metadata.firebaseStorageDownloadTokens).to.deep.equal(
```
tohhsinpei, line 1110: Probably don't need to check deep equality since it's a string?

tohhsinpei (commented):
Seems reasonable. Would it be a good idea to make `this.downloadTokens` a set?

tonyjhuang (approved):

PR #4384, opened 2022-03-31
Update TaskQueue retry config to match SDK/container contract options

taeold:
At some point, I assume we updated name of TQ retry config options that ends with `*Duration` into `*Seconds`.


colerogers (approved):
lgtm

inlined (commented):
LGTM. Good catch.

PR #4358, opened 2022-03-24
Fix issue with importing Storage Emulator data with nested directory structure

tohhsinpei:
#### Issue
This PR fixes #4326.

Importing any Storage data exported prior to v10.3.0 causes the emulator to crash.

#### Cause
In #3647 (released as part of v10.3.0), we flattened the emulator directory structure. However, this is not backwards-compatible with previously exported data, which still contains the nested directory structure. In particular, when we copy the imported data, we copy over the old nested directories.

### Scenarios Tested
Tested against [example](https://github.com/charlesfries/firebase-tools-repro) provided in issue and verified that crash does not occur and that we are able to view the file:
repro-fixed

Also wrote two integration tests, one testing import of old nested structure, another testing import of new flattened structure. These will be run as part of `npm run test:storage-emulator-integration`.

Issue #4326, opened 2022-03-18
Firebase Storage Emulator crashes unexpectedly since v10.3.0


DevSchmidtchen:
### [REQUIRED] Environment info

**firebase-tools:** 10.4.0

**Platform:** macOS Monterey (12.3)

### [REQUIRED] Test case

- Set up an app with storage emulator and import the existing emulator data (`firebase emulators:start --import ./emulator_data`)

### [REQUIRED] Steps to reproduce

1. Start local emulators including storage emulator (`firebase emulators:start --import ./emulator_data`)2. 
2. Open Emulator UI and go to storage tab

### [REQUIRED] Expected behavior

The storage contents are shown and you can browse through them and upload new files. This works using version `10.2.2` of firebase emulator but since the update it doesn't work anymore.

### [REQUIRED] Actual behavior

The emulator crashes.

```
[debug] [2022-03-18T09:40:32.246Z] Error: ENOENT: no such file or directory, open '/var/folders/pn/tjh55fz97z34qp9h8q6w72bw0000gn/T/firebase/storage/blobs/artistconnect-app.appspot.com%2Fwatermark.wav'
    at openSync (node:fs:585:3)
    at Persistence.readBytes (/Users/matti/.nvm/versions/node/v16.14.0/lib/node_modules/firebase-tools/lib/emulator/storage/persistence.js:37:36)
    at StorageLayer.getBytes (/Users/matti/.nvm/versions/node/v16.14.0/lib/node_modules/firebase-tools/lib/emulator/storage/files.js:98:38)
    at StorageLayer.handleGetObject (/Users/matti/.nvm/versions/node/v16.14.0/lib/node_modules/firebase-tools/lib/emulator/storage/files.js:83:49)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async /Users/matti/.nvm/versions/node/v16.14.0/lib/node_modules/firebase-tools/lib/emulator/storage/apis/firebase.js:74:35
[error] 
[error] Error: An unexpected error has occurred.
```

AlexVFornazieri:
Hey guys,

I have a similar issue here, in my case happen when clone project in a fresh new folder, in PC or in a pipeline CI/CD process. But in my case, not uploaded any files in my emulated storage eat. So the "storage_export/blob" and "storage_export/metadata" folder are empty and git don't track then. It cause a erro that don't find that directories in storage emulator startup.

[...]

Issue #4322, opened 2022-03-17
Storage emulator import/export data error


rulaman123:
Hi, after starting the storage emulator with import the ui only shows a white page and i get that error:

```
Failed to load resource: the server responded with a status of 404 (Not Found)
2react-dom.production.min.js:225 FirebaseError: Firebase Storage: Object '...' does not exist. (storage/object-not-found)
```

### [REQUIRED] Environment info

**firebase-tools:** v10.4.0

**Platform:** Windows

### [REQUIRED] Steps to reproduce

Upload image to storage emulator. 
Export image and close emulator.
Open storage emulator with import data.
Error.

xero88:
same issue for me, I throught it was related to my M1, but it's seams not. 

On my old mac (where it's ok) : firebase-cli version is 10.2.1
On my new mac (where it's bugged) : firebase-cli version is 10.4.0

ermartinez13:
I've been running into the same issue when I went from version 10.2.1 to 10.4.1. I downgraded to version 10.3.1 and found the bug still present. Only when I went back to v10.2.1 did the bug disappear.

torstenek:
Adding some more info:

### [REQUIRED] Environment info

I can reproduce this on 
- MB Pro M1 Max 
- MP Pro 2019

### [REQUIRED] Steps to reproduce

1. Start with an empty $FIREBASE_EMULATOR_DATA directory. 

2. Start the emulator with 

`firebase emulators:start --project myproject -only 'storage' --import $FIREBASE_EMULATOR_DATA --export-on-exit`

Now upload, for example, an image `myfile.jpg` to the bucket (for example by using the web UI). The problem is the same if you upload it via the firebase dart package. 

3. Stop the emulator gracefully and note the "Export complete" log entry indicating that everything went OK. 

4. Start the emulators again

`firebase emulators:start --project myproject -only 'storage' --import $FIREBASE_EMULATOR_DATA --export-on-exit`

Note the log message

`Could not find file "myproject%252Fmyfile.jpg" in storage export. `

5. Verify that the file is indeed missing from the storage emulator.

torstenek:
The sequence %252F in the filename could be the result of a double URI encoding.

tonyjhuang:
Potential duplicate of https://github.com/firebase/firebase-tools/issues/4326

sceee:
Jumping in here as I'm facing the same issue and have more information to add:
[...]

PR #4387, opened 2022-03-31
Fix duplicate import for Storage Emulator

tohhsinpei:
### Description
Follow up to #4358. With this PR we should be able to close out #4326 and #4322.

#### Issue
Previously, I made a mistake in writing the new import logic, causing a duplicate metadata file to be written to the `blobs` directory. You can see this when you export the data again:
duplicate-import
Note the presence of the `test_upload.jpg.json` file; we expect to see only the `test_upload.jpg` file.

#### Cause
I thought that we needed to copy over the metadata as a file on disk when importing the emulator data. This is false; we only need to copy over its contents into the `StoredFile`. Only on export do we turn the metadata into a JSON file, which goes into the `metadata`, not `blobs`, directory.

### Scenarios Tested
Added an integration test checking that the imported emulator data contains only the expected file blobs. As before, we do this by checking the exported data.

### DIFF
--- a/scripts/storage-emulator-integration/import/tests.ts
+++ b/scripts/storage-emulator-integration/import/tests.ts
@@ -1,4 +1,6 @@
 import * as path from "path";
+import * as fs from "fs-extra";
+import { expect } from "chai";
 import supertest = require("supertest");
 
 import { createTmpDir } from "../../../src/test/emulators/fixtures";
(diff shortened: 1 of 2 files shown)


tonyjhuang (approved):
Good catch, thanks for the fix

tonyjhuang:
Did this bug have any practical impact or was it just that the files saved to the local file system was incorrect?

tohhsinpei:
> Did this bug have any practical impact or was it just that the files saved to the local file system was incorrect?

The latter, I think.

PR #4388, opened 2022-03-31
Container contract updates

taeold:
Partial revert of https://github.com/firebase/firebase-tools/pull/4240 where we transformed EventFilters to list of arrays. We now prefer to keep the object structure.

To support path-pattern operator on event filters, we add new eventFilterPathPattern property. While we are at it, we also adds channel property to event trigger definition.

Note that most of the affected changes are in test files :/

Relevant SDK change: https://github.com/firebase/firebase-functions/pull/1070

### DIFF
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -48,7 +48,7 @@ export interface CallableTriggered {
   callableTrigger: CallableTrigger;
 }
 
-type EventFilterKey = "resource" | "topic" | "bucket" | "alerttype" | string;
+type EventFilterKey = "resource" | "topic" | "bucket" | "alerttype" | "appid" | string;
 
 /** API agnostic version of a Cloud Function's event trigger. */
 export interface EventTrigger {
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -48,7 +48,7 @@ export interface CallableTriggered {
   callableTrigger: CallableTrigger;
 }
 
-type EventFilterKey = "resource" | "topic" | "bucket" | string;
+type EventFilterKey = "resource" | "topic" | "bucket" | "alerttype" | string;
 
 /** API agnostic version of a Cloud Function's event trigger. */
 export interface EventTrigger {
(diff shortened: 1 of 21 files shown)


src/deploy/functions/release/planner.ts
```diff
@@ -196,10 +196,7 @@ export function changedV2PubSubTopic(want: backend.Endpoint, have: backend.Endpo
   if (have.eventTrigger.eventType !== v2events.PUBSUB_PUBLISH_EVENT) {
     return false;
   }
-
-  return (
-    backend.findEventFilter(have, "topic")?.value !== backend.findEventFilter(want, "topic")?.value
-  );
+  return have.eventTrigger.eventFilters.topic !== want.eventTrigger.eventFilters.topcic;
```
colerogers, line 199: typo - `topcic` to `topic`

src/deploy/functions/backend.ts
```diff
@@ -48,16 +48,7 @@ export interface CallableTriggered {
...
-
-// One or more event filters restrict the set of events delivered to an EventTrigger.
-interface EventFilter {
-  attribute: EventFilterAttribute;
-  value: string;
-
-  // if left unspecified, equality is used.
-  operator?: "match-path-pattern";
-}
+type EventFilterKey = "resource" | "topic" | "bucket" | "alerttype" | string;
```
colerogers, line 51: while we're here, we can also add `appid` for alerts that use it

colerogers (changes requested):
couple things, then we're g2g

colerogers (approved):

src/deploy/functions/backend.ts
```diff
@@ -48,7 +48,7 @@ export interface CallableTriggered {
   callableTrigger: CallableTrigger;
 }
 
-type EventFilterKey = "resource" | "topic" | "bucket" | string;
+type EventFilterKey = "resource" | "topic" | "bucket" | "alerttype" | string;
```
inlined: I wonder if it's silly to keep this type.

src/gcp/cloudfunctionsv2.ts
```diff
@@ -485,6 +477,9 @@ export function functionFromEndpoint(endpoint: backend.Endpoint, source: Storage
   return gcfFunction;
 }
 
+/**
+ *
```
inlined, line 481: wat

inlined (commented):
Woof. Lots of work all over again. TY for this.

PR #4391, opened 2022-04-01
firebase-frameworks integration

jamesdaniels:
Serve & deploy, no init experience yet—depending on timing I might opt for that in a follow on.

[go/framework-aware-firebase-tools](http://go/framework-aware-firebase-tools)
[b/227694244](https://buganizer.corp.google.com/issues/227694244)

Will build the web app, deploy the static content to hosting, and the backing Cloud Function (gen2). 

Right now we can't deploy alongside any other Cloud Functions, if a Cloud Function is needed, until #4419 lands. The work-around is to call `firebase deploy --only hosting` which will still deploy the backing function.

* Add `firebase-frameworks` dep
* Require and call out to `firebase-frameworks`, if `hosting.source` option is present
* Fix for deploying GCFv2 rewrites for the first time
  * ~~Ensure that functions are always acted on before hosting at each step~~
  * Remove any rewrites that refer to a gen2 function being deployed before creating the version
  * Patch the rewrites back in when finalizing the version
* Allow developers to refer to CF3v2 functions in rewriting using `function` and `region`, rather than `run: {...}`
* Refactor `deploy/index.js` into Typescript, forgive the anys

TODO:
- [x] Clean up the gen2 rewrite solution a bit, perhaps I should extract that into a seperate PR?
- [x] Add some quick test coverage
- [x] Support multiple hosting targets in firebase-frameworks
- [x] Add handling for the `--preview` flag
- [x] Add types where I can, rather than using any everywhere

Confirmed working with Next.js, Nuxt, Nuxt3, and custom web frameworks. [See the other side of the integration on the firebase-frameworks repo](https://github.com/FirebaseExtended/firebase-framework-tools/blob/main/src/index.ts#L23).

[Tarball attached for Googlers](https://drive.google.com/file/d/1aE_5y317WYausgma57AAQcKEbdo9vVJb/view?usp=sharing&resourcekey=0-fPkaS_7x9C8aVcHc7vldnw).

FYI @inlined @bkendall @mbleigh


jamesdaniels:
FYI @inlined first pass on the GCFv2 chicken/egg problem in 6d8ec34

src/deploy/hosting/prepare.ts
```diff
@@ -9,7 +9,7 @@ import * as deploymentTool from "../../deploymentTool";
 /**
  *  Prepare creates versions for each Hosting site to be deployed.
  */
-export async function prepare(context: any, options: any): Promise<void> {
+export async function prepare(context: any, options: any, payload: any): Promise<void> {
```
joehan, line 12: If possible, it would be great to give these real types while you're here. 

options has a generic type that can be used for almost any command (defined in options.ts).
 In other deploy flows, we've defined types for Context/Payload as well (see https://github.com/firebase/firebase-tools/blob/master/src/deploy/extensions/args.ts#L3 and https://github.com/firebase/firebase-tools/blob/master/src/deploy/functions/args.ts#L9)

This also has the added benefit of silencing the many, many lint warnings that pop up in this file during reviews.

src/deploy/hosting/prepare.ts
```diff
@@ -44,6 +44,22 @@ export async function prepare(context: any, options: any): Promise<void> {
       labels: deploymentTool.labels(),
     };
 
+    // Keep a copy of the data in the payload, to patch in later
+    payload.hosting ||= {};
```
joehan, line 48: TIL about ||=  😄

src/deploy/hosting/prepare.ts
```diff
@@ -44,6 +44,22 @@ export async function prepare(context: any, options: any): Promise<void> {
...
+    payload.hosting.config ||= {};
+    payload.hosting.config[deploy.site] = { ...data.config };
+
+    // Filter out any rewrites that point to GCFv2 functions that are being deployed
+    Object.entries(payload.functions?.backend?.endpoints || {}).forEach(([region, endpoints]) => {
```
joehan, line 53: Strongly prefer `for.. of...` over forEach - go/tsstyle#iterating-containers - applies for both of these loops

src/deploy/hosting/prepare.ts
```diff
@@ -44,6 +44,22 @@ export async function prepare(context: any, options: any): Promise<void> {
...
+      Object.entries(endpoints as {}).forEach(([serviceId, endpoint]) => {
+        if ((endpoint as any).platform === "gcfv2") {
+          data.config.rewrites = data.config.rewrites.filter(
+            (rewrite: any) => rewrite.run?.serviceId !== serviceId && rewrite.run?.region !== region
```
joehan, line 57: The typing inside this loops seems a bit code smelly to me - I don't like that we are casting to {} then to any. I'd prefer to give this some strong types so someone unfamiliar with the code can easily know what to expect here.

src/deploy/hosting/release.ts
```diff
@@ -6,7 +6,7 @@ import * as utils from "../../utils";
 /**
  *  Release finalized a Hosting release.
  */
-export async function release(context: any, options: any): Promise<void> {
+export async function release(context: any, options: any, payload: any): Promise<void> {
```
joehan, line 9: Same comment as prepare - adding types here would be great.

src/deploy/index.js
```diff
@@ -53,36 +53,54 @@ var deploy = function (targetNames, options, customContext = {}) {
...
-      );
+  return new Promise((resolve) => {
+    if (targetNames.includes("hosting") && options.config.get("hosting.source")) {
+      resolve(require("firebase-frameworks").prepare(targetNames, context, options));
+    } else {
+      resolve();
     }
+  })
+    .then(function () {
+      // Ensure that functions predeploy & deploy gets called before hosting that way
```
joehan, line 64: This seems like a bit of a roundabout way to do things. AFAICT, the order for these is defined here, https://github.com/firebase/firebase-tools/blob/master/src/commands/deploy.js#L15 and it looks like functions is already before hosting. 

Is there some case that I'm missing where we cant just rely on that ordering?

src/previews.ts
```diff
@@ -12,6 +12,7 @@ interface PreviewFlags {
   deletegcfartifacts: boolean;
   artifactregistry: boolean;
   emulatoruisnapshot: boolean;
+  frameworkawareness: boolean;
```
joehan, line 15: You added this preview flag but afaict, it isn't referenced anywhere in the code. Should we be checking this flag before we `await require('firebase-frameworks')` in serve/index.ts and deploy/index.ts?

joehan (changes requested):

src/deploy/index.js
jamesdaniels: Ah I see, it's firebase-frameworks that's just pushing onto the targets array. I'll fix that and drop the ordering force.

src/previews.ts
jamesdaniels: Yeah that's what I'm thinking, will also use in the init experience but I haven't taken a pass on that yet.

src/deploy/hosting/prepare.ts
jamesdaniels: Yeah, just fighting Object.entries and any, strong types will help a lot.

src/deploy/hosting/prepare.ts
jamesdaniels: Ended up ditching both loops, since I could just solve in a single filter pass on the config... 🤦

joehan (approved):
Code LGTM, but like you mentioned in the TODOs. better test coverage would be great. If you are able to add that, rerequest from me and i'll review those too/

yuchenshi (approved):

src/deploy/hosting/release.ts
```diff
@@ -17,11 +18,12 @@ export async function release(context: any, options: any): Promise<void> {
   await Promise.all(
     context.hosting.deploys.map(async (deploy: any) => {
       utils.logLabeledBullet(`hosting[${deploy.site}]`, "finalizing version...");
-      const finalizeResult = await client.patch(
-        `/${deploy.version}`,
-        { status: "FINALIZED" },
-        { queryParams: { updateMask: "status" } }
-      );
+
+      const config = convertConfig(deploy.config);
```
inlined, line 22: This is expanding scope from your previous work, but I wonder if we should also make `convertConfig` take the functions payload and detect that a function rewrite was a GCFv2 function and turn it into a Run rewrite. That'll fix the edge case where people must know to use the service rewrite for gen 2 functions until July.

src/deploy/index.js
```diff
@@ -53,36 +54,49 @@ var deploy = function (targetNames, options, customContext = {}) {
...
 
-  for (var i = 0; i < targetNames.length; i++) {
-    var targetName = targetNames[i];
-    var target = TARGETS[targetName];
-
-    if (!target) {
-      return Promise.reject(
-        new FirebaseError(clc.bold(targetName) + " is not a valid deploy target", { exit: 1 })
-      );
+  return new Promise((resolve) => {
```
inlined, line 57: Can we use modern TypeScript and just make `deploy` an async function?

inlined (approved):
I don't feel like I innately get what's going on, but to the best of my understanding this LGTM. Has firebase-frameworks been peer reviewed as well or was that a solo effort?

src/deploy/hosting/release.ts
jamesdaniels: I don't think that's a big lift from what I've already built here, happy to take a swing.

jamesdaniels:
@joehan added some tests, please take a look.

@inlined modified `convertConfig` to convert `function: ...` to `run: { ... }` automatically for GCFv2 endpoints (both existing and pending) and rewrote `deploy/index` into Typescript, as you suggested.

joehan (approved):
Tests LGTM

jamesdaniels:
FYI @taeold looks like your fix to the deploy refactor wasn't complete, the functions v1&2 groups might not exist. Adding a nullish on this branch.

PR #4395, opened 2022-04-03
Add blocking trigger type

colerogers:
Adds deployment code for blocking trigger types

```
```

Requirements:
- allows at most one blocking event type in a deployment
- reduces the trigger options of all triggers to the OR of each trigger
- does not overwrite identity platform options on function updates
- does not overwrite identity platform URIs on delete if endpoint.uri !== resource uri
- sets the invoker to allUsers on create & update


src/gcp/identityPlatform.ts
```diff
@@ -0,0 +1,234 @@
...
+    config.blockingFunctions = {
+      triggers: {
+        beforeCreate: {
+          functionUri: "",
+        },
+        beforeSignIn: {
+          functionUri: "",
+        },
+      },
+      forwardInboundCredentials: {},
```
taeold, line 173: Per typing of `BlockingFunctionsConfig`, every property is optional. I think most our code assumes that these property may not exists. Why do we go about initializing an empty object?

taeold (changes requested):

src/deploy/functions/backend.ts
```diff
@@ -131,6 +131,17 @@ export interface TaskQueueTriggered {
   taskQueueTrigger: TaskQueueTrigger;
 }
 
+export interface BlockingTrigger {
+  eventType: string;
+  accessToken?: boolean;
+  idToken?: boolean;
+  refreshToken?: boolean;
```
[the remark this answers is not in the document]
colerogers: Yep that's correct, these variables are actually booleans. Here's the explanation that identity platform gives
```
Select OAuth token credentials to be passed into the attached Cloud Function.
```

src/deploy/functions/release/fabricator.ts
```diff
@@ -54,6 +55,7 @@ export interface FabricatorArgs {
 const rethrowAs =
   <T>(endpoint: backend.Endpoint, op: reporter.OperationType) =>
   (err: unknown): T => {
+    logger.error((err as Error).message);
```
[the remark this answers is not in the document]
colerogers: So I noticed this during testing, by re-throwing we don't actually log out the internal error that happened. This makes it hard to troubleshoot and can help with debugging customer errors in the future

src/deploy/functions/release/fabricator.ts
```diff
@@ -248,6 +250,13 @@ export class Fabricator {
           })
           .catch(rethrowAs(endpoint, "set invoker"));
       }
+    } else if (backend.isBlockingTriggered(endpoint)) {
+      // set invoker to all users
```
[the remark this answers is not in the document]
colerogers: 👍

src/deploy/functions/release/fabricator.ts
```diff
@@ -317,6 +326,11 @@ export class Fabricator {
...
```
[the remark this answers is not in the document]
colerogers: 👍

src/deploy/functions/release/fabricator.ts
```diff
@@ -354,6 +368,8 @@ export class Fabricator {
       invoker = endpoint.httpsTrigger.invoker;
     } else if (backend.isTaskQueueTriggered(endpoint)) {
       invoker = endpoint.taskQueueTrigger.invoker;
+    } else if (backend.isBlockingTriggered(endpoint)) {
+      invoker = ["public"];
```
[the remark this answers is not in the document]
colerogers: Yeah the out-of-band rule makes sense. I'm specifically violating it in the Auth Blocking case since the `setTrigger` will fail on non-public functions. I figured that a failed deployment should be a big enough thing to "bend" this rule. Let me know if my logic makes sense here

src/deploy/functions/services/authBlocking.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  if (
+    endpoint.blockingTrigger.eventType === events.v1.BEFORE_CREATE_EVENT ||
+    endpoint.blockingTrigger.eventType === events.v2.BEFORE_CREATE_EVENT
+  ) {
+    blockingConfig.triggers = {
+      beforeCreate: {
+        functionUri: endpoint.uri!,
+      },
+      beforeSignIn: blockingConfig.triggers?.beforeSignIn,
```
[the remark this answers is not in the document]
colerogers: Yep, that looks like a better solution!

src/deploy/functions/services/authBlocking.ts
```diff
@@ -0,0 +1,149 @@
...
+  endpoint.blockingTrigger.refreshToken =
+    wantBackend.resourceOptions.identityPlatform?.refreshToken || false;
+}
+
+/**
+ * Registers the auth blocking trigger to identity platform. On updates, we don't touch the options.
+ * @param endpoint the blocking endpoint
+ * @param update if this registration is an update
+ */
+export async function registerAuthBlockingTriggerToIdentityPlatform(
```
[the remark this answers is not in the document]
colerogers: SGTM I'll shorten this and the un-register function

src/deploy/functions/services/authBlocking.ts
```diff
@@ -0,0 +1,149 @@
...
+    };
+  } else {
+    throw new FirebaseError("Invalid auth blocking trigger type.");
+  }
+
+  if (!update) {
+    blockingConfig.forwardInboundCredentials = {
+      idToken: (endpoint.blockingTrigger.idToken || false).toString(),
+      accessToken: (endpoint.blockingTrigger.accessToken || false).toString(),
+      refreshToken: (endpoint.blockingTrigger.refreshToken || false).toString(),
```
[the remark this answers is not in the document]
colerogers: Yeah I'm preserving out-of-band resource changes

src/deploy/functions/services/authBlocking.ts
```diff
@@ -0,0 +1,149 @@
...
+  endpoint: backend.Endpoint & backend.BlockingTriggered
+): Promise<void> {
+  const blockingConfig = await identityPlatform.getBlockingFunctionsConfig(endpoint.project);
+  if (
+    endpoint.blockingTrigger.eventType === events.v1.BEFORE_CREATE_EVENT ||
+    endpoint.blockingTrigger.eventType === events.v2.BEFORE_CREATE_EVENT
+  ) {
+    if (
+      !blockingConfig.triggers?.beforeCreate?.functionUri ||
+      endpoint.uri !== blockingConfig.triggers.beforeCreate.functionUri
```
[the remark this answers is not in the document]
colerogers: 👍

src/deploy/functions/services/authBlocking.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+
+/**
+ * Un-registers the auth blocking trigger from identity platform. If the endpoint uri is not on the resource, we do nothing.
+ * @param endpoint the blocking endpoint
+ */
+export async function unregisterAuthBlockingTriggerFromIdentityPlatform(
+  endpoint: backend.Endpoint & backend.BlockingTriggered
+): Promise<void> {
+  const blockingConfig = await identityPlatform.getBlockingFunctionsConfig(endpoint.project);
```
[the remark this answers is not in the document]
colerogers: 👍

src/deploy/functions/services/authBlocking.ts
```diff
@@ -0,0 +1,149 @@
...
+    endpoint.blockingTrigger.eventType === events.v2.BEFORE_CREATE_EVENT
+  ) {
+    if (
+      !blockingConfig.triggers?.beforeCreate?.functionUri ||
+      endpoint.uri !== blockingConfig.triggers.beforeCreate.functionUri
+    ) {
+      return;
+    }
+    blockingConfig.triggers = {
+      beforeCreate: {},
```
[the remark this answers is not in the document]
colerogers: 👍

src/deploy/functions/services/index.ts
```diff
@@ -1,25 +1,40 @@
...
   readonly api: string;
 
   // dispatch functions
   requiredProjectBindings:
     | ((projectNumber: string, policy: iam.Policy) => Promise<Array<iam.Binding>>)
     | undefined;
   ensureTriggerRegion: (ep: backend.Endpoint & backend.EventTriggered) => Promise<void>;
+  ensureTriggerIsValid: (
+    ep: backend.Endpoint & backend.BlockingTriggered,
+    want: backend.Backend
```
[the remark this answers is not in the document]
colerogers: Sure, I'll change it

src/gcp/cloudfunctions.ts
```diff
@@ -4,16 +4,36 @@ import { FirebaseError } from "../error";
...
+import { BEFORE_CREATE_EVENT, BEFORE_SIGN_IN_EVENT } from "../functions/events/v1";
 
 export const API_VERSION = "v1";
 const client = new Client({ urlPrefix: functionsOrigin, apiVersion: API_VERSION });
 
+const BLOCKING_LABEL = "deployment-blocking";
+
+const BLOCKING_LABEL_KEY_TO_EVENT: Record<
+  string,
+  typeof BEFORE_CREATE_EVENT | typeof BEFORE_SIGN_IN_EVENT
```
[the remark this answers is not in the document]
colerogers: 👍

src/gcp/cloudfunctions.ts
```diff
@@ -487,6 +507,16 @@ export function endpointFromFunction(gcfFunction: CloudFunction): backend.Endpoi
...
+  } else if (gcfFunction.labels?.[BLOCKING_LABEL]) {
+    // trigger = inferAuthBlockingDetailsFromFunction(gcfFunction, additionalDetailsCache);
+    trigger = {
+      blockingTrigger: {
+        eventType: BLOCKING_LABEL_KEY_TO_EVENT[gcfFunction.labels[BLOCKING_LABEL]],
+      },
+    };
+
+    uri = gcfFunction.httpsTrigger!.url;
+    securityLevel = gcfFunction.httpsTrigger!.securityLevel;
```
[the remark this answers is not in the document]
colerogers: That won't work since we'd be reassigning trigger to be an httpsTrigger for the label functions and overwrite what's currently there

src/deploy/functions/prepare.ts
```diff
@@ -186,6 +187,7 @@ export function inferDetailsFromExisting(
   usedDotenv: boolean
 ): void {
   for (const wantE of backend.allEndpoints(want)) {
+    serviceForEndpoint(wantE).copyResourceOptionsToEndpoint(wantE as any, want);
```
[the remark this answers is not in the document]
colerogers: Ya we have to cast it to any since we have the function signature as `ep: backend.Endpoint & backend.BlockingTriggered`

src/gcp/identityPlatform.ts
colerogers: Good point, I'll remove it

src/deploy/functions/prepare.ts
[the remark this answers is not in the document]
colerogers: You're correct, this is pretty brittle. Maybe I should combine these into one function? I was trying to not iterate over the endpoints another time since we already do it quite a bit

src/deploy/functions/backend.ts
```diff
@@ -131,6 +131,17 @@ export interface TaskQueueTriggered {
   taskQueueTrigger: TaskQueueTrigger;
 }
 
+export interface BlockingTrigger {
+  eventType: string;
```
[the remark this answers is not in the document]
colerogers: Yeah that's a good call, thanks!

src/deploy/functions/prepare.ts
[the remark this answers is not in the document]
colerogers: combined them into a single function that's invoked from the prepare stage

src/deploy/functions/services/index.ts
```diff
@@ -28,37 +43,60 @@ export const NoOpService: Service = {
...
 };
+
 /** A pubsub service object */
 export const PubSubService: Service = {
   name: "pubsub",
   api: "pubsub.googleapis.com",
   requiredProjectBindings: undefined,
   ensureTriggerRegion: noop,
+  ensureTriggerIsValid: noop,
+  copyResourceOptionsToEndpoint: noop,
```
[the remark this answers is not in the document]
colerogers: After moving the resource options code into it's own function, I was able to remove this Service function property. We're only left with `ensureTriggerIsValid` which seems like an okay function to have for every service.

src/deploy/functions/backend.ts
```diff
@@ -143,6 +157,8 @@ export function endpointTriggerType(endpoint: Endpoint): string {
     return endpoint.eventTrigger.eventType;
   } else if (isTaskQueueTriggered(endpoint)) {
     return "taskQueue";
+  } else if (isBlockingTriggered(endpoint)) {
+    return "blocking";
```
inlined, line 161: endpoint.blockingTrigger.eventType. It's much more interesting to know this is a beforeUserCreated event than that it's some generic blocking function.

src/deploy/functions/prepare.ts
```diff
@@ -271,3 +272,29 @@ function maybeCopyTriggerRegion(wantE: backend.Endpoint, haveE: backend.Endpoint
   }
   wantE.eventTrigger.region = haveE.eventTrigger.region;
 }
+
+/** Figures out the blocking endpoint options by taking the OR of every trigger option and reassigning that value back to the endpoint. */
```
inlined, line 276: Why is this the correct behavior? If I have a beforeUserCreated function I must have the same security options as a beforeUserLoggedIn function?

src/deploy/functions/release/fabricator.ts
```diff
@@ -21,6 +21,7 @@ import * as reporter from "./reporter";
 import * as run from "../../../gcp/run";
 import * as scheduler from "../../../gcp/cloudscheduler";
 import * as utils from "../../../utils";
+import * as authBlocking from "../services/authBlocking";
```
inlined, line 24: Are we really going to have different files for auth and authBlocking?

src/deploy/functions/release/fabricator.ts
```diff
@@ -248,6 +250,13 @@ export class Fabricator {
           })
           .catch(rethrowAs(endpoint, "set invoker"));
       }
+    } else if (backend.isBlockingTriggered(endpoint)) {
+      // Blocking functions should always be public
+      await this.executor
+        .run(async () => {
+          await gcf.setInvokerCreate(endpoint.project, backend.functionName(endpoint), ["public"]);
```
inlined, line 257: Here and elsewhere, it's `public` because it's an auth blocking function, not because it's a blocking function. Please make the code forwards compatible to other blocking trigger types.

src/deploy/functions/release/fabricator.ts
```diff
@@ -519,6 +541,15 @@ export class Fabricator {
     }
   }
 
+  async registerBlockingTrigger(
+    endpoint: backend.Endpoint & backend.BlockingTriggered,
+    update: boolean
+  ): Promise<void> {
+    await this.executor
+      .run(() => authBlocking.registerTrigger(endpoint, update))
```
inlined, line 549: Again, this is only appropriate if it's an auth blocking function.

src/deploy/functions/release/planner.ts
```diff
@@ -235,6 +235,8 @@ export function checkForIllegalUpdate(want: backend.Endpoint, have: backend.Endp
       return "a scheduled";
     } else if (backend.isTaskQueueTriggered(e)) {
       return "a task queue";
+    } else if (backend.isBlockingTriggered(e)) {
+      return "a blocking";
```
inlined, line 239: We should be probably using event type here. I didn't have event types for eventTriggers because I wasn't sure when/how we'll be able to upgrade from v1 types to v2 types.

src/deploy/functions/services/authBlocking.ts
```diff
@@ -0,0 +1,127 @@
...
+ * Registers the auth blocking trigger to identity platform. On updates, we don't touch the options.
+ * @param endpoint the blocking endpoint
+ * @param update if this registration is an update
+ */
+export async function registerTrigger(
+  endpoint: backend.Endpoint & backend.BlockingTriggered,
+  update: boolean
+): Promise<void> {
+  const newBlockingConfig = await identityPlatform.getBlockingFunctionsConfig(endpoint.project);
+  const oldBlockingConfig = { ...newBlockingConfig };
```
inlined, line 66: Since you're modifying `newBlockingConfig` and it's an interface, not a simple Record<string, string> I'd feel more comfortable if you used `utils.deepCopy`

src/deploy/functions/validate.ts
```diff
@@ -8,14 +8,16 @@ import * as fsutils from "../../fsutils";
...
 import * as utils from "../../utils";
 import * as secrets from "../../functions/secrets";
+import { serviceForEndpoint } from "./services";
 
 /** Validate that the configuration for endpoints are valid. */
 export function endpointsAreValid(wantBackend: backend.Backend): void {
-  functionIdsAreValid(backend.allEndpoints(wantBackend));
+  const endpoints = backend.allEndpoints(wantBackend);
+  functionIdsAreValid(endpoints);
+  endpoints.forEach((ep) => serviceForEndpoint(ep).validateTrigger(ep as any, wantBackend));
```
inlined, line 17: nit: google style guide forbids `forEach`. Just use `for( of )` instead.

src/functions/events/v2.ts
```diff
@@ -9,7 +9,14 @@ export const STORAGE_EVENTS = [
 
 export const FIREBASE_ALERTS_PUBLISH_EVENT = "google.firebase.firebasealerts.alerts.v1.published";
 
+export const BEFORE_CREATE_EVENT = "providers/cloud.auth/eventTypes/user.beforeCreate";
+
+export const BEFORE_SIGN_IN_EVENT = "providers/cloud.auth/eventTypes/user.beforeSignIn";
+
+export const AUTH_BLOCKING_EVENTS = [BEFORE_CREATE_EVENT, BEFORE_SIGN_IN_EVENT] as const;
```
inlined, line 16: IDK if I really love having this defined as a separate event type. I saw some places where only one type is handled and this also gets in the way of GCIP offering a new v2 api (not that they're super inclined to anyway).

In reality, events.v1 and events.v2 isn't about GCF versions as much as it is "eventflow" or "eventarc". It's OK that these event types will be handled by a v2 function and it's possible that we'll be forced to add support for all of events.v1 in v2 functions.

src/gcp/identityPlatform.ts
```diff
@@ -0,0 +1,221 @@
...
+ * Helper function to set the blocking function config to identity platform.
+ * @param project GCP project ID or number
+ * @param blockingConfig the blocking functions configuration to update
+ * @returns the blocking functions config
+ */
+export async function setBlockingFunctionsConfig(
+  project: string,
+  blockingConfig: BlockingFunctionsConfig
+): Promise<BlockingFunctionsConfig> {
+  const config =
```
inlined, line 188: Nit: Would this be easier if you just used `proto.fieldMasks` to create your field mask instead of setting it to "blockingFunctions"? Then you wouldn't need to merge both beforeCreate and beforeSignIn before setting it.

src/deploy/functions/services/authBlocking.ts
```diff
@@ -0,0 +1,127 @@
...
+  } else {
+    newBlockingConfig.triggers = {
+      ...newBlockingConfig.triggers,
+      beforeSignIn: {
+        functionUri: endpoint.uri!,
+      },
+    };
+  }
+
+  if (!update) {
```
inlined, line 84: Why don't we modify on updates? Is that not allowed?

src/test/deploy/functions/services/authBlocking.spec.ts
```diff
@@ -0,0 +1,527 @@
...
+      };
+      getConfig.resolves(blockingConfig);
+
+      await authBlocking.registerTrigger(ep, false);
+
+      expect(setConfig).to.not.have.been.called;
+    });
+  });
+
+  describe("unregisterAuthBlockingTriggerFromIdentityPlatform", () => {
```
inlined, line 372: You can't use spaces here?

inlined (changes requested):

src/deploy/functions/prepare.ts
colerogers: It's part of the trigger proposal - http://go/cf3-auth-blocking-triggers

Since these are resource level options, they must be the same. Instead of erroring if a user sets the `beforeCreate` & `beforeSignIn` options to different values, we just take the OR of them. This will need to be made very clear in our docs

src/deploy/functions/release/fabricator.ts
colerogers: Good catch, I'll rename it

src/gcp/identityPlatform.ts
colerogers: So funny thing about their API... if you try and use a field mask that has the inner property that changed, like `blockingFunctions.triggers.beforeCreate.functionUri`, it will not actually pickup the changes.

src/deploy/functions/services/authBlocking.ts
colerogers: I'm preserving out-of-band resource changes. That seemed to fit with our current model, let me know if you disagree

src/test/deploy/functions/services/authBlocking.spec.ts
colerogers: Oops, leftover from a previous name, I'll change it to `unregisterTrigger`

src/deploy/functions/services/authBlocking.ts
colerogers: Changed to `cloneDeep()`

src/deploy/functions/validate.ts
colerogers: 👍

src/functions/events/v2.ts
colerogers: Would you recommend just dropping v1 events entirely?

src/deploy/functions/backend.ts
```diff
@@ -131,6 +132,19 @@ export interface TaskQueueTriggered {
   taskQueueTrigger: TaskQueueTrigger;
 }
 
+export interface BlockingTrigger {
+  eventType:
+    | typeof events.v1.AUTH_BLOCKING_EVENTS[number]
+    | typeof events.v2.AUTH_BLOCKING_EVENTS[number];
+  accessToken?: boolean;
+  idToken?: boolean;
+  refreshToken?: boolean;
```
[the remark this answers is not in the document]
colerogers: @inlined How about we go with a record structure for options and for auth blocking triggers we'd assume they'd use these three options. So the trigger will now be
```
export interface BlockingTrigger {
  eventType:
    | typeof events.v1.AUTH_BLOCKING_EVENTS[number]
    | typeof events.v2.AUTH_BLOCKING_EVENTS[number];
  options: Record<string, any>;
}
```

What do you think?

src/deploy/functions/release/fabricator.ts
colerogers: Added a check
```
if (backend.isBlockingTriggered(endpoint) && AUTH_BLOCKING_EVENTS.includes(endpoint.blockingTrigger.eventType)) { ... }
```

src/deploy/functions/release/fabricator.ts
colerogers: Changed it to be
```
await this.executor
      .run(() => services.serviceForEndpoint(endpoint).registerTrigger(endpoint, update))
      .catch(rethrowAs(endpoint, "register blocking trigger"));
```

src/deploy/functions/release/fabricator.ts
```diff
@@ -545,9 +560,12 @@ export class Fabricator {
...
     update: boolean
   ): Promise<void> {
-    await this.executor
-      .run(() => authBlocking.registerTrigger(endpoint, update))
-      .catch(rethrowAs(endpoint, "register blocking trigger"));
+    this.triggerQueue = this.triggerQueue.then(async () => {
+      await this.executor
+        .run(() => services.serviceForEndpoint(endpoint).registerTrigger(endpoint, update))
+        .catch(rethrowAs(endpoint, "register blocking trigger"));
+    });
```
inlined, line 567: Why don't we make the `triggerQueue` be a feature of the Auth service? That way we run auth trigger updates in serial but can run other trigger updates in parallel.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -201,18 +201,9 @@ function parseEndpoints(
       requireKeys(prefix + ".blockingTrigger", ep.blockingTrigger, "eventType");
       assertKeyTypes(prefix + ".blockingTrigger", ep.blockingTrigger, {
         eventType: "string",
-        accessToken: "boolean",
```
inlined: Depending how thorough you want to be, you can detect the eventType is an auth blocking event type and then do the assertKeyTypes for the inner options object

src/deploy/functions/services/auth.ts
```diff
@@ -83,9 +82,9 @@ export async function registerTrigger(
 
   if (!update) {
```
inlined, line 83: I still don't think I understand why we only set these fields on function create instead of update.

inlined (approved):
Looking really good! I'd personally make the trigger queue more narrowly scoped (e.g. inside auth) so we can parallelize as much as possible and it's not clear to me why we don't update options when the trigger is updated. Otherwise LGTM.

src/deploy/functions/release/fabricator.ts
colerogers: I was trying to avoid defining a class for the Auth Blocking service, but parallelizing makes a lot of sense here, I'll make the change

src/deploy/functions/runtimes/discovery/v1alpha1.ts
colerogers: I ended up taking out the explicit eventType per your last review. The reason why I didn't compare each option is because other blockingTriggers will have a different set of options

src/test/deploy/functions/validate.spec.ts
```diff
@@ -198,6 +199,101 @@ describe("validate", () => {
         /they have fewer than 2GB memory/
       );
     });
+
+    it("Disallows multiple beforeCreate blocking", () => {
```
taeold, line 203: nit - prefer lowercase `disallow` (same comment for tests below).

src/test/deploy/functions/services/auth.spec.ts
```diff
@@ -0,0 +1,495 @@
...
+const authBlockingService = new auth.AuthBlockingService();
+
+describe("authBlocking", () => {
+  let getConfig: sinon.SinonStub;
+  let setConfig: sinon.SinonStub;
+
+  beforeEach(() => {
+    getConfig = sinon
+      .stub(identityPlatform, "getBlockingFunctionsConfig")
+      .throws("Unexpected call to getBlockingFunctionsConfig");
```
taeold, line 25: nit* for async functions, you'd use `reject`. I guess this works since `await` is forgiving 🤷🏼

src/test/deploy/functions/services/auth.spec.ts
```diff
@@ -0,0 +1,495 @@
...
+            refreshToken: false,
+          },
+        },
+      };
+      getConfig.resolves(blockingConfig);
+      setConfig.resolves(newBlockingConfig);
+
+      await authBlockingService.registerTrigger(ep);
+
+      expect(blockingConfig).to.deep.equal(newBlockingConfig);
```
taeold, line 204: huh I'm having a brain freeze, but how does this pass? `blockingConfig` and `newBlockingConfig` clearly looks different to me.

taeold (approved):
Few nitpicks, but overall looks good.

src/deploy/functions/services/auth.ts
```diff
@@ -3,122 +3,148 @@ import * as identityPlatform from "../../../gcp/identityPlatform";
...
+          functionUri: endpoint.uri!,
+        },
+      };
+    }
 
-  if (!update) {
     newBlockingConfig.forwardInboundCredentials = {
       idToken: endpoint.blockingTrigger.options?.idToken || false,
       accessToken: endpoint.blockingTrigger.options?.accessToken || false,
       refreshToken: endpoint.blockingTrigger.options?.refreshToken || false,
```
inlined, line 99: I like your intention to preserve OOB changes. I think what I would do is say something like

```
newBlockingConfig.forwardInboundCredentials = { ...oldBlockingConfig.forwardInboundCredentials, ...endpoint.blockingTrigger.options }
```

This would keep options that were set out of band unless it was specified in the customer's code. The `|| false` you have should generally be unnecessary because the API server will automatically fill zero values for any missing field.

src/deploy/functions/services/auth.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    await identityPlatform.setBlockingFunctionsConfig(endpoint.project, newBlockingConfig);
+  }
+
+  /**
+   * Registers the auth blocking trigger to identity platform.
+   * @param ep the blocking endpoint
+   */
+  registerTrigger(ep: backend.Endpoint & backend.BlockingTriggered): Promise<void> {
+    this.triggerQueue = this.triggerQueue.then(() => this.registerTriggerLocked(ep));
```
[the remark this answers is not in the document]
inlined: I pushed Cole to this implementation. The problem is that we update functions concurrently and it's possible that we update the beforeCreate function at the same time as beforeSignIn. Since the process of updating involves reading a whole config, changing a field, and saving a whole config, it's possible that two racing updates will read the same original config, change one field, and only one write will win.

By using a promise chain we guarantee that the racing updates will happen sequentially. Think of it as running inside a mutex.

src/deploy/functions/services/auth.ts
```diff
@@ -3,122 +3,148 @@ import * as identityPlatform from "../../../gcp/identityPlatform";
...
-    )
-  ) {
-    throw new FirebaseError(
-      `Can only create at most one Auth Blocking Trigger for ${endpoint.blockingTrigger.eventType} events`
-    );
+export class AuthBlockingService implements Service {
+  name: Name;
+  api: string;
+  triggerQueue: Promise<void>;
+  unregisterQueue: Promise<void>;
```
inlined, line 12: nit: unused var

src/deploy/functions/services/index.ts
```diff
@@ -77,26 +77,18 @@ export const FirebaseAlertsService: Service = {
...
   "google.cloud.pubsub.topic.v1.messagePublished": PubSubService,
   "google.cloud.storage.object.v1.finalized": StorageService,
   "google.cloud.storage.object.v1.archived": StorageService,
   "google.cloud.storage.object.v1.deleted": StorageService,
   "google.cloud.storage.object.v1.metadataUpdated": StorageService,
   "google.firebase.firebasealerts.alerts.v1.published": FirebaseAlertsService,
-  "providers/cloud.auth/eventTypes/user.beforeCreate": AuthBlockingService,
-  "providers/cloud.auth/eventTypes/user.beforeSignIn": AuthBlockingService,
+  "providers/cloud.auth/eventTypes/user.beforeCreate": authBlockingService,
+  "providers/cloud.auth/eventTypes/user.beforeSignIn": authBlockingService,
```
inlined, line 91: nit: the capitalization here confuses me. Are these Classes or variables?

src/test/deploy/functions/services/auth.spec.ts
inlined: Agreed. I think this is happening because we're mutating return values of blockingConfig? Which seems wrong here... Maybe `setConfig` should resolve w/o any return value (unless we depend on it) and we shoudl be verifying that `setConfig` was called with `newBlockingConfig`.

src/deploy/functions/backend.ts
```diff
@@ -3,6 +3,7 @@ import * as gcf from "../../gcp/cloudfunctions";
 import * as gcfV2 from "../../gcp/cloudfunctionsv2";
 import * as utils from "../../utils";
 import * as runtimes from "./runtimes";
+import * as events from "../../functions/events";
```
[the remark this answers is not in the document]
colerogers: Removed in #4443

src/deploy/functions/prepare.ts
```diff
@@ -271,3 +273,35 @@ function maybeCopyTriggerRegion(wantE: backend.Endpoint, haveE: backend.Endpoint
...
+export function inferBlockingDetails(want: backend.Backend): void {
+  const authBlockingEndpoints = backend
```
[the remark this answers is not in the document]
colerogers: Actually here I do a check in `filter` to see if the eventType is an Auth Blocking Trigger event

src/deploy/functions/release/fabricator.ts
```diff
@@ -519,6 +556,18 @@ export class Fabricator {
     }
   }
 
+  async registerBlockingTrigger(
+    endpoint: backend.Endpoint & backend.BlockingTriggered,
+    update: boolean
+  ): Promise<void> {
+    this.triggerQueue = this.triggerQueue.then(async () => {
```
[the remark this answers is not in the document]
colerogers: I think your checking out an old piece of code, but we're just adding the next task in the queue to the `then` function. When the current task is fulfilled, the next task will be executed.

src/deploy/functions/runtimes/node/parseTriggers.ts
```diff
@@ -68,6 +68,10 @@ export interface TriggerAnnotation {
     };
     invoker?: string[];
   };
+  blockingTrigger?: {
+    eventType: string;
+    options?: Record<string, any>;
```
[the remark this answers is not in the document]
colerogers: Fixed in #4443

src/deploy/functions/services/index.ts
```diff
@@ -1,64 +1,94 @@
...
-  requiredProjectBindings:
-    | ((projectNumber: string, policy: iam.Policy) => Promise<Array<iam.Binding>>)
-    | undefined;
+  requiredProjectBindings?: (
+    projectNumber: string,
+    policy: iam.Policy
+  ) => Promise<Array<iam.Binding>>;
   ensureTriggerRegion: (ep: backend.Endpoint & backend.EventTriggered) => Promise<void>;
+  validateTrigger: (
+    ep: backend.Endpoint & backend.BlockingTriggered,
```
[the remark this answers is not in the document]
colerogers: Fixed in #4443

src/test/deploy/functions/validate.spec.ts
colerogers: fixed in #4443

src/test/deploy/functions/services/auth.spec.ts
colerogers: fixed in #4443, we no longer check for a modification to the object

src/deploy/functions/services/auth.ts
colerogers: fixed in #4443

src/deploy/functions/services/auth.ts
colerogers: fixed in #4443

PR #4397, opened 2022-04-04
Fix metadata parse error response for multipart uploads in Storage Emulator

tohhsinpei:
#### Issue
Currently when invalid metadata is passed in a multipart upload for the Storage emulator, the response looks like this:
```
{"error":{"code":400,"message":{}}}
```

#### Context
I encountered this error recently while trying to make a new directory in the Emulator UI. This [UI fix](https://github.com/firebase/firebase-tools-ui/pull/727) should take care of it but we've not cut a new version yet I believe.


tonyjhuang:
Thanks for the cleanup :)

tonyjhuang (approved):

PR #4398, opened 2022-04-04
Adding extensions summary table

joehan:
### Description
Adding extensions summary table & refactoring extensions emulator into an EmulatorInstance.

This one spiraled on me a little bit. I first implemented the extensions table, then I realized I wanted it to print out at the end (and only on emulators-start). I then realized that because I didn't make this an EmulatorInstance in the first place, it was really hard to access it outside of controller.ts. I ended up adding stub methods to fulfill the EmulatorInstance type, so that I could make use of the EmulatorRegistry more fully. 

### Scenarios Tested
Running emulators:start with extensions:
Screen Shot 2022-04-04 at 3 43 09 PM

Without extensions:
Screen Shot 2022-04-04 at 4 42 11 PM

### DIFF
--- a/src/commands/emulators-start.ts
+++ b/src/commands/emulators-start.ts
@@ -7,6 +7,7 @@ import { Emulators, EMULATORS_SUPPORTED_BY_UI } from "../emulator/types";
 import * as clc from "cli-color";
 import { Constants } from "../emulator/constants";
 import { logLabeledWarning } from "../utils";
+import { ExtensionsEmulator } from "../emulator/extensionsEmulator";
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const Table = require("cli-table");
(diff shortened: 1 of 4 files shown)


src/commands/emulators-start.ts
```diff
@@ -94,7 +93,13 @@ module.exports = new Command("emulators:start")
...
         .filter((v) => v)
     );
-
+    let extensionsTable: string = "";
+    if (EmulatorRegistry.isRunning(Emulators.EXTENSIONS)) {
+      const extensionsEmulatorInstance = EmulatorRegistry.get(
+        Emulators.EXTENSIONS
+      ) as ExtensionsEmulator;
+      extensionsTable = extensionsEmulatorInstance.extensionsInfoTable(options);
+    }
```
elvisun, line 102: this lgtm, just some additional thoughts:

printing some post-startup instruction might not be an extension specific issue. wdyt of adding a `getInstruction` method to `EmulatorInstance`, where each `EmulatorInstance` would implement this. Then at the end of `emulators-start` command we can print the post-startup instruction for all running emulators. Could also propose a standardized way to print these instructions that all emulators follow.

elvisun (approved):

PR #4401, opened 2022-04-05
Adding fields to support local extensions in InstanceSpec and ManifestInstanceSpec

joehan:
### Description
Adding new fields to InstanceSpec and ManifestInstanceSpec to support local extensions, as well as some caching getters that will be useful.

### Scenarios Tested
I verified that all of the current flows still work:
- ext:install --local
- ext:update --local
- ext:configure --local
- emulators:start
- deploy
- ext:export
I also caught (and fixed) a few existing bugs here. Turns out we broke ext:export with locally installed extensions on a project  when we added support for secrets 😬 It should be fixed now


src/deploy/extensions/planner.ts
```diff
@@ -21,7 +22,12 @@ export interface ManifestInstanceSpec {
   instanceId: string;
   params: Record<string, ParamBindingOptions>;
   ref?: refs.Ref;
-  paramSpecs?: extensionsApi.Param[];
+  localPath?: string;
+  // Used by getExtensionVersion, getExtension, and getExtensionSpec.
+  // You should stronly prefer accessing via those methods
+  extensionVersion?: extensionsApi.ExtensionVersion;
+  extension?: extensionsApi.Extension;
```
elvisun, line 29: I'm leaning towards removing these two fields. There are many optional fields here and they are getting populated at different states with different uses of "optional". It can make things confusing pretty quickly - at what stage is what optional, should I be populating this field, etc... So there's benefit to keeping the properties of `ManifestInstanceSpec` to a minimum here.

what do you think of removing extensionVersion and extension here but keeping the util functions to retrieve them lazily if we need to. I tweaked `getExtensionVersion` slightly to make it work.

```
export async function getExtensionVersion(
  i: InstanceSpec | ManifestInstanceSpec
): Promise<extensionsApi.ExtensionVersion> {
  if ('extensionVersion' in i && !i.extensionVersion) {
    if (!i.ref) {
      throw new FirebaseError(
        `Can't get ExtensionVersion for ${i.instanceId} because it has no ref`
      );
    }
    i.extensionVersion = await extensionsApi.getExtensionVersion(refs.toExtensionVersionRef(i.ref));
    return i.extensionVersion;
  } else {
    if (!i.ref) {
      throw new FirebaseError(
        `Can't get ExtensionVersion for ${i.instanceId} because it has no ref`
      );
    }
    return await extensionsApi.getExtensionVersion(refs.toExtensionVersionRef(i.ref));
  }
}
```

elvisun (commented):
mostly lgtm just one minor thing

src/deploy/extensions/planner.ts
joehan: So, I ended up refactoring these types a bit further to extend a common base type before I saw this comment. LMK if you think this is more understandable/if we should switch back so we can strip some fields out of manifest instance spec. I think its a little cleaner overall this way, but you're definitely more familiar with the part of the code that uses manifestInstanceSpec

src/deploy/extensions/planner.ts
```diff
@@ -17,11 +30,8 @@ import { ParamBindingOptions } from "../../extensions/paramHelper";
...
  */
-export interface ManifestInstanceSpec {
-  instanceId: string;
+export interface ManifestInstanceSpec extends InstanceSpec {
   params: Record<string, ParamBindingOptions>;
-  ref?: refs.Ref;
-  paramSpecs?: extensionsApi.Param[];
 }
 
 // TODO(lihes): Rename this to something like DeploymentInstanceSpec.
```
elvisun, line 37: thanks for taking care of my todo :)

src/deploy/extensions/planner.ts
elvisun: ah I like this base type idea, that's pretty clean!

do you think pulling these 2 fields into DeploymentInstanceSpec make sense? because they aren't really used by the manifest
```
  extensionVersion?: extensionsApi.ExtensionVersion;
  extension?: extensionsApi.Extension;
```

joehan: I think ExtensionVersion is 'used', since we use getExtensionVersion under the hood in getExtensionSpec. We could remove it from the type & make the corresponding changes to getExtensionVersion, but i feel like it comes out uglier (since we end up needing to accept and handle DeploymentInstanceSpec | ManifestInstanceSpec, instead of being able to use the base type.)

src/deploy/extensions/planner.ts
joehan: np :) Removed this as well

src/deploy/extensions/planner.ts
elvisun: sg let's go with what you have here :) thanks a lot for exploring different options and cleaning up these types!

elvisun (approved):

PR #4402, opened 2022-04-05
Ignore empty values when writing to .env files

joehan:
### Description
While testing out some extensions, I noticed that we were writing empty values to the .env file. This lead to some awkward behavior, since it wasn't clear whether we should treat this as "" or undefined at deploy/emulation time. To clarify this, I'm changing this behavior to filter out empty optional params when we are writing to the manifest.

Note that there is no way to pass in "" to a required param, so this will only apply to optional params.

### Scenarios Tested
`ext:install firebase/storage-resize-images  --local` and verified that empty param values were not written to the .env file:
Screen Shot 2022-04-05 at 4 09 55 PM

### DIFF
--- a/src/extensions/manifest.ts
+++ b/src/extensions/manifest.ts
@@ -174,6 +174,7 @@ async function writeEnvFiles(
 ): Promise<void> {
   for (const spec of specs) {
     const content = Object.entries(spec.params)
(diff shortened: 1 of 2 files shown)


elvisun (approved):
You probably already did but just a reminder to test the deploy behavior :)

PR #4405, opened 2022-04-06
Allow local extensions in ext:* --local commands

elvisun:
### Description

Allow local extensions in ext:* --local commands

#### Install

Screen Shot 2022-04-07 at 12 14 41 PM

#### Update

Screen Shot 2022-04-07 at 12 14 48 PM

#### Configure

Screen Shot 2022-04-07 at 12 15 00 PM

#### Configure but with edited extension.yaml

Screen Shot 2022-04-07 at 12 15 11 PM

#### Deploy (doesn't work yet but nothing catastrophic)

Screen Shot 2022-04-07 at 12 15 21 PM


src/commands/ext-install.ts
```diff
@@ -184,46 +190,33 @@ export default new Command("ext:install [extensionName]")
     }
   });
 
-async function infoInstallBySource(
```
elvisun, line 187: these info functions have side effects... I refactored them all together

joehan:
This looks great! However, before we merge this to master, I think it would be god to at least add a try/catch or filter somewhere so that we just ignore local extensions on deploy instead of erroring out

src/extensions/manifest.ts
```diff
@@ -149,18 +149,44 @@ export function instanceExists(instanceId: string, config: Config): boolean {
   return !!config.get("extensions", {})[instanceId];
 }
 
-export function getInstanceRef(instanceId: string, config: Config): refs.Ref {
+/**
+ * Gets the instance's extension ref string or local path given an instanceId.
+ */
+export function getInstanceSource(instanceId: string, config: Config): string {
```
joehan, line 155: Consider a different word than source here - source already has a meaning (ExtensionSource). Maybe 'getInstanceTarget'?

joehan (approved):
One naming nit, but this lgtm overall

PR #4406, opened 2022-04-06
Minor refactoring of function deploy

taeold:
Just moving around code in `prepare.ts` phase of the function deploy to make it easy to support multi-codebase deploys. This is a simple refactor that should introduces no change to deploy behavior. I've added annotation to each section to make it slightly easier to understand what'a going on in each section.

There are few light changes:

* We reduce scope of `backend.checkAvailability` check to just the endpoints that's being deployed (to allow deploy to proceed even if region of endpoint NOT being deployed is down).  
* We reduce scope `ensureServiceAgentRoles`  check to just the endpoints that's being deployed (to reduce IAM calls to bare minimum)

### DIFF
--- a/src/deploy/functions/args.ts
+++ b/src/deploy/functions/args.ts
@@ -5,12 +5,24 @@ import * as deployHelper from "./functionsDeployHelper";
 
 // These types should probably be in a root deploy.ts, but we can only boil the ocean one bit at a time.
 
+interface CodebasePayload {
+  wantBackend: backend.Backend;
+  haveBackend: backend.Backend;
+}
+
 // Payload holds the output of what we want to build + what we already have.
 export interface Payload {
-  functions?: {
-    wantBackend: backend.Backend;
-    haveBackend: backend.Backend;
-  };
+  codebase?: CodebasePayload;
+}
+
+export interface Source {
+  // Filled in the "prepare" phase.
+  functionsSourceV1?: string;
+  functionsSourceV2?: string;
+
+  // Filled in the "deploy" phase.
+  sourceUrl?: string;
+  storage?: Record<string, gcfV2.StorageSource>;
 }
 
 // Context holds cached values of what we've looked up in handling this request.
@@ -22,15 +34,11 @@ export interface Context {
 
   // Filled in the "prepare" phase.
   config?: projectConfig.ValidatedSingle;
-  functionsSourceV1?: string;
-  functionsSourceV2?: string;
-  runtimeConfigEnabled?: boolean;
   artifactRegistryEnabled?: boolean;
   firebaseConfig?: FirebaseConfig;
 
-  // Filled in the "deploy" phase.
-  sourceUrl?: string;
-  storage?: Record<string, gcfV2.StorageSource>;
+  // Filled in the "prepare" and "deploy" phase.
+  source?: Source;
 }
 
 export interface FirebaseConfig {
--- a/src/deploy/functions/args.ts
+++ b/src/deploy/functions/args.ts
@@ -12,7 +12,7 @@ interface CodebasePayload {
 
 // Payload holds the output of what we want to build + what we already have.
 export interface Payload {
-  function?: CodebasePayload;
+  functions?: CodebasePayload;
 }
 
 export interface Source {
--- a/src/deploy/functions/args.ts
+++ b/src/deploy/functions/args.ts
@@ -12,7 +12,7 @@ interface CodebasePayload {
 
 // Payload holds the output of what we want to build + what we already have.
 export interface Payload {
-  functions?: CodebasePayload;
+  function?: CodebasePayload;
 }
 
 export interface Source {
--- a/src/deploy/functions/args.ts
+++ b/src/deploy/functions/args.ts
@@ -12,7 +12,7 @@ interface CodebasePayload {
 
 // Payload holds the output of what we want to build + what we already have.
 export interface Payload {
-  codebase?: CodebasePayload;
+  functions?: CodebasePayload;
 }
 
 export interface Source {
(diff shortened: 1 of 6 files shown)


src/deploy/functions/args.ts
```diff
@@ -5,12 +5,24 @@ import * as deployHelper from "./functionsDeployHelper";
...
+  haveBackend: backend.Backend;
+}
+
 // Payload holds the output of what we want to build + what we already have.
 export interface Payload {
-  functions?: {
-    wantBackend: backend.Backend;
-    haveBackend: backend.Backend;
-  };
+  codebase?: CodebasePayload;
```
inlined, line 15: This is a large object that is passed around for all deployment targets. It's still probably appropriate to have them in side a "functions" key. Also, are we not going to have an array of codebases or map of them soon?

src/deploy/functions/args.ts
```diff
@@ -22,15 +34,11 @@ export interface Context {
...
-  functionsSourceV2?: string;
-  runtimeConfigEnabled?: boolean;
   artifactRegistryEnabled?: boolean;
   firebaseConfig?: FirebaseConfig;
 
-  // Filled in the "deploy" phase.
-  sourceUrl?: string;
-  storage?: Record<string, gcfV2.StorageSource>;
+  // Filled in the "prepare" and "deploy" phase.
+  source?: Source;
```
inlined, line 41: This probably needs to be per-codebase, right?

src/deploy/functions/deploy.ts
```diff
@@ -29,11 +29,27 @@ async function uploadSourceV1(context: args.Context, region: string): Promise<vo
...
   const res = await gcfv2.generateUploadUrl(context.projectId, region);
   const uploadOpts = {
-    file: context.functionsSourceV2!,
-    stream: fs.createReadStream(context.functionsSourceV2!),
+    file: context.source!.functionsSourceV2!,
+    stream: fs.createReadStream(context.source!.functionsSourceV2!),
   };
   await gcs.upload(uploadOpts, res.uploadUrl);
-  context.storage = { ...context.storage, [region]: res.storageSource };
+  context.source!.storage = { ...context.source!.storage, [region]: res.storageSource };
```
inlined, line 36: Also, FYI, the Functions team backed down on their insistence that we always deploy from the same region as the function. We can unroll this to be "pick a region from the list of regions" like we did for v1 if you want.

src/deploy/functions/deploy.ts
```diff
@@ -29,11 +29,27 @@ async function uploadSourceV1(context: args.Context, region: string): Promise<vo
...
+    if (!v) {
+      throw new FirebaseError(
+        errMsg +
+          "This should never happen. Please file a bug at https://github.com/firebase/firebase-tools"
+      );
+    }
+  };
+  assertExists(context.config, "Functions config unexpectedly empty.");
+  assertExists(context.source, "Functions sources unexpectedly empty.");
+  assertExists(context.source?.functionsSourceV1, "Functions v1 source unexpectedly empty.");
```
inlined, line 51: Won't this be missing if there were no functions targeting v1?

src/deploy/functions/deploy.ts
```diff
@@ -47,18 +63,11 @@ export async function deploy(
...
 ): Promise<void> {
-  if (!context.config) {
-    return;
-  }
-
-  if (!context.functionsSourceV1 && !context.functionsSourceV2) {
-    return;
-  }
-
+  assertPreconditions(context, options, payload);
```
inlined, line 66: I don't think this is correct. The previous code helped make sure that we silently short-circuit the functions deploy code if someone makes a deploy but the functions directory doesn't exist. At least verify that this works for a new project w/o inlitializing functions and one where the functions directory exists but the samples are still commented out & therefore no functions are to be deployed.

inlined (commented):
I have mixed feelings. On one hand, I always envisioned having multiple codebase objects like you do now. On the other hand, having codebase as an attribute of a single backend meant it was easier for the deployment planner to move a function rather than delete + recreate. As long as you worry about that case then I'm happy.

In the meantime, also double check that changing those early exits to assertions doesn't break edge cases. IIRC it did before when I tried similar.

src/deploy/functions/args.ts
taeold: Yes exactly. I'm trying to prepare the stage so that adding codebase -> <whatever> mapping seems more obvious.

src/deploy/functions/args.ts
taeold: Yep. Being done in https://github.com/firebase/firebase-tools/pull/4419.

src/deploy/functions/deploy.ts
taeold: Oh interesting. I thought returning early was a mistake - I'll double check the behavior and bring back the old behavior.

taeold: Deploy unexpectedly failed on a project w/ functions directly with 0 function. Reverted back to the old logic. Thanks for the catch!

src/deploy/functions/deploy.ts
taeold: Resolved by reverting back to the old check style.

src/deploy/functions/prepareFunctionsUpload.ts
```diff
@@ -18,32 +18,26 @@ import * as projectConfig from "../../functions/projectConfig";
...
+      throw new FirebaseError(
+        "Cloud Runtime Config is currently experiencing issues, " +
+          "which is preventing your functions from being deployed. " +
+          "Please wait a few minutes and then try to deploy your functions again." +
+          "\nRun `firebase deploy --except functions` if you want to continue deploying the rest of your project."
+      );
     }
   }
-
-  config.firebase = context.firebaseConfig;
```
inlined, line 45: What happened to the `firebase` key?

inlined (approved):
I wonder if you'll want to have an array of `codebase` with a wantBackend and maybe a haveBackend and a source. Then in prepare you'll track the source for every codebase and read all backends as wantBackend but then you'll split haveBackend by region + id rather than its codebase label? I'm just so antsy about that migration case.

taeold:
@inlined would love for yall to take a look at the following PR on how we can go about implementing multi codebase deploy!

Issue #4396, opened 2022-04-04
File uploads in storage emulator on m1 mac failing


rishisv:
### [REQUIRED] Environment info

**firebase-tools:** 10.5.0

**Platform:** macOS version 12.2.1, Apple M1 Pro chip

### [REQUIRED] Test case

* Launch the storage emulator, localhost:9595/storage and upload a large file. 
* Error is seen in the console: 
`Uncaught (in promise) FirebaseError: Firebase Storage: An unknown error occurred, please check the error payload for server response. (storage/unknown)`

<img width="1102" alt="image" src="https://user-images.githubusercontent.com/59505526/161490387-f1c200c2-b32c-473b-bc79-f441ab0f98f1.png">

### [REQUIRED] Steps to reproduce

* Install firebase-tools, launch emulator locally, go to the storage emulator webpage and upload a large image

### [REQUIRED] Expected behavior

* No errors

### [REQUIRED] Actual behavior
* Error seen with large files > ~500KB.

rishisv:
There's a related issue #4346 that talks about 4mb file sizes and on older version of firebase tools but I can't even upload ~500kb files with the latest version. My teammates on x86/intel mac do not have this issue.

tohhsinpei:
Fixed in #4407

Issue #4364, opened 2022-03-26
Storage Emulator Upload Error: Range invalid for upload data.


Vortec4800:
### [REQUIRED] Environment info

**firebase-tools:** 10.5.0

**Platform:** macOS 12.3 on M1 Max

### [REQUIRED] Test case

```swift
private func uploadSelectedImage(_ selectedImage: UIImage) async {
	do {
		let storageRef = Storage.storage().reference().child("gallery/image.jpg")
		let metadata = StorageMetadata()
		metadata.contentType = "image/jpg"
		if let data = selectedImage.jpegData(compressionQuality: 0.8) {
			let uploadMetadata = try await storageRef.putDataAsync(data, metadata: metadata)
			
			print("\(uploadMetadata)")
		}
	} catch {
		print("Error uploading image: \(error)")
	}
}
```

### [REQUIRED] Steps to reproduce

Upload an image using the provided function, after connecting to Firebase Storage Emulator.

Storage rules are to allow writes if authenticated.

```
rules_version = '2';
service firebase.storage {
  match /b/{bucket}/o {
    match /{allPaths=**} {
      allow read, write: if request.auth!=null;
    }
  }
}
```

### [REQUIRED] Expected behavior

Image to upload to Storage Emulator and print `uploadMetadata` to console.

### [REQUIRED] Actual behavior

Image actually does successfully upload looking in the Storage Emulator UI, however the error catch happens with the following output:

```
2022-03-26 15:10:39.827585-0500 ConnectDataTest[13671:170319] chunk fetcher completion has kStatusUnknown upload status for headers <CFBasicHash 0x6000035fb600 [0x1de715af0]>{type = immutable dict, count = 9,
entries =>
	0 : Connection = <CFString 0x6000020ed560 [0x1de715af0]>{contents = "keep-alive"}
	1 : X-Powered-By = Express
	2 : Access-Control-Expose-Headers = <CFString 0x10a5423d0 [0x1de715af0]>{contents = "content-type,x-firebase-storage-version,x-goog-upload-url,x-goog-upload-status,x-goog-upload-command,x-gupload-uploadid,x-goog-upload-header-content-length,x-goog-upload-header-content-type,x-goog-upload-protocol,x-goog-upload-status,x-goog-upload-chunk-granularity,x-goog-upload-control-url"}
	3 : Content-Type = <CFString 0x600003594940 [0x1de715af0]>{contents = "application/json; charset=utf-8"}
[...]

PR #4407, opened 2022-04-06
Fix missing upload status header in resumable upload finalise response

### LINKED ISSUE #4346: Error uploading large files to emulator storage since 10.3.0

**firebase-tools:** 10.4.0

**Platform:** macOS

### [REQUIRED] Test case

I have a simple file input dialog and I'm using [uploadBytesResumable](https://firebase.google.com/docs/reference/js/storage.md#uploadbytesresumable) from the Javascript SDK to upload to the storage emulator. This works fine for smaller files, but when I try to upload a larger file (~4MB), I receive an error in the console.

```
const uploadTask = uploadBytesResumable(
  storageRef,
  file.data
);

uploadTask.catch((error) => {
  console.log(error);
  console.log(uploadTask.snapshot);
});
```

Upload a larger file to the storage emulator. I've also been able to reproduce by uploading a larger file through the Emulator UI.

The file uploads successfully to the storage emulator.

`FirebaseError: Firebase Storage: An unknown error occurred, please check the error payload for server response. (storage/unknown)`

Screenshot 2022-03-21 at 22 27 53

One difference I noticed between large files and small files is that the _resumable property of the UploadTask is only true for the larger files.

tohhsinpei:
#### Issue
This PR fixes #4346, #4364, and #4396.

Any large (> ~500kB) upload to the Storage Emulator gives a console error.

On web:
```
Uncaught (in promise) FirebaseError: Firebase Storage: An unknown error occurred, please check the error payload for server response. (storage/unknown)
```

On iOS:
```
2022-03-31 16:50:05.822789-0400 Example[51380:621771] offset 438491 exceeds data length 438491
2022-03-31 16:50:05.823434-0400 Example[51380:622628] Range invalid for upload data.  offset: 438491	length: 8491	dataLength: 438491	expected UploadLength: 438491
2022-03-31 16:50:08.017868-0400 Example[51380:622083] [javascript] Possible Unhandled Promise Rejection (id: 0):
```

On the emulator UI, the upload is shown with a loading spinner indefinitely. However, on refresh we see that the file was in fact uploaded successfully.

This issue started in v10.3.0.

#### Cause
In #4250 we refactored the upload logic. In doing so we forgot to include the header `X-Goog-Upload-Status` with value `final` in the response for the `finalize` request. My guess is that the client expects to make further `upload` requests and errors out when there are no more bytes to upload.

### Scenarios Tested
Tested on web and verified that console error no longer appears and that the header is included in the response. Also included a check for the latter in an existing integration test for resumable uploads.

### DIFF
--- a/scripts/storage-emulator-integration/tests.ts
+++ b/scripts/storage-emulator-integration/tests.ts
@@ -2105,14 +2105,17 @@ describe("Storage emulator", () => {
           })
           .expect(200);
 
-        await supertest(STORAGE_EMULATOR_HOST)
+        const uploadStatus = await supertest(STORAGE_EMULATOR_HOST)
           .put(uploadURL.pathname + uploadURL.search)
           .set({
             // No Authorization required in finalize
             "X-Goog-Upload-Protocol": "resumable",
             "X-Goog-Upload-Command": "upload, finalize",
           })
-          .expect(200);
+          .expect(200)
+          .then((res) => res.header["x-goog-upload-status"]);
+
+        expect(uploadStatus).to.equal("final");
 
         await supertest(STORAGE_EMULATOR_HOST)
           .get(`/v0/b/${storageBucket}/o/test_upload.jpg`)
--- a/src/emulator/storage/apis/firebase.ts
+++ b/src/emulator/storage/apis/firebase.ts
@@ -346,6 +346,7 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
         throw err;
       }
 
+      res.header("x-goog-upload-status", "final");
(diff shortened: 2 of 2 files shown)


tonyjhuang:
Wow nice find! Do we need to do this in `gcloud.ts` as well?

tohhsinpei:
> Wow nice find! Do we need to do this in `gcloud.ts` as well?

No... The way the Admin SDK [does it](https://github.com/firebase/firebase-tools/blob/19d3e24b488b4b96bc4428284648f17404b0146d/src/emulator/storage/apis/gcloud.ts#L139-L140) is a bit different. It'll just be one POST to start then one PUT to upload and finalise.

tonyjhuang (approved):

PR #4409, opened 2022-04-06
Prefer host from request in storage resumable upload.

### LINKED ISSUE #4374: Storage emulator returning wrong host for upload URL's

**firebase-tools:** 10.5.0

**Platform:** macOS

1. create docker contianer with storage emulator in it and make it listen to "host": "0.0.0.0" (so it can be accessed from outside the container)
2. create a second container and try to upload a file to the emulator 

what will happen is that it will try to connect to 0.0.0.0 and it will fail

the emulator should not return the bind host, instead return the req.hostname

emulator uses the bing host to determine the host

related issue: https://github.com/googleapis/nodejs-storage/issues/1611

yuchenshi:
### Description

Fixes https://github.com/firebase/firebase-admin-node/issues/1562
Fixes #4374

### Scenarios Tested

See tests added.

### DIFF
--- a/src/emulator/auth/utils.ts
+++ b/src/emulator/auth/utils.ts
@@ -170,48 +170,18 @@ export function logError(err: Error): void {
 
 /**
  * Return a URL object with Auth Emulator protocol, host, and port populated.
- * @param req a express request to emulator server; used to infer host
- * @return the construted URL object
+ *
+ * Compared to EmulatorRegistry.url, this functions prefers the configured host
+ * and port, which are likely more useful when the link is opened on the same
+ * device running the emulator (assuming developers click on the link printed on
+ * terminal or Emulator UI).
  */
 export function authEmulatorUrl(req: express.Request): URL {
-  // WHATWG URL API has no way to create from parts, so let's use a minimal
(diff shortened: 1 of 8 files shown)


src/emulator/auth/utils.ts
```diff
@@ -170,48 +170,18 @@ export function logError(err: Error): void {
...
-      url.hostname = "[::1]";
-    } else if (info.host.includes(":")) {
-      url.hostname = `[${info.host}]`; // IPv6 addresses need to be quoted using brackets.
-    } else {
-      url.hostname = info.host;
-    }
-    url.port = info.port.toString();
+  if (EmulatorRegistry.getInfo(Emulators.AUTH)) {
+    return EmulatorRegistry.url(Emulators.AUTH);
   } else {
```
tonyjhuang, line 182: unnecessary else

tonyjhuang (approved):
Thanks for the fix!

src/emulator/auth/utils.ts
yuchenshi: Oops I had automerge on. Next time then...

PR #3647, opened 2021-08-06
Emulator fixes

rhodgkins:
### Description

When using the emulator against `@google-cloud/storage`, I noticed a few of the headers were missing, this mainly caused any validation on download to fail (as the emulated endpoint was missing the `x-goog-hash` header).

I've also included some values for the other `x-goog` headers.

### Scenarios Tested

N/A

### Sample Commands

N/A

PR #4411, opened 2022-04-06
Fix cross-platform incompatibility with Storage Emulator exports

tohhsinpei:
### Description
This PR fixes https://github.com/firebase/firebase-tools/issues/4322#issuecomment-1084657623, hopefully the last follow up to #3647. The Storage emulator is unable to import data previously exported on a different platform because we didn't account for different file path formats (e.g. `\` delimiter on Windows but `/` on Linux) when flattening the directory structure.

Note re: the fix implemented here: ideally I would store and export everything in POSIX file format, then do the obvious replacement when importing. But I don't think that's a good idea for backwards compatibility reasons. Lmk if you have thoughts.

### Scenarios Tested
Added integration test to verify that emulator is able to import data previously exported on Windows.

### DIFF
--- a/scripts/storage-emulator-integration/import/windows-emulator-data/storage_export/metadata/fake-project-id.appspot.com%5Ctest_upload.jpg.json
+++ b/scripts/storage-emulator-integration/import/windows-emulator-data/storage_export/metadata/fake-project-id.appspot.com%5Ctest_upload.jpg.json
@@ -1,0 +1,20 @@
+{
+  "name": "test_upload.jpg",
+  "bucket": "fake-project-id.appspot.com",
+  "contentType": "application/octet-stream",
+  "metageneration": 1,
+  "generation": 1648084940926,
+  "storageClass": "STANDARD",
+  "contentDisposition": "inline",
+  "cacheControl": "public, max-age=3600",
+  "contentEncoding": "identity",
+  "downloadTokens": [
+    "c3c71086-95a8-445d-96e7-f625972de4b0"
+  ],
+  "etag": "PQJQBXRweACX9yRsBEInQjOJ/0s",
+  "timeCreated": "2022-03-24T01:22:20.926Z",
+  "updated": "2022-03-24T01:22:20.926Z",
+  "size": 0,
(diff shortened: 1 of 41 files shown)


src/emulator/storage/files.ts
```diff
@@ -610,3 +613,11 @@ export class StorageLayer {
     }
   }
 }
+
+/** Returns file separator used in given path, either '\\' or '/'. */
+function getPathSep(decodedPath: string): string {
+  // Suffices to check first separator, which occurs immediately after bucket name
+  const bucketSuffix = ".appspot.com";
```
tonyjhuang, line 620: buckets don't have to end with .appspot.com. The guidelines for bucket naming state:

"""
Bucket names can only contain lowercase letters, numeric characters, dashes (-), underscores (_), and dots (.). Spaces are not allowed.
"""

See https://cloud.google.com/storage/docs/naming-buckets

tohhsinpei: Done.

scripts/storage-emulator-integration/import/windows-emulator-data/storage_export/metadata/fake-project-id.appspot.com%5Ctest_upload.jpg.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "test_upload.jpg",
```
tonyjhuang, line 2: can you include a test with a nested directory structure

tonyjhuang (commented):
thanks for the fix! LGTM outside of a test case

scripts/storage-emulator-integration/import/windows-emulator-data/storage_export/metadata/fake-project-id.appspot.com%5Ctest_upload.jpg.json
tohhsinpei: Done.

tonyjhuang (approved):

PR #4415, opened 2022-04-07
Catch OSX port 5000 in-use

jamesdaniels:
OSX is reserving port 5000 in a way that Node's http server is not erroring when you listen on it. Add a special case for this.


src/serve/hosting.ts
```diff
@@ -44,6 +45,34 @@ function startServer(options: any, config: any, port: number, init: TemplateServ
...
+      });
+    }
+  };
+
+  // On OSX, some ports may be reserved by the OS in a way that node http doesn't detect.
+  // Starting in MacOS 12.3 it does this with port 5000 our default port. This is a bad
+  // enough devexp that we should special case and ensure it's available.
+  if (process.platform === "darwin") {
+    try {
+      execSync(`lsof -i :${port}`);
```
joehan, line 68: Looking through this, it seems like this only checks if the initial port is being used by osX, and will still break out if a subsequent attempt hits a osx occupied port. Should we include this `if (platform == 'darwin') lsof` in portInUse() as well?

joehan (approved):
Thanks for fixing this, this was super annoying! One potential extra edge case to cover, but this already looks like a nice improvement.

src/serve/hosting.ts
jamesdaniels: portInUse recurses, so it should work in subsequent attempts

jamesdaniels: see line 54

jamesdaniels:
@joehan thanks for the quick turn around on your reviews today!

PR #4417, opened 2022-04-07
Create Build type with conversion to Backend

Berlioz:
Creating a new type which is backwards-compatible with backend.Backend in preparation for parameterization during CF3 discovery. This code is currently unreachable, but the type should be fully defined before I start parsing YAML into it.

Now available in in-the-right-repository flavor! My understanding is that comments on the other PR have been resolved at this point.


src/deploy/functions/build.ts
```diff
@@ -0,0 +1,397 @@
...
+  } else if (typeof from === "string") {
+    throw new FirebaseError("CEL evaluation of expression '" + from + "' not yet supported");
+  }
+  return from;
+}
+
+function resolveString(from: string | Expression<string> | null): string {
+  if (from == null) {
+    return "";
+  }
```
inlined, line 43: nit: Maybe check for "{{ }}" and throw here too.

src/deploy/functions/build.ts
```diff
@@ -0,0 +1,397 @@
...
+// Trigger definitions for endpoints that listen to CloudEvents emitted by other systems (or legacy
+// Google events for GCF gen 1)
+interface EventTrigger {
+  eventType: string;
+  eventFilters: Record<string, Expression<string>>;
+
+  // whether failed function executions should retry the event execution.
+  // Retries are indefinite, so developers should be sure to add some end condition (e.g. event
+  // age)
+  retry: boolean | Expression<boolean> | null;
```
inlined, line 88: nit: here and elsewhere you could have saved yourself a lot of time by defining `Expression<T>` to be

```
// TOOD RE list
Expression<T extends string | number | boolean> = T extends string ? string | null : T | `{{ ${infer expr} }}` | null
```

And then you can just say

```
retry: Expression<boolean>
```

Or if you want to leave `Expression` as the string type, you could define `Field<T> = T | Expression<T> | null`

src/deploy/functions/build.ts
```diff
@@ -0,0 +1,397 @@
...
+        timeoutSeconds: timeout,
+        ...trigger,
+      };
+      proto.renameIfPresent(bkEndpoint, endpoint, "maxInstances", "maxInstances", resolveInt);
+      proto.renameIfPresent(bkEndpoint, endpoint, "minInstances", "minInstances", resolveInt);
+      proto.renameIfPresent(bkEndpoint, endpoint, "concurrency", "concurrency", resolveInt);
+      proto.copyIfPresent(bkEndpoint, endpoint, "ingressSettings");
+      if (endpoint.vpc) {
+        bkEndpoint.vpc = {
+          connector: endpoint.vpc.connector,
```
inlined, line 267: connector might be a template

src/deploy/functions/build.ts
```diff
@@ -0,0 +1,397 @@
...
+// TODO(vsfan): resolve build.Params
+// TODO(vsfan): handle Expression<T> types
+export function discoverParams(build: Build): backend.Backend {
+  const bkEndpoints: Array<backend.Endpoint> = [];
+  for (const endpointId of Object.keys(build.endpoints)) {
+    const endpoint = build.endpoints[endpointId];
+
+    let regions = endpoint.region;
+    if (typeof regions === "undefined") {
+      regions = ["functionsDefaultRegion"];
```
inlined, line 229: Let's use the actual functionsDefaultRegion

src/deploy/functions/build.ts
```diff
@@ -0,0 +1,397 @@
...
+    }
+    trigger = { httpsTrigger: bkHttps };
+  } else if ("callableTrigger" in endpoint) {
+    trigger = { callableTrigger: {} };
+  } else if ("blockingTrigger" in endpoint) {
+    throw new FirebaseError("blocking triggers not supported");
+  } else if ("eventTrigger" in endpoint) {
+    const bkEvent: backend.EventTrigger = {
+      eventType: endpoint.eventTrigger.eventType,
+      eventFilters: endpoint.eventTrigger.eventFilters,
```
inlined, line 301: Don't you need to map over these and resolveString?

src/deploy/functions/build.ts
```diff
@@ -0,0 +1,397 @@
...
+  retryCount?: number | Expression<number> | null;
+  maxRetrySeconds?: number | Expression<number> | null;
+  minBackoffSeconds?: number | Expression<number> | null;
+  maxBackoffSeconds?: number | Expression<number> | null;
+  maxDoublings?: number | Expression<number> | null;
+}
+
+interface ScheduleTrigger {
+  schedule: string | Expression<string>;
+  timeZone: string;
```
inlined, line 133: I wonder if this should be an Expression as well. People could conceivably use a SelectInput

src/deploy/functions/build.ts
```diff
@@ -0,0 +1,397 @@
...
+    if (endpoint.eventTrigger.serviceAccount) {
+      bkEvent.serviceAccountEmail = endpoint.eventTrigger.serviceAccount;
+    }
+    if (endpoint.eventTrigger.region) {
+      bkEvent.region = resolveString(endpoint.eventTrigger.region);
+    }
+    trigger = { eventTrigger: bkEvent };
+  } else if ("scheduleTrigger" in endpoint) {
+    const bkSchedule: backend.ScheduleTrigger = {
+      schedule: endpoint.scheduleTrigger.schedule,
```
inlined, line 313: resolve strings

src/deploy/functions/build.ts
```diff
@@ -0,0 +1,397 @@
...
+        resolveInt
+      );
+      bkTaskQueue.retryConfig = bkRetryConfig;
+    }
+    if (endpoint.taskQueueTrigger.invoker) {
+      bkTaskQueue.invoker = endpoint.taskQueueTrigger.invoker.map((sa) => resolveString(sa));
+    }
+    trigger = { taskQueueTrigger: bkTaskQueue };
+  } else {
+    throw new FirebaseError("unknown trigger type");
```
inlined, line 394: If you've done the typing right you should be able to just say `assertExhaustive(endpoint)` and it'll fail to compile if we ever add a new trigger type.

src/deploy/functions/build.ts
```diff
@@ -0,0 +1,397 @@
...
+type Param = StringParam;
+
+function isMemoryOption(value: backend.MemoryOptions | any): value is backend.MemoryOptions {
+  return value == null || [128, 256, 512, 1024, 2048, 4096, 8192].includes(value);
+}
+
+/** Converts a build specification into a Backend representation, with all Params resolved and interpolated */
+// TODO(vsfan): resolve build.Params
+// TODO(vsfan): handle Expression<T> types
+export function discoverParams(build: Build): backend.Backend {
```
inlined, line 222: nit: I would personally call this something like `resolveBackend` since it isn't "discovering" params.

inlined (approved):
This looks fine. I'm curious how you plan to actually do interpolation though. Will you run a CEL interpreter for each resolveNumber, resolveString, etc?

src/deploy/functions/build.ts
```diff
@@ -0,0 +1,397 @@
...
+
+// Trigger definitions for endpoints that should be called as a delegate for other operations.
+// For example, before user login.
+interface BlockingTrigger {
+  eventType: string;
+}
+
+// Trigger definitions for endpoints that listen to CloudEvents emitted by other systems (or legacy
+// Google events for GCF gen 1)
+interface EventTrigger {
```
inlined, line 81: Holy crap. After days of @taeold nerd snping me, I finally figured out how to use type programming to make a build just a type transformation on an Endpoint.

```
export type RecursiveKeyOf<T extends object> = {
  [Key in keyof T & (string | number)]:
    T[Key] extends any[]
    ? `${Key}`
    : T[Key] extends object
      ? `${Key}` | `${Key}.${RecursiveKeyOf<T[Key]>}`
      : `${Key}`;
}[keyof T & (string | number)];

export type HeadOf<T extends string> = [
  T extends `${infer Head}.${infer Tail}` ? Head : T
][number]

export type TailsOf<T extends string, Head extends string> = [
  T extends `${Head}.${infer Tail}` ? Tail : never
][number];

export type Parameterized<T> =
  T extends Array<infer A>
  ? Array<Parameterized<A>>
  : T extends object
    ? { [Key in keyof T & string]: Parameterized<T[Key]> }
    : T | string | null | undefined;

xport type Parameterize<T extends object, Paths extends RecursiveKeyOf<T>> = ParameterizeUnsafe<T, Paths>;

type ParameterizeUnsafe<T extends object, Paths extends string> = {
  [Key in keyof T]:
  Key extends Paths
  ? Parameterized<T[Key]>
  : Key extends HeadOf<Paths>
    ? ParameterizeUnsafe<T[Key] & object, TailsOf<Paths, Key>>
    : T[Key]
};

// Demo
interface Endpoint {
  platform: "gcfv1" | "gcfv2";
  entrypoint: number;
  eventTrigger: {
    eventType: number;
    filters: Record<string, string>;
  };
};
```

`Parameterized<Endpoint, "entrypoint" | "eventTrigger.eventType">` is equivalent to 

```
{
  platform: "gcfv1" | "gcfv2
  entrypoint: number | string | undefined | null
  eventTrigger: {
    eventType: number | string | undefined | null
    filters: Record<string, string>
  }
}
```

src/deploy/functions/build.ts

src/deploy/functions/build.ts

src/deploy/functions/build.ts
Berlioz: I think that's reasonable. Done.

src/deploy/functions/build.ts

src/deploy/functions/build.ts

src/deploy/functions/build.ts
Berlioz: Done.

PR #4419, opened 2022-04-08
Allow and deploy multiple codebases on deploy.

taeold:
## What's new
Let's say we are given `firebase.json` that specifies multiple function configuration (which became possible w/ https://github.com/firebase/firebase-tools/pull/4269):

```json
{
  "functions": [
    {
      "source": "functions",
      "codebase": "foo"
    },
    {
      "source": "moreFunctions",
      "codebase": "bar"
    }
  ]
}
```

With this change, we now deploy functions discovered in both codebases:

```bash

i  deploying functions
...
i  functions: preparing codebase foo for deployment
i  functions: preparing codebase bar for deployment
i  functions: packaged functions(45.24 KB) for uploading
i  functions: preparing moreFunctions directory for uploading...
i  functions: packaged moreFunctions (48.39 KB) for uploading
✔  functions: moreFunctions folder uploaded successfully
i  functions: creating Node.js 14 function [foo]myFn(us-central1)...
i  functions: creating Node.js 14 function [foo]anotherfn(us-central1)...
i  functions: creating Node.js 14 function [bar]otherfn(us-west1)...
```

It's possible to just deploy functions in a specific codebase using the `--only` flag:

```bash
$ firebase deploy --only functions:foo

i  deploying functions
...
i  functions: preparing codebase foo for deployment
i  functions: packaged functions(45.24 KB) for uploading
i  functions: creating Node.js 14 function [foo]myFn(us-central1)...
i  functions: creating Node.js 14 function [foo]anotherfn(us-central1)...
```

## Implementation Note
* Most of the deployment logic remains the same except that we are now book-keeping `sources` and `want/haveBackends` for each of the target codebase.
* Changes in prepare stage of deployment looks big at first glance, but it's just wrapping backend discovery logic inside a for loop while moving around few pieces to ensure that we don't duplicate work like enabling necessary GCP APIs.
* At the time of creating deployment plan to be executed by the fabricator, we simply merge the plan generated for each codebase. Fabricator works just fine.
* We add logic to prevent setting codebase label on functions for fns in the "default" codebase. This should preserve the old behavior and only use the label for intentional codebase usages.


src/deploy/functions/functionsDeployHelper.ts
```diff
@@ -127,6 +127,84 @@ export function getEndpointFilters(options: { only?: string }): EndpointFilter[]
...
+    id = `[${fn.codebase}]${id}`;
+  }
+  return id;
+}
+
+/**
+ * Returns list of codebases specified in firebase.json filtered by --only filters if present.
+ */
+export function targetCodebases(config: ValidatedConfig, filters?: EndpointFilter[]): string[] {
+  const codebasesFromConfig = new Set(Object.values(config).map((c) => c.codebase));
```
colerogers, line 142: I notice that your using quite a bit of Set's here, I've received feedback from @inlined to try to stay away from them since they're not really idiomatic of js. I see here why they're useful, so I'm a bit conflicted

src/deploy/functions/prepare.ts
```diff
@@ -78,114 +74,117 @@ export async function prepare(
...
-    projectDir: options.config.projectDir,
-    runtime: context.config.runtime || "",
-  };
-  const runtimeDelegate = await runtimes.getRuntimeDelegate(delegateContext);
-  logger.debug(`Validating ${runtimeDelegate.name} source`);
-  await runtimeDelegate.validate();
-  logger.debug(`Building ${runtimeDelegate.name} source`);
-  await runtimeDelegate.build();
+  context.sources = {};
+  const codebaseUsesEnvs = new Set<string>();
```
colerogers, line 78: Same here, do we want to just use an array?

src/deploy/functions/prepare.ts
```diff
@@ -78,114 +74,117 @@ export async function prepare(
...
+    const userEnvs = functionsEnv.loadUserEnvs(userEnvOpt);
+    logger.debug(`Analyzing ${runtimeDelegate.name} backend spec`);
+    const wantBackend = await runtimeDelegate.discoverSpec(runtimeConfig, firebaseEnvs);
+    wantBackend.environmentVariables = { ...userEnvs, ...firebaseEnvs };
+    for (const endpoint of backend.allEndpoints(wantBackend)) {
+      endpoint.environmentVariables = wantBackend.environmentVariables;
+      endpoint.codebase = codebase;
+    }
+    wantBackends[codebase] = wantBackend;
+    if (functionsEnv.hasUserEnvs(userEnvOpt)) {
```
colerogers, line 118: we can add `&& !codebaseUsesEnvs.includes(codebase)` to keep duplicates out if we switch to using an array

src/deploy/functions/release/planner.ts
```diff
@@ -92,37 +99,33 @@ export function calculateUpdate(want: backend.Endpoint, have: backend.Endpoint):
 
 /**
  * Create a plan for deploying all functions in one region.
- * @param want the desired state
```
colerogers, line 95: Should we still document what's in `PlanArgs` here?

colerogers (commented):
Some q's

src/deploy/functions/functionsDeployHelper.ts
```diff
@@ -127,6 +127,84 @@ export function getEndpointFilters(options: { only?: string }): EndpointFilter[]
...
+ * Breakup backends by codebase.
+ *
+ * An endpoint is part a codebase if:
+ *   1. Endpoint is associated w/ the current codebase (duh).
+ *   2. Endpoint name matches name of an endoint we want to deploy
+ *
+ * Condition (2) might feel wrong but is a practical conflict resolution strategy as it makes migrating a function
+ * from one codebase to another straightforward.
+ */
+export function groupByCodebase(
```
colerogers, line 174: Maybe change this to something like `groupHaveBackendByCodebase`

src/deploy/functions/prepare.ts
taeold: At this point, we are guaranteed to have unique codebases, so I'll keep that out for brevity!

src/deploy/functions/release/planner.ts
taeold: I think going to `PlanArgs` to read its documentation works . WDYT?

src/deploy/functions/checkIam.ts
```diff
@@ -128,7 +132,7 @@ function reduceEventsToServices(services: Array<Service>, endpoint: backend.Endp
  * @param existingPolicy the project level IAM policy
  * @param serviceAccount the IAM service account
  * @param role the role you want to grant
- * @returns
+ * @return
```
colerogers, line 135:
```suggestion
 * @returns the correct IAM binding
```

colerogers (approved):
lgtm

inlined (commented):
LGTM with some post-review nits.

PR #4427, opened 2022-04-11
Added eventarc triggering

apascal07:


src/emulator/functionsEmulator.ts
```diff
@@ -623,6 +623,13 @@ export class FunctionsEmulator implements EmulatorInstance {
               definition.schedule
             );
             break;
+          case Constants.SERVICE_EVENTARC:
```
huangjeff5, line 626: It seems like this won't ever be Constants.SERVICE_EVENTARC becuase getFunctionService(definition) doesn't return that value

apascal07: Good catch, thanks!

huangjeff5 (approved):

PR #4431, opened 2022-04-12
Fix issue where v2 deploys are failing due to bad refactoring.

taeold:
I decided to make a small change AFTER getting approval for https://github.com/firebase/firebase-tools/pull/4406. This is a bad idea in general, and espeically bad idea now that I've completely broken v2 function deploys.

Here a fix to correctly upload source code for v2 function.


colerogers (approved):

inlined (commented):
LGTM

PR #4435, opened 2022-04-12
Add CPU option

inlined:
Add a `cpu` option that will eventually allow V2 functions to have custom CPU shapes. This does nothing yet to wire up those options.

While I was at it, I used my new type system foo to improve the schema enforcement tools. Now if you have a type `{foo: string}` you cannot accidentally create a schema `{ foo: "number" }`. You can only have a schema value of `"number" | "omit" | ((val: number) => boolean` (I added boolean validators to handle enum types or union types).


src/deploy/functions/backend.ts
```diff
@@ -211,6 +217,9 @@ export interface SecretEnvVar {
   version?: string;
 }
 
+/**
+ *
```
taeold, line 221:
```suggestion
 * Returns full resource name of a secret version.
```

src/deploy/functions/runtimes/discovery/parsing.ts
```diff
@@ -2,8 +2,24 @@ import { FirebaseError } from "../../../../error";
...
+      ? "boolean"
+      : T extends unknown[]
+      ? "array"
+      : T extends object
+      ? "object"
+      : never)
+  | "omit"
+  | ((t: T) => boolean);
+/**
+ *
```
taeold, line 20:
```suggestion
 * Asserts that all yaml contains all required keys specified in the schema.
 ```

File: src/deploy/functions/runtimes/discovery/parsing.ts [hunk 3]
```diff
@@ -14,40 +30,50 @@ export function requireKeys<T extends object>(prefix: string, yaml: T, ...keys:
   }
 }
 
-export function assertKeyTypes<T extends Object>(
+/**
+ *
```
taeold, line 34:
```suggestion
 * Asserts that runtime types of the given object matches the type specified in the schema.
```

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -85,17 +85,17 @@ function parseEndpoints(
 
   assertKeyTypes(prefix, ep, {
     region: "array",
-    platform: "string",
+    platform: (platform) => backend.AllFunctionsPlatforms.includes(platform),
```
taeold, line 88: dang.

taeold (approved):

PR #4439, opened 2022-04-13
Convert track.js to TS.

yuchenshi:
Actual diff:

https://github.com/firebase/firebase-tools/compare/3a613f1cc82a07093f37c700c1542a6801825b19...ys/tsify-track


src/track.ts
```diff
@@ -44,5 +41,4 @@ function track(action, label, duration) {
 
 // New code should import track by name so that it can be stubbed
 // in unit tests. Legacy code still imports as default.
-track.track = track;
-module.exports = track;
+export default track;
```
mbleigh, line 44: We have generally avoided default exports when converting to TS, instead preferring to change all importing files to use `import { track } from "../track"` style instead.

yuchenshi: Done.

mbleigh (approved):

Issue #737, opened 2018-04-11
Deployment fails due to size constraint / inclusion of node_modules


swftvsn:
### Version info

3.18.2

### Platform Information

OS X

### Steps to reproduce

run firebase deploy

### Expected behavior

The deployment does not contain node_modules

### Actual behavior

My deploy fails due to too big size. (117MB) This is expected. What I did not expect, was to find out that the zip generated by `firebase deploy` command includes the node_modules folder. 

Is this intentional or a regression? I also noticed that the node_modules folder contain all development deps, like typescript (36MB) etc. The highest size seems to be google-gax (53.8MB), @google-gloud (47.6MB). (GRPC seems to be the root cause for bloat on both)

Is there anything I can do to make 
- firebase deploy command to drop my development dependencies or
- to not deploy node_modules at all
?

swftvsn:
OK, so googling around I found out that I can do

```
npm install
npm run build
npm prune --production
firebase deploy --only functions
npm install
```

to make sure that all deps are there to run build and then remove devDeps and then deploy and then install everything back.

swftvsn:
Before zipping the node_modules is almost 400Mb. 😄 I need to lose some weight here. The libs also contain sharp with its binaries as also pdfium that do not compress that well.

laurenzlong:
This was just fixed in 3.18.3

PR #4440, opened 2022-04-13
Set up a dummy server to be used for emulating Eventarc #737

huangjeff5:


src/emulator/eventarcEmulator.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  constructor(private args: EventarcEmulatorArgs) {}
+
+  createHubServer(): express.Application {
+    const hub = express();
+
+    hub.use(express.json())
+
+    const helloWorldRoute = `/hello_world`;
+    const publishEventsRoute = `/v1/projects/:project_id/locations/:location/channels/:channel::publishEvents`;
```
apascal07, line 32: Double colon intentional? Do we need to escape the character here?

huangjeff5: Yep, it's intentional. Otherwise it gets picked up as a variable.

PR #4441, opened 2022-04-13
Fix function deploy bug

taeold:
Fixes a bug introduced in https://github.com/firebase/firebase-tools/pull/4431 where deployment will fail if the source doesn't specify both v1 and v2 triggers.

On one hand, I'm glad I running integrations test that catches issue like this. This also pushes me towards having multiple integration tests that test many different deployment scenario instead of having one gigantic deployment that tests all triggers?


taeold:
Fixed in https://github.com/firebase/firebase-tools/pull/4391/files

PR #4442, opened 2022-04-13
Adds blocking function discovery - Emulator

colerogers:
- Adds `prepareEndpoints` to check for violations and infer endpoint details. 
- Adds `performPostLoadOperations` to do operations after all triggers have loaded.
- Adds `addBlockingTrigger` to register a blocking trigger.

After the SDK changes have finalized, I'll update the `triggers-end-to-end-tests`.


src/emulator/functionsEmulator.ts
```diff
@@ -647,6 +675,44 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+          Authorization: "Bearer owner",
+        },
+        data: {
+          blockingFunctions: this.blockingFunctionsConfig,
+        },
+        json: true,
+      })
+      .then(() => {
+        return Promise.resolve();
+      })
```
taeold, line 706: This `then` doesn't seem useful. Is it serving some purpose I'm not aware of?

src/emulator/functionsEmulator.ts
```diff
@@ -647,6 +675,44 @@ export class FunctionsEmulator implements EmulatorInstance {
     }
   }
 
+  performPostLoadOperations(): Promise<void> {
```
taeold, line 678: nit* I think this function can be simplified if it were written as `async` function -no need to deal with promises or promise chains.

src/emulator/functionsEmulator.ts
```diff
@@ -795,6 +861,39 @@ export class FunctionsEmulator implements EmulatorInstance {
...
   }
 
+  addBlockingTrigger(url: string, blockingTrigger: BlockingTrigger): boolean {
+    logger.debug(`addBlockingTrigger`, JSON.stringify({ blockingTrigger }));
+
+    const eventType = blockingTrigger.eventType;
+    if (AUTH_BLOCKING_EVENTS.includes(eventType as any)) {
+      const accessToken = blockingTrigger.options!.accessToken as boolean;
+      const idToken = blockingTrigger.options!.idToken as boolean;
+      const refreshToken = blockingTrigger.options!.refreshToken as boolean;
```
taeold, line 871: nit* I generally like to keep variable declaration close to its use - so move this line to below or pull up `this.blockingFunctionsConfig.forwardInboundCredentials`?

src/emulator/functionsEmulator.ts
```diff
@@ -795,6 +861,39 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+        };
+      }
+      this.blockingFunctionsConfig.forwardInboundCredentials = {
+        accessToken,
+        idToken,
+        refreshToken,
+      };
+    }
+
+    return true;
```
taeold, line 894: I think this function needs to return false if it fails to add the blocking trigger for any reason. This could happen if the if-condition in L868 fails for some reason.

src/emulator/functionsEmulator.ts
```diff
@@ -647,6 +675,44 @@ export class FunctionsEmulator implements EmulatorInstance {
     }
   }
 
+  performPostLoadOperations(): Promise<void> {
+    if (
+      this.blockingFunctionsConfig.triggers?.beforeCreate?.functionUri === "" &&
+      this.blockingFunctionsConfig.triggers.beforeSignIn?.functionUri === ""
```
taeold, line 681:
```suggestion
      this.blockingFunctionsConfig.triggers.?beforeSignIn?.functionUri === ""
```

src/emulator/functionsEmulator.ts
```diff
@@ -647,6 +675,44 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+    ) {
+      return Promise.resolve();
+    }
+
+    const authEmu = EmulatorRegistry.get(Emulators.AUTH);
+    if (!authEmu) {
+      return Promise.resolve();
+    }
```
taeold, line 689: IIRC, functions emulator always starts before the auth emulator. Is there some mechanism to have the functions reloaded after auth emulator start or does it work in some other way?

taeold (commented):
overall lgtm, some questions and nits.

src/emulator/functionsEmulator.ts
colerogers: So the functions emulator gets the `start` method called before the auth emulator, but this code is executed in the `connect` stage which is always called after `start`, so we should be okay referencing auth here.

src/emulator/functionsEmulator.ts
colerogers: I was following convention from the other API requests in `functionsEmulator`, I'll remove it and make the whole function async

src/emulator/functionsEmulator.ts
colerogers: Good catch thanks

src/emulator/functionsEmulator.ts
colerogers: Sure, fixed

src/emulator/functionsEmulator.ts
colerogers: Ahh, I had them as separate vars from a previous iteration where I was combining them in this function, fixed it

src/emulator/functionsEmulator.ts
colerogers: 👍

taeold (approved):

lisajian (approved):

PR #4443, opened 2022-04-13
Fix comments from Auth Blocking Triggers PR

colerogers:
This changed addresses post merge comments from #4395


src/deploy/functions/services/auth.ts
```diff
@@ -143,7 +143,10 @@ export class AuthBlockingService implements Service {
    * Un-registers the auth blocking trigger from identity platform. If the endpoint uri is not on the resource, we do nothing.
    * @param ep the blocking endpoint
    */
-  unregisterTrigger(ep: backend.Endpoint & backend.BlockingTriggered): Promise<void> {
+  unregisterTrigger(ep: backend.Endpoint): Promise<void> {
+    if (!backend.isBlockingTriggered(ep)) {
+      return Promise.resolve(); // this should never happen
+    }
     this.triggerQueue = this.triggerQueue.then(() => this.unregisterTriggerLocked(ep));
```
taeold, line 150: Q: So I understand that we are building a chain of promises to make sure trigger registration happens sequentially.

But doesn't _something_ have to wait for all of these chain of promises are all fulfilled before we declare "hey deployment is done!". If we don't do this, then aren't we just banking on it being completed before the CLI process finishes (at which it will make sure all promises are fulfilled)?

taeold (approved):

src/deploy/functions/services/auth.ts
inlined: We return the promise below. By modifying the promise in place we're returning a promise for the latest enqueued operation. Fabricator waits on this promise before resolving setTrigger or unsetTrigger.

inlined (approved):

PR #4447, opened 2022-04-14
Add emulator support for multi codebase functions setup.

taeold:
Emulator will now load all functions discoverable in all codebases defined in user's repository.

The change ended up being fairly easy to implement, as we simply needed to `for` loop through the array of functions config and setup N emulatable backend (vs 1) for the FunctionsEmulator to load. This required a minor change on how triggers are invoked - the Functions Emulator will now take care of finding the correct emulatable backend associated w/ the trigger on invocation instead of taking it as an argument.

I've taken a separate PR to test this new feature of the emulator to make PR reviews slightly less burdensome:  https://github.com/firebase/firebase-tools/pull/4448


scripts/emulator-tests/functionsEmulator.spec.ts
```diff
@@ -131,12 +131,11 @@ function useFunctions(triggers: () => {}): void {
 
   // eslint-disable-next-line @typescript-eslint/unbound-method
   functionsEmulator.invokeTrigger = (
-    backend: EmulatableBackend,
```
joehan, line 134: Looks like we can get rid of the import for EMulatableBackend too.

joehan (approved):
This mostly LGTM! However, I think this breaks the backendInfo api slightly (https://github.com/firebase/firebase-tools/blob/master/src/emulator/functionsEmulator.ts#L931). That API assumes that everything that isn't part of an extension is part of one single emulatable backend for cf3, which is no longer true. Now, this API is only used by the extensions emulator UI, and so this is not important, but you may want to fix this in order to make future functions emulator UI work easier.

PR #4448, opened 2022-04-14
Use new codebase feature to split triggers for emulator e2e test 

taeold:
Emulator e2e test will now comprise 3 codebase:

1) Codebase that defines v1 triggers

2) Codebase that defines v2 triggers

3) Codebase that defines triggers that activate event/http triggers

This is strictly not necessary but a good way to exercise multi-codebase support in the emulator.

While implementing multi-codebase support, I found a small race condition in the Pubsub Emulator when registering trigger. Reorganizing where we `await` to squash that bug.


src/emulator/pubsubEmulator.ts
```diff
@@ -201,6 +209,7 @@ export class PubsubEmulator implements EmulatorInstance {
 
     this.ensureFunctionsClient();
 
+    console.log(JSON.stringify(triggers));
```
joehan, line 212: Clean this up?

joehan (approved):
LGTM once you get rid of the stray console log

PR #4449, opened 2022-04-15
Wire up CPU options

inlined:
Works for everything but 128MB functions with `gcf_gen1` CPU because we actually need 128MiB functions to qualify for that RAM amount. It doesn't seem like 128MB even works.


taeold (commented):
I'll review this PR once we get approval to move to MiB (vs MB)?

src/deploy/functions/release/fabricator.ts
```diff
@@ -337,12 +338,19 @@ export class Fabricator {
     }
 
     const mem = endpoint.availableMemoryMb || backend.DEFAULT_MEMORY;
-    if (mem >= backend.MIN_MEMORY_FOR_CONCURRENCY && endpoint.concurrency !== 1) {
-      await this.setConcurrency(
-        endpoint,
-        serviceName,
-        endpoint.concurrency || DEFAULT_GCFV2_CONCURRENCY
-      );
+    const hasCustomCPU = endpoint.cpu !== backend.memoryToGen1Cpu(mem);
```
taeold, line 341: having a brain fart, but should this be comparing the `endpoint.cpu` to `memoryToGen2Cpu` instead?

In english words, I would've guessed that:

```
hasCustomCPU = endpoint.cpu does not match default settings for Gen 2 memory -> CPU mapping.
```

src/deploy/functions/release/fabricator.ts
```diff
@@ -431,8 +439,24 @@ export class Fabricator {
         .catch(rethrowAs(endpoint, "set invoker"));
     }
 
-    if (endpoint.concurrency) {
-      await this.setConcurrency(endpoint, serviceName, endpoint.concurrency);
+    // Ideally we'd avoid a read of the Cloud Run service. We need to read if we
+    // believe a setting (CPU or concurrency) has changed because the user
+    // changed their mind or if GCF has stamped over the user's choice (we're
+    // using custom VM shapes).
+    const usingCustomCPU =
```
taeold, line 446: nit* `hasCustomCPU` for consistency with above.

src/deploy/functions/release/fabricator.ts
```diff
@@ -606,3 +647,38 @@ export class Fabricator {
...
+    logger.debug(
+      `Waiting for service ${service.metadata.name} to be ready. ` +
+        `Status is ${JSON.stringify(service.status?.conditions)}`
+    );
+    return false;
+  } else if (readyCondition?.status === "True") {
+    return true;
+  }
+  logger.debug(
+    `Service ${service.metadata.name} has unexpected ready status ${JSON.stringify(
```
taeold, line 677:
```suggestion
    `Service ${service.metadata.name} has unexpected status ${JSON.stringify(
```

src/deploy/functions/release/fabricator.ts
```diff
@@ -466,23 +490,40 @@ export class Fabricator {
...
+        service = await run.replaceService(serviceName, service);
+
+        // Now we need to wait for reconciliation or we might delete the docker
+        // image while the service is still rolling out a new revision.
+        let retry = 0;
+        while (!exports.serviceIsResolved(service)) {
+          await backoff(retry, 2, 30);
+          retry = retry + 1;
+          service = await run.getService(serviceName);
+        }
```
taeold, line 526: Isn't this an infinite loop? or are we expecting `serviceIsResolved` to throw an error at some point?

src/gcp/run.ts
```diff
@@ -113,6 +126,9 @@ export interface IamPolicy {
   etag?: string;
 }
 
+/**
+ *
+ */
```
taeold, line 131: nit* couple of unexpectedly empty docstring comments here and below.

src/test/deploy/functions/release/fabricator.spec.ts
```diff
@@ -767,55 +828,224 @@ describe("Fabricator", () => {
...
+          traffic: [],
+          url: "",
+          address: {
+            url: "",
+          },
+        },
+      };
+    });
+
+    it("returns false if the observed generation isn't hte metageneration", () => {
```
taeold, line 1015:
```suggestion
    it("returns false if the observed generation isn't the metageneration", () => {
```

src/test/deploy/functions/release/fabricator.spec.ts
```diff
@@ -767,55 +828,224 @@ describe("Fabricator", () => {
...
+    });
+
+    it("returns true if resolved", () => {
+      service.status!.observedGeneration = 2;
+      service.metadata.generation = 2;
+      service.status!.conditions[0].status = "True";
+      service.status!.conditions[0].type = "Ready";
+
+      expect(() => fabricator.serviceIsResolved(service)).to.throw;
+    });
```
taeold, line 1046: the test description and expectatoin doesn't seem to match here.

taeold (commented):
overall looks great, some comments to consider b/4 merging.

src/deploy/functions/prepare.ts
```diff
@@ -34,6 +34,9 @@ function hasUserConfig(config: Record<string, unknown>): boolean {
   return Object.keys(config).length > 1;
 }
 
+/**
+ *
+ */
```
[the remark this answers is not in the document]
inlined: Can't remove; linter just adds it back. Will just add a comment.

src/deploy/functions/release/fabricator.ts
inlined: I can change the name to be more clear or add a comment. This code is correct; gcf gen 1 is the "built in" memory that we get automatically and v2 is a default we use for our custom VM shapes.

src/deploy/functions/release/fabricator.ts
inlined: Sure

src/deploy/functions/release/fabricator.ts
inlined: serviceIsResolved returns true once observedGeneration == spec.generation and Condition[Type=Ready].State is True

src/deploy/functions/release/fabricator.ts
inlined: We're specifically talking about the Status of Type=Ready so I think the comment is correct.

src/gcp/run.ts
inlined: Will just document

src/test/deploy/functions/release/fabricator.spec.ts
inlined: Odd... That shouldn't pass. I'll dive deeper.

taeold (approved):

PR #4450, opened 2022-04-16
Add blocking function fetch logic

lisajian:
### Description

Adds logic to fetch configured blocking functions and perform fetch on impacted endpoints. 

Corresponding internal bug: b/192387243

### Scenarios Tested

`npm test` passes

### Sample Commands

N/A

### DIFF
--- a/src/emulator/auth/operations.ts
+++ b/src/emulator/auth/operations.ts
@@ -1,7 +1,7 @@
 import { URLSearchParams } from "url";
 import { decode as decodeJwt, sign as signJwt, JwtHeader } from "jsonwebtoken";
 import * as express from "express";
-import fetch, { Response } from "node-fetch";
+import fetch from "node-fetch";
 import AbortController from "abort-controller";
 import { ExegesisContext } from "exegesis-express";
 import {
@@ -154,7 +154,7 @@ async function signUp(
 ): Promise<Schemas["GoogleCloudIdentitytoolkitV1SignUpResponse"]> {
   assert(!state.disableAuth, "PROJECT_DISABLED");
   let provider: string | undefined;
-  const updates: Omit<Partial<UserInfo>, "localId" | "providerUserInfo"> = {
+  let updates: Omit<Partial<UserInfo>, "localId" | "providerUserInfo"> = {
     lastLoginAt: Date.now().toString(),
   };
 
@@ -229,11 +229,12 @@ async function signUp(
   let extraClaims;
   if (!user) {
     if (reqBody.email) {
-      await fetchBlockingFunctionWithTimeoutAndUpdateUser(
+      const fetchResponse = await fetchBlockingFunction(
         state,
         BlockingFunctionEvents.BEFORE_CREATE,
         updates
       );
+      updates = fetchResponse.user;
     }
     if (reqBody.localId) {
       user = state.createUserWithLocalId(reqBody.localId, updates);
@@ -243,11 +244,13 @@ async function signUp(
     }
   } else {
     if (reqBody.email) {
-      extraClaims = await fetchBlockingFunctionWithTimeoutAndUpdateUser(
+      const fetchResponse = await fetchBlockingFunction(
         state,
         BlockingFunctionEvents.BEFORE_SIGN_IN,
         updates
       );
+      updates = fetchResponse.user;
+      extraClaims = fetchResponse.extraClaims;
     }
     user = state.updateUserByLocalId(user.localId, updates);
   }
@@ -257,7 +260,7 @@ async function signUp(
     localId: user.localId,
     displayName: user.displayName,
     email: user.email,
-    ...(provider ? issueTokens(state, user, provider, extraClaims) : {}),
+    ...(provider ? issueTokens(state, user, provider, { extraClaims }) : {}),
   };
 }
 
@@ -1380,7 +1383,7 @@ async function signInWithEmailLink(
 
   state.deleteOobCode(reqBody.oobCode);
 
-  const updates: Omit<Partial<UserInfo>, "localId" | "providerUserInfo"> = {
+  let updates: Omit<Partial<UserInfo>, "localId" | "providerUserInfo"> = {
     email,
     emailVerified: true,
     emailLinkSignin: true,
@@ -1396,11 +1399,12 @@ async function signInWithEmailLink(
     if (userFromIdToken) {
       user = state.updateUserByLocalId(userFromIdToken.localId, updates);
     } else {
-      await fetchBlockingFunctionWithTimeoutAndUpdateUser(
+      const fetchResponse = await fetchBlockingFunction(
         state,
         BlockingFunctionEvents.BEFORE_CREATE,
         updates
       );
+      updates = fetchResponse.user;
       user = state.createUser(updates);
     }
   } else {
@@ -1408,11 +1412,13 @@ async function signInWithEmailLink(
     assert(!userFromIdToken || userFromIdToken.localId === user.localId, "EMAIL_EXISTS");
     user = state.updateUserByLocalId(user.localId, updates);
   }
-  const { extraClaims } = await fetchBlockingFunctionWithTimeoutAndUpdateUser(
+  const fetchResponse = await fetchBlockingFunction(
     state,
     BlockingFunctionEvents.BEFORE_SIGN_IN,
     updates
   );
+  updates = fetchResponse.user;
+  const extraClaims = fetchResponse.extraClaims;
   user = state.updateUserByLocalId(user.localId, updates);
 
   const response = {
@@ -1568,33 +1574,36 @@ async function signInWithIdp(
     oauthExpiresIn: coercePrimitiveToString(response.oauthExpireIn),
   };
   if (response.isNewUser) {
-    const updates = {
+    let updates: Partial<UserInfo> = {
       ...accountUpdates.fields,
       lastLoginAt: Date.now().toString(),
       providerUserInfo: [providerUserInfo],
       tenantId: state instanceof TenantProjectState ? state.tenantId : undefined,
     };
-    await fetchBlockingFunctionWithTimeoutAndUpdateUser(
+    const fetchResponse = await fetchBlockingFunction(
       state,
       BlockingFunctionEvents.BEFORE_CREATE,
       updates,
       oauthTokens
     );
+    updates = fetchResponse.user;
     user = state.createUser(updates);
     response.localId = user.localId;
   } else {
     if (!response.localId) {
       throw new Error("Internal assertion error: localId not set for exising user.");
     }
-    const updates = {
+    let updates = {
       ...accountUpdates.fields,
     };
-    extraClaims = await fetchBlockingFunctionWithTimeoutAndUpdateUser(
+    const fetchResponse = await fetchBlockingFunction(
       state,
       BlockingFunctionEvents.BEFORE_SIGN_IN,
       updates,
       oauthTokens
     );
+    extraClaims = fetchResponse.extraClaims;
+    updates = fetchResponse.user;
     user = state.updateUserByLocalId(response.localId, updates, {
       upsertProviders: [providerUserInfo],
     });
@@ -1617,7 +1626,7 @@ async function signInWithIdp(
     user = state.updateUserByLocalId(user.localId, { lastLoginAt: Date.now().toString() });
     return {
       ...response,
-      ...issueTokens(state, user, providerId, { signInAttributes, ...extraClaims }),
+      ...issueTokens(state, user, providerId, { signInAttributes, extraClaims }),
     };
   }
 }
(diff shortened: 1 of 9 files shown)


src/emulator/auth/operations.ts
```diff
@@ -225,14 +228,29 @@ function signUp(
...
     } else {
       user = state.createUser(updates);
     }
   } else {
+    if (reqBody.email) {
+      extraClaims = await fetchBlockingFunctionWithTimeoutAndUpdateUser(
+        state,
+        BlockingFunctionEvents.BEFORE_SIGN_IN,
+        updates
+      );
```
yuchenshi, line 252:
```suggestion
      ).extraClaims;
```

yuchenshi: nvm this works but just the variable name is misleading

src/emulator/auth/operations.ts
```diff
@@ -2830,6 +2904,246 @@ function updateTenant(
   return state.updateTenant(reqBody, ctx.params.query.updateMask);
 }
 
+// TODO: Timeout is 60s. Should we make the timeout an emulator configuration?
+async function fetchBlockingFunctionWithTimeoutAndUpdateUser(
```
lisajian, line 2908: - Split logic for fetching from logic that applies modifiable fields
- Rename to something shorter, e.g. `fetchBlockingFunction`
- Generally will want to return update instead of having side effect

yuchenshi (changes requested):
BLOCKING: I think passing `updates` to `fetchBlockingFunction` is too fragile especially given that `updates` are only a tiny fraction of touched fields on most sign-ins. Let's talk offline to brainstorm a better high-level design.

We also need some tests for what was passed *into* the blocking function (i.e. request body) -- that would have caught some issues caused by the statement above.

yuchenshi:
Please kindly re-request review when this is ready for another round of review.

src/emulator/auth/state.ts
```diff
@@ -345,6 +353,16 @@ export abstract class ProjectState {
     return this.getUserByLocalIdAssertingExists(localId);
   }
 
+  getAllPhoneNumbersByLocalId(localId: string): string[] {
```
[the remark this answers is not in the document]
lisajian: Ohh, good call, thanks!

src/emulator/auth/state.ts
```diff
@@ -57,6 +57,14 @@ export abstract class ProjectState {
 
   abstract get enableEmailLinkSignin(): boolean;
 
+  abstract get includeAccessToken(): boolean;
```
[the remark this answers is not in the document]
lisajian: That makes sense, updated!

lisajian:
Newest revision includes the following changes:
- Modified impacted endpoints to call blocking functions for new/existing users
- Refactored endpoints (e.g. signInWithEmailLink and signInWithPhoneNumber) so it's easier to see when there's a new vs existing user
- Changed `fetchBlockingFunction()` so it takes a user instead of an update and only returns the updated fields instead of modifying `UserInfo` in place
- Changed `BeforeCreate` calls so the requests contains a pending `localId` and as much of the `UserInfo` available as possible prior to writing to store
- Changed `BeforeSignIn` calls so requests contains the `UserInfo` object with any updates from `BeforeCreate` updates

Hopefully I addressed everything, but let me know if I'm missing anything! Thanks for taking a look

src/emulator/auth/operations.ts
```diff
@@ -2830,6 +2920,254 @@ function updateTenant(
...
+    assert(
+      userRecord.updateMask,
+      "BLOCKING_FUNCTION_ERROR_RESPONSE: ((Response UserRecord is missing updateMask.))"
+    );
+    const mask = userRecord.updateMask;
+    const fields = mask.split(",");
+
+    for (const field of fields) {
+      if (
+        Object.values(BlockingFunctionModifiableFields).includes(
```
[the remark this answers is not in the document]
yuchenshi: Blocking functions are user code and they can return anything they like. For example, a developer may be using plain JavaScript without type checking.

src/emulator/auth/operations.ts
```diff
@@ -223,12 +227,39 @@ function signUp(
...
+      );
+      updates = { ...updates, ...blockingResponse.updates };
+    }
+
+    user = state.createUserWithLocalId(localId, updates);
+    assert(user, "DUPLICATE_LOCAL_ID");
+
+    if (reqBody.email) {
+      if (!user.disabled) {
+        const blockingResponse = await fetchBlockingFunction(
```
yuchenshi, line 250: I believe the current logic calls the `beforeSignIn` blocking function even for an admin request (e.g. from AdQmin SDKs or Emulator UI). Is this the right behavior?

src/emulator/auth/operations.ts
```diff
@@ -1393,23 +1462,23 @@ function signInWithEmailLink(
     isNewUser,
   };
 
-  if (
-    (state.mfaConfig.state === "ENABLED" || state.mfaConfig.state === "MANDATORY") &&
-    user.mfaInfo?.length
-  ) {
+  if (isMfaEnabled(state, user)) {
```
yuchenshi, line 1465: For disabled users with MFA, should the API also return an error here?

src/emulator/auth/operations.ts
```diff
@@ -2800,6 +2994,276 @@ function updateTenant(
...
+    const res = await fetch(url, {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify(reqBody),
+      signal: controller.signal,
+    });
+    const text = await res.text();
+    assert(
+      res.ok,
+      `BLOCKING_FUNCTION_ERROR_RESPONSE: ((HTTP request to ${url} returned an error: ${text}))`
```
yuchenshi, line 3057:
```suggestion
      `BLOCKING_FUNCTION_ERROR_RESPONSE: ((HTTP request to ${url} returned HTTP error ${res.status}: ${text}))`
```

src/emulator/auth/operations.ts
```diff
@@ -2800,6 +2994,276 @@ function updateTenant(
...
+    const mask = userRecord.updateMask;
+    const fields = mask.split(",");
+
+    for (const field of fields) {
+      switch (field) {
+        case "displayName":
+        case "photoUrl":
+        case "disabled":
+        case "emailVerified":
+          (updates as any)[field] = userRecord[field];
```
yuchenshi, line 3102: Is `as any` still needed? If so, maybe try `updates[field as keyof BlockingFunctionUpdates]`

src/emulator/auth/operations.ts
```diff
@@ -2800,6 +2994,276 @@ function updateTenant(
...
+          break;
+        case "customClaims":
+          validateSerializedCustomClaims(userRecord.customClaims!);
+          updates.customAttributes = userRecord.customClaims;
+          break;
+        // Session claims are only returned in beforeSignIn. For more info,
+        // see https://cloud.google.com/identity-platform/docs/blocking-functions#modifying_a_user
```
yuchenshi, line 3109: If a developer makes a mistake and returns `sessionClaims` from `beforeCreate`, the implementation will try to parse it instead of ignores it. Is there anyway we can address this?

src/emulator/auth/operations.ts
```diff
@@ -2830,6 +2920,254 @@ function updateTenant(
...
+        email: providerUserInfo.email,
+        uid: providerUserInfo.providerId,
+      };
+      provider_data.push(provider);
+    }
+  }
+  if (user.localId) {
+    const allPhoneNumbers = state.getAllPhoneNumbersByLocalId(user.localId);
+    for (const phoneNumber of allPhoneNumbers) {
+      const provider: Provider = {
```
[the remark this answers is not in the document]
yuchenshi: I think the Auth Emulator implementation is funkier than the internal link -- since the Auth Emulator tries to keep `user.providerUserInfo` synced with phone numbers when writing to DB, Line 3191 will transform that entry into `provider_data`. The logic here will probably add yet another entry to `provider_data`, creating a duplicate. Feel free to add a test to prove me wrong.

src/emulator/auth/operations.ts
```diff
@@ -2800,6 +2994,276 @@ function updateTenant(
...
+      provider_data.push(provider);
+    }
+  }
+  if (user.localId) {
+    const allPhoneNumbers = user.providerUserInfo
+      ? user.providerUserInfo.filter((info) => !!info.phoneNumber).map((info) => info.phoneNumber)
+      : [];
+    for (const phoneNumber of allPhoneNumbers) {
+      const provider: Provider = {
+        provider_id: PROVIDER_PASSWORD,
```
yuchenshi, line 3200:
```suggestion
        provider_id: PROVIDER_PHONE,
```

src/emulator/auth/operations.ts
```diff
@@ -2830,6 +2920,254 @@ function updateTenant(
...
+        uid: phoneNumber,
+        phone_number: phoneNumber,
+      };
+      provider_data.push(provider);
+    }
+  }
+  if (user.email && (user.passwordHash || user.emailLinkSignin)) {
+    const provider: Provider = {
```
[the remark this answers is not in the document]
yuchenshi: Same concern as above

src/emulator/auth/state.ts
```diff
@@ -53,14 +54,21 @@ export abstract class ProjectState {
...
     for (let i = 0; i < 10; i++) {
       // Try this for 10 times to prevent ID collision (since our RNG is
       // Math.random() which isn't really that great).
       const localId = randomId(28);
-      const user = this.createUserWithLocalId(localId, props);
-      if (user) {
-        return user;
+      if (!this.users.has(localId) && !this.pendingLocalIds.has(localId)) {
+        // Create a pending localId until user is created
+        this.pendingLocalIds.add(localId);
```
yuchenshi, line 70: This creates a memory leak if the blocking functions throws and the `localId` is never used. (Theoretically, if this happens enough times we will even run out of IDs.) I think this is probably acceptable, but worth a comment to document the tradeoff.

yuchenshi (commented):
Partial LGTM. I have not reviewed the tests line by line yet, but I believe we're still lacking coverage for blocking functions _request_ payloads. See comments inline.

src/emulator/auth/operations.ts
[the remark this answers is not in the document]
lisajian: Ah true, I'm not sure what the backend behavior would be but I added a bit of logic to coerce both displayName and photoUrl to string and similarly for the booleans

src/test/emulators/auth/mfa.spec.ts
```diff
@@ -462,4 +469,210 @@ describeAuthEmulator("mfa enrollment", ({ authApi, getClock }) => {
...
+        {
+          blockingFunctions: {},
+        },
+        "blockingFunctions"
+      );
+      expect(nock.isDone()).to.be.true;
+      nock.cleanAll();
+    });
+
+    it("mfaSignIn:finalize should update modifiable fields before sign in", async () => {
```
yuchenshi, line 487: nit: add a test that blocking function is not called if MFA is present on user but *not* performed, and `displayName` is not updated.

src/test/emulators/auth/mfa.spec.ts
```diff
@@ -462,4 +469,210 @@ describeAuthEmulator("mfa enrollment", ({ authApi, getClock }) => {
...
+
+      const { mfaPendingCredential, mfaEnrollmentId } = await authApi()
+        .post("/identitytoolkit.googleapis.com/v1/accounts:signInWithPassword")
+        .query({ key: "fake-api-key" })
+        .send({ email, password })
+        .then((res) => {
+          expectStatusCode(200, res);
+          expect(res.body).not.to.have.property("idToken");
+          expect(res.body).not.to.have.property("refreshToken");
+          const mfaPendingCredential = res.body.mfaPendingCredential as string;
```
yuchenshi, line 504: nit: extract a helper function for extracting mfaPendingCredential and mfaEnrollmentId from response (mainly to clean the code up visually, since the exact response format is not the focus of the test here)

src/test/emulators/auth/signUp.spec.ts
```diff
@@ -572,4 +581,273 @@ describeAuthEmulator("accounts:signUp", ({ authApi }) => {
...
+          expect(res.body.displayName).to.equal(DISPLAY_NAME);
+          expect(decoded!.payload.name).to.equal(DISPLAY_NAME);
+          expect(decoded!.payload.picture).to.equal(PHOTO_URL);
+          expect(decoded!.payload.email_verified).to.be.true;
+          expect(decoded!.payload).to.have.property("customAttribute").equals("custom");
+          expect(decoded!.payload).to.have.property("sessionAttribute").equals("session");
+        });
+    });
+
+    it("should disable email/password user if set", async () => {
```
yuchenshi, line 771:
```suggestion
    it("should disable new user if set", async () => {
```

src/test/emulators/auth/signUp.spec.ts
```diff
@@ -572,4 +581,273 @@ describeAuthEmulator("accounts:signUp", ({ authApi }) => {
...
+          expect(res.body).not.to.have.property("idToken");
+          expect(res.body).not.to.have.property("refreshToken");
+
+          expect(res.body.displayName).to.equal(displayName);
+          expect(res.body.localId).to.be.a("string").and.not.empty;
+        });
+
+      // Shouldn't trigger nock calls
+      expect(nock.isDone()).to.be.false;
+      nock.cleanAll();
```
yuchenshi, line 850: This should be in a `afterEach` block so that the cleanup still happens even if the test fails somehow

yuchenshi (approved):
LGTM! Please add at least one test for the blocking function request payload though (i.e. the `jwt` that is sent to the blocking function). This reduces chances of regression in the future.

src/emulator/auth/operations.ts
lisajian: Sick, updated!

lisajian:
> LGTM! Please add at least one test for the blocking function request payload though (i.e. the `jwt` that is sent to the blocking function). This reduces chances of regression in the future.

Added to signInWithEmailLink

PR #4451, opened 2022-04-16
Add support for secrets in gcfv2 endpoints

taeold:
We now read `secretEnvironmentVariables` property from the endpoint manifest and include the property when creating/updating v1 and v2 functions.

Side note - if secretEnvVar in the manifest is missing `secret` property (i.e. doesn't specify the secret resource name), we'll assume that resource name matches the `key` property. This will be the case for all CF3 functions.


inlined (approved):

colerogers (approved):

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -243,6 +243,24 @@ function parseEndpoints(
...
+      parsed,
+      ep,
+      "secretEnvironmentVariables",
+      "secretEnvironmentVariables",
+      (senvs: ManifestEndpoint["secretEnvironmentVariables"]) => {
+        if (senvs && senvs.length > 0) {
+          ep.secretEnvironmentVariables = [];
+          for (const { key, secret } of senvs) {
+            ep.secretEnvironmentVariables.push({
+              key: key,
```
TheIronDev, line 256: Nit, you can use shorthand instead: 

```
{
key,
}
```

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -243,6 +243,24 @@ function parseEndpoints(
...
+              secret: secret || key, // if secret is undefined, assume env var key == secret name
+              projectId: project,
```
TheIronDev, line 258: 2 Nits (feel free to ignore)

* rename `project` to `projectId` (its more descriptive)
* if you rename `project` to `projectId`, then you can also do shorthand here

src/deploy/functions/validate.ts
```diff
@@ -140,7 +140,7 @@ export async function secretsAreValid(projectId: string, wantBackend: backend.Ba
  * Ensures that all endpoints specifying secret environment variables target platform that supports the feature.
  */
 function validatePlatformTargets(endpoints: backend.Endpoint[]) {
-  const supportedPlatforms = ["gcfv1"];
+  const supportedPlatforms = ["gcfv1", "gcfv2"];
```
TheIronDev, line 143: Nit:

```
const supportedPlatforms = new Set(["gcfv1", "gcfv2"]);
const unsupported = endpoints.filter((e) => !supportedPlatforms.has(e.platform));
```

Nit x2: I don't know how often this is called, but maybe it makes more sense to have this defined as a const up top?

src/gcp/cloudfunctionsv2.ts
```diff
@@ -92,6 +92,13 @@ export interface EventFilter {
   value: string;
 }
 
+export interface SecretEnvVar {
```
TheIronDev, line 95: Nit: can you add documentation for what this is and what each of the fields do?

src/gcp/cloudfunctionsv2.ts
```diff
@@ -104,6 +111,7 @@ export interface ServiceConfig {
   timeoutSeconds?: number;
   availableMemory?: string;
   environmentVariables?: Record<string, string>;
+  secretEnvironmentVariables?: Array<SecretEnvVar>;
```
TheIronDev, line 114: Nit: `SecretEnvVar[]` ?

TheIronDev (approved):

src/deploy/functions/runtimes/discovery/v1alpha1.ts
taeold: I'm going to prefer to keep `project` here since that's the name of the property we use to store project identifiers in `backend.Endpoint`. Is that a good idea? I dunno, but that change would touch LOTS of files and I'd suggest that we do that in a separate PR if we want to make that change.

src/deploy/functions/validate.ts
taeold: You and @inlined should have a talk! @inlined does not like Sets, especially in our code where N is almost always < 50 where using Set (which has every so slightly different API from arrays) doesn't give us much!

src/gcp/cloudfunctionsv2.ts
taeold: huh I finally read go/tsstyle#arrayt-type. Sounds like that is perferable.

src/deploy/functions/validate.ts
TheIronDev: Team convention is king. :)

That said, my thoughts are its less about efficiency and more about intention/code clarity.

`Set` limits the purpose of the variable/code to just checking for existence.

`Array` leaves the implementation open ended. Maybe the array going to be used with `has`, `reduce`, or `map`. Someone else won't know unless they read through the code to find how/why the array is being used.

PR #4460, opened 2022-04-20
Read/write events envvars in manifest export + emulate flows

huangjeff5:
### Description

Update planner.ts (have/want methods to incorporate events configuration)
Update ext:export command to write env config to environment variables
Update manifest-based emulate mode to set events-related config in environment variables while emulating functions 

[API Proposal](https://docs.google.com/document/d/1qeM93yB3tvWjRSQboDX0IYbRKNeEeSTjt8OuvVjOtYg/edit#heading=h.x94y9zzc13xn)


src/extensions/askUserForEventsConfig.ts
```diff
@@ -0,0 +1,112 @@
+import * as _ from "lodash";
+import { onceWithJoin } from "./utils";
```
joehan, line 2: I think this is unused now

src/extensions/askUserForEventsConfig.ts
```diff
@@ -0,0 +1,112 @@
+import * as _ from "lodash";
```
joehan, line 1: This looks unused.

joehan (approved):
One nit about unused deps, but LGTM otherwise.
Note: this was also thoroughly review on Private

Issue #4412, opened 2022-04-07
Firebase Functions Emulator Considers 0.0.0.0 An External Network Resource


marcfrankel:
### [REQUIRED] Environment info

**firebase-tools:** 10.6.0

**Platform:** macOS

### [REQUIRED] Test case

If your Storage emulator is bound to port 0.0.0.0 and you connect to it

### [REQUIRED] Steps to reproduce

In your Firebase.json file bind the storage emulator to the usual port but to host: "0.0.0.0" instead of "localhost" or "127.0.0.1"

Then perform any storage operation where you request that local bucket/instance (but bound to 0.0.0.0) URL.

### [REQUIRED] Expected behavior

The expected behavior is that the IP address 0.0.0.0 or 127.0.0.1 should also be considered an internal/local address and should not trigger this warning.

I believe the code in question causing this small bug is here: https://github.com/firebase/firebase-tools/blob/7271c958eea82b8d223c0e581a94e3b5b7e57619/src/emulator/functionsEmulatorRuntime.ts#L368

In case you're curious our use case for why we need to bind to 0.0.0.0 is to be able to connect to the local emulator instance from React Native on our physical phones.

### [REQUIRED] Actual behavior

Output similar to the following from the emulator command line:
`⚠  External network resource requested! - URL: "http://0.0.0.0:9199/b/local_bucketg/o/CT6BO6EQzr6kzo61Ky2hQEcMCjJy%2FC99C62D6-ED0D-4DBE-94CD-BAEF6FC30AE6.png?"`

I will admit this is more of just a tiny annoyance, but it also seems like a simple fix by just adding more to the above line's if statement.

joehan:
Hey @marcfrankel, thanks for reporting this (and providing a code point!).  I'll make a PR to fix this shortly.

tonyjhuang:
Reopening to track better error messaging

taeold:
I think my issue that 0.0.0.0 can technically refer to anything. 0.0.0.0. It's a wildcard, and some user may use it connect to external services.

I understand your situation and your mild annoyance, but since the implementation can go either way I'm going to choose the lazy way and keep the status quo. Feel free to disagree but also feel free to make a contribution. I'll escalate the issue to a wider audience internally to see what the "right" implementation then.

PR #4462, opened 2022-04-20
Improving localhost detection for functions emulator

joehan:
### Description
Detect calls to 127.0.0.1 as localhost and don't display external HTTP call warnings for them. This (partly) addresses #4412 - however, I excluded 0.0.0.0 from this because we should not be using it as a destination address ever (see http://whois.arin.net/rest/nets;q=0.0.0.0?showDetails=true). In a separate PR, we'll block 0.0.0.0 and add messaging telling users to use 127.0.0.1 or localhost instead.


src/emulator/functionsEmulatorUtils.ts
```diff
@@ -131,3 +131,10 @@ export function compareVersionStrings(a?: string, b?: string) {
 
   return 0;
 }
+
+/**
+ * Check if a url is localhost
+ */
+export function isLocalHost(href: string): boolean {
+  return !!href.match(/^(http(s)?:\/\/)?(localhost|127.0.0.1)/);
```
yuchenshi, line 139:
```suggestion
  return !!href.match(/^(http(s)?:\/\/)?(localhost|127.0.0.1|\[\::1])/);
```

yuchenshi (approved):

PR #4464, opened 2022-04-21
Supporting events flows in extensions emulator

huangjeff5:
### Description

This is a feature branch used to hold all extensions emulator features related to events.


src/emulator/functionsEmulator.ts
```diff
@@ -1119,6 +1156,11 @@ export class FunctionsEmulator implements EmulatorInstance {
       process.env.PUBSUB_EMULATOR_HOST = pubsubHost;
     }
 
+    const eventarcEmulator = this.getEmulatorInfo(Emulators.EVENTARC);
+    if (eventarcEmulator) {
+      process.env.CLOUD_EVENTARC_EMULATOR_HOST = `http://${formatHost(eventarcEmulator)}`;
```
huangjeff5, line 1161: Should this be CLOUD_EVENTARC_EMULATOR_HOST or EVENTARC_EMULATOR_HOST?

apascal07: CLOUD_EVENTARC_EMULATOR_HOST, resolved in sync!

src/emulator/eventarcEmulator.ts
```diff
@@ -0,0 +1,169 @@
+import * as _ from "lodash";
```
joehan, line 1: This is unused - looks like it slipped in accidentally?

src/emulator/eventarcEmulator.ts
```diff
@@ -0,0 +1,169 @@
...
+import * as express from "express";
```
joehan, line 2: Nit - I prefer to add a newline between imports from external deps (like express) and imports from other files in this project

src/emulator/eventarcEmulator.ts
```diff
@@ -0,0 +1,169 @@
...
+import { CloudEvent } from "./events/types";
+import { EmulatorRegistry } from "./registry";
+
+interface CustomEventTrigger {
+  projectId: string;
+  triggerName: string;
+  eventTrigger: EventTrigger;
+}
+
+interface RequestWithRawBody extends express.Request {
```
joehan, line 19: In firebase-functions, we have this exact same type and we call it Request. For consistency, can we just call this Request (or import the firebase-functions type?)

apascal07: I believe other emulators (e.g. Functions emulator) also call it RequestWithRawBody which is what this is consistent with. What do you think?

joehan: Ahh, good point - you're right that its better to be consistent within this repo than with firebase-functions. Good call.

src/emulator/eventarcEmulator.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  private logger = EmulatorLogger.forEmulator(Emulators.EVENTARC);
+  private customEvents: { [key: string]: CustomEventTrigger[] } = {};
+
+  constructor(private args: EventarcEmulatorArgs) {}
+
+  createHubServer(): express.Application {
+    const registerTriggerRoute = `/emulator/v1/projects/:project_id/triggers/:trigger_name(*)`;
+    const registerTriggerHandler: express.RequestHandler = (req, res) => {
+      const projectId = req.params.project_id;
```
joehan, line 39: Should we be returning an error if these are missing on the request?

src/emulator/eventarcEmulator.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  async start(): Promise<void> {
+    const { host, port } = this.getInfo();
+    const server = this.createHubServer().listen(port, host);
+    this.destroyServer = createDestroyer(server);
+    return Promise.resolve();
+  }
+
+  async connect(): Promise<void> {
+    // wip. what to do here?
```
joehan, line 146: Probably nothing - in other emulators, we use this to set up any file watchers we need (ie to watch for changes to security rules in Storage/RTDB, or to watch for code changes in functions).

Can just replace this with `return Promise.resolve()`

src/emulator/functionsEmulator.ts
```diff
@@ -271,7 +271,7 @@ export class FunctionsEmulator implements EmulatorInstance {
 
     // The URL for the function that the other emulators (Firestore, etc) use.
     // TODO(abehaskins): Make the other emulators use the route below and remove this.
-    const backgroundFunctionRoute = `/functions/projects/:project_id/triggers/:trigger_name`;
+    const backgroundFunctionRoute = `/functions/projects/:project_id/triggers/:trigger_name(*)`;
```
joehan, line 274: why did we need this change"

joehan (approved):
Some small fixes requested, but LGTM!

src/emulator/functionsEmulator.ts
apascal07: trigger_name is of format <function name>-<fully qualified channel name with forward slashes> so it doesn't parse the path correctly unless we specify that the rest of the path should be used for trigger_name.

apascal07: Err, tags got stripped. Meant to say of format "function name" - "fully qualified channel path with forward slashes".

PR #4474, opened 2022-04-25
Use mebibytes over hybrid mega/mebi

inlined:
V1 always used "Mebibytes" (base 2) but used the term "Megabytes" (ambiguous base 2 or 10) for a unitless "availableMemoryMb". In v2, the available memory field has units. In those units, "M" is base 10 and "Mi" is base 2.

This corrects the previous erroneous behavior where we had base 2 quantities and base 10 units.


src/gcp/cloudfunctionsv2.ts
```diff
@@ -386,6 +386,9 @@ export async function deleteFunction(cloudFunction: string): Promise<Operation>
   }
 }
 
+/**
+ *
+ */
```
colerogers, line 391: Empty ts doc comment

src/gcp/cloudfunctionsv2.ts
```diff
@@ -507,6 +516,9 @@ export function functionFromEndpoint(endpoint: backend.Endpoint, source: Storage
   return gcfFunction;
 }
 
+/**
+ *
+ */
```
colerogers, line 521: Empty ts doc comment

colerogers (changes requested):
lgtm, delete the empty comments before merge

inlined (commented):
Yeah, I don't know why my lint fixer is adding empty comments. Fixed.

colerogers (approved):

Issue #4414, opened 2022-04-07
Invalid Refresh token in emulator with Identity Platform


Stf-F:
### [REQUIRED] Environment info

**firebase-tools:** 10.2.0

**Platform:** Mac OS 12.1
 **Firebase SDK version:** 9.6.2
 **Node:** 16.13.1
**npm:** 8.1.2 (via nvm use 16)

### [REQUIRED] Test case

```javascript
      const unsubscribe = onAuthStateChanged(
        auth,
        async (user) => {
          if (user) {
              try {
                await user?.getIdToken(true); // <------ This yields token refresh errors
              } catch (e) {
                throw e;
            }
            // rest of logic here
          }
        },
      );
  ```

### [REQUIRED] Steps to reproduce
On localhost, against the Auth emulators: 
- In a Firebase project with firebase Auth working, upgrade to Identity Platform  
- Generate a sign in link on the server side via the admin SDK with the tenant ID ( [as per the official docs](https://cloud.google.com/identity-platform/docs/multi-tenancy-managing-tenants#generating_email_action_links))
- Authenticate with the link
- (Update the claims on the server side)
- Invoke the `getIdToken(true)` to force refresh the token
- Expect a crash

### [REQUIRED] Expected behavior

-  Being able to use the `getIdToken(true)` against the emulator when signing in with Identify Platform tokens.  

### [REQUIRED] Actual behavior

![image](https://user-images.githubusercontent.com/52630104/162050548-add3161d-ee12-4afe-8dd2-3323d12efb7b.png)

yuchenshi:
Looks like the token refresh endpoint isn't tenant ID aware (yet). @lisajian for further triaging.

Stf-F:
> Looks like the token refresh endpoint isn't tenant ID aware (yet). @lisajian for further triaging.

@yuchenshi, please also note that tenanted users created in localhost do not appear in the 'Authentication' tab of the emulator. Yet, they do exist and are successfully created as a [getUserByEmail()](https://cloud.google.com/identity-platform/docs/multi-tenancy-managing-tenants#getting_a_user) would successfully fetch them. Let me know, I am happy to file another bug report for this related issue if necessary.

yuchenshi:
@Stf-F The Emulator UI does not support multi-tenancy at this moment. We'll let you know if / when support is added.

lisajian:
[...]

PR #4475, opened 2022-04-25
Make grantToken tenant-aware

lisajian:
### Description

Adds multi-tenancy support on grantToken endpoint by pulling tenant ID from refresh token. Changes include:
- Make refresh token minting / validating stateless by serializing RefreshTokenRecord instead of representing with random string 
- Changing server side logic to check request body's refresh token for tenant ID (note that `grantToken` is the only endpoint in which the tenant ID cannot be obtained from the request path, query params, or ID token in the request body)
- Added additional unit tests for `grantToken` edge cases

Fixes https://github.com/firebase/firebase-tools/issues/4414. 

### Scenarios Tested

`npm test` passes

### Sample Commands

N/A

### DIFF
--- a/src/emulator/auth/state.ts
+++ b/src/emulator/auth/state.ts
@@ -27,8 +27,6 @@ export abstract class ProjectState {
   private localIdForPhoneNumber: Map<string, string> = new Map();
   private localIdsForProviderEmail: Map<string, Set<string>> = new Map();
   private userIdForProviderRawId: Map<string, Map<string, string>> = new Map();
-  private refreshTokens: Set<string> = new Set();
-  private refreshTokensForLocalId: Map<string, Set<string>> = new Map();
   private oobs: Map<string, OobRecord> = new Map();
   private verificationCodes: Map<string, PhoneVerificationRecord> = new Map();
   private temporaryProofs: Map<string, TemporaryProofRecord> = new Map();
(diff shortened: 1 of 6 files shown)


src/emulator/auth/state.ts
```diff
@@ -872,9 +855,11 @@ export type Config = {
 };
 
 interface RefreshTokenRecord {
+  _AuthEmulatorRefreshTokenRecord: string;
```
yuchenshi, line 858: Drop `Record`

src/emulator/auth/state.ts
```diff
@@ -922,6 +907,19 @@ interface TemporaryProofRecord {
...
 }
 
+export function encodeRefreshToken(refreshTokenRecord: RefreshTokenRecord): string {
+  return Buffer.from(JSON.stringify(refreshTokenRecord), "utf8").toString("base64");
+}
+
+export function decodeRefreshToken(refreshTokenString: string): RefreshTokenRecord {
+  try {
+    const json = Buffer.from(refreshTokenString, "base64").toString("utf8");
+    return JSON.parse(json) as RefreshTokenRecord;
```
yuchenshi, line 917: Let's validate the object has the `_AuthEmulatorRefreshToken` field to prevent misuse of other types of tokens, e.g. custom tokens, fake IDP tokens, mfa pending credentials

src/emulator/auth/state.ts
lisajian: Done!

src/emulator/auth/state.ts
lisajian: Cool, missed that assertion, thanks!

yuchenshi (approved):

yuchenshi:
CHANGELOG please

PR #4476, opened 2022-04-25
Adding deprecation warning for --params flag

joehan:
### Description
Adds a deprecation warning for the `--params` flag. This flag is being removed in firebase-tools v11, in favor of using extension manifest + `firebase deploy --only extensions`


src/extensions/warnings.ts
```diff
@@ -119,3 +120,14 @@ export async function displayWarningsForDeploy(instancesToCreate: InstanceSpec[]
   }
   return experimental.length > 0 || eapExtensions.length > 0;
 }
+
+/**
+ * paramsFlagDeprecationWarning displays a warning about the future depreaction of the --params flag.
+ */
+export function paramsFlagDeprecationWarning() {
```
elvisun, line 127: maybe showParamsFlagDeprecationWarning

elvisun (approved):

PR #4477, opened 2022-04-26
Fix broken caching in ContainerCleaner

inlined:
This was a subtle one. Apologies to all the bugs that I've let languish. I was both pressed to make sure we met I/O deadlines and thought we'd stop using Google Container Registry months ago. It wasn't until I was pouring over debug logs that I noticed the same file was being listed over and over.

`async` is nifty, but the result is returned in a different iteration of the runloop than the argument. Given the line

The following code is broken:

```
let cache: T = undefined;

function getThing(): Promise<T> {
  if (!cache) {
    // bug here: cache can be assigned to a year after getThingAsync() was called
    // we'll call getThingsAsync several times until the first one resolves and we
    // finally update the cache
    cache = await getThingAsync();
  }
  return cache;
}
```

In this case, you need to actually be caching the Promise itself so that subsequent calls will all block on the first call of `getThingsAsync` to resolve.

Verified the fix reduces calls with:

```
❯ firebase deploy --force --debug --only functions | grep "GET https://us.gcr.io" | wc -l
      27
# Rebuild firebase-tools with the fix
❯ firebase deploy --force --debug --only functions | grep "GET https://us.gcr.io" | wc -l
       9
```

Unfortunately I think this bug caused a polynomial increase in REST API calls. Reading N functions caused N tree traversals of a tree width N. This will take out an order of magnitude API calls and address both deploy time and bandwidth concerns.


src/deploy/functions/containerCleaner.ts
```diff
@@ -360,20 +360,24 @@ export interface Stat {
...
-        children: raw.child,
-      };
+    if (!(path in this.cache)) {
+      this.cache[path] = retry(() => this.client.listTags(path)).then((res) => {
+        return {
+          tags: res.tags,
+          digests: Object.keys(res.manifest),
+          children: res.child,
+        };
+      });
```
Youmoo, line 380: Sorry to bother. I think using `cache = Object.create(null)` is safer. What if `path` here is `toString`?

taeold: That would also work, but I'm not sure what you mean with "`path` here is `toString`"! Can you elaborate?

taeold (approved):

src/deploy/functions/containerCleaner.ts
Youmoo: Sure. pass `toString` as the argument to the method `ls(path)`.

inlined: It seems a little overboard, but sure, why not.

Issue #4221, opened 2022-02-25
Unmet peer dependency warning (marked-terminal -> marked)


ebeloded:
### [REQUIRED] Environment info

**firebase-tools:** 10.2.1

**Platform:** macOS

### [REQUIRED] Test case

Installing firebase tools produces peer dependency warning

![CleanShot 2022-02-25 at 11 06 18](https://user-images.githubusercontent.com/5859177/155651821-75bae13f-6c19-402c-bda3-b6c9b916d825.png)

### [REQUIRED] Steps to reproduce

```pnpm install firebase-tools```

### [REQUIRED] Expected behavior

Expected to see no warning

### [REQUIRED] Actual behavior

Seeing the warning

joehan:
Thanks for reporting this! Upgrading to the latest version of marked-terminal would fix this - however, the latest version of marked-terminal requires node 14 & this repository is still on node 12 due to other dependencies. When we upgrade the node version for firebase-tools, we will definitely upgrade to the latest version of marked-terminal as well.

Fortunately, for now, this warning is nothing to worry about. We've verified that everything that we use marked & marked-terminal for is rendering properly. If you notice any places where markdown is being rendered strangely, please open a new issue to let us know. Otherwise, I'll leave this issue open for now as a reminder to fix this when able.

bkendall:
This will be fixed by #4481 when we release the new breaking version of the CLI. That won't be for a little bit, but Soon(tm)

PR #4481, opened 2022-04-26
update marked and npm

bkendall:
### Description

Now that we've upgraded node, I'm upgrading marked & marked-terminal so that npm@8.7 can work without issues. Pinning the version of npm in our processes to 8.7 for sanity, but should be good to go.


joehan (approved):
Love to see it - this also closes #4221

kossnocorp:
This PR makes `firebase-tools` incompatible with Node.js v15 and below. See: https://github.com/mikaelbr/marked-terminal/pull/127

kossnocorp:
@bkendall thanks, you're right.

PR #4485, opened 2022-04-27
Limited CEL evaluator for Functions params

Berlioz:
Handles the subset of CEL defined in the container contract doc with only one function entry point, but will require the caller to override the type of the returned value with `as`, i think.


src/deploy/functions/cel.ts
```diff
@@ -0,0 +1,176 @@
...
+export type IdentityExpression = CelExpression;
+export type EqualityExpression = CelExpression;
+export type TernaryExpression = CelExpression;
+
+export type Literal = string | number | boolean;
+type L = "string" | "number" | "boolean";
+
+const identityRegexp = /{{ params\.(\S+) }}/;
+const equalityRegexp = /{{ params\.(\S+) == (.+) }}/;
+const ternaryRegexp = /{{ params\.(\S+) == (.+) ? (.+) : (.+) }/;
```
inlined, line 14: Missing the ending curly

inlined, line 14: Ternary expressions should not be required to also be an equality expression. It's totally valid to have:

```
const isProd = defineBool("IS_PROD", {
  default: projectId.eq("myProject"),
});

export const myFunc = onRequest({
  minIntsances: isProd.then(10, 0),
}, (req, res) = > res.send("OK"));
```

src/deploy/functions/cel.ts
```diff
@@ -0,0 +1,176 @@
...
+    throw new FirebaseError("malformed CEL equality expression '" + expr + "'");
+  }
+
+  const lhs = params[match[1]];
+  let rhs: Literal;
+  if (typeof lhs === "undefined") {
+    throw new FirebaseError(
+      "CEL equality expression LHS '" + match[1] + "' was not resolvable to a param"
+    );
+  } else if (typeof lhs === "string") {
```
inlined, line 79: Can you not say `rhs = resolveLiteral(typeof lhs, match[2])`?

src/deploy/functions/cel.ts
```diff
@@ -0,0 +1,176 @@
...
+
+function resolveLiteral(wantType: L, value: string): Literal {
+  if (paramRegexp.exec(value)) {
+    throw new FirebaseError(
+      "CEL tried to evaluate param." + value + " in a context which only permits literal values"
+    );
+  }
+
+  if (wantType === "number") {
+    if (isNaN(parseInt(value))) {
```
inlined, line 154: don't parseInt. Technically speaking we should probably be figuring out how to disambiguate between ints and floats as params since they're different types. At minimum we shouldn't parseInt because `cpu` can be a fractional amount.

inlined (commented):
Just a few overall thoughts:

1. I'm curious if we'll actually be able to get through this with only RegEx instead of an AST. Won't you have to do a lot of this over for resolving default params?
2. Should regexes be bracketed with "^$"? Or that will get in the way of interpolating in strings
3. I'm very curious to see how you're going to surface these CEL errors in a broader context (e.g. cannot set `cpu` to a string expression)

Berlioz:
This branch has changed, uh. A lot. The entire backing type system has changed, specifically. Closing this PR and reopening.

PR #4492, opened 2022-04-27
Serialize loading functions backend for Functions Emulator to avoid race conditions.

taeold:
Container contract based function trigger discovery process spins up a control API server to announce /functions.yaml. When we load multiple functions backend in parallel, they sometimes end up having conflict over port. Serializing the load resolves the issue, but slows down the overall loading process.

Slowdown may be a big hit to extensions emulator, delaying the emulator initialization by few more seconds. Setting up a PR to discuss changes.

### DIFF
--- a/src/emulator/functionsEmulator.ts
+++ b/src/emulator/functionsEmulator.ts
@@ -438,7 +438,6 @@ export class FunctionsEmulator implements EmulatorInstance {
   }
 
   async connect(): Promise<void> {
-    const loadTriggerPromises: Promise<void>[] = [];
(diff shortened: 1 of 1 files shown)


joehan:
I think this is the right thing to do - its better to work correctly than to be broken quickly, and so we should merge this in for now.

There are probably some ways we can improve this, tho, to get back to loading in parallel. We could rework this to pass in a separate port to each loadTriggers, or we could implement better port collision detection logic for the container contract API server.

joehan (approved):
Approved, but let's at least add a note to explore parallelizing this again

PR #4493, opened 2022-04-28
Filter out invalid prefixes and items from listing.

yuchenshi:
### Description

This filters out invalid items and prefixes from Firebase Storage list API results. Similar to production logic at go/firebase-storage-backend-valid-path (Googler only).

### Scenarios Tested

See tests added. Manually tested in Emulator UI too: created a folder, and it shows up in parent folder correctly. Inside the new folder, the view is empty with no placeholder showing up as files.

### DIFF
--- a/scripts/storage-emulator-integration/tests.ts
+++ b/scripts/storage-emulator-integration/tests.ts
@@ -1721,7 +1721,6 @@ describe("Storage emulator", () => {
             items: [],
           });
         });
-
         context("with folder placeholders", () => {
           beforeEach(async function (this) {
             this.timeout(TEST_SETUP_TIMEOUT);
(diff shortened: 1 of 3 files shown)


src/emulator/storage/apis/firebase.ts
```diff
@@ -517,3 +518,28 @@ function setObjectHeaders(
     res.setHeader("Content-Language", metadata.contentLanguage);
   }
 }
+
+function validPrefix(prefix: string) {
```
tonyjhuang, line 522: nit: isValidPrefix

tonyjhuang, line 522: nit: return type?

scripts/storage-emulator-integration/tests.ts
```diff
@@ -1710,6 +1721,63 @@ describe("Storage emulator", () => {
             items: [],
           });
         });
+
+        context("with folder placeholders", () => {
```
tonyjhuang, line 1725: can you include a test that validates that invalid paths are filtered out of the list response please, e.g. `//foo/` `bar//` `/baz//qux/`

src/emulator/storage/apis/firebase.ts
yuchenshi: This is intentionally matching naming in storage backend (for easier cross-reference). I'm open to review a CL that renames it internally for consistency though.

tonyjhuang: I don't think the two need to move in lockstep, we can clean up the internal version too but let's try to follow the style guide here

scripts/storage-emulator-integration/tests.ts
yuchenshi: Done.

src/emulator/storage/apis/firebase.ts
yuchenshi: Done.

tonyjhuang (approved):

src/emulator/storage/apis/firebase.ts
yuchenshi: Sure, renamed

PR #4498, opened 2022-04-28
Adds preview flag guarded code path to start using Builds in CF3 deploys

Berlioz:
Is there any schema or something I need to change somewhere to make the `functionsparam`preview a valid open-sesame target?


inlined (approved):
I still worry about implementation skew here, and I'd want you to be able to release incremental features before being totally finished, but it's your feature.

PR #4501, opened 2022-04-29
Add events config logic to ext:configure flow

huangjeff5:
### Description

In CLI we have an explicit reconfigure flow. We need to make this work with events (i.e. end user can change allowed event types + location)

Screen Shot 2022-04-29 at 12 32 18 PM


joehan (approved):
LGTM overall, with one optional improvement.

PR #4503, opened 2022-04-29
Add total events count when displaying allowed events prompt

huangjeff5:
### Description

For large # of events, user doesn't know how many events there are and it can get overwhelming. A total count helps this.

Screen Shot 2022-04-29 at 12 30 38 PM


joehan (commented):
Hmmm, i think the diff got messed up here, could you fix that?

src/extensions/askUserForEventsConfig.ts
```diff
@@ -67,8 +67,8 @@ export async function askForAllowedEventTypes(
       default: preselectedTypes ?? [],
       // @TODO(b/229170748): Link to docs / audit the copy with UX.
       message:
-        "Please select the events that this extension is permitted to emit. " +
-        "You can implement your own handlers that trigger when these events are emitted to customize the extension's behavior. ",
+        `Please select the events [${eventTypes.length} total] that this extension is permitted to emit. ` +
```
joehan, line 70: Consider rewording to "Please select the events [${events.length} types total]"

As is, it sort of sounds like this extension will only emit X events total, instead of X types

PR #4505, opened 2022-04-29
Add events config logic to ext:configure flow

huangjeff5:
### Description

In CLI we have an explicit reconfigure flow. We need to make this work with events (i.e. end user can change allowed event types + location)

Screen Shot 2022-04-29 at 12 32 18 PM

### DIFF
--- a/src/commands/ext-configure.ts
+++ b/src/commands/ext-configure.ts
@@ -131,7 +131,9 @@ export default new Command("ext:configure <extensionInstanceId>")
       return;
     }
     if (!projectId) {
-      throw new FirebaseError(`Project ID must be provided when re-configuring an instance outside of local mode.`);
+      throw new FirebaseError(
+        `Project ID must be provided when re-configuring an instance outside of local mode.`
+      );
     }
     // TODO(b/220900194): Remove everything below and make --local the default behavior.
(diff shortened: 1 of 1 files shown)


src/commands/ext-configure.ts
```diff
@@ -46,6 +47,9 @@ export default new Command("ext:configure <extensionInstanceId>")
   .before(diagnoseAndFixProject)
   .action(async (instanceId: string, options: Options) => {
     const projectId = getProjectId(options);
+    if (!projectId) {
```
joehan, line 50: This isn;t true - you can run `firebase ext:configure --local` without a project id to reconfigure an extension in the manifest

joehan (changes requested):

src/commands/ext-configure.ts
```diff
@@ -115,29 +130,20 @@ export default new Command("ext:configure <extensionInstanceId>")
       manifest.showPreviewWarning();
       return;
     }
-
+    if (!projectId) {
```
joehan, line 133: Hmmm, I think this is actually redundant - we call needProjectId on l144, which will error out if project id is undefined.

joehan (approved):
LGTM!

PR #4512, opened 2022-05-03
Fix update flow such that users can turn off events

huangjeff5:
### Description

While testing I noticed some weirdness where events could not be turned off via CLI. After digging deeper I tried to simplify the logic such that:

- if events are turned off, it always includes allowed_event_types/eventarc_channel the update mask but undefined fields in the body, so those fields get emptied out in Spanner.
- if events are turned on, it always includes allowed_event_types/eventarc_channel 

If the extension doesn't emit extensions in general, the update flow will still pass allowed_event_types/eventarc_channel in the update mask + undefined fields in the body. This isn't ideal from a code cleanliness standpoint but seems like it will require a bigger refactor to explicitly handle this differently.

### Scenarios Tested

Verified by testing on autopush
firebase update firestore-stripe-payments-autopush1  firebase-internal-testing/firestore-stripe-payments --project=pavelj-extensions1

### DIFF
--- a/src/extensions/extensionsApi.ts
+++ b/src/extensions/extensionsApi.ts
@@ -389,11 +389,7 @@ export async function configureInstance(args: {
       },
     },
   };
-  if (!args.canEmitEvents) {
-    reqBody.data.config.allowedEventTypes = undefined;
-    reqBody.data.config.eventarcChannel = undefined;
-    reqBody.updateMask += ",config.allowed_event_types,config.eventarc_channel";
-  } else {
+  if (args.canEmitEvents) {
     if (args.allowedEventTypes === undefined || args.eventarcChannel === undefined) {
       throw new FirebaseError(
         `This instance is configured to emit events, but either allowed event types or eventarc channel is undefined.`
@@ -401,8 +397,8 @@ export async function configureInstance(args: {
     }
     reqBody.data.config.allowedEventTypes = args.allowedEventTypes;
     reqBody.data.config.eventarcChannel = args.eventarcChannel;
-    reqBody.updateMask += ",config.allowed_event_types,config.eventarc_channel";
   }
+  reqBody.updateMask += ",config.allowed_event_types,config.eventarc_channel";
   return patchInstance(reqBody);
 }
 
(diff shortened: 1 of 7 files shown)


src/extensions/extensionsApi.ts
```diff
@@ -388,14 +389,19 @@ export async function configureInstance(args: {
       },
     },
   };
-  if (args.allowedEventTypes !== undefined) {
+  if (!args.canEmitEvents) {
+    reqBody.data.config.allowedEventTypes = undefined;
+    reqBody.data.config.eventarcChannel = undefined;
```
joehan, line 394: Isn't it unnecessary to explicitly set these to undefined here? IIUC correctly, just including the fields in updateMask should be sufficient

src/extensions/extensionsApi.ts
```diff
@@ -430,14 +437,20 @@ export async function updateInstance(args: {
...
   }
-  if (args.allowedEventTypes !== undefined) {
-    body.config.allowedEventTypes = args.allowedEventTypes;
-    updateMask += ",config.allowed_event_types";
-  }
 
-  if (args.eventarcChannel !== undefined) {
+  if (!args.canEmitEvents) {
+    body.config.allowedEventTypes = undefined;
+    body.config.eventarcChannel = undefined;
```
joehan, line 443: Same as above - isn;t just updateMask enough?

src/extensions/extensionsApi.ts
```diff
@@ -430,14 +437,20 @@ export async function updateInstance(args: {
...
+  } else {
+    if (args.allowedEventTypes === undefined || args.eventarcChannel === undefined) {
+      throw new FirebaseError(
+        `This instance is configured to emit events, but either allowed event types or eventarc channel is undefined.`
+      );
+    }
+    body.config.allowedEventTypes = args.allowedEventTypes;
     body.config.eventarcChannel = args.eventarcChannel;
-    updateMask += ",config.eventarc_channel";
+    updateMask += ",config.allowed_event_types,config.eventarc_channel";
```
joehan, line 453: You have this same line in both parts of an if else  statement - you can just move it up a level.

In fact, i think you can simplify this whole logic quite a bit to:
```
if (args.canEmitEvents) {
 if (args.allowedEventTypes === undefined || args.eventarcChannel === undefined) {
      throw new FirebaseError(
        `This instance is configured to emit events, but either allowed event types or eventarc channel is undefined.`
      );
    }
    body.config.allowedEventTypes = args.allowedEventTypes;
    body.config.eventarcChannel = args.eventarcChannel;
}
updateMask += ",config.allowed_event_types,config.eventarc_channel"; // or just include this when you define update mask.
```

joehan (changes requested):

src/extensions/extensionsApi.ts
huangjeff5: Done

src/extensions/extensionsApi.ts
huangjeff5: Done

joehan (approved):

PR #4514, opened 2022-05-03
Inlined.cpu.3

inlined:
1. Added validation to CPU config
2. When tests weren't failing, I realized you need to say `throw()` not `throw` in Chai Expect, which meant none of our throwing tests were actually testing anything. Used find + replace to turn all `throw;` into `throw();` and fixed broken tests.


taeold (approved):

taeold:
🤯 on `.throw` vs `.throw()`

PR #4523, opened 2022-05-05
Misc bugfixes to trigger -> build -> backend chain

Berlioz:
Addresses some problems I noticed while testing locally:
1) To get the default serviceAccountEmail, you set the field to null in backend.backend, not "default"
2) I forgot about SecretEnvVariables entirely
3) None of this makes any sense in the context of CF3v2 (the code in runtimes/discovery should take care of that, right?)

### DIFF
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -114,7 +114,7 @@ export async function prepare(
     const userEnvs = functionsEnv.loadUserEnvs(userEnvOpt);
     const envs = { ...userEnvs, ...firebaseEnvs };
     let wantBackend: backend.Backend;
-    if (previews.functionsparams) {
+    if (previews.functionsparams && !previews.functionsv2) {
       const wantBuild = await runtimeDelegate.discoverBuild(runtimeConfig, firebaseEnvs);
       wantBackend = build.resolveBackend(wantBuild, userEnvs);
     } else {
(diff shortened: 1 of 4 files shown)


src/deploy/functions/prepare.ts
```diff
@@ -114,7 +114,7 @@ export async function prepare(
     const userEnvs = functionsEnv.loadUserEnvs(userEnvOpt);
     const envs = { ...userEnvs, ...firebaseEnvs };
     let wantBackend: backend.Backend;
-    if (previews.functionsparams) {
+    if (previews.functionsparams && !previews.functionsv2) {
```
inlined, line 117: previews.functionsv2 is gone now.

src/deploy/functions/runtimes/node/parseTriggers.ts
```diff
@@ -302,7 +302,7 @@ export function addResourcesToBuild(
     project: projectId,
     entryPoint: annotation.entryPoint,
     runtime: runtime,
-    serviceAccount: annotation.serviceAccountEmail || "default",
+    serviceAccount: annotation.serviceAccountEmail || null,
```
inlined, line 305: It's quite possible that `undefined` will trigger different behavior than `null`. They behave differently in `JSON.stringify` and `proto.fieldMasks` includes nulls but not undefined too. This likely has the side effect of resetting service accounts on every deploy to their default.

inlined (approved):
Sorry I got so caught up in I/O stuff that I missed this. If you ever wait more than a full day for a review, please ping me.

inlined:
This has been sitting around for about a month; should this be committed or closed?

Berlioz:
We're closing this one; the relevant changes in here have been folded into other PRs.

PR #4526, opened 2022-05-06
Support relative paths when emulating local extensions

joehan:
### Description
Support relative paths when emulating local extensions. Turns out we needed to resolve this to an absolute path so that the functions emulator can require it from a separate process consistently.

### Scenarios Tested
I was able to emulate an extension located at "../"

### DIFF
--- a/src/emulator/functionsEmulator.ts
+++ b/src/emulator/functionsEmulator.ts
@@ -1012,6 +1012,7 @@ export class FunctionsEmulator implements EmulatorInstance {
   }
 
   getNodeBinary(backend: EmulatableBackend): string {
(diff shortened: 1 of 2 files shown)


apascal07 (approved):

src/emulator/functionsEmulator.ts
```diff
@@ -1012,6 +1012,7 @@ export class FunctionsEmulator implements EmulatorInstance {
   }
 
   getNodeBinary(backend: EmulatableBackend): string {
+    this.logger.log("DEBUG", `We're at ${__dirname}`);
```
huangjeff5, line 1015: Remove?

huangjeff5 (approved):
LGTM with one small nit + fixing lint issues.

PR #4527, opened 2022-05-06
Improve error message when parsing function source fails.

taeold:
Previously, an invalid function source would show error message like this:

```bash

i  deploying functions
...
Error: Failed to parse backend specification:
- YAMLException incomplete explicit mapping pair; a key node is missed; or followed by a non-tabulated empty line at line 1, column 65:
     ...  function source: ReferenceError: aa is not defined
```

This is surprising - a YAMLException? A backend specification? Often times, the full error message would curt short, and debugging the issue required the user to carefully inspect the debug log.

We improve the error message to the following:

```bash

i  deploying functions
...
Error: Failed to load function definition from source: Failed to generate manifest from function source: ReferenceError: aa is not defined
```

We hid the YAML portion of the error and make sure to relay the full error message returned from the underlying Functions Control API (i.e. the serve responsible for loading and advertising the `functions.yaml` baked into the Functions SDK)

### DIFF
--- a/src/deploy/functions/runtimes/discovery/index.ts
+++ b/src/deploy/functions/runtimes/discovery/index.ts
@@ -34,6 +34,9 @@ export function yamlToBackend(
   }
 }
 
+/**
+ *
+ */
 export async function detectFromYaml(
   directory: string,
   project: string,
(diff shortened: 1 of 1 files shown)


jhuleatt (approved):
🥇 Thanks @taeold! This will make error messages way more readable

PR #4530, opened 2022-05-06
Inlined.cpu fixes

inlined:
Bugbash fixes for myself & others.

1. Adds validation to CPU + memory combinations so errors are reported earlier
2. Forces the default memory in GCFv2 to be 256MiB instead of 256MB because we make everything MiB. Allowing MB crashes pantheon on update.


colerogers (approved):
@inlined Is the validation logic in a separate PR?

inlined:
Sorry, they got lost in a rebase. Fixed.

PR #4535, opened 2022-05-09
Fixing broken tests on next

joehan:
### Description
While sync next and master, I noticed we introduced a bug in ext:install. Then, I accidentally merged a bug fix directly to https://github.com/firebase/firebase-tools/commits/next 🤦 , which broke the tests. This PR fixes those tests (and finishes the bug fix)

### Scenarios Tested
The tests pass.
firebase ext:install works correctly when version is 'latest' or unspecified.


bkendall (approved):
Thanks for the fix!

PR #4543, opened 2022-05-11
Correctly copy over secret environment variables

inlined:
Fixes #4540

PR #4544, opened 2022-05-12
Add typing to proto converters and fix breaks

inlined:
After fixing #4543 I decided that I now know how to make `proto.renameIfPresent` more strongly typed and made a change that would have caught the error before it was ever a bug.

Did that and started fixing the build breaks it caught. Since I noticed a lot of converter code for `build` was using the same key name, I also added `proto.convertIfPresent`.

The build code was broken for scheduled functions. That's because the old `triggerParser` uses durations whereas the container contract uses seconds (so we can use number params). The code looks extra wonky because `backend.Backend` _also_ uses durations. As an exercise for the reader, someone should probably make `backend.Backend` use seconds and convert to/from duration in `endpointFromFunction` and `functionFromEndpoint`. duration is silly and should be in as little of our type system as possible.


src/deploy/functions/build.ts
```diff
@@ -364,77 +364,104 @@ function discoverTrigger(endpoint: Endpoint): backend.Triggered {
...
+        resolveInt
+      );
+      proto.renameIfPresent(
+        bkSchedule.retryConfig,
+        endpoint.scheduleTrigger.retryConfig,
+        "minBackoffDuration",
+        "minBackoffSeconds",
+        (seconds) => proto.durationFromSeconds(resolveInt(seconds))
+      );
+      proto.renameIfPresent(
```
colerogers, line 394: Is this identical to L374?

src/gcp/cloudfunctionsv2.ts
```diff
@@ -230,7 +230,12 @@ export function mebibytes(memory: string): number {
     const suffix = matches[3] || "";
     bytes = quantity * BYTES_PER_UNIT[suffix as MemoryUnit];
   }
-  return bytes / (1 << 20);
+  bytes = bytes / (1 << 20);
+  const typed = bytes as backend.MemoryOptions;
+  if (backend.AllMemoryOptions.includes(typed)) {
```
colerogers, line 235: should there be a `!` here?

colerogers (commented):
Couple Q's

src/deploy/functions/build.ts
inlined: yup

src/gcp/cloudfunctionsv2.ts
inlined: yup

PR #4549, opened 2022-05-17
fix error messaging in apps:sdkconfig

### LINKED ISSUE #4007: firebase apps:sdkconfig non-interactive output misleads, says "has multiple apps" when there are 0 apps

**firebase-tools:**  9.22.0
**Platform:** Mac

- Create new firebase project, e.g. named "a-project-with-no-apps"
- `firebase use a-project-with-no-apps`
- `firebase apps:sdkconfig web --json`

The JSON output indicates "Project a-project-with-no-apps has no apps. Create an app first." or similar.

Output for this example project "a-project-with-no-apps" is:

```
{
  "status": "error",
  "error": "Project a-project-with-no-apps has multiple apps, must specify an app id."
}
```

which is misleading - the command can't produce output in this non-interactive mode because there are 0 apps, not multiple apps.

bkendall:
### Description

The error messaging when trying to get apps:sdkconfig was confusing and inconsistent when in interactive vs. non-interactive mode. Unify the error messaging to make a bit more sense and to be consistent when there are no apps.

Fixes #4007 

### Scenarios Tested

image


src/commands/apps-sdkconfig.ts
```diff
@@ -55,78 +60,78 @@ module.exports = new Command("apps:sdkconfig [platform] [appId]")
...
-        throw err;
+      const apps = await listFirebaseApps(projectId, appPlatform);
+      // Fail out early if there's no apps.
+      checkForApps(apps, appPlatform);
+      // if there's only one app, we don't need to prompt interactively
+      if (apps.length === 1) {
+        // If there's only one, use it.
+        appId = apps[0].appId;
+        appPlatform = apps[0].platform;
+      } else if (options.nonInteractive) {
```
joehan, line 83: How does this behave when there are 0 apps? Do we show the correct error?

joehan (approved):
LGTM as long as this behaves correctly when there are 0 apps

src/commands/apps-sdkconfig.ts
bkendall: Above, on L77, it checks for apps and will throw an error if there are no apps.

joehan: Derp, just saw that you had a screenshot of this case. LGTM

PR #4556, opened 2022-05-17
Fixes for Node 18

bkendall:
### Description

This has been an ✨  adventure ✨ .

- updated the node types to v14, which is the minimum we now support.
- updated `@google-cloud/pubsub` to fix some typing issues
- fixed a bunch of other typing issues in our codebase - nothing too significant

Then, it turns out that Node 18 changed how `localhost` is resolved. In short, node uses the system to resolve the dns address `localhost`, but does not sort the returned results any more - which used to put the v4 address first. Now, if v6 is available, it will use that address instead.

What this means is that we have some services that were starting on `localhost`, but binding only to the ipv6 loopback address (`::1`). Others were being forced to start on `127.0.0.1`. When we would start looking to see if a port was being used (to see if an emulator was alive), we let the DNS tell us where to look, and sometimes that was wrong.

In v14.18.0, `dns.setDefaultResultOrder` was added to give control over the ordering. We're setting it in the CLI now to always give us ipv4 first, keeping the existing behavior through node 18. We *also* have to make sure we pass the flag in for the UI with this same value, so that the UI knows how to connect to the emulators.

This needs to be tested on a couple more platforms I think, but it should keep the status quo.

What we *really* should do is listen on `::`, so that `127.0.0.1` and `::1` (both options for `localhost`) work, but I think that will require more thought (especially since WSL *doesn't* work with that, and then we're having to support this instead - though maybe WSL v2 does work with ipv6 and it's moot now... will have to check).

### DIFF
--- a/src/emulator/portUtils.ts
+++ b/src/emulator/portUtils.ts
@@ -135,12 +135,15 @@ export async function checkPortOpen(port: number, host: string): Promise<boolean
 }
 
 /**
- * Wait for a port to close on the given host. Checks every 250ms for up to 30s.
+ * Wait for a port to close on the given host. Checks every 250ms for up to 60s.
  */
 export async function waitForPortClosed(port: number, host: string): Promise<void> {
   const interval = 250;
-  const timeout = 60000;
+  const timeout = 60_000;
   try {
+    if (host === "localhost") {
+      host = "127.0.0.1";
+    }
     await tcpport.waitUntilUsedOnHost(port, host, interval, timeout);
   } catch (e: any) {
     throw new FirebaseError(`TIMEOUT: Port ${port} on ${host} was not active within ${timeout}ms`);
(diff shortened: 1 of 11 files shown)


yuchenshi (approved):

src/emulator/portUtils.ts
```diff
@@ -1,9 +1,12 @@
 import * as pf from "portfinder";
 import * as tcpport from "tcp-port-used";
+import * as dns from "dns";
 
 import { FirebaseError } from "../error";
 import { logger } from "../logger";
 
+dns.setDefaultResultOrder("ipv4first");
```
yuchenshi, line 8: Okay I don't love this idea either, but can we feature detect this?
```suggestion
if (dns.setDefaultResultOrder) {
  dns.setDefaultResultOrder("ipv4first");
}
```

bkendall: We _can_, but I've also updated the engines here to `v14.18`, where this function was added. Do you still think it's worth having the check?

For what it's worth, I've discovered that a dependency does `require('node:process')`, which support for was only added in node 14.18 anyway, so... if it breaks, it'd break in glorious fashion.

bkendall:
I can confirm this worked on Windows 11 in WSL and Command Prompt. Yay!

PR #4558, opened 2022-05-18
Run integration tests on windows-latest as well

joehan:
### Description
Lots of users are on Windows - we should be testing on Windows too.

This started small and quickly ballooned - it turns out there were a LOT of things breaking our tests on Windows. I managed to get 3.5/6 of our integration tests working, but I've sunk a lot of time into this already and need to take a break to focus on other work before getting back to it. 

A quick summary of the changes:
- Adds a 'integration-windows' test block to node-test.yml
- Switch intergration-helpers to use cross-spawn instead of spawn
- Minor changes to helper scripts in order to work in batch
- Separate out import/export tests from triggers-end-to-end-test. These tests were unrelated to triggers, so they probably should have been separate in the first place. They were also breaking on Windows.
- Refactor triggers-end-to-end-tests to not restart the emulator suite between every trigger type. This was unnecessary + slowed the test down, and also broke on Windows.

Known issues:
- On Windows, tests the repeatedly restart the emulator suite fail due to ports being taken already. by previous runs of the emulator I'm not sure exactly why, but it seems like emulator hub is not shutting down fully & releasing the port before the next test starts.
- The Extension emulator test seems to be firing before `npm i` finishes - it fails with a 'module not found' error.


scripts/emulator-import-export-tests/tests.ts
```diff
@@ -0,0 +1,559 @@
+import { expect } from "chai";
```
joehan, line 1: FYI, the code in this file used to be part of triggers-end-to-end-test, and was moved over basically unchanged.

scripts/triggers-end-to-end-tests/tests.ts
```diff
@@ -42,16 +38,7 @@ function readConfig(): FrameworkOptions {
...
-  return (data: unknown) => {
-    if (typeof data !== "string" && !Buffer.isBuffer(data)) {
-      throw new Error(`data is not a string or buffer (${typeof data})`);
-    }
-    return data.includes(msg);
-  };
-}
-
-describe("database and firestore emulator function triggers", () => {
+describe("function triggers", () => {
```
joehan, line 41: These changes look very large, but they are really just reorganizing these tests to share a single beforeAll hook & to run as part of the same emulator run.

bkendall (approved):

bkendall:
I don't *love* this, but I think having tests on Windows is better than not. Let's give it a whirl.

PR #4559, opened 2022-05-18
add firepit-builder build script

bkendall:
### Description

This script is missing to make building this container easier.


joehan:
Do we document anywhere when this should be used? Maybe add a note to go/publish-via-cloud-build?

PR #4564, opened 2022-05-19
CF3v2 support for Build type

Berlioz:
This implements an equivalent to the parseEndpoints() function that converts a CF3v2 YAML manifest into a Backend, which instead spits out a Build. resolve() can then be called on the Build to get a Backend. Currently this is intended to be exactly equivalent to the directly-to-Backend code path, but parameterization features will be hung off the resolve() function on Build in the future.

Currently, this code path will not be enabled for production users outside of preview features.


src/deploy/functions/build.ts
```diff
@@ -309,7 +320,7 @@ export function resolveBackend(build: Build, userEnvs: Record<string, string>):
         "environmentVariables",
         "labels"
       );
-      // proto.copyIfPresent(bkEndpoint, endpoint, "secretEnvironmentVariables");
+      proto.copyIfPresent(bkEndpoint, endpoint, "secretEnvironmentVariables");
```
colerogers, line 323: nit (feel free to disregard) - what are your thoughts on naming these two endpoints `buildEp` and `backendEp`? I was a bit confused to what this generic `endpoint` was at first.

Berlioz: Honestly, I think it's a great idea.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -40,6 +41,26 @@ export interface Manifest {
...
+export function buildFromV1Alpha1(
+  yaml: unknown,
+  project: string,
+  region: string,
+  runtime: runtimes.Runtime
+): build.Build {
+  const manifest = JSON.parse(JSON.stringify(yaml)) as Manifest;
+  const bd: build.Build = build.empty();
+  bd.requiredAPIs = parseRequiredAPIs(manifest);
+  bd.endpoints = {};
```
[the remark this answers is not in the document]
Berlioz: Sure does.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -80,17 +101,8 @@ function parseRequiredAPIs(manifest: Manifest): backend.RequiredAPI[] {
...
 
-function parseEndpoints(
-  manifest: Manifest,
-  id: string,
-  project: string,
-  defaultRegion: string,
-  runtime: runtimes.Runtime
-): backend.Endpoint[] {
-  const allParsed: backend.Endpoint[] = [];
+function assertManifestEndpoint(ep: ManifestEndpoint, id: string) {
```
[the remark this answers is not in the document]
Berlioz: Done.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -146,19 +158,206 @@ function parseEndpoints(
...
+      timeZone: ep.scheduleTrigger.timeZone || "",
+      retryConfig: {},
+    };
+    if (ep.scheduleTrigger.retryConfig) {
+      st.retryConfig = {
+        retryCount: ep.scheduleTrigger.retryConfig.retryCount,
+        maxDoublings: ep.scheduleTrigger.retryConfig.maxDoublings,
+      };
+      if (ep.scheduleTrigger.retryConfig.maxRetryDuration) {
+        st.retryConfig.maxRetrySeconds = secondsFromDuration(
```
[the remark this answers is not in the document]
Berlioz: Later on; the build gets resolved into a backend, prompting for user input for anything not living in the .env and cloud secret manager. I think we'll have to jigger this code later, to allow Expression<number> types in addition to number types, but that might involve code in the actual SDK too.

src/deploy/functions/build.ts
```diff
@@ -366,13 +378,29 @@ function discoverTrigger(endpoint: Endpoint): backend.Triggered {
       schedule: resolveString(endpoint.scheduleTrigger.schedule),
       timeZone: resolveString(endpoint.scheduleTrigger.timeZone),
     };
-    proto.renameIfPresent(
-      bkSchedule,
-      endpoint.scheduleTrigger,
-      "retryConfig",
-      "retryConfig",
-      resolveInt
+    const bkRetry: backend.ScheduleRetryConfig = {};
```
taeold, line 381: vsfan/danielylee: Typings didn't catch this bug - why?

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -40,6 +41,25 @@ export interface Manifest {
...
 
+/** Returns a Build from a v1alpha1 Manifest. */
+export function buildFromV1Alpha1(
+  yaml: unknown,
+  project: string,
+  region: string,
+  runtime: runtimes.Runtime
+): build.Build {
+  const manifest = JSON.parse(JSON.stringify(yaml)) as Manifest;
+  const bd: build.Build = build.empty();
```
taeold, line 52: TODO: consider adding the following check as done in `backendFromV1Alpha1`.

```
  assertKeyTypes("", manifest, {
    specVersion: "string",
    requiredAPIs: "array",
    endpoints: "object",
  });
```

taeold (approved):

src/deploy/functions/build.ts
Berlioz: resolveInt can accept a string as an argument and return an integer, because it has to be able to handle CEL expressions. 

therefore, passing it to renameIfPresent was coincidentally able to hide this bug--the retryconfig parameters being passed in as string durations instead of as numbers--from the type system.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
Berlioz: Done.

PR #4567, opened 2022-05-19
Update e2e extensions emulator test to include events

huangjeff5:
### Description

Update e2e test such that:
- We are using a version that emits events
- Custom handler fires and writes to firestore
- Test verifies that firestore result matches expectation

### DIFF
--- a/scripts/extensions-emulator-tests/run.sh
+++ b/scripts/extensions-emulator-tests/run.sh
@@ -8,8 +8,6 @@ source scripts/set-default-credentials.sh
   npm install
 )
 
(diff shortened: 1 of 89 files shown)


scripts/extensions-emulator-tests/run.sh
```diff
@@ -9,3 +9,5 @@ source scripts/set-default-credentials.sh
 )
 
 mocha scripts/extensions-emulator-tests/tests.ts
+
```
joehan, line 12: remove these extra newlines - should just be one at the end of the file

joehan (approved):

PR #4568, opened 2022-05-19
rewrite rtdb in typescript, using apiv2

### LINKED ISSUE #4083: Convert to TS: src/rtdb.js
The following file `src/rtdb.js` should be converted to Typescript.

bkendall:
### Description

Rewrite rtdb.js to typescript, using apiv2.

Fixes #4083

### Scenarios Tested

Tested deploying rules, both valid an invalid.


src/rtdb.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export async function updateRules(
+  projectId: string,
+  instance: string,
+  src: any,
+  options: { dryRun?: boolean } = {}
+): Promise<void> {
+  let path = ".settings/rules.json";
+  if (options.dryRun) {
+    path += "?dryRun=true";
```
joehan, line 19: Since we're moving to apiv2, consider switching this to use ClientRequestOptions.queryParams instead of just appending them to path

joehan (approved):
LGTM with one nit

src/rtdb.ts
bkendall: Easy enough. Done.

PR #4569, opened 2022-05-19
migrate management API calls to apiv2

bkendall:
### Description

Migrates all the API calls in the management folder to apiv2. Done by using `nock` in the tests (which validates the calls being made) and then replacing the actual calls, so I know the requests are the same.

### Scenarios Tested

- [x] Test making/removing/listing apps
- [x] Test working with a database


src/test/management/apps.spec.ts
```diff
@@ -61,21 +64,22 @@ function generateWebAppList(counts: number): WebAppMetadata[] {
   }));
 }
 
-describe("App management", () => {
+describe.only("App management", () => {
```
joehan, line 67: Forgot an only

joehan (changes requested):
Left a .only in here

joehan (commented):
Other than the .only, this looks good to me (once the 2 cases you listed are tested

src/appdistribution/client.ts
```diff
@@ -196,7 +196,7 @@ export class AppDistributionClient {
         { emails: string[] },
         BatchRemoveTestersResponse
       >({
-        method: "POST",
+        method: "DELETE",
```
joehan, line 199: Was this just broken before?

joehan (approved):

src/appdistribution/client.ts
bkendall: I changed the wrong place.

PR #4582, opened 2022-05-23
Add RTDB Triggers into function deploy

colerogers:
WIP

### DIFF
--- a/src/deploy/functions/services/database.ts
+++ b/src/deploy/functions/services/database.ts
@@ -1,0 +1,18 @@
+import * as backend from "../backend";
+import { FirebaseError } from "../../../error";
+
+/**
+ * Sets a database event trigger's region to the function region.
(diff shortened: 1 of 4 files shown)


src/deploy/functions/services/database.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Sets a database event trigger's region to the function region.
+ * @param endpoint the database endpoint
+ */
+export function ensureDatabaseTriggerRegion(
+  endpoint: backend.Endpoint & backend.EventTriggered
+): Promise<void> {
+  if (!endpoint.eventTrigger.region) {
+    endpoint.eventTrigger.region = endpoint.region;
```
inlined, line 12: Don't we need to default the region to the CLI's default region as well?

inlined (approved):

PR #4584, opened 2022-05-24
Fixes issue where --local flag would cause extensions commands to error out

### LINKED ISSUE #4577: `error: unknown option '--local'` when trying to install an extension using CLI
**firebase-tools:** 11.0.0
**Platform:**: Ubuntu

### [REQUIRED] Test case
```sh
$ firebase ext:install typesense/firestore-typesense-search --local --project=xxx
error: unknown option '--local'
```

Also checked with `--help`. But not giving enough info:
```
$ firebase ext:install --help
Usage: firebase ext:install [options] [extensionName]

install an official extension if [extensionName] or [extensionName@version] is provided; or run with `-i` to see all available extensions.

Options:
  -f, --force  automatically accept all interactive prompts
  -h, --help   output usage information
```

I was trying to install an extension and try it in emulator. For that I followed as per [firebase extension page](https://firebase.google.com/products/extensions/typesense-firestore-typesense-search#:~:text=How%20to%20install%20this%20extension).

Screenshot from 2022-05-20 11-31-44

I don't know. I was trying. I hope this should install the extension locally so that I can test with emulator.

```
error: unknown option '--local'
```
Screenshot from 2022-05-20 11-19-59

Run the command with --debug flag, and include the logs below: same result

Additionally, all the blue links in steps 3, 4, 5 are broken.
Screenshot from 2022-05-20 11-32-57

joehan:
### Description
`ext:*` commands should not error out when run with --local, even though it was deprecated (fixes #4577)

### Scenarios Tested
Now with a warning message
Screen Shot 2022-05-24 at 1 29 57 PM

### DIFF
--- a/src/commands/ext-configure.ts
+++ b/src/commands/ext-configure.ts
@@ -34,6 +34,7 @@ marked.setOptions({
 export default new Command("ext:configure <extensionInstanceId>")
   .description("configure an existing extension instance")
   .withForce()
+  .option("--local", "deprecated")
   .before(requirePermissions, [
     "firebaseextensions.instances.update",
     "firebaseextensions.instances.get",
--- a/src/commands/ext-configure.ts
+++ b/src/commands/ext-configure.ts
@@ -50,6 +50,9 @@ export default new Command("ext:configure <extensionInstanceId>")
           `See https://firebase.google.com/docs/extensions/manifest for more details.`
       );
     }
+    if (options.local) {
+      utils.logLabeledWarning(logPrefix, "As of firebase-tools@11.0.0, the `--local` flag is no longer required, as it is the default behavior.")
+    }
 
     const config = manifest.loadConfig(options);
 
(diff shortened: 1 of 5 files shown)


src/commands/ext-configure.ts
```diff
@@ -34,6 +34,7 @@ marked.setOptions({
 export default new Command("ext:configure <extensionInstanceId>")
   .description("configure an existing extension instance")
   .withForce()
+  .option("--local", "deprecated")
```
bkendall, line 37: Strongly, strongly consider printing a warning if `--local` is included. IIRC, you should be able to check `options.local`.

bkendall (changes requested):

src/commands/ext-configure.ts
joehan: Good point - done

bkendall (approved):

PR #4585, opened 2022-05-25
Migrate API call in eventarcTrigger to stop using requests library

huangjeff5:


apascal07 (approved):

bkendall (approved):
I appreciate you.

huangjeff5:
> I appreciate you.

:)

That being said, I had to change our server response to return JSON (previously it accepted strings). Ideally we shouldn't have to do that and the client can just handle the response. Do you happen to know if there's a setting that enables v2 client to handle non-JSON responses?

bkendall:
Yup - you want to use `Client.request` and set all the options yourself. You can then do what you want with the body once you have it (string, xml, binary, gibberish). But since _most_ APIs are JSON based 'round these parts, that's the default we went with.

bkendall:
Link example: https://github.com/firebase/firebase-tools/blob/master/src/test/apiv2.spec.ts#L127-L132

PR #4590, opened 2022-05-26
Fix env file optional params

yamankatby:
### Description

Fallback to the default value if the param not required and not provided

fixes https://github.com/firebase/extensions/issues/924


src/extensions/extensionsHelper.ts
```diff
@@ -177,15 +177,15 @@ export function populateDefaultParams(
...
-    if (!paramVars[param.param]) {
-      if (param.default !== undefined && param.required) {
-        newParams[param.param] = param.default;
-      } else if (param.required) {
-        throw new FirebaseError(
-          `${param.param} has not been set in the given params file` +
-            " and there is no default available. Please set this variable before installing again."
-        );
-      }
+    if (paramVars[param.param]) continue;
```
dackers86, line 180: Changes look good. The first line could be an unnecessary change so can probably leave as it was.

dackers86 (changes requested):

yamankatby:
Hey @bkendall,

done, update test to cover the new case

jeffdh5 (approved):

joehan (changes requested):
Thanks for contributing! However, I think this fix is a bit off. The check you removed is correct and should be there - optional params should be null if they are omitted. As I mentioned in the linked issue, I think the extension shoul have those params as required. However, this also highlighted some weirdness in a different check we have: https://github.com/firebase/firebase-tools/blob/master/src/extensions/askUserForParam.ts#L61 is trying to check 'undefined' against the list of options. I'll make a PR to clean that up.

yamankatby:
Hi @joehan, Thank you for your review!
I made this change to make the behavior of the CLI installation match the Firebase UI installation, In the Firebase UI, you are forced to choose the default value even if the param is optional.

https://user-images.githubusercontent.com/35961879/173414696-a9590145-7fd0-4d22-98e7-916ca0baa871.mov

joehan:
Hey @yamankatby - the behavior you are describing (being forced to pick a value) is exactly what 'required' is meant to do. 

https://github.com/firebase/firebase-tools/pull/4646 should clean up the errors you were running into where you could not omit optional SELECT and MULTISELECT params. I also suggested in the original thread that we should make these params required.

PR #4592, opened 2022-05-27
Emulator - Fix Auth Blocking Trigger registration

colerogers:
Fixes Auth Blocking Trigger registration in the Emulator.


inlined (approved):

src/emulator/functionsEmulator.ts
```diff
@@ -859,31 +852,32 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+        beforeSignIn: {
+          functionUri: url,
+        },
+      };
     }
 
+    this.blockingFunctionsConfig.forwardInboundCredentials = {
+      accessToken: blockingTrigger.options!.accessToken as boolean,
+      idToken: blockingTrigger.options!.idToken as boolean,
+      refreshToken: blockingTrigger.options!.refreshToken as boolean,
```
lisajian, line 878: Ah my bad, I didn't catch this in https://github.com/firebase/firebase-tools/pull/4442 - would it make more sense to just change the `BlockingTrigger` interface to the following?

```
export interface BlockingTrigger {
  eventType: string;
  options?: Record<string, boolean>; // previously Record<string, unknown>
}
```

From what I can tell, the options corresponds to `forwardInboundCredentials`, so you wouldn't need to do any casting here if you changed that interface definition? Although, if there's some other funkiness happening (e.g. if `blockingTrigger.options!.accessToken` could actually be a string "true"/"false" or a number 0/1), you could do the double exclamation mark trick to convert it to an actual boolean (e.g. `!!blockingTrigger.options!.accessToken`)

colerogers: So we have it set to `unknown` for additional blocking trigger options in the future (if that ever happens). Since these get sourced from our SDK, we want to use `unknown` over `any` for a little more type safety.

Thanks for the shorthand cast notation suggestion, updated it!!

lisajian (approved):

PR #4600, opened 2022-05-31
fix: Change photoUrl to photoURL in cloud functions auth provider payload

macksal:
### Description

The Firebase SDKs expect the Photo URL property on a user payload to be named `photoURL` but it is returned from the emulator as `photoUrl`. The property name is changed correctly for the top-level user info, but not for the array of third-party provider info. See [#2887](https://github.com/firebase/firebase-tools/issues/2887) for more context.

### Scenarios Tested

I haven't manually tested this as it is a clear typo from copying over the types from the REST API. All tests are still passing.


macksal:
I see based on [the type definitions in firebase-admin-node](https://github.com/firebase/firebase-admin-node/blob/a98a3cc275f9fee0474ff1905d2217a600d5f57e/src/auth/index.ts#L76) that uid should perhaps also be returned here? If someone can sanity check this, it would also be a trivial change to add here.

haythamhakim:
any updates on this? The inconsistency is causing issues.

yuchenshi (approved):

PR #4608, opened 2022-06-01
Crash emulator when an invalid function is emulated

huangjeff5:
### Description

Currently, if we try to emulate an invalid function and we detect things like formatting issues, we will catch the issue without processing the function. This isn't ideal... it's easy for those issues to get lost in the logs and for users to spend unnecessary time debugging.

We ran into this a couple times with the extension events emulator work; we weren't sure why our custom event handler wasn't firing (it didn't follow the strict cf3v2 naming conventions) and realized that it wasn't being registered in the first place. Now we'll throw an explicit error and force the user to handle it before starting emulators.

Screen Shot 2022-06-02 at 1 05 00 PM

### DIFF
--- a/src/emulator/functionsEmulator.ts
+++ b/src/emulator/functionsEmulator.ts
@@ -576,7 +576,7 @@ export class FunctionsEmulator implements EmulatorInstance {
         // To match prod behavior, only validate functionName
         functionIdsAreValid([{ ...definition, id: definition.name }]);
       } catch (e: any) {
-        throw new FirebaseError(`functions[${definition.id}]: Invalid function id: ${e.message}`)
+        throw new FirebaseError(`functions[${definition.id}]: Invalid function id: ${e.message}`);
       }
 
       let added = false;


yuchenshi:
I'm okay with the change, but how big is the estimated fraction of developers potentially affected? Will this break save + hot reload dev flows? What if my IDE (e.g. IntelliJ) saves my file automatically all the time without explicit action?

joehan (approved):
I _think_ this is more desirable behavior, since its better to error clearly than to do so silently. I also think the hot reload concerns are pretty minor - I imagine that not many changes will actually invalidate a functionID during development. Additionally, folks using an IDE that autosaves are likely going to run into issues today that break hot reloading - it seems pretty likely that you'd autosave at point where the code is invalid for reasons other than bad function Ids anyway.

Regardless, this is probably worth running past Daniel or someone on the CF3 side, since I suspect it will mostly affect the UX for cf3v2 functions

huangjeff5:
> I'm okay with the change, but how big is the estimated fraction of developers potentially affected? Will this break save + hot reload dev flows? What if my IDE (e.g. IntelliJ) saves my file automatically all the time without explicit action?

Technically it affects all functions emulator users, but practically it will affect cf3v2 users more because they have stricter guidelines on function naming.

In the case of hot reloading, I'm not sure I understand the question fully. When I think of React hot reload, they handle uncompilable code by showing you where the compile error is, but continues to run. Is that how our hot-reload works - where errors get highlighted but the hot-reloader still runs and listens to changes? That's the user experience I'd like to get to.

huangjeff5:
Ok so to confirm it does indeed break hot reload. Attached a screenshot to show the UX. I still think this is better than silently failing but the best scenario is:

- code throws error
- hot reloader catches the error, highlights it in terminal UI, and continues hot reloading

But I think that will be a bigger project outside of this PR. I still think what we have here is better than the status quo so I would go with merging if people are okay.

huangjeff5:
Another point I realized: Hot reloading already crashes if the code has issues, but not if the function ID has issues.

So crashing at the trigger validation logic is more consistent with our UX than the status quo. And in the future, we can figure out a path forward for both that's consistent - but for now it makes more sense to keep it consistent.

taeold (approved):

huangjeff5:
Quick recap from discussion with Daniel on Friday:
- Even though breaking hot reloading is not ideal, this is something new we introduced while supporting CF3v2 emulation.
- This PR will get us to a consistent state, where both invalid IDs and non-compiling JS code will cause hot reloading to fail. Even if this is not the ideal state we want to be in, a consistent UX is better than an inconsistent UX.
- Daniel will follow up in the future to revamp our hot reload feature so that it doesn't crash and possibly stops execution at the error so that the user knows what happened.

yuchenshi (approved):

PR #4609, opened 2022-06-02
feat: CRC32C Base64 Fix

danielbankhead:
### Description

Dan from GCS DPE 👋🏽

This issue landed in our repo: https://github.com/googleapis/nodejs-storage/issues/1975

After some digging, the current CRC32C implementation in the storage emulator ignores the first few bytes of the hash when converted to base64 (notice the `"----"`):
https://github.com/firebase/firebase-tools/blob/81ec80303ced030d12c2c444da698686b17fa483/src/emulator/storage/crc.ts#L44-L54

In `@google-cloud/storage` v6, we now correctly check the entire CRC32C value.

### Scenarios Tested

Tested against `@google-cloud/storage` @ v6.0.1's `CRC32C` class.

Replication notes:
```ts
import {CRC32C} from '@google-cloud/storage';

const crc32c = new CRC32C();

crc32c.update(/* Buffer */);

const result = crc32c.toString();
```

I see `@google-cloud/storage` is pulled in via `firebase-admin` - once the package is updated to v6+ there,  a handy new [CRC32C](https://github.com/googleapis/nodejs-storage/blob/1d7d075b82fd24ea3c214bd304cefe4ba5d8be5c/src/crc32c.ts) class can be used for testing.

Fixes: https://github.com/googleapis/nodejs-storage/issues/1975


bkendall (approved):

bkendall:
This looks great. I'm assigning @yuchenshi to double check, but this seems perfectly reasonable.

I pushed the code to another branch to get all the tests to run as well (since that doesn't happen automatically when PRs come from forks). Happy to re-run it if necessary.

yuchenshi:
This makes sense, but I think we should update the `@google-cloud/storage` dependency in `firebase-admin` ASAP and roll out that fix too before releasing this. @lahirumaramba WDYT?

src/emulator/storage/crc.ts
```diff
@@ -43,12 +45,14 @@ export function crc32c(bytes: Buffer): number {
 
 /**
  * Adapted from:
- *  - https://github.com/googleapis/nodejs-storage/blob/0c1fa3934a52a608366a8c6c798c43516dd03dbf/src/file.ts#L1406-L1409
+ *  - https://github.com/googleapis/nodejs-storage/blob/1d7d075b82fd24ea3c214bd304cefe4ba5d8be5c/src/crc32c.ts
```
danielduhh, line 48: Should we just link the main branch?

src/emulator/storage/crc.ts
```diff
@@ -43,12 +45,14 @@ export function crc32c(bytes: Buffer): number {
...
  */
 export function crc32cToString(crc32cValue: number | string): string {
-  if (typeof crc32cValue === "string") {
-    crc32cValue = Number(crc32cValue);
-  }
-  // Does the reverse of https://stackoverflow.com/q/25096737/849645
-  return "----" + Buffer.from([crc32cValue]).toString("base64");
+  const value = typeof crc32cValue === "string" ? Number.parseInt(crc32cValue) : crc32cValue;
+
+  // Allocating 4 bytes to write an unsigned 32-bit integer
```
danielduhh, line 53: Would be helpful to link to something that explains the 4 bytes

danielduhh (commented):
Thank you @danielbankhead ! Non blocking comments...

src/emulator/storage/crc.ts
danielbankhead: This permalink keeps track of the file exactly as referenced (for context purposes) - this can help future debugging; like the link that was removed helped me trace where the code was copied from + date/time

src/emulator/storage/crc.ts
danielbankhead: Sg, updated the documentation

lahirumaramba:
> This makes sense, but I think we should update the `@google-cloud/storage` dependency in `firebase-admin` ASAP and roll out that fix too before releasing this. @lahirumaramba WDYT?

`@google-cloud/storage v6` is on TypeScript 4 so we need to upgrade TypeScript support on `firebase-admin` (a breaking change) before we can upgrade the dependencies. We are currently planning this major version bump and trying to cut an expedited release, but it will take a while and won't be done asap.

danielbankhead:
@lahirumaramba thanks for the update, @yuchenshi we could merge this in as to unblock customers in the short-term, while enhancing once `@google-cloud/storage` is updated upstream in `firebase-admin`. What do you think?

yuchenshi (approved):

yuchenshi:
@bkendall Let's rerun the tests and merge if everything works

PR #4611, opened 2022-06-02
Add commands for Crashlytics mapping files

mrichards:
Adds `crashlytics:mappingfile:generateid` and `crashlytics:mappingfile:upload`. 

Also includes a refactor commit to generalize the buildtools jar utilities previously
encapsulated in `crashlytics:symbols:upload`.

[See internal doc go/crashlytics-cli-mapping-files for API review]


src/commands/crashlytics-mappingfile-generateid.ts
```diff
@@ -0,0 +1,39 @@
...
+  debug: boolean | null;
+}
+
+interface JarOptions {
+  resourceFilePath: string;
+}
+
+export default new Command("crashlytics:mappingfile:generateid")
+  .description(
+    "Generate a mapping file id and write it to an Android resource file, which will be built into the app."
```
bkendall, line 17:
```suggestion
    "generate a mapping file id and write it to an Android resource file, which will be built into the app"
```

Following guidance from go/firebase-cli-contributor#create-a-file-for-your-command - please adjust the other options/commands in this PR

src/crashlytics/buildToolsJarHelper.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (fs.existsSync(JAR_CACHE_DIR)) {
+    logger.debug(
+      `Deleting Jar cache at ${JAR_CACHE_DIR} because the CLI was run with a newer Jar version`
+    );
+    rimraf.sync(JAR_CACHE_DIR);
+  }
+  utils.logBullet("Downloading crashlytics-buildtools.jar to " + jarPath);
+  utils.logBullet(
+    "For open source licenses used by this command, look in the META-INF directory in the buildtools.jar file"
+  );
```
bkendall, line 45: Is this something that's absolutely required to be printed every time? I don't know other jars that do this in the CLI - even apps usually bury this logic deep in some menu. If it's required, maybe it can be a debug log? I'm just not sure that it adds a whole lot of value for a user (and then I'd start wondering if I _needed_ to look at the licenses).

src/crashlytics/buildToolsJarHelper.ts
```diff
@@ -0,0 +1,73 @@
...
+    jarPath = process.env.LOCAL_JAR;
+  }
+  return jarPath;
+}
+
+/**
+ * Helper function to invoke the given set of arguments on the the executable jar
+ */
+export function runBuildtoolsCommand(jarFile: string, args: string[], debug: boolean): void {
+  args = ["-jar", jarFile].concat(args);
```
bkendall, line 61: I would _highly_ recommend not modifying the input (since it's passed by reference), but instead creating a new variable for your args.

And, if you want to be real typescripty:

```typescript
const fullArgs = [
  "-jar",
  "jarFile",
  ...args,
  "-clientName",
  "firebase-cli;...",
];
```

src/commands/crashlytics-mappingfile-upload.ts
```diff
@@ -0,0 +1,63 @@
...
+      throw new FirebaseError("set <mappingFile> to a valid mapping file path");
+    }
+    const mappingFilePath = mappingFile;
+    const resourceFilePath = options.resourceFile ? options.resourceFile : "";
+
+    const jarFile = await fetchBuildtoolsJar();
+    const jarOptions: JarOptions = { app, mappingFilePath, resourceFilePath };
+
+    utils.logBullet(`Uploading mapping file: ${mappingFilePath}`);
+    const uploadArgs = buildArgs({ ...jarOptions });
```
bkendall, line 41: Should `buildArgs(jarOptions)` be equivalent?

bkendall (changes requested):

src/crashlytics/buildToolsJarHelper.ts
mrichards: This is only printed when a new version of the buildtools jar is downloaded, which only happens the first time a Crashlytics command is executed or when we publish an update.

src/commands/crashlytics-mappingfile-generateid.ts
rachelsaunders: Is there any way to mention that the mapping file is for deobfuscating stack traces?

src/commands/crashlytics-mappingfile-upload.ts
```diff
@@ -0,0 +1,63 @@
...
+  .option(
+    "--resource-file <resourceFile>",
+    "Path to the Android resource XML file that includes the mapping file id."
+  )
+  .option("--debug", "print debug output and logging from the underlying uploader tool")
+  .action(async (mappingFile: string, options: CommandOptions) => {
+    const app = getGoogleAppID(options) || "";
+    const debug = !!options.debug;
+    if (!mappingFile) {
+      throw new FirebaseError("set <mappingFile> to a valid mapping file path");
```
rachelsaunders, line 32: Could we give more detail here about the required format of the path?

src/commands/crashlytics-mappingfile-upload.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    utils.logBullet(`Uploading mapping file: ${mappingFilePath}`);
+    const uploadArgs = buildArgs({ ...jarOptions });
+    runBuildtoolsCommand(jarFile, uploadArgs, debug);
+    utils.logBullet("Successfully uploaded mapping file");
+  });
+
+function getGoogleAppID(options: CommandOptions): string | null {
+  if (!options.app) {
+    throw new FirebaseError("Set the --app option to a valid Firebase app id and try again");
```
rachelsaunders, line 48: @bkendall What's our guidance about "and try again" in errors?  I don't feel strongly either way, but I think we should be consistent.

src/commands/crashlytics-mappingfile-upload.ts
```diff
@@ -0,0 +1,63 @@
...
+  app: string;
+  mappingFilePath: string;
+  resourceFilePath: string;
+}
+
+export default new Command("crashlytics:mappingfile:upload <mappingFile>")
+  .description("Upload a mapping file to deobfuscate stack traces.")
+  .option("--app <appID>", "The app id of your Firebase app")
+  .option(
+    "--resource-file <resourceFile>",
```
rachelsaunders, line 24: Is there an error if the resource file path isn't valid?

rachelsaunders (changes requested):

src/crashlytics/buildToolsJarHelper.ts
mrichards: Thank you! My ts n00bishness is fully exposed here. :)

src/commands/crashlytics-mappingfile-upload.ts
```diff
@@ -0,0 +1,63 @@
...
+    "Path to the Android resource XML file that includes the mapping file id."
+  )
+  .option("--debug", "print debug output and logging from the underlying uploader tool")
+  .action(async (mappingFile: string, options: CommandOptions) => {
+    const app = getGoogleAppID(options) || "";
```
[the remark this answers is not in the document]
mrichards: `options.app` is required. If throwing is the preferred behavior for a missing argument, then that's what we want. I'll remove the `|| ""` in the next commit.

src/commands/crashlytics-mappingfile-upload.ts
mrichards: Good catch, I removed the "and try again" part.

src/commands/crashlytics-mappingfile-upload.ts
mrichards: I added a check in the calling function, in the next commit.

src/commands/crashlytics-mappingfile-upload.ts
mrichards: I added examples of the path to all the errors in the next commit.

src/commands/crashlytics-mappingfile-upload.ts
mrichards: Yes, this was a copy-paste oversight. Fixed in next commit.

src/crashlytics/buildToolsJarHelper.ts
```diff
@@ -0,0 +1,73 @@
...
+  args = args.concat(["-clientName", "firebase-cli;crashlytics-buildtools"]);
+  const outputs = spawn.sync("java", args, {
+    stdio: debug ? "inherit" : "pipe",
+  });
+
+  if (outputs.status || 0 > 0) {
```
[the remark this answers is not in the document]
samedson: Somehow this was in my original commit. My hunch is this was meant to be status != 0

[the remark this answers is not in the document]
mrichards: Updated to `!=`

samedson:
Updated the JAR version. Tested `crashlytics:symbols:upload` Matt is gonna do the new commands.

src/commands/crashlytics-mappingfile-generateid.ts
```diff
@@ -0,0 +1,45 @@
...
+      );
+    }
+    const jarFile = await fetchBuildtoolsJar();
+    const jarOptions: JarOptions = { resourceFilePath };
+
+    utils.logBullet(`Updating resource file: ${resourceFilePath}`);
+    const uploadArgs = buildArgs(jarOptions);
+    runBuildtoolsCommand(jarFile, uploadArgs, debug);
+    utils.logBullet("Successfully updated mapping file id");
+  });
```
bkendall, line 41: [optional] it looks like to me that this returns `undefined`. Would it be possible to return something useful for machine output (i.e. the command run with `--json`)? 

Thinking about it a bit myself - if it's going to the JAR and doesn't really have a good way of parsing it out, might be more tricky than it's worth. `--json` output isn't required, but it does make it nicer to deal with if the CLI is run in a script

src/commands/crashlytics-mappingfile-generateid.ts
```diff
@@ -0,0 +1,45 @@
...
+  resourceFilePath: string;
+}
+
+export const command = new Command("crashlytics:mappingfile:generateid")
+  .description(
+    "generate a mapping file id and write it to an Android resource file, which will be built into the app"
+  )
+  .option(
+    "--resource-file <resourceFile>",
+    "Path to the Android resource XML file that will be created or updated."
```
bkendall, line 22: general guidance around descriptions and options: https://github.com/firebase/firebase-tools/blob/master/CONTRIBUTING.md#create-a-file-for-your-command

```suggestion
    "path to the Android resource XML file that will be created or updated"
```

src/commands/crashlytics-mappingfile-upload.ts
```diff
@@ -0,0 +1,73 @@
...
+  mappingFilePath: string;
+  resourceFilePath: string;
+}
+
+export const command = new Command("crashlytics:mappingfile:upload <mappingFile>")
+  .description("upload a ProGuard/R8-compatible mapping file to deobfuscate stack traces")
+  .option("--app <appID>", "The app id of your Firebase app")
+  .option(
+    "--resource-file <resourceFile>",
+    "Path to the Android resource XML file that includes the mapping file id."
```
bkendall, line 25:
```suggestion
    "path to the Android resource XML file that includes the mapping file id"
```

src/commands/crashlytics-mappingfile-upload.ts
```diff
@@ -0,0 +1,73 @@
...
+  )
+  .option("--debug", "print debug output and logging from the underlying uploader tool")
```
bkendall, line 27: This is a global option - it shouldn't need to be redefined and should be able to be safely removed

src/commands/crashlytics-mappingfile-generateid.ts
```diff
@@ -0,0 +1,45 @@
...
+  )
+  .option("--debug", "print debug output and logging")
```
bkendall, line 24: This is a global option - this should be able to be removed.

src/commands/crashlytics-mappingfile-upload.ts
```diff
@@ -0,0 +1,73 @@
...
+      throw new FirebaseError(
+        "set <mappingFile> to a valid mapping file path, e.g. app/build/outputs/mapping.txt"
+      );
+    }
+    const mappingFilePath = mappingFile;
+
+    const resourceFilePath = options.resourceFile;
+    if (!resourceFilePath) {
+      throw new FirebaseError(
+        "set <resourceFile> to a valid Android resource file path, e.g. app/main/res/values/strings.xml"
```
bkendall, line 41: Same here - probably want to specify the flag and/or the variable, not just the variable

src/commands/crashlytics-mappingfile-upload.ts
```diff
@@ -0,0 +1,73 @@
...
+    "--resource-file <resourceFile>",
+    "Path to the Android resource XML file that includes the mapping file id."
+  )
+  .option("--debug", "print debug output and logging from the underlying uploader tool")
+  .action(async (mappingFile: string, options: CommandOptions) => {
+    const app = getGoogleAppID(options);
+    const debug = !!options.debug;
+    if (!mappingFile) {
+      throw new FirebaseError(
+        "set <mappingFile> to a valid mapping file path, e.g. app/build/outputs/mapping.txt"
```
bkendall, line 33: Should this maybe specify the flag rather than the variable? Below would match what's in generateid:

```suggestion
        "set `--mapping-file <mappingFile>` to a valid mapping file path, e.g. app/build/outputs/mapping.txt"
```

src/commands/crashlytics-symbols-upload.ts
```diff
@@ -36,33 +29,22 @@ interface JarOptions {
...
-const JAR_URL = `https://dl.google.com/android/maven2/com/google/firebase/firebase-crashlytics-buildtools/${JAR_VERSION}/firebase-crashlytics-buildtools-${JAR_VERSION}.jar`;
 
 export const command = new Command("crashlytics:symbols:upload <symbolFiles...>")
-  .description("Upload symbols for native code, to symbolicate stack traces.")
+  .description("upload symbols for native code, to symbolicate stack traces")
   .option("--app <appID>", "the app id of your Firebase app")
-  .option("--generator [breakpad|csym]", "the symbol generator being used, defaults to breakpad.")
+  .option("--generator [breakpad|csym]", "the symbol generator being used, default is breakpad")
   .option("--dry-run", "generate symbols without uploading them")
   .option("--debug", "print debug output and logging from the underlying uploader tool")
```
bkendall, line 38: This is a global option and should be able to be omitted...

src/crashlytics/buildToolsJarHelper.ts
```diff
@@ -0,0 +1,69 @@
...
+
+const JAR_VERSION = "2.9.1";
+const JAR_URL = `https://dl.google.com/android/maven2/com/google/firebase/firebase-crashlytics-buildtools/${JAR_VERSION}/firebase-crashlytics-buildtools-${JAR_VERSION}.jar`;
+
+/**
+ * Returns the path to the jar file, downloading it if necessary.
+ */
+export async function fetchBuildtoolsJar(): Promise<string> {
+  // If you set LOCAL_JAR to a path it will override the downloaded buildtools.jar
+  if (process.env.LOCAL_JAR) {
```
bkendall, line 24: `LOCAL_JAR` is a pretty generic variable. Would `CRASHLYTICS_LOCAL_JAR` be better/safer, if more verbose?

src/crashlytics/buildToolsJarHelper.ts
```diff
@@ -0,0 +1,69 @@
...
+  const fullArgs = ["-jar", jarFile, ...args, "-clientName", "firebase-cli;crashlytics-buildtools"];
+  const outputs = spawn.sync("java", fullArgs, {
+    stdio: debug ? "inherit" : "pipe",
+  });
+
+  if (outputs.status !== 0) {
+    if (!debug) {
+      utils.logWarning(outputs.stdout?.toString() || "An unknown error occurred");
+    }
+    throw new FirebaseError("Command failed");
```
bkendall, line 67: Maybe provide a bit more context to the error?

```suggestion
    throw new FirebaseError("java command failed: args: ${fullArgs}");
```

bkendall (changes requested):
Few comments, but overall looks good!

src/commands/crashlytics-mappingfile-generateid.ts
```diff
@@ -0,0 +1,44 @@
...
+        "set --resource-file <resourceFile> to an Android resource file path, e.g. app/src/main/res/values/crashlytics.xml"
+      );
+    }
+    const jarFile = await fetchBuildtoolsJar();
+    const jarOptions: JarOptions = { resourceFilePath };
+
+    utils.logBullet(`Updating resource file: ${resourceFilePath}`);
+    const uploadArgs = buildArgs(jarOptions);
+    runBuildtoolsCommand(jarFile, uploadArgs, debug);
+    utils.logBullet("Successfully updated mapping file id");
```
bkendall, line 39: What is `file id` referring to?

src/commands/crashlytics-mappingfile-upload.ts
```diff
@@ -0,0 +1,72 @@
...
+
+interface JarOptions {
+  app: string;
+  mappingFilePath: string;
+  resourceFilePath: string;
+}
+
+export const command = new Command("crashlytics:mappingfile:upload <mappingFile>")
+  .description("upload a ProGuard/R8-compatible mapping file to deobfuscate stack traces")
+  .option("--app <appID>", "The app id of your Firebase app")
```
bkendall, line 22:
```suggestion
  .option("--app <appID>", "the app id of your Firebase app")
```

bkendall (commented):
Could you also add an entry to CHANGELOG.md? Something like

```markdown
- Adds ....
```

src/commands/crashlytics-mappingfile-generateid.ts
mrichards: Bryan & I discussed in person and decided to keep the return `undefined` for now, largely because parsing the output of the executable jar is going to be brittle. We'll revisit if we get feedback from customers using the `--json` switch.

bkendall (approved):

PR #4623, opened 2022-06-06
Find the function region for a functions rewrite automatically.

akongara-goog:
### Description

If a function is deployed in one region, it will be automatically added
to the rewrite. If the function is deployed in multiple regions, the
rewrite must specify a function region.

### Scenarios Tested

Deployed a project with a function not in us-central1 and verified that the rewrite worked.
Moved the function to a different region with a functions-only deploy. Then did a hosting-only deploy that picked up the new function region.

### Tests

All of these include/require a Hosting deploy. The setup regarding Functions should be as follows:

- while deploying the first function (0 -> 1 functions) with 1 region:
  - [x] specifying the region in `firebase.json` - should succeed
  - [x] with automatic region detection - should succeed
  - [x] specifying a different region in `firebase.json` than in code - should error
  - [x] specifying a rewrite to a function that doesn't exist and isn't being deployed - should error?
- deploying the first function (0 -> 1 functions) with multiple regions
  - [x] specifying any valid region in `firebase.json` that matches code - should succeed
  - [x] specifying a region that is invalid - should fail
  - [x] omitting a region in `firebase.json` - should error
- updating a function (1 -> 1 functions) when changing the region:
  - [x] specifying the new region in `firebase.json` - should succeed
  - [x] with automatic region detection - should succeed
  - [x] specifying a different region in `firebase.json` than in code - should error
  - [x] specifying a rewrite to a function that doesn't exist and isn't being deployed - should error?
- deploying without Functions code available
  - [x] having the region specified in `firebase.json` that matches the state of prod - should succeed
  - [x] automatic region detection - should work
  - [x] specifying an invalid region in `firebase.json` - should error


src/deploy/hosting/convertConfig.ts
```diff
@@ -1,7 +1,10 @@
 import { FirebaseError } from "../../error";
 import { HostingConfig, HostingRewrites, HostingHeaders } from "../../firebaseConfig";
+import { logger } from "../../logger";
```
bkendall, line 3: If this isn't used, please remove it.

src/deploy/hosting/convertConfig.ts
```diff
@@ -57,27 +60,99 @@ export async function convertConfig(
...
+      If the function is deployed in multiple regions, you must specify a region.`);
+    }
+
+    if (matchingBackends.length === 1) {
+      const endpoint = matchingBackends[0];
+      if (endpoint && isHttpsTriggered(endpoint)) {
+        return endpoint;
+      }
+    }
+    return undefined;
```
bkendall, line 90:
```suggestion
    return;
```

`return` with no parameter will return `undefined`

src/deploy/hosting/convertConfig.ts
```diff
@@ -57,27 +60,99 @@ export async function convertConfig(
...
+
+    const matchingBackends = backendsForId.filter((endpoint) => {
+      return (
+        (!region || endpoint.region === region) && (!platform || endpoint.platform === platform)
+      );
+    });
+
+    if (matchingBackends.length > 1) {
+      throw new FirebaseError(`More than one backend found for function name: ${serviceId}.
+      If the function is deployed in multiple regions, you must specify a region.`);
```
bkendall, line 81:
```suggestion
      throw new FirebaseError(
        `More than one backend found for function name: ${serviceId}. If the function is deployed in multiple regions, you must specify a region.`
      );
```

String literals are just that: literal. With the newline and spaces, those will show up in the error message and be really weird. The above is valid from the formatter, AFAIK.

src/deploy/hosting/convertConfig.ts
```diff
@@ -57,27 +60,99 @@ export async function convertConfig(
...
-      if (endpoint && isHttpsTriggered(endpoint) && endpoint.platform === "gcfv2") return endpoint;
+      if (!wantBackend) {
+        continue;
+      }
+      const endpoint = endpointFromBackend(wantBackend, serviceId, platform, region);
+      if (endpoint) {
+        return endpoint;
+      }
     }
     return undefined;
```
bkendall, line 107: Again here, `return;` is sufficient.

```suggestion
    return undefined;
```

src/deploy/hosting/convertConfig.ts
```diff
@@ -57,27 +60,99 @@ export async function convertConfig(
...
     if (pendingEndpoint) return pendingEndpoint;
     const backend = await existingBackend(context);
     return allEndpoints(backend).find(
       (it) =>
         isHttpsTriggered(it) &&
-        it.platform === "gcfv2" &&
+        it.platform === platform &&
         it.id === serviceId &&
-        it.region === region
+        (region === undefined || it.region === region)
```
bkendall, line 123: Does region specifically need to be undefined? Or would an empty string be a case that needs to be considered?

Normally, you wouldn't make a check against `=== undefined`, preferring instead `!thing` (since undefined is falsey). But if `""` (empty string) is technical valid, and not the same as "not defined", then what is written is correct (and I would put a comment there saying why)

src/deploy/hosting/convertConfig.ts
```diff
@@ -57,27 +60,99 @@ export async function convertConfig(
...
     );
   };
 
+  const findEndpointWithValidRegion = async (
+    rewrite: HostingRewrites,
+    context: Context
+  ): Promise<backend.Endpoint | undefined> => {
+    if ("function" in rewrite) {
```
bkendall, line 131: `in` can be fickle since it technically checks the object being looked at _and any object in its prototype chain_. That said, it's better to avoid it in my experience.

Would checking simply for `rewrite.function` be accurate? Or can it not be done for some reason?

src/deploy/hosting/convertConfig.ts
```diff
@@ -57,27 +60,99 @@ export async function convertConfig(
     return out;
   }
 
-  const endpointBeingDeployed = (serviceId: string, region: string = "us-central1") => {
+  const endpointFromBackend = (
+    targetBackend: backend.Backend,
+    serviceId: string,
+    platform?: string,
+    region?: string
+  ): backend.Endpoint | undefined => {
```
bkendall, line 68: Since you've got a couple places where `platform` isn't specified, but `region` is, why not make this take an option array so you can more easily call the function (and stop writing `undefined` in a  few places)

```suggestion
  const endpointFromBackend = (opts: {
    targetBackend: backend.Backend;
    serviceId: string;
    platform?: string;
    region?: string;
  }): backend.Endpoint | undefined => {
```

src/deploy/hosting/convertConfig.ts
```diff
@@ -57,27 +60,99 @@ export async function convertConfig(
...
+        existingBackend,
+        rewrite.function,
+        undefined,
+        rewrite.region
+      );
+      if (endpointAlreadyDeployed) {
+        return endpointAlreadyDeployed;
+      }
+    }
+    return undefined;
```
bkendall, line 153:
```suggestion
    return;
```

src/test/deploy/hosting/convertConfig.spec.ts
```diff
@@ -300,3 +360,131 @@ describe("convertConfig", () => {
...
+      errorString: "More than one backend found for function name",
+      finalize: true,
+    },
+  ];
+
+  for (const {
+    name,
+    context = DEFAULT_CONTEXT,
+    input,
+    payload = DEFAULT_PAYLOAD,
```
bkendall, line 479: IMO, having the DEFAULT_* variables set here is an anti-pattern. You want all the context for each test case _in the test case_, so they should be defined above in the cases. Please go ahead and make this change.

(this is also why I was very confused why payload was being added in the other set of tests - I didn't understand why it's _now_ returning a payload, instead of it just being changed)

bkendall (changes requested):

src/deploy/hosting/convertConfig.ts
akongara-goog: I should have made this just ( !region || ...stuff ) like elsewhere. region = "" doesn't make have a reasonable meaning separate from undefined.

src/deploy/hosting/convertConfig.ts
akongara-goog: So I looked here:
https://www.typescriptlang.org/docs/handbook/advanced-types.html
https://levelup.gitconnected.com/user-defined-type-guards-in-typescript-fad639e4944f

And it seems like rewrite.function is not valid. The other ways is to use instanceof (which basically only works on classes) and a type predicate (which needs a named type, which we don't have). It seems like we have to leave it as is or redefine the type union and add some type predicates.

src/deploy/hosting/convertConfig.ts
akongara-goog: Fixed.

src/deploy/hosting/convertConfig.ts
akongara-goog: Created a new interface for this because it's used in multiple places.

src/test/deploy/hosting/convertConfig.spec.ts
akongara-goog: Fixed.

src/deploy/functions/release/index.ts
```diff
@@ -93,6 +93,12 @@ export async function release(
   const allErrors = summary.results.filter((r) => r.error).map((r) => r.error) as Error[];
   if (allErrors.length) {
     const opts = allErrors.length === 1 ? { original: allErrors[0] } : { children: allErrors };
+    console.log("Functions deploy failed.");
+    logger.info("Functions deploy failed.");
+    for (const error of allErrors) {
+      console.log(JSON.stringify(error, null, 2));
+      logger.info(JSON.stringify(error, null, 2));
+    }
```
bkendall, line 101: These probably need to be simplified in some way? Maybe `logger.debug` the errors if you like, but my hunch is that the console.log and logger.info should be removed

bkendall (approved):

src/deploy/functions/release/index.ts
akongara-goog: Done.

PR #4624, opened 2022-06-06
Fix error when frameworkawareness is on and there is no hosting config in firebase.json…

joehan:
### Description
Currently, if previews.frameworkawareness is true & firebase.json has no hosting entry, the emulator suite throws a `TypeError: Cannot read property 'source' of undefined`


bkendall (approved):
LG. Turned off auto-merge. Do you want a CHANGELOG?

src/emulator/controller.ts
```diff
@@ -406,7 +406,7 @@ export async function startAll(
 
   if (previews.frameworkawareness) {
     const config = options.config.get("hosting");
-    if (Array.isArray(config) ? config.some((it) => it.source) : config.source) {
+    if (Array.isArray(config) ? config.some((it) => it.source) : config?.source) {
```
jamesdaniels, line 409: Can we do a similar fix for deploy?

PR #4629, opened 2022-06-07
remove api.request. finally.

bkendall:
### Description

I'm removing `api.request`. It's not used anywhere. ^_^

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -1,14 +1,10 @@
 "use strict";
 
 var _ = require("lodash");
-var querystring = require("querystring");
-var request = require("request");
 var url = require("url");
 
 var { Constants } = require("./emulator/constants");
-var { FirebaseError } = require("./error");
 const { logger } = require("./logger");
-var { responseToError } = require("./responseToError");
 var scopes = require("./scopes");
 var utils = require("./utils");
 var CLI_VERSION = require("../package.json").version;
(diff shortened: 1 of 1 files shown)


mbleigh (approved):

mikehardy:
@bkendall / @mbleigh  Oh sadness, it was used quite a bit

https://gist.github.com/puf/e00c34dd82b35c56e91adbc3a9b1c412

This was how I was pushing things to hosting.

Yes, it's an internal API but it's been a good friend for years. What would be a suitable replacement for deploying a single file to hosting?

mikehardy:
Direct use of REST API I guess, it is all I can find at the moment https://firebase.google.com/docs/hosting/api-deploy - going to maintain a pin on v11.0.0 here pending comment. Cheers

bkendall:
Ah we knew about that gist, but weren't sure how much it was used... guess we can put at least one up on the board! I'll see if I can get Puf to take a look at it :)

> What would be a suitable replacement for deploying a single file to hosting?

Well, a couple options come to mind. But, I'm going to assume a couple things: (1) you have many, many, many files - so many that `hosting deploy` still takes too long (to re-hash, gzip, upload changed ones, etc) and (2) you know which ones change from version to version.

I'd recommend using the API to fetch your current Version, and use [clone](https://firebase.google.com/docs/reference/hosting/rest/v1beta1/sites.versions/clone) with `finalize: false`. Then you can call `populateFiles` on the version to update the files necessary, upload them (to the URL in the response to `populateFiles`, and then finalize the Version and create a new Release!

mikehardy:
My use case is more about pushing a single file without having the rest of the repo in place. However, your comment (and the whole change here) has made me re-examine my previous constraint of "don't want the whole hosting site in place" and I think I can relax that, at which point it's just a normal hosting site publish and I can quit hacking around.

Was a great hack while it lasted though :-). Thanks

joehinkle11:
I'm thinking about making a fork which allows for you to add a flag to your `firebase.json` files like this:

```json
{
  "hosting": {
    "never_delete_remote_files": true,
    "public": "public",
    "ignore": [
      "firebase.json",
      "**/.*",
      "**/node_modules/**"
    ]
  }
}
```

This way you can upload new files and override files the same way as before with `firebase deploy`, but if you delete a file, it would remain on the server.

joehinkle11:
I made a PR allowing for the CLI to more or less support single file uploads. @mikehardy, I'm curious if this meets your needs.

https://github.com/firebase/firebase-tools/pull/4761

mikehardy:
That looks neat @joehinkle11 - my goal wasn't so much to save disk space though, it was just to avoid having the repo for "it's hard to generate" reasons, but I re-worked how I do things and no longer need the functionality myself, so I won't be able to collaborate here since I designed the need out, personally.

PR #4632, opened 2022-06-08
WIP: Use Build-to-Backend instead of Backend in the main CF3v2 deploy path.

Berlioz:
This PR does 3 fairly spicy things:

1) Sets the behavior of serviceAccountEmail in the build type to at least something which seems to work when I tested locally. The last time this change came up, there were suggestions that this behavior (setting serviceAccountEmail to `undefined` in the backend if not otherwise present to force the default service account) would result in incorrect behavior, but I confess I didn't entirely follow why. I'll be following up on this one before this PR stops being a WIP.

2) Flips detectFromYaml to return Promise<build.Build> instead of Promise<backend.Backend>, and plumbs that changed type through the calls that it makes and its test suite. As a consequence, we can now remove the feature guard from Builds in deploy/prepare.ts. **Merging this PR will cause all CF3v2 deploys to use the Build type.**

3) Modifies the Functions emulator to also do a discoverBuild+resolveBackend instead of a discoverSpec.

### DIFF
--- a/src/deploy/functions/build.ts
+++ b/src/deploy/functions/build.ts
@@ -331,6 +331,8 @@ export function resolveBackend(build: Build, userEnvs: Record<string, string>): 
       }
       if (bdEndpoint.serviceAccount) {
         bkEndpoint.serviceAccountEmail = bdEndpoint.serviceAccount;
+      } else {
+        bkEndpoint.serviceAccountEmail = undefined;
       }
 
       bkEndpoints.push(bkEndpoint);
--- a/src/deploy/functions/runtimes/discovery/index.ts
+++ b/src/deploy/functions/runtimes/discovery/index.ts
@@ -7,6 +7,7 @@ import { promisify } from "util";
 import { logger } from "../../../../logger";
 import * as api from "../../.../../../../api";
 import * as backend from "../../backend";
+import * as build from "../../build";
 import * as runtimes from "..";
 import * as v1alpha1 from "./v1alpha1";
 import { FirebaseError } from "../../../../error";
@@ -34,11 +35,32 @@ export function yamlToBackend(
   }
 }
 
+export function yamlToBuild(
+  yaml: any,
+  project: string,
+  region: string,
+  runtime: runtimes.Runtime
+): build.Build {
+  try {
+    if (!yaml.specVersion) {
+      throw new FirebaseError("Expect manifest yaml to specify a version number");
+    }
+    if (yaml.specVersion === "v1alpha1") {
+      return v1alpha1.buildFromV1Alpha1(yaml, project, region, runtime);
+    }
+    throw new FirebaseError(
+      "It seems you are using a newer SDK than this version of the CLI can handle. Please update your CLI with `npm install -g firebase-tools`"
+    );
+  } catch (err: any) {
+    throw new FirebaseError("Failed to parse build specification", { children: [err] });
+  }
+}
+
 export async function detectFromYaml(
   directory: string,
   project: string,
   runtime: runtimes.Runtime
-): Promise<backend.Backend | undefined> {
+): Promise<build.Build | undefined> {
   let text: string;
   try {
     text = await exports.readFileAsync(path.join(directory, "functions.yaml"), "utf8");
@@ -53,7 +75,7 @@ export async function detectFromYaml(
 
   logger.debug("Found functions.yaml. Got spec:", text);
   const parsed = yaml.load(text);
-  return yamlToBackend(parsed, project, api.functionsDefaultRegion, runtime);
+  return yamlToBuild(parsed, project, api.functionsDefaultRegion, runtime);
 }
 
 export async function detectFromPort(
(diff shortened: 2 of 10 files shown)


src/test/deploy/functions/runtimes/discovery/index.spec.ts
```diff
@@ -7,19 +7,20 @@ import * as api from "../../../../../api";
...
 
-const ENDPOINT: backend.Endpoint = {
+const ENDPOINT: build.Endpoint = {
   ...MIN_ENDPOINT,
-  id: "id",
   platform: "gcfv2",
   project: "project",
-  region: api.functionsDefaultRegion,
   runtime: "nodejs16",
+  region: ["us-central1"],
```
taeold, line 22:
```suggestion
  region: [api.functionsDefualtRegion],
```

src/test/deploy/functions/runtimes/discovery/index.spec.ts
```diff
@@ -29,24 +30,25 @@ const YAML_OBJ = {
 
 const YAML_TEXT = yaml.dump(YAML_OBJ);
 
-const BACKEND: backend.Backend = backend.of(ENDPOINT);
+// const BACKEND: backend.Backend = backend.of(ENDPOINT);
```
taeold, line 33:
```suggestion
```

src/deploy/functions/runtimes/discovery/index.ts
```diff
@@ -7,6 +7,7 @@ import { promisify } from "util";
 import { logger } from "../../../../logger";
 import * as api from "../../.../../../../api";
 import * as backend from "../../backend";
+import * as build from "../../build";
 import * as runtimes from "..";
 import * as v1alpha1 from "./v1alpha1";
 import { FirebaseError } from "../../../../error";
```
taeold, line 13: <i class='graphite__hidden'>[Re: line 17]<br/></i>

* is `yamlToBackend` still needed<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/firebase/firebase-tools/4632?utm_source=unchanged-line-comment">Graphite</a>.</span>

taeold (commented):
Hmph I'm not sure why the end-to-end test is failing.

src/deploy/functions/build.ts
```diff
@@ -331,6 +331,8 @@
       }
       if (bdEndpoint.serviceAccount) {
         bkEndpoint.serviceAccountEmail = bdEndpoint.serviceAccount;
+      } else {
+        bkEndpoint.serviceAccountEmail = undefined;
```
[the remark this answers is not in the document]
Berlioz: It's not _technically_ the same, since I don't think setting something undefined or null are equivalent (at least, it breaks my test cases), but it should be the same as far as the CF3 backend is concerned. That said, I'm having some difficulty getting renameIfPresent to work in the test cases.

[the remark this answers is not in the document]
Berlioz: Specifically, when testing the back-converted Backend object:
```
      AssertionError: expected { Object (requiredAPIs, endpoints, ...) } to deeply equal { Object (requiredAPIs, endpoints, ...) }
      + expected - actual

               "platform": "gcfv2"
               "project": "project"
               "region": "region"
               "runtime": "node14"
      -        "serviceAccountEmail": [null]
               "timeoutSeconds": 60
             }
           }
         }

```

Which is wild since a Backend can't even _have_ null as the value of serviceAccountEmail, I think?

src/deploy/functions/runtimes/discovery/index.ts
Berlioz: It's not needed anymore. My plan was to rip it (and backendFromV1Alpha1) out only after this release, when we were sure we wouldn't need to go backwards.

src/test/deploy/functions/runtimes/discovery/index.spec.ts
Berlioz: Done.

src/test/deploy/functions/runtimes/discovery/index.spec.ts
Berlioz: Done.

Berlioz:
> Hmph I'm not sure why the end-to-end test is failing.

We're not supposed to be able to run that one locally, right?

src/deploy/functions/build.ts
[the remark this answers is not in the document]
inlined: If I can frustrate you even more, there's a difference between setting it to undefined and simply not setting it.

```
> const obj = {}
undefined
> "a" in obj
false
> obj.a
undefined
> obj.a = undefined
undefined
> "a" in obj
true
> obj.a
undefined
> delete obj.a
true
> "a" in obj
false
```

Is it wrong to use `renameIfPresent` for now which will leave the field missing instead of undefined? I also need to do a really scary change where `renameIfPresent` actually copies nulls so we can unbreak the container contract, but I'll do that later.

inlined (approved):

PR #4638, opened 2022-06-10
Prompt users for some types of String and Number params

Berlioz:
This PR moves all param resolution types and code out of `build.ts` into a new file, `params.ts`, where it will stay. 

In addition, we implement the minimum version of user-defined params from the CLI--integer and string types, with text input only.


src/deploy/functions/params.ts
```diff
@@ -0,0 +1,255 @@
...
+      throw new FirebaseError(
+        "Build specified string parameter " + param.param + " with unsupported input type 'select'"
+      );
+    case "text":
+    default:
+      let prompt = `Enter a value for ${param.label || param.param}:`;
+      if (param.description) {
+        prompt += ` \n(${param.description})`;
+      }
+      return await promptOnce({
```
inlined, line 214: Doesn't this mean the user is going to get prompted on ever deploy? We need to write the params back to the appropriate .env file.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,255 @@
...
+        prompt += ` \n(${param.description})`;
+      }
+      const res = await promptOnce({
+        name: param.param,
+        type: "number",
+        default: param.default,
+        message: prompt,
+      });
+      if (!Number.isInteger(res)) {
+        // / TODO: what do we do here? is there a way to force prompts to only give us integers?
```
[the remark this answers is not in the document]
Berlioz: Ack.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,255 @@
...
+      return promptIntParam(param);
+    default:
+      throw new FirebaseError(
+        "Build specified parameter " + param.param + " with unsupported type"
+      );
+  }
+}
+
+async function promptStringParam(param: StringParam): Promise<string> {
+  if (!param.input) {
```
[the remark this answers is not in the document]
Berlioz: Ack.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,255 @@
...
+    return "";
+  } else if (/{{ params\.(\S+) }}/.test(from)) {
+    const match = /{{ params\.(\S+) }}/.exec(from);
+    if (!match) {
+      return "";
+    }
+    const referencedParamValue = paramValues[match[1]];
+    if (typeof referencedParamValue !== "string") {
+      throw new FirebaseError(
+        "Referenced numeric parameter '" +
```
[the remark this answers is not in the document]
Berlioz: Ack.

src/deploy/functions/params.ts
Berlioz: Made a note of that, but I think this PR is large enough already.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,255 @@
+import { FirebaseError } from "../../error";
+import { promptOnce } from "../../prompt";
+import * as build from "./build";
+
+/**
+ * Resolves a numeric field in a Build to an an actual numeric value.
+ * Fields can be null, literal, or a {{ }} delimited CEL expression referencing param values.
+ * Currently only the CEL literal {{ params.PARAMNAME }} is implemented.
+ */
+export function resolveInt(
```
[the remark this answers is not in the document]
Berlioz: Ack.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,333 @@
...
+  switch (param.input.type) {
+    case "select":
+      throw new FirebaseError(
+        "Build specified string parameter " + param.param + " with unsupported input type 'select'"
+      );
+    case "hardcoded":
+      return resolvedDefault || "";
+    case "text":
+    default:
+      let prompt = `Enter a value for ${param.label || param.param}:`;
```
inlined, line 287: let's always mention the param name and then add the label as optional if it exists.

E.g.

```
let prompt = `Enter a value for ${param.param}`
if (param.label) {
  prompt = `${prompt} (${param.label})`
}
```


(or vice versa; don't know which should be in the parens.

inlined (commented):
It's really good to see this all taking shape. What is the "hardCoded" input type for? Just tests? If so, it might be better to just stub the prompt library rather than extending the type system.

src/deploy/functions/build.ts
```diff
@@ -360,40 +348,43 @@ function discoverTrigger(endpoint: Endpoint): backend.Triggered {
     const bkEventFilters: Record<string, string> = {};
     for (const key in endpoint.eventTrigger.eventFilters) {
```
inlined, line 349:
```
for (const [key, value] of Object.entries(endpoint.eventTrigger.eventFiltesr)) {
  bkEventFilters[key] = params.resovleString(value, paramValues);
}
```

src/deploy/functions/build.ts
```diff
@@ -360,40 +348,43 @@ function discoverTrigger(endpoint: Endpoint): backend.Triggered {
...
     };
     if (endpoint.eventTrigger.serviceAccount) {
       bkEvent.serviceAccountEmail = endpoint.eventTrigger.serviceAccount;
     }
     if (endpoint.eventTrigger.region) {
-      bkEvent.region = resolveString(endpoint.eventTrigger.region);
+      bkEvent.region = params.resolveString(endpoint.eventTrigger.region, paramValues);
     }
     trigger = { eventTrigger: bkEvent };
   } else if ("scheduleTrigger" in endpoint) {
```
inlined, line 369: Be very careful/explicit with the `in` operator. It's almost never what you want as an iterator (you probably want `of`) and it will return true here even if you said `endpoint.scheduleTrigger = undefined`. It'll also return true (IIRC) if the property was implemented in a base class.

Different codebases have different rules around membership tests.  The reigning standard is `{}.hasOwnProperty.call(endpoint, "scheduleTrigger")` but that's arcane as can be. We generally just check for truthiness unless we know the field is a number or bool.

src/deploy/functions/build.ts
```diff
@@ -360,40 +348,43 @@ function discoverTrigger(endpoint: Endpoint): backend.Triggered {
     const bkEventFilters: Record<string, string> = {};
     for (const key in endpoint.eventTrigger.eventFilters) {
       if (typeof key === "string") {
```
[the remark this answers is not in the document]
Berlioz: It can't, but the check was required because I was using `in` instead of `of` in the loop; with that fixed this check is no longer necessary and has been removed.

src/deploy/functions/build.ts
```diff
@@ -360,40 +348,43 @@ function discoverTrigger(endpoint: Endpoint): backend.Triggered {
...
-      bkEvent.region = resolveString(endpoint.eventTrigger.region);
+      bkEvent.region = params.resolveString(endpoint.eventTrigger.region, paramValues);
     }
     trigger = { eventTrigger: bkEvent };
   } else if ("scheduleTrigger" in endpoint) {
     const bkSchedule: backend.ScheduleTrigger = {
-      schedule: resolveString(endpoint.scheduleTrigger.schedule),
-      timeZone: resolveString(endpoint.scheduleTrigger.timeZone),
+      schedule: params.resolveString(endpoint.scheduleTrigger.schedule, paramValues),
+      timeZone: params.resolveString(endpoint.scheduleTrigger.timeZone, paramValues),
```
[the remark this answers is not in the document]
Berlioz: It defaults to "America/Los_Angeles" I think, before it reaches this point.

src/deploy/functions/build.ts
```diff
@@ -310,9 +269,33 @@ export function resolveBackend(build: Build, userEnvs: Record<string, string>):
         timeoutSeconds: timeout,
         ...trigger,
       };
-      proto.renameIfPresent(bkEndpoint, bdEndpoint, "maxInstances", "maxInstances", resolveInt);
-      proto.renameIfPresent(bkEndpoint, bdEndpoint, "minInstances", "minInstances", resolveInt);
-      proto.renameIfPresent(bkEndpoint, bdEndpoint, "concurrency", "concurrency", resolveInt);
+      proto.renameIfPresent(
```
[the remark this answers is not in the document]
Berlioz: Done. Clever.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,333 @@
+import { logger } from "../../logger";
+import { FirebaseError } from "../../error";
+import { promptOnce } from "../../prompt";
+import * as build from "./build";
+
+type CEL = build.Expression<string> | build.Expression<number> | build.Expression<boolean>;
+function isCEL(expr: string | number | boolean): expr is CEL {
+  return typeof expr === "string" && expr.startsWith("{{") && expr.endsWith("}}");
```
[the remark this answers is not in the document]
Berlioz: Ack. Done.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,333 @@
...
+  // Optional human-facing value for this option (e.g. "US Central (Iowa)" instead of value
+  // "us-central1")
+  label?: string;
+
+  // Actual value of the parameter if this option is selected
+  value: T;
+}
+
+export interface SelectInput<T> {
+  type?: "select";
```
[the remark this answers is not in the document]
Berlioz: Done.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,333 @@
...
+ * Currently only the CEL identity {{ params.PARAMNAME }} is implemented.
+ */
+export function resolveString(
+  from: string | build.Expression<string>,
+  paramValues: Record<string, build.Field<string | number | boolean>>
+): string {
+  if (from == null) {
+    return "";
+  } else if (/{{ params\.(\S+) }}/.test(from)) {
+    const match = /{{ params\.(\S+) }}/.exec(from);
```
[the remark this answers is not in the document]

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,333 @@
...
+  value: T;
+}
+
+export interface SelectInput<T> {
+  type?: "select";
+
+  select: Array<SelectOptions<T>>;
+}
+
+export interface DefaultOnly {
```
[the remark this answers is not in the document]
Berlioz: Testing, now deprecated because I remembered stubs exist.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,333 @@
...
+import { FirebaseError } from "../../error";
+import { promptOnce } from "../../prompt";
+import * as build from "./build";
+
+type CEL = build.Expression<string> | build.Expression<number> | build.Expression<boolean>;
+function isCEL(expr: string | number | boolean): expr is CEL {
+  return typeof expr === "string" && expr.startsWith("{{") && expr.endsWith("}}");
+}
+function dependenciesCEL(expr: CEL): string[] {
+  return /params\.(\w+)/.exec(expr)?.slice(1) || [];
```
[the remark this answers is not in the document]
Berlioz: \w = word characters (which I think is right), \S = non-whitespace

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,333 @@
...
+  type: string,
+  expr: CEL,
+  currentEnv: Record<string, ParamValue>
+): ParamValue {
+  const deps = dependenciesCEL(expr);
+  const allDepsFound = deps.every((dep) => {
+    return Object.keys(currentEnv).includes(dep);
+  });
+  if (!allDepsFound) {
+    throw new FirebaseError("");
```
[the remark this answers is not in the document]

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,333 @@
...
+): Promise<Record<string, ParamValue>> {
+  const paramValues: Record<string, ParamValue> = {};
+
+  for (const param of params) {
+    if (userEnvs.hasOwnProperty(param.param)) {
+      if (canSatisfyParam(param, userEnvs[param.param])) {
+        paramValues[param.param] = userEnvs[param.param];
+        continue;
+      } else {
+        throw new FirebaseError("");
```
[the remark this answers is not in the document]

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,333 @@
...
+/**
+ * Returns the resolved value of a user-defined Functions parameter.
+ * Functions params are defined by the output of the Functions SDK, but their value is not set until deploy-time.
+ *
+ * For most param types, we check the contents of the dotenv files first for a matching key, then interactively prompt the user.
+ * When the CLI is running in non-interactive mode or with the --force argument, it is an error for a param to be undefined in dotenvs.
+ */
+async function promptParam(param: Param, resolvedDefault?: ParamValue): Promise<ParamValue> {
+  if (resolvedDefault !== undefined && !canSatisfyParam(param, resolvedDefault)) {
+    throw new FirebaseError("");
```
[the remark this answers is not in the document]

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,333 @@
...
+      if (isCEL(param.default)) {
+        paramDefault = resolveDefaultCEL(param.type || "", param.default, paramValues);
+      } else {
+        paramDefault = param.default;
+      }
+      paramValues[param.param] = await promptParam(param, paramDefault);
+    } else {
+      paramValues[param.param] = await promptParam(param);
+    }
+  }
```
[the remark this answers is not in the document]
Berlioz: Done-ish.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,333 @@
...
+ * Tests whether a mooted ParamValue literal is of the correct type to be the value for a Param.
+ */
+function canSatisfyParam(param: Param, value: ParamValue): boolean {
+  switch (param.type) {
+    case "string":
+      return typeof value === "string";
+    case "int":
+      return typeof value === "number" && Number.isInteger(value);
+    default:
+      throw new FirebaseError("Build specified parameter " + param + " with unsupported type");
```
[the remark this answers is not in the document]
Berlioz: Ack.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,333 @@
...
+ *
+ * For most param types, we check the contents of the dotenv files first for a matching key, then interactively prompt the user.
+ * When the CLI is running in non-interactive mode or with the --force argument, it is an error for a param to be undefined in dotenvs.
+ */
+async function promptParam(param: Param, resolvedDefault?: ParamValue): Promise<ParamValue> {
+  if (resolvedDefault !== undefined && !canSatisfyParam(param, resolvedDefault)) {
+    throw new FirebaseError("");
+  }
+
+  switch (param.type) {
```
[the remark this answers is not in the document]
Berlioz: Ack.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,333 @@
...
+    const defaultToText: TextInput<string> = { text: {} };
+    param.input = defaultToText;
+  }
+
+  switch (param.input.type) {
+    case "select":
+      throw new FirebaseError(
+        "Build specified string parameter " + param.param + " with unsupported input type 'select'"
+      );
+    case "hardcoded":
```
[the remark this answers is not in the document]
Berlioz: Testing, again; again, pointless when stubs exist. Fixed.

src/functions/env.ts
```diff
@@ -240,6 +240,21 @@ export function hasUserEnvs({
...
 
+/**
+ * Write a new environment variable into a dotenv file.
+ *
+ * Identifies one and only one dotenv file to touch using the same rules as loadUserEnvs().
+ */
+export function writeUserEnv(
+  key: string,
+  value: string,
+  { functionsSource, projectId, projectAlias, isEmulator }: UserEnvsOpts
```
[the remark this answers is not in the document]
Berlioz: Done.

src/functions/env.ts
```diff
@@ -240,6 +240,21 @@ export function hasUserEnvs({
   return findEnvfiles(functionsSource, projectId, projectAlias, isEmulator).length > 0;
 }
 
+/**
+ * Write a new environment variable into a dotenv file.
+ *
+ * Identifies one and only one dotenv file to touch using the same rules as loadUserEnvs().
+ */
+export function writeUserEnv(
```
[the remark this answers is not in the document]
Berlioz: Done.

src/deploy/functions/build.ts

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,342 @@
...
+
+/**
+ * Resolves a string field in a Build to an an actual string.
+ * Fields can be literal or a {{ }} delimited CEL expression referencing param values.
+ * Currently only the CEL identity {{ params.PARAMNAME }} is implemented.
+ */
+export function resolveString(
+  from: string | build.Expression<string>,
+  paramValues: Record<string, build.Field<string | number | boolean>>
+): string {
```
inlined, line 50: please write the above and below function again to have fewer indentations and be more readable. This is a great example for another reason to value early exit for errors & putting most of the body at main indentation level. Notice that the first if/else if are on LOC 51 & 53. Line 77 is where this condition stops. By the time I see the backup case of "return from" I no longer remember what that's the edge case for.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,342 @@
...
+  // Optional human-facing value for this option (e.g. "US Central (Iowa)" instead of value
+  // "us-central1")
+  label?: string;
+
+  // Actual value of the parameter if this option is selected
+  value: T;
+}
+
+export interface SelectInput<T> {
+  type?: "select";
```
inlined, line 160: why is this resolved? you need to remove the `?`

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,342 @@
...
+/**
+ * Tests whether a mooted ParamValue literal is of the correct type to be the value for a Param.
+ */
+function canSatisfyParam(param: Param, value: ParamValue): boolean {
+  if (param.type === "string") {
+    return typeof value === "string";
+  } else if (param.type === "int") {
+    return typeof value === "number" && Number.isInteger(value);
+  } else {
+    assertExhaustive(param);
```
inlined, line 210: nit: you don't need to put this in an else. The only time you'd put it in a block would be if you wanted to do a select statement:

```
select(param.type) {
case "string": return typeof value === "string";
case "int": return typeof value === "int" && Number.isInteger(value);
default: assertExhaustive();
}
```

Even then, I don't know if you need to put it in a default. We're using a special `never` type to hint to the compiler that this piece of code will never run. It helps the compiler recognize that enum types have been all matched and therefore we've seen all code paths that can happen.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,342 @@
...
+            " resolved to value from dotenv files " +
+            userEnvs[param.param] +
+            " of wrong type"
+        );
+      }
+    }
+
+    let paramDefault: ParamValue | undefined = param.default;
+    if (paramDefault && isCEL(paramDefault)) {
+      paramDefault = resolveDefaultCEL(param.type || "string", paramDefault, paramValues);
```
inlined, line 249: Why did you need the `|| "string"`? The type of `param.type` is `"string" | "int"`. Either type should be optional in the definition of `Param` or you shouldn't have a backup and this will fail to compile if you ever _did_ need a fallback value.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,342 @@
...
+}
+
+/**
+ * A param defined by the SDK may resolve to:
+ * - the value of a Cloud Secret in the same project with name == param name (not implemented yet), but only if it's a SecretParam
+ * - a literal value of the same type already defined in one of the .env files with key == param name
+ * - the value returned by interactively prompting the user
+ *   - the default value of the prompt comes from the SDK via param.default, which may be a literal value or a CEL expression
+ *   - if the default CEL expression is not resolvable--it depends on a param whose value is not yet known--we throw an error
+ *   - yes, this means that the same set of params may or may not throw depending on the order the SDK provides them to us in
```
inlined, line 221: Or, more interestingly, the value of one param can set the defaults of many params. The sendgrid extension, for example, asks for a collection name three different times for three different features and then asks you for a key inside that collection. Most people probably have a single collection with three keys in it. Using this behavior the customer can say:

```
const coll1 = params.defineString("COLL_1", { description: "The collection where you do foo" });
const coll2 = params.defineString("COLL_2", { description: "The collection where you bar", default: coll1 });
const coll3 = params.defineString("COLL_3", { description: "The collection where you baz", default: coll2 });
```

Then the installer saves typing in a collection name up to twice.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,342 @@
...
+        continue;
+      } else {
+        throw new FirebaseError(
+          "Parameter " +
+            param.param +
+            " resolved to value from dotenv files " +
+            userEnvs[param.param] +
+            " of wrong type"
+        );
+      }
```
inlined, line 244: Again, you can make this simpler with early exit conditions and a bit of JS tricks/style:

```
for (const param of params.filter(param => userEnvs.hasOwnProperty(parm.param)) {
  if (!canSatisfyParam(param, userEnvs[param.param])) {
    throw new FirebaseError(
      `Parameter ${param.param} resolved to value from dotenv files ` +
      `${userEnvs[param.param]}  of wrong type`
    );
  }
 paramValues[param.param] = userEnvs[param.param];
}
```

It's 30% smaller and has a 50% lower [cyclomatic complexity](https://www.perforce.com/blog/qac/what-cyclomatic-complexity#:~:text=Cyclomatic%20complexity%20(CYC)%20is%20a,the%20more%20complex%20the%20code.)

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,342 @@
...
+    let paramDefault: ParamValue | undefined = param.default;
+    if (paramDefault && isCEL(paramDefault)) {
+      paramDefault = resolveDefaultCEL(param.type || "string", paramDefault, paramValues);
+    }
+    if (paramDefault && !canSatisfyParam(param, paramDefault)) {
+      throw new FirebaseError(
+        "Parameter " + param.param + " has default value " + paramDefault + " of wrong type"
+      );
+    }
+    paramValues[param.param] = await promptParam(param, paramDefault);
```
inlined, line 256: Can we at least add a TODO to track that a param was added & we need to add it to the appropriate dotenv file?

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,342 @@
...
+ * Functions params are defined by the output of the Functions SDK, but their value is not set until deploy-time.
+ *
+ * For most param types, we check the contents of the dotenv files first for a matching key, then interactively prompt the user.
+ * When the CLI is running in non-interactive mode or with the --force argument, it is an error for a param to be undefined in dotenvs.
+ */
+async function promptParam(param: Param, resolvedDefault?: ParamValue): Promise<ParamValue> {
+  if (param.type === "string") {
+    return promptStringParam(param, resolvedDefault as string);
+  } else if (param.type === "int") {
+    return promptIntParam(param, resolvedDefault as number);
```
inlined, line 273: I know I'm being pedantic here, but we should be using `resolvedDefault as number | undefined`. As it currently stands, casting to number/string (I'm pretty sure) is performing a null assertion. This means someone can accidentally make promptIntParam and promptStringParam require the resolvedDefault parameter and everything will still compile but blow up at runtime

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,342 @@
...
+    case "select":
+      throw new FirebaseError(
+        "Build specified string parameter " + param.param + " with unsupported input type 'select'"
+      );
+    case "text":
+    default:
+      let prompt = `Enter a value for ${param.param}`;
+      if (param.label) {
+        prompt = `${prompt} (${param.label})`;
+      }
```
inlined, line 295: I _suspect_ the better developer experience is to say `Enter a value for ${param.label || param.param}` (making the env var name meaningless when the user supplied a human readable label). If you think that's not obvious we should bring up the point in the CF3 chat room.

src/functions/env.ts
```diff
@@ -240,6 +240,24 @@ export function hasUserEnvs({
   return findEnvfiles(functionsSource, projectId, projectAlias, isEmulator).length > 0;
 }
 
+/* eslint-disable */
```
inlined, line 243: Why are you disabling this?

src/test/deploy/functions/params.spec.ts
```diff
@@ -0,0 +1,134 @@
+import * as chai from "chai";
+import * as sinon from "sinon";
+chai.use(require("chai-as-promised"));
+const expect = chai.expect;
```
inlined, line 4: Do we still need to use `chai.use(require("chai-as-promised"));`? I think that's old. Most tests just have

```
import { expect } from "chai"`
```

src/test/deploy/functions/params.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+const prompt = require("../../../prompt");
```
inlined, line 5: 1. Why are you using requires instead of imports?
2. imports should have one block of external imports followed by a newline followed by a block of internal imports.

src/test/deploy/functions/params.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+  it("can use a provided literal", async () => {
+    const paramsToResolve: params.Param[] = [
+      {
+        param: "foo",
+        default: "bar",
+        type: "string",
+        input: { type: "text", text: {} },
+      },
+    ];
+    promptOnce.returns(Promise.resolve("bar"));
```
inlined, line 57: I think you can say `promptOnce.resolves("bar")`

src/test/deploy/functions/params.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    await expect(params.resolveParams(paramsToResolve, "", userEnv)).to.eventually.be.rejected;
+  });
+
+  it("can use a provided literal", async () => {
+    const paramsToResolve: params.Param[] = [
+      {
+        param: "foo",
+        default: "bar",
+        type: "string",
+        input: { type: "text", text: {} },
```
inlined, line 54: is `text: {}` required? It'd be a little weird if there was a meaningful difference between an empty object and a missing object (proto APIs aside)

src/test/deploy/functions/params.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+      {
+        param: "bar",
+        default: "{{ params.foo }}",
+        type: "string",
+        input: { type: "text", text: {} },
+      },
+    ];
+    promptOnce.returns(Promise.resolve("baz"));
+    await params.resolveParams(paramsToResolve, "", {});
+    expect(promptOnce.getCall(1).args[0].default).to.eq("baz");
```
inlined, line 80: clever!

src/test/deploy/functions/runtimes/node/parseTriggers.spec.ts
```diff
@@ -74,8 +78,8 @@ describe("addResourcesToBuild", () => {
       ...BASIC_BACKEND_ENDPOINT,
       httpsTrigger: {},
     });
-    const convertedBackend: backend.Backend = build.resolveBackend(expected, {});
-    expect(convertedBackend).to.deep.equal(expectedBackend);
+    const convertedBackend = resolveBackend(expected);
+    expect(convertedBackend).to.eventually.deep.equal(expectedBackend);
```
inlined, line 82: When an expectation is called with an `eventually` modifier it returns a promise you need to await.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,342 @@
...
+ */
+export function resolveString(
+  from: string | build.Expression<string>,
+  paramValues: Record<string, build.Field<string | number | boolean>>
+): string {
+  if (from == null) {
+    return "";
+  } else if (from.includes("{{") && from.includes("}}")) {
+    let output = from;
+    const matches = /{{ params\.(\w+) }}/.exec(from);
```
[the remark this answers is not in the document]
Berlioz: So I'm pretty sure we can't use `match` here because there's no way to pull multiple matches out of the call _and_ get capturing groups, unless I'm really misreading the MDN docs and random stack overflow questions. So, we're dealing with this by doing the `exec` iterate thing.

src/deploy/functions/params.ts
Berlioz: Took a stab at it.

src/deploy/functions/params.ts
Berlioz: Oops. Dumb. Ack.

src/deploy/functions/params.ts
Berlioz: Ah, so it's basically just checking if the compiler thinks that the type of the parameter is `never`?

src/deploy/functions/params.ts
Berlioz: I think this might have just been when it was `type?` instead of 'type' and I threw that in to make the compiler happy without realizing what I was doing. Fixed, now.

src/deploy/functions/params.ts
Berlioz: Done

src/deploy/functions/params.ts
Berlioz: Done. Agreed.

src/deploy/functions/params.ts
Berlioz: Done, though I think there's a pretty good chance that an implementation in env.ts is the next thing I work on and we can get rid of this TODO.

src/functions/env.ts
Berlioz: eslint is extremely not happy about the function parameters not being used anywhere in the body of the function, I think.

src/test/deploy/functions/params.spec.ts
Berlioz: Oh. Well. That makes more sense.

src/test/deploy/functions/params.spec.ts
Berlioz: You sure can.

src/test/deploy/functions/params.spec.ts
Berlioz: It is required, though I'm also not...actually sure what its purpose is, come to think of it. I'll catch up in chat tomorrow.

src/test/deploy/functions/runtimes/node/parseTriggers.spec.ts
Berlioz: Fixed, I think.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,332 @@
...
+import { FirebaseError } from "../../error";
+import { promptOnce } from "../../prompt";
+import * as build from "./build";
+import { assertExhaustive } from "../../functional";
+
+type CEL = build.Expression<string> | build.Expression<number> | build.Expression<boolean>;
+function isCEL(expr: string | number | boolean): expr is CEL {
+  return typeof expr === "string" && expr.includes("{{") && expr.includes("}}");
+}
+function dependenciesCEL(expr: CEL): string[] {
```
inlined, line 11: nit: newline between functions

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,332 @@
...
+  return /{{ params\.\w+ }}/.exec(expr)?.slice(1) || [];
```
inlined, line 12: Aren't you missing a capture group here?

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,332 @@
...
+ * Currently only the CEL identity {{ params.PARAMNAME }} is implemented.
+ */
+export function resolveInt(
+  from: number | build.Expression<number>,
+  paramValues: Record<string, build.Field<string | number | boolean>>
+): number {
+  if (typeof from === "number") {
+    return from;
+  }
+  if (!/{{ params\.(\w+) }}/.test(from)) {
```
inlined, line 27: just do an `exec` call once and test whether the result is truthy or not. It'll also save you from doing the null assertion.

inlined, line 27: Nit: should we anchor the expression or match any part of a string?

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,332 @@
...
+ * It's an error to call this on a CEL expression that depends on params not already known in the currentEnv.
+ */
+function resolveDefaultCEL(
+  type: string,
+  expr: CEL,
+  currentEnv: Record<string, ParamValue>
+): ParamValue {
+  const deps = dependenciesCEL(expr);
+  const allDepsFound = deps.every((dep) => {
+    return Object.keys(currentEnv).includes(dep);
```
inlined, line 178: nit: `hasOwnProperty` is oblique when called "properly" but is much faster here.

```
const allDepsFound = deps.every(dep => {}.hasOwnProperty.call(currentEnv, dep))
```

Also, I don't really care if you want to just test for truthiness since you know that currentEnv isn't going to have null values:

```
const allDepsFound = deps.every((dep) => !!currentEnv[dep])
```

(`!!` is a way to cast into a boolean to silence type concerns)

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,332 @@
...
+ *   - after prompting, the resolved value of the param is written to the most specific .env file available
+ */
+export async function resolveParams(
+  params: Param[],
+  projectId: string,
+  userEnvs: Record<string, ParamValue>
+): Promise<Record<string, ParamValue>> {
+  const paramValues: Record<string, ParamValue> = {};
+
+  for (const param of params.filter((param) => userEnvs.hasOwnProperty(param.param))) {
```
inlined, line 228: FYI, you're calling filter twice with inverted fields. There's a tool for this in `src/functional.ts`. You could theoretically say:

```
const [resolved, outstanding] = functional.partition(params, (param) => userEnvs.hasOwnProperty(param.param));

const typeMismatch = resolved.find((param) => !canSatisfyParam(param, userEnvs[param.param]));
if (typeMismatch) {
  throw new Error(...);
}

for (const param of outstanding) {
  ...
}
```

This is just an FYI, not an instance to change.

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,332 @@
...
+  } else if (param.type === "int") {
+    return promptIntParam(param, resolvedDefault as number | undefined);
+  }
+  assertExhaustive(param);
+}
+
+async function promptStringParam(param: StringParam, resolvedDefault?: string): Promise<string> {
+  if (!param.input) {
+    const defaultToText: TextInput<string> = { type: "text", text: {} };
+    param.input = defaultToText;
```
inlined, line 278: fyi: the reason you need to cast here is because `text` is a string, not a literal. Other techniques you can do:

1. Assign to `param.input` directly while keeping `"text"` a literal string
2. Keep `"text"` from type-erasing to string by saying `{ type: "text", text: {}} as const`
3. Keep as is

Just an FYI, not a request to change.

inlined (approved):
Double check the comments about capture groups, but I don't think we need another back & forth.

src/deploy/functions/params.ts
Berlioz: You're right; and the really spooky thing is that the tests are passing on this commit. I'll figure out...whatever exec is returning here, and add a case.

Berlioz: ...which means this function has to go from `exec` to `match` also.

src/deploy/functions/params.ts

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,339 @@
+import { logger } from "../../logger";
+import { FirebaseError } from "../../error";
+import { promptOnce } from "../../prompt";
+import * as build from "./build";
+import { assertExhaustive } from "../../functional";
+
+type CEL = build.Expression<string> | build.Expression<number> | build.Expression<boolean>;
+
+function isCEL(expr: string | number | boolean): expr is CEL {
```
bkendall, line 9: Again, can the type of what's passed in be `unknown`? It's also odd a number or boolean could be passed in, even though they'd never pass the check

```suggestion
function isCEL(expr: unknown): expr is CEL {
```

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,339 @@
...
+ * Currently only the CEL identity {{ params.PARAMNAME }} is implemented.
+ */
+export function resolveInt(
+  from: number | build.Expression<number>,
+  paramValues: Record<string, build.Field<string | number | boolean>>
+): number {
+  if (typeof from === "number") {
+    return from;
+  }
+  const match = /\A{{ params\.(\w+) }}\z/.exec(from);
```
bkendall, line 36: What are the \A and \z for?

src/deploy/functions/params.ts
```diff
@@ -0,0 +1,339 @@
...
+  if (!match) {
+    throw new FirebaseError("CEL evaluation of expression '" + from + "' not yet supported");
```
bkendall, line 38: prefer template strings.

```suggestion
    throw new FirebaseError(`CEL evaluation of expression "${from}" not yet supported`);
```

PR #4646, opened 2022-06-13
Improve support for optional SELECT and MULTISELECT params

joehan:
### Description
Fixes the issue in https://github.com/firebase/extensions/issues/924 - previously, this code was erroring out if you omitted a value for an optional SELECT or MULTISELECT param, even though it should not.

### Scenarios Tested
Able to deploy firestore-bigquery-export while omitting values for 
TIME_PARTITIONING_FIELD_TYPE and WILDCARD_IDS.


src/extensions/askUserForParam.ts
```diff
@@ -47,28 +47,26 @@ export function checkResponse(response: string, spec: Param): boolean {
...
         utils.logWarning(spec.validationErrorMessage || genericWarn);
         valid = false;
       }
-    });
+    }
   }
 
   if (spec.type && (spec.type === ParamType.MULTISELECT || spec.type === ParamType.SELECT)) {
-    _.forEach(responses, (r) => {
+    for (const r of responses) {
```
huangjeff5, line 62: Is it reliable to assume that if a response was not provided, then it wasn't required? Maybe we want to check the spec.params to verify required/not required - for example, in case the frontend had a bug.

huangjeff5 (approved):

PR #4649, opened 2022-06-14
Add updated rules handling to CLI

sam-gc:


src/gcp/resourceManager.ts
```diff
@@ -98,3 +101,42 @@ export async function addServiceAccountToRoles(
...
+    const bindingIndex = findIndex(
+      projectPolicy.bindings,
+      (binding: Binding) => binding.role === roleName
+    );
+
+    if (bindingIndex === -1) {
+      return false;
+    }
+
+    const binding = projectPolicy.bindings[bindingIndex];
```
bkendall, line 133: native js makes this easier

```suggestion
    const binding = projectPolicy.bindings.find((b: Binding) => b.role === roleName);
    if (!binding) {
      return false;
    }
```

src/rulesDeploy.ts
```diff
@@ -100,6 +108,48 @@ export class RulesDeploy {
     return { latestName, latestContent };
   }
 
+  async checkStorageRulesIamPermissions(rulesContent: string | undefined): Promise<void> {
```
bkendall, line 111:
```suggestion
  async checkStorageRulesIamPermissions(rulesContent?: string): Promise<void> {
```

src/rulesDeploy.ts
```diff
@@ -100,6 +108,48 @@ export class RulesDeploy {
...
+    if (rulesContent?.match(CROSS_SERVICE_FUNCTIONS) === null) {
```
bkendall, line 112: Is it necessary to check for `null` explicitly? Would this be sufficent?

```suggestion
    if (!rulesContent?.match(CROSS_SERVICE_FUNCTIONS)) {
```

bkendall (commented):
Couple tiny unsolicited things for you :)

src/rulesDeploy.ts
sam-gc: I'd like to keep this one as-is. This matches the logic we have in the Console. Based on my understanding of `match()`, it _shouldn't_ matter, but since there is a subtle semantic difference between strict null check and a general falsey check, I want to keep the two in sync

src/test/rulesDeploy.spec.ts
```diff
@@ -337,6 +339,87 @@ describe("RulesDeploy", () => {
...
+      const FIXTURE_DIR = path.resolve(__dirname, "fixtures/rulesDeployCrossService");
+      const CROSS_SERVICE_OPTIONS: { cwd: string; project: string; config: any } = {
+        cwd: FIXTURE_DIR,
+        project: "test-project",
+        config: null,
+      };
+      CROSS_SERVICE_OPTIONS.config = Config.load(CROSS_SERVICE_OPTIONS, false);
+
+      beforeEach(() => {
+        (gcp.rules.getLatestRulesetName as sinon.SinonStub).resolves(null);
```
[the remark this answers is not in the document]
sam-gc: Yeah this test file could use some love. Good fix-it week item? :)

src/rulesDeploy.ts
```diff
@@ -100,6 +109,48 @@ export class RulesDeploy {
...
+    // We have cross-service rules. Now check the P4SA permission
+    const projectNumber = await getProjectNumber(this.options);
+    const saEmail = `service-${projectNumber}@gcp-sa-firebasestorage.iam.gserviceaccount.com`;
+    try {
+      if (await serviceAccountHasRoles(projectNumber, saEmail, [CROSS_SERVICE_RULES_ROLE], true)) {
+        return;
+      }
+
+      // Prompt user to ask if they want to add the service account
+      const addRole = await promptOnce(
```
bkendall, line 126: You _might_ want to check if `this.options.nonInteractive` is true - for someone in a CI/CD environment a prompt has the chance of stalling out the command.

bkendall (approved):

PR #4654, opened 2022-06-15
Remove AR experiment

inlined:
Always use Artifact Registry and always clean up Artifact Registry. In a major version bump we can stop cleaning up GCR on normal operations.


src/gcp/cloudfunctions.ts
```diff
@@ -383,9 +381,7 @@ export async function updateFunction(
   // a protobuf.Empty value, so we have to manually add it in the missing case.
   try {
     const headers: Record<string, string> = {};
-    if (previews.artifactregistry) {
```
taeold, line 386: It looks like we can get rid of this preview flag altogther in `previews.ts`

taeold (approved):

src/gcp/cloudfunctions.ts
inlined: Yup. Looks like I accidentally forgot to save that file.

inlined:
DO NOT MERGE until we are ready with public documentation changes.

PR #4658, opened 2022-06-16
add better messaging when the emulator UI isn't started

### LINKED ISSUE #3781: `firebase emulators:start --only hosting` does not start emulator-ui

**firebase-tools:** 9.18.0

**Platform:** Ubuntu

React project builds to the `build` folder, run `firebase emulators:start --only hosting` 

Should run the emulator and be visible in the Emultor-UI

Does not run the Emulator-UI and displays the following warning:
```bash
⚠  emulators: The Emulator UI requires a project ID to start. Configure your default project with 'firebase use' or pass the --project flag.
```
The suggestions do not fix the problem, I have .firebaserc and even when setting `--project my-project-id` it does not work, nor with `GCLOUD_PROJECT=my-project-id` system variable

bkendall:
### Description

Adds better messaging when the emulator UI isn't started.

Also, uses the typescript version of `no-unused-vars` to turn off the irritating "ENUM value isn't used" warnings.

Fixes #3781

### DIFF
--- a/src/emulator/controller.ts
+++ b/src/emulator/controller.ts
@@ -747,7 +747,10 @@ export async function startAll(
     hubLogger.logLabeled(
       "WARN",
       "emulators",
-      "The Emulator UI requires a project ID to start. Configure your default project with 'firebase use' or pass the --project flag."
+      "The Emulator UI is not starting, either because none of the emulated " +
+        "features interact with the Emulator UI or it cannot determine the " +
+        "Project ID. Configure your default project with 'firebase use' or " +
+        "pass the --project flag"
     );
   }
 
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -39,6 +39,7 @@ module.exports = {
         "no-invalid-this": "off", // Turned off in favor of @typescript-eslint/no-invalid-this.
         "@typescript-eslint/no-invalid-this": ["error"],
         eqeqeq: ["error", "always", { null: "ignore" }],
+        "no-unused-vars": "off", // Off in favor of @typescript-eslint/no-unused-vars.
 
         "@typescript-eslint/ban-types": "warn", // TODO(bkendall): remove, allow to error.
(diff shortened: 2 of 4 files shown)


src/emulator/controller.ts
```diff
@@ -747,7 +747,10 @@ export async function startAll(
     hubLogger.logLabeled(
       "WARN",
       "emulators",
-      "The Emulator UI requires a project ID to start. Configure your default project with 'firebase use' or pass the --project flag."
+      "The Emulator UI is not starting, either because none of the emulated " +
+        "features interact with the Emulator UI or it cannot determine the " +
```
rachelsaunders, line 751: What does "features" mean in this context?

src/emulator/controller.ts
```diff
@@ -747,7 +747,10 @@ export async function startAll(
...
+        "Project ID. Configure your default project with 'firebase use' or " +
```
rachelsaunders, line 752: I think we could just direct them towards passing the --project flag.  Understanding "which project the CLI runs against" isn't the easiest thing to grok for a lot of folks :-/  But passing the --project flag is quick for troubleshooting that the Emulator UI isn't starting.

rachelsaunders (changes requested):

src/emulator/controller.ts
bkendall: is "products" any better? i.e. "hosting", "functions", "auth", "database" - what we can actually emulate

src/emulator/controller.ts
bkendall: Sure!

src/emulator/controller.ts
rachelsaunders: yep, "products" is more appropriate.

I'm not sure what "none of the emulated products interact with the Emulator UI" means, though?  Aren't all the products that work with the Emulator Suite also available in the Emulator UI?

bkendall: wordsmith'd

yuchenshi (approved):

rachelsaunders (commented):
LGTM

PR #4659, opened 2022-06-16
remove some lodash functions

bkendall:
### Description

Doing a light pass of some easier functions to remove. Making @joehan happier. 😄


joehan (approved):
❤️

PR #4660, opened 2022-06-16
no unused vars

bkendall:
### Description

I'm removing a ton of instances where a variable is defined, but unused. This has been a thorn in my side forever :)


src/test/extensions/resolveSource.spec.ts
```diff
@@ -1,45 +0,0 @@
-import { expect } from "chai";
```
joehan, line 1: Delete this file

src/throttler/throttler.ts
```diff
@@ -46,7 +46,7 @@ export interface ThrottlerStats {
   elapsed: number;
 }
 
-interface TaskData<T, R> {
+interface TaskData<T> {
```
joehan, line 49: Hmmm, looks like this may have been a typo & possibly should be used? `TaskData.wait.resolve` is of type `(R: any) => void`- i suspect that this should actually be something like `(val: R) => void`

joehan (approved):
2 small notes, but lgtm otherwise

src/throttler/throttler.ts
bkendall: You... are... right! This lead me down a weird path of fixing a handful of types.

src/test/extensions/resolveSource.spec.ts
bkendall: It is - that's why there's no diff on the other side :)

src/deploy/hosting/uploader.ts
```diff
@@ -212,7 +208,7 @@ export class Uploader {
   }
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
joehan, line 210: Remove

joehan (approved):
Type changes lgtm too, 1 more eslint comment to kill

PR #4666, opened 2022-06-17
remove isArray, filter, reject, isNull

bkendall:
### Description

removes `isArray`, `filter`, `reject`, and `isNull`

Also fixes a bug where the `checkValidTargetFilters` was checking _every character in the only string_ instead of each thing split by a comma. That was interesting to find.

### DIFF
--- a/src/checkValidTargetFilters.ts
+++ b/src/checkValidTargetFilters.ts
@@ -49,12 +49,6 @@ export async function checkValidTargetFilters(options: Options): Promise<void> {
       );
     }
     const targetsForFunctions = targetsForTypes(only, "functions");
-    console.error(
-      "uhh",
-      targetsForFunctions,
-      targetsHaveFilters(...targetsForFunctions),
-      targetsHaveNoFilters(...targetsForFunctions)
-    );
     if (
       targetsForFunctions.length &&
       targetsHaveFilters(...targetsForFunctions) &&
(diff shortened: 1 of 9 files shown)


src/checkValidTargetFilters.ts
```diff
@@ -24,10 +22,10 @@ export function checkValidTargetFilters(options: Options): Promise<void> {
     if (options.except) {
       return reject(new FirebaseError("Cannot specify both --only and --except"));
     }
-    if (targetContainsFilter(["database", "storage", "hosting"])) {
+    if (targetContainsFilter(["database", "storage"])) {
```
joehan, line 25: This logic seems outdated - shouldn't this trigger for all VALID_TARGETS except functions & hosting? 

alid targets are:   
"database",
  "storage",
  "firestore",
  "functions",
  "hosting",
  "remoteconfig",
  "extensions",

joehan (changes requested):
Found a logic error that we should fix while were here

src/checkValidTargetFilters.ts
bkendall: I've revamped this file to be a little better to understand. Let me know what you think.

src/checkValidTargetFilters.ts
```diff
@@ -1,37 +1,65 @@
...
       return reject(
         new FirebaseError(
-          "Filters specified with colons (e.g. --only functions:func1,functions:func2) are only supported for functions"
+          "Filters specified with colons (e.g. --only functions:func1,functions:func2) are only supported for functions and hosting"
         )
       );
     }
-    if (targetContainsFilter(["functions"]) && targetDoesNotContainFilter(["functions"])) {
+    const targetsForFunctions = targetsForTypes(only, "functions");
+    console.error(
```
joehan, line 52: Remove this

joehan (approved):
LGTM except for the leftover console.error

PR #4667, opened 2022-06-17
fix some _.map replacements

bkendall:
### Description

Unfortunately, in some cases where `_.map(undefined)` may have happened, `undefined.map` throws a worse error. I've gone through our couple PRs removing `map` and have fixed up places where that was an issue.


joehan (approved):
Ahhh, good catch

PR #4668, opened 2022-06-17
Clean up uses of _.has

joehan:
### Description
Removes _.has from the codebase.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -61,9 +61,9 @@ export class Config {
       );
     }
 
-    // Move the deprecated top-level "rules" ket into the "database" object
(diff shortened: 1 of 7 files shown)


src/config.ts
```diff
@@ -168,7 +168,7 @@ export class Config {
   }
 
   has(key: string) {
-    return _.has(this.data, key);
+    return this.data[key];
```
bkendall, line 171: To keep the return type the same:

```suggestion
    return !!this.data[key];
```

bkendall (approved):
One comment but looks ok.

PR #4670, opened 2022-06-17
remove more lodash functions - sortBy, assign, etc

bkendall:
### Description

Removing startsWith, endsWith, assign, sortBy, find... and a couple gets that let me remove the file from imports

### DIFF
--- a/src/extensions/listExtensions.ts
+++ b/src/extensions/listExtensions.ts
@@ -1,4 +1,3 @@
-import * as _ from "lodash";
 import * as clc from "cli-color";
 import Table = require("cli-table");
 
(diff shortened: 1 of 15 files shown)


src/extensions/listExtensions.ts
```diff
@@ -28,21 +27,23 @@ export async function listExtensions(projectId: string): Promise<any> {
...
+    (a, b) => new Date(b.createTime).valueOf() - new Date(a.createTime).valueOf()
+  );
   const formatted: Record<string, string>[] = [];
   sorted.forEach((instance) => {
-    let extension = _.get(instance, "config.extensionRef", "");
+    let extension = instance.config.extensionRef || "";
     let publisher;
     if (extension === "") {
-      extension = _.get(instance, "config.source.spec.name", "");
+      extension = instance.config.source.spec.name || "";
```
joehan, line 38: Do we need to null guard  this with optional chaining?

joehan (approved):

src/extensions/listExtensions.ts
bkendall: according to the types, no. it all seemed to be there. that's a lot of trust in the types though

PR #4683, opened 2022-06-27
Add the ability to ability to skip deploying no-op functions

TheIronDev:
This function adds the ability to skip deploying functions that have
not changed.

A common usecase for this is a user that only updated their hosting
files (/public) and then ran `firebase deploy`.

**Actionable Items:**

- [x] Hash function
- [x] Hash used .env
- [x] Hash secret versions
- [ ] Add a `firebase-cli-upload-hash` label
- [ ] Fetch the labels of the latest uploaded function
- [ ] Short-circuit upload if hashes match
- [ ] Skip short-circuit if the user has `--force`
- [ ] Display cli messaging informing functions were not updated (but they can force it)

### DIFF
--- a/src/deploy/functions/deploy.ts
+++ b/src/deploy/functions/deploy.ts
@@ -12,10 +12,11 @@ import * as gcf from "../../gcp/cloudfunctions";
 import * as gcfv2 from "../../gcp/cloudfunctionsv2";
 import * as backend from "./backend";
 import crypto from "crypto";
+import * as secrets from "../../functions/secrets";
 
 setGracefulCleanup();
 
-async function generateSourceHash(source: args.Source): Promise<any> {
+async function generateSourceHash(source: args.Source, wantBackend: backend.Backend): Promise<any> {
   const hash = crypto.createHash("sha256");
   const sourceFile = source.functionsSourceV2 || source.functionsSourceV1;
   // Hash the contents of the source file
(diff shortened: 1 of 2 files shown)


src/deploy/functions/deploy.ts
```diff
@@ -11,9 +11,32 @@ import * as gcs from "../../gcp/storage";
...
+async function generateSourceHash(source: args.Source): any {
+  const hash = crypto.createHash("sha256");
+  const sourceFile = source.functionsSourceV2 || source.functionsSourceV1;
+  // Hash the contents of the source file
+  if (sourceFile) {
+    const readStream = fs.createReadStream(sourceFile);
+    readStream.pipe(hash);
+    await new Promise((resolve, reject) => {
+      hash.on("end", () => resolve(hash.read()));
+      readStream.on("error", (err) => reject(err));
```
bkendall, line 27:
```suggestion
      readStream.on("error", reject);
```

src/deploy/functions/deploy.ts
```diff
@@ -11,9 +11,32 @@ import * as gcs from "../../gcp/storage";
...
+    });
+  }
+
+  // Hash any possible .env additions
+  hash.push({
+    ...process.env,
```
bkendall, line 33: Is this... all the environment variables of the shell? That's... rather a lot. Is there a way to pull the is it worth trying to pull just the relevant ones?

This worries me because even in my SSH'd shell, I have some variables that look like they change every time I log in which are never going to be relevant to deploying my functions. However, that would prevent this from ever working because my shell changes all the time. What do you think?

src/deploy/functions/deploy.ts
```diff
@@ -11,9 +11,32 @@ import * as gcs from "../../gcp/storage";
...
+  });
+
+  // TODO(tystark) Hash the contents of the secret versions; rpc needed (unless we already made this call earlier)
+  return hash.read();
```
bkendall, line 37: Consider `hash.read().toString('hex')` or similar

bkendall (changes requested):

src/deploy/functions/deploy.ts
TheIronDev: Shoot, I just tested it and you're right! (I thought it was limited to args passed, which would have included the .env

inlined:
Please ping when I should give this another look.

TheIronDev:
Closing this pull request while I implement functionality in smaller chunks.

PR #4685, opened 2022-06-28
Fix Blocking Function signature in Emulator 

colerogers:
- Fixes a small bug where blocking functions treated as an `event` function instead of an `http` function.


taeold (approved):

lisajian (commented):
I unfortunately don't have enough context to review this change - is there a bug thread that I can read to get more info? Thanks!

taeold:
@lisajian go/cf3-container-contract is a start, but unfortunately that document is quite a mouthful. I think the best way would be for you to hang out with either me or @colerogers (or both!) over a quick chat?

PR #4692, opened 2022-06-29
Statically allocate socket path for each function emulator runtime process.

taeold:
This is first of several refactoring change we plan on making on the Functions Emulator to migrate bespoke IPC protocol between the emulator and the emulated functions.

Today, the invoking an HTTP function involves an intricate dance between the emulator and the emulated function:

1) On HTTP function invocation, Functions Emulator designates a socket path ([node doc](https://nodejs.org/api/net.html#net_identifying_paths_for_ipc_connections)). The chosen socket path is passed to the emulated process via IPC message sends.

2) Emulated function spins up an ephemeral express server, binds to the socket path, and responds to the HTTP requests.

This change proposes that instead of generating the socket path on _every_ request, we do it once at the emulated function process creation time.

FWIW, the sock path that was being "dynamically" generated at function invocation already generates the same socket path for each emulated function anyway. The change here simply generates a random socket path and saves it along with other metadata associated with the emulated function process.

There are obvious follow ups to this change (e.g. why do we spin up an ephemeral server on every function call?). This PR intentionally isn't tackling them to make the PR small. You can expect them on later PRs.

### DIFF
--- a/src/emulator/functionsEmulatorShared.ts
+++ b/src/emulator/functionsEmulatorShared.ts
@@ -1,9 +1,11 @@
-import * as _ from "lodash";
-import { CloudFunction } from "firebase-functions";
 import * as os from "os";
 import * as path from "path";
-import * as express from "express";
 import * as fs from "fs";
+import { randomBytes } from "crypto";
+
+import * as _ from "lodash";
+import * as express from "express";
+import { CloudFunction } from "firebase-functions";
 
 import * as backend from "../deploy/functions/backend";
 import { Constants } from "./constants";
@@ -270,7 +272,7 @@ export function getEmulatedTriggersFromDefinitions(
   );
 }
 
-export function getTemporarySocketPath(pid: number, cwd: string): string {
+export function getTemporarySocketPath(): string {
   // See "net" package docs for information about IPC pipes on Windows
   // https://nodejs.org/api/net.html#net_identifying_paths_for_ipc_connections
   //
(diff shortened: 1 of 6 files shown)


src/emulator/functionsEmulatorShared.ts
```diff
@@ -1,9 +1,11 @@
-import * as _ from "lodash";
-import { CloudFunction } from "firebase-functions";
 import * as os from "os";
 import * as path from "path";
-import * as express from "express";
 import * as fs from "fs";
+import { randomBytes } from "crypto";
+
```
joehan, line 5: unneeded newline?

joehan (approved):
Sorry for the slow review on this, code looks good but I kept going back and forth on whether this change is going in the right direction. I'm a little hesitant about moving socketPath further 'up' the code (and if sockets are the right choice over ports), but I like getting rid of the awkward 'get the socket from the lastArgs' and I think this is cleaner as a whole

taeold:
@joehan Thanks for the feedback. As I play with the code, I beginning to think of unix domain sockets as equivalent of PORTs since something like this works just fine:

```js
const app = express();
app.listen(/* insert socket file path */)
```

I think runtime being unaware whether it's talking over unix socket or TCP port is a good thing. The ultimate goal I'm chasing after is to allow the GCF's functions framework work just as well as a runtime harness. That way, onboarding a new language/runtime to the emulator would have an easy integration path with follow up work to support various debugging features like allow debugger to attach to the process.

PR #4694, opened 2022-06-30
Add Auth Blocking Integration Tests

colerogers:
- Adds integration tests for auth blocking functions
- Updates packages

### DIFF
--- a/scripts/triggers-end-to-end-tests/triggers/index.js
+++ b/scripts/triggers-end-to-end-tests/triggers/index.js
@@ -99,12 +99,14 @@ exports.createUserFromAuth = functions.https.onRequest(async (req, res) => {
   await createUserWithEmailAndPassword(auth, "email@gmail.com", "mypassword");
 
   res.json({ created: "ok" });
+  return Promise.resolve();
 });
 
 exports.signInUserFromAuth = functions.https.onRequest(async (req, res) => {
   await signInWithEmailAndPassword(auth, "email@gmail.com", "mypassword");
 
   res.json({ done: "ok" });
+  return Promise.resolve();
 });
 
 exports.writeToDefaultStorage = functions.https.onRequest(async (req, res) => {
--- a/scripts/triggers-end-to-end-tests/triggers/index.js
+++ b/scripts/triggers-end-to-end-tests/triggers/index.js
@@ -17,18 +17,19 @@ const PUBSUB_SCHEDULED_TOPIC = "firebase-schedule-pubsubScheduled";
 const STORAGE_FILE_NAME = "test-file.txt";
 
 const pubsub = new PubSub();
+
 admin.initializeApp();
 
 // firebase.initializeApp();
 // const auth = firebase.auth();
 // auth.useEmulator("http://localhost:9099");
 
-exports.createUserFromAuth = functions.https.onRequest(async (req, res) => {
+exports.createUserFromAuth = functions.https.onRequest((req, res) => {
   // await auth.createUserWithEmailAndPassword("email@gmail.com", "password");
   res.json({ created: "ok" });
 });
 
-exports.signInUserFromAuth = functions.https.onRequest(async (req, res) => {
+exports.signInUserFromAuth = functions.https.onRequest((req, res) => {
   // await auth.signInWithEmailAndPassword("email@gmail.com", "password");
   res.json({ created: "ok" });
 });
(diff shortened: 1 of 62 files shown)


scripts/triggers-end-to-end-tests/triggers/index.js
```diff
@@ -70,6 +95,18 @@ exports.writeToAuth = functions.https.onRequest(async (req, res) => {
   res.json({ created: "ok" });
 });
 
+exports.createUserFromAuth = functions.https.onRequest(async (req, res) => {
+  await createUserWithEmailAndPassword(auth, "email@gmail.com", "mypassword");
```
taeold, line 99: tangent - I think you can create user using the Admin SDK (https://firebase.google.com/docs/auth/admin/manage-users#create_a_user)

scripts/triggers-end-to-end-tests/triggers/index.js
```diff
@@ -70,6 +95,18 @@ exports.writeToAuth = functions.https.onRequest(async (req, res) => {
...
+
+  res.json({ created: "ok" });
+});
+
+exports.signInUserFromAuth = functions.https.onRequest(async (req, res) => {
+  await signInWithEmailAndPassword(auth, "email@gmail.com", "mypassword");
```
taeold, line 105: In theory, we don't need to init and use the JS SDK - instead we can call the REST API directly (https://firebase.google.com/docs/reference/rest/auth/#section-sign-in-email-password).

This will be more bloated, but could be less weird (weird to me since we are using JS SDK on server-side)[

scripts/triggers-end-to-end-tests/triggers/index.js
```diff
@@ -16,8 +25,24 @@ const PUBSUB_SCHEDULED_TOPIC = "firebase-schedule-pubsubScheduled";
...
+    apiKey: "fake-api-key",
+    projectId: `${FIREBASE_PROJECT}`,
+    authDomain: `${FIREBASE_PROJECT}.firebaseapp.com`,
+    storageBucket: `${FIREBASE_PROJECT}.appspot.com`,
+    appId: "fake-app-id",
+  },
+  "TRIGGERS_END_TO_END"
+);
+const auth = getAuth(app);
+connectAuthEmulator(auth, "http://localhost:9099");
```
taeold, line 44:
```suggestion
connectAuthEmulator(auth, process.env.FIREBASE_AUTH_EMULATOR_HOST);
```

This is probably less fragile in case we change the port assignment for the auth emulator.

taeold (commented):
Minor quibble about using/initializing JS SDK in server-side code.

scripts/triggers-end-to-end-tests/triggers/index.js
colerogers: Unfortunately, using the Admin SDK here will not trigger auth blocking functions

taeold: Dang... does that make sense? Is that how it works in prod as well?

scripts/triggers-end-to-end-tests/triggers/index.js
colerogers: hmm, yeah that makes sense, lemme try and put something together that use it

scripts/triggers-end-to-end-tests/triggers/index.js
colerogers: Agreed, I'll fix it, thanks

scripts/triggers-end-to-end-tests/triggers/index.js
```diff
@@ -16,8 +25,24 @@ const PUBSUB_SCHEDULED_TOPIC = "firebase-schedule-pubsubScheduled";
...
+const app = initializeApp(
+  {
+    apiKey: "fake-api-key",
+    projectId: `${FIREBASE_PROJECT}`,
+    authDomain: `${FIREBASE_PROJECT}.firebaseapp.com`,
+    storageBucket: `${FIREBASE_PROJECT}.appspot.com`,
+    appId: "fake-app-id",
+  },
+  "TRIGGERS_END_TO_END"
+);
```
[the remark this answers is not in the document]
colerogers: Ideally it isn't, but the two SDK's each require an init

scripts/triggers-end-to-end-tests/triggers/index.js
taeold: Well if we can't get rid of using admin sdk entirely... then I'm fine leaving this as is!

taeold (approved):
TIL: You can use the JS SDK in Node.js without too much trouble.

scripts/triggers-end-to-end-tests/triggers/index.js
colerogers: Yeah, we'd be able to get rid of the JS SDK not the admin SDK

PR #4700, opened 2022-07-01
Record etags in .firebaserc and check them before Extension deploys

joehan:
### Description
During ext:export and deploy, record the etags of all instances on them project. Then, before subsequent deploys, compare the current etags to the stored ones to detect out of band changes. If any are detected, warn and ask the user if they want to continue.

### Scenarios Tested
I tested and confirmed the following cases:
ext:export, then deploy (no warning)
deploy, make a change to extension manifest, deploy again (no warning)
ext:export, deploy, make a change via console, deploy (warning)


src/extensions/etags.ts
```diff
@@ -0,0 +1,42 @@
...
+// and returns the ids on any instances have different etags.
+export function detectEtagChanges(
+  rc: RC,
+  projectId: string,
+  instances: { instanceId: string; etag?: string }[]
+): string[] {
+  const lastDeployedEtags = rc.getEtags(projectId).extensionInstances;
+  const currentEtags = etagsMap(instances);
+  // If we don't have a record of the last deployed state, detect no changes.
+  if (!lastDeployedEtags || !Object.keys(lastDeployedEtags).length) {
```
huangjeff5, line 21: If we don't have tags recorded but already have an exported manifest, and we deployed, that would override silently right? How does this compare to forcing them to export first before deploying?

joehan: Unfortunately, if we don't have this if statement to skip when there is no recorded etags, we'll trigger this for _everyone_ on the first deployment on the version this releases in. I didn't want to inaccurately trigger this for everyone, so i figured it was better to only trigger if we actually have etags.

Note that we also have a separate check that warns if you have existing instances that don't match what you are deploying - this should cover some of these cases.

Issue #4706, opened 2022-07-04
Get an error: An unexpected error has occurred, when try to get runtimeconfig(firebase functions:config:get)


maylorsan:
### Environment info

**firebase-tools:** 11.2.0

**node**:  16.15.0

**Platform:** macOS

### Test case

Run command:
```
firebase functions:config:get
```

Get error:
<img width="572" alt="Screen Shot 2022-07-04 at 11 38 29" src="https://user-images.githubusercontent.com/51034456/177116608-c136d132-d2a8-4458-9fac-b8726e2d1c2a.png">

firebase-debug.log
```
[debug] [2022-07-04T08:09:43.150Z] >>> [apiv2][query] GET https://runtimeconfig.googleapis.com/v1beta1/projects/*project_id*/configs/env/variables/backups/project_id [none]
[debug] [2022-07-04T08:09:43.209Z] TypeError: Cannot read properties of undefined (reading 'map')
    at traverseVariables (/usr/local/lib/node_modules/firebase-tools/lib/functionsConfig.js:90:38)
    at Object.materializeConfig (/usr/local/lib/node_modules/firebase-tools/lib/functionsConfig.js:95:11)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Promise.all (index 1)
    at async Object.materializeAll (/usr/local/lib/node_modules/firebase-tools/lib/functionsConfig.js:105:5)
    at async Command.actionFn (/usr/local/lib/node_modules/firebase-tools/lib/commands/functions-config-get.js:32:20)
[error] 
[error] Error: An unexpected error has occurred.
```

**If downgrade to firebase-tools version 10.2.2 it is fixed.**

### Steps to reproduce

Run commands:
```
firebase login
firebase init
firebase functions:config:get
```

### Expected behavior

Getting variables from runtimeconfig

### Actual behavior

Getting an error

haroldadmin:
I can confirm that I get this error too, but only on CI builds in GitHub Actions. It works fine locally.

Here's a screenshot of the error:
<img width="651" alt="Screenshot 2022-07-04 at 2 47 09 PM" src="https://user-images.githubusercontent.com/24315306/177123985-6734e945-7508-474a-831c-256d789c7d30.png">

I can confirm the following:
- Firebase CLI is installed on the workflow runner
- The `TOKEN` environment variable is correctly set to the token generated using `firebase login:ci`
- This command used to run successfully on CI until a few days ago.

[...]

PR #4707, opened 2022-07-04
Fix bug with undefined variables

maylorsan:
### Description

A bug related to this [issue](https://github.com/firebase/firebase-tools/issues/4706) has been fixed.

### Scenarios Tested

When running command: `firebase functions:config:get` get an error: 
image

I found out that I have 2 configurations, but 1(configs/stripe) does not have a variable values(it is empty config).

Screen Shot 2022-07-04 at 14 08 37

After that, I added a check in the function materializeConfig check for this case

### Sample Commands

```
firebase login
firebase init
firebase functions:config:get
```

### DIFF
--- a/src/functionsConfig.ts
+++ b/src/functionsConfig.ts
@@ -127,6 +127,9 @@ export async function materializeConfig(configName: string, output: any): Promis
   };
 
   const variables = await runtimeconfig.variables.list(configName);
+
+  if (!variables) return output;
+
   await traverseVariables(variables);
   return output;
 }


maylorsan (commented):
@taeold Can u review it?

src/gcp/runtimeconfig.ts
```diff
@@ -50,7 +50,7 @@ function deleteConfig(projectId: string, configId: string): Promise<any> {
     });
 }
 
-function listVariables(configPath: string): Promise<any> {
+function listVariables(configPath: string): Promise<{ name: string }[] | undefined> {
```
taeold, line 53: what do you think that if instead of returning either an arry or an undefined, we always returned an array by doing something like this:

```
...
    .then((resp) => {
      return Promise.resolve(resp.body.variables || []);
    });
```

Then we won't need any of the `if (!variables) return {};` jazz.

Issue #4703, opened 2022-07-02
11.2.0 Functions no longer deploy


nery:
### [REQUIRED] Environment info

**firebase-tools:**  11.2.0

**Platform:** Ubuntu, macOS
**Node:**  14.19.0
**Package Manager** NPM

### [REQUIRED] Test case

Create a new project and include a new `helloWorld` function. 
Update firebase config with key value pairs of both `string` and `object` values.
Try to deploy said function.

### [REQUIRED] Steps to reproduce

Type `firebase deploy --only functions:helloWorld` or `firebase deploy --only functions` in a console on a newly created project.

### [REQUIRED] Expected behavior

The firebase function(s) are deployed.

### [REQUIRED] Actual behavior

Deployment fails in all cases. Temporary resolution is rolling back to using an older version of `firebase-tools@11.0.0`

```logs
[2022-07-02T05:21:46.801Z] >>> [apiv2][query] GET https://runtimeconfig.googleapis.com/v1beta1/projects/***/configs [none]

[2022-07-02T05:21:46.920Z] <<< [apiv2][status] GET https://runtimeconfig.googleapis.com/v1beta1/projects/***/configs 200

[2022-07-02T05:21:46.921Z] <<< [apiv2][body] GET https://runtimeconfig.googleapis.com/v1beta1/projects/***/configs {*******}

[2022-07-02T05:21:46.923Z] >>> [apiv2][query] GET https://runtimeconfig.googleapis.com/v1beta1/projects/***/configs/cloudtask****url/variables [none]

[2022-07-02T05:21:47.013Z] <<< [apiv2][status] GET https://runtimeconfig.googleapis.com/v1beta1/projects/***/configs/cloudtask****url/variables 200

[2022-07-02T05:21:47.013Z] <<< [apiv2][body] GET https://runtimeconfig.googleapis.com/v1beta1/projects/***/configs/cloudtask****url/variables {}

[2022-07-02T05:21:47.014Z] Got unexpected error from Runtime Config; it has no status code:Cannot read property 'map' of undefined {"level":"debug"}
```
*Some pats have been scrubbed for data sensitivity reasons*

anandsathe67:
Running into this as well. Disappointed how a release can be put out when basic deployment fails. Confirming downgrade to 11.0.0 works

AriDevC:
I'm facing the same issue, after updating firebase-tools.

dawogfather:
[...]

PR #4709, opened 2022-07-06
Fix check valid target filters bug

TravisSperry:
### Description

This addresses a bug introduced in 11.2.0 where only options are parsed incorrectly and when specifying "except" targets "only" targets are always determined to be present.

The following error began to propagate on a build pipeline that had no changes to the deploy scripts. The only change was the 11.2.0 release of`firebase-tools`. Downgrading to 11.1.0 resolved the issue.
```
Error: Cannot specify both --only and --except
```
Screen Shot 2022-07-06 at 9 00 01 AM

This also adds test coverage to `checkValidTargetFilters.ts`.

Closes: #4703 

### Scenarios Tested

Unit test coverage.

### Sample Commands

N/A

### DIFF
--- a/src/test/checkValidTargetFilters.spec.ts
+++ b/src/test/checkValidTargetFilters.spec.ts
@@ -1,6 +1,4 @@
-import * as chai from "chai";
-chai.use(require("chai-as-promised"));
-const expect = chai.expect;
+import { expect } from "chai";
 
 import { Options } from "../options";
 import { RC } from "../rc";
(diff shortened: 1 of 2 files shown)


src/test/checkValidTargetFilters.spec.ts
```diff
@@ -0,0 +1,72 @@
+import * as chai from "chai";
+chai.use(require("chai-as-promised"));
+const expect = chai.expect;
```
bkendall, line 3: We hook in `chai-as-promised` in our bootstrap file in tests, so you can reduce this to:

```suggestion
import { expect } from "chai";
```

bkendall:
This looks promising. I'll take a look at this and test it out later today.

Left one comment for you, but I'll get the tests running in the meantime.

TravisSperry:
@bkendall sounds good! Thanks!

bkendall (approved):

Issue #4697, opened 2022-07-01
Invalid service account: default while deploying functions with firebase-tools@11.2.0


ibrahimduran:
### [REQUIRED] Environment info

**firebase-tools:**
11.2.0

**Platform:**
Windows 11 - 21H2

### [REQUIRED] Test case

Firebase cli fails to deploy functions for new project. Downgrading to 11.1.0 works.

### [REQUIRED] Steps to reproduce

1. Install `firebase-tools@11.2.0`
2. Create firebase project or use existing one (?)
3. Initialize functions codebase using `firebase init`
4. Run deployment by `firebase deploy --debug --only functions`

### [REQUIRED] Expected behavior

Successful deployment. 

### [REQUIRED] Actual behavior

Fails deployment by `Invalid function service account requested: default.` error.

```
[2022-07-01T07:32:03.535Z] >>> [apiv2][body] POST https://cloudfunctions.googleapis.com/v1/projects/my-project/locations/us-central1/functions {"name":"projects/my-project/locations/us-central1/functions/helloWorld","sourceUploadUrl":"...","entryPoint":"helloWorld","runtime":"nodejs16","dockerRegistry":"ARTIFACT_REGISTRY","labels":{"deployment-tool":"cli-firebase"},"httpsTrigger":{"securityLevel":"SECURE_ALWAYS"},"serviceAccountEmail":"default","environmentVariables":{"FIREBASE_CONFIG":"{\"projectId\":\"my-project\",\"storageBucket\":\"my-project.appspot.com\",\"locationId\":\"europe-west\"}","GCLOUD_PROJECT":"my-project"},"timeout":"60s"}
[2022-07-01T07:32:03.816Z] <<< [apiv2][status] POST https://cloudfunctions.googleapis.com/v1/projects/my-project/locations/us-central1/functions 400
[2022-07-01T07:32:03.816Z] <<< [apiv2][body] POST https://cloudfunctions.googleapis.com/v1/projects/my-project/locations/us-central1/functions {"error":{"code":400,"message":"Invalid function service account requested: default. Please visit https://cloud.google.com/functions/docs/troubleshooting for in-depth troubleshooting documentation.","status":"INVALID_ARGUMENT"}}
!  functions: failed to create function projects/my-project/locations/us-central1/functions/helloWorld
Failed to create function projects/my-project/locations/us-central1/functions/helloWorld
```

Schmale97:
I am also experiencing this when deploying from ubuntu based docker image

[...]

PR #4714, opened 2022-07-06
Fix bug where function deployment failed when service account is not specified

taeold:
This bug affects projects using Firebase Functions SDK version 3.19.0 and below.

The CLI incorrectly sets `serviceAccountEmail` to "default" when no service account is specified in user code. Going forward, we should probably just remove the "default" as a shorthand - users can just set serviceAccount to undefined.

Fixes https://github.com/firebase/firebase-tools/issues/4697

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Fix bug where function deployed failed with message "Invalid function service account requested: default." (#4714).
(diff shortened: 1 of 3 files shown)


inlined:
I think you mean they can set the service account to null? FWIW I think a lot of these are broken and we're going to need a separate sprint to make sure settings can be reset.

inlined (approved):

taeold:
@inlined Agreed. That's a pretty rough/undefined territory we need to get a better handle on.

PR #4715, opened 2022-07-06
Ensure artifact registry API on function deploys.

taeold:
We only check but not enable Artifact Registry API on function deploys. Since our last release, all Function deploys requires AR API.


taeold (commented):
Add changelog.

TheIronDev (approved):
Lgtm

inlined (approved):

PR #4722, opened 2022-07-08
fix filter check to allow firestore targets

### LINKED ISSUE #4698: Can no longer only deploy firestore rules or firestore indexes

**firebase-tools:** 11.2.0

**Platform:** any

### [REQUIRED] Test case

Just try to run the command

Run `firestore deploy --only firestore:indexes`

Only the firestore indexes would be deployed (and not the rules)

An error:

> Filters specified with colons (e.g. --only functions:func1,functions:func2) are only supported for functions and hosting

This change was introduced in https://github.com/firebase/firebase-tools/pull/4666 and I'm just wondering if this was intentional, since the commit message doesn't say anything about changing this behavior.

Until this commit, I was able to deploy only the rules or only the index (or both), but now I'm no longer able to do so.

Is there a reason behind this change, or was this an accidental side effect of a cleanup action?

bkendall:
### Description

Fixes #4698

`--only firestore:rules` was not working (or `:indexes`) because the check only allowed `hosting` or `functions` to specify targets like that. This check was *fixed* when it was rewritten to actually return errors, but it didn't include `firestore` in the allowed list for targets.

### Scenarios Tested

* `firebase deploy --only firestore:rules`
* `firebase deploy --except firestore:rules`
* `firebase deploy --only hosting`
* `firebase deploy --only hosting:main`

### DIFF
--- a/src/checkValidTargetFilters.ts
+++ b/src/checkValidTargetFilters.ts
@@ -44,7 +44,7 @@ export async function checkValidTargetFilters(options: Options): Promise<void> {
     if (targetsForNonFilteredTypes.length && targetsHaveFilters(...targetsForNonFilteredTypes)) {
       return reject(
         new FirebaseError(
-          "Filters specified with colons (e.g. --only functions:func1,functions:func2) are only supported for functions and hosting"
+          "Filters specified with colons (e.g. --only functions:func1,functions:func2) are only supported for functions, hosting, and firestore"
         )
       );
     }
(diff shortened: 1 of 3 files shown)


yuchenshi (approved):

src/checkValidTargetFilters.ts
```diff
@@ -38,13 +38,13 @@ export async function checkValidTargetFilters(options: Options): Promise<void> {
       return reject(new FirebaseError("Cannot specify both --only and --except"));
     }
     const nonFilteredTypes = VALID_DEPLOY_TARGETS.filter(
-      (t) => !["hosting", "functions"].includes(t)
+      (t) => !["hosting", "functions", "firestore"].includes(t)
```
dalenguyen, line 41: Hi @bkendall, just want to chime in. Does it also support storage? 

Our pipeline also deploy storage rules with --only

```
firebase deploy --only storage:rules 
```

PR #4723, opened 2022-07-08
remove dns flag from emulator UI startup

bkendall:
### Description

This flag causes issues starting the emulator UI in firepit. Unfortunately, this may break connections from the UI to the emulators, but we're going to have to deal with that a different way.


yuchenshi (approved):

yuchenshi:
Is there any chance that we can remove the flag only if firepit is running? If there's no straightforward way, feel free to ignore this comment for now

bkendall:
Yeah, I don't think there is, unfortunately. There may be an env variable? But I don't know it

PR #4733, opened 2022-07-11
Get rid of _.isPlainObject

joehan:
### Description
Get rid of _.isPlainObject(). Also, fix test-functions-config so that I could verify that these changes were safe.

isPlainObject behaves slightly differently than `typeof == 'object' && !==null` - it returns false if the object prototype has a constructor. However, AFAICT, the difference isn't relevant in these 2 uses.

### Scenarios Tested
Screen Shot 2022-07-11 at 10 14 42 AM

### DIFF
--- a/scripts/test-functions-config.js
+++ b/scripts/test-functions-config.js
@@ -18,7 +18,6 @@ var tmp = require("tmp");
 
 var api = require("../lib/api");
 var scopes = require("../lib/scopes");
-var { configstore } = require("../lib/configstore");
 
 var projectId = process.argv[2] || "functions-integration-test";
 var localFirebase = __dirname + "/../lib/bin/firebase.js";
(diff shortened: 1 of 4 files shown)


scripts/test-functions-config.js
```diff
@@ -29,7 +29,6 @@ var preTest = function () {
   var dir = tmp.dirSync({ prefix: "cfgtest_" });
   tmpDir = dir.name;
   fs.copySync(projectDir, tmpDir);
-  api.setRefreshToken(configstore.get("tokens").refresh_token);
```
taeold, line 32: is this intentional?

taeold (approved):

scripts/test-functions-config.js
bkendall: It does exist in apiv2, if that matters. I'm not entirely sure though :)

bkendall (approved):

PR #4737, opened 2022-07-13
Fix invoker type mismatch for https functions.

akongara-goog:
### Description

Fix invoker type mismatch for https functions. Invoker was string[] in TriggerAnnotation but ServiceAccount (a.k.a string) in build.HttpsTrigger. This was causing problems with explicitly set invokers.

### Scenarios Tested

- Deploying with no specified invoker.
- Deploying with invoker "public".
- Deploying with invoker "private".
- Deploying with invoker set to 1 service account.
- Deploying with invoker set to 2 service accounts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Fix invoker type mismatch for https functions.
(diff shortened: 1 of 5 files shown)


taeold (approved):
I wish we had some test case for `discoverTrigger` function in `build.ts`.

It seems like that is where the bug was hidden, and unsurprisingly that function isn't being tested at all.

taeold (approved):
Can we add an entry on CHANGELOG.md?

akongara-goog:
> Can we add an entry on CHANGELOG.md?

Done.

> I wish we had some test case for `discoverTrigger` function in `build.ts`.
> 
> It seems like that is where the bug was hidden, and unsurprisingly that function isn't being tested at all.

Added a test with just a couple of basic cases that we can add on to.

PR #4738, opened 2022-07-13
utility files cleanup with tests

bkendall:
### Description

Some very trivial error message updates, typing, and tests.

### DIFF
--- a/src/requireConfig.ts
+++ b/src/requireConfig.ts
@@ -1,11 +1,18 @@
 import { FirebaseError } from "./error";
 import { Options } from "./options";
 
+/**
+ * Rejects if there is no config in `options`.
+ */
 export async function requireConfig(options: Options): Promise<void> {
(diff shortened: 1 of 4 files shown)


src/requireConfig.ts
```diff
@@ -1,11 +1,18 @@
 import { FirebaseError } from "./error";
 import { Options } from "./options";
 
+/**
+ * Rejects if there is no config in `options`.
+ */
 export async function requireConfig(options: Options): Promise<void> {
```
joehan, line 7: nit: no longer needs to be marked async since there's no await

joehan (approved):

src/requireConfig.ts
bkendall: but, but, it's so nice to keep it there! doesn't let it _not_ return a Promise by accident too (even though the return type there helps)

Issue #4735, opened 2022-07-13
Deploy functions not working


arthurgubaidullin:
### [REQUIRED] Environment info

**firebase-tools:** 11.2.2

**Platform:** macOS

### [REQUIRED] Test case

`firebase deploy --only functions`

### [REQUIRED] Steps to reproduce

`firebase deploy --only functions`

### [REQUIRED] Expected behavior

Must works.

### [REQUIRED] Actual behavior

Throw error: `Error: CEL evaluation of expression 'undefined' not yet supported`

NDA.

bkendall:
Deploying functions from my test projects seems to be working just fine. There's not enough information here to debug your issues. Can you provide a [reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) that shows this error?

ericvera:
I received the same error when starting the functions emulator with 12.2.2. Works fine with 12.2.1.

taeold:
@ericvera Do you mind sharing some of your code that's surfacing these errors. I'm having some trouble trying to reproduce the issue.

r-dev-limited:
I can see same error with emulator

ianitsky:
I received the following error:



[debug] [2022-07-15T15:27:12.832Z] <<< [apiv2][body] PATCH https://cloudfunctions.googleapis.com/v1/projects/*/locations/us-central1/functions/* {"error":{"code":400,"message":"The request has errors","status":"INVALID_ARGUMENT","details":[{"@type":"type.googleapis.com/google.rpc.BadRequest","fieldViolations":[{"field":"vpc_connector","description":"A Cloud Function and a Serverless VPC Access connector must be located in the same region."}]}]}}

But the function and the VPC connector are on the same region.

The upload works fine on version 11.1.0

bkendall:
@ianitsky could you open a separate issue? This is going to be tracking the `"CEL evaluation"` issue, which I don't think you're experiencing. Thanks

arthurgubaidullin:
`firebase-tools` `v11.3.0` also affected. `v11.1.0` not affected.

micro-luke:
I see this issue as well - for me it only occurs when I include a "task" function in the deployment (just started experimenting with it). If I comment out the task function everything is deploying.
firebase-tools - 11.3.0
firebase-admin - 10.3.0

radvansky-tomas:
Same error if you try to run emulator suite

```
[...]

PR #4739, opened 2022-07-13
'gracefully' handle cases when undefined is passed into resolveX

Berlioz:
Speculative: this fix addresses the failure case in https://github.com/firebase/firebase-tools/issues/4735 where `undefined` is somehow getting passed into resolveInt or resolveBoolean at runtime in violation of the typescript annotations. While impossible to verify in light of the absence of a repro, we know that the corresponding functions looked like this in v12.2.1:

```
function resolveInt(from: number | Expression<number> | null): number {	
  if (from == null) {	
    return 0;	
  } else if (typeof from === "string") {	
    throw new FirebaseError("CEL evaluation of expression '" + from + "' not yet supported");	
  }	
  return from;	
}
```

meaning that plumbing `undefined` through the function is highly likely to reverse the regression.

However, this doesn't address the more fundamental problem that we don't _know where undefined is creeping in_ yet, which makes me cautious of a band-aid solution like this.

### DIFF
--- a/src/deploy/functions/params.ts
+++ b/src/deploy/functions/params.ts
@@ -30,7 +30,7 @@ export function resolveInt(
   from: number | build.Expression<number>,
   paramValues: Record<string, build.Field<string | number | boolean>>
 ): number {
-  if (typeof from === "number") {
+  if (typeof from === "number" || typeof from === "undefined") {
     return from;
   }
   const match = /\A{{ params\.(\w+) }}\z/.exec(from);
@@ -58,7 +58,7 @@ export function resolveString(
   from: string | build.Expression<string>,
   paramValues: Record<string, build.Field<string | number | boolean>>
 ): string {
-  if (!isCEL(from)) {
+  if (!isCEL(from) || typeof from === "undefined") {
     return from;
   }
   let output = from;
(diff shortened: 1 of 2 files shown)


src/deploy/functions/params.ts
```diff
@@ -30,7 +30,7 @@ export function resolveInt(
   from: number | build.Expression<number>,
   paramValues: Record<string, build.Field<string | number | boolean>>
 ): number {
-  if (typeof from === "number") {
+  if (typeof from === "number" || typeof from === "undefined") {
```
bkendall, line 33: Consider just doing `from === undefined`, which _should_ give you the same behavior (but seem a little less clunky/extra-stepy)

inlined:
Let's work on getting that repro. I'm in the middle of some heavy surgery on build.ts that fixes typing so that we worry about undefined and null correctly (warning... template magic incoming).

I'd like to test the repro against the heavier handed fix if possible.

src/deploy/functions/params.ts
inlined: `undefined` is a variable, not a constant. JavaScript best practices is always to compare `typeof`.

inlined: Whelp, my overly pedantic coding style is based on an outdated concern (and one which would almost certainly not happen w/o a malicious NPM module). ICIMIY: `undefined` is technically a variable, not a constant. So in the past you could assign `undefined = 5` and things that had a `typeof (x) === "undefined"` would not `== undefined`.

When I tried to reproduce this insanity though it seems like Node fixed this by _still allowing_ assignment to undefined, but making it not do anything.

inlined (commented):
Catching up on all my notifications in FIFO and ran across this pretty early. Is this still of value after my PR or were we racing to fix the same issues?

inlined:
Ping? Am I to review this still?

Berlioz:
The type safety changes that went in last week likely fix this problem in a less, uh, ham-fisted way.

Issue #2837, opened 2020-11-18
Functions emulator does not honor timeoutSeconds


xaphod:
### [REQUIRED] Environment info

**firebase-tools:** 8.16.2

**Platform:** macOS 10.15.7

### [REQUIRED] Test case

```
exports.zip = functions
  .runWith({ memory: '1GB', timeoutSeconds: 540 })
  .https
  .onRequest(zip);
```

### [REQUIRED] Steps to reproduce

Run function on emulator such that it takes a long time to execute.

### [REQUIRED] Expected behavior

I expect that the function times out in 540 seconds.

### [REQUIRED] Actual behavior

The function times out in about 60 seconds (I did not time it, that is a guess).

Is it a known issue that the emulator doesn't respect `timeoutSeconds`, or did I do it wrong?

samtstern:
@xaphod as far as I can tell timeout is working.  I wrote these two functions:

```js
const functions = require('firebase-functions');

exports.withTimeout = functions.runWith({
  timeoutSeconds: 30
}).https.onRequest((request, response) => {
  setTimeout(() => {
    response.send("Hello from Firebase!");
  }, 45 * 1000);
});

exports.withoutTimeout = functions.runWith({
  memory: '1GB'
}).https.onRequest((request, response) => {
  setTimeout(() => {
    response.send("Hello from Firebase!");
  }, 45 * 1000);
});
```

When I run the one with the 30s timeout:
```
i  functions: Beginning execution of "withTimeout"
⚠  functions: Your function timed out after ~30s. To configure this timeout, see
      https://firebase.google.com/docs/functions/manage-functions#set_timeout_and_memory_allocation.
>  /Users/samstern/Projects/firebase-tools/lib/emulator/functionsEmulatorRuntime.js:668
>                  throw new Error("Function timed out.");
```

When I run the one without:
```
i  functions: Beginning execution of "withoutTimeout"
i  functions: Finished "withoutTimeout" in ~44s
```

samtstern:
I also made sure it's not about going over 60s:
```js
exports.veryLongTimeout = functions.runWith({
  timeoutSeconds: 120
}).https.onRequest((request, response) => {
  setTimeout(() => {
    response.send("Hello from Firebase!");
  }, 100 * 1000);
});
```

Logs:
```
i  functions: Beginning execution of "veryLongTimeout"
i  functions: Finished "veryLongTimeout" in ~99s
```

xaphod:
[...]

PR #4745, opened 2022-07-14
Fix "bug" where function timeout option did not apply to Functions Emulator.

taeold:
Several months ago, we started to refactor the Functions Emulator with the goal of the Emulator implement more closely resembling how GCF works.

One way the Emulator and GCF differed was how timeout was enforced - in GCF that work was done outside of the runtime process whereas in the emulated environment it was enforced by the runtime process itself.

Our work on redesigning the emulator still has much more work to do. It was premature of me to remove a well working feature - again, hubris got the best of me ("I'll add the feature back in just a week or two!".

Many users have spoken out - so I'm proposing here we bring back the timeout feature I ripped out few months ago. It'll be our responsibility to preserve this feature in the redesign-ed the Emulator.

Fixes https://github.com/firebase/firebase-tools/issues/2837

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Fix bug where function timeout couldn't be configured in the Functions Emulator. (#4745)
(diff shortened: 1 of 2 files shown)


colerogers (approved):

absemetov:
When I use cli v11.* (11.4.0) my functions not work Error("Function timed out.").
v 10.* latest 10.9.2 work good
[10.9.2](https://github.com/firebase/firebase-tools/issues/2837#issuecomment-1198990122)

PR #4750, opened 2022-07-15
Rewrite functions emulator integration test to avoid using serialized trigger hack

taeold:
Functions Emulator integration tests depended on a unique feature called "serialized triggers" that allows the emulated function process to dynamically load new function definition on per-testcase basis.

Mechanically, the integration test target serializes/stringifies the test trigger definition and sends it over to the runtime process, and the runtime process `eval()`s it to load the function definition.

This is pretty convenient and allow the integration test to re-use single process for all test cases.

However, it gets in the way of simplifying the runtime environment. So I rewrote all integration tests to instead write out the test trigger definition to `index.js` and have the runtime load it from source - just as it does in non-test scenarios.

As expected, the exec time for emulator integration test jumped from ~2min -> 3min (we are now spawning and killing a worker process per test case instead of re-using one). I think this is a reasonable tradeoff.

The test coverage remains exactly the same - I just rewrote how the integration tests are setup and executed. I like the result  too - especially the runtime integration test which is no longer strongly coupled to how the Functions Emulator is implemented. This will be helpful later when we (maybe) break out runtime implementation to be outside the CLI.

### DIFF
--- a/.github/workflows/node-test.yml
+++ b/.github/workflows/node-test.yml
@@ -50,7 +50,7 @@ jobs:
 
       - run: npm i -g npm@8.7
       - run: npm ci
-      - run: npm test
+      - run: npm test -- -- --forbid-only
 
       - uses: codecov/codecov-action@v3
         if: matrix.node-version == '14'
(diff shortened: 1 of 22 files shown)


bkendall (approved):
LGTM on everything but the *.spec.ts files (which I assume are slightly rewrites and I didn't look that closely). Someone on the CF3 team should probably 👍 the tests themselves.

colerogers (approved):
🥵 loooong files, but all the test cases look mostly the same to me, lgtm

PR #4759, opened 2022-07-20
sets groundwork for removing login with token

bkendall:
### Description

Lays groundwork for fully deprecating token support, in favor of service accounts.

### Scenarios Tested

- Auth with `GOOGLE_APPLICATION_CREDENTIALS`
- `--token` still working (with warnings)
- `FIREBASE_TOKEN` still working (with warnings)
- `login:ci` still working (with warnings)

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,11 +174,11 @@ Use `firebase:deploy --only remoteconfig` to update and publish a project's Fire
 
 The Firebase CLI can use one of four authentication methods listed in descending priority:
 
-- **User Token** - **DEPRECATED: this will be removed in a future major version of `firebase-tools`; prefer using a Service Account** - provide an explicit long-lived Firebase user token generated from `firebase login:ci`. Note that these tokens are extremely sensitive long-lived credentials and are not the right option for most cases. Consider using service account authorization instead. The token can be set in one of two ways:
+- **User Token** - **DEPRECATED: this authentication method will be removed in a future major version of `firebase-tools`; use a service account to authenticate instead** - provide an explicit long-lived Firebase user token generated from `firebase login:ci`. Note that these tokens are extremely sensitive long-lived credentials and are not the right option for most cases. Consider using service account authorization instead. The token can be set in one of two ways:
   - Set the `--token` flag on any command, for example `firebase --token="<token>" projects:list`.
   - Set the `FIREBASE_TOKEN` environment variable.
 - **Local Login** - run `firebase login` to log in to the CLI directly as yourself. The CLI will cache an authorized user credential on your machine.
-- **Service Account** - set the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to point to the path of a JSON service account key file. See Cloud's [Getting started with authentication](https://cloud.google.com/docs/authentication/getting-started) page for more details.
+- **Service Account** - set the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to point to the path of a JSON service account key file. For more details, see Google Cloud's [Getting started with authentication](https://cloud.google.com/docs/authentication/getting-started) guide.
 - **Application Default Credentials** - if you use the `gcloud` CLI and log in with `gcloud auth application-default login`, the Firebase CLI will use them if none of the above credentials are present.
 
 ### Multiple Accounts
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -174,11 +174,11 @@ Use `firebase:deploy --only remoteconfig` to update and publish a project's Fire
...
+- **User Token** - **DEPRECATED: this will be removed in a future major version of `firebase-tools`; prefer using a Service Account** - provide an explicit long-lived Firebase user token generated from `firebase login:ci`. Note that these tokens are extremely sensitive long-lived credentials and are not the right option for most cases. Consider using service account authorization instead. The token can be set in one of two ways:
   - Set the `--token` flag on any command, for example `firebase --token="<token>" projects:list`.
   - Set the `FIREBASE_TOKEN` environment variable.
 - **Local Login** - run `firebase login` to log in to the CLI directly as yourself. The CLI will cache an authorized user credential on your machine.
-- **Service Account** - set the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to point to the path of a JSON service account key file.
+- **Service Account** - set the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to point to the path of a JSON service account key file. See Cloud's [Getting started with authentication](https://cloud.google.com/docs/authentication/getting-started) page for more details.
```
rachelsaunders, line 181:
```suggestion
- **Service Account** - set the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to point to the path of a JSON service account key file. For more details, see Google Cloud's [Getting started with authentication](https://cloud.google.com/docs/authentication/getting-started) guide.
```

src/requireAuth.ts
```diff
@@ -53,8 +53,16 @@ export async function requireAuth(options: any): Promise<void> {
   let tokenOpt = utils.getInheritedOption(options, "token");
   if (tokenOpt) {
     logger.debug("> authorizing via --token option");
+    utils.logWarning(
+      "Authenticating with `--token` is deprecated and will be removed in a future major version of `firebase-tools`. " +
+        "Prefer using a service account key with GOOGLE_APPLICATION_CREDENTIALS: https://cloud.google.com/docs/authentication/getting-started"
```
rachelsaunders, line 58:
```suggestion
        "Instead, use a service account key with `GOOGLE_APPLICATION_CREDENTIALS`: https://cloud.google.com/docs/authentication/getting-started"
```

rachelsaunders (changes requested):

README.md
```diff
@@ -224,21 +224,14 @@ or `HTTP_PROXY` value in your environment to the URL of your proxy (e.g.
...
 compatible with CI and other headless environments.
 
-1. On a machine with a browser, install the Firebase CLI.
-2. Run `firebase login:ci` to log in and print out a new [refresh token](https://developers.google.com/identity/protocols/OAuth2)
-   (the current CLI session will not be affected).
-3. Store the output token in a secure but accessible way in your CI system.
+Complete the following steps to run Firebase commands in a CI environment. Find detailed instructions for each step in Google Cloud's [Getting started with authentication](https://cloud.google.com/docs/authentication/getting-started) guide.
 
-There are two ways to use this token when running Firebase commands:
+ 1. Create a service account and grant it the appropriate level of access to your project.
```
rachelsaunders, line 229:
```suggestion
1. Create a service account and grant it the appropriate level of access to your project.
```

rachelsaunders (approved):

src/index.ts
```diff
@@ -14,7 +14,10 @@ program.option(
 );
 program.option("--account <email>", "the Google account to use for authorization");
 program.option("-j, --json", "output JSON instead of text, also triggers non-interactive mode");
-program.option("--token <token>", "supply an auth token for this command");
+program.option(
+  "--token <token>",
+  "DEPRECATED - will be removed in a future major version - supply an auth token for this command"
+);
```
mbleigh, line 20: Technically `--token` accepts both refresh tokens and access tokens. I'd like us to stop accepting refresh tokens (because they are long-lived user-attached bearer tokens), but I'm not 100% sure I want to stop accepting access tokens (since they are short-lived).

Maybe we should instrument how often `--token` gets used with access tokens to help decide.

mbleigh (approved):

src/index.ts
bkendall: I'm not sure I could tell the difference between the two, but... maybe. IMO, I think SAs are cleaner - and if we're not giving a way to generate or show that token (it would have to be yoinking the CLI token or using gcloud?), I'm disinclined to keep the flag around for (what I would guess is) a small number of use cases.

bkendall:
Thank you @rachelsaunders and @mbleigh

steveoh:
Wait, will this make firebase tools incompatible with google identity federation authentication to firebase? @bkendall You say using a SA is cleaner, I assume that means the json key from a SA, which for us requires rotation etc where federation is much more elegant.

What is the path forward for CI with firebase tools and deployments?

We're currently investigating why function deployments have stopped working recently and came across this. While I don't think it's the reason it brought up some questions.


```
Service account must be of the form 'service-account@' or 'service-account@{project-id}.iam.gserviceaccount.com'

Functions deploy had errors with the following functions:
	maps(us-central1)

Unable to set the invoker for the IAM policy on the following functions:
	maps(us-central1)

Some common causes of this:

- You may not have the roles/functions.admin IAM role. Note that roles/functions.developer does not allow you to change IAM policies.

- An organization policy that restricts Network Access on your project.
```

mbleigh:
@steveoh can you describe the specific way you're using `--token` now? The thing we don't want to support anymore is using long-lived user-based bearer credentials (OAuth refresh tokens) as a parameter for authentication. We don't want to inadvertently block other valid use cases, so I'd like to understand yours more.

steveoh:
We don't explicitly use `--token` as far as I can tell. Our CI uses https://github.com/google-github-actions/auth. Since we use the `create_credentials_file: true`, `GOOGLE_APPLICATION_CREDENTIALS` is set and hopefully used by the firebase tools. 

```
      - name: 🗝️ Authenticate to Google Cloud
        id: auth
        uses: google-github-actions/auth@v0
        with:
          access_token_scopes: 'email, openid, https://www.googleapis.com/auth/cloud-platform, https://www.googleapis.com/auth/firebase'
          workload_identity_provider: ${{ secrets.IDENTITY_PROVIDER }}
          service_account: ${{ secrets.SERVICE_ACCOUNT_EMAIL }}
          create_credentials_file: true
```

mbleigh:
That workflow should be totally supported on an ongoing basis, and if it broke it would be a bug we'd fix 👍

bkendall:
TIL about that action - that's pretty neat! Thanks for flagging this use case for us to keep an eye on

steveoh:
https://github.com/FirebaseExtended/action-hosting-deploy is another popular one but it doesn't have support for workload federation when I last used it so we switched to firebase tools directly. It would be great to have supported github actions for firebase ;)

Issue #4470, opened 2022-04-23
ENAMETOOLONG when uploading to Storage Emulator


starleaf1:
### [REQUIRED] Environment info

**firebase-tools:**
10.7.1

**Platform:**
Ubuntu 20.04.4

### [REQUIRED] Test case

I have a React app set to run on Firebase Storage Emulator. I tried uploading to Storage and an error occurred. Debug log reveals that it's an `ENAMETOOLONG` error.

### [REQUIRED] Steps to reproduce

  1. Create a Firebase web project
  2. Set up Storage emulator and hook-up the app
  3. Upload a file with `uploadBytes` into Storage Emulator

### [REQUIRED] Expected behavior

File gets uploaded and shows up in Emulator UI.

### [REQUIRED] Actual behavior

Firebase emulator crashes, showing message `An unexpected error has occurred.` in console

[firebase-debug.log](https://github.com/firebase/firebase-tools/files/8559162/firebase-debug.log)

joehan:
@tohhsinpei Could you take a look at this issue? I see that you've handled some Storage Emulator issues in the past. If you're the wrong person, could you reasssign?

starleaf1:
I included the wrong log
[firebase-debug.log](https://github.com/firebase/firebase-tools/files/8559159/firebase-debug.log)
.

justmorenoise:
Same issue on macOs 12.3.1
Firebase 9.6.6
Firebase Tools 10.8.0

```
(node:93295) UnhandledPromiseRejectionWarning: Error: ENAMETOOLONG: name too long, open '/var/folders/s4/xdtcj1yx4n34g1l_s5nlzp9m0000gn/T/firebase/storage/blobs/70f39d01-140e-456f-aac4-8cde0738144a_b_comicsflow.appspot.com_o_user_projects%252F2UpPaUVOEtHeawJAGyAEOJUI6qlY%252Fprojects%252Fab71b588-df19-4d8d-8684-c8ba612b38c8%252Fpages%252F2258b808-0f92-4471-a264-84b3ceb37e1f%252Factivities%252Fb9e9bb92-c008-4b63-9696-07dfcda952e7'
    at Object.openSync (fs.js:498:3)
    at Object.writeFileSync (fs.js:1524:35)
    at Object.appendFileSync (fs.js:1563:6)
    at Persistence.appendBytes (/usr/local/lib/node_modules/firebase-tools/lib/emulator/storage/persistence.js:25:16)
[...]

PR #4771, opened 2022-07-21
fixed ENAMETOLONG storage emulator issue 

Yuangwang:
### Description

The storage emulator has a bug with error "ENAMETOOLONG". This is caused by some file names being too long for the os. This PR adds an internal mapping from filepath -> unique file identifier so now all file names will have a fixed length when stored on disk.
https://github.com/firebase/firebase-tools/issues/4470

### Scenarios Tested

Ran the storage integration tests and added test to check for the case where a file name is extremely long


src/emulator/storage/persistence.ts
```diff
@@ -74,10 +82,19 @@ export class Persistence {
   }
 
   renameFile(oldName: string, newName: string): void {
+    this._diskPathMap.set(newName, uuid.v4());
     renameSync(this.getDiskPath(oldName), this.getDiskPath(newName));
```
tonyjhuang, line 86: do you need to move the file? it seems like just updating the pointer in diskPathMap is sufficient since we're essentially implementing a virtual file system pointer now

src/emulator/storage/persistence.ts
```diff
@@ -2,12 +2,19 @@ import { openSync, closeSync, readSync, unlinkSync, renameSync, mkdirSync } from
...
 
-/** Helper for disk I/O operations. */
+/**
+ * Helper for disk I/O operations.
+ * Assigns a unique identifier to each file and stores it on disk based on that identifier
+ */
 export class Persistence {
   private _dirPath!: string;
+  // Mapping from emulator filePaths to unique identifiers on disk
+  private _diskPathMap: Map<string, string>;
```
tonyjhuang, line 14: I'm a little concerned that we're not consistent in how we're using maintaining `diskPathMap`, i.e. deleting a file on disk doesn't remove the key from the map, any method using `getDiskPath` can either use the map or default to reading directly from the disk, etc.

If we're introducing virtual file pointers, can we stay consistent and avoid ever reading the absolute file path passed into these methods?

src/emulator/storage/persistence.ts
```diff
@@ -74,10 +82,19 @@ export class Persistence {
...
 
   getDiskPath(fileName: string): string {
+    const shortenedDiskPath = this._diskPathMap.get(fileName);
+    if (shortenedDiskPath !== undefined) {
+      return path.join(this._dirPath, encodeURIComponent(shortenedDiskPath));
+    }
     return path.join(this._dirPath, encodeURIComponent(fileName));
   }
+
+  addDiskPathMapping(fileName: string, diskPath: string): void {
```
tonyjhuang, line 97: I see this disk path map as an implementation detail of this class. Is there a way we can avoid exposing this to consumers of this class?

src/test/emulators/storage/README.md
```diff
@@ -0,0 +1,43 @@
+# Firebase Storage emulator
```
tonyjhuang, line 1: thanks for the README! could you submit this in a separate PR? I don't think it's related to this name issue

scripts/storage-emulator-integration/rules/manager.test.ts
```diff
@@ -63,8 +65,7 @@ describe("Storage Rules Manager", function () {
     // Write rules to file
     const fileName = "storage.rules";
     const testDir = createTmpDir("storage-files");
-    const persistence = new Persistence(testDir);
```
tonyjhuang, line 66: why can't we use Persistence anymore?

src/emulator/storage/files.ts
```diff
@@ -585,18 +587,22 @@ export class StorageLayer {
...
+      const filepath = this.path(metadata.bucket, fileName);
+
+      const decodedBlobPath = decodeURIComponent(blobPath);
       const blobDiskPath = this._persistence.getDiskPath(decodedBlobPath);
 
       const file = new StoredFile(metadata, blobDiskPath);
-      this._files.set(decodedBlobPath, file);
+      this._files.set(filepath, file);
 
+      this._persistence.addDiskPathMapping(filepath, decodedBlobPath);
```
tonyjhuang, line 605: If I'm understanding this correctly, new files added to the emulator are stored on disk as uuids, imported files maintain their existing formatting. IMO let's coalesce all files, existing or new, to use the new randomly generated ID system as opposed to maintaining both. 

This will also have the added benefit of not needing to expose `addDiskPathMapping` from `Persistence`

src/emulator/storage/persistence.ts
Yuangwang: You bring up some good points, I'll be a bit more careful in staying consistent and make sure the virtual file system is the only one in use

src/emulator/storage/files.ts
Yuangwang: Right now the imported files are copied into the same unique id format as the new files created in the emulator, but I needed the addDiskPathMapping to add the created ids into the map. 
I get what you're saying though and I see why we wouldn't want to expose addDiskPathMapping so I'll try to bake the mapping logic into the persistence layer itself to maintain a cleaner interface.

src/emulator/storage/persistence.ts
```diff
@@ -22,6 +29,7 @@ export class Persistence {
   }
 
   appendBytes(fileName: string, bytes: Buffer): string {
+    this._diskPathMap.set(fileName, uuid.v4());
```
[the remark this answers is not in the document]
Yuangwang: I originally thought that appendBytes was used only as a create new file method. I looked at it again after seeing your comment and I see that it's used in resumable upload to append bytes to existing file. I'll add an extra check here to make sure its only generating the uuid when a new file is made.

src/emulator/storage/persistence.ts
Yuangwang: Yep good catch! I'll simplify this

src/emulator/storage/persistence.ts
Yuangwang: Will remove this, found a workaround so I don't need to expose it

src/emulator/storage/persistence.ts
```diff
@@ -67,17 +81,25 @@ export class Persistence {
...
           resolve();
         }
       });
     });
   }
 
   renameFile(oldName: string, newName: string): void {
-    renameSync(this.getDiskPath(oldName), this.getDiskPath(newName));
+    const oldNameId = this._diskPathMap.get(oldName)!;
+    this._diskPathMap.set(newName, oldNameId);
```
tonyjhuang, line 93: do you want to delete the old file path key from `diskPathMap`?

src/emulator/storage/persistence.ts
```diff
@@ -67,17 +81,25 @@ export class Persistence {
...
+          this._diskPathMap = new Map();
           resolve();
         }
       });
     });
   }
 
   renameFile(oldName: string, newName: string): void {
-    renameSync(this.getDiskPath(oldName), this.getDiskPath(newName));
+    const oldNameId = this._diskPathMap.get(oldName)!;
```
tonyjhuang, line 92: This is actually a little bit more nuanced than I thought it was. The way our upload scheme works, is that in-progress uploads get stored to [tmp staging directory](https://github.com/firebase/firebase-tools/blob/master/src/emulator/storage/upload.ts#L113), so we need to actually move the file to the proper emulator directory here.

That being said, because in progress uploads are not being tracked by `Files._files`, we don't technically need a staging directory since only tracked files are exported. 

In short, could you remove the staging directory logic in another PR (with tests) so that we can maintain the logic of changing the virtual pointer here (instead of actually moving the file on disk)?

src/emulator/storage/persistence.ts
```diff
@@ -67,17 +81,25 @@ export class Persistence {
...
   }
 
   getDiskPath(fileName: string): string {
-    return path.join(this._dirPath, encodeURIComponent(fileName));
+    const shortenedDiskPath = this._diskPathMap.get(fileName)!;
+    return path.join(this._dirPath, encodeURIComponent(shortenedDiskPath));
+  }
+
+  copyFromExternalPath(sourcePath: string, newName: string): void {
+    this._diskPathMap.set(newName, uuid.v4());
```
tonyjhuang, line 102: small nit, can you move these instances of `uuid.v4()` to a helper method

tonyjhuang (changes requested):

src/emulator/storage/persistence.ts
Yuangwang: From what I could tell `Upload.path` actually just a filename is in the format of `${uploadId}_b_${bucketId}_o_${objectId}` and isn't actually pointing to a different directory. 

In all cases where `Upload.path` is used it is going through the Persistence layer and bing mapped to a virtual pointer too. So generally even during the staging step, files should be stored in the final emulator directory and there shouldn't be a staging directory. 

I think the map looks like 
`{uploadId_b_bucketId_o_objectId: dirPath/UUID}`
during staging and should look like `{bucketId/objectId: dirPath/UUID}` after, which this code should be able to handle. Ultimately I'm not sure if the suggested changes are needed since there is no temp staging directory.

Please let me know if I'm missing something and there are changes that need to be made here

tonyjhuang:
Btw could you also update CHANGELOG.md to force a version bump, thanks!

src/emulator/storage/persistence.ts
tonyjhuang: synced offline, you're right I was misremembering the current state of the code and spoke before I verified my statements. There's no separate staging dir so this pattern still works as is

src/emulator/storage/files.ts
```diff
@@ -539,9 +539,11 @@ export class StorageLayer {
     const metadataDirPath = path.join(storageExportPath, "metadata");
     await fse.ensureDir(metadataDirPath);
 
-    for await (const [p, file] of this._files.entries()) {
-      const metadataExportPath = path.join(metadataDirPath, encodeURIComponent(p)) + ".json";
-
+    for await (const [, file] of this._files.entries()) {
+      // get filename from file path, needed to make sure the metadata and blob file have the same filename
+      const diskFileName = file.path.replace(/^.*[\\\/]/, "");
```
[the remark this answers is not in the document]
Yuangwang: I'm able to consolidate the two fields but I still need some reference to the hashed name to store so the metadata and the blob files can have the same. 
I'm thinking of exposing a get method for `Persistence.diskPathMap` to allow for this without the regex

src/emulator/storage/files.ts
```diff
@@ -539,9 +527,13 @@ export class StorageLayer {
     const metadataDirPath = path.join(storageExportPath, "metadata");
     await fse.ensureDir(metadataDirPath);
 
-    for await (const [p, file] of this._files.entries()) {
-      const metadataExportPath = path.join(metadataDirPath, encodeURIComponent(p)) + ".json";
-
+    for await (const [, file] of this._files.entries()) {
```
tonyjhuang, line 530: do we want to move the copy on L524 to this loop as well? Seems like we can avoid copying untracked files if we do

src/emulator/storage/files.ts
```diff
@@ -585,19 +577,18 @@ export class StorageLayer {
...
 
-      const blobDiskPath = this._persistence.getDiskPath(decodedBlobPath);
-
-      const file = new StoredFile(metadata, blobDiskPath);
-      this._files.set(decodedBlobPath, file);
+      const filepath = this.path(metadata.bucket, fileName);
 
-      fse.copyFileSync(blobAbsPath, blobDiskPath);
+      this._persistence.copyFromExternalPath(blobAbsPath, filepath);
+      const file = new StoredFile(metadata);
```
tonyjhuang, line 590: nit: inline

src/emulator/storage/persistence.ts
```diff
@@ -54,6 +57,7 @@ export class Persistence {
   deleteFile(fileName: string, failSilently = false): void {
     try {
       unlinkSync(this.getDiskPath(fileName));
+      this._diskPathMap.delete(fileName);
```
tonyjhuang, line 60: tiny nit: please only keep lines that can actually throw inside `try` blocks so it's clear what errors are being caught

src/emulator/storage/persistence.ts
```diff
@@ -67,17 +71,34 @@ export class Persistence {
...
-    return path.join(this._dirPath, encodeURIComponent(fileName));
+    const shortenedDiskPath = this.getDiskFileName(fileName);
+    return path.join(this._dirPath, encodeURIComponent(shortenedDiskPath));
+  }
+
+  getDiskFileName(fileName: string): string {
+    return this._diskPathMap.get(fileName)!;
+  }
+
+  copyFromExternalPath(sourcePath: string, newName: string): void {
```
tonyjhuang, line 96: test?

tonyjhuang (commented):
Mostly LGTM, some minor nits remain

src/emulator/storage/persistence.ts
Yuangwang: Will add a unit test for the new methods

src/emulator/storage/files.ts
Yuangwang: Yeah I think that makes a lot of sense willdo

src/emulator/storage/files.ts
```diff
@@ -607,8 +608,7 @@ export class StorageLayer {
 
 /** Returns file separator used in given path, either '\\' or '/'. */
 function getPathSep(decodedPath: string): string {
-  // Suffices to check first separator, which occurs immediately after bucket name.
-  // Bucket naming guidelines: https://cloud.google.com/storage/docs/naming-buckets
-  const firstSepIndex = decodedPath.search(/[^a-z0-9-_.]/g);
+  // Checks for the first matching file separator
+  const firstSepIndex = decodedPath.search(/[\/|\\\\]/g);
```
tonyjhuang, line 612: this is such a crazy looking regex :P

tonyjhuang (approved):

src/emulator/storage/files.ts
Yuangwang: yeah haha lots of escaping characters

PR #4781, opened 2022-07-25
use user-project quota to test IAM permissions

bkendall:
### Description

`x-goog-quota-user` is the [system parameter](https://cloud.google.com/apis/docs/system-parameters) that can be used to charge a specific user the quota for a call. Since the CLI does `testIamPermissions` quite a bit, the CLI's project uses quite a bit of this quota when we really should be 'charging' the user for it. Adding the header will do that.

I also cleaned up a few types and tiny bits of logic (no more lodash!) in the `iam.ts` file, but nothing drastic!

### Scenarios Tested

Hosting and  Functions deploys.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,2 @@
-- Fix bug where function timeout couldn't be configured in the Functions Emulator. (#4745)
+- Fixes bug where function timeout couldn't be configured in the Functions Emulator. (#4745)
+- Adjusts API call to test IAM permissions to use the users' project's quota.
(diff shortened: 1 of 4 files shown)


joehan (approved):
LGTM, tho we should also run through a few commands using a project number and a project alias. I took a look through where testIamPermissions is used and i don't _think_ those cases will break, but seems like its worth testing.

bkendall:
Confirmed a project alias worked! Huzzah. Project number is still an identifier, so should work. And, as far as I know, if it's not valid, it's ignored, so... fairly safe.

Issue #4756, opened 2022-07-19
firebase auth:emulators have different json data additionalUserInfo profile at production


rymesaint:
### [REQUIRED] Environment info

**firebase-tools:** 11.3.0

**Platform:** macOS

### [REQUIRED] Test case

Login using emulator
Login using firebase auth production

### [REQUIRED] Steps to reproduce

- Login via Google Account

- Get Return AdditionalUserInfo > Profile (JSON)

### [REQUIRED] Expected behavior
```
{at_hash: g-F0nYiggy_hbu8HhTKlPw, exp: 1658240233, azp: 143834371075-h8jqsht30aa13mtdct3tv0q5pp9tf4tn.apps.googleusercontent.com, nonce: GskHkFD0RDzqRhqreHGwocGdppgJP0tszXM34KFpQ44, picture: https://lh3.googleusercontent.com/a-/AFdZucqc4iha2GZgr-FD3Oy0MwZGwLwQGCAqbXW5BuASUw=s96-c, locale: en, iss: https://accounts.google.com, email_verified: true, sub: 104779650814885939520, aud: 143834371075-h8jqsht30aa13mtdct3tv0q5pp9tf4tn.apps.googleusercontent.com, family_name: Fauzan, iat: 1658236633, email: rymetutor@gmail.com, name: Ridwan Fauzan, given_name: Ridwan}
```

### [REQUIRED] Actual behavior

```
{family_name: Fauzan, locale: en, id: 104779650814885939520, granted_scopes: openid https://www.googleapis.com/auth/userinfo.profile https://www.googleapis.com/auth/userinfo.email, verified_email: true, email: rymetutor@gmail.com, picture: https://lh3.googleusercontent.com/a-/AFdZucqc4iha2GZgr-FD3Oy0MwZGwLwQGCAqbXW5BuASUw=s96-c, given_name: Ridwan, name: Ridwan Fauzan}
```

lisajian:
Hey there @rymesaint! Wanted to clarify that some of the claims that are missing are intentionally ignored by the Auth Emulator, e.g. `exp`, `iat`, `at_hash` - for more information about ignored fields, check out the inline documentation of this interface: https://github.com/firebase/firebase-tools/blob/ae788e76a8ea45e1c967bd6e37078a991c89dc6b/src/emulator/auth/operations.ts#L3308 

[...]

PR #4783, opened 2022-07-26
Updates Google rawUserInfo to be closer to production

lisajian:
### Description

Fixes https://github.com/firebase/firebase-tools/issues/4756. Changes `id` -> `sub` and `verified_email` -> `email_verified` for Google provider

### Scenarios Tested

`npm run test` passes

### Sample Commands

N/A

### DIFF
--- a/src/emulator/auth/operations.ts
+++ b/src/emulator/auth/operations.ts
@@ -2627,11 +2627,11 @@ function fakeFetchUserInfoFromIdp(
       response.lastName = claims.family_name;
       response.rawUserInfo = JSON.stringify({
         granted_scopes,
-        id: rawId,
+        sub: rawId,
         name: displayName,
         given_name: claims.given_name,
         family_name: claims.family_name,
-        verified_email: emailVerified,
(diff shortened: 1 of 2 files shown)


yuchenshi (changes requested):
Let's hold off on this and discuss internally. I believe the Google OIDC endpoints can return either `email_verified` or `verified_email` depending on the endpoint being called -- search for both terms here: https://github.com/googleapis/google-api-go-client/blob/main/oauth2/v1/oauth2-api.json

This may be related to whether ID token or access token is being used during sign-in.

lisajian:
> Let's hold off on this and discuss internally. I believe the Google OIDC endpoints can return either `email_verified` or `verified_email` depending on the endpoint being called -- search for both terms here: https://github.com/googleapis/google-api-go-client/blob/main/oauth2/v1/oauth2-api.json
> 
> This may be related to whether ID token or access token is being used during sign-in.

Ahh interesting, sounds good, will hold off for now, thanks for pointing that out!

PR #4786, opened 2022-07-26
v1alpha1 wire format needs to support Field types

Berlioz:
The overarching problem we need to solve here is that v1alpha1's parser worked by converting the wire format YAML into an object of type Manifest, which is defined using types imported from Backend (i.e backend.ServiceConfiguration and backend.Triggered), which naturally do not support CEL expression types in its various fields. Therefore, the type checking in `assertKeyTypes` would break any attempts by the SDK to actually send Fields to the CLI.

### DIFF
--- a/src/deploy/functions/build.ts
+++ b/src/deploy/functions/build.ts
@@ -14,9 +14,8 @@ export interface Build {
   params: params.Param[];
 }
 
-/* A utility function that returns an empty Build. */
 /**
- *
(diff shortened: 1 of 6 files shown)


Berlioz:
This PR hasn't been reviewed yet, and the merge with #4791 is sufficiently nightmarish that I was working on it in a separate branch. Closing out this PR and re-opening with the new branch.

PR #4787, opened 2022-07-26
Persist param values to disk on flag enable; pull int and boolean params from disk

Berlioz:
Going at the Big Build Followup PR and hacking off separately reviewable chunks to make the eventual nightmare PR more readable.


src/deploy/functions/build.ts
```diff
@@ -224,12 +224,36 @@ export async function resolveBackend(
   const projectId = userEnvOpt.projectId;
   let paramValues: Record<string, Field<string | number | boolean>> = {};
   if (previews.functionsparams) {
-    paramValues = await params.resolveParams(build.params, projectId, userEnvs);
+    paramValues = await params.resolveParams(build.params, projectId, envWithTypes(userEnvs));
+  }
+  if (previews.parampersistence) {
+    const toWrite: Record<string, string> = {};
+    for (const paramName of Object.keys(paramValues).filter((pn) => !userEnvs.hasOwnProperty(pn))) {
```
taeold, line 231: nit* Since we are already have a for loop, I think writing out the filter condition make the code more readable (to me):

```
for (const paramName of Object.keys(paramValues)) {
  if (userEnvs.hasOwnProperty(paramName)) {
     continue;
  }
   toWrite[paramName] = paramValues[paramName]?.toString() || "";
}
```

taeold (dismissed):
https://github.com/firebase/firebase-tools/pull/4721#issuecomment-1206638634

I think we are sorta kinda converging on having `project-id` environment variable file be the MOST specific .env file, and given that it might make senes for params env var values on `.env.projectId` always. WDYT?

taeold (commented):
woops sorry - just meant to comment.

src/deploy/functions/build.ts
```diff
@@ -224,12 +224,36 @@ export async function resolveBackend(
   const projectId = userEnvOpt.projectId;
   let paramValues: Record<string, Field<string | number | boolean>> = {};
   if (previews.functionsparams) {
-    paramValues = await params.resolveParams(build.params, projectId, userEnvs);
+    paramValues = await params.resolveParams(build.params, projectId, envWithTypes(userEnvs));
+  }
+  if (previews.parampersistence) {
```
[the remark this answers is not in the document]
Berlioz: Fair enough. Done.

src/deploy/functions/build.ts
Berlioz: Yeah, I'd agree. Fixed.

src/deploy/functions/build.ts
```diff
@@ -224,12 +224,36 @@ export async function resolveBackend(
...
   }
 
   return toBackend(build, paramValues);
 }
 
+function envWithTypes(rawEnvs: Record<string, string>): Record<string, string | number | boolean> {
+  const out: Record<string, string | number | boolean> = {};
+  for (const envName of Object.keys(rawEnvs)) {
+    const value = rawEnvs[envName];
+    if (!isNaN(+value)) {
```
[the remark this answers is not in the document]
Berlioz: By ~~copying something from stack overflow~~ looking it up, i think !isNan + isFInite + doesn't contain the character e is good enough here, and I think it's more readable/dependable than relying on a regex

src/deploy/functions/build.ts
```diff
@@ -224,12 +224,39 @@ export async function resolveBackend(
...
+      }
+      toWrite[paramName] = paramValues[paramName]?.toString() || "";
+    }
+    writeUserEnvs(toWrite, userEnvOpt);
   }
 
   return toBackend(build, paramValues);
 }
 
+function envWithTypes(rawEnvs: Record<string, string>): Record<string, string | number | boolean> {
```
inlined, line 243: I'm nervous that casting this into a type now will cause problems. I worry that a customer has some reason to use "true" as a string. On first deploy it'll succeed because they'll be prompted for a string and provide `"true"`. But when it comes back, it'll come back as `true` and fail to update.

src/deploy/functions/build.ts
```diff
@@ -224,12 +224,39 @@ export async function resolveBackend(
...
+  const out: Record<string, string | number | boolean> = {};
+  for (const envName of Object.keys(rawEnvs)) {
+    const value = rawEnvs[envName];
+    if (!isNaN(+value) && isFinite(+value) && !value.includes("e")) {
```
inlined, line 247: I assume this is intended to support ints instead of floats (I thought we supported both parameter types). Technically "1e3" is an int but would fail this test. I think you might just want to check for +value being betwee than `Number.MIN_SAFE_INTEGER` and `Number.MAX_SAFE_INTEGER`

inlined (approved):
LGTM, though I'm very nervous about casting a string into a value based on what it looks like, not whether it looks like the type that's expected of it (e.g. 1e3 looks like a float but can be cast to an int or a string)

src/deploy/functions/build.ts
Berlioz: Yeah, I'm worried about that too. I've been working on a branch where ParamValue is now a struct which contains both a raw actual value and some information about what types it can be applied to, but that system also interacts with Secret params so it doesn't make much sense to try and fit it into this branch. Call it a fast-follow sort of thing?

src/deploy/functions/build.ts
Berlioz: The hilarious thing is that the reason it's like this in the first place is that my first swing at this _did_ allow 1e3 and daniel had some reservations about it--I've come around to his position personally because we expect to support params in multiple languages and in most languages that aren't javascript, I don't think you can write int literals like that.

src/deploy/functions/build.ts
```diff
@@ -311,6 +312,67 @@ function envWithTypes(rawEnvs: Record<string, string>): Record<string, string |
   return out;
 }
 
+// Utility class to make it more fluent to use proto.convertIfPresent
+// The class usese const lambdas so it doesn't loose the this context when
+// passing Resolver.resolveFoo as a proto.convertIfPresent arg.
+// The class also recognizes that if the input is not null the output cannot be
+// null.
+class Resolver {
```
inlined, line 320: Is this a copy + paste or a move of what I wrote? Do we need the definition twice?

Issue #4679, opened 2022-06-24
v2 Cloud Function deploy doesn't respect `cpu` config


mqp:
### [REQUIRED] Environment info

**firebase-tools:** 11.1.0

**Platform:** Arch Linux

### [REQUIRED] Test case

Define a function like this:

```typescript
import { onRequest } from 'firebase-functions/v2/https'

const opts = { cpu: 2 }
export const test = onRequest(opts, async (_req, res) => {
  res.status(200).json({})
})
```

### [REQUIRED] Steps to reproduce

Deploy the function.

`firebase deploy --only functions:test`

Observe on the dashboard that it is created with a CPU of only 1.

To get the CPU value to be respected, I had to fix `firebase-functions` to pass through the `cpu` field from the options into the trigger annotation, and then I had to fix `firebase-tools` to copy the `cpu` field from the annotation into the endpoint spec. Then it worked.

### [REQUIRED] Expected behavior

The configured CPU is applied, as documented in https://firebase.google.com/docs/functions/beta/manage-functions.

### [REQUIRED] Actual behavior

The configured CPU is not applied.

inlined:
Confirmed fixed in the attached pull request. This should be fixed in our next minor CLI version, which we're working on as our top priority.

PR #4791, opened 2022-07-28
Improve typing and optional propagation in functions codebase

inlined:
This is from me picking back up an old attempt to improve the typing in proto.*ifPresent so we don't need to use the `as` keyword anymore. It turns out that improving typing exposed a rather haphazard patchwork of null handling. 

There's a couple of ways we could have better handled null values. The way that this change opts to implement reset-to-null is that we send the null in the server request and restore the backend API's default values. The other option, which I might have chosen were I to do this again, is to set defaults wherever is necessary during the `prepare` phase. I think we should have a serious discussion during work hours whether we should follow-up by removing nullish annotations on the GCP types and the backend type.

Along the way I made a few well-intentioned changes that may or may not have been welcome in such a large diff:
1. To help me increase the density of boilerplate in build.ts I added my own resolver type to fit tersely in common patterns (map to resolve or resolve a list of fields)
2. To further improve readability and type safety of build.ts, I pulled the trigger on an old TODO and made all our logical types hold `number` seconds rather than `proto.Duration` (string).
3. I tried renaming `serviceAccountEmail` to `serviceAccount` since I don't actually believe this is a whole email where it is called `email`. OFC I made sure we accept either `serviceAccountEmail` and `serviceAccount` in the container contract. We can forever except the old one but (if we ever document the protocol) would document the terser one.

This PR fixes compiler errors and unit tests. We should still do a lot more testing and some more work. For example, I tried resetting memory, min, and max instances for a v1 & v2 function. In all cases I had to cast `null as any` to reset a field. Min/max instances worked correctly, but after too long debugging I noticed that the firebase-functions harness doesn't actually forward null memory amounts (something we should add as a unit test there). Additionally, I noticed that the test "detectFromPort > passes a smoke test" in discovery/index.spec.ts causes mocha to take about 20 seconds to tear down the process even after tests report they successfully run.

Fixes https://github.com/firebase/firebase-tools/issues/4679, https://github.com/firebase/firebase-tools/issues/4735


src/gcp/cloudfunctionsv2.ts
```diff
@@ -454,10 +457,15 @@ export function functionFromEndpoint(endpoint: backend.Endpoint, source: Storage
...
-    "serviceAccountEmail",
     "ingressSettings",
     "timeoutSeconds"
   );
+  proto.renameIfPresent(
+    gcfFunction.serviceConfig,
+    endpoint,
+    "serviceAccountEmail",
+    "serviceAccount"
+  );
```
taeold, line 468: similarly here - I don't think endpoint.vpc is resettable as written.

src/deploy/functions/backend.ts
```diff
@@ -169,9 +169,14 @@ export const AllIngressSettings: IngressSettings[] = [
   "ALLOW_INTERNAL_AND_GCLB",
 ];
 export type MemoryOptions = 128 | 256 | 512 | 1024 | 2048 | 4096 | 8192 | 16384 | 32768;
-export const AllMemoryOptions: MemoryOptions[] = [
-  128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768,
-];
+const allMemoryOptions: MemoryOptions[] = [128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768];
+
+/**
+ *
```
Berlioz, line 175: Seems like eslint "helpfully" added this blank comment

src/gcp/cloudscheduler.ts
```diff
@@ -52,22 +52,25 @@ export interface Job {
...
+    maxRetryDuration?: string | null;
+    minBackoffDuration?: string | null;
+    maxBackoffDuration?: string | null;
+    maxDoublings?: number | null;
   };
 }
 
+/**
+ *
+ */
```
taeold, line 73:
```suggestion
```

src/deploy/functions/params.ts
```diff
@@ -232,7 +232,10 @@ export async function resolveParams(
 ): Promise<Record<string, ParamValue>> {
   const paramValues: Record<string, ParamValue> = {};
 
-  for (const param of params.filter((param) => userEnvs.hasOwnProperty(param.param))) {
+  const [resolved, outstanding] = partition(params, (param) => {
```
Berlioz, line 235: nit: my instinct is that "resolved" might not be the right variable name here given that them already being in the .env files is more like "doesn't need to be resolved"--maybe something like `fromEnvs` and `toResolve`, off the top of my head?

src/emulator/functionsEmulatorShared.ts
```diff
@@ -125,6 +115,9 @@ export class EmulatedTrigger {
   }
 }
 
+/**
+ *
```
Berlioz, line 119: another loose eslint comment

src/emulator/functionsEmulatorShared.ts
```diff
@@ -265,6 +254,9 @@ export function getEmulatedTriggersFromDefinitions(
   );
 }
 
+/**
+ *
```
Berlioz, line 258: tons of these in this file

Berlioz:
I don't think I can give approval on this PR as a whole, but I've gone through the bits intimately connected to what I'm working on (params.ts/backend.ts/build.ts and the v1alpha1 parser) and this seems generally sensible. I'm not looking forward to the merge conflict, but still.

I recall you showed me a pretty scary looking metaprogramming based refactor of resolveInt/String/Boolean in params.ts earlier which doesn't seem to have made it into the final PR--I assume this is intentional?

taeold (commented):
First pass. Will take another look soon.

src/gcp/cloudfunctions.ts
```diff
@@ -107,31 +111,31 @@ export interface CloudFunction {
   entryPoint: string;
   runtime: runtimes.Runtime;
   // Seconds. Default = 60
```
taeold, line 113:
```suggestion
  // Default = 60s
```

src/gcp/cloudfunctions.ts
```diff
@@ -583,8 +589,16 @@ export function functionFromEndpoint(
     dockerRegistry: "ARTIFACT_REGISTRY",
   };
 
-  proto.copyIfPresent(gcfFunction, endpoint, "labels");
+  // N.B. It has the same effect to set labels to the empty object as it does to
+  // set it to null, except the former is more effective for adding automatic
+  // lables for thigns like deployment-callable
```
taeold, line 594:
```suggestion
  // lables for things like deployment-callable
```

src/test/deploy/functions/runtimes/discovery/v1alpha1.spec.ts
```diff
@@ -20,6 +20,226 @@ async function resolveBackend(bd: build.Build): Promise<backend.Backend> {
 }
 
 describe("buildFromV1Alpha", () => {
+  describe("null handling", () => {
```
taeold, line 23: this is wonderful

src/test/deploy/functions/runtimes/discovery/v1alpha1.spec.ts
```diff
@@ -290,10 +501,10 @@ describe("buildFromV1Alpha", () => {
           ...blockingTrigger,
         },
       });
-      expect(resolveBackend(parsed)).to.eventually.deep.equal(expectedBackend);
+      await expect(resolveBackend(parsed)).to.eventually.deep.equal(expectedBackend);
```
taeold, line 504: oh man. great catch.

taeold (approved):

src/deploy/functions/prepare.ts
```diff
@@ -261,19 +261,20 @@ export function inferDetailsFromExisting(
 
     // If the instance size is set out of bounds or was previously set and is now
     // unset we still need to remember it so that the min instance price estimator
-    // is accurate.
-    if (!wantE.availableMemoryMb && haveE.availableMemoryMb) {
+    // is accurate. If, on the other hand, we have a null value for availableMemoryMb
+    // we need to keep taht null (meaning "use defaults").
```
TheIronDev, line 265: nit:
taht -> that

src/deploy/functions/runtimes/discovery/parsing.ts
```diff
@@ -50,32 +68,44 @@ export function assertKeyTypes<T extends object>(
...
+    if (value === null) {
+      if (schemaType.endsWith("?")) {
+        continue;
+      }
+      throw new FirebaseError(`Expected ${fullKey}} to be type ${schemaType}; was null`);
+    }
+    if (schemaType.endsWith("?")) {
+      schemaType = schemaType.slice(0, schemaType.length - 1);
+    }
+    if (schemaType === "string") {
```
TheIronDev, line 90: Could all of this be DRY'd up with:

```
if (typeof value !== schemaType) {
  throw new FirebaseError(`Expected ${fullKey} to be ${schemaType}; was ${typeof value}`);
  throw new FirebaseError(`Expected ${fullKey} to be type ${schemaType}; was ${typeof value}`);
}
```

colerogers (approved):
🥵 🥵 🥵

src/gcp/cloudfunctions.ts
```diff
@@ -628,20 +642,21 @@ export function functionFromEndpoint(
...
     gcfFunction,
     endpoint,
-    "timeout",
-    "timeoutSeconds",
-    proto.durationFromSeconds
+    "availableMemoryMb",
+    (mem) => mem as backend.MemoryOptions
+  );
+  proto.convertIfPresent(gcfFunction, endpoint, "timeout", "timeoutSeconds", (sec) =>
+    sec ? proto.durationFromSeconds(sec) : null
```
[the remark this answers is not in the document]
inlined: SGTM

PR #4796, opened 2022-07-28
pin a specific version of cli-color

bkendall:
### Description

Pin versions of `cli-color` and `es5-ext`. Updated versions of these libraries are unfortunately unusable by us.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "body-parser": "^1.19.0",
     "chokidar": "^3.0.2",
     "cjson": "^0.3.1",
-    "cli-color": "^2.0.2",
+    "cli-color": "2.0.1",
     "cli-table": "0.3.11",
     "commander": "^4.0.1",
     "configstore": "^5.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -94,14 +94,15 @@
...
-    "cli-color": "^2.0.2",
+    "cli-color": "2.0.1",
     "cli-table": "0.3.11",
     "commander": "^4.0.1",
     "configstore": "^5.0.1",
     "cors": "^2.8.5",
     "cross-env": "^5.1.3",
     "cross-spawn": "^7.0.1",
     "csv-parse": "^5.0.4",
+    "es5-ext": "0.10.53",
```
mbleigh, line 105: Should this be an [override](https://stackoverflow.com/questions/58930922/google-cloud-functions-custom-build-step) instead / in addition?

mbleigh (approved):

package.json
bkendall: The override functionality isn't working for me (I suspect `npm-shrinkwrap.json` has to do with it)

PR #4804, opened 2022-08-01
Added support for publishing pre-release Extension versions

apascal07:


src/extensions/extensionsHelper.ts
```diff
@@ -405,6 +411,7 @@ export async function publishExtensionVersionFromLocalSource(args: {
   rootDirectory: string;
   nonInteractive: boolean;
   force: boolean;
+  stage: "rc" | "alpha" | "beta" | "stable";
```
joehan, line 414: Can you check what the behavior is when you pass something other than these options? AFAICT, there is nothing in place rn to prevent users from passing other values.

joehan, line 414: Also, it would be good to declare this type outside of this function, since we may need to reference it elsewhere.

src/extensions/extensionsHelper.ts
```diff
@@ -423,6 +430,32 @@ export async function publishExtensionVersionFromLocalSource(args: {
   );
   validateSpec(subbedSpec);
 
+  if (args.stage !== "stable") {
```
joehan, line 433: This function probably should have had tests already, but as we add more functionality, it needs them more. Can you add a test that checks the behavior when different values are passed for stage?

joehan (changes requested):

src/extensions/extensionsHelper.ts
apascal07: Added explicit check.

src/extensions/extensionsHelper.ts
apascal07: Added tests.

joehan (approved):

PR #4805, opened 2022-08-01
add storage emulator support to init.js

tonyjhuang:
### Description

Adding support to Firebase Hosting's `init.js` to support setting the emulator host and port for the storage SDK when using the `useEmulators=true` query param


bkendall (changes requested):
Could you add a CHANGELOG entry for this? Thanks

bkendall (approved):

Issue #3823, opened 2021-10-13
Emulators start with --import fails whether storage_export/blobs or storage_export/metadata empty folders do not exist


charliejuc:
**firebase-tools: 9.20.0**

 

**Platform: Pop OS**

### [REQUIRED] Test case and Steps to reproduce

```
firebase emulators:export storeExport
```

Then after upload to git  and download the code, empty folders are not uploaded to version control: storeExport/storage_export/blobs or storeExport/storage_export/metadata

Then when you run:

```
firebase emulators:start --import storeExport
```

Previous command will fail.

### [REQUIRED] Expected behavior

Emulator should not require empty folders to work.

### [REQUIRED] Actual behavior

Fails when the above mentioned folders do not exist.

tonyjhuang:
This should be fixed by 33647

rhodgkins:
I'm not sure that #3647 will fix this - the `metadata` and `blobs` directories will always be empty if there's nothing to export. Usually for stuff like this I'd add in an empty `.gitadd` file but not sure if that's really appropriate!

tonyjhuang:
I see, my mistake. We can update the emulator to handle importing a directory structure with missing `metadata` and `blobs ` folders at some point in the future, thanks for raising the issue @charliejuc

tonyjhuang:
Filed internal bug to track, b/240637117

abhis3:
Fix is merged in, should be in the next emulator release. Closing issue.

PR #4806, opened 2022-08-01
Fix Storage import failure due to missing directories from export

abhis3:
### Description

When exporting the storage emulator, if a directory is empty we never write it to the export folder. This causes a bug when importing as these folders are missing, causing the emulator to terminate prematurely.

Fix in response to #3823 

### Scenarios Tested

Manually tested this change with local `firebase-tools` build against both populated exports and empty exports generated by the emulator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,2 @@
 - Adds check for callable functions when discovering Hosting rewrite endpoints. (#4792)
+- Fixes Storage import failure due to missing directories. (#3823)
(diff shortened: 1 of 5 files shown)


tonyjhuang (approved):
🔥

tonyjhuang (commented):
nice! test?

abhis3:
> nice! test?

I wanted to add a test but I don't believe we have any testing for import/export functionality and it'll require a bit of work to implement imo. I'd prefer to add in those tests separately and test import and export more exhaustively.

Happy to add a TODO if that works for you

bkendall:
@abhis3 would it be possible to create a test in `scripts/storage-emulator-integration/import/tests.ts` for this? I think more of the storage emulator is tested via those functions (since files.spec.ts isn't that comprehensive).

Yuangwang (approved):

tonyjhuang (approved):

Issue #4599, opened 2022-05-31
Storage emulator interprets the machine's current time as UTC without converting it


macksal:
### [REQUIRED] Environment info

**firebase-tools:** 11.0.1

**Platform:** OSX Catalina 10.15.7

### [REQUIRED] Test case

`broken.test.ts`: 

```ts
import * as rulesTest from "@firebase/rules-unit-testing";
import {equal} from "assert";
describe("Storage", function () {
    it("should populate timeCreated with approximately the current time", async () => {
        const env = await rulesTest.initializeTestEnvironment({});
        await env.clearStorage();
        await env.withSecurityRulesDisabled(async ctx => {
            const now = new Date();
            const ref = ctx.storage().ref("object1");
            await ref.put(new ArrayBuffer(1024));
            const meta = await ref.getMetadata();
            const timeCreated = new Date(meta.timeCreated);
            console.info("timeCreated: ", timeCreated);
            console.info("current time: ", now);
            // allow 5s of error either side
            equal(timeCreated.getTime() < now.getTime() + 5000, true);
            equal(timeCreated.getTime() > now.getTime() - 5000, true);
        });
        await env.cleanup();
    });
});
export {};
```

### [REQUIRED] Steps to reproduce

1. Set up a new firebase project with the storage emulator enabled.
2. Create a package for testing with these dependencies only, and the Mocha test case from above:

```json
"devDependencies": {
    "@firebase/rules-unit-testing": "^2.0.2",
    "@types/mocha": "^9.1.1",
    "firebase": "^9.8.2",
    "firebase-tools": "^11.0.1",
    "mocha": "^10.0.0",
    "ts-mocha": "^10.0.0",
    "typescript": "^4.7.2"
}
```

3. Run the test case with `yarn firebase emulators:exec --only storage 'ts-mocha broken.test.ts'`

### [REQUIRED] Expected behavior

4. The test should pass.

### [REQUIRED] Actual behavior

The test fails with the below output. I am running this at 12:38 PM in my timezone (Australia/Brisbane). 

```
timeCreated:  2022-05-31T12:38:44.572Z
current time:  2022-05-31T02:38:44.561Z
```

[...]

PR #4813, opened 2022-08-02
Change storage metadata to be UTC only

Yuangwang:
### Description

Fixes issue #4599 by changing references of date.get*() methods into date.getUTC*(). This makes sure everything is always set to UTC time in the storage metadata.
### Scenarios Tested

Added unit test for toSerializedDate to make sure that the serialized date is always set to UTC and tested manually so that the bug is fixed.


abhis3 (approved):

src/test/emulators/storage/metadata.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { expect } from "chai";
+import { toSerializedDate } from "../../../emulator/storage/metadata";
+
+describe("toSerializedDate", () => {
+  it("correctly serializes date", () => {
+    const testDate = new Date("2022-01-01T00:00:00.000Z");
```
tonyjhuang, line 6: can you init this date with a custom timezone to verify it converts to UTC properly?

tonyjhuang (approved):

abhis3:
@Yuangwang I think you need a `-` in the CHANGELOG

Yuangwang:
> @Yuangwang I think you need a `-` in the CHANGELOG

Good catch, I'll add it in

PR #4816, opened 2022-08-02
Fix schedule function deploy and delete

colerogers:
Fixes schedule function deploy & delete


src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -31,13 +37,22 @@ type EventTrigger = backend.EventTrigger & {
...
+};
+
 export type ManifestEndpoint = Base &
   backend.Triggered &
   Partial<backend.HttpsTriggered> &
   Partial<backend.CallableTriggered> &
   Partial<{ eventTrigger: EventTrigger }> &
   Partial<backend.TaskQueueTriggered> &
   Partial<backend.BlockingTriggered> &
+  Partial<{ scheduleTrigger: ScheduleTrigger }> &
```
taeold, line 55: Can we add a note on why we needed to make duplicate declartion of ScheduleTrigger (*Seconds vs *Duration)?

src/test/deploy/functions/runtimes/discovery/v1alpha1.spec.ts
```diff
@@ -307,6 +307,63 @@ describe("buildFromV1Alpha", () => {
       await expect(resolveBackend(parsed)).to.eventually.deep.equal(expectedBackend);
     });
 
+    it("copies v1 schedules", async () => {
```
taeold, line 310: Can we add 2 test cases? One that uses *Duration another that uses *Seconds?

taeold: Let's do this in a follow up PR.

taeold (approved):

src/test/deploy/functions/runtimes/discovery/v1alpha1.spec.ts
colerogers: There's already a test that uses *Seconds on L367

src/deploy/functions/runtimes/discovery/v1alpha1.ts
colerogers: Added on L40

PR #4817, opened 2022-08-02
Clean up Storage Emulator integration tests

tonyjhuang:
### Description

Cleans up set up and tear down logic for the storage emulator integration tests and dedupes a lot of common operations.

Also:
*  fixes some conformance tests (tests that point to production)
* bumps sdk versions to 9.9.1


scripts/storage-emulator-integration/tests.ts
```diff
@@ -87,41 +91,53 @@ describe("Storage emulator", () => {
...
+      await resetStorageEmulator(STORAGE_EMULATOR_HOST);
+    }
+  }
 
-      // TODO: We should not need a real credential for emulator tests, but
-      //       today we do.
-      const credential = fs.existsSync(path.join(__dirname, SERVICE_ACCOUNT_KEY))
-        ? admin.credential.cert(readJson(SERVICE_ACCOUNT_KEY))
-        : admin.credential.applicationDefault();
+  before(async function (this) {
```
Yuangwang, line 101: After making all of this setup and teardown logic common, does running individual tests in this file still work? just want to confirm

scripts/storage-emulator-integration/tests.ts
```diff
@@ -1373,87 +1232,44 @@ describe("Storage emulator", () => {
...
       });
       await page.addScriptTag({
-        url: "https://www.gstatic.com/firebasejs/7.24.0/firebase-auth.js",
+        url: "https://www.gstatic.com/firebasejs/9.9.1/firebase-auth-compat.js",
       });
       await page.addScriptTag({
-        url: TEST_CONFIG.useProductionServers
-          ? "https://www.gstatic.com/firebasejs/7.24.0/firebase-storage.js"
-          : "https://storage.googleapis.com/fir-tools-builds/firebase-storage.js",
+        url: "https://www.gstatic.com/firebasejs/9.9.1/firebase-storage-compat.js",
```
Yuangwang, line 1240: Why don't we need "https://storage.googleapis.com/fir-tools-builds/firebase-storage.js" anymore?

scripts/storage-emulator-integration/tests.ts
```diff
@@ -2111,121 +1877,171 @@ describe("Storage emulator", () => {
         });
       });
     });
+  });
 
-    emulatorSpecificDescribe("Non-SDK Endpoints", () => {
-      beforeEach(async () => {
-        await resetStorageEmulator(STORAGE_EMULATOR_HOST);
+  describe("Non-SDK Endpoints", () => {
```
Yuangwang, line 1881: These tests will run when pointed to prod servers too now right? Do they all work when pointed to prod?

scripts/storage-emulator-integration/tests.ts
tonyjhuang: yup

scripts/storage-emulator-integration/tests.ts
tonyjhuang: They *should* work when pointed to prod but many are broken. We'll need to do another round of changes to fix the conformance tests but for now I'm re-enabling these when pointing tests to prod servers to make this expectation explicit.

scripts/storage-emulator-integration/tests.ts
tonyjhuang: Not sure why this was here in the first place, it seems like was designed to support testing against custom built js sdk version but this seems undesirable from a testing perspective

scripts/storage-emulator-integration/tests.ts
Yuangwang: Do you mind adding a TODO to indicate that these don't yet work? We can remove it once we get around to fixing them later.

Yuangwang (approved):
Small comment, other than that LGTM

scripts/storage-emulator-integration/tests.ts
tonyjhuang: done

scripts/storage-emulator-integration/tests.ts
```diff
@@ -2111,121 +1877,171 @@ describe("Storage emulator", () => {
...
+      Buffer.from(IMAGE_FILE_BASE64, "base64"),
+      tmpDir
+    );
 
-        await page.evaluate(
-          (IMAGE_FILE_BASE64, filename) => {
-            const auth = (window as any).auth as firebase.auth.Auth;
+    beforeEach(async function (this) {
+      this.timeout(TEST_SETUP_TIMEOUT);
+      await resetEmulatorState();
```
abhis3, line 1893: I know it was like this before, but I don't think we need to reset the emulator state every time here right? Some tests like generating a new token with invalid values will fail regardless of the emulator state. Any way we can make it so we're not resetting every time and only when it's needed?

I think this might be one of our time sinks when running tests but I'm not 100% sure

src/test/emulators/fixtures.ts
```diff
@@ -232,36 +232,3 @@ service firebase.storage {
 `,
   },
 };
-
-/*
```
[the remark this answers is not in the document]
tonyjhuang: hmm, the delta here is me removing it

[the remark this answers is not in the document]
abhis3: Oh good call, yup false alarm

scripts/storage-emulator-integration/tests.ts
tonyjhuang: discussed offline, resetting local state is very fast, resetting prod state may take longer but is worth it for maintaining a purely hermetic testing environment

PR #4821, opened 2022-08-03
Advanced CF3 param features: regexp-validated, selector, boolean, secret, resource-list, boolean-type

Berlioz:
Major architectural change: ParamValue has changed from `string | number | boolean` to an object which contains the raw value, type information, and critically, whether the param came from CSM and therefore should never be written to disk.

Potential regression: we used to fail when reading params from dotenvs if the value didn't "look" like the right type. This current implementation lets people do stupid things with string-to-number/boolean conversion if they want to, which is...the spirit of Javascript, right?


TheIronDev:
Git is calling out a few lint errors

src/deploy/functions/params.ts
```diff
@@ -168,8 +185,108 @@ export interface SelectInput<T> {
...
+  rawValue: string;
+  isSecret: boolean;
+
+  constructor(
+    rawValue: string,
+    secret: boolean,
+    types: { string?: boolean; boolean?: boolean; number?: boolean }
+  ) {
+    this.rawValue = rawValue;
+    this.isSecret = secret;
```
TheIronDev, line 239: boolean consistency nit:

I'd recommending either omit `isSecret` or add the `is` prefix to the other three.

src/deploy/functions/params.ts
```diff
@@ -168,8 +185,108 @@ export interface SelectInput<T> {
...
+        this.legalString = false;
+        this.legalBoolean = true;
+        this.legalNumber = false;
+        break;
+      case "int":
+        this.legalString = false;
+        this.legalBoolean = false;
+        this.legalNumber = true;
+        break;
+      case "secret":
```
TheIronDev, line 263: Is fallthrough intentional? Can you add a comment to confirm if it is or not?

src/deploy/functions/params.ts
```diff
@@ -168,8 +185,108 @@ export interface SelectInput<T> {
...
+          `ParamValue.narrowType() called on a param of type ${isType}, which cannot have values from .env files.`
+        );
+    }
+  }
+
+  toString(): string {
+    return this.rawValue;
+  }
+
+  legalString: boolean;
```
TheIronDev, line 275: Nit: move these class members to the top (with docs)

src/deploy/functions/params.ts
```diff
@@ -168,8 +185,108 @@ export interface SelectInput<T> {
...
+ *   the type of the Param.
+ * - ParamValues coming from Cloud Secrets Manager will have a string type field set
+ *   and isSecret = true, telling the Build process not to write the value to .env files.
+ */
+export class ParamValue {
+  rawValue: string;
+  isSecret: boolean;
+
+  constructor(
+    rawValue: string,
```
TheIronDev, line 234: nit: consider parameter properties:

go/tsstyle#parameter-properties

src/deploy/functions/params.ts
```diff
@@ -233,32 +357,37 @@ export async function resolveParams(
...
-    }
-    paramValues[param.name] = userEnvs[param.name];
+  for (const param of resolved) {
+    const valueFromEnvs = Object.create(userEnvs[param.name]) as ParamValue;
+    valueFromEnvs.narrowType(param.type);
+    paramValues[param.name] = valueFromEnvs;
   }
 
-  if (nonInteractive && outstanding.length > 0) {
+  const [needSecret, needPrompt] = partition(outstanding, (param) => {
```
TheIronDev, line 370: Nit, remove unnecessary braces:

```
const [needSecret, needPrompt] = partition(outstanding, (param) => param.type === "secret");
```

src/deploy/functions/params.ts
```diff
@@ -143,14 +152,22 @@ export interface IntParam extends ParamBase<number> {
   input?: TextInput<number> | SelectInput<number>;
 }
 
-export interface TextInput<T, Extensions = {}> { // eslint-disable-line
+export interface BooleanParam extends ParamBase<number> {
+  type: "boolean";
+
+  // If omitted, defaults to TextInput<number>
```
taeold, line 158:
```suggestion
  // If omitted, defaults to TextInput<boolean>
```

src/deploy/functions/params.ts
```diff
@@ -168,8 +185,108 @@ export interface SelectInput<T> {
...
+  ) {
+    this.rawValue = rawValue;
+    this.isSecret = secret;
+    this.legalString = types.string || false;
+    this.legalBoolean = types.boolean || false;
+    this.legalNumber = types.number || false;
+  }
+
+  // Call this when we figure out the actual type of a ParamValue from the .envs
+  narrowType(isType: Param["type"]): void {
```
taeold, line 246: nit* `isType` as throwing me off b/c I expected it to be a boolean value per naming convention but actually is a string.

src/deploy/functions/params.ts
```diff
@@ -233,32 +357,37 @@ export async function resolveParams(
...
-        "Parameter " +
-          param.name +
-          " resolved to value from dotenv files " +
-          userEnvs[param.name] +
-          " of wrong type"
-      );
-    }
-    paramValues[param.name] = userEnvs[param.name];
+  for (const param of resolved) {
+    const valueFromEnvs = Object.create(userEnvs[param.name]) as ParamValue;
```
taeold, line 365: uh I think I need some help understanding this. ParaValue is a class, and we are casting a plain object to be a `ParamValue` object? and calling `.narrowType` on it works? Why can't we use `param` directly here?

src/deploy/functions/params.ts
```diff
@@ -267,83 +396,271 @@ export async function resolveParams(
...
+        secretParam.name
+      } and managed by Firebase Hosting (Firebase Hosting Service Agent will be granted Secret Admin role on this secret).\nEnter a value for ${
+        secretParam.label || secretParam.name
+      }:`,
+    });
+    await secretManager.createSecret(projectId, secretParam.name, {});
+    await secretManager.addVersion(projectId, secretParam.name, secretValue);
+    return secretValue;
+  } else if (!metadata.secretVersion) {
+    // the secret exists, but version "latest" doesnt, which...is questionably possible?
```
taeold, line 426: it could, but it may be disabled. we should be checking the state of the secrets in this code.

src/deploy/functions/params.ts
```diff
@@ -267,83 +396,271 @@ export async function resolveParams(
...
+  }
+
+  // we need to test if the hosting service account can actually read
```
taeold, line 429: q: why hosting service account?

src/deploy/functions/params.ts
```diff
@@ -267,83 +396,271 @@ export async function resolveParams(
...
+  // secretManager.ensureServiceAgentRole(metadata.secret, [`what address goes here`], "roles/secretmanager.admin");
+
+  return secretManager.accessSecretVersion(projectId, secretParam.name, "latest");
```
taeold, line 432: this feels like a big no no - user may not have access to secret values (and are encouraged not to have access unless strictly necessary).

I didn't think we can use secret values can ever be used inside a CEL expression - why do we try to access it here?

src/deploy/functions/params.ts
```diff
@@ -267,83 +396,271 @@ export async function resolveParams(
...
+}
+
+async function getSecretMetadata(
```
taeold, line 435: nit* This function feels like it should either live in `gcp/secret` or `functoins/secret`

src/deploy/functions/params.ts
```diff
@@ -267,83 +396,271 @@ export async function resolveParams(
...
+  if (!metadata.secret) {
+    const secretValue = await promptOnce({
+      name: secretParam.name,
+      type: "password",
+      message: `This secret will be stored in Cloud Secret Manager (https://cloud.google.com/secret-manager/pricing) as ${
+        secretParam.name
+      } and managed by Firebase Hosting (Firebase Hosting Service Agent will be granted Secret Admin role on this secret).\nEnter a value for ${
+        secretParam.label || secretParam.name
+      }:`,
+    });
```
taeold, line 421: We might want to rethink this flow (consider not implemented for the first cut?

One design goal in implementing `functions:secret:set` command was to try to avoid having users type out the secrets in command line since that may be logged - see [go/cf3-secrets](http://go/cf3-secrets) for details.

src/deploy/functions/params.ts
```diff
@@ -267,83 +396,271 @@ export async function resolveParams(
...
   } else if (param.type === "int") {
-    return promptIntParam(param, resolvedDefault as number | undefined);
+    const provided = await promptIntParam(param, resolvedDefault as number | undefined);
+    return new ParamValue(provided.toString(), false, { number: true });
+  } else if (param.type === "boolean") {
+    const provided = await promptBooleanParam(param, resolvedDefault as boolean | undefined);
+    return new ParamValue(provided.toString(), false, { boolean: true });
+  } else if (param.type === "secret") {
+    throw new FirebaseError(
+      `Somehow ended up trying to interactively prompt for secret parameter ${param.name}, which should never happen.`
```
taeold, line 483: I'm a little confused why `handleSecret` implements prompting for secret params instead of doing it here (where I'd naturally look for prompting logic).

src/deploy/functions/params.ts
```diff
@@ -267,83 +396,271 @@ export async function resolveParams(
...
+  }
+}
+
+async function promptResourceString(
+  prompt: string,
+  input: ResourceInput,
+  projectId: string,
+  resolvedDefault?: string
+): Promise<string> {
+  const notFound = new FirebaseError(`No instances of ${input.resource.type} found.`);
```
taeold, line 592: nit* this error is only used once - i don't think a var is necessary.

src/deploy/functions/params.ts
```diff
@@ -267,83 +396,271 @@ export async function resolveParams(
...
+      return promptText<string>(prompt, { type: "text" }, resolvedDefault, (res: string) => res);
+  }
+}
+
+type retryInput = { message: string };
+async function promptText<T extends RawParamValue>(
+  prompt: string,
+  input: TextInput<T>,
+  resolvedDefault: T | undefined,
+  converter: (res: string) => T | retryInput
```
taeold, line 619: I think it make sense to make this argument optional instead of having call pass identity function.

src/deploy/functions/params.ts
```diff
@@ -267,83 +396,271 @@ export async function resolveParams(
...
+          `Input did not match provided validator ${userRe.toString()}, retrying...`
+      );
+      return promptText<T>(prompt, input, resolvedDefault, converter);
+    }
   }
+  const converted = converter(res);
+  if (typeof converted === "object") {
+    logger.error(converted.message);
+    return promptText<T>(prompt, input, resolvedDefault, converter);
+  }
```
taeold, line 640: hmm this feel overlapping with validationRegex and a bit confusing to understand without looking at how this argument is actually used by the caller.

One alternative would be to define `promptNumber` or `promptBoolean` to specifically cover those cases.

(no action required if you disagree)

src/deploy/functions/params.ts
```diff
@@ -267,83 +396,271 @@ export async function resolveParams(
...
+    message: prompt,
+    choices: input.select.map((option: SelectOptions<T>): ListItem => {
+      return {
+        checked: false,
+        name: option.label,
+        value: option.value.toString(),
+      };
+    }),
+  });
+  return converter(response);
```
taeold, line 665: q: why is it safe to not worry about "conversion gone wrong" here?

src/gcp/storage.ts
```diff
@@ -229,6 +239,27 @@ export async function getBucket(bucketName: string): Promise<BucketResponse> {
   }
 }
 
+/**
+ * Gets the list of storage buckets associated with a specific project from GCP.
+ * Ref: https://cloud.google.com/storage/docs/json_api/v1/buckets/list
+ * @param {string} bucketName name of the storage bucket
+ * @return a bucket resource object
+ */
+export async function listBuckets(projectId: string): Promise<Array<string>> {
```
taeold, line 248: note: we are going to have to deal with pagination at some point. There are other implementation for dealing with pagination in this codebase.

src/deploy/functions/params.ts
inlined: Ostensibly the types were correct when the developer defined the select input, but Daniel is right. Some jerk could try an int param with a select dialog that return "foo". We might as well have an error message if the select option's value can't be coerced into the param's type. Parsing is the worst.

src/gcp/storage.ts
inlined: I have thoughts on how dynamically paginated lists could have been implemented, it's not clear to me that `prompt` has support for a dynamic list of a select prompt. This means Victor has to query for all buckets every time before he can present UX. Most cloud products only allow 1K buckets and a bucket object is probably 1-10K, so that means the worst possible case is that we make a 10MB download. I can live with that given that this user is probably overjoyed that we even work for their project.

src/test/deploy/functions/params.spec.ts
```diff
@@ -80,12 +98,12 @@ describe("resolveParams", () => {
         name: "foo",
         default: "bar",
         type: "string",
-        input: { type: "text", text: {} },
```
inlined, line 83: Isn't this the wrong field we deleted?

inlined (commented):
Holy cow this project is maturing!

src/deploy/functions/build.ts
```diff
@@ -287,33 +287,29 @@ export async function resolveBackend(
       nonInteractive
     );
 
-    // TODO(vsfan@): when merging secrets support into the Build, make sure we aren't writing those to disk.
     const toWrite: Record<string, string> = {};
     for (const paramName of Object.keys(paramValues)) {
-      if (userEnvs.hasOwnProperty(paramName)) {
+      const paramValue = paramValues[paramName];
+      if (userEnvs.hasOwnProperty(paramName) || paramValue === undefined || paramValue.isSecret) {
```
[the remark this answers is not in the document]
Berlioz: Fixed. There's no case in which it can be undefined, and tsc doesn't even complain if I remove the check entirely.

src/deploy/functions/params.ts
Berlioz: Wow, let's. Uh. Pretend I never wrote this code. Yeah.

src/deploy/functions/params.ts
Berlioz: Presumably, it will be, once we implement more resource types than just bucket.

src/deploy/functions/params.ts
Berlioz: Wouldn't that only make sense if T was `string` itself? There's no implicit way to make an arbitrary type out of a string, is there?

src/deploy/functions/params.ts
Berlioz: Done.

src/deploy/functions/params.ts
Berlioz: Done.

src/deploy/functions/params.ts
Berlioz: Done.

src/deploy/functions/params.ts
```diff
@@ -278,59 +349,45 @@ function resolveDefaultCEL(
...
-      case "int":
-        value = +rawValue;
-        break;
-      case "string":
-      default:
-        value = rawValue;
-        break;
-    }
+    const value = new ParamValue(rawValue, true, { string: true });
     paramValues[secretParam.name] = value;
```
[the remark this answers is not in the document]
Berlioz: Wow, good catch. I totally forgot to add the check I was intending on in CEl resolution:
` const dependsOnSecret = deps.some((dep) => currentEnv[dep].secret)`
and then throw the same error that you'd get if you depended on a param name that didn't exist.

Added that just now.

src/deploy/functions/params.ts
```diff
@@ -340,26 +397,27 @@ function resolveDefaultCEL(
...
  */
- async function handleSecret(secretParam: SecretParam, projectId: string): Promise<string> {
+async function handleSecret(secretParam: SecretParam, projectId: string): Promise<string> {
   const metadata = await getSecretMetadata(projectId, secretParam.name, "latest");
   if (!metadata.secret) {
     const secretValue = await promptOnce({
       name: secretParam.name,
       type: "password",
-      message: `This secret will be stored in Cloud Secret Manager (https://cloud.google.com/secret-manager/pricing) as ${secretParam.name} and managed by Firebase Hosting (Firebase Hosting Service Agent will be granted Secret Admin role on this secret).\nEnter a value for ${secretParam.label || secretParam.name}:`,
+      message: `This secret will be stored in Cloud Secret Manager (https://cloud.google.com/secret-manager/pricing) as ${
```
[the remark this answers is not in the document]
Berlioz: eslint happened to it, which is infuriating.

src/deploy/functions/params.ts
```diff
@@ -402,24 +460,41 @@ async function getSecretMetadata(
...
+async function promptBooleanParam(
+  param: BooleanParam,
+  resolvedDefault?: boolean
+): Promise<boolean> {
   if (!param.input) {
     const defaultToText: TextInput<boolean> = { type: "text" };
     param.input = defaultToText;
   }
-  const isTruthyInput = (res: string) => ['true', 'y', 'yes', '1'].includes(res.toLowerCase());
+  const isTruthyInput = (res: string) => ["true", "y", "yes", "1"].includes(res.toLowerCase());
```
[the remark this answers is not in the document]
Berlioz: That's currently inconsistent with how we handle it on the SDK side, though naturally such things could always be changed.

src/test/deploy/functions/params.spec.ts
Berlioz: Technically, it isn't--the type: field is mandatory until #4874 goes in, since I separated out the 10 SWE-hour banana into a different PR. Whichever one goes second is just going to have to fix merge conflicts, but it should be trivial.

src/deploy/functions/params.ts

src/deploy/functions/params.ts
Berlioz: It sure is, but a comment is a good idea.

src/deploy/functions/params.ts

src/deploy/functions/params.ts
Berlioz: hmm, well, `paramType` is presumably pretty inoffensive....

src/deploy/functions/params.ts
```diff
@@ -168,8 +185,81 @@ export interface SelectInput<T> {
...
+  legalBoolean: boolean;
+  // Whether this param value can be sensibly interpreted as a number
+  legalNumber: boolean;
+
+  constructor(
+    private readonly rawValue: string,
+    readonly secret: boolean,
+    types: { string?: boolean; boolean?: boolean; number?: boolean }
+  ) {
+    this.rawValue = rawValue;
```
TheIronDev, line 240: Assigning `rawValue``secret` in constructor body are both redundant (see https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties)

TheIronDev (commented):
Lgtm!

src/deploy/functions/params.ts
taeold: ah ha this is where I slip out the "YAGNI" card, but I think that's not productive comment on a month old PR. sgtm.

src/deploy/functions/params.ts
```diff
@@ -267,83 +366,298 @@ export async function resolveParams(
...
+ */
+async function handleSecret(secretParam: SecretParam, projectId: string): Promise<string> {
+  const metadata = await getSecretMetadata(projectId, secretParam.name, "latest");
+  if (!metadata.secret) {
+    const secretValue = await promptOnce({
+      name: secretParam.name,
+      type: "password",
+      message: `This secret will be stored in Cloud Secret Manager (https://cloud.google.com/secret-manager/pricing) as ${
+        secretParam.name
+      } and managed by Firebase Hosting (Firebase Hosting Service Agent will be granted Secret Admin role on this secret).\nEnter a value for ${
```
taeold, line 388: I don't think this is true - especially the Hosting SA part. We always use end user credentials in Firebase CLI don't we?

src/deploy/functions/params.ts
```diff
@@ -267,83 +366,298 @@ export async function resolveParams(
...
+        secretParam.label || secretParam.name
+      } is in illegal state ${metadata.secretVersion.state}`
+    );
+  }
+
+  secretManager.ensureServiceAgentRole(
+    metadata.secret,
+    [`${projectId}@appspot.gserviceaccount.com`],
+    "roles/secretmanager.admin"
+  );
```
taeold, line 415: I'm not sure why we need to do this. Firebase CLI always use end-user credentials to operate on Cloud resources unlike Extensions.

src/deploy/functions/params.ts
taeold: I still think this 🤔

src/deploy/functions/params.ts
taeold: Still not sure - is there reason why secretParam given super special treatment?

taeold (changes requested):

src/deploy/functions/params.ts
taeold: oh good point. Didn't consider that aspect.

src/deploy/functions/params.ts
Berlioz: That's correct. Removed.

src/deploy/functions/params.ts
Berlioz: Yup.

Berlioz:
Okay, talked things over with @taeold and it turns out that a lot of the complexity around secrets, intended to give parity with extensions, just doesn't make sense in a CF3 world. Cutting back the handling of secret params to basically just be a validation step now.

src/deploy/functions/params.ts
taeold: still feel this way, not a blocking comment though.

taeold (approved):

PR #4823, opened 2022-08-03
Fix bug where invalid CPU was set for 16GiB functions.

taeold:
Today, Firebase CLI fails to deploy cf3v2 function w/ 16GiB and no cpu setting:

```js
export const fn = onCall({ memory: "16GiB"}, () => "hello");
```

```
$ firebase deploy

Error: The following functions have invalid CPU settings fn. Valid CPU options are (0.08, 1], 2, 4, 6, 8, or "gcf_gen1"
```

Found that we are mapping 16GiB => 3 CPU which then fails to validate. Proposing that we map the CPU setting to 2 CPU instead.

### DIFF
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -228,7 +228,7 @@ export function memoryToGen2Cpu(memory: MemoryOptions): number {
     2048: 1,
     4096: 2,
     8192: 2,
-    16384: 3,
+    16384: 2,
     32768: 4,
   }[memory];
 }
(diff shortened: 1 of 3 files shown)


colerogers (approved):

src/deploy/functions/backend.ts
```diff
@@ -228,7 +228,7 @@ export function memoryToGen2Cpu(memory: MemoryOptions): number {
     2048: 1,
     4096: 2,
     8192: 2,
-    16384: 3,
+    16384: 2,
```
inlined, line 231: Should have been 4 according to a quick survey where we create a 16GB function in the Cloud Console

inlined (approved):

PR #4824, opened 2022-08-03
Make storage emulator content type parser case insensitive

Yuangwang:
### Description
Make storage emulator content type parser case insensitive, fixes part of #3953

### Scenarios Tested
Manual testing with the setup described in #3953 and also added unit test for this specific case


tonyjhuang (approved):
thanks for the fix!

PR #4828, opened 2022-08-03
Populate resource correctly in storage rules evaluation

abhis3:
### Description

Fix an error (#4329) where resource was not populate correctly during rules evaluation.

### Scenarios Tested

Tested locally against local emulator codebase before and after the change. The repro'd error is now no longer present.


scripts/storage-emulator-integration/tests.ts
```diff
@@ -1,15 +1,15 @@
+import { Bucket, CopyOptions } from "@google-cloud/storage";
```
abhis3, line 1: This is from some auto-formatting that I ran

Yuangwang (approved):
LGTM

PR #4829, opened 2022-08-04
Extend function deploy integration tests to cover more scenarios.

taeold:
Function deploy test will now deploy and check the result of:

1. Task queue functions
2. Functions with secrets
3. V1 scheduled with retry configs

While adding these test cases, I found out that neither scheduled functions or task queue functions preserves existing configuration on updates. I made changes to scheduled function in this PR to fix that b/c it was easy. Fixing the scenario for TQ function will be a little more involved, so I simply disabled the test case for now. Let's revisit in subsequent PRs.


scripts/functions-deploy-tests/tests.ts
```diff
@@ -6,12 +6,17 @@ import * as functions from "firebase-functions";
...
+import * as scheduler from "../../src/gcp/cloudscheduler";
 import { Endpoint } from "../../src/deploy/functions/backend";
+import { getGlobalDefaultAccount } from "../../src/auth";
+import { setRefreshToken } from "../../src/apiv2";
 
 const FIREBASE_PROJECT = process.env.GCLOUD_PROJECT || "";
 const FIREBASE_DEBUG = process.env.FIREBASE_DEBUG || "";
 const FUNCTIONS_DIR = path.join(__dirname, "functions");
-const FNS_COUNT = 12;
+const FNS_COUNT = 16;
```
Berlioz, line 19: this is a completely unrelated request, but how hard would it be to set up something which let you run only a specific test case that was failing?

src/gcp/cloudscheduler.ts
```diff
@@ -159,17 +167,28 @@ export async function createOrReplaceJob(job: Job): Promise<any> {
...
-function isIdentical(job: Job, otherJob: Job): boolean {
-  return (
-    job &&
-    otherJob &&
-    job.schedule === otherJob.schedule &&
-    job.timeZone === otherJob.timeZone &&
-    _.isEqual(job.retryConfig, otherJob.retryConfig)
-  );
+function needUpdate(existingJob: Job, newJob: Job): boolean {
+  if (!existingJob) {
```
Berlioz, line 174: was it intentional that we're now only checking whether one of the parameters is truthy?

Berlioz (approved):

scripts/functions-deploy-tests/tests.ts
taeold: That's a good idea. I think that requires larger change on how this test case currently runs its test - feel free to throw your idea in if you have any. I'll also see if there is a nice way of doing that 🤔

src/gcp/cloudscheduler.ts
taeold: I thought it was redundant! But since it's safer to check for newJob too, I'll reinsert the logic.

scripts/functions-deploy-tests/run.sh
```diff
@@ -4,6 +4,9 @@ set -e # Immediately exit on failure
 # Globally link the CLI for the testing framework
 ./scripts/npm-link.sh
 
+# Create a secret for testing if it doesn't exist
+firebase functions:secrets:get TOP --project $GCLOUD_PROJECT || (echo secret | firebase functions:secrets:set --data-file=- TOP --project $GCLOUD_PROJECT -f)
+
 (cd scripts/functions-deploy-tests/functions && npm i)
 
-mocha scripts/functions-deploy-tests/tests.ts
+mocha scripts/functions-deploy-tests/tests.ts
```
inlined, line 12: Aren't you missing a newline?

scripts/functions-deploy-tests/tests.ts
```diff
@@ -286,31 +386,34 @@ describe("firebase deploy", function (this) {
     for (const e of Object.values(endpoints)) {
       expect(e).to.include({
         availableMemoryMb: 128,
-        // TODO: Fix bug where timeout is being updated, not inferred from existing.
-        // timeoutSeconds: 42,
-        maxInstances: 42,
+        timeoutSeconds: 60,
```
inlined, line 389: This changed from a non-standard to a default value. Is that intentional?

inlined (commented):
LGTM

PR #4830, opened 2022-08-04
Optimize Storage Emulator integration test runs

tonyjhuang:
### Description

Shares a single puppeteer page across all test cases within "Firebase Endpoints". This reduces our test run time from ~45s to <15s for the whole file, but we will have to make sure we don't save any state to the page to avoid breaking hermetic state.


scripts/storage-emulator-integration/tests.ts
```diff
@@ -1254,25 +1234,31 @@ describe("Storage emulator", () => {
...
+    afterEach(async function (this) {
+      this.timeout(EMULATORS_SHUTDOWN_DELAY_MS);
+      await page.evaluate(async () => {
+        await firebase.auth().signOut();
+      });
     });
 
-    after(async () => {
+    after(async function (this) {
+      this.timeout(EMULATORS_SHUTDOWN_DELAY_MS);
```
abhis3, line 1253: We set this in line 104, I think we can remove this

abhis3: Unless we want the same timeout applied to Firebase Endpoints and the entire test suite?

abhis3: Moving to other PR

abhis3 (approved):
LGTM

PR #4831, opened 2022-08-04
Fix bug where CPU mapping for v2 16GiB function was incorrect.

taeold:
Should be 4 CPU instead of 2 CPU according to what happens on GCF deploys.


taeold:
Duplicate of https://github.com/firebase/firebase-tools/pull/4845

PR #4832, opened 2022-08-04
Read name from metadata as backup for gcs upload

### LINKED ISSUE #3953: Storage Emulator does not match Google Cloud Storage library API

Yuangwang:
### Description

Uses metadata.name as a backup to query.name in gcs uploads, this makes the storage emulator compliant with gcs api spec, fixes #3953 
### Scenarios Tested
Tested via the python gcs sdk and also by adding two new integration tests here.

### DIFF
--- a/scripts/storage-emulator-integration/tests.ts
+++ b/scripts/storage-emulator-integration/tests.ts
@@ -239,9 +239,7 @@ describe("Storage emulator", () => {
           expect(uploadURL.searchParams?.get("name")).to.equal(fileName);
         });
 
-        it.only("should handle multipart upload with name only in metadata", async () => {
-          const contentTypeHeader =
-            "multipart/related; boundary=b1d5b2e3-1845-4338-9400-6ac07ce53c1e";
+        it("should handle multipart upload with name only in metadata", async () => {
           const body = Buffer.from(`--b1d5b2e3-1845-4338-9400-6ac07ce53c1e\r
 content-type: application/json\r
 \r
@@ -259,7 +257,7 @@ hello there!
             .send(body)
             .set({
               Authorization: "Bearer owner",
-              "content-type": contentTypeHeader,
+              "content-type": "multipart/related; boundary=b1d5b2e3-1845-4338-9400-6ac07ce53c1e",
             })
             .expect(200)
             .then((res) => res.body.name);
--- a/scripts/storage-emulator-integration/tests.ts
+++ b/scripts/storage-emulator-integration/tests.ts
@@ -246,6 +246,46 @@ describe("Storage emulator", () => {
           expect(fileMetadata).to.deep.include(metadata);
         });
 
+        it("should handle resumable upload with name only in metadata", async () => {
+          const fileName = "test_upload.jpg";
+          const uploadURL = await supertest(STORAGE_EMULATOR_HOST)
+            .post(`/upload/storage/v1/b/${storageBucket}/o?uploadType=resumable`)
+            .send({ name: fileName })
+            .set({
+              Authorization: "Bearer owner",
+            })
+            .expect(200)
+            .then((res) => new URL(res.header["location"]));
+          expect(uploadURL.searchParams?.get("name")).to.equal(fileName);
+        });
+
+        it.only("should handle multipart upload with name only in metadata", async () => {
+          const contentTypeHeader =
+            "multipart/related; boundary=b1d5b2e3-1845-4338-9400-6ac07ce53c1e";
+          const body = Buffer.from(`--b1d5b2e3-1845-4338-9400-6ac07ce53c1e\r
+content-type: application/json\r
+\r
+{"name":"test_upload.jpg"}\r
+--b1d5b2e3-1845-4338-9400-6ac07ce53c1e\r
+content-type: text/plain\r
+\r
+hello there!
+\r
+--b1d5b2e3-1845-4338-9400-6ac07ce53c1e--\r
+`);
+          const fileName = "test_upload.jpg";
+          const responseName = await supertest(STORAGE_EMULATOR_HOST)
+            .post(`/upload/storage/v1/b/${storageBucket}/o?uploadType=multipart`)
+            .send(body)
+            .set({
+              Authorization: "Bearer owner",
+              "content-type": contentTypeHeader,
+            })
+            .expect(200)
+            .then((res) => res.body.name);
+          expect(responseName).to.equal(fileName);
+        });
+
         it("should return an error message when uploading a file with invalid metadata", async () => {
           const fileName = "test_upload.jpg";
           const errorMessage = await supertest(STORAGE_EMULATOR_HOST)
--- a/scripts/storage-emulator-integration/tests.ts
+++ b/scripts/storage-emulator-integration/tests.ts
@@ -225,10 +225,14 @@ describe("Storage emulator", () => {
 
           expect(fileMetadata).to.deep.include(metadata);
         });
-
+        // TODO(b/241151246): Fix conformance tests.
         it("should handle resumable upload with name only in metadata", async () => {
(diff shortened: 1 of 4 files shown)


scripts/storage-emulator-integration/tests.ts
```diff
@@ -226,6 +226,46 @@ describe("Storage emulator", () => {
...
+            .send({ name: fileName })
+            .set({
+              Authorization: "Bearer owner",
+            })
+            .expect(200)
+            .then((res) => new URL(res.header["location"]));
+          expect(uploadURL.searchParams?.get("name")).to.equal(fileName);
+        });
+
+        it.only("should handle multipart upload with name only in metadata", async () => {
```
tonyjhuang, line 242: remove `only`

scripts/storage-emulator-integration/tests.ts
```diff
@@ -226,6 +226,46 @@ describe("Storage emulator", () => {
...
+          const contentTypeHeader =
```
tonyjhuang, line 243: inline?

src/emulator/storage/apis/gcloud.ts
```diff
@@ -219,6 +209,11 @@ export function createCloudEndpoints(emulator: StorageEmulator): Router {
       if (emulatorInfo === undefined) {
         return res.sendStatus(500);
       }
+      const name = getValidName(req.query, req.body);
+      if (name == null) {
```
tonyjhuang, line 213: go/js-style#features-equality-checks, please use ===

src/emulator/storage/apis/gcloud.ts
```diff
@@ -237,11 +232,17 @@ export function createCloudEndpoints(emulator: StorageEmulator): Router {
...
     let metadataRaw: string;
     let dataRaw: Buffer;
+    let name: string | undefined;
     try {
       ({ metadataRaw, dataRaw } = parseObjectUploadMultipartRequest(
         contentTypeHeader!,
         await reqBodyToBuffer(req)
       ));
+      name = getValidName(req.query, JSON.parse(metadataRaw));
+      if (name == null) {
```
tonyjhuang, line 242: would this ever be true? name is `string | undefined`, I don't see null as an option above

src/emulator/storage/apis/gcloud.ts
Yuangwang: The goal is to catch null and undefined with this check I think its an exception to the rule, go/js-style#features-equality-checks-exceptions

scripts/storage-emulator-integration/tests.ts
Yuangwang: good catch!

src/emulator/storage/apis/gcloud.ts
Yuangwang: this check should catch undefined and null. I can change it to name=== undefined too, but I think this is a bit cleaner in my eyes if anyone ever introduces a null state in the future. Lmk what you think

src/emulator/storage/apis/gcloud.ts
```diff
@@ -237,11 +232,17 @@ export function createCloudEndpoints(emulator: StorageEmulator): Router {
     // Multipart upload
     let metadataRaw: string;
     let dataRaw: Buffer;
+    let name: string | undefined;
     try {
       ({ metadataRaw, dataRaw } = parseObjectUploadMultipartRequest(
         contentTypeHeader!,
         await reqBodyToBuffer(req)
       ));
+      name = getValidName(req.query, JSON.parse(metadataRaw));
```
tonyjhuang, line 241: why is this in the try catch?

src/emulator/storage/apis/gcloud.ts
tonyjhuang: you still need to declare the type on L235 as nullable. In general I would prefer we avoid any type coercion if we can in typescript since it's fraught with unexpected behavior, see https://www.youtube.com/watch?v=et8xNAc2ic8

What about just using `string | null` as opposed to `string | undefined | null` if you want to keep it to a single equality check?

src/emulator/storage/apis/gcloud.ts
tonyjhuang: See comments below, since `getValidName` returns `string | undefined`, you should only need `name === undefined` here since `getValidName` can't return null

src/emulator/storage/apis/gcloud.ts
```diff
@@ -411,3 +412,8 @@ function sendObjectNotFound(req: Request, res: Response): void {
     });
   }
 }
+
+function getValidName(query: Query, metadata: IncomingMetadata): string | undefined {
```
tonyjhuang, line 416: tiniest of nits, would also prefer a more descriptive name, go/java-practices/naming#method-name-expectations

What is a valid name? In what contexts should you use this? Something like `getIncomingFileNameFromRequest` is a little easier to parse esp if we're not adding a comment

src/emulator/storage/apis/gcloud.ts
Yuangwang: yeah thats fair, I'll swap it to name === undefined

src/emulator/storage/apis/gcloud.ts
Yuangwang: yeah thats a fair point about the type coercion, I'll swap this out to name === undefined since name can only be `string | undefined` as you mentioned

src/emulator/storage/apis/gcloud.ts
Yuangwang: Good catch, I'll move it out

src/emulator/storage/apis/gcloud.ts
Yuangwang: Makes sense, thanks for the suggestion too! I'll add it in

scripts/storage-emulator-integration/tests.ts
```diff
@@ -225,10 +225,14 @@ describe("Storage emulator", () => {
 
           expect(fileMetadata).to.deep.include(metadata);
         });
-
+        // TODO(b/241151246): Fix conformance tests.
         it("should handle resumable upload with name only in metadata", async () => {
+          const expectedHost = TEST_CONFIG.useProductionServers
```
tonyjhuang, line 230: You can probably extract this to the top level `describe` to save yourself from having to declare this in every test case

tonyjhuang (approved):
Thanks for the fix!

Issue #4780, opened 2022-07-25
 Cloud Function and a Serverless VPC Access connector must be located in the same region.


ianitsky:
### [REQUIRED] Environment info

**firebase-tools: 11.3.0

**Platform: Ubuntu

### [REQUIRED] Test case

Create a VPC and add on runWith and execute:

firebase deploy --only functions

`
const runtimeOpts: RuntimeOptions = {
    vpcConnector: functions.config().vpc_connector.name,
    vpcConnectorEgressSettings: 'PRIVATE_RANGES_ONLY'
}
`

`functions.runWith(runtimeOpts)`

### [REQUIRED] Steps to reproduce

Create a VPC and add on runWith and execute:

firebase deploy --only functions

`const runtimeOpts: RuntimeOptions = {
    vpcConnector: functions.config().vpc_connector.name,
    vpcConnectorEgressSettings: 'PRIVATE_RANGES_ONLY'
}
functions.runWith(runtimeOpts)

### [REQUIRED] Expected behavior

Must works.

### [REQUIRED] Actual behavior

Throw error on terminal and show follow error on log:

[debug] [2022-07-15T15:27:12.832Z] <<< [apiv2][body] PATCH https://cloudfunctions.googleapis.com/v1/projects/*/locations/us-central1/functions/* {"error":{"code":400,"message":"The request has errors","status":"INVALID_ARGUMENT","details":[{"@type":"type.googleapis.com/google.rpc.BadRequest","fieldViolations":[{"field":"vpc_connector","description":"A Cloud Function and a Serverless VPC Access connector must be located in the same region."}]}]}}

The upload works fine on version 11.1.0

moldy530:
I'm seeing this as well on v11.4.1 of firebase tools with the firebase-functions sdk v3.22

taeold:
Hey @ianitsky @moldy530 - thanks for taking time to report an issue.

Asking the obvious questions 

1) Can you give it another try with `v11.4.2` CLI (released yesterday). We fixed a load of issues with function configuration in the release.

2) What is the region of the function (which, I presume is the same region of the VPC). Do you happen to be deploying the function to multiple regions w/ options like `regions: ['us-central1', 'europe-west4']`?

I tried deploying a function with very simple VPC setup (function in `us-central1`, vpc connector also in `us-central1`) using the latest Functions SDK & CLI and wondering if there is a bug is hiding in more advanced setup.

coolbho3k:
[...]

PR #4834, opened 2022-08-05
Fix bug where vpc connector was incorrectly converted.

taeold:
Functions with VPC connector setting written using the old Functions SDK that still relies on parseTrigger module can't be deployed because we have a bug in transforming shorthand vpc connector e.g. `hello-vpc` into the correct long resource form e.g. `projects/my-project/location/us-central1/connectors/hello-vpc`.

Here, we consolidate the logic for expanding VPC connector name in the `backend` phase instead of the `build` phase since region info isn't available during a build phase.

Fixes https://github.com/firebase/firebase-tools/issues/4780

### DIFF
--- a/src/test/deploy/functions/runtimes/node/parseTriggers.spec.ts
+++ b/src/test/deploy/functions/runtimes/node/parseTriggers.spec.ts
@@ -368,6 +368,38 @@ describe("addResourcesToBuild", () => {
     await expect(convertedBackend).to.eventually.deep.equal(expectedBackend);
   });
 
+  it("should expand vpc connector in w/ shorthand form", async () => {
+    const trigger: parseTriggers.TriggerAnnotation = {
+      ...BASIC_TRIGGER,
+      httpsTrigger: {},
+      vpcConnector: "hello-vpc",
+    };
+
+    const result = build.empty();
+    parseTriggers.addResourcesToBuild("project", "nodejs16", trigger, result);
+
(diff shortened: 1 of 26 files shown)


src/test/deploy/functions/runtimes/node/parseTriggers.spec.ts
```diff
@@ -368,6 +368,38 @@ describe("addResourcesToBuild", () => {
...
+
+    const expectedBackend: backend.Backend = backend.of({
+      ...BASIC_BACKEND_ENDPOINT,
+      httpsTrigger: {},
+      vpc: {
+        connector: `projects/project/locations/${api.functionsDefaultRegion}/connectors/hello-vpc`,
+      },
+    });
+    const convertedBackend = resolveBackend(expected);
+    await expect(convertedBackend).to.eventually.deep.equal(expectedBackend);
```
colerogers, line 400: any reason that you are using `eventually` here instead of awaiting `resolveBackend`?

colerogers (approved):

src/test/deploy/functions/runtimes/node/parseTriggers.spec.ts
taeold: Copy-and-paste! I think I'd done what you said, but will keep this style since other test cases are also written in this style.

inlined (commented):
LGTM

Issue #1602, opened 2019-08-15
firebase.json 'functions.ignore' bloats package upload


repentsinner:
### Environment info

**firebase-tools:**
```
firebase --version
7.2.2
```

**Platform:**
Both MacOS (10.14.6 (18G87)) & Windows 10 development environments.

### Test case

MCVE: Firebase functions template + a 'functions.ignore' section in `firebase.json`

### Steps to reproduce

Create a firebase project.

Run firebase init functions with default options, and add it to your new project:
```
firebase init functions
```

Uncomment the helloWorld sample function in `index.js` and deploy:

```bash
⋊> ~/D/f/functions firebase deploy                                                                                    11:02:38

=== Deploying to 'test-function-deploy-3a957'...

i  deploying functions
i  functions: ensuring necessary APIs are enabled...
✔  functions: all necessary APIs are enabled
i  functions: preparing functions directory for uploading...
i  functions: packaged functions (22.95 KB) for uploading
✔  functions: functions folder uploaded successfully
i  functions: updating Node.js 8 function helloWorld(us-central1)...
✔  functions[helloWorld(us-central1)]: Successful update operation. 

✔  Deploy complete!

Project Console: https://console.firebase.google.com/project/test-function-deploy-3a957/overview
⋊> ~/D/f/functions
```

Note the deployed packaged functions size (22.95KB).

Edit `firebase.json` to include a 'functions.ignore' section:

```json
{
  "functions": {
    "ignore": []
  }
}
```

Deploy again:

```bash
⋊> ~/D/f/functions firebase deploy                                                                                    11:03:24

=== Deploying to 'test-function-deploy-3a957'...

i  deploying functions
i  functions: ensuring necessary APIs are enabled...
✔  functions: all necessary APIs are enabled
i  functions: preparing functions directory for uploading...
i  functions: packaged functions (9.07 MB) for uploading
✔  functions: functions folder uploaded successfully
i  functions: updating Node.js 8 function helloWorld(us-central1)...
✔  functions[helloWorld(us-central1)]: Successful update operation. 

✔  Deploy complete!

Project Console: https://console.firebase.google.com/project/test-function-deploy-3a957/overview
⋊> ~/D/f/functions 
```

Note that the deployed package has grown from 22.95**KB** to 9.07**MB**.

### Expected behavior

I expect that if things are ignored, the deployed package gets smaller rather than larger.

### Actual behavior

[...]

PR #4838, opened 2022-08-08
Explicitly add `functions.ignore` options to firebase.json

taeold:
We are hoping that this will improve discoverability of the `ignore` field while we think through how we want to document different firebase.json options for the functions project.

Partially address lack of documentation on `firebase.json` fields for functions discussed in https://github.com/firebase/firebase-tools/issues/1602#issuecomment-521761044.


src/init/features/functions/index.ts
```diff
@@ -7,6 +8,9 @@ import { previews } from "../../../previews";
 import { Options } from "../../../options";
 import { ensure } from "../../../ensureApiEnabled";
 
+/**
+ * Setup a new firebase project for functions.
```
egilmorez, line 12: Nit: as a verb, this has a space: "Set up"

egilmorez (approved):
One nit, thanks!

PR #4839, opened 2022-08-08
Set request.path variable correctly in storage emulator

tonyjhuang:
### Description

Previously, when evaluating `request.path` in security rules, the Storage Emulator was setting it to the exact path that the user was passing in, e.g.:

```
firebase.storage().ref('dir/subdir/').listAll(); // Sets request.path to 'dir/subdir'
```

This is incorrect, according to the docs: https://firebase.google.com/docs/rules/rules-language#building_conditions. The correct `request.path` should be "b/<bucket>/o/dir/subdir". This PR fixes that bug.

### Scenarios Tested

Added an integration test


abhis3 (approved):
LGTM, thanks for the fix!

Yuangwang (approved):
LGTM

Issue #4803, opened 2022-08-01
Cannot deploy V2 Cloud Functions with more than 8 GiB of memory


0xjei:
### [REQUIRED] Environment info

**firebase-tools**: *^11.4.1*

**Platform**: *Ubuntu*

### [REQUIRED] Test case

You can reproduce the error by defining a V2 Cloud Function as below:

```javascript
import { onCall } from 'firebase-functions/v2/https'

export const dummycf = onCall({
  timeoutSeconds: 3600,
  memory: "16GiB",
  cpu: 2
}, (request) => {
  console.log(`This V2 CF w/ 16GiB of memory and 2 vCPUs should work fine!`)
})
```

### [REQUIRED] Steps to reproduce

You just need to deploy the Cloud Function by running:

```bash 
firebase deploy --only functions:test
```

### [REQUIRED] Expected behavior

The Cloud Function must be correctly deployed with 4 vCPUs and 16 GiB of memory. The V2 public preview should support the memory and cpu values as reported in the [documentation](https://firebase.google.com/docs/functions/beta/manage-functions).

### [REQUIRED] Actual behavior

Running the command above results in the following error:

```bash
> firebase deploy --only functions:dummycf

=== Deploying to 'mpc-phase2-suite-test'...

i  deploying functions
Running command: npm --prefix "$RESOURCE_DIR" run build

> mpc-phase2-suite-backend@0.0.1 build
> tsc

✔  functions: Finished running predeploy script.
i  functions: ensuring required API cloudfunctions.googleapis.com is enabled...
i  functions: ensuring required API cloudbuild.googleapis.com is enabled...
i  artifactregistry: ensuring required API artifactregistry.googleapis.com is enabled...
✔  artifactregistry: required API artifactregistry.googleapis.com is enabled
✔  functions: required API cloudbuild.googleapis.com is enabled
✔  functions: required API cloudfunctions.googleapis.com is enabled
i  functions: preparing codebase default for deployment
i  functions: Loaded environment variables from .env.

Error: available memory must be a supported value, if present
```

taeold:
Hey @0xjei - thanks for reporting and apologies for the issue you are seeing.

I tried giving 16GiB function a spin and it gave my setup no problem:

```
exports.sixteen = functionsv2.https.onCall({ memory: "16GiB", cpu: 2 }, () => { return { foo: "bar"}; });
```

[...]

PR #4845, opened 2022-08-09
Fix bug where wrong CPU config was set for 16/32GiB functions

taeold:
On Cloud Run, 16 GiB memory requires at least 4 CPUs. 32 GiB memory requires at least 8 CPUs.

https://cloud.google.com/run/docs/configuring/memory-limits#cpu-minimum

Partial fix for https://github.com/firebase/firebase-tools/issues/4803


inlined (approved):
Changelog please

PR #4850, opened 2022-08-09
Fix bug where functions:secrets:get command did not call requirePermissions.

taeold:
`requirePermission` in turn calls `requireAuth` which ultimately executes to mint refresh token for the Firebase CLI to use to generate auth tokens. By not calling `requirePermission` in `functions:secrets:set` command, API calls to secret manager did not include Authorization token if not already cached locally. For CI systems, this meant that secret manager API calls always failed with 401.

Should fix broken functions integration test unless it is broken b/c other code changes.


bkendall:
Do you mean `requirePermissions`?

taeold:
@bkendall Yes, but ultimately `requireAuth` was the function call that mattered.

colerogers (approved):

PR #4851, opened 2022-08-10
Emulator analytics

yuchenshi:


bkendall (approved):

src/extensions/extensionsApi.ts
```diff
@@ -476,7 +476,7 @@ export async function listExtensionVersions(
     }>(`/publishers/${publisherId}/extensions/${extensionId}/versions`, {
       queryParams: {
         filter,
-        show_prereleases: String(show_prereleases),
+        showPrereleases: String(showPrereleases),
```
yuchenshi, line 479:
```suggestion
        show_prereleases: String(showPrereleases),
```

yuchenshi: Does the backend accept either casing of the query parameter? If not, we should revert. (I'm guessing either works if it is a One Platform API, but we may want to play safe)

PR #4852, opened 2022-08-10
[Function-NoOp] Implement hashing of function  (env + secrets + source)

TheIronDev:
### Description

Implement generating a hash for: 

* Backend environment variables
* Endpoint secret versions
* Backend's fs source

This PR does not wire up functionality, but instead exposes functionality that could be used for #4876

### Scenarios Tested

- [x] Files matching
- [x] Files not matching
- [x] Secret versions matching
- [x] Secret versions not matching
- [x] Env variables matching
- [x] Env variables not matching


src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,39 @@
+import * as fs from "fs";
+import * as crypto from "crypto";
+import * as args from "../args";
+import * as secrets from "../../../functions/secrets";
+import * as backend from "../backend";
+
+export async function getLocalHash(source: args.Source, wantBackend: backend.Backend) {
```
inlined, line 7: Make this take an exact source file and then call the function twice: once for sourceV1 and once for sourceV2. The former may contain a dynamically generated runtimeconfig.json and the latter definitely won't.

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  return hash.read().toString("hex");
+}
+
+// Hash the secret versions.
+function getSecretVersions(wantBackend: backend.Backend): Record<string, string> {
+  const endpointsById = Object.values(wantBackend?.endpoints || {});
+  const endpointsList: backend.Endpoint[] = endpointsById
+    .map((endpoints) => Object.values(endpoints))
+    .reduce((memo, endpoints) => [...memo, ...endpoints], []);
```
inlined, line 34: I think you can just replace this with `backend.allEndpoints(wantBackend)`

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,39 @@
...
+  return secrets.of(endpointsList).reduce((memo, { secret, version }) => {
+    if (version) memo[secret] = version;
```
inlined, line 36: 1. I'm pretty sure inlining an if statement c-style is against the style rules
2. Are we absolutely sure version will be present? If so we should maybe assert rather than having a hash that might make functions not deploy irrespective of secret versions changing. If version may not be present we have to handle it.

src/test/functions/cache/hash.spec.ts
```diff
@@ -0,0 +1,44 @@
+import { expect } from "chai";
+import { getLocalHash } from "../../../deploy/functions/cache/hash";
+import * as backend from "../../../deploy/functions/backend";
+import { resolve } from "path";
+
+const EMPTY_WANT_BACKEND = {
```
inlined, line 6: nit: can just use `backend.empty()`

src/test/functions/cache/hash.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+  requiredAPIs: [],
+  environmentVariables: {},
+  endpoints: {},
+} as backend.Backend;
```
inlined, line 10: nit: for variables you can just declare them to be the type you want instead of casting them.

src/test/functions/cache/hash.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+    const source2 = { functionsSourceV1: file2 };
+
+    const hash1 = await getLocalHash(source1, EMPTY_WANT_BACKEND);
+    const hash2 = await getLocalHash(source2, EMPTY_WANT_BACKEND);
+
+    expect(hash1).to.equal(hash2);
+    done();
+  });
+
+  // TODO(tystark) test case for secret versions
```
inlined, line 42: Are you adding this before checking in?

src/deploy/functions/cache/hash.ts
TheIronDev: !! You're absolutely right!

src/deploy/functions/cache/hash.ts
TheIronDev: 1. Updated!
2. As long as this is invoked at the end of the `prepare` stage then version will be present. The tentative plan is to invoke this after the secret versions are set.

src/test/functions/cache/hash.spec.ts
TheIronDev: Updated!

src/test/functions/cache/hash.spec.ts
TheIronDev: Good point! Updated (er, `empty()` makes the cast go away, but noted for everywhere else.)

src/test/functions/cache/hash.spec.ts
TheIronDev: Yep yep! Both have been added as of an hour ago. The fs piping has not been resolving though...

src/deploy/functions/cache/hash.ts
TheIronDev: I don't quite follow. I'll ping you offline.

TheIronDev: Spoke offline, I understand now. Updated the signature.

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,55 @@
...
+export async function getBackendHash(backend: Backend, sourceFile?: string) {
+  const hash = crypto.createHash("sha256");
+
+  // If present, hash the contents of the source file
+  if (sourceFile) {
+    await new Promise((resolve, reject) => {
+      const readStream = fs.createReadStream(sourceFile);
+      readStream.on("error", (err) => reject(err));
+      readStream.on("data", (chunk) => chunk && hash.push(chunk));
+      readStream.on("end", () => resolve(hash.digest("hex")));
```
taeold, line 20: nit* I don't think this digest is necessary.

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,55 @@
...
+    hash.push(JSON.stringify(backend.environmentVariables));
+  }
+
+  // Hash the secret versions.
+  const secretVersions = getSecretVersions(backend);
+  if (hasSecretVersions(secretVersions)) {
+    hash.push(JSON.stringify(secretVersions));
+  }
+
+  return hash.read().toString("hex");
```
taeold, line 35: q: When I read documentation for crypto module, I expected `hash.digest('hex')`.  Is this roughly the same thing?

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,55 @@
...
+function hasSecretVersions(secretVersions: Record<string, string>) {
+  return Object.keys(secretVersions).length;
+}
+
+// Hash the secret versions.
+function getSecretVersions(backend: Backend): Record<string, string> {
+  const endpointsList: Endpoint[] = allEndpoints(backend);
+  return secrets.of(endpointsList).reduce((memo, { secret, version }) => {
+    if (version) {
+      memo[secret] = version;
```
taeold, line 51: Is there a bug here where if I'm using different versions of secret (for some reason) for different functions, I loose info?

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,55 @@
+import * as fs from "fs";
+import * as crypto from "crypto";
+import * as secrets from "../../../functions/secrets";
+import { Backend, Endpoint, EnvironmentVariables, allEndpoints } from "../backend";
+
+/**
+ * Retrieves the unique hash given a {@link Backend} and sourceFile.
+ * @param backend Backend of a set of functions
+ * @param sourceFile Packaged file contents of functions
```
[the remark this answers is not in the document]
inlined: Maybe we could use a different name to resolve this confusion. This is the zip file that will be sent to `generateUploadUrl`

src/deploy/functions/cache/hash.ts
inlined: Well, it's ambiguous since the Promise isn't annotated as returning anything and IDK if TypeScript can deduce it here. I personally think that this Promise should return the hash, not the digest, so that the env vars and secrets configs can be pushed onto the same Hash object. I think I would probably create a utility for adding a file to a hash and then call something like:

```
const hash = crypto.createHash("sha256");
await addFileTohash(sourceFile, hash);
await addFileToHash(.env, hash);
await addFileTohash(.env.alias, hash)
await addFileToHash(.env.alias, hash)
hash.push(JSON.stringify(endpoint.secrets));
```

Though that seems super wasteful becuase the source & dotenv hashes are really expensive to generate (file I/O) and are the same for an entire backend. So I'm starting to think that you can do a concatination of multiple hashes as long as we can guarantee that the string doesn't get too long (and if it does... I guess we can hash it again?).

So for a whole project you'll share a hash of the env files. And within a codbase, all v1 functions share one source hash and all v2 functions share another source hash. Then finally, each endpoint will have a hash based on the config of the secrets it depends upon.

But hashes are themselves pretty long and we may not be able to hold a whole hash in a label value, so... maybe an endpoint's hash is `hash(hash(env) + hash(source) + hash(secrets))`

src/deploy/functions/cache/hash.ts
inlined: Yes, I would recommend `hash.digest('hex')` here as the more direct API call. `hash.read()` is meant as a way of making a hash usable as a pipe. In order to even use the `read` command, for example, you are supposed to wait for the `readable` event.

src/deploy/functions/cache/hash.ts
TheIronDev: This is totally a bug. Good thing this function is being blown away by an endpoint version. 👍

src/deploy/functions/cache/hash.ts
TheIronDev: Updated!

TheIronDev: + thanks for talking through the difference.

src/deploy/functions/cache/hash.ts
TheIronDev: Updated the generation of the hashes to be `hash(hash(env) + hash(source) + hash(secrets))`.

#4876 will be updated to reflect this shortly

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  // If present, hash the contents of the source file
+  if (sourceFile) {
+    await new Promise((resolve, reject) => {
+      const readStream = fs.createReadStream(sourceFile);
+      readStream.on("error", (err) => reject(err));
+      readStream.on("data", (chunk) => chunk && hash.push(chunk));
+      readStream.on("end", () => resolve(hash.digest("hex")));
+    });
+  }
```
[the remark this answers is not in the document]
TheIronDev: Neat! I lost a giant chunk of time getting the Promise + callback mix to play nicely. So... thanks for calling this out, updated to use `fs/promise` now.

src/deploy/functions/cache/hash.ts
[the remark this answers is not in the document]
TheIronDev: I renamed it to `pathToGeneratedPackageFile`, but am down to changing to a better name as well. 👍

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,81 @@
...
+
+function hasEnvironmentVariables(environmentVariables: EnvironmentVariables): boolean {
+  return !!Object.keys(environmentVariables).length;
+}
+
+function hasSecretVersions(secretVersions: Record<string, string>): boolean {
+  return !!Object.keys(secretVersions).length;
+}
+
+// Hash the secret versions.
```
colerogers, line 73: I think this comment is describing the wrong function

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,81 @@
...
+function getSecretVersions(endpoint: Endpoint): Record<string, string> {
```
colerogers, line 74: I feel like this might be generic enough that we should put it in `backend.ts`, thoughts?

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,81 @@
...
+    hash.update(JSON.stringify(secretVersions));
+  }
+
+  return hash.digest("hex");
+}
+
+/**
+ * Generates a unique hash derived from the hashes generated from the package source, environment variables, and endpoint secrets.
+ */
+export function getEndpointHash(sourceHash: string, envHash: string, secretsHash: string) {
```
colerogers, line 56: linter: return type

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  return hash.digest("hex");
+}
+
+/**
+ * Retrieves a hash generated from the secrest of {@link Endpoint}.
+ * @param endpoint Endpoint
+ */
+export function getSecretsHash(endpoint: Endpoint) {
```
colerogers, line 41: linter: return type

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  return hash.digest("hex");
+}
+
+/**
+ * Retrieves the unique hash given a pathToGeneratedPackageFile.
+ * @param backend Backend of a set of functions
+ * @param pathToGeneratedPackageFile Packaged file contents of functions
+ */
+export async function getSourceHash(pathToGeneratedPackageFile?: string) {
```
colerogers, line 25: linter: return type

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,81 @@
...
+  if (hasEnvironmentVariables(backend.environmentVariables)) {
+    hash.update(JSON.stringify(backend.environmentVariables));
+  }
+
+  return hash.digest("hex");
+}
+
+/**
+ * Retrieves the unique hash given a pathToGeneratedPackageFile.
+ * @param backend Backend of a set of functions
```
colerogers, line 22: linter: unused param

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,81 @@
+import { readFile } from "node:fs/promises";
+import * as crypto from "crypto";
+import { Backend, Endpoint, EnvironmentVariables } from "../backend";
+
+/**
+ * Generates a hash from the environment variables of a {@link Backend}.
+ * @param backend Backend of a set of functions
+ */
+export function getEnvironmentVariablesHash(backend: Backend) {
```
colerogers, line 9: linter: return type

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,81 @@
...
+export function getEndpointHash(sourceHash: string, envHash: string, secretsHash: string) {
+  const hash = crypto.createHash("sha256");
+
+  const combined = [envHash, sourceHash, secretsHash].join("");
+  hash.update(combined);
+
+  return hash.digest("hex");
+}
+
+function hasEnvironmentVariables(environmentVariables: EnvironmentVariables): boolean {
```
colerogers, line 65: same here, should we put this in backend.ts?

src/deploy/functions/cache/hash.ts
TheIronDev: Yep, updated!

src/deploy/functions/cache/hash.ts
TheIronDev: Ty!

src/deploy/functions/cache/hash.ts
TheIronDev: Ty!

src/deploy/functions/cache/hash.ts
TheIronDev: Removed, ty!

src/deploy/functions/cache/hash.ts
TheIronDev: 👍

src/deploy/functions/cache/hash.ts
TheIronDev: Eh, its a 1-liner that didn't need its own function, so I removed it. 🤷

src/deploy/functions/cache/hash.ts
TheIronDev: 👍

src/deploy/functions/cache/hash.ts
TheIronDev: You're right! I moved it to secrets.ts instead though.

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,69 @@
...
+  const hash = crypto.createHash("sha256");
+
+  // Hash the contents of the dotenv variables
+  const hasEnvironmentVariables = !!Object.keys(backend.environmentVariables).length;
+  if (hasEnvironmentVariables) {
+    hash.update(JSON.stringify(backend.environmentVariables));
+  }
+
+  return hash.digest("hex");
+}
```
taeold, line 20: I think we should get environment variable hash for an endpoint and not from the backend. It's possible that an endpoint might have different set of environment variables from another endpoint in the same backend. For example, you can go to Pantheon and manually add an env var entry for a function.

taeold (commented):
Should we update the name of this PR?

It sounds like we are actually headed into implementation where you can no-op deploy an endpoint instead of no-op deploying an entire codebase since hash is now unique per endpoint.

src/deploy/functions/cache/hash.ts
TheIronDev: Spoke offline. The understanding we came to was that in instances of wantBackend, `backend.environmentVariables === endpoint.environmentVariables`.

So in the interest of conserving the # of hashing requests, I'm keeping it on the `Backend` level and allowing it to be shared with the hashing of multiple Endpoints.

colerogers (approved):
lgtm, but I'd also get @taeold's sign-off

src/deploy/functions/cache/hash.ts
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Retrieves the unique hash given a pathToGeneratedPackageFile.
+ * @param pathToGeneratedPackageFile Packaged file contents of functions
+ */
+export async function getSourceHash(pathToGeneratedPackageFile?: string): Promise<string> {
```
taeold, line 26: q: When is `pathToGeneratedPackageFile` undefined during function deploy?

TheIronDev: Good point! It shouldn't be! Updated!

src/test/functions/secrets.spec.ts
```diff
@@ -176,6 +176,36 @@ describe("functions/secret", () => {
...
+
+      const endpoint = {
+        ...ENDPOINT,
+        secretEnvironmentVariables: [secret1, secret2, secret3],
+      };
+
+      expect(secrets.getSecretVersions(endpoint)).to.deep.eq({
+        [secret1.secret]: secret1.version,
+        [secret3.secret]: secret3.version,
+      });
```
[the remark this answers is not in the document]
TheIronDev: Because secret2 does not have a version.

Or do you think `[secret2.secret]: undefined` is a better assumption?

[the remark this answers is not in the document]
taeold: But secrets will always have a version? It will be resolved by the time this function gets called no?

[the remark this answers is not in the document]
TheIronDev: `SecretEnvVar.version` is optional, so it's "possible" its not defined yet... but I agree the test is confusing things.

Updated!

TheIronDev:
> Just some nits on testing - I think secret versions. will always exists. Just worried if the assumption that it may not exist is relied on some important way 🤔

Makes sense! I updated the tests to clear away confusion about the state of versions.

src/test/functions/secrets.spec.ts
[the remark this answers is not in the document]
taeold: It is optional, but at the time this function is called, it won't be! We always resolve secret versions during deploy.

PR #4860, opened 2022-08-11
Fix uploads from android sdk to the storage emulator

Yuangwang:
### Description

Fixes issue where blocked uploads from the android sdk crash the storage emulator

### Scenarios Tested

Tested manually from the android sdks and modified integration tests to accommodate the new flow.


scripts/storage-emulator-integration/tests.ts
```diff
@@ -1949,23 +1949,17 @@ hello there!
...
         .post(`/v0/b/${storageBucket}/o/test_upload.jpg?uploadType=resumable&name=test_upload.jpg`)
         .set({
           // Authorization missing
           "X-Goog-Upload-Protocol": "resumable",
           "X-Goog-Upload-Command": "start",
         })
-        .expect(200)
-        .then((res) => new URL(res.header["x-goog-upload-url"]));
+        .expect(403)
+        .then((res) => res.header["x-goog-upload-status"]);
```
abhis3, line 1960: Curious why we're doing this over making a new URL like we do elsewhere?

src/test/emulators/storage/files.spec.ts
```diff
@@ -80,19 +80,21 @@ describe("files", () => {
...
+        const initialUpload = _uploadService.startResumableUpload({
           bucketId: "bucket",
           objectId: "dir%2Fobject",
           metadataRaw: "{}",
-        }).id;
-        _uploadService.continueResumableUpload(uploadId, Buffer.from("hello world"));
-        const upload = _uploadService.finalizeResumableUpload(uploadId);
+        });
+        _uploadService.startAuthorizedResumableUpload(initialUpload);
+        _uploadService.continueResumableUpload(initialUpload.id, Buffer.from("hello world"));
```
abhis3, line 96: Potentially dumb question, why are we appending the bytes in "hello world" to the buffer? I see it was there before but not making a ton of sense to me.

src/emulator/storage/files.ts
```diff
@@ -331,6 +331,37 @@ export class StorageLayer {
...
+        bucket: upload.bucketId,
+        contentType: upload.metadata.contentType || "application/octet-stream",
+        contentDisposition: upload.metadata.contentDisposition,
+        contentEncoding: upload.metadata.contentEncoding,
+        contentLanguage: upload.metadata.contentLanguage,
+        cacheControl: upload.metadata.cacheControl,
+        customMetadata: upload.metadata.metadata,
+      },
+      this._cloudFunctions,
+      Buffer.alloc(0)
```
tonyjhuang, line 348: This is potentially going to give us incorrect security rule results if there's a security rule that checks the `request.size` variable.

This is also essentially why we wait until the file is fully uploaded before evaluating security rules, because we need an accurate file size before we can get a fully formed rules verdict.

tonyjhuang: e.g. the following security rule with _always_ fail given the above logic

```
allow write: if request.resource.size > 0;
```

abhis3 (approved):

src/emulator/storage/upload.ts
```diff
@@ -50,6 +50,9 @@ export type StartResumableUploadRequest = {
 /** Error that signals a resumable upload that's expected to be active is not. */
 export class UploadNotActiveError extends Error {}
 
+/** Error that signals a resumable upload that shouldn't be finalized is */
```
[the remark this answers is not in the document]
Yuangwang: I'll add the missing period

scripts/storage-emulator-integration/conformance/firebase.endpoints.test.ts
```diff
@@ -173,6 +173,40 @@ describe("Firebase Storage endpoint conformance tests", () => {
...
           .expect(403);
       });
+
+      it("should return 400 when calling finalize on an already finalized resumable upload", async () => {
+        const uploadURL = await supertest(firebaseHost)
+          .post(
+            `/v0/b/${storageBucket}/o/test_upload.jpg?uploadType=resumable&name=test_upload.jpg`
+          )
+          .set({
+            Authorization: "Bearer owner",
```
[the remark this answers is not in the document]
Yuangwang: I tested against prod as well, and I notice that prod returns a 200 response. In prod this makes sense because there was an initial rules check and a second finalize is probably seen as a no op. In emulator however we aren't doing the initial rules check and this error is indication of a failed check in the android sdk so I'm not sure if I should match prod behavior or return a different code here.

In any case the emulator stability bug will be fixed but the error message from the android sdk won't make much sense if I return 200 response. What are your thoughts here?

scripts/storage-emulator-integration/conformance/firebase.endpoints.test.ts
```diff
@@ -173,6 +173,37 @@ describe("Firebase Storage endpoint conformance tests", () => {
           })
           .expect(403);
       });
+
+      it("should return 403 when resumable upload is unauthenticated and finalize is called again", async () => {
```
Yuangwang, line 177: Pointed tests against prod as well and now it passes on both the emulator and prod.

scripts/storage-emulator-integration/conformance/firebase.endpoints.test.ts
```diff
@@ -173,6 +173,40 @@ describe("Firebase Storage endpoint conformance tests", () => {
           })
           .expect(403);
       });
+
+      it("should return 400 when calling finalize on an already finalized resumable upload", async () => {
```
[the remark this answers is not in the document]
Yuangwang: gcs endpoints don't seem to have an explicit finalize step (https://cloud.google.com/storage/docs/performing-resumable-uploads) so there is no equivalent test that could be added in gcs.endpoints.test, instead I noticed that gcs endpoint tests weren't testing the upload step so I added a test for that.

src/emulator/storage/files.ts
```diff
@@ -122,7 +122,8 @@ export class StorageLayer {
     private _rulesValidator: FirebaseRulesValidator,
     private _adminCredsValidator: AdminCredentialValidator,
     private _persistence: Persistence,
-    private _cloudFunctions: StorageCloudFunctions
+    private _cloudFunctions: StorageCloudFunctions,
+    private _uploadService: UploadService
```
tonyjhuang, line 126: do you need this?

src/emulator/storage/upload.ts
```diff
@@ -172,13 +176,34 @@ export class UploadService {
...
+  public setResponseCode(uploadId: string, code: number): void {
+    const upload = this._uploads.get(uploadId);
+    if (upload) {
+      upload.prevResponseCode = code;
+    }
+  }
+
+  /**
+   * Gets previous response code.
+   * In the case the previous response code doesn't exist (after importing) return 200
```
tonyjhuang, line 201: upload ids don't get persisted across import/export

src/emulator/storage/apis/firebase.ts
```diff
@@ -342,9 +347,14 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
         upload = uploadService.finalizeResumableUpload(uploadId);
       } catch (err) {
         if (err instanceof NotFoundError) {
+          uploadService.setResponseCode(uploadId, 404);
```
tonyjhuang, line 350: do you not need to update the GCS API?

src/emulator/storage/upload.ts
Yuangwang: yeah thats true, in the case where someone exports and then imports data I was thinking we would just return 200 when finalize is called twice. 

It didn't seem to make sense to persist this previous response code on export since that would mean adding it to the metadata which seemed strange and different from prod metadata.

In any case returning 200 here would keep the emulator from crashing and as far as I know theres no equivalent flow to this import/export on prod so this wouldn't be introducing any divergence.

src/emulator/storage/apis/firebase.ts
tonyjhuang: That's fair, I more meant like what if you were to send direct HTTP requests to `finalize` via `supertest`, but we can say that's out of scope.

tonyjhuang (approved):

PR #4866, opened 2022-08-12
[Function-NoOp] Implement Hash from Label

TheIronDev:
### Description

Given a hash label from a gcf function, apply it to the Backend's Endpoint.

### Scenarios Tested

- [x] Given a hash label, does it apply to Backend's Endpoint

### DIFF
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -359,6 +359,10 @@ export type Endpoint = TargetIds &
     // This is a temporary fix to address https://github.com/firebase/firebase-tools/issues/4171
     // GCFv1 can be http or https and GCFv2 is always https
     securityLevel?: gcf.SecurityLevel;
+
+    // "Hash" is a value derived from function labels that is used to check if there are any changes
+    // between the serverside and local copies of the function.
+    hash?: string;
   };
 
 export interface RequiredAPI {
--- a/src/gcp/cloudfunctions.ts
+++ b/src/gcp/cloudfunctions.ts
@@ -11,23 +11,16 @@ import * as projectConfig from "../functions/projectConfig";
 import { Client } from "../apiv2";
 import { functionsOrigin } from "../api";
 import { AUTH_BLOCKING_EVENTS } from "../functions/events/v1";
+import {
+  BLOCKING_EVENT_TO_LABEL_KEY,
+  BLOCKING_LABEL,
+  BLOCKING_LABEL_KEY_TO_EVENT,
+  CODEBASE_LABEL,
+  HASH_LABEL,
+} from "./constants";
 
 export const API_VERSION = "v1";
-export const CODEBASE_LABEL = "firebase-functions-codebase";
-export const HASH_LABEL = "firebase-functions-hash";
 const client = new Client({ urlPrefix: functionsOrigin, apiVersion: API_VERSION });
-
-export const BLOCKING_LABEL = "deployment-blocking";
-
-const BLOCKING_LABEL_KEY_TO_EVENT: Record<string, typeof AUTH_BLOCKING_EVENTS[number]> = {
-  "before-create": "providers/cloud.auth/eventTypes/user.beforeCreate",
-  "before-sign-in": "providers/cloud.auth/eventTypes/user.beforeSignIn",
-};
-
-const BLOCKING_EVENT_TO_LABEL_KEY: Record<typeof AUTH_BLOCKING_EVENTS[number], string> = {
-  "providers/cloud.auth/eventTypes/user.beforeCreate": "before-create",
-  "providers/cloud.auth/eventTypes/user.beforeSignIn": "before-sign-in",
-};
 
 interface Operation {
   name: string;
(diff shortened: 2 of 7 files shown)


src/gcp/cloudfunctions.ts
```diff
@@ -559,6 +560,9 @@ export function endpointFromFunction(gcfFunction: CloudFunction): backend.Endpoi
     );
   }
   endpoint.codebase = gcfFunction.labels?.[CODEBASE_LABEL] || projectConfig.DEFAULT_CODEBASE;
+  if (gcfFunction.labels?.[HASH_LABEL]) {
+    endpoint.hash = gcfFunction.labels?.[HASH_LABEL];
```
colerogers, line 564: Don't we also need the reverse logic to add the hash label to the function from the endpoint? Look inside `functionFromEndpoint()` on L527 for examples of how we do that with the other label types

taeold, line 564: nit* since you already checked for existence of `HASH_LABEL`, I think you can omit the `?` (think TS is smart enough to deduce that).

taeold (dismissed):

src/gcp/cloudfunctions.ts
taeold: +1. Is it being done in another PR?

src/deploy/functions/backend.ts
```diff
@@ -359,6 +359,10 @@ export type Endpoint = TargetIds &
     // This is a temporary fix to address https://github.com/firebase/firebase-tools/issues/4171
     // GCFv1 can be http or https and GCFv2 is always https
     securityLevel?: gcf.SecurityLevel;
+
+    // "Hash" is a value derived from function labels that is used to check if there are any changes
+    // between the serverside and local copies of the function.
+    hash?: string;
```
inlined, line 365: Sometimes I wish an Endpoint was a class so we could default fields like this to "" and not worry about nullish values.

src/gcp/cloudfunctions.ts
inlined: This should implement functionFromEndpoint and endpointFromFunction in both cloudfunctions.ts and cloudfunctionsv2.ts

inlined: Yes, any time you're in the clause of an if statement that did a null check, the compiler stops requiring the ?. And in fact we should remove it to
A) Prove to ourselves that we've handled all edge cases
B) Ensure that this code, if copy + pasted, will only work when similarly guarded

src/test/gcp/cloudfunctions.spec.ts
```diff
@@ -455,6 +455,30 @@ describe("cloudfunctions", () => {
...
     });
+
+    it("should derive hash from labels", () => {
+      expect(
+        cloudfunctions.endpointFromFunction({
+          ...HAVE_CLOUD_FUNCTION,
+          httpsTrigger: {},
+          labels: {
+            ...CLOUD_FUNCTION.labels,
+            [cloudfunctions.CODEBASE_LABEL]: "my-codebase",
```
inlined, line 466: In general I wonder if we should delete the labels in endpointFromFunction so that tests don't have to worry about both the label and the property implementation of these features when creating fake data.

src/gcp/cloudfunctions.ts
TheIronDev: Updated! Thanks for calling out!

TheIronDev: Good catch! Updated!

src/test/gcp/cloudfunctions.spec.ts
TheIronDev: I noticed this part. I was concerned about unintended side-effects, so I left the forced assignment as-is.

src/gcp/constants.ts
```diff
@@ -0,0 +1,13 @@
...
+export const HASH_LABEL = "firebase-functions-hash";
+export const BLOCKING_LABEL = "deployment-blocking";
+export const BLOCKING_LABEL_KEY_TO_EVENT: Record<string, typeof AUTH_BLOCKING_EVENTS[number]> = {
+  "before-create": "providers/cloud.auth/eventTypes/user.beforeCreate",
+  "before-sign-in": "providers/cloud.auth/eventTypes/user.beforeSignIn",
+};
+export const BLOCKING_EVENT_TO_LABEL_KEY: Record<typeof AUTH_BLOCKING_EVENTS[number], string> = {
+  "providers/cloud.auth/eventTypes/user.beforeCreate": "before-create",
+  "providers/cloud.auth/eventTypes/user.beforeSignIn": "before-sign-in",
+};
```
colerogers, line 13: A downside for putting these here is that this directory is for all GCP services, not just functions. Maybe defining them in `src/functions` would be a better option. But @taeold might have some opinions about yanking these

colerogers (commented):
LGTM, just the open issue about the constants file that @taeold should sign off on

src/gcp/constants.ts
taeold: Agree with @colerogers here. Technically `gcp/cloudfunctions.ts` doesn't feel that great either, but at least that file is mainly used by us.

taeold (approved):
lgtm - mone minor comment on `gcp/constants.ts`.

PR #4869, opened 2022-08-13
Fix storage gcs conformance tests

tonyjhuang:
### Description

* Fixes default metadata field bugs in https://github.com/firebase/firebase-tools/issues/3398#issuecomment-1159821364 by setting appropriate default metadata fields depending on which API is used
* Introduces [media upload protocol](https://cloud.google.com/storage/docs/uploading-objects#uploading-an-object) support to both GCS and Firebase APIs 
* Adds support to GCS API for `/storage/v1/b/:bucketId/o/:objectId` url pattern


scripts/storage-emulator-integration/conformance/firebase-js-sdk.test.ts
```diff
@@ -472,36 +456,35 @@ describe("Firebase Storage JavaScript SDK conformance tests", () => {
...
-      //     md5Hash: "string",
-      //     metageneration: "string",
-      //     name: "string",
-      //     size: "number",
-      //     timeCreated: "string",
-      //     type: "string",
-      //     updated: "string",
-      //   });
-      // });
+      it.only("should return file metadata", async () => {
```
abhis3, line 459: `.only`

src/emulator/storage/apis/firebase.ts
```diff
@@ -120,7 +120,8 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
         data = gunzipSync(data);
       }
       res.setHeader("Accept-Ranges", "bytes");
-      res.setHeader("Content-Type", metadata.contentType);
+      res.setHeader("Content-Type", metadata.contentType || "application/octet-stream");
+      res.setHeader("Content-Disposition", metadata.contentDisposition || "inline");
```
abhis3, line 124: Shouldn't these fields be set already when uploading on object to the bucket?

tonyjhuang: So the way Content-Type works:

*  is if it's not passed in the upload request, the metadata field is empty but when fetching it, it defaults to "application/octet-stream"
* if the file name has a supported file extension e.g. .jpg, I believe the server automatically fills the field with the appropriate content type, e.g. image/jpg
* if you pass in an explicit metadata field, that overrides the above behaviors


This change adds support for the first bullet, the third bullet already works in the emulator. We don't have support for the inferred content type, unclear if we want to prioritize something like that.

scripts/storage-emulator-integration/conformance/firebase.endpoints.test.ts
```diff
@@ -71,6 +73,75 @@ describe("Firebase Storage endpoint conformance tests", () => {
...
+  describe("metadata", () => {
+    it("should set default metadata", async () => {
+      const fileName = "dir/someFile";
+      const encodedFileName = "dir%2FsomeFile";
+      await supertest(firebaseHost)
+        .post(`/v0/b/${storageBucket}/o?name=${fileName}`)
+        .set(authHeader)
+        .send(Buffer.from("hello world"))
+        .expect(200)
+        .then((res) => res.body.name);
```
Yuangwang, line 85: Are you using this result anywhere?

scripts/storage-emulator-integration/conformance/firebase.endpoints.test.ts
```diff
@@ -71,6 +73,75 @@ describe("Firebase Storage endpoint conformance tests", () => {
...
+  });
+
+  describe("media upload", () => {
+    it("should default to media upload if upload type is not provided", async () => {
+      await supertest(firebaseHost)
+        .post(`/v0/b/${storageBucket}/o?name=${ENCODED_TEST_FILE_NAME}`)
+        .set(authHeader)
+        .send(Buffer.from("hello world"))
+        .expect(200)
+        .then((res) => res.body.name);
```
Yuangwang, line 134: Are you using this result anywhere?

scripts/storage-emulator-integration/conformance/gcs.endpoints.test.ts
```diff
@@ -65,107 +83,136 @@ describe("GCS endpoint conformance tests", () => {
...
+  });
+
+  describe("Metadata", () => {
+    it("should set default metadata", async () => {
+      await supertest(storageHost)
+        .post(`/upload/storage/v1/b/${storageBucket}/o?name=${TEST_FILE_NAME}`)
+        .set(authHeader)
+        .send(Buffer.from("hello world"))
+        .expect(200)
+        .then((res) => res.body.name);
```
Yuangwang, line 112: Are you using this result anywhere?

package.json
```diff
@@ -202,6 +204,7 @@
     "@typescript-eslint/parser": "^5.9.0",
     "chai": "^4.3.4",
     "chai-as-promised": "^7.1.1",
+    "deep-equal-in-any-order": "^2.0.0",
```
bkendall, line 207: Looking at the way this is used, I think you should be able to omit this and do

```typescript
expect(Object.keys(obj)).to.have.members([/*...*/]);
```

https://www.chaijs.com/api/bdd/#method_members

LMK if that doesn't work for you.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "firebase-tools",
-  "version": "11.6.0",
+  "version": "11.7.0",
```
bkendall, line 3: Make sure to fix up issues like this and the CHANGELOG removing everything. This in particular is a dangerous inclusion.

package.json
```diff
@@ -23,6 +23,7 @@
     "mocha": "nyc mocha 'src/test/**/*.{ts,js}'",
     "mocha:storage-spec": "mocha 'src/test/emulators/storage/*.spec.ts'",
     "mocha:storage-emulator-integration": "mocha 'scripts/storage-emulator-integration/tests.ts'",
+    "mocha:storage-debug": "mocha 'scripts/storage-emulator-integration/conformance/*.test.ts'",
```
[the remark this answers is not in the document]
bkendall: +1 - this, and the above `storage` specific entries really shouldn't be here.

You can run individual tests _really_ easily (outside of the `scripts` directory, anyway) doing `npx mocha src/test/file.spec.ts`

bkendall (changes requested):
This seems to touch a lot of files... is that intended?

package.json
tonyjhuang: This isn't quite equivalent because `have.members` checks for set inclusion:

`expect([1,2,3]).to.have.members([1,2])`

The above evaluates to true whereas I want an exact match on the keys

scripts/storage-emulator-integration/conformance/firebase.endpoints.test.ts
tonyjhuang: nope good catch.

scripts/storage-emulator-integration/conformance/firebase.endpoints.test.ts
tonyjhuang: nope removed

package.json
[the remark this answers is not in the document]
tonyjhuang: removed

bkendall (approved):
Thanks for addressing those bits :)

Yuangwang (approved):
LGTM, thanks for the tests and refactor!

PR #4873, opened 2022-08-15
Only start eventarc emulator when there is at least one emulatable backend

huangjeff5:
### Description

Eventarc emulator starts even when Functions emulator doesn't start. This PR fixes that.


yuchenshi (approved):

joehan (approved):
LGTM

PR #4874, opened 2022-08-15
Differentiate param input types by property name instead of "type" field

Berlioz:


inlined (approved):
You don't need the `text: {}` field for `{ textInput: {} }`

PR #4876, opened 2022-08-15
[Function-NoOp] Apply Hash to wantBackend's endpoints

TheIronDev:
### Description

This PR adds the functionality to generate hashes for each endpoint given a list of `Backends[]`

### Scenarios Tested

- [x] Test wantBackend with 1 endpoint
- [x] Test wantBackend with 2 endpoints
- [x] Test setting multiple wantBackends

### DIFF
--- a/src/deploy/functions/cache/applyHash.ts
+++ b/src/deploy/functions/cache/applyHash.ts
@@ -1,0 +1,56 @@
+import { Backend, allEndpoints } from "../backend";
+import * as args from "../args";
+import {
+  getEndpointHash,
+  getEnvironmentVariablesHash,
+  getSecretsHash,
+  getSourceHash,
+} from "./hash";
+
+/**
+ *
+ * Updates all the CodeBase {@link Backend}, applying a hash to each of their {@link Endpoint}.
+ */
+export async function applyBackendHashToBackends(
+  wantBackends: Record<string, Backend>,
+  context: args.Context
+): Promise<void> {
+  // envHash
+  const envHash = getEnvHash(wantBackends);
+  for (const [codebase, wantBackend] of Object.entries(wantBackends)) {
+    const source = context?.sources?.[codebase]; // populated earlier in prepare flow
+    const sourceV1Hash = source?.functionsSourceV1
+      ? await getSourceHash(source?.functionsSourceV1)
+      : undefined;
+    const sourceV2Hash = source?.functionsSourceV2
+      ? await getSourceHash(source?.functionsSourceV2)
+      : undefined;
+    applyBackendHashToBackend(wantBackend, envHash, sourceV1Hash, sourceV2Hash);
+  }
+}
+
+/**
+ * Updates {@link Backend}, applying a unique hash to each {@link Endpoint}.
+ */
+export function applyBackendHashToBackend(
+  wantBackend: Backend,
+  envHash: string,
+  sourceV1Hash?: string,
+  sourceV2Hash?: string
+): void {
+  for (const endpoint of allEndpoints(wantBackend)) {
+    const secretsHash = getSecretsHash(endpoint);
+    const isV2 = endpoint.platform === "gcfv2";
+    const sourceHash = isV2 ? sourceV2Hash : sourceV1Hash;
+    endpoint.hash = getEndpointHash(sourceHash, envHash, secretsHash);
+  }
+}
+
+/** @returns hash of first wantBackend */
+function getEnvHash(wantBackends: Record<string, Backend>): string {
+  const backends = Object.values(wantBackends);
+  if (backends.length) {
+    return getEnvironmentVariablesHash(backends[0]);
+  }
+  return "";
+}
--- a/src/deploy/functions/cache/applyHash.ts
+++ b/src/deploy/functions/cache/applyHash.ts
@@ -15,8 +15,6 @@ export async function applyBackendHashToBackends(
   wantBackends: Record<string, Backend>,
   context: args.Context
 ): Promise<void> {
-  // envHash
-  const envHash = getEnvHash(wantBackends);
   for (const [codebase, wantBackend] of Object.entries(wantBackends)) {
     const source = context?.sources?.[codebase]; // populated earlier in prepare flow
     const sourceV1Hash = source?.functionsSourceV1
@@ -25,6 +23,7 @@ export async function applyBackendHashToBackends(
     const sourceV2Hash = source?.functionsSourceV2
       ? await getSourceHash(source?.functionsSourceV2)
       : undefined;
+    const envHash = getEnvironmentVariablesHash(wantBackend);
     applyBackendHashToBackend(wantBackend, envHash, sourceV1Hash, sourceV2Hash);
   }
 }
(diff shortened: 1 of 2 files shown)


src/deploy/functions/cache/applyHash.ts
```diff
@@ -0,0 +1,31 @@
+import * as backend from "../backend";
+import * as args from "../args";
+
+/**
+ *
```
inlined, line 5: Let's just get in the habit of adding comments since all of our linters add empty ones otherwise.

src/deploy/functions/cache/applyHash.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export async function applyBackendHashToBackend(wantBackend: backend.Backend, source?: args.Source) {
+  // Blocked by #4852
+  // const hash = await getBackendHash(wantBackend: backend.Backend, source?: args.Source);
+  const hash = "Hash" + wantBackend + source;
+  backend.allEndpoints(wantBackend).forEach((endpoint) => {
+    // Blocked by #4866
+    // endpoint.hash = hash;
```
inlined, line 26: Interesting. I hadn't thought about making `hash` a separate field that's redundant with the label. I don't know if it's necessary. I don't know if it hurts anything.

inlined (commented):
I'm glad that you are calculating different hashes for different codebases, but again, you need to calculate a different hash for your v1 functions and your v2 functions because they come from different zip files.

src/deploy/functions/cache/applyHash.ts
TheIronDev: Agreed, updated!

src/deploy/functions/cache/applyHash.ts
TheIronDev: It is a little redundant, but I am a fan of the separate field for 3 reasons:

1. Documenting the field.
2. Only relying on the behavior of `label`'s `Record<string, string>` feels brittle and error prone. Theres a few points we re-assign the labels during the `release` phase. Although we spread the existing labels (so nothing is "lost"), I want to avoid important fields getting unintentionally clobbered.
3. Prior similar art of `Endpoint.codebase`

inlined: All sound like pretty compelling reasons. Thanks!

src/deploy/functions/cache/applyHash.ts
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Updates {@link Backend}, applying a unique hash to each {@link Endpoint}.
+ */
+export function applyBackendHashToBackend(
+  wantBackend: Backend,
+  envHash: string,
+  sourceV1Hash?: string,
+  sourceV2Hash?: string
+): void {
+  allEndpoints(wantBackend).forEach((endpoint: Endpoint) => {
```
colerogers, line 41: style guides prefer `for(...)` over `forEach(...)`

src/deploy/functions/cache/applyHash.ts
```diff
@@ -0,0 +1,57 @@
...
+    const secretsHash = getSecretsHash(endpoint);
+    const isV2 = endpoint.platform === "gcfv2";
+    const sourceHash = isV2 ? sourceV2Hash : sourceV1Hash;
+    // Blocked by #4866
```
colerogers, line 45: Is this still blocked? If so, maybe we should make this PR merge into that one and remove this comment

colerogers (changes requested):

src/deploy/functions/cache/applyHash.ts
TheIronDev: Ah it got merged, no longer blocked, thanks for catching!!

src/deploy/functions/cache/applyHash.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ *
+ * Updates all the CodeBase {@link Backend}, applying a hash to each of their {@link Endpoint}.
+ */
+export async function applyBackendHashToBackends(
+  wantBackends: Record<string, Backend>,
+  context: args.Context
+): Promise<void> {
+  // envHash
```
taeold, line 18: nit: stray comment?

src/deploy/functions/cache/applyHash.ts
```diff
@@ -0,0 +1,56 @@
...
+  const envHash = getEnvHash(wantBackends);
```
taeold, line 19: I think this logic is wrong. Each codebase has its own set of environment variables, optionally specified by `.env` in its source directory.

src/test/deploy/functions/cache/applyHash.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+      expect(getEndpointHash).to.have.been.calledWith(
+        "source=backend2_sourceV2",
+        "env=shared_environment_variables",
+        "secret=secret2"
+      );
+      expect(endpoint2.hash).to.equal(
+        "source=backend2_sourceV2&env=shared_environment_variables&secret=secret2"
+      );
+      // Note: env variables only derived from the first backend
+      // (since all backends in wantBackends shared the same environment variables)
```
taeold, line 114: Commented above - I don't think this is true. Each codebase can have its own set of environment variable.

taeold (changes requested):

src/deploy/functions/cache/applyHash.ts
TheIronDev: Updated! Thank you!

src/deploy/functions/cache/applyHash.ts
TheIronDev: Yep, removed!

src/deploy/functions/cache/applyHash.ts
TheIronDev: Thanks for calling that out. Its safer to repeat the calls for those instances. Updated!

src/test/deploy/functions/cache/applyHash.spec.ts
TheIronDev: Thanks for calling out, removed comment and updated test!

src/deploy/functions/cache/applyHash.ts
```diff
@@ -0,0 +1,46 @@
...
+      : undefined;
+    const envHash = getEnvironmentVariablesHash(wantBackend);
+    applyBackendHashToBackend(wantBackend, envHash, sourceV1Hash, sourceV2Hash);
+  }
+}
+
+/**
+ * Updates {@link Backend}, applying a unique hash to each {@link Endpoint}.
+ */
+export function applyBackendHashToBackend(
```
taeold, line 34: nit* Documentation makes me think this function name should be `applyBackendHashToEndpoints` 🤔. But no strong opinions on naming internal api 🤷🏼‍♂️

src/test/deploy/functions/cache/applyHash.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+        codebase: "backend1",
+        secretEnvironmentVariables: [
+          {
+            key: "key",
+            secret: "secret1",
+            projectId: "projectId",
+            version: "1",
+          },
+        ],
+      } as backend.Endpoint;
```
taeold, line 52: nit* I'd try to avoid type assertion. It make it really hard to refactor.

Prefer `const endpoint1: backend.Endpoint` if appropriate,

taeold (approved):

src/test/deploy/functions/cache/applyHash.spec.ts
TheIronDev: Updated!

src/deploy/functions/cache/applyHash.ts
TheIronDev: `applyBackendHashToEndpoints` definitely read better. Updated!

PR #4882, opened 2022-08-15
Throw instead of prompting for CF3 param values in non-interactive mode

Berlioz:
Should be pretty self-explanatory.

### DIFF
--- a/src/deploy/functions/params.ts
+++ b/src/deploy/functions/params.ts
@@ -3,6 +3,7 @@ import { FirebaseError } from "../../error";
 import { promptOnce } from "../../prompt";
 import * as build from "./build";
 import { assertExhaustive, partition } from "../../functional";
+import { Options } from "../../options";
 
 type CEL = build.Expression<string> | build.Expression<number> | build.Expression<boolean>;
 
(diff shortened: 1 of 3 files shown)


src/deploy/functions/params.ts
```diff
@@ -248,6 +249,15 @@ export async function resolveParams(
   }
 
   for (const param of outstanding) {
+    if (nonInteractive) {
```
inlined, line 252: Rather than have a developer iterate through `firebase deploy` calls until one finally succeeds, let's move this outside of the for loop and include all missing params in one error. Something like:

```
if (outstanding.length && nonInteractive) {
  const envNames = outstanding.map(p => p.name).join(", ");
  throw new FirebaseError(
    `In non-interactive mode but have no value for the following environment variables: ${envNames}\n` +
    "To continue, either run `firebase deploy` with an interactive terminal, or add values to a dotenv file. " +
    "For information regarding how to use dotenv files, see https://firebase.google.com/docs/functions/config-env"
  );
}

inlined (approved):

src/deploy/functions/params.ts
Berlioz: Done, good call.

PR #4888, opened 2022-08-16
Implement removing Endpoint from Backend

TheIronDev:
### Description

This commit implements removing an endpoint from a backend.
Useful for instances where we would want to... remove an endpoint, from Backend.

Possibly in the `prepare` step.:wq

### Scenarios Tested

- [x] Test removing an endpoint from a backend

### DIFF
--- a/src/test/deploy/functions/backend.spec.ts
+++ b/src/test/deploy/functions/backend.spec.ts
@@ -597,6 +597,12 @@ describe("Backend", () => {
       region: "europe-west1",
     };
 
+    const endpointREMOVED: backend.Endpoint = {
+      ...endpointUS,
+      id: "endpointREMOVED",
+      region: "europe-west1",
+    };
+
     const bkend: backend.Backend = {
       ...backend.empty(),
     };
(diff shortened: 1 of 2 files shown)


src/test/deploy/functions/backend.spec.ts
```diff
@@ -629,6 +635,19 @@ describe("Backend", () => {
       expect(backend.someEndpoint(bkend, (fn) => fn.id === "missing")).to.be.false;
     });
 
+    it("removeEndpoints", () => {
+      // Add our endpoint
+      bkend.endpoints[endpointEU.region] = { [endpointREMOVED.id]: endpointREMOVED };
```
taeold, line 640: nit* A more idiomatic way of creating a backend from endpoint is to use `of` fn:

```
const b = backend.of(endpointREMOVED)
```

src/test/deploy/functions/backend.spec.ts
```diff
@@ -629,6 +635,19 @@ describe("Backend", () => {
...
+      // Verify its there given our someEndpoint check
+      expect(backend.someEndpoint(bkend, (fn) => fn.id === "endpointREMOVED")).to.be.true;
```
taeold, line 642: You can also use `hasEndpoint` fn:

```
expect(backend.hasEndpoint(b, endpointREMOVED)).to.be.true;
```

taeold (approved):
lgtm - curious why we need this?

inlined (commented):
Can I suggest a different approach altogether?

What if we modify `fabricator` so that it puts these functions in a fourth category: skippedEndpoints. Then we can retain them in the release phase if we want to use log lines or report metrics on codebases or functions skipped.

inlined:
Note: putting skipped functions in the release phase will mean in the long run that we probably want to create a plan in the fabricator during deploy because we'll want to see if we should even upload the source or if all functions that depend on it are being skipped.

TheIronDev:
> Note: putting skipped functions in the release phase will mean in the long run that we probably want to create a plan in the fabricator during deploy because we'll want to see if we should even upload the source or if all functions that depend on it are being skipped.

It also means that we still end up uploading on gcs+gcf the function during the deploy phase.

Stripping out the functions during the prepare phase was the current plan to skip the upload step, but I can see the definite pros to doing it in the fabricator too. Will fast follow on where is the best place to drop asap.

TheIronDev:
The goal of this PR was to provide a way to remove `Endpoint` from `backend.Backend`, which would be performed during the `prepare` phase of a deploy.

After further discussion, I am pivoting the design to redirect Endpoints into a different Changeset during the `release` phase, which removes the need for removing endpoints.

If in the future there arises a need to remove `Endpoint`s from `backend.Backend` then I will revive this PR.

PR #4890, opened 2022-08-16
Single project mode

christhompsongoogle:
Add a singleProjectMode flag to the firebase config which defaults to true (on init and if missing when read).

Pass the flag to firestore in the form of a warning for now.

Requires the latest version of firestore to be cut including CL/466521585

Tested: Manual testing for configuration and firestore emulator.


src/firebaseConfig.ts
```diff
@@ -167,6 +167,7 @@ export type EmulatorsConfig = {
     host?: string;
     port?: number;
   };
+  single_project_mode?: boolean;
```
bkendall, line 170: I don't think this should be underscore_case - generally variables are camelCase

bkendall (changes requested):
Is the update to `npm-shrinkwrap` intended?

src/firebaseConfig.ts
christhompsongoogle: Updated to singleProjectMode

christhompsongoogle:
> Is the update to `npm-shrinkwrap` intended?

No, I don't know how to revert it now. I've tried rebasing and checking out the master file. I resorted to regenerated the shrinkwrap to see if it would match :)

src/init/features/emulators.ts
```diff
@@ -100,6 +100,11 @@ export async function doSetup(setup: any, config: any) {
     ]);
   }
 
+  // Set the default behavior to be single project mode.
+  if (setup.config.emulators.singleProjectMode !== undefined) {
```
bkendall, line 104: If `singleProjectMode` in `firebase.json` is set to `false`, this condition means that `singleProjectMode` will be enabled. I'm sure that's not the intention.

Do you mean to just check if it's truthy?

```suggestion
  if (setup.config.emulators.singleProjectMode) {
```

bkendall (changes requested):
Please:

1. Add a `CHANGELOG` entry
2. Fix the CI tests

src/init/features/emulators.ts
christhompsongoogle: Fixed, nice find. After some testing this was backwards. The intention here is to set if it it doesn't exist, create the config, otherwise leave it as the user set it. The default should be true, but the current behavior is false.

bkendall (approved):

bkendall (approved):

kovan:
I am getting megabytes and megabytes of log entries with the message:

>  [2022-10-21T19:43:17.950Z] Multiple projectIds are not recommended in single project mode. Requested project ID ****, but the emulator is configured for ***. This warning will become an error in the future. To opt-out of single project mode add/set the '"single_project_mode": false' property in the firebase.json emulators config

To the point that it is impacting the performance of the build machine. 

Anyway, following the message, I have tried setting the mentioned flag "single_project_mode" to false in firebase.json, but the flag seems to be rejected by the emulator.
Investigating a bit more, it looks like the flag it is spelled as "singleProjectMode" in the [JSON schema](https://github.com/firebase/firebase-tools/blob/master/schema/firebase-config.json#L269)

I haven't yet tested it with that spelling, I will try to report back when I do.

bkendall:
@kovan please open a new issue with the relevant information. thanks!

PR #4899, opened 2022-08-18
[Function-NoOp] Update Fabricator to handle Endpoints to be Skipped

TheIronDev:
### Description

Updates `fabricator.ts` to handle log output for functions that should be skipped.

### Scenarios Tested

- [x] Given a changeset with an endpointToSkip, make sure the correct log output is returned

### DIFF
--- a/src/deploy/functions/release/fabricator.ts
+++ b/src/deploy/functions/release/fabricator.ts
@@ -125,11 +125,11 @@ export class Fabricator {
       this.logOpStart("creating", endpoint);
       upserts.push(handle("create", endpoint, () => this.createEndpoint(endpoint, scraper)));
     }
-    for (const endpoint of changes.endpointsToSkip) {
-      upserts.push(handle("skip", endpoint, () => this.skipEndpoint(endpoint)));
-    }
     if (changes.endpointsToSkip.length) {
-      upserts.push(this.skipEndpointsCompleted(changes.endpointsToSkip));
+      for (const endpoint of changes.endpointsToSkip) {
+        utils.logSuccess(this.getLogSuccessMessage("skip", endpoint));
+      }
+      utils.logSuccess(this.getSkippedDeployingNopOpMessage(changes.endpointsToSkip));
     }
     for (const update of changes.endpointsToUpdate) {
       this.logOpStart("updating", update.endpoint);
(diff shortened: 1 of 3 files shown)


src/deploy/functions/release/fabricator.ts
```diff
@@ -125,6 +125,10 @@ export class Fabricator {
       this.logOpStart("creating", endpoint);
       upserts.push(handle("create", endpoint, () => this.createEndpoint(endpoint, scraper)));
     }
+    for (const endpoint of changes.endpointsToSkip) {
+      this.logOpStart("skipping", endpoint);
+      upserts.push(handle("skip", endpoint, () => this.skipEndpoint()));
```
taeold, line 130: Doesn't this lead to kind of a funny deploy log like:

```
Skipping function 1...
Finished skipping function...
```

It's not a total blocker, but I'd expect something simpler like:

```
Not deploying function 1 - no change since last deploy (hash=???) 
Not deploying function 2 - no change since last deploy (hash=???) 
```

Or even something more general like:

```
Skipping deployment of following functions since CLI detected no changes:
  - Function 1 (hash=?)
  - Function 2 (hash=?)

To force deploy these function, run command XXX 
```

src/deploy/functions/release/fabricator.ts
```diff
@@ -125,6 +125,12 @@ export class Fabricator {
       this.logOpStart("creating", endpoint);
       upserts.push(handle("create", endpoint, () => this.createEndpoint(endpoint, scraper)));
     }
+    for (const endpoint of changes.endpointsToSkip) {
+      upserts.push(handle("skip", endpoint, () => this.skipEndpoint(endpoint)));
+    }
+    if (changes.endpointsToSkip.length) {
+      upserts.push(this.skipEndpointsCompleted(changes.endpointsToSkip));
```
taeold, line 132: This seems a bit roundabout way to log messages.

Do you think we'll need this wrapping around operation in the future. YAGNI would suggest just logging the messages here directly.

TheIronDev: TBH it did feel / seem roundabout, but didn't want to stray too far from existing conventions. 🤷

src/deploy/functions/release/fabricator.ts
TheIronDev: Updated so messaging is closer to that!

taeold (approved):

PR #4900, opened 2022-08-18
[Function-NoOp] Add a preview flag for skipping no-op function deploys

TheIronDev:
Adding `skipdeployingnoopfunctions ` as a preview flag.

Eventually, when `skipdeployingnoopfunctions ` is enabled, functions that are detected to be no-ops will not be re-deployed.

Example usage:

```
$ FIREBASE_CLI_PREVIEWS=skipdeployingnoopfunctions; firebase deploy
```

### DIFF
--- a/src/previews.ts
+++ b/src/previews.ts
@@ -12,7 +12,7 @@ interface PreviewFlags {
   frameworkawareness: boolean;
   functionsparams: boolean;
   crossservicerules: boolean;
-  skipnoopfunctiondeploy: boolean;
+  skipdeployingnoopfunctions: boolean;
 }
 
 export const previews: PreviewFlags = {
@@ -27,7 +27,7 @@ export const previews: PreviewFlags = {
   frameworkawareness: false,
   functionsparams: false,
   crossservicerules: false,
-  skipnoopfunctiondeploy: false,
+  skipdeployingnoopfunctions: false,
 
   ...(configstore.get("previews") as Partial<PreviewFlags>),
 };
--- a/src/previews.ts
+++ b/src/previews.ts
@@ -12,6 +12,7 @@ interface PreviewFlags {
   frameworkawareness: boolean;
   functionsparams: boolean;
   crossservicerules: boolean;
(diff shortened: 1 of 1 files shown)


src/previews.ts
```diff
@@ -26,6 +27,7 @@ export const previews: PreviewFlags = {
   frameworkawareness: false,
   functionsparams: false,
   crossservicerules: false,
+  skipnoopfunctiondeploy: false,
```
taeold, line 30: *nit - just for clarity, I'd suggest flag that starts with `functions` as a prefix - e.g. `functionsnoopdeploy`

taeold (approved):

bkendall (changes requested):
I'm not sure that this is the appropriate way to achieve the end goal - let's continue to chat.

src/previews.ts
TheIronDev: My read on this is `functionsnoopdeploy` suggests we are enabling no-op deploys. What we are trying to do is to enable skipping no-op deploys.

Maybe  `skipdeployingnoopfunctions` ?

TheIronDev:
> I'm not sure that this is the appropriate way to achieve the end goal - let's continue to chat.

I think we were at a misunderstanding of this flag being used as a `disable` vs `enable` flag. The goal/intention is this will be used to enable a new feature.

TheIronDev:
@taeold + @bkendall I updated the flag. LMKWYT

src/previews.ts
taeold: (Just playing with words here. I don't really care that much about preview flag words. I'd consult with @egilmorez or @jhuleatt if you think we will eventually expose this feature flag to our customers.)

Here's my definition:

**no-op deploy**: Feature to detect and skip deployment of functions that hasn't changed since last deployment

So to me, skipping 'no-op deploy' means flag to disable no-op deploy which is why it's confusing to me unless this is actually what you are trying to do here!

taeold (approved):

bkendall (approved):
This makes sense now :)

PR #4902, opened 2022-08-18
Gracefully close rules runtime on storage emulator stop

tonyjhuang:
### Description

Prevents the Firebase Rules runtime process exit signal from throwing an Error and preventing data export from happening on emulator stoppage.

Fixes https://github.com/firebase/firebase-tools/issues/3799

Removes `/internal/setRules` endpoint since file watching fulfills the same purpose.


etienne-ditchlabs:
By removing the endpoint, it is breaking: https://github.com/firebase/firebase-js-sdk/blob/df8604e4dada5874ce0513e3376ded130b6231b4/packages/rules-unit-testing/src/impl/rules.ts#L75

PR #4904, opened 2022-08-18
newly defined params should be visible in process.env to user code

Berlioz:
welp

I think it's correct to not do anything special in the emulator, since the emulator runs in non-interactive mode, so any attempt to have defined a new param through there would result in a throw anyways? Is that correct?

### DIFF
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -112,19 +112,32 @@ export async function prepare(
     const userEnvs = functionsEnv.loadUserEnvs(userEnvOpt);
     const envs = { ...userEnvs, ...firebaseEnvs };
     const wantBuild: build.Build = await runtimeDelegate.discoverBuild(runtimeConfig, firebaseEnvs);
-    const wantBackend: backend.Backend = await build.resolveBackend(
+    const { backend: wantBackend, envs: resolvedEnvs } = await build.resolveBackend(
       wantBuild,
       userEnvOpt,
       userEnvs,
       options.nonInteractive
     );
+
+    let hasEnvsFromParams = false;
     wantBackend.environmentVariables = envs;
+    for (const envName of Object.keys(resolvedEnvs)) {
+      const envValue = resolvedEnvs[envName]?.toString();
+      if (
+        envValue &&
+        !Object.prototype.hasOwnProperty.call(wantBackend.environmentVariables, envName)
+      ) {
+        wantBackend.environmentVariables[envName] = envValue;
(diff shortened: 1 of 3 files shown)


src/emulator/functionsEmulator.ts
```diff
@@ -513,7 +513,8 @@ export class FunctionsEmulator implements EmulatorInstance {
         projectAlias: this.args.projectAlias,
       };
       const discoveredBuild = await runtimeDelegate.discoverBuild(runtimeConfig, environment);
-      const discoveredBackend = await resolveBackend(discoveredBuild, userEnvOpt, environment);
+      const discoveredBackend = (await resolveBackend(discoveredBuild, userEnvOpt, environment))
```
taeold, line 516: nit* I think this could also work:

```
const { backend: discoveredBackend } = await resolveBackend(discoveredBuild, userEnvOpt, environment)`
```

Not 100% convinced I like this more than just

```
const resolution = await resolveBackend(discoveredBuild, userEnvOpt, environment)`
const discoveredBackend = resolution.backend
```

but `(await foo()).bar` always bothered me a little.

src/deploy/functions/prepare.ts
```diff
@@ -112,19 +112,32 @@ export async function prepare(
...
     const wantBuild: build.Build = await runtimeDelegate.discoverBuild(runtimeConfig, firebaseEnvs);
-    const wantBackend: backend.Backend = await build.resolveBackend(
+    const { backend: wantBackend, envs: resolvedEnvs } = await build.resolveBackend(
       wantBuild,
       userEnvOpt,
       userEnvs,
       options.nonInteractive
     );
+
+    let hasEnvsFromParams = false;
```
taeold, line 122: see L175 - you might wan to add tracking to see how much usage param is getting.

src/deploy/functions/prepare.ts
```diff
@@ -112,19 +112,32 @@ export async function prepare(
...
     wantBackend.environmentVariables = envs;
+    for (const envName of Object.keys(resolvedEnvs)) {
+      const envValue = resolvedEnvs[envName]?.toString();
```
taeold, line 125: dang TS can't deduce that `resolveEnvs[envName]` will never be undefined?

taeold (approved):
just some nits.

src/emulator/functionsEmulator.ts
TheIronDev: +1, tossing it in a variable makes it easier to read.

TheIronDev (approved):

Issue #3528, opened 2021-06-23
Make JSON schema more usable


samtstern:
Follow up to https://github.com/firebase/firebase-tools/pull/3505

TODO:

- [ ] Publish the schema somewhere
- [ ] Include `$schema` in `firebase.json` templates
- [ ] Make schema errors more visible in the logs

samtstern:
@abeisgoat think we could publish the schema on `firebase.tools`?

segevfiner:
You also need to allow "$schema" in the schema itself.

![image](https://user-images.githubusercontent.com/24731903/186179752-6d52ed7f-ad14-459a-b01a-d184e787f364.png)

dominicbartl:
@samtstern I created an updated PR to allow the top-level `$schema` property. It would be awesome if you could review this very simple change or add someone else as a reviewer. :smiley:

joehan:
Happened to be working on this, but never saw this issue. - thanks for highlighting it @dominicbartl! 

We've also added this to the[ JSON schema store](https://www.schemastore.org/json/), so many IDEs should pick it up by default (or once you add the necessary plugins): https://github.com/SchemaStore/schemastore/blob/master/src/api/json/catalog.json#L1615

PR #4919, opened 2022-08-23
Allow top-level $schema property

segevfiner:
### Description

Related to https://github.com/firebase/firebase-tools/issues/3528

This allows specifying the schema for validation and completion like so:
```json
  "$schema": "./node_modules/firebase-tools/schema/firebase-config.json",
```

Without getting an error about `$schema`.

### DIFF
--- a/schema/firebase-config.json
+++ b/schema/firebase-config.json
@@ -8,6 +8,9 @@
         }
     },
     "properties": {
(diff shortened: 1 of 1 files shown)


cedvdb:
@joehan no one is assigned for review here

PR #4922, opened 2022-08-25
Metric for param usage

Berlioz:
Based on a suggestion in #4904. Seperate PR because I have a question: is adding this one line enough to define a new metric, or do I need to edit a configuration file somewhere else?


taeold (approved):

src/deploy/functions/prepare.ts
```diff
@@ -142,6 +142,11 @@ export async function prepare(
     if (functionsEnv.hasUserEnvs(userEnvOpt) || hasEnvsFromParams) {
       codebaseUsesEnvs.push(codebase);
     }
+
+    if (wantBuild.params.length > 0) {
+      // TODO(vsfan@): distinguish between params using secrets and those without
+      void track("functions_params_in_build", "env_only");
```
inlined, line 148: Are we logging the metrics here because we don't have params saved in the backend?

Typically we do all our metrics publishing in deploy/release/reporter.ts where we can report a bunch of metrics in parallel and wait for their completion. It bothers me to see that we've started casting our metrics promises to void and dropping them on the floor. @bkendall does this bother you as well, or am I worrying about something de jour? If the former, maybe we need to start saving these promises in `pendingMetricsWrites` array and `await Promise.all(pendingMetricsWrites)` at the end of the method (they should be all resolved or near resolution since we're doing async IAM stuff later in this function).

Also, if I were to design metrics as I wanted to use them, I'd consider a "secret param" something totally different. I'd maybe have "data_params" event with a lable for the bucket of how many params were being used: 0, 1, 2, 3, 4, 5+.
Then I'd have "secret_params" with the same buckets and maaaaaybe "nonparam_secrets" that would count the total number of referenced secrets that weren't developed using params"

inlined: Also, what I'd really love to know is how many functions were configured by parameters; that's the killer feature, right? That might require a metrics publish in resolveBackend though.

Issue #4797, opened 2022-07-29
Version 11.2.2 and higher do not attach secrets to functions upon deploy when using Firebase Functions SDK <v.3.20.0


jacobsandlund:
### [REQUIRED] Environment info

**firebase-tools: 11.2.2,11.3.0 or 11.4.0 (has bug)**

**firebase-tools: 10.9.2 and 11.1.0 succeeds (no bug)**

**Platform: macOS**

### [REQUIRED] Test case

Have a function with `runWith` `secrets`:
```
export const playground = functions
  .runWith({
    secrets: [
      'TEST_SECRET',
    ],
  })
  .https.onRequest(app());
```

### [REQUIRED] Steps to reproduce

Deploy the function:

```
firebase deploy --project projectId --config firebase.json --only functions:playground
```

### [REQUIRED] Expected behavior

The secret should be attached to the function, as it is in 10.9.2 and 11.1.0.

### [REQUIRED] Actual behavior

There is no log about attaching the TEST_SECRET in the buggy versions when deploying, whereas 10.9.2 and 11.1.0 do attach the secret. Inside the function, inspecting `process.env.TEST_SECRET` shows undefined in the buggy versions.

taeold:
Hi @jacobsandlund - apologies for the issue you are seeing.

Thank you for sharing your experiments. You are right - we shipped a bug.

While we work on getting a hotfix out, I suggest you try updating the `firebase-functions` to the latest version. It seems like the bug only exists on older versions of the functions SDK.

Schmale97:
I am on firebase-functions version 3.22.0 and firebase-tools versions 11.4.0 and my secrets aren't being picked up when I emulate the cloud functions

taeold:
@Schmale97 Do you mind sharing a quick example of your setup?

I tried to reproduce your issue using the latest Firebase CLI (11.4.2) and Functions SDK (3.22.0) and running the emulator, but things seemed to work on my setup:

```
const functions = require("firebase-functions");

exports.helloSecrets = functions.runWith({ secrets: ["MY_SECRET"] }).https.onRequest((request, response) => {
  functions.logger.info("MY_SECRET!", {secret: process.env.MY_SECRET});
  response.send("Hello from Firebase!");
 });
``` 

```
$ firebase emulators:start
i  functions: Beginning execution of "helloSecrets"
>  {"secret":"*****","severity":"INFO","message":"MY_SECRET!"}
```

Schmale97:
[...]

PR #4926, opened 2022-08-26
Fix bug where secrets were not parsed when using old firebase functions sdk versions

taeold:
Partially fixes https://github.com/firebase/firebase-tools/issues/4797


src/deploy/functions/runtimes/node/parseTriggers.ts
```diff
@@ -140,6 +140,9 @@ export function useStrategy(): Promise<boolean> {
   return Promise.resolve(true);
 }
 
+/**
+ * Parse trigger annotations in sourceDir to generate backed.Build.
```
TheIronDev, line 144: nit: `{@link backend.Build}`

+others

src/deploy/functions/runtimes/node/parseTriggers.ts
```diff
@@ -377,12 +389,24 @@ export function addResourcesToBuild(
     "timeout",
     proto.secondsFromDuration
   );
+  if (annotation.secrets) {
+    const secretEnvs: backend.SecretEnvVar[] = [];
+    for (const secret of annotation.secrets) {
```
TheIronDev, line 394: nit: Why not:

```
const secretEnvs: backend.SecretEnvVar[] = annotation.secrets.map((secret) => ({
        secret,
        projectId,
        key: secret,
      }));
```

TheIronDev (approved):
LGTM,
Can you also write a test for this?

src/deploy/functions/runtimes/node/parseTriggers.ts
taeold: We don't use api-documenter for CLI 🤔 Do you think that annotation is useful regardless?

src/deploy/functions/runtimes/node/parseTriggers.ts
taeold: ctrl+c/ctrl+v 😛 (see few lines below). I think this one -changed.

colerogers (approved):

PR #4928, opened 2022-08-26
Set a unique exit code on the error when publishing extension with version lower than current

pavelgj:


src/extensions/extensionsHelper.ts
```diff
@@ -532,7 +532,8 @@ export async function publishExtensionVersionFromLocalSource(args: {
         `${args.publisherId}/${args.extensionId}`
       )}'. Please make sure this version is greater than the current version (${clc.bold(
         extension.latestVersion
-      )}) inside of extension.yaml.\n`
+      )}) inside of extension.yaml.\n`,
+      { exit: 104 }
```
huangjeff5, line 536: How was 104 picked? Should we move this to a constant somewhere?

pavelgj: 103 is used below... 104 is unused. Single use constants are usually frowned upon.

huangjeff5 (approved):

PR #4931, opened 2022-08-29
Rules emulator improvements

sam-gc:
This makes rules features available in emulator


src/emulator/storage/rules/types.ts
```diff
@@ -20,8 +26,10 @@ export interface SourceFile {
 }
 
 export interface RuntimeActionResponse {
-  id: number;
+  id?: number;
+  server_request_id?: number;
```
bkendall, line 30: is `server_request_id` dictated by what the JAR returns or something? We don't use underscore-case except in some very limited cases. rulesVersion below, which extends this interface, seems to work - would serverRequestId work here?

bkendall (commented):
Looks like the storage-emulator-integration tests are failing to run. Should take a look into that as well.

sam-gc:
Whoops, updated tests

scripts/storage-emulator-integration/rules/manager.test.ts
```diff
@@ -21,6 +21,7 @@ const EMULATOR_LOAD_RULESET_DELAY_MS = 2000;
 describe("Storage Rules Manager", function () {
   const rulesRuntime = new StorageRulesRuntime();
   const opts = { method: RulesetOperationMethod.GET, file: {}, path: "/b/bucket_0/o/" };
+  const projectId = "project-id";
```
yuchenshi, line 24:
```suggestion
  const projectId = "demo-project-id";
```

scripts/storage-emulator-integration/rules/runtime.test.ts
```diff
@@ -367,7 +367,7 @@ async function testIfPermitted(
   }
 
   const { permitted, issues } = await loadResult.ruleset.verify(
-    verificationOpts,
+    { ...verificationOpts, projectId: "project-id" },
```
yuchenshi, line 370:
```suggestion
    { ...verificationOpts, projectId: "demo-project-id" },
```

src/emulator/storage/rules/runtime.ts
```diff
@@ -318,7 +338,29 @@ export class StorageRulesRuntime {
...
-    const response = (await this._sendRequest(runtimeActionRequest)) as RuntimeActionVerifyResponse;
+
+    return this._completeVerifyWithRuleset(opts.projectId, runtimeActionRequest);
+  }
+
+  private async _completeVerifyWithRuleset(
+    projectId: string,
+    runtimeActionRequest: RuntimeActionBundle,
+    overrideId?: number
+  ): Promise<
```
yuchenshi, line 349: This returns a Promise of a Promise. Please unnest this.

yuchenshi (approved):

christhompsongoogle (approved):

src/emulator/storage/rules/runtime.ts
```diff
@@ -318,7 +338,27 @@ export class StorageRulesRuntime {
...
+  ): Promise<{
+    permitted?: boolean;
+    issues: StorageRulesIssues;
+  }> {
+    const response = (await this._sendRequest(
+      runtimeActionRequest,
+      overrideId
+    )) as RuntimeActionVerifyResponse;
+
+    if ("context" in response) {
```
tonyjhuang, line 358: Where does this string come from? can we change this to a constant

sam-gc: It's a named key and here used for type narrowing:
https://github.com/firebase/firebase-tools/pull/4931/files/e4ec6a0bfda1cdb917262919a6ed53f274be9858#diff-caf325e78d96d4559be60aff9779ca8a8090299c4078d6366eae20deb9322eb3R52-R55

Since it's a key in the type and used for narrowing, I'd prefer not moving it to a const, even though that does technically work per the type checker. This is just a language quirk but it's essentially no different than writing `response.context`

PR #4937, opened 2022-08-31
Add v2 Remote Config triggers to deploy

colerogers:
```
```

### DIFF
--- a/src/test/deploy/functions/services/remoteConfig.spec.ts
+++ b/src/test/deploy/functions/services/remoteConfig.spec.ts
@@ -1,0 +1,47 @@
+import { expect } from "chai";
+import { Endpoint } from "../../../../deploy/functions/backend";
+import * as remoteConfig from "../../../../deploy/functions/services/remoteConfig";
+
+const projectNumber = "123456789";
+
+const endpoint: Endpoint = {
+  id: "endpoint",
+  region: "us-central1",
+  project: projectNumber,
+  eventTrigger: {
+    retry: false,
+    eventType: "google.firebase.remoteconfig.remoteConfig.v1.updated",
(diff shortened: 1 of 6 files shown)


TheIronDev (approved):
LGTM

src/test/deploy/functions/services/remoteConfig.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+    await remoteConfig.ensureRemoteConfigTriggerRegion(ep);
+
+    expect(ep.eventTrigger.region).to.eq("global");
+  });
+
+  it("should error if the trigger location is not global", () => {
+    const ep = { ...endpoint };
+    ep.eventTrigger.region = "us-west1";
+
+    expect(() => remoteConfig.ensureRemoteConfigTriggerRegion(ep)).to.throw(
```
taeold, line 43: Hm I'm surprise this passes. I didn't expect this to work since `ensureRemoteConfigTriggerRegion` is an async function an dyou are never awaiting it.

A proper way to test this would be to check the condition `to.be.eventually.rejected`...

taeold (approved):
lgtm minor comment about test case

src/test/deploy/functions/services/remoteConfig.spec.ts
colerogers: So `ensureRemoteConfigTriggerRegion` isn't an async function, we just return a void promise to satisfy the interface in `services/index`. We are actually throwing an error, not rejecting a promise, so the function isn't `thenable` and can't use the `eventually.rejected` syntax. We also do this for `database` & `firebasealerts` functions

PR #4940, opened 2022-08-31
[Function-NoOp] Update to sha1 and fix source hashing

TheIronDev:
### Description

This commit fixes 2 bugs found during E2E testing of Skipping No-Op Functions.

2 Problems were discovered during local testing:

1. Using a sha256 exceeded the length requirements of labels. [More Info](https://cloud.google.com/resource-manager/docs/creating-managing-labels#requirements)
1. The package source file was different every time due to `.runtimeconfig.json` regeneration and metadata mismatching.

### Testing

Enable the preview feature

```
$ node path/to/bin/firebase.js --open-sesame skipdeployingnoopfunctions
```

Deploy some code

```
$ node path/to/bin/firebase.js deploy
```

Deploy some code again, without changes

```
$ node path/to/bin/firebase.js deploy
```

Deploy some code again, after changing some code

```
$ node path/to/bin/firebase.js deploy
```

### Scenarios Tested

- [x] Uploading Function that did not previously have a hash - should update
- [x] Uploading Function whose local hash matches the server copy - should skip
- [x] Uploading Function whose local hash does not match the server copy - should update

### Sample Commands

```
$ firebase --open-sesame skipdeployingnoopfunctions
$ firebase deploy
```

### DIFF
--- a/src/deploy/functions/args.ts
+++ b/src/deploy/functions/args.ts
@@ -14,8 +14,6 @@ export interface Source {
   // Filled in the "prepare" phase.
   functionsSourceV1?: string;
   functionsSourceV2?: string;
-
-  // Filled in the "prepare" phase.
   functionsSourceV1Hash?: string;
   functionsSourceV2Hash?: string;
 
--- a/src/deploy/functions/args.ts
+++ b/src/deploy/functions/args.ts
@@ -15,6 +15,10 @@ export interface Source {
   functionsSourceV1?: string;
   functionsSourceV2?: string;
 
+  // Filled in the "prepare" phase.
+  functionsSourceV1Hash?: string;
+  functionsSourceV2Hash?: string;
+
   // Filled in the "deploy" phase.
   sourceUrl?: string;
   storage?: gcfV2.StorageSource;
--- a/src/deploy/functions/cache/applyHash.ts
+++ b/src/deploy/functions/cache/applyHash.ts
@@ -1,30 +1,24 @@
 import { Backend, allEndpoints } from "../backend";
 import * as args from "../args";
-import {
-  getEndpointHash,
-  getEnvironmentVariablesHash,
-  getSecretsHash,
-  getSourceHash,
-} from "./hash";
+import { getEndpointHash, getEnvironmentVariablesHash, getSecretsHash } from "./hash";
 
 /**
  *
  * Updates all the CodeBase {@link Backend}, applying a hash to each of their {@link Endpoint}.
  */
-export async function applyBackendHashToBackends(
+export function applyBackendHashToBackends(
   wantBackends: Record<string, Backend>,
   context: args.Context
-): Promise<void> {
+): void {
   for (const [codebase, wantBackend] of Object.entries(wantBackends)) {
     const source = context?.sources?.[codebase]; // populated earlier in prepare flow
-    const sourceV1Hash = source?.functionsSourceV1
-      ? await getSourceHash(source?.functionsSourceV1)
-      : undefined;
-    const sourceV2Hash = source?.functionsSourceV2
-      ? await getSourceHash(source?.functionsSourceV2)
-      : undefined;
     const envHash = getEnvironmentVariablesHash(wantBackend);
-    applyBackendHashToEndpoints(wantBackend, envHash, sourceV1Hash, sourceV2Hash);
+    applyBackendHashToEndpoints(
+      wantBackend,
+      envHash,
+      source?.functionsSourceV1Hash,
+      source?.functionsSourceV2Hash
+    );
   }
 }
 
--- a/src/deploy/functions/cache/applyHash.ts
+++ b/src/deploy/functions/cache/applyHash.ts
@@ -19,12 +19,10 @@ export function applyBackendHashToBackends(
     }
     const source = context?.sources?.[codebase]; // populated earlier in prepare flow
     const envHash = getEnvironmentVariablesHash(wantBackend);
-    const filtersFilteredByCodebase =
-      context.filters?.filter((filter) => filter.codebase === codebase) || [];
     applyBackendHashToEndpoints(
       wantBackend,
       envHash,
-      filtersFilteredByCodebase,
+      context.filters || [],
       source?.functionsSourceV1Hash,
       source?.functionsSourceV2Hash
     );
--- a/src/deploy/functions/cache/applyHash.ts
+++ b/src/deploy/functions/cache/applyHash.ts
@@ -1,6 +1,7 @@
 import { Backend, allEndpoints } from "../backend";
 import * as args from "../args";
 import { getEndpointHash, getEnvironmentVariablesHash, getSecretsHash } from "./hash";
+import { EndpointFilter, isCodebaseFiltered, isEndpointFiltered } from "../functionsDeployHelper";
 
 /**
  *
@@ -11,11 +12,19 @@ export function applyBackendHashToBackends(
   context: args.Context
 ): void {
   for (const [codebase, wantBackend] of Object.entries(wantBackends)) {
+    // If an entire codebase is filtered, then don't set the hash for the functions in the codebase.
+    // This effectively forces all the functions to deploy without the duplication check.
+    if (isCodebaseFiltered(codebase, context.filters || [])) {
+      continue;
+    }
     const source = context?.sources?.[codebase]; // populated earlier in prepare flow
     const envHash = getEnvironmentVariablesHash(wantBackend);
+    const filtersFilteredByCodebase =
+      context.filters?.filter((filter) => filter.codebase === codebase) || [];
     applyBackendHashToEndpoints(
       wantBackend,
       envHash,
+      filtersFilteredByCodebase,
       source?.functionsSourceV1Hash,
       source?.functionsSourceV2Hash
     );
@@ -28,6 +37,7 @@ export function applyBackendHashToBackends(
 function applyBackendHashToEndpoints(
   wantBackend: Backend,
   envHash: string,
+  endpointFilters: EndpointFilter[],
   sourceV1Hash?: string,
   sourceV2Hash?: string
 ): void {
@@ -35,6 +45,10 @@ function applyBackendHashToEndpoints(
     const secretsHash = getSecretsHash(endpoint);
     const isV2 = endpoint.platform === "gcfv2";
     const sourceHash = isV2 ? sourceV2Hash : sourceV1Hash;
+    // If the endpoint is in the filtered list, then skip setting a hash (effectively forcing a deploy).
+    if (isEndpointFiltered(endpoint, endpointFilters)) {
+      continue;
+    }
     endpoint.hash = getEndpointHash(sourceHash, envHash, secretsHash);
   }
 }
--- a/src/deploy/functions/cache/hash.ts
+++ b/src/deploy/functions/cache/hash.ts
@@ -8,7 +8,7 @@ import { getSecretVersions } from "../../../functions/secrets";
  * @param backend Backend of a set of functions
  */
 export function getEnvironmentVariablesHash(backend: Backend): string {
-  const hash = crypto.createHash("sha256");
+  const hash = crypto.createHash("sha1");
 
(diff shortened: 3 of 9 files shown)


src/deploy/functions/prepareFunctionsUpload.ts
```diff
@@ -102,13 +112,14 @@ async function packageSource(
       filesize(archive.pointer()) +
       ") for uploading"
   );
-  return tmpFile;
+  const hash = fileHashes.join(".");
+  return { pathToSource: tmpFile, hash: hash };
```
colerogers, line 116: Couldn't we shorten this to `return { pathToSource: tmpFile, hash };`

colerogers (approved):

src/deploy/functions/args.ts
```diff
@@ -15,6 +15,10 @@ export interface Source {
   functionsSourceV1?: string;
   functionsSourceV2?: string;
 
+  // Filled in the "prepare" phase.
+  functionsSourceV1Hash?: string;
+  functionsSourceV2Hash?: string;
```
taeold, line 20: nit* I see that we are repeating "Filled in the "prepare" phase." line from 14, and could see these new properties bunched together with what we have w/ L15-16.

src/deploy/functions/cache/hash.ts
```diff
@@ -8,7 +8,7 @@ import { getSecretVersions } from "../../../functions/secrets";
  * @param backend Backend of a set of functions
  */
 export function getEnvironmentVariablesHash(backend: Backend): string {
-  const hash = crypto.createHash("sha256");
+  const hash = crypto.createHash("sha1");
```
taeold, line 11: nit* Does `sha1` HAVE to be consistent across all env vars? Just wondering if we should just define a helper function in this module to make sure we don't repeat `crypto.createHash("sha1");` again and again and instead just call `createHash()` helper function.

src/deploy/functions/args.ts
taeold: We could, also, create a `functionSource` object with these properties:

source: {
  v1: {
    source: String
    hash: String
  }
}

Just a thought - up to you to decide if this is cleaner in any way.

src/deploy/functions/prepareFunctionsUpload.ts
taeold, line 116: Can we add return type for packageSource fn?

src/deploy/functions/prepareFunctionsUpload.ts
```diff
@@ -102,13 +112,14 @@ async function packageSource(
       filesize(archive.pointer()) +
       ") for uploading"
   );
-  return tmpFile;
+  const hash = fileHashes.join(".");
```
taeold, line 115: woah isn't this hash like really long? One hash per every file joined via "."?

src/deploy/functions/prepareFunctionsUpload.ts
```diff
@@ -71,8 +78,11 @@ async function packageSource(
   try {
     const files = await fsAsync.readdirRecursive({ path: sourceDir, ignore: ignore });
     for (const file of files) {
+      const name = path.relative(sourceDir, file.name);
+      const fileHash = await getSourceHash(file.name);
+      fileHashes.push(fileHash);
```
taeold, line 83: nit* This seems a bit inefficient given that we have to read the file from disk twice (once to generate a hash, another by the archiver itself.

Can't we create a hash from archived zip once?

taeold: offline discussion - no b/c metadata changes the hash of the archive...

src/deploy/functions/cache/hash.ts
```diff
@@ -24,7 +24,7 @@ export function getEnvironmentVariablesHash(backend: Backend): string {
  * @param pathToGeneratedPackageFile Packaged file contents of functions
  */
 export async function getSourceHash(pathToGeneratedPackageFile: string): Promise<string> {
-  const hash = crypto.createHash("sha256");
+  const hash = crypto.createHash("sha1");
 
   // Hash the contents of the source file
```
taeold, line 29: why is this still reading the archived file - not the hash we generated in prepareFunctionUpload?

src/deploy/functions/prepareFunctionsUpload.ts
TheIronDev: Good catch!

src/deploy/functions/args.ts
TheIronDev: Removed the redundant comment.

As for `functionSource` object. While I like the object, I'd like to minimize the # of places I touch in this PR.

src/deploy/functions/cache/hash.ts
TheIronDev: Helper method it is!

src/deploy/functions/prepareFunctionsUpload.ts
TheIronDev: Updated!

src/deploy/functions/prepareFunctionsUpload.ts
TheIronDev: 🙃 At this point yeah its a super long string. By the time we get to storing it on the endpoint we will sha1 the long string into a shorter string.

An alternative is to pass around something like: `interface FileHash  { filename: string; hash: string}`.

WDYT?

src/deploy/functions/cache/hash.ts
TheIronDev: Ah good catch! The comments are misleading. We aren't hashing the archive anymore, we are hashing the files contained within the archive.

Updated comments!

src/deploy/functions/prepareFunctionsUpload.ts
TheIronDev: Unfortunately we can't hash the archived zip due to metadata. 🐼

src/deploy/functions/prepareFunctionsUpload.ts
taeold: hmm I thought you'd create an "intermediate" hash in this function:

```js
  const hash = crypto.createHash("sha1"); // or md5 or whatever
...
  hash.update(fileHash);
```

Hash of a hash of a hash of a hash of a...

To think about it - probably doesn't matter too much whether we carry around a hash or some really long string 🤷🏼.

src/deploy/functions/cache/applyHash.ts
```diff
@@ -1,30 +1,33 @@
...
-    const sourceV1Hash = source?.functionsSourceV1
-      ? await getSourceHash(source?.functionsSourceV1)
-      : undefined;
-    const sourceV2Hash = source?.functionsSourceV2
-      ? await getSourceHash(source?.functionsSourceV2)
-      : undefined;
     const envHash = getEnvironmentVariablesHash(wantBackend);
-    applyBackendHashToEndpoints(wantBackend, envHash, sourceV1Hash, sourceV2Hash);
+    const filtersFilteredByCodebase =
+      context.filters?.filter((filter) => filter.codebase === codebase) || [];
```
taeold, line 23: IIRC, when codebase === undefined, it should match the idchunks against all codebases:

https://github.com/firebase/firebase-tools/blob/2f8f0cb2f1660c18ca888c37ef705cc558147114/src/deploy/functions/functionsDeployHelper.ts#L5-L7

We should either include filters whose codebase is undefined or... not do this filtering at all? I'm not 100% sure if this filtering is necessary.

src/deploy/functions/cache/hash.ts
```diff
@@ -2,52 +2,37 @@ import { readFile } from "node:fs/promises";
 import * as crypto from "crypto";
 import { Backend, Endpoint } from "../backend";
 import { getSecretVersions } from "../../../functions/secrets";
+import { BinaryLike } from "crypto";
```
taeold, line 5: nit* prefer grouping std lib imports

src/deploy/functions/cache/hash.ts
```diff
@@ -2,52 +2,37 @@ import { readFile } from "node:fs/promises";
...
-  const hash = crypto.createHash("sha256");
-
   // Hash the contents of the dotenv variables
   const hasEnvironmentVariables = !!Object.keys(backend.environmentVariables).length;
-  if (hasEnvironmentVariables) {
-    hash.update(JSON.stringify(backend.environmentVariables));
-  }
-
-  return hash.digest("hex");
+  return createHash(hasEnvironmentVariables ? JSON.stringify(backend.environmentVariables) : "");
```
taeold, line 14: nit* hasEnviromentVariable feels slightly less useful w/ the refactored code. Seems like it's equilvalent to write something like

```
return createHash(JSON.stringify(backend.environmentVariables || {}))
```

(though this might break your tests. I think it's cool to ignore this comment if this feels too picky)

src/deploy/functions/cache/hash.ts
```diff
@@ -2,52 +2,37 @@ import { readFile } from "node:fs/promises";
...
-export async function getSourceHash(pathToGeneratedPackageFile: string): Promise<string> {
-  const hash = crypto.createHash("sha256");
-
-  // Hash the contents of the source file
-  const data = await readFile(pathToGeneratedPackageFile);
-  hash.update(data);
-
-  return hash.digest("hex");
+export async function getSourceHash(pathToFile: string): Promise<string> {
+  // Hash the contents of a file
```
taeold, line 22:
```suggestion
  // Hash the contents of a file, ignoring metadata.
  // Excluding metadata in the hash is important because some
  // files are dynamically generated on deploy.
```

src/deploy/functions/cache/hash.ts
```diff
@@ -2,52 +2,37 @@ import { readFile } from "node:fs/promises";
...
-
   // Hash the secret versions.
   const secretVersions = getSecretVersions(endpoint);
   const hasSecretVersions = !!Object.keys(secretVersions).length;
-  if (hasSecretVersions) {
-    hash.update(JSON.stringify(secretVersions));
-  }
-
-  return hash.digest("hex");
+  return createHash(hasSecretVersions ? JSON.stringify(secretVersions) : "");
```
taeold, line 35: similar comment to removing `hasEnvironmentVariable`.

src/deploy/functions/functionsDeployHelper.ts
```diff
@@ -204,3 +204,17 @@ export function groupEndpointsByCodebase(
   // defined in other project repositories.
   return grouped;
 }
+
+/** Checks if a codebase should be filtered */
+export function isCodebaseFiltered(codebase: string, filters: EndpointFilter[]) {
```
taeold, line 209:
```suggestion
export function isCodebaseFiltered(codebase: string, filters: EndpointFilter[]): boolean {
```

taeold (approved):
One possible bug, some nits.

As long as we address the possible bug I pointed out, 👍🏼 on merging so stamping an early approval.

src/deploy/functions/prepareFunctionsUpload.ts
TheIronDev: What gets passed around doesn't matter too much, as long as the final hash we store as a label isn't large. 🤷

src/deploy/functions/cache/applyHash.ts
TheIronDev: Thats a very good point (and it also causes the bug that you mentioned.) Removed the filtered filter, which fixed the bug (verified with TDD).

src/deploy/functions/cache/hash.ts
TheIronDev: Updated!

src/deploy/functions/cache/hash.ts
TheIronDev: Updated!

src/deploy/functions/cache/hash.ts
TheIronDev: You're right, `JSON.stringify({})` isn't any more or less useful than `""`.  Updated! (And tests didn't break!)

PR #4943, opened 2022-08-31
Full-ish featured CEL evaluation for param defaults/option values.

Berlioz:
This is about as far as the regular expression based parsing approach is going to take us. My personal opinion is that this is good enough for a MVP release, mostly because going into nested expressions will require a full AST-based shift-reduce parser and I'm still living in denial about having to write one eventually.

Handles:
- {{ params.foo }}
- {{ params.foo == 24 }}
- {{ params.foo == params.bar }}
- {{ params.foo == 24 ? "asdf" : "jkl;" }}
- {{ params.foo ? "asdf" : "jkl;" }}, when foo is of boolean type
- type safety
- the special case of interpolating into a string
Does not handle:
- {{ params.foo == params.bar ? "asdf" : "jkl;" }}
- Nesting expressions with boolean type into the conditional of a ternary operator in general


src/deploy/functions/params.ts
```diff
@@ -16,7 +17,7 @@ interface ListItem {
 type CEL = build.Expression<string> | build.Expression<number> | build.Expression<boolean>;
 
 function isCEL(expr: string | number | boolean): expr is CEL {
-  return typeof expr === "string" && expr.includes("{{") && expr.includes("}}");
+  return isCelExpression(expr);
 }
```
taeold, line 21: nit* doesn't seem like a useful helper fn anymore.

src/deploy/functions/params.ts
```diff
@@ -66,29 +54,22 @@ export function resolveString(
   from: string | build.Expression<string>,
```
taeold, line 54: looks like we should update the docstring for this fn. Likewise for other `resolve` functions

src/deploy/functions/params.ts
```diff
@@ -66,29 +54,22 @@ export function resolveString(
...
-    if (!referencedParamValue.legalString) {
+  while ((match = celCapture.exec(output)) != null) {
+    infiniteLoopGuard += 1;
+    if (infiniteLoopGuard > 100) {
       throw new FirebaseError(
-        "Referenced string parameter '" +
-          match[1] +
-          "' resolved to non-string value " +
-          referencedParamValue
+        `CEL string expression ${from} spectacularily broke the CEL parser; we would very much appreciate you opening a Github issue.`
```
taeold, line 65: lol is this too cute of an error message?

We should at least include why we think it's broken - e.g. too many nested params?

src/deploy/functions/params.ts
```diff
@@ -66,29 +54,22 @@ export function resolveString(
...
-  const paramCapture = /{{ params\.(\w+) }}/g;
+  const celCapture = /{{ (.+?) }}/;
+  let infiniteLoopGuard = 0;
   let match: RegExpMatchArray | null;
-  while ((match = paramCapture.exec(from)) != null) {
-    const referencedParamValue = paramValues[match[1]];
-    if (!referencedParamValue.legalString) {
+  while ((match = celCapture.exec(output)) != null) {
+    infiniteLoopGuard += 1;
+    if (infiniteLoopGuard > 100) {
```
taeold, line 63: q: why is it that `resolveString` is the only param that resolves nested CEL expressions?

src/deploy/functions/cel.ts
```diff
@@ -0,0 +1,345 @@
...
+  }
+}
+
+function assertType(wantType: L, paramName: string, paramValue: ParamValue) {
+  if (
+    (wantType === "string" && !paramValue.legalString) ||
+    (wantType === "number" && !paramValue.legalNumber) ||
+    (wantType === "boolean" && !paramValue.legalBoolean)
+  ) {
+    throw new ExprParseError(`illegal type coercion of param ${paramName} to type ${wantType}`);
```
taeold, line 96: nit* I think all FirebaseError messages are expected to be read by the end user and hence start with capital letters (like a sentence) rather than being an error message (like go)/

src/test/deploy/functions/cel.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+          FOO: boolV(true),
+        });
+      }).to.throw(ExprParseError);
+    });
+
+    it("pulls number parameters", () => {
+      const params: Record<string, ParamValue> = {
+        FOO: numberV(22),
+        BAR: boolV(true),
+        BAZ: new ParamValue("quox", false, { string: true }),
```
taeold, line 35: nit* `stringV("quox")`? or is this intentional?

src/test/deploy/functions/cel.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+        })
+      ).to.be.true;
+      expect(
+        resolveExpression("boolean", "{{ params.FOO == 11 }}", {
+          FOO: numberV(22),
+        })
+      ).to.be.false;
+      expect(
+        resolveExpression("boolean", '{{ params.FOO == "bar baz" }}', {
+          FOO: new ParamValue("bar baz", false, { string: true }),
```
taeold, line 72: nit* how is this test different from using `stringV` to define a string param?

src/test/deploy/functions/cel.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+      expect(
+        resolveExpression("boolean", "{{ params.FOO == true }}", {
+          FOO: boolV(true),
+        })
+      ).to.be.true;
+      expect(
+        resolveExpression("boolean", "{{ params.FOO == false }}", {
+          FOO: boolV(true),
+        })
+      ).to.be.false;
```
taeold, line 89: this feels sneaky since it doesn't match the test description (LHS === RHS).

src/test/deploy/functions/cel.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+          FOO: numberV(22),
+          BAR: numberV(22),
+        })
+      ).to.be.true;
+      expect(
+        resolveExpression("boolean", "{{ params.FOO == params.BAR }}", {
+          FOO: numberV(22),
+          BAR: numberV(11),
+        })
+      ).to.be.false;
```
taeold, line 137: also feels a bit sneaky since it doesn't match the test desription.

src/test/deploy/functions/cel.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+        });
+      }).to.throw(ExprParseError);
+      expect(() => {
+        resolveExpression("number", "{{ params.FOO == 22 ? params.BAZ : params.QUOZ }}", {
+          FOO: numberV(11),
+        });
+      }).to.throw(ExprParseError);
+    });
+
+    it("it provides resolved parameters when the ternary expression calls for them", () => {
```
taeold, line 201: Looking at these, I wish `>, >=, <, <=` were also supported, but I also don't know if there are usecase for it 🤷🏼

taeold (approved):
Awesome set of tests!

Frankly, I didn't look at the cel implementation too closely since unit test seemed pretty comprehensive. I will learn as we fix any bug that comes out of it.

src/test/deploy/functions/cel.spec.ts
Berlioz: Oops. I had them all typed out, decided it was a farce later, and did a find-replace with the helper functions. Must've missed this one.

src/test/deploy/functions/cel.spec.ts
Berlioz: Bad copy-paste job instead of intentionally sneaky, but yes, I'm fixing it

src/test/deploy/functions/cel.spec.ts
Berlioz: Bad copy-paste job instead of intentionally sneaky, but yes, I'm fixing it

src/deploy/functions/params.ts
Berlioz: I agree.

src/deploy/functions/params.ts
Berlioz: Done.

src/deploy/functions/params.ts
Berlioz: This is an escape hatch that users ~*~should never see~*~; it's there to prevent infinite loops in string interpolation which I can't see could ever happen

Berlioz: oh no

Berlioz: it could happen

src/test/deploy/functions/cel.spec.ts
Berlioz: Keep Tuned (TM)

src/test/deploy/functions/cel.spec.ts
Berlioz: it isnt

src/deploy/functions/params.ts
Berlioz: This is interpolation, not nesting. The revised version of this function should make that more clear.

src/deploy/functions/cel.ts
Berlioz: Got it. That makes sense.

src/deploy/functions/params.ts
Berlioz: Okay, it can't happen now, so we can get rid of this cute message.

PR #4945, opened 2022-09-01
[Function-NoOp] Implement an escape hatch

TheIronDev:
### Description

This commit implements the `--only` escape hatch to force functions
to be deployed, even if they are identical to the copy found on gcf.

### Scenarios Tested

- [x] No filters
- [x] Codebase filter
- [x] Codebase + endpoint ids filter

### DIFF
--- a/src/deploy/functions/cache/applyHash.ts
+++ b/src/deploy/functions/cache/applyHash.ts
@@ -1,6 +1,7 @@
 import { Backend, allEndpoints } from "../backend";
 import * as args from "../args";
 import { getEndpointHash, getEnvironmentVariablesHash, getSecretsHash } from "./hash";
+import { EndpointFilter } from "../functionsDeployHelper";
 
 /**
  *
@@ -11,14 +12,21 @@ export function applyBackendHashToBackends(
   context: args.Context
 ): void {
   for (const [codebase, wantBackend] of Object.entries(wantBackends)) {
-    const source = context?.sources?.[codebase]; // populated earlier in prepare flow
-    const envHash = getEnvironmentVariablesHash(wantBackend);
-    applyBackendHashToEndpoints(
-      wantBackend,
-      envHash,
-      source?.functionsSourceV1Hash,
-      source?.functionsSourceV2Hash
-    );
+    // If an entire codebase is filtered, then don't set the hash for the functions in the codebase.
+    // This effectively forces all the functions to deploy without the duplication check.
+    if (!isCodebaseFiltered(codebase, context.filters || [])) {
+      const source = context?.sources?.[codebase]; // populated earlier in prepare flow
+      const envHash = getEnvironmentVariablesHash(wantBackend);
+      const codebaseFilters =
+        context.filters?.filter((filter) => filter.codebase === codebase) || [];
+      applyBackendHashToEndpoints(
+        wantBackend,
+        envHash,
+        codebaseFilters,
+        source?.functionsSourceV1Hash,
+        source?.functionsSourceV2Hash
+      );
+    }
   }
 }
 
@@ -28,6 +36,7 @@ export function applyBackendHashToBackends(
 function applyBackendHashToEndpoints(
   wantBackend: Backend,
   envHash: string,
+  codebaseFilters: EndpointFilter[],
   sourceV1Hash?: string,
   sourceV2Hash?: string
 ): void {
@@ -35,6 +44,20 @@ function applyBackendHashToEndpoints(
     const secretsHash = getSecretsHash(endpoint);
     const isV2 = endpoint.platform === "gcfv2";
     const sourceHash = isV2 ? sourceV2Hash : sourceV1Hash;
-    endpoint.hash = getEndpointHash(sourceHash, envHash, secretsHash);
+    // If the endpoint is in the filtered list, then skip setting a hash (effectively forcing a deploy).
+    if (!isEndpointFiltered(endpoint.id, codebaseFilters)) {
+      endpoint.hash = getEndpointHash(sourceHash, envHash, secretsHash);
+    }
   }
+}
+
+function isCodebaseFiltered(codebase: string, filters: EndpointFilter[]) {
+  return filters.some(
+    (filter) =>
+      (!filter?.idChunks || filter?.idChunks?.length === 0) && filter.codebase === codebase
+  );
+}
+
+function isEndpointFiltered(id: string, filters: EndpointFilter[]) {
+  return filters.some((filter) => filter?.idChunks?.some((filterId) => filterId === id));
 }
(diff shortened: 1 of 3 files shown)


src/deploy/functions/cache/applyHash.ts
```diff
@@ -34,13 +36,28 @@ export async function applyBackendHashToBackends(
...
     const sourceHash = isV2 ? sourceV2Hash : sourceV1Hash;
-    endpoint.hash = getEndpointHash(sourceHash, envHash, secretsHash);
+    // If the endpoint is in the filtered list, then skip setting a hash (effectively forcing a deploy).
+    if (!isEndpointFiltered(endpoint.id, codebaseFilters)) {
+      endpoint.hash = getEndpointHash(sourceHash, envHash, secretsHash);
+    }
   }
 }
+
+function isCodebaseFiltered(codebase: string, filters: EndpointFilter[]) {
```
colerogers, line 54: This looks like it could live in `functionsDeployHelper.ts`, no?

src/deploy/functions/cache/applyHash.ts
```diff
@@ -34,13 +36,28 @@ export async function applyBackendHashToBackends(
...
+  return filters.some(
+    (filter) =>
+      (!filter?.idChunks || filter?.idChunks?.length === 0) && filter.codebase === codebase
+  );
+}
+
+function isEndpointFiltered(id: string, filters: EndpointFilter[]) {
```
colerogers, line 61: Same here, could it live in `functionsDeployHelper.ts`?

src/deploy/functions/cache/applyHash.ts
```diff
@@ -34,13 +36,28 @@ export async function applyBackendHashToBackends(
...
   sourceV1Hash?: string,
   sourceV2Hash?: string
 ): void {
   for (const endpoint of allEndpoints(wantBackend)) {
     const secretsHash = getSecretsHash(endpoint);
     const isV2 = endpoint.platform === "gcfv2";
     const sourceHash = isV2 ? sourceV2Hash : sourceV1Hash;
-    endpoint.hash = getEndpointHash(sourceHash, envHash, secretsHash);
+    // If the endpoint is in the filtered list, then skip setting a hash (effectively forcing a deploy).
+    if (!isEndpointFiltered(endpoint.id, codebaseFilters)) {
```
colerogers, line 48: We could move this to the top of the `for` loop and flip the `if` condition 
```
if (isEndpointFiltered(endpoint.id, codebaseFilters)) {
  continue;
}
```
Thoughts?

src/deploy/functions/cache/applyHash.ts
```diff
@@ -1,30 +1,32 @@
...
-      ? await getSourceHash(source?.functionsSourceV1)
-      : undefined;
-    const sourceV2Hash = source?.functionsSourceV2
-      ? await getSourceHash(source?.functionsSourceV2)
-      : undefined;
-    const envHash = getEnvironmentVariablesHash(wantBackend);
-    applyBackendHashToEndpoints(wantBackend, envHash, sourceV1Hash, sourceV2Hash);
+    // If an entire codebase is filtered, then don't set the hash for the functions in the codebase.
+    // This effectively forces all the functions to deploy without the duplication check.
+    if (!isCodebaseFiltered(codebase, context.filters || [])) {
```
colerogers, line 17: We could flip this to 
```
if (isCodebaseFiltered(codebase, context.filters || [])) {
  continue;
}
```
and remove the indented block, thoughts?

src/deploy/functions/cache/applyHash.ts
```diff
@@ -34,13 +36,28 @@ export async function applyBackendHashToBackends(
...
+    if (!isEndpointFiltered(endpoint.id, codebaseFilters)) {
+      endpoint.hash = getEndpointHash(sourceHash, envHash, secretsHash);
+    }
   }
 }
+
+function isCodebaseFiltered(codebase: string, filters: EndpointFilter[]) {
+  return filters.some(
+    (filter) =>
+      (!filter?.idChunks || filter?.idChunks?.length === 0) && filter.codebase === codebase
```
taeold, line 57:
```suggestion
      (!filter.idChunks || filter.idChunks.length === 0) && filter.codebase === codebase
```

I _think_ this works since you've already checked for existence of filter and `idChunks`?

taeold: Can we add a small comment on what each conditions do? It's taking me some time to parse this condition (I may be just a bit slow this Friday....)

IIUC - I think we want to know if any of the filter we have JUST sets the codebase property?

src/deploy/functions/cache/applyHash.ts
```diff
@@ -34,13 +36,28 @@ export async function applyBackendHashToBackends(
...
+  );
+}
+
+function isEndpointFiltered(id: string, filters: EndpointFilter[]) {
+  return filters.some((filter) => filter?.idChunks?.some((filterId) => filterId === id));
+}
```
taeold, line 63: I think you might want to juse use `endpointMatchesFilter` function.

There is a bit of nuance here w/ grouped function (https://firebase.google.com/docs/functions/organize-functions) that's a bit complicated, and this helper doesn't correctly compensate grouped function filters.

src/deploy/functions/cache/applyHash.ts
TheIronDev: The negative is making it harder to read. Updated!

src/deploy/functions/cache/applyHash.ts
TheIronDev: Good point, moved it!

src/deploy/functions/cache/applyHash.ts
TheIronDev: 👍

src/deploy/functions/cache/applyHash.ts
TheIronDev: Updated!

src/deploy/functions/cache/applyHash.ts
TheIronDev: Good point. I cleaned up the condition and added a bunch of comments.

src/deploy/functions/cache/applyHash.ts
TheIronDev: 🙇  !!! You're absolutely right! Thank you for catching this!

src/deploy/functions/cache/applyHash.ts
```diff
@@ -28,13 +36,18 @@ export function applyBackendHashToBackends(
 function applyBackendHashToEndpoints(
   wantBackend: Backend,
   envHash: string,
+  codebaseFilters: EndpointFilter[],
```
taeold, line 39: nit* I wonder if more appropriate var name for this is "deploy filter" or something - not sure what you mean by codebaseFilters here.

taeold (approved):

src/deploy/functions/cache/applyHash.ts
TheIronDev: Fair point. I wanted to convey that it was filtered by codebase already... but thats an implementation detail that the function doesn't need to call out. 👍 Renamed back to `filters`.

PR #4950, opened 2022-09-03
Support for non-== comparisons inside CEL

Berlioz:
fastest cmd-c cmd-v in the west


taeold (approved):
These tests cases are sooo good.

PR #4965, opened 2022-09-09
Adds support for codebases in `firebase init functions`

blidd-google:
The `firebase init functions` flow is unaware of the CF3 codebases feature, which allows users to organize their functions in multiple directories and deploy them separately. To initialize a codebase, users must create a new directory to house the codebase, create all the necessary files in the directory, and edit their project's `firebase.json` file to specify the codebase configuration details.

We update the `firebase init functions` flow to walk users through the initialization of new codebases, asking for user input on configuration details when necessary and automating the process of setting up the functions codebases structure.


src/init/features/functions/index.ts
```diff
@@ -30,6 +22,104 @@ export async function doSetup(setup: any, config: any, options: Options) {
       ensure(projectId, "runtimeconfig.googleapis.com", "unused", true),
     ]);
   }
+  setup.functions = {};
+
+  if (!config.src.functions) {
+    // if functions have not been initialized yet
+    setup.config.functions = [];
+    return codebaseSetup(setup, config, false);
+  } else {
```
taeold, line 31: nit* This else block seems unnecessary since the previous `if` block will always exit.

src/init/features/functions/index.ts
```diff
@@ -30,6 +22,104 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+ *  User dialogue to set up configuration for functions codebase.
+ */
+async function codebaseSetup(setup: any, config: Config, reinit: boolean): Promise<any> {
+  if (reinit) {
+    const choices = setup.config.functions.map((cbconfig: any) => {
+      return {
+        name: cbconfig["codebase"],
+        value: cbconfig["codebase"],
+      };
+    });
```
taeold, line 70: It looks like we have an opportunity to make the flow smarter - if there is only 1 codebase, we shouldn't need the user to select from our choices.

src/init/features/functions/index.ts
```diff
@@ -30,6 +22,104 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+
+/**
+ *  User dialogue to set up configuration for functions codebase.
+ */
+async function codebaseSetup(setup: any, config: Config, reinit: boolean): Promise<any> {
+  if (reinit) {
+    const choices = setup.config.functions.map((cbconfig: any) => {
+      return {
+        name: cbconfig["codebase"],
+        value: cbconfig["codebase"],
```
taeold, line 68: this is equivalent to:

`.map((cfg) => ({ name: config.codebase, value: config.codebase })`

Also - it looks like name and value points to the same value. Is that intentional?

src/init/features/functions/index.ts
```diff
@@ -30,6 +22,104 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+      };
+    });
+    const codebase = await promptOnce({
+      type: "list",
+      message: "Which codebase would you like to re-initialize?",
+      choices,
+    });
+    const cbconfig = configForCodebase(setup.config.functions, codebase);
+    _.set(setup, "functions.source", cbconfig.source);
```
taeold, line 77: I don't have strong opinions, but several folks prefer we stop using lodash since it often doesn't add too much value.

This is equivalent to:

```
setup.functions.source = cbconfig.source;
```

right?

Also - I'm a bit confused since I thought `setup.config.functions` was an array of configs?

src/init/features/functions/index.ts
```diff
@@ -30,6 +22,104 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+    _.set(setup, "functions.codebase", cbconfig.codebase);
+
+    logger.info("Re-initializing " + clc.bold(`codebase ${codebase}...`));
+    logger.info();
```
taeold, line 81: nit* I think I'd rather include `\n` in the previous log rather than calling `logger.info()`.

Sometimes, logger we are using injects prefix/labels (e.g. `logger.info("foo")` prints `[TIMESTAMP] foo` and it's sometimes wrong to assume that logger.info() will print out an empty newline.

src/init/features/functions/index.ts
```diff
@@ -30,6 +22,104 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+  } else {
+    logger.info();
+    logger.info("Let's create a new codebase to contain and manage your functions.");
```
taeold, line 84:
```suggestion
    logger.info("Let's create a new codebase for your functions.");
```

src/init/features/functions/index.ts
```diff
@@ -30,6 +22,104 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+    logger.info("A directory corresponding to the codebase will be created in your project");
+    logger.info("with sample code pre-configured.");
```
taeold, line 86: similar concern here - let's use a single `logger.info()` for each line rather than rely on that it won't print anything.

src/init/features/functions/index.ts
```diff
@@ -30,6 +22,104 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+
+    const source = await promptOnce({
+      type: "input",
+      message: "Where would you like to initialize your new functions source?",
+      default: "functions",
+    });
+    const codebase = await promptOnce({
+      type: "input",
+      message: "What should be the name of this codebase?",
+      default: source, // TODO: codebase-name-ify the dir name
```
taeold, line 102: are we not going to implement this on the first round of PR?

src/init/features/functions/index.ts
```diff
@@ -30,6 +22,104 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+    });
+    setup.config.functions.push({
+      source: source,
+      codebase: codebase,
+      ignore: ["node_modules", ".git", "firebase-debug.log", "firebase-debug.*.log"],
```
taeold, line 107: Small scope creep - this setup is specific to JS/TS. Can we find a place to do this in lang-specific setup?

src/init/features/functions/index.ts
```diff
@@ -30,6 +22,104 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+    });
+    // checks if the updated functions codebase config is valid
+    // also checks if the user-specified codebase name and source are valid
```
taeold, line 110: What happens if codebase name is invalid? Do we throw an error? 

I like the idea of giving an opportunity for user to retry instead of abruptly ending the init flow. wdyt?

src/init/features/functions/npm-dependencies.ts
```diff
@@ -14,8 +14,9 @@ export function askInstallDependencies(setup: any, config: any): Promise<void> {
   ]).then(() => {
     if (setup.npm) {
       return new Promise<void>((resolve) => {
+        console.log("CALL SPAWN");
```
taeold, line 17: nit* stray console log?

src/test/init/features/functions.spec.ts
```diff
@@ -0,0 +1,236 @@
+import * as sinon from "sinon";
+import { Config } from "../../../config";
+import * as prompt from "../../../prompt";
+import { Setup } from "../../../init";
+import { doSetup } from "../../../init/features/functions";
+import { Options } from "../../../options";
+import { RC } from "../../../rc";
+import { expect } from "chai";
+import * as _ from "lodash";
```
taeold, line 9: *nit: Unenforced but nice to follow rules when writing imports:

1) Separate out imports based on "builtin" vs "external" vs "local"

```
import foo from "fs"

import bar from "some-lib"

import car from "../local"
```

2. I tend to group named import and wildcard importt

```
import * from froo
import * from bar
import { foobar } from foobar
import { a, b, c } from alphabets
```

src/test/init/features/functions.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+      { projectDir: "test", cwd: "test" }
+    ),
+  };
+}
+
+describe("functions", () => {
+  const sandbox: sinon.SinonSandbox = sinon.createSandbox();
+  let promptOnceStub: sinon.SinonStub;
+  let promptStub: sinon.SinonStub;
+  // let spawnStub: sinon.SinonStub;
```
taeold, line 40: ?

src/test/init/features/functions.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+  let askWriteProjectFileStub: sinon.SinonStub;
+  let emptyConfig: Config;
+  let options: Options;
+
+  beforeEach(() => {
+    promptOnceStub = sandbox.stub(prompt, "promptOnce").throws("Unexpected promptOnce call");
+    promptStub = sandbox.stub(prompt, "prompt").throws("Unexpected prompt call");
+    // spawnStub = sandbox.stub(spawn, "spawn");
```
taeold, line 48: ?

src/test/init/features/functions.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+      nonInteractive: false,
+      interactive: false,
+      debug: false,
+      config: emptyConfig,
+      rc: new RC(),
+    };
+  });
+
+  afterEach(() => {
+    sandbox.restore();
```
taeold, line 68:
```suggestion
    sandbox.verifyAndRestore();
```

To make sure mocks are verified/satisfied.

src/test/init/features/functions.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  describe("doSetup", () => {
+    describe("with an uninitialized Firebase project repository", () => {
+      it("creates a new javascript codebase with the correct configuration", async () => {
+        const setup = { config: {}, rcfile: {} };
+        promptOnceStub.onFirstCall().resolves("testsource");
+        promptOnceStub.onSecondCall().resolves("testcodebase");
+        promptOnceStub.onThirdCall().resolves("javascript");
+
+        // eslint?
```
taeold, line 79: Can we elaborate? `?` doesn't sound very convincing.

src/test/init/features/functions.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+        promptStub.onFirstCall().callsFake((functions: any): Promise<void> => {
+          functions.lint = true;
+          return Promise.resolve();
+        });
+        // install dependencies?
```
taeold, line 84: Can we elaborate? `?` doesn't sound very convincing.

src/test/init/features/functions.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+        promptStub.onSecondCall().callsFake((setup: any): Promise<void> => {
+          setup.npm = false;
+          return Promise.resolve();
+        });
+        askWriteProjectFileStub = sandbox.stub(emptyConfig, "askWriteProjectFile");
+        askWriteProjectFileStub.resolves(true);
```
taeold, line 90: is it possible to do this setup in `beforeEach` like rest of the mock setups?

src/test/init/features/functions.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+        const writeFilesExpected = [
+          `${setupExpected.source}/package.json`,
+          `${setupExpected.source}/.eslintrc.js`,
+          `${setupExpected.source}/index.js`,
+          `${setupExpected.source}/.gitignore`,
+        ];
+
+        await doSetup(setup, emptyConfig, options);
+
+        expect(_.get(setup, "config.functions.0")).to.deep.equal(setupExpected);
```
taeold, line 107: nit* let's get rid of use of lodash.

taeold: nit* I think I'd prefer seeing a plain object:

```
.to.deep.equal({
          source: "testsource",
          codebase: "testcodebase",
          ignore: ["node_modules", ".git", "firebase-debug.log", "firebase-debug.*.log"],
          predeploy: ['npm --prefix "$RESOURCE_DIR" run lint'],
        })
```

when writing test cases.

src/test/init/features/functions.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  describe("doSetup", () => {
+    describe("with an uninitialized Firebase project repository", () => {
+      it("creates a new javascript codebase with the correct configuration", async () => {
+        const setup = { config: {}, rcfile: {} };
+        promptOnceStub.onFirstCall().resolves("testsource");
```
taeold, line 75: Here and onward: I think it'd be slightly better if we refactored `testsource` as a test variables `e.g. `const testSource = "testsource"` instead of peppering this magic string throughout this test case.

src/test/init/features/functions.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+        );
+      });
+
+      it("creates a new typescript codebase with the correct configuration", async () => {
+        const setup = { config: {}, rcfile: {} };
+        promptOnceStub.onFirstCall().resolves("testsource");
+        promptOnceStub.onSecondCall().resolves("testcodebase");
+        promptOnceStub.onThirdCall().resolves("typescript");
+        // promptStub.withArgs(setup.functions, FUNCTIONS_ESLINT_QUESTION).resolves(true);
+        // promptStub.onFirstCall().resolves(true); // use eslint?
```
taeold, line 119: ?

src/test/init/features/functions.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+        promptStub.onFirstCall().callsFake((functions: any): Promise<void> => {
+          functions.lint = true;
+          return Promise.resolve();
+        });
+        promptStub.onSecondCall().resolves(); // install dependencies?
```
taeold, line 124: Can we elaborate? ? doesn't sound very convincing.

src/test/init/features/functions.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+        ];
+
+        await doSetup(setup, config, options);
+
+        expect(_.get(setup, "config.functions")).to.deep.equal(setupExpected);
+        expect(askWriteProjectFileStub.getCalls().map((call) => call.args[0])).to.deep.equal(
+          writeFilesExpected
+        );
+      });
+      // TODO: legacy config
```
taeold, line 233: going to do the todo?

src/init/features/functions/index.ts
blidd-google: I wondered about that and decided against handling the single-codebase case in the name of "code-conciseness" (but really I was just being lazy). But agreed

src/init/features/functions/index.ts
blidd-google: yep, the displayed "name" should be the codebase name, and the "value" I want is also the codebase name so that I can retrieve the full config object by calling `configForCodebase(setup.config.functions, codebase)`.

src/init/features/functions/index.ts
blidd-google: I agree about the lodash thing... I would be perfectly happy to stop using it entirely.

So from my understanding of the previous developer's code, `setup.functions` keeps track of data needed for the rest of the functions initialization process. Things like `source`, `codebase`, `lint`, tracks configuration data in the `setup.functions` object that the next functions in the control flow can access to determine how to configure the new codebase.

 `setup.config.functions` refers to the new functions config object we are going to write to `firebase.json` as soon as we are done. So the two are distinct, but maybe `setup.functions` is not aptly named. Maybe something like `setup.options.functions` might be a little more explicit?

src/init/features/functions/index.ts
blidd-google: Totally fair. Not a fan of so many empty `logger.info` statements floating around anyway

src/init/features/functions/index.ts
blidd-google: So the utility I wrote does three things to the name of the directory provided by the user in the previous prompt:
1) truncates the string to the first 63 characters,
2) converts all uppercase chars to lowercase,
3) replaces all invalid chars with "_".

This seems to me reasonable and simple enough?

src/test/init/features/functions.spec.ts
blidd-google: Ah I think I decided to not do it because the "legacy config" test I had in mind wouldn't actually test anything that hadn't been covered by the previous tests

src/test/init/features/functions.spec.ts
blidd-google: So I remember why I did it this way--depending on the unit test, we will have different config objects, so we need to stub out the `askWriteProjectFile` function per test

src/functions/projectConfig.ts
```diff
@@ -34,7 +34,7 @@ function validateSingle(config: FunctionConfig): ValidatedSingle {
   }
   if (config.codebase.length > 63 || !/^[a-z0-9_-]+$/.test(config.codebase)) {
     throw new FirebaseError(
-      "Invalid codebase name. Codebase must be less than 63 characters and " +
+      "Invalid codebase name. Codebase must be less than or equal to 63 characters and " +
```
taeold, line 37: "less than 64"?

src/functions/projectConfig.ts
```diff
@@ -84,3 +84,13 @@ export function configForCodebase(config: ValidatedConfig, codebase: string): Va
   }
   return codebaseCfg;
 }
+
+/**
+ * Given an initial naming attempt, suggest a similar but valid codebase name
+ */
+export function suggestCodebaseName(name: string): string {
```
taeold, line 91: Can we add a test case for this function?

src/init/features/functions/index.ts
```diff
@@ -30,6 +26,106 @@ export async function doSetup(setup: any, config: any, options: Options) {
       ensure(projectId, "runtimeconfig.googleapis.com", "unused", true),
     ]);
   }
+  setup.functions = {};
+  if (!config.src.functions) {
+    // if functions have not been initialized yet
```
taeold, line 31: I will consider removing this comment if you think it falls under [go/tott/563](https://goto.google.com/tott/563).

src/init/features/functions/index.ts
```diff
@@ -30,6 +26,106 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+    setup.config.functions = [];
+    return codebaseSetup(setup, config, false);
+  }
+  // if functions have already been initialized, make sure config is validated and normalized
+  setup.config.functions = normalizeAndValidate(setup.config.functions);
+
+  logger.info();
```
taeold, line 38: nit* using newline chars instead of empty logger.info  calls?

src/init/features/functions/index.ts
```diff
@@ -30,6 +26,106 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+    default: "new",
+    choices,
+  });
+  return codebaseSetup(setup, config, initOpt === "reinit");
+}
+
+/**
+ *  User dialogue to set up configuration for functions codebase.
+ */
+async function codebaseSetup(setup: any, config: Config, reinit: boolean): Promise<any> {
```
taeold, line 64: nit* Just to offer different perspective - I don't see why we wouldn't define 2 different functions (`reinitCodebase` and `setupNewCodebase`) now that there is very little shared code between when `reinit=true` and `reinit=false`.

src/init/features/functions/index.ts
```diff
@@ -30,6 +26,106 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+    logger.info("\nRe-initializing " + clc.bold(`codebase ${codebase}...\n`));
+  } else {
+    logger.info("Let's create a new codebase for your functions.");
+    logger.info("A directory corresponding to the codebase will be created in your project");
+    logger.info("with sample code pre-configured.\n");
+
+    logger.info("See https://firebase.google.com/docs/functions/organize-functions for");
+    logger.info("more information on organizing your functions using codebases.\n");
+
+    logger.info("Functions can be deployed with " + clc.bold("firebase deploy") + ".\n");
```
taeold, line 94:
```suggestion
    logger.info(`Functions can be deployed with ${clc.bold("firebase deploy")}.\n`);
```

Yay string interpolation.

src/init/features/functions/index.ts
```diff
@@ -30,6 +26,106 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+
+    const source = await promptOnce({
+      type: "input",
+      message: "Where would you like to initialize your new functions source?",
+      default: "functions",
+    });
+    const codebase = await promptOnce({
```
taeold, line 101: It looks like we never examine the user-input for codebase. We should make sure user input is a valid codebase name, and if now prompt user to try again.

You might consider something like this:

```
    let attempts = 0
    while (true) {
      if (attempts >= MAX_ATTEMPTS) {
        throw new FirebaseError("Exceeded max attempt to input valid codebase name.");
      }
      prefix = await promptForCodebase(errMsg);
      attempts += 1;
    }
```

src/init/features/functions/index.ts
```diff
@@ -30,6 +26,106 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+      type: "input",
+      message: "What should be the name of this codebase?",
+      default: suggestCodebaseName(source),
+    });
+    let functionsConfig: FunctionsConfig = setup.config.functions.slice().concat({
```
taeold, line 106: q: why do we need to call `slice()`?

src/init/features/functions/index.ts
```diff
@@ -30,6 +26,106 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+      source: source,
+      codebase: codebase,
```
taeold, line 108:
```suggestion
      source,
      codebase,
```

^ is equivalent and shorter!

src/init/features/functions/index.ts
```diff
@@ -30,6 +26,106 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+    // checks if the updated functions codebase config is valid
+    // also checks if the user-specified codebase name and source are valid
+    try {
+      functionsConfig = normalizeAndValidate(functionsConfig);
+      setup.config.functions = functionsConfig;
+      setup.functions.source = source;
+      setup.functions.codebase = codebase;
+    } catch (err: any) {
+      logger.info(err);
+      return codebaseSetup(setup, config, false);
```
taeold, line 119: So if something goes wrong, we are asking the developer to try the flow again?

1) In what scenarios could things go wrong?

2) Can you experiment and see what kind of output devx this provides? I'm not sure if the previous `logger.info` call is informative enough for developers to understand what they did wrong.

src/init/features/functions/index.ts
```diff
@@ -52,11 +148,15 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
-  ]);
+  const cbconfig = configForCodebase(setup.config.functions, setup.functions.codebase);
+  switch (language) {
+    case "javascript":
+      cbconfig.ignore = ["node_modules", ".git", "firebase-debug.log", "firebase-debug.*.log"];
+      break;
+    case "typescript":
+      cbconfig.ignore = ["node_modules", ".git", "firebase-debug.log", "firebase-debug.*.log"];
+      break;
+    // add other cases as more languages are supported
```
taeold, line 159:
```suggestion
```

(I think this comment is unnecessary! The code makes it pretty obvious what needs to be done to support more languages.)

src/init/features/functions/index.ts
taeold: ah `setup.config.functions` vs `setup.config`. My eyes failed me 😛

src/init/features/functions/index.ts
blidd-google: fair point!

src/init/features/functions/index.ts
blidd-google: So my thought process was that `normalizeAndValidate(config)` will check the user-input for me when validating the config object (with the user-input codebase name inserted inside it). If `normalizeAndValidate` fails, then the error is logged and `initNewCodebase` is called recursively and the developer must try to input another codebase name. This may be too subtle of a way to check user input; I can see how explicitly checking the user-input codebase name would be more clear.

src/init/features/functions/index.ts
blidd-google: not sure actually. I don't know if it's completely obvious that checking `(!config.src.functions)` means I'm checking if cloud functions has been initialized yet. It also feels unnecessary to refactor this out into a function  `isFunctionsInitialized` (or something like that) which will end up being 1 line of code. Ideally, there would be some state in `Config` that tracks whether `functions` has been initialized which I could check here. Maybe I will leave it for now, and revisit this question of "initialized" state in the future?

blidd-google: Those are some very helpful tips about writing comments, btw!

CHANGELOG.md
```diff
@@ -0,0 +1,2 @@
+- update `firebase init functions` flow to be codebase-aware.
```
taeold, line 1: Please capitalize first sentence! Also, I think the rule is one changelog entry per PR. Since something like adding unit test isn't that relevant to the end-user of the CLI, I'd remove the entry.

src/functions/projectConfig.ts
```diff
@@ -25,25 +25,42 @@ export function normalize(config?: FunctionsConfig): NormalizedConfig {
   return [config];
 }
 
+/**
+ *
+ */
```
taeold, line 30: Please document or remove.

src/functions/projectConfig.ts
```diff
@@ -25,25 +25,42 @@ export function normalize(config?: FunctionsConfig): NormalizedConfig {
...
-  }
+  validateCodebase(config.codebase);
 
   return { ...config, source: config.source, codebase: config.codebase };
 }
 
-function assertUnique(config: ValidatedConfig, property: keyof ValidatedSingle) {
+/**
+ *
+ */
```
taeold, line 54: Please document or remove.

src/init/features/functions/index.ts
```diff
@@ -30,6 +30,130 @@ export async function doSetup(setup: any, config: any, options: Options) {
...
+  let codebase: string;
+
+  let attempts = 0;
+  while (true) {
+    if (attempts >= MAX_ATTEMPTS) {
+      throw new FirebaseError(
+        "Exceeded max number of attempts to input valid source. Please restart."
+      );
+    }
+    attempts++;
```
taeold, line 84: nit* This can be reduced to:

```
if (attempts++ >= MAX_ATTEMPTS) {
...
}
```

src/test/functions/projectConfig.spec.ts
```diff
@@ -120,4 +120,26 @@ describe("projectConfig", () => {
...
+    it("truncates names that are too long", () => {
+      expect(
+        projectConfig.suggestCodebaseName(
+          "1234567812345678123456781234567812345678123456781234567812345678"
+        )
+      ).to.equal("123456781234567812345678123456781234567812345678123456781234567");
+    });
+
+    it("suggests valid name based on name with multiple issues", () => {
+      expect(projectConfig.suggestCodebaseName("Codebase#1")).to.equal("codebase_1");
```
taeold, line 142: nit* I might also assert that suggested codebase is also a valid codebase by running it through `validateCodebase`

taeold (approved):
Minor nits overall lgtm.

Please update your PR title to something more descriptive.

src/test/functions/projectConfig.spec.ts
blidd-google: so after feedback from the API council I don't actually use the`suggestCodebaseName` utility lol. Should I keep the function in `projectConfig.ts` and this corresponding unit test in case it is useful in the future, or remove it for now?

Issue #3919, opened 2021-11-22
Can't deploy all functions consistently. Fails most of the time.


MarwanZeitoun:
Hello; 

I have a project with 169 cloud functions and need to set up automatic deployment to firebase in GitLab, but the deployment is always failed. 

I tried with various node versions, also with various access tokens, but always got the same error.
Locally one of the developers is able to deploy without any issue. But locally in my environment, it fails. Even when replicating the versions of node and firebase.

here is my .yml file
`
image: node:14.16.1
before_script:
  - cd functions
  - npm I
  
 script:
    - npx firebase use way-backend
    - npx firebase deploy --only functions --token $FIREBASE_TOKEN
  only:
    refs:
      - master
    changes:
      - functions/**/*
` 

first warnings like this  with many functions
`
⚠ functions: got "Quota Exceeded" error while trying to update projects/xproject/locations/us-central1/functions/xxxxx. Waiting to retry...
`
and then  got this one
`
Functions deploy had errors with the following functions:
	function1(us-central1)
	function2(us-central1)
	.... etc
Because there were errors creating or updating functions, the following functions were not deleted
To delete these, use firebase functions:delete
`
and not all functions throw the error only a few functions
and finally got this one

`
i  functions: cleaning up build files...
Cleaning up project directory and file based variables
00:01
ERROR: Job failed: exit code 1
`

Thanks in advance

abeisgoat:
Hey we're aware of issues when deploying over 100 functions, it's best to deploy the functions you changed specifically instead of the whole lot when you're at this scale.

MarwanZeitoun:
> Hey we're aware of issues when deploying over 100 functions, it's best to deploy the functions you changed specifically instead of the whole lot when you're at this scale.

usually, we deploy only changed functions, but we need to find a way to set up CI/CD on this project, also when we deploy partially the config changes are still undeployed!

ggirotto:
@abeisgoat I'm also facing a quota issue, although I'm trying to deploy just 20 functions. Can you confirm if it's related or it's another issue? The output that I have in Functions dashboard is:

```
[...]

PR #4967, opened 2022-09-09
Reduce functions deployment concurrency to 10

longuniquename:
### Description

There is an old issue we were facing with functions deployments reaching quotas: #3919. 

From Firebase [documentation](https://firebase.google.com/docs/functions/manage-functions#deploy_functions):

> When deploying large numbers of functions, you may exceed the standard quota and receive HTTP 429 or 500 error messages. To solve this, deploy functions in groups of 10 or fewer.

So why don't we just update this concurrency value from 40 to 10?

### Scenarios Tested

Deployed 90 functions with:

```sh
```

### DIFF
--- a/src/deploy/functions/release/index.ts
+++ b/src/deploy/functions/release/index.ts
@@ -62,7 +62,7 @@ export async function release(
   const functionExecutor: executor.QueueExecutor = new executor.QueueExecutor({
     retries: 30,
     backoff: 20000,
-    concurrency: 40,
+    concurrency: 10,
     maxBackoff: 40000,
   });
 


rhodgkins:
https://github.com/firebase/firebase-tools/issues/3919#issuecomment-1243786372:
>I'm planning to review #4967 soon, but my feeling is that fighting the GCF quota limit isn't fool-proof way, especially since quota is different depending on your project's state, location, history, etc.

@taeold even if this isn't a perfect fix, could this be the base for at least it being user configurable or something maybe by an environmental variable? Then the default can be kept to the current value (40), but allows lower values if needed?  I only suggest a environmental variable as I know command flags (which would probably make more sense for this) need to go through a whole review process. 
The currently function deployment is crippled with constant quota exceeded errors, so any fix (even if its temporary) has to be better than the current state.

taeold:
@longuniquename @rhodgkins @dikkeduif All great suggestions. We haven't done any deep investigations on what kind of value work best per function number, and the answer probably depends on the API quota allotted to the project.

I really like @rhodgkins's suggestion on exposing environment variable to configure this - that seems like the right first step. What if we do that and table this PR for later when we have some data that 10 is a better number (I'm not sure why our docs suggest 10 tbh)? If you have any data you can share with us, I'd deeply appreciate it.

rhodgkins:
Thanks for the update @taeold 🙂 

>If you have any data you can share with us, I'd deeply appreciate it.

If that's aimed at me I don't have anything I'm afraid. We've got a few environments/projects and there's no consistency on failed deployments due to quota errors. I could deploy locally without issue to one project, it could then fail deploying locally to a different project and then the Cloud Build might or might not succeed for either of the projects when I then try on there a while later 🤷‍♂️ all the projects have the same quota.

Also, as an aside, for all those projects the `Write requests per minute` quota is not editable anyway so we can't even increase the quota. 

quota screenshot

dikkeduif:
Regardless of the quotas I think firebase deploy should just auto retry to deploy. It's obvious we want to deploy our functions, no need to exit the process and tell us which ones failed to deploy. Is there a way to auto retry the ones that failed to deploy because of quota errors?

The way it works now is very inconsistent, sometimes it succeeds to deploy, sometimes it fails half of the functions and it's hard to understand exactly why.

Ideally it should be fire and forget.

PR #4968, opened 2022-09-09
[Function-NoOp] Fix Deploying after an Only Deploy

TheIronDev:
### Description

This commit fixes the behavior of turtle deploys such that forced no-op deploys (initiated by `--only functions:___`) still generate a hash.

This fixes the use case where a user forcefully deploys an individual function (using `--only`), then, without any changes, deploys the function without the `--only` flag.

Before this commit, functions deployed with `--only` would not have a hash. This would cause subsequent deploys to always deploy, even if they were unchanged.

After this commit, hash calculation is now stored as a label and is still compared against the local hash to detect for changes.

### Scenarios Tested

With the preview flag enabled...

* deploy a function
  * expected: function is deployed with hash stored as a label
  * actual: function is deployed with hash stored as a label
* deploy that function with the `--only` target filtering
  * expected: function is deployed with hash stored as a label
  * actual: function is deployed with hash stored as a label
* deploy a function (hash will get stored)
  * expected: function is skipped
  * actual: function is skipped

### Sample Commands

```
$ firebase deploy --only functions:helloWorld

...

$ firebase deploy 

(helloWorld is not deployed because its unchanged!)
```


src/deploy/functions/prepare.ts
```diff
@@ -341,6 +346,24 @@ function maybeCopyTriggerRegion(wantE: backend.Endpoint, haveE: backend.Endpoint
...
+        endpoint.targetedByOnly = true;
+      }
+      continue;
+    }
+    for (const endpoint of allEndpoints(wantBackend)) {
+      endpoint.targetedByOnly = isEndpointFiltered(endpoint, endpointFilters);
+    }
+  }
+}
+
```
taeold, line 366: I think this is equivalent to:

```js
for (const wantBackend of Object.values(wantBackends) {
  for (const endpoint of allEndpoints(wantBackend)) {
    if (endpointMatchesAnyFilter(endpoint, endpointFilters) {
      endpoint.targetByOnly = true
    }
  }
}
```

src/test/deploy/functions/prepare.spec.ts
```diff
@@ -159,4 +159,83 @@ describe("prepare", () => {
...
+      backends = { backend1, backend2 };
+    });
+
+    it("should mark targeted codebases", () => {
+      const filters = [{ codebase: "backend1" }];
+      // Execute
+      prepare.updateEndpointTargetedStatus(backends, filters);
+
+      // Expect
+      expect(endpoint1InBackend1.targetedByOnly).to.equal(true);
```
taeold, line 209: nit* prefer `expect(foo).to.be.true`

taeold (approved):
minor suggetions. Otherwise 🔥

src/test/deploy/functions/prepare.spec.ts
TheIronDev: Updated!

src/deploy/functions/prepare.ts
TheIronDev: Updated with the slight change of:

```
endpoint.targetByOnly = endpointMatchesAnyFilter(endpoint, endpointFilters);
```

PR #4970, opened 2022-09-09
Make FirebaseConfig properties available as internal parameters

Berlioz:
Based on bugbash feedback from @taeold and @colerogers.

We repurpose the `.secret` field of ParamValue (which wasn't being used anyway after we washed our hands of handling Secrets ourselves) to be `.internal`, which designates a param that we generated which shouldn't be written to disk or the runtime process.env.

We make the fields of FirebaseConfig--project id, RTDB URL, and storage bucket--available as internal params to the resolution process.

In order for this to do anything useful, the SDK will have to export corresponding special Expressions that users can use in their Functions; these Expressions will not add anything to the manifest (since they'll be generated by this code anyway) but will have to have their .value() function special cased to read from process.env.firebase_config instead.

### DIFF
--- a/src/deploy/functions/params.ts
+++ b/src/deploy/functions/params.ts
@@ -6,6 +6,7 @@ import { assertExhaustive, partition } from "../../functional";
 import * as secretManager from "../../gcp/secretManager";
 import { listBuckets } from "../../gcp/storage";
 import { isCelExpression, resolveExpression } from "./cel";
+import { FirebaseConfig } from "./args";
 
 // A convinience type containing options for Prompt's select
 interface ListItem {
@@ -199,7 +200,7 @@ export class ParamValue {
 
   constructor(
     private readonly rawValue: string,
-    readonly secret: boolean,
+    readonly internal: boolean,
     types: { string?: boolean; boolean?: boolean; number?: boolean }
   ) {
     this.legalString = types.string || false;
@@ -236,8 +237,7 @@ function resolveDefaultCEL(
 ): RawParamValue {
   const deps = dependenciesCEL(expr);
   const allDepsFound = deps.every((dep) => !!currentEnv[dep]);
-  const dependsOnSecret = deps.some((dep) => currentEnv[dep].secret);
-  if (!allDepsFound || dependsOnSecret) {
+  if (!allDepsFound) {
     throw new FirebaseError(
       "Build specified parameter with un-resolvable default value " +
         expr +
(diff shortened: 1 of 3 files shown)


src/deploy/functions/params.ts
```diff
@@ -326,12 +326,37 @@ export async function resolveParams(
...
+  });
+  defaultParams["GCLOUD_PROJECT"] = new ParamValue(config.projectId, true, {
+    string: true,
+    boolean: false,
+    number: false,
+  });
+  defaultParams["STORAGE_BUCKET"] = new ParamValue(config.storageBucket, true, {
+    string: true,
+    boolean: false,
+    number: false,
```
taeold, line 355: I'm pretty sure our type for FirebaseConfig is wrong. `databaseURL` and `storageBucket` may be empty for a project that didn't initialize neither of these products. Should we instead conditionally setup those parameters?

taeold (approved):
Super nice. 1 comment.

src/deploy/functions/params.ts
```diff
@@ -105,12 +105,21 @@ type ParamBase<T extends string | number | boolean> = {
   input?: ParamInput<T>;
 };
 
+/**
+ *
+ */
```
taeold, line 110: add documentation or remove?

src/deploy/functions/params.ts
```diff
@@ -334,11 +343,13 @@ export async function resolveParams(
...
-    number: false,
-  });
+  if (config.databaseURL !== "") {
+    defaultParams["DATABASE_URL"] = new ParamValue(config.databaseURL, true, {
+      string: true,
+      boolean: false,
+      number: false,
+    });
+  }
   defaultParams["PROJECT_ID"] = new ParamValue(config.projectId, true, {
```
taeold, line 353: nit* I wonder if we need PROJECT_ID. We reserve GCLOUD_PROJECT and not PROJECT_ID - what would happen if user had defined `PROJECT_ID` in the dotenv file 🤔

taeold (approved):

src/deploy/functions/params.ts
Berlioz: The user's definition wins. I added a test case to make this clear.

src/deploy/functions/params.ts

src/deploy/functions/params.ts
Berlioz: I think so, yeah.

colerogers (approved):
lgtm

PR #4983, opened 2022-09-15
Writing params to .env.local in emulator mode should not error if param exists in .env.projectId

Berlioz:
This is kind of an ugly special case, but the previous behavior got flagged as surprising during a bugbash and on reflection I think they're right. Also, bulking out the test cases, including fixing one test that just was not testing what it was purporting to.

### DIFF
--- a/src/functions/env.ts
+++ b/src/functions/env.ts
@@ -4,7 +4,7 @@ import * as path from "path";
 
 import { FirebaseError } from "../error";
 import { logger } from "../logger";
-import { logBullet } from "../utils";
+import { logBullet, logWarning } from "../utils";
 
 const FUNCTIONS_EMULATOR_DOTENV = ".env.local";
 
@@ -264,20 +264,19 @@ export function writeUserEnvs(toWrite: Record<string, string>, envOpts: UserEnvs
   if (Object.keys(toWrite).length === 0) {
     return;
   }
-
   const { functionsSource, projectId, projectAlias, isEmulator } = envOpts;
-  const envFiles = findEnvfiles(functionsSource, projectId, projectAlias, isEmulator);
-  const projectScopedFileName = `.env.${projectId}`;
 
-  const projectScopedFileExists = envFiles.includes(projectScopedFileName);
-  if (!projectScopedFileExists) {
-    createEnvFile(envOpts);
+  const allEnvFiles = findEnvfiles(functionsSource, projectId, projectAlias, isEmulator);
+  const targetEnvFile = envOpts.isEmulator ? FUNCTIONS_EMULATOR_DOTENV : `.env.${envOpts.projectId}`;
+  const targetEnvFileExists = allEnvFiles.includes(targetEnvFile);
+  if (!targetEnvFileExists) {
+    logger.debug(`Creating ${targetEnvFile}...`);
+    fs.writeFileSync(path.join(envOpts.functionsSource, targetEnvFile), "", { flag: "wx" });
   }
 
-  const currentEnvs = loadUserEnvs(envOpts);
   for (const k of Object.keys(toWrite)) {
     validateKey(k);
-    if (currentEnvs.hasOwnProperty(k)) {
+    if (isAlreadyDefinedWrite(k, envOpts)) {
       throw new FirebaseError(
         `Attempted to write param-defined key ${k} to .env files, but it was already defined.`
       );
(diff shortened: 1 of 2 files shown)


src/functions/env.ts
```diff
@@ -264,44 +264,67 @@ export function writeUserEnvs(toWrite: Record<string, string>, envOpts: UserEnvs
...
+  const targetEnvFile = envOpts.isEmulator
+    ? FUNCTIONS_EMULATOR_DOTENV
+    : `.env.${envOpts.projectId}`;
+  const targetEnvFileExists = allEnvFiles.includes(targetEnvFile);
+  if (!targetEnvFileExists) {
+    fs.writeFileSync(path.join(envOpts.functionsSource, targetEnvFile), "", { flag: "wx" });
+    if (!isEmulator) {
+      logBullet(
+        clc.yellow(clc.bold("functions: ")) +
+          `Created new local file ${targetEnvFile} to store param values. We suggest explicitly adding or excluding this file from version control.`
```
Berlioz, line 279: Genuinely have no idea what the wording of this message should be.

src/functions/env.ts
```diff
@@ -264,44 +264,67 @@ export function writeUserEnvs(toWrite: Record<string, string>, envOpts: UserEnvs
...
+/**
+ * Returns whether we should error on trying to write a key because it's already
+ * defined in the .env fields. This seems like a simple presence, check, but...
+ *
+ * For emulator deploys, it's legal to write a key to .env.local even if it's
+ * already defined in .env.projectId. This is a special case designed to follow
+ * the principle of least surprise for emulator users.
+ */
+export function isAlreadyDefinedWrite(key: string, envOpts: UserEnvsOpts): boolean {
+  const definedInEnv = loadUserEnvs(envOpts).hasOwnProperty(key);
```
inlined, line 313: This method is called in a loop. Are we really doing up to 4 file reads for every environment variable?

src/functions/env.ts
```diff
@@ -264,44 +264,67 @@ export function writeUserEnvs(toWrite: Record<string, string>, envOpts: UserEnvs
...
   logBullet(
-    clc.cyan(clc.bold("functions: ")) +
-      `Writing new parameter values to disk: ${projectScopedFileName}`
+    clc.cyan(clc.bold("functions: ")) + `Writing new parameter values to disk: ${targetEnvFile}`
   );
   for (const k of Object.keys(toWrite)) {
     fs.appendFileSync(
-      path.join(functionsSource, projectScopedFileName),
+      path.join(functionsSource, targetEnvFile),
       formatUserEnvForWrite(k, toWrite[k])
```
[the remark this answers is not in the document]
Berlioz: Good catch. Not anymore.

src/functions/env.ts
Berlioz: Fixed, though I have some serious concerns about the readability of my fix.

src/test/functions/env.spec.ts
```diff
@@ -324,18 +342,66 @@ FOO=foo
...
       expect(() =>
         env.writeUserEnvs(
-          { FOO: "bar" },
+          { FOO: "baz" },
           { projectId: "project", projectAlias: "alias", functionsSource: tmpdir }
         )
       ).to.throw(FirebaseError);
     });
 
+    it("is fine writing a key that already exists in any .env not not .env.local, in emulator mode", () => {
```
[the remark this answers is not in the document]
Berlioz: Fixed.

src/functions/env.ts
Berlioz: Done.

src/functions/env.ts
```diff
@@ -264,44 +264,72 @@ export function writeUserEnvs(toWrite: Record<string, string>, envOpts: UserEnvs
...
+    );
   }
 
-  const currentEnvs = loadUserEnvs(envOpts);
+  // Throw if any of the keys are duplicate (note special case if emulator) or malformed
+  const fullEnvs = loadUserEnvs(envOpts);
+  const prodEnvs = isEmulator
+    ? loadUserEnvs({ ...envOpts, isEmulator: false })
+    : loadUserEnvs(envOpts);
+  checkForDuplicateKeys(isEmulator || false, Object.keys(toWrite), fullEnvs, prodEnvs);
```
inlined, line 288: super nit: if you're only trying to null/undefined coalesce and not coalesce false `|| false` the new fancy syntax is `??`

src/functions/env.ts
```diff
@@ -264,44 +264,72 @@ export function writeUserEnvs(toWrite: Record<string, string>, envOpts: UserEnvs
...
-  return fileToWrite;
+/**
+ * Errors if any of the provided keys are aleady defined in the .env fields.
+ * This seems like a simple presence check, but...
+ *
+ * For emulator deploys, it's legal to write a key to .env.local even if it's
+ * already defined in .env.projectId. This is a special case designed to follow
+ * the principle of least surprise for emulator users.
+ */
+export function checkForDuplicateKeys(
```
inlined, line 312: Forgive my ignorance, but aren't we only prompting and writing params that weren't defined? Should the errors in this method be user-facing errors, or assertion failures that exit with code 2 (internal error)

inlined (approved):
Are we always re-prompting for all params in emulator mode? I wouldn't mind that. It could be a neat feature where we default to prod but ask again. Without this, though, I'm wondering whether our defense against writing a variable we would never have prompted for is dead code that could be replaced by assertions.

PR #4984, opened 2022-09-15
Display roles and APIs during ext:install and ext:update

joehan:
### Description
Display info about roles and APIs during ext:install and ext:update.

This was removed when we made the manifest changes - however, we want to ensure that users are informed of important changes to their project, so were reversing that decision.

### Scenarios Tested
Screen Shot 2022-09-15 at 2 42 26 PM


ifielker (approved):
LGTM

huangjeff5 (approved):

PR #4993, opened 2022-09-19
[Function-NoOp] Integration test to verify skipping deploy is WAI

TheIronDev:
### Description

Integration test to verify skipping deploy is WAI

### Scenarios Tested

- [x] NoOp Function Deploys are skipped
- [ ] Non-NoOp Function Deploys are skipped

### DIFF
--- a/scripts/functions-deploy-tests/tests.ts
+++ b/scripts/functions-deploy-tests/tests.ts
@@ -11,6 +11,7 @@ import * as tasks from "../../src/gcp/cloudtasks";
 import * as scheduler from "../../src/gcp/cloudscheduler";
 import { Endpoint } from "../../src/deploy/functions/backend";
 import { requireAuth } from "../../src/requireAuth";
+import { previews } from "../../src/previews";
 
 const FIREBASE_PROJECT = process.env.GCLOUD_PROJECT || "";
 const FIREBASE_DEBUG = process.env.FIREBASE_DEBUG || "";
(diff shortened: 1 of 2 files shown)


scripts/functions-deploy-tests/tests.ts
```diff
@@ -260,6 +263,83 @@ describe("firebase deploy", function (this) {
...
+        schedule: "every 30 minutes",
+        retryCount: 3,
+        minBackoffSeconds: 42,
+        maxRetrySeconds: 42,
+        maxDoublings: 42,
+        maxBackoffSeconds: 42,
+      },
+    };
+
+    previews.skipdeployingnoopfunctions = true;
```
colerogers, line 335: q: do we plan on removing this preview flag? If so, we should put a TODO here to yank it

colerogers (commented):
let me know when this is out of draft state and I'll do another pass

scripts/functions-deploy-tests/tests.ts
```diff
@@ -260,6 +263,83 @@ describe("firebase deploy", function (this) {
     }
   });
 
+  it("skips duplicate deploys functions with runtime options", async () => {
+    const opts: Opts = {
```
taeold, line 267: It looks like this configuration must match the configuration set in the previous test case in order for the test case to work as intended.

Can we make that coupling explicit? e.g. pull out configuration as a test variable.

scripts/functions-deploy-tests/tests.ts
TheIronDev: Updated!

scripts/functions-deploy-tests/tests.ts
TheIronDev: They don't need to match the previous one, I was just copy/pasting. I updated the opts to not be the giant list (to reduce distractions from the purpose of the test.)

TheIronDev:
@taeold Any idea why test returns this:

```
- YAMLException incomplete explicit mapping pair; a key node is missed; or followed by a non-tabulated empty line at line 1, column 62:
     ... rom function source: SyntaxError: Unexpected token '?'
```

Might be an issue with the project I am using....

taeold (approved):

colerogers (approved):

PR #4994, opened 2022-09-19
Implement Experiments feature

inlined:
This is a WIP. I still need to add the experiments command (which, ironically, until API council approves will be hidden behind a preview/experiment).

TODO: Add experiments:enable/disable/list/describe/clear commands (follow-up PR?)
TODO: Get representatives of each experiment to provide a shortDescription, verify if any should be public, and verify any assertEnabled calls.


src/experiments.ts
```diff
@@ -0,0 +1,182 @@
...
+export const EXPERIMENTS = Object.freeze([
+  // meta:
+  experiment("experiments", "enables the experiments family of commands"),
+
+  // Realtime Database experiments
+  experiment("rtdbrules", "TODO"),
+  experiment("rtdbmanagement", "TODO"),
+
+  // Extensions experiments
+  experiment("ext", "Enables the 'ext:sources:create' command"),
```
inlined, line 40: @pavelgj Can you reword or approve these commands?

1. Do you like the short description?
2. Do you want this to show up in `firebase experiments:list` or just in `firebase experiments:describe`?
2. Do you want to provide a full description and/or docs URL for customers?

src/experiments.ts
```diff
@@ -0,0 +1,182 @@
...
+  experiment("extdev", "Enalbes the 'ext:dev' family of commands"),
+
+  // Cloud Functions for Firebase experiments
+  experiment("pythonfunctions", "Python support for Cloud Functions for Firebase", {
+    fullDescription:
+      "Adds the ability to initializea and deploy Cloud " +
+      "Functions for Firebase in Python. While this feature is experimental " +
+      "breaking API changes are allowed in MINOR API revisions",
+  }),
+  experiment("golang", "Golang support for Cloud Functions for Firebase. Does not work.", {
```
inlined, line 50: @taeold I wonder if we should just delete this code path for now. WDYT? Maybe we can keep the folder but not load the runtime delegate in the main list?

src/experiments.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+  ),
+  experiment("functionsparams", "Adds support for paramaterizing functions deployments"),
+  experiment(
+    "skipdeployingnoopfunctions",
+    "Detect that there have been no changes to a function and skip deployment"
+  ),
+
+  // Emulator experiments
+  experiment("emulatoruisnapshot", "Load pre-release versions of hte emulator UI"),
```
inlined, line 84: @yuchenshi (Feel free to delegate to someone who now owns this). Do you want to have a different short description? A full description? Should this be an experiment that customers can see when they call `firebase experiments:list`?

src/experiments.ts
```diff
@@ -0,0 +1,182 @@
...
+    fullDescription:
+      "Adds support for popular web frameworks such as Next.js " +
+      "Nuxt, Netlify, Angular, and Vite-compatible frameworks. Firebase is " +
+      "committed to support these platforms long-term, but a manual migration " +
+      "may be required when the non-experimental support for these frameworks " +
+      "is released",
+  }),
+
+  // Access experiments
+  experiment("crossservicerules", "Allow Firebase Rules to reference resources in other services"),
```
inlined, line 97: @sam-gc Looks like you added this preview. Can you sign off or reword the short description? Should we make this experiment visible when customers call `firebase experiments:list`? If so, can you also provide a full description and/or docs?

Can you also verify/correct my usage of this experiment? Later in this PR I assert that the experiment is enabled, which breaks the test. In the "experiments" world, if we see a clear intent to use an experiment (e.g. if there are rules with cross service dependencies) we should assert that the experiment is enabled rather than fail silently. I turned this into an assert, but that causes a unit test to fail and I want to defer to you for the correct behavior.

src/experiments.ts
```diff
@@ -0,0 +1,177 @@
...
+    shortDescription,
+  } as Experiment<Name>;
+}
+
+export const EXPERIMENTS = Object.freeze([
+  // meta:
+  experiment("experiments", "enables the experiments family of commands"),
+
+  // Realtime Database experiments
+  experiment("rtdbrules", "TODO"),
```
inlined, line 36: @jsdt @lepatryk Can you provide a short description for these experiments? Should they be discoverable with the command `firebase experiments:enable`? If so, we should consider adding a full description and/or docs URL.

inlined:
A note to all people tagged for review on your experiment. If your experiment is going to have external users, you should add a log whenever executing experimental codepaths that mentions "experimental" in bold. E.g. "Deploying functions with **experimental** support for Python" or "Deploying hosting with **experimental** support for web frameworks"

src/experiments.ts
lepatryk: Hmm, what is this. @jsdt is this the rules canarying that Ryan was working on? I'm not sure we should make them discoverable right now.

src/handlePreviewToggles.ts
```diff
@@ -1,36 +1,39 @@
...
+  console.log(`${bold(red("Error:"))} Did not recognize preview feature ${bold(name || "")}`);
   process.exit(1);
 }
 
-export function handlePreviewToggles(args: string[]) {
-  const isValidPreview = has(previews, args[1]);
+/**
+ * Implement --open-sesame and --close-sesame
+ */
+export function handlePreviewToggles(args: string[]): boolean {
```
bkendall, line 15: Does this _need_ to return a boolean? Looks like it could just be `: void`

src/experiments.ts
inlined: Leaving them undiscoverable is the default. But we should probably still have a short description. I guess we can make it "for internal testing only"?

src/handlePreviewToggles.ts
inlined: The function returned falsey (undefined) when neither command was used, and this was used in if statements elsewhere. I formalized it by making it boolean and only returning false (or exiting)

src/experiments.ts
inlined: Resolved offline

src/experiments.ts
inlined: Chose to delete

src/experiments.ts
inlined: Will Not Fix (Obsolete). Flag was removed in https://github.com/firebase/firebase-tools/pull/5002

inlined:
ping?

src/commands/index.ts
```diff
@@ -69,6 +72,14 @@ export function load(client: any): any {
   client.experimental = {};
   client.experimental.functions = {};
   client.experimental.functions.shell = loadCommand("experimental-functions-shell");
+  if (experiments.isEnabled("experiments")) {
```
bkendall, line 75: I assume this is eventually going to be removed before submitting? How would I get to these commands otherwise?

src/deploy/index.ts
```diff
@@ -56,9 +56,10 @@ export const deploy = async function (
   const postdeploys: Chain = [];
   const startTime = Date.now();
 
-  if (previews.frameworkawareness && targetNames.includes("hosting")) {
+  if (targetNames.includes("hosting")) {
```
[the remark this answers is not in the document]
bkendall: That sounds right to me. None of the framework awareness stuff should work if we don't have a `source` path. 

What I'm a tiny bit afraid of is that this _will_ break someone who _may_ have set `frameworkawareness` but never set `source`. I'm not sure if I want to consider it a breaking change because it _is_ something someone opted into, but that's the consideration in my mind.

src/experiments.ts
```diff
@@ -0,0 +1,172 @@
...
+export interface Experiment {
+  shortDescription: string;
+  fullDescription?: string;
+  public?: boolean;
+  docsUri?: string;
+  default?: boolean;
+}
+
+// Utility method to ensure there are no typos in defining ALL_EXPERIMENTS
+function experiments<Keys extends string>(exp: Record<Keys, Experiment>): Record<Keys, Experiment> {
```
bkendall, line 16: I'm confused to what this is doing. Wouldn't just defining `ALL_EXPERIMENTS: Record<string, Experiments>` below enforce the same thing?

src/experiments.ts
```diff
@@ -0,0 +1,172 @@
...
+    return null;
+  }
+
+  // N.B. I personally would use < (name.length + malformed.length) * 0.2
+  // but this logic matches src/index.ts. I neither want to change something
+  // with such potential impact nor to create divergent behavior.
+  return Object.keys(ALL_EXPERIMENTS).filter(
+    (name) => leven(name, malformed) < malformed.length * 0.4
+  );
+}
```
bkendall, line 120:
```suggestion
}

```

src/commands/index.ts
bkendall: Ah, I see previews still works. :)

src/handlePreviewToggles.ts
```diff
@@ -1,36 +1,39 @@
...
-import { unset, has } from "lodash";
 import { bold, red } from "colorette";
 
-import { configstore } from "./configstore";
-import { previews } from "./previews";
+import * as experiments from "./experiments";
 
-function _errorOut(name?: string) {
-  console.log(bold(red("Error:")), "Did not recognize preview feature", bold(name || ""));
+function errorOut(name?: string): void {
```
bkendall, line 7: `name` is never used. Remove or call `errorOut` with a name?

bkendall (changes requested):

src/commands/experiments-clear.ts
```diff
@@ -0,0 +1,26 @@
...
+      return;
+    }
+
+    logger.error(`Cannot find experiment ${bold(experiment)}`);
+    const potentials = experiments.experimentNameAutocorrect(experiment)!;
+    if (potentials.length === 1) {
+      logger.error(`Did you mean ${potentials[0]}?`);
+    } else if (potentials.length) {
+      logger.error(
+        `Did you mean ${potentials.slice(0, -1).join(",")} or ${potentials[potentials.length - 1]}?`
```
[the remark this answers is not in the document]
inlined: Sure, but I also changed the signature of last to never return undefined since that only happened when someone passed a non-array which should never happen due to the signature of the method.

src/experiments.ts
inlined: If I called this a `Record<string, Experiments>` then the keys would be of type `string` and I can't use `keyof ALL_EXPERIMENTS` to make a typedef of experiment names. We need a literal, to preserve key names, but then if I don't type it, you neither get code complete for an experiment nor protection against mistyping the experiment. Using this utility makes the compiler enforce the type of values but infer the type of keys.

src/handlePreviewToggles.ts
inlined: How about instead I just start passing name in? That seems like it'll provide better value

src/commands/experiments-clear.ts
```diff
@@ -0,0 +1,29 @@
...
+import { last } from "../utils";
+
+export const command = new Command("experiments:clear <experiment>")
+  .description("clear preferences for an experiment on this machine")
+  .action((experiment: string) => {
+    if (experiments.isValidExperiment(experiment)) {
+      experiments.setEnabled(experiment, null);
+      experiments.flushToDisk();
+      logger.info(`Cleared preferences for experiment ${bold(experiment)}`);
+      return;
```
bkendall, line 16: Should this throw instead?

src/commands/experiments-describe.ts
```diff
@@ -0,0 +1,29 @@
...
+  .action((experiment: string) => {
+    if (!experiments.isValidExperiment(experiment)) {
+      logger.error(`Cannot find experiment ${bold(experiment)}`);
+      const potentials = experiments.experimentNameAutocorrect(experiment);
+      if (potentials.length === 1) {
+        logger.error(`Did you mean ${potentials[0]}?`);
+      } else if (potentials.length) {
+        logger.error(`Did you mean ${potentials.slice(0, -1).join(",")} or ${last(potentials)}?`);
+      }
+      return;
```
bkendall, line 19: Should this throw instead?

src/commands/experiments-list.ts
```diff
@@ -0,0 +1,25 @@
...
+  for (const [name, exp] of enabled) {
+    table.push(["y", name, exp.shortDescription]);
+  }
+  for (const [name, exp] of disabled) {
+    if (!exp.public) {
+      continue;
+    }
+    table.push(["n", name, exp.shortDescription]);
+  }
+  logger.info(table.toString());
```
bkendall, line 24: One detail that can be punted on until later: for `--json` (machine readable) output, we could export the table as a JSON-ish object (just return an object in this method). Would that be easy enough to do?

bkendall (approved):

src/commands/experiments-clear.ts
inlined: Actually, Michael asked us to remove this feature for now.

src/commands/experiments-describe.ts
inlined: Done

src/commands/experiments-list.ts
inlined: I couldn't get this to work. You'll have to teach me how --json works with the commands library

src/commands/experiments-clear.ts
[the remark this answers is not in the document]
inlined: I'm an idiot and last can return undefined if the array is non-empty. Will fix

PR #4996, opened 2022-09-20
Single project mode for the auth emulator.

christhompsongoogle:
Single project mode for the Auth emulator. If singleProjectMode is set, the emulator will reject calls for project IDs that aren't the default project ID.

### DIFF
--- a/src/emulator/auth/index.ts
+++ b/src/emulator/auth/index.ts
@@ -13,6 +13,8 @@ export interface AuthEmulatorArgs {
   projectId: string;
   port?: number;
   host?: string;
+  singleProjectModeWarning?: boolean;
+  singleProjectModeError?: boolean;
 }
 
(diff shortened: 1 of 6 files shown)


src/emulator/auth/index.ts
```diff
@@ -13,6 +13,8 @@ export interface AuthEmulatorArgs {
   projectId: string;
   port?: number;
   host?: string;
+  singleProjectModeWarning?: boolean;
```
yuchenshi, line 16: Can we have an `enum` instead of two booleans? Like `ENFORCED`, `WARN`, and `DISABLED`?

yuchenshi (approved):

src/emulator/auth/index.ts
christhompsongoogle: Added

PR #5001, opened 2022-09-21
Update GCF v2 API call to include filter condition for listing v2 functions

taeold:
GCF CP is going to return both 1st Gen and 2nd Gen functions (2-3 weeks) in the ListFunctions API and now require specific filter condition for the API to return only the V2 functions as before.

We update the listFunctions API to include a filter condition to fetch only the v2 functions as done prior to the change. 

Consequently making an update to use the latest, stable v2 api endpoint for GCF V2 calls.

### DIFF
--- a/src/gcp/cloudfunctionsv2.ts
+++ b/src/gcp/cloudfunctionsv2.ts
@@ -1,6 +1,6 @@
 import * as clc from "colorette";
 
-import { Client } from "../apiv2";
+import { Client, ClientVerbOptions } from "../apiv2";
 import { FirebaseError } from "../error";
 import { functionsV2Origin } from "../api";
 import { logger } from "../logger";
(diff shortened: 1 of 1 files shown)


src/gcp/cloudfunctionsv2.ts
```diff
@@ -19,7 +19,7 @@ import {
   HASH_LABEL,
 } from "../functions/constants";
 
-export const API_VERSION = "v2alpha";
+export const API_VERSION = "v2";
```
colerogers, line 22: Only `v2` is introducing this change, right?

Otherwise all old CLI installs will break

colerogers (approved):

src/gcp/cloudfunctionsv2.ts
taeold: Yes old CLIs would display double entries. I don't think we have other choice though do we?

colerogers: Oh yeah, I'm not sure there's another way out of this then

PR #5002, opened 2022-09-22
Update emulator version and move csr out of feature preview

sam-gc:


taeold (approved):
Do you want to remove the preview flag entirely by deleting it from preview.ts?

sam-gc:
Oh my bad, I thought I did. Updated

PR #5011, opened 2022-09-23
Adds test lab triggers to firebase deploy

colerogers:


src/deploy/functions/services/testLab.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @param endpoint the test lab endpoint
+ */
+export function ensureTestLabTriggerRegion(
+  endpoint: backend.Endpoint & backend.EventTriggered
+): Promise<void> {
+  if (!endpoint.eventTrigger.region) {
+    endpoint.eventTrigger.region = "global";
+  }
+  if (endpoint.eventTrigger.region !== "global") {
+    throw new FirebaseError("A test lab trigger must specify 'global' trigger location");
```
taeold, line 15: nit* Does Test Lab team have a product name preference? e.g. test lab vs testlab vs TestLab vs Test Lab?

taeold (approved):

TheIronDev (approved):
LGTM

src/deploy/functions/services/testLab.ts
colerogers: All sites/docs/team pages show Test Lab, but in the CLI we tend to keep everything lowercase. I'll change for this service and maybe we can revisit the other implemented services later

PR #5015, opened 2022-09-26
[Do not review] Add support for pinTag rewrite option

inlined:
This does not yet actually release the feature since the backend changes are still pending. Instead this creates a tag, garbage collects if there are more than 500 tags, and then logs the tag that would have been sent as a pin.

Along the way my OCD got the better of me. The codebase predates some of the type definitions that have really improved the functions codebase, so I added them in. This revealed that there was a lot of conversion between superfluous types, which included a lot of `any`s. 

A quick list of changes made:
1. Canged/added typing. Documented some of functions' fields in `Context`, added more documentation for the hosting API, used more specific typing in the Hosting codebase so the type system guaranteed we didn't need to handle certain edge cases, and removed the local type cast in hosting deploy so that we either have firebase.json types or api types. Moved hosting deploy code to use the API library for stronger typing of requets as well.
2. Refactored a utility for Cloud Run from the functions codebase into the API file so it could be reused.
3. I combined `normalizedConfigs` and `validate` into one `config` file that has much more granular methods that are easier to unit test.
4. Dramatically simplified the logic for looking up functions for hosting redirects.

Some additional changes I think we should do:
1. Get rid of the first pass of convertConfig; just create the version with labels only in `prepare` and remove the `finalize` param from `convertConfig` because it is called only in `release.
2. Change `prepare` slightly so that `HostingDeploy` can be guaranteed to have a non-null `version`.

This change has waaaaaay more LOC than there were meaningful changes due to things like indentation changing or consolidating files. Feel free to ask me to work together to figure out how to break this into more manageable chunks.


inlined:
Leaving this PR up so I can browse locally w/o switching branches, but this is being broken up into smaller reviews.

inlined:
Sub-PRs are committing everything

PR #5016, opened 2022-09-26
Define updateService to make run updates feel strongly consistent

inlined:
This is a fragment of the monster monolith that is #5015. In that PR I want the ability to modify run instances and I need the ability to wait for changes to propagate. This is already part of `fabricator`, so I just move that utility into `gcp/run.ts`. I call it `updateService` since it acts like a normal GCP update command, but just manages the replaceService and eventual consistency delays.

Feel free to scan through this one. It's basically a cut & paste PR.


TheIronDev (approved):
Lgtm!

taeold (approved):

PR #5017, opened 2022-09-26
Consolidate config utilities and add better typing

inlined:
This is a fragment of the monster monolith that is https://github.com/firebase/firebase-tools/pull/5015.

Part of the point of this PR is to consolidate normalizeConfig and validate into one utility that has more testable/composable units. The LOC for validate can be ignored; they're just copy+pasted (sans any type adjustment). The LOC for the rest of the config functions are the meat of the change but are mostly copy + paste.

Some other notable changes:
1. We use a proper Context and Options typed object everywhere
2. We got rid of an unnecessary re-definition of the firebase.json types that even omit type info
3. the function formerly called normalizeConfigs now _always_ performs resolveSite. It felt like a massive code smell that normalizeConfigs had a cache that ignores options.resolveTargets. In the one place where we don't want to resolve targets, we now just use the individual components of `hostingConfig`


bkendall (changes requested):

src/hosting/config.ts
```diff
@@ -0,0 +1,262 @@
...
+import { dirExistsSync } from "../fsutils";
+import { resolveProjectPath } from "../projectPath";
+import path from "path";
+
+// TODO: Consider putting this type along with the Options type.
+// I haven't tried to do this yet because true options embeds classes, not just
+// interfaces.
+// We should consider either refactoring options so it can be more easily mocked,
+// or creating a utility for constructing options easily in tests.
+export interface MockableOptions {
```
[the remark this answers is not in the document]
inlined: This is explicitly not a partial. It is explicitly requiring exactly and only the parts of Option that are necessary for this codebase to work.

src/deploy/hosting/release.ts
```diff
@@ -17,14 +19,14 @@ export async function release(context: any, options: any, payload: Payload): Pro
...
-    context.hosting.deploys.map(async (deploy: any) => {
+    context.hosting.deploys.map(async (deploy) => {
       utils.logLabeledBullet(`hosting[${deploy.site}]`, "finalizing version...");
 
       const config = await convertConfig(context, payload, deploy.config, true);
       const data = { status: "FINALIZED", config };
       const queryParams = { updateMask: "status,config" };
 
-      const finalizeResult = await client.patch(`/${deploy.version}`, data, { queryParams });
+      const finalizeResult = await client.patch(`/${deploy.version!}`, data, { queryParams });
```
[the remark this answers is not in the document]
inlined: Per offline conversation added a throw earlier if version is undefined. As one of the last changes to be made in this repo I'm going to make HostingDeploy require a site and a version through the type system and not a null assertion.

src/hosting/config.ts
```diff
@@ -0,0 +1,256 @@
...
+        `Couldn't find specified i18n root directory ${bold(
+          config.i18n.root
+        )} in public directory ${bold(config.public)}`
+      );
+    }
+  }
+}
+
+/**
+ * Converts all configs from having a target to having a soruce
```
Berlioz, line 214: Typo

src/hosting/config.ts
```diff
@@ -0,0 +1,256 @@
...
+
+  const exceptTargets = exceptOption.split(",");
+  if (exceptTargets.includes("hosting")) {
+    return [];
+  }
+
+  const exceptValues = exceptTargets
+    .filter((t) => t.startsWith("hosting:"))
+    .map((t) => t.replace("hosting:", ""));
+  const toReject = matchingConfigs(configs, exceptValues, /* assertMatch=*/ false);
```
Berlioz, line 101: Is that comment just a stray from development? I can't tell what it's trying to say.

scripts/npm-link.sh
```diff
@@ -3,3 +3,5 @@ set -e
 
 echo "Running npm link..."
 npm link
+
+chmod u+x ./lib/bin/firebase.js
```
Berlioz, line 7: extreme nitpicking, but wouldn't +rx be more appropriate here?

src/hosting/config.ts
```diff
@@ -0,0 +1,256 @@
...
+
+/**
+ * Returns a subset of configs that match the only string
+ */
+export function filterOnly(configs: HostingMultiple, onlyString?: string): HostingMultiple {
+  if (!onlyString) {
+    return configs;
+  }
+
+  let onlyTargets = onlyString.split(",");
```
Berlioz, line 70: Dumb question: do we control totally the fields of the Options struct? Are there any situations in which doing a split on "," without stripping whitespace could bite us?

src/hosting/config.ts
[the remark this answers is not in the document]
bkendall: Ah, maybe [Pick](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys) of Options could be useful, but don't think it's blocking.

src/options.ts
```diff
@@ -26,6 +27,14 @@ export interface Options {
...
   rc: RC;
 
+  // Hosting-specific (including web frameworks) options
+  normalizedHostingConfig?: HostingResolved[];
+  site?: string; // When using a single hosting site, this may be filled by firedata
+  expires?: `${number}${"h" | "d" | "m"}`;
+
+  // Emulator options:
+  port?: number;
+
```
bkendall, line 37: Comment got lost/outdated:

So I think we should avoid putting [product]-specific things into the Options object here. My thoughts before were to make sure the hosting deploy path could specify their requirements in their functions in a way that Options plus something (e.g. extending Options) could satisfy the requirement.

My fear here is that `port` shows up in Options and starts being used. If I see `options.expires`, I'm going to have no idea what that's for.

inlined: Per offline conversation (which admittedly I'm quoting here before you had much of an opportunity to contribute) I repurposed `MockableOptions` as `HostingOptions`. I still pretty strongly believe I should leave the fields in the top-level `Options` object so that I can be protected from drift with the `Implements` assertion and so that I don't need to cast `Options as unknown as HostingOptions`. If you really strongly believe I should leave fields undocumented and violate the type system to do so, I'll grumble but comply.

scripts/npm-link.sh
inlined: Was always created with read-access, but added it to be explicit.

src/hosting/config.ts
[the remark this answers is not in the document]
inlined: After playing in the Typescript playground I found I am wrong. Pick can work on classes and return an interface. If you think it is more readable I can write a new metaprogramming library called `RecursivePick<object, RecursiveKeyOf<object>>` and instead say:

```
type HostingOptions = RecursivePick<Options, "project" | "cwd" | "configPath" | "only" | "except" | "config.src", "rc.requireTarget"> & {
  normalizedHostingConfig?: Array<HostingResolved>;
  expires?: `${number}${"h" | "d" | "m"}`
}
```

But I personally think it reads better to redefine the fields I depend on and use metaprogramming to verify I haven't changed any field's type.

src/hosting/config.ts
```diff
@@ -0,0 +1,245 @@
...
+}
+
+/** Validates hosting configs for semantic correctness. */
+export function validate(configs: HostingMultiple, options: HostingOptions): void {
+  for (const config of configs) {
+    validateOne(config, options);
+  }
+}
+
+function validateOne(config: HostingMultiple[number], options: HostingOptions): void {
```
bkendall, line 130: What does `[number]` do in this context? I've never seen this before (and you can imagine google searching "typescript `[number]`" doesn't exactly give me the right results haha)

src/test/hosting/config.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+      );
+    });
+  });
+
+  describe("filterOnly", () => {
+    const tests: Array<
+      {
+        desc: string;
+        cfg: HostingMultiple;
+        only: string | undefined;
```
bkendall, line 103: nit (same below): 
```suggestion
        only?: string;
```

src/test/hosting/config.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+      });
+    }
+  });
+
+  const PUBLIC_DIR_ERROR_PREFIX = /Must supply a "public" directory/;
+  describe("validate", () => {
+    const tests: Array<{
+      desc: string;
+      site: HostingSingle;
+      wantErr: RegExp | false;
```
bkendall, line 249: idea: `wantErr?: RegExp` could eliminate all the places where you have `wantErr: false`. Follows a nice go-pattern of "if the default value works, don't bother specifying it". The check in the test would still be `!t.wantErr`, but you wouldn't have to specify `wantErr: false` in each happy test case (which should be pretty much assumed unless you have a desired error specified)

bkendall (approved):
Couple nits and questions, but LGTM

src/hosting/config.ts
inlined: Type arithmetic. What happens when you index a HostingMultiple with a number? Since HostingMultiple is an array type, you get the element type of HostingMultiple. You would _think_ that would be a HostingSingle, but HostingSingle allows both site and target to be undefined and the array syntax of HostingMultiple always requires site or target to be defined.

inlined: Oh, fun fact, you can also index by a literal string. For example `{ foo: string }["foo"]` is `string`. I think I do that elsewhere to say a `HostingOptions["duration"]` without separately exporting the duration field's type.

src/test/hosting/config.spec.ts
inlined: Done

src/test/hosting/config.spec.ts
inlined: Removed. I had done that to improve the readability of the test matrix. There is no desired outcome; there was no `want` for me to set when I expected things to go smoothly. So instead saying `wantErr: false` felt like I was making this test actually _assert_ something, as tests usually do.

src/firebaseConfig.ts
```diff
@@ -72,18 +69,25 @@ type HostingBase = {
...
 };
 
-type HostingSingle = HostingBase & {
+export type HostingSingle = HostingBase & {
   site?: string;
   target?: string;
 } & Deployable;
 
-type HostingMultiple = (HostingBase &
+export type HostingMultiple = (HostingBase &
```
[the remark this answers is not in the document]
inlined: Also, as TL, you have the power to wave your hands and ask me to remove this HostingMultiple vs HostingSingle nonsense since the new utilities return a HostingResolved[] instead. 🙃

PR #5022, opened 2022-09-27
Inlined.more api

inlined:
Add missing methods in the hosting API file. Add missing tests. Fix some types to use string literal types instead of enums (much easier to provide in tests). Use the API library in the deploy codebase.


src/test/hosting/api.spec.ts
```diff
@@ -266,6 +266,125 @@ describe("hosting", () => {
     });
   });
 
+  describe("createVersion", () => {
+    afterEach(nock.cleanAll);
+
+    it("should make the API requests to create a version", async () => {
+      const VERSION = { status: "CREATED" } as const;
```
bkendall, line 273: What does `as const` do?

src/test/hosting/api.spec.ts
```diff
@@ -266,6 +266,125 @@ describe("hosting", () => {
...
+      },
+      fileCount: 0,
+      versionBytes: 0,
+    };
+    const VERSION_2 = {
+      ...VERSION_1,
+      name: `projects/-/sites/${SITE}/versions/v2`,
+    };
+
+    it("Returns a single page of versions", async () => {
```
bkendall, line 350:
```suggestion
    it("returns a single page of versions", async () => {
```

src/deploy/hosting/release.ts
```diff
@@ -15,38 +14,40 @@ export async function release(context: Context, options: Options, payload: Paylo
...
-
-      const finalizeResult = await client.patch(`/${deploy.version!}`, data, { queryParams });
-
-      logger.debug(`[hosting] finalized version for ${deploy.site}:${finalizeResult.body}`);
+      const update: Partial<api.Version> = {
+        status: "FINALIZED",
+        config: await convertConfig(context, payload, deploy.config, /* finalize= */ true),
+      };
+
+      const parts = deploy.version!.split("/");
```
bkendall, line 27: Can we just check for `deploy.version`?

src/deploy/hosting/release.ts
```diff
@@ -15,38 +14,40 @@ export async function release(context: Context, options: Options, payload: Paylo
...
+      const versionId = parts[parts.length - 1];
+      const finalizedVersion = await api.updateVersion(deploy.site, versionId, update);
+
+      logger.debug(
+        `[hosting] finalized version for ${deploy.site}:${JSON.stringify(
+          finalizedVersion,
+          null,
+          2
+        )}`
```
bkendall, line 36: I'd prefer we don't bloat the debug lines with newlines if we can help it. It makes it impossible to get the entire log by just looking for a single line (i.e. `cat log | grep "some request"` doesn't work - the rest of the object is on new lines)

src/deploy/hosting/release.ts
```diff
@@ -15,38 +14,40 @@ export async function release(context: Context, options: Options, payload: Paylo
...
-        `/projects/${projectNumber}/sites/${deploy.site}${channelSegment}/releases`,
-        { message: options.message || null },
-        { queryParams: { versionName: deploy.version! } }
+      const release = await api.createRelease(
+        deploy.site,
+        context.hostingChannel || "live",
+        deploy.version!
       );
-      logger.debug("[hosting] release:", releaseResult.body);
+      logger.debug("[hosting] release:", JSON.stringify(release, null, 2));
```
bkendall, line 50: Same here. I'd prefer not adding in the newlines.

bkendall (changes requested):

src/deploy/hosting/release.ts
inlined: I assume you mean check instead of using the null assertion. That's fixed in the upstream PR and merged into here.

bkendall (approved):

PR #5032, opened 2022-09-28
[Function-NoOp] Enable Skipping NoOp Deploys by Default

TheIronDev:
### Description

This commit flips the "skip no-op deploys" flag to default to `true`.

When enabled, `firebase deploy` will compare the last deployed instances of a function with the current copy and, if they are the same, will skip uploading and deploying the function.


bkendall:
Add a CHANGELOG entry, please.

bkendall (approved):

taeold (approved):

PR #5034, opened 2022-09-28
dependency updates

bkendall:
### Description

1. `googleapis` is a `111MB` dependency... but it's only used in tests! so I moved it to devDeps so we can shrink the install size of `firebase-tools` even more
2. `vm2` had an audit warning, so I updated that too.


joehan (approved):
Love to see it, good catch!

PR #5040, opened 2022-09-29
Include roles and apis for Tasks and Secrets during DisplayExtensionInfo

joehan:
### Description
When displaying info about an extension being installed or updated, also show role + API for Cloud tasks and role for Secrets, if applicable.

I'm not listing the API for Secrets at this stage, because it is enabled by the client, not by the extensions backend, so we display a message earlier in the install.

### Scenarios Tested
Install an extension with tasks:
Screen Shot 2022-09-28 at 5 26 58 PM

Install an extension with Secrets:
Screen Shot 2022-09-29 at 10 39 22 AM


src/extensions/displayExtensionInfo.ts
```diff
@@ -39,13 +43,17 @@ export async function displayExtInfo(
     if (spec.license) {
       lines.push(`**License**: ${spec.license}`);
     }
-    lines.push(`**Source code**: ${spec.sourceUrl}`);
+    if (spec.sourceUrl) {
```
huangjeff5, line 46: Should this be optional?

joehan: Good point, it should be - done.

huangjeff5 (approved):

PR #5042, opened 2022-09-29
Set default values properly for selector params

Berlioz:


inlined (approved):

taeold (commented):
Can we merge this on launch.next instead?

PR #5045, opened 2022-09-30
Adds emulator support for v2 rtdb triggers

colerogers:
This change tries to make the minimum amount of changes to the functions emulator to support v2 rtdb triggers. Once we get some breathing room in the future, I'd like to refactor how triggers are added and possibly merge some of the deploy/emulator code so we aren't re-building similar logic.

### DIFF
--- a/src/emulator/functionsEmulator.ts
+++ b/src/emulator/functionsEmulator.ts
@@ -589,7 +589,9 @@ export class FunctionsEmulator implements EmulatorInstance {
             added = await this.addRealtimeDatabaseTrigger(
               this.args.projectId,
               key,
-              definition.eventTrigger
+              definition.eventTrigger,
+              signature,
+              definition.region
             );
             break;
           case Constants.SERVICE_PUBSUB:
@@ -721,21 +723,12 @@ export class FunctionsEmulator implements EmulatorInstance {
     }
   }
 
-  async addRealtimeDatabaseTrigger(
-    projectId: string,
-    key: string,
-    eventTrigger: EventTrigger
-  ): Promise<boolean> {
-    const databaseEmu = EmulatorRegistry.get(Emulators.DATABASE);
-    if (!databaseEmu) {
-      return false;
-    }
-
-    const result: string[] | null = DATABASE_PATH_PATTERN.exec(eventTrigger.resource);
+  private getV1DatabaseApiAttributes(projectId: string, key: string, eventTrigger: EventTrigger) {
+    const result: string[] | null = DATABASE_PATH_PATTERN.exec(eventTrigger.resource!);
     if (result === null || result.length !== 3) {
       this.logger.log(
         "WARN",
-        `Event function "${key}" has malformed "resource" member. ` + `${eventTrigger.resource}`
+        `Event function "${key}" has malformed "resource" member. ` + `${eventTrigger.resource!}`
       );
       throw new FirebaseError(`Event function ${key} has malformed resource member`);
     }
@@ -748,11 +741,9 @@ export class FunctionsEmulator implements EmulatorInstance {
       topic: `projects/${projectId}/topics/${key}`,
     });
 
-    logger.debug(`addRealtimeDatabaseTrigger[${instance}]`, JSON.stringify(bundle));
-
-    let setTriggersPath = "/.settings/functionTriggers.json";
+    let apiPath = "/.settings/functionTriggers.json";
     if (instance !== "") {
-      setTriggersPath += `?ns=${instance}`;
+      apiPath += `?ns=${instance}`;
     } else {
       this.logger.log(
         "WARN",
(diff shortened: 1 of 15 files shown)


src/emulator/functionsEmulator.ts
```diff
@@ -819,7 +862,7 @@ export class FunctionsEmulator implements EmulatorInstance {
     logger.debug(`addPubsubTrigger`, JSON.stringify({ eventTrigger }));
 
     // "resource":\"projects/{PROJECT_ID}/topics/{TOPIC_ID}";
-    const resource = eventTrigger.resource;
+    const resource = eventTrigger.resource!;
```
TheIronDev, line 865: is this ! necessary here?

TheIronDev (approved):
Lgtm!

src/emulator/functionsEmulator.ts
colerogers: We make def.resource `string | undefined` now

src/emulator/functionsEmulator.ts
```diff
@@ -748,24 +741,74 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+  ) {
+    const instance =
+      eventTrigger.eventFilters?.instance || eventTrigger.eventFilterPathPatterns?.instance;
+    if (!instance) {
+      throw new FirebaseError("A database instance must be supplied.");
+    }
+
+    const ref = eventTrigger.eventFilterPathPatterns?.ref;
+    if (!ref) {
+      throw new FirebaseError("A database reference must be supplied.");
```
taeold, line 771: q: What does it mean for these values to not exist (in other words - can these conditions be triggered if users are using the function sdk correctly?)

src/emulator/functionsEmulator.ts
```diff
@@ -748,24 +741,74 @@ export class FunctionsEmulator implements EmulatorInstance {
...
+
+    const bundle = JSON.stringify({
+      name: `projects/${projectId}/locations/${region}/triggers/${key}`,
+      path: ref,
+      event: eventTrigger.eventType,
+      topic: `projects/${projectId}/topics/${key}`,
+      namespacePattern: instance,
+    });
+
+    const apiPath = "/.settings/functionTriggers.json";
```
taeold, line 782: does difference in how we construct the apiPath for v1 and v2 mean that v2 rtdb emulation only works on the default RTDB instance? or existence of `namespacePattern` imply that it's a v2 trigger?

(whatever ends up being true - can we add a comment explaining this?)

taeold (approved):

src/emulator/functionsEmulator.ts
inlined: No, but I appreciate treating this as untrusted input. We could have new SDK authors pop these assertions because they have bugs in their implementation.

src/emulator/functionsEmulatorRuntime.ts
```diff
@@ -1048,11 +1048,9 @@ async function main(): Promise<void> {
...
           const rawBody = (req as RequestWithRawBody).rawBody;
           let reqBody = JSON.parse(rawBody.toString());
-          if (req.headers["content-type"]?.includes("cloudevent")) {
-            if (EventUtils.isBinaryCloudEvent(req)) {
-              reqBody = EventUtils.extractBinaryCloudEventContext(req);
-              reqBody.data = req.body;
-            }
+          if (EventUtils.isBinaryCloudEvent(req)) {
+            reqBody = EventUtils.extractBinaryCloudEventContext(req);
+            reqBody.data = req.body;
```
inlined, line 1053: I wonder if we actually wanted a third option: check for contenttype to include cloudevent but then throw that structured encoding is not supported if isBinaryCloudEvent is false.

src/emulator/functionsEmulatorShared.ts
```diff
@@ -50,10 +58,11 @@ export interface EventSchedule {
 }
 
 export interface EventTrigger {
-  resource: string;
+  resource: string | undefined;
```
inlined, line 61: `resource?: string`

src/functions/events/v2.ts
```diff
@@ -1,4 +1,4 @@
-export const PUBSUB_PUBLISH_EVENT = "google.cloud.pubsub.topic.v1.messagePublished" as const;
```
inlined, line 1: Do you know why `as const` was there in the first place? Was it used as a string literal type somewhere & isn't now?

src/emulator/functionsEmulator.ts
colerogers: `namespacePattern` implies that we are using the v2 interface. Under the hood, the `?ns=...` query param will be ignored for v2 triggers, but I figured it would be better to show that explicitly here.

colerogers: Added a comment explaining this

src/emulator/functionsEmulator.ts
colerogers: Correct, using the Functions SDK as it currently is will never achieve this state

src/functions/events/v2.ts
colerogers: I'm not sure why we had it like this, maybe leftover from a different implementation?

src/emulator/functionsEmulatorShared.ts
colerogers: Fixed

src/emulator/functionsEmulatorRuntime.ts
colerogers: We are actually removing the check for content type to include cloudevent since it's not in the spec - https://cloud.google.com/eventarc/docs/workflows/cloudevents#payload-format

PR #5046, opened 2022-09-30
Implement library for adding tags to run service based on rewrites

inlined:
One of the last fragment of the monster monolith that is https://github.com/firebase/firebase-tools/pull/5015.

This change defines the library that we'll eventually use in convertConfigs.ts. From the original PR, I found it easier to convert all `pinTag: true` in firebase.json types to `tag: "TODO"` in convertConfig's map comprehension. Then I would make one call to replace all the TODO tags with tags I looked up or set. This allowed me to make a single call to expensive APIs like listVersions for all sites. If this obvious hack is too offensive, I can look at redesigning the runTags library to use a caching iterative solution.


src/hosting/runTags.ts
```diff
@@ -0,0 +1,183 @@
...
+ * from passed in services that don't match a valid tag.
+ * This makes no actual server-side changes to these services; you must then
+ * call run.updateService to save these changes. We divide this responsiblity
+ * because we want to possibly insert a new tagged target before saving.
+ */
+export async function gcTagsForServices(project: string, services: run.Service[]): Promise<void> {
+  // region -> service -> tags
+  // We cannot simplify this into a single map because we might be mixing project
+  // id and number.
+  const validTags: Record<string, Record<string, Set<string>>> = {};
```
bkendall, line 24: Fun pattern I use to carry your comment of "region -> service -> tags" through to code for readability is to name this `validTagsByServiceByRegion`. Then I know `validTagsByServiceByRegion[region][service] = tag` without any questions or referring to other comments. (Long because it's nested, but I think it _is_ nice!)

src/hosting/runTags.ts
```diff
@@ -0,0 +1,183 @@
...
+      validTags[rewrite.run.region][rewrite.run.serviceId] =
+        validTags[rewrite.run.region][rewrite.run.serviceId] || new Set<string>();
+      validTags[rewrite.run.region][rewrite.run.serviceId].add(rewrite.run.tag);
+    }
+  }
+
+  // Erase all traffic targets that have an expired tag and no serving percentage
+  for (const service of services) {
+    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+    console.log(`gcpIds of metadta ${JSON.stringify(service.metadata)} is ${run.gcpIds(service)}`);
```
bkendall, line 45: Remove, or move to `logger.debug`

src/hosting/runTags.ts
```diff
@@ -0,0 +1,183 @@
...
+        if (!traffic.tag.startsWith("fh-")) {
+          return traffic;
+        }
+        if (validTags[region]?.[serviceId]?.has(traffic.tag)) {
+          return traffic;
+        }
+        return null;
+      })
+      // Note: the filter command doesn't update the type info to drop null
+      .filter((t) => t !== null) as run.TrafficTarget[];
```
bkendall, line 67: Why map _then_ filter? Couldn't this just have been `.filter` that checks the same conditions and returns true (otherwise false)? (I saw this elsewhere too and came back to check - I believe this is the only place where the type doesn't change (map returns the same values) which is why it could be skipped)

src/hosting/runTags.ts
```diff
@@ -0,0 +1,183 @@
...
+      throw new FirebaseError(
+        `Assertion failed: service ${service.metadata.name} has no latestRevision traffic target`
+      );
+    }
+    const latestRevision = latestRevisionTarget.revisionName;
+    const alreadyTagged = service.spec.traffic.find(
+      (target) => target.revisionName === latestRevision && target.tag
+    );
+    if (alreadyTagged) {
+      tags[region][serviceId] = alreadyTagged.tag!;
```
bkendall, line 165: It's fairly obvious, but put a comment here that the null assertion is okay 'cause of the above check, please :)

src/test/hosting/runTags.spec.ts
```diff
@@ -0,0 +1,307 @@
...
+    runTags.ensureLatestRevisionTagged.throws(
+      new Error("Unexpected runTags.ensureLatestRevisionTagged")
+    );
+    runTags.gcTagsForServices.rejects(new Error("Unepxected runTags.gcTagsForServices"));
+    runTags.setRewriteTags.rejects(new Error("Unexpected runTags.setRewriteTags call"));
+    runTags.setGarbageCollectionThreshold.restore();
+  });
+
+  afterEach(() => {
+    sinon.verifyAndRestore();
```
bkendall, line 140: I don't see that these tests use mocks (`sinon.mock`), so `sinon.restore()` is sufficient.

bkendall (approved):

inlined:
Well shit. I merged into more-api. I'm going to merge the last one into master, which will include this.

PR #5047, opened 2022-09-30
Wire up pinTag

inlined:
Adds a new alternative input for functions rewrites that encapsulate options into an object. The new functions rewrite form and run rewrites get a pinTag option. When set, we will resolve or add a new tag to the targeted service and send it in our ServingConfig.


inlined:
Holy 🤬 git went sideways on this one. Don't worry about this until we can get the other three PRs merged in. Then I'll do whatever is necessary to get a clean diff.

src/gcp/proto.ts
```diff
@@ -235,3 +235,30 @@ export function formatServiceAccount(serviceAccount: string, projectId: string):
   }
   return `serviceAccount:${serviceAccount}`;
 }
+
+/**
+ * Remove keys whose values are undefined.
+ * When we write an interface { foo?: number } there are three possible
+ * forms: { foo: 1 }, {}, and { foo: undefined }. The latter surprises
+ * most people and make unit test comparison flaky. This cleans up.
```
TheIronDev, line 243: nit: `This cleans that up`.

src/hosting/config.ts
```diff
@@ -224,6 +231,41 @@ export function resolveTargets(
...
+          function: {
+            functionId: rewrite.function,
+            // Do not set pinTag so we can track how often it is used
+          },
+        };
+        delete (modern as unknown as LegacyFunctionsRewrite).region;
+        if ("region" in rewrite && typeof rewrite.region === "string") {
+          modern.function.region = rewrite.region;
+        }
+        // Somebody shoot me... why is it taht these type narrowings compile
```
TheIronDev, line 257: nit: that

src/hosting/config.ts
```diff
@@ -224,6 +231,41 @@ export function resolveTargets(
...
+        // but are somehow necessary?
```
TheIronDev, line 258: I'm guessing because `HostingRewrites` is union typed between a bunch of different things, that while technically legal, you have to give it a strong winky face to get the code to behave (that and we don't know the shape of the config's rewrites when we invoke `normalize()`.

TheIronDev (approved):
LGTM

inlined:
Thanks for the review @TheIronDev. Since this is Hosting, I'm going to also wait for a review by @bkendall.

src/hosting/config.ts
inlined: But `if (typeof rewrite.function === "string")` should indicate that rewrite is a `LegacyFunctionRewrite`, which is where `region` is defined. `rewrite` should have already been typed, but the check against function being a string is only typing that field I think. I may try a different appraoch.

src/gcp/proto.ts
```diff
@@ -235,3 +235,30 @@ export function formatServiceAccount(serviceAccount: string, projectId: string):
...
   return `serviceAccount:${serviceAccount}`;
 }
+
+/**
+ * Remove keys whose values are undefined.
+ * When we write an interface { foo?: number } there are three possible
+ * forms: { foo: 1 }, {}, and { foo: undefined }. The latter surprises
+ * most people and make unit test comparison flaky. This cleans that up.
+ */
+export function pruneUndefiends(obj: unknown): void {
```
bkendall, line 245: Does this have tests?

I'm really confused about what this is solving. The default value of any key that doesn't exist on an object _is_ `undefined`, so when I see below `if (keyable[key] === undefined) delete keyable[key]` I think "well, if a test was testing for the existence of something, this wouldn't have changed anything".  Maybe you can point me to a unit test that is flakey without this function?

src/deploy/functions/backend.ts
```diff
@@ -509,14 +497,13 @@ interface PrivateContextFields {
...
  */
 export async function existingBackend(context: Context, forceRefresh?: boolean): Promise<Backend> {
-  const ctx = context as Context & PrivateContextFields;
-  if (!ctx.loadedExistingBackend || forceRefresh) {
-    await loadExistingBackend(ctx);
+  if (!context.loadedExistingBackend || forceRefresh) {
+    await loadExistingBackend(context);
   }
-  return ctx.existingBackend;
+  return context.existingBackend!;
```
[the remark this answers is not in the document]
inlined: Now that it's grown I could call it LoadFunctionContext. But yes, loadExistingBackend makes existingBackend exist. Added comment.

src/deploy/hosting/context.ts
```diff
@@ -2,9 +2,10 @@ import { HostingResolved } from "../../firebaseConfig";
 import { Context as FunctionsContext } from "../functions/args";
 
 export interface HostingDeploy {
+  // Note: a HostingMultiple[number] is a stronger guarantee than a HostingSingle
+  // because at least one of site and target must exist.
```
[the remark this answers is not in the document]
inlined: Because I added the comment in an earlier PR and in a later PR added a stricter type in response to your comments.

PR #5048, opened 2022-09-30
Rename frameworkawareness to webframeworks

inlined:
What it says on the tin. Would like to get this in the same minor release as experiments since this is a pre-release breaking change which also implies minor.


inlined:
@TheIronDev beat me to it

PR #5052, opened 2022-09-30
moves esbuild to an optional dependency; requires node 16 for frameworks

bkendall:
### Description

`esbuild` as a dependency causes issues with npm@6 since it will fail installs for optional dependencies where engines don't match (from npm-shrinkwrap). Moving `esbuild` to optionalDependencies allows node 14 installs (which default to npm@6) to continue working.

`firebase-frameworks` has a requirement of `node >= 16.0.0` and `npm` will complain if its codepaths are used on lower versions of node. Enforcing a requirement of node 16 or higher in the awareness codepaths will prevent issues from arising there _and_ should give us confidence that `npm` is >=8, where the optional dependencies issue above won't happen either (and therefore `esbuild` should have been installed with `firebase-tools`).

But, as a precaution, we must move and check the import for `esbuild` later in the runtime so that we don't try to load a dependency that may have failed to install (notably on node 14).

### Scenarios Tested

I've been testing with a local nextjs app on nodes 14 and 16 with npms 6 and 8. It's behaving correctly so far as I can tell and installing the package with this change on node 14 does work again.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -25,7 +25,6 @@
         "cross-env": "^5.1.3",
         "cross-spawn": "^7.0.3",
         "csv-parse": "^5.0.4",
-        "esbuild": "^0.15.7",
         "exegesis": "^4.1.0",
         "exegesis-express": "^4.0.0",
         "express": "^4.16.4",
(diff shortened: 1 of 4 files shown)


yuchenshi (approved):

buu700:
I've also been running into issues with npm 8.19.2 and node 16.16.0, which this may have already fixed (not 100% sure). To reproduce the issue, run `npm install firebase-tools esbuild ; npm ci`; it will fail with an error message about a bunch of missing esbuild packages with version 0.15.8.

As a temporary workaround, I created a version of the package with the shrinkwrap file removed (`firebase-tools@github:buu700/firebase-tools-tmp`).

PR #5055, opened 2022-10-01
fix: slow startup if no network is available

### LINKED ISSUE #5054: Painfully slow emulator start if no network is available

**firebase-tools:** 11.9.0

**Platform:** Windows

### [REQUIRED] Test case
Disable WIFI / unplug ethernet and start the emulator. 

Disable WIFI / unplug ethernet and start the emulator. You'll notice the emulator starts very very slowly because the `npm` command (almost) never returns.

https://github.com/firebase/firebase-tools/blob/9e524c18f4038dfcadbab637c9f661165feb1753/src/deploy/functions/runtimes/node/versioning.ts#L62

The emulator actually starts in a reasonable amount of time.

The emulator does not start in < 60s.

stfsy:
Preventing slow startups (> 60s) if no network is available. Tried to pick a reasonable timeout. Currently from Barcelona the NPM commands take 500ms to return. 

### Description
Fixes #5054


src/deploy/functions/runtimes/node/versioning.ts
```diff
@@ -45,6 +46,7 @@ export function getFunctionsSDKVersion(sourceDir: string): string | void {
     const child = spawn.sync("npm", ["list", "firebase-functions", "--json=true"], {
       cwd: sourceDir,
       encoding: "utf8",
+      timeout: NPM_COMMAND_TIMEOUT_MILLIES,
```
colerogers, line 49: We actually removed this spawn in a recent commit, do you mind pulling in the changes from master?

stfsy: @colerogers thank you very much, rebased from master

src/deploy/functions/runtimes/node/versioning.ts
```diff
@@ -16,6 +16,7 @@ interface NpmShowResult {
 }
 
 const MIN_SDK_VERSION = "2.0.0";
+const NPM_COMMAND_TIMEOUT_MILLIES = 3000;
```
colerogers, line 19: We discussed this internally and think that `10000` would be a more appropriate timeout for customers with slower connectivity or an intermittent issue

stfsy: @colerogers ok good point. I updated the value as per your recommendation

colerogers (approved):
Thanks for your contribution!

stfsy:
@colerogers anything I need to? any more prerequisites for it getting merged? ☺

stfsy:
@colerogers Thanks for merging!

PR #5059, opened 2022-10-04
Allow use of the source directory in-lieu of public

jamesdaniels:
validateOnce was not allowing for start of web frameworks


jamesdaniels:
Actually, looks like we have a deeper regression I don't know if this is valid.

PR #5060, opened 2022-10-04
Fix Hosting validation

inlined:
I accidentally forgot to add `config.validate` in `config.hostingConfig` which is a good thing because that utility assumes web frameworks has already pre-processed all hosting configs to turn "source" into "public" where necessary and the method is used by web frameworks itself.

Instead we remove the validate call in `firebase serve` and add a validate call in `firebase deploy`


TheIronDev:
Testing locally I am still seeing deployment failures:

https://webframeworks-test-1.web.app/

TheIronDev (approved):
LGTM

PR #5061, opened 2022-10-05
Enable webframework awareness for Github Action Deploys

TheIronDev:
### Description

This commit enables webframework awareness for Github Action Deploys.

In addition, places where we assert the end-user to enable a feature are omitted when deploying using the Github Action.


TheIronDev:
Closing this in lieu of the better fix found in: #5069

PR #5064, opened 2022-10-05
Fix function deploy test

taeold:
A bad regex in the test condition!

Fixed test result: https://github.com/firebase/firebase-tools/actions/runs/3191733913


TheIronDev (approved):
Whups!!!!

PR #5068, opened 2022-10-06
Pass singleProjectMode config down to the database emulator and bump the version to 4.10.0.

christhompsongoogle:
Pass singleProjectMode config down to the database emulator and bump the version to 4.10.0.


bkendall (approved):

fredzqm (approved):
LG

PR #5069, opened 2022-10-06
feat(experiments) Add support for FIREBASE_CLI_EXPERIMENTS

TheIronDev:
### Description

Enables the ability to set experiments via env variables.

The intention behind this is to enable features during Github CI/CD hooks.

### Scenarios Tested

* Enabling an experiment via env var
* Attempting to enable a non-experiment via Env variable (it doesn't work)

### Sample Commands

Locally

```bash
FIREBASE_CLI_EXPERIMENTS=experiment1,experiment2 firebase deploy
```

Github Action

```yml
  - uses: FirebaseExtended/action-hosting-deploy@v0
     with:
       # ...
     env:
       FIREBASE_CLI_EXPERIMENTS: webframeworks,pintag
```


src/test/experiments.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { expect } from "chai";
+import {
+  enableExperimentsFromCliEnvVariable,
+  getSnapshotOfExperimentPreferences,
+} from "../experiments";
+
+describe("experiments", () => {
+  describe("enableExperimentsFromCliEnvVariable", () => {
+    it("should enable some experiments", () => {
+      process.env.FIREBASE_CLI_EXPERIMENTS = "experiments,not_an_experiment";
```
bkendall, line 10: This will persist into other tests after this is run. Consider creating a `before` that either stores or sets this to `""`, then in the test set it and run the function, then restore or reset to `""` in an `after`. ~Probably will have to run the `enableExperimentsFromCliEnvVariable` again too to reset it.~ Ah, it's not persisted to disk if flush isn't called. Nice break.

(this is one of the things I don't like about env var functionality - it's really tricky when it comes to things like this)

src/experiments.ts
```diff
@@ -164,6 +164,33 @@ export function setEnabled(name: ExperimentName, to: boolean | null): void {
...
 }
 
+/**
+ * Returns the unsaved set of preferences.
+ */
+export function getSnapshotOfExperimentPreferences() {
+  return {
+    ...localPreferences(),
+  };
+}
```
bkendall, line 174: I don't love having to add this just for testing. Could `isEnabled` be used to check? That's how the real code would do it and it's probably wise to match that anyway.

bkendall (changes requested):

src/test/experiments.spec.ts
TheIronDev: Yeah I didn't realize the flush disk thing at first either, had a bunch of code to side-step mutating the config.

Was great.... ....

Anyway, added some beforeEach / afterEach to reset the env variable.

src/experiments.ts
TheIronDev: It could! Removed `getSnapshot` and am using `isEnabled` instead.

src/bin/firebase.ts
```diff
@@ -161,7 +164,7 @@ if (!handlePreviewToggles(args)) {
 
   // determine if there are any non-option arguments. if not, display help
   args = args.filter((arg) => {
-    return arg.indexOf("-") < 0;
+    return !arg.includes("-");
   });
   if (!args.length) {
```
bkendall, line 169: Oh man, I hate when I'm like "but you can make it _even better_". totally optional, and I'm 90% confident this is right (please check), but:

```typescript
const hasNotFlagParams = args.some((a) => a.includes("-"));
if (!hasNotFlagParams) {
```

src/test/experiments.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  before(() => {
+    originalCLIState = process.env.FIREBASE_CLI_EXPERIMENTS;
+  });
+
+  beforeEach(() => {
+    process.env.FIREBASE_CLI_EXPERIMENTS = originalCLIState;
+  });
+
+  afterEach(() => {
+    process.env.FIREBASE_CLI_EXPERIMENTS = originalCLIState;
```
bkendall, line 16: This still leaves the experiments enabled that are manipulated in the test. Should this run through all the experiments and reset them to not enabled, then call `enableExperimentsFromCliEnvVariable` after resetting the env var?

src/bin/firebase.ts
TheIronDev: I think I hit the 10% in finding they are not functional equivalent. 

```
// before
args = ['x', '-']
args = args.filter((arg) => !arg.includes("-")); // ['x']
!x.length; // false

// after
args = ['x', '-']
hasNotFlagParams =args.some((a) => a.includes("-"));
!hasNotFlagParams; // true
```

src/test/experiments.spec.ts
TheIronDev: True. I decided to just manually reset "experiments" in the test. 🤷

src/bin/firebase.ts
TheIronDev: I think I'm going to leave as-is for now. (The code change happened automagically from IntelliJ 🤷 )

bkendall (approved):

bolds07:
there should be a file in documentation or even a section on readme to find a list of all env vars... it is basically impossible to someone guess it.

I could only find this because of someone on stackoverflow

wiktor-cobry:
Having trouble adding this parameter to the workflow.

```YAML
jobs:
  build_and_deploy:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - run: npm ci && npm run build
      - uses: FirebaseExtended/action-hosting-deploy@v0
        with:
          repoToken: '${{ secrets.GITHUB_TOKEN }}'
          firebaseServiceAccount: '${{ secrets.FIREBASE_SERVICE_ACCOUNT_redacted}}'
          channelId: live
          projectId: redacted
          env:
            FIREBASE_CLI_EXPERIMENTS: webframeworks
```

I'm getting a linting error:  'A mapping was not expected' under the Firebase CLI Experiments. I'm not too familiar with YAML so assistance would be appreciated.

PR #5077, opened 2022-10-06
Fix Hosting deploys

bkendall:
### Description

1. `TypeError: Cannot read properties of undefined (reading 'join')`

An import that was `import path from "path"` wasn't working correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,2 @@
-- Enable single project mode for the auth emulator (#5068).
+- Enables single project mode for the auth emulator (#5068).
+- Fixes issue deploying to Hosting with i18n enabled.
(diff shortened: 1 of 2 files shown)


inlined (approved):
Curious regression. I'm wondering how this was avoided in the original code (maybe it already did the respective check)

TheIronDev (approved):
LGTM!

src/deploy/hosting/convertConfig.ts
```diff
@@ -79,18 +80,44 @@ export function findEndpointForRewrite(
 
 /**
  * convertConfig takes a hosting config object from firebase.json and transforms it into
- * the valid format for sending to the Firebase Hosting REST API
+ * the valid format for sending to the Firebase Hosting REST API.
+ *
+ * TODO: this currently lists remote backends (functions) and attemtps to validate them.
+ * We currently catch 403 issues and handle them, but it's probably not the best solution
+ * to have a required permission in functions when a deploy may "only" be to Hosting.
```
steveoh, line 87: **suggestion**: convert this to an issue

steveoh (commented):
Thanks!

drummerjolev:
Thanks @bkendall! Any ETA on the version release? I'm using the Firebase GH action, which has been broken since this morning.

Though https://github.com/FirebaseExtended/action-hosting-deploy/issues/242 could also be a solution.

PR #5078, opened 2022-10-06
Set environment variable necessary to be a custom events source

inlined:
The admin SDK does/will read the CloudEvent source field from an environment variable so that firebase can inject an experiment ID, function ID, or service ID rather than requiring customers to encode their deployment name in their codebase (not even possible for extensions).


taeold (approved):
🤔 Does emulator need similar treatment?

inlined:
Yes it does. Where would I put that?

PR #5079, opened 2022-10-06
Remove getInfoHostString and use connectable hostnames.

yuchenshi:
### Description

This is a refactor with a subtle behavior change to use correctly-formatted URLs with hostname fixes. This marks the start of a quest to fix IPv6 and DNS related problems.

When given a `0.0.0.0` hostname, we'll try to connect over `127.0.0.1` instead to make sure max client compatibility. `EmulatorRegistry.getInfo` now returns the transformed hostname by default, since this is usually the desirable behavior. A new helper method is added for creating an API client to an emulator -- a common enough pattern that worth standardizing. This PR also uses `EmulatorRegistry.url` for most URL constructions and has removed `getInfoHostString`, which causes more confusions than it helps.

### Scenarios Tested

Unit test added for the new `connectableHostname` helper.

Basic manual testing like starting the emulators. 

I'll rely on our CI to catch any remaining errors.

### Sample Commands

N/A


src/commands/emulators-start.ts
```diff
@@ -95,22 +95,23 @@ function printEmulatorOverview(options: any): void {
...
+          url.pathname = `/${emulator}`;
+          uiLink = stylizeLink(url.toString());
+        }
 
-        return [
-          emulatorName,
-          EmulatorRegistry.getInfoHostString(info),
-          isSupportedByUi && uiInfo ? stylizeLink(`${uiUrl}/${emulator}`) : clc.blackBright("n/a"),
-        ];
+        return [emulatorName, info.host, info.port, uiLink];
```
bkendall, line 109: I'm certain the reason we had the full links before was so that they were clickable from a terminal - I remember it being a specific request! Could we maintain that?

src/emulator/commandUtils.ts
```diff
@@ -276,17 +276,19 @@ function processKillSignal(
         const pids: number[] = [];
 
         const emulatorsTable = new Table({
-          head: ["Emulator", "Host:Port", "PID"],
+          head: ["Emulator", "Host", "Port", "PID"],
```
bkendall, line 279: Similar here - this will make things un-clickable. Could we maintain that?

src/emulator/controller.ts
```diff
@@ -502,14 +501,7 @@ export async function startAll(
...
     experiments.assertEnabled("webframeworks", "emulate a web framework");
-    const emulators: EmulatorInfo[] = [];
-    if (experiments.isEnabled("webframeworks")) {
-      for (const e of EMULATORS_SUPPORTED_BY_UI) {
-        const info = EmulatorRegistry.getInfo(e);
-        if (info) emulators.push(info);
-      }
-    }
-    await prepareFrameworks(targets, options, options, emulators);
+    await prepareFrameworks(targets, options, options);
```
bkendall, line 504: We will probably need to explicitly test that the frameworks stuff works. It looks fine, but it's sensitive to disruption right now.

src/emulator/pubsubEmulator.ts
```diff
@@ -25,7 +25,7 @@ interface Trigger {
 }
 
 export class PubsubEmulator implements EmulatorInstance {
-  pubsub: PubSub;
+  private _pubsub: PubSub | undefined;
```
bkendall, line 28: it's redundant to prefix private fields with `_`. Please omit.

bkendall (changes requested):
I think we'll want to manually test some webframeworks stuff here for sanity.

src/emulator/pubsubEmulator.ts
yuchenshi: The getter is also called `pubsub` and we'd have a naming conflict if `_` were to be removed.

src/commands/emulators-start.ts
yuchenshi: Good catch! I've restored the behavior. Added some comments too for background.

src/emulator/commandUtils.ts
yuchenshi: Good catch! I've restored the behavior. Added some comments too for background.

src/emulator/controller.ts
yuchenshi: As discussed offline, I've reverted my changes to frameworks and added a comment on the code in question.

src/emulator/controller.ts
```diff
@@ -505,6 +505,8 @@ export async function startAll(
     const emulators: EmulatorInfo[] = [];
     if (experiments.isEnabled("webframeworks")) {
       for (const e of EMULATORS_SUPPORTED_BY_UI) {
+        // TODO: Double check if this actually works -- we're early in the startup
```
christhompsongoogle, line 508: nit: add name/ldap

src/emulator/functionsEmulator.ts
```diff
@@ -1149,41 +1132,11 @@ export class FunctionsEmulator implements EmulatorInstance {
...
-    const pubsubEmulator = this.getEmulatorInfo(Emulators.PUBSUB);
-    if (pubsubEmulator) {
-      const pubsubHost = formatHost(pubsubEmulator);
-      process.env.PUBSUB_EMULATOR_HOST = pubsubHost;
-    }
 
+    setEnvVarsForEmulators(envs);
     const eventarcEmulator = this.getEmulatorInfo(Emulators.EVENTARC);
     if (eventarcEmulator) {
+      // TODO: Why do we set the protocol here, but no protocol during emulators:exec?
```
christhompsongoogle, line 1139: nit: add name/ldap

christhompsongoogle:
Question: What's the semantic difference between "isRunning" and "getInfo" - is info populated before isRunning? Why switch from one to the other?

christhompsongoogle (approved):

src/emulator/controller.ts
yuchenshi: I don't know the right contact yet but I've started a thread for it internally. If I cannot identify a contact, then I'll just create an internal bug for this TODO.

yuchenshi:
> Question: What's the semantic difference between "isRunning" and "getInfo" when checking if an Emulator is running/active/enabled - is info populated before isRunning? Why switch from one to the other?

It is more or less equivalent, and there's no actual difference in timing. But `isRunning` is preferred because it captures the intention better.

Deeper reason: When using `getInfo`, if later one needs to construct a URL, they will naturally get the `host` and `port` from the info and try simple string interpolation, which works most of the time but fails in the edge cases in PR description. It is much safer to just do a boolean test and use `EmulatorRegistry.url` or `EmulatorRegistry.client` later if needed.

Future consideration: I'm about to create follow-up PRs to change the way emulator listens on hosts (with DNS in mind) and it is necessary to audit the situations where the `host` is absolutely needed. Less `getInfo` means less places to track down and potentially modify later.

Side benefit: `isRunning` does less work than `getInfo` -- a bit more performance, less moving pieces.

bkendall (approved):

PR #5083, opened 2022-10-07
Assign emulator ports and resolve hostnames upfront.

yuchenshi:
### Description

This PR explicitly resolves hostnames like `localhost` to IP addresses like `127.0.0.1`, `::1` before listening for each emulator and assigns all the ports all at once early in the process.

It also modifies the Emulator Hub to listen on as many addresses as it can to increase the chance of successful discovery (no matter the client is v4- or v6-only). If we like what we're seeing here, we can apply this to most other Node.js-based emulators (see comments inline).

The new test process use the more reliable net Socket instead of `tcp-port-used`. There's only one instance of it left to test for port closed (occupied by emulator), and I wonder if we still need it. (Also, the new test process is more or less similar to `portfinder` and maybe we can remove the dependency of it one day, after we update all the other usages.)

Since the ports are all allocated early, this may fix the frameworks autoinit integration.

### Scenarios Tested

Manually started the emulators, observed what ports are allocated with `--debug`, including when ports are partially occupied. ProTip: You can occupy a specific address with `python3 -m http.server --bind ::1 9999` IPv4 (or `127.0.0.1` for IPv4). I've tested that when the preferred IPv4 address is occupied, the process gives an error. A warning is given when secondary addresses are occupied.


src/emulator/portUtils.ts
```diff
@@ -149,3 +160,249 @@ export async function waitForPortClosed(port: number, host: string): Promise<voi
...
+  // bind to only one of the addresses resolved from hostname.
+  database: true,
+  firestore: true,
+  "firestore.websocket": true,
+  pubsub: true,
+
+  // Listening on multiple addresses to maximize the chance of discovery.
+  hub: false,
+
+  // TODO: Modify the following emulators to listen on multiple addresses.
```
yuchenshi, line 177: Q: Do want this to happen in the same release or later?

bkendall (approved):

christhompsongoogle (approved):

PR #5086, opened 2022-10-07
Small fixes for run tag pinning

inlined:
Few small changes:

1. When I wrote this code I thought `version` was an ID like most other components we get from the API library. Unfortunately it's actually a name, so using a name as a tag didn't work. We now strip the ID out of the name and pass it to the runTags library
2. As a minor simplification/resiliency, I'm getting the latest revision name from the Service status directly rather than the traffic target status. This required a bugfix to correct the name of the field.
3. When testing e2e, Hosting was creating a new (failed) revision while modifying traffic target. This is because updateService was deleting the template's name. This is necessary when modifying the template to avoid conflicts, but if you do it any other time, you'll spin up a new revision because the template doesn't match the previous version. So I moved the delete call from updateFunction and into fabricator (where we actually do intend to modify the template)


mbleigh (approved):
I'm not nagging on the lint warnings since they aren't changed lines, but should I be?

inlined:
One could have been made better, but in general those are warnings because we decided to catch `any` for errors rather than `unknown` and `Endpoint` has a lot of fields that are optional but filled in during the deploy process.

Issue #215, opened 2016-09-02
List and delete deployments


Nowaker:
I've just noticed by Firebase hosting usage is now almost 1 GB. Quite a lot given the fact our website is only 20 MB.

```
nowaker@nwkr-desktop ~/projekty/virtkick/website (git)-[master] % du -hs build 
20M     build
```

It seems all previous deployments are kept by Firebase and they're visible at https://console.firebase.google.com/project/PROJECTNAME/hosting/main.

Deleting 90 deployments one by one manually is out of question. There's a big need to have a way to list deployments via CLI and delete them.

brendanlim:
@Nowaker This is definitely on our radar and something we're looking to improve

zhenhaolei:
hey @brendanlim any updates here? We are also seeing infinite loading in the hosting module, looks like we have too many deployments :( 

Thanks! 

`Error: too_big: The data requested exceeds the maximum size that can be accessed with a single request.
    at r (rs=AON9PunUVzISsT9OTTHxul9qDyFtbYjNGA:8597)
    at H (third_party/javascript/firebase/firebase_js_minified.jslib:126)
    at Object.eval [as H] (third_party/javascript/firebase/firebase_js_minified.jslib:206)
    at eval (third_party/javascript/firebase/firebase_js_minified.jslib:190)
    at Kh.g.Id (third_party/javascript/firebase/firebase_js_minified.jslib:196)
    at yh.Id (third_party/javascript/firebase/firebase_js_minified.jslib:186)
    at qh.eval [as zg] (third_party/javascript/firebase/firebase_js_minified.jslib:184)
    at th (third_party/javascript/firebase/firebase_js_minified.jslib:178)
    at WebSocket.ua.onmessage (third_party/javascript/firebase/firebase_js_minified.jslib:177)
    at WebSocket.b [as __zone_symbol___onmessage] (rs=AON9PunUVzISsT9OTTHxul9qDyFtbYjNGA:8568)
    at w.invokeTask (rs=AON9PunUVzISsT9OTTHxul9qDyFtbYjNGA:8611)
    at u.runTask (rs=AON9PunUVzISsT9OTTHxul9qDyFtbYjNGA:8601)
    at WebSocket.invoke (rs=AON9PunUVzISsT9OTTHxul9qDyFtbYjNGA:8613)
`

Nowaker:
[...]

PR #5087, opened 2022-10-07
add additional site config information to hosting:sites:list

bkendall:
### Description

There's useful information in Hosting's API that we can present in the CLI - specifically if Cloud Logging is enabled or how many versions are retained. We should display that!

image

This is a tiny step towards addressing #215 which is _ooollllddddd_ haha. But to be able to view this in the CLI would be good.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,3 @@
 - Enable single project mode for the database emulator (#5068).
 - Ravamp emulator networking to assign ports early and explictly listen on IP addresses (#5083).
+- Adds additional Hosting site information to `hosting:sites:list`.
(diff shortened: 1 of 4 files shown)


EvanCesanek:
This is a great idea! Two suggestions/requests:

1. Please include this information in the return value of `firebase hosting:list:sites`, rather than simply displaying it. This is important for using the CLI as a module.
2. Please allow access to even more of the information available via the Hosting API. For instance, I'd love a command like `firebase hosting:list:releases <siteId>` that returns the output of ``apiClient.get(`/projects/${project}/sites/${site}/releases`)``. Alternatively, to avoid introducing another command, maybe return some key information about the current release for each site in `firebase hosting:list:sites` (_e.g._, **message**, releaseTime, version.status).

PR #5090, opened 2022-10-07
Fix Functions emulator not starting for frameworks.

yuchenshi:
### Description

Fix regression introduced by #5083 where code block with side effects were moved.

### Scenarios Tested

Started a project with no functions but with frameworks, and the Functions emulator now start correctly.

### DIFF
--- a/src/emulator/controller.ts
+++ b/src/emulator/controller.ts
@@ -349,18 +349,23 @@ export async function startAll(
   }
 
   const listenConfig = {} as Record<PortName, EmulatorListenConfig>;
+  if (emulatableBackends.length) {
+    // If we already know we need Functions (and Eventarc), assign them now.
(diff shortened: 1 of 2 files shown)


src/emulator/controller.ts
```diff
@@ -324,29 +324,9 @@ export async function startAll(
   }
 
   const emulatableBackends: EmulatableBackend[] = [];
-  const projectDir = (options.extDevDir || options.config.projectDir) as string;
```
yuchenshi, line 327: This block has been moved back to where it was

TheIronDev:
Any chance you can write some tests to reduce further regression opportunities?

bkendall (approved):

yuchenshi:
> Any chance you can write some tests to reduce further regression opportunities?

I have some ideas -- I think we need something like the existing emulator integration tests for the web frameworks in general, but that is a bigger ask. Happy to discuss more offline.

PR #5091, opened 2022-10-07
Check for functions permissions if web frameworks adds a functions dep

inlined:
Previously frameworks might add functions as a deploy target and we will not have called requirePermissions on the functions permissions. This ensures that those permissions are checked before we proceed with the deploy.

### DIFF
--- a/src/deploy/index.ts
+++ b/src/deploy/index.ts
@@ -17,6 +17,8 @@ import * as RemoteConfigTarget from "./remoteconfig";
 import * as ExtensionsTarget from "./extensions";
 import { prepareFrameworks } from "../frameworks";
 import { HostingDeploy } from "./hosting/context";
+import { requirePermissions } from "../requirePermissions";
+import { TARGET_PERMISSIONS } from "../commands/deploy";
 
 const TARGETS = {
   hosting: HostingTarget,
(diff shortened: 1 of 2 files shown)


TheIronDev (approved):

src/deploy/index.ts
```diff
@@ -61,7 +63,12 @@ export const deploy = async function (
     const config = options.config.get("hosting");
     if (Array.isArray(config) ? config.some((it) => it.source) : config.source) {
       experiments.assertEnabled("webframeworks", "deploy a web framework to hosting");
+      const usedToTargetFunctions = targetNames.includes("functions");
       await prepareFrameworks(targetNames, context, options);
+      const nowTargetsFunctions = targetNames.includes("functions");
+      if (nowTargetsFunctions && !usedToTargetFunctions) {
+        await requirePermissions(TARGET_PERMISSIONS["functions"]);
```
bkendall, line 70: This is going to break every GH action deployment using webframeworks because they don't have the permissions to deploy functions (such is my understanding). We should have a better story of how to upgrade those users' SAs before we push this change.

bkendall: Then again, if they're broken _now_, that's unlikely to matter.

inlined: We need to focus on a sprint to give customers preview channels for web frameworks + pinned tags. I can play with the code, but I don't know how to test outside of production.

PR #5101, opened 2022-10-10
Fix web framework options not applied in hosting emulator.

yuchenshi:
### Description

Fix a few webframeworks things like rewrites, redirects, and headers (including autoinit cookie) not working in the hosting emulator.

The previous implementation modifies the the result from `hostingConfig` (cached in `options.normalizedHostingConfig`), which is not used in the emulator code path. I've added a new codepath `hostingConfigForEmulator` in this PR and handles it correctly.

### Scenarios Tested

Started the emulator in a new Next.js app with the following code added:

```js
// ...
import { useEffect } from 'react'
import { getFunctions } from 'firebase/functions'
import { getAuth } from 'firebase/auth'

export default function Home() {
  useEffect(() => {
    console.log(getFunctions());
    console.log(getAuth());
  }, []);
  // ...
}
// ...
```

The app started just fine, and the console log indicates that the Auth emulator is used. `curl` to a JavaScript file shows the right cookie, and a non-existent path gets rewrited to the catch-all function (as indicated by CLI logs).

### Sample Commands

`firebase emulators:start`

### DIFF
--- a/src/frameworks/index.ts
+++ b/src/frameworks/index.ts
@@ -11,7 +11,7 @@ import * as process from "node:process";
 import * as semver from "semver";
 
 import { needProjectId } from "../projectUtils";
-import { hostingConfig } from "../hosting/config";
+import { hostingConfig, hostingConfigForEmulator } from "../hosting/config";
 import { listSites } from "../hosting/api";
 import { getAppConfig, AppPlatform } from "../management/apps";
 import { promptOnce } from "../prompt";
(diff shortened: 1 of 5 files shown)


inlined:
This is less surgical than #5102. Let's use that instead (though you do admittedly fix a warning error that says "deploy" when it might be emulating, that error message should now never happen)

yuchenshi:
Obsoleted by #5102.

PR #5104, opened 2022-10-10
Block deploying web frameworks' SSR to preview channels

inlined:
Traditionally, hosting does not automatically deploy functions in rewrites. This means that when someone deploys to a preview channel, their static assets are deployed but their dynamic content still points to prod.

Unlike traditional hosting for the last 6 years, web frameworks _always deploy with their ssr function if necessary_. This means that deploying to a preview channel would modify prod!!

This quick PR blocks deploys of SSR'd hosting sites to preview channels to prevent this bug.

### DIFF
--- a/src/deploy/index.ts
+++ b/src/deploy/index.ts
@@ -67,6 +67,11 @@ export const deploy = async function (
       await prepareFrameworks(targetNames, context, options);
       const nowTargetsFunctions = targetNames.includes("functions");
       if (nowTargetsFunctions && !usedToTargetFunctions) {
+        if (context.hostingChannel && !experiments.isEnabled("pintags")) {
+          throw new FirebaseError(
+            "Web frameworks with dynamic content do not yet support deploying to preview channels"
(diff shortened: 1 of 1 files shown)


bkendall (approved):

joenewbry:
Hey wondering if there will ever be able to deploy a web frameworks with SSR to a preview channel?

inlined:
It's something we'd very much like to support. Keep watching our release notes/announcements

inlined:
Hopefully that wasn't too long to wait!

At Google I/O we released "dynamic preview channels". In traditional hosting, you enable this by adding `"pinTag": true` in the (2nd gen) functions or run rewrite. In web frameworks, this is auto-enrolled with the latest CLI. To get this new behavior, run `npm install --global firebase-tools`. If you're using our GitHub action for CI/CD, it won't currently be granted the functions permissions that are now needed; to fix this, reinstall the GitHub action with `firebase init`.

PR #5109, opened 2022-10-11
fix missing message on deploy

### LINKED ISSUE #5107: Firebase Hosting: `firebase deploy` ignores the -m argument in version 11.14.1
In version 11.14.1 of `firebase-tools`, the `firebase deploy` command seems to completely ignore the -m argument (description), and the description in empty in the Firebase Hosting dashboard.
Version 11.13.0 was working fine in that regard.

I believe this change to be the offending one, where it seems that the release message was dropped in the new api.createRelease function

https://github.com/firebase/firebase-tools/commit/3a92daf7d663afb3d08790640f44bc77435c9d4e#diff-7d0f3d70809a7e16334fc0ebd9d0b60757f72a416dd11397b62c37ef41b5ed1fL51

bkendall:
### Description

The message was dropped (because Options was removed?) on release. This restores that functionality and adds tests!

### Scenarios Tested

image

Fixes #5107

### DIFF
--- a/src/hosting/api.ts
+++ b/src/hosting/api.ts
@@ -451,6 +451,8 @@ export async function cloneVersion(
   return pollRes;
 }
 
+type PartialRelease = Partial<Pick<Release, "message" | "type">>;
+
 /**
  * Create a release on a channel.
  * @param site the site for the version.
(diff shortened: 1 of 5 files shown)


src/hosting/api.ts
```diff
@@ -460,13 +460,14 @@ export async function cloneVersion(
 export async function createRelease(
   site: string,
   channel: string,
-  version: string
+  version: string,
+  partialRelease?: Partial<Pick<Release, "message" | "type">>
```
taeold, line 464: q: I notice that types don't necessarily match here and in the `.post` call:

```
Partial<Pick<Release, "message" | "type">>
Partial<Pick<Release, "message" >>
```

Intentional? Do you want to define an intermediate type instead of retyping this fairly long type def?

taeold (approved):

inlined (approved):

src/hosting/api.ts
bkendall: Yeah, meant to make a new type, then didn't. Thanks.

PR #5118, opened 2022-10-12
Unbreak default ints in text input

Berlioz:


taeold (approved):

src/deploy/functions/params.ts
```diff
@@ -590,7 +590,7 @@ async function promptText<T extends RawParamValue>(
   resolvedDefault: T | undefined,
   converter: (res: string) => T | retryInput
 ): Promise<T> {
-  const res = await promptOnce({
+  const res: any = await promptOnce({
```
inlined, line 593: why not `typeof resolvedDefaut | string`? Or just call `resolvedDefautl.toString()` in the prompt call?

src/deploy/functions/params.ts
```diff
@@ -605,7 +605,7 @@ async function promptText<T extends RawParamValue>(
       return promptText<T>(prompt, input, resolvedDefault, converter);
     }
   }
-  const converted = converter(res);
+  const converted = converter(typeof res === "string" ? res : res.toString());
```
inlined, line 608: fun fact: `string` also implements `toString` so that you can avoid ternaries like this.

inlined (approved):
Works, but left some nits on style.

joehan (approved):

src/deploy/functions/params.ts
Berlioz: I didn't want to do the .toString() thing in this call because I wanted as small a footprint of a change in an emergency patch as possible; I'd agree that it'd probably be just as good/better.

The :any annotation is unnecessary with that tip you gave about string.toString() below, since all it's doing is reassuring tsc that it's not going to have to call `never.toString()` down below. I'm getting rid of it.

src/deploy/functions/params.ts
Berlioz: You know, I knew that, but I apparently didn't _know_ that.

PR #5137, opened 2022-10-17
List Params support in CEL

Berlioz:
"This creaky regular expression based hack of an interpreter for our limited CEL subset can't possibly handle more features!," said Frog, as he built another feature on top of it


src/deploy/functions/params.ts
```diff
@@ -444,6 +497,52 @@ async function promptParam(
...
+    }
+    prompt += "\nSelect an option with the arrow keys, and use Enter to confirm your choice. ";
+    return promptSelectMultiple<string>(
+      prompt,
+      param.input,
+      resolvedDefault,
+      (res: string[]) => res
+    );
+  } else if (isTextInput(param.input)) {
+    prompt = `Enter a list of strings for ${param.label || param.name}:`;
```
inlined, line 534: This will probably get more complex if/when we transition to inputting with a delimiter. If your delimiter is newline, you probably have to accept double newlines as the way you end your prompt. IDK if the prompt library supports that, so we _might_, _for now_ stick to just comma delimiters.

src/deploy/functions/params.ts
```diff
@@ -584,6 +688,10 @@ async function promptResourceString(
 }
 
 type retryInput = { message: string };
+function shouldRetry(obj: any): obj is retryInput {
+  return (obj as retryInput).message !== undefined;
```
inlined, line 692: Do you also need to verify that obj is indeed an object? It looks like the previous code this is replacing used to check for that.

src/deploy/functions/params.ts
```diff
@@ -636,9 +744,36 @@ async function promptSelect<T extends RawParamValue>(
...
   return converted;
 }
+
+async function promptSelectMultiple<T extends string>(
+  prompt: string,
+  input: ListSelectInput,
+  resolvedDefault: T[] | undefined,
+  converter: (res: string[]) => T[] | retryInput
+): Promise<T[]> {
+  const response = await promptOnce({
```
inlined, line 760: In firebase init we explain to people that they should use space to select a list item and return to choose the selection. We should probably add that to the prompt here. Similarly, when using a text input for list items, we should explain the delimiter usage.

src/deploy/functions/runtimes/discovery/parsing.ts
```diff
@@ -106,6 +107,14 @@ export function assertKeyTypes<T extends object>(
       }
       continue;
     }
+    if (schemaType === "List") {
```
inlined, line 110: Technically you've defined "List?" as a type. Do you want to check for value === null first or is that case already handled?

src/test/deploy/functions/cel.spec.ts
```diff
@@ -3,16 +3,168 @@ import { resolveExpression, ExprParseError } from "../../../deploy/functions/cel
...
+          BAR: stringV("asdf"),
+        })
+      ).to.deep.equal(["asdf"]);
+    });
+
+    it("can handle a list with multiple elements", () => {
+      expect(
+        resolveExpression(
+          "string[]",
+          '{{ params.FOO == params.FOO ? [ "foo", params.BAR, {{ params.BAR }} ] : [] }}',
```
inlined, line 73: Shouldn't the last field be `"{{ params.BAR }}"` (with the quotes)

inlined (commented):
The winds are blowing in the direction of making lists be delimited strings rather than JSON expressions. This would make it easier to represent in GUI dialogs and reads easier in .env files.

src/deploy/functions/params.ts
Berlioz: Good catch.

src/deploy/functions/runtimes/discovery/parsing.ts
Berlioz: It looks to me like the conditional checks immediately above that one cover this case, unless I'm badly misinterpreting your comment.

src/deploy/functions/cel.ts
```diff
@@ -94,11 +107,72 @@ export function resolveExpression(
...
+ * - A reference to a param value (params.FOO) which must resolve with type string
+ * - A sub-CEL expression {{ params.BAR == 0 ? "a" : "b" }} which must resolve with type string
+ */
+function resolveList(
+  wantType: "string",
+  list: string,
+  params: Record<string, ParamValue>
+): string[] {
+  if (!list.startsWith("[") || !list.endsWith("]")) {
+    throw new ExprParseError("");
```
taeold, line 143:
```suggestion
    throw new ExprParseError("Invalid list: must start with '[' and end with ']'");
```

src/deploy/functions/cel.ts
```diff
@@ -371,7 +466,20 @@ function resolveLiteral(wantType: L, value: string): Literal {
     );
   }
 
-  if (wantType === "number") {
+  if (wantType === "string[]") {
+    // N.B: value being a literal list that can just be JSON.parsed should be guaranteed
+    // by the preprocessLists() invocation at the beginning of CEL resolution
+    const parsed = JSON.parse(value);
+    if (!Array.isArray(parsed)) {
+      throw new ExprParseError("");
```
taeold, line 474: Can we include more descriptive parser error message?

src/deploy/functions/params.ts
```diff
@@ -206,15 +244,31 @@ export class ParamValue {
...
+    this.delimiter = ",";
+  }
+
+  static fromList(ls: string[], delimiter = ","): ParamValue {
+    const pv = new ParamValue(ls.join(delimiter), false, { list: true });
+    pv.setDelimiter(delimiter);
+    return pv;
+  }
+
+  setDelimiter(delimiter: string) {
```
taeold, line 270: nit* feels very Java-ish 🤔

src/deploy/functions/params.ts
```diff
@@ -444,6 +509,50 @@ async function promptParam(
...
+    }
+    prompt += "\nSelect an option with the arrow keys, and use Enter to confirm your choice. ";
+    return promptSelectMultiple<string>(
+      prompt,
+      param.input,
+      resolvedDefault,
+      (res: string[]) => res
+    );
+  } else if (isTextInput(param.input)) {
+    prompt = `Enter a list of strings for ${param.label || param.name}:`;
```
taeold, line 538: can we expand the prompt to help instruct users how they can input the delimeter-separated list input?

taeold (approved):

src/deploy/functions/params.ts
```diff
@@ -123,6 +143,12 @@ export function isSelectInput<T>(input: ParamInput<T>): input is SelectInput<T>
 export function isResourceInput<T>(input: ParamInput<T>): input is ResourceInput {
   return {}.hasOwnProperty.call(input, "resource");
 }
+/**
+ * Determines whether an Input field value can be coerced to ListSelectInput.
```
taeold, line 147: The documentation and function name doesn't seem to match

```suggestion
 * Determines whether an Input field value can be coerced to MultiSelectInput.
```

src/deploy/functions/params.ts
```diff
@@ -532,8 +653,9 @@ async function promptIntParam(param: IntParam, resolvedDefault?: number): Promis
       }
       return +res;
     });
-  }
-  if (isTextInput(param.input)) {
+  } else if (isMultiSelectInput(param.input)) {
+    throw new FirebaseError("Non-list params cannot have list selector inputs");
```
taeold, line 657:
```suggestion
    throw new FirebaseError("Non-list params cannot have multi selector inputs");
```

src/deploy/functions/params.ts
```diff
@@ -492,6 +611,8 @@ async function promptStringParam(
       prompt += ` \n(${param.description})`;
     }
     return promptResourceString(prompt, param.input, projectId, resolvedDefault);
+  } else if (isMultiSelectInput(param.input)) {
+    throw new FirebaseError("Non-list params cannot have list selector inputs");
```
taeold, line 615:
```suggestion
    throw new FirebaseError("Non-list params cannot have multi selector inputs");
```

src/deploy/functions/params.ts
```diff
@@ -462,6 +579,8 @@ async function promptBooleanParam(
     }
     prompt += "\nSelect an option with the arrow keys, and use Enter to confirm your choice. ";
     return promptSelect<boolean>(prompt, param.input, resolvedDefault, isTruthyInput);
+  } else if (isMultiSelectInput(param.input)) {
+    throw new FirebaseError("Non-list params cannot have list selector inputs");
```
taeold, line 583:
```suggestion
    throw new FirebaseError("Non-list params cannot have multi selector inputs");
```

src/deploy/functions/params.ts
taeold: +1

PR #5140, opened 2022-10-18
Track which web framework is deployed

inlined:
Adds a track command for the "hosting_deploy" metric with a label for the kind of site deployed. If it's not a web framework, we use "classic" as a sentinel. If there is a function related to the framework, we add "_ssr".

### DIFF
--- a/src/frameworks/index.ts
+++ b/src/frameworks/index.ts
@@ -248,7 +248,7 @@ export async function prepareFrameworks(
   context: any,
   options: any,
   emulators: EmulatorInfo[] = []
-) {
+): Promise<string[]> {
   // `firebase-frameworks` requires Node >= 16. We must check for this to avoid horrible errors.
   const nodeVersion = process.version;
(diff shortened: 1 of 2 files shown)


src/frameworks/index.ts
```diff
@@ -403,6 +411,7 @@ export async function prepareFrameworks(
       config.public = relative(projectRoot, hostingDist);
       if (wantsBackend) codegenFunctionsDirectory = codegenProdModeFunctionsDirectory;
     }
+    deployedFrameworks.push(`${framework}${codegenFunctionsDirectory ? "_ssr" : ""}`);
```
bkendall, line 414: What is `codegenFunctionsDirectory` typically, here? This _looks_ like it could potentially contain user data which makes me ask the question.

inlined: It has the customer’s directory structure, but we’re not actually using it anywhere. We’re just testing whether it is undefined to decide whether to put the “_ssr” suffix on the framework

bkendall (approved):

PR #5141, opened 2022-10-18
Next.js public directory is optional, ignore if missing

jamesdaniels:
Next.js public directory is optional, ignore if missing. This is failing if immediately calling serve/deploy after init, as the hello-world template does not have a public directory.


jamesdaniels:
closing in favor of #5142

PR #5143, opened 2022-10-18
Add warning for emulated database functions

colerogers:
- Adds warning if the function region != us-central1


taeold (approved):
Let's a changelog entry. Seems like this might surprise/break some user.

TheIronDev (approved):

Issue #5134, opened 2022-10-16
Functions emulator timeout running parallel requests directly after start-up


gewfy:
This is very similar to #4807  but occurs when the first requests to the emulator are in parallel

### [REQUIRED] Environment info

**firebase-tools:** 11.14.1

**Platform:** macOS

### [REQUIRED] Test case

Bare minimum http end-point
```javascript
import {onRequest} from 'firebase-functions/v2/https';

export const foo = onRequest({}, (req, res) => {
  res.send('hello');
});
```

### [REQUIRED] Steps to reproduce

1. Start firebase emulator
2. After emulator is up and running, run two curl commands in parallel - do not make any other requests before these two:
```
curl http://localhost:5001/some-project/us-central1/foo & curl http://localhost:5001/some-project/us-central1/foo
```

### [REQUIRED] Expected behavior

Both curl commands should return `hello`

### [REQUIRED] Actual behavior

Response from server is HTTP 500 with body:
```
{"errno":-2,"code":"ENOENT","syscall":"connect","address":"/var/folders/4s/nwqmrp5n4bv0cbt0knyzym2h0000gp/T/fire_emu_17fd97dff0e6f47b.sock"}
```

`firebase-debug.log` reads:
```
[debug] [2022-10-16T20:34:14.536Z] [work-queue] {"queueLength":1,"workRunningCount":0}
[debug] [2022-10-16T20:34:14.536Z] [work-queue] {"queueLength":0,"workRunningCount":1}
[debug] [2022-10-16T20:34:14.536Z] Accepted request GET /demo-29k-cupcake/us-central1/foo --> us-central1-foo
[debug] [2022-10-16T20:34:14.538Z] [functions] Runtime ready! Sending request! {"metadata":{"emulator":{"name":"functions"},"message":"[functions] Runtime ready! Sending request!"}}
[debug] [2022-10-16T20:34:14.538Z] [functions] Got req.url=/demo-29k-cupcake/us-central1/foo, mapping to path=/ {"metadata":{"emulator":{"name":"functions"},"message":"[functions] Got req.url=/demo-29k-cupcake/us-central1/foo, mapping to path=/"}}
[info] i  functions: Loaded environment variables from .env. 
[debug] [2022-10-16T20:34:14.548Z] [worker-pool] addWorker(us-central1-foo) {"metadata":{"emulator":{"name":"functions"},"message":"[worker-pool] addWorker(us-central1-foo)"}}
[...]

PR #5149, opened 2022-10-20
Fix parallel requests in the functions emulator

colerogers:
Addresses #5134 and adds an e2e test for parallel requests

### DIFF
--- a/scripts/integration-helpers/framework.ts
+++ b/scripts/integration-helpers/framework.ts
@@ -45,6 +45,7 @@ const AUTH_BLOCKING_CREATE_V2_LOG =
   "========== AUTH BLOCKING CREATE V2 FUNCTION METADATA ==========";
 const AUTH_BLOCKING_SIGN_IN_V2_LOG =
   "========== AUTH BLOCKING SIGN IN V2 FUNCTION METADATA ==========";
+const HTTPS_V2_FUNCTION_LOG = "========== HTTPS V2 FUNCTION ==========";
 
 interface ConnectionInfo {
   host: string;
@@ -143,6 +144,7 @@ export class TriggerEndToEndTest extends EmulatorEndToEndTest {
   authBlockingCreateV2TriggerCount = 0;
   authBlockingSignInV2TriggerCount = 0;
   rtdbV2TriggerCount = 0;
+  httpsV2TriggerCount = 0;
 
   rtdbFromFirestore = false;
   firestoreFromRtdb = false;
@@ -179,6 +181,7 @@ export class TriggerEndToEndTest extends EmulatorEndToEndTest {
     this.authBlockingCreateV2TriggerCount = 0;
     this.authBlockingSignInV2TriggerCount = 0;
     this.rtdbV2TriggerCount = 0;
+    this.httpsV2TriggerCount = 0;
   }
 
   /*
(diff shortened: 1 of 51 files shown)


scripts/integration-helpers/framework.ts
```diff
@@ -336,6 +342,10 @@ export class TriggerEndToEndTest extends EmulatorEndToEndTest {
     });
   }
 
+  triggerHttpsFunction(): Promise<Response> {
+    return this.invokeHttpFunction("triggerHttpsFunction");
+  }
+
```
taeold, line 348: hmm I'm not sure I understand this indirection

Why can't we call `invokeHttpFunction` directly in `test.ts` file, similar to how its done here:

https://github.com/firebase/firebase-tools/blob/master/scripts/triggers-end-to-end-tests/tests.ts#L415

Instead of checking to see if a log was emitted, we can rely on the HTTP response to indicate that the request succeeded.

scripts/triggers-end-to-end-tests/triggers/index.js
```diff
@@ -43,6 +55,18 @@ const app = initializeApp(
...
+  const url = `http://localhost:${[FUNCTIONS_PORT, FIREBASE_PROJECT, FUNCTIONS_REGION, name].join(
+    "/"
+  )}`;
+  return fetch(url);
+}
+
+exports.triggerHttpsFunction = functions.https.onRequest(async (req, res) => {
+  await invokeHttpsFunction("httpsv2reaction");
+  res.json({ triggered: true });
+});
```
taeold, line 68: Similar to comment above - instead of having a http function which invokes another http functoin, it looks like we can cut this middle man and just have test script invoke the http function directly.

Was there problem with that approach?

scripts/triggers-end-to-end-tests/v2/index.js
```diff
@@ -23,12 +23,18 @@ const AUTH_BLOCKING_CREATE_V2_LOG =
...
 const RTDB_LOG = "========== RTDB V2 FUNCTION ==========";
+const HTTPS_FUNCTION_LOG = "========== HTTPS V2 FUNCTION ==========";
 
 const PUBSUB_TOPIC = "test-topic";
 const START_DOCUMENT_NAME = "test/start";
 
 admin.initializeApp();
 
+exports.httpsv2reaction = functionsV2.https.onRequest((req, res) => {
+  console.log(HTTPS_FUNCTION_LOG);
```
taeold, line 34: this seems unnecessary for http function since the test script can check the status of the http request directly (assuming that we don't need http function calling another http function wrapping)

src/test/emulators/functionsRuntimeWorker.spec.ts
```diff
@@ -99,16 +111,18 @@ describe("FunctionsRuntimeWorker", () => {
       const worker = new RuntimeWorker(workerPool.getKey("trigger"), new MockRuntimeInstance());
       const counter = new WorkerStateCounter(worker);
 
+      worker.readyForWork();
```
taeold, line 114: (sorry for adding busy work) can we update the title to include the new `created` state?

taeold (approved):
some minor comment on tests

taeold (approved):
Can we add a changelog entry too?

src/test/emulators/functionsRuntimeWorker.spec.ts
colerogers: updated

scripts/triggers-end-to-end-tests/v2/index.js
colerogers: fixed

scripts/integration-helpers/framework.ts
colerogers: Much cleaner, thanks!

scripts/triggers-end-to-end-tests/triggers/index.js
colerogers: No problem, but I wanted to invoke the test functions in a similar way for consistency (and I lazily didn't see the direct invoking test at the bottom).

PR #5160, opened 2022-10-24
Added support for publishing Extension versions from remote repo

apascal07:
Added support to the CLI to publish Extension versions from remote repo (i.e. verifiable versions).


src/extensions/extensionsHelper.ts
```diff
@@ -506,52 +531,284 @@ export async function publishExtensionVersionFromLocalSource(args: {
...
+        choices: stageOptions,
+        default: defaultStage,
+      });
+    } else {
+      stage = defaultStage;
+    }
+  }
+
+  // Fetch and validate Extension from remote repo.
+  const archiveUri = path.join(repoUri, `archive/${sourceRef}.zip`);
```
joehan, line 694: path.join is only appropriate for local paths, not for URIs. This will strip the second / in `https://`, because it interprets it as an empty segment in a local path

src/commands/ext-dev-publish.ts
```diff
@@ -21,7 +25,16 @@ marked.setOptions({
  */
 export const command = new Command("ext:dev:publish <extensionRef>")
   .description(`publish a new version of an extension`)
-  .option(`-s, --stage <stage>`, `release stage (supports "rc", "alpha", "beta", and "stable")`)
+  .option(`-s, --stage <stage>`, `release stage (supports "alpha", "beta", "rc", and "stable")`)
+  .option(
+    `--repo <repo>`,
+    `Public Git repo URI (only required for first version, cannot be changed)`
+  )
+  .option(`--ref <ref>`, `commit hash, branch, or tag to build from the repo (defaults to HEAD)`)
```
joehan, line 33: Seems like it doesn't actually default to HEAD atm - if it is omitted, the code here will treat it like a local source.

src/extensions/extensionsApi.ts
```diff
@@ -606,8 +606,10 @@ export async function undeprecateExtensionVersion(extensionRef: string): Promise
  */
 export async function publishExtensionVersion(
   extensionVersionRef: string,
-  packageUri: string,
-  extensionRoot?: string
+  packageUri?: string,
```
joehan, line 609: The param list for this is getting pretty long, and it would be easy to mix up the order of arguments when calling it. Consider  instead:
```
publishExtensionVersion(args: {
  extensionVersionRef: string,
  packageUri?: string,
  extensionRoot?: string,
  repoUri?: string,
  sourceRef?: string
})
```
If you want to get really fancy, you might be able to use a disjoint union type to make it a type error to call this with both packageUri and extensionRoot at the same time

src/extensions/extensionsHelper.ts
```diff
@@ -85,7 +91,12 @@ export const AUTOPOULATED_PARAM_PLACEHOLDERS = {
   DATABASE_INSTANCE: "project-id-default-rtdb",
   DATABASE_URL: "https://project-id-default-rtdb.firebaseio.com",
 };
-export type ReleaseStage = "stable" | "alpha" | "beta" | "rc";
+export const resourceTypeToNiceName: Record<string, string> = {
```
joehan, line 94: Exported consts usually are CAPS_CASE (https://google.github.io/styleguide/tsguide.html#identifiers-constants)

src/extensions/extensionsHelper.ts
```diff
@@ -85,7 +91,12 @@ export const AUTOPOULATED_PARAM_PLACEHOLDERS = {
...
+  "firebaseextensions.v1beta.function": "Cloud Function",
+};
+export type ReleaseStage = "alpha" | "beta" | "rc" | "stable";
+const repoRegex = new RegExp("^https://github.com/[^/]+/[^/]+$");
```
joehan, line 98: I think the `.`  in `github.com` needs to be escaped, otherwise it matches any single character

src/extensions/extensionsHelper.ts
```diff
@@ -369,6 +380,26 @@ export async function promptForValidInstanceId(instanceId: string): Promise<stri
...
+export async function promptForValidRepoURI(): Promise<string> {
+  let repoIsValid = false;
+  let extensionRoot = "";
+  while (!repoIsValid) {
+    extensionRoot = await promptOnce({
+      type: "input",
+      message: "Enter the repo URI where this Extension's source code is located:",
+    });
+    if (!repoRegex.test(extensionRoot)) {
+      logger.info("Repo URI follows this format: https://github.com/<user>/<repo>");
```
joehan, line 395:
```suggestion
      logger.info("Repo URI must follow this format: https://github.com/<user>/<repo>");
```

src/extensions/extensionsHelper.ts
```diff
@@ -369,6 +380,26 @@ export async function promptForValidInstanceId(instanceId: string): Promise<stri
...
+/**
+ * Prompts for a valid repo URI.
+ */
+export async function promptForValidRepoURI(): Promise<string> {
+  let repoIsValid = false;
+  let extensionRoot = "";
+  while (!repoIsValid) {
+    extensionRoot = await promptOnce({
+      type: "input",
+      message: "Enter the repo URI where this Extension's source code is located:",
```
joehan, line 392:
```suggestion
      message: "Enter the GitHub repo URI where this Extension's source code is located:",
```

src/extensions/extensionsHelper.ts
```diff
@@ -503,52 +531,284 @@ export async function publishExtensionVersionFromLocalSource(args: {
...
+  let extension: Extension | undefined;
+  try {
+    extension = await getExtension(extensionRef);
+  } catch (err: any) {
+    // Silently fail and continue the publish flow if Extension not found.
+  }
+
+  // Prompt for repo URI and validate that it hasn't changed if previously set.
+  if (args.repoUri && !repoRegex.test(args.repoUri)) {
+    throw new FirebaseError("Repo URI follows this format: https://github.com/<user>/<repo>");
```
joehan, line 598:
```suggestion
    throw new FirebaseError("Repo URI must follow this format: https://github.com/<user>/<repo>");
```

src/extensions/extensionsHelper.ts
```diff
@@ -503,52 +531,284 @@ export async function publishExtensionVersionFromLocalSource(args: {
...
+  }
+  let repoUri = args.repoUri || extension?.repoUri;
+  if (!repoUri) {
```
joehan, line 601: This looks like it will prompt even if noninteractive=true. It should error out if repoUri is undefined and noninteractive is true

src/extensions/extensionsHelper.ts
```diff
@@ -503,52 +531,284 @@ export async function publishExtensionVersionFromLocalSource(args: {
...
+    repoUri = await promptForValidRepoURI();
+  }
+  if (extension?.repoUri) {
```
joehan, line 604: Optional: Consider unnesting this if statement, its a little hard to follow as is. Something like:
```
if (extension?.repoUri && extension.repoUri !== repoUri) {
  throw ...
}

if (!extension?.repoUri) {
  logger.info("You are about to associate...")
  const confirmed = await confirm(...)
  if (!confirmed) return
}
```

src/extensions/extensionsHelper.ts
```diff
@@ -503,52 +531,284 @@ export async function publishExtensionVersionFromLocalSource(args: {
...
+    });
+    if (!confirmed) {
+      return;
+    }
+  }
+
+  // Prompt for Extension root and default to last root used.
+  let extensionRoot = args.extensionRoot;
+  let defaultRoot = "/";
+  if (!extensionRoot) {
```
joehan, line 640: This logic should to be cleaned up - currently, it doesn't default to the previous versions root when noninteractive = true. I think it can be unnested a bit as well.

src/extensions/extensionsHelper.ts
```diff
@@ -503,52 +531,284 @@ export async function publishExtensionVersionFromLocalSource(args: {
...
+        choices: stageOptions,
+        default: defaultStage,
+      });
+    } else {
+      stage = defaultStage;
+    }
+  }
+
+  // Fetch and validate Extension from remote repo.
+  const archiveUri = `${repoUri}/archive/${sourceRef}.zip`;
```
joehan, line 694: If this can be slow, consider adding a logger.info message saying something like "Downloading and validating source code..."

src/extensions/extensionsHelper.ts
```diff
@@ -503,52 +531,284 @@ export async function publishExtensionVersionFromLocalSource(args: {
...
+      extensionRoot,
+      repoUri,
+      args.sourceRef
+    );
+    publishSpinner.succeed(` Successfully published ${clc.bold(extensionRef)}`);
+  } catch (err: any) {
+    publishSpinner.fail();
+    if (err.status === 404) {
+      throw new FirebaseError(
+        marked(
```
joehan, line 755: This string is long and repeated in publishFromLocalSource - separate it out into a constant?

joehan (changes requested):

src/commands/ext-dev-publish.ts
```diff
@@ -21,7 +25,16 @@ marked.setOptions({
  */
 export const command = new Command("ext:dev:publish <extensionRef>")
   .description(`publish a new version of an extension`)
-  .option(`-s, --stage <stage>`, `release stage (supports "rc", "alpha", "beta", and "stable")`)
+  .option(`-s, --stage <stage>`, `release stage (supports "alpha", "beta", "rc", and "stable")`)
+  .option(
+    `--repo <repo>`,
+    `Public Git repo URI (only required for first version, cannot be changed)`
```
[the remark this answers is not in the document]
apascal07: Yes, it's currently supported as is. Changed the wording to "first version from repo".

src/commands/ext-dev-publish.ts
apascal07: This is working as intended. We will need to wait for a major CLI version to change the default behavior. For now, --ref has to be specified for repo publishing even though there's a prompt for it.

src/extensions/extensionsHelper.ts
apascal07: This isn't part of my change, the order got flipped when merging I think. I can change it if you'd like.

src/extensions/extensionsHelper.ts
apascal07: Good catch!

src/extensions/extensionsHelper.ts
apascal07: I had exactly that originally but Will thought the wording was too aggressive as an error so I changed it to "follows".

src/extensions/extensionsHelper.ts
apascal07: Good idea, thanks!

src/extensions/extensionsHelper.ts
apascal07: Fixed the non-interactive case but can't see any way to unnest unless we do the HTTP call even if an argument has already been passed in to override it.

src/commands/ext-dev-publish.ts
joehan: I'm not sure this is WAI in all cases. Right now, this says it defaults to HEAD. However, if i run `firebase ext:dev:publish --repo=github.com/my/repo --root=myroot`,
it will attempt to publish the local directory, even though I passed in multiple flags to publish a remote source.

src/extensions/extensionsHelper.ts
joehan: Ooops, was reviewing on autopilot. Feel free to, but certainly not blocking

src/extensions/extensionsHelper.ts
joehan:
```suggestion
const repoRegex = new RegExp("^https://github\.com/[^/]+/[^/]+$");
```

joehan (changes requested):

src/commands/ext-dev-publish.ts
apascal07: Ah I see, you're right, good catch. I misunderstood what you meant at first. Changed it to throw an error if `ref` isn't specified, since we still want to default the behavior to local source until a major version.

src/commands/ext-dev-publish.ts
```diff
@@ -45,15 +58,33 @@ export const command = new Command("ext:dev:publish <extensionRef>")
...
-    const res = await publishExtensionVersionFromLocalSource({
-      publisherId,
-      extensionId,
-      rootDirectory: extensionYamlDirectory,
-      nonInteractive: options.nonInteractive,
-      force: options.force,
-      stage: options.stage ?? "stable",
-    });
+    if ((options.repo || options.root) && !options.ref) {
+      throw new FirebaseError("`ref` must be set if `repo` or `root` are set.");
```
joehan, line 62: Should this error or just set ref to HEAD?

apascal07: Good point, done. Changed this condition to default to new path if any of the arguments are set, and it'll default to HEAD later when it prompts.

joehan (approved):

joehan:
Could you add a CHANGELOG entry for this?

PR #5164, opened 2022-10-24
Fixing handling of automatic rewrites detection to consider backends that don't exist but are being deployed.

akongara-goog:
### Description

Fixing handling of automatic rewrites detection to consider backends that don't exist but are being deployed (in the current deploy).

### Scenarios Tested

- Ran hosting-rewrites integration test.
- Manually deployed functions with and without explicit invoker.

### DIFF
--- a/src/deploy/hosting/convertConfig.ts
+++ b/src/deploy/hosting/convertConfig.ts
@@ -14,7 +14,7 @@ import { logger } from "../../logger";
 
 /**
  * extractPattern contains the logic for extracting exactly one glob/regexp
- * from a Hosting rewrite/redirect/header specification
+ * from a Hosting rewrite/redirect/header specification.
  */
 function extractPattern(type: string, source: HostingSource): api.HasPattern {
   let glob: string | undefined;
@@ -42,7 +42,10 @@ function extractPattern(type: string, source: HostingSource): api.HasPattern {
 }
 
 /**
- * Finds an endpoint suitable for deploy at a site given an id and optional region
+ * Finds an endpoint suitable for deploy at a site given an id and optional region.
+ *
+ *  @returns a pair of matching endpoint, if found, and a boolean indicating whether
+ *  the function exists in some region.
  */
 export function findEndpointForRewrite(
   site: string,
@@ -146,26 +149,20 @@ export async function convertConfig(
       let endpoint: backend.Endpoint | undefined = undefined;
       let fnIsPresent = false;
       for (const backend of wantBackends) {
-        const possibleEndpoint = findEndpointForRewrite(deploy.config.site, backend, id, region);
-        if (possibleEndpoint) {
-          [endpoint, fnIsPresent] = possibleEndpoint;
-        }
+        [endpoint, fnIsPresent] = findEndpointForRewrite(deploy.config.site, backend, id, region);
       }
-      console.log(fnIsPresent);
-      let fnIsPresentInExisting = false;
       if (!endpoint) {
-        [endpoint, fnIsPresentInExisting] = findEndpointForRewrite(
+        const [existingEndpoint, fnIsPresentInExisting] = findEndpointForRewrite(
           deploy.config.site,
           haveBackend,
           id,
           region
         );
-      }
-      console.log(fnIsPresentInExisting);
-      if (!endpoint) {
-        console.log("no endpoint");
-        if (fnIsPresent || fnIsPresentInExisting) {
-          console.log("no endpoint");
+        [endpoint, fnIsPresent] = [existingEndpoint, fnIsPresent || fnIsPresentInExisting];
+
+        if (fnIsPresent) {
+          // If the function is a firebase-managed function, we consider it an error for
+          // a rewrite not to point to a valid endpoint.
           throw new FirebaseError(
             `Unable to find a valid endpoint for function. Functions matching the rewrite
   are present but in the wrong region.`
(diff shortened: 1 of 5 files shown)


src/deploy/hosting/convertConfig.ts
```diff
@@ -49,17 +49,16 @@ export function findEndpointForRewrite(
   targetBackend: backend.Backend,
   id: string,
   region: string | undefined
-): backend.Endpoint | undefined {
+): [backend.Endpoint | undefined, boolean] {
```
bkendall, line 52: I'd maybe prefer returning an object rather than a tuple like this. Object can be way better documented (and generally follow style guide for T/Js much better).

Also, on that note, I'm not sure what the bool is for. Can you add some docs for this?

akongara-goog: Done. The bool is for "we saw this function somewhere", which means it's a firebase function and we complain if things don't look right.

src/deploy/hosting/convertConfig.ts
```diff
@@ -96,9 +104,16 @@ export async function convertConfig(
   // rewrites to see if it's necessary.
   const hasBackends = !!deploy.config.rewrites?.some((r) => "function" in r || "run" in r);
 
-  // We need to be able to do a rewrite to an existing function that is may not
-  // even be part of Firebase's control or a function that we're currently
-  // deploying.
+  // We need to be able to do a rewrite to an existing function that may not be
+  // under Firebase's control or a function that we're currently deploying.
+
+  let wantBackends: backend.Backend[] = [];
```
inlined, line 110: You don't need to remember each backend by codebase. Consider merging them all together with

```
const wantBackend = backend.merge(
  ...Object.values(functionsPayload.functions || {}).map(c => c.wantBackend)
)
```

Then you can get rid of for loops later.

src/deploy/hosting/convertConfig.ts
```diff
@@ -136,8 +151,43 @@ export async function convertConfig(
...
+      let functionIsInDeployingBackends = false;
+      for (const backend of wantBackends) {
+        const deployingEndpointSearch = findEndpointForRewrite(
+          deploy.config.site,
+          backend,
+          id,
+          region
+        );
+        functionIsInDeployingBackends =
+          functionIsInDeployingBackends || deployingEndpointSearch.endpointExistsSomewhere;
```
inlined, line 164: I feel like "endpointExistsSomewhere" isn't super clear because the endpoint may not have existed somewhere (it could have been in the wrong region). I wonder if "foundMatchingId" or something similar might convey the technical meaning more crisply.

src/deploy/hosting/convertConfig.ts
```diff
@@ -136,8 +151,43 @@ export async function convertConfig(
       }
       const id = rewrite.function.functionId;
       const region = rewrite.function.region;
-      const endpoint = findEndpointForRewrite(deploy.config.site, haveBackend, id, region);
+      let endpoint: backend.Endpoint | undefined = undefined;
+      let functionIsInDeployingBackends = false;
```
inlined, line 155: again, this seems like a place where backend.merge could have helped. What if above you merged not only the wantBackend but the exitingBackend as well?

The main reason I would argue against simplifying would be to set you up for more advanced error handling where you notice that a function currently exists but is about to get deleted.
1. I don't know if that would be necessary
2. You could still use some utility methods/comprehensions to do that. Off the top of my head:

```
wantBackend = backend.merge(...Object.keys(payload.functions || {}).map(c => c.wantBackend));
haveBackendTargets = backend.merge(...Object.keys(payload.functions || {}).map(c => c.haveBackend));
// Just putting this in a backend since your utility methods use them already
// You can use any predicate, but I'm going to rely on the structure of Backend to avoid N^2 performance
toDelete = backend.matchingBackend(haveBackendTargets, (e) => wantBackend.endpoints[e.region]?.[e.id]);
```

src/test/deploy/hosting/convertConfig.spec.ts
```diff
@@ -59,39 +61,152 @@ describe("convertConfig", () => {
...
+      want: {
+        rewrites: [
+          {
+            glob: "/foo",
+            function: FUNCTION_ID,
+            functionRegion: "us-central2",
+          },
+        ],
+      },
+      functionsPayload: {
```
inlined, line 73: Ooof. Sorry you had to re-add these test fixtures after my rush/foolishness.

inlined (approved):

src/deploy/hosting/convertConfig.ts
akongara-goog: Done.

src/deploy/hosting/convertConfig.ts
akongara-goog: Agreed. Renamed.

src/deploy/hosting/convertConfig.ts
akongara-goog: I added a test case for function-being-deleted and fixed it. Turned out to be reasonably easy to do. I'm leaving haveBackend and wantBackend separate because it's much simpler to understand that way, especially for the function-being-deleted case.

src/test/deploy/hosting/convertConfig.spec.ts
akongara-goog: Np, happens

src/deploy/hosting/convertConfig.ts
akongara-goog: Done.

PR #5166, opened 2022-10-24
Use template for emulator update details

tohhsinpei:
To prevent silly oversights like #5151.


src/emulator/downloadableEmulators.ts
```diff
@@ -25,53 +25,66 @@ const EMULATOR_INSTANCE_KILL_TIMEOUT = 4000; /* ms */
 const CACHE_DIR =
   process.env.FIREBASE_EMULATORS_PATH || path.join(os.homedir(), ".cache", "firebase", "emulators");
 
+const EMULATOR_VERSIONS = {
```
bkendall, line 28: I might recommend up-leveling all the values that have to change into a single object, like the checksums and sizes. Otherwise, it might be even easier to change _this_ value but not go down below and adjust the others.

tohhsinpei: Done.

bkendall (approved):
Reasonable improvement :)

PR #5168, opened 2022-10-25
Update the flag name in the warning message for singleProjectMode to match the actual flag

christhompsongoogle:
Update the flag name in the warning message for singleProjectMode to match the actual flag

### DIFF
--- a/src/emulator/auth/server.ts
+++ b/src/emulator/auth/server.ts
@@ -374,7 +374,7 @@ export async function createApp(
         `Multiple projectIds are not recommended in single project mode. ` +
         `Requested project ID ${projectId}, but the emulator is configured for ` +
         `${defaultProjectId}. This warning will become an error in the future. To opt-out of ` +
-        `single project mode add/set the \'"single_project_mode"\' false' property in the` +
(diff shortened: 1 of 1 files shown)


src/emulator/auth/server.ts
```diff
@@ -373,9 +373,8 @@ export async function createApp(
       const errorString =
         `Multiple projectIds are not recommended in single project mode. ` +
         `Requested project ID ${projectId}, but the emulator is configured for ` +
-        `${defaultProjectId}. This warning will become an error in the future. To opt-out of ` +
-        `single project mode add/set the \'"singleProjectMode"\' false' property in the` +
-        ` firebase.json emulators config.`;
+        `${defaultProjectId}. To opt-out of single project mode add/set the ` +
+        `\'"singleProjectMode"\' false' property in the firebase.json emulators config.`;
```
bkendall, line 377: I don't see that this message is actually changing the flag name. Is that correct?

bkendall (approved):

src/emulator/auth/server.ts
christhompsongoogle: you're right - that's odd. The repo flag is single_project_mode though I checked.

PR #5176, opened 2022-10-26
Inlined.web frameworks label

inlined:
Add a label to hosting versions with the type of web framework being used. This uses the same id as the google analytics metric, so I moved that track code to the deploy function as well.

Separately, I moved HostingResolved to src/hosting so that I don't feel guilty adding new fields for the sake of communication between web frameworks and hosting.

### DIFF
--- a/src/deploy/hosting/prepare.ts
+++ b/src/deploy/hosting/prepare.ts
@@ -6,6 +6,7 @@ import { Context } from "./context";
 import { Options } from "../../options";
 import { HostingOptions } from "../../hosting/options";
 import { zipIn } from "../../functional";
+import { track } from "../../track";
 
 /**
  *  Prepare creates versions for each Hosting site to be deployed.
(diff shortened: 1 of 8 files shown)


src/deploy/hosting/prepare.ts
```diff
@@ -25,12 +26,24 @@ export async function prepare(context: Context, options: HostingOptions & Option
     return Promise.resolve();
   }
 
-  const version: Omit<api.Version, api.VERSION_OUTPUT_FIELDS> = {
-    status: "CREATED",
-    labels: deploymentTool.labels(),
-  };
   const versions = await Promise.all(
-    configs.map((config) => api.createVersion(config.site, version))
+    configs.map(async (config) => {
```
TheIronDev, line 30: Please add tests for this change

TheIronDev (approved):

src/deploy/hosting/prepare.ts
inlined: Done

TheIronDev (approved):

Issue #5180, opened 2022-10-26
TypeScript Firebase Cloud Functions don't recognize ES modules (can't `import` npm packages)


tdkehoe:
### [REQUIRED] Environment info

firebase-admin@11.2.0
firebase-tools@11.15.0
firebase@9.12.1
node: 'v18.11.0'
npm: '8.19.2'

**firebase-tools:**

macOS Big Sur 11.7

**Platform:**

### [REQUIRED] Test case

This is the default  `package.json` when you spin up Firebase Cloud Functions.

```
{
  "compilerOptions": {
    "module": "commonjs",
    "noImplicitReturns": true,
    "noUnusedLocals": true,
    "outDir": "lib",
    "sourceMap": true,
    "strict": true,
    "target": "es2017"
  },
  "compileOnSave": true,
  "include": [
    "src"
  ]
}
```

### [REQUIRED] Steps to reproduce

npm install -g @angular/cli
ng new CloudFunctions
cd CloudFunctions
npm install -g firebase-tools
npm audit fix --force
firebase init

Select `Emulators`.
Select `Create a new project`.
Select `Functions Emulator` and `Firestore Emulator`.

firebase login
firebase init functions

Select `TypeScript`.

firebase emulators:start

Error: `functions/lib/index.js does not exist, can't deploy Cloud Functions`

In `package.json` change `"main": "lib/index.js",` to `"main": "src/index.ts",`.

Error: `SyntaxError: Cannot use import statement outside a module`

This is the error I can't fix in TypeScript.

First, that error message should say `SyntaxError: Cannot use import statement outside an ES module`. As written the error message suggests that you could use `import` in a CommonJS module, which is debatable.

I can fix the error by switching to JavaScript. This suggests that the problem is in `tsconfig.json`, which JavaScript doesn't use.

Change `index.ts` to `index.js`.
In `package.json` change `"main": "lib/index.js",` to `"main": "src/index.js",`.
Add `"type": "module",`.

Success! To test that the JavaScript Cloud Functions are working with ES modules, add this cloud function to `index.js`:

```js
export const makeUppercase = functions.firestore.document('messages/{docId}').onCreate((snap, context) => {
    const original = snap.data().original;
    functions.logger.log('Uppercasing', context.params.docId, original);
    const uppercase = original.toUpperCase();
    return snap.ref.set({ uppercase }, { merge: true });
});
```

[...]

PR #5183, opened 2022-10-26
Update typescript/package.nolint.json

tdkehoe:
### Description

Added `"type": "module",` for ES modules. This doesn't make ES modules work, please see  https://github.com/firebase/firebase-tools/issues/5180

Changed `"main": "lib/index.js",` to `"main": "src/index.ts",` to match directory structure and file extension.

Updated dependency versions of `firebase-admin`, `firebase-functions`, and `typescript`.

These changes are identical to my pull request for `typescript/package.lint.json`.

### DIFF
--- a/templates/init/functions/typescript/package.nolint.json
+++ b/templates/init/functions/typescript/package.nolint.json
@@ -1,5 +1,6 @@
 {
   "name": "functions",
+  "type": "module",
   "scripts": {
     "build": "tsc",
     "build:watch": "tsc --watch",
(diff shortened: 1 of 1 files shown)


inlined:
We will update the dependencies in the project initialization, but depending on `js` is correct, not `ts`. TypeScript is a layer that sits above vanilla `node` and must be compiled into Node for the function to work. That's why we create the functions.json with:

```
"predeploy": [
  "npm --prefix \"$RESOURCE_DIR\" run build"
]
```

tdkehoe:
When you spin up `firebase init functions` and select TypeScript it creates the directory structure `functions/src/index.ts`. `main` in `functions/package.json` has to match the directory structure.

PR #5195, opened 2022-10-31
Paths that have middleware (Next.js) should route through Cloud Functions

jamesdaniels:
Skip moving files that might match middleware to the code generated hosting directory. We don't want these on the CDN, they should passed through to the backing Cloud Function.

* Skip copying files to hosting that match middleware
* Next export succeeds even when there's middleware, it's no longer a usable shortcut
* If there is middleware, we need a Cloud Function
* Give up entirely on trying to skip Cloud Functions for app directory (fallbacks aren't showing in manifest)
* ISR is a reason to spin up a Cloud Function
* Console logging reasons why a Cloud Function is needed

Tidying things up:

* Bumping the next.js devDep
* Adding types where we didn't have any before
* Cleaning up the public directory generation function
* Reduce verbosity by using readJSON
* Cleanup the logic for wantsBackend

TODO:

- [x] ~~Pass the real port to Next dev server for `firebase serve`~~
- [ ] Ship the port/host fix to the Cloud Functions dep https://github.com/FirebaseExtended/firebase-framework-tools/pull/50


jamesdaniels:
https://github.com/FirebaseExtended/firebase-framework-tools/issues/42

src/frameworks/next/index.ts
```diff
@@ -300,14 +331,17 @@ export async function ɵcodegenFunctionsDirectory(sourceDir: string, destDir: st
...
   const { default: next } = relativeRequire(dir, "next");
   const nextApp = next({
     dev: true,
     dir,
+    hostname: hostingEmulatorInfo?.host,
+    port: hostingEmulatorInfo?.port,
   });
   const handler = nextApp.getRequestHandler();
   await nextApp.prepare();
+  // TODO can we check for middleware and error if we don't have hostingEmulatorInfo
```
jamesdaniels, line 344: I wasn't able to find a good way to hook into `firebase serve` and get the port number. IMO we're steering away from serve anyway, so let's punt. Taking a TODO to see if we can error but we shouldn't hold this up

jamesdaniels: The manifest files (in dev mode) don't have middleware, we could potentially just look for a middleware.js/ts file & warn

src/frameworks/next/index.ts
```diff
@@ -183,79 +223,70 @@ export async function init(setup: any) {
...
+
+    const isReactServerComponent = route.dataRoute.endsWith(".rsc");
+    const contentDist = join(
+      sourceDir,
+      distDir,
+      "server",
+      isReactServerComponent ? "app" : "pages"
+    );
+
+    const parts = path.split("/").filter((it) => !!it);
```
jamesdaniels, line 279: FYI we need filter, since `["/"].split("/") => ["", ""]`. It was redundant with the slice

src/frameworks/next/index.ts
```diff
@@ -183,79 +223,70 @@ export async function init(setup: any) {
...
+    const partsOrIndex = parts.length > 0 ? parts : ["index"];
+
+    const htmlPath = `${join(...partsOrIndex)}.html`;
+    await mkdir(join(destDir, dirname(htmlPath)), { recursive: true });
+    await copyFile(join(contentDist, htmlPath), join(destDir, htmlPath));
```
jamesdaniels, line 284: IMO if we're not able to find files we should error, since that's likely a misunderstanding with the directory structure and we want developers to file issues

jamesdaniels: Especially now that we aren't leaning on next export

src/frameworks/next/index.ts
```diff
@@ -157,10 +179,28 @@ export async function build(dir: string): Promise<BuildResult> {
...
+      );
+    }
+    console.log("");
+  }
+
   return { wantsBackend, headers, redirects, rewrites };
 }
 
-/**
+/** q
```
jamesdaniels, line 203: Stray character.

TheIronDev: Remove it?

TheIronDev:
Please add unit tests to capture what is changing in `src/frameworks`

src/frameworks/next/index.ts
jamesdaniels: Yep, will do, was just reviewing myself

jamesdaniels:
> Please add unit tests to capture what is changing in `src/frameworks`

Any thoughts on unit testing strategy here? These changes should be captured in `scripts/webframeworks-deploy-tests` still functioning & deploying the same number of files (despite a new route)—admittedly it's a functional test. 

I am planning on breaking out some unit tests but I think that needs to be part of a larger refactor.

jamesdaniels:
Handing off to monogram

Issue #1288, opened 2019-05-14
Firestore emulator sending incorrect event payloads to Functions emulator


samtstern:
I just ran a real Cloud Function (non-Firebase) triggered on `onWrite` for the document `foo/bar`.  

I created the document (no previous data) and logged out the two arguments to the function and got this:

**context**
```js
{
    eventId: "6e7a9aae-5e22-4414-abaf-cb6b0a3363a2-0",
    eventType: "providers/cloud.firestore/eventTypes/document.write",
    notSupported: {},
    resource: "projects/fir-dumpster/databases/(default)/documents/foo/bar",
    timestamp: "2019-05-14T01:49:26.587519Z",
}
```

**data**
```js
{
    oldValue: {},
    updateMask: {},
    value: {
      createTime: "2019-05-14T01:49:26.587519Z",
      fields: {
        hey: {
          stringValue: "now",
        },
      },
      name: "projects/fir-dumpster/databases/(default)/documents/foo/bar",
      updateTime: "2019-05-14T01:49:26.587519Z",
    },
}
```

However if I do the same thing in the emulator suite I get:

**context**
```js
{
  "eventId": "68ded781-de59-4adb-aff2-a8b6597a9725",
  "timestamp": "2019-05-14T01:59:17.900Z",
  "eventType": "providers/cloud.firestore/eventTypes/document.write",
  "resource": {
    "name": "projects/fir-dumpster/databases/(default)/documents/foo/bar",
    "service": "firestore.googleapis.com"
  }
}
```

**data** 
```js
{
  "value": {
    "name": "projects/fir-dumpster/databases/(default)/documents/foo/bar",
    "fields": {
      "hello": {
        "stringValue": "world"
      }
    },
    "createTime": "2019-05-14T01:59:17.900Z",
    "updateTime": "2019-05-14T01:59:17.900Z"
  },
  "updateMask": {}
}
```

The difference is that in the emulator we get `context.resource` is an Object and in prod we get `context.resource` is a String.

samtstern:
cc @abeisgoat

samtstern:
Note that I am working around this in https://github.com/firebase/firebase-tools/pull/1260

samtstern:
I think there is something really funky going on here ... I can't find a single solution for `resource` assumptions that works for both RTDB and Firestore.

samtstern:
I checked for RTDB in prod and I got:

**context**
```
[...]

PR #5196, opened 2022-10-31
Auto-upgrade/downgrade implicit concurrency with CPU changes

inlined:
Potential fix for #1288 if we decide to take it. We can decide to:

1. Auto downgrade from 80 concurrents to 1 concurrent, but not auto-upgrade (may confuse users, but avoids race condition bugs)
2. Apply the 1/80 concurrent defaults whenever concurrency is not specified (most in-line with other fields)
3. Keep the current behavior that concurrency is only set implicitly on first deploy

### DIFF
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -213,7 +213,7 @@ export async function prepare(
   for (const [codebase, { wantBackend, haveBackend }] of Object.entries(payload.functions)) {
     inferDetailsFromExisting(wantBackend, haveBackend, codebaseUsesEnvs.includes(codebase));
     await ensureTriggerRegions(wantBackend);
-    resolveCpu(wantBackend);
+    resolveCpuAndConcurrency(wantBackend);
     validate.endpointsAreValid(wantBackend);
     inferBlockingDetails(wantBackend);
   }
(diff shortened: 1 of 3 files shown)


TheIronDev (approved):

inlined:
Updated to also upgrade concurrency when it is not specified and CPU grows to support it.

src/deploy/functions/prepare.ts
```diff
@@ -321,17 +321,18 @@ export function inferDetailsFromExisting(
...
       wantE.cpu = haveE.cpu;
     }
 
+    // N.B. concurrency has different defaults based on CPU. If the customer
+    // only specifies CPU and they change that specification to < 1, we should
+    // turn off concurrency. We'll not do the opposite and turn on concurrency
+    // if there are >1 CPU because this could expose race conditions. They
+    // either need to start on concurrency where they've always needed to
+    // handle race conditions, or they should explicitly enable.
+    // We'll hanndle this in setCpuAndConcurrency
```
taeold, line 334: This comment seems inaccurate w/ latest implementation. We do upgrade concurrency regardless, ignoring race condition.

taeold (approved):

Issue #4266, opened 2022-03-07
"Invalid source token" code 9 error during function deployment


trex-quo:
### [REQUIRED] Environment info

**firebase-tools:** 10.1.4 

We are not able to upgrade to the most recent version, as it breaks the emulator suite. I saw that this issue was supposed to be fixed in the most recent release, however the emulators still fail to start up with the most recent version. Our functions config values are still not being loaded in properly when we start the emulators, so we are using 10.1.4 until further notice.

**Platform:** Docker image cimg/node:16.14.0 via CircleCI

### [REQUIRED] Test case

Here is error output from our CircleCI build:
```
i  functions: updating Node.js 16 function functionName1(us-central1)...
i  functions: updating Node.js 16 function functionName2(us-central1)...
i  functions: updating Node.js 16 function functionName3(us-central1)...

Functions deploy had errors with the following functions:
	functionName1(us-central1)
	functionName2(us-central1)
	functionName3(us-central1)
i  functions: cleaning up build files...
Error: There was an error deploying functions:
- Error Failed to update function functionName1 in region us-central1
- Error Failed to update function functionName2 in region us-central1
- Error Failed to update function functionName3 in region us-central1

Exited with code exit status 123
CircleCI received exit code 123
```

When looking at the function logs, I have seen a variety of vague errors that I cannot find the cause of. Here is one of these errors:
```
{
  "protoPayload": {
    "@type": "type.googleapis.com/google.cloud.audit.AuditLog",
    "status": {
      "code": 14,
      "message": "The service has encountered an error during container import. Please try again later"
    },
    "authenticationInfo": {
      "principalEmail": "REDACTED"
    },
    "serviceName": "cloudfunctions.googleapis.com",
    "methodName": "google.cloud.functions.v1.CloudFunctionsService.UpdateFunction",
    "resourceName": "REDACTED"
  },
  ...
  "severity": "ERROR",
  ...
}
```

And here is another:
```
{
  "protoPayload": {
    "@type": "type.googleapis.com/google.cloud.audit.AuditLog",
    "status": {
      "code": 9,
      "message": "Invalid source token"
    },
    "authenticationInfo": {
      "principalEmail": "REDACTED"
    },
    "serviceName": "cloudfunctions.googleapis.com",
    "methodName": "google.cloud.functions.v1.CloudFunctionsService.UpdateFunction",
    "resourceName": "REDACTED"
  },
  "severity": "ERROR",
  ...
}
```

[...]

PR #5198, opened 2022-10-31
Fixes source token expiration by token refresh

blidd-google:
Developers have reported a variety of issues while attempting large function deployments (see https://github.com/firebase/firebase-tools/issues/4266). Specifically, developers were encountering an `Invalid source token` error message after successfully deploying a subset of the functions, indicating that at least part of the issue stemmed from the source token re-use hack (used to expedite deployments by re-using the same container for each function).

On particularly large deployments (~100+), the process can take >30 minutes. The source token acquired after the first function deployment expires at 30 minutes, resulting in the failure of subsequent function deployments that included the expired source token.

This change tracks the expiration of acquired source tokens, and after about 25 minutes, discards the old token and instructs the next function deployment to request a new source token. The remaining functions wait until the new source token is acquired before starting deployment. A 5 minute buffer is provided to basically guarantee that no subsequent function deployments will include the expired token. After about 25 minutes, the second token also expires and the next function requests a new token, etc., etc.


src/deploy/functions/release/sourceTokenScraper.ts
```diff
@@ -1,5 +1,10 @@
 import { logger } from "../../../logger";
-import { Timer } from "./timer";
+
+enum TokenState {
+  "NONE",
+  "VALID",
+  "INVALID",
+}
```
taeold, line 7: tangent: enums are controversial in TS for some reason.

https://blog.logrocket.com/why-typescript-enums-suck/

Internally, I think we prefer string literals:

```
type TokenState = "none" | "valid" | "invalid"
```

src/deploy/functions/release/sourceTokenScraper.ts
```diff
@@ -8,40 +13,54 @@ import { Timer } from "./timer";
...
+
+  checkTokenExpired(): boolean {
+    if (this.expiration === undefined) {
+      throw new Error();
+    }
+    return process.hrtime.bigint() >= this.expiration;
+  }
+
+  calculateTokenExpiry(): bigint {
+    const now = process.hrtime.bigint();
```
taeold, line 51: Personally, I think using `hrtime` with its `bigint` silliness is unnecessarily precise, especially if the time range we care about is in MINUTES not MICROSECONDS.

I'd suggest we stick to `Date.now()` and forget about big ints.

src/deploy/functions/release/sourceTokenScraper.ts
```diff
@@ -8,40 +13,54 @@ import { Timer } from "./timer";
...
-  tokenPromise(): Promise<string | undefined> {
-    if (this.tokenRefreshRequired) {
-      this.tokenRefreshRequired = false;
-      this.tokenRefreshTimer = new Timer(this.tokenValidDurationMs);
-      return Promise.resolve(undefined);
+  async getToken(): Promise<string | undefined> {
+    if (this.tokenState === TokenState.NONE) {
+      this.tokenState = TokenState.INVALID;
+      return undefined;
+    } else if (this.tokenState === TokenState.INVALID) {
```
taeold, line 31: I found that token states as described here bit confusing. `tokenState == INVALID` makes me think that token is invalid, but token doesn't actually exist at this point.

What about another name that closely maps to the state of the scraper, like `tokenFetchState` instead of `tokenState` and `FETCHING` and `INVALID`?

src/deploy/functions/release/sourceTokenScraper.ts
```diff
@@ -8,40 +13,54 @@ import { Timer } from "./timer";
...
+        this.tokenState = TokenState.INVALID;
+        this.promise = new Promise((resolve) => (this.resolve = resolve));
+        return undefined;
+      }
+      return this.promise;
     }
-    return this.promise;
+  }
+
+  checkTokenExpired(): boolean {
```
taeold, line 43: nit* I usually name my boolean function with "is" like `isTokenExpired`

src/deploy/functions/release/sourceTokenScraper.ts
```diff
@@ -8,40 +13,54 @@ import { Timer } from "./timer";
...
+    if (this.expiration === undefined) {
+      throw new Error();
```
taeold, line 45: This error is user visible - we should try our best to describe the situation and also call out what users should do when they see the issue.

src/deploy/functions/release/fabricator.ts
```diff
@@ -210,9 +210,11 @@ export class Fabricator {
     if (apiFunction.httpsTrigger) {
       apiFunction.httpsTrigger.securityLevel = "SECURE_ALWAYS";
     }
-    apiFunction.sourceToken = await scraper.tokenPromise();
+    // apiFunction.sourceToken = await scraper.tokenPromise();
```
taeold, line 213: nit - stray comment?

src/deploy/functions/release/sourceTokenScraper.ts
```diff
@@ -1,37 +1,64 @@
...
-  // Token Promise will return undefined for the first caller
-  // (because we presume it's this function's source token we'll scrape)
-  // and then returns the promise generated from the first function's onCall
-  tokenPromise(): Promise<string | undefined> {
-    if (this.firstCall) {
-      this.firstCall = false;
-      return Promise.resolve(undefined);
+  isTokenExpired(): boolean {
+    if (this.expiry === undefined) {
+      throw new FirebaseError("failed to check expiry: no token exists");
```
taeold, line 42: This error message makes sense to you and me but won't to almost all users trying to deploy their function.

I think this is a good example of a user facing error message - describe the situation in coarsely and suggest next steps:

https://github.com/firebase/firebase-tools/blob/2ebefe2d8c5e33e9fc8c078235d02579ccdbf036/src/deploy/functions/prepareFunctionsUpload.ts#L36-L40

In our case, hitting this if statement should never happen and would be considered a bug. Maybe we can suggest that they file a issue.

src/deploy/functions/release/sourceTokenScraper.ts
```diff
@@ -1,37 +1,64 @@
...
     }
-    return this.promise;
+    return Date.now() >= this.expiry;
   }
 
   get poller() {
     return (op: any) => {
+      console.log(`${op.response.name}: polling`)
```
taeold: nit* let's remove console logs

src/deploy/functions/release/sourceTokenScraper.ts
```diff
@@ -1,37 +1,64 @@
...
       if (op.metadata?.sourceToken || op.done) {
         const [, , , /* projects*/ /* project*/ /* regions*/ region] =
           op.metadata?.target?.split("/") || [];
         logger.debug(`Got source token ${op.metadata?.sourceToken} for region ${region as string}`);
+        console.log(
```
taeold: let's remove console logs.

src/deploy/functions/release/timer.ts
```diff
@@ -1,14 +1,30 @@
 /** Measures the time taken from construction to the call to stop() */
 export class Timer {
-  private readonly start: bigint;
+  private readonly start: bigint; // nanoseconds
```
taeold: are changes to this file still needed? If not, let's remove them.

src/deploy/functions/release/sourceTokenScraper.ts
blidd-google: ah gotcha. working on a more descriptive message.

src/deploy/functions/release/sourceTokenScraper.ts
blidd-google: 'throw new FirebaseError(
        "Your deployment is checking the expiration of a source token that has not yet been polled. " +
          "Hitting this case should never happen and should be considered a bug. " +
          "Please file an issue at https://github.com/firebase/firebase-tools/issues"
      );'

src/deploy/functions/release/sourceTokenScraper.ts
blidd-google:
```
throw new FirebaseError(
        "Your deployment is checking the expiration of a source token that has not yet been polled. " +
          "Hitting this case should never happen and should be considered a bug. " +
          "Please file an issue at https://github.com/firebase/firebase-tools/issues"
      );
```
What about something like this?

src/deploy/functions/release/sourceTokenScraper.ts
```diff
@@ -1,28 +1,52 @@
...
+      return undefined;
+    } else if (this.fetchState === "FETCHING") {
+      return this.promise; // wait until we get a source token
+    } else if (this.fetchState === "VALID") {
+      if (this.isTokenExpired()) {
+        this.fetchState = "FETCHING";
+        this.promise = new Promise((resolve) => (this.resolve = resolve));
+        return undefined;
+      }
+      return this.promise;
```
taeold, line 36: Should we throw an error here instead of silently succeeding? Can see argument for either.

src/test/deploy/functions/release/sourceTokenScraper.spec.ts
```diff
@@ -31,14 +31,63 @@ describe("SourcTokenScraper", () => {
...
+    scraper.poller({
+      metadata: {
+        sourceToken: "magic token #2",
+        target: "projects/p/locations/l/functions/f",
+      },
+    });
+    await expect(scraper.getToken()).to.eventually.equal("magic token #2");
+  });
+
+  it("concurrent requests for source token", async () => {
```
taeold, line 69: Nice test!!

taeold (approved):
Lgtm

src/deploy/functions/release/sourceTokenScraper.ts
blidd-google: By throwing an error, do you mean throwing an error and then catching it in an enclosing `try { await getToken(); ... } catch (expiredErr) {...}` block, and then retrying?

taeold: I mean that in theory we should never reach this codepath - fetchState can only be in one of three states, and we have if condition for all of them. So we could throw (which could happen if we misunderstood this code for somee reason) or just return the promise (which might leave the CLI in an undefined state). I have slight preference for the former.

src/deploy/functions/release/sourceTokenScraper.ts
```diff
@@ -1,28 +1,55 @@
...
-  // Token Promise will return undefined for the first caller
-  // (because we presume it's this function's source token we'll scrape)
-  // and then returns the promise generated from the first function's onCall
-  tokenPromise(): Promise<string | undefined> {
-    if (this.firstCall) {
-      this.firstCall = false;
-      return Promise.resolve(undefined);
+  isTokenExpired(): boolean {
+    if (this.expiry === undefined) {
+      throw new FirebaseError(
```
inlined, line 45: Can we update this to throw with code: 2? We track this in Google Analytics for internal assertion errors.

src/test/deploy/functions/release/sourceTokenScraper.spec.ts
```diff
@@ -31,14 +31,63 @@ describe("SourcTokenScraper", () => {
...
+      metadata: {
+        sourceToken: "magic token",
+        target: "projects/p/locations/l/functions/f",
+      },
+    });
+    await expect(scraper.getToken()).to.eventually.equal("magic token");
+    const timeout = (duration: number): Promise<void> => {
+      return new Promise<void>((resolve) => setTimeout(resolve, duration));
+    };
+    await timeout(50);
```
inlined, line 58: In the future, try to instrument code without sleeping in your test. You can inject a clock for example.

PR #5201, opened 2022-11-01
Add missing host and port to Next.js custom server

leoortizz:
### Description

- Pass host and port to Next.js custom server from hosting emulator info. 
- Fallback to `DEFAULT_HOST` and `DEFAULT_PORTS.hosting` since the `serve` command is not passing the emulator info to `prepareFrameworks`.
- Add a TODO to do pass emulator info to `prepareFrameworks` from `serve`.
- Add `hostingEmulatorInfo` parameter to the `Framework`'s `getDevModeHandle` type

### Scenarios Tested

A redirect in a Next.js Middleware should redirect to the appropriate route now, before it was erroring because the host and port were `undefined`.


src/frameworks/next/index.ts
```diff
@@ -303,17 +303,17 @@ export async function ɵcodegenFunctionsDirectory(sourceDir: string, destDir: st
  * Create a dev server.
  */
 export async function getDevModeHandle(
-  dir: Parameters<NonNullable<Framework['getDevModeHandle']>>[0],
-  hostingEmulatorInfo: Parameters<NonNullable<Framework['getDevModeHandle']>>[1]
+  dir: Parameters<NonNullable<Framework["getDevModeHandle"]>>[0],
```
jamesdaniels, line 306: I think we can simplify the typings here a little bit, take on a TODO to namespace it

src/serve/index.ts
```diff
@@ -24,10 +24,10 @@ export async function serve(options: any): Promise<void> {
   options.port = parseInt(options.port, 10);
   if (targetNames.includes("hosting") && config.extract(options).some((it: any) => it.source)) {
     experiments.assertEnabled("webframeworks", "emulate a web framework");
-    
+
     // TODO: pass emulator info as the fourth parameter so that the dev server
     //       can use host and port
-    await prepareFrameworks(targetNames, options, options); 
+    await prepareFrameworks(targetNames, options, options);
```
jamesdaniels, line 30: Could we pass something in the shape of an emulatorInfo here for now? e.g, `[{name: 'hosting', port: options.port, ... }]`

jamesdaniels (changes requested):
I started a PR to add hostname and port to the custom server in prod, mind taking a peek? We can cut a canary build to test.

https://github.com/FirebaseExtended/firebase-framework-tools/pull/50

jamesdaniels:
Also let's add a line to the changelog

src/frameworks/next/index.ts
leoortizz: Indeed. I replaced it with the explicit types required and added a TODO to reuse it from the `Frameworks` interface

src/serve/index.ts
leoortizz: I harcoded it and changed the TODO to get the "real emulator info"

PR #5205, opened 2022-11-03
Update pubsub version/md5/byte size.

christhompsongoogle:
Update pubsub version/md5/byte size.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+Updated the pubsub emulator to v0.7.1.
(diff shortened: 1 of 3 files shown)


yuchenshi (approved):
Changlog please

christhompsongoogle:
Hi [fixl](https://github.com/fixl), this is a repackage of the cloud pubsub emulator v0.7.1. We don't add any changes on top of it to incorporate it in the emulator suite. In that case I expect the issue to still be outstanding.

Issue #5008, opened 2022-09-23
Debug mode in 11.10.0 broke for me (re #4886)


spicemix:
### [REQUIRED] Environment info

**firebase-tools:**
11.10.0

**Platform:**

MacOS latest intel

### [REQUIRED] Test case

That may be more work for me than fixing this is for you, this release is only a few hours old, giving you early heads up.

### [REQUIRED] Steps to reproduce

I use `firebase emulators:start --inspect-functions 9229` rather than just `firebase emulators:start` to enter debug mode. This works perfectly for my typescript functions on 11.9.0 but the functions just hang on 11.10.0.

### [REQUIRED] Expected behavior

However, if I do just use `firebase emulators:start` in 11.10.0, everything works fine. I just can't debug this way.

### [REQUIRED] Actual behavior

Well, I think it's a bit of hassle to run logs and clean them up for public consumption, but I think you'll appreciate the feedback that the fix in #4886 broke for me in practice in debug mode only (which you mention is different now). Being able to debug is handy so I will roll back to 11.9.0 until you have this fixed thanks!

spicemix:
✅ Fixed for me in 11.11.0 thank you so very much

spicemix:
Ooooops my global update workflow `npm-check -ug` that had worked for years just broke somehow on me in the most sly way...it reported installation, with no error, but in fact didn't update my global deps. So I had tested under 11.9.0 instead of 11.11.0, without knowing. The problem still exists for me in 11.12.0. I never actually tested 11.11.0. So I have to reopen this thank you.

tomas-carv-com:
I have similar issue, every time I run emulators with `--inspect-functions` I cant connect to them or they dont function correctly. Triggers does not work etc.

tomas-carv-com:
And I see these 2 PRs which are affecting --inspect-functions

https://github.com/firebase/firebase-tools/commit/454be140a47502a2aad128d8e8053e278c295213
https://github.com/firebase/firebase-tools/commit/b1099a2c3b0a104124eda5d8bf4b48e3aac38c37

I mean now debugging is completely broken. I cant use app if debugging is on

taeold:
We think we have a fix and are working on a patch.

If anyone has time to help - would love to have folks test out the dl-fn-emu-inspect-bug branch which contains our proposed fix on their workflow.

Steps for running a local version of the Firebase CLI is available at https://github.com/firebase/firebase-tools/blob/master/CONTRIBUTING.md#getting-started.

Issue #5050, opened 2022-09-30
Version 11.10.0 no longer calls onFinalize background function


johnnyoshika:
### [REQUIRED] Environment info

**firebase-tools:** 11.10.0 (also tested 11.12.0)

**Platform:** Windows

### [REQUIRED] Test case

```
export const onFileFinalize = functions.storage
  .object()
  .onFinalize(async object => {
  
  console.log('Hello');
});
```

Then upload using the Web SDK's `uploadBytesResumable()`.

### [REQUIRED] Steps to reproduce

Start the emulator, upload file with `uploadBytesResumable()`, then observe that `onFinalize()` background function doesn't get triggered.

### [REQUIRED] Expected behavior

Expect `onFinalize()` background function to get triggered. This worked properly in firebase-tools@11.9.0, so this is a regression that was introduced in version 11.10.0.

### [REQUIRED] Actual behavior

Background function doesn't get triggered.

bkendall:
I've seen something similar recently. There should be a log for the storage emulator - could you provide the output of that?

johnnyoshika:
@bkendall I can't find the storage emulator log. Can you point me in the right direction? I do see `firebase-debug.log`  though.

Here's what I see printed in `firebase-debug.log` when I upload a file using firebase-tools@11.10.0:

```
[debug] [2022-09-30T21:24:06.695Z] Sep. 30, 2022 2:24:06 P.M. io.gapi.emulators.netty.HttpVersionRoutingHandler channelRead
INFO: Detected non-HTTP/2 connection.
 {"metadata":{"emulator":{"name":"firestore"},"message":"Sep. 30, 2022 2:24:06 P.M. io.gapi.emulators.netty.HttpVersionRoutingHandler channelRead\r\nINFO: Detected non-HTTP/2 connection.\r\n"}}
[debug] [2022-09-30T21:24:07.997Z] >>> [apiv2][query] POST http://localhost:5001/functions/projects/demo-project/trigger_multicast [none]
[...]

PR #5211, opened 2022-11-04
Fix bug where event triggered functions failed in debug mode

taeold:
A classic case of deadlock!

Internally, the Functions Emulator maintains a work queue which processes unit of work called "tasks" with configured parallelism.

When Functions Emulator runs in debug mode, the work queue runs in `SEQUENTIAL` mode where tasks are processed one by one in FIFO order.

When event functions are triggered via the multicast route (e.g. storage and auth triggers), Functions Emulator submits a task per function associated with the event to the work queue where each task makes a call to the event function via its HTTP endpoint. Let's call this task an "invocation task".

When the Function Emulator receives an HTTP request for the function invocation, it submits a new task to the work queue to handle the http request. Let's call this task a "http task".

So, a call to the multicast route begets zero or more "invocation task" (one per trigger). Each "invocation task" begets exactly one "http task".

The code today is written such that an "invocation task" will wait for the corresponding "http task" to complete.

In debug mode this causes the invocation task to wait forever - the "http task" is stuck behind its "invocation task" in the queue. It never has chance to be processed.

This PR proposes that "invocation task" doesn't wait for the corresponding "http task" to complete, effectively breaking the deadlock.

The result is that "invocation task" now fire-and-forget an "http task". AFAIK, this doesn't have any negative effect because "invocation task" didn't do anything with the result of the "http task" anyway.

Fixes https://github.com/firebase/firebase-tools/issues/5008, https://github.com/firebase/firebase-tools/issues/5050

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,2 @@
 - Add support for object list using certain Admin SDKs (#5208)
+- Fix bug where emulated event triggered function broke in debug mode (#5211)
--- a/scripts/triggers-end-to-end-tests/tests.inspect.ts
+++ b/scripts/triggers-end-to-end-tests/tests.inspect.ts
@@ -10,6 +10,7 @@ const FIREBASE_PROJECT = process.env.FBTOOLS_TARGET_PROJECT || "";
  * parallel emulator subprocesses.
  */
 const TEST_SETUP_TIMEOUT = 80000;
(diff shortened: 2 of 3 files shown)


TheIronDev (approved):
Lgtm!!

colerogers (approved):
sweet fix! 👏

PR #5212, opened 2022-11-04
Handle Next.js rewrites/redirects/headers

leoortizz:
### Description
- Handle Next.js rewrites/redirects/headers incompatible with `firebase.json` in Cloud Functions
- Filter out Next.js prerendered routes that matches rewrites/redirects/headers rules from SSG content directory


src/frameworks/next/index.ts
```diff
@@ -149,11 +150,25 @@ export async function build(dir: string): Promise<BuildResult> {
...
+      if (
+        has ||
+        destination.includes("http") // filter out rewrites to third parties
+      )
+        return undefined;
+
+      try {
+        new RE2(source);
+        new RE2(destination);
+
```
jamesdaniels, line 166: Set `wantsBackend` to true

src/frameworks/next/index.ts
```diff
@@ -149,11 +150,25 @@ export async function build(dir: string): Promise<BuildResult> {
...
+        return { source, destination };
+      } catch (error) {
+        // regex not supported by RE2 cannot be transformed into firebase.json rewrites
+        return undefined;
```
jamesdaniels, line 170: Set `wantsBackend` to true

src/frameworks/next/index.ts
```diff
@@ -149,11 +150,25 @@ export async function build(dir: string): Promise<BuildResult> {
...
-      if (has) return undefined;
-      return { source, destination };
+      if (
+        has ||
+        destination.includes("http") // filter out rewrites to third parties
+      )
+        return undefined;
+
+      try {
+        new RE2(source);
```
jamesdaniels, line 164: let's do similar tests for redirects and headers

src/frameworks/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+  } catch (error) {
+    return false;
+  }
+}
+
+/**
+ * Wheter the given URL is an external URL.
+ */
+export function isThirdPartyUrl(url: string): boolean {
+  return url.includes("http");
```
jamesdaniels, line 21: I'd do `.startsWith` and change it to just isUrl

src/frameworks/next/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+ * According to the Next.js documentation:
+ * ```md
+ *  To match a regex path you can wrap the regex in parentheses
+ *  after a parameter, for example /post/:slug(\\d{1,}) will match /post/123
+ *  but not /post/abc.
+ * ```
+ * See: https://nextjs.org/docs/api-reference/next.config.js/redirects#regex-path-matching
+ */
+export function pathHasRegex(path: string): boolean {
+  for (let i = 0; i < path.length; i++) {
```
jamesdaniels, line 12: This would probably be more readable, tunable with a regex, mind converting?

src/frameworks/next/index.ts
```diff
@@ -177,18 +187,18 @@ export async function build(dir: string): Promise<BuildResult> {
...
       if (has) return false;
 
-      if (
-        supportsFrameworkRegex(source) &&
-        supportsFrameworkRegex(destination) &&
-        isUrl(destination) === false
-      ) {
-        return true;
-      } else {
+      if (pathHasRegex(source) || isUrl(destination)) {
```
jamesdaniels, line 190: Maybe we could break these into util functions? `filter(rewriteSupportedByFirebase)` that would make them easier to test & reuse in codegenPublicDir too.

would you mind taking on the work to utilize in codegenPublicDir in this PR? the idea being if the regex matches the path (for !supportedByFirebase) we should not copy that file into the CDN

jamesdaniels (changes requested):

jamesdaniels:
Add changelog

src/frameworks/next/utils.ts
leoortizz: @TheIronDev in the last commit I pushed a working regex for that.. I think it can get tricky with `.includes` because of paths like this:
```
/post/\\(escapedparentheses\\)/:slug(\\d{1,}) <--- this has a regex and escaped parenthesis
```

src/frameworks/next/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  return false;
+}
+
+/**
+ * Remove double backslashes from a string
+ */
+export function cleanEscapedChars(path: string): string {
+  return path.replace(/\\/g, "");
```
jamesdaniels, line 29: Shouldn't this replace with a single "\", rather than empty? mind getting some more edge cases covered in the test?

src/frameworks/next/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+ * ```md
+ *  To match a regex path you can wrap the regex in parentheses
+ *  after a parameter, for example /post/:slug(\\d{1,}) will match /post/123
+ *  but not /post/abc.
+ * ```
+ * See: https://nextjs.org/docs/api-reference/next.config.js/redirects#regex-path-matching
+ */
+export function pathHasRegex(path: string): boolean {
+  // finds parentheses that are not preceded by double backslashes
+  const regex = /(?<!\\)\(/;
```
jamesdaniels, line 16: I don't think it would hurt readability to make this a one liner, e.g, `return /(?<!\\)\(/.test(path)`

src/frameworks/utils.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Whether the given string is a URL.
+ */
+export function isUrl(url: string): boolean {
+  return url.startsWith("http");
```
jamesdaniels, line 5: we should make sure this is a protocol, /^https?:\/\//, so say `httpendpoint/foo/bar` wouldn't trip it

jamesdaniels (changes requested):

src/frameworks/utils.ts
leoortizz: @jamesdaniels 🤦‍♂️ right, thanks! I also added the example you mentioned as test cases

src/frameworks/next/utils.ts
leoortizz: @jamesdaniels right.. that should be the case. Reviewing it now Next.js has this on their docs:

> The following characters (, ), {, }, :, *, +, ? are used for regex path matching, so when used in the source as non-special values they must be escaped by adding

I'm fixing it and adding test cases, will push

src/frameworks/next/interfaces.ts
```diff
@@ -0,0 +1,31 @@
+import type { Header, Rewrite, Redirect } from "next/dist/lib/load-custom-routes";
+
```
leoortizz, line 2: Unfortunately `Redirect` is not extendable because its members are not statically known, so this is not possible:
```
export interface RoutesManifestRedirect extends Redirect {
  regex: string;
}
```

src/frameworks/next/index.ts
```diff
@@ -136,26 +127,71 @@ export async function build(dir: string): Promise<BuildResult> {
...
+  const manifest = JSON.parse(manifestBuffer.toString()) as Manifest;
   const {
     headers: nextJsHeaders = [],
     redirects: nextJsRedirects = [],
     rewrites: nextJsRewrites = [],
   } = manifest;
-  const headers = nextJsHeaders.map(({ source, headers }) => ({ source, headers }));
+
+  const headers = nextJsHeaders
+    .filter((header) => {
```
TheIronDev, line 138: Nit: Please remove side-effects from the filter function.

Maybe instead:

```
const hasUnsupportedHeader = nextJsHeaders.some((header) => {
  return !isHeaderSupportedByFirebase(header);
});
if (hasUnsupportedHeader) {
 wantsBackend = true;
}

const headers = nextJsHeaders
  .filter((header) => isHeaderSupportedByFirebase(header))
  .map(({ source, headers }) => ({
      // clean up unnecessary escaping
      source: cleanEscapedChars(source),
      headers,
    }));
```

src/frameworks/next/index.ts
```diff
@@ -136,26 +127,71 @@ export async function build(dir: string): Promise<BuildResult> {
...
-  const nextJsRewritesToUse = Array.isArray(nextJsRewrites)
-    ? nextJsRewrites
-    : nextJsRewrites.beforeFiles || [];
+    .filter((redirect) => {
+      if (redirect.internal) return false;
+
+      if (isRedirectSupportedByFirebase(redirect)) {
+        return true;
+      } else {
+        wantsBackend = true;
```
TheIronDev, line 159: Same here

src/frameworks/next/index.ts
```diff
@@ -136,26 +127,71 @@ export async function build(dir: string): Promise<BuildResult> {
...
   const rewrites = nextJsRewritesToUse
-    .map(({ source, destination, has }) => {
-      // Can we change i18n into Firebase settings?
-      if (has) return undefined;
-      return { source, destination };
+    .filter((rewrite) => {
+      if (isRewriteSupportedByFirebase(rewrite)) {
+        return true;
+      } else {
+        wantsBackend = true;
```
TheIronDev, line 186: Ditto here

src/frameworks/next/index.ts
```diff
@@ -210,10 +246,37 @@ export async function ɵcodegenPublicDirectory(sourceDir: string, destDir: strin
...
+        join(
+          sourceDir,
+          distDir,
+          "routes-manifest.json" // TODO: get this from next/constants
+        )
+      ),
+    ]);
+
     const prerenderManifest = JSON.parse(prerenderManifestBuffer.toString());
+    const routesManifest = JSON.parse(routesManifestBuffer.toString()) as Manifest;
```
TheIronDev, line 267: Is it possible `routesManifestBuffer` or `prerenderManifestBuffer` is not JSON.parse-able?

Should we try-catch guard this to prevent errors?

src/frameworks/next/index.ts
```diff
@@ -210,10 +246,37 @@ export async function ɵcodegenPublicDirectory(sourceDir: string, destDir: strin
...
+
+    const rewritesToUse = getNextjsRewritesToUse(routesManifest.rewrites);
+    const rewritesNotSupportedByFirebase = rewritesToUse?.filter(
+      (rewrite) => isRewriteSupportedByFirebase(rewrite) === false
```
TheIronDev, line 271: Nit, these can be simplified:

```
(rewrite) => !isRewriteSupportedByFirebase(rewrite)
```

src/frameworks/next/index.ts
```diff
@@ -210,10 +246,37 @@ export async function ɵcodegenPublicDirectory(sourceDir: string, destDir: strin
...
+    );
+    const redirectsNotSupportedByFirebase = routesManifest.redirects?.filter(
+      (redirect) => isRedirectSupportedByFirebase(redirect) === false
+    );
+    const headersNotSupportedByFirebase = routesManifest.headers?.filter(
```
TheIronDev, line 276: If you guard with an empty array you can remove a bunch of the if-blocks later on.

```
const headersNotSupportedByFirebase = routesManifest.headers?.filter(
      (header) => isHeaderSupportedByFirebase(header) === false
    ) || [];


```

src/frameworks/next/interfaces.ts
TheIronDev: :(

src/frameworks/next/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+export function cleanEscapedChars(path: string): string {
+  return path.replace(/\\/g, "");
+}
+
+/**
+ * Whether a Next.js rewrite is supported by Firebase.
+ */
+export function isRewriteSupportedByFirebase(rewrite: Rewrite): boolean {
+  if ("has" in rewrite || pathHasRegex(rewrite.source) || isUrl(rewrite.destination)) {
```
TheIronDev, line 31: Simplify Nit:

```
return !("has" in rewrite || pathHasRegex(rewrite.source) || isUrl(rewrite.destination));
```

TheIronDev: + others

TheIronDev, line 31: What is the significance of "has" here? Can you add a comment explaining how this is being checked?

src/frameworks/next/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+  return true;
+}
+
+/**
+ * Firebase Rewrites that can be supported by firebase.json.
+ */
+export function getNextjsRewritesToUse(
+  nextJsRewrites: Manifest["rewrites"]
+): RoutesManifestRewrite[] {
+  return Array.isArray(nextJsRewrites) ? nextJsRewrites : nextJsRewrites?.beforeFiles || [];
```
TheIronDev, line 66: I see 3 possible return values packed into 1 line.
Can we break up this ternary to make it easier to read?

```
if (Array.isArray(nextJsRewrites)) {
  return nextJsRewrites;
}

return nextJsRewrites?.beforeFiles || [];
```

Also, are we certain `nextJsRewrites?.beforeFiles` is an array? should we instead check `Array.isArray` on it to be safe?

src/frameworks/next/index.ts
jamesdaniels: probably better to refactor to readJSON & throw rather than get too crazy IMO

src/frameworks/next/index.ts
```diff
@@ -210,10 +246,37 @@ export async function ɵcodegenPublicDirectory(sourceDir: string, destDir: strin
...
+    ]);
+
     const prerenderManifest = JSON.parse(prerenderManifestBuffer.toString());
+    const routesManifest = JSON.parse(routesManifestBuffer.toString()) as Manifest;
+
+    const rewritesToUse = getNextjsRewritesToUse(routesManifest.rewrites);
+    const rewritesNotSupportedByFirebase = rewritesToUse?.filter(
+      (rewrite) => isRewriteSupportedByFirebase(rewrite) === false
+    );
+    const redirectsNotSupportedByFirebase = routesManifest.redirects?.filter(
```
jamesdaniels, line 273: Can we destructure with defaults, `const { redirects=[] } = routesManifest` like we do in build? would make it so this couldn't be null

jamesdaniels: and we wouldn't have to guard

src/frameworks/next/index.ts
```diff
@@ -222,6 +285,29 @@ export async function ɵcodegenPublicDirectory(sourceDir: string, destDir: strin
           continue;
         }
 
+        if (rewritesNotSupportedByFirebase) {
```
jamesdaniels, line 288: If we destructure / guard, then we can simplify this (and the others) as

```ts
if (rewritesNotSupportedByFirebase.find(rewrite => new RegExp(rewrite.regex).test(path))) {
  continue;
}```

**Review by jamesdaniels — CHANGES_REQUESTED** [review]

File: src/frameworks/next/index.ts (hunk 20 above)
Comment (jamesdaniels) [reply src/frameworks/next/index.ts:288]: As an optimization you would pre init all the regex before the `for`, `e.g, const rewriteRegexsNotSupportedByFirebase = rewritesNotSupportedByFirebase.map(rewrite => new RegExp(rewrite.regex))`

File: src/frameworks/next/utils.ts (hunk 18 above)
Comment (leoortizz) [reply src/frameworks/next/utils.ts]: @TheIronDev [Next.js won't build](https://nextjs.org/docs/messages/routes-must-be-array) if this is not an array(or object with arrays). Since we always use this after the build step we should not need to check if it's an array

**Review by jamesdaniels — CHANGES_REQUESTED** [review]

File: src/frameworks/next/index.ts (hunk 20 above)
Comment (leoortizz) [reply src/frameworks/next/index.ts]: @jamesdaniels I changed that to destructure with defaults and also instantiated the regexes before the loop as suggested.

Then instead of
```js
if (rewritesNotSupportedByFirebase.find(rewrite => new RegExp(rewrite.regex).test(path))) {
  continue;
}
```

I assigned the loop result to a variable because I think it's easier/quicker to understand as the variable name replaces a comment that may become outdated if this changes at some point. Also as the name is quite long, IMO a variable name can help. Let me know your thoughts

src/frameworks/next/utils.ts
leoortizz: @TheIronDev `has` is used for [Header, Cookie, and Query Matching](https://nextjs.org/docs/api-reference/next.config.js/rewrites#header-cookie-and-query-matching). As it can't be translated to firebase.json we say it's not supported. I improved the functions docs linking to both the `firebase.json` and Next.js docs

src/frameworks/next/index.ts
```diff
@@ -230,15 +231,15 @@ export async function ɵcodegenPublicDirectory(sourceDir: string, destDir: strin
       }
     }
 
-    const [prerenderManifestBuffer, routesManifestBuffer] = await Promise.all([
-      readFile(
+    const [prerenderManifest, routesManifest] = await Promise.all([
+      readJSON<any>(
```
jamesdaniels, line 235: `<any>` is implied

src/frameworks/next/index.ts
```diff
@@ -136,26 +128,55 @@ export async function build(dir: string): Promise<BuildResult> {
   }
 
   const manifestBuffer = await readFile(join(dir, distDir, "routes-manifest.json"));
-  const manifest: Manifest = JSON.parse(manifestBuffer.toString());
+  const manifest = JSON.parse(manifestBuffer.toString()) as Manifest;
```
jamesdaniels, line 131: I directed incorrectly, `manifest: Manifest` is still preferred over `as Manifest` at Google

jamesdaniels (approved):
Address nits if you have the chance, I'll kick off integration tests

src/frameworks/next/utils.ts
jamesdaniels: @leoortizz `path.replace(/\\([\(\){}\:\+\?])/g, (a,b) => b)` should do it IMO. WDYT?

```
> '/\\(\\)\\{\\}\\:\\+\\?/:slug'.replace(/\\([\(\){}\:\+\?])/g, (a,b) => b)
'/(){}:+?/:slug'
```

src/frameworks/next/index.ts
leoortizz: @jamesdaniels I replaced it with the typed `readJSON`

src/frameworks/next/index.ts
leoortizz: @jamesdaniels `any` was not implied as the function required a type. As it might be necessary sometimes I added `any` as a default

src/frameworks/next/utils.ts
leoortizz: @jamesdaniels this seem to fix it, thanks! I implemented it and adjusted the unit tests, should be all working now

jamesdaniels (approved):

jamesdaniels:
merged in #5316, thanks @leoortizz

PR #5222, opened 2022-11-07
[hosting:github] - Show warning on 404

pschoffer:
### Description

When running `firebase init hosting:github` we use OAuth token to fetch a public key for secrets. If this token doesn't have access to the given repository 403 is returned by GitHub on public repositories. However, on private repositories, 404 is returned instead.

This change adds a catch condition for `404` as well as a corresponding warning message.

### Scenarios Tested

1) Have a private repo in an organization
2) Revoke the grant for that organization
3) Try to setup github action in that repository via `firebase init hosting:github`

*Actual Result*
No warning or any feedback is shown to the user only the prompt is reactivated.
actual

*Fixed Result*
The warning is shown.

fixed


pschoffer:
It looks like this was addressed in another PR - https://github.com/firebase/firebase-tools/pull/5643

(I probably should not have bothered opening the PR several months ago )

Issue #5153, opened 2022-10-21
cannot deploy custom event triggers (onCustomEventPublished)


taishi55:
### [REQUIRED] Environment info

**firebase-tools:** 11.15.0

**Platform:** macOS

### [REQUIRED] Test case

?

### [REQUIRED] Steps to reproduce

1. create a project and install stripe extension with setup
2. initialize firebase function
3. add 
```
import { onCustomEventPublished } from "firebase-functions/v2/eventarc";

export const customevent = onCustomEventPublished(
    {
      eventType: "com.stripe.v1.customer.subscription,
      channel: "locations/us-west1/channels/firebase",
      region: "us-west1",
    },
    (event) => { ... });
```
5. deploy the function
6. you will see an error and fail the deployment

### [REQUIRED] Expected behavior

Successful deployment

### [REQUIRED] Actual behavior
```
Could not create or update Cloud Run service customevent, Container Healthcheck failed. Revision 'customevent-00001-rem' is not ready and cannot serve traffic. The user-provided container failed to start and listen on the port defined provided by the PORT=8080 environment variable. Logs for this revision might contain more information.

```

inlined:
We have additional work to do to create the channel on demand when a function is created. If you deploy the Extension that writes to the channel first and then deploy the function that listens to it, does that work for now?

taishi55:
The problem was using `config()`. Use secret manager instead of using config(). Then you can deploy without a problem

PR #5226, opened 2022-11-08
Creates eventarc channel in before deploying custom event function

blidd-google:
Some customers were experiencing deployment failures while attempting to deploy custom event triggered functions using the `onCustomEventPublished` providers (https://github.com/firebase/firebase-tools/issues/5153).

Our hypothesis is that the customer was attempting to deploy the custom event function _before_ creating the Eventarc channel on which the function would be listening for events, and as a result, GCF threw an error back at the customer.

This PR patches the missing Eventarc channel issue by:
1. detecting which channel the function was expecting to listen on,
2. checking if the channel already exists,
3. if it doesn't exist, creating the new channel via the Eventarc API, and
4. deploying the function.


src/deploy/functions/release/fabricator.ts
```diff
@@ -295,6 +303,36 @@ export class Fabricator {
...
+        .run(async () => {
+          try {
+            const op: { name: string } = await eventarc.createChannel({ name: channel });
+            return await poller.pollOperation<eventarc.Channel>({
+              ...eventarcPollerOptions,
+              pollerName: `create-${channel}-${endpoint.region}-${endpoint.id}`,
+              operationResourceName: op.name,
+            });
+          } catch (err: any) {
+            if (err.status === 409) {
```
inlined, line 325: Can we add a comment that we get a CONFLICT because the channel has been created already and therefore we don't need to worry about the channel already existing?

src/gcp/eventarc.ts
```diff
@@ -0,0 +1,100 @@
...
+    return undefined;
+  }
+  return res.body;
+}
+
+/**
+ * Creates a channel.
+ */
+export async function createChannel(channel: Channel): Promise<Operation> {
+  const body: Partial<Channel> = cloneDeep(channel);
```
inlined, line 69: We needed to clone into a `Partial<Channel>` because we previously deleted the name. Now that we're not doing this we don't need to clone the body or use partial types.

src/test/deploy/functions/release/fabricator.spec.ts
```diff
@@ -481,6 +488,97 @@ describe("Fabricator", () => {
       );
     });
 
+    it.only("handles already existing eventarc channels", async () => {
```
inlined, line 491: What does `it.only` do rather than `it`?

inlined (commented):
I'm sure the channel didn't come with a project ID from the container contract. Do we need to normalize the channel that we get from functions.yaml into a full resource name? If so, where does that happen?

src/test/deploy/functions/release/fabricator.spec.ts
blidd-google: Ah I forgot to take that out, hadn't cleaned up the PR before creating the draft!

If some unit tests are specified with `it.only`, only they will run when the `npm run test` command is executed.

src/deploy/functions/release/fabricator.ts
blidd-google: Re: channel normalization, I believe the normalization happens here: https://github.com/firebase/firebase-tools/blob/master/src/deploy/functions/runtimes/discovery/v1alpha1.ts

blidd-google: and yes agreed about adding a comment

src/gcp/eventarc.ts
```diff
@@ -0,0 +1,95 @@
...
+  /** If set, the channel will grant publish permissions to the 2P provider. */
+  provider?: string;
+
+  // BEGIN oneof transport
+  pubsubTopic?: string;
+  // END oneof transport
+
+  state?: "PENDING" | "ACTIVE" | "INACTIVE";
+
+  /** When the channel is PENDING, this token must be sent to the provider */
```
egilmorez, line 27: Suggest backticks for the literal `PENDING`

src/gcp/eventarc.ts
```diff
@@ -0,0 +1,95 @@
...
+  const res = await client.put<Channel, Channel>(channel.name, channel, {
+    queryParams: {
+      updateMask: fieldMasks(channel).join(","),
+    },
+  });
+  return res.body;
+}
+
+/**
+ * Deletes a chanel.
```
egilmorez, line 91: Typo: "channel"

egilmorez (commented):
Some nits for you, thanks for sending!

src/gcp/eventarc.ts
```diff
@@ -0,0 +1,95 @@
+import { Client } from "../apiv2";
+import { eventarcOrigin } from "../api";
+import { last } from "lodash";
+import { fieldMasks } from "./proto";
+
+export const API_VERSION = "v1";
+
+export interface Channel {
+  /** User provided name. Must be projects/p/locations/l/channels/c */
```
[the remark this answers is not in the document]
egilmorez: Hm, in that case, is this even information the 3P developer needs?  

Is the user-specified name the one they handle in their code?  Do they ever see the after-resolution name at all?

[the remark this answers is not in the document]
blidd-google: I think the 3P developer will only need their user-specified name. As far as I can tell, the full channel name is only visible via the Eventarc console. Do you think the comment is unnecessary?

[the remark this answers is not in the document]
egilmorez: I'd say the detail about the name is only necessary if we expect the user to view it in the Eventarc console.  If not, then it's unnecessary and potentially confusing.

It's still breaking my brain that you can create multiple user-defined names for multiple channels (true?) but they all resolve to one single specific name/path, but I don't always understand everything :)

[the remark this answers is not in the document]
blidd-google: I can see how the comment would cause confusion for 3p developers, so I would be okay with removing it. 

The channel resources are associated with a project and location, so I think the goal was to keep it simple for 3p developers by providing reasonable defaults for project + location. I think the average use case is simple enough that we won't see too much confusion about channel naming conventions (since they don't really have to worry about anything but providing a channel ID), but maybe we will get some feedback about this as the trigger becomes more popular...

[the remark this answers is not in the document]
egilmorez: Oh, OK, I get it now -- p and location (or some or all of that path) are variables.  

Nonetheless, I like the plan of leaving out the detail unless we see demand for it.  Thanks!

egilmorez (approved):
LGTM, thanks!

Issue #4982, opened 2022-09-15
Uploading a .png image with .putData() creates an octet-stream object instead of a png object in storage emulator


Wizzel1:
### [REQUIRED] Environment info

**firebase-tools:** 11.9.0

**Platform:** Windows 10

### [REQUIRED] Steps to reproduce

1. Create an app that has access to the emulator.
2. Convert a png image to bytes
3. Upload the data with `.putData(data, SettableMetadata(contentType: 'image/png'));`
4. Check the result in storage emulator

### [REQUIRED] Expected behavior
As in version 11.7.0, I expected to have a image/png object added to the storage emulator.

### [REQUIRED] Actual behavior
Instead it adds a application/octet-stream.

tonyjhuang:
what sdk are you using?

Wizzel1:
I use the flutter SDK https://github.com/firebase/flutterfire/tree/master/packages/firebase_storage

balazss:
I have a similar issue with `11.8.1` when using `uploadBytesResumable` with Node client. Only happens with PNGs.

Error is:
`error FirebaseError: Firebase Storage: An unknown error occurred, please check the error payload for server response. (storage/unknown)`

sebschre:
I also have a similar issue with any file that is larger than ~200kB:
`FirebaseError: Firebase Storage: An unknown error occurred, please check the error payload for server response. (storage/unknown)`
The type is always `application/octet-stream` and the size is always `262.14 kB`

LanderBeeuwsaert:
We have the same issue when we do upload() via firebase admin to the storage emulator, since updating from 11.9.0 to 11.16.0

rulamon:
I've been having a similar issue as well in the emulator: when adding a file to the emulator through firebase admin's upload() the type is always application/octet-stream.

Yuangwang:
Hi, I think the discussion here has been for two separate issues.

1. Some of you are having issues with resumable uploads erroring out and also having the type `application/octet-stream` (@sebschre, @balazss, and maybe @Wizzel1). I think this was reported before (https://github.com/firebase/firebase-tools/issues/4959) and fixed in this pr (https://github.com/firebase/firebase-tools/pull/5018). Try updating to the latest  firebase-tools release and your issue should hopefully be fixed, if not please report back and we can take a closer look!
2. There is another issue that I was able to reproduce where some admin sdks are not setting the file type properly, its always being set to `application/octet-stream`. The fix for this is a WIP. 

[...]

PR #5229, opened 2022-11-09
Fix storage admin sdk content type

Yuangwang:
### Description

Fixes the content type admin sdk issues mentioned here https://github.com/firebase/firebase-tools/issues/4982

### Scenarios Tested
Tested via nodejs admin sdk and python gcs sdk. Also added new tests for this specific issue

### DIFF
--- a/scripts/storage-emulator-integration/conformance/gcs.endpoints.test.ts
+++ b/scripts/storage-emulator-integration/conformance/gcs.endpoints.test.ts
@@ -312,7 +312,7 @@ describe("GCS endpoint conformance tests", () => {
 
         expect(res.text).to.include("Bad content type.");
       });
-      
+
       it("should upload content type properly from x-upload headers", async () => {
         const returnedMetadata = await supertest(storageHost)
           .post(`/upload/storage/v1/b/${storageBucket}/o?uploadType=multipart`)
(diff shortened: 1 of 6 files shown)


scripts/storage-emulator-integration/conformance/gcs.endpoints.test.ts
```diff
@@ -251,6 +251,26 @@ describe("GCS endpoint conformance tests", () => {
         expect(returnedMetadata.contentType).to.equal(customMetadata.contentType);
         expect(returnedMetadata.contentDisposition).to.equal(customMetadata.contentDisposition);
       });
+
+      it("should upload content type properly from x-upload-content-type headers", async () => {
```
tonyjhuang, line 255: can you add a test for the media upload protocol above as well?

src/emulator/storage/upload.ts
```diff
@@ -44,7 +44,7 @@ export type MediaUploadRequest = {
 export type MultipartUploadRequest = {
   bucketId: string;
   objectId: string;
-  metadataRaw: string;
+  metadataRaw: object;
```
tonyjhuang, line 47: small nit: can you rename this `metadata` if we're passing in the object

tonyjhuang (approved):
two small comments, overall lgtm thanks for the fix!

abhis3 (approved):

PR #5235, opened 2022-11-11
Copy project .npmrc to function dist

cjsewell:
To support private repositories and custom npm configs, copy the projects .npmrc file to the functions dist dir before deploying


src/frameworks/index.ts
```diff
@@ -513,6 +514,12 @@ ${firebaseDefaults ? `__FIREBASE_DEFAULTS__=${JSON.stringify(firebaseDefaults)}\
         // continue
       });
 
+      if (fileExistsSync(getProjectPath(".npmrc"))) {
```
jamesdaniels, line 517: I'd rather use the async APIs so we won't have any hiccups if we want to parallelize this work later, mind changing?

src/frameworks/index.ts
```diff
@@ -513,6 +514,12 @@ ${firebaseDefaults ? `__FIREBASE_DEFAULTS__=${JSON.stringify(firebaseDefaults)}\
...
+        await copyFile(getProjectPath(".npmrc"), join(functionsDist, ".npmrc")).catch(() => {
```
jamesdaniels, line 518: I think we can drop the catch and let the error throw

jamesdaniels: Otherwise looks good. Mind adding a changelog entry?

jamesdaniels (changes requested):

src/frameworks/index.ts
cjsewell: No worries, done!

src/frameworks/index.ts
```diff
@@ -519,6 +519,12 @@ ${firebaseDefaults ? `__FIREBASE_DEFAULTS__=${JSON.stringify(firebaseDefaults)}\
         // continue
       });
 
+      await pathExists(getProjectPath(".npmrc")).then( async (exists) => {
+        if (exists){
```
jamesdaniels, line 523: can simplify to `if (await pathExists(getProjectPath(".npmrc")) { ... }`

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Copy projects .npmrc file if it exists to functions dist directory (#5235)
```
jamesdaniels, line 1: I'd say `Respect .npmrc in backends spun up for web frameworks` or something like that, scope it to frameworks

jamesdaniels (changes requested):
Thanks. Two nits, happy to merge once addressed

cjsewell:
No worries, should be all good now

jamesdaniels (approved):

Issue #4952, opened 2022-09-05
Unexpected Error while deploying using "firebase deploy --only functions" in firebase-tools@11.8.0 (latest version).


rahuld600:
### [REQUIRED] Environment info

**firebase-tools:** 11.8.0 (latest release)

**Platform:** : macOS

### [REQUIRED] Test case

If I make a function in your index.js file, and then I try to run using:
firebase deploy --only functions:nameOfYourFunction,
I get an Unexpected Error at the end.

### [REQUIRED] Steps to reproduce

Using npm or whatever package manager you prefer, run these on your terminal,

> npm install -g firebase-tools
> firebase login
> firebase init functions
> 

Now under the functions folder, make sure you have installed node modules using `npm install`,
and finally you create a basic function you like on index.js after you are done importing stuffs, for example:

> 
> const functions = require('firebase-functions');
> const admin = require('firebase-admin');
> admin.initializeApp();
> 
> exports.yourFunctionName = functions.https.onRequest(async(req,res)=>{
>     functions.logger.log(res.body); //just for an example.
> })
> 

Now when you try to do:

`firebase deploy --only functions:yourFunctionName`

You'll see this error message in the end:

> [debug] [2022-09-05T11:21:08.962Z] Building nodejs source
> [debug] [2022-09-05T11:21:08.963Z] Could not find functions.yaml. Must use http discovery
> [debug] [2022-09-05T11:21:08.972Z] Error: spawn ./node_modules/.bin/firebase-functions ENOENT
>     at Process.ChildProcess._handle.onexit (node:internal/child_process:283:19)
>     at onErrorNT (node:internal/child_process:476:16)
>     at processTicksAndRejections (node:internal/process/task_queues:83:21)
> [error] 
> [error] Error: An unexpected error has occurred.
> [debug] [2022-09-05T11:21:09.063Z] Serving at port 9005

### [REQUIRED] Expected behavior

Expected behaviour is that the function gets deployed successfully without any error.

### [REQUIRED] Actual behavior

Actual behaviour is that function normally get's deployed successfully and just after this update, this issue came into play.

rahuld600:
I could fix this issue just by downgrading the version of firebase tools to v10.0.0 using:
`npm install -g firebase-tools@10.0.0`

rahuld600:
Since there might be something wrong with the new v11.8.0, so It would be really awesome if the engineers took a look at what's up with the latest release...so I am keeping this issue open.

[...]

PR #5236, opened 2022-11-11
Pubsub2

kossnocorp:
### Description

This change solves the bug with pub/sub function v2 deployment caused by the `eventFilters` missing `topic` field:

```
FirebaseError Error: Pub/Sub event trigger is missing topic: {
  "eventType": "google.cloud.pubsub.topic.v1.messagePublished",
  "retry": false,
  "eventFilters": {
    "resource": "projects/XXX/topics/edit-delayed2"
  }
}
```

During the deployment, the API also needs the `resource` field:

```
⚠  functions: HTTP Error: 400, Cannot create trigger projects/chirrapp-staging/locations/us-central1/triggers/publishdelayed2-555223: The request was invalid: generic::invalid_argument: event type google.cloud.pubsub.topic.v1.messagePublished not supported: attribute resource not found within event type
```

I split the PR into two commits, [one adds fallback to `resource`](https://github.com/firebase/firebase-tools/commit/d0dca6cd6973e299e5f19164f4e5116d58ef75c3), and [another that renames `topic` to `resource`](https://github.com/firebase/firebase-tools/commit/d0bb14e6a3b38053cfc9cf475aa03d4618f9b1d6).

I'm not sure if it worked before.

### Scenarios Tested

I've changed this in my local `node_modules` and managed to deploy the function.

### DIFF
--- a/src/gcp/cloudfunctionsv2.ts
+++ b/src/gcp/cloudfunctionsv2.ts
@@ -489,16 +489,17 @@ export function functionFromEndpoint(
       eventType: endpoint.eventTrigger.eventType,
     };
     if (gcfFunction.eventTrigger.eventType === PUBSUB_PUBLISH_EVENT) {
-      if (!endpoint.eventTrigger.eventFilters?.topic) {
+      if (!endpoint.eventTrigger.eventFilters?.topic && !endpoint.eventTrigger.eventFilters?.resource) {
         throw new FirebaseError(
           "Error: Pub/Sub event trigger is missing topic: " +
             JSON.stringify(endpoint.eventTrigger, null, 2)
         );
       }
-      gcfFunction.eventTrigger.pubsubTopic = endpoint.eventTrigger.eventFilters.topic;
+      gcfFunction.eventTrigger.pubsubTopic =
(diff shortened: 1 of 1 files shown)


moshe-moneytor (approved):

kossnocorp:
@moshe-moneytor I'm curious, did you have the same error and the fixed helped you?

stephanos:
Do we know if there is a workaround for this? Or is everyone that uses `onMessagePublished` from `firebase-functions/v2` blocked by this? I've just spent half an hour migrating to v2 only to be stock on this :-/

kossnocorp:
No workaround afaik, I just edited node_modules 🤦‍♂️

kossnocorp:
@bkendall @joehan @yuchenshi could you please take a look?

mqp:
Hey, this is really broken, it's totally impossible to deploy pub/sub v2 functions without patching this.

kossnocorp:
@taeold could you please also take a look at this one, since you started #4952, it's suboptimal to use a manually patched package in order to deploy functions, the fix is trivial.

MahmoudAbuDaya:
We are also manually patching this. Why isn't it being reviewed and merged?

PR #5238, opened 2022-11-11
Add sharp library to Next.js build

cfofiu:
### Description

Checking for /app dir and nextImage to add the sharp library to the Next.js build.

### Scenarios Tested

In order to add the sharp library the code is looking for two things:
* if the new routes `/app` folder exists
* if `isNextImageImported` is set to `true` in `.next/export-marker.json`


src/frameworks/next/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Check if Next.js is forced to serve the source image as-is instead of being oprimized
+ * by setting `unoptimized: true` in next.config.js.
+ * https://nextjs.org/docs/api-reference/next/image#unoptimized
+ *
+ * @param sourceDir location of the source directory
+ * @param distDir location of the dist directory
+ * @return true if image optimization is disabled
+ */
+export async function hasUnoptimizedImage(sourceDir: string, distDir: string): Promise<boolean> {
+  const imageManifest: ImageManifest = (await readJSON(
```
jamesdaniels, line 36: we can drop `: ImageManifest` the `as` is preferred

src/frameworks/next/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+import { readJSON } from "fs-extra";
+import type { ExportMarker, ImageManifest } from "./interfaces";
+
+/**
+ * Check if `/app` directory is used in the Next.js project.
+ * @param sourceDir location of the source directory
+ * @return true if app directory is used in the Next.js project
+ */
+export function usesAppDirRouter(sourceDir: string): boolean {
+  return existsSync(join(sourceDir, "app")) ? true : false;
```
jamesdaniels, line 12: can drop the ternary
would rather we use an async API so we can parallelize down the road with reworking

though I'd prefer we do something a little more robust here. I was thinking we could [check for keys in the app-path-routes-manifest.json](https://github.com/firebase/firebase-tools/blob/0a7b583ef3d5aedc007f085fa7ecf0ed645d9777/src/frameworks/next/index.ts#L109). that said we're doing this because app dir is unstable... hmmmm, maybe we just drop the ternary and presence of app dir at all trips

jamesdaniels (changes requested):
Add changelog entry

src/frameworks/next/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+/**
+ * Check if the project is using the next/image component based on the export-marker.json file.
+ * @param sourceDir location of the source directory
+ * @return true if the Next.js project uses the next/image component
+ */
+export async function usesNextImage(sourceDir: string, distDir: string): Promise<boolean> {
+  const exportMarker: ExportMarker = (await readJSON(
```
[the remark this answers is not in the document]
jamesdaniels: I directed incorrectly, while the larger ecosystem is moving to 'as' in preparation for 'satisifies', Google style-guide is still sticking with `const foo: Bar =`

src/frameworks/next/utils.ts
jamesdaniels: drop `as ImageManifest` redundant

src/frameworks/next/utils.ts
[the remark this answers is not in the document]
cfofiu: I can keep `: ExportMarker` but if I remove `as ExportMarker` I get "Unsafe assignment of an `any` value".

src/frameworks/next/utils.ts
cfofiu: Same as line 20. Please advice.

src/frameworks/next/utils.ts
cfofiu: Agreed, the ternary wasn't necessary since `existsSync` returns a boolean. I am checking the `app-path-routes-manifest.json` now.

src/frameworks/next/utils.ts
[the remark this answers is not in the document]
jamesdaniels: actually can we push up to a generic on readJson? `readJson<ExportMarker>`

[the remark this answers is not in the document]
jamesdaniels: that's in leos PR, i'm working to merge now. can you add some quick unit tests over these new utils?

PR #5245, opened 2022-11-16
Support eventarc trigger channel param

Berlioz:
`resolveChannelName` needs to move to build resolution instead of parsing, because if `channel` is provided as a param, the param has to be resolved into a string literal _before_ we start doing string processing on it.

### DIFF
--- a/src/deploy/functions/build.ts
+++ b/src/deploy/functions/build.ts
@@ -106,7 +106,7 @@ export interface EventTrigger {
 
   // The name of the channel where the function receives events.
   // Must be provided to receive CF3v2 custom events.
-  channel?: string;
+  channel?: string | Expression<string>;
 }
 
 export interface TaskQueueRateLimits {
@@ -425,7 +425,7 @@ export function toBackend(
       regions = [api.functionsDefaultRegion];
     }
     for (const region of regions) {
-      const trigger = discoverTrigger(bdEndpoint, region, r);
+      const trigger = discoverTrigger(bdEndpoint, region, bdEndpoint.project, r);
 
       if (typeof bdEndpoint.platform === "undefined") {
         throw new FirebaseError("platform can't be undefined");
@@ -498,7 +498,12 @@ export function toBackend(
   return bkend;
 }
 
-function discoverTrigger(endpoint: Endpoint, region: string, r: Resolver): backend.Triggered {
+function discoverTrigger(
+  endpoint: Endpoint,
+  region: string,
+  projectId: string,
+  r: Resolver
+): backend.Triggered {
   if (isHttpsTriggered(endpoint)) {
     const httpsTrigger: backend.HttpsTrigger = {};
     if (endpoint.httpsTrigger.invoker === null) {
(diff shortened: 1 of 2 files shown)


src/deploy/functions/build.ts
```diff
@@ -586,3 +595,32 @@ function discoverTrigger(endpoint: Endpoint, region: string, r: Resolver): backe
   }
   assertExhaustive(endpoint);
 }
+
+function interpolateEventArcChannel(channel: string, projectId: string): string {
```
Berlioz, line 599: This is largely a copy-paste from the removed function in v1alpha1.ts, with two changes I'd like to draw attention to:
1) the original function accepted defaultRegion as a parameter, which turns out to always be populated by calling `api.functionsDefaultRegion` if you follow the call chain up.
2) the projectId was once provided directly from a variable passed into the discovery process by the runtime delegate. replacing it with `endpoint.project` is safe in this context because `endpoint.project` is always populated from that same variable in v1alpha1.ts.

src/deploy/functions/build.ts
```diff
@@ -586,3 +595,32 @@ function discoverTrigger(endpoint: Endpoint, region: string, r: Resolver): backe
...
+  if (!channel.includes("/")) {
+    const location = api.functionsDefaultRegion;
+    const channelId = channel;
+    return "projects/" + projectId + "/locations/" + location + "/channels/" + channelId;
+  }
+
+  const CHANNEL_NAME_REGEX = new RegExp(
```
colerogers, line 606: nit: can we move this constant to the top of the file?

src/deploy/functions/build.ts
```diff
@@ -586,3 +595,32 @@ function discoverTrigger(endpoint: Endpoint, region: string, r: Resolver): backe
   }
   assertExhaustive(endpoint);
 }
+
+function interpolateEventArcChannel(channel: string, projectId: string): string {
+  if (!channel.includes("/")) {
+    const location = api.functionsDefaultRegion;
```
taeold, line 601: hmm part of me is surprised that this isn't `region` - shouldn't eventarc channel be in the same region as the function.

src/deploy/functions/build.ts
```diff
@@ -586,3 +595,32 @@ function discoverTrigger(endpoint: Endpoint, region: string, r: Resolver): backe
...
+      "channels\\/" +
+      "(?<channel>[A-Za-z\\d\\-_]+)"
+  );
+  const match = CHANNEL_NAME_REGEX.exec(channel);
+  if (!match?.groups) {
+    throw new FirebaseError("Invalid channel name format: " + channel);
+  }
+  const matchedProjectId = match.groups.project;
+  const location = match.groups.location;
+  const channelId = match.groups.channel;
```
taeold, line 620: nit* equivalent to:

```
const { project: matchedProjectId, location, channel: channeld } = match.groups
```

I can see how this might look less readable though 🤔

src/deploy/functions/build.ts
```diff
@@ -586,3 +595,32 @@ function discoverTrigger(endpoint: Endpoint, region: string, r: Resolver): backe
...
+  if (matchedProjectId) {
+    return "projects/" + matchedProjectId + "/locations/" + location + "/channels/" + channelId;
```
taeold, line 622: I think we shouldn't support cross-project channels. IIUC, eventarc doesn't support such a thing anyway no?

I understand this is mostly copy-and-paste code... but just checking to see if we know the answer to issues like cross-region triggering or cross-project triggering. If we don't know the answers, then we should investigate and maybe add some documentation to it at https://firebase.google.com/docs/functions/beta/custom-events.

Issue #5158, opened 2022-10-22
Web Framework Hosting - Sveltekit


xariusdesai:
### [REQUIRED] Environment info

**firebase-tools:**
11.15.0

**Platform:**
macOS

### [REQUIRED] Steps to reproduce

Create skeleton Sveltekit project
firebase experiments:enable webframeworks
Firebase init hosting
Svelte project auto detected
No other changes to other config files
firebase serve - site working as expected using Vite development server
firebase deploy - builds sveltekit app and then fails during build process

### [REQUIRED] Expected behavior

Build process finishes and deploys to hosting

### [REQUIRED] Actual behavior

Get the following error:
Error: An unexpected error has occurred.

Looking at the log, it seems as if it is trying to get to the client folder, but the path has my base path in twice.
[firebase-debug.log](https://github.com/firebase/firebase-tools/files/9845124/firebase-debug.log)

kddige:
Might be duplicate of #5144 - no solution yet, but problem might have been found looking at the comments.

Nushio:
We'll need a new svelte adapter for this to work, as well as a very small amount of work to the firebase-tool. I'll send the pr later this afternoon for the firebase-tool but I need a bit more time to figure out the best path forward (Either adapt svelte-firebase-adapter to fit this tool, or, make a new tool altogether)

simonnepomuk:
I wrote an [adapter for SvelteKit](https://www.npmjs.com/package/sveltekit-adapter-firebase) which works fine for the _traditional_ Firebase deployment. I'm now trying to integrate web framework support to leverage the features that come with it but I'm a bit stuck. My approach was to build everything and utilize the [custom express integration](https://firebase.google.com/docs/hosting/express). I've [opened a PR](https://github.com/simonnepomuk/monorepo/pull/2) to show what my approach is and filed an issue for framework support [here](https://github.com/FirebaseExtended/firebase-framework-tools/issues/53).   
Any input is very much appreciated!

simonnepomuk:
@TheIronDev  I think [this PR](https://github.com/firebase/firebase-tools/pull/5248) in combination with [this one](https://github.com/simonnepomuk/monorepo/pull/2) should fix the issue.

TheIronDev:
Hi @simonnepomuk ,

First, apologies for the delayed response. And also thank you for opening up the PR(s)!!!!

[...]

PR #5248, opened 2022-11-16
Web Frameworks support: Use ES6 module syntax and use main file in express integration

simonnepomuk:
### Description

This should fix two issues:  
- When trying to integrate custom web frameworks through express the generated handle function tries to import the root path (where the `package.json` lies). Actually, it should import the file specified in the  `main` property of the `package.json`
- When trying to integrate custom web frameworks through express with `"type": "module"` in `package.json` the generated function fails to execute since the generated code contains a `require(...)` statement.

Should close:  
 - [Issue in firebase-framework-tools repo](https://github.com/FirebaseExtended/firebase-framework-tools/issues/53)
 - #5158

### Scenarios Tested

I linked the `firebase-tools` to my branch using `npm link` and ran the tests in the [firebase-framework-tools repo](https://github.com/FirebaseExtended/firebase-framework-tools). Is this how it should be done? I needed to replace the site ID in a couple of files to a site ID which I have access to. 

This was also tested in the branch of the [web framework support PR](https://github.com/simonnepomuk/monorepo/pull/2) of my SvelteKit adapter with `emulators:start` and `deploy`.

### DIFF
--- a/src/frameworks/express/index.ts
+++ b/src/frameworks/express/index.ts
@@ -48,13 +48,15 @@ async function getBootstrapScript(
   const allowRecursion = !entry;
   if (!entry) {
     const {
-      packageJson: { name },
+      packageJson: { name, main },
     } = await getConfig(root);
     try {
-      entry = require(root);
+      entry = require(join(root, main));
       bootstrapScript = `const bootstrap = Promise.resolve(require('${name}'))`;
     } catch (e) {
-      entry = await dynamicImport(root).catch(() => undefined);
+      entry = await dynamicImport(join(root, main)).catch((err: Error) =>
+        console.log("Unable to find main entry point", err)
+      );
       bootstrapScript = `const bootstrap = import('${name}')`;
     }
   }
@@ -63,7 +65,7 @@ async function getBootstrapScript(
   if (typeof handle === "function") {
     return (
       bootstrapScript +
-      ";\nexports.handle = async (req, res) => (await bootstrap).handle(req, res);"
+      ";\nexport const handle = async (req, res) => (await bootstrap).handle(req, res);"
     );
   }
   if (typeof app === "function") {
@@ -72,7 +74,7 @@ async function getBootstrapScript(
       if (typeof express.render === "function") {
         return (
           bootstrapScript +
-          ";\nexports.handle = async (req, res) => (await bootstrap).app(req, res);"
+          ";\nexport const handle = async (req, res) => (await bootstrap).app(req, res);"
         );
       }
     } catch (e) {
--- a/src/frameworks/index.ts
+++ b/src/frameworks/index.ts
@@ -523,10 +523,9 @@ ${firebaseDefaults ? `__FIREBASE_DEFAULTS__=${JSON.stringify(firebaseDefaults)}\
       // TODO move to templates
       await writeFile(
         join(functionsDist, "server.js"),
-        `const { onRequest } = require('firebase-functions/v2/https');
-const server = import('firebase-frameworks');
-exports.ssr = onRequest((req, res) => server.then(it => it.handle(req, res)));
-`
+        `import { onRequest } from 'firebase-functions/v2/https';
+         const server = import('firebase-frameworks');
+         export const ssr = onRequest((req, res) => server.then(it => it.handle(req, res)));`
       );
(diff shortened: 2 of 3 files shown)


simonnepomuk:
I've done some additional testing and it looks like the import is correct, but the generated code needs to be adapted to support ES modules. I'll try to revert the changes to the import path and see if the `custom` integration tests in the `framework-tools` repo still work.

simonnepomuk:
@jamesdaniels 
I think this might need some functionality to determine if `"type": "module"` exists in the `package.json` and either load the adapted code from this PR or the preexisting code that is currently being replaced by this PR. Otherwise it probably breaks CJS code.

jamesdaniels (changes requested):
While I'm testing on my machine, mind adding a changelog entry?

simonnepomuk:
> While I'm testing on my machine, mind adding a changelog entry?

Will do.

src/frameworks/index.ts
```diff
@@ -523,10 +523,9 @@ ${firebaseDefaults ? `__FIREBASE_DEFAULTS__=${JSON.stringify(firebaseDefaults)}\
...
       await writeFile(
         join(functionsDist, "server.js"),
-        `const { onRequest } = require('firebase-functions/v2/https');
-const server = import('firebase-frameworks');
-exports.ssr = onRequest((req, res) => server.then(it => it.handle(req, res)));
-`
+        `import { onRequest } from 'firebase-functions/v2/https';
+         const server = import('firebase-frameworks');
+         export const ssr = onRequest((req, res) => server.then(it => it.handle(req, res)));`
       );
```
austincrim, line 529: Agree with your comment on the PR that we will need some logic to retain the old CJS behavior as well.

Should be straightforward with something like `if (packageJson.type === 'module') ...`.

We could extract the snippets to `const`s, keep them as inline literals, or even bring them out into the `templates` folder like the `TODO` comment suggests, but I'll defer to @jamesdaniels on that one.

src/frameworks/express/index.ts
```diff
@@ -63,7 +65,7 @@ async function getBootstrapScript(
   if (typeof handle === "function") {
     return (
       bootstrapScript +
-      ";\nexports.handle = async (req, res) => (await bootstrap).handle(req, res);"
+      ";\nexport const handle = async (req, res) => (await bootstrap).handle(req, res);"
```
austincrim, line 68: Same comment regarding CJS support

src/frameworks/express/index.ts
```diff
@@ -72,7 +74,7 @@ async function getBootstrapScript(
       if (typeof express.render === "function") {
         return (
           bootstrapScript +
-          ";\nexports.handle = async (req, res) => (await bootstrap).app(req, res);"
+          ";\nexport const handle = async (req, res) => (await bootstrap).app(req, res);"
```
austincrim, line 77: Same comment regarding CJS support

src/frameworks/index.ts
jamesdaniels: I could see

```ts
const importStatements = packageJson.type === 'module' ?
  "import { onRequest } from 'firebase-functions/v2/https';" : 
  "const { onRequest } = require('firebase-functions/v2/https');"`
```

if we wanted to DRY it up

```ts
const importsFrom = [["onRequest"], "firebase-functions/v2/https"];
const importStatements = importsFrom.map(([imports, from]) => packageJson.type === "module" ? ... : ...);
...
```

austincrim: We'll need the same for the export, right? Might be easier/clearer to just repeat the whole block.

jamesdaniels: SG, let's just repeat the block

austincrim:
@simonnepomuk Let me know if you're still up for getting this merged. If not, we can adapt your PR and move forward ourselves, with attribution back to your work.

simonnepomuk:
> @simonnepomuk Let me know if you're still up for getting this merged. If not, we can adapt your PR and move forward ourselves, with attribution back to your work.

Hey @austincrim, thx for the review. My interest in using firebase dwindled over the last months because of some regulatory issues. Feel free to move forward.

austincrim:
#5540

PR #5267, opened 2022-11-25
Add support for Firestore TTL

JU-2094:
### Description

API change to add Firestore TTL support for the CLI.
The implementation of the design change is tracked at b/249528585

### Scenarios Tested

User Journeys tested:

* Set and Unset TTL in an existing Field override.
* Set and Unset TTL in a new configuration file.
* Don't affect existing TTL policies created without the CLI for backward compatibility.
* See other TTL policies not present in the configuration file.
* Don't affect TTL policies if the `ttl` property is undefined. Only take effect when is explicitly set.
* Remove field overrides and TTL policies when the `--force` flag is used. 
* Verify there's only 1 TTL per collection.
* Verify that existing behavior remains if not using new property.
* Verify the order of field overrides in the configuration file do not affect.
* Verify the type of the `ttl` property is enforced to be boolean.


src/firestore/indexes.ts
```diff
@@ -334,11 +339,17 @@ export class FirestoreIndexes {
    * Validate that an arbitrary object is safe to use as an {@link Spec.FieldOverride}.
    * @param field
    */
-  validateField(field: any): void {
+  validateField(field: any, collectionsWithTtl: Set<string>): void {
     validator.assertHas(field, "collectionGroup");
     validator.assertHas(field, "fieldPath");
     validator.assertHas(field, "indexes");
 
+    /** Only one TTL per Collection */
```
HansiMou, line 347: Is this check needed? Backend has the same check and should delegate this to backend.

src/firestore/indexes.ts
```diff
@@ -395,7 +416,7 @@ export class FirestoreIndexes {
     const url = field.name;
```
HansiMou, line 416: nit: The comments for this method should be updated(on line 413; cannot directly comment on that line for some reason)

src/firestore/indexes.ts
JU-2094: It was agreed to have it in the design doc. 

The deployment of the indexes & field overrides is sequential. In a first pass all fields are validated. 
Then a second pass will be done for actual patching. I added this check to catch it earlier and show an error message from CLI instead of 4xx, 5xx codes.

Removing the asserts for some edge cases is not a problem. It would prompt a 403 or 500 error with the same messages as GCP console.

HansiMou: IMO it's better to let backend decide. This is not an accurate check either since right now we have 1 `UP` ttl config per collection. This limit could also be changed in the future.

JU-2094: SG. I'll remove this check. The HTTP error has clear information on this.

HansiMou:
LGTM from TTL backend.

bkendall (approved):

PR #5275, opened 2022-11-28
Fix bin paths for angular & vite when using NPM workspaces

TatsuyaYamamoto:
### Description

When `firebase-tools` detects Next.js, it tries to run `next build` and `next export`. `firebase-tools` creates "next bin" path with `path.join` to run `next export` but currently, it supports the case only that “node_modules/.bin/next" is in the same directory as `.next` dir.

This PR makes firebase-tools also load “node_modules/.bin/next” in another dir. In my case, it's a useful fix when using npm workspace.

(1): next bin that Current `firebase-tools` can load
(2): next bin that I want to make `firebase-tools` load

```
repo_root/
    ├ node_modules/.bin/next (2)
    ├ workspace_packages/
    │    └ next-app/
    │        ├ .next/
    │        ├ node_modules/.bin/next (1)
```

#### non-monorepo

```
$ npx create-next-app@latest firebase-tools-test
$ cd firebase-tools-test

$ echo "export default () => <div>static page</div>" > pages/static-page.js
$ echo "module.exports = {images: { unoptimized: true }};" > next.config.js

$ npm link firebase-tools
$ npx firebase init
$ npx firebase-tools deploy --only hosting

$ ls -1 .firebase/*/hosting/*.html
.firebase/***/hosting/404.html
.firebase/***/hosting/index.html
.firebase/***/hosting/static-page.html 👈 expected!
```

#### monorepo

```
$ mkdir firebase-tools-test-monorepo
$ cd firebase-tools-test-monorepo
$ echo '{ "workspaces": ["./packages/next-app"] }' > package.json

$ mkdir packages
$ cd packages
$ npx create-next-app@latest next-app
$ echo "export default () => <div>static page</div>" > next-app/pages/static-page.js
$ echo "module.exports = {images: { unoptimized: true }};" > next-app/next.config.js
$ cd ..

$ npm link firebase-tools
$ npx firebase init
$ jq '.hosting.source="packages/next-app"' firebase.json > firebase.new.json
$ cat firebase.new.json > firebase.json
$ npx firebase-tools deploy --only hosting

$ ls -1 .firebase/*/hosting/*.html
.firebase/***/hosting/404.html
.firebase/***/hosting/index.html
.firebase/***/hosting/static-page.html 👈 expected!
```


src/frameworks/next/index.ts
```diff
@@ -40,7 +40,11 @@ interface Manifest {
 }
 
 const CLI_COMMAND = join(
-  "node_modules",
+  require.resolve("next"), // ./dist/server/next.js (ref: https://github.com/vercel/next.js/blob/canary/packages/next/package.json)
```
jamesdaniels, line 43: Could we instead get the NPM root somehow? `execSync("npm root", { cwd: ... })` making the assumption here is spooking me. Also we should add similar logic to the angular integration

jamesdaniels (changes requested):
Should add a changelog too

src/frameworks/next/index.ts
TatsuyaYamamoto: > Could we instead get the NPM root somehow?

I changed the following and re-tried "Scenarios Tested”. It looks fine.

https://github.com/firebase/firebase-tools/pull/5275/files#diff-372d171b067c11f7dbdea434bc98b44c356278e71e0f1251758e4ca25517c0b8R43

> Also we should add similar logic to the angular integration

I changed similarly vite and anular integrations. (5507397)

> Should add a changelog too

[TODO] I’ll update PR title and chagelog.md after changing policy is decided.

atanaskanchev:
Would this fix a similar issue for other monorepos like turborepo/nx as reported [here](https://github.com/FirebaseExtended/firebase-framework-tools/issues/37)?

src/frameworks/angular/index.ts
```diff
@@ -20,7 +20,11 @@ export const name = "Angular";
 export const support = SupportLevel.Experimental;
 export const type = FrameworkType.Framework;
 
-const CLI_COMMAND = join("node_modules", ".bin", process.platform === "win32" ? "ng.cmd" : "ng");
+const CLI_COMMAND = join(
+  execSync("npm root").toString().trim(),
```
jamesdaniels, line 24: additional thought, we should make sure this is relative to the source directory

we should write it as a function: 

```ts
function getCLI(cwd: string) {
  return join(
    execSync("npm root", { cwd }),
    ...
```

jamesdaniels (changes requested):

jamesdaniels:
Addressed in #5640, closing this out. Thanks for your help

Issue #5270, opened 2022-11-25
Function Deploy - Precondition failed. Cannot update a GCF function without sourceUrl


ensgenc:
Hi, What is this GCF function without sourceUrl,

The same package deploys in other environments, but only one environment has this error.

Environment:
Microsoft azure ubuntu 20.04

![Screen Shot 2022-11-26 at 01 46 01](https://user-images.githubusercontent.com/6282510/204061218-d0d748d7-fa7c-41ba-91e6-0bc4368f9190.png)
![Screen Shot 2022-11-26 at 01 46 15](https://user-images.githubusercontent.com/6282510/204061225-29a8e60a-1be2-42ec-807e-166028442a58.png)

Do you have any idea

thanks

taeold:
Hey @ensgenc. Is this something that's happening consistently on your function deploys?

If you can provide us with a small, reproducible example, we'd appreciate it!

PR #5280, opened 2022-11-30
Fix bug using --only filter to disable no-op deploy fails deploy because function source isn't uploaded

taeold:
Fixes https://github.com/firebase/firebase-tools/issues/5270


TheIronDev (approved):
Lgtm!

Issue #5290, opened 2022-12-01
functions:delete fails to match function name unless dot notation is used


twaddington:
### Environment info

**firebase-tools:** 11.16.1
**Platform:** macOS 12.6.1

### Test case

```
// index.ts
export * as foo from "./providers/foo.js";

// foo.ts
export const bar = functions.pubsub
  .schedule("every day 09:00")
  .onRun(async () => {
    console.log("Hello Firebase");
    return null;
  });

```

### Steps to reproduce

```
// Fails with error
$ firebase functions:delete foo-bar
Error: The specified filters do not match any existing functions in project.

// Works as expected
$ firebase functions:delete foo.bar
...

// Deploy works the same with either wording
$ firebase deploy --only functions:foo-bar
$ firebase deploy --only functions:foo.bar
```

### Expected behavior

The command `firebase functions:delete foo-bar` should succeed since that's how the function name appears when listed via `firebase functions:list`.

### Actual behavior

The command `firebase functions:delete foo-bar` fails with an error.

twaddington:
From poking around the source it looks like the problem is here:

https://github.com/firebase/firebase-tools/blob/9b4c23398838a79ef406f2d25ce41ece2b445286/src/commands/functions-delete.ts#L36-L39

The deployment command uses a different parsing function here:

https://github.com/firebase/firebase-tools/blob/9b4c23398838a79ef406f2d25ce41ece2b445286/src/deploy/functions/functionsDeployHelper.ts#L59-L81

PR #5292, opened 2022-12-01
Fix functions:delete id filter matching.

twaddington:
Fixes https://github.com/firebase/firebase-tools/issues/5290.


taeold (approved):
🔥  Thank you!

PR #5294, opened 2022-12-02
Pubsub force shutdown if shutdown isn't working.

christhompsongoogle:
Pubsub force shutdown if shutdown isn't working.


src/emulator/pubsubEmulator.ts
```diff
@@ -11,6 +11,14 @@ import { FirebaseError } from "../error";
...
 import { SignatureType } from "./functionsEmulatorShared";
 import { CloudEvent } from "./events/types";
+import { exec } from "child_process";
+
+// Finds processes with "pubsub-emulator" in the description and runs `kill` if any exist
+// Since the pubsub emulator doesn't export any data, force-killing will not affect export-on-exit
+// Note the `[p]` is a workaround to avoid selecting the currently running `ps` process.
+const PUBSUB_KILL_COMMAND =
+  "pubsub_pids=$(ps aux | grep '[p]ubsub-emulator' | awk '{print $2}');" +
+  " if [ ! -z '$pubsub_pids' ]; then kill -9 $pubsub_pids; fi;";
```
bkendall, line 21: I'm a little afraid of what this would do on Windows - it's very obviously not going to work (because it's using unix tooling), but will this not cause issues when we try to stop?

src/emulator/pubsubEmulator.ts
```diff
@@ -62,7 +70,19 @@ export class PubsubEmulator implements EmulatorInstance {
   }
 
   async stop(): Promise<void> {
-    await downloadableEmulators.stop(Emulators.PUBSUB);
+    try {
+      await downloadableEmulators.stop(Emulators.PUBSUB);
+    } catch (e: unknown) {
+      this.logger.logLabeled("DEBUG", "pubsub", JSON.stringify(e));
+      exec(PUBSUB_KILL_COMMAND, (err, stdout) => {
```
bkendall, line 77: Yeah, here - if we do this on Windows, are issues silently logged and it moves on?

src/emulator/pubsubEmulator.ts
christhompsongoogle: Good point - I need to get a windows machine to test on. I'll hold off on this righ tnow

src/emulator/pubsubEmulator.ts
christhompsongoogle: Added a windows check and will follow-up if we see any windows issues (and test once I have a machine to play with)

src/emulator/pubsubEmulator.ts
christhompsongoogle: Added a windows check PTAL

src/emulator/pubsubEmulator.ts
```diff
@@ -62,7 +70,21 @@ export class PubsubEmulator implements EmulatorInstance {
   }
 
   async stop(): Promise<void> {
-    await downloadableEmulators.stop(Emulators.PUBSUB);
+    try {
+      await downloadableEmulators.stop(Emulators.PUBSUB);
+    } catch (e: unknown) {
+      this.logger.logLabeled("DEBUG", "pubsub", JSON.stringify(e));
+      if (process.platform === "win32") {
```
bkendall, line 77: This check looks like the opposite of what you wanted. Do you want `!==`?

bkendall (changes requested):

src/emulator/pubsubEmulator.ts
christhompsongoogle: Correct thanks for pointing it out

src/emulator/pubsubEmulator.ts
```diff
@@ -62,7 +70,21 @@ export class PubsubEmulator implements EmulatorInstance {
   }
 
   async stop(): Promise<void> {
-    await downloadableEmulators.stop(Emulators.PUBSUB);
+    try {
+      await downloadableEmulators.stop(Emulators.PUBSUB);
+    } catch (e: unknown) {
+      this.logger.logLabeled("DEBUG", "pubsub", JSON.stringify(e));
+      if (process.platform !== "win32") {
+        exec(PUBSUB_KILL_COMMAND, (err, stdout) => {
```
bkendall, line 78: I just noticed: exec isn't synchronous, should this be wrapped in a Promise and awaited, or use execSync?

bkendall (changes requested):

stfsy:
@christhompsongoogle having this PR merged would be very, very handy :) Can I support somehow?

src/emulator/pubsubEmulator.ts
christhompsongoogle: This can be a fire-and-forget so long as Node doesn't terminate before executing it. In that case I think we'll just be missing the output. At any rate I think wrapping in a promise like this should do the trick.

christhompsongoogle:
> @christhompsongoogle having this PR merged would be very, very handy :) Can I support somehow?

Sorry I've been a little distracted on this one. I'll have this in by Jan 6 once our release process has exited the holiday freeze.

src/emulator/pubsubEmulator.ts
```diff
@@ -62,7 +70,26 @@ export class PubsubEmulator implements EmulatorInstance {
   }
 
   async stop(): Promise<void> {
-    await downloadableEmulators.stop(Emulators.PUBSUB);
+    try {
+      await downloadableEmulators.stop(Emulators.PUBSUB);
+    } catch (e: unknown) {
+      this.logger.logLabeled("DEBUG", "pubsub", JSON.stringify(e));
+      if (process.platform !== "win32") {
+        // The exec is wrapped in a promise so we can block on it's completion via `await`.
```
bkendall, line 78: This is fine, but it might be cleaner to use [execSync](https://nodejs.org/api/child_process.html#child_processexecsynccommand-options)

src/emulator/pubsubEmulator.ts
```diff
@@ -62,7 +70,26 @@ export class PubsubEmulator implements EmulatorInstance {
...
+        const execPromise = new Promise((resolve) => {
+          exec(PUBSUB_KILL_COMMAND, (err, stdout) => {
+            if (err) {
+              this.logger.logLabeled("DEBUG", "pubsub", JSON.stringify(err));
+            }
+            if (stdout) {
+              this.logger.logLabeled("DEBUG", "pubsub", JSON.stringify(stdout));
+            }
+            resolve(null);
```
bkendall, line 87: Passing `null` isn't necessary.

```suggestion
            resolve();
```

bkendall (approved):

bgazzera:
For anyone coming at this issue when using an Alpine-based linux distro (most Docker images), the fix won't work because of the way `ps aux` works in Alpine (it reverses the pid & user columns in the output).

When terminating the pubsub emulator you will see a message like:
```
Stopping Pub/Sub Emulator 
sh: invalid number 'root'
sh: invalid number 'root'
```
A solution is to install the procps package, like so: `apk --no-cache add procps`

PR #5298, opened 2022-12-02
Update default Storage rules to contain example cross-srevice-rules

sam-gc:
To help visibility for this feature


templates/init/storage/storage.rules
```diff
@@ -1,4 +1,8 @@
 rules_version = '2';
+
+// Craft rules based on data in your Firestore database
```
joehan, line 3: Consider including a relevant docs links here?

joehan (approved):
Good idea!

templates/init/storage/storage.rules
sam-gc: Ah good point. This copy was approved with PM so let me double check about the link and get back to this

yuchenshi: Please move this into a `match` block otherwise it won't work when you uncomment it as-is

sam-gc: It's meant not to be uncommentable like that (to prevent accidentally opening things up)

sam-gc: I'm going to go ahead and merge this as-is. These rules match the Console and in Console there's a banner already with a link to the docs.

PR #5305, opened 2022-12-05
Improve handling of bad versions during ext:install

joehan:
### Description
Noticed that ext:install was not working correctly for stripe/firestore-stripe-payments, because there was a badly formed extension in our backend that did not have a version number. I'm gonna investigate that, but figured I should also make this code less brittle.


src/extensions/extensionsHelper.ts
```diff
@@ -834,13 +834,17 @@ export async function diagnoseAndFixProject(options: any): Promise<void> {
  * 2. Infer "latest" as the version if not provided
  */
 export async function canonicalizeRefInput(extensionName: string): Promise<string> {
```
elvisun, line 836: the word "name" has lost its value (at least to me...) consider naming this `input` and the variable `inferredRef`

src/test/deploy/extensions/planner.spec.ts
```diff
@@ -26,13 +26,12 @@ describe("Extensions Deployment Planner", () => {
...
-        .resolves([
-          extensionVersion("0.1.0"),
-          extensionVersion("0.1.1"),
-          extensionVersion("0.2.0"),
-        ]);
+      listExtensionVersionsStub = sinon.stub(extensionsApi, "listExtensionVersions").resolves([
+        extensionVersion("0.1.0"),
+        extensionVersion("0.1.1"),
+        extensionVersion("0.2.0"),
+        extensionVersion(), // Explicitly test that this doesn't break on bad data
```
elvisun, line 33: 👍 👍

elvisun (approved):

PR #5306, opened 2022-12-06
Added support for CF3v2 functions in the Extensions emulator.

pavelgj:


src/extensions/billingMigrationHelper.ts
```diff
@@ -41,7 +41,16 @@ function hasRuntime(spec: ExtensionSpec, runtime: string): boolean {
   const specVersion = spec.specVersion || defaultSpecVersion;
   const defaultRuntime = defaultRuntimes[specVersion];
   const resources = spec.resources || [];
-  return resources.some((r) => runtime === (r.properties?.runtime || defaultRuntime));
+  return resources.some((r) => {
+    switch (r.type) {
```
joehan, line 45: Seems like this could use getResourceRuntime to dedupe code a bit?

src/extensions/emulator/triggerHelper.ts
```diff
@@ -4,38 +4,93 @@ import {
...
-    entryPoint: resource.name,
-    platform: "gcfv1",
-  };
-  const properties = resource.properties || {};
-  proto.convertIfPresent(etd, properties, "timeoutSeconds", "timeout", proto.secondsFromDuration);
-  proto.convertIfPresent(etd, properties, "regions", "location", (str: string) => [str]);
-  proto.copyIfPresent(etd, properties, "availableMemoryMb");
-  if (properties.httpsTrigger) {
-    etd.httpsTrigger = properties.httpsTrigger;
+  const resourceType = resource.type;
```
joehan, line 22: Its a little awkward to define this here and then switch between resource.type and resourceType throughout the code. Prefer getting rid of this and just using resource.type

joehan (approved):
Some small improvments but mostly LGTM. If you're feeling inspired, consider adding an extension with v2 functions in it to the integration test as well: https://github.com/firebase/firebase-tools/tree/master/scripts/extensions-emulator-tests

src/extensions/emulator/triggerHelper.ts
pavelgj: I had to define this var... basically if resource.type is not v1 AND not v2 then typescript infers the type of resource as `never`, so I can't use `resource` after the second if...

src/extensions/billingMigrationHelper.ts
pavelgj: Ah, good catch! Done!

Issue #4605, opened 2022-06-01
Functions cannot access Firebase Secrets, when i debug (inspect) mode


if-fi:
### [REQUIRED] Environment info

**firebase-tools:11.0.1**

**Platform:MacOS**

### [REQUIRED] Test case

```javascript
export const testSecretAccess = functions
  .region("europe-west3")
  .runWith({
    secrets: ["TEST_API_USER"],
  })
  .https.onRequest(async (request, response) => {

    functions.logger.info(`SECRET: ${process.env.TEST_API_USER}`);

    response.send({
      status: process.env.TEST_API_USER ? "success" : "failed",
    });
  });
```

### [REQUIRED] Steps to reproduce

1. Add a secret: `firebase functions:secrets:set TEST_API_USER`
2. Run functions emulator in debug mode: `firebase emulators:start --inspect-functions --only functions`
3. Trugger the function `http://localhost:5001/<PROJECT_NAME>/europe-west3/testSecretAccess`

### [REQUIRED] Expected behavior

The value of the secret is logged, and "success" is returned as a response.

Log:
```
>  {"severity":"INFO","message":"SECRET: test"}
```

Response:
```
{"status":"success"}
```

### [REQUIRED] Actual behavior

Log:
```
>  {"severity":"INFO","message":"SECRET: undefined"}
```

Response:
```
{"status":"failed"}
```

_NB: The expected result is achieved, when we run the emulator **without** debug mode: `firebase emulators:start --only functions`. **The probelm appears only in debug mode.**_

Adam-Beno:
Running into the same issue:
```
firebase-tools@11.8.0

Platform: Windows 10
```

justinkrol:
I'm also running into this issue.
```
firebase-tools@11.13.0
```
Platform: macOS

@taeold is this a known limitation of the `--inspect-options` flag?

Thaina:
I still got this warning in 11.16.0 when run in firebase emulator with `--inspect-functions`

>  {"severity":"WARNING","message":"No value found for secret parameter \"DIALOGFLOW_KEYFILE\". A function can only access a secret if you include the secret in the function's dependency array."}
> 
> 
>  {"severity":"WARNING","message":"No value found for secret parameter \"FACEBOOK_TOKENS\". A function can only access a secret if you include the secret in the function's dependency array."}

![image](https://user-images.githubusercontent.com/1042507/201863017-14e45748-d7d1-408b-ab40-6985cfc5abcb.png)

[...]

Issue #5062, opened 2022-10-05
Extension Emulator don’t load secretEnvs silently.


kazuooooo:
### [REQUIRED] Environment info

**firebase-tools: `v11.13.0`**

**Platform: `macOS`**

### [REQUIRED] Test case

1. install [stripe firebase extension](https://firebase.google.com/products/extensions/stripe-firestore-stripe-payments)
2. run extensions by emulator
3. Call `ext-firestore-stripe-payments-createPortalLink` in browser

```tsx
import { httpsCallable } from 'firebase/functions'

const functionRef = httpsCallable<{ returnUrl: string }, { url: string }>(
  functions,
  'ext-firestore-stripe-payments-createPortalLink'
)
const { data } = await functionRef({
  returnUrl: window.location.href
})
window.location.assign(data.url)
```

### [REQUIRED] Steps to reproduce

The same above.

### [REQUIRED] Expected behavior

Emulator load from`STRIPE_API_KEY` and `STRIPE_WEBHOOK_SECRET` from Secret Manager.

```tsx
STRIPE_API_KEY=projects/${param:PROJECT_NUMBER}/secrets/firestore-stripe-payments-STRIPE_API_KEY/versions/latest
STRIPE_WEBHOOK_SECRET=projects/${param:PROJECT_NUMBER}/secrets/firestore-stripe-payments-STRIPE_WEBHOOK_SECRET-fd8d/versions/latest
```

### [REQUIRED] Actual behavior

Stripe Raise this error, because No API key provided from Firebase Runtime.

```tsx
>  {"severity":"INFO","message":"⚙️ Creating checkout session for doc [nNEzStB2wCK3QbX9FAg7]."}
>  {"severity":"INFO","message":"⚙️ Creating customer object for [0uwT7MhNb8pjHjJZ1DMPhW8z9Wct]."}
>  {"severity":"ERROR","message":"❗️[Error]: Failed to create customer for [0uwT7MhNb8pjHjJZ1DMPhW8z9Wct]: You did not provide an API key. You need to provide your API key in the Authorization header, using Bearer auth (e.g. 'Authorization: Bearer YOUR_SECRET_KEY'). See https://stripe.com/docs/api#authentication for details, or we can help at https://support.stripe.com/."}
>  {"severity":"ERROR","message":"❗️[Error]: Checkout session creation failed for doc [nNEzStB2wCK3QbX9FAg7]: Cannot read properties of null (reading 'stripeId')"}
```

[...]

PR #5308, opened 2022-12-07
Load secrets when emulating functions with --inspect-function flag

joehan:
### Description
Load secrets when emulating functions with --inspect-function flag. Fixes https://github.com/firebase/firebase-tools/issues/4605 and #5062 
### Scenarios Tested
Tested that extensions and CF3 codebases correctly load secrets, both with and without the --inspect-functions flag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,2 @@
 - Add support for Firestore TTL (#5267)
+- Fix bug where secrets were not loaded when emulating functions with `--inpsect-functions`. (#4605)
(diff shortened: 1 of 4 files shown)


taeold (approved):
🔥

gbourne1:
Secrets are loaded at emulator startup with the " --inspect-functions" flag, but hot reloads (changing the code and watch reloads) cause the secrets to return to undefined. The emulator needs to then be restarted for the secrets to load again.

PR #5310, opened 2022-12-07
Fix output buffer pipe issue

### LINKED ISSUE #3286: firebase cli --json output is truncated to 64000 bytes when piped to another command

**firebase-tools:** 9.10.0

**Platform:** Ubuntu 20.04

### [REQUIRED] Test case

With a project where this outputs more than 64KB, run:

```
firebase ext:list --json | jq
```
This is failing for me, with the output truncating at 64000 bytes

Enable several Firebase extensions in a project
Run a command that pipes the json output to another command, eg:
```firebase ext:list --json | jq```

(or maybe run any firebase cli command that supports --json output, and produces > 64KB).

Valid json should be piped.

When piped to another command the JSON seems to be terminated at 64000.

Using the `unbuffer` command seems to workaround this, but ideally this wouldn't be required.

eg, this works as expected:

```
unbuffer firebase ext:list --json | jq
```

maneesht:
Fix #3286 
Most likely we're exiting earlier than the pipe can write extra information in the buffer. Tested this with the latest version of flutterfire CLI as it relies on the output of projects:list.


guidezpl:
@bkendall hey, can I get your review on this? it's affecting a downstream Flutter package (https://github.com/invertase/flutterfire_cli/issues/150)

joehan (approved):

joehan:
@maneesht Could you add a CHANGELOG for thiss?

PR #5316, opened 2022-12-08
Leoortizz nextjs rewrites

jamesdaniels:
Googler mirror of #5212 for integration tests


src/frameworks/next/utils.ts
```diff
@@ -0,0 +1,103 @@
...
+export function pathHasRegex(path: string): boolean {
+  // finds parentheses that are not preceded by double backslashes
+  return /(?<!\\)\(/.test(path);
+}
+
+/**
+ * Remove double backslashes from a string
+ */
+export function cleanEscapedChars(path: string): string {
+  return path.replace(/\\/g, "");
```
jamesdaniels, line 24: @leoortizz `path.replace(/\\([\(\){}\:\+\?])/g, (a,b) => b)` should do it IMO. WDYT?

```
> '/\\(\\)\\{\\}\\:\\+\\?/:slug'.replace(/\\([\(\){}\:\+\?])/g, (a,b) => b)
'/(){}:+?/:slug'
```

jamesdaniels: "The following characters (, ), {, }, :, *, +, ? are used for regex path matching, so when used in the source as non-special values they must be escaped by adding \\ before them"

PR #5320, opened 2022-12-09
Paths that have middleware (Next.js) should route through Cloud Functions

leoortizz:
### Description

Wrapping up #5195


src/frameworks/next/index.ts
```diff
@@ -390,6 +391,22 @@ export async function ɵcodegenFunctionsDirectory(sourceDir: string, destDir: st
...
+      pathExists(join(dir, "middleware.ts")),
+    ]);
+
+    if (middlewareJs || middlewareTs) {
+      throw new FirebaseError(
+        `${clc.bold("firebase serve")} does not support Next.js Middleware. Please use ${clc.bold(
+          "firebase emulators:start"
+        )} instead.`
+      );
+    }
```
leoortizz, line 407: @jamesdaniels had some issues trying to import those files to check the exports to know whether we should error or not, so for now it's just checking if some of the middleware files exists

src/frameworks/next/index.ts
```diff
@@ -1,10 +1,25 @@
...
 import type { NextConfig } from "next";
+import {
+  APP_PATH_ROUTES_MANIFEST,
+  EXPORT_MARKER,
+  IMAGES_MANIFEST,
+  MIDDLEWARE_MANIFEST,
+  PAGES_MANIFEST,
+  PRERENDER_MANIFEST,
+  ROUTES_MANIFEST,
+} from "next/constants";
```
jamesdaniels, line 13: Are these going to work in prod, since next is just a dev dep? We might need to dynamically import them or bundle

src/frameworks/next/index.ts
jamesdaniels: can we add a dev/prod switch to isUsingMiddleware?

src/frameworks/next/utils.ts
```diff
@@ -150,3 +151,10 @@ export async function hasUnoptimizedImage(sourceDir: string, distDir: string): P
   );
   return imageManifest.images.unoptimized;
 }
+
+/**
+ * Whether Next.js middleware is being used
+ */
+export function isUsingMiddleware(middleware: MiddlewareManifest["middleware"]): boolean {
```
jamesdaniels, line 158: I'd add the manifest reading in here

jamesdaniels: test coverage

src/frameworks/next/index.ts
```diff
@@ -84,77 +96,104 @@ export async function build(dir: string): Promise<BuildResult> {
...
-        !(
-          ["/_app", "/", "/_error", "/_document", "/404"].includes(it) ||
-          prerenderedRoutes.includes(it) ||
-          dynamicRoutes.includes(it)
-        )
-    );
-    // TODO log these as a reason why Cloud Functions are needed
-    if (!anyDynamicRouteFallbacks && unrenderedPages.length === 0) {
-      wantsBackend = false;
+  const { isNextImageImported } = await readJSON<ExportMarker>(join(dir, distDir, EXPORT_MARKER));
```
jamesdaniels, line 110: I know outside of the scope here but maybe we could refactor into an usingImageOptimization util fn?

jamesdaniels (changes requested):

src/frameworks/next/index.ts
leoortizz: right, forgot about that. I used them as types moved all to a constants file

src/frameworks/next/index.ts
leoortizz: done! am planning to do the same for almost everything that affects `reasonsForBackend`, e.g. `isUsingAppDirectory`, `isUsingFallback`, `isUsingRevalidate`, ...

src/frameworks/next/index.ts
```diff
@@ -84,77 +95,95 @@ export async function build(dir: string): Promise<BuildResult> {
...
-      wantsBackend = false;
+  if (await isUsingImageOptimization(join(dir, distDir))) {
+    reasonsForBackend.push(`Image Optimization`);
+  }
+
+  const appPathRoutesManifestPath = join(dir, distDir, APP_PATH_ROUTES_MANIFEST);
+  const appPathRoutesManifestJSON = fileExistsSync(appPathRoutesManifestPath)
+    ? await readJSON(appPathRoutesManifestPath)
+    : {};
+  const usingAppDirectory = Object.keys(appPathRoutesManifestJSON).length > 0;
```
jamesdaniels, line 113: mind refactor to isUsingAppDirectory? Maybe we can port some of these checks into that util

jamesdaniels (changes requested):
Looking good. I'm going to pull down and run this locally since we have a lot of refactoring here, I kicked off the integration tests. Add changelog when you can.

src/frameworks/next/index.ts
leoortizz: sure, will do that. I'm about to do the same refactors as mentioned [here](https://github.com/firebase/firebase-tools/pull/5320#discussion_r1046604060)

jamesdaniels (changes requested):

src/frameworks/next/utils.ts
```diff
@@ -145,8 +153,49 @@ export async function usesNextImage(sourceDir: string, distDir: string): Promise
...
+  if (isDevMode) {
+    const [middlewareJs, middlewareTs] = await Promise.all([
+      pathExists(join(dir, "middleware.js")),
+      pathExists(join(dir, "middleware.ts")),
+    ]);
+
+    return middlewareJs || middlewareTs;
+  } else {
+    const middlewareManifest: MiddlewareManifest = await readJSON<MiddlewareManifest>(
+      join(dir, MIDDLEWARE_MANIFEST)
```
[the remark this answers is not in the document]
jamesdaniels: why this didn't trip the integration test I do not know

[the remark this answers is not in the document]
jamesdaniels: need distDir on there too

src/frameworks/next/index.ts
jamesdaniels: actually let's stick with just that the file exists, like is in the util... it is unstable after all, next could break the contract

jamesdaniels (approved):
Just add changelog and we're good

PR #5321, opened 2022-12-10
Nuxt support

cfofiu:
### Description

This PR handles basic Nuxt 2 and Next 3 support.

### Scenarios Tested

Covers basic Nuxt 3 with limited configuration.
Nuxt 2 has a couple more use cases, but the following should be handled:
* `ssr: false`
* `ssr: true`
* `target: 'server'`
* `target: 'static'`
* custom generate directory

### Sample Commands
Deploying Nuxt 2 and 3 codebase via `firebase emulators:start`.


src/frameworks/nuxt/index.ts
```diff
@@ -1,33 +1,39 @@
 import { copy, pathExists } from "fs-extra";
 import { readFile } from "fs/promises";
-import { basename, join } from "path";
+import { join } from "path";
 import { gte } from "semver";
-import { BuildResult, findDependency, FrameworkType, relativeRequire, SupportLevel } from "..";
+import { findDependency, FrameworkType, relativeRequire, SupportLevel } from "..";
 
-export const name = "Nuxt";
+export const name = "Nuxt 3";
```
jamesdaniels, line 7: I think we can drop the 3 here.

src/frameworks/nuxt/index.ts
```diff
@@ -1,33 +1,39 @@
...
   const configFilesExist = await Promise.all([
     pathExists(join(dir, "nuxt.config.js")),
     pathExists(join(dir, "nuxt.config.ts")),
   ]);
+
   const anyConfigFileExists = configFilesExist.some((it) => it);
   if (!anyConfigFileExists && !nuxtDependency) return;
-  return { mayWantBackend: true };
+  if (!version) throw new Error("Unable to find the nuxt dep.");
+  if (gte(version, "3.0.0")) return { mayWantBackend: true };
```
jamesdaniels, line 23: Let's keep it gte 3.0.0-0 so we capture all the RCs

src/frameworks/nuxt/index.ts
```diff
@@ -1,33 +1,39 @@
...
+  const version = nuxtDependency?.version;
   const configFilesExist = await Promise.all([
     pathExists(join(dir, "nuxt.config.js")),
     pathExists(join(dir, "nuxt.config.ts")),
   ]);
+
   const anyConfigFileExists = configFilesExist.some((it) => it);
   if (!anyConfigFileExists && !nuxtDependency) return;
-  return { mayWantBackend: true };
+  if (!version) throw new Error("Unable to find the nuxt dep.");
```
jamesdaniels, line 22: IMO we can drop these, let it fall back to the can't find framework error

src/frameworks/nuxt2/index.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/**
+ * Get the Nuxt app
+ * @param cwd
+ * @return Nuxt app object
+ */
+async function getNuxtAppForBuild(cwd: string) {
+  let nuxt: any = null;
+  try {
```
jamesdaniels, line 63: IMO we should be able to drop the try/catch

src/frameworks/nuxt2/index.ts
```diff
@@ -0,0 +1,130 @@
...
+  const version = nuxtDependency?.version;
+  const configFilesExist = await Promise.all([
+    pathExists(join(dir, "nuxt.config.js")),
+    pathExists(join(dir, "nuxt.config.ts")),
+  ]);
+  const anyConfigFileExists = configFilesExist.some((it) => it);
+
+  if (!anyConfigFileExists && !nuxtDependency) return;
+  if (!version) throw new Error("Unable to find the nuxt dep.");
+  if (lt(version, "3.0.0")) return { mayWantBackend: true };
```
jamesdaniels, line 23: 3.0.0-0 so it doesn't trip on RCs

src/frameworks/nuxt2/index.ts
```diff
@@ -0,0 +1,130 @@
...
+  const nuxtDependency = findDependency("nuxt", { cwd: dir, depth: 0, omitDev: false });
+  const version = nuxtDependency?.version;
+  const configFilesExist = await Promise.all([
+    pathExists(join(dir, "nuxt.config.js")),
+    pathExists(join(dir, "nuxt.config.ts")),
+  ]);
+  const anyConfigFileExists = configFilesExist.some((it) => it);
+
+  if (!anyConfigFileExists && !nuxtDependency) return;
+  if (!version) throw new Error("Unable to find the nuxt dep.");
```
jamesdaniels, line 22: drop

src/frameworks/nuxt2/index.ts
```diff
@@ -0,0 +1,130 @@
...
+		for the `.nuxt` directory in the destination directory.
+	*/
+  await copy(join(sourceDir, ".nuxt"), join(destDir, ".nuxt"));
+
+  /*
+		When using `SSR: false`, we need to copy the `nuxt.config.js` to the destination directory (`functions`)
+		This is because `loadNuxt` (called from `firebase-frameworks`) will look
+		for the `nuxt.config.js` file in the destination directory.
+		*/
+  if (nuxtConfig.ssr === false) {
```
jamesdaniels, line 124: nit, `!nuxtConfig.ssr`... that said if ssr is false should we skip entirely—by returning wantsBackend: false in build?

src/frameworks/nuxt2/index.ts
```diff
@@ -0,0 +1,130 @@
+import { copy, pathExists } from "fs-extra";
+import { readFile } from "fs/promises";
+import { join } from "path";
+import { lt } from "semver";
+import { findDependency, FrameworkType, relativeRequire, SupportLevel } from "..";
+
+export const name = "Nuxt 2";
```
jamesdaniels, line 7: Can just call nuxt IMO

src/frameworks/nuxt2/index.ts
```diff
@@ -0,0 +1,130 @@
...
+export const support = SupportLevel.Experimental;
+export const type = FrameworkType.Toolchain;
```
jamesdaniels, line 9: Metaframework

src/frameworks/nuxt/index.ts
```diff
@@ -1,33 +1,39 @@
...
-import { basename, join } from "path";
+import { join } from "path";
 import { gte } from "semver";
-import { BuildResult, findDependency, FrameworkType, relativeRequire, SupportLevel } from "..";
+import { findDependency, FrameworkType, relativeRequire, SupportLevel } from "..";
 
-export const name = "Nuxt";
+export const name = "Nuxt 3";
 export const support = SupportLevel.Experimental;
 export const type = FrameworkType.Toolchain;
```
jamesdaniels, line 9: huh, why was this toolchain, should be metaframework

cfofiu: Well, [Toolchain](https://github.com/firebase/firebase-tools/blob/4e013b40a3d4069c3758235544d62d12e8c003d3/src/frameworks/index.ts#L89) seems to map to Vite?

src/frameworks/nuxt2/index.ts
cfofiu: We still need a backend for `loadNuxt` to run

src/frameworks/nuxt2/index.ts
cfofiu: implemented a small refactor; removed the try/catch block

src/frameworks/nuxt2/index.ts
jamesdaniels: if you return `wantsBackend: false` in build it won't try to bundle up a functions directory at all

src/frameworks/nuxt/index.ts
```diff
@@ -1,39 +1,53 @@
...
-  ]);
-  const anyConfigFileExists = configFilesExist.some((it) => it);
+  const nuxtDependency = findDependency("nuxt", {
+    cwd: dir,
+    depth: 0,
+    omitDev: false,
+  }) as NuxtDependency;
+
+  const version = nuxtDependency?.version;
+  const anyConfigFileExists = await nuxtConfigFilesExist(dir);
```
jamesdaniels, line 31: nit, unused?

src/frameworks/nuxt2/index.ts
```diff
@@ -0,0 +1,155 @@
...
+/**
+ * Copy the static files to the destination directory whether it's a static build or server build.
+ * @param root
+ * @param dest
+ */
+export async function ɵcodegenPublicDirectory(root: string, dest: string) {
+  const nuxt = await getNuxtApp(root);
+  const nuxtConfig = await nuxt.loadNuxtConfig();
+  const { ssr, target } = nuxtConfig;
+  /*
```
jamesdaniels, line 105: nit: spacing

jamesdaniels (approved):
LGTM better than what we have. Add changelog and can you fix Austin's email in a force-push?

src/frameworks/nuxt2/index.ts
cfofiu: Comments formatting updated

PR #5325, opened 2022-12-12
Nextjs sharp

jamesdaniels:
Googler mirror of #5238 for integration tests


jamesdaniels:
Don't need a PR turns out, just push to a branch

Issue #5251, opened 2022-11-16
EvaluationException when using crosss-service security rule with Emulator


johnnyoshika:
### [REQUIRED] Environment info

**firebase-tools:** 11.16.1

**Platform:** Windows

### [REQUIRED] Test case

Use cross-service security rule (i.e. `firestore.get()`) in storage.rules:

```
rules_version = '2';
service firebase.storage {
  match /b/{bucket}/o {
    match /courses/{courseId}/files/{fileName} {
      allow get: if firestore.get(/databases/(default)/documents/courses/$(courseId)).data.users[request.auth.uid].exists;
    }
  }
}
```

Add the following document to Emulator's Firestore at `courses/course1`:

```
{
  users: {
    user1: { exists: true }
  }
}
```

### [REQUIRED] Steps to reproduce

Download file from storage using the JavaScript client SDK:

```
    const storage = getStorage();
    const fileRef = ref(storage, 'courses/4VnY71FQeT6XDAGNDjSSX/files/foo.txt');
    const url = await getDownloadURL(fileRef);
    const response = await fetch(url);
```

### [REQUIRED] Expected behavior

Based on this announcement: https://firebase.blog/posts/2022/09/announcing-cross-service-security-rules, I expected cross-service security rules to be supported automatically for firebase-tools@11.10.0 and above.

### [REQUIRED] Actual behavior

`firestore.get()` inside the storage rules results in this exception, so the security rule never allows the request:

```
  com.google.firebase.rules.runtime.common.EvaluationException: Error: C:\Users\...\storage.rules line [5], column [31]. Service call error. Function: [firestore.get], Argument: [path_value {
  segments {
    simple: "databases"
  }
  segments {
    simple: "(default)"
  }
  segments {
    simple: "documents"
  }
  segments {
    simple: "courses"
  }
  segments {
    simple: "4VnY71FQeT6XDAGNDjSSX"
  }
}
].
```

sam-gc:
I'm unable to reproduce this issue. Using your rules and version `11.16.1`, everything works as expected on my end. Just to double check, you're also using the Firebase Auth emulator as well right?

johnnyoshika:
@sam-gc I'll try to reproduce it in a fresh environment. I'll share the repo if I can.

johnnyoshika:
Just to double check, you're also using the Firebase Auth emulator as well right?

>

johnnyoshika:
[...]

PR #5328, opened 2022-12-13
Fix errors in cross-service rules when the Firestore document contains nullValue fields

sam-gc:
Fixes https://github.com/firebase/firebase-tools/issues/5251

The issue is due to a mismatch between what's returned from Firestore and what the Rules emulator expects in Firestore-shaped data


src/emulator/storage/rules/runtime.ts
```diff
@@ -455,6 +457,31 @@ async function fetchFirestoreDocument(
   }
 }
 
+/**
+ * Deeply replaces objects matching {nullValue: null} with {nullValue: 0}.
+ * This corresponds to how Rules emulator handles the null_value type (as an
```
yuchenshi, line 462: This doesn't seem right. To quote the protobuf reference docs:

> The JSON representation for `NullValue` is JSON `null`.

https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#nullvalue

This should be special-cased by the Java protobuf library already. In other words, if the Rules emulator doesn't accept it, that's a bug within the Rules emulator and should be fixed there. Client side fixes like this may mask bigger problems.

yuchenshi (changes requested):

Issue #5354, opened 2022-12-21
Deploying the same function twice does not skip subsequent function deploy


mercury2269:
### [REQUIRED] Environment info

**firebase-tools:** 11.19.0

**Platform:** macOS

### [REQUIRED] Test case

setup multiple configuration values in
```
firebase functions:config:set akey.avalue=1
firebase functions:config:set akey.bvalue=1
firebase functions:config:set akey.cvalue=1
firebase functions:config:set akey.dvalue=1
firebase functions:config:set bkey.avalue=1
firebase functions:config:set bkey.bvalue=1
firebase functions:config:set bkey.cvalue=1
firebase functions:config:set bkey.dvalue=1
```

### [REQUIRED] Steps to reproduce

1. Build your test function
2. Deploy your test function
3. Deploy your test function again

### [REQUIRED] Expected behavior

The second function deploy with the same exact code base, configuration and secrets should be skipped

### [REQUIRED] Actual behavior
The function is deployed again. 

### Discovered issue
Runtime configuration hash is not sorted/deterministic and produces a different hash every time given there are enough configuration objects.

PR #5355, opened 2022-12-21
Bug fix: Cloud Functions Runtime Config Hash Value

mercury2269:
### Description

We've been trying to figure out why in our project functions are re-deployed every time, even when the source code and function configuration doesn't change. As we found out, the runtime configuration produces new hash every time. Because during the plan step the retrieved runtime configuration is not sorted, every time it is pulled it's produces different order of configuration keys. And when the configuration is appended to hash string, it produces a new hash every time.   

Solution: 
Convert runtime configuration object tree into a sorted array.

Please see the following reported issue: 
https://github.com/firebase/firebase-tools/issues/5354

### Scenarios Tested

Re-deploying functions previously deployed with the same code base and configuration.  

cc @TheIronDev


src/deploy/functions/prepareFunctionsUpload.ts
```diff
@@ -88,7 +90,8 @@ async function packageSource(
       });
     }
     if (typeof runtimeConfig !== "undefined") {
-      const runtimeConfigString = JSON.stringify(runtimeConfig, null, 2);
```
TheIronDev, line 91: There is a chance mutating the runtimeConfig string may break function upload behavior.
(@taeold / @inlined can keep me honest here.)

To mitigate that, can you use a separate variable to push into the hash?

```
const runtimeConfigString = JSON.stringify(runtimeConfig, null, 2);
      // In order for hash to be consistent, configuration object tree must be sorted by key, only possible with arrays.
 const runtimeConfigHashString = JSON.stringify(convertToSortedKeyValueArray(runtimeConfig));
hashes.push(runtimeConfigHashString);
```

TheIronDev:
@mercury2269 Thank you for opening this pull request and catching this bug!

Overall the code looks good, I made a comment requesting the config that is added to the archiver be left alone.

Would you be able to add a unit test for this change?

Unfortunately the spec file doesn't exist yet. You would need to create: `src/test/deploy/functions/prepareFunctionsUpload.spec.ts`

 You don't have to write a unit test for the entire file, just the parts that you are making changes to.

src/deploy/functions/prepareFunctionsUpload.ts
taeold: IIUC, this code is mutating the value that we pass into the hash, not the underlying runtime config value.

If I understood correctly, that looks safe!

mercury2269: @TheIronDev @taeold I think it makes sense for the configuration file in the archive to be pretty printed anyways. I went ahead and added a separate formatted string variable. 

Also, as suggested, added a few tests for the sort function. Please have a look @TheIronDev.

taeold (approved):
LGTM. Do you mind adding an entry to the Changelog.md describing the change?

Thank you for your contribution!

mercury2269:
@taeold no problem, added.

alexpchin:
It would be great to merge this 👍 🥇

taeold:
@mercury2269 Do you mind resolving the merge conflict? I think there is a merge conflict on CHANGELOG.md file.

mercury2269:
> @mercury2269 Do you mind resolving the merge conflict? I think there is a merge conflict on CHANGELOG.md file.

done @taeold

pedrofsantoscom:
are there any updates to when this PR can be merged?

taeold (approved):

edenhermelin:
hey @taeold @mercury2269 
I updated to the latest version and looks like the deployment still doesn't skip functions that are deployed with the exact same codebase. I'm wondering if I'm facing the same issue or there is something else.
Do you have a suggestion for how can I investigate further? how can I compare the runtime configs between deployments and make sure they are the same?

Thanks!

mercury2269:
> hey @taeold @mercury2269 I updated to the latest version and looks like the deployment still doesn't skip functions that are deployed with the exact same codebase. I'm wondering if I'm facing the same issue or there is something else. Do you have a suggestion for how can I investigate further? how can I compare the runtime configs between deployments and make sure they are the same?
> 
> Thanks!

Hi @edenhermelin,

The caching feature is working for us now, so it may be an unrelated bug. You can follow contribution readme on how to setup and debug locally [here](https://github.com/firebase/firebase-tools/blob/master/CONTRIBUTING.md)

PR #5360, opened 2022-12-22
Custom emulator auth handler support

grimsteel:
### Description

This adds support for custom auth handlers. A custom auth handler can be set by updating the emulator config through the REST api. It's used when [outputting auth action urls to the terminal](https://firebase.google.com/docs/emulator-suite/connect_auth#emulated_email_email_link_and_anonymous_authentication)

I created a related PR in firebase-tools-ui for adding a way to change this through the emulator UI

### Scenarios Tested

* Developers can update the custom auth handler url through the REST api and Emulator UI
* Setting the handler url to an empty string will once again make the emulator output urls using the default auth handler (typically "localhost:9099/emulator/action")

### Sample Commands

No commands changed

Sample request to update handler URL:
`PATCH http://localhost:9099/emulator/v1/projects/:targetProjectId/config`
body:
```json
{
  "notification": {
    "sendEmail": {
      "callbackUri": "https://example.com/auth/handler"
    }
  }
}
```

### DIFF
--- a/scripts/gen-auth-api-spec.ts
+++ b/scripts/gen-auth-api-spec.ts
@@ -456,6 +456,17 @@ function addEmulatorOperations(openapi3: any): void {
         },
         type: "object",
       },
+      notification: {
+        properties: {
+          sendEmail: {
+            properties: { 
+              callbackUri: { type: "string" },
+            },
+            type: "object",
+          },
+        },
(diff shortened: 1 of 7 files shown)


grimsteel:
Could somebody approve the workflow runs?

PR #5376, opened 2022-12-28
fix: prevent lint from failing

stfsy:
### Description
At least on my Windows macine, linting fails because the RecursiveElement type is unused. I move the alraedy existing `eslint disable ..` comment one line up to prevent linting from failing.

### Scenarios Tested
Ran lint command on WIndows machine.


stfsy:
@taeold IDK maybe it is only on my machine. Apparently linting in GH workflows works as expected. Your call.

PR #5378, opened 2022-12-29
Fix.

tinkersprojects:
from this issue:
[https://github.com/firebase/firebase-tools/issues/5369](https://github.com/firebase/firebase-tools/issues/5369
)
Its trying to run `npm ls --omit=dev --all --json` in a command line but the output is tooo large for the buffer. this makes the JSON parse fail.

this fix will make it read from where is getting its data from and not from a limited buffer

### DIFF
--- a/src/frameworks/next/index.ts
+++ b/src/frameworks/next/index.ts
@@ -348,11 +348,17 @@ export async function ɵcodegenFunctionsDirectory(sourceDir: string, destDir: st
     // macs and older Node versions; either way, we should avoid taking on any deps in firebase-tools
     // Alternatively I tried using @swc/spack and the webpack bundled into Next.js but was
     // encountering difficulties with both of those
-    const dependencyTree: NpmLsReturn = JSON.parse(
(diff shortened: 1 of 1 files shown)


jamesdaniels:
Thanks for identifying the root cause @tinkersprojects, I took a different approach in #5393. Would you mind giving that a spin? I'm streaming the npm ls stdout to a stream-json chain and ignoring everything but the keys... hopefully that should work without resorting to reading lockfiles

NiedziolkaMichal:
This solution doesn't work with my project, which uses yarn instead of npm and therefore doesn't have `package-lock.json`.

jamesdaniels:
@NiedziolkaMichal @tinkersprojects since I don't have a repro would you mind emailing me your package.json & lockfiles? jamesdaniels@google.com Once I have a repro I'm sure I'll be able to find a workable solution for both NPM, yarn

NiedziolkaMichal:
[package+lock.zip](https://github.com/firebase/firebase-tools/files/10385425/package%2Block.zip)

jamesdaniels:
Addressing in #5499

PR #5389, opened 2023-01-03
upgrade superstatic

### LINKED ISSUE #4879: Hosting i18n rewrites for 404.html not supported by emulator
Running `firebase emulators:start` in local folder, with i18n rewrites, following the example found in this document: https://firebase.google.com/docs/hosting/i18n-rewrites#priority-order-example

Browsing to `localhost:5000/`
Changed cookie to `fr` -> displays the correct `fr/index.html`

Browsing to `localhost:5000/awesome-page.html`
**Expected behavior**
I expect the version under `/fr/404.html` to be served.
Instead, only the root 404.html is served.

- OS: Windows 11
- Browser Chrome
- Version 11.6.0

Note: I did not try to deploy to see the behavior live

### LINKED ISSUE #5133: Hosting Emulator: Geting 404 error page instead of index.html
I have recently updated firebase tools by `npm update -g firebase-tools`
and the expected hosting emulator behavior has changed.
I am getting 404 page instead of index.html.

**Expected behavior**

when accessing `localhost:5000` I should get `localhost:5000/index.html`
but i am getting `404` not found page
also I am getting `localhost:5000/index.html` when using url `localhost:5000/index.html`.

**To Reproduce**

I have tried installing emulator suite from scratch on another machine
and still getting 404 not found page instead of index.html
so this bug may have been introduced in new versions of firebase-tools

OS: windows
Browser :n/a
firebase-tools Version :11.14.4

bkendall:
### Description

Upgrades superstatic to 9.0.1. Reintroduces the i18n fix and the Windows paths fix.

Fixes #4879 
Fixes #5133

### Scenarios Tested

- Local Hosting site with basic testing


bkendall:
My i18n test didn't work, so I'm checking in on it.

src/serve/hosting.ts
```diff
@@ -52,13 +52,13 @@ function startServer(options: any, config: any, port: number, init: TemplateServ
   const server = superstatic({
     debug: false,
     port: port,
-    host: options.host,
+    hostname: options.host,
     config: config,
     compression: true,
-    cwd: detectProjectRoot(options),
+    cwd: detectProjectRoot(options) || undefined,
```
joehan, line 58: Little confused why this `|| undefined` is needed - is this meant to cast falsey values to undefined?

bkendall: Yeah, detectProjectRoot returns `string | null`, which is just irritating. This is why I don't use `null`.

joehan (approved):

PR #5396, opened 2023-01-05
Add database:import command for non-atomic import

tohhsinpei:
### Description
[b/261669787](http://b/261669787)
API proposal: [go/firebase-database-import](http://goto.google.com/firebase-database-import)

Implementation:
- Take JSON file input and stream top-level objects using `JSONStream`. Assume individual objects can fit in memory.
- For each JSON object, split into 1 MB chunks (partially based on [firebase-import](https://github.com/FirebaseExtended/firebase-import/blob/master/src/firebase-import.js)). PUT chunks in parallel.


src/databaseChunkUploader.ts
```diff
@@ -0,0 +1,77 @@
...
+  constructor(private dbUrl: URL, file: string) {
+    this.client = new Client({ urlPrefix: dbUrl.origin, auth: true });
+
+    const data = { json: JSON.parse(file), pathname: dbUrl.pathname };
+    const chunkedData = this.chunkData(data);
+    this.chunks = chunkedData.chunks || [data];
+  }
+
+  public async upload(overwrite: boolean): Promise<any> {
+    return Promise.all(
```
tohhsinpei, line 30: Maybe set a concurrency limit here?

src/commands/database-import.ts
```diff
@@ -0,0 +1,89 @@
...
+      importer = new DatabaseImporter(dbUrl, inputString);
+    } catch (err: any) {
+      throw new FirebaseError("Invalid data; couldn't parse JSON object, array, or value.", {
+        original: err,
+        exit: 2,
+      });
+    }
+
+    try {
+      await importer.execute();
```
fredzqm, line 77: We want to make sure the path is empty first.

Maybe issue an shallow query and throw an error to suggest use `database:remove --disable-triggers` first.

Think about the use case a bit more. We should only allow importing to an empty location. Overwriting existing data is likely due to wrong import path.
Our internal restore tool only allow restoring to an empty database by default. I think ask them to delete the path is reasonable.

src/database/import.ts
```diff
@@ -0,0 +1,80 @@
...
+    const chunkedData = this.chunkData(data);
+    this.chunks = chunkedData.chunks || [data];
+    this.client = new Client({ urlPrefix: dbUrl.origin, auth: true });
+  }
+
+  /**
+   * Writes the chunked data to RTDB. Any existing data at the specified location will be overwritten.
+   */
+  async execute(): Promise<any> {
+    return Promise.all(
```
fredzqm, line 36: Yeah, we definitely want to limit the concurrency. Maybe a configurable flag that defaults to 5?

Not sure, can discuss more in the API doc.

fredzqm (changes requested):
Love your implementation. Really concise and clean.
It's good to start with for now.

Using `PUT` allows us to override one path at a time.
However, consider a case with a large list of small elements. (The total list exceeds 1MB) This would create tons of splits.
If we multi-path `PATCH`, we can batch many small path and have fewer tiny chunks.

This is an optional optimization.

src/commands/database-import.ts
tohhsinpei: Done, did it inside `DatabaseImporter` though.

src/database/import.ts
tohhsinpei: Done. But not configurable for now. Not sure that it's necessary.

src/database/import.ts
```diff
@@ -0,0 +1,114 @@
...
+ */
+export default class DatabaseImporter {
+  chunks: Data[];
+  private client: Client;
+  private limit = pLimit(CONCURRENCY_LIMIT);
+
+  constructor(private dbUrl: URL, file: string, private chunkSize = MAX_CHUNK_SIZE) {
+    let data;
+    try {
+      data = { json: JSON.parse(file), pathname: dbUrl.pathname };
```
lepatryk, line 33: I'd like this command to support at least 100GB json files. So parsing everything to memory at once isn't really practical. https://github.com/FirebaseExtended/firebase-import uses JSONStream. Can we use it here as well?

tohhsinpei: Done. Although, note that
1) `firebase-import` still loads the whole file into memory, not sure why, we'll not do that here
2) `JSONStream` assumes that the top-level objects can fit into memory but that the entire file should be streamed. I think this is a reasonable, maybe even necessary assumption.

src/commands/database-import.ts
```diff
@@ -0,0 +1,89 @@
...
+      },
+      options
+    );
+    if (!confirm) {
+      throw new FirebaseError("Command aborted.");
+    }
+
+    const inputString =
+      options.data ||
+      (await utils.streamToString(infile ? fs.createReadStream(infile) : process.stdin));
```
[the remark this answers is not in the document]
tohhsinpei: For now I have `JSONStream` parse and stream the top-level objects. We assume that individual objects can fit in memory.

tohhsinpei:
@fredzqm TBC on doing multi-path PATCH for now. If we do, we'll need to do something else to make sure that each batch of updates also doesn't exceed the request payload size. But shouldn't be hard.

fredzqm:
Multi-path PATCH is optional.

Singe-path PUT would deliver a lot of value already, so I'm happy to get that algorithm in and iterate later.

I would leave whether to use multi-path PATCH in the first version up to you. @tohhsinpei

src/database/import.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  private readAndWriteChunks(inStream: NodeJS.ReadableStream): Promise<any> {
+    const { dbUrl } = this;
+    const chunkData = this.chunkData.bind(this);
+    const writeChunk = this.writeChunk.bind(this);
+    const getJoinedPath = this.joinPath.bind(this);
+
+    const readChunks = new stream.Transform({ objectMode: true });
+    readChunks._transform = function (chunk: { key: string; value: any }, _, done) {
```
fredzqm, line 76: Very cool. I am reading its docs: https://nodejs.org/api/stream.html#implementing-a-readable-stream

Thinking about how to control the size of the `chunk` emitted.
Shall we set `highWaterMark` higher to have larger chunk size?

fredzqm (changes requested):

src/commands/database-import.ts
```diff
@@ -0,0 +1,96 @@
...
+  .before(requireDatabaseInstance)
+  .before(populateInstanceDetails)
+  .before(printNoticeIfEmulated, Emulators.DATABASE)
+  .action(async (path: string, infile, options) => {
+    if (!path.startsWith("/")) {
+      throw new FirebaseError("Path must begin with /");
+    }
+
+    if (options.only && !options.only.startsWith("/")) {
+      throw new FirebaseError("Data path must begin with /");
```
joehan, line 43: This error should mention the name of the flag that is incorrect

src/commands/database-import.ts
```diff
@@ -0,0 +1,96 @@
...
+  )
+  .option(
+    "--only <dataPath>",
+    "import only data at this path in the JSON file (if omitted, import entire file)"
+  )
+  .before(requirePermissions, ["firebasedatabase.instances.update"])
+  .before(requireDatabaseInstance)
+  .before(populateInstanceDetails)
+  .before(printNoticeIfEmulated, Emulators.DATABASE)
+  .action(async (path: string, infile, options) => {
```
joehan, line 37: Please add a type for infile and options. 

For options, I suggest extending the type defined in options.ts with any command  specific fields (like we do for hosting in hosting/options.ts)

This will fix the 'unsafe member access' linter issues that are all over this file.

joehan, line 37: Note: We support a global flag --json, which makes a command output whatever is returned by the action function as JSON. To that end, it would be nice to include a return type for the function in action.

Right now, this is Promise<void>, which is fine. However, think about if there is anything that this should return that would be useful for CI/automated use cases.

src/database/import.ts
```diff
@@ -0,0 +1,166 @@
...
+        )
+        .pipe(readChunks)
+        .pipe(writeChunks)
+        .on("data", (res: any) => responses.push(res))
+        .on("error", reject)
+        .once("end", () => resolve(responses));
+    });
+  }
+
+  private writeChunk(chunk: Data): Promise<any> {
```
joehan, line 111: Replace `any` with a real return type

src/database/import.ts
```diff
@@ -0,0 +1,166 @@
...
+      done();
+    };
+
+    return new Promise((resolve, reject) => {
+      const responses: any[] = [];
+      inStream
+        .pipe(JSONStream.parse(this.jsonPath))
+        .on("error", (err: any) =>
+          reject(
+            new FirebaseError("Invalid data; couldn't parse JSON object, array, or value.", {
```
joehan, line 97: Could you test this error case to see how it appears to users? I want to make sure the original error is visible to them without the --debug flag

src/test/database/import.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  it("chunks data in top-level objects", async () => {
+    nock("https://test-db.firebaseio.com").get("/foo.json?shallow=true").reply(200);
+    nock("https://test-db.firebaseio.com").put("/foo/a.json", "100").reply(200);
+    nock("https://test-db.firebaseio.com")
+      .put("/foo/b.json", JSON.stringify([true, "bar", { f: { g: 0, h: 1 }, i: "baz" }]))
+      .reply(200);
+
+    const importer = new DatabaseImporter(dbUrl, DATA_STREAM, /* importPath= */ "/");
+    const responses = await importer.execute();
```
joehan, line 41: nit - prefer newlines between the arrange, act, and assert parts of the test

package.json
```diff
@@ -120,6 +120,7 @@
     "google-auth-library": "^7.11.0",
     "inquirer": "^8.2.0",
     "js-yaml": "^3.13.1",
+    "JSONStream": "^1.3.5",
```
joehan, line 123: We should not add this import. It is a small, deprecated utility that hasn't been maintained for 4 years, (https://github.com/dominictarr/JSONStream/blob/master/index.js#L17). Use the already imported json-stream instead.

Take a look at https://github.com/firebase/firebase-tools/pull/4027 for an example of how to replace JSONstream with stream-json

joehan (changes requested):

src/commands/database-import.ts
```diff
@@ -0,0 +1,96 @@
...
+import { logger } from "../logger";
+import { printNoticeIfEmulated } from "../emulator/commandUtils";
+import { promptOnce } from "../prompt";
+import { populateInstanceDetails } from "../management/database";
+import { realtimeOriginOrEmulatorOrCustomUrl } from "../database/api";
+import { requireDatabaseInstance } from "../requireDatabaseInstance";
+import { requirePermissions } from "../requirePermissions";
+
+export const command = new Command("database:import <path> [infile]")
+  .description("non-atomically import JSON file to the specified path")
```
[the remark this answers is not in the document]
tohhsinpei: Done.

src/commands/database-import.ts
```diff
@@ -0,0 +1,96 @@
...
+    "--instance <instance>",
+    "use the database <instance>.firebaseio.com (if omitted, use default database instance)"
+  )
+  .option(
+    "--disable-triggers",
+    "suppress any Cloud functions triggered by this operation, default to true",
+    true
+  )
+  .option(
+    "--only <dataPath>",
```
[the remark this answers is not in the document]
tohhsinpei: Going with `--filter`.

src/database/import.ts
tohhsinpei: Done.

src/commands/database-import.ts
tohhsinpei: Done.

src/commands/database-import.ts
```diff
@@ -0,0 +1,96 @@
...
+import { printNoticeIfEmulated } from "../emulator/commandUtils";
+import { promptOnce } from "../prompt";
+import { populateInstanceDetails } from "../management/database";
+import { realtimeOriginOrEmulatorOrCustomUrl } from "../database/api";
+import { requireDatabaseInstance } from "../requireDatabaseInstance";
+import { requirePermissions } from "../requirePermissions";
+
+export const command = new Command("database:import <path> [infile]")
+  .description("non-atomically import JSON file to the specified path")
+  .option("-f, --force", "pass this option to bypass confirmation prompt")
```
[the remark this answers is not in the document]
tohhsinpei: Done.

CHANGELOG.md
```diff
@@ -1 +1,2 @@
 - Fixes an issue where `ext:dev:init` would fail due to a missing CHANGELOG.md file (#5530).
+- Adds `database:import` command for non-atomic imports (#5396).
```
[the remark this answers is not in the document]
tohhsinpei: No public doc yet.

src/database/import.ts
tohhsinpei: Fixed:
```
Error: Invalid data; couldn't parse JSON object, array, or value. Unexpected COLON(":") in state COMMA
```

package.json
tohhsinpei: Done.

src/commands/database-import.ts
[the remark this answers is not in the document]
tohhsinpei: `--force` just bypasses the `? You are about to import data to https://<instance>.firebaseio.com/. Are you sure? (y/N)` prompt. We still check if the imported path is empty and tell the user to delete if not.

src/database/import.ts
```diff
@@ -0,0 +1,166 @@
...
+  ) {
+    this.jsonPath = this.computeJsonPath(dataPath);
+    this.client = new Client({ urlPrefix: dbUrl.origin, auth: true });
+  }
+
+  /**
+   * Writes the chunked data to RTDB. Any existing data at the specified location will be overwritten.
+   */
+  async execute(): Promise<any> {
+    await this.checkLocationIsEmpty();
```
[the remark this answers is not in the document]
tohhsinpei: See above.

src/database/import.ts
tohhsinpei: Not sure. We're in `objectMode`, so `highWaterMark` defaults to 16 (objects, not bytes). It's hard to know how many objects we should buffer.

src/commands/database-import.ts
tohhsinpei: Maybe total number of bytes written. Would be a good follow up. I'd prefer to get this PR in first if that's ok.

src/database/import.ts
fredzqm: Can you experiment with a large file and log each request's size?

If it ends up sending one request for lots of tiny locations, it would be very slow.
We want to make sure the batch size is reasonable large for this command to be efficient and useful.

fredzqm (approved):

fredzqm:
Impl LGTM. Curious about the actual performance of it.

src/database/import.ts
```diff
@@ -0,0 +1,172 @@
...
+    private dataPath: string,
+    private chunkSize = MAX_CHUNK_SIZE
+  ) {
+    this.client = new Client({ urlPrefix: dbUrl.origin, auth: true });
+  }
+
+  /**
+   * Writes the chunked data to RTDB. Any existing data at the specified location will be overwritten.
+   */
+  async execute(): Promise<any> {
```
joehan, line 47: Could you take an another pass through this file and try to clean up all `any`s that are not absolutely necessary? It seems like we should be able to add a real return type for writeChunk since we own the API it is calling & should know what to expect back. Once we add that, we should be able to propogate the real type up through the call stack too.

If there is somewhere in here where we can't know the type, we should use `unknown` instead of `any`.

src/database/import.ts
```diff
@@ -0,0 +1,172 @@
...
+      // Leaf node, cannot be chunked
+      return { chunks: null, size: JSON.stringify(json).length };
+    } else {
+      // Children node
+      let size = 2; // {}
+
+      const chunks = [];
+      let hasChunkedChild = false;
+
+      for (const key of Object.keys(json)) {
```
joehan, line 143: Optional - since you are accessing json[key] inside this loop, it would be slightly cleaner imo to use Object.entries instead 

```
for (const [key, val] of Object.entries(json))
```

joehan (commented):
If possible, please minimize `any`s in this code.

src/database/import.ts
tohhsinpei: Done.

src/database/import.ts
tohhsinpei: Done.

src/database/import.ts
```diff
@@ -0,0 +1,172 @@
...
+    const chunkData = this.chunkData.bind(this);
+    const writeChunk = this.writeChunk.bind(this);
+    const getJoinedPath = this.joinPath.bind(this);
+
+    const readChunks = new stream.Transform({ objectMode: true });
+    readChunks._transform = function (chunk: { key: string; value: any }, _, done) {
+      const data = { json: chunk.value, pathname: getJoinedPath(dbUrl.pathname, chunk.key) };
+      const chunkedData = chunkData(data);
+      const chunks = chunkedData.chunks || [data];
+      chunks.forEach((chunk: Data) => this.push(chunk));
```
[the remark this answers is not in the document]
tohhsinpei: Done.

tohhsinpei:
> > Can you experiment with a large file and log each request's size?
> 
> If it ends up sending lots of tiny requests, it would be very slow.
> We want to make sure the batch size is reasonable large for this command to be efficient and useful.

I just increased `MAX_CHUNK_SIZE` to 10 MB. So it will be able to write objects of up to 10 MB in one go. If the file is a lot of small objects at the top level then this won't help, though.

joehan (approved):

PR #5413, opened 2023-01-11
Don't use Next.js internal redirects for the backend test

jamesdaniels:


jamesdaniels:
Turns out trailingSlashes is the culprit here, going to punt to #5445

PR #5419, opened 2023-01-12
 Remove go discovery module for CF3

taeold:
Maybe one day we will go back and re-invest here. For now let's remove the code to avoid extra effort to keep these modules up-to-date as we extend capabilities of runtime delegate module.

PR #5422, opened 2023-01-13
Refactor how node binary is retrieved for functions/extensions.

taeold:
Today, Functions Emulator make strong assumption that all functions target node runtime and pass around references like `nodeBinary`, `nodeMajorVersion`, etc.

This PR refactors the logic to abstract away `node` into `bin` and `runtime`. The logic for choosing the appropriate `bin` for the function is lifted to `RuntimeDelegate`, same piece of code that holds together other `nodejs` runtime specific logic. 

The refactored required small changes to the extensions emulator - looping in @joehan to see if he has any objections.

### DIFF
--- a/src/deploy/functions/runtimes/node/index.ts
+++ b/src/deploy/functions/runtimes/node/index.ts
@@ -66,7 +66,6 @@ export class Delegate {
   // to decide whether to use the JS export method of discovery or the HTTP container contract
   // method of discovery.
   _sdkVersion: string | undefined = undefined;
-
   get sdkVersion(): string {
     if (this._sdkVersion === undefined) {
       this._sdkVersion = versioning.getFunctionsSDKVersion(this.sourceDir) || "";
--- a/src/deploy/functions/runtimes/node/index.ts
+++ b/src/deploy/functions/runtimes/node/index.ts
@@ -103,35 +103,34 @@ export class Delegate {
       }
     }
 
-    // If the requested version is the same as the host, let's use that
     if (semver.major(requestedVersion) === semver.major(hostVersion)) {
       logLabeledSuccess("functions", `Using node@${semver.major(hostVersion)} from host.`);
-    } else {
-      // Otherwise we'll warn and use the version that is currently running this process.
-      if (process.env.FIREPIT_VERSION) {
-        logLabeledWarning(
-          "functions",
-          `You've requested "node" version "${semver.major(
-            requestedVersion
-          )}", but the standalone Firebase CLI comes with bundled Node "${semver.major(
-            hostVersion
-          )}".`
-        );
-        logLabeledSuccess(
-          "functions",
-          `To use a different Node.js version, consider removing the standalone Firebase CLI and switching to "firebase-tools" on npm.`
-        );
-      } else {
-        logLabeledWarning(
-          "functions",
-          `Your requested "node" version "${semver.major(
-            requestedVersion
-          )}" doesn't match your global version "${semver.major(
-            hostVersion
-          )}". Using node@${semver.major(hostVersion)} from host.`
-        );
-      }
+      return process.execPath;
     }
+
+    if (!process.env.FIREPIT_VERSION) {
+      logLabeledWarning(
+        "functions",
+        `Your requested "node" version "${semver.major(
+          requestedVersion
+        )}" doesn't match your global version "${semver.major(
+          hostVersion
+        )}". Using node@${semver.major(hostVersion)} from host.`
+      );
+      return process.execPath;
+    }
+
+    // Otherwise we'll warn and use the version that is currently running this process.
+    logLabeledWarning(
+      "functions",
+      `You've requested "node" version "${semver.major(
+        requestedVersion
+      )}", but the standalone Firebase CLI comes with bundled Node "${semver.major(hostVersion)}".`
+    );
+    logLabeledSuccess(
+      "functions",
+      `To use a different Node.js version, consider removing the standalone Firebase CLI and switching to "firebase-tools" on npm.`
+    );
     return process.execPath;
   }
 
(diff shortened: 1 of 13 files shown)


taeold:
* master
  * **PR #5419** <a href="https://app.graphite.dev/github/pr/firebase/firebase-tools/5419" target="_blank">Graphite</a> 
    * **PR #5422** <a href="https://app.graphite.dev/github/pr/firebase/firebase-tools/5422" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/firebase/firebase-tools/5422?utm_source=stack-comment).

src/deploy/functions/runtimes/node/index.ts
```diff
@@ -66,13 +66,75 @@ export class Delegate {
...
   _sdkVersion: string | undefined = undefined;
-  get sdkVersion() {
+  get sdkVersion(): string {
     if (this._sdkVersion === undefined) {
       this._sdkVersion = versioning.getFunctionsSDKVersion(this.sourceDir) || "";
     }
     return this._sdkVersion;
   }
 
+  _bin = "";
```
TheIronDev, line 76: Can you write some docs on this (+ other functions) ?

src/deploy/functions/runtimes/node/index.ts
```diff
@@ -66,13 +66,75 @@ export class Delegate {
...
+        logLabeledSuccess(
+          "functions",
+          `Using node@${semver.major(localNodeVersion)} from local cache.`
+        );
+        return localNodePath;
+      }
+    }
+
+    // If the requested version is the same as the host, let's use that
+    if (semver.major(requestedVersion) === semver.major(hostVersion)) {
```
TheIronDev, line 107: Nit: reduce if-else stack:

```
if (semver.major(requestedVersion) === semver.major(hostVersion)) {
    logLabeledSuccess("functions", `Using node@${semver.major(hostVersion)} from host.`);
    return process.execPath;
}

if (!process.env.FIREPIT_VERSION) {
  logLabeledWarning(
          "functions",
          `Your requested "node" version "${semver.major(
            requestedVersion
          )}" doesn't match your global version "${semver.major(
            hostVersion
          )}". Using node@${semver.major(hostVersion)} from host.`
        );
    return process.execPath;
}


// Otherwise we'll warn and use the version that is currently running this process.
logLabeledWarning(
  "functions",
  `You've requested "node" version "${semver.major(
    requestedVersion
  )}", but the standalone Firebase CLI comes with bundled Node "${semver.major(
    hostVersion
  )}".`
);
logLabeledSuccess(
  "functions",
  `To use a different Node.js version, consider removing the standalone Firebase CLI and switching to "firebase-tools" on npm.`
);
return process.execPath;
```

TheIronDev (approved):

TheIronDev:
Can you write tests for these new functions too?

src/emulator/extensionsEmulator.ts
```diff
@@ -230,15 +230,16 @@ export class ExtensionsEmulator implements EmulatorInstance {
     const functionsDir = path.join(extensionDir, "functions");
     // TODO(b/213335255): For local extensions, this should include extensionSpec instead of extensionVersion
     const env = Object.assign(this.autoPopulatedParams(instance), instance.params);
-    const { extensionTriggers, nodeMajorVersion, nonSecretEnv, secretEnvVariables } =
+    const { extensionTriggers, runtime, nonSecretEnv, secretEnvVariables } =
       await getExtensionFunctionInfo(instance, env);
     const emulatableBackend: EmulatableBackend = {
       functionsDir,
+      runtime,
+      bin: process.execPath,
```
joehan, line 238: IIUC these changes correctly, this means we'll always use the version of node that is executing the CLI, and will effectively ignore the runtime. Is that intended? Does the code work without setting `bin` here?

joehan (commented):
This change makes a lot of sense to me and mostly looks good. However, I get a little worried whenever we start messing with paths in the emulator.

Could you test the following cases:
- Emulate an extension, confirm it executes
- Emulate an extension, confirm it uses the right version of node (even when different from the system version)

taeold:
@joehan 

> Could you test the following cases:
Emulate an extension, confirm it executes

Wait doesn't the extensions emulator exercise this?

src/deploy/functions/runtimes/node/index.ts
taeold: Most of these fields are documented in the abstract class. I'd prefer if we don't duplicate the comments?

https://github.com/firebase/firebase-tools/blob/e7d25e292746132fd6084b027fec010c1d912a1a/src/deploy/functions/runtimes/index.ts#L65-L68

joehan:
@taeold Totally forgot we had that test running on CI - it definitely does cover that case

src/deploy/functions/runtimes/node/index.ts
taeold: I like it! ❤️

src/extensions/emulator/specHelper.ts
```diff
@@ -94,27 +93,34 @@ export function getFunctionProperties(resources: Resource[]) {
...
-
   if (invalidRuntimes.length) {
     throw new FirebaseError(
       `The following runtimes are not supported by the Emulator Suite: ${invalidRuntimes.join(
         ", "
       )}. \n Only Node runtimes are supported.`
     );
   }
-  return Math.max(...versions);
+  return runtimes[0];
```
joehan, line 125: Nit: I'd prefer to keep this behavior of returning the highest version of whatever type of runtime we are looping through, in case we ever do switch to using the right version of node for each backend.

joehan (approved):

PR #5440, opened 2023-01-18
add `file:` dependency support for web frameworks

austincrim:
### Description

This PR [re-introduces](https://github.com/FirebaseExtended/firebase-framework-tools/blob/a1eb7c9167781fe35fe6b3910e558caf035b49b1/src/frameworks/index.ts#L87) the ability to bundle and deploy a [local dependency](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#local-paths) to Firebase Functions (when using web frameworks) by specifying the dependency version as `file:./some-path`.

If the specified path is a folder, it will run `npm pack` and copy the tarball into the deployed `functions` folder. If the specified path is a file, it will copy it directly to the `functions` folder.


src/frameworks/index.ts
```diff
@@ -503,6 +503,27 @@ export async function prepareFrameworks(
       packageJson.engines ||= {};
       packageJson.engines.node ||= NODE_VERSION;
 
+      for (const [name, version] of Object.entries(
+        packageJson.dependencies as Record<string, string>
+      )) {
+        if (version.startsWith("file:")) {
+          const path = version.split(":")[1];
```
jamesdaniels, line 510: let's get more strict here, `version.replace(/^file:/, '')`, we probably need to be smarter about relative paths too

jamesdaniels (changes requested):

src/frameworks/index.ts
austincrim: @jamesdaniels what specific concerns do you have about relative paths?

jamesdaniels: if it's relative, e.g, `file:../foo` we probably want to resolve it relative to the source directory rather than the cwd, which i believe is what the default is

jamesdaniels (approved):
Look into relative dirs and add a changelog, otherwise LGTM

src/frameworks/index.ts
austincrim: in my initial testing, relative paths seem to be working as intended without any modifications. there might be some scenarios I missed but things like `file:../foo.js` and `file:../test_dir` were resolved and copied/packed correctly.

PR #5443, opened 2023-01-20
Explicitly check for channel existence before creating it

pavelgj:
calling createChannel on an existing channel with existing triggers can result in 400 error

### DIFF
--- a/src/deploy/functions/release/fabricator.ts
+++ b/src/deploy/functions/release/fabricator.ts
@@ -316,7 +316,7 @@ export class Fabricator {
       await this.executor
         .run(async () => {
           try {
-            if (await eventarc.getChannel(channel) !== undefined) {
+            if ((await eventarc.getChannel(channel)) !== undefined) {
               return;
             }
             const op: { name: string } = await eventarc.createChannel({ name: channel });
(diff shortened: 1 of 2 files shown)


src/deploy/functions/release/fabricator.ts
```diff
@@ -316,6 +316,9 @@ export class Fabricator {
       await this.executor
         .run(async () => {
           try {
+            if ((await eventarc.getChannel(channel)) !== undefined) {
+              return;
+            }
```
taeold, line 321: In L330, We explicitly check for error code 409 to see if the channel already exists. Are we saying 409 is the wrong code?

pavelgj: createChannel will sometimes return 400 if the trigger exists and has triggers (which is common if you already functions listening to events on that channel), and 409 if exists and no triggers...

taeold: :(. Can we add comment on why we can't rely on 409? Future me might accidentally remove this condition thinking that 409 check is sufficient.

taeold (approved):

src/deploy/functions/release/fabricator.ts
pavelgj: added a comment and rebased on master

dreadjr:
@pavelgj Thank you!! Any idea when this will get released?

PR #5445, opened 2023-01-20
Pass `trailingSlash` from Next.js config to firebase.json

leoortizz:


src/frameworks/next/index.ts
```diff
@@ -95,7 +95,11 @@ export async function build(dir: string): Promise<BuildResult> {
   });
 
   const reasonsForBackend = [];
-  const { distDir } = await getConfig(dir);
+  const {
+    distDir,
+    // trailingSlash defaults to false in Next.js: https://nextjs.org/docs/api-reference/next.config.js/trailing-slash
+    trailingSlash = false,
```
jamesdaniels, line 101: can we push the default into getConfig?

leoortizz: I tried using the third parameter in `loadConfig` but ends up that it [overrides the user config](https://github.com/vercel/next.js/blob/440560fbbb26d06233b41738c1e99e246291e7e7/packages/next/src/server/config.ts#L878). I'll dig into it a little bit more

jamesdaniels: I'm thinking here https://github.com/firebase/firebase-tools/blob/e02f4fd1f13d68df66551ff6e014327c34b18e5a/src/frameworks/next/index.ts#L451

leoortizz: gotcha, the same way as `distDir`.. I missed it. Just pushed it

jamesdaniels (approved):
LGTM, add changelog

PR #5417, opened 2023-01-12
Add Astro framework

Geoffrey-Pliez:
### Description

added support for astro framework

### Scenarios Tested

Tested with astro config to output: server and adapter: nodejs (middleware mode)

PR #5469, opened 2023-01-30
Fix valid choice of node

Geoffrey-Pliez:
### Description
In dev mode, if the local version of node > 16,  ssr functions does not start and the following message is displayed : 

!!  functions: Failed to load function definition from source: FirebaseError: package.json in functions directory has an engines field which is unsupported. Valid choices are: {"node": 10|12|14|16}

In order not to receive the message and to be able to modify the version number, we retrieve the version in the package.json file at the root of the project

### Scenarios Tested

firebase emulators:start with web framework and local node > 16


austincrim:
Hey @Geoffrey-Pliez, thanks for the contribution.

If I'm understanding it correctly, you're seeing errors when running the Functions emulator and using a version of Node higher than 16 (e.g. 17 or 18)? I don't get any errors when doing so locally. 

If you're still hitting this, could you provide more detailed reproduction instructions I can try, then we can tackle to fix after confirming it's still an issue.

Geoffrey-Pliez:
Hi @austincrim,

Yes that's fine.

when firebase creates package.json file in .firebase/{site}/functions folder for SSR mode and locally, it gets local value from node via [prepareFramework] (https://github.com/firebase/firebase-tools/ blob/17de2ab9efc85a9ff1fc637fa673ed58e06be874/src/frameworks/index.ts#L111) and then [generates](https://github.com/firebase/firebase-tools/blob/17de2ab9efc85a9ff1fc637fa673ed58e06be874/src/frameworks/index.ts#L512) the package.json file. If this version is greater than 16, when launching the "SSR" function, the error message Valid choices are: {"node": 10|12|14|16} appears and as a result, the function emulator does not not create the ssr function.
So the local emulator does not work like the production server and does not allow to exploit the __session cookie.

In this case, my temporary solution is to grab the project's package.json with the desired engine value on the prod side.

But another solution would be not to check the version of node when launching the function when you are in dev mode.

I hope my explanation was clear enough

austincrim:
@Geoffrey-Pliez sounds like I'm understanding the problem correctly, but I'm still unable to reproduce it.

Have you tried with the latest version of `firebase-tools`? If it's still occurring then, could you create and share a minimal reproduction repository?

Geoffrey-Pliez:
@austincrim 
here is a [repository](https://github.com/Geoffrey-Pliez/firebase-tools-debug), you have to create a project on firebase to test it.

For the problem to occur, the 2 conditions of the code [here](https://github.com/firebase/firebase-tools/blob/17de2ab9efc85a9ff1fc637fa673ed58e06be874/src/frameworks/index.ts#L405) must be true.

Here is the log:
Detected at Astro codebase. This is an experimental integration, proceed with caution.

Debugger attached.
    astro v2.0.11 started in 48ms
  
   ┃ Local http://localhost:3000/
   ┃ Network use --host to expose
  
Debugger attached.
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE package: 'firebase-frameworks@0.6.2',
npm WARN EBADENGINE required: { node: '^16.0.0 || ^18.0.0' },
npm WARN EBADENGINE current: { node: 'v19.1.0', npm: '9.4.2' }
npm WARN EBADENGINE }

up to date in 1s

22 packages are looking for funding
   run `npm fund` for details
Waiting for the debugger to disconnect...
09:45:32 [astro] update C:/Users/gunbl/Dev/firebase-tools-debug/.firebase/debug-7611f/functions/package-lock.json
! functions: The following emulators are not running, calls to these services from the Functions emulator will affect production: auth, firestore, database, pubsub, storage
i hosting[debug-7611f]: Serving hosting files from: dist\client
+ hosting[debug-7611f]: Local server: http://127.0.0.1:5000
i functions: Watching "C:\Users\gunbl\Dev\firebase-tools-debug\.firebase\debug-7611f\functions" for Cloud Functions...
**!! functions: Failed to load function definition from source: FirebaseError: package.json in functions directory has an engines field which is unsupported. Valid choices are: {"node": 10|12|14|16}**

Deploys to runtimes below Node.js 10 are now disabled in the Firebase CLI. Existing Node.js 8 functions will stop executing at a future date. Update existing functions to Node.js 10 or greater as soon as possible.

┌┌─────────────────────────────── ────────────┐
│ ✔ All emulators ready! It is now safe to connect your app. │
└└─────────────────────────────── ────────────┘

┌───────────┬────────────────┐
│ Emulator │ Host:Port │
├───────────┼────────────────┤
│ Functions │ 127.0.0.1:5001 │
├───────────┼────────────────┤
│ Hosting │ 127.0.0.1:5000 │
└───────────┴────────────────┘
   Emulator Hub running at 127.0.0.1:4400
   Other reserved ports: 4500

Geoffrey-Pliez:
@austincrim  I forgot to mention that I am using the code from pull request Add Astro framework #5417

austincrim:
Thanks @Geoffrey-Pliez, I see the error now. It only occurs [if using a Node version of 19](https://github.com/firebase/firebase-tools/blob/d6826ee9e9c2dc5b8e320953aa794be7ba2fdcdb/src/deploy/functions/runtimes/node/parseRuntimeAndValidateSDK.ts#L12-L20) or greater, despite what the error message says.

I wonder if there any potential behavior discrepancies we could see by emulating a function with an different and unsupported Node version than what gets deployed, though I suppose that could already happen if you deploy with, say, 16 but are using 10 locally. cc @jamesdaniels 

In the meantime @Geoffrey-Pliez, could you remove the `astro.js` file and what seem to be a lot of import formatting changes from this PR, just leaving the relevant change?

src/frameworks/index.ts
```diff
@@ -608,8 +608,8 @@ exports.${functionId} = onRequest(${JSON.stringify(
   }
 }
 
-function codegenDevModeFunctionsDirectory() {
-  const packageJson = {};
+function codegenDevModeFunctionsDirectory(projectPath) {
+  const packageJson = require(`${projectPath}/package.json`)
```
jamesdaniels, line 612: Rather than read the package.json, can we change it to `{ engines: { node: NODE_VERSION }}` to address, otherwise we're likely to get extra dependencies and such that the Cloud Functions middleware shim doesn't need.

jamesdaniels (changes requested):

src/frameworks/index.ts
Geoffrey-Pliez: @jamesdaniels 
No because NODE_VERSION is the local value of node.  [see](https://github.com/firebase/firebase-tools/blob/17de2ab9efc85a9ff1fc637fa673ed58e06be874/src/frameworks/index.ts#L111)

For extras, they are deleted [here](https://github.com/firebase/firebase-tools/blob/17de2ab9efc85a9ff1fc637fa673ed58e06be874/src/frameworks/index.ts#L506).
But I would add this:
```suggestion
  delete packageJson.scripts
```

On the other hand, I report an inconsistencies in the code because 

- the version of node must be [>= 16](https://github.com/firebase/firebase-tools/blob/17de2ab9efc85a9ff1fc637fa673ed58e06be874/src/frameworks/index.ts#L261) but only accepts 16 max so we can only use version 16.

-  And package.JSON starts always empty, so the following code is not consistent whether for dev or build
```suggestion
  packageJson.main = "server.js";
      delete packageJson.devDependencies;
      packageJson.dependencies ||= {};
      packageJson.dependencies["firebase-frameworks"] ||= FIREBASE_FRAMEWORKS_VERSION;
      packageJson.dependencies["firebase-functions"] ||= FIREBASE_FUNCTIONS_VERSION;
      packageJson.dependencies["firebase-admin"] ||= FIREBASE_ADMIN_VERSION;
      packageJson.engines ||= {};
      packageJson.engines.node ||= NODE_VERSION;
```

- I also get the message that the firebase-functions version is outdated. 

so I propose that we get Package.json from projectPath in the prepareFramework function


I can code the proposal in this PR

Geoffrey-Pliez: **Other proposal:** 

replace `codegenFunctionsDirectory = codegenDevModeFunctionsDirectory;`
by  `codegenFunctionsDirectory = Framework["ecodegenDevModeFunctionsDirectory"] || codegenDevModeFunctionsDirectory` 

So that we are free to return the package.json
that we want depending on the project

Issue #5415, opened 2023-01-12
Error: Failed to get Firebase project undefined


haode333:
### [REQUIRED] Environment info

```
[debug] [2023-01-12T04:55:23.454Z] CLI Version:   11.20.0
[debug] [2023-01-12T04:55:23.454Z] Platform:      win32
[debug] [2023-01-12T04:55:23.454Z] Node Version:  v19.3.0
```

**firebase-tools:** 11.20.0

**Platform:** win32

### [REQUIRED] Test case

N/A

### [REQUIRED] Steps to reproduce

1. install with npm
2. run ```firebase login```
3. run ```firebase init```
4. run ```firebase deploy```

### [REQUIRED] Expected behavior

Successfully deploy

### [REQUIRED] Actual behavior

[debug] [2023-01-12T04:55:23.451Z] ----------------------------------------------------------------------
[debug] [2023-01-12T04:55:23.453Z] Command:       C:\Users\---\scoop\apps\nodejs\current\node.exe C:\Users\---\scoop\apps\nodejs\current\bin\node_modules\firebase-tools\lib\bin\firebase.js deploy
[debug] [2023-01-12T04:55:23.454Z] CLI Version:   11.20.0
[debug] [2023-01-12T04:55:23.454Z] Platform:      win32
[debug] [2023-01-12T04:55:23.454Z] Node Version:  v19.3.0
[debug] [2023-01-12T04:55:23.454Z] Time:          Wed Jan 11 2023 22:55:23 GMT-0600 (北美中部标准时间)
[debug] [2023-01-12T04:55:23.454Z] ----------------------------------------------------------------------
[debug] 
[debug] [2023-01-12T04:55:23.507Z] >>> [apiv2][query] GET https://firebase.googleapis.com/v1beta1/projects/undefined [none]
[debug] [2023-01-12T04:55:23.746Z] <<< [apiv2][status] GET https://firebase.googleapis.com/v1beta1/projects/undefined 403
[debug] [2023-01-12T04:55:23.746Z] <<< [apiv2][body] GET https://firebase.googleapis.com/v1beta1/projects/undefined {"error":{"code":403,"message":"The caller does not have permission","status":"PERMISSION_DENIED"}}
[debug] [2023-01-12T04:55:23.747Z] HTTP Error: 403, The caller does not have permission
[debug] [2023-01-12T04:55:23.748Z] FirebaseError: HTTP Error: 403, The caller does not have permission
[...]

PR #5483, opened 2023-02-02
Added checker to see if the project ID is valid

aalej:
### Description

Proposed fix for #5415 

When there is no currently selected project, an error is encountered when running `firebase deploy`. Looks like the cause is the [getFirebaseProject](https://github.com/firebase/firebase-tools/blob/e8a59e64d85dee0273e58e1bc5eb86860e5f660d/src/getDefaultHostingSite.ts#L10) method [[method code](https://github.com/firebase/firebase-tools/blob/master/src/management/projects.ts#L430)]. getFirebaseProject creates a GET request even if the projectId is `undefined`, which raises the error.

Checking if the projectId is valid using the [needProjectId](https://github.com/firebase/firebase-tools/blob/e8a59e64d85dee0273e58e1bc5eb86860e5f660d/src/projectUtils.ts#L31) method before creating the GET request should show and error message, that that there is currently no active project.
Example:
```
Error: No project active, but project aliases are available.

Run firebase use <alias> with one of these options:

  default (project_id_0)
  staging (project_id_1)
```

### Scenarios Tested

1. Run `firebase init`, select hosting
2. Run `firebase use --add` to add another project alias
3. Run `firebase use --clear` to unselect currently used project
4. Running `firebase use` should output say that no project is active:
```
No project is currently active, and no aliases have been created..

Project aliases for /<path>:

  default (project_id_0)
  staging (project_id_1)
```
5. Running `firebase deploy` shows trying to deploy to `undefined`:
```
Error: Failed to get Firebase project undefined. Please make sure the project exists and your account has permission to access it.
```


christhompsongoogle (approved):

bkendall (changes requested):
Please add a line to the CHANGELOG.md

```
- Catches error when attempting to deploy without a project (#5415).
```

bkendall (approved):

PR #5490, opened 2023-02-06
[frameworks] better nuxt3 support

DevJoghurt:
### Description

This pull request added better support for nuxt3.
- fixed some bugs for production build, aligned with current nuxt v3.1.2 implementation
- added local dev support
IMPORTANT: local dev only works if this pull request in superstatic will be merged -> https://github.com/firebase/superstatic/pull/389

### Scenarios Tested

firebase emulator:start
firebase deploy

### Sample Commands

firebase emulator:start
firebase deploy


jamesdaniels:
Thanks for this @DevJoghurt, we'll take a look.

jamesdaniels:
PTAL @cfofiu

src/frameworks/nuxt/index.ts
```diff
@@ -31,47 +41,119 @@ export async function discover(dir: string): Promise<{ mayWantBackend: true } |
...
   await warnIfCustomBuildScript(root, name, DEFAULT_BUILD_SCRIPT);
 
-  await buildNuxt(nuxtApp);
-  return { wantsBackend: true };
-}
+  /**
+   * currently genrate is only supported by using nuxi generate or nuxi build --prerender
+   * moreover, it is currently still experimental
+   */
+  const prerender = false;
```
jamesdaniels, line 62: We can turn this into a CLI experiment

DevJoghurt: Prerender is one of several ways to build a nuxt app. This way the whole app is generated locally and no backend is needed. Other build options such as spa or ssr can be configured via config for individual routes, for which prerender is then set to false.
To use prerender as a whole app, you would need to use a cli argument, such as "firebase deploy --nuxt-preprender" or there is a way to specify prerender as config via firebase.json.

jamesdaniels: Ah gotcha, I'll have to think about that more then... 

I've been wondering if we should just shell out to `npm run build` in all these integrations and change the default build step warning to something like `we noticed your build step is not just "${default}". Here be dragons—do what you want—we expect the compiled artifacts to be in ${distDir} at the end of the day`. That way if they want prerender they just change the build step in their package.json. 

Doing all the building in Node has theoretical benefits... but it seems the cost of maintaining these build steps and adding tons of framework specific options outweighs those. As an external contributor to our framework integrations do you agree or have any other feedback about our choice of abstraction?

DevJoghurt: Yes, that makes perfect sense, as it would really be more maintenance free.
But of course it would also be possible to throw out the config for prerender, because the user can really overwrite it in the nuxt.config. The `_generate` config only affects a complete static build of the page anyway. Building the app with different strategies is always possible on route level too. 
Furthermore, with a build script it is no longer so easy to see if the nuxt app really needs a backend or not, since you don't have access to the configuration of nuxt.
I would of course find building the app via node nicer and also more practical but could of course understand the decision.

DevJoghurt: Maybe I can implement both things. If the user defines a build script, then this is used. If there is no build script, the implemented node builder is used.
Here I would take out prerender and read via the generated build config of loadNuxt if a backend is needed or not.

jamesdaniels: Let me and the team think on this strategy a bit before we reshuffle, in the meantime we'll review the rest of this PR.

austincrim:
@DevJoghurt Could you provide more details about what production build bugs this fixes? That will help a lot with review/testing.

DevJoghurt:
@austincrim 
Yeah sure, sorry for the lack of information.

All bugfixes for prod build:
https://github.com/firebase/firebase-tools/blob/f70dad2656446a19f0fca3487e8b5e9e520d810e/src/frameworks/nuxt/index.ts#L64-L79
I have customized the build config for nuxt so that env variables are also added when building the app. Another todo here would be to remove the prerender config again, so that it can be configured by the user via the nuxt.config, as already discussed in the previous thread.

https://github.com/firebase/firebase-tools/blob/f70dad2656446a19f0fca3487e8b5e9e520d810e/src/frameworks/nuxt/index.ts#L81-L91
These are minor changes that cover possible edge cases, such as deleting the output folder and writing the types. In principle, this is simply an adaptation of how nuxi (cli of nuxt 3) builds the app.

https://github.com/firebase/firebase-tools/blob/f70dad2656446a19f0fca3487e8b5e9e520d810e/src/frameworks/nuxt/index.ts#L110
The output folder from the firebase build has caused me problems more often, so I think it's safer to delete it again before building the app.

https://github.com/firebase/firebase-tools/blob/f70dad2656446a19f0fca3487e8b5e9e520d810e/src/frameworks/nuxt/index.ts#L118
The server build tool nitro from nuxt is tracing all packages and adds them as bundledDependencies to the package.json. Since the firebase cli performs an 'npm install', the packages added by the user and traced by nitro are uninstalled again. Therefore, the bundledDependencies must be added as dependencies.

Added local dev with firebase emulator:
https://github.com/firebase/firebase-tools/blob/f70dad2656446a19f0fca3487e8b5e9e520d810e/src/frameworks/nuxt/index.ts#L131-L159
For the local development via firebase emulator i also followed the implementation of nuxi (the cli of nuxt) and adapted it as far as possible to the firebase cli. In principle, most of the functions in utils.ts and prepare.ts are also necessary for this.
As a todo here is still the automatic reload of the nuxt runtime on certain file changes.

I would still add all the todos listed here if you let me know that the principle procedure in this push request is ok.

austincrim:
@DevJoghurt thanks for the detail!

We are moving forward with a different strategy for Nuxt dev mode (#5551) that spawns `nuxt dev` from the `firebase` CLI. If you could remove the dev mode-related parts of this PR, we can proceed with reviewing just the bug fix changes.

src/frameworks/nuxt/index.ts
austincrim: @DevJoghurt Is it possible to detect the `prerender` config from a user's config/project setup?

cc @jamesdaniels

DevJoghurt:
@austincrim 
Sure.
I also tried using spawn first, since it would be much easier that way, of course. However, I had many problems with my more complex nuxt3 projects. See my comments in the new pull request https://github.com/firebase/firebase-tools/pull/5551#issuecomment-1449715510

austincrim:
@DevJoghurt thanks for your input on that PR!

If you're still interested in moving forward with this PR, could you remove the dev mode related changes?

If not, let us know and we can try to adapt your work into a separate PR, with full attribution.

DevJoghurt:
> @DevJoghurt thanks for your input on that PR!
> 
> If you're still interested in moving forward with this PR, could you remove the dev mode related changes?
> 
> If not, let us know and we can try to adapt your work into a separate PR, with full attribution.

Thank you.
I removed all dev related changes and cleaned up code. It should be ready for merging now

JamieCurnow:
@DevJoghurt @TheIronDev @cfofiu any movement on this? 🙏

DevJoghurt:
> @DevJoghurt @TheIronDev @cfofiu any movement on this? 🙏

Do you still have problems with the current firebase cli version?
I was actually planning to close this pull request.

https://github.com/firebase/firebase-tools/blob/f70dad2656446a19f0fca3487e8b5e9e520d810e/src/frameworks/nuxt/index.ts#L118

The only problem I currently have is that when deploying packages are not installed, because nuxt takes bundledDependencies as a reference and this causes problems with the current procedure of the firebase cli. But I solve this with a custom nitro preset.

Maybe @jamesdaniels  can fix this last bug or the workflow of the firebase cli changes to take bundledDependencies into account.

JamieCurnow:
@DevJoghurt Yea that was the root of my issue too and basically the only thing that stopped it working "out of the box". I've also made a custom nitro preset to get around this. Here it is for anyone else who faces this issue:

https://gist.github.com/JamieCurnow/374e69a2c3dd3b4952f967aa00a02938

I'm going to write up a little guide on Medium and link out to a demo minimal repo. Thanks for all your work and the comments around this! Such a great feature - especially having deploy previews of both front and backend with the pinTag feature 😍

JamieCurnow:
I've ended up making a nuxt module to help with this:
https://www.npmjs.com/package/nuxt-firebase-webframework

Feedback much appreciated 😊

PR #5496, opened 2023-02-08
Fix regex page matcher for Next.js middlewares version 1

leoortizz:
### Description

Projects using the first version of the Next.js middleware were failing to deploy due to a difference in `middleware-manifest.json`. The first version was in Next.js until [v12.1.6](https://github.com/vercel/next.js/commit/b188fab3360855c28fd9407bd07c4ee9f5de16a6). Version 2 was introduced in [Next.js 12.2.0](https://nextjs.org/docs/messages/middleware-upgrade-guide). This PR handles the middleware page matchers for version 1.


src/frameworks/next/index.ts
```diff
@@ -277,13 +277,19 @@ export async function ɵcodegenPublicDirectory(sourceDir: string, destDir: strin
   }
 
   const [middlewareManifest, prerenderManifest, routesManifest] = await Promise.all([
-    readJSON<MiddlewareManifest>(join(sourceDir, distDir, "server", MIDDLEWARE_MANIFEST)),
+    readJSON<MiddlewareManifest | MiddlewareManifestV1>(
```
jamesdaniels, line 280: can we instead make `type MiddlewareManifest = MiddlewareManifestV1 | MiddlewareManifestV2`

jamesdaniels: alt, if the middleware manifest type is exported by Next 12 we could add next 12 as a named devDep in the package.json and leverage that. e.g, `"next12": "npm:next@^12", "next13": "npm:next@^13"`

jamesdaniels (changes requested):

jamesdaniels:
Are there any other Next 12 gotchas? E.g, rewrites/redirects/headers?

src/frameworks/next/index.ts
leoortizz: I initially tried having `next12` as a separate dev dep but Next.js has some internal imports from `next/dist/...` that breaks the build. I can investigate further

leoortizz:
> Are there any other Next 12 gotchas? E.g, rewrites/redirects/headers?

I can check but I don't think so. Middlewares was still in beta before Next.js 12.2 so that was the main thing we're dealing with that has changed

src/frameworks/next/index.ts
jamesdaniels: @leoortizz if you tried that and it didn't work well, let's punt. Not worth the time for Next 12, which we'll probably drop at some point

jamesdaniels:
Add changelog

src/frameworks/next/index.ts
```diff
@@ -288,7 +287,9 @@ export async function ɵcodegenPublicDirectory(sourceDir: string, destDir: strin
   ]);
 
   const middlewareMatcherRegexes = Object.values(middlewareManifest.middleware)
-    .map((it) => it.matchers)
+    .map((page: MiddlewareManifest["middleware"]["page"]) =>
```
TheIronDev, line 290: Test this branch? (bonus points illustrating a test -case a previously broken case and showing it works now.)

TheIronDev (changes requested):
Please write a test to "prove" this addresses the mismatch middleware issue.

jamesdaniels (approved):

PR #5498, opened 2023-02-09
Overhaul `ext:dev:init`

joehan:
### Description
Making a bunch of improvements to `ext:dev:init`. The `hello-world` extension created by this command now:
- Includes an example integration test using the Emulator suite
- Passes `npm run lint` on creation
- Includes a `lint:fix` script
- Instructions on how to manually test using the emulator suite

### Scenarios Tested
I tested all 8 combos (js/ts, lint/nolint, npm install/no npm install), and verified that:
- ext:dev:init succeeds
- The resulting extension can be deployed
- The resulting extension can be emulated
- The included tests pass
- When applicable, the linter passes.


templates/extensions/typescript/WELCOME.md
```diff
@@ -1,9 +1,22 @@
-This directory now contains the source files for a simple extension called **greet-the-world**. To try out this extension right away, install it in an existing Firebase project by running:
+s directory now contains the source files for a simple extension called **greet-the-world**. You can try it out right away in the Firebase Emulator suite: first, compile your code by running:
```
huangjeff5, line 1: nit: s ==> This

templates/extensions/typescript/integration-test.ts
```diff
@@ -0,0 +1,13 @@
+import axios from "axios";
+import { expect } from "chai";
+
+describe("greet-the-world", () => {
+  it("should respond with the configured greeting", async () => {
+    const expected = "Hello World from greet-the-world";
+
+    const httpFunctionUri = "http://localhost:5001/demo-test/us-central1/ext-greet-the-world-greetTheWorld/";
```
huangjeff5, line 8: This assumes functions emulator always runs on port 5001, maybe we can make that a commandline option?

pavelgj (approved):

templates/extensions/typescript/integration-test.ts
joehan: We hard code that port in the firebase.json we write (see integration-test.json). There might be a better way to pipe it through from the emulator as an environment variable (and then rely on the default port discovery mechanism instead of hard coding it), but this should work for now.

In the case when 5001 is taken, this will error out with a clear message, too.

huangjeff5 (approved):

PR #5504, opened 2023-02-10
Hosting SSR region

chalosalvador:
### Description
Add the ability to set the region of a web framework's generated SSR function.

- Add initialization step to choose region to host server-side content
- Validate region in `prepareFrameworks`
- Add custom region in `functions.yaml` and the cloud function

### Scenarios Tested
1. Initialize a firebase hosting project with a web framework
  - Verify that the "choose a region" step is added to the initialization steps.
  - Verify that the `firebase.json` file includes the `frameworksBackend` property with the chosen region.
  - Run `firebase deploy` with and without the necessity of a backend.
      - When deploying with necessity of a backend, verify that the `functions.yaml` contains the correct region and the `server.js` file contains the cloud function with the correct region added to it.
  - Verify that the cloud function is actually created in the correct region.
2. Manually change the region in `firebase.json` and verify that the CLI asks to delete or keep the current deployed cloud function with the old region.
3. Set an invalid region the `firebase.json` and verify that the CLI fails with the appropriate error message.


src/frameworks/index.ts
```diff
@@ -560,11 +567,12 @@ ${firebaseDefaults ? `__FIREBASE_DEFAULTS__=${JSON.stringify(firebaseDefaults)}\
       if (bootstrapScript) await writeFile(join(functionsDist, "bootstrap.js"), bootstrapScript);
 
       // TODO move to templates
+      const regionArg = ssrRegion !== DEFAULT_REGION ? `{region: "${ssrRegion}"}, ` : "";
```
jamesdaniels, line 570: We should also add all the frameworksBackend stuff, e.g, `JSON.stringify(frameworksBackend)`

src/init/features/hosting/index.ts
```diff
@@ -116,10 +116,31 @@ export async function doSetup(setup: any, config: any): Promise<void> {
...
+      { name: "us-central1 (Iowa)", value: "us-central1" },
+      { name: "us-west1 (Oregon)", value: "us-west1" },
+      { name: "us-east1 (South Carolina)", value: "us-east1" },
+      { name: "europe-west1 (Belgium)", value: "europe-west1" },
+      { name: "asia-east1 (Taiwan)", value: "asia-east1" },
+    ];
+
+    await promptOnce(
+      {
+        name: "region",
```
jamesdaniels, line 129: add a default of the DEFAULT_REGION

src/init/features/hosting/index.ts
```diff
@@ -116,10 +116,31 @@ export async function doSetup(setup: any, config: any): Promise<void> {
       await WebFrameworks[setup.hosting.whichFramework].init!(setup, config);
     }
 
+    const regionChoices = [
```
jamesdaniels, line 119: See prior comment

jamesdaniels (changes requested):

src/frameworks/index.ts
```diff
@@ -307,6 +309,12 @@ export async function prepareFrameworks(
     if (publicDir) {
       throw new Error(`hosting.public and hosting.source cannot both be set in firebase.json`);
     }
+    const ssrRegion = (frameworksBackend?.region as string) ?? DEFAULT_REGION;
+    if (!ALLOWED_SSR_REGIONS.includes(ssrRegion)) {
+      const notValidRegionError = `Hosting config for site ${site} places server-side content in region ${ssrRegion} which is not known. Valid regions are us-west1, us-central1, us-east1, europe-west1, asia-east1`;
+      console.error(notValidRegionError);
+      throw new Error(notValidRegionError);
```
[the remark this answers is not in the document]
chalosalvador: If we only throw the error, it logs it in `firebase-debug.log` but it doesn't print the error in the console, it only shows `Error: An unexpected error has occurred.`

src/frameworks/index.ts
```diff
@@ -309,6 +317,12 @@ export async function prepareFrameworks(
     if (publicDir) {
       throw new Error(`hosting.public and hosting.source cannot both be set in firebase.json`);
     }
+    const ssrRegion = (frameworksBackend?.region as string) ?? DEFAULT_REGION;
+    if (!allowedRegionsValues.includes(ssrRegion)) {
+      const notValidRegionError = `Hosting config for site ${site} places server-side content in region ${ssrRegion} which is not known. Valid regions are us-west1, us-central1, us-east1, europe-west1, asia-east1`;
```
jamesdaniels, line 322: Map over allowed regions in the message, `ALLOWED_SSR_REGIONS.map(r => r.name).join(', ')`

src/frameworks/index.ts
```diff
@@ -576,11 +589,12 @@ ${firebaseDefaults ? `__FIREBASE_DEFAULTS__=${JSON.stringify(firebaseDefaults)}\
       if (bootstrapScript) await writeFile(join(functionsDist, "bootstrap.js"), bootstrapScript);
 
       // TODO move to templates
+      const firstArg = frameworksBackend ? `${JSON.stringify(frameworksBackend)}, ` : "";
```
jamesdaniels, line 592: Simplify to `JSON.stringify(frameworksBackend || {})`, that way space punc can stay in the onReq string template

jamesdaniels (changes requested):

jamesdaniels (approved):
Made a few small changes, let's :shipit:

src/frameworks/index.ts
```diff
@@ -109,8 +109,15 @@ const SupportLevelWarnings = {
 export const FIREBASE_FRAMEWORKS_VERSION = "^0.6.0";
 export const FIREBASE_FUNCTIONS_VERSION = "^3.23.0";
 export const FIREBASE_ADMIN_VERSION = "^11.0.1";
-export const DEFAULT_REGION = "us-central1";
 export const NODE_VERSION = parseInt(process.versions.node, 10).toString();
+export const DEFAULT_REGION = "us-central1";
+export const ALLOWED_SSR_REGIONS = [
```
luc122c, line 114: Hi, is there a motive behind limiting the regions? For example, we currently use `europe-west2` (London) but it seems this configuration would be disallowed?

tamerxkilinc: That's because cloud Functions V2 are only available in those regions.

gustavopch: @tamerxkilinc This page https://cloud.google.com/functions/docs/locations?hl=en#tier_2_pricing shows that Cloud Functions v2 are currently supported in many more regions. Does that mean we can already use them or are there other limiting factors?

tamerxkilinc: Should be possible to use the new regions now

PR #5507, opened 2023-02-10
Alternative to using esbuild for next.config.js 

jamesdaniels:


jamesdaniels:
Sticking with esbuild for now, as we found some complex cases this didn't work out of box for

Issue #5509, opened 2023-02-11
Installing 11.22.0 results in --inspect-functions only supported for Node.js runtimes and debugger will not attach


Zelfapp:
### [REQUIRED] Environment info

**firebase-tools:** 11.22.0

**Platform:** Manjaro Linux

**Nodejs version is:** 16.16.0

### [REQUIRED] Test case

Using `npm i -g firebase-tools` to install 11.22.0 and then trying to start emulator with `--inspect-functions` I can no longer attach a debugger in 11.22.0. Instead I get the warning during startup that `--inspect-functions only supported for Node.js runtimes`. So I tried this fix on an [issue](https://github.com/firebase/firebase-tools/issues/4885#issuecomment-1320730178) that is now closed, by running `/usr/local/bin/firebase is:npm i node@16 --save-dev` in my functions directory. However, this did not solve the problem. I completely removed the `node_modules` directory and reinstalled everything. Tried completely uninstalling firebase-tools, which is installed globally, and then reinstalling, but the only thing that worked was to downgrade to 11.21.0.

Please advise, what is the upgrade path to 11.22.0? Thanks.

### [REQUIRED] Steps to reproduce

1. From functions directory `firebase emulators:start --import=../saved-data --export-on-exit --inspect-functions` add the `--inspect functions` flag to start in debug mode.
2. Using vscode or webstorm attempt to attach a debugger to listen on port 9229.

### [REQUIRED] Expected behavior

In firestore emulator console should see:

- `>  Debugger listening on ws://127.0.0.1:9229/...` during emulator startup when using `--inspect-functions` flag.
- `>  Debugger attached.` when attaching a debugger from an IDE.

### [REQUIRED] Actual behavior

During firebase emulator startup I see this warning: `--inspect-functions only supported for Node.js runtimes` and I cannot attach a debugger to listen on port 9229.

joerozek:
Experiencing the same problem on Max OSX 12.5, node 16.15

Zelfapp:
> Experiencing the same problem on Max OSX 12.5, node 16.15

Give my post a thumbs up and it will get more attention. I still have not found a resolution.

tanaka-baseed:
[functionsEmulator.ts](https://github.com/firebase/firebase-tools/blob/master/src/emulator/functionsEmulator.ts)
`      if (!emulatableBackend.bin?.startsWith("node")) {
        this.logger.log("WARN", "--inspect-functions only supported for Node.js runtimes.");`

[...]

PR #5516, opened 2023-02-14
Fix bug where --inspect-functions flag always fails.

taeold:
Fixes https://github.com/firebase/firebase-tools/issues/5509

### DIFF
--- a/src/emulator/functionsEmulator.ts
+++ b/src/emulator/functionsEmulator.ts
@@ -690,7 +690,8 @@ export class FunctionsEmulator implements EmulatorInstance {
     // In debug mode, we eagerly start the runtime processes to allow debuggers to attach
     // before invoking a function.
     if (this.args.debugPort) {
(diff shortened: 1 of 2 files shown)


src/emulator/functionsEmulator.ts
```diff
@@ -690,7 +690,8 @@ export class FunctionsEmulator implements EmulatorInstance {
     // In debug mode, we eagerly start the runtime processes to allow debuggers to attach
     // before invoking a function.
     if (this.args.debugPort) {
-      if (!emulatableBackend.bin?.startsWith("node")) {
+      console.log(emulatableBackend.bin);
```
jamesdaniels, line 693: Drop console.log?

joehan, line 693: Clean up the console.log (or switch to logger.debug if we need this info)

joehan (approved):
LGTM after you remove the console.log

Issue #5517, opened 2022-10-24
Functions discovery breaks with yarn/npm/pnpm workspaces


andrewkrippner:
### [REQUIRED] Version info

**node:**
v16.15.0

**firebase-functions:**
v4.0.1

**firebase-tools:**
v11.15.0

**firebase-admin:**
v11.2.0

### [REQUIRED] Test case

```
firebase init functions
>Don't set up a default project 
>Typescript
>ESLint: No
>Install Deps: No

Add {
    "projects": {
        "default": "default"
    }
} to /.firebaserc

Uncomment helloWorld function

yarn set version berry
Add [nodeLinker: node-modules](https://yarnpkg.com/configuration/yarnrc#nodeLinker) in your .yarnrc.yml file
Add "workspaces": ["functions"] to /package.json

cd functions
yarn build
cd ..

yarn up firebase-functions@4.0.1
```

### [REQUIRED] Steps to reproduce

```firebase functions:shell```

### [REQUIRED] Expected behavior

Functions found by firebase and run in emulator and deploy properly, as they do before ```yarn up firebase-functions@4.0.1``` is ran

### [REQUIRED] Actual behavior

i  functions: Loaded functions: 
No functions emulated.

### Were you able to successfully deploy your functions?

No, I receive the error:

The following functions are found in your project but do not exist in your local source code:
*

PR #5518, opened 2023-02-14
Lookup more places to find Firebase Functions SDK in developer's workspace.

taeold:
Fixes https://github.com/firebase/firebase-tools/issues/5517.


TheIronDev (approved):
Lgtm


Nit: create separate functions to test the path creations

PR #5548, opened 2023-02-23
Firebase CLI Supports Multiple Firestore Databases

VicVer09:
b/267551167
go/add-database-id-to-firebase-cli

### Description
To support multiple databases in the Firebase CLI, commands should handle an optional databaseId to specify which database it should be applied to.

### Details

- firebase config schema reflects firestore config can be single as before or accepts multiple databases/targets
- fsConfig mimics the logic in src/database/rulesConfig.ts to parse the possible Single/Multiple configs into an array of rules/indexes/databaseId for deploy command
- deploy uses this parsed config to promise all index/rules changes
- mechanism for updating rules/indexes is unchanged, all that is necessary is to pass the proper databaseId to the indexes API from the config
- ~~delete command will do a check to make sure there is only one database present, otherwise a databaseId must be specified~~ delete --database will be (default) if none provided
- ~~provisioning api skeleton started for getting the database list, this will be expanded in future to support more commands (marked as todo) for now keep this file simple~~
- controller for now only accepts default database configuration, since emulator does not yet support multiple databases

#### Deploy

- checkout branch (see go/firebase-cli-contributor for detailed setup guide)
- `npm link`
- in a separate directory:
- create gcp project (using gcloud, api, or ui)
- create firebase project or using firebase tools
- add project to multiple databases allowlist go/firestore-enable-multiple-databases
- `firebase init` with firestore enabled in a separate directory
- create default database (using gcloud, api, or ui)
- create a named database (using api)
- configure rules for both databases in separate files (sample https://github.com/firebase/quickstart-js/blob/master/firestore/firestore.indexes.json)
- configure indexes for both in separate files (sample https://firebase.google.com/docs/firestore/security/get-started#deny-all)

configure firebase.json to contain:
```
    {
      "database": "(default)",
      "rules": "firestore.rules",
      "indexes": "firestore.indexes.json"
    },
    {
      "database": "named",
      "rules": "named.rules",
      "indexes": "named.indexes.json"
    }
  ]
```

- run `firebase deploy --only firestore`
- check composite indexes are updated using gcloud:
- `gcloud firestore indexes composite list --database named`
- `gcloud firestore indexes composite list --database \(default\)`
- check rules are updated for default by checking ui
- check rules are updated for named db by trying to write a document to named database using api
- modify rules/indexes slightly and deploy again, verify the changes are picked up

#### Commands
- add a document to database in firebase ui or using firestore api
- add a composite index to database in firebase ui or firestore api (can skip if did this in deploy testing above)
- test delete by running: `firebase firestore:delete --all-collections --recursive --database="(default)"` or with named db
- test list indexes by running `firebase firestore:indexes` or `firebase firestore:indexes --database="named"`


src/commands/firestore-delete.ts
```diff
@@ -84,6 +85,10 @@ export const command = new Command("firestore:delete [path]")
       "including all collections and documents. Any other flags or arguments will be ignored."
   )
   .option("-f, --force", "No confirmation. Otherwise, a confirmation prompt will appear.")
+  .option(
+    "--database <databaseId>",
+    "Database ID for database to delete from. Mandatory if more than 1 database present."
```
rwhogg, line 90: [Nitpick] "database present" => "database is present"

src/firestore/fsConfig.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export interface ParsedFirestoreConfig {
+  database: string;
+  rules?: string;
+  indexes?: string;
+}
+
+export function getFirestoreConfig(projectId: string, options: Options): ParsedFirestoreConfig[] {
+  const fsConfig = options.config.src.firestore;
+  if (fsConfig === undefined) {
```
rwhogg, line 13: Why `=== undefined` instead of just a falsy check?

src/firestore/fsConfig.ts
```diff
@@ -0,0 +1,78 @@
...
+      const databaseId = fsConfig.database || `(default)`;
+      return [{ rules: fsConfig.rules, indexes: fsConfig.indexes, database: databaseId }];
+    } else {
+      logger.debug("Possibly invalid database config: ", JSON.stringify(fsConfig));
+      return [];
+    }
+  }
+
+  const results: ParsedFirestoreConfig[] = [];
+  for (const c of fsConfig) {
```
rwhogg, line 47: This block of the function (the loop across the `c` of fsConfig) is a bit confusing IMO. Would it be possible to explain some of the thinking here a bit more?

My apologies.

VicVer09: the confusing part here is mostly from logic to support options using `--only`.

target-to-database is a one-to-many mapping that the user can configure using https://firebase.google.com/docs/cli/targets#set-up-deploy-target-hosting

think of it this way, we will only include the config if there is no `--only` clause, or if the database is included in the target mapping passed to the `--only` option
for the target case, we first check that the target mapping exists, and if it is then push the rules/indexes specified for each database in the mapping

the logic is equivalent to https://github.com/firebase/firebase-tools/blob/master/src/database/rulesConfig.ts#L62 used for rtdb (goal here being to be as consistent as possible to how rtdb is configured in firebase.json

src/firestore/util.ts
```diff
@@ -29,14 +31,15 @@ export function parseIndexName(name?: string): IndexName {
   }
 
   const m = name.match(INDEX_NAME_REGEX);
-  if (!m || m.length < 4) {
+  if (!m || m.length < 5) {
```
rwhogg, line 34: What is the 4->5 increase needed for?

src/firestore/provisioning.ts
```diff
@@ -0,0 +1,54 @@
+import { firestoreOrigin } from "../api";
+import { Client } from "../apiv2";
+import { FirebaseError } from "../error";
+
+// todo add enums for possible string values
```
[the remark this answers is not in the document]
VicVer09: no this will be deferred to provisioning API story

src/firestore/util.ts
VicVer09: yep

src/firestore/fsConfig.ts
VicVer09: main motivation was to stay consistent with rtdb. Either would work however there is an edge case if fsConfig is an empty object `{}` we will miss adding an empty configuration for (default). there should not be any repercussions to this (either way deploy would be a no-op) but my preference would be to stay as close to the rtdb implementation as possible

src/firestore/util.ts
bkendall, line 34: If there aren't tests for this file, this feels like a really good time to write some. If there are tests, it would be good to update them to check this logic out! (I'm skiddish around regexes - always better to have some tests with them)

src/test/firestore/indexes.spec.ts
```diff
@@ -117,6 +118,29 @@ describe("IndexNameParsing", () => {
...
+      databaseId: "(default)",
+      collectionGroupId: "collection",
+      fieldPath: "abc123",
+    });
+  });
+
+  it("should parse an index name from a named database correctly", () => {
+    const name =
+      "/projects/myproject/databases/named-db/collectionGroups/collection/indexes/abc123/";
+    expect(util.parseIndexName(name)).to.eql({
```
bkendall, line 130: Okay, there are tests for `util`, but they're in the `indexes.spec.ts`  file? That's... strange. I'd expect tests for `util` functions to be in `util.spec.ts`... maybe not worth fixing in this CL (because it'll be a lot of extra changes to sort though), but maybe follow up to address that.

bkendall (changes requested):
I did a once over and there's some things to address. You also should include an entry in `CHANGELOG.md`. Once you've addressed comments from me, feel free to re-request my review which will reset my bit

src/commands/firestore-indexes-list.ts
```diff
@@ -13,13 +13,18 @@ export const command = new Command("firestore:indexes")
     "Pretty print. When not specified the indexes are printed in the " +
       "JSON specification format."
   )
+  .option(
+    "--database <databaseId>",
+    "Database ID of the firestore database from which to list indexes. (default) if none provided."
+  )
   .before(requirePermissions, ["datastore.indexes.list"])
   .before(warnEmulatorNotSupported, Emulators.FIRESTORE)
   .action(async (options: any) => {
```
joehan, line 22: Optional: While you're making changes to these commands, it would be nice to give options a type other than 'any' - this would silence the many unsafe assignment warnings  throughout these commands. 

We have a base type in  [options.ts](https://github.com/firebase/firebase-tools/blob/master/src/options.ts)  that contains the fields that are present in all commands. Ideally, you'd do something like hosting does in [hosting/options.ts](https://github.com/firebase/firebase-tools/blob/master/src/hosting/options.ts), where we extend the base type with fields specific to the command, and then use that.

src/commands/firestore-indexes-list.ts
```diff
@@ -13,13 +13,18 @@ export const command = new Command("firestore:indexes")
...
     const indexApi = new fsi.FirestoreIndexes();
 
-    const indexes = await indexApi.listIndexes(options.project);
-    const fieldOverrides = await indexApi.listFieldOverrides(options.project);
+    const databaseId = options.database || "(default)";
```
joehan, line 25: Nit: prefer ?? over || here since we only want to fall back to "(default)" if options.database is undefined

src/deploy/firestore/release.ts
```diff
@@ -1,25 +1,36 @@
...
-  const rulesFile = options.config.src.firestore?.rules;
-  if (!rulesFile) {
-    throw new FirebaseError(
-      `Invalid firestore config: ${JSON.stringify(options.config.src.firestore)}`
-    );
-  }
-  await rulesDeploy.release(rulesFile, RulesetServiceType.CLOUD_FIRESTORE);
+  const rulesContext = _.get(context, "firestore.rules");
+  return Promise.all(
+    rulesContext.map((ruleContext: any) => {
```
joehan, line 20: Suggest making this an async function - then, you can omit the Promise.resolves inside it and just return what you want

joehan, line 20: ruleContext should have a real type.

src/deploy/firestore/release.ts
```diff
@@ -1,25 +1,36 @@
...
   }
 
-  const rulesFile = options.config.src.firestore?.rules;
-  if (!rulesFile) {
-    throw new FirebaseError(
-      `Invalid firestore config: ${JSON.stringify(options.config.src.firestore)}`
-    );
-  }
-  await rulesDeploy.release(rulesFile, RulesetServiceType.CLOUD_FIRESTORE);
+  const rulesContext = _.get(context, "firestore.rules");
```
joehan, line 18: Prefer optional chaining over _.get:
```
const rulesContext = context.?firestore.?rules
```

src/firestore/delete.ts
```diff
@@ -79,7 +86,9 @@ export class FirestoreDelete {
     this.path = this.path.replace(/(^\/+|\/+$)/g, "");
 
     this.allDescendants = this.recursive;
-    this.root = "projects/" + project + "/databases/(default)/documents";
+
+    // databaseId guaranteed non null
```
joehan, line 90: The strong typing you used for 'options' makes this comment redundant

src/firestore/provisioning.ts
joehan, line 5: Consider a string literal union instead of an enum: https://www.typescriptlang.org/docs/handbook/literal-types.html#string-literal-types

joehan: Also, TODOs should have owners or bugs attached:
`// TODO(VicVer09): ... `

src/firestore/provisioning.ts
```diff
@@ -0,0 +1,54 @@
...
+  private apiClient: Client;
+
+  constructor(project: string) {
+    this.project = project;
+    this.apiClient = new Client({ urlPrefix: firestoreOrigin, apiVersion: "v1" });
+  }
+
+  /**
+   * Lists databases for project
+   * todo b/267473272 refactor this into a comprehensive package for Provisioning API
```
joehan, line 30:
```suggestion
   * TODO(b/267473272): refactor this into a comprehensive package for Provisioning API
```

joehan (changes requested):

src/deploy/firestore/prepare.ts
```diff
@@ -7,6 +7,17 @@ import utils = require("../../utils");
...
 
+export interface RulesContext {
+  databaseId: string;
+  rulesFile: string;
+}
+
+export interface IndexContext {
+  databaseId: string;
+  indexesFileName: string;
+  indexesSrc: any;
```
rwhogg, line 18: What is `indexesSrc`? Is there a more-specific type you can give it than any?

src/deploy/firestore/release.ts
```diff
@@ -3,34 +3,31 @@ import * as _ from "lodash";
...
-          `Invalid firestore config for ${databaseId} database: ${JSON.stringify(
-            options.config.src.firestore
-          )}`
-        );
-        return Promise.resolve();
-      }
-      return Promise.resolve(
-        rulesDeploy.release(rulesFile, RulesetServiceType.CLOUD_FIRESTORE, databaseId)
+  const rulesContext: RulesContext[] = context?.firestore?.rules;
+  rulesContext.map(async (ruleContext: any) => {
```
rwhogg: Is the cast to `any` needed?

src/deploy/firestore/prepare.ts
VicVer09: indexesSrc is an object describing the raw index source spec found in the config.

the issue here is this may represent the old v1beta1 indexes spec or the new v1/v1 format. The FirestoreIndexes.deploy function does the messy work of converting it from `any` into `Spec.IndexFile` so to minimize changes i think it's acceptable to keep it as `any` here.

i will rename it to `indexesRawSrc` to be more specific

src/deploy/firestore/release.ts
VicVer09: no it is not - good catch

src/test/firestore/indexes.spec.ts
VicVer09: good idea, this isn't a huge change - moved

src/commands/firestore-indexes-list.ts
VicVer09: good idea - done!

src/emulator/controller.ts
```diff
@@ -594,7 +594,10 @@ export async function startAll(
     }
 
     const config = options.config;
-    const rulesLocalPath = config.src.firestore?.rules;
+    // emulator does not support multiple databases config yet
+    // todo b/269787702
+    const rulesLocalPath =
+      config.src.firestore && "rules" in config.src.firestore && config.src.firestore?.rules;
```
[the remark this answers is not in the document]
VicVer09: refactored this to use fsConfig

VicVer09:
addressed comments
- added changelog.md changes
- removed provisioning api
- removed as many instances of `any` as possible

as per firebase api proposal decided to remove requirement to pass database for delete command - instead use `(default)` to stay consistent with indexes

src/deploy/firestore/prepare.ts
```diff
@@ -5,45 +5,60 @@ import { loadCJSON } from "../../loadCJSON";
...
+export interface IndexContext {
+  databaseId: string;
+  indexesFileName: string;
+  indexesRawSpec: any; // could be the old v1beta1 indexes spec or the new v1/v1 format
+}
 
 /**
  * Prepares Firestore Rules deploys.
  * @param context The deploy context.
  * @param options The CLI options object.
```
rwhogg, line 24: Please update the docblock

src/deploy/firestore/prepare.ts
```diff
@@ -5,45 +5,60 @@ import { loadCJSON } from "../../loadCJSON";
...
+  context.firestore.rules.push({
+    databaseId,
+    rulesFile,
+  } as RulesContext);
 }
+
 /**
  * Prepares Firestore Indexes deploys.
  * @param context The deploy context.
  * @param options The CLI options object.
```
rwhogg, line 42: Ditto

src/deploy/firestore/prepare.ts
```diff
@@ -65,6 +80,30 @@ export default async function (context: any, options: any): Promise<void> {
...
+  );
+  if (!firestoreConfigs || firestoreConfigs.length === 0) {
+    return;
+  }
+
+  context.firestore = context.firestore || {};
+  context.firestore.indexes = [];
+  context.firestore.rules = [];
+  const rulesDeploy: RulesDeploy = new RulesDeploy(options, RulesetServiceType.CLOUD_FIRESTORE);
+  _.set(context, "firestore.rulesDeploy", rulesDeploy);
```
joehan, line 95: This _.set is not necessary because 3 lines above you ensure that context.firestore exists.
```suggestion
  context.firestore.rulesDeploy = rulesDeploy;
```

joehan: This is also the last use of lodash in the file i beleive, so you can remove it's import too 🧹

src/deploy/firestore/prepare.ts
```diff
@@ -65,6 +80,30 @@ export default async function (context: any, options: any): Promise<void> {
...
+
+  firestoreConfigs.forEach((firestoreConfig: fsConfig.ParsedFirestoreConfig) => {
```
joehan, line 97: Normal for loops are strongly preferred over foreach, as they lead to less confusion about scope (go/tsjs-practices/iteration#avoid-using-foreach-when-other-approaches-are-better)

src/deploy/firestore/release.ts
```diff
@@ -1,25 +1,33 @@
...
+      const rulesFile = ruleContext.rulesFile;
+      if (!rulesFile) {
+        logger.error(
+          `Invalid firestore config for ${databaseId} database: ${JSON.stringify(
+            options.config.src.firestore
+          )}`
+        );
+        return;
+      }
+      await rulesDeploy.release(rulesFile, RulesetServiceType.CLOUD_FIRESTORE, databaseId);
```
joehan, line 30: Since ruleDeploy.release returns Promise<void>, i think you can return it instead of awaiting it here
```suggestion
      return rulesDeploy.release(rulesFile, RulesetServiceType.CLOUD_FIRESTORE, databaseId);
```

src/firestore/options.ts
```diff
@@ -0,0 +1,27 @@
+import { FirebaseConfig } from "../firebaseConfig";
+
+/**
+ * The set of fields that the Firestore commands need from Options.
+ * It is preferable that all codebases use this technique so that they keep
+ * strong typing in their codebase but limit the codebase to have less to mock.
```
joehan, line 6: ❤️

joehan (approved):
Some minor fixes, but this is at LGTM. Since this is a lot of new functionality, I'd recommend running a bug bash with your team before release to iron out any issues that the review missed as well.

rwhogg (approved):

bkendall (approved):
One comment, but the rest looks good. Thanks.

src/firestore/options.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+export interface FirestoreOptions {
```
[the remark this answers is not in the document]
VicVer09: good idea - done!

PR #5551, opened 2023-02-24
Nuxt dev mode support

leoortizz:


tsconfig.json
```diff
@@ -9,10 +9,8 @@
     "allowSyntheticDefaultImports": true,
     "target": "es2017",
     "sourceMap": true,
-    "typeRoots": [
-      "node_modules/@types",
-      "src/types"
-    ]
+    "skipLibCheck": true,
```
leoortizz, line 12: this is a workaround for having Nuxt as a `devDependency`, otherwise it breaks the build

DevJoghurt:
I have tested this pull request with my more complex nuxt3 apps and have the following comments.
The current implementation does not support monorepo setup. For this you could change the call a little bit and see where the app is located and then just let the nuxt cli do everything else:
```js  
  // get root dir of nuxt app
  const cmdParams = ['dev'];
  if(dir.split(process.cwd()).length > 1){
    cmdParams.push(dir.split(process.cwd() + '/')[1])
  }
  const host = new Promise<string>((resolve) => {
    const serve = spawn(CLI_COMMAND, cmdParams);
  ...
```

I have also noticed many URL errors in my more complex apps, the following fix seems to be the way forward:
```js
 // workaround for non-standard Nuxt URLs
 if (queryString.includes("vue&")) {
   fetchURL = fetchURL.replace(/\=$/, "").replace(/\=&/, "&");
 }
```
However, I still have massive problems for example with tailwind (@nuxtjs/tailwindcss) or with the new dev tool (@nuxt/devtools). I think the problem will also occur with many other use cases, but i have not yet been able to look deeper to find out exactly where the problem lies.

leoortizz:
> I have tested this pull request with my more complex nuxt3 apps and have the following comments. The current implementation does not support monorepo setup. For this you could change the call a little bit and see where the app is located and then just let the nuxt cli do everything else:
> 
> ```js
>   // get root dir of nuxt app
>   const cmdParams = ['dev'];
>   if(dir.split(process.cwd()).length > 1){
>     cmdParams.push(dir.split(process.cwd() + '/')[1])
>   }
>   const host = new Promise<string>((resolve) => {
>     const serve = spawn(CLI_COMMAND, cmdParams);
>   ...
> ```
> 
> I have also noticed many URL errors in my more complex apps, the following fix seems to be the way forward:
> 
> ```js
>  // workaround for non-standard Nuxt URLs
>  if (queryString.includes("vue&")) {
>    fetchURL = fetchURL.replace(/\=$/, "").replace(/\=&/, "&");
>  }
> ```
> 
> However, I still have massive problems for example with tailwind (@nuxtjs/tailwindcss) or with the new dev tool (@nuxt/devtools). I think the problem will also occur with many other use cases, but i have not yet been able to look deeper to find out exactly where the problem lies.

> The current implementation does not support monorepo setup

Thanks for the great feedback! I'll investigate the problems you mentioned.

leoortizz:
> I have tested this pull request with my more complex nuxt3 apps and have the following comments. The current implementation does not support monorepo setup. For this you could change the call a little bit and see where the app is located and then just let the nuxt cli do everything else:
> 
> ```js
>   // get root dir of nuxt app
>   const cmdParams = ['dev'];
>   if(dir.split(process.cwd()).length > 1){
>     cmdParams.push(dir.split(process.cwd() + '/')[1])
>   }
>   const host = new Promise<string>((resolve) => {
>     const serve = spawn(CLI_COMMAND, cmdParams);
>   ...
> ```

I tried a monorepo with Nuxt3 projects and also had issues running it. Turns out the errors were also happening by running the Nuxt project directly, without Firebase. I found the fixes for those errors here: https://github.com/nuxt/nuxt/issues/14146#issuecomment-1397344559

Does this fixes your issues with monorepos? If not can you share more details about the errors you had and ways to reproduce it?

> I have also noticed many URL errors in my more complex apps, the following fix seems to be the way forward:
> 
> ```js
>  // workaround for non-standard Nuxt URLs
>  if (queryString.includes("vue&")) {
>    fetchURL = fetchURL.replace(/\=$/, "").replace(/\=&/, "&");
>  }
> ```
> 
> However, I still have massive problems for example with tailwind (@nuxtjs/tailwindcss) or with the new dev tool (@nuxt/devtools). I think the problem will also occur with many other use cases, but i have not yet been able to look deeper to find out exactly where the problem lies.

I was able to reproduce the errors with `@nuxtjs/tailwindcss` and the small change I pushed fixed that. Can you try and see if your errors are gone? If not can you share the errors you get?

For `@nuxt/devtools` the problem seems to be related to our proxy which doesn't support WebSockets. The WebSockets support is being implemented.

DevJoghurt:
@leoortizz 
Yes, I can confirm that even my larger projects are now working as expected.

Regarding the monorepo setup, I currently only tested yarn or npm with workspaces, there the error seems to be different.
I was able to fix the issue by not explicitly setting the CLI_COMMAND to the path of node_modules, but simply using nuxt dev directly. Is there a special reason for taking the path ("node_modules/.bin/nuxt")?

src/frameworks/nuxt/index.ts
```diff
@@ -30,10 +39,10 @@ export async function discover(dir: string): Promise<{ mayWantBackend: true } |
   const version = nuxtDependency?.version;
   const anyConfigFileExists = await nuxtConfigFilesExist(dir);
 
-  if (!anyConfigFileExists && !nuxtDependency) return;
-  if (version && gte(version, "3.0.0-0")) return { mayWantBackend: true };
+  if (!anyConfigFileExists && !nuxtDependency) return { publicDirectory };
```
jamesdaniels, line 42: This should return void no?

jamesdaniels (changes requested):

src/frameworks/nuxt/index.ts
leoortizz: probably yes, I'll refactor this function as it got messy with my changes

src/test/frameworks/nuxt/helpers.ts
```diff
@@ -0,0 +1,7 @@
+import type { NuxtOptions } from "../../../frameworks/nuxt/interfaces";
```
TheIronDev, line 1: Nit: this file seems unnecessary if its only being used in the spec file.

src/frameworks/nuxt/index.ts
```diff
@@ -75,3 +84,29 @@ export async function ɵcodegenFunctionsDirectory(sourceDir: string, destDir: st
   await copy(join(sourceDir, ".output", "server"), destDir);
   return { packageJson: { ...packageJson, ...outputPackageJson }, frameworksEntry: "nuxt3" };
 }
+
+export async function getDevModeHandle(dir: string) {
```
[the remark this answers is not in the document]
TheIronDev: Thanks for writing tests for this!

src/test/frameworks/nuxt/helpers.ts
leoortizz: @TheIronDev it's also being used here: https://github.com/firebase/firebase-tools/blob/a476540d2a908bfd5168e5342dfdf4c1a6c5040b/src/frameworks/nuxt/index.ts#L14

src/frameworks/nuxt/index.ts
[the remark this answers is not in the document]
leoortizz: Thanks for pushing for it 🤝

TheIronDev (approved):
Approving. I'm Awaiting a James Approval before merging this in.

src/test/frameworks/nuxt/helpers.ts
leoortizz: @TheIronDev sorry, just realized you're not referring to `frameworks/nuxt/interfaces.ts` but to `test/frameworks/nuxt/helpers.ts`. I expected this file to have more stuff but it wasn't necessary. Thanks, removing it now.

jamesdaniels (approved):
Couple nits, I dropped them in—otherwise let's go!

PR #5572, opened 2023-03-06
FIREBASE_FRAMEWORKS_BUILD_TARGET

chalosalvador:
### Description

- Add support for `FIREBASE_FRAMEWORKS_BUILD_TARGET` environment variable to override the default build target for `firebase serve`, `firebase emulator:start` and `firebase emulators:exec`.
- Valid options `dev` or `prod`

- TODO: Support other targets for Angular.

### Scenarios Tested

- Tested the environment variable running the different frameworks: Next.js, Svelte, React, Vite, Lit.

### Sample Commands

```
$ NODE_ENV=production firebase emulators:start

$ FIREBASE_FRAMEWORKS_BUILD_TARGET=production firebase emulators:start

$ NODE_ENV=production FIREBASE_FRAMEWORKS_BUILD_TARGET=staging firebase emulators:start

```


src/frameworks/index.ts
```diff
@@ -404,16 +406,23 @@ export async function prepareFrameworks(
       support,
     } = WebFrameworks[framework];
     console.log(`Detected a ${name} codebase. ${SupportLevelWarnings[support] || ""}\n`);
+    const hostingEmulatorInfo = emulators.find((e) => e.name === Emulators.HOSTING);
+
+    const frameworksBuild = getFrameworksBuildMode(options);
+
     // TODO allow for override
-    const isDevMode = context._name === "serve" || context._name === "emulators:start";
+    const isDevMode =
```
jamesdaniels, line 414: can we do context._name !== "deploy" rather? also if !isDevMode let's ahead and set NODE_ENV=production

src/emulator/commandUtils.ts
```diff
@@ -59,6 +59,9 @@ export const FLAG_TEST_PARAMS = "--test-params <params.env file>";
 export const DESC_TEST_PARAMS =
   "A .env file containing test param values for your emulated extension.";
 
+export const FLAG_FRAMEWORKS_BUILD = "--frameworks-build [environment]";
+export const DESC_FRAMEWORKS_BUILD = "Specify an environment build to preview locally.";
```
jamesdaniels, line 63: let's add prod and dev options to the message

jamesdaniels (changes requested):

src/frameworks/index.ts
chalosalvador: Typescript complains with `Cannot assign to 'NODE_ENV' because it is a read-only property` if we try to assign something to `process.env.NODE_ENV`.

src/frameworks/index.ts
```diff
@@ -626,6 +630,22 @@ function codegenDevModeFunctionsDirectory() {
...
 }
 
+function getFrameworksBuildMode(options: any): FrameworksBuild {
+  let frameworksBuild: FrameworksBuild = options.frameworksBuild;
+  if (frameworksBuild) {
+    // TODO validate frameworksBuild with other possible values (for Angular)
+    if (!FRAMEWORKS_BUILD_OPTIONS.includes(frameworksBuild)) {
+      throw new FirebaseError(`Invalid value for frameworksBuild: ${frameworksBuild}`);
+    }
+  } else if (process.env.NODE_ENV === "production") {
```
jamesdaniels, line 640: Add TODO about other languages, e.g, Python

jamesdaniels (approved):
LGTM, let's wait on the API proposal review since it's this week. Please add a changelog entry in the meantime

PR #5589, opened 2023-03-12
Tab completion

sassanh:
### Description

It adds `npm run generate:completion-files` to generate `completion.bash` and `completion.fish` files. It also adds `firebase completion <shell>` command to print those files into stdout (`<shell>` can be "bash" or "fish").

The goal is to let the user easily add bash/fish completion so that they can use firebase easier.

### Scenarios Tested

I built and installed firebase-cli on my machine and then ran `firebase completion fish | source`, then ran `firebase ` and then used tab completion to write commands and options and then removed them and wrote other commands and options using tab completion.
I also ran `source <(firebase completion fish)` in bash and did some tests in bash.

For example writing `firebase appdistribution:distribute --groups-file ` and then pressing tab should show files or writing `firebase a` followed by tab should show this in bash:
```
appdistribution:distribute      apps:android:sha:delete         apps:sdkconfig
appdistribution:testers:add     apps:android:sha:list           auth:export
appdistribution:testers:remove  apps:create                     auth:import
apps:android:sha:create         apps:list                       
```
and this in fish:
```
appdistribution:distribute                                                          (upload a release binary)
appdistribution:testers:add                                                          (add testers to project)
appdistribution:testers:remove                                                (remove testers from a project)
apps:android:sha:create                                      (add a SHA certificate hash for a given app id.)
apps:android:sha:delete                                   (delete a SHA certificate hash for a given app id.)
apps:android:sha:list                                   (list the SHA certificate hashes for a given app id.)
apps:create            (create a new Firebase app. [platform] can be IOS, ANDROID or WEB (case insensitive).)
apps:list  (list the registered apps of a Firebase project. Optionally filter apps by [platform]: IOS, ANDR…)
apps:sdkconfig  (print the Google Services config of a Firebase app. [platform] can be IOS, ANDROID or WEB …)
auth:export                                     (Export accounts from your Firebase project into a data file)
auth:import                         (import users into your Firebase project from a data file(.csv or .json))
```

### Sample Commands

```bash
npm run generate:completion-files
source <(firebase completion bash)
firebase completion bash >> ~/.bashrc
firebase completion fish | source
firebase completion fish >> ~/.config/fish/config.fish
```


joehan:
Hey @sassanh, thanks for the contribution, I _love_ the idea of adding autocomplete for firebase commands. However, this PR changes the public surface of firebase-tools by adding a new command `firebase completion`. We try to be very careful about adding new commands, and have an internal review process for any changes to the API surface. I'm going to spend some time looking for a good way to do this without adding a new command. This will stay unreviewed for a little bit - but don't worry, it is still on my plate!

sassanh:
Hey @joehan, thanks for informing me about this! I understand you are being careful about adding new commands. It's ok if this PR stays unreviewed.

An alternative to changing the public surface would be to introduce a new npm package for the purpose of providing completion files. It wouldn't be as convenient for the end user though.

clari7744:
It looks like this PR has stagnated; I'm here two years later still looking for Zsh tab completion for Firebase's (mostly unintuitive) commands. Any progress on adding Zsh completion so I can download it manually, or on reviewing this PR so that it can be added to Firebase proper?

PR #5592, opened 2023-03-14
Add 2nd gen firestore triggers to firebase deploy

colerogers:


TheIronDev (approved):

src/deploy/functions/services/firestore.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Sets a firestore event trigger's region to the firestore database region.
+ * @param endpoint the firestore endpoint
+ */
+export async function ensureFirestoreTriggerRegion(
+  endpoint: backend.Endpoint & backend.EventTriggered
+): Promise<void> {
+  const db = await firestore.getDatabase(
+    endpoint.project,
+    endpoint.eventTrigger.eventFilters?.database || "(default)"
+  );
```
taeold, line 15: q: in `functions/services/database` we don't make similar guarantee like we've done here by making a call to get RTDB info. What's the difference here?

src/deploy/functions/services/index.ts
```diff
@@ -117,6 +119,17 @@ const testLabService: Service = {
   unregisterTrigger: noop,
 };
 
+/** A firestore service object */
+const firestoreService: Service = {
+  name: "firestore",
+  api: "firestore.googleapis.com",
+  requiredProjectBindings: noopProjectBindings,
```
taeold, line 126: huh this is a little surprising (that everything would just work w/o additional IAM fixing). Do you expect this to change later?

src/deploy/functions/services/firestore.ts
```diff
@@ -0,0 +1,25 @@
...
+  const dbRegion = db.locationId;
+  if (!endpoint.eventTrigger.region) {
+    endpoint.eventTrigger.region = dbRegion;
```
taeold, line 18: Q: Aren't Firestore db location multi-region? (e.g. `nam5` vs `us-central1`)? Is it possible to have multi-region triggers?

taeold: Ah ha sounds like it should be possible, just like storage regions:

https://cloud.google.com/eventarc/docs/locations#gcs-triggers

taeold: Should we be following GCS implementation and do more granular checks (e.g. is this region in this multiregion):

https://github.com/firebase/firebase-tools/blob/3a6e0f05e544da914dcf900d6d149aeed5068d9f/src/deploy/functions/services/storage.ts#L58-L61

or is Firestore implementation much more straighforward?

src/deploy/functions/services/firestore.ts
colerogers: For RTDB, we explicitly call it out in the (docs)[https://firebase.google.com/docs/functions/beta/database-events#setting_the_function_location]. It would be a little harder to have verification since the GET (Rest API)[https://firebase.google.com/docs/reference/rest/database/database-management/rest] requires the region to be in the query. We _could_ write some custom logic with a LIST request, but on the original proposal we said that docs should mitigate this risk of deployment failure.

src/deploy/functions/services/index.ts
colerogers: Nope, this integration is a standard Eventarc one, if you look at the other services, only storage requires custom project bindings since they have a __special__ service account

src/deploy/functions/services/firestore.ts
colerogers: Yep Firestore can be multi-region triggers. To my knowledge, Firestore is more straightforward and just requires the trigger to be in the exact same region as the db. I haven't seen any documentation that function regions are limited like storage and should be fine to live in a location outside of the multi-region that Eventarc/Firestore lives

taeold: As written, we'd reject trigger in `us-central1` because that's not the same as `nam5` even though it is techincally part of `nam5`. Is that what we want?

taeold (approved):

src/deploy/functions/services/firestore.ts
colerogers: resolved offline

Issue #5333, opened 2022-12-14
New function deployed via CLI had unexpected value for EVENTARC_CLOUD_EVENT_SOURCE


willbattel:
### [REQUIRED] Environment info

**firebase-tools:** 11.18.0

**Platform:** macOS

### [REQUIRED] Test case

See below.

### [REQUIRED] Steps to reproduce

1. Deploy new functions to Firebase project **with other already-deployed functions**. Trigger does not seem to matter as this is happening with our Callable, Firestore, and Pub/Sub triggered functions.
2. View function settings in Google Cloud Console.
3. Under `Runtime environment variables`, check value for `EVENTARC_CLOUD_EVENT_SOURCE`.

### [REQUIRED] Expected behavior

I'm not 100% sure of this, as I just noticed in the console that `EVENTARC_CLOUD_EVENT_SOURCE` is pointing to a different function of ours, but I would presume this value should reflect the function. I don't know exactly what this variable is used for, but if the value is meant to be shared between functions then I would expect it to not have one (and only one) of our functions' names included in it- which isn't the function we are viewing.

### [REQUIRED] Actual behavior

`EVENTARC_CLOUD_EVENT_SOURCE` is set to `projects/MY-PROJECT-ID/locations/us-central1/functions/ANOTHER-ONE-OF-OUR-FUNCTIONS`.

This does not (currently) negatively affect us; I just wanted to report it as it seems strange and could be problematic.

<img width="578" alt="Screenshot 2022-12-14 at 1 17 33 AM" src="https://user-images.githubusercontent.com/13490339/207521094-58018a2d-a101-4689-aec4-3219df910e03.png">

pascalfiv:
I noticed the same behaviour.

This seams to be in relation to the order of how my functions are listed in the index.js

index.js:
```javascript 
....

exports.function1 = functions
  .region("europe-west6")
  .https.onCall(function1);

exports.function2 = functions
  .region("europe-west6")
  .runWith({ secrets: ["..."], timeoutSeconds: 120})
  .https.onCall(function2);

exports.function3 = functions
  .region("europe-west6")
  .runWith({ secrets: ["..."] })
  .https.onCall(function3);
```

in function1 and function2 - the name of function3 is referenced in EVENTARC_CLOUD_EVENT_SOURCE.

Have a good one :)

taeold:
Thank you for reporting the bug @willbattel - we will shipping a fix in the next release!

PR #5597, opened 2023-03-15
Fix bug where EVENTARC_CLOUD_EVENT_SOURCE environment variable was correctly set for some functions.

taeold:
Object containing environment variable was mistakenly shared for all function.

This is usually okay as CF3 doesn't allow user to define per-function environment variable. However, Firebase-defined `EVENTARC_CLOUD_EVENT_SOURCE` env var is unique per function and due to this bug, all but one function has correct value for it.

Fixes https://github.com/firebase/firebase-tools/issues/5333.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Fix bug where EVENTARC_CLOUD_EVENT_SOURCE environment variable was correctly set for some functions. (#5597)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- Fix bug where EVENTARC_CLOUD_EVENT_SOURCE environment variable was correctly set for some functions. (#5597)
```
```suggestion
- Fixes bug where EVENTARC_CLOUD_EVENT_SOURCE environment variable was incorrectly set for some functions. (#5597)
```

joehan (approved):

joehan:
Ahhh, missed that you had automerge on - mind fixing the CHANGELOG in a separate PR?

PR #5605, opened 2023-03-15
Remove call to Cloud Run API and set CPU & concurrency in GCF API instead

blidd-google:
The GCF API was recently updated to expose the concurrency and CPU fields for the Cloud Run service underlying 2nd gen Cloud Functions. Setting concurrency and CPU used to require a separate request to the Cloud Run API, which meant an extra API call per 2nd gen function deploy (for functions with non-default concurrency and CPU values). This PR eliminates the extra API call to the Cloud Run service and sets the concurrency and CPU values in the GCF API in the same API request as creating or updating the function. 

Deploying functions via the GCF API consumes the user's project quota for the Cloud Run API; reducing the number of Cloud Run API calls should lead to less quota-related failures on large function deployments.

### DIFF
--- a/src/gcp/cloudfunctionsv2.ts
+++ b/src/gcp/cloudfunctionsv2.ts
@@ -115,10 +115,12 @@ export interface ServiceConfig {
 
   timeoutSeconds?: number | null;
   availableMemory?: string | null;
+  availableCpu?: string | null;
   environmentVariables?: Record<string, string> | null;
   secretEnvironmentVariables?: SecretEnvVar[] | null;
   maxInstanceCount?: number | null;
   minInstanceCount?: number | null;
+  maxInstanceRequestConcurrency?: number | null;
   vpcConnector?: string | null;
   vpcConnectorEgressSettings?: VpcConnectorEgressSettings | null;
   ingressSettings?: IngressSettings | null;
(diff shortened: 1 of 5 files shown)


src/gcp/cloudfunctionsv2.ts
```diff
@@ -467,10 +469,26 @@ export function functionFromEndpoint(
...
   gcfFunction.serviceConfig.availableMemory = mem > 1024 ? `${mem / 1024}Gi` : `${mem}Mi`;
   proto.renameIfPresent(gcfFunction.serviceConfig, endpoint, "minInstanceCount", "minInstances");
   proto.renameIfPresent(gcfFunction.serviceConfig, endpoint, "maxInstanceCount", "maxInstances");
+  proto.renameIfPresent(
+    gcfFunction.serviceConfig,
+    endpoint,
+    "maxInstanceRequestConcurrency",
+    "concurrency"
+  );
+  proto.convertIfPresent(gcfFunction.serviceConfig, endpoint, "availableCpu", "cpu", (cpu) => {
```
inlined, line 482: Let's leave a note or guard to only do this when this is a 2nd gen function (we'll probably eventually use the v2 API to configure v1 and v2 functions)

src/gcp/cloudfunctionsv2.ts
```diff
@@ -467,10 +469,26 @@ export function functionFromEndpoint(
   );
   // Memory must be set because the default value of GCF gen 2 is Megabytes and
   // we use mebibytes
-  const mem: number = endpoint.availableMemoryMb || backend.DEFAULT_MEMORY;
+  const mem = endpoint.availableMemoryMb || backend.DEFAULT_MEMORY;
   gcfFunction.serviceConfig.availableMemory = mem > 1024 ? `${mem / 1024}Gi` : `${mem}Mi`;
   proto.renameIfPresent(gcfFunction.serviceConfig, endpoint, "minInstanceCount", "minInstances");
   proto.renameIfPresent(gcfFunction.serviceConfig, endpoint, "maxInstanceCount", "maxInstances");
+  proto.renameIfPresent(
```
inlined, line 476: The old code defaulted to 80 concurrents if concurrency was not set and CPU was >= 1. We should do the same still.

blidd-google: The `resolveCpuAndConcurrency` function in the prepare phase already provides the default — are there any edge cases in which the prepare phase does not happen, but the release phase does?

src/test/deploy/functions/release/fabricator.spec.ts
```diff
@@ -938,141 +845,6 @@ describe("Fabricator", () => {
...
-
-    it("wraps errors", async () => {
-      scheduler.createOrReplaceJob.rejects(new Error("Fail"));
-      await expect(fab.upsertScheduleV1(ep)).to.eventually.be.rejectedWith(
-        reporter.DeploymentError,
-        "upsert schedule"
-      );
-    });
-  });
-
```
taeold, line 1075: don't think you mean't to remove this test.

blidd-google: whoops, thanks for catching that!

src/gcp/cloudfunctionsv2.ts
inlined: It's fine. I may have just been programming more defensively than necessary in previous versions (or, more likely, it became clear that resolveCpuAndConcurrency had to be written to support inferencing on update after I had already written the fabricator code)

taeold (approved):

PR #5610, opened 2023-03-16
Fix Vite js-sdk auto init

chalosalvador:
### Description

- JS SDK auto init was not working for Vite projects and the browser was showing the following error when calling any firebase method like `getAuth()`
```
Uncaught FirebaseError: Firebase: Need to provide options, when not being deployed to hosting via source. (app/no-options).
    at initializeApp (api.ts:146:25)
    at getApp (api.ts:206:12)
    at getAuth (index.ts:73:44)
    at main.ts:14:14
```
- The error was that the __FIREBASE_DEFAULTS__ cookie was not  being set in the browser for Vite projects and stopping the js-sdk to be auto initialized.
- `frameworks/index.ts` was setting the "Set-Cookie" header to be transmitted only to JS files, but Vite is requesting for TS files, e.g. `http://localhost:5000/src/main.ts`. If we transmit the cookie to TS files as well, the issue is fixed.

### Scenarios Tested

- Initialize a Vite project.
- Add firebase sdk as dependency
- Call `getAuth()` from `'firebase/auth'` and confirm the firebase app is auto initialized without calling `initializeApp()`.


jamesdaniels (approved):
Add changelog entry

PR #5612, opened 2023-03-17
Rename function name in python template

cbenhagen:
### Description

2nd gen function names cannot contain capital letters or underscores. The change is not very pythonic but I do not see other solutions at the moment. Maybe someone can point me to the right repo / channels to request 2nd gen function names with underscores or to some documentation why they are not available.


cbenhagen:
Fixed in https://github.com/firebase/firebase-tools/pull/5690

Issue #5621, opened 2023-03-21
Installing firebase-tools (latest, 19.25) and running "firebase use default" fails with module p-limit not found


gunnarmein-ts:
### [REQUIRED] Environment info

npm install firebase-tools --location=global
run on a github ubuntu worker around 3:40 pm pst

**firebase-tools:**
cannot give you the version that quickly, need to modify job. we presume it was 19.25 since it was published a couple of hours ago.

Ubuntu

**Platform:**

### [REQUIRED] Test case

install firebase-tools and run firebase use default

### [REQUIRED] Steps to reproduce

npm install firebase-tools --location=global
firebase use default

### [REQUIRED] Expected behavior

not crashing

### [REQUIRED] Actual behavior

<img width="1287" alt="image" src="https://user-images.githubusercontent.com/86687645/226762459-5d8e70f5-fddf-4611-8afd-9402a30162a6.png">

We hope we are wrong but we don't see how. The script ran fine this morning. Thanks for looking!

SpencerMartel:
Same issue as#5619

joehan:
This should be resolved now in version 11.25.1 - apologies for the trouble!

Issue #5619, opened 2023-03-21
Error: Cannot find module 'p-limit'


bkendall:
### [REQUIRED] Environment info

**firebase-tools:** v11.25.0

**Platform:** macOS

### [REQUIRED] Test case

see below

### [REQUIRED] Steps to reproduce

```bash
nvm i 18 
npm i -g firebase-tools
firebase login
```

Turned out I _was_ logged in - downgrading to v11.24 worked fine.

### [REQUIRED] Expected behavior

Not a crash!

### [REQUIRED] Actual behavior

```
» firebase login
node:internal/modules/cjs/loader:1078
  throw err;
  ^

Error: Cannot find module 'p-limit'
Require stack:
- /Users/bkend/.nvm/versions/node/v18.15.0/lib/node_modules/firebase-tools/lib/database/import.js
- /Users/bkend/.nvm/versions/node/v18.15.0/lib/node_modules/firebase-tools/lib/commands/database-import.js
- /Users/bkend/.nvm/versions/node/v18.15.0/lib/node_modules/firebase-tools/lib/commands/index.js
- /Users/bkend/.nvm/versions/node/v18.15.0/lib/node_modules/firebase-tools/lib/index.js
- /Users/bkend/.nvm/versions/node/v18.15.0/lib/node_modules/firebase-tools/lib/bin/firebase.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1075:15)
    at Module._load (node:internal/modules/cjs/loader:920:27)
    at Module.require (node:internal/modules/cjs/loader:1141:19)
    at require (node:internal/modules/cjs/helpers:110:18)
    at Object.<anonymous> (/Users/bkend/.nvm/versions/node/v18.15.0/lib/node_modules/firebase-tools/lib/database/import.js:8:19)
    at Module._compile (node:internal/modules/cjs/loader:1254:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
    at Module.load (node:internal/modules/cjs/loader:1117:32)
    at Module._load (node:internal/modules/cjs/loader:958:12)
    at Module.require (node:internal/modules/cjs/loader:1141:19)
    at require (node:internal/modules/cjs/helpers:110:18)
    at Object.<anonymous> (/Users/bkend/.nvm/versions/node/v18.15.0/lib/node_modules/firebase-tools/lib/commands/database-import.js:8:18)
    at Module._compile (node:internal/modules/cjs/loader:1254:14)
[...]

PR #5625, opened 2023-03-22
Switches integration tests to package firebase-tools

joehan:
### Description
Switches integration tests to package firebase-tools (as we would before publishing a new version to NPM) instead of npm i && npm link.

This should catch issues like #5619 and #5621 in the future, before they get merged in.


scripts/clean-install.sh
```diff
@@ -0,0 +1,27 @@
...
+  rm -rf ./clean
+  rm -rf ./firebase-tools-*.tgz
+  echo "Artifacts deleted"
+}
+
+trap cleanup EXIT
+
+echo "Installing clean-publish"
+npm install -g clean-publish
+echo "Installed clean-publish."
```
bkendall, line 16: This should be unnecessary if you're using `npx` below.

scripts/clean-install.sh
```diff
@@ -0,0 +1,27 @@
...
+
+echo "Running clean-publish --without-publish, as we would before publishing to npm..."
+npx clean-publish --without-publish --before-script ./scripts/clean-shrinkwrap.sh --temp-dir clean
+echo "Packaging cleaned firebase-tools"
+cd ./clean
+PACKED=$(npm pack --pack-destination ../ | tail -n 1)
+cd ..
```
bkendall, line 23: Why not leave it in `clean` so it can automatically be cleaned up and removed in one command (above) rather than two? (I'm also guilty of leaving some packed versions in my directory, so I'm a little afraid of your cleanup command anyway)

bkendall (approved):

scripts/clean-install.sh
joehan: Good call, done

Issue #5618, opened 2023-03-21
execSync('firebase functions:config:get') is truncated to 8192 bytes


oxc:
### [REQUIRED] Environment info

**firebase-tools:** 11.18.0

**Platform:** macOS

### [REQUIRED] Test case

Create a functions config that is larger than 8192 bytes in size. Run `execSync('firebase functions:config:get')` on an affected system, and the output will be truncated to 8192 bytes.

So far, we encountered this only on macOS.

Piping this to `| cat` seems to serve as a workaround, even though that seems to contradict the nodejs docs below, which claim that pipes are asynchronous on POSIX systems.

### [REQUIRED] Steps to reproduce

```
firebase functions:config:set foo.bar="$(curl 'https://lipsum.com/feed/json?what=bytes&amount=10000')"
# will be truncated to 8192 bytes
echo "require('child_process').execSync('firebase functions:config:get', { encoding: 'utf8' })" | node -p -
```

```
# prints fully on our test systems
echo "require('child_process').execSync('firebase functions:config:get | cat', { encoding: 'utf8' })" | node -p -
```

### [REQUIRED] Expected behavior

I expect the full json object to be printed to stdout.

### [REQUIRED] Actual behavior

The output gets truncated at 8192 bytes.

### Additional information

As mentioned in the following nodejs docs, stdout might not be fully flushed when process.exit() is called.

https://nodejs.org/api/process.html#a-note-on-process-io
https://nodejs.org/api/process.html#processexitcode

> Calling process.exit() will force the process to exit as quickly as possible even if there are still asynchronous operations pending that have not yet completed fully, including I/O operations to process.stdout and process.stderr.

This seems to be what is happening here. The json object is printed with `logger.info` here:
https://github.com/firebase/firebase-tools/blob/b177da255568c14803fae9629f9492a8f22af936/src/commands/functions-config-get.ts#L33

The `process.exit()` is called here: 
https://github.com/firebase/firebase-tools/blob/b177da255568c14803fae9629f9492a8f22af936/src/command.ts#L220

edenhermelin:
any updates on this one?

joehan:
[...]

PR #5626, opened 2023-03-23
Use a safer exit when terminating CLI commands

Berlioz:
This is a shot at addressing https://github.com/firebase/firebase-tools/issues/5618, which is in brief caused by the fact that process.exit() does not guarantee that stdout/stderr are fully flushed before exiting on all platforms. 

Setting process.exitCode causes an exit (https://nodejs.org/api/process.html#process_process_exit_code), but outstanding call handlers that have not been cleaned up will cause the process to hang. Since I have no way of proving that firebase-tools does not contain any leaky handlers, I propose setting a hard timeout after which we fall back on process.exit(), while using metrics to track whether this is actually happening.

### DIFF
--- a/src/command.ts
+++ b/src/command.ts
@@ -217,7 +217,12 @@ export class Command {
               ])
             );
           }
-          process.exit();
+          process.exitCode = 0;
+          var timeout = setTimeout(function (cmd: any) {
+            track(cmd, "terminated_with_open_handlers");
+            process.exit(0);
+          }, 2000, this.name);
+          timeout.unref(); 
         })
         .catch(async (err) => {
           if (getInheritedOption(options, "json")) {
--- a/src/command.ts
+++ b/src/command.ts
@@ -66,7 +66,6 @@ export class Command {
    *
    * @example
    *   command.option("-d, --debug", "turn on debugging", false)
-   *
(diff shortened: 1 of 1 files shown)


src/command.ts
```diff
@@ -217,7 +216,16 @@ export class Command {
               ])
             );
           }
-          process.exit();
+          process.exitCode = 0;
+          const timeout = setTimeout(
+            (cmd: any) => {
```
joehan, line 221: Pretty sure this should just be string type

joehan, line 221: Think this needs to be an async function as well & await track, so that the track actually happens

src/command.ts
```diff
@@ -217,7 +216,16 @@ export class Command {
...
+              track(cmd, "terminated_with_open_handlers");
+              process.exit(0);
+            },
+            2000,
```
joehan, line 225: Consider pushing this even higher than 2s - since this is a failsafe, I'm ok with 10s even

joehan:
Few comments, but this is a very good idea

joehan (approved):
LGTM, though we should double check if those integration test failures are related - not sure if they are flakes or not.

Issue #3319, opened 2021-04-28
customAttributes is not completely preserved when exporting and importing


1betatsu:
### [REQUIRED] Environment info

**firebase-tools:**
9.10.0

**Platform:**
macOS
### [REQUIRED] Test case
Set custom claim `{"admin":true}` by any SDK or something.
Export accounts from firebase by firebase auth:export, and import the exported file by firebase auth:import.
Export accounts again, then you will find the customAttributes miss double quotations like `{admin:true}`.
 

### [REQUIRED] Steps to reproduce
Set Custom Claims by [Go SDK](https://github.com/firebase/firebase-admin-go/blob/4121c50b610f0ce408a2978cf14b306becf276d3/snippets/auth.go#L295-L308)
```
// Set admin privilege on the user corresponding to uid.
claims := map[string]interface{}{"admin": true}
err = client.SetCustomUserClaims(ctx, uid, claims)
if err != nil {
	log.Fatalf("error setting custom claims %v\n", err)
}
```
Export accounts
```
firebase auth:export ACCOUNT_FILE
```
Import the exported file.
```
firebase auth:import ACCOUNT_FILE \
    --hash-algo=YOURS \
    --hash-key=YOURS \
    --salt-separator=YOURS \
    --rounds=YOURS \
    --mem-cost=YOURS
```

### [REQUIRED] Expected behavior
Custom claims are preserved as `{"admin":true}`

### [REQUIRED] Actual behavior
Double quotations were disappeared as `{admin:true}`

thomasmburke:
Just tested this and it appears this behavior is specific to the CSV export, so if you are looking to preserve custom claims for users upon import and export via firebase-tools the JSON file type is a valid workaround as the team looks to fix buggy behavior highlighted here pertaining to CSV file types.

PR #5630, opened 2023-03-24
quote entire custom claims object and escape inner quotes

thomasmburke:
### Description

fixes [3319](https://github.com/firebase/firebase-tools/issues/3319)

### Scenarios Tested

* Export of a user with a single custom claim
* Export of a user with multiple custom claims
* Export of a user with a custom claim that contains a "
* Export of a user with a custom claim that's value is not quota, e.g. `true` or 99`
* Export of a user with no custom claims
* Export of users with a mix of the above

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,2 @@
 - Releases Cloud Firestore emulator v1.16.1, which adds support for read_time in ListCollectionIds.
+- Fixes auth:export with csv format for users with custom claims
--- a/src/accountExporter.ts
+++ b/src/accountExporter.ts
@@ -84,7 +84,10 @@ function transUserToArray(user: any): any[] {
   arr[24] = user.lastLoginAt;
   arr[25] = user.phoneNumber;
   arr[26] = user.disabled;
-  arr[27] = user.customAttributes;
+  // quote entire custom claims object and escape inner quotes with quotes
+  arr[27] = user.customAttributes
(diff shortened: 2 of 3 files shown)


thomasmburke:
### Note about tests

New test may be valid, but was a bit tricky. I was sort of suspecting `spyWrite.getCall(0).args[0]` to have a custom claim of `"{ ""customBoolean"": true, ""customString"": ""test"", ""customInt"": 99 }"`, but that wasn't the case so I ended up just comparing it to `userList` which escapes quotes with `\n` instead of `""`

### Solutions considered

Elected to stick with the default [csv-parse `escape`](https://csv.js.org/parse/options/escape/) character which is `"` so that we could properly parse claims that have quotes in the value if a user did that for some reason.

Considered [csv-parse `ignore_last_delimiters`](https://csv.js.org/parse/options/ignore_last_delimiters/) but since we know the possible forms that a claims object will come in quoting felt like the cleanest approach to avoid picking up delimiters in the claims object

CHANGELOG.md
```diff
@@ -1 +1,2 @@
 - Releases Cloud Firestore emulator v1.16.1, which adds support for read_time in ListCollectionIds.
+- Fixes auth:export with csv format for users with custom claims
```
```suggestion
- Fixes auth:export with csv format for users with custom claims. (#3319)
```

joehan (approved):
1 suggestion for the changelog, but lgtm otherwise

PR #5637, opened 2023-03-29
Adding proof-of-concept for sourcing frameworks docs here in GH

egilmorez:
As mooted in our internal doc planning, here is a start in adding source files to GitHub.

More files and a copy script must follow, but it may make sense to submit the script separately.
If it doesn't we, can add that to this PR.

Thanks!


egilmorez:
Note:  "mirrored" source also submitted for review in internal cl/520377806.

Issue #5636, opened 2023-03-28
Error emulating and deploying Python Functions


RafaelZasas:
**note** I believe the issue is in this file:
src/functions/python.ts
On lines 21 and 22. (Although I do expect it to be using the source command, it says source is not found)

### Environment info

**firebase-tools:** 11.25.1

**Platform:** Ubuntu 22.10

### Test case
Enable experimental python functions, uncomment default hello world function and run emulators or deploy.

### Expected behavior

Emulators run or deploy.

### Actual behavior

I get an error saying that the source command can not be found, however I have validated that i can activate the venv using source in my terminal.

See `firebase-debug.log`

```
[debug] [2023-03-28T17:52:38.613Z] Customer code is not Node
[debug] [2023-03-28T17:52:38.614Z] Validating python source
[debug] [2023-03-28T17:52:38.614Z] Building python source
[debug] [2023-03-28T17:52:38.615Z] Could not find functions.yaml. Must use http discovery
[debug] [2023-03-28T17:52:38.620Z] Running command with virtualenv: command=source, args=["/home/rafael/dev/protea/web/functions/venv/bin/activate","&&","python3.10","-c","\"import firebase_functions; import os; print(os.path.dirname(firebase_functions.__file__))\""]
[debug] [2023-03-28T17:52:38.632Z] Running admin server with args: ["python3.10","private/serving.py"] and env: {"FIREBASE_CONFIG":"{\"projectId\":\"protea-technology-services\",\"storageBucket\":\"protea-technology-services.appspot.com\",\"locationId\":\"us-central\"}","GCLOUD_PROJECT":"protea-technology-services","ADMIN_PORT":"8081"} in /home/rafael/dev/protea/web/functions
[debug] [2023-03-28T17:52:38.633Z] Running command with virtualenv: command=source, args=["/home/rafael/dev/protea/web/functions/venv/bin/activate","&&","python3.10","private/serving.py"]
[debug] [2023-03-28T17:52:38.642Z] stderr: /bin/sh: 1: source: not found

[debug] [2023-03-28T17:53:09.185Z] FetchError: request to http://127.0.0.1:8081/__/quitquitquit failed, reason: connect ECONNREFUSED 127.0.0.1:8081
    at ClientRequest.<anonymous> (/home/rafael/.nvm/versions/node/v16.19.1/lib/node_modules/firebase-tools/node_modules/node-fetch/lib/index.js:1491:11)
    at ClientRequest.emit (node:events:513:28)
[...]

PR #5638, opened 2023-03-29
Fix `/bin/sh :1 source not found` error by replacing `source` cmd with `.`

RafaelZasas:
### Description

Fixing issue with python-functions implementation. See Issue: https://github.com/firebase/firebase-tools/issues/5636
Was running into an issue where the `source` command was not found.
Replaced `source` with `.` and tested it locally, and it worked.

### Scenarios Tested

I first changed the file in my global node modules, ran the emulators and it worked as expected. 
I then installed the firebase-tools from my fork and it worked too.

### Sample Commands
N/A


RafaelZasas:
Hi, I need to deploy asap. Is there any update on this?

RafaelZasas:
@taeold, @joehan  It is getting slightly frustrating that nobody is communicating at all here... I have to change the source code in node modules each time I want to update firebase-tools. I've validated that the simple change of `source` to `.` works for both emulators and production- in fact it is the only way I can deploy to production at all.

I understand that python functions is still experimental and there is probably more pressing issues to deal with but I can try to help iron it out if need be- I just need someone to point me in the right direction. A little communication would be greatly appreciated.

Salakar (approved):

taeold:
Superseded by https://github.com/firebase/firebase-tools/pull/5818! thanks for the suggestion!

PR #5640, opened 2023-03-30
Vite and NPM workspace fixes

jamesdaniels:
* Better error message if a framework isn't detected
* ~~Fail detection for vite if the config can't be loaded (seeing this with SvelteKit mostly)~~
* Handle NPM workspaces for both Vite and Angular
* Error if the node_module binary isn't found
* Force vite build to production mode


src/frameworks/index.ts
```diff
@@ -233,11 +234,30 @@ function scanDependencyTree(searchingFor: string, dependencies = {}): any {
   return;
 }
 
+function getNpmRoot(cwd: string): string | undefined {
+  return spawnSync('npm', ['root'], { cwd }).stdout?.toString().trim();
+}
+
+export function getNodeModuleBin(name: string, cwd: string) {
+  const cantFindExecutable = new FirebaseError(`Cloud not find the ${name} executable.`);
```
jamesdaniels, line 242: "could not find"

src/frameworks/vite/index.ts
```diff
@@ -58,7 +53,9 @@ export async function discover(dir: string, plugin?: string, npmDependency?: str
   const anyConfigFileExists = configFilesExist.some((it) => it);
   if (!anyConfigFileExists && !findDependency("vite", { cwd: dir, depth, omitDev: false })) return;
   if (npmDependency && !additionalDep) return;
-  const { appType, publicDir: publicDirectory, plugins } = await getConfig(dir);
+  const config = await getConfig(dir);
+  if (!config) return;
+  const { appType, publicDir: publicDirectory, plugins } = config;
```
leoortizz, line 58: I think that empty lines would make this easier to read
```suggestion

  const config = await getConfig(dir);
  if (!config) return;

  const { appType, publicDir: publicDirectory, plugins } = config;
```

src/frameworks/vite/index.ts
```diff
@@ -68,20 +65,21 @@ export async function build(root: string) {
...
   await warnIfCustomBuildScript(root, name, DEFAULT_BUILD_SCRIPT);
 
-  await build({ root });
+  await build({ root, mode: "production" });
 }
 
 export async function ɵcodegenPublicDirectory(root: string, dest: string) {
   const viteConfig = await getConfig(root);
-  const viteDistPath = join(root, viteConfig.build.outDir);
+  const viteDistPath = join(root, viteConfig!.build.outDir);
```
leoortizz, line 73: shouldn't this handle undefined `viteConfig`?

```ts
  const viteConfig = await getConfig(root);
  if(!viteConfig) ..

  const viteDistPath = join(root, viteConfig.build.outDir);
```

jamesdaniels: We fail earlier in the build process, if discover fails, see the throw in frameworks/index.ts but might be worth being defensive

jamesdaniels: I found the issues with sveltekit that i was guarding here, will ditch the try/catch

src/frameworks/index.ts
```diff
@@ -233,11 +234,26 @@ function scanDependencyTree(searchingFor: string, dependencies = {}): any {
   return;
 }
 
+export function getNodeModuleBin(name: string, cwd: string) {
+  const cantFindExecutable = new FirebaseError(`Could not find the ${name} executable.`);
+  const npmBin = spawnSync(NPM_COMMAND, ["bin"], { cwd }).stdout?.toString().trim();
+  if (!npmBin) {
+    throw cantFindExecutable;
+  }
+  const path = `${join(npmBin, name)}${process.platform === "win32" ? ".cmd" : ""}`;
```
leoortizz, line 243: `.cmd` should not be needed for Windows as `cross-spawn` is being used

leoortizz (approved):
LGTM

PR #5645, opened 2023-03-31
Removing slashes and editing the linter ignore file to allow them (as…

egilmorez:
Removing slashes and editing the linter ignore file to allow them (as required by DevSite).

Thanks!

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,3 +4,4 @@ lib
 node_modules
 standalone
 templates
+/src/frameworks/docs/**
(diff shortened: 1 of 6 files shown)


.eslintignore
```diff
@@ -4,3 +4,4 @@ lib
 node_modules
 standalone
 templates
+/src/frameworks/docs/**
```
jamesdaniels, line 7: Put in `.eslintrc.js` (`ignorePatterns`) instead

egilmorez: Oh -- thanks!!

jamesdaniels:
whoops, need it in the prettier ignore too @egilmorez

PR #5667, opened 2023-04-06
Auth integration

leoortizz:


src/monospace/index.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  const { rid } = initFirebaseResponse;
+
+  // Poll for response from the user
+  const authorizedProject = await pollAuthorizedProject(rid);
+
+  if (project === authorizedProject) return;
+
+  // TODO: create a .firebaserc in user’s folder like this:
```
hsubox76, line 35: Can this return the project instead? The VS Code extension has logic to manage reading/writing `.firebaserc` and can take it from here.

src/requireAuth.ts
```diff
@@ -36,8 +37,13 @@ function getAuthClient(config: GoogleAuthOptions): GoogleAuth {
  */
 async function autoAuth(options: any, authScopes: string[]): Promise<void> {
   const client = getAuthClient({ scopes: authScopes, projectId: options.project });
-  const token = await client.getAccessToken();
-  token !== null ? apiv2.setAccessToken(token) : false;
+
+  if (process.env.MONOSPACE_DAEMON_PORT) {
+    await setupMonospace(client, options.project);
```
hsubox76, line 42: If we end up getting the short-lived 10 minute token here (I think that is the new plan going forward?) I think it needs to be refreshed for plugin usage. Could be done on a timer or could check TTL every time requireAuth is called and get a fresh one if needed (the way requireAuth is written right now though, it will bail if a `user` is found before reaching autoAuth).

src/monospace/index.ts
```diff
@@ -0,0 +1,114 @@
+import fetch from "node-fetch";
+
+import { FirebaseError } from "../error";
+
+import type { GetInitFirebaseResponse, InitFirebaseResponse } from "./interfaces";
+
+const MONOSPACE_DAEMON_PORT = process.env.MONOSPACE_DAEMON_PORT;
```
hsubox76, line 7: I won't have this variable set when the extension loads, the extension will need to grab it in a different way, then set the env variable, and then call CLI methods. I think the assignment here will be too early since it's in the global space and will happen when the code loads, before I have time to run anything. Can the assignment be done inside `initFirebase()`?

leoortizz: @hsubox76 I just changed it

src/monospace/index.ts
```diff
@@ -0,0 +1,126 @@
...
+  if (initFirebaseResponse.success === false) {
+    throw new Error(String(initFirebaseResponse.error));
+  }
+
+  const { rid } = initFirebaseResponse;
+
+  // Poll for response from the user
+  const authorizedProject = await pollAuthorizedProject(rid);
+
+  createFirebaseRc(projectRoot, authorizedProject);
```
hsubox76, line 23: Do we have a separate path where I can skip this? It looks like the only function I have control over calling is requireAuth() so it seems like you'd have to add a param to requireAuth() and pass it all the way down here.

leoortizz: > Do we have a separate path where I can skip this? It looks like the only function I have control over calling is requireAuth() so it seems like you'd have to add a param to requireAuth() and pass it all the way down here.

currently we don't. Can you somehow set an environment variable so we can detect that it's running in the plugin?

leoortizz: Added a condition to return the authorized project if it's coming from the extension(through `options.isVSCE`)

src/monospace/index.ts
leoortizz: It's returning the project now if `options.isVSCE` is true

src/options.ts
```diff
@@ -30,4 +30,7 @@ export interface BaseOptions {
 export interface Options extends BaseOptions {
   // TODO(samstern): Remove this once options is better typed
   [key: string]: unknown;
+
+  // whether it's coming from the VS Code Extension
+  isVSCE?: true;
```
leoortizz, line 35: not sure about this but thought it would be great to add as it's a new option. Didn't want to add it to `BaseOptions` as it will only be available through the extension. Happy to move it to a proper place if needed

hsubox76: This seems ok to me as long as it doesn't break anything for the CLI.

hsubox76 (approved):

src/options.ts
leoortizz: @joehan Since it will only be present in the options when coming from the extension I think it will always be `true`, otherwise it would be `undefined`. From what I see it can never be `false`, but I can change if needed

src/monospace/index.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  firebaseRc.addProjectAlias("default", authorizedProject);
+
+  return firebaseRc.save();
+}
+
+/**
+ * Whether this is a Monospace environment
+ */
+export async function isMonospaceEnv(): Promise<boolean> {
```
joehan, line 146: Why promisify this? Seems very unnecessary

leoortizz: tbh I'm not sure too, I followed @jamesdaniels suggestions, maybe he had some idea about checking this in some async way but it ended up being just checking for the environment variable

jamesdaniels: the thought originally there is that @hsubox76 could add in some override logic for the VSCE that was async. AFIAK she ended up taking a different approach

joehan (approved):

src/monospace/index.ts
```diff
@@ -0,0 +1,155 @@
...
+  // FIXME: This is not being reached as the process exits before a new call is made
+
+  // Error response: User canceled without authorizing any project
+  if (error === "USER_CANCELED") {
+    // The user hasn’t authorized any project.
+    // Display appropriate error message.
+    throw new FirebaseError("User canceled without authorizing any project");
+  }
+
+  throw new FirebaseError(`Unhandled /get-init-firebase-response error: ${error}`);
```
joehan, line 84: nit: Would be cleaner as:
```suggestion
  throw new FirebaseError(`Unhandled /get-init-firebase-response error`, {original: error});
```

joehan (approved):

src/monospace/index.ts
leoortizz: I removed the Promise from it as it doesn't need to be async anymore

src/monospace/index.ts
```diff
@@ -72,7 +72,8 @@ async function pollAuthorizedProject(rid: string): Promise<string | null> {
     return pollAuthorizedProject(rid);
   }
 
-  // FIXME: This is not being reached as the process exits before a new call is made
+  // TODO: Review this. It's not being reached as the process exits before a new
```
hsubox76, line 75: If this isn't being reached, what happens when the user cancels the dialog and how is that handled?

leoortizz: It's being handled in this block as `userResponse.success` is coming as `false`: https://github.com/firebase/firebase-tools/blob/556df35b0532bebefedc94c595b8272b92ee5a80/src/monospace/index.ts#L57-L59


in the tests I did I never received `error` as `USER_CANCELED`

hsubox76 (approved):

PR #5668, opened 2023-04-07
Breaking changes to ext:dev:* commands

joehan:
### Description
Breaking changes to the ext:dev:* family of commands, as approved in go/firex-ext-dev-breaks

- Removes the extdev flag and makes these standard commands.
- Renames ext:dev:publish -> ext:dev:upload
- Moves ext:dev:extensions:delete to be a flagged behavior of ext:dev:deprecate
- Delete the ext:dev:emulators:* commands (deprecated in v11)
- Delete the ext:sources:create command 
- Delete the ext:dev:unpublish command (deprecated)

TODO:
- Replace all options: any with real types 
- Manual testing of ext:dev:deprecate


src/commands/index.ts
```diff
@@ -86,25 +86,14 @@ export function load(client: any): any {
...
-    client.ext.dev.delete = loadCommand("ext-dev-extension-delete");
-    client.ext.dev.usage = loadCommand("ext-dev-usage");
-  }
+  client.ext.dev = {};
+  client.ext.dev.init = loadCommand("ext-dev-init");
+  client.ext.dev.list = loadCommand("ext-dev-list");
+  client.ext.dev.register = loadCommand("ext-dev-register");
+  client.ext.dev.deprecate = loadCommand("ext-dev-deprecate");
+  client.ext.dev.undeprecate = loadCommand("ext-dev-undeprecate");
+  client.ext.dev.upload = loadCommand("ext-dev-upload");
```
apascal07, line 95: I think we wanted to keep the `:publish` alias?

src/commands/ext-install.ts
```diff
@@ -42,11 +41,8 @@ marked.setOptions({
  */
 export const command = new Command("ext:install [extensionName]")
   .description(
-    "install an official extension if [extensionName] or [extensionName@version] is provided; " +
-      (experiments.isEnabled("extdev")
-        ? "install a local extension if [localPathOrUrl] or [url#root] is provided; install a published extension (not authored by Firebase) if [publisherId/extensionId] is provided "
-        : "") +
-      "or run with `-i` to see all available extensions."
+    "add a published extension to firebase.json if [publisherId/extensionId] is provided;" +
```
apascal07, line 44: Do we want to keep "published" naming here or use "uploaded extension" since it can be any extension in the registry?

src/commands/ext-dev-deprecate.ts
```diff
@@ -4,77 +4,141 @@ import * as semver from "semver";
...
+  versionPredicate: string,
+  options: ExtDevDeprecateOptions
+) {
+  if (versionPredicate) {
+    throw new FirebaseError("");
+  }
+  const extRef = refs.toExtensionRef(extensionRef);
+  utils.logLabeledWarning(
+    logPrefix,
+    "If you delete this extension, developers won't be able to install it. " +
```
apascal07, line 118: Do we still allow reconfigures if the extension is deleted?

src/commands/ext-dev-deprecate.ts
```diff
@@ -4,77 +4,141 @@ import * as semver from "semver";
...
+      "For developers who currently have this extension installed, " +
+      "it will continue to run and will appear as unpublished when " +
+      "listed in the Firebase console or Firebase CLI."
```
apascal07, line 121: I would love to add a note here suggesting deprecating the extension instead.

src/commands/ext-dev-deprecate.ts
```diff
@@ -4,77 +4,141 @@ import * as semver from "semver";
...
+  );
+  utils.logLabeledWarning(
+    logPrefix,
+    "This is a permanent action" +
+      `Once deleted, you may never use the extension name '${clc.bold(extRef)}' again.`
```
apascal07, line 126: Nit: Can you refactor this string as a single literal? I think this might currently read "permanent actionOnce deleted".

src/commands/ext-dev-deprecate.ts
joehan: Not sure, though i suspect we do not as there won't be an extension version there to reconfigure with?

src/commands/ext-dev-deprecate.ts
```diff
@@ -4,77 +4,141 @@ import * as semver from "semver";
...
+  utils.logLabeledSuccess(logPrefix, "successfully deprecated extension version(s).");
+}
+
+async function deleteExt(
+  extensionRef: refs.Ref,
+  versionPredicate: string,
+  options: ExtDevDeprecateOptions
+) {
+  if (versionPredicate) {
+    throw new FirebaseError("");
```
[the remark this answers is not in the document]
joehan: Realized that this made more sense to just let users omit a version predicate if they want to affect all versions (for both deprecate and delete)

src/commands/ext-dev-deprecate.ts
```diff
@@ -4,77 +4,141 @@ import * as semver from "semver";
...
+  message: string;
+}
 
 /**
  * Deprecate all extension versions that match the version predicate.
  */
 export const command = new Command("ext:dev:deprecate <extensionRef> <versionPredicate>")
   .description("deprecate extension versions that match the version predicate")
   .option("-m, --message <deprecationMessage>", "deprecation message")
+  .option("-d, --delete", "delete the entire extension instead of deprecating it")
```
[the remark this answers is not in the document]
joehan: Done - though that wouldn't be a breaking change regardless

src/commands/ext-dev-deprecate.ts
```diff
@@ -4,77 +4,147 @@ import * as semver from "semver";
...
+  }
+  if (!publisherId || !extensionId) {
+    throw new FirebaseError(
+      `Error parsing publisher ID and extension ID from extension reference '${clc.bold(
+        refs.toExtensionRef(extensionRef)
+      )}'. Please use the format '${clc.bold("<publisherId>/<extensionId>")}'.`
+    );
+  }
+
+  // Error out if --delete and version predicate is provided
```
apascal07, line 70: Still outdated comment here I think?

src/commands/ext-dev-deprecate.ts
```diff
@@ -4,77 +4,147 @@ import * as semver from "semver";
...
+    })
+    .filter((extensionVersion) => {
+      if (extensionVersion.state === "DEPRECATED" && !options.force) {
+        return false;
       }
-    } else {
-      throw new FirebaseError("No extension versions matched the version predicate.");
+      const message =
+        extensionVersion.state === "DEPRECATED" ? ", will overwrite deprecation message" : "";
+      // TODO: Clean up usage of extensionVersion.state
```
apascal07, line 87: Curious, why?

src/commands/ext-dev-deprecate.ts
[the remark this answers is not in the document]
apascal07: Can you set a message on `FirebaseError`?

src/commands/index.ts
apascal07: Do we need `client.ext.dev.publish = loadCommand("ext-dev-publish");` here?

src/commands/ext-dev-deprecate.ts
joehan: Clarified this comment a bit - right now, it displays either "DEPRECATED" or "PUBLISHED", and the later does not align with our current terminology.

apascal07 (approved):

PR #5674, opened 2023-04-10
Increases timeout of flaky tests to 5000ms

joehan:
### Description
Increases the timeouts of some tests that were a little flaky: https://github.com/firebase/firebase-tools/actions/runs/4658954278/jobs/8245248507#step:6:3799


jamesdaniels (approved):

jamesdaniels:
I’ve been experimenting with an alt approach to NPM namespace detection that is faster but I think this is good enough for now.

PR #5678, opened 2023-04-10
Removes experimental warning from ext-install and deploy

joehan:
### Description
Removes warning for experimental extensions from ext:install and deploy. Leaving the EAP warnings as is for now, since the language still seems relevant.


apascal07 (approved):
Nice, thanks!

Issue #4800, opened 2022-07-31
Undeleteable function with `TypeError: Cannot read properties of undefined (reading 'service')`


ky28059:
### [REQUIRED] Environment info

**firebase-tools:** 11.3.0

**Platform:** Windows

### [REQUIRED] Test case

Possibly related to #4770.

I upgraded a function from v1 to v2 and tried to deploy it, but the deploy failed with `Error: Pass the --force option to deploy functions that increase the minimum bill`. However, the function was created on Firebase without a trigger, and all subsequent attempts to redeploy, even after downgrading the function to v1, fail with `Error: An unexpected error has occurred.`. Running with --debug gives:
```
[2022-07-31T21:45:23.477Z] >>> [apiv2][query] GET https://cloudfunctions.googleapis.com/v2alpha/projects/gunnwatt/locations/-/functions [none]
[2022-07-31T21:45:24.205Z] <<< [apiv2][status] GET https://cloudfunctions.googleapis.com/v2alpha/projects/gunnwatt/locations/-/functions 200
[2022-07-31T21:45:24.206Z] <<< [apiv2][body] GET https://cloudfunctions.googleapis.com/v2alpha/projects/gunnwatt/locations/-/functions {"functions":[{"name":"projects/gunnwatt/locations/us-central1/functions/api","buildConfig":{"build":"projects/108805079121/locations/us-central1/builds/51bcc649-8ec1-4e2b-af37-5d8c1ca3b4ec","runtime":"nodejs14","entryPoint":"api","source":{"storageSource":{"bucket":"gcf-v2-sources-108805079121-us-central1","object":"api/function-source.zip"}},"sourceProvenance":{"resolvedStorageSource":{"bucket":"gcf-v2-sources-108805079121-us-central1","object":"api/function-source.zip","generation":"1659293271683162"}}},"state":"FAILED","updateTime":"2022-07-31T18:49:52.861395303Z","labels":{"deployment-tool":"cli-firebase"},"stateMessages":[{"severity":"ERROR","type":"CloudRunServiceNotFound","message":"Cloud Run service projects/gunnwatt/locations/us-central1/services/api for the function was not found. The function will not work correctly. Please redeploy."}],"environment":"GEN_2"}]}
[2022-07-31T21:45:24.368Z] TypeError: Cannot read properties of undefined (reading 'service')
    at C:\Users\Kevin\AppData\Roaming\npm\node_modules\firebase-tools\lib\deploy\functions\backend.js:197:113
[...]

PR #5684, opened 2023-04-12
Fix broken Functions CLI experience for projects with incomplete GCF 2nd Gen functions.

taeold:
The codebase assumes that "serviceConfig" property must be present in all 2nd Gen Functions in GCF. This is an incorrect assumption - 2nd Gen GCF functions may be missing a Cloud Run service if the build never succeeded.

Also refactors `backend` module a bit to avoid calling out to Cloud Run to retrieve concurrency & cpu config - it's available on the GCF response now!

Fixes https://github.com/firebase/firebase-tools/issues/4800


src/gcp/cloudfunctionsv2.ts
```diff
@@ -671,63 +669,78 @@ export function endpointFromFunction(gcfFunction: CloudFunction): backend.Endpoi
...
+        gcfFunction.serviceConfig,
+        "egressSettings",
+        "vpcConnectorEgressSettings"
+      );
+    }
+    const serviceName = gcfFunction.serviceConfig.service;
+    if (!serviceName) {
+      logger.debug(
+        "Got a v2 function without a service name." +
+          "Maybe we've migrated to using the v2 API everywhere and missed this code"
```
TheIronDev, line 735: Who is "we" and do we need to surface this to the end user?

TheIronDev: Ah, I'm seeing this was mentioned in a later line, feel free to ignore too. :+1:

TheIronDev (approved):

src/gcp/cloudfunctionsv2.ts
taeold: We is us the Firebase team. I think this error might be too edge-cases too edge-casey to need to surface to the user (and hopefully not very common).

Berlioz (approved):
LGTM

src/metaprogramming.ts
```diff
@@ -88,6 +88,22 @@ type DeepPickUnsafe<T, Keys extends string> = {
...
+ * type Foo = {
+ *     a?: string
+ *     b?: number
+ *     c?: object
+ * }
+ *
+ * type Bar = RequireKeys<Foo, "a" | "b">
+ * // Property "a" and "b" are now required.
+ */
+export type RequireKeys<T extends object, Keys extends keyof T> = T & {
```
inlined, line 103: Would you like me to make a DeepRequire like we have DeepOmit and DeepPick? Then you can say DeepRequire<Foo, "a.b" | "c">

PR #5687, opened 2023-04-12
Prevent GCF from automatically running the build script.

taeold:
GCF will now run `npm run build` script as part of function deploys:

https://cloud.google.com/functions/docs/release-notes#April_11_2023

This is a breaking change. We want to shield Cloud Functions for Firebase user from the breaking change and manually disable the behavior in CF3 deploys. We will enable the GCF feature in the next major CLI release.


TheIronDev (approved):

colerogers (approved):
yay breaking changes :sigh

PR #5688, opened 2023-04-13
Let's have Tails play with some turtles

inlined:
I want to jump in on the design doc and help design a version of discovery that answers Mono's needs, can be coupled with our own classes to implement Turtles, and uses POJO to power incremental frameworks (though I suspect a thick class is needed for each runtime).

I know there's a lot of TODOs here and there's probably a lot that doesn't work. Can you two rip this PR to shreds to help me bootstrap on the complexities I cannot ignore? Feel free to patch the PR as well.


src/tails/node.ts
```diff
@@ -0,0 +1,229 @@
...
+    dependencies: [{ name: "@angular/core" }, { name: "@angular/cli" }],
+    buildCommand: "ng build",
+    devCommand: "ng run",
+  },
+  {
+    name: "core:astro",
+    parent: "nodejs",
+    requiredFiles: [["astro.config.mjs", "astro.config.cjs", "astro.config.js", "astro.config.ts"]],
+    dependencies: [{ name: "astrojs" }],
+    canEmbed: ["core:svelte", "core:react"],
```
jamesdaniels, line 31: also core:svelte-vite, core:react-vite, since they might take a dep on a version of vite rather than leave it transitive

src/tails/node.ts
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  packageManagerInstallCommand(): string | null {
+    const packages: string[] = [];
+    if (this.pkgMgr === "yarn") {
+      packages.push("yarn");
+    }
+    if (this.lang === "ts-global") {
+      packages.push("typescript");
+    }
```
taeold, line 105: What's the purpose of this? Wouldn't typesscript be part of `devDependencies`?

src/tails/node.ts
```diff
@@ -0,0 +1,229 @@
...
+      hasYarn = await fs.exists("yarn.lock");
+    })(),
+    (async () => {
+      hasTsconfig = await fs.exists("tsconfig.json");
+    })(),
+  ]);
+  if (!pkgJsonRaw) {
+    return null;
+  }
+  // TODO: Find out why pkgJsonRaw is never. TypeScript isn't seeing the assignment
```
jamesdaniels, line 214: You should destructure your Promise.all, `const [pkgJsonRaw, hasYarn, hasTsconfig] = Promise.all([ fs.read("package.json"), fs.exists("yarn.lock"), fs.exists("tsconfig.json")]);`

inlined: That's how I tried to do it, but I think Promise.all got angry with me when it had nonhomogenous return types. I think whoever did the type annotation for Promise.all did `<Promise<T>[]>` and not `<T extends any[]>`, which would have captured tuples correctly.

inlined: Found it. I was on the right track about template inference. If I add "as const" to the end of the array it turns it into a tuple and I can destructure. Hopefully this shit gets figured out in the new version of TypeScript when you can infer const.

src/tails/node.ts
taeold:
```
npm install --global typescript
```

waaa didn't realize people actually do this!

TheIronDev:
```
npm install --global typescript
```

Er, I'm one of those people that only does this. º_º

PR #5701, opened 2023-04-17
Adding extensions to firebase init

joehan:
### Description
Adds `extensions` as an option to `firebase init`. If selected, this will set up an empty `extensions` section in `firebase.json` and create an empty `extensions` directory. If called from a directory that already has extensions set up, it will prompt users if they want to overwrite.

### Scenarios Tested
Screenshot 2023-04-17 at 11 28 07 AM
init a fresh directory
Screenshot 2023-04-17 at 11 32 26 AM
init a directory that already has extensions:
Screenshot 2023-04-17 at 11 34 21 AM

### DIFF
--- a/src/extensions/manifest.ts
+++ b/src/extensions/manifest.ts
@@ -1,10 +1,12 @@
 import * as clc from "colorette";
 import * as path from "path";
+import * as fs from 'fs-extra';
+
 import * as refs from "./refs";
 import { Config } from "../config";
 import { getExtensionSpec, ManifestInstanceSpec } from "../deploy/extensions/planner";
 import { logger } from "../logger";
-import { promptOnce } from "../prompt";
+import { confirm, promptOnce } from "../prompt";
 import { readEnvFile } from "./paramHelper";
 import { FirebaseError } from "../error";
 import * as utils from "../utils";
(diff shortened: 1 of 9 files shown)


src/extensions/manifest.ts
```diff
@@ -60,6 +62,31 @@ export async function writeToManifest(
...
+        message: `firebase.json already contains extensions:\n${currentExtensions}\nWould you like to overwrite them?`,
+        nonInteractive: options.nonInteractive,
+        force: options.force,
+        default: false,
+      }))
+    ) {
+      return;
+    }
+  }
+  config.set("extensions", {});
```
elvisun, line 87: This doesn't delete the `.env` files for the extensions. Rather than trying to overwrite it, can we just throw if something already exists to keep things simple?

huangjeff5 (approved):

elvisun (approved):

pavelgj (approved):

src/extensions/manifest.ts
joehan: This is a little awkward as is, but intentional. I'd like init to be idempotent so that it can more safely be used into scripts - init is meant as an interactive command, but some folks still try to use it non interactively.

I considered cleaning up the .env files, with a prompt, only in interactive mode, but leaving them doesn't have any ill effects.

PR #5702, opened 2023-04-17
Added secret manager API enablement on ext:install

pavelgj:


joehan (approved):
LGTM, we should cover ext:update as well.

PR #5707, opened 2023-04-18
Allow changing of the Angular configuration

jamesdaniels:
@chalosalvador Adding ng support, WDYT? Probably worth allowing the entire target to be overridden too, not just the configuration but we can tackle later.

### DIFF
--- a/src/frameworks/angular/index.ts
+++ b/src/frameworks/angular/index.ts
@@ -6,6 +6,7 @@ import { copy, pathExists } from "fs-extra";
 import { mkdir } from "fs/promises";
 
 import {
+  BUILD_TARGET_PURPOSE,
   BuildResult,
   Discovery,
   findDependency,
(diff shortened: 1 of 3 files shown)


src/frameworks/angular/index.ts
```diff
@@ -154,8 +169,26 @@ export async function ɵcodegenFunctionsDirectory(sourceDir: string, destDir: st
   return { bootstrapScript, packageJson };
 }
 
+export async function getValidBuildTargets(purpose: BUILD_TARGET_PURPOSE, dir: string) {
+  const validTargetNames = new Set(["development", "production"]);
+  const { workspaceProject, browserTarget, serverTarget, serveTarget } = await getContext(dir);
+  const { target } = ((purpose === "serve" && serveTarget) || serverTarget || browserTarget)!;
```
jamesdaniels, line 175: We need a union of browser and server, not just this order of operations

src/frameworks/angular/index.ts
```diff
@@ -227,18 +260,7 @@ async function getContext(dir: string) {
...
-      ? "production"
-      : target.defaultConfiguration;
-    if (!configuration)
-      throw new Error("No production or default configutation found for prerender.");
-    if (configuration !== "production")
-      console.warn(
-        `Using ${configuration} configuration for the prerender, we suggest adding a production target.`
-      );
-    prerenderTarget = { project, target: "prerender", configuration };
+    prerenderTarget = { project, target: "prerender", configuration: "production" };
```
jamesdaniels, line 263: turns out both development and production are valid values in Angular, so win

PR #5709, opened 2023-04-19
[Fix] Astro relatives dirs

Geoffrey-Pliez:
### Description
Fix a problem in relative paths

Indeed, the paths retrieved in the astro config begin with file:/// and the pathname begins with '/' so the relative function returns a bad value.

This generates an error in the build.

###  scenario
launch firebase deploy on Windows


Geoffrey-Pliez:
Hi @jamesdaniels
Don't forget to approve and merge this fix.
Thanks in advance. ;)

Geoffrey-Pliez:
Hi @jamesdaniels
I found a cross-platform solution with my latest commit ;)

PR #5710, opened 2023-04-19
replace obsolete npm bin command

Geoffrey-Pliez:
npm bin command was removed in npm version 9.

https://github.blog/changelog/2022-10-24-npm-v9-0-0-released/

The proposed solution is to get the root and join the .bin folder.

### DIFF
--- a/src/frameworks/index.ts
+++ b/src/frameworks/index.ts
@@ -234,10 +234,8 @@ function scanDependencyTree(searchingFor: string, dependencies = {}): any {
 
 export function getNodeModuleBin(name: string, cwd: string) {
   const cantFindExecutable = new FirebaseError(`Could not find the ${name} executable.`);
-  const npmBin = spawnSync("npm", ["bin"], { cwd }).stdout?.toString().trim();
-  if (!npmBin) {
-    throw cantFindExecutable;
-  }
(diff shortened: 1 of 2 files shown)


src/frameworks/index.ts
```diff
@@ -234,10 +234,8 @@ function scanDependencyTree(searchingFor: string, dependencies = {}): any {
 
 export function getNodeModuleBin(name: string, cwd: string) {
   const cantFindExecutable = new FirebaseError(`Could not find the ${name} executable.`);
-  const npmBin = spawnSync("npm", ["bin"], { cwd }).stdout?.toString().trim();
-  if (!npmBin) {
-    throw cantFindExecutable;
-  }
+  const npmRoot = (_a = (0, cross_spawn_1.sync)("npm", ["root"], { cwd }).stdout) === null || _a === void 0 ? void 0 : _a.toString().trim();
```
jamesdaniels, line 237: thanks for the fix, mind making the change to the typescript (pre-compiled) source?

jamesdaniels (changes requested):

src/frameworks/index.ts
jamesdaniels: i can modify the PR if you'd rather, NBD

Geoffrey-Pliez: Sorry, i was a bit too fast.
Is it good like this ?
Otherwise feel free to correct

jamesdaniels (approved):

PR #5712, opened 2023-04-19
Misc fixes for VS Code Extension

hsubox76:
### Description

Miscellaneous fixes needed for the VS Code extension to work.

- Make `prepare` public, the extension needs it to fill in options.
- cast a type in `functionsEmulator.ts`, avoid an annoying TS error in the webpack build process
- `WebFrameworks` tries to read frameworks config dirs at runtime, they aren't there in the extension, can maybe be bundled later, for now we can try/catch so it doesn't throw
- Need an `Options` type in the extension that does not have a catch-all index. Created a `BaseOptions` extended by `Options`. The rest of the CLI code should still be able to use `Options` unchanged.


joehan (approved):
Changes LGTM, looks like you just need to `npm run format`

PR #5715, opened 2023-04-20
Ensure spec.params and spec.systemParams are not empty

joehan:
### Description
Due to the conversion to and from proto wire format, required repeated fields of extensionSpec would be undefined instead of empty. This caused some unexpected behavior, including erroring out when trying to ext:install or ext:configure an extension that has no params.

This change ensures that spec.param and spec.systemParams will be [] instead of undefined, and adds some defensive checking in other places to prevent similar issues.

### DIFF
--- a/src/commands/ext-configure.ts
+++ b/src/commands/ext-configure.ts
@@ -75,9 +75,8 @@ export const command = new Command("ext:configure <extensionInstanceId>")
       instanceId,
       projectDir: config.projectDir,
     });
-
     const [immutableParams, tbdParams] = partition(
-      spec.params.concat(spec.systemParams ?? []),
+      (spec.params ?? []).concat(spec.systemParams ?? []),
       (param) => param.immutable ?? false
     );
     infoImmutableParams(immutableParams, oldParamValues);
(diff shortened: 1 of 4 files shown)


src/commands/ext-configure.ts
```diff
@@ -75,9 +75,8 @@ export const command = new Command("ext:configure <extensionInstanceId>")
       instanceId,
       projectDir: config.projectDir,
     });
-
     const [immutableParams, tbdParams] = partition(
-      spec.params.concat(spec.systemParams ?? []),
+      (spec.params ?? []).concat(spec.systemParams ?? []),
```
huangjeff5, line 79: Do we need these checks here if we are checking it at the api level?

huangjeff5 (approved):

src/commands/ext-configure.ts
joehan: Not really - this is just some extra defensive null checking in case we reintroduce this in the future accidentally. I have a low priority goal to make ext:install and ext:configure work offline for local extensions, and doing so could skip over the code in extensionsApi that ensures these are populated.

PR #5716, opened 2023-04-20
Nuxt, baseURL, and SSG fixes

jamesdaniels:
* Handle Nuxt bundled dependencies
* Add fast dev-mode to Nuxt 2
* Handle ssr true/false in a more conventional way
* Call generate when targeting ssr: false
* Import all the integrations, rather than dynamically require
* Leave it to the framework to build an SPA
* Handle Nuxt3 baseURL
* SPAs for Nuxt3 and Angular now respect baseUrl
* If the Nuxt version couldn't be determined, assume Nuxt3
* Nuxt used process.cwd in a couple places, guard against this
* Fix serveOptimizedImages + SPA


src/frameworks/index.ts
```diff
@@ -107,7 +109,7 @@ const SupportLevelWarnings = {
   ),
 };
 
-export const FIREBASE_FRAMEWORKS_VERSION = "^0.7.0";
+export const FIREBASE_FRAMEWORKS_VERSION = "0.6.1-canary.35a57cb";
```
jamesdaniels, line 112: This is in ^0.8.0 now

luc122c: Are we able to test this with experiments enabled on the Firebase CLI?

jamesdaniels: Once merged and cut, yes. If you'd like to help verify the changes in this branch, you can install it manually via `npm i -g github:firebase/firebase-tools#jamesdaniels_nuxtFixes`

src/frameworks/nuxt2/index.ts
```diff
@@ -1,139 +1,116 @@
...
+  // TODO pack this
+  await copy(configFilePath, join(destDir, basename(configFilePath)));
 
   return { packageJson: { ...packageJson }, frameworksEntry: "nuxt" };
 }
+
+export async function getDevModeHandle(cwd: string) {
+  const host = new Promise<string>((resolve) => {
+    const cli = getNodeModuleBin("nuxt", cwd);
+    const serve = spawn(cli, ["dev"], { cwd: cwd });
```
leoortizz, line 101:
```suggestion
    const serve = spawn(cli, ["dev"], { cwd });
```

leoortizz (approved):
LGTM! Left a couple nits. Tried dev mode and deployments, all working fine.

jamesdaniels:
Going to clean up my use of `!` in a couple spots, manually test the baseUrl w/SPA changes.

PR #5717, opened 2023-04-21
Go, go, gadget pintag!

jamesdaniels:
* pinTags true was not working on run rewrites
* de-dup GC API calls
* enable by default for web frameworks
* tag was missing off of run rewrites

### DIFF
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -36,8 +36,6 @@ import { generateServiceIdentity } from "../../gcp/serviceusage";
 import { applyBackendHashToBackends } from "./cache/applyHash";
 import { allEndpoints, Backend } from "./backend";
 import { assertExhaustive } from "../../functional";
-import { hostingConfig } from "../../hosting/config";
-import { getExistingRewrites } from "../hosting/convertConfig";
 
 export const EVENTARC_SOURCE_ENV = "EVENTARC_CLOUD_EVENT_SOURCE";
 function hasUserConfig(config: Record<string, unknown>): boolean {
@@ -98,35 +96,6 @@ export async function prepare(
     runtimeConfig,
     context.filters
   );
-
-  const configs = hostingConfig(options);
-  for (const config of configs) {
-    for (const rewrite of config.rewrites || []) {
-      let serviceId: string | undefined;
-      if (
-        "function" in rewrite &&
-        typeof rewrite.function === "object" &&
-        rewrite.function.pinTag
-      ) {
-        const backends = await backend.existingBackend(context);
-        const region = ("region" in rewrite && rewrite.region) || "us-central1";
-        serviceId = backends.endpoints[region][rewrite.function.functionId]?.runServiceId;
-      }
-      if ("run" in rewrite && (rewrite.run.pinTag || rewrite.run.tag)) {
-        serviceId = rewrite.run.serviceId;
-      }
-      if (!serviceId) continue;
-      const existingRewrites = await getExistingRewrites(context.projectId, config.site);
-      const matchingRewrite = existingRewrites.find(
-        (it) =>
-          ("glob" in it && "source" in rewrite && it.glob === rewrite.source) ||
-          ("regex" in it && "regex" in rewrite && it.regex === rewrite.regex)
-      );
-      if (!(matchingRewrite && "run" in matchingRewrite)) continue;
-      if (matchingRewrite.run.tag || matchingRewrite.run.serviceId !== serviceId) continue;
-      throw new FirebaseError("ya need to enable pintags on prod yo!");
-    }
-  }
 
   // == Phase 2. Resolve build to backend.
   const codebaseUsesEnvs: string[] = [];
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -36,6 +36,9 @@ import { generateServiceIdentity } from "../../gcp/serviceusage";
 import { applyBackendHashToBackends } from "./cache/applyHash";
 import { allEndpoints, Backend } from "./backend";
 import { assertExhaustive } from "../../functional";
+import { hostingConfig } from "../../hosting/config";
+import { getExistingRunRewrites } from "../hosting/convertConfig";
+import { ensureTargeted } from "../../functions/ensureTargeted";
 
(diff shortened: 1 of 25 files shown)


src/deploy/index.ts
```diff
@@ -60,19 +62,32 @@ export const deploy = async function (
...
+      for (const config of configs) {
+        for (const rewrite of config.rewrites || []) {
+          let serviceIdToPin: string|undefined;
+          if ("function" in rewrite && typeof rewrite.function === "object" && rewrite.function.pinTag) {
+            serviceIdToPin = rewrite.function.functionId;
+          } else if ("run" in rewrite && rewrite.run.pinTag) {
+            serviceIdToPin = rewrite.run.serviceId;
+          }
+          if (serviceIdToPin) {
+            await requirePermissions(TARGET_PERMISSIONS["functions"]);
```
jamesdaniels, line 82: pull this into the memo

jamesdaniels:
FYI @inlined have a work in progress on enablement, PTAL at the logic. Will clean up, get some test coverage, and bash. 

After experimenting I found I had to do this ASAP or I couldn't do things like force the "only" and "targets".

src/deploy/functions/prepare.ts
```diff
@@ -97,6 +100,44 @@ export async function prepare(
...
+        const codespace = Object.keys(wantBuilds).find(codespace => {
+          return Object.keys(wantBuilds[codespace].endpoints).includes(serviceIdToPin!);
+        });
+        if (!codespace) throw new FirebaseError(`Can't find codespace exporting ${serviceIdToPin}`);
+        const { region: regions, platform } = wantBuilds[codespace].endpoints[serviceIdToPin];
+        if (platform !== "gcfv2") throw new FirebaseError("pintag reqs gen2 yo!");
+        if (!regions) throw new Error("huh...");
+        if (context.hostingChannel) {
+          const existingRewrites = await getExistingRunRewrites(context.projectId, config.site);
+          const matchingRewrite = existingRewrites.find((it) => it.serviceId === serviceIdToPin && regions.includes(it.region));
```
jamesdaniels, line 128: edge case but could be multiple, test for that

src/deploy/functions/prepare.ts
```diff
@@ -97,6 +100,44 @@ export async function prepare(
     context.filters
   );
 
+  const configs = hostingConfig(options);
+  for (const config of configs) {
+    for (const rewrite of config.rewrites || []) {
+      let serviceIdToPin: string | undefined;
+      if (
+        "function" in rewrite &&
```
jamesdaniels, line 108: Dry up these tests, use them in multiple spots

inlined:
In our architecture so far, Web Frameworks is Hosting orchestrating Functions. I'd like to keep the separation of the three as much as possible.

It looks like this PR is solely targeting the onboarding case. If that's the case, I think we should try a much more surgical approach.

1. In src/deploy/hosting.ts, detect that we need to guardPreviewTagging if firebase.json has pinTags = true and hostingChannel is not live
2. In guardPreviewTragging we detect whether live has a rewrite for the same path with a run rewrite that is not tagged
3. If firebase.json indicates a run rewrite and we have a match, throw
4. If firebase.json indicates a functions rewrite, get the run service from existingBackend(). If the function exists and the service Id matches, throw
5. If the function exists and the service ID doesn't match, don't throw
6. If the function doesn't exist, then live is broken and we don't need to protect the customer from anything. Don't throw. FWIW this can only happen if the customer deleted the service since their last deploy. They're an edge case and they're already holding the gun to their foot.

src/deploy/functions/prepare.ts
```diff
@@ -37,8 +37,7 @@ import { applyBackendHashToBackends } from "./cache/applyHash";
 import { allEndpoints, Backend } from "./backend";
 import { assertExhaustive } from "../../functional";
 import { hostingConfig } from "../../hosting/config";
-import { getExistingRunRewrites } from "../hosting/convertConfig";
-import { ensureTargeted } from "../../functions/ensureTargeted";
+import { getExistingRewrites } from "../hosting/convertConfig";
```
inlined, line 40: I specifically think we should try to solve this problem in just deploy/hosting/prepare.ts and nowhere else. I don't understand any need to put this in functions code.

jamesdaniels: Agreed, now that I don't need to detect the codebase, just haven't moved it yet

src/deploy/hosting/convertConfig.ts
```diff
@@ -89,22 +89,16 @@ export function findEndpointForRewrite(
   };
 }
 
-const existingRunRewritesMemo = new Map<string, Promise<api.RunRewrite[]>>();
-export function getExistingRunRewrites(
-  projectId: string,
-  site: string,
-  channelName: string = "live"
-) {
+const existingRewritesMemo = new Map<string, Promise<api.Rewrite[]>>();
```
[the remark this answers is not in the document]
jamesdaniels: No need ATM, in a prior iteration I was keying off something else. I'll simplify

inlined:
This is time sensitive and I seem to be the only person with sufficient context of both functions and hosting codebases (and tag pinning) so I tried my hand at James' general idea. See #5731

src/firebaseConfig.ts
```diff
@@ -65,6 +65,7 @@ export type RunRewrite = {
     serviceId: string;
     region?: string;
     pinTag?: boolean;
+    tag?: string;
```
jamesdaniels, line 68: I might be doing the wrong thing here, clarify

src/deploy/hosting/convertConfig.ts
```diff
@@ -232,16 +232,20 @@ export async function convertConfig(
...
           ...rewrite.run,
         },
       };
-      if (apiRewrite.run.tag) {
+      if (apiRewrite.run.tag || rewrite.run.pinTag) {
         experiments.assertEnabled("pintags", "pin to a run service revision");
       }
+      if (rewrite.run.pinTag) {
+        delete apiRewrite.run.pinTag;
+        apiRewrite.run.tag ||= runTags.TODO_TAG_NAME;
```
jamesdaniels, line 247: I might be doing the wrong thing here, clarify

jamesdaniels:
Closing in favor of #5742

PR #5722, opened 2023-04-24
Cache project metadata calls when computing the default service account

andrew-teirney:
### Description

When trying to deploy a single function from a firebase project with hundreds of functions it was found that it took a while before the remote deployment started. This was tracked down to lots of calls to get the firebase project metadata to get the default service account for each backend endpoint, this was evident by looking at the `firebase-debug.log` whilst waiting for things to progress.

This changeset looks to add caching to the getFirebaseProject call when computing the defaultServiceAccount such that for a given project only one call is ever made.

This greatly reduces the number of calls to the project metadata endpoint whilst at the same time reducing the amount of time before the actual deployment starts.

I have assumed that changes in the metadata across calls to get the default service account is not possible, as I believe the data that is read from the metadata in this context are immutable. I have also assumed that if the call fails for whatever reason, then it would just treat all requests for metadata as failed. There might be some more elegant ways of doing this, but I am happy to rerun if a failure is encountered given the time savings these changes make.

### DIFF
--- a/src/deploy/functions/ensure.ts
+++ b/src/deploy/functions/ensure.ts
@@ -1,5 +1,7 @@
 import * as clc from "colorette";
 
+import type { FirebaseProjectMetadata } from "../../types/project";
+
 import { ensure } from "../../ensureApiEnabled";
 import { FirebaseError, isBillingError } from "../../error";
 import { logLabeledBullet, logLabeledSuccess } from "../../utils";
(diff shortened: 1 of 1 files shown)


joehan:
@andrew-teirney Thanks for the contribution!  I think this is a very nice efficiency improvement and LGTM. I'm going to defer to @taeold for approval before merging though, since they know a bit more about this part of the codebase

joehan (approved):

taeold (approved):
This is an excellent contribution! Thank you so much!

ankit-gautam23 (approved):
LGTM

PR #5725, opened 2023-04-24
Adds new extensions TOS api

joehan:
### Description
Adds methods to interact with new extensions TOS private API. In follow up PRs, I'll add calls to these at the appropriate  places in the flow.

### DIFF
--- a/src/extensions/tos.ts
+++ b/src/extensions/tos.ts
@@ -1,0 +1,116 @@
+import { Client } from "../apiv2";
+import { extensionsTOSOrigin } from "../api";
+import { logger } from "../logger";
+import { confirm } from "../prompt";
+import { FirebaseError } from "../error";
+
+const VERSION = "v1";
+const extensionsTosUrl = (tos: string) => `https://firebase.google.com/terms/extensions/${tos}`;
+
+export interface TOS {
+  name: string;
+  lastAcceptedVersion: string;
+  lastAcceptedTime: string;
+  latestTosVersion: string;
+}
+export type PublisherTOS = TOS;
+export type AppDevTOS = TOS;
+
+const apiClient = new Client({ urlPrefix: extensionsTOSOrigin, apiVersion: VERSION });
+
+export async function getAppDeveloperTOSStatus(projectId: string): Promise<AppDevTOS> {
+  const res = await apiClient.get<AppDevTOS>(`/projects/${projectId}/appdevtos`);
+  return res.body;
+}
+
+export async function acceptAppDeveloperTOS(
+  projectId: string,
+  tosVersion: string,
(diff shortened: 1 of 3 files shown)


src/extensions/tos.ts
```diff
@@ -0,0 +1,120 @@
...
+  if (currentAcceptance.lastAcceptedVersion === currentAcceptance.latestTosVersion) {
+    logger.debug(
+      `Latest version of app developer TOS is ${currentAcceptance.lastAcceptedVersion}, already accepted.`
+    );
+    return currentAcceptance;
+  } else {
+    // Display link to TOS, prompt for acceptance
+    const tosLink = extensionsTosUrl("appdev");
+    logger.info(
+      `To continue, you must accept the [Extensions app developer terms of service](${tosLink})`
```
apascal07, line 108: Possibly nit: Capitalize and use full name "Firebase Extensions App Developer Terms of Service" and "Firebase Extensions Publisher Terms of Service" I think.

src/extensions/tos.ts
```diff
@@ -0,0 +1,120 @@
...
+    );
+    if (
+      await confirm({
+        ...options,
+        message: "Do you accept the Extensions publisher terms of service?",
+      })
+    ) {
+      return acceptPublisherTOS(projectId, currentAcceptance.latestTosVersion);
+    }
+    throw new FirebaseError("Command terminated becuase publisher TOS was not accepted");
```
apascal07, line 86: Suggestion: "You must accept the terms of service to continue."

src/extensions/tos.ts
```diff
@@ -0,0 +1,120 @@
+import { Client } from "../apiv2";
+import { extensionsTOSOrigin } from "../api";
+import { logger } from "../logger";
+import { confirm } from "../prompt";
+import { FirebaseError } from "../error";
+
+const VERSION = "v1";
+const extensionsTosUrl = (tos: string) => `https://firebase.google.com/terms/extensions/${tos}`;
```
apascal07, line 8: Suggestion: have a constant for each -- IMO it's nice to be able to search constant strings if they need replacing, where they're used, etc.

src/extensions/tos.ts
joehan: Was assuming that this is another case of 'Not Everything Needs To Be In Caps Case'

apascal07: I think this is one case that it's important To Be In Caps since it refers to the legal document with the same title, not some abstract terms.

src/extensions/tos.ts
joehan: Done in the follow up PR, good call

PR #5726, opened 2023-04-24
Allow user to configure `database:import` chunk size and concurrency

tohhsinpei:
### Description
[b/278767412](http://b/278767412)
API proposal: [go/firebase-database-import](http://go/firebase-database-import)

### Sample Commands
```
firebase database:import / data.json --chunk-size 20 --concurrency 10
```

### DIFF
--- a/src/database/import.ts
+++ b/src/database/import.ts
@@ -9,9 +9,6 @@ import { Client, ClientResponse } from "../apiv2";
 import { FirebaseError } from "../error";
 import * as pLimit from "p-limit";
 
-const MAX_CHUNK_SIZE = 1024 * 1024 * 10;
-const CONCURRENCY_LIMIT = 5;
-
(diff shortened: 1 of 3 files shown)


src/database/import.ts
```diff
@@ -31,15 +28,17 @@ type ChunkedData = {
...
-  private limit = pLimit(CONCURRENCY_LIMIT);
+  private limit: pLimit.Limit;
 
   constructor(
     private dbUrl: URL,
     private inStream: stream.Readable,
     private dataPath: string,
-    private chunkSize = MAX_CHUNK_SIZE
+    private chunkBytes: number,
+    concurrency: number
```
joehan, line 38: Any reason this can't be private as well?

joehan (approved):

src/database/import.ts
tohhsinpei: It can be; I just don't need it outside of the constructor.

PR #5730, opened 2023-04-25
Upgrade tooling to node 18, remove explicit support for node 14

joehan:
### Description
Upgrades tooling to node 18, removes explicit support for node 14


joehan:
Spent a bunch of time debugging the import-export integration test, to no avail. I'm increasingly convinced that we need to redesign it to be less fragile - it currently spins up and tears down the emulators 6 times, and has a hard coded wait between each. However, the test cases are consistently overlapping and causing it to fail.

bkendall (approved):

Igloczek:
It would be really helpful to me, and probably to others, if the discontinuation of Node.js 14 support was mentioned in the 12.0.0 release notes. Thanks!

bkendall:
@Igloczek good call. Done. 🙂

PR #5731, opened 2023-04-25
Block deploys to preview channels if would break prod

inlined:
Thanks to @jamesdaniels clever idea, we are able to dramatically cut scope by erroring if deploying to a preview would break prod rather than simply trying to modify prod while deploying to a preview.

This is an alternative approach to #5717 that is intended to be more surgical.

### DIFF
--- a/src/deploy/hosting/prepare.ts
+++ b/src/deploy/hosting/prepare.ts
@@ -108,7 +108,6 @@ export async function unsafePins(
   }
 
   if (!Object.keys(targetTaggedRewrites).length) {
-    console.error("No targets with tags");
     return [];
   }
 
--- a/src/deploy/hosting/prepare.ts
+++ b/src/deploy/hosting/prepare.ts
@@ -75,7 +75,11 @@ function rewriteTarget(source: HostingSource): string {
 }
 
 /**
- *
+ * Returns a list of rewrite targets that would break in prod if deployed.
+ * People use tag pinning so that they can deploy to preview channels without
+ * modifying production. This assumption is violated if the live channel isn't
+ * actually pinned. This method returns "unsafe" pins, where we're deploying to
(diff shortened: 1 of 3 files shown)


src/deploy/hosting/prepare.ts
```diff
@@ -52,3 +61,70 @@ export async function prepare(context: Context, options: HostingOptions & Option
...
+      // This function is new. It can't be pinned elsewhere
+      if (!endpoint) {
+        continue;
+      }
+      targetTaggedRewrites[target] = `${region}/${endpoint.runServiceId || endpoint.id}`;
+    }
+  }
+
+  if (!Object.keys(targetTaggedRewrites).length) {
+    console.error("No targets with tags");
```
jamesdaniels, line 111: No need to error right?

jamesdaniels (approved):
LGTM, doing some verification now.

src/deploy/index.ts
```diff
@@ -67,11 +67,6 @@ export const deploy = async function (
       await prepareFrameworks(targetNames, context, options);
       const nowTargetsFunctions = targetNames.includes("functions");
       if (nowTargetsFunctions && !usedToTargetFunctions) {
-        if (context.hostingChannel && !experiments.isEnabled("pintags")) {
```
jamesdaniels, line 70: Let's revert this change for now and capture enabling this experiment by default in a follow on, SG?

src/deploy/hosting/prepare.ts
```diff
@@ -52,3 +61,69 @@ export async function prepare(context: Context, options: HostingOptions & Option
...
+    return source.source;
+  } else if ("regex" in source) {
+    return source.regex;
+  } else {
+    assertExhaustive(source);
+  }
+}
+
+/**
+ *
```
taeold, line 78: do you mind adding a short summary of why certain pins are considered unsafe?

PR #5733, opened 2023-04-25
Frameworks package.json fixes

jamesdaniels:
* Remove scripts from package.json
* Convert bundled deps to deps
* Make engine picking less bad

### DIFF
--- a/src/frameworks/index.ts
+++ b/src/frameworks/index.ts
@@ -30,6 +30,7 @@ import { ensureTargeted } from "../functions/ensureTargeted";
 import { implicitInit } from "../hosting/implicitInit";
 import { fileExistsSync } from "../fsutils";
 import { logWarning } from "../utils";
+import { conjoinOptions } from "./utils";
 
 // Use "true &&"" to keep typescript from compiling this file and rewriting
 // the import statement into a require
@@ -121,12 +122,6 @@ export const ALLOWED_SSR_REGIONS = [
   { name: "europe-west1 (Belgium)", value: "europe-west1" },
   { name: "asia-east1 (Taiwan)", value: "asia-east1" },
 ];
-
-function conjoinOptions(opts: any[], conjunction: string = "and", separator: string = ","): string {
-  if (opts.length === 1) return opts[0].toString();
-  const lastElement = opts.slice(-1)[0];
-  return `${opts.slice(0, -1).join(`${separator} `)}${separator} ${conjunction} ${lastElement}`;
-}
 
 const DEFAULT_FIND_DEP_OPTIONS: FindDepOptions = {
   cwd: process.cwd(),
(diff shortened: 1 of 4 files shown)


src/frameworks/index.ts
```diff
@@ -120,6 +122,12 @@ export const ALLOWED_SSR_REGIONS = [
   { name: "asia-east1 (Taiwan)", value: "asia-east1" },
 ];
 
+function conjoinOptions(opts: any[], conjunction: string, seperator: string = ","): string {
+  if (opts.length === 1) return opts[0].toString();
+  const lastElement = opts.slice(-1)[0];
+  return `${opts.slice(0, -1).join(`${seperator} `)}, ${conjunction} ${lastElement}`;
```
jamesdaniels, line 128: use separator for this last comma

leoortizz, line 128: just a typo 😛
```suggestion
function conjoinOptions(opts: any[], conjunction: string, separator: string = ","): string {
  if (opts.length === 1) return opts[0].toString();
  const lastElement = opts.slice(-1)[0];
  return `${opts.slice(0, -1).join(`${separator} `)}, ${conjunction} ${lastElement}`;
```

src/frameworks/index.ts
```diff
@@ -346,7 +354,7 @@ export async function prepareFrameworks(
     }
     const ssrRegion = frameworksBackend?.region ?? DEFAULT_REGION;
     if (!allowedRegionsValues.includes(ssrRegion)) {
-      const validRegions = allowedRegionsValues.join(", ");
+      const validRegions = conjoinOptions(allowedRegionsValues, "or");
```
chalosalvador, line 357:
```suggestion
      const validRegions = conjoinOptions(allowedRegionsValues, "and");
```

src/frameworks/index.ts
```diff
@@ -542,7 +550,20 @@ export async function prepareFrameworks(
...
+        logWarning(
+          `This integration expects Node version ${conjoinOptions(
+            VALID_ENGINES.node,
+            "or"
+          )}. You're running version ${NODE_VERSION}, problems may be encountered.`
+        );
+      }
+      packageJson.engines.node ||= engine.toString();
+      delete packageJson.scripts;
+      delete packageJson.devDependencies;
```
jamesdaniels, line 565: we're already removing devDeps above, we should consolidate

src/frameworks/index.ts
```diff
@@ -542,7 +550,20 @@ export async function prepareFrameworks(
...
+      delete packageJson.bundledDependencies;
```
jamesdaniels, line 566: warn and convert to deps

leoortizz (approved):
LGTM. I added tests for `conjoinOptions` and fixed some issues with the returned string

PR #5735, opened 2023-04-25
Do multi-path `PATCH` instead of `PUT` in `database:import`

tohhsinpei:
### Description
[b/278767412](http://b/278767412)

[Context](https://github.com/firebase/firebase-tools/pull/5396#pullrequestreview-1243356339)


src/database/import.ts
```diff
@@ -115,19 +127,20 @@ export default class DatabaseImporter {
...
 
-  private writeChunk(chunk: Data): Promise<ClientResponse<JsonType>> {
+  private doWriteBatch(batch: Batch): Promise<ClientResponse<JsonType>> {
     const doRequest = (): Promise<ClientResponse<JsonType>> => {
       return this.client.request({
-        method: "PUT",
-        path: chunk.pathname + ".json",
-        body: JSON.stringify(chunk.json),
+        method: "PATCH",
+        path: "/.json",
```
fredzqm, line 142: If there is a long common ancestor, we should extract that into `path`, to reduce bytes in the payload.

I think the previous abstraction `chunk: Data` is still good.

src/database/import.ts
```diff
@@ -190,3 +203,35 @@ export default class DatabaseImporter {
...
+
+  _transform(chunk: Data, _: BufferEncoding, callback: stream.TransformCallback) {
+    const chunkSize = JSON.stringify(chunk.json).length;
+    if (this.size + chunkSize > this.maxSize) {
+      this.push(this.batch);
+      this.batch = {};
+      this.size = 0;
+    }
+    this.batch[chunk.pathname] = chunk.json;
+    this.size += chunkSize;
```
fredzqm, line 227: `chunkSize` don't include `chunk.pathname`.
I worry that this can still hit our 10MB incoming request limit. Let's include `pathname.length` as well to be conservative.

10MB may not be the most optimal chunk size. Can we set a lower default number? 1MB is already quite sizable.

src/database/import.ts
```diff
@@ -190,3 +203,35 @@ export default class DatabaseImporter {
...
+  private size = 0;
+
+  constructor(private maxSize: number, opts?: stream.TransformOptions) {
+    super({ ...opts, objectMode: true });
+  }
+
+  _transform(chunk: Data, _: BufferEncoding, callback: stream.TransformCallback) {
+    const chunkSize = JSON.stringify(chunk.json).length;
+    if (this.size + chunkSize > this.maxSize) {
+      this.push(this.batch);
```
fredzqm, line 222: Can you convert `Batch` into a `Data`?
Find the longest common ancestor of all data.

fredzqm (changes requested):

src/database/import.ts
tohhsinpei: The casework is a bit gnarly... For example, we can PATCH the payload `{"a": "foo"}` to `/.json` but we can't PATCH `"foo"` to `/a.json`. Also can't PATCH an array (have to convert to object).

tohhsinpei: Gave it a shot anyway, lmk what you think.

src/database/import.ts
tohhsinpei: Done.

src/database/import.ts
tohhsinpei: See above.

src/database/import.ts
```diff
@@ -190,3 +213,63 @@ export default class DatabaseImporter {
...
+class BatchChunks extends stream.Transform {
+  private batch: Data = { json: {}, pathname: "" };
+  private size = 0;
+
+  constructor(private maxSize: number, opts?: stream.TransformOptions) {
+    super({ ...opts, objectMode: true });
+  }
+
+  _transform(chunk: Data, _: BufferEncoding, callback: stream.TransformCallback) {
+    const chunkSize = JSON.stringify(chunk.json).length + chunk.pathname.length;
```
tohhsinpei, line 230: Don't love this extra call to `JSON.stringify`, should probably make it part of `Data`?

src/database/import.ts
```diff
@@ -190,3 +213,63 @@ export default class DatabaseImporter {
...
+      this.push(this.batch);
+      this.batch = { json: {}, pathname: "" };
+      this.size = 0;
+    }
+    if (this.size === 0) {
+      this.batch = chunk;
+    } else {
+      const newPathname = this._findLongestCommonPrefix(chunk.pathname, this.batch.pathname);
+      const batchKey = this.batch.pathname.substring(newPathname.length + 1); // +1 to trim leading slash
+      const chunkKey = chunk.pathname.substring(newPathname.length + 1);
```
fredzqm, line 241: The iterative approach would work. Not sure if `String.substring` trigger a copy. In that case, it would be O(n^2).

Might be more safe if we find the common prefix of `[]Data` in the end. Less objects allocated.

src/commands/database-import.ts
```diff
@@ -25,7 +25,7 @@ interface DatabaseImportOptions extends Options {
   concurrency?: string;
 }
 
-const MAX_CHUNK_SIZE_MB = 10;
+const MAX_CHUNK_SIZE_MB = 1;
```
fredzqm, line 28: Could you update the cli args parsing as well?

https://github.com/firebase/firebase-tools/blob/a48cb559d6e7525da5d16b1761802e41fb16368b/src/commands/database-import.ts#L50

src/database/import.ts
tohhsinpei: Not sure I understand your comment. Do you mean you want to do all the batching, then find the prefix for each batch?

src/database/import.ts
tohhsinpei: Ok, I got it to count while chunking.

src/commands/database-import.ts
tohhsinpei: Done.

fredzqm: Also here. 
When someone supplied an invalid int, we fail with invalid flag. Is `parseInt(..., 10)` providing a default value?

https://github.com/firebase/firebase-tools/pull/5735/files/a48cb559d6e7525da5d16b1761802e41fb16368b..a169285aa30c0b714e2829d579427f8e74dda0e7#diff-0fb4f1de717a5e0cb768aa8c8feaefeea2c9c8376619eecc8d13f7cc3cacfec2R65

src/database/import.ts
fredzqm: Yeah. The chunk size can be a rough estimate. Doesn't need to be accurate.

I worry a bit that `substring` on each chunk will create a copy and make it O(n^2).
- Collect batches until the estimated size reach the payload size.
- Compact batches, find common prefix, send them out.

tohhsinpei: Done.

fredzqm (approved):

src/commands/database-import.ts
```diff
@@ -47,7 +47,7 @@ export const command = new Command("database:import <path> [infile]")
     "--filter <dataPath>",
     "import only data at this path in the JSON file (if omitted, import entire file)"
   )
-  .option("--chunk-size <mb>", "max chunk size in megabytes, default to 10 MB")
+  .option("--chunk-size <mb>", "max chunk size in megabytes, default to 1 MB")
   .option("--concurrency <val>", "concurrency limit, default to 5")
   .before(requirePermissions, ["firebasedatabase.instances.update"])
   .before(requireDatabaseInstance)
```
fredzqm, line 53: In line 65, 
```
    const chunkMegabytes = options.chunkSize ? parseInt(options.chunkSize, 1) : MAX_CHUNK_SIZE_MB;
```

tohhsinpei: Good catch, thanks!

joehan: That is correct as is - the second arg of parseInt is the radix, not the default. Users will pass in the chunkSize arg in base 10, so it should stay as is

tohhsinpei: That's the radix. Should be base 10.

joehan (approved):

Issue #5734, opened 2023-04-25
Firebase Emulators download failed, expected 64907894 bytes but got 64723779


markmxr:
### [REQUIRED] Environment info

**firebase-tools:** 11.29.0

**Platform:** macOS, Ubuntu

### [REQUIRED] Test case

None

### [REQUIRED] Steps to reproduce

In a project with Firebase Emulators set up,
1) npm uninstall firebase-tools
1 optional) npm uninstall -g firebase-tools
2) npm install firebase-tools
3) firebase emulators:start

### [REQUIRED] Expected behavior

Firebase Emulators downloads and boots up as normally.

### [REQUIRED] Actual behavior

Firebase Emulators appears to download to 100%, but emulators are aborted with the error
`Error: download failed, expected 64907894 bytes but got 64723779`

avilao:
I have the same issue.

DavidWeiss2:
I have the same issue. node 16.19.1

Lerobindesbois:
same problem :-( v18.16.0. and Firebase CLI 11.29.0 


![Capture d’écran, le 2023-04-25 à 19 58 32](https://user-images.githubusercontent.com/69640665/234432999-d3747797-8677-4355-b03b-ff188c8b7b0f.png)

rhys-mondo:
Ran into the same issue with v11.29.0

For now, have reverted back to using my previous version (v11.24.1) which is working again 👍

endrureza:
I have the same problem, here is the debug log

```
[debug] [2023-04-26T00:56:27.915Z] ----------------------------------------------------------------------
[debug] [2023-04-26T00:56:27.916Z] Command:       /opt/homebrew/Cellar/node@18/18.16.0/bin/node /Users/tabsquare/Library/pnpm/global/5/node_modules/firebase-tools/lib/bin/firebase.js setup:emulators:firestore --debug
[debug] [2023-04-26T00:56:27.917Z] CLI Version:   11.29.0
[debug] [2023-04-26T00:56:27.917Z] Platform:      darwin
[debug] [2023-04-26T00:56:27.917Z] Node Version:  v18.16.0
[debug] [2023-04-26T00:56:27.918Z] Time:          Wed Apr 26 2023 07:56:27 GMT+0700 (Western Indonesia Time)
[debug] [2023-04-26T00:56:27.918Z] ----------------------------------------------------------------------
[debug] 
[...]

PR #5737, opened 2023-04-26
Updated expectedSize and expectedChecksum

aalej:
### Description

Proposed fix for #5734.

Looks like the cause is the incorrect expectedSize and expectedChecksum [here](https://github.com/firebase/firebase-tools/blob/0d2acc5f8e17c8baf0126d2daf419285fa45e731/src/emulator/downloadableEmulators.ts#L37). `expectedSize` is 64723779 and `expectedChecksum` is 963b46bce1a2f8ca5dfc504ab7b2c901.

### Scenarios Tested

1. Deleting the `.cache/firebase/emulators/cloud-firestore-emulator-v1.17.2.jar`. 
2. Run `firebase emulators:start --project demo-proj --only firestore` to re-install the emulator.
3. Firestore emulator is downloaded as intended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,1 @@
-- Fixes an issue where downloading Firestore emulator fails due to incorrect expectedSize and expectedChecksum (#5734).
+- Fixes an issue where downloading Firestore emulator fails due to incorrect expectedSize and expectedChecksum. (#5734)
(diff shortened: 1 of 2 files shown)


trakhimenok:
I wonder if you can put some check in place to prevent this in future?

joehan:
This was an odd one - we have tests in place to catch this, but we changed the JAR after running them.  Ended up fixing this issue by putting the correct JAR in storage.

PR #5741, opened 2023-04-26
deploying --only hosting also deploys tagged functions

inlined:
DO NOT MERGE UNTIL API COUNCIL GIVES THE LGTM IN THE DISCUSSION THREAD

Surgical implementation to add any functions specifically tagged to the deploy targets even when --only hosting is specified.

Tested manually in addition to using unit tests

### DIFF
--- a/src/deploy/hosting/prepare.ts
+++ b/src/deploy/hosting/prepare.ts
@@ -10,11 +10,9 @@ import { track } from "../../track";
 import * as utils from "../../utils";
 import { HostingSource } from "../../firebaseConfig";
 import * as backend from "../functions/backend";
+import { ensureTargeted } from "../../functions/ensureTargeted";
 
-/**
- *  Prepare creates versions for each Hosting site to be deployed.
- */
-export async function prepare(context: Context, options: HostingOptions & Options): Promise<void> {
+function handlePublicDirectoryFlag(options: HostingOptions & Options): void {
   // Allow the public directory to be overridden by the --public flag
   if (options.public) {
     if (Array.isArray(options.config.get("hosting"))) {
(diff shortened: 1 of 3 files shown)


src/deploy/hosting/prepare.ts
```diff
@@ -23,6 +21,41 @@ export async function prepare(context: Context, options: HostingOptions & Option
...
+  options: HostingOptions & Options
+): Promise<boolean> {
+  // This must be called before modifying hosting config because we turn it from
+  // a scalar to an array now
+  handlePublicDirectoryFlag(options);
+
+  let addedFunctions = false;
+  for (const c of config.hostingConfig(options)) {
+    for (const r of c.rewrites || []) {
+      if (!("function" in r) || typeof r.function !== "object" || !r.function.pinTag) {
```
jamesdaniels, line 40: add same logic for r.run.pinTag too

src/deploy/index.ts
```diff
@@ -77,6 +78,16 @@ export const deploy = async function (
     }
   }
 
+  if (
```
jamesdaniels, line 81: we need to remove the functions unshift in frameworks/index#prepareFrameworks for this logic to work

src/deploy/hosting/prepare.ts
inlined: Don't need it because we don't support deploying a Run service.

src/deploy/index.ts
inlined: Oh, really good point. I think we should just remove the guard against `!targetName.includes("functions")` to simplify things.

src/deploy/hosting/prepare.ts
jamesdaniels: oh fair point

src/deploy/hosting/prepare.ts
```diff
@@ -23,6 +21,41 @@ export async function prepare(context: Context, options: HostingOptions & Option
...
+        continue;
+      }
+
+      const existing = await backend.existingBackend(context);
+      const endpoint =
+        existing.endpoints[r.function.region || "us-central1"][r.function.functionId];
+      options.only = ensureTargeted(options.only, endpoint.codebase || "default", endpoint.id);
```
jamesdaniels, line 47: only might be undefined, can handle either here or default in ensureTargeted

src/deploy/index.ts
jamesdaniels: removing the guard would break `--only hosting,functions` as it would then append to only

inlined: Actually, we need to remove the guard. 

`--only hosting,functions` will still be `--only hosting,functions` after ensureTargeted, but `--only hosting,functions:someOtherFunction` may need to be `--only hosting,functions:ssr,functions:someOtherFunction`

bkendall (approved):

jamesdaniels (approved):

PR #5747, opened 2023-04-27
Small fixes

jamesdaniels:
* Remove specific handling for web frameworks
* Fix only=undefined handling, in the case of preview channels

### DIFF
--- a/src/deploy/hosting/prepare.ts
+++ b/src/deploy/hosting/prepare.ts
@@ -30,10 +30,6 @@ export async function addTaggedFunctionsToOnlyString(
   context: Context,
   options: HostingOptions & Options
 ): Promise<boolean> {
-  if (!options.only) {
-    return false;
-  }
-
   // This must be called before modifying hosting config because we turn it from
   // a scalar to an array now
   handlePublicDirectoryFlag(options);
--- a/src/frameworks/index.ts
+++ b/src/frameworks/index.ts
@@ -26,6 +26,7 @@ import { FirebaseError } from "../error";
 import { requireHostingSite } from "../requireHostingSite";
 import { HostingRewrites } from "../firebaseConfig";
 import * as experiments from "../experiments";
+import { ensureTargeted } from "../functions/ensureTargeted";
 import { implicitInit } from "../hosting/implicitInit";
 import { fileExistsSync } from "../fsutils";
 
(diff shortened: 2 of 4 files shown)


src/deploy/hosting/prepare.ts
```diff
@@ -30,10 +30,6 @@ export async function addTaggedFunctionsToOnlyString(
   context: Context,
   options: HostingOptions & Options
 ): Promise<boolean> {
-  if (!options.only) {
```
inlined, line 33: This was here because a deployment that doesn't have an `--only` flag doesn't need to update `--only`; it's already deploying everything.

src/frameworks/index.ts
```diff
@@ -502,13 +501,6 @@ export async function prepareFrameworks(
         },
       ]);
 
-      if (!targetNames.includes("functions")) {
```
inlined, line 505: This included a nice optimization that we knew the codebase even if the function hadn't been deployed yet which does/should/can mean we don't actually compile all other functions' source code to fiugre out which one matches this SSR function.

inlined (commented):
Do we have repros for the issues you're trying to fix here?

src/deploy/hosting/prepare.ts
jamesdaniels: only is undefined when deploying to a preview channel, we should remain targeted

src/frameworks/index.ts
jamesdaniels: Fair, I can keep. Will pivot to the fix discussed in your branch

jamesdaniels: Ok thought about this some more, i'll keep the ensureTargeted code and drop unshifting functions, leave that to prepare. Broken case otherwise is `--only hosting,functions`

jamesdaniels: As we don't want to populate "only" in that case

jamesdaniels:
> Do we have repros for the issues you're trying to fix here?

A) if using web frameworks, only the web frameworks functions are being deployed on your branch
B) `firebase hosting:channel:deploy XYZ` is deploying all functions, not just the pinned ones

src/frameworks/index.ts
jamesdaniels: Ah that's correct, maybe i messed something else up when i was playing

PR #5753, opened 2023-04-28
Lift the pinTags experiment

jamesdaniels:


inlined:
Do we want to yank the experiment or default it to true so that customers can disable without downgrading if there are bugs?

bkendall (approved):

PR #5757, opened 2023-04-28
moves permissions check to be earlier and conditional. 

inlined:
1. Move the code for checking for GCF permissions to deploy/index where we can only do the check if functions is added by pintags
2. Make the functions permission check based on a set of permissions (and the same list by commands/deploy) rather than a fixed role name
3. Add fallback support in addTaggedFunctionsToOnlyString so that the function can limp along when backend.existingBackend fails because we haven't checked for functions permissions yet.

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -18,7 +18,7 @@ export const VALID_DEPLOY_TARGETS = [
   "remoteconfig",
   "extensions",
 ];
-export const TARGET_PERMISSIONS: Record<string, string[]> = {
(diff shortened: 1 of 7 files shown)


jamesdaniels:
This should work but I don't really like the try/catch. Would it be better to add pinned functions to context? Parse first, `Record<region: string, functionIds: string[]>`, validate permissions if in Github Actions, and then append to only/targets?

bkendall (commented):
I really can't speak to if this is the correct fix, but codewise it looks alright.

inlined:
Took PR feedback. Also, this new code allows us to remove the web frameworks code for manipulating deploy targets since the function will always be pulled in thanks to tag pinning.

jamesdaniels (approved):
LGTM

PR #5762, opened 2023-04-30
Adding new console links to ext:dev:register and ext:dev:upload

joehan:
### Description
Add links to publisher profile after ext:dev:register, and to the extension details page after ext:dev:upload/publish.

### DIFF
--- a/src/commands/ext-dev-register.ts
+++ b/src/commands/ext-dev-register.ts
@@ -34,7 +34,7 @@ export const command = new Command("ext:dev:register")
       message: msg,
       default: projectId,
     });
-    let profile: PublisherProfile
+    let profile: PublisherProfile;
     try {
       profile = await registerPublisherProfile(projectId, publisherId);
     } catch (err: any) {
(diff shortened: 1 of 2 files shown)


src/commands/ext-dev-register.ts
```diff
@@ -55,10 +57,14 @@ export const command = new Command("ext:dev:register")
...
       );
     }
-    return utils.logLabeledSuccess(
+    utils.logLabeledSuccess(
       logPrefix,
       `Publisher ID '${clc.bold(publisherId)}' has been registered to project ${clc.bold(
         projectId
+      )}. View and edit your profile at ${utils.consoleUrl(
+        projectId,
+        `/publisher/${publisherId}/dashboard`
```
elvisun, line 66: is publisherId needed?

I thought the URL is 
https://firebase.corp.google.com/project/extensions-staging-test/publisher/dashboard

elvisun: Can also drop `/dashboard` and rely on console redirect to decide which view to show by default, in case we change it in the future

elvisun (approved):

src/commands/ext-dev-register.ts
joehan: Huh, i think there are some redirects in place - the link used here works too. Will switch this to the simplest one though, good call.

PR #5769, opened 2023-05-01
Update firebase emulator ui to 1.11.6

christhompsongoogle:
Update firebase emulator ui to 1.11.6


src/emulator/downloadableEmulators.ts
```diff
@@ -45,9 +45,9 @@ const EMULATOR_UPDATE_DETAILS: { [s in DownloadableEmulators]: EmulatorUpdateDet
   ui: experiments.isEnabled("emulatoruisnapshot")
     ? { version: "SNAPSHOT", expectedSize: -1, expectedChecksum: "" }
     : {
-        version: "1.11.5",
+        version: "1.11.6",
         expectedSize: 3063444,
```
christhompsongoogle, line 49: This is an odd coincidence but I checked it twice .. thee bytes are the same

joehan: LOL, thanks for double checking this, i would have been concerned

joehan (approved):

PR #5774, opened 2023-05-02
Angular i18n, baseHref, and optimized images

jamesdaniels:
* Add support for Angular i18n
* Add support for Angular baseUrl
* Move context logic into utils
* Shell out all targets, best not to have conditional logic
* deploy.serveOptimizedImages trips requirement for backend
* Validate i18n locales
* Use baseUrl in rewrite (ng & nextjs)
* Angular needs an i18n root of "/"—I only departed from this with NextJS to aid debugging
* Add Angular to the WF integration test
* Bump firebase-frameworks version
* Give sharp a version range
* Reset env vars and memos after prepare


src/frameworks/angular/index.ts
```diff
@@ -53,35 +52,20 @@ export async function init(setup: any, config: any) {
...
     });
-  } else {
-    await scheduleTarget(browserTarget);
-    if (serverTarget) await scheduleTarget(serverTarget);
   }
-
-  const wantsBackend = !!serverTarget;
-
-  return { wantsBackend };
+  const wantsBackend = !!serverTarget || ngOptimizedImage;
```
leoortizz, line 66: maybe we could log the reasons for backend as we do with Next.js?

jamesdaniels: absolutely, i expect more reasons as Angular is on a roll lately

src/frameworks/angular/utils.ts
```diff
@@ -0,0 +1,305 @@
+import type { Target } from "@angular-devkit/architect";
+import type { ProjectDefinition } from "@angular-devkit/core/src/workspace";
+import type { WorkspaceNodeModulesArchitectHost } from "@angular-devkit/architect/node";
+
+import { AngularI18nConfig } from "./interfaces";
+import { relativeRequire } from "../utils";
+import { FirebaseError } from "../../error";
+import { join } from "path";
+
+async function localesForTarget(
```
leoortizz, line 10: would be great to have unit tests for this

chalosalvador (commented):
@jamesdaniels what's a working `angular.json` file to test this? I'm using a basic i18n set up and getting an error (`Error: Failed to lookup view "index" in views directory`) when running with `firebase emulators:start`, it is working when running with `ng serve`.

jamesdaniels:
@chalosalvador the failure is due to your server.ts not being adapted for localization, you'll need to make the following changes:

```ts
// The Express app is exported so that it can be used by serverless Functions.
export function app(locale: string): express.Express {
  const server = express();
  const distFolder = join(process.cwd(), `dist/hosting/browser/${locale}`);
```

Note the local string being passed into app and used for the distFolder. We should def document this & set it up out of box with the starter template.

src/frameworks/constants.ts
```diff
@@ -27,9 +27,10 @@ export const FEATURE_REQUEST_URL =
   "https://github.com/firebase/firebase-tools/issues/new?template=feature_request.md";
 export const MAILING_LIST_URL = "https://goo.gle/41enW5X";
 
-export const FIREBASE_FRAMEWORKS_VERSION = "^0.9.0";
+export const FIREBASE_FRAMEWORKS_VERSION = "0.10.0-canary.1e4546c";
```
jamesdaniels, line 30: Bump once we have a version

src/frameworks/index.ts
```diff
@@ -249,7 +249,7 @@ export async function prepareFrameworks(
       `\n${frameworksCallToAction(SupportLevelWarnings[support](name), docsUrl, "   ")}\n`
     );
     // TODO allow for override
-    const isDevMode = context._name === "serve" || context._name === "emulators:start";
+    const isDevMode = false; // context._name === "serve" || context._name === "emulators:start";
```
jamesdaniels, line 252: revert

leoortizz (approved):
LGTM! I was able to deploy Angular i18n projects in both MacOS and Windows. Found an issue in Windows with `spawn` and pushed a fix, all working now!

The error messages also helped me fix my project without leaving the terminal, amazing job @jamesdaniels!

PR #5785, opened 2023-05-03
Update functions template

jhuleatt:
Match Node's default of adding Firebase-Admin as a dep

### Description

Auto-import Firebase Admin SDK, like Node does:

https://github.com/firebase/firebase-tools/blob/90bf9499079a3ae3e6bb9cb953cb1e00051389fe/templates/init/functions/javascript/package.lint.json#L16-L18

### DIFF
--- a/templates/init/functions/python/requirements.txt
+++ b/templates/init/functions/python/requirements.txt
@@ -1,1 +1,2 @@
 firebase_functions~=0.1.0
+firebase-admin~=6.1.0


taeold (approved):

templates/init/functions/python/requirements.txt
```diff
@@ -1 +1,2 @@
-firebase_functions~=0.1.0
\ No newline at end of file
+firebase_functions~=0.1.0
+firebase-admin~=6.1.0
```
jhuleatt, line 2: @taeold per our chat, it sounds like this line isn't actually necessary. Is there a risk of weird behavior if the `firebase-admin` version here gets out of sync with the `firebase-admin` version expected by `firebase_functions`?

PR #5789, opened 2023-05-03
fix: don't assume nextjs app root is workspace root

OscarLundstrom:
### Description

Configuring a NextJs app in `firebase.json` that doesn't have its own `package.json` results in firebase failing to detect the framework:

```json
// firebase.json
...
"hosting": {
    "source": "./path/to/next-js-app-project-without-package-json"
}
...
```
This assumption is often not true in a monorepo.

This fix removes that assumption, which shouldn't be an issue because the remaining checks should still be able to reliably determine the framework.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Fix framework detection limitation for NextJs applications that aren't co-located with their package.json (monorepo)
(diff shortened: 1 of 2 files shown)


ibrahimAboelsuod:
In an nx monorepo `findDependency()` fails to find next. Because it thinks the source path is the project root when now it's not.

PR #5793, opened 2023-05-04
Implement module to link stack to GitHub repository

blidd-google:
🥞 <--> github


src/init/features/turtles/repo.ts
```diff
@@ -0,0 +1,168 @@
...
+        {
+          name: "Continue",
+          value: "continue",
+        },
+        {
+          name: "Cancel",
+          value: "cancel",
+        },
+      ],
+    });
```
taeold, line 121: Hmm I think I might consider extracting this out as a separate function.

A single function that handles API calls, polls, and user prompts smells a little.

Think it's easier for me to read code that codes like:

```
getConnection
if (stage == PENDING USER OAUTH)
  open browser
poll until state is updated (?)
```

It'd at least be eaiser to test out each components?

src/commands/internaltesting-turtles-init.ts
```diff
@@ -0,0 +1,14 @@
+import { Command } from "../command";
+import { linkGitHubRepository } from "../init/features/turtles/repo";
+import { Options } from "../options";
+import { needProjectId } from "../projectUtils";
+import requireInteractive from "../requireInteractive";
+
+export const command = new Command("internaltesting:turtles:init")
```
taeold, line 7: nit* controversial: we should use "frameworks" instead of "turtle"

why? 🤷🏼

src/init/features/turtles/repo.ts
```diff
@@ -0,0 +1,173 @@
...
+import { promptOnce } from "../../../prompt";
+
+const gcbPollerOptions: Omit<poller.OperationPollerOptions, "operationResourceName"> = {
+  apiOrigin: cloudbuildOrigin,
+  apiVersion: "v2",
+  masterTimeout: 25 * 60 * 1_000,
+  maxBackoff: 10_000,
+};
+
+function extractRepoSlugFromURI(remoteUri: string): string | undefined {
```
taeold, line 16: nit* not asking for test here, but I'd be happy with a example use case in the comments, e.g.

```
extractRepoSlugFromURL("AAAAAA")
=> "BBBBBBBBBBBBBB"
```

src/init/features/turtles/repo.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+  return match[1];
+}
+
+function generateConnectionId(stackId: string): string {
+  return `turtles-${stackId}-conn`;
+}
+
+function generateRepositoryId(): string | undefined {
+  return `turtles-repo`;
```
taeold, line 29: can we add a comment here that this is totally arbitrary decision (and the implication of this - e.g. this means we will maintain one repo per project per region... right?)

src/init/features/turtles/repo.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  let remoteUri = await promptRepositoryURI(projectId, location, connectionId);
+  while (remoteUri === "") {
+    await utils.openInBrowser("https://github.com/apps/google-cloud-build/installations/new");
+    await promptOnce({
+      type: "input",
+      message:
+        "Press any key once you have finished configuring your installation's access settings.",
+    });
+    remoteUri = await promptRepositoryURI(projectId, location, connectionId);
```
taeold, line 51: is there some delay between installing the GitHub app and it showing up on Cloud build?

wondering if we should insert some delay here or have some clarification here ("Uh-oh. We still don't have your GitHub installation token, so we are going to open up the installation window again") before we jump back into the infinite loop.

src/init/features/turtles/repo.ts
```diff
@@ -0,0 +1,173 @@
...
+async function promptRepositoryURI(
+  projectId: string,
+  location: string,
+  connectionId: string
+): Promise<string> {
+  const resp = await gcb.fetchLinkableRepositories(projectId, location, connectionId);
+  if (!resp.repositories || resp.repositories.length === 0) {
+    throw new FirebaseError(
+      "The GitHub App does not have access to any repositories. Please configure " +
+        "your app installation permissions at https://github.com/settings/installations."
```
taeold, line 68: 👍🏼

taeold (approved):

src/init/features/turtles/repo.ts
```diff
@@ -0,0 +1,184 @@
...
+/**
+ * Exported for unit testing.
+ */
+export async function getOrCreateRepository(
+  projectId: string,
+  location: string,
+  connectionId: string,
+  remoteUri: string
+): Promise<gcb.Repository> {
+  const repositoryId = generateRepositoryId();
```
svnsairam, line 154: Will repositoryId be constant?

src/init/features/turtles/repo.ts
```diff
@@ -0,0 +1,184 @@
...
+  projectId: string,
+  location: string,
+  connectionId: string
+): Promise<gcb.Connection> {
+  let conn: gcb.Connection;
+  try {
+    conn = await gcb.getConnection(projectId, location, connectionId);
+  } catch (err: unknown) {
+    if ((err as FirebaseError).status === 404) {
+      const op = await gcb.createConnection(projectId, location, connectionId);
```
svnsairam, line 128: Why createConnection returns an Operation instead of Connection object?

src/init/features/turtles/repo.ts
```diff
@@ -0,0 +1,184 @@
...
+      conn = await poller.pollOperation<gcb.Connection>({
+        ...gcbPollerOptions,
+        pollerName: `create-${location}-${connectionId}`,
+        operationResourceName: op.name,
+      });
```
svnsairam, line 133: T poller waits for Promise and resolves when the operation is completed. But can these be used after every await or should these be used in certain usecases(here this is not used with getConnection but used with createConnection).

svnsairam (approved):

src/init/features/turtles/repo.ts
taeold: Because Cloud Build's createConnection API is an asynchronous API that returns a Long Running Operation:

https://google.aip.dev/151

src/init/features/turtles/repo.ts
blidd-google: The `getConnection` API simply returns a Connection, so there's no need to poll since there is no Operation to poll on. `createConnection` on the other hand does require polling until the Operation completes.

src/init/features/turtles/repo.ts
blidd-google: Right now, yes, but this may be subject to change in the future depending on how our design evolves!

PR #5801, opened 2023-05-05
Default to open rules when emulating demo- projects without matching targets

### LINKED ISSUE #5170: Can no longer run emulators with --project=dummy
In Version 11.13.0 of firebase-tools (and previous versions), I was able to launch the emulator with
`firebase emulators:start --import my-backup --inspect-functions --project=dummy`

Since upgrading to 11.15.0, I get
`Error: Deploy target xxx not configured for project dummy. Configure with...`
(where xxx is a real deployment target).

I don't want the emulators to start with the same project name as a 'real' project (as I was doing previously)...

joehan:
Change the Storage emulator to fall back to default open rules for 'demo-' projects if storage config refers to targets that are not defined for the project in  `.firebaserc`.
Fixes #5170

### DIFF
--- a/src/emulator/storage/rules/config.ts
+++ b/src/emulator/storage/rules/config.ts
@@ -62,7 +62,7 @@ export function getStorageRulesConfig(
         storageLogger.logLabeled(
           "BULLET",
           "storage",
-          `Detected demo project ID "${projectId}", using a default (open) rules configuration.`
+          `Detected demo project ID "${projectId}", using a default (open) rules configuration. Storage targets in firebase.json will be ignored.`
         );
         return defaultStorageRules();
       }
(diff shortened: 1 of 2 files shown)


kroikie (approved):

src/emulator/storage/rules/config.ts
```diff
@@ -49,18 +48,37 @@ export function getStorageRulesConfig(
...
-      results.push({ resource, rules: getSourceFile(targetConfig.rules, options) });
-    });
+    const targets = rc.target(projectId, "storage", targetConfig.target);
+    if (targets.length === 0) {
+      // Fall back to open if this is a demo project
+      if (Constants.isDemoProject(projectId)) {
+        storageLogger.logLabeled(
+          "BULLET",
+          "storage",
+          `Detected demo project ID "${projectId}", using a default (open) rules configuration.`
```
christhompsongoogle, line 65: Super-mega-nit: Can we alter this log very slightly for debugging purposes so we know if it was logged here or in the above case?

christhompsongoogle (approved):

PR #5804, opened 2023-05-07
Refactored ext:dev:upload with new features in mind.

apascal07:
This PR consists of a few things:

- Refactoring of the ext:dev:upload code into smaller functions that are called by both the local and GitHub upload paths.
- Adding quality of life features and information like extension metadata, next version by stage, warnings for not uploading pre-release versions first, callouts that stable versions will be submitted for review, etc.
- Adding restrictions on upload from local to pre-release only.

Screenshot 2023-05-08 at 9 08 41 AM
Screenshot 2023-05-08 at 9 08 56 AM
Screenshot 2023-05-08 at 9 13 50 AM
Screenshot 2023-05-08 at 9 15 24 AM
Screenshot 2023-05-08 at 9 55 56 AM

### DIFF
--- a/src/extensions/extensionsHelper.ts
+++ b/src/extensions/extensionsHelper.ts
@@ -95,7 +95,7 @@ export const resourceTypeToNiceName: Record<string, string> = {
 };
 export type ReleaseStage = "alpha" | "beta" | "rc" | "stable";
 const repoRegex = new RegExp(`^https:\/\/github\.com\/[^\/]+\/[^\/]+$`);
-const stageOptions = ["alpha", "beta", "rc", "stable"];
+const stageOptions = ["rc", "alpha", "beta", "stable"];
 
 /**
  * Turns database URLs (e.g. https://my-db.firebaseio.com) into database instance names
@@ -248,6 +248,13 @@ export function validateSpec(spec: any) {
   }
   if (!spec.version) {
     errors.push("extension.yaml is missing required field: version");
+  } else if (!semver.valid(spec.version)) {
+    errors.push(`version ${spec.version} in extension.yaml is not a valid semver`);
+  } else {
+    const version = semver.parse(spec.version)!;
+    if (version.prerelease.length > 0 || version.build.length > 0) {
+      errors.push("version field in extension.yaml does not support pre-release annotations");
+    }
   }
   if (!spec.license) {
     errors.push("extension.yaml is missing required field: license");
@@ -388,7 +395,7 @@ export async function promptForValidRepoURI(): Promise<string> {
   while (!repoIsValid) {
     extensionRoot = await promptOnce({
       type: "input",
-      message: "Enter the GitHub repo URI where this Extension's source code is located:",
+      message: "Enter the GitHub repo URI where this extension's source code is located:",
     });
     if (!repoRegex.test(extensionRoot)) {
       logger.info("Repo URI must follow this format: https://github.com/<user>/<repo>");
@@ -397,6 +404,92 @@ export async function promptForValidRepoURI(): Promise<string> {
     }
   }
   return extensionRoot;
+}
+
+/**
+ * Prompts for a valid extension root.
+ *
+ * @param defaultRoot the default extension root
+ */
+export async function promptForValidExtensionRoot(defaultRoot: string): Promise<string> {
+  let rootIsValid = false;
+  let extensionRoot = "";
+  while (!rootIsValid) {
+    extensionRoot = await promptOnce({
+      type: "input",
+      message:
+        "Enter this extension's root directory in the repo (defaults to previous root if set):",
+      default: defaultRoot,
+    });
+    // TODO: Change this to a real check.
+    if (false) {
+      logger.info(`${extensionRoot} is not a valid directory path.`);
+    } else {
+      rootIsValid = true;
+    }
+  }
+  return extensionRoot;
+}
+
+/**
+ * Prompts for the extension version's release stage.
+ *
+ * @param versionByStage map from stage to the next version to upload
+ * @param autoReview whether the stable version will be automatically sent for review on upload
+ * @param allowStable whether to allow stable versions
+ * @param hasVersions whether there have been any pre-release versions uploaded already
+ */
+async function promptForReleaseStage(args: {
+  versionByStage: Map<string, string>;
+  autoReview: boolean;
+  allowStable: boolean;
+  hasVersions: boolean;
+  nonInteractive: boolean;
+  force: boolean;
+}): Promise<ReleaseStage> {
+  const defaultStage = "rc";
+  let stage: ReleaseStage;
+  const choices = [
+    { name: `Release candidate (${args.versionByStage.get("rc")})`, value: "rc" },
+    { name: `Alpha (${args.versionByStage.get("alpha")})`, value: "alpha" },
+    { name: `Beta (${args.versionByStage.get("beta")})`, value: "beta" },
+  ];
+  if (args.allowStable) {
+    const stableChoice = {
+      name: `Stable (${args.versionByStage.get("stable")}${
+        args.autoReview ? ", automatically sent for review" : ""
+      })`,
+      value: "stable",
+    };
+    choices.push(stableChoice);
+  }
+  stage = await promptOnce({
+    type: "list",
+    message: "Choose the release stage:",
+    choices: choices,
+    default: defaultStage,
+  });
+  if (stage === "stable" && !args.hasVersions) {
+    logger.info(
+      `${clc.bold(
+        clc.yellow("Warning:")
+      )} It's highly recommended to first upload a pre-release version before choosing stable.`
+    );
+    const confirmed = await confirm({
+      nonInteractive: args.nonInteractive,
+      force: args.force,
+      default: false,
+    });
+    if (!confirmed) {
+      stage = await promptOnce({
+        type: "list",
+        message: "Choose the release stage:",
+        choices: choices,
+        default: defaultStage,
+      });
+    }
+  }
+  return stage;
 }
 
 export async function ensureExtensionsApiEnabled(options: any): Promise<void> {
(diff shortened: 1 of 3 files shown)


src/extensions/extensionsHelper.ts
```diff
@@ -667,58 +869,48 @@ export async function uploadExtensionVersionFromGitHubSource(args: {
...
+  const rootDirectory = await fetchExtensionSource(repoUri, sourceRef, extensionRoot);
+  const extensionSpec = await validateExtensionSpec(rootDirectory, args.extensionId);
+  validateVersion(extensionRef, extensionSpec.version, extension?.latestVersion);
+  const { versionByStage, hasVersions } = await getNextVersionByStage(
+    extensionRef,
+    extensionSpec.version
+  );
+  const autoReview =
+    !!extension?.latestApprovedVersion ||
+    ["PENDING", "APPROVED", "REJECTED"].includes(latestVersion?.listing?.state ?? "");
```
joehan, line 881: Nit: Took a few tries to grok what this line does. Might be simpler as:
```
!!extension?.latestApprovedVersion ||
latestVersion?.listing?.state === "PENDING ||
latestVersion?.listing?.state === "APPROVED ||
latestVersion?.listing?.state === "REJECTED;
```

src/extensions/extensionsHelper.ts
```diff
@@ -904,33 +1117,35 @@ export function getPublisherProjectFromName(publisherName: string): number {
...
+export function displayReleaseNotes(args: {
+  extensionRef: string;
+  newVersion: string;
+  autoReview: boolean;
+  releaseNotes?: string;
+  sourceUri?: string;
+}): void {
+  const source = args.sourceUri || "Local source";
+  const releaseNotesMessage = args.releaseNotes
+    ? `${clc.bold("Release notes:")}\n${marked(args.releaseNotes)}`
```
joehan, line 1136:
```suggestion
    ? `${clc.bold("Release notes:")}\n${marked(args.releaseNotes)}\n`
```

src/extensions/extensionsHelper.ts
```diff
@@ -824,7 +1037,7 @@ export async function uploadExtensionVersionFromLocalSource(args: {
     throw err;
   }
   await deleteUploadedSource(objectPath);
-  return res;
+  return;
```
joehan, line 1040: Need to return res here. We need to return the extensionVersion from ext:dev:upload so that it supports the --json flag.

src/extensions/extensionsHelper.ts
```diff
@@ -479,51 +568,78 @@ export async function incrementPrereleaseVersion(
...
+      semver.inc(`${newVersion}-${stage}`, "prerelease", undefined, stage)!,
+    ])
+  );
+  let extensionVersions: ExtensionVersion[] = [];
+  try {
+    extensionVersions = await listExtensionVersions(extensionRef, `id="${newVersion}"`, true);
+    for (const extensionVersion of extensionVersions) {
+      const version = semver.parse(extensionVersion.spec.version)!;
+      // Extensions only support semvers of format: <major>.<minor>.<patch>-<stage>.<count>
+      const stage = String(version.prerelease[0]).split(".")[0];
```
joehan, line 592: This will break if version.prerelease is length 0.
```suggestion
      const stage = String(version.prerelease.?[0]).split(".")[0];
```

src/extensions/extensionsHelper.ts
```diff
@@ -479,51 +568,78 @@ export async function incrementPrereleaseVersion(
...
+  newVersion: string
+): Promise<{ versionByStage: Map<string, string>; hasVersions: boolean }> {
+  const versionByStage = new Map(
+    ["rc", "alpha", "beta"].map((stage) => [
+      stage,
+      semver.inc(`${newVersion}-${stage}`, "prerelease", undefined, stage)!,
+    ])
+  );
+  let extensionVersions: ExtensionVersion[] = [];
+  try {
```
joehan, line 587: This function is giving me some bad code smells - right now, it is very error driven, and throws+catches errors during the normal & intended behavior of handling stable versions.
Could we avoid this if we filter out all ev's that don't have a prerelease?

joehan: Or, alternatively, we should error out if a stable version with this id already exists, since the upload will definitely fail then.

src/extensions/extensionsHelper.ts
```diff
@@ -479,51 +568,78 @@ export async function incrementPrereleaseVersion(
...
-    args.stage
-  );
-
+/**
+ * Validates the release notes.
+ *
+ * @param rootDirectory the directory with the extension source
+ * @param newVersion the new extension version
+ */
+function validateReleaseNotes(rootDirectory: string, newVersion: string, required: boolean) {
```
joehan, line 638: I'd prefer not to add 'required' as an argument to this function - it spreads the logic of when release notes are necessary across the codebase instead of keeping confined to a single function.

src/extensions/extensionsHelper.ts
```diff
@@ -534,43 +650,156 @@ async function validateExtensionSpec(args: {
...
+    } catch (err: any) {
+      // Silently fail and continue if extension has no latest version.
+    }
+    // TODO: Fix this logic.
+    const source = extension.repoUri
+      ? `${new URL(
+          latestVersion?.extensionRoot ?? "",
+          extension.repoUri
+        )} (use --repo and --root to modify)`
+      : "-";
```
joehan, line 745: Nit: would preferable to say 'built from local source' instead of -

src/extensions/extensionsHelper.ts
```diff
@@ -534,43 +650,156 @@ async function validateExtensionSpec(args: {
...
+      return "-";
+  }
+}
+
+/**
+ * Fetches the extension and its latest version and displays all metadata.
+ *
+ * @param extensionRef the ref of the extension
+ */
+async function displayExtensionHeader(extensionRef: string): Promise<{
```
joehan, line 726: Would be better to not combine the logic of displaying a header with the logic for fetching backend code. 

I'd prefer to make this only handle the logging, and do the fetching elsewhere:
```
async function displayExtensionHeader(e: Extension, latestEv?: ExtensionVersion): Promise<void>
```

joehan (commented):
Early drive by to hopefully save some back and forth

src/extensions/extensionsHelper.ts
apascal07: I agree it kinda sucks but it's limited by the API. Also, the try/catch is only there to handle 404s, not for anything else. I can refactor to make the try/catch around the API fetch only.

The stable is outside of the main logic because regardless of what exists already, the stable version pulled from extension.yaml is what would be uploaded next (whether it succeeds or is valid or not).

We do error out prior to this in `validateVersion()` if `newVersion` is less than or equal to the latest version.

apascal07: Refactored it to hopefully make it a bit clearer/less sketchy looking.

src/extensions/extensionsHelper.ts
```diff
@@ -534,43 +617,145 @@ async function validateExtensionSpec(args: {
...
-        )}'. Please increment the version inside of extension.yaml.\n`,
+        )}'. Increment the version inside of extension.yaml and try again.\n`,
         { exit: 103 }
       );
     }
   }
+}
+
+/** Unpacks extension state into a more specific string. */
+export function unpackExtensionState(extension: Extension) {
```
joehan, line 668: Unit test for this would be good.

joehan (commented):
Few more things to clean up, but looking pretty good!

src/extensions/extensionsHelper.ts
```diff
@@ -399,6 +406,88 @@ export async function promptForValidRepoURI(): Promise<string> {
...
+  let rootIsValid = false;
+  let extensionRoot = "";
+  while (!rootIsValid) {
+    extensionRoot = await promptOnce({
+      type: "input",
+      message:
+        "Enter this extension's root directory in the repo (defaults to previous root if set):",
+      default: defaultRoot,
+    });
+    // TODO: Add real directory path validation.
```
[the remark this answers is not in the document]
apascal07: This is more for valid paths but with extraneous symbols. The roots should be fairly trivial in most cases but they haven't been always.

src/extensions/extensionsHelper.ts
```diff
@@ -399,6 +406,88 @@ export async function promptForValidRepoURI(): Promise<string> {
...
+  if (args.allowStable) {
+    const stableChoice = {
+      name: `Stable (${args.versionByStage.get("stable")}${
+        args.autoReview ? ", automatically sent for review" : ""
+      })`,
+      value: "stable",
+    };
+    choices.push(stableChoice);
+  }
+  stage = await promptOnce({
```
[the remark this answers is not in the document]
apascal07: Done. Force is only for confirmations, no?

src/extensions/extensionsHelper.ts
```diff
@@ -441,89 +530,83 @@ async function archiveAndUploadSource(extPath: string, bucketName: string): Prom
...
-    return semver.inc(latestVersion, "prerelease", undefined, stage)!;
+async function getNextVersionByStage(
+  extensionRef: string,
+  newVersion: string
+): Promise<{ versionByStage: Map<string, string>; hasVersions: boolean }> {
+  let extensionVersions: ExtensionVersion[] = [];
+  try {
+    extensionVersions = await listExtensionVersions(extensionRef, `id="${newVersion}"`, true);
+  } catch (err: any) {
+    // Silently fail if there are no extension versions exist.
```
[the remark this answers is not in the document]
apascal07: Done.

src/extensions/extensionsHelper.ts
```diff
@@ -441,89 +530,83 @@ async function archiveAndUploadSource(extPath: string, bucketName: string): Prom
...
+  const versionByStage = new Map(
+    ["rc", "alpha", "beta"].map((stage) => [
+      stage,
+      semver.inc(`${newVersion}-${stage}`, "prerelease", undefined, stage)!,
+    ])
+  );
+  extensionVersions
+    .map((extensionVersion) => semver.parse(extensionVersion.spec.version)!)
+    .filter((version) => !!version.prerelease)
+    .forEach((version) => {
```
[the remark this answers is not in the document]
apascal07: Done.

src/extensions/extensionsHelper.ts
```diff
@@ -248,6 +248,13 @@ export function validateSpec(spec: any) {
   }
   if (!spec.version) {
     errors.push("extension.yaml is missing required field: version");
+  } else if (!semver.valid(spec.version)) {
+    errors.push(`version ${spec.version} in extension.yaml is not a valid semver`);
+  } else {
+    const version = semver.parse(spec.version)!;
+    if (version.prerelease.length > 0 || version.build.length > 0) {
+      errors.push("version field in extension.yaml does not support pre-release annotations");
```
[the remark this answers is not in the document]
apascal07: Done.

src/extensions/extensionsHelper.ts
```diff
@@ -596,16 +785,21 @@ export async function uploadExtensionVersionFromGitHubSource(args: {
...
+    latestVersion = await getExtensionVersion(`${extensionRef}@latest`);
   } catch (err: any) {
-    // Silently fail and continue the publish flow if Extension not found.
+    // Silently fail and continue if extension is new or has no latest version set.
   }
+  displayExtensionHeader(extensionRef, extension, latestVersion?.extensionRoot);
 
-  // Prompt for repo URI and validate that it hasn't changed if previously set.
+  if (args.stage && !stageOptions.includes(args.stage)) {
+    throw new FirebaseError(`--stage flag only supports the following values: ${stageOptions}`);
```
joehan, line 799: Super nit - this currently prints out with no spaces between the entries in stage options
```suggestion
    throw new FirebaseError(`--stage flag only supports the following values: ${stageOptions.join(', ')}`);
```

src/extensions/extensionsHelper.ts
```diff
@@ -769,23 +923,48 @@ export async function uploadExtensionVersionFromLocalSource(args: {
...
-    publisherId: args.publisherId,
-    extensionId: args.extensionId,
-    rootDirectory: args.rootDirectory,
-    latestVersion: extension?.latestVersion,
-    stage: args.stage,
-  });
+  const localStageOptions = ["rc", "alpha", "beta"];
+  if (args.stage && !localStageOptions.includes(args.stage)) {
+    throw new FirebaseError(
+      `--stage flag only supports the following values when used with --local: ${localStageOptions}`
```
joehan, line 940: Same as above - super nit - this currently prints out with no spaces between the entries in stage options:
```suggestion
      `--stage flag only supports the following values when used with --local: ${localStageOptions.join(', ')}`
```

src/test/extensions/extensionsHelper.spec.ts
```diff
@@ -977,6 +895,108 @@ describe("extensionsHelper", () => {
...
+        "test",
+        "1.0.0"
+      );
+      expect(Array.from(versionByStage.entries())).to.eql(Array.from(expected.entries()));
+      expect(hasVersions).to.eql(true);
+    });
+  });
+
+  describe("unpackExtensionState", () => {
+    it("should return correct state", () => {
```
joehan, line 954: Better to separate these into multiple it blocks, one for each call to unpackExtensionState. That way, they will fail in a more granular manner

joehan (approved):
LGTM after a minor refactor in one of the unit tests.

PR #5808, opened 2023-05-08
Remove 'request' from 'functions:shell'

### LINKED ISSUE #2215: [v13.0.0 Blocker] Refactor `functions:shell` to not use request
I am trying to run "npm install -g firebase-tools" command on linux OS but it gives me the above error every time.On the other hand when I try to run "npx create-react-app app-name" command, it works fine. I searched the solution on the internet but did not find any acceptable solution. Is there any way to fix this problem.

joehan:
### Description
Reviving an old PR from Bryan to get rid of 'request', for good.

### Scenarios Tested
I ran some light testing to sanity check that this didn't blow up the world, and had no issues with basic testing of https/firestore functions.

Fixes #2215


joehan:
Managed to break this in a few tests (ie an https function that sends a non JSON string back). Don't think this can safely make it into tomorrow's release, but will hopefully get it in ASAP after with minimal breaking changes to the API

src/localFunction.ts
```diff
@@ -33,25 +35,122 @@ export default class LocalFunction {
...
+    options?: HttpsOptions
+  ): { path: string; opts: HttpsOptions } {
+    let path = "/";
+    let opts: HttpsOptions = {};
+    // Just path
+    if (typeof pathOrOptions === "string" && !options) {
+      path = pathOrOptions;
+    } // error case, 2 options
+    else if (!!pathOrOptions && !!options && typeof pathOrOptions !== "string") {
+      throw new Error(
```
taeold, line 143: I had hard time evaluating this series of if statement in my head.

wdyt about this implementation?

```js
  // Case: No arguments provided
  if (!pathOrOptions && !options) {
    return { path: "/", opts: {} };
  }

  // Case: pathOrOptions is provided as a string
  if (typeof pathOrOptions === "string") {
    return { path: pathOrOptions, opts: options || {} };
  }

  // Case: pathOrOptions is an object (HttpsOptions), and options is not provided
  if (typeof pathOrOptions !== "string" && !options) {
    return { path: "/", opts: pathOrOptions };
  }

  // Error case: Invalid combination of arguments
  if (typeof pathOrOptions !== "string" || !options) {
    throw new Error(
      `Invalid argument combination: Expected a string and/or HttpsOptions, got ${typeof pathOrOptions} and ${typeof options}`
    );
  }

  // Default return, though this point should not be reached
  return { path: "/", opts: {} };
```

taeold (approved):
🥳 amazing - thanks for this work! small suggestions, otherwise 🚢

src/localFunction.ts
joehan: Oooh, that is much more readable, thanks!

src/localFunction.ts
```diff
@@ -33,25 +35,122 @@ export default class LocalFunction {
...
+          .then((res) => {
+            this.requestCallBack(undefined, res, res.body);
+          })
+          .catch((err) => {
+            this.requestCallBack(err);
+          });
+        return HTTPS_SENTINEL;
+      };
+    };
+    const verbWithoutReqBodyFactory = (
```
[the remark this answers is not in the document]
joehan: Good idea, done

PR #5815, opened 2023-05-09
[Fix] spawnSync from cross-spawn

Geoffrey-Pliez:
### Description

Use of package cross-pawn instead of child-process for cross-platform compatibility

### Scenarios Tested

on Windows, child_process.spawnSync does not always find the npm command and therefore the getNPMRoot function returns null.

Indeed spawnSync returned the error "spawnSync npm ENOENT"


jamesdaniels:
Thanks, we actually just caught too and merged in a fix on the next branch w/ #5774. This should be out the door shortly

PR #5659, opened 2023-04-05
Python frameworks

jamesdaniels:

PR #5825, opened 2023-05-10
Flask documentation & experimental callouts

jamesdaniels:
* Add flask doc
* Add flask to the overview
* Fixed the express link in overview
* Add note on experimental integrations


src/frameworks/docs/flask.md
```diff
@@ -0,0 +1,60 @@
...
+{% include "docs/hosting/_local_variables.html" %}
+
+<link rel="stylesheet" type="text/css" href="/styles/docs.css" />
+
+# Integrate Flask
+
+With the Firebase framework-aware {{cli}}, you can deploy your Flask application
+to Firebase and serve dynamic content to your users.
+
+Note: Flask support in Framework-aware {{hosting}} is coming soon.
```
egilmorez, line 16: The distinction may not be very clear at this point.  Note that we basically conflate framework-aware CLI and framework-aware Hosting in pages like https://firebase.google.com/docs/hosting/frameworks/angular

src/frameworks/docs/flask.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+You can [view your deployed app](/docs/hosting/test-preview-deploy#view-changes)
+on its live site.
+
+## Serve static content
+
+When you call deploy, your Flask applications's
+[static files](https://flask.palletsprojects.com/en/2.3.x/quickstart/#static-files)
+will be deployed to Firebase Hosting automatically—these files will be served
```
egilmorez, line 59: Suggest &mdash; or double dashes.

Also, for these and "will be deployed," present tense works, and is preferred.

https://developers.google.com/style/tense

src/frameworks/docs/flask.md
```diff
@@ -0,0 +1,60 @@
...
+```shell
+firebase deploy
+```
+
+You can [view your deployed app](/docs/hosting/test-preview-deploy#view-changes)
+on its live site.
+
+## Serve static content
+
+When you call deploy, your Flask applications's
```
egilmorez, line 57: Should this be just "When you deploy" or possibly "When you call `deploy`?"

src/frameworks/docs/frameworks-overview.md
```diff
@@ -51,4 +51,8 @@ See the detailed guide for your preferred framework:
 
 * [Angular Universal](/docs/hosting/frameworks/angular)
 * [Next.js] (/docs/hosting/frameworks/nextjs)
-* [Frameworks with Express.js](/docs/hosting/frameworks/ßexpress)
+* [Flask](/docs/hosting/frameworks/flask)
```
egilmorez, line 54: Is this the "coming soon" Flask support promised in the note at https://firebase.google.com/docs/hosting/frameworks/frameworks-overview?

If so, we need to modify the note.  And even if not, I'd really like to add that note into source, modified or otherwise (or it will be overwritten).  

LMK if you can do that in this PR James.  Basically the task is:

-- create _coming-soon.md file in /includes

-- include it in the overview by sticking <<_includes/_coming-soon.md>> between the paragraphs, same is in the live version.

jamesdaniels: Can do. This is the same integration, it's still coming soon but we wanted to get the docs up.

src/frameworks/docs/flask.md
jamesdaniels: Hard to tell in the monospace font github uses but that's an mdash, will work on tense

egilmorez: Oh cool -- I actually typed 'ampersand-"mdash"-semicolon' in my comment, but GH rendered it :D

src/frameworks/docs/frameworks-overview.md
egilmorez: Thanks!

src/frameworks/docs/flask.md
jamesdaniels: I can understand both but arguably the implementation *ATM* is that the CLI is framework aware and doing all the heavy lifting. IMO we should make it consistent.

egilmorez: Sounds true!  But right here, in line, what is the difference/distinction between

"With the Firebase framework-aware {{cli}}, you can deploy your Flask application to Firebase and serve . . ."

and

"Flask support in Framework-aware {{hosting}}"?

It sounds to me like we are saying "here's a thing, and the same thing is coming soon."  If there's a meaningful difference between what you get after you finish the steps in this guide and what you'll get coming soon, then I think we need more words to clarify the difference between those two things.

jamesdaniels:
Closing in favor of #5659 5659

PR #5834, opened 2023-05-14
Increases timeOut for detectFromPort to avoid sporadic deployment fails

himanshu810e:
Fixes: 
Root cause of following error
Error: User code failed to load. Cannot determine backend specification

### DIFF
--- a/src/deploy/functions/runtimes/discovery/index.ts
+++ b/src/deploy/functions/runtimes/discovery/index.ts
@@ -60,7 +60,7 @@ export async function detectFromPort(
   port: number,
   project: string,
   runtime: runtimes.Runtime,
-  timeout = 10_000 /* 10s to boot up */
+  timeout = 60_000 /* 10s to boot up */
 ): Promise<build.Build> {
   // The result type of fetch isn't exported
   let res: { text(): Promise<string> };


src/deploy/functions/runtimes/discovery/index.ts
```diff
@@ -60,7 +60,7 @@ export async function detectFromPort(
   port: number,
   project: string,
   runtime: runtimes.Runtime,
-  timeout = 10_000 /* 10s to boot up */
+  timeout = 50_000 /* 10s to boot up */
```
taeold, line 63: Thanks for suggesting the fix. Do you have any insights on 50s as magic number? (vs 30 or 20 or etc)?

taeold, line 63: Can we update the comment to match the updated timeout?

himanshu810e: Nope, just an arbitrary value that works, didn't have time to profile it. The intention was to solely increase the timeout. I wanted to parameterize it but I am not sure how to go about that.

himanshu810e: Done!

nomagick:
@taeold @joehan @hsubox76 
I need this fix real bad. 
For me, half of the time emulator doesn't work because of this.
Could you release this ASAP?

himanshu810e:
> 

So far I am literally replacing the code using good old sed, but yeah all it needs is an approval and its closed but IDK what's taking so long

mpsq:
This is causing major issues for larger repos -- is there any chance this can be merged?

harveylee:
We were also having intermittent issues with the emulator not being able to load the functions. The symptoms were consistent with the functions discovery process timing out. The issue occurred more frequently when trying to start multiple emulators together (e.g. hosting, database, storage and functions) vs. just the functions emulator (i.e. when there's less load on the machine it seems to be more successful).

I am using Debian on WSL v1.2.5.0, Windows 11 v10.0.22621, with the project directory under `~`.

I can confirm that extending timeout when `detectFromPort` is called from 10 seconds to 30 seconds on our project leads to improved reliability of the functions emulator starting. (We do this by directly editing the `functions/runtimes/discovery/index.js` file in the pnpm global dependency folder.)

mpsq:
@blidd-google Would it be possible to merge this PR :pray: ?

Isma-acv:
In my case, even increasing the time it only worked once, when I ran it again the same error came back

Isma-acv:
In my case, even increasing the time it only worked once, when I ran it again the same error came back

Isma-acv:
In my case, even increasing the time it only worked once, when I ran it again the same error came back.

image

PR #5837, opened 2023-05-14
Add support for nodejs20 (in preview)

willbattel:
To resolve the following error when trying to deploy with Node 20:

`Error: package.json in functions directory has an engines field which is unsupported. Valid choices are: {"node": 10|12|14|16|18}`

This is a mirror of https://github.com/firebase/firebase-tools/pull/5319 but changed for `nodejs20`, which [is now available as a Functions runtime in preview](https://cloud.google.com/functions/docs/concepts/nodejs-runtime). I simply copied the changes in that PR. If a Firebaser finds these changes to be correct then of course feel free to merge, otherwise I would greatly appreciate it if someone else added support swiftly so that we may deploy with this runtime. Thanks!


taeold:
Amazing - thank you for your contribution!!

taeold (approved):

PR #5465, opened 2023-01-25
Add option to initialize python functions.

taeold:
Add Python as an option for `firebase init functions`.

The python choice is guarded behind the `pythonfunctions` experiment. You can enable the experiment with `firebase experiments:enable pythonfunctions`.

This PR is a fork of https://github.com/firebase/firebase-tools/pull/4653. With this PR, https://github.com/firebase/firebase-tools/pull/4653 is now obsolete.

PR #5846, opened 2023-05-16
Svnsairam setup turtlestack cli

svnsairam:


TheIronDev:
I believe you're going to want to feature-gate this.

src/commands/init.ts
```diff
@@ -70,6 +70,11 @@ const choices = [
     name: "Extensions: Set up an empty Extensions manifest",
     checked: false,
   },
+  {
+    value: "frameworkstack",
```
taeold, line 74: can I suggest the name "frameworks" instead?

src/commands/init.ts
```diff
@@ -70,6 +70,11 @@ const choices = [
     name: "Extensions: Set up an empty Extensions manifest",
     checked: false,
   },
+  {
```
taeold, line 73: this should be behind a experiment flag - we don't want users to see this as a choice.

Checkout https://github.com/firebase/firebase-tools/pull/5465 to see an example of how you can create a new experimental feature and to use it to change behavior of init!

src/init/features/frameworkstack/index.ts
```diff
@@ -0,0 +1,66 @@
...
+      type: "input",
+      default: "acme-inc-web",
+      message: "Create a name for your service [1-64 characters]",
+    },
+    setup.frameworkstack
+  );
+
+  await promptOnce(
+    {
+      name: "regionName",
```
taeold, line 41: nit* prefer `region` over `regionName`

PR #5867, opened 2023-05-19
Improve error message when function deployment fails due to quota issues

taeold:
When function deployment fails due to insufficient quota on Cloud Run, the error message is difficult to understand:

```
i  functions: creating Node.js 18 function helloWorld(us-east4)...
⚠  functions: HTTP Error: 400, Could not create Cloud Run service helloworld. spec.template.spec.containers.resources.limits.cpu: Invalid value specified for cpu. For the specified value, maxScale may not exceed 10.
Consider running your workload in a region with greater capacity, decreasing your requested cpu-per-instance, or requesting an increase in quota for this region if you are seeing sustained usage near this limit, see https://cloud.google.com/run/quotas. Your project may gain access to further scaling by adding billing information to your account.
```

Updated error message, which we hope is clearer and actionable:

```
i  functions: creating Node.js 18 (2nd Gen) function helloWorld(us-east4)...
⚠  functions: Your current project quotas don't allow for the current max instances setting of 100. Either reduce this function's maxInstances, or request a quota increase on the underlying Cloud Run service at https://cloud.google.com/run/quotas.
⚠  functions: You can adjust the max instances value in your function's runtime options:
	setGlobalOptions({maxInstances: 10})
```

### DIFF
--- a/src/deploy/functions/functionsDeployHelper.ts
+++ b/src/deploy/functions/functionsDeployHelper.ts
@@ -125,6 +125,16 @@ export function getEndpointFilters(options: { only?: string }): EndpointFilter[]
 }
 
 /**
+ * Get human friendly name for the given function platform
+ */
+export function getHumanFriendlyPlatformName(platform: backend.Endpoint["platform"]): string {
+  if (platform === "gcfv1") {
+    return "1st Gen";
+  }
+  return "2nd Gen";
+}
+
+/**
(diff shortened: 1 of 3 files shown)


jhuleatt (approved):
Awesome, thanks Daniel! I think this will clear up a lot of confusion around quota issues.

PR #5882, opened 2023-05-23
Adding package-lock.json to standalone/

joehan:
### Description
Adding a package-lock to standalone/. This was flagged as a potential security issue in internal bug 279235012


jamesdaniels (approved):
Thanks for jumping on this

Issue #5875, opened 2023-05-22
Firebase Cloud Functions 2nd Gen Scheduler does not work correctly on the emulators.


gladevise:
Firebase Cloud Functions 2nd Gen Scheduler does not work correctly on the emulators.

When I start the emulators and try to execute a function with `firebase functions:shell` , the following error occurs.

```sh
firebase > scheduleTestV2()
'Successfully invoked function.'
firebase > ⚠  functions: TypeError: req.header is not a function
    at httpFunc (/home/username/workspace/tests/firebase-cloud-functions-schedule-v2-not-working-emulator/functions/node_modules/.pnpm/firebase-functions@4.3.1_firebase-admin@11.8.0/node_modules/firebase-functions/lib/v2/providers/scheduler.js:63:26)
    at /home/username/workspace/tests/firebase-cloud-functions-schedule-v2-not-working-emulator/functions/node_modules/.pnpm/firebase-functions@4.3.1_firebase-admin@11.8.0/node_modules/firebase-functions/lib/v2/trace.js:16:28
    at /home/username/.nvm/versions/node/v18.16.0/lib/node_modules/firebase-tools/lib/emulator/functionsEmulatorRuntime.js:524:16
    at runFunction (/home/username/.nvm/versions/node/v18.16.0/lib/node_modules/firebase-tools/lib/emulator/functionsEmulatorRuntime.js:506:15)
    at runCloudEvent (/home/username/.nvm/versions/node/v18.16.0/lib/node_modules/firebase-tools/lib/emulator/functionsEmulatorRuntime.js:523:11)
    at processBackground (/home/username/.nvm/versions/node/v18.16.0/lib/node_modules/firebase-tools/lib/emulator/functionsEmulatorRuntime.js:490:16)
    at /home/username/.nvm/versions/node/v18.16.0/lib/node_modules/firebase-tools/lib/emulator/functionsEmulatorRuntime.js:689:27
    at Layer.handle [as handle_request] (/home/username/.nvm/versions/node/v18.16.0/lib/node_modules/firebase-tools/node_modules/express/lib/router/layer.js:95:5)
    at next (/home/username/.nvm/versions/node/v18.16.0/lib/node_modules/firebase-tools/node_modules/express/lib/router/route.js:144:13)
    at next (/home/username/.nvm/versions/node/v18.16.0/lib/node_modules/firebase-tools/node_modules/express/lib/router/route.js:140:7)
⚠  Your function was killed because it raised an unhandled error.
[...]

PR #5891, opened 2023-05-24
Fixes v2 scheduled function emulation by changing signature to HTTP instead of Cloud Event

blidd-google:
The functions emulator was treating 2nd gen scheduled functions as triggered by "Cloud Events" when they should have been treating them as triggered by HTTP. 1st gen scheduled functions should continue to be classified as "Event"-triggered. Under the hood, 2nd gen scheduled functions are simply HTTP functions that listen to a Cloud Scheduler, while 1st gen scheduled functions listen on Pub/Sub topics for events published by a Cloud Scheduler.

Fixes https://github.com/firebase/firebase-tools/issues/5875.


taeold (approved):
Great find!

avilao:
Any chance someone can take a look at this? 🙏

Issue #5890, opened 2023-05-24
Deployment of Nextjs project fails after updating to 12.2.0


ydalir:
Hello, we ran into an issue today with our firebase deployment. The github action we use installs the latest firebase-tools version, e.g. 12.2.0, which breaks our build. Explicitly setting the firebase-tools version back to 12.1.0 fixes the issue. We suspect the issue may be some of the changes around rewrites in https://github.com/firebase/firebase-tools/pull/5716.

Because the problem arose in a CI environment, posting logs and running commands was difficult. Let me know if you need more information than provided here.

### [REQUIRED] Environment info

**firebase-tools:** 12.2.0

**Platform:** Ubuntu 22.04.2 LTS (Github Action)

### [REQUIRED] Test case

```
jobs:
  deploy-staging:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - name: Install
        run: yarn install
      - uses: FirebaseExtended/action-hosting-deploy@v0
        with:
          repoToken: ***
          firebaseServiceAccount: ***
          projectId: ***
          channelId: live
          entrypoint: ./***
        env:
          FIREBASE_CLI_EXPERIMENTS: webframeworks
```
### [REQUIRED] Steps to reproduce

Trigger github action with above workflow configuration

### [REQUIRED] Expected behavior

Github Action correctly builds and deploys entire Nextjs project to firebase, or fails and throws an error.

### [REQUIRED] Actual behavior

Github Action does not build project correctly but does not throw an error, and deploys mangled build.

gabriellimoni:
Same issue here but using Gitlab CI. Downgrading to version 12.1.0 fixes

jamesdaniels:
I wonder if it's building a development bundle? The other thing we added was a new environment variable. `FIREBASE_FRAMEWORKS_BUILD_TARGET` it's inferred from NODE_ENV out of box. 

If you set this environment variable to "production" does this work around the CI issue?

jamesdaniels:
This should be resolved in firebase-tools `12.2.1`. 

Sorry the interruption in your deployment pipelines & thank you for trying out the web frameworks experiment!

sam-hyperact:
Still not resolved for me, just tried. Stopped working between 12.1 & 12.2. The build is still failing. It triggers this error in the build logs:

```
2023-05-25T07:05:45.6217026Z npm WARN EBADENGINE Unsupported engine ***	
[...]

PR #5892, opened 2023-05-24
Always assume build target of production on deploy

jamesdaniels:
I'm guessing this is the root cause of #5890


chalosalvador (approved):
lgtm

Issue #5880, opened 2023-05-23
Webframeworks do not deploy nextjs ssr{projectId} function when using project aliases


pzmudzinski:
### [REQUIRED] Environment info

**firebase-tools:12.1.0**

**Platform:macOS**

### [REQUIRED] Test case

Repository with two firebase projects set as aliases. Add Nextjs project inside it.

### [REQUIRED] Steps to reproduce

```
firebase use {alias}
firebase deploy --only hosting
```

### [REQUIRED] Expected behavior

`ssr{projectId}` function should be deployed but it does not and see warning in deploy command:
```
hosting[...]: Unable to find a valid endpoint for function `ssr...`, but still including it in the config
```
I got it working when using default project. It stopped working after adding new project alias and trying to deploy the same website under different project. I am assuming that's
[logs.txt](https://github.com/firebase/firebase-tools/files/11543080/logs.txt)
 the reason but maybe something got broken with latest versions of CLI?

### [REQUIRED] Actual behavior

After deploying page returns 404.
Logs attached:
[logs.txt](https://github.com/firebase/firebase-tools/files/11543089/logs.txt)

pzmudzinski:
**Update**: downgrading firebase-tools back to 11.30.0 fixes the issue so this is definitely buggy there ❗

Snugug:
I'm having a similar problem with Astro (so can't downgrade to 11.x).  When trying to deploy, I get the following output after successfully building the static bits of my site:

```
i  hosting: The following function(s) are pinned to site {projectId} and will be deployed as well: ssr{projectId}

=== Deploying to '{projectId}'...

i  deploying functions, hosting

Error: No function matches given --only filters. Aborting deployment.
 ELIFECYCLE  Command failed with exit code 1.
```

rodydavis:
> I'm having a similar problem with Astro (so can't downgrade to 11.x). When trying to deploy, I get the following output after successfully building the static bits of my site:
> 
> ```
> i  hosting: The following function(s) are pinned to site {projectId} and will be deployed as well: ssr{projectId}
> 
> === Deploying to '{projectId}'...
> 
> i  deploying functions, hosting
> 
> Error: No function matches given --only filters. Aborting deployment.
>  ELIFECYCLE  Command failed with exit code 1.
> ```
> 
> I'm using Astro hybrid rendering.
> 
> I'm not using project aliases.

This is happening with astro standalone mode too:

```
output: 'server',
  adapter: node({
[...]

PR #5898, opened 2023-05-25
Support Astro hybrid rendering

jamesdaniels:

### DIFF
--- a/src/frameworks/astro/index.ts
+++ b/src/frameworks/astro/index.ts
@@ -25,7 +25,7 @@ export async function discover(dir: string): Promise<Discovery | undefined> {
   if (!getAstroVersion(dir)) return;
   const { output, publicDir: publicDirectory } = await getConfig(dir);
   return {
-    mayWantBackend: output === "server",
+    mayWantBackend: output !== "static",
     publicDirectory,
   };
 }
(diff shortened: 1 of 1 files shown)


Snugug:
Thanks for working on this! I'm testing it out using `firebase emulators:start` and it's rendering the site from the server, which is great, but it's also not using the pre-rendered pages as I'd expect. Is there additional config I need to get prerendering to work under emulators?

Snugug:
> @Snugug by default we use fast dev-mode (HMR) for `emulators:start`. To emulate behavior closest to production use `FIREBASE_FRAMEWORKS_BUILD_TARGET=production firebase emulators:start`

Thanks! This got the static site running as expected, but routes I expect the server to handle aren't working.  I get a blank white page with the following error:

```
Function us-central1-ssr{{project}} does not exist, valid functions are:
```

`{{project}}` is my project name. There are no valid functions listed. Is there other config I'm missing?

austincrim:
@jamesdaniels lgtm.

@Snugug I was able to run locally and deploy using the Astro starter project with the following `astro.config.mjs`:
```js
import { defineConfig } from 'astro/config'

import node from '@astrojs/node'

// https://astro.build/config
export default defineConfig({
  output: 'hybrid',
  adapter: node({
    mode: 'middleware'
  }),
  experimental: {
    hybridOutput: true
  }
})
```

I added `export const prerender = false` to one of my `.astro` files and then ran `FIREBASE_FRAMEWORKS_BUILD_TARGET=production firebase emulators:start` with the version of `firebase-tools` that James specified in the original post.

austincrim (approved):

jamesdaniels:
@Snugug maybe you're issue is the same as #5880? Are you using aliases?

Snugug:
> @Snugug maybe your issue is the same as #5880? Are you using aliases?

I am not; only have `default` in `.firebaserc`

Issue #5869, opened 2023-05-20
[NEXT]: new framework with app router is unaccessible after deployment to firebase/cloud-functions


RichardIvan:
### [REQUIRED] Environment info

global firebase v - 11.30.0

Repository is using pnpm. The next app is within a workspace.

project firebase
"firebase": "^9.22.0",
"firebase-tools": "^12.1.0"

firebase config
```json
{
  "hosting": {
    "source": "./",
    "ignore": [
      "firebase.json",
      "**/.*",
      "**/node_modules/**"
    ],
    "frameworksBackend": {
      "region": "europe-west1",
      "cpu": 1,
      "maxInstances": 1,
      "invoker": "public",
      "cors": true,
      "ingressSettings": "ALLOW_ALL"
    }
  }
}

```

**firebase-tools:**

**Platform:**
macOS

### [REQUIRED] Test case
N/A

### [REQUIRED] Steps to reproduce
firebase deploy of the frameworks enabled nextjs app

### [REQUIRED] Expected behavior
example.com/api/endpoint does not throw 500, Invariant: failed to start render worker Error: listen EADDRNOTAVAIL: address not available

### [REQUIRED] Actual behavior
Local, emulator works as expected including nextjs /app/api/* function
Deployment is successful and the Nextjs framework app.

When deployed, GET request function works as expected.
When trying to access POST request, we see a failure even if the functions are the same - basically trying to return 200 status with empty object.

Requesting POST API endpoint fails with 500. "Invariant: failed to start render worker Error: listen EADDRNOTAVAIL: address not available"
Logs in GCP don't provide much of additional information.

justinh00k:
I am also getting this error on the cloud function of my NextJs firebase hosting app.
`"Invariant:` failed to start render worker Error: listen EADDRNOTAVAIL: address not available 199.36.158.100"

rchaz:
I'm hitting the same error

rchaz:
Error on the SSR function deployment post upgrade to nextjs13 app router - 

Invariant: failed to start render worker Error: listen EADDRNOTAVAIL: address not available 199.36.158.100
at Server.setupListenHandle [as _listen2] (node:net:1723:21)
at listenInCluster (node:net:1788:12)
at GetAddrInfoReqWrap.doListen [as callback] (node:net:1937:7)
at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:109:8) {
code: 'EADDRNOTAVAIL',
[...]

PR #5912, opened 2023-05-29
fix Next.js hostname issue

michaelangrivera:
### Description

Attempts to fix https://github.com/firebase/firebase-tools/issues/5869.

I noticed that the `hostname` is populated here:
https://github.com/FirebaseExtended/firebase-framework-tools/blob/4ae1bb0cedd3423e102c4360a2c91d644fa5627b/src/next.js/index.ts#LL30C22-L30C22 , which is the root of the issue [here](https://github.com/firebase/firebase-tools/issues/5869) and here https://github.com/vercel/next.js/issues/49777

Perhaps `req` object is not resolving correctly without explicit naming? 🤷 

Also on a side note, can you allow us to customize this script? Should be simple to implement considering it's almost already implemented via `FrameworksBackendOptions`. You can just add another option to point to a output `server.js` file. More flexibility, the better, IMO. 

### Scenarios Tested

From my own experimentation, explicitly defining the return seems to fix the issue (weirdly enough). I've deployed a Next app (13.4.2) to my URL here https://staging.activitystreak.app/. 

### Sample Commands

NA

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,9 +72,6 @@
     "url": "https://github.com/firebase/firebase-tools/issues"
   },
   "homepage": "https://github.com/firebase/firebase-tools",
-  "publishConfig": {
-    "registry": "https://wombat-dressing-room.appspot.com"
-  },
   "nyc": {
     "require": [
       "ts-node/register"
--- a/package.json
+++ b/package.json
@@ -72,6 +72,9 @@
     "url": "https://github.com/firebase/firebase-tools/issues"
   },
   "homepage": "https://github.com/firebase/firebase-tools",
+  "publishConfig": {
+    "registry": "https://wombat-dressing-room.appspot.com"
+  },
   "nyc": {
     "require": [
       "ts-node/register"
--- a/src/frameworks/index.ts
+++ b/src/frameworks/index.ts
@@ -461,7 +461,7 @@ export async function prepareFrameworks(
       //   // continue
       // });
 
-      const npmrcPath = getProjectPath(".npmrc");
+      const npmrcPath = getProjectPath("..", "..", ".npmrc");
       if (await pathExists(npmrcPath)) {
         console.log("\x1b[36m%s\x1b[0m", `copying ${npmrcPath} to ${functionsDist}`);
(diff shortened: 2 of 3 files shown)


michaelangrivera:
@jamesdaniels

jamesdaniels:
Unfortunately changes to the `FrameworksBackendOptions` aren't really tenable, as it would require we put together internal documentation, presenting to the Firebase API council, and justify it against the long term roadmap. Since it's an API surface that we'd need to support long-term.

We'll look into this hostname issue though, thanks for your explorations.

slecoustre-vulog:
To see there is the same error when you deploy on aws.  I think it comes from nextjs 

https://github.com/vercel/next.js/issues/49777

@michaelangrivera @jamesdaniels

n00dl3s:
Not sure if its exclusive to firebase, seems to be happening in a dev environment using this template. The only thing I did was update all dependencies to their newest versions. My workaround thus far was to use a PUT on a different route to handle the request. 

https://github.com/saleor/nextjs-commerce

DhruvRaiPuri:
> Unfortunately "quick"/"simple" changes to the `FrameworksBackendOptions` aren't really tenable, as it would require we put together internal documentation, present to the Firebase API council, and justify it against the long term roadmap—since it's an API surface that we'd need to support long-term.
> 
> We'll look into this hostname issue though, thanks for your explorations.

I understand that change management processes have to be adhered to, but an app I have started building has failed with cloud run service/function failing with hostname related issue.

I tried adding the hostname '0.0.0.0' manually in Cloud Run service's environment variables by 'editing' the cloud run service definition, but it fails with image not found error - Image 'us-central1-docker.pkg.dev/<redacted project name>/gcf-artifacts/ssr<redacted project name>:version_1' not found.

Now, I have 2 options - 
1. Wait for Firebase's bureaucratic processes to get around to approving this fix
2. Go to a different provider like Vercel or Netlify

I want to wait for option 1 - but there should be an expected timeline for that to happen. Is it possible to give a timeline?

PR #5913, opened 2023-05-30
Ensure Extensions TOS api is enabled before calling it

### LINKED ISSUE #5855: Firebase Deploy Fails TOS acceptance
I can run cmd to deploy project locally, but it fails in CI with:

```
extensions: By installing an extension instance onto a Firebase project, you accept the Firebase Extensions User Terms of Service: https://firebase.google.com/terms/extensions/user

Error: HTTP Error: 403, Firebase Extensions Terms of Service Private API has not been used in project XYZ before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/firebaseextensionstos-pa.googleapis.com/overview?project=XYZ then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
yes: standard output: Broken pipe
```

First of all, that link does not work / exist.

Second, TOS has been accepted, I run that cmd locally and accepted TOS (even though this should not be the required thing if you run CI)

Also there should be a way to do that manually if is such big of a deal.

joehan:
### Description
Ensure that the TOS api is enabled before we call it. Fixes #5855

### DIFF
--- a/src/extensions/tos.ts
+++ b/src/extensions/tos.ts
@@ -5,6 +5,7 @@ import { confirm } from "../prompt";
 import { FirebaseError } from "../error";
 import { logPrefix } from "./extensionsHelper";
 import * as utils from "../utils";
+import { ensure } from "../ensureApiEnabled";
 
 const VERSION = "v1";
 const extensionsTosUrl = (tos: string) => `https://firebase.google.com/terms/extensions/${tos}`;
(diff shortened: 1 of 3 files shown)


apascal07 (approved):

src/extensions/tos.ts
```diff
@@ -107,6 +109,7 @@ export async function acceptLatestAppDeveloperTOS(
   ) {
     throw new FirebaseError("You must accept the terms of service to continue.");
   }
+  await ensure(projectId, extensionsTOSOrigin, "extensions", true);
```
pavelgj, line 112: can't do this... it's a private API

pavelgj (changes requested):

PR #5917, opened 2023-05-31
Switch ext:dev:init to default billingRequire to true

joehan:
### Description
Switch `ext:dev:init` to default billingRequire to true.  Since Cloud Functions  require billing now, this should almost always be true.


nohe427 (approved):

nohe427:
Thank you!

apascal07 (approved):

PR #5925, opened 2023-06-01
Remove LOCATION param from extension.yaml template.

joehan:
### Description
Remove LOCATION param from `extension.yaml` template. This param is no automatically injected for all extensions using v1 functions, so no need to include it here.


nohe427 (approved):

ifielker (approved):
LGTM

templates/extensions/extension.yaml
```diff
@@ -39,8 +39,6 @@ resources:
     description: >-
```
pavelgj, line 39: Doesn't let me comment on the lines above for some reason... but on line 18 the sourceUrl is
`sourceUrl: https://github.com/firebase/firebase-tools/tree/master/templates/extensions`

Can we change it to something that's more obvious that it needs to be changed.
ex: `https://github.com/MY_ORG/MY_REPO/MY_EXTENSION`

PR #5954, opened 2023-06-09
Emulators in the VSCode plugin

christhompsongoogle:
Add an emulator suite panel for launching the emulators in the VSCode extension


firebase-vscode/common/messaging/protocol.ts
```diff
@@ -117,6 +131,26 @@ export interface ExtensionToWebviewParamsMap {
...
+  notifyRunningEmulatorInfo: RunningEmulatorInfo ;
+  notifyEmulatorImportFolder: { folder: string };
 }
 
 export type MessageParamsMap = WebviewToExtensionParamsMap | ExtensionToWebviewParamsMap;
+
+/**
+ * Info to display in the UI while the emulators are running
+ */
+export interface RunningEmulatorInfo {
```
hsubox76, line 144: There's a `types.d.ts` file adjacent to this one, these interfaces are probably a better fit there (that's where Message lives, and ChannelWithId which seems on par with these - info sent thru messages).

firebase-vscode/src/cli.ts
```diff
@@ -172,6 +177,33 @@ export async function initHosting(options: { spa: boolean; public: string }) {
   await initAction("hosting", commandOptions);
 }
 
+export async function emulatorsStart(firebaseJson: FirebaseConfig, emulatorUiSelections: EmulatorUiSelections) {
+  const commandOptions = await getCommandOptions(firebaseJson, {
+    config: { ...firebaseJson },
```
hsubox76, line 182: Doesn't command.prepare fill this in automatically?

src/command.ts
```diff
@@ -281,8 +281,8 @@ export class Command {
       setupLoggers();
     }
 
-    if (getInheritedOption(options, "config")) {
-      options.configPath = getInheritedOption(options, "config");
+    if (getInheritedOption(options, "configPath")) {
```
hsubox76, line 284: I see some rearranging of config. Does this affect the way the CLI works? I think it might. I think the name of the option that you can type in the CLI is "--config" (search for it [here](https://firebase.google.com/docs/cli#the_firebasejson_file)) and this is honoring the way it comes in as user input. I don't fully understand the CLI but I think that this might mean the user has to type `--configPath` to do the same thing which is kind of a breaking change.

firebase-vscode/src/options.ts
```diff
@@ -1,19 +1,20 @@
...
+import * as commandUtils from "../../src/emulator/commandUtils";
 
 /**
  * User-facing CLI options
  * Passed to command.prepare()
  */
 
-interface CliOptions extends Omit<BaseOptions, "config"> {
-  config: string;
+interface CliOptions extends BaseOptions {
```
hsubox76, line 15: See the other comment about config but I think the thing is that `command.prepare` takes CLI options as passed in directly from user input, and then it does some renaming, and this `CliOptions` interface reflects the type that `command.prepare` takes, and in that case, `config` is a string.

firebase-vscode/webviews/SidebarApp.tsx
```diff
@@ -141,6 +153,7 @@ export function SidebarApp() {
           }}
         />
       )}
+      {firebaseJson && <EmulatorPanel firebaseJson={firebaseJson} />}
```
hsubox76, line 156: Probably gate this on `!!projectId` so that it's consistent with the deploy/hosting init panels? I guess you don't need to be logged in to use emulators, or else also `!!userEmail`.

firebase-vscode/webviews/EmulatorPanel.tsx
```diff
@@ -0,0 +1,249 @@
...
+  // TODO(christhompson): Add more info and sort this.
+  function formatEmulatorRunningInfo(emulatorInfos: EmulatorInfo[]): string {
+    return emulatorInfos
+      .map((info) => info.name)
+      .filter((name) => name !== "logging")
+      .join("<br/>");
+  }
+
+  return (
+    <PanelSection>
```
hsubox76, line 162: Can you pass it a `title` prop? Then it'll have a nice header bar and also be collapsible.

firebase-vscode/webviews/EmulatorPanel.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+  function toggleExportOnExit() {
+    console.log("toggle export on exit");
+    const selections: EmulatorUiSelections = emulatorUiSelections;
+    selections.exportStateOnExit = !selections.exportStateOnExit;
+    setEmulatorUiSelectionsAndSaveToWorkspace(selections);
+  }
+
+  function projectIdChanged(event: React.ChangeEvent<HTMLInputElement>) {
+    console.log("projectIdChanged: " + event.target.value);
```
hsubox76, line 132: Use webLogger for these? FYI it doesn't do `.log` because the treatment of `.log()` in the CLI logger seemed a little different and I didn't know what was up with it, so `info` or `debug` maybe.

firebase-vscode/common/messaging/protocol.ts
christhompsongoogle: Done

firebase-vscode/webviews/SidebarApp.tsx
christhompsongoogle: Really we only care about firebaseJson existence - we could combine that with userEmail if we want to only show the panel after the user has logged in, something like

`!!userEmail && !!firebaseJson && <...>`

WDYT?

firebase-vscode/webviews/EmulatorPanel.tsx
christhompsongoogle: Done

firebase-vscode/webviews/EmulatorPanel.tsx
christhompsongoogle: Whoops, thanks for pointing that out, these should have been removed/logged already

src/command.ts
christhompsongoogle: Good point I should have opened this for discussion - I was getting an undefined error by this, likely because this makes assumptions about the structure of config vs configPath. I assumed at the time that this was just a typo mistake that hadn't been caught yet. I can't find anything on the internet about it

@joehan do you have any thoughts?

firebase-vscode/src/options.ts
christhompsongoogle: Hmm - I see what you're saying - this config though is overloaded to the point of being fragile. At some point the `options.config` represents the firebase.json compiled config right? How can we separate the two so as not to cross wires. Is that what command.prepare does, replace options.config with an object instead of a string?

src/command.ts
christhompsongoogle: Nvm reverting this and making it work :)

firebase-vscode/src/cli.ts
```diff
@@ -172,6 +177,32 @@ export async function initHosting(options: { spa: boolean; public: string }) {
   await initAction("hosting", commandOptions);
 }
 
+export async function emulatorsStart(firebaseJson: FirebaseConfig, emulatorUiSelections: EmulatorUiSelections) {
+  const commandOptions = await getCommandOptions(undefined, {
```
joehan, line 181: Seems like this  should take firebaseJson as the first arg instead of undefined?

firebase-vscode/src/cli.ts
```diff
@@ -172,6 +177,32 @@ export async function initHosting(options: { spa: boolean; public: string }) {
...
+export async function stopEmulators() {
+  await stopAllEmulators();
+}
+
+export function listRunningEmulators(): EmulatorInfo[] {
+  return EmulatorRegistry.listRunningWithInfo();
+}
+
+export function getEmulatorUiUrl(): string | undefined {
+  const asdf = EmulatorRegistry.url(Emulators.UI);
```
joehan, line 202: asdf -> a real variable name

firebase-vscode/src/workflow.ts
```diff
@@ -369,10 +373,54 @@ export function setupWorkflow(
...
         { projectId, folderPath: currentOptions.cwd });
-      
+
       await fetchChannels();
     }
   }
+  broker.on(
+    "launchEmulators",
+    async ({ firebaseJson, emulatorUiSelections }) => {
+      emulatorsStart(firebaseJson, emulatorUiSelections).then(() => {
```
joehan, line 383: Prefer await emualtorsStart() over promise chaining, especially sincd were already in an async function

firebase-vscode/webviews/EmulatorPanel.tsx
```diff
@@ -0,0 +1,250 @@
...
+  /**
+   * Called when import folder changes.
+   */
+  function selectedImportFolder(event: any) {
+    event.preventDefault();
+    broker.send("selectEmulatorImportFolder");
+  }
+
+  function toggleExportOnExit() {
+    webLogger.debug("toggle export on exit");
```
joehan, line 126: nit: include what it is being toggled to

firebase-vscode/webviews/EmulatorPanel.tsx
```diff
@@ -0,0 +1,250 @@
...
+      .join("<br/>");
+  }
+
+  return (
+    <PanelSection title="Emulators">
+      <h2>Launch the Emulator Suite</h2>
+      {/* TODO(christhompson): Insert some education links or tooltips here. */}
+      <Spacer size="xxlarge" />
+      Current project ID:
+      {/* TODO(christhompson): convert this into a demo- prefix checkbox or something. */}
```
joehan, line 168: I like this idea a lot - much easier way to let peeps know about demo- projects

src/emulator/controller.ts
```diff
@@ -854,6 +843,16 @@ export async function startAll(
     await startEmulator(loggingEmulator);
   }
 
+  if (showUI && !shouldStart(options, Emulators.UI)) {
+    hubLogger.logLabeled(
+      "WARN",
+      "emulators",
+      "The Emulator UI is not starting, either because none of the emulated " +
+        "products have an interaction layer in Emulator UI or it cannot " +
```
joehan, line 851: Nit: Consider simplifying this message a bit - 'interaction layer' feels like unnecessary jargon here, for example

joehan (approved):

firebase-vscode/src/cli.ts
christhompsongoogle: Whoops, thanks for that

firebase-vscode/src/workflow.ts
christhompsongoogle: good point, fixed!

firebase-vscode/webviews/EmulatorPanel.tsx
christhompsongoogle: Done

src/emulator/controller.ts
christhompsongoogle: Done

firebase-vscode/src/cli.ts
christhompsongoogle: This is a remnant of a previous build - removed this for now since the workflow.ts caches the active firebaseJSON there's no need to pass it around.

firebase-vscode/src/options.ts
```diff
@@ -12,16 +12,20 @@ import * as commandUtils from "../../src/emulator/commandUtils";
  * Passed to command.prepare()
  */
 
-interface CliOptions extends BaseOptions {
-  exportOnExit?: boolean | string
-  import?: string
+interface CliOptions extends Omit<BaseOptions, "config"> {
+  config: string;
+  exportOnExit?: boolean | string;
```
hsubox76, line 17: I just realized, are these legit CLI options that are used in the CLI? If so maybe we can add it to BaseOptions and the CLI can benefit from this typing too?

hsubox76 (approved):

firebase-vscode/src/options.ts
christhompsongoogle: Good point - fixed

PR #5968, opened 2023-06-12
VSCode plugin refactoring and UX

hsubox76:
Refactoring:
- Use CLI's existing file read/write utilities for firebaserc and firebase.json - create a separate file, config.ts, for the read/writes
- Streamline some messages
- Add VSCode settings for logging, npm path, and turning on web frameworks

UX fixes:
- Add "SplitButton" dropdown picker for preview channels (replaces radio buttons) - used code provided by Roman
- Extract strings used in UX into one file, for easy changes in the future
- Add progress indicator for hosting init as web frameworks init can take several seconds


firebase-vscode/.vscodeignore
```diff
@@ -16,3 +16,4 @@ vsc-extension-quickstart.md
 webpack.*.js
 ../
 *.zip
+node_modules/
```
hsubox76, line 19: After moving some deps from devdependencies to dependencies in a previous PR, vscode started bundling the dependencies, which are not needed, as webpack should bundle them all.

firebase-vscode/common/messaging/protocol.ts
```diff
@@ -4,20 +4,16 @@
...
   /**
-   * Ask extension for env variables
+   * Ask extension for initial data
    */
-  getEnv: {};
-  /**
-   * User management
-   */
-  getUsers: {};
+  getInitialData: {};
```
hsubox76, line 16: This is bundling getEnv/getUsers/getChannels/getSelectedProject into one message since they're all requested by the webview upon init and not at any other time (they may be *pushed* by the extension at other times but not pulled by the webview)

firebase-vscode/package.json
```diff
@@ -3,7 +3,7 @@
   "displayName": "firebase-vscode",
   "publisher": "firebase",
   "description": "VSCode Extension for Firebase",
-  "version": "0.0.9",
+  "version": "0.0.22",
```
hsubox76, line 6: FYI I have to keep bumping the version every time I want to push to Monospace to test

firebase-vscode/src/configs.ts
```diff
@@ -0,0 +1,144 @@
+import * as fs from "fs";
+import * as path from "path";
+import * as vscode from "vscode";
+import { workspace } from "vscode";
+import { ExtensionBrokerImpl } from "./extension-broker";
+import { updateOptions, currentOptions } from "./options";
+import { RC } from "../../src/rc";
+import { Config } from "../../src/config";
+import { pluginLogger } from "./logger-wrapper";
```
christhompsongoogle, line 9: nit: to distance from the existing "Config.ts" can we change the name here to configUtils or configParsers or the like?

firebase-vscode/webviews/components/ui/SplitButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+          onClick={() => setMenuOpen(true)}
+          appearance={appearance || "secondary"}
+          {...(props as any)}
+        >
+          <Icon icon="chevron-down" />
+        </VSCodeButton>
+      </div>
+    </>
+  );
+};
```
christhompsongoogle, line 53: nit: NL

christhompsongoogle (approved):

yuchenshi (approved):
High level LGTM (not a line-by-line review)

firebase-vscode/package.json
joehan: If monospace will allow it, consider using prerelease tags (ie `0.0.22-alpha.0`) instead of patch versions when testing

firebase-vscode/src/configs.ts
hsubox76: You mean change the name of this file? Done.

firebase-vscode/src/config-files.ts
```diff
@@ -0,0 +1,144 @@
...
+      pluginLogger.error(e.message);
+      throw e;
+    }
+  }
+  try {
+    firebaseJSON = Config.load({ configPath: path.join(configPath, 'firebase.json') });
+  }
+  catch (e) {
+    if (e.message.includes('could not locate')
+      || e.message.includes('Could not load config file')) {
```
joehan, line 67: This error message matching seems a bit brittle - might be worth changing Config.load to return an error with status 404 and checking e.status instead

joehan (approved):

firebase-vscode/package.json
hsubox76: Sounds good, I will give it a try.

PR #5969, opened 2023-06-12
Increased create extensions instance timeout to 1h to match the backend

pavelgj:


pavelgj:
Updated the CHANGELOG.

PR #5974, opened 2023-06-13
Use AppEngine v1alpha api version

tonyjhuang:
### Description

Changes app engine endpoint for fetching default bucket to `v1alpha` as `v1` no longer supports requests from free tier projects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,1 @@
-- Update App Engine api version to v1alpha (#5955)
+- Fixed issue where Storage rules could not be deployed to projects without a billing plan. (#5955)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- Update App Engine api version to v1alpha (#5955)
```
```suggestion
- Fixed issue where Storage rules could not be deployed to projects without a billing plan. (#5955)
```

joehan (approved):
LGTM for a temporary fix - also, take a look at https://github.com/firebase/firebase-tools/pull/5975 for a longer term fix

joehan (changes requested):
Actually, looks like the storage deploy integration test is failing with this change. Could you test this change out a little more to confirm that it will actually fix things?

PR #5987, opened 2023-06-15
Patch fix for issues caused by adding type check in #5906

aalej:
### Description

 
Type checking `reqBody.email` and `reqBody.password` introduced in #5906 caused some issues with the Auth Emulator UI.

Request made by the emulator UI ofter contain a blank values by default. Creating a user via the Emulator UI using phone number will provide a blank email and password string. This will result in the request received by the Auth Emulator to look like:
```
reqBody: {
  customAttributes: '',
  displayName: '',
  photoUrl: '',
  email: '',
  password: '',
  phoneNumber: '+1230104124',
  ....
} 
```

Adding type checking of `reqBody.email` and `reqBody.password` causes the emulator to always assert a valid email address value. Proposed solution is to remove type check for both `reqBody.email` and `reqBody.password`. 

To keep the solution in #5906, add a condition to specifically check if `reqBody.email` is an empty string and a provider is defined `(reqBody.email === "" && provider)`.

### Scenarios Tested

| Scenario                                 | Prod                                              | Emulator                                          | PR                                                |
|------------------------------------------|---------------------------------------------------|---------------------------------------------------|---------------------------------------------------|
| email="" password=""                     | Error (auth/invalid-email)                        | Creates a user with blank identifier and provider | Error (auth/invalid-email)                        |
| email="" password="123456"               | Error (auth/missing-email)                        | Error (auth/missing-email)                        | Error (auth/missing-email)                        |
| email="test@gmail.com" password="123456" | Creates a user with identifier and email provider | Creates a user with identifier and email provider | Creates a user with identifier and email provider |

Emulator UI
1. Tested creating a user using only a phone number
2. Tested toggling the verified email toggle in the Auth Emulator UI


src/emulator/auth/operations.ts
```diff
@@ -198,13 +198,13 @@ async function signUp(
     }
   }
 
-  if (typeof reqBody.email === "string") {
+  if (reqBody.email || (reqBody.email === "" && provider)) {
```
christhompsongoogle, line 201: Nit: add a comment here describing the conditions that we're testing for and why

christhompsongoogle (approved):

src/emulator/auth/operations.ts
aalej: Added a comment describing the additional condition

aalej:
Closing this PR, changes are fixed in https://github.com/firebase/firebase-tools/pull/6127

PR #5989, opened 2023-06-15
Support lifecycle hooks for function codebases

sam-throne:
Sample command:

`firebase deploy --only functions:metrics:report,functions:metrics:process`

Previous behavior:
Run lifecycle hooks of all codebases in the project

New behavior:
Run lifecycle hooks of relevant codebases only (`metrics` in the case of the sample command)


sam-throne:
fix #6002

PR #5993, opened 2023-06-15
Discovery: Added node runtime.

svnsairam:
### Description

Created this branch in reference to: https://github.com/firebase/firebase-tools/pull/5963

1. node.ts -> Finds best matching commands for the codebase.
2. node.spec.ts -> Tests the functionality.


src/frameworks/compose/discover/runtime/node.ts
```diff
@@ -0,0 +1,276 @@
...
+    packageJSON: PackageJSON,
+    packageManager: string
+  ): Promise<Record<string, string>> {
+    try {
+      let dependencies = {};
+      if (packageManager === NPM) {
+        dependencies = await this.getDependenciesForNPM(fs, packageJSON);
+      } else if (packageManager === YARN) {
+        dependencies = await this.getDependenciesForYARN(fs, packageJSON);
+      }
```
taeold, line 136: suggestion - let's just get deps from package.json, regardless of package manager.

src/frameworks/compose/discover/runtime/node.ts
```diff
@@ -0,0 +1,276 @@
...
+        detectedCommands: this.detectedCommands(
+          packageManager,
+          packageJSON.scripts,
+          matchedFramework
+        ),
+      };
+
+      return runtimeSpec;
+    } catch (error: any) {
+      throw new FirebaseError(`Failed to indentify commands for codebase: ${error}`);
```
taeold, line 273: nit*: This could also catch errors like "failed to parse engine"  and the "Failed to identity command" seems less relevant..

src/test/frameworks/compose/discover/runtime/node.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+          },
+        }),
+      });
+      const packageJSON: PackageJSON = {
+        dependencies: {
+          express: "^4.18.2",
+        },
+        devDependencies: {
+          nodemon: "^2.0.12",
+          mocha: "^9.1.1",
```
taeold, line 91: nit* looks like this can be extracted as a common variable to use in both MockFS and here.

src/test/frameworks/compose/discover/runtime/node.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+        run: {
+          cmd: "yarn run start",
+          env: { NODE_ENV: "production" },
+        },
+      };
+
+      expect(actual).to.deep.equal(expected);
+    });
+
+    it("scripts have higher preference over framework commands", () => {
```
taeold, line 198:
```suggestion
    it("should prefer scripts over framework commands", () => {
```

src/test/frameworks/compose/discover/runtime/node.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+        "@algolia/autocomplete-core": "npm:1.7.1",
+        mocha: "^9.1.1",
+      };
+
+      expect(actual).to.deep.equal(expected);
+    });
+  });
+
+  describe("detectedCommands", () => {
+    it("scripts are run using yarn and framework commands prepend npx", () => {
```
taeold, line 163:
```suggestion
    it("should prepend npx to framework commands", () => {
```

src/test/frameworks/compose/discover/runtime/node.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+          cmd: "npx next dev",
+          env: { NODE_ENV: "dev" },
+        },
+        run: {
+          cmd: "yarn run start",
+          env: { NODE_ENV: "production" },
+        },
+      };
+
+      expect(actual).to.deep.equal(expected);
```
taeold, line 237: q: this test looks pretty identical to the previous test case 🤔  what is the difference?

src/test/frameworks/compose/discover/runtime/node.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+    });
+  });
+
+  describe("analyseCodebase", () => {
```
taeold, line 241: let's also test situations where the analyzeCodebase would throw an error.

taeold (changes requested):

src/test/frameworks/compose/discover/runtime/node.spec.ts
svnsairam: the preference is given to `commands present in scripts` over `commands present in framework specs`.

src/frameworks/compose/discover/runtime/node.ts
```diff
@@ -0,0 +1,276 @@
...
+  }
+
+  getRuntimeName(): string {
+    return NODE_RUNTIME_ID;
+  }
+
+  getNodeImage(engine: Record<string, string> | undefined): string {
+    // If no version is mentioned explicitly, assuming application is compatible with latest version.
+    if (!engine) {
+      const latest = VALID_ENGINES.node[VALID_ENGINES.node.length - 1];
```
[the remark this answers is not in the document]
svnsairam: changing to just use support node version = 18.

src/frameworks/compose/discover/runtime/node.ts
svnsairam: Done

src/frameworks/compose/discover/runtime/node.ts
```diff
@@ -0,0 +1,276 @@
...
+    matchedFramework: FrameworkSpec | null
+  ): Command {
+    let runCommand: Command = { cmd: "", env: { NODE_ENV: "production" } };
+    if (scripts?.start) {
+      runCommand.cmd = this.executeScript(packageManager, "start");
+    } else if (matchedFramework && matchedFramework.commands?.run) {
+      runCommand = matchedFramework.commands.run;
+      runCommand = this.executeFrameworkCommand(packageManager, runCommand);
+    }
+
```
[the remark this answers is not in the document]
svnsairam: It can be  either using `node index.js` as fallback command. Or throwing an error saying command not found

src/frameworks/compose/discover/runtime/node.ts
```diff
@@ -0,0 +1,276 @@
...
+import { RuntimeSpec } from "../types";
+import { frameworkMatcher } from "../frameworkMatcher";
+import { LifecycleCommands } from "../types";
+import { Command } from "../types";
+import { join } from "path";
+import { logger } from "../../../../logger";
+import { FirebaseError } from "../../../../error";
+import { VALID_ENGINES } from "../../../constants";
+import { conjoinOptions } from "../../../utils";
+export interface PackageJSON {
```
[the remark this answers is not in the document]
svnsairam: Done

src/frameworks/compose/discover/runtime/node.ts
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    return `npm install --global ${packages.join(" ")}`;
+  }
+
+  installCommand(packageManager: string): string | undefined {
+    if (packageManager === NPM) {
+      return "npm ci";
+    } else if (packageManager === YARN) {
+      return "yarn install";
```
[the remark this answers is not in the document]
svnsairam: If lock files are present we can use `npm ci` for npm 
and `yarn install --frozen-lockfile` for yarn. As they avoid generation a new lock file and fails if there is any update required for dependencies.

And if lock files are not present we can stick with `npm install` and `yarn install`.

src/frameworks/compose/discover/runtime/node.ts
```diff
@@ -0,0 +1,216 @@
...
+    try {
+      let installCmd = "npm install";
+      if (await fs.exists(PACKAGE_LOCK_JSON)) {
+        installCmd = "npm ci";
+      }
+
+      if (packageManager === "yarn") {
+        installCmd = "yarn install";
+        if (await fs.exists(YARN_LOCK)) {
+          installCmd = "yarn install --frozen-lockfile";
```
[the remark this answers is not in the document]
svnsairam: The commands `yarn install --frozen-lockfile` and `npm ci` avoid creating a new lock file. And they install dependencies in the versions mentioned in these lock files. And fail if an update is required. This speeds up installing process.

src/test/frameworks/compose/discover/runtime/node.spec.ts
svnsairam: Moved some common attributed to before() method.

src/frameworks/compose/discover/runtime/node.ts
```diff
@@ -0,0 +1,216 @@
...
+  private readonly runtimeRequiredFiles: string[] = [PACKAGE_JSON];
+  private readonly contentCache: Record<string, boolean> = {};
+
+  // Checks if the codebase is using Node as runtime.
+  async match(fs: FileSystem): Promise<boolean | null> {
+    const areAllFilesPresent = await Promise.all(
+      this.runtimeRequiredFiles.map((file) => fs.exists(file))
+    );
+
+    return Promise.resolve(areAllFilesPresent.every((present) => present));
```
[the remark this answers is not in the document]
svnsairam: Done

src/frameworks/compose/discover/runtime/node.ts
```diff
@@ -0,0 +1,216 @@
...
+
+export interface PackageJSON {
+  dependencies?: Record<string, string>;
+  devDependencies?: Record<string, string>;
+  scripts?: Record<string, string>;
+  engines?: Record<string, string>;
+}
+type PackageManager = "npm" | "yarn";
+
+const supportedNodeVersion = 18;
```
taeold, line 18: nit* let's make this a string[] - we know it's going to be a list soon.

src/frameworks/compose/discover/runtime/node.ts
```diff
@@ -0,0 +1,216 @@
...
+    return areAllFilesPresent.every((present) => present);
+  }
+
+  getRuntimeName(): string {
+    return NODE_RUNTIME_ID;
+  }
+
+  getNodeImage(engine: Record<string, string> | undefined): string {
+    // If no version is mentioned explicitly, assuming application is compatible with latest version.
+    if (!engine) {
```
taeold, line 43: you'd want to also check that `engine.node` isn't undefined.

src/frameworks/compose/discover/runtime/node.ts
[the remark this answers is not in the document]
taeold: per offline discussion: we are going write up a document on how Node.js first.

src/frameworks/compose/discover/runtime/node.ts
svnsairam: Done

src/frameworks/compose/discover/runtime/node.ts
svnsairam: Done

src/frameworks/compose/discover/runtime/node.ts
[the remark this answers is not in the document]
svnsairam: After referring this https://docs.npmjs.com/cli/v8/commands/npm-ci, I thought we should prefer clean install of dependencies as we run the build, install commands in the docker containers.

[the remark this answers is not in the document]
taeold: I agree with your last comment - I'm just pointing out here that if we are going to fall back on `npm install` or `yarn install`, then I'm asking what purpose does `npm ci` or `yarn install --lockfile` serves. I really don't know the answer.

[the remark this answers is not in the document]
svnsairam: I will keep the as `install` for now and document these for later reference.

src/frameworks/compose/discover/runtime/node.ts
```diff
@@ -0,0 +1,276 @@
...
+    matchedFramework: FrameworkSpec | null
+  ): Command {
+    let devCommand: Command = { cmd: "", env: { NODE_ENV: "dev" } };
+    if (scripts?.dev) {
+      devCommand.cmd = this.executeScript(packageManager, "dev");
+    } else if (matchedFramework && matchedFramework.commands?.dev) {
+      devCommand = matchedFramework.commands.dev;
+      devCommand = this.executeFrameworkCommand(packageManager, devCommand);
+    }
+
```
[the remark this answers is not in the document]
taeold: So what are we doing here? Nothing?

src/frameworks/compose/discover/runtime/node.ts
```diff
@@ -0,0 +1,276 @@
...
+    matchedFramework: FrameworkSpec | null
+  ): Command {
+    let buildCommand: Command = { cmd: "" };
+    if (scripts?.build) {
+      buildCommand.cmd = this.executeScript(packageManager, "build");
+    } else if (matchedFramework && matchedFramework.commands?.build) {
+      buildCommand = matchedFramework.commands.build;
+      buildCommand = this.executeFrameworkCommand(packageManager, buildCommand);
+    }
+
```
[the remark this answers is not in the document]
svnsairam: I noticed that `build` command is available from `package.json` only.

src/frameworks/compose/discover/runtime/node.ts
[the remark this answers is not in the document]
svnsairam: Here too same as `build` command.

src/frameworks/compose/discover/types.ts
```diff
@@ -77,4 +79,19 @@ export interface RuntimeSpec {
...
+}
+
+export interface FrameworkHooks {
+  // Programmatic hook with access to filesystem and nodejs API to inspect the workspace.
+  // Primarily intended to gather hints relevant to the build.
+  afterInstall?: (b: AppBundle) => AppBundle;
+
+  // Programmatic hook with access to filesystem and nodejs API to inspect the build artifacts.
+  // Primarily intended to informs what assets should be deployed.
+  afterBuild?: (b: AppBundle) => AppBundle;
```
taeold, line 96: nit* do we need this in this PR?

src/frameworks/compose/discover/runtime/node.ts
```diff
@@ -0,0 +1,212 @@
...
+    return command;
+  }
+
+  getBuildCommand(
+    packageManager: PackageManager,
+    scripts: Record<string, string> | undefined,
+    matchedFramework: FrameworkSpec | null
+  ): Command {
+    let buildCommand: Command = { cmd: "" };
```
taeold, line 129: I think I'd prefer returning `undefined` if the command is empty to denote that we have no build command. Do you think it's better to return an empty string instead?

svnsairam: I think undefined is better than empty string as we can check if command is available or not.

src/frameworks/compose/discover/types.ts
svnsairam: Moving it to different PR.

src/frameworks/compose/discover/runtime/node.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    return buildCommand.cmd === "" ? undefined : buildCommand;
+  }
+
+  getDevCommand(
+    packageManager: PackageManager,
+    scripts: Record<string, string> | undefined,
+    matchedFramework: FrameworkSpec | null
+  ): Command | undefined {
+    let devCommand: Command = { cmd: "", env: { NODE_ENV: "dev" } };
```
taeold, line 145: `NODE_ENV: "dev"` isn't really a common convention. Please use "development" instead (e.g. see https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables)

taeold (approved):

PR #5997, opened 2023-06-16
Refactored ext:install to use the latest extension metadata.

apascal07:
There are a few changes in this PR:

- Removed the ability to list official extensions to choose from. Instead, not passing in an extension ref results in an error that suggests going to Extensions Hub.
- Cleaned up various TODO artifacts from pre-v11.
- Added displaying of new metadata including review status, latest versions, build source URI, Extensions Hub links, deprecation state, installing an outdated version, as well as the contents of the extension like resources and events.
- Added error when trying to install a deprecated extension (this would fail at deploy time).
- When installing an extension, it will now cascade from `latest_approved_version` to `latest_version` to support "offline" use cases.

Example of installing the latest version of a published extension:
Screenshot 2023-06-16 at 8 22 01 AM

Example of installing a deprecated extension version (with no verifiable source):
Screenshot 2023-06-16 at 8 29 29 AM

Example of installing an older unreviewed version of an extension:
Screenshot 2023-06-16 at 8 34 24 AM

Example of installing an unpublished extension:
Screenshot 2023-06-16 at 8 40 01 AM


src/commands/ext-install.ts
```diff
@@ -51,60 +51,69 @@ export const command = new Command("ext:install [extensionName]")
...
+        (!extension.latestApprovedVersion &&
+          extension.latestVersion &&
+          semver.gt(extension.latestVersion, extensionVersion.spec.version))
+      ) {
+        logger.info(
+          `You are about to install extension version ${clc.bold(
+            extensionVersion.spec.version
+          )} which is older than the latest ${
+            extension.latestApprovedVersion ? "accepted version" : "version"
+          } ${clc.bold(extension.latestApprovedVersion!)}.`
```
Rafer45, line 109: What happens if latestApprovedVersion is null here? Should we display `extensions.latestVersion` instead of `latestApprovedVersion`?

apascal07: Good catch, fixed!

i14h:
beautiful ...

src/extensions/displayExtensionInfo.ts
```diff
@@ -54,33 +103,29 @@ export async function displayExtInfo(
...
+export function displayEvents(spec: ExtensionSpec) {
+  const lines = spec.events?.map((event: EventDescriptor) => {
+    return `  - ${clc.magenta(event.type)}${event.description ? `: ${event.description}` : ""}`;
+  });
+  return (
+    clc.bold("Events emitted by this extension:\n") + (lines?.length ? lines.join("\n") : " - None")
+  );
+}
+
+export function displayResources(spec: ExtensionSpec) {
```
ifielker, line 119: This doesn't take TaskQueue resources into account. They are automatically created for "taskQueueTrigger" functions.
This also doesn't take Secrets into account. They are under params with type "secret".
We show both of these types of resources when installing through the console. They should be here too.

joehan:
Looks like there are a few tests that still reference old methods that need to be cleaned up

src/extensions/displayExtensionInfo.ts
apascal07: Good eye! I'll ensure parity with console.

apascal07:
Updated with task queues, secrets, and external services:

Screenshot 2023-06-17 at 2 26 12 AM

@ifielker

src/extensions/displayExtensionInfo.ts
```diff
@@ -18,33 +29,77 @@ const TASKS_ROLE = "cloudtasks.enqueuer";
...
-  extensionName: string,
-  publisher: string,
+ * @param spec the extension spec
+ * @param extensionVersion the extension version
+ * */
+export async function displayExtensionVersionInfo(
   spec: ExtensionSpec,
-  published = false
+  extensionVersion?: ExtensionVersion,
+  latestApprovedVersion?: string,
```
joehan, line 40: Nit: consider an params object here - it seems like it would be easy for callers to accidentally reverse the order on latestApprovedVersion and latestVersion

src/extensions/displayExtensionInfo.ts
```diff
@@ -18,33 +29,77 @@ const TASKS_ROLE = "cloudtasks.enqueuer";
...
+    switch (extensionVersion.listing?.state) {
+      case "APPROVED":
+        reviewStatus = clc.bold(clc.green("Accepted"));
+        break;
+      case "REJECTED":
+        reviewStatus = clc.bold(clc.red("Rejected"));
+        break;
+      default:
+        reviewStatus = clc.bold(clc.yellow("Unreviewed"));
+        break;
```
joehan, line 75: nit: break is unnecessary on the last block of a switch statement, especially since this is a default block (and should always be last)

src/extensions/displayExtensionInfo.ts
```diff
@@ -54,33 +109,59 @@ export async function displayExtInfo(
...
+
+export function displayResources(spec: ExtensionSpec) {
+  const lines = spec.resources.map((resource: Resource) => {
+    let type: string = resource.type;
+    switch (resource.type) {
+      case "firebaseextensions.v1beta.function":
+        type = "Cloud Function V1";
+        break;
+      case "firebaseextensions.v1beta.v2function":
+        type = "Cloud Function V2";
```
joehan, line 140: I think the official product name is Cloud Function (1st gen)/(2nd gen)

ie: https://codelabs.developers.google.com/codelabs/cloud-starting-cloudfunctions-v2#0

joehan (approved):
This looks great! A few small fixes to make, but close enough to LGTM

src/commands/ext-install.ts
```diff
@@ -51,60 +51,72 @@ export const command = new Command("ext:install [extensionName]")
...
         "As of firebase-tools@11.0.0, the `--local` flag is no longer required, as it is the default behavior."
       );
     }
-
+    if (!extensionRef) {
+      throw new FirebaseError(
+        "Extension ref is required to install. To see a full list of available extensions, go to Extensions Hub (https://extensions.dev/extensions)."
+      );
+    }
+    let source: ExtensionSource | undefined;
```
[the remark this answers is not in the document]
apascal07: Hmm, if I don't then `if (!source && !extensionVersion)` complains about using them before initialization. Isn't `MyType | undefined` the expanded version of an optional/nullable in TS?

PR #6000, opened 2023-06-16
Fixes to make web frameworks work for plugin

hsubox76:
1) `WebFrameworks` in `src/frameworks/index.ts` was supposed to be imported from the new `frameworks.ts` file which uses imports (stable, bundleable) instead of runtime requires (causes problems in webpack bundle for VSCode plugin). James told me this was probably a bad rebase and it should have been checked in.

2) A couple of places in the CLI code use `require`, but after being run through webpack, it replaces it with webpack's own require, which works slightly differently (mainly, it expects modules only, not files). There are 2 workarounds for this.
    1) Probably the most robust one is to use `createRequire(import.meta.url)` but unfortunately you can't use `import.meta.url` unless the `target` is changed from `commonjs` to `node16` or `nodenext`, which I don't think the CLI codebase is ready for.
    2) The other workaround is to use a string replacer loader in webpack to replace specific appearances of `require` at compile time with `__non_webpack_require__`. TS hates this so we need a `@ts-ignore`, and we don't want to replace all instances of `require` (the webpack one works fine and probably better on actual modules), just specific ones.

### DIFF
--- a/firebase-vscode/webpack.common.js
+++ b/firebase-vscode/webpack.common.js
@@ -85,6 +85,37 @@ const extensionConfig = {
           ],
         },
       },
+      {
+        test: /frameworks\/utils\.ts$/,
+        loader: "string-replace-loader",
+        options: {
+          multiple: [
+            {
+              search: "require.resolve",
+              replace: "__non_webpack_require__.resolve",
+              strict: true
+            },
+            {
+              search: "require(path",
+              replace: '__non_webpack_require__(path',
+              strict: true,
+            },
+          ],
+        },
+      },
+      {
+        test: /dynamicImport.js$/,
+        loader: "string-replace-loader",
+        options: {
+          multiple: [
+            {
+              search: "require.resolve",
+              replace: "__non_webpack_require__.resolve",
+              strict: true
+            },
+          ],
+        },
+      },
     ],
   },
   plugins: [
--- a/firebase-vscode/webpack.common.js
+++ b/firebase-vscode/webpack.common.js
@@ -85,37 +85,6 @@ const extensionConfig = {
           ],
         },
       },
-      {
-        test: /frameworks\/utils\.ts$/,
-        loader: "string-replace-loader",
-        options: {
-          multiple: [
-            {
-              search: "require.resolve",
-              replace: "__non_webpack_require__.resolve",
-              strict: true
-            },
-            {
-              search: "require(path",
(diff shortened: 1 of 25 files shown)


firebase-vscode/webpack.common.js
```diff
@@ -85,6 +85,37 @@ const extensionConfig = {
...
+        loader: "string-replace-loader",
+        options: {
+          multiple: [
+            {
+              search: "require.resolve",
+              replace: "__non_webpack_require__.resolve",
+              strict: true
+            },
+            {
+              search: "require(path",
```
joehan, line 99: This feels quite fragile - consider at least adding some explicit comments in`frameworks/utils.ts` warning about this string replace

hsubox76:
> I don't love string-replace-loader, as it feels a bit hacky + fragile, but I trust your judgement on whether it is the best choice here.

I think the most robust choice is probably to use [`createRequire(import.meta.url)`](https://nodejs.org/api/module.html#modulecreaterequirefilename) but it won't work with TypeScript until/if the CLI can use `node16` or `nodenext` targets: https://stackoverflow.com/questions/73494747/ts1343-the-import-meta-meta-property-is-only-allowed-when-the-module-opti

firebase-vscode/webpack.common.js
hsubox76: Added a comment.

hsubox76:
> > I don't love string-replace-loader, as it feels a bit hacky + fragile, but I trust your judgement on whether it is the best choice here.
> 
> I think the most robust choice is probably to use [`createRequire(import.meta.url)`](https://nodejs.org/api/module.html#modulecreaterequirefilename) but it won't work with TypeScript until/if the CLI can use `node16` or `nodenext` targets: https://stackoverflow.com/questions/73494747/ts1343-the-import-meta-meta-property-is-only-allowed-when-the-module-opti

Actually I switched to using a conditional, it's not great either but it's less fragile than the string replace. It checks to see if `__webpack_require__` is defined which if true, means it's in webpack, and it means __non_webpack_require__ also exists, and it will sub it in for `require`. Strangely you can't check for `typeof __non_webpack_require__` directly. Solution's described in this comment: https://github.com/webpack/webpack/issues/4175#issuecomment-342931035

Let me know what you think.

yuchenshi (approved):

Issue #5934, opened 2023-06-04
firebase emulators:start crashing in nextjs app.


kevinnammour:
### [REQUIRED] Environment info

**firebase-tools:** 12.3.0

**Platform:** Windows

### [REQUIRED] Steps to reproduce
`npx create-next-app@latest .`

- Typescript: Yes
- Eslint: Yes
- TailwindCSS: No
- src directory: No
- app router: No

`firebase init`

Choose:
- **Hosting**: Configure files for Firebase Hosting and (optionally) set up GitHub Action deploys
- **Emulators**: Set up local emulators for Firebase products

- Detected an existing Next.js codebase in the current directory, should we use this? **Yes**
- In which region would you like to host server-side content, if applicable? **europe-west1 (Belgium)**
- Set up automatic builds and deploys with GitHub? **No**
- Which Firebase emulators do you want to set up? **Hosting emulator**
- Which port do you want to use for the hosting emulator? **5000**
- Would you like to enable the Emulator UI? **Yes**
- Which port do you want to use for the Emulator UI (leave empty to use any available port)? 
- Would you like to download the emulators now? **Yes**

`firebase emulators:start`

Emulator│ Host:Port
Hosting  │ 127.0.0.1:5000 

### [REQUIRED] Expected behavior
The server will stay running once I open 127.0.0.1:5000

### [REQUIRED] Actual behavior
Error [ERR_HTTP_HEADERS_SENT]: Cannot render headers after they are sent to the client
Error: An unexpected error has occurred.
- error uncaughtException: Error [ERR_STREAM_WRITE_AFTER_END]: write after end

And the server crashed.

[debug] [2023-06-04T07:35:29.740Z] Error [ERR_STREAM_WRITE_AFTER_END]: write after end
    at new NodeError (node:internal/errors:399:5)
    at ServerResponse.end (node:_http_outgoing:1010:15)
    at ServerResponse.end (C:\Users\User\Desktop\nextfire\node_modules\next\dist\compiled\compression\index.js:22:783)
    at Responder.handleError (C:\Users\User\AppData\Roaming\npm\node_modules\firebase-tools\node_modules\superstatic\lib\responder.js:96:14)
    at C:\Users\User\AppData\Roaming\npm\node_modules\firebase-tools\node_modules\superstatic\lib\middleware\files.js:142:29
[error] 
[error] Error: An unexpected error has occurred.

kevinnammour:
[...]

Issue #5840, opened 2023-05-15
Hosting rewrites are not working on emulators when using Next.js with webframeworks


nikita-mishchenko:
### [REQUIRED] Environment info

**firebase-tools:** 12.0.0

**Platform:** macOS

### [REQUIRED] Test case

Example `firebase.json`

```
{
    "hosting": {
        ...
        "rewrites": [
            {
                "source": "/api/helloworld",
                "function": {
                    "functionId": "helloworld",
                    "region": "europe-west1"
                }
            },
        ...
        ]
    }
    ...
}
```
Should add rewrite to the hosting emulator.

### [REQUIRED] Steps to reproduce

1) Add function rewrite rule to the hosting block in firebase.json
2) Add some simple firebase function `helloworld`
3) Have some Next.js application
4) Enable webframeworks experiments
5) Run `firebase emulators:start`
6) Call `http://<emulators_host>:<hosting_port>/api/helloworld`

### [REQUIRED] Expected behavior

Receive a response from `helloworld` firebase function

### [REQUIRED] Actual behavior

Receive a 404 error page from Next.js app

### Why is this happening?

This bug is related to pull request [#5582](https://github.com/firebase/firebase-tools/pull/5582). It was released in version `11.25.3`.

Firebase hosting emulator is basically the [superstatic](https://github.com/firebase/superstatic)  server. `firebase-tools` using it to create a dev server with the `superstatic` `strict` stack which is `["redirects", "headers", "files", "rewrites", "missing"]`. Also, `firebase-tools` adding functions before and after `files` part. So the resulting request processing pipeline is the following:

1. Check redirects
2. Check headers
3. Check firebase service urls
4. Check files
5. Check Next.js app response
6. Check rewrites
7. Check missing

In the pull request [#5582](https://github.com/firebase/firebase-tools/pull/5582) `createServerResponseProxy` function was replaced with the `simpleProxy` function. But `createServerResponseProxy` had the following piece of code, that is currently missing:

```
if (proxiedRes.statusCode === 404) {
    next();
}
```

It means that before, if the route is not found in the Next.js app the request pipeline would continue to evaluate `rewrites` and `missing`. And now, it just returns the response from Node.js app and ignores `rewrites` and `missing`.
As the result the whole `rewrites` section under `hosting` in `firebase.json` is ignored.

nikita-mishchenko:
@jamesdaniels Seems like fix is not working for me. Getting the following error message after upgrade to 12.3.0:

```
[...]

PR #6005, opened 2023-06-19
Update proxyResponse method

chalosalvador:
### Description

Fix for #5934 and #5840 

### Scenarios Tested

Followed the steps from #5934  and #5840


src/frameworks/utils.ts
```diff
@@ -63,20 +63,47 @@ export async function warnIfCustomBuildScript(
...
+  proxiedRes.end = new Proxy(proxiedRes.end.bind(proxiedRes), {
+    apply: (target: any, thisArg, args) => {
+      target.call(thisArg, ...args);
+      if (proxiedRes.statusCode === 404) {
+        next();
+      } else {
+        for (const [fn, args] of buffer) {
+          (res as any)[fn](...args);
+        }
+        res.end(...args);
```
alexastrum, line 102: We should probably empty the buffer after playing it back on the `res` object to release memory sooner rather than later.
```
res.end(...args);
buffer.length = 0;
```

src/frameworks/utils.ts
```diff
@@ -63,20 +63,47 @@ export async function warnIfCustomBuildScript(
...
-    if (statusCode === 404) {
-      return next();
-    }
-    const headers = "getHeaders" in response ? response.getHeaders() : response.headers;
-    original.writeHead(statusCode, statusMessage, headers);
-    response.pipe(original);
-  };
+export function proxyResponse(req: IncomingMessage, res: ServerResponse, next: () => void) {
+  const proxiedRes = new ServerResponse(req);
+  const buffer: [string, any[]][] = [];
```
alexastrum, line 68: Very ingenious. I like this pattern a lot.
Should also make troubleshooting easier in the future.

alexastrum (approved):

src/frameworks/utils.ts
```diff
@@ -63,20 +63,47 @@ export async function warnIfCustomBuildScript(
...
+  proxiedRes.write = new Proxy(proxiedRes.write.bind(proxiedRes), {
+    apply: (target: any, thisArg, args) => {
```
leoortizz, line 70: can we type `target` somehow?

src/frameworks/utils.ts
leoortizz, line 68: nit: I think this can be more readable if we use one of the `[]` as `Array<>`
```suggestion
  const buffer: Array<[string, any[]]> = [];
```
side note: can we replace the `any` somehow?

src/frameworks/utils.ts
```diff
@@ -63,20 +63,47 @@ export async function warnIfCustomBuildScript(
...
+    },
+  });
+  proxiedRes.end = new Proxy(proxiedRes.end.bind(proxiedRes), {
+    apply: (target: any, thisArg, args) => {
+      target.call(thisArg, ...args);
+      if (proxiedRes.statusCode === 404) {
+        next();
+      } else {
+        for (const [fn, args] of buffer) {
+          (res as any)[fn](...args);
```
leoortizz, line 100: can we avoid any here somehow?

src/frameworks/utils.ts
```diff
@@ -63,20 +63,47 @@ export async function warnIfCustomBuildScript(
...
-      return;
-    }
-    if (statusCode === 404) {
-      return next();
-    }
-    const headers = "getHeaders" in response ? response.getHeaders() : response.headers;
-    original.writeHead(statusCode, statusMessage, headers);
-    response.pipe(original);
-  };
+export function proxyResponse(req: IncomingMessage, res: ServerResponse, next: () => void) {
```
leoortizz, line 66: I think it would be great if you could add a brief description to what this function is doing now, I'm glad it works but I have no idea why it fixes the errors :joy:

leoortizz (approved):
LGTM! I have no idea why this works though 😂 

Before merging make sure to add changelog

src/frameworks/utils.ts
chalosalvador: I added some comments.

alexastrum (approved):

CHANGELOG.md
```diff
@@ -1 +1,2 @@
 - Run lifecycle hooks for specific codebases. (#6011)
+- Fixes issue that caused firebase emulators:start to crash in Next.js apps (#6005)
```
```suggestion
- Fixed issue causing `firebase emulators:start` to crash in Next.js apps (#6005)
```

amondnet (approved):

PR #6010, opened 2023-06-20
Init flow frameworks cli.

svnsairam:
### Description

CLI for framework stack.

Reference PR: https://github.com/firebase/firebase-tools/pull/5846


src/commands/init.ts
```diff
@@ -71,6 +72,15 @@ const choices = [
     checked: false,
   },
 ];
+
+if (isEnabled("internaltesting")) {
```
taeold, line 76: please create a new experiment (e.g. `frameworks`). `internaltesting` is for another experiment.

src/commands/init.ts
```diff
@@ -71,6 +72,15 @@ const choices = [
...
+  choices.push({
+    value: "frameworkstack",
```
taeold, line 78: let's use the name `frameworks` more consistently. I seem `frameworkstack` and `frameworkstackapi` and would prefer seeing `frameworks`

src/init/features/frameworkstack/constants.ts
```diff
@@ -0,0 +1,15 @@
+export const DEFAULT_REGION = "us-central1";
+export const ALLOWED_REGIONS = [
+  { name: "us-east1 (South California)", value: "us-east1" },
+  { name: "us-east4 (Northern Virginia)", value: "us-east4" },
+  { name: "us-east5 (Columbus)", value: "us-east5" },
+  { name: "us-central1 (Iowa)", value: "us-central1" },
+  { name: "us-central2 (Chicago, IL)", value: "us-central2" },
+  { name: "us-south1 (Dallas)", value: "us-south1" },
+  { name: "us-west1 (Oregon)", value: "us-west1" },
```
taeold, line 9: Where did we get this list?

If we are making this up, then I'd just pick `us-central1` only

src/init/features/frameworkstack/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  setup.frameworkstack = {};
+
+  utils.logBullet("First we need a few details to create your service.");
+
+  await promptOnce(
+    {
+      name: "serviceName",
+      type: "input",
+      default: "acme-inc-web",
+      message: "Create a name for your service [1-64 characters]",
```
taeold, line 37: It looks like we aren't validating this requirement at all. Can we validate it and reprompt if it doesn't meet our requirement?

src/init/index.ts
```diff
@@ -29,6 +29,7 @@ const featureFns = new Map<string, (setup: any, config: any, options?: any) => P
   ["project", features.project], // always runs, sets up .firebaserc
   ["remoteconfig", features.remoteconfig],
   ["hosting:github", features.hostingGithub],
+  ["frameworkstack", features.frameworkstack],
```
taeold, line 32: This chioce has to be inside an expeirment. We don't want this choice to show up for most users.

src/init/features/frameworkstack/constants.ts
svnsairam: This was according to doc shared here: https://www.figma.com/file/soCklCIUh3XVQQZPQm9Olx/Firestack?type=design&node-id=1004-6227&t=q4xvLZwgtAsR9hbn-0

src/init/features/frameworkstack/index.ts
svnsairam: The reason to leave this is, this validation should happen on backend, similar to 
the case where a new `project id` is taken as input (it should be between 6 to 32 characters) and the error is shown when the `Creating Google Cloud Platform project` step is executed. WDYT?

src/commands/init.ts
svnsairam: done

src/commands/init.ts
svnsairam: done

src/init/index.ts
svnsairam: done

src/init/features/frameworks/index.ts
```diff
@@ -0,0 +1,68 @@
...
+import { isEnabled } from "../../../experiments";
+
+/**
+ * Setup new frameworks project.
+ */
+export async function doSetup(setup: any, config: Config, options: Options): Promise<void> {
+  const projectId = setup?.rcfile?.projects?.default;
+  if (projectId && !isEnabled("frameworks")) {
+    await requirePermissions({ ...options, project: projectId });
+    await Promise.all([ensure(projectId, "placeholder.googleapis.com", "unused", true)]);
```
taeold, line 24: nit* we might want to comment this out for now? Does this actually work?

taeold (approved):

src/init/features/frameworks/index.ts
svnsairam: The line to get the `projectId` works, but the we can remove `if` condition for now. The condition was using in other places.

Issue #6002, opened 2023-06-17
Cloud Function lifecycle hooks don't respect codebase filters


sam-throne:
### [REQUIRED] Environment info

**firebase-tools:** 12.3.1

**Platform:** macOS

### [REQUIRED] Test case

```
"functions": [
  {
    "predeploy": "echo predeploy codebaseA",
    "source": "codebaseA",
    "codebase": "codebasea"
  },
  {
    "predeploy": "echo predeploy codebaseA",
    "source": "codebaseB",
    "codebase": "codebaseb"
  },
]
```

### [REQUIRED] Steps to reproduce

```
firebase deploy --only functions:codebasea
```

### [REQUIRED] Expected behavior

Only "predeploy codebaseA" is printed

### [REQUIRED] Actual behavior

predeploy codebaseA AND predeploy codebaseB are printed

I have a potential fix in #5989

aalej:
Hi @sam-throne, thanks for reporting this and for submitting a PR. I was able to reproduce the issue you mentioned. Running `firebase deploy --only functions:<codebase>` runs the predeploy script for both codebases.

I'll notify our engineers about this so that they can take a look, and I’ll be marking this as reproducible. One odd thing I noticed is that your codebases are named “codebaseA” and “codebaseB”. From what I know, codebase names can only contain lowercase letters.

sam-throne:
Thanks @aalej I updated the example to lowercase

blidd-google:
Hey @sam-throne, thanks for reporting this issue. I have implemented a simple fix in a PR out here: https://github.com/firebase/firebase-tools/pull/6011. Note that predeploy hooks won't won't run for individually specified functions, so we will need to revisit this issue and restructure the deploy logic in the future.

sam-throne:
Hi @blidd-google thanks for your time. Your PR appears to be a regression however. Predeploy hooks **did** previously run for individually specified functions, it's just that all of them ran.

IMO if you release with the current behavior you will cause a lot of pain for people who have built workarounds for the current system. I encourage you to take a look at the PR I attached which has much better behavior

blidd-google:
Hey @sam-throne, thanks for the followup and pointing this out — opened a PR against yours  at https://github.com/firebase/firebase-tools/pull/6023 and added some additional fixes to make sure we don't break any pre-existing behavior.

PR #6011, opened 2023-06-20
Run lifecycle hooks for specific codebases

blidd-google:
Fixes https://github.com/firebase/firebase-tools/issues/6002. Further investigation has surfaced another issue where deploying individual functions specified by the `--only` flag will not run any predeploy hooks, but will table this issue for another fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Run lifecycle hooks for specific codebases. (#6011)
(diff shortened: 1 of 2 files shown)


colerogers (approved):

sam-throne:
The issue with hooks not running when deploying individual functions did not exist before! That's a new regression introduced by this PR

PR #6012, opened 2023-06-20
Frameworks

svnsairam:
### Description
Frameworks URL


src/api.ts
```diff
@@ -173,6 +173,11 @@ export const serviceUsageOrigin = utils.envOverride(
   "FIREBASE_SERVICE_USAGE_URL",
   "https://serviceusage.googleapis.com"
 );
+// TODO: Change to correct URL later.
```
alexastrum, line 176: Please assign an issue or a contributor to this TODO.

src/frameworks/compose/discover/frameworks.ts
```diff
@@ -0,0 +1,117 @@
...
+export type BuildOutputOnlyFields = "createTime" | "updateTime" | "sourceRef";
+
+interface BuildSource {
+  codeBaseSource?: CodebaseSource;
+}
+
+interface Status {
+  code: number;
+  message: string;
+  details: any[];
```
alexastrum, line 53: Any chance we can have something more specific than `any` here?

src/api.ts
svnsairam: removing this - internal details.

src/frameworks/compose/discover/frameworks.ts
svnsairam: In the backend its type is declared as `Any` and I wasn't sure what that type looks like as it comes from backend.

src/api.ts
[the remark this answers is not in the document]
svnsairam: Done

src/frameworks/compose/discover/frameworkstackApi.ts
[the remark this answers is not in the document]
svnsairam: Going forward do we use src/api/ for all api calls?

src/api/frameworks.ts
```diff
@@ -0,0 +1,117 @@
...
+  statusDetail: string;
+  cancelRequested: boolean;
+  apiVersion: string;
+}
+
+export interface Operation {
+  name: string;
+  metadata?: OperationMetadata;
+  done: boolean;
+  // oneof result
```
taeold, line 78: q: what do these mean 🤔 ?

taeold (approved):

src/api/frameworks.ts
svnsairam: These are part of Long Running Operation object and they are returned as response from the rpc method, 
name: Service assigned name which is unique. For HttpMapping, the `name` is a resource name ending with `operations/{unique_id}`.
OperationMetadata: Service specific metadata associated with the operation. The is the metadata_type of the Operation.
done: If `false` then Operation is still running, if `true` operation is completed.

PR #6013, opened 2023-06-21
Fixed a bug in the hosting config for emulators. 

christhompsongoogle:
Also fixed an issue where an empty folder creates an issue loading the side panel.

### DIFF
--- a/firebase-vscode/src/workflow.ts
+++ b/firebase-vscode/src/workflow.ts
@@ -145,12 +145,17 @@ export function setupWorkflow(
 ) {
   extensionContext = context;
 
-  // Get user-defined VSCode settings.
-  const workspaceConfig = workspace.getConfiguration(
-    'firebase',
-    vscode.workspace.workspaceFolders[0].uri
-  );
-  const shouldDebug: boolean = workspaceConfig.get('debug');
+  var shouldDebug: boolean;
+  if (vscode.workspace.workspaceFolders) {
+    // Get user-defined VSCode settings.
+    const workspaceConfig = workspace.getConfiguration(
+      'firebase',
+      vscode.workspace.workspaceFolders[0].uri
+    );
+    shouldDebug = workspaceConfig.get('debug');
+  } else {
+    shouldDebug = false;
+  }
 
   /**
    * Logging setup for logging to console and to file.
@@ -158,13 +163,14 @@ export function setupWorkflow(
   // Sets up CLI logger to log to console
   process.env.DEBUG = 'true';
   setupLoggers();
+  
+  // Only log to file if firebase.debug extension setting is true.
+  if (shouldDebug) {
   // Re-implement file logger call from ../../src/bin/firebase.ts to not bring
   // in the entire firebase.ts file
   const rootFolders = getRootFolders();
   const filePath = path.join(rootFolders[0], 'firebase-plugin-debug.log');
   pluginLogger.info('Logging to path', filePath);
-  // Only log to file if firebase.debug extension setting is true.
-  if (shouldDebug) {
     logger.add(
       new transports.File({
         level: "debug",
--- a/firebase-vscode/webviews/EmulatorPanel.tsx
+++ b/firebase-vscode/webviews/EmulatorPanel.tsx
@@ -33,14 +33,20 @@ const DEFAULT_EMULATOR_UI_SELECTIONS: EmulatorUiSelections = {
  */
 export function EmulatorPanel({
   firebaseJson,
+  projectId,
 }: {
   firebaseJson: FirebaseConfig;
+  projectId?: string | undefined;
 }) {
   if (!firebaseJson) {
     throw Error("Expected a valid FirebaseConfig.");
   }
+  var defaultState = DEFAULT_EMULATOR_UI_SELECTIONS;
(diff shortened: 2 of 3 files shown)


firebase-vscode/src/workflow.ts
```diff
@@ -145,26 +145,32 @@ export function setupWorkflow(
...
-    'firebase',
-    vscode.workspace.workspaceFolders[0].uri
-  );
-  const shouldDebug: boolean = workspaceConfig.get('debug');
+  var shouldDebug: boolean;
+  if (vscode.workspace.workspaceFolders) {
+    // Get user-defined VSCode settings.
+    const workspaceConfig = workspace.getConfiguration(
+      'firebase',
+      vscode.workspace.workspaceFolders[0].uri
```
yuchenshi, line 153: nit: shall we display a warning if multiple folders are opened? For example, "the Firebase Extension will use config from the first folder" or something

firebase-vscode/src/workflow.ts
```diff
@@ -145,26 +145,32 @@ export function setupWorkflow(
...
+    );
+    shouldDebug = workspaceConfig.get('debug');
+  } else {
+    shouldDebug = false;
```
yuchenshi, line 157: Please just set the default value on initialization

yuchenshi (approved):

joehan (approved):

firebase-vscode/webviews/EmulatorPanel.tsx
```diff
@@ -33,14 +33,20 @@ const DEFAULT_EMULATOR_UI_SELECTIONS: EmulatorUiSelections = {
  */
 export function EmulatorPanel({
```
hsubox76, line 34: nit: I just noticed that this is at the top level of `webviews/`. Can you put it inside `components/` with the other components?

firebase-vscode/src/workflow.ts
```diff
@@ -145,26 +145,32 @@ export function setupWorkflow(
...
   extensionContext = context;
 
-  // Get user-defined VSCode settings.
-  const workspaceConfig = workspace.getConfiguration(
-    'firebase',
-    vscode.workspace.workspaceFolders[0].uri
-  );
-  const shouldDebug: boolean = workspaceConfig.get('debug');
+  var shouldDebug: boolean;
+  if (vscode.workspace.workspaceFolders) {
```
hsubox76, line 149: As I mentioned in the chat, this may get the user past init without errors but they will run into errors when the extension tries to find the root folder - we should try to figure out if we should allow people to use the extension at all without providing a root folder. I think we should probably prompt them to provide a path or open a folder first

firebase-vscode/webviews/EmulatorPanel.tsx
```diff
@@ -33,14 +33,20 @@ const DEFAULT_EMULATOR_UI_SELECTIONS: EmulatorUiSelections = {
...
   firebaseJson,
+  projectId,
 }: {
   firebaseJson: FirebaseConfig;
+  projectId?: string | undefined;
 }) {
   if (!firebaseJson) {
     throw Error("Expected a valid FirebaseConfig.");
   }
+  var defaultState = DEFAULT_EMULATOR_UI_SELECTIONS;
```
hsubox76, line 44: `let`, or actually probably use `const` and a ternary?

firebase-vscode/webviews/EmulatorPanel.tsx
```diff
@@ -74,7 +80,9 @@ export function EmulatorPanel({
   });
 
   broker.on("notifyEmulatorImportFolder", ({ folder }) => {
-    webLogger.debug(`notifyEmulatorImportFolder received in sidebar: ${folder}`);
+    webLogger.debug(
+      `notifyEmulatorImportFolder received in sidebar: ${folder}`
+    );
     emulatorUiSelections.importStateFolderPath = folder;
```
hsubox76, line 86: Sorry I didn't notice this before but if you're using React state you probably shouldn't be mutating the state variable directly? Same in other places.

firebase-vscode/webviews/EmulatorPanel.tsx
christhompsongoogle: Done.

firebase-vscode/webviews/EmulatorPanel.tsx
christhompsongoogle: In that case - I updated this to create a new object and set that instead. I didn't find any problems with the former but I can see how it might make things act strangely.

firebase-vscode/webviews/EmulatorPanel.tsx
christhompsongoogle: Done

firebase-vscode/src/workflow.ts
christhompsongoogle: This is a little fragile but works in the near-term - the extension loads and lets you select a project. I agree we should have some explicit checking in the future

firebase-vscode/src/workflow.ts
christhompsongoogle: Done.

PR #6015, opened 2023-06-21
Missed a few changes from comments

christhompsongoogle:
Missed a few changes from comments


firebase-vscode/webviews/components/EmulatorPanel.tsx
```diff
@@ -247,3 +267,20 @@ export function EmulatorPanel({
     </PanelSection>
   );
 }
+
+/**
+ * Formats a project ID with a demo prefix if we're in offline mode, or uses the
+ * regular ID if we're hosting.
```
joehan, line 273:
```suggestion
 * regular ID if we're in hosting only mode.
```

joehan (approved):

PR #6023, opened 2023-06-22
Run lifecycle hooks for specific functions

blidd-google:
Unfortunately, previous changes to the lifecycle hook detection logic implemented in https://github.com/firebase/firebase-tools/pull/6011 introduces a potential regression for customers who deploy specific functions, not running any lifecycle hooks when deploying using `--only functions:[specific-function]`. Because in the predeploy stage we don't yet know which functions belong to which codebases, we would like to preserve the behavior of running the lifecycle hooks for all codebases when a specific function is deployed using the `--only` flag. This is an unfortunate consequence of the fact that functions discovery happens _after_ the predeploy stage.

In the future, when we have a chance to refactor, we may be able to address this issue, but for now, the predeploy hooks should run for those codebases specified by `--only`, and all predeploy hooks will run for individually specified functions.

Thanks to @sam-throne for identifying this issue.


src/deploy/lifecycleHooks.ts
```diff
@@ -133,15 +133,46 @@ function getReleventConfigs(target: string, options: Options) {
...
-  return targetConfigs.filter((config: any) => {
-    return !config.target || onlyTargets.includes(config.target);
-  });
+  if (target === "functions") {
+    let onlyConfigs = [];
+    const matched = onlyTargets.reduce(
+      (matched: object, target: string) => ({ ...matched, [target]: false }),
+      {}
+    );
+    targetConfigs.forEach((config: any) => {
```
taeold, line 148: nit* prefer `for` over `forEach`

taeold (approved):

sam-throne:
Thanks for making this change! This will allow me to dramatically speed up my team's build process since we use a monorepo and are currently having to build them all when deploying any function.

The caveat that all hooks will run when deploying individual functions like `firebase deploy --only functions:api-store` is fine, since it can easily be worked around by specifying all codebases (e.g. `firebase deploy --only functions:default:api-store`).

I would also encourage you to document the behavior here: https://firebase.google.com/docs/functions/organize-functions?gen=2nd

sam-throne:
Do you have an ETA on when a release including this PR will be made available on NPM?

blidd-google:
The changes were included in the latest release (12.4.2)!

PR #6042, opened 2023-06-26
Integrate discovery with composer

svnsairam:
### Description

Integrating frameworks discover with composer.


src/frameworks/compose/discover/index.ts
```diff
@@ -1,26 +1,52 @@
...
-      NODE_ENV: "PRODUCTION",
-    },
-    installCommand: "npm install",
-    buildCommand: "npm run build",
-    startCommand: "npm run start",
+const availableRuntimes: Runtime[] = [new NodejsRuntime()];
+
+/**
+ *
+ */
```
taeold, line 13: nit* Looks like you accidentally moved the docstring for discover to `availableRuntimes`

src/frameworks/compose/driver/docker.ts
```diff
@@ -152,16 +153,18 @@ export class DockerDriver implements Driver {
...
     this.buildStage(DOCKER_STAGE_INSTALL, ".");
   }
 
   build(): void {
-    this.dockerfileBuilder
-      .fromLastStage(DOCKER_STAGE_BUILD)
-      .copyForFirebase(".", ".")
-      .run(this.spec.buildCommand);
+    this.dockerfileBuilder.fromLastStage(DOCKER_STAGE_BUILD).copyForFirebase(".", ".");
+    if (this.spec.detectedCommands && this.spec.detectedCommands.build) {
```
taeold, line 165: nit* `this.spec.detectedCommands?.build` for shorter if statement..

src/frameworks/compose/driver/local.ts
```diff
@@ -18,12 +19,20 @@ export class LocalDriver implements Driver {
   }
 
   install(): void {
-    const [cmd, ...args] = this.spec.installCommand.split(" ");
+    let cmd = "";
+    let args: string[] = [];
+    if (this.spec.installCommand) {
```
taeold, line 24: not sure why we run anything if `installCommand` is empty. Shouldn't this be:

```
if (this.spec.installCommand) {
  ...
  this.execCmd(cmd, args)
}
```

src/frameworks/compose/driver/local.ts
```diff
@@ -18,12 +19,20 @@ export class LocalDriver implements Driver {
...
+      [cmd, ...args] = this.spec.installCommand.split(" ");
+    }
     this.execCmd(cmd, args);
   }
 
   build(): void {
-    const [cmd, ...args] = this.spec.buildCommand.split(" ");
+    let cmd = "";
+    let args: string[] = [];
+    if (this.spec.detectedCommands && this.spec.detectedCommands.build) {
```
taeold, line 33: same here - why are we running an empty `cmd`?

src/frameworks/compose/index.ts
```diff
@@ -1,31 +1,36 @@
 import { AppBundle } from "./interfaces";
 import { getDriver, Mode } from "./driver";
 import { discover } from "./discover";
+import { FrameworkSpec, FileSystem } from "./discover/types";
 
 /**
  * Run composer in the specified execution context.
  */
-export function compose(mode: Mode): AppBundle {
+export async function compose(
```
taeold, line 9: please also update `src/commands/internaltesting-frameworks-compose.ts` call to `compose` to await on the result.

src/frameworks/compose/index.ts
```diff
@@ -1,31 +1,36 @@
...
+  fs: FileSystem,
+  allFrameworkSpecs: FrameworkSpec[]
+): Promise<AppBundle> {
   let bundle: AppBundle = { version: "v1alpha" };
-  const spec = discover();
+  const spec = await discover(fs, allFrameworkSpecs);
   const driver = getDriver(mode, spec);
 
-  if (spec.startCommand) {
+  if (spec.detectedCommands && spec.detectedCommands.run) {
```
taeold, line 18: nit* same as `if (spec.detectedCommands?.run)`

(does TS not like this?)

src/frameworks/compose/discover/index.ts
```diff
@@ -1,26 +1,52 @@
...
+  allFrameworkSpecs: FrameworkSpec[]
+): Promise<RuntimeSpec> {
+  try {
+    let discoveredRuntime = undefined;
+    for (const runtime of availableRuntimes) {
+      if (await runtime.match(fs)) {
+        if (!discoveredRuntime) {
+          discoveredRuntime = runtime;
+        } else {
+          throw new FirebaseError("Multiple runtimes discovered for the codebase");
```
taeold, line 25: didn't we try to update the error message to be more helpful 🤔  I forgot if we ever landed on what this should say instead or we decided to punt on it.

src/frameworks/compose/index.ts
svnsairam: Noticed that unit tests are failing due to this.

src/frameworks/compose/index.ts
svnsairam: Previously, `spec` can be `null` if `package.json` is not present. So TS doesn't like `?` symbol on optionals and suggested to use `spec && spec.detectedCommands && spec.detectedCommands.run`. 

Later, I changed so that `spec` won't be null even if `package.json` is not found. It contains other useful parameter like `install` command etc. So, we can use `?` notation.

src/frameworks/compose/discover/index.ts
svnsairam: Adding what runtimes are found may help user? WDYT

src/frameworks/compose/discover/index.ts
svnsairam: Oh yes, may be due `Auto Fix all` suggestion

src/frameworks/compose/driver/local.ts
svnsairam: we can avoid this

src/frameworks/compose/driver/docker.ts
```diff
@@ -152,16 +153,18 @@ export class DockerDriver implements Driver {
       .fromLastStage(DOCKER_STAGE_INSTALL)
       .workdir("/home/firebase/app")
       .envs(this.spec.environmentVariables || {})
-      .copyForFirebase("package.json", ".")
-      .run(this.spec.installCommand);
+      .copyForFirebase("package.json", ".");
+    if (this.spec.installCommand) {
+      this.dockerfileBuilder.run(this.spec.installCommand);
```
taeold, line 158: We won't need to run this step if install command is undefined.

src/frameworks/compose/driver/docker.ts
```diff
@@ -152,16 +153,18 @@ export class DockerDriver implements Driver {
...
     this.buildStage(DOCKER_STAGE_INSTALL, ".");
   }
 
   build(): void {
-    this.dockerfileBuilder
-      .fromLastStage(DOCKER_STAGE_BUILD)
-      .copyForFirebase(".", ".")
-      .run(this.spec.buildCommand);
+    this.dockerfileBuilder.fromLastStage(DOCKER_STAGE_BUILD).copyForFirebase(".", ".");
+    if (this.spec.detectedCommands?.build) {
```
taeold, line 165: same here - don't think we need to run anything if build step doesn't exist no?

src/frameworks/compose/discover/index.ts
```diff
@@ -1,26 +1,53 @@
...
+): Promise<RuntimeSpec> {
+  try {
+    let discoveredRuntime = undefined;
+    for (const runtime of availableRuntimes) {
+      if (await runtime.match(fs)) {
+        if (!discoveredRuntime) {
+          discoveredRuntime = runtime;
+        } else {
+          throw new FirebaseError(
+            `Conflit occured as multiple runtimes ${discoveredRuntime.getRuntimeName()}, ${runtime.getRuntimeName()} are discovered within the codebase.`
```
taeold, line 23:
```suggestion
            `Conflict occurred as multiple runtimes ${discoveredRuntime.getRuntimeName()}, ${runtime.getRuntimeName()} are discovered within the codebase.`
```

src/frameworks/compose/discover/index.ts
```diff
@@ -1,26 +1,53 @@
...
+          );
+        }
+      }
+    }
+
+    if (!discoveredRuntime) {
+      throw new FirebaseError("Unable to identify a runtime for the codebase");
```
taeold, line 30: nit* Include list of supported runtime in the error message.

taeold (approved):
Couple of unaddressed comment - please address. Otherwise lgtm.

PR #6044, opened 2023-06-27
Git to cloud build

svnsairam:
### Description
1. Link Git repo created in cloud build to init flow.
2. With the git repo information make a call to create stack using frameworks API.

interfaces.ts:  contains all the types and interfaces used in frameworks api.
repo.ts: If any rpc call returns long running operations, the methods here converts Operation object to required response objects.
frameworks.ts: Contains the actual calls to backend rpc's.

index.ts: This is the init flow which promts questions to users and calls the framewoks api to createStack.

repo.spec.ts: Tests the method present in repo.ts.


o.txt
taeold, line 1: Accidentally checked in?

src/init/features/composer/repo.ts
```diff
@@ -25,10 +25,6 @@ function extractRepoSlugFromURI(remoteUri: string): string | undefined {
   return match[1];
 }
 
-function generateConnectionId(stackId: string): string {
-  return `composer-${stackId}-conn`;
-}
-
```
taeold, line 31: q: why do we remove this 🤔 . Do you think it's better to re-use stack id as the name of the connection id?

src/api/frameworks/operationsCoverter.ts
```diff
@@ -0,0 +1,30 @@
...
+const frameworksPollerOptions: Omit<poller.OperationPollerOptions, "operationResourceName"> = {
+  apiOrigin: frameworksOrigin,
+  apiVersion: API_VERSION,
+  masterTimeout: 25 * 60 * 1_000,
+  maxBackoff: 10_000,
+};
+
+/**
+ * Creates Stack object from long running operations.
+ */
```
taeold, line 16: nit* Personally, I'd prefer:

1. keeping the HTTP client for Frameworks in `/gcp` like the rest of the http client for GCP apis.

2. Leaving the responsibility of instantiating the poller at the client side. That way, the client can control the poller's configuration like timeout and backoff that's appropriate for the command. (as is done in src/init/features/composer/repo.ts)

Any particular reason you wanted to deviate from the convention and set the structure like this?

taeold (changes requested):

o.txt
svnsairam: Yeah missed to delete in last push.

src/init/features/composer/repo.ts
svnsairam: I felt the user might be happy to see the stackId that they provided in the cloud build than appending something to it. So, I removed this. And if I have to use the connectionID in any other place I need to append `composer-${}-conn` this may lead to changing connectionID in multiple places(if required in future). To keep things simple I felt `stackId` is sufficient.

src/api/frameworks/operationsCoverter.ts
svnsairam: In case of `src/init/features/composer/repo.ts` it uses `gcp` api's. But I think the `frameworksAPI` is different from `/gcp` family. WDYT?

taeold: all firebase apis are part of gcp.

src/init/features/composer/repo.ts
taeold: sgtm. We can easily change this later if we change our mind.

src/frameworks/compose/api/interfaces.ts
taeold, line 1: I'd put this in gcp/frameworks.ts instead - bundling api functions w/ types in the same file seems to be the pattern in `gcp/` files.

src/frameworks/compose/api/repo.ts
```diff
@@ -0,0 +1,30 @@
...
+  projectId: string,
+  location: string,
+  stackInput: Omit<Stack, StackOutputOnlyFields>
+): Promise<Stack> {
+  const op = await gcp.createStack(projectId, location, stackInput);
+  const stack = await poller.pollOperation<Stack>({
+    ...frameworksPollerOptions,
+    pollerName: `create-${projectId}-${location}-${stackInput.name}`,
+    operationResourceName: op.name,
+  });
```
taeold, line 27: I'd still encourage doing this directly where we need it, e.g. in `init/features/frameworks/index.ts`

src/init/features/composer/repo.ts
taeold, line 1: Can we move this under `frameworks?` `composer` is the old location where we were writing the frameworks project files... now that we have aligned on frameworks, let's move it there.

taeold (approved):
nits, overall lgtm.

PR #6045, opened 2023-06-27
Add Remote Config Emulator

kroikie:
### Description

Add Remote Config support to the Emulator Suite. The RC Emulator will allow developers to locally emulate different Remote Config parameter values and see the impact on their app.

Goals:
- Allow developers to develop their apps around different RC parameter values.
- Code to interact with the RC emulator should be the same as those to interact with RC production.

Non goals:
- Emulate client side properties to the emulated RC backend. The RC emulator will serve the parameter values that are selected regardless of the client that is fetching. Eg: An Android client app can fetch parameter values meant for the browser if the browser value is selected.
- Handle template versioning.

The RC Emulator is behind an experiment flag `rcemulator`

### Elements added
- Server that reads local template file to populate initial in-memory template and updates in-memory template based on UI changes or changes to the local template file.
- Respond to fetch requests from client apps
- Handle template updates (PUT) requests.
- Function trigger for template updates.


src/firebaseConfig.ts
```diff
@@ -4,10 +4,11 @@
 // the schema/ directory. After changing this file you will need to run
 // 'npm run generate:json-schema' to regenerate the schema files.
 //
-
 import { RequireAtLeastOne } from "./metaprogramming";
 import type { HttpsOptions } from "firebase-functions/v2/https";
 import { IngressSetting, MemoryOption, VpcEgressSetting } from "firebase-functions/v2/options";
+// Sourced from - https://docs.microsoft.com/en-us/javascript/api/@azure/keyvault-certificates/requireatleastone?view=azure-node-latest
+import { Emulators } from "./emulator/types";
```
joehan, line 11: This line is causing the errors, though i am not entirely clear why. I think importing it is causing the strictNullChecks to in `npm run generate:json-schema to be applied across the entire codebase, which fails/

src/emulator/remoteconfig/server.ts
```diff
@@ -0,0 +1,78 @@
...
+    emulator.loadTemplate();
+    res.json(emulator.template);
+  });
+
+  // Please note you should set the Remote Config minimal fetch interval to 0, so it refreshes every time
+  // Otherwise caching will cause you headaches
+  app.post("/v1/projects/:projectId/namespaces/firebase:fetch", (req, res) => {
+    res.header("etag", `etag-${req.params["projectId"]}-firebase-fetch--${new Date().getTime()}`);
+    res.header(
+      "Access-Control-expose-Headers",
```
joehan, line 26: Not sure if it has any effect, but i think this should be Expose-Headers 
(https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers)

src/emulator/remoteconfig/server.ts
```diff
@@ -0,0 +1,78 @@
...
+      }
+    }
+  });
+
+  return Promise.resolve(app);
+}
+
+function generateClientResponse(template: any): any {
+  const parameters = template["parameters"];
+  const entriesObj: any = {};
```
joehan, line 70: Can this be a real type? Maybe Record<string, boolean>?

src/firebaseConfig.ts
```diff
@@ -4,10 +4,10 @@
 // the schema/ directory. After changing this file you will need to run
 // 'npm run generate:json-schema' to regenerate the schema files.
 //
-
 import { RequireAtLeastOne } from "./metaprogramming";
 import type { HttpsOptions } from "firebase-functions/v2/https";
 import { IngressSetting, MemoryOption, VpcEgressSetting } from "firebase-functions/v2/options";
+// Sourced from - https://docs.microsoft.com/en-us/javascript/api/@azure/keyvault-certificates/requireatleastone?view=azure-node-latest
```
joehan, line 10: Remove this?

src/test/emulators/remoteconfig/templates.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+
+describe("templates", () => {
+  it("can remove !isEmulator parameters", () => {
+    const extractedTemplate = RemoteConfigEmulator.extractEmulator(emulatorTemplate);
+    expect(extractedTemplate).to.deep.equal(prodTemplate);
+  });
+
+  it("can prepare template for emulator", () => {
+    const emulatorReadyTemplate = RemoteConfigEmulator.prepareEmulatorTemplate(prodTemplate);
+    console.log(JSON.stringify(emulatorReadyTemplate, null, 2));
```
joehan, line 15: Clean up console logs

src/firebaseConfig.ts
kroikie: Done!

src/emulator/remoteconfig/server.ts
kroikie: Done!

src/emulator/remoteconfig/server.ts
```diff
@@ -0,0 +1,78 @@
+import * as express from "express";
+import { RemoteConfigEmulator } from "./index";
+import * as cors from "cors";
+
+/**
+ * @param defaultProjectId
```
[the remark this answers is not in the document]
kroikie: Done!

src/emulator/remoteconfig/cloudFunctions.ts
```diff
@@ -0,0 +1,117 @@
...
+      }
+    } catch (e: any) {
+      err = e as Error;
+    }
+
+    if (err || errStatus.length > 0) {
+      this.logger.logLabeled(
+        "WARN",
+        "functions",
+        "Firebase Remote Config function was not triggered due to emulation error. Please file a bug."
```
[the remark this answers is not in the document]
kroikie: Done!

src/emulator/remoteconfig/server.ts
kroikie: In this case the `boolean` would be a [`RemoteConfigParameter`](https://firebase.google.com/docs/reference/remote-config/rest/v1/RemoteConfig#RemoteConfigParameter) which is a complex object. I'm happy to add that type in a follow up PR if you think it would be necessary.

src/test/emulators/remoteconfig/templates.spec.ts
kroikie: Done!

package.json
```diff
@@ -120,6 +121,7 @@
     "inquirer": "^8.2.0",
     "js-yaml": "^3.13.1",
     "jsonwebtoken": "^9.0.0",
+    "JSONStream": "^1.2.1",
```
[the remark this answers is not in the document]
kroikie: I believe this is used for compression on one of RCs put endpoints:

https://github.com/firebase/firebase-tools/pull/6045/commits/6752acaec49703559d04408e53495fefad3b53e0#diff-625686f1018012cd5df45bb4a52e971fe7601bf8a068d6006db5dc8f413f66a6R43

src/emulator/remoteconfig/cloudFunctions.ts
```diff
@@ -0,0 +1,123 @@
...
+      if (cloudEventRes.status !== 200) {
+        errStatus.push(cloudEventRes.status);
+      }
+    } catch (e: any) {
+      err = e as Error;
+    }
+
+    if (err || errStatus.length > 0) {
+      this.logger.logLabeled(
+        "WARN",
```
joehan, line 73: ERROR instead?

src/emulator/remoteconfig/cloudFunctions.ts
```diff
@@ -0,0 +1,123 @@
...
+        "functions",
```
joehan, line 74: Should these be labeled "functions" or "remote config"?

src/emulator/remoteconfig/cloudFunctions.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (!this.enabled) {
+      return;
+    }
+
+    const errStatus: Array<number> = [];
+    let err: Error | undefined;
+    try {
+      /** Legacy Google Events */
+      const eventBody = this.createLegacyEventRequestBody(action, object);
+      const eventRes = await this.client!.post(this.multicastPath, eventBody);
```
joehan, line 51: Rather than using a nonnull assertion, consider just adding an if statement to check at the top of the function:
```
if (!this.client) {
  // throw or log and error as you see appropriate
   this.logger.LogLabelledError("ERROR", "remote config", "Tried to dispatch events but did not have a functions client. This is likely a bug with initializing the remote config emulator - please report")
   return;
}
```

src/emulator/remoteconfig/cloudFunctions.ts
```diff
@@ -0,0 +1,123 @@
...
+      }
+    } catch (e: any) {
+      err = e as Error;
+    }
+
+    if (err || errStatus.length > 0) {
+      this.logger.logLabeled(
+        "WARN",
+        "functions",
+        `Firebase Remote Config function was not triggered due to emulation error. Please file a bug.`
```
joehan, line 75:
```suggestion
        `Firebase Remote Config triggered function was not triggered due to emulation error. Please file a bug.`
```

src/emulator/remoteconfig/cloudFunctions.ts
```diff
@@ -0,0 +1,123 @@
...
+export interface RemoteConfigEventPayload {
+  description: string;
+  updateOrigin: string;
+  updateTime: string;
+  updateType: string;
+  updateUser: any;
+  versionNumber: number;
+}
+
+export class RemoteConfigCloudFunctions {
```
joehan, line 24: Nit: This name could be a bit clearer. IIUC, this class mainly handles event dispatching - maybe RemoteConfigEventDispatcher?

src/emulator/remoteconfig/index.ts
```diff
@@ -0,0 +1,301 @@
+import * as utils from "../../utils";
```
joehan, line 1: Nit: Split internal/external imports.

src/emulator/remoteconfig/index.ts
```diff
@@ -0,0 +1,301 @@
...
+}
+
+export interface ValidTemplate {
+  valid: boolean;
+  msg: string;
+}
+
+export class RemoteConfigEmulator implements EmulatorInstance {
+  private destroyServer?: () => Promise<void>;
+  private _app?: express.Express;
```
joehan, line 36: Nit: This _Xyz style is a bit dated - its mostly used for pure JS implementations to denote something as private by convention (since for a long time, JS did not have any concept of private properties). Since typescript lets us actually declare fields private, theres no need to do this anymore.

src/emulator/remoteconfig/index.ts
```diff
@@ -0,0 +1,301 @@
...
+      // Adding a small delay prevents that at very little cost.
+      await new Promise((res) => setTimeout(res, 5));
+
+      this._logger.logLabeled(
+        "BULLET",
+        "remoteconfig",
+        `Change detected, updating template for Remote Config...`
+      );
+      this.updateTemplateSource(templatePath);
+      this.loadTemplate(this._templateSource);
```
joehan, line 109: Might want to use debounce here (https://github.com/firebase/firebase-tools/blob/master/src/utils.ts#L710) to prevent this from being called excessively when chokidar fires multiple times in quick succession.

We do something very similar when watching for changes to functions files (https://github.com/firebase/firebase-tools/blob/27957b30c5ed67e5fdf05a6dea3240eb2f9f7ef1/src/emulator/functionsEmulator.ts#L447), and it generally improves that feel of the watch experience.

src/emulator/remoteconfig/index.ts
```diff
@@ -0,0 +1,301 @@
...
+   * Update incoming template to be used by the emulator server.
+   * If necessary `!isEmulator` conditional values are added to
+   * each parameter.
+   *
+   * @param template A valid remote config template.
+   */
+  static prepareEmulatorTemplate(template: any): any {
+    const emulatorTemplate = cloneDeep(template);
+    const emulatorParameters = emulatorTemplate["parameters"] || {};
+    for (const parameterName of Object.keys(emulatorParameters)) {
```
joehan, line 136: Can simplify this a lil to `for (const [parameterName, emulatorParameter] of Object.entries(emulatorParameters)) {`

src/emulator/remoteconfig/index.ts
```diff
@@ -0,0 +1,301 @@
...
+   * The object must be a valid Remote Config template and have valid !isEmulator values.
+   *
+   * @param {any} template A RemoteConfigTemplate object to be validated.
+   *
+   * @returns {any} Object indicating the validity of template.
+   * The returned object contains two parameters `valid` and `msg`. `valid` is true if
+   * the template is valid and false otherwise. `msg` describes the reason for the `valid`
+   * value.
+   */
+  static validateRemoteConfigEmulatorTemplate(template: any): ValidTemplate {
```
joehan, line 167: High level feedback: Right now, this seems to just return the first reason a RC template is invalid. I think it would be better to instead try validate the whole thing and return a list of all of the problems.

Instead of just returning every time you find an issue, consider having a array of validationMessages that you continually add to. Then, at the end, you can check the length of that array to see if the template is valid, and either join the error messages together or just return them as an array in ValidTemplate.

joehan, line 167: Not sure if `any` is the ideal type here - if this is user input that we have no clue about the shape of, `unknown` is more appropriate. Otherwise, it would be great to strengthen this type.

src/emulator/remoteconfig/index.ts
```diff
@@ -0,0 +1,301 @@
...
+   * The object must have valid parameters, parameter groups, conditions, and an etag.
+   *
+   * @param {any} template A RemoteConfigTemplate object to be validated.
+   *
+   * @returns {ValidTemplate} indicating the validity of template.
+   * The returned object contains two parameters `valid` and `msg`. `valid` is true if
+   * the template is valid and false otherwise. `msg` describes the reason for the `valid`
+   * value.
+   */
+  static validateRemoteConfigTemplate(template: any): ValidTemplate {
```
joehan, line 228: Same as above - consider letting this return multiple messages when you implement template validation, and consider whether template really needs to be `any`

src/emulator/remoteconfig/index.ts
```diff
@@ -0,0 +1,301 @@
...
+    return {
+      valid: true,
+      msg: "template is valid",
+    };
+  }
+
+  static extractEmulator(emulatorTemplate: any): any {
+    const nonEmulatorTemplate = cloneDeep(emulatorTemplate);
+    const emulatorParameters = nonEmulatorTemplate["parameters"] || {};
+    for (const parameterName of Object.keys(emulatorParameters)) {
```
joehan, line 252: Same as above - prefer Object.entries here if you're gonna use both the key and value.

src/emulator/remoteconfig/server.ts
```diff
@@ -0,0 +1,79 @@
...
+  app.use(cors("*" as any));
+
+  app.put("/revert", (req, res) => {
+    emulator.loadTemplate();
+    res.json(emulator.template);
+  });
+
+  // Please note you should set the Remote Config minimal fetch interval to 0, so it refreshes every time
+  // Otherwise caching will cause you headaches
+  app.post("/v1/projects/:projectId/namespaces/firebase:fetch", (req, res) => {
```
joehan, line 24: This is gonna see `:fetch` as an express param, so this will match `/v1/projects/:projectId/namespaces/firebase:foo `as well.

I _think_ you can escape it like `/v1/projects/:projectId/namespaces/firebase[:]fetch` to prevent this

src/emulator/remoteconfig/server.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+  });
+
+  return Promise.resolve(app);
+}
+
+function generateClientResponse(template: any): any {
+  const parameters = template["parameters"];
+  const entriesObj: any = {};
+  for (const parameterName of Object.keys(parameters)) {
```
joehan, line 72: Prefer Object.entries here since you use the value too

joehan (changes requested):
Took a pass thru, some style feedabck and minor things, but this is looking pretty good. Could you also add some instructions for testing this out to the description?

src/emulator/remoteconfig/cloudFunctions.ts
kroikie: fair, agreed.

src/emulator/remoteconfig/cloudFunctions.ts
```diff
@@ -0,0 +1,123 @@
+import * as uuid from "uuid";
```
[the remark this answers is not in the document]
kroikie: done

src/emulator/remoteconfig/cloudFunctions.ts
kroikie: IMO `functions` since this error would occur when a dev is writing a RC function.

src/emulator/remoteconfig/cloudFunctions.ts
kroikie: done

src/emulator/remoteconfig/server.ts
kroikie: done

src/emulator/remoteconfig/server.ts
kroikie: Done

src/emulator/remoteconfig/index.ts
kroikie: done

saenzramiro:
Any news on this PR? 🤓

I think having an emulator for RC is crucial nowadays.

PR #6050, opened 2023-06-28
Support Sign in with Apple token revocation on the Auth Emulator

### LINKED ISSUE #6028: [Sign in with Apple] Token revocation not supported in Auth Emulator
When deleting an account of a user that has signed in using Sign in with Apple, Apple stipulates that the app developer also revokes the user's access/refresh tokens. This has been implemented in GCIP, and is exposed on the Firebase Apple SDK (JavaScript and Android SDK pending).

We should implement this on the Emulator as well - people are trying to use this, and receive errors when calling the API:
* https://github.com/firebase/firebase-ios-sdk/issues/11469
* https://github.com/firebase/firebase-ios-sdk/issues/9906#issuecomment-1528679012

Internal: backend implementation in cl/517195342

peterfriese:
### Description

This implements basic support for the Firebase Authentication token revocation endpoint that is required to support Sign in with Apple. 

This fixes #6028

### DIFF
--- a/src/emulator/auth/operations.ts
+++ b/src/emulator/auth/operations.ts
@@ -52,6 +52,7 @@ export const authOperations: AuthOps = {
     accounts: {
       createAuthUri,
       delete: deleteAccount,
+      revokeToken,
       lookup,
       resetPassword,
       sendOobCode,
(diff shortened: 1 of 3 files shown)


src/test/apiv2.spec.ts
```diff
@@ -158,7 +158,7 @@ describe("apiv2", () => {
         method: "GET",
         path: "/path/to/foo",
       });
```
christhompsongoogle, line 160: It would be great to cover this with a test to protect against regressions

src/emulator/auth/operations.ts
```diff
@@ -680,6 +681,23 @@ function createSessionCookie(
...
+  reqBody: Schemas["GoogleCloudIdentitytoolkitV2RevokeTokenRequest"]
+  // ctx: ExegesisContext
+): Schemas["GoogleCloudIdentitytoolkitV2RevokeTokenResponse"] {
+  assert(!state.disableAuth, "PROJECT_DISABLED");
+
+  assert(reqBody.idToken, "MISSING_LOCAL_ID");
+  assert(reqBody.providerId, "MISSING_PROVIDER_ID");
+  assert(reqBody.token, "MISSING_TOKEN");
+  assert(reqBody.tokenType, "MISSING_TOKEN_TYPE");
+
```
christhompsongoogle, line 695: I was expecting some kind of token deletion from the ProjectState - how does the revokeToken function emulate the revokeToken behavior in production?

PR #6053, opened 2023-06-28
Migrates functions metrics to GA4 

blidd-google:
Follow-up to https://github.com/firebase/firebase-tools/pull/6016. This PR enables the CLI to send functions metrics data to our new GA4 property. For now, we are intentionally limiting the metrics we track in GA4 to the metrics we are tracking in UA today, but in the next few weeks we will follow up with an expanded set of metrics that will enhance our ability to assess the health of CF3 deployments.

To avoid duplicate reporting to UA and GA4, this PR also stops sending functions metrics to UA.

### DIFF
--- a/src/deploy/functions/release/reporter.ts
+++ b/src/deploy/functions/release/reporter.ts
@@ -1,8 +1,9 @@
 import * as backend from "../backend";
 import * as clc from "colorette";
 
+import * as args from "../args";
 import { logger } from "../../../logger";
-import { track } from "../../../track";
+import { track, trackGA4 } from "../../../track";
 import * as utils from "../../../utils";
 import { getFunctionLabel } from "../functionsDeployHelper";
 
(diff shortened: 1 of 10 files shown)


src/deploy/functions/release/reporter.ts
```diff
@@ -56,30 +57,83 @@ export class AbortedDeploymentError extends DeploymentError {
...
     if (!result.error) {
       totalSuccesses++;
+      if (
+        context?.codebaseDeployEvents &&
+        context.codebaseDeployEvents[result.endpoint.codebase || "default"] !== undefined
+      ) {
+        context.codebaseDeployEvents[result.endpoint.codebase || "default"]
+          .fn_deploy_num_successes++;
+      }
       reports.push(track("function_deploy_success", tag, result.durationMs));
```
joehan, line 100: Could you remove these calls to track? They'll stop collecting data in a few days, and  getting rid of them now make it a little easier to merge old & new metrics in BQ

src/deploy/functions/release/reporter.ts
```diff
@@ -56,30 +57,83 @@ export class AbortedDeploymentError extends DeploymentError {
...
     } else if (result.error instanceof AbortedDeploymentError) {
       totalAborts++;
+      if (
+        context?.codebaseDeployEvents &&
+        context.codebaseDeployEvents[result.endpoint.codebase || "default"] !== undefined
```
joehan, line 105: Nit: I think this conditional can be reduced down to:
```
        context?.codebaseDeployEvents.?[result.endpoint.codebase || "default"] !== undefined
```

joehan (approved):
A few small things to change, but this LGTM - thanks for the rapid turnaround!

PR #6054, opened 2023-06-28
Normalized extension root path before usage in ext:dev:upload.

apascal07:


src/extensions/extensionsHelper.ts
```diff
@@ -817,11 +810,13 @@ export async function uploadExtensionVersionFromGitHubSource(args: {
   if (!extensionRoot) {
     const defaultRoot = "/";
     if (!args.nonInteractive) {
-      extensionRoot = await promptForValidExtensionRoot(defaultRoot);
+      extensionRoot = await promptForExtensionRoot(defaultRoot);
     } else {
       extensionRoot = defaultRoot;
     }
   }
+  const normalizedRoot = path.join(extensionRoot).replaceAll(/^[.\/]*|[\/]*$/g, "");
```
joehan, line 818: Consider if path.normalize works here here instead: https://www.w3schools.com/nodejs/met_path_normalize.asp

apascal07: Good suggestion, thanks!

joehan (approved):

PR #6056, opened 2023-06-28
VSCode plugin: Workarounds for functions code + Monospace settings

hsubox76:
### 2 workarounds for some functions code issues:
1. Some functions files call functions as members of `module.exports` or `exports` as a hack to make sure those functions can be stubbed when called in a test by other functions in the same file. Example: https://github.com/firebase/firebase-tools/blob/master/src/deploy/functions/release/planner.ts#L299 This doesn't work in the plugin's current TS/webpack setup - it would need to switch tsconfig to "module: commonjs" which creates a larger number of problems. Fixed with a search/replace for now - long term solution, maybe refactor the functions files so that the functions that need to be stubbed are moved to separate files.
2. The trigger parsing flow relies on 2 uncompiled JS files existing in the same dir. Again the long term fix might be complex, so for now I'm using WebpackCopyPlugin to make sure those 2 files get copied to the plugin's dist dir at compile time.

### Some changes to default settings:
Monospace wants logs on by default, and wants them to default to the path /tmp/firebase-plugin.log. I'm setting this to only be the default for the Monospace build for now by using the package json swapping script. For the desktop build, we need to figure out a "default path". Windows doesn't have a /tmp path. I can use os.tmpdir() and I'm not sure what it does on windows, but on Mac it goes to a new directory with a randomly generated name in /var. On linux it seems to go to /tmp. The Mac case seems unacceptable and the windows case needs testing. So for now we'll only put defaults on for Monospace.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,0 @@
-- Increased extension instance create poll timeout to 1h to match backend (#5969).
-- Refactored `ext:install` to use the latest extension metadata. (#5997)
-- Fixed issue where missing trigger warnings would be wrongly displayed when emulating extensions with HTTPS triggers. (#6055)
-- Normalized extension root path before usage in `ext:dev:upload`. (#6054)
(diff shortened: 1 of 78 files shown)


joehan (approved):
LGTM - number 1 seems like something that we could fix relatively easily with some minor refactoring, number 2 seems quite a bit more involved unfortunately, but i'll defer to CF3 folks there.

prakhar1989 (approved):

PR #6063, opened 2023-06-29
Make firebase:database:list to always use the RTDB management API

fredzqm:
### Description

`database:instances:list` shall always use the public one platform API. The firedata API should be turned down soon.

The `rtdbmanagement` experiment keeps the output format backward compatible.
When `rtdbmanagement` is enabled, use the table output.
When `rtdbmanagement` is disabled, return databases name in a plain text.

`--json` works in both case.

Internal context: http://b/289243438, [go/rtdb-instances-list-cli](https://goto.google.com/rtdb-instances-list-cli)

### Sample Commands

```
> FIREBASE_CLI_PREVIEWS=rtdbmanagement firebase database:instances:list
> firebase database:instances:list
> firebase database:instances:list --json
> firebase database:instances:list --json --location us-central1
```


src/commands/database-instances-list.ts
```diff
@@ -18,71 +17,48 @@ import {
...
+      for (const instance of instances) {
+        logger.info(instance.name);
+      }
+      logger.info(`Project ${options.project} has ${instances.length} database instances`);
+      return instances;
     }
-    logger.info(`Project ${options.project} has ${instances.length} database instances`);
+    const tableHead = ["Database Instance Name", "Location", "Type", "State"];
+    const table = new Table({ head: tableHead, style: { head: ["green"] } });
+    instances.forEach((db) => {
```
joehan, line 58: Nit: Prefer `for.. of...` over `forEach`

joehan (approved):

fredzqm:
Should we put anything in the release note? It changes the impl of the cmd, no behavior change intended.

joehan:
Probably worth adding a note to the changelog, since users could see out of date data on old versions - something like 'Fixed an issue where `database:list` would have inaccurate results'

PR #6066, opened 2023-06-29
Frameworks: Call "next build" from command line instead of import

hsubox76:
During the Next.js frameworks deploy, there is a step where web frameworks has to call `next build`. But instead of calling it on the command line, it imports the code from the Next.js node_modules dir and runs it. If run in an environment like the VSCode plugin, which does not shut down after a successful run, this causes there to be some kind of in-memory state that persists on a second call to `next build`, which Next is not expecting, and breaks it. Since this Next command was designed to run once from the command line and shut down, and run fresh the next time it is run, I don't think this use case is supported.

I've changed it to use `spawn` to call NextJS from the provided executable in `node_modules/.bin`, similar to how it's done in the web frameworks config for Astro: https://github.com/firebase/firebase-tools/blob/master/src/frameworks/astro/index.ts#L45

I'm using spawn instead of spawnSync because it takes a while and want to make sure the user is getting output during the Nextjs build (at least in the debug logs).

Sorry about the package.json changes, trying to come up with a solution where the generated package.json for VSCE packaging goes into a .gitignored dist/ folder or something instead of changing the original package.json.


prakhar1989 (approved):

leoortizz (approved):
LGTM

leoortizz:
@alexastrum FYI

PR #6068, opened 2023-06-30
genSignedUrl POST endpoint for firebase-storage emulator

abegizaw3:

### DIFF
--- a/src/emulator/storage/apis/firebase.ts
+++ b/src/emulator/storage/apis/firebase.ts
@@ -16,6 +16,7 @@ import {
 } from "../upload";
 import { reqBodyToBuffer } from "../../shared/request";
 import { ListObjectsResponse } from "../files";
+import { time } from "node:console";
 import { createHmac } from "node:crypto";
 import { privateKey } from "../constants";
 /**
--- a/src/emulator/storage/apis/firebase.ts
+++ b/src/emulator/storage/apis/firebase.ts
@@ -16,7 +16,6 @@ import {
 } from "../upload";
 import { reqBodyToBuffer } from "../../shared/request";
 import { ListObjectsResponse } from "../files";
-import { time } from "node:console";
 import { createHmac } from "node:crypto";
 import { privateKey } from "../constants";
 /**
--- a/src/emulator/storage/apis/firebase.ts
+++ b/src/emulator/storage/apis/firebase.ts
@@ -130,10 +130,6 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
     return res.json(new OutgoingFirebaseMetadata(metadata));
   });
 
-  /**
-   * Is object ID encoded? And where
-   * @date 6/27/2023 - 4:02:12 PM
-   */
   firebaseStorageAPI.post(`/b/:bucketId/o/:objectId[(:)]generateSignedUrl`, async (req, res) => {
     const timeToLive = req.body.ttl;
     const msInAWeek = 7 * 24 * 60 * 60 * 1000;
--- a/src/emulator/storage/apis/firebase.ts
+++ b/src/emulator/storage/apis/firebase.ts
@@ -16,7 +16,9 @@ import {
 } from "../upload";
 import { reqBodyToBuffer } from "../../shared/request";
 import { ListObjectsResponse } from "../files";
-
+import { time } from "node:console";
+import { createHmac } from "node:crypto";
+import { privateKey } from "../constants";
 /**
  * @param emulator
  */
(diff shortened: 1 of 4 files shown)


src/emulator/storage/apis/firebase.ts
```diff
@@ -128,6 +130,48 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
+        },
+      });
+    }
+    const currentDate = new Date()
+      .toISOString()
+      .replaceAll("-", "")
+      .replaceAll(":", "")
+      .replaceAll(".", "");
+
+    const unsignedUrl = `${req.protocol}://${req.hostname}:${req.socket.localPort}/v0/b/${req.params.bucketId}/o/${req.params.objectId}?alt=media&X-Firebase-Date=${currentDate}&X-Firebase-Expires=${timeToLive}`;
```
jmwski, line 151: Is `${req.params.objectId}` url encoded?

src/emulator/storage/apis/firebase.ts
```diff
@@ -128,6 +130,48 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
+          message: `Unauthorized User`,
+        },
+      });
+    }
+
+    const signature = createHmac("sha256", privateKey).update(unsignedUrl).digest("base64");
+
+    const signedUrl = `${unsignedUrl}&X-Firebase-Signature=${signature}`;
+
+    return res.status(200).json(signedUrl);
```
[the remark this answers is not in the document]
jmwski: (looking at http://go/signed-url-proposal) code sample

src/emulator/storage/apis/firebase.ts
```diff
@@ -128,6 +130,48 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
     return res.json(new OutgoingFirebaseMetadata(metadata));
   });
 
+  firebaseStorageAPI.post(`/b/:bucketId/o/:objectId[(:)]generateSignedUrl`, async (req, res) => {
+    const timeToLive = req.body.ttl;
```
jmwski, line 134: Looks like it's going to be `req.body.ttlInMillis`, per [go/signedurl-js-proposal](http://goto.google.com/signedurl-js-proposal).

src/emulator/storage/apis/firebase.ts
```diff
@@ -128,6 +130,48 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
+    }
+    const currentDate = new Date()
+      .toISOString()
+      .replaceAll("-", "")
+      .replaceAll(":", "")
+      .replaceAll(".", "");
+
+    const unsignedUrl = `${req.protocol}://${req.hostname}:${req.socket.localPort}/v0/b/${req.params.bucketId}/o/${req.params.objectId}?alt=media&X-Firebase-Date=${currentDate}&X-Firebase-Expires=${timeToLive}`;
+
+    const authorized = await storageLayer.authenticateUser({
```
jmwski, line 153: Could you move the security rules check to the top of the route. If they fail, we should exit with 403 immediately.

src/emulator/storage/apis/firebase.ts
jmwski, line 151: I think we should split out express js details from url construction. 

Maybe create an object like `CreateSignedUrlRequest` similar to https://github.com/firebase/firebase-tools/blob/57dd7fef80d9b9294da41bb93cf25a1108674459/src/emulator/storage/files.ts#L90 and supply all the properties needed to generate signed urls. We could move this to a new file and add a helper:

```
func generateSignedUrl(req: CreateSignedUrlRequest): CreateSignedUrlResponse {
  ...
}
```

All input validation / parsing can remain here, but I'd move the `(timeToLive < 0 || timeToLive > msInAWeek)` inside this new helper. The helper should throw when out of range ttls are passed.

We can hardcode `https` instead of using `req.protocol`.

Instead of passing `hostname` and `port` separately, we can pass them as one parameter called `baseUrl` `${req.hostname}:${req.socket.localPort}`.

jmwski: I would also split out all the details related to computing the signature into another helper called by `generateSignedUrl(...)` (it can live in the same file). Specifically, this logic:

createHmac("sha256", privateKey).update(unsignedUrl).digest("base64");

src/emulator/storage/apis/firebase.ts
jmwski: naming nit: `authenticate` is different than `authorize`: https://auth0.com/docs/get-started/identity-fundamentals/authentication-and-authorization

src/emulator/storage/constants.ts
```diff
@@ -0,0 +1,3 @@
+import { randomBytes } from "node:crypto";
+
+export const privateKey = randomBytes(32).toString("hex");
```
jmwski, line 3: nit: Call this something more descriptive: `SIGNED_URL_PRIVATE_KEY`. 

I think it might make sense for us to actually hardcode a key (call `randomBytes(32).toString("hex")` in your terminal and copy/paste the output). If we generate a new one each time, signed urls will become invalid if a dev restarts the emulator (probably not uncommon).

src/emulator/storage/files.ts
```diff
@@ -40,6 +40,13 @@ export class StoredFile {
   }
 }
 
+/**  Parsed request object for {@link StorageLayer#authenticateUser}. */
+export type CreateObjectRequest = {
```
jmwski, line 44: I'd this, `generateSignedUrl` is not creating an object.

src/emulator/storage/apis/firebase.ts
```diff
@@ -128,6 +130,48 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
     return res.json(new OutgoingFirebaseMetadata(metadata));
   });
 
+  firebaseStorageAPI.post(`/b/:bucketId/o/:objectId[(:)]generateSignedUrl`, async (req, res) => {
```
jmwski, line 133: Right now I think this will create signed urls for objects that don't exist. We should return a 404 in that case.

src/emulator/storage/apis/firebase.ts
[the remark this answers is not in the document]
abegizaw3: I accidentally overlooked this part

src/emulator/storage/apis/firebase.ts
jmwski: Discussed offline, rulesValidator is a private member of storageLayer, so the new `generateSignedUrl` should call it.

src/emulator/storage/apis/firebase.ts
abegizaw3: yes

jmwski:
Just wanted to confirm, are the comments that were resolved in this PR going to be addressed in https://github.com/firebase/firebase-tools/tree/abeG-wip-getRequest?

abegizaw3:
> Just wanted to confirm, are the comments that were resolved in this PR going to be addressed in https://github.com/firebase/firebase-tools/tree/abeG-wip-getRequest?

yes, they have all been fixed in a working draft. Is it not updating?

PR #6078, opened 2023-07-05
VSCode plugin: Handle service accounts better

hsubox76:
The plugin's handling of service accounts was very fragile - essentially it was assuming that if calling the CLI's getAllAccounts() and other account-checking methods returned nothing, but requireAuth() succeeded without error, that there must be a service account (that it had no information about). It also wasn't equipped to properly handle switching between a service account and a logged-in account.

This PR modifies `requireAuth()` and `autoAuth()` to return the email of the service account. It seems `autoAuth()` only really is called when a service account is involved, and it writes the service account token to an in-memory variable in apiv2.ts. When the CLI makes requests, this in-memory variable takes precedence over getting a token from a standard logged-in Google OAuth2 account. So we have to make sure to wipe it if the plugin user switches accounts from a service account to the OAuth2 account.

Also changed the plugin code to pass around the user instead of just the email, so that various bits of code can check the `type` property to see if it's a service account. (It used to check the email when all service accounts email fields were just "service_account", but now that we store the actual service account email, it has to check a second field.)

We need the service account email for debugging, and so that users can correctly grant any required permissions to it (especially important for Monospace).


firebase-vscode/webviews/components/AccountSection.tsx
```diff
@@ -14,20 +14,25 @@ import { ServiceAccountUser } from "../../common/types";
...
+  user,
   allUsers,
   isMonospace,
 }: {
-  userEmail: string | null;
+  user: UserWithType | ServiceAccountUser | null;
   allUsers: Array<User | ServiceAccountUser> | null;
   isMonospace: boolean;
 }) {
+  console.log(user, allUsers);
```
joehan, line 30: Remove

joehan (approved):

firebase-vscode/webviews/components/AccountSection.tsx
hsubox76: Done.

Issue #6074, opened 2023-07-05
functions:secrets:set does not destroy the stale version of the secret


kitfit-dave:
### Environment info

**firebase-tools:** 12.4.2

**Platform:** macOS 13.4

### Test case

Updating a secret that was previously created and is used in one or more functions claims to "destroy the stale version" of the secret after the re-deploy, but does not do so.

### Steps to reproduce

- `firebase functions:secrets:set MY_KEY`
- answer Y to `Do you want to re-deploy the functions and destroy the stale version of secret MY_KEY`

### Expected behavior

new MY_KEY version to be added
previous MY_KEY version to be in "Destroyed" state

### Actual behavior

new MY_KEY version to be added
previous MY_KEY version still in "Enabled" state

Note: subsequent `firebase functions:secrets:prune` does identify and destroy the old version

aalej:
Hi @kitfit-dave, thanks for bringing this to our attention. I was able to replicate this. Let me check what we can do for this issue or bring someone here who can provide more context about this behavior.

taeold:
Looks like this bug is left over from https://github.com/firebase/firebase-tools/issues/4459 where we pulled out the feature to remove secrets on a successful `set` operation due to a nasty bug that removed almost all secrets IN USE 😱. Working on a bug-free, safer version now.

PR #6080, opened 2023-07-05
Prune old versions when creating new secret version using `secrets:set` command

taeold:
Fixes https://github.com/firebase/firebase-tools/issues/6074.

To make it easy to identify secret versions associated with existing functions, I made small refactoring changes on couple of functions in `src/gcp/cloudfunctionsv2.ts` as part of this PR.


src/functions/secrets.ts
```diff
@@ -21,12 +23,25 @@ import { logWarning } from "../utils";
 import { promptOnce } from "../prompt";
 import { validateKey } from "./env";
 import { logger } from "../logger";
-import { functionsOrigin } from "../api";
 import { assertExhaustive } from "../functional";
 import { needProjectId } from "../projectUtils";
 
 const FIREBASE_MANAGED = "firebase-managed";
 
+const gcfV1PollerOptions: Omit<poller.OperationPollerOptions, "operationResourceName"> = {
```
blidd-google, line 31: *nit: we have the same poller options specified in `fabricator.ts`, should we keep them all in the same place?

blidd-google (approved):
LGTM

src/functions/secrets.ts
taeold: For some crazy reason... I have to make a copy here other wise some of the value in the options are empty. I wish I knew what was going on... but not going to bother for now. Adding a comment.

svnsairam (approved):

Issue #6089, opened 2023-07-06
Unable to run v2 functions on firebase functions:shell 


ilonawong:
### [REQUIRED] Environment info

**firebase-tools:** 12.4.2

**Platform:** MacOS

### [REQUIRED] Test case
Any v2 function (specifically, `onDocumentWritten`)

### [REQUIRED] Steps to reproduce
run firebase functions:shell, try and call the v2 function with any combination of params

```
v2_function_test({ 
      data: { before, after }
})

v2_function_test({ before, after })
``` 

and this error pops up:
```
Error: Cannot determine payload type, datacontenttype is undefined, failing out

functions: Error: Error: Cannot parse event payload.
    at createBeforeSnapshot (/Users/iwong/repos/partiful/node_modules/firebase-functions/lib/v2/providers/firestore.js:140:15)
    at makeChangedFirestoreEvent (/Users/iwong/repos/partiful/node_modules/firebase-functions/lib/v2/providers/firestore.js:171:37)
    at func (/Users/iwong/repos/partiful/node_modules/firebase-functions/lib/v2/providers/firestore.js:237:32)

```

If I try to add datacontenttype as a param, it just gets nested inside a "data" object.

### [REQUIRED] Expected behavior
Be able to run v2 functions locally

### [REQUIRED] Actual behavior
It doesn't, error above

taeold:
Hey @ilonawong - thank you for filing this bug report. Confirmed that this is an issue for functions shell - we'll work on getting something out soon to support 2nd Gen functions in `functions:shell` command.

ilonawong:
Thanks @taeold! Any approximate timeline? We converted a bunch of our functions over and we're limited to reverting to v1 to run yarn shell right now haha.

taeold:
@ilonawong Going for mid next week 🙏🏼

PR #6092, opened 2023-07-07
Rewrite `src/localFunction.js` in TypeScript.

taeold:
`src/localFunction.js` is used to power the `functions:shell` command. We need to make substantial changes to it to properly support 2nd Gen functions (https://github.com/firebase/firebase-tools/issues/6089). Rewriting the source in modern TypeScript as a prep.


src/functionsShellCommandAction.ts
```diff
@@ -99,7 +98,7 @@ export const actionFunction = async (options: Options) => {
           if (emulator.emulatedFunctions.includes(trigger.id)) {
             const localFunction = new LocalFunction(trigger, emulator.urls, emulator);
             const triggerNameDotNotation = trigger.name.replace(/-/g, ".");
-            _.set(context, triggerNameDotNotation, localFunction.call);
+            context[triggerNameDotNotation] = localFunction.makeFn();
```
joehan, line 101: Can you double check that this still works as expected for functions with -'s in the name? `_.set` will treat `triggerNameDotNotation` as a path (https://www.geeksforgeeks.org/lodash-_-set-method/)  but `context[triggerNameDotNotation` will set it as a single top level attribute.

src/localFunction.ts
```diff
@@ -0,0 +1,225 @@
...
+import { AuthMode } from "./emulator/events/types";
+
+/**
+ * @class
+ * @this LocalFunction
+ * @param {object} trigger
+ * @param {object=} urls
+ * @param {object=} controller
+ */
+export default class LocalFunctionClass {
```
joehan, line 16: Nit: Probably cleaner to just name this LocalFunction

src/localFunction.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  constructor(
+    private trigger: EmulatedTriggerDefinition,
+    urls: Record<string, string>,
+    private controller: FunctionsEmulatorShell
+  ) {
+    this.url = urls[trigger.id];
+  }
+
+  private substituteParams(resource: string, params: Record<string, string>) {
```
joehan, line 27:
```suggestion
  private substituteParams(resource: string, params: Record<string, string>): string {
```

src/localFunction.ts
```diff
@@ -0,0 +1,225 @@
...
+    const wildcardRegex = new RegExp("{[^/{}]*}", "g");
```
joehan, line 28: Nit: consider moving this out of the function so that we don't have to compile it on every call

src/localFunction.ts
```diff
@@ -0,0 +1,225 @@
...
+    return resource.replace(wildcardRegex, (wildcard: string) => {
+      const wildcardNoBraces = wildcard.slice(1, -1); // .slice removes '{' and '}' from wildcard
+      const sub = params?.[wildcardNoBraces];
+      return sub || wildcardNoBraces + utils.randomInt(1, 9);
+    });
+  }
+  private constructCallableFunc(data: any, opts: any) {
```
joehan, line 35: How tricky would it be to have stronger types here? I think data could be `string | object | undefined`, and opts can be `opts?: { instanceIdToken?: string } = {}` (or you can make a shared type that covers all the different `opts` in this file

src/localFunction.ts
```diff
@@ -0,0 +1,225 @@
...
+          uid: auth.uid ?? "",
+          token: auth.token || {},
+        },
+      };
+    }
+    // Default to admin
+    return { admin: true };
+  }
+
+  makeFirestoreValue(input?: any) {
```
joehan, line 99: Should input be unknown here, since I believe its an arbitrary user provided value?

src/localFunction.ts
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  private isDatabaseFn(eventTrigger: Required<EmulatedTriggerDefinition>["eventTrigger"]) {
+    return utils.getFunctionsEventProvider(eventTrigger.eventType) === "Database";
+  }
+  private isFirestoreFunc(eventTrigger: Required<EmulatedTriggerDefinition>["eventTrigger"]) {
+    return utils.getFunctionsEventProvider(eventTrigger.eventType) === "Firestore";
+  }
+
+  private triggerEvent(data: any, opts: any) {
```
joehan, line 145: Same thought as above - can we improve this typing?

joehan:
Love to see it!

src/functionsShellCommandAction.ts
taeold: great catch!!

taeold: Reverting back to using lodash :(

src/localFunction.ts
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  private substituteParams(resource: string, params?: Record<string, string>): string {
+    if (!params) {
+      return resource;
+    }
+    return resource.replace(this.paramWildcardRegex, (wildcard: string) => {
+      const wildcardNoBraces = wildcard.slice(1, -1); // .slice removes '{' and '}' from wildcard
+      const sub = params?.[wildcardNoBraces];
+      return sub || wildcardNoBraces + utils.randomInt(1, 9);
```
joehan, line 43:
```suggestion
      return sub || `${wildcardNoBraces}${utils.randomInt(1, 9)}`;
```

src/localFunction.ts
```diff
@@ -0,0 +1,240 @@
...
+          uid: auth.uid ?? "",
+          token: auth.token || {},
+        },
+      };
+    }
+    // Default to admin
+    return { admin: true };
+  }
+
+  makeFirestoreValue(input?: unknown) {
```
joehan, line 110: nit: return type

src/localFunction.ts
```diff
@@ -0,0 +1,240 @@
...
+    }
+    const currentTime = new Date().toISOString();
+    return {
+      fields: encodeFirestoreValue(input),
+      createTime: currentTime,
+      updateTime: currentTime,
+    };
+  }
+
+  private requestCallBack(err: unknown, response: request.Response, body: string | object) {
```
joehan, line 130:
```suggestion
  private requestCallBack(err: unknown, response: request.Response, body: string | object): void {
```

joehan (approved):
LGTM - as a nit, consider taking a pass through and adding some return types on the methods of LocalFunction

PR #6096, opened 2023-07-07
Support 2nd Gen functions in `functions:shell` command

taeold:
Fixes https://github.com/firebase/firebase-tools/issues/6089


src/utils.ts
```diff
@@ -341,8 +341,8 @@ export function getFunctionsEventProvider(eventType: string): string {
     const provider = last(parts[1].split("."));
     return _.capitalize(provider);
   }
-  // New event types:
-  if (/google.pubsub/.exec(eventType)) {
+  // v1 event types:
```
jhuleatt, line 344:
```suggestion
  // 1st gen event types:
```

^nit

blidd-google (approved):

PR #6130, opened 2023-07-13
Better message for api enablement failures

hsubox76:
`ensure` is called in a number of places to ensure the user has API permissions on cloud APIs that are about to be used. If permission is denied with a 403, in most cases it spits out the error verbatim as received from the endpoint which looks something like this: "HTTP Error: 403, Permission denied to enable service [artifactregistry.googleapis.com]" - where whatever service it is will be in place of "artifactregistry.google.com".

Functions try/catches `ensure` and wraps the error in a nice user-friendly message that gives the user the GCP console url to enable the API: https://github.com/firebase/firebase-tools/blob/master/src/deploy/functions/ensure.ts#L51 but every other usage of `ensure` seems to just spit out the less helpful error above.

This change adds a user-friendly message similar to the one Functions currently throws, for all services. It just modifies the original's error message and re-throws it, so that Functions won't be affected (it will pick up the status and continue to throw its bespoke error message instead of this new one).

### DIFF
--- a/src/ensureApiEnabled.ts
+++ b/src/ensureApiEnabled.ts
@@ -40,6 +40,12 @@ export async function check(
   return isEnabled;
 }
 
+const apiPermissionDeniedRegex = new RegExp(/Permission denied to enable service \[([.a-zA-Z]+)\]/);
+
+function isPermissionError(e: { context?: { body?: { error?: { status?: string } } } }): boolean {
+  return e.context?.body?.error?.status === "PERMISSION_DENIED";
+}
+
 /**
  * Attempt to enable an API on the specified project (just once).
  *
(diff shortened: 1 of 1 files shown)


src/ensureApiEnabled.ts
```diff
@@ -68,8 +74,27 @@ async function enable(projectId: string, apiName: string): Promise<void> {
       )} can't be enabled until the upgrade is complete. To upgrade, visit the following URL:
 
 https://console.firebase.google.com/project/${projectId}/usage/details`);
+    } else if (isPermissionError(err)) {
+      // Recognize permission denied errors on APIs and provide users the
+      // GCP console link to easily enable the API.
+      const permissionsError = apiPermissionDeniedRegex.exec((err as Error).message);
+      if (permissionsError && permissionsError[1]) {
```
prakhar1989, line 81: Wont this fail if permissionsError.length <2

hsubox76: Yes, but if it fails that means my regex failed and I don't have a capturing group with the string of the API name, so I can't go forward with constructing this message anyway. It will fall back to throwing the original error which should be the most useful thing we have.

src/ensureApiEnabled.ts
```diff
@@ -40,6 +40,12 @@ export async function check(
   return isEnabled;
 }
 
+const apiPermissionDeniedRegex = new RegExp(/Permission denied to enable service \[([.a-zA-Z]+)\]/);
```
joehan, line 43: This is not relevant for the CLI (since the RegExp is freshly loaded on each command), but may be for VSCode - repeatedly calling `exec` on the same RegExp can lead to some unexpected results, since exec sets RegExp.lastIndex (https://stackoverflow.com/questions/11477415/why-does-javascripts-regex-exec-not-always-return-the-same-value)

If the VSCode extension might hit this error multiple times from a long running process, consider instantiating this RegExp inside of `enable`

joehan (approved):
LGTM with one suggestion. Also, could you add an entry for this to the CLI changelog as well?

src/ensureApiEnabled.ts
hsubox76: Thanks for the catch, I've always used `match` but eslint seems to prefer `exec` so this is my first time using it, good to learn about the differences.

PR #6142, opened 2023-07-17
Jw/demo checkout

abegizaw3:


src/emulator/storage/apis/firebase.ts
```diff
@@ -93,13 +93,18 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
+
     try {
       // Both object data and metadata get can use the same handler since they share auth logic.
       ({ metadata, data } = await storageLayer.getObject({
         bucketId: req.params.bucketId,
         decodedObjectId: decodeURIComponent(req.params.objectId),
+        url: `${req.protocol}://${req.hostname}:${req.socket.localPort}`,
         authorization: req.header("authorization"),
         downloadToken: req.query.token?.toString(),
+        urlSignature: req.query["X-Firebase-Signature"] as string,
```
jmwski, line 105: We need all three of `X-Firebase-Signature`, `X-Firebase-Date` and `X-Firebase-Expires` to be present in order for us to have a usable signed url. 

It might make sense to add a nested struct:

```
export type SignedUrlParams = {
  urlSignature?: string;
  urlUsableSeconds?: string;
  urlTtlSeconds?: number;
}

export type GetObjectRequest = {
  bucketId: string;
  decodedObjectId: string;
  url?: string;
  authorization?: string;
  downloadToken?: string;
  signedUrl?: SignedUrlParams;
};
```

You could add a helper that validates these parameters that is called inside storage layer's `getObject`:

```
private validateSignedUrl(signedUrl: SignedUrlParams) {...}
```

This helper could do everything you're doing on L257-274. Having a single `SignedUrlParams` struct would make `createUnsignedUrl`'s argument list shorter.

src/emulator/storage/apis/firebase.ts
```diff
@@ -128,6 +140,47 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
 
+  firebaseStorageAPI.post(`/b/:bucketId/o/:objectId[(:)]generateSignedUrl`, async (req, res) => {
+    let signedUrlObject: SignedUrlResponse;
+
+    try {
+      signedUrlObject = await storageLayer.generateSignedUrl({
+        bucketId: req.params.bucketId,
+        decodedObjectId: decodeURIComponent(req.params.objectId),
+        authorization: req.header("authorization"),
+        originalUrl: `${req.protocol}://${req.hostname}:${req.socket.localPort}`,
```
jmwski, line 151: nit: Also call this baseUrl. If you want you could also abstract this into a helper since we use it in the GET request too.

package.json
```diff
@@ -39,7 +39,9 @@
     "test:triggers-end-to-end": "bash ./scripts/triggers-end-to-end-tests/run.sh",
     "test:triggers-end-to-end:inspect": "bash ./scripts/triggers-end-to-end-tests/run.sh inspect",
     "test:storage-deploy": "bash ./scripts/storage-deploy-tests/run.sh",
-    "test:storage-emulator-integration": "bash ./scripts/storage-emulator-integration/run.sh"
+    "test:storage-emulator-integration": "bash ./scripts/storage-emulator-integration/run.sh",
+	"mocha-units": "mocha 'src/test/emulators/storage/files.spec.ts'"
```
maneesht, line 43: Let's get rid of this change

src/emulator/storage/apis/firebase.ts
```diff
@@ -128,6 +140,47 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
+        ttlSeconds:
+          req.body.ttlSeconds === undefined ? SIGNED_URL_DEFAULT_TTL_SECONDS : req.body.ttlSeconds,
+      });
+    } catch (err) {
+      if (err instanceof NotFoundError) {
+        return res.status(404).json({
+          error: {
+            code: 404,
+            message: `Object in bucket does not exist`,
```
jmwski, line 160: Could we have `generateSignedUrl` provide us the message? It could be more descriptive and include the object path.

scripts/storage-emulator-integration/multiple-targets/.firebaserc
```diff
@@ -13,4 +13,4 @@
     }
   },
   "etags": {}
-}
+}
```
maneesht, line 16: Remove

scripts/storage-emulator-integration/run.sh
```diff
@@ -19,3 +19,4 @@ mocha scripts/storage-emulator-integration/import/tests.ts
 mocha scripts/storage-emulator-integration/multiple-targets/tests.ts
 
 mocha scripts/storage-emulator-integration/conformance/*.test.ts
+
```
maneesht, line 22: Remove

src/emulator/storage/apis/firebase.ts
```diff
@@ -93,13 +93,18 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
   firebaseStorageAPI.get("/b/:bucketId/o/:objectId", async (req, res) => {
     let metadata: StoredFileMetadata;
     let data: Buffer;
+
```
maneesht, line 96: nit: remove

src/emulator/storage/apis/firebase.ts
```diff
@@ -128,6 +140,47 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
+            code: 404,
+            message: `Object in bucket does not exist`,
+          },
+        });
+      }
+      if (err instanceof ForbiddenError) {
+        return res.status(403).json({
+          error: {
+            code: 403,
+            message: `Permission denied. No WRITE permission.`,
```
jmwski, line 168: I'd leave the WRITE permission part out. Can you also send the message from `generateSignedUrl`? It helps people debugging if the error message appears where the error originated.

src/emulator/storage/apis/firebase.ts
```diff
@@ -128,6 +140,47 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
     return res.json(new OutgoingFirebaseMetadata(metadata));
   });
 
+  firebaseStorageAPI.post(`/b/:bucketId/o/:objectId[(:)]generateSignedUrl`, async (req, res) => {
+    let signedUrlObject: SignedUrlResponse;
```
jmwski, line 144: nit: Maybe call it `signedUrlResponse: SignedUrlResponse;`

src/emulator/storage/files.ts
```diff
@@ -125,12 +156,72 @@ export class StorageLayer {
...
+   * @returns {Promise<boolean>}
+   */
+  async generateSignedUrl(request: CreateSignedUrl): Promise<SignedUrlResponse> {
+    const metadata = this.getMetadata(request.bucketId, request.decodedObjectId);
+
+    const currentDate = this.getCurrentDate();
+
+    const timeToLive = request.ttlSeconds;
+    if (!Number.isInteger(timeToLive)) {
+      throw new BadRequestError("Invalid TTL Parameter");
```
jmwski, line 174: nit: Have a more descriptive error: "Invalid TTL: ${timeToLive} is not an integer."

src/emulator/storage/files.ts
```diff
@@ -125,12 +156,72 @@ export class StorageLayer {
...
+    }
+
+    if (timeToLive < SIGNED_URL_MIN_TTL_SECONDS || timeToLive > SIGNED_URL_MAX_TTL_SECONDS) {
+      throw new BadRequestError(`TTL specified is less than 0 or more than allowed max (2 week)`);
```
jmwski, line 178: Use the constants in the error message:

```
throw new BadRequestError(`Invalid TTL: ${timeToLive} is out of range. Must be > ${SIGNED_URL_MIN_TTL_SECONDS} and < ${SIGNED_URL_MAX_TTL_SECONDS}`);
```

src/emulator/storage/apis/firebase.ts
jmwski: Also, it's a `RulesetOperationMethod.GET` permission, isn't it?

src/emulator/storage/files.ts
```diff
@@ -40,12 +46,29 @@ export class StoredFile {
...
+
 /**  Parsed request object for {@link StorageLayer#getObject}. */
 export type GetObjectRequest = {
   bucketId: string;
   decodedObjectId: string;
+  url?: string;
   authorization?: string;
   downloadToken?: string;
+  urlSignature?: string;
+  urlUsableSeconds?: string | undefined;
```
maneesht, line 70: The `?` operator already marks the field as optional, so `| undefined` is redundant

src/emulator/storage/files.ts
```diff
@@ -125,12 +156,72 @@ export class StorageLayer {
...
+      this._projectId,
+      request.authorization
+    );
+
+    if (!authorized) {
+      throw new ForbiddenError();
+    }
+
+    if (!metadata) {
+      throw new NotFoundError("Object in bucket does not exist");
```
jmwski, line 195: More descriptive error: `Object "/b/${bucketId}/o/${objectId}" does not exist.`

src/test/emulators/storage/files.spec.ts
```diff
@@ -173,3 +435,20 @@ describe("files", () => {
     const getPersistenceTmpDir = () => `${tmpdir()}/firebase/storage/blobs`;
   });
 });
+function getCurrentDate(): string {
+  return new Date().toISOString().replaceAll("-", "").replaceAll(":", "").replaceAll(".", "");
```
maneesht, line 439: You should just try and re-use the existing `getCurrentDate` from `files.ts`

src/emulator/storage/files.ts
```diff
@@ -125,12 +156,72 @@ export class StorageLayer {
...
+      bucketId: request.bucketId,
+      decodedObjectId: request.decodedObjectId,
+      url: request.originalUrl,
+      urlUsableSeconds: currentDate,
+      urlTtlSeconds: request.ttlSeconds,
+    });
+
+    const signature = createSignature(unsignedUrl);
+
+    const signedUrl = `${unsignedUrl}&X-Firebase-Signature=${encodeURIComponent(signature)}`;
```
jmwski, line 208: WDYT of collapsing these 3 lines:

```
return {
  signed_url: `${unsignedUrl}&X-Firebase-Signature=${encodeURIComponent(createSignature(unsignedUrl))}`
}
```

src/emulator/storage/files.ts
```diff
@@ -125,12 +156,72 @@ export class StorageLayer {
...
+
+    if (!authorized) {
+      throw new ForbiddenError();
+    }
+
+    if (!metadata) {
+      throw new NotFoundError("Object in bucket does not exist");
+    }
+
+    const unsignedUrl = createUnsignedUrl({
```
jmwski, line 198: nit: I think having a helper for this is a bit overkill. We could just inline a variable:

```
const unsigned = `${url}/v0/b/${bucketId}/o/${encodeURIComponent(
    decodedObjectId
  )}?alt=media&X-Firebase-Date=${urlUsableSeconds}&X-Firebase-Expires=${urlTtlSeconds}`
```

That way someone reading this function can see the full structure of the url.

src/emulator/storage/files.ts
```diff
@@ -125,12 +156,72 @@ export class StorageLayer {
...
+    };
+  }
+
   createBucket(id: string): void {
     if (!this._buckets.has(id)) {
       this._buckets.set(id, new CloudStorageBucketMetadata(id));
     }
   }
 
+  getCurrentDate() {
```
jmwski, line 221: Add a comment above this with the intended resulting timestamp format: `YYYYMMDD'T'HHMMSS'Z'`

src/emulator/storage/files.ts
```diff
@@ -149,14 +240,58 @@ export class StorageLayer {
...
-    if (!authorized) {
+
+    let checkAuth = false;
+
+    if (!hasValidDownloadToken) {
+      if (request.urlSignature) {
+        const prevSignature = request.urlSignature;
+
+        if (!request.urlUsableSeconds || !request.urlTtlSeconds) {
+          throw new BadRequestError(`Invalid ${!request.urlUsableSeconds ? "Date" : "TTL"}`);
```
jmwski, line 258: It'd be great if this error contained the name of the missing parameter. For instance, if `urlUsableSeconds` is missing, the error should read:

```
Missing required parameter X-Firebase-Date: YYYYMMDD'T'HHMMSS'Z'
```

Similar for urlTtlSeconds.

src/emulator/storage/files.ts
```diff
@@ -149,14 +240,58 @@ export class StorageLayer {
...
+        }
+        const start = convertDateToMS(request.urlUsableSeconds);
+
+        if (!start || !Number.isInteger(request.urlTtlSeconds)) {
+          throw new BadRequestError(`Invalid ${!start ? "Date" : "TTL"}`);
```
jmwski, line 263: Split out a separate error message for each of these cases: 

```
if (!start) {
  throw new BadRequestError(`Invalid format for X-Firebase-Date, expected YYYYMMDD'T'HHMMSS'Z'`) 
}
if (!Number.isInteger(request.urlTtlSeconds)) {
   throw new BadRequestError(`Invalid format for X-Firebase-Expires, not an integer.'`) 
}
```

src/emulator/storage/files.ts
```diff
@@ -149,14 +240,58 @@ export class StorageLayer {
...
+
+        const unsignedUrl = createUnsignedUrl({
+          bucketId: request.bucketId,
+          decodedObjectId: request.decodedObjectId,
+          url: request.url as string,
+          urlUsableSeconds: request.urlUsableSeconds,
+          urlTtlSeconds: request.urlTtlSeconds,
+        });
+
+        const isCorrect = createSignature(unsignedUrl) === prevSignature;
```
jmwski, line 284: You can inline this:
```
if (createSignature(unsignedUrl) !== prevSignature) {
   throw new ForbiddenError("Invalid signed url signature")
} else {
  ...
}
```

src/emulator/storage/files.ts
```diff
@@ -149,14 +240,58 @@ export class StorageLayer {
...
     const metadata = this.getMetadata(request.bucketId, request.decodedObjectId);
 
     // If a valid download token is present, skip Firebase Rules auth. Mainly used by the js sdk.
     const hasValidDownloadToken = (metadata?.downloadTokens || []).includes(
       request.downloadToken ?? ""
     );
-    let authorized = hasValidDownloadToken;
-    if (!authorized) {
+
+    let checkAuth = false;
```
jmwski, line 251: nit: Call this `hasValidSignedUrl`?

src/emulator/storage/files.ts
```diff
@@ -166,8 +301,9 @@ export class StorageLayer {
         request.authorization
       );
     }
+
     if (!authorized) {
-      throw new ForbiddenError("Failed auth");
+      throw new ForbiddenError("Permission denied. No READ permission");
```
jmwski, line 306: For the permission denied errors, we can omit the error message. This is done elsewhere.

src/emulator/storage/index.ts
```diff
@@ -103,7 +103,15 @@ export class StorageEmulator implements EmulatorInstance {
     this._uploadService.reset();
   }
 
+  /**
+   * Description placeholder
+   * @async
+   */
   async start(): Promise<void> {
+    EmulatorLogger.forEmulator(Emulators.STORAGE).log(
```
jmwski, line 111: nit: Get rid of this log.

src/test/emulators/storage/files.spec.ts
```diff
@@ -120,16 +126,49 @@ describe("files", () => {
...
         });
 
+        const currentDate = getCurrentDate();
+
+        const unsignedUrl = createUnsignedUrl({
+          bucketId: "bucket",
+          decodedObjectId: "dir%2Fobject",
+          urlTtlSeconds: 100,
+          urlUsableSeconds: currentDate,
+          url: "localhost:9000",
```
jmwski, line 160: nit: `http://localhost:9000`

src/test/emulators/storage/files.spec.ts
```diff
@@ -157,6 +196,229 @@ describe("files", () => {
           })
         ).to.be.rejectedWith(NotFoundError);
       });
+
+      it("should throw an error if TTL in Seconds aren't passed", () => {
```
jmwski, line 200: `signed url get should throw error if missing TTL`

src/test/emulators/storage/files.spec.ts
```diff
@@ -157,6 +196,229 @@ describe("files", () => {
...
+            bucketId: "bucket",
+            decodedObjectId: "dir%2Fobject",
+            urlSignature: "mockSignature",
+            urlTtlSeconds: SIGNED_URL_MIN_TTL_SECONDS,
+            urlUsableSeconds: getAdjustedDate(-1),
+          })
+        ).to.be.rejectedWith(BadRequestError);
+      });
+
+      it("should throw an error if the URL has expired. TTL is 1 Week(Max)", () => {
```
jmwski, line 240: Isn't this testing the same thing as the previous test?

src/test/emulators/storage/files.spec.ts
```diff
@@ -157,6 +196,229 @@ describe("files", () => {
...
+          })
+        ).to.be.rejectedWith(BadRequestError);
+      });
+
+      it("should throw an error if anything in the url was changed from what it originally was", () => {
+        const storageLayer = getStorageLayer(ALWAYS_TRUE_RULES_VALIDATOR);
+        const unsignedUrl = createUnsignedUrl({
+          bucketId: "10",
+          decodedObjectId: "dir%2Fobject",
+          url: "localhost:9000",
```
jmwski, line 272: nit: `http://localhost:9000`

src/test/emulators/storage/files.spec.ts
```diff
@@ -157,6 +196,229 @@ describe("files", () => {
...
+          })
+        ).to.be.rejectedWith(BadRequestError);
+      });
+
+      it("should throw an error if TTL is below the min or above the max time", () => {
+        const storageLayer = getStorageLayer(ALWAYS_TRUE_RULES_VALIDATOR);
+
+        const cases = [SIGNED_URL_MIN_TTL_SECONDS, SIGNED_URL_MAX_TTL_SECONDS];
+
+        cases.forEach((time) => {
```
jmwski, line 349: nit: `cases.forEach((case) => ...)`

src/test/emulators/storage/files.spec.ts
```diff
@@ -157,6 +196,229 @@ describe("files", () => {
...
+
+        const unsignedUrl = `${tempUnsignedUrl.replace("X-Firebase-Date=*", signedDate![0])}`;
+
+        const signature = createSignature(unsignedUrl);
+
+        const actualSignedUrl = `${unsignedUrl}&X-Firebase-Signature=${encodeURIComponent(
+          signature
+        )}`;
+
+        expect(signedUrlObject.signed_url).to.equal(actualSignedUrl);
```
jmwski, line 420: In this case `signedUrlObject` is the "actual" value, `actualSignedUrl` is the expectation right?

src/test/emulators/storage/files.spec.ts
```diff
@@ -173,3 +435,20 @@ describe("files", () => {
...
+ * @date 7/12/2023 - 4:28:20 PM
+ *
+ * @param {number} changeBy
+ * @returns {string}
+ */
+function getAdjustedDate(changeBy: number): string {
+  const newDate = new Date();
+  const adjutedDate = new Date(newDate.getTime() + changeBy * SECONDS_TO_MS_FACTOR);
+
+  return adjutedDate.toISOString().replaceAll("-", "").replaceAll(":", "").replaceAll(".", "");
```
jmwski, line 453: You can add a helper that does:

```
function getSignedUrlTimestampFor(date: Date): string = {
   return date.toISOString().replaceAll("-", "").replaceAll(":", "").replaceAll(".", "")
}
```

And then call it everywhere you do this replacement (in tests + implementation).

src/emulator/storage/files.ts
abegizaw3: the reason i have it as a helper is because i use this same block of code in my post reuqest, get request, and testing when creating unsignedUrls

src/test/emulators/storage/files.spec.ts
abegizaw3: would this be doing the same thing but just passing in what the first 2 lines are calculating?

src/test/emulators/storage/files.spec.ts
abegizaw3: I was thinking of it as I am testing signedUrlObject, so in my method I expect it to be the actualSignedUrl that I know is correct

src/test/emulators/storage/files.spec.ts
abegizaw3: I was testing that the max and min time both work

src/emulator/storage/files.ts
abegizaw3: for clarity purposes, i think having a variable would be easier to read. Especially since we don't inline the isLive variable a few lines above.

src/emulator/storage/files.ts
abegizaw3: For readability, i think keeping signature by itself then collapsing the next 2 lines might make more sense

src/emulator/storage/apis/firebase.ts
abegizaw3: also if we make SignedUrlParams optional, i think we should make the parameters in it not optional

abegizaw3: mhmmm actually nvm ^

jmwski (commented):
Looks good, I had a few more cleanup suggestions.

jmwski:
This can be done in a separate PR, but it would be great if we had SDK conformance tests, similar to those in:

* https://github.com/firebase/firebase-tools/blob/master/scripts/storage-emulator-integration/conformance/firebase-js-sdk.test.ts

But using the regular SDK (non-admin). Those changes aren't done yet, but you could work on the setup code for it and verify that you can successfully connect and do some basic ops.

package.json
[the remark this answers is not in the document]
abegizaw3: i was thinking of removing it since it was just for personal use. I could keep it if you think it'll be helpful for the future though

src/test/emulators/storage/files.spec.ts
abegizaw3: I like the idea of the helper, but in my opinion `   return getSignedUrlTimestampFor(new Date(new Date().getTime() + changeBy * SECONDS_TO_MS_FACTOR));
` is hard to read and it isnt very obvious what is happening. I think I am confused on what should be inlined and what should be separated for readability purposes

src/emulator/storage/apis/firebase.ts
```diff
@@ -108,7 +115,14 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
         return res.status(403).json({
           error: {
             code: 403,
-            message: `Permission denied. No READ permission.`,
+            message: err.message,
+          },
+        });
+      } else if (err instanceof BadRequestError) {
+        return res.status(400).json({
```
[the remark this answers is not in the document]
abegizaw3: what are the cases where we dont throw a specific error code. I would like to return some constant to represent it, but I dont know what to name it

src/emulator/storage/apis/firebase.ts
```diff
@@ -93,25 +93,32 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
-      } else if (err instanceof ForbiddenError) {
-        return res.status(403).json({
+      const errorCode = errorToHttpCode(err);
+      if (errorCode != -1) {
+        return {
           error: {
-            code: 403,
-            message: `Permission denied. No READ permission.`,
+            code: errorToHttpCode(err),
+            message: (err as Error).message,
```
maneesht, line 117: Why can't you just use the `Error` type on line 111?

package.json
```diff
@@ -39,8 +39,9 @@
     "test:triggers-end-to-end": "bash ./scripts/triggers-end-to-end-tests/run.sh",
     "test:triggers-end-to-end:inspect": "bash ./scripts/triggers-end-to-end-tests/run.sh inspect",
     "test:storage-deploy": "bash ./scripts/storage-deploy-tests/run.sh",
-    "test:storage-emulator-integration": "bash ./scripts/storage-emulator-integration/run.sh"
-  },
+    "test:storage-emulator-integration": "bash ./scripts/storage-emulator-integration/run.sh",
+	"test:storage-emulator": "mocha 'src/test/emulators/storage/**/*.spec.{ts, js}'"
```
maneesht, line 43: Just a note to remove this before we merge it in.

src/emulator/storage/apis/firebase.ts
```diff
@@ -128,6 +135,32 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
+      signedUrlResponse = await storageLayer.generateSignedUrl({
+        bucketId: req.params.bucketId,
+        decodedObjectId: decodeURIComponent(req.params.objectId),
+        authorization: req.header("authorization"),
+        baseUrl: `${req.protocol}://${req.hostname}:${req.socket.localPort}`,
+        ttlSeconds: req.body.ttlSeconds ?? SIGNED_URL_DEFAULT_TTL_SECONDS,
+      });
+    } catch (err) {
+      const errorCode = errorToHttpCode(err);
+      if (errorCode != -1) {
```
maneesht, line 150: `===`. Can you also address the rest of the linter issues?

src/emulator/storage/apis/firebase.ts
```diff
@@ -549,3 +582,17 @@ function isValidNonEncodedPathString(path: string): boolean {
 function removeAtMostOneTrailingSlash(path: string): string {
   return path.replace(/\/$/, "");
 }
+
+function errorToHttpCode(err: unknown) {
```
maneesht, line 586: This should be `Error` instead of unknown

src/emulator/storage/files.ts
```diff
@@ -149,14 +245,47 @@ export class StorageLayer {
...
+        const isCorrect = createSignature(unsignedUrl) === request.signedUrl.signature;
+        if (!isCorrect) {
+          throw new ForbiddenError("Invalid Url");
+        }
+      } else {
+        checkAuth = true;
+      }
+    }
+
+    if (checkAuth) {
```
jmwski, line 288: We can get rid of this variable and just do. We should check auth by default, and then make the exceptions clear.

```
if (!(hasValidDownloadToken || hasValidSignedUrl)) {
  ...
}
```

src/test/emulators/storage/files.spec.ts
jmwski: In typical testing language, "actual*" is the thing that gets produced by the logic under test, and "expected*" is the hard-coded expectation.

src/test/emulators/storage/files.spec.ts
```diff
@@ -173,3 +450,16 @@ describe("files", () => {
...
+ * get the current date + changeBy SECONDS
+ * @date 7/12/2023 - 4:28:20 PM
+ *
+ * @param {number} changeBy
+ * @returns {string}
+ */
+function getAdjustedDate(changeBy: number): string {
+  const currentDate = new Date().getTime();
+  const adjutedDate = new Date(currentDate + changeBy * SECONDS_TO_MS_FACTOR);
+  return getSignedUrlTimestampFor(adjutedDate);
```
jmwski, line 464: This function can be one line:
```
function getAdjustedDate(changeBy: number): string {
  return getSignedUrlTimestampFor(new Date(new Date().getTime() + changeBy * SECONDS_TO_MS_FACTOR));
}
```

src/emulator/storage/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const SIGNED_URL_MIN_TTL_SECONDS = 1;
+export const SIGNED_URL_MAX_TTL_SECONDS = 7 * 24 * 60 * 60;
+export const SIGNED_URL_DEFAULT_TTL_SECONDS = 60 * 60;
+export const SECONDS_TO_MS_FACTOR = 1000;
+export const SIGNED_URL_PRIVATE_KEY =
+  "5c39d11c20b96241c42e07218724d6726ee27c3db33eb43256c21f92bb1f52dc";
+export const INVALID_TTL_PARAM = "X-Firebase-Expires: TTL";
```
jmwski, line 7: Apologies if I was not clear, can we hardcode all instances of header names into a constant:

```
export const X_FIREBASE_EXPIRES = "X-Firebase-Expires";
export const X_FIREBASE_DATE = "X-Firebase-Date";
export const X_FIREBASE_SIGNATURE = "X-Firebase-Signature";
```

and then replace every string literal "X-Firebase-Expires" with `${X_FIREBASE_EXPIRES}`. The goal is to just have just one place where we define these string.

src/emulator/storage/apis/firebase.ts
```diff
@@ -128,6 +135,32 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
 
+  firebaseStorageAPI.post(`/b/:bucketId/o/:objectId[(:)]generateSignedUrl`, async (req, res) => {
+    let signedUrlResponse: CreateSignedUrlResponse;
+    try {
+      signedUrlResponse = await storageLayer.generateSignedUrl({
+        bucketId: req.params.bucketId,
+        decodedObjectId: decodeURIComponent(req.params.objectId),
+        authorization: req.header("authorization"),
+        baseUrl: `${req.protocol}://${req.hostname}:${req.socket.localPort}`,
+        ttlSeconds: req.body.ttlSeconds ?? SIGNED_URL_DEFAULT_TTL_SECONDS,
```
jmwski, line 146: `req.body.ttlSeconds` is not necessarily a Number right? I think we need to validate it. It's worth extracting a helper then calling it to validate ttlSeconds in both the post and get requests.

src/test/emulators/storage/files.spec.ts
abegizaw3: the only reason I have it like this is for clarity purposes. It is hard to tell what the function is doing in one line

src/emulator/storage/constants.ts
abegizaw3: Makes sense. I assumed we just wanted the invalid headers.

src/emulator/storage/apis/firebase.ts
abegizaw3: i validate it in storageLayer

src/emulator/storage/apis/firebase.ts
```diff
@@ -93,25 +93,32 @@ export function createFirebaseEndpoints(emulator: StorageEmulator): Router {
...
       // Both object data and metadata get can use the same handler since they share auth logic.
       ({ metadata, data } = await storageLayer.getObject({
         bucketId: req.params.bucketId,
         decodedObjectId: decodeURIComponent(req.params.objectId),
         authorization: req.header("authorization"),
         downloadToken: req.query.token?.toString(),
+        signedUrl: {
+          signature: req.query["X-Firebase-Signature"] as string,
+          usableSeconds: req.query["X-Firebase-Date"] as string,
+          ttlSeconds: Number(req.query["X-Firebase-Expires"]),
```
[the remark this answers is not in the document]
abegizaw3: so in the documentation, parseInt would truncate decimals. Do we want to allow users to use decimals and truncate them behind the scene?

src/emulator/storage/files.ts
abegizaw3: checkAuth just makes it more obvious of why we are making this check. With the introduction of one variable, we make it easy for a user to see why we would checkAuth following when it gets set to true.

abegizaw3: But if you think your way is better we can hop on a call to talk about it more

src/test/emulators/storage/files.spec.ts
abegizaw3: in the same file, there was a test that has 
` expect(metadata.contentType).to.equal("mime/type");
        expect(data.toString()).to.equal("Hello, World!");`
so I just followed the format that was already there. Should I change the tests in the file as well? I feel like it would be confusing if we test both ways

PR #6149, opened 2023-07-18
Clarifying the difference between experimental and preview docs for c…

egilmorez:
ontributors!

Also adding astro.md as a model for creating experimental guide docs.

Thanks!


puf:
Great work @egilmorez! Very happy to see this coming together.

src/frameworks/docs/astro.md
```diff
@@ -0,0 +1,75 @@
...
+
+You can [view your deployed app](/docs/hosting/test-preview-deploy#view-changes)
+on its live site.
+
+## Pre-render dynamic content
+
+Astro will prerender all pages to static files and will work on Firebase Hosting
+when hybrid rendering is enable for the project. Configure `output: 'hybrid'` as
+shown in
+[converting a static site to hybrid rendering](https://docs.astro.build/en/guides/server-side-rendering/#converting-a-static-site-to-hybrid-rendering`).
```
austincrim, line 60:
```suggestion
Astro will prerender all pages to static files and will work on Firebase Hosting without any configuration changes.

If you need a small set of pages to SSR, configure `output: 'hybrid'` as
shown in
[converting a static site to hybrid rendering](https://docs.astro.build/en/guides/server-side-rendering/#converting-a-static-site-to-hybrid-rendering`).
```

`output: 'hybrid'` is only needed if you want to opt-in some pages to server rendering. Astro's default, no configuration mode will prerender everything and work on Firebase Hosting.

src/frameworks/docs/astro.md
```diff
@@ -0,0 +1,75 @@
...
+
+With these settings prerendering is still the default, but you can opt in to SSR by
+adding `const prerender = false` at the top of any Astro page. Similarly,  where
+server rendering is the default you can opt in to prerendering by adding 
+`const prerender = true`.
```
austincrim, line 65:
```suggestion
With these settings prerendering is still the default, but you can opt in to SSR by
adding `const prerender = false` at the top of any Astro page. Similarly, in `output: 'server'` where
server rendering is the default you can opt in to prerendering by adding 
`const prerender = true`.
```

austincrim (approved):

alexastrum:
Looks great! Probably worth doing the same for the other experimental frameworks.

One think I'm on the fence on is using placeholders like `{{firebase_cli}}` for experimental frameworks. These are not processed by GitHub, so might be confusing to external contributors.

egilmorez:
Thanks Austin and Alex!

100% -- it would be great for the team to make a push to add pages for the other experimental frameworks.

And crap you're right, those product variables should be replaced for the experimental guides.

Issue #6150, opened 2023-07-18
Vulnerability in deps. 


JeanMeche:
Hi, I have the following deps in my project, 

```
├─┬ firebase-tools@12.4.0
│ └─┬ proxy-agent@5.0.0
│   └─┬ pac-proxy-agent@5.0.0
│     └─┬ pac-resolver@5.0.1
│       └─┬ degenerator@3.0.4
│         └── vm2@3.9.19
```

`vm2` is subject to a [critical vulnerability](https://github.com/advisories/GHSA-cchq-frgv-rjh5). 
`proxy-agent` has been updated, [`6.3.0` contains the fix](TooTallNate/proxy-agents/releases/tag/proxy-agent%406.3.0).

PR #6160, opened 2023-07-19
Updating proxy-agent

joehan:
### Description
Updating proxy-agent from 5.0.0 to 6.3.0. 5.0.0 depends on vm2@3.9.19, which [has a critical vulnerability](https://github.com/advisories/GHSA-cchq-frgv-rjh5).

In order to pick up a safe version of vm2, we need to update to a new major version of proxy-agent. This new version has proxy from env built in as the default behavior (https://www.npmjs.com/package/proxy-from-env), so we no longer need to provide the proxy URI directly. I also removed the ability to set the proxy from apiv2.Client options, since it was not used anywhere and does not work nicely with the new API for proxy-agent. 

Fixes https://github.com/firebase/firebase-tools/issues/6150

### DIFF
--- a/src/apiv2.ts
+++ b/src/apiv2.ts
@@ -1,7 +1,7 @@
 import { AbortSignal } from "abort-controller";
 import { URL, URLSearchParams } from "url";
 import { Readable } from "stream";
-import * as ProxyAgent from "proxy-agent";
+import { ProxyAgent } from "proxy-agent";
 import * as retry from "retry";
 import AbortController from "abort-controller";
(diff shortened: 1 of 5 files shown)


src/apiv2.ts
```diff
@@ -326,12 +326,8 @@ export class Client {
       compress: options.compress,
     };
 
-    if (this.opts.proxy) {
```
bkendall, line 329: I imagine if `opts.proxy` isn't used any longer, do we need to drop it from the options in the types?

joehan: Good point - done

bkendall (approved):

PR #6166, opened 2023-07-20
Next.js/glob fix

### LINKED ISSUE #6163: firebase deploy crashes due to next.js image optimization check, version 12.4.5

**firebase-tools:** 12.4.5

The issue seems to have been introduced with commit 7d2a90e6063937daaef9c9a882339991570584e2

**Platform:** ubuntu-lastest (FirebaseExtended/action-hosting-deploy@v0)

### [REQUIRED] Test case

Proprietary source code using "next/image" (image optimization with Node.js), hosted on firebase using Firebase Hosting and Firebase Functions. The code is built using GitHub Actions and deployed using FirebaseExtended/action-hosting-deploy@v0, which uses the latest firebase-tools version. The GitHub Actions job config is below.

Screenshot 2023-07-20 005700

1. Run the FirebaseExtended/action-hosting-deploy@v0 GitHub action, which pulls the latest firebase-tools version, in this case 12.4.5
2. action-hosting-deploy runs firebase deploy
3. firebase deploy crashes in **_isUsingNextImageInAppDirectory_**

firebase deploy should run successfully and identify Next.js image optimization occurences.

**Note:** even with the previous version (12.4.4), firebase-tools does not identify the usage of the image optimization in my private repository and I have to create an empty server action for the image optimization to work, otherwise it does not identify the image optimization usage and does not generate a firebase function. If I create an empty server action, then a function is generated and image optimization works as a result of the function being generated for a different purpose

firebase deploy crashes with the stack trace shown below 

Screenshot 2023-07-20 005207

leoortizz:
### Description

Fixes #6163

The usage of `Glob` caused problems in the CLI build. This PR replaces `Glob` with `sync`. 

In other parts of the frameworks code `glob` is being used as `import * as glob from "glob"`: https://github.com/firebase/firebase-tools/blob/29747a824caa5589a4e04851a4ca0fa92491f03e/src/frameworks/index.ts#L8

 but I could not find a way to stub it with Sinon for the unit tests, that's why `Glob` was used.

@joehan can you help confirming `sync` won't cause problems? Maybe because the CLI `glob` version is old? The CLI uses `7.1.2` but the latest is `10.3.3`.

### Scenarios Tested

`deploy`

### DIFF
--- a/src/frameworks/next/utils.ts
+++ b/src/frameworks/next/utils.ts
@@ -249,7 +249,7 @@ export async function isUsingNextImageInAppDirectory(
     join(projectDir, nextDir, "server", "**", "*client-reference-manifest.js")
   );
 
-  for await (const filepath of files) {
+  for (const filepath of files) {
     const fileContents = await readFile(filepath);
 
     // Return true when the first file containing the next/image component is found
--- a/src/frameworks/next/utils.ts
+++ b/src/frameworks/next/utils.ts
@@ -2,7 +2,7 @@ import { existsSync } from "fs";
 import { pathExists } from "fs-extra";
 import { basename, extname, join, posix } from "path";
 import { readFile } from "fs/promises";
-import { Glob } from "glob";
+import * as glob from "glob";
 import type { PagesManifest } from "next/dist/build/webpack/plugins/pages-manifest-plugin";
 import { coerce } from "semver";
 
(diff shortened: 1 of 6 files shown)


alexastrum (approved):

src/frameworks/next/utils.ts
```diff
@@ -245,7 +245,7 @@ export async function isUsingNextImageInAppDirectory(
   projectDir: string,
   nextDir: string
 ): Promise<boolean> {
-  const { found: files } = new Glob(
+  const files = globSync(
```
joehan, line 248: I think this should work fine - however, we probaly dont need the `await` on l252 any more.

Can you remove that and then test this again?

leoortizz: done, thanks! Just added an image test to the deploy tests too, I'll check if everything goes well

joehan (approved):
Love to see the added test too!

leoortizz:
> Love to see the added test too!

Looks like a few more changes are necessary, on it

PR #6185, opened 2023-07-26
Vscode version 0.0.24

christhompsongoogle:


firebase-vscode/package.json
```diff
@@ -22,12 +22,12 @@
       "properties": {
         "firebase.debug": {
           "type": "boolean",
-          "default": false,
+          "default": true,
```
hsubox76, line 25: Revert all the changes in this file except the version update in line 6 - all the rest is from running `swap-pkg.js monospace`. We should put a check in at some point to prevent this from being checked in.

christhompsongoogle: Done.

joehan (approved):

PR #6202, opened 2023-07-31
Better handling when updating from LOCATION to system param

joehan:
### Description
Recently, we switched official extensions from using hard coded LOCATION params to a system provided param. This led to some cases during updates where the original value chosen for LOCATION  would get replaced by a new value. This PR adds special handling for this case to smoothly use the LOCATION param as the value of the system param. Addtionally, during configures, we prompt for undefined immutable params (which should help CLI users repair affected instances).

While I was here, I also cleaned up some related, unused code.

### Scenarios Tested
I `ext:install`d `firestore-bigquery-export@0.1.32', which used a 'LOCATION' param, in `us-east1`. Then, I updated it to the latest version, which does not have `LOCATION`. I then confirmed that the manifest contained `firebaseextensions.v1beta.function/location = us-east1`, the old `LOCATION` value was removed, and there was no prompt for the value.

I also confirmed that you `ext:configure` lets you set immutable params that are undefined.

### DIFF
--- a/src/commands/ext-configure.ts
+++ b/src/commands/ext-configure.ts
@@ -76,9 +76,10 @@ export const command = new Command("ext:configure <extensionInstanceId>")
       instanceId,
       projectDir: config.projectDir,
     });
+    const params = (spec.params ?? []).concat(spec.systemParams ?? []);
     const [immutableParams, tbdParams] = partition(
-      (spec.params ?? []).concat(spec.systemParams ?? []),
-      (param) => param.immutable ?? false
+      params,
+      (param) => (param.immutable && !!oldParamValues[param.param]) ?? false
     );
     infoImmutableParams(immutableParams, oldParamValues);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Fix bug where functions:secrets:\* family of commands did not work when Firebase CLI is authenticated via GOOGLE_APPLICATION_CREDENTIALS (#6190)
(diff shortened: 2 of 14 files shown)


src/commands/ext-configure.ts
```diff
@@ -76,9 +76,10 @@ export const command = new Command("ext:configure <extensionInstanceId>")
       instanceId,
       projectDir: config.projectDir,
     });
+    const params = (spec.params ?? []).concat(spec.systemParams ?? []);
     const [immutableParams, tbdParams] = partition(
-      (spec.params ?? []).concat(spec.systemParams ?? []),
-      (param) => param.immutable ?? false
+      params,
+      (param) => (param.immutable && !!oldParamValues[param.param]) ?? false
```
elvisun, line 82: Is this the same as `!!(param.immutable && oldParamValues[param.param])`

elvisun (approved):
Do we have this problem when updating from the console?

src/commands/ext-configure.ts
elvisun: I found it to be a little easier to reason about than the `boolean|undefine && boolean` (TIL this can be undefined..)

joehan:
@elvisun I think we do - may need to bug you later for guidance on where the equivalent code lives over there

src/commands/ext-configure.ts
joehan: It is not - TS will automatically cast (string) to (bool) in an if statement, but the && operator will not

Issue #6187, opened 2023-07-26
"Could not find the ng executable." Error


jwowillo:
### [REQUIRED] Environment info

**firebase-tools:**

Firebase CLI Version: 12.4.6

**Platform:**

Ubuntu

### [REQUIRED] Test case

`firebase deploy --only hosting`

.firebaserc:
```
{
  "projects": {
    "default": "county-collector"
  },
  "targets": {},
  "etags": {}
}
```

firebase.json:
```
{
  "hosting": {
    "source": ".",
    "ignore": [
      "firebase.json",
      "**/.*",
      "**/node_modules/**"
    ],
    "frameworksBackend": {
      "region": "us-central1"
    }
  }
}
```

Outputs:
`Error: Could not find the ng executable.`

### [REQUIRED] Steps to reproduce

1. `npm install @angular/cli`: Note that ng executable path from `which ng` is `/usr/local/bin/ng`
2. Install Firebase CLI
3. `firebase deploy --only hosting`

### [REQUIRED] Expected behavior

Correctly deploys angular app or describes documentation to change ng path.

### [REQUIRED] Actual behavior

Fails to find ng executable.

```
[2023-07-26T15:09:51.656Z] > command requires scopes: ["email","openid","https://www.googleapis.com/auth/cloudplatformprojects.readonly","https://www.googleapis.com/auth/firebase","https://www.googleapis.com/auth/cloud-platform"]
[2023-07-26T15:09:51.657Z] > authorizing via signed-in user (<EMAIL>)
[2023-07-26T15:09:51.657Z] [iam] checking project county-collector for permissions ["firebase.projects.get","firebasehosting.sites.update"]
[2023-07-26T15:09:51.659Z] >>> [apiv2][query] POST https://cloudresourcemanager.googleapis.com/v1/projects/<PROJECT-NAME>:testIamPermissions [none]
[2023-07-26T15:09:51.662Z] >>> [apiv2][(partial)header] POST https://cloudresourcemanager.googleapis.com/v1/projects/<PROJECT-NAME>:testIamPermissions x-goog-quota-user=projects/county-collector
[2023-07-26T15:09:51.663Z] >>> [apiv2][body] POST https://cloudresourcemanager.googleapis.com/v1/projects/<PROJECT-NAME>:testIamPermissions {"permissions":["firebase.projects.get","firebasehosting.sites.update"]}
[...]

Issue #6173, opened 2023-07-25
Could not find the next executable


MiftahClassifieds:
### Environment info node 16.16.0
npm 8.11.0

**firebase-tools:** 12.4.4

**Platform:** nextjs

###  Steps to reproduce:
*1* create nextjs project
*2* firebase init
*3* firebase deploy

###  Actual behavior:
[info] 
   Thank you for trying our early preview of Next.js support on Firebase Hosting.
   During the preview, support is best-effort and breaking changes can be expected. Proceed with caution.

   Documentation: https://firebase.google.com/docs/hosting/frameworks/nextjs
   File a bug: https://github.com/firebase/firebase-tools/issues/new?template=bug_report.md
   Submit a feature request: https://github.com/firebase/firebase-tools/issues/new?template=feature_request.md

   We'd love to learn from you. Express your interest in helping us shape the future of Firebase Hosting: https://goo.gle/41enW5X

[error] 
[error] Error: Could not find the next executable.

aalej:
Hi @MiftahClassifieds, thanks for reaching out to us. I’m using Next.js v13.4.9 and firebase-tools v12.4.4 to replicate this issue while waiting for the fix for [#6163](https://github.com/firebase/firebase-tools/issues/6163) to be released. So far, I haven’t encountered any issues while deploying. 

Let me know if I may have made a mistake or missed anything; the steps I took were:

1. Run `npx create-next-app`
2. Run `cd my-app`
3. Downgrade Next.js to `13.4.9`
3. Run `firebase init hosting --project <project_id>`
4. Run `firebase deploy`

Could you let us know which version of Next.js you’re using and share with us the output of running `firebase deploy --debug`? Passing the `--debug` flag provides a detailed log that we can use to check what caused the error. Please remove any PII (personal identifiable information) if needed.

malikmubashar:
Exectly same issue here.
I tried more than 10 time with different projects and nextjs versions but all in vain. Always shows the same error

 "Could not find the next executable".

malikmubashar:
Here are the complete log.




[debug] [2023-07-26T13:58:04.741Z] ----------------------------------------------------------------------
[debug] [2023-07-26T13:58:04.745Z] Command:       C:\Program Files\nodejs\node.exe C:\Users\<username>\AppData\Roaming\npm\node_modules\firebase-tools\lib\bin\firebase.js deploy --debug
[...]

PR #6203, opened 2023-08-01
Disambiguate logging for framework discovery

joehan:
### Description
Minor improvement to logs to help debug issues like #6187 and #6173. My suspicion right now is that these issues represent a bunch of different cases where either npm is not available on the system (ie Firepit) or where dependencies have unusual paths in `node_modules`. I'm hoping these logs help us fix at least some of those cases.

cc @alexastrum who knows much more about this frameworks


blidd-google (approved):
LGTM

Issue #6204, opened 2023-07-18
Type Mismatch Problems While Setting region Property for Gen 2 Firebase Functions using TypeScript and a Parameterized Variable


BenJackGill:
**Issue Description:**

I am having trouble setting the `region` property of my Gen 2 Firebase Functions using TypeScript and a parameterized variable due to type mismatch issues. Here's a sample of my `.env` file:

```env
LOCATION="australia-southeast1"
```

I have attempted to define the `region` both globally and at the function level.

**1. Defining region globally using `setGlobalOptions`**

I would prefer to define the `region` globally using `setGlobalOptions`.

The global `region` property on `setGlobalOptions` has a type of `GlobalOptions.region?: string | Expression<string> | ResetValue | undefined`. This is slightly different from the function level type `HttpsOptions.region?: string | string[] | Expression<string> | ResetValue | undefined` as the global level only allows for a singular `string` and no `string[]`.

No matter what I do at the global level, it fails. All three of these attempts give a deployment error `Error: Illegal type coercion of param LOCATION to type string[]`:

```TypeScript
setGlobalOptions({
  region: defineString("LOCATION"),
});
```

```TypeScript
setGlobalOptions({
  region: defineString("LOCATION") as unknown as string,
});
```

```TypeScript
setGlobalOptions({
  region: defineList("LOCATION") as unknown as string,
});
```

Therefore, I am unsure how to define `region` on `setGlobalOptions` using a parameterized variable.

**2. Defining region at the function level**

When I tried defining `region` at the function level, the IDE showed `region` as `HttpsOptions.region?: string | string[] | Expression<string> | ResetValue | undefined`.

I thought `defineString()` would work as it returns a `string`. Although there were no errors in the IDE, during deployment I received this error `Error: Illegal type coercion of param LOCATION to type string[]`.

```TypeScript
import { onCall } from "firebase-functions/v2/https";
import { testValidators } from "../../../services/tests";
import { defineString } from "firebase-functions/params";

export const functionName = onCall(
  {
    region: defineString("LOCATION"),
  },
  (request) => {
    testValidators(request);
  }
);
```

Next, I tried to define the `region` using `defineList()`, but that produced a type error in the IDE:

```TypeScript
import { onCall } from "firebase-functions/v2/https";
import { testValidators } from "../../../services/tests";
import { defineString } from "firebase-functions/params";

export const functionName = onCall(
  {
    region: defineList("LOCATION"),
  },
  (request) => {
[...]

PR #6205, opened 2023-08-01
Fix type mismatch for parametrized function region

blidd-google:
Developers attempting to parametrize `region` using `setGlobalOptions` or at the function level were running into type mismatch errors at deploy time. This is due to the fact that the deploy logic expects the developer to provide a List param for `region`. This PR adjust the deploy logic to first attempt resolving `region` a List param, and if a type mismatch occurs, trying to resolve `region` as a String param instead.

Fixes https://github.com/firebase/firebase-tools/issues/6204.


CHANGELOG.md
```diff
@@ -1,3 +1 @@
-- Fixed issue where the Extensions emulator would error when emualting local extensions with no params. (#6271)
```
Berlioz, line 1: Is it intention that we're blowing away the existing CHANGELOG file?

Berlioz (approved):
LGTM modulo the changelog thing

paslavsky:
Hi all,
Are there any plans to release it?

Issue #6194, opened 2023-07-28
Storage hangs without completing or errors when uploading or getting files in parallel


jellynoone:
Hi, @abhis3.
I'm still experiencing this issue, although, I should add it doesn't happen all that consistantly.
Here are the following specs:
Device: `MacBook Pro (Retina, 13-inch, Mid 2014)`
OS: `macOS Big Sur, 11.7.8`
Firebase tools: `firebase-tools@12.4.5`

**Additional context**
The error doesn't just happen when uploading many files but also when downloading many of them in parallel, for instance, in context of a messaging application this would be when downloading in-chat media, like images or videos.

On my end, if I try to download sometimes 30 or sometimes more images, but at around 60 the failure rate is almost consistant. At the same time, I get the following output in the command line, where the emulators are run:
```
{"result":{"permit":true},"id":82,"status":"ok"}
{"context":{"path":"/databases/(default)/documents/..."},"action":"fetch_firestore_document","server_request_id":83}
```
The images themselves are at average between 300kb - 500kb large.

I'm running into this problem when running a Flutter application, where a call `writeToFile` for the desired images, more or less so independent of one another i.e. it is coincidental that they are called in parallel as this depends on the images found on the particular screen.

_Originally posted by @jellynoone in https://github.com/firebase/firebase-tools/issues/4918#issuecomment-1649195257_

jellynoone:
I'm reposting my comment in the linked issue since I'm still experiencing it. I'm also attaching my other responses to that issue in hopes of being helpful:

https://github.com/firebase/firebase-tools/issues/4918#issuecomment-1649205950
I also took a look at: https://github.com/firebase/firebase-tools/commit/dec8a3f0b2b3fa990da86af8b8d8926f8cf3d24e and I think I noticed a few things could be improved:
1. The test for this bug is incorrect
2. The number of test files being downloaded at the same time should probably be increased

Full reference to the test in question:
https://github.com/firebase/firebase-tools/blob/8e3e2161b1da8e97bff8edbbe570d3dd15ec5f2a/scripts/storage-emulator-integration/conformance/firebase-js-sdk.test.ts#L492-L513

The specifically possibly incorrect part:
https://github.com/firebase/firebase-tools/blob/8e3e2161b1da8e97bff8edbbe570d3dd15ec5f2a/scripts/storage-emulator-integration/conformance/firebase-js-sdk.test.ts#L502-L509

[...]

PR #6209, opened 2023-08-01
Storage tests

christhompsongoogle:
https://github.com/firebase/firebase-tools/issues/6194

Test with the following, adapted from firebase-tools/scripts/storage-emulator-integration/run.sh

./scripts/clean-install.sh
source scripts/set-default-credentials.sh
firebase setup:emulators:storage
mocha scripts/storage-emulator-integration/conformance/firebase-js-sdk.test.ts -g spammed

### DIFF
--- a/firebase-vscode/package.json
+++ b/firebase-vscode/package.json
@@ -22,12 +22,12 @@
       "properties": {
         "firebase.debug": {
           "type": "boolean",
-          "default": false,
+          "default": true,
           "description": "Enable writing debug-level messages to the file provided in firebase.debugLogPath (requires restart)"
         },
         "firebase.debugLogPath": {
           "type": "string",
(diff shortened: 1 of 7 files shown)


firebase-vscode/package.json
```diff
@@ -133,5 +133,8 @@
     "webpack": "^5.75.0",
     "webpack-cli": "^5.0.1",
     "webpack-merge": "^5.8.0"
-  }
+  },
```
joehan, line 136: Suspect these slipped in due to local vscode dev, probably should not be included here.

joehan (approved):
Change looks fine to me, just need to revert the vscode bits

Mrvirt222 (commented):
Oooooooooohhhhhh my god

PR #6211, opened 2023-08-02
Fixing a couple typos and adding version info.

egilmorez:
Making these tweak in preparation for running the sync script.  Thanks!

### DIFF
--- a/src/frameworks/docs/nextjs.md
+++ b/src/frameworks/docs/nextjs.md
@@ -14,7 +14,7 @@ Using the {{firebase_cli}}, you can deploy your Next.js Web apps to Firebase and
 serve them with {{firebase_hosting}}. The {{cli}} respects your Next.js settings and
 translates them to Firebase settings with zero or minimal extra configuration on
 your part. If your app includes dynamic server-side logic, the {{cli}} deploys that
-logic to {{cloud_functions_full}}. The CLI targets Next.js 13.4.7.
+logic to {{cloud_functions_full}}. The latest supported Next.js version is 13.4.7.
 
 <<_includes/_preview-disclaimer.md>>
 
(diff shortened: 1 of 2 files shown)


src/frameworks/docs/nextjs.md
```diff
@@ -14,14 +14,12 @@ Using the {{firebase_cli}}, you can deploy your Next.js Web apps to Firebase and
 serve them with {{firebase_hosting}}. The {{cli}} respects your Next.js settings and
 translates them to Firebase settings with zero or minimal extra configuration on
 your part. If your app includes dynamic server-side logic, the {{cli}} deploys that
-logic to {{cloud_functions_full}}.
+logic to {{cloud_functions_full}}. The CLI targets Next.js 13.4.7.
```
leoortizz, line 17: I think this should say that 13.4.7 is the latest supported version so folks don't think previous versions won't work?

egilmorez: Is it true that, if they want to use a different version from the one pinned, then they need to explicitly set it in a next.js config file?

That's the impression I got from Arthur, but I may be off base. If pinning is "retroactive," and allows any earlier version, your suggestion should work fine.  If users need to edit some config, maybe we should add a line about that and break this into a paragraph.

leoortizz: That Next.js version is only the server version, used at [firebase-frameworks](https://github.com/FirebaseExtended/firebase-framework-tools/blob/main/package.json#L94). The server should be backwards compatible.

The user can only use a different version by adding an older version of `firebase-frameworks` as a dependency on their project, but that downgrades `firebase-frameworks` and not just the Next.js version. Bottom line is the user can not choose a specific version.

Changing the `firebase-frameworks` version was just a workaround for when we had an outage caused by a broken Next.js patch version, so it's not something that should be encouraged as we're using a pinned Next.js version now.

egilmorez: I think "pinning" does not mean what I thought it meant :)

Using your wording to be safe.  Thanks!

leoortizz: oh, "pinning" the version there means that we dropped the caret ^ from the Next.js dependency in package.json, that way no version other than the one defined is used

leoortizz (approved):

PR #6213, opened 2023-08-02
Web frameworks - Fix Astro config loading

austincrim:
### Description

The web frameworks integration with Astro was broken as of `astro@2.9.7` due to internal changes on config loading. I went with the quick and easy fix here, but it would be ideal to not depend on Astro internals to resolve and load the configuration.

### Scenarios Tested

Tested with `astro@2.9.6`, and `astro@2.9.7`.


src/frameworks/astro/utils.ts
```diff
@@ -10,15 +12,31 @@ export function getBootstrapScript() {
...
-    join(astroDirectory, "dist", "core", "config", "config.js")
-  );
-  const logging: any = undefined; // TODO figure out the types here
-  const { astroConfig: config } = await openConfig({ cmd: "build", cwd, logging });
+  const version = getAstroVersion(cwd);
+
+  let config;
+  if (gte(version!, "2.9.7")) {
+    const { resolveConfig } = await dynamicImport(
+      join(astroDirectory, "dist", "core", "config", "config.js")
```
taeold, line 20: nit* Seems like this line and L26 is identical - do we want to extract this out as a variable so we can more easily identify difference between how we handle versions <2.9.7 and >=2.9.7?

taeold (approved):

PR #6238, opened 2023-08-09
VSCode plugin: fix predeploy call

hsubox76:
The predeploy command for the CLI is called in a pretty tricky way that doesn't work in the VSCode plugin:

1) It relies on `process.env.execPath` to get the Node executable. In the VSCode plugin environment this actually contains the VSCode executable which is kind of annoying. I talked to Joe and the best guess for why the CLI uses `process.env.execPath` instead of just `node` is so that it will work in the standalone version of the CLI.

2) It calls `cross-env-shell.js`, an executable in a `bin` directory in `node_modules/cross-env` and it gets the path using require.resolve(). The node_modules dir path isn't accessible from the vscode packaged dist folder.

To fix this I detect if we're in a vscode extension environment by looking for `process.env.VSCODE_CWD` which should be pretty reliable?

Then in each case above:

1) call just `node` instead. This should be fine, node should always be available in a vscode plugin environment.
2) add the cross-env-shell dist directory to the list of things webpack copies over into the vscode dist output folder, and then use __dirname to make sure it finds that path

Tested a deploy with the CLI as well and it seems like it continues to output the same string as before, as long as VSCODE_CWD isn't set.


joehan (approved):
LGTM

PR #6256, opened 2023-08-15
Vscode emulators

christhompsongoogle:


firebase-vscode/src/workflow.ts
```diff
@@ -230,6 +234,43 @@ export async function setupWorkflow(
     broker.send("notifyPreviewChannelResponse", { id: response });
   });
 
+  broker.on(
```
hlshen, line 237: do we want all the emulator registration in workflow (what does workflow mean)? I think the amount of code warrants an emulator.ts file.

firebase-vscode/src/cli.ts
```diff
@@ -289,3 +293,29 @@ export async function deployToHosting(
...
   return { success: true, hostingUrl: "", consoleUrl: "" };
 }
+
+export async function emulatorsStart(emulatorUiSelections: EmulatorUiSelections) {
+  const commandOptions = await getCommandOptions(undefined, {
+    ...currentOptions,
+    project: emulatorUiSelections.projectId,
+    exportOnExit: emulatorUiSelections.exportStateOnExit,
+    import: emulatorUiSelections.importStateFolderPath,
+    only: emulatorUiSelections.mode === "hosting" ? "hosting" : ""
```
hlshen, line 303: can we make an enum for modes? eventually we'll add firemat :)

hlshen (changes requested):

hsubox76 (approved):

hsubox76:
FYI to Chris and any other reviewers, I can make the modifications we discussed in another PR after this is merged (gate emulators feature based on a VSCode setting, try to find a better solution for the string-replace TODOs), as I know there's some urgency to get the current state of work checked in.

firebase-vscode/src/cli.ts
```diff
@@ -289,3 +293,29 @@ export async function deployToHosting(
...
+  // Adjusts some options, export on exit can be a boolean or a path.
+  commandUtils.setExportOnExitOptions(commandOptions as commandUtils.ExportOnExitOptions);
+  return startAllEmulators(commandOptions, /*showUi=*/ true);
+}
+
+export async function stopEmulators() {
+  await stopAllEmulators();
+}
+
+export function listRunningEmulators(): EmulatorInfo[] {
```
joehan, line 314: @hlshen this is how we'd grab the port for firemat

PR #6257, opened 2023-08-15
Refactor webview registration to cut down on boilerplate

tjlav5:
Visual no-op; but cuts down on the boilerplate needed to register a new webview.


firebase-vscode/package.json
```diff
@@ -64,7 +64,7 @@
       "firebase": [
         {
           "type": "webview",
-          "id": "firebase.sidebarView",
+          "id": "sidebar",
```
hsubox76, line 67: Any chance of collisions with other extensions creating a webview named "sidebar" or will this be scoped to the parent "firebase"?

hsubox76 (approved):
Just one question, but it's likely you researched it already.

PR #6259, opened 2023-08-15
Consolidate extension-state into context

tjlav5:
* Ability to easily re-use in other webviews
* Quick access to state via hooks
* Signals are the new hot-ness, surgical UI updates
* BONUS: enable strict-mode


firebase-vscode/webviews/globals/app.tsx
```diff
@@ -0,0 +1,10 @@
+import React, { ReactNode, StrictMode } from "react";
+import { ExtensionStateProvider } from "./extension-state";
+
+export function App({ children }: { children: ReactNode }): JSX.Element {
```
hsubox76, line 4: Comment on the purpose of this (to wrap every webview in a common template, which to start off with is just StrictMode)

firebase-vscode/webviews/globals/extension-state.tsx
```diff
@@ -0,0 +1,64 @@
...
+    broker.on("notifyUserChanged", ({ user }) => {
+      state.selectedUserEmail.value = user.email;
+    });
+
+    broker.on("notifyProjectChanged", ({ projectId }) => {
+      state.projectId.value = projectId;
+    });
+
+    broker.send("getInitialData");
+  }, [state]);
```
hsubox76, line 55: If I understand useEffect correctly, this will add these listeners and send "getInitialData" every time state changes, is that bad? Or do signals make it so that `state` doesn't seem to change and so it won't trigger this?

firebase-vscode/webviews/globals/extension-state.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+const ExtensionState =
+  createContext<ReturnType<typeof createExtensionState>>(null);
+
+export function ExtensionStateProvider({
+  children,
+}: {
+  children: ReactNode;
+}): JSX.Element {
+  const state = createExtensionState();
```
hsubox76, line 33: Will this run every time the component re-renders? Would it create a duplicate ExtensionState?

firebase-vscode/webviews/globals/app.tsx
```diff
@@ -0,0 +1,10 @@
+import React, { ReactNode, StrictMode } from "react";
```
hsubox76, line 1: Nit: here and with extension-state.tsx - I think it's a convention to name all React components Pascal case although I know there is an edge case where some OS's have some issues with Pascal case, but we should probably have a consistent naming convention. So if you're concerned about that we could convert the existing components to snake case.

firebase-vscode/webviews/SidebarApp.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { StrictMode, useEffect, useState } from "react";
```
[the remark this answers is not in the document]
tjlav5: no need to import it here, that was a vestige from experimenting, removed.

firebase-vscode/webviews/globals/extension-state.tsx
```diff
@@ -0,0 +1,64 @@
+import React, { createContext, ReactNode, useContext, useEffect } from "react";
+import { broker } from "./html-broker";
+import { signal, computed } from "@preact/signals-react";
+import { User } from "../types/auth";
+
+export enum Environment {
+  UNESPECIFIED,
```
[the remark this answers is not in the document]
tjlav5: good catch, fixed

firebase-vscode/webviews/globals/app.tsx
tjlav5: Ack, that's right, and added a comment.

firebase-vscode/webviews/globals/extension-state.tsx
tjlav5: So it would change anytime the reference to "state" changes, not the value itself.

firebase-vscode/webviews/globals/extension-state.tsx
tjlav5: Technically very much yes, but that's why it's all the way up at the top of the React tree. I'll add some comments.

firebase-vscode/webviews/globals/app.tsx
tjlav5: ack - yeah, I'm on the fence here... since this file is exporting a component and a hook, does the component take precedence? also, funny you mentioned snake-case, because g3 is very much a snake-case environment, but it seems like this repo is more of a hyphen-env?

tjlav5: Going to keep as-is, but will give this some more thought and try to come up with concrete style-guide on when to use kebab vs pascal.

Issue #6193, opened 2023-07-27
next.config.js bundle timeout deploys a broken build


rasendubi:
https://github.com/firebase/firebase-tools/pull/5691 [introduced](https://github.com/firebase/firebase-tools/commit/e5981f2e0f40ff87139039cb5ac5004231f2e69f#diff-372d171b067c11f7dbdea434bc98b44c356278e71e0f1251758e4ca25517c0b8R459) a 10s timeout for bundling next.config.js. This bundling uses `npx esbuild next.config.js` command and it may timeout for a variety of reasons (most likely, network-related).

When this happens, the code fallbacks to copying next.config.js to the output. However, if next.config.js imports any other local file, copying next.config.js is not enough (as the other file needs to be copied as well).

This led to our CI silently deploying a broken build to production (sometimes) which then failed with hard to understand errors.

There are a couple of possible fixes, and I'm not sure which one is better:
- remove timeout for esbuild
- add `esbuild` to firebase-tools' dependencies, so bundling does not depend on network speed
- in fallback, copy all source files to output
- in fallback, stop the build

cc @jamesdaniels @leoortizz 

### [REQUIRED] Environment info

**firebase-tools:** anything after https://github.com/firebase/firebase-tools/commit/e5981f2e0f40ff87139039cb5ac5004231f2e69f (11.30.0+)

**Platform:** any

### [REQUIRED] Test case

next.config.js:
```js
require('./other');

module.exports = {};
```

other.js:
```js
```

### [REQUIRED] Steps to reproduce

- flush npm/npx cache: `rm -rf "$(npm config get cache)/_npx"`
- run `env FIREBASE_CLI_EXPERIMENTS=webframeworks firebase deploy`
- loop until fails

### [REQUIRED] Expected behavior

It works

### [REQUIRED] Actual behavior

SSR is deployed but fails at runtime.

### Workaround
The simplest workaround is to add `esbuild` to your project's dependencies. This will cause `npx` to use project-installed version, so it can't timeout due to network

leoortizz:
Hey @rasendubi, thanks for reporting. I discussed this with @alexastrum and we decided to increase the timeout to 60 seconds to workaround the network issues. It will be available in the next release.

rasendubi:
Hey @leoortizz. Increasing the timeout doesn't sound like a solution.

It surely decreases the chance of failure but I still don't like that there
is a 0.1% chance that it will silently deploy crap instead of my app. I
[...]

PR #5499, opened 2023-02-09
Fix Next.js deployments on Windows

leoortizz:
### Description

Wrapping up @jamesdaniels PR #5393 aiming to fix #5369. Ended up that the error was only happening on Windows due to `spawn` executing commands which were not recognized by the terminal. The fix was to replace `spawn` and `spawnSync` with its equivalent from [`cross-spawn`](https://www.npmjs.com/package/cross-spawn), which is a cross platform implementation of `spawn` that's [already used in `firebase-tools`](https://github.com/firebase/firebase-tools/blob/master/package.json#L111).

### Scenarios Tested

Tested this on both MacOs and Windows devices and `firebase deploy` worked following these steps:

1. Create a project using `npx create-next-app@latest`
2. Build this branch with `npm run build`
3. From the Next.js project folder, run `node ../firebase-tools/lib/bin/firebase.js init hosting`
4. `node ../firebase-tools/lib/bin/firebase.js deploy`

### Sample Commands

--

PR #5393, opened 2023-01-04
Stream npm ls when bundling Next.js

jamesdaniels:

PR #6269, opened 2023-08-16
Remove fallback for npx esbuild next.config.js

alexastrum:
### Description

Fixed: Next.config.js bundle timeout deploys a broken build #6193

### DIFF
--- a/src/frameworks/next/index.ts
+++ b/src/frameworks/next/index.ts
@@ -484,52 +484,44 @@ export async function ɵcodegenFunctionsDirectory(sourceDir: string, destDir: st
   // Alternatively I tried using @swc/spack and the webpack bundled into Next.js but was
   // encountering difficulties with both of those
   if (existsSync(join(sourceDir, "next.config.js"))) {
-    try {
-      const productionDeps = await new Promise<string[]>((resolve) => {
-        const dependencies: string[] = [];
-        const npmLs = spawn("npm", ["ls", "--omit=dev", "--all", "--json=true"], {
-          cwd: sourceDir,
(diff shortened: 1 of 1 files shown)


alexastrum:
@leoortizz 

I see that the fallback was added here: https://github.com/firebase/firebase-tools/commit/d46ae1fbc6d9ba6f5f4694d8308ca172fc476db4

Would removing it create issues on Windows? My understanding is that the `ls` command should still work.

Are there any edge cases? While increasing the timeout is the short-term workaround that is a low hanging fruit, a permanent fix for #6193 should involve (at least allowing) removal of the fallback.

leoortizz:
> @leoortizz
> 
> I see that the fallback was added here: [d46ae1f](https://github.com/firebase/firebase-tools/commit/d46ae1fbc6d9ba6f5f4694d8308ca172fc476db4)
> 
> Would removing it create issues on Windows? My understanding is that the `ls` command should still work.
> 

#5499 also had #5393, so it had more than one change. The fallback wasn't related to Windows. The Windows fix on that PR was using `cross-spawn` instead of `spawn` from `child_process`, as Node.js's spawn requires an alternative implementation for Windows.

> Are there any edge cases? While increasing the timeout is the short-term workaround that is a low hanging fruit, a permanent fix for #6193 should involve (at least allowing) removal of the fallback.

I'm currently investigating it, the issue can be related to different `npx` versions, mainly the way that `npx --yes` is handled

leoortizz (approved):

leoortizz:
@alexastrum I just created #6287 that adds another fix to `npx esbuild`

PR #6289, opened 2023-08-22
add Vite and related frameworks readme

austincrim:
### Description
Adds a guide for the Svelte web framework integration.

### Outstanding question

React, Preact, Svelte, and Lit are all built on the Vite integration, thus their instructions are essentially identical. 

We already have `react.md` and we could still create separate READMEs for each framework, though that would result in some duplication. 

Alternatively, we could create a single `vite.md` and reference all the compatible frameworks in a single README, though that could be less discoverable for those looking for a specific framework. @egilmorez let me know your thoughts.


egilmorez:
On the outstanding question -- if this were for Devsite production docs, my answer would be yes, let's have different "container files" where the h1 or title names the specific framework, and then use an include across all four or five files (using Django variables for any framework-specific details).

But while we are still presenting the docs right here in the repo, the single Vite file would be a lot easier to maintain.

austincrim:
@egilmorez I think I'm in agreement. I can remove `react.md` and `svelte.md` and include a generic `vite.md` in this PR, if you can confirm that's the direction we should go.

egilmorez:
SG Austin.  Let's keep the fullest version of the content in vite.md

On removing the others though -- as a nod to better discoverability we *could* keep very brief .md files in each of the corresponding directories and in those files just basically clarify that this framework is built on the Vite integration, and that users should see the Vite guide for full guidance.

austincrim:
@egilmorez done!

src/frameworks/docs/vite.md
```diff
@@ -0,0 +1,51 @@
...
+## Serve static content
+
+After initializing Firebase, you can serve static content with the standard
+deployment command:
+
+```shell
+firebase deploy
+```
+
+You can [view your deployed app](/docs/hosting/test-preview-deploy#view-changes)
```
egilmorez, line 50: Just realized that this will need to be an absolute link to reach Devsite: https://firebase.google.com/docs/hosting/test-preview-deploy#view-changes

src/frameworks/docs/vite.md
```diff
@@ -0,0 +1,51 @@
...
+release is not subject to any SLA or deprecation policy and may receive limited
+or no support.
+
+## Before you begin
+
+Before you get started deploying your app to Firebase,
+review the following requirements and options:
+
+- Firebase CLI version 12.1.0 or later. Make sure to
+  [install the CLI](/docs/cli#install_the_firebase_cli) using your preferred
```
egilmorez, line 18: Just realized that this will need to be an absolute link to reach Devsite: https://firebase.google.com/docs/cli#install_the_firebase_cli

src/frameworks/docs/vite.md
```diff
@@ -0,0 +1,51 @@
...
+## Before you begin
+
+Before you get started deploying your app to Firebase,
+review the following requirements and options:
+
+- Firebase CLI version 12.1.0 or later. Make sure to
+  [install the CLI](/docs/cli#install_the_firebase_cli) using your preferred 
+	method.
+- Optional: An existing Vite project. You can create one with 
+	`npm create vite@latest` or let the Firebase CLI
```
egilmorez, line 21: This link isn't doing much lifting in a page of this scope, so let's just remove it:

initialize a new project for you.

egilmorez (commented):
Thanks Austin!  Found a few things we should fix before merging, but otherwise looks great.

austincrim:
@egilmorez comments addressed, looks like we had one flaky E2E test but that shouldn't be related to my cnanges.

egilmorez (approved):
Thanks for working through this Austin!  Much appreciated.

PR #6303, opened 2023-08-25
Add `nuxt.md` to docs

JulienMartel:
Adds a guide for the `Nuxt` web framework integration, following `sveltekit.md`.


src/frameworks/docs/sveltekit.md
```diff
@@ -17,7 +17,7 @@ Before you get started deploying your app to Firebase,
 review the following requirements and options:
 
 - Firebase CLI version 12.1.0 or later. Make sure to
-  [install the CLI](/docs/cli#install_the_firebase_cli)
+  [install the CLI](https://github.com/firebase/firebase-tools#installation)
```
egilmorez, line 20: In a similar PR from Austin I suggested this absolute link for a fix:

https://firebase.google.com/docs/cli#install_the_firebase_cli

But if there's a reason to keep everything in GH, or if only npm works for the frameworks CLI, we can use this (it needs to be fixed in a few other places I'm afraid).

src/frameworks/docs/nuxt.md
```diff
@@ -0,0 +1,64 @@
...
+release is not subject to any SLA or deprecation policy and may receive limited
+or no support.
+
+## Before you begin
+
+Before you get started deploying your app to Firebase,
+review the following requirements and options:
+
+- Firebase CLI version 12.1.0 or later. Make sure to
+  [install the CLI](https://github.com/firebase/firebase-tools#installation)
```
egilmorez, line 20: In a similar PR from Austin I suggested this absolute link for a fix:

https://firebase.google.com/docs/cli#install_the_firebase_cli

But if there's a reason to keep everything in GH, or if only npm works for the frameworks CLI, we can use this (it needs to be fixed in a few other places I'm afraid).

src/frameworks/docs/nuxt.md
```diff
@@ -0,0 +1,64 @@
...
+  using your preferred method.
+- Optional: Billing enabled on your Firebase project
+  (required if you plan to use SSR)
```
egilmorez, line 23: If an existing Nuxt project is required, would that mean that "Initialize a new project" as detailed below doesn't work?

egilmorez (commented):
Thanks Julien!  Here are a few things to look/questions.

src/frameworks/docs/sveltekit.md
JulienMartel: Ah, I was looking for an official page in the docs, thanks!

The one [I linked to in the change](https://github.com/firebase/firebase-tools#standalone-binary) has a command for installing "Standalone Binary" as well, but I prefer the google.com link!

I'll go ahead and look for other occurrences to replace too.

src/frameworks/docs/nuxt.md
JulienMartel: Hmm let me refactor this.

JulienMartel: This seems to refer to two separate "projects": one being the firebase hosting project, and the other is referring to your nuxt project.

src/frameworks/docs/sveltekit.md
egilmorez: Thanks for getting the Astro doc fixed!

egilmorez (approved):
Thanks!

src/frameworks/docs/nuxt.md
leoortizz: @egilmorez thanks for pointing that out! @JulienMartel I was checking and actually the Nuxt implementation does not have an init method, therefore Nuxt isn't an option in `firebase init hosting`.

We can remove `Initialize a new project` in this PR and add it back after adding the `init` method to Nuxt, which should be simple.

leoortizz (changes requested):

src/frameworks/docs/nuxt.md
JulienMartel: Sounds good. Thanks for catching that.

egilmorez (approved):
Looks like everything is resolved; let's get those link fixes in!  Thanks very much for getting those.

leoortizz:
@joehan any idea why the integration tests are always being skipped in this PR even though the other checks are passing? It's blocking the merge

PR #6304, opened 2023-08-25
Allow flask entry file as an init  config option

chalosalvador:
### Description

- Detect flask app and ask for the entry file name
- Save the entry file  in firebase.json under `frameworksBackend.flask`
- Update firebase config schema

### Scenarios Tested
- Init with an existing Flask app with entry file = app.py.
- Init with an existing Flask app with entry file = main.py.
- Init with a blank project.
- Start emulators
- Deploy app

### DIFF
--- a/src/firebaseConfig.ts
+++ b/src/firebaseConfig.ts
@@ -110,6 +110,10 @@ interface FrameworksBackendOptions extends HttpsOptions {
   region?: string;
   // Invoker can only be public
   invoker?: "public";
+  // To allow dynamic entry file for Flask apps
+  flask: {
+    entryFile?: string;
+  };
 }
 
 export type HostingBase = {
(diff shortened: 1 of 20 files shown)


src/firebaseConfig.ts
```diff
@@ -110,6 +110,10 @@ interface FrameworksBackendOptions extends HttpsOptions {
   region?: string;
   // Invoker can only be public
   invoker?: "public";
+  // To allow dynamic entry file for Flask apps
+  flask: {
```
leoortizz, line 114: should `flask` be optional?
```suggestion
  flask?: {
```

src/frameworks/interfaces.ts
```diff
@@ -79,24 +85,32 @@ export interface Framework {
...
+
+export type CodegenPublicDirectoryOptions = {
+  dest?: string;
+  target?: string;
+  context?: {
+    project: string;
+    site: string;
+  };
+  frameworksBackend?: HostingBase["frameworksBackend"];
+};
```
leoortizz, line 112: can't we make `options`(the function param), `dest`, `target` and `context` required as they were before? That way we won't need to check and throw in every framework as those fields were already required. I think it should be covered by the type definition

leoortizz, line 112: can't we make `options`(the function param), `dest`, `target` and `context` required as they were before? That way we won't need to check and throw in every framework as those fields were already required. I think it should be covered by the type definition

alexastrum (approved):

PR #6309, opened 2023-08-29
add nuxt init method for firebase hosting

JulienMartel:
Nuxt included in the list of frameworks to choose from when initializing a new project using `firebase init hosting`


src/frameworks/nuxt/index.ts
```diff
@@ -111,3 +112,16 @@ export async function getConfig(dir: string): Promise<NuxtOptions> {
   const { loadNuxtConfig } = await relativeRequire(dir, "@nuxt/kit");
   return await loadNuxtConfig(dir);
 }
+
+/**
+ * Utility method used during project initialization.
+ */
+export function init(setup: any, config: any) {
+  return new Promise<void>((resolve) => {
```
inlined, line 120: If you're returning a promise just to conform to an interface, you can just return Promise.resolve(). The callback passed to `new Promise` is run synchronously.

jamesdaniels (approved):

PR #6314, opened 2023-08-31
VSCode plugin: Set up working unit test framework

hsubox76:
The default test pipeline provided with the VSCode starter repo uses `tsc` to compile. This won't work here, as we're using a number of webpack modifications and hacks to make this extension work with CLI code.

I've changed the pipeline to compile with webpack using multiple entry points (all test/suite/**.test.ts files) to multiple output files, which the default test runner scripts will run. The test runner scripts themselves are ts files and are compiled with tsc (the tsconfig should only compile runTest.ts and index.ts and avoid compiling *test.ts files, which should be left to webpack).

One caveat is that sinon does not work out of the box to stub methods from other files, due to being incompatible with ES6 (imported ES6 modules are considered non-writable). There are some workarounds for this.

This also doesn't have any tests for the UI. VSCode Extension Tester seems like a promising tool for doing that: https://github.com/redhat-developer/vscode-extension-tester/wiki It uses selenium-webdriver.


firebase-vscode/package.json
```diff
@@ -94,7 +94,8 @@
     "build": "npm run copyfiles && webpack --config webpack.prod.js --devtool hidden-source-map",
     "build:extension": "webpack --config webpack.prod.js --config-name extension",
     "build:sidebar": "npm run copyfiles && webpack --config webpack.prod.js --config-name sidebar",
-    "pretest": "npm run dev ; npm run lint && tsc -p src/test/tsconfig.test.json",
+    "test-compile": "npm run copyfiles && webpack --config src/test/webpack.test.js",
```
hlshen, line 97: should prepublish or package be running tests?

firebase-vscode/src/test/tsconfig.test.json
```diff
@@ -6,6 +6,7 @@
 		"outDir": "../../dist/test",
 		"sourceMap": true,
 		"rootDir": "./",
-		"strict": true,
-	}
+		"strict": true
+	},
+	"exclude": ["**/*.test.ts"]
```
hlshen, line 11: can you explain this? i'm confused why we're excluding here...

firebase-vscode/src/test/webpack.test.js
```diff
@@ -0,0 +1,56 @@
...
+          testFiles[fileName] = path.resolve(__dirname, "suite", file);
+        }
+        resolve(testFiles);
+      }
+    );
+  });
+
+async function getTestConfig() {
+  const testFiles = await getTestFiles();
+  console.log(testFiles);
```
hlshen, line 29: checking if this is a purposeful log

firebase-vscode/src/test/webpack.test.js
```diff
@@ -0,0 +1,56 @@
...
+const getTestFiles = () =>
+  new Promise((resolve, reject) => {
+    glob(
+      "**/**.test.ts",
+      { cwd: path.resolve(__dirname, "suite") },
+      (err, files) => {
+        if (err) {
+          reject(e(err));
+        }
+        const testFiles = {};
```
hlshen, line 17: I'm guessing this is necessary? Didn't think we'd have to search for test files manually.

hlshen (approved):

firebase-vscode/package.json
hsubox76: I think it can be added later when the tests do something. My preference would be I would want an overarching release process that runs each of the steps (test, build, package) separately, for more fine grained control.

firebase-vscode/src/test/tsconfig.test.json
hsubox76: `tsc` is only used to compile the 2 test scaffolding files (runTest.ts and index.ts). We don't want it to run on the src files, as it won't work without webpack. All the *.test.ts files should be run through webpack, not raw tsc. The tsc process (last step in `pretest`) uses the tsconfig to determine what files it compiles, the webpack process sort of overrides this by specifying specific `entry` files, and when you do that, `exclude` won't exclude anything those files import.

firebase-vscode/src/test/webpack.test.js
hsubox76: I'll remove this.

firebase-vscode/src/test/webpack.test.js
hsubox76: We need to be very specific in telling webpack exactly which files are test suite files, and make sure not to include the test scaffolding files. On second thought I think maybe the webpack compile process is good enough to use on the test scaffolding files too and we don't need two compile processes. I think it's good to keep the processes separate somewhat as they do different things (runTest.ts/index.ts set up the test process, *.test.ts are the actual tests, which may import CLI code, dependencies, etc).

Or let me know if I misunderstood the question.

firebase-vscode/package.json
joehan: +1 to keeping each script to a single resposibility

joehan (approved):

PR #6332, opened 2023-09-07
Resolve vpc value prior to checking it's value

### LINKED ISSUE #6327: Unable to deploy with VPC from external shared project

**firebase-tools:** 12.5.2

**Platform:** macos

### [REQUIRED] Test case

Trying to use a shared vpc from a different project with `.env` and

```
FUNCTION_VPC="projects/shared/locations/us-central1/connectors/shared"
```

```js
const vpc = params.defineString('FUNCTION_VPC');
setGlobalOptions({
  vpcConnector: vpc
});
```

Try to deploy with these values and firebase tools mutates the `vpcConnector` string to an invalid value.

```
functions: HTTP Error: 400, Could not update Cloud Run service projects/my-project/locations/us-central1/services/function. spec.template.metadata.annotations: Expected a VPC access connector name like projects/*/locations/*/connectors/*, but obtained projects/not-shared-project/locations/us-central1/connectors/{{the full value of the FUNCTION_VPC}}
```

I assume the issue is around https://github.com/firebase/firebase-tools/blob/09c8d88ff5f1b4c4e617064c7da729b2301d905b/src/deploy/functions/runtimes/node/parseTriggers.ts#L502-L514

But my connector has a `/` in it so it shouldn't get modified. I wonder if it has to do with the `Expression<string>` type that has a `.value()` not being invoked or something until later? Or maybe the `/` is encoded when read from the environment file?

The VPC value I specify is used

The string is mutated and the deployment fails.

steveoh:
### Description

fixes #6327

### DIFF
--- a/src/deploy/functions/build.ts
+++ b/src/deploy/functions/build.ts
@@ -494,10 +494,10 @@ export function toBackend(
         nullsafeVisitor((cpu) => (cpu === "gcf_gen1" ? cpu : r.resolveInt(cpu)))
       );
       if (bdEndpoint.vpc) {
+        bkEndpoint.vpc = { connector: params.resolveString(bdEndpoint.vpc.connector, paramValues) };
         if (bdEndpoint.vpc.connector && !bdEndpoint.vpc.connector.includes("/")) {
           bdEndpoint.vpc.connector = `projects/${bdEndpoint.project}/locations/${region}/connectors/${bdEndpoint.vpc.connector}`;
         }
-        bkEndpoint.vpc = { connector: params.resolveString(bdEndpoint.vpc.connector, paramValues) };
         proto.copyIfPresent(bkEndpoint.vpc, bdEndpoint.vpc, "egressSettings");
       } else if (bdEndpoint.vpc === null) {
         bkEndpoint.vpc = null;
--- a/src/deploy/functions/build.ts
+++ b/src/deploy/functions/build.ts
@@ -498,7 +498,7 @@ export function toBackend(
         if (bdEndpoint.vpc.connector && !bdEndpoint.vpc.connector.includes("/")) {
           bdEndpoint.vpc.connector = `projects/${bdEndpoint.project}/locations/${region}/connectors/${bdEndpoint.vpc.connector}`;
         }
-        
+
         bkEndpoint.vpc = { connector: bdEndpoint.vpc.connector };
         proto.copyIfPresent(bkEndpoint.vpc, bdEndpoint.vpc, "egressSettings");
       } else if (bdEndpoint.vpc === null) {
--- a/src/deploy/functions/build.ts
+++ b/src/deploy/functions/build.ts
@@ -494,10 +494,12 @@ export function toBackend(
         nullsafeVisitor((cpu) => (cpu === "gcf_gen1" ? cpu : r.resolveInt(cpu)))
       );
       if (bdEndpoint.vpc) {
-        bkEndpoint.vpc = { connector: params.resolveString(bdEndpoint.vpc.connector, paramValues) };
+        bdEndpoint.vpc = { connector: params.resolveString(bdEndpoint.vpc.connector, paramValues) };
         if (bdEndpoint.vpc.connector && !bdEndpoint.vpc.connector.includes("/")) {
           bdEndpoint.vpc.connector = `projects/${bdEndpoint.project}/locations/${region}/connectors/${bdEndpoint.vpc.connector}`;
         }
+        
+        bkEndpoint.vpc = { connector: bdEndpoint.vpc.connector };
         proto.copyIfPresent(bkEndpoint.vpc, bdEndpoint.vpc, "egressSettings");
       } else if (bdEndpoint.vpc === null) {
         bkEndpoint.vpc = null;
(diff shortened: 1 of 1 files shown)


steveoh:
I'd love to got a workflow approval for this to see the tests run!

steveoh:
@joehan 🔔

joehan:
Hey @steveoh, thanks for the @ on this, it managed to get lost in the mix. I've approved the test run and it looks like there is a unit test case failing.

I've also assigned some folks who work on CF3 to review.

steveoh:
Thanks @joehan. 

For time savings this is the failed test

```
1) addResourcesToBuild
       should expand vpc connector in w/ shorthand form:

      AssertionError: expected { Object (requiredAPIs, endpoints, ...) } to deeply equal { Object (requiredAPIs, endpoints, ...) }
      + expected - actual

               "project": "project"
               "region": "us-central1"
               "runtime": "nodejs16"
               "vpc": {
      -          "connector": "hello-vpc"
      +          "connector": "projects/project/locations/us-central1/connectors/hello-vpc"
               }
             }
           }
         }
```

I think this test is wrong and should be updated. The values are equivalent basically and the long form could be seen as more correct™ as it allows for more use cases? Thoughts?

src/deploy/functions/build.ts
```diff
@@ -494,10 +494,10 @@ export function toBackend(
         nullsafeVisitor((cpu) => (cpu === "gcf_gen1" ? cpu : r.resolveInt(cpu)))
       );
       if (bdEndpoint.vpc) {
+        bkEndpoint.vpc = { connector: params.resolveString(bdEndpoint.vpc.connector, paramValues) };
```
colerogers, line 497: You want to resolve the `bdEndpoint.vpc.connector` and save that back into `bdEndpoint` so the `if` check below works. Something like this should do it:
```
if (bdEndpoint.vpc) {
        **bdEndpoint.vpc.connector = params.resolveString(bdEndpoint.vpc.connector, paramValues);**
        if (bdEndpoint.vpc.connector && !bdEndpoint.vpc.connector.includes("/")) {
          bdEndpoint.vpc.connector = `projects/${bdEndpoint.project}/locations/${region}/connectors/${bdEndpoint.vpc.connector}`;
        }
        **bkEndpoint.vpc = { connector: bdEndpoint.vpc.connector };**
        proto.copyIfPresent(bkEndpoint.vpc, bdEndpoint.vpc, "egressSettings");
      } ...
```

colerogers (commented):
small bug, please fix and rerun the tests

colerogers:
> Thanks @joehan.
> 
> For time savings this is the failed test
> 
> ```diff
> 1) addResourcesToBuild
>        should expand vpc connector in w/ shorthand form:
> 
>       AssertionError: expected { Object (requiredAPIs, endpoints, ...) } to deeply equal { Object (requiredAPIs, endpoints, ...) }
>       + expected - actual
> 
>                "project": "project"
>                "region": "us-central1"
>                "runtime": "nodejs16"
>                "vpc": {
>       -          "connector": "hello-vpc"
>       +          "connector": "projects/project/locations/us-central1/connectors/hello-vpc"
>                }
>              }
>            }
>          }
> ```
> 
> I think this test is wrong and should be updated. The values are equivalent, as long as `hello-vpc` is in that project and region, and the long form could be seen as more correct™ as it allows for the shared vpc use case? Thoughts?

If you change your code to save to `bdEndpoint` and not `bkEndpoint`, the test should pass

steveoh:
@colerogers does there need to be another test case written for this use case?

src/deploy/functions/build.ts
steveoh: thanks!

steveoh:
@joehan can we approve the test runs?

steveoh:
ughhh, the codespace wasn't set up with prettier...

colerogers:
@steveoh please run `npm run mocha` to test everything out locally

src/deploy/functions/build.ts
```diff
@@ -494,10 +494,12 @@ export function toBackend(
         nullsafeVisitor((cpu) => (cpu === "gcf_gen1" ? cpu : r.resolveInt(cpu)))
       );
       if (bdEndpoint.vpc) {
+        bdEndpoint.vpc = { connector: params.resolveString(bdEndpoint.vpc.connector, paramValues) };
```
colerogers, line 497: You are writing over the entire `vpc` object which will remove the field `egressSettings` if it exists

colerogers:
```
bdEndpoint.vpc.connector = params.resolveString(bdEndpoint.vpc.connector, paramValues);
```

steveoh: ok committed that fix.

steveoh:
> @steveoh please run `npm run mocha` to test everything out locally

ok, i never checked out the repo. I debugged in my node_modules and was hoping your ci would do the trick but it is a pain with the workflows requiring a review. I'll check it locally and test it out.

steveoh:
ok tests passed locally @joehan

steveoh:
ok tests passed locally. Can I get a CI review approval to cross the finish line @joehan?

dwight cracking nuckles

joehan (approved):

joehan:
LGTM - will get this out in the next release later this week. Thanks for the contribution @steveoh!

PR #6334, opened 2023-09-07
Pull settings (and generic listeners) out of workflow

tjlav5:


firebase-vscode/src/core/index.ts
```diff
@@ -0,0 +1,38 @@
+import vscode, { Disposable } from "vscode";
+import { ExtensionBrokerImpl } from "../extension-broker";
+import { registerConfig } from "./config";
+import { registerEmulators } from "./emulators";
```
hlshen, line 4: should cli be in core as well?

firebase-vscode/src/workflow.ts
```diff
@@ -299,9 +259,9 @@ export async function setupWorkflow(
     currentFramework = undefined;
     // Note: discover() takes a few seconds. No need to block users that don't
     // have frameworks support enabled.
+    const { useFrameworks } = getSettings();
     if (useFrameworks) {
-      currentFramework =
-        useFrameworks && (await discover(currentOptions.cwd, false));
```
hlshen, line 304: Not sure if you know but what was the original use of this?

hlshen (approved):

firebase-vscode/src/core/index.ts
tjlav5: arguably in utils, but if we do move this, it'll be a later CL, trying to keep these changes small since we don't have test coverage

firebase-vscode/src/workflow.ts
tjlav5: idk - but it's in the if-statement already, so there's really no need for it

PR #6337, opened 2023-09-08
Don't prompt for OOB change confirmation in noninteractive mode

### LINKED ISSUE #6321: Deploying Firebase extensions not considering `--force` and `--non-interactive`

**firebase-tools:** 12.4.7

**Platform:** Ubuntu 22.04 ([Github Actions Image](https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20230821.1))

### [REQUIRED] Test case

Deploying from continuous integration server (here using Github Actions) may lead to the following warning message when updating extensions:

```
The following instances may have been changed in the Firebase console or by another machine since the last deploy from this machine.
	storage-resize-images
	fs-bq-agreements
        ...
```
We ALWAYS deploy (extensions included) from Github Actions and NEVER from the console. However, we never commit back the new extensions etags after deployment because we don't want de deployment pipeline to change the release code. This is most probably what triggers the warning in the first place.

Consequently we deploy with the following command: `firebase-tools --non-interactive deploy --force`, so that we always force deploy what's in the code base which is our only source of truth.

Unfortunately, the previous command cannot complete cause of the following confirmation prompt, which obviously cannot be answered on a CI server:
```
If you proceed with this deployment, those changes will be overwritten. To avoid this, run `firebase ext:export` to sync these changes to your local extensions manifest.
? Do you wish to continue deploying these extension instances? (y/N) 
```

1. Deploy a least one Firebase extension.
2. Commit your up to date `.firebaserc` after deployment.
3. Update your firebase extension so a new deployment is needed (e.g increment extension version)
4. Deploy your new extension version
5. Reset your `.firebaserc` to discard the last etags changes (this should give you the etags of step 2).
6. Deploy one last time using `--non-interactive` and `--force`.

=> You are still prompted for override confirmation.

The CLI should force deploy, overriding anyway without entering interactive mode.

The CLI enters the interactive mode and ask wether to continue with the deployment.

joehan:
### Description
Pass `--noninteractive` and `--force` to prompts for out of band changes to extensions instances. Fixes #6321 
### Scenarios Tested
See https://github.com/aalej/issues-6321 - this no longer hangs in noninteractive mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,3 @@
 - Fixed an issue where builds from https://firebase.tools could not run commands that spawn `npm`. (#6132)
-- Fixed an issue where `--noninteractive` and `--force` were not respected in some extension deploys. (#6321)
+- Fixed an issue where `--non-interactive` and `--force` were not respected in some extension deploys. (#6321)
 - Fixed the regex in extensions changelog parser to lazy match the version prefix to allow matching higher versions (#6326)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,3 @@
 - Fixed an issue where builds from https://firebase.tools could not run commands that spawn `npm`. (#6132)
+- Fixed an issue where `--noninteractive` and `--force` were not respected in some extension deploys. (#6321)
 - Fixed the regex in extensions changelog parser to lazy match the version prefix to allow matching higher versions (#6326)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,2 +1,3 @@
 - Fixed an issue where builds from https://firebase.tools could not run commands that spawn `npm`. (#6132)
+- Fixed an issue where `--noninteractive` and `--force` were not respected in some extension deploys. (#6321)
```
aalej, line 2: nit: `--noninteractive` -> `--non-interactive`

pavelgj (approved):

aalej (approved):

PR #6342, opened 2023-09-08
Move quickstart functionality to disposable

tjlav5:
* and out of workflow


hlshen (approved):
I guess this one is core?? feels the least core out of the other ones but not sure where else this would go

PR #6348, opened 2023-09-09
Turndown workflow file

tjlav5:
Updated project-picking at the same time (could only have one picker at a time)


firebase-vscode/src/core/project.ts
```diff
@@ -67,6 +124,6 @@ async function promptUserForProject(projects: FirebaseProjectMetadata[]) {
     description: p.displayName,
   }));
 
-  const projectId = await vscode.window.showQuickPick(items);
-  return projectId;
+  const item = await vscode.window.showQuickPick(items);
```
hlshen, line 127: Why this change? Can we add the return type to eht func

firebase-vscode/src/core/project.ts
```diff
@@ -42,17 +59,57 @@ export function registerProject(broker: ExtensionBrokerImpl): Disposable {
...
+        });
+
+        if (projectId) {
+          currentProjectId.value = projectId;
+        }
+      } catch (e) {
+        pluginLogger.error(e);
+      }
+    } else if (isServiceAccount.value) {
+      return;
```
hlshen, line 103: Can this check be moved up? To avoid unnecessary async call?

hlshen (approved):

firebase-vscode/src/core/project.ts
tjlav5: Which async call? I don't want to make too many logic changes since I'm not that familiar w/ monospace.

firebase-vscode/src/core/project.ts
tjlav5: I'm using label/description now, instead of just a string.

PR #6357, opened 2023-09-12
Added a check to see if export target folder is empty

### LINKED ISSUE #6313: emulators:export doesn't check if the target folder is empty...

**firebase-tools:** 12.4.6

**Platform:** Fedora

### [REQUIRED] Test case

If you run `firebase emulators:export .`, your working directory will be rimraf'd.

Run `firebase emulators:export .`

The command makes sure the path is empty and warns of deleting everything there. I see there's a `--force` flag, which I guess should prevent this check, but I didn't use it, yet still got this insane result.

Everything just disappears, all the code you haven't pushed yet gets completely obliterated...

aalej:
### Description

Fixes #6313 

Running `firebase emulators:export .` will delete everything in the current directory

Steps to reproduce:
1. Run `firebase init firestore --project <project_id>`
   - Setup emulator
2. Run `firebase emulators:start`
3. In a new terminal, under the same directory, run `firebase emulators:export .`
   - Entire directory is deleted

### Scenarios Tested

1. Run `firebase init firestore --project <project_id>`
   - Setup emulator
2. Run `firebase emulators:start`
3. In a new terminal, under the same directory, run `firebase emulators:export .`
4. Receive prompt:
```
The directory /Users/<path>/test is not empty. Existing files in this directory will be overwritten. Do you
 want to continue? (y/N)
```
5. Test passing `--force` flag to override checks
   - Entire directory is deleted

### Sample Commands

`firebase emulators:export`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,3 +2,4 @@
   - Removed nano precision in timestamp used in Firestore emulator (#5893)
   - Fixed a bug where query behaves differently from production.
 - Fixed an issue where very long command outputs would be cut off. (#3286)
+- Fixed an issue where `emulators:export` does not check if the target folder is empty. (#6313)
--- a/src/emulator/controller.ts
+++ b/src/emulator/controller.ts
@@ -961,7 +961,8 @@ export async function exportEmulatorData(exportPath: string, options: any, initi
 
   // Check if there is already an export there and prompt the user about deleting it
   const existingMetadata = HubExport.readMetadata(exportAbsPath);
-  if (existingMetadata && !(options.force || options.exportOnExit)) {
+  const isExportDirEmpty = fs.readdirSync(exportAbsPath).length === 0;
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- Fixed an issue where `emulators:export` does not check if the target folder is empty. (#6313)
```
```suggestion
- Fixed an issue where `emulators:export` did not check if the target folder is empty. (#6313)
```

src/emulator/controller.ts
```diff
@@ -971,7 +972,7 @@ export async function exportEmulatorData(exportPath: string, options: any, initi
 
     const prompt = await promptOnce({
```
joehan, line 973: Optional - consider using [confirm](https://github.com/firebase/firebase-tools/blob/master/src/prompt.ts#L104) here instead - it already has some of this logic for handling force/non-interactive baked in

joehan (approved):
LGTM with one optional improvement

PR #6358, opened 2023-09-12
Revert "Fix output buffer pipe issue"

joehan:
Reverts firebase/firebase-tools#5310 - I missed that this also removed process.exit(), which breaks the emulator suite in a number of ways (and is why the triggers-end-to-end integration test is currently failing)


maneesht (approved):
Approved

hlshen (approved):

PR #6359, opened 2023-09-12
Ensure long outputs are written to stdout before exiting process

### LINKED ISSUE #3286: firebase cli --json output is truncated to 64000 bytes when piped to another command

joehan:
### Description
A safer take on https://github.com/firebase/firebase-tools/pull/5310 - ensures that we're done writing to stdout before we exit. Fixes #3286

This is necessary due to some slightly unusual behavior of console.log/streams in Node. Writing to stdout (aka calling console.log) can usually be thought of as synchronous action - however, if you write a large enough buffer, it instead behaves asynchronously - https://nodejs.org/docs/v18.17.1/api/stream.html#:~:text=is%20not%20safe.-,Buffering,-%23 lays this out in more detail.


src/command.ts
```diff
@@ -223,19 +226,25 @@ export class Command {
             );
           }
           process.exit();
+        })
+        .then(async (result) => {
```
maneesht, line 230: Any reason why this is here?

src/command.ts
```diff
@@ -371,7 +380,7 @@ export class Command {
    * @return an async function that executes the command.
    */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  runner(): (...a: any[]) => Promise<void> {
+  runner(): (...a: any[]) => Promise<any> {
```
maneesht, line 383: I'm assuming this change is unrelated to the changes above?

src/command.ts
joehan: Derp, no - remnant of a previous approach I tried that i didn't clean up/

src/command.ts
joehan: This is actually the 'result' that we're printing out in the main part of this change. 

This was incorrectly typed before - runner() is actually returning whatever is returned by the command's action function.

scripts/clean-install.sh
```diff
@@ -9,6 +9,7 @@ function cleanup() {
 
 trap cleanup EXIT
 
+rm -rf ./clean || true
```
joehan, line 12: This is sort of an unrelated improvement - triggers-end-to-end-test was  sometimes failing because `./clean` already existed when we went to write it, and it was annoying me while I was debugging this.

src/command.ts
maneesht: Got it. Thanks!

maneesht (approved):
LGTM

Issue #6381, opened 2023-09-20
Failed to list Firebase projects: Unable to parse JSON: SyntaxError: Unexpected token < in JSON at position 0


chris-arta:
### Environment info

**firebase-tools: 12.5.4**

**Platform: Ubuntu 20.04.6 LTS (hosted on GCE)**

### Test case

List Firebase projects using firebase-tools

### Steps to reproduce

`npm install -g firebase-tools`
`firebase login --no-localhost`
`firebase projects:list`

### Expected behavior

Display a list of the current projects associated with the current logged in user, including: Project Display Name, Project ID, Project Number and Resource Location ID

### [REQUIRED] Actual behavior

`firebase projects:list --debug`
```
[2023-09-20T15:43:02.525Z] > command requires scopes: ["email","openid","https://www.googleapis.com/auth/cloudplatformprojects.readonly","https://www.googleapis.com/auth/firebase","https://www.googleapis.com/auth/cloud-platform"]
[2023-09-20T15:43:02.526Z] > authorizing via signed-in user (chris@artafinance.com)
⠋ Preparing the list of your Firebase projects[2023-09-20T15:43:02.532Z] >>> [apiv2][query] GET https://firebase.googleapis.com/v1beta1/projects pageSize=1000
⠙ Preparing the list of your Firebase projects[2023-09-20T15:43:02.696Z] <<< [apiv2][status] GET https://firebase.googleapis.com/v1beta1/projects 403
[2023-09-20T15:43:02.697Z] <<< [apiv2][body] GET https://firebase.googleapis.com/v1beta1/projects [omitted]
[2023-09-20T15:43:02.697Z] Unable to parse JSON: SyntaxError: Unexpected token < in JSON at position 0
✖ Preparing the list of your Firebase projects
[2023-09-20T15:43:02.714Z] FirebaseError: Unable to parse JSON: SyntaxError: Unexpected token < in JSON at position 0
    at RetryOperation._fn (/home/chris/.nvm/versions/node/v18.12.1/lib/node_modules/firebase-tools/lib/apiv2.js:257:39)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

Error: Failed to list Firebase projects. See firebase-debug.log for more info.

Having trouble? Try again or contact support with contents of firebase-debug.log
```

### More info
The reported behavior started recently (not sure about the exact date, but most likely in the last couple of weeks)
Also impacts other firebase-tools commands, like `firebase use <project-name>`
[...]

PR #6387, opened 2023-09-21
Better logging when nonJSON errors are returned

joehan:
### Description
Better logging when non-JSON errors are returned on calls expecting JSON responses.

In some cases (usually related to auth or restricted regions), API calls are rejected by GFE and an HtML response is returned instead of the expected JSON (ie #6381). In these cases, let's force the full body to be printed in debug logs so that we can help users more easily.


fredzqm (approved):

fredzqm (changes requested):
nvm, this is from one year ago.

PR #6409, opened 2023-09-27
Support non-ASCII file names in `appdistribution:distribute`

tonybaroneee:
### Description

Adds support for non-ASCII characters in app binary file names that are sent over the wire using the `X-Goog-Upload-File-Name` HTTP header.

Leverages `encodeURIComponent` to encode the file name.

### DIFF
--- a/src/appdistribution/client.ts
+++ b/src/appdistribution/client.ts
@@ -92,7 +92,7 @@ export class AppDistributionClient {
       method: "POST",
       path: `/upload/v1/${appName}/releases:upload`,
       headers: {
-        "X-Goog-Upload-File-Name": distribution.getFileName(),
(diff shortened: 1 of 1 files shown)


src/appdistribution/client.ts
```diff
@@ -92,7 +92,7 @@ export class AppDistributionClient {
       method: "POST",
       path: `/upload/v1/${appName}/releases:upload`,
       headers: {
-        "X-Goog-Upload-File-Name": distribution.getFileName(),
+        "X-Goog-Upload-File-Name": encodeURIComponent(distribution.getFileName()),
```
kaibolay, line 95: Does this header use URI encoding (i.e. how is it decoded on the receiving end)?

tonybaroneee: Yep, we already decode this on the server-side. This is actually the same technique that the Firebase console uses today.

PR #6411, opened 2023-09-28
Fix issue #5561 emulators not starting with vite

sotovviktor:
### Description

Fixes https://github.com/firebase/firebase-tools/issues/5561

stdout data on widnows has ANSI characters in the URL preventing resolution of the promise. This PR fixes it.

### DIFF
--- a/src/frameworks/vite/index.ts
+++ b/src/frameworks/vite/index.ts
@@ -3,6 +3,7 @@ import { spawn } from "cross-spawn";
 import { existsSync } from "fs";
 import { copy, pathExists } from "fs-extra";
 import { join } from "path";
+const stripAnsi = require("strip-ansi");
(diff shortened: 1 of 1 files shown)


src/frameworks/vite/index.ts
```diff
@@ -3,6 +3,7 @@ import { spawn } from "cross-spawn";
 import { existsSync } from "fs";
 import { copy, pathExists } from "fs-extra";
 import { join } from "path";
+const stripAnsi = require("strip-ansi");
```
jamesdaniels, line 6: can we change to an import? e.g, `import * as stripAnsi from "strip-ansi"`

jamesdaniels (changes requested):
Add changelog entry and see about converting to an import statement and g2g

jamesdaniels:
I addressed feedback in #6588

PR #6420, opened 2023-09-29
Use correct bucket name field when reading DefaultBucket for storage

abhis3:
### Description

Use correct bucket name field when reading in DefaultBucket proto.

Here is the Default Bucket proto format: https://firebase.google.com/docs/reference/rest/storage/rest/v1alpha/projects.defaultBucket#DefaultBucket

This will return a string with a format "projects/X/buckets/X", which we then need to strip to the last part of the path which is just the bucket name to match the API behavior that already existed.

### DIFF
--- a/src/gcp/storage.ts
+++ b/src/gcp/storage.ts
@@ -133,6 +133,14 @@ interface ListBucketsResponse {
   ];
 }
 
+interface GetDefaultBucketResponse {
+  name: string;
+  location: string;
+  bucket: {
+    name: string;
+  };
+}
+
 /** Response type for obtaining the storage service agent */
 interface StorageServiceAccountResponse {
   email_address: string;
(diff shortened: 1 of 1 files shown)


src/gcp/storage.ts
```diff
@@ -143,19 +151,19 @@ export async function getDefaultBucket(projectId: string): Promise<string> {
...
       throw new FirebaseError(
         "Your project is being set up. Please wait a minute before deploying again."
       );
     }
-    return response.body.name;
+    return response.body.bucket.name.split("/").pop()!;
   } catch (err: any) {
     logger.info(
-      "\n\nThere was an issue deploying your Storage rules. Verify that your project has a Google App Engine instance setup at https://console.cloud.google.com/appengine and try again. If this issue persists, please contact support."
+      "\n\nThere was an issue deploying your functions. Verify that your project has a Google App Engine instance setup at https://console.cloud.google.com/appengine and try again. If this issue persists, please contact support."
```
joehan, line 166: Quick sanity check - is this still accurate/ relevant? If someone doesn't have a default storage bucket set up, do they still need to go create an App engine instance, or is there something else they can do?

joehan (approved):
LGTM with one piece of copy to fix

src/gcp/storage.ts
abhis3: As of now it's the correct path of action. We also have a new createDefaultBucket endpoint that we can ask clients to call which right now under the hood among other things will create an AppEngine instance. https://firebase.google.com/docs/reference/rest/storage/rest/v1alpha/projects.defaultBucket/create

I would say we change the copy once the endpoint hits v1beta as opposed to v1alpha

Issue #6364, opened 2023-09-15
Cloud Function deploy fails for Firestore-triggered function


dominoFire:
### [REQUIRED] Environment info

**firebase-tools:** 12.4.7

**Platform:** Windows

### [REQUIRED] Test case

When I try to deploy a triggered Python, 2nd-gen cloud function, it fails with a database not found error.

### [REQUIRED] Steps to reproduce

1. Create a function in your main.py:

```python
@firestore_fn.on_document_created(document='a/{aId}/b/{bId}')
def do_magic(event: Event[DocumentSnapshot]) -> None:
    """
    Do magic
    """


    return None
```

2. Run `firebase deploy` on the function.

### [REQUIRED] Expected behavior

Deploys succeeds.

### [REQUIRED] Actual behavior

Deploy fails with the following message:

[firebase-debug.log](https://github.com/firebase/firebase-tools/files/12629198/firebase-debug.log)

```

=== Deploying to 'propio-app'...

i  deploying functions
i  functions: preparing codebase default for deployment
i  functions: ensuring required API cloudfunctions.googleapis.com is enabled...
i  functions: ensuring required API cloudbuild.googleapis.com is enabled...
i  artifactregistry: ensuring required API artifactregistry.googleapis.com is enabled...
+  functions: required API cloudfunctions.googleapis.com is enabled
+  functions: required API cloudbuild.googleapis.com is enabled
+  artifactregistry: required API artifactregistry.googleapis.com is enabled
i  functions: Loading and analyzing source code for codebase default to determine what to deploy
 * Serving Flask app 'serving'
 * Debug mode: off

WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.
 * Running on http://127.0.0.1:8081
Press CTRL+C to quit

127.0.0.1 - - [15/Sep/2023 13:13:27] "GET /__/functions.yaml HTTP/1.1" 200 -

2023-09-15 13:13:27,719 - werkzeug - INFO - 127.0.0.1 - - [15/Sep/2023 13:13:27] "GET /__/functions.yaml HTTP/1.1" 200 -

127.0.0.1 - - [15/Sep/2023 13:13:27] "GET /__/quitquitquit HTTP/1.1" 200 -

2023-09-15 13:13:27,728 - werkzeug - INFO - 127.0.0.1 - - [15/Sep/2023 13:13:27] "GET /__/quitquitquit HTTP/1.1" 200 -

[...]

PR #6442, opened 2023-10-11
Ignore Firestore emulator host on functions deploy

blidd-google:
Some users were running into an error at functions deploy time due to their `FIRESTORE_EMULATOR_HOST` environment variable being set. The deploy logic should not be reading the emulator host environment variable; this PR fixes the issue by only checking the Firestore origin host URL on deploys.

Fixes https://github.com/firebase/firebase-tools/issues/6364.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,2 @@
+- Ignore Firestore emulator host on functions deploy. (#6442)
 - Added support for enabling, disabling, and displaying Point In Time Recovery enablement state on Firestore databases (#6388)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1 +1,2 @@
+- Ignore Firestore emulator host on functions deploy. (#6442)
```
```suggestion
- Ignore `FIRESTORE_EMULATOR_HOST` environment variable on functions deploy. (#6442)
```

joehan (approved):

PR #6447, opened 2023-10-13
Update min node version to 18, add 20 to supported versions & CI

joehan:
### Description
Upgrades tooling to node 20, removes explicit support for node 16


bkendall (changes requested):
You probably want to include the Dockerfiles in this as well

joehan:
Merging this in with framework tests not passing - we'll come back for those later

Issue #5873, opened 2023-05-22
FirebaseError: Failed to find location of Firebase Functions SDK.


sharadregoti:
When trying to start the Cloud Functions emulator that uses Python, an error is being raised. 

### [REQUIRED] Environment info

**firebase-tools:**
firebase-tools: 12.1.0

**Platform:**
Platform: linux (windows wsl2)

Node: v16.17.0

Python: v3.11.3

### [REQUIRED] Steps to reproduce
1. Follow this getting started guide of firebase function for python
https://firebase.google.com/docs/functions/get-started?gen=2nd#about_this_tutorial

### [REQUIRED] Expected behavior
No error should have occured & I should have gotten a URL for testing function

### [REQUIRED] Actual behavior
Whenever i run `firebase emulators:start`

i  emulators: Starting emulators: functions, firestore, storage
⚠  functions: The following emulators are not running, calls to these services from the Functions emulator will affect production: auth, database, hosting, pubsub
i  firestore: Firestore Emulator logging to firestore-debug.log
✔  firestore: Firestore Emulator UI websocket is running on 9150.
i  ui: Emulator UI logging to ui-debug.log
i  functions: Watching "/home/sharad/personal/salarygpt/functions" for Cloud Functions...
⬢  functions: Failed to load function definition from source: FirebaseError: Failed to find location of Firebase Functions SDK. Did you forget to run '. "/home/sharad/personal/salarygpt/functions/venv/bin/activate" && python3.11 -m pip install -r requirements.txt'?
i  functions: Watching "/home/sharad/personal/salarygpt/js-functions" for Cloud Functions...
⚠  functions: Your requested "node" version "18" doesn't match your global version "16". Using node@16 from host.
✔  functions: Loaded functions definitions from source: addmessage, makeuppercase.
✔  functions[us-central1-addmessage]: http function initialized (http://127.0.0.1:5001/salarygpt-c0c20/us-central1/addmessage).
✔  functions[us-central1-makeuppercase]: firestore function initialized.

┌─────────────────────────────────────────────────────────────┐
│ ✔  All emulators ready! It is now safe to connect your app. │
│ i  View Emulator UI at http://127.0.0.1:4000/               │
└─────────────────────────────────────────────────────────────┘

┌───────────┬────────────────┬─────────────────────────────────┐
│ Emulator  │ Host:Port      │ View in Emulator UI             │
├───────────┼────────────────┼─────────────────────────────────┤
[...]

PR #6455, opened 2023-10-18
Add python 3.12 to firebase functions runtimes.

nitinl:
### Description
Firebase functions now support the Python 3.12 runtime. This commit updates cli-tools to work for python 3.12 as well. This should also resolve https://github.com/firebase/firebase-tools/issues/5873 for users in a Python 3.12 environment.


dcgudeman:
what's the status of this issue?

Issue #6453, opened 2023-10-16
Firebase function redeploy error after re-setting secrets


mvarchdev:
### [REQUIRED] Environment info

**firebase-tools: 12.6.2**

**Platform: macOS (tried also Ubuntu)**

### [REQUIRED] Test case
Create firebase project, enable firebase functions, create firebase functions source code that in some way uses secrets. Set secrets, deploy functions.

### [REQUIRED] Steps to reproduce

1. Set firebase functions secrets with `firebase functions:secrets:set`
2. Deploy functions that are using that secrets
3. Set secrets again, when asked if needed to redeploy and destroy stale version, type yes

### [REQUIRED] Expected behavior

It should set secrets, then deploy functions and then when secrets are set again, it should redeploy correctly.

### [REQUIRED] Actual behavior

First set of secrets is working ok, then functions deployment its ok, but when you will try to re-set secrets, it will ask you if you want to redeploy functions, when you will reply yes, It will create this error:
```
marosvarchola@Maross-MacBook-Pro glassfinder-firebase-functions % firebase functions:secrets:set ALGOLIA_ID        
? Enter a value for ALGOLIA_ID [hidden]
✔  Created a new secret version projects/129938725887/secrets/ALGOLIA_ID/versions/7
i  2 functions are using stale version of secret ALGOLIA_ID:
        onGlassesUpdateIndexAlgolia(europe-west3)
        rebuildGlassesIndex(europe-west3)
? Do you want to re-deploy the functions and destroy the stale version of secret ALGOLIA_ID? Yes
i  Updating function onGlassesUpdateIndexAlgolia(europe-west3)...
i  Updating function rebuildGlassesIndex(europe-west3)...
⚠  functions: HTTP Error: 400, Validation failed for trigger projects/glass-finder-c5b53/locations/europe-west3/triggers/onglassesupdateindexalgolia-785296: The request was invalid: invalid argument: event type google.cloud.firestore.document.v1.written not supported: attribute topic not found within event type
⚠  functions:  failed to update function projects/glass-finder-c5b53/locations/europe-west3/functions/onGlassesUpdateIndexAlgolia

Error: Failed to update function projects/glass-finder-c5b53/locations/europe-west3/functions/onGlassesUpdateIndexAlgolia
```

Logs:
```
marosvarchola@Maross-MacBook-Pro glassfinder-firebase-functions % firebase --debug functions:secrets:set ALGOLIA_ID
[...]

PR #6456, opened 2023-10-19
Fix bug where 2nd Gen firestore functions were mistakenly parsed as pubsub function

taeold:
Firestore trigger also includes pubsubTopic since they use pubsub as transport. This triggered a bug where 2nd Gen Firestore bugs were being parsed as pubsub function which manifests in bugs like https://github.com/firebase/firebase-tools/issues/6453.

Fixes https://github.com/firebase/firebase-tools/issues/6453


aalej (approved):
LGTM! 

Tried out re-deploying 2nd gen and 1st gen Firestore Functions after changing the secrets being used.

PR #6471, opened 2023-10-23
GetStorageDefaultBucket through Storage API

### LINKED ISSUE #6467: [v13.0.0 Blocker] GetStorageDefaultBucket through Storage API
We currently get the Default Storage Bucket by calling a FireData API that will be deprecated soon. We need to switch to using the Storage API for this. This will require some new permissions (included in Storage Bucket Viewer role) - we should also document these and ensure that when they are missing we error cleanly.

abhis3:
GetStorageDefaultBucket through Storage API.

This change is breaking as it introduces a new permission, it rolls in changes from https://github.com/firebase/firebase-tools/commit/77b3690ff3cd8f283782ab58513b71243080027e and https://github.com/firebase/firebase-tools/commit/6ff08fc7101f995b5a695c72f169a7d19c80c23c into one PR.

Fixes #6467


joehan (approved):
Thanks!

tcf909:
Hello, you pushed this to the public before the actual ACL was available to be added / managed in the cloud console:

image

PR #6480, opened 2023-10-26
Angular v17 support 

jamesdaniels:
Adding support for [Angular's new build system](https://angular.dev/tools/cli/esbuild) based off esbuild and vite. This promises to dramatically simplify the integration, we'll work to depreciate the old builders once this is well adopted.

* Updated the documentation to reflect Angular v17
* Renamed the existing builders to LEGACY_* so it's clearer the code we can deprecate later
* For application builder:
  * SSR is an option, we don't have to infer it from another target
  * There is always a "browser" sub-folder, making outDir handling easier
  * Building the browser, server, and pre-rendering are all in a single command! wohoo!
  * Server target is an MJS now, switching the function template to use import
* For the es-build builder:
  * While it's not a legacy builder, we have to treat it the same in terms of path (no "browser" folder)
* Cleaned up some of the mess in getContext & adding more guardrails. It should be much clearer now the precedence assuming build targets:
  1. ng-deploy
  2. prerender
  3. build / server

### DIFF
--- a/src/frameworks/angular/index.ts
+++ b/src/frameworks/angular/index.ts
@@ -36,18 +36,12 @@ export const docsUrl = "https://firebase.google.com/docs/hosting/frameworks/angu
 
 const DEFAULT_BUILD_SCRIPT = ["ng build"];
 
-/**
- *
- */
 export async function discover(dir: string): Promise<Discovery | undefined> {
   if (!(await pathExists(join(dir, "package.json")))) return;
   if (!(await pathExists(join(dir, "angular.json")))) return;
   return { mayWantBackend: true, publicDirectory: join(dir, "src", "assets") };
 }
 
-/**
- *
- */
 export async function init(setup: any, config: any) {
   execSync(
     `npx --yes -p @angular/cli@latest ng new ${setup.projectId} --directory ${setup.hosting.source} --skip-git`,
@@ -70,9 +64,6 @@ export async function init(setup: any, config: any) {
   }
 }
 
-/**
- *
- */
 export async function build(dir: string, configuration: string): Promise<BuildResult> {
   const {
     targets,
@@ -106,9 +97,6 @@ export async function build(dir: string, configuration: string): Promise<BuildRe
   return { wantsBackend, i18n, rewrites, baseUrl };
 }
 
-/**
- *
- */
 export async function getDevModeHandle(dir: string, configuration: string) {
   const { targetStringFromTarget } = relativeRequire(dir, "@angular-devkit/architect");
   const { serveTarget } = await getContext(dir, configuration);
@@ -133,9 +121,6 @@ export async function getDevModeHandle(dir: string, configuration: string) {
   return simpleProxy(await host);
 }
 
-/**
- *
- */
 export async function ɵcodegenPublicDirectory(
   sourceDir: string,
   destDir: string,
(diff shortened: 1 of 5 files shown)


src/frameworks/angular/utils.ts
```diff
@@ -268,67 +294,90 @@ export async function getContext(dir: string, targetOrConfiguration?: string) {
...
+    throw new FirebaseError(`No build target on ${project}`);
+  }
+  const buildOrBrowserTargetOptions = await architectHost.getOptionsForTarget(buildOrBrowserTarget);
+  if (!buildOrBrowserTargetOptions) {
+    const targetString = targetStringFromTarget(buildOrBrowserTarget);
+    throw new FirebaseError(`Couldn't find options for ${targetString}.`);
   }
 
-  const baseHref = browserTargetOptions.baseHref || "/";
+  const baseHref = buildOrBrowserTarget.baseHref || "/";
```
jamesdaniels, line 307: whoops

src/frameworks/angular/utils.ts
```diff
@@ -268,67 +288,81 @@ export async function getContext(dir: string, targetOrConfiguration?: string) {
...
   if (typeof baseHref !== "string") {
-    throw new FirebaseError(
-      `baseHref on ${targetStringFromTarget(browserTarget)} was not a string`
-    );
+    const targetString = targetStringFromTarget(buildOrBrowserTarget);
+    throw new FirebaseError(`baseHref on ${targetString} was not a string`);
   }
 
+  const buildTargetOptions = buildTarget && (await architectHost.getOptionsForTarget(buildTarget));
+  const ssr = !!buildTargetOptions?.ssr || !!serverTarget;
```
jamesdaniels, line 308: should not fall back to serverTarget if using application builder

src/frameworks/docs/angular.md
```diff
@@ -50,42 +50,14 @@ firebase deploy
 ## Pre-render dynamic content
 
 To prerender dynamic content in Angular, you need to set up Angular Universal.
```
9kubczas4, line 52:
```suggestion
To prerender dynamic content in Angular, you need to set up Angular SSR.
```

src/frameworks/angular/index.ts
```diff
@@ -67,10 +67,10 @@ export async function init(setup: any, config: any) {
 export async function build(dir: string, configuration: string): Promise<BuildResult> {
```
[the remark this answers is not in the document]
jamesdaniels: Good point, I should rework the bootstrap. It should be part of Angular's create routine now

egilmorez (approved):
Updates to docs/angular.md look good, thanks for getting that!!

src/frameworks/angular/index.ts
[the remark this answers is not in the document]
alan-agius4: Since you are using @angular/cli@latest, `@nguniversal/express-engine` should not be used as this is only supported for version 16 and before.

alan-agius4 (approved):
LGTM

ovitrif:
When will we see this in prod?

alexastrum:
> When will we see this in prod?

In the next minor release of Firebase CLI tools.

jnizet:
I still can't see the documentation introduced in this commit in https://firebase.google.com/docs/hosting/frameworks/frameworks-overview. It still mentions Angular Universal. Is it normal?

PR #6486, opened 2023-11-01
Svn main 001 rem

svnsairam:
### Description

Changed stack -> backend
I will change stack to backend in rest of the places in other PR's


src/commands/frameworks-stacks-create.ts
```diff
@@ -4,8 +4,8 @@ import { needProjectId } from "../projectUtils";
 import requireInteractive from "../requireInteractive";
 import { doSetup } from "../init/features/frameworks";
 
-export const command = new Command("stacks:create")
-  .description("Create a stack in a Firebase project")
+export const command = new Command("backends:create")
```
abhis3, line 7: Small semantic nit, but should this be `backend:create` as opposed to `backends:create`? If we're creating an instance of a backend it seems to me the more logical choice will be the singular version of the noun as opposed to the plural form here. (same comment throughout this PR)

abhis3 (approved):

PR #6488, opened 2023-11-01
Changed stack to backend

svnsairam:
### Description

Change stack to backend in GET Stack and LIST Stack methods.
Show the stack details in Table format.

I will change stack to backend in rest of the places in other PR's


src/commands/frameworks-stacks-list.ts
```diff
@@ -4,27 +4,39 @@ import { needProjectId } from "../projectUtils";
...
 
-    let backends;
+    let backendsList;
     try {
-      backends = await gcp.listBackend(projectId, location);
-      /**
-       * TODO print this in a prettier way.
-       */
-      logger.info(backends);
+      backendsList = await gcp.listBackend(projectId, location);
```
abhis3: This is kind of weird imo that GCP exposes it like this without the plural form, which goes against AIP best practices https://google.aip.dev/132, specifically "The remainder of the RPC name should be the plural form of the resource being listed."

Nothing we can do I guess just wanted to call this out

abhis3 (approved):

src/commands/index.ts
```diff
@@ -154,9 +154,9 @@ export function load(client: any): any {
   if (experiments.isEnabled("internalframeworks")) {
     client.frameworks = {};
     client.frameworks.stacks = {};
-    client.frameworks.stacks.list = loadCommand("frameworks-stacks-list");
+    client.frameworks.stacks.list = loadCommand("frameworks-backends-list");
```
abhis3, line 157: I assume we're not responsible for changing the client.frameworks.stacks.list? Or at least not in this PR?

svnsairam: I deleted the file "frameworks-stacks-list.ts" and created "frameworks-backends-list.ts" file. So, we should replace the values here.

abhis3 (approved):

Issue #6485, opened 2023-11-01
Reusage of --force flag for function deployment with minimum instances (minimum bill)


Buckwich:
### [REQUIRED] Environment info

**firebase-tools:** 12.4.4 (should be the same on latest version)

**Platform:** Docker on Ubuntu on Github Action Runner

### [REQUIRED] Test case

Any normal firebase project with some functions. Optionally it is using a CI/CD (non-interactive) to deploy. Adjusting the minInstances to increase the minimum bill of the project, in our case 1 is enough.

```
.runWith({
  minInstances:1
})
```
### [REQUIRED] Steps to reproduce

Deploy the adjusted function optimally via a non interactive terminal, ie. github action. 

### [REQUIRED] Expected behavior

The function should be deployed with minInstance=1. 

### [REQUIRED] Actual behavior

Error message: `Pass the --force option to deploy functions that increase the minimum bill` and deployment is canceled. 

The problem is that we can't use the `--force` flag as this would also auto delete functions. From the help section this behaviour is also not clear. There the force flag is only documented to delete functions:
```
 -f, --force              delete Cloud Functions missing from the current working directory without confirmation
```

### Solution Idea

The `--force` flag should not be reused as a helper for other stuff than the deletion. Probably a separate flag for that would suffice, eg. `--allow-bill-increase`. 

That flag would replace force in this file: https://github.com/firebase/firebase-tools/blob/ca4349d91047811f68ac615e94db249d5be1a9ee/src/deploy/functions/prompts.ts#L135

aalej:
Hi @Buckwich, thanks for raising this issue and providing detailed feedback. Passing the `--force` flag should only delete functions which are found in your project but do not exist in your local source code.

Just to make sure we’re on the same page here and to get a better understanding of your use case, do you have functions in your project that do not exist in your local source code? 

I do agree with your feedback here, a better description for the `--force` flag could be added, and if there are scenarios where the `--force` flag would not be ideal to use, adding a separate flag would be a good approach. For now, I’ll mark this as a feature request.

Buckwich:
Thanks for your response. Normally we don't have functions without source code, but it has happened and the error prevented us from deploying a corrupt state of our project. If we would use `--force` by default we would miss that valid error.

colerogers:
[...]

PR #6500, opened 2023-11-08
Updated  `--force` description for deploy command

aalej:
### Description

Feedback from #6485 

Updated the `--force` description to use the default description for `--force`. Current description is that `--force` is that it's used to automatically delete Cloud Functions missing from the current working directory without confirmation

```
Options:
  -f, --force              delete Cloud Functions missing from the current working directory without confirmation
```

Looks like it's used to
- automatically deploy functions with a failure policy https://github.com/firebase/firebase-tools/blob/ca4349d91047811f68ac615e94db249d5be1a9ee/src/deploy/functions/prompts.ts#L51
- automatically deploy functions that increase the minimum bill https://github.com/firebase/firebase-tools/blob/ca4349d91047811f68ac615e94db249d5be1a9ee/src/deploy/functions/prompts.ts#L135
- automatically delete Cloud Functions missing from the current working directory without confirmation https://github.com/firebase/firebase-tools/blob/ca4349d91047811f68ac615e94db249d5be1a9ee/src/deploy/functions/prompts.ts#L82

### Scenarios Tested

Running `firebase deploy --help`
```
Usage: firebase deploy [options]

deploy code and assets to your Firebase project

Options:
  -f, --force              automatically accept all interactive prompts
```

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -45,9 +45,7 @@ export const TARGET_PERMISSIONS: Record<(typeof VALID_DEPLOY_TARGETS)[number], s
 
 export const command = new Command("deploy")
   .description("deploy code and assets to your Firebase project")
-  .withForce(
-    "delete Cloud Functions missing from the current working directory without confirmation"
-  )
+  .withForce()
   .option("-p, --public <path>", "override the Hosting public directory specified in firebase.json")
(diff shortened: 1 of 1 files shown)


joehan (approved):

aalej:
I think this is addressed in #6506, going to close this PR for now.

Issue #6526, opened 2023-11-15
Why does firebase login:use <email> require you to be in a Firebase project?


Yukinosuke-Takada:
I wanted to see all Firebase projects that were associated with another account I've added to the firebase-cli. I used the `firebase login:use <email>` command but got some error that said I needed to be in a Firebase project. I made a firebase project (`firebase init`) just to change the account. I find it strange since other `firebase login` commands do not require you to be in a Firebase project.

### [REQUIRED] Environment info

**firebase-tools:** 12.9.0

**Platform:** macOS 14.0

### [REQUIRED] Test case

I get an error: 

```
Error: Could not determine active Firebase project directory
```

when I use command:

```bash
firebase login:use <email>
```

in a directory that is not a Firebase project.

### [REQUIRED] Steps to reproduce

Go to any directory that is not a Firebase project.

```bash
firebase login:use <email>
```

### [REQUIRED] Expected behavior

Should be able to change your account inside firebase-cli outside of Firebase projects.

### [REQUIRED] Actual behavior

Gives Error that says it requires you to be in a Firebase project

```
Error: Could not determine active Firebase project directory
```

aalej:
Hey @Yukinosuke-Takada, thanks for raising this. I was able to reproduce the error you mentioned. Let me raise this to our team to see what we can do here and try to bring someone to provide additional context about this behavior.

aalej:
I discussed this with our team, currently the `firebase login:use <email>` command tells the CLI to use the specified identity for the active Firebase project directory. It seems like this is working as intended.

However, we do agree that it is a reasonable feature to make this command work outside of a Firebase project directory. So for now I’ll be marking this as a feature request. If you’re up for it, feel free to submit a PR!

Yukinosuke-Takada:
@aalej Thanks for the reply. I've read the source code myself and understand why this was implemented like this.

I've made a PR of this feature, that makes it possible to use this command outside a Firebase project directory!

joehan:
Closing this as your PR has been merged and will roll out in the next release. Thanks for the suggestion and contribution!

PR #6533, opened 2023-11-20
Feature: login:use can be used outside of Firebase project to change the global default account.

Yukinosuke-Takada:
### Description

 
issue: #6526 
(feature): It makes it possible to use `firebase login:use <email>` command outside a Firebase project which previously raised an error that said `Error: Could not determine active Firebase project directory`. The new feature DOES NOT EFFECT how we use this command in a Firebase project. It works exactly the same as before (except for raising an error if you try to use the current default account as the new one. See minor change below).  When you use the command outside of a Firebase project, it now changes the global default account to the specified account.

(minor change): In addition to the feature request in the issue, It raises an error when you try to change to the account that's already in use. prior to this, it did not check if it was the same account and doing unnecessary executions.

### Scenarios Tested

- regardless of the directory:
    - Checked if an error is raised when they try to set the current account as the new account. 
    - Checked if an error is raised when they try to use an email not added to the CLI.

- when setting a different account inside of a Firebase project:
    - Checked if nothing changed from prior to implementing this feature. (No code was changed so it should work the same)

- when setting a different account outside of a Firebase project:
    - Checked if `new accounts`'s User and Token was set as global default account and removed from `additionalAccounts`, and the `previous global account`'s  User and Token was added to the `additionalAccounts`. 

### Sample Commands

1. In a Firebase project:
    (As before) When you set a different email.
    ```
    firebase login:use <email>

    ✔  Set default account <email> for current project directory.
    ```

    (New) When you try to set the account to the one already set.
    ```
    firebase login:use <email>

    Error: Already using account <email> for this project directory.
    ```

 2. Outside of a Firebase project:
    (New) When you set a different email to the global default account.
    ```
    firebase login:use <email>

    ✔  Set global default account to <email>.
    ```

    (New) When you try to set the account to the one already set.
    ```
    firebase login:use <email>

    Error: Already using account <email> for the global default account.
    ```

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -188,9 +188,7 @@ export async function loginAdditionalAccount(useLocalhost: boolean, email?: stri
     utils.logWarning(`Already logged in as ${resultEmail}.`);
     updateAccount(newAccount);
   } else {
-    const additionalAccounts = getAdditionalAccounts();
-    additionalAccounts.push(newAccount);
-    configstore.set("additionalAccounts", additionalAccounts);
+    addAdditionalAccount(newAccount);
   }
 
   return newAccount;
@@ -725,4 +723,10 @@ export async function logout(refreshToken: string) {
       original: err,
     });
   }
+}
+
+export function addAdditionalAccount(account: Account) {
+  const additionalAccounts = getAdditionalAccounts();
+  additionalAccounts.push(account);
+  configstore.set("additionalAccounts", additionalAccounts);
 }
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -725,7 +725,11 @@ export async function logout(refreshToken: string) {
   }
 }
 
-export function addAdditionalAccount(account: Account) {
+/**
+ * adds an account to the list of additional accounts.
+ * @param account the account to add.
+ */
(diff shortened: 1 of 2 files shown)


Yukinosuke-Takada:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/firebase/firebase-tools/pull/6533/checks?check_run_id=18850202671) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

Done

joehan (approved):
Thanks for the contribution, this looks great! Could you also add an entry for this to CHANGELOG.md?

```
- Fixed an issue where `login:use` would not work outside of a Firebase project directory. (#6526)
```

Yukinosuke-Takada:
> Thanks for the contribution, this looks great! Could you also add an entry for this to CHANGELOG.md?
> 
> ```
> - Fixed an issue where `login:use` would not work outside of a Firebase project directory. (#6526)
> ```

Included!

joehan (approved):

PR #6548, opened 2023-11-27
Ensure API is enabled for frameworks commands.

taeold:
Before running any frameworks command, we ensure that the API is enabled.

### DIFF
--- a/src/gcp/frameworks.ts
+++ b/src/gcp/frameworks.ts
@@ -1,6 +1,9 @@
 import { Client } from "../apiv2";
+import { needProjectId } from "../projectUtils";
 import { frameworksOrigin } from "../api";
+import { ensure } from "../ensureApiEnabled";
 
+export const API_HOST = new URL(frameworksOrigin).host;
 export const API_VERSION = "v1alpha";
 
 const client = new Client({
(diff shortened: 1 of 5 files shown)


src/gcp/frameworks.ts
```diff
@@ -162,3 +165,11 @@ export async function createBuild(
 
   return res.body;
 }
+
+/**
+ * Ensure that Frameworks API is enabled on the project.
+ */
+export async function ensureApiEnabled(options: any): Promise<void> {
+  const projectId = needProjectId(options);
+  return await ensure(projectId, API_HOST, "frameworks", true);
```
tonyjhuang, line 174: I could be wrong but didn't we want to ensure all of the dependent APIs as well (in case the user manually toggles any of them off?)

taeold: Hmm yeah let me tackle that problem in the follow up PR.

taeold:
### Merge activity

* **Dec 6, 4:33 PM**: [Graphite](https://app.graphite.dev/github/pr/firebase/firebase-tools/6548) rebased this pull request as part of a merge.

Issue #6450, opened 2023-10-16
Functions emulator tries to start if it just sees a functions folder present


CodingDoug:
### Environment info

**firebase-tools:** 12.7.0

**Platform:** macOS

### Test case

Functions emulator starts with nothing more than a functions folder present.

### Steps to reproduce

```
$ firebase init emulators
... choose only Firestore
... take all the defaults

$ firebase emulators:start
... see that only Firestore starts
... ^C

$ mkdir functions
$ firebase emulators:start
... see that Functions emulator runs
```

The emulator generates the error:

> ⬢  functions: Failed to load function definition from source: FirebaseError: Could not detect language for functions at /path/to/functions

But the emulator continues to run on the default 5001 port.

### Expected behavior

I expect that the functions emulator not try to run at all.

IMO the only time it should try to run is if there is something in my firebase.json that suggests it should (perhaps a value for port under emulators/functions/port).

### Actual behavior

Explained above.

aalej:
Hi @CodingDoug, thanks for raising this issue and for providing a detailed report. I was able to reproduce the behavior you mentioned. I’ll notify our engineering team about this issue and mark this as reproducible.

PR #6555, opened 2023-11-28
BREAKING: Ignore 'functions' directory in a Firebase project without a functions config in firebase.json

taeold:
Historically, Firebase CLI reserved directory named `functions` inside a Firebase project directory for use by Cloud Functions for Firebase. This meant that if `functions` directory exists, it was assumed to hold CF3 source code.

However, we recognize that there are valid use case for using `functions` directory name, especially in monorepo-like setups.

With this change, Firebase CLI will only reserve `functions` directory when there is a `functions` configuration entry in firebase.json that does not specify the `source` directory.

This change will be released in the next breaking change of the Firebase CLI (v13).

Fixes https://github.com/firebase/firebase-tools/issues/6450


taeold:
* master
  * **PR #6555** <a href="https://app.graphite.dev/github/pr/firebase/firebase-tools/6555?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/firebase/firebase-tools/6555?utm_source=stack-comment).

joehan (approved):

amedeedaboville:
Note to the maintainers, sorry this references work from >1.5 yrs ago, but I just hit this issue while upgrading:

This breaking change isn't on the "official" [release notes](https://firebase.google.com/support/release-notes/cli#version_1300_-_december_06_2023) of the Firebase CLI docs.

The changelog for 13.0.0 has 3 breaking changes, whereas the [v13 Release in Github](https://github.com/firebase/firebase-tools/releases/tag/v13.0.0) has 4. 

As such, it's very hard to find out about the change and what to do about it (as most users aren't trawling the Github Releases for the changelog). I only found this issue after googling an error message and applying the fix proposed in #6672. I recommend adding it to the changelog on the firebase website, and also suggesting a fix like the one in #6672.

PR #6556, opened 2023-11-29
Better error message when failing to fetch MOTD

joehan:
### Description
Clarifying this error message - this call is best effort, but if it fails there may be an issue with the users proxy or connectivity.

### DIFF
--- a/src/fetchMOTD.ts
+++ b/src/fetchMOTD.ts
@@ -51,7 +51,7 @@ export function fetchMOTD(): void {
         configstore.set("motd.fetched", Date.now());
       })
       .catch((err) => {
-        utils.logWarning("Unable to fetch the CLI MOTD and remote config. This is not a fatal error, but may indicate an issue with your proxy set up.");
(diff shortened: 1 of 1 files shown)


src/fetchMOTD.ts
```diff
@@ -51,7 +51,7 @@ export function fetchMOTD(): void {
         configstore.set("motd.fetched", Date.now());
       })
       .catch((err) => {
-        utils.logWarning("Unable to fetch the CLI MOTD and remote config.");
+        utils.logWarning("Unable to fetch the CLI MOTD and remote config. This is not a fatal error, but may indicate an issue with your proxy set up.");
```
bkendall, line 54: It's not always a proxy. But "network connection" may be generic enough.

```suggestion
        utils.logWarning("Unable to fetch the CLI MOTD and remote config. This is not a fatal error, but may indicate an issue with your network connection.");
```

Also, fix the formatting errors.

bkendall (approved):

PR #6558, opened 2023-11-30
check whether app `not-found` is static

leoortizz:
### Description

Prevent requiring a Cloud Function when the `not-found` page is static or non-existent in the project. At the moment, the `not-found` component in the app router is being identified as dynamic due to inconsistencies with Next.js manifests.

### Scenarios Tested

- not having a `not-found` component
- having a static `not-found` component
- having a dynamic `not-found` component with the route segment config `export const dynamic = 'force-dynamic'`

### DIFF
--- a/src/frameworks/next/index.ts
+++ b/src/frameworks/next/index.ts
@@ -47,6 +47,7 @@ import {
   getHeadersFromMetaFiles,
   cleanI18n,
   getNextVersion,
+  hasStaticAppNotFoundComponent,
 } from "./utils";
 import { NODE_VERSION, NPM_COMMAND_TIMEOUT_MILLIES, SHARP_VERSION, I18N_ROOT } from "../constants";
 import type {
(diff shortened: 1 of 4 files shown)


src/frameworks/next/index.ts
```diff
@@ -208,13 +209,21 @@ export async function build(dir: string): Promise<BuildResult> {
...
     if (appPathsManifest) {
-      const unrenderedServerComponents = getNonStaticServerComponents(
+      let unrenderedServerComponents = getNonStaticServerComponents(
         appPathsManifest,
         appPathRoutesManifest,
         prerenderedRoutes,
         dynamicRoutes
       );
 
+      if (unrenderedServerComponents.includes("/_not-found")) {
```
jamesdaniels, line 219: nit, if you store index then the delete will be easier `delete unrenderedServerComponents[index]` then it can remain a const

jamesdaniels (changes requested):
Add changelog, explore nit

jamesdaniels (approved):
LGTM

src/frameworks/next/index.ts
jamesdaniels: Even better thanks

PR #6562, opened 2023-12-01
Add supported range warnings

jamesdaniels:
* Adding supported range warnings to NextJS, Angular, Nuxt, Astro, and Vite
* Use this supported range for the bootstrap scripts
* Remove supported version from the NextJS doc, that was unnecessarily scaring users away from Firebase IMO

Example NextJS:
Screenshot 2023-12-02 at 8 55 57 PM

Example Svelte:
Screenshot 2023-12-02 at 8 53 49 PM


src/frameworks/sveltekit/index.ts
```diff
@@ -11,7 +11,8 @@ export const name = "SvelteKit";
 export const support = SupportLevel.Experimental;
 export const type = FrameworkType.MetaFramework;
 export const discover = viteDiscoverWithNpmDependency("@sveltejs/kit");
-export { getDevModeHandle } from "../vite";
+
+export { getDevModeHandle, supportedRange } from "../vite";
 
 export async function build(root: string) {
```
jamesdaniels, line 17: sveltekit should probably have it's own supported range

src/frameworks/utils.ts
```diff
@@ -360,21 +361,36 @@ export function relativeRequire(dir: string, mod: string) {
...
+  const allButLast = opts.slice(0, -1);
   return `${allButLast.join(`${separator} `)}${separator} ${conjunction} ${lastElement}`;
 }
 
-export function frameworksCallToAction(message: string, docsUrl = DEFAULT_DOCS_URL, prefix = "") {
-  return `${prefix}${message}
+export function frameworksCallToAction(
+  message: string,
+  docsUrl = DEFAULT_DOCS_URL,
+  prefix = "",
```
leoortizz, line 377: nit: maybe prefix could be handled internally in the function based on the params? e.g.:
```ts
const prefix = framework ? "   " : "";
```
just thinking about reducing the parameters here

src/frameworks/sveltekit/index.ts
leoortizz: indeed but maybe after 2.0 is released? It seems like it won't be released anytime soon: https://github.com/sveltejs/kit/milestone/5

leoortizz (approved):
LGTM. Left only one suggestion.

src/frameworks/sveltekit/index.ts
jamesdaniels: Fair enough

src/frameworks/utils.ts
jamesdaniels: i'll refactor this in a follow on

Issue #6564, opened 2023-12-03
deploy failing due to incorrect Hosting API URL


leoortizz:
### [REQUIRED] Environment info

**firebase-tools:** latest

**Platform:** macOS, GitHub action

### [REQUIRED] Test case

We started having errors with deployments on GitHub actions, then I tried to replicate locally and got the same error. The final error we get on deploy is:
`Error: Unable to parse JSON: SyntaxError: Unexpected token < in JSON at position 0`

While debugging I found that `listSites` here:
https://github.com/firebase/firebase-tools/blob/73c4c9649ea1e8a2c8bb9e5df56ee7209cd3d1c8/src/hosting/runTags.ts#L25

returns this object:
```
[
    {
      name: 'projects/<project-id>/sites/<site-id>',
      defaultUrl: 'https://<project-id>.web.app/',
      appId: '....',
      type: 'DEFAULT_SITE'
    }
]
```

Then, the URL that the Firebase CLI is calling here: 
https://github.com/firebase/firebase-tools/blob/73c4c9649ea1e8a2c8bb9e5df56ee7209cd3d1c8/src/hosting/runTags.ts#L26
https://github.com/firebase/firebase-tools/blob/61523b9086d54ab8370d962f6710387f2256b192/src/hosting/api.ts#L448-L450

is:
```
https://firebasehosting.googleapis.com/v1beta1/projects/-/sites/projects/<project-id>/sites/<site-id>/versions
```
Note `/sites` showing up twice. Basically the URL is invalid so it’s 404ing, leading to a HTML 404 page from Hosting being returned in the API call, which expects a JSON.

This one works:
```
https://firebasehosting.googleapis.com/v1beta1/projects/-/sites/<site-id>/versions
```

I was able to deploy locally with https://github.com/firebase/firebase-tools/pull/6565, however I don't know where the fix should be applied in case a similar problem occurs elsewhere.

### [REQUIRED] Steps to reproduce

`firebase deploy`

### [REQUIRED] Expected behavior

deploy works

### [REQUIRED] Actual behavior

deploy errors out because of invalid URL in Hosting API

PR #6565, opened 2023-12-03
use only site id from site name

leoortizz:
### Description

See https://github.com/firebase/firebase-tools/issues/6564 for context

from:
`projects/project-id/sites/site-id`
use only `site-id`

### Sample Commands

firebase deploy

### DIFF
--- a/src/hosting/runTags.ts
+++ b/src/hosting/runTags.ts
@@ -1,3 +1,4 @@
+import { basename } from "node:path";
 import * as run from "../gcp/run";
 import * as api from "./api";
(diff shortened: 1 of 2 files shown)


src/hosting/runTags.ts
```diff
@@ -1,3 +1,4 @@
+import { basename } from "node:path";
```
jamesdaniels, line 1: Should probably pull in the posix version

jamesdaniels (changes requested):

jamesdaniels (approved):
Add changelog and LGTM

PR #6573, opened 2023-12-05
First crack at getting webframeworks tests back up and running

joehan:
### Description
Attempting to update deps to get tests passing again. Not qutie successful yet.

### DIFF
--- a/scripts/webframeworks-deploy-tests/angular/angular.json
+++ b/scripts/webframeworks-deploy-tests/angular/angular.json
@@ -194,5 +194,8 @@
         }
       }
     }
+  },
+  "cli": {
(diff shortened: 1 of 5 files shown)


jamesdaniels:
* Updated ng to v17
* Updated nextjs to v14
* Fixed a couple bugs in #6480, we weren't handling all legacy builders
* Added `--exit` to mocha so it won't time out on failure
* Added `--retries 2` so it will handle slow to boot up Cloud Functions emulator
* Broke up some tests that were had more than one fetch, leading to a high likelyhood of timeout

jamesdaniels (approved):

PR #6576, opened 2023-12-06
Prompt required IAM permission during frameworks onboarding.

taeold:
Before the CLI attempts to create a connection to GitHub using the Cloud Build API, the CLI now prompts the user if they want the CLI to grant the necessary IAM permissions on the Cloud Build Service Agent (as documented in the [Cloud Build public doc](https://cloud.google.com/build/docs/automating-builds/github/connect-repo-github?generation=2nd-gen#iam_perms)):

```
i To create a new GitHub connection, Secret Manager Admin role (roles/secretmanager.admin) is required on the Cloud Build Service Agent.
? Grant the required role to the Cloud Build Service Agent? Yes
✅ Successfully granted the required role!
```

### DIFF
--- a/src/init/features/frameworks/repo.ts
+++ b/src/init/features/frameworks/repo.ts
@@ -1,12 +1,14 @@
 import * as clc from "colorette";
 
 import * as gcb from "../../../gcp/cloudbuild";
+import * as rm from "../../../gcp/resourceManager";
 import * as poller from "../../../operation-poller";
 import * as utils from "../../../utils";
 import { cloudbuildOrigin } from "../../../api";
 import { FirebaseError } from "../../../error";
 import { logger } from "../../../logger";
 import { promptOnce } from "../../../prompt";
+import { getProjectNumber } from "../../../getProjectNumber";
 
 export interface ConnectionNameParts {
   projectId: string;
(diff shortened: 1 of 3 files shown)


taeold:
* master
  * **PR #6576** <a href="https://app.graphite.dev/github/pr/firebase/firebase-tools/6576?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/firebase/firebase-tools/6576?utm_source=stack-comment).

src/init/features/frameworks/repo.ts
```diff
@@ -81,6 +83,10 @@ export async function linkGitHubRepository(
   logger.info(clc.bold(`\n${clc.yellow("===")} Connect a GitHub repository`));
   const existingConns = await listFrameworksConnections(projectId);
   if (existingConns.length < 1) {
```
tonyjhuang, line 85: This may not be the best sentinel for whether the SA is set up properly. Users can always go into their IAM policy and remove the role grant. The more futureproof way of doing this is to read their IAM policy and verify the service account is still provisioned correctly.

I'm happy to leave it as is for M2 but we should probably note this in a line comment or track in a bug

taeold: > This may not be the best sentinel for whether the SA is set up properly. Users can always go into their IAM policy and remove the role grant.

AFAIK, this is the only code path right now where we need the IAM grant on the GCB service account.  If we are reusing an existing connection, there is no need to make an IAM grant. It's only when setting up the connection for the very first time when we need the IAM grant.

A sub-goal here was to make sure we don't ask for granting this IAM policy unless it's necessary since it's yet another friction for onboarding.

PR #6587, opened 2023-12-07
Never ask to delete Firestore TTL indexes

### LINKED ISSUE #6569: TTL rules creates indexes that cannot be seen in firebase console, but firebase-tools asks if they should be deleted

12.9.1

**firebase-tools:**

macOS

**Platform:**

### [REQUIRED] Test case

Create a TTL rule in the cloud console
Deploy an update to you firestore indexes

firebase-tools will show an error like:

i  firestore: The following field overrides are defined in your project but are not present in your firestore indexes file:
	[ttl_failure.ttl_expiration] -- (ASCENDING) (DESCENDING) (CONTAINS) TTL(ACTIVE)
	[ttl_success.ttl_expiration] -- (ASCENDING) (DESCENDING) (CONTAINS) TTL(ACTIVE)
? Would you like to delete these field overrides? Selecting no will continue the rest of the deployment. No

Provided above.

I would expect firebase-tools to either ignore TTL based indexes (like it appears that firebase console does) OR
Let firebase-tools provide a way to express these TTL rules.

Me, as a user, gets confused if I should try and create the indexes that are mentioned in the output from the deployment. But if I do that, will I overwrite the indexes with a version that does not actually support TTL?

matt-oakes:
### Description

This PR fixes #6569 which asks the user if they want to delete the TTL index when they're deploying unless they have a copy of the index in their JSON file. This is likely not desireable as the TTL index is hidden from the Firebase Console, so it's likely inteded as an implementation detail of the TTL system. With this change, the Firebase CLI will now also ignore it.

### Scenarios Tested

I have tested this by running using the locally build version of the CLI with this fix without my workaround index. The CLI then deploys without asking to delete the TTL index.


joehan (approved):
This looks great! Thanks for the contribution. Could you also add an entry for this to CHANGELOG.md:
```
- Fixed an issue where `deploy` would wrongly prompt to delete Firestore TTL indexes. (#6569)
```

JU-2094:
Q. 
Did you tested for adding SFI exemptions and using TTL policy? 

- (https://firebase.google.com/docs/firestore/query-data/index-overview#indexing_best_practices) - Add single-field exemptions for your TTL fields. 
At first glance I think that is affected. Added another comment in the https://github.com/firebase/firebase-tools/issues/6569
indexing in the field is enabled by default, that's the difference in the configuration file.

joehan:
@JU-2094 I'll wait for your review/approval to merge this since you have the domain expertise on Firestore

JU-2094:
Discussed offline with @joehan , we will instead update the public docs https://firebase.google.com/docs/firestore/query-data/indexing and the README to avoid misunderstandings with the warning message.

This diff warning was intentional to spot differences with SFI exemptions and TTL field. When the user adds SFI on the field with TTL. Removing that should be optional and not enforced (e.g. adding a flag when deploying the command).

matt-oakes:
Thanks for the replies.

@JU-2094 The details of this are a bit beyond me, but I'm happy to hand this over to you as a starting point to implement it the correct way. I still think that the warning showing up is confising behaviour, but if it does show up then It would be amazing if the error message told you exactly what to add to the indexes JSON file to keep the existing index and TLS settings as they are.

PR #6588, opened 2023-12-07
Strip special characters from vite output

jamesdaniels:
Addressing feedback in #6411, fixed in inaccuracy i just merged into changelog


leoortizz (approved):
LGTM

do we need the same for other frameworks?

src/frameworks/vite/index.ts
```diff
@@ -105,7 +106,8 @@ export async function getDevModeHandle(dir: string) {
     const serve = spawn(cli, [], { cwd: dir });
     serve.stdout.on("data", (data: any) => {
       process.stdout.write(data);
-      const match = data.toString().match(/(http:\/\/.+:\d+)/);
+      const dataWithoutAnsiCodes = stripAnsi(data.toString());
+      const match = dataWithoutAnsiCodes.match(/(http:\/\/.+:\d+)/);
```
leoortizz, line 110: nit: we could strip ansi only if there's a match to avoid doing it unnecessarily?

PR #6589, opened 2023-12-07
Rolling firepit back to node 18

joehan:
### Description
Made these changes locally to fix the issues we ran into, now just need to commit them.


jamesdaniels (approved):
Lgtm

Issue #6138, opened 2023-07-17
In frameworks mode, unable to deploy hosting in case configuration names are changed


itherocojp:
In my Angular project, the configuration production has been renamed to live and the development has been done to dev, and the stage has been added currently.

In this situation, when I try to deploy my app to firebase hosting, an error below occurs.

```
Error: Configuration 'production' is not set in the workspace.
```

I know the frameworks mode is now experimental. It would be much better if we could switch it to a different configuration name.

Or are there any good ways to solve this?

Currently I do this.
1. Turn off frameworks mode.
2. Remove source and frameworks region from firebase.json and add public to it.
3. Run ng build and firebase deploy

Thank you.

---

```
"dependencies": {
    "@angular/animations": "^16.1.5",
    "@angular/common": "^16.1.5",
    "@angular/compiler": "^16.1.5",
    "@angular/core": "^16.1.5",
    "@angular/fire": "^7.6.1",
    "@angular/forms": "^16.1.5",
    "@angular/platform-browser": "^16.1.5",
    "@angular/platform-browser-dynamic": "^16.1.5",
    "@angular/router": "^16.1.5",
    "rxjs": "~7.8.0",
    "tslib": "^2.3.0",
    "zone.js": "~0.13.0",
    "firebase": "^7.0 || ^8.0"
  },
```

9kubczas4:
The `custom` configuration name could be specified using the `FIREBASE_FRAMEWORKS_BUILD_TARGET` env variable.

On MacOS: `FIREBASE_FRAMEWORKS_BUILD_TARGET='live' firebase deploy`
On Windows: `set FIREBASE_FRAMEWORKS_BUILD_TARGET='live' firebase deploy`

Also in PR #6594 there is an improvement, the error message for this particular situation will be more meaningful: `Unable to determine the application to deploy, specify a target via the FIREBASE_FRAMEWORKS_BUILD_TARGET environment variable`.

PR #6594, opened 2023-12-07
feat: display meaningful error msg when cannot determine target

9kubczas4:
### Description

Display a meaningful error message when the cannot determine the target.
It would be helpful in scenarios when the user changed default configuration names, for example, `production` to `live`.
This change allows us to understand better what needs to be done by the developer to fix this issue.

Previous error message: `Error: Configuration 'production' is not set in the workspace.`
Current error message: `Unable to determine the application to deploy, specify a target via the FIREBASE_FRAMEWORKS_BUILD_TARGET environment variable`.

Related to #6138.

### Scenarios Tested

1. Create Angular application (v17)
2. Rename the `production` configuration to `live` - adjust `angular.json`
3. Deploy to Firebase using `FIREBASE_FRAMEWORKS_BUILD_TARGET='live' firebase deploy`


jamesdaniels:
Few small things, thanks

src/frameworks/angular/utils.ts
```diff
@@ -10,6 +10,7 @@ import { BUILD_TARGET_PURPOSE } from "../interfaces";
 import { AssertionError } from "assert";
 import { assertIsString } from "../../utils";
 import { coerce } from "semver";
+import { JsonObject } from "@angular-devkit/core";
```
leoortizz, line 13: change it to `import type` and move to the beginning of the file as others Angular imports are

PR #6608, opened 2023-12-12
also check `page` from `appPathRoutesManifest` to get RSC headers

leoortizz:
### Description

Also check if the path from `appPathRoutesManifest` ends with `page` for the RSC `.meta` headers, adding the app directory pages headers to `firebase.json` headers. 

`/page` is for [pages](https://nextjs.org/docs/app/building-your-application/routing/pages-and-layouts), `/route` is for [route handlers](https://nextjs.org/docs/app/building-your-application/routing/route-handlers).


jamesdaniels (approved):
Add changelog otherwise LGTM

PR #6615, opened 2023-12-13
Fix/storage emulator json stream

benjaminpetersen:
### Description

Sometimes the storage emulator fails to resolve and the network request remains in a pending state.
The problem comes from the java subprocess. When a received chunk includes a non-json log or 2 json objects the parse command fails, logs, and returns. 
The fix I implemented is finding matching brackets and parsing the content in those.

Linked issue: #6194

### Scenarios Tested

[FIXED]: When users add `debug(...)` to storage.rules files, the requests intermittently hang.
[FIXED]: When hundreds of requests go out sometimes response get tangled.

### DIFF
--- a/src/emulator/jsonUtils.ts
+++ b/src/emulator/jsonUtils.ts
@@ -1,0 +1,88 @@
+type MatchingBracketsAccumulator = {
+  matching: "none" | "square" | "curly";
+  squareUnclosed: number;
+  curlyUnclosed: number;
+  matches: number[];
+  escapes: number;
+  isString: boolean;
+};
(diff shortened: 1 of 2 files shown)


benjaminpetersen:
It seems I can't agree to the CLA

abhis3 (approved):

abhis3:
Thanks for the fix!! [🚀](https://emojipedia.org/rocket)

christhompsongoogle:
Thank you Benjamin, would you be willing to add some test coverage for this case?

Issue #6636, opened 2023-12-22
Firebase Deploy with Skipped Functions Prevent Deployment with Hosting Rewrite to Function


chriswaterbury:
### [REQUIRED] Environment info

**firebase-tools:**13.0.2

**Platform:** macOS, Ubuntu

### [REQUIRED] Test case

Firebase functions that do not update during a deployment with Skipped (No changes detected) will cause an error if hosting rewrites to a function that didn't update (Skipped (No changes detected)).

Ex hosting config:
```
"hosting": {
    "public": "public",
    "ignore": [
      "firebase.json",
      "**/.*",
      "**/node_modules/**"
    ],
    "rewrites": [
      {
        "source": "/**",
        "function": "helloWorld"
      }
    ]
  }
```

### [REQUIRED] Steps to reproduce

1. firebase init functions
3. write a function and firebase deploy
4. firebase init hosting
5. add a hosting rewrite to your deployed firebase hosting
7. firebase deploy
8. function should be Skipped (No changes detected)
9. Error: HTTP Error: 400, Cloud Run service 'hostingFunction' does not exist in region 'region' in this project.

### [REQUIRED] Expected behavior

firebase deploy should deploy even if changes to a hosting rewrite function aren't made

### [REQUIRED] Actual behavior

```
[debug] [2023-12-22T19:58:05.682Z] ----------------------------------------------------------------------
[debug] [2023-12-22T19:58:05.683Z] Command:       /Users/userhome/.nvm/versions/node/v18.18.2/bin/node /Users/userhome/.nvm/versions/node/v18.18.2/bin/firebase deploy --debug
[debug] [2023-12-22T19:58:05.683Z] CLI Version:   13.0.2
[debug] [2023-12-22T19:58:05.683Z] Platform:      darwin
[debug] [2023-12-22T19:58:05.683Z] Node Version:  v18.18.2
[debug] [2023-12-22T19:58:05.684Z] Time:          Fri Dec 22 2023 14:58:05 GMT-0500 (Eastern Standard Time)
[debug] [2023-12-22T19:58:05.684Z] ----------------------------------------------------------------------
[debug] 
[debug] [2023-12-22T19:58:05.715Z] > command requires scopes: ["email","openid","https://www.googleapis.com/auth/cloudplatformprojects.readonly","https://www.googleapis.com/auth/firebase","https://www.googleapis.com/auth/cloud-platform"]
[...]

PR #6658, opened 2024-01-02
Fix hosting rewrite deployment bug for skipped functions

colerogers:
Addresses #6636


chriswaterbury:
Any ETA on this? This bug a huge pain in our CI/CD pipeline

inlined (approved):
I'm cautious whether this will fix all problems, but there's no controversy with this change so I'm willing to LGTM and see. You've probably looked at this in more detail than me.

Issue #5623, opened 2023-03-22
Firebase auth emulator in multi-tenant mode only exports users tied to the default tenant


damoonrashidi:
When using the auth emulator in a multi-tenant setup, users aren't exported or imported for any tenant other than the default tenant.

### [REQUIRED] Environment info

**firebase-tools:** 11.24.1

**Platform:** macOS

### [REQUIRED] Test case

1. start the auth emulator with `firebase emulators:start --export-on-exit=./users --import=./users`
2. Serve [this file](https://gist.github.com/damoonrashidi/758af4b71fb9df19c2e64cf7b4d7fe75) locally.
3. Follow steps to reproduce

### [REQUIRED] Steps to reproduce

1. Start the auth emulator with `firebase emulators:start --export-on-exit=./users --import=./users`
2. Using the [provided web ui](https://gist.github.com/damoonrashidi/758af4b71fb9df19c2e64cf7b4d7fe75) with firebase auth emulator running, create a user via the Google Provider button.
4. See that the tenant is available in the tenant select dropdown in the emulator auth UI, and that the user is listed for tenant `abc-123` and not the default tenant.
5. Stop the emulator via `SIGINT` (`ctrl+c`)
6. See that the exported user json file does not include the user that was available in the emulator UI.

### [REQUIRED] Expected behavior

The exported users should include all users, segmented by tenant.

### [REQUIRED] Actual behavior

The exported users only include users tied to the default tenant.

edupsousa:
I don't know if it's related but since a few days ago we noticed that users created in other tenants than the default using the Google Auth Provider are not being listed in the popup sign-in screen.

KennyBZA:
This bug is still present on current version.

Environment:
> firebase-tools: 12.4.2
> Platform: macOS

damoonrashidi:
I would try to submit a fix for this myself but it feels like it could potentially be a breaking change to change the format of the exported users and I don't want to take that liberty.

I have a local workaround where generated test users are stored in a json that looks like this

```typescript
[
  tenantId1: [user1, user2],
  tenantId2: [user3, user4]
]
```

And then another script that goes through that json and seeds the local emulator with the users for each tenant, but of course that script needs to be run every time the emulator is started which is pretty annoying.

mikenigels:
I have a project suffering from this issue as well.

@damoonrashidi is the workaround you came up with something you can share with the community? I was just about to develop a similar set of scripts while waiting for an official fix, when I found your post.

[...]

PR #6683, opened 2024-01-12
Fixes issue where Auth emulator sign in with Google only shows default tenant

aalej:
### Description

As mentioned [here](https://github.com/firebase/firebase-tools/issues/5623#issuecomment-1577324485), "other tenants than the default using the Google Auth Provider are not being listed in the popup sign-in screen". This may be because the link created looks like `http://127.0.0.1:9099/emulator/auth/handler?<other_query>&tid=second`. The link has a query of `tid` not `tenantId`

We are currently looking for the `tenantId` query.
```ts
const tenantId = req.query.tenantId as string | undefined;
```

Given that we may be using `tenantId` in the past(or other sdks), keeping both instead of replacing `tenantId` with `tid` might be a better option

### Scenarios Tested

See https://github.com/aalej/issues_5623-tid for a test case

Using the Auth emulator and Hosting emulator
1. Sign in with "default" tenant
2. Sign out
3. Sign in with "second" tenant
    1.  Users from the "default" must not be displayed
4. Sign out
5. Sign in with "default" tenant
    1.  Only users from the "default" must be displayed
6. Sign out
7. Sign in with "second" tenant
    1.  Only users from the "second" must be displayed

### Sample Commands

`firebase emulators:start --project <project_id>`

### DIFF
--- a/src/emulator/auth/handlers.ts
+++ b/src/emulator/auth/handlers.ts
@@ -172,7 +172,7 @@ export function registerHandlers(
     res.set("Content-Type", "text/html; charset=utf-8");
     const apiKey = req.query.apiKey as string | undefined;
     const providerId = req.query.providerId as string | undefined;
-    const tenantId = req.query.tenantId as string | undefined;
+    const tenantId = (req.query.tenantId || req.query.tid) as string | undefined;
     if (!apiKey || !providerId) {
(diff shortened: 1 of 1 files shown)


joehan (approved):
🔥, thanks Alej!

PR #6684, opened 2024-01-12
Prevent pinTags + minInstances from any functions, not just frameworks

inlined:
We previously removed pinTags from the web frameworks configuration options because it is a foot gun, but customers could still hurt themselves if they turn on pinTags with a manual Function rewrite. This prevents the footgun at the API layer.


Berlioz (approved):
LGTM

trepca:
@inlined so how can I use minInstances for my webframework deployment?

PR #6702, opened 2024-01-19
Adds rudimentary email enumeration protection for auth emulator

aalej:
### Description

Added rudimentary support for email enumeration protection on the Auth emulator.
Mainly covers these endpoints:
- identitytoolkit.googleapis.com/v1/accounts:signInWithPassword
- identitytoolkit.googleapis.com/v1/accounts:createAuthUri
- identitytoolkit.googleapis.com/v1/accounts:sendOobCode

#### Reference
Auth Methods: https://firebase.google.com/docs/reference/js/auth.md#fetchsigninmethodsforemail
Email Enumeration Protection: https://cloud.google.com/identity-platform/docs/admin/email-enumeration-protection#overview

#### Manual Testing
Tried to match Auth emulator results with the results from manual testing in https://github.com/aalej/auth-eep-testing

### Sample Commands

`firebase emulators:start`

### Caveats

The public docs for [Email enumeration protection](https://cloud.google.com/identity-platform/docs/admin/email-enumeration-protection#enable) uses this endpoint which currently does not exist in the auth emulator from what I can tell([apiSpec.ts](https://github.com/firebase/firebase-tools/blob/master/src/emulator/auth/apiSpec.ts))

```
https://identitytoolkit.googleapis.com/admin/v2/projects/PROJECT_ID/config?updateMask=emailPrivacyConfig
```

The emulator however does has this endpoint(the admin is not included in the path). See https://github.com/firebase/firebase-tools/blob/5f6c816f4b4d3ca3caba1b35cc638b010b76c92e/src/emulator/auth/apiSpec.ts#L2325
```
https://identitytoolkit.googleapis.com/v2/projects/PROJECT_ID/config?updateMask=emailPrivacyConfig
```

The `emailPrivacyConfig` can be enabled by this curl command
```
curl -X PATCH -d '{"emailPrivacyConfig":{"enableImprovedEmailPrivacy":true}}' \
    -H 'Authorization: Bearer owner' \
    -H 'Content-Type: application/json' -H 'X-Goog-User-Project: <PROJECT_ID>' \
    "http://127.0.0.1:9099/identitytoolkit.googleapis.com/v2/projects/<PROJECT_ID>/config?updateMask=emailPrivacyConfig"
```

Alternatively, the endpoint specific to the emulator can be used to set the configs:
```
http://127.0.0.1:9099/emulator/v1/projects/<project_id>/config
```


joehan (approved):
This looks great to me! Gonna find _someone_ from auth to take a quick pass too

src/test/emulators/auth/oob.spec.ts
```diff
@@ -357,4 +358,41 @@ describeAuthEmulator("accounts:sendOobCode", ({ authApi, getClock }) => {
...
+      .post("/identitytoolkit.googleapis.com/v1/accounts:sendOobCode")
+      .set("Authorization", "Bearer owner")
+      .send({ email: user.email, requestType: "PASSWORD_RESET", returnOobLink: true })
+      .then((res) => {
+        expectStatusCode(400, res);
+        expect(res.body.error).to.have.property("message").equals("EMAIL_NOT_FOUND");
+      });
+  });
+
+  it("should return email when sending a password reset to non-existent user with improved email privacy enabled", async () => {
```
renkelvin, line 374: This email should be sent only when the user exists. See https://cloud.google.com/identity-platform/docs/admin/email-enumeration-protection#overview

renkelvin (approved):
Thanks, LGTM with a comment on a test case.

src/test/emulators/auth/oob.spec.ts
aalej: Just to verify. Are you referring to this line “When you make a password reset request, a verification email is sent only if the email address exists”. If so, I think the name of the test case I wrote may be a little bit misleading.

In this case I’m referring to the email address to be returned in the api response. When an email does not exists, a password reset email would not be sent, just that the api response would look like
```json
{
    "kind": "identitytoolkit#GetOobConfirmationCodeResponse",
    "email": "non_existent_email@fake.fake"
}
```

Should I leave a comment in the test case describing this? Or just change the test name to “should return email address...”

LMK in case I misunderstood anything.

renkelvin: Thanks for the clarification. I think "should return email address.." sounds more accurate.

aalej: Thanks, updated the test title.

Issue #5962, opened 2023-06-12
Sveltekit dev module variable is true on prod deployment


danielfrg:
### [REQUIRED] Environment info

**firebase-tools:** 11.28.0 and the latest [Github Action](https://github.com/FirebaseExtended/action-hosting-deploy)

**Platform:**

### [REQUIRED] Test case

Any sveltekit project that uses the [dev module](https://kit.svelte.dev/docs/modules#$app-environment-dev)

### [REQUIRED] Steps to reproduce

For example I use it to show stuff in the UI when I am in dev mode:

```
{#if dev}
   ... stuff ...
{/if}
```

### [REQUIRED] Expected behavior

In prod the content should be ignored since the dev variable is false. This works fine with doing a regular svelte build. For example doing an static build and deploying that to firebase. When using the frameworks this variable is set to `true` for some reason.

### [REQUIRED] Actual behavior

When doing firebase deploy it doesn't seem to set this variable for some reason, probably because the adaptar is ignored. Not 100% sure why.

caidanw:
I was having trouble with this as well and finally found the answer after some research. It looks like what is happening is that somewhere in the firebase-tools codebase (not sure where yet) the env var `NODE_ENV` is being set to `development` instead of `production`.

My research let me to find that when running `firebase deploy` and checking the build output in the `.svelte-kit/output` dir, I can see dev specific code and the presence of `.svelte-kit/output/server/chunks/dev-ssr.js`. I can also see the presence of `.firebase/[PROJECT-ID]/functions/chunks/dev-ssr.js` as well.

I found a few other github issues that referenced a similar problem but with the env var `FIREBASE_FRAMEWORKS_BUILD_TARGET`
- https://github.com/firebase/firebase-tools/issues/5907
- https://github.com/firebase/firebase-tools/issues/6041
- https://github.com/firebase/firebase-tools/issues/5896

There is a lot of mention in those tickets that the issue related to `FIREBASE_FRAMEWORKS_BUILD_TARGET` has been fixed in firebase-tools `v12.3.0`, which was weird since I am on `v12.4.1`.

I tried using `FIREBASE_FRAMEWORKS_BUILD_TARGET=production firebase deploy` but that didn't help. After investigating the code I came across the env var `NODE_ENV` in:
https://github.com/firebase/firebase-tools/blob/3d3f5ef4ee2a57d9749b86a11ffb94a98776bbc3/src/frameworks/utils.ts#L296-L303

[...]

PR #6710, opened 2024-01-23
Fix invalid assignment error

rrousselGit:
~~This reverts commit d0972f7bbb31050bd7cfbcab718fe1c0fb8faa9a.~~

~~That ENV changes breaks bundling, as Webpack replaces `process.env.NODE_ENV = target;` to `"development" = target`, which is not valid~~

Fixed by making the expression non-statically analysable, which disables Webpack's shenanigans

### Sample Commands

cc @leoortizz

### DIFF
--- a/src/frameworks/vite/index.ts
+++ b/src/frameworks/vite/index.ts
@@ -89,12 +89,16 @@ export async function build(root: string, target: string) {
 
   const originalNodeEnv = process.env.NODE_ENV;
 
-  let envKey = 'NODE_ENV'
+  let envKey = "NODE_ENV";
+  // Voluntarily making .env[key] not statically analyzable to avoid
+  // Webpack from converting it to "development" = target;
   process.env[envKey] = target;
 
   await build({ root, mode: target });
   process.chdir(cwd);
 
+  // Voluntarily making .env[key] not statically analyzable to avoid
+  // Webpack from converting it to "development" = target;
   process.env[envKey] = originalNodeEnv;
 
   return { rewrites: [{ source: "**", destination: "/index.html" }] };
--- a/src/frameworks/vite/index.ts
+++ b/src/frameworks/vite/index.ts
@@ -89,7 +89,7 @@ export async function build(root: string, target: string) {
 
   const originalNodeEnv = process.env.NODE_ENV;
 
-  let envKey = "NODE_ENV";
+  const envKey = "NODE_ENV";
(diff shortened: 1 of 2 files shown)


leoortizz (commented):
@rrousselGit would `process.env[“NODE_ENV”]` help on this? Trying to not reopen #5962

rrousselGit:
Let me check real quick.

jamesdaniels:
q, where are we bundling the cli with the webpack environment plugin? it would also negatively affect the emulator's environment testing

jamesdaniels:
e.g, `src/frameworks/utils.ts#getFrameworksBuildTarget`

rrousselGit:
> @rrousselGit would `process.env[“NODE_ENV”]` help on this? Trying to not reopen #5962

Unfortunately no. But good call. With one extra step, I was able to fix the error without removing it.

rrousselGit:
> q, where are we bundling the cli with the webpack environment plugin? it would also negatively affect the emulator's environment testing & the bundle that is shipped to hosting

I'm not sure which line exactly is triggering this, but I'd assume it'd be somewhere in there: https://github.com/firebase/firebase-tools/blob/84c5ba67914d25593d6cb8cc89d1b5552edf5684/firebase-vscode/webpack.common.js

src/frameworks/vite/index.ts
```diff
@@ -88,15 +88,18 @@ export async function build(root: string, target: string) {
   process.chdir(root);
 
   const originalNodeEnv = process.env.NODE_ENV;
-  // @ts-expect-error - NODE_ENV is `development` when building for production
-  // temporarily replace with target during build
-  process.env.NODE_ENV = target;
+
+  const envKey: string = "NODE_ENV";
```
leoortizz, line 92: nit: mind adding a comment here to explain why `: string` is necessary?

leoortizz (approved):
LGTM

src/frameworks/vite/index.ts
rrousselGit, line 92:
```suggestion
  // Downcasting as `string` as otherwise it is inferred as `readonly 'NODE_ENV'`,
  // but `env[key]` expects a non-readonly variable.
  const envKey: string = "NODE_ENV";
```

rrousselGit: Done

PR #6711, opened 2024-01-23
Replace disabled TextField with plain text instead

rrousselGit:
### Description

Screenshot 2024-01-23 at 22 53 56


firebase-vscode/webviews/components/EmulatorPanel.tsx
```diff
@@ -186,14 +179,11 @@ export function EmulatorPanel({
...
-      <VSCodeTextField
-        disabled={true}
-        className="in-line"
-        value={emulatorUiSelections.projectId}
-        onChange={(event) => projectIdChanged(event)}
-      ></VSCodeTextField>
+      <span>
+        {"Current project ID: "}
+        {/* TODO(christhompson): convert this into a demo- prefix checkbox or something. */}
+        <kbd>{emulatorUiSelections.projectId}</kbd>
```
hlshen, line 185: what is kbd? can we avoid the box styling and just leave it as text (in the span).

hlshen (approved):

firebase-vscode/webviews/components/EmulatorPanel.tsx
rrousselGit: It's like `this`.
I thought it'd be looking better, but I don't mind removing it.

rrousselGit: Switched to bold. What do you think?
I want to differentiate it from the label. It feels odd to me otherwise

Screenshot 2024-01-25 at 08 27 51

hlshen: ye let's just keep that! I think you might be merging into the wrong branch.

rrousselGit: Ah yes, I started from the previous "revert" Pr to be able to work locally

PR #6724, opened 2024-01-29
Fix project prompt

rrousselGit:
### Description

Two issues were found with the project prompt:
- If the user aborts the modal without picking a project, the code threw an undefined error
- If the prompt was opened before projects were loaded, an empty list would be shown

This PR handles the undefined case and causes the picker to await for projects to be loaded.

### DIFF
--- a/firebase-vscode/package-lock.json
+++ b/firebase-vscode/package-lock.json
@@ -947,9 +947,9 @@
       "dev": true
     },
     "node_modules/@types/node": {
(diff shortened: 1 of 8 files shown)


hlshen (approved):

hlshen:
Can we write a test for this?

rrousselGit:
Sure. Let me merge the e2e PR and then update this PR to add an e2e test

PR #6725, opened 2024-01-29
Add firebase.selectProject command

rrousselGit:
### Description

This exposes a `firebase.selectProject` command, which users can bind to a shortcut.

This command will be reused by statusbar changes for Firemat

### DIFF
--- a/firebase-vscode/src/core/project.ts
+++ b/firebase-vscode/src/core/project.ts
@@ -12,7 +12,9 @@ import { updateFirebaseRCProject } from "../config-files";
 import { globalSignal } from "../utils/globals";
 
 /** Available projects */
-export const projects = globalSignal<Record<string, FirebaseProjectMetadata[]>>({});
+export const projects = globalSignal<Record<string, FirebaseProjectMetadata[]>>(
+  {}
+);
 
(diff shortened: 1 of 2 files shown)


firebase-vscode/src/core/project.ts
```diff
@@ -74,48 +76,53 @@ export function registerProject({
...
-        vscode.extensions.getExtension("google.monospace");
-      process.env.MONOSPACE_DAEMON_PORT =
-        monospaceExtension.exports.getMonospaceDaemonPort();
-      try {
-        const projectId = await selectProjectInMonospace({
-          projectRoot: currentOptions.value.cwd,
-          project: undefined,
-          isVSCE: true,
-        });
+  const command = vscode.commands.registerCommand(
```
hlshen, line 79: do we need to change the webview to call this command?

rrousselGit: The webview invokes the command through the HTMLBroker. Hence the "broker.on" at the end of the changes

hlshen: ah I missed that!

hlshen (approved):

PR #6726, opened 2024-01-30
Improve FAH create & delete

inlined:
1. Create now sets deployment-tool so we can track where users are coming from
2. Create now sets a rollout policy like the Console
3. Delete now works
4. Delete no longer makes buildId required

### DIFF
--- a/src/commands/apphosting-backends-delete.ts
+++ b/src/commands/apphosting-backends-delete.ts
@@ -5,7 +5,6 @@ import { FirebaseError } from "../error";
 import { promptOnce } from "../prompt";
 import { logger } from "../logger";
 import { DEFAULT_REGION } from "../init/features/apphosting/constants";
-import { last } from "../utils";
 import * as utils from "../utils";
 import * as apphosting from "../gcp/apphosting";
 
@@ -30,7 +29,10 @@ export const command = new Command("apphosting:backends:delete")
   .action(async (options: Options) => {
     const projectId = needProjectId(options);
     let location = options.location as string;
-    let backendId = options.backend as string;
+    const backendId = options.backend as string;
+    if (!backendId) {
+      throw new FirebaseError("Backend id can't be empty.");
+    }
 
     if (!location) {
       const allowedLocations = (await apphosting.listLocations(projectId)).map(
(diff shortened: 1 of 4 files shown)


src/commands/apphosting-backends-delete.ts
```diff
@@ -47,21 +45,27 @@ export const command = new Command("apphosting:backends:delete")
       });
     }
 
+    let backend: apphosting.Backend;
+    if (backendId) {
```
bkendall, line 49: Okay, this is odd. According to the approved proposal, the backend ID is supposed to be a required parameter, not a flag. That eliminates the need to pick a backend at all.

src/test/init/apphosting/index.spec.ts
```diff
@@ -66,10 +67,10 @@ describe("operationsConverter", () => {
         repository: cloudBuildConnRepo.name,
         rootDirectory: "/",
       },
-      labels: {},
+      labels: deploymentTool.labels(),
     };
 
-    it("should createBackend", async () => {
+    it("should createBackend with rollout policy", async () => {
```
bkendall, line 73: Nothing here tests to make sure the rollout policy was called. Please add the appropriate logic here to test for that

bkendall (changes requested):

src/commands/apphosting-backends-delete.ts
inlined: Irrespective of the proposal, do you like this better? It seemed odd to me that location could be browsed but backend couldn't. If you like this better we can push it through API council. As an EAP API, do you feel like it is OK to release and then get approval, or should I splice this into another CL until after approval?

src/test/init/apphosting/index.spec.ts
inlined: I may need to revert the label. I originally wrote the test to exercise this but realized it was for a smaller utility function and the larger flow is completely untested. I can either:

1. Move creation of the rollout policy into a utility function and test it. This would be a trivial whitebox test that is unlikely to catch anything though.
2. Add a new test that covers the overall flow of the entiere init process.
3. Cry that init functions are generally sparsely tested in this codebase.

You make the call and I'll follow.

src/commands/apphosting-backends-delete.ts
bkendall: I personally don't like it better, because I hate the prompt selector flow in a CLI 🤷‍♂️ It's not a wrong thing to do, but I'm assuming that there were product discussions that brought us to the API proposal (as that's a requirement at the top of the document) and that the pattern in the proposal was agreed upon by all the relevant stakeholders already. If you'd like to make a change to what was proposed there, starting back in the process, involving the right people, and getting an addendum would be appropriate.

That said, I thought the general pattern was to:

1. require backendID as a parameter in the command
2. default to looking at us-central1 as we 'traditionally' have

If someone is outside of us-central1, they get to include the `--location` flag.

Having this flow also makes it more complex. Sticking with the picker, you have to also consider what happens in non-interactive environment (like a CI system). As written, this would hang, I believe, which may be unexpected (instead of simply failing out).

src/test/init/apphosting/index.spec.ts
bkendall: Ah, yeah, this isn't actually testing `doSetup`. Probably just revert the test name then. init function coverage is kinda a known weak point.

bkendall (changes requested):

src/commands/apphosting-backends-delete.ts
inlined: (If it's chosen that we should keep the existing dialog, I agree that we need to check for nonInteractive)

inlined:
Controversial UX changes have been removed and test name has been reverted. PTAL.

src/gcp/apphosting.ts
```diff
@@ -197,10 +197,20 @@ export interface RolloutPolicy {
   // end oneof trigger
   stages?: RolloutStage[];
   disabled?: boolean;
+
+  // TODO: This will be undefined if disabled is not true, right?
```
bkendall, line 201: Yes, it should be.

bkendall (approved):

PR #6730, opened 2024-01-30
Add support for automated tests in App Distribution

lfkellogg:
### Description

Adding `--test-devices` and related flags to `appdistribution:distribute` action. This will trigger the new Automated Tester feature.

### Scenarios Tested

- Kick off tests for a release
- Kick off tests async
- Kick off tests with username/password
- Kick off tests with username/password and resource IDs
- Kick off tests with username but not password (error case)
- Kick off tests with username resource but not password resource (error case)
- Kick off tests with username/password resource but not username/password (error case)

### Sample Commands

`firebase appdistribution:distribute --app <app id> <path> --test-devices <devices>`
`firebase appdistribution:distribute --app <app id> <path> --test-devices <devices> --test-async`
`firebase appdistribution:distribute --app <app id> <path> --test-devices <devices> --test-username <user> --test-password <password>`
`firebase appdistribution:distribute --app <app id> <path> --test-devices <devices> --test-username <user> --test-password <password> --test-username-resource <resource id> --test-password-resource <resource id>`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,2 @@
 - Added rudimentary email enumeration protection for auth emulator. (#6702)
+- You can now run customized automated tests on your Android apps in App Distribution, with the Automated Tester feature (beta). This feature automatically runs tests on your Android apps on virtual and physical devices at different API levels. To learn how to run an automated test, see [Run an automated test for Android apps](https://firebase.google.com/docs/app-distribution/android-automated-tester). (#6730)
--- a/src/appdistribution/client.ts
+++ b/src/appdistribution/client.ts
@@ -6,113 +6,16 @@ import { Distribution } from "./distribution";
 import { FirebaseError } from "../error";
 import { Client } from "../apiv2";
 import { appDistributionOrigin } from "../api";
-
-/**
- * Helper interface for an app that is provisioned with App Distribution
- */
-export interface AabInfo {
-  name: string;
-  integrationState: IntegrationState;
-  testCertificate: TestCertificate | null;
-}
-
-export interface TestCertificate {
-  hashSha1: string;
-  hashSha256: string;
-  hashMd5: string;
-}
-
-/** Enum representing the App Bundles state for the App */
-export enum IntegrationState {
-  AAB_INTEGRATION_STATE_UNSPECIFIED = "AAB_INTEGRATION_STATE_UNSPECIFIED",
-  INTEGRATED = "INTEGRATED",
-  PLAY_ACCOUNT_NOT_LINKED = "PLAY_ACCOUNT_NOT_LINKED",
-  NO_APP_WITH_GIVEN_BUNDLE_ID_IN_PLAY_ACCOUNT = "NO_APP_WITH_GIVEN_BUNDLE_ID_IN_PLAY_ACCOUNT",
-  APP_NOT_PUBLISHED = "APP_NOT_PUBLISHED",
-  AAB_STATE_UNAVAILABLE = "AAB_STATE_UNAVAILABLE",
-  PLAY_IAS_TERMS_NOT_ACCEPTED = "PLAY_IAS_TERMS_NOT_ACCEPTED",
-}
-
-export enum UploadReleaseResult {
-  UPLOAD_RELEASE_RESULT_UNSPECIFIED = "UPLOAD_RELEASE_RESULT_UNSPECIFIED",
-  RELEASE_CREATED = "RELEASE_CREATED",
-  RELEASE_UPDATED = "RELEASE_UPDATED",
-  RELEASE_UNMODIFIED = "RELEASE_UNMODIFIED",
-}
-
-export interface Release {
-  name: string;
-  releaseNotes: ReleaseNotes;
-  displayVersion: string;
-  buildVersion: string;
-  createTime: Date;
-  firebaseConsoleUri: string;
-  testingUri: string;
-  binaryDownloadUri: string;
-}
-
-export interface ReleaseNotes {
-  text: string;
-}
-
-export interface UploadReleaseResponse {
-  result: UploadReleaseResult;
-  release: Release;
-}
-
-export interface BatchRemoveTestersResponse {
-  emails: string[];
-}
-
-export interface Group {
-  name: string;
-  displayName: string;
-  testerCount?: number;
-  releaseCount?: number;
-  inviteLinkCount?: number;
-}
-
-export interface CreateReleaseTestRequest {
-  releaseTest: ReleaseTest;
-}
-
-export enum TestState {
-  IN_PROGRESS = "IN_PROGRESS",
-  PASSED = "PASSED",
-  FAILED = "FAILED",
-  INCONCLUSIVE = "INCONCLUSIVE",
-}
-
-export interface TestDevice {
-  model: string;
-  version: string;
-  locale: string;
-  orientation: string;
-}
-
-export interface DeviceExecution {
-  device: TestDevice;
-  state?: TestState;
-  failedReason?: string;
-  inconclusiveReason?: string;
-}
-
-export interface FieldHints {
-  usernameResourceName?: string;
-  passwordResourceName?: string;
-}
-
-export interface LoginCredential {
-  username?: string;
-  password?: string;
-  fieldHints?: FieldHints;
-}
-
-export interface ReleaseTest {
-  name?: string;
-  deviceExecutions: DeviceExecution[];
-  loginCredential?: LoginCredential;
-}
+import {
+  AabInfo,
+  BatchRemoveTestersResponse,
+  DeviceExecution,
+  Group,
+  LoginCredential,
+  ReleaseTest,
+  TestDevice,
+  UploadReleaseResponse,
+} from "./types";
 
 /**
  * Makes RPCs to the App Distribution server backend.
@@ -321,7 +224,6 @@ export class AppDistributionClient {
           loginCredential,
         },
       });
-      utils.logSuccess(`Release test created successfully`);
       return response.body;
     } catch (err: any) {
       throw new FirebaseError(`Failed to create release test ${err}`);
--- a/src/appdistribution/client.ts
+++ b/src/appdistribution/client.ts
@@ -98,8 +98,8 @@ export interface DeviceExecution {
 }
 
 export interface FieldHints {
-  usernameResource?: string;
-  passwordResource?: string;
+  usernameResourceName?: string;
+  passwordResourceName?: string;
 }
 
 export interface LoginCredential {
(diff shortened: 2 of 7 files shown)


rebehe (approved):

src/commands/appdistribution-distribute.ts
```diff
@@ -43,13 +50,39 @@ export const command = new Command("appdistribution:distribute <release-binary-f
     "--groups-file <file>",
     "path to file with a comma separated list of group aliases to distribute to",
   )
+  .option(
+    "--test-devices <string>",
+    "semicolon-separated list of devices to run automated tests on, in the format 'model=<model-id>,version=<os-version-id>,locale=<locale>,orientation=<orientation>'. Run 'gcloud firebase test android|ios models list' to see available devices",
```
MaesterChestnut, line 55: can you add a note in here about this feature being in beta to keep the language in line with our other CLI tools?

lfkellogg: Done! I added all the similar updates from the fastlane plugin: https://github.com/firebase/fastlane-plugin-firebase_app_distribution/commit/e14eb83330b14f20bd7bc6214429e6cb9e74dc47

CHANGELOG.md
```diff
@@ -1 +1,2 @@
 - Added rudimentary email enumeration protection for auth emulator. (#6702)
+- You can now run customized automated tests on your Android apps in App Distribution, with the Automated Tester feature (beta). This feature automatically runs tests on your Android apps on virtual and physical devices at different API levels. To learn how to run an automated test, see [Run an automated test for Android apps](https://firebase.google.com/docs/app-distribution/android-automated-tester). (#6730)
```
rebehe, line 2: Question for the Firebase CLI reviewer - we heard there might be a way to mark something as "beta" in the Firebase CLI. Is this applicable here?

rebehe: Nvm - we found go/firebase-cli-contributor#hide-the-command-behind-a-flag, but given this is not a new command (just a new optional param on an existing command), it's not applicable here

src/appdistribution/client.ts
```diff
@@ -72,17 +72,63 @@ export interface Group {
   inviteLinkCount?: number;
 }
 
+export interface CreateReleaseTestRequest {
+  releaseTest: ReleaseTest;
+}
+
+export enum TestState {
```
joehan, line 79: Optional: Consider a string union type here instead (ie `IN_PROGRESS" | "PASSED" | "FAILED" | "INCONCLUSIVE`). They tend to be a bit easier to work with IMO.

src/appdistribution/client.ts
```diff
@@ -260,4 +306,41 @@ export class AppDistributionClient {
...
+    try {
+      const response = await this.appDistroV1AlphaClient.request<ReleaseTest, ReleaseTest>({
+        method: "POST",
+        path: `${releaseName}/tests`,
+        body: {
+          deviceExecutions: devices.map((d) => this.mapDeviceToExecution(d)),
+          loginCredential,
+        },
+      });
+      utils.logSuccess(`Release test created successfully`);
```
joehan, line 324: Nit: I see we do it throughout this file, but it is better separation of concerns to do this logging at the call site instead of inside this client file.

src/appdistribution/client.ts
```diff
@@ -260,4 +306,41 @@ export class AppDistributionClient {
...
+    releaseName: string,
+    devices: TestDevice[],
+    loginCredential?: LoginCredential,
+  ): Promise<ReleaseTest> {
+    try {
+      const response = await this.appDistroV1AlphaClient.request<ReleaseTest, ReleaseTest>({
+        method: "POST",
+        path: `${releaseName}/tests`,
+        body: {
+          deviceExecutions: devices.map((d) => this.mapDeviceToExecution(d)),
```
joehan, line 320: Can simplify to:
```suggestion
          deviceExecutions: devices.map(this.mapDeviceToExecution),
```

src/appdistribution/client.ts
```diff
@@ -72,17 +72,63 @@ export interface Group {
   inviteLinkCount?: number;
 }
 
+export interface CreateReleaseTestRequest {
```
joehan, line 75: Nit: Consider splitting these types off into their own file `src/appdistribution/types.ts`

src/appdistribution/options-parser-util.ts
```diff
@@ -62,3 +64,105 @@ export function getAppName(options: any): string {
...
+  if (!value && file) {
+    ensureFileExists(file);
+    value = fs.readFileSync(file, "utf8");
+  }
+
+  if (!value) {
+    return [];
+  }
+
+  // The value is split into a string[]
```
joehan, line 84: Nit: This comment isn't very helpful - it basically just explains the type signature of `.split()`, which well known, but doesn't explain what the rest of this  chain does. Consider removing it or improving it.

src/appdistribution/options-parser-util.ts
```diff
@@ -62,3 +64,105 @@ export function getAppName(options: any): string {
...
+  const deviceStrings = value
+    .split(/[;\n]/)
+    .map((entry) => entry.trim())
+    .filter((entry) => !!entry);
+
+  return deviceStrings.map((str) => parseTestDevice(str));
```
joehan, line 90: Nit: Since you're already chaining here, you could go a step further and: 
```
return value
    .split(/[;\n]/)
    .map((entry) => entry.trim())
    .filter((entry) => !!entry)
    .map((str) => parseTestDevice(str));
```

src/commands/appdistribution-distribute.ts
```diff
@@ -43,13 +50,42 @@ export const command = new Command("appdistribution:distribute <release-binary-f
     "--groups-file <file>",
     "path to file with a comma separated list of group aliases to distribute to",
   )
+  .option(
+    "--test-devices <string>",
+    "semicolon-separated list of devices to run automated tests on, in the format 'model=<model-id>,version=<os-version-id>,locale=<locale>,orientation=<orientation>'. Run 'gcloud firebase test android|ios models list' to see available devices. Note: This feature is in beta.",
```
joehan, line 55: I don't love that this feature tells users to go use gcloud for this. I assume that listing available models is basically a single API call to an API that we've already implemented a client for. Why not just add an equivalent command to firebase-tools?

src/commands/appdistribution-distribute.ts
```diff
@@ -43,13 +50,42 @@ export const command = new Command("appdistribution:distribute <release-binary-f
...
+  )
+  .option(
+    "--test-devices-file <string>",
+    "path to file containing a list of semicolon- or newline-separated devices to run automated tests on, in the format 'model=<model-id>,version=<os-version-id>,locale=<locale>,orientation=<orientation>'. Run 'gcloud firebase test android|ios models list' to see available devices. Note: This feature is in beta.",
+  )
+  .option("--test-username <string>", "username for automatic login")
+  .option("--test-password <string>", "password for automatic login")
```
joehan, line 62: Password is likely a sensitive field, right? It is not appropriate to supply it as a flag, since that will be persisted in bash history.

src/commands/appdistribution-distribute.ts
```diff
@@ -168,4 +204,65 @@ export const command = new Command("appdistribution:distribute <release-binary-f
...
+          );
+      }
+    }
+  }
+  throw new FirebaseError("It took longer than expected to process your test, please try again.", {
+    exit: 1,
   });
+}
+
+function delay(ms: number) {
```
joehan, line 262: nit: Add a return type.

joehan (changes requested):
Code mostly looks good, with a few small style fixes suggested. However, this change needs an approved API proposal since it changes that public API of the CLI. I have some concerns about the API for this command that would be better discussed there. I couldn't find one at go/firebase-api-tracker - could you link one if I missed it, or start one up?

src/appdistribution/client.ts
```diff
@@ -72,17 +72,63 @@ export interface Group {
...
+  name?: string;
+  deviceExecutions: DeviceExecution[];
+  loginCredential?: LoginCredential;
+}
+
 /**
  * Makes RPCs to the App Distribution server backend.
  */
 export class AppDistributionClient {
-  appDistroV2Client = new Client({
```
kaibolay, line 79: The `V2` was due to `import { Client } from "../apiv2";` but I think renaming based on the OnePlatform API version probably makes sense

src/appdistribution/client.ts
lfkellogg: Thanks for the tip! Done.

src/appdistribution/client.ts
lfkellogg: Sounds good. I'll at least fix this new case just to avoid propagating the problem.

src/appdistribution/client.ts
lfkellogg: Done

src/appdistribution/client.ts
lfkellogg: Done

src/appdistribution/options-parser-util.ts
lfkellogg: Removed

src/appdistribution/options-parser-util.ts
lfkellogg: Good call, done

src/commands/appdistribution-distribute.ts
lfkellogg: Addressed this in [go/atad-cli-api-review](http://go/atad-cli-api-review).

src/commands/appdistribution-distribute.ts
lfkellogg: Done

src/appdistribution/client.ts
lfkellogg: OK so this actually now causes a lint error because of potential `this` binding issues:

```
/Users/lkellogg/dev/firebase-tools/src/appdistribution/client.ts
  223:41  error  Avoid referencing unbound methods which may cause unintentional scoping of `this`.
If your function does not access `this`, you can annotate it with `this: void`, or consider using an arrow function instead  @typescript-eslint/unbound-method
```

What do you think, I feel like I'd rather have a verbose lambda then have to add `this: void` to the method signature...

lfkellogg: Oh never mind, I just moved the map function to types.ts (which I think makes sense) and now it's all good.

joehan (approved):

src/appdistribution/options-parser-util.ts
```diff
@@ -62,3 +64,109 @@ export function getAppName(options: any): string {
...
+  }
+  return { model, version, locale, orientation };
+}
+
+/**
+ * Takes option values for username and password related options and returns a LoginCredential
+ * object that can be passed to the API.
+ */
+export function getLoginCredential(
+  username?: string,
```
joehan, line 133: Optional: This would be a good spot to use a parameter object - with 5 optional fields all with the same type, it seems very easy to accidentally pass in args in the wrong order. A param object makes this a bit more explicit and readable:
```
function getLoginCredential( args: {
  username?: string,
  password?: string,
  passwordFile?: string,
  usernameResourceName?: string,
  passwordResourceName?: string,
})

lfkellogg: I like it, done

PR #6731, opened 2024-01-31
Remove development-only files from Next.js deployments

### LINKED ISSUE #6715: Web frameworks - Next.js - Functions deployment is huge due to .next/cache
### Environment info

**firebase-tools:** 13.1.0

**Platform:** macOS

### Test case

Create a new next.js project and then attempt to deploy it to Firebase, the functions folder will be huge like 100MB or upto 300MB.

### Steps to reproduce

Create a new next.js project and then attempt to deploy it to Firebase, the functions folder will be huge like 100MB or upto 300MB.

### Expected behavior

The `.next/cache` folder should be ignored by Firebase Functions.

### Actual behavior

The `.next/cache` folder is being copied over to `.firebase/app-name/functions/.next/cache`

leoortizz:
### Description

Remove development-only files from deployed files. Closes #6715

### Scenarios Tested

see test case added in 1f579d7601b63a0d3d9f3da5e7f3afb27f89cdfb


src/frameworks/next/utils.ts
```diff
@@ -421,3 +421,17 @@ export async function hasStaticAppNotFoundComponent(
...
 }
+
+/**
+ * Remove development files from the .next directory for production deployments.
+ */
+export async function removeDevFiles(distDirPath: string): Promise<void> {
+  const cacheDevFiles = join(distDirPath, "cache", "**", "*-development*");
+  const cacheEslintFiles = join(distDirPath, "cache", "eslint");
+
+  const filesToRemove = globSync(`{${cacheDevFiles},${cacheEslintFiles}}`, {
```
jamesdaniels, line 432: alt, `[cacheDevFiles,cacheEslintFiles].join(",")`

why not use async here?

jamesdaniels:
Would be great if we didn't copy those files to start with but that would probably make the code more complex, wdyt?

src/frameworks/next/utils.ts
leoortizz: I followed whatever's already used here, I'm hesitant to alter the glob usage per this: #6166. 

We have `node-glob` v7 which only supports async with a callback. If we bump it to latest we could use the `await` syntax without manually creating a Promise, can we do that?

leoortizz:
> Would be great if we didn't copy those files to start with but that would probably make the code more complex, wdyt?

indeed, I can make that change. I thought it would be slower than just removing the files but it's probably similar

src/frameworks/next/utils.ts
leoortizz: switched to the async glob, PTAL. However, AFAIK we can't create stubs of `glob` that way, so I couldn't write unit tests for it. Either way, added an E2E test which is better in this case.

jamesdaniels (approved):

PR #6743, opened 2024-02-05
FAH uses ids using the same naming policy as rolloutPolicy

inlined:
Additionally added listRollouts and fixed listBuilds so that it paginates over multiple requests and preserves the list of unavailable regions (and handles them)

### DIFF
--- a/src/gcp/apphosting.ts
+++ b/src/gcp/apphosting.ts
@@ -356,8 +356,8 @@ export async function listBuilds(
   backendId: string,
 ): Promise<ListBuildsResponse> {
   const name = `projects/${projectId}/locations/${location}/backends/${backendId}/builds`;
-  let pageToken: string | undefined = undefined;
-  const res: ListBuildsResponse = {
+  let pageToken: string | undefined;
+  const res: implements ListBuildsResponse = {
     builds: [],
     unreachable: [],
   };
@@ -365,8 +365,8 @@ export async function listBuilds(
   do {
     const queryParams: Record<string, string> = pageToken ? { pageToken } : {};
     const int = await client.get<ListBuildsResponse>(name, { queryParams });
-    res.builds.splice(res.builds.length, 0, ...(int.body.builds || []));
-    res.unreachable?.splice(res.unreachable.length, 0, ...(int.body.unreachable || []));
+    res.builds.push(...(int.body.builds || []));
+    res.unreachable.push(...(int.body.unreachable || []));
     pageToken = int.body.nextPageToken;
   } while (pageToken);
 
@@ -520,7 +520,6 @@ export async function ensureApiEnabled(options: any): Promise<void> {
 /**
  * Generates the next build ID to fit with the naming scheme of the backend API.
  * @param counter Overrides the counter to use, avoiding an API call.
- * @return
  */
 export async function getNextRolloutId(
   projectId: string,
@@ -531,7 +530,7 @@ export async function getNextRolloutId(
   const date = new Date();
   const year = date.getUTCFullYear();
   // Note: month is 0 based in JS
-  const month = String(date.getUTCMonth()).padStart(2, "0");
+  const month = String(date.getUTCMonth() + 1).padStart(2, "0");
   const day = String(date.getUTCDay()).padStart(2, "0");
 
   if (counter) {
--- a/src/gcp/apphosting.ts
+++ b/src/gcp/apphosting.ts
@@ -346,26 +346,19 @@ export async function listBuilds(
   backendId: string,
 ): Promise<ListBuildsResponse> {
   const name = `projects/${projectId}/locations/${location}/backends/${backendId}/builds`;
-  let nextPageToken: string | null = null;
+  let pageToken: string | undefined = undefined;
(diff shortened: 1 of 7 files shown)


src/gcp/apphosting.ts
```diff
@@ -334,6 +347,33 @@ export async function getBuild(
...
+  const res: ListBuildsResponse = {
+    builds: [],
+    unreachable: [],
+  };
+
+  do {
+    const queryParams: Record<string, string> = pageToken ? { pageToken } : {};
+    const int = await client.get<ListBuildsResponse>(name, { queryParams });
+    res.builds.splice(res.builds.length, 0, ...(int.body.builds || []));
+    res.unreachable?.splice(res.unreachable.length, 0, ...(int.body.unreachable || []));
```
bkendall, line 369: Is it cleaner to use `.push`?

```suggestion
    res.builds.push(...(int.body.builds || []));
    res.unreachable?.push(...(int.body.unreachable || []));
```

src/gcp/apphosting.ts
```diff
@@ -389,11 +429,24 @@ export async function listRollouts(
...
+  const res: ListRolloutsResponse = {
+    rollouts: [],
+    unreachable: [],
+  };
+
+  do {
+    const queryParams: Record<string, string> = pageToken ? { pageToken } : {};
+    const int = await client.get<ListRolloutsResponse>(name, { queryParams });
+    res.rollouts.splice(res.rollouts.length, 0, ...(int.body.rollouts || []));
+    res.unreachable?.splice(res.unreachable.length, 0, ...(int.body.unreachable || []));
```
bkendall, line 444: Similar to above... is `push` more clear?

src/gcp/apphosting.ts
```diff
@@ -334,6 +347,33 @@ export async function getBuild(
...
+/**
+ * List Builds by backend
+ */
+export async function listBuilds(
+  projectId: string,
+  location: string,
+  backendId: string,
+): Promise<ListBuildsResponse> {
+  const name = `projects/${projectId}/locations/${location}/backends/${backendId}/builds`;
+  let pageToken: string | undefined = undefined;
```
bkendall, line 359: Technically I don't think you need to specify `= undefined` - that can be dropped.

```suggestion
  let pageToken: string | undefined;
```

src/gcp/apphosting.ts
```diff
@@ -460,3 +516,53 @@ export async function ensureApiEnabled(options: any): Promise<void> {
   const projectId = needProjectId(options);
   return await ensure(projectId, API_HOST, "frameworks", true);
 }
+
+/**
+ * Generates the next build ID to fit with the naming scheme of the backend API.
+ * @param counter Overrides the counter to use, avoiding an API call.
+ * @return
```
bkendall, line 523: I _think_ you can drop an empty `@return` line - though I'm not certain

src/gcp/apphosting.ts
```diff
@@ -460,3 +516,53 @@ export async function ensureApiEnabled(options: any): Promise<void> {
...
+export async function getNextRolloutId(
+  projectId: string,
+  location: string,
+  backendId: string,
+  counter?: number,
+): Promise<string> {
+  const date = new Date();
+  const year = date.getUTCFullYear();
+  // Note: month is 0 based in JS
+  const month = String(date.getUTCMonth()).padStart(2, "0");
```
bkendall, line 534: So should you be adding 1?

src/gcp/apphosting.ts
```diff
@@ -460,3 +516,53 @@ export async function ensureApiEnabled(options: any): Promise<void> {
...
+  const day = String(date.getUTCDay()).padStart(2, "0");
+
+  if (counter) {
+    return `build-${year}-${month}-${day}-${String(counter).padStart(3, "0")}`;
+  }
+
+  // Note: must use exports here so that listRollouts can be stubbed in tests.
+  const builds = await (exports as { listRollouts: typeof listRollouts }).listRollouts(
```
bkendall, line 542: Huh, but `exports.listRollouts` gives you type errors or something? That's unfortunate :(

src/test/gcp/apphosting.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+    });
+
+    function idPrefix(date: Date): string {
+      const year = date.getUTCFullYear();
+      const month = String(date.getUTCDay()).padStart(2, "0");
+      const day = String(date.getUTCDay()).padStart(2, "0");
+      return `build-${year}-${month}-${day}`;
+    }
+
+    it("Should handle explicit counters", async () => {
```
bkendall, line 24: These generally shouldn't be capitalized (though I'm not sure I have any style guidance to point to). When I see this, I read it as a sentence, and "it Should do this" looks abnormal (compared to "it should do this")

bkendall (approved):
style/cleanup comments, but looks fine to me

src/gcp/apphosting.ts
inlined: 😱 I modified the code and dropped the +1. Good catch and thank god for comments.

PR #6754, opened 2024-02-08
FAH bugfixes

inlined:
1. Fixes issue where the --branch param on builds:create is ignored
2. Fixes issue where the build ID uses day of week instead of day of month

### DIFF
--- a/src/commands/apphosting-builds-create.ts
+++ b/src/commands/apphosting-builds-create.ts
@@ -16,12 +16,12 @@ export const command = new Command("apphosting:builds:create <backendId>")
     const buildId =
       (options.buildId as string) ||
       (await apphosting.getNextRolloutId(projectId, location, backendId));
-    const branch = options.branch as string;
+    const branch = (options.branch as string | undefined) ?? "main";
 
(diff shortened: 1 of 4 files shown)


src/commands/apphosting-builds-create.ts
```diff
@@ -16,12 +16,12 @@ export const command = new Command("apphosting:builds:create <backendId>")
     const buildId =
       (options.buildId as string) ||
       (await apphosting.getNextRolloutId(projectId, location, backendId));
-    const branch = options.branch as string;
+    const branch = (options.branch as string | undefined) ?? "main";
```
bkendall, line 19: A really neat improvement to this would be to see if they have a RolloutPolicy and pull the "default" branch from there. But that can come later.

bkendall (approved):

src/commands/apphosting-builds-create.ts
inlined: File a bug and assign it to me. I agree that would be much better, but IIRC rollouts is still an "internal tool" that hasn't gone through API review.

PR #6766, opened 2024-02-13
Disable switching Firebase project when emulators are running

rrousselGit:


hlshen (approved):
Let's get a test for this as well

PR #6768, opened 2024-02-13
fixes iterating issue when listing no turtles

bkendall:
### Description

When listing turtles, if there are none, we need to have _something_ to try and iterate over, otherwise it breaks.

### DIFF
--- a/src/commands/apphosting-backends-list.ts
+++ b/src/commands/apphosting-backends-list.ts
@@ -28,7 +28,7 @@ export const command = new Command("apphosting:backends:list")
       );
     }
 
-    const backends = backendRes?.backends;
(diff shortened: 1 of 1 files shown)


blidd-google (approved):

src/commands/apphosting-backends-list.ts
```diff
@@ -28,7 +28,7 @@ export const command = new Command("apphosting:backends:list")
       );
     }
 
-    const backends = backendRes?.backends;
+    const backends = backendRes.backends ?? [];
```
rrousselGit, line 31: This change feels weird. `backendRes.backends` is non-nullable, so technically that `??` is unreachable

Maybe there's a type issue here.

PR #6769, opened 2024-02-13
silence the turtles

bkendall:
### Description

The turtles were being a little too vocal.


inlined (approved):
We should also do a rename to "apphosting" now that we have an approved name. That was supposed to be a rapid follow-up after M2.

PR #6775, opened 2024-02-16
Light control flow cleanups within backends:create

tonyjhuang:

### DIFF
--- a/src/init/features/apphosting/index.ts
+++ b/src/init/features/apphosting/index.ts
@@ -29,7 +29,7 @@ const apphostingPollerOptions: Omit<poller.OperationPollerOptions, "operationRes
 /**
  * Set up a new App Hosting backend.
  */
-export async function doSetup(setup: any, projectId: string): Promise<void> {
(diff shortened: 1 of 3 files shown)


src/init/features/apphosting/index.ts
```diff
@@ -117,92 +94,103 @@ export async function doSetup(setup: any, projectId: string): Promise<void> {
...
-    default: DEFAULT_REGION,
-    message:
-      "Please select a region " +
-      `(${clc.yellow("info")}: Your region determines where your backend is located):\n`,
-    choices: locations.map((loc) => ({ value: loc })),
-  })) as string;
+/**
+ * Prompts the user for a backend id and verifies that it doesn't match a pre-existing backend.
+ */
+async function promptNonexistentBackend(projectId: string, location: string): Promise<string> {
```
blidd-google, line 112: *nit: `promptNewBackendId`? Nonexistent might imply we're not about to create a new backend with that ID, but honestly works fine too

PR #6764, opened 2024-02-12
Improve FAH onboarding flow to connect backends with SCMs

blidd-google:
This PR fixes a number of different issues with connecting repos to FAH backends:

* Lists all accessible repos (not limited to 20 anymore) and enables autocomplete prompting
* Links to the FAH dashboard at the end of the create flow
* Allows users to link additional GitHub accounts / orgs and connect their repos

PR #6781, opened 2024-02-17
turtle onboarding fixes

bkendall:
### Description

Fixed up a couple things:

1. use `validateOnly` to pre-flight the `CreateRollout` call to not hit issues where the build isn't immediately available
1. fixes `:delete` command to require the backend ID as an argument, not a flag


src/gcp/cloudbuild.ts
```diff
@@ -158,10 +158,18 @@ export async function fetchLinkableRepositories(
   projectId: string,
   location: string,
   connectionId: string,
-): Promise<LinkableRepositories> {
+): Promise<Repository[]> {
   const name = `projects/${projectId}/locations/${location}/connections/${connectionId}:fetchLinkableRepositories`;
-  const res = await client.get<LinkableRepositories>(name);
-  return res.body;
+  let pageToken = "";
```
tonyjhuang, line 163: do you mind reverting this file? this will be addressed very shortly in https://github.com/firebase/firebase-tools/pull/6764/files#diff-2d40d8ede700bdabb5e1e6f7141916628fd1dfc44a55e68582be14ad1cc44398

we might as well avoid the merge conflict now

src/init/features/apphosting/index.ts
```diff
@@ -212,9 +213,32 @@ export async function onboardRollout(
...
+  };
+
+  let tries = 0;
+  let done = false;
+  while (!done) {
+    tries++;
+    try {
+      await apphosting.createRollout(projectId, location, backendId, buildId, rolloutBody, true);
+      done = true;
+    } catch (err: unknown) {
```
tonyjhuang, line 227: can we be more specific with the error code here?

src/gcp/cloudbuild.ts
blidd-google: Yes apologies if it was unclear that I was including this in my PR

src/init/features/apphosting/repo.ts
```diff
@@ -135,9 +135,9 @@ async function promptRepositoryUri(
   const remoteUriToConnection: Record<string, gcb.Connection> = {};
   for (const conn of connections) {
     const { location, id } = parseConnectionName(conn.name)!;
-    const resp = await gcb.fetchLinkableRepositories(projectId, location, id);
-    if (resp.repositories && resp.repositories.length > 0) {
-      for (const repo of resp.repositories) {
+    const repos = await gcb.fetchLinkableRepositories(projectId, location, id);
```
blidd-google, line 138: We're probably gonna run into some merge conflicts here, I reworked the logic here as well (https://github.com/firebase/firebase-tools/pull/6764/files#diff-a82ad0512cae837b1f50b301e857637957d3b8dc5fb4d2c8095b7a2853857179)

src/init/features/apphosting/index.ts
```diff
@@ -212,9 +213,32 @@ export async function onboardRollout(
...
     build: `projects/${projectId}/locations/${location}/backends/${backendId}/builds/${buildId}`,
-  });
+  };
+
+  let tries = 0;
+  let done = false;
+  while (!done) {
+    tries++;
+    try {
+      await apphosting.createRollout(projectId, location, backendId, buildId, rolloutBody, true);
```
tonyjhuang, line 225: can we include a arg comment /* validateOnly= */ true

src/init/features/apphosting/index.ts
```diff
@@ -212,9 +213,32 @@ export async function onboardRollout(
   const buildId = await apphosting.getNextRolloutId(projectId, location, backendId, 1);
   const buildOp = await apphosting.createBuild(projectId, location, backendId, buildId, buildInput);
 
-  const rolloutOp = await apphosting.createRollout(projectId, location, backendId, buildId, {
+  const rolloutBody = {
     build: `projects/${projectId}/locations/${location}/backends/${backendId}/builds/${buildId}`,
-  });
+  };
+
+  let tries = 0;
```
tonyjhuang, line 220: This only gives us ~5 seconds of total delay (slightly more depending on how long CreateRollout takes). Is this enough? Perhaps we should consider more retries or adding exponential backoff

bkendall: It's been enough this far. Working on improving it on the backend as well.

src/init/features/apphosting/index.ts
bkendall: Blurg. I hate those. Named the value though, which helps.

src/init/features/apphosting/index.ts
bkendall: I don't know what you mean.

`unknown` is allowed and expected - you can't be more specific else you get an error  like this: `Catch clause variable type annotation must be 'any' or 'unknown' if specified.`

Can I check to see if it's an expected error for this scenario? Sure.

tonyjhuang (approved):

PR #6797, opened 2024-02-20
Provision default compute service account

tonyjhuang:
### Description

Creates the default compute service account if it doesn't exist when creating new backends. Also adds a --service-account flag to override the default behavior.


blidd-google (approved):
LGTM

PR #6836, opened 2024-03-05
Respect GOOGLE_CLOUD_QUOTA_PROJECT env var when it is set

### LINKED ISSUE #6801: Add support for `GOOGLE_CLOUD_QUOTA_PROJECT` across Firebase API client
We're running into an issue running `firebase-tools deploy` in Google Cloud Build, when invoking the command inside a build step owned by one GCP project, deploying to Firebase owned by another GCP project.

Essentially, the Firebase Management API isn't enabled in the project that owns the build runner, but it is enabled in the project that owns the target of the `firebase-tools deploy` command.

When `firebase-tools deploy` performs its pre-deploy checks, it uses the quota project of the Google Cloud Build environment, not the project of the target deployment.

### Environment info

We are running `firebase-tools` inside Google Cloud Build (Ubuntu).

### Expected behavior

I believe the `GOOGLE_CLOUD_QUOTA_PROJECT` environment variable needs to be honored by `firebase-tools` base API client. Under the hood, I believe what this does is it sets the `x-goog-user-project` on the outgoing requests to various Google APIs, so those requests are associated with the project of my choice (and not necessarily the environment).

### Actual behavior

`GOOGLE_CLOUD_QUOTA_PROJECT` isn't supported, and the project associated with the environment (or its service account) is used for quota/billing to API requests instead of the desired target project.

This results in this error:

```
npx firebase-tools@latest deploy --debug --project $TARGET_PROJECT
...
[2024-02-21T00:36:31.601Z] <<< [apiv2][body] GET https://firebase.googleapis.com/v1beta1/projects/$TARGET_PROJECT {"error":{"code":403,"message":"Firebase Management API has not been used in project $SOURCE_PROJECT_NUMBER before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/firebase.googleapis.com/overview?project=$SOURCE_PROJECT_NUMBER then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.","status":"PERMISSION_DENIED","details":[{"@type":"type.googleapis.com/google.rpc.Help","links":[{"description":"Google developers console API activation","url":"https://console.developers.google.com/apis/api/firebase.googleapis.com/overview?project=587340091677"}]},{"@type":"type.googleapis.com/google.rpc.ErrorInfo","reason":"SERVICE_DISABLED","domain":"googleapis.com","metadata":{"consumer":"projects/$SOURCE_PROJECT_NUMBER","service":"firebase.googleapis.com"}}]}}
````

joehan:
### Description
When GOOGLE_CLOUD_QUOTA_PROJECT is set, we should set `x-goog-user-project` on outgoing API requests.

This is a common standard accross many google libraries (https://cloud.google.com/docs/quotas/set-quota-project), and we've had a few different cases where this would be useful. It also fixes #6801.

However, I don't have the most thorough understand of how the various x-goog headers work, and whether this is the correct/narrowest one to set here. I'd love a review from someone who is well versed in this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
 - Released Firestore emulator v1.19.2, which fixes some bugs affecting client SDKs when in Datastore Mode.
 - Fix demo projects + web frameworks with emulators (#6737)
 - Fix Next.js static routes with server actions (#6664)
+- Fixed an issue where `GOOGLE_CLOUD_QUOTA_PROJECT` was not correctly respected. (#6801)
--- a/src/apiv2.ts
+++ b/src/apiv2.ts
@@ -18,7 +18,10 @@ import * as FormData from "form-data";
 const pkg = require("../package.json");
 const CLI_VERSION: string = pkg.version;
 
-const GOOG_QUOTA_USER = "x-goog-quota-user";
+const GOOG_QUOTA_USER_HEADER = "x-goog-quota-user";
+
+const GOOG_USER_PROJECT_HEADER = "x-goog-user-project";
+const GOOGLE_CLOUD_QUOTA_PROJECT = process.env.GOOGLE_CLOUD_QUOTA_PROJECT;
 
 export type HttpMethod = "GET" | "PUT" | "POST" | "DELETE" | "PATCH" | "OPTIONS" | "HEAD";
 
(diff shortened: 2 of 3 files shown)


joehan:
> You might want to look for `x-goog-user-project` in the codebase and think about the behavior we want - for instance I manually set it to the current project for auth recently but maybe the env var should override that?

I think you were actually setting x-goog-quota-user  (https://cloud.google.com/apis/docs/system-parameters#definitions), which is closely related but only affects quota (x-goog-user-project affects quota and billing). I _think_ x-goog-user-project is the right one to use here for consistency with other tools.

Either way, that is the current behavior here - we always overwrite any heades provided at the call site with this value.

Issue #6814, opened 2024-02-25
Can not deploy gen2 trigger cloud functions with custom service account


sconix:
### Environment info

**firebase-tools:** 13.3.1

**Platform:** Debian

### Test case

```javascript
export const automatedBackupFromFirestoreV2 = onSchedule(
  {
    region: 'europe-west1',
    schedule: 'every 168 hours',
    serviceAccount: 'compute-engine-default@*****.iam.gserviceaccount.com',
  },
  async () => {});
```

This happens also with onDocument* firestore triggers, basically anything that uses EventArc. GCloud tool seems to have separate trigger-service-account input argument so it is at least possible to define the service account to be used, but can not get it to work with firebase cloud functions. Have tried using the setGlobalOptions as well, but no luck with that neither.

This used to be an issue with the onCall and onRequest as well, but those now work with the latest version of firebase-tools.

### Steps to reproduce

Have any kind of trigger cloud function and configure it to use custom service account and have a project that does not have the default compute engine service account which EventArc uses when no service account is defined.

### Expected behavior

Deploy works without the error about missing default compute engine service account and the defined service account is used.

### Actual behavior

```bash
[2024-02-25T18:58:14.315Z] Functions deploy failed.
[2024-02-25T18:58:14.315Z] {
  "endpoint": {
    "id": "automatedBackupFromFirestoreV2",
    "project": "*****",
    "region": "europe-west1",
    "entryPoint": "automatedBackupFromFirestoreV2",
    "platform": "gcfv2",
    "runtime": "nodejs18",
    "scheduleTrigger": {
      "schedule": "every 168 hours",
      "timeZone": null,
      "retryConfig": {}
    },
    "labels": {
      "deployment-tool": "cli-firebase"
    },
    "ingressSettings": null,
    "availableMemoryMb": null,
    "serviceAccount": "compute-engine-default@*****.iam.gserviceaccount.com",
    "timeoutSeconds": null,
    "maxInstances": null,
    "minInstances": null,
    "concurrency": 80,
    "vpc": null,
    "environmentVariables": {
      "BACKUPS_BUCKET": "\"gs://*****",
      "FIREBASE_CONFIG": "{\"projectId\":\"*****\",\"databaseURL\":\"*****\",\"storageBucket\":\"*****\",\"locationId\":\"europe-west\"}",
      "GCLOUD_PROJECT": "*****",
[...]

PR #6859, opened 2024-03-08
Propogate explicit service account to eventarc

inlined:
Another part of the fix to #6814. Similar to #6858, this propagates the function's service account through to the invoking service.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+v2 event functions with explicit service accounts trigger eventarc to use that service account (#6859)
--- a/src/gcp/cloudfunctionsv2.ts
+++ b/src/gcp/cloudfunctionsv2.ts
@@ -550,6 +550,9 @@ export function functionFromEndpoint(endpoint: backend.Endpoint): InputCloudFunc
       eventType: endpoint.eventTrigger.eventType,
       retryPolicy: "RETRY_POLICY_UNSPECIFIED",
     };
+    if (endpoint.serviceAccount) {
+      gcfFunction.eventTrigger.serviceAccountEmail = endpoint.serviceAccount;
+    }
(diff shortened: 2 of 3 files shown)


colerogers (approved):

inlined:
This is unfortunate. I can go through the process of creating an API proposal to have two separate fields. As hard-coding a possibly non-existent service account might have been considered a bug, we may instead roll forward where the trigger service account defaults to the service account used within the container and can be overridden. 

Can you help me understand your use case a bit better? I assume you were using a lower-privileged service account in your functions to ensure bugs don't have a large blast radius?

aablsk:
Dear @inlined thank you for your immediate response! 🙇

We would absolutely appreciate the API being amended to have two separate fields.

The motivation for multiple service accounts is exactly as you assumed to adhere to best practices and minimise any potential bugs or security incidents have highly limited blast radius (e.g. the service account for the CloudRun service would usually not have permissions to invoke itself). Additionally, it makes it easier to manage/maintain the service accounts permission if the service account is more narrowly scoped thus further reducing the risk of unnecessary permissions that are kept due to uncertainty of them being in use.

**Our setup**
Our use-case involves deploying our firebase functions with firebase-cli and then adjusting parameters that cannot be set through firebase with terraform. firebase functions are imported into terraform after the first deployment to have them in IaC as well and allow us to set parameters that cannot be set through the firebase function config. This is how we used to set the service account for the Event Arc trigger (and this would still hypothetically work). We also import the corresponding pub/sub subscription into Terraform so we can set e.g. dead letter queue behavior.

Unfortunately, when changing the service account of a Event Arc Trigger for a function it needs to be re-created, which also means that the corresponding Pub/Sub topic and Pub/Sub subscription are re-created. This means that we either would have to re-import the subscription on every deployment or we need to adhere to how firebase-cli manages the services accounts (one service account for both trigger and the service).

Thank you for your time and consideration @inlined! 🙇

Issue #6759, opened 2024-02-09
Error Adjusting Traffic to Previous Cloud Run Revisions Deployed via Firebase CLI(Cloud Functions)


maylorsan:
### Environment info
**firebase-tools:** 13.2.1
**Platform:** macOS 14.0 (23A344)

### Test case
A Cloud Run service deployed via Cloud Functions, where deployments are triggered through the Firebase CLI. The service experiences failures when attempting to route traffic back to previous revisions, receiving errors related to fetching container image metadata from Google Artifact Registry.

### Steps to reproduce
1. Deploy Cloud Functions that trigger updates to a Cloud Run service using the Firebase CLI with the following command:

```
firebase deploy
```

2. Confirm the deployment is successful and the new revision is serving traffic.
3. Attempt to adjust traffic back to a previous revision using either the Google Cloud Console or the gcloud CLI:
```
gcloud run services update-traffic "SERVICE_NAME" --region "europe-central2" --to-revisions "PREVIOUS_REVISION=100"
```
4. Observe the error message indicating the failure due to the inability to fetch metadata for the container image from the Artifact Registry.

### Expected behavior
It should be possible to adjust traffic to any revision of the Cloud Run service deployed via Firebase CLI without encountering errors related to container image imports. Adjusting traffic should be seamless across revisions, allowing for flexible traffic management and rollback capabilities.

### Actual behavior
When trying to adjust traffic to a previous revision of a Cloud Run service deployed via Firebase CLI, the operation fails with an error related to fetching metadata for the container image from Google Artifact Registry. The error specifically states that the container image cannot be imported because the metadata cannot be fetched, implying that the image or its manifest could not be found.

**Error Message:**
[...]

PR #6861, opened 2024-03-08
Make GCF artifact cleanup guarded by an experiment

inlined:
DO NOT MERGE UNTIL API COUNCIL HAS APPROVED THE CHANGE

Adds a default-on experiment that can be turned off to prevent functions from cleaning up old images. Will unblock customers unhappy in #6759 

With this change, you can disable image cleanup with:

```
firebase experiments:disable deletegcfartifactsondeploy
```


colerogers (approved):

petergrau:
We really would like to have this feature b/c at the moment we cannot re-route traffic to old deployments. Can you make an estimate when it will be merged? @inlined?

inlined:
The API council can take some time. We decided to rename the experiment so that "deletegcfartifacts" and "automaticallydeletegcfartifacts" make sense together. Going to merge this PR imminently and then it will be part of next week's CLI release

Issue #6774, opened 2024-02-15
Functions Python version conflict 3.12 and 3.11


curtis-jotson:
### [REQUIRED] Environment info

**firebase-tools: 13.3.0**

**Platform: macOS, Ubuntu (Github Actions)**

### [REQUIRED] Test case

Fresh Python functions project with both Python 3.11 and 3.12 installed

### [REQUIRED] Steps to reproduce

1. Start a new (or use existing) Firebase project and init Python functions
2. Install Python 3.12
3. Build the venv
4. Install pip requirements (firebase-functions, firebase-admin)
5. Build a function
6. Attempt to run the emulators and it complains that you're not using Python 3.11
7. Install Python 3.11
8. Build the venv using 3.11
9. Install pip requirements using 3.11
10. Emulators run
11. Attempt to deploy the functions and it complains that you're not using Python 3.12

### [REQUIRED] Expected behavior

Emulators should work with Python 3.12 and deploy should work with Python 3.11 when both are installed.

### [REQUIRED] Actual behavior

Emulators fail to start the Python functions if the venv was setup using 3.12
Deploy fails when venv was setup using 3.11 if 3.12 is also installed

antopolskiy:
I've got the same issue: since firebase-tools switched to version 13.3.0, my Github Actions builds fail to deploy Python Cloud functions (I am using python3.11, and with firebase-tools@13.2.1 it was working fine).

```
Run npx firebase-tools deploy --only functions --force --json --project *** --token ***
  
npm WARN exec The following package was not found and will be installed: firebase-tools@13.3.0
***
  "status": "error",
  "error": "Failed to find location of Firebase Functions SDK. Did you forget to run '. \"/home/runner/work/webapp/webapp/functions-python/venv/bin/activate\" && python3.12 -m pip install -r requirements.txt'?"
*** 
```

curtis-jotson:
@antopolskiy So far our fix has been to pin `firebase-tools` to 13.2.1 for now.

kmsbernard:
I got the same error and found a fix. 

This error occurred when you omitted `runtime` config in `firebase.json`. If your runtime uses python 3.11, it should looks like following:
```json
{
  "functions": [
    {
      "source": "functions",
      "codebase": "default",
      "runtime": "python311",  <--- here

    }
  ],
}
```

If the config is omitted, `firebase-tools` attempt to find latest python version. 
[...]

PR #6866, opened 2024-03-10
Strengthen typing of Runtimes+Langauges and their support timelines.

inlined:
- Factors out Runtimes into a new file so that they can be imported into firebaseConfig.ts and have a single source of truth and prevent the error encountered in https://github.com/firebase/firebase-tools/issues/6774#issuecomment-1953399977
- Adds a formal concept of languages and the ability to discriminate runtimes per language
- Makes all information about a runtime table driven so there can never be a runtime without all necessary metadata
- Adds the ability to dynamically fetch the latest runtime for a language
- Adds support timelines for all runtimes
- Unifies runtime support checks across all runtimes/languages. There are now separate warnings for when deprecation is upcoming (90d), when a runtime is deprecated, and when a runtime is decommissioned.


src/deploy/functions/runtimes/python/index.ts
```diff
@@ -30,9 +29,15 @@ export async function tryCreateDelegate(
     logger.debug("Customer code is not Python code.");
     return;
   }
-  const runtime = context.runtime ? context.runtime : LATEST_VERSION;
-  if (!runtimes.isValidRuntime(runtime)) {
-    throw new FirebaseError(`Runtime ${runtime} is not a valid Python runtime`);
+  const runtime = context.runtime ? context.runtime : supported.latest("python");
```
joehan, line 32: Nit: `context.runtime ?? supported.latest("python")` is probably cleaner than a ternary

src/test/deploy/functions/checkIam.spec.ts
```diff
@@ -21,7 +21,7 @@ const BINDING = {
 const SPEC = {
   region: "us-west1",
   project: projectNumber,
-  runtime: "nodejs14",
+  runtime: "nodejs14" as const,
```
joehan, line 24: TIL that `as const` satisfies enums without having to import them - good to know!

joehan (approved):
MInor nits, but lgtm!

src/test/deploy/functions/checkIam.spec.ts
inlined: yes, "as const" will make a string literal type rather than just a string type.

PR #6875, opened 2024-03-12
Add web app selection to app hosting onboarding

mathu97:
### Description

Adds a prompt during App Hosting onboarding to make sure user enters a Firebase web app that they would like to associate their Backend with.

### Sample Commands

For interactive app selection:

- `firebase apphosting:backends:create --project=<project-name>`

Otherwise: 

- `firebase apphosting:backends:create --app="<firebase-web-app>" --project=<project-name>`
- `firebase apphosting:backends:create -a <firebase-web-app> --project=<project-name>`

**Note: Don't merge till backend appId field is supported by App Hosting backend**


src/init/features/apphosting/index.ts
```diff
@@ -62,7 +66,30 @@ export async function doSetup(
     }
   }
 
-  logBullet("First we need a few details to create your backend.");
+  logBullet("First we need a few details to create your backend.\n");
+
+  let webAppId: string;
+
+  const existingUserProjectWebApps = new Map(
+    (await listFirebaseApps(projectId, AppPlatform.WEB)).map((obj) => [
```
mathu97, line 74: this lists up to a max of 100 apps, do we need to consider the case of a user having > 100 apps in a project?

src/init/features/apphosting/index.ts
```diff
@@ -62,7 +66,30 @@ export async function doSetup(
...
+      // displayName can be null, use name instead if so. Example - displayName: "mathusan-web-app", name: "projects/mathusan-fwp/webApps/1:461896338144:web:426291191cccce65fede85"
+      obj.displayName ?? obj.name,
+      obj.appId,
+    ]),
+  );
+
+  if (webAppName) {
+    if (!existingUserProjectWebApps.get(webAppName)) {
+      throw new FirebaseError(`The web app '${webAppName}' does not exist in project ${projectId}`);
```
tonyjhuang, line 83: Let's move all of this to a helper so that the top level control flow remains clear

src/init/features/apphosting/index.ts
```diff
@@ -132,6 +160,56 @@ export async function doSetup(
   logSuccess(`Your site is now deployed at:\n\thttps://${backend.uri}`);
 }
 
+/**
+ * Prompts the user for the web app that they would like to associate their backend with
+ * @param projectId user's projectId
+ * @param existingUserProjectWebApps a map of a user's web apps to their ids
+ * @return the name and ID of a web app
+ */
+async function promptFirebaseWebAppId(
```
tonyjhuang, line 169: We should create a web app if there are none, are you planning on adding that in another PR?

mathu97: not that's implemented here, see: https://github.com/firebase/firebase-tools/pull/6875/files#diff-1f7254b41cdcf666adebb8a934648480cbb16a8abffe6145254f564aed2ec315R198

mathu97: I actually just implemented that functionality: https://github.com/firebase/firebase-tools/pull/6875/files#diff-1f7254b41cdcf666adebb8a934648480cbb16a8abffe6145254f564aed2ec315R168

src/init/features/apphosting/index.ts
```diff
@@ -132,6 +144,96 @@ export async function doSetup(
...
+async function getOrCreateWebAppId(
+  projectId: string,
+  webAppName: string | null,
+  backendId: string,
+): Promise<string[]> {
+  let webAppId: string;
+
+  const existingUserProjectWebApps = new Map(
+    (await listFirebaseApps(projectId, AppPlatform.WEB)).map((obj) => [
+      // displayName can be null, use name instead if so. Example - displayName: "mathusan-web-app", name: "projects/mathusan-fwp/webApps/1:461896338144:web:426291191cccce65fede85"
```
tonyjhuang, line 162: I'm sure this will be addressed in the API review but it may be better to print out just the actual name of the web app, ie `1:461896338144:web:426291191cccce65fede85`

src/init/features/apphosting/index.ts
```diff
@@ -132,6 +144,96 @@ export async function doSetup(
...
+      obj.displayName ?? obj.name,
+      obj.appId,
+    ]),
+  );
+
+  if (existingUserProjectWebApps.size < 1) {
```
tonyjhuang, line 168: == 0?

src/init/features/apphosting/index.ts
```diff
@@ -132,6 +144,96 @@ export async function doSetup(
...
+ *
+ * @param projectId user's projectId
+ * @param webAppName (optional) name of web app
+ * @return app name and app id
+ */
+async function getOrCreateWebAppId(
+  projectId: string,
+  webAppName: string | null,
+  backendId: string,
+): Promise<string[]> {
```
tonyjhuang, line 157: how about instead of Promise<string[]>, type WebApp = [string, string];

or even better, type WebApp = {name: string, id: string}

src/init/features/apphosting/index.ts
```diff
@@ -132,6 +144,96 @@ export async function doSetup(
...
+  );
+
+  if (existingUserProjectWebApps.size < 1) {
+    // create a web app using backend id
+    const newWebApp = await createWebApp(projectId, { displayName: backendId });
+    return [newWebApp.displayName, newWebApp.appId];
+  }
+
+  if (webAppName) {
+    if (!existingUserProjectWebApps.get(webAppName)) {
```
tonyjhuang, line 175: prefer `=== undefined` to avoid js equality bs

src/init/features/apphosting/index.ts
```diff
@@ -132,6 +144,96 @@ export async function doSetup(
...
+    ]),
+  );
+
+  if (existingUserProjectWebApps.size < 1) {
+    // create a web app using backend id
+    const newWebApp = await createWebApp(projectId, { displayName: backendId });
+    return [newWebApp.displayName, newWebApp.appId];
+  }
+
+  if (webAppName) {
```
tonyjhuang, line 174: test coverage?

tonyjhuang (changes requested):

src/init/features/apphosting/index.ts
```diff
@@ -27,8 +27,12 @@ import { DEFAULT_REGION } from "./constants";
 import { ensure } from "../../../ensureApiEnabled";
 import * as deploymentTool from "../../../deploymentTool";
 import { DeepOmit } from "../../../metaprogramming";
+import { AppPlatform, createWebApp, listFirebaseApps } from "../../../management/apps";
+import * as inquirer from "inquirer";
+import * as fuzzy from "fuzzy";
 
 const DEFAULT_COMPUTE_SERVICE_ACCOUNT_NAME = "firebase-app-hosting-compute";
+const CREATE_NEW_WEB_APP_CHOICE = "CREATE_NEW_WEB_APP";
```
qtz1, line 35: nit: can we call this "CREATE_NEW_FIREBASE_WEB_APP"? Web app is an overloaded term and there was a time when we referred to app hosting backends as web apps (and sometimes we still do)

qtz1 (changes requested):
Left a couple comments but lgtm

src/init/features/apphosting/index.ts
mathu97: do you mean we show the `appId` instead of the `appName` when `displayName` isn't available?

src/init/features/apphosting/index.ts
mathu97: yep this is much better!

src/init/features/apphosting/index.ts
mathu97: added

src/init/features/apphosting/index.ts
mathu97: ah good call out! Updated the naming here.

src/init/features/apphosting/index.ts
```diff
@@ -132,6 +144,96 @@ export async function doSetup(
...
+    message:
+      "Which of the following Firebase web apps would you like to associate your backend with?",
+    source: searchWebApps(Array.from(existingUserProjectWebApps.keys())),
+  });
+
+  if (webAppName === CREATE_NEW_WEB_APP_CHOICE) {
+    const newAppDisplayName = await promptOnce({
+      type: "input",
+      name: "webAppName",
+      message: "Enter a unique name for your web app",
```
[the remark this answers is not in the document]
mathu97: ya you're right! This isn't a requirement. It seems even if there are existing projects we should allow the user to create a new app using the given backendId. However, what happens when there is an existing project with the backendId that they provide? Doesn't seem great to just error out. 

cc @tonyjhuang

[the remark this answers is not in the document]
tonyjhuang: Let's run this flow by Julia to see if we can land on flow that works for both CLI and Console

tonyjhuang (commented):
LGTM pending product confirmation

src/init/features/apphosting/index.ts
tonyjhuang: I actually meant `1:461896338144:web:426291191cccce65fede85` instead of `projects/mathusan-fwp/webApps/1:461896338144:web:426291191cccce65fede85`

src/init/features/apphosting/index.ts
```diff
@@ -85,6 +85,18 @@ export async function doSetup(
...
+  let webAppId: string | undefined;
+
+  try {
+    const webApp = await apps.getOrCreateWebApp(projectId, webAppName, backendId);
+    if (webApp) {
+      webAppId = webApp.id;
+      logSuccess(`Firebase web app set to ${webApp.name}.\n`);
+    }
+  } catch (e) {
+    logWarning("Unable to associate your App Hosting backend to a Firebase web app.");
```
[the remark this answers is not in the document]
tonyjhuang: Is this still the intended product behavior? ie if the backend returns a quota error, we're currently failing open and proceeding with backend creation

tonyjhuang (changes requested):

src/init/features/apphosting/index.ts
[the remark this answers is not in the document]
mathu97: no i missed this, thanks for catching!

tonyjhuang (approved):

PR #6886, opened 2024-03-18
Simplify GitHub repo connect flow

tonyjhuang:
### Description

Removes the "configure existing connection" choice from the repo picker. 

Also aims to simplify the top level `linkGitHubRepository()` by adding some helper methods

### DIFF
--- a/src/init/features/apphosting/repo.ts
+++ b/src/init/features/apphosting/repo.ts
@@ -102,8 +102,8 @@ export async function linkGitHubRepository(
     );
   }
 
-  var repoRemoteUri: string | undefined;
-  var connection: gcb.Connection;
+  let repoRemoteUri: string | undefined;
+  let connection: gcb.Connection;
   do {
     if (repoRemoteUri === ADD_CONN_CHOICE) {
       existingConns.push(
@@ -134,7 +134,7 @@ export async function linkGitHubRepository(
 }
 
 /**
- * Creates a new GCB GitHub connection and ensures that it is full configured on the GitHub
+ * Creates a new GCB GitHub connection resource and ensures that it is fully configured on the GitHub
  * side (ie associated with an account/org and some subset of repos within that scope).
  * Copies over Oauth creds from the sentinel Oauth connection to save the user from having to
  * reauthenticate with GitHub.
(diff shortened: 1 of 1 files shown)


src/init/features/apphosting/repo.ts
```diff
@@ -87,84 +91,111 @@ export async function linkGitHubRepository(
...
-    let refreshedConn = conn;
-    while (refreshedConn.installationState.stage !== "COMPLETE") {
-      refreshedConn = await promptAppInstall(conn);
-    }
-    existingConns.push(refreshedConn);
+
+  if (existingConns.length === 0) {
+    await ensureSecretManagerAdminGrant(projectId);
+    existingConns.push(
+      await ensureFullyInstalledConnection(projectId, location, generateConnectionId(), oauthConn),
```
mathu97, line 101: NIT: the naming here doesn't make me think that a new connection would be created.

src/init/features/apphosting/repo.ts
```diff
@@ -87,84 +91,111 @@ export async function linkGitHubRepository(
...
+    );
   }
 
-  let { remoteUri, connection } = await promptRepositoryUri(projectId, existingConns);
-  while (CONFIGURE_REMOTE_URI_CHOICES.includes(remoteUri)) {
-    if (remoteUri === ADD_REPO_CHOICE) {
```
mathu97, line 115: why remove this? I like that it makes the UX simpler but would this not mean that we would be creating a new Connection whenever the user wants to make a change to the repos available to the Cloud Build Github app?

Though i'm not too sure about the consequences of having many connections ... are there any impact on billing?

mathu97 (commented):
LGTM as long as this was tested manually. There doesn't seem to be any logic changes - mostly refactoring.

src/init/features/apphosting/repo.ts
tonyjhuang: Updated to `createFullyInstalledConnection()`

src/init/features/apphosting/repo.ts
tonyjhuang: GCB `connection` resources are free so it doesn't hurt to create multiple. The idea is that once we have a Firebase GitHub app, we can dedupe `connections` on behalf of the user in our redirect handler.

mathu97 (approved):

PR #6887, opened 2024-03-18
Add Developer Connect API client

mathu97:
### Description

Similar to the Cloud Build Client, this is a helper that will make working with the Developer Connect API easier.

### DIFF
--- a/src/gcp/devConnect.ts
+++ b/src/gcp/devConnect.ts
@@ -1,0 +1,244 @@
+import { Client } from "../apiv2";
+import { developerConnectOrigin, developerConnectP4SAOrigin } from "../api";
+
+const PAGE_SIZE_MAX = 100;
+
+export const client = new Client({
+  urlPrefix: developerConnectOrigin,
+  auth: true,
+  apiVersion: "v1",
+});
+
+export interface OperationMetadata {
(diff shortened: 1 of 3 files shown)


src/gcp/devConnect.ts
```diff
@@ -0,0 +1,244 @@
...
+ */
+export async function listConnections(projectId: string, location: string): Promise<Connection[]> {
+  const conns: Connection[] = [];
+  const getNextPage = async (pageToken = ""): Promise<void> => {
+    const res = await client.get<{
+      connections: Connection[];
+      nextPageToken?: string;
+    }>(`/projects/${projectId}/locations/${location}/connections`, {
+      queryParams: {
+        pageSize: PAGE_SIZE_MAX,
```
tonyjhuang, line 166: this seems at odds with `fetchLinkableGitRespositories()` below, is PAGE_SIZE_MAX intended to be the max value or the default value?

tonyjhuang (approved):

src/gcp/devConnect.ts
tonyjhuang: Maybe we should have a listConnections vs listAllConnections?

PR #6889, opened 2024-03-19
DeveloperConnect: make `fetchLinkableGitRepositories` list all linkable git repositories

mathu97:
### Description

Instead of fetching a certain page of linkable git repos, `fetchLinkableGitRepositories` has been updated to fetch all linkable git repositories, iterating over all pages. This will behave similar to `listConnections` (renamed to `listAllConnections`)

Follow up from: https://github.com/firebase/firebase-tools/pull/6887/files/d8841df93bac3c0dbcb8773783db627015842094#r1530608858

### DIFF
--- a/src/gcp/devConnect.ts
+++ b/src/gcp/devConnect.ts
@@ -1,7 +1,7 @@
 import { Client } from "../apiv2";
 import { developerConnectOrigin, developerConnectP4SAOrigin } from "../api";
 
-const PAGE_SIZE_MAX = 100;
+const PAGE_SIZE_MAX = 1000;
 
 export const client = new Client({
   urlPrefix: developerConnectOrigin,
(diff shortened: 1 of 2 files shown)


src/gcp/devConnect.ts
```diff
@@ -181,22 +184,33 @@ export async function listConnections(projectId: string, location: string): Prom
...
-      pageToken,
-    },
-  });
+  const repos: LinkableGitRepository[] = [];
 
-  return res.body;
+  const getNextPage = async (pageToken = ""): Promise<void> => {
+    const res = await client.get<LinkableGitRepositories>(name, {
+      queryParams: {
+        PAGE_SIZE_MAX,
```
tonyjhuang, line 198: ooc why default to 100 here instead of 1000 if our goal is to list all results?

mathu97: no specific reason. But 1000 makes sense for both I think. Less calls to the APIs. Even if 1000 is more than the supported max for page_size, the API should just return the max page_size it can from what I understand.

tonyjhuang (approved):

PR #6899, opened 2024-03-21
Add support for grantaccess

abhis3:
Add support for apphosting:secrets:grantaccess

### DIFF
--- a/src/commands/apphosting-secrets-grantaccess.ts
+++ b/src/commands/apphosting-secrets-grantaccess.ts
@@ -6,7 +6,7 @@ import { requireAuth } from "../requireAuth";
 import * as secrets from "../functions/secrets";
 import { requirePermissions } from "../requirePermissions";
 import * as apphosting from "../gcp/apphosting";
-import { grantSecretAccess } from "../init/features/apphosting";
+import { grantSecretAccess } from "../init/features/apphosting/secrets";
 
 export const command = new Command("apphosting:secrets:grantaccess <secretName>")
   .description("grant service accounts permissions to the provided secret")
--- a/src/commands/apphosting-secrets-grantaccess.ts
+++ b/src/commands/apphosting-secrets-grantaccess.ts
@@ -5,11 +5,8 @@ import { FirebaseError } from "../error";
 import { requireAuth } from "../requireAuth";
 import * as secrets from "../functions/secrets";
 import { requirePermissions } from "../requirePermissions";
-import { logger } from "../logger";
 import * as apphosting from "../gcp/apphosting";
-import * as secretManager from "../gcp/secretManager";
-import { setIamPolicy } from "../gcp/secretManager";
-import * as iam from "../gcp/iam";
+import { grantSecretAccess } from "../init/features/apphosting";
 
 export const command = new Command("apphosting:secrets:grantaccess <secretName>")
   .description("grant service accounts permissions to the provided secret")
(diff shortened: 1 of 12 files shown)


src/commands/apphosting-secrets-grantaccess.ts
```diff
@@ -0,0 +1,113 @@
...
+    "secretmanager.secrets.update",
+    "secretmanager.versions.add",
+    "secretmanager.secrets.setIamPolicy",
+  ])
+  .action(async (secretName: string, options: Options) => {
+    const projectId = needProjectId(options);
+    const projectNumber = await needProjectNumber(options);
+
+    if (!options.location) {
+      logger.error(
```
tonyjhuang, line 33: whats the difference between logger.error() and throwing a FirebaseError?

src/commands/apphosting-secrets-grantaccess.ts
```diff
@@ -0,0 +1,113 @@
...
+        role: "roles/secretmanager.viewer",
+        members: [`serviceAccount:${serviceAccounts.buildServiceAccount}`],
+      },
+    ];
+
+    try {
+      await setIamPolicy(secret, bindings);
+    } catch (err: any) {
+      throw new FirebaseError(
+        `Failed to set IAM bindings ${bindings.toString()} on secret: ${secret.name}. Ensure you have the permissions to do so and try again.`,
```
tonyjhuang, line 86: "Ensure you have the permissions to do so and try again."

Don't we ensure the user has permissions on L21?

src/commands/apphosting-secrets-grantaccess.ts
```diff
@@ -0,0 +1,113 @@
...
+        { original: err },
+      );
+    }
+
+    logger.info(`Successfully set IAM bindings ${bindings.toString()} on secret: ${secret.name}.`);
```
tonyjhuang, line 91: https://github.com/firebase/firebase-tools/blob/master/src/init/features/apphosting/index.ts#L118

tonyjhuang (commented):
tests?

src/commands/apphosting-secrets-grantaccess.ts
abhis3: Changed to `logSuccess`

src/commands/apphosting-secrets-grantaccess.ts
abhis3: Looks like logger.error() is just more poorly formatted for our use case, changed to FirebaseError

src/commands/apphosting-secrets-grantaccess.ts
abhis3: true, but keeping that line now that it will be enforced by the calling function, not within the CLI function itself

src/init/features/apphosting/index.ts
```diff
@@ -161,6 +162,83 @@ function defaultComputeServiceAccountEmail(projectId: string): string {
...
+      ],
+    },
+    {
+      role: "roles/secretmanager.viewer",
+      members: [`serviceAccount:${serviceAccounts.buildServiceAccount}`],
+    },
+  ];
+
+  try {
+    await secretManager.setIamPolicy(secret, bindings);
```
tonyjhuang, line 231: have you tested this out in prod? I'm pretty sure you want to merge the new bindings into the existing binding before setting. Otherwise you overwrite existing grants

tonyjhuang (changes requested):

src/init/features/apphosting/index.ts
```diff
@@ -161,6 +162,83 @@ function defaultComputeServiceAccountEmail(projectId: string): string {
   return `${DEFAULT_COMPUTE_SERVICE_ACCOUNT_NAME}@${projectId}.iam.gserviceaccount.com`;
 }
 
+function defaultCloudBuildServiceAccount(projectNumber: string): string {
```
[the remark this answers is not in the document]
inlined: Can also factor out the implementation from src/deploy/functions/checkIam.ts

src/init/features/apphosting/index.ts
inlined: Can confirm. Consider adding helper methods like those used in src/gcp/cloudfunctions.ts (setInvokerCreate and setInvokerUpdate, where the former blindly writes and the latter merges)

src/init/features/apphosting/index.ts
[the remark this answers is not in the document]
abhis3: Looks like this isn't always the case https://github.com/firebase/firebase-tools/blob/master/src/init/features/apphosting/index.ts#L160-L162

I'd prefer to scope it to the secrets.ts file in this case

src/init/features/apphosting/index.ts
abhis3: Yeah good call, I tested it but it was with some inherited permissions that couldn't be overridden. Fixed

src/init/features/apphosting/index.ts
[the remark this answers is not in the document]
tonyjhuang: The service account you're referring to is specific to App Hosting (see go/firestack-compute-runner) so it's reasonable to me that it's in the apphosting index file.

In these service accounts' case, you're referring to Cloud Build's and Compute Engine's service accounts so it's appropriate to move them to those service's files.

To echo what @inlined mentioned, we'll eventually want to share a decent chunk of logic with the functions secrets flow so making sure these live in common files will make our lives a tiny bit easier in the near future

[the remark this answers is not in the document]
abhis3: I see thanks for the context, changed and refactored the existing references

src/init/features/apphosting/secrets.ts
```diff
@@ -0,0 +1,98 @@
...
+    },
+  ];
+
+  try {
+    const existingBindings = (await secretManager.getIamPolicy(secret)).bindings;
+    const updatedBindings = updateBindings(existingBindings, newBindings);
+    await secretManager.setIamPolicy(secret, updatedBindings);
+  } catch (err: any) {
+    throw new FirebaseError(
+      `Failed to set IAM bindings ${JSON.stringify(newBindings)} on secret: ${secret.name}. Ensure you have the permissions to do so and try again.`,
```
tonyjhuang, line 70: small nit: this is going to print a confusing error message if `secretManager.getIamPolicy` fails on L65. Would recommend pulling that out to its own try catch

src/init/features/apphosting/secrets.ts
```diff
@@ -0,0 +1,98 @@
...
+      { original: err },
+    );
+  }
+
+  logSuccess(`Successfully set IAM bindings on secret ${secret.name}.\n`);
+}
+
+function updateBindings(existing: iam.Binding[], newBindings: iam.Binding[]): iam.Binding[] {
```
tonyjhuang, line 78: Is there a common IAM file that we can move this to? This seems generally useful outside of the context of apphosting secrets

tonyjhuang (changes requested):
Thanks for the changes! Just a few small nits :)

src/init/features/apphosting/secrets.ts
abhis3: Done!

src/init/features/apphosting/secrets.ts
abhis3: Yeah that's fair I moved it to src/gcp/iam, I was hesitant but there are some helper methods in there so I think it would be beneficial

src/gcp/iam.ts
```diff
@@ -222,3 +219,33 @@ export async function testIamPermissions(
...
+export function generateUpdatedIamBindings(
+  existingBindings: Binding[],
+  newBindings: Binding[],
+): Binding[] {
+  const updatedBindings: Binding[] = existingBindings ? [...existingBindings] : [];
+
+  for (const newBinding of newBindings) {
+    const existingRoleIndex = updatedBindings.findIndex((b) => b.role === newBinding.role);
+
+    if (existingRoleIndex >= 0) {
```
tonyjhuang, line 237: Can you double check this is necessary? I _think_ IAM dedupes/coalesces roles but it's been a while since I've played with the API.

Also, tests? :)

abhis3: Ahh you're right this is completely unnecessary, let's let the API handle it. Removed the method and the secrets.spec.ts has a test to ensure this

tonyjhuang (approved):
Ty for the changes 🔥🔥

PR #6900, opened 2024-03-21
Vector config support

NickChittle:
### Description

Add support for the vectorConfig in Firestore Indexes, this is similar to the existing 'order' and 'arrayConfig' options that currently exist.

### Scenarios Tested

Tested vector configs with and without other existing index field types.

### Sample Commands

`firebase deploy --only firestore`

### DIFF
--- a/src/firestore/api-sort.ts
+++ b/src/firestore/api-sort.ts
@@ -180,6 +180,7 @@ export function compareFieldOverride(a: Spec.FieldOverride, b: Spec.FieldOverrid
  *   1) Field path.
  *   2) Sort order (if it exists).
  *   3) Array config (if it exists).
+ *   4) Vector config (if it exists).
  */
 function compareIndexField(a: API.IndexField, b: API.IndexField): number {
   if (a.fieldPath !== b.fieldPath) {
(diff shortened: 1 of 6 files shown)


src/firestore/api-sort.ts
```diff
@@ -225,6 +230,19 @@ function compareArrayConfig(a?: API.ArrayConfig, b?: API.ArrayConfig): number {
   return ARRAY_CONFIG_SEQUENCE.indexOf(a) - ARRAY_CONFIG_SEQUENCE.indexOf(b);
 }
 
+function compareVectorConfig(a?: API.VectorConfig, b?: API.VectorConfig): number {
+  if (a === undefined) {
```
rwhogg, line 234: I'd prefer using `!a` rather than `a === undefined` just in case an errant `null` gets in here.

src/firestore/api-sort.ts
```diff
@@ -225,6 +230,19 @@ function compareArrayConfig(a?: API.ArrayConfig, b?: API.ArrayConfig): number {
...
+    if (b === undefined) {
+      return 0;
+    } else {
+      return 1;
+    }
+  } else if (b === undefined) {
+    return -1;
```
rwhogg, line 241: Is this ordering deliberate? If `a` has a higher `dimension` than `b`, then this function will have a positive return value. But if `b` is `undefined` and `a` is not (and thus presumably has non-zero degree), this function will have a negative return value.

rwhogg (approved):

joehan (approved):

PR #6904, opened 2024-03-22
Fix "could not assert Secret Manager permissions" Cloud Build error

mathu97:
Users are often running into this `could not assert Secret Manager permissions` error when onboarding to apphosting.

This occurs when the user has no Cloud Build repo connections (including the sentinel Oauth connection) and the secret manager admin role has not been granted to the Cloud Build P4SA. So when the user goes to create an Oauth connection first but Cloud Build has no permissions to save the oath token in secret manager, it throws the above error. 

To fix this we just need to ensure that the correct secret manager privileges are set to the Cloud Build P4SA before even creating the initial sentinel Oauth connection.

### DIFF
--- a/src/init/features/apphosting/repo.ts
+++ b/src/init/features/apphosting/repo.ts
@@ -92,7 +92,6 @@ export async function linkGitHubRepository(
 ): Promise<gcb.Repository> {
   utils.logBullet(clc.bold(`${clc.yellow("===")} Set up a GitHub connection`));
   // Fetch the sentinel Oauth connection first which is needed to create further GitHub connections.
-  await ensureSecretManagerAdminGrant(projectId);
   const oauthConn = await getOrCreateOauthConnection(projectId, location);
   const existingConns = await listAppHostingConnections(projectId);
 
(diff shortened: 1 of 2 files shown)


src/init/features/apphosting/repo.ts
```diff
@@ -92,11 +92,11 @@ export async function linkGitHubRepository(
 ): Promise<gcb.Repository> {
   utils.logBullet(clc.bold(`${clc.yellow("===")} Set up a GitHub connection`));
   // Fetch the sentinel Oauth connection first which is needed to create further GitHub connections.
+  await ensureSecretManagerAdminGrant(projectId);
```
tonyjhuang, line 95: This change requires all callers of this command to have access to fetch the IAM policy (to ensure that the CB service account has SecretManager permissions), which I don't believe is intentional.

We probably want to refactor getOrCreateOauthConnection to ensure the grant prior to creating the oauth connection if it doesnt exist

mathu97: ah good point, ya we only want to check for this when users go to create an oauth connection.

tonyjhuang (approved):
Can we add a test?

src/init/features/apphosting/repo.ts
```diff
@@ -174,7 +173,20 @@ async function getOrCreateOauthConnection(
   projectId: string,
   location: string,
 ): Promise<gcb.Connection> {
-  let conn = await getOrCreateConnection(projectId, location, APPHOSTING_OAUTH_CONN_NAME);
```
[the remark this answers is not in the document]
mathu97: ya we can but I do like that it helps with readability.

PR #6915, opened 2024-03-26
Onboard users to App Hosting using Developer Connect

mathu97:
### Description

Enables app hosting onboarding using developer connect. This path is gated by a flag `--with-dev-connect` or `-w`. The changes in this PR should not affect or change the current flow of onboarding using Cloud Build repositories.

### Sample Commands

FIREBASE_DEVELOPERCONNECT_P4SA_URL=<p4sa_url> CLOUD_SECRET_MANAGER_URL=<secret_manager_url> FIREBASE_DEVELOPERCONNECT_URL=<dev_connect_url> firebase apphosting:backends:create --project=<firebase_project> --with-dev-connect

**Note**: This flow is currently only being tested with Developer Connect autopush. Staging and Prod envs are not ready yet.


src/commands/apphosting-backends-create.ts
```diff
@@ -13,11 +13,27 @@ export const command = new Command("apphosting:backends:create")
     "specify the service account used to run the server",
     "",
   )
+  .option(
```
mathu97, line 16: add test

src/commands/apphosting-connections-list.ts
```diff
@@ -0,0 +1,15 @@
+import { Command } from "../command";
```
mathu97, line 1: move to a separate PR (this is only helpful for testing)

src/commands/apphosting-connections-reset.ts
```diff
@@ -0,0 +1,15 @@
+import { Command } from "../command";
```
mathu97, line 1: move to a separate PR (this is only helpful for testing)

src/commands/index.ts
```diff
@@ -165,6 +165,9 @@ export function load(client: any): any {
     client.apphosting.backends = {};
     client.apphosting.backends.list = loadCommand("apphosting-backends-list");
     client.apphosting.backends.create = loadCommand("apphosting-backends-create");
+    client.apphosting.connections = {};
```
mathu97, line 168: move to a separate PR (this is only helpful for testing)

src/init/features/apphosting/index.ts
```diff
@@ -36,6 +39,105 @@ const apphostingPollerOptions: Omit<poller.OperationPollerOptions, "operationRes
...
+
+  await Promise.all([
+    ensure(projectId, developerConnectOrigin(), "apphosting", true),
+    ensure(projectId, cloudbuildOrigin(), "apphosting", true),
+    ensure(projectId, secretManagerOrigin(), "apphosting", true),
+    ensure(projectId, cloudRunApiOrigin(), "apphosting", true),
+    ensure(projectId, artifactRegistryDomain(), "apphosting", true),
+  ]);
+
+  logBullet("All required APIs are enabled");
```
mathu97, line 60: remove

src/gcp/devConnect.ts
```diff
@@ -147,6 +170,7 @@ export async function getConnection(
   location: string,
   connectionId: string,
 ): Promise<Connection> {
+  location = LOCATION_OVERRIDE;
```
mathu97, line 173:
```suggestion
  location = LOCATION_OVERRIDE ?? location
```
Move LOCATION_OVERRIDE to env variable in `api.ts`. The `LOCATION_OVERRIDE` can be removed later, but it is  helpful for debugging dev connect in different environments.

src/init/features/apphosting/githubConnections.ts
```diff
@@ -0,0 +1,431 @@
...
+  }
+  const { projectId, location, id } = match.groups as unknown as ConnectionNameParts;
+  return {
+    projectId,
+    location,
+    id,
+  };
+}
+
+export function isApphostingConnection(name: string): boolean {
```
mathu97, line 50: move to separate PR

src/init/features/apphosting/index.ts
```diff
@@ -351,3 +453,39 @@ export async function orchestrateRollout(
   }
   return { rollout, build };
 }
+
+/**
+ * lists dev connect connections
+ */
+export async function listDeveloperConnectAppHostingConnections(
```
mathu97, line 460: move to separate Pr

src/init/features/apphosting/index.ts
```diff
@@ -351,3 +453,39 @@ export async function orchestrateRollout(
...
+  for (let i = 0; i < appHostingConnections.length; i++) {
+    const connection = appHostingConnections[i];
+    logBullet(connection.name);
+  }
+}
+
+/**
+ * deletes all dev connect connections
+ */
+export async function resetDeveloperConnectAppHostingConnections(
```
mathu97, line 477: move to separate PR

src/init/features/apphosting/index.ts
```diff
@@ -36,6 +39,105 @@ const apphostingPollerOptions: Omit<poller.OperationPollerOptions, "operationRes
   masterTimeout: 25 * 60 * 1_000,
   maxBackoff: 10_000,
 };
+/**
+ * Set up a new App Hosting backend.
+ */
+export async function doSetupWithDeveloperConnect(
```
mathu97, line 45: might be better to use the original `doSetup` function with a flag to use dev connect instead of cloud build

src/init/features/apphosting/githubConnections.ts
```diff
@@ -0,0 +1,431 @@
...
+        `\t  --role="roles/secretmanager.admin\n`,
+    );
+    throw new FirebaseError("Insufficient IAM permissions to create a new connection to GitHub");
+  }
+
+  try {
+    await rm.addServiceAccountToRoles(projectId, dcsaEmail, ["roles/secretmanager.admin"], true);
+  } catch (e: any) {
+    // if the dev connect P4SA doesn't exist in the project, generate one
+    if (e?.code === 400) {
```
mathu97, line 300: Add test coverage for this

tonyjhuang (changes requested):

src/gcp/devConnect.ts
```diff
@@ -3,7 +3,14 @@ import { developerConnectOrigin, developerConnectP4SAOrigin } from "../api";
 import { generateServiceIdentity } from "./serviceusage";
 
 const PAGE_SIZE_MAX = 1000;
-const LOCATION_OVERRIDE = "us-central1"; // TODO(mathusan): for debugging only
+
+/**
+ * LOCATION_OVERRIDE is used to test Developer Connect in different environments.
+ *
+ * Autopush -> 'us-central1'
+ * Staging -> 'us-west1'
```
tonyjhuang, line 11: I wonder if we should include this information, I don't see any other references to autopush/staging envs in this repo

tonyjhuang: Also are we planning on removing this after we launch? How is it different from calling `backends:create --location`?

tonyjhuang (changes requested):
Almost there :)

src/init/features/apphosting/githubConnections.ts
```diff
@@ -0,0 +1,437 @@
...
+      projectId,
+      dcsaEmail,
+      ["roles/secretmanager.admin"],
+      /* skipAccountLookup= */ true,
+    );
+  } catch (e: any) {
+    // if the dev connect P4SA doesn't exist in the project, generate one
+    if (e?.code === 400 || e?.status === 400) {
+      await devConnect.generateP4SA(projectNumber);
+      await rm.addServiceAccountToRoles(
```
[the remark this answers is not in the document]
mathu97: > You're right in that calling GenerateServiceIdentity multiple times after the P4SA already exists will always return a done operation (eg from your testing), but this may not be the case for a new onboarding project that doesn't already have a p4sa.

No I meant on the first call. I've tested it a couple times, creating a new project from scratch. On first call to GenerateServiceIdentity we get an operation with the name `finished.DONE_OPERATION ` with `done = true`. And polling with that operation name results in a 400 error. I can add the polling logic and skip polling if the returned operation is already done.

src/gcp/devConnect.ts
mathu97: > I wonder if we should include this information, I don't see any other references to autopush/staging envs in this repo
ya I think it might cause some confusion. I'll remove. 

> Also are we planning on removing this after we launch? How is it different from calling backends:create --location?

Yes this will be removed as soon as developer connect production is good to go. The reason for this is because we can only create backends in `us-central1` at the moment, and if we were to test in dev connect in staging we need to use `us-west1` because dev connect staging is only spun up in that location.

tonyjhuang (approved):

Issue #6896, opened 2024-03-20
SignatureDoesNotMatch error when using GOOGLE_CLOUD_QUOTA_PROJECT environment variable


jdcoldsmith:
### Environment info

**firebase-tools:** 13.5.2

**Platform:** Ubuntu

### Issue
Hello! I have recently run into an issue using the `GOOGLE_CLOUD_QUOTA_PROJECT` environment variable that was recently respected in PR #6836. The way our application works is that we use one master service account that has permissions to each of the projects we would like to deploy functions to. For better quota management I started using the `GOOGLE_CLOUD_QUOTA_PROJECT` environment variable to prevent all of our usage lumping together into the project where our master service account was created. Upon using this variable the Firebase CLI asked me to give the "roles/serviceusage.serviceUsageConsumer" permission which I added to the service account, but now whenever I deploy functions I get a `SignatureDoesNotMatch` error claiming that there is something wrong with my "secret key or signing method". You can see the full error below. I have scoured the internet looking for anyone running into something similar to no avail so I thought I would ask you smart people for some help.😊 Thank you for your time!

![image](https://github.com/firebase/firebase-tools/assets/65235172/49f80168-1b4a-439c-86bd-e8535337c528)

aalej:
Hey @jdcoldsmith, thanks for reaching out and sharing your observations. Sorry to hear you encountered this issue. I’m able to reproduce this on my end, let me raise this to our engineering team so they can take a look.

jdcoldsmith:
@aalej Thank you for checking this out in such a timely manner!😊

inlined:
I'll look into this. It may require escalating to a few teams within Google, so please bear with me if that takes a while. It looks like the source upload to Cloud Functions is failing, but that upload is going to a project that is not under your control (so that you don't have to pay for it). This technical detail may mean that we cannot respect the env var for function source uploads.

inlined:
We believe that GCS signed URLs may not be compatible with this feature, so yesterday's release of the CLI omits the header on that specific API call. This shouldn't cause you to run into any particular quota issues because the upload was never quota'd against your project as far as I'm aware.

jdcoldsmith:
Thank you very much for you help!

PR #6917, opened 2024-03-26
Ignore quota project in GCF source uploads

inlined:
Best fix for #6896 that we can do for now. I strongly suspect that GCS just can't handle signed URL uploads with quota projects.

### DIFF
--- a/src/deploy/functions/deploy.ts
+++ b/src/deploy/functions/deploy.ts
@@ -31,12 +31,18 @@ async function uploadSourceV1(
     stream: fs.createReadStream(source.functionsSourceV1!),
   };
   if (process.env.GOOGLE_CLOUD_QUOTA_PROJECT) {
-    logLabeledWarning("functions", "GOOGLE_CLOUD_QUTOA_PROJECT is not usable when uploading source for Cloud Functions.");
+    logLabeledWarning(
+      "functions",
+      "GOOGLE_CLOUD_QUTOA_PROJECT is not usable when uploading source for Cloud Functions.",
+    );
   }
-  await gcs.upload(uploadOpts, uploadUrl, {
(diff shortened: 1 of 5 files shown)


joehan (approved):

src/deploy/functions/deploy.ts
```diff
@@ -51,7 +62,13 @@ async function uploadSourceV2(
     file: source.functionsSourceV2!,
     stream: fs.createReadStream(source.functionsSourceV2!),
   };
-  await gcs.upload(uploadOpts, res.uploadUrl);
+  if (process.env.GOOGLE_CLOUD_QUOTA_PROJECT) {
+    logLabeledWarning(
+      "functions",
+      "GOOGLE_CLOUD_QUTOA_PROJECT is not usable when uploading source for Cloud Functions.",
```
jdcoldsmith, line 68: Typo here. Should be "GOOGLE_CLOUD_QUOTA_PROJECT"

src/deploy/functions/deploy.ts
```diff
@@ -30,9 +30,20 @@ async function uploadSourceV1(
     file: source.functionsSourceV1!,
     stream: fs.createReadStream(source.functionsSourceV1!),
   };
-  await gcs.upload(uploadOpts, uploadUrl, {
-    "x-goog-content-length-range": "0,104857600",
-  });
+  if (process.env.GOOGLE_CLOUD_QUOTA_PROJECT) {
+    logLabeledWarning(
+      "functions",
+      "GOOGLE_CLOUD_QUTOA_PROJECT is not usable when uploading source for Cloud Functions.",
```
jdcoldsmith, line 36: Typo here. Should be "GOOGLE_CLOUD_QUOTA_PROJECT"

PR #6928, opened 2024-03-27
Remove reference to now cleaned up experiment

inlined:
Saw the experiment in a log while reproing a bug and realized that the experiment is long committed.


joehan (approved):
Format it, but lgtm!

PR #6935, opened 2024-03-29
Support pagination in listBackends

inlined:
Also start reporting unreachable regions as we should add a disclaimer for those.


abhis3 (approved):
thanks for the change!

PR #6939, opened 2024-04-01
Chose service accounts dialog

inlined:
Part 1 of 4 for secrets:set

### DIFF
--- a/src/apphosting/secrets/dialogs.ts
+++ b/src/apphosting/secrets/dialogs.ts
@@ -7,116 +7,142 @@ import * as iam from "../../gcp/iam";
 import * as utils from "../../utils";
 import { logger } from "../../logger";
 
-interface Metadata { location: string, id: string, accounts: string[] };
+interface Metadata {
+  location: string;
+  id: string;
+  accounts: string[];
+}
 
-export function serviceAccountsForBackend(projectNumber: string, backend: apphosting.Backend): string[] {
-    if (backend.serviceAccount) {
-        return [backend.serviceAccount];
-    }
-    return [iam.getDefaultCloudBuildServiceAgent(projectNumber), iam.getDefaultComputeEngineServiceAgent(projectNumber)];
+export function serviceAccountsForBackend(
+  projectNumber: string,
+  backend: apphosting.Backend,
+): string[] {
+  if (backend.serviceAccount) {
+    return [backend.serviceAccount];
+  }
+  return [
+    iam.getDefaultCloudBuildServiceAgent(projectNumber),
+    iam.getDefaultComputeEngineServiceAgent(projectNumber),
+  ];
 }
 
 export function toMetadata(projectNumber: string, backends: apphosting.Backend[]): Metadata[] {
-    const metadata: Metadata[] = [];
-    for (const backend of backends) {
-        const [,,,location,,id] = backend.name.split("/");
-        metadata.push({ location, id, accounts:  serviceAccountsForBackend(projectNumber, backend) });
+  const metadata: Metadata[] = [];
+  for (const backend of backends) {
+    const [, , , location, , id] = backend.name.split("/");
+    metadata.push({ location, id, accounts: serviceAccountsForBackend(projectNumber, backend) });
+  }
+  return metadata.sort((left, right) => {
+    const cmplocation = left.location.localeCompare(right.location);
+    if (cmplocation) {
+      return cmplocation;
     }
-    return metadata.sort((left, right) => {
-        const cmplocation = left.location.localeCompare(right.location);
-        if (cmplocation) {
-            return cmplocation;
-        }
-        return left.id.localeCompare(right.id);
-    });
+    return left.id.localeCompare(right.id);
+  });
 }
 
 export function tableForBackends(metadata: Metadata[]): [headers: string[], rows: string[][]] {
-    const headers = ["location", "backend", metadata.findIndex(val => val.accounts.length > 1) === -1 ? "service account" : "service accounts"];
-    const rows = metadata.map(m => [m.location, m.id, m.accounts.join(", ")]);
-    return [headers, rows];
+  const headers = [
+    "location",
+    "backend",
+    metadata.findIndex((val) => val.accounts.length > 1) === -1
+      ? "service account"
+      : "service accounts",
+  ];
+  const rows = metadata.map((m) => [m.location, m.id, m.accounts.join(", ")]);
+  return [headers, rows];
 }
 
-export const WARN_NO_BACKENDS = 
-    "To use this secret, your backend's service account must have secret accessor permission. " +
-    "It does not look like you have a backend yet. After creating a backend, grant access with " +
-    clc.bold("firebase apphosting:secrets:grantAccess");
+export const WARN_NO_BACKENDS =
+  "To use this secret, your backend's service account must have secret accessor permission. " +
+  "It does not look like you have a backend yet. After creating a backend, grant access with " +
+  clc.bold("firebase apphosting:secrets:grantAccess");
 
 export const GRANT_ACCESS_IN_FUTURE = `To grant access in the future, run ${clc.bold("firebase apphosting:secrets:grantaccess")}`;
 
-export async function selectBackendServiceAccounts(projectNumber: string, projectId: string, options: any): Promise<string[]> {
-    const listBackends = await apphosting.listBackends(projectId, "-")
+export async function selectBackendServiceAccounts(
+  projectNumber: string,
+  projectId: string,
+  options: any,
+): Promise<string[]> {
+  const listBackends = await apphosting.listBackends(projectId, "-");
 
-    if (listBackends.unreachable.length) {
-        utils.logLabeledWarning(
-            "apphosting",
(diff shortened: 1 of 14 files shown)


src/apphosting/secrets/dialogs.ts
```diff
@@ -0,0 +1,148 @@
...
+  return [
+    iam.getDefaultCloudBuildServiceAgent(projectNumber),
+    iam.getDefaultComputeEngineServiceAgent(projectNumber),
+  ];
+}
+
+export function toMetadata(projectNumber: string, backends: apphosting.Backend[]): Metadata[] {
+  const metadata: Metadata[] = [];
+  for (const backend of backends) {
+    const [, , , location, , id] = backend.name.split("/");
```
tonyjhuang, line 32: Let's include a line comment with an example of a raw backend.name value

src/apphosting/secrets/dialogs.ts
```diff
@@ -0,0 +1,148 @@
...
+    metadata.push({ location, id, accounts: serviceAccountsForBackend(projectNumber, backend) });
+  }
+  return metadata.sort((left, right) => {
+    const cmplocation = left.location.localeCompare(right.location);
```
tonyjhuang, line 36: If we're sorting, consider adding a second sort column for name (will help with users with multi-regional backends)

src/apphosting/secrets/dialogs.ts
```diff
@@ -0,0 +1,148 @@
...
+    utils.logLabeledWarning("apphosting", WARN_NO_BACKENDS);
+    return [];
+  }
+
+  if (listBackends.backends.length === 1) {
+    const grant = await prompt.confirm({
+      nonInteractive: options.nonInteractive,
+      default: true,
+      message:
+        "To use this secret, your backend's service account must have secret accessor permission. Would you like to grant it now?",
```
tonyjhuang, line 88: Do we want to tell the user the name of the backend?

src/apphosting/secrets/dialogs.ts
```diff
@@ -0,0 +1,148 @@
...
+    if (grant) {
+      return serviceAccountsForBackend(projectNumber, listBackends.backends[0]);
+    }
+    utils.logLabeledBullet("apphosting", GRANT_ACCESS_IN_FUTURE);
+    return [];
+  }
+
+  const metadata: Metadata[] = toMetadata(projectNumber, listBackends.backends);
+
+  // Use JSON.stringify because deep copmarison is annoying in JS. Because the order of the service account list should be deterinistic,
```
tonyjhuang, line 99: typo in comparison

src/apphosting/secrets/dialogs.ts
```diff
@@ -0,0 +1,148 @@
...
+  // this shouldn't need a sort command.
+  const test = JSON.stringify(metadata[0].accounts);
+  const allSharedAccounts = metadata.every((val) => JSON.stringify(val.accounts) === test);
```
tonyjhuang, line 102: Alternatively, you can do:

new Set(metadata.map(val => JSON.stringify(val.accounts)).size === 1

src/apphosting/secrets/dialogs.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+    utils.logLabeledBullet("apphosting", GRANT_ACCESS_IN_FUTURE);
+    return [];
+  }
+
+  const metadata: Metadata[] = toMetadata(projectNumber, listBackends.backends);
+
+  // Use JSON.stringify because deep copmarison is annoying in JS. Because the order of the service account list should be deterinistic,
+  // this shouldn't need a sort command.
+  const test = JSON.stringify(metadata[0].accounts);
```
tonyjhuang, line 101: firstBackendServiceAccounts?

src/apphosting/secrets/dialogs.ts
```diff
@@ -0,0 +1,148 @@
...
+  const allSharedAccounts = metadata.every((val) => JSON.stringify(val.accounts) === test);
+  if (allSharedAccounts) {
+    const grant = await prompt.confirm({
+      nonInteractive: options.nonInteractive,
+      default: true,
+      message:
+        "To use this secret, your backend's service account must have secret accessor permission. All of your backends use " +
```
tonyjhuang, line 108: Is the intention of this method to be shared between `set` and `grantAccess`? If so, this messaging feels overly verbose for the `grantAccess` usecase

tonyjhuang (changes requested):
Thanks for the PR! Left some comments

src/apphosting/secrets/dialogs.ts
```diff
@@ -0,0 +1,148 @@
+import * as clc from "colorette";
+const Table = require("cli-table");
+
+import * as apphosting from "../../gcp/apphosting";
+import * as prompt from "../../prompt";
+import * as iam from "../../gcp/iam";
+import * as utils from "../../utils";
+import { logger } from "../../logger";
+
+interface Metadata {
```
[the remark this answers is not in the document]
inlined: Done

src/apphosting/secrets/dialogs.ts
```diff
@@ -0,0 +1,148 @@
...
+  location: string;
+  id: string;
+  accounts: string[];
+}
+
+export function serviceAccountsForBackend(
```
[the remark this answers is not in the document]
inlined: Done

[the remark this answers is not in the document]
inlined: That function is not forwards compatible with explicit SAs and I'm not sure how to use the map of SAs when there will soon only be one.

My intention was to:
1. Finish the set command reasonably independently
2. Migrate all the existing source to appropraite locations
3. Reconcile the various commands

I still need to check with Abhi why the viewer permission is needed for Build. That may change things in the future.

src/apphosting/secrets/dialogs.ts
inlined: Done

src/apphosting/secrets/dialogs.ts
inlined: That's what the next condition does (`return left.id.localeCompare(right.id)`)

src/apphosting/secrets/dialogs.ts
inlined: If they only have one backend that seems gratuitous, no?

src/apphosting/secrets/dialogs.ts
inlined: Done

src/apphosting/secrets/dialogs.ts
inlined: I'm not sure I agree, but this also sounds like a discussion we can have once we're ready to reevaluate `secrets:grantaccess` and whether it should support the dialog.

src/gcp/iam.ts
```diff
@@ -4,6 +4,16 @@ import { Client } from "../apiv2";
 
 const apiClient = new Client({ urlPrefix: iamOrigin(), apiVersion: "v1" });
 
+/** Returns the default cloud build service agent */
+export function getDefaultCloudBuildServiceAgent(projectNumber: string): string {
```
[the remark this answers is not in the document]
inlined: Prevoiusly gcb had the helpers for the default compute service agent as well. Historically we've thrown common IAM concerns (where possible) in the IAM folder. We could create a `src/gcp/computeengine` file that literally only has one const if you desire.

If we move into gcb and gce files, I would suggest that we shorten the functions to `getDefaultServiceAgent`

src/apphosting/secrets/dialogs.ts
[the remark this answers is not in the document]
inlined: Hmmm... Given the fact that the build account needs an extra role I see what this was doing. I'll reconcile. This will be annoying and I need to go rehearse for I/O, so I'd like to get sign-off on what I have with a pinky promise to reconcile/fix this when I continue with the change tonight.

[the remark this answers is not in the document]
tonyjhuang: sg, lets ship this for now and we can revisit later, esp much of this will change with the pending service account API updates

src/apphosting/secrets/dialogs.ts
tonyjhuang: sg. Personally, I would assume users would understand this point when they call `secrets:grantAccess`, but happy to re-eval later

src/gcp/iam.ts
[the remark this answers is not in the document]
tonyjhuang: Separating out by service files feels more extensible to me but I'll leave it up to you

src/apphosting/secrets/dialogs.ts
tonyjhuang: We also reconfirm the backend in `backends:delete` if a backend id is passed in but the `location` is omitted. I'll leave it up to you here

tonyjhuang (approved):
Left a few followup comments but overall lgtm thanks for the change!

src/apphosting/secrets/dialogs.ts
inlined: Oh, forgot to mention here that `every` short circuits but `map` does not. It's a trivial optimization, but seems correct to use in this case.

PR #6940, opened 2024-04-01
FAH ensure secret

inlined:
Part two of secrets:set. Adds a utility to ensure a secret is created

### DIFF
--- a/src/apphosting/secrets/index.ts
+++ b/src/apphosting/secrets/index.ts
@@ -10,48 +10,51 @@ import * as prompt from "../../prompt";
  * If a secret exists, we verify the user is not trying to change the region and verifies a secret
  * is not being used for both functions and app hosting as their garbage collection is incompatible
  * (client vs server-side).
+ * @returns true if a secret was created, false if a secret already existed, and null if a user aborts.
  */
 export async function upsertSecret(
   project: string,
   secret: string,
   location?: string,
-): Promise<boolean> {
+): Promise<boolean | null> {
+  let existing: gcsm.Secret;
   try {
-    const existing = await gcsm.getSecret(project, secret);
-    const replication = existing.replication?.userManaged;
-    if (
-      location &&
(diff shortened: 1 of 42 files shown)


src/apphosting/secrets/index.ts
```diff
@@ -0,0 +1,57 @@
...
+      });
+      if (!stopTracking) {
+        return false;
+      }
+      delete existing.labels[FIREBASE_MANAGED];
+      await gcsm.patchSecret(project, secret, existing.labels);
+    }
+    // TODO: consider whether we should prompt a user who has an unmanaged secret to enroll in version control.
+    // This may not be a great idea until version control is actually implemented.
+  } catch (err: any) {
```
tonyjhuang, line 50: can we update this try catch to just wrap L20 to keep the try block focused? As it stands now, if `patchSecret` on L46 errors out for whatever reason, the user gets an unexpected error message

src/init/features/apphosting/secrets.ts
```diff
@@ -73,6 +76,7 @@ export async function grantSecretAccess(
   }
 
   try {
+    // TODO: Merge with existing bindings with the same role
```
tonyjhuang, line 79: This is intentional, IAM will dedupe/coalesce roles fo rus

tonyjhuang (approved):
lgtm besides a couple nits, feel free to merge after addressing

src/init/features/apphosting/secrets.ts
inlined: Would like to verify. If so, I believe this is new.

abhis3: I can confirm I verified this manually when testing the CLI

You can see that in the original version of the commit there was an `updateBindings` method https://github.com/firebase/firebase-tools/pull/6899/commits/0fb470c874e9c7a31c56d623af8195080f098909#diff-1de7c7b0fb2e58db23edd1a7905f884d38b83d93d51ed5edd39a0c082cc3de49L78-L98

PR #6941, opened 2024-04-01
Add FAH config library

inlined:
Basic FAH config library


tonyjhuang (approved):
small nit but feel free to submit after

PR #6944, opened 2024-04-01
Block web browser requests to export endpoint

joehan:
### Description
Block web browser requests to export endpoint. 

### Scenarios Tested
Verified that calls from within the CLI still work by running `firebase emulators:start --export-on-exit ./exports`
Verifed that calls from a browser do not work by spinning up a basic Hosting site that calls this endpoint, and running it in the hosting emulator.


src/emulator/hub.ts
```diff
@@ -98,6 +98,11 @@ export class EmulatorHub extends ExpressBasedEmulator {
     });
 
     app.post(EmulatorHub.PATH_EXPORT, async (req, res) => {
+      if (req.headers.origin) {
+        res.status(401).json({
```
yuchenshi, line 102: This should be 403 since additional authentication (like a header wouldn't help)

yuchenshi (approved):

src/emulator/hub.ts
yuchenshi:
```suggestion
        res.status(403).json({
```

PR #6946, opened 2024-04-02
Reconcile existing secrets management code.

inlined:
Updated the original grant access code to accept a secret and a MultiServiceAccounts so that it can be used to grant access to multiple accounts at the same time. Logic for verifying a secret exists or getting a backend's service account are trivial and were moved into the command.

Updated the original secrets dialog to use the new MultiServiceAccounts type so that we can remember which accounts need the special access for Cloud Build to pin versions.

### DIFF
--- a/src/apphosting/secrets/dialogs.ts
+++ b/src/apphosting/secrets/dialogs.ts
@@ -1,34 +1,17 @@
 import * as clc from "colorette";
 const Table = require("cli-table");
 
+import { MultiServiceAccounts, ServiceAccounts, serviceAccountsForBackend, toMulti } from ".";
 import * as apphosting from "../../gcp/apphosting";
 import * as prompt from "../../prompt";
-import * as gce from "../../gcp/computeEngine";
-import * as cloudbuild from "../../gcp/cloudbuild";
 import * as utils from "../../utils";
 import { logger } from "../../logger";
 
 interface BackendMetadata {
   location: string;
   id: string;
-  accounts: string[];
-}
-
-/**
- * Finds the explicit service account used for a backend or, for legacy cases,
- * the defaults for GCB and compute.
- */
-export function serviceAccountsForBackend(
-  projectNumber: string,
-  backend: apphosting.Backend,
-): string[] {
-  if (backend.serviceAccount) {
-    return [backend.serviceAccount];
-  }
-  return [
-    cloudbuild.getDefaultServiceAccount(projectNumber),
-    gce.getDefaultServiceAccount(projectNumber),
-  ];
+  build: string;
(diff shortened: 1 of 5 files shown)


src/apphosting/secrets/index.ts
```diff
@@ -3,70 +3,74 @@ import * as iam from "../../gcp/iam";
...
-  runServiceAccount: string;
-} {
-  // TODO: For now we will always return the default CBSA and CESA. When the getBackend call supports returning
-  // the attached service account in a given backend/location then return that value instead.
-  // Sample Call: await apphosting.getBackend(projectId, location, backendId); & make this function async
-  return {
-    buildServiceAccount: gcb.getDefaultServiceAccount(projectNumber),
-    runServiceAccount: gce.getDefaultServiceAccount(projectNumber),
-  };
+/** Interface for holding the service account pair for a given Backend. */
```
tonyjhuang, line 12: While this feels cleaner, we may be overengineering here. Adding these interfaces (and the methods supporting them) will add more code surface to refactor when the backend changes land (which should be imminently). Although it isn't perfect, I would actually recommend treating run and build service accounts the same to save us from having to refactor all of these methods _again_ EOW

ie `serviceAccountsforBackend(): [string]`

tonyjhuang (changes requested):

src/apphosting/secrets/index.ts
tonyjhuang: Chatted offline, we will revisit this after the demo

tonyjhuang (approved):

src/apphosting/secrets/index.ts
inlined: Technically this makes the other suggested change redundant, but oh well. Per offline conversation we'll evaluate and change after the current stack is committed, otherwise I worry that there will be a _beast_ of a merge reconciliation in the downstream branches.

FWIW, I was worried because I wasn't sure that private build pools would bind to a specific service account. From a quick glance, it does not seem to be the case, so we can steer into just using a single list of service accounts in the future.

src/apphosting/secrets/dialogs.ts
```diff
@@ -1,34 +1,17 @@
...
-): string[] {
-  if (backend.serviceAccount) {
-    return [backend.serviceAccount];
-  }
-  return [
-    cloudbuild.getDefaultServiceAccount(projectNumber),
-    gce.getDefaultServiceAccount(projectNumber),
-  ];
+  build: string;
+  run: string;
```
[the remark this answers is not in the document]
inlined: Done (though using camelCase)

PR #6947, opened 2024-04-02
Finally wrap up the secrets:set command 🎉 

inlined:
Implement the full spec of the `firebase apphosting:secrets:set` command.

Confirmed this works by checking the Cloud Console to see that the secret was created with the correct permissions and that apphosting.yaml was created. Did not test a deploy because I am not aware whether the bug where values aren't propagated to runtime environments has been fixed.

This will need further polish. E.g. we need to discuss what to do in a non-interactive environment or when `--force` is set. We may also have to update the YAML handling to make sure we're only doing delta edits to minimize git diffs.

### DIFF
--- a/src/apphosting/config.ts
+++ b/src/apphosting/config.ts
@@ -12,21 +12,14 @@ export interface RunConfig {
   maxInstances?: number;
 }
 
-interface HasSecret {
-  secret: string;
-  value: never;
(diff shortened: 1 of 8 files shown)


abhis3 (approved):
🚀🚀🚀

PR #6948, opened 2024-04-02
Add support for apphosting:secrets:describe, alias for functions:secrets:get to align with gcloud

abhis3:
### Description

Add support for apphosting:secrets:describe
Alias for functions:secrets:get to align with gcloud

### DIFF
--- a/src/commands/apphosting-secrets-describe.ts
+++ b/src/commands/apphosting-secrets-describe.ts
@@ -17,7 +17,6 @@ export const command = new Command("apphosting:secrets:describe <secretName>")
   .action(async (secretName: string, options: Options) => {
     const projectId = needProjectId(options);
     const versions = await listSecretVersions(projectId, secretName);
-    console.log(versions[0]);
 
     const table = new Table({
       head: ["Name", "Version", "Status", "Create Time"],
--- a/src/commands/apphosting-secrets-describe.ts
+++ b/src/commands/apphosting-secrets-describe.ts
@@ -1,0 +1,30 @@
+import { Command } from "../command";
+import { Options } from "../options";
+import { needProjectId } from "../projectUtils";
+import { logger } from "../logger";
+import { requireAuth } from "../requireAuth";
+import { listSecretVersions } from "../gcp/secretManager";
+import * as secretManager from "../gcp/secretManager";
+import { requirePermissions } from "../requirePermissions";
+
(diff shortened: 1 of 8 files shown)


src/commands/apphosting-secrets-describe.ts
```diff
@@ -0,0 +1,29 @@
...
+    const versions = await listSecretVersions(projectId, secretName);
+
+    const table = new Table({
+      head: ["Name", "Version", "Status", "Create Time"],
+      style: { head: ["yellow"] },
+    });
+    for (const version of versions) {
+      table.push([secretName, version.versionId, version.state, version.createTime]);
+    }
+    logger.info(table.toString());
```
joehan, line 28: Applies for all 3 commands: These command should support the global `--json` flag, so that they can easily be used as part of scripts. To do this, you just need to return a JSON representation of the values displayed.

See https://github.com/firebase/firebase-tools/blob/master/src/commands/ext-dev-list.ts for an example of what this looks like

src/commands/apphosting-secrets-describe.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const Table = require("cli-table");
+
+export const command = new Command("apphosting:secrets:describe <secretName>")
+  .description("Get metadata for secret and its versions.")
+  .before(requireAuth)
+  .before(secretManager.ensureApi)
+  .before(requirePermissions, ["secretmanager.secrets.get"])
+  .action(async (secretName: string, options: Options) => {
+    const projectId = needProjectId(options);
```
joehan, line 18: It looks like `apphosting:secrets:describe`, `functions:secrets:describe`, and `functions:secrets:get` are all identical commands. Big picture, the API could maybe be simplified to a single `secrets:describe` command. 

Small picture, we should define this action function once and use it in all 3 commands.

joehan (changes requested):

src/commands/apphosting-secrets-describe.ts
abhis3: Agreed re:long term, this is a priority in the future as we continue to merge the functions and apphosting secrets functionality.

`apphosting:secrets:describe` has a different table so going to keep it as is but I'll extract the `functions:secrets` stuff out

joehan (approved):

PR #6949, opened 2024-04-02
Print message to view Firestore database in console after creation or restore

rwhogg:
### Description

After you create a Firestore database, we will now display a link to that database in the Firebase console. 

We also do this after you restore a Firestore database, but also note that you should wait until the restore is finished before using it.


VicVer09 (approved):
LGTM, pending rebase and a more detailed description with some screenshots

joehan (approved):

PR #6952, opened 2024-04-02
Remove distinction between build and run accounts

inlined:
Luckily the distinction was easier to remove than to write!

### DIFF
--- a/src/apphosting/secrets/dialogs.ts
+++ b/src/apphosting/secrets/dialogs.ts
@@ -1,7 +1,7 @@
 import * as clc from "colorette";
 const Table = require("cli-table");
 
-import { MultiServiceAccounts, ServiceAccounts, serviceAccountsForBackend, toMulti } from ".";
+import { serviceAccountsForBackend } from ".";
 import * as apphosting from "../../gcp/apphosting";
 import * as prompt from "../../prompt";
 import * as utils from "../../utils";
@@ -13,8 +13,7 @@ import * as env from "../../functions/env";
 interface BackendMetadata {
   location: string;
   id: string;
-  buildServiceAccount: string;
-  runServiceAccount: string;
+  serviceAccounts: string[];
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/apphosting/secrets/dialogs.ts
```diff
@@ -27,7 +27,11 @@ export function toMetadata(
   for (const backend of backends) {
     // Splits format projects/<unused>/locations/<location>/backends/<id>
     const [, , , location, , id] = backend.name.split("/");
-    metadata.push({ location, id, serviceAccounts: serviceAccountsForBackend(projectNumber, backend) });
+    metadata.push({
+      location,
+      id,
+      serviceAccounts: serviceAccountsForBackend(projectNumber, backend),
```
tonyjhuang, line 33: is updating this field to be a `Set` an option?

src/apphosting/secrets/dialogs.ts
```diff
@@ -39,7 +43,10 @@ export function toMetadata(
 }
 
 const matchesServiceAccounts = (target: BackendMetadata) => (test: BackendMetadata) => {
-  return target.serviceAccounts.length === test.serviceAccounts.length && target.serviceAccounts.every(sa => test.serviceAccounts.indexOf(sa) != -1);
+  return (
+    target.serviceAccounts.length === test.serviceAccounts.length &&
+    target.serviceAccounts.every((sa) => test.serviceAccounts.indexOf(sa) !== -1)
```
tonyjhuang, line 48: Having BackendMetadata.serviceAccounts be a `Set` could help making equality checks a little more straightforward. Otherwise, we end up with an order-dependent equality check which _may_ be unexpected

tonyjhuang (approved):
Thanks for the refactor! Left an optional suggestion for you

tonyjhuang (changes requested):
Actually I see there are some issues in the apphosting commands files, see src/commands/apphosting-secrets-set.ts and src/commands/apphosting-secrets-grantaccess.ts

PR #6956, opened 2024-04-03
Properly create a new secret when using apphosting:secrets:set

abhis3:
### Description

Fixes a bug where `apphosting:secrets:set` did not create a secret version after creating the underlying secret.

### DIFF
--- a/src/commands/apphosting-secrets-set.ts
+++ b/src/commands/apphosting-secrets-set.ts
@@ -40,6 +40,7 @@ export const command = new Command("apphosting:secrets:set <secretName>")
   )
   .action(async (secretName: string, options: Options) => {
     const howToAccess = `You can access the contents of the secret's latest value with ${clc.bold(`firebase apphosting:secrets:access ${secretName}`)}`;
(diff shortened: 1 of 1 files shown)


src/commands/apphosting-secrets-set.ts
```diff
@@ -62,22 +63,27 @@ export const command = new Command("apphosting:secrets:set <secretName>")
...
     const accounts = await dialogs.selectBackendServiceAccounts(projectNumber, projectId, options);
+
     // If we're not granting permissions, there's no point in adding to YAML either.
     if (!accounts.buildServiceAccounts.length && !accounts.runServiceAccounts.length) {
-      logWarning(
-        `To use this secret your backend, you must grant access. You can do so in the future with ${clc.bold("firebase apphosting:secrets:grantAccess")}`,
-      );
+      logWarning(grantAccess);
+
+      // TODO: For existing secrets, enter the grantSecretAccess dialog only when the necessary permissions don't exist.
```
inlined, line 86: Don't know how we'll ever know this since we don't know what backends this maps code to but 🤷

inlined (approved):

PR #6957, opened 2024-04-03
support for apphosting:secrets:access

abhis3:
### Description

Add support for apphosting:secrets:access


blidd-google (approved):
lookz gud 2 me

PR #6958, opened 2024-04-03
Await dangling promise

inlined:
<shame cube/>


inlined:
Dup of #6956. Closing.

PR #6969, opened 2024-04-05
App Hosting - Allow users to manage their github installation when "missing a repo" option is chosen

mathu97:
This is a bug fix. Currently when the `Missing a repo? Select this option to configure your GitHub connection settings` option is chosen a new connection is re-created with existing oauth credentials so nothing happens.

Now, when user selects this option a new connection is created but without the GitHub Installation ID, this sends user to the Dev Connect installation management page from which they can create a new installation with another GitHub org or account, or change the repo access settings for an existing GitHub app installation.

### DIFF
--- a/src/apphosting/githubConnections.ts
+++ b/src/apphosting/githubConnections.ts
@@ -116,6 +116,7 @@ export async function linkGitHubRepository(
           location,
           generateConnectionId(),
           oauthConn,
+          /** withNewInstallation= */ true,
         ),
       );
     }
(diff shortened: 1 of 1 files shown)


src/apphosting/githubConnections.ts
```diff
@@ -149,9 +150,10 @@ async function createFullyInstalledConnection(
   location: string,
   connectionId: string,
   oauthConn: devConnect.Connection,
+  withNewInstallation = false,
 ): Promise<devConnect.Connection> {
   let conn = await createConnection(projectId, location, connectionId, {
-    appInstallationId: oauthConn.githubConfig?.appInstallationId,
+    appInstallationId: withNewInstallation ? undefined : oauthConn.githubConfig?.appInstallationId,
```
tonyjhuang, line 156: can you add a param comment to the docstring on when you would want to set this and what the default behavior is?

tonyjhuang (approved):
LGTM left one comment

Issue #6975, opened 2024-04-09
can't download firestore emulator due to quota project error


leemhenson:
### [REQUIRED] Environment info

**firebase-tools:13.6.1**

**Platform:macOS**

### [REQUIRED] Test case

Use firestore emulator.

### [REQUIRED] Steps to reproduce

`firebase.json` contents:

```json
{
  "emulators": {
    "auth": {
      "host": "0.0.0.0",
      "port": 9099
    },
    "functions": {
      "host": "0.0.0.0",
      "port": 5001
    },
    "firestore": {
      "host": "0.0.0.0",
      "port": 8080
    },
    "storage": {
      "host": "0.0.0.0",
      "port": 9199
    },
    "ui": {
      "enabled": true
    },
    "singleProjectMode": true
  },
  "extensions": {
    "firestore-send-email": "firebase/firestore-send-email@0.1.31"
  },
  "firestore": {
    "rules": "firestore.rules",
    "indexes": "firestore.indexes.json"
  },
  "functions": [
    {
      "source": ".",
      "runtime": "nodejs20",
      "isolate": true,
      "codebase": "default",
      "ignore": [
        "node_modules",
        ".git",
        "firebase-debug.log",
        "firebase-debug.*.log"
      ]
    }
  ],
  "storage": {
    "rules": "storage.rules"
  }
}
```

### [REQUIRED] Expected behavior

I should be able to successfully boot the emulators.

### [REQUIRED] Actual behavior

```
firestore: downloading cloud-firestore-emulator-v1.19.4.jar...

Error: download failed, status 401: <?xml version='1.0' encoding='UTF-8'?><Error><Code>AuthenticationRequiredRequesterPays</Code><Message>Requester pays bucket access requires authentication.</Message></Error>
```

I can fix the issue by patching the call to `request` in `downloadToTmp` so that it passes the `ignoreQuotaProject: true` option. I don't know if there are any negative implications for that.

https://github.com/firebase/firebase-tools/blob/master/src/downloadUtils.ts#L20-L26

leemhenson:
Yes we do have `GOOGLE_CLOUD_QUOTA_PROJECT` set - I think this was to work around an issue in the past that was very similar to this one. I have removed my patch, unset that env var and deleted the cached emulator jar and yes the emulators are downloaded correctly now.

aalej:
[...]

PR #6976, opened 2024-04-09
add ignoreQuotaProject flag to downloadToTmp

leemhenson:
### Description

Fix for issue reported here: https://github.com/firebase/firebase-tools/issues/6975

The presence of a `GOOGLE_CLOUD_QUOTA_PROJECT` env var causes the firestore emulator download to fail. 

### Scenarios Tested

Tested with and without the environment variable.


joehan (approved):

joehan:
Thanks for the contribution!

Issue #6651, opened 2023-12-29
Angular SSR 17 in Firebase : listen EADDRINUSE: address already in use :::8080


alexcibotari:
### [REQUIRED] Environment info

**firebase-tools:** 13.0.2

**Platform:** Windows, Linux

### [REQUIRED] Test case

I deploy a fully workable version from the emulator in Firebase Functions of the Angular SSR application.

### [REQUIRED] Steps to reproduce

- Enabled experiment webframeworks.
- Create an application angular ssr from cli with standalone componenet
- run it locally
- deploy it in the firebase.
During Local build I have next warning 

````console
▲ [WARNING] Module 'undici' used by 'node_modules/@firebase/auth/dist/node-esm/index.js' is not ESM
▲ [WARNING] Module '@grpc/grpc-js' used by 'node_modules/@firebase/firestore/dist/index.node.mjs' is not ESM
▲ [WARNING] Module '@grpc/proto-loader' used by 'node_modules/@firebase/firestore/dist/index.node.mjs' is not ESM
▲ [WARNING] Module 'undici' used by 'node_modules/@firebase/functions/dist/esm-node/index.node.esm.js' is not ESM
▲ [WARNING] Module 'undici' used by 'node_modules/@firebase/storage/dist/node-esm/index.node.esm.js' is not ESM
  CommonJS or AMD dependencies can cause optimization bailouts.
  For more information see: https://angular.io/guide/build#configuring-commonjs-dependencies
````

### [REQUIRED] Expected behavior

Should work exactly as in the emulator.

### [REQUIRED] Actual behavior

SSR Page returns: Service Unavailable

Logs 

````console
Default STARTUP TCP probe succeeded after 1 attempt for container "ssr-project-name-1" on port 8080.

Uncaught exception

Error: listen EADDRINUSE: address already in use :::8080
    at Server.setupListenHandle [as _listen2] (node:net:1872:16)
    at listenInCluster (node:net:1920:12)
    at Server.listen (node:net:2008:7)
    at te.listen (file:///workspace/dist/ssr-project-name/server/server.mjs:81:3210)

at $k (file:///workspace/dist/ssr-project-name/server/server.mjs:99:5579)
at file:///workspace/dist/ssr-project-name/server/server.mjs:99:5665
at ModuleJob.run (node:internal/modules/esm/module_job:217:25)
at async ModuleLoader.import (node:internal/modules/esm/loader:316:24)
[...]

PR #6977, opened 2024-04-09
fix: throw an error when server.ts calls run function

9kubczas4:
### Description

There was a following change in main server file template in Angular 17:

```ts
// Webpack will replace 'require' with '__webpack_require__'
// '__non_webpack_require__' is a proxy to Node 'require'
// The below code is to ensure that the server is run only when not requiring the bundle.
declare const __non_webpack_require__: NodeRequire;
const mainModule = __non_webpack_require__.main;
const moduleFilename = mainModule && mainModule.filename || '';
if (moduleFilename === __filename || moduleFilename.includes('iisnode')) {
  run();
}
```

to

```ts
run();
```

The guard to call `run` function was removed. In the result there is an issue when someone tries to deploy Angular 17+ app with SSR.

I've added a check in the deployment script, if `run` function is called in `server.ts` then error with message: `For SSR to work properly, please remove the `run` function call from the `server.ts` file.` will be thrown.

The solution will provide better guidance to resolve this issue: https://github.com/firebase/firebase-tools/issues/6651.

### Scenarios Tested

Angular 17+:
- commented out `run` call in `server.ts` - deployment was successful
- removed `run` call in `server.ts` - deployment was successful
- `server.ts` has got `run` call and SSR is enabled in `angular.json` - an error with guidance was thrown
- `server.ts` has got `run` call and SSR is disabled in `angular.json` - deployment was successful


9kubczas4:
Hi @jamesdaniels,

What do you think about this change?
It could help with problem mentioned in this issue: https://github.com/firebase/firebase-tools/issues/6651.

When server.ts will contain `run()` call then meaningful error will be thrown.

PR #6981, opened 2024-04-10
Adding support for <product>_EMULATOR_BINARY_PATH for local emualtor dev

joehan:
### Description
Added  new environment variables `<product>_EMULATOR_BINARY_PATH`.

`<product>_EMULATOR_BINARY_PATH` tells the emulator suite to skip downloading/checking a downloadable emulator and instead use the binary at the provided path.

### Sample Commands
```
// Replace FIRESTORE with any other product, or UI
export FIRESTORE_EMULATOR_BINARY_PATH=path/to/firestore/emulator

// And the emulator suite will use your local version at that path
firebase emulators:start --only firestore 
```


src/emulator/downloadableEmulators.ts
```diff
@@ -323,7 +323,12 @@ export function _getCommand(
   args: { [s: string]: any },
 ): DownloadableEmulatorCommand {
   const baseCmd = Commands[emulator];
-
+  // If a commandOverride is passed, we use that instead but pass in the optional args.
+  const commandOverride = process.env[`${emulator.toUpperCase()}_EMULATOR_COMMAND`];
+  if (commandOverride) {
+    baseCmd.binary = commandOverride;
+    baseCmd.args = [];
```
joehan, line 330: @yuchenshi could you sanity check me on this? I'm pretty sure we need to omit these required args for the 'blaze run' case to work - take a look at the bae commands for the java emulators, and see how we pass in java args that won't work with 'blaze run' IIUC.

yuchenshi (approved):

PR #6986, opened 2024-04-11
Have tests folder architecture match src folder.

rrousselGit:
Also renames `firematSuite/Test` to `firebaseSuite/Test`


firebase-vscode/src/test/suite/project.test.ts
```diff
@@ -1,23 +0,0 @@
-import assert from "assert";
```
rrousselGit, line 1: This isn't deleted, but moved. Git is drunk here

hlshen (approved):

PR #6987, opened 2024-04-11
Edit configs at the AST level to preserve formatting and comments

inlined:
Improves FAH secrets set command with AST-based YAML editing.

Replaces js-yaml in our codebase with yaml for consistency (though nowhere else needs the AST API). Unfortunately, js-yaml is still a dependency for our tooling, so it's still in package-lock.json

Want review from Tony for FAH stuff and Joe for CLI dependency changes + double checking extensions changes


src/apphosting/config.ts
```diff
@@ -1,8 +1,9 @@
 import * as path from "path";
 import { writeFileSync } from "fs";
-import * as yaml from "js-yaml";
+import * as yaml from "yaml";
 
 import * as fs from "../fsutils";
+import { NodeType } from "yaml/dist/nodes/Node";
```
joehan, line 6: This seems like an internal type for the library, and they explicitly don't use it in their public API. https://github.com/eemeli/yaml/blob/main/src/public-api.ts

Do you need this type? Could this be refactored not to use it?

joehan (commented):
Overall, change looks good, extensions code looks fine, and `yaml` seems like a clearly better library to use than `js-yaml`.  I'd love if you could get rid of the usage of NodeType though.

src/apphosting/config.ts
inlined: I think this is an oversight for it not to be exported. AFAICT none of the types are exported (note in public-api, Node is imported, and is a mandatory base class of Document<T, Strict> but is not exported in that file either.

Frankly, the type annotations are a mess; I had it working and removed it all.  I could hypothetically remove the API call, but it would instead involve yet more casting as any/unknown.

src/apphosting/config.ts
```diff
@@ -55,12 +56,45 @@ export function yamlPath(cwd: string): string | null {
 }
 
 /** Load apphosting.yaml */
-export function load(yamlPath: string): Config {
+export function load(yamlPath: string): yaml.Document {
```
tonyjhuang, line 59: optional nit but these methods all feel like good candidates to wrap in a `yaml.ts` that can be shared across firebase products.

Especially the main logic in getEnv and setEnv could likely be genericized and `get/setEnv` can use those helpers.

src/apphosting/config.ts
```diff
@@ -55,12 +56,45 @@ export function yamlPath(cwd: string): string | null {
...
 }
 
 /** Save apphosting.yaml */
-export function store(yamlPath: string, config: Config): void {
-  writeFileSync(yamlPath, yaml.dump(config));
+export function store(yamlPath: string, document: yaml.Document): void {
+  writeFileSync(yamlPath, document.toString());
+}
+
+export function getEnv(document: yaml.Document, variable: string): Env | undefined {
```
tonyjhuang, line 69: jsdocs for exported functions please

tonyjhuang, line 69: consider renaming to findEnv

src/apphosting/config.ts
```diff
@@ -55,12 +56,45 @@ export function yamlPath(cwd: string): string | null {
...
+  if (!document.has("env")) {
+    document.set("env", document.createNode([env]));
+    return;
+  }
+  const envs = document.get("env") as yaml.YAMLSeq<NodeType<Env>>;
+
+  // The type system in this library is... not great at propagating type info
+  const envYaml = document.createNode(env);
+  for (let i = 0; i < envs.items.length; i++) {
+    if ((envs.items[i].get("variable") as unknown) === env.variable) {
```
tonyjhuang, line 92: This is very nitpicky but line items in yaml are not guaranteed to be unique. This will replace the first instance of `env.variable` but will leave duplicates in place

src/commands/apphosting-secrets-set.ts
```diff
@@ -90,12 +91,14 @@ export const command = new Command("apphosting:secrets:set <secretName>")
 
     // Note: The API proposal suggested that we would check if the env exists. This is stupidly hard because the YAML may not exist yet.
     let path = config.yamlPath(process.cwd());
-    let yaml: config.Config = {};
+    let projectYaml: yaml.Document;
```
tonyjhuang, line 94: As this method gets longer, I would recommend breaking it out into helper functions to help users more easily parse through the top level control flow, as we do here: https://github.com/firebase/firebase-tools/blob/master/src/apphosting/index.ts#L99

tonyjhuang (changes requested):

src/apphosting/config.ts
inlined: Generally it's better to build abstraction layers once you actually have a second use for the abstraction. Here, loading a document is 2LOC (can be one). get/setEnv are 10 and 18LOC respectively. I'm frankly not even sure how I'd factor them into a utility that isn't just a more skillful usage of the yaml library itself (e.g. maybe you want me to add a push() method?).

Factoring out YAML wrapper over the library while we're still building an understanding of how to use the library well reeks of premature optimization.

src/apphosting/config.ts
```diff
@@ -98,3 +102,46 @@ export function setEnv(document: yaml.Document, env: Env): void {
 
   envs.add(envYaml);
 }
+
+export async function maybeAddSecretToYaml(secretName: string): Promise<void> {
```
tonyjhuang, line 106: jsdoc for exported function please

tonyjhuang (approved):
LGTM except one comment

PR #6988, opened 2024-04-11
Cleanup apphosting flows

mathu97:
Changes to apphosting command descriptions and wording.

### DIFF
--- a/src/apphosting/githubConnections.ts
+++ b/src/apphosting/githubConnections.ts
@@ -230,7 +230,7 @@ async function promptCloneUri(
   const cloneUri = await promptOnce({
     type: "autocomplete",
     name: "cloneUri",
-    message: `Which of the following repositories would you like to deploy? ${clc.bold("(Use arrow keys or type to search)")}`,
+    message: `Which of the following repositories would you like to deploy?`,
     source: (_: any, input = ""): Promise<(inquirer.DistinctChoice | inquirer.Separator)[]> => {
       return new Promise((resolve) =>
         resolve([
--- a/src/apphosting/githubConnections.ts
+++ b/src/apphosting/githubConnections.ts
@@ -230,7 +230,7 @@ async function promptCloneUri(
   const cloneUri = await promptOnce({
     type: "autocomplete",
     name: "cloneUri",
-    message: `Which of the following repositories would you like to deploy?`,
+    message: "Which repository would you like to deploy?",
     source: (_: any, input = ""): Promise<(inquirer.DistinctChoice | inquirer.Separator)[]> => {
       return new Promise((resolve) =>
         resolve([
--- a/src/apphosting/githubConnections.ts
+++ b/src/apphosting/githubConnections.ts
@@ -230,7 +230,7 @@ async function promptCloneUri(
   const cloneUri = await promptOnce({
     type: "autocomplete",
     name: "cloneUri",
-    message: "Which of the following repositories would you like to deploy?",
+    message: `Which of the following repositories would you like to deploy? ${clc.yellow("(Use arrow keys or type to search)")}`,
     source: (_: any, input = ""): Promise<(inquirer.DistinctChoice | inquirer.Separator)[]> => {
       return new Promise((resolve) =>
         resolve([
--- a/src/apphosting/githubConnections.ts
+++ b/src/apphosting/githubConnections.ts
@@ -230,7 +230,7 @@ async function promptCloneUri(
   const cloneUri = await promptOnce({
     type: "autocomplete",
     name: "cloneUri",
-    message: `Which of the following repositories would you like to deploy? ${clc.yellow("(Use arrow keys or type to search)")}`,
+    message: `Which of the following repositories would you like to deploy? ${clc.bold("(Use arrow keys or type to search)")}`,
     source: (_: any, input = ""): Promise<(inquirer.DistinctChoice | inquirer.Separator)[]> => {
       return new Promise((resolve) =>
         resolve([
--- a/src/apphosting/index.ts
+++ b/src/apphosting/index.ts
@@ -1,5 +1,3 @@
-import * as clc from "colorette";
-
 import * as repo from "./repo";
 import * as poller from "../operation-poller";
 import * as apphosting from "../gcp/apphosting";
@@ -69,9 +67,7 @@ export async function doSetup(
       name: "region",
       type: "list",
       default: DEFAULT_REGION,
-      message:
-        "Please select a region " +
-        `(${clc.yellow("info")}: Your region determines where your backend is located):\n`,
+      message: "Select a region to host your backend",
       choices: allowedLocations.map((loc) => ({ value: loc })),
     })) as string);
 
--- a/src/apphosting/index.ts
+++ b/src/apphosting/index.ts
@@ -135,7 +135,6 @@ export async function doSetup(
 
   logSuccess(`Successfully created backend:\n\t${backend.name}`);
   logSuccess(`Your backend is now deployed at:\n\thttps://${backend.uri}`);
-  logBullet("A certificate is being provisioned for your backend, this may take up to ten minutes");
 }
 
 /**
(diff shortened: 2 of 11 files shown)


src/commands/apphosting-secrets-set.ts
```diff
@@ -63,17 +63,12 @@ export const command = new Command("apphosting:secrets:set <secretName>")
...
-    }
-
     const version = await gcsm.addVersion(projectId, secretName, secretValue);
     logSuccess(`Created new secret version ${gcsm.toSecretVersionResourceName(version)}`);
-    logSuccess(howToAccess);
+    logBullet(howToAccess);
 
     // If the secret already exists, we want to exit once the new version is added
     if (!created) {
-      logWarning(grantAccess);
```
mathu97, line 76: Instead of logging this when the secret is not created. We should try to check if the service accounts already have access. If they don't then we should log. 

When the user creates a new secret but declines to grant access to the service accounts then following versions of the secret will not have access either.

mathu97: @tonyjhuang let me know what you think?

src/commands/apphosting-secrets-set.ts
```diff
@@ -39,7 +39,7 @@ export const command = new Command("apphosting:secrets:set <secretName>")
     'File path from which to read secret data. Set to "-" to read the secret data from stdin.',
   )
   .action(async (secretName: string, options: Options) => {
-    const howToAccess = `You can access the contents of the secret's latest value with ${clc.bold(`firebase apphosting:secrets:access ${secretName}`)}`;
+    const howToAccess = `You can access the contents of the secret's latest value with ${clc.bold(`firebase apphosting:secrets:access ${secretName}\n`)}`;
     const grantAccess = `To use this secret in your backend, you must grant access. You can do so in the future with ${clc.bold("firebase apphosting:secrets:grantAccess")}`;
```
tonyjhuang, line 43: secrets:grantaccess  not secrets:grantAccess

src/commands/apphosting-secrets-set.ts
```diff
@@ -39,7 +39,7 @@ export const command = new Command("apphosting:secrets:set <secretName>")
     'File path from which to read secret data. Set to "-" to read the secret data from stdin.',
   )
   .action(async (secretName: string, options: Options) => {
-    const howToAccess = `You can access the contents of the secret's latest value with ${clc.bold(`firebase apphosting:secrets:access ${secretName}`)}`;
+    const howToAccess = `You can access the contents of the secret's latest value with ${clc.bold(`firebase apphosting:secrets:access ${secretName}\n`)}`;
```
tonyjhuang, line 42: I know this was already here but would prefer to inline these comments: see go/java-practices/constants#when-to-use-a-constant-versus-an-inlined-literal

Relevant passage:
"""
If, however, the value is used just once, or its meaning is self-explanatory in context, prefer inlining the literal. The indirection involved in using the constant makes its usage slightly less readable (when one wants to see the value), so it should be done only if it's worth it.
"""

This is the java style guide but it's widely applicable imo

src/commands/apphosting-secrets-set.ts
tonyjhuang: IMO this current flow seems reasonable to me. If the secret previously existed and the user is adding a new version, it seems appropriate to expect the secret is already correctly provisioned and we can exit out early

tonyjhuang (approved):
left a few small nits

src/commands/apphosting-backends-delete.ts
```diff
@@ -10,7 +10,7 @@ import { printBackendsTable } from "./apphosting-backends-list";
 
 export const command = new Command("apphosting:backends:delete <backend>")
   .description("delete a backend from a Firebase project")
```
tonyjhuang, line 12: update description?

src/commands/apphosting-secrets-set.ts
mathu97: ah good catch!

src/commands/apphosting-secrets-set.ts
mathu97: +1 on inlining, I was thinking the same.

src/apphosting/index.ts
```diff
@@ -122,7 +122,7 @@ export async function doSetup(
 
   if (!confirmRollout) {
     logSuccess(`Successfully created backend:\n\t${backend.name}`);
-    logSuccess(`Your site will be deployed at:\n\thttps://${backend.uri}`);
+    logSuccess(`Your backend will be deployed at:\n\thttps://${backend.uri}`);
```
mathu97, line 125: @tonyjhuang the wording "site" feels odd because we aren't using that anywhere else. I think "app" or "backend" would make more sense. Let me know what you think.

src/apphosting/index.ts
```diff
@@ -135,7 +135,8 @@ export async function doSetup(
   });
 
   logSuccess(`Successfully created backend:\n\t${backend.name}`);
-  logSuccess(`Your site is now deployed at:\n\thttps://${backend.uri}`);
+  logSuccess(`Your backend is now deployed at:\n\thttps://${backend.uri}`);
+  logBullet("A certificate is being provisioned for your backend, this may take up to ten minutes");
```
mathu97, line 139: @tonyjhuang let me know what you think of the wording here.

mathu97: nvm removing this as @inlined is working on this.

src/apphosting/githubConnections.ts
```diff
@@ -230,7 +230,7 @@ async function promptCloneUri(
   const cloneUri = await promptOnce({
     type: "autocomplete",
     name: "cloneUri",
-    message: "Which of the following repositories would you like to deploy?",
+    message: `Which of the following repositories would you like to deploy?`,
```
tonyjhuang, line 233: Optional terse check: "Which repository would you like to deploy?"

src/apphosting/index.ts
tonyjhuang: backend sgtm

src/apphosting/repo.ts
```diff
@@ -214,7 +214,7 @@ async function promptRepositoryUri(
   const remoteUri = await promptOnce({
     type: "autocomplete",
     name: "remoteUri",
-    message: "Which of the following repositories would you like to deploy?",
+    message: `Which of the following repositories would you like to deploy?`,
```
tonyjhuang, line 217: Same suggestion as above

src/apphosting/secrets/dialogs.ts
```diff
@@ -143,37 +142,38 @@ export async function selectBackendServiceAccounts(
...
 
   const metadata: BackendMetadata[] = toMetadata(projectNumber, listBackends.backends);
 
   if (metadata.every(matchesServiceAccounts(metadata[0]))) {
-    utils.logLabeledBullet(
-      "apphosting",
-      "To use this secret, your backend's service account must have secret accessor permission. All of your backends use " +
-        (sameServiceAccount(metadata[0]) ? "service account " : "service accounts ") +
+    utils.logBullet(
+      "To use this secret, your backend's service account must have secret accessor permission.",
```
tonyjhuang, line 153: This feels awkward to me for some reason. "To use this secret, your backend's service account must be granted access."?

src/apphosting/secrets/dialogs.ts
```diff
@@ -143,37 +142,38 @@ export async function selectBackendServiceAccounts(
...
+        (sameServiceAccount(metadata[0]) ? "service account: " : "service accounts: ") +
         serviceAccountDisplay(metadata[0]) +
-        ". Granting access to one backend will grant access to all backends.",
+        ".\nGranting access to one backend will grant access to all backends.",
     );
     const grant = await prompt.confirm({
       nonInteractive: options.nonInteractive,
       default: true,
-      message: "Would you like to grant it now?",
+      message: "Would you like to grant access now?",
```
tonyjhuang, line 164: Grant access to all backends now?

tonyjhuang (approved):
LGTM barring a few (optional) nits

PR #6993, opened 2024-04-12
Add various missing "dispose" calls

rrousselGit:
### Description

While working on configs, I spotted a bunch of resources that weren't released. It shouldn't have a huge impact, but that's probably better anyway.

### DIFF
--- a/firebase-vscode/webviews/globals/html-broker.ts
+++ b/firebase-vscode/webviews/globals/html-broker.ts
@@ -1,22 +1,72 @@
+import { useEffect, useState } from "react";
 import { Broker, createBroker } from "../../common/messaging/broker";
 import {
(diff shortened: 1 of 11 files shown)


firebase-vscode/webviews/globals/html-broker.ts
```diff
@@ -1,22 +1,72 @@
+import { useEffect, useState } from "react";
 import { Broker, createBroker } from "../../common/messaging/broker";
 import {
   ExtensionToWebviewParamsMap,
   WebviewToExtensionParamsMap,
 } from "../../common/messaging/protocol";
 
+export function useBrokerListener<
```
rrousselGit, line 8: These were already added in the private repository but I forgot to sync those here.

hlshen (approved):

PR #6994, opened 2024-04-12
update FAH root dir prompt

tonyjhuang:

### DIFF
--- a/src/apphosting/index.ts
+++ b/src/apphosting/index.ts
@@ -93,7 +93,7 @@ export async function doSetup(
     type: "input",
     default: "/",
     message:
-      "Specify the path to the root of the app you would like to deploy (relative to the repo root)",
+      "Specify your app's root directory relative to your repository",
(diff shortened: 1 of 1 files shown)


src/apphosting/index.ts
```diff
@@ -92,8 +92,7 @@ export async function doSetup(
     name: "rootDir",
     type: "input",
     default: "/",
-    message:
-      "Specify the path to the root of the app you would like to deploy (relative to the repo root)",
+    message: "Specify your app's root directory relative to your repository",
```
blidd-google, line 95: "Relative to your repository root" maybe? Don't feel strongly about this though

blidd-google (approved):

src/apphosting/index.ts
tonyjhuang: Oop sorry, had auto merge turned on. I'll keep it as is for now

PR #6996, opened 2024-04-12
Poll till apphosting backend deletion completes

mathu97:
We should be polling the long running operation till completion.


src/apphosting/index.ts
```diff
@@ -361,3 +361,19 @@ export async function orchestrateRollout(
   }
   return { rollout, build };
 }
+
+/**
+ * Deletes the given backend. Polls till completion.
+ */
+export async function deleteBackend(
```
tonyjhuang, line 368: nit: `deleteBackendAndPoll()` if you want to stay consistent with ``createBackendAndPoll()` above

tonyjhuang, line 368: test?

tonyjhuang (changes requested):

tonyjhuang (approved):

PR #7000, opened 2024-04-15
Handle malformed config files.

rrousselGit:
#6998 should be merged first

### Description

Currently, if a config file is missing or broken, the extension fails to start.
This PR refactors the config parsing
to handle errors.

In terms of UI, I've experimented with:
- Showing the error in the sidebar
- Showing an error message using a toast

I think a toast feels nicer:

Screenshot 2024-04-15 at 12 16 26


firebase-vscode/src/core/config.ts
```diff
@@ -8,11 +8,18 @@ import { RC, RCData } from "../../../src/rc";
...
+import { ValueOrError } from "../../common/messaging/protocol";
 import { onChange } from "../utils/signal";
-import { pluginLogger } from "../logger-wrapper";
+import { Result, ResultError } from "../result";
+import { FirebaseConfig } from "../firebaseConfig";
+import { effect } from "@preact/signals-react";
 
-export const firebaseRC = globalSignal<RC | undefined>(undefined);
-export const firebaseConfig = globalSignal<Config | undefined>(undefined);
+export const firebaseRC = globalSignal<Result<RC | undefined> | undefined>(
```
hlshen, line 17: imo this type system is getting more complicated than it needs to be.

firebase-vscode/src/core/config.ts
```diff
@@ -39,10 +46,24 @@ export async function updateFirebaseRCProject(
   rc.save();
 }
 
+
 function notifyFirebaseConfig(broker: ExtensionBrokerImpl) {
   broker.send("notifyFirebaseConfig", {
-    firebaseJson: firebaseConfig.value?.data,
-    firebaseRC: firebaseRC.value?.data,
+    firebaseJson: firebaseConfig.value?.switchCase<
```
hlshen, line 52: similar thoughts here. Is there enough value prop to warrant this level of code to return a value. Is it not sufficient to error out on read and maintain firebaseConfig as value | undefined.

firebase-vscode/src/core/config.ts
```diff
@@ -120,33 +161,32 @@ export function _readRC(): RC | undefined {
...
       return undefined;
-    } else {
-      pluginLogger.error(e.message);
-      throw e;
     }
+
+    return config;
+  });
+
+  if (result instanceof ResultError && (result.error as any).status === 404) {
```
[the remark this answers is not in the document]
rrousselGit: This differentiates between "No config file" from "There's a config file, but its content is incorrect".

This enables the UI to react differently based on this information. "No file" would prompt for `firebase init` once that's implemented, but an error would just show a toast about incorrect configs

firebase-vscode/src/core/config.ts
rrousselGit: `value | undefined` is not enough.
We have 3 separate states:
- no value because configs are still lloading
- configs are loaded, but the config file is missing
- configs are loaded, and the file is malformed


The current type differentiates all those states.

`undefined` means "Configs" are loading
`ResultValue(undefined)` means configs loading has completed, and the file doesn't exist
`ResultError(...)` means there's an error in the config file.

firebase-vscode/src/core/config.ts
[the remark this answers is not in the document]
hlshen: ohhh I didn't realize this line was a specific result error. So only on 404 where we get no value back. I guess that's self explanatory but I'd appreciate some comments generally in this file.

firebase-vscode/src/core/config.ts
hlshen: comments in this function would be appreciated too

hlshen (approved):

PR #7001, opened 2024-04-15
Run vscode tests in the CI

rrousselGit:


.github/workflows/node-test.yml
```diff
@@ -38,6 +38,34 @@ jobs:
       - run: npm i -g npm@9.5
       - run: npm ci
       - run: npm run lint:changed-files
+      - run: npm run lint
+        working-directory: firebase-vscode
+
+  vscode_unit:
+    runs-on: windows-latest
```
rrousselGit, line 45:
```suggestion
    # Using windows to bypass an error thrown by VScode
    # when run in an environment that does not have a screen.
    runs-on: windows-latest
```

hlshen (approved):
approved, going to let Joe take a look at this too

.github/workflows/node-test.yml
```diff
@@ -38,6 +38,36 @@ jobs:
...
+        node-version:
+          - "18"
+          - "20"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: npm
+          cache-dependency-path: npm-shrinkwrap.json
```
joehan, line 59: Should this be firebase-vscode/[package-lock.json instead (or in addition)? I assume we want to cache the vscode deps here.

.github/workflows/node-test.yml
```diff
@@ -38,6 +38,36 @@ jobs:
...
+
+      - run: npm i -g npm@9.5
+      - run: npm ci
```
joehan, line 62: Double checking - if you remove this line, do the tests still pass?

joehan (approved):
Some questions/potential minor efficiency improvements, but very glad to be adding these to CI.

.github/workflows/node-test.yml
rrousselGit, line 62:
```suggestion
```

.github/workflows/node-test.yml
```diff
@@ -38,6 +38,35 @@ jobs:
...
+        node-version:
+          - "18"
+          - "20"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: npm
+          cache-dependency-path: npm-shrinkwrap.json
```
rrousselGit, line 59:
```suggestion
          cache-dependency-path: firebase-vscode/package-lock.json
```

.github/workflows/node-test.yml
rrousselGit: I doubt it. We have to run `npm i` in both the root and the vscode folder.
The `npm i -g` above doesn't count as it only updates the npm version and nothing else.

.github/workflows/node-test.yml
```diff
@@ -38,6 +38,35 @@ jobs:
...
+
+      - run: npm i -g npm@9.5
```
rrousselGit, line 61:
```suggestion
      - run: npm i -g npm@9.5
      - run: npm ci
```

.github/workflows/node-test.yml
```diff
@@ -38,6 +38,36 @@ jobs:
...
+      - run: npm ci
+
```
rrousselGit, line 63:
```suggestion
```

PR #7003, opened 2024-04-15
upgrade dependencies in test project

bkendall:
### Description

Does what it says on the box. Upgrades dependencies in an (empty) test project. If tests pass, should be safe enough?


joehan (approved):
Thanks!

PR #7011, opened 2024-04-16
Improve getting started experience

inlined:
1. Tells customers they can ^C once the rollout has been created
2. Turns the rollout in progress message to a spinny
3. Adds a spinny to wait on TLS to be ready. To be forwards compatible with the day that we can just serve all default domains on a wildcard cert, this dialog only pops up after an initial check verifies that TLS is not yet ready.

Not sure about some of the prose I've moved about. Feel encouraged to nitpick and/or add other nitpickers as reviewers.

### DIFF
--- a/src/apphosting/index.ts
+++ b/src/apphosting/index.ts
@@ -148,7 +148,6 @@ export async function doSetup(
   });
 
   await setDefaultTrafficPolicy(projectId, location, backendId, branch);
-  logSuccess(`Successfully created backend:\n\t${backend.name}`);
 
   const confirmRollout = await promptOnce({
     type: "confirm",
@@ -169,7 +168,7 @@ export async function doSetup(
   // TODO: Previous versions of this command printed the URL before the rollout started so that
   // if a user does exit they will know where to go later. Should this be re-added?
   const createRolloutSpinner = ora(
-    "Starting a new rollout; this make take a few minutes. It's safe to exit now.",
+    "Starting a new rollout; this may take a few minutes. It's safe to exit now.",
   ).start();
   await orchestrateRollout(projectId, location, backendId, {
     source: {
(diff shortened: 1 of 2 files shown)


inlined:
Reduced scope to just awaiting TLS now that other changes have been implemented. Changed the rollout text a bit since we may to continue waiting for TLS. Also, I removed a "..." because it seems redundant/passive aggressive along with a spinner.

src/apphosting/index.ts
```diff
@@ -121,6 +148,7 @@ export async function doSetup(
   });
 
   await setDefaultTrafficPolicy(projectId, location, backendId, branch);
+  logSuccess(`Successfully created backend:\n\t${backend.name}`);
```
tonyjhuang, line 151: Remove, this is already on L138

src/apphosting/index.ts
```diff
@@ -147,7 +178,13 @@ export async function doSetup(
       },
     },
   });
-  createRolloutSpinner.succeed(`Your backend is now deployed at:\n\thttps://${backend.uri}`);
+  createRolloutSpinner.succeed("Rollout complete");
+  if (!(await tlsReady(url))) {
+    const tlsSpinner = ora("Waiting for TLS certificate");
```
tonyjhuang, line 183: you probably want tlsSpinner = ora(...).start()

src/apphosting/index.ts
```diff
@@ -134,11 +162,14 @@ export async function doSetup(
...
 
+  const url = `https://${backend.uri}`;
   logBullet(
     `You may also track this rollout at:\n\t${consoleOrigin()}/project/${projectId}/apphosting`,
   );
+  // TODO: Previous versions of this command printed the URL before the rollout started so that
+  // if a user does exit they will know where to go later. Should this be re-added?
   const createRolloutSpinner = ora(
-    "Starting a new rollout... This make take a few minutes. It's safe to exit now.",
+    "Starting a new rollout; this make take a few minutes. It's safe to exit now.",
```
tonyjhuang, line 172: typo: this may take

src/apphosting/index.ts
tonyjhuang, line 183: "Finalizing your backend's TLS certificate; this may take a few minutes."

tonyjhuang (approved):
Left some nits

src/apphosting/index.ts
inlined: ugh. merge issues. I had previously moved and now it got turned into an add. TY for the catch.

src/apphosting/index.ts
inlined: Whoops! Maybe Ora isn't as easy to use as I thought!

PR #7025, opened 2024-04-21
Update action-hosting-deploy version

178inaba:
### Description

FirebaseExtended/action-hosting-deploy runtime has been upgraded to node20[^1].
Since it has not been released, we have modified it to specify the version using a hash value[^2].

[^1]: https://github.com/FirebaseExtended/action-hosting-deploy/commit/ac8041b3b04337509168113bf98b95879df22322
[^2]: https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-a-commits-sha-for-release-management

### Scenarios Tested

I tested it manually.
I used the official documentation for the steps.
https://firebase.google.com/docs/hosting/github-integration

### Sample Commands

There are no command or flag changes.

### DIFF
--- a/src/experiments.ts
+++ b/src/experiments.ts
@@ -223,7 +223,7 @@ export function assertEnabled(name: ExperimentName, task: string): void {
           "FIREBASE_CLI_EXPERIMENTS",
         )} environment variable to ${filename}, like so: ${italic(`
 
-- uses: FirebaseExtended/action-hosting-deploy@v0
+- uses: FirebaseExtended/action-hosting-deploy@ac8041b3b04337509168113bf98b95879df22322
   with:
     ...
   env:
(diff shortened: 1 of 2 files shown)


oyvindwe:
This is a poor solution, and not forward compatible. Could we please get a release of the GitHub action instead?

178inaba:
@oyvindwe 
I am not authorized to release `FirebaseExtended/action-hosting-deploy`.

I also hope that a release will take place.
Because it can be fixed in `@v1` once released.
However, the release seems to be delayed.
Therefore, I temporarily fixed it in this way to avoid the warning.

178inaba:
Since action-hosting-deploy has been released, I will close this pull request.
https://github.com/FirebaseExtended/action-hosting-deploy/releases/tag/v0.7.2

PR #7041, opened 2024-04-23
add manual check for turtle service accounts

bkendall:
### Description

Instead of checking for an error message coming back from the API, we can proactively check for service accounts and their permissions. This change adjusts the flow to be:

1. test for `actAs` permissions on the service account
2. create the service account (with the appropriate role) if it did not exist
3. create the backend

This should be significantly more reliable since we won't be relying on error message parsing.

### DIFF
--- a/src/test/apphosting/index.spec.ts
+++ b/src/test/apphosting/index.spec.ts
@@ -11,12 +11,12 @@ import {
   deleteBackendAndPoll,
   promptLocation,
   setDefaultTrafficPolicy,
+  ensureAppHostingComputeServiceAccount,
 } from "../../apphosting/index";
 import * as deploymentTool from "../../deploymentTool";
 import { FirebaseError } from "../../error";
 
-describe("operationsConverter", () => {
-  const sandbox: sinon.SinonSandbox = sinon.createSandbox();
+describe("apphosting setup functions", () => {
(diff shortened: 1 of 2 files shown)


src/test/apphosting/index.spec.ts
```diff
@@ -108,68 +110,6 @@ describe("operationsConverter", () => {
...
-        location,
-        backendId,
-        cloudBuildConnRepo,
-        /* serviceAccount= */ null,
-        webAppId,
-      );
-
-      // CreateBackend should be called twice; once initially and once after the service account was created
-      expect(createBackendStub).to.be.calledTwice;
-      expect(createServiceAccountStub).to.be.calledOnce;
```
kmcnellis, line 138: Shouldn't we still have these expects?       
expect(createServiceAccountStub).to.be.calledOnce;
expect(addServiceAccountToRolesStub).to.be.calledOnce;

kmcnellis (approved):

kmcnellis (approved):

src/test/apphosting/index.spec.ts
bkendall: Done.

mathu97 (approved):

PR #7068, opened 2024-04-29
Cleaning up changes from Data Connect merge

joehan:
### Description
Cleaned up the public API for apiv2 and some errant pacakge.json entries.

 Specifically: 
- Made 'BaseRequesOptions.path' and 'ClientOptions.urlPrefix' required again.
- Removed getHeaders, and moved that functionality to FbToolsAuthClient
- Made `client.getAccessToken()` private again.
- Made `accessToken` and `refreshToken` public

I think this approach leaves the public API in a cleaner state. I did need to expose 'access/refreshToken' - however these were already referenced in the public API through their setters.

Thanks @bkendall for most of these suggestions.

### Scenarios Tested
Still successfully able to authenticate calls through the cloudSQL connector.


src/apiv2.ts
```diff
@@ -312,18 +304,7 @@ export class Client {
...
-    const map: Record<string, string> = {};
-    for (const h of opts.headers!) {
-      map[h[0]] = h[1];
-    }
-    return map;
-  }
-
-  public async getAccessToken(): Promise<string> {
+  private async getAccessToken(): Promise<string> {
     // Runtime fetch of Auth singleton to prevent circular module dependencies
```
bkendall, line 308: I think this comment is obsolete now!

src/apiv2.ts
```diff
@@ -312,18 +304,7 @@ export class Client {
...
-    await this.addAuthHeader(opts);
-    const map: Record<string, string> = {};
-    for (const h of opts.headers!) {
-      map[h[0]] = h[1];
-    }
-    return map;
-  }
-
-  public async getAccessToken(): Promise<string> {
+  private async getAccessToken(): Promise<string> {
```
bkendall, line 307: Maybe it's better to yoink this as an exported helper - that way you don't have to re-implement it in your Client, and you don't have to directly export the private variables. Thoughts? It doesn't need anything from Client, so it should be simple to do.

then it looks kinda neat, that you could do  `apiv2.getAccessToken()` on a whim.

src/gcp/cloudsql/fbToolsAuthClient.ts
```diff
@@ -33,13 +37,21 @@ export class FBToolsAuthClient extends AuthClient {
...
-    return client.getHeaders();
+
+  public async getRequestHeaders(): Promise<Record<string, string>> {
+    const token = await this.getAccessToken();
+    return {
+      Connection: "keep-alive",
+      "User-Agent": `FirebaseCLI/${CLI_VERSION}`,
+      "X-Client-Version": `FirebaseCLI/${CLI_VERSION}`,
+      Authorization: `Bearer ${token.token}`,
+    };
```
bkendall, line 55: For my money, maybe make these first three either a static Map that you can iterate through and return here (adding Authorization). Then you can have apiv2 iterate over the same list.

Added bonus, you could then have a "I only want these headers" type list so that if future things are added to `apiv2.StandardHeadersMap`, you could filter them out pretty easily.

bkendall (changes requested):

PR #7069, opened 2024-04-29
Firebase init updates

sam-gc:


src/init/features/genkit.ts
```diff
@@ -0,0 +1,66 @@
...
+      },
+    ]);
+    if (!setup.continueFunctions) {
+      logger.info("Stopped Genkit initialization");
+      return;
+    }
+
+    // Functions with genkit should always be typescript
+    setup.languageOverride = "typescript";
+    await functionsSetup(setup, config, options);
```
joehan, line 29: Can you test how this behaves when initing multiple products at once? Particularly, i'd be concerned about initing gen kit + python or javascript functions

joehan, line 29: What about cases where I am not initing functions but I already am in a firebase directory with typescript functions? Does this work as you want?

src/init/features/genkit.ts
```diff
@@ -0,0 +1,66 @@
...
+  } catch (e) {
+    logger.error("Genkit initialization failed...");
+    return;
+  }
+
+  logger.info("To use the Genkit CLI, run:");
+  logger.info("    npm install genkit -g");
+}
+
+function wrapSpawn(cmd: string, args: string[], projectDir: string): Promise<void> {
```
joehan, line 46: This code duplicates a lot of https://github.com/firebase/firebase-tools/blob/master/src/init/features/functions/npm-dependencies.ts#L6 - consider separating out a common helper and using it in both places/

src/init/features/genkit.ts
sam-gc: If you're already in a firebase directory with typescript functions, this will prompt you to initialize a new codebase because the default (`functions/`) is taken. This is expected and wanted; `genkit init` assumes you're in a clean node environment

sam-gc: Yep -- it works in that case. Both if the directory already exists and if you select both functions and genkit from the init menu. Functions seems to always init first but I've added a `delete setup.languageOverride` to ensure it still works even in reverse

src/init/features/genkit.ts
sam-gc: Done.

joehan (approved):

PR #7077, opened 2024-04-30
Ensure APIs and required permissions on all Data Connect commands

joehan:
### Description
For all FDC commands, ensure the required APIs are enabled, and check for required permissions beforehand. These checks helps avoid partial deployments that fail halfway thru due to a single missing permission or API.

Eventually, I'd like to move some of these CloudSQL permission checks to near their call site so that people who don't have <insert sensitive cSQL access here> can still deploy Data Connect. This seems relevant for enterprise-type teams who likely want to limit cSQL connect/create access/

### Scenarios Tested
Sanity checked that commands all still work, and new APIs will be enabled successfully.


src/commands/deploy.ts
```diff
@@ -47,7 +47,29 @@ export const TARGET_PERMISSIONS: Record<(typeof VALID_DEPLOY_TARGETS)[number], s
...
+    "cloudsql.instances.get",
+    "cloudsql.instances.update",
+    "cloudsql.instances.create",
+    "cloudsql.instances.connect",
+    "cloudsql.databases.create",
+    "cloudsql.databases.update",
+    "cloudsql.users.create",
+    "firebasedataconnect.connectors.create",
+    "firebasedataconnect.connectors.delete",
+    "firebasedataconnect.operations.get",
```
lisajian, line 61: Can you move this up / down so the connectors permissions are together?

src/commands/deploy.ts
```diff
@@ -47,7 +47,29 @@ export const TARGET_PERMISSIONS: Record<(typeof VALID_DEPLOY_TARGETS)[number], s
     "firebaserules.releases.update",
   ],
   remoteconfig: ["cloudconfig.configs.get", "cloudconfig.configs.update"],
-  dataconnect: [], // TODO: add required permissions
+  dataconnect: [
```
lisajian, line 50: Is this just supposed to be the smallest set of permissions that people will need to use the CLI?

src/commands/dataconnect-sql-migrate.ts
```diff
@@ -1,22 +1,28 @@
...
 
 export const command = new Command("dataconnect:sql:migrate [serviceId]")
   .description("migrates your CloudSQL database's schema to match your local DataConnect schema")
-  // .before(requirePermissions, ["dataconnect.services.list", "dataconnect.schemas.list", "dataconnect.connectors.list"])
+  .before(requirePermissions, [
+    "firebasedataconnect.services.list",
+    "firebasedataconnect.schemas.list",
+    "firebasedataconnect.schemas.update",
+    "cloudsql.instances.connect",
+    "cloudsql.users.create",
```
lisajian, line 19: Not familiar with the migration flow - what is the cloudsql.users.create permission used for?

src/commands/deploy.ts
joehan: Yes (or at least, specifically, the `firebase deploy` command). I know some of these won't be used in all cases (ie you won't need `cloudsql.instances.create` if you are using an existing instance) - I'd like to come back and refine this to support those cases later

src/commands/dataconnect-sql-migrate.ts
joehan: We need to create the IAM users that we connect to the DB as - see https://github.com/firebase/firebase-tools/blob/master/src/gcp/cloudsql/connect.ts#L113

lisajian (approved):
LGTM, just a nit about moving firebasedataconnect.operations.get

fredzqm (approved):

src/commands/dataconnect-sql-migrate.ts
fredzqm, line 19: Can we add a TODO to make `"cloudsql.users.create"` optional?

So just `roles/cloudsql.client` should theoretically be able to perform schema migration if already granted the access.

PR #7078, opened 2024-04-30
Add basic readme and icon for 'store' page.

hlshen:


firebase-vscode/package.json
```diff
@@ -1,7 +1,8 @@
 {
   "name": "firebase-vscode",
-  "displayName": "firebase-vscode",
+  "displayName": "Firebase Extension",
```
joehan, line 3: Prefer any other name to avoid clashing with the separate product Firebase Extensions. How about just Firebase?

joehan (approved):

PR #7079, opened 2024-04-30
update apphosting import to es6

hlshen:

### DIFF
--- a/src/apphosting/app.ts
+++ b/src/apphosting/app.ts
@@ -153,4 +153,4 @@ function isQuotaError(error: any): boolean {
   return code === 429;
 }
 
-export = webApps;
+export default webApps;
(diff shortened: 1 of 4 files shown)


src/apphosting/app.ts
```diff
@@ -153,4 +153,4 @@ function isQuotaError(error: any): boolean {
   return code === 429;
 }
 
-export = webApps;
+export default webApps;
```
bkendall, line 156: default exports shouldn't (generally) be used: go/tsjs-style#exports

bkendall (changes requested):
Why is this necessary?

src/apphosting/app.ts
bkendall: why wouldn't it just be `export const webApps` way above?

hlshen: Changed to export const. 

I'm not sure why I can't respond to your other message, but the firebase-vscode targets module: "es2020" which restricts usage of export assignments.

bkendall: Hmm. firebase-tools's target could probably get updated one of these days too...

bkendall (approved):

mathu97 (approved):

PR #7083, opened 2024-05-01
Handle schema migration errors when validateOnly=false

joehan:
### Description
2 big fixes for schema migration here:
- First, we now handle schema migration errors when validateOnly=false
- Second, we will actually set role to `firebaseowner _dbid_public` before doing migrations, so that it owns any tables create

### Scenarios Tested
I tested that deploy triggers schema migration even on a fresh project
I also tested that I can login on one account, run a schema migration that creates tables, then log in on another account and migrate those tables.

### DIFF
--- a/src/dataconnect/schemaMigration.ts
+++ b/src/dataconnect/schemaMigration.ts
@@ -3,13 +3,12 @@ import { format } from "sql-formatter";
 
 import { IncompatibleSqlSchemaError, Diff } from "./types";
 import { upsertSchema } from "./client";
-import { execute, setupIAMUser } from "../gcp/cloudsql/connect";
+import { execute, firebaseowner, setupIAMUser } from "../gcp/cloudsql/connect";
 import { promptOnce } from "../prompt";
 import { logger } from "../logger";
 import { Schema } from "./types";
 import { Options } from "../options";
 import { FirebaseError } from "../error";
-import { REQUIRED_EXTENSIONS_COMMANDS } from "./provisionCloudSql";
 import { needProjectId } from "../projectUtils";
 import { logLabeledWarning } from "../utils";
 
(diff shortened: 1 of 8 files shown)


src/dataconnect/schemaMigration.ts
```diff
@@ -53,51 +54,74 @@ export async function migrateSchema(
   if (!instanceId) {
     throw new FirebaseError(`tried to migrate schema but ${instanceId} was undefined`);
   }
-  const iamUser = await setupIAMUser(instanceId, databaseId, options);
   try {
-    // TODO(b/330596914): Handle cases where error only comes back after validateOnly=false
-    await upsertSchema(schema, /** validateOnly=*/ true);
+    await upsertSchema(schema, validateOnly);
```
fredzqm, line 58:
```
if validateOnly && existingSchema.schemaValidation == "NONE" {
   // Call upsertSchema(existingSchemaWithSchemaValidationStrict, validateOnly=false)
   // This ensure the backend connectivity to the Cloud SQL instance is provisioned.
   // Without it, `upsertSchema(schema, validateOnly=true)` below may success with false positives
}
```

fredzqm: Can you make sure to test `dataconnect:sql:diff` on a fresh new instance created by Firebase Console?

joehan: Just tested, and it reports no differences. However, I'm not sure how to safely fix this - we can't call `upsertSchema(schema, validateOnly=true)` from sql:diff - if it succeeds, it would update the schema instead of just showing us the difference

joehan: Thanks for the offline chat - implemented your suggestion and verified that both diff and migrate work on the 'fresh  outta console' state.

fredzqm (approved):

PR #7084, opened 2024-05-01
Update dependencies in functions initialization template

blidd-google:
Updating dependencies to the latest version of the firebase-admin (v12.1.0) and firebase-functions (v5.0.0) SDKs.


joehan (approved):
LGTM - can you do this for the extensions templates too?

joehan (changes requested):
Add a changelog please

joehan (approved):

PR #7088, opened 2024-05-01
add fah tos check to backends:create

tonyjhuang:


src/commands/apphosting-backends-create.ts
```diff
@@ -23,6 +25,7 @@ export const command = new Command("apphosting:backends:create")
   )
   .before(ensureApiEnabled)
   .before(requireInteractive)
+  .before(requireTosAcceptance(APPHOSTING_TOS_ID))
```
joehan, line 28: Sanity check that this works when you're running as a service account (`firebase logout`, then `export GOOGLE_APPLICATION_CREDENTIALS=path/to/key.json`, then hit this)

I'm not sure how firedata's api works when the caller is not a user.

src/requireTosAcceptance.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Example:
+ *   new Command(...)
+ *     .description(...)
+ *     .before(requireTosAcceptance(APPHOSTING_TOS_ID)) ;
+ *
+ * Note: When supporting new products, be sure to update `consoleLandingPage` above to avoid surfacing
+ * generic ToS error messages.
+ **/
+export default function requireTosAcceptance(tosId: TosId): (options: Options) => Promise<void> {
```
joehan, line 22: Prefer not using default exports go/tsjs-style#exports

joehan (changes requested):

src/commands/apphosting-backends-create.ts
joehan: Following up from internal convo - maybe add a check inside requireTosAcceptance to skip the call if this is a service account?

tonyjhuang: According to Tyler, Firedata will soon reply with ACCEPTED for all service account tos checks

src/requireTosAcceptance.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+async function requireTos(tosId: TosId): Promise<void> {
+  const res = await getTosStatus();
+  if (isProductTosAccepted(res, tosId)) {
+    return;
+  }
+  const console = consoleLandingPage.get(tosId) || consoleOrigin();
+  throw new FirebaseError(
+    `Your account is missing the required terms of service for this action. Please accept the Terms of Service and try again. ${console}`,
```
joehan, line 33: nit: s/"is missing"/"has not accepted"

src/test/gcp/firedata.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    nock.enableNetConnect();
+  });
+
+  describe("getTosStatus", () => {
+    it("should return parsed GetTosStatusResponse", () => {
+      nock("https://mobilesdk-pa.googleapis.com")
+        .get("/v1/accessmanagement/tos:getStatus")
+        .reply(200, SAMPLE_RESPONSE);
+
+      expect(getTosStatus()).to.eventually.deep.equal(SAMPLE_RESPONSE as GetTosStatusResponse);
```
joehan, line 52: You need to either return this or await it: https://www.chaijs.com/plugins/chai-as-promised/ - as is, it is not actually running this assertion.

src/test/requireTosAcceptance.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    await requireTosAcceptance(APP_CHECK_TOS_ID)(SAMPLE_OPTIONS);
+  });
+
+  it("should throw error if not accepted", async () => {
+    nock("https://mobilesdk-pa.googleapis.com")
+      .get("/v1/accessmanagement/tos:getStatus")
+      .reply(200, SAMPLE_RESPONSE);
+
+    await expect(requireTosAcceptance(APPHOSTING_TOS_ID)(SAMPLE_OPTIONS)).to.be.rejectedWith(
+      "Your account is missing the required terms of service for this action. Please accept the Terms of Service and try again. https://console.firebase.google.com/project/_/apphosting",
```
joehan, line 71: Nit - As is, this is a pure change detector for the message string. I'd suggest narrowing this down to checking just some relevant portion of the string (ie: 'terms of service') so that it is less fragile.

joehan (approved):
A bunch of minor test issues that you shoudl fix, but code lgtm

PR #7096, opened 2024-05-02
Retry user creation when role is not ready

joehan:
### Description
(semi) Fix the issue you ran into yesterday - retry user creation  a few times with simple backoff when role is not ready.


src/gcp/cloudsql/cloudsqladmin.ts
```diff
@@ -162,28 +162,50 @@ export async function createUser(
...
-  );
-  const opName = `projects/${projectId}/operations/${op.body.name}`;
-  const pollRes = await operationPoller.pollOperation<User>({
-    apiOrigin: cloudSQLAdminOrigin(),
-    apiVersion: API_VERSION,
-    operationResourceName: opName,
-    doneFn: (op: Operation) => op.status === "DONE",
-  });
-  return pollRes;
+  const maxRetries = 3;
```
hlshen, line 165: looks fine, but surprised we don't have some template/package for retries?

hlshen (approved):

PR #7099, opened 2024-05-03
fix copied functions dist dir files for Next.js when source config ends with slash

### LINKED ISSUE #7062: Next.js SSR Deployment fails due to wrong firebase-tools creating wrong folder

13.7.5

**firebase-tools:**

macOS

**Platform:**

### [REQUIRED] Test case

Deployment of Next using Firebase Webframeworks due to firebase creating wrong ssr folder

run `firebase deploy --only hosting` when using next.js 

image

`.next`  should be in the a sub directory of the `functions`

I get `functions.next` instead. This then leads to firebase uploading the files without the ssr files

This then leads to a server error
Error: ENOENT: no such file or directory, open '/workspace/.next/BUILD_ID'

leoortizz:
### Description

`path.replace` was breaking when source dir was e.g. `web/`, causing the final folder to be `.hosting/functions.next` instead of `.hosting/functions/.next`. Fixes #7062

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,3 @@
 - Increased the timeout for waiting for emulators to start to 60s. (#7091)
 - Fixes infinite loop when trying to create a Hosting site.
+- Fix copied functions dist dir files for Next.js when source config ends with slash (#7099)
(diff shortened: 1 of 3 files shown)


jamesdaniels (approved):
Does this still work in windows?

leoortizz:
> Does this still work in windows?

yep, it also fixes the same issue on Windows. `join` FTW!

PR #7100, opened 2024-05-03
Improve App Hosting Web App selection UX

mathu97:
### Description

Firebase web app is created automatically during onboarding. The prompt for a web app is now removed.


src/apphosting/app.ts
```diff
@@ -34,22 +32,7 @@ async function getOrCreateWebApp(
...
-  }
-
-  const existingUserProjectWebApps = new Map(
-    webAppsInProject.map((obj) => [
-      // displayName can be null, use app id instead if so. Example - displayName: "mathusan-web-app", appId: "1:461896338144:web:426291191cccce65fede85"
-      obj.displayName ?? obj.appId,
-      obj.appId,
-    ]),
-  );
+  const existingUserProjectWebApps = firebaseAppsToMap(webAppsInProject);
```
tonyjhuang, line 35: Should we update the method name and comment?

mathu97: hmm i left the the name as is because it still gets the webapp if the user provides the web app via a flag. Wondering if we should remove this too.

tonyjhuang (approved):

PR #7104, opened 2024-05-03
Handle new schemas too.

joehan:
### Description
Silly oversight on my part - forgot to cover cases where the schema doesn't exist yet.

### DIFF
--- a/src/dataconnect/schemaMigration.ts
+++ b/src/dataconnect/schemaMigration.ts
@@ -178,7 +178,15 @@ async function promptForSchemaMigration(
 // the backend will not have the necesary permissions to check cSQL for differences.
 // We fix this by upserting the currently deployed schema with schemaValidation=strict,
 async function ensureServiceIsConnectedToCloudSql(serviceName: string) {
-  const currentSchema = await getSchema(serviceName);
+  let currentSchema
+  try {
+    currentSchema = await getSchema(serviceName);
+  } catch (err: any) {
+    if (err.status === 404) {
+      return;
+    }
+    throw err;
+  }
   if (
     !currentSchema.primaryDatasource.postgresql ||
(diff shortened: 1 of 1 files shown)


fredzqm (approved):

dconeybe (approved):

src/dataconnect/schemaMigration.ts
```diff
@@ -178,7 +178,15 @@ async function promptForSchemaMigration(
 // the backend will not have the necesary permissions to check cSQL for differences.
 // We fix this by upserting the currently deployed schema with schemaValidation=strict,
 async function ensureServiceIsConnectedToCloudSql(serviceName: string) {
-  const currentSchema = await getSchema(serviceName);
+  let currentSchema;
+  try {
+    currentSchema = await getSchema(serviceName);
+  } catch (err: any) {
+    if (err.status === 404) {
```
fredzqm, line 185: Second thought on this. This is gonna be tricky.

If some customers starts with empty schema, then `ensureServiceIsConnectedToCloudSql` wouldn't do it its job.

`sql:diff` and `sql:migrate` will run into the false positive bugs as well.

```
export ensureServiceIsConnectedToCloudSql(serviceName: string, dataSource: string) {
  try {
    currentSchema = await getSchema(serviceName);
    if (currentSchema.dataSource == dataSource) {
       throw new Error("cannot migrate SQL schema of a different Cloud SQL instance already linked. Please consider use a new service.");
    }
    // rest of the logics. upsertSchema(currentSchema, /** validateOnly=*/ false) if not already in Strict mode.
  } catch (err: any) {
    if (err.status === 404) {
       await upsertSchema(/*just datasource no source*/, /** validateOnly=*/ false);
       return;
    }
    throw err;
  }
]
```

PR #7110, opened 2024-05-05
@invertase/scheduled emulator

exaby73:

### DIFF
--- a/src/emulator/functionsEmulator.ts
+++ b/src/emulator/functionsEmulator.ts
@@ -64,6 +64,7 @@ import { setEnvVarsForEmulators } from "./env";
 import { runWithVirtualEnv } from "../functions/python";
 import { Runtime } from "../deploy/functions/runtimes/supported";
 import { CronJob } from "cron";
+import { ScheduledEmulator } from "./scheduledEmulator";
 
 const EVENT_INVOKE_GA4 = "functions_invoke"; // event name GA4 (alphanumertic)
 
@@ -129,7 +130,7 @@ export interface FunctionsEmulatorArgs {
  * IPC connection info of a Function Runtime.
  */
 export class IPCConn {
-  constructor(readonly socketPath: string) {}
+  constructor(readonly socketPath: string) { }
 
   httpReqOpts(): http.RequestOptions {
     return {
@@ -145,7 +146,7 @@ export class TCPConn {
   constructor(
     readonly host: string,
     readonly port: number,
-  ) {}
+  ) { }
 
   httpReqOpts(): http.RequestOptions {
     return {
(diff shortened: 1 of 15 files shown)


src/emulator/functionsEmulator.ts
```diff
@@ -1631,7 +1788,7 @@ export class FunctionsEmulator implements EmulatorInstance {
         return;
       }
     }
-    let debugBundle;
+    let debugBundle: any;
```
cabljac, line 1791: Would it be possible to remove this `any`? I think it could be typed as this?

https://github.com/firebase/firebase-tools/blob/bfb9c32188c272409b59d93e0bdd162a52f9c2c7/src/emulator/functionsEmulatorShared.ts#L88

src/emulator/scheduledEmulator.ts
```diff
@@ -0,0 +1,86 @@
...
+      // We assume that the schedule is a cron expression
+      const hasSeconds = schedule.split(" ").length === 6;
+      scheduleData = later.parse.cron(schedule, hasSeconds);
+    } else {
+      // We assume that the schedule is a text expression
+      scheduleData = later.parse.text(schedule);
+    }
+
+    if (scheduleData.schedules.length === 0) {
+      this.logger.log(
```
cabljac, line 76: Is it OK not to throw something here?

cabljac (commented):
Left a couple of comments, otherwise looks good to me

src/emulator/functionsEmulator.ts
cabljac: `let debugBundle: FunctionsRuntimeBundle[debug]` or something like that

src/emulator/scheduledEmulator.ts
exaby73: Something like:

```ts
    if (scheduleData.schedules.length === 0) {
      throw new FirebaseError(
        `Failed to parse ${isCron ? "cron" : "text"} schedule for timer ${id}: ${schedule}`,
      );
    }
```

scripts/triggers-end-to-end-tests/tests.ts
```diff
@@ -187,16 +187,8 @@ describe("function triggers", () => {
...
-      this.timeout(EMULATOR_TEST_TIMEOUT);
-
-      const response = await test.writeToScheduledPubsub();
-      expect(response.status).to.equal(200);
-      await new Promise((resolve) => setTimeout(resolve, EMULATORS_WRITE_DELAY_MS));
-    });
-
     it("should have have triggered cloud functions", () => {
-      expect(test.pubsubTriggerCount).to.equal(2);
+      expect(test.pubsubTriggerCount).to.equal(1);
```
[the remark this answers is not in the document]
exaby73: Although I am not completely sure, this may be because of https://github.com/firebase/firebase-tools/blob/35926f853478c959059fc97cea201d468dcb66b4/scripts/triggers-end-to-end-tests/tests.ts#L177-L183. This was always like this but some tests were removed therefore this change to `1`

src/emulator/scheduledEmulator.ts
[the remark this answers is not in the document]
exaby73: Reference to `scheduled` already exists in the codebase. For example: V1 scheduled. Hence we chose to call it Scheduled for consistency sake

PR #7114, opened 2024-05-06
App Hosting: Log the selected location only when user is prompted for one

mathu97:
### Description

This log looks out of place when the location is auto-selected for the user. Moved it so that it is logged only when the user is prompted to select a location.


tonyjhuang:
I understand the rationale but IMO it's still worth letting the user know which region we are selecting for them. WDYT?

PR #7119, opened 2024-05-07
Handle invalid connector errors

joehan:
### Description
Handle invalid connector errors - see b/331486949 for further discussion of backend improvements.

Also, some minor code cleanup - parallelized listServices calls because I was annoyed at how slow they were while testing, and cleaned up some logging inconsistencies.

### Scenarios Tested
Diff:
Screenshot 2024-05-06 at 8 31 23 PM

Migrate:
Screenshot 2024-05-06 at 8 31 08 PM

Deploy:
Screenshot 2024-05-06 at 8 31 46 PM
Screenshot 2024-05-06 at 8 31 56 PM


fredzqm (changes requested):

src/dataconnect/schemaMigration.ts
```diff
@@ -55,29 +56,35 @@ export async function migrateSchema(args: {
...
-    const diffs = await handleIncompatibleSchemaError({
-      ...args,
-      incompatibleSchemaError: incompatible,
-      instanceId,
-      databaseId,
-    });
-    // Then, try to upsert schema again. If there still is an error, just throw it now
-    await upsertSchema(schema, validateOnly);
-    return diffs;
+    if (invalid) {
```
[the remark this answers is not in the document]
fredzqm: I think the returning both invalid connectors and schema migration errors in the backend is at risk to make I/O.

We can do a stop gap solution to call `migrateSchema` twice in all callers, so the second call will capture the "invalid connector" errors. Still perform schema migration before recreate connectors, but this adds a prompt between (a) and (b) above. Acceptable stop gap while we improve backend.

src/dataconnect/errors.ts
```diff
@@ -0,0 +1,42 @@
...
+    throw err;
+  }
+  const details: any[] = original.details;
+  const incompatibles = details.filter((d) => d["@type"] === INCOMPATIBLE_SCHEMA_ERROR_TYPESTRING);
+  // Should never get multiple incompatible schema errors
+  return incompatibles[0];
+}
+
+export function isInvalidConnectorError(err: { message?: string }): boolean {
+  return err.message?.includes("The request was invalid: invalid connector") ?? false;
```
fredzqm, line 22: Relying on the error message is pretty bad.

LG for now. Though let me try to make backend return proper failed precondition error details and code "INVALID_CONNECTOR".

This change would probably make the cut off.

http://shortn/_9PGdmqCjVx

src/dataconnect/schemaMigration.ts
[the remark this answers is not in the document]
joehan: First, to clarify, this does not delete connectors during `sql:migrate` as is (since validateOnly=true). It would be reasonable to cut off before offering any database schema migration during 'sql:migrate' if there are invalid , I can make that change.

Good call on splitting the confirmations and the actual changes - will do so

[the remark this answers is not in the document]
joehan: Made a pretty large refactor here to match this new beahvior, PTAL. I tested a bunch of cases locally and the behavior now matches what you've described here.

src/dataconnect/errors.ts
joehan: Now using these new errors.

src/dataconnect/schemaMigration.ts
```diff
@@ -173,26 +213,84 @@ async function promptForSchemaMigration(
...
     }
-    throw err;
   }
   if (
     !currentSchema.primaryDatasource.postgresql ||
     currentSchema.primaryDatasource.postgresql.schemaValidation === "STRICT"
   ) {
-    // Only want to do this coming from console half deployed state. If the current schema is "STRICT" mode,
-    // or if there is not postgres attached, don't try this.
+    // If the current schema is "STRICT" mode, don't do thins.
```
fredzqm, line 293: `s/thins/this`

src/dataconnect/schemaMigration.ts
```diff
@@ -43,60 +46,97 @@ export async function migrateSchema(args: {
...
+      "tried to migrate schema but instance name was not provided in dataconnect.yaml",
+    );
   }
+  const instanceId = instanceName?.split("/").pop();
+
+  const serviceName = schema.name.replace(`/schemas/${SCHEMA_ID}`, "");
   try {
-    const serviceName = schema.name.replace(`/schemas/${SCHEMA_ID}`, "");
-    await ensureServiceIsConnectedToCloudSql(serviceName);
+    await ensureServiceIsConnectedToCloudSql(serviceName, instanceName, databaseId);
```
fredzqm, line 67: Code health wise, I think we can extract a helper 

```
const (instanceName, databaseName) = await ensureServiceIsConnectedToCloudSql(schema);
```

This logic is shared between `sql:diff` and `sql:migrate`.

Do it outside `try catch`, don't think errors from this method should be handled in the same way.

src/dataconnect/schemaMigration.ts
```diff
@@ -43,60 +46,97 @@ export async function migrateSchema(args: {
...
+    const migrationMode = incompatible
+      ? await promptForSchemaMigration(
+          options,
+          databaseId,
+          incompatible,
+          allowNonInteractiveMigration,
+        )
+      : "none";
+    // First, error out if we aren't making all changes
+    if (migrationMode === "none" && incompatible) {
```
fredzqm, line 97: No AI needed. 

Note that since backend checks for exact match of SQL schema. 

`migrationMode === "safe"` is in an awkward unusable state, since it will fail below:
```
    // Then, try to upsert schema again. If there still is an error, just throw it now
    await upsertSchema(schema, validateOnly);
```

Rosalyn's intern project "incompatible" mode would address this hole.

Ideas to improve it before there, we could suggest developers to use `deploy --force` to update FDC schema ignoring schema validation error. `schemaValidation=NONE`

Definitely out of scope for this PR. Post-IO polish idea.

fredzqm (approved):

src/dataconnect/schemaMigration.ts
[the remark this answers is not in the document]
fredzqm: Thanks a bunch for the refactor! Looks great!

src/dataconnect/schemaMigration.ts
fredzqm: [For follow up]

non-force `firebase deploy` should block on schema migration being fully done. This following prompt makes sense
```
    const choices = err.destructive
      ? [
          { name: "Execute all changes (including destructive changes)", value: "all" },
          // This set schemaValidation=NONE in the second `upsertSchema` 
          { name: "Skip any destructive changes", value: "safe" }, 
          { name: "Abort changes", value: "abort" }, // non-interactive mode behavior
        ]
      : [
          { name: "Execute changes", value: "safe" },
          { name: "Abort changes", value: "abort" }, // non-interactive mode behavior
        ];
```

Interactive `deploy --force` prompts as well, but should provide more options to force deploy ignoring schema diffs.

```
    const choices = err.destructive
      ? [
          { name: "Execute all changes (including destructive changes)", value: "all" },
          // This set schemaValidation=NONE in the second `upsertSchema` 
          { name: "Skip any destructive changes", value: "safe" }, 
          // This set schemaValidation=NONE in the second `upsertSchema` 
          { name: "Skip all changes", value: "skip" }, // non-interactive mode behavior
          { name: "Abort changes", value: "abort" },
        ]
      : [
          { name: "Execute changes", value: "safe" },
          // This set schemaValidation=NONE in the second `upsertSchema` 
          { name: "Skip all changes", value: "skip" }, // non-interactive mode behavior
          { name: "Abort changes", value: "abort" },
        ];
```

src/dataconnect/schemaMigration.ts
joehan: Good call - took a slightly different approach to have better separation of concerns, but still dryed this out

PR #7121, opened 2024-05-07
Improved data connect init

joehan:
### Description
- Use list APIs for choosing instance and DB.
- Clean up some todos in template files.
- Skip provisioning question if using existing instance and DB.
- Removed 'what directory do you want to use' question.

### Scenarios Tested
-New instance, new db
- Existing instance, existing db
- exisiting instance, new db


src/init/features/dataconnect/index.ts
```diff
@@ -52,29 +53,60 @@ export async function doSetup(setup: Setup, config: Config): Promise<void> {
...
+    }
+  }
+  if (cloudSqlInstanceId === "") {
+    newInstance = true;
+    cloudSqlInstanceId = await promptOnce({
+      message: `What ID would you like to use for your new CloudSQL instance?`,
+      type: "input",
+      default: `dataconnect-test`,
+    });
+  }
```
fredzqm, line 85: Playing around with the CLI a bit more. Just an idea for consideration.

For customers that just finished console onboard, they already have a CSQL instance created, would be it nice ask them to pick an existing cloud sql instance as the first question?

If they picked an existing instance, then no need to ask about region. Basically. Right now, need to remember the region of my instance in order to pick it in CLI.

PR #7129, opened 2024-05-08
App Hosting: Take optional web app Id instead of  web app name

mathu97:
There can be multiple Firebase apps with the same name. Need to use appId instead.

### DIFF
--- a/src/commands/apphosting-backends-create.ts
+++ b/src/commands/apphosting-backends-create.ts
@@ -11,7 +11,7 @@ export const command = new Command("apphosting:backends:create")
   .description("create a Firebase App Hosting backend")
   .option(
     "-a, --app <webAppId>",
-    "specify an existing Firebase web app to associate your App Hosting backend with",
+    "specify an existing Firebase web app's ID to associate your App Hosting backend with",
   )
   .option("-l, --location <location>", "specify the location of the backend", "")
   .option(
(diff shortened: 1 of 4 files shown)


src/commands/apphosting-backends-create.ts
```diff
@@ -10,7 +10,7 @@ import { requireTosAcceptance } from "../requireTosAcceptance";
 export const command = new Command("apphosting:backends:create")
   .description("create a Firebase App Hosting backend")
   .option(
-    "-a, --app <webApp>",
+    "-a, --app <webAppId>",
     "specify an existing Firebase web app to associate your App Hosting backend with",
```
tonyjhuang, line 14: nit: specify an existing Firebase web app's ID to ...

src/test/apphosting/app.spec.ts
```diff
@@ -42,7 +42,7 @@ describe("app", () => {
         webApps.getOrCreateWebApp(projectId, "nonExistentWebApp", backendId),
       ).to.be.rejectedWith(
         FirebaseError,
-        "The web app 'nonExistentWebApp' does not exist in project projectId",
+        "The web app with Id 'nonExistentWebApp' does not exist in project projectId",
```
tonyjhuang, line 45: same comment here, you can likely drop `with Id`

tonyjhuang (approved):

PR #7133, opened 2024-05-08
A bunch of FDC polish

joehan:
### Description
- Updated to new, better templates
- Updated location handling in init
- Added free trial terms
- Switched deletion prompt to a warning telling users go to console to delete.

### Scenarios Tested
Init with existing instance:
Screenshot 2024-05-08 at 12 01 55 PM

Init with new instance:
Screenshot 2024-05-08 at 12 02 38 PM


src/init/features/dataconnect/index.ts
```diff
@@ -25,6 +25,12 @@ export async function doSetup(setup: Setup, config: Config): Promise<void> {
...
     default: "dataconnect",
   });
+  // TODO: Guided prompts to set up connector auth mode and generate
+  const connectorId = await promptOnce({
+    message: "What ID would you like to use for your connector?",
+    type: "input",
+    default: "my-connector",
+  });
   // Hardcoded locations for when there is no project set up.
   let locationOptions = [
```
fredzqm, line 35: We can move `locationOptions` block right before the prompt?

Don't need to list FDC locations when developer already picked the csql instance.

src/init/features/dataconnect/index.ts
```diff
@@ -42,38 +48,24 @@ export async function doSetup(setup: Setup, config: Config): Promise<void> {
...
     const instances = await cloudsql.listInstances(setup.projectId);
-    const instancesInLocation = instances.filter((i) => i.region === locationId);
-    const choices = instancesInLocation.map((i) => {
-      return { name: i.name, value: i.name };
+    const choices = instances.map((i) => {
+      return { name: i.name, value: i.name, location: i.region };
     });
-    choices.push({ name: "Create a new instance", value: "" });
-    if (instancesInLocation.length) {
+    choices.push({ name: "Create a new instance", value: "", location: "" });
```
fredzqm, line 60: Add an option with link to create the instance in cloud console instead.

nit: we can add a "Refresh the cloud sql instance list option" here for them to resume after clicking create button in console.

nit: add a TODO to only add "Create a free trial instance" if there isn't one already.

We can sort of predict whether they will run into free trial instance there.

fredzqm (approved):
Thank you!

src/init/features/dataconnect/index.ts
fredzqm: Oh, we already have `checkForFreeTrialInstance`.

Make it a blocking comment then. Let's move its logics before prompt for "Create a free trial instance"

src/init/features/dataconnect/index.ts
joehan: Done.

fredzqm (approved):

PR #7140, opened 2024-05-09
Generate .graphqlrc on startup

hlshen:

### DIFF
--- a/firebase-vscode/src/data-connect/language-client.ts
+++ b/firebase-vscode/src/data-connect/language-client.ts
@@ -91,7 +91,7 @@ export function setupLanguageClient(
 
   const generateYamlFile = async () => {
     const basePath = vscode.workspace.rootPath;
-    const filePath = ".firebase/.dataconnect-graphqlrc.yaml"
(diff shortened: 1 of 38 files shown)


firebase-vscode/src/data-connect/language-client.ts
```diff
@@ -75,38 +78,70 @@ export function setupLanguageClient(context) {
...
 
+  const generateYamlFile = async () => {
+    const basePath = vscode.workspace.rootPath;
+    const filePath = ".firebase/.graphqlrc";
+    const fileUri = vscode.Uri.file(`${basePath}/${filePath}`);
+    const folderPath = ".firebase";
+    const folderUri = vscode.Uri.file(`${basePath}/${folderPath}`);
+
+    // TODO: Expand to multiple services
+    const config = configs.values[0];
```
yuchenshi, line 100: Would you mind short-circuiting this when `configs.values.length === 0`? Otherwise it throws. Feel free to put in an empty config

yuchenshi (approved):

PR #7143, opened 2024-05-09
Fix init shape

joehan:
### Description
This was previously sticking the prompted value into the object we use to build firebase.json, which was causing all sorts of funky behavior.


hlshen (approved):

hlshen (approved):

dconeybe (commented):
I tested locally and it fixes my problem. Thanks!

PR #7144, opened 2024-05-09
Block safe migrations because they are exteremely unsafe

joehan:
### Description
Cutting this becuase it leads to too many broken states. Basically, whenever a change includes both a safe and destructive change to the same table (IE renaming a field), atlasgo bundles them into a single ALTER table command. If we skip this and set schemaValidation to false, its extremely easy to successfully deploy connectors that will fail with SQL errors every single time.

### DIFF
--- a/src/dataconnect/schemaMigration.ts
+++ b/src/dataconnect/schemaMigration.ts
@@ -42,8 +42,8 @@ export async function migrateSchema(args: {
   const { options, schema, allowNonInteractiveMigration, validateOnly } = args;
 
   const { serviceName, instanceId, instanceName, databaseId } = getIdentifiers(schema);
-  await ensureServiceIsConnectedToCloudSql(serviceName, instanceName, databaseId);
   try {
+    await ensureServiceIsConnectedToCloudSql(serviceName, instanceName, databaseId);
     await upsertSchema(schema, validateOnly);
     logger.debug(`Database schema was up to date for ${instanceId}:${databaseId}`);
   } catch (err: any) {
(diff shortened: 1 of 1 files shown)


src/dataconnect/schemaMigration.ts
```diff
@@ -42,8 +42,8 @@ export async function migrateSchema(args: {
   const { options, schema, allowNonInteractiveMigration, validateOnly } = args;
 
   const { serviceName, instanceId, instanceName, databaseId } = getIdentifiers(schema);
-  await ensureServiceIsConnectedToCloudSql(serviceName, instanceName, databaseId);
   try {
+    await ensureServiceIsConnectedToCloudSql(serviceName, instanceName, databaseId);
```
joehan, line 46: Moved this back inside the try catch so that it can detect and repair out of band SQL schema changes. The try-catch still rethrows errors that we don't have an easy way of handling, so this won't swallow other errors.

src/dataconnect/schemaMigration.ts
```diff
@@ -187,7 +192,6 @@ async function promptForSchemaMigration(
     const choices = err.destructive
       ? [
           { name: "Execute all changes (including destructive changes)", value: "all" },
-          { name: "Execute only safe changes", value: "safe" },
```
fredzqm, line 190: nit: change this below option from `safe` to `all` as it means the same thing as `safe` in this context.

Then `if (incompatibleSchemaError.destructive && choice === "safe")` would be no longer needed.

rosalyntan (approved):

src/dataconnect/schemaMigration.ts
fredzqm: I would suggest add some error handling within `ensureServiceIsConnectedToCloudSql` to ignore any failed precondition errors.

`ensureServiceIsConnectedToCloudSql` helps ensure SQL is connected, at least tried to make the connection. Even if it failed to commit, the next step should surface the correct schema migration error if any.

PR #7149, opened 2024-05-09
Using android/ios default ports

joehan:
### Description
Andorid and IOS default to 9510 - for now, lets hardcode grpc endpoint to that port. In the future, we should merge http and grpc to a single port.

### DIFF
--- a/src/emulator/constants.ts
+++ b/src/emulator/constants.ts
@@ -13,7 +13,7 @@ export const DEFAULT_PORTS: { [s in Emulators]: number } = {
   auth: 9099,
   storage: 9199,
   eventarc: 9299,
-  dataconnect: 9399,
+  dataconnect: 9509,
 };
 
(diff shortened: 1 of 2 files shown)


yuchenshi (approved):

dconeybe (approved):

src/emulator/constants.ts
```diff
@@ -13,7 +13,7 @@ export const DEFAULT_PORTS: { [s in Emulators]: number } = {
   auth: 9099,
   storage: 9199,
   eventarc: 9299,
-  dataconnect: 9399,
+  dataconnect: 9509,
```
yuchenshi, line 16: This changes the HTTP port, let's revert

yuchenshi: Done in https://github.com/firebase/firebase-tools/pull/7163

PR #7157, opened 2024-05-10
Display accurate message for INACCESSIBLE_SCHEMA error

fredzqm:
### Description

1. Interpret and handle `INACCESSIBLE_SCHEMA` from the backend. Print accurate error message. In practice, `INACCESSIBLE_SCHEMA` should occur rarely:
   - CSQL was just created or in the middle of a restart.
   - CSQL is down or overloaded.
   - Switch to a new CSQL database, so no chance to establish connectivity.
3. Make sure errors from LRO are handled properly. Tested with running `firebase deploy` after deleting the P4SA IAM user, so `INCOMPATIBLE_SCHEMA` error are returned in Apply after provisioning connectivity.

There was a subtle typo `s/orignal/original`, failed with:
Screenshot 2024-05-09 at 8 54 06 PM

### Scenarios Tested

`firebase sql:migrate:diff` when schema is accessible vs not.
Screenshot 2024-05-09 at 8 02 31 PM


src/dataconnect/errors.ts
```diff
@@ -5,28 +5,23 @@ const PRECONDITION_ERROR_TYPESTRING = "type.googleapis.com/google.rpc.Preconditi
...
-  const details: any[] = original.details;
-  const incompatibles = details.filter((d) =>
-    d["@type"]?.includes(INCOMPATIBLE_SCHEMA_ERROR_TYPESTRING),
-  );
   // Should never get multiple incompatible schema errors
-  return incompatibles[0];
+  const incompatible = incompatibles[0];
+  // Extract the violation type from the precondition error detail.
+  const preconditionErrs = errorDetails(err, PRECONDITION_ERROR_TYPESTRING);
+  incompatible.violationType = preconditionErrs[0].violations[0].type;
```
joehan, line 16: This seems fragile to future changes - are we sure that we'll always return only one precondition error detail (maybe)? Are we sure that it will always contain exactly 1 violation (probably not)? 

Would prefer something at least like like:
```
if (preconditonErrs.length) {
const preconditionErr = preconditionErrs[0]
incompatible.violationType = preconditonErr.violations.some(v => v.type === "INACCESSIBLE_SCHEMA") ? "INACCESSIBLE_SCHEMA" : "INCOMPATIBLE_SCHEMA";
}
```

src/dataconnect/schemaMigration.ts
```diff
@@ -315,11 +315,29 @@ async function ensureServiceIsConnectedToCloudSql(
...
+        const message =
+          "Cannot access your CloudSQL database to validate schema. " +
+          "The following SQL statements can setup a new database schema.\n" +
+          error.diffs.map(toString).join("\n");
+        logLabeledWarning("dataconnect", message);
+        logLabeledWarning("dataconnect", "Some SQL resources may already exist.");
+      }
+      break;
+    default:
+      throw new FirebaseError("Unknown schema violation type: " + error.violationType);
```
joehan, line 339: Include the error message here too - if this ever get hit, it will be very helpful to have it.

src/dataconnect/types.ts
```diff
@@ -48,12 +48,16 @@ export interface File {
...
 export interface IncompatibleSqlSchemaError {
-  // A list of differences between the two schema with instrucitons how to resolve them.
+  // A list of differences between the two schema with instructions how to resolve them.
   diffs: Diff[];
   // Whether any of the changes included are destructive.
   destructive: boolean;
+
+  // The failed precondition validation type.
+  // Either "INCOMPATIBLE_SCHEMA" or "INACCESSIBLE_SCHEMA".
+  violationType: string;
```
joehan, line 60:
```suggestion
  violationType: "INCOMPATIBLE_SCHEMA" | "INACCESSIBLE_SCHEMA" | string;
```

Good spot for a string union type, tho we likely should keep the 'string' in there since there's no guarantee other types wouldn't be returned in the future.

joehan (changes requested):

src/dataconnect/errors.ts
fredzqm: This is true right now. If backend returns `IncompatibleSqlSchemaError`, the first `PreconditionFailure` should be that.

Though you made a good point on future changes. Let me make it detect the first violation type that match.

src/dataconnect/schemaMigration.ts
fredzqm: Unfortunately, at this point, we don't have access to the original error message. I can include the whole diff error detail here.

fredzqm (commented):
Addressed feedbacks and tested again

PR #7159, opened 2024-05-10
Cloud SQL create and update polish

fredzqm:
### Description

1. Explain why update instance is needed
2. Make create database optional. It fails if CSQL is in the middle of some operations.
    Common example is linking an instance just created in cloud console.

3. Update the create setting a bit to match the default query insights configs. 
Otherwise, any instances created via CLI would override it to 0... I consider this a CSQL bug, but see also http://cl/629192474
 
When customer click edit instance, do nothing and then click save, they see this:
Screenshot 2024-05-09 at 8 58 40 PM

3. CSQL database creation is optional because the backend would try that as well. If there is an ongoing LRO, we CLI can skip this step.

### Scenarios Tested

Print update reason:
Screenshot 2024-05-10 at 10 39 32 AM

### DIFF
--- a/src/dataconnect/provisionCloudSql.ts
+++ b/src/dataconnect/provisionCloudSql.ts
@@ -27,13 +27,14 @@ export async function provisionCloudSql(args: {
     const existingInstance = await cloudSqlAdminClient.getInstance(projectId, instanceId);
     silent || utils.logLabeledBullet("dataconnect", `Found existing instance ${instanceId}.`);
     connectionName = existingInstance?.connectionName || "";
-    if (!checkInstanceConfig(existingInstance, enableGoogleMlIntegration)) {
+    const why = checkInstanceConfig(existingInstance, enableGoogleMlIntegration);
+    if (why) {
       // TODO: Return message from checkInstanceConfig to explain exactly what changes are made
       silent ||
         utils.logLabeledBullet(
           "dataconnect",
           `Instance ${instanceId} settings not compatible with Firebase Data Connect. ` +
-            `Updating instance to enable Cloud IAM authentication and public IP. This may take a few minutes...`,
+            `Updating instance to ${why}. This may take a few minutes...`,
         );
       await promiseWithSpinner(
         () =>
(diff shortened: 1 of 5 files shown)


src/dataconnect/provisionCloudSql.ts
```diff
@@ -27,13 +27,14 @@ export async function provisionCloudSql(args: {
     const existingInstance = await cloudSqlAdminClient.getInstance(projectId, instanceId);
     silent || utils.logLabeledBullet("dataconnect", `Found existing instance ${instanceId}.`);
     connectionName = existingInstance?.connectionName || "";
-    if (!checkInstanceConfig(existingInstance, enableGoogleMlIntegration)) {
+    const why = checkInstanceConfig(existingInstance, enableGoogleMlIntegration);
+    if (why) {
       // TODO: Return message from checkInstanceConfig to explain exactly what changes are made
```
rosalyntan, line 32: This TODO can be removed now, right?

src/dataconnect/provisionCloudSql.ts
```diff
@@ -95,23 +106,23 @@ export async function provisionCloudSql(args: {
 export function checkInstanceConfig(
   instance: Instance,
   requireGoogleMlIntegration: boolean,
-): boolean {
+): string | false {
   const settings = instance.settings;
   // CloudSQL instances must have public IP enabled to be used with Firebase Data Connect.
   if (!settings.ipConfiguration?.ipv4Enabled) {
-    return false;
+    return "enable public IP";
```
rosalyntan, line 113: Should this add to an array of reasons, so this method returns all the things that need to be updated, rather than just the first one?

src/dataconnect/provisionCloudSql.ts
```diff
@@ -95,23 +106,23 @@ export async function provisionCloudSql(args: {
 export function checkInstanceConfig(
```
rosalyntan, line 106: nit: Maybe rename to `getUpdateReason` and have this return an array of string reasons?

rosalyntan (approved):

src/dataconnect/provisionCloudSql.ts
fredzqm: Done

src/dataconnect/provisionCloudSql.ts
fredzqm: Done

src/dataconnect/provisionCloudSql.ts
fredzqm: Good idea. Changed it to append to the reason string, so will display it as a list.

fredzqm:
Thank  you for the review~ Addressed feedbacks~

PR #7162, opened 2024-05-10
Readd platform specific dependencies to npm-shrinkwrap.json

joehan:
### Description
A while back (shortly after 13.7.5), we were affected by https://github.com/npm/cli/issues/4828, which caused some platform specific, transitive, optional dependencies to be removed from npm-shrinkwrap.json. Unbeknownst to me, IDX was using NIX to build `firebase-tools` and set it up as a builtin. Since `linux-64x` version of `@esbuild` was not in npm-shrinkwrap.json, this started to fail, and so IDX has been stuck on 13.7.5

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -980,6 +980,54 @@
         "node": ">=16"
       }
     },
+    "node_modules/@esbuild/android-arm": {
+      "version": "0.17.16",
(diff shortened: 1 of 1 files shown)


bkendall (approved):
Looks good. so long as that check doesn't remove things from shrinkwrap and it passes, I don't see a problem here

PR #7164, opened 2024-05-10
Change default database name to `postgres`.

yuchenshi:
This DB name is always present in most Postgres setups.

### DIFF
--- a/src/init/features/dataconnect/index.ts
+++ b/src/init/features/dataconnect/index.ts
@@ -113,7 +113,7 @@ export async function doSetup(setup: Setup, config: Config): Promise<void> {
     cloudSqlDatabase = await promptOnce({
       message: `What ID would you like to use for your new database in ${cloudSqlInstanceId}?`,
       type: "input",
-      default: `dataconnect`,
+      default: `postgres`,
     });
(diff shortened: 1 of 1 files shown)


yuchenshi:
Converting to draft pending internal discussion

src/init/features/dataconnect/index.ts
```diff
@@ -113,7 +113,7 @@ export async function doSetup(setup: Setup, config: Config): Promise<void> {
     cloudSqlDatabase = await promptOnce({
       message: `What ID would you like to use for your new database in ${cloudSqlInstanceId}?`,
       type: "input",
-      default: `dataconnect`,
+      default: `postgres`,
```
fredzqm, line 116: Is the goal to improve database in emulator?

This changes the prod Cloud SQL database name, which I think a separate name it more clear. I agree that we should make the local emulator url as default as possible.

https://stackoverflow.com/questions/3582552/what-is-the-format-for-the-postgresql-connection-string-url

yuchenshi: Let's continue the internal discussion on b/339903547 please

PR #7166, opened 2024-05-11
Fix login in idx

hlshen:


firebase-vscode/webviews/components/ProjectSection.tsx
```diff
@@ -7,14 +7,26 @@ import React from "react";
...
+  isMonospace,
 }: {
-  userEmail: string | null;
+  user: UserWithType | ServiceAccountUser | null;
   projectId: string | null | undefined;
+  isMonospace: boolean;
 }) {
+  const userEmail = user.email;
+
+  if (isMonospace && user?.type === "service_account") {
```
joehan, line 27: Nonblocking: Could we talk thru this case on monday? Is there a normal path where this would come up? If so, i'm a bit concerned that it will end up using the random service account gerenated for IDX, which is probably not functional or intended

joehan (approved):

PR #7208, opened 2024-05-22
`sql:migrate --force` shouldn't prompt

fredzqm:
### Description

`firebase dataconnect:sql:migrate --force` should never prompt.

`firebase dataconnect:sql:migrate` should prompt for schema migrations and then abort if there is invalid connectors. This way developers can see the SQL before taking the suggesting and run it `--force`.

#### Without invalid connectors
`firebase dataconnect:sql:migrate`
Screenshot 2024-05-22 at 12 20 37 PM

`firebase dataconnect:sql:migrate --force`
Screenshot 2024-05-22 at 12 21 27 PM

#### With invalid connectors
`firebase dataconnect:sql:migrate`
Screenshot 2024-05-22 at 12 23 41 PM

`firebase dataconnect:sql:migrate --force`
Screenshot 2024-05-22 at 12 24 44 PM

### DIFF
--- a/src/dataconnect/schemaMigration.ts
+++ b/src/dataconnect/schemaMigration.ts
@@ -50,10 +50,10 @@ export async function diffSchema(schema: Schema): Promise<Diff[]> {
 export async function migrateSchema(args: {
   options: Options;
   schema: Schema;
-  allowNonInteractiveMigration: boolean;
+  /** true for `dataconnect:sql:migrate`, false for `deploy` */
   validateOnly: boolean;
 }): Promise<Diff[]> {
-  const { options, schema, allowNonInteractiveMigration, validateOnly } = args;
+  const { options, schema, validateOnly } = args;
 
   const { serviceName, instanceId, instanceName, databaseId } = getIdentifiers(schema);
   await ensureServiceIsConnectedToCloudSql(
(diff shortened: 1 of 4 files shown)


src/dataconnect/schemaMigration.ts
```diff
@@ -205,44 +199,51 @@ async function handleIncompatibleSchemaError(args: {
 async function promptForSchemaMigration(
   options: Options,
   databaseName: string,
-  err: IncompatibleSqlSchemaError,
-  allowNonInteractiveMigration: boolean,
-): Promise<"none" | "safe" | "all"> {
+  err: IncompatibleSqlSchemaError | undefined,
+  validateOnly: boolean,
+): Promise<"none" | "all"> {
```
rosalyntan, line 204: Since there's only two options now, should this return `Promise<false | true>`? Also does `maybeExecuteSchemaMigration` better describe what this method does?

fredzqm: This method doesn't execute schema migration. It determines whether to execute it later.

Well, keeping it as string merely because we will have a "compatible" mode soon. At least 3 value enum is needed here.

rosalyntan (approved):

fredzqm:
Thanks!

PR #7230, opened 2024-05-28
Make prettify more tolerant to missing fields.

yuchenshi:
### Description

Fixes errors like this:

```
[debug] [2024-05-28T23:02:54.833Z] TypeError: Cannot read properties of undefined (reading 'file')
    at prettify (\some\path\node_modules\firebase-tools\lib\dataconnect\graphqlError.js:7:39)
```

### Scenarios Tested

A customer has encountered this when the error `connector source must not contain anonymous operations` was thown. The existing code fails during processing, and swallows the actual error. I guess these are indeed without file names or paths for now. We may need to improve the error reporting, but for now, the point is that errors like these do exist.

This PR make sure when it happens again we will see the actual error message.

### Sample Commands

`firebase deploy`


joehan (approved):
LGTM

PR #7232, opened 2024-05-29
Add framework for GA4 Analytics to VSCode

hlshen:

### DIFF
--- a/firebase-vscode/src/analytics.ts
+++ b/firebase-vscode/src/analytics.ts
@@ -1,0 +1,36 @@
+import {
+  env,
+  TelemetryLogger,
+  TelemetrySender,
+} from "vscode";
+import { pluginLogger } from "./logger-wrapper";
+import { AnalyticsParams, trackVSCode } from "./track";
+
+
+export class AnalyticsLogger {
+    readonly logger: TelemetryLogger
+    constructor() {
+        this.logger = env.createTelemetryLogger(new GA4TelemetrySender(pluginLogger));
+    }
+}
+
+class GA4TelemetrySender implements TelemetrySender {
+  constructor(readonly pluginLogger) {}
+
+  sendEventData(
+    eventName: string,
+    data?: Record<string, any> | undefined,
+  ): void {
+    if (!env.isTelemetryEnabled) {
+      pluginLogger.warn("Telemetry is not enabled.")
+    }
+    if (!data) {
+      return;
+    }
+    trackVSCode(eventName, data as AnalyticsParams);
+  }
+
+  sendErrorData(error: Error, data?: Record<string, any> | undefined): void {
+    // n/a
+  }
+}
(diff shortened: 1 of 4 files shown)


firebase-vscode/src/analytics.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (!env.isTelemetryEnabled) {
+      pluginLogger.warn("Telemetry is not enabled.")
+    }
+    if (!data) {
+      return;
+    }
+    trackVSCode(eventName, data as AnalyticsParams);
+  }
+
+  sendErrorData(error: Error, data?: Record<string, any> | undefined): void {
```
joehan, line 33: I like the idea here, but be careful what we include in error events when we implement this later. 

Events should not include identifying info like file paths or service ids. If youre relying on error.message to name name the event, its very easy to accidentally include this.

firebase-vscode/src/analytics.ts
```diff
@@ -0,0 +1,36 @@
...
+  env,
+  TelemetryLogger,
+  TelemetrySender,
+} from "vscode";
+import { pluginLogger } from "./logger-wrapper";
+import { AnalyticsParams, trackVSCode } from "./track";
+
+
+export class AnalyticsLogger {
+    readonly logger: TelemetryLogger
```
joehan, line 11: Add semicolons where they are missing.

firebase-vscode/src/analytics.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+class GA4TelemetrySender implements TelemetrySender {
+  constructor(readonly pluginLogger) {}
+
+  sendEventData(
+    eventName: string,
+    data?: Record<string, any> | undefined,
+  ): void {
+    if (!env.isTelemetryEnabled) {
```
joehan, line 24: Quick sanity check - Is this a VSCode wide setting? If so, users need to explicitly opt in for isTelemetryEnabled to be true, right?

firebase-vscode/src/analytics.ts
```diff
@@ -0,0 +1,36 @@
...
+      pluginLogger.warn("Telemetry is not enabled.")
```
joehan, line 25: We should return here, otherwise we just log that telemetry is disabled and then send telemetry 😨

joehan (changes requested):

firebase-vscode/src/analytics.ts
hlshen: I believe the telemetrylogger from vscode handles this. Let me confirm

firebase-vscode/src/analytics.ts
hlshen: correct. I need to look into how to prompt users to enable. But this is a must check for metrics within vscode

firebase-vscode/src/analytics.ts
hlshen: never mind.  we should definitely return

joehan (approved):

PR #7233, opened 2024-05-30
Automatically start the emulator when the extension starts

rrousselGit:
### Description

This starts the data-connect emulator when the extension starts.

Alongside it, the emulator states is now visible in the statusbar. And clicking on it opens the RC file.

This also removes the previous "start/stop emulator" buttons and commands.


firebase-vscode/package.json
```diff
@@ -42,14 +42,6 @@
       {
         "command": "firebase.selectProject",
         "title": "Firebase: Switch project"
-      },
```
rrousselGit, line 45: I assumed there's no reason to keep those commands if the emulator is always started :)

firebase-vscode/src/core/emulators.ts
```diff
@@ -156,115 +151,98 @@ export class EmulatorsController implements Disposable {
...
-      {
-        location: vscode.ProgressLocation.Window,
-        cancellable: false,
-        title: "Stopping emulators",
-      },
-      async (progress) => {
-        progress.report({ increment: 0 });
-
+  async startEmulators() {
+    this.emulatorStatusItem.text = "$(loading~spin) Emulators: Starting...";
```
hlshen, line 167: "Dataconnect Toolkit: Starting..."

firebase-vscode/src/core/emulators.ts
```diff
@@ -156,115 +151,98 @@ export class EmulatorsController implements Disposable {
...
-          status: "stopping",
-          infos: this.emulators.value.infos,
+          status: "running",
+          infos: {
+            uiUrl: getEmulatorUiUrl(),
+            displayInfo: listRunningEmulators(),
+          },
         };
-        await stopEmulators();
+        this.emulatorStatusItem.text = "Emulators: running";
```
hlshen, line 185: (Dataconnect Logo) Toolkit: running

firebase-vscode/src/core/index.ts
```diff
@@ -53,9 +54,22 @@ export async function registerCore({
   });
 
   const emulatorsController = new EmulatorsController(broker);
+  // Start the emulators when the extension starts.
+  emulatorsController.startEmulators();
+
+  const openRcCmd = vscode.commands.registerCommand(
+    "firebase.openFirebaseRc",
+    () => {
+      for (const root of getRootFolders()) {
```
hlshen, line 63: I think you can use `getConfigPath()` from the same import file instead of looping over `getRootFolders()`

firebase-vscode/src/test/integration/fishfood/emulator_status.ts
```diff
@@ -9,8 +9,8 @@ firebaseTest(
     const workbench = await browser.getWorkbench();
     const statusBar = new StatusBar(workbench);
 
-    expect(await statusBar.emulatorsStatus.getText()).toBe(
-      "No emulator running"
+    expect(await statusBar.emulatorsStatus.getText()).toContain(
+      "Emulators: starting"
```
hlshen, line 13: tests to show Logo + "Toolkit"

hlshen (changes requested):
I take it the PSQL connection will be in a followup?

firebase-vscode/src/core/emulators.ts
hlshen: Ignore the above.

We only want to show "(FDC Logo) Emulators: running" when the user has explicitly connected to psql

firebase-vscode/src/core/index.ts
rrousselGit: I figured there's the case of multiroot workspaces.

rrousselGit:
Let me update this with the postgres logic

src/rc.ts
```diff
@@ -240,9 +240,7 @@ export class RC {
   }
 
   setDataconnect(localConnectionString: string) {
-    if (!this.data.dataconnectEmulatorConfig) {
```
hlshen, line 243: @joehan confirming this doesn't break anything in cli

joehan: Seems like this was previously broken, LGTM

PR #7235, opened 2024-05-30
Adding VSCE to our publish script, and removing remnants of twitter

joehan:
### Description
Adding VSCE to our release script and removing remnants of broken twitter integration. Now, on every CLI release, we will cut a patch release of VSCode. If you want to make a major/minor version update, you can manually increase package.json.

### Scenarios Tested
Ran this script against a different bucket, and successfully built + uploaded the 2 VSIX files


firebase-vscode/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# Change Log
```
joehan, line 1: I made this change to automate our changelog - we can put change log entries for the next release under NEXT, and the script will automatically replace it with the next version when its run.

scripts/publish.sh
```diff
@@ -87,6 +81,10 @@ npm version $VERSION
 NEW_VERSION=$(jq -r ".version" package.json)
 echo "Made a $VERSION version."
 
+echo "Preparing VSCode extension..."
+./scripts/publish-vscode.sh patch
```
joehan, line 85: Hardcoding this to patch right now - open to switching this to match the CLI version type (ie if we release a minor version of the CLI, the accompanying VSCE release will be minor) - lmk what you prefer.

firebase-vscode/package.json
```diff
@@ -191,12 +191,11 @@
   "scripts": {
```
hlshen, line 191: did my changes for this file not merge?

scripts/publish.sh
hlshen: I don't think it should match. VSCode features upgrades will definitely differ.

What does the CLI do? Can we specify the update "level" through changelog?

hlshen (approved):

yuchenshi (approved):

hlshen (approved):

PR #7241, opened 2024-05-31
Make vector warnings best effort

joehan:
### Description
Catch any errors from 'fdc build' in this context so that we always try to start the emulator.


src/emulator/dataconnectEmulator.ts
```diff
@@ -35,21 +35,25 @@ export class DataConnectEmulator implements EmulatorInstance {
...
+          );
+        } else {
+          this.logger.logLabeled(
+            "WARN",
+            "Data Connect",
+            "Operations that use vector_embed will make calls to production Vertex AI",
+          );
+        }
       }
+    } catch (err: any) {
```
fredzqm, line 55: I don't think this try-catch is needed. 

If `fdc build` returns non zero exit code, it failed to read sources [here](https://source.corp.google.com/piper///depot/google3/third_party/firebase/dataconnect/emulator/cli/build/build.go;l=31-34;rcl=639106823).

Compile errors are still zero exit code with gql error part of output json.

Do we have a repro of this bug?

yuchenshi: If you look into the implementation of `.build` below, any output on stderr will be considered fatal, including verbose logs. This may be overkill. Can we agree on the desirable behavior? I think most tools have the main output (e.g. JSON) on stdout and logging should be done freely on stderr

fredzqm: I agree.

We should throw err based on exit code. Stderr goes in the error message in case of failure, but otherwise debug logs

joehan: Fair points - changed build to only throw on nonzero error codes. I do still think this should be a best effort check in this context, so I think it makes sense to keep this try/catch.

src/emulator/dataconnectEmulator.ts
```diff
@@ -35,24 +35,27 @@ export class DataConnectEmulator implements EmulatorInstance {
...
+      this.logger.log("DEBUG", `'fdc build' failed with error: ${err.message}`);
     }
     return start(Emulators.DATACONNECT, {
-      ...this.args,
```
joehan, line 55: While testing, I noticed this was leading to a bunch of 'Ignored unrecognized arg' debug logs, so I cleaned it up.

src/emulator/downloadableEmulators.ts
```diff
@@ -295,6 +295,10 @@ const Commands: { [s in DownloadableEmulators]: DownloadableEmulatorCommand } =
       "local_connection_string",
       "project_id",
       "service_location",
+      "disable_sdk_generation",
```
joehan, line 298: Somewhat unrelated, but updated this list of flags to match the binary.

src/emulator/dataconnectEmulator.ts
```diff
@@ -111,11 +114,16 @@ export class DataConnectEmulator implements EmulatorInstance {
         original: res.error,
       });
     }
-    if (res.stderr) {
+    if (res.status !== 0) {
       throw new FirebaseError(
         `Unable to build your Data Connect schema and connectors: ${res.stderr}`,
```
yuchenshi, line 119:
```suggestion
        `Unable to build your Data Connect schema and connectors (exit code ${res.status}): ${res.stderr}`,
```

src/emulator/dataconnectEmulator.ts
yuchenshi: I think we should keep `auto_download`. It's probably not needed in this context because `.build` happens to download the binary, but this probably won't survive a refactor or two.

yuchenshi (approved):

fredzqm (approved):

PR #7240, opened 2024-05-31
Improved handling of 'fresh from console' services

joehan:
### Description
During `firebase init dataconnect`, if there are any services with either no schema or no GQL files n the schema, offer to pull the config from that service.

Also, stop prompting users for connector ID.

### Scenarios Tested
I tested:
 service with no schema:
Screenshot 2024-05-31 at 10 37 56 AM

service with empty schema:
Screenshot 2024-05-31 at 10 13 47 AM

New service (existing flow):
Screenshot 2024-05-31 at 10 38 27 AM

PR #7259, opened 2024-06-03
Address comment from #7240

joehan:
### Description
One last comment from #7240 - we weren't actually prompting in this case, but I moved the if statement to make the code a bit more readable.


src/init/features/dataconnect/index.ts
```diff
@@ -175,17 +175,17 @@ async function promptForService(setup: Setup, info: RequiredInfo): Promise<Requi
 async function promptForCloudSQLInstance(setup: Setup, info: RequiredInfo): Promise<RequiredInfo> {
   if (setup.projectId) {
     const instances = await cloudsql.listInstances(setup.projectId);
-    let choices = instances.map((i) => {
-      return { name: i.name, value: i.name, location: i.region };
-    });
+      if (instances.length) {
```
fredzqm, line 178: I still see a bug here. It would still prompt "Which CloudSQL instance would you like to use?" when there is no instance available in the target region.

We want to do `choices.length` after it's filtered by region.

```
let choices = instances.map((i) => {
    return { name: i.name, value: i.name, location: i.region };
});
if (choices.length) {
      const freeTrialInstanceId = await checkForFreeTrialInstance(setup.projectId);
      if (!freeTrialInstanceId) {
        choices.push({ name: "Create a new instance", value: "", location: "" });
      }
      info.cloudSqlInstanceId = await promptOnce({
        type: "list",
        choices,
      });
```

PR #7309, opened 2024-06-11
Add new framework for dataconect e2e test

joehan:
### Description
Setting up a framework to test more complicated schema migration and deployment cases. I've started with just a few simple cases, but we can add more in the future.

The main changes here:
- Every case is run on a separate, fresh DB and service, so no false no-op positives
- Cases are now 'sequences' of schemas/connectors to deploy, to let use test migration from different starting states


scripts/dataconnect-test/tests.ts
```diff
@@ -28,45 +63,128 @@ async function list() {
...
+  return await cli.exec("deploy", FIREBASE_PROJECT, args, __dirname, /** quiet=*/ false, {
+    FIREBASE_CLI_EXPERIMENTS: "dataconnect",
+  });
+}
+
+function toPath(p: string) {
+  return path.join(__dirname, p);
+}
+
+function getRandomString(length: number): string {
```
fredzqm, line 98: surprised that this doesn't have a native util

scripts/dataconnect-test/tests.ts
```diff
@@ -28,45 +63,128 @@ async function list() {
...
+  let result = "";
+  for (let i = 0; i < length; i++) {
+    result += SUFFIX_CHAR_SET.charAt(Math.floor(Math.random() * SUFFIX_CHAR_SET.length));
+  }
+  return result;
+}
+
+// Each test run should use a random serviceId and databaseId.
+function newTestRun(): { serviceId: string; databaseId: string } {
+  const serviceId = `cli-e2e-service-${getRandomString(6)}`;
```
fredzqm, line 109: Consider using the same random string to correspond in logs?

scripts/dataconnect-test/tests.ts
```diff
@@ -28,45 +63,128 @@ async function list() {
...
     expect(FIREBASE_PROJECT).not.to.equal("", "No FBTOOLS_TARGET_PROJECT env var set.");
+    const info = newTestRun();
+    serviceId = info.serviceId;
+    databaseId = info.databaseId;
+    await requireAuth({});
+  });
+
+  afterEach(async function (this) {
+    this.timeout(10000);
+    fs.rmSync(toPath("fdc-test"), { recursive: true, force: true });
```
fredzqm, line 157: Extract a constant or variable for `toPath("fdc-test")`?

See it in many places

fredzqm (approved):

scripts/dataconnect-test/tests.ts
joehan: we could use crypto.getRandomByes and cast it to a string, but crypto is not available in all node builds (https://nodejs.org/api/crypto.html#determining-if-crypto-support-is-unavailable) and I generally try to avoid using it

joehan:
😢  Chasing down access to the target project so that we can enable the data connect API and run tests on it

PR #7315, opened 2024-06-12
Auth working in IDX?

joehan:

### DIFF
--- a/src/requireAuth.ts
+++ b/src/requireAuth.ts
@@ -37,9 +37,7 @@ function getAuthClient(config: GoogleAuthOptions): GoogleAuth {
  * @param authScopes scopes to be obtained.
  */
 async function autoAuth(options: Options, authScopes: string[]): Promise<void | string> {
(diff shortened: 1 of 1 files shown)


src/requireAuth.ts
```diff
@@ -52,7 +50,10 @@ async function autoAuth(options: Options, authScopes: string[]): Promise<void |
     // Make sure any error here doesn't block the CLI, but log it.
     logger.debug(`Error getting account credentials.`);
   }
-
+  if (process.env.MONOSPACE_ENV && token && clientEmail) {
```
joehan, line 53: Note: we should maybe replace this MONOSPACE check with a check whether clientEmail is a user email (as opposed to a service account email). IIUC, this would make ADCs set through GCloud work more consistently with some of our login flows. Likely do need to iron out some details around logout with it tho

hlshen: ++ we'll need a huge rework on logout anyways i think...

hlshen (approved):

PR #7316, opened 2024-06-12
refactor webviews

hlshen:


firebase-vscode/webpack.common.js
```diff
@@ -219,11 +219,11 @@ const extensionConfig = {
   },
 };
 
-function makeWebConfig(entryName) {
+function makeWebConfig(entryName, entryPath = "") {
   return {
     name: entryName,
     mode: "none", // this leaves the source code as close as possible to the original (when packaging we set this to 'production')
-    entry: `./webviews/${entryName}.entry.tsx`,
+    entry: `./webviews/${entryPath}${entryName}.entry.tsx`,
```
joehan, line 226: Right now, this is fragile to mismatched slashes (ie 'myEntryPath/' works, but 'myEntryPath' does not). You probably want to use path.join to construct this..

firebase-vscode/webpack.common.js
```diff
@@ -309,4 +309,9 @@ module.exports = [
     .filter((filename) => filename.match(/\.entry\.tsx/))
     .map((filename) => filename.replace(/\.entry\.tsx/, ""))
     .map((name) => makeWebConfig(name)),
+  ...fs
```
joehan, line 312: Nit - this is a lot of logic to inline. For clarity, consider making these 2 long chained calls into variables ie:
```
const webConfigs =   ...fs
    .readdirSync("webviews")
    .filter((filename) => filename.match(/\.entry\.tsx/))
    .map((filename) => filename.replace(/\.entry\.tsx/, ""))
    .map((name) => makeWebConfig(name));
const dataconnectWebConfigs = <blah>

module.exports = [ extensionConfig, ...webConfigs, ...dataconnectWebConfigs ];

joehan (approved):
LGTM after at least the path fix.

firebase-vscode/webpack.common.js
hlshen: done

firebase-vscode/webpack.common.js
hlshen: done

PR #7318, opened 2024-06-13
Move tests to be colocated with source files.

yuchenshi:
### Description

e.g. A test file `src/tests/foo/bar.spec.js` is moved to `src/foo/bar.spec.js`.

This helps restructuring directories later.

Each commit can be reviewed as a smaller unit.

### Scenarios Tested

See GitHub unit and integration test results.

### Sample Commands

npm test


joehan:
LGTM since all tests are passing. I know this is a hackweek draft rn, but I'm in favor of the change.

yuchenshi:
Commits all cleaned up and reworded. I've enabled auto-rebase-and-merge too. This now just needs an approval.

joehan (approved):

PR #7322, opened 2024-06-13
Disable auth check for downloading Dataconnect emulator. Remove emulator UI references

hlshen:
Remove emulator UI references in vscode. Remove need for Auth to download Dataconnect emulator.


firebase-vscode/webviews/components/EmulatorPanel.tsx
```diff
@@ -1,140 +1,142 @@
-import { VSCodeButton, VSCodeLink } from "@vscode/webview-ui-toolkit/react";
-import React, { useEffect } from "react";
-import { Spacer } from "./ui/Spacer";
-import { broker, useBroker, useBrokerListener } from "../globals/html-broker";
-import { PanelSection } from "./ui/PanelSection";
-import { FirebaseConfig } from "../../../src/firebaseConfig";
-import { EmulatorInfo } from "../../../src/emulator/types";
-import { webLogger } from "../globals/web-logger";
-import { DEFAULT_EMULATOR_UI_SELECTIONS } from "../../common/messaging/protocol";
+// DISABLED UNTIL MULTIPLE EMULATORS IS SUPPORTED
```
hlshen, line 1: Actually not sure if I should just delete the emulator UI piece. I want to keep this code around but this level of commenting out is getting out of hand ... Open to thoughts

joehan: Just delete it. We have source control, no need to comment out code to save it for later. We can always dig it back up later.

joehan (approved):

Issue #7207, opened 2024-05-22
Hosting doesn't take dotenv files from different environments


yanqianglu:
### [REQUIRED] Environment info

**firebase-tools:** 13.10.0

**Platform:** macOS

### [REQUIRED] Test case

See "Steps to reproduce" below.

### [REQUIRED] Steps to reproduce

```
# .firebaserc
{
  "projects": {
    "prod": "my-project-prod",
    "stg": "my-project-stg"
  }
}
```

```
# .envfile
FOO_1=BAR
```

```
# .env.stg file
FOO_2=BAR
```

### [REQUIRED] Expected behavior

It should take the `.env.stg` file as part of the build.

```
> firebase use stg && firebase deploy

  ▲ Next.js 14.2.3

  - Environments: .env, .env.stg


   Creating an optimized production build ...

 ✓ Compiled successfully

   Linting and checking validity of types ...

...


i  functions: Loaded environment variables from .env, .env.stg.
```

### [REQUIRED] Actual behavior

It only takes `.env` file and ignored `.env.stg` file. The `.env.stg` is included as part of the functions deployment though.

```
> firebase use stg && firebase deploy

  ▲ Next.js 14.2.3

  - Environments: .env


   Creating an optimized production build ...

 ✓ Compiled successfully

   Linting and checking validity of types ...


...



i  functions: Loaded environment variables from .env, .env.stg.
```

chalosalvador:
@yanqianglu If your pages use SSR you can rename your dotenv file to `.env.<PROJECT-ID>` instead of `.env.<PROJECT-ALIAS>` and you should be able to read  the environment variables in that file. You can read more about it in the [Parameterized configuration](https://firebase.google.com/docs/functions/config-env?gen=2nd#node.js) for Cloud Functions.

I have opened a PR (#7323) to handle the issue when using SSG in Next.js. But the idea is the same, we would read the values from the `.env.<PROJECT-ID>`.

chalosalvador:
@yanqianglu  the PR #7323 with the fix has been merged and will be available in the next CLI release.

PR #7323, opened 2024-06-14
Allow dotenv files from different environments for SSR and SSG in Next.js

chalosalvador:
### Description

This PR is intended to fix the issue #7207 by checking if the .env.<PROJECT-ID> file exists and make it available for the build process.

### Scenarios Tested

1. Initialized a hosting project with Next.js
2. Created a `.env.<PROJECT-ID>` file in the root of the Next.js project
3. Created a page that uses SSR and accesses the environment variable in `.env.<PROJECT-ID>`
4. Created a page that uses SSG and accesses the environment variable in `.env.<PROJECT-ID>`
5. Deployed the project to Firebase hosting. 
6. Make sure the environment variable in the `.env.<PROJECT-ID>` is available in a server side rendered page.
7. Make sure the environment variable in the `.env.<PROJECT-ID>` is available in a static generated page.


src/frameworks/next/index.ts
```diff
@@ -126,6 +126,22 @@ export async function build(
...
   const env = { ...process.env };
 
+  // Check if the .env.<PROJECT-ID> file exists and make it available for the build process
+  if (context?.projectId) {
+    const envFile = join(dir, `.env.${context.projectId}`);
+
+    if (await pathExists(envFile)) {
+      const envConfig = await readFile(envFile);
+      const envConfigString = envConfig.toString();
+      const envConfigLines = envConfigString.split("\n");
```
jamesdaniels, line 136: I worry this is a bit naive of a parser for .env, we need to make sure that we handle escaping and multiline files... maybe we should see if there's some reputable NPM libs?

jamesdaniels (changes requested):
Can we add some test coverage?

src/frameworks/next/index.ts
chalosalvador: It turns out there's actually a function available ([parseStrict](https://github.com/firebase/firebase-tools/blob/chalosalvador/dotenv-files/src/functions/env.ts#L192)) that parses .env files in a more accurate way.

src/frameworks/next/index.ts
```diff
@@ -126,6 +127,18 @@ export async function build(
...
 
+  // Check if the .env.<PROJECT-ID> file exists and make it available for the build process
+  if (context?.projectId) {
+    const projectEnvPath = join(dir, `.env.${context.projectId}`);
+
+    if (await pathExists(projectEnvPath)) {
+      const projectEnvVars = parseStrict((await readFile(projectEnvPath)).toString());
+
+      // Merge the parsed variables with the existing environment variables
+      Object.assign(env, projectEnvVars);
```
jamesdaniels, line 138: dotenv does not override environment variables by default if i recall, so this logic will likely need inverting `{ ...projectEnvVars, ...process.env }`

jamesdaniels (changes requested):
See small comment on overrides, add tests, and changelog—otherwise LGTM

src/frameworks/next/index.ts
chalosalvador: @jamesdaniels I just made this change, added tests for the `parseStrict` function and added the changelog.

jamesdaniels (approved):

PR #7332, opened 2024-06-15
Add a manifest file next to each test fixture.

yuchenshi:
### Description

This allows moving test files and / or fixture directories independently using TypeScript refactors to update imports.

No more `../../` as untyped strings. `tsc` will error if there's a mismatch.

This method won't work for templates -- I'll find a different way for those in a follow-up PR.

### Scenarios Tested

See tests on CI.

### Sample Commands

`npm run mocha`.


joehan (approved):
Nice idea, lgtm!

PR #7338, opened 2024-06-17
Update dataconnect init template

fredzqm:
The previous example were unfortunately encouraging insecure practices. We should suggest `auth.uid` in our init template.


templates/init/dataconnect/mutations.gql
```diff
@@ -1,30 +1,35 @@
...
+# mutation CreateRecipient($emailId: UUID) @auth(level: USER) {
 #   recipient_insert(data: {
 #     emailId: $emailId,
-#     userUid: $uid
+#     userUid_expr: "auth.uid"
 #   })
 # }
-# mutation DeleteEmail($emailId: UUID, $uid: String) @auth(level: PUBLIC) {
-## <type>_ delete lets you delete rows from your table.
-#   recipient_delete(key: {emailId: $emailId, userUid: $uid})
```
fredzqm, line 29: Deleting before we support the secure version in core module.

joehan (approved):
LGTM, tho maybe we should change the docs as well.

fredzqm:
I have cl/644125400 sent out to Mark

PR #7339, opened 2024-06-17
Better handling of connection string

joehan:
### Description
This PR does 2 things:
- Adding Postgres setup to 'setup:emulators:dataconnect'
- Error out if no connection string is set when we try to connect to postgres.

### Scenarios Tested
Running 'firebase setup:emulators:dataconnect' from outside a project directory, then inside a project directory
Screenshot 2024-06-17 at 4 06 06 PM

Try to start emulator with a connection string configured, error out, run the suggested command, and then successfully run the emulator
Screenshot 2024-06-17 at 4 25 21 PM

### DIFF
--- a/src/commands/setup-emulators-dataconnect.ts
+++ b/src/commands/setup-emulators-dataconnect.ts
@@ -1,11 +1,32 @@
 import { Command } from "../command";
-import { downloadEmulator } from "../emulator/download";
 import { Emulators } from "../emulator/types";
+import { Options } from "../options";
+import { DEFAULT_POSTGRES_CONNECTION } from "../init/features/emulators";
+import { promptOnce } from "../prompt";
+import { logger } from "../logger";
+import { downloadIfNecessary } from "../emulator/downloadableEmulators";
 
 const NAME = Emulators.DATACONNECT;
 
 export const command = new Command(`setup:emulators:${NAME}`)
   .description(`downloads the ${NAME} emulator`)
-  .action(() => {
-    return downloadEmulator(NAME);
+  .action(async (options: Options) => {
+    await downloadIfNecessary(NAME);
+    if (!options.config) {
+      logger.info(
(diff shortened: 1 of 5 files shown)


src/commands/setup-emulators-dataconnect.ts
```diff
@@ -1,11 +1,32 @@
...
+    const defaultConnectionString =
+      dataconnectEmulatorConfig?.postgres?.localConnectionString ?? DEFAULT_POSTGRES_CONNECTION;
+    // TODO: Download Postgres
+    const localConnectionString = await promptOnce({
+      type: "input",
+      name: "localConnectionString",
+      message: `What is the connection string of the local Postgres instance you would like to use with the Data Connect emulator?`,
+      default: defaultConnectionString,
+    });
+    options.rc.setDataconnect(localConnectionString);
```
fredzqm, line 31: So we decide to keep `firebase setup:emulators:dataconnect` instead of `firebase init emulators dataconnect`.

You probably know a lot more than I do on this. Is this a common pattern?

[non-blocking]

`init emulators` also setup port and other things. Do you think it make sense to include them or somehow re-use?

Screenshot 2024-06-17 at 4 38 40 PM

fredzqm (approved):

src/commands/setup-emulators-dataconnect.ts
joehan: We have both commands for all of the (downloadable) emulators - I went with suggesting `firebase setup:emulators:dataconnect` for this one because it was an easier 1 liner than `init emulators`, and we do not support `init emulators <productName>` (and doing so would require an API proposal + a surprising amount of rewiring in `init`) 

I did a bit of digging, and the raison d'etre for `setup:emulators:product` is for use in CI/CD environment set up scripts (ie Docker files or as a cached Github action step). With that in mind, I've gated the prompt so that it only happens in interactive environments.

I omitted port configuration to match other `setup:emulators:<product>` commands. I also don't think they are relevant for _most_ users.

PR #7349, opened 2024-06-20
update emulator to 1.2.2

hlshen:


joehan (approved):
lgtm after adding a changelog entry

Issue #7292, opened 2024-06-07
Firebase-tools v13.11.0 has crashlytics buildtools v3 that does not work on Intel-based Macs


Altair4Ru:
### [REQUIRED] Environment info
**firebase-tools:** 13.11.0+
**Platform:** MacOS

### [REQUIRED] Test case
The thing is, Crashlytics buildtools v3 now have dump_syms.bin for MacOS built for arm64 arch only, while it should be built as a universal binary
```
.//firebase-crashlytics-buildtools-2.9.9/dump_syms/macos/dump_syms.bin:   Mach-O 64-bit executable x86_64
.//firebase-crashlytics-buildtools-3.0.0/dump_syms/macos/dump_syms.bin:   Mach-O 64-bit executable arm64
```

### [REQUIRED] Steps to reproduce
Just call firebase-tools -- firebase crashlytics:symbols:upload --app={app_id_here} on Intel-based Mac

### [REQUIRED] Expected behavior

The symbols got uploaded without any errors

### [REQUIRED] Actual behavior

```
[CRASHLYTICS LOG DEBUG] Crashlytics Buildtools initialized.
[CRASHLYTICS LOG DEBUG] Extracting dump_syms from dump_syms/macos/dump_syms.bin to /<ProjectPath>/.crashlytics/dump_syms.bin
[CRASHLYTICS LOG DEBUG] Breakpad symbol generator initialized: /<ProjectPath>/.crashlytics/dump_syms.bin
...
java.io.IOException: Cannot run program "/<ProjectPath>/.crashlytics/dump_syms.bin": error=86, Bad CPU type in executable
  at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
  at com.google.firebase.crashlytics.buildtools.ndk.internal.breakpad.BreakpadSymbolGenerator.generateSymbols(BreakpadSymbolGenerator.java:119)
  at com.google.firebase.crashlytics.buildtools.Buildtools.generateNativeSymbolFiles(Buildtools.java:306)
  at com.google.firebase.crashlytics.buildtools.Buildtools.generateNativeSymbolFiles(Buildtools.java:280)
  at com.google.firebase.crashlytics.buildtools.CommandLineHelper.executeGenerateSymbols(CommandLineHelper.java:193)
  at com.google.firebase.crashlytics.buildtools.CommandLineHelper.executeCommand(CommandLineHelper.java:125)
[...]

PR #7357, opened 2024-06-21
Fix compatibility issue with Intel-based Macs uploading symbols

mrober:
This pulls in the latest Crashlytics buildtools jar, that fixes the Intel-based Macs compatibility issue.


mrober:
#7292

joehan (approved):

PR #7361, opened 2024-06-21
Refactor: extract template helpers into templates.ts

yuchenshi:
### Description

This makes template file paths all relative to `/templates/` and removes a lot of `__dirname` usages.

It allows source files to be moved around using TypeScript refactoring without breaking template paths.

Note that unlike #7332, moving template files themselves still requires manually updating string literals. (I've considered adding manifest files to templates but gave up since `/templates` are not in `/src`.)

### Scenarios Tested

See integration tests. Not every change is covered by tests though -- we may have to rely on careful code reviews on others.

### Sample Commands

N/A

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -150,8 +150,8 @@ are unavailable to Pull Requests coming from forks of the repository.
 | `src`           | Contains shared/support code for the commands             |
 | `src/bin`       | Contains the runnable script. You shouldn't need to touch this content. |
 | `src/commands`  | Contains code for the commands, organized by one-file-per-command with dashes. |
-| `src/templates` | Contains static files needed for various reasons (inittemplates, login success HTML, etc.) |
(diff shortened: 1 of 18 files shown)


CONTRIBUTING.md
```diff
@@ -150,8 +150,8 @@ are unavailable to Pull Requests coming from forks of the repository.
 | `src`           | Contains shared/support code for the commands             |
 | `src/bin`       | Contains the runnable script. You shouldn't need to touch this content. |
 | `src/commands`  | Contains code for the commands, organized by one-file-per-command with dashes. |
-| `src/templates` | Contains static files needed for various reasons (inittemplates, login success HTML, etc.) |
 | `src/test`      | Contains test helpers. Actual tests (`*.spec.ts`) should be colocated with source files. |
+| `templates`     | Contains static files needed for various reasons (init templates, login success HTML, etc.) |
```
yuchenshi, line 154: This is a mere correction to the documentation. No files are actually moved.

hlshen (approved):

PR #7368, opened 2024-06-24
Correction of a problem where export to dotenv fails if the config value is operated with empty characters in the 1st gen of functions.

ytetsuro:
### Description

Correction of a problem where export to dotenv fails if the config value is operated with empty characters in the 1st gen of functions.

### DIFF
--- a/src/functions/runtimeConfigExport.ts
+++ b/src/functions/runtimeConfigExport.ts
@@ -178,7 +178,7 @@ const CHARACTERS_TO_ESCAPE_SEQUENCES: Record<string, string> = {
 
 function escape(s: string): string {
   // Escape newlines, tabs, backslashes and quotes
-  return s.replace(/[\n\r\t\v\\"']/g, (ch) => CHARACTERS_TO_ESCAPE_SEQUENCES[ch]);
+  return (s || '').replace(/[\n\r\t\v\\"']/g, (ch) => CHARACTERS_TO_ESCAPE_SEQUENCES[ch]);
 }
 
 /**


src/functions/runtimeConfigExport.ts
```diff
@@ -178,7 +178,7 @@ const CHARACTERS_TO_ESCAPE_SEQUENCES: Record<string, string> = {
 
 function escape(s: string): string {
   // Escape newlines, tabs, backslashes and quotes
-  return s.replace(/[\n\r\t\v\\"']/g, (ch) => CHARACTERS_TO_ESCAPE_SEQUENCES[ch]);
+  return (s || '').replace(/[\n\r\t\v\\"']/g, (ch) => CHARACTERS_TO_ESCAPE_SEQUENCES[ch]);
```
colerogers, line 181: So the issue is that s is coming into this function as undefined from the env map. Let's change where this is actually being called below.

src/functions/runtimeConfigExport.ts
```diff
@@ -178,7 +178,7 @@ const CHARACTERS_TO_ESCAPE_SEQUENCES: Record<string, string> = {
...
 }
 
 /**
```
colerogers, line 184: On line 188, let's change it to something like
```
const lines = envs.map(({ newKey, value }) => `${newKey}="${value === undefined ? "" : escape(value)}"`);
```

ytetsuro: Thanks, fixed it.

taeold (commented):
This is great contribution - thank you!

Do you mind adding an entry in README.md that summarizes the change?

ytetsuro:
@taeold 

I consider this to be a bug fix.
I can't seem to find a place in README.md to mention change log.

Does this mean that a summary of the config:export command should be written in the following section?

https://github.com/firebase/firebase-tools/blob/master/README.md?plain=1#L144

I don't mind writing, but I am not good at English.
I would like to add the following sentence, is that OK?

`To migrate for environment variables, export your existing environment configurations using the firebase functions:config:export command.`

taeold:
A bug fix does have a place in the CHANGELOG.md. I usually write something short like:

"Fix bug where dotenv didn't support emtpy string as value"

PR #7371, opened 2024-06-24
App Hosting: add github branch validation

mathu97:
Ensures that the branch entered by user is a valid branch that is available in their repo.


src/apphosting/githubConnections.ts
```diff
@@ -257,6 +259,32 @@ async function promptCloneUri(
...
 }
 
+/**
+ * Prompts the user for a GitHub branch and validates that the given branch
+ * actually exists. User is re-prompted until they enter a valid branch.
+ */
+export async function promptGitHubBranch(repoLink: devConnect.GitRepositoryLink) {
+  const { lookupMap } = await devConnect.listAllBranches(repoLink.name);
+
+  let branch = DEFAULT_GITHUB_BRANCH;
```
tonyjhuang, line 269: You can remove this assignment since it's never used

src/apphosting/githubConnections.ts
```diff
@@ -257,6 +259,32 @@ async function promptCloneUri(
...
+  do {
+    branch = await promptOnce({
+      name: "branch",
+      type: "input",
+      default: DEFAULT_GITHUB_BRANCH,
```
tonyjhuang, line 274: optional: personally I would inline this and leave file-level comments to values that are used in multiple places. IMO this is easier to read:

`promptOnce({default: "main"})` vs `promptOnce({default: DEFAULT_BRANCH})`

src/apphosting/githubConnections.ts
```diff
@@ -257,6 +259,32 @@ async function promptCloneUri(
...
+      message: "Pick a branch for continuous deployment",
+    });
+
+    if (!lookupMap.has(branch)) {
+      utils.logWarning(
+        "the given branch does not exist, please enter a valid branch for your repo",
+      );
+    }
+  } while (!lookupMap.has(branch));
```
tonyjhuang, line 283: very optional suggestion, if you want to change this to use positive booleans instead (go/tott/685):

```
while true:
  branch = ...
  if (map.has(branch)) return branch
  utils.logWarning(...)
```

src/gcp/devConnect.ts
```diff
@@ -234,6 +234,46 @@ export async function listAllLinkableGitRepositories(
   return repos;
 }
 
+/**
+ * Lists all branches for a given repo. Returns an array of branches
+ * and a map to make lookups more efficient.
+ */
+interface ListAllBranchesResponse {
+  branches: string[];
+  lookupMap: Map<string, boolean>;
```
tonyjhuang, line 243: if this is just to help with lookups, how about `Set`? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

tonyjhuang (approved):
left some small comments feel free to merge after addressing

src/gcp/devConnect.ts
mathu97: makes sense!

PR #7373, opened 2024-06-24
Move 'location' from firebase.json to dataconnect.yaml

joehan:
### Description
Previously, location was provided as part of firebase.json:
```
## firebase.json
dataconnect: {
  source: 'dataconnect',
  location: 'us-east4',
}
```
Now, it is provided via dataconnect.yaml
```
specVersion: "v1alpha"
serviceId: "other-other-dataconnect"
location: "us-east4"
```

### Scenarios Tested
Tested the following cases using an emulator built from HEAD:
- Deployed dataconnect, and confirmed that it uses the right location
- firebase init dataconnect writes location to dataconnect.yaml, not firebase.json
- Ran SDK generation (via dataconnect:sdk:generate and via emulator:start), and confirmed that the correct location is passed through to the generated SDKs by searching the generated code.

I also tested the cases where someone still has the old way of setting location - we show a warning if location is set in firebase.json, and error out if it is missing in dataconnect.yaml:
Screenshot 2024-06-24 at 3 10 49 PM

### DIFF
--- a/src/dataconnect/fileUtils.ts
+++ b/src/dataconnect/fileUtils.ts
@@ -39,7 +39,7 @@ export async function readDataConnectYaml(sourceDirectory: string): Promise<Data
 }
 
 function validateDataConnectYaml(unvalidated: any): DataConnectYaml {
-  // TODO: Add validation
+  // TODO: Use json schema for validation here!
   if (!unvalidated["location"]) {
     throw new FirebaseError("Missing required field 'location' in dataconnect.yaml");
   }
--- a/src/dataconnect/fileUtils.ts
+++ b/src/dataconnect/fileUtils.ts
@@ -13,14 +13,13 @@ export function readFirebaseJson(config: Config): DataConnectMultiple {
     return [];
   }
   const validator = (cfg: any) => {
-    if (!cfg["source"] && !cfg["location"]) {
+    if (!cfg["source"]) {
       throw new FirebaseError(
-        "Invalid firebase.json: DataConnect requires `source` and `location`",
+        "Invalid firebase.json: DataConnect requires `source`",
       );
     }
     return {
       source: cfg["source"],
-      location: cfg["location"],
     };
   };
   const configs = config.get("dataconnect");
@@ -30,7 +29,7 @@ export function readFirebaseJson(config: Config): DataConnectMultiple {
     return configs.map(validator);
   } else {
     throw new FirebaseError(
-      "Invalid firebase.json: dataconnect should be of the form { source: string, location: string }",
+      "Invalid firebase.json: dataconnect should be of the form { source: string }",
     );
   }
 }
@@ -43,6 +42,9 @@ export async function readDataConnectYaml(sourceDirectory: string): Promise<Data
 
 function validateDataConnectYaml(unvalidated: any): DataConnectYaml {
   // TODO: Add validation
+  if (!unvalidated["location"]) {
+    throw new FirebaseError("Missing required field 'location' in dataconnect.yaml");
+  }
   return unvalidated as DataConnectYaml;
 }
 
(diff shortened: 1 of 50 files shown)


src/dataconnect/fileUtils.ts
```diff
@@ -43,6 +40,9 @@ export async function readDataConnectYaml(sourceDirectory: string): Promise<Data
 
 function validateDataConnectYaml(unvalidated: any): DataConnectYaml {
   // TODO: Add validation
+  if (!unvalidated["location"]) {
+    throw new FirebaseError("Missing required field 'location' in dataconnect.yaml");
+  }
   return unvalidated as DataConnectYaml;
```
fredzqm, line 46: OOC, Q: do we want to validate other fields as well?

Is there a way to make json schema do it for us?

src/dataconnect/prompts.ts
```diff
@@ -21,3 +21,10 @@ export async function promptDeleteConnector(
     utils.logLabeledSuccess("dataconnect", `Connector ${connectorName} deleted`);
   }
 }
+
+export function locationInFirebaseJsonWarning() {
```
fredzqm, line 25: Is this used? I saw the same string in `src/config.ts`

src/emulator/dataconnectEmulator.ts
```diff
@@ -139,16 +137,22 @@ export class DataConnectEmulator implements EmulatorInstance {
...
       "-v=2",
       "generate",
-      `--service_location=${args.locationId}`,
       `--config_dir=${args.configDir}`,
       `--connector_id=${args.connectorId}`,
     ];
     const res = childProcess.spawnSync(commandInfo.binary, cmd, { encoding: "utf-8" });
+
+    logger.debug(res.stderr);
     if (res.error) {
```
fredzqm, line 146: I think we would want to remove this. `--logtostderr` will re-direct all logs to stderr. Could be negligence on my end.

Have we tested it?

src/init/features/dataconnect/index.ts
```diff
@@ -69,7 +69,9 @@ export async function doSetup(setup: Setup, config: Config): Promise<void> {
 
   if (!config.has("dataconnect")) {
     config.set("dataconnect.source", dir);
-    config.set("dataconnect.location", info.locationId);
+  } else {
+    // Remove location if it was previously set.
+    config.set("dataconnect", { source: dir });
```
fredzqm, line 74: Q: why this branch?

Isn't this equivalent to always do `config.set("dataconnect", { source: dir });`?

src/dataconnect/prompts.ts
joehan: Oops, this was accidentally left in, removing.

src/emulator/dataconnectEmulator.ts
joehan: This does pipe all logs to stderr - however, it is the stderr of the childprocess, so it is not visible to the end user unless we do this.. This pipes that to the main logger at debug level, so that we can at least use it for debugging issues. 
We could alternatively log at info level if we want these to always be displayed, which looks like:
Screenshot 2024-06-24 at 6 28 19 PM
Its a bit noisy, but not too bad overall.

src/dataconnect/fileUtils.ts
joehan: That's a good point, and we defiitely should leverage the json schema here. Adding a todo for now.

src/init/features/dataconnect/index.ts
joehan: Good point. Fixed.

src/emulator/dataconnectEmulator.ts
fredzqm: Oh, thanks for the test run!

The info logs here looks helpful~ I was worrying if `if (res.error)` would be true in case of any stderror. Doesn't seem to be that case.

fredzqm (approved):

PR #7377, opened 2024-06-25
fix grammar: setup -> set up 

jarrensj:
fixing grammar for setup -> set up

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Below is a brief list of the available commands and their function:
 | **login:use**  | Set the default account to use for this project                                                                                                 |
 | **use**        | Set active Firebase project, manage project aliases.                                                                                            |
 | **open**       | Quickly open a browser to relevant project resources.                                                                                           |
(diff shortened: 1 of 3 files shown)


joehan (approved):

joehan:
Thanks for the contribution. Looks like you need to run 'npm run format', and then this should be good to go.

jarrensj:
> Thanks for the contribution. Looks like you need to run 'npm run format', and then this should be good to go.

Thank you!

package.json
```diff
@@ -222,7 +222,7 @@
     "eslint": "^8.56.0",
     "eslint-config-google": "^0.14.0",
     "eslint-config-prettier": "^9.1.0",
-    "eslint-plugin-jsdoc": "^48.0.1",
+    "eslint-plugin-jsdoc": "^48.4.0",
```
joehan, line 225: This seems unrelated to the typo fix - any reason for updating this?

jarrensj: This was added to the PR in the commit I had run the format command.

I can remove this from the PR.

Issue #7250, opened 2024-06-03
The command line is too long.


busanTeddyBear:
### [REQUIRED] Environment info

**firebase-tools : 13.10.2**

**Platform : Windows**

**Next.js : 14.2.3**

### [REQUIRED] Test case

N/A

### [REQUIRED] Steps to reproduce

1. run ```firebase deploy``` command

### [REQUIRED] Expected behavior

N/A

### [REQUIRED] Actual behavior

I'm using Next.js with hosting integration.
I'm getting the following error when deploying:

~~~
...

Building a Cloud Function to run this application. This is needed due to:
 • middleware
 • non-static component /[locale]/[...rest]/page
 • non-static component /[locale]/page

Unable to bundle next.config.mjs for use in Cloud Functions, proceeding with deploy but problems may be encountered.
The command line is too long.

...
~~~

busanTeddyBear:
My ```next.config.mjs``` file code is:

```mjs
import createNextIntlPlugin from "next-intl/plugin";

const withNextIntl = createNextIntlPlugin();

/** @type {import("next").NextConfig} */
const nextConfig = {};

export default withNextIntl(nextConfig);
```

leoortizz:
Hey @busanTeddyBear, can you please share your package.json?

busanTeddyBear:
@leoortizz 

Yes, of course.

This is the /hosting/package.json file content:

```json
{
    "name": "hosting",
    "version": "0.1.0",
    "private": true,
    "scripts": {
        "dev": "next dev",
        "build": "next build",
        "start": "next start",
        "lint": "next lint"
    },
    "dependencies": {
        "@heroicons/react": "^2.1.3",
        "@nextui-org/react": "^2.4.1",
        "framer-motion": "^11.2.10",
        "next": "^14.2.3",
        "next-intl": "^3.15.0",
        "next-themes": "^0.3.0",
        "react": "^18.3.1",
        "react-device-detect": "^2.2.3",
        "react-dom": "^18.3.1",
        "zustand": "^4.5.2"
    },
    "devDependencies": {
        "@types/node": "^20.14.2",
        "@types/react": "^18.3.3",
        "@types/react-dom": "^18.3.0",
        "autoprefixer": "^10.4.19",
        "eslint": "^8.57.0",
        "eslint-config-next": "^14.2.3",
        "eslint-plugin-perfectionist": "^2.10.0",
        "postcss": "^8.4.38",
        "prettier": "^3.3.2",
[...]

PR #7395, opened 2024-06-27
Use esbuild lib instead of npx

chalosalvador:
### Description

Fix for this https://github.com/firebase/firebase-tools/issues/7250 and https://github.com/firebase/firebase-tools/issues/6193 where `esbuild` execution was throwing an error saying "Command line too long" on Windows.

It tries to find an installed version of `esbuild` using `npm which`, if it is not found then it tries to install it. Once it is able to find a path to `esbuild`, it dynamically imports it and uses it to bundle the next.config file.

If the version of the installed esbuild doesn't match the required version, it will show a warning in the terminal.

### Scenarios Tested

- Tested with a project with a setup like described in https://github.com/firebase/firebase-tools/issues/7250 on a Windows machine and MacOS.


leoortizz:
@jamesdaniels this would address #7250 and #6193, WDYT?

jamesdaniels:
We use to do this. The trouble here is that firebase-tools pins dependencies. So we were pinning to a version of esbuild that was not supported on customer's machines, be it node version or architecture. 

Perhaps we could be smart and do some equivalent of "npm install" and "npm which"

src/frameworks/next/index.ts
```diff
@@ -16,6 +16,7 @@ import { parser } from "stream-json";
 import { pick } from "stream-json/filters/Pick";
 import { streamObject } from "stream-json/streamers/StreamObject";
 import { fileExistsSync } from "../../fsutils";
+import * as esbuild from "esbuild";
```
jamesdaniels, line 19: The concern here is that we'd need to pin esbluild, this led to numerous problems before since firebase-tools would be pinned to versions of esbuild that didn't support the node version developers were on or even their architecture (apple silicon was the source of many issues filed). If we could do something clever like how NPX installs on demand, then dynamically load from that path, then we'd have the best of both worlds IMO

jamesdaniels (changes requested):

src/frameworks/next/index.ts
```diff
@@ -74,8 +77,8 @@ import {
   ROUTES_MANIFEST,
   APP_PATH_ROUTES_MANIFEST,
   APP_PATHS_MANIFEST,
-  ESBUILD_VERSION,
   SERVER_REFERENCE_MANIFEST,
+  ESBUILD_VERSION,
```
jamesdaniels, line 81: Can we bump this dependency, change to a semver range, and warn if the existing global esbuild does not match? I understand esbuild is 0.x right now which makes this all riskier

jamesdaniels: `^0.19.2` should be fine IMO, they could break any point though

chalosalvador: I have updated the PR to use `^0.19.2` and show a warning if the installed version does not match.

jamesdaniels (approved):

src/frameworks/next/utils.ts
```diff
@@ -489,3 +492,52 @@ export async function whichNextConfigFile(dir: string): Promise<NextConfigFileNa
...
+export function findEsbuildPath(): string | null {
+  try {
+    const esbuildBinPath = execSync("npx which esbuild", { encoding: "utf8" }).trim();
+    const globalVersion = getGlobalEsbuildVersion();
+    if (globalVersion && !satisfies(globalVersion, ESBUILD_VERSION)) {
+      console.warn(
+        `Warning: Global esbuild version (${globalVersion}) does not match the required version (${ESBUILD_VERSION}).`,
+      );
+    }
+    return resolve(dirname(esbuildBinPath), "../esbuild");
```
leoortizz, line 508: does `../esbuild` resolves correctly on all platforms? I'm under the impression we need to use `join` here

chalosalvador: @leoortizz `resolve` will indeed resolve correctly on all platforms as it normalizes the path and converts it to an absolute path.

leoortizz (approved):

PR #7405, opened 2024-07-01
App Hosting: GitHub integration improvements

mathu97:
1. UX Update: Prompt for GitHub account / org first, then list repos for that account / org
2. Go directly to GitHub to install Firebase App Hosting GitHub app on a new account
3. Go directly to GitHub to manage an existing installation


src/apphosting/githubConnections.ts
```diff
@@ -155,10 +170,10 @@ async function createFullyInstalledConnection(
   location: string,
   connectionId: string,
   oauthConn: devConnect.Connection,
-  withNewInstallation = false,
+  withExistingInstallationId: string | undefined = undefined,
```
mathu97, line 173: this can be changed to just `installationId` since a connection should not be created without an installationId.

src/apphosting/githubConnections.ts
```diff
@@ -98,39 +99,55 @@ export async function linkGitHubRepository(
...
   const oauthConn = await getOrCreateOauthConnection(projectId, location);
-  const existingConns = await listAppHostingConnections(projectId, location);
+  let installationId = await promptGitHubInstallation(projectId, location, oauthConn);
 
-  if (existingConns.length === 0) {
-    existingConns.push(
-      await createFullyInstalledConnection(projectId, location, generateConnectionId(), oauthConn),
+  while (installationId === ADD_ACCOUNT_CHOICE) {
+    utils.logBullet(
+      "Install the Firebase App Hosting GitHub app on a new account to enable access to those repositories",
```
taeold, line 106: nit* not blocking

I wonder how surprising it is for users to ultimately be redirected to DevConnect. It looks friendly enough, and even says you should close the page, but I wonder if we should give a hint here about what to expect when a popup is open.

src/apphosting/githubConnections.ts
```diff
@@ -180,6 +195,107 @@ async function createFullyInstalledConnection(
   return conn;
 }
 
+async function manageInstallation(connection: devConnect.Connection): Promise<void> {
```
taeold, line 198: nit* it looks like "manage" is an intentional choice of word here but I was confused because I expected it to be "update" or "configure". Curious why you ended up deciding on "manage".

src/apphosting/githubConnections.ts
```diff
@@ -180,6 +195,107 @@ async function createFullyInstalledConnection(
...
+  utils.logBullet(
+    "Manage the Firebase App Hosting GitHub app to enable access to GitHub repositories",
+  );
+  const targetUri = connection.githubConfig?.installationUri;
+  if (!targetUri) {
+    throw new Error("No installation given to manage");
```
taeold, line 204: should we be throwing `FirebaseError` here?

If this is a user facing error, is there any guidance you can provide to help user "debug" this state (probalby just.. to try again?)

src/apphosting/githubConnections.ts
```diff
@@ -180,6 +195,107 @@ async function createFullyInstalledConnection(
...
+  );
+  if (connectionsMatchingInstallation.length === 0) {
+    return null;
+  }
+
+  if (connectionsMatchingInstallation.length > 1) {
+    // return the oldest connection
+    const sorted = devConnect.sortConnectionsByCreateTime(connectionsMatchingInstallation);
+    return sorted[0];
+  }
```
taeold, line 233: q: is there a situation where this is a valid state?

(if not, maybe we can comment on this being invalid but we'll go ahead and handle the situation gracefully by picking the oldest connection)

src/apphosting/githubConnections.ts
```diff
@@ -180,6 +195,107 @@ async function createFullyInstalledConnection(
...
+
+/**
+ * A "valid" installation is either the user's account itself or any orgs they
+ * have access to that the GitHub app has been installed on.
+ */
+export async function listValidInstallations(
+  projectId: string,
+  location: string,
+  connection: devConnect.Connection,
+) {
```
taeold, line 284: can we type the return value for this fn?

taeold (approved):

taeold (commented):
Can you add a changelog entry?

src/apphosting/githubConnections.ts
qtz1, line 106: Maybe this is a marissa question (this working of "adding an account" is on the console too). "Account" really means user or organization. I wonder if users are aware that "account" can also be an organization?

src/apphosting/githubConnections.ts
```diff
@@ -98,39 +99,55 @@ export async function linkGitHubRepository(
...
     }
 
-    const selection = await promptCloneUri(projectId, existingConns);
-    repoCloneUri = selection.cloneUri;
-    connection = selection.connection;
-  } while (repoCloneUri === ADD_CONN_CHOICE);
+    repoCloneUri = await promptCloneUri(projectId, connectionMatchingInstallation);
+  } while (repoCloneUri === MANAGE_INSTALLATION_CHOICE);
 
   // Ensure that the selected connection exists in the same region as the backend
```
qtz1, line 146: where's the logic that ensures this? or is this a todo?

qtz1 (approved):

src/apphosting/githubConnections.ts
mathu97: I thought "manage" makes sense  because it takes them to the installation page from which users can either, update the installation's repository access, suspend the installation, or uninstall the GitHub app. I think "configure" could work here as well.

Screenshot 2024-07-11 at 9 47 43 AM

src/apphosting/githubConnections.ts
mathu97: ya i think `FirebaseError` makes sense here and the error message here could be better. Something like "Failed to get installation URI. Please try again."?

src/apphosting/githubConnections.ts
mathu97: No I don't believe so. With the changes in this PR we want to ensure there isn't more than one connection per installation. But in the Console and CLI currently we are creating a connection and then choosing an installation, which makes it possible for there to be more than one connection for the same installation. So we should handle it gracefully.

I'll add a comment here.

src/apphosting/githubConnections.ts
mathu97: ah this seems like it was an old TODO. Will remove. When fetching or creating a connection we use the region that the user wants to create their backend in.

src/apphosting/githubConnections.ts
mathu97: hmm we could add some messaging here but I feel that this would make it too verbose. Ideally we redirect to a firebase branded success page which we can look into separately.

mathu97: talked about this with @qtz1 offline

PR #7406, opened 2024-07-02
Set a reasonable default dataconnect generated SDK output path

fredzqm:
### Description

Provide a default output path within `.dataconnect` folder, so by default we don't check-in generated SDK.

The existing behavior going through `firebase init dataconnect:sdk` flow with the default option is horrendous bad...

Screenshot 2024-07-01 at 8 01 54 PM

### Sample Commands

Screenshot 2024-07-01 at 7 48 19 PM

### DIFF
--- a/src/init/features/dataconnect/sdk.ts
+++ b/src/init/features/dataconnect/sdk.ts
@@ -106,7 +106,7 @@ export async function doSetup(setup: Setup, config: Config): Promise<void> {
       type: "input",
       default:
         newConnectorYaml.generate.kotlinSdk?.outputDir ||
-        `./../.dataconnect/generated/${newConnectorYaml.connectorId}/kotlin-sdk`,
+        `./../.dataconnect/generated/${newConnectorYaml.connectorId}/kotlin-sdk/src/main/kotlin/${newConnectorYaml.connectorId}`,
     });
     const pkg = await promptOnce({
       message: "What package name do you want to use for your Kotlin SDK?",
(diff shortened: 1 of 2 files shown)


maneesht (approved):
If you can wait for @cherylEnkidu and @dconeybe's reviews for Swift/Kotlin to cover all platforms, that'd be great

src/init/features/dataconnect/sdk.ts
```diff
@@ -101,7 +104,9 @@ export async function doSetup(setup: Setup, config: Config): Promise<void> {
     const outputDir = await promptOnce({
       message: `What directory do you want to write your Kotlin SDK code to? (If not absolute, path will be relative to '${connectorInfo.directory}')`,
       type: "input",
-      default: newConnectorYaml.generate.kotlinSdk?.outputDir,
+      default:
+        newConnectorYaml.generate.kotlinSdk?.outputDir ||
+        `./../.dataconnect/generated/${newConnectorYaml.connectorId}/kotlin-sdk`,
```
cherylEnkidu, line 109: For Kotlin SDK, it would be better to use `./../.dataconnect/generated/${newConnectorYaml.connectorId}/kotlin-sdk/src/main/kotlin/${newConnectorYaml.connectorId}`

cherylEnkidu (approved):

src/init/features/dataconnect/sdk.ts
fredzqm: Sure~

PR #7411, opened 2024-07-03
Use emulator issues stream notification to initiate schema reload

hlshen:


firebase-vscode/src/data-connect/emulator-stream.ts
```diff
@@ -59,19 +62,24 @@ export async function runEmulatorIssuesStream(
...
-export function displayAndHandleIssue(issue: EmulatorIssue, isPostgresEnabled: Signal<boolean>) {
+export function displayAndHandleIssue(
+  issue: EmulatorIssue,
+  isPostgresEnabled: Signal<boolean>,
+  schemaReloadFunc: () => void,
+) {
   const issueMessage = `Data Connect Emulator: ${issue.kind.toString()} - ${issue.message}`;
   if (issue.severity === Severity.ALERT) {
     vscode.window.showErrorMessage(issueMessage);
-  } else if (issue.severity === Severity.NOTICE) {
```
rosalyntan, line 66: Not blocking for this PR, but I wonder if we should chat to discuss if there should be a standard way for VSCode to handle different severity levels (i.e. would you prefer the `FILE_RELOAD` notification to be sent with `DEBUG` severity level, so you can keep displaying `NOTICE` as a warning message?)

firebase-vscode/src/data-connect/emulator-stream.ts
```diff
@@ -36,13 +37,15 @@ export async function runEmulatorIssuesStream(
...
   isPostgresEnabled: Signal<boolean>,
+  schemaReloadFunc: () => void,
 ) {
   const obsErrors = await getEmulatorIssuesStream(configs, fdcEndpoint);
   const obsConverter = {
     next(nextResponse: EmulatorIssueResponse) {
       if (nextResponse.result?.issues?.length) {
         for (const issue of nextResponse.result.issues) {
-          displayAndHandleIssue(issue, isPostgresEnabled);
+          console.log("HAROLD: ", issue)
```
rosalyntan, line 47: leftover log!

rosalyntan (approved):

Issue #2131, opened 2020-04-14
Functions emulator should only ignore what is in `functions.ignore`


samtstern:
### [REQUIRED] Environment info

**firebase-tools:** all

**Platform:** all

### [REQUIRED] Test case

Right now we have the following file watcher in the Functions emulator:
https://github.com/firebase/firebase-tools/blob/bd6a23be0ed9a7b06aa178aa66c94e6cb6ce8045/src/emulator/functionsEmulator.ts#L238

It ignores certain files based on best practices, however in `firebase.json` the developer can customize `functions.ignore`, we should match that exactly with our file watcher.

The main use case for this is when developers want to use local node modules, we are excluding them by matching all `node_modules`

### [REQUIRED] Steps to reproduce

N/A

### [REQUIRED] Expected behavior

See above

### [REQUIRED] Actual behavior

See above

TacB0sS:
Question: I was under the impression that **function.ignore** is also responsible for determining which part of the directory tree and files would not be deployed to the cloud function!

On the one hand serving locally I don't want to ignore the **node_modules** cause I work on the [project and the library](https://stackoverflow.com/questions/61011366/tsc-watch-rebuilds-project-but-not-including-changes-in-node-modules) and would like the function to listen on changes there while on the other hand I do **not** want to deploy the **node_modules** folder to the cloud function.

I believe this makes the **function.ignore** have a dual functionality that is potentially conflicting therefore imo should not be coupled.

Perhaps another entry in the **firebase.json** file would be a better fit?

samtstern:
@TacB0sS thanks for the feedback.  As you said, `functions.ignore` controls what is actually uploaded to Cloud Functions.  We want the local development environment to help you deploy confidently, so ignoring the same files means you will have almost the same experience locally and  in prod.  If you're relying on something that is in `node_modules` locally your functions will break in production!

Does that makes sense or am I misunderstanding your concern?

TacB0sS:
This is our firebase.json file:

```
{
  "functions": {
    "source": "app-backend",
    "ignore": [
      "src",
      ".config",
      "dist-test",
      "deploy.js",
      "node_modules"
    ],
    "predeploy": [
      "npm --prefix ./app-backend run build"
    ]
  },
  "hosting": {
    "public": "app-frontend/dist/dev",
    "redirects": [
      {
        "source": "/backend/api/:rPath*",
[...]

PR #7414, opened 2024-07-05
Pass down functions.ignore values in firebase.json to functions emulator

dikatok:
Pass down `functions.ignore` values in `firebase.json` to functions emulator so that supposedly ignored directories/files will not trigger reload.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Pass down `functions.ignore` values in `firebase.json` to functions emulator so that supposedly ignored directories/files will not trigger reload. (#7414)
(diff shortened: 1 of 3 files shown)


dikatok:
Whoops, apparently I forgot to checkout my new branch and committed the changes to the master branch of my fork, is that okay?

joehan (approved):
Thanks for the contribution! Lgtm, but could you add an entry to changelog.md for this change?

dikatok:
@joehan do you mean `changelog.md` in the root directory? is it supposed to be empty?

dikatok:
fixes #2131

PR #7420, opened 2024-07-08
Use only a single account in VSCode. Support IDX Auth

hlshen:
* Remove concept of multiple accounts in VSCode
* Pass in currentOptions object to CLI auth commands
* IDX auth is picked up by VSCode


firebase-vscode/src/cli.ts
```diff
@@ -32,116 +28,75 @@ import * as commandUtils from "../../src/emulator/commandUtils";
...
+    if (!options.user || !options.tokens) {
+      return false;
+    }
+
+    const optionsUser = options.user as User;
+    const optionsTokens = options.tokens as Tokens;
+    return (
+      account &&
+      account.user.email === optionsUser.email &&
+      account.tokens.access_token === optionsTokens.access_token
```
joehan, line 59: Do we want to compare access_token here? Is email sufficient to determine whether this is from the right account?

src/requireAuth.ts
```diff
@@ -52,7 +49,14 @@ async function autoAuth(options: Options, authScopes: string[]): Promise<void |
     // Make sure any error here doesn't block the CLI, but log it.
     logger.debug(`Error getting account credentials.`);
   }
+  if (process.env.MONOSPACE_ENV && token && clientEmail) {
+    // Within monospace, this a OAuth token for the user, so we make it the active user.
+    setActiveAccount(options, { user: {email: clientEmail}, tokens: {access_token: token} });
+    setGlobalDefaultAccount({ user: { email: clientEmail }, tokens: { access_token: token } });
 
+    // project is also slected in monospace auth flow
```
joehan, line 57: nit: selected

joehan (approved):
I think this mostly looks good! Need to run `npm run format` from the firebase-tools directory to make the linter happy tho.

Approvingh for now, and we can run thru it together in more detail later

joehan (approved):

Issue #5520, opened 2023-02-14
Emulator using `--project` with `demo-` prefix ignoring `.secret.local`


fedenusy:
### [REQUIRED] Environment info

**firebase-tools:** 11.23.1

**Platform:** ubuntu

### [REQUIRED] Test case

1. Initialize project
2. Uncomment `helloWorld` function and make it `runWith` any secret
3. Add `functions/.secret.local` containing a value for the secret
4. Run emulators using `--project` with `demo-` prefix 
5. Observe `.secret.local` getting ignored:
<img width="1055" alt="image" src="https://user-images.githubusercontent.com/1056523/218879454-8bdf5ca2-5c87-4424-b1c5-c09584da48dd.png">

### [REQUIRED] Steps to reproduce
1. `mkdir test && cd test`
2. `firebase init` > functions > any existing project > typescript
3. Change `functions/src/index.ts` to what's shown below.
4. Add `functions/.secret.local` as shown below.
5. `cd functions && npm run build && firebase emulators:start --only functions --project demo-project-id`

`functions/src/index.ts`:
```typescript
import * as functions from "firebase-functions";
import { defineSecret } from "firebase-functions/params";

defineSecret('SUPA_SECRET')

export const helloWorld = functions.runWith({ secrets: ['SUPA_SECRET'] }).https.onRequest((request, response) => {
  functions.logger.info("Hello logs!", {structuredData: true});
  response.send("Hello from Firebase!");
});
```

`functions/.secret.local`:
```
SUPA_SECRET=shhhh
```

NB: I also tried placing `.secret.local` at the repo root, and got the same result.

### [REQUIRED] Expected behavior
Emulators [should pick up `.secret.local`](https://firebase.google.com/docs/functions/config-env#secrets_and_credentials_in_the_emulator).

### [REQUIRED] Actual behavior

```
[2023-02-14T22:46:11.959Z] > command requires scopes: ["email","openid","https://www.googleapis.com/auth/cloudplatformprojects.readonly","https://www.googleapis.com/auth/firebase","https://www.googleapis.com/auth/cloud-platform"]
[2023-02-14T22:46:11.959Z] > authorizing via signed-in user (federico.nusymowicz@farmpro.ag)
i  emulators: Starting emulators: functions {"metadata":{"emulator":{"name":"hub"},"message":"Starting emulators: functions"}}
[...]

PR #7422, opened 2024-07-08
Fix bug where secret values provided in env files were parsed as list.

taeold:
Values for secret params defined using `defineSecret` are not supposed to be provided in env files. Instead, secret values should be created and managed in Cloud Secret Manager.

However, users often do include values for secret params in their env files, and today CLI incorrectly treats it as a list due to a bug that assumed that secret values will never be provided in the env files.

The change here treats values for secrets in env files as string.

Fixes https://github.com/firebase/firebase-functions/issues/1567

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,3 @@
 - Fixes framework support for Nuxt ^3.12 by correctly calling loadNuxtConfig() (#7375)
+- Fixes bug where secret values provided in env files were parsed as list (#7422)
+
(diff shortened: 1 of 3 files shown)


blidd-google (approved):

tzappia:
Nice to see this fixed! FYI, one of the reasons secrets end up in .env files is because of a bug when using the emulator and a demo- project. As a workaround to the bug, putting an empty secret in the .env file works, but perhaps some devs forget that this is just a workaround and the var should just be empty. See more here: https://github.com/firebase/firebase-tools/issues/5520#issuecomment-1746641238

PR #7438, opened 2024-07-11
Update the default Postgres connection string to be more descriptive

fredzqm:


src/init/features/emulators.ts
```diff
@@ -13,7 +13,8 @@ interface EmulatorsInitSelections {
 }
 
 // postgresql://localhost:5432 is a default out of the box value for most installations of Postgres
-export const DEFAULT_POSTGRES_CONNECTION = "postgresql://localhost:5432?sslmode=disable";
+export const DEFAULT_POSTGRES_CONNECTION = "postgresql://postgresusername:postgrespassword@localhost:5432/databasename?sslmode=disable";
```
joehan, line 16: This seems more like a placeholder value than a default value. 

We chose `postgresql://localhost:5432` originally because it will be right for some number of users. This will be correct for no users.

joehan (changes requested):
I'm open to switching this default, but I'm not sure this is going to actually be better for users.

Also, need to run 'npm run format'

src/init/features/emulators.ts
fredzqm: Do you remember if there is a way to reproduce cases that this might be correct for some users?

I was never able to get the original connection string to work. Always need to edit it.

My interpretation is to spell out enough defaults, so this also serves as a good documentation.

Issue #7401, opened 2024-06-30
Functions emulator should not requires authenticated secret manager calls if .secret.local is defined


glorat:
### [REQUIRED] Environment info

**firebase-tools: 13.12.0**

**Platform: macOS**

### [REQUIRED] Test case

A firebase function that depends on a secret. E.g.
```
export const processUploadedFile = functions
  .region(FUNCTION_REGION)
  .runWith({ secrets: FUNCTION_SECRETS })
  .storage.object()
  .onFinalize(processUploadedFileHandler)
```
Also have a .secret.local file that defines all required secrets

### [REQUIRED] Steps to reproduce

With no firebase or application default credentials in place (or being offline)
`firebase --project demo-my-project-id emulators:start`

### [REQUIRED] Expected behavior

Since .secret.local is defined, we should not worry if secrets manager is inaccessible (due to lack of credentials or being offline) since we assume user has populated .secret.local properly. I expect any issues to be handled at runtime with 
`⬢  functions: Unable to access secret environment variables from Google Cloud Secret Manager. Make sure the credential used for the Functions Emulator have access or provide override values in /.../functions/.secret.local:
`
as correctly handled by `resolveSecretEnvs`

### [REQUIRED] Actual behavior

`⬢  functions: Failed to load function definition from source: FirebaseError: HTTP Error: 401, Request is missing required authentication credential. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.`

Through debugging with local firebase-tools, this is happening in the functions deploy phase, where the issue is at
`resolveParams` which has the following code
```  
const [needSecret, needPrompt] = partition(outstanding, (param) => param.type === "secret");
  for (const param of needSecret) {
    await handleSecret(param as SecretParam, firebaseConfig.projectId);
  }
```
which requires caller to have access to the secret manager APIs to succeed.

Commenting out the call to handleSecret gives me the expected behaviour - if I have no credentials or even if I'm offline, the emulator starts with functions and my full stack operates as expected.

### Background
The background is that I'm trying to get the emulator to work while fully offline as I'll be travelling without internet for a while and have been following advice from #3916 and elsewhere. This is the only issue that needs to be resolved for me to work offline.

### Workaround
I have built a local copy of firebase-tools and commented out the `await handleSecret` line and am okay now.

### Proposed fix
[...]

PR #7443, opened 2024-07-12
Fixes secret params resolution in functions emulator

blidd-google:
The functions emulator will try to resolve secrets twice - we disable the secret resolution in deploy flow's `resolveBackend` (which is reused by the emulator to discover functions triggers) so the emulator can access handle emulator-specific secret resolution (i.e. prioritizing .secret.local).

Fixes https://github.com/firebase/firebase-tools/issues/7401.


aalej (commented):
This LGTM! Tested the PR using https://github.com/aalej/issues-7401?tab=readme-ov-file#pr-test. Could you add a changelog entry?

src/deploy/functions/build.ts
```diff
@@ -279,13 +279,15 @@ export async function resolveBackend(
   userEnvOpt: UserEnvsOpts,
   userEnvs: Record<string, string>,
   nonInteractive?: boolean,
+  isEmulator = false,
```
joehan, line 282: Optional: this list of args is quite long, and it would be easy to transpose nonInteractive and isEmulator. Consider switching to a parameter object. ie:
```
export async function resolveBackend( args: {
  build: Build,
  firebaseConfig: FirebaseConfig,
  userEnvOpt: UserEnvsOpts,
  userEnvs: Record<string, string>,
  nonInteractive?: boolean,
  isEmulator?: boolean,
  })
  ```

joehan (approved):
LGTM, with a suggestion

PR #7455, opened 2024-07-16
Data Connect commands work outside project root

joehan:
### Description
Switched dataconect commands to use config.path instead of path, so that command will work outside of the root directory.

### Scenarios Tested
Tested each command from projectRoot, projectRoot/dataconnect, and projectRoot/dataconnect/my-connector, and verfied that they work as expected.


fredzqm (approved):
Looks good. Good to know about `options.config`

PR #7458, opened 2024-07-16
Compatible Mode

bkhushi:
Added CLI support for schema migration compatibility mode

Testing: Pointed the CLI to staging environment, enabled the experiment, deployed a schema, then deployed schema again minus a table (did not prompt to deploy to Postgres database with the experiment enabled)


src/dataconnect/schemaMigration.ts
```diff
@@ -62,6 +63,23 @@ export async function migrateSchema(args: {
     databaseId,
     /* linkIfNotConnected=*/ true,
   );
+
+  if (experiments.isEnabled("compatiblemode")) {
+    if (schema.primaryDatasource.postgresql?.schemaValidation) {
```
lisajian, line 68: Ah, I see that `getIdentifiers()` gets called. By the time you hit this block, it's ensured that postgresql is defined. You can probably just simplify to:

```
if (experiments.isEnabled("compatiblemode")) {
   schema.primaryDatasource.postgresql.schemaValidation = "COMPATIBLE";
}
```

src/dataconnect/schemaMigration.ts
```diff
@@ -62,6 +63,23 @@ export async function migrateSchema(args: {
...
+          database: databaseId,
+          cloudSql: {
+            instance: instanceName,
+          },
+          schemaValidation: "COMPATIBLE",
+        },
+      };
+    }
+  }
+
```
rosalyntan, line 82: Looks great! Can you also add this into the `diffSchema` method as well? It might be worth extracting this into a helper method, to reduce duplication as well.

lisajian (commented):
Did you get a chance to try the CLI command locally? If so, would you mind putting some of the validation steps that you performed in the PR description?

src/experiments.ts
```diff
@@ -134,6 +134,13 @@ export const ALL_EXPERIMENTS = experiments({
     default: true,
     public: false,
   },
+
+  compatiblemode: {
+    shortDescription: "Enable schema migrations in Compatible Mode",
+    fullDescription: "Enable schema migrations in Compatible Mode",
+    default: true,
```
rosalyntan, line 141: I believe `default: false` (or just remove this field entirely) since we're not launching yet.

src/dataconnect/schemaMigration.ts
bkhushi: It still gives an error without the if statements

joehan (approved):
Some nits, but LGTM!

rosalyntan (approved):

lisajian (approved):

PR #7461, opened 2024-07-17
Ident sub init functions

joehan:
### Description
To steal a line from Fred, this was floor hanging fruit for clearing up confusion around 2 init options:

### Scenarios Tested

Screenshot 2024-07-17 at 11 15 28 AM


fredzqm (approved):
Thank you!

PR #7462, opened 2024-07-17
Add app folder platform check

hlshen:


src/dataconnect/fileUtils.ts
```diff
@@ -106,3 +106,23 @@ export async function pickService(
   }
   return serviceInfo;
 }
+
+// case insensitive indicators for supported app platforms
+const WEB_INDICATORS = ["package.json", "package-lock.json"];
+const IOS_INDICATORS = ["info.plist", "project.pbxproj", "podfile", "package.swift"];
```
joehan, line 112: @aashishpatil-g to review this list. Based on my minimal knowledge, i think we'd want to look for any file with .xcodeproj or .xcworkspace file extensions.

src/dataconnect/fileUtils.ts
```diff
@@ -106,3 +106,23 @@ export async function pickService(
...
+const ANDROID_INDICATORS = ["androidmanifest.xml", "build.gradle"];
+
+// given a directory, determine the platform type
+export async function getPlatformFromFolder(dirPath: string) {
+  const fileNames = await fs.readdir(dirPath);
```
joehan, line 117: We probably want to dive down into child directories here.

src/dataconnect/fileUtils.ts
```diff
@@ -106,3 +106,23 @@ export async function pickService(
...
   return serviceInfo;
 }
+
+// case insensitive indicators for supported app platforms
+const WEB_INDICATORS = ["package.json", "package-lock.json"];
+const IOS_INDICATORS = ["info.plist", "project.pbxproj", "podfile", "package.swift"];
+const ANDROID_INDICATORS = ["androidmanifest.xml", "build.gradle"];
+
+// given a directory, determine the platform type
+export async function getPlatformFromFolder(dirPath: string) {
```
joehan, line 116: Would love a unit test for this - you can stub out fs.readdir, and then add tests for a few different file lists. Right now these are pretty simple, but it will be nice to have tests to prevent accidental breaks in the future.

joehan (changes requested):
Some potential improvements, but overall approach LGTM. Would like unit tests for this tho.

src/dataconnect/fileUtils.ts
aashishpatil-g: Yes that is correct - look for folders whose names end in suffix .xcodeproj or .xcworkspace 

The project.pbxproj is an internal file within the .xcodeproj package folder, so won't be available at project folder level.

src/dataconnect/fileUtils.ts
hlshen: I'm still not certain how much we want to dive down. We are not supporting multiple app detections yet, right?

src/dataconnect/fileUtils.ts
hlshen: Are you suggesting we also check the selected folder's name for .xcodeproj or .xcworkspace?

src/dataconnect/fileUtils.ts
maneesht: Yes, but the user might have their app multiple directories in

src/dataconnect/fileUtils.ts
hlshen: are you ok with mock-fs dependency?

src/dataconnect/fileUtils.ts
hlshen: added, please review

src/dataconnect/fileUtils.ts
joehan: Yeah, that seems safe to add (and it'll be useful for me in unit testing the init refactor)

src/dataconnect/fileUtils.ts
```diff
@@ -106,3 +106,30 @@ export async function pickService(
   }
   return serviceInfo;
 }
+
+// case insensitive indicators for supported app platforms
+const WEB_INDICATORS = ["package.json", "package-lock.json"];
```
joehan, line 111: Consider checking for a directory named "node_modules" as another potential signal

src/dataconnect/fileUtils.ts
joehan: I think we went back and forth on this a bit in the meeting, but it is probably fine to omit downward recursion for now.

joehan (approved):
Thanks for unit tests!

src/dataconnect/fileUtils.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+        file1: "contents",
+        podfile: "cocoa pods yummy",
+        "androidmanifest.xml": "file found second :(",
+      },
+      output: Platform.IOS,
+    },
+  ];
+  for (const c of cases) {
+    it(c.desc, async () => {
+      mockfs({ [c.folderName]: c.folderItems });
```
joehan, line 95: I think you need to add an afterEach running mockfs.restore - otherwise this will leak across tests

joehan (changes requested):

src/dataconnect/fileUtils.ts
aashishpatil-g: These will be inside the app folder. In Finder the xcodeproj and .xcworkspace appear as files but at a filesystem level these are folders.

src/dataconnect/fileUtils.ts
```diff
@@ -106,3 +106,30 @@ export async function pickService(
...
+const WEB_INDICATORS = ["package.json", "package-lock.json", "node_modules"];
+const IOS_INDICATORS = ["info.plist", "project.pbxproj", "podfile", "package.swift"];
+const IOS_FOLDER_INDICATORS = [".xcworkspace", ".xcodeproj"];
+const ANDROID_INDICATORS = ["androidmanifest.xml", "build.gradle"];
+
+// given a directory, determine the platform type
+export async function getPlatformFromFolder(dirPath: string) {
+  // check for ios folder indicator
+  if (IOS_FOLDER_INDICATORS.some((indicator) => path.basename(dirPath).endsWith(indicator))) {
+    return Platform.IOS;
```
aashishpatil-g, line 120: Is this comparing the app folder path or a file / folder inside the app folder. (related to my other comment)

src/dataconnect/fileUtils.ts
hlshen: to confirm, a user would select an app folder, and within that folder would be an .xcodeproj folder

aashishpatil-g: Yes that is correct. Either a folder with suffix .xcodeproj or a folder with suffix .xcworkspace

joehan (approved):

aashishpatil-g (approved):

src/dataconnect/fileUtils.ts
```diff
@@ -106,3 +106,29 @@ export async function pickService(
   }
   return serviceInfo;
 }
+
+// case insensitive exact match indicators for supported app platforms
+const WEB_INDICATORS = ["package.json", "package-lock.json", "node_modules"];
+const IOS_INDICATORS = ["info.plist", "project.pbxproj", "podfile", "package.swift"];
```
aashishpatil-g, line 112: project.pbxproj is a file internal to the .xcodeproj folder (macOS package). We shouldn't create anything as a peer to this since this folder is internal to xcode and managed by xcode. I'd suggest not using this as an ios indicator. Instead just rely on xcodeproj and xcworkspace.

hlshen: done

PR #7464, opened 2024-07-17
Clearer errors on wrong service/connector id

joehan:
### Description
Screenshot 2024-07-17 at 3 59 46 PM

### Scenarios Tested
Tested that wrong connector or service ID throw the right error, and tested the correct service IDs, connector IDs, or serviceID:schema do not

### DIFF
--- a/src/deploy/dataconnect/prepare.ts
+++ b/src/deploy/dataconnect/prepare.ts
@@ -1,14 +1,17 @@
+import * as clc from "colorette";
+
 import { Options } from "../../options";
 import { load } from "../../dataconnect/load";
 import { readFirebaseJson } from "../../dataconnect/fileUtils";
 import { logger } from "../../logger";
 import * as utils from "../../utils";
 import { needProjectId } from "../../projectUtils";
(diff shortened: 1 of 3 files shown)


src/deploy/dataconnect/prepare.ts
```diff
@@ -28,6 +31,25 @@ export default async function (context: any, options: Options): Promise<void> {
   for (const si of serviceInfos) {
     si.deploymentMetadata = await build(options, si.sourceDirectory);
   }
+  const unmatchedFilters = filters?.filter((f) => {
+    // filter out all filters that match no service
+    const serviceMatched = serviceInfos.some((s) => s.dataConnectYaml.serviceId === f.serviceId);
```
fredzqm, line 36: Is this equivalent to this?
```
    return serviceInfos.some((s) => {
          return (
            s.dataConnectYaml.serviceId === f.serviceId &&
            (f.connectorId ? s.connectorInfo.some((c) => c.connectorYaml.connectorId === f.connectorId: true)
          );
        });
```
Feels more readable

fredzqm (approved):

PR #7467, opened 2024-07-18
Fix functions:secrets:set - stop erroring in non-interactive flows without force option

### LINKED ISSUE #7466: (Functions)The redeploy option is undefined, but an error message is displayed asking to use the redeploy option.
I use secrets with firebase functions.
I get the following error when I try to overwrite the secrets I am using.
However, the redeploy option does not exist.
It is also not possible to bypass redeploy as there is no option not to perform redeploy.

```
Error: Missing required options (redeploy) while running in non-interactive mode:
- redeploy Do you want to re-deploy the functions and destroy the stale version of secret FOO?
```

**firebase-tools:**

13.13.3

**Platform:**

Mac OS 14.4

### [REQUIRED] Test case

```bash
$ firebase functions:secrets:set --data-file secrets.local --redeploy FOO
```

```bash
$ firebase functions:secrets:set --data-file secrets.local --redeploy FOO
```

If secrets exist, they are redeployed.

Show following error.

```
error: unknown option '--redeploy'
```

Pull Request

https://github.com/firebase/firebase-tools/pull/7465

colerogers:
Fixes #7466


ytetsuro:
@colerogers 

Thank you.
If this PR is merge, it would be difficult to overwrite the secret manager value and terminate successfully without running redeploy.
Am I correct in my understanding?
I would rather not redeploy as I want to deploy stably.

src/commands/functions-secrets-set.ts
```diff
@@ -81,6 +82,12 @@ export const command = new Command("functions:secrets:set <KEY>")
         endpointsToUpdate.map((e) => `${e.id}(${e.region})`).join("\n\t"),
     );
 
+    if (options.nonInteractive && !options.force) {
+      throw new FirebaseError(
+        `Unable to destroy the stale version of secret ${secret.name} with the --force flag`,
```
joehan, line 87: THe error message here seems to be for cases were --force is included, but this if statement covers cases where it is not included.

colerogers:
@ytetsuro I think I understand the issue better, our behavior for running `functions:secrets:set` with responding `no`  to the prompt was different than running `functions:secrets:set --non-interactive`. The latter was producing an error that the former was not. This PR should now fix that.

joehan (approved):

PR #7469, opened 2024-07-18
feat: update supported range for angular

### LINKED ISSUE #7418: Web Frameworks Support for Angular 18
Hello,

I would like to request Firebase hosting web framework support for Angular 18

9kubczas4:
### Description

Angular 18 was released in May. There were no breaking changes between version 17 and 18 in terms of deployment to Firebase. This change fixes #7418, so when the user is creating a new Angular app using Firebase CLI, exactly via `firebase init`, then with this change the created version of an Angular App will be 18.

### Scenarios Tested

- Deploy Angular 18 app to Firebase
- Deploy Angular 18 app (SSG)
- Deploy Angular 18 app (SSR)


src/frameworks/angular/index.ts
```diff
@@ -36,7 +36,7 @@ export const docsUrl = "https://firebase.google.com/docs/hosting/frameworks/angu
 
 const DEFAULT_BUILD_SCRIPT = ["ng build"];
 
-export const supportedRange = "14 - 17";
+export const supportedRange = "15 - 18";
```
9kubczas4, line 39: btw @jamesdaniels shouldn't it be `16-18` based on https://angular.dev/reference/releases#actively-supported-versions?

jamesdaniels (approved):

PR #7479, opened 2024-07-22
Add CMEK support in Firebase CLI for Firestore databases

jinyangtang:
### Description

This lets Firebase CLI create a CMEK (Customer-managed encryption keys) Firestore database and view the key associated with a CMEK database.
- Create can take a `--kms-key-name` option to create a CMEK database. (No change for Update because we currently don't support update a CMEK database's key)
- Get displays `Kms Key Name` and `Active Key Versions` for CMEK databases.

### Scenarios Tested

*   [x] Creating a new database with no options, non-CMEK database is created by default
*   [x] Creating a new database with `kms-key-name` provided, a CMEK database is created
*   [x] Fail to create a database because `--kms-key-name` is misspelled
*   [x] Fail to create a database because `--kms-key-name` is provided with an empty value
*   [x] Fail to create a database because an invalid key name is provided in `--kms-key-name` option
*   [x] Fail to create a database because user didn't grant correct key permission to Firestore P4SA
*   [x] Get a CMEK database, it shows correct key and in-use key versions in `Kms Key Name` and `Active Key Version` rows
*   [x] Get a CMEK database, it shows only `Kms Key Name` when `Active Key Version` is not available from backend yet right after database creation (it needs 10min to populate)
*   [x] Get a non-CMEK database, it shows **no** `Kms Key Name` and `Active Key Version` rows

### Sample Commands

- `firebase firetore:databases:create <database>` # as before, to create a non-CMEK database
- `firebase firetore:databases:create <database> --kms-key-name <key>`


src/commands/firestore-databases-create.ts
```diff
@@ -63,13 +67,21 @@ export const command = new Command("firestore:databases:create <database>")
...
+    }
+
     const databaseResp: types.DatabaseResp = await api.createDatabase(
       options.project,
       database,
       options.location,
       type,
       deleteProtectionState,
       pointInTimeRecoveryEnablement,
+      cmekConfig,
```
joehan, line 84: Optional: Create database has 7 args - consider switching this to a parameter object (ie https://go/tott/550)

joehan (approved):
LGTM with some small suggestions

joehan:
Could you add a CHANGELOG entry for this?

src/commands/firestore-databases-create.ts
```diff
@@ -25,6 +25,10 @@ export const command = new Command("firestore:databases:create <database>")
     "--point-in-time-recovery <enablement>",
     "Whether to enable the PITR feature on this database, for example 'ENABLED' or 'DISABLED'. Default is 'DISABLED'",
   )
+  .option(
+    "--kms-key-name <kmsKeyName>",
+    "The resource ID of a Cloud KMS key. If set, the database created will be a Customer-managed Encryption Key (CMEK) database encrypted with this key. This feature is allowlist only in initial launch.",
```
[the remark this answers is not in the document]
jinyangtang: Sure, have added.

src/commands/firestore-databases-create.ts
jinyangtang: Have switched to a param object, thanks

PR #7485, opened 2024-07-23
Start of init refactor

joehan:
### Description
Splitting up the prompting and actuation logic of `init dataconnect` to set us up to init within VSCode. This is just data connect for now, but i'd like to follow this loose pattern for all inits.


fredzqm (approved):

src/init/features/dataconnect/index.ts
```diff
@@ -55,7 +56,19 @@ const defaultConnector = {
   ],
 };
 
+// doSetup is split into 2 phases - ask questions and then actuate files and API calls based on those answers.
 export async function doSetup(setup: Setup, config: Config): Promise<void> {
+  const info = await askQuestions(setup, config);
```
hlshen, line 61: In this setup, vscode will need both a doSetup function and a askQuestions function. Is this the expectation? I was thinking we could limit VSCode to need an askQuestions function, which has a set output type.

joehan: actuate is also exported - I'd suggest the VSCode implements it's own askQuestions() that returns RequiredInfo, and then directly call actuate

hlshen (approved):

PR #7502, opened 2024-07-29
Update marked and marked-terminal

joehan:
### Description
Updating marked + marked-terminal to latest versions, and make the necessary code changes to their usage.

Note: @types/marked-terminal doesn't cover the latest version, so we need to do  `markedTerminal() as any` to make TS happy. As a todo, I'll open a PR for those types and remove this hack when its released

### Scenarios Tested
Ran some com
Screenshot 2024-07-29 at 11 26 28 AM
mands that hit this and confirmed that the markdown text still shows up as expected.


ifielker (approved):
LGTM

PR #7503, opened 2024-07-29
Respect --force during deploy --only storage

### LINKED ISSUE #7499: `--force` does not apply to `firebase deploy --only storage`, so programmatically provision are not possible
### Environment info

**firebase-tools:** `13.14.2` (also tested with `13.0.0`)

**Platform:** macOS

### Test case

_I'd really like to provide this, but this is just a empty Firebase project, with nothing in it._

The steps to initate the Firebase project looks something like this:

- Add Firestore to project: `firebase projects:addfirebase "testproject-dev"`
- Use project: `firebase use testproject-dev`
- Enable API's: `gcloud services enable firestore.googleapis.com --project=testproject-dev`
- Create default Firestore DB: `firebase firestore:databases:create "(default)" --location "eur3" --project "testproject-dev"`
- Setup the defaultLocation for Firestore with the API endpoint: `/projects/${project_id}/defaultLocation:finalize?locationId=${location_id}`
- Initiate storage with `firebase init storage --project testproject-dev`
- Now try and deploy the storage: `firebase deploy --only storage --force --project testproject-dev`

### Steps to reproduce

- Follow the steps in the test case to initate the Firebase project, where defaultLocation are set, and a Firestore database are created
- Run `firebase init storage --project testproject-dev` to initate the Firebase Storage part
- Run `firebase deploy --only storage --force --project testproject-dev`
- See that a `Y/n` prompt shows up. That's my problem,  as I've added `--force` to the statement

### Expected behavior

I expect the `--force` param to force the IAM roles to be granted with**out** user interaction (_edit: fixed the missing 'out'_). 

### Actual behavior

I'm prompted with:

```
? Cloud Storage for Firebase needs an IAM Role to use cross-service rules. Grant the new role? (Y/n)
```

Even though the `-f, --force` states: "_bypass interactive prompts:_". I expect and need the prompt to be bypassed and answered with `Yes` (Y).

Right now, I'm unable to make a programmatically provisioned Firebase stack, which I'd like to do. For this project, I'm not going to use Terraform, due to a few known limits around Firebase V2 functions, but also other decisions. I also think the use of `--force` should be applied for this kind of yes/no questions.

If `--force` was not meant to handle the IAM Role specific prompt, I expect another flag to enforce a "yes" answer, on a prompt like `firebase deploy storage`. There's possibly other situations, where this applies, too.

joehan:
### Description
Automatically accept prompts during `firebase deploy --force --only storage`. Fixes #7499


aalej (approved):
LGTM! Tested using test case in #7499. Good catch updating the other prompt!

PR #7508, opened 2024-07-30
Explictly enable gen SDK and schema extension flag in fdc

fredzqm:


joehan (approved):
LGTM, but ideally wait until we cut the new emulator release to merge this in

Issue #7346, opened 2024-06-19
Sharp out of date


pm0u:
### [REQUIRED] Environment info

**firebase-tools:** 13.9.0

**Platform:** MacOS Sonoma 14.5

### [REQUIRED] Test case

`yarn create next-app test && cd test && yarn add sharp@latest && firebase deploy`

### [REQUIRED] Expected behavior

I can use the latest version of sharp.

### [REQUIRED] Actual behavior

When using the frameworks integration, `sharp` is out of date here: https://github.com/FirebaseExtended/firebase-framework-tools/blob/main/packages/firebase-frameworks/package.json#L58 which means that any projects that would like to deploy to firebase are unable to have an up to date version of sharp as it conflicts.

I tried setting `"overrides": { "sharp": "0.33.4" }` in package.json to match my installed version but the deploy errored out with no explanation so I guess there must be some breaking change somewhere.

leoortizz:
Hey @pm0u! The support to sharp `0.33.4` was released recently, can you please try it and report back here?

PR #7517, opened 2024-08-01
chore: update sharp version

emwp:
### Description
Updating sharp version to address issue related on issue: https://github.com/firebase/firebase-tools/issues/7346
Sharp version needs to be updated in two places, there is already a PR on the other repo:
https://github.com/FirebaseExtended/firebase-framework-tools/pull/227

### Scenarios Tested
Ran tests locally for Next.js 12 to 14 and Angular 14 to 17.
Also running locally with the updated sharp version from this PR: https://github.com/FirebaseExtended/firebase-framework-tools/pull/227 for firebase-framework.

### Sample Commands
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,4 +3,4 @@
 - Improved error message thrown when missing permissions to deploy to hosting. (#7190)
 - Improved default values for 'init dataconnect:sdk'.
 - Fixed issue where `functions:secrets:set` wrongly errorred in non-interactive flows without force option (#7467)
-- Updated `sharp` dependency version to 0.33.4.
+- Updated `sharp` dependency version to 0.33.4. (#7517)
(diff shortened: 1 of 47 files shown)


emwp:
> LGTM! Just add changelog then we can merge. Thanks!

Done!

CHANGELOG.md
```diff
@@ -3,3 +3,4 @@
 - Improved error message thrown when missing permissions to deploy to hosting. (#7190)
 - Improved default values for 'init dataconnect:sdk'.
 - Fixed issue where `functions:secrets:set` wrongly errorred in non-interactive flows without force option (#7467)
+- Updated `sharp` dependency version to 0.33.4.
```
leoortizz, line 6: the PR number would go here
```suggestion
- Updated `sharp` dependency version to 0.33.4. (#7517)
```

leoortizz (approved):

src/frameworks/constants.ts
```diff
@@ -32,7 +32,7 @@ export const FIREBASE_FRAMEWORKS_VERSION =
   DEFAULT_FIREBASE_FRAMEWORKS_VERSION;
 export const FIREBASE_FUNCTIONS_VERSION = "^4.5.0";
 export const FIREBASE_ADMIN_VERSION = "^11.11.1";
-export const SHARP_VERSION = "^0.32.1";
+export const SHARP_VERSION = "^0.33.4";
```
leoortizz, line 35: just realized ^0.33.4 would break for people using old CLI versions after seeing the E2E test failure, to keep it backwards compatible we need to support both versions
```suggestion
export const SHARP_VERSION = "^0.32 || ^0.33";
```

leoortizz (changes requested):

leoortizz (approved):

PR #7520, opened 2024-08-01
Update init dataconnect with movie schema

HYACCCINT:
### Description

Update the auto generated files for firebase init dataconnect to use a movie schema, add package names to swift and kotlin in connector.yaml


templates/init/dataconnect/connector.yaml
```diff
@@ -5,9 +5,11 @@ authMode: "PUBLIC"
...
 #     outputDir: <Path where you want the generated SDK to be written to, relative to this file>
-#     package: "@firebasegen/my-connector"
+#     package: "@firebasegen/default"
 #     packageJSONDir: < Optional. Path to your Javascript app's package.json>
 #   swiftSdk:
 #     outputDir: <Path where you want the generated SDK to be written to, relative to this file>
+#     package: "firebasegen/default"
 #   kotlinSdk:
 #     outputDir: <Path where you want the generated SDK to be written to, relative to this file>
+#     package: com.google.firebase.dataconnect.connectors.default
```
fredzqm, line 15: Denver suggests to change it to

`connectors.default`

fredzqm (approved):

templates/init/dataconnect/connector.yaml
fredzqm: https://github.com/firebase/firebase-tools/blob/master/src/init/features/dataconnect/sdk.ts#L127

BTW, it's already changed here

dconeybe: +1 to fredzqm's comment above

HYACCCINT: Do we still need this line in this pr then?
`com.google.firebase.dataconnect.connectors.default` is what was generated when I ran `firebase init dataconnect:sdk`

templates/init/dataconnect/queries.gql
```diff
@@ -1,56 +1,82 @@
...
 
 # # @auth() directives control who can call each operation.
-# # Only admins should be able to list all emails, so we use NO_ACCESS
-# query ListEmails @auth(level: NO_ACCESS) {
-#   emails {
-#     id, subject, text, sent
-#     from {
-#       name
-#     }
+# # List all movies, anyone can list all movies, the auth level is set to PUBLIC
```
joehan, line 4:
```suggestion
# # Anyone should be able to list all movies, so the auth level is set to PUBLIC
```

templates/init/dataconnect/queries.gql
```diff
@@ -1,56 +1,82 @@
...
-#   }) {
-#     id subject sent
-#     content: text # Select the `text` field but alias it as `content` in the response.
-#     sender: from { name address uid }
-#     # <field>_on_<foreign_key_field> makes it easy to grab info from another table
-#     # Here, we use it to grab all the recipients of the email.
-#     to: recipients_on_email {
-#       user { name address uid }
+
+# # Logged in user can lists all their reviews and movie titles associated with the review
```
joehan, line 19:
```suggestion
# # Logged in user can list all their reviews and movie titles associated with the review 
```

templates/init/dataconnect/schema.gql
```diff
@@ -1,28 +1,47 @@
-# # Example schema for simple email app
-# type User @table(key: "uid") {
-# 	uid: String!
-# 	name: String!
-# 	address: String!
+# # Example schema for simple movie
```
joehan, line 1:
```suggestion
# # Example schema for simple movie review app
```

templates/init/dataconnect/schema.gql
```diff
@@ -1,28 +1,47 @@
...
+
+# # Users
+# # Suppose a user can leave reviews for movies
+# # user-reviews is a one to many relationship, movie-reviews is a one to many relationship, movie:user is a many to many relationship
```
joehan, line 5:
```suggestion
# # user  -> reviews is a one to many relationship,
# # movie -> reviews is a one to many relationship
# # movie <-> user is a many to many relationship
```

joehan (changes requested):

templates/init/dataconnect/mutations.gql
```diff
@@ -1,35 +1,50 @@
...
-#     text: $content,
-#     subject: $subject,
-#     # Server values let your service populate sensitive data.
-#     # Users are only allowed to create emails sent from their account.
-#     fromUid_expr: "auth.uid",
-#     # Server values let your service populate data for you
-#     # Here, we use sent_date: { today: true } to set 'sent' to today's date.
-#     sent_date: { today: true }
-#   })
+# # Upsert (update or insert) a user based on their username
```
yuchenshi, line 18: I think this is more like "updating the username of the current user ID", no? It's certainly not based on the username, but based on `auth.uid`.

templates/init/dataconnect/queries.gql
```diff
@@ -1,56 +1,82 @@
...
-#     to: recipients_on_email {
-#       user { name address uid }
+
+# # Logged in user can list all their reviews and movie titles associated with the review 
+# # Since the query requires the uid of the current authenticated user, the auth level is set to USER
+# query ListUserReviews @auth(level: USER) {
+#   user(key: {id_expr: "auth.uid"}) {
+#     id
+#     username
+#     reviews: reviews_on_user {
```
yuchenshi, line 25: Consider adding a comment explaining aliasing and a comment explaining `<field>_on_<foreign_key_field>`, like the old example did.

templates/init/dataconnect/queries.gql
```diff
@@ -1,56 +1,82 @@
...
-#     to: recipients_on_email {
-#       user { name address uid }
+# # Search for movies, actors, and reviews
+# query SearchMovie(
+#   $titleInput: String
+#   $genre: String
+# ) @auth(level: PUBLIC) {
+#   movies(
+#     where: {
+#       _or: [
```
yuchenshi, line 71: Shouldn't this be `_and` (filtering based on both if both are passed)?

templates/init/dataconnect/schema.gql
```diff
@@ -1,28 +1,49 @@
...
-# 	name: String!
-# 	address: String!
+# # Example schema for simple movie review app
+
+# # Users
+# # Suppose a user can leave reviews for movies
+# # user  -> reviews is a one to many relationship,
+# # movie -> reviews is a one to many relationship
+# # movie <-> user is a many to many relationship
+# type User
```
yuchenshi, line 8: nit: `type User @table {` on the same line for readability. Ditto below

templates/init/dataconnect/schema.gql
```diff
@@ -1,28 +1,49 @@
...
+#   @table {
+#   id: String! @col(name: "user_auth")
+#   username: String! @col(name: "username", dataType: "varchar(50)")
+#   # The following are generated from the @ref in the Review table
```
yuchenshi, line 12: But the code does not contain `@ref` literally. Shall we reword this? e.g. "generated by the `user: User!` field in the Review table"

templates/init/dataconnect/mutations.gql
```diff
@@ -1,35 +1,50 @@
...
+#   $rating: Int!
+#   $reviewText: String!
+# ) @auth(level: USER) {
+#   review_upsert(
+#     data: {
+#       userId_expr: "auth.uid"
+#       movieId: $movieId
+#       rating: $rating
+#       reviewText: $reviewText
+#       reviewDate_date: { today: true }
```
yuchenshi, line 40: nit: Given that `reviewDate` has a `@default` in the schema, I think this is a good opportunity to leave this out but just add a comment explaining that it is set by schema default.

```suggestion
#       # reviewDate defaults to today in the schema. No need to set it manually.
```

yuchenshi (approved):
LGTM with nits

templates/init/dataconnect/connector.yaml
```diff
@@ -9,5 +9,7 @@ authMode: "PUBLIC"
 #     packageJsonDir: < Optional. Path to your Javascript app's package.json>
 #   swiftSdk:
 #     outputDir: <Path where you want the generated SDK to be written to, relative to this file>
+#     package: "firebasegen/default"
 #   kotlinSdk:
 #     outputDir: <Path where you want the generated SDK to be written to, relative to this file>
+#     package: com.connectors.default
```
dconeybe, line 15: Please get rid of the `com.` prefix for the kotlin packge; just `connectors.default`

dconeybe (changes requested):

templates/init/dataconnect/schema.gql
```diff
@@ -1,28 +1,46 @@
...
 
-# type Email @table {
-# 	subject: String!
-# 	sent: Date!
-# 	text: String!
-# 	from: User!
+# # Movies
+# type Movie @table {
+#   # The below parameter values are generated by default with @table, and can be edited manually.
+#   # implicitly calls @col to generates a column name. ex: @col(name: "movie_id")
```
yuchenshi, line 19:
```suggestion
#   # implies directive `@col(name: "movie_id")`, generating a column name
```

I'd avoid the word "call" with directives since directives aren't functions or methods.

templates/init/dataconnect/schema.gql
```diff
@@ -1,28 +1,46 @@
...
+#   # for UUID fields, @default(expr: "uuidV4()") is also implicitly called
```
yuchenshi, line 20: I don't think this is true. If the entire `id` field is omitted then `id: UUID! @default(expr: "uuidV4()")` is added, but I don't think we add defaults to explicit UUID-typed fields. Otherwise, there's no way to declare a UUID field _without_ a default, even if the developer wants to.

Feel free to file a bug if you feel like we should either change the behavior or the documentation, or both.

yuchenshi (approved):

templates/init/dataconnect/schema.gql
HYACCCINT: That makes sense! Updated!

joehan (approved):
LGTM

templates/init/dataconnect/schema.gql
yuchenshi: This line doesn't seem to be updated on my side. Is a git commit/push needed?

HYACCCINT: Hmmm, just pushed again, is it updated now?

yuchenshi: Yes, thanks!

dconeybe (approved):

PR #7522, opened 2024-08-02
Detect platform during `init dataconnect:sdk` instead of asking

joehan:
### Description
Asking users to type in paths is the worst. Just knowing is better.

Still have a few more things to test here:

- [x] Test iOS autodetection
- [x] Test Android autodetect
- [x] Test out the total fail case where we cannot detect at all

### Scenarios Tested
Tested auto-detection of a web app, from the firebase.json directory and from the app directory
Screenshot 2024-08-01 at 5 07 13 PM
Screenshot 2024-08-01 at 5 06 54 PM
IOS (tested with `friendlyeats-ios` app)
Screenshot 2024-08-02 at 11 53 09 AM
Android (tested with `friendlyeats-android` app)
Screenshot 2024-08-02 at 11 48 16 AM

### DIFF
--- a/src/dataconnect/fileUtils.ts
+++ b/src/dataconnect/fileUtils.ts
@@ -110,7 +110,7 @@ export async function pickService(
 // case insensitive exact match indicators for supported app platforms
 const WEB_INDICATORS = ["package.json", "package-lock.json", "node_modules"];
 const IOS_INDICATORS = ["info.plist", "podfile", "package.swift"];
-const ANDROID_INDICATORS = ["androidmanifest.xml", "build.gradle"];
+const ANDROID_INDICATORS = ["androidmanifest.xml", "build.gradle", "build.gradle.kts"];
 
 // endswith match
 const IOS_INDICATORS_2 = [".xcworkspace", ".xcodeproj"];
--- a/src/dataconnect/fileUtils.ts
+++ b/src/dataconnect/fileUtils.ts
@@ -132,3 +132,8 @@ export async function getPlatformFromFolder(dirPath: string) {
 
   return Platform.UNDETERMINED;
 }
+
+export async function directoryHasPackageJson(dirPath: string) {
+  const fileNames = await fs.readdir(dirPath);
+  return fileNames.some((f) => f.toLowerCase() === "package.json");
+}
--- a/src/dataconnect/types.ts
+++ b/src/dataconnect/types.ts
@@ -135,7 +135,7 @@ export interface Generate {
 export interface JavascriptSDK {
   outputDir: string;
   package: string;
-  packageJSONDir?: string;
+  packageJsonDir?: string;
 }
 
 export interface SwiftSDK {
--- a/src/init/features/dataconnect/sdk.ts
+++ b/src/init/features/dataconnect/sdk.ts
@@ -28,6 +28,7 @@ export type SDKInfo = {
   connectorYamlContents: string;
   connectorInfo: ConnectorInfo;
   shouldGenerate: boolean;
+  displayIOSWarning: boolean;
 };
 export async function doSetup(setup: Setup, config: Config): Promise<void> {
   const sdkInfo = await askQuestions(setup, config);
@@ -36,6 +37,7 @@ export async function doSetup(setup: Setup, config: Config): Promise<void> {
 
 async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
   const serviceCfgs = readFirebaseJson(config);
+  // TODO: This current approach removes comments from YAML files. Consider a different approach that won't.
   const serviceInfos = await Promise.all(
     serviceCfgs.map((c) => load(setup.projectId || "", config, c.source)),
   );
@@ -91,7 +93,7 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
         choices: [
           { name: "iOS (Swift)", value: Platform.IOS },
           { name: "Web (JavaScript)", value: Platform.WEB },
-          { name: "Androd (Kotlin)", value: Platform.ANDROID },
+          { name: "Android (Kotlin)", value: Platform.ANDROID },
         ],
       });
     }
@@ -102,6 +104,7 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
     newConnectorYaml.generate = {};
   }
 
+  let displayIOSWarning = false;
   if (targetPlatform === Platform.IOS) {
     const outputDir =
       newConnectorYaml.generate.swiftSdk?.outputDir ||
@@ -109,9 +112,11 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
         connectorInfo.directory,
         path.join(appDir, `generated/${newConnectorYaml.connectorId}`),
       );
-    const pkg = camelCase(newConnectorYaml.connectorId);
+    const pkg =
+      newConnectorYaml.generate.swiftSdk?.package ?? camelCase(newConnectorYaml.connectorId);
     const swiftSdk = { outputDir, package: pkg };
     newConnectorYaml.generate.swiftSdk = swiftSdk;
+    displayIOSWarning = true;
   }
 
   if (targetPlatform === Platform.WEB) {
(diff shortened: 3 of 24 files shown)


src/dataconnect/types.ts
```diff
@@ -135,7 +135,7 @@ export interface Generate {
 export interface JavascriptSDK {
   outputDir: string;
   package: string;
-  packageJSONDir?: string;
+  packageJsonDir?: string;
```
joehan, line 138: Separate change - this was just incorrect before

src/init/features/dataconnect/sdk.ts
```diff
@@ -54,19 +63,42 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
...
+      }),
+    );
+    const platformGuess = await getPlatformFromFolder(appDir);
+    if (platformGuess !== Platform.UNDETERMINED) {
+      logLabeledSuccess("dataconnect", `Detected ${platformGuess} app in directory ${appDir}`);
+      targetPlatform = platformGuess;
+    } else {
+      // If we still can't autodetect, just ask the user
+      logLabeledBullet("dataconnect", "Couldn't automatically detect your app's platform.");
+      targetPlatform = await promptOnce({
```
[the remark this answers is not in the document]
joehan: Gonna open a separate PR after to do this and to trigger the SDK flow from `init dataconnect` since both will require similar minor refactors

src/dataconnect/fileUtils.ts
```diff
@@ -110,7 +110,7 @@ export async function pickService(
 // case insensitive exact match indicators for supported app platforms
 const WEB_INDICATORS = ["package.json", "package-lock.json", "node_modules"];
 const IOS_INDICATORS = ["info.plist", "podfile", "package.swift"];
-const ANDROID_INDICATORS = ["androidmanifest.xml", "build.gradle"];
+const ANDROID_INDICATORS = ["androidmanifest.xml", "build.gradle", "build.gradle.kts"];
```
joehan, line 113: IIUC 'build.gradle.kts' is a slightly new equivalent of `build.gradle` (and is what we use in `friendlyeats-android`), so we should check for it too

yuchenshi (approved):
Edge case: If both web and Android indicators are present in the same directory, do we want to treat it as web (as implemented right now) or undetermined? If the former, what is the best order of priority?

src/init/features/dataconnect/sdk.ts
```diff
@@ -54,19 +62,38 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
...
+    } else {
+      // If we still can't autodetect, just ask the user
+      logBullet("Couldn't automatically detect your app's platform.");
+      targetPlatform = await promptOnce({
+        message: "Which platform do you want to set up a generated SDK for?",
+        type: "list",
+        choices: [
+          { name: "iOS (Swift)", value: Platform.IOS },
+          { name: "Web (JavaScript)", value: Platform.WEB },
+          { name: "Androd (Kotlin)", value: Platform.ANDROID },
```
maneesht, line 94: nit: `Android`

src/init/features/dataconnect/sdk.ts
```diff
@@ -75,53 +102,52 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
...
-        `./../gensdk/${newConnectorYaml.connectorId}/swift-sdk`,
-    });
+  if (targetPlatform === Platform.IOS) {
+    const outputDir =
+      newConnectorYaml.generate.swiftSdk?.outputDir ||
+      path.relative(
+        connectorInfo.directory,
+        path.join(appDir, `generated/${newConnectorYaml.connectorId}`),
+      );
     const pkg = camelCase(newConnectorYaml.connectorId);
```
maneesht, line 112: Do we want to check the current `newConnector.generate.swiftSdk.package` value here and default to the connector if there isn't already a value?

src/dataconnect/fileUtils.ts
maneesht: @dconeybe - can you check that we aren't missing anything here?

src/dataconnect/types.ts
maneesht: SGTM

src/init/features/dataconnect/sdk.ts
yuchenshi:
```suggestion
          { name: "Android (Kotlin)", value: Platform.ANDROID },
```

src/init/features/dataconnect/sdk.ts
```diff
@@ -75,53 +102,52 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
...
-        newConnectorYaml.generate.kotlinSdk?.outputDir ||
-        `./../gensdk/${newConnectorYaml.connectorId}/kotlin-sdk`,
-    });
+
+  if (targetPlatform === Platform.ANDROID) {
+    const outputDir =
+      newConnectorYaml.generate.kotlinSdk?.outputDir ||
+      path.relative(
+        connectorInfo.directory,
+        path.join(appDir, `generated/${newConnectorYaml.connectorId}`),
```
dconeybe, line 149: Omit `${newConnectorYaml.connectorId}` from the output dir as the kotlin codegen will automatically append directories corresponding to the "package", which usually incorporates the connector name.

For example, I tested this out on a connector named "foo". The code was then generated into `generated/foo/connectors/foo/` when it should have gone into `generated/connectors/foo/`.

dconeybe, line 149: This is a good start, but for the vast majority of Android projects this output directory will be nonsensical. For small, single-module Android projects that use the conventional directory structure, a better output directory would be `app/src/main`. However, `app` is only a convention, and it can be named anything. Maybe consider looking for `app/src/main` and using that directory if it exists.

dconeybe (changes requested):
Two random comments:

1. An irrelevant message about adding the project to Xcode is emitted when initializing the Kotlin SDK. Consider only emitting this message when configuring the Swift SDK:

```
i  Generated SDK code for foo
i  Please follow the instructions here to add your generated sdk to your XCode project:
        https://firebase.google.com/docs/data-connect/gp/ios-sdk#set-client
```

2. Any comments in the yaml file are discarded. I would expect comments to remain after "editing" the yaml file to add the new information.

src/init/features/dataconnect/sdk.ts
joehan: Good call.

src/init/features/dataconnect/sdk.ts
joehan: Good idea - added a check for that directory. If its not found, should fall back to the appDir (the current behavior), or are there other structure we ought to try?

joehan: Done

joehan:
> Two random comments:
> 
> 1. An irrelevant message about adding the project to Xcode is emitted when initializing the Kotlin SDK. Consider only emitting this message when configuring the Swift SDK:
> 
> ```
> i  Generated SDK code for foo
> i  Please follow the instructions here to add your generated sdk to your XCode project:
>         https://firebase.google.com/docs/data-connect/gp/ios-sdk#set-client
> ```
Good catch, will refactor to avoid that.

> 2. Any comments in the yaml file are discarded. I would expect comments to remain after "editing" the yaml file to add the new information.
This one is an unfortunate side effect of us marshalling + unmarshalling the YAML to edit it. I think the effort required to avoid this is not worth the time ATM. Adding a TODO to remember this tho

src/init/features/dataconnect/sdk.ts
```diff
@@ -142,12 +147,15 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
...
+      ? path.join(appDir, "app/src/main")
+      : appDir;
+
     const outputDir =
       newConnectorYaml.generate.kotlinSdk?.outputDir ||
-      path.relative(
-        connectorInfo.directory,
-        path.join(appDir, `generated/${newConnectorYaml.connectorId}`),
-      );
+      path.relative(connectorInfo.directory, path.join(baseDir, `generated`));
```
dconeybe, line 158: Do not add `generated` if `app/src/main` is used above. Only add `generated` if `appDir` is used above.

src/init/features/dataconnect/sdk.ts
```diff
@@ -142,12 +147,15 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
   }
 
   if (targetPlatform === Platform.ANDROID) {
+    // app/src/main is a common practice for Andorid, but not explicitly required.
+    // If it is present, we'll use it. Otherwise, we fall back to the app directory.
+    const baseDir = fs.existsSync(path.join(appDir, "app/src/main"))
+      ? path.join(appDir, "app/src/main")
```
dconeybe, line 153: I just realized, this should be `app/src/main/java`, not `app/src/main`.

src/init/features/dataconnect/sdk.ts
```diff
@@ -142,12 +147,15 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
...
+      : appDir;
```
dconeybe, line 154: Suggest `path.join(appDir, "generated")` instead of `appDir`, then remove the logic to append `"generated"` from line 158 below.

dconeybe (dismissed):

dconeybe:
My final comments were not addressed. Please fix as the output is suboptimal. 

Specifically, use `app/src/main/java` instead of `app/src/main` and only append `generated` if _not_ using `app/src/main/java`.

PR #7524, opened 2024-08-02
Listen for changes to extensions during deploy

ifielker:
Listen for changes to extensions during deploy


src/deploy/extensions/planner.ts
```diff
@@ -123,6 +126,64 @@ export async function have(projectId: string): Promise<DeploymentInstanceSpec[]>
   });
 }
 
+/**
+ * wantDynamic checks the users code for Extension SDKs usage and returns
```
joehan, line 130:
```suggestion
 * wantDynamic checks the user's code for Extension SDKs usage and returns
```

src/deploy/extensions/planner.ts
```diff
@@ -123,6 +126,64 @@ export async function have(projectId: string): Promise<DeploymentInstanceSpec[]>
...
+  projectNumber: string;
+  aliases: string[];
+  projectDir: string;
+  extensions: Record<string, DynamicExtension>;
+  emulatorMode?: boolean;
+}): Promise<DeploymentInstanceSpec[]> {
+  const instanceSpecs: DeploymentInstanceSpec[] = [];
+  const errors: FirebaseError[] = [];
+  for (const [instanceId, ext] of Object.entries(args.extensions)) {
+    const autoPopulatedParams = await getFirebaseProjectParams(args.projectId, args.emulatorMode);
```
joehan, line 145: A lot of this parameter handling logic is (nearly) duplicated here and in want(). Consider extracting into a separate function and calling it from both.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -422,6 +440,49 @@ function parseEndpointForBuild(
...
+  }
+  if (ex.localPath) {
+    refOrPath++;
+  }
+  if (refOrPath === 0) {
+    throw new FirebaseError("Expected either extension reference or local path in extension " + id);
+  }
+  if (refOrPath > 1) {
+    throw new FirebaseError(
+      "Multiple definitions for extension " +
```
joehan, line 464: Super nit: Prefer \`string ${templates}\` over string concatenation

src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -422,6 +440,49 @@ function parseEndpointForBuild(
...
+
+  let refOrPath = 0;
+  if (ex.ref) {
+    refOrPath++;
+  }
+  if (ex.localPath) {
+    refOrPath++;
+  }
+  if (refOrPath === 0) {
+    throw new FirebaseError("Expected either extension reference or local path in extension " + id);
```
joehan, line 460: Super nit: Prefer \`string ${templates}\` over string concatenation

src/extensions/runtimes/common.ts
```diff
@@ -0,0 +1,129 @@
...
+import { getFirebaseConfig } from "../../functionsConfig";
+import {
+  EndpointFilter as Filter,
+  targetCodebases,
+} from "../../deploy/functions/functionsDeployHelper";
+import { logger } from "../../logger";
+
+export { DynamicExtension } from "../../deploy/functions/build";
+
+const savedLoggerSilent = (logger as any).silent;
```
joehan, line 15: I don't like that `(logger as any).slient` is subverting the type system. Could you either change the type of logger so that this is not necessary, or implement silenceLogging() in logger.ts?

src/extensions/runtimes/common.ts
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * Extracts extensions from build records
+ * @param builds The builds to examine
+ * @param filters The filters to use
+ * @returns a record of extensions by extensionId
+ */
+export function extractExtensionsFromBuilds(builds: Record<string, Build>, filters?: Filter[]) {
+  const extRecords: Record<string, DynamicExtension> = {};
+  Object.entries(builds).forEach(([codebase, build]) => {
```
joehan, line 83: Prefer `for ... of...` over `forEach` (or you can implement this with .map/.filter` if you want to go functional) [go/tsjs-practices/iteration](https://goto.google.com/tsjs-practices/iteration)

src/extensions/types.ts
```diff
@@ -247,14 +251,65 @@ export interface Param {
...
 }
 
 export interface ParamOption {
   value: string;
   label?: string;
 }
+
+// Typeguard for ExtensionSpec. (We often get "specs" from parsing yaml).
+// This helps decide if it's actually a spec or just some random yaml.
+// Any is reasonable here since this is a typeguard.
```
joehan, line 273: Nit: I think this is a textbook use case for `unknown` over `any`: https://stackoverflow.com/questions/51439843/unknown-vs-any

Optional since i think it will require a decent bit of refactoring here.

src/extensions/types.ts
```diff
@@ -247,14 +251,65 @@ export interface Param {
...
+      if (!validSysParams) {
+        break;
+      }
+    }
+  } else {
+    validSysParams = false;
+  }
+  return (
+    typeof spec === "object" &&
+    Boolean(spec.name && spec.version && validResources && validParams && validSysParams)
```
joehan, line 313: Nit: Is this `Boolean` cast necessary?

src/deploy/extensions/prepare.ts
```diff
@@ -15,37 +15,53 @@ import { displayWarningsForDeploy, outOfBandChangesWarning } from "../../extensi
...
+// be true when called from functions/prepare (since extensions/prepare would
+// not be called). It is necessary otherwise you can get the same questions
+// and notifications twice (e.g. delete these extensions?)
+async function prepareHelper(
+  context: Context,
+  options: Options,
+  payload: Payload,
+  wantExtensions: planner.DeploymentInstanceSpec[],
+  noDeleteExtensions: planner.DeploymentInstanceSpec[],
+  isPrimaryCall: boolean,
```
joehan, line 40: `isPrimaryCall` strikes me as a code smell here - it seems like it may be a sign that we ought to split out the prompting code from the rest of the prepare code here.

If you want to avoid refactors here, maybe rename `isPrimaryCall` to `shouldDeleteExtensions` - AFAICT, that is the only behavior we actually gate on this.

joehan (changes requested):
Could you also manually test the existing flows and add screenshots in the description? Our automated tests _should_ cover those cases, but they only cover non-interactive flows.

src/deploy/extensions/planner.ts
ifielker: I tried this, and other than the first 4 lines, it turned into a big mess of if(isDynamic) everywhere. Complicated by the different types of extension (one is Record<string, string>, one is Record<string, DynamicExtension>). So I don't think this actually would be better after all.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
ifielker: Done.

src/deploy/functions/runtimes/discovery/v1alpha1.ts
ifielker: Done.

src/extensions/runtimes/common.ts
ifielker: Done. (I added silent to the Logger interface.)

src/extensions/runtimes/common.ts
ifielker: Done.

src/extensions/types.ts
ifielker: Refactored, N/A

src/deploy/extensions/prepare.ts
ifielker: Renamed.

joehan (approved):
Looks like you need to `npm run format`, but LGTM otherwise

PR #7530, opened 2024-08-06
Dont try to createDatabase when we know the instance is not ready

joehan:
### Description
Thanks @maneesht for finding this - during `firebase init`, we don't want to wait 15 minutes for instance creation. However, database creation will never succeed until the instance is ready. We should skip database creation when we know it won't work, and rely on the next deploy to create it.

### DIFF
--- a/src/dataconnect/provisionCloudSql.ts
+++ b/src/dataconnect/provisionCloudSql.ts
@@ -3,6 +3,7 @@ import * as utils from "../utils";
 import { grantRolesToCloudSqlServiceAccount } from "./checkIam";
 import { Instance } from "../gcp/cloudsql/types";
 import { promiseWithSpinner } from "../utils";
+import { logger } from "../logger";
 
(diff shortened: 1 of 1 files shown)


src/dataconnect/provisionCloudSql.ts
```diff
@@ -106,6 +111,7 @@ export async function provisionCloudSql(args: {
     } else {
       // Skip it if the database is not accessible.
       // Possible that the CSQL instance is in the middle of something.
+      logger.debug(`Unexpected error from CloudSQL: ${err}`);
```
maneesht, line 114: Do we want to keep this debug log?

maneesht (approved):

src/dataconnect/provisionCloudSql.ts
joehan: Yep, it might be useful in cases users hit this in the wild.

maneesht (approved):

fredzqm (approved):

PR #7532, opened 2024-08-06
Add writeConnector to file utils

hlshen:


src/dataconnect/fileUtils.ts
```diff
@@ -132,3 +132,38 @@ export async function getPlatformFromFolder(dirPath: string) {
 
   return Platform.UNDETERMINED;
 }
+
+// Generates sdk yaml based on platform, and returns a modified connectorYaml
+export function generateSdkYaml(
+  platform: Platform,
+  connectorYaml: ConnectorYaml,
+  connectorYamlPath:string,
```
joehan, line 140: Npm run format on firebase-tools

joehan (commented):
I think we bumped into each other on this one and https://github.com/firebase/firebase-tools/pull/7522/files - can you double check that the logic here is still up to date after merging master?

src/dataconnect/fileUtils.ts
```diff
@@ -137,3 +137,38 @@ export async function directoryHasPackageJson(dirPath: string) {
...
+  const relPath = path.relative(connectorYamlFolder, appFolder);
+  const outputDir = path.join(relPath, "dataconnect-generated");
+  if (!connectorYaml.generate) {
+    connectorYaml.generate = {};
+  }
+  if (platform === Platform.WEB) {
+    connectorYaml.generate.javascriptSdk = {
+      outputDir,
+      package: `@firebasegen/${connectorYaml.connectorId}`,
+      packageJsonDir: appFolder,
```
joehan, line 157: Edge case to check for when using this function - If i don't provide an app folder, what does this write to connector.yaml? Does it leave an empty `packageJsonDir:`?

src/dataconnect/fileUtils.ts
```diff
@@ -137,3 +137,38 @@ export async function directoryHasPackageJson(dirPath: string) {
...
+      package: connectorYaml.connectorId,
+    };
+  }
+  if (platform === Platform.ANDROID) {
+    connectorYaml.generate.kotlinSdk = {
+      outputDir,
+      package: `connectors.${connectorYaml.connectorId}`,
+    };
+  }
+
```
joehan, line 172: Nit: Unnecessary newline?

joehan (approved):
LGTM, thanks for adding tests! One case to call out

src/dataconnect/fileUtils.ts
hlshen: hrm, are you suggesting we check if appFolder is empty? outputDir is also being generated from appFolder

joehan: This is looking forward to follow up PRs that start using this function, so no need to block on this: 'packageJsonDir' is optional, to support users that use package managers other than npm. We should make sure that the UI makes it easy for users to omit it, and we should make sure that the final `connector.yaml` doesn't have an empty `packageJsonDir` field:

```
generate:
 webSdk:
    outputDir: ../../dir
    package: connectors.default
    packageJsonDir:
```

hlshen: ah, this would be for CLI init flow no? The auto detect doesn't have any web indicators aside from npm indiciators.

Issue #6779, opened 2024-02-16
Firebase deployment not working when using @ngx-env/builder


devomr:
### [REQUIRED] Environment info

**firebase-tools: 13.3.0**

**Platform: Windows**

### [REQUIRED] Test case

Angular SSR project using @ngx-env/builder and Firebase Hosting.

[angular.json](https://github.com/firebase/firebase-tools/files/14315580/angular.json)
[firebase.json](https://github.com/firebase/firebase-tools/files/14315583/firebase.json)
[package.json](https://github.com/firebase/firebase-tools/files/14315596/package.json)

### [REQUIRED] Steps to reproduce

1. Create Angular SSR project (version 17.1.0)
2. Add @ngx-env/builder (version 17.1.3) by running the command _ng add @ngx-env/builder_
3. Configure Firebase Hosting for detected codebase (Angular)

### [REQUIRED] Expected behavior

Deploy Angular application to Firebase Hosting.

### [REQUIRED] Actual behavior

The deployment not working and the following error is displayed _@ngx-env/builder:application (test:build:production) is not a recognized builder. Please check your angular.json_

Hope I provided all the necessary details.
Note: This issue started with firebase-tools version 13.0.0. I had a look on the source code and from that version was introduced a check that throws the error message which I encounter.

Really appreciate any help how can I fix this problem.

9kubczas4:
Hi @devomr,

After recent changes `firebase-tools` are supporting all builders, including `@ngx-env/builder`.
I've done tests using `firebase-tools@14.5.1`, `angular@20.0.0` and `@ngx-env/builder@19.0.4`.
All tests were done with this repository: https://github.com/9kubczas4/ngx-builder-sample.

Issue #6602, opened 2023-12-08
Error: @angular-builders/custom-webpack:browser is not a recognized builder. Please check your angular.json


ginobean:
### [REQUIRED] Environment info

13.0.1

**firebase-tools:**

Ubuntu

**Platform:**

### [REQUIRED] Steps to reproduce

firebase deploy (using @angular-builders/custom-webpack:browser as the builder).

### [REQUIRED] Expected behavior

Successful deployment.

### [REQUIRED] Actual behavior

Deployment fails with an error.

---
Interestingly, I downgraded from version from 13.0.1 to firebase-tools@12 and was able to successfully deploy thereafter..

Have NO idea what the error means or what the resolution should be..

ginobean:
[angular.json](https://github.com/firebase/firebase-tools/files/13621217/angular.json)

jamesdaniels:
With the support for Angular v17 and the new esbuild/vite based application builder, we now require explicit support for builders in our source code.

These new application builders broke the NPM command sequence and output directory structures that have been stable for years. While they are certainly a welcome improvement, we can no longer have those assumptions hardcoded.

ATM we're supporting known first party builders (Angular, Angular DevKit, and nguniversal). In this case it seems we need to add another browser builder, which shouldn't be a big deal, but obviously we'll want to check assumptions and write some test coverage.

Does that make sense?

ginobean:
That sounds good. Alternatively, could you use a config file to hold the list of supported builders? Might make it easier to add new builders going forwards, and allow us (users) to add to the list..

jamesdaniels:
In the meantime it looks like you're using the gen1 schematic from AngularFire, so `ng deploy` should still work. If you have a problem with that command file over in the AngularFire repo

ginobean:
I'll just stay on firebase-tools version 12 for now..

9kubczas4:
Hi @ginobean,

After recent changes, `firebase-tools` are supporting all builders, including `@angular-builders/custom-webpack`.
I've done tests using `firebase-tools@14.5.1`, `angular@19.0.0` and `@angular-builders/custom-webpack@19.0.1`.

PR #7557, opened 2024-08-14
feat: support custom webpack and ngx-env builders

9kubczas4:
### Description

This PR supports the custom Angular builders requested in the following issues:
- https://github.com/firebase/firebase-tools/issues/6779
- https://github.com/firebase/firebase-tools/issues/6602

### Scenarios Tested
Deployed Angular applications which uses following builders:
- custom webpack
- ngx-env builder.


src/frameworks/angular/utils.ts
```diff
@@ -79,10 +79,21 @@ const enum ExpectedBuilder {
   LEGACY_DEVKIT_SSR_DEV_SERVER = "@angular-devkit/build-angular:ssr-dev-server",
 }
 
+const enum ThirdPartyBuilder {
```
jamesdaniels, line 82: Qq, could we just simplify all this and only look after the last colon and expect it to conform to the builder-type? e.g, ":application", ":dev-server", ":server", ":browser", etc.

bboysokol: @jamesdaniels Regarding to this PR, I created another one with the requested changes:
https://github.com/firebase/firebase-tools/pull/8248

jamesdaniels (approved):
LGTM, thanks for taking to the finish line

PR #7574, opened 2024-08-20
Add --dry-run to deploy

joehan:
### Description
Adds a new `--dry-run` flag to `firebase deploy`. `firebase deploy --dry-run` will run any validations or build steps, and then report information about what change would have been made, without making any changes to your project.

Under the hood, this means that we only run the prepare step, and skip release/deploy. I reviewed `prepare.ts` for each product to ensure they don't make any production changes, and moved some validation from later stages to prepare where it belongs.

### DIFF
--- a/src/dataconnect/provisionCloudSql.ts
+++ b/src/dataconnect/provisionCloudSql.ts
@@ -42,7 +42,7 @@ export async function provisionCloudSql(args: {
     const why = getUpdateReason(existingInstance, enableGoogleMlIntegration);
     if (why) {
       const cta = dryRun
-        ? `It will be updated on you next deploy.`
+        ? `It will be updated on your next deploy.`
         : `Updating instance. This may take a few minutes...`;
       silent ||
         utils.logLabeledBullet(
--- a/src/dataconnect/provisionCloudSql.ts
+++ b/src/dataconnect/provisionCloudSql.ts
@@ -22,6 +22,7 @@ export async function provisionCloudSql(args: {
   enableGoogleMlIntegration: boolean;
   waitForCreation: boolean;
   silent?: boolean;
+  dryRun?: boolean;
 }): Promise<string> {
   let connectionName = ""; // Not used yet, will be used for schema migration
   const {
@@ -32,6 +33,7 @@ export async function provisionCloudSql(args: {
     enableGoogleMlIntegration,
     waitForCreation,
     silent,
+    dryRun,
   } = args;
   try {
     const existingInstance = await cloudSqlAdminClient.getInstance(projectId, instanceId);
@@ -46,15 +48,17 @@ export async function provisionCloudSql(args: {
             `Updating instance. This may take a few minutes...` +
             why,
         );
-      await promiseWithSpinner(
-        () =>
-          cloudSqlAdminClient.updateInstanceForDataConnect(
-            existingInstance,
-            enableGoogleMlIntegration,
-          ),
-        "Updating your instance...",
-      );
-      silent || utils.logLabeledBullet("dataconnect", "Instance updated");
+      if (!dryRun) {
+        await promiseWithSpinner(
+          () =>
+            cloudSqlAdminClient.updateInstanceForDataConnect(
+              existingInstance,
+              enableGoogleMlIntegration,
+            ),
+          "Updating your instance...",
+        );
+        silent || utils.logLabeledBullet("dataconnect", "Instance updated");
+      }
     }
   } catch (err: any) {
     // We only should catch NOT FOUND errors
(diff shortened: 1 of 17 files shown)


src/dataconnect/provisionCloudSql.ts
```diff
@@ -73,34 +77,44 @@ export async function provisionCloudSql(args: {
...
   } catch (err: any) {
     if (err.status === 404) {
+      if (dryRun) {
+        silent ||
+          utils.logLabeledBullet(
+            "dataconnect",
+            `Database ${databaseId} not found. It will be created on your next deploy.`,
+          );
+      }
       // Create the database if not found.
```
rosalyntan, line 118: Does this need to be wrapped in an `else`? Otherwise it'll run even in `dryRun`, right?

src/deploy/dataconnect/deploy.ts
```diff
@@ -96,6 +96,7 @@ export default async function (
           return Promise.resolve();
         }
         const enableGoogleMlIntegration = requiresVector(s.deploymentMetadata);
+        utils.logLabeledBullet("dataconnect", "Checking for CloudSQL resources...");
```
rosalyntan, line 99: nit: This log looks duplicated from line 83, should it be removed there?

src/deploy/dataconnect/prepare.ts
```diff
@@ -55,6 +58,37 @@ export default async function (context: any, options: Options): Promise<void> {
     filters,
   };
   utils.logLabeledBullet("dataconnect", `Successfully prepared schema and connectors`);
+  if (options.dryRun) {
+    utils.logLabeledBullet("dataconnect", "Checking for CloudSQL resources...");
+    return;
```
rosalyntan, line 63: Just checking -- does this need to return early? Since below, you pass `options.dryRun` into the `provisionCloudSql` call anyway?

src/deploy/dataconnect/prepare.ts
```diff
@@ -55,6 +58,37 @@ export default async function (context: any, options: Options): Promise<void> {
...
+          projectId,
+          locationId: parseServiceName(s.serviceName).location,
+          instanceId,
+          databaseId,
+          enableGoogleMlIntegration,
+          waitForCreation: true,
+          dryRun: options.dryRun,
+        });
+      }),
+  );
```
rosalyntan, line 91: This `await` section looks duplicated from deploy.ts, should it be removed there?

src/deploy/functions/prepare.ts
```diff
@@ -276,9 +277,12 @@ export async function prepare(
   // ===Phase 7. Finalize preparation by "fixing" all extraneous environment issues like IAM policies.
   // We limit the scope endpoints being deployed.
   await backend.checkAvailability(context, matchingBackend);
-  await ensureServiceAgentRoles(projectId, projectNumber, matchingBackend, haveBackend);
+  // TODO: CheckServiceAgentRoles when dryRun = true
```
ifielker, line 280: I assume you will do this TODO before you submit this PR?

src/deploy/functions/prepare.ts
```diff
@@ -276,9 +277,12 @@ export async function prepare(
...
+  options.dryRun ||
+    (await ensureServiceAgentRoles(projectId, projectNumber, matchingBackend, haveBackend));
   await validate.secretsAreValid(projectId, matchingBackend);
-  await ensure.secretAccess(projectId, matchingBackend, haveBackend);
+  // TODO: CheckSecretAccess when dryRun = true
```
ifielker, line 284: Also here

src/deploy/index.ts
```diff
@@ -117,7 +122,7 @@ export const deploy = async function (
 
   await chain(predeploys, context, options, payload);
   await chain(prepares, context, options, payload);
-  await chain(deploys, context, options, payload);
+  if (!options.dr) await chain(deploys, context, options, payload);
```
ifielker, line 125: s/dr/dryRun/

src/deploy/extensions/prepare.ts
```diff
@@ -122,6 +122,7 @@ async function prepareHelper(
   if (payload.instancesToDelete.length) {
     logger.info(deploymentSummary.deletesSummary(payload.instancesToDelete));
     if (
+      !options.dryRun &&
```
ifielker, line 125: So --dry run assumes you will be deleting things? Just want to make sure, since we are thinking of changing 
--force to not delete things.. ? Not saying it's wrong, just we need to make sure that --dry-run and --force are on the same page.. whatever that is.

src/deploy/functions/prepare.ts
ifielker, line 1: Are we expecting --dry-run not to enable APIs? Because I think it is still doing that.
There are service accounts being created to possibly. And I know there's somewhere in code that it asks if you want to delete functions no longer defined in the code base. I didn't see the if !options.dryRun for that one anywhere...

joehan: We generally need to enable APIs since we need them in order to actually dry run the API calls (and api enablement is low impact + reversible).

Good catch on the functions prompts, will add a conditional there too.

milyes (approved):

src/deploy/functions/prepare.ts
joehan: Double checked this - that prompt is in the release stage, which won't be executed during --dry-run, so no need for a conditional

joehan (commented):
Sorry for the sloppiness here, and ty for the detailed reviews!

src/dataconnect/provisionCloudSql.ts
joehan: Yes it does, good catch

milyes (approved):

milyes (approved):

src/dataconnect/provisionCloudSql.ts
```diff
@@ -32,29 +33,33 @@ export async function provisionCloudSql(args: {
...
+    dryRun,
   } = args;
   try {
     const existingInstance = await cloudSqlAdminClient.getInstance(projectId, instanceId);
     silent || utils.logLabeledBullet("dataconnect", `Found existing instance ${instanceId}.`);
     connectionName = existingInstance?.connectionName || "";
     const why = getUpdateReason(existingInstance, enableGoogleMlIntegration);
     if (why) {
+      const cta = dryRun
+        ? `It will be updated on you next deploy.`
```
rosalyntan, line 45: nit: s/you/your

rosalyntan (approved):

src/deploy/dataconnect/prepare.ts
```diff
@@ -55,6 +58,36 @@ export default async function (context: any, options: Options): Promise<void> {
     filters,
   };
   utils.logLabeledBullet("dataconnect", `Successfully prepared schema and connectors`);
+  if (options.dryRun) {
```
rosalyntan, line 61: From chatting offline -- we should invoke `diffSchema` here to surface migration diffs in `dryRun`

PR #7578, opened 2024-08-21
Fdc sql perms

tammam-g:
### Description

This pull request changes how we setup SQL permissions inside postgres as part of the dataconnect:migrate command. The purpose of this change is to reduce the radius of permissions we grant to improve the database security.

The new migration flow is the following:
 - Revoke cloudsqlsuperuser from firebase owner provide backward compatibility.
 - Creates and sets up owner/writer/reader firebase SQL roles.
 - Grant cloudsqlsuperuser access to these roles.
 - Grant dataconnect P4SA access to the writer role.
 - Execute `Create Extension` commands with "cloudsqlsuperuser" role.
 - Execute remaining diffs under the IAM user firebaseowner_* role .

This change is backward compatible and effects already existing users minimally. We automatically revoke the cloudsqlsuperuser permission from firebaseowner to avoid cyclic dependencies/errors with the new roles setup but we don't revoke any other permissions that have been granted to PUBLIC.

Note, downgrading CLI version will require manual intervention, specifically users will need to manually run `REVOKE "firebaseowner_postgres_public" FROM "cloudsqlsuperuser"`. 

## Test 1
 1. Setup new SQL instance and FDC service.
 2. Ran the original `firebase dataconnect:sql:migrate` to apply schema changes and extensions.
 3. Dropped extensions and schema changes from CloudSQL console.
 4. Ran the new  `firebase dataconnect:sql:migrate` and confirmed diffs applied with no errors.

## Test 2
 1. Created a new SQL instance and FDC service. 
 2. Ran the new  `firebase dataconnect:sql:migrate` and confirmed diffs applied, extensions installed with no errors. 
 3. Confirmed dataconnect P4SA can still read and write to the db and use vector extension search.

### Sample Commands

No changes.


src/gcp/cloudsql/connect.ts
```diff
@@ -105,82 +106,102 @@ export async function execute(
...
+  const projectId = needProjectId(options);
+
+  // 0. Get the current identity
+  const { user, mode } = await getIAMUser(options);
+
+  // 1. Create an IAM user for the current identity.
+  await cloudSqlAdminClient.createUser(projectId, instanceId, mode, user);
+
+  // 2. Setup FDC required SQL roles and permissions.
+  await setupSQLPermissions(instanceId, databaseId, options, true);
```
fredzqm, line 191: `silent = true`

In general, literal argument should be named. Hard to guess what it's for.

src/gcp/cloudsql/permissions.ts
```diff
@@ -0,0 +1,267 @@
...
+          AND grantee.rolname = '${granteeRole}';
+
+        -- If no rows were found, raise an exception
+        IF role_count = 0 THEN
+            RAISE EXCEPTION 'Role "%", is not granted to role "%".', '${grantedRole}', '${granteeRole}';
+        END IF;
+    END $$;
+`;
+  try {
+    await executeSqlCmdsAsIamUser(options, instanceId, databaseId, [checkCmd], true);
```
fredzqm, line 64: `silent = true`

src/gcp/cloudsql/permissions.ts
```diff
@@ -0,0 +1,267 @@
...
+  const projectId = needProjectId(options);
+  const requiredPermissions = [
+    "cloudsql.databases.create",
+    "cloudsql.databases.update",
+    "cloudsql.instances.connect",
+    "cloudsql.instances.create",
+    "cloudsql.instances.get",
+    "cloudsql.instances.list",
+    "cloudsql.instances.update",
+    "cloudsql.users.create",
```
fredzqm, line 87: Only `cloudsql.users.create` and `cloudsql.instances.connect` matters for CLI's use case

We already have a check in `dataconnect:sql:migrate`. We should remove `cloudsql.users.create` there, so non-admin developers can run `sql:migrate` as well. And perform admin check in application logics.

https://github.com/firebase/firebase-tools/blob/d4ac884423e9cca014e3a31e619c6b2e176caa02/src/commands/dataconnect-sql-migrate.ts#L14-L19

src/gcp/cloudsql/permissions.ts
```diff
@@ -0,0 +1,267 @@
...
+  // 1. Create a temporary builtin user
+  const superuser = "firebasesuperuser";
+  const temporaryPassword = utils.generateId(20);
+  await cloudSqlAdminClient.createUser(
+    projectId,
+    instanceId,
+    "BUILT_IN",
+    superuser,
+    temporaryPassword,
+  );
```
fredzqm, line 222: I think `executeSqlCmdsAsSuperUser` below already does this.

src/gcp/cloudsql/permissions.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  // 2. Upsert dataconnenct P4SA user in case it's not created.
+  const projectNumber = await needProjectNumber(options);
+  const { user: dataconnectDBUser, mode } = toDatabaseUser(getDataConnectP4SA(projectNumber));
+  await cloudSqlAdminClient.createUser(projectId, instanceId, mode, dataconnectDBUser);
```
fredzqm, line 227: nit: move P4SA IAM user creation to the caller next to the user IAM user creation.
I was looking for that there.

This way `setupSQLPermissions` only deals with SQL permissions.

src/gcp/cloudsql/permissions.ts
```diff
@@ -0,0 +1,267 @@
...
+    projectId,
+    instanceId,
+    "BUILT_IN",
+    superuser,
+    temporaryPassword,
+  );
+
+  // 2. Upsert dataconnenct P4SA user in case it's not created.
+  const projectNumber = await needProjectNumber(options);
+  const { user: dataconnectDBUser, mode } = toDatabaseUser(getDataConnectP4SA(projectNumber));
```
fredzqm, line 226: call it `fdcP4SAUser`

src/gcp/cloudsql/connect.ts
```diff
@@ -105,82 +106,102 @@ export async function execute(
...
+
+  // 3. Grant firebaseowner role to the IAM user.
+  const grants = [`GRANT "${firebaseowner(databaseId)}" TO "${user}"`];
+
+  await executeSqlCmdsAsSuperUser(options, instanceId, databaseId, grants, true);
```
fredzqm, line 196: nit: I was looking for GRANT current user owner in `setupSQLPermissions`.

Re-naming suggestions to the line easier: What about?
- `setupIAMUser` just upserts two IAM user (user and P4SA)
- `setupSQLPermissions` just deals with SQL permissions. The last grant is also included in the big setup SQL array.

The caller calls both.

fredzqm (commented):
Looks really good! High code standard! I only have minor nits left.

Do you think it is safe to release this regularly?
Can you test it with an database setup using old CLI and verify the upgrade process?

After that, LGTM

src/gcp/cloudsql/connect.ts
tammam-g: It's not possible to name arguments in typescript/javascript afaik. Do you mean creating a variable for it `const silent = true`?

src/gcp/cloudsql/permissions.ts
tammam-g: True, I forgot to remove this.

src/gcp/cloudsql/permissions.ts
tammam-g: My intuition here is the `setupSQLPermissions` is what would run if a user runs `dataconnect:sql:setup` . While `setupIamUser` is only responsible for setting up the IamUser which uses `setupSQLPermissions` to make sure everything is setup before granting IAM user the owner role. I can change it but that's the reason I kept p4sa setup here.

src/gcp/cloudsql/connect.ts
tammam-g: Responsed in another thread. Willing to change if you still think it's better to combine them, the biggest reason is that I wanted `setupSQLPermissions` to be what runs in `dataconnect:sql:setup`.

src/gcp/cloudsql/permissions.ts
tammam-g: Done

tammam-g:
> Do you think it is safe to release this regularly? Can you test it with an database setup using old CLI and verify the upgrade process?

It should be safe. I tried it on a new database, used the current latest version of `firebase deploy`, then changed the schema a bit and ran the new version `fb deploy`.  A message saying `Detected cloudsqlsuperuser was previously given to firebase owner, revoking to improve database security.` is printed (silently the setup runs) and the diff is then executed.

src/gcp/cloudsql/permissions.ts
tammam-g: Thanks, good catch. I remove `cloudsql.users.create` from the migrate  requirements. I updated the isAdmin permissions to be check for less permissions but I think `cloudsql.instances.login` and `cloudsql.instances.get` should remain.

src/gcp/cloudsql/connect.ts
tammam-g: Added a gsdoc comment

src/gcp/cloudsql/permissions.ts
tammam-g: Moved p4sa creation and grants to setupIamUsers after our convo

src/gcp/cloudsql/connect.ts
tammam-g: Moved p4sa creation and grants to setupIamUsers after our convo

src/gcp/cloudsql/permissions.ts
fredzqm: Technically, you can connect via sql-proxy without `cloudsql.instances.login`.

It's pretty confusing, but `login` is only needed for manual IAM authentication, not used in CLI.

https://cloud.google.com/sql/docs/postgres/iam-logins#logging-in-as-a-user

fredzqm:
Looks good to me~ Left one comment on `cloudsql.instances.login`.

@tammam-g Can you update CHANGELOG?

cc @joehan in case you want to take a look. Optional though.

fredzqm (approved):

src/api.ts
```diff
@@ -151,6 +151,11 @@ export const githubClientSecret = () =>
 
 export const dataconnectOrigin = () =>
   utils.envOverride("FIREBASE_DATACONNECT_URL", "https://firebasedataconnect.googleapis.com");
+export const dataconnectP4SADomain = () =>
```
joehan, line 154: Not blocking: Is this something that will need to be overridden ever? It seems like all p4sas will have this domain. If it won't be overridden, you could simplify this to just a normal constant in schemaMigration.ts

src/gcp/cloudsql/permissions.ts
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// Returns true if "grantedRole" is granted to "granteeRole" and false otherwise.
+// Throw an error if commands fails due to another reason like connection issues.
+export async function checkRoleIsGranted(
+  options: Options,
+  instanceId: string,
+  databaseId: string,
+  grantedRole: string,
+  granteeRole: string,
```
joehan, line 28: Consider using the parameter object pattern here - this is a long list of args of the same type and it would be easy for a future dev to transpose args. See go/tott/550

joehan (approved):
Some messaging suggestions but this mostly lgtm

src/gcp/cloudsql/permissions.ts
tammam-g: Ok, removed login.

src/dataconnect/schemaMigration.ts
```diff
@@ -202,20 +205,65 @@ async function handleIncompatibleSchemaError(args: {
...
-        username: iamUser,
-      },
-    );
-    return incompatibleSchemaError.diffs;
+        firebaseowner(databaseId),
+        (await getIAMUser(options)).user,
+      ))
+    ) {
+      throw new FirebaseError(
+        `Command aborted. Only users with role firebaseowner role can run migrations.`,
```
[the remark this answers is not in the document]
tammam-g: Updated the message to specify it's sql role that we are checking, do you think this is better?

src/gcp/cloudsql/permissions.ts
```diff
@@ -0,0 +1,237 @@
...
+  return `firebasereader_${databaseId}_public`;
+}
+
+export function firebasewriter(databaseId: string) {
+  return `firebasewriter_${databaseId}_public`;
+}
+
+// Returns true if "grantedRole" is granted to "granteeRole" and false otherwise.
+// Throw an error if commands fails due to another reason like connection issues.
+export async function checkRoleIsGranted(
```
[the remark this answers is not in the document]
tammam-g: Done

src/gcp/cloudsql/permissions.ts
tammam-g: I'm tempted to leave this as is since many of our existing functions take instanceId, databaseId.

PR #7589, opened 2024-08-23
update data connect client, dataplaneclient, and emulator client to v…

hlshen:
Do not release until emulator updates endpoints


joehan (changes requested):
Tests in dataconnect/client.spec.ts are failing - looks like you just need to update the references to v1alpha in there as well.

joehan (approved):

joehan (approved):

PR #7600, opened 2024-08-27
temporarily remove extensions deploy from functions

ifielker:
Temporarily remove extensions deploy from functions.
This is causing issues with unexpected deletion of extensions during functions deploy.


joehan (approved):
Maybe add a comment adding some extra context here

PR #7611, opened 2024-08-30
Update message when users try to update from 1st to 2nd Gen function

taeold:
CLI has always block the transition 1st Gen to 2nd Gen in the CLI - change here simply update the product name and point users to our public doc that documents changes necessary to transition to 2nd gen function.

Aside: Cloud Run functions doesn't allow for 1st and 2nd gen functions to share the same name.


src/deploy/functions/release/planner.ts
```diff
@@ -316,9 +316,8 @@ export function checkForIllegalUpdate(want: backend.Endpoint, have: backend.Endp
 export function checkForV2Upgrade(want: backend.Endpoint, have: backend.Endpoint): void {
   if (want.platform === "gcfv2" && have.platform === "gcfv1") {
     throw new FirebaseError(
-      `[${getFunctionLabel(
-        have,
-      )}] Upgrading from GCFv1 to GCFv2 is not yet supported. Please delete your old function or wait for this feature to be ready.`,
+      `[${getFunctionLabel(have)}] Upgrading from 1st Gen to 2nd Gen is not yet supported. ` +
+        "Please review https://firebase.google.com/docs/functions/2nd-gen-upgrade before migrating to 2nd Gen.",
```
egilmorez, line 320: I always feel slightly rude making this comment, but our developer style guide advises against "please."  I'd suggest the simple and direct "See <link>."

src/deploy/functions/release/planner.ts
```diff
@@ -316,9 +316,8 @@ export function checkForIllegalUpdate(want: backend.Endpoint, have: backend.Endp
 export function checkForV2Upgrade(want: backend.Endpoint, have: backend.Endpoint): void {
   if (want.platform === "gcfv2" && have.platform === "gcfv1") {
     throw new FirebaseError(
-      `[${getFunctionLabel(
-        have,
-      )}] Upgrading from GCFv1 to GCFv2 is not yet supported. Please delete your old function or wait for this feature to be ready.`,
+      `[${getFunctionLabel(have)}] Upgrading from 1st Gen to 2nd Gen is not yet supported. ` +
```
egilmorez, line 319: When do we actually display this?  On deployment of a scheduled function I guess?

milyes (approved):
https://github.com/firebase/firebase-tools/blob/0d5f7492511a141eb235e27fe5d0a160a61af4f5/schema/firebase-config.json

src/deploy/functions/release/planner.ts
taeold: On `firebase deploy` when we detect that there exists 1st Gen function whose name is same as a 2nd Gen function you have in your source.

joehan (approved):

PR #7615, opened 2024-08-31
DataConnect + PGLite prototype.

joehan:
### Description
Got a working prototype of the DataConnect emualtor backed by PGLite. If you want to use this instead of spinning up your own Postgres, run `firebase experiments:enable fdcpglite` and the start up the emulator (make sure port 5432 is free!)

Note: Previously, this had some unshippable changes - those have been fixed after the latest commits.

### Scenarios Tested
Start up the DataConnect emulator with no Postgres server running beforehand, and then watch it successfully spin up a PGLite database, force migrate it to a simple schema, and then curl the dataplane. I tested this both on a gLinux cloudtop and on my m1 Mac.

### TODO Testing

- [ ] Test on Windows
- [ ] Test on Intel mac
- [ ] Test on IDX


src/emulator/dataconnect/pg-gateway/auth/cert.ts
```diff
@@ -0,0 +1,100 @@
...
+    // TODO: get peer cert and validate through hook
+    const isValid = false;
+
+    // const isValid = await this.auth.validateCredentials(
+    //   {
+    //     username: this.username,
+    //     certificate:  this.socket.getPeerCertificate(),
+    //   },
+    //   this.connectionState,
+    // );
```
rosalyntan, line 78: Does this need to be handled? (not sure if this is directly copied from `pg-gateway` or not)

src/emulator/dataconnect/pgliteServer.ts
```diff
@@ -0,0 +1,119 @@
...
+      });
+
+      const extendedQueryPatch: PGliteExtendedQueryPatch = new PGliteExtendedQueryPatch(connection);
+
+      socket.on("end", () => {
+        logger.debug("Postgres client disconnected");
+      });
+    });
+    const listeningPromise = new Promise<void>((resolve) => {
+      server.listen(5432, "127.0.0.1", () => {
```
rosalyntan, line 52: Is this always the port, or is this something configurable?

src/emulator/dataconnect/pgliteServer.ts
```diff
@@ -0,0 +1,119 @@
...
+  async getDb(): Promise<PGlite> {
+    if (this.db) {
+      return this.db;
+    }
+    // Not all schemas will need vector installed, but we don't have an good way
+    // to swap extensions after starting PGLite, so we always include it.
+    const vector = (await dynamicImport("@electric-sql/pglite/vector")).vector;
+    const uuidOssp = (await dynamicImport("@electric-sql/pglite/contrib/uuid_ossp")).uuid_ossp;
+    return PGlite.create({
+      // dataDir?: string;
```
rosalyntan, line 71: nit: remove the commented params?

src/emulator/dataconnect/pgliteServer.ts
```diff
@@ -0,0 +1,119 @@
...
+  getMessages,
+  type PostgresConnection,
+  FrontendMessageCode,
+  BackendMessageCode,
+} from "./pg-gateway/index";
+import { fromNodeSocket } from "./pg-gateway/platforms/node";
+import { logger } from "../../logger";
+
+export class PostgresServer {
+  public callCount = 0;
```
rosalyntan, line 20: Does this get used anywhere?

src/emulator/dataconnect/pg-gateway/auth/cert.ts
joehan: No need for us to actually authenticate the toolkit's creds - PGlite runs in single user mode, so we are using 'cleartext' auth and will never hit this code. I removed this while debugging some import issues that arose from converting these files to cjs.

src/emulator/dataconnect/pgliteServer.ts
joehan: Ahh, good catch, we should make this configurable. I'll put it in firebase.json#emulators.dataconnect.postgresPort/Host and use these as the defaults

src/emulator/dataconnect/pgliteServer.ts
```diff
@@ -0,0 +1,120 @@
...
+    this.database = database;
+  }
+}
+
+export class PGliteExtendedQueryPatch {
+  isExtendedQuery = false;
+
+  constructor(public connection: PostgresConnection) {}
+
+  async *filterResponse(message: Uint8Array, response: Uint8Array) {
```
joehan, line 97: Note from @hlshen - add reference to the pglite bug and a todo to remove this.

src/emulator/dataconnectEmulator.ts
```diff
@@ -92,7 +112,11 @@ export class DataConnectEmulator implements EmulatorInstance {
       this.usingExistingEmulator = false;
     }
     if (!isVSCodeExtension()) {
-      await this.connectToPostgres();
+      await this.connectToPostgres(
```
joehan, line 115: Probably move this to this.args.autostartPostgres block

src/emulator/dataconnectEmulator.ts
```diff
@@ -25,6 +26,9 @@ export interface DataConnectEmulatorArgs {
   auto_download?: boolean;
   rc: RC;
   config: Config;
+  autostartPostgres: boolean;
```
joehan, line 29: Note: let's switch this this autoconnectToPostgres, so that it meshes with the DataConnectToolkitRegistry harold is working on.

hlshen (approved):
Approved with plans to get rid of pg-gateway folder

PR #7620, opened 2024-09-02
Improve error message thrown when secret dataFile path does not exist

aalej:
### Description

`firebase functions:secrets:set SECRET_API_KEY --data-file ./non-existent.json` throws 
```
``` 

when file does not exist

### Scenarios Tested

`firebase functions:secrets:set SECRET_API_KEY --data-file ./secrets/test.jso` throws
```
Error: File not found: ./secrets/test.jso
```

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -915,5 +915,12 @@ export function readSecretValue(prompt: string, dataFile?: string): Promise<stri
   if (dataFile && dataFile !== "-") {
     input = dataFile;
   }
-  return Promise.resolve(fs.readFileSync(input, "utf-8"));
+  try {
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -915,5 +915,12 @@ export function readSecretValue(prompt: string, dataFile?: string): Promise<stri
   if (dataFile && dataFile !== "-") {
     input = dataFile;
   }
-  return Promise.resolve(fs.readFileSync(input, "utf-8"));
+  try {
+    return Promise.resolve(fs.readFileSync(input, "utf-8"));
```
joehan, line 919: Unrelated to your change, but this feels like it should be an async function and just return fs.readFileSync();

PR #7625, opened 2024-09-04
Skipping partial html files generated due to enabling ppr in Nextjs

chalosalvador:
### Description

When partial pre-rendering (`experimental: {ppr: true}` is enabled in `next.config.js` Next.js includes dynamic pages in the `prerender-manifest.json` and it generates the `.html` files with partial html and those files are copied over by firebase to the hosting folder causing the pages to render empty or invalid html pages.

This PR add a check to skip partial html files from being copies over to the hosting folder.

### Scenarios Tested
- Enable ppr in next.config.js
- Deploy to firebase hosting or run emulators in production mode.
- Partial html files are skipped and the pages are SSR'd


src/frameworks/next/index.ts
```diff
@@ -536,6 +536,15 @@ export async function ɵcodegenPublicDirectory(
       let defaultDestPath = isDefaultLocale && join(destDir, basePath, ...destPartsOrIndex);
       if (!fileExistsSync(sourcePath) && fileExistsSync(`${sourcePath}.html`)) {
         sourcePath += ".html";
+        // Check if the HTML file is partial (doesn't have closing </html> tag)
+        const content = await readFile(sourcePath, "utf8");
+        if (!content.includes("</html>")) {
```
jamesdaniels, line 541: can we do a similar pass adding these files to `reasonsForBackend`

chalosalvador: @jamesdaniels I added this and moved some common logic to utils functions and added some tests as well.

src/frameworks/next/utils.ts
```diff
@@ -549,3 +550,59 @@ export function installEsbuild(version: string): void {
...
+      };
+    }
+  }
+  return pagesManifestLikePrerender;
+}
+
+/**
+ * Check if an HTML file is partial (doesn't have closing </html> tag)
+ */
+export async function isPartialHTML(filePath: string): Promise<boolean> {
```
jamesdaniels, line 578: q, in my test project I see a *.prefetch.rsc file next to PPR content in the dist dir, is that a suitable test that wouldn't require reading the file?

chalosalvador: The problem with using the `*.prefetch.rsc` file is that it is present even for files that are non-partial html. If we use that as a test then we would be skipping valid (fully generated) html files.

src/frameworks/next/index.ts
```diff
@@ -290,6 +294,24 @@ export async function build(
...
+
+    for (const [path, route] of Object.entries(allRoutes)) {
+      const appPathRoute = getAppPathRoute(route, appPathRoutesManifest);
+      const contentDist = getContentDist(dir, distDir, appPathRoute);
+      const sourceParts = path.split("/").filter((it) => !!it);
+      const sourcePartsOrIndex = sourceParts.length > 0 ? sourceParts : ["index"];
+
+      const sourcePath = join(contentDist, ...sourcePartsOrIndex);
+
+      if (await isPartialHTML(`${sourcePath}.html`)) {
```
jamesdaniels, line 311: Let's test if .html exists first, that should reduce some of the cost

src/frameworks/next/index.ts
```diff
@@ -290,6 +294,24 @@ export async function build(
...
+        reasonsForBackend.add(`partial HTML file ${path}`);
+      }
+    }
```
leoortizz, line 314: we Promise.all this
```suggestion
    await Promise.all(
      Object.entries(allRoutes).map(async ([path, route]) => {
        const appPathRoute = getAppPathRoute(route, appPathRoutesManifest);
        const contentDist = getContentDist(dir, distDir, appPathRoute);
        const sourceParts = path.split("/").filter((it) => !!it);
        const sourcePartsOrIndex = sourceParts.length > 0 ? sourceParts : ["index"];

        const sourcePath = join(contentDist, ...sourcePartsOrIndex);

        if (await isPartialHTML(`${sourcePath}.html`)) {
          reasonsForBackend.add(`partial HTML file ${path}`);
        }
      }),
    );
```

src/frameworks/next/utils.ts
jamesdaniels: So digging in a bit on an experimental project, it seems as though the best test is the "postponed" key in the `.meta` file, we're already reading that for headers. Mind refactoring to use the presence of that key, rather than reading the HTML? That will also get us PPR for data routes, etc.

jamesdaniels (changes requested):

src/frameworks/next/utils.ts
chalosalvador: Nice! Sorry for not noticing that property before. Just refactored it to use that property from the .meta files.

jamesdaniels (approved):

CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- Fixed issue with ppr routes not rendering correctly.
```
leoortizz, line 1: please always link the PR in the changelog
```suggestion
- Fixed issue with ppr routes not rendering correctly. (#7625)
```

CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- Fixed issue with ppr routes not rendering correctly. (#7625)
```
leoortizz, line 1: I think we should also always mention the framework
```suggestion
- Fixed Next.js issue with PPR routes not rendering correctly. (#7625)
```

leoortizz (approved):

PR #7635, opened 2024-09-10
update cli support for dataconnect toolkit and webhooks

hlshen:


src/emulator/dataconnectToolkitController.ts
```diff
@@ -0,0 +1,50 @@
+import { EmulatorInstance, EmulatorInfo } from "./types";
+import { FirebaseError } from "../error";
+import * as portUtils from "./portUtils";
+import * as express from "express";
```
joehan, line 4: Remove

src/emulator/hub.ts
```diff
@@ -50,7 +51,8 @@ export class EmulatorHub extends ExpressBasedEmulator {
     const data = fs.readFileSync(locatorPath, "utf8").toString();
     const locator = JSON.parse(data) as Locator;
 
-    if (locator.version !== this.CLI_VERSION) {
+    // TODO: In case the locator file format is change, handle issues with format incompatability
```
joehan, line 54:
```suggestion
    // TODO: In case the locator file format is changed, handle issues with format incompatibility
```

joehan (commented):
Went through this together in person, and its very close to LGTM. I'd love to get some of the low hanging fruit TODOs in tho

src/emulator/commandUtils.ts
```diff
@@ -464,9 +465,11 @@ export async function emulatorExec(script: string, options: any): Promise<void>
   try {
     const showUI = !!options.ui;
     ({ deprecationNotices } = await controller.startAll(options, showUI, true));
+    sendVSCodeMessage({ message: VSCODE_MESSAGE.EMULATORS_STARTED });
     exitCode = await runScript(script, extraEnv);
     await controller.onExit(options);
   } finally {
+    sendVSCodeMessage({ message: VSCODE_MESSAGE.EMULATORS_START_ERRORED });
```
rrousselGit, line 472: Wouldn't having this in the `finally` block _always_ notify the emulator as errored?

joehan: +1 - this should be in a new catch, and there should be a separate finally block for controller.cleanShutdown()

joehan (approved):
LGTM after the requested fixes

src/emulator/commandUtils.ts
hlshen: Oh I just didn't read that at all. good finally, I mean catch

joehan (approved):

PR #7652, opened 2024-09-12
Removing LocalConnectionString

joehan:
### Description
In the new wonderful world of pglite, localconnectionstring is confusing and not very useful for 95% of users.
This PR:

- Removes LocalConnectionString from .firebaserc
- Keeps the FIREBASE_DATACONNECT_POSTGRESQL_STRING env var override for the rare cases we do want this.
-  Gets rid of these questions from init flows.

### DIFF
--- a/src/emulator/dataconnectEmulator.ts
+++ b/src/emulator/dataconnectEmulator.ts
@@ -90,8 +90,14 @@ export class DataConnectEmulator implements EmulatorInstance {
       enable_output_generated_sdk: this.args.enable_output_generated_sdk,
     });
     this.usingExistingEmulator = false;
-    if (this.args.autoconnectToPostgres) {
-      // TODO: Skip starting our own PG server if localConnectString is set
+    if (dataConnectLocalConnString()) {
+      this.logger.logLabeled(
+        "INFO",
+        "Data Connect",
+        `FIREBASE_DATACONNECT_POSTGRESQL_STRING is set to ${dataConnectLocalConnString()}`,
+      );
+    }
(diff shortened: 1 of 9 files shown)


src/emulator/dataconnectEmulator.ts
```diff
@@ -90,8 +89,14 @@ export class DataConnectEmulator implements EmulatorInstance {
...
-    if (this.args.autoconnectToPostgres) {
-      // TODO: Skip starting our own PG server if localConnectString is set
+    if (dataConnectLocalConnString()) {
+      this.logger.logLabeled(
+        "INFO",
+        "Data Connect",
+        `FIREBASE_DATACONNECT_POSTGRESQL_STRING is set to ${dataConnectLocalConnString()} - using that instead of starting a new database`,
+      );
+    }
+    if (this.args.autoconnectToPostgres && !dataConnectLocalConnString()) {
```
hlshen, line 99: Here we check if dataConnectLocalConnString doesn't exist, then on line 114 we use it if the experiment isn't enabled.

Should the flow on line 92 call connectToPostgres()?

hlshen (changes requested):

src/emulator/dataconnectEmulator.ts
joehan: Good catch - ended up refactoring this slightly to make more sense. Now:
- the experiment controls wether we start pglite
- Autoconnect just controls whether we connect right away
- if you provide an override in the env var, we still connect to it if appropriate

hlshen (approved):

rosalyntan (approved):

PR #7655, opened 2024-09-13
Update the Firestore init() flow to handle non-default databases

trippedout:
### Description

Update the Firestore init() flow to handle non-default databases, by allowing the user to input the name of the database they want to initialize. 

This is a catch for older projects created in Cloud, which have a (default) non-Firestore Native database in their projects before they add Firebase. New projects shouldn't be affected by this change.

The only other thought I had was maybe it would be better to have input of database name during init *by default* but defaulting to (default) with option to enter your own without first hitting a 404. 

### Scenarios Tested

Projects created with older (default) database, deleting that and creating a new Firestore database with a different name.

### Sample Commands

Nothing new, just additional ask for database name if (default) doesnt exist, which would fail out without reason previously.


joehan (approved):
Thanks a ton for implementing this!

PR #7657, opened 2024-09-13
Add View Docs button

joehan:
### Description
Adding a button to better expose the cool new docs feature:
Screenshot 2024-09-13 at 1 34 50 PM
Clicking it opens the docs in your default browser. We could move this into a Webview at some point, but this seems simpler for now.

### Scenarios Tested
Click the button, view the docs.


hlshen:
Only question is do we want this in the explorer panel as a header button?

hlshen (approved):

PR #7659, opened 2024-09-16
Remove authMode

fredzqm:
Tested with existing `connector.yaml`. Works fine with any command.

I think the json schema is only used for VS Code linter.


joehan (approved):
Yep, you're correct about the JSON schema only being used by vscode linter plugins

PR #7662, opened 2024-09-16
Add firebase-path option for vscode

tjlav5:
Allow folks to specify a specific firebase binary

### DIFF
--- a/firebase-vscode/package.nls.json
+++ b/firebase-vscode/package.nls.json
@@ -3,7 +3,8 @@
   "ext.config.dataConnect.skipToAppFolderSelect": "Skip app folder selection prompt. If false (default), trying to generate SDK config will open a confirmation modal.",
   "ext.config.debug": "Enable writing debug-level messages to the file provided in `#firebase.debugLogPath#` (requires restart)",
   "ext.config.debugLogPath": "If `#firebase.debug#` is `true`, Append debug-level messages to the provided file (requires restart)",
+  "ext.config.firebasePath": "Path to the `Firebase` module, e.g. `./node_modules/firebase`",
   "ext.config.hosting.useFrameworks": "Enable web frameworks",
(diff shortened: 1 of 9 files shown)


firebase-vscode/package.nls.json
```diff
@@ -3,7 +3,8 @@
   "ext.config.dataConnect.skipToAppFolderSelect": "Skip app folder selection prompt. If false (default), trying to generate SDK config will open a confirmation modal.",
   "ext.config.debug": "Enable writing debug-level messages to the file provided in `#firebase.debugLogPath#` (requires restart)",
   "ext.config.debugLogPath": "If `#firebase.debug#` is `true`, Append debug-level messages to the provided file (requires restart)",
+  "ext.config.firebasePath": "Path to the `Firebase` module, e.g. `./node_modules/firebase`",
```
hlshen, line 6: should it be "firebase-tools"?

hlshen (changes requested):

firebase-vscode/package.nls.json
tjlav5: ehh Firebase-tools is more-so the repo, and Firebase is the module (the built binary)

firebase-vscode/src/core/webhook.ts
```diff
@@ -9,7 +9,7 @@ export function registerWebhooks() {
   app.use(bodyParser.json()); // for parsing application/json
 
   const server = createServer(app);
-  const port = 40000; // TODO: generate port dynamically with this as default, and pass as env var to integrated terminal
+  const port = 40001; // TODO: generate port dynamically with this as default, and pass as env var to integrated terminal
```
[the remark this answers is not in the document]
tjlav5: Good catch, added

hlshen (approved):

PR #7667, opened 2024-09-17
Round 1 of bug bash fixes

joehan:
### Description
A few small fixes for issues we found during the bug bash. Specifically:
- Get rid of the last remnants of the pglite experiment flag.
- Fixed a mismatch between VSCode engine and VSCode types.
- Mention that you need to quote keys in the variables input. Kept it light here in case we want to replace this component wholesale.
- Clearer error messages when an emulators port is taken, including the number of the process blocking that port.
- Add --yes to the npx command.


src/emulator/portUtils.ts
```diff
@@ -495,3 +495,13 @@ export function listenSpecsToString(specs: ListenSpec[]): string {
     })
     .join(",");
 }
+
+export function findProcessBlockingPort(port: number): number | undefined {
+  try {
+    const res = execSync(`lsof -t -i :${port} `, {});
```
hlshen, line 501: sanity checking that this is our best option to check port

src/emulator/dataconnect/pgliteServer.ts
```diff
@@ -51,13 +53,22 @@ export class PostgresServer {
       });
     });
     const listeningPromise = new Promise<void>((resolve) => {
+      server.on("error", (e: any) => {
```
hlshen, line 56: Do we want to allow users to run both psql on their machine and PGLite?

Maybe the logs should suggest users to change their port instead

joehan: Good call, forgot i made this configurable - reworded slightly to suggest changing this.

src/emulator/portUtils.ts
joehan: Hmmm, we should probably use https://github.com/marionebl/lsofi instead to make sure this works on windows too. Will swap it over

PR #1014, opened 2018-11-14
Support v1beta2 indexes API

samtstern:
Enables #889 

The new indexes API better supports:

  * Collection group indexes
  * Array indexes
  * Future index types (more flexible)
  * Single-field index controls

Example of pretty-printing index config with some field exemptions:
image

PR #7674, opened 2024-09-17
Small improvements to FDC init and JSON Schema. #1014

mbleigh:
1. Fixes the JSON schema for connectors to allow either list or single for generators, and adds llmTools support.
1. Changes default service id (no my-).
1. Elides "this will take a few minutes" when not creating an instance.


joehan (approved):

joehan:
Dataconnect Deploy tests were broken on master and will be fixed by https://github.com/firebase/firebase-tools/pull/7678 - no need to worry about the failure

PR #7678, opened 2024-09-18
Switch primaryDatasource -> datasources

joehan:
### Description
Switching primaryDatasource -> datasources, as part of the move to v1beta apis.

### DIFF
--- a/src/deploy/dataconnect/deploy.ts
+++ b/src/deploy/dataconnect/deploy.ts
@@ -88,10 +88,12 @@ export default async function (
         return !filters || filters?.some((f) => si.dataConnectYaml.serviceId === f.serviceId);
       })
       .map(async (s) => {
-        const primaryDatasource = s.schema.datasources.find((d) => d.postgresql);
-        if (primaryDatasource) {
-          const instanceId = primaryDatasource.postgresql?.cloudSql.instance.split("/").pop();
-          const databaseId = primaryDatasource.postgresql?.database;
+        const postgresDatasource = s.schema.datasources.find(d => d.postgresql);
+        if (postgresDatasource) {
+          const instanceId = postgresDatasource.postgresql?.cloudSql.instance
+            .split("/")
+            .pop();
+          const databaseId = postgresDatasource.postgresql?.database;
           if (!instanceId || !databaseId) {
             return Promise.resolve();
           }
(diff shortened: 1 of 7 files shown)


hlshen (approved):
super nit: should we make it a helper func? I know it's one line of code but could be `findPrimaryDataSource(schema).postgresql....`

LGTM either way

src/init/features/dataconnect/index.ts
```diff
@@ -262,16 +262,17 @@ async function promptForService(
         info.serviceId = serviceName.serviceId;
         info.locationId = serviceName.location;
         if (choice.schema) {
-          if (choice.schema.primaryDatasource.postgresql?.cloudSql.instance) {
+          const primaryDatasource = choice.schema.datasources.find((d) => d.postgresql);
```
rosalyntan, line 264: nit: Can multiple datasources specify `postgresql`? Also given that the backend resource doesn't specify a "primary" datasource anymore -- is the "primary" datasource intended to just be the first in the list?

Maybe for now we can just pull the first datasource for now and verify it has a `postgresql` field?

src/deploy/dataconnect/deploy.ts
```diff
@@ -88,22 +88,23 @@ export default async function (
         return !filters || filters?.some((f) => si.dataConnectYaml.serviceId === f.serviceId);
       })
       .map(async (s) => {
-        const instanceId = s.schema.primaryDatasource.postgresql?.cloudSql.instance
-          .split("/")
-          .pop();
-        const databaseId = s.schema.primaryDatasource.postgresql?.database;
-        if (!instanceId || !databaseId) {
-          return Promise.resolve();
+        const primaryDatasource = s.schema.datasources.find((d) => d.postgresql);
```
rosalyntan, line 91: nit: For future-proofing (since the "primary" datasource won't necessarily be the Postgres one) would it make more sense to name this `postgresDatasource`?

rosalyntan (approved):

src/init/features/dataconnect/index.ts
joehan: This is definitely a short sighted solution on my end, because we will definitely need a pretty major overhaul of the init code once we support multiple datasources.

PR #7691, opened 2024-09-19
Better handling for port conflicts and errors in Data Connect emulator

joehan:
### Description
Better handling for port conflicts and errors in Data Connect emulator. Also cleaned up more remaining logic to pick up running data connect emulators.

### Scenarios Tested
Screenshot 2024-09-19 at 3 51 57 PM
Screenshot 2024-09-19 at 3 31 24 PM

### DIFF
--- a/src/emulator/dataconnectEmulator.ts
+++ b/src/emulator/dataconnectEmulator.ts
@@ -99,15 +99,24 @@ export class DataConnectEmulator implements EmulatorInstance {
       } else {
         connStr = `postgres://${this.args.postgresHost ?? "127.0.0.1"}:${this.args.postgresPort ?? 5432}/${dbId}?sslmode=disable`;
         const pgServer = new PostgresServer(dbId, "postgres");
-        const server = await pgServer.createPGServer(
-          this.args.postgresHost,
-          this.args.postgresPort,
-        );
-        this.logger.logLabeled(
-          "INFO",
-          "Data Connect",
-          `Started up Postgres server, listening on ${server.address()?.toString()}`,
-        );
+        try {
+          const server = await pgServer.createPGServer(
+            this.args.postgresHost,
+            this.args.postgresPort,
+          );
+          server.on("error", (err) => {
+            this.logger.logLabeled("ERROR", "Data Connect", `Postgres threw an unexpected error, shutting down the Data Connect emulator: ${err}`);
+            this.stop();
+          });
+          this.logger.logLabeled(
+            "INFO",
+            "Data Connect",
+            `Started up Postgres server, listening on ${server.address()?.toString()}`,
+          );
+        } catch(err: any) {
+            this.logger.logLabeled("ERROR", "Data Connect", `Postgres threw an unexpected error, shutting down the Data Connect emulator: ${err}`);
+            this.stop();
+        }
       }
       await this.connectToPostgres(connStr, dbId, serviceId);
     }
--- a/src/emulator/dataconnectEmulator.ts
+++ b/src/emulator/dataconnectEmulator.ts
@@ -1,5 +1,6 @@
 import * as childProcess from "child_process";
 import { EventEmitter } from "events";
+const lsofi = require("lsofi");
 
 import { dataConnectLocalConnString } from "../api";
 import { Constants } from "./constants";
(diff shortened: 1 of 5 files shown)


src/emulator/downloadableEmulators.ts
```diff
@@ -500,10 +500,15 @@ async function _runBinary(
...
+      if (data.toString().includes("address already in use")) {
+        const message = `${description} has exited because its configured port ${command.port} is already in use. Are you running another copy of the emulator suite?`;
+        logger.logLabeled("ERROR", emulator.name, message);
+      }
     });
 
-    emulator.instance.on("error", (err) => {
-      handleEmulatorProcessError(emulator.name, err, command.port);
+    emulator.instance.on("error", async (err) => {
+      await handleEmulatorProcessError(emulator.name, err, command.port);
```
yuchenshi, line 511: I don't think this does anything. Event listener callbacks are not monitoring Promise returns and a rejection or a completion won't have visible effects.

To communicate this better, you'd probably want the following instead (this promise is not being handled):

```suggestion
      void handleEmulatorProcessError(emulator.name, err, command.port);
```

src/emulator/storage/rules/runtime.ts
```diff
@@ -167,8 +167,8 @@ export class StorageRulesRuntime {
     });
 
     // This catches error when spawning the java process
-    this._childprocess.on("error", (err) => {
-      handleEmulatorProcessError(Emulators.STORAGE, err);
+    this._childprocess.on("error", async (err) => {
+      await handleEmulatorProcessError(Emulators.STORAGE, err);
```
yuchenshi, line 171: ditto

src/emulator/downloadableEmulators.ts
```diff
@@ -500,10 +500,15 @@ async function _runBinary(
...
+
+      if (data.toString().includes("address already in use")) {
+        const message = `${description} has exited because its configured port ${command.port} is already in use. Are you running another copy of the emulator suite?`;
+        logger.logLabeled("ERROR", emulator.name, message);
+      }
     });
 
-    emulator.instance.on("error", (err) => {
-      handleEmulatorProcessError(emulator.name, err, command.port);
+    emulator.instance.on("error", async (err: any) => {
```
yuchenshi, line 510:
```suggestion
    emulator.instance.on("error", (err: any) => {
```

src/emulator/storage/rules/runtime.ts
```diff
@@ -167,8 +167,8 @@ export class StorageRulesRuntime {
     });
 
     // This catches error when spawning the java process
-    this._childprocess.on("error", (err) => {
-      handleEmulatorProcessError(Emulators.STORAGE, err);
+    this._childprocess.on("error", async (err: any) => {
```
yuchenshi, line 170:
```suggestion
    this._childprocess.on("error", (err: any) => {
```

yuchenshi (changes requested):
Have we considered automatically picking a pgLite port? I don't feel like requiring an exact port by default is that useful

joehan:
> Have we considered automatically picking a pgLite port? I don't feel like requiring an exact port by default is that useful

That's a good point. Switched this to be more similar to the standard behavior of emulators (find a port unless one is set in firebase.json)

src/emulator/dataconnectEmulator.ts
```diff
@@ -97,12 +98,31 @@ export class DataConnectEmulator implements EmulatorInstance {
...
+        if (this.args.postgresPort) {
+          const process = await lsofi(this.args.postgresPort);
+          if (process) {
+            const errMessage =
+              `Data Connect: Unable to start PGLite server on port ${this.args.postgresPort} because it is already in use by process number ${process}. This may occur if you are running another instance of Postgres.` +
+              ` You can choose a different port by setting 'firebase.json#emulators.dataconnect.postgresPort', or you can unset that field to automatically find an open port.`;
+            throw new FirebaseError(errMessage);
+          }
+        }
+        const port = this.args.postgresPort || (await findOpenPort(5432));
```
yuchenshi, line 111: This should follow Firestore emulator websocket's example to find a port early and avoid conflicts: https://github.com/firebase/firebase-tools/blob/0314586ffffcfe286caacb067465c1817286c806/src/emulator/controller.ts#L398

I'm approving to unblock but please clean this up as a follow up. Please also remove `findOpenPort` -- it only checks `127.0.0.1`, not `postgresHost`, while the `resolveHostAndAssignPorts` as linked above handles all these edge cases

yuchenshi (approved):

hlshen (approved):

src/emulator/dataconnectEmulator.ts
yuchenshi: Addressed in #7727 (no pun intended)

PR #7693, opened 2024-09-20
Include SDK generation as part of `firebase init dataconnect`

rosalyntan:
### Description

Include `firebase init dataconnect:sdk` functionality as part of `firebase init dataconnect`.

Also updated the app directory prompt: "Where is your app directory? Leave blank to set up a generated SDK in your current directory."

### Scenarios Tested

- [x] `firebase init dataconnect` in a folder with `build.gradle`
- [x] `firebase init dataconnect` in a folder without platform indicators, (y) when prompted for gen SDKs
- [x] `firebase init dataconnect` in a folder without platform indicators, (n) when prompted for gen SDKs

### Sample Commands

`firebase init dataconnect` in a folder with `build.gradle`:
Screenshot 2024-09-23 at 12 40 54 PM


src/init/features/dataconnect/index.ts
```diff
@@ -65,17 +66,27 @@ const defaultConnector = {
 
 // doSetup is split into 2 phases - ask questions and then actuate files and API calls based on those answers.
 export async function doSetup(setup: Setup, config: Config): Promise<void> {
-  const info = await askQuestions(setup, config);
+  const info = await askQuestions(setup);
   await actuate(setup, config, info);
+
+  const promptForSDKGeneration = await confirm({
+    message: `Would you like to configure generated SDKs now?`,
+    default: true,
```
fredzqm, line 74: Default to true or false?

I think we can run a quick framework detection, if something is detected, go straight to SDK setup w/o prompting.
If nothing is detected, then ask if they want to setup SDK and default to false.

https://github.com/firebase/firebase-tools/blob/master/src/init/features/dataconnect/sdk.ts#L70-L71

fredzqm: Existing code, but I think "which connector" question should be asked after picking the framework.

https://github.com/firebase/firebase-tools/blob/master/src/init/features/dataconnect/sdk.ts#L39-L65

rosalyntan: Updated to run framework detection and proceed with gen SDK setup w/o prompting when a framework is detected.

I don't think we can simplify by including "skip" as an option in the framework question, as if a framework isn't detected we prompt for an app directory first before asking the framework question, which results in weird UX if we make them enter an app directory before they can skip out of the gen SDK flow.

fredzqm: Works for me. Thanks!

fredzqm (approved):

PR #7694, opened 2024-09-20
Set default state for sidebar. Update fallback for emulator start

hlshen:
Screenshot 2024-09-20 at 12 32 28 AM

### DIFF
--- a/firebase-vscode/src/core/emulators.ts
+++ b/firebase-vscode/src/core/emulators.ts
@@ -92,7 +92,7 @@ export class EmulatorsController implements Disposable {
         // notify UI to show reset
         this.broker.send("notifyEmulatorsHanging", true);
       }
-    }, 10000); // default 15 seconds spin up time
+    }, 10000); // default 10 seconds spin up time
   }
 
   public setEmulatorsStopping() {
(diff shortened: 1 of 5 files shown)


firebase-vscode/src/core/emulators.ts
```diff
@@ -82,17 +82,14 @@ export class EmulatorsController implements Disposable {
...
-        await this.findRunningCliEmulators();
-        if (this.emulators.status === "stopped") {
-          vscode.window.showErrorMessage("Emulators failed to start.");
-        }
+      if (this.emulators.status === "starting") {
+        // notify UI to show reset
+        this.broker.send("notifyEmulatorsHanging", true);
       }
-    }, 15000); // default 15 seconds spin up time
+    }, 10000); // default 15 seconds spin up time
```
joehan, line 92: Nit: update comment to reflect the new timeout

firebase-vscode/src/core/emulators.ts
```diff
@@ -82,17 +82,14 @@ export class EmulatorsController implements Disposable {
     this.emulators.status = "starting";
     this.notifyEmulatorStateChanged();
 
-
+    // TODO: handle rare cases where user cancels on their own, and starts again withing 10 seconds
```
joehan, line 85: Probably just need to save the return value of setTimeou[clearTimeout()](https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout) when we shut down.

joehan (approved):
I think we can fix the TODO pretty easily, but this LGTM otherwise.

PR #7703, opened 2024-09-20
Merge Public Preview launch branch into master

hlshen:


firebase-vscode/src/core/config.ts
```diff
@@ -95,7 +91,7 @@ function notifyFirebaseConfig(broker: ExtensionBrokerImpl) {
     >(
       (value) => ({ value: value?.data, error: undefined }),
       (error) => ({ value: undefined, error: `${error}` }),
-    ),
+    ) as any,
```
joehan, line 94: @hlshen The typing in this file is pretty shoddy, and we really should not use `as any` to paper it over like this.  I'm gonna send a follow up PR to touch it up.

joehan: Done!

joehan (approved):

PR #7707, opened 2024-09-21
When no project is detected, add an "open folder" button.

rrousselGit:
### Description

The new button relies on the VSCode's built-in open command.

### DIFF
--- a/firebase-vscode/src/core/index.ts
+++ b/firebase-vscode/src/core/index.ts
@@ -12,6 +12,7 @@ import { registerQuickstart } from "./quickstart";
 import { registerOptions } from "../options";
 import { upsertFile } from "../data-connect/file-utils";
 import { registerWebhooks } from "./webhook";
+import { setupExecuteCommandMockable } from "../utils/test_hooks";
 
 export async function registerCore(
   broker: ExtensionBrokerImpl,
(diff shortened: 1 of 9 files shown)


firebase-vscode/src/core/index.ts
```diff
@@ -39,6 +40,14 @@ export async function registerCore(
     vscode.env.openExternal(vscode.Uri.parse(href));
   });
 
+  const executeCommand = setupExecuteCommandMockable(context);
+
+  context.subscriptions.push({
```
hlshen, line 45: assign to value and add to Disposable at bottom of file

hlshen (changes requested):
Approved besides small cleanup.

firebase-vscode/src/core/index.ts
rrousselGit: Why is this preferable?

hlshen: Instead of writing docs, just having the named variable could really help readability :)

rrousselGit: What doc? :)

This is IMO more maintainable to the old way of returning a Disposable.
It makes it much more obvious what's disposed and what isn't. And it makes it much simpler to guarantee that the disposal happens in the correct order too.

In the grand scheme of things, is it really something that should block this PR?
If you insist, I can change it. But I really think this is a better way of disposing objects.

hlshen: Docs as in comments. It's less important in this PR, but the larger config PR has >10 + subscriptions that I have to read in detail to understand what's going on. Adding a name to them, and returning the disposables together will give me a quick look at all the subscriptions in a file/method. 

If this is truly something you want to keep, then I'll ask for a comment describing each "context.subscriptions.push" block.

rrousselGit: Do you have an example of what that comment would be?
I'm a bit at loss at what you're asking here.

PR #7725, opened 2024-09-24
Pipe through app folder selection to init dataconnect:sdk

joehan:
### Description
Stop asking for app folder twice - just pipe through the selection made in the VSCode UI

### DIFF
--- a/firebase-vscode/src/data-connect/terminal.ts
+++ b/firebase-vscode/src/data-connect/terminal.ts
@@ -15,11 +15,10 @@ export function setTerminalEnvVars(envVar: string, value: string) {
   environmentVariables[envVar] = value;
 }
 
-export function runCommand(command: string, cwd?: string) {
+export function runCommand(command: string) {
   const terminalOptions: TerminalOptions = {
     name: "Data Connect Terminal",
     env: environmentVariables,
-    cwd,
   };
   const terminal = vscode.window.createTerminal(terminalOptions);
   terminal.show();
--- a/firebase-vscode/src/data-connect/terminal.ts
+++ b/firebase-vscode/src/data-connect/terminal.ts
@@ -11,16 +11,16 @@ const executionOptions: vscode.ShellExecutionOptions = {
   env: environmentVariables,
 };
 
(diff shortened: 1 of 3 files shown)


src/init/features/dataconnect/sdk.ts
```diff
@@ -75,11 +76,13 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
   } else {
     // If we aren't, ask the user where their app is, and try to autodetect from there
     logBullet(`Couldn't automatically detect your app directory.`);
-    appDir = await promptForDirectory({
-      config,
-      message:
-        "Where is your app directory? Leave blank to set up a generated SDK in your current directory.",
-    });
+    appDir =
```
fredzqm, line 79: Should this happen before line 71?

If `FDC_APP_FOLDER` is specified, but `process.cwd()` is detected to be a iOS app, it should still respect `FDC_APP_FOLDER`.

fredzqm (approved):
Thanks! 

LG after one suggestion.

src/init/features/dataconnect/sdk.ts
joehan: Good call, fixed.

firebase-vscode/src/data-connect/terminal.ts
```diff
@@ -11,16 +11,16 @@ const executionOptions: vscode.ShellExecutionOptions = {
...
 export function setTerminalEnvVars(envVar: string, value: string) {
   environmentVariables[envVar] = value;
 }
 
-export function runCommand(command: string) {
+export function runCommand(command: string, cwd?: string) {
+  const terminalOptions: TerminalOptions = {
+    name: "Data Connect Terminal",
+    env: environmentVariables,
+    cwd,
```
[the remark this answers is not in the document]
joehan: Nope, cleaned up cwd. I originally thought we could just run the command from the chosen directory, but realized it wouldn't add to the right firebase.json if the chosen dir was outside of the current firebase directory.

PR #7727, opened 2024-09-24
Clean up port finding logic for dataconnect.postgres

joehan:
### Description
As suggested by @yuchenshi, switch the Postgres server port finding logic to the same logic we use for firestore websockets.

### Scenarios Tested
Tested emulators:start with the default port blocked and not blocked - both cases work correctly.

### DIFF
--- a/src/emulator/dataconnectEmulator.ts
+++ b/src/emulator/dataconnectEmulator.ts
@@ -10,7 +10,7 @@ import { FirebaseError } from "../error";
 import { EmulatorLogger } from "./emulatorLogger";
 import { RC } from "../rc";
 import { BuildResult, requiresVector } from "../dataconnect/types";
-import { findOpenPort, listenSpecsToString } from "./portUtils";
+import { listenSpecsToString } from "./portUtils";
 import { Client, ClientResponse } from "../apiv2";
 import { EmulatorRegistry } from "./registry";
 import { logger } from "../logger";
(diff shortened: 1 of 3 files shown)


src/emulator/dataconnectEmulator.ts
```diff
@@ -90,27 +89,28 @@ export class DataConnectEmulator implements EmulatorInstance {
...
+              `Data Connect: Unable to start PGLite server on port ${pgPort} because it is already in use by process number ${process}. This may occur if you are running another instance of Postgres.` +
               ` You can choose a different port by setting 'firebase.json#emulators.dataconnect.postgresPort', or you can unset that field to automatically find an open port.`;
             throw new FirebaseError(errMessage);
           }
         }
-        const port = this.args.postgresPort || (await findOpenPort(5432));
-        const server = await pgServer.createPGServer(this.args.postgresHost, port);
-        connStr = `postgres://${this.args.postgresHost ?? "127.0.0.1"}:${port}/${dbId}?sslmode=disable`;
+        const server = await pgServer.createPGServer(pgHost, pgPort);
+        connStr = `postgres://${pgHost}:${pgPort}/${dbId}?sslmode=disable`;
```
yuchenshi, line 113: Please use `connectableHostname` to handle `pgHost` being 0.0.0.0 etc and quote IPv6 addresses. Similar to https://github.com/firebase/firebase-tools/blob/0e2ab54c522d40ec9fe874e108332427e4886e0a/src/emulator/registry.ts#L187-L192

You may want to use the `new URL`  constructor FWIW

yuchenshi (approved):

src/emulator/dataconnectEmulator.ts
joehan: Done

PR #7730, opened 2024-09-25
Fix more issues with prepared statements and extended query protocol

joehan:
### Description
Bug bash (and Tyler's demoing) uncovered more problems with our handling of extended query protocol and prepared statements:. Fixes b/369172372

- If we're in the midst of a pipeline and the PGLite sends a ErrorMessage, we need to end the pipeline.
- We previously were holding onto prepared statements indefinitely. This lead to issues where the emulator tries to re-prepare statements with the same name. Now, we `DEALLOCATE ALL`  after a new client is authenticated.
- Correctly pipe through socket errors so that we stop orphaning emulators.

### Scenarios Tested
I was using the web quickstart app to repro the original issue.  I confirmed that the movies app works after this change.

### DIFF
--- a/src/emulator/dataconnect/pgliteServer.ts
+++ b/src/emulator/dataconnect/pgliteServer.ts
@@ -43,8 +43,8 @@ export class PostgresServer {
         async onAuthenticated() {
           // Every time we see a new authentication exchange, we need to throw away previously prepared statements -
           // Clients may not know of these statements, and may attempt to reuse the same name.
-          await db.query("DEALLOCATE ALL")
-        }
+          await db.query("DEALLOCATE ALL");
+        },
       });
 
       const extendedQueryPatch: PGliteExtendedQueryPatch = new PGliteExtendedQueryPatch(connection);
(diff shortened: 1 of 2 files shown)


yuchenshi (approved):

src/emulator/dataconnect/pgliteServer.ts
```diff
@@ -39,13 +39,22 @@ export class PostgresServer {
           // pglite wrongly sends.
           return extendedQueryPatch.filterResponse(data, result);
         },
+
+        async onAuthenticated() {
+          // Every time we see a new authentication exchange, we need to throw away previously prepared statements -
+          // Clients may not know of these statements, and may attempt to reuse the same name.
+          await db.query("DEALLOCATE ALL");
```
yuchenshi, line 46: I still don't think this is quite right -- what if there are two clients connected at the same time? Like, one tries to prepare a statement and then the other one authenticates. 

Shouldn't we be deallocating when clients disconnect instead? Prepared statements should be per connection

PR #7731, opened 2024-09-25
Revert the minimum Functions SDK version and add logging for extensions features using v5.1.0

colerogers:
Revert the min functions SDK version for proper functions trigger parsing using the container contract

Add a log for using SDK version under v5.1.0 with the new extensions features


joehan (approved):
LGTM after a changelog and formatting

PR #7733, opened 2024-09-25
Reorder `firebase init dataconnect` prompts for better local onboarding flow.

rosalyntan:
### Description

- If there are existing services / instances, use those for setting up local files
    - Until b/368609569 is fixed, this is only for projects with billing. After, should work regardless of billing state
- If there are no existing services / instances:
    - If no billing:
        - Set up local files with defaults
    - If yes billing:
        - Ask if they want to configure backend resource, otherwise set up local files with defaults

### Scenarios Tested

- [x] project with no billing
- [x] project with billing with existing services
- [x] project with billing with no services, (y) for configuring backend resources
- [x] project with billing with no services, (n) for configuring backend resources

### Sample Commands

project with billing with existing services:
Screenshot 2024-09-25 at 9 40 59 PM

project with billing with no services, (n) for configuring backend resources:
Screenshot 2024-09-25 at 9 39 15 PM

### DIFF
--- a/src/init/features/dataconnect/index.ts
+++ b/src/init/features/dataconnect/index.ts
@@ -107,16 +107,20 @@ async function askQuestions(setup: Setup): Promise<RequiredInfo> {
   if (setup.projectId) {
     isBillingEnabled ? await ensureApis(setup.projectId) : await ensureSparkApis(setup.projectId);
   }
-
   info = await checkExistingInstances(setup, info, isBillingEnabled);
 
-  const shouldConfigureBackend = isBillingEnabled
-    ? await confirm({
-        message: `Would you like to configure your backend resources now?`,
-        default: false,
-      })
-    : false;
-
+  const requiredConfigUnset =
+    info.serviceId === "" ||
+    info.cloudSqlInstanceId === "" ||
+    info.locationId === "" ||
+    info.cloudSqlDatabase === "";
+  const shouldConfigureBackend =
+    isBillingEnabled && requiredConfigUnset
+      ? await confirm({
+          message: `Would you like to configure your backend resources now?`,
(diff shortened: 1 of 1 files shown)


src/init/features/dataconnect/index.ts
```diff
@@ -222,90 +246,77 @@ function subConnectorYamlValues(replacementValues: { connectorId: string }): str
...
+  const existingServicesAndSchemas = await Promise.all(
+    existingServices.map(async (s) => {
+      return {
+        service: s,
+        schema: await getSchema(s.name),
+      };
+    }),
+  );
+  if (existingServicesAndSchemas.length) {
+    const choices: { name: string; value: any }[] = existingServicesAndSchemas.map((s) => {
```
joehan, line 270: Nit: 
```suggestion
    const choices: { name: string; value: { schema: Schema, service: Service } | undefined }[] = existingServicesAndSchemas.map((s) => {
```

src/init/features/dataconnect/index.ts
```diff
@@ -222,90 +246,77 @@ function subConnectorYamlValues(replacementValues: { connectorId: string }): str
...
+      const serviceName = parseServiceName(s.service.name);
+      return {
+        name: `${serviceName.location}/${serviceName.serviceId}`,
+        value: s,
+      };
+    });
+    choices.push({ name: "Create a new service", value: undefined });
+    const choice: { service: Service; schema: Schema } = await promptOnce({
```
joehan, line 278: If you set it above, you can remove this type declaration here because promptOnce will return a defined type.

joehan (approved):
Small type nits but lgtm otherwise

PR #7735, opened 2024-09-26
Add a test-only GitRepoLinks create command

kmcnellis:
### Description

Add a new command for testing, that allows creation of a new gitRepoLink, with a configurable connection ID.  This allows us to test the flow for creating a DevConnect repo & test changing the linked repo via the REST api.

Mostly just moves around existing code in src/apphosting/githubConnections.ts, and allows for passing & testing a new parameter for the connection ID.

### Scenarios Tested

Tested creating a backend to ensure it still works, and tested the new command

### Sample Commands
firebase experiments:enable internaltesting

firebase apphosting:repos:create --project=kiana-firestack-bash --gitconnection=test-connection

firebase apphosting:repos:create --project=kiana-firestack-bash


src/apphosting/githubConnections.ts
```diff
@@ -90,13 +90,32 @@ const ADD_ACCOUNT_CHOICE = "@ADD_ACCOUNT";
...
 /**
- * Prompts the user to link their backend to a GitHub repository.
+ * Prompts the user to create a GitHub connection.
  */
-export async function linkGitHubRepository(
+export async function createGitHubConnection(
   projectId: string,
   location: string,
-): Promise<devConnect.GitRepositoryLink> {
+  connectionId: string | null,
```
mathu97, line 98: Nit

```suggestion
  connectionId: string?,
```

src/apphosting/githubConnections.ts
```diff
@@ -90,13 +90,32 @@ const ADD_ACCOUNT_CHOICE = "@ADD_ACCOUNT";
...
+): Promise<devConnect.Connection> {
   utils.logBullet(clc.bold(`${clc.yellow("===")} Import a GitHub repository`));
+
+  if (connectionId) {
+    // Check if the connection already exists.
+    try {
+      utils.logBullet(
+        clc.bold(`${clc.yellow("===")} Check if connection ${connectionId} already exists`),
```
mathu97, line 106: is this log necessary?

src/apphosting/githubConnections.ts
```diff
@@ -90,13 +90,32 @@ const ADD_ACCOUNT_CHOICE = "@ADD_ACCOUNT";
 const MANAGE_INSTALLATION_CHOICE = "@MANAGE_INSTALLATION";
 
 /**
- * Prompts the user to link their backend to a GitHub repository.
+ * Prompts the user to create a GitHub connection.
  */
-export async function linkGitHubRepository(
+export async function createGitHubConnection(
```
mathu97, line 95: Nit: should this be called `getOrCreateGithubConnection` since we'll be returning an existing connection if a connection with the `connectionId` exists?

src/apphosting/githubConnections.ts
```diff
@@ -117,21 +136,30 @@ export async function linkGitHubRepository(
     installationId = await promptGitHubInstallation(projectId, location, oauthConn);
   }
 
-  let connectionMatchingInstallation = await getConnectionForInstallation(
```
mathu97, line 120: getting rid of this would mean we'll be making duplicate connections for the same github account/org when we can re-use them.

src/apphosting/githubConnections.ts
```diff
@@ -117,21 +136,30 @@ export async function linkGitHubRepository(
...
+
+/**
+ * Prompts the user to link their backend to a GitHub repository.
+ */
+export async function linkGitHubRepository(
+  projectId: string,
+  location: string,
+  createConnectionId: string | null,
+): Promise<devConnect.GitRepositoryLink> {
+  const connectionMatchingInstallation = await createGitHubConnection(
```
mathu97, line 158: Nit: The variable name `connectionMatchingInstallation` doesn't really make sense anymore.

src/apphosting/githubConnections.ts
kmcnellis: This doesn't work. 

src/apphosting/githubConnections.ts:98:17 - error TS8020: JSDoc types can only be used inside documentation comments.

98   connectionId: string?,
                   ~~~~~~~

src/apphosting/githubConnections.ts
kmcnellis: Happy to remove.

src/apphosting/githubConnections.ts
kmcnellis: SGTM

src/apphosting/githubConnections.ts
mathu97: oops should be `connectionId?: string`, updated.

src/apphosting/githubConnections.ts
kmcnellis: Yeah, good point. Didn't mean to remove that

src/apphosting/githubConnections.ts
kmcnellis: done

src/apphosting/githubConnections.ts
kmcnellis: done

src/apphosting/githubConnections.ts
```diff
@@ -117,36 +130,58 @@ export async function linkGitHubRepository(
...
 
-  if (!connectionMatchingInstallation) {
-    connectionMatchingInstallation = await createFullyInstalledConnection(
-      projectId,
-      location,
-      generateConnectionId(),
-      oauthConn,
-      installationId,
-    );
+    if (!createConnectionId || matchingConnectionId === createConnectionId) {
```
mathu97, line 141: is the `matchingConnectionId === createConnectionId` comparison necessary? We are already checking if a connection with `createConnectionId` already exists on line 102, and returning if so.

src/apphosting/githubConnections.ts
```diff
@@ -90,13 +90,26 @@ const ADD_ACCOUNT_CHOICE = "@ADD_ACCOUNT";
...
-export async function linkGitHubRepository(
+export async function getOrCreateGithubConnection(
   projectId: string,
   location: string,
-): Promise<devConnect.GitRepositoryLink> {
+  createConnectionId?: string,
+): Promise<devConnect.Connection> {
   utils.logBullet(clc.bold(`${clc.yellow("===")} Import a GitHub repository`));
+
+  if (createConnectionId) {
```
mathu97, line 102: would be great to add a unittest for this

src/apphosting/githubConnections.ts
```diff
@@ -90,13 +90,26 @@ const ADD_ACCOUNT_CHOICE = "@ADD_ACCOUNT";
...
+    // Check if the connection already exists.
+    try {
+      const connection = await devConnect.getConnection(projectId, location, createConnectionId);
+      utils.logBullet(`Reusing existing connection ${createConnectionId}`);
+      return connection;
+    } catch (e) {
```
mathu97, line 108: It may be good to verify that this is a 404 error and continue to throw if not. We do this in other places as well, see: 
- https://github.com/firebase/firebase-tools/blob/master/src/apphosting/githubConnections.ts#L316
- https://github.com/firebase/firebase-tools/blob/master/src/apphosting/githubConnections.ts#L502

src/commands/apphosting-repos-create.ts
```diff
@@ -0,0 +1,23 @@
+import { Command } from "../command";
+import { Options } from "../options";
+import { needProjectId } from "../projectUtils";
+import requireInteractive from "../requireInteractive";
+import { createGitRepoLink } from "../apphosting";
+import { ensureApiEnabled } from "../gcp/apphosting";
+import { APPHOSTING_TOS_ID } from "../gcp/firedata";
+import { requireTosAcceptance } from "../requireTosAcceptance";
+
+export const command = new Command("apphosting:repos:create")
```
mathu97, line 10: qq: do we expect users to use this command?

If it's just for us maybe it should be behind an experiment?

kmcnellis: It's hidden under "    if (experiments.isEnabled("internaltesting")) {"

mathu97 (approved):

src/apphosting/githubConnections.ts
kmcnellis: Done

kmcnellis: Also added tests for the linkGitHubRepository without a named connection, as they were missing

mathu97 (approved):

PR #7736, opened 2024-09-26
Fix FDC query template.

rosalyntan:
### Description

Came across this issue before in previous bug bashes, where the default query template isn't able to run, but I wasn't sure if it was something weird with my state. I reproduced it while testing a different PR though so I do think it is an actual issue.


joehan (approved):
Good catch - this looks correct against the schema now since there is no director field and it specifically says Movie -> MovieMetadata is 1-1

PR #7753, opened 2024-09-26
Update add data

hlshen:


firebase-vscode/src/data-connect/ad-hoc-mutations.ts
```diff
@@ -162,46 +176,79 @@ query {
...
+        return { kind: Kind.BOOLEAN, value: false };
+      case "Date":
+        return {
+          kind: Kind.STRING,
+          value: new Date().toISOString().substring(0, 10),
+        };
+      case "Float":
+        return { kind: Kind.FLOAT, value: "0" };
+      case "Int":
+        return { kind: Kind.INT, value: "0" };
```
yuchenshi, line 237:
```suggestion
        return { kind: Kind.INT, value: 0 };
```

I think?

firebase-vscode/src/data-connect/ad-hoc-mutations.ts
```diff
@@ -162,46 +176,79 @@ query {
...
+      case "Int64":
+        return { kind: Kind.INT, value: "0" };
+      case "String":
+        return { kind: Kind.STRING, value: "" };
+      case "Timestamp":
+        return { kind: Kind.STRING, value: new Date().toISOString() };
+      case "UUID":
+        return { kind: Kind.STRING, value: "1111-2222-3333-44444" };
```
yuchenshi, line 245:
```suggestion
        return { kind: Kind.STRING, value: "11111111222233334444555555555555" };
```

to make it actually parse. Removing dashes to align with the default output format (so people don't copy a UUID and wonder if they should manually insert dashes)

firebase-vscode/src/data-connect/ad-hoc-mutations.ts
```diff
@@ -162,46 +176,79 @@ query {
...
+        value: defaultValue,
+      });
     }
-    mutation.push(`${functionSpacing}})`, "}"); // closing braces/paren
-    return mutation.join("\n");
+
+    return {
+      kind: Kind.OPERATION_DEFINITION,
+      operation: OperationTypeNode.MUTATION,
+      name: { kind: Kind.NAME, value: singularName },
```
yuchenshi, line 200: Let's keep it anonymous

```suggestion
```

firebase-vscode/src/data-connect/ad-hoc-mutations.ts
```diff
@@ -162,46 +176,79 @@ query {
...
+      selectionSet: {
+        kind: Kind.SELECTION_SET,
+        selections: [
+          {
+            kind: Kind.FIELD,
+            name: { kind: Kind.NAME, value: `${singularName}_insert` },
```
yuchenshi, line 206: I bet this generates `PascalCase_insert` instead of `camelCase_insert`

yuchenshi (approved):

firebase-vscode/src/data-connect/ad-hoc-mutations.ts
tjlav5: If it's anonymous, do we run into problems with folks pressing AddData multiple times? Or do we make a new file each time? I forget how the extension works here

hlshen: I check if there's a file first, so it won't make a bunch of files

firebase-vscode/src/data-connect/ad-hoc-mutations.ts
tjlav5: counter point... empty string with an LSP error message might be better, to make it clear they need to provide a truly unique ID?

otherwise, if we're keeping a valid ID, I'd stick to something easier to read, like all 0's: "00000000000000000000..."

firebase-vscode/src/data-connect/ad-hoc-mutations.ts
```diff
@@ -162,46 +176,79 @@ query {
...
+      case "Int64":
+        return { kind: Kind.INT, value: "0" };
+      case "String":
+        return { kind: Kind.STRING, value: "" };
+      case "Timestamp":
+        return { kind: Kind.STRING, value: new Date().toISOString() };
+      case "UUID":
+        return { kind: Kind.STRING, value: "1111-2222-3333-44444" };
+      case "Vector":
+        return { kind: Kind.LIST, values: [] };
```
tjlav5, line 246: I forget where we landed w/ this during the bash... should this be omitted or an empty list?

firebase-vscode/src/data-connect/terminal.ts
```diff
@@ -43,7 +43,7 @@ export function runTerminalTask(
           resolve(`Successfully executed ${taskName} with command: ${command}`);
         } else {
           reject(
-            new Error(`Failed to execute ${taskName} with command: ${command}`),
+            new Error(`Exit code: ${e.exitCode}. Failed to execute ${taskName} with command: ${command}`),
```
tjlav5, line 46: nit, how about:

`{${e.exitCode}}: Failed to execute....`

firebase-vscode/src/data-connect/ad-hoc-mutations.ts
hlshen: IntValueNode takes in a string for value

firebase-vscode/src/data-connect/ad-hoc-mutations.ts
hlshen: You are a good better

firebase-vscode/src/data-connect/ad-hoc-mutations.ts
hlshen: Not following, for vectors?

joehan: I think the point here is that an empty vector is very likely to be invalid (due to the size directive), so it may be unhelpful to put it there. FWIW, i don't see a great way to handle required vectors here either.

yuchenshi (approved):

PR #7762, opened 2024-09-27
Include the error message as well as error code in the error message

fredzqm:
Tested here. Having the error message makes it so much easy to tell what happened.

We can polish up additional error handling later.

To prod:
Screenshot 2024-09-27 at 11 19 18 AM

To emulator:
Screenshot 2024-09-27 at 11 23 33 AM

### DIFF
--- a/firebase-vscode/src/data-connect/service.ts
+++ b/firebase-vscode/src/data-connect/service.ts
@@ -16,6 +16,7 @@ import { firebaseRC } from "../core/config";
 import {
   dataconnectDataplaneClient,
   executeGraphQL,
+  DATACONNECT_API_VERSION,
 } from "../../../src/dataconnect/dataplaneClient";
 import {
   ExecuteGraphqlRequest,
@@ -246,8 +247,7 @@ export class DataConnectService {
       }
       const client = new Client({
         urlPrefix: endpoint,
-        apiVersion: "v1beta",
-        auth: true,
+        apiVersion: DATACONNECT_API_VERSION,
       });
       const resp = await executeGraphQL(client, servicePath, prodBody);
       return this.handleEmulatorResponse(resp);
--- a/firebase-vscode/src/data-connect/service.ts
+++ b/firebase-vscode/src/data-connect/service.ts
@@ -13,14 +13,17 @@ import { EmulatorsController } from "../core/emulators";
 import { dataConnectConfigs } from "../data-connect/config";
 
 import { firebaseRC } from "../core/config";
-import { executeGraphQL } from "../../../src/dataconnect/dataplaneClient";
+import {
+  dataconnectDataplaneClient,
+  executeGraphQL,
+} from "../../../src/dataconnect/dataplaneClient";
 import {
   ExecuteGraphqlRequest,
   ExecuteGraphqlResponse,
   ExecuteGraphqlResponseError,
   Impersonation,
 } from "../dataconnect/types";
-import { ClientResponse } from "../apiv2";
+import { Client, ClientResponse } from "../apiv2";
 import { InstanceType } from "./code-lens-provider";
 import { pluginLogger } from "../logger-wrapper";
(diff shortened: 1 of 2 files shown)


firebase-vscode/src/data-connect/service.ts
```diff
@@ -243,23 +234,23 @@ export class DataConnectService {
...
+      const endpoint = this.emulatorsController.getLocalEndpoint();
+      if (!endpoint) {
+        throw new DataConnectError(
+          `Emulator isn't running. Please start your emulator!`,
+        );
+      }
+      const client = new Client({
+        urlPrefix: endpoint,
+        apiVersion: "v1beta",
+        auth: true,
```
yuchenshi, line 250: Please revert this to use `"x-mantle-admin": "all"` instead. `auth: true` may send production auth tokens to the emulator and may run into issues when developing offline.

If the `Client` API doesn't do what we want, then using `fetch` is also okay for now. We can always clean code up later but let's not mess with the behavior.

firebase-vscode/src/data-connect/service.ts
```diff
@@ -78,47 +81,35 @@ export class DataConnectService {
...
-    assertExecutionResult(json);
-
-    return json;
-  }
-
-  private async handleInvalidResponse(response: Response): Promise<never> {
-    const cause = await this.decodeResponse(response);
-
-    throw new DataConnectError(
-      `Request failed with status ${response.status}`,
```
fredzqm, line 111: It swallowed the error message here.

firebase-vscode/src/data-connect/service.ts
```diff
@@ -243,23 +234,23 @@ export class DataConnectService {
...
-      return this.handleResponse(resp);
+      const endpoint = this.emulatorsController.getLocalEndpoint();
+      if (!endpoint) {
+        throw new DataConnectError(
+          `Emulator isn't running. Please start your emulator!`,
+        );
+      }
+      const client = new Client({
+        urlPrefix: endpoint,
+        apiVersion: "v1beta",
```
joehan, line 249: Probably should use DATACONNECT_API_VERSION here

firebase-vscode/src/data-connect/service.ts
```diff
@@ -78,47 +81,35 @@ export class DataConnectService {
...
       ExecuteGraphqlResponse | ExecuteGraphqlResponseError
     >,
   ): Promise<ExecutionResult> {
-    if (!(clientResponse.status >= 200 && clientResponse.status < 300)) {
+    if (!(response.status >= 200 && response.status < 300)) {
       const errorResponse =
-        clientResponse as ClientResponse<ExecuteGraphqlResponseError>;
+        response as ClientResponse<ExecuteGraphqlResponseError>;
       throw new DataConnectError(
-        `Request failed with status ${clientResponse.status}`,
```
fredzqm, line 89: It swallowed the error message here.

yuchenshi (approved):

joehan (approved):

firebase-vscode/src/data-connect/service.ts
fredzqm: What about remove `auth: true` here?

`x-mantle-admin` doesn't matter.

firebase-vscode/src/data-connect/service.ts
fredzqm: Done

firebase-vscode/src/data-connect/service.ts
fredzqm: I tested executing against emulator. Everything works fine, both error and success.

fredzqm: Here are some emulator logs: https://paste.googleplex.com/6332029630414848

Good catch on not sending auth token. Emulator has V(3) debug logs for the auth token, won't log by default even if someone sends a real token to it.

PR #7764, opened 2024-09-27
Update ReadME. Fix case for extension dependency

hlshen:


rosalyntan (approved):

firebase-vscode/package.json
```diff
@@ -14,7 +14,7 @@
     "Other"
   ],
   "extensionDependencies": [
-    "graphql.vscode-graphql-syntax"
+    "GraphQL.vscode-graphql-syntax"
```
joehan, line 17: Screenshot 2024-09-27 at 12 09 41 PM
It kinda looks like this is supposed to be lowercase, no?

fredzqm (approved):

yuchenshi (commented):
I meant to approve this PR but I forgot if it's approve, Approve or aPProVe, sorry!

firebase-vscode/package.json
hlshen: If you copy the id it's capitalized in both VSCode and open-vsx.  This fix is targeted at fixing open-VSX builds. Locally it works fine and installs the dependency.

PR #7771, opened 2024-09-30
App Hosting emulator - package manager detection

mathu97:
App hosting emulator will detect the package manager and run it's dev command on `firebase emulators:start`

Supported package managers:
1. npm
2. pnpm
3. yarn

### DIFF
--- a/src/emulator/apphosting/serve.ts
+++ b/src/emulator/apphosting/serve.ts
@@ -2,45 +2,48 @@
  * Start the App Hosting server.
  * @param options the Firebase CLI options.
  */
-import { isIPv4 } from "net";
-import { checkListenable } from "../portUtils";
-import { wrapSpawn } from "../../init/spawn";
-import {discover} from "./utils";
+
+import { spawn } from "cross-spawn";
+import { discoverPackageManager } from "./utils";
 
 /**
  * Spins up a project locally by running the project's dev command.
  */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export async function start(options: any): Promise<{ port: number }> {
-  let port = options.port;
-  while (!(await availablePort(options.host, port))) {
-    port += 1;
-  }
 
-  serve(options, port);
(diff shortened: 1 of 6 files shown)


src/emulator/apphosting/serve.ts
```diff
@@ -2,37 +2,55 @@
...
+  return await hostUrl;
 }
 
 /**
  * Exported for unit testing
  */
-export async function serve(options: any, port: string) {
-  // TODO: update to support other package managers and frameworks other than NextJS
-  await wrapSpawn("npm", ["run", "dev", "--", "-H", options.host, "-p", port], process.cwd());
+export function getHostUrlFromString(data: string): string | undefined {
```
taeold, line 51: Can you clarify what this function is doing? It looks like it's doing few things like

1) ignoring URLs that starts with https://  (why?)
2) stripping out "/" (why?)

src/emulator/apphosting/serve.ts
```diff
@@ -2,37 +2,55 @@
...
+   * */
+
+  const hostUrl = new Promise<string>((resolve, reject) => {
+    const serve = spawn(packageManager, ["run", "dev"], {
+      cwd: rootDir,
+    });
+
+    serve.stdout.on("data", (data: any) => {
+      process.stdout.write(data);
+      const url = getHostUrlFromString(data.toString());
```
taeold, line 36: I think we can reasonably expect that the

1) dev server will run on localhost
2) dev server will respect the `PORT` environment variable

I feel like these expectation is probably as reasonable as expecting the dev server to print out the URL. wdyt?

mathu97: +1 reading from STDOUT definitely does not feel right, and I think these assumptions are valid.

src/emulator/apphosting/serve.ts
mathu97: outdated - not using the approach of reading host/port from STDOUT anymore

taeold (approved):

PR #7777, opened 2024-10-01
init dataconnect:sdk should preserve existing packageJsonDir if not overridden

fredzqm:


yuchenshi (approved):

src/init/features/dataconnect/sdk.ts
```diff
@@ -141,6 +141,7 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
     const javascriptSdk: JavascriptSDK = {
       outputDir,
       package: pkg,
+      packageJsonDir: newConnectorYaml.generate.javascriptSdk?.packageJsonDir,
```
maneesht, line 144: I remember there being a prompt for this at some point elsewhere in the code actually

src/init/features/dataconnect/sdk.ts
```diff
@@ -141,6 +141,7 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
...
     };
 
     if (
```
maneesht, line 147: I think below we should prompt even if there isn't a package Json dir

fredzqm: Feel free to tweak it.

If there is no `package.json`, what would `packageJsonDir` do? Try to `npm i`, but fails?

PR #7780, opened 2024-10-02
Modernize `use.ts` and stop requiring auth for use

joehan:
### Description
Modernized 'use.ts' to use async/await, and make `firebase use` work when not logged in. In cases where users don't actually have access to the project, they'll hit normal auth errors later when they execute commands, which is fine.

### DIFF
--- a/src/commands/use.ts
+++ b/src/commands/use.ts
@@ -1,7 +1,7 @@
 import * as clc from "colorette";
 
 import { Command } from "../command";
-import { listFirebaseProjects } from "../management/projects";
+import { listFirebaseProjects, getFirebaseProject } from "../management/projects";
 import { FirebaseProjectMetadata } from "../types/project";
 import { logger } from "../logger";
 import { Options } from "../options";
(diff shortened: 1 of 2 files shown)


src/commands/use.ts
```diff
@@ -62,48 +69,36 @@ export const command = new Command("use [alias_or_project_id]")
...
-            utils.makeActiveProject(options.projectRoot, newActive);
-            logger.info("Now using project", clc.bold(newActive));
-          } else {
-            // no alias or project recognized
-            return utils.reject("Invalid project selection, " + verifyMessage(newActive));
-          }
-        });
+        makeActiveProject(options.projectRoot, newActive);
+        logger.info("Now using alias", clc.bold(newActive), "(" + resolvedProject + ")");
+      } else if (project) {
```
aalej, line 94: I think the tests are failing since `project` never gets assigned a value. It looks like we assigned the value using https://github.com/firebase/firebase-tools/blob/495cb72c8e7a4270a4e9d602fddc9ee8fdca53d6/src/commands/use.ts#L65-L68, but we removed it here since `getFirebaseProject` requires the user to be logged in .

It might be possible to change `project` -> `resolvedProject`

joehan (commented):
I need to fix the tests that this is breaking, no need to review yet

PR #7781, opened 2024-10-02
Apphosting emulator - environment variable support

mathu97:
Upon starting the App hosting emulator, `apphosting.yaml` and `apphosting.local.yaml` configuration files are read for environment variables and made available to the user's app.

### DIFF
--- a/src/emulator/apphosting/environments.ts
+++ b/src/emulator/apphosting/environments.ts
@@ -53,7 +53,7 @@ export async function loadAppHostingYaml(
 
 /**
  * Loads in apphosting.yaml & apphosting.local.yaml, giving
- * apphosting.local.yaml precedence if one is present.
+ * apphosting.local.yaml precedence if present.
  */
 export async function getLocalAppHostingConfiguration(
   sourceDirectory: string,
--- a/src/emulator/apphosting/environments.ts
+++ b/src/emulator/apphosting/environments.ts
@@ -1,8 +1,13 @@
 import { readFileFromDirectory, wrappedSafeLoad } from "../../utils";
+import { pathExists } from "fs-extra";
+import { join } from "path";
+import { logger } from "./utils";
+import { Emulators } from "../types";
 
 export type EnvironmentAvailability = "BUILD" | "RUNTIME";
 
 const APPHOSTING_YAML = "apphosting.yaml";
+const APPHOSTING_LOCAL_YAML = "apphosting.local.yaml";
 
 interface AppHostingYaml {
   env?: {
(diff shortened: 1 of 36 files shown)


src/emulator/apphosting/environments.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  return { environmentVariables, secrets };
+}
+
+/**
+ * Loads in apphosting.yaml & apphosting.local.yaml, giving
+ * apphosting.local.yaml precedence if present.
+ */
+export async function getLocalAppHostingConfiguration(
```
taeold, line 60: FYI: I wrote a very similar code for loading .env files few years ago.

https://github.com/firebase/firebase-tools/blob/48f08ce0d7a29101c6ef717cf71d8563e157a824/src/functions/env.ts#L366

interesting (for me) to see how we approached the problem slightly differently.

src/emulator/apphosting/index.ts
```diff
@@ -12,24 +12,23 @@ interface AppHostingEmulatorArgs {
  * environment for testing App Hosting features locally.
  */
 export class AppHostingEmulator implements EmulatorInstance {
-  private logger = EmulatorLogger.forEmulator(Emulators.APPHOSTING);
   constructor(private args: AppHostingEmulatorArgs) {}
 
   async start(): Promise<void> {
-    this.logger.logLabeled("INFO", Emulators.APPHOSTING, "starting apphosting emulator");
+    logger.logLabeled("INFO", Emulators.APPHOSTING, "starting apphosting emulator");
```
taeold, line 18: q: why did we decide not to use the emulator logger? I assumed that would be the preferred/consistent way for the emulator.

src/emulator/apphosting/serve.ts
```diff
@@ -32,8 +33,12 @@ export async function start(): Promise<{ hostname: string; port: number }> {
 async function serve(port: number): Promise<void> {
   const rootDir = process.cwd();
   const packageManager = await discoverPackageManager(rootDir);
+  const apphostingLocalConfig = await getLocalAppHostingConfiguration(rootDir);
 
-  await wrapSpawn(packageManager, ["run", "dev"], rootDir, { PORT: port });
+  await wrapSpawn(packageManager, ["run", "dev"], rootDir, {
+    ...apphostingLocalConfig.environmentVariables,
```
taeold, line 39: secret support coming later?

taeold (approved):
minor nits

src/emulator/apphosting/index.ts
mathu97: We are using the emulator logger. Just refactored so that we don't have to keep instantiating the emulator logger in every file we want to log: https://github.com/firebase/firebase-tools/pull/7781/files/eddd7aed4d0ac0a20b5bf56373c79b8257fc1187#diff-5c4c65d13162d36f52f131c06e42731e0ee91acf7064fc5d2a9568ccb2f4a799R6

src/emulator/apphosting/serve.ts
mathu97: yes!

src/emulator/apphosting/environments.ts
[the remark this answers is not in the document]
mathu97: yep that makes sense

PR #7784, opened 2024-10-03
make logs default to .firebase/logs/

hlshen:
Remove setting for conditionally enabling logs.

Set default logs path to .firebase/logs/firebase-extension-debug.log

### DIFF
--- a/firebase-vscode/src/logger-wrapper.ts
+++ b/firebase-vscode/src/logger-wrapper.ts
@@ -35,25 +35,17 @@ for (const logLevel in pluginLogger) {
 /**
  * Logging setup for logging to console and to file.
  */
-export function logSetup({
-  shouldWriteDebug,
-  debugLogPath,
-}: {
-  shouldWriteDebug: boolean;
-  debugLogPath: string;
-}) {
+export function logSetup() {
   // Log to console (use built in CLI functionality)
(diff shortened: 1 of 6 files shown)


firebase-vscode/src/logger-wrapper.ts
```diff
@@ -35,25 +35,17 @@ for (const logLevel in pluginLogger) {
...
 
   // Log to file
   // Only log to file if firebase.debug extension setting is true.
-  if (shouldWriteDebug) {
     // Re-implement file logger call from ../../src/bin/firebase.ts to not bring
     // in the entire firebase.ts file
     const rootFolders = getRootFolders();
-    const filePath =
-      debugLogPath || path.join(rootFolders[0], "firebase-plugin-debug.log");
+    const filePath = path.join(rootFolders[0], ".firebase", "logs", "firebase-extension-debug.log");
```
yuchenshi, line 48: nit:

```suggestion
    const filePath = path.join(rootFolders[0], ".firebase", "logs", "fdc-extension-debug.log");
```

Also, a prefix isn't really necessary because it's in the `.firebase` directory. Just `extension-debug.log` will also do. Up to you

yuchenshi (approved):

joehan (approved):

PR #7785, opened 2024-10-03
adding emulator support for SDK-defined extensions

ifielker:
Adding emulator support for SDK-defined extensions

Also fixed various other bugs I found along the way.

### DIFF
--- a/src/emulator/eventarcEmulator.ts
+++ b/src/emulator/eventarcEmulator.ts
@@ -76,7 +76,7 @@ export class EventarcEmulator implements EmulatorInstance {
       return { projectId, triggerName, eventTrigger, key };
     };
 
-    const removeTriggerRoute = `/emulator/v1/remove/projects/:project_id/triggers/:trigger_name`;
+    const removeTriggerRoute = `/emulator/v1/remove/projects/:project_id/triggers/:trigger_name(*)`;
     const removeTriggerHandler: express.RequestHandler = (req, res) => {
       try {
         const { projectId, triggerName, eventTrigger, key } = getTriggerIdentifiers(req);
--- a/src/emulator/functionsEmulator.ts
+++ b/src/emulator/functionsEmulator.ts
@@ -639,6 +639,7 @@ export class FunctionsEmulator implements EmulatorInstance {
     const toRemove = Object.keys(this.triggers).filter((recordKey) => {
       const record = this.getTriggerRecordByKey(recordKey);
       if (record.backend.codebase !== emulatableBackend.codebase) {
+        // Order is important here. This needs to go before any other checks.
         // We are only loading one codebase, don't delete triggers from another.
         return false;
       }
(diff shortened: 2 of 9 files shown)


src/emulator/eventarcEmulator.ts
```diff
@@ -76,7 +76,7 @@ export class EventarcEmulator implements EmulatorInstance {
       return { projectId, triggerName, eventTrigger, key };
     };
 
-    const removeTriggerRoute = `/emulator/v1/remove/projects/:project_id/triggers/:trigger_name`;
+    const removeTriggerRoute = `/emulator/v1/remove/projects/:project_id/triggers/:trigger_name(*)`;
```
joehan, line 79: Quick sanity check since this seems like an unrelated bug fix - could you briefly explain why we made this change?

src/emulator/functionsEmulator.ts
```diff
@@ -594,6 +638,10 @@ export class FunctionsEmulator implements EmulatorInstance {
     // Remove any old trigger definitions
     const toRemove = Object.keys(this.triggers).filter((recordKey) => {
       const record = this.getTriggerRecordByKey(recordKey);
+      if (record.backend.codebase !== emulatableBackend.codebase) {
+        // We are only loading one codebase, don't delete triggers from another.
+        return false;
+      }
```
joehan, line 644: Not sure either way, but should this be below the `if (force)` conditional?

joehan (approved):
LGTM with some small questions - also, can you add a changelog?

src/emulator/eventarcEmulator.ts
ifielker: Sure: During testing I noticed that removing eventarc triggers didn't work and always gave an error message. I eventually figured out that it was because the POST request was not getting to the removeTrigger handler, but rather hitting the fallthrough "*" handler instead. I investigated a bit more and realized it was because the trigger name has slashes (/) in it. So I looked at the addTriggerRoute (because that one was working) and saw the (*) on the end. Tried that same pattern for the removeTriggerRoute, and then it started working and removing the eventarc triggers when it was supposed to.

src/emulator/functionsEmulator.ts
ifielker: No. The order is important here, and we should not be removing triggers from a different backend. This is what was happening before: If you had 2 codebases, then you loop through the codebases with loadTriggers and force=true. So codebaseA would delete all existing triggers, then add trigger1, trigger2 say, but then during the next loop, loadTriggers(codebaseB, true) would be run and it would delete trigger1, trigger2 and then add it's own triggers. So it clobbered the triggers that came before. BUT it wasn't too big of a problem because the eventarc remove was broken and the removeEventTriggers were not really implemented. So stuff stuck around by accident. It's fixed to work properly now: each codebase is only allowed to optionally remove its own triggers. (So the fixed eventarc deletion will not cause problems.)

PR #7787, opened 2024-10-04
fix: prevent failure when deleting cloud function with missing schedule

### LINKED ISSUE #4795: Function deploy fails to delete scheduled function without Cloud Scheduler job

**firebase-tools:** 11.4.0

**Platform:** macOS

### [REQUIRED] Test case

```js

export const v1scheduled = functions.pubsub.schedule("every 30 minutes").onRun(() => {});
```

1. Deploy a scheduled function
2. Delete the Cloud Scheduler job associated w/ the scheduled function
3. Try deleting the function via `firebase functions:delete <fn>`
 
Delete the pubsub topic and the function

Command fails:

HTTP 404: Job not found

You have to go manually delete the function in Cloud console. This is annoying, especially if you are running something like `firebase deploy --force` to clean up old functions as it marks the entire deployment as failed.

crossan007:
Closes: 4795

### Description
Fixes https://github.com/firebase/firebase-tools/issues/4795#issuecomment-2393788829

When deletion of a `pubsub.schedule` Cloud Function partially fails, manual interaction is required to fully delete the function.

In my case, deletion of the Cloud Function was unpredictably failing after deleting the associated Cloud Schedule, leaving the function unable to be deleted by a deployment even with the `--force` CLI flag

This MR changes the behavior of Cloud Functions deployment so that a function which is supposed to have an associated schedule will be deleted if the schedule no longer exists.

#### Using the latest build of `master` @ `a54e4ff20cc0aca11d89e2f6b31c35f063918d87`
1. Deploy a Cloud Function (v1) using `firebase deploy`
2. Confirm the function exists https://console.cloud.google.com/functions/list
3. Confirm the schedule for the function exists https://console.cloud.google.com/cloudscheduler
4. Delete the schedule associated with the function
  a.  I am unsure what specific scenarios cause the schedule to be deleted without the function being deleted, but it's happened to me a number of times
  b.  Notice the `Trigger` column in the Firebase console becomes blank:
373668886-11645a9a-b9ee-44e5-a496-d60816517304
5. Remove the function from the deployment source code
6. Run `firebase deploy` 
  a.  Observe the failure in the deploy logs: `HTTP Error: 404, Job not found.`
  b.  Observe the final failure notice in the deploy logs:
    ```
    Error: There was an error deploying functions:
      Error Failed to upsert schedule function <function name> in region us-central1
    ```

#### Using the changes from this merge request

> repeat steps 1-5

6. Run `firebase deploy` 
  a.  Observe the warning in the deploy logs:
      ```
      HTTP Error: 404, Job not found.
      Trigger for dispatchCampaign not found, continuing with deletion of function
      ```
    b.  Confirm the deployment succeeds and the affected functions have been removed.


crossan007:
@taeold and @christhompsongoogle  are tagged in the associated issue

src/deploy/functions/release/fabricator.ts
```diff
@@ -214,7 +214,38 @@ export class Fabricator {
...
+       * For example, if an account lacking the `pubsub.topics.delete` permission
+       * attempts to delete a function with a Pub/Sub trigger, the trigger will be
+       * deleted but the topic will not. This will cause the function deletion
+       * to fail on subsequent attempts.
+       * 
+       */
+      if (
+          err instanceof reporter.DeploymentError && 
+          err.op == "delete schedule" && 
+          (err.original as FirebaseError).message == "HTTP Error: 404, Job not found."
```
joehan, line 234: Drive by: Probably safer to just check if err.status == 404 - relying on the exact string is vulnerable to breaking in the future.

src/deploy/functions/release/fabricator.ts
```diff
@@ -214,7 +214,38 @@ export class Fabricator {
...
+        ) {
+          logger.warn(`Trigger for ${endpoint.id} not found. Continuing to delete function.`);
```
joehan, line 236:
```suggestion
          logger.warn(`Cloud Scheduler job for ${endpoint.id} not found. It may have been deleted out of band. Continuing to delete function.`);
```

src/deploy/functions/release/fabricator.ts
```diff
@@ -214,7 +214,38 @@ export class Fabricator {
...
+      }
+      else if (
+          err instanceof reporter.DeploymentError && 
+          err.op == "delete topic" && 
+          err.message.startsWith("Failed to delete topic function")
```
joehan, line 241: This concerns me - if the topic deletion fails for a generic reason, I don't like orphaning the resource. If we can check for specific failures (ie 404s), I'd be more comfortable with this

joehan (commented):
Thanks for the contribution! I added some early review feedback, and assigned some folks with more context on functions to give a more thorough pass.

Could you also add a changelog entry?

src/deploy/functions/release/fabricator.ts
crossan007: I considered that; however, I wasn't sure whether the backend here may re-use the 404 status code for a scenario where we *would not* want to continue cleanup.

I can adjust this either way based on guidance from someone familiar with the possible failure modes here.

joehan:
Fixes #4795

PR #7788, opened 2024-10-04
Use quota metric to check CSQL free trial eligibility

joehan:
### Description
Use quota metric `sqladmin.googleapis.com/fdc_lifetime_free_trial_per_project` to check CSQL free trial eligibility.

### Scenarios Tested
Screenshot 2024-10-04 at 1 19 51 PM

### DIFF
--- a/src/dataconnect/freeTrial.ts
+++ b/src/dataconnect/freeTrial.ts
@@ -10,14 +10,19 @@ export function freeTrialTermsLink(): string {
 const FREE_TRIAL_METRIC = "sqladmin.googleapis.com/fdc_lifetime_free_trial_per_project";
 
 // Checks whether there is already a free trial instance on a project.
-export async function checkFreeTrialInstanceEligibility(projectId: string): Promise<boolean> {
-
+export async function checkFreeTrialInstanceUsed(projectId: string): Promise<boolean> {
+  const past7d = new Date();
+  past7d.setDate(past7d.getDate() - 7);
   const query: CmQuery = {
-    filter:`metric.type="${FREE_TRIAL_METRIC}"`,
+    filter: `metric.type="serviceruntime.googleapis.com/quota/allocation/usage" AND metric.label.quota_metric = "${FREE_TRIAL_METRIC}"`,
+    "interval.endTime": new Date().toJSON(),
+    "interval.startTime": past7d.toJSON(),
   };
   const ts = await queryTimeSeries(query, projectId);
-  console.log(ts);
-  return false;
+  if (ts.length) {
+    return ts[0].points.some((p) => p.value.int64Value);
+  }
+  return true;
 }
 
 export async function getFreeTrialInstanceId(projectId: string): Promise<string | undefined> {
@@ -33,7 +38,7 @@ export function printFreeTrialUnavailable(
   if (!instanceId) {
     utils.logLabeledBullet(
       "data connect",
-      "The CloudSQL free trial has already been used on this project."
+      "The CloudSQL free trial has already been used on this project.",
     );
     utils.logLabeledBullet(
       "data connect",
@@ -49,8 +54,8 @@ export function printFreeTrialUnavailable(
     `To use a different database in the same instance, ${clc.bold(`change the ${clc.blue("instanceId")} to "${instanceId}"`)} in ` +
     `${clc.green(configYamlPath)}. (Also, update the ${clc.blue("database")} field (i.e. DB name in the instance) ` +
     `and ${clc.blue("location")} as needed.)`;
-  utils.logLabeledBullet("data connect", reuseHint);
-  utils.logLabeledBullet(
+  utils.logLabeledError("data connect", reuseHint);
+  utils.logLabeledError(
     "data connect",
     `Or you may create a new (paid) CloudSQL instance at https://console.cloud.google.com/sql/instances`,
   );
--- a/src/dataconnect/freeTrial.ts
+++ b/src/dataconnect/freeTrial.ts
@@ -24,11 +24,10 @@ export async function checkFreeTrialInstanceUsed(projectId: string): Promise<boo
       return ts[0].points.some((p) => p.value.int64Value);
     }
     return true;
-  } catch(err: any) {
+  } catch (err: any) {
     // If the metric doesn't exist, free trial is not used.
     return false;
   }
-
 }
 
 export async function getFreeTrialInstanceId(projectId: string): Promise<string | undefined> {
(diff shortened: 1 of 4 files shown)


src/dataconnect/freeTrial.ts
```diff
@@ -25,9 +59,9 @@ export function printFreeTrialUnavailable(
     `To use a different database in the same instance, ${clc.bold(`change the ${clc.blue("instanceId")} to "${instanceId}"`)} in ` +
     `${clc.green(configYamlPath)}. (Also, update the ${clc.blue("database")} field (i.e. DB name in the instance) ` +
     `and ${clc.blue("location")} as needed.)`;
-  utils.logLabeledBullet("dataconnect", reuseHint);
-  utils.logLabeledBullet(
-    "dataconnect",
+  utils.logLabeledError("data connect", reuseHint);
```
yuchenshi, line 62: Do we need all of these to be errors? The above can be seen as an error, but this one here and below are more like hints (informational). I don't have a strong preference here, but if you do decide to go with all errors, please also update the logs in the if-branch above

src/dataconnect/freeTrial.ts
```diff
@@ -6,17 +7,50 @@ export function freeTrialTermsLink(): string {
   return "https://firebase.google.com/pricing";
 }
 
+const FREE_TRIAL_METRIC = "sqladmin.googleapis.com/fdc_lifetime_free_trial_per_project";
+
 // Checks whether there is already a free trial instance on a project.
-export async function checkForFreeTrialInstance(projectId: string): Promise<string | undefined> {
+export async function checkFreeTrialInstanceUsed(projectId: string): Promise<boolean> {
+  const past7d = new Date();
+  past7d.setDate(past7d.getDate() - 7);
```
yuchenshi, line 15: Now I think about it, what does this actually do? Will the CLI prevent creating a new no-cost trial instance even if you've deleted your previous one 6 days ago, and you just have to wait a bit more? Or is this reflecting a backend limitation?

fredzqm: `fdc_lifetime_free_trial_per_project` is monotonically increasing.
If you create and delete the no cost trial cloud sql instance, this metric would remain at 1.
At that state, Cloud SQL won't allow re-creating the free trial instance.

What Joe implemented is correct. 7d is an arbitrary number in case recent metric hasn't been uploaded to Chemist.

src/dataconnect/provisionCloudSql.ts
```diff
@@ -68,9 +69,9 @@ export async function provisionCloudSql(args: {
     if (err.status !== 404) {
       throw err;
     }
-    const freeTrialInstanceId = await checkForFreeTrialInstance(projectId);
-    if (freeTrialInstanceId) {
-      printFreeTrialUnavailable(projectId, freeTrialInstanceId, configYamlPath);
+    const freeTrialInstanceId = await getFreeTrialInstanceId(projectId);
+    if (await checkFreeTrialInstanceUsed(projectId)) {
+      printFreeTrialUnavailable(projectId, configYamlPath, freeTrialInstanceId);
       throw new FirebaseError("Cannot create another no-cost trial Cloud SQL instance.");
```
fredzqm, line 75: "No-cost trial Cloud SQL instance is no longer available"

`another` is no longer accurate given CSQL don't support re-create.

src/init/features/dataconnect/index.ts
```diff
@@ -327,8 +327,10 @@ async function checkExistingInstances(
     // If we've already chosen a region (ie service already exists), only list instances from that region.
     choices = choices.filter((c) => info.locationId === "" || info.locationId === c.location);
     if (choices.length) {
-      const freeTrialInstanceId = await checkForFreeTrialInstance(setup.projectId);
-      if (!freeTrialInstanceId) {
+      const freeTrialInstanceId = await getFreeTrialInstanceId(setup.projectId);
+      if (await checkFreeTrialInstanceUsed(setup.projectId)) {
+        logBullet(`Found an existing CloudSQL free trial instance ${freeTrialInstanceId}`);
```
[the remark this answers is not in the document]
fredzqm: +1 `checkFreeTrialInstanceUsed` can return true even if there is no free trial instance running.

This log may be confusing if c.location doesn't match freeCsql.location.
I would recommend looping through `choices` and append suffix to the displayed free trial instance.

Change above to:
```
    let choices = instances.map((i) => {
      let display = i.name;
      if (i.settings.userLabels?.["firebase-data-connect"] === "ft") {
         display += " (no cost trial)";
      }
      return { name: display, value: i.name, location: i.region };
    });
```

fredzqm (approved):

src/dataconnect/freeTrial.ts
```diff
@@ -6,17 +7,50 @@ export function freeTrialTermsLink(): string {
...
+  instanceId?: string,
 ) {
+  if (!instanceId) {
+    utils.logLabeledBullet(
+      "data connect",
+      "The CloudSQL free trial has already been used on this project.",
+    );
+    utils.logLabeledBullet(
+      "data connect",
+      `You may create or use a paid CloudSQL instance by visiting https://console.cloud.google.com/sql/instances`,
```
fredzqm, line 50: Or try it out in a new project.

src/dataconnect/freeTrial.ts
```diff
@@ -6,17 +7,50 @@ export function freeTrialTermsLink(): string {
...
 export function printFreeTrialUnavailable(
   projectId: string,
-  instanceId: string,
   configYamlPath: string,
+  instanceId?: string,
 ) {
+  if (!instanceId) {
+    utils.logLabeledBullet(
+      "data connect",
+      "The CloudSQL free trial has already been used on this project.",
```
fredzqm, line 46: Suggest " You may try a different project."?

src/dataconnect/freeTrial.ts
joehan: Will do - I updated them for msotly visual reasons - it was a bit awkward having these semantically linked messages print out with different color badges.

src/dataconnect/freeTrial.ts
joehan: Yep, this is enforced on the backend too.

src/dataconnect/provisionCloudSql.ts
joehan: Reworded

src/init/features/dataconnect/index.ts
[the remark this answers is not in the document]
joehan: Good call, refactored this, and added the region to the display name as well.

PR #7789, opened 2024-10-05
Maintenance

ifielker:
- fixing some lint warnings 
- replacing some chained promises with async/await

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -55,7 +55,7 @@ function isObject(value: unknown): value is Record<string, unknown> {
 }
 
 /**
- * Safely gets a status from an unknown object if
+ * Safely gets a status from an unknown object if it has one.
  * @param err The error to get the status of
  * @param defaultStatus a default status if there is none
  * @return the err status, a default status or DEFAULT_STATUS
--- a/src/error.ts
+++ b/src/error.ts
@@ -34,6 +34,41 @@ export class FirebaseError extends Error {
 }
 
 /**
+ * Safely gets an error message from an unknown object
+ * @param err an unknown error type
+ * @param defaultMsg an optional message to return if the err is not Error or string
+ * @return An error string
+ */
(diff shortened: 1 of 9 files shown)


src/error.ts
```diff
@@ -33,6 +33,41 @@ export class FirebaseError extends Error {
   }
 }
 
+/**
+ * Safely gets an error message from an unknown object
+ * @param err an unknown error type
+ * @param defaultMsg an optional message to return if the err is not Error or string
+ * @return An error string
+ */
+export function getErrMsg(err: unknown, defaultMsg?: string): string {
```
joehan, line 42: Love this -  those `unsafe any access` warnings on every `catch(err: any)` are very obnoxious

src/error.ts
```diff
@@ -33,6 +33,41 @@ export class FirebaseError extends Error {
...
+  }
+  return JSON.stringify(err);
+}
+
+function isObject(value: unknown): value is Record<string, unknown> {
+  return typeof value === "object" && value !== null;
+}
+
+/**
+ * Safely gets a status from an unknown object if
```
joehan, line 58:
```suggestion
 * Safely gets a status from an unknown object if it has one.
```

src/extensions/extensionsApi.ts
```diff
@@ -458,14 +467,14 @@ export async function getExtensionVersion(extensionVersionRef: string): Promise<
       populateSpec(res.body.spec);
     }
     return res.body;
-  } catch (err: any) {
-    if (err.status === 404) {
+  } catch (err: unknown) {
+    if (isObject(err) && err.status === 404) {
```
joehan, line 471: Should this just use getErrStatus?

joehan (approved):

src/extensions/extensionsApi.ts
ifielker: yup, I wrote this before I wrote that, and forgot to update it.

PR #7790, opened 2024-10-06
FDC SDK init use the same outputDir pattern for both CLI and VS Code

fredzqm:
### Description

Currently, VS Code button configure outputDir to `dataconnect-generated` under `appDir`. CLI configures outputDir to `generated/${language}/${package}`.

I like the idea of the `dataconnect-generated` as the gen SDK convention. Clear name to remind customers that this is generated by dataconnect. We also should add a sub folder within it to signal language in case there are >1 apps in the folder.

How about `dataconnect-generated/{js,swift,kotlin}/${package}` as the gen SDK output path to consolidate both ideas?
Note: `${package}` is part of `outputDir` of `JS` and `Dart`, but implicitly included for `Swift` and `Kotlin`.

Updated the default package generation algorithm based on offline discussions.

- ConnectorID: `default`
- JS: `@firebasegen/default-connector`
- Dart: `default_connector`
- Android: `connectors.default`
- IOS: `DefaultConnector`

Other bug fixes:
- Always override the gen SDK config. Previously, we could get in state with old package but new outputDir...
- `Javascript.packageJsonDir` is always set to appDir. There is little downside since emulator would ignore it gracefully. We should improve emulator logging though: http://cl/682785528

### Scenarios Tested

Added lots of unit tests for each platform. Run with `npx mocha src/dataconnect/fileUtils.spec.ts`.


src/init/features/dataconnect/sdk.ts
```diff
@@ -74,9 +71,9 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
     targetPlatform = cwdPlatformGuess;
   } else {
     // If we aren't, ask the user where their app is, and try to autodetect from there
-    logBullet(`Couldn't automatically detect your app directory.`);
+    logBullet(`Couldn't automatically detect your app in directory ${appDir}.`);
     appDir =
-      process.env[FDC_APP_FOLDER] ??
+      process.env[FDC_APP_FOLDER] ||
```
fredzqm, line 76: BUG: When `process.env[FDC_APP_FOLDER]` is an empty string, `??` would take it.
BUG: in the terminal started from gen SDK button, `FDC_APP_FOLDER` is set, so subsequent init sdk call won't prompt. Logging the appDir makes it clear what's happening.

src/init/features/dataconnect/sdk.ts
```diff
@@ -109,103 +105,83 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
...
-
     const javascriptSdk: JavascriptSDK = {
-      outputDir,
-      package: pkg,
-      packageJsonDir: newConnectorYaml.generate.javascriptSdk?.packageJsonDir,
+      outputDir: path.relative(connectorDir, path.join(appDir, `dataconnect-generated/js`)),
+      package: `@firebasegen/${connectorYaml.connectorId}`,
+      // If appDir has package.json, this would install JS SDK there.
+      // Otherwise, emulator would ignore it. Always add it here in case `package.json` is added later.
+      packageJsonDir: path.relative(connectorDir, appDir),
```
fredzqm, line 155: https://github.com/firebase/firebase-tools/pull/7777#discussion_r1783526292

Maneesh brought up a good point about missing `package.json`.

I don't think there is any harm to always set it. Emulator can ignore it and print a more clear debug log: http://cl/682785528

src/init/features/dataconnect/sdk.ts
```diff
@@ -109,103 +106,90 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
...
     // app/src/main/kotlin and app/src/main/java are conventional for Android,
     // but not required or enforced. If one of them is present (preferring the
-    // "kotlin" directory), use it. Otherwise, fall back to the app directory.
-    let baseDir = path.join(appDir, `generated/kotlin`);
+    // "kotlin" directory), use it. Otherwise, fall back to the dataconnect-generated dir.
     for (const candidateSubdir of ["app/src/main/java", "app/src/main/kotlin"]) {
       const candidateDir = path.join(appDir, candidateSubdir);
-      if (fs.existsSync(candidateDir)) {
-        baseDir = candidateDir;
+      if (dirExistsSync(candidateDir)) {
```
fredzqm, line 185: Had to change this to use the helper because `fs.existsSync` don't work well work mock-fs

src/init/features/dataconnect/sdk.ts
maneesht: `      // If appDir has package.json, Emulator would install JS SDK for every reload.` - this is not accurate. We only add the dependency if it doesn't already exist in the package.json. So it would add the first time, and then skip the rest.

Also, fwiw, this doesn't automatically install the JS SDK, it automatically installs your generated SDK package, and your package manager will automatically install the JS SDK depending on which one you're using.

What's the reason to remove the prompt? It's an optional field, so I feel like we might want to just keep the prompt.

src/dataconnect/fileUtils.spec.ts
```diff
@@ -127,81 +129,254 @@ describe("getPlatformFromFolder", () => {
...
+      desc: string;
+      appDir: string;
+      output: any;
+    }[] = [
+      {
+        desc: "basic",
+        appDir: appFolderBase,
+        output: {
+          outputDir: "../dataconnect-generated/dart/test-connector",
+          package: "test-connector",
```
maneesht, line 340: This should be an underscore for Dart, fwiw

fredzqm: Done

src/init/features/dataconnect/sdk.ts
```diff
@@ -104,103 +101,90 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
...
-    const pkg = newConnectorYaml.generate.dartSdk?.package ?? newConnectorYaml.connectorId;
+  if (targetPlatform === Platform.FLUTTER) {
     const dartSdk: DartSDK = {
-      outputDir,
-      package: pkg,
+      outputDir: path.relative(
+        connectorDir,
+        path.join(appDir, `dataconnect-generated/dart/${snakeCase(connectorYaml.connectorId)}`),
+      ),
+      package: snakeCase(connectorYaml.connectorId),
```
fredzqm, line 165: @maneesht brought this up in chat.

`default` isn't a valid dart package name, but out default connector name is `default`. Need a better pattern here.

fredzqm:
Updated the default package. Ready for another pass

src/init/features/dataconnect/sdk.ts
maneesht: This is fine with me

src/dataconnect/fileUtils.spec.ts
```diff
@@ -127,81 +129,254 @@ describe("getPlatformFromFolder", () => {
       const platform = await getPlatformFromFolder(c.folderName);
       expect(platform).to.equal(c.output);
     });
-    afterEach(() => {
-      mockfs.restore();
-    });
   }
+  afterEach(() => {
```
maneesht, line 133: Should this be moved out? It seems valuable to keep it where it was before and reset the mocks in between tests

maneesht: Ah, actually, this seems correct.

src/init/features/dataconnect/sdk.ts
```diff
@@ -104,103 +101,89 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
...
-    const pkg =
-      newConnectorYaml.generate.javascriptSdk?.package ??
-      `@firebasegen/${connectorInfo.connectorYaml.connectorId}`;
-
+    const pkg = `${connectorYaml.connectorId}-connector`;
     const javascriptSdk: JavascriptSDK = {
-      outputDir,
-      package: pkg,
-      packageJsonDir: newConnectorYaml.generate.javascriptSdk?.packageJsonDir,
+      outputDir: path.relative(connectorDir, path.join(appDir, `dataconnect-generated/js/${pkg}`)),
```
maneesht, line 147: This seems different from the swiftSdk `outputDir`

fredzqm: Aim for the same pattern `dataconnect-generated/{js,swift,kotlin}/${package}` for all platforms.

`${package}` is part of `outputDir` of `JS` and `Dart`, but implicitly included for `Swift` and `Kotlin`.

maneesht: Why would it be a breaking change? Isn't this just for when the user hasn't created a generated SDK for the platform yet?

src/dataconnect/fileUtils.spec.ts
```diff
@@ -127,81 +129,254 @@ describe("getPlatformFromFolder", () => {
...
+      appDir: string;
+      output: any;
+    }[] = [
+      {
+        desc: "basic",
+        appDir: appFolderBase,
+        output: {
+          outputDir: "../dataconnect-generated/js/default-connector",
+          package: "@firebasegen/default-connector",
+          packageJsonDir: "..",
```
maneesht, line 163: Wait, I think this should only add a `packageJsonDir` if one's detected...

maneesht (approved):

fredzqm:
Thanks for the detailed review!

PR #7798, opened 2024-10-07
Update text for SDK Configuration button

hlshen:
Screenshot 2024-10-07 at 1 38 47 PM

### DIFF
--- a/firebase-vscode/CHANGELOG.md
+++ b/firebase-vscode/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Updated internal firebase-tools dependency to 13.20.2
 - Updated default debug-log output to .firebase/logs directory
 - [Fixed] Fixed an issue where emulator startup would hang
+- Updated text for SDK Configuration button
 
 ## 0.10.1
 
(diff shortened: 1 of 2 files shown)


fredzqm:
What do folks think of "Add SDK to my app"?

yuchenshi (approved):
I'm okay with whatever wording

fredzqm:
https://github.com/search?q=repo%3Afirebase%2Ffirebase-tools+%22Configure+Generated+SDK%22&type=code

Do we want to update the code lens name as well?

hlshen:
I'm a bit indecisive. I feel that once you're in the yaml file, it makes sense to think of configuring settings. Thus "Configure Generated SDK" makes more sense.

joehan (approved):

fredzqm:
> I'm a bit indecisive. I feel that once you're in the yaml file, it makes sense to think of configuring settings. Thus "Configure Generated SDK" makes more sense. 

"Configure Generated SDK" makes sense to us, but not really for developers.

On the connector.yaml, maybe "Use Connector in the app".

Can come later. I think js and kotlin would work fine, but iOS needs additional steps.
CLI displays those messages. VS Code should do the same or we actually figure out how to import it in the app properly.

PR #7801, opened 2024-10-07
[FDC] Remove confusing prompts

fredzqm:
### Description

1. Remove prompt at the end of init dataconnect:sdk
2. If developer picked an existing empty service, don't ask them if they want to override local gql with templated files.

### Scenarios Tested

Pick up empty service. Don't override any GQL.
Screenshot 2024-10-07 at 12 13 44 PM

Generate SDK. Don't prompt any more.
Screenshot 2024-10-07 at 12 14 21 PM

### DIFF
--- a/src/init/features/dataconnect/index.ts
+++ b/src/init/features/dataconnect/index.ts
@@ -314,7 +314,6 @@ async function checkExistingInstances(
           });
         }
       }
-      console.error(info.schemaGql);
     } else {
       info = await promptForService(info);
     }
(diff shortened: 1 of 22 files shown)


src/init/features/dataconnect/index.ts
```diff
@@ -305,6 +314,7 @@ async function checkExistingInstances(
           });
         }
       }
+      console.error(info.schemaGql);
```
joehan, line 317:
```suggestion
```

fredzqm: Good catch

PR #7802, opened 2024-10-07
Fz/init flow

fredzqm:

### DIFF
--- a/src/init/features/dataconnect/index.ts
+++ b/src/init/features/dataconnect/index.ts
@@ -73,17 +73,9 @@ export async function doSetup(setup: Setup, config: Config): Promise<void> {
   if (cwdPlatformGuess !== Platform.UNDETERMINED) {
     await sdk.doSetup(setup, config);
   } else {
-    const promptForSDKGeneration = await confirm({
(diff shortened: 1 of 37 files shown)


src/init/features/dataconnect/index.ts
```diff
@@ -74,22 +74,29 @@ const defaultSchema = { path: "schema.gql", content: SCHEMA_TEMPLATE };
 
 // doSetup is split into 2 phases - ask questions and then actuate files and API calls based on those answers.
 export async function doSetup(setup: Setup, config: Config): Promise<void> {
-  const info = await askQuestions(setup);
+  const isBillingEnabled = setup.projectId ? await checkBillingEnabled(setup.projectId) : false;
```
joehan, line 77: Optional, canbe done later: Checking billing status is probably appropriate to do in init/index.ts and then passed around in setup, since it won't change from product to product.

joehan (approved):

src/init/features/dataconnect/index.ts
fredzqm: Already passed this into `askQuestions`.

Only called once

PR #7804, opened 2024-10-07
Use 'default' alias if there are multiple aliases

joehan:
### Description
Fixes a weird quirk of project selection behavior where we only respected the 'default' alias if there were no other aliases

### Scenarios Tested
Ran a command with 2 aliases in my rc, and confirmed that it correctly used 'default'


src/command.ts
```diff
@@ -334,25 +335,33 @@ export class Command {
...
+      : undefined;
+    options.project = options.project ?? activeProject;
     // support deprecated "firebase" key in firebase.json
     if (options.config && !options.project) {
       options.project = options.config.defaults.project;
     }
 
     const aliases = rc.projects;
-    const rcProject = get(aliases, options.project);
+    const rcProject = options.project ? aliases[options.project] : undefined;
```
fredzqm, line 351: What does rcProject stands for?

fredzqm (approved):

PR #7806, opened 2024-10-08
Updated requireAuth to include getting the access token

maneesht:
Made sure that the access token is fetched when the user is in the `MONOSPACE_ENV` path

### DIFF
--- a/src/requireAuth.ts
+++ b/src/requireAuth.ts
@@ -7,7 +7,7 @@ import { FirebaseError } from "./error";
 import { logger } from "./logger";
 import * as utils from "./utils";
 import * as scopes from "./scopes";
-import { Tokens, User } from "./types/auth";
+import { Tokens, TokensWithExpiration, User } from "./types/auth";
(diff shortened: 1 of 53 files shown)


src/requireAuth.ts
```diff
@@ -53,9 +53,18 @@ async function autoAuth(options: Options, authScopes: string[]): Promise<void |
...
     setActiveAccount(options, {
       user: { email: clientEmail },
-      tokens: { access_token: token },
+      tokens: {
+        access_token: token,
+        expires_at: client.cachedCredential?.credentials.expiry_date,
+      } as TokensWithExpiration, // We need the refresh_token here.
+    });
+    setGlobalDefaultAccount({
+      user: { email: clientEmail },
```
joehan, line 62: Nit: This object is duplicated between the calls to setActive and setGlobal. Consider declaring it once as a variable and reusing.

joehan (approved):
LGTM with one style fix

PR #7810, opened 2024-10-08
Always include the default template the data connect

fredzqm:
Pick up new empty service from console.

From an empty folder
Screenshot 2024-10-08 at 1 36 05 PM

From a Firebase project already has `dataconnect`.
Screenshot 2024-10-08 at 1 37 14 PM


joehan (approved):
LGTM

PR #7811, opened 2024-10-08
Centralize logs when not in a firebase directory

joehan:
### Description
When not inside a firebase directory, write logs to `~/.cache/firebase/vscode`. `~/.cache/firebase` is already used by other places in the CLI.

### DIFF
--- a/firebase-vscode/src/logger-wrapper.ts
+++ b/firebase-vscode/src/logger-wrapper.ts
@@ -47,8 +47,8 @@ export function logSetup() {
     // in the entire firebase.ts file
     const rootFolders = getRootFolders();
     // Default to a central path, but write files to a local path if we're in a Firebase directory.
-    let filePath = "~/.cache/firebase/vscode/vsce-debug.log";
-    if (fs.existsSync(path.join(rootFolders[0], "firebase.json")) { 
+    let filePath = path.join("~", ".cache", "firebase", "logs", "vsce-debug.log");
+    if (fs.existsSync(path.join(rootFolders[0], "firebase.json"))) { 
       filePath = path.join(rootFolders[0], ".firebase", "logs", "vsce-debug.log");
     }
     pluginLogger.info("Logging to path", filePath);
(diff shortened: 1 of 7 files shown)


firebase-vscode/src/logger-wrapper.ts
```diff
@@ -45,7 +46,11 @@ export function logSetup() {
     // Re-implement file logger call from ../../src/bin/firebase.ts to not bring
     // in the entire firebase.ts file
     const rootFolders = getRootFolders();
-    const filePath = path.join(rootFolders[0], ".firebase", "logs", "vsce-debug.log");
+    // Default to a central path, but write files to a local path if we're in a Firebase directory.
+    let filePath = "~/.cache/firebase/vscode/vsce-debug.log";
```
yuchenshi, line 50:
```suggestion
    let filePath = "~/.cache/firebase/logs/vsce-debug.log";
```

And we also need a random segment to prevent conflicts across multiple windows.

And speaking of windows... Does it work on Windows (capital W)?

yuchenshi: And... one last thing, please use `require('os').homedir()` instead of `~` for portability

firebase-vscode/src/logger-wrapper.ts
```diff
@@ -45,7 +46,11 @@ export function logSetup() {
     // Re-implement file logger call from ../../src/bin/firebase.ts to not bring
     // in the entire firebase.ts file
     const rootFolders = getRootFolders();
-    const filePath = path.join(rootFolders[0], ".firebase", "logs", "vsce-debug.log");
+    // Default to a central path, but write files to a local path if we're in a Firebase directory.
+    let filePath = path.join("~", ".cache", "firebase", "logs", "vsce-debug.log");
```
hlshen, line 50: Is it more confusing to have two locations? 

The only real use case of having the logs without firebase.json is for firebase init.

I'm wondering if we can log to the central location always, to not lose the logs after firebase init.

joehan: I like keeping the .log files located in the project directory when possible - we already write the firebase-debug.log and <emulator>-debug.log's there

firebase-vscode/src/logger-wrapper.ts
joehan: Good calls, done. Ignoring the random segment bit for now - different instances will write to the same log, which is fine for now imo.

hlshen (approved):

PR #7826, opened 2024-10-11
Replace rimraf dependencies by built-in Node APIs

echedaleux:
### Description

This PR removes the rimraf package and replaces it with the native fs.rmSync method from Node.js, as part of the [Ecosystem Performance](https://e18e.dev/guide/) initiative ([more details about rimraf here](https://github.com/es-tooling/module-replacements/blob/main/docs/modules/rimraf.md)). This change simplifies the codebase and eliminates the need for an external package, aligning with performance and sustainability goals.
- Updated clean script to use fs.rmSync
- Removed rimraf from dependencies


joehan (approved):
Thanks a ton for this contribution! LGTM

PR #7831, opened 2024-10-14
Apphosting emulator - local secrets support via `apphosting:config:export` command

mathu97:
Adds a new command `apphosting:config:export` to export secrets (and other configurations in the future).

This new command reads from the chosen apphosting files present in the user's project codebase, hits Google Secret Manager to get the values of theses secrets, and writes them to the `apphosting.local.yaml` file. The `apphosting.local.yaml` file is automatically loaded into the app hosting emulator when `firebase emulators:start` is run.


src/apphosting/config.ts
```diff
@@ -32,16 +36,21 @@ export interface Config {
...
+}
+
 /**
  * Finds the path of apphosting.yaml.
  * Starts with cwd and walks up the tree until apphosting.yaml is found or
  * we find the project root (where firebase.json is) or the filesystem root;
  * in these cases, returns null.
  */
-export function yamlPath(cwd: string): string | null {
+export function yamlPath(cwd: string, yamlFileName: string): string | null {
```
mathu97, line 50:
```suggestion
export function yamlPath(cwd: string, fileName: string): string | null {
```

mathu97: add tests here

src/apphosting/config.ts
```diff
@@ -54,7 +63,41 @@ export function yamlPath(cwd: string): string | null {
     }
     dir = parent;
   }
-  return resolve(dir, "apphosting.yaml");
+  return resolve(dir, APPHOSTING_BASE_YAML_FILE);
+}
+
+export function allYamlPaths(cwd: string): string[] | null {
```
mathu97, line 69: Add jsdoc

mathu97: Add unit tests

src/apphosting/config.ts
```diff
@@ -54,7 +63,41 @@ export function yamlPath(cwd: string): string | null {
...
+    if (parent === dir) {
+      break;
+    }
+    dir = parent;
+  }
+
+  return files.length > 0 ? files : null;
+}
+
+function listAppHostingYamlsInCWD(cwd: string): string[] {
```
mathu97, line 93: add jsdoc + unit tests

src/apphosting/secrets/index.ts
```diff
@@ -164,3 +167,73 @@ export async function upsertSecret(
   // This may not be a great idea until version control is actually implemented.
   return false;
 }
+
+export async function getAppHostingConfigToExport(
```
mathu97, line 171: add jsdoc + unit tests

src/apphosting/secrets/index.ts
```diff
@@ -164,3 +167,73 @@ export async function upsertSecret(
...
+    name: "apphosting-yaml",
+    type: "list",
+    message: "Which environment would you like to export secrets from Secret Manager for?",
+    choices: listOptions,
+  });
+
+  return fileToExportPath;
+}
+
+function getEnvironmentNameFromYamlFileName(fileName: string): string {
```
mathu97, line 230: jsdoc + unit tests

src/apphosting/secrets/index.ts
```diff
@@ -164,3 +167,73 @@ export async function upsertSecret(
...
+    config = {
+      ...config,
+      ...baseConfig,
+    };
+  }
+
+  return config;
+}
+
+async function promptForAppHostingFileToExportSecretsFrom(fileNameToPathMap: Map<string, string>) {
```
mathu97, line 198: jsdoc + unit tests

src/apphosting/utils.ts
```diff
@@ -0,0 +1,30 @@
+import { readFileFromDirectory, wrappedSafeLoad } from "../utils";
+import { Config as AppHostingYaml, AppHostingReadableConfiguration } from "./config";
+
+/**
+ * Reads an apphosting.*.yaml file, parses, and returns in an easy to use
+ * format.
+ */
+export async function loadAppHostingYaml(
```
mathu97, line 8: unit tests

src/commands/apphosting-config-export.ts
```diff
@@ -0,0 +1,64 @@
...
+  )
+  .before(requireAuth)
+  .before(secretManager.ensureApi)
+  .before(requirePermissions, ["secretmanager.versions.access"])
+  .action(async (options: Options) => {
+    const projectId = needProjectId(options);
+    const currentDir = process.cwd();
+
+    // Get all apphosting yaml files ignoring the apphosting.local.yaml file
+    const yamlFilePaths = allYamlPaths(currentDir)?.filter(
```
tonyjhuang, line 29: where are we using the value of the --secrets flag?

src/commands/apphosting-config-export.ts
```diff
@@ -0,0 +1,64 @@
...
+      (path) => !path.endsWith(APPHOSTING_LOCAL_YAML),
+    );
+
+    if (!yamlFilePaths) {
+      logger.warn("No apphosting YAMLS found");
```
tonyjhuang, line 34: how about "No apphosting.yaml found"

src/commands/apphosting-config-export.ts
```diff
@@ -0,0 +1,64 @@
...
+      return;
+    }
+
+    // Load apphosting.local.yaml file if it exists. Secrets should be added to the env list in this object and written back to the apphosting.local.yaml
+    const apphostingLocalConfigPath = yamlPath(currentDir, APPHOSTING_LOCAL_YAML);
+    const localAppHostingConfig = await loadAppHostingYaml(apphostingLocalConfigPath ?? undefined);
```
tonyjhuang, line 40: smallest of nits but why are we swapping the order of these variable parts here?

apphosting local config path
vs
local apphosting config

src/commands/apphosting-config-export.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const configsToExport = await getConfigToExport(yamlFilePaths);
```
tonyjhuang, line 42: nit: singular `configToExport`

src/commands/apphosting-config-export.ts
```diff
@@ -0,0 +1,64 @@
...
+    const secretsToExport = configsToExport.secrets;
+    if (!secretsToExport) {
+      logger.warn("No secrets found to export in the choosen apphosting files");
+      return;
+    }
+
+    const secretsToInjectAsEnvs = await fetchSecrets(projectId, secretsToExport);
```
tonyjhuang, line 49: optional naming: secretMaterial

src/commands/apphosting-config-export.ts
```diff
@@ -0,0 +1,64 @@
...
+        availability: ["RUNTIME"],
+      });
+    }
+
+    // update apphosting.local.yaml
+    localAppHostingConfig.writeToFile(
+      apphostingLocalConfigPath ?? join(currentDir, APPHOSTING_LOCAL_YAML),
+    );
+
+    logger.info(`Wrote Secrets as environment variables to ${APPHOSTING_LOCAL_YAML}.`);
```
tonyjhuang, line 63: nit: "secrets"

src/emulator/apphosting/config.ts
```diff
@@ -1,75 +1,31 @@
...
-import { readFileFromDirectory, wrappedSafeLoad } from "../../utils";
 import { logger } from "./utils";
 import { Emulators } from "../types";
+import { APPHOSTING_BASE_YAML_FILE } from "../../apphosting/config";
+import { AppHostingYamlConfig, loadAppHostingYaml } from "../../apphosting/yaml";
 
-type EnvironmentAvailability = "BUILD" | "RUNTIME";
-
-const APPHOSTING_YAML = "apphosting.yaml";
 const APPHOSTING_LOCAL_YAML = "apphosting.local.yaml";
```
tonyjhuang, line 8: duplicated?

src/emulator/apphosting/config.ts
```diff
@@ -1,75 +1,31 @@
...
-  }
-
-  return { environmentVariables, secrets };
-}
-
 /**
  * Loads in apphosting.yaml & apphosting.local.yaml, giving
  * apphosting.local.yaml precedence if present.
  */
 export async function getLocalAppHostingConfiguration(
```
tonyjhuang, line 14: small suggestion: if we make this environment-agnostic, this will help us provide apphosting linters for environments much more easily in the future, eg

getAppHostingConfiguration(environment)  // you can pass in 'staging' or 'local' for example

src/emulator/apphosting/config.ts
```diff
@@ -78,18 +34,17 @@ export async function getLocalAppHostingConfiguration(
...
-      ...apphostingBaseConfig.environmentVariables,
-      ...apphostingLocalConfig.environmentVariables,
-    },
-    secrets: {
-      ...apphostingBaseConfig.secrets,
-      ...apphostingLocalConfig.secrets,
-    },
-  };
+  if (!mainConfig) {
+    return await loadAppHostingYaml();
```
tonyjhuang, line 46: no arg function call?

src/emulator/apphosting/serve.ts
```diff
@@ -39,8 +39,9 @@ export async function start(options?: StartOptions): Promise<{ hostname: string;
 async function serve(port: number, startCommand?: string): Promise<void> {
   const rootDir = process.cwd();
   const apphostingLocalConfig = await getLocalAppHostingConfiguration(rootDir);
+
   const environmentVariablesToInject = {
-    ...apphostingLocalConfig.environmentVariables,
+    ...apphostingLocalConfig.environmentVariablesAsRecord(),
```
tonyjhuang, line 44: why as Record?

tonyjhuang: This seems like a unique enough use case that I would advocate for doing the mapping to Record type here as opposed to lifting it to the class definition.

src/emulator/apphosting/config.ts
```diff
@@ -1,75 +1,31 @@
...
       "SUCCESS",
       Emulators.APPHOSTING,
-      `${APPHOSTING_YAML} found, loading configuration`,
+      `${APPHOSTING_BASE_YAML_FILE} found, loading configuration`,
     );
-    apphostingBaseConfig = await loadAppHostingYaml(sourceDirectory, APPHOSTING_YAML);
+
+    mainConfig = await loadAppHostingYaml(join(sourceDirectory, APPHOSTING_BASE_YAML_FILE));
+    console.log(`${JSON.stringify(mainConfig._loadedAppHostingYaml)}`);
+    console.log(`main config: ${JSON.stringify(mainConfig.environmentVariables)}`);
```
tonyjhuang, line 28: remove

src/emulator/apphosting/config.ts
```diff
@@ -1,75 +1,31 @@
...
  * Loads in apphosting.yaml & apphosting.local.yaml, giving
  * apphosting.local.yaml precedence if present.
  */
 export async function getLocalAppHostingConfiguration(
   sourceDirectory: string,
-): Promise<AppHostingConfiguration> {
-  let apphostingBaseConfig: AppHostingConfiguration = {};
-  let apphostingLocalConfig: AppHostingConfiguration = {};
+): Promise<AppHostingYamlConfig> {
+  let mainConfig: AppHostingYamlConfig | undefined;
```
tonyjhuang, line 17: optional nit: you can simplify some of the logic in this function if you init this to a default/empty AppHostingYamlConfig.

config = loadBaseConfig() || emptyConfig();
config.merge(loadLocalConfig());
return config

src/apphosting/yaml.ts
```diff
@@ -0,0 +1,133 @@
...
+  private _secrets: Map<string, Secret>;
+
+  constructor(filePath?: string) {
+    this.filePath = filePath;
+    this._loadedAppHostingYaml = {};
+    this._environmentVariables = new Map();
+    this._secrets = new Map();
+  }
+
+  async init() {
```
tonyjhuang, line 31: optional but I like to avoid introducing state into my data objects. Not knowing whether an object is fully "initialized" or not leads to awkward code like the helper later in this file.

The danger here is someone new to the codebase uses this constructor without calling `init` after, leading to all sorts of fun behavior.

How about a factory method?

export class ... {
  // Loads a config from file
  static loadFromFile(filePath): Promise<AppHostingYamlConfig> {..}
  
  // Creates an empty config
  static empty(): AppHostingYamlConfig {..}
 
  /* use one of the factory methods above */
  private constructor(..) {..}
}

src/apphosting/yaml.ts
```diff
@@ -0,0 +1,133 @@
...
+    return newObject;
+  }
+
+  addEnvironmentVariable(env: EnvironmentVariable) {
+    this._environmentVariables.set(env.variable, env);
+  }
+
+  addSecret(secret: Secret) {
+    this._secrets.set(secret.variable, secret);
+  }
```
tonyjhuang, line 68: thoughts on getting rid of these and letting clients access and mutate the fields directly?

config.secrets.set(key, val)

doesn't seem too bad to me

src/apphosting/secrets/index.ts
```diff
@@ -164,3 +168,115 @@ export async function upsertSecret(
...
+export async function getConfigToExport(
+  allAppHostingYamlPaths: string[],
+): Promise<AppHostingYamlConfig> {
+  const fileNameToPathMap: Map<string, string> = new Map();
+  for (const path of allAppHostingYamlPaths) {
+    const fileName = basename(path);
+    fileNameToPathMap.set(fileName, path);
+  }
+  const baseFilePath = fileNameToPathMap.get(APPHOSTING_BASE_YAML_FILE);
+  const appHostingfileToExportPath = await promptForAppHostingYaml(fileNameToPathMap);
```
tonyjhuang, line 230: we need to account for the case where the user specified `export --secrets apphosting.staging.yaml`

tonyjhuang (changes requested):

src/apphosting/config.ts
```diff
@@ -54,7 +57,51 @@ export function yamlPath(cwd: string): string | null {
...
+  return files.length > 0 ? files : null;
+}
+
+/**
+ * Lists all apphosting.*.yaml files in the given directory.
+ */
+export function list(cwd: string): string[] {
+  const paths: string[] = [];
+  for (const file of fs.listFiles(cwd)) {
+    if (file.startsWith("apphosting.") && file.endsWith(".yaml")) {
```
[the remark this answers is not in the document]
mathu97: is that file name invalid? Can we not have environments named "foo.bar"?

src/apphosting/config.ts
```diff
@@ -54,7 +57,51 @@ export function yamlPath(cwd: string): string | null {
...
+ * This function starts at the provided directory (`cwd`) and traverses
+ * upwards through the file system until it finds a `firebase.json` file
+ * (indicating the project root) or reaches the root of the filesystem.
+ * Along the way, it collects the paths of all encountered `apphosting.*.yaml` files.
+ *
+ * @param cwd The directory to start the search from.
+ * @returns An array of strings representing the paths to all found `apphosting.*.yaml` files,
+ *          or `null` if no such files are found.
+ */
+export function allYamlPaths(cwd: string): string[] | null {
```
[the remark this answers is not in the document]
mathu97: +1, what about `discoverConfigsInProject`? as its also checking in the current directory as well

src/commands/apphosting-config-export.ts
mathu97: missed this, thanks for catching it!

src/commands/apphosting-config-export.ts
mathu97: no reason, makes more sense to keep them consistent!

src/apphosting/secrets/index.ts
mathu97: updated: implemented logic for this.

src/apphosting/yaml.ts
mathu97: totally makes sense, refactored it to use the Factory pattern

src/apphosting/yaml.ts
mathu97: I like that the setters hide implementation, and the getters do a transformation so that the values are in `Record<string, string>[]` format which is easier to use by clients, compared to a Map.

src/apphosting/secrets/index.ts
```diff
@@ -215,25 +215,33 @@ export async function fetchSecrets(
  * to ensure the exported secrets accurately reflect the chosen environment.
  *
  * @param allAppHostingYamlPaths An array of paths to all available `apphosting.*.yaml` files.
- * @returns A Promise that resolves to an `AppHostingYamlConfig` object representing the combined
```
tonyjhuang: To be clear, you can still say

@returns the final merged config to export secrets from

I was mainly suggesting you remove the type information

src/apphosting/secrets/index.ts
```diff
@@ -215,25 +215,33 @@ export async function fetchSecrets(
...
- *          configuration to be used for exporting configurations.
  */
 export async function getConfigToExport(
```
tonyjhuang, line 219: I wonder if this should be a helper method in config.ts, ie

GetConfigForEnvironment(...);

This type of logic seems broadly useful for non-secret related workflows

src/apphosting/yaml.ts
```diff
@@ -16,47 +17,42 @@ export type Secret = Omit<Env, "availability" | "value">;
...
-  constructor(filePath?: string) {
-    this.filePath = filePath;
-    this._loadedAppHostingYaml = {};
-    this._environmentVariables = new Map();
-    this._secrets = new Map();
-  }
-
-  async init() {
-    // If the file doesn't exist don't load it in, instead it will be used to write to later
+  static async loadFromFile(filePath: string): Promise<AppHostingYamlConfig> {
```
tonyjhuang, line 23: jsdocs for public methods

src/apphosting/yaml.ts
```diff
@@ -16,47 +17,42 @@ export type Secret = Omit<Env, "availability" | "value">;
...
+      config._environmentVariables = parsedEnvs.environmentVariables;
+      config._secrets = parsedEnvs.secrets;
     }
+
+    return config;
   }
 
-  get environmentVariables(): EnvironmentVariable[] {
-    return this.mapToEnv(this._environmentVariables);
+  static empty() {
```
tonyjhuang, line 41: jsdoc pls

src/apphosting/yaml.ts
```diff
@@ -81,53 +77,44 @@ export class AppHostingYamlConfig {
...
-  const apphostingConfig = new AppHostingYamlConfig(filePath);
-  await apphostingConfig.init();
+  return {
+    environmentVariables,
+    secrets,
+  };
+}
 
-  return apphostingConfig;
+function mapToEnv(map: Map<string, Env>): Env[] {
```
tonyjhuang, line 118: naming nit: mapToArray?

src/apphosting/yaml.ts
```diff
@@ -81,53 +77,44 @@ export class AppHostingYamlConfig {
...
   }
 
-  writeToFile(customFilePath?: string) {
-    if (!this.filePath && !customFilePath) {
-      throw Error("No apphosting yaml file path provided to write to");
+  /**
+   * Loads the given file if it exists and updates it. If
+   * it does not exist a new file will be created.
+   */
+  async writeToFile(filePath: string) {
```
tonyjhuang, line 84: naming nit: this feels somewhat surprising. I would rename this to something like `upsertFile`, `write` _usually_ indicates a full overwrite

src/commands/apphosting-config-export.ts
```diff
@@ -26,28 +30,36 @@ export const command = new Command("apphosting:config:export")
...
-    const apphostingLocalConfigPath = yamlPath(currentDir, APPHOSTING_LOCAL_YAML);
-    const localAppHostingConfig = await loadAppHostingYaml(apphostingLocalConfigPath ?? undefined);
+    const localApphostingConfigPath = yamlPath(currentDir, APPHOSTING_LOCAL_YAML_FILE);
+    let localAppHostingConfig: AppHostingYamlConfig = AppHostingYamlConfig.empty();
 
-    const configsToExport = await getConfigToExport(yamlFilePaths);
-    const secretsToExport = configsToExport.secrets;
+    if (localApphostingConfigPath) {
+      localAppHostingConfig = await AppHostingYamlConfig.loadFromFile(localApphostingConfigPath);
+    }
```
tonyjhuang, line 48: nit: try to declare variables close to where they're used. 

go/java-style#s4.8.2.2-variables-limited-scope

For example:

```
var a;
<15 lines of code>
a += 1
```

```
<15 lines of code>
var a;
a += 1;
```

In the second example, it's much easier to see there's a bug

src/commands/apphosting-config-export.ts
```diff
@@ -26,28 +30,36 @@ export const command = new Command("apphosting:config:export")
...
+
+    const configToExport = await getConfigToExport(
+      yamlFilePaths,
+      options.secrets as string | undefined,
+    );
```
tonyjhuang, line 53: super small nit: Is there a way to make this more self documenting?

getConfigToExport(yamlFilePaths, options.secrets) 

doesn't really mean anything to me. Alternatively, this means a lot

loadConfigForEnvironment("staging")

I'd recommend taking a pass through this method and seeing if we can make some of this easier to parse at first glance

tonyjhuang (changes requested):

src/apphosting/config.ts
```diff
@@ -36,15 +37,15 @@ export interface Config {
...
 /**
- * Finds the path of apphosting.yaml.
- * Starts with cwd and walks up the tree until apphosting.yaml is found or
+ * Finds the path of a yaml file.
+ * Starts with cwd and walks up the tree until yamlFileName is found or
  * we find the project root (where firebase.json is) or the filesystem root;
  * in these cases, returns null.
  */
-export function yamlPath(cwd: string, fileName: string): string | null {
+export function yamlPath(cwd: string, yamlFileName: string): string | null {
```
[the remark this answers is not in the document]
mathu97: No it doesn't have to be a yaml file, and I agree that the function should be renamed. Also now that you mention it I don't think this function was intended to search the entire project. Its currently being used by the `apphosting:secrets:*` commands, which i'm guessing so that the command still works when it's run in some subdirectory of the project? i.e `apphosting:secrets:set` should work in project root or in some subdirectory of the project as well.

src/apphosting/config.ts
```diff
@@ -36,15 +37,15 @@ export interface Config {
 }
 
 /**
- * Finds the path of apphosting.yaml.
- * Starts with cwd and walks up the tree until apphosting.yaml is found or
+ * Finds the path of a yaml file.
```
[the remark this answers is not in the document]
mathu97: you mean absolute path right? I'll add a comment

src/apphosting/config.ts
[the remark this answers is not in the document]
mathu97: Refactored it a little bit to return the root directory of the apphosting project instead. This makes it easier to be re-used and is more readable.

src/apphosting/config.ts
```diff
@@ -33,15 +38,15 @@ export interface Config {
 }
 
 /**
- * Finds the path of apphosting.yaml.
- * Starts with cwd and walks up the tree until apphosting.yaml is found or
- * we find the project root (where firebase.json is) or the filesystem root;
- * in these cases, returns null.
+ * Finds the project root for apphosting backends.
+ * Starts with cwd and walks up the path until a apphosting.yaml file is found
```
tonyjhuang, line 42: grammar nit: an apphosting.yaml

src/apphosting/config.ts
```diff
@@ -33,15 +38,15 @@ export interface Config {
...
+ *
+ * Eample path that's returned: "/home/my-project"
```
tonyjhuang, line 44: nit: Example

src/apphosting/config.ts
```diff
@@ -33,15 +38,15 @@ export interface Config {
...
  */
-export function yamlPath(cwd: string): string | null {
+export function discoverProjectRoot(cwd: string): string | null {
```
tonyjhuang, line 46: projectRoot or backendRoot? is firebase.json guaranteed to be colocated with apphosting.yaml?

src/apphosting/config.ts
```diff
@@ -54,7 +59,41 @@ export function yamlPath(cwd: string): string | null {
...
+    if (configsInCwd.length === 0) {
+      return null;
+    }
+
+    return configsInCwd;
+  }
+
+  const configsAtRoot = listAppHostingFilesInPath(projectRoot);
+  if (configsAtRoot.length === 0) {
+    return null;
```
tonyjhuang, line 83: why null and not an empty list?

src/apphosting/secrets/index.ts
```diff
@@ -164,3 +174,88 @@ export async function upsertSecret(
...
+  let environmentConfigFilePath = baseFilePath;
+  if (!environmentConfigFile) {
+    // If file is not provided, prompt the user
+    environmentConfigFilePath = await promptForAppHostingYaml(
+      fileNameToPathMap,
+      "Which environment would you like to export secrets from Secret Manager for?",
+    );
+  }
+
+  return await loadConfigForEnvironment(environmentConfigFilePath, baseFilePath);
```
tonyjhuang, line 246: optional suggestion: This logic is still a little bit confusing to me. There's a lot of mixed conditionals and different handling of the base config depending on the inputs, along with aliasing vars (L237) that makes it a bit hard to understand what's going on.

This is very optional but it seems to me that there are two cases we need to handle here: 
1. User passed an explicit config file
2. User did not and we need to prompt.

My intuition says branching this function earlier may make this clearer, for example:

```
function(userGivenConfigFile) {
  allConfigs = discoverConfigs()

  // Get path for user-specified config file
  if (userGivenConfigFile) {
    userGivenConfigFilePath = allConfigs[userGivenConfigFile]
  } else {
    userGivenConfigFilePath = promptForConfig(allConfigs);   
  }
  
  // Load the config with any appropriate overrides
  if (userGivenConfigFile == BASE_YAML_FILE) {
    return AppHostingConfig.loadFile(BASE_YAML_FILE);
  }
  baseYamlFilePath = allConfigs[BASE_YAML_FILE];
  return loadConfigForEnvironment(userGivenConfigFile, baseYamlFilePath);
  }
}
```

src/apphosting/yaml.ts
```diff
@@ -0,0 +1,129 @@
...
+import * as yaml from "yaml";
+import * as jsYaml from "js-yaml";
+import { fileExistsSync } from "../fsutils";
+import { FirebaseError } from "../error";
+
+export type EnvironmentVariable = Omit<Env, "secret">;
+export type Secret = Omit<Env, "availability" | "value">;
+
+/**
+ * AppHostingYamlConfig is an object representing an Apphosting Yaml configuration
```
tonyjhuang, line 13: nit: representing an apphosting.yaml configuration

src/apphosting/yaml.ts
```diff
@@ -0,0 +1,129 @@
...
+ * present in the user's codebase (i.e 'apphosting.yaml', 'apphosting.staging.yaml', etc).
+ *
+ * This class is used to abstract away the parsing and handling logic of these
```
tonyjhuang, line 16: imo you can drop this

src/apphosting/yaml.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  addSecret(secret: Secret) {
+    this._secrets.set(secret.variable, secret);
+  }
+
+  /**
+   * Merges this AppHostingYamlConfig with another config, the incoming config
+   * has precedence if there are any conflicting configurations.
+   * */
+  merge(anotherAppHostingYamlConfig: AppHostingYamlConfig) {
```
tonyjhuang, line 79: optional: anotherAppHostingYamlConfig -> other

src/commands/apphosting-config-export.ts
```diff
@@ -0,0 +1,64 @@
...
+      localAppHostingConfigPath = resolve(projectRoot, APPHOSTING_LOCAL_YAML_FILE);
+      if (fs.fileExistsSync(localAppHostingConfigPath)) {
+        localAppHostingConfig = await AppHostingYamlConfig.loadFromFile(localAppHostingConfigPath);
+      }
+    }
+
+    const configToExport = await loadConfigToExport(cwd, environmentConfigFile);
+    const secretsToExport = configToExport.secrets;
+    if (!secretsToExport) {
+      logger.warn("No secrets found to export in the choosen apphosting files");
```
tonyjhuang, line 45: nit: chosen

Also lets align on App Hosting or `apphosting.yaml`. "apphosting" isn't really a thing

src/commands/apphosting-config-export.ts
```diff
@@ -0,0 +1,64 @@
...
+      localAppHostingConfig.addEnvironmentVariable({
+        variable: key,
+        value: value,
+        availability: ["RUNTIME"],
+      });
+    }
+
+    // update apphosting.local.yaml
+    localAppHostingConfig.upsertFile(
+      localAppHostingConfigPath ?? join(cwd, APPHOSTING_LOCAL_YAML_FILE),
```
tonyjhuang, line 60: should this not upsert at the project root?

src/emulator/apphosting/config.ts
```diff
@@ -1,95 +1,44 @@
...
       Emulators.APPHOSTING,
-      `${APPHOSTING_LOCAL_YAML} found, loading configuration`,
+      `${APPHOSTING_LOCAL_YAML_FILE} found, loading configuration`,
+    );
+
+    const localConfig = await AppHostingYamlConfig.loadFromFile(
+      join(sourceDirectory, APPHOSTING_LOCAL_YAML_FILE),
     );
-    apphostingLocalConfig = await loadAppHostingYaml(sourceDirectory, APPHOSTING_LOCAL_YAML);
+    config.merge(localConfig);
```
tonyjhuang, line 40: duplicated with above?

tonyjhuang (changes requested):

src/apphosting/config.ts
mathu97: yes backend root makes more sense. I don't think firebase.json needs to be colocated wiht apphosting.yaml. We can assume the backend root is wherever apphosting.yaml is.

src/apphosting/config.ts
mathu97: It just makes call site cleaner so we don't need to check for empty array there. But looking at this again I can see call sites not knowing this and checking for both null and empty array. Will just go with empty array.

src/commands/apphosting-config-export.ts
mathu97: yep, I was a not sure what to do if the root could not be found (i.e no `apphosting.yaml` is present). This implementation lets them still export secrets even if they only had an environment specific app hosting config file, but I think it makes sense to force users to initialize their apphosing.yaml file and expect it to be there if they're using apphosting.

src/apphosting/secrets/index.ts
```diff
@@ -164,3 +174,88 @@ export async function upsertSecret(
...
+  projectId: string,
+  secrets: Secret[],
+): Promise<Map<string, string>> {
+  let secretsKeyValuePairs: Map<string, string>;
+
+  try {
+    const secretPromises: Promise<[string, string]>[] = secrets.map(async (secretConfig) => {
+      const [name, version] = getSecretNameParts(secretConfig.secret!);
+
+      const value = await gcsm.accessSecretVersion(projectId, name, version);
```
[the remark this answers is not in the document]
mathu97: Although the secret fetching line is blocking, it doesn't block the other callbacks from running. So this map should be fetching secrets concurrently.

src/apphosting/secrets/index.ts
```diff
@@ -164,3 +174,88 @@ export async function upsertSecret(
...
+
+  return await loadConfigForEnvironment(environmentConfigFilePath, baseFilePath);
+}
+
+/**
+ * secret expected to be in format "myApiKeySecret@5",
+ * "projects/test-project/secrets/secretID", or
+ * "projects/test-project/secrets/secretID/versions/5"
+ */
+export function getSecretNameParts(secret: string): [string, string] {
```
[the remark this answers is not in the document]
mathu97: you mean in the [AppHostingConfigYaml class](https://github.com/firebase/firebase-tools/pull/7831/files#diff-a6d463faf3917b2ef694e5d72c1d740ca6e081c033a3eafc47effdeeaeca7c1fR63)? Having it here will let us re-use it in other places like the [`secrets:access` command](https://github.com/firebase/firebase-tools/pull/7831/files#diff-a262be25868e90d5a5e44fe68848d3772c7b53985194cbe9c4317e818f88609bR18)

src/apphosting/secrets/index.ts
mathu97: makes sense to me, this is definitely more clear!

src/commands/apphosting-config-export.ts
tonyjhuang: I understand now, it upserts at the original location or writes a new apphosting.local.yaml file at cwd

I think this is an acceptable fall back. Maybe we can make this clear in the console output, eg 

"Exported to local config at /full/path/to/existing/apphosting.local.yaml."

vs

"
Exported to local config at cwd/apphosting.local.yaml.
Make sure this file is stored at your backend root folder (usually next to apphosting.yaml and firebase.json).
"

src/apphosting/config.ts
```diff
@@ -64,26 +65,17 @@ export function discoverProjectRoot(cwd: string): string | null {
...
 
 /**
- * Lists absolute paths for `apphosting.*.yaml` configs at project root, if
- * a project root cannot be found it returns files from cwd.
+ * Lists absolute paths for `apphosting.*.yaml` configs at project root
  */
-export function discoverConfigsAtProjectRoot(cwd: string): string[] | null {
-  const projectRoot = discoverProjectRoot(cwd);
+export function discoverConfigsAtBackendRoot(cwd: string): string[] {
+  const projectRoot = discoverBackendRoot(cwd);
```
tonyjhuang: tiny nit: backendRoot

src/commands/apphosting-config-export.ts
```diff
@@ -29,20 +30,22 @@ export const command = new Command("apphosting:config:export")
 
     // Load apphosting.local.yaml file if it exists. Secrets should be added to the env list in this object and written back to the apphosting.local.yaml
     let localAppHostingConfig: AppHostingYamlConfig = AppHostingYamlConfig.empty();
-    const projectRoot = discoverProjectRoot(cwd);
-    let localAppHostingConfigPath: string | undefined;
+    const backendRoot = discoverBackendRoot(cwd);
+    if (!backendRoot) {
+      throw new FirebaseError(
+        "Missing apphosting.yaml: This command requires an apphosting.yaml configuration file. Please create one and try again.",
```
tonyjhuang, line 36: How about

Please run `firebase init` and try again.

src/emulator/apphosting/config.ts
```diff
@@ -1,44 +1,30 @@
...
  * apphosting.local.yaml precedence if present.
  */
-export async function getLocalAppHostingConfiguration(
-  sourceDirectory: string,
-): Promise<AppHostingYamlConfig> {
-  const config: AppHostingYamlConfig = AppHostingYamlConfig.empty();
+export async function getLocalAppHostingConfiguration(cwd: string): Promise<AppHostingYamlConfig> {
+  // Get all apphosting yaml files ignoring the apphosting.local.yaml file
+  const appHostingConfigPaths = discoverConfigsAtBackendRoot(cwd).filter(
+    (path) => !path.endsWith(APPHOSTING_LOCAL_YAML_FILE),
```
tonyjhuang: Seems like we're filtering out the local yaml file here and then trying to find it in in the map below (L28)?

tonyjhuang (changes requested):

src/apphosting/config.ts
mathu97: fixed in a follow up commit

src/apphosting/utils.ts
```diff
@@ -15,16 +15,16 @@ export function getEnvironmentName(apphostingYamlFileName: string): string {
...
  * (e.g: "apphosting.staging.yaml" => "/cwd/apphosting.staging.yaml"), this function
- * will prompt the user to choose an apphosting configuration. It returns the path
- * of the chosen apphosting configuration.
+ * will prompt the user to choose an app hosting configuration. It returns the path
+ * of the chosen app hosting configuration.
  */
 export async function promptForAppHostingYaml(
   apphostingFileNameToPathMap: Map<string, string>,
-  promptMessage = "Please select an apphosting config:",
+  promptMessage = "Please select an app hosting config:",
```
tonyjhuang, line 27: nit: App Hosting

src/apphosting/yaml.ts
```diff
@@ -18,9 +18,9 @@ export class AppHostingYamlConfig {
   private _secrets: Map<string, Secret>;
 
   /**
-   * Reads in the apphosting yaml file found in filePath, parses the secrets and
+   * Reads in the app hosting yaml file found in filePath, parses the secrets and
```
tonyjhuang, line 21: apphosting.yaml or App Hosting (yaml) config here and elsewhere

tonyjhuang (approved):
thanks for all the changes! 🔥🔥

PR #7832, opened 2024-10-14
Focus terminal for shell tasks.

yuchenshi:
### Description

Makes it easier to answer prompts and harder to hit the `init` button twice.

### Scenarios Tested

Not tested :(

### Sample Commands

Click the `firebase init`, deploy, and start emulator buttons in left panel.


joehan (approved):

yuchenshi:
Marking as draft since this is untested

firebase-vscode/src/data-connect/terminal.ts
```diff
@@ -54,15 +54,15 @@ export function runTerminalTask(
...
-        new vscode.ShellExecution(command, executionOptions),
-      ),
+    const task = new vscode.Task(
+      { type },
+      vscode.TaskScope.Workspace,
+      taskName,
+      "firebase",
+      new vscode.ShellExecution(command, executionOptions),
     );
+    task.presentationOptions = { focus: true };
```
yuchenshi, line 64: On a second thought, maybe we don't need focus for running the emulators? That command almost never prompts for anything interactively. Focusing the terminal for init and deploy makes sense though

hlshen: I agree, we can add it to the call signature

yuchenshi: Done

PR #7833, opened 2024-10-14
First pass at auto generating sdk configs

maneesht:
Fixed:
* Return type. Originally the object returned by `apps-sdkconfig` was typed as any, and didn't match the correct `AppConfigurationData` type. I returned the right object instead.
* When file name is not specified via `out` and the command attempted to get the default file name, the terminal would output `undefined` and error out. Now the correct output file is specified and used.
* Implemented `relativeTo` when prompting for directories
```bash
➜  dataconnect git:(mtewani/idx-updates) node ~/source/firebase-tools/lib/bin/firebase.js apps:init -o false

Error: Cannot autoconfigure an app because the experiment appsinit is not enabled. To enable appsinit run firebase experiments:enable appsinit

➜  app git:(mtewani/idx-updates) node ~/source/firebase-tools/lib/bin/firebase.js apps:init
✔  Detected WEB app in directory /Users/mtewani/source/quickstart-js/dataconnect/app
✔ Downloading configuration data for your Firebase WEB app
App configuration is written in firebase-js-config.json

How to use your JS SDK Config:
ES Module:
import { initializeApp } from 'firebase/app';
import json from './firebase-js-config.json';
initializeApp(json); // or copy and paste the config directly from the json file here
// CommonJS Module:
const { initializeApp } = require('firebase/app');
const json = require('./firebase-js-config.json');
initializeApp(json); // or copy and paste the config directly from the json file here
```


joehan (commented):
Some small stuff around typing & where helper functions live, but this is mostly LGTM!

src/commands/apps-create.ts
```diff
@@ -194,22 +194,25 @@ export const command = new Command("apps:create [platform] [displayName]")
...
-        default:
-          throw new FirebaseError("Unexpected error. This should not happen");
-      }
-
+      const appData = await sdkInit(appPlatform, options);
       logPostAppCreationInformation(appData, appPlatform);
       return appData;
     },
   );
+export async function sdkInit(appPlatform: AppPlatform, options: any) {
```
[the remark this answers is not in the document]
maneesht: Should we move everything into a new file? Also, we should put that info somewhere, I think other commands have other functions besides just the Command

src/commands/apps-sdkconfig.ts
```diff
@@ -1,56 +1,24 @@
...
+  writeConfigToFile,
 } from "../management/apps";
-import { needProjectId } from "../projectUtils";
-import { getOrPromptProject } from "../management/projects";
-import { FirebaseError } from "../error";
 import { requireAuth } from "../requireAuth";
 import { logger } from "../logger";
-import { promptOnce } from "../prompt";
 import { Options } from "../options";
+import * as path from "path";
```
joehan, line 18: Nit: prefer grouping this with the other external imports at the top of the file.

src/commands/apps-create.ts
```diff
@@ -72,93 +43,6 @@ function logPostAppCreationInformation(
...
-    return appData;
-  } catch (err: any) {
-    spinner.fail();
-    throw err;
-  }
-}
-
 export const command = new Command("apps:create [platform] [displayName]")
   .description(
     "create a new Firebase app. [platform] can be IOS, ANDROID or WEB (case insensitive).",
```
joehan, line 48: While you're here, could you add a type for options on this command as well?

joehan (approved):
Made a driveby type cleanup commit, and a few other nitty things, but LGTM!

package.json
```diff
@@ -100,6 +101,7 @@
   },
   "dependencies": {
     "@electric-sql/pglite": "^0.2.16",
+    "@firebasegen/default-connector": "file:dataconnect-generated/js/default-connector",
```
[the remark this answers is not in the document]
maneesht: It is! Nice catch

src/commands/apps-configure.ts
```diff
@@ -0,0 +1,112 @@
...
+import { requireAuth } from "../requireAuth";
+import { logger } from "../logger";
+import { Options } from "../options";
+import { needProjectId } from "../projectUtils";
+
+export interface AppsConfigureOptions extends Options {
+  out?: string | boolean;
+}
+
+function logUse(platform: AppPlatform, filePath: string) {
```
joehan, line 25:
```suggestion
function logUse(platform: AppPlatform, filePath: string): void {
```

joehan (approved):
LGTM after formatting, with 1 suggestion

package.json
```diff
@@ -28,6 +28,7 @@
     "test": "npm run lint:quiet && npm run test:compile && npm run mocha",
     "test:client-integration": "bash ./scripts/client-integration-tests/run.sh",
     "test:compile": "tsc --project tsconfig.compile.json",
+    "test:management": "npm run lint:quiet && npm run test:compile && nyc mocha 'src/management/*.spec.{ts,js}'",
```
joehan, line 31: This is probably unecessary since these unit tests are covered by `npm run mocha`. All these other e2e tests have special commands because they use code from `./scripts` and do things that are not appropriate for unit tests (ie calls to prod with credentials)

src/management/apps.spec.ts
```diff
@@ -1,7 +1,9 @@
+import * as mockfs from "mock-fs";
 import { expect } from "chai";
 import * as sinon from "sinon";
 import * as fs from "fs";
 import * as nock from "nock";
+import FileSystem from "mock-fs/lib/filesystem";
```
joehan, line 6: Missed this on the last pass, but its generally bad practice to dip into the lib of our imports. If this is not an exported type that you can get from mockfs, we should probably just use Record<string, any> instead in its place

joehan (commented):
Tiny things

PR #7836, opened 2024-10-15
Import + export support for the data connect emulator

joehan:
### Description
Adds support for data persistence between runs of the data conect emulator, as described in [go/fdc-emu-import-api](http://goto.google.com/fdc-emu-import-api)

### Scenarios Tested
- Set dataDir, ran a few CREATES, shut down the emulator, and verified they were still there on restart.
- Tested out `emulator:export` and confirmed that it outputs a tarball database dump in the export directory.
- Tested out `--import` and confirmed that the same data is there after import.
- Curled the clearData endpoint and confirmed that it clears out data from all tables, and that the emulator still works and can write data afterwards.

### Follow up work
- Add a Clear Data admin endpoint that empties out this directory, and give VSCode users an easy button to trigger it.
- E2E test for the data connect emulator that covers import/export


src/emulator/dataconnect/pgliteServer.ts
```diff
@@ -19,11 +21,16 @@ import { logger } from "../../logger";
...
   private database: string;
+  private dataDirectory?: string;
+  private importPath?: string;
 
-  public db: PGlite | undefined;
+  public db: PGlite | undefined = undefined;
   public async createPGServer(host: string = "127.0.0.1", port: number): Promise<net.Server> {
-    const db: PGlite = await this.getDb();
-    await db.waitReady;
+    const getDb: () => Promise<PGlite> = () => {
```
yuchenshi, line 29: nit: How about `this.getDb.bind(this)`?

src/emulator/dataconnect/pgliteServer.ts
```diff
@@ -50,41 +58,79 @@ export class PostgresServer {
...
+      const uuidOssp = (await dynamicImport("@electric-sql/pglite/contrib/uuid_ossp")).uuid_ossp;
+      const pgliteArgs: PGliteOptions = {
+        username: this.username,
+        database: this.database,
+        debug: 0,
+        extensions: {
+          vector,
+          uuidOssp,
+        },
+        // TODO:  Use dataDir + loadDataDir to implement import/export.
```
yuchenshi, line 85:
```suggestion
```

src/emulator/dataconnect/pgliteServer.ts
```diff
@@ -50,41 +58,79 @@ export class PostgresServer {
...
-      // TODO:  Use dataDir + loadDataDir to implement import/export.
-      // dataDir?: string;
-      // loadDataDir?: Blob | File;
-    });
+    return this.db;
+  }
+
+  public async clearDb(): Promise<void> {
+    // if (this.dataDirectory) {
+    //   rmSync(this.dataDirectory, { recursive: true });
```
yuchenshi, line 103: What are these commented lines for?

src/emulator/dataconnectEmulator.ts
```diff
@@ -102,14 +104,21 @@ export class DataConnectEmulator implements EmulatorInstance {
           `FIREBASE_DATACONNECT_POSTGRESQL_STRING is set to ${clc.bold(connStr)} - using that instead of starting a new database`,
         );
       } else if (pgHost && pgPort) {
-        const pgServer = new PostgresServer(dbId, "postgres");
-        const server = await pgServer.createPGServer(pgHost, pgPort);
+        const dataDirectory = this.args.config.get("emulators.dataconnect.dataDirectory");
```
yuchenshi, line 107: What if this is not set?

src/emulator/dataconnectEmulator.ts
```diff
@@ -102,14 +104,21 @@ export class DataConnectEmulator implements EmulatorInstance {
...
+        );
+        const server = await this.postgresServer.createPGServer(pgHost, pgPort);
         const connectableHost = connectableHostname(pgHost);
         connStr = `postgres://${connectableHost}:${pgPort}/${dbId}?sslmode=disable`;
-        server.on("error", (err) => {
+        server.on("error", (err: any) => {
           if (err instanceof FirebaseError) {
             this.logger.logLabeled("ERROR", "Data Connect", `${err}`);
           } else {
+            console;
```
yuchenshi, line 121:
```suggestion
```

src/emulator/dataconnectEmulator.ts
```diff
@@ -170,6 +179,22 @@ export class DataConnectEmulator implements EmulatorInstance {
     return Emulators.DATACONNECT;
   }
 
+  async clearData(): Promise<void> {
```
yuchenshi, line 182: Can one still clear the DB even if running a separate postgres? I'd say yes, but any concerns?

src/emulator/hub.ts
```diff
@@ -164,6 +166,25 @@ export class EmulatorHub extends ExpressBasedEmulator {
       res.status(200).json({ enabled: true });
     });
 
+    app.post(EmulatorHub.PATH_CLEAR_DATA_CONNECT, async (req, res) => {
+      // TODO: Sanity check that this is needed.
```
yuchenshi, line 170: Probably needed

src/emulator/hubExport.ts
```diff
@@ -289,6 +304,25 @@ export class HubExport {
...
   }
+
+  private async exportDataConnect(metadata: ExportMetadata): Promise<void> {
+    void trackEmulator("emulator_export", {
+      initiated_by: this.options.initiatedBy,
+      emulator_name: Emulators.DATACONNECT,
+    });
+    const dataconnectExportPath = path.join(this.tmpDir, metadata.dataconnect!.path);
+    if (fs.existsSync(dataconnectExportPath)) {
+      fse.removeSync(dataconnectExportPath);
```
yuchenshi, line 315: Shouldn't we do this after checking for running?

yuchenshi (approved):

src/emulator/dataconnect/pgliteServer.ts
joehan: Sorry, accidentally sent this out for review earlier than intended - the implementation here has some minor discrepancies from what was approved and needs some clean up.

Will re-request once its in a better state

src/emulator/dataconnectEmulator.ts
joehan: It will be undefined, and pglite will fall back to using an in-memory DB instead of a file backed one.

src/emulator/dataconnectEmulator.ts
joehan: That's a good product q - I think it's reasonable to offer, but if they're connecting to a real DB that becomes a pretty scary big red button.

src/emulator/hubExport.ts
joehan: Good call.

src/emulator/dataconnect/pgliteServer.ts
joehan: Ahh, that was what I was looking for, thanks! Despite many years of JS, `this` still screws me up 😬

src/emulator/dataconnectEmulator.ts
yuchenshi: But given the code as it is written right now, this seems to noop if it's not using pglite.

I think it should either do what it says on the label, or fail with a clear error message (e.g. "unimplemented"), but not noop.

PR #7839, opened 2024-10-15
Improved handling when `emulators:export` cannot read the metadata file.

joehan:
### Description
Improved handling when `emulators:export` cannot read the metadata file - errors here weren't valid JSON, so this was giving an unclear JSON.parse error.


kroikie (approved):
LGTM

PR #7856, opened 2024-10-21
Cleaning up remnants of old vsce publish script

joehan:
### Description
Cleaning up the VSCE publish script - this is being moved to kokoro, so it is longer needed here.


scripts/publish.sh
hlshen, line 1: You don't need to check for gsutil anymore

hlshen (approved):

scripts/publish.sh
joehan: Good catch, removed.

PR #7857, opened 2024-10-21
Set FIREBASE_BINARY as "firebasePath" user setting

hlshen:


firebase-vscode/src/utils/settings.ts
```diff
@@ -7,17 +7,23 @@ export interface Settings {
...
-  "npx -y firebase-tools@latest";
+// TODO: Temporary fallback for bashing, this should probably point to the global firebase binary on the system
+const DEFAULT_FIREBASE_BINARY = "npx -y firebase-tools@latest";
 
 export function getSettings(): Settings {
   const config = workspace.getConfiguration("firebase");
 
+  // TODO: Consider moving side effect out of getSettings
+  // Persist env var as path setting when path setting doesn't exist
+  if (process.env.FIREBASE_BINARY && !config.get<string>("firebasePath")) {
```
yuchenshi, line 18: According to http://b/373897889 we should check and set user preference (not per-workspace preference)

yuchenshi (changes requested):

firebase-vscode/src/utils/settings.ts
hlshen: Setting the setting type to `scope: application` should make it _only_ a user setting. I can add ConfigurationTarget.Global to ensure it?

firebase-vscode/package.json
```diff
@@ -63,7 +63,8 @@
         },
         "firebase.firebasePath": {
           "type": "string",
-          "markdownDescription": "%ext.config.firebasePath%"
+          "markdownDescription": "%ext.config.firebasePath%",
+          "scope": "application"
```
yuchenshi, line 67: Upon a closer look, I think we want `machine-overridable`. To quote the **scope** section in the [schema](https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema), emphasis mine: 

> `machine` - Machine specific settings that can be set only in user settings or only in remote settings. For example, **an installation path which shouldn't be shared across machines**.
> `machine-overridable` - Machine specific settings that can be overridden by workspace or folder settings.

The `-overridable` is mostly so that developers can pin to a different CLI / Node.js version for a certain workspace folder if they really want to. We won't programatically set this, it's all up to the developer

firebase-vscode/src/utils/settings.ts
yuchenshi: `ConfigurationTarget.Global` sounds good, and we need this because we want to keep it manually overridable per workspace if developers really want to (see the other comment).

yuchenshi (approved):

PR #7860, opened 2024-10-22
replace strip-ansi with stripVTControlCharacters from node:util

jsaguet:
### Description

This PR removes the dependency on `strip-ansi` and uses `stripVTControlCharacters` from `node:util` instead which is a built-in replacement.


joehan (approved):

joehan:
Thanks for the contribution! Did some quick testing and this looks great to me.

PR #7862, opened 2024-10-22
Fix rollouts:create to handle backend regionality & other fixes

blidd-google:
Fixes the following issues:
* Prompts user to select a region if multiple regions contain a backend with the given backend ID; if --force flag is passed and region is ambiguous, the command fails
* Lists branches with autocomplete when no branch or commit is specified
* Removes confirmation prompt after branch or commit is specified
* Fixes short form flags (-b for --git-branch, -g for --git-commit, -i removed)
* Sends request to stable v1beta API instead of v1alpha


src/apphosting/backend.ts
```diff
@@ -0,0 +1,482 @@
...
+  if (location) {
+    if (!allowedLocations.includes(location)) {
+      throw new FirebaseError(
+        `Invalid location ${location}. Valid choices are ${allowedLocations.join(", ")}`,
+      );
+    }
+  }
+
+  location =
+    location || (await promptLocation(projectId, "Select a location to host your backend:\n"));
```
joehan, line 102: Do these commands support noninteractive use cases? If so, we should double check this code - it seems like it will prompt even in noninteractive mode if there is no location.

src/apphosting/backend.ts
```diff
@@ -0,0 +1,482 @@
...
+
+  const gitRepositoryLink: GitRepositoryLink = await githubConnections.linkGitHubRepository(
+    projectId,
+    location,
+  );
+
+  const rootDir = await promptOnce({
+    name: "rootDir",
+    type: "input",
+    default: "/",
```
joehan, line 112: Should this default to . instead? / can easily be mistaken for an absolute path.

src/apphosting/backend.ts
```diff
@@ -0,0 +1,482 @@
...
+  }
+}
+
+/**
+ * Prompts the user for a backend id and verifies that it doesn't match a pre-existing backend.
+ */
+async function promptNewBackendId(
+  projectId: string,
+  location: string,
+  prompt: any,
```
joehan, line 266: Use a stronger type than any here

src/apphosting/backend.ts
```diff
@@ -0,0 +1,482 @@
...
+): Promise<string> {
+  while (true) {
+    const backendId = await promptOnce(prompt);
+    try {
+      await apphosting.getBackend(projectId, location, backendId);
+    } catch (err: any) {
+      if (err.status === 404) {
```
joehan, line 273: Ingrid recently added type safe helpers for getting status or message from an error (https://github.com/firebase/firebase-tools/blob/1d4b580cc5c572b71dffe2fc0c0024dc75ca7dd6/src/error.ts#L63) - this would be a good place to use them to silence these check warnings. 

Applies throughout.

src/apphosting/backend.ts
```diff
@@ -0,0 +1,482 @@
+import * as clc from "colorette";
```
joehan, line 1: Optional - this files has a mix of prompting functions and API calling functions. In the past, I've see this lead to really large and unwieldy files. If you'd like to get ahead of that, consider splitting this into  backendPrompts.ts and backend.ts

src/commands/apphosting-rollouts-create.ts
```diff
@@ -7,13 +7,12 @@ import { createRollout } from "../apphosting/rollout";
 
 export const command = new Command("apphosting:rollouts:create <backendId>")
   .description("create a rollout using a build for an App Hosting backend")
-  .option("-l, --location <location>", "specify the region of the backend", "us-central1")
-  .option("-i, --id <rolloutId>", "id of the rollout (defaults to autogenerating a random id)", "")
+  .option("-l, --location <location>", "specify the region of the backend", "-")
```
joehan, line 10: These flag changes are breaking changes, and ideally should be part of the next major version., unless App Hosting has a strong argument for allowing breaks in minor versions.

joehan (changes requested):
Concerned about the breaking flag changes

blidd-google:
> Concerned about the breaking flag changes

The `apphosting:rollouts:create` feature was added without announcement, so our team is okay with the risk of a breaking change in a minor CLI release given that the likelihood we actually break anybody is very low.

taeold (approved):
+1 for making this breaking change.

joehan (approved):

nolafs:
I currently using version and the manual rollout option is missing:

error: apphosting:rollouts:create is not a Firebase command

PR #7866, opened 2024-10-23
Clarify Java compatibility requirements

mltsy:
### Description

Update Java compatibility deprecation message to clarify how to get the necessary JRE.

For those less familiar with the Java ecosystem, it's not clear that you can't get modern JREs from Java.com:
- https://stackoverflow.com/questions/72438409/firebase-tools-no-longer-supports-java-version-before-11-please-upgrade-to-java
- https://stackoverflow.com/questions/79105227/why-does-java-com-still-recommend-java-8-when-there-are-multiple-newer-lte-rele?noredirect=1#comment139487042_79105227


joehan:
Thanks for contributing this! I think this is a good clarification, and I'll check about getting the docs updated too

PR #7875, opened 2024-10-25
Always log reauth error messages

joehan:
### Description
Always log reauth errors.

This error gets used in many places accross the CLI, and it is easy to accidentally swallow the user facing error message here. Rather than digging through all of the code base to address this, we'll just log this here.


NickChittle (approved):
LGTM

PR #7882, opened 2024-10-28
Clean up logs when using FIREBASE_DATACONNECT_POSTGRESQL_STRING

joehan:
### Description
Fixing a couple logging bugs that made it seem like PGLite was running even when FIREBASE_DATACONNECT_POSTGRESQL_STRING was used.

### Scenarios Tested
With FIREBASE_DATACONNECT_POSTGRESQL_STRING (corrected the capitalization inconsistency on the log label)
Screenshot 2024-10-28 at 3 29 00 PM

Without: 
Screenshot 2024-10-28 at 3 27 00 PM


src/emulator/emulatorLogger.ts
```diff
@@ -310,6 +310,9 @@ You can probably fix this by running "npm install ${systemLog.data.name}@latest"
       case "BULLET":
         utils.logLabeledBullet(label, text, "info", mergedData);
         break;
+      case "INFO":
+        utils.logLabeledBullet(label, text, "info", mergedData);
```
yuchenshi, line 314: Why do we need this in addition to `bullet`?

yuchenshi (approved):

PR #7885, opened 2024-10-30
Improve detection for outdated standalone CLI versions.

yuchenshi:
### Description

This uses a check to catch edge cases where `npm` is present but the `firebase`
command is installed as standalone.

### Scenarios Tested

Manually tested on my machine with both `npm` and standalone CLI.

### Sample Commands

(Just start the VSCE.)

### DIFF
--- a/firebase-vscode/src/extension.ts
+++ b/firebase-vscode/src/extension.ts
@@ -1,5 +1,5 @@
 import * as vscode from "vscode";
-import { spawnSync } from 'child_process';
+import { spawnSync } from "child_process";
 import * as semver from "semver";
 
 import { ExtensionBroker } from "./extension-broker";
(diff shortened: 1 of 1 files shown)


firebase-vscode/src/extension.ts
```diff
@@ -1,5 +1,5 @@
 import * as vscode from "vscode";
-import { spawnSync } from 'child_process';
+import { spawnSync } from "child_process";
```
yuchenshi, line 2: Lots of prettier diffs. Maybe we should add prettier lint to the CI to make sure everyone else runs it.

hlshen: ++ I'll look into it

hlshen (approved):

PR #7896, opened 2024-11-01
Fix breakage with app hosting beta API switch

inlined:
Fixes the following error message that started cropping up in `firebase apphosting:backends:create` when we switched to the beta API.

```
Error: Request to https://firebaseapphosting.googleapis.com/v1beta/projects/inlined-junkdrawer/locations/us-central1/backends/scuba2/traffic?updateMask=rolloutPolicy had HTTP Error: 400, Invalid JSON payload received. Unknown name "stages" at 'traffic.rollout_policy': Cannot find field.
```


taeold (approved):

hlshen:
Stages is being used in apphosting/backend.ts - please fix

PR #7897, opened 2024-11-01
Moving the emulator UI server to firebase-tools.

christhompsongoogle:
Move the server.ts file to firebase-tools so that the emulator UI can be served in-process with the firebase CLI.

Some questions are inline with the code (marked with "FIXME") for second opinions


src/emulator/ui/server.ts
```diff
@@ -0,0 +1,118 @@
...
+  const webDir = path.join(zipDirPath, 'client');
+  app.use(express.static(webDir));
+  // Required for the router to work properly.
+  app.get('*', function (_, res) {
+    res.sendFile(path.join(webDir, 'index.html'));
+  });
+
+  if (listenOptions.length == 0) {
+    console.error(`Failed to start UI server, listenOptions empty`);
+    process.exit(1);
```
yuchenshi, line 79: Don't exit the process, throw an error instead. It's in the same process as the CLI now. Same below

src/emulator/ui/server.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+  var server = null;
+  for (const opts of listenOptions) {
+    server = createServer(app).listen(opts);
+    server.once('listening', () => {
+      console.log(`Web / API server started at ${opts.host}:${opts.port}`);
```
yuchenshi, line 85: This log (or really, most logs in this file) isn't useful any more now we don't have a separate process. In case we do need to log, use `EmulatorLogger` instead of `console.log`.

src/emulator/ui/server.ts
```diff
@@ -0,0 +1,118 @@
...
+    res.sendFile(path.join(webDir, 'index.html'));
+  });
+
+  if (listenOptions.length == 0) {
+    console.error(`Failed to start UI server, listenOptions empty`);
+    process.exit(1);
+  }
+  var server = null;
+  for (const opts of listenOptions) {
+    server = createServer(app).listen(opts);
```
yuchenshi, line 83: This will only return the last server to `ui.ts` and only one destroyer will be created in `ui.ts`. This leads to servers not properly shutting down. `ExpressBasedEmulator` handles all the listening etc. correctly -- let's just inherit the class and reuse the logic there. No need to duplicate the work (Line 77-97 here).

I think the `createApp` function in a separate file is hindering our ability to reuse code in this case. Consider making it a method in class EmulatorUI instead. e.g. `override createExpressApp() { ... }`

yuchenshi (changes requested):

src/emulator/ui/server.ts
christhompsongoogle: you're right - I want to keep this in line with the server.ts in firebase-tools-ui but maybe that's a painful goal to have - this should work better, though still flawed

yuchenshi: I don't think we need to keep this in line with server.ts in firebase-tools-ui. After this PR gets merged, the file in the UI can be rewritten to serve development only. That file may get removed or changed so much that it no longer resembles anything here.

yuchenshi (approved):
Approach looks good. One nit inline. Please run `npm run fmt` to fix linter errors

src/emulator/ui.ts
```diff
@@ -13,55 +17,100 @@ export interface EmulatorUIOptions {
...
+export class EmulatorUI extends ExpressBasedEmulator {
 
-  start(): Promise<void> {
+  constructor(private args: EmulatorUIOptions) {
+    super({
+      listen: args.listen,
+    });
+  }
+
+  override async start(): Promise<void> {
```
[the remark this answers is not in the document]
christhompsongoogle: Since this one includes a projectID this strips that out before calling the super method

src/emulator/ui.ts
```diff
@@ -10,58 +14,99 @@ import { ALL_EXPERIMENTS, ExperimentName, isEnabled } from "../experiments";
...
+    ).filter((experimentName) => isEnabled(experimentName));
+    const emulatorGaSession = emulatorSession();
 
-    const session = emulatorSession();
-    if (session) {
-      env[Constants.FIREBASE_GA_SESSION] = JSON.stringify(session);
-    }
+    await downloadableEmulators.downloadIfNecessary(Emulators.UI);
+    const downloadDetails = downloadableEmulators.getDownloadDetails(Emulators.UI);
+    const webDir = path.join(downloadDetails.unzipDir!!, "client");
```
joehan, line 47: Might be misunderstanding what this does, but why do we need 2 !'s here? IIUC, this is just asserting that it is non-null twice

src/emulator/ui.ts
```diff
@@ -10,58 +14,99 @@ import { ALL_EXPERIMENTS, ExperimentName, isEnabled } from "../experiments";
...
+    // them using client SDKs. For features that involve multiple emulators or
+    // hard to accomplish using client SDKs, consider adding an API below.
+    app.get(
+      "/api/config",
+      this.jsonHandler(async () => {
+        const hubDiscoveryUrl = new URL(
+          `http://${EmulatorRegistry.url(Emulators.HUB).host}/emulators`,
+        );
+        const emulatorsRes = await fetch(hubDiscoveryUrl.toString());
+        const emulators = (await emulatorsRes.json()) as any;
```
joehan, line 59: Nit: This is a bit code smelly and fragile - if emulatorsRes ever returns something other than a properly formatted JSON, this will crash with a "... on noniterable object" error

Might be worth refactoring this to fail cleaner, but not blocking

joehan (approved):
Some nits, but LGTM overall

src/emulator/ui.ts
yuchenshi: If I'm not wrong, this can be done in-memory using `EmulatorRegistry` to directly get the hub instance. No need for HTTP at all.

christhompsongoogle: Nothing a leaky abstraction won't fix - PTAL

christhompsongoogle: Question: do we need EmulatorHub.PATH_EMULATORS anymore?

src/emulator/hub.ts
```diff
@@ -96,7 +107,7 @@ export class EmulatorHub extends ExpressBasedEmulator {
           ...info,
         };
       }
-      res.json(body);
+      res.json(this.getRunningEmulatorsMapping());
```
yuchenshi, line 110: Yo, please remove the duplicate code above now that you've created a function.

src/emulator/ui.ts
```diff
@@ -1,67 +1,114 @@
...
+    // Exposes the host and port of various emulators to facilitate accessing
+    // them using client SDKs. For features that involve multiple emulators or
+    // hard to accomplish using client SDKs, consider adding an API below.
+    app.get(
+      "/api/config",
+      this.jsonHandler(async () => {
+        const hubDiscoveryUrl = new URL(
+          `http://${EmulatorRegistry.url(Emulators.HUB).host}/emulators`,
+        );
+        const emulatorsRes = await fetch(hubDiscoveryUrl.toString());
```
yuchenshi, line 60: Like, also remove fetch?

src/emulator/ui.ts
yuchenshi: > Question: do we need EmulatorHub.PATH_EMULATORS anymore?

Yes, that's part our public API and used for an SDK too

yuchenshi (approved):

src/emulator/ui.ts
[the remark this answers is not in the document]
yuchenshi: I don't see `start()` taking an argument anywhere, here or the base class. Mind pointing me in the right direction?

PR #7907, opened 2024-11-05
Preparing for genkit 0.9.0

ifielker:
preparing for genkit version 0.9.0+


src/init/features/genkit/index.ts
```diff
@@ -0,0 +1,581 @@
...
+    }
+
+    // Functions with genkit should always be typescript
+    setup.languageOverride = "typescript";
+    await functionsSetup(setup, config, options);
+    delete setup.languageOverride;
+    logger.info();
+  }
+
+  const projectDir: string = `${config.projectDir}/${setup.functions.source}`;
```
pavelgj, line 96:
```suggestion
  const projectDir = `${config.projectDir}/${setup.functions.source}`;
```

ifielker: done

src/init/features/genkit/index.ts
```diff
@@ -0,0 +1,581 @@
...
+    throw new FirebaseError(
+      "Please update your firebase-tools. Alternatively you can set a GENKIT_DEV_VERSION environment variable to choose a genkit version < 1.0.0",
+    );
+  } else if (semver.gte(genkitVersion, "0.6.0")) {
+    sampleVersion = "0.9.0";
+    useInit = false;
+  } else {
+    sampleVersion = "";
+    useInit = true;
+  }
```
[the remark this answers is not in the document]
ifielker: The whole reason we're doing this PR is because we don't want things to break and have no solution.
With this setup, I can submit this PR now, people can update to it, and it will continue to allow them to install 0.5.17 (the latest) up until we release 0.9.0 and then it will install 0.9.0. 
If I don't do this, then I can't submit this PR until sometime after we release 0.9.0 and there will be a period where it is just broken with no "just update to the latest firebase-tools" fix. 
It would also not notify people to update their firebase-tools because a newer version (>=1.0.0) is available. 
If we had this kind of logic in place from the beginning, then nothing would break when we release 0.9.0. So I'm a bit confused why you want me to remove it. People do not necessarily update their firebase-tools unless they have a reason to do so. 
I wouldn't want them to be "stuck" getting `firebase init genkit` installs for 0.9.0 when the latest version of genkit is 1.5.7 and there have been breaking change and the docs are for 1.5.7 and they just have a bad time trying to figure out what's going on and end up not liking genkit because it's "broken". I'd much rather have them getting a message saying "you need to update your firebase-tools, or explicitly pick an "old" version of genkit to install". If we do it this way then, we can remove support for older versions and add support for newer versions and not be coupled to the release of genkit versions without breaking anything. I feel very strongly that this is a good solution.

src/init/features/genkit/index.ts
```diff
@@ -0,0 +1,595 @@
...
+      label: "Google AI",
+      plugin: "@genkit-ai/googleai",
+      package: `@genkit-ai/googleai@${genkitVersion}`,
+    },
+    vertexai: {
+      label: "Google Cloud Vertex AI",
+      plugin: "@genkit-ai/vertexai",
+      package: `@genkit-ai/vertexai@${genkitVersion}`,
+    },
+    ollama: {
```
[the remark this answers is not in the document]
ifielker: done.

src/init/features/genkit/index.ts
```diff
@@ -0,0 +1,595 @@
...
+        logger.info(`    cd ${setup.functions.source} && genkit start`);
+      } else {
+        await wrapSpawn(
+          "npm",
+          ["install", "-g", `genkit-cli@${genkitInfo.genkitVersion}`],
+          projectDir,
+        );
+        await genkitSetup(options, genkitInfo, projectDir);
+        logger.info("Start the Genkit developer experience by running:");
+        logger.info(`    cd ${setup.functions.source} && npm run genkit:start`);
```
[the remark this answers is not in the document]
ifielker: We *absolutely* need these. I know you know all the details about how to start genkit and get it running and the right order to do those things, but the genkit newbie does not (i.e. people like me who spent too much trying with a "broken" genkit trying to debug what was wrong.)

src/init/features/genkit/index.ts
```diff
@@ -0,0 +1,595 @@
...
+  } catch (err) {
+    logger.error(`Failed to generate sample file: ${err}`);
+    process.exit(1);
+  }
+}
+
+/**
+ * Updates package.json with Genkit-expected fields.
+ */
+async function updatePackageJson(nonInteractive: boolean, projectDir: string) {
```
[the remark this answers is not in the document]
ifielker: Recall that even with instructions, I had a bad time trying to figure out why the UI was "broken" and spent quite some time trying to fix "step 7" before moving on to "step 8". And the solution was "you need to run step 7, then not open the UI yet, then run step 8, then go back and open the UI from step 7". That's why I added package.json scripts like `genkit:start` that runs step 7 and step 8 in order together, so there's no confusion. (and also `genit:stop` which just stops the ui). That way people are not left in a "ok, it's installed, now what?" state, but rather they have an easy way to get everything going so they can focus on playing with a working genkit and not on trying to "fix a broken step 7" like I was.

src/init/features/genkit/index.ts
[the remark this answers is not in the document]
pavelgj: support 0.5 is fine... 
I'm more thinking about the `semver.gte(genkitVersion, "1.0.0")` case. I think that if version is `semver.gte(genkitVersion, "0.6.0")` we pin it to 0.9. until we update the firebase cli in the future.

src/init/features/genkit/index.ts
```diff
@@ -0,0 +1,581 @@
...
+  }
+  const existingPackageJson = JSON.parse(fs.readFileSync(packageJsonPath, "utf8"));
+  const choice = nonInteractive
+    ? "overwrite"
+    : await promptWriteMode("Would you like to update your package.json with suggested settings?");
+  const packageJson = {
+    main: "lib/index.js",
+    scripts: {
+      "genkit:start": "genkit ui:start && GENKIT_ENV=dev tsx --watch src/genkit-sample.ts",
+      "genkit:stop": "genkit ui:stop",
```
pavelgj, line 554:
```suggestion
      "genkit:start": "genkit start -- tsx --watch src/genkit-sample.ts"
```

templates/genkit/firebase.0.9.0.template
```diff
@@ -0,0 +1,63 @@
...
+      config: {
+        temperature: 1,
+      },
+    });
+
+		// Handle the response from the model API. In this sample, we just
+		// convert it to a string, but more complicated flows might coerce the
+		// response into structured output or chain the response into another
+		// LLM call, etc.
+    return llmResponse.text;
```
pavelgj:
```suggestion

    // Handle the response from the model API. In this sample, we just
    // convert it to a string, but more complicated flows might coerce the
    // response into structured output or chain the response into another
    // LLM call, etc.
    return llmResponse.text;
```

templates/genkit/firebase.0.9.0.template
```diff
@@ -0,0 +1,63 @@
...
+$GENKIT_CONFIG_IMPORTS
+$GENKIT_MODEL_IMPORT
+
+// From the Firebase plugin, import the functions needed to deploy flows using
+// Cloud Functions.
+import {firebaseAuth} from "@genkit-ai/firebase/auth";
+import {onFlow} from "@genkit-ai/firebase/functions";
+
+// Log debug output to tbe console.
+logger.setLogLevel("debug");
```
pavelgj:
```suggestion
// Import the Genkit core libraries and plugins.
import {genkit} from 'genkit';
$GENKIT_CONFIG_IMPORTS
$GENKIT_MODEL_IMPORT

// From the Firebase plugin, import the functions needed to deploy flows using
// Cloud Functions.
import {firebaseAuth} from "@genkit-ai/firebase/auth";
import {onFlow} from "@genkit-ai/firebase/functions";

```

ifielker: done

templates/genkit/firebase.0.9.0.template
ifielker: I'm not seeing that in the current version of the template. I might have fixed it along the way (possibly it was tabs before?)

src/init/features/genkit/index.ts
```diff
@@ -0,0 +1,634 @@
...
+    genkitVersion = execFileSync("npm", ["view", "genkit", "version"]).toString();
+  }
+
+  if (!genkitVersion) {
+    throw new FirebaseError("Unable to determine genkit version to install");
+  }
+
+  if (semver.gte(genkitVersion, "1.0.0")) {
+    // We don't know about this version.
+    throw new FirebaseError(
```
pavelgj, line 71: as discussed please change this to a warning

ifielker: done

pavelgj (approved):

PR #7917, opened 2024-11-07
App Hosting Emulator: Monorepo support

mathu97:
- Monorepo support via `rootDirectory` config in `firebase.json` App Hosting emulator config
- Emulator now can run even if an apphosting.yaml file is not present in the user's project

### DIFF
--- a/src/apphosting/config.ts
+++ b/src/apphosting/config.ts
@@ -8,6 +8,8 @@ import * as prompt from "../prompt";
 import * as dialogs from "./secrets/dialogs";
 import { AppHostingYamlConfig } from "./yaml";
 import { FirebaseError } from "../error";
+import { detectProjectRoot } from "../detectProjectRoot";
+import { logWarning } from "../utils";
 
 export const APPHOSTING_BASE_YAML_FILE = "apphosting.yaml";
 export const APPHOSTING_LOCAL_YAML_FILE = "apphosting.local.yaml";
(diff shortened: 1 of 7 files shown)


src/apphosting/config.ts
```diff
@@ -39,12 +40,28 @@ export interface Config {
 }
 
 /**
- * Finds the project root for apphosting backends.
- * Starts with cwd and walks up the path until an apphosting.yaml file is found
+ * Returns the absolute path for an app hosting backend root.
  *
- * Example path that's returned: "/home/my-project"
+ * If backendRootRelativeToProjectRoot is provided then this function will
+ * try to return the absolute path of the backend root
```
tonyjhuang, line 46: small nit: I think backendRootRelativeToProjectRoot should be it's own function, its functionality is completely distinct from when it's not passed in

src/apphosting/config.ts
```diff
@@ -67,12 +84,13 @@ export function discoverBackendRoot(cwd: string): string | null {
 /**
  * Lists absolute paths for `apphosting.*.yaml` configs at backend root
  */
-export function discoverConfigsAtBackendRoot(cwd: string): string[] {
-  const backendRoot = discoverBackendRoot(cwd);
+export function discoverConfigsAtBackendRoot(cwd: string, rootDir?: string): string[] {
```
tonyjhuang, line 87: This function might look better if you pass in a single directory path and resolve the `cwd` vs `rootDir` beforehand, eg

```
backendRoot = backendRoot ? join(cwd, backendRoot) : cwd
discoverConfigs(backendRoot)
```

tonyjhuang (changes requested):

src/emulator/apphosting/config.ts
```diff
@@ -11,18 +11,29 @@ import { AppHostingYamlConfig } from "../../apphosting/yaml";
...
+  if (!baseFilePath && !localFilePath) {
+    return AppHostingYamlConfig.empty();
+  }
+
+  // If one of them exists ...
+  if (!baseFilePath || !localFilePath) {
+    return await AppHostingYamlConfig.loadFromFile((baseFilePath || localFilePath)!);
+  }
+
+  return await loadConfigForEnvironment(localFilePath, baseFilePath);
```
taeold, line 38: Hmm doesn't this load `apphosting.{env}.yaml`? IIUC, we don't load environment configs.

Would it make more sense to call `.merge()` here directly?

src/emulator/apphosting/config.spec.ts
```diff
@@ -9,76 +8,125 @@ import * as configImport from "../../apphosting/config";
...
+  apphostingYamlConfigOne.addSecret({ variable: "randomSecretThree", secret: "secretThree" });
+
+  const apphostingYamlConfigTwo = AppHostingYamlConfig.empty();
+  apphostingYamlConfigTwo.addEnvironmentVariable({
+    variable: "randomEnvOne",
+    value: "envOne",
+  });
+  apphostingYamlConfigTwo.addEnvironmentVariable({
+    variable: "randomEnvTwo",
+    value: "blah",
```
taeold, line 41: nit* I think the test case would be a lot clearer if these override markers are self descriptive, e.g.

"value: "THIS IS FROM SECOND CONFIG""

src/emulator/apphosting/config.spec.ts
```diff
@@ -9,76 +8,125 @@ import * as configImport from "../../apphosting/config";
...
-      apphostingYamlConfigThree.addSecret({ variable: "randomSecretOne", secret: "bleh" });
-      apphostingYamlConfigThree.addSecret({ variable: "randomSecretTwo", secret: "secretTwo" });
-      apphostingYamlConfigThree.addSecret({ variable: "randomSecretFour", secret: "secretFour" });
-
-      loadAppHostingYamlStub.onFirstCall().returns(apphostingYamlConfigThree);
-      loadAppHostingYamlStub.onSecondCall().returns(apphostingYamlConfigTwo);
-
-      const apphostingConfig = await getLocalAppHostingConfiguration("test");
+      loadAppHostingYamlStub.onFirstCall().returns(apphostingYamlConfigTwo);
+      loadAppHostingYamlStub.onSecondCall().returns(apphostingYamlConfigOne);
```
taeold, line 127: nit* pretty tricky since it's unclear which file is associated with these reads.

When I was writing tests for similar functionality, I ended up creating a tmpdir and actually wrote files. Not asking you to do it here but something to consider if you find later that these test are brittle:

https://github.com/firebase/firebase-tools/blob/7a500afbcdbbd631751c7c8a606c50aeb18930ac/src/functions/env.spec.ts#L292

taeold (approved):

src/emulator/apphosting/config.spec.ts
mathu97: ya this is a good way to go about it, will take a look!

tonyjhuang (approved):

PR #7920, opened 2024-11-08
Make emulator links clickable

mathu97:
Makes emulator links clickable from terminal by prepending the links with `http://`


tonyjhuang:
This affects all emulators right? Can you make sure this is intentional?

tonyjhuang:
How does hosting's emulator handle this?

PR #7940, opened 2024-11-13
Increase memory for server function on Next.js with image optimization

chalosalvador:

### DIFF
--- a/src/frameworks/index.ts
+++ b/src/frameworks/index.ts
@@ -55,6 +55,7 @@ import { resolveProjectPath } from "../projectPath";
 import { logger } from "../logger";
 import { WebFrameworks } from "./frameworks";
 import { constructDefaultWebSetup } from "../fetchWebSetup";
+import { isUsingImageOptimization } from "./next/utils";
 
 export { WebFrameworks };
 
@@ -537,6 +538,12 @@ ${
       if (bootstrapScript) await writeFile(join(functionsDist, "bootstrap.js"), bootstrapScript);
 
       // TODO move to templates
+
+      if (frameworksBackend && framework === "next") {
+        if (await isUsingImageOptimization(getProjectPath(), ".next")) {
+          frameworksBackend.memory = "512MiB";
+        }
+      }
 
       if (packageJson.type === "module") {
         await writeFile(


src/frameworks/index.ts
```diff
@@ -538,6 +539,12 @@ ${
 
       // TODO move to templates
 
+      if (frameworksBackend && framework === "next") {
+        if (await isUsingImageOptimization(getProjectPath(), ".next")) {
+          frameworksBackend.memory = "512MiB";
+        }
+      }
```
leoortizz, line 546: instead of doing it here, I’d move this to the Next.js files and return `frameworksBackend` overrides from `ɵcodegenFunctionsDirectory`, @jamesdaniels WDYT?

src/frameworks/index.ts
```diff
@@ -538,6 +539,12 @@ ${
 
       // TODO move to templates
 
+      if (frameworksBackend && framework === "next") {
+        if (await isUsingImageOptimization(getProjectPath(), ".next")) {
```
leoortizz, line 543: don't we need to do this only for Next 15?

leoortizz, line 543: `.next` can be customized, so we should use `distDir` instead, which should be easy after we move this to `ɵcodegenFunctionsDirectory` as we have it there

chalosalvador: The memory issue is happening on Next 14 and 13 as well 🫠

src/frameworks/index.ts
chalosalvador: I refactored this to do it in the Next.js file.

src/frameworks/index.ts
chalosalvador: Not needed anymore since the check is happening in the Next.js file.

src/frameworks/index.ts
```diff
@@ -537,25 +538,24 @@ ${
       if (bootstrapScript) await writeFile(join(functionsDist, "bootstrap.js"), bootstrapScript);
 
       // TODO move to templates
-
+      const mergedFrameworksBackend = {
+        ...functionsDirectoryFrameworksBackend,
+        ...frameworksBackend,
```
leoortizz, line 543: I think `functionsDirectoryFrameworksBackend` should come after so it can override the default values?

chalosalvador: I'm thinking that the user could set the memory to a higher value, in that case we should not override it.

src/frameworks/next/index.ts
```diff
@@ -677,8 +679,14 @@ export async function ɵcodegenFunctionsDirectory(
   }
 
   // Add the `sharp` library if app is using image optimization
+
+  let frameworksBackend: FrameworksBackendOptions | undefined;
   if (await isUsingImageOptimization(sourceDir, distDir)) {
     packageJson.dependencies["sharp"] = SHARP_VERSION;
+    // TODO check if it is only for Next 15
+    frameworksBackend = {
+      memory: "512MiB",
```
leoortizz, line 688: maybe move this to a constant?

PR #7948, opened 2024-11-13
Migrate some API calls to resourcemanager.googleapis.com

joehan:
### Description
Migrate some API calls from `firebase.googleapis.com` to `cloudresourcemanger.googleapis.com`. This addresses some issues with using the CLI with non-Firebase Cloud projects.

### Scenarios Tested
I tested a bunch of commands (including init and deploy for GCP TOS products), and verified that they work.

### DIFF
--- a/src/management/projects.ts
+++ b/src/management/projects.ts
@@ -2,7 +2,7 @@ import * as clc from "colorette";
 import * as ora from "ora";
 
 import { Client } from "../apiv2";
-import { FirebaseError } from "../error";
+import { FirebaseError, getErrStatus } from "../error";
 import { pollOperation } from "../operation-poller";
 import { Question, promptOnce } from "../prompt";
 import * as api from "../api";
@@ -46,6 +46,11 @@ const firebaseAPIClient = new Client({
   urlPrefix: api.firebaseApiOrigin(),
   auth: true,
   apiVersion: "v1beta1",
+});
+
+const resourceManagerClient = new Client({
+  urlPrefix: api.resourceManagerOrigin(),
+  apiVersion: "v1",
 });
 
 export async function createFirebaseProjectAndLog(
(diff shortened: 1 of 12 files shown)


src/management/projects.ts
```diff
@@ -411,6 +417,19 @@ export async function getFirebaseProject(projectId: string): Promise<FirebasePro
     });
     return res.body;
   } catch (err: any) {
+    if (getErrStatus(err) === 404) {
+      try {
+        const info = await getProject(projectId);
+        // TODO: Update copy based on Rachel/Yvonne's feedback.
```
joehan, line 423: This messaging is coming later as part of supporting projects that don't have firebase enabled.

src/types/project/index.d.ts
```diff
@@ -13,7 +13,7 @@ export interface FirebaseProjectMetadata {
   name: string /* The fully qualified resource name of the Firebase project */;
   projectId: string;
   projectNumber: string;
-  displayName: string;
+  displayName?: string;
```
joehan, line 16: Based on usage, this always should have been optional anyway.

src/management/projects.ts
bkendall: Consider putting a debug log in here, just in case someone is having _some_ issue, you'll know this path is being followed (and they're not seeing messaging they should be)

src/management/projects.ts
```diff
@@ -423,3 +442,21 @@ export async function getFirebaseProject(projectId: string): Promise<FirebasePro
     );
   }
 }
+
+export interface ProjectInfo {
+  projectNumber: string;
```
bkendall, line 447: You know what would probably be _really_ nice, is making this, you know, a number. Similarly with the createTime - at _minimum_ that should be a number, but better yet might be that it's a Date.

bkendall (approved):

src/management/projects.ts
joehan: https://cloud.google.com/resource-manager/reference/rest/v1/projects#resource:-project Agree fully, but not up to me - the API returns these both as strings, so I'm matching that representation.

src/management/projects.ts
joehan: Good call, done

PR #7953, opened 2024-11-15
Change type from number to integer for the emulator ports.

jfbenckhuijsen:
### Description

Changed the firebase-config.json Schema file. The ports of the emulators are changed from number to integer. Main reason for this is when using the schema in a generator like jsonschema2pojo, the generator will generate java.lang.Double as field instead of java.lang.Integer. As ports can only be integers, this setup better reflects the field.

### DIFF
--- a/schema/firebase-config.json
+++ b/schema/firebase-config.json
@@ -364,7 +364,7 @@
                             "type": "string"
                         },
                         "port": {
-                            "type": "number"
+                            "type": "integer"
                         },
                         "startCommandOverride": {
                             "type": "string"
@@ -379,7 +379,7 @@
                             "type": "string"
                         },
                         "port": {
-                            "type": "number"
+                            "type": "integer"
                         }
                     },
                     "type": "object"
@@ -391,7 +391,7 @@
                             "type": "string"
                         },
                         "port": {
-                            "type": "number"
+                            "type": "integer"
                         }
                     },
                     "type": "object"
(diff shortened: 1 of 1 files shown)


joehan (changes requested):
I agree with the gist of this change - however, this file is generated from the contents of `src/firebaseConfig.ts` by running `npm run generate:json-schema`. It shouldn't be edited manually (and our test suite will not pass if it is).

If you can find a way to change `src/firebaseConfig.ts` to get this outcome, go for it!

jfbenckhuijsen:
> I agree with the gist of this change - however, this file is generated from the contents of `src/firebaseConfig.ts` by running `npm run generate:json-schema`. It shouldn't be edited manually (and our test suite will not pass if it is).
> 
> If you can find a way to change `src/firebaseConfig.ts` to get this outcome, go for it!

Looks like that should be doable with 

`    /**
     * @items.type integer
     */
`

Just a quick question, the UI field has a port specified as `integer | string`. When would you specify a string?

joehan:
> Just a quick question, the UI field has a port specified as `number | string`. When would you specify a string?
That's a good question - from some git blame searching, it looks like at some point in the distant past, we had `port: string` for all of the emulator configs, and when we switched them over to the more appropriate `number`, we set this one to accept both, presumably for backward compatibility. 
Switching it is _technically_ a breaking change, but I suspect almost everyone uses number at this point. I'll make myself a note to switch this over on the next major version of firebase-tools, just in case.

joehan (approved):

PR #7962, opened 2024-11-18
Added alias support and alias appdistribution:groups:*

joehan:
### Description
Added a new Command.alias() support and use it to alias `appdistribution:group:*` to `appdistribution:groups:*`

### Scenarios Tested
Confirmed that `appdistribution:groups:*` and `appdistribution:group:*` both call the correct command.


src/commands/appdistribution-group-create.ts
```diff
@@ -6,6 +6,7 @@ import { getProjectName } from "../appdistribution/options-parser-util";
 
 export const command = new Command("appdistribution:group:create <displayName> [alias]")
   .description("create group in project")
+  .alias("appdistribution:groups:create")
```
kaibolay, line 9: Should this be the canonical form and `group:*` the alias?

What would the effect be? I guess we don't want both forms to show up in the help text - only the plural form. Or at least give the plural form more prominence.

kaibolay (approved):

src/commands/index.ts
```diff
@@ -29,10 +29,6 @@ export function load(client: any): any {
   client.appdistribution.group.list = loadCommand("appdistribution-group-list");
```
kaibolay, line 29: I assume keeping `client.appdistribution.groups` and removing `client.appdistribution.group` makes more sense.

..and probably the commands themselves could/should be renamed: `appdistribution-group-*` to `appdistribution-groups-`.

kaibolay (approved):

src/commands/index.ts
joehan: Renamed the files, and updated this file to make the commands available under both the old and new names when using firebase-tools as a module.

kaibolay: Thanks!

src/commands/index.ts
```diff
@@ -26,9 +26,10 @@ export function load(client: any): any {
   client.appdistribution.testers.add = loadCommand("appdistribution-testers-add");
   client.appdistribution.testers.delete = loadCommand("appdistribution-testers-remove");
   client.appdistribution.group = {};
```
kaibolay, line 28: sorry, one more: `client.appdistribution.groups` (i.e. plural)?

joehan: See line 32 - we copy it over from client.appdistribution.group instead of calling loadCommand 2x, so that we don't register the commands 2x

kaibolay: What I meant was swapping `group` for `groups` on lines 29-31 and reversing the assignment on line 32. But it doesn't really matter. We're deep in supernit territory. Sorry.

PR #7963, opened 2024-11-18
Apphosting emulator: Add apphositng.local.yaml to `.gitignore`

mathu97:
Automatically add `apphositng.local.yaml` file to `.gitignore` file whenever user exports secrets using `apphosting config:export` or when the app hosting emulator is being initialized.


src/apphosting/config.ts
```diff
@@ -243,6 +244,9 @@ export async function exportConfig(
   // update apphosting.local.yaml
   await localAppHostingConfig.upsertFile(localAppHostingConfigPath);
   logger.info(`Wrote secrets as environment variables to ${APPHOSTING_LOCAL_YAML_FILE}.`);
+
+  updateOrCreateGitignore(backendRoot, [APPHOSTING_LOCAL_YAML_FILE]);
```
tonyjhuang, line 248: should this be projectRoot?

mathu97: I guess that comes down to preference. I'm guessing having a single .gitignore at project root is more common even in monorepos, so we can go with that?

mathu97: discussed offline to create or update .gitignore at projectroot

src/utils.ts
```diff
@@ -918,3 +918,23 @@ export function readSecretValue(prompt: string, dataFile?: string): Promise<stri
...
+ * Updates or creates a .gitignore file with the given entries in the given path
+ */
+export function updateOrCreateGitignore(dirPath: string, entries: string[]) {
+  const gitignorePath = path.join(dirPath, ".gitignore");
+
+  if (fs.existsSync(gitignorePath)) {
+    let content = fs.readFileSync(gitignorePath, "utf-8");
+    for (const entry of entries) {
+      if (!content.includes(entry)) {
+        content += `${entry}\n`;
```
tonyjhuang, line 931: Should this be `\n${entry}` to account for .gitignores without a new line at the end of the file?

src/utils.ts
```diff
@@ -918,3 +918,23 @@ export function readSecretValue(prompt: string, dataFile?: string): Promise<stri
...
+      }
+    }
+
+    fs.writeFileSync(gitignorePath, content);
+    return;
+  }
+
+  fs.writeFileSync(gitignorePath, entries.join("\n"));
```
tonyjhuang, line 939: This is a good opportunity for an early return: go/tott/651

tonyjhuang (approved):
Two small comments

taeold (approved):

src/utils.ts
mathu97: uhm, `${entry}\n` should still work because we're using `.includes` above instead of `===`

tonyjhuang: Sorry I'm not following. I was saying we may want to append `\n${entry}` in case content was something like the following: `line1\nline2`. 

Appending `${entry}\n` would result in the following: `line1\nline2${entry}\n` which would be problematic

PR #7964, opened 2024-11-18
Apphosting emulator: Clean up secrets when writing to apphosting.local.yaml

mathu97:
Remove secrets when writing to apphosting.local.yaml file. This helps avoid confusion for users as secrets from this file aren't read in when the emulator is run.


src/apphosting/yaml.ts
```diff
@@ -69,6 +69,10 @@ export class AppHostingYamlConfig {
     this._secrets.set(secret.variable, secret);
   }
 
+  resetSecrets() {
```
tonyjhuang, line 72: optional nit: clearSecrets

src/apphosting/config.ts
```diff
@@ -242,8 +242,9 @@ export async function exportConfig(
     });
   }
 
-  // update apphosting.local.yaml
-  await localAppHostingConfig.upsertFile(localAppHostingConfigPath);
+  // remove secrets to avoid confusion as they are not read anyways.
+  localAppHostingConfig.resetSecrets();
```
tonyjhuang, line 246: test?

tonyjhuang (approved):
two small comments

src/apphosting/config.ts
mathu97: added

PR #7941, opened 2024-11-13
upgrade superstatic to 9.1.0

bkendall:
### Description

Updates superstatic to v9.1.0.

Fixes #7929
Fixes #7173

### Scenarios Tested

A basic website still worked well with this version. Works on node 22 as well!

PR #7975, opened 2024-11-20
Upgrade superstatic in package.json to 9.1.0

f0urfingeredfish:
### Description

Follow up to https://github.com/firebase/firebase-tools/pull/7941 to also upgrade `superstatic` in the package.json. https://github.com/firebase/firebase-tools/pull/7941#issuecomment-2488945873

### Sample Commands

npm install uses superstatic@9.1.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+Updates `superstatic` to `9.1.0` in package.json.
--- a/package.json
+++ b/package.json
@@ -152,7 +152,7 @@
     "sql-formatter": "^15.3.0",
     "stream-chain": "^2.2.4",
     "stream-json": "^1.7.3",
-    "superstatic": "^9.0.3",
+    "superstatic": "^9.1.0",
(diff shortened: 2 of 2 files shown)


melyux:
Thanks for this, was wondering why it wasn't working. Let's hope it's merged soon

moekify:
Is this happening? We're waiting for this fix to alleviate the underlying security issue

moekify:
@bkendall Any update? Wondering what's the holdup with something seemingly small like this? 🤔 @melyux Would appreciate if this gets pushed out soon :)

bkendall:
Thanks for raising this - looks as though our npm-shrinkwrap isn't being published as expected, so we'll have to address that some other time. With the (American) holidays and limited time I have already, didn't manage to see and get around to this. But now we're here! I've made #8026 with a bit of updates (since this PR is out of date a smidge and we're getting ready for a release like right now), so that should come out very soon :)

moekify:
Thank you! Appreciated :)

PR #7982, opened 2024-11-21
Cleaning up some leftovers from #7836

joehan:
### Description
Fixes a few small things that I meant to include in #7836:
- ClearData now works for non-pglite instances
- ImportExportEmualtors now includes all the emulators that support import/export
- (sorta unreleated, but noticed it while working) Removed a keybinding for a command that no longer exists.

### DIFF
--- a/src/emulator/dataconnectEmulator.ts
+++ b/src/emulator/dataconnectEmulator.ts
@@ -1,4 +1,5 @@
 import * as childProcess from "child_process";
+import * as pg from "pg";
 import { EventEmitter } from "events";
 import * as clc from "colorette";
 import * as path from "path";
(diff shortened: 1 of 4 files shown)


src/emulator/dataconnectEmulator.ts
```diff
@@ -180,6 +181,9 @@ export class DataConnectEmulator implements EmulatorInstance {
   async clearData(): Promise<void> {
     if (this.postgresServer) {
       await this.postgresServer.clearDb();
+    } else {
+      const conn = new pg.Client(dataConnectLocalConnString());
```
yuchenshi, line 185: Do we need to close the connection?

yuchenshi (approved):

src/emulator/dataconnectEmulator.ts
joehan: Yes we do, good catch.

PR #7989, opened 2024-11-21
Add default value for `dataDir` in `init dataconnect`

joehan:
### Description
Add default value for `dataDir` in `init dataconnect`, and makes it configurable via `init emulators:dataconnect`.

I intentionally held this back until post Thanksgiving freeze so that this persistence behavior is opt in for now.

### DIFF
--- a/src/emulator/initEmulators.ts
+++ b/src/emulator/initEmulators.ts
@@ -8,7 +8,7 @@ import { Emulators } from "./types";
 import { exportConfig } from "../apphosting/config";
 import { detectProjectRoot } from "../detectProjectRoot";
 
-type InitFn = () => Promise<Record<string, string> | null>;
+type InitFn = (currentConfig?: Record<string, any>) => Promise<Record<string, string> | null>;
 type AdditionalInitFnsType = Partial<Record<Emulators, InitFn>>;
 
 export const AdditionalInitFns: AdditionalInitFnsType = {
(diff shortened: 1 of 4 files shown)


src/emulator/initEmulators.ts
```diff
@@ -46,6 +46,18 @@ export const AdditionalInitFns: AdditionalInitFnsType = {
 
     return mapToObject(additionalConfigs);
   },
+  [Emulators.DATACONNECT]: async () => {
+    const dataDir = await promptOnce({
+      name: "dataDir",
+      type: "input",
+      default: "./postgresData",
```
fredzqm, line 53: Should this be something within the `dataconnect` directory?

In the order of my preference:
- `./dataconnect/.dataconnect/pglite_data` (hidden folder, out of sight)
- `./dataconnect/pglite_data` (next to schema and connector sources)
- `./dataconnect-generated/pglite_data` (next to generated SDK's default path)

fredzqm (approved):

src/init/features/dataconnect/index.ts
```diff
@@ -79,6 +79,8 @@ export async function doSetup(setup: Setup, config: Config): Promise<void> {
     isBillingEnabled ? await ensureApis(setup.projectId) : await ensureSparkApis(setup.projectId);
   }
   const info = await askQuestions(setup, isBillingEnabled);
+  // Most users will want to perist data between emulator runs, so set this to a reasonable default.
+  config.set("emulators.dataconnect.dataDir", "./postgresData");
```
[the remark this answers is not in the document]
joehan: Good catch, will do

src/emulator/initEmulators.ts
joehan: That seems reasonable - went with the first option, except that we use the configured data connect source directory.

src/init/features/dataconnect/index.ts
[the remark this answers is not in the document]
joehan: Good call.

src/emulator/initEmulators.ts
joehan: Ended up refactoring this a bit because when you pass a default to inquirer, it forces blank answers to that default. Instead, we'll just ask if you want to enable persistence, and tell you what field to change if you want a different directory.

PR #7990, opened 2024-11-22
App Hosting Emulator: Change startCommandOverride to startCommand

mathu97:


taeold (approved):
Can we update the docs asap?

https://firebase.google.com/docs/app-hosting/emulate

PR #8001, opened 2024-11-26
Support for AI Testing Agent

kaibolay:
Add support for `--test-case-ids` and `--test-case-ids-file` to `appdistribution:distribute` to launch .

### Description

The Automated Testing Agent is a beta feature of App Distribution. Once test cases are created in the Firebase Console, then now can be launched automatically after uploading a new binary.

### Scenarios Tested

Tested `firebaseappdistribution:distribute` with and without `--test-case-ids` flag.

### Sample Commands

```
firebase firebaseappdistribution:distribute --app "1:987654321:android:deadbeef" app.apk \
    --test-devices="model=MediumPhone.arm,version=34,locale=en,orientation=portrait" \
    --test-case-ids="case1,case2"
```

### DIFF
--- a/src/commands/appdistribution-distribute.ts
+++ b/src/commands/appdistribution-distribute.ts
@@ -9,6 +9,7 @@ import {
   IntegrationState,
   UploadReleaseResult,
   TestDevice,
+  ReleaseTest,
 } from "../appdistribution/types";
 import { FirebaseError, getErrMsg, getErrStatus } from "../error";
 import { Distribution, DistributionFileType } from "../appdistribution/distribution";
@@ -16,8 +17,8 @@ import {
   ensureFileExists,
   getAppName,
   getLoginCredential,
-  getTestDevices,
-  getTestersOrGroups,
+  parseTestDevices,
+  parseIntoStringArray,
 } from "../appdistribution/options-parser-util";
 
 const TEST_MAX_POLLING_RETRIES = 40;
@@ -35,19 +36,21 @@ function getReleaseNotes(releaseNotes: string, releaseNotesFile: string): string
 }
 
 export const command = new Command("appdistribution:distribute <release-binary-file>")
-  .description("upload a release binary")
+  .description(
+    "upload a release binary, optionally distribute it to testers and/or run automated tests",
+  )
   .option("--app <app_id>", "the app id of your Firebase app")
   .option("--release-notes <string>", "release notes to include")
   .option("--release-notes-file <file>", "path to file with release notes")
-  .option("--testers <string>", "a comma separated list of tester emails to distribute to")
+  .option("--testers <string>", "a comma-separated list of tester emails to distribute to")
   .option(
     "--testers-file <file>",
-    "path to file with a comma separated list of tester emails to distribute to",
+    "path to file with a comma- or newline-separated list of tester emails to distribute to",
   )
-  .option("--groups <string>", "a comma separated list of group aliases to distribute to")
+  .option("--groups <string>", "a comma-separated list of group aliases to distribute to")
   .option(
     "--groups-file <file>",
-    "path to file with a comma separated list of group aliases to distribute to",
+    "path to file with a comma- or newline-separated list of group aliases to distribute to",
   )
   .option(
     "--test-devices <string>",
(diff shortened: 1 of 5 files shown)


lfkellogg (approved):
Looks great!

src/commands/appdistribution-distribute.ts
```diff
@@ -210,56 +224,78 @@ export const command = new Command("appdistribution:distribute <release-binary-f
...
+              releaseName,
+              testDevices,
+              loginCredential,
+              `${appName}/testCases/${testCaseId}`,
+            ),
+          );
+        }
+      }
+      const releaseTests = await Promise.all(releaseTestPromises);
+      utils.logSuccess(`${releaseTests.length} release test(s) started successfully`);
```
Fry-kun, line 248: "[all] started successfully" -- what if some could not be started due to out-of-quota or other reasons?
There doesn't seem to be any handling of that; this phrasing might be deceptive.

kaibolay: From the perspective of the App Distribution API, the "release test" has been started as soon as `createReleaseTest()` returns successfully.

If the test matrix that's used "under the hood" fails for some reason (incl. out-of-quota), this will be caught below (line 279) - unless `--test-non-blocking` is specified. In that case the customer has to check the Firebase Console.

I understand that the TestLab API distinguishes the setup phase from running the actual test, but App Distribution doesn't.

src/commands/appdistribution-distribute.ts
```diff
@@ -210,56 +224,78 @@ export const command = new Command("appdistribution:distribute <release-binary-f
...
-        await awaitTestResults(releaseTest.name!, requests);
+        const releaseTestNames = new Set(releaseTests.map((rt) => rt.name!!));
+        await awaitTestResults(releaseTestNames, requests);
       }
     }
   });
 
 async function awaitTestResults(
-  releaseTestName: string,
+  releaseTestNames: Set<string>,
```
Fry-kun, line 257: Nit: I dislike passing by reference and modifying the original object - it would get messy if caller decides to reuse the object for anything else.
As a bonus, if you make a local copy, you could separate 1 vs many test cases & adjust the messages to use that.

src/commands/appdistribution-distribute.ts
```diff
@@ -210,56 +224,78 @@ export const command = new Command("appdistribution:distribute <release-binary-f
...
   requests: AppDistributionClient,
 ): Promise<void> {
   for (let i = 0; i < TEST_MAX_POLLING_RETRIES; i++) {
-    utils.logBullet("the automated tests results are pending");
+    utils.logBullet("the automated test results are pending");
```
joehan, line 261: Not new code, but log lines should generally be capitalized
```suggestion
    utils.logBullet("The automated test results are pending...");
```

src/commands/appdistribution-distribute.ts
```diff
@@ -210,56 +224,78 @@ export const command = new Command("appdistribution:distribute <release-binary-f
...
-    for (const execution of releaseTest.deviceExecutions) {
-      switch (execution.state) {
-        case "PASSED":
-        case "IN_PROGRESS":
+    for (const releaseTestName of releaseTestNames) {
+      const releaseTest = await requests.getReleaseTest(releaseTestName);
+      if (releaseTest.deviceExecutions.every((e) => e.state === "PASSED")) {
+        releaseTestNames.delete(releaseTestName);
+        if (releaseTestNames.size === 0) {
+          utils.logSuccess("automated test(s) passed!");
```
joehan, line 268:
```suggestion
          utils.logSuccess("Automated test(s) passed!");
```

joehan (approved):

src/commands/appdistribution-distribute.ts
kaibolay: Done

src/commands/appdistribution-distribute.ts
kaibolay: Done.

src/commands/appdistribution-distribute.ts
kaibolay: Done.

src/commands/appdistribution-distribute.ts
```diff
@@ -210,56 +224,78 @@ export const command = new Command("appdistribution:distribute <release-binary-f
...
+              `${appName}/testCases/${testCaseId}`,
+            ),
+          );
+        }
+      }
+      const releaseTests = await Promise.all(releaseTestPromises);
+      utils.logSuccess(`${releaseTests.length} release test(s) started successfully`);
       if (!options.testNonBlocking) {
-        await awaitTestResults(releaseTest.name!, requests);
+        const releaseTestNames = new Set(releaseTests.map((rt) => rt.name!!));
```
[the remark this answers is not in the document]
kaibolay: Unfortunately `?? ""` or something similar doesn't work, since there's no reasonable default that could be used. If the `ReleaseTest` resource returned by the API doesn't have the `name` property set, then the subsequent code can't work.

I could add an explicit check and error message, but that seems to be overkill. There are many other edge-cases which are much more deserving explicit error checking. This one here can't really happen - unless the API is completely b0rken.

I've replaced the `!!` with a single `!`

PR #8011, opened 2024-11-29
Fixing issue with firebase init genkit on Windows

ifielker:
Fixing issue with firebase init genkit on Windows.

### DIFF
--- a/src/extensions/runtimes/node.ts
+++ b/src/extensions/runtimes/node.ts
@@ -29,6 +29,26 @@ export const SDK_GENERATION_VERSION = "1.0.0";
 export const FIREBASE_FUNCTIONS_VERSION = ">=5.1.0";
 export const TYPESCRIPT_VERSION = "^4.9.0";
 
+/**
+ * execNpm runs the npm command with the given args.
+ * If it fails, it tries a Windows friendly version of the command.
+ * @param args The args to run npm with
+ * @return The output of the command as a string
+ */
+export function execNpm(args: readonly string[]): string {
+  try {
+    return execFileSync("npm", args).toString();
+  } catch (err: unknown) {
+    try {
+      // For Windows you might need "npm.cmd"
+      return execFileSync("npm.cmd", args).toString();
+    } catch (innerErr: unknown) {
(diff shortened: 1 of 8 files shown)


src/init/features/genkit/index.ts
```diff
@@ -61,7 +59,18 @@ async function getGenkitVersion(): Promise<GenkitInfo> {
...
+    try {
+      genkitVersion = await spawnWithOutput("npm", ["view", "genkit", "version"]);
+    } catch (err: unknown) {
+      throw new FirebaseError(
+        "Unable to determine which genkit version to install.\n" +
+          `npm Error: ${getErrMsg(err)}\n\n` +
+          "For a possible workaround run\n  npm view genkit version\n" +
+          "and then set an environment variable:\n" +
+          "  export GENKIT_DEV_VERSION=<output from previous command>\n" +
+          "and try the firebase init command again",
```
joehan, line 71:
```suggestion
          "and run `firebase init` again",
```

src/extensions/runtimes/node.ts
```diff
@@ -482,7 +483,7 @@ export async function writeSDK(
   // Build it
   logLabeledBullet("extensions", `running 'npm --prefix ${shortDirPath} run build'`);
   try {
-    execFileSync("npm", ["--prefix", dirPath, "run", "build"]);
+    await spawnWithOutput("npm", ["--prefix", dirPath, "run", "build"]);
```
joehan, line 486: Nit: Since we don't use the output, should we be using wrapSpawn in this file instead?

joehan (approved):
LGTM with some nits - thanks for throwing in some extra code cleanup too!

src/init/features/genkit/index.ts
ifielker: Done.

src/extensions/runtimes/node.ts
ifielker: wrapSpawn requires project directory which is not available here. We may not use the output, but it doesn't really matter. Maybe in future I can refactor wrapSpawn to have project directory be optional.

PR #8016, opened 2024-12-02
Fix load  when datasource is undefined

joehan:
### Description
Fix a 'cannot read properties of undefined' error when datasource is not set. Load gets used in some places where this is not relevant (ie `dataconnect:sdk;generate`), so we shouldn't error if this is missing, but rather handle missing fields at the call site when necessary.
Thanks for catching this Denver!


dconeybe (approved):
Gotta love those single-character bug fixes! Thanks for addressing this.

PR #8017, opened 2024-12-02
Add no-undeclared-imports eslint rule

joehan:
### Description
Add no-undeclared-imports eslint rule - this means importing something not declared in package.json (like an indirect dependency) will be a linter error. 

This also caught a few cases of us using types from indirect dependencies - this isn't a runtime error (since types get washed away during compile), but it is ugly, so I added those as devDeps.


joehan:
Broke the linter, will resend once i fix that.

bkendall (approved):

mathu97 (approved):

PR #8021, opened 2024-12-03
Adding better error messaging when using non Firebase projects

joehan:
### Description
Adding better error messaging when using projects that do not have Firebase enabled.


src/command.ts
```diff
@@ -415,12 +430,15 @@ export class Command {
...
-      const options = last(args);
-      await this.prepare(options);
+      const options = await this.prepare(last(args));
 
       for (const before of this.befores) {
         await before.fn(options, ...before.args);
       }
+      if (this.firebaseRequired && options.project && !(await isFirebaseProject(options.project))) {
+        printMigrateToFirebaseMessage(options.project);
+        throw new FirebaseError("This command requires your project to be migrated to Firebase.");
```
rachelsaunders, line 440:
```suggestion
        throw new FirebaseError("This command requires that your Google Cloud project has Firebase added to it.");
```

src/commands/deploy.ts
```diff
@@ -97,19 +101,38 @@ export const command = new Command("deploy")
...
+      options.filteredTargets.includes("functions") ||
+      options.filteredTargets.includes("hosting") ||
+      options.filteredTargets.includes("remoteconfig") ||
+      options.filteredTargets.includes("storage")
+    ) {
+      const projectId = needProjectId(options);
+      if (!(await isFirebaseProject(projectId))) {
+        printMigrateToFirebaseMessage(projectId);
+        throw new FirebaseError(
+          "Some of the products in this deployment require this project to be migrated to Firebase.",
```
rachelsaunders, line 124:
```suggestion
          "Some of the products in this deployment require that your Google Cloud project has Firebase added to it.",
```

src/utils.ts
```diff
@@ -42,6 +43,14 @@ export function consoleUrl(project: string, path: string): string {
   return `${api.consoleOrigin()}/project/${project}${path}`;
 }
 
+/**
+ * Create a Migrate to Firebase Console URL for the specified project.
+ */
+export function migrateToFirebaseLink(project: string): string {
```
rachelsaunders, line 49: Do you have flexibility on this verb "migrate"?  It's not really a "migrate to Firebase" but more like "add Firebase" if that's possible?

rachelsaunders (changes requested):

src/utils.ts
joehan: Definitely - I took 'migrate' from the query parameter in the console link, but if 'add Firebase' is more accurate lets use that

rachelsaunders: ya, from the dev perspective, it's messaged as "add Firebase"

e.g., https://firebase.google.com/docs/reference/firebase-management/rest/v1beta1/projects/addFirebase

rachelsaunders (changes requested):

rachelsaunders (approved):

PR #8023, opened 2024-12-03
Implement FDC connector evolution warnings.

rosalyntan:
### Description

Surface connector evolution warnings for Firebase Data Connect during `firebase deploy`, gated behind the `fdcconnectorevolution` experiment flag. See go/fdc-connector-evolution-cli for a detailed description of desired behavior.

### Scenarios Tested

- [x] `firebase deploy`
- [x] `firebase deploy --force`
- [x] `firebase deploy --non-interactive`
- [x] `firebase deploy --non-interactive --force`
- [x] `firebase deploy --dry-run`
- [x] `firebase deploy --dry-run --force`
- [x] `firebase deploy --dry-run --non-interactive`
- [x] `firebase deploy --dry-run --non-interactive --force`

### Sample Commands

`firebase deploy`
Screenshot 2024-12-18 at 4 23 17 PM


src/dataconnect/build.ts
```diff
@@ -1,15 +1,80 @@
...
-export async function build(options: Options, configDir: string): Promise<DeploymentMetadata> {
-  const buildResult = await DataConnectEmulator.build({ configDir });
+export async function build(
+  options: Options,
+  configDir: string,
+  dryRun?: boolean,
+): Promise<DeploymentMetadata> {
+  const args: { configDir: string; projectId?: string } = { configDir };
+  if (experiments.isEnabled("fdcconnectorevolution") && options.projectId) {
+    process.env["DATA_CONNECT_PREVIEW"] = "conn_evolution";
```
joehan, line 17: Can the emulator take multiple preview flags at once? If so, this should probably append the flag to whatever the current env var is

src/dataconnect/build.ts
```diff
@@ -1,15 +1,80 @@
...
-    throw new FirebaseError(
-      `There are errors in your schema and connector files:\n${buildResult.errors.map(prettify).join("\n")}`,
+    if (buildResult.errors.filter((w) => !w.extensions?.warningLevel).length) {
+      // Throw immediately if there are any build errors in the GraphQL schema or connectors.
+      throw new FirebaseError(
+        `There are errors in your schema and connector files:\n${buildResult.errors.map(prettify).join("\n")}`,
+      );
+    }
+    const interactiveAcks = buildResult.errors.filter(
+      (w) => w.extensions?.warningLevel && w.extensions?.warningLevel === "INTERACTIVE_ACK",
```
joehan, line 29: This is redundant currently - can simplify to
```suggestion
      (w) => w.extensions?.warningLevel === "INTERACTIVE_ACK",
```
because undefined !=INTERACTIVE_ACK

src/dataconnect/build.ts
```diff
@@ -1,15 +1,80 @@
...
+    );
+    const requiredAcks = buildResult.errors.filter(
+      (w) => w.extensions?.warningLevel && w.extensions?.warningLevel === "REQUIRE_ACK",
```
joehan, line 32:
```suggestion
      (w) => w.extensions?.warningLevel === "REQUIRE_ACK",
```

src/dataconnect/build.ts
```diff
@@ -1,15 +1,80 @@
...
+    ];
+    if (requiredAcks.length) {
+      utils.logLabeledWarning(
+        "dataconnect",
+        `There are changes in your schema or connectors that may break your existing applications. These changes require explicit acknowledgement to proceed. You may either reject the changes and update your sources with the suggested workaround(s), if any, or acknowledge these changes and proceed with the deployment:\n` +
+          prettifyWithWorkaround(requiredAcks),
+      );
+      if (options.nonInteractive && !options.force) {
+        throw new FirebaseError(
+          "Explicit acknowledgement required for breaking schema or connector changes.",
```
joehan, line 46: We should explicitly tell users to include --force to go forward with the change
```suggestion
          "Explicit acknowledgement required for breaking schema or connector changes. Rerun this command with --force to deploy these changes",
```

src/dataconnect/build.ts
```diff
@@ -1,15 +1,80 @@
...
+  const args: { configDir: string; projectId?: string } = { configDir };
+  if (experiments.isEnabled("fdcconnectorevolution") && options.projectId) {
+    process.env["DATA_CONNECT_PREVIEW"] = "conn_evolution";
+    args.projectId = options.projectId;
+  }
+  const buildResult = await DataConnectEmulator.build(args);
   if (buildResult?.errors?.length) {
-    throw new FirebaseError(
-      `There are errors in your schema and connector files:\n${buildResult.errors.map(prettify).join("\n")}`,
+    if (buildResult.errors.filter((w) => !w.extensions?.warningLevel).length) {
```
joehan, line 22: Now that we're adding some more conditional logic to this function, its probably worth separating this out into a 'handleBuildErrors' function and adding unit tests for it.

joehan (commented):
Mostly LGTM, just a few nits and one place that I think we should add unit tests.

src/dataconnect/build.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+import * as sinon from "sinon";
+import * as prompt from "../prompt";
+import { handleBuildErrors } from "./build";
+import { GraphqlError } from "./types";
+
+describe("handleBuildErrors", () => {
+  let promptOnceStub: sinon.SinonStub;
+  before(() => {
+    sinon.verifyAndRestore();
+  });
```
rosalyntan, line 11: @joehan not sure why, but this test fails with `Attempted to wrap promptOnce which is already wrapped` without this clause here, even though the same thing is called in `afterEach`.

joehan: Oooooh, that means we must have an unrestored stub somewhere in the code - let me see if i can find it :/

joehan: I think /src/apphosting/utils.spec.ts is the offender - if you add an afterEach  to that and call sinon.verifyAndRestore(), are you able to remove this?

joehan (approved):
LGTM if fixing /src/apphosting/utils.spec.ts lets us clean this up

src/dataconnect/build.spec.ts
rosalyntan: That fixed it, thanks!!

PR #8024, opened 2024-12-04
Make run local always present

hlshen:
Run local will start the emulator if user agrees.

### DIFF
--- a/firebase-vscode/src/data-connect/execution.ts
+++ b/firebase-vscode/src/data-connect/execution.ts
@@ -24,12 +24,14 @@ import { DataConnectError, toSerializedError } from "../../common/error";
 import { OperationLocation } from "./types";
 import { InstanceType } from "./code-lens-provider";
 import { DATA_CONNECT_EVENT_NAME } from "../analytics";
+import { EmulatorsController } from "../core/emulators";
 
 export function registerExecution(
   context: ExtensionContext,
   broker: ExtensionBrokerImpl,
   dataConnectService: DataConnectService,
   telemetryLogger: TelemetryLogger,
+  emulatorsController: EmulatorsController,
 ): Disposable {
   const treeDataProvider = new ExecutionHistoryTreeDataProvider();
   const executionHistoryTreeView = vscode.window.createTreeView(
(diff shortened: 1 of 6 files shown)


firebase-vscode/src/data-connect/terminal.ts
```diff
@@ -99,16 +98,25 @@ export function registerTerminalTasks(
...
       { focus: true },
     );
+  };
+  const startEmulatorsTaskBroker = broker.on("runStartEmulators", () => {
+    startEmulatorsTask();
   });
+  const startEmulatorsCommand = vscode.commands.registerCommand(
+    "firebase.emulators.start",
+    startEmulatorsTask,
+  );
```
fredzqm, line 114: Out of curiosity while I try to understand this code.
Excuse my ignorance. Very new to VS Code codebase.

What's the difference between `broker` and `vscode.commands`?

My guess is `broker` communicated any events between UI and extension process, so UI is never blocking. `vscode.commands` seem like something that can be triggered asynchronously from code, but so can `broken.send`

firebase-vscode/src/data-connect/terminal.ts
```diff
@@ -99,16 +98,25 @@ export function registerTerminalTasks(
...
       cmd += ` --export-on-exit ${settings.exportPath}`;
     }
+    vscode.window.showInformationMessage("Starting emulators... Please see terminal.");
     runTerminalTask(
       "firebase emulators",
       cmd,
       { focus: true },
     );
+  };
+  const startEmulatorsTaskBroker = broker.on("runStartEmulators", () => {
```
fredzqm, line 108: I saw `broker.on("runStartEmulators")` twice in the code base.

Here and [in emulator.ts](https://github.com/firebase/firebase-tools/blob/77620c041c76a2c6bb4f218e2dc70ab4f845adae/firebase-vscode/src/core/emulators.ts#L24-L26). Would it make sense to dedup them?

Move to `emulator.ts` and re-use `startEmulators` method?

firebase-vscode/src/data-connect/execution.ts
```diff
@@ -80,8 +82,43 @@ export function registerExecution(
...
+        "Trying to execute an operation on the emulator, but it isn't started yet. " +
+          "Do you want to start it?",
+        { modal: false }, // TODO: Consider making modal. blocked by testing infra
+        yes,
+        always,
+      );
+
+      // If the user selects "always", we update User settings.
+      if (result === always) {
+        configs.update(alwaysStartEmulator, true, ConfigurationTarget.Global);
```
fredzqm, line 106: If customer choose `alwaysStartEmulator` option, it sounds like we shouldn't show the above prompt again.

Shall we add a branch to skip it?

fredzqm (approved):
Thank you! Great step to improve discoverability of "run local"!

[Not blocking] Would leave it up to you~
I see this PR adds a warning to ask if user want to start the emulator.
I personally prefer to skip this question. The intend of "run local" is to run against the emulator. They have already made the intend super clear.
Starting local emulator won't break anything.

I think we can show a warning that says "Automatically started the emulator. Please retry "run local" after it's started."

firebase-vscode/src/data-connect/terminal.ts
hlshen: As you said, broker is used specifically for communicating between Webviews and the extension process. The Webviews have no concept of "VSCode", so they would not be able to call `vscode.comamnds` directly.

firebase-vscode/src/data-connect/execution.ts
hlshen: Good catch! I should add a test for that too. I somehow missed that logic.

fredzqm:
We brought this up in local tooling sync with Rose and Joe.

I think we agree that a warning w/o confirmation would suffice.

PR #8025, opened 2024-12-04
make adhoc files saved by default

hlshen:


maneesht (approved):
LGTM. Thanks for the change!

PR #8036, opened 2024-12-05
Adding a smoke test for the data connect emulator

joehan:
### Description
Adding a simple smoke test for the data connect emulator. This will mainly catch wrong checksum/bytes. We'll add more meaningful tests later.

### Scenarios Tested
Confirmed that the test passes on the current HEAD. Then, editted downloadableEmulators.ts to have a bad checksum, and confirmed that the test fails.


maneesht (commented):
I think this might be missing the script itself?

joehan:
> I think this might be missing the script itself?

Yep, forgot to 'git add' 🤦 Fixed

maneesht (approved):
LGTM. Thanks for the change!

joehan:
Windows tests are failing for no apparent reason, and I'm going slightly insane trying to debug them. Gonna submit running only on linux for now

PR #8039, opened 2024-12-06
Allow a Callable to include a Genkit Action annotation

inlined:
In preparation for Genkit 1.0, we are going to have a `onCallGenkit` function declaration in `firebase-functions/https`. This will be a "subclass" of callable functions.

To represent this, "Genkit callables" are represented in memory as a callableTriggered with a new "genkitAction" property. Actual Cloud Functions will be created with the label deployed-callable: true still, but will also include a 'genkit-action' label as well, which will eventually be used to annotate the Firebase Console.

To allow users to migrate from `@genkit-ai/firebase/functions:onFlow` to `firebase-functions/https:onCallGenkit` we need to relax the restrictions around converting function types in function updates. This change allows an HTTPS function to become a Callable function. I also noticed a bug where you could convert between any type of non-auth-context Firestore function to any type of auth-context Firestore function and fixed it.

Since the `firebase-functions` code is not yet written, I tested the CLI code with the following genkit 0.9-style code:

```
import {onFlow, noAuth } from "@genkit-ai/firebase/functions";
import { defineSecret } from "firebase-functions/params";
import { gemini15Flash, googleAI } from '@genkit-ai/googleai';
import { genkit, z } from 'genkit';

const genAIKey = defineSecret("GOOGLE_GENAI_API_KEY");

const ai = genkit({
    plugins: [googleAI()],
    model: gemini15Flash,
});

export const tellJoke = onFlow(ai, {
    name: "jokeTeller",
    inputSchema: z.string().optional(),
    outputSchema: z.string(),
    authPolicy: noAuth(),
    httpsOptions: {
        secrets: [genAIKey],
    }
}, async (type?: string) => {
    const { text } = await ai.generate(`Tell me a ${type ?? "dad"} joke`);
    return text;
});

/* Uncomment to deploy a callable trigger. Used to test for upgrade compatibility. */
// delete (tellJoke.__trigger as any).httpsTrigger;
// (tellJoke.__trigger as any).callableTrigger = { genkitAction: "flows/jokeTeller" };
```

### DIFF
--- a/src/functions/events/v2.ts
+++ b/src/functions/events/v2.ts
@@ -33,10 +33,6 @@ export const FIRESTORE_EVENTS = [
 
 export const FIREALERTS_EVENT = "google.firebase.firebasealerts.alerts.v1.published";
 
-export const FIRESTORE_EVENT_REGEX = /^google\.cloud\.firestore\.document\.v1\.[^\.]*$/;
-export const FIRESTORE_EVENT_WITH_AUTH_CONTEXT_REGEX =
-  /^google\.cloud\.firestore\.document\.v1\..*\.withAuthContext$/;
-
 export type Event =
   | typeof PUBSUB_PUBLISH_EVENT
   | (typeof STORAGE_EVENTS)[number]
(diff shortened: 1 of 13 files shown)


inlined:
Note: to get this code to work, I had to add a couple of extra configs to my tsconfig.json:
```
"module": "NodeNext",
"esModuleInterop": true,
"moduleResolution": "nodenext",
"skipLibCheck": true,
```

I really really don't like the last one because it's just ignoring a valid compiler error in Genkit, but the other ones should possibly be added to our template configs?

src/functions/events/v2.ts
```diff
@@ -46,3 +42,17 @@ export type Event =
   | typeof TEST_LAB_EVENT
   | (typeof FIRESTORE_EVENTS)[number]
   | typeof FIREALERTS_EVENT;
+
+// Why can't auth context be removed? This is map was added to correct a bug where a regex
+// allowed any non-auth type to be converted to any auth type, but we should follow up for why
+// a functon can't opt into reducing PII.
+export const CONVERTABLE_EVENTS: Partial<Record<Event, Event>> = {
```
taeold, line 49: why is this called "covertable" when we use this mapping to check whether event is not convertable?

taeold (approved):

src/functions/events/v2.ts
inlined: Because it holds the list of events that _are_ convertable. Not convertable means not in the map

PR #8049, opened 2024-12-09
ensureServiceIsConnectedToCloudSql only need to call backend when schemaValidation==NONE

fredzqm:
[low priority] Minor tweak.

`postgresql.schemaValidation !== "NONE"` can be used to detect if backend has been linked to Cloud SQL.

We previously used `=== "STRICT"` because at that time only `NONE` and `STRICT` were valid, so it's equivalent to `!== "NONE"`.

We forgot to tweak this when adding compatible mode.


joehan (approved):

fredzqm:
Thanks for the review! Will verify locally before merging.

PR #8051, opened 2024-12-10
Fixing init dataconnect on some empty schema cases

joehan:
### Description
Fixes an issue where `firebase init dataconnect` would error out when importing services that have schemas with no GQL files. Thanks @maneesht for debugging this with me!

### DIFF
--- a/src/init/features/dataconnect/index.spec.ts
+++ b/src/init/features/dataconnect/index.spec.ts
@@ -1,5 +1,6 @@
 import * as sinon from "sinon";
 import { expect } from "chai";
+import * as fs from "fs-extra";
 
 import * as init from "./index";
 import { Config } from "../../../config";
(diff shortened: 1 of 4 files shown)


src/init/features/dataconnect/index.ts
```diff
@@ -196,7 +197,11 @@ async function writeFiles(config: Config, info: RequiredInfo) {
     for (const f of info.schemaGql) {
       await config.askWriteProjectFile(join(dir, "schema", f.path), f.content);
     }
+  } else {
+    // Even if the schema is empty, lets give them an emty .gql file to get started.
+    fs.ensureFileSync(join(dir, "schema", "schema.gql"));
```
maneesht, line 202: Do you think it's worth also filling in that schema with the commented out schema? I could go either way

src/init/features/dataconnect/index.ts
```diff
@@ -196,7 +197,11 @@ async function writeFiles(config: Config, info: RequiredInfo) {
     for (const f of info.schemaGql) {
       await config.askWriteProjectFile(join(dir, "schema", f.path), f.content);
     }
+  } else {
+    // Even if the schema is empty, lets give them an emty .gql file to get started.
```
maneesht, line 201: Nit:
```suggestion
    // Even if the schema is empty, lets give them an empty .gql file to get started.
```

maneesht (approved):

src/init/features/dataconnect/index.spec.ts
```diff
@@ -33,6 +36,7 @@ describe("init dataconnect", () => {
       expectedSource: string;
       expectedFiles: string[];
       expectCSQLProvisioning: boolean;
+      expectEnsureSchemaGQL: boolean;
```
maneesht, line 39: No change required, but this could've been optional, and you wouldn't have to change any of the existing test-cases

PR #8065, opened 2024-12-14
Change how FDC determines whether creating a CSQL free trial is allowed.

rosalyntan:
### Description

Currently, the CLI determines whether it can create a CSQL free trial instance by querying for the presence of the free trial metric. However, some projects (mostly internal testing projects) have been allowlisted to have unlimited free trial instances, and creating a CSQL instance via the CLI fails for these projects if they previously had a free trial instance.

This PR updates the provisioning logic use the backend error for determining whether a free trial instance can be created, instead of just the presence of the free trial metric.

### Scenarios Tested

- [x] `firebase deploy` for a non-allowlisted project that previously had a free trial instance
- [x] `firebase deploy` for an allowlisted project that previously had a free trial instance


fredzqm (approved):
Thanks @rosalyntan for the change!

This makes sense. So `firebase init` will continue to use the metric to determine whether or not to show the free trial option, but `firebase deploy` will try it first.

joehan (approved):

PR #8086, opened 2024-12-30
fix dot env files in windows

### LINKED ISSUE #8076: dotenv files ignored when deploying nextjs apps

**firebase-tools: 13.29.1**

**Platform: Windows**

### [REQUIRED] Test case

Nextjs project with .env.* files 

1. Create new nextjs project
2. Create .env.production file
3. Set up firebase
4. Create "production" alias
5. Deploy

Expected firebase-tools to pick up .env.production

.env.production is not picked up

### Problem

[Here ](https://github.com/firebase/firebase-tools/blob/4e7b4925e9b929d71eea4e9481f412161881055a/src/frameworks/index.ts#L528 )we're supposed to find all environment variable files. The issue is that glob [doesn't support backslashes](https://github.com/isaacs/node-glob?tab=readme-ov-file#windows). They're instead used for escaping something or another and node's path.join (in getProjectPath) does return paths with backslashes.

### Potential solution

Replace all backslashes with forwards 

`const getProjectPath = (...args: string[]) => join(projectRoot, source, ...args).replaceAll("\\","/");`

leoortizz:
### Description

Fixes #8076 by telling glob to use Windows path delimiters on Windows.

### Scenarios Tested

Tested with an `.env.production` file on Windows

### Sample Commands

firebase deploy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,3 @@
 - Fixed a Data Connect emulator issue where prepared statements would be persisted after terminated connections.
 - Added a warning when deploying a Genkit function without a secret as this is likely a bug (#8138)
+- Fixed `.env.*` files for web frameworks in Windows (#8086)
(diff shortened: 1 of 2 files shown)


chalosalvador (approved):
LGTM

PR #8097, opened 2025-01-04
Override `nan` to support node 22 (and earlier)

### LINKED ISSUE #8096: Continuing build error from older `ajv` under Node 22. Needs updated native interfaces.
NOTE: This is a continuation of #8088 and #8091 There were still problems building due to the older 6.x `ajv` embedded in the angular tools. That can't be upgraded due to API incompatibilities, but there's a better fix. On inspection, the root cause is that `nan` -- the node native interfaces -- was at 2.17.0 which not generating the right calls for modern v8 (see, e.g. https://github.com/firebase/firebase-tools/blob/ba3fa9954591723f33c2c890217d8407db71fc0a/npm-shrinkwrap.json#L14639C5-L14644C7). The current release is 2.22.0 and is backwards compatible all the way back to node 10.

The simplest fix is to `override` the `nan` package to its current version. 

P.S. I don't regret the earlier `ajv` fix as it has much nicer Typescript support now ;)

**firebase-tools:** 13.29.1 (with or without #8091 patch)

**Platform:** Nix on MacOS 15.2, Nixos Linux-x86_64 with Node 22 LTS

### [REQUIRED] Test case

Build from source under Node 22 on Nix. Easiest way to see this is to use nix: `nix-build -A firebase-tools`

1. Check out clean repo
2. rm node_modules and cache
3. `nvm use 22 --lts`
4. Build it

It should build.

Failures building ajv and node-gyp due to generating incompatible code for V8 in Node 22

sarahec:
### Description

Overrides `nan` (node native interfaces) to its current 2.22.0. This adds support for node 22 for any native V8 interfaces and fixes a remaining 6.x ajv compilation issue due to the angular tools.

Pay particular attention to the `npm-shrinkwrap` diffs -- you can see the nan upgrade happening in exactly two places.

Fixes #8096 

### Scenarios Tested

All tests not requiring a local firebase project. Also, this compiles on nixos with the sandbox (which builds all from source including extensions -- it's a very strict build which also runs the tests.) 

### Sample Commands

n/a

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -14637,9 +14637,10 @@
       }
     },
     "node_modules/nan": {
-      "version": "2.17.0",
-      "resolved": "https://registry.npmjs.org/nan/-/nan-2.17.0.tgz",
-      "integrity": "sha512-2ZTgtl0nJsO0KQCjEpxcIr5D+Yv90plTitZt9JBfQvVJDS5seMl3FOvsh3+9CoYWXf/1l5OaZzzF6nDm4cagaQ==",
+      "version": "2.22.0",
+      "resolved": "https://registry.npmjs.org/nan/-/nan-2.22.0.tgz",
+      "integrity": "sha512-nbajikzWTMwsW+eSsNm3QwlOs7het9gGJU5dDZzRTQGk03vyBOauxgI4VakDzE0PtsGTmXPsXTbbjVhRwR5mpw==",
+      "license": "MIT",
       "optional": true
     },
     "node_modules/nanoid": {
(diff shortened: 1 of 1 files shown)


sarahec:
One question that came up on the Nix side: should I just run `npm upgrade nan` to change the shrinkwrap and not add anything in the package.json? 

In my mind, the `override` in package.json makes it explicit that we're using a different (newer) `nan` than we would otherwise get if we regenerated the shrinkwrap from scratch (but I could be wrong).

joehan:
> One question that came up on the Nix side: should I just run `npm upgrade nan` to change the shrinkwrap and not add anything in the package.json?
> 
> In my mind, the `override` in package.json makes it explicit that we're using a different (newer) `nan` than we would otherwise get if we regenerated the shrinkwrap from scratch (but I could be wrong).

I think I prefer just changing the shrinkwrap here, (scoping the override to just `re2`) - it previously had `nan:"^2.17.0"`for the affected dependency, which should resolve to the largest 2.x.y version on a fresh regen. 

The override also adds a bit of potential tech debt - if/when a dependency switches to nan@3.x.y, this would pin us to 2.x.x and potentially break things.

sarahec:
Agreed. Removed the `override` keyword and kept the shrinkwrap changes.

sarahec:
Rebased atop master

PR #8103, opened 2025-01-06
Fix relative dataDir issue

joehan:
### Description

Resolve dataDir and other paths relative to firebase.json instead of cwd. Fixes b/384546990

### Scenarios Tested
Pulled down the web data connect codelab, and added a dataDir to firebase.json. Confirmed that it runs successfully from the directory with firebase.json, and also from ./app


fredzqm (approved):
Thank you Joe!

PR #8106, opened 2025-01-08
Fixing some warnings due to union types

joehan:
### Description
Fixing up some issues from the recent 'ajv' upgrade - namely:
- AllowUnionTypes:true - typescript-json-schema can generate these, so we need to allow them.
- Actually depend on ajv-formats - this slipped past our no-undeclared-imports check because it was a require before.

### Scenarios Tested
Before:
Screenshot 2025-01-08 at 10 01 13 AM
After:
Screenshot 2025-01-08 at 10 01 20 AM

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -14,6 +14,7 @@
         "@google-cloud/pubsub": "^4.5.0",
         "abort-controller": "^3.0.0",
         "ajv": "^8.17.1",
+        "ajv-formats": "3.0.1",
         "archiver": "^7.0.0",
         "async-lock": "1.4.1",
         "body-parser": "^1.19.0",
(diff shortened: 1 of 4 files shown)


bkendall:
Looks like you've got some issues to sort out before this is ready... :/

joehan:
> Looks like you've got some issues to sort out before this is ready... :/

I believe most of these failures were actually caused by an NPM outage today - rerunning the tests now to confirm. Will fix and rerequest later if there are still issues

joehan:
The test failures are unrelated and on non-blocking tests. The VSCode failures should be fixed by https://github.com/firebase/firebase-tools/pull/8107, and the data connect e2e test is flaky.

bkendall (approved):
Seems fine to me - though there's a lot removed from firebase-config, so maybe just confirm that that's cool

joehan:
> Seems fine to me - though there's a lot removed from firebase-config, so maybe just confirm that that's cool

Sanity checked it and it looks like it is still working correctly. The reduced file length is because the new version of typescript-json-schema makes (much) better use of $ref to avoid redundancies.

PR #8118, opened 2025-01-15
Suggest DNS friendly service ids

joehan:
### Description
`init dataconnect` should suggest DNS friendly service ides.

### DIFF
--- a/src/init/features/dataconnect/index.ts
+++ b/src/init/features/dataconnect/index.ts
@@ -141,7 +141,9 @@ async function askQuestions(setup: Setup, isBillingEnabled: boolean): Promise<Re
       }))
     );
   } else {
-    info.serviceId = info.serviceId || basename(process.cwd());
+    // Ensure that the suggested name is DNS compatible
+    const defaultServiceId = toDNSCompatibleId(basename(process.cwd()));
+    info.serviceId = info.serviceId || defaultServiceId;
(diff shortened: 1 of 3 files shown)


src/init/features/dataconnect/index.ts
```diff
@@ -447,3 +449,20 @@ async function locationChoices(setup: Setup) {
...
 }
+
+/**
+ * Converts any string to a DNS friendly service ID.
+ */
+export function toDNSCompatibleId(id: string): string {
+  let defaultServiceId = basename(id)
+    .toLowerCase()
+    .replaceAll(/[^a-z0-9-]/g, "")
+    .slice(0, 61);
```
lisajian, line 460: We allow resource names up to 63 characters~

lisajian (commented):
Sorry if I missed it in our convo yesterday, but would it be possible to suggest DNS friendly connector IDs too? Thanks!

joehan:
> Sorry if I missed it in our convo yesterday, but would it be possible to suggest DNS friendly connector IDs too? Thanks!

We already do! We either use the existing connector names (when importing from prod) or 'default' as the connector id

src/init/features/dataconnect/index.ts
joehan: Derp, I saw 61 in the regex and turned off my brain. Fixed.

lisajian (approved):
Thanks Joe!

PR #8127, opened 2025-01-22
Remove label annotations for genkit functions; they break

inlined:
Ran into an error testing out `onCallGenkit`. Turns out that Run labels can only have [a-z0-9_-]. We need both capital letters and slashes. Run annotations supports that, but they are unavailable until after a migration to CRF.

### DIFF
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -42,6 +42,9 @@ export interface HttpsTriggered {
 
 /** API agnostic version of a Firebase callable function. */
 export type CallableTrigger = {
+  // NOTE: This is currently unused because GCF 2nd gen labels do not support
+  // the characterset that may be in a genkit action name.
+  // This should be set as a Cloud Run attribute once we move to Cloud Run Functions.
   genkitAction?: string;
 };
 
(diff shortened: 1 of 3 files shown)


joehan (approved):
Change seems fine to me, though I don't know the use case here. If it makes sense here, consider `label encoding` the genkit action name instead of omitting it entirely?

inlined:
I'm not sure encoding is the best idea because the encoded value will be visible to the users through the Cloud console. I can chat with the genkit team and re-add if they want (since it would provide deployment metrics, though the pipeline would be torn down and rebuilt in CRF). In the meantime I'd like to just unblock the launch.

Not adding relnotes because it's a bugfix of an unreleased feature. LMK if you disagree and I'll add them

joehan:
> I'm not sure encoding is the best idea because the encoded value will be visible to the users through the Cloud console. I can chat with the genkit team and re-add if they want (since it would provide deployment metrics, though the pipeline would be torn down and rebuilt in CRF). In the meantime I'd like to just unblock the launch.
> 
> Not adding relnotes because it's a bugfix of an unreleased feature. LMK if you disagree and I'll add them

SGTM on both points

PR #8136, opened 2025-01-23
Support node 22

joehan:
### Description
Fixes a bunch of places where our dependencies were using builtin methods that were deprecated in Node 22.

TODO before merge:

- [x] Switch to a real version of superstatic instead of pointing at a git branch
- [x] Add node 22 to CI 

### Scenarios Tested
Ran the tests, and ran through a bunch of common commands without getting deprecation notices.


joehan:
Started to move Ci to node 22, but doing it piecemeal:

1.  In this PR, added 22 for the 'cheap' tests to verify they work. 
2. Next, I'll make these required checks (and remove the redundant versions)
3. In a follow up PR, i'll move the integration tests over, and confirm that they work. I'll force merge that and then update the branch protection rules with the new required checks.

src/unzip.spec.ts
```diff
@@ -13,7 +13,7 @@ describe("unzip", () => {
   });
 
   after(async () => {
-    await fs.promises.rmdir(tempDir, { recursive: true });
+    await fs.promises.rm(tempDir, { recursive: true });
```
bkendall, line 16: just checking that this is intentional :)

bkendall (approved):

src/unzip.spec.ts
joehan: Yep, `rmdir` recursive was deprecated (and the deprecation notice says to use this instead.

PR #8138, opened 2025-01-25
Add prompt if user deploys a genkit function without a secret

inlined:
Redundant warning with the SDK after @taeold made a good point that the error isn't visible during deployment. Probably worth keeping the other one around anyway since this doesn't warn if the function already existed.

### DIFF
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -524,14 +524,11 @@ export async function warnIfNewGenkitFunctionIsMissingSecrets(
   );
 
   if (newAndMissingSecrets.length) {
-    const msgBase =
+    const message =
       `The function(s) ${newAndMissingSecrets.map((e) => e.id).join(", ")} use Genkit but do not have access to a secret. ` +
       "This may cause the function to fail if it depends on an API key. To learn more about granting a function access to " +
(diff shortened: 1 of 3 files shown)


src/deploy/functions/prepare.ts
```diff
@@ -495,3 +498,41 @@ export async function loadCodebases(
...
+      `The function(s) ${newAndMissingSecrets.map((e) => e.id).join(", ")} use Genkit but do not have access to a secret. ` +
+      "This may cause the function to fail if it depends on an API key. To learn more about granting a function access to " +
+      "secrets, see https://firebase.google.com/docs/functions/config-env?gen=2nd#secret_parameters.";
+    if (options.nonInteractive) {
+      throw new FirebaseError(msgBase);
+    }
+    if (!(await prompt.confirm({ message: `${msgBase} Continue?` }))) {
+      throw new FirebaseError("Aborted");
+    }
+  }
```
taeold, line 537: nit* prompts.confirm has logic to short-circuit if nonInteractive is passed. So could be re-written to:

```ts
    logLabledBullet(msgBase);
    if (!(await prompt.confirm({ nonInteractive: options.nonInteractive, force: options.force }))) {
      throw new FirebaseError("Aborted");
    }
```

Slightly different behavior in messaging.

taeold (approved):

PR #8145, opened 2025-01-27
fix: support `default` and `reqHandler` exports in Angular

alan-agius4:
Updated the export structure to ensure both `default` and `reqHandler` exports function when using Angular.


src/frameworks/angular/index.ts
```diff
@@ -243,19 +243,24 @@ exports.handle = function(req,res) {
 };\n`;
   } else if (serverOutputPath) {
     bootstrapScript = `
-    const app = new Promise((resolve) => {
+    const app = new Promise((resolve, reject) => {
       setTimeout(() => {
```
alan-agius4, line 247: JFYI: The setTimeout here seems to be quite brittle, and I doubt it is longer needed.

jamesdaniels (approved):
PTAL @leoortizz. This should get us up and running with the Angular 19's default server.ts

9kubczas4 (approved):

src/frameworks/angular/index.ts
```diff
@@ -36,7 +36,7 @@ export const docsUrl = "https://firebase.google.com/docs/hosting/frameworks/angu
 
 const DEFAULT_BUILD_SCRIPT = ["ng build"];
 
-export const supportedRange = "16 - 18";
+export const supportedRange = "16 - 19";
```
leoortizz, line 39: @jamesdaniels FYI

leoortizz:
thanks @alan-agius4! I added changelog and increased the supported Angular version to 19.

PR #8150, opened 2025-01-29
Fdc brownfield setup

tammam-g:
### Description
Support brownfield database setup. When we detect the database schema already have other tables (brownfield or greenfield but then a table was added manually) we ask the user if they want to setup the database as greenfield. 
1- They say yes -> We transfer schema and table ownerships to firebase roles, and grant previous roles writer access only.
2- They say no -> We setup the database in brownfield mode which means that deployed connectors have writer permissions but SQL migration will stop working through cli.

Note: We should find a way to automatically turn on compat mode when database is setup as greenfield from brownfield to avoid accidentally dropping important tables.

Note 2: I think I can improve the brownfield experience by allowing firebase deploy to run if the database diff isn't incompatible (say only drops tables and remove fields). Right now sql:migrate will fail if database is found as brownfield which will fail firebase deploy.

#### Fresh greenfield setup
Setup greenfield database from scratch using firebase deploy/sql:migrate -> no change

#### Brownfield to greenfield setup
I created one database with one table (not owned by firebase) and then user agrees to set it up as greenfield.

Screenshot 2025-01-31 at 4 13 02 PM
 
Screenshot 2025-01-31 at 4 14 41 PM

#### Brownfield setup test

I create a new database with one table not owned by firebase and firebase isn't setup.
Screenshot 2025-01-31 at 4 30 57 PM

User runs `firebase deploy`, we then detect it's brownfield and ask the user if they want to setup the database as greenfield. -> User chooses to keep project as brownfield which sets up firebase writer/reader roles and p4sa access but deployment fails.

Screenshot 2025-01-31 at 4 31 38 PM

Confirming the user which has firebase writer role now can edit and select but can't create new tables.
Screenshot 2025-01-31 at 4 31 56 PM

Screenshot 2025-01-31 at 4 37 58 PM

#### BrownField Default Permissions Test
Finally, I test default permissions when database is setup. Firebase writer/reader should be able to see and edit tables created by other users in the schema ("public") after the setup is done in brownfield mode.

Screenshot 2025-01-31 at 4 40 50 PM

#### Additional flags test
Trying `--force` when migrating a  brownfield database fails.

Screenshot 2025-01-31 at 4 50 53 PM

### Sample Commands
`firebase dataconnect:sql:setup`


src/gcp/cloudsql/permissions.ts
```diff
@@ -183,27 +213,52 @@ function readerRolePermissions(databaseId: string, superuser: string, schema: st
...
+}
 
-    // Set reader defaults for new tables.
-    // Only the owner of the schema can set defaults.
-    `SET ROLE = '${firebaseowner(databaseId)}';`,
-    `ALTER DEFAULT PRIVILEGES IN SCHEMA "${schema}" GRANT SELECT ON TABLES TO "${firebaseReaderRole}";`,
-    `ALTER DEFAULT PRIVILEGES IN SCHEMA "${schema}" GRANT SELECT, USAGE ON SEQUENCES TO "${firebaseReaderRole}";`,
-    `ALTER DEFAULT PRIVILEGES IN SCHEMA "${schema}" GRANT EXECUTE ON FUNCTIONS TO "${firebaseReaderRole}"`,
-    `SET ROLE = cloudsqlsuperuser`,
+function defaultPermissions(databaseId: string, schema: string, ownerRole: string) {
```
fredzqm, line 219: Nice cleanup.

Would it make sense to port those changes to backend code as well to keep them reasonably aligned?

I know they should do the same thing for green field.

To make this sync easier in the future, it might make sense to split caller logics to separate files.

We can split logics that aren't copied to the backend [permissions.go](https://source.corp.google.com/piper///depot/google3/firebase/dataconnect/shared/sqlpermission/permissions.go;l=53?q=f:permission&ss=piper%2FGoogle%2FPiper:google3%2Ffirebase%2Fdataconnect%2F) into `permissions_setup.ts`. It can have `getSchemaMetaData‎ `, `greenFieldSchemaSetup`, `brownfieldSqlSetup`, etc.

src/gcp/cloudsql/permissions.ts
```diff
@@ -221,22 +276,316 @@ export async function setupSQLPermissions(
...
+  const projectNumber = await needProjectNumber(options);
+  const { user: fdcP4SAUser } = toDatabaseUser(getDataConnectP4SA(projectNumber));
+
   const sqlRoleSetupCmds = concat(
     // For backward compatibality we sometimes need to revoke some roles.
     revokes,
 
     // We shoud make sure schema exists since this setup runs prior to executing the diffs.
-    [`CREATE SCHEMA IF NOT EXISTS "public"`],
+    [`CREATE SCHEMA IF NOT EXISTS "${schema}"`],
```
fredzqm, line 288: In the backend code, I moved this to [`ownerRolePermissions`](https://source.corp.google.com/piper///depot/google3/firebase/dataconnect/shared/sqlpermission/permissions.go;l=53?q=f:permission&ss=piper%2FGoogle%2FPiper:google3%2Ffirebase%2Fdataconnect%2F) because there is a chance for the schema to be created with `firebasesuperuser` as the owner, and then switch to `firebaseowner`.

src/gcp/cloudsql/permissions.ts
```diff
@@ -221,22 +276,316 @@ export async function setupSQLPermissions(
...
+  // We have determined firebase writer exists but schema/table owner isn't firebaseowner -> Brownfield
+  return {
+    name: schema,
+    owner: schemaOwner,
+    setupStatus: SchemaSetupStatus.BrownField,
+    tables: tables,
+  };
+}
+
+export async function getUserAccessLevel(
```
fredzqm, line 402: Is it used?

src/gcp/cloudsql/permissions.ts
```diff
@@ -221,22 +276,316 @@ export async function setupSQLPermissions(
...
+  schemaInfo: SchemaMetaData,
+  options: Options,
+  silent: boolean = false,
+) {
+  const schema = schemaInfo.name;
+
+  // Step 1: Grant firebasesuperuser access to the original owner
+  const uniqueTablesOwners = [...new Set(schemaInfo.tables.map((t) => t.owner))];
+  const grants = uniqueTablesOwners.map((owner) => `GRANT ${owner} TO ${FIREBASE_SUPER_USER}`);
+  await executeSqlCmdsAsSuperUser(options, instanceId, databaseId, grants, silent);
```
fredzqm, line 489: nit: any reason, this couldn't be executed in the same `executeSqlCmdsAsSuperUser` as below.

Each `executeSqlCmdsAsSuperUser` would call UpsertUser with random password. Adds a few seconds of delay.

src/gcp/cloudsql/permissions.ts
```diff
@@ -221,22 +276,316 @@ export async function setupSQLPermissions(
...
+        "Seems like the database is already setup. Would you like to rerun the setup process?",
+      ),
+      default: false,
+    });
+    if (continueSetup) {
+      await greenFieldSchemaSetup(instanceId, databaseId, schema, options, silent);
+      return true;
+    }
+  } else {
+    logger.info(`Detected schema "${schema}" isn't fully setup or setup as brownfield project.`);
```
fredzqm, line 548: nit: Shall we log `schemaInfo.setupStatus` here to help debugging?

src/gcp/cloudsql/permissions.ts
```diff
@@ -221,22 +276,316 @@ export async function setupSQLPermissions(
...
+  });
+
+  if (continueSetup) {
+    await setupBrownfieldAsGreenfield(instanceId, databaseId, schemaInfo, options, silent);
+    return true;
+  } else {
+    logger.info(
+      clc.yellow(
+        "Setting up database in brownfield mode.\n" +
+          "Note: SQL migrations can't be done through FDC in this mode.",
```
fredzqm, line 584: done through `clc.bold("firebase dataconnect:sql:migrate")`

Be specific about what command we are referring to.

src/dataconnect/schemaMigration.ts
```diff
@@ -351,10 +360,21 @@ async function handleIncompatibleSchemaError(args: {
         ${commandsToExecuteBySuperUser.join("\n")}`);
     }
 
-    // TODO (tammam-g): at some point we would want to only run this after notifying the admin but
-    // until we confirm stability it's ok to run it on every migration by admin user.
-    if (userIsCSQLAdmin) {
-      await setupIAMUsers(instanceId, databaseId, options);
+    const schemaInfo = await getSchemaMetaData(instanceId, databaseId, DEFAULT_SCHEMA, options);
```
fredzqm, line 363: Should we check `setupStatus` the first thing in `handleIncompatibleSchemaError`?

When `handleIncompatibleSchemaError` is called, we have detected some diffs in `sql:migrate`. We should surface any permission setup error before any SQL migration specific errors.

fredzqm (changes requested):
Great work! Excited to have brown field support very soon!

src/gcp/cloudsql/permissions.ts
```diff
@@ -221,22 +276,316 @@ export async function setupSQLPermissions(
...
+  }
+
+  // We need to setup the database
+  if (schemaInfo.tables.length === 0) {
+    logger.info(`Found no tables in schema "${schema}", assuming greenfield project.`);
+    await greenFieldSchemaSetup(instanceId, databaseId, schema, options, silent);
+    return true;
+  }
+
+  if (options.nonInteractive || options.force) {
```
joehan, line 558: Should we include 'force' here? It seems like unexpected behavior to throw if --force is passed in interactive mode.

src/gcp/cloudsql/permissions.ts
```diff
@@ -221,22 +276,316 @@ export async function setupSQLPermissions(
...
+    throw new FirebaseError(
+      `Schema "${schema}" isn't setup and can only be setup in interactive mode.`,
+    );
+  }
+  const currentTablesOwners = [...new Set(schemaInfo.tables.map((t) => t.owner))];
+  logger.info(
+    `We found some existing object owners [${currentTablesOwners.join(", ")}] in your cloudsql "${schema}" schema.`,
+  );
+
+  const continueSetup = await confirm({
```
joehan, line 568: Nit: 'continueSetup' seems like a misnomer here - if its false, we still call brownfield SQL setup. Rename to 'shouldSetupGreenfield'?

src/gcp/cloudsql/permissions.ts
joehan: +1 to splitting this up into multiple files - I'd love to keep the prompting logic in a separate file from the SQL command logic

src/gcp/cloudsql/permissions.ts
joehan: Unused, please remove.

src/gcp/cloudsql/permissions.ts
```diff
@@ -221,22 +276,316 @@ export async function setupSQLPermissions(
...
+  schemaInfo: SchemaMetaData,
+  options: Options,
+  silent: boolean = false,
+): Promise<boolean> {
+  const schema = schemaInfo.name;
+  // Step 0: Check current user can run setup and upsert IAM / P4SA users
+  logger.info(`Attempting to Setup SQL schema "${schema}".`);
+  const userIsCSQLAdmin = await iamUserIsCSQLAdmin(options);
+  if (!userIsCSQLAdmin) {
+    throw new FirebaseError(`Only users with 'roles/cloudsql.admin' can setup SQL schemas.`);
```
joehan, line 532:
```suggestion
    throw new FirebaseError(`Missing required IAM permission to setup SQL schemas. SQL schema setup requires 'roles/cloudsql.admin' or an equivalent role.`);
```

Just to clarify that custom roles (or owner/editor) also works

joehan (changes requested):

src/gcp/cloudsql/permissions.ts
tammam-g: I want to replace checkSQLRoleIsGranted with this. Will make it a separate PR.

src/gcp/cloudsql/permissions.ts
tammam-g: I move `setupSQLPermissions` to the command file `dataconnect-sql-setup.ts`, this way interactivity is not part of permissions and it's within the command file which makes sense to me. I do import the `setupSQLPermissions` in `schemaMigration.ts` but I also feel this make sense, it basically makes it clear that we are running the setup workflow within the migration.

src/gcp/cloudsql/permissions.ts
tammam-g: No reason, good suggestion. I think ideally we can have executeSqlCmdsAsSuperUser create a transaction on all the passed commands. Will make brownfield setup more robust against say a user cancelling midway of our sql execution.

src/gcp/cloudsql/permissions.ts
tammam-g: Added, I think most people only use the deploy command which runs the sql:migrate command so that's why I thought to keep it open ended.

src/gcp/cloudsql/permissions.ts
```diff
@@ -221,22 +276,316 @@ export async function setupSQLPermissions(
...
+    databaseId,
+    /** cmd=*/ [
+      `SELECT
+    CASE
+        WHEN EXISTS (SELECT 1 FROM pg_namespace WHERE nspname = '${schema}')
+        THEN TRUE
+        ELSE FALSE
+    END AS schema_exists,
+    (SELECT pg_get_userbyid(nspowner) FROM pg_namespace WHERE nspname = '${schema}') AS schema_owner;`,
+    ],
```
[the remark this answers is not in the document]
tammam-g: Made the change. It does make reading the returned row less intuitive.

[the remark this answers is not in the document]
tammam-g: Simplified it in the same function instead of making another helper.

src/gcp/cloudsql/connect.ts
```diff
@@ -200,18 +201,6 @@ export async function setupIAMUsers(
   );
```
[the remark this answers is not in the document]
tammam-g: Done

src/dataconnect/schemaMigration.ts
```diff
@@ -237,10 +237,19 @@ export async function grantRoleToUserInSchema(options: Options, schema: Schema)
   }
 
   // Run the database roles setup. This should be idempotent.
-  await setupIAMUsers(instanceId, databaseId, options);
+  const schemaInfo = await getSchemaMetaData(instanceId, databaseId, DEFAULT_SCHEMA, options);
+  if (schemaInfo.setupStatus === SchemaSetupStatus.GreenField) {
```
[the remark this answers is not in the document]
tammam-g: Thanks for catching this, I forgot to test the grants command.

src/gcp/cloudsql/permissions.ts
tammam-g: I think it makes sense because this is a one time thing and it's important the customer acknowledges the changes, if we force yes the change is irreversible.

src/dataconnect/schemaMigration.ts
tammam-g: I think we should but let's leave it outside this PR since this changes the current greenfield behaviour.

src/gcp/cloudsql/permissions.ts
tammam-g: Done

fredzqm:
> Note 2: I think I can improve the brownfield experience by allowing firebase deploy to run if the database diff isn't incompatible (say only drops tables and remove fields). Right now sql:migrate will fail if database is found as brownfield which will fail firebase deploy.

It should already work in the way that you want.

`sql:migrate` talks to backend to determine if there are any diffs. If there are no IncompatibleSchemaError returned, it would actually skip any migration steps. Brown field customer can use `firebase deploy` to deploy FDC Schema & Connectors as long as there is no diff.

src/commands/dataconnect-sql-setup.ts
```diff
@@ -0,0 +1,132 @@
...
+      message: clc.yellow(
+        "Seems like the database is already setup. Would you like to rerun the setup process?",
+      ),
+      default: false,
+    });
+    if (rerunSetup) {
+      await greenFieldSchemaSetup(instanceId, databaseId, schema, options, silent);
+      logger.info(clc.green("Database setup complete."));
+      return true;
+    }
```
fredzqm, line 58: If `!rerunSetup`, should we abort?

Honestly, don't find this prompt that valuable.

If they run `sql:setup`, the intent to setup SQL is already clear. It feels redundant to ask them again.

Logging is enough here.

src/gcp/cloudsql/permissions.ts
fredzqm: `commands/**` should only have scaffolding for each comment.

Any re-usable logics should be defined in the `dataconnect/*` package.


For the sake of making it easier to map it to the [backend permissions.go](https://source.corp.google.com/piper///depot/google3/firebase/dataconnect/shared/sqlpermission/permissions.go;l=53?q=f:permission&ss=piper%2FGoogle%2FPiper:google3%2Ffirebase%2Fdataconnect%2F), can you move any logics not in backend to a separate file?

src/commands/dataconnect-sql-setup.ts
```diff
@@ -0,0 +1,132 @@
...
+} from "../gcp/cloudsql/permissions";
+import { setupIAMUsers } from "../gcp/cloudsql/connect";
+import { logger } from "../logger";
+import { getIdentifiers } from "../dataconnect/schemaMigration";
+import { confirm } from "../prompt";
+import * as clc from "colorette";
+
+// Sets up all FDC roles (owner, writer, and reader).
+// Granting roles to users is done by the caller.
+// Returns true if schema is setup as greenfield
```
fredzqm, line 28: "Return true" makes it hard to read.

Consider returns a union:
`SchemaSetupStatus.GreenField | SchemaSetupStatus.BrownField`

src/dataconnect/schemaMigration.ts
```diff
@@ -236,11 +236,28 @@ export async function grantRoleToUserInSchema(options: Options, schema: Schema)
...
   }
 
-  // Run the database roles setup. This should be idempotent.
-  await setupIAMUsers(instanceId, databaseId, options);
+  // Make sure we have the right setup for the requested role grant.
+  const schemaInfo = await getSchemaMetaData(instanceId, databaseId, DEFAULT_SCHEMA, options);
+  let isGreenfieldSetup = schemaInfo.setupStatus === SchemaSetupStatus.GreenField;
+  if (schemaInfo.setupStatus === SchemaSetupStatus.GreenField) {
+    logger.info(`Detected schema "${DEFAULT_SCHEMA}" is setup in greenfield mode. Skipping Setup.`);
+  } else if (
```
fredzqm, line 244: nit: use switch case

src/commands/dataconnect-sql-setup.ts
tammam-g: Done

src/dataconnect/schemaMigration.ts
tammam-g: Done

fredzqm (approved):
Thanks you! Looks good.

I trust you to test it well. There are loads of edge cases.

Let double check to make sure it doesn't introduce regression to green field onboard before submisson.

joehan:
> Thanks you! Looks good.
> 
> I trust you to test it well. There are loads of edge cases.
> 
> Let double check to make sure it doesn't introduce regression to green field onboard before submisson.

+1 - this feature is complex enough that its probably worth a bug bash (offline one if you prefer)

src/gcp/cloudsql/permissions_setup.ts
```diff
@@ -0,0 +1,384 @@
...
+import { logger } from "../../logger";
+import { confirm } from "../../prompt";
+import * as clc from "colorette";
+import { FirebaseError } from "../../error";
+import { needProjectNumber } from "../../projectUtils";
+import { executeSqlCmdsAsIamUser, executeSqlCmdsAsSuperUser, getIAMUser } from "./connect";
+import { concat } from "lodash";
+import { getDataConnectP4SA, toDatabaseUser } from "./connect";
+
+export type TableMetaData = {
```
joehan, line 23: Nit: MetaData -> Metadata - it is one word, not two.

Applies for schemaMetaData too, and the getter functions

src/gcp/cloudsql/permissions_setup.ts
```diff
@@ -0,0 +1,384 @@
...
+  const userIsCSQLAdmin = await iamUserIsCSQLAdmin(options);
+  if (!userIsCSQLAdmin) {
+    throw new FirebaseError(
+      `Missing required IAM permission to setup SQL schemas. SQL schema setup requires 'roles/cloudsql.admin' or an equivalent role.`,
+    );
+  }
+  await setupIAMUsers(instanceId, databaseId, options);
+
+  if (schemaInfo.setupStatus === SchemaSetupStatus.GreenField) {
+    logger.info(`Seems like the database is already setup. Reruning the setup process.`);
```
joehan, line 119:
```suggestion
    logger.info(`Database ${databaseId} has already been setup. Rerunning setup to repair any missing permissions.`);
```

src/gcp/cloudsql/permissions_setup.ts
```diff
@@ -0,0 +1,384 @@
...
+  if (schemaInfo.tables.length === 0) {
+    logger.info(`Found no tables in schema "${schema}", assuming greenfield project.`);
+    await greenFieldSchemaSetup(instanceId, databaseId, schema, options, silent);
+    logger.info(clc.green("Database setup complete."));
+    return SchemaSetupStatus.GreenField;
+  }
+
+  if (options.nonInteractive || options.force) {
+    throw new FirebaseError(
+      `Schema "${schema}" isn't setup and can only be setup in interactive mode.`,
```
joehan, line 136:
```suggestion
      `Schema "${schema}" isn't set up and can only be set up in interactive mode.`,
```
Had to look this one up to confirm, but generally setup is a noun, set up is a verb: https://grammarist.com/spelling/set-up-vs-setup/

joehan (approved):
LGTM with some minor wording nits

PR #8151, opened 2025-01-29
Add handling for REQUIRE_FORCE issues.

rosalyntan:

### DIFF
--- a/src/dataconnect/build.ts
+++ b/src/dataconnect/build.ts
@@ -41,6 +41,17 @@ export async function handleBuildErrors(
       `There are errors in your schema and connector files:\n${errors.map(prettify).join("\n")}`,
     );
   }
+
+  const requiredForces = errors.filter((w) => w.extensions?.warningLevel === "REQUIRE_FORCE");
+  if (requiredForces.length) {
+    utils.logLabeledError(
+      "dataconnect",
+      `There are changes in your schema or connectors that will result in broken behavior:\n` +
+        prettifyWithWorkaround(requiredForces),
+    );
+    throw new FirebaseError("Rerun this command with --force to deploy these changes.");
+  }
+
   const interactiveAcks = errors.filter((w) => w.extensions?.warningLevel === "INTERACTIVE_ACK");
   const requiredAcks = errors.filter((w) => w.extensions?.warningLevel === "REQUIRE_ACK");
   const choices = [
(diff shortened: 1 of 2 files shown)


src/dataconnect/graphqlError.ts
```diff
@@ -17,6 +17,8 @@ export function prettifyWithWorkaround(errs: GraphqlError[]): string {
   const table = new Table({
     head: ["Issue", "Workaround", "Reason"],
     style: { head: ["yellow"] },
+    colWidths: [75, 75, 75],
```
rosalyntan, line 20: @joehan not sure if there's a better way of determining a default column size. Is there a good way to get the terminal window size programmatically?

src/dataconnect/graphqlError.ts
```diff
@@ -1,5 +1,5 @@
 import { GraphqlError } from "./types";
-const Table = require("cli-table");
+const Table = require("cli-table3");
```
joehan, line 2: This most likely should be cli-table, not cli-table3 - cli-table3 looks like a transitive dependency (since it is only listed in shrinkwrap, not in package.json) and therefore we can't use it directly. 

Node is weird, and this will work on you local mahcine, but will break once you ship it. We have a linter check to catch these, but apparently it only catches imports, not requires

joehan (changes requested):
LGTM other than the cli-table3 issue

src/dataconnect/graphqlError.ts
rosalyntan: Ahh I changed it to `cli-table3` in order to use the `wordWrap` functionality (which doesn't look like it's available in `cli-table`), otherwise the error message and workarounds in the table are too long for the table to render properly.

Could I add `cli-table3` into package.json? Or is there a better way within `cli-table` to format things properly? FWIW from https://github.com/cli-table/cli-table3/tree/master it looks like `cli-table` is unmaintained and `cli-table3` is the current version.

joehan: Oooh, if they are API compatible lets just get rid of our dependency on cli-table and use cli-table3. Would you be willing to upgrade it across the codebase so that we don't have 2 redundant deps?

rosalyntan: Done in #8176

joehan (approved):

PR #8158, opened 2025-01-30
Added Generated React Bindings

maneesht:


src/init/features/dataconnect/sdk.ts
```diff
@@ -135,14 +140,41 @@ export function generateSdkYaml(
...
+      packageJsonDir,
     };
+    const packageJson = await resolvePackageJson(appDir);
+    if (packageJson) {
+      const frameworksUsed = getFrameworksFromPackageJson(packageJson);
+      frameworksUsed.forEach((framework) => {
+        javascriptSdk[framework] = true;
+      });
+    }
+    // const unusedFrameworks = SUPPORTED_FRAMEWORKS.filter(framework => !frameworksUsed.includes(framework));
```
maneesht, line 159: There's code here that prompts the user if we don't detect angular or react to see if they want to add one. I ended up removing it as it creates complexity with the call hierarchy since `prompt` requires `Setup`

maneesht: Alternatively, this could go into `askQuestions`

CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- Added code generation of react hooks for Data Connect
```
```suggestion
- Added code generation of React hooks for Data Connect
```

src/dataconnect/fileUtils.ts
```diff
@@ -152,3 +160,48 @@ export async function getPlatformFromFolder(dirPath: string) {
...
+export const SUPPORTED_FRAMEWORKS: (keyof SupportedFrameworks)[] = ["react"];
+function lookForFramework(
+  key: "dependencies" | "devDependencies",
+  packageJson: PackageJSON,
+  frameworksUsed: (keyof SupportedFrameworks)[],
+) {
+  if (!packageJson[key]) {
+    return frameworksUsed;
+  }
+  Object.keys(packageJson[key]!).forEach((dep) => {
```
joehan, line 188: This currently mutates the frameworksUsed arg - would prefer to not have side effect driven code. Maybe simplify to something like:

```
function getFrameworksFromPackageJson(  
  packageJson: PackageJSON,
): (keyof SupportedFrameworks)[] {
  const dependencies = Object.keys(packageJson.dependencies?? {});
  const devDependencies = Object.keys(packageJson.devDependencies ?? {})

  const matches = [...dependencies, ...devDependencies].filter(
     dep => SUPPORTED_FRAMEWORKS.includes(dep));
   // Using set to deduplicate saves us from the extra .includes check
   return [... new Set(matches)];
}
```

joehan (commented):
Some minor style suggestions and potential simplifications, but looking mostly good to me!

src/init/features/dataconnect/sdk.ts
```diff
@@ -101,12 +108,35 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
...
     targetPlatform,
     connectorYaml,
     connectorInfo.directory,
     appDir,
   );
+  if (targetPlatform === Platform.WEB) {
+    const unusedFrameworks = SUPPORTED_FRAMEWORKS.filter(
+      (framework) => newConnectorYaml!.generate?.javascriptSdk![framework],
+    );
+    if (unusedFrameworks.length > 0) {
```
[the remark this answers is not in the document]
joehan: 🤦‍♀️ you right, brain fart on my part.

joehan (approved):

PR #8162, opened 2025-02-01
Remove AbortController polyfill as it is no longer needed and use native fetch where possible

v1rtl:
### Description

This PR removes a redundant `AbortController` polyfill, since it's supported natively in Node.js since v14.18, and Firebase minimum target is 18. It also removes fetch polyfill where it's not needed anymore, however I had to leave it be in one place because of usage of custom HTTP agents, which native fetch doesn't support yet. node-fetch-native supports proxies but not custom http agents based on what I know.

### Scenarios Tested

I've run `npm test`, and a bunch of unrelated tests started failing (assertions). Will see what I can do. Linting passes though

### DIFF
--- a/src/apiv2.ts
+++ b/src/apiv2.ts
@@ -1,12 +1,9 @@
-import { AbortSignal } from "abort-controller";
 import { URL, URLSearchParams } from "url";
 import { Readable } from "stream";
 import { ProxyAgent } from "proxy-agent";
 import * as retry from "retry";
-import AbortController from "abort-controller";
 import fetch, { HeadersInit, Response, RequestInit, Headers } from "node-fetch";
 import util from "util";
-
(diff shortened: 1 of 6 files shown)


src/apiv2.ts
```diff
@@ -374,7 +371,7 @@ export class Client {
       reqTimeout = setTimeout(() => {
         controller.abort();
       }, options.timeout);
-      fetchOptions.signal = controller.signal;
+      fetchOptions.signal = controller.signal as RequestInit["signal"];
```
talentlessguy, line 374: this is due to node-fetch AbortController types not being up-to-date with the runtime AbortController which supports more features

MichaelDeBoey:
@fredzqm @joehan @taeold Anything you still want us to do in order to get this one merged?

PR #8169, opened 2025-02-05
fix: #8168 - enforce webframeworks only when needed

### LINKED ISSUE #8168: Hosting deploy wrongly assumes all-or-none webframeworks
### Environment info

**firebase-tools:**
13.30.0

**Platform:**
Ubuntu

### Test case

1. Create a firebase.json with two `hosting` targets: one using webframeworks and one normal (i.e. "target: framey" with `source` and "target: classic" with `public`).
2. Attempt `firebase deploy --only hosting:classic`

### Steps to reproduce

(same as test case above)

### Expected behavior

Deploy of "classic" target to succeed. Of course a deploy of `hosting:framey` should fail... but I wouldn't expect `classic` to fail because it doesn't use webframeworks.

### Actual behavior

Deployment errors out with "Cannot deploy a web framework from source because the experiment webframeworks is not enabled. To enable webframeworks run firebase experiments:enable webframeworks."

I suspect [this line](https://github.com/firebase/firebase-tools/blob/bfeeae01902b5a5b38831c3e4358872e62485022/src/deploy/index.ts#L72) is the problem. The tools shouldn't be asserting `webframeworks` whenever _any_ target requires it... it should only be asserting it for the specific target if it uses it.

fivecar:
### Description

In deployments where `--only hosting:boo` is used, enforce webframeworks enablement only when the target actually uses webframeworks. This PR:
- Refactors `prepareFrameworksIfNeeded` to make testing easier
- Introduces 5 unit tests
- Parses `--only` and matches webframeworks requirement only for webframeworks sites

Fixes #8168 

### Scenarios Tested

Created a firebase.json with one webframeworks site and one normal site.
- Disabled webframeworks and tried deploying `--only hosting:frameworkysite`, and verified error message
- Enabled webframeworks and tried the same deployment, verifying successful deploy
- Disabled webframeworks and deployed `--only hosting:nonframeworksite` and verified success
- Enabled webframeworks and deployed the same site, verifying success

And ran all unit tests clean.


src/deploy/index.ts
```diff
@@ -69,7 +69,17 @@ export const deploy = async function (
...
   if (targetNames.includes("hosting")) {
     const config = options.config.get("hosting");
-    if (Array.isArray(config) ? config.some((it) => it.source) : config.source) {
+    if (
+      Array.isArray(config)
+        ? config.some(
+            (it) =>
+              it.source &&
+              (!options.only.includes("hosting:") ||
+                new RegExp(`\\bhosting:${it.target}\\b`).exec(options.only)),
```
fivecar, line 78: I understand this is lame, and reflects my lack of understanding of all the config flags/etc and how best to parse this. But the point is to make this logic, "Enforce webframeworks if either all sites need to be deployed, or if the site being deployed itself has `source`").

fivecar:
@aalej : This is the PR I submitted for #8168, which you reproduced. Would you mind directing this to the right engineer to approve? Thank you!

aalej:
@fivecar, already shared this PR with our engineering team when I raised the issue. Thanks again for opening up this PR!

src/deploy/index.ts
chalosalvador: @fivecar I just pushed [this commit](https://github.com/firebase/firebase-tools/pull/8169/commits/00cc1fc2183bb62f31cb16928e438a4f6cb2d503) to improve readability and to handle the case when using `firebase deploy` which was erroring out.
All the functionality should remain the same.

chalosalvador (approved):
Thanks for working on this @fivecar 🚀

src/deploy/index.ts
fivecar: Awesome - it looks so much better, @chalosalvador . Thanks for your help and support!

PR #8174, opened 2025-02-05
fix(genkit): Better messages around running the Genkit sample.

ifielker:
* Improve messaging around how to login with gcloud prior to running the Genkit sample. 
* Add a description and default value for the sample flow

### DIFF
--- a/src/init/features/genkit/index.ts
+++ b/src/init/features/genkit/index.ts
@@ -124,6 +124,17 @@ export interface GenkitSetup extends Setup {
   [key: string]: unknown;
 }
 
+function showStartMessage(config: Config, command: string) {
+  logger.info("\nLogin to Google Cloud using:");
+  logger.info(
+    clc.bold(
+      clc.green(`    gcloud auth application-default login --project ${config.options.project}\n`),
+    ),
+  );
+  logger.info("Then start the Genkit developer experience by running:");
+  logger.info(clc.bold(clc.green(`    ${command}`)));
+}
+
 /**
  * doSetup is the entry point for setting up the genkit suite.
  */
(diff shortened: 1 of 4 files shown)


src/init/features/genkit/index.ts
```diff
@@ -186,8 +197,7 @@ export async function doSetup(setup: GenkitSetup, config: Config, options: Optio
           projectDir,
         );
         await genkitSetup(options, genkitInfo, projectDir);
-        logger.info("Start the Genkit developer experience by running:");
-        logger.info(`    cd ${setup.functions.source} && npm run genkit:start`);
+        showStartMessage(config, `cd ${setup.functions.source} && npm run genkit:start`);
```
inlined, line 200: Here and elsewhere s/genkit:start/genkit start

Also, since this is a functions integration, maybe it should be `genkit start -- firebase emulators:start --inspect-functions`?

inlined (approved):

src/init/features/genkit/index.ts
ifielker: I just tried that manually, and it doesn't seem to connect to the UI. Maybe we can add that as a second option in another PR once we figure out why it's not connecting?

ifielker: Also genkit:start is a script in the package.json. It runs this: `genkit start -- tsx --watch src/genkit-sample.ts`. I could have just printed that command, but I felt it was important to 1) Have a place where users could look up the command to run if they forget. 2) Have a way to have different commands for different versions. If say 1.0.0 has 'genkit start -- tsc --watch src/genkit-sample.ts' but maybe 2.0.0 has genkit someNewCommand index.ts. The the instructions don't change and people can still just run the familiar `npm run genkit:start` and it will do the right thing. And they can go into the package.json to see what the new right thing is. If we just print the command, then people might successfully run it, then close the window and then not remember how they ran it the first time. So I'd like to keep the script.

PR #8176, opened 2025-02-05
Move cli-table dep to cli-table3.

rosalyntan:
### Description

Per https://github.com/cli-table/cli-table3/blob/master/README.md, cli-table is unmaintained and cli-table3 is API-compatible with cli-table, and includes some desired features like word-wrapping options.


joehan (approved):
LGTM, tho we might want to also use import instead of require.

rosalyntan:
> LGTM, tho we might want to also use import instead of require.

Done! From chatting offline -- looks like it wasn't fully API-compatible (rip) so made a couple minor changes in a few files to get things to build.

src/emulator/extensionsEmulator.ts
```diff
@@ -324,7 +324,7 @@ export class ExtensionsEmulator implements EmulatorInstance {
         );
         table.push([
           apiToWarn.apiName,
-          apiToWarn.instanceIds,
+          apiToWarn.instanceIds.toString(),
```
joehan, line 327:
```suggestion
          apiToWarn.instanceIds.join(", "),
```
[].toString doesn't put spaces between elements, so its kind of ugly

joehan (approved):
LGTM with one minor thing! Thanks for taking this detour 😄

PR #8197, opened 2025-02-10
Update introspection api endpoint to V1

hlshen:
Data Connect APIs are swapping to v1. Updated the introspection endpoint.


lisajian (commented):
Do we need to update [firebase-vscode/src/test/integration/empty/init.ts](https://github.com/firebase/firebase-tools/blob/ab425647e2ec65d19c652fa68cc1a69f2dd032df/firebase-vscode/src/test/integration/empty/init.ts#L106)?

lisajian (approved):

PR #8215, opened 2025-02-13
Don't surface insecure operations errors in VSCode.

rosalyntan:
### Description

Currently, VSCode is flagging insecure operation errors in the editor. However, these errors are not currently accurate since the connector source isn't being compared against a deployed source, and can't differentiate between existing insecure operations and new insecure operations. This PR suppresses insecure operation issues in VSCode for now.

### DIFF
--- a/firebase-vscode/src/data-connect/core-compiler.ts
+++ b/firebase-vscode/src/data-connect/core-compiler.ts
@@ -48,6 +48,10 @@ function convertGQLErrorToDiagnostic(
   const perFileDiagnostics: Record<string, Diagnostic[]> = {};
   const dcPath = configs.values[0].path;
   for (const error of gqlErrors) {
+    if (error.warningLevel)  {
+      // Don't surface connector evolution or insecure operation issues for now; we need to be able to compare with a deployed source for these to have meaning.
+      continue;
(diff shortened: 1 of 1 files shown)


hlshen (approved):

fredzqm:
@rosalyntan 

Q: any reasons that VS Code linter couldn't poll latest FDC sources to do the diff?
Is it able to pick up application default credentials?

As a more general approach, maybe the emulator should skip the diff errors when it couldn't fetch latest backend sources as opposed to report incorrect errors.

rosalyntan:
> @rosalyntan
> 
> The insecure linter error is helpful when it is correct.
> 
> Q: any reasons that VS Code linter couldn't poll latest FDC sources to do the diff? Is it able to pick up application default credentials?

Polling the latest FDC sources will be done in b/396453481 -- this turns off the warnings until they're done as the existing INSECURE warnings are too noisy.
> 
> As a more general approach, maybe the emulator should skip returning diff warnings when it couldn't fetch latest backend sources as opposed to report incorrect warnings.

Yeah, though we'd need to differentiate between "couldn't fetch" due to NOT_FOUND, vs. other errors.

PR #8216, opened 2025-02-13
FDC Emulator Update

maneesht:
Fixes:
* Issue where react web app users get a crash when calling `firebase init dataconnect` due to misconfiguration of `prompt`
* Issue where web developers using versions lower than 11.3.0 of firebase get a missing export error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,3 @@
-- Fixed issue where react-based web apps were crashing when users initialize Data Connect.
-- Updated the Firebase Dat Connect local toolkit to v.1.8.0, which:
+- Fixed issue where `firebase init dataconnect` would crash on React-based web apps.
+- Updated the Firebase Data Connect local toolkit to v.1.8.1, which:
   - Fixed issue where users who are using a version lower than 11.3.0 of `firebase` get a "missing import" error.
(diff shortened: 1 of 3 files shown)


src/init/features/dataconnect/sdk.ts
```diff
@@ -116,24 +116,30 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
...
+              value: frameworkStr,
+              name: frameworkStr,
+              checked: false,
+            })),
           },
         ],
       );
-      for (const framework of additionalFrameworks.features) {
-        newConnectorYaml!.generate!.javascriptSdk![framework] = true;
+      if (additionalFrameworks.fdcFrameworks) {
```
joehan, line 139: Suprised this is needed - does prompt return undefined when you choose no fields in a multiselect?

joehan (approved):

CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+- Fixed issue where react-based web apps were crashing when users initialize Data Connect.
+- Updated the Firebase Dat Connect local toolkit to v.1.8.0, which:
```
```suggestion
- Updated the Firebase Data Connect local toolkit to v.1.8.1, which:
```

CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+- Fixed issue where react-based web apps were crashing when users initialize Data Connect.
```
```suggestion
- Fixed issue where `firebase init dataconnect` would crash on React-based web apps.
```

joehan (approved):

src/init/features/dataconnect/sdk.ts
maneesht: Ah no, I just removed that. I did found out that if the field is already defined (like I was using features before), then it just skips the prompt entirely: https://github.com/firebase/firebase-tools/blob/e8fc1ed459e18152106b6d9718147868f649566d/src/prompt.ts#L57

PR #8229, opened 2025-02-18
Print warning about --location removal.

annajowang:
### Description
Print warning about --location removal in next major version

### Scenarios Tested
Tested the following commands with and without --location:
apphosting:backends:get, apphosting:backends:create, apphosting:backends:delete,
apphosting:backends:list, apphosting:builds:create, apphosting:rollouts:create , apphosting:builds:get


src/commands/apphosting-backends-create.ts
```diff
@@ -23,9 +24,15 @@ export const command = new Command("apphosting:backends:create")
   .before(requireInteractive)
   .before(requireTosAcceptance(APPHOSTING_TOS_ID))
   .action(async (options: Options) => {
+    if (options.location !== undefined) {
```
taeold, line 27: nit

```suggestion
    if (options.location) {
```

taeold (approved):

PR #8236, opened 2025-02-20
Update vscode to 0.13.1

hlshen:
Update VSCode version to 0.13.1

- Updated internal `firebase-tools` dependency to 13.31.2
- [Fixed] Insecure operations no longer surfaced as errors.

### DIFF
--- a/firebase-vscode/CHANGELOG.md
+++ b/firebase-vscode/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## NEXT
 
+## 0.13.1
+
+- Updated internal `firebase-tools` dependency to 13.31.2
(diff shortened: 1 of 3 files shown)


firebase-vscode/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## NEXT
 
+## 0.13.1
+
+- Updated internal `firebase-tools` dependency to 13.31.2
+- [Fixed] Insecure operations no longer surfaced as errors.
```
rosalyntan, line 6: I think we can omit this from the changelog -- the issue this fixed is behind an experiment flag still, so it wouldn't have affected any developers.

PR #8237, opened 2025-02-20
Added env var to magically import data connect service from console

joehan:
### Description
Slightly hacky trick to make onboarding from console experience smoother - when `FDC_SERVICE_ID` is set to an existing service ID, we'll automatically assume you chose that one and proceed with `init`.

### DIFF
--- a/src/init/features/dataconnect/index.ts
+++ b/src/init/features/dataconnect/index.ts
@@ -30,10 +30,10 @@ const SCHEMA_TEMPLATE = readTemplateSync("init/dataconnect/schema.gql");
 const QUERIES_TEMPLATE = readTemplateSync("init/dataconnect/queries.gql");
 const MUTATIONS_TEMPLATE = readTemplateSync("init/dataconnect/mutations.gql");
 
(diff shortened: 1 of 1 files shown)


src/init/features/dataconnect/index.ts
```diff
@@ -19,7 +19,7 @@ import { Schema, Service, File, Platform } from "../../../dataconnect/types";
 import { parseCloudSQLInstanceName, parseServiceName } from "../../../dataconnect/names";
 import { logger } from "../../../logger";
 import { readTemplateSync } from "../../../templates";
-import { logBullet } from "../../../utils";
+import { logBullet, envOverride } from '../../../utils';
```
rosalyntan, line 22: nit: did these get changed from `"` to `'` intentionally?

rosalyntan (approved):
Nice, thanks!

src/init/features/dataconnect/index.ts
joehan: Nope, no clue why my auto import did that... fixing for cosnistency

PR #8257, opened 2025-02-25
Switch env var for isVSCE check

### LINKED ISSUE #7471: Incorrect path resolve when other VS Code extensions run firebase related command

**firebase-tools:** 13.13.3

**Platform:** macOS Sonoma 14.5

### [REQUIRED] Test case

Use Orta.vscode-jest extension, configure `firebase emulators:exec "jest"` as its command, run jest unit test.

1. Install Orta.vscode-jest extension through VS Code extension marketplace.
2. Create a bash file say `test.sh`, and put the following content into it:
    ```
    #!/bin/bash

    PARAMS="$@"
    yarn firebase emulators:exec "jest ${PARAMS}"
    ```
3. In VS Code `settings.json`, add `"jest.jestCommandLine": "./test.sh"` or put `./test.sh` into `Jest: Jest Command Line` in Jest extension settings.
4. Configure a jest unit test
5. Run jest unit test through CodeLens (i.e. clicking the run test button in the unit test file)

Unit test is passed.

Received error, test aborted:
```
TestRun "my-app:runTest: orta.vscode-jest:TestProvider:my-app:32 (0)" started
node:fs:448
    return binding.readFileUtf8(path, stringToFlags(options.flag));
                   ^

Error: ENOENT: no such file or directory, open '/Users/user/Projects/my-app/node_modules/firebase-tools/lib/templates/hosting/init.js'
    at readFileSync (node:fs:448:20)
    at readTemplateSync (/Users/user/Projects/my-app/node_modules/firebase-tools/lib/templates.js:17:34)
    at Object.<anonymous> (/Users/user/Projects/my-app/node_modules/firebase-tools/lib/hosting/implicitInit.js:12:56)
    at Module._compile (node:internal/modules/cjs/loader:1358:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1416:10)
    at Module.load (node:internal/modules/cjs/loader:1208:32)
    at Module._load (node:internal/modules/cjs/loader:1024:12)
    at Module.require (node:internal/modules/cjs/loader:1233:19)
    at require (node:internal/modules/helpers:179:18)
    at Object.<anonymous> (/Users/user/Projects/my-app/node_modules/firebase-tools/lib/frameworks/index.js:25:24) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/user/Projects/my-app/node_modules/firebase-tools/lib/templates/hosting/init.js'
}

Node.js v20.14.0
```

### Workaround
In fact, if I remove the following extension check in`src/templates.ts`:
```typescript
if (isVSCodeExtension()) {
    // In the VSCE, the /templates directory is copied into dist, which makes it
    // right next to the compiled files (from various sources including this
    // TS file). See CopyPlugin in `../firebase-vscode/webpack.common.js`.
    return resolve(__dirname, "templates", relPath);
}
```
the issue will be suppressed and the unit test is passed.

joehan:
### Description
`VSCODE_CWD` is set whenever you're running within VSCode, so its not safe to assume we've been webpacked if it is set. Instead, we define our own variable and use that.
Fixes #7471.

### DIFF
--- a/firebase-vscode/src/extension.ts
+++ b/firebase-vscode/src/extension.ts
@@ -93,7 +93,7 @@ async function checkCLIInstallation(): Promise<void> {
     const latestVersion = (await latestVersionRes.json())?.["dist-tags"]?.[
       "latest"
     ];
-    const env = { ...process.env, VSCODE_CWD: "" };
+    const env = { ...process.env, VSCODE_CWD: "", IS_FIREBASE_VSCE: "" };
     const versionRes = spawnSync("firebase", ["--version"], {
       env,
       shell: process.platform === "win32",
--- a/firebase-vscode/src/extension.ts
+++ b/firebase-vscode/src/extension.ts
@@ -25,6 +25,7 @@ import {
 import { env } from "./core/env";
 
 import { suggestGraphqlSyntaxExtension } from "./data-connect/graphql-syntax-highlighter";
+import { setIsVSCodeExtension } from "../../src/vsCodeUtils";
 
 // This method is called when your extension is activated
 export async function activate(context: vscode.ExtensionContext) {
@@ -93,7 +94,8 @@ async function checkCLIInstallation(): Promise<void> {
     const latestVersion = (await latestVersionRes.json())?.["dist-tags"]?.[
       "latest"
     ];
-    const env = { ...process.env, VSCODE_CWD: "", IS_FIREBASE_VSCE: "" };
+    setIsVSCodeExtension(true);
+    const env = { ...process.env, VSCODE_CWD: "" };
     const versionRes = spawnSync("firebase", ["--version"], {
       env,
       shell: process.platform === "win32",
--- a/src/hosting/implicitInit.ts
+++ b/src/hosting/implicitInit.ts
@@ -8,8 +8,6 @@ import { EmulatorRegistry } from "../emulator/registry";
 import { EMULATORS_SUPPORTED_BY_USE_EMULATOR, Emulators } from "../emulator/types";
 import { readTemplateSync } from "../templates";
 
-const INIT_TEMPLATE = readTemplateSync("hosting/init.js");
-
 export interface TemplateServerResponse {
   // __init.js content with only initializeApp()
   js: string;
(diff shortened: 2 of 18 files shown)


joehan:
Debugging this now, seems like this is causing extension activation to fail

joehan:
Should be working now - tested both through the VSCE development host and by manually packaging a VSIX and installing it

yuchenshi (commented):
This seems fine but it's too confusing with a real env var.

Have we considered using a global variable instead, such as `globalThis.WEBPACK_IS_VSCE`? Or, can we at least add comments to the code saying it's not set but just defined by webpack?

joehan:
> This seems fine but it's too confusing with a real env var.
> 
> Have we considered using a global variable instead, such as `globalThis.WEBPACK_IS_VSCE`? Or, can we at least add comments to the code saying it's not set but just defined by webpack?

I'll try out the global variable idea.

joehan:
Refactored as you suggested - it required a little bit of finagling to make work with our build/lint rules tho

firebase-vscode/src/extension.ts
```diff
@@ -93,7 +93,7 @@ async function checkCLIInstallation(): Promise<void> {
     const latestVersion = (await latestVersionRes.json())?.["dist-tags"]?.[
       "latest"
     ];
-    const env = { ...process.env, VSCODE_CWD: "" };
+    const env = { ...process.env, VSCODE_CWD: "", IS_FIREBASE_VSCE: "" };
```
yuchenshi, line 96: Revert

src/vsCodeUtils.ts
```diff
@@ -1,8 +1,10 @@
-import { env } from "process";
-
+// @eslint-disable @typescript-eslint/no-namespace
```
yuchenshi, line 1: Still bunch of linting issues. I wonder if I'm the one overcomplicating this.

I feel like `@ts-ignore` inline is the best way to go. We'd only do it one time anyway. 

The other idea is we can just code `isVSCodeExtension()` to  return false and then use the define the whole expression to true in webpack.

yuchenshi (approved):
General approach lgtm, see nits inline. Your call

src/vsCodeUtils.ts
joehan: Refactored this after a rabbit hole + pairing for a bit. TLDR webpack plugin ordering is a mess with ts-loader, and we aren't really sure why some webpack based approaches were broken. Rather than do something hacky, we went with a simple setter that's called from the activation function.

src/hosting/implicitInit.ts
```diff
@@ -8,8 +8,6 @@ import { EmulatorRegistry } from "../emulator/registry";
 import { EMULATORS_SUPPORTED_BY_USE_EMULATOR, Emulators } from "../emulator/types";
 import { readTemplateSync } from "../templates";
 
-const INIT_TEMPLATE = readTemplateSync("hosting/init.js");
```
joehan, line 11: Had to move this into the function so that it wasn't evaluated before activation.

yuchenshi: It's exactly this kind of stuff that makes me hate mutable state so much that I even awakened the mighty webpack. Well, at least we know what the tradeoffs are and we can deal with it another day

PR #8261, opened 2025-02-26
Split up emulator hub message to better match how its used

joehan:
### Description
This log line was misleading to some users, and it mainly used across 2 env vars (one for host, one for port).


rosalyntan (approved):

yuchenshi:
nit: It's two fields but one env var `FIREBASE_EMULATOR_HUB`. I don't have a problem with this change though

PR #8262, opened 2025-02-26
Remove `--location` from apphosting:backends:delete

annajowang:
### Description

Remove `--location` from apphosting:backends:delete. 

If more than one backend with a given Id is found, we let the users choose which ones they want to delete. The backends are deleted sequentially.

`--force` is not allowed if more than one backend is found.

### Scenarios Tested

`firebase apphosting:backends:delete -f friendlybugs-codelab` where there are 3 friendlybugs-codelab backends and a firebase errror was thrown.

`firebase apphosting:backends:delete friendlybugs-codelab` where two were selected to be deleted.

`firebase apphosting:backends:delete friendlybugs-codelab` where only one was left and the user was not made to select any backends.

### DIFF
--- a/src/apphosting/backend.ts
+++ b/src/apphosting/backend.ts
@@ -430,6 +430,64 @@ export async function getBackendForLocation(
 }
 
 /**
+ * Fetches backends of the given backendId and lets the user choose if more than one is found.
+ */
+export async function chooseBackends(
+  projectId: string,
+  backendId: string,
+  chooseBackendPrompt: string,
+  force?: boolean,
+): Promise<apphosting.Backend[]> {
+  let { unreachable, backends } = await apphosting.listBackends(projectId, "-");
+  if (unreachable && unreachable.length !== 0) {
+    logWarning(
+      `The following locations are currently unreachable: ${unreachable.join(",")}.\n` +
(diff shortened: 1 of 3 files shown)


src/apphosting/backend.ts
```diff
@@ -429,6 +429,64 @@ export async function getBackendForLocation(
...
+  backendId: string,
+  chooseBackendPrompt: string,
+  force?: boolean,
+): Promise<apphosting.Backend[]> {
+  let { unreachable, backends } = await apphosting.listBackends(projectId, "-");
+  if (unreachable && unreachable.length !== 0) {
+    logWarning(
+      `The following locations are currently unreachable: ${unreachable.join(",")}.\n` +
+        "If your backend is in one of these regions, please try again later.",
+    );
```
taeold, line 446: do we need to return here? or want to continue with rest of the command?

taeold (approved):

src/apphosting/backend.ts
annajowang: I think it's fine to continue. This is just a warning and the backend they care about may not even be impacted.  This also matches the behavior of `getBackendForAmbiguousLocation`

PR #8264, opened 2025-02-27
Remove --location from apphosting:backends:create and prompt for primary region

annajowang:
### Description

Remove --location from apphosting:backends:create and prompt for primary region

### Scenarios Tested

`firebase apphosting:backends:create`

`firebase apphosting:backends:create --location asia-east1` should return error

### DIFF
--- a/src/apphosting/backend.ts~
+++ b/src/apphosting/backend.ts~
@@ -88,9 +88,6 @@ export async function doSetup(
   // possible to reduce the likelihood that the subsequent Cloud Build fails. See b/336862200.
   await ensureAppHostingComputeServiceAccount(projectId, serviceAccount);
 
-  // TODO(https://github.com/firebase/firebase-tools/issues/8283): The "primary region"
-  // is still "locations" in the V1 API. This will change in the V2 API and we may need to update
-  // the variables and API methods we're calling under the hood when fetching "primary region".
   const location = await promptLocation(
     projectId,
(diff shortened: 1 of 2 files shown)


src/apphosting/backend.ts
```diff
@@ -89,17 +88,10 @@ export async function doSetup(
...
-    if (!allowedLocations.includes(location)) {
-      throw new FirebaseError(
-        `Invalid location ${location}. Valid choices are ${allowedLocations.join(", ")}`,
-      );
-    }
-  }
-
-  location =
-    location || (await promptLocation(projectId, "Select a location to host your backend:\n"));
+  const location = await promptLocation(
```
taeold, line 91: nit* should we rename var and helper func to primaryLocation to avoid confusion?

taeold (approved):

src/apphosting/backend.ts
annajowang: I had considered that, but ultimately decided not to because underneath the hood, promptLocations is calling the listLocations API method. So in the long term, even after any API changes, I would assume that "promptLocations" will always be prompting for locations and the API layer would just introduce a new "listPrimaryRegions" method. I think I would only call something promptPrimaryRegion if it's calling listPrimaryRegions under the hood.

But you're right this is confusing. I'm adding a TODO comment to explain things.

PR #8268, opened 2025-02-27
Add a new command to setup a cleanup policy for functions artifacts

taeold:
Add command `functions:artifacts:setpolicy`.

This new command leverages [Artifact Registry's cleanup policies](https://cloud.google.com/artifact-registry/docs/repositories/cleanup-policy) feature. Cleanup policies are a managed service feature that automatically removes container images based on configurable rules, providing a more robust and transparent solution.

The command simplifies the process of configuring these policies for Cloud Run functions repositories. By default, the command sets up a policy that deletes images older than 3 days. Users can easily adjust this retention period using the `--days` option.

The command handles checking if the user already has a cleanup policy set up. The cli command will check if a clean up policy exists, and if so, will prompt the user if they wish to proceed.


src/commands/functions-artifacts-setup-cleanup-policy.ts
```diff
@@ -0,0 +1,191 @@
...
+  .before(requireAuth)
+  .before(artifactregistry.ensureApiEnabled)
+  .before(requirePermissions, [
+    "artifactregistry.repositories.update",
+    "artifactregistry.versions.delete",
+  ])
+  .action(async (options: any) => {
+    const projectId = needProjectId(options);
+    const location = options.location || "us-central1";
+    const daysToKeep = parseInt(options.days || "5", 10);
```
inlined, line 46: documentation says "3" not 5

src/commands/functions-artifacts-setup-cleanup-policy.ts
```diff
@@ -0,0 +1,191 @@
...
+          `Please deploy your functions first using: ` +
+            `${clc.bold(`firebase deploy --only functions`)}`,
+        );
+        return;
+      }
+      throw err;
+    }
+
+    if (options.none) {
+      if (artifacts.hasCleanupOptOut(repository)) {
```
inlined, line 65: I wonder if we should also check that there is no policy. A customer may have opted out of the policy before, set a policy in the Cloud Console, and then assumed the CLI would be able to modify it.

src/commands/functions-artifacts-setup-cleanup-policy.ts
```diff
@@ -0,0 +1,191 @@
...
+        logBullet(
+          `Please deploy your functions first using: ` +
+            `${clc.bold(`firebase deploy --only functions`)}`,
+        );
+        return;
+      }
+      throw err;
+    }
+
+    if (options.none) {
```
inlined, line 64: It feels like there's a lot of content here that could be moved to a tested utility file. E.g. deleteCleanupPolicy, setCleanupPolicy

src/commands/functions-artifacts-setup-cleanup-policy.ts
```diff
@@ -0,0 +1,191 @@
...
+import { requirePermissions } from "../requirePermissions";
+import { requireAuth } from "../requireAuth";
+import { logBullet, logSuccess } from "../utils";
+import * as artifactregistry from "../gcp/artifactregistry";
+import * as artifacts from "../functions/artifacts";
+
+/**
+ * Command to set up a cleanup policy for Cloud Run functions container images in Artifact Registry
+ */
+export const command = new Command("functions:artifacts:setup-cleanup-policy")
```
inlined, line 16: Did we actually get approval to be the first command to use kebab-case? Also why setup-cleanup-policy and not set-cleanup-policy? This can be used post-setup.

src/commands/functions-artifacts-setup-cleanup-policy.ts
```diff
@@ -0,0 +1,191 @@
...
+  )
+  .withForce("Automatically create or modify cleanup policy")
+  .before(requireAuth)
+  .before(artifactregistry.ensureApiEnabled)
+  .before(requirePermissions, [
+    "artifactregistry.repositories.update",
+    "artifactregistry.versions.delete",
+  ])
+  .action(async (options: any) => {
+    const projectId = needProjectId(options);
```
inlined, line 44: Add a check that the user cannot specify both --days and --none

src/functions/artifacts.spec.ts
```diff
@@ -0,0 +1,318 @@
+import { expect } from "chai";
+import * as sinon from "sinon";
+
+import * as artifactregistry from "../gcp/artifactregistry";
+import * as artifacts from "./artifacts";
+
+describe("functions artifacts", () => {
+  describe("makeRepoPath", () => {
+    it("should construct a valid repo path", () => {
+      const path = artifacts.makeRepoPath("my-project", "us-central1");
```
inlined, line 10: IDK if I feel comfortable with this hard-coding the repository ID, especially since it might change with CRF

src/functions/artifacts.spec.ts
```diff
@@ -0,0 +1,318 @@
...
+
+  describe("generateCleanupPolicy", () => {
+    it("should generate a valid cleanup policy with the correct days", () => {
+      const policy = artifacts.generateCleanupPolicy(7);
+      expect(policy).to.have.property(artifacts.CLEANUP_POLICY_ID);
+      expect(policy[artifacts.CLEANUP_POLICY_ID].id).to.equal(artifacts.CLEANUP_POLICY_ID);
+      expect(policy[artifacts.CLEANUP_POLICY_ID].action).to.equal("DELETE");
+      expect(policy[artifacts.CLEANUP_POLICY_ID].condition).to.deep.include({
+        tagState: "ANY",
+        olderThan: "604800s", // 7 days in seconds
```
inlined, line 108: nit: I often find it easier to write something like `${60 * 24 * 7}s` to make it clear that the math is right.

src/commands/functions-artifacts-setup-cleanup-policy.ts
```diff
@@ -0,0 +1,191 @@
...
+        logSuccess(`Successfully opted out from cleanup policy for ${clc.bold(repoPath)}`);
+        return;
+      } catch (err: unknown) {
+        throw new FirebaseError("Failed to opt-out from artifact registry cleanup policy", {
+          original: err as Error,
+        });
+      }
+    }
+
+    if (isNaN(daysToKeep) || daysToKeep <= 0) {
```
inlined, line 114: I wonder if we should allow zero with a fudge factor (e.g. 5m)

src/functions/artifacts.ts
```diff
@@ -0,0 +1,140 @@
...
+export const OPT_OUT_LABEL_KEY = "firebase-functions-cleanup-opted-out";
+
+const SECONDS_IN_DAY = 24 * 60 * 60;
+
+/**
+ * Construct the full path to a repository in Artifact Registry
+ *
+ * @returns The full path to the repository
+ */
+export function makeRepoPath(projectId: string, location: string): string {
```
inlined, line 27: maybe add a 3rd parameter "repoName: string = GCF_REPO"

src/gcp/artifactregistry.ts
```diff
@@ -1,14 +1,32 @@
...
-export const API_VERSION = "v1beta2";
+export const API_VERSION = "v1";
 
 const client = new Client({
   urlPrefix: artifactRegistryDomain(),
   auth: true,
   apiVersion: API_VERSION,
 });
 
+export function ensureApiEnabled(options: any): Promise<void> {
```
inlined, line 15: nit: It would be cleaner if we just passed a projectId rather than propagating an untyped object.

src/gcp/artifactregistry.ts
```diff
@@ -21,8 +39,71 @@ export interface Operation {
...
+ */
+export async function getRepository(repoPath: string): Promise<Repository> {
+  const res = await client.get<Repository>(repoPath);
+  return res.body;
+}
+
+/**
+ * Update an Artifact Registry repository.
+ */
+export async function updateRepository(repo: Partial<Repository>): Promise<Repository> {
```
inlined, line 80: FYI: I've improved the state of the art here on what you can do instead of Partials.

```ts
type RepositoryOutputOnlyFields = 'foo', 'bar.baz';
assertImplements<RepsotoryOutputOnlyFields, DeepKeysOf<Repository>>();

// ...
export async function updateRepository(repo: DeepOmit<Repository, RepositoryOutputOnlyFields>): Promise<Repository> {
```

src/gcp/artifactregistry.ts
```diff
@@ -21,8 +39,71 @@ export interface Operation {
...
+    queryParams: { updateMask: updateMask.join(",") },
+  });
+  return res.body;
+}
+
+export type Permissions = { permissions: string[] };
+
+/**
+ * Test IAM permissions for an Artifact Registry resource.
+ *
```
inlined, line 96: This doesn't seem to be implemented anywhere; it may be worth yanking. testIamPermissions is meant for UX design and will fail open (suggesting the user has permissions they don't actually ahve)

src/commands/functions-artifacts-setup-cleanup-policy.ts
taeold: Oh I think the reviewers missed this - I didn't realize that there was no precedent for kebab-case.

src/commands/functions-artifacts-setup-cleanup-policy.ts
taeold: Fixed

src/commands/functions-artifacts-setup-cleanup-policy.ts
taeold: good call. fixed.

src/commands/functions-artifacts-setup-cleanup-policy.ts
taeold: Done!

src/commands/functions-artifacts-setup-cleanup-policy.ts
taeold: Good idea.

src/functions/artifacts.spec.ts
taeold: I feel okay since this is documented in GRF docs and has not changed since beginning of GCF.

src/commands/functions-artifacts-setup-cleanup-policy.ts
taeold: Like that.

src/functions/artifacts.ts
taeold: Done.

src/gcp/artifactregistry.ts
taeold: Very neato.

src/gcp/artifactregistry.ts
taeold: Removed.

src/commands/functions-artifacts-setup-cleanup-policy.ts
taeold: Resolved to rename the command `functions:artifacts:setpolicy`

src/functions/artifacts.ts
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * Extract an existing cleanup policy from the repository if it exists
+ *
+ * @returns The existing policy if found, undefined otherwise
+ */
+export function findExistingPolicy(
+  repository: artifactregistry.Repository,
+): artifactregistry.CleanupPolicy | undefined {
+  if (!repository.cleanupPolicies) {
```
inlined, line 43: Super nit: this can all be `return repository?.cleanupPolicies?.[CLEANUP_POLICY_ID]`

inlined (approved):
LGTM. I assume this can go in as reverse compatible and in the vNext branch you're going to take a swing at checking the policy on deployment and removing the ArtifactCleaner (and its related experiments)?

src/commands/functions-artifacts-setpolicy.ts
```diff
@@ -0,0 +1,178 @@
...
+    "--days <days>",
+    "Number of days to keep container images before deletion. Default is 3 days.",
+    "3",
+  )
+  .option(
+    "--none",
+    "Opt-out from cleanup policy. This will prevent suggestions to set up a cleanup policy during initialization and deployment.",
+  )
+  .before((options) => {
+    if (options.days && options.none) {
```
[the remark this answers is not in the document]
taeold: Since --days default to 3, we don't have to check that case!

src/functions/artifacts.ts
taeold: i like one liners.

taeold:
@inlined yep that's the plan. Added that work in go/firebase-tools-14, and will work on it soon.

PR #8271, opened 2025-02-27
Remove --location from apphosting:rollouts:create and error when more than one backend is found.

annajowang:
### Description

Remove --location from apphosting:rollouts:create and error when more than one backend is found.

### Scenarios Tested

`firebase apphosting:rollouts:create chickenbugs` where chickenbugs is the only backend with the name => should succeed
`firebase apphosting:rollouts:create chicken` where there are multiple backends with this name => should error
`firebase apphosting:rollouts:create chicken --location us-central1` => should complain about unknown option --location

### DIFF
--- a/src/apphosting/backend.ts
+++ b/src/apphosting/backend.ts
@@ -443,7 +443,7 @@ export async function getBackendForAmbiguousLocation(
   let { unreachable, backends } = await apphosting.listBackends(projectId, "-");
   if (unreachable && unreachable.length !== 0) {
     logWarning(
-      `The following locations are currently unreachable: ${unreachable}.\n` +
+      `The following locations are currently unreachable: ${unreachable.join(", ")}.\n` +
         "If your backend is in one of these regions, please try again later.",
     );
   }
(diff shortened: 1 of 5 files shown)


src/apphosting/backend.ts
```diff
@@ -474,3 +474,33 @@ export async function getBackendForAmbiguousLocation(
...
+  projectId: string,
+  backendId: string,
+): Promise<apphosting.Backend> {
+  let { unreachable, backends } = await apphosting.listBackends(projectId, "-");
+  backends = backends.filter(
+    (backend) => apphosting.parseBackendName(backend.name).id === backendId,
+  );
+  if (backends.length > 1) {
+    throw new FirebaseError(
+      `You have multiple backends with the same ${backendId} ID. This is not allowed until we can support more locations. ` +
```
taeold, line 492: since we already have the backends, think it would be nice to list out locations where the conflict is arising.

src/apphosting/backend.ts
```diff
@@ -474,3 +474,33 @@ export async function getBackendForAmbiguousLocation(
...
+    );
+  }
+  if (backends.length === 1) {
+    return backends[0];
+  }
+  if (unreachable && unreachable.length !== 0) {
+    logWarning(
+      `Backends with the following primary regions are unreachable: ${unreachable.join(", ")}.\n` +
+        "If your backend is in one of these regions, please try again later.",
+    );
```
taeold, line 503: should we return here? otherwise, it looks like we are throwing.

taeold (approved):

src/apphosting/backend.ts
annajowang: hm this was intentional on my end.
If no backend is found, no matter what, I wanted to throw an error. 

but IF there are unreachable locations I wanted to warn before throwing that error. the unreachable location may impact the backend the user wants to get, but it may also be completely unreleated.

PR #8277, opened 2025-03-02
Migrate to biomejs

taeold:


taeold:
biome 1 was not type aware.

PR #8285, opened 2025-03-03
List regions of duplicate backend in warning/error message.

annajowang:
### Description

List regions of duplicate backend in warning/error message.

### Scenarios Tested

`firebase apphosting:backends:get chicken` where there are multiple backends called chicken

`firebase apphosting:backends:get chickenbugs` where there's only one backend called chickenbugs

`firebase apphosting:rollouts:create chicken` where there are multiple backends called chicken


src/apphosting/backend.ts~
taeold, line 1: lol what is this haha

src/apphosting/backend.ts
```diff
@@ -541,8 +541,9 @@ export async function getBackend(
     (backend) => apphosting.parseBackendName(backend.name).id === backendId,
   );
   if (backends.length > 1) {
+    const regions = backends.map((b) => apphosting.parseBackendName(b.name).location);
```
taeold, line 544: nit* location for consistency?

taeold (approved):

PR #8305, opened 2025-03-10
Add support for secrets in the App Hosting emulator

inlined:
Part of an ongoing conversation. Not ready to review until that conversation resolves.

### DIFF
--- a/src/emulator/apphosting/serve.spec.ts
+++ b/src/emulator/apphosting/serve.spec.ts
@@ -10,6 +10,8 @@ import * as projectPathImport from "../../projectPath";
 import { AppHostingYamlConfig } from "../../apphosting/yaml";
 import * as emulatorRegistry from "../registry";
 import * as emulatorEnvs from "../env";
+import * as secrets from "../../gcp/secretManager";
+import { FirebaseError } from "../../error";
 
 describe("serve", () => {
   let checkListenableStub: sinon.SinonStub;
@@ -20,6 +22,7 @@ describe("serve", () => {
   let resolveProjectPathStub: sinon.SinonStub;
   let listRunningWithInfoStub: sinon.SinonStub;
   let setEnvVarsForEmulatorsStub: sinon.SinonStub;
+  let accessSecretVersionStub: sinon.SinonStub;
 
   beforeEach(() => {
     checkListenableStub = sinon.stub(portUtils, "checkListenable");
@@ -34,6 +37,8 @@ describe("serve", () => {
 
     resolveProjectPathStub.returns("");
     detectStartCommandStub.returns("npm run dev");
+
+    accessSecretVersionStub = sinon.stub(secrets, "accessSecretVersion");
   });
 
   afterEach(() => {
@@ -50,9 +55,7 @@ describe("serve", () => {
 
     it("should use user-provided port if one is defined", async () => {
       checkListenableStub.onFirstCall().returns(true);
-      configsStub.getLocalAppHostingConfiguration.returns(
-        Promise.resolve(AppHostingYamlConfig.empty()),
-      );
+      configsStub.getLocalAppHostingConfiguration.resolves(AppHostingYamlConfig.empty());
 
       const res = await serve.start({ port: 9999 });
       expect(res.port).to.equal(9999);
@@ -62,9 +65,8 @@ describe("serve", () => {
       checkListenableStub.onFirstCall().returns(false);
       checkListenableStub.onSecondCall().returns(false);
       checkListenableStub.onThirdCall().returns(true);
-      configsStub.getLocalAppHostingConfiguration.returns(
-        Promise.resolve(AppHostingYamlConfig.empty()),
-      );
+      configsStub.getLocalAppHostingConfiguration.resolves(AppHostingYamlConfig.empty());
+
       const res = await serve.start();
       expect(res.port).to.equal(DEFAULT_PORTS.apphosting + 2);
     });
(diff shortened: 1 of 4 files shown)


src/emulator/apphosting/serve.ts
```diff
@@ -36,12 +40,43 @@ export async function start(options?: StartOptions): Promise<{ hostname: string;
...
 
-  serve(port, options?.startCommand, options?.rootDirectory);
+  await serve(options?.projectId, port, options?.startCommand, options?.rootDirectory);
 
   return { hostname, port };
 }
 
+const secretResourceRegex =
+  /^projects\/([^/]+)\/secrets\/([^/]+)(?:\/versions\/((?:latest)|\d+))?$/;
+const secretShorthandRegex = /^([^/@]+)(?:@((?:latest)|\d+))?$/;
```
annajowang, line 50: Can there be comments that show examples of what these match

src/emulator/apphosting/serve.spec.ts
```diff
@@ -62,25 +65,108 @@ describe("serve", () => {
...
-      );
+      configsStub.getLocalAppHostingConfiguration.resolves(AppHostingYamlConfig.empty());
 
       await serve.start({ startCommand });
 
       expect(spawnWithCommandStringStub).to.be.called;
       expect(spawnWithCommandStringStub.getCall(0).args[0]).to.eq(startCommand);
     });
+
+    it("Should pass plaintext environment varialbes", async () => {
```
annajowang, line 85: "varialbes"

annajowang (approved):

src/emulator/apphosting/serve.ts
```diff
@@ -55,13 +90,17 @@ async function serve(
   for (const env of apphostingLocalConfig.environmentVariables) {
     environmentVariablesAsRecord[env.variable] = env.value!;
   }
+  await Promise.all(
+    apphostingLocalConfig.secrets?.map(async (secret) => {
```
taeold, line 94: We should update the logic for merge:

https://github.com/firebase/firebase-tools/blob/bbc33ec1d3130893030daa8e0f17be11e01f9aef/src/apphosting/yaml.ts#L85

Where we remove secrets if they are overridden by an env var in apphosting.local.yaml

Otherwise, there is no way to override the secret to prevent it from fetching it during emulator start.

inlined (commented):
Based on Daniel's catch that env wasn't merging right, I did the cleanup that I was delaying for another CL.

IDK why the appyhosting yaml file was made to look like idiomatic java instead of idiomatic javascript. Were we afraid to use a little functional programming? Ripping all that out fixed the bug, made future code less fragile, and removed about 300LOC.

src/apphosting/yaml.ts
```diff
@@ -48,43 +48,26 @@ export class AppHostingYamlConfig {
...
-    for (const [key, value] of other._environmentVariables) {
-      this._environmentVariables.set(key, value);
+  merge(other: AppHostingYamlConfig, allowSecretsToBecomePlaintext: boolean = true) {
+    if (!allowSecretsToBecomePlaintext) {
+      const wereSecrets = Object.entries(this.env)
+        .filter(([, env]) => env.secret)
+        .map(([key]) => key);
+      if (wereSecrets.some((key) => other.env[key]?.value)) {
+        throw new FirebaseError(
+          `Cannot convert secret to plaintext in ${other.filename ? other.filename : "apphosting yaml"}`,
```
taeold, line 62:
```suggestion
          `Cannot convert secret to plaintext in ${other.filename ? other.filename : "apphosting.yaml"}`,
```

taeold: alternatively, `other.filename ?? "apphosting.yaml"`

src/apphosting/yaml.ts
```diff
@@ -3,38 +3,38 @@ import { readFileFromDirectory, wrappedSafeLoad } from "../utils";
...
   /**
    * Reads in the App Hosting yaml file found in filePath, parses the secrets and
    * environment variables, and returns an object that makes it easier to
    * programatically read or manipulate the App Hosting config.
    */
   static async loadFromFile(filePath: string): Promise<AppHostingYamlConfig> {
-    const config = new AppHostingYamlConfig();
     if (!fileExistsSync(filePath)) {
-      throw new FirebaseError("Cannot load AppHostingYamlConfig from given path, it doesn't exist");
+      throw new FirebaseError("Cannot load App Hosting yaml from given path, it doesn't exist");
```
taeold, line 28: nit* include `filePath` in the error

src/emulator/apphosting/serve.ts
```diff
@@ -36,12 +40,53 @@ export async function start(options?: StartOptions): Promise<{ hostname: string;
...
+      throw new FirebaseError(
+        `Cannot load secret ${match[1]} without a project. ` +
+          `Please use ${clc.bold("firebase use")} or pass the --project flag.`,
+      );
+    }
+    projectId = project;
+    secretId = match[1];
+    version = match[2] || "latest";
+  }
+  return await secrets.accessSecretVersion(projectId, secretId, version);
```
taeold, line 85: can we try/catch this to print user friendly log when status == 401 with suggestion to use `firebase apphosting:secrets:grantaccess` command?

src/apphosting/yaml.spec.ts
```diff
@@ -1,195 +1,52 @@
...
-      apphostingYaml.addEnvironmentVariable({
-        variable: "TEST",
-        value: "overwritten_value",
-      });
-
-      expect(JSON.stringify(apphostingYaml.environmentVariables)).to.equal(
-        JSON.stringify([{ variable: "TEST", value: "overwritten_value" }]),
-      );
+describe("merge", () => {
+  it("merges incoming apphosting yaml config with precendence", () => {
```
taeold, line 5:
```suggestion
  it("merges incoming apphosting.yaml config with precendence", () => {
```

taeold (approved):

src/apphosting/yaml.ts
inlined: This shoudl only ever happen in tests where we start with `.empty()` instead of loading from a file. I wanted something different so that we didn't accidentally start printing invalid file names if we didn't actually know what was right.

src/apphosting/yaml.spec.ts
inlined: Technically incorrect, no? It's about merging apphosting.local.yaml over apphosting.emulators.yaml over apphosting.yaml

Issue #4954, opened 2022-09-07
Check permission for`artifactregistry.packages.delete` before attempting to clean up container images


ylemkimon:
### [REQUIRED] Environment info

**firebase-tools:** 11.8.0

**Platform:** Ubuntu

### [REQUIRED] Test case

See below.

### [REQUIRED] Steps to reproduce

1. Create a Firebase Functions project and deploy.

```sh
firebase init functions
firebase deploy
```

2. Create a service account with `Cloud Functions Developer`, `Firebase Hosting Admin
` permission and `Service Account User` of the App Engine default service account in https://console.cloud.google.com/iam-admin/serviceaccounts 
4. Create and download a key.
5. Deploy the project with the service account.

```sh
firebase logout
GOOGLE_APPLICATION_CREDENTIALS=[path to downloaded key] firebase deploy
```

### [REQUIRED] Expected behavior

Build artifacts are cleaned up:

- The request should be called with a (Cloud Build or Cloud Functions) service account/agent.
- If a permission error occurs, it should not fail silently.
- The `artifactregistry.packages.delete` permission requirement should be documented.

### [REQUIRED] Actual behavior

```
i  functions: cleaning up build files... 
[2022-09-07T04:05:43.056Z] >>> [apiv2][query] DELETE https://artifactregistry.googleapis.com/v1beta2/projects/:project/locations/us-central1/repositories/gcf-artifacts/packages/hello_world [none]
[2022-09-07T04:05:43.057Z] >>> [apiv2][query] GET https://us.gcr.io/v2/:project/gcf/us-central1/tags/list [none]
[2022-09-07T04:05:43.361Z] <<< [apiv2][status] GET https://us.gcr.io/v2/:project/gcf/us-central1/tags/list 200
[2022-09-07T04:05:43.361Z] <<< [apiv2][body] GET https://us.gcr.io/v2/:project/gcf/us-central1/tags/list {"child":[],"manifest":{},"name":":project/gcf/us-central1","tags":[]}
[2022-09-07T04:05:43.362Z] Could not find image for function projects/:project/locations/us-central1/functions/helloWorld
[2022-09-07T04:05:44.238Z] <<< [apiv2][status] DELETE https://artifactregistry.googleapis.com/v1beta2/projects/:project/locations/us-central1/repositories/gcf-artifacts/packages/hello_world 403
[...]

Issue #8164, opened 2025-02-04
webframeworks deploy fails (404) by attempting delete of artifact


fivecar:
### Environment info

**firebase-tools:**
firebase-tools@13.29.3

**Platform:**
Ubuntu (GitHub Action)

### Test case

Using an Astro project with webframeworks deployment, a GitHub Action Yaml like this causes failure:
```yaml
      - name: Deploy site to Firebase Hosting
        uses: FirebaseExtended/action-hosting-deploy@v0
        with:
          repoToken: ${{ secrets.GITHUB_TOKEN }}
          firebaseServiceAccount: ${{ secrets.FIREBASE_SERVICE_ACCOUNT_WAVEZ_PROD }}
          channelId: live
          projectId: wavez-prod
          target: site
        env:
          FIREBASE_CLI_EXPERIMENTS: webframeworks
```

### Steps to reproduce

Run the GitHub Action

### Expected behavior

Successful deploy with no errors

### Actual behavior

The actual SSR function is deployed successfully (i.e. runs just fine), but the deployment script follows up with a "failure" during cleanup:
```console
[2025-02-04T06:55:15.275Z] >>> [apiv2][query] GET https://cloudfunctions.googleapis.com/v2/projects/wavez-prod/locations/us-central1/operations/operation-1738651964031-62d4b74b441a5-8f575101-8491b3b2 [none]
  [2025-02-04T06:55:15.627Z] <<< [apiv2][status] GET https://cloudfunctions.googleapis.com/v2/projects/wavez-prod/locations/us-central1/operations/operation-1738651964031-62d4b74b441a5-8f575101-8491b3b2 200
[...]

Issue #4757, opened 2022-07-20
"Unhandled error cleaning up build images." - what rights should I give to the GCP project?


akauppi:
### [REQUIRED] Environment info

- firebase-tools: 11.3.0
- Platform: Mac (does not matter)

### [REQUIRED] Test case

skip

### [REQUIRED] Steps to reproduce

skip

### [REQUIRED] Expected behavior

Firebase would clean up Cloud Function deployment cache, as it aims to do.

### [REQUIRED] Actual behavior

The cleanup fails. 

### Context

I am deploying a Firebase backend, using CI (Cloud Build).

Here's an excerpt from the log:

```
Step #6: === Deploying to 'groundlevel-jun-22'...
Step #6: 
Step #6: i  deploying firestore, functions
...
Step #6: i  functions: ensuring required API cloudfunctions.googleapis.com is enabled...
Step #6: i  functions: ensuring required API cloudbuild.googleapis.com is enabled...
Step #6: i  artifactregistry: ensuring required API artifactregistry.googleapis.com is enabled...
Step #6: ✔  artifactregistry: required API artifactregistry.googleapis.com is enabled
Step #6: ✔  functions: required API cloudbuild.googleapis.com is enabled
Step #6: ✔  functions: required API cloudfunctions.googleapis.com is enabled
Step #6: i  functions: preparing codebase default for deployment
Step #6: i  functions: preparing ./functions directory for uploading...
Step #6: i  functions: packaged /workspace/packages/backend/functions (53.92 KB) for uploading
Step #6: i  firestore: latest version of ./firestore.rules already up to date, skipping upload...
Step #6: ✔  firestore: deployed indexes in ./firestore.indexes.json successfully
Step #6: ✔  functions: ./functions folder uploaded successfully
Step #6: ✔  firestore: released rules ./firestore.rules to cloud.firestore
Step #6: i  functions: updating Node.js 16 function userInfoShadow_2(europe-west2)...
Step #6: ⚠  functions: failed to update function projects/groundlevel-jun-22/locations/europe-west2/functions/userInfoShadow_2
Step #6: Failed to update function projects/groundlevel-jun-22/locations/europe-west2/functions/userInfoShadow_2
Step #6: 
Step #6: Functions deploy had errors with the following functions:
Step #6: 	userInfoShadow_2(europe-west2)
Step #6: i  functions: cleaning up build files...
[...]

PR #8324, opened 2025-03-16
[BREAKING(functions)] Update post-deploy function logic to stop aggressively deleting conatiners

taeold:
Today, CLI attempts to clean up GCF artifacts (container images) in Artifact Registry post function deploys.

Historically, this post-deploy cleanup task had not been too reliable, and the resulting warning logs have confused developers.

We are trying a different approach - instead of trying to clean up the artifacts on every deploy, we'll make it easy for developers to set up a [Artifact Registry Cleanup Policy](https://cloud.google.com/artifact-registry/docs/repositories/cleanup-policy) on repository used by Cloud Run functions.

Cleanup policies are a managed service feature that automatically removes container images based on configurable rules, providing a more robust and transparent solution.

As suggested in this change, we'll automatically set up a clean up policy post-function deployment if we detect that a clean up policy doesn't exist yet.

We'll soon be updating the documentations to clarify what developers should do in order to avoid small monthly cost for Firebase Function's use of Artifact Registry.

Fixes https://github.com/firebase/firebase-tools/issues/4954, https://github.com/firebase/firebase-tools/issues/4954, https://github.com/firebase/firebase-tools/issues/8164, https://github.com/firebase/firebase-tools/issues/4757

### DIFF
--- a/src/deploy/functions/release/index.ts
+++ b/src/deploy/functions/release/index.ts
@@ -3,20 +3,20 @@ import * as clc from "colorette";
 import { Options } from "../../../options";
 import { logger } from "../../../logger";
 import { reduceFlat } from "../../../functional";
+import * as utils from "../../../utils";
 import * as args from "../args";
 import * as backend from "../backend";
-import * as containerCleaner from "../containerCleaner";
 import * as planner from "./planner";
 import * as fabricator from "./fabricator";
 import * as reporter from "./reporter";
 import * as executor from "./executor";
 import * as prompts from "../prompts";
-import * as experiments from "../../../experiments";
 import { getAppEngineLocation } from "../../../functionsConfig";
 import { getFunctionLabel } from "../functionsDeployHelper";
 import { FirebaseError } from "../../../error";
 import { getProjectNumber } from "../../../getProjectNumber";
 import { release as extRelease } from "../../extensions";
+import * as artifacts from "../../../functions/artifacts";
 
 /** Releases new versions of functions and extensions to prod. */
 export async function release(
@@ -104,13 +104,11 @@ export async function release(
   const wantBackend = backend.merge(...Object.values(payload.functions).map((p) => p.wantBackend));
   printTriggerUrls(wantBackend);
 
-  const haveEndpoints = backend.allEndpoints(wantBackend);
-  const deletedEndpoints = Object.values(plan)
-    .map((r) => r.endpointsToDelete)
-    .reduce(reduceFlat, []);
-  if (experiments.isEnabled("automaticallydeletegcfartifacts")) {
-    await containerCleaner.cleanupBuildImages(haveEndpoints, deletedEndpoints);
-  }
+  await setupArtifactCleanupPolicies(
+    options,
+    options.projectId!,
+    Object.keys(wantBackend.endpoints),
+  );
 
   const allErrors = summary.results.filter((r) => r.error).map((r) => r.error) as Error[];
   if (allErrors.length) {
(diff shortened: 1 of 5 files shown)


chalosalvador:
@taeold I opened the PR #8318 which may be related to this PR. It updates the `packagePath` method to handle the correct names for the artifacts in GCF v1 and v2 (#8164)

taeold:
@chalosalvador that's a good change, but I think I'm going to suggest that Functions still move on from trying to maintain the container cleanup logic. The logic is still used in `functions:deletegcfartifacts` command which users can use to manually purge their Artifact Repository repo if needed, so the PR is definitely helpful.

src/deploy/functions/release/index.ts
```diff
@@ -105,12 +106,8 @@ export async function release(
...
 
   const haveEndpoints = backend.allEndpoints(wantBackend);
-  const deletedEndpoints = Object.values(plan)
-    .map((r) => r.endpointsToDelete)
-    .reduce(reduceFlat, []);
-  if (experiments.isEnabled("automaticallydeletegcfartifacts")) {
-    await containerCleaner.cleanupBuildImages(haveEndpoints, deletedEndpoints);
-  }
+
+  await checkArtifactCleanupPolicies(options.projectId!, haveEndpoints);
```
inlined, line 110: Won't this leak if a customer had a lone function in a region and they cleaned it up before we could check policies?

src/deploy/functions/release/index.ts
```diff
@@ -144,3 +141,119 @@ export function printTriggerUrls(results: backend.Backend): void {
...
+ * and automatically sets up policies where needed.
+ *
+ * The policy is only set up when:
+ * 1. No cleanup policy exists yet
+ * 2. No other cleanup policies exist (beyond our own if we previously set one)
+ * 3. User has not explicitly opted out
+ */
+async function checkArtifactCleanupPolicies(
+  projectId: string,
+  endpoints: backend.Endpoint[],
```
inlined, line 156: 1. This seems to unnecessarily couple the function to backend.Endpoint when there would be less coupling to just check one or more regions.
2. backend.Backend.endpoints is a map keyed on region. Why not just use `Object.keys(wantBackend.endpoints)` to get a list of regions?

src/deploy/functions/release/index.ts
```diff
@@ -144,3 +141,119 @@ export function printTriggerUrls(results: backend.Backend): void {
...
+          hasPolicy: false,
+          hasOptOut: false,
+          hasOtherPolicies: false,
+          error: err,
+        };
+      }
+    }),
+  );
+
+  const regionsWithErrors = regionResults
```
inlined, line 203: It might be more idiomatic to use `Promise.allSettled()` and then filter out the resolved vs rejected promises. Here, `Promise.all()` is going to fail on the first error and you'll get effectively undefined behavior.

src/deploy/functions/release/index.ts
```diff
@@ -144,3 +141,119 @@ export function printTriggerUrls(results: backend.Backend): void {
...
+            `Failed to set up artifact cleanup policy for region ${result.region}:`,
+            err,
+          );
+          regionsNeedingWarning.push(result.region);
+          return { region: result.region, success: false, error: err };
+        }
+      }),
+    );
+
+    const failedSetups = setupResults.filter((r) => !r.success);
```
inlined, line 236: `Promise.allSettled` again.

src/deploy/functions/release/index.ts
```diff
@@ -144,3 +141,119 @@ export function printTriggerUrls(results: backend.Backend): void {
...
+    (result) => !result.hasPolicy && !result.hasOptOut && !result.hasOtherPolicies && !result.error,
+  );
+
+  const regionsNeedingWarning: string[] = [];
+
+  if (regionsToSetup.length > 0) {
+    utils.logLabeledSuccess(
+      "functions",
+      `Configuring a cleanup policy for repositories in ${regionsToSetup.join(", ")}. ` +
+        `Images older than ${artifacts.DEFAULT_CLEANUP_DAYS} days will be automatically deleted.`,
```
inlined, line 217: It gives me the heebies to depend on a const like this inside the function. Can we at least make it a parameter and default it to the const?

Also, why use the default instead of prompting with this as the default value of the prompt?

taeold: Interesting. I've never even considered introducing a prompt at the end of the deployment, thinking that it's an abrupt change in behavior of how functions deployment has worked for a long time.

Happy to make it a prompt and refactor things a bit to share more code between this and the command if you like the idea of prompting at the end.

src/deploy/functions/release/index.ts
```diff
@@ -144,3 +142,65 @@ export function printTriggerUrls(results: backend.Backend): void {
...
+ * and automatically sets up policies where needed.
+ *
+ * The policy is only set up when:
+ *   1. No cleanup policy exists yet
+ *   2. No other cleanup policies exist (beyond our own if we previously set one)
+ *   3. User has not explicitly opted out
+ *
+ * In non-interactive mode:
+ *   - With force flag: applies the default cleanup policy
+ *   - Without force flag: warns and does not apply policy
```
inlined, line 157: Aborts, right?

src/deploy/functions/release/index.ts
```diff
@@ -144,3 +142,65 @@ export function printTriggerUrls(results: backend.Backend): void {
...
+    utils.logLabeledWarning(
+      "functions",
+      `Failed to set up cleanup policy for repositories in ${regionsWithErrors.length > 1 ? "regions" : "region"} ` +
+        `${regionsWithErrors.join(", ")}.` +
+        "This could result in a small monthly bill as container images accumulate over time.",
+    );
+    throw new FirebaseError(
+      `Functions successfully deployed but could not set up cleanup policy in ` +
+        `${regionsWithErrors.length > 1 ? "regions" : "region"} ${regionsWithErrors.join(", ")}.` +
+        `Pass the--force option to automatically set up a cleanup policy or` +
```
inlined, line 202: "the --force"

src/functions/artifacts.ts
```diff
@@ -37,6 +38,27 @@ export function makeRepoPath(
   return `projects/${projectId}/locations/${location}/repositories/${repoName}`;
 }
 
+/** For testing purposes */
+
+export const getRepoCache = new Map<string, artifactregistry.Repository>();
```
inlined, line 43: Two things:
1. Why Map? We always use Record, not Map
2. I specifically hate using global cache objects because they tend to cause subtle test errors (and/or cause prod to use cache when it shouldn't). That's why the backend code puts its cache in `context`. I know we're really pressed for time here; can we at least add a TODO to resolve a discussion about this?

src/functions/artifacts.ts
```diff
@@ -182,3 +204,100 @@ export function hasSameCleanupPolicy(
...
+      } catch (err) {
+        logger.debug(`Failed to set up artifact cleanup policy for region ${location}:`, err);
+        return { region: location, success: false, error: err };
+      }
+    }),
+  );
+
+  for (const result of setupRepos) {
+    if (result.status === "rejected") {
+      // should not happen
```
inlined, line 288: Nit: I'm personally a fan of throwing if it should never happen (can use status 1 to mean "Firebase screwed up" in telemetry)

inlined: Alternatively, just use the promise settled result instead of 'success" and "err"

inlined (approved):
I like the compromise to the internal discussion that the deployment will succeed but we'll exit with non-zero if the there is no policy, no opt-out, is non-interactive, and force isn't set.

PR #8328, opened 2025-03-17
Swap pkg for yao/pkg

joehan:
### Description
pkg is no longer maintained, and does not support anything after node 18. We will be dropping support for node 18 in the next major version, so we're switching to a forked version that does support newer node versions.

Long term, we'll probably move to SEAs, but not yet.

Thanks a ton @aalej for working on this and testing it out!


yuchenshi (approved):
Sure if it works it works. I don't know what to review.

PR #8330, opened 2025-03-18
Fix bug in the set policy command where --none flag no longer works.

taeold:
Accidentally introduced a bug where `--none` flag couldn't be specified.

Also refactor the default policy day to be a global constant (will use later) and set it to 1 instead of 3.

### DIFF
--- a/src/commands/functions-artifacts-setpolicy.ts
+++ b/src/commands/functions-artifacts-setpolicy.ts
@@ -26,18 +26,12 @@ export const command = new Command("functions:artifacts:setpolicy")
   )
   .option(
     "--days <days>",
-    "Number of days to keep container images before deletion. Default is 3 days.",
-    "3",
+    `Number of days to keep container images before deletion. Default is ${artifacts.DEFAULT_CLEANUP_DAYS} day.`,
   )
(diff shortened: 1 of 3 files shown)


src/commands/functions-artifacts-setpolicy.ts
```diff
@@ -26,18 +26,13 @@ export const command = new Command("functions:artifacts:setpolicy")
   )
   .option(
     "--days <days>",
-    "Number of days to keep container images before deletion. Default is 3 days.",
-    "3",
+    `Number of days to keep container images before deletion. Default is ${artifacts.DEFAULT_CLEANUP_DAYS} day.`,
+    `${artifacts.DEFAULT_CLEANUP_DAYS}`,
```
inlined, line 30: I missed this in another PR, but can we actually tell the difference between getting a default value and no value here? In other words is there a way to know the difference between `--days 1 --none` and `--none`?

taeold: Fixed it so that we can differentiate the 2 states.

inlined (approved):

PR #8333, opened 2025-03-18
Add support for apphosting.emulator.yaml

inlined:
The App Hosting emulator can now load values from apphosting.emulator.yaml. This file is designed to be safe to commit, and therefore cannot convert a secret value to a plaintext value.

### DIFF
--- a/src/emulator/apphosting/config.ts
+++ b/src/emulator/apphosting/config.ts
@@ -8,8 +8,11 @@ import {
 import { AppHostingYamlConfig } from "../../apphosting/yaml";
 
 /**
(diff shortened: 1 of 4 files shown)


src/emulator/apphosting/config.ts
```diff
@@ -15,27 +16,33 @@ export async function getLocalAppHostingConfiguration(
 ): Promise<AppHostingYamlConfig> {
```
mathu97, line 16: We should update the JSDoc to reflect the new behaviour of us loading in the apphosting.emulator.yaml file as well.

mathu97 (approved):

PR #8337, opened 2025-03-19
bump min vscode version to 0.19.0 

hlshen:
CLI is removing support for node < 20. Bumping up minimum VSCode version to match.


joehan:
Moved this over to next, but lgtm otherwise

joehan (approved):

PR #8339, opened 2025-03-19
Improve fdc permissions setup

tammam-g:
This PR improves the FDC schema setup:
1- Make sure IAM user is upserted before sql:diff
2- Refactor brownfield setups to revoke owners after setup to avoid getting left with IAM roles.
3- Use transactions for schema setup commands to avoid problems when setup fails in intermediate state (e.g if we assign IAM role to firebasesuperuser then firebasesuperuser becomes IAM role and it becomes not possible to login using that role).
4- If schema is setup as brownfield then schema migration will fail without prompting the user to setup. If users decide to change their setup then they should run the setup command explicitly (this change doesn't affect fresh database, they are still automatically setup). 
5- Improve messaging (tell users to turn on compate mode)

Scenarios tested:
- Completely fresh database is automatically setup as greenfield without any prompting
- Brownfield database with IAM tables setup as greenfield 
- Ctrl+c half way through brownfield setup and confirming IAM role wasn't assigned to firebasesuperuser (transaction wasn't commited)

### DIFF
--- a/src/dataconnect/schemaMigration.ts
+++ b/src/dataconnect/schemaMigration.ts
@@ -30,15 +30,31 @@ import * as cloudSqlAdminClient from "../gcp/cloudsql/cloudsqladmin";
 
 import * as errors from "./errors";
 
-async function setupSchemaIfNecessary(instanceId: string, databaseId: string, options: Options) {
+async function setupSchemaIfNecessary(
+  instanceId: string,
+  databaseId: string,
+  options: Options,
+): Promise<SchemaSetupStatus.GreenField | SchemaSetupStatus.BrownField> {
   await setupIAMUsers(instanceId, databaseId, options);
   const schemaInfo = await getSchemaMetadata(instanceId, databaseId, DEFAULT_SCHEMA, options);
   if (
     schemaInfo.setupStatus !== SchemaSetupStatus.BrownField &&
     schemaInfo.setupStatus !== SchemaSetupStatus.GreenField
   ) {
-    await setupSQLPermissions(instanceId, databaseId, schemaInfo, options, /* silent=*/ true);
+    return await setupSQLPermissions(
+      instanceId,
+      databaseId,
+      schemaInfo,
+      options,
+      /* silent=*/ true,
+    );
+  } else {
+    logger.info(
+      `Detected schema "${schemaInfo.name}" is setup in ${schemaInfo.setupStatus} mode. Skipping Setup.`,
+    );
   }
+
+  return schemaInfo.setupStatus;
 }
 
 export async function diffSchema(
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,3 +6,4 @@
 - Increase emulator UI body parser limit to match Storage emulator maximum. (#8329)
 - Fixed Data Connect setup issues for fresh databases due to IAM user not being created. (#8335)
 - Fixed an issue where `ext:install` used POSIX file seperators on Windows machines. (#8326)
+- Allow FDC permissions to better handle tables owned by IAM roles. (#8339)
```
joehan, line 9:
```suggestion
- Improved Data Connect SQL permissions to better handle tables owned by IAM roles. (#8339)
```
FDC isn't an acronym that external users would know

src/dataconnect/schemaMigration.ts
```diff
@@ -43,6 +44,15 @@ export async function diffSchema(
   );
   let diffs: Diff[] = [];
 
+  await setupIAMUsers(instanceId, databaseId, options);
```
joehan, line 47: Nit: This chunk is (nearly) exactly repeated between the two functions, consider DRYing out with a helper function

src/dataconnect/schemaMigration.ts
```diff
@@ -383,19 +403,11 @@ async function handleIncompatibleSchemaError(args: {
...
     const schemaInfo = await getSchemaMetadata(instanceId, databaseId, DEFAULT_SCHEMA, options);
     if (schemaInfo.setupStatus !== SchemaSetupStatus.GreenField) {
-      const newSetupStatus = await setupSQLPermissions(
-        instanceId,
-        databaseId,
-        schemaInfo,
-        options,
-        /* silent=*/ true,
+      throw new FirebaseError(
+        `Brownfield database are protected from dataconnect SQL changes.\n` +
```
joehan, line 407:
```suggestion
        `Brownfield database are protected from SQL changes by Data Connect.\n` +
```

src/dataconnect/schemaMigration.ts
```diff
@@ -383,19 +403,11 @@ async function handleIncompatibleSchemaError(args: {
...
+          `You can use the SQL diff generated by 'firebase dataconnect:sql:diff' to assist you in applying the required changes to your CloudSQL database. Connector deployment will succeed when there is no required diff changes.` +
+          `If you would like Dataconnect to control your database you can run 'firebase dataconnect:sql:setup'`,
```
joehan, line 409:
```suggestion
          `If you would like Data Connect to manage your database schema, run 'firebase dataconnect:sql:setup'`,
```

src/gcp/cloudsql/connect.ts
```diff
@@ -11,6 +11,7 @@ import { logger } from "../../logger";
 import { FirebaseError } from "../../error";
 import { Options } from "../../options";
 import { FBToolsAuthClient } from "./fbToolsAuthClient";
+import { log } from "console";
```
joehan, line 14: Looks like this accidentally slipped in, remove.

src/gcp/cloudsql/connect.ts
```diff
@@ -90,21 +92,32 @@ export async function execute(
...
+  if (opts.transaction) {
+    sqlStatements.unshift("BEGIN;");
+    sqlStatements.push("COMMIT;");
+  }
   for (const s of sqlStatements) {
     logFn(`Executing: '${s}'`);
     try {
       results.push(await conn.query(s));
     } catch (err) {
+      logFn("Rolling back transaction");
```
joehan, line 113: Let's log the error too (at least at debug level)

src/gcp/cloudsql/permissions_setup.ts
```diff
@@ -114,20 +117,36 @@ export async function setupSQLPermissions(
   }
   await setupIAMUsers(instanceId, databaseId, options);
 
+  let runGreenfieldSetup = false;
   if (schemaInfo.setupStatus === SchemaSetupStatus.GreenField) {
+    runGreenfieldSetup = true;
     logger.info(
-      `Database ${databaseId} has already been setup. Rerunning setup to repair any missing permissions.`,
+      `Database ${databaseId} has already been setup as Greenfield project. Rerunning setup to repair any missing permissions.`,
```
joehan, line 124:
```suggestion
      `Database ${databaseId} has already been setup as greenfield project. Rerunning setup to repair any missing permissions.`,
```

src/gcp/cloudsql/permissions_setup.ts
```diff
@@ -114,20 +117,36 @@ export async function setupSQLPermissions(
...
     );
-    await greenFieldSchemaSetup(instanceId, databaseId, schema, options, silent);
-    return SchemaSetupStatus.GreenField;
-  } else {
-    logger.info(`Detected schema "${schema}" setup status is ${schemaInfo.setupStatus}.`);
   }
 
-  // We need to setup the database
   if (schemaInfo.tables.length === 0) {
+    runGreenfieldSetup = true;
```
joehan, line 129: Nit: We have inconsistent capitalization on greenfield and brownfield (some places it's greenField, some it's greenfield). Wikipedia says 'greenfield' (https://en.wikipedia.org/wiki/Greenfield_project), so let's settle on that

src/gcp/cloudsql/permissions_setup.ts
```diff
@@ -153,6 +172,12 @@ export async function setupSQLPermissions(
 
   if (shouldSetupGreenfield) {
     await setupBrownfieldAsGreenfield(instanceId, databaseId, schemaInfo, options, silent);
+    logger.info(clc.green("Database setup complete."));
+    logger.info(
+      clc.yellow(
+        "IMPORTANT: please uncomment 'schemaValidation: \"COMPATIBLE\"' in your dataconnect.yaml file to avoid dropping any existing tables by mistake.",
```
joehan, line 178: (can be in a follow up) Should we just uncomment it by default in this case? Seems like a foot gun otherwise

joehan (approved):
Some ntis and suggestions, but mostly LGTM

joehan:
Could you also add some Scenarios Tested in the PR description?

src/gcp/cloudsql/permissions_setup.ts
tammam-g: Are you aware of an easy way to do the uncommenting?

PR #8342, opened 2025-03-20
Provide firebase login credentials to Data connect emulator

joehan:
### Description
Provide default credentials to Data Connect emulator so that it can make calls to the vertex and data connect APIs.


hlshen (approved):
Approved; Do we need to add any safety checks for account not existing?

src/commands/dataconnect-sdk-generate.ts
```diff
@@ -42,10 +43,12 @@ export const command = new Command("dataconnect:sdk:generate")
         return;
       }
       for (const conn of serviceInfo.connectorInfo) {
+        const account = getProjectDefaultAccount(options.projectRoot);
```
rosalyntan, line 46: I think it's just `fdc build` and `fdc dev` that need the credentials (not `fdc generate`)

joehan: Yeah, I figured we ought to provide them  as future proofing.

rosalyntan (approved):

PR #8354, opened 2025-03-21
Set extraEnv in vscode process

joehan:
### Description
Fixes an issue where extraEnv wasn't actually set for code executed as part of the extension. 

Also, added some logging to help debug where prod requests are going.

### Scenarios Tested
Verified that when I set the DATA_CONNECT_URL env var to staging, Run (production) sends requests to staging.


src/dataconnect/dataplaneClient.ts
```diff
@@ -1,4 +1,5 @@
 import { dataconnectOrigin } from "../api";
+export { dataconnectOrigin } from "../api"
```
rosalyntan, line 2: nit: `"../api";`

rosalyntan (approved):
Thanks!!

PR #8356, opened 2025-03-21
Increase size limit for JSON uploads to storage

### LINKED ISSUE #8355: PayloadTooLargeError when uploading 130kb json with Content-Type application/json

**firebase-tools:**
13.35.1

**Platform:**
macOS 13.7.4

### [REQUIRED] Test case

```
# Generate a 130kb json file
printf '{"k":"%s"}' "$(head -c 130000 < /dev/zero | tr '\0' 'a')" > large.json
# Run the emulator without the Content-Type header, this works
firebase emulators:exec --log-verbosity DEBUG --only storage -- 'curl "localhost:9199/upload/storage/v1/b/bucket/o?name=test&uploadType=media" -d "@large.json" -o /dev/null -w "%{http_code}"'
# ...
200✔  Script exited successfully (code 0)
# Run the same request with Content-Type header set to application/json
firebase emulators:exec --log-verbosity DEBUG --only storage -- 'curl "localhost:9199/upload/storage/v1/b/bucket/o?name=test&uploadType=media" -d "@large.json" -H "Content-Type: application/json" -o /dev/null -w "%{http_code}"'
# ...
413PayloadTooLargeError: request entity too large

Run the test case above.

A json object uploaded with the content-type set should be successful, with a 200 status response and the object created.

This should work because in production Cloud Storage, uploading with the content-type header set is the preferred way to ensure that HTTP responses from the bucket use the same content-type header. This is important for serving and metadata tracking.

Uploading with "content-type: application/json" fails because the request size limit is `102400` bytes, which is the default.

The default is set here in server.ts:
https://github.com/firebase/firebase-tools/blob/7a7f3a5cd1edf159b2ecdcc2d6f075b96ad0a2d6/src/emulator/storage/server.ts#L64-L68

This should either be increased, or if the upload endpoint is hit with the content type header, the emulator should respond with a more informative error message like "Json content types not supported for emulator uploads".

joehan:
### Description
The Storage emulator was applying an artificially low limit on upload size when Content-Type: application/json was set. Fixes #8355

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,2 @@
 - Fix bug in Auth emulator where accounts:lookup is case-sensitive for emails (#8344)
+- Fixed an issue where the Storage emulator applied too low of a limit for files uploaded with Content-Type: application/json. (#8355)
--- a/src/emulator/storage/server.ts
+++ b/src/emulator/storage/server.ts
@@ -64,7 +64,7 @@ export function createApp(
   app.use(
     express.json({
       type: ["application/json"],
-      limit: "130mb"
+      limit: "130mb",
     }),
   );
 
--- a/src/emulator/storage/server.ts
+++ b/src/emulator/storage/server.ts
@@ -64,6 +64,7 @@ export function createApp(
   app.use(
     express.json({
       type: ["application/json"],
+      limit: "130mb"
     }),
   );
 


joehan:
> irebase emulators:exec --log-verbosity DEBUG --only storage -- 'curl "localhost:9199/upload/storage/v1/b/bucket/o?name=test&uploadType=media" -d "@large.json" -H "Content-Type: application/json" -o /dev/null -w "%{http_code}"

Good catch, i just repro'd this too. Investigating before merging - also need to double check that this is actually new behavior

mgraczyk:
@joehan FWIW there is a longstanding, separate bug where the emulator hangs forever when I upload certain files and set the Content-Type header.

This has been an issue for at least a year but I just rerun my tests automatically when this happens, and it never happens with prod GCS so I ignored it.

mgraczyk:
@joehan I just looked, the problem is that the handler for [the upload POST endpoint](https://github.com/firebase/firebase-tools/blob/6c7245f6a6dbffdb2bc1d794234a3f0b1704c7b4/src/emulator/storage/apis/gcloud.ts#L257) tries to [read the body here](https://github.com/firebase/firebase-tools/blob/6c7245f6a6dbffdb2bc1d794234a3f0b1704c7b4/src/emulator/storage/apis/gcloud.ts#L346C22-L346C37), but it has already been read by express.json [here](https://github.com/firebase/firebase-tools/blob/6c7245f6a6dbffdb2bc1d794234a3f0b1704c7b4/src/emulator/storage/server.ts#L65)

One fix would be to check for json in  here:


As written, it will hang forever in `reqBodyToBuffer` whenever the body has already been read:

https://github.com/firebase/firebase-tools/blob/6c7245f6a6dbffdb2bc1d794234a3f0b1704c7b4/src/emulator/shared/request.ts#L7

Since express.json writes an empty object `{}` for non-json Conten-Type headers, you can't just check for `typeof request.body === "object"`, and may need to check the headers before the call to `reqBodyToBuffer` or something like that.

andersonaddo:
@joehan what's the status on this? This is still a problem, as far as I can see

joehan:
Hey @andersonaddo - this PR had been abandoned because it turned out the issue was a bit more complex than I originally thought. Closing it in favor of #10800, which I beleive actually fixes the issue.

PR #8357, opened 2025-03-24
Can grant access to emails (user, group, SA)

inlined:
Part of the new work to encourage the use of test secrets in the emulator.

Unfortunately there is no general API to figure out whether an email address is a service account, user, or group. Fortunately, IAM will return an error with a clear message on what type of principal the secret was on error so we can brute force our way to success ¯\_(ツ)_/¯

### DIFF
--- a/src/apphosting/secrets/index.ts
+++ b/src/apphosting/secrets/index.ts
@@ -98,9 +98,8 @@ export async function grantSecretAccess(
     );
   }
 
+  const updatedBindings = existingBindings.concat(newBindings);
   try {
-    // TODO: Merge with existing bindings with the same role
-    const updatedBindings = existingBindings.concat(newBindings);
     await gcsm.setIamPolicy({ projectId, name: secretName }, updatedBindings);
   } catch (err: unknown) {
     throw new FirebaseError(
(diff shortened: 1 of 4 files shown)


src/apphosting/secrets/index.ts
```diff
@@ -112,6 +111,69 @@ export async function grantSecretAccess(
...
+  // already, why not just accept the value without leaking its type?
+  // Note: we keep typeGuesses outside of the loop so that we learn the type of principal an email is once across all secrets.
+  const typeGuesses = Object.fromEntries(emails.map((email) => [email, "user"]));
+  for (const secretName of secretNames) {
+    let existingBindings;
+    try {
+      existingBindings = (await gcsm.getIamPolicy({ projectId, name: secretName })).bindings || [];
+    } catch (err: unknown) {
+      throw new FirebaseError(
+        `Failed to get IAM bindings on secret: ${secretName}. Ensure you have the permissions to do so and try again.`,
```
taeold, line 136: What if we link to https://cloud.google.com/secret-manager/docs/manage-access-to-secrets#required-roles to guide the users a little more here?

I'd love to link to a Firebase doc, but seems like we don't have a doc.

src/commands/apphosting-secrets-grantaccess.ts
```diff
@@ -9,10 +9,13 @@ import * as apphosting from "../gcp/apphosting";
...
 
-export const command = new Command("apphosting:secrets:grantaccess <secretName>")
-  .description("grant service accounts permissions to the provided secret")
+export const command = new Command("apphosting:secrets:grantaccess <secretNames>")
+  .description(
+    "grant service accounts permissions to the provided secret(s). Can pass one or more secrets, separated by a comma",
+  )
   .option("-l, --location <location>", "backend location", "-")
   .option("-b, --backend <backend>", "backend name")
+  .option("-u, --emails <emails>", "comma delmited list user or group emails")
```
taeold, line 18: Want to update the short flag too or keep it?

```suggestion
  .option("-e, --emails <emails>", "comma delmited list user or group emails")
```

src/commands/apphosting-secrets-grantaccess.ts
```diff
@@ -54,5 +73,9 @@ export const command = new Command("apphosting:secrets:grantaccess <secretName>"
 
     const accounts = secrets.toMulti(secrets.serviceAccountsForBackend(projectNumber, backend));
 
-    await secrets.grantSecretAccess(projectId, projectNumber, secretName, accounts);
+    await Promise.all(
```
taeold, line 76: Prefer `Promise.allSettled`?

taeold (approved):

src/commands/apphosting-secrets-grantaccess.ts
```diff
@@ -73,7 +73,7 @@ export const command = new Command("apphosting:secrets:grantaccess <secretNames>
 
     const accounts = secrets.toMulti(secrets.serviceAccountsForBackend(projectNumber, backend));
 
-    await Promise.all(
+    await Promise.allSettled(
```
taeold, line 76: Wanna explicitly handle errors here?

Or we could just go back to using `Promise.all()` instead... 🤷🏼 

e.g.
```
const results = await Promise.allSettled(
  secretList.map((secretName) => 
    secrets.grantSecretAccess(projectId, projectNumber, secretName, accounts)
  )
);

const failures = results.filter(result => result.status === 'rejected');
if (failures.length > 0) {
  throw new Error(`Failed to grant access to ${failures.length}/${results.length} secrets`);
 }
```

taeold (approved):

PR #8361, opened 2025-03-25
[BREAKING(apphosting)] Deprecate apphosting.local.yaml in favor of apphosting.emulator.yaml

inlined:
1. Removes apphosting export command as we no longer recommend storing secrets in plaintext.
2. Offers to create apphosting.emulator.yaml if it does not exist
3. If creating apphosting.emulator.yaml, offers to overwrite any value in apphosting.yaml
4. When overwriting secrets, detects that a secret reference ID already exists and confirms whether the user actually wants to reuse the secret reference
5. For secrets in { ...apphosting.yaml[env], ...apphosting.emulator.yaml[env]} prompts for a set of emails to grant access
6. Prints detailed logs on how to grant access if projectId is missing

projectId may become necessary, depending on some changes @jamesdaniels is looking at, but this should all be forwards compatible with these changes.


src/apphosting/config.spec.ts
```diff
@@ -240,60 +242,226 @@ env:
...
+      fs.fileExistsSync.withArgs(`/project/${config.APPHOSTING_EMULATORS_YAML_FILE}`).returns(true);
+
+      await config.maybeGenerateEmulatorYaml("projectId", "/project");
+
+      expect(prompt.confirm).to.not.have.been.called;
+      expect(store).to.not.have.been.called;
+    });
+
+    // This allows us to prompt to give devs access to prod keys
+    it("Returns existing config even if the user does not create apphosting.emulator.yaml", async () => {
```
taeold, line 285:
```suggestion
    it("returns existing config even if the user does not create apphosting.emulator.yaml", async () => {
```

src/apphosting/config.spec.ts
```diff
@@ -240,60 +242,226 @@ env:
...
+      // Do not create emulator file
+      prompt.confirm.resolves(false);
+      loadFromFile.resolves(existingYaml);
+
+      await expect(
+        config.maybeGenerateEmulatorYaml("projectId", "/project"),
+      ).to.eventually.deep.equal(toEnvList(existingYaml.env));
+    });
+
+    it("Returns overwritten config", async () => {
```
taeold, line 299:
```suggestion
    it("returns overwritten config", async () => {
```

taeold (approved):

annajowang (approved):

src/apphosting/config.ts
```diff
@@ -215,160 +212,129 @@ export async function maybeAddSecretToYaml(
...
-      "Which environment would you like to export secrets from Secret Manager for?",
-    );
-  }
+    let secretRef: string;
+    let action: "reuse" | "create" | "pick-new" = "pick-new";
+    while (action === "pick-new") {
+      secretRef = await prompt.promptOnce({
+        type: "input",
+        message: `What would you like to name the secret reference for ${name}?`,
+        // TODO: kebab case + test
```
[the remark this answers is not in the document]
inlined: Actually fixed it and forgot to remove the TODO 😆

PR #8363, opened 2025-03-25
Fix bug where we revoke cloudsqlsuperuser role from firebasesuperuser

tammam-g:
Last week we added a change to revoke user roles from firebasesuperuser once database setup is complete. This was an issue because if we assign IAM role to firebasesuperuser and we don't revoke it then cloudsql starts treating firebasesuperuser as IAM role and we can't access it anymore. 

However, an edge case we didn't consider is cloudsqlsuperuser owning tables/schem, this will basically remove cloudsqlsuperuser access from firebasesuperuser.

Scenarios tested:
- Fresh Database Setup
- Reproduce bug and confirm this fixes it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
 - Fix bug in Auth emulator where accounts:lookup is case-sensitive for emails (#8344)
 - firebase apphosting:secrets:grantAccess can now grant access to emails and can grant multiple secrets at once (#8357)
 - firebase apphosting:secrets:set now has flows to help with test secrets (#8359)
-- Fix bug where we revoke cloudsqlsuperuser from firebasesuperuser (#8363)
+- Fixed an issue where `sql:setup` would incorrectly remove the `cloudsqlsuperuser` role from `firebasesuperuser` (#8363)


CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
 - Fix bug in Auth emulator where accounts:lookup is case-sensitive for emails (#8344)
 - firebase apphosting:secrets:grantAccess can now grant access to emails and can grant multiple secrets at once (#8357)
 - firebase apphosting:secrets:set now has flows to help with test secrets (#8359)
+- Fix bug where we revoke cloudsqlsuperuser from firebasesuperuser (#8363)
```
joehan, line 4:
```suggestion
- Fixed an issue where `sql:setup` would incorrectly remove the `cloudsqlsuperuser` role from `firebasesuperuser` (#8363)
```
Did this only happen on setup? Or also on other deployments?

joehan (approved):

CHANGELOG.md
tammam-g: On setup, it happened to use on deploy because it was the first deploy

PR #8368, opened 2025-03-26
Prompted users before overwriting connector when running SDK Init

maneesht:
image


src/config.ts
```diff
@@ -229,9 +232,14 @@ export class Config {
     fs.removeSync(this.path(p));
   }
 
-  askWriteProjectFile(p: string, content: any, force?: boolean, confirmByDefault?: boolean) {
+  async askWriteProjectFile(
+    p: string,
```
joehan, line 236: Nit in not your code - p is a completely non descriptive variable name. Probably should rename to fileName

src/init/features/dataconnect/sdk.ts
```diff
@@ -224,10 +224,17 @@ export async function generateSdkYaml(
   return connectorYaml;
 }
 
-export async function actuate(sdkInfo: SDKInfo) {
+export async function actuate(sdkInfo: SDKInfo, config?: Config) {
```
joehan, line 227: Is there a reason that this is optional (and we have the fallback behavior)? Look like actuate is only ever called in one place, and config can't be undefined there, so it feels like we can simplify to just:

```
    logBullet(`Writing your new SDK configuration to ${connectorYamlPath}`);
    await config.askWriteProjectFile(
      path.relative(config.projectDir, connectorYamlPath),
      sdkInfo.connectorYamlContents,
    );
```

joehan (commented):
Mostly LGTM with some minor simplification suggestions - and thank you for cleaning things up in config.ts while you were here!

src/init/features/dataconnect/sdk.ts
maneesht: Ah, the reason for that was that we call `actuate` from the tests, and don't have a `Config` to pass to `actuate`

maneesht: Alternatively, I was thinking of doing something like this:
```
export async function actuate(sdkInfo: SDKInfo, config: Config, skipCheck?: boolean) {
  const connectorYamlPath = `${sdkInfo.connectorInfo.directory}/connector.yaml`;
  logBullet(`Writing your new SDK configuration to ${connectorYamlPath}`);
  if(skipCheck) {
    await fs.writeFile(connectorYamlPath, sdkInfo.connectorYamlContents);
  } else {
    await config.askWriteProjectFile(
      path.relative(config.projectDir, connectorYamlPath),
      sdkInfo.connectorYamlContents,
    );
  }
  ```

Comment (joehan) [reply src/init/features/dataconnect/sdk.ts]: It's bad practice to have branch behavior for tests - then, we're testing something different than what were actually running. You can just make a config in the test like so:
```
new Config({}, { projectDir: "test", cwd: "test" }),
```
You'll also want to stub out askWriteProjectFile in the test (see https://github.com/firebase/firebase-tools/blob/master/src/init/features/functions.spec.ts#L80 for an example)

maneesht: That's fair, I should do this the *right* way then :-)

joehan (approved):

PR #8371, opened 2025-03-26
Set `FIREBASE_DATA_CONNECT_EMUALTOR_HOST`instead of `FIREBASE_DATACONNECT_EMUALTOR_HOST`

joehan:
### Description
🤦‍♀️ 

Also minor clean up for some ugly error handling in pglite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - `apphosting:secrets:grantAccess` can now grant access to emails and can grant multiple secrets at once (#8357)
 - `apphosting:secrets:set` now has flows to help with test secrets (#8359)
 - Updated `init dataconnect` templates to `v1`.
+- Fixed a bug where the emulator suite would set `FIREBASE_DATACONNECT_EMUALTOR_HOST` instead of `FIREBASE_DATA_CONNECT_EMUALTOR_HOST`.
 - Fixed a bug in Auth emulator where accounts:lookup is case-sensitive for emails (#8344)
 - Fixed a bug where function deploys didn't support shorthand notation and parameterized service account (#8366)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - `apphosting:secrets:grantAccess` can now grant access to emails and can grant multiple secrets at once (#8357)
 - `apphosting:secrets:set` now has flows to help with test secrets (#8359)
 - Updated `init dataconnect` templates to `v1`.
+- Fixed a bug where the emulator suite would set `FIREBASE_DATACONNECT_EMUALTOR_HOST` instead of `FIREBASE_DATA_CONNECT_EMUALTOR_HOST`.
```
maneesht, line 14:
```suggestion
- Fixed a bug where the emulator suite would set `FIREBASE_DATACONNECT_EMULATOR_HOST` instead of `FIREBASE_DATA_CONNECT_EMULATOR_HOST`.
```

src/emulator/env.ts
```diff
@@ -46,6 +46,9 @@ export function setEnvVarsForEmulators(
         break;
       case Emulators.DATACONNECT:
         env[Constants.FIREBASE_DATACONNECT_EMULATOR_HOST] = host;
```
maneesht, line 48: I *think* you need to include the protocol for the emulator, so:
```suggestion
        env[Constants.FIREBASE_DATACONNECT_EMULATOR_HOST] = `http://${host}`;
```

maneesht (changes requested):

maneesht (approved):

src/emulator/env.ts
yuchenshi: For the record the protocol should never be set, see https://github.com/firebase/firebase-tools/pull/8396

PR #8380, opened 2025-03-27
Cleaning up logging in dataconnect:sql:setup

joehan:
### Description
A couple logging clean ups:
- Respect the silent flag as much as possible in permissionsSetup
- Move Google Analytics logs to debug level
- Move some default case logs for SQL set up to debug level
- Rename file to use camelcase.

### Scenarios Tested
Screenshot 2025-03-27 at 2 46 20 PM

### DIFF
--- a/src/gcp/cloudsql/permissionsSetup.ts
+++ b/src/gcp/cloudsql/permissionsSetup.ts
@@ -1,3 +1,5 @@
+import * as clc from "colorette";
+
 import { Options } from "../../options";
 import {
   firebaseowner,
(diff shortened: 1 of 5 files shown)


src/gcp/cloudsql/permissionsSetup.ts
```diff
@@ -343,21 +340,23 @@ export async function setupBrownfieldAsGreenfield(
   const schema = schemaInfo.name;
 
   const firebaseOwnerRole = firebaseowner(databaseId, schema);
-  const uniqueTablesOwners = filterTableOwners(schemaInfo, databaseId);
+  const nonFirebasetablesOwners = [...new Set(schemaInfo.tables.map((t) => t.owner))].filter(
```
tammam-g, line 343: Not sure if this is intentional or not but this is a regression, I think we should keep filterTableOwners or make sure to be also filter CLOUDSQL_SUPER_USER.

src/gcp/cloudsql/permissionsSetup.ts
```diff
@@ -394,8 +393,8 @@ export async function brownfieldSqlSetup(
 ) {
   const schema = schemaInfo.name;
 
-  // Step 1: Grant firebasesuperuser access to the original owner.
-  const uniqueTablesOwners = filterTableOwners(schemaInfo, databaseId);
```
tammam-g, line 398: Same as above

tammam-g (changes requested):

src/gcp/cloudsql/permissionsSetup.ts
joehan: Definitely, this was just a screwed up merge. Fixing now

tammam-g (approved):

PR #8394, opened 2025-04-01
fix(functions,emulators): loadTriggers() with top level await

muryoh:
### Description

Context: ESM functions with firestore triggers
The expected behavior of `loadTriggers()` is to
try and `require()` the function entrypoint,
falling back to dynamic `import()` when node raises a ERR_REQUIRE_ESM error. But node may also raise an ERR_REQUIRE_ASYNC_MODULE error when the module has top level awaits

This commit simply makes it so ERR_REQUIRE_ASYNC_MODULE also falls back to `import()`

### Scenarios Tested

Tested on our (private) repo

Let me know if tests need to be added/updated :)

### DIFF
--- a/src/deploy/functions/runtimes/node/triggerParser.js
+++ b/src/deploy/functions/runtimes/node/triggerParser.js
@@ -21,7 +21,7 @@ async function loadModule(packageDir) {
   try {
     return require(packageDir);
   } catch (e) {
-    if (e.code === "ERR_REQUIRE_ESM") {
+    if (e.code === "ERR_REQUIRE_ESM" && e.code !== "ERR_REQUIRE_ASYNC_MODULE") {
       const modulePath = require.resolve(packageDir);
       // Resolve module path to file:// URL. Required for windows support.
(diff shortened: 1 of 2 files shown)


taeold:
This is great contribution 🎉 . Think we can add tests later when functions bring support for Node24.

src/deploy/functions/runtimes/node/triggerParser.js
```diff
@@ -21,7 +21,7 @@ async function loadModule(packageDir) {
   try {
     return require(packageDir);
   } catch (e) {
-    if (e.code === "ERR_REQUIRE_ESM") {
+    if (e.code === "ERR_REQUIRE_ESM" && e.code !== "ERR_REQUIRE_ASYNC_MODULE") {
```
neelance, line 24: This condition seems wrong. `e.code !== "ERR_REQUIRE_ASYNC_MODULE"` is always true if `e.code === "ERR_REQUIRE_ESM"`.

JFGHT: @taeold friendly reminder.

PR #8396, opened 2025-04-02
Remove protocol from DATA_CONNECT_EMULATOR_HOST.

yuchenshi:
### Description

The protocol should never be set according to go/emulator-suite-admin-sdk and setting it breaks the Node.js Admin SDK.

Symptoms include error messages like `getaddrinfo ENOTFOUND http` because it's trying to resolve the word `http` via DNS.

The other environment variable `FIREBASE_DATA_CONNECT_EMULATOR_HOST` is consumed by the JS SDK, which requires a protocol right now. We've decided not to remove the protocol for now to avoid breaking the JS SDK (e.g. when used within the Functions emulator where this env var is automatically set).

### Scenarios Tested

Not tested

### Sample Commands

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,2 @@
 - Fixed a bug where the Admin SDK fails with ENOTFOUND when automatically connecting to the Data Connect emulator when run in the Functions emulator. (#8379)
+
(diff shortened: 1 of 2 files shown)


maneesht (approved):

joehan (approved):
Thank you for adding the detailed comments too!

PR #8398, opened 2025-04-02
Fix issue where prompt would print boolean if false

maneesht:
When no frameworks are detected, `Which falseframeworks` prints out. Instead, we should just print `Which frameworks`
Before:
```
? Which falseframeworks would you like to generate SDKs for? Press Space to select features, then Enter to confirm your choices. (Press <space> to select, <a> to toggle all, <i> to invert selection, and <enter> to proceed)
❯◯ react
 ◯ angular
```
After:
```
? Which frameworks would you like to generate SDKs for? Press Space to select features, then Enter to confirm your choices. (Press <space> to select, <a> to toggle all, <i> to invert selection, and <enter> to proceed)
❯◯ react
 ◯ angular
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,2 @@
-* Fix issue with prompt when Data Connect prompts user for framework generation.
+- Fixed issue with prompt when Data Connect prompts user for framework generation.
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1 @@
+* Fix issue with prompt when Data Connect prompts user for framework generation.
```
```suggestion
- Fixed issue with prompt when Data Connect prompts user for framework generation.
```

joehan (approved):

PR #8411, opened 2025-04-08
Update inquirer library

inlined:
The new inquirer library is said to have many fixes and performance improvements. Upgrade our use. While I intended to only upgrade a bit at a time, I kept pulling at the string trying to fix the graphical glitches and was never successful. Will continue trying to fix glitches in a dependent PR

Accidentally closed the last PR while trying to fix the conflicts. Recreated.


src/prompt.ts
```diff
@@ -1,170 +1,276 @@
...
+};
+
+// Personal hack deviating from inquirer to allow string values to propagate
+// as strings or enum arrays without needing an explicit type at the call stie.
+type MaybeLiteral<Value> = Value extends string ? Value : never;
+
+/**
+ * Options for the checkbox function.
+ *
+ * Epxorted because Inquirer does not export its own input configs anymore. Some unused
```
joehan, line 124:
```suggestion
 * Ecported because Inquirer does not export its own input configs anymore. Some unused
```

src/prompt.ts
```diff
@@ -1,170 +1,276 @@
...
-    autocomplete: AutocompletePrompt.AutocompleteQuestionOptions<T>;
+export { Separator } from "@inquirer/prompts";
+
+export interface BasicOptions<T> {
+  message: string;
+  default?: T;
+  force?: boolean;
+  nonInteractive?: boolean;
+}
+
```
joehan, line 12:
```suggestion

/**
* Exported for testing only
*/
```

src/prompt.ts
```diff
@@ -1,170 +1,276 @@
...
+ * Will abort if nonInteractive and not force
+ */
+export async function confirm(opts: ConfirmConfig): Promise<boolean>;
+
+export async function confirm(opts: string | ConfirmConfig) {
+  if (typeof opts === "string") {
+    opts = { message: opts };
+  } else {
+    if (opts.force) {
+      // TODO: Should we print what we've forced?
```
joehan, line 91: Good question - I like the idea (especially since it gives a 'paper trail' when running noninteractive commands). We probably need to audit the places we use this tho to make sure it doesn't lead to too many redundant logs

src/prompt.ts
```diff
@@ -1,170 +1,276 @@
...
-export async function promptOnce<A>(question: Question, options: Options = {}): Promise<A> {
-  // Need to replace any .'s in the question name - otherwise, Inquirer puts the answer
-  // in a nested object like so: `"a.b.c" => {a: {b: {c: "my-answer"}}}`
-  question.name = question.name?.replace(/\./g, "/") || "question";
-  await prompt(options, [question]);
-  return options[question.name];
+/**
+ * Options for the checkbox function.
+ *
+ * Epxorted because Inquirer does not export its own input configs anymore. Some unused
```
joehan, line 157:
```suggestion
 * Ecported because Inquirer does not export its own input configs anymore. Some unused
```

src/prompt.ts
```diff
@@ -1,170 +1,276 @@
...
+    ...opts,
+    loop: false,
+  });
 }
 
 /**
- * Confirm if the user wants to continue
+ * Options for the number function.
+ *
+ * Epxorted because Inquirer does not export its own input configs anymore. Some unused
```
joehan, line 186:
```suggestion
 * Exported because Inquirer does not export its own input configs anymore. Some unused
```

src/prompt.ts
```diff
@@ -1,170 +1,276 @@
...
   }
+
+  return (await inquirer.number({ ...opts, required: true }))!;
 }
 
 /**
- * Prompts for a directory name, and reprompts if that path does not exist
+ * Options for the checkbox function.
+ *
+ * Epxorted because Inquirer does not export its own input configs anymore. Some unused
```
joehan, line 223:
```suggestion
 * Exported because Inquirer does not export its own input configs anymore. Some unused
```

joehan (approved):
Took a relatively detailed pass through this + played around with it locally, and I think it looks good. Some nits, but LGTM.

Thanks for taking the time to drive this significant refactor!

PR #8412, opened 2025-04-08
Code to create emulators.yaml uses backend root not project root

inlined:
Quicker fix than I realized.

We may want to re-add some of the previous behavior to detect multiple environments and determine which to use as the base config. I was more focused on making breaking changes before the major release, but that was interesting/useful, esp if people only have apphosting[.environment].yaml and not a generic apphosting.yaml.

That might require more though though as we may want to consolidate backendRoot/apphosting.mybackend.yaml and /apphosting.yaml for mult-backend setups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,1 @@
-fix: Prompt to create apphosting.emulator.yaml works with backends that are not at the project.root (#8412)
+- Fixed an issue where the prompt to create apphosting.emulator.yaml did not work with backends that are not at the project.root (#8412)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,1 @@
-- Upgraded `inquirer` library to address some visual bugs with prompting (#8389)
+fix: Prompt to create apphosting.emulator.yaml works with backends that are not at the project.root (#8412)
--- a/src/apphosting/config.ts
+++ b/src/apphosting/config.ts
@@ -218,11 +218,11 @@ export async function maybeAddSecretToYaml(
  */
 export async function maybeGenerateEmulatorYaml(
   projectId: string | undefined,
-  repoRoot: string,
+  backendRoot: string,
 ): Promise<Env[] | null> {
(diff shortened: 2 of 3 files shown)


inlined:
NOTE: This PR is part of my chain for debugging cursor glitches, but if approved, I'll probably rebase onto next directly, commit there, and then re-pull it into inlined.inquirer.

mathu97 (approved):

annajowang (approved):

CHANGELOG.md
```diff
@@ -0,0 +1 @@
+fix: Prompt to create apphosting.emulator.yaml works with backends that are not at the project.root (#8412)
```
```suggestion
- Fixed an issue where the prompt to create apphosting.emulator.yaml did not work with backends that are not at the project.root (#8412)
```

Clean up to match our usual style.

src/apphosting/config.ts
```diff
@@ -218,11 +218,11 @@ export async function maybeAddSecretToYaml(
  */
 export async function maybeGenerateEmulatorYaml(
   projectId: string | undefined,
-  repoRoot: string,
+  backendRoot: string,
 ): Promise<Env[] | null> {
   // Even if the app is in /project/app, the user might have their apphosting.yaml file in /project/apphosting.yaml.
   // Walk up the tree to see if we find other local files so that we can put apphosting.emulator.yaml in the right place.
-  const basePath = dynamicDispatch.discoverBackendRoot(repoRoot) || repoRoot;
+  const basePath = dynamicDispatch.discoverBackendRoot(backendRoot) || backendRoot;
```
joehan, line 225:
```suggestion
  const basePath = dynamicDispatch.discoverBackendRoot(backendRoot) ?? backendRoot;
```
Nit: IIUC, ?? is preferred over || for undefined fallbacks

joehan (approved):

src/apphosting/config.ts
inlined: I personally prefer a style where you only use `??` when you know `null` and/or `0` is a valid value.

PR #8419, opened 2025-04-09
Keep artifact registry dry run off for policy changes

### LINKED ISSUE #8418: New cleanup policy applying in Dry Run mode

**firebase-tools:**
14.1.0

**Platform:**
Linux
### [REQUIRED] Test case
In a project with no gcf-artifacts artifact registry cleanup policy set, apply the Artifact registry cleanup policy with the default 1 day when prompted. 

Apply policy. Look at policy applied in the web-based console.

Expected the policy to be there and be in 'Delete artifacts' mode.

Policy is present but in 'Dry run' mode.

From the coding in this repository and the messaging, it looks like it's intended to set up in the 'Delete artifacts' mode rather than 'Dry run' so I couldn't immediately see where it went wrong. It did happen on 3 projects though so it wasn't a fluke of not have the artifact registry administrator role as happened on the first project.

colerogers:
Fixes #8418 

It seems that the [Repository resource](https://cloud.google.com/artifact-registry/docs/reference/rest/v1/projects.locations.repositories#resource:-repository) has the field `cleanupPolicyDryRun` set to `true` by default on the GCF repos. This change fixes forces that field to be `false` when using the `firebase functions:artifacts:setpolicy` command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Keep artifact registry dry run off for policy changes #8419
--- a/src/functions/artifacts.ts
+++ b/src/functions/artifacts.ts
@@ -173,6 +173,7 @@ export async function setCleanupPolicy(
       ...repository.cleanupPolicies,
       ...generateCleanupPolicy(daysToKeep),
     },
+    cleanupPolicyDryRun: false,
     labels,
   };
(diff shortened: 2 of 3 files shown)


taeold (approved):
wow great catch 😭

PR #8429, opened 2025-04-15
Let users create paid CSQl instances

joehan:
### Description
Allow users to create paid CSQL instances when the free trial has already been used.

### Scenarios Tested
Ran firebase init on a project that already had used the free trial:
Screenshot 2025-04-15 at 12 09 44 PM
Verified that the created instance has the `nt` label value:
Screenshot 2025-04-15 at 12 10 21 PM
TODO once instance provisioning completes - deploy a schema that uses vectors and confirm that it can enable Google ML extension.


fredzqm (approved):
Thank you!!

PR #8443, opened 2025-04-18
Support Firebase sutdio port forwarding in emulator suite

joehan:
### Description
Did some rewiring to support Firebase Studios port forwarding logic. 

 In Firebase Studio, if you serve something on localhost:4000, it will be available on `4000-idx-<workspaceId>.<clusterId>.cloudworkstations.dev`. We need to provide this cloudworkstations.dev URL to the various emalators and SDKs so that corss emulator calls point to the right place.

Rough design:

- When this env var is set, we'll update the host/port for emulators to be `host: <originalHost>-<MONOSPACE_PORT_FORWARDING_HOST> port: 80`. We'll provide these new values in the env vars that the Functions/app hosting emulators use, and we'll provide these in the /api/configs endpoint that the emulator UI uses. 

Stuff that's still broken:
- CORS problems abound - haven't really dug into these yet, but pretty much every call across hosts fails.
- The UI has [some wonky logic](https://github.com/firebase/firebase-tools-ui/blob/master/src/components/common/EmulatorConfigProvider.tsx#L256) that forces the Websocket Host onto the same host as the HTTP host. In IDX, sicne they are on different ports, they are gonna be on different hosts.
- Probably more!

### Using it in Firebase Studio
If you want to test against this in Firebase Studio, the easiest way is probably to use NPX - in any firebase command, replace `firebase` with:
`npx -y firebase/firebase-tools#jh-idx-hax` - so for example:
```
npx -y firebase/firebase-tools#jh-idx-hax emulators:start
```

### DIFF
--- a/src/emulator/env.ts
+++ b/src/emulator/env.ts
@@ -88,7 +88,7 @@ export async function getCredentialsEnvironment(
   return credentialEnv;
 }
 
-export function maybeUseMonospacePortForwarding(emulatorInfos: EmulatorInfo[]): EmulatorInfo[]{
+export function maybeUseMonospacePortForwarding(emulatorInfos: EmulatorInfo[]): EmulatorInfo[] {
   const portForwardingHost = process.env.MONOSPACE_PORT_FORWARDING_HOST;
   if (process.env.MONOSPACE_ENV && portForwardingHost) {
     for (const info of emulatorInfos) {
@@ -99,7 +99,7 @@ export function maybeUseMonospacePortForwarding(emulatorInfos: EmulatorInfo[]): 
       }
       const url = `${info.port}-${portForwardingHost}`;
       info.host = url;
-      info.listen = info.listen?.map(l => {
+      info.listen = info.listen?.map((l) => {
         l.address = url;
         l.port = 80;
         return l;
--- a/src/emulator/env.ts
+++ b/src/emulator/env.ts
@@ -109,7 +109,7 @@ export function maybeUseMonospacePortForwarding(emulatorInfos: EmulatorInfo[]): 
       if (fsInfo.webSocketPort) {
         fsInfo.webSocketHost = `${fsInfo.webSocketPort}-${portForwardingHost}`;
         fsInfo.webSocketPort = 80;
-      } 
+      }
     }
   }
   return emulatorInfos;
--- a/src/emulator/env.ts
+++ b/src/emulator/env.ts
@@ -85,3 +85,19 @@ export async function getCredentialsEnvironment(
   }
   return credentialEnv;
 }
+
+export function maybeUseMonospacePortForwarding(emulatorInfos: EmulatorInfo[]): EmulatorInfo[]{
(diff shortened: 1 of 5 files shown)


src/emulator/ui.ts
```diff
@@ -47,29 +48,23 @@ export class EmulatorUI extends ExpressBasedEmulator {
...
     // Exposes the host and port of various emulators to facilitate accessing
     // them using client SDKs. For features that involve multiple emulators or
-    // hard to accomplish using client SDKs, consider adding an API below.
+    // hard to accomplish using client SDKs, consider adding an API below
     app.get(
       "/api/config",
       this.jsonHandler(() => {
+        const emulatorInfos = (hub! as EmulatorHub).getRunningEmulatorsMapping();
+        maybeUseMonospacePortForwarding(Object.values(emulatorInfos));
         const json = {
```
richieforeman, line 60: Super nit: is there a structural type for this?

src/emulator/env.ts
```diff
@@ -14,6 +15,7 @@ export function setEnvVarsForEmulators(
   env: Record<string, string | undefined>,
   emulators: EmulatorInfo[],
 ): void {
+  maybeUseMonospacePortForwarding(emulators);
```
avolkovi, line 18: let's not use internal codenames in public facing code, just "maybeUsePortForwarding"

src/emulator/env.ts
```diff
@@ -85,3 +87,30 @@ export async function getCredentialsEnvironment(
   }
   return credentialEnv;
 }
+
+export function maybeUseMonospacePortForwarding(emulatorInfos: EmulatorInfo[]): EmulatorInfo[] {
+  const portForwardingHost = process.env.MONOSPACE_PORT_FORWARDING_HOST;
```
avolkovi, line 92: Can we make the Environment variable more generic? PORT_FORWARDING_HOST?

src/emulator/env.ts
```diff
@@ -85,3 +87,30 @@ export async function getCredentialsEnvironment(
...
+  if (process.env.MONOSPACE_ENV && portForwardingHost) {
```
avolkovi, line 93: can this variable be FIREBASE_STUDIO_ENV or just not be checked at all?

avolkovi (changes requested):

src/emulator/env.ts
joehan: This one already exists and is in use elsewhere (in firebase-vscode). However, it is  somewhat redundant here, so I'll remove it

src/emulator/env.ts
joehan: Sure - renaming to FIREBASE_PORT_FORWARDING_HOST

src/emulator/ui.ts
joehan: Added one, good call

avolkovi (approved):
this is much cleaner!! thank you

PR #8451, opened 2025-04-22
Fix adhoc queries

hlshen:
Fix an issue where adhoc mutations weren't passing the document correctly for execution.


fredzqm (approved):
Thanks!

PR #8453, opened 2025-04-22
fix init genkit wrong package for version

ifielker:
Previously the genkit cli was part of the genkit package. Now it's part of genkit-cli. They were in sync for a long time so we didn't notice the error.


src/init/features/genkit/index.ts
```diff
@@ -66,7 +66,7 @@ async function getGenkitVersion(): Promise<GenkitInfo> {
     genkitVersion = process.env.GENKIT_DEV_VERSION;
   } else {
     try {
-      genkitVersion = await spawnWithOutput("npm", ["view", "genkit", "version"]);
+      genkitVersion = await spawnWithOutput("npm", ["view", "genkit-cli", "version"]);
```
pavelgj, line 69: that's not right either... downstream you use this version to install both `genkit` and `genkit-cli` packages... 
We need separate `getGenkitCLIVersion()` and `getGenkitRuntimeVersion()` functions.

ifielker: We only install 'genkit'  if the version is < 0.6.0. That can only happen when the environment variable is used to override the latest. So looking for genkit-cli if the environment variable isn't set is the correct behaviour since it will always be > 0.6.

ifielker: Also when do we want to drop support for the super old 0.6.0 and below?

src/init/features/genkit/index.ts
```diff
@@ -183,40 +208,19 @@ export async function doSetup(setup: GenkitSetup, config: Config, options: Optio
...
-      } else {
-        await wrapSpawn(
-          "npm",
-          ["install", "-g", `genkit-cli@${genkitInfo.genkitVersion}`],
-          projectDir,
-        );
-        await genkitSetup(options, genkitInfo, projectDir);
-        showStartMessage(config, `cd ${setup.functions.source} && npm run genkit:start`);
-      }
+      await wrapSpawn("npm", ["install", "-g", `genkit-cli@${genkitInfo.cliVersion}`], projectDir);
```
pavelgj, line 213: where do you install `genkit@${genkitInfo.genkitVersion}`?

src/init/features/genkit/index.ts
```diff
@@ -39,49 +39,62 @@ import {
...
+  vertexVersion: string;
+  googleAiVersion: string;
   templateVersion: string;
-  useInit: boolean;
   stopInstall: boolean;
 }
 
 // This is the next breaking change version past the latest template.
 const UNKNOWN_VERSION_TOO_HIGH = "2.0.0";
+const MIN_VERSION = "0.6.0";
```
pavelgj, line 51: 0.9.0?
we kinda technically have https://www.npmjs.com/package/genkit/v/0.6.0-dev.2

ifielker: The code didn't change. I just put it into a variable.

PR #8461, opened 2025-04-24
Initial scaffolding for Firebase MCP Server.

mbleigh:
This adds a new `firebase mcp` command that starts an MCP server. I've scaffolded various namespaces of tools to activate to make it configurable. It relies on the existing CLI auth and project detection logic to get the "current" project.

I had to do some very rough low-level short-circuiting of the normal command flow because (a) the logger breaks MCP when used as stdio transport and (b) for some reason when starting the server as part of a `Command` action it exits immediately instead of waiting for input in a way I haven't been able to debug yet.

### TODO List (incomplete)

- [ ] Implement a bunch more tools
- [ ] Figure out how to allow specifying a working directory (Cline doesn't do the right thing by default, for example)
- [ ] Make everything much more resilient


mbleigh:
@joehan it's been a while since I contributed to the CLI...are the CI tests just flaky? What's the best method for getting a green PR?

joehan:
> @joehan it's been a while since I contributed to the CLI...are the CI tests just flaky? What's the best method for getting a green PR?

Yeah, the VSCode tests have been flaky, but they are not required for merging.  Harold has it on his plate to fix them, but feel free to ignore them unless you are directly working on VSCE

src/bin/mcp.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { silenceStdout } from "../logger";
+silenceStdout();
+
+import { FirebaseMcpServer } from "../mcp/index";
+
+const cmd = new Command("mcp").before(requireAuth);
+
+export async function mcp() {
+  const options: any = {};
```
joehan, line 14: No action required. mostly just a note to self:
cmd.prepare is probably in need of a refactor to be strongly typed and return Promise<Options> instead of mutating the input.

src/bin/firebase.ts
```diff
@@ -11,160 +12,11 @@ if (!semver.satisfies(nodeVersion, pkg.engines.node)) {
...
-
-if (!handlePreviewToggles(args)) {
-  // determine if there are any arguments. if not, display help
-  if (!args.length) {
-    client.cli.help();
-  } else {
-    cmd = client.cli.parse(process.argv);
-  }
+// we short-circuit the normal process for MCP
+if (isEnabled("mcp") && process.argv[2] === "experimental:mcp") {
```
joehan, line 16: Probably OK to just make the command `firebase mcp`, since its already behind an experiment flag

src/mcp/tool.ts
```diff
@@ -0,0 +1,31 @@
+import { CallToolResult } from "@modelcontextprotocol/sdk/types";
+import { z, ZodTypeAny } from "zod";
+import { zodToJsonSchema } from "zod-to-json-schema";
+
+export interface ServerTool<InputSchema extends ZodTypeAny = ZodTypeAny> {
+  mcp: {
+    name: string;
+    description?: string;
+    inputSchema: any;
```
joehan, line 9: Could probably make this more strongly typed. It looks like this is expected to be JSON schema, so  we could either roll our own limited version of that, or use https://github.com/StefanTerdell/zod-to-json-schema/blob/master/src/parseTypes.ts#L58 (i think it is exported?)

src/mcp/tools/project/get_sdk_config.ts
```diff
@@ -0,0 +1,49 @@
+/* eslint camelcase: 0 */
```
joehan, line 1: Dumb question - why do we need to not use camelcase names for these tools?

src/mcp/tools/project/list_apps.ts
```diff
@@ -0,0 +1,34 @@
+/* eslint camelcase: 0 */
+
+import { z } from "zod";
+import { tool } from "../../tool.js";
+import { mcpError, toContent } from "../../util.js";
+import { AppPlatform, listFirebaseApps } from "../../../management/apps.js";
+
+const PLATFORM_MAP: Record<string, AppPlatform> = {
```
joehan, line 8: Repeated in a few places - wanna just put this in src/management/apps.js?

src/mcp/tools/project/list_apps.ts
```diff
@@ -0,0 +1,34 @@
...
+        .describe("the specific platform to list (omit to list all platforms)"),
+    }),
+    annotations: {
+      title: "List Firebase Apps",
+      readOnlyHint: true,
+    },
+  },
+  async ({ platform }, { projectId }) => {
+    if (!projectId) return mcpError("No current project detected.");
+    const apps = await listFirebaseApps(projectId, PLATFORM_MAP[platform || ""] || AppPlatform.ANY);
```
joehan, line 31: Nit: IIUC, this will return all apps if I call it with platform= 'foo'.  Is there something in MCP that enforces the enum? Or do we need to do so here?

src/track.ts
```diff
@@ -352,9 +352,9 @@ function isDebugMode(): boolean {
   if (account?.user.email.endsWith("@google.com")) {
     try {
       require("../tsconfig.json");
-      logger.debug(
-        `Using Google Analytics in DEBUG mode. Emulators (+ UI) events will be shown in GA Debug View only.`,
-      );
+      // logger.debug(
```
joehan, line 355: Uncomment this? It only shows up for googlers anyway

src/mcp/index.ts
```diff
@@ -0,0 +1,65 @@
...
+        activeFeatures: this.activeFeatures,
+      },
+    });
+  }
+
+  async mcpCallTool(request: CallToolRequest): Promise<CallToolResult> {
+    const toolName = request.params.name;
+    const toolArgs = request.params.arguments;
+    const tool = this.getTool(toolName);
+    if (!tool) throw new Error(`Tool '${toolName}' could not be found.`);
```
joehan, line 57: Nit: Might be helpful to show a different message when tool exists but is not active.

src/mcp/tools/project/get_sdk_config.ts
```diff
@@ -0,0 +1,49 @@
...
+      app_id: z.string().nullish().describe("the specific app id to fetch"),
+    }),
+    annotations: {
+      title: "Get Firebase SDK Config",
+      readOnlyHint: true,
+    },
+  },
+  async ({ platform: inputPlatform, app_id: appId }, { projectId }) => {
+    if (!projectId) return mcpError("No current project detected.");
+    let platform = PLATFORM_MAP[inputPlatform || ""];
```
joehan, line 33: Nit: Prefer ?? over || in cases like this.

src/mcp/tools/project/list_apps.ts
joehan, line 31: ?? over ||

joehan (approved):
LGTM overall, mostly nits and suggesting that we just call the command mcp

src/bin/firebase.ts
mbleigh: So I actually intend to remove the experiment flag before we launch - most MCP servers are installed and run with an `npx` command so I want to make sure this can be run on a fresh machine's install without needing to have done any pre-work (I have additional work to do around not requiring auth etc).

So the idea is someone could just add `npx -y firebase-tools experimental:mcp` as their MCP server command and even if they've never used Firebase before it gets them started. If we require an experiment flag, they'd have to download the CLI separately.

I hoped that there was a command line flag to enable an experiment but don't see one so this is my current approach. Maybe I should add that flag instead.

src/mcp/tool.ts
mbleigh: Yeah, ideally this would have a slightly better type but the MCP SDK has its own idea of what the type is that isn't exported well and fundamentally this is an internal implementation detail so I'm going to leave it as-is for now.

src/mcp/tools/project/list_apps.ts
mbleigh: Decided to just rework this a little to avoid needing the map.

src/mcp/tools/project/get_sdk_config.ts
mbleigh: Technically we could make all of this stuff camelCase, but:

1. Models are trained heavily on Python tools and `snake_case` seems to be the most common.
2. I thought it was better to have the exports be named consistently with the underlying tool named consistently with the file...but it's really a choice to do consistency and disable the lint error over inconsistency but lint passes.

TBH I hate ESLint 😆

src/mcp/tools/project/list_apps.ts
mbleigh: Yeah the MCP SDK enforces schema on the way in so the Zod enum should give us coverage here.

src/mcp/index.ts
mbleigh: I think I'll leave as is because realistically this is basically a "should never happen" error. The client would have to call a tool that was not returned in the listTools response.

src/track.ts
mbleigh: Ah yeah this was back when I was trying to do MCP as a normal command and the logs going to stdout were blowing things up. Uncommented.

PR #8462, opened 2025-04-24
Add in disable_user to auth calls

nohe427:
### Description

This adds in an authentication ability to disable users based on their UID in a Firebase project. This is used in an additional effort for a different project.

### Scenarios Tested

n/a, will be tested in a separate, larger PR

### Sample Commands

No commands are being added in this PR.


nohe427:
@joehan - Does this match project conventions?

@mbleigh - FYI

nohe427:
Replaced by #8480 8480

PR #8464, opened 2025-04-25
Dont require auth for demo projects

joehan:
### Description
If using a `demo-` project, we should not require auth. Doing so is pointless since we should not make any prod calls (and if we do, they will hit a nonexistant project and fail)

Also fixed a place where we'd make a prod call for 'demo-' projects when emulating webframeworks. This call would always fail and we'd use a fallback value, so instead we'll skip the call and just use the fallback


davideast (approved):
LGTM

PR #8468, opened 2025-04-27
[MCP] Flexible directory selection and command line arguments.

mbleigh:
This makes the scaffold more robust to handling MCP launch in different environments.

- If started up without a `--dir` command or `PROJECT_ROOT` env, there are tools made available to set the project dir.
- Each tool call dynamically detects project etc. to allow for the above.
- Also adds support for `--only` to activate a subset of features.

Overall this makes the scaffold more robust to a variety of clients - using in Claude Code will probably be non-directory-specific, using in Cursor will usually be directory-specific.


src/bin/mcp.ts
```diff
@@ -1,19 +1,37 @@
...
+
+const STARTUP_MESSAGE = `
+This is a running process of the Firebase MCP server. This command should only be executed by an MCP client. An example MCP client configuration might be:
 
-const cmd = new Command("mcp").before(requireAuth);
+{
+  "mcpServers": {
+    "firebase": {
+      "command": "firebase",
+      "args": ["experimental:mcp", "--dir", "/path/to/firebase/project"]
```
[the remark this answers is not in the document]
mbleigh: I don't think we want that in the startup message because I'd rather have good defaults (i.e. eventually only turn on features present in `firebase.json`), but it should be in docs/README.

mbleigh:
@joehan can you give this one a runthrough?

src/mcp/README.md
```diff
@@ -0,0 +1,43 @@
+# Firebase MCP Server
+
+The Firebase MCP Server provides assistive utilities to compatible AI-assistent
```
joehan, line 3:
```suggestion
The Firebase MCP Server provides assistive utilities to compatible AI-assistant
```

src/mcp/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Example: Cursor
+
+In `.cursor/mcp.json` in your workspace directory, add the Firebase MCP server:
+
+```json
+{
+  "mcpServers": {
+    "firebase": {
+      "command": "npx",
```
joehan, line 23: I see we are recommending npx commands in cursor rules, etc. Unfortunately, now that package.json has multiple binaries, npx works a lot less nicely with us and we'll need to pass the --package flag (see https://github.com/npm/npx) to specify which binary to run.

Do we actually need to have this as a separate binary? It seems like `experimental:mcp` should be sufficient

src/mcp/index.ts
```diff
@@ -10,55 +10,104 @@ import {
...
+  constructor(options: { activeFeatures?: ServerFeature[]; projectRoot?: string }) {
     this.activeFeatures = options.activeFeatures;
-    this.cliOptions = options.cliOptions;
     this.server = new Server({ name: "firebase", version: SERVER_VERSION });
-    this.server.registerCapabilities({ tools: { listChanged: false } });
+    this.server.registerCapabilities({ tools: { listChanged: true } });
     this.server.setRequestHandler(ListToolsRequestSchema, this.mcpListTools.bind(this));
     this.server.setRequestHandler(CallToolRequestSchema, this.mcpCallTool.bind(this));
+    this.projectRoot =
+      options.projectRoot ||
```
joehan, line 39: Nit: ?? over || for undefined fallbacks

src/mcp/tools/core/get_firebase_directory.ts
```diff
@@ -0,0 +1,28 @@
+import { z } from "zod";
+import { tool } from "../../tool.js";
+import { mcpError, toContent } from "../../util.js";
+import { detectProjectRoot } from "../../../detectProjectRoot.js";
+
+export const get_firebase_directory = tool(
+  {
+    name: "get_firebase_directory",
+    description:
+      "Gets the current Firbase project directory. If this has been set using the `set_firebase_directory` tool it will return that, otherwise it will look for a PROJECT_ROOT environment variable or the current working directory of the running Firebase MCP server.",
```
joehan, line 10:
```suggestion
      "Gets the current Firebase project directory. If this has been set using the `set_firebase_directory` tool it will return that, otherwise it will look for a PROJECT_ROOT environment variable or the current working directory of the running Firebase MCP server.",
```

src/mcp/tools/core/set_firebase_directory.ts
```diff
@@ -0,0 +1,43 @@
...
+      title: "Set Firebase Project Directory",
+      idempotentHint: true,
+    },
+  },
+  ({ dir }, { host }) => {
+    if (dir === null) {
+      host.setProjectRoot(null);
+      return Promise.resolve(
+        toContent(
+          `Firebase MCP project directory setting deleted. New project root is: ${host.projectRoot || "unset"}`,
```
joehan, line 30: Seems like this should never get hit from looking at the code, but seems fine to have a fallback

src/track.ts
```diff
@@ -352,9 +352,9 @@ function isDebugMode(): boolean {
   if (account?.user.email.endsWith("@google.com")) {
     try {
       require("../tsconfig.json");
-      logger.debug(
-        `Using Google Analytics in DEBUG mode. Emulators (+ UI) events will be shown in GA Debug View only.`,
-      );
+      // logger.debug(
```
joehan, line 355: Undo this (or  change the if statement to not trigger in MCP mode)

joehan (approved):
Code looks good to me, one larger question about whether we need multiple binaries

src/mcp/README.md
mbleigh: !!! Oops, I forgot to remove the second binary which is now redundant. Done!

PR #8472, opened 2025-04-28
Add in the get_auth_user tool

nohe427:
### Description

This adds in the get_auth_user tool to the MCP to lookup specific users from within Firebase authentication.

### Scenarios Tested

Have yet to test

### DIFF
--- a/src/gcp/auth.ts
+++ b/src/gcp/auth.ts
@@ -1,6 +1,56 @@
 import { Client } from "../apiv2";
 import { identityOrigin } from "../api";
 
+interface MfaEnrollment {
+  mfaEnrollmentId: string;
+  displayName: string;
+  enrolledAt: string;
+  phoneInfo?: string;
+  emailInfo?: {
+    emailAddress: string;
+  };
+  unobfuscatedPhoneInfo?: string;
+}
+
+interface UserInfo {
+  localId: string;
+  email: string;
+  displayName: string;
+  language: string;
+  photoUrl: string;
+  timeZone: string;
+  dateOfBirth: string;
+  passwordHash: string;
+  salt: string;
+  version: number;
+  emailVerified: boolean;
+  passwordUpdatedAt: number;
+  providerUserInfo: {
+    providerId: string;
+    displayName: string;
+    photoUrl: string;
+    federatedId: string;
+    email: string;
+    rawId: string;
+    screenName: string;
+    phoneNumber: string;
+  }[];
+  validSince: string;
+  disabled: boolean;
+  lastLoginAt: string;
+  createdAt: string;
+  screenName: string;
+  customAuth: boolean;
+  rawPassword?: string;
+  phoneNumber: string;
+  customAttributes: string;
+  emailLinkSignin: boolean;
+  tenantId: string;
+  mfaInfo: MfaEnrollment[];
+  initialEmail: string;
+  lastRefreshAt: string;
+}
+
 const apiClient = new Client({ urlPrefix: identityOrigin(), auth: true });
 
 /**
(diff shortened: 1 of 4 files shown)


src/mcp/tools/auth/index.ts
```diff
@@ -0,0 +1,6 @@
+/* eslint camelcase: 0 */
```
joehan, line 1: Probably not for this PR, but it would be great if we could put this in .eslintrc.js instead of having this at the top of every file. Ideally, we'd exempt the mcp directory from the camlecase rule.

src/gcp/auth.ts
```diff
@@ -36,3 +86,49 @@ export async function updateAuthDomains(project: string, authDomains: string[]):
...
+ * @param limit the amount of results to return.
+ * @param offset the amount to page through the results
+ * @return an array of user info
+ */
+export async function getAuthUser(
+  project: string,
+  email?: string,
+  phone?: string,
+  uid?: string,
+  limit = 1000,
```
joehan, line 105: This seems wonky... I'd expect getAuthUser to return a single UserInfo, not an array.
I'd recommend renaming this to `searchAuthUser`. 

Alternatively, you could:
- get rid of the limit and offset arguments, and switch the return type to UserInfo
- hard code the limit to 1 & unpack the results within this function
- return an error if there are no matching users

src/gcp/auth.ts
```diff
@@ -1,6 +1,56 @@
...
+    screenName: string;
+    phoneNumber: string;
+  }[];
+  validSince: string;
+  disabled: boolean;
+  lastLoginAt: string;
+  createdAt: string;
+  screenName: string;
+  customAuth: boolean;
+  rawPassword?: string;
```
joehan, line 44: Probably safe to omit input only fields.

src/gcp/auth.ts
```diff
@@ -36,3 +86,49 @@ export async function updateAuthDomains(project: string, authDomains: string[]):
...
+export async function getAuthUser(
+  project: string,
+  email?: string,
+  phone?: string,
+  uid?: string,
+  limit = 1000,
+  offset = 0,
+): Promise<{ recordCount: string; userInfo: UserInfo }[]> {
+  const expression: { email?: string; phoneNumber?: string; userId?: string } = {};
+  if (email !== undefined) {
```
joehan, line 109: Are these `!== undefined` checks needed? I think we could probably simplify this whole block to:
```
const expressions = { email, phoneNumber: phone, userId: uid };
```

src/gcp/auth.ts
nohe427: I was hoping to use this for 2 seperate tools. The first is to get one and only one user, the second is to list all users or search for a particular user.

src/gcp/auth.ts
nohe427: Good call, removed.

src/gcp/auth.ts
mbleigh: I'd probably have separate `listUsers({project,limit?})` and `findUser({project,email?,uid?,phone?})` commands, it's a little weird to do both in one.

nohe427: Ok, I can do that.

nohe427:
Fixed some JSON formatting. This should be g2g:

image

src/mcp/tools/auth/index.ts
nohe427: Nevermind, fixed in this CL.

src/gcp/auth.ts
```diff
@@ -36,3 +85,41 @@ export async function updateAuthDomains(project: string, authDomains: string[]):
   );
   return res.body.authorizedDomains;
 }
+
+/**
+ * getAuthUser
```
joehan, line 90: Nit: Update this.
```suggestion
 * findUser searches for an auth user in a project.
```

joehan (approved):

nohe427:
Closed in favor of #8478

PR #8479, opened 2025-04-29
Enhance `firebase init apphosting` to support local source deploys

blidd-google:
In order to support deploys from local source to App Hosting, the CLI needs a way to identify which backend to target with the local source code. `firebase init apphosting` provides an interactive flow to guide users through the process of either creating a new backend or linking an existing backend to a local app, then writes the configuration to `firebase.json` under a new section identified by key `apphosting`.


src/init/features/apphosting.ts
```diff
@@ -1,17 +1,143 @@
...
+  utils.logSuccess("Firebase initialization complete!");
+}
+
+// Exported for unit testing.
+export async function promptExistingBackend(projectId: string): Promise<string> {
+  while (true) {
+    const backendId = await promptOnce({
+      name: "backendId",
+      type: "input",
+      message: "Provide a backend name:",
```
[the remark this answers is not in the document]
leslie546: Hi, can you help me with my app hosting that I'm stuck with firebase CLI

src/init/features/apphosting.ts
```diff
@@ -1,17 +1,155 @@
...
+                name: parseBackendName(result.original.name).id,
+                value: parseBackendName(result.original.name).id,
+              };
+            }),
+        ]),
+      );
+    },
+  });
+  return backendId;
+}
```
annajowang, line 142: Can you put this function in https://github.com/firebase/firebase-tools/blob/master/src/apphosting/backend.ts

src/init/features/apphosting.ts
```diff
@@ -1,17 +1,155 @@
...
+      return new Promise((resolve) =>
+        resolve([
+          ...fuzzy
+            .filter(input, backends, {
+              extract: (backend) => parseBackendName(backend.name).id,
+            })
+            .map((result) => {
+              return {
+                name: parseBackendName(result.original.name).id,
+                value: parseBackendName(result.original.name).id,
```
annajowang, line 134: The original UX concern is that the user would have to know off the top of their head the name of their backends. 

This is slightly better since they don't have to know the exact name, but afaict we're not displaying the list of backends for them.

what I originally had in mind was to just list all the backends like:
https://github.com/firebase/firebase-tools/blob/49e4897f5bfceeb4e7995d0dd5f31b483610f825/src/apphosting/backend.ts#L452

this requires zero mental work on the user while autocompleting still requires some thinking

src/init/features/apphosting.ts
```diff
@@ -1,17 +1,155 @@
...
+    const backend = await createBackend(
+      projectId,
+      location,
+      backendId,
+      /* serviceAccount= */ null,
+      /* repository= */ undefined,
+      webApp?.id,
+    );
+    createBackendSpinner.succeed(`Successfully created backend!\n\t${backend.name}\n`);
+  } else {
```
annajowang, line 85: there is 0% chance that "createLink" is something other than "link" or "create". If we are pretty certain that something won't happen, it's not worth clouding the code to catch those cases. I would keep the code cleaner with just:
```
if (createOrLink === "link") {
    backendConfig.backendId = await promptExistingBackend(projectId);
  } else {
 ....
  }
```

src/init/features/apphosting.ts
annajowang: Since we are on a time crunch and we want to merge this quickly to rebase the other PR, i'm fine with a TODO instead to address this comment and the one below, but please make sure to address asap so it does not linger.

annajowang (approved):
approval once some comments are addressed

src/init/features/apphosting.ts
blidd-google: The above actually implements both - this will list the user's backends while also filtering the list based on the user's typed input. We implemented this UX for prompting users to select a repository during the backend onboarding flow.

annajowang: ok sg, sorry didn't confirm if code was showing users the list.

PR #8480, opened 2025-04-29
[MCP] disable_auth_user

nohe427:
### Description

This adds in an authentication ability to disable users based on their UID in a Firebase project via MCP

### Scenarios Tested

Manually tested using the MCP Inspector


src/mcp/tools/auth/disable_auth_user.ts
```diff
@@ -0,0 +1,35 @@
...
+  {
+    name: "disable_auth_user",
+    description: "Disables or enables a user based on a UID.",
+    inputSchema: z.object({
+      uid: z.string().describe("The localId or UID of the user to disable or enable"),
+      disabled: z.boolean().describe("true disables the user, false enables the user"),
+    }),
+    annotations: {
+      title: "Disable or enable a particular user",
+      readOnlyHint: true,
```
mbleigh, line 16: This isn't readonly - it should probably be marked as dangerous and idempotent. See https://modelcontextprotocol.io/docs/concepts/tools#tool-definition-structure

src/mcp/tools/auth/disable_auth_user.ts
```diff
@@ -0,0 +1,35 @@
...
+      if (res) {
+        return toContent(`User ${uid} as been ${disabled ? "disabled" : "enabled"}`);
+      }
+      return toContent(`Failed to ${disabled ? "disable" : "enable"} user ${uid}`);
+    } catch (err: unknown) {
+      if (err instanceof Error) {
+        return mcpError(err.message);
+      } else {
+        return mcpError("unknown error");
+      }
```
mbleigh, line 32: Could you refactor `mcpError` to do this check?

Check for (1) is it an error (if so, use the message) or (2) is it a string (if so, return it) or (3) is it something else (return unknown error)

That way we can just reliably do `return mcpError(err)`

mbleigh (approved):
Approved with a couple little nits.

src/mcp/tools/auth/disable_auth_user.ts
nohe427: Fixed

src/mcp/tools/auth/disable_auth_user.ts
nohe427: Whoops, good catch

PR #8482, opened 2025-04-30
[MCP] Add auth claims

nohe427:

### DIFF
--- a/src/gcp/auth.ts
+++ b/src/gcp/auth.ts
@@ -191,12 +191,12 @@ export async function setCustomClaim(
   if (user.uid !== uid) {
     throw new Error(`Could not find ${uid} in the auth db, please check the uid again.`);
   }
-  let attributeJson = new Map<string, string | number | boolean>();
-  if (user.customAttributes !== undefined && user.customAttributes !== "") {
-    attributeJson = JSON.parse(user.customAttributes) as Map<string, string | number | boolean>;
-  }
   let reqClaim = JSON.stringify(claim);
   if (options?.merge) {
+    let attributeJson = new Map<string, string | number | boolean>();
+    if (user.customAttributes !== undefined && user.customAttributes !== "") {
+      attributeJson = JSON.parse(user.customAttributes) as Map<string, string | number | boolean>;
+    }
     reqClaim = JSON.stringify({ ...attributeJson, ...claim });
   }
   const res = await apiClient.post<
--- a/src/gcp/auth.ts
+++ b/src/gcp/auth.ts
@@ -13,7 +13,8 @@ interface MfaEnrollment {
 }
 
 interface UserInfo {
-  localId: string;
+  uid?: string;
+  localId?: string;
   email: string;
   displayName: string;
   language: string;
@@ -177,19 +178,18 @@ export async function disableUser(
 export async function setCustomClaim(
   project: string,
   uid: string,
-  claim: string,
-  value: string | number | boolean,
+  claim: Record<string, unknown>,
+  options?: {merge?: boolean}
 ): Promise<UserInfo> {
   let user = await findUser(project, undefined, undefined, uid);
   if (user.localId !== uid) {
     throw new Error(`Could not find ${uid} in the auth db, please check the uid again.`);
   }
-  const newClaim = { [claim]: value };
   let attributeJson = new Map<string, string | number | boolean>();
   if (user.customAttributes !== undefined && user.customAttributes !== "") {
     attributeJson = JSON.parse(user.customAttributes) as Map<string, string | number | boolean>;
   }
-  const reqClaim = JSON.stringify({ ...attributeJson, ...newClaim });
+  const reqClaim = JSON.stringify({ ...attributeJson, ...claim });
   const res = await apiClient.post<
     { customAttributes: string; targetProjectId: string; localId: string },
     SetAccountInfoResponse
--- a/src/gcp/auth.ts
+++ b/src/gcp/auth.ts
@@ -189,7 +189,7 @@ export async function setCustomClaim(
   if (user.customAttributes !== undefined && user.customAttributes !== "") {
     attributeJson = JSON.parse(user.customAttributes) as Map<string, string | number | boolean>;
   }
-  const reqClaim = JSON.stringify({ ...newClaim, ...attributeJson });
+  const reqClaim = JSON.stringify({ ...attributeJson, ...newClaim });
   const res = await apiClient.post<
     { customAttributes: string; targetProjectId: string; localId: string },
     SetAccountInfoResponse
--- a/src/gcp/auth.ts
+++ b/src/gcp/auth.ts
@@ -42,7 +42,7 @@ interface UserInfo {
   screenName: string;
   customAuth: boolean;
   phoneNumber: string;
-  customAttributes: string;
+  customAttributes?: string;
   emailLinkSignin: boolean;
   tenantId: string;
   mfaInfo: MfaEnrollment[];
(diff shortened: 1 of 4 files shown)


src/gcp/auth.ts
```diff
@@ -165,3 +165,42 @@ export async function disableUser(
...
+ * @param claim the key value in the custom claim.
+ * @param value the value in the custom claim.
+ * @return the results of the accounts update request.
+ */
+export async function setCustomClaim(
+  project: string,
+  uid: string,
+  claim: string,
+  value: string | number | boolean,
+): Promise<UserInfo> {
```
mbleigh, line 182: At the API layer I'd prefer this be:

```ts
setCustomClaims(project: string, uid: string, claims?: Record<string,unknown>, options?: {merge?: boolean})
```

It's fine if the tool only sets one claim at a time, but if the underlying API supports multiple let's do multiple.

Also I believe custom claims can be more than just scalar values, e.g. `{roles: ["admin", "moderator"]}`

src/gcp/auth.ts
```diff
@@ -165,3 +165,42 @@ export async function disableUser(
...
+  let user = await findUser(project, undefined, undefined, uid);
+  if (user.localId !== uid) {
+    throw new Error(`Could not find ${uid} in the auth db, please check the uid again.`);
+  }
+  const newClaim = { [claim]: value };
+  let attributeJson = new Map<string, string | number | boolean>();
+  if (user.customAttributes !== undefined && user.customAttributes !== "") {
+    attributeJson = JSON.parse(user.customAttributes) as Map<string, string | number | boolean>;
+  }
+  const reqClaim = JSON.stringify({ ...newClaim, ...attributeJson });
```
mbleigh, line 192: Isn't this backwards? We want the newClaims to override the existing values.

src/mcp/tools/auth/set_auth_claims.ts
```diff
@@ -0,0 +1,31 @@
...
+export const set_auth_claims = tool(
+  {
+    name: "set_auth_claims",
+    description: "Sets a list of specific claims for a user.",
+    inputSchema: z.object({
+      uid: z.string().describe("the UID or localId of the user to update"),
+      claim: z.string().describe("the key value int he custom claim to update"),
+      value: z
+        .union([z.string(), z.boolean(), z.number()])
+        .describe("the value of the custom claim"),
```
mbleigh, line 15: Hmm, do we want to allow setting multiple claims at once, or just one? If just one, it should probably be `set_auth_claim`.

src/mcp/tools/auth/set_auth_claims.ts
```diff
@@ -0,0 +1,31 @@
+import { z } from "zod";
+import { tool } from "../../tool.js";
+import { mcpError, toContent } from "../../util.js";
+import { setCustomClaim } from "../../../gcp/auth.js";
+
+export const set_auth_claims = tool(
+  {
+    name: "set_auth_claims",
+    description: "Sets a list of specific claims for a user.",
```
mbleigh, line 9: I've seen advice that these should be more descriptive.

```suggestion
    description: "Sets custom claims on a specific user's account. Use to create trusted values associated with a user e.g. marking them as an admin. Claims are limited in size and should be succinct in name and value.",
```

src/mcp/tools/auth/set_auth_claims.ts
```diff
@@ -0,0 +1,31 @@
...
+    inputSchema: z.object({
+      uid: z.string().describe("the UID or localId of the user to update"),
```
mbleigh, line 11: What is a `localId`? I think it's always a uid, no?

src/mcp/tools/auth/set_auth_claims.ts
```diff
@@ -0,0 +1,31 @@
...
+      claim: z.string().describe("the key value int he custom claim to update"),
```
mbleigh, line 12:
```suggestion
      claim: z.string().describe("the name (key) of the claim to update, e.g. 'admin'"),
```

src/mcp/tools/auth/set_auth_claims.ts
```diff
@@ -0,0 +1,31 @@
...
+      value: z
+        .union([z.string(), z.boolean(), z.number()])
+        .describe("the value of the custom claim"),
+    }),
+    annotations: {
+      title: "Set a custom claim on a specific user.",
+      destructiveHint: true,
```
mbleigh, line 19:
```suggestion
```

I don't think this is destructive per se (unless we want to mark *all* write ops as destructive). I think since there's a `readonlyHint` and a `destructiveHint` that leaves a middleground for "neither hint" which means it's not readonly but it's also not destructive.

nohe427: Ahh fair. I can change this

src/mcp/tools/auth/set_auth_claims.ts
nohe427: When the API returns with user info from identity platform, it always returns a `localId` rather than a UID

src/gcp/auth.ts
nohe427: You're right. Fixing

src/mcp/tools/auth/set_auth_claims.ts
mbleigh: Let's correct the output to always change "localId" to "uid" in MCP responses then. Nobody should have to know about that particular bit of sausage 🙃

src/mcp/tools/auth/set_auth_claims.ts
nohe427: I would prefer one at a time since I imagine most folks that interact with the MCP would say : give user@myApp.com custom claim admin = true. WDYT?

mbleigh: I'm good with it, but let's change the name accordingly.

src/gcp/auth.ts
nohe427: What should merge do? Should it overwrite all existing claims or should it change the value of:

`roles: admin` to `roles: [admin, moderator]`?

src/mcp/tools/auth/set_auth_claims.ts
nohe427: One last question: for the value arg... Would the type be `z.unknown()` or perhaps something else?

mbleigh: Ugh, might need to experiment a bit with Gemini to figure out what's possible. If you do `z.any()` it will blow up. Maybe for now do:

```ts
z.union([
  z.string(),  
  z.number(),
  z.boolean(),
  z.record(z.union(z.string(),z.number(),z.boolean())),
  z.array(z.union(z.string(),z.number,z.boolean())
]);
```

Which allows for arrays of scalars and objects with scalars but not deep nesting.

src/gcp/auth.ts
mbleigh: Simple merge. So if {merge: true} you do `{...oldValues, newValues}` and if `{merge: false}` you do `{newValues}` (overwrite all claims).

The tool should always merge.

src/mcp/tools/auth/set_auth_claims.ts
nohe427: Done!

src/mcp/tools/auth/set_auth_claims.ts
nohe427: Fixed!

src/gcp/auth.ts
nohe427: Done!

src/gcp/auth.ts
```diff
@@ -165,3 +171,45 @@ export async function disableUser(
...
+  options?: { merge?: boolean },
+): Promise<UserInfo> {
+  let user = await findUser(project, undefined, undefined, uid);
+  if (user.uid !== uid) {
+    throw new Error(`Could not find ${uid} in the auth db, please check the uid again.`);
+  }
+  let attributeJson = new Map<string, string | number | boolean>();
+  if (user.customAttributes !== undefined && user.customAttributes !== "") {
+    attributeJson = JSON.parse(user.customAttributes) as Map<string, string | number | boolean>;
+  }
```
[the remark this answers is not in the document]
nohe427: Ahh fair call. Migrating to the merge block

PR #8487, opened 2025-04-30
[MCP] All tools all the time w/ instructive errors

mbleigh:
This adds instructive error messages for signed out users and users without a project directory set. It also adds automatic error checks for `authRequired` and `projectRequired` tools. We can remove these checks once emulators are supported.

### DIFF
--- a/src/mcp/errors.ts
+++ b/src/mcp/errors.ts
@@ -1,6 +1,14 @@
-import { mcpError } from "./util";
+import { commandExistsSync, mcpError } from "./util";
 
 export const NO_PROJECT_ERROR = mcpError(
-  "No active project was found. Use the `set_firebase_directory` command to set the project directory (containing firebase.json).",
+  'No active project was found. Use the `set_firebase_directory` tool to set the project directory to an absolute folder location containing a firebase.json config file. Alternatively, change the MCP server config to add [...,"--dir","/absolute/path/to/project/directory"] in its command-line arguments.',
   "PRECONDITION_FAILED",
 );
+
+export function mcpAuthError() {
+  const cmd = commandExistsSync("firebase") ? "firebase" : "npx -y firebase-tools";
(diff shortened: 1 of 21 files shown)


src/mcp/errors.ts
```diff
@@ -1,6 +1,14 @@
...
 
 export const NO_PROJECT_ERROR = mcpError(
-  "No active project was found. Use the `set_firebase_directory` command to set the project directory (containing firebase.json).",
+  'No active project was found. Use the `set_firebase_directory` tool to set the project directory to an absolute folder location containing a firebase.json config file. Alternatively, change the MCP server config to add [...,"--dir","/absolute/path/to/project/directory"] in its command-line arguments.',
   "PRECONDITION_FAILED",
 );
+
+export function mcpAuthError() {
+  const cmd = commandExistsSync("firebase") ? "firebase" : "npx -y firebase-tools";
+  return mcpError(`The user is not currently logged into the Firebase CLI, which is required to use this tool. Please instruct the user to execute this shell command to sign in or to configure [Application Default Credentials][ADC] on their machine.
```
joehan, line 10: Should we add a docs link explaining how to set up ADC? IDK if that's actually helpful for AIs.

src/mcp/tools/auth/disable_auth_user.ts
```diff
@@ -17,10 +17,13 @@ export const disable_auth_user = tool(
       destructiveHint: true,
       idempotentHint: true,
     },
+    _meta: {
```
joehan, line 20: Clean up unused imports in this file (and the rest of the tools)

joehan (approved):
LGTM, after tiny cleanups

src/mcp/errors.ts
mbleigh: Oh yes I literally intended to that's why it looks like a Markdown link.

PR #8492, opened 2025-05-01
[MPC] FDC Query Generation

nohe427:


nohe427:
image

src/mcp/tools/dataconnect/generate_dataconnect_operation.ts
```diff
@@ -0,0 +1,35 @@
...
+    description: "Generates a Firebase Data Connect Operation based on the deployed app schema.",
+    inputSchema: z.object({
+      prompt: z
+        .string()
+        .describe("A description of the query or mutation operation you would like to generate"),
+      service: z
+        .string()
+        .describe(
+          "The service id or name of the deployed Firebase Data Connect Schema in format: projects/<project-id>/locations/<location>/services/<service-name>",
+        ),
```
mbleigh, line 18: Should be `.nullish()` and described as optional, saying that the service from firebase.json will be used if none provided. Should also accept just an id without the full name string which should then list services to look up which one it is.

src/mcp/tools/dataconnect/generate_dataconnect_operation.ts
```diff
@@ -0,0 +1,35 @@
+import { z } from "zod";
+import { tool } from "../../tool.js";
+import { toContent } from "../../util.js";
+import { generateOperation } from "../../../gif/fdcExperience.js";
+
+export const generate_dataconnect_operation = tool(
+  {
+    name: "generate_dataconnect_operation",
+    description: "Generates a Firebase Data Connect Operation based on the deployed app schema.",
```
mbleigh, line 9:
```suggestion
    description: "Generates a single Firebase Data Connect query or mutation based on the currently deployed schema and the provided prompt.",
```

src/mcp/tools/dataconnect/generate_dataconnect_operation.ts
```diff
@@ -0,0 +1,35 @@
...
+    inputSchema: z.object({
+      prompt: z
+        .string()
+        .describe("A description of the query or mutation operation you would like to generate"),
```
mbleigh, line 13: If there's any guidance we have about how to make a "good" prompt we should include a blurb here.

nohe427: https://cloud.google.com/gemini/docs/discover/write-prompts

Added some of this to the blurb

src/mcp/tools/dataconnect/generate_dataconnect_operation.ts
nohe427: Done

src/mcp/tools/dataconnect/generate_dataconnect_operation.ts
```diff
@@ -0,0 +1,47 @@
...
+    _meta: {
+      requiresProject: true,
+      requiresAuth: true,
+      // TODO: Create an endpoint to check for GiF activiation.
+    },
+  },
+  async ({ prompt, service }, { projectId, host }) => {
+    const serviceInfo = await pickService(
+      projectId!,
+      Config.load({ cwd: host.projectRoot })!,
```
fredzqm, line 41: nit: Just added `config` as a context passed to all tools.

We can re-use that

src/mcp/tools/dataconnect/generate_dataconnect_operation.ts
```diff
@@ -0,0 +1,47 @@
...
+      prompt: z
+        .string()
+        .describe(
+          "Write the prompt like you're talking to a person, describe the task you're trying to accomplish and give details that are specific to the users requst",
+        ),
+      service: z
+        .string()
+        .nullish()
+        .describe(
+          "Optional: Uses the service from the firebase.json file if nothing provided. The service id or name of the deployed Firebase Data Connect Schema in format: projects/<project-id>/locations/<location>/services/<service-name>.",
```
fredzqm, line 24: This should just be the serviceID.

`pickService` only expects serviceID, not the full resource~

It's probably simpler for the LLM to reason about.

fredzqm (commented):
LGTM.

Want @hlshen to check the GiF part a bit more

src/mcp/tools/dataconnect/generate_dataconnect_operation.ts
nohe427: Done!

src/mcp/tools/dataconnect/generate_dataconnect_operation.ts
nohe427: Done, I believe

src/mcp/tools/dataconnect/generate_dataconnect_operation.ts
```diff
@@ -0,0 +1,42 @@
...
+        .string()
+        .describe(
+          "Write the prompt like you're talking to a person, describe the task you're trying to accomplish and give details that are specific to the users requst",
+        ),
+      serviceId: z
+        .string()
+        .nullish()
+        .describe(
+          "Optional: Uses the serviceId from the firebase.json file if nothing provided. The serviceId of the deployed Firebase resource.",
+        ),
```
fredzqm, line 24: This is effectively the generate_connector in console. It would use the deployed schema in the backend~

It's a great starting point, but I think we would want one with local schema.

LGTM. Unblock you~

fredzqm (approved):

src/mcp/tools/dataconnect/generate_dataconnect_operation.ts
nohe427: Yes, I agree. Not sure if GiF can handle local schema. Maybe we can fix in future update if available?

PR #8497, opened 2025-05-01
Add metrics for MCP

joehan:
### Description
Add simple metrics for MCP - now we get # of calls per tool, # of errors per tool, and # of times list tools is called.

### DIFF
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -120,7 +120,7 @@ export class FirebaseMcpServer {
     try {
       const config = Config.load({ cwd: this.projectRoot });
       const res = await tool.fn(toolArgs, { projectId: await this.getProjectId(), host: this, config });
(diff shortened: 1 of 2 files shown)


src/mcp/index.ts
```diff
@@ -115,8 +117,11 @@ export class FirebaseMcpServer {
     if (tool.mcp._meta?.requiresProject && !projectId) return NO_PROJECT_ERROR;
 
     try {
-      return tool.fn(toolArgs, { projectId: await this.getProjectId(), host: this });
+      const res = tool.fn(toolArgs, { projectId: await this.getProjectId(), host: this });
+      await trackGA4("mcp_tool_call", { tool_name: toolName, error: 0 });
+      return res;
     } catch (err: unknown) {
+      await trackGA4("mcp_tool_call", { tool_name: toolName, error: 1 });
```
mbleigh, line 124: It should also mark as an error if `isError` is true in the returned content.

mbleigh (approved):

PR #8509, opened 2025-05-05
Update MCP README.md

fredzqm:
### Description

Improve onboard instructions

### Scenarios Tested

Screenshot 2025-05-05 at 12 06 47 PM

### DIFF
--- a/src/mcp/README.md
+++ b/src/mcp/README.md
@@ -12,6 +12,14 @@ as an argument to operate against the project of your working directory.
 For clients that don't operate within a specific workspace, the Firebase MCP
 server makes tools available to read and write a project directory.
 
+
+### Enable Firebase CLI experiments
+
+Firebase MCP server command aren't visible unless you enable the experiements.
+```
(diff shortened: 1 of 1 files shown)


src/mcp/README.md
```diff
@@ -12,6 +12,14 @@ as an argument to operate against the project of your working directory.
...
 server makes tools available to read and write a project directory.
 
+
+### Enable Firebase CLI experiments
+
+Firebase MCP server command aren't visible unless you enable the experiements.
+```
+npx -y firebase-tools experiments:enable mcp
+```
+
```
mbleigh, line 22: I'm planning to remove the need to enable the experiment before I/O -- that's why it has the `experimental` name in the command itself.

fredzqm: When we do that, we can strip `experiments` in instructions & cmd.

Trying to make it easier for folks to test it out right now. Every time, I try to send folks this cmd. 😄

PR #8510, opened 2025-05-05
Remove globs from default Firestore rules

joehan:
### Description
Remove globs from default Firestore rules. The replacement rules are sourced from our docs, and are effectively closed to all reads/writes

### DIFF
--- a/templates/init/firestore/firestore.rules
+++ b/templates/init/firestore/firestore.rules
@@ -1,16 +1,14 @@
 service cloud.firestore {
   match /databases/{database}/documents {
-    match /{document=**} {
-      // This rule allows anyone with your database reference to view, edit,
-      // and delete all data in your database. It is useful for getting
-      // started, but it is configured to expire after 30 days because it
-      // leaves your app open to attackers. At that time, all client
-      // requests to your database will be denied.
-      //
-      // Make sure to write security rules for your app before that time, or
-      // else all client requests to your database will be denied until you
-      // update your rules.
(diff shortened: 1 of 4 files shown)


templates/init/firestore/firestore.rules
```diff
@@ -1,16 +1,14 @@
 service cloud.firestore {
   match /databases/{database}/documents {
-    match /{document=**} {
```
avolkovi, line 3: we want to match inside a collection, there are no top level documents in firestore

can you also make the change to RTDB and storage rules?

avolkovi (changes requested):

templates/init/firestore/firestore.rules
joehan: Done

templates/init/firestore/firestore.rules
```diff
@@ -1,16 +1,13 @@
...
-      // Make sure to write security rules for your app before that time, or
-      // else all client requests to your database will be denied until you
-      // update your rules.
-      allow read, write: if request.time < timestamp.date({{IN_30_DAYS}});
+    // Security rules are closed by default - if you don't have a rule that grants access,
+    // clients will not be able to read or write to that location in Firestore.
+    // Below is an example of how to open up a collection to only authenticated users
+    // For more help developing your security rules, see https://firebase.google.com/docs/rules/basics
+    match /some_collection/{document} {
+      allow read, write: if request.auth != null;
```
avolkovi, line 10: wondering if we should keep the time sensitive rules... @sam-gc what do you think?

PR #8515, opened 2025-05-07
App Hosting: Remove sentinel oauth connections

mathu97:
This PR removes the use of a sentinel oauth developer connect connection (named `firebase-app-hosting-github-oauth`). Now if a new dev connect connection is required we look for an app hosting connection that has the oauth credentials so we can create a new connection without making the user go through the oauth dance. The main benefit of this is to make it easier to set up app hosting via Terraform.

Tested the following flows:
1. Creating a new backend without any existing connections -> **Passed**
2. Creating another backend under same installation but new repo (should re-use connection) -> **Passed**
3. Creating a backend in a new location should create a new connection in that location -> **Passed**
4. Create another backend under different installation and new repo (should create another installation) -> **Passed**
5. Repo Linking -> Should be able to create a new git repo link a repo to an existing connection using the internal command -> **Passed**
6. Should just work for existing configs (they have a app-hosting-oauth connection) -> **Passed**


src/apphosting/githubConnections.ts
```diff
@@ -357,27 +352,23 @@ export async function listValidInstallations(
...
-      // Cloud build P4SA requires the secret manager admin role.
-      // This is required when creating an initial connection which is the Oauth connection in our case.
-      await ensureSecretManagerAdminGrant(projectId);
-      conn = await createConnection(projectId, location, APPHOSTING_OAUTH_CONN_NAME);
-    } else {
-      throw err;
-    }
+  const completedConnections = await listAppHostingConnections(projectId, location);
+  if (completedConnections.length > 0) {
+    return completedConnections[0];
```
annajowang, line 365: can you include a comment on why we know [0] is the connection we want?
underlying API call does not seem to be using the "orderBy" field: https://cloud.google.com/developer-connect/docs/api/reference/rest/v1/projects.locations.connections/list

if any connection returned from the list is fine, can you add an explicit comment on why?

src/apphosting/githubConnections.ts
```diff
@@ -357,27 +352,23 @@ export async function listValidInstallations(
...
   }
 
+  const connectionId = generateConnectionId();
```
annajowang, line 368: nit: since we're not using this anywhere else i think we could call generateConnectionId() inline in `createConnectino()`

annajowang (approved):

src/apphosting/githubConnections.ts
mathu97: Added comment.

PR #8516, opened 2025-05-07
Add support for deploying local source to App Hosting

blidd-google:
`firebase deploy` now deploys source code on the user's local machine to the corresponding backend on App Hosting. Guides the user through backend creation if the backend does not exist yet.

Note that `firebase deploy` relies on a new `apphosting` section in the `firebase.json` file, which can be initialized using the enhanced `firebase init apphosting` command (implemented here https://github.com/firebase/firebase-tools/pull/8479).


annajowang:
some changes from https://github.com/firebase/firebase-tools/pull/8479/files are showing up in this PR

could you separate these?

src/apphosting/backend.ts
```diff
@@ -250,12 +251,31 @@ export async function ensureAppHostingComputeServiceAccount(
       );
     }
   }
+  if (deployFromSource) {
```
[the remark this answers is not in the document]
annajowang: ^ for better code comprehension e.g. letting reader know that deploy from source has the source code bundled in storage buckets so we need access to that bucket.

but also, now I see below that `roles/storage.objectViewer` is also added in provisionDefaultComputeServiceAccount() below, which is called in this ensureAppHostingComputeServiceAccount()

and I'm not sure why?

either there's a better way to do it to make things more clear natively or let's document what we're doing and explain any weird/confusing stuff.

annajowang (commented):
initial review of just the first few files in https://github.com/firebase/firebase-tools/compare/b5697e6209372606c2692bafe253c0fc9f9280dd..fdc6255ec928bac4a3871572793d856703674a06?diff=split&w

I skimmed the others. Could you just pull out src/gcp/apphosting.ts,  src/gcp/devConnect.ts, and src/gcp/storage.ts changes into other PRs, and I can take a quick look and then stamp, before continuing with this PR

src/deploy/apphosting/deploy.ts
```diff
@@ -0,0 +1,91 @@
...
+      // Unfortunately, requests for a non-existent bucket from the GCS API sometimes return 403 responses as well as 404s.
+      // We must attempt to create a new bucket on both 403s and 404s.
+      if (errStatus === 403 || errStatus === 404) {
+        logBullet(
+          `Creating Cloud Storage bucket in ${loc} to store App Hosting source code uploads at ${bucketName}...`,
+        );
+        try {
+          await gcs.createBucket(projectId, {
+            name: bucketName,
+            location: "US-CENTRAL1",
```
annajowang, line 40: does the bucket location also need to match the apphosting location? or are we just creating a bucket for each apphosting location?

src/deploy/apphosting/deploy.ts
```diff
@@ -0,0 +1,91 @@
...
+  const backendLocations = context.backendLocations.values() || [];
+  for (const loc of backendLocations) {
+    const bucketName = `firebaseapphosting-sources-${options.projectNumber}-${loc.toLowerCase()}`;
+    try {
+      await gcs.getBucket(bucketName);
+    } catch (err) {
+      const errStatus = getErrStatus((err as FirebaseError).original);
+      // Unfortunately, requests for a non-existent bucket from the GCS API sometimes return 403 responses as well as 404s.
+      // We must attempt to create a new bucket on both 403s and 404s.
+      if (errStatus === 403 || errStatus === 404) {
```
annajowang, line 33: Is this from reading documentation somewhere or is this from testing/experimenting?

I'm wondering if it's the case that errStatus === 403 =>  no permission to see the bucket => and if they don't have permission to see the bucket they definitely don't have permission to create any new bucket? https://cloud.google.com/storage/docs/access-control/iam-roles seems to support this.
If so, it would be better to fail immediately in the 403 case.

annajowang (commented):
reviewed up to src/deploy/apphosting/deploy.ts

src/deploy/apphosting/index.ts
```diff
@@ -0,0 +1,5 @@
+import prepare from "./prepare";
+import deploy from "./deploy";
+import release from "./release";
+
+export { prepare, deploy, release };
```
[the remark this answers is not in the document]
blidd-google: The deploy/apphosting module is being imported here and the functions added to these chains: https://github.com/firebase/firebase-tools/blob/49e4897f5bfceeb4e7995d0dd5f31b483610f825/src/deploy/index.ts#L128

src/deploy/apphosting/deploy.ts
```diff
@@ -0,0 +1,91 @@
...
+import { needProjectId } from "../../projectUtils";
+import { logBullet, logWarning } from "../../utils";
+import { Context } from "./args";
+import { createArchive } from "./util";
+
+/**
+ * Zips and uploads App Hosting source code to Google Cloud Storage in preparation for
+ * build and deployment. Creates storage buckets if necessary.
+ */
+export default async function (context: Context, options: Options): Promise<void> {
```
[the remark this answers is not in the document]
blidd-google: Since the function is a default export from the apphosting/deploy module, it doesn't need a name and can be imported and used like so: `import deploy from "./deploy"; deploy(...)`.

The other deploy functions do the same thing:
https://github.com/firebase/firebase-tools/blob/49e4897f5bfceeb4e7995d0dd5f31b483610f825/src/deploy/dataconnect/deploy.ts#L19

src/deploy/apphosting/deploy.ts
blidd-google: We're creating a bucket on demand for every location that has an deploy-from-source backend in it.

For example: the user is trying to deploy from source backend A in us-central1, but there's no bucket yet in us-central1 so we create it. There is also a backend B in asia-east1, but the user hasn't tried to deploy B from source yet, so we don't create that bucket.

src/deploy/apphosting/deploy.ts
blidd-google: Looking at the owner / editor permissions under https://cloud.google.com/storage/docs/access-control/iam-roles#basic-roles-intrinsic, some users could actually could have `storage.buckets.create` permission, but not `storage.buckets.get` permission. So some of them might legitimately run into the case where they can't check if a bucket exists, but they can create one. Failing immediately on 403s might actually cause their source deploys to fail consistently, so my thinking was that there is less harm in attempting to create the bucket anyway, and if the create operation fails with a 403, we then exit.

annajowang: ok if you ran into it, other users will too. thanks for confirming.

src/deploy/apphosting/prepare.ts
```diff
@@ -0,0 +1,182 @@
...
+  const backendIds: string[] = [];
+  for (const selector of selectors) {
+    // if the user passes the "apphosting" selector, we default to deploying all backends
+    // listed in the user's firebase.json App Hosting config.
+    if (selector === "apphosting") {
+      return backendConfigs;
+    }
+    if (selector.startsWith("apphosting:")) {
+      const backendId = selector.replace("apphosting:", "");
+      if (selector.length > 0) {
```
annajowang, line 165: Should this be `if (backendId.length > 0)` ?

src/deploy/apphosting/prepare.ts
```diff
@@ -0,0 +1,182 @@
...
+    return [];
+  }
+
+  const filtered = [];
+  for (const id of backendIds) {
+    const cfg = backendConfigs.find((cfg) => cfg.backendId === id);
+    if (cfg) {
+      filtered.push(cfg);
+    }
+  }
```
annajowang, line 180: feel like this could all be replaced by
```
filtered = backendConfigs.filter(cfg => backendIds.include(cfg.BackendId));
```

src/deploy/apphosting/prepare.ts
```diff
@@ -0,0 +1,182 @@
...
+      );
+      createBackendSpinner.succeed(`Successfully created backend!\n\t${backend.name}\n`);
+      context.backendConfigs.set(cfg.backendId, cfg);
+      context.backendLocations.set(cfg.backendId, location);
+    } else if (filteredBackends.length === 1) {
+      const backend = filteredBackends[0];
+      const { location } = parseBackendName(backend.name);
+      if (cfg.alwaysDeployFromSource === false) {
+        continue;
+      }
```
annajowang, line 93: nit: move this above defining the two constants, since you don't need them.
for readability, delay creating new variables until right before they're first needed.

src/deploy/apphosting/prepare.ts
```diff
@@ -0,0 +1,182 @@
...
+          logBullet(
+            `Your deployment preferences have been saved to firebase.json. On future invocations of "firebase deploy", your local source will be deployed to my-backend. You can edit this setting in your firebase.json at any time.`,
+          );
+        } else {
+          confirmDeploy = true;
+        }
+        if (!confirmDeploy) {
+          logWarning(`Skipping deployment of backend ${cfg.backendId}`);
+          continue;
+        }
```
annajowang, line 126: I don't think `confirmDeploy` could ever be falsey except within the `if (!options.force) {` if statement. So i think we can keep `confirmDeploy` and the logwarning all inside the if statement and get rid of this extra `else`

src/deploy/apphosting/prepare.ts
```diff
@@ -0,0 +1,182 @@
...
+        projectId,
+        location,
+        cfg.backendId,
+        null,
+        undefined,
+        webApp?.id,
+      );
+      createBackendSpinner.succeed(`Successfully created backend!\n\t${backend.name}\n`);
+      context.backendConfigs.set(cfg.backendId, cfg);
+      context.backendLocations.set(cfg.backendId, location);
```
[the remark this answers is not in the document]
annajowang: This is repeated below:
``` 
context.backendConfigs.set(cfg.backendId, cfg);
context.backendLocations.set(cfg.backendId, location);
```
can you move this to the bottom of the for loop

src/deploy/apphosting/release.ts
```diff
@@ -0,0 +1,59 @@
...
+        },
+      }),
+    );
+  }
+
+  const multipleRolloutsMessage = `Starting new rollouts for backends ${Array.from(context.backendConfigs.keys()).join(", ")}`;
+  const singleRolloutMessage = `Starting a new rollout for backend ${Array.from(context.backendConfigs.keys()).join(", ")}`;
+  logBullet(
+    `${rollouts.length > 1 ? multipleRolloutsMessage : singleRolloutMessage}; this may take a few minutes. It's safe to exit now.`,
+  );
```
annajowang, line 48: i think just saying "Starting rollout(s) for backend(s) ...." is fine and does not degrade user experience.

src/commands/deploy.ts
```diff
@@ -23,6 +23,7 @@ export const VALID_DEPLOY_TARGETS = [
   "remoteconfig",
   "extensions",
   "dataconnect",
+  "apphosting",
 ];
 export const TARGET_PERMISSIONS: Record<(typeof VALID_DEPLOY_TARGETS)[number], string[]> = {
```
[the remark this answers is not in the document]
blidd-google: added calls with `ensureApiEnabled` to check if App Hosting API is enabled. Double checking with team which specific IAM permissions are required

src/deploy/apphosting/prepare.ts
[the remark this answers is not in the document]
blidd-google: I see your point about having the create logic in the same module, I've refactored it into a doSetupSourceDeploy() function in https://github.com/firebase/firebase-tools/blob/master/src/apphosting/backend.ts. However, I think de-duping might be a bit premature; refactoring the duplicate logic into a helper function would require introducing new arguments to that helper (i.e. we would need an optional `backendId` arg since source deploys won't need to prompt for a backend ID, and same thing for `webAppName`), just to avoid repeating three function calls. So as of now, from a readability perspective I'm leaning on the side of keeping separate doSetup and doSetupSourceDeploy. But as the logic expands in the future, I'm open to revisiting the refactor option. WDYT?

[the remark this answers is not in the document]
annajowang: yeah keeping similar code close together, so anyone making changes in one is more likely to not miss making the same changes in the other is my main concern. agree deduping would be a big task right now that can be left for later given timeline.

src/deploy/apphosting/util.ts
```diff
@@ -0,0 +1,93 @@
...
+  const archive = archiver("zip");
+
+  // We must ignore firebase-debug.log or weird things happen if you're in the public dir when you deploy.
+  const ignore = config.ignore || ["node_modules", ".git"];
+  ignore.push("firebase-debug.log", "firebase-debug.*.log");
+  const gitIgnorePatterns = await parseGitIgnorePatterns();
+  ignore.push(...gitIgnorePatterns);
+
+  if (!projectRoot) {
+    projectRoot = process.cwd();
```
annajowang, line 32: do we want to let the user known what directory is being used as their projectRoot?
I think it would be a big deal if the user accidentally deploys the wrong set of files, but I don't know if we're preventing / reducing chances of that happening at some point prior to this.

src/deploy/apphosting/util.ts
```diff
@@ -0,0 +1,93 @@
...
+      resolve([]);
+      return;
+    }
+    const fileStream: fs.ReadStream = fs.createReadStream(absoluteFilePath);
+    const rl: readline.Interface = readline.createInterface({
+      input: fileStream,
+      crlfDelay: Infinity,
+    });
+    rl.on("line", (line: string) => {
+      if (line.startsWith("#") || line.trim() === "") {
```
annajowang, line 67: nit: would feel better if line was trimmed before checking if it starts with "#" as well.

src/deploy/apphosting/util.ts
```diff
@@ -0,0 +1,93 @@
...
+      "Could not read source directory. Remove links and shortcuts and try again.",
+      { original: err as Error, exit: 1 },
+    );
+  }
+  return tmpFile;
+}
+
+async function parseGitIgnorePatterns(filePath = ".gitignore"): Promise<string[]> {
+  const absoluteFilePath: string = path.resolve(filePath);
+  const lines: string[] = [];
```
annajowang, line 55: nit: move `lines` to inside promise?

src/deploy/apphosting/util.ts
blidd-google: Added some log lines. Users can cancel the upload file stream if they realize they uploaded the wrong files.

src/deploy/apphosting/util.ts
```diff
@@ -0,0 +1,94 @@
...
+  } catch (err: unknown) {
+    throw new FirebaseError(
+      "Could not read source directory. Remove links and shortcuts and try again.",
+      { original: err as Error, exit: 1 },
+    );
+  }
+  return { projectSourcePath: projectRoot, zippedSourcePath: tmpFile };
+}
+
+async function parseGitIgnorePatterns(filePath = ".gitignore"): Promise<string[]> {
```
annajowang, line 53: I'd really like us to  not write code where we don't have to. but since we're on a time crunch:
could you leave a TODO to say that this could potentially be replaced by
https://www.npmjs.com/package/parse-gitignore

so whoever touches this next can consider doing it.

annajowang: actually, as bryan said this package essentially does 
"fs.ReadFileSync("path").filter(startsWith("#"))"

I realize it's not clear to me why we need 30+ lines of code instead of just
```
return fs.ReadFileSync(".gitinore").filter( (line) => line.trim().startsWith("#") || line.trim() === ""))
```

blidd-google: Fair enough. The event listening approach is kind of an old school way to do it. Made the changes

src/apphosting/backend.ts
```diff
@@ -340,6 +388,7 @@ async function provisionDefaultComputeServiceAccount(projectId: string): Promise
       "roles/firebaseapphosting.computeRunner",
       "roles/firebase.sdkAdminServiceAgent",
       "roles/developerconnect.readTokenAccessor",
+      "roles/storage.objectViewer",
```
Yuangwang, line 391: I think this is always added even in non source deploy cases whenever there is a 404? Why is that?

blidd-google: It's added in case the customer wants to use deploy the source in the future - if we can add the storage.objectViewer role now, then they won't run into the IAM propagation delay problem in the future.

src/deploy/apphosting/util.ts
```diff
@@ -0,0 +1,74 @@
...
+      { original: err as Error, exit: 1 },
+    );
+  }
+  return { projectSourcePath: projectRoot, zippedSourcePath: tmpFile };
+}
+
+function parseGitIgnorePatterns(filePath = ".gitignore"): string[] {
+  const absoluteFilePath = path.resolve(filePath);
+  const lines = fs
+    .readFileSync(absoluteFilePath)
```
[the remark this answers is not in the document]
blidd-google: added an edge case handling .gitignore not found

PR #8523, opened 2025-05-08
[MCP] Organize MCP tools && update README

fredzqm:
Core tools like `firebase_get_project`, `firebase_set_project`, `firebase_consult_assistant` should always be available.

Directory tools like `set_project_directory`, `get_project_directory` is only available if `--dir` is not set.

Other tools are populated based on the `--only` flags.

image


src/mcp/tools/index.ts
```diff
@@ -3,11 +3,28 @@ import { ServerFeature } from "../types.js";
...
+    toolDefs.push(...directoryTools);
+  }
+  if (!activeFeatures || !activeFeatures.length) {
+    activeFeatures = Object.keys(tools) as ServerFeature[];
+  }
+  for (const key of activeFeatures) {
+    toolDefs.push(...tools[key]);
+  }
+  return toolDefs;
+}
```
fredzqm, line 25: @mbleigh Let me now what you think.

I think this function makes the intention more clear.

The difference between directory tools and core tools were confusing.

mbleigh (approved):
This mostly seems fine to me. One thing I'm wondering about is having another way to slice commands - I think the ones that call GiF should potentially be able to be disabled / excluded by `--only` separately from other tools.

Like I might want the rest of the tools but don't necessarily want consult_firebase, generate_schema, and generate_operation (especially if GiF isn't active in my project)

PR #8525, opened 2025-05-08
[MCP] getDownloadUrl - Firebase Storage

nohe427:


nohe427:
Loosely based on the admin sdk of the same method: https://github.com/firebase/firebase-admin-node/blob/master/src/storage/index.ts#L74

src/mcp/tools/storage/get_download_url.ts
```diff
@@ -0,0 +1,29 @@
+import { z } from "zod";
+import { tool } from "../../tool.js";
+import { toContent } from "../../util.js";
+import { getDownloadUrl } from "../../../gcp/storage.js";
+
+export const get_storage_object_download_url = tool(
```
fredzqm, line 6: nit: `get_object_download_url`.

It will become `storage_get_object_download_url` after this block.

https://github.com/firebase/firebase-tools/blob/master/src/mcp/tools/index.ts#L31

fredzqm (approved):
Looks great!

src/mcp/tools/storage/get_download_url.ts
nohe427: Done!

src/mcp/tools/storage/get_download_url.ts
```diff
@@ -0,0 +1,35 @@
...
+      readOnlyHint: true,
+    },
+    _meta: {
+      requiresProject: true,
+      requiresAuth: true,
+    },
+  },
+  async ({ bucket, objectPath }, { projectId }) => {
+    if (!bucket) {
+      if (!projectId) {
```
fredzqm, line 27: This is also redundant. `requiresProject` enabled a check before calling.

https://github.com/firebase/firebase-tools/blob/284ad53bdb7a1921b37ccc70e09d7f9ee10ab47f/src/mcp/index.ts#L110

nohe427: Ahh, thank you! Removed!

PR #8533, opened 2025-05-09
Refactoring logging to better support MCP server

joehan:
### Description
Refactor logging so that it plays nicely with MCP.

### Scenarios Tested
- Tested that when running as MCP, we write firebase-debug.log and don't leak any logs to STDOUT by running a bunch of tools from inspector.
- Tested out a few basic commands from the command line to ensure that we didn't break normal logging


fredzqm (approved):
Looks reasonable

Issue #8336, opened 2025-03-18
Effectively/Logically hard coded max python version; exceeding results in silent failure and false pseudo-success


paulpv:
I have Python 3.13 installed.
```
% python3.13 --version
Python 3.13.1
```

I am trying to follow the guide at https://firebase.google.com/docs/functions/get-started?gen=2nd and after run firebase functions with:
```
firebase init functions
...
firebase emulators:start
```

The emulator says either...
```
Failed to load function definition from source: FirebaseError: Failed to find location of Firebase Functions SDK: Missing virtual environment at venv directory. Did you forget to run 'python3.12 -m venv venv'?
```
... or, if I run `cd ./functions && python3 -m venv venv`...
```
Failed to load function definition from source: FirebaseError: Failed to find location of Firebase Functions SDK. Did you forget to run '. "./functions/venv/bin/activate" && python3.12 -m pip install -r requirements.txt'?
```

I don't see that logged step documented anywhere.  
Apparently you have to run the equivalent of:
```
pushd functions
python3 -m venv venv
source venv/bin/activate
pip install --upgrade pip
pip install -r requirements.txt
deactivate
popd
firebase emulators:start
```

**But that is all moot because it still fails!**

Looking at `firebase-debug.log`: **EMPHASIS MINE**
```
...
[info] i  functions: Watching "/Users/pv/Dev/GitHub/swooby/swoo.by/functions" for Cloud Functions... {"metadata":{"emulator":{"name":"functions"},"message":"Watching \"/Users/pv/Dev/GitHub/swooby/swoo.by/functions\" for Cloud Functions..."}}
[debug] [2025-03-18T21:38:02.543Z] Customer code is not Node
[debug] [2025-03-18T21:38:02.543Z] Validating python source
[debug] [2025-03-18T21:38:02.543Z] Building python source
[debug] [2025-03-18T21:38:02.543Z] Could not find functions.yaml. Must use http discovery
*****
[debug] [2025-03-18T21:38:02.545Z] Running command with virtualenv: command=., args=["\"/Users/pv/Dev/GitHub/swooby/swoo.by/functions/venv/bin/activate\"","&&","python3.12","-c","\"import firebase_functions; import os; print(os.path.dirname(firebase_functions.__file__))\""]
[debug] [2025-03-18T21:38:02.553Z] stderr: /bin/sh: python3.12: command not found
*****

[...]

PR #8534, opened 2025-05-09
chore: Update Cloud Run functions runtime versions

taeold:
Updates the Node.js and Python runtime versions to align with the latest supported runtimes on Cloud Run (https://cloud.google.com/functions/docs/runtime-support)

Notably, this changes the default runtime of python to be 3.13 instead of 3.12.

Partially addresses https://github.com/firebase/firebase-tools/issues/8336


src/deploy/functions/runtimes/supported/types.ts
```diff
@@ -75,37 +75,43 @@ export const RUNTIMES = runtimes({
     friendly: "Node.js 18",
     status: "GA",
     deprecationDate: "2025-04-30",
```
inlined, line 77: While you're editing this file can yo ualso update Node16 and below to be "decommissioned" for their status?

inlined (approved):

PR #8547, opened 2025-05-12
Adding use_project tool

joehan:
### Description
Adding use_project tool to set the active project.
### Scenarios Tested
Screenshot 2025-05-12 at 11 32 37 AM
Screenshot 2025-05-12 at 11 32 51 AM

### DIFF
--- a/src/mcp/tools/core/use_project.ts
+++ b/src/mcp/tools/core/use_project.ts
@@ -1,0 +1,32 @@
+import { z } from "zod";
+import { tool } from "../../tool.js";
+import { setNewActive } from "../../../commands/use.js";
+import { mcpError, toContent } from "../../util.js";
+
+export const use_project = tool(
+  {
+    name: "use_project",
+    description: "Select a Firebase Project to use for subsequent tool calls.",
+    inputSchema: z.object({
+      project: z.string().describe("The project id, number, or alias to use."),
+    }),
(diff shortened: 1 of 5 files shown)


mbleigh:
Did you forget to check in the actual tool code?

fredzqm:
> Did you forget to check in the actual tool code?

+1 I was looking for that file as well.

joehan:
🤦 I did. Just pushed it up

fredzqm (approved):

src/mcp/tools/core/use_project.ts
```diff
@@ -0,0 +1,32 @@
...
+    description: "Select a Firebase Project to use for subsequent tool calls.",
+    inputSchema: z.object({
+      project: z.string().describe("The project id, number, or alias to use."),
+    }),
+    annotations: {
+      title: "Select a Firebase Project to use for subsequent tool calls.",
+      readOnlyHint: false,
+    },
+    _meta: {
+      requiresAuth: true,
```
fredzqm, line 18: Does it require `auth`?

`firebase use` can be run w/o logging in, right?

joehan: We check that you have access to the project if it is a real one, so its probably simpler to requrie auth for now.

We can always revist this once we add emulator support.

PR #8548, opened 2025-05-12
[MCP] firestore_query_collection

nohe427:

### DIFF
--- a/src/gcp/firestore.ts
+++ b/src/gcp/firestore.ts
@@ -33,6 +33,38 @@ export interface Database {
   etag: string;
 }
 
+export interface StructuredQuery {
+  from: { collectionId: string; allDescendants: boolean }[];
+  where?: {
+    fieldFilter: {
+      field: { fieldPath: string };
+      op:
+        | "OPERATOR_UNSPECIFIED"
+        | "LESS_THAN"
+        | "LESS_THAN_OR_EQUAL"
+        | "GREATER_THAN"
+        | "GREATER_THAN_OR_EQUAL"
+        | "EQUAL"
+        | "NOT_EQUAL"
+        | "ARRAY_CONTAINS"
+        | "ARRAY_CONTAINS_ANY"
+        | "IN"
+        | "NOT_IN";
+      value: FirestoreValue;
+    };
+  };
+  orderBy?: {
+    field: { fieldPath: string };
+    direction: "ASCENDING" | "DESCENDING" | "DIRECTION_UNSPECIFIED";
+  }[];
+  limit?: number;
+}
+
+interface RunQueryResponse {
+  document?: FirestoreDocument;
+  readTime?: string;
+}
+
 export enum DayOfWeek {
   MONDAY = "MONDAY",
   TUEDAY = "TUESDAY",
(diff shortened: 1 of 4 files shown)


src/mcp/tools/firestore/query_collection.ts
```diff
@@ -0,0 +1,106 @@
...
+      //   .string()
+      //   .nullish()
+      //   .describe("Database id to use. Defaults to `(default)` if unspecified."),
+      collectionPath: z
+        .string()
+        .describe(
+          "A collection path (e.g. `collectionName/` or `parentCollection/parentDocument/collectionName`)",
+        ),
+      filter: z.object({
+        where: z
```
fredzqm, line 24: So this will support only one filter and one orderBy?

fredzqm (commented):
Curious how it work in practice. Can you attach a screenshot of it?

src/mcp/tools/firestore/query_collection.ts
mbleigh: My original idea was that it would be `where: {[fieldName]: eqValue}` (you could do multiple field comparisons but all equals) but Gemini can't do dicts.

I'm actually mostly fine with this formulation, an LLM should be able to pick it up. I'd make it an array tho, not a single object:

```ts
{where: [{field: "foo", "operator": "EQUALS", value: "bar"}, ...]}
```

Also please make sure to test this against the Gemini API

nohe427: Ok, if we want the Where to be a filter, we would then check if `where.length === 1` and if it does, build a field filter. If its longer than `1` we would build a composite filter. Should the composite filter be an `AND` or an `OR` and how would we expose this to the user?

mbleigh: It's always an AND, we don't expose to the user, and I think you can just always do a composite filter with one or more entries in it (don't need to special case n=1)

nohe427: I'll kick the tires on it and see what I get

nohe427:
Screenshot 2025-05-12 at 2 56 28 PM

Preview of a simple query

src/mcp/tools/firestore/query_collection.ts
nohe427: In Gemini, I cannot get it to pick from a union type, I am kind of stuck on what value to provide here. I tried or(), any() and unknown() for the value field but Gemini doesn't seem to want to pick it up..

nohe427:
I cooked something up. The challenge was I was getting an error and its because I am missing a composite index on the table for compound queries. Figured out how to surface that error. Will push updated code soon

fredzqm:
> I cooked something up. The challenge was I was getting an error and its because I am missing a composite index on the table for compound queries. Figured out how to surface that error. Will push updated code soon

Good enough for now.
We can move fast and break things~

Composite index is always a pain point for Firestore. Yeah, it probably would only work well with a single field filter.

fredzqm (approved):

src/gcp/firestore.ts
```diff
@@ -164,6 +207,47 @@ export async function getDocuments(
...
+    });
+    const out: { documents: FirestoreDocument[] } = { documents: [] };
+    res.body.map((r) => {
+      if (r.document) {
+        out.documents.push(r.document);
+      }
+    });
+    return out;
+  } catch (err: FirebaseError | unknown) {
+    throw JSON.stringify(err);
```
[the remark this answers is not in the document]
nohe427: So the weird part is I need to convert the whole error into JSON to get the URL for building the composite index. Otherwise the MCP error is just a generic 400 error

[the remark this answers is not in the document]
nohe427: Done

PR #8551, opened 2025-05-12
Add backend for CloudAICompanion calls

hlshen:
Create service calls to CloudAiCompanion from FDC Extension


src/dataconnect/cloudAICompanionClient.ts
```diff
@@ -0,0 +1,100 @@
+import { ClientResponse, Client } from "../apiv2";
```
fredzqm, line 1: How does this differ from code here?

https://github.com/firebase/firebase-tools/blame/58f56134e0a3436caf834691594c923ae5ec70f1/src/gif/fdcExperience.ts

src/dataconnect/types.ts
```diff
@@ -249,3 +249,74 @@ interface ImpersonationUnauthenticated {
 export type Impersonation = ImpersonationAuthenticated | ImpersonationUnauthenticated;
 
 /** End Dataplane Client Types */
+
+/** Start Gemini client types */
+
+export interface CallCloudAiCompanionRequest {
```
fredzqm, line 255: Let's separate the FDC API's types and Cloud AI Companion (TAF)'s types.

They are quite different~

src/dataconnect/cloudAICompanionClient.ts
hlshen: It doesn't. Unfortunately had this code sitting around for awhile. Do you mind if we push the merge to post I/O?

src/dataconnect/types.ts
fredzqm: Yeah, would be nice in short term~

src/dataconnect/cloudAICompanionClient.ts
fredzqm: Happy to unblock you in short term~

Though let's clean them up soon. It's fine to check-in unused experimental codes to master~

Feature branch is useful when we want to hide a launch from public in the `FirebasePrivate` repos. If it's already public, don't think there are much values keeping it un-merged for long.

fredzqm (approved):

PR #8552, opened 2025-05-12
feat(mcp): Automatically detect features.

mbleigh:
This PR makes the default behavior (when no `--only` arg is supplied) be to only activate features that are detected in the current project environment. This relies first on `firebase.json` configs, but falls back to API activation checks in the current project.


fredzqm (approved):

fredzqm (commented):
No AI. Some thoughts.

After the meta init call, we will want to update the list of detected features.
This gets into the "dynamic" tool list problem we discussed before again.

PR #8553, opened 2025-05-12
feat: pin runtimes to latest supported version on functions init.

taeold:
To avoid surprises, we want to pin the runtime version associated with a functions codebase. e.g. Version of the runtime associated with a codebase shouldn't change when CLI version changes.

We'll start on this journey by changing our init template to explicitly pin the runtime on init.


ig-raiderler:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

inlined (approved):

PR #8554, opened 2025-05-13
[FDC init] Move the default value to actuate

fredzqm:
### Description

Move the default values from `askQuestions` to `actuate`.
This way the MCP tool can pass in an empty config and lets the `actuate` step populates the standard default value.

### Scenarios Tested

- In a Blaze project, create a FDC service in console
- Downgrade the project,
- Run `firebase init` locally, it is still able to finds the right service and pulls it down.
- `firebase deploy` requires billing as usually.


src/init/features/dataconnect/index.ts
```diff
@@ -266,24 +267,15 @@ function subConnectorYamlValues(replacementValues: { connectorId: string }): str
   return replaced;
 }
 
-async function promptForExistingServices(
-  setup: Setup,
-  info: RequiredInfo,
-  isBillingEnabled: boolean,
-): Promise<RequiredInfo> {
-  if (!setup.projectId || !isBillingEnabled) {
-    // TODO(b/368609569): Don't gate this behind billing once backend billing fix is rolled out.
```
fredzqm, line 275: Got rid of the `isBillingEnabled` check here because we no longer needs it.

FDC API can be used without billing.

rosalyntan (approved):

PR #8558, opened 2025-05-13
Adding messaging APIs and mcp tools that take advantange of it

gsiddh:
Demo: https://screencast.googleplex.com/cast/NDcyMTMxMzg3NTc1NTAwOHxiODY4MTk1MC01NA

### DIFF
--- a/src/mcp/tools/messaging/send_message_to_fcm_token.ts
+++ b/src/mcp/tools/messaging/send_message_to_fcm_token.ts
@@ -1,0 +1,33 @@
+import { z } from "zod";
+import { tool } from "../../tool.js";
+import { mcpError, toContent } from "../../util.js";
+import { sendMessageToFcmToken } from "../../../messaging/sendMessage.js";
+
+export const send_message_to_fcm_token = tool(
+  {
+    name: "send_message_to_fcm_token",
+    description: "Sends a message to FCM Token",
+    inputSchema: z.object({
+      fcmToken: z.string(),
+      title: z.string().optional(),
+      body: z.string().optional(),
+    }),
+    annotations: {
+      title: "Send message to FCM Token",
+      readOnlyHint: true,
+    },
+    _meta: {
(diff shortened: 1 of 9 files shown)


src/mcp/tools/messaging/send_message_to_fcm_token.ts
```diff
@@ -0,0 +1,33 @@
...
+      title: "Send message to FCM Token",
+      readOnlyHint: true,
+    },
+    _meta: {
+      requiresAuth: true,
+      requiresProject: true,
+    },
+  },
+  async ({ fcmToken, title, body }, { projectId }) => {
+    if (projectId === undefined) {
```
fredzqm, line 25: This isn't necessary. We have this check in the caller if `requiresProject: true,`

https://github.com/firebase/firebase-tools/blob/master/src/mcp/index.ts#L133

src/mcp/tools/messaging/send_message_to_fcm_topic.ts
```diff
@@ -0,0 +1,33 @@
...
+      title: "Send message to an FCM Topic",
+      readOnlyHint: true,
+    },
+    _meta: {
+      requiresAuth: true,
+      requiresProject: true,
+    },
+  },
+  async ({ topic, title, body }, { projectId }) => {
+    if (projectId === undefined) {
```
fredzqm, line 25: dito

fredzqm (approved):

src/mcp/tools/messaging/send_message_to_fcm_token.ts
gsiddh: Done.

src/mcp/tools/messaging/send_message_to_fcm_topic.ts
gsiddh: Done.

PR #8562, opened 2025-05-13
[MCP] firebase init dataconnect tool 

fredzqm:
Start with FDC, will add other products to the list gradually.

Screenshot 2025-05-13 at 10 17 57 AM

### DIFF
--- a/src/mcp/tools/core/index.ts
+++ b/src/mcp/tools/core/index.ts
@@ -5,6 +5,7 @@ import { get_project } from "./get_project.js";
 import { use_project } from "./use_project.js";
 import { get_sdk_config } from "./get_sdk_config.js";
 import { list_apps } from "./list_apps.js";
+import { init } from "./init.js";
 
 export const coreTools: ServerTool[] = [
   get_project,
@@ -12,4 +13,5 @@ export const coreTools: ServerTool[] = [
   list_apps,
   get_sdk_config,
   consult_assistant,
+  init,
 ];
--- a/src/mcp/tools/core/index.ts
+++ b/src/mcp/tools/core/index.ts
@@ -2,7 +2,14 @@ import type { ServerTool } from "../../tool.js";
 
 import { consult_assistant } from "./consult_assistant.js";
 import { get_project } from "./get_project.js";
+import { use_project } from "./use_project.js";
 import { get_sdk_config } from "./get_sdk_config.js";
 import { list_apps } from "./list_apps.js";
 
-export const coreTools: ServerTool[] = [get_project, list_apps, get_sdk_config, consult_assistant];
+export const coreTools: ServerTool[] = [
+  get_project,
+  use_project,
+  list_apps,
+  get_sdk_config,
+  consult_assistant,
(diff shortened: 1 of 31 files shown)


src/mcp/tools/core/init.ts
```diff
@@ -0,0 +1,75 @@
...
+              "The Firebase Data Connect service ID to initialize. Default to match the current folder name.",
+            ),
+          locationId: z
+            .string()
+            .default("us-central1")
+            .describe("The GCP region ID to set up the Firebase Data Connect service."),
+          cloudSqlInstanceId: z
+            .string()
+            .optional()
+            .describe("The GCP Cloud SQL instance ID to use in the Firebase Data Connect service."),
```
mbleigh, line 27: Provide an example and what happens if it's not supplied.

src/mcp/tools/core/init.ts
```diff
@@ -0,0 +1,75 @@
...
+          cloudSqlDatabase: z
+            .string()
+            .optional()
+            .describe("The Postgres database ID to use in the Firebase Data Connect service."),
```
mbleigh, line 31: Provide an example/default and what happens if it's not supplied.

src/mcp/tools/core/init.ts
```diff
@@ -0,0 +1,75 @@
+import { z } from "zod";
+import { tool } from "../../tool.js";
+import { toContent } from "../../util.js";
+import { actuate, Setup, SetupInfo } from "../../../init/index.js";
+
+export const init = tool(
+  {
+    name: "init",
+    description: "Initialize the Firebase Products. Provide the desired features information.",
```
mbleigh, line 9: Probably need a longer/more robust description.

src/mcp/tools/core/init.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    const setup: Setup = {
+      config: config?.src,
+      rcfile: rc?.data,
+      projectId: projectId,
+      features: [...featuresList],
+      featureInfo: featureInfo,
+    };
+    // Set force to true to avoid prompting the user for confirmation.
+    await actuate(setup, config, { force: true });
```
mbleigh, line 70: Is there a way we can make `force` a top-level tool param that defaults to false? Then catch conflicts and return an error describing the conflict and that `{force: true}` is needed to proceed?

src/mcp/tools/core/index.ts
```diff
@@ -5,11 +5,13 @@ import { get_project } from "./get_project.js";
...
 import { list_apps } from "./list_apps.js";
+import { init } from "./init.js";
 
 export const coreTools: ServerTool[] = [
   get_project,
   use_project,
   list_apps,
   get_sdk_config,
   consult_assistant,
+  init,
```
mbleigh, line 16: Are you confident in landing additional products in `init` before release cut on Thursday? I don't want to release init and it only works for one product, would rather leave commented out for now if so

mbleigh (approved):

PR #8564, opened 2025-05-13
[MCP] Add remaining "core" tools

TrCaM:
Tested with cursor that the tools are working


src/mcp/tools/core/create_android_app.ts
```diff
@@ -0,0 +1,35 @@
...
+  {
+    name: "create_android_app",
+    description: "Creates a new Android app in your Firebase project.",
+    inputSchema: z.object({
+      displayName: z.string().optional().describe("The user-friendly display name for your Android app."),
+      packageName: z.string().describe("The package name for your Android app (e.g., com.example.myapp)."),
+    }),
+    annotations: {
+      title: "Create Android App",
+      destructiveHint: false,
```
fredzqm, line 16: nit: `readOnlyHint: false,`

dito for all tools.

src/mcp/tools/core/create_android_app.ts
```diff
@@ -0,0 +1,35 @@
...
+  },
+  async ({ displayName, packageName }, { projectId }) => {
+    try {
+      const androidApp = await createAndroidApp(projectId!, {
+        displayName,
+        packageName,
+      });
+      
+      return toContent(androidApp);
+    } catch (error) {
```
fredzqm, line 31: This isn't necessary.

Any error throw will be caught and converted to `mcpError` [here](https://github.com/firebase/firebase-tools/blob/master/src/mcp/index.ts#L149)

dito for all tools.

fredzqm (changes requested):

src/mcp/tools/core/create_android_app.ts
TrCaM: Edit all except get_admin_sdk_config tool

src/mcp/tools/core/create_android_app.ts
TrCaM: Gotcha

fredzqm (approved):

src/mcp/tools/core/create_android_app.ts
```diff
@@ -0,0 +1,37 @@
+import { z } from "zod";
+import { tool } from "../../tool.js";
+import { toContent } from "../../util.js";
+import { createAndroidApp } from "../../../management/apps.js";
+
+export const create_android_app = tool(
+  {
+    name: "create_android_app",
```
mbleigh, line 8: I would recommend consolidating the three "create app" methods into a single method with:

```ts
create_app({
  display_name: z.string(),
  platform: z.enum(['web', 'ios', 'android']),
  android_config: z.object({package_name: z.string()}).optional(),
  ios_config: z.object({bundle_id: z.string(), app_store_id: z.string()}).optional()
})
```

TrCaM: Done. I also included the original message so that LLM is aware of the failure reason

mbleigh (approved):

TrCaM:
Just added some extra message prefix and suffix to hint LLM that they can fetch the sdk config after creating the app

PR #8566, opened 2025-05-13
Go look up compute service account instead of guessing

joehan:
### Description
Previously, we always assumed that the default compute service account was always ${projectNumber}-compute@developer.gserviceaccount.com - however, it is now configurable. Instead, we'll try to look it up, and fall back to the standard default if we cannot.


felansu:
A question, how i can specify the service account? 
can i specify in `firebase deploy --only extensions` command?

Thanks

MrLibya:
I think you missed 2 points: 

- `/src/gcp` line: `14`
-`/src/deploy/functions/ensure.ts` line:  `31`

Those 2 have hard-coded default service compute account.

Also the problem is still present, When running this command to deploy extensions:

```
firebase deploy --only extensions
```

Now it will read the correct default service account which is for me `compute-default@PROJECT_ID.iam.gserviceaccount.com`, But after deploy finish we will have the error saying it can't find default service account `number-compute@...`


```
- update user
; RESOURCE_ERROR at /deployments/firebase-ext-user/resources/extensionName: {"ResourceType":"gcp-types/cloudfunctions-v2beta:projects.locations.functions","ResourceErrorCode":"404","ResourceErrorMessage":{"code":404,"message":"Service account projects/-/serviceAccounts/PROJECT_NUMBER-compute@developer.gserviceaccount.com was not found.","status":"NOT_FOUND","statusMessage":"Not Found","requestPath":"https://cloudfunctions.googleapis.com/v2beta/projects/PROJECT_ID/locations/us-central1/functions","httpMethod":"POST"}}

```

joehan:
@MrLibya Thank you for catching those! Will fix.

For the issues you are still seeing - did you ever set a new Default Comput Service account on the project (ie did you run https://cloud.google.com/sdk/gcloud/reference/alpha/compute/project-info/set-default-service-account)? If not, this fix won't address the issue.

Apologies for the inconvenience here all - I'm still trying to wrap my head around the full scope of this issue, so its very possible that there are downstream issues that are breaking in the same way.

MrLibya:
@joehan Yes our new service account is set as default, And I confirm ur solution did half of the work, in debug I see the deploy process is catching up the new default service account, However after deploy the extension it have  function v2, That function is looking for the old default service account, I look into the code and I can't see where is that, Maybe is something on the cloud ?

joehan:
> @joehan Yes our new service account is set as default, And I confirm ur solution did half of the work, in debug I see the deploy process is catching up the new default service account, However after deploy the extension it have function v2, That function is looking for the old default service account, I look into the code and I can't see where is that, Maybe is something on the cloud ?

Yeah, that sounds like it will be a backend fix - I'll dig around and try to get it to the correct team

inlined (approved):

RyosukeNishi:
@joehan 
> Yeah, that sounds like it will be a backend fix - I'll dig around and try to get it to the correct team

Is there a public issue for the backend fix? I just want to make sure I can take notice when the fix is done.

MrLibya:
Any news on the backend side ? @joehan

PR #8569, opened 2025-05-14
[MPC] Add a way to print out a help markdown 

fredzqm:
After updating a MCP tool, folks can run this command and update the documentation table.

```
firebase experimental:mcp --help
```


src/mcp/tools/index.ts
```diff
@@ -26,19 +26,40 @@ export function availableTools(fixedRoot: boolean, activeFeatures?: ServerFeatur
...
+
+/**
+ * Generates a markdown table of all available tools and their descriptions.
+ * This is used for generating documentation.
+ */
+export function markdownDocsOfTools(): string {
+  const allTools = availableTools(false, []);
+  let doc = `
+| Tool Name                        | Feature Group | Description         |
+| -------------------------------- | ------------- | ------------------- |`;
```
mbleigh, line 55: If you're not actually going to consistently pad it in the content, may as well just drop the extra spaces.

src/mcp/tool.ts
```diff
@@ -15,6 +15,7 @@ export interface ServerToolContext {
 export interface ServerTool<InputSchema extends ZodTypeAny = ZodTypeAny> {
   mcp: {
     name: string;
+    productGroup?: string;
```
mbleigh, line 18: Please move this to `_meta`. The `mcp` field is meant to be exactly compatible with MCP protocol. Also call it `feature` not `productGroup`.

mbleigh (commented):
I'd kinda rather have this be emitted by an undocumented flag like `--generate-tool-list` instead of `--help`. I think `--help` should probably emit something closer to the MCP config docs that are emitted if you start it up in interactive mode.

src/mcp/tool.ts
fredzqm: Sure

fredzqm:
> I'd kinda rather have this be emitted by an undocumented flag like `--generate-tool-list` instead of `--help`. I think `--help` should probably emit something closer to the MCP config docs that are emitted if you start it up in interactive mode.

SG. Changed the name of the flag.

Wasn't sure either~

src/mcp/README.md
```diff
@@ -34,26 +34,43 @@ In `.cursor/mcp.json` in your workspace directory, add the Firebase MCP server:
...
-| `dataconnect_get_connector`      | `dataconnect` | Get the Firebase Data Connect Connectors in the project, which includes the pre-defined GraphQL queries accessible to client SDKs.                                                                             |
-| `storage_get_rules`              | `storage`     | Retrieves the Firebase Cloud Storage Rules for the default bucket.                                                                                                                                             |
-| `send_message_to_fcm_token`      | `messaging`   | Sends a push message to an FCM Token.                                                                                                                                                                          |
-| `send_message_to_fcm_topic`      | `messaging`   | Sends a push message to an FCM Topic.                                                                                                                                                                          |
+Generate the  below table using the following command:
+
+```bash
+firebase experimental:mcp --generate-tool-docs
+```
+
```
mbleigh, line 42:
```suggestion
```

We don't need to document how to do this - it should be for our own internal use.

mbleigh (approved):

src/mcp/README.md
mbleigh: How about `CONTRIBUTING.md` in the mcp directory since that's where contribution guides go for GH generally.

PR #8576, opened 2025-05-14
Moving downloadable emulator details to JSON, adding sha256 checksum

joehan:
### Description
Firebase Studio wants to predownload emulators, and this change will make it simpler to do so:
- Move the emulator download details to a JSON file so they can be parsed more easily
- Add a SHA256 checksum (since nix has deprecated MD5). This is more secure than MD5 anyway, so I'm gonna switch the CLI to use it as well (and get rid of MD5) in an upcoming release - holding off on that for now to derisk the change before IO.

I generated these checksums using the following command on macos:
```
shasum -a 256 <fileName>
```

### Scenarios Tested
Tested that the emulators still start up correctly, download if missing, and use the cache otherwise.


src/emulator/types.ts
```diff
@@ -185,6 +185,7 @@ export interface EmulatorUpdateDetails {
   version: string;
   expectedSize: number;
   expectedChecksum: string;
+  expectedChecksumSHA256: string; // TODO: Use this for validation within the CLI as well.
```
maneesht, line 188: This doesn't match the field in the JSON file

src/emulator/downloadableEmulatorInfo.json
```diff
@@ -0,0 +1,60 @@
+{
```
joehan, line 1: Ahhhh, i accidentally nuked this file somehow - fixing

christhompsongoogle:
Is the idea to switch to json completely and remove the js definitions? If so SGTM - if not we have a double source of truth problem. Also we'll have to adjust some documentation on emulator releases which is a small lift

joehan:
> Is the idea to switch to json completely and remove the js definitions? If so SGTM - if not we have a double source of truth problem. Also we'll have to adjust some documentation on emulator releases which is a small lift

Yeah, that is the intent - all the JS objects in downloadableEmulators.ts are taking the version number, checksums, etc from this JSON file.

And ack on emulator release docs - I'll take a pass thru moma and update the ones I find. I'm also gonna CC folks who I know release emualtors on this

src/emulator/downloadableEmulatorInfo.json
```diff
@@ -0,0 +1,60 @@
...
+    }
+  },
+  "pubsub": {
+    "version": "0.8.14",
+    "expectedSize": 66786933,
+    "expectedChecksum": "a9025b3e53fdeafd2969ccb3ba1e1d38",
+    "expectedChecksumSHA256": "4fbeefd8ecb32b10e5ab522e5d8748e0c2b13891c471c0c327c04632dcc75a8d"
+  },
+  "dataconnect": {
+    "darwin": {
```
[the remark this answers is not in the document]
joehan: Yeah, this format is expected to stay the same - we need to do special things for dataconnect because Golang can't easily compile to cross platform binaries. I think the safest way to parse is to assume that the existing shape stays in place, but allow for extra keys at the top level/inside UI or dataconnect.

maneesht (approved):

PR #8578, opened 2025-05-14
chore(mcp): Consolidate FCM tools into a single send_message tool.

mbleigh:


fredzqm:
Looks fine. Will let @gsiddh take the final pass~

tl;dr; There is a limit on number of tools we can register, so smaller number of versatile tools is preferred.

src/messaging/interfaces.ts
```diff
@@ -28,4 +28,6 @@ export interface Notification {
    * The notification body
    */
   body?: string;
+  /** URL of an image to include in the notification. */
+  image?: string;
```
gsiddh, line 32: we can leave the image out for the first pass.

src/messaging/sendMessage.ts
```diff
@@ -16,23 +16,32 @@ const apiClient = new Client({
...
+      title: options.title,
+      body: options.body,
+      image: options.image,
     };
-    const message: TokenMessage = {
-      token: fcmToken,
+    if (!options.token && !options.topic) {
+      throw new FirebaseError("Must supply either token or topic to send FCM message.");
+    }
+    const message: Message = {
```
gsiddh, line 42: not too familiar with typescript - would it use the appropriate underlying type TopicMessage and TokenMessage?

src/messaging/interfaces.ts
mbleigh: Why leave it out? Is there a problem with including?

src/messaging/sendMessage.ts
mbleigh: Message is already the union type, so it's appropriate to use here.

src/messaging/interfaces.ts
gsiddh: we can keep it. no issues. Was just reducing testing scope.

gsiddh (approved):

PR #8579, opened 2025-05-14
[MCP] Starter Implementation of dataconnect emulator support

tammam-g:
### Description

Adds a `useEmulator` option to dataconnect tools.

### Scenarios Tested

QraphqQL Execute
QraphqQL Execute Readonly
Mutation Execute
Query Execute

### Things that don't work
- Auth (USER, NO_ACCESS) gated mutations/queries don't work for both prod and emulator.
- Some clients couldn't find the emulator, I was able to pinpoint that the issue is because of default temp directory results   here (https://github.com/firebase/firebase-tools/blob/12f569e63913973484f57c5a636169b0f5b2ed45/src/emulator/hub.ts#L67). It seems like some clients are overriding the environment variables which is where this sourced from. We need to either do a fallback to expected port/host or find a more robust way for defining a global locator path.

### Sample Commands
Screenshot 2025-05-14 at 5 36 11 PM
image


src/mcp/index.ts
```diff
@@ -153,6 +169,7 @@ export class FirebaseMcpServer {
       config: Config.load(options, true) || new Config({}, options),
       rc: loadRC(options),
       accountEmail,
+      emulatorHubClient: await this.getEmulatorHubClient(),
```
fredzqm, line 172: `host: this` is already passed into the `ServerToolContext`.

Tools can access emulator hub via `getEmulatorHubClient`. Don't think we need to pass it into `ServerToolContext` again.

src/mcp/tools/dataconnect/execute_graphql.ts
```diff
@@ -23,6 +26,7 @@ export const execute_graphql = tool(
         .describe(
           "A stringified JSON object containing variables for the operation. MUST be valid JSON.",
         ),
+      useEmulator: z.boolean().optional().describe("Target the DataConnect emulator if true."),
```
fredzqm, line 29: `default(false)`

src/mcp/tools/dataconnect/execute_graphql.ts
```diff
@@ -33,13 +37,40 @@ export const execute_graphql = tool(
       requiresAuth: true,
     },
   },
-  async ({ query, serviceId, variables: unparsedVariables }, { projectId, config }) => {
+  async (
+    { query, serviceId, variables: unparsedVariables, useEmulator },
+    { projectId, config, emulatorHubClient },
+  ) => {
+    if (!projectId) {
```
fredzqm, line 44: This is already checked here.

https://github.com/firebase/firebase-tools/blob/c64476f2a372d1d61c0fde10a0cf623ddd59b99f/src/mcp/index.ts#L147

src/mcp/tools/dataconnect/execute_graphql.ts
```diff
@@ -33,13 +37,40 @@ export const execute_graphql = tool(
...
+        );
+      }
+      apiClient = new Client({
+        urlPrefix: emulatorDetails.url,
+        apiVersion: dataplane.DATACONNECT_API_VERSION,
+        auth: false,
+      });
+    } else {
+      apiClient = dataplane.dataconnectDataplaneClient();
+    }
```
fredzqm, line 67: Could you extract this block into a helper method?

They are shared across 4 tools.

fredzqm (changes requested):

src/mcp/emulator/dataconnectEmulatorController.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  try {
+    const emulators = await hubClient.getEmulators();
+    const dcEmulatorInfo = emulators[Emulators.DATACONNECT];
+
+    if (dcEmulatorInfo) {
+      return formatEndpoint(dcEmulatorInfo);
+    }
+  } catch (error) {
+    console.warn(
```
fredzqm, line 34: `console.warn` won't work well in MCP server.

cc @jonathanedey Do you know how to produce a good debugging logs in the MCP server?

src/mcp/emulator/dataconnectEmulatorController.ts
```diff
@@ -0,0 +1,40 @@
...
+  const host = emulatorInfo.host.includes(":") ? `[${emulatorInfo.host}]` : emulatorInfo.host;
+  return {
+    host: emulatorInfo.host,
+    port: emulatorInfo.port,
+    url: `http://${host}:${emulatorInfo.port}`,
+  };
+}
+
+export async function getDataConnectEmulatorDetails(
+  hubClient?: EmulatorHubClient,
```
fredzqm, line 20: This looks quite complicated. Can we have a simple helper?

```
export async function dataconnectEmulatorClient(hubClient?: EmulatorHubClient): Promise<Client> {
  // Call hub to figure out the port
  // Construct the API client based on info.
}
```

Upon any errors, just throw them out. Let the caller (LLM) know that the emulator hub isn't running or the just dataconnect emulator isn't running.

fredzqm (changes requested):

src/mcp/emulator/dataconnectEmulatorController.ts
mbleigh: Use the `logger` - when running in `--debug` mode it should be capturing things in `firebase-debug.log`.

src/mcp/index.ts
tammam-g: Done

src/mcp/tools/dataconnect/execute_graphql.ts
tammam-g: Done

src/mcp/emulator/dataconnectEmulatorController.ts
tammam-g: Done

fredzqm (approved):

PR #8580, opened 2025-05-14
Provision App Hosting compute service account during init flow

blidd-google:
We must provision the App Hosting compute service account before calling `createBackend`.


src/commands/init.ts
```diff
@@ -48,7 +48,7 @@ let choices: {
   },
   {
     value: "apphosting",
-    name: "App Hosting: Configure an apphosting.yaml file for App Hosting",
+    name: "App Hosting: Enable web app deployments with App Hosting",
```
kmcnellis, line 51: This seems wrong to me? are we enabling deployments, or making a backend?  I think the flow as-written doesn't link to github?

src/init/features/apphosting.ts
```diff
@@ -31,6 +34,26 @@ export async function doSetup(setup: Setup, config: Config): Promise<void> {
       "Firebase App Hosting requires billing to be enabled on your project. Please enable billing by following the steps at https://cloud.google.com/billing/docs/how-to/modify-project",
     );
   }
+  await ensureApiEnabled({ projectId });
+  await ensureRequiredApisEnabled(projectId);
+  // N.B. Deploying a backend from source requires the App Hosting compute service
```
kmcnellis, line 39: Are we actually attaching the the storage.objectViewer role right now?
Also, even if we were doing that, this comment is about the storage.objectViewer role, but its on the block of code that actually provisions the SA, and we'd need to do that even if the use isn't using deploy from source

src/init/features/apphosting.ts
kmcnellis, line 1: In     
```const backend = await createBackend(
      projectId,
      location,
      backendId,
      /* serviceAccount= */ null,
      /* repository= */ undefined,
      webApp?.id,
    );
```
It's a real trap right now to make a backend without attaching a git repo (or doing deploy from source), since AFAIK we don't have a command to add one later

kmcnellis (changes requested):

src/commands/init.ts
blidd-google: cc @jpreid7396 my understanding is that the description is meant to be generic enough to cover set up for deploy from source, but also any other. I'll defer to Julia here about the language specifics

src/init/features/apphosting.ts
blidd-google: Yeah, the storage.objectViewer role change isn't relevant until the follow-up deploy from source PR https://github.com/firebase/firebase-tools/pull/8516. Removing the comment.

kmcnellis (approved):

PR #8582, opened 2025-05-14
Getting autoinit working with the App Hosting emulator

jamesdaniels:
## What's new

Pretty simple one, wire up `FIREBASE_CONFIG`, `FIREBASE_WEBAPP_CONFIG`, and a few other important environment variables to the App Hosting emulator to support SDK autoinit—from there trip `@firebase/utils` postinstall step, if present. This is intended to fail open and rely on the JS SDK error messages, if the developer is relying on autoinit.

The only "meh" bit is cross referencing the apphosting emulator config with the new `.apphosting` config entry, to look up the `backendId`. We should consider aligning these more in the future.

## Cases tested

* Missing/invalid project id
* Missing/invalid backend id
* Multiple apphosting firebase.json entries
* No network access
* Multiple versions of Firebase
* Versions of Firebase that don't have the postinstall step

## Known issues

* PNPM still has issues, logging a warning for now
* Sometimes the NextJS build cache can lead to old values being picked up—the developer can work around this with with an `rm -rf .next/cache`

I will document these limitations. IMO the best solution to both of these is to use autoinit via Cookie—like Web Frameworks does—but that will take a bit more work as we'll need to implement a proxy, I will address post-I/O.

## Discovered issues

While working on this feature I discovered a couple issues with the App Hosting emulator, namely that the child process isn't always cleanly terminated and never is when using `emulators:exec`. I'll file bugs and if other folk don't have bandwidth I can work on that while addressing the above *known issues*


src/emulator/apphosting/serve.ts
```diff
@@ -167,3 +201,44 @@ export function getEmulatorEnvs(): Record<string, string> {
...
+async function getBackendAppConfig(
+  projectId: string,
+  backendId: string,
+): Promise<WebConfig | undefined> {
+  if (Constants.isDemoProject(projectId)) {
+    return constructDefaultWebSetup(projectId);
+  }
+
+  const backendsList = await apphosting.listBackends(projectId, "-").catch(() => undefined);
+  const backend = backendsList?.backends.find((b) => b.name.split("/").at(-1) === backendId);
```
jamesdaniels, line 232: Does `.at(-1)` work in all the supported runtimes, consider using `pop()` instead

src/emulator/apphosting/serve.ts
```diff
@@ -167,3 +201,44 @@ export function getEmulatorEnvs(): Record<string, string> {
...
+  try {
+    firebaseUtilPath = dirname(
+      require.resolve("@firebase/util/package.json", { paths: [rootDirectory] }),
+    );
+  } catch (err: unknown) {
+    return;
+  }
+
+  await spawnWithCommandString("npm run postinstall", firebaseUtilPath, env).catch((err) => {
+    logWarning(`Failed to run postinstall script for @firebase/util: ${getError(err).message}.`);
```
jamesdaniels, line 219: Error message does not explain what the issue here might be, expand or skip

src/emulator/apphosting/serve.ts
```diff
@@ -167,3 +201,44 @@ export function getEmulatorEnvs(): Record<string, string> {
...
+): Promise<WebConfig | undefined> {
+  if (Constants.isDemoProject(projectId)) {
+    return constructDefaultWebSetup(projectId);
+  }
+
+  const backendsList = await apphosting.listBackends(projectId, "-").catch(() => undefined);
+  const backend = backendsList?.backends.find((b) => b.name.split("/").at(-1) === backendId);
+
+  if (!backend) {
+    logWarning(`Unable to lookup details for backend ${backendId}.`);
```
jamesdaniels, line 235: Explain what the problem is

src/emulator/controller.ts
```diff
@@ -968,11 +970,28 @@ export async function startAll(
...
    */
-  const apphostingConfig = options.config.src.emulators?.[Emulators.APPHOSTING];
+  const apphostingEmulatorConfig = options.config.src.emulators?.[Emulators.APPHOSTING];
 
   if (listenForEmulator.apphosting) {
+    const rootDirectory = apphostingEmulatorConfig?.rootDirectory;
+    const backendRoot = resolveProjectPath({}, rootDirectory ?? "./");
+
+    // It doesn't seem as though App Hosting emulator supports multiple backends, infer the correct one
+    // from the root directory.
```
jamesdaniels, line 980: File an issue saying we should add backendId to the emulators.apphosting config and add TODO

src/emulator/apphosting/serve.ts
```diff
@@ -118,8 +132,28 @@ async function serve(
...
+      environmentVariablesToInject["FIREBASE_WEBAPP_CONFIG"] ||= JSON.stringify(webappConfig);
+      environmentVariablesToInject["FIREBASE_CONFIG"] ||= JSON.stringify({
+        databaseURL: webappConfig.databaseURL,
+        storageBucket: webappConfig.storageBucket,
+        projectId: webappConfig.projectId,
+      });
+      await tripFirebasePostinstall(backendRoot, environmentVariablesToInject);
+    }
+  } else {
+    logWarning("No projectId or backendId provided. Skipping Firebase WebApp config.");
```
jamesdaniels, line 154: Explain the repercussions of this or skip the error entirely

src/emulator/apphosting/serve.ts
```diff
@@ -167,3 +201,44 @@ export function getEmulatorEnvs(): Record<string, string> {
...
 }
+
+async function tripFirebasePostinstall(
+  rootDirectory: string,
+  env: Record<string, string>,
+): Promise<void> {
+  let firebaseUtilPath;
+  try {
+    firebaseUtilPath = dirname(
+      require.resolve("@firebase/util/package.json", { paths: [rootDirectory] }),
```
jamesdaniels, line 212: If for whatever reason there is more than one version of `@firebase/util` this is non-deterministic on which one it patches. Should patch all of them.

src/emulator/apphosting/serve.ts
```diff
@@ -115,11 +129,34 @@ async function serve(
     value.value ? value.value : await loadSecret(projectId, value.secret!),
   ]);
 
-  const environmentVariablesToInject = {
+  const environmentVariablesToInject: NodeJS.ProcessEnv = {
     ...getEmulatorEnvs(),
     ...Object.fromEntries(await Promise.all(resolveEnv)),
+    NODE_ENV: process.env.NODE_ENV,
```
jamesdaniels, line 135: move to the top of the list so the yaml files override them

src/emulator/apphosting/serve.ts
```diff
@@ -167,3 +204,83 @@ export function getEmulatorEnvs(): Record<string, string> {
...
+  dependencies?: Record<string, Dependency>;
+};
+
+async function tripFirebasePostinstall(
+  rootDirectory: string,
+  env: NodeJS.ProcessEnv,
+): Promise<void> {
+  const npmLs = spawnSync("npm", ["ls", "@firebase/util", "--json", "--long"], {
+    cwd: rootDirectory,
+    shell: process.platform === "win32",
```
blidd-google, line 221: We only want to run `npm ls` in a shell if the user is on windows?

src/emulator/apphosting/serve.ts
```diff
@@ -167,3 +204,83 @@ export function getEmulatorEnvs(): Record<string, string> {
...
+  if (Constants.isDemoProject(projectId)) {
+    return constructDefaultWebSetup(projectId);
+  }
+
+  if (!backendId) {
+    return undefined;
+  }
+
+  const backendsList = await apphosting.listBackends(projectId, "-").catch(() => undefined);
+  const backend = backendsList?.backends.find((b) => b.name.split("/").pop() === backendId);
```
blidd-google, line 271: we have a utility for parsing the backend ID from the full resource name called `parseBackendName` in src/apphosting/backend.ts

src/emulator/controller.ts
```diff
@@ -968,11 +970,28 @@ export async function startAll(
    * app hosting emulator may depend on other emulators (i.e auth, firestore,
    * storage, etc).
    */
-  const apphostingConfig = options.config.src.emulators?.[Emulators.APPHOSTING];
+  const apphostingEmulatorConfig = options.config.src.emulators?.[Emulators.APPHOSTING];
 
   if (listenForEmulator.apphosting) {
+    const rootDirectory = apphostingEmulatorConfig?.rootDirectory;
```
blidd-google, line 976: How does the root directory here relate to the root directory specified per-backend by the user in firebase.json?

blidd-google (approved):
left some comments / nits but otherwise LGTM

src/emulator/controller.ts
jamesdaniels: I think we need to put some thought into that, the main difference IMO is that emulatorConfig rootDir is relative to the json. I think my matcher works well enough for now but as `firebase init apphosting` is new I might be missing some cases. Will play around.

PR #8584, opened 2025-05-15
[MCP] Support init for database

fredzqm:
Mostly refactors to make existing RTDB init flow fit into the MCP flow.

Init from an empty folder

image

`firebase init database`
Screenshot 2025-05-14 at 5 08 42 PM


src/mcp/tools/core/init.ts
```diff
@@ -15,6 +16,16 @@ export const init = tool(
...
         // TODO: Add all the features here.
+        database: z.object({
+          rulesFilename: z
+            .string()
+            .default("database.rules.json")
+            .describe("The file to use for Realtime Database Security Rules."),
+          rules: z
+            .string()
+            .default(DEFAULT_RULES)
+            .describe("The security rules to use for Realtime Database Security Rules."),
```
mbleigh, line 27: Make both of these `.optional()` in addition to having defaults defined.

mbleigh (approved):
It's a little lopsided to have this as our second `init` when we don't otherwise have RTDB tools, but 🤷 . Let's target Firestore next.

src/mcp/tools/core/init.ts
fredzqm: Sure

fredzqm:
> It's a little lopsided to have this as our second `init` when we don't otherwise have RTDB tools, but 🤷 . Let's target Firestore next.

Only doing it because I already got familiar with its init code~ Firestore and storage will come very soon.

PR #8585, opened 2025-05-15
feat(mcp): Adds validate_rules tools.

mbleigh:
Refactors rules to be common functions relied on for both Firestore and Storage. Adds `validate_rules` tools with nice formatting.

image


src/mcp/tools/rules/validate_rules.ts
```diff
@@ -0,0 +1,137 @@
...
+      } else {
+        // If neither source nor source_file is provided, default to empty string for validation.
+        // testRuleset might still return general errors not tied to specific lines.
+        rulesSourceContent = "";
+      }
+
+      const result = await testRuleset(projectId!, [
+        // The name "firestore.rules" is a convention for testRuleset,
+        // actual fileName from issues will be used in formatting.
+        { name: "firestore.rules", content: rulesSourceContent },
```
fredzqm, line 123: If this name don't matter, can we do something generic like `test.rules`?

Right now, it looks like a bug if this is used to test `storage.rules`.

fredzqm (approved):

PR #8588, opened 2025-05-15
[MCP] Init Firestore

fredzqm:


mbleigh (approved):
Have you tested this out much in an actual client? How well does it work?

fredzqm:
> Have you tested this out much in an actual client? How well does it work?

It works in the MCP inspector.

I tested it in cursor. It ran into the above stdout spam issue. Will try to debug it.

PR #8594, opened 2025-05-15
chore(mcp): remove GiF tools for now until ToS is worked out

mbleigh:
We're not quite ready to add these tools since there are some questions around ToS and usage.


fredzqm:
What's the concern here?

Those tools are top of the listed and provides the most values.

fredzqm (approved):

fredzqm:
Let's wait until the discussion with Thevi concludes~

PR #8599, opened 2025-05-15
chore(mcp): Refactor client directory config logic.

mbleigh:
This makes it so that project directory is stored on a per-client basis instead of globally, and also provides better instructions in `get_environment` for what happens if there is no active environment.

### DIFF
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -73,7 +73,7 @@ export class FirebaseMcpServer {
   }
 
   private get clientConfigKey() {
-    return `mcp.clientConfigs.${this.clientInfo?.name || "<unknown-client>"}`;
+    return `mcp.clientConfigs.${this.clientInfo?.name || "<unknown-client>"}${this.optionsRoot ? `:${this.optionsRoot}` : ""}`;
   }
 
   getStoredClientConfig(): ClientConfig {
@@ -91,10 +91,6 @@ export class FirebaseMcpServer {
     await this.ready();
     if (this.cachedProjectRoot) return this.cachedProjectRoot;
     const storedRoot = this.getStoredClientConfig().projectRoot;
-    if (storedRoot) {
-      this.cachedProjectRoot = storedRoot;
-      return storedRoot;
-    }
     this.cachedProjectRoot =
       storedRoot || this.optionsRoot || process.env.PROJECT_ROOT || process.cwd();
     return this.cachedProjectRoot;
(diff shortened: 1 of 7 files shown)


src/mcp/index.ts
```diff
@@ -50,15 +53,53 @@ export class FirebaseMcpServer {
...
+  }
+
+  async detectProjectRoot(): Promise<string> {
+    await this.ready();
+    if (this.cachedProjectRoot) return this.cachedProjectRoot;
+    const storedRoot = this.getStoredClientConfig().projectRoot;
+    if (storedRoot) {
+      this.cachedProjectRoot = storedRoot;
+      return storedRoot;
+    }
```
fredzqm, line 97: nit: the above if branch looks redundant. Already taken care of below

src/mcp/index.ts
```diff
@@ -8,7 +8,7 @@ import {
   ListToolsResult,
 } from "@modelcontextprotocol/sdk/types.js";
 import { checkFeatureActive, mcpError } from "./util.js";
-import { SERVER_FEATURES, ServerFeature } from "./types.js";
+import { ClientConfig, SERVER_FEATURES, ServerFeature } from "./types.js";
 import { availableTools } from "./tools/index.js";
 import { ServerTool, ServerToolContext } from "./tool.js";
 import { configstore } from "../configstore.js";
```
fredzqm, line 14: Not very familiar with this. Likely a dumb question

What is config store? Is it backed by a local file somewhere?

mbleigh: Yeah. All device-level Firebase config is stored there. It's in `~/.config/configstore/firebase-tools.json`

fredzqm (approved):

mbleigh:
> When you say per-client basis?
> 
> If I open two cursor windows and two VS Code with GitHub co-pilot, are there 2 project dir or 4 project dir cached?

I just pushed some tweaks that update the behavior to key the root by any of `--dir`, `PROJECT_ROOT`, or `cwd()`. So this means the only time project roots will be shared between different workspaces is when the MCP Client has multiple instances that both start in the same working directory with no `--dir` or `PROJECT_ROOT`.

Unfortunately, I think this is what Cline does today 😦  but it should cover most other clients pretty consistently without surprises.

src/mcp/index.ts
mbleigh: You're right!

PR #8601, opened 2025-05-15
Save file before execution

hlshen:
Save the active text editor before running an execution.

Tested locally.


firebase-vscode/src/data-connect/execution/execution.ts
```diff
@@ -333,6 +333,7 @@ export function registerExecution(
             ? DATA_CONNECT_EVENT_NAME.RUN_LOCAL
             : DATA_CONNECT_EVENT_NAME.RUN_PROD,
         );
+        vscode.window.activeTextEditor?.document.save();
```
yuchenshi, line 336:
```suggestion
        await vscode.window.activeTextEditor?.document.save();
```

I think https://code.visualstudio.com/api/references/vscode-api#TextDocument

yuchenshi (approved):

PR #8604, opened 2025-05-15
Make emulator hub locator path more robust to environment changes

tammam-g:
### Description

Some clients couldn't discover emulator due to tempdir path being different depending the client, by using ~/.cache/firebase we should have more robust path.
 
### Scenarios Tested
Running dataconnect emulator
Running dataconnect MCP and communicating with emulator from previously failing client (RooCode)

### Sample Commands

image


joehan (changes requested):
Does this file get cleaned up after you kill the emulator? If not, I'm worried that all subsequent runs will see a running emualtor

tammam-g:
> Does this file get cleaned up after you kill the emulator? If not, I'm worried that all subsequent runs will see a running emualtor

Just tested it and it does cleanup, works same exact way as the previous one worked.

fredzqm (approved):

fredzqm:
> Does this file get cleaned up after you kill the emulator? If not, I'm worried that all subsequent runs will see a running emualtor

The `EmulatorHubClient` is just a HTTP client that tries to reach the emulator.

Not something that consumes a lot of resource, right?

fredzqm (changes requested):
(offline discussion)

This one looks risky

fredzqm:
@joehan @yuchenshi

Can you take a look at this small but risky change?

It sounds like some MCP client sets the `$TMPDIR` and messed up CLI's resolution algorithm. @tammam-g  is trying to move the locator file to a fixed path.

src/emulator/hub.ts
```diff
@@ -64,9 +64,7 @@ export class EmulatorHub extends ExpressBasedEmulator {
   }
 
   static getLocatorFilePath(projectId: string): string {
-    const dir = os.tmpdir();
-    const filename = `hub-${projectId}.json`;
-    return path.join(dir, filename);
+    return path.join(os.homedir(), ".cache", "firebase", `hub-${projectId}.json`);
```
yuchenshi, line 67: What if the parent directories do not exist? Can we find some existing code in the rest of the CLI code base that creates the parents and reuse the logic?

PR #8605, opened 2025-05-15
initial apphosting mcp tool

bkendall:
### Description

Adds an initial MCP tool for App Hosting.

This combines the Backend and it's singleton Traffic resource into one so that the LLM can make some inferences about the traffic information!

It also adds a tool so that it can fetch build or service logs!


src/mcp/tools/apphosting/list_backends.ts
```diff
@@ -0,0 +1,46 @@
...
+    _meta: {
+      requiresAuth: true,
+      requiresProject: true,
+    },
+  },
+  async ({ location } = {}, { projectId }) => {
+    if (!projectId) return NO_PROJECT_ERROR;
+    if (!location) location = "-";
+    const data: (Backend & { traffic: Traffic })[] = [];
+    const backends = await listBackends(projectId, location);
```
joehan, line 38: We might want a different message for no backends - ie 
```suggestion
    const backends = await listBackends(projectId, location);
    if (!backends.length) return toContent("No backends exist on project ${projectId} ${location !== "-" ? `in ${location}` :""}
```
Not sure what the LLMs will like more

joehan (approved):

src/mcp/tools/apphosting/list_backends.ts
```diff
@@ -0,0 +1,46 @@
...
+      title: "List App Hosting backends.",
+      readOnlyHint: true,
+    },
+    _meta: {
+      requiresAuth: true,
+      requiresProject: true,
+    },
+  },
+  async ({ location } = {}, { projectId }) => {
+    if (!projectId) return NO_PROJECT_ERROR;
```
fredzqm, line 35: This is already taken care of by the caller.

https://github.com/firebase/firebase-tools/blob/214d190ecd98844c1cb90d22afb51f584854775c/src/mcp/index.ts#L189

src/mcp/tools/apphosting/list_backends.ts
```diff
@@ -0,0 +1,46 @@
...
+  {
+    name: "list_backends",
+    description:
+      "Retrieves a list of App Hosting backends in the current project. The `uri` is the public URL of the backend.",
+    inputSchema: z.object({
+      location: z
+        .string()
+        .optional()
+        .default("-")
+        .describe("Limit the listed backends to this region."),
```
fredzqm, line 23: Maybe explain to LLM that by default it would list all regions

fredzqm (approved):

src/mcp/tools/apphosting/list_backends.ts
bkendall: Sure, but I still have to assert that projectId isn't undefined, otherwise I have to do checks all over the place. Does this hurt that much to bother?

src/mcp/tools/apphosting/list_backends.ts
bkendall: Done.

src/mcp/tools/apphosting/list_backends.ts
fredzqm: Yeah, it's a bit annoying that `projectId` is nullable even if `requiresProject` is true above.

Maybe we can do a clean up to supply a dummy `projectId: "missing-project"` if there there is no project present.

fredzqm (changes requested):
Looks good to me

I'd defer to @joehan on when the code freeze ends and we are ready to adding more tools~

src/mcp/tools/apphosting/list_backends.ts
bkendall: I'll go with it, sure. Though it makes me feel weird that it doesn't understand that an empty list represents this state.

fredzqm (approved):
LG since we are exiting code freeze.

src/mcp/tools/run/fetch_logs.ts
```diff
@@ -0,0 +1,30 @@
...
+import { tool } from "../../tool.js";
+import { toContent } from "../../util.js";
+import { NO_PROJECT_ERROR } from "../../errors.js";
+import { fetchServiceLogs } from "../../../gcp/run.js";
+
+export const fetch_logs = tool(
+  {
+    name: "fetch_logs",
+    description:
+      "Fetches recent logs for a Cloud Run service. Includes details such as the message, severity, and timestamp.",
```
joehan, line 11: Probably include something about how you can use this to see app hosting server side logs, so that LLMs know to call it when the user asks something like "Show me the logs for my app hosting backend"

joehan (approved):
Code LGTM - but should we enable this tool when you enable app hosting tools too?

src/mcp/tools/apphosting/fetch_logs.ts
```diff
@@ -0,0 +1,86 @@
+import { z } from "zod";
+import { tool } from "../../tool.js";
+import { toContent } from "../../util.js";
+import {
+  Backend,
+  getBackend,
+  getTraffic,
+  listBuilds,
+  parseBackendName,
```
joehan, line 9: Remove

src/mcp/tools/apphosting/fetch_logs.ts
```diff
@@ -0,0 +1,86 @@
...
+    },
+  },
+  async ({ buildLogs, backendId, location } = {}, { projectId }) => {
+    projectId ||= "";
+    location ||= "";
+    if (!backendId) {
+      return toContent(`backendId must be specified.`);
+    }
+    const backend = await getBackend(projectId, location, backendId);
+    // if (location === "-") location = parseBackendName(backend.name).location;
```
joehan, line 52: Is this needed, or can it be removed?

src/mcp/tools/apphosting/fetch_logs.ts
```diff
@@ -0,0 +1,86 @@
...
+    const traffic = await getTraffic(projectId, location, backendId);
+    const data: Backend & { traffic: Traffic } = { ...backend, traffic };
+
+    if (buildLogs) {
+      const builds = await listBuilds(projectId, location, backendId);
+      // builds.builds.sort((a, b) => a.c)
```
joehan, line 58: Remove?

joehan (approved):
LGTM after cleanup

PR #8606, opened 2025-05-15
[MCP] Require project root

fredzqm:


src/mcp/index.ts
```diff
@@ -185,6 +185,9 @@ export class FirebaseMcpServer {
 
     const projectId = await this.getProjectId();
     const accountEmail = await this.getAuthenticatedUser();
+    if (!this.cachedProjectRoot) {
+      return mcpError("No project root set. Please set the project root before calling tools.");
```
mbleigh, line 189:
```suggestion
      return mcpError("No project root set. Use the `firebase_update_environment` tool to set a directory or add a `--dir <directory>` option to the Firebase MCP Server command line flags.");
```

mbleigh (approved):
FWIW I think with my earlier PR today this should basically never happen, but better safe than sorry.

fredzqm:
I think I ran into this after your change got in~ Miew, edge cases.

PR #8610, opened 2025-05-16
[MCP] Create list_projects tool + slight modifications to list_apps #2

TrCaM:
Create list_projects tool + slight modifications to list_apps

For some reason cursor keep breaking the tool without the change. It works much more consistently after removing the optional() part

### DIFF
--- a/src/mcp/tools/core/list_apps.ts
+++ b/src/mcp/tools/core/list_apps.ts
@@ -9,8 +9,8 @@ export const list_apps = tool(
     description: "Retrieves apps registered in the current Firebase project.",
     inputSchema: z.object({
       platform: z
-        .enum(["ios", "android", "web"])
-        .nullish()
+        .enum(["ios", "android", "web", "all"])
+        .default("all")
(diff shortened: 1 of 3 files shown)


src/mcp/tools/core/list_apps.ts
```diff
@@ -23,10 +23,15 @@ export const list_apps = tool(
...
-    );
-    return toContent(apps);
+    try {
+      const apps = await listFirebaseApps(
+        projectId!,
+        platform === "all" ? AppPlatform.ANY : (platform.toUpperCase() as AppPlatform),
+      );
+      return toContent(apps);
+    } catch (err: any) {
+      const originalMessage = err.original ? `: ${err.original.message}` : "";
```
fredzqm, line 33: Wonder if this logic is worth porting to the upstream exception handler.

It feels like a generic enough improvement.

https://github.com/firebase/firebase-tools/blob/master/src/mcp/index.ts#L223

fredzqm (approved):
The change look good & low risk.

Left some optional thoughts for consideration.

PR #8616, opened 2025-05-16
Pin standalone to yao-pkg/pk@6.4.1

joehan:
### Description
yao-pkg/pkg@6.5.0 is breaking the Firepit build job because it points at node binaries that were not uploaded to Github (see https://github.com/yao-pkg/pkg-fetch/pull/82). Until that is fixed, pin to the last known good version.

### Scenarios Tested
Used this when repairing the issues from last release, and it worked correctly.

### DIFF
--- a/standalone/package-lock.json
+++ b/standalone/package-lock.json
@@ -16,7 +16,7 @@
         "user-home": "^2.0.0"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


aalej (approved):
This change looks good to me! Tested running the build script locally. 

Though it looks like a patch was released in 6.5.1, might also be a good option to pin to that version.

PR #8636, opened 2025-05-21
Update function deploy to add necessary Genkit monitoring permissions when deploying Genkit function

bryanatkinson:
### Description

Genkit Monitoring requires that the service account running the Genkit code has permission to write metrics, traces and logs. This change adds those permissions to the default service account when deploying as a Firebase Function.

### Scenarios Tested

* Tested with newly added Genkit function and ensure the permissions are set on the first firebase deploy
* Added unit tests

### DIFF
--- a/src/deploy/functions/checkIam.spec.ts
+++ b/src/deploy/functions/checkIam.spec.ts
@@ -161,8 +161,8 @@ describe("checkIam", () => {
           projectId,
           projectNumber,
           backend.of(wantFn),
-          backend.empty()
-        )
+          backend.empty(),
+        ),
       ).to.not.be.rejected;
       expect(storageStub).to.have.been.calledOnce;
       expect(getIamStub).to.have.been.calledOnce;
@@ -194,12 +194,12 @@ describe("checkIam", () => {
           projectId,
           projectNumber,
           backend.of(wantFn),
-          backend.empty()
-        )
+          backend.empty(),
+        ),
       ).to.be.rejectedWith(
         "We failed to modify the IAM policy for the project. The functions " +
           "deployment requires specific roles to be granted to service agents," +
-          " otherwise the deployment will fail."
+          " otherwise the deployment will fail.",
       );
       expect(storageStub).to.have.been.calledOnce;
       expect(getIamStub).to.have.been.calledOnce;
@@ -256,7 +256,7 @@ describe("checkIam", () => {
         projectId,
         projectNumber,
         backend.of(wantFn),
-        backend.empty()
+        backend.empty(),
       );
 
       expect(storageStub).to.have.been.calledOnce;
@@ -312,7 +312,7 @@ describe("checkIam", () => {
       projectId,
       projectNumber,
       backend.of(wantFn),
-      backend.of(haveFn)
+      backend.of(haveFn),
     );
 
     expect(storageStub).to.have.been.calledOnce;
@@ -365,7 +365,7 @@ describe("checkIam", () => {
       projectId,
       projectNumber,
       backend.of(wantFn),
-      backend.empty()
+      backend.empty(),
     );
 
     expect(getIamStub).to.have.been.calledOnce;
@@ -401,7 +401,7 @@ describe("checkIam", () => {
       projectId,
       projectNumber,
       backend.of(wantFn),
-      backend.of(haveFn)
+      backend.of(haveFn),
     );
 
     expect(getIamStub).to.not.have.been.called;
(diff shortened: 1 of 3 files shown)


andrewbrook (approved):

bryanatkinson:
> 1. We only do IAM stuff on new function creation because it requires an OWNER permission rather than an editor permission and we want large teams to be able to have finer grained ACLs
> 2. I'm concerned that this is granting the binding to the default compute service account, rather than the accounts that the actual functions run as (which defaults to the compute SA but is not guaranteed to be so).
> 
> For 2, it's not clear that this is technically correct. @taeold do you know why obtainDefaultComputeServiceAgentBindings is here? Is this part of the bug/feature request that we don't allow configuring Eventarc to use custom service accounts yet?
> 
> I would consider a new dedicated function that
> 
> 1. Finds all _new_ endpoints that are Genkit callable endpoints
> 2. Finds the set of service accounts they run as (defaulting to the default compute SA)
> 3. Adds those role bindings for the SA to the project.

I have updated to move this into a separate function called in the prepare step, and to find all service accounts for new Genkit endpoints only. Let me know if this doesn't match what you were thinking.

src/deploy/functions/checkIam.ts
```diff
@@ -168,6 +180,52 @@ export async function obtainDefaultComputeServiceAgentBindings(
...
+  projectId: string,
+  projectNumber: string,
+  want: backend.Backend,
+  have: backend.Backend,
+  dryRun?: boolean,
+): Promise<void> {
+  const wantEndpoints = backend.allEndpoints(want).filter(isGenkitEndpoint);
+  const haveEndpoints = backend.allEndpoints(have).filter(isGenkitEndpoint);
+  const newEndpoints = wantEndpoints.filter(
+    (wantE) => !haveEndpoints.find((haveE) => haveE.id === wantE.id),
```
inlined, line 201: Technically you should compare the region too. There's a helper here:

```
wantEndopints.filter(backend.missingEndpoint(haveEndpoints))
```

src/deploy/functions/checkIam.ts
```diff
@@ -168,6 +180,52 @@ export async function obtainDefaultComputeServiceAgentBindings(
...
+  );
+
+  if (newEndpoints.length === 0) {
+    return;
+  }
+
+  const defaultComputeServiceAgent = await gce.getDefaultServiceAccount(projectNumber);
```
inlined, line 208: It'd be ideal to only calculate this if necessary. I'm not sure what this code would do on a project that has disabled the default SA.

src/deploy/functions/checkIam.ts
```diff
@@ -215,8 +288,8 @@ export async function ensureServiceAgentRoles(
     iam.printManualIamConfig(requiredBindings, projectId, "functions");
     utils.logLabeledBullet(
       "functions",
-      "Could not verify the necessary IAM configuration for the following newly-integrated services: " +
-        `${newServices.map((service) => service.api).join(", ")}` +
+      "Could not verify the necessary IAM configuration for the following newly-integrated services or endpoints: " +
```
inlined, line 291: endpoints is an internal term because we were experimenting with code that would eventually become App Hosting. It is meaningless to users.

inlined (approved):

src/deploy/functions/checkIam.spec.ts
```diff
@@ -244,6 +244,363 @@ describe("checkIam", () => {
...
+      ).to.be.rejectedWith(
+        "We failed to modify the IAM policy for the project. The functions " +
+          "deployment requires specific roles to be granted to service agents," +
+          " otherwise the deployment will fail.",
+      );
+      expect(getIamStub).to.have.been.calledOnce;
+      expect(getIamStub).to.have.been.calledWith(projectNumber);
+      expect(setIamStub).to.have.been.calledOnce;
+    });
+    it("should not update policy if it already has necessary bindings", async () => {
```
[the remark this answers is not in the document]
bryanatkinson: Fixed

src/deploy/functions/checkIam.ts
bryanatkinson: Nice! Updated.

src/deploy/functions/checkIam.ts
bryanatkinson: Updated to only make this call if the default is needed.

src/deploy/functions/checkIam.ts
bryanatkinson: Reverted this back to just say "services"

PR #8648, opened 2025-05-23
Split confirmWriteProjectFile out of askWriteProjectFile

fredzqm:
Doing some refactor to make split prompt and file write actuation easier. 

1. Moved logging into `config.writeProjectFile` method. Audited all use cases of it and update surrounding logs to avoid duplications.
2. Extract a `confirmWriteProjectFile` out of `askWriteProjectFile`.

It's very common in `doSetup` to use `askWriteProjectFile`.
After the refactor, we can make `askQuestions` calls `confirmWriteProjectFile` and saves the response. `actuate` calls `askWriteProjectFile`.

### From Empty Folder

Screenshot 2025-05-22 at 5 57 29 PM

### Re-running init again

Screenshot 2025-05-22 at 5 58 24 PM

It's arguably better than the existing behavior, which ask for confirmation even if there is no change in Firestore / storage rules.

#### [Existing] Re-running init

Screenshot 2025-05-22 at 5 50 14 PM

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -17,7 +17,6 @@ import * as utils from "./utils";
 import { getValidator, getErrorMessage } from "./firebaseConfigValidate";
 import { logger } from "./logger";
 import { loadCJSON } from "./loadCJSON";
-import { b } from "@electric-sql/pglite/dist/pglite-DqRPKYWs";
 const parseBoltRules = require("./parseBoltRules");
 
 export class Config {
--- a/src/config.ts
+++ b/src/config.ts
@@ -260,7 +260,7 @@ export class Config {
       message: "File " + clc.underline(path) + " already exists. Overwrite?",
       default: !!confirmByDefault,
     });
-    if (shouldWrite) {
+    if (!shouldWrite) {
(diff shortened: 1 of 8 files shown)


src/config.ts
```diff
@@ -233,36 +241,45 @@ export class Config {
...
-    if (existingContent && existingContent !== content && !force) {
-      next = await confirm({
-        message: "File " + clc.underline(path) + " already exists. Overwrite?",
-        default: !!confirmByDefault,
-      });
+    const shouldWrite = await confirm({
+      message: "File " + clc.underline(path) + " already exists. Overwrite?",
+      default: !!confirmByDefault,
+    });
+    if (shouldWrite) {
```
joehan, line 263: This seems reversed - shouldn't it be !shouldWrite

joehan (approved):
One bug to fix, but lgtm otherwise

src/config.ts
fredzqm: Thanks a bunch!

Missed the case when file changed in manual testing.

PR #8653, opened 2025-05-23
Firebase init + deploy should setup a new project without going to console

fredzqm:


CHANGELOG.md
```diff
@@ -3,3 +3,4 @@
 - Fixed crash when starting the App Hosting emulator in certain applications (#8624)
 - Fixed issue where, with `webframeworks` enabled, `firebase init hosting` re-prompts users for source. (#8587)
 - Update typescript version in functions template to avoid build issue with @google-cloud/storage depedency (#8194)
+- [added] Support `firebase init` + `firebase deploy` flow on a fresh new project without going to Console (#8653)
```
joehan, line 6:
```suggestion
- Improved `firebase init firestore` + `firebase deploy` flow on a fresh new project without going to Console (#8653)
```

src/deploy/firestore/deploy.ts
```diff
@@ -1,10 +1,41 @@
...
 import { FirestoreApi } from "../../firestore/api";
+import * as types from "../../firestore/api-types";
 import { logger } from "../../logger";
 import * as utils from "../../utils";
 import { RulesDeploy, RulesetServiceType } from "../../rulesDeploy";
 import { IndexContext } from "./prepare";
+import { FirestoreConfig } from "../../firebaseConfig";
+import { sleep } from "../../utils";
+
+async function createDatabase(context: any, options: any): Promise<void> {
```
joehan, line 12: Nit: Should use Options (from options.ts) here instead of any - that should clear out a bunch of these type warnings.

src/deploy/firestore/deploy.ts
```diff
@@ -45,22 +76,33 @@ async function deployIndexes(context: any, options: any): Promise<void> {
...
-          )} successfully for ${databaseId} database`,
-        );
-      });
+      try {
+        await firestoreIndexes.deploy(options, indexes, fieldOverrides, databaseId);
+      } catch (err: any) {
+        if (err.status !== 404) {
+          throw err;
+        }
+        // It might take a while for the database to be created.
```
joehan, line 85: Is there a cleaner way to handle this case rather than justign waiting? Is there an operation that we ought to be polling on?

Can be done in a follow up if its annoyingly complex.

src/init/features/firestore/indexes.ts
```diff
@@ -53,6 +53,9 @@ async function getIndexesFromConsole(
     const res = await Promise.all([indexesPromise, fieldOverridesPromise]);
     return JSON.stringify(indexes.makeIndexSpec(res[0], res[1]), null, 2);
   } catch (e: any) {
+    if (e.status === 404 || e.code === 403) {
```
joehan, line 56: Catching 403s seems possibly unsafe - won't this catch and silence cases where the user does not have the required permission?

src/deploy/firestore/deploy.ts
fredzqm: Done

src/deploy/firestore/deploy.ts
```diff
@@ -1,10 +1,42 @@
...
+import * as types from "../../firestore/api-types";
 import { logger } from "../../logger";
 import * as utils from "../../utils";
 import { RulesDeploy, RulesetServiceType } from "../../rulesDeploy";
 import { IndexContext } from "./prepare";
+import { FirestoreConfig } from "../../firebaseConfig";
+import { sleep } from "../../utils";
+import { Options } from "../../options";
+
+async function createDatabase(context: any, options: Options): Promise<void> {
```
joehan, line 13: All the nonnull assertions in here are giving code smells that we might hit accidental NPEs from some edge case here. Can we clean up the typing at all to defend against this?

joehan (commented):
LGTM after some minor type cleanup

src/deploy/firestore/deploy.ts
fredzqm: Sure, I can do that here~

The typesafe-ness in `firebase deploy` in general is quite bad. Most args are passed as `any`  right now.

PR #8660, opened 2025-05-27
[MCP] firebase_init support storage

fredzqm:
Screenshot 2025-05-27 at 8 28 37 AM

### DIFF
--- a/src/init/features/storage.ts
+++ b/src/init/features/storage.ts
@@ -3,22 +3,54 @@ import * as clc from "colorette";
 import { logger } from "../../logger";
 import { input } from "../../prompt";
 import { readTemplateSync } from "../../templates";
+import { Config } from "../../config";
+import { Setup } from "..";
+import { FirebaseError } from "../../error";
+
+export interface RequiredInfo {
+  rulesFilename: string;
+  rules: string;
+  writeRules: boolean;
+}
 
 const RULES_TEMPLATE = readTemplateSync("init/storage/storage.rules");
+const DEFAULT_RULES_FILE = "storage.rules";
 
-export async function doSetup(setup: any, config: any): Promise<void> {
-  setup.config.storage = {};
-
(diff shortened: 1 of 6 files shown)


src/init/features/storage.ts
```diff
@@ -3,22 +3,54 @@ import * as clc from "colorette";
...
   logger.info("and publish them with " + clc.bold("firebase deploy") + ".");
   logger.info();
 
-  const storageRulesFile = await input({
+  const info: RequiredInfo = {
+    rulesFilename: DEFAULT_RULES_FILE,
+    rules: RULES_TEMPLATE,
+    writeRules: true,
+  };
+  info.rulesFilename = await input({
```
joehan, line 31: Separate from your PR: IDK why we even ask this in init, we probably should just always use  {current value from firebase.json} || storage.rules

src/mcp/tools/core/init.ts
```diff
@@ -78,11 +78,30 @@ export const init = tool(
...
+            rules_filename: z
+              .string()
+              .optional()
+              .default("storage.rules")
+              .describe("The file to use for Firebase Storage Security Rules."),
+            rules: z
+              .string()
+              .optional()
+              .describe(
+                "The security rules to use for Firebase Storage Security Rules. Default to open rules that deny all access.",
```
joehan, line 92: Confusing sentence here - are these rules actually open?
```
open rules that deny all access.",
```

joehan (approved):
LGTM with some nits and thoughts about existing storage init

src/init/features/storage.ts
fredzqm: Err, we kind of have this pattern of asking rule name.

I guess to confirm it in most cases so people know what this file is for~

src/mcp/tools/core/init.ts
fredzqm: Good catch! Shouldn't trust AI auto complete too much

PR #8661, opened 2025-05-27
Allow multiSelect

### LINKED ISSUE #8657: Invalid type "multiselect" error in extensions.yaml during firebase ext:dev:upload

**firebase-tools: 14.4.0**

**Platform: macOS 15.4.1 (24E263)**

### [REQUIRED] Test case

When uploading an extension using firebase ext:dev:upload, an error occurs if the extensions.yaml file includes a type: multiselect parameter. According to the Firebase Extensions YAML reference (https://firebase.google.com/docs/extensions/reference/extension-yaml), multiselect (one word) is listed as a valid type for the type parameter in param definitions. However, the upload fails with an error indicating that multiselect is invalid.

Example extensions.yaml snippet:
```yaml
params:
  - param: ALERTS
    label: Alerts
    type: multiselect
    options:
      - label: Option 1
        value: option1
      - label: Option 2
        value: option2
```

1. Create an extensions.yaml file with a parameter using type: multiselect as shown above.
1. Run the command firebase ext:dev:upload to upload the extension.
1. Observe the error message.

The firebase ext:dev:upload command should successfully upload the extension, recognizing multiselect as a valid type for a parameter, as documented in the Firebase Extensions YAML reference (https://firebase.google.com/docs/extensions/reference/extension-yaml).

The upload fails with the following error:

```bash
Error: The extension.yaml has the following errors: 
  - Invalid type multiselect for param ALERTS. Valid types are select, multiSelect, string, selectResource, secret
```

The error message suggests multiSelect (camelCase) as a valid type, which contradicts the documentation's use of multiselect (lowercase, one word).

joehan:
### Description
Fixes #8657


aalej (commented):
qq, do we plan to support both `multiselect` and `multiSelect`? If we're only planning on supporting `multiSelect`, I think we can remove `multiselect`

htsuruo:
@joehan @aalej 

>edit: we might also want to update

This is just documentation. 
The update is also needed, but following line is more important and must be updated if `multiSelect` is correct.
https://github.com/firebase/firebase-tools/blob/e4b79eae9e93647f486204021130b986c89f9769/schema/extension-yaml.json#L124

aalej (approved):
LGTM! Verified the schema update locally.

image

PR #8664, opened 2025-05-28
Firestore mcp emulator

tammam-g:


src/gcp/firestore.ts
```diff
@@ -149,7 +154,7 @@ export async function getDatabase(
   database: string,
   allowEmulator: boolean = false,
 ): Promise<Database> {
-  const apiClient = allowEmulator ? emuOrProdClient : prodOnlyClient;
+  const apiClient = getClient(allowEmulator);
```
fredzqm, line 157: （offline discussion)

Not a big fan of the `allowEmulator` boolean + the env override. Can do a simple refactor to change them to take a `apiClient` override option.
If the client wants to use emulator, it can override.

src/mcp/tools/firestore/commit_document.ts
```diff
@@ -0,0 +1,71 @@
+import { z } from "zod";
+import { tool } from "../../tool.js";
+import { mcpError, toContent } from "../../util.js";
+import { commitDocument, FirestoreDocument, FirestoreValue } from "../../../gcp/firestore.js";
+import { convertInputToValue, firestoreDocumentToJson } from "./converter.js";
+import { getFirestoreEmulatorHost } from "./emulator.js";
+import { NO_PROJECT_ERROR } from "../../errors.js";
+
+export const commit_document = tool(
```
fredzqm, line 9: Let's separate it out to a separate PR

PR #8672, opened 2025-05-29
v20 made outputPath optional, fix

jamesdaniels:


src/frameworks/angular/utils.ts
```diff
@@ -411,10 +411,10 @@ export async function getBrowserConfig(sourceDir: string, configuration: string)
     architectHost.getBuilderNameForTarget(buildOrBrowserTarget),
   ]);
 
-  assertIsString(targetOptions?.outputPath);
+  const buildOutputPath = typeof targetOptions?.outputPath === "string" ? targetOptions.outputPath : join("dist", buildOrBrowserTarget.project);
```
jamesdaniels, line 414: Maybe more appropriate for an assert `string | null | undefined`?

PR #8674, opened 2025-05-29
Improve log messages on skipped App Hosting source deploys

blidd-google:
Fixes misleading log messages during firebase deploy by hiding logs for skipped App Hosting backends and clearly listing which ones were skipped.

To improve user feedback, spinners will now be displayed during service account and web app creation to indicate that the operations are in progress.


src/deploy/apphosting/prepare.ts
```diff
@@ -118,39 +120,43 @@ export default async function (context: Context, options: Options): Promise<void
...
+        context.backendLocations.set(cfg.backendId, location);
+      }
+    } else {
+      skippedBackends.push(...notFoundBackends);
+    }
   }
+  logLabeledWarning(
+    "apphosting",
+    `Skipping deployment of backend(s) ${skippedBackends.map((cfg) => cfg.backendId).join(", ")}.`,
+  );
```
annajowang, line 159: do we want to hide this log under some  `if (skippedBackends.length > 0):` ?

blidd-google: Yes - thanks for catching

PR #8684, opened 2025-05-30
Improve automation for env vars for Data Connect init.

yuchenshi:
### Description

See CHANGELOG.md

### Scenarios Tested

Screenshot_1748644971

### Sample Commands

`firebase init`

`firebase init dataconnect`


src/init/features/project.ts
```diff
@@ -120,14 +120,22 @@ export async function doSetup(setup: any, config: any, options: any): Promise<vo
...
-  // If the user presented a project with `--project`, try to fetch that project.
   if (options.project) {
+    // If the user presented a project with `--project`, try to fetch that project.
     logger.debug(`Using project from CLI flag: ${options.project}`);
     projectMetaData = await getFirebaseProject(options.project);
   } else {
-    projectMetaData = await projectChoicePrompt(options);
-    if (!projectMetaData) {
-      return;
+    const projectEnvVar = utils.envOverride("GCLOUD_PROJECT", "");
```
fredzqm, line 128: Defer to @joehan on this~

Is this a reasonable env name?

fredzqm (approved):

src/init/features/project.ts
joehan: I think it probably should be GOOGLE_CLOUD_PROJECT or FIREBASE_PROJECT

https://github.com/GoogleCloudPlatform/nodejs-getting-started/issues/251 GCLOUD_PROJECT is a deprecated CF3 env var - obviously not gonna affect this, but I'd prefer to keep consistency accross products when possible

We also have some code in place that populates GCLOUD_PROJECT to support very old firebase-functions versions, so it'll be nice to avoid any interactions with that by using GOOGLE_CLOUD_PROJECT instead

joehan (approved):
LGTM, but probably switch the env var

src/init/features/project.ts
yuchenshi: Okay I've changed it to `FIREBASE_PROJECT`. Later maybe we can make it a drop-in replacement of `--project`. Or not. The script can work either way

Issue #8689, opened 2025-06-02
framework: npm timeout in big projects breaks deploy


tharvik:
when deploy w/ a framework, `npm list --json` is called a few times via `findDependency`. in some larger project, the relatively low timeout (10s) will trigger and make the function throw when parsing it as JSON.
note: the `result.stdout` check is never taken as it is a `Buffer`, not a `string`

I've got a small patch to `utils.envOverride("NPM_COMMAND_TIMEOUT_MILLIES", …)` in `constants.ts` but I'd even argue that removing the whole timeout check is a saner fix (w/ this Halting issue of Turing complet languages ya know). WDYT?

looptheloop88:
Hi @tharvik, thanks for bringing this matter to our attention. Please go ahead and submit a pull request for our engineers to review.

tharvik:
version removing timeout https://github.com/firebase/firebase-tools/pull/8690

leoortizz:
Hey @tharvik, thank you for sharing your issue and for the PR that can fix it. 

To better understand the issue, what framework are you using? Additionally, if you can share the dependencies in your project's `package.json` which is causing the issue, would be great for us to investigate it.

tharvik:
we are using NextJS. the deps list is not really relevant as it is more the machine running the deploy that causes the issue. FWIW, we are using `pnpm` to handle deps so `npm list` seems to be doing more computing than with regular npm systems.
you can trigger it more easily by reducing the usage of `firebase-tools deploy` with [`cpulimit`](https://cpulimit.sourceforge.net/).

anyway here goes the deps
```json
{
  "dependencies": {
    "@chakra-ui/icons": "^2.2.4",
    "@chakra-ui/layout": "^2.3.1",
    "@chakra-ui/next-js": "^2.4.2",
    "@chakra-ui/react": "^2.10.7",
    "@dnd-kit/core": "^6.3.1",
    "@dnd-kit/modifiers": "^9.0.0",
    "@dnd-kit/utilities": "^3.2.2",
    "@emmetio/xml-diff": "^2.2.0",
    "@emotion/react": "^11.14.0",
    "@emotion/styled": "^11.14.0",
    "@heroicons/react": "^2.2.0",
    "@sentry/nextjs": "^9.18.0",
    "@tiptap/extension-blockquote": "^2.12.0",
    "@tiptap/extension-bold": "^2.12.0",
    "@tiptap/extension-bullet-list": "^2.12.0",
[...]

PR #8690, opened 2025-06-02
framework: remove npm timeout

tharvik:
### Description

see #8689

### Scenarios Tested

`firebase deploy` with my project now works as it is not stopped when listing deps.

### Sample Commands

nothing changed there


tharvik:
closed in favor of #8702

PR #8691, opened 2025-06-02
Update typings for Gemini in Firebase api calls

hlshen:
Add types file for Cloud AI Companion experience calls.
Add chat history as an optional input

### DIFF
--- a/src/gif/fdcExperience.ts
+++ b/src/gif/fdcExperience.ts
@@ -1,61 +1,19 @@
 import { Client } from "../apiv2";
 import { cloudCompanionOrigin } from "../api";
+import {
+  ChatExperienceResponse,
+  CloudAICompanionMessage,
+  CloudAICompanionRequest,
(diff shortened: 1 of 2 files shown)


src/gif/fdcExperience.ts
```diff
@@ -1,75 +1,37 @@
 import { Client } from "../apiv2";
 import { cloudCompanionOrigin } from "../api";
+import {
```
joehan: Nit: Could we rename this directory to 'gemini'? GIF still reads as the image format for me (and definitely will for external contributore)

joehan (approved):

joehan:
Also looks like this broke the vsce unit tests - lets fix those first.

hlshen:
> Also looks like this broke the vsce unit tests - lets fix those first.

This is unrelated to this PR. However, I have a fix in #8693

PR #8697, opened 2025-06-03
Fix issue where values for results aren't being set

### LINKED ISSUE #8694: Alias name is empty / undefined when associating the app with the Firebase project
When I followed [this step](https://firebase.google.com/codelabs/firestore-web#3) in the firestore codelab tutorial to associate my app with my Firebase project (run `firebase use --add`), the alias name did not show up on the firebase CLI. It shows `created alias  for undefined`.

Image

aalej:
### Description

Fixes #8694

Looks like we never set the value of `results` in
https://github.com/firebase/firebase-tools/blob/c8dd88aadc6269419bfd1a47da6d5fe93da3f006/src/commands/use.ts#L106-L122
so `results.project` and `results.alias` are always `undefined`

### Scenarios Tested

Verified that old behavior matched current behavior
#### Old behavior (14.2.0)
```
$ firebase --version
14.2.0
$ firebase use
Active Project: default (aalej-6162)

Project aliases for /Users/PATH/Desktop/firebase-tools/issues/8694:

* default (aalej-6162)

Run firebase use --add to define a new project alias.
$ firebase use --add
? Which project do you want to add? aalej-6162-2
? What alias do you want to use for this project? (e.g. staging) staging

Created alias staging for aalej-6162-2.
Now using alias staging (aalej-6162-2)
```

#### Current behavior

```
$ firebase --version
14.5.1
$ firebase use
Active Project: default (aalej-6162)

Project aliases for /Users/PATH/Desktop/firebase-tools/issues/8694:

* default (aalej-6162)

Run firebase use --add to define a new project alias.
$ firebase use --add 
✔ Which project do you want to add? aalej-6162-2
✔ What alias do you want to use for this project? (e.g. staging) staging

Created alias staging for aalej-6162-2.
Now using alias staging (aalej-6162-2)
```


jiechuns:
Thank you for the quick fix!

PR #8698, opened 2025-06-03
Fix "empty" unit test

hlshen:
Fixes the empty unit test, which sometimes fails due to lack of workspace.

Small buffer added to sidebar loading.


joehan:
This looks fine to me, except that the integration tests are failing on it still 😬 
Screenshot 2025-06-03 at 9 28 51 AM

PR #8700, opened 2025-06-03
[MCP] Support firestore emulator in MCP server

tammam-g:
Adds support for emulator in firestore

image

image

image

image

### DIFF
--- a/src/mcp/tools/firestore/query_collection.ts
+++ b/src/mcp/tools/firestore/query_collection.ts
@@ -3,6 +3,7 @@ import { tool } from "../../tool.js";
 import { mcpError, toContent } from "../../util.js";
 import { queryCollection, StructuredQuery } from "../../../gcp/firestore.js";
 import { convertInputToValue, firestoreDocumentToJson } from "./converter.js";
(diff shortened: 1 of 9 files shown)


src/mcp/tools/firestore/query_collection.ts
```diff
@@ -74,6 +75,7 @@ export const query_collection = tool(
         .number()
         .describe("The maximum amount of records to return. Default is 10.")
         .optional(),
+      use_emulator: z.boolean().default(false).describe("Target the Firestore emulator if true."),
```
fredzqm, line 78: nit: Shall we do `.optional().default(false)`?

I recall Gemini has some trouble with nullable(), but works better with optional().

fredzqm:
Since Firebase MCP is publicly announced, let's start adding CHANGELOG.md notes.

src/mcp/tools/firestore/query_collection.ts
tammam-g: Haven't had problems in gemini (used it from roocode/cline). If it turns out to be a problem we can change it everywhere in one PR cause this is consistent with data connect.

PR #8706, opened 2025-06-04
[MCP] Simplify and standardize how we get a given emulator endpoint

tammam-g:
Unifies how we get the emulator endpoint string for a given emulator. 

Firebase storage based on this PR.

Examples confirming firestore and dataconnect still work

image

image


joehan (approved):
Good call splitting this out!

PR #8728, opened 2025-06-08
Clean up 'storage.rules' file created during 'npm run test'

aalej:
### Description

Running `npm run test` creates a file `/test/storage.rules`.
image

This PR change creates a temporary directory where the test can run. Also removes the temporary directory after tests. Copied similar code from https://github.com/firebase/firebase-tools/blob/master/src/appdistribution/options-parser-util.spec.ts

### Scenarios Tested

Verified that `npm run test` creates and removes the temporary directory created.

### DIFF
--- a/src/init/features/storage.spec.ts
+++ b/src/init/features/storage.spec.ts
@@ -1,12 +1,15 @@
 import { expect } from "chai";
 import * as _ from "lodash";
 import * as sinon from "sinon";
+import * as tmp from "tmp";
+import { rmSync } from "node:fs";
 
 import { Config } from "../../config";
 import { askQuestions, actuate } from "./storage";
 import * as prompt from "../../prompt";
 
 describe("storage", () => {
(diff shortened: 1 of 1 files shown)


aalej:
For some reason, on the initial run `npm run test:storage-emulator-integration` failed https://github.com/firebase/firebase-tools/actions/runs/15520718312/job/43693672180. But on every succeeding run, it has been successful. Maybe it was flaky?

joehan:
> For some reason, on the initial run `npm run test:storage-emulator-integration` failed https://github.com/firebase/firebase-tools/actions/runs/15520718312/job/43693672180. But on every succeeding run, it has been successful. Maybe it was flaky?

Ya, that one has falked before (usually jsut due to timeouts). If it is passing on subsequent runs, i would not worry about it (the individual runs do not share any state with each other)

joehan:
Closing in favor of https://github.com/firebase/firebase-tools/pull/8734

PR #8732, opened 2025-06-09
Improved metrics for firebase studio and init

joehan:
### Description
Slightly expanding the metrics we collect:
- New user metric to track usage from Firebase Studio
- New `products_init` metric to better understand how init is used.

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -15,6 +15,7 @@ import { Options } from "../options";
 import { isEnabled } from "../experiments";
 import { readTemplateSync } from "../templates";
 import { FirebaseError } from "../error";
+import { trackGA4 } from "../track";
 
 const homeDir = os.homedir();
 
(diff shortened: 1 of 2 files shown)


src/commands/init.ts
```diff
@@ -246,6 +249,10 @@ export async function initAction(feature: string, options: Options): Promise<voi
   if (!fsutils.fileExistsSync(config.path(".gitignore"))) {
     config.writeProjectFile(".gitignore", GITIGNORE_TEMPLATE);
   }
+  const duration = Math.floor((process.uptime() - start) * 1000);
+
+  await trackGA4("product_init", { products_initialized: setup.features?.join(",") }, duration);
```
fredzqm, line 254: Emm, will this be easy to parse in GA?

We may have `firestore,dataconnect`, `dataconnect`, `dataconnect,storage` all mixed up.

It will be hard to find number of Firestore init for example.

Does GA support a list of labels?

joehan: There isn't a great option ,unfortunately. For deploy, we instead did 1 custom field per product, and it ended up being nearly impossible to parse within GA's explorer tool. I think this ugly approach will actually make things easier - we'll have an easy count for each combination that gets used, and we can use contains filters to find all that include 'Firestore', for example

PR #8735, opened 2025-06-10
Allow MCP to use ADC

### LINKED ISSUE #8671: [MCP] MCP server not auth with Application Default Credentials (ADC)
## Summary

Firebase tools MCP not auth with ADC

## Bug Info 

- **Affected Tool(s):** All tools
- **MCP Client:**  @modelcontextprotocol/inspector
- **Operating System:** Macos 14.7

### Steps to Reproduce
1. Run command: gcloud auth application-default login
2. Call tools firebase tools mcp: npx @modelcontextprotocol/inspector --cli npx -y firebase-tools@latest experimental:mcp --method tools/call --tool-name  firebase_create_project --tool-arg project_id="{{SAMPLE_NAME_HERE}}"

If run tool/list. I got this meta data:
```
  "_meta": {
    "projectRoot": "{{ dir }}",
    "projectDetected": false,
    "authenticatedUser": null,
    "activeFeatures": [],
    "detectedFeatures": []
  }
```

joehan:
### Description
Fixes #8671.

### Scenarios Tested
Ran `firebase logout`, `gcloud auth application-default login`, and then tested a tool that required auth and confirmed that it did not error out.

### DIFF
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -174,7 +174,8 @@ export class FirebaseMcpServer {
 
   async getAuthenticatedUser(): Promise<string | null> {
     try {
-      return await requireAuth(await this.resolveOptions());
+      const email = await requireAuth(await this.resolveOptions());
+      return email ?? "Application default credentials";
     } catch (e) {
       return null;
     }
@@ -221,12 +222,8 @@ export class FirebaseMcpServer {
     projectId = projectId || "";
 
     const accountEmail = await this.getAuthenticatedUser();
-    if (tool.mcp._meta?.requiresAuth) {
-      try {
-        await requireAuth(await this.resolveOptions());
-      } catch (e: any) {
-        return mcpAuthError();
-      }
+    if (tool.mcp._meta?.requiresAuth && !accountEmail) {
+      return mcpAuthError();
     }
 
     const options = { projectDir: this.cachedProjectRoot, cwd: this.cachedProjectRoot };
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -221,8 +221,12 @@ export class FirebaseMcpServer {
     projectId = projectId || "";
 
     const accountEmail = await this.getAuthenticatedUser();
-    if (tool.mcp._meta?.requiresAuth && !accountEmail) {
-      return mcpAuthError();
+    if (tool.mcp._meta?.requiresAuth) {
+      try {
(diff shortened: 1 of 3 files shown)


src/mcp/index.ts
```diff
@@ -221,8 +221,12 @@ export class FirebaseMcpServer {
     projectId = projectId || "";
 
     const accountEmail = await this.getAuthenticatedUser();
-    if (tool.mcp._meta?.requiresAuth && !accountEmail) {
-      return mcpAuthError();
+    if (tool.mcp._meta?.requiresAuth) {
+      try {
+        await requireAuth(await this.resolveOptions());
```
fredzqm: How it is different then calling `getAuthenticatedUser()`?

```
  async getAuthenticatedUser(): Promise<string | null> {
    try {
      return await requireAuth(await this.resolveOptions());
    } catch (e) {
      return null;
    }
  }
```

fredzqm: Correct me if I am wrong

It sounds like `getAuthenticatedUser()` currently returns:
- string for `firebase login`
- null for `gcloud auth application-default login`
- Throw error if there is neither

Should we document that clearly in `requireAuth()`.

Maybe change `this.getAuthenticatedUser()` to encode that information property

joehan: requireAuth can return null when you are authenticated via a service account, but it errors when the user is not authenticated. getAuthenticated catches those errors and returns null. Under the previous logic, this triggered the conditional that returned mcpAuthError()

joehan: Good call - refactored this. Now, getAuthenticatedUser() returns:
email for firebase login
'Application default credentials when using ADC
null for gcloud auth application-default login

and we skip the redundant requireAuth call.

src/mcp/index.ts
```diff
@@ -174,7 +174,8 @@ export class FirebaseMcpServer {
 
   async getAuthenticatedUser(): Promise<string | null> {
     try {
-      return await requireAuth(await this.resolveOptions());
+      const email = await requireAuth(await this.resolveOptions());
+      return email ?? "Application default credentials";
```
fredzqm: nit: "Application Default Credentials"

Could you also update the tsdoc on `requireAuth`? It wasn't obvious what `null` means reading it.

Thanks!

fredzqm (approved):
Thank you!

PR #8742, opened 2025-06-11
add debounce to codelens reload

hlshen:
Prevents Codelens from flickering while typing.

### DIFF
--- a/firebase-vscode/src/data-connect/code-lens-provider.ts
+++ b/firebase-vscode/src/data-connect/code-lens-provider.ts
@@ -6,6 +6,7 @@ import { Disposable } from "vscode";
 import { Signal } from "@preact/signals-core";
 import { dataConnectConfigs, firebaseRC } from "./config";
 import { EmulatorsController } from "../core/emulators";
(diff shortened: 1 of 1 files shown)


firebase-vscode/src/data-connect/code-lens-provider.ts
```diff
@@ -17,6 +18,39 @@ abstract class ComputedCodeLensProvider implements vscode.CodeLensProvider {
...
+
+    vscode.workspace.onDidChangeTextDocument((event) => {
+      if (this.document && event.document === this.document) {
+        this.debouncedParse();
+      }
+    });
+
+    vscode.window.onDidChangeActiveTextEditor((editor) => {
+      if (editor) {
+        this.document = editor.document;
```
yuchenshi, line 44: Will this parse non-graphql documents too?

joehan: Nope - this code implements the code lens provider but does not register it. We tell vscode what files to apply this for when we register it (see src/dataconnect/index.ts)

PR #8750, opened 2025-06-12
MVP working transaction fix

tammam-g:


joehan (approved):
LGTM, but I coauthored.

src/emulator/dataconnect/pgliteServer.ts
```diff
@@ -130,18 +128,89 @@ export class PostgresServer {
...
+    const tempOldDb = await PGlite02.create({
+      loadDataDir: dumpDir,
+      extensions,
+    });
+
+    const dumpResult = await pgDump({ pg: tempOldDb, args: ["--verbose", "--verbose"] });
+    await tempOldDb.close();
+    await oldDb.close();
+
+    // 4. Nuke old data directory
```
yuchenshi: Can we do this step last so that if something goes wrong during the import, the developer can still go back to an older CLI version? 

Even better, what if we create a subdirectory in dataDir, say `pg17` and only write to that subdirectory starting from this PR? In this way, the old and new CLI versions should not mess with each other. Ever time we change the on-disk format we can move to a different subdir name.

yuchenshi (changes requested):
Code looks good except one compatibility concern inline

src/emulator/dataconnect/pgliteServer.ts
joehan: That is a good call, implemented it:
When we migrate you, we'll copy your old dataDir over to dataDir/pg16. Then, we will use dataDir/pg17 going forward.

yuchenshi: Thanks, this looks great. I think we can make it even better by leaving existing data prefixless instead of `pg16`. In this way, downgrades are 100% safe out of the box without changing the dataDir settings.

yuchenshi (approved):

Issue #8543, opened 2025-05-12
NextJs Hosting not picking up .env.develop only .env.production


JackMF:
### [REQUIRED] Environment info

**firebase-tools:** 14.3.1

**Platform:** macOs

### [REQUIRED] Test case
set up a next js project following this guide: [https://firebase.google.com/docs/hosting/frameworks/nextjs](https://firebase.google.com/docs/hosting/frameworks/nextjs)

### [REQUIRED] Steps to reproduce

1. Make your repo support mutiple firebase projects by changing your .firebaserc to have

```
{
  "projects": {
    "develop": "project-dev-id",
    "production": "project-prod-id"
  }
}
```

2. Have 3 env files:
`.env`
`.env.develop`
`.env.production`

3. Deploy project by using:

```
firebase use develop
firebase deploy

```

### [REQUIRED] Expected behavior

I would expect the project to build with the `.env.develop` vars

### [REQUIRED] Actual behavior

The project builds with the `.env.production`

```
info  - Loaded env from .env.production

info  - Loaded env from .env
```

Note this also happens if you set the develop env file to be called: `.env.project-dev-id`.

leoortizz:
Hey @JackMF, thanks for reporting your issue. In this case you should be able to use a [parametrized configuration](https://firebase.google.com/docs/functions/config-env?gen=2nd#params) to achieve your goal. Please take a look at this comment: https://github.com/firebase/firebase-tools/issues/7207#issuecomment-2168321754.

In this case you should modify your `.env` files to use the proper project ids.

Let me know if that worked for you.

JackMF:
Thanks for you feedback. This worked :D

Issue #8471, opened 2025-04-28
Mode Support for Vite Deployments


hkropp:
I have a vite app and would like to use firebase to support different modes during deployment.

Vite uses modes to pick up environment variables like so:
```
.env                # loaded in all cases
.env.local          # loaded in all cases, ignored by git
.env.[mode]         # only loaded in specified mode
.env.[mode].local   # only loaded in specified mode, ignored by git
```
From the vite documentation: https://vite.dev/guide/env-and-mode#env-files

In my case, the Firebase API key among other variables is defined inside the `.env.[mode]` configuration files.

So that, if I deploy to staging or production, it would be great to be able to define the mode during `firebase deploy --only hosting` such that the right configuration is picked up.

Currently, one gets the following warning if changing the build command in `package.json`:

```
WARNING: Your package.json contains a custom build that is being ignored. Only the Vite default build script (e.g, "vite build") is respected. If you have a more advanced build process you should build a custom integration https://firebase.google.com/docs/hosting/express
```

It would be great if the vite modes and the defined environments in `.firebaserc` would be the same. So, in case of using `firebase use ENV`, which would also be the selected build mode for vite.
This behavior would be similar to the current behavior using the Firebase emulator, where the development mode is being used.

leoortizz:
Hey @hkropp, thanks for reporting your issue. In this case you should be able to use a [parametrized configuration](https://firebase.google.com/docs/functions/config-env?gen=2nd#params) to achieve your goal. Please take a look at this comment: https://github.com/firebase/firebase-tools/issues/7207#issuecomment-2168321754 which is related to a similar issue.

In this case you should modify your `.env` files to be named with the related project ids.

Let me know if that solves your main issue.

Issue #7631, opened 2024-09-08
Can't use arbitrary value as Vite mode to build non-production code


yegorius:
### Environment info

**firebase-tools:** 13.16.0

**Platform:** Linux

### Test case

Any Vite based project which uses modes.

### Steps to reproduce

We disable AppCheck in local dev environments. We only enable it in prod or staging environments.
```
if (!dev) initializeAppCheck();
```
We also store FB configs in .env files for prod and staging FB projects (`.env.production` and `.env.development`). Building production code is fine, `firebase deploy` calls `vite build --mode production` by default.
**But how do we build staging code?**
I checked the code of firebase-tools and it seems we can use `FIREBASE_FRAMEWORKS_BUILD_TARGET` to override build target, so let's set it to `development`. Now comes the problem: when `FIREBASE_FRAMEWORKS_BUILD_TARGET` is set to `development` it also causes `NODE_ENV` to be `development` which, in its own turn, causes `dev` from `$app/environment` to be `true` (SvelteKit uses [benmccann/esm-env](https://github.com/benmccann/esm-env)). So now we have code built for staging env with `dev` set to `true`. OK, let's try setting `FIREBASE_FRAMEWORKS_BUILD_TARGET` to `staging`. Unfortunately it's not possible because this variable only supports two values `production` and `development` and bails out if anything else is supplied.
Eventually, we might find a workaround by once again monkey-patching `src/frameworks/vite/index.ts` but the problem is wider. What if someone has multiple Vite modes like: `production-us`, `production-eu`, `staging-us` and `staging-eu`? How do we pass arbitrary values to `vite build --mode`? Or what if someone has additional build step after the code is built but before it's deployed?
The problem here is that `firebase deploy` not only deploys, but also builds the code. It tries to do this auto-magically. And while doing so it takes away some freedom from the developers. After many rounds of discussions and patches, It might, eventually, get it right, but this is going to be another layer of abstraction to deal with.
My proposal is simple: don't build the code in `firebase deploy`. Let developers build the code correctly and just deploy it. It's a win-win situation: firebase-tools devs have less work to do, and fb-tools users have less things to figure out.
As in Unix philosophy:
> Make each program do one thing well. To do a new job, build afresh rather than complicate old programs by adding new "features".

### Expected behavior

`dev` is `false`

### Actual behavior

`dev` is `true`

plungarini:
Same here. +1

leoortizz:
[...]

PR #8779, opened 2025-06-23
docs: add section on configuring different environments for web frameworks

leoortizz:
### Description

Suggestion to mention parameterized configuration docs from Cloud Functions in Web Frameworks, due to many issues raised before with users that did not know how to configure different .env files when deploying to different projects/environments.

For reference, see: #7631, #8543, #8471


egilmorez (commented):
Sorry for the delay (pls don't hesitate to ping internally)!

Is there any existing content we could link to?  Is this close to the desired guidance: https://firebase.google.com/docs/functions/config-env?gen=1st#deploying_multiple_sets_of_environment_variables_2?

This sentence feels like it needs a little unpacking with an example, but if a link can help, that would work too.

egilmorez:
(The 2nd gen version to headline would be https://firebase.google.com/docs/functions/config-env?gen=2nd#deploying_multiple_sets_of_environment_variables)

leoortizz:
> Sorry for the delay (pls don't hesitate to ping internally)!
> 
> Is there any existing content we could link to? Is this close to the desired guidance: https://firebase.google.com/docs/functions/config-env?gen=1st#deploying_multiple_sets_of_environment_variables_2?
> 
> This sentence feels like it needs a little unpacking with an example, but if a link can help, that would work too.

thanks @egilmorez! I made a few changes based on your feedback, let me know what you think!

src/frameworks/docs/frameworks-overview.md
```diff
@@ -50,6 +50,16 @@ live site:
 1. Run `firebase deploy` from the terminal.
 2. Check your website on: `SITE_ID.web.app` or `PROJECT_ID.web.app` (or your custom domain, if you set one up).
 
+## Configure different environments
+
+You can deploy multiple sets of environment variables for different project environments, such as staging and production.
```
egilmorez, line 55: One more thing: People using those tools don't have lots of visibility into how functions enables the flow, do they? Therefore it might seem a little abrupt to be suddenly linked inside functions documentation.

I think we can solve that with a little extra glue like, " Like Cloud Functions for Firebase, this <tooling/adapter/whatever> supports the [dotenv](https://www.npmjs.com/package/dotenv)  file format for loading environment variables specified in a .env file."  I'd suggest that as a second sentence right after the one in line 55.

leoortizz:
```suggestion
You can deploy multiple sets of environment variables for different project environments, such as staging and production.

Like Cloud Functions for Firebase, this tooling supports the [dotenv](https://www.npmjs.com/package/dotenv)  file format for loading environment variables specified in a .env file.
```

egilmorez (approved):
LG, thanks!

PR #8780, opened 2025-06-24
Update message to show correct timeout

aalej:
### Description

When setting `FUNCTIONS_DISCOVERY_TIMEOUT` to increase the functions timeout, if the function discovery takes too long, the error message still indicates that the timeout is 10000(`Timeout after 10000`).
```
$ export FUNCTIONS_DISCOVERY_TIMEOUT=25
$ firebase emulators:start --project demo-project
i  emulators: Starting emulators: functions, extensions
i  emulators: Detected demo project ID "demo-project", emulated services will use a demo configuration and attempts to access non-emulated services for this project will fail.
i  functions: Watching "/Users/PATH/8776/functions" for Cloud Functions...
⚠  functions: Your requested "node" version "22" doesn't match your global version "20". Using node@20 from host.
Serving at port 8593

⬢  functions: Failed to load function definition from source: FirebaseError: User code failed to load. Cannot determine backend specification. Timeout after 10000. See https://firebase.google.com/docs/functions/tips#avoid_deployment_timeouts_during_initialization'
```

### Scenarios Tested

Verified that the error message now uses the set value `FUNCTIONS_DISCOVERY_TIMEOUT`(`Timeout after 25000`).
```
$ export FUNCTIONS_DISCOVERY_TIMEOUT=25
$ firebase emulators:start --project demo-project
i  emulators: Starting emulators: functions, extensions
i  emulators: Detected demo project ID "demo-project", emulated services will use a demo configuration and attempts to access non-emulated services for this project will fail.
i  functions: Watching "/Users/PATH/8776/functions" for Cloud Functions...
⚠  functions: Your requested "node" version "22" doesn't match your global version "20". Using node@20 from host.
Serving at port 8944

⬢  functions: Failed to load function definition from source: FirebaseError: User code failed to load. Cannot determine backend specification. Timeout after 25000. See https://firebase.google.com/docs/functions/tips#avoid_deployment_timeouts_during_initialization'
```

### DIFF
--- a/src/deploy/functions/runtimes/discovery/index.ts
+++ b/src/deploy/functions/runtimes/discovery/index.ts
@@ -79,12 +79,13 @@ export async function detectFromPort(
   timeout = 10_000 /* 10s to boot up */,
 ): Promise<build.Build> {
   let res: Response;
+  const discoveryTimeout = getFunctionDiscoveryTimeout() || timeout;
   const timedOut = new Promise<never>((resolve, reject) => {
     setTimeout(() => {
       const originalError = "User code failed to load. Cannot determine backend specification.";
-      const error = `${originalError} Timeout after ${timeout}. See https://firebase.google.com/docs/functions/tips#avoid_deployment_timeouts_during_initialization'`;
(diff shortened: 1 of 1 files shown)


joehan (approved):
Good catch!

PR #8785, opened 2025-06-25
Improve App Hosting compute service account flow for source deploys

blidd-google:
In the firebase init apphosting flow, we preemptively enable ensure the App Hosting compute service accounts exists and add the required IAM roles. However, we don't actually need the compute service account for any of the init steps. So instead of failing the init and deploy flows when a service account is still being provisioned, we optimistically proceed with the remaining steps.


src/apphosting/backend.ts
```diff
@@ -267,39 +266,14 @@ export async function ensureAppHostingComputeServiceAccount(
...
-    ) {
-      await addServiceAccountToRoles(
-        projectId,
-        defaultComputeServiceAccountEmail(projectId),
-        ["roles/storage.objectViewer"],
-        /* skipAccountLookup= */ true,
-      );
-    }
-  }
+  await provisionDefaultComputeServiceAccount(projectId);
```
annajowang, line 276: it's not clear to me why this was moved from the above to here. can you explain?
It seems like this leads to a potential unnecessary call of `createServiceAccount()` even if the AlreadyExists error is ultimately ignored.

src/apphosting/backend.ts
```diff
@@ -267,39 +266,14 @@ export async function ensureAppHostingComputeServiceAccount(
...
-      await provisionDefaultComputeServiceAccount(projectId);
     } else if (err.status === 403) {
       throw new FirebaseError(
         `Failed to create backend due to missing delegation permissions for ${sa}. Make sure you have the iam.serviceAccounts.actAs permission.`,
         { original: err },
       );
     }
   }
-  // N.B. To deploy from source, the App Hosting Compute Service Account must have
-  // the storage.objectViewer IAM role. For firebase-tools <= 14.3.0, the CLI does
```
annajowang, line 281: why do we not need this anymore?

src/init/features/apphosting.ts
```diff
@@ -45,19 +45,16 @@ export async function doSetup(setup: Setup, config: Config): Promise<void> {
...
-    await ensureAppHostingComputeServiceAccount(
-      projectId,
-      /* serviceAccount= */ "",
-      /* deployFromSource= */ true,
-    );
-    spinner.succeed("App Hosting compute Service account is ready");
+    await ensureAppHostingComputeServiceAccount(projectId, /* serviceAccount= */ "");
+    spinner.succeed("App Hosting compute service account is ready");
   } catch (err) {
     if ((err as FirebaseError).status === 400) {
```
annajowang: do we know for sure that a 400 error always means that the account is still being provisioned or could it be another error ? I don't know where this 400 is coming from

annajowang: actually, if i'm reading the internal chat thread correctly, 400 is from the chunk of code you've deleted in  281 - 302 ?

it seems like we'd want to do either of the following:
1) keep the IAM role setting, but ignore any 400 error from it and just log a warning saying:
    a) the SA is being provisioned in the background and 
    b) IAM roles needed for other actions may not be ready yet. 
and I assume, at some later point when the user does need the iam roles, the CLI will attempt to set them again for the SA? in which case i feel like (b) is optional
2) OR get rid of IAM role setting (as you've done in this PR). raise any errors that come from provisioning the SA. or if no errors, we just log that the SA is being provisioned in the background.

src/apphosting/backend.ts
blidd-google: Due to what I think is nondeterministic behavior from the IAM API responses, the service account was sometimes only being created with the `storage.objectViewer` role and none of the other roles. To avoid this problem, I consolidated the API calls to add the service account IAM roles into the `provisionAppHostingComputeServiceAccount` function.

src/apphosting/backend.ts
blidd-google: The `provisionDefaultComputeServiceAccount` function couples the service account creation logic with the service account role grant logic. So if the service account already exists, the `createServiceAccount` call will be skipped, but so will the `addServiceAccountToRoles` call. The `addServiceAccountToRoles` call is necessary for users who have used the CLI previously and whose existing service accounts don't have the storage.objectViewer role, which is required for deploy from source.

If you find that confusing, we could consider de-coupling the creation and role granting logic.

src/init/features/apphosting.ts
blidd-google: Unfortunately it seems the IAM API errors are somewhat non-deterministic; the `addServiceAccountToRoles` API call will sometimes return the 400 error when the service account doesn't exist yet, causing the init flow to error out. This is what Kiana was running into. 

If it is just an error due to propagation delay, we would prefer not to interrupt the init flow, so I added the logic to ignore the 400 error. However, I think it is also valid to fail out on any error in case the issue isn't related to propagation delay.

src/apphosting/backend.ts
annajowang: forgot the contents of `provisionDefaultComputeServiceAccount` until now. ok makes sense, thanks for explaining.

src/init/features/apphosting.ts
annajowang: ok got it `addServiceAccountToRoles` is still being called in `provisionDefaultComputeServiceAccount`. 

If we know that this 400 is coming from `addServiceAccountToRoles`, can we do this try-catch for the 400 in provisionDefaultComputeServiceAccount.

ideally, the errors should be handled as close as possible to the source. putting it here gives the reader no context, makes maintenance difficult, and could potentially try-catch errors we should not be catching.

src/apphosting/backend.ts
```diff
@@ -267,39 +266,14 @@ export async function ensureAppHostingComputeServiceAccount(
...
     if (!(err instanceof FirebaseError)) {
       throw err;
-    }
-    if (err.status === 404) {
-      await provisionDefaultComputeServiceAccount(projectId);
     } else if (err.status === 403) {
       throw new FirebaseError(
         `Failed to create backend due to missing delegation permissions for ${sa}. Make sure you have the iam.serviceAccounts.actAs permission.`,
         { original: err },
       );
```
annajowang, line 273: Since we got rid of that final `throw err` in apphosting.ts, i believe we need to add `throw err` here for any error that passed the first two if statements?

And then with that, I'm wondering if the error handling in prepare.ts is no longer needed and can be removed?

annajowang (approved):
lgtm with a final comment on error handling refactoring

thanks for fixing this!

PR #8790, opened 2025-06-26
Make the platform optional to give more space for LLMs to configure its value.

visumickey:


src/mcp/tools/core/list_apps.ts
```diff
@@ -26,7 +26,9 @@ export const list_apps = tool(
     try {
       const apps = await listFirebaseApps(
         projectId!,
-        platform === "all" ? AppPlatform.ANY : (platform.toUpperCase() as AppPlatform),
+        platform === "" || platform === "all"
```
mbleigh:
```suggestion
        !platform || platform === "all"
```

mbleigh (approved):

src/mcp/tools/core/list_apps.ts
visumickey: Done

PR #8793, opened 2025-06-26
Add the extended query patch back in

joehan:
### Description
And around we go again 😬 Adding the extended query patch back in. While debugging, I also ended up rehauling our logging to be quite a bit easier to understand, so adding that in too.

So, it turns out that the upstream PGlite issues were partially fixed (https://github.com/electric-sql/pglite/issues/223), but we still do sometimes return an extra, unexpected ReadyForQuery message. Adding back in the patch handles that. Fortunately, I also found what was causing the transaction issues:

- ReadyForQuery messages include 2 parts: a 'Z' (which indicates that this is RFQ), and a 'I' or 'T' or 'R' indicating whether you are currently in a transcation, idle, or rolled back.
- Previously, we injected a RFQ on the first sync after errors (https://github.com/firebase/firebase-tools/blame/5a119439fa66214e5813b503b4fe0cc06939d512/src/emulator/dataconnect/pgliteServer.ts#L193). This RFQ was always ZI, which would be wrong when executing a prepared statement within a transaction (aka something we do all the time).
- 
### Scenarios Tested
I ran a bunch of the scenarios that caused us issues in the past - transactions, rolled back transations, upserts, etc.


joehan:
Gonna merge for now and follow up with more fixes as needed

PR #8794, opened 2025-06-26
Read the new dataDir for exporting behaviors

christhompsongoogle:
Read the new dataDir for exporting behaviors


joehan (approved):

joehan:
Code looks good, but import/export test is failing. Is that expected (and going to be fixed before merging to master)?

christhompsongoogle:
> Code looks good, but import/export test is failing. Is that expected (and going to be fixed before merging to master)?

It might be hitting the same issue as I did - firebase tools is not updating. This runs locally - let me look into this further

PR #8807, opened 2025-07-02
feat: add utilities to support initiatlizing various ai tools for Firebase

taeold:


src/init/features/aitools/promptUpdater.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+    // Stub fs.readFileSync for prompt files
+    readFileSyncStub = sandbox.stub(fs, "readFileSync");
+
+    // Mock prompt file contents
+    readFileSyncStub.withArgs(sinon.match(/FIREBASE\.md$/))
+      .returns(`<firebase_base_context version="0.0.1">
+# Firebase CLI Context
+
+Base Firebase content
+</firebase_base_context>`);
```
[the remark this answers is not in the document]
taeold: no it doesn't. files in prompt/ are eventually wrapped in these xml tags to denote their haash.

src/init/features/aitools/gemini.ts
```diff
@@ -0,0 +1,106 @@
...
+    config: Config,
+    projectPath: string,
+    enabledFeatures: string[],
+  ): Promise<AIToolConfigResult> {
+    const files: AIToolConfigResult["files"] = [];
+
+    // Part 1: Configure the main gemini-extension.json file.
+    const extensionPath = `${GEMINI_DIR}/gemini-extension.json`;
+    const extensionTemplate = readTemplateSync("init/aitools/gemini-extension.json");
+    const extensionConfig = extensionTemplate.replace("{{PROJECT_PATH}}", projectPath);
```
hlshen: nit: newConfigRaw or newRaw to follow existing[raw/Config] pattern

src/init/features/aitools/gemini.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    let extensionUpdated = false;
+    try {
+      const existingRaw = config.readProjectFile(extensionPath);
+      const existingConfig = JSON.parse(existingRaw);
+      const newConfig = JSON.parse(extensionConfig);
+
+      if (!deepEqual(existingConfig, newConfig)) {
```
hlshen, line 56: We're basically going to enforce this exact template to users. Is there no use cases where users will have their own modifications?

src/init/features/aitools/gemini.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+    files.push({ path: extensionPath, updated: extensionUpdated });
+
+    // Part 2: Generate feature-specific context files (e.g., FIREBASE-BASE.md).
+    const baseContent = generateFeaturePromptSection("base");
+    const basePath = `${CONTEXTS_DIR}/FIREBASE-BASE.md`;
+    const baseResult = await replaceFirebaseFile(config, basePath, baseContent);
+    files.push({ path: basePath, updated: baseResult.updated });
+
+    if (enabledFeatures.includes("functions")) {
```
hlshen: Can we move this below and combine the two `if(enabledFeatures)` blocks.

src/init/features/aitools/gemini.ts
```diff
@@ -0,0 +1,106 @@
+import { Config } from "../../../config";
```
hlshen, line 1: So turns out gemini CLI and Gemini Code Assist settings files are different. Can we rename this gemini_cli.ts?

hlshen: chatted offline - seems the extensions are applied to settings.json

PR #8818, opened 2025-07-08
Always set includeDebugDetails to true in VS Code

fredzqm:
### Description

When `includeDebugDetails` in `Impersonation` is set to true, backend will return verbose debug information like SQL error, CEL runtime error.

We want to always set it to true in trusted clients like Firebase Console and VS Code.

Reference: [Backend proto](https://source.corp.google.com/piper///depot/google3/google/firebase/dataconnect/v1main/graphql_service.proto;l=180)


fredzqm:
@hlshen Here is a quick PR~

hlshen (approved):

PR #8820, opened 2025-07-08
Fixed an issue where the FDC emulator did not have ADC provided

### LINKED ISSUE #8819: DataConnect field _embed requires ADC to be set up with gcloud

**firebase-tools:** v14.9.0

**Platform:** macOS 15.5

**Firebase Data Connect extension:** v1.5.0

### [REQUIRED] Test case

https://github.com/aalej/fdc-embed-adc-mcve

Using the test case

1. Run `gcloud auth application-default revoke`
2. Run `gcloud auth revoke`
3. Run `firebase login`
4. Run `firebase use --add PROJECT_ID --alias default`
5. Open the Firebase Data Connnect extension
6. Click the "Start emulators" button
7. Open `./dataconnect/RestaurantEmbedd_insert.gql`, then click "Run (local)"

The VS Code extension makes an ADC using the credentials from `firebase login` and the mutation succeeds

The mutation errors with 
```
failed to build http client for calling Vertex. Error failed to set up an OAuth token source. Error google: could not find default credentials. See https://cloud.google.com/docs/authentication/external/set-up-adc for more information
```

joehan:
### Description
Fixes #8819.

### Scenarios Tested
Tested using the repro provided, and confirmed that we can make a vector call locally.
Screenshot 2025-07-08 at 2 11 47 PM

### DIFF
--- a/src/emulator/controller.ts
+++ b/src/emulator/controller.ts
@@ -603,7 +603,6 @@ export async function startAll(
       );
     }
 
-
     // TODO(b/213241033): Figure out how to watch for changes to extensions .env files & reload triggers when they change.
     const functionsEmulator = new FunctionsEmulator({
       projectId,


aalej:
I think something odd is happening here. When running with `firebase emulators:start --project PROJECT_ID`, the mutation works, but when starting the emulators with the VS Code extension, it does not work.

#### With `firebase emulators:start --project PROJECT_ID` 

It works fine

Screenshot 2025-07-09 at 7 23 07 PM

#### With the "Start emulators" button from Firebase Data Connect VS Code extension

It does not work

Screenshot 2025-07-09 at 7 25 09 PM

aalej:
Arghhh, disregard the previous comment. Turns out the reason why running via VS Code extension does not work is because it runs `npx -y firebase-tools@latest emulators:start --project PROJECT_ID` which does not contain the fix since the fix is only available in the local copy via `npm link` 🤦

aalej (approved):
LGTM! Verified the fix for Data Connect and tested Functions emulator ADC. I think `npm run format` should fix those failing issue?

PR #8824, opened 2025-07-10
Fix top level writeStream breaking functions.

### LINKED ISSUE #8821: uncaughtException: EROFS: read-only file system, open 'pglite-debug.log'

**firebase-tools:**

v14.9.0

**Platform:**

macOS , Ubuntu

### [REQUIRED] Test case

Deploying a cloud function (v1) with firebase-tools v14.9.0 or v14.8.0 results in this error every time the function is called before it even reaches the code

Image

```
uncaughtException: EROFS: read-only file system, open 'pglite-debug.log' Error: EROFS: read-only file system, open 'pglite-debug.log'
Uncaught exception
Error: EROFS: read-only file system, open 'pglite-debug.log'
Error: Process exited with code 16 at process.<anonymous> (/layers/google.nodejs.functions-framework/functions-framework/node_modules/@google-cloud/functions-framework/build/src/invoker.js:96:22) at process.emit (node:events:536:35) at process.emit (node:domain:489:12) at process.exit (node:internal/process/per_thread:184:15) at sendCrashResponse (/layers/google.nodejs.functions-framework/functions-framework/node_modules/@google-cloud/functions-framework/build/src/logger.js:49:9) at process.<anonymous> (/layers/google.nodejs.functions-framework/functions-framework/node_modules/@google-cloud/functions-framework/build/src/invoker.js:85:44) at process.emit (node:events:536:35) at process.emit (node:domain:489:12) at process._fatalException (node:internal/process/execution:188:25)
```

We reproduced the issue on ~50 different cloud functions (all v1), database triggers, cron jobs, pubsub, ... all had the same behavior
We event tries to remove all code and only deploy a console.log in the function content and had the same behabior
The issue went away once we downgraded firebase-tools to v14.7.0 or lower
The issue also wasn't reproducible on cloud functions v2 (cloud run functions)

The error stacktrace doesn't help much but it seems to be caused by https://github.com/firebase/firebase-tools/pull/8750 which added @electric-sql/pglite, which seems to cause some automatic file write on boot which is not allowed on cloud functions v1

Deploy a cloud function v1 using firebase-tools v14.9.0 or v14.8.0
Try to run it

The function to run normally

The function crash with code 16 and a EROFS exception

joehan:
### Description
In a recent PR, I added a top level call to createWriteStream. This is generally no problem, unless you are running in read only file system, in which case it breaks everything. Fixes #8821

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,1 @@
-- Improve App Hosting compute service account flow for source deploys. (#8785)
-- Fixed an issue with `ext:configure` where params without default values could not be set. (#8810)
-- Updated Data Connect emulator to v2.9.0, which:
-  - Fixed an issue in Data Connect where indexes over 63 characters broke schema migration.
-  - Added support for `string_pattern` filters in Data Connect. These allow you to filter string fields using regex or `LIKE` semantics.
-- Fixed an issue where the Data Connect emulator wasn't provided application default credentials. (#8819)
+- Fixed an issue where `firebase-tools` could not be used within v1 Cloud Functions due to trying to write to a read only file.
--- a/src/emulator/dataconnect/pgliteServer.ts
+++ b/src/emulator/dataconnect/pgliteServer.ts
@@ -37,7 +37,6 @@ END
 $do$;`;
 
 const decoder = new StringDecoder();
-const pgliteDebugLog = fs.createWriteStream("pglite-debug.log");
 
 export class PostgresServer {
   private baseDataDirectory?: string;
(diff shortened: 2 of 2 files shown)


taeold (approved):

aalej (approved):
LGTM! Tested deploying functions with these dependencies
```json
  "dependencies": {
    "firebase-admin": "^12.6.0",
    "firebase-functions": "^6.0.1",
    "firebase-tools": "github:firebase/firebase-tools#jh-fix-v1-funcs"
  },
```
  
With code
```js
const functions = require("firebase-functions/v1");
const logger = require("firebase-functions/logger");
const firebaseTools = require("firebase-tools")

exports.helloWorldV1 = functions.https.onRequest((request, response) => {
    logger.info("Hello logs!", { structuredData: true });
    const appsListDesc = firebaseTools.getCommand("apps:list").description()
    logger.log(appsListDesc)
    response.send(`Hello from Firebase! using v1 functions<br>app:list - ` + appsListDesc);
});
```

Invoking function outputs
  
image

PR #8846, opened 2025-07-14
implement mcp logging

antholeole:
### Description

implement MCP logging protocol https://modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging

### Scenarios Tested

I used the MCP inspector. Note that the spec is not well defined in that it does not specify which log level is default. I opened a PR against the inspector to send a default log level (https://github.com/modelcontextprotocol/inspector/pull/611) but this works without it, you just have to explicitly set log level in the inspector


joehan (approved):

joehan:
LGTM, feel free to merge whenever

PR #8848, opened 2025-07-14
feat: Add experimental support to discover functions via dynamic generation of functions.yaml

taeold:
feat: Add experimental file-based discovery via env var

Adds a new environment variable, FIREBASE_FUNCTIONS_DISCOVERY_OUTPUT_PATH, to enable an experimental file-based function discovery mechanism.

- If FIREBASE_FUNCTIONS_DISCOVERY_OUTPUT_PATH=true, a temporary file path will be generated for the discovery manifest (functions.yaml).
- If FIREBASE_FUNCTIONS_DISCOVERY_OUTPUT_PATH is set to a path, that path will be used to store the discovery manifest.

This change also includes a refactoring of the discovery codebase to use the modern fs/promises API instead of util.promisify.


src/deploy/functions/runtimes/node/index.ts
```diff
@@ -208,48 +195,81 @@ export class Delegate {
...
+    });
+
+    childProcess.stdout?.on("data", (chunk: Buffer) => {
+      logger.info(chunk.toString("utf8"));
+    });
+
+    return childProcess;
+  }
+
+  execAdmin(config: backend.RuntimeConfigValues, envs: backend.EnvironmentVariables): ChildProcess {
```
blidd: Could you explain why you are wrapping the spawnFunctionsProcess() method with the execAdmin() method? A comment could be helpful here

src/deploy/functions/runtimes/node/index.ts
```diff
@@ -208,48 +195,81 @@ export class Delegate {
...
+    return this.spawnFunctionsProcess(config, envs);
+  }
+
+  serveAdmin(
+    port: string,
+    config: backend.RuntimeConfigValues,
+    envs: backend.EnvironmentVariables,
+  ): Promise<() => Promise<void>> {
+    const childProcess = this.spawnFunctionsProcess(config, { ...envs, PORT: port });
+    childProcess.stderr?.on("data", (chunk: Buffer) => {
```
blidd: Why are we attaching the stderr listener here in `serveAdmin`, but the stdout listener in `spawnFunctionsProcess`?

src/deploy/functions/runtimes/discovery/index.ts
```diff
@@ -130,3 +128,77 @@ export async function detectFromPort(
...
+  manifestPath: string,
+  project: string,
+  runtime: Runtime,
+  timeout = 10_000,
+): Promise<build.Build> {
+  return new Promise((resolve, reject) => {
+    let stderrBuffer = "";
+    let resolved = false;
+
+    const discoveryTimeout = getFunctionDiscoveryTimeout() || timeout;
```
joehan, line 148: Nit: ?? over ||

src/deploy/functions/runtimes/discovery/index.ts
```diff
@@ -130,3 +128,77 @@ export async function detectFromPort(
...
+      stderrBuffer += chunk.toString();
+    });
+
+    childProcess.on("exit", async (code: number | null) => {
+      if (!resolved) {
+        clearTimeout(timer);
+        resolved = true;
+
+        if (code !== 0 && code !== null) {
+          const errorMessage = stderrBuffer.trim() || `Discovery process exited with code ${code}`;
```
joehan: Same - ?? over ||

src/deploy/functions/runtimes/node/index.ts
```diff
@@ -208,48 +195,81 @@ export class Delegate {
...
+    const env: NodeJS.ProcessEnv = {
+      ...envs,
+      FUNCTIONS_CONTROL_API: "true",
+      HOME: process.env.HOME,
+      PATH: process.env.PATH,
+      NODE_ENV: process.env.NODE_ENV,
+      // Web Frameworks fails without this environment variable
+      __FIREBASE_FRAMEWORKS_ENTRY__: process.env.__FIREBASE_FRAMEWORKS_ENTRY__,
+    };
+    if (Object.keys(config || {}).length) {
```
joehan, line 223: Based on types, this || {} shouldn't be necessary. Either remove or fix the arg type?

joehan (approved):
LGTM, thanks for adding integration test coverage

src/deploy/functions/runtimes/node/index.ts
```diff
@@ -280,13 +300,37 @@ export class Delegate {
...
-      const port = await portfinder.getPortPromise({ port: basePort });
-      const kill = await this.serveAdmin(port.toString(), config, env);
-      try {
-        discovered = await discovery.detectFromPort(port, this.projectId, this.runtime);
-      } finally {
-        await kill();
+      const discoveryPath = process.env.FIREBASE_FUNCTIONS_DISCOVERY_OUTPUT_PATH;
+      if (discoveryPath) {
+        let manifestPath: string;
+        if (discoveryPath === "true") {
```
blidd: *nit: we could potentially avoid another nested if-else if we did something like

```
if (!discoveryPath) {
  ...
} else if (discoveryPath === "true") {
  ...
} else {
  // discoveryPath has a custom path specified
  ...
}
```

This is mostly just a subjective preference though

src/deploy/functions/runtimes/node/index.ts
```diff
@@ -208,48 +195,81 @@ export class Delegate {
...
+    port: string,
+    config: backend.RuntimeConfigValues,
+    envs: backend.EnvironmentVariables,
+  ): Promise<() => Promise<void>> {
+    const childProcess = this.spawnFunctionsProcess(config, { ...envs, PORT: port });
+    childProcess.stderr?.on("data", (chunk: Buffer) => {
+      logger.error(chunk.toString("utf8"));
+    });
+
+    return Promise.resolve(async () => {
```
blidd, line 269: The nested promises are a bit difficult to grok. Does `serveAdmin` need to be an async function? Could we directly return the kill function, i.e. `serveAdmin(...): () => Promise<void>`?

src/deploy/functions/runtimes/node/index.ts
taeold: For the config || {} thing - might need it since config comes from runtime/API responses that could be null despite the types.
 
Added a comment explaining why we're being defensive here.

src/deploy/functions/runtimes/node/index.ts
taeold: I agree is getting a bit narly, but I'd like to refactor this later as a separate change to avoid unintetional breakage for now. left a todo.

src/deploy/functions/runtimes/node/index.ts
taeold: fixed inconsistency.

src/deploy/functions/runtimes/node/index.ts
taeold: i like this

src/deploy/functions/runtimes/node/index.ts
taeold: yep added a comment

PR #8854, opened 2025-07-15
[MCP] implement {get,set}_data and {validate,get}_rules for RTDB MCP

antholeole:


src/mcp/tools/database/get_data.ts
```diff
@@ -0,0 +1,63 @@
...
+export const get_data = tool(
+  {
+    name: "get_data",
+    description: "Returns RTDB data from the specified location",
+    inputSchema: z.object({
+      databaseUrl: z
+        .string()
+        .optional()
+        .describe(
+          "connect to the database at url. If omitted, use default database instance <project>-default-rtdb.firebaseio.com. Can point to emulator URL (e.g. localhost:6000/<instance>)",
```
nguymichael: nit: the domain is different than the one actually used.

I don't feel strongly either way, but should we force the input to specify the full url? Regardless default-rtdb + us-central1 fit a majority of cases.

src/mcp/tools/database/get_data.ts
```diff
@@ -0,0 +1,63 @@
...
+  async ({ path, databaseUrl }, { projectId, host }) => {
+    if (!path.startsWith("/")) {
+      return mcpError(`paths must start with '/' (you passed '${path}')`);
+    }
+
+    const dbUrl = new url.URL(
+      databaseUrl
+        ? `${databaseUrl}/${path}.json`
+        : `https://${projectId}-default-rtdb.us-central1.firebasedatabase.app/${path}.json`,
+    );
```
nguymichael, line 48: nit: path is already prefixed with `/` (though the url should work anyways). If we're using node: path.join?

src/mcp/tools/database/validate_rules.ts
```diff
@@ -0,0 +1,50 @@
...
+  async ({ databaseUrl, rules }, { projectId, host }) => {
+    const dbUrl =
+      databaseUrl ?? `https://${projectId}-default-rtdb.us-central1.firebasedatabase.app`;
+
+    const client = new Client({ urlPrefix: dbUrl });
+
+    try {
+      await updateRulesWithClient(client, rules, { dryRun: true });
+    } catch (e: unknown) {
+      host.logger.debug(`failed to update rules at url ${dbUrl}`);
```
nguymichael: nit: failed to validate

src/mcp/tools/database/get_data.ts
antholeole: > I don't feel strongly either way, but should we force the input to specify the full url? Regardless default-rtdb + us-central1 fit a majority of cases.

I'm confused by the question here. true that i used another URL then I said I used, updated that. not sure what you mean by ...should we force the input...

nguymichael: I meant required vs. optional

antholeole: ahhhh i see. I would prefer it optional but if ur taking a hard stance we can make it required. reason being is that an llm is calling these tools -  a human might be able to figure out their project name pretty trivially, but it might require another tool call from the LLM. these tools are supposed to be as low friction as possible imo

nguymichael: My stance here is pretty soft -- slightly toward the optional side. I only mention since I'm not sure how the LLM may handle two modes of inputs (i.e. empty v. specify). But again -- I haven't used MCP before so this may really just be a non-issue.

PR #8855, opened 2025-07-16
adhoc queries and mutations support list and enum

hlshen:
Added unit tests for mutation creation.

Minimal change to queries to support lists. Future revamp to utilize live schema is planned.


firebase-vscode/src/data-connect/ad-hoc-mutations.ts
```diff
@@ -258,6 +195,95 @@ query {
...
+export function makeAdHocMutation(
+  fields: GraphQLInputField[],
+  singularName: string,
+): OperationDefinitionNode {
+  const argumentFields: ObjectFieldNode[] = [];
+
+  for (const field of fields) {
+    const type = getNamedType(field.type);
+    let defaultValue: ValueNode | undefined;
+    if (isEnumType(type)) {
```
hlshen, line 206: 207-211, and 220-222 are the changes. Needed to move this function out to export to tests.

firebase-vscode/src/data-connect/ad-hoc-mutations.ts
```diff
@@ -67,9 +67,12 @@ export function registerAdHoc(
       var hasField = false;
       let query = "{\n";
       for (const field of ast.fields!) {
-        // We unwrap NonNullType to obtain the actual type
+        // We unwrap NonNullType and ListType to obtain the actual type
         let fieldType = field.type;
-        if (fieldType.kind === Kind.NON_NULL_TYPE) {
+        while (
```
stephenarosaj, line 72: why `while` instead of `if`?

hlshen: Could be a list of non_null(required) types.

firebase-vscode/src/data-connect/ad-hoc-mutations.ts
```diff
@@ -258,6 +195,95 @@ query {
...
+      return { kind: Kind.BOOLEAN, value: false };
+    case "Date":
+      return {
+        kind: Kind.STRING,
+        value: new Date().toISOString().substring(0, 10),
+      };
+    case "Float":
+      return { kind: Kind.FLOAT, value: "0" };
+    case "Int":
+      return { kind: Kind.INT, value: "0" };
```
joehan, line 271: Should Float and Int default values actually be quoted? Seems wrong to me but maybe im missing something

firebase-vscode/src/test/suite/src/dataconnect/ad-hoc-mutation.test.ts
```diff
@@ -0,0 +1,153 @@
...
+  test("should generate a mutation with a different singular name", () => {
+    const fields = [mockFields.string];
+    const singularName = "Item";
+    const result = makeAdHocMutation(fields, singularName);
+    const printedResult = print(result).replace(/\s/g, "");
+
+    const expected = 'mutation{item_insert(data:{field1:""})}';
+    assert.strictEqual(printedResult, expected);
+  });
+});
```
joehan: Newline at end of file

joehan (approved):
Small things, but lgtm after that

firebase-vscode/src/data-connect/ad-hoc-mutations.ts
hlshen: IntValueNode and FloatValueNode both take string values

PR #8857, opened 2025-07-17
Remove an unnecessary includeDebugDetails

fredzqm:
It was in the wrong place. `includeDebugDetails` should be inside `impersonate`.


fredzqm:
@hlshen not so important, but would be nice to get a stamp on this one line PR. 😄

Issue #8758, opened 2025-06-15
__name__ beeing stripped from firestore:indexes cmd


glumb:
**Why it matters**
The issue is that some queries require a non standard `__name__` sorting, like
```ts
const query = await db.collection("Auditlog")
  .where("tenantID", "in", ["IeZfE0yKWs0kGBl1IG5M"])
  .orderBy("_meta.dateCreated", "desc")
  .limit(50)
  .get();
```

wants me to create this index (maybe this is another bug? as I read that the __name__ shall automatically sort according to the last sort in the index)

![Image](https://github.com/user-attachments/assets/93cbdddd-aed9-4f61-b03e-7afc1d72c5b7)

When I now create this index online in the firebase console, download the indexes via the client and redeploy them to another project, the `__name__` sorting information got lost, as its now creating an index sorting by DESCENDING for `__name__`

### [REQUIRED] Environment info

**firebase-tools:** 14.7.0

**Platform:** macOS

### [REQUIRED] Test case

just call `firebase firestore:indexes`. the `__name__` field is stripped

### [REQUIRED] Steps to reproduce

1.  have an index in firestore
2. call `firebase firestore:indexes`. 
3. the `__name__` field is stripped 

### [REQUIRED] Expected behavior

`__name__` is present

### [REQUIRED] Actual behavior

the `__name__` field is stripped due to https://github.com/firebase/firebase-tools/blob/843b6b4d8ac5e523cd5346007a2bffb4b0076ca5/src/firestore/api.ts#L187-L191

aalej:
Hey @glumb, thanks for the detailed report and for opening up a PR for this! Let me know if I may be wrong here, but I looked around other repositories and I think the code snippet you're using is something like:
```js
const query = await db.collection("Auditlog")
           .where("tenantID", "in", ["IeZfE0yKWs0kGBl1IG5M"])
           .orderBy("_meta.dateCreated", "desc")
           .orderBy('tenantID', 'asc')
           .limit(50)
           .get();
```
Which is similar to the code you provided in https://github.com/firebase/firebase-js-sdk/issues/9097.

I was able to reproduce the issue using the code snippet you provided in the JS SDK repository. The generated index looks like

![Image](https://github.com/user-attachments/assets/362f0607-c588-4c4d-a8db-21ac596655ae)

Running `firebase firestore:indexes` outputs
```json
{
  "indexes": [
    {
      "collectionGroup": "Auditlog",
      "queryScope": "COLLECTION",
[...]

PR #8860, opened 2025-07-17
fix(firestore): smart filtering of __name__ fields in index listings

### LINKED ISSUE #7629: Incorrect firebase firestore:indexes output generated

**firebase-tools:** 13.6.0

**Platform:** macOS

### [REQUIRED] Test case

Given 2 composite indexes, created via the `FAILED_PRECONDITION` provided link

Screenshot 2024-09-07 at 06 33 54

Run the `firebase firestore:indexes` CLI command to output the indexes. As you will see, the indexes are duplicated, because the `__name__` is stripped off https://github.com/firebase/firebase-tools/blob/master/src/firestore/api.ts#L187

```
    {
      "collectionGroup": "random-collection",
      "queryScope": "COLLECTION_GROUP",
      "fields": [
        {
          "fieldPath": "a",
          "order": "ASCENDING"
        },
        {
          "fieldPath": "b",
          "order": "DESCENDING"
        }
      ]
    },
    {
      "collectionGroup": "random-collection",
      "queryScope": "COLLECTION_GROUP",
      "fields": [
        {
          "fieldPath": "a",
          "order": "ASCENDING"
        },
        {
          "fieldPath": "b",
          "order": "DESCENDING"
        }
      ]
    },
```

The `__name__` to also be included, as that is the difference between both

 
The current duplication is a problem because when you take the index file from the sandbox env and deploy it to production, you think all the indexes are created, but in fact only one of the 2 is. And thus you reach a `FAILED_PRECONDITION`  on production

While this example is exaggerated in the sense that probably you would not have 2 indexes like this, the bigger issue is if you have an index with a `__name__` order that is not exactly as the default sort order, then it's not properly generated when moving from sandbox to production

TorbenWetter:
## Problem

The recent fix for #7629 introduced a regression where `firebase deploy --only firestore` fails with "409 index already exists" errors. The original fix was too aggressive, preserving ALL `__name__` fields instead of filtering intelligently.

## Root Cause

The previous implementation completely removed the filtering logic for `__name__` fields. While this fixed the duplicate index issue (#7629), it introduced deployment conflicts because:

1. The API returns indexes with explicit `__name__` fields
2. User spec files typically don't include implicit `__name__` fields  
3. The comparison logic treated these as different indexes
4. Deployment attempted to create "new" indexes that already existed

## Solution

This PR implements smart filtering that:
- Keeps `__name__` fields with DESCENDING order (explicit/meaningful)
- Removes `__name__` fields with ASCENDING order (implicit/default)
- Fixes both duplicate index issues (#7629) and deployment conflicts (#8859)

## Implementation

The fix modifies the `listIndexes` method in `src/firestore/api.ts` to filter `__name__` fields intelligently:

```typescript
const fields = index.fields.filter((field: types.IndexField) => {
  if (field.fieldPath !== "__name__") {
    return true;
  }
  return field.order === types.Order.DESCENDING;
});
```

## Changes

- Enhanced `listIndexes` method with intelligent `__name__` field filtering
- Added 5 comprehensive test cases covering all scenarios
- Updated changelog with user-facing documentation

## Testing

- All existing tests pass with no regressions
- New test suite `IndexListingWithNameFields` validates the fix
- Covers filtering, preservation, and edge cases
- Handles various index configurations correctly

## Impact

This change is backward compatible and isolated to index listing logic. It resolves the critical deployment failures in v14.11.0 while maintaining the semantic meaning of indexes.

Fixes #7629, #8859


aalej:
Thanks for opening up this PR @TorbenWetter!

I'm testing the changes and it seems like the error is still occurring. Tried with indexes

`firestore.indexes.json`

```json
{
  "indexes": [
    {
      "collectionGroup": "widgets",
      "queryScope": "COLLECTION",
      "fields": [
        { "fieldPath": "foo", "arrayConfig": "CONTAINS" },
        { "fieldPath": "bar", "mode": "DESCENDING" }
      ]
    },
    {
      "collectionGroup": "widgets",
      "queryScope": "COLLECTION",
      "fields": [
        { "fieldPath": "foo", "arrayConfig": "CONTAINS" },
        { "fieldPath": "bar", "mode": "ASCENDING" }
      ]
    }
  ]
}
```

1. Run `firebase deploy --only firestore:indexes`

```
$ firebase deploy --only firestore:indexes

=== Deploying to 'PROJECT_ID'...

i  deploying firestore
i  firestore: reading indexes from firestore.indexes.json...
i  cloud.firestore: checking firestore.rules for compilation errors...
✔  cloud.firestore: rules file firestore.rules compiled successfully
i  firestore: deploying indexes...
✔  firestore: deployed indexes in firestore.indexes.json successfully for (default) database

✔  Deploy complete!
```

2. Wait a couple of minutes to let the indexes build
   - Status can be checked in https://console.firebase.google.com/project/PROJECT_ID/firestore/databases/-default-/indexes
3. Run `firebase deploy --only firestore:indexes` again

```
$ firebase deploy --only firestore:indexes

=== Deploying to 'PROJECT_ID'...

i  deploying firestore
i  firestore: reading indexes from firestore.indexes.json...
i  cloud.firestore: checking firestore.rules for compilation errors...
✔  cloud.firestore: rules file firestore.rules compiled successfully
i  firestore: deploying indexes...
i  firestore: The following indexes are defined in your project but are not present in your firestore indexes file:
        (widgets) -- (foo,CONTAINS) (bar,DESCENDING)
✔ Would you like to delete these indexes? Selecting no will continue the rest of the deployment. No

Error: Request to https://firestore.googleapis.com/v1/projects/PROJECT_ID/databases/(default)/collectionGroups/widgets/indexes had HTTP Error: 409, index already exists
```

CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- **firestore**: Fixed issue where `__name__` fields with DESCENDING order were incorrectly filtered from index listings, causing duplicate index issues (#7629) and deployment conflicts (#8859). The fix now preserves `__name__` fields with explicit DESCENDING order while filtering out implicit ASCENDING `__name__` fields.
```
joehan:
```suggestion
- Fixed issue where `__name__` fields with DESCENDING order were incorrectly filtered from index listings, causing duplicate index issues (#7629) and deployment conflicts (#8859). The fix now preserves `__name__` fields with explicit DESCENDING order while filtering out implicit ASCENDING `__name__` fields.
```

joehan (approved):
Taking a closer look at the issue @aalej found.

joehan:
Hmmm, it seems like the behavior here is tricker than we initially realized - in the example @aalej shared, the 'implied' direction for __name__ is decending when the other fields are descending
Screenshot 2025-07-17 at 9 22 09 AM
I'm gonna test out some minor changes to this PR to make it follow that behavior, and check with someone on the firestore team to confirm this behavior.

joehan:
Ahhh, turns out we documented it - https://firebase.google.com/docs/firestore/query-data/index-overview#default_ordering_and_the_name_field the default ordering follows the ordering of the last field in the index.

aalej:
I think in issue https://github.com/firebase/firebase-tools/issues/7629, there's an instance where the default ordering for `__name__` did not follow the last field in the index. 

There's also a case where the indexes deployed changed, here's a repro - https://github.com/firebase/firebase-tools/issues/8758#issuecomment-2992044133

PR #8871, opened 2025-07-19
fix(mcp): Address auth issues with Firebase Studio environment

mbleigh:
- Removes auth check from list tools call when in Firebase Studio
- Adds timeouts to feature checks and auth generally
- Centralized place to check environment stuff instead of messing with variables
- Creates a special `FIREBASE_MCP_DEBUG_LOG` variable that when used will log the entire MCP transcript to the specified file.

### DIFF
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -240,6 +240,13 @@ export class FirebaseMcpServer {
     }
   }
 
+  async filteredAvailableTools(): Promise<ServerTool[]> {
+    const excludedNames: string[] = [];
+    // exclude create_project from firebase studio since there's a better way there
+    if (isFirebaseStudio()) excludedNames.push("firebase_create_project");
+    return this.availableTools.filter((t) => !excludedNames.includes(t.mcp.name));
+  }
+
   async mcpListTools(): Promise<ListToolsResult> {
     await Promise.all([this.detectActiveFeatures(), this.detectProjectRoot()]);
     const hasActiveProject = !!(await this.getProjectId());
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -240,13 +240,6 @@ export class FirebaseMcpServer {
     }
   }
 
-  async filteredAvailableTools(): Promise<ServerTool[]> {
-    const excludedNames: string[] = [];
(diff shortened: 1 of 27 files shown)


src/requireAuth.ts
```diff
@@ -104,6 +116,8 @@ export async function requireAuth(options: any): Promise<string | null> {
     );
   } else if (user && (!isExpired(tokens) || tokens?.refresh_token)) {
     logger.debug(`> authorizing via signed-in user (${user.email})`);
+  } else if (skipAutoAuth) {
```
joehan, line 119: We have monospace specific logic in autoAuth that sets the user from the returned access token as the active user (so that it appears like the user is logged in). I'm concerned about the case where these tokens are present but expired - MCP server will tell them to login, but I think `firebase login`  might not work as expected.

I think we probably need to clean up the auth strategy overall ASAP - what you have here is probably still an improvement over the current state tho

src/requireAuth.ts
```diff
@@ -44,13 +47,19 @@ async function autoAuth(options: Options, authScopes: string[]): Promise<null |
 
   let clientEmail;
   try {
-    const credentials = await client.getCredentials();
+    const timeoutSeconds = isFirebaseMcp() ? 5000 : 15000; // shorter timeout for MCP
```
joehan: Clean up unused or pass it into the timeout func

src/mcp/index.ts
```diff
@@ -204,28 +228,37 @@ export class FirebaseMcpServer {
...
     } catch (e) {
+      this.log("debug", `error in requireAuth: ${e}`);
       return null;
     }
   }
 
+  async filteredAvailableTools(): Promise<ServerTool[]> {
+    const excludedNames: string[] = [];
+    // exclude create_project from firebase studio since there's a better way there
+    if (isFirebaseStudio()) excludedNames.push("firebase_create_project");
```
joehan: Can you also exclude the firebase_app_create tool?

joehan: Update: Gonna do this within the prototyper in studio by filtering the listTools result. This whole chnk of logic can be removed

PR #8878, opened 2025-07-22
Rename tests

aalej:
### Description

Renamed a couple of test files since they are getting included in the build
image
image

### Scenarios Tested

`npm run test`


joehan (approved):
Good catch!

PR #8879, opened 2025-07-22
Configure apptesting

jakeouellette:
### Description

Simplify setup of App Testing: When running `firebase init apptesting`, onboard onto storage, cloud run, and artifact registry. Additionally, setup the initial service account used.

### Scenarios Tested

Manually tested running this against the three scenarios:
* A project that had not been setup before
* A project that had everything, but the role was not added to the service account
* A project that was already set up

### Sample Commands

No change, but the command is `firebase init apptesting`, it will ask one additional question now:

✔ Firebase App Testing runs tests in Cloud Run using a service account, create the account? Yes


src/apptesting/ensureProjectConfigured.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (!serviceAccountExistsAndIsRunner) {
+    const grant = await confirm(
+      "Firebase App Testing runs tests in Cloud Run using a service account, create the account?",
+    );
+    if (!grant) {
+      logBullet(
+        "You, or your project administrator, should run the following command to grant the required role:\n\n" +
+          `\tgcloud projects add-iam-policy-binding ${projectId} \\\n` +
+          `\t  --member="serviceAccount:${serviceAccount} \\\n` +
+          `\t  --role="${TEST_RUNNER_ROLE}\n`,
```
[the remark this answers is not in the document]
jakeouellette: This removes the newline, but it did catch the missing quota, not applying it, but adding the quote properly.

src/apptesting/ensureProjectConfigured.spec.ts
```diff
@@ -0,0 +1,144 @@
+import * as ensureApiEnabled from "../ensureApiEnabled";
+import * as iam from "../gcp/iam";
+import * as sinon from "sinon";
```
kjelko, line 1: external imports (sinon and chai) should go first

src/apptesting/ensureProjectConfigured.ts
```diff
@@ -0,0 +1,77 @@
...
+import * as iam from "../gcp/iam";
+import { confirm } from "../prompt";
+
+const TEST_RUNNER_ROLE = "roles/firebaseapptesting.testRunner";
+const TEST_RUNNER_SERVICE_ACCOUNT_NAME = "firebaseapptesting-test-runner";
+
+export async function ensureProjectConfigured(projectId: string) {
+  await ensure(projectId, appTestingOrigin(), "storage", false);
+  await ensure(projectId, appTestingOrigin(), "run", false);
+  await ensure(projectId, appTestingOrigin(), "artifactregistry", false);
```
kjelko, line 15: Does the customer project actually need artifactregistry if they're just reading from our public one?

src/apptesting/ensureProjectConfigured.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const serviceAccountExistsAndIsRunner = await serviceAccountHasRoles(
+    projectId,
+    serviceAccount,
+    [TEST_RUNNER_ROLE],
+    true,
+  );
+  if (!serviceAccountExistsAndIsRunner) {
+    const grant = await confirm(
+      "Firebase App Testing runs tests in Cloud Run using a service account, create the account?",
```
kjelko: should we mention the service account name here?

src/init/features/apptesting/index.ts
```diff
@@ -23,6 +24,9 @@ export async function askQuestions(setup: Setup): Promise<void> {
         })),
     },
   };
+  if (setup.projectId) {
```
kjelko, line 27: is it possible this wouldn't be set? what happens if that's the case?

kjelko (commented):
nice! mostly just small nits

src/apptesting/ensureProjectConfigured.ts
jakeouellette: According to this, https://cloud.google.com/artifact-registry/docs/integrate-cloud-run#permissions, but not sure if it becomes untrue if it is actually a public repository. It fixed an error I had, though.

src/apptesting/ensureProjectConfigured.ts
jakeouellette: added

src/init/features/apptesting/index.ts
jakeouellette: I tried to do this, and I think I need an empty account, because it always seems to fall back to a valid project.  That said, this is prevalent (with no throws) in many init calls. I'm happy to defer to @joehan  on best practices here

src/apptesting/ensureProjectConfigured.ts
joehan: Yes, the API must be enabled if the call is being made with creds from the user project. Since it fixed an issue, I suspect this is the case on the backend

jakeouellette:
It seems the broken CI test is unrelated to us. (is about a firebase refresh command, seems to be happening in other prs)

src/init/features/apptesting/index.ts
joehan: Yeah, projectId can be not set if you run 'firebase init' and then choose 'dont set up a project yet'. I think your approach here follows best practices tho - when no project is set, we just want to set up local files and skip any provisioning/API enablement

PR #8881, opened 2025-07-22
fix: fix ext:export command

huangjeff5:
The command currently doesn't export system params at all. The params are already exposed from the backend. We just need to merge them into the created .env.


joehan (approved):
Add a change log entry, but lgtm otherwise!

PR #8888, opened 2025-07-23
Adding artifactsOnly mode and fixing Docker image build

joehan:
### Description

Making some improvements so that releases are smoother in the future.

- Added an artifactsOnly mode, which skips the package publish and just publishes Firepit artifacts to Github and the Docker image. Both of these actions are effectively idempotent
- Fixed an issue where the Docker image build script was running with the wrong Docker context and couldn't find package.json/package-lock.json


scripts/publish/cloudbuild.yaml
```diff
@@ -141,7 +141,7 @@ options:
       path: /root/.ssh
 
 substitutions:
-  _VERSION: ""
+  # _VERSION: ""
```
hlshen: what's this change for?

hlshen (changes requested):
AI caught some quote issues

scripts/publish/cloudbuild.yaml
joehan: Derp, it was accidental

PR #8903, opened 2025-07-28
Fixing an issue where MCP server would still hang on studio

joehan:
### Description
More Studio MCP fixes - if we're gonna not use ADC on Studio, we need to be consistent about it.

### Scenarios Tested
Waiting for a workspace to warm up to test this properly.


fredzqm (approved):

joehan:
Screenshot 2025-07-29 at 11 37 55 AM
Tested on studio and works like a charm

PR #8904, opened 2025-07-29
Fetch project from studio

samedson:
### Description

Features:
 - Implements a way to fetch the active Firebase Project when running the CLI inside of Firebase Studio so that they sync up. When Studio and the CLI disagree, implements a way for the user to reconcile them. Note that the `--project` parameter will still work as an override.
 - Updates `firebase use` to update the active project in Firebase Studio.

Progress:

- [x] Fetch the active project from Firebase Studio
- [x] Give the user a way to reconcile active projects
- [x] Write the active project to configstore
- [x] Update `firebase use` to update the active project
- [x] Provide a way to signal to Firebase Studio that the active project has been updated (eg. writing to configstore or a separate file)
- [x] Is it wonky if there's a conflict and you run `firebase use <new-project>`? You go through conflict resolution and then could set the project to a third project (maybe we skip conflict resolution if you are calling the `use` command)?

### Scenarios Tested

```
MONOSPACE_ENV=1 WORKSPACE_SLUG="<studio-workspace-id>" firebase apps:list --project <another-project-id>
✔ Preparing the list of your Firebase apps

... list of apps from <another-project-id>
```

### Sample Commands

```
$ MONOSPACE_ENV=1 WORKSPACE_SLUG="<studio-workspace-id>" firebase apps:list
? Found different active Firebase Projects in the Firebase CLI and your Firebase Studio 
Workspace. Which project would you like to set as your active project?
  Set <studio-project-id> from Firebase Studio as my active project in both places
❯ Set <cli-project-id> from Firebase CLI as my active project in both places
Updating Studio Workspace active project to match Firebase CLI '<cli-project-id>'
✔ Preparing the list of your Firebase apps

... list of apps from <cli-project-id>
```


src/management/studio.ts
```diff
@@ -0,0 +1,158 @@
...
+      message += ` (original: ${err.original.message})`;
+    }
+    logger.error(`Failed to fetch Firebase Project from current Studio Workspace: ${message}`);
+    // We're going to fail gracefully so that the caller can handle the error
+    return undefined;
+  }
+}
+
+async function writeStudioProjectToConfigStore(options: Options, studioProjectId: string) {
+  if (options.projectRoot) {
```
samedson, line 104: I was curious what the desired behavior should be without a project root. This is when there's no firebase.json in your directory chain?

I did a quick sanity check and noticed there's different behavior when you have a firebase.json and when you don't:

```
➜  firebase-tools git:(fetch-project-from-studio) MONOSPACE_ENV=1 WORKSPACE_SLUG="studio-48028039" firebase apps:list
✔ Preparing the list of your Firebase apps
No apps found.

➜  firebase-tools git:(fetch-project-from-studio) touch firebase.json

➜  firebase-tools git:(fetch-project-from-studio) ✗ MONOSPACE_ENV=1 WORKSPACE_SLUG="studio-48028039" firebase apps:list
? Found different active Firebase Projects in the Firebase CLI and your Firebase Studio 
Workspace. Which project would you like to set as your active project? (Use arrow keys)
❯ Set greeting-generator-spibe from Firebase Studio as my active project in both places
  Set dumpsterfire-6 from Firebase CLI as my active project in both places
  
  ➜  firebase-tools git:(fetch-project-from-studio) ✗ rm -f firebase.json                      
  
➜  firebase-tools git:(fetch-project-from-studio) MONOSPACE_ENV=1 WORKSPACE_SLUG="studio-48028039" firebase use dumpsterfire-5

Error: firebase use must be run from a Firebase project directory.

Run firebase init to start a project directory in the current folder.
```

My hunch is a lot of apps won't have a firebase.json for their studio project.

So maybe the user experience is:
  - ✅  If you haven't called firebase use, then we use the studio project, no conflict resolution.
  - ✅  If you have called firebase use, it'll force you through the `init` flow, which creates a firebase.json
  
 So lots of words to say, this seems fine to me, but what do you think?

src/management/studio.ts
```diff
@@ -0,0 +1,158 @@
...
+    logger.error(
+      `Failed to update active Firebase Project for current Studio Workspace: ${message}`,
+    );
+  }
+  recordStudioProjectSyncTime();
+}
+
+/**
+ * Records the last time we synced the Studio project in Configstore.
+ * Conviently, this triggers the file watcher.
```
samedson: Suggestion: "This line is important to trigger a file watcher in Firebase Studio that keeps the UI in sync"

I think a future reader will may not have context

samedson (commented):
LGTM! Just one suggestion

src/management/studio.ts
joehan: Yeah, this was pretty much my thinking as well when I noticed this edge case - if you're not in a Firebase project directory, we can't set your active project, we just pick up your studio project.

joehan (approved):

src/management/studio.ts
joehan: Good call, this was too vague.

samedson:
Just rebased onto master FYI to see if it fixes the test failures

PR #8911, opened 2025-07-30
feat: Add support for function prefixes

taeold:
This change adds a new `prefix` property to firebase functions configuration in firebase.json. The new property can be used to prepend function name with the given prefix for all functions in the codebase.

Note that prefix is also used for all secret associated with the codebase to allow different codebases to refer to different configuration value.

The new feature also requires two new validation rules for function configurations in `firebase.json`:

1.  The `prefix` property, if specified, must only contain lowercase letters, numbers, and hyphens.
2.  The combination of `source` and `prefix` must be unique across all function configurations.

These changes ensure that function deployments with prefixes are valid and that there are no conflicts when deploying multiple functions from the same source directory.


src/deploy/functions/prepare.ts
```diff
@@ -155,7 +155,7 @@ export async function prepare(
     }
 
     for (const endpoint of backend.allEndpoints(wantBackend)) {
-      endpoint.environmentVariables = { ...wantBackend.environmentVariables } || {};
+      endpoint.environmentVariables = { ...(wantBackend.environmentVariables || {}) };
```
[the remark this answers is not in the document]
inlined: I'm not sure what Gemini is saying. This probably could just be `{ ...wantBackend.environmentVariables }` (no backup). `...undefined` in an object destructure yields nothing. Frustratingly though it throws in an array destructure.

src/deploy/functions/build.ts
```diff
@@ -651,3 +651,42 @@ function discoverTrigger(endpoint: Endpoint, region: string, r: Resolver): backe
...
+ * Prefixes all endpoint IDs and secret names in a build with a given prefix.
+ * This ensures that functions and their associated secrets from different codebases
+ * remain isolated and don't conflict when deployed to the same project.
+ */
+export function applyPrefix(build: Build, prefix: string): void {
+  if (!prefix) {
+    return;
+  }
+  const newEndpoints: Record<string, Endpoint> = {};
+  for (const id of Object.keys(build.endpoints)) {
```
inlined: Suuuper nit that you can disregard. You could also have said `for (const [id, endpoint] of Object.entries(build.endpoints))`

src/emulator/functionsEmulator.ts
```diff
@@ -220,6 +222,7 @@ export class FunctionsEmulator implements EmulatorInstance {
 
   private staticBackends: EmulatableBackend[] = [];
   private dynamicBackends: EmulatableBackend[] = [];
+  private watchers: chokidar.FSWatcher[] = [];
```
inlined, line 225: What is this? It doesn't seem to be used anywhere

src/deploy/functions/prepare.spec.ts
```diff
@@ -16,14 +20,86 @@ describe("prepare", () => {
     region: "region",
     project: "project",
     entryPoint: "entry",
-    runtime: "nodejs16",
+    runtime: "nodejs22",
```
inlined: nit: this my be more evergreen to replace with `latest("nodejs")`

src/emulator/functionsEmulator.ts
taeold: It's used later to add the watcher we create for functions emulator.

Learned that our integration test never cleans up the watcher, leaving the process hanging.

PR #8928, opened 2025-08-02
Add `product_init_mcp` and fix param of `product_init`

fredzqm:
In Google Analytics, `products_initialized` is always empty because the `init()` shifts it to an empty list in the end.

Want to figure out how many time `init dataconnect` was called.


fredzqm:
@joehan Please take a look~ Simple PR to fix init's GA metrics.

I was trying to figure out how many people ran `firebase init dataconnect`

joehan (approved):

PR #8933, opened 2025-08-04
Prototype adding Setup Instructions to Firebase MCP

samedson:
https://screencast.googleplex.com/cast/NjA2NjA0NjM5Mjg2MDY3Mnw1ZWI4YjYzYS1kNg


joehan:
Michael added support for prompts in https://github.com/firebase/firebase-tools/pull/8965 - this might be more appropriate as prompt instead of a tool

PR #8947, opened 2025-08-06
[FDC] `firebase init dataconnect:sdk` don't need confirmation to update connector.yaml

fredzqm:
Simplify it to only ask one question about "App Platform"

### Before

Screenshot 2025-08-06 at 1 18 27 PM

### After

Screenshot 2025-08-06 at 1 15 32 PM

We can make the one-linter script set `_FDC_APP_FOLDER`, so the cmd would skips the "app folder question"

### DIFF
--- a/src/init/features/dataconnect/sdk.ts
+++ b/src/init/features/dataconnect/sdk.ts
@@ -249,11 +249,9 @@ export async function generateSdkYaml(
 export async function actuate(sdkInfo: SDKInfo, config: Config) {
   const connectorYamlPath = `${sdkInfo.connectorInfo.directory}/connector.yaml`;
   logBullet(`Writing your new SDK configuration to ${connectorYamlPath}`);
-  await config.askWriteProjectFile(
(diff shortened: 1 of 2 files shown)


src/init/features/dataconnect/sdk.ts
```diff
@@ -68,17 +68,8 @@ async function askQuestions(setup: Setup, config: Config): Promise<SDKInfo> {
   }
 
   // First, lets check if we are in an app directory
-  let appDir = process.env[FDC_APP_FOLDER] || process.cwd();
+  const appDir = process.env[FDC_APP_FOLDER] || process.cwd();
   let targetPlatform = await getPlatformFromFolder(appDir);
-  if (targetPlatform === Platform.NONE && !process.env[FDC_APP_FOLDER]?.length) {
```
maneesht: I'm not sure that we can just assume that wherever the firebase.json is, that that's where the user's app is going to be as well

yuchenshi (approved):

PR #8948, opened 2025-08-06
update emulators to 2.11.0

hlshen:
- Updated the Firebase Data Connect local toolkit to v2.11.0, which includes the following changes:
  - [Fixed] Kotlin code generation with enums
  - [Fixed] Deploying schemas with enums would report false breaking changes
  - [Added] Support for ordering results by aggregate fields (min/max/count/sum/avg).
  - [Added] Support for built-in and user-defined enums to LLM Tools generation.

### DIFF
--- a/src/emulator/downloadableEmulatorInfo.json
+++ b/src/emulator/downloadableEmulatorInfo.json
@@ -54,28 +54,28 @@
   },
   "dataconnect": {
     "darwin": {
-      "version": "2.10.1",
-      "expectedSize": 29336416,
-      "expectedChecksum": "328ae0a354505430f1c162edf37da776",
-      "expectedChecksumSHA256": "ede5dc299045151f228b2adc346f47ad5d3ee10c3c9d7528df5fc0f0c6d70808",
-      "remoteUrl": "https://storage.googleapis.com/firemat-preview-drop/emulator/dataconnect-emulator-macos-v2.10.1",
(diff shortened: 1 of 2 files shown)


src/emulator/downloadableEmulatorInfo.json
```diff
@@ -54,28 +54,28 @@
...
     },
     "win32": {
-      "version": "2.10.1",
-      "expectedSize": 29829632,
-      "expectedChecksum": "82e82e35c1728a214db37967dba751e7",
-      "expectedChecksumSHA256": "ba509e921c692ac60c1138b5b9c83a23b4e11feb339f08df5f1cfced44cffc91",
-      "remoteUrl": "https://storage.googleapis.com/firemat-preview-drop/emulator/dataconnect-emulator-windows-v2.10.1",
-      "downloadPathRelativeToCacheDir": "dataconnect-emulator-2.10.1.exe"
+      "version": "2.11.0",
+      "expectedSize": 29719040,
```
rosalyntan, line 66: Just want to double check that the cut was made from a synced client? `expectedSize` going down for all three platforms is a bit weird.

fredzqm (approved):

rosalyntan (changes requested):
see comment re: `expectedSize` discrepancy

src/emulator/downloadableEmulatorInfo.json
hlshen: thanks for checking. I'm synced and updated with main

PR #8963, opened 2025-08-13
fix: only show runtime config deprecation warning for user-defined config

### LINKED ISSUE #8925: Functions config deprecation warning even though I'm not using that API

**firebase-tools:** 14.11.1

**Platform:** OSX

Every time I deploy, I see the deprecation warning for `functions.config() API is deprecated.`. I'm not using and have never used that API. How can I figure out what's triggering this call and be certain that either I remove it, or that it will not stop me from deploying?

```
 DEPRECATION NOTICE: Action required to deploy after Dec 31, 2025
  
  functions.config() API is deprecated.
  Cloud Runtime Configuration API, the Google Cloud service used to store function configuration data, will be shut down on December 31, 2025. As a result, you must migrate away from using functions.config() to continue deploying your functions after December 31, 2025.
  
  What this means for you:
  
  - The Firebase CLI commands for managing this configuration (functions:config:set, get, unset, clone, and export) are deprecated. These commands no longer work after December 31, 2025.
  - firebase deploy command will fail for functions that use the legacy functions.config() API after December 31, 2025.
  
  Existing deployments will continue to work with their current configuration.
  
  See your migration options at: https://firebase.google.com/docs/functions/config-env#migrate-to-dotenv
```

I should not receive a deprecation warning for an API I'm not using. Or it should point me to the specific usage so I can remove it, which is especially important if it's a transitive dependency that I'm not aware of. 

I'm warned that I can't deploy next year due to an API I do not use.

taeold:
## Summary
Fixes false positive deprecation warnings for users who aren't using `functions.config()` and updates deprecation timeline.

## Problem
Users were seeing runtime config deprecation warnings even without using Runtime Config, because the default Firebase config (`projectId`, `storageBucket`) is always present for v1 functions.

## Solution
1. Only show the warning when runtime config contains keys other than the default `firebase` key
2. Updated deprecation date to March 2026 to give users more time to migrate

## Test plan
- [x] Build and lint pass
- [x] Warning only shows for actual Runtime Config usage
- [x] Default Firebase config doesn't trigger warning

Fixes #8925

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
-- [Fixed] Runtime config deprecation warning no longer shows for v1 functions that only use default Firebase config. (#8963)
+- Updated untime config deprecation warning to no longer shows for v1 functions that only use default Firebase config. (#8963)
+- Updated Data Connect emulator to v2.11.1, which:
+  - [added] Add an app watch that collects embedded GQL into the connector folder.
+  - [fixed] Handle foreign key constraint error as FailedPrecondition.
 - [Added] Support for creating Firestore Enterprise databases using `firestore:databases:create --edition enterprise`. (#8952)
 - [Added] Support for Firestore Enterprise database index configurations. (#8939)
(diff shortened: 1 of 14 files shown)


jhuleatt (approved):

tscislo:
Thanks for this, when there will be a version with this change available @taeold ?

PR #8965, opened 2025-08-13
[MCP] Adds support for prompts.

mbleigh:
Adds support for prompts in the Firebase MCP server along with a "deploy" prompt as a first entry.

### DIFF
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -401,14 +401,14 @@ export class FirebaseMcpServer {
     try {
       const messages = await prompt.fn(promptArgs, promptsCtx);
       await this.trackGA4("mcp_get_prompt", {
-        prompt_name: promptName,
+        tool_name: promptName,
       });
       return {
         messages,
       };
     } catch (err: unknown) {
       await this.trackGA4("mcp_get_prompt", {
-        prompt_name: promptName,
+        tool_name: promptName,
         error: 1,
       });
       // TODO: should we return mcpError here?


src/mcp/index.ts
```diff
@@ -324,6 +352,78 @@ export class FirebaseMcpServer {
...
+    await this.detectProjectRoot();
+    const promptName = req.params.name;
+    const promptArgs = req.params.arguments;
+    const prompt = this.getPrompt(promptName);
+    if (!prompt) {
+      throw new Error(`Prompt '${promptName}' could not be found.`);
+    }
+
+    let projectId = await this.getProjectId();
+    // TODO: add requiresProject to prompts
```
schnecle: This one's interesting - would we expect any prompt to be automatically hydrated with the project id if that is part of he input schema? Is that how it works in tools?

src/mcp/prompts/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  crashlytics: [],
+  apphosting: [],
+  database: [],
+};
+
+function addFeaturePrefixToPrompts(feature: string, prompts: ServerPrompt[]): ServerPrompt[] {
+  return prompts.map((p) => {
+    const newPrompt = { ...p };
+    newPrompt.mcp = { ...p.mcp };
+    newPrompt.mcp.name = `${feature}_${p.mcp.name}`;
```
schnecle: Consider prefixing with "firebase" as well at least in the short term. All prompts will show up as commands in alphabetical order when you type /, which can be quite confusing if you have multiple mcp servers present in gemini cli.

mbleigh: The `deploy` prompt is being very intentionally *not* prefixed. I need to think about this some more to figure out what the best approach is and how other clients handle this.

src/mcp/index.ts
mbleigh: See what I've done in `src/mcp/prompts/core/deploy.ts` -- the project id is provided as context to the prompt and can be used in the text of the prompt if it's needed/useful.

package.json
```diff
@@ -273,5 +273,6 @@
     "node-fetch": {
       "whatwg-url": "^14.0.0"
     }
-  }
+  },
+  "packageManager": "pnpm@10.13.1+sha256.0f9ed48d808996ae007835fb5c4641cf9a300def2eddc9e957d9bbe4768c5f28"
```
bkendall: What's this?

mbleigh: An oopsie from accidentally using `pnpm` locally. It's sooooo much better than `npm` but I didn't mean to introduce it here. Reverted.

src/mcp/prompts/index.ts
mbleigh: For now I've made it so everything is `firebase:<feat>:<name>` except if explicitly opted out of namespacing with `omitPrefix`.

schnecle: Sounds good to me.

PR #8979, opened 2025-08-15
feat(mcp): Adds `dataconnect_compile` tool to MCP server.

mbleigh:
This PR does a few things:

1. Adds a new `dataconnect_compile` tool that builds FDC schema/connectors and returns any errors.
2. Adds a new `mcpalpha` experiment so that we can gate MCP features on a flag to keep things on `master`.
3. Adds a new `firebase:dataconnect:schema` prompt that is only available with `mcpalpha` enabled.


src/experiments.ts
```diff
@@ -147,6 +147,11 @@ export const ALL_EXPERIMENTS = experiments({
     default: true,
     public: false,
   },
+  mcpalpha: {
+    shortDescription: "Opt-in to early MCP features before they're widely released.",
+    default: false,
+    public: false,
```
joehan: WDYT about making this public?

src/mcp/prompts/dataconnect/schema.ts
```diff
@@ -0,0 +1,74 @@
+import { z } from "zod";
```
joehan: Remove

src/mcp/tools/dataconnect/compile.ts
```diff
@@ -0,0 +1,47 @@
+import { z } from "zod";
+import { tool } from "../../tool";
+import { pickService } from "../../../dataconnect/fileUtils";
+import { compileErrors } from "../../util/dataconnect/compile";
+
+export const compile = tool(
+  {
+    name: "compile",
+    description:
+      "Use this to compile Firebase Data Connect schema and/or operations and check for build errors.",
```
joehan:
```suggestion
      "Use this to compile Firebase Data Connect schema, operations, and/or connectors and check for build errors.",
```

src/mcp/tools/dataconnect/compile.ts
```diff
@@ -0,0 +1,47 @@
...
+    inputSchema: z.object({
+      error_filter: z
+        .enum(["all", "schema", "operations"])
```
joehan, line 13:
```suggestion
        .enum(["all", "schema", "connectors"])
```

src/mcp/util/dataconnect/compile.ts
```diff
@@ -0,0 +1,21 @@
+import { prettify } from "../../../dataconnect/graphqlError";
+import { ServiceInfo } from "../../../dataconnect/types";
```
joehan: Remove

src/experiments.ts
mbleigh: Done!

src/mcp/index.ts
```diff
@@ -227,6 +227,10 @@ export class FirebaseMcpServer {
   }
 
   get availablePrompts(): ServerPrompt[] {
+    this.log(
```
fredzqm: Is this change intended?

src/mcp/tools/dataconnect/compile.ts
```diff
@@ -0,0 +1,47 @@
+import { z } from "zod";
+import { tool } from "../../tool";
+import { pickService } from "../../../dataconnect/fileUtils";
+import { compileErrors } from "../../util/dataconnect/compile";
+
+export const compile = tool(
+  {
+    name: "compile",
+    description:
+      "Use this to compile Firebase Data Connect schema, operations, and/or connectors and check for build errors.",
```
fredzqm, line 10: Can we do `dataconnect_build`?

I want to add `dataconnect:sdk:generate` as part of this as well. When customer change any GQL sources, it naturally makes sense to re-generated SDKs.

Later of course

src/mcp/tools/dataconnect/compile.ts
```diff
@@ -0,0 +1,47 @@
...
+    inputSchema: z.object({
+      error_filter: z
+        .enum(["all", "schema", "connectors"])
```
fredzqm: "connectors" optional doesn't make sense.

If schema is invalid, connectors will return no error, which is mis-leading. This should be either "schema" or "all".

src/mcp/util/dataconnect/compile.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export async function compileErrors(
+  configDir: string,
+  errorFilter?: "all" | "schema" | "operations",
+) {
+  const errors = (await DataConnectEmulator.build({ configDir })).errors;
+  return (
+    errors
+      ?.filter((e) => {
+        const isOperationError = ["query", "mutation"].includes(e.path?.[0] as string);
```
fredzqm, line 12: Not all clients follow the `query` & `mutation` convention. Only our CLI init template does.

We should probably not support `errorFilter` for now.

I can find a correct way to do this later.

src/mcp/prompts/dataconnect/schema.ts
```diff
@@ -0,0 +1,73 @@
...
+    arguments: [
+      {
+        name: "prompt",
+        description:
+          "describe the schema you want generated or the edits you want to make to your existing schema",
+        required: true,
+      },
+    ],
+    annotations: {
+      title: "Deploy to Firebase",
```
fredzqm: "title" looks off.

I am wondering if it makes more sense to rely on the GiF backend more, which is known to have good accuracy.

I have concerns over lack of eval & performance visibility in local MCP prompts. It gets worse if this tool shadows the `dataconnect_generate_schema` tool.

src/mcp/prompts/dataconnect/schema.ts
```diff
@@ -0,0 +1,73 @@
...
+Project ID: ${projectId || "<NONE>"}
+${renderServices(fdcServices)}${renderErrors(buildErrors)}
+
+==== USER PROMPT ====
+
+${prompt}
+
+==== TASK INSTRUCTIONS ====
+
+1. If Data Connect is marked as \`<UNCONFIGURED>\`, first run the \`firebase_init\` tool with \`{dataconnect: {}}\` arguments to initialize it.
```
fredzqm, line 63: NOTE: for myself. We can pass in appIdea here to let GiF generates the schema and connectors for us

src/mcp/prompts/dataconnect/schema.ts
fredzqm: [not blocking] Feel free to check this in.

I am trying to figure out where that line should be. Maybe it's edit vs create.

Use GiF API to generate the initial schema and new operations.

Use the local prompts to iterate and edit them. Cursor probably has more local context.

Without evaluation, it's really a pain to figure how the experience is.

src/mcp/util/dataconnect/compile.ts
mbleigh: No, I believe this will always be correct -- the path is like `["query", "users", "someField"]` because it's mapping `query { users { someField }}`. I believe it will always work.

src/mcp/tools/dataconnect/compile.ts
mbleigh: I'll just drop this for now, it can just see all the errors at once.

src/mcp/index.ts
mbleigh: Removed.

src/mcp/prompts/dataconnect/schema.ts
mbleigh: I don't think our visibility into GiF is all that much better, though I'd like us to come up with an informal eval system utilizing Gemini CLI for the MCP server generally. This is currently flagged off in the "mcpalpha" experiment so I'm not too worried about it one way or the other for now.

src/dataconnect/loadAll.ts
```diff
@@ -0,0 +1,7 @@
+import { load, readFirebaseJson } from "./load";
+import type { Config } from "../config";
+
+export function loadAll(projectId: string, config: Config) {
```
fredzqm: Since your PR, I added `loadAll` in a different file~

https://github.com/firebase/firebase-tools/blob/2f0b457b299019ada540f0ebd96c64070af19578/src/dataconnect/load.ts#L61C23-L61C30

src/mcp/util/dataconnect/compile.ts
fredzqm: Oh, I see. This is the GQL path error.

I got it mixed up with the file name.

fredzqm (approved):
Thank you Michael!! Love it!!! 🔥 🔥 🔥

PR #8980, opened 2025-08-15
BYO bucket for v2 functions uploads with runfunctions enabled

inlined:
As a way to test things progressively, when the "runfunctions" experiment is enabled, always choose to use your own bucket rather than using GCF's generate upload source function. To keep costs under control, objects are deleted after 1d, which is the lowest available retention policy.

Verified to work with the runfunctions experiment both enabled and disabled


inlined:
@annajowang Here's the PR I told you about offline. After writing the initial version, I looked up code to answer Abhii's q and realized there's a lot of overlap. I consolidated the code (where FAH's was better than what I had!). While I was there I got nerd sniped; I simplified some code, replaced Map with the canonical Record, and made some stuff more parallelized.

/gemini review

src/deploy/apphosting/deploy.ts
```diff
@@ -24,77 +24,56 @@ export default async function (context: Context, options: Options): Promise<void
...
-      cfg.backendId,
-      `gs://firebaseapphosting-sources-${options.projectNumber}-${backendLocation.toLowerCase()}/${path.basename(zippedSourcePath)}`,
-    );
-  }
+      const { bucket, object } = await gcs.uploadObject(
+        {
+          file: zippedSourcePath,
+          stream: fs.createReadStream(zippedSourcePath),
+        },
+        `firebaseapphosting-sources-${options.projectNumber}-${backendLocation.toLowerCase()}`,
```
annajowang: This is the same long string as the prefix of the one below. can this be a shared string const?

src/gcp/storage.ts
```diff
@@ -223,9 +224,10 @@ export async function upload(
   uploadUrl: string,
   extraHeaders?: Record<string, string>,
   ignoreQuotaProject?: boolean,
-): Promise<any> {
-  const url = new URL(uploadUrl);
-  const localAPIClient = new Client({ urlPrefix: url.origin, auth: false });
+): Promise<{ generation: string | null }> {
+  const url = new URL(uploadUrl, storageOrigin());
+  const signedUrl = url.searchParams.has("GoogleAccessId");
```
annajowang: nit `isSignedUrl`

src/deploy/apphosting/deploy.ts
```diff
@@ -24,77 +24,57 @@ export default async function (context: Context, options: Options): Promise<void
...
-            lifecycle: {
-              rule: [
-                {
-                  action: {
-                    type: "Delete",
-                  },
-                  condition: {
-                    age: 30,
-                  },
+  await Promise.all(
```
taeold, line 27: nit* consider `.allSettled`.?

(aside - I really dislike the boilerplate required to properly catch and fire errors when using `.allSettled` - wondering if you know of an easier way?)

src/deploy/functions/deploy.ts
```diff
@@ -48,33 +50,83 @@ async function uploadSourceV1(
...
+  // N.B. Should we upload to multiple regions? For now, just pick the first one.
+  // Uploading to multiple regions might slow upload and cost the user money if they
+  // pay their ISP for bandwidth, but having a bucket per region would avoid cross-region
+  // fees from GCP.
   const region = v2Endpoints[0].region; // Just pick a region to upload the source.
-  const res = await gcfv2.generateUploadUrl(projectId, region);
   const uploadOpts = {
     file: source.functionsSourceV2!,
-    stream: fs.createReadStream(source.functionsSourceV2!),
+    stream: (exports as { createReadStream: typeof createReadStream }).createReadStream(
```
taeold, line 77: aside: dang why wouldn't `exports.createReadStream` work here??

src/deploy/functions/deploy.ts
```diff
@@ -48,33 +50,83 @@ async function uploadSourceV1(
...
+      );
+    }
+    const res = await gcfv2.generateUploadUrl(projectId, region);
+    await gcs.upload(uploadOpts, res.uploadUrl, undefined, true /* ignoreQuotaProject */);
+    return res.storageSource;
   }
-  await gcs.upload(uploadOpts, res.uploadUrl, undefined, true /* ignoreQuotaProject */);
-  return res.storageSource;
+
+  // New behavior: BYO bucket since we're moving away from the GCF API.
```
taeold: can we leave this comment out for now?

src/deploy/functions/deploy.ts
```diff
@@ -48,33 +50,83 @@ async function uploadSourceV1(
...
+    projectId,
+    createMessage: `Creating Cloud Storage bucket in ${region} to store Functions source code uploads at ${bucketName}...`,
+    req: {
+      name: bucketName,
+      location: region,
+      lifecycle: {
+        rule: [
+          {
+            action: { type: "Delete" },
+            condition: { age: 1 }, // Delete objects after 1 day
```
taeold: is it yagni if we make this configurable?

(e.g. we occasionally get support request from users wanting to "get their source code")

src/deploy/functions/deploy.ts
```diff
@@ -48,33 +50,83 @@ async function uploadSourceV1(
...
+          },
+        ],
+      },
+    },
+  });
+  await gcs.upload(
+    uploadOpts,
+    `${bucketName}/${source.functionsSourceV2Hash}.zip`,
```
taeold: nit* define a variable t store zip filename to be used here and below

taeold (approved):
good for functions

src/deploy/apphosting/deploy.ts
inlined: All settled would be new error handling not present in the prevoius code, so I'll continue a conversation on whether this needs to be a follow-up.

Given that this is deploy not release and has no changes, what behavior would we want to have if we have partial success?

src/deploy/functions/deploy.ts
inlined: This is what we do elsewhere so I did it here. It works but you don't get type enforcement.

src/deploy/functions/deploy.ts
inlined: Will adjust to say while using the runfunctions experiment.

src/deploy/functions/deploy.ts
inlined: Let's start with a safe default (not spending unbounded money as time progresses) and add configurability as a feature in the future

annajowang (approved):

Issue #8985, opened 2025-08-16
Firebase Tools 14.12+ Database Rules Fail To Deploy


enchorb:
### [REQUIRED] Environment info

**firebase-tools:** `14.12.1`

**Platform:** All plafroms

### [REQUIRED] Test case

Deploy `firebase deploy --config firebase.json` 

Test database rules file

```
{
  "rules": {
    "users": {
      "$uid": {
        ".read": "$uid === auth.uid",
        ".write": "$uid === auth.uid"
      }
    }
  }
}
```

### [REQUIRED] Steps to reproduce

On `firebase-tools` versions `14.10.1` and before this works fine. On `14.12.1` I get the below error. This same config file has been working for years with no issue

### [REQUIRED] Expected behavior

Rules deploy

### [REQUIRED] Actual behavior

```
i  deploying database, storage, firestore, functions
i  database: checking rules syntax...

Error: Syntax error in database rules:

1:2: Expected 'rules' property.
```

ryzizub:
I have same issue, created PR to fix that

PR #8986, opened 2025-08-16
fix: update updateRulesWithClient in realtime-database to include src

ryzizub:
… parameter

### Description

I updated to latest version of firebase-tools and had troubles deploying the realtime database rules. Find out that source of realtime database is not included so those are incorrectly passed

Same issue is mentioned here https://github.com/firebase/firebase-tools/issues/8985

### Sample Commands

I did `firebase deploy` only

Before:
Snímek obrazovky 2025-08-16 v 21 52 12

After:
Snímek obrazovky 2025-08-16 v 21 52 20

### DIFF
--- a/src/rtdb.ts
+++ b/src/rtdb.ts
@@ -29,7 +29,7 @@ export async function updateRules(
   );
   const client = new Client({ urlPrefix: origin });
 
-  return updateRulesWithClient(client, options);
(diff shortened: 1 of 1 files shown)


jhuleatt:
@antholeole looks like this bug was introduced in https://github.com/firebase/firebase-tools/pull/8854 - can you please give this fix PR a review?

Issue #8676, opened 2025-05-29
[MCP] Crashlytics - Ability to fetch actual crash report


freak4pc:
## Summary

Hey there :) 
First of all, thank you for making the MCP available for us, it has a huge potential to build super powerful flows.

Right now I can get the top issues from the MCP, but it's only the first step to building a meaningful flow from it.
The next big step would be the ability to ask the MCP to retrieve the crash report itself of one of these listed crashes.

i.e., the equivalent of pressing this button in the console:

<img width="724" alt="Image" src="https://github.com/user-attachments/assets/2e7fb6ab-5af8-4eb0-bc0e-018ac4a9f138" />

(or alternatively pressing the TXT button)

if cursor can get back the crash info as a response that would allow building an end-to-end triage flow where you ask for the latest crash, pick it, and ask Cursor to suggest an appropriate fix based on the context of the codebase (Then, with a GitHub MCP, it can also open a PR with the suggestion, etc.)

This feels like a natural complement to the MCP capabilities, but happy to hear your thoughts on this :)

Thanks 🙏

arutkayb:
This ticket could be closed, I guess. As I tested now, the official Firebase plugin (firebase@claude-plugins-official) provides the following information correctly:
                                                                                   
  - Stack trace (exceptions, threads)
  - Logs & Breadcrumbs (logs, breadcrumbs)                                                                                                                            
  - Custom Keys (customKeys)
  - Device info (manufacturer, model, marketing name)                                                                                                                 
  - Memory & Storage                                    
  - OS & App version
  - Blame frame (file, line, symbol)
  - Process state (background/foreground)
  - Issue variants

freak4pc:
Indeed!

ldp940622:
Seems not to get stack information from threads other than the crash thread.

PR #8995, opened 2025-08-18
Add a couple of tools to fetch issue details and samples for a given issue id.

visumickey:
Addresses feature request https://github.com/firebase/firebase-tools/issues/8676

### DIFF
--- a/src/mcp/tools/crashlytics/get_issue_details.ts
+++ b/src/mcp/tools/crashlytics/get_issue_details.ts
@@ -1,7 +1,7 @@
 import { z } from "zod";
-import { tool } from "../../tool.js";
-import { mcpError, toContent } from "../../util.js";
-import { getIssueDetails } from "../../../crashlytics/getIssueDetails.js";
+import { tool } from "../../tool";
+import { mcpError, toContent } from "../../util";
+import { getIssueDetails } from "../../../crashlytics/getIssueDetails";
 
 export const get_issue_details = tool(
   {
--- a/src/mcp/tools/crashlytics/get_sample_crash.ts
+++ b/src/mcp/tools/crashlytics/get_sample_crash.ts
@@ -1,7 +1,7 @@
 import { z } from "zod";
-import { tool } from "../../tool.js";
-import { mcpError, toContent } from "../../util.js";
-import { getSampleCrash } from "../../../crashlytics/getSampleCrash.js";
+import { tool } from "../../tool";
+import { mcpError, toContent } from "../../util";
+import { getSampleCrash } from "../../../crashlytics/getSampleCrash";
 
 export const get_sample_crash = tool(
   {
--- a/src/mcp/tools/crashlytics/index.ts
+++ b/src/mcp/tools/crashlytics/index.ts
@@ -1,7 +1,7 @@
-import type { ServerTool } from "../../tool.js";
-import { get_issue_details } from "./get_issue_details.js";
-import { list_top_issues } from "./list_top_issues.js";
-import { get_sample_crash } from "./get_sample_crash.js";
+import type { ServerTool } from "../../tool";
+import { get_issue_details } from "./get_issue_details";
+import { list_top_issues } from "./list_top_issues";
+import { get_sample_crash } from "./get_sample_crash";
 
 export const crashlyticsTools: ServerTool[] = [
   list_top_issues,


src/crashlytics/fetchIssueDetails.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+
+function parseProjectId(appId: string): string | undefined {
+  const appIdParts = appId.split(":");
+  if (appIdParts.length > 1) {
+    return appIdParts[1];
+  }
+  return undefined;
+}
```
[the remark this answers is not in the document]
visumickey: Will do this in a future iteration.

src/mcp/tools/crashlytics/get_sample_crash.ts
```diff
@@ -0,0 +1,50 @@
...
+      issue_id: z
+        .string()
+        .describe(
+          "The issue Id for which the sample crash needs to be fetched. This is the value of the field `id` in the list of issues. Defaults to the first id in the list of issues.",
+        ),
+      variant_id: z
+        .string()
+        .optional()
+        .describe(
+          "The variant Id that maps to an issue Id for which the sample crash needs to be fetched.",
```
raymondlam: We can include that this is optional

visumickey: Should we include? Just marking the field as optional would make it to the LLM that this is an optional parameter. Don't think the description needs it. Have you seen this otherwise?

raymondlam: The phrasing is a bit off in `for which the sample crash needs to be fetched`, I think we should rewrite this so the behavior is more clear.

`The variant id to filter the list of sample crashes by.` or `Filters the sample crashes to the specific variant id`

src/mcp/tools/crashlytics/get_sample_crash.ts
```diff
@@ -0,0 +1,50 @@
...
+    inputSchema: z.object({
+      app_id: z
+        .string()
+        .describe(
+          "AppId for which the issues list should be fetched. For an Android application, read the mobilesdk_app_id value specified in the google-services.json file for the current package name. For an iOS Application, read the GOOGLE_APP_ID from GoogleService-Info.plist. If neither is available, use the `firebase_list_apps` tool to find an app_id to pass to this tool.",
+        ),
+      issue_id: z
+        .string()
+        .describe(
+          "The issue Id for which the sample crash needs to be fetched. This is the value of the field `id` in the list of issues. Defaults to the first id in the list of issues.",
```
raymondlam, line 19: `Defaults to the first id in the list of issues.` should be removed from the schema. I don't know if we should be assuming that this was called after an list_top_issues call.

visumickey: It is a train of tools that backtracks. When this tool gets called, if it does not have an issue id - we are suggesting it to call the get_top_issues and then get the sample crash for that issue. Atleast, that is how I had defined it. The order of tool call cannot be guaranteed. But I'm using backtracking methods to give a sequence of tooling.

src/mcp/tools/crashlytics/get_sample_crash.ts
visumickey: Made the change. One practice I'm trying to follow in the parameters for the MCP tool is to make them explanatory even without the context of the MCP tool. This gives LLM a huge advantage to pick the parameter based MCP tool. But, sometimes that complicates the description of the field.

src/crashlytics/getSampleCrash.ts
```diff
@@ -0,0 +1,59 @@
...
+  projectId: string,
+  appId: string,
+  issueId: string,
+  variantId?: string,
+  sampleCount?: number,
+): Promise<string> {
+  try {
+    const queryParams = new URLSearchParams();
+    queryParams.set("filter.issue.id", `${issueId}`);
+    queryParams.set("page_size", String(sampleCount!));
```
joehan: Don't use nonnull assertions on optional arguments. Either make this argument required, provide a default value, or wrap this in a conditional.

src/crashlytics/getSampleCrash.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (variantId) {
+      queryParams.set("filter.issue.variant_id", `${variantId!}`);
```
joehan:
```suggestion
      queryParams.set("filter.issue.variant_id", variantId);
```
Unecessary to wrap this in a template string or use a nonnull assertion, since you are already in a conditional.

src/mcp/tools/crashlytics/get_issue_details.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const get_issue_details = tool(
+  {
+    name: "get_issue_details",
+    description: "Gets the details about a specific crashlytics issue.",
+    inputSchema: z.object({
+      app_id: z
+        .string()
+        .describe(
+          "AppId for which the issues list should be fetched. For an Android application, read the mobilesdk_app_id value specified in the google-services.json file for the current package name. For an iOS Application, read the GOOGLE_APP_ID from GoogleService-Info.plist. If neither is available, use the `firebase_list_apps` tool to find an app_id to pass to this tool.",
```
joehan:
```suggestion
          "The App ID for which the issues list should be fetched. For an Android application, read the mobilesdk_app_id value specified in the google-services.json file for the current package name. For an iOS Application, read the GOOGLE_APP_ID from GoogleService-Info.plist. If neither is available, use the `firebase_list_apps` tool to find an app_id to pass to this tool.",
```

joehan (approved):
Code LGTM - probably worth testing this with some agents to see if the strategy you used for the descriptions works like we hope

joehan:
Also add a Changelog please

src/crashlytics/getIssueDetails.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+import { FirebaseError } from "../error";
+import { crashlyticsApiOrigin } from "../api";
+
+chai.use(chaiAsPromised);
+const expect = chai.expect;
+
+describe("getIssueDetails", () => {
+  const projectId = "my-project";
+  const appId = "1:1234567890:android:abcdef1234567890";
+  const requestProjectId = "1234567890";
```
schnecle: nit: use projectNumber to disambiguate the human readable version

src/crashlytics/getIssueDetails.ts
```diff
@@ -0,0 +1,49 @@
...
+  } catch (err: any) {
+    logger.debug(err.message);
+    throw new FirebaseError(
+      `Failed to fetch the issue details for the Firebase Project ${projectId}, AppId ${appId}, IssueId ${issueId}. Error: ${err}.`,
+      { original: err },
+    );
+  }
+}
+
+function parseProjectId(appId: string): string | undefined {
```
schnecle: naming: I'd call this parseProjectNumber

src/crashlytics/getSampleCrash.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+import { FirebaseError } from "../error";
+import { crashlyticsApiOrigin } from "../api";
+
+chai.use(chaiAsPromised);
+const expect = chai.expect;
+
+describe("getSampleCrash", () => {
+  const projectId = "my-project";
+  const appId = "1:1234567890:android:abcdef1234567890";
+  const requestProjectId = "1234567890";
```
schnecle: same nit

src/mcp/tools/crashlytics/get_issue_details.ts
```diff
@@ -0,0 +1,37 @@
...
+          "The issue ID for which the details needs to be fetched. This is the value of the field `id` in the list of issues. Defaults to the first id in the list of issues.",
+        ),
+    }),
+    annotations: {
+      title: "Gets the details of a specific issue.",
+      readOnlyHint: true,
+    },
+    _meta: {
+      requiresAuth: true,
+      requiresProject: true,
```
schnecle: One thing I encountered as I was playing around with this is that for some of these analysis tools, you may want to run them from a different directory. Is it easier if we allow you to pass in a project id? It certainly makes it a bit harder to test.

src/mcp/tools/crashlytics/get_sample_crash.ts
schnecle: In general pushing all of this instruction into the argument description feels funky to me. It might be the way we have to go, but it feels like it should be an anti-pattern. I don't have great reasons for it though.

schnecle (approved):

src/crashlytics/getSampleCrash.ts
visumickey: Good catch! Fixed it.

src/crashlytics/getSampleCrash.ts
visumickey: Done.

src/mcp/tools/crashlytics/get_issue_details.ts
visumickey: Done.

src/mcp/tools/crashlytics/get_issue_details.ts
visumickey: Good catch. I'm removing all the references of using a projectID and instead use the AppId to get the project number that is used in the API calls.

src/crashlytics/getSampleCrash.spec.ts
visumickey: Done. Modified it to be called a project number.

src/crashlytics/getIssueDetails.ts
visumickey: Done. Modified it to be called a project number.

src/crashlytics/getIssueDetails.spec.ts
visumickey: Done. Modified it to be called a project number.

src/mcp/tools/crashlytics/get_sample_crash.ts
visumickey: Marking this as resolved. Happy to iterate on this separately.

src/mcp/tools/crashlytics/get_sample_crash.ts
visumickey: I think we should discuss about options here. Will land this change and we can circle on this offline.

PR #8997, opened 2025-08-19
feat: Add unit tests for src/gcp/artifactregistry.ts

joehan:
### Description
From Jules:

This commit adds unit tests for the functions in src/gcp/artifactregistry.ts. This is the first step in adding test coverage to the src/gcp directory.


taeold (commented):
I like Jules, but honestly these unit test feel unhelpful.

e.g. it feels like we are just mocking the API and then expecting the return value to equal the mocked return value.

```js
const a = mocks("b").returns(1)

expect(a()).to.equal(1)
```

PR #9000, opened 2025-08-19
feat: Add `configDir` to specify a directory for .env files

taeold:
This change introduces a new property, `configDir`, to the `functions` configuration in `firebase.json`. This property allows users to specify a directory where `.env` files are located, providing more flexibility in project structure.

The `configDir` property is supported for both deployment and the local functions emulator.

- The `configDir` path is resolved relative to the `firebase.json` file.
- If `configDir` is not specified, the behavior remains unchanged, and `.env` files are looked for in the functions source directory.

Unit and emulator integration tests have been added to verify the new functionality.


src/emulator/functionsEmulator.ts
```diff
@@ -1385,12 +1387,15 @@ export class FunctionsEmulator implements EmulatorInstance {
...
-    const projectInfo = {
+    const projectInfo: functionsEnv.UserEnvsOpts = {
       functionsSource: backend.functionsDir,
       projectId: this.args.projectId,
       projectAlias: this.args.projectAlias,
       isEmulator: true,
     };
+    if (backend.configDir) {
+      projectInfo.configDir = backend.configDir;
+    }
```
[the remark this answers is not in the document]
inlined: I disagree with this feedback. Having an explicit undefined value leads to subtle bugs.

src/deploy/functions/prepare.ts
```diff
@@ -124,6 +124,7 @@ export async function prepare(
     const firebaseEnvs = functionsEnv.loadFirebaseEnvs(firebaseConfig, projectId);
     const userEnvOpt: functionsEnv.UserEnvsOpts = {
       functionsSource: options.config.path(config.source),
+      configDir: config.configDir ? options.config.path(config.configDir) : undefined,
```
inlined: I generally don't like present but undefined. I'd have copyIfPresent TBH

src/emulator/controller.ts
```diff
@@ -540,6 +540,7 @@ export async function startAll(
       }
       emulatableBackends.push({
         functionsDir,
+        configDir: cfg.configDir ? path.join(projectDir, cfg.configDir) : undefined,
```
inlined: same distaste for present but undefined

PR #9005, opened 2025-08-20
add bun as a supported package manager

nektro:
### Description

Adds https://bun.com as a supported package manager option by successfully detecting its lockfile.

### Scenarios Tested

Adds tests too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ ui-debug.log
 test_output.log
 scripts/emulator-tests/functions/index.js
 yarn.lock
+bun.lock
 .npmrc
 
 .DS_Store
--- a/src/emulator/apphosting/developmentServer.spec.ts
+++ b/src/emulator/apphosting/developmentServer.spec.ts
@@ -51,4 +51,16 @@ describe("utils", () => {
 
     expect(await detectPackageManager("./")).to.equal("yarn");
   });
+
+  it("returns bun if bun.lock file fond", async () => {
+    pathExistsStub.callsFake((...args) => {
+      if (args[0] === "bun.lock") {
+        return true;
+      }
+
+      return false;
+    });
+
+    expect(await detectPackageManager("./")).to.equal("bun");
+  });
 });
--- a/src/emulator/apphosting/developmentServer.ts
+++ b/src/emulator/apphosting/developmentServer.ts
@@ -9,7 +9,7 @@ export const logger = EmulatorLogger.forEmulator(Emulators.APPHOSTING);
 /**
  * Supported package managers. This mirrors production.
  */
-export type PackageManager = "npm" | "yarn" | "pnpm";
+export type PackageManager = "npm" | "yarn" | "pnpm" | "bun";
 
 /**
  * Returns the package manager used by the project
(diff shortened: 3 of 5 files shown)


abhis3:
Hi @nektro thanks so much for the contribution!

It looks like this change is only for Bun support in the emulator. In order to accept this change, we would need to add support for Bun in the actual buildpacks as that's used during the production deploy process to App Hosting.

You can see where we have the necessary infra to support other package managers here such as npm, yarn, pnpm, etc https://github.com/GoogleCloudPlatform/buildpacks/tree/main/cmd/nodejs

The frustrating part is that we have an internal repository that is mapped to the open source repo I linked, meaning that I don't think it's possible for others to submit a PR to the buildpacks repo, since the changes first need to be made internally and then ported over to open source to ensure we remain in a consistent state.

Do you mind sharing a bit more about your use case and why you use Bun, in general we'd love to learn more! In addition, please file a feature request on our UserVoice at https://firebase.uservoice.com/forums/948424-general?category_id=501599. I went through the existing requests and didn't see anything related to Bun, but we routinely check this as we do our planning cycles on what to work on next :)

nektro:
> Hi @nektro thanks so much for the contribution!

> It looks like this change is only for Bun support in the emulator. In order to accept this change, we would need to add support for Bun in the actual buildpacks as that's used during the production deploy process to App Hosting.
> 
> You can see where we have the necessary infra to support other package managers here such as npm, yarn, pnpm, etc [GoogleCloudPlatform/buildpacks@`main`/cmd/nodejs](https://github.com/GoogleCloudPlatform/buildpacks/tree/main/cmd/nodejs?rgh-link-date=2025-08-20T20%3A26%3A55Z)

thanks for the heads up! i'll send a PR over there too

> The frustrating part is that we have an internal repository that is mapped to the open source repo I linked, meaning that I don't think it's possible for others to submit a PR to the buildpacks repo, since the changes first need to be made internally and then ported over to open source to ensure we remain in a consistent state.

https://github.com/GoogleCloudPlatform/buildpacks/blob/main/CONTRIBUTING.md#code-reviews notes that github prs are accepted.

> Do you mind sharing a bit more about your use case and why you use Bun, in general we'd love to learn more! In addition, please file a feature request on our UserVoice at [firebase.uservoice.com/forums/948424-general?category_id=501599](https://firebase.uservoice.com/forums/948424-general?category_id=501599). I went through the existing requests and didn't see anything related to Bun, but we routinely check this as we do our planning cycles on what to work on next :)

I work at @oven-sh and am happy to follow up on this if anything comes up. Bun is by a decent margin the fastest npm-compilatible package manager i've used and so i love to use it for that reason.

the specific impetus for this patch was a user request https://x.com/elkriefy/status/1957914397893861432
we also get a number of users reporting using firebase https://github.com/oven-sh/bun/issues?q=is%3Aissue%20state%3Aopen%20firebase
apologies for the lack of uservoice thread

abhis3:
Amazing thanks for the context, that's great to know I'll also surface this to the team! But overall we welcome this change and your contributions, and would love to support Bun (we have a few fans already here on the team).

Feel free to ping this thread when you have a PR out in the other repo and I can help get the right eyes on it. I briefly synced with the relevant team and just 2 additional things I want to point out so that there isn't too much churn:

1. Do you mind creating an issue in the repo describing what the change you want to make is? That way you don't make a potentially large PR for something that wasn't on the team's radar
2. There is some support that you'll need to add to some other files, specifically our platform specific buildpacks. I _believe_ that this should account for all of the places this change will need to be made to support Bun, but I _could_ be missing something so please quickly double check if possible. I'm linking the nodejs.go file here as an example for things in that directory https://github.com/GoogleCloudPlatform/buildpacks/blob/main/pkg/nodejs/nodejs.go#L66

But overall excited for this change and thanks for your contributions! Please let me know and I'm happy to interface and engage the right teams on our end to get this out!

PR #9015, opened 2025-08-20
Random attempt at new ergonomics where copyIfPresent can be inlined

inlined:
Just having fun with ergonomics.


inlined:
I think I also might just play with Zod

PR #9025, opened 2025-08-21
Allow apphosting:backends:create to be called with --non-interactive

annajowang:
### Description

Allow for apphosting:backends:create to be called with --non-interactive. This adds two flags that are required in --non-interactive mode, but if provided even without --non-interactive, they will reduce the number of user interactions required.

### Scenarios Tested

firebase apphosting:backends:create --primary-region=europe-west4 --backend=test-non-interactive
firebase apphosting:backends:create --primary-region=europe-west4 --backend=test-non-interactive --non-interactive
firebase apphosting:backends:create --primary-region=europe-west4 --non-interactive

firebase apphosting:backends:create


src/apphosting/backend.ts
```diff
@@ -73,8 +73,12 @@ async function awaitTlsReady(url: string): Promise<void> {
  */
 export async function doSetup(
   projectId: string,
+  force: boolean,
   webAppName: string | null,
+  backendId: string | null,
   serviceAccount: string | null,
+  primaryRegion: string | null,
+  rootDir: string | undefined,
```
taeold: i'd suggest aligning on using `null` or `undefined` (I'd prefer `undefined`) and not both.

src/apphosting/backend.ts
```diff
@@ -85,30 +89,34 @@ export async function doSetup(
...
-
-  const gitRepositoryLink: GitRepositoryLink = await githubConnections.linkGitHubRepository(
-    projectId,
-    location,
-  );
+  const location = primaryRegion
+    ? primaryRegion
+    : await promptLocation(projectId, "Select a primary region to host your backend:\n");
+
+  if (rootDir == null && !force) {
```
taeold: nit* prefer `===` instead of `==` since `==` often has surprising type casting behavior

Also, IF you are changing to `===`, be aware that `null === undefined` evalutes to false.

src/apphosting/backend.ts
```diff
@@ -85,30 +89,34 @@ export async function doSetup(
...
+    });
+  }
 
-  const gitRepositoryLink: GitRepositoryLink = await githubConnections.linkGitHubRepository(
-    projectId,
-    location,
-  );
+  let gitRepositoryLink: GitRepositoryLink | undefined;
+  let branch: string | undefined;
+  if (!force) {
```
taeold: I'd love to see documentation around what `force` does in the setup. It sounds like `force` allows users to not provide certain configuration values?

src/apphosting/backend.ts
```diff
@@ -85,30 +89,34 @@ export async function doSetup(
...
+    // prefers. We could reduce the number of questions asked by letting people
+    // enter tag:<pattern>?
+    branch = await githubConnections.promptGitHubBranch(gitRepositoryLink);
+    logSuccess(`Repo linked successfully!\n`);
+  }
 
-  logBullet(`${clc.yellow("===")} Set up your backend`);
-  const backendId = await promptNewBackendId(projectId, location);
-  logSuccess(`Name set to ${backendId}\n`);
+  if (backendId == null) {
```
taeold: seems a bit odd to me that setting `force` would still prompt here. Maybe it should fail if `backendId` is not set and you also set `force`.

src/apphosting/backend.ts
```diff
@@ -127,6 +135,14 @@ export async function doSetup(
   );
   createBackendSpinner.succeed(`Successfully created backend!\n\t${backend.name}\n`);
 
+  if (force) {
+    return;
```
taeold, line 152: q: why return instead of moving forward with rest of setup (setting up traffic policy, starting a rollout, etc.)

src/apphosting/backend.ts
annajowang: was relying on command checking for backendId when force == true. but you're right, this doSetup() can be called by anything. Added a check for non-interactive mode.

src/apphosting/backend.ts
annajowang: In nonInteractive mode, no github repo was connected. There's nothing to rollout. Added inline comment to explain

src/apphosting/backend.ts
annajowang: lower level methods were expecting a mix of "string | null" and "string | undefined". but made all undefined and cleaned things up

src/apphosting/backend.ts
```diff
@@ -73,44 +73,63 @@ async function awaitTlsReady(url: string): Promise<void> {
...
 export async function doSetup(
   projectId: string,
-  webAppName: string | null,
-  serviceAccount: string | null,
+  nonInteractive: boolean,
+  webAppName: string | undefined,
+  backendId: string | undefined,
+  serviceAccount: string | undefined,
+  primaryRegion: string | undefined,
+  rootDir: string | undefined,
```
taeold:
```suggestion
  webAppName?: string,
  backendId?: string,
  serviceAccount?: string,
  primaryRegion?: string,
  rootDir?: string,
```

src/apphosting/backend.ts
```diff
@@ -73,44 +73,63 @@ async function awaitTlsReady(url: string): Promise<void> {
...
-  const location = await promptLocation(
-    projectId,
-    "Select a primary region to host your backend:\n",
-  );
+  let location = primaryRegion;
+  let gitRepositoryLink: GitRepositoryLink | undefined;
+  let branch: string | undefined;
+  if (nonInteractive) {
+    if (backendId === undefined || primaryRegion === undefined) {
+      throw new FirebaseError("non-interactive mode requires a backendId and primaryRegion");
```
taeold:
```suggestion
      throw new FirebaseError("non-interactive mode requires a --backend and --primary-region");
```

src/apphosting/backend.ts
```diff
@@ -73,44 +73,63 @@ async function awaitTlsReady(url: string): Promise<void> {
...
+    }
+  } else {
+    if (location === undefined) {
```
taeold: nit* more of a stylistic thing, I'd usually write `if(!location)` for brevity.

taeold (approved):

jpreid7396:
One thing I'm struggling with in the prompt is making sure the AI first detects that there is already an active Firebase project. It seems to always ask which project to use, even if there is an active project. I don't think this is a launch blocker, but curious if @mbleigh has ideas to fix this

**Edit:** I think it's because I was removing firebase.json each time. Does removing firebase.json unset the active project in the CLI?

src/apphosting/backend.ts
annajowang: I already have an error like this using `--backend` and `--primary-region` in the command file that would be triggered for this case. 
At this level, anything could be calling this function, so I referenced the args of the function, not the CLI command options.

PR #9029, opened 2025-08-22
feat(mcp/prompts/crashlytics): Add mcp prompts for prioritizing and fixing Crashlytics issues

schnecle:
### Description

* Add mcp prompts for prioritizing and fixing Crashlytics issues.
* Removed the firebase prefix in favor of just the feature name because there is incredibly limited real estate in the Gemini CLI command name space.

### Scenarios Tested

Attached Gemini CLI to MCP locally and ran the prioritize_issues and fix_issue commands.

### Sample Commands
```
$ gemini
> /crashlytics:prioritize_issues --app_id="(optional) your mobile-app-id here" --issues_count="(optional) number of issues to prioritize" --prompt="(optional) any additional instructions you want to add"
> /crashlytics:fix_issue --issue_id="your issue-id here"  --app_id="(optional) your mobile-app-id here"
```


src/mcp/prompts/crashlytics/fix_issue.ts
```diff
@@ -0,0 +1,89 @@
...
+        required: false,
+      },
+      {
+        name: "issue_id",
+        description: "A Crashlytics issue id.",
+        required: true,
+      },
+    ],
+    annotations: {
+      title: "Prioritize Crashlytics Issues",
```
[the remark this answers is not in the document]
visumickey: +1. I think there is a mismatch here.

src/mcp/prompts/crashlytics/prioritize_top_issues.ts
```diff
@@ -0,0 +1,79 @@
...
+
+Follow the steps below taking note of any user instructions provided above.
+
+1. If there is no active user, prompt the user to run \`firebase login\` in an interactive terminal before continuing.
+2. If there is no active app id, then do the following:
+   2a. If this is an Android app, read the mobilesdk_app_id value specified in the google-services.json file
+   2b. If this is an iOS app, read the GOOGLE_APP_ID from GoogleService-Info.plist file
+   2a. If you can't find either of the above, ask the user for the app id.
+3. Use the \`crashlytics_list_top_issues\` tool to fetch up to 20 issues.
+5. Use the \'crashlytics_list_top_versions\' tool to fetch the top versions for this app.
```
raymondlam: We skipped 4.

src/mcp/prompts/crashlytics/fix_issue.ts
```diff
@@ -0,0 +1,89 @@
...
+  async ({ app_id, issue_id }, { config, accountEmail }) => {
+    return [
+      {
+        role: "user" as const,
+        content: {
+          type: "text",
+          text: `
+Your goal is to fix a specific issue from Crashlytics.
+
+Active user: ${accountEmail || "<NONE>"}
```
raymondlam, line 36: Why is the accountEmail needed?

src/mcp/prompts/crashlytics/fix_issue.ts
```diff
@@ -0,0 +1,89 @@
...
+Active app: ${app_id || "<NONE>"}
+Issue id: ${issue_id}
+
+Contents of \`firebase.json\` config file:
+
+\`\`\`json
+${config.readProjectFile("firebase.json", { fallback: "<FILE DOES NOT EXIST>" })}
```
raymondlam: Is the firebase.json needed as well?

src/mcp/prompts/crashlytics/fix_issue.ts
```diff
@@ -0,0 +1,89 @@
...
+
+1. If there is no active user, prompt the user to run \`firebase login\` in an interactive terminal before continuing.
+2. If there is no active app id, then do the following:
+   2a. If this is an Android app, read the mobilesdk_app_id value specified in the google-services.json file
+   2b. If this is an iOS app, read the GOOGLE_APP_ID from GoogleService-Info.plist file
+   2a. If you can't find either of the above, ask the user for the app id.
+3. Make sure you have a good understanding of the code structure and where different functionality exists
+4. Use the \`crashlytics_get_issue_details\` and \'crashlytics_get_sample_crash_for_issue\' tools to get more context on the issue.
+5. Read the files that exist in the stack trace of the issue to understand the crash deeply.
+5. Determine the root cause of the crash.
```
raymondlam, line 58: There are two 5.

src/mcp/prompts/crashlytics/fix_issue.ts
```diff
@@ -0,0 +1,89 @@
...
+6. Write out a plan using the following criteria:
+   6a. Write out a description of the issue and including
+       * A brief description of the cause of the issue
+       * A determination of your level of confidence in the cause of the issue
+       * A determination of which library is at fault, this codebase or a dependent library
+       * A determination for how complex the fix will be
+   6b. The plan should include relevant files to change
+   6c. The plan should include a test plan to verify the fix
```
raymondlam, line 66: Should we also mention test cases? Or is it not applicable to every issue.

src/mcp/prompts/crashlytics/fix_issue.ts
visumickey: +1 Why do we need user details for this journey? Is this for notes addition scenarios?

src/mcp/prompts/crashlytics/fix_issue.ts
```diff
@@ -0,0 +1,89 @@
...
+    return [
+      {
+        role: "user" as const,
+        content: {
+          type: "text",
+          text: `
+Your goal is to fix a specific issue from Crashlytics.
+
+Active user: ${accountEmail || "<NONE>"}
+Active app: ${app_id || "<NONE>"}
```
visumickey, line 37: AppId cannot be none. Needs a valid AppID.

src/mcp/prompts/crashlytics/fix_issue.ts
```diff
@@ -0,0 +1,89 @@
...
+Issue id: ${issue_id}
+
+Contents of \`firebase.json\` config file:
```
visumickey: Should we hierarchically position the need of how to pick the appID:

1. If Android, google service.json file.
2. If iOS, GoogleServic-Info.plist
3. Finally firebase.json.

FWIW firebase.json is heavily project indicative and not sure if that can provide details of the app ID. We might have to get the list of appIDs with the project ID if we were to go that route.

src/mcp/prompts/crashlytics/fix_issue.ts
```diff
@@ -0,0 +1,89 @@
...
+
+\`\`\`json
+${config.readProjectFile("firebase.json", { fallback: "<FILE DOES NOT EXIST>" })}
+\`\`\`
+
+## Steps
+
+Follow the steps below taking note of any user instructions provided above.
+
+1. If there is no active user, prompt the user to run \`firebase login\` in an interactive terminal before continuing.
```
visumickey: Would'nt this be already taken care? I mean the login situation. You cannot call the tools without the login setup right?

src/mcp/prompts/crashlytics/fix_issue.ts
```diff
@@ -0,0 +1,89 @@
...
+2. If there is no active app id, then do the following:
+   2a. If this is an Android app, read the mobilesdk_app_id value specified in the google-services.json file
```
visumickey: GoogleServices.json file is a common file with multiple package ids and its corresponding appId. The way we need to do that is to say pick the appID corresponding to the package name of the current application as specified in the application bundle.

src/mcp/prompts/crashlytics/fix_issue.ts
```diff
@@ -0,0 +1,89 @@
...
+Active app: ${app_id || "<NONE>"}
+Issue id: ${issue_id}
+
+Contents of \`firebase.json\` config file:
+
+\`\`\`json
+${config.readProjectFile("firebase.json", { fallback: "<FILE DOES NOT EXIST>" })}
+\`\`\`
+
+## Steps
```
visumickey, line 52: Let's merge this change. but I definitely feel that this prompt needs iteration. This is still quite high level to guide the model to do what we want it to do. Happy to work together to get this few levels deeper.

src/mcp/prompts/crashlytics/prioritize_top_issues.ts
```diff
@@ -0,0 +1,79 @@
+import { prompt } from "../../prompt";
+
+export const prioritize_issues = prompt(
+  {
+    name: "prioritize_issues",
+    omitPrefix: false,
+    description: "Fetch and prioritize issues from Crashlytics.",
```
visumickey: Slightly confused with this description - this feels very close to top 'n' issue MCP tool. I assumed the journey here would be be about given an issue id, it would help prioritize. But the journey I read here is "get me the top 'n' issues and prioritize them for me" which is very close to "get the top 'n' issues for me".

src/mcp/prompts/crashlytics/fix_issue.ts
schnecle: We do actually need this to make sure you are logged in and able to call tools.

src/mcp/prompts/crashlytics/fix_issue.ts
schnecle: What do you mean by test cases? Like unit test cases?

src/mcp/prompts/crashlytics/fix_issue.ts
schnecle: We leave the app_id argument as optional to give the developer flexibility in where they want to run this. In the instructions below, it allows for reading the value out of the google-services file or getting it manually. This is different than the tools because it is explicitly invoked by the developer not by the LLM (at least gemini CLI is not yet capable of doing this). That's why this is left as app_id or NONE.

src/mcp/prompts/crashlytics/fix_issue.ts
schnecle: yeah, this was in here previously when we were passing the project id separately. Now that we aren't, I don't think we need the firebase.json file.

src/mcp/prompts/crashlytics/fix_issue.ts
schnecle: This is a prompt, so it could be the first thing that you do. You can't call tools without logging in.

src/mcp/prompts/crashlytics/fix_issue.ts
schnecle: Good call out. We actually ran into something similar to this with multiple google services files based on environment (test, beta, prod). I'll try to refine this.

src/mcp/prompts/crashlytics/prioritize_top_issues.ts
schnecle: The thing that is different here is that it is instructing it about the actual prioritization. The tool just fetches issues against the API, but it actually doesn't do any prioritization. This tells the LLM to grab 20 issues, and then order them based on certain criteria.

src/mcp/prompts/crashlytics/fix_issue.ts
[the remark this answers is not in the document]
schnecle: Done.

src/mcp/prompts/crashlytics/fix_issue.ts
```diff
@@ -0,0 +1,89 @@
...
+   <A plan for how to fix the issue>
+
+   **Test**
+   <a plan for how to test that the issue has been fixed an protect against regressions>
+
+7. Present the plan to the user and get approval before making the change.
+8. Fix the issue.
+   9a. Be mindful of API contracts and do not add fields to resources without a clear way to populate those fields
+   9b. If there is not enough information in the crash report to find a root cause, describe why you cannot fix the issue instead of making a guess.
+9. Ask the developer if they would like you to test the fix for them.
```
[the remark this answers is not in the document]
schnecle: Done.

src/mcp/prompts/crashlytics/prioritize_top_issues.ts
```diff
@@ -0,0 +1,79 @@
...
+7. If the user instructions do not include statements about prioritization, then prioritize the returned issues using the following criteria:
+  7a. The app versions for the issue include the most recent version of the app.
+  7b. The number of users experiencing the issue across variants
+  7c. The volume of crashes
+8. Return the top ${issue_count || 5} issue ids, with a brief description of the issue in the following format: 
+   * Issue <issue id>
+       * Description: <issue description>
+       * Rationale: <the reason this issue was prioritized in the way it was>
+
+The point of this command is to surface information only.
```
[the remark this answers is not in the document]
schnecle: Done.

src/mcp/prompts/crashlytics/fix_issue.ts
schnecle: Removed.

src/mcp/prompts/crashlytics/fix_issue.ts
schnecle: Done.

src/mcp/prompts/crashlytics/fix_issue.ts
schnecle: I wonder if for this first pass, we just instruct it to ask the developer what it should do.

schnecle: It would look like this:

```
╭────────────────────────────────────╮                                                                                                                                                         
│  > /crashlytics:prioritize_issues  │                                                                                                                                                         
╰────────────────────────────────────╯                                                                                                                                                         
                                                                                                                                                                                               
✦ Okay, I will prioritize the top 5 issues from Crashlytics. First, I need to find the App ID. I see multiple google-services.json files in the project. I will read them to                   
  find the mobilesdk_app_id.
 ╭───────────────────────────────────────────╮
 │ ✔ ReadManyFiles Will attempt to read and concatenate files using patterns:                                 │
 │    a/fake/path…                                                                                                                                         │
 │                                                                                                                                                                   │
 │    ### ReadManyFiles Result (Target Dir: `/a/fake/path/to/an/app`)                                                  │
 │                                                                                                                                                                   │
 │    Successfully read and concatenated content from **4 file(s)**.                                                     │
 │                                                                                                                                                                   │
 │    **Processed Files:**                                                                                                                            │
 │    - `fake/app/src/beta/google-services.json`                                                                                      │
 │    - `fake/app/src/debug/google-services.json`                                                                                   │
 │    - `fake/app/src/internal/google-services.json`                                                                                 │
 │    - `fake/app/src/release/google-services.json`                                                                                 │
╰───────────────────────────────────────────╯
✦ I found multiple mobilesdk_app_id values in the google-services.json files. Please choose which one you'd like me to use:

   * beta: <beta app id>
   * debug: <debug app id>
   * internal: <internal app id>
   * release: <release app id>

```

src/mcp/prompts/crashlytics/fix_issue.ts
visumickey: Aah, got it. That makes sense.

src/mcp/prompts/crashlytics/fix_issue.ts
visumickey: Interesting. Have your tried running MCP server without logging in? I thought that is a prerequisite for the MCP server to even kick start. I have not seen this login requirement in other MCP tools - so the question.

src/mcp/prompts/crashlytics/fix_issue.ts
visumickey: I love this. On the other hand, the scenario could also mean one fat google-services.json file used for all environments where the appID is picked based on the current packageID. Curious how CLI handles those scenarios.

src/mcp/prompts/crashlytics/prioritize_top_issues.ts
visumickey: Sounds good. I like the fact that the prioritization is controlled. So, might be an extra capability we are introducing for the developers to allow them to play around with prioritization. Let's move ahead with this change and then iterate.

src/mcp/prompts/crashlytics/fix_issue.ts
raymondlam: Yea, adding unit tests in addition to making the fix. The "test plan" is a bit ambiguous whether it would create unit tests.

src/mcp/prompts/crashlytics/fix_issue.ts
schnecle: When you aren't logged in, the tool calls just fail with an auth error and then it tells you to log in. Though super interestingly, the cli agent seems to be really eager to skip this check and will move straight on to looking for the app id. Pretty much no matter what I say to it.

src/mcp/prompts/crashlytics/fix_issue.ts
schnecle: Do you have an example google-services file? I can test it out.

src/mcp/prompts/crashlytics/fix_issue.ts
schnecle: I think it's a good idea. Let's tackle that as a follow up.

PR #9030, opened 2025-08-22
Jules unit tests for apphosting MCP tools

joehan:
### Description
From Jules - unit tests for the app_hosting tools.


bkendall:
I'm fine with these as-is. Might improve them later, but good enough for now.

PR #9032, opened 2025-08-22
[FDC] VS Code SDK init button should always run CLI

fredzqm:
We have been making lots of improvements to `firebase init dataconnect`.

VS Code should always delegate to CLI as opposed to manipulate the `connector.yaml` itself.


fredzqm:
@hlshen Please take a look~

Given our recent improvements to `init dataconnect:sdk`, the VS Code portion has drifted apart. The easiest way forward is to delegate to CLI al the time.

hlshen (approved):

PR #9033, opened 2025-08-23
[FDC] Update init template

fredzqm:
* Leave GQL files uncommented. 
* Remove comments about gen SDK in `connector.yaml`. They are going away soon.
* Document STRICT and COMPATIBLE mode in `dataconnect.yaml`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -110,7 +110,7 @@ module.exports = {
       rules: {},
     },
     {
-      files: ["src/mcp/tools/**/*.ts"],
+      files: ["src/mcp/tools/**/*.ts", "src/mcp/prompts/**/*.ts"],
       rules: { camelcase: "off" },
     },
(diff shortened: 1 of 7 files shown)


fredzqm:
@joehan here are some low-hanging fruit to improve init template.

Mainly want to leave GQL files uncommented. Remove the comment is an extra step, not adding any value.

fredzqm:
I can double check. Pretty sure that's the case already 

From here: https://github.com/firebase/firebase-tools/blob/cd601f9ceec2d255ad08b5ecbe98f9eb2f78920b/src/init/features/dataconnect/index.ts#L278

If operation generation failed, it won't save any connector.

joehan:
Note to self: add a changelog entry for this later - we can bundle all the varous changes into 'Improved `init dataconnect` templates'

PR #9037, opened 2025-08-25
Jules tests core tools

joehan:
### Description
From Jules:

Tests for a bunch of the core tools.  It originally covered the last 2 tools in the directory as well, but the test files were causing compiler issues that i didn't want to untangle.


nohe427:
@joehan - Could you look over src/mcp/tools/core/init.spec.ts

PR #9044, opened 2025-08-27
Update FDC local toolkit to v2.12.0

nguymichael:
### Description

<Title>


src/emulator/downloadableEmulatorInfo.json
```diff
@@ -54,28 +54,28 @@
   },
   "dataconnect": {
     "darwin": {
-      "version": "2.11.2",
+      "version": "2.12.0",
       "expectedSize": 29447008,
```
joehan, line 58: Is this really the exact same size, to the byte?

nguymichael: For posterity, here's the output of the deploy script: https://paste.googleplex.com/6198273630797824#l=7

oddly... it's the same.

PR #9045, opened 2025-08-27
[MCP] Always include core MCP tools + return instructions from firebase_init

fredzqm:
While testing this, I realized that all of the core MCP tools were not included.

Screenshot 2025-08-27 at 3 35 48 PM


src/mcp/tools/index.ts
```diff
@@ -17,6 +17,9 @@ export function availableTools(activeFeatures?: ServerFeature[]): ServerTool[] {
   if (!activeFeatures?.length) {
     activeFeatures = Object.keys(tools) as ServerFeature[];
   }
+  if (!activeFeatures.includes("core")) {
```
fredzqm, line 20: Emm, this is embarrassing

PR #9055, opened 2025-08-29
[FDC] Fix `firebase dataconnect:sdk:generate --watch`

fredzqm:
While switching `fdc generate` to `fdc sdk generate` cmd, I found this bug.

`fdc sdk generate --watch` would hang with no output and wait for any file changes.

The previous behavior hangs because the cmd never exits.
This PR changed it to pipe stdout, so developers can see what it's doing behind the scenes.

Screenshot 2025-09-02 at 3 43 36 PM


fredzqm:
[not urgent] While swapping `fdc generate` to `fdc sdk generate`, I found this bug.

Please take a look~ Not in a rush

src/commands/dataconnect-sdk-generate.ts
```diff
@@ -20,36 +21,32 @@ export const command = new Command("dataconnect:sdk:generate")
...
+      logger.warn(`Run ${clc.bold("firebase init dataconnect:sdk")} to configure a generated SDK.`);
+      logger.warn(
+        `See https://firebase.google.com/docs/data-connect/web-sdk for more details of how to configure generated SDKs.`,
+      );
+      return;
+    }
+    for (const serviceInfo of serviceInfosWithSDKs) {
+      const configDir = serviceInfo.sourceDirectory;
+      const account = getProjectDefaultAccount(options.projectRoot);
+      await DataConnectEmulator.generate({
```
yuchenshi: When watch is specified, this will hang on watching the first service / connector instead of watching them all.

Please use different logic for watch and do something like `Promise.race` to terminate upon any command exit.

src/emulator/dataconnectEmulator.ts
```diff
@@ -238,39 +236,44 @@ export class DataConnectEmulator implements EmulatorInstance {
...
+            // Command failed
+            reject(new Error(`Command failed with exit code ${code}`));
+          }
+        });
+
+        proc.on("error", (err) => {
+          // Handle errors like command not found
+          reject(err);
+        });
+      } catch (e: any) {
```
yuchenshi, line 263: Are you sure errors like missing Rosetta will be caught by this block, not the `proc.on("error")` above? I doubt `spawn` throws synchronously (as opposed to `spawnSync`).

Consider rejecting all errors in the Promise constructor and chaining `.catch(...)` to handle errors there

src/commands/dataconnect-sdk-generate.ts
```diff
@@ -20,36 +21,32 @@ export const command = new Command("dataconnect:sdk:generate")
...
-        const output = await DataConnectEmulator.generate({
-          configDir,
-          connectorId: conn.connectorYaml.connectorId,
-          watch: options.watch,
-          account,
-        });
-        logger.info(output);
-        logger.info(`Generated SDKs for ${conn.connectorYaml.connectorId}`);
-      }
+      logLabeledSuccess("dataconnect", `Successfully Generated SDKs`);
```
yuchenshi: This log line is misleading for `watch` where the commands are terminated or run to completion

yuchenshi (changes requested):

src/emulator/dataconnectEmulator.ts
fredzqm: This logic is copied from `_runBinary` [here](https://github.com/firebase/firebase-tools/blob/a0033aa9be00a9b751b53e42778b0542b3d5efd2/src/emulator/downloadableEmulators.ts#L383)

To answer your question, honestly I am not sure. I don't know how to re-produce the original error.

We should probably extract a helper to run long-running emulator-ish binary process.

Can we do that separately? This bug here is pretty bad.

src/commands/dataconnect-sdk-generate.ts
fredzqm: Good point. Done

src/commands/dataconnect-sdk-generate.ts
fredzqm: Done

src/commands/dataconnect-sdk-generate.ts
```diff
@@ -20,36 +21,42 @@ export const command = new Command("dataconnect:sdk:generate")
...
-        });
-        logger.info(output);
-        logger.info(`Generated SDKs for ${conn.connectorYaml.connectorId}`);
-      }
+    });
+
+    if (options.watch) {
+      await Promise.race(generateRuns);
+    } else {
+      await Promise.all(generateRuns);
```
yuchenshi: This runs them all at once and may cause the stdout to interleave each other. The previous code containing for-loop works a lot better here.

src/emulator/dataconnectEmulator.ts
yuchenshi: Okay fair enough. Feel free to split that into a follow-up

yuchenshi (changes requested):

src/commands/dataconnect-sdk-generate.ts
fredzqm: Okay, sure. Did the refactor to make it serial.

yuchenshi (approved):

PR #9062, opened 2025-09-01
fix(emulator, auth): allow SMS MFA finalization with obfuscated number

mikehardy:
### Description

Hi there 👋  - I'm one of the maintainers of react-native-firebase on behalf of Google

I'm trying to enable MFA testing with SMS as the second factor for the iOS platform, and I noticed a problem. This patch fixes it.

SMS MFA works using firebase-js-sdk and firebase-android-sdk against the emulator - these send unobfuscated phone numbers for the finalization step allowing successful enrollment location

firebase-ios-sdk sends an obfuscated phone number during the finalization step though and the enrollment isn't found, leading to SMS MFA failures against the emulator despite the same code working against cloud auth.

By allowing the enrollment lookup to check both unobfuscated and obfuscated numbers for enrollment id, the emulator will successfully finalize an MFA login with SMS as the second factor, on all platforms

### Scenarios Tested

See description

### Sample Commands

See descriptoin

### DIFF
--- a/src/emulator/auth/operations.ts
+++ b/src/emulator/auth/operations.ts
@@ -2221,7 +2221,14 @@ async function mfaSignInFinalize(
 
   let { user, signInProvider } = parsePendingCredential(state, reqBody.mfaPendingCredential);
   const enrollment = user.mfaInfo?.find(
-    (enrollment) => enrollment.unobfuscatedPhoneInfo === phoneNumber,
+    (enrollment) => {
+      return
+        // all but firebase-ios-sdk finalize with unobfuscated phone number which we find easily...
+        enrollment.unobfuscatedPhoneInfo === phoneNumber ||
+        // ...but firebase-ios-sdk finalizes with obfuscated
+        // This works against cloud auth, so emulator should attempt to find enrollment obfuscated as well.
+        ('+********' + enrollment.unobfuscatedPhoneInfo.substring(enrollment.unobfuscatedPhoneInfo.length - 4)) === phoneNumber;
+    }
   );
 
   const { updates, extraClaims } = await fetchBlockingFunction(


src/emulator/auth/operations.ts
```diff
@@ -2221,7 +2221,14 @@ async function mfaSignInFinalize(
...
   let { user, signInProvider } = parsePendingCredential(state, reqBody.mfaPendingCredential);
   const enrollment = user.mfaInfo?.find(
-    (enrollment) => enrollment.unobfuscatedPhoneInfo === phoneNumber,
+    (enrollment) => {
+      return
+        // all but firebase-ios-sdk finalize with unobfuscated phone number which we find easily...
+        enrollment.unobfuscatedPhoneInfo === phoneNumber ||
+        // ...but firebase-ios-sdk finalizes with obfuscated
+        // This works against cloud auth, so emulator should attempt to find enrollment obfuscated as well.
+        ('+********' + enrollment.unobfuscatedPhoneInfo.substring(enrollment.unobfuscatedPhoneInfo.length - 4)) === phoneNumber;
```
mikehardy: The code review bot makes a reasonable point.

My response / updated patch would be more verbose as suggested for readability, and re-use the existing `obfuscatePhoneNumber` method I did not realize existed.

For your consideration:

```suggestion
        // All but firebase-ios-sdk finalize with unobfuscated phone number.
        if (enrollment.unobfuscatedPhoneInfo === phoneNumber) {
          return true;
        }
        
        // But firebase-ios-sdk finalizes with an obfuscated number. This works against
        // cloud auth, so emulator should attempt to find enrollment obfuscated as well.
        if (obfuscatePhoneNumber(enrollment.unobfuscatedPhoneInfo) === phoneNumber) {
          return true;
        }
        
        return false;
```

src/emulator/auth/operations.ts
```diff
@@ -2221,7 +2221,20 @@ async function mfaSignInFinalize(
 
   let { user, signInProvider } = parsePendingCredential(state, reqBody.mfaPendingCredential);
   const enrollment = user.mfaInfo?.find(
-    (enrollment) => enrollment.unobfuscatedPhoneInfo === phoneNumber,
+    (enrollment) => {
+        // All but firebase-ios-sdk finalize with unobfuscated phone number.
```
mikehardy: linter might not be happy with this spacing nit...

```suggestion
      // All but firebase-ios-sdk finalize with unobfuscated phone number.
```

mikehardy: Actually, the linter argument with the patch is much more involved than that - with apologies, I did not run lint locally, I proposed the patch via textbox-coding in the github web UI (since my effective patch for react-native-firebase CI/CD was against the transpiled JS files...) - I can pull the branch local and run lint though if needed, though you may have a local workspace set up to do so quite efficiently - just let me know

davidjadczyk:
@mikehardy is it possible, that the issue still remains after initial MFA rollout, when you want to re-verify? Cause the initial rollout works seamlessly with the initial verification. The phonenumber is unobscured also in the console. But after I try to sign in a 2nd time, which prompts for revivification, I get a code with corresponding to the phone number obscured in the console, and entering it just results in 'resolve-signin-failed'. I have done some research and your PR was the closest I came to a helpful fix.

Also there is following issue in https://github.com/firebase/firebase-ios-sdk/issues/11079 still unresolved. Could they be related?

mikehardy:
This is on my list to retest today, glad I built it in to our test app versus testing as a one off initially!

mikehardy:
I had no problem today when I re-tested this using react-native-firebase 23.5.0 (but with updated firebase-ios-sdk 12.6.0 in it), and current firebase-tools (14.26.0). 

You can see my full test scenario with logs coming out of the emulator, emulator Auth UI visible, and logs coming out of the test app while recording - approx 3min 30secs (and reencoded to crush it down to approx 6MB) to do the full sign up + verify email + enroll phone SMS MFA + sign out + start sign in again + request code for MFA + verify code + success

it may be that our test scenarios are different thus the video - perhaps there is still something "off" here ?


https://github.com/user-attachments/assets/a7d4d885-cf0b-4ef1-8707-58f5168682b3

davidjadczyk:
@mikehardy thank you for the test. That is really odd. I did my test almost identical (just with flutter ui) and I also get the code with an  obscure phone number in the console. But in my app it is not working after the login. 

I will do more digging cause I am now not sure if its maybe a firebase_auth (flutter) issue or an implementation issue on my side, as this is the result:


piia-mfa-issue-1
piia-mfa-issue-2
piia-mfa-issue-3
piia-mfa-issue-4

If I find anything related that indicates issues about your issue, I will notify again. Thank you for your efforts!

mikehardy:
@DJ2695 you've probably already done this - but only just in case - triple check that you have the latest firebase-tools so you are running the emulator that accepts the hashed numbers from firebase-ios-sdk. Also, before that patch to the emulator *it worked against real auth in the cloud*. It always worked there as far as I know, before the patch here. Just a couple things to check in case you hadn't. Either way, does seem like a mystery. If you figure it out I'd love to know the reason - I personally worked pretty hard getting MFA working up from iOS through react-native, against real auth and emulator auth so I'm interested in the whole area. Cheers

PR #9065, opened 2025-09-01
feat(mcp, firestore): Add 'select' parameter to get_documents tool

### LINKED ISSUE #9043: [MCP] Add client-side field masking to firestore_get_documents for token limit management
### Problem
The Firebase MCP server's `firestore_get_documents` tool retrieves entire documents from Firestore, which is expected behavior. However, when documents contain large amounts of data, the MCP response exceeds the maximum allowed tokens (25K), causing the tool to fail.

### Proposed Solution
Add client-side field masking functionality to the MCP tool that filters the response **after** receiving the full document from Firestore, but **before** sending it to the LLM.

### Suggested Implementation
Add an optional `mask` parameter that filters the response on the MCP server side:

```javascript
mcp__firebase__firestore_get_documents({
  paths: ['collection/document'],
  mask: ['field1', 'field2', 'nested.field']
})
```

### Technical Details
- Firestore would still return the complete document (normal behavior)
- MCP server would apply field filtering before sending response to LLM
- This prevents token limit exceeded errors
- Maintains Firestore's standard behavior while solving MCP-specific constraints

### Use Case
Large documents (e.g., documents with large arrays, complex nested objects) cause MCP tools to fail due to response size limits. Developers need to inspect only specific fields from these documents.

### Current Behavior
- Firestore returns full document ✓
- MCP forwards entire response to LLM
- Large responses exceed 25K token limit ❌
- Tool fails with 'response exceeds maximum allowed tokens'

### Expected Behavior  
- Firestore returns full document ✓
- MCP applies field mask to response
- Filtered response stays within token limits ✓
- Tool succeeds with relevant data only

boraberkcetintas:
### Description

Adds a 'select' parameter to the `firestore_get_documents` MCP tool. This allows developers to specify which fields of a document to return, preventing token limit errors when dealing with large documents.

The `select` parameter accepts an array of field paths (e.g., ['field1', 'nested.field']). The MCP server will still fetch the full document from Firestore but will apply this field mask before sending the response to the LLM.

If `select` is not provided or is an empty array, the full document is returned. The `__path__` field is always preserved in the response for context.

### Scenarios Tested

A comprehensive suite of unit tests has been added to cover the new functionality.

### Sample Commands

firebase firestore documents get "my-collection/my-doc" --select "field1,nested.field"

Fixes  #9043


src/mcp/tools/firestore/get_documents.ts
```diff
@@ -20,6 +35,12 @@ export const get_documents = tool(
         .describe(
           "One or more document paths (e.g. `collectionName/documentId` or `parentCollection/parentDocument/collectionName/documentId`)",
         ),
+      select: z
+        .array(z.string())
+        .optional()
+        .describe(
+          "A list of field paths to return. If not specified, returns the entire document. E.g. `['field1', 'nested.field2']`.",
```
joehan:
```suggestion
          "A list of fields to return. Nested fields can be accessed by separating each field with a period (ie 'nested.nested2.field') If not specified, returns the entire document. E.g. `['field1', 'nested.field2']`.",
```

joehan (approved):
Code LGTM, but gonna have someone from the Firestore team take a look as well - the Gemini suggested change seems like it might be a good idea, but IDK if it is a hallucination or not.

boraberkcetintas:
@joehan Gemini’s idea seems fair. I developed the suggested method, but I couldn’t figure out how to use the Firestore Emulator and MCP together to test it myself. Since the Admin SDK has a select feature, I think this approach could work.

I also built the package locally with npm run build and tested it using npm link. In addition, if the Admin SDK supports it, we could also make direct API requests as an alternative.

API reference: https://firebase.google.com/docs/firestore/reference/rest/v1beta1/projects.databases.documents/batchGet

joehan (approved):

joehan:
Looks like you just need to run 'npm run format' to fix the linter issues, and then this LGTM

boraberkcetintas:
Closing - the get_documents MCP tool was removed from this repo in favor of the OneMCP proxy (#9957). The feature would need to be implemented on the OneMCP server side instead.

PR #9071, opened 2025-09-03
Improving Crashlytics MCP prompts

maxl0rd:
### Description

feat(mcp): Improving Crashlytics MCP prompt

- Consolidating the prioritize_issues and fix_issue into a single connect prompt
- Encouraging use of numbered lists to make user selection easier
- Minor formatting and copy edits


src/mcp/prompts/crashlytics/connect.ts
```diff
@@ -0,0 +1,127 @@
+import { prompt } from "../../prompt";
+
+export const connect = prompt(
+  {
+    name: "connect",
+    omitPrefix: false,
+    description: "Access a Firebase application's Crashlytics data.",
+    arguments: [
+      {
+        name: "prompt",
```
schnecle: Should we just remove this? I think this is kind of no longer necessary because you are entering into a conversation immediately with the agent, so any specific instructions will be passed along afterward. We could instead have it tell the user how it intends to prioritize crashes and let them edit those instructions?

src/mcp/prompts/crashlytics/connect.ts
```diff
@@ -0,0 +1,127 @@
...
+Once you have confirmed that the user is logged in to Firebase, and confirmed the
+id for the application that they want to access, then you can ask the user what actions
+they would like to perform. Here are some possibilities and instructions follow below:
+
+  1. Prioritize the most impactful stability issues
+  2. Diagnose and propose a fix for a crash
+
+## Instructions for Using Crashlytics Data
+
+Follow the steps below, taking note of any other instructions provided.
```
schnecle: This might just become "Follow the steps below" if we remove the additional argument.

src/mcp/prompts/crashlytics/connect.ts
```diff
@@ -0,0 +1,127 @@
...
+        * A determination of your level of confidence in the cause of the issue
+        * A determination of which library is at fault, this codebase or a dependent library
+        * A determination for how complex the fix will be
+    6b. The plan should include relevant files to change
+    6c. The plan should include a test plan to verify the fix
+    6d. Use the following format for the plan:
+
+    ## Cause
+    <A description of the root cause leading to the issue>
+      - **Confidence**: <one of "not at all confident", "somewhat confident", "confident", "very confident">
```
schnecle: I also wonder if we should remove this. I don't think I've ever seen it less than confident lol

maxl0rd: Yea

src/mcp/prompts/crashlytics/connect.ts
maxl0rd: Done

src/mcp/prompts/crashlytics/connect.ts
maxl0rd: Removed

PR #9072, opened 2025-09-03
[VS Code] Fix the no project emulator experience

fredzqm:
Previously, if a project isn't available locally:
- No local toolkit is running
- No code lens is available
- Cannot start or discover emulator

This PR addressed numerous edge cases to make sure the no project experience is smooth.

When project is missing,
- VS Code starts the toolkit, so there is error linter.
- VS Code will show the "run local" code lens. No "run prod" because project is missing.
- The emulator hub would still starts to set the locator file (e.g. `/var/folders/6x/5fwhc57j2b10smkj7m5fgsjw00k66j/T/hub-local-demo.json`). This allows VS Code to discover the projectless emulator easily.

Tested: https://screencast.googleplex.com/cast/NjAyMjU3NjcyNjg2Nzk2OHxiNzU5MDVjNC1hMA

### DIFF
--- a/firebase-vscode/src/data-connect/config.ts
+++ b/firebase-vscode/src/data-connect/config.ts
@@ -334,7 +334,10 @@ export class ResolvedDataConnectConfigs {
     const dataConnectConfig = this.findEnclosingServiceForPath(path);
     const serviceId = dataConnectConfig?.value.serviceId;
     const locationId = dataConnectConfig?.dataConnectLocation;
-    return `projects/${projectId || EmulatorHub.MISSING_PROJECT_PLACEHOLDER}/locations/${locationId}/services/${serviceId}`;
+    // FDC emulator can service multiple services keyed by serviceId.
+    // ${projectId} and ${locationId} aren't used to resolve emulator service.
+    projectId = projectId || EmulatorHub.MISSING_PROJECT_PLACEHOLDER;
+    return `projects/${projectId}/locations/${locationId}/services/${serviceId}`;
   }
 }
 


yuchenshi (changes requested):
Let's try to agree on a single default project name. I'd say `demo-local` or something explicit like `demo-no-project`.

Right now, different values are used in different places, such as `""`, `"p"` and `"local-demo"`. The last one may even conflict with a real project with that name. (`demo-*` is reserved but `local-*` is not)

fredzqm:
> Let's try to agree on a single default project name. I'd say `demo-local` or something explicit like `demo-no-project`.
> 
> Right now, different values are used in different places, such as `""`, `"p"` and `"local-demo"`. The last one may even conflict with a real project with that name. (`demo-*` is reserved but `local-*` is not)

What's your recommendation? I don't mind any of them. Just want this to work

BTW, project doesn't matter for FDC emulator. We key based on just service_id.

fredzqm:
@yuchenshi 

What do you think? There is no project. The only thing that needs to be specified is the discovery file when project is missing.

```
  let filename = `hub_no_project.json`;
    if (projectId) {
      filename = `hub-${projectId}.json`;
    }
```

Better?

firebase-vscode/src/core/emulators.ts
```diff
@@ -179,9 +179,8 @@ export class EmulatorsController implements Disposable {
   }
 
   private getHubClient(): EmulatorHubClient | undefined {
-    const projectId = firebaseRC.value?.tryReadValue?.projects?.default;
-    // TODO: think about what to without projectID, in potentially a logged out mode
-    const hubClient = new EmulatorHubClient(projectId!);
+    const projectId = firebaseRC.value?.tryReadValue?.projects?.default || "";
```
yuchenshi:
```suggestion
    const projectId = firebaseRC.value?.tryReadValue?.projects?.default;
```

firebase-vscode/src/data-connect/config.ts
```diff
@@ -305,27 +305,35 @@ export class ResolvedDataConnectConfig {
...
+    return dc;
   }
 
-  getApiServicePathByPath(projectId: string, path: string) {
+  getApiServicePathByPath(projectId: string | undefined, path: string): string {
     const dataConnectConfig = this.findEnclosingServiceForPath(path);
     const serviceId = dataConnectConfig?.value.serviceId;
     const locationId = dataConnectConfig?.dataConnectLocation;
-    return `projects/${projectId}/locations/${locationId}/services/${serviceId}`;
+    return `projects/${projectId || "p"}/locations/${locationId}/services/${serviceId}`;
```
yuchenshi: Do we still need this? If so, I'd suggest

```suggestion
    return `projects/${projectId || "demo-ignored"}/locations/${locationId}/services/${serviceId}`;
```

yuchenshi:
@fredzqm I recommend `demo-no-project` because it will be easier to tell or debug in logs or support cases later. I believe a demo project ID works better than the empty string. A few emulators like Auth and UI have troubles with no project ID and will either be skipped or throw an error. Instead of coding branching logic everywhere downstream, a fake project ID should work out of the box.

joehan (changes requested):
Let's just have this default to demo-no-project

fredzqm:
SGTM. Just changed the PR to use `demo-no-project`

src/emulator/controller.ts
```diff
@@ -851,17 +834,11 @@ export async function startAll(
...
-        "Cannot start the Pub/Sub emulator without a project: run 'firebase init' or provide the --project flag",
-      );
-    }
-
     const pubsubAddr = legacyGetFirstAddr(Emulators.PUBSUB);
     const pubsubEmulator = new PubsubEmulator({
       host: pubsubAddr.host,
       port: pubsubAddr.port,
-      projectId,
+      projectId: projectId,
```
yuchenshi: revert this

src/emulator/controller.ts
```diff
@@ -1104,14 +1080,7 @@ function getListenConfig(
...
   const projectId = options.project;
-  if (!projectId) {
-    throw new FirebaseError(
-      "Could not determine project ID, make sure you're running in a Firebase project directory or add the --project flag.",
-      { exit: 1 },
-    );
-  }
-
-  const hubClient = new EmulatorHubClient(projectId);
+  const hubClient = new EmulatorHubClient(projectId || "");
```
yuchenshi:
```suggestion
  const hubClient = new EmulatorHubClient(projectId);
```

yuchenshi (approved):

firebase-vscode/src/data-connect/config.ts
fredzqm: Yes, when `.firebaserc` is missing. `project` is undefined here.

Sure, I will use the no-project placeholder.

yuchenshi: Just add a comment that the Data Connect emulator ignores the project ID and will return the right response pretending that there's only one project

fredzqm:
@joehan I think this PR still need a LGTM from you because there was a "requested change" before.

CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- `firebase emulator:start` use a default project if no project can be found. (#9072)
```
```suggestion
- Changed `firebase emulator:start` to use a default project `demo-no-project` if no project can be found. (#9072)
```

joehan (approved):

CHANGELOG.md
fredzqm: Err, my bad. Shouldn't have enable auto-merge.

PR #9076, opened 2025-09-04
feat(functions): add remoteSource to functions config

taeold:
Add new configuration - `remoteSource` for functions in firebase.json.

`remoteSource` can be used to reference functions source code that exists in git repositories, like in GitHub. Firebase CLI will clone, discover, and deploy functions in the remote source at deploy time.

This PR only adds the new configuration without actually supporting the deploy. That will come in subsequent PRs!


src/config.ts
```diff
@@ -78,15 +78,23 @@ export class Config {
...
+    // neither a local source nor a remoteSource configured.
     if (this.get("functions")) {
       if (this.projectDir && fsutils.dirExistsSync(this.path(Config.DEFAULT_FUNCTIONS_SOURCE))) {
-        if (Array.isArray(this.get("functions"))) {
-          if (!this.get("functions.[0].source")) {
+        const funcs = this.get("functions");
+        if (Array.isArray(funcs)) {
+          // Preserve historical behavior (inject for the first entry when needed)
+          // but also ensure we don't inject if remoteSource is present.
+          const firstHasSource = this.get("functions.[0].source");
```
inlined: Why is this not applied to the whole array?

src/functions/projectConfig.ts
```diff
@@ -64,7 +109,34 @@ function validateSingle(config: FunctionConfig): ValidatedSingle {
...
   }
 
-  return { ...config, source: config.source, codebase: config.codebase };
+  // Narrow to validated shapes
+  const { codebase, prefix, ignore, configDir, predeploy, postdeploy } = config;
+  if (source) {
+    const validated: ValidatedLocalSingle = {
+      source,
+      codebase,
+      ...(runtime ? { runtime } : {}),
```
inlined: Why is this not just "runtime"? because you don't want to insert undefineds?

src/functions/projectConfig.ts
```diff
@@ -136,3 +221,39 @@ export function configForCodebase(config: ValidatedConfig, codebase: string): Va
...
+  return c;
+}
+
+/**
+ * Resolve the directory used for .env files.
+ * - Local: returns `configDir` if set, otherwise `source`.
+ * - Remote: returns `configDir` if set, otherwise `undefined`.
+ */
+export function resolveConfigDir(c: ValidatedSingle): string | undefined {
+  if (c.configDir) return c.configDir;
```
inlined: nit: why not `return c.configDir || c.source`? Why do you care about isLocalConfig which is just a check for c.source anyway?

inlined (approved):

src/config.ts
taeold: Huh that's an interesting point. Taking the opportunity to fix that behavior.

src/functions/projectConfig.ts
taeold: I blame gemini

src/functions/projectConfig.ts
taeold: yeah good call. I think I had originally had the code throw an error when remote source didn't defined configDir but decided to relax that constraint.

src/functions/projectConfig.ts
taeold: ah this was because i didn't want keys to be checked during unit testing..

PR #9077, opened 2025-09-04
feat(functions): add remoteSource utility and path resolution helper

taeold:
Add utility function required for remote source implementations.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,12 +1,12 @@
 {
   "name": "firebase-tools",
-  "version": "14.25.0",
+  "version": "14.25.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "firebase-tools",
-      "version": "14.25.0",
+      "version": "14.25.1",
       "license": "MIT",
       "dependencies": {
         "@apphosting/build": "^0.1.6",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "firebase-tools",
-  "version": "14.25.0",
+  "version": "14.25.1",
   "description": "Command-Line Interface for Firebase",
   "main": "./lib/index.js",
   "bin": {
--- a/src/appUtils.ts
+++ b/src/appUtils.ts
@@ -177,24 +177,29 @@ async function packageJsonToAdminOrWebApp(
   dirPath: string,
   packageJsonFile: string,
 ): Promise<App[]> {
-  const fullPath = path.join(dirPath, packageJsonFile);
-  const packageJson = JSON.parse((await fs.readFile(fullPath)).toString()) as PackageJSON;
-  const allDeps = getAllDepsFromPackageJson(packageJson);
-  const detectedApps = [];
-  if (allDeps.includes("firebase-admin") || allDeps.includes("firebase-functions")) {
-    detectedApps.push({
-      platform: Platform.ADMIN_NODE,
-      directory: path.dirname(packageJsonFile),
-    });
+  try {
+    const fullPath = path.join(dirPath, packageJsonFile);
+    const packageJson = JSON.parse((await fs.readFile(fullPath)).toString()) as PackageJSON;
+    const allDeps = getAllDepsFromPackageJson(packageJson);
+    const detectedApps = [];
+    if (allDeps.includes("firebase-admin") || allDeps.includes("firebase-functions")) {
+      detectedApps.push({
+        platform: Platform.ADMIN_NODE,
+        directory: path.dirname(packageJsonFile),
+      });
+    }
+    if (allDeps.includes("firebase") || detectedApps.length === 0) {
+      detectedApps.push({
+        platform: Platform.WEB,
+        directory: path.dirname(packageJsonFile),
+        frameworks: getFrameworksFromPackageJson(packageJson),
+      });
+    }
+    return detectedApps;
+  } catch (err) {
+    // If there is a malformed package.json, don't crash
+    return [];
   }
-  if (allDeps.includes("firebase") || detectedApps.length === 0) {
-    detectedApps.push({
-      platform: Platform.WEB,
-      directory: path.dirname(packageJsonFile),
-      frameworks: getFrameworksFromPackageJson(packageJson),
-    });
-  }
-  return detectedApps;
 }
 
 const WEB_FRAMEWORKS: Framework[] = Object.values(Framework);
@@ -336,6 +341,7 @@ export async function detectFiles(dirPath: string, filePattern: string): Promise
       "**/coverage/**", // Test coverage reports
     ],
     absolute: false,
+    maxDepth: 4,
   };
   return glob(`**/${filePattern}`, options);
 }
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -23,6 +23,7 @@ export interface ScheduleTrigger {
   schedule?: string;
   timeZone?: string | null;
   retryConfig?: ScheduleRetryConfig | null;
+  attemptDeadlineSeconds?: number | null;
 }
 
 /** Something that has a ScheduleTrigger */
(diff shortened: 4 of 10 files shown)


src/deploy/functions/remoteSource.ts
```diff
@@ -0,0 +1,236 @@
...
+      const origin = `${repository}@${ref}/${dir ?? ""}`;
+      if (sha) {
+        logLabeledBullet("functions", `verified functions.yaml for ${origin}; using commit ${sha}`);
+      } else {
+        logLabeledBullet("functions", `verified functions.yaml in remote source (${origin})`);
+      }
+    } catch {
+      const origin = `${repository}@${ref}/${dir ?? ""}`;
+      logLabeledBullet("functions", `verified functions.yaml in remote source (${origin})`);
+    }
```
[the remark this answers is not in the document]
taeold: good idea

src/deploy/functions/remoteSource.ts
```diff
@@ -0,0 +1,236 @@
...
+    if (attempt < retries && isTransientGitError(stderr)) {
+      await delay(backoffMs * Math.max(1, attempt + 1));
+      continue;
+    }
+    return res;
+  }
+  // Should not reach here, but return last if it exists.
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  return last!;
+}
```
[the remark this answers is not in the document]
taeold: good idea

src/deploy/functions/remoteSource.ts
```diff
@@ -0,0 +1,221 @@
...
+   * Clones a Git repo to a temporary directory and returns the absolute path
+   * to the source directory. Verifies that a `functions.yaml` manifest exists
+   * before returning.
+   *
+   * @param repository Remote Git URL (e.g. https://github.com/org/repo)
+   * @param ref Git ref to fetch (tag/branch/commit)
+   * @param dir Optional subdirectory within the repo to use
+   * @param gitClient Optional Git client for testing/injection
+   * @returns Absolute path to the checked‑out source directory
+   */
```
jhuleatt: doesn't jsdoc usually go on top of the function definition?

src/deploy/functions/remoteSource.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  const tmpDir = tmp.dirSync({
+    prefix: "firebase-functions-remote-",
+    unsafeCleanup: true,
+  });
+
+  if (!isGitAvailable()) {
+    throw new FirebaseError(
+      "Git is required to deploy functions from a remote source. " +
+        "Please install Git from https://git-scm.com/downloads and try again.",
```
jhuleatt: There's a tool called giget that can fetch code without requiring `git`. Might be worth looking at how they do it https://github.com/unjs/giget?tab=readme-ov-file#features

src/deploy/functions/remoteSource.ts
```diff
@@ -0,0 +1,221 @@
...
+ * Throws a FirebaseError with guidance if it is missing.
+ */
+export function requireFunctionsYaml(codeDir: string): void {
+  const functionsYamlPath = path.join(codeDir, "functions.yaml");
+  if (!fs.existsSync(functionsYamlPath)) {
+    throw new FirebaseError(
+      `The remote repository is missing a required deployment manifest (functions.yaml).\n\n` +
+        `For your security, Firebase requires a static manifest to deploy functions from a remote source. ` +
+        `This prevents the execution of arbitrary code on your machine during the function discovery process.\n\n` +
+        `To resolve this, clone the repository locally, inspect the code for safety, and deploy it as a local source.`,
```
jhuleatt:
```suggestion
        `If you trust this repository and want to use it anyway, clone the repository locally, inspect the code for safety, and deploy it as a local source.`,
```

src/deploy/functions/remoteSource.ts
taeold: actually going to try to remove requirement to have git installed!

src/deploy/functions/remoteSource.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+
+describe("remoteSource", () => {
+  describe("requireFunctionsYaml", () => {
+    afterEach(() => {
+      mockfs.restore();
+    });
+
+    it("should not throw if functions.yaml exists", () => {
+      mockfs({
+        "/app/functions.yaml": "runtime: nodejs16",
```
jhuleatt: just a heads up that this is a really old Node version

src/deploy/functions/remoteSource.ts
```diff
@@ -0,0 +1,135 @@
...
+import * as downloadUtils from "../../downloadUtils";
+import * as unzipModule from "../../unzip";
+
+/**
+ * Downloads a remote source to a temporary directory and returns the absolute path
+ * to the source directory. 
+ * 
+ * @param repository Remote URL (e.g. https://github.com/org/repo) or shorthand (org/repo)
+ * @param ref Git ref to fetch (tag/branch/commit)
+ * @param subDir Optional subdirectory within the repo to use
```
jhuleatt, line 20: `destDir` is missing from jsdoc. Is that the spot to put the fetched source code?

src/deploy/functions/remoteSource.ts
```diff
@@ -0,0 +1,135 @@
...
+ * @return Absolute path to the checked‑out source directory
+ */
+export async function getRemoteSource(
+  repository: string,
+  ref: string,
```
jhuleatt, line 25: could `ref` be optional and default to latest? or is that too risky?

src/deploy/functions/remoteSource.ts
```diff
@@ -0,0 +1,135 @@
...
+        subDir,
+        `Subdirectory '${subDir}' in remote source must not escape the repository root.`,
+      )
+    : rootDir;
+
+  if (subDir && !dirExistsSync(sourceDir)) {
+    throw new FirebaseError(`Directory '${subDir}' not found in repository ${repository}@${ref}`);
+  }
+
+  const origin = `${repository}@${ref}${subDir ? `/${subDir}` : ""}`;
```
jhuleatt, line 79: `path.join` might be a little safer, but not a big deal since this is just to print a message

src/deploy/functions/remoteSource.ts
```diff
@@ -0,0 +1,135 @@
...
+        rootDir,
+        subDir,
+        `Subdirectory '${subDir}' in remote source must not escape the repository root.`,
+      )
+    : rootDir;
+
+  if (subDir && !dirExistsSync(sourceDir)) {
+    throw new FirebaseError(`Directory '${subDir}' not found in repository ${repository}@${ref}`);
+  }
+
```
jhuleatt, line 78: Are there easy checks to do here to see if this is valid functions source? See if there is a package.json that has `firebase-functions` as a dependency?

src/deploy/functions/remoteSource.ts
taeold: I actually wonder if defaulting ref to `main` is an antipattern (e.g. we want users to use a "fixed" label like "v2.0.0" instead of floating label like "main")

src/deploy/functions/remoteSource.ts
taeold: for now, I think functions.yaml is all you need (which is another utility function). Later part of deploy will probably catch errors like you noted since it will try to deploy using the normal function deploy.

src/deploy/functions/remoteSource.ts
taeold: good call

src/deploy/functions/remoteSource.ts
taeold: Leaving it as is for now

src/deploy/functions/remoteSource.ts
taeold: good catch!

src/deploy/functions/remoteSource.spec.ts
taeold: silly gemini

src/deploy/functions/remoteSource.ts
taeold: oh turns out this is an URL identifier, so path.join may not work

PR #9083, opened 2025-09-05
Adding nock based tests for the API methods in src/gcp

joehan:
### Description
From Jules:

Nock based tests for API methods in src/gcp. These are a bit rote, but are helpful as breakage detectors.

### DIFF
--- a/src/gcp/cloudfunctions.ts
+++ b/src/gcp/cloudfunctions.ts
@@ -472,16 +472,6 @@ async function list(projectId: string, region: string): Promise<ListFunctionsRes
 }
 
 /**
- * List all existing Cloud Functions in a project and region.
- * @param projectId the Id of the project to check.
- * @param region the region to check in.
- */
-export async function listFunctions(projectId: string, region: string): Promise<CloudFunction[]> {
-  const res = await list(projectId, region);
-  return res.functions;
-}
-
-/**
  * List all existing Cloud Functions in a project.
  * @param projectId the Id of the project to check.
  */
--- a/src/gcp/artifactregistry.spec.ts
+++ b/src/gcp/artifactregistry.spec.ts
@@ -1,0 +1,143 @@
+import { expect } from "chai";
+import * as nock from "nock";
+import * as sinon from "sinon";
+import * as artifactRegistry from "./artifactregistry";
+import { artifactRegistryDomain } from "../api";
+import * as api from "../ensureApiEnabled";
+
+const API_VERSION = "v1";
+const PROJECT_ID = "test-project";
+const REGION = "us-central1";
+const REPO = "test-repo";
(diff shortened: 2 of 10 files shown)


src/gcp/cloudfunctions.ts
```diff
@@ -471,16 +471,6 @@ async function list(projectId: string, region: string): Promise<ListFunctionsRes
   }
 }
 
-/**
- * List all existing Cloud Functions in a project and region.
- * @param projectId the Id of the project to check.
- * @param region the region to check in.
- */
-export async function listFunctions(projectId: string, region: string): Promise<CloudFunction[]> {
```
joehan, line 479: Turns out these are unused outside of tests - safe to remove.

aalej:
The `should retry creating a user if built-in role is not read` test in `src/gcp/cloudsql/cloudsqladmin.spec.ts` seems to be failing when run locally

image

joehan:
Ok, think I fixed it - turns out the amount of time we were waiting for each retry was greater than the test timeout 🤦

aalej (approved):
LGTM! 
`
    "mocha": "nyc --reporter=html mocha 'src/gcp/artifactregistry.spec.ts' 'src/gcp/auth.spec.ts' 'src/gcp/cloudbilling.spec.ts' 'src/gcp/cloudbuild.spec.ts' 'src/gcp/cloudfunctions.spec.ts' 'src/gcp/cloudfunctions.ts' 'src/gcp/cloudfunctionsv2.spec.ts' 'src/gcp/cloudlogging.spec.ts' 'src/gcp/cloudsql/cloudsqladmin.spec.ts' 'src/gcp/cloudsql/cloudsqladmin.spec.ts'",`
    
 outputs:
 ```
 ...
   cloudlogging
    listEntries
      ✔ should resolve with a list of log entries on success
      ✔ should reject if the API call fails

  cloudsqladmin
    iamUserIsCSQLAdmin
      ✔ should return true if user has required permissions
      ✔ should return false if user does not have required permissions
      ✔ should return false on IAM error
    listInstances
      ✔ should return a list of instances on success
      ✔ should handle allowlist error
    getInstance
      ✔ should return an instance on success
      ✔ should update an instance with google ml integration
      ✔ should throw if instance is in a failed state
    instanceConsoleLink
      ✔ should return the correct console link
    createInstance
      ✔ should create an instance
    updateInstanceForDataConnect
      ✔ should update an instance
    Databases
      ✔ should list databases
      ✔ should get a database
      ✔ should create a database
      ✔ should delete a database
    Users
      ✔ should create a user
      ✔ should retry creating a user if built-in role is not ready
      ✔ should get a user
      ✔ should delete a user
      ✔ should list users


  124 passing (338ms)
```

PR #9100, opened 2025-09-09
PoC of `firebase login` tool

joehan:
### Description
Assisted by Jules:
Implement 'core_login' tool that logs you into Firebase. To use this tool, first call it with no auth code argument. It will get back a login URI, which the user should follow and get an auth code. Call the tool again with that authCode to complete the login.

### Scenarios Tested
- Run 'firebase logout'
- Run the tool via MCP inspector, complete the flow.
- Run 'firebase login', get 'already logged in' message

### DIFF
--- a/src/mcp/tools/core/login.ts
+++ b/src/mcp/tools/core/login.ts
@@ -3,11 +3,14 @@ import { tool } from "../../tool";
 import { loginPrototyper } from "../../../auth";
 import { FirebaseMcpServer } from "../../../mcp";
 import { toContent, mcpError } from "../../util";
-
+import { User, UserCredentials } from "../../../types/auth";
 const LoginInputSchema = z.object({
   authCode: z.string().optional().describe("The authorization code from the login flow"),
 });
 
+export type ServerWithLoginState = FirebaseMcpServer & {
+  authorize?: (authCode: string) => Promise<UserCredentials>;
+}
 export const login = tool(
   {
     name: "login",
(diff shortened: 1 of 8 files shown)


src/mcp/tools/index.ts
```diff
@@ -27,7 +27,7 @@ export function availableTools(activeFeatures?: ServerFeature[]): ServerTool[] {
 }
 
 const tools: Record<ServerFeature, ServerTool[]> = {
-  core: addFeaturePrefix("firebase", coreTools),
+  core: addFeaturePrefix("core", coreTools),
```
fredzqm: Emm, please revert.

This changes the name of many MCP tools from `firebase_init` to `core_init`.

src/mcp/tools/core/login.ts
```diff
@@ -0,0 +1,50 @@
...
+    name: "login",
+    description: "Logs the user into Firebase.",
+    inputSchema: LoginInputSchema,
+    _meta: {
+      requiresAuth: false,
+    },
+  },
+  async (input: z.infer<typeof LoginInputSchema>, ctx: { host: FirebaseMcpServer }) => {
+    const { authCode } = input;
+    const serverWithState = ctx.host as any;
```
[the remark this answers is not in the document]
fredzqm: Why `as any`? Use property types?

src/mcp/tools/core/login.ts
```diff
@@ -0,0 +1,50 @@
...
+        const creds = await serverWithState.__login_authorize(authCode);
+        delete serverWithState.__login_authorize;
+        return toContent(`Successfully logged in as ${creds.user.email}`);
+      } catch (e: any) {
+        delete serverWithState.__login_authorize;
+        return mcpError(`Login failed: ${e.message}`);
+      }
+    } else {
+      const prototyper = await loginPrototyper();
+      serverWithState.__login_authorize = prototyper.authorize;
```
fredzqm: Instead of a function, can we save the result there?

If `result` is defined, then we know there is a pending login_request.

src/mcp/tools/index.ts
joehan: Ahhhh, good catch, didn't see the Jules slipped this in here. Fixed

src/mcp/tools/core/login.ts
joehan: We cannot (at least without exposing the inards of the login code and having poor encapsulation). The authorize function depends on a random value that is generated on the first call - saving the function like this is much simpler than exposing that value and saving it directly

src/mcp/tools/index.ts
```diff
@@ -63,10 +63,7 @@ export function markdownDocsOfTools(): string {
 | Tool Name | Feature Group | Description |
 | --------- | ------------- | ----------- |`;
   for (const tool of allTools) {
-    let feature = tool.mcp?._meta?.feature || "";
-    if (feature === "firebase") {
-      feature = "core";
-    }
+    const feature = tool.mcp?._meta?.feature || "";
```
fredzqm: revert this. Is this intended?

src/mcp/tools/core/login.ts
```diff
@@ -0,0 +1,55 @@
...
+  authCode: z.string().optional().describe("The authorization code from the login flow"),
+});
+
+export type ServerWithLoginState = FirebaseMcpServer & {
+  authorize?: (authCode: string) => Promise<UserCredentials>;
+};
+export const login = tool(
+  {
+    name: "login",
+    description: "Logs the user into Firebase.",
```
fredzqm: nit: Logs the user into Firebase, including both CLI and MCP server.

src/mcp/tools/core/login.ts
fredzqm: SGTM. Thanks for adding types.

src/mcp/tools/index.ts
joehan: Nope, Jules at it again

PR #9116, opened 2025-09-12
Added create-firebase-app template

maneesht:
Does the following:
* Adds FDC NextJS template
* Fixes issue where if you select a react or nextjs template, the CLI doesn't detect that a generated SDK needs to be created for that app.
image
image


templates/dataconnect-nextjs/package.json
```diff
@@ -0,0 +1,11 @@
+{
```
joehan: For the sake of organization as we potentially add more of these in the future - can we put this in templates/dataconnect/nextjs? Then, we could have templates/dataconnect/angular, etc in the future and keep them clearly grouped

templates/dataconnect-nextjs/web-app/.gitignore
```diff
@@ -0,0 +1,41 @@
...
+/out/
+
+# production
+/build
+
+# misc
+.DS_Store
+*.pem
+
+# debug
```
joehan, line 27: Can we put firebase-debug.log, dataconnect-debug.log, etc... in here as well?

templates/dataconnect-nextjs/web-app/README.md
```diff
@@ -0,0 +1,36 @@
+This is a [Next.js](https://nextjs.org) project bootstrapped with [`create-next-app`](https://nextjs.org/docs/app/api-reference/cli/create-next-app).
+
+## Getting Started
+
+First, run the development server:
```
joehan:
```suggestion
First, run the development server using your package manager of choice:
```

templates/dataconnect-nextjs/web-app/README.md
```diff
@@ -0,0 +1,36 @@
...
+bun dev
+```
+
+Open [http://localhost:3000](http://localhost:3000) with your browser to see the result.
+
+You can start editing the page by modifying `app/page.tsx`. The page auto-updates as you edit the file.
+
+This project uses [`next/font`](https://nextjs.org/docs/app/building-your-application/optimizing/fonts) to automatically optimize and load [Geist](https://vercel.com/font), a new font family for Vercel.
+
+## Learn More
```
joehan, line 23: Let's add in relevant dataconnect docs links here too

templates/dataconnect-nextjs/web-app/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+To learn more about Next.js, take a look at the following resources:
+
+- [Next.js Documentation](https://nextjs.org/docs) - learn about Next.js features and API.
+- [Learn Next.js](https://nextjs.org/learn) - an interactive Next.js tutorial.
+
+You can check out [the Next.js GitHub repository](https://github.com/vercel/next.js) - your feedback and contributions are welcome!
+
+## Deploy on Vercel
```
joehan: Probably best to remove this whole section (or better yet, replace with deploy on Firebase!)

templates/dataconnect-nextjs/web-app/src/app/providers.tsx
```diff
@@ -0,0 +1,53 @@
...
+  // NOTE: Avoid useState when initializing the query client if you don't
+  //       have a suspense boundary between this and the code that may
+  //       suspend because React will throw away the client on the initial
+  //       render if it suspends and there is no boundary
+  const queryClient = getQueryClient()
+
+  return (
+    <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+  )
+}
```
joehan: Newlines at EOF?

package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Command-Line Interface for Firebase",
   "main": "./lib/index.js",
   "bin": {
-    "firebase": "./lib/bin/firebase.js"
+    "firebase": "./lib/bin/firebase.js",
+    "create-dataconnect-nextjs": "./lib/templates/index.js"
```
joehan, line 9: 😬 this is unfortunate, but putting a second bin entry screws with npx:

Previously, this will work, and we've documented it in a lot of places:
```
npx firebase-tools deploy
```
When we add a second bin, we now need to specify which to run (per https://www.npmjs.com/package/npx#description):
```
npx -p firebase firebase-tools deploy
```

src/init/features/dataconnect/sdk.ts
```diff
@@ -77,10 +77,24 @@ export async function askQuestions(setup: Setup): Promise<void> {
...
-        await createNextApp(newUniqueId("web-app", listFiles(cwd)));
+        info.apps = [
+          {
+            directory: path.resolve("./dataconnect-nextjs-app"),
+            platform: Platform.WEB,
+            frameworks: ["react"],
+          },
+        ];
+        await createNextApp();
         break;
```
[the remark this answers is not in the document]
fredzqm: This review feedback is legit. Please address~

The cmd should take a unique folder to ensure it doesn't try to override an existing folder.

src/init/features/dataconnect/sdk.ts
```diff
@@ -77,10 +77,24 @@ export async function askQuestions(setup: Setup): Promise<void> {
     });
     switch (choice) {
       case "react":
+        info.apps = [
+          {
+            directory: path.resolve("./", "web-app"),
```
fredzqm: Have you seen any issues in the prior logic when apps list is left empty?

actuate() would detect the apps just generated when apps list is empty.

A couple of reasons why I didn't include it in previous revisions:
- Make it easy to test that app detection works on the template. Note: MCP tool always runs in the actuate() with an empty apps list.
- If template generation fails (e.g. npx not found), it shouldn't setup gen SDK.

maneesht: When apps are left empty, as the PR description says, no generated SDK is configured, and so the user is left with an app that doesn't do anything. 

Also, if the user is using a "react" or "nextjs" app, they most likely want to set up the generated SDK.

If the template fails, then yes, of course, we shouldn't set up the SDKs for it, and I'll take care of that right now.

But there's basically no situation where the developer chooses to have a react/nextjs dataconnect app but doesn't want a generated sdk set up for them.

package.json
maneesht: I ended up adding a new entry as you suggested. Any way I can test that out?

src/init/features/dataconnect/sdk.ts
fredzqm: The PR description "Fixes issue where if you select a react or nextjs template, the CLI doesn't detect that a generated SDK needs to be created for that app." is incorrect.

Test out the existing code ([demo video](https://screencast.googleplex.com/cast/NjU1MDkxMTU5MjY5Mzc2MHxmMzRiYmRiMC00Zg)), which already setups generated SDK correctly.

You are right. When customers pick a template, they always want to setup gen SDK. Though it's more robust to let `detectApps` take care of it. 

https://github.com/firebase/firebase-tools/blob/e175bb6689ab9c16b24352be39f842684e05cb66/src/init/features/dataconnect/sdk.ts#L180-L183

maneesht: I removed my logic and ran into this:
image

fredzqm: Thanks for the context!

This happens because `detectApp` tries to avoid folder with dataconnect configs or generate SDK.

https://github.com/firebase/firebase-tools/blob/e175bb6689ab9c16b24352be39f842684e05cb66/src/dataconnect/appFinder.ts#L101

Yeah, we might want to narrow down the exclude list to avoid those kind of surprises. Good catch.

For the default app name, let's avoid `dataconnect`. Data connect is a feature used in the app, shouldn't be part of the default folder. Similarly, `firestore-web-app`, `storage-web-app` doesn't make sense.

Possible candidates to choose from:
- `web-app`
- `react-app`
- `next-app`

We use `web-app` and `flutter_app` today.

And BTW, this discussion reminded me of an edge case that I forgot to handle.
Template creation error shouldn't be fatal. Quick https://github.com/firebase/firebase-tools/pull/9119

templates/dataconnect/nextjs/src/firebase/init.ts
```diff
@@ -0,0 +1,4 @@
+import { initializeApp } from 'firebase/app';
+export const app = initializeApp(
+  /* Replace with sdkConfig */
+);
```
joehan: newline

maneesht: Added. Thanks!

src/init/features/dataconnect/sdk.ts
maneesht: Addressed the `web-app` concern

templates/dataconnect/nextjs/src/app/page.tsx
```diff
@@ -0,0 +1,103 @@
...
+              src/app/page.tsx
+            </code>
+            .
+          </li>
+          <li className="tracking-[-.01em]">
+            Save and see your changes instantly.
+          </li>
+        </ol>
+
+        <div className="flex gap-4 items-center flex-col sm:flex-row">
```
taeold, line 28: can we remove this or point to FAH?

joehan: +1 - actually gonna block merge on this one.

joehan (changes requested):

PR #9117, opened 2025-09-12
[VS Code] Remove unused codes associated with gca cmds

fredzqm:


fredzqm:
@hlshen Here is a clean up PR to remove dead code.

Please take a look~

PR #9119, opened 2025-09-13
[FDC init] Handle error of template create

fredzqm:
If app template creation fails for any reason (e.g. npm or flutter command is not found or stale), FDC init flow should log an error and continue without generating SDKs.

Screenshot 2025-09-12 at 9 24 49 PM

### DIFF
--- a/src/init/features/dataconnect/sdk.ts
+++ b/src/init/features/dataconnect/sdk.ts
@@ -75,18 +75,22 @@ export async function askQuestions(setup: Setup): Promise<void> {
         { name: "no", value: "no" },
       ],
     });
-    switch (choice) {
-      case "react":
-        await createReactApp(newUniqueId("web-app", listFiles(cwd)));
-        break;
-      case "next":
-        await createNextApp(newUniqueId("web-app", listFiles(cwd)));
-        break;
-      case "flutter":
-        await createFlutterApp(newUniqueId("flutter_app", listFiles(cwd)));
-        break;
-      case "no":
-        break;
+    try {
+      switch (choice) {
+        case "react":
+          await createReactApp(newUniqueId("web-app", listFiles(cwd)));
+          break;
+        case "next":
+          await createNextApp(newUniqueId("web-app", listFiles(cwd)));
+          break;
+        case "flutter":
+          await createFlutterApp(newUniqueId("flutter_app", listFiles(cwd)));
+          break;
+        case "no":
+          break;
+      }
+    } catch (err: any) {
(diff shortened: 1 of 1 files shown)


src/init/features/dataconnect/sdk.ts
```diff
@@ -75,18 +75,22 @@ export async function askQuestions(setup: Setup): Promise<void> {
...
+          await createNextApp(newUniqueId("web-app", listFiles(cwd)));
+          break;
+        case "flutter":
+          await createFlutterApp(newUniqueId("flutter_app", listFiles(cwd)));
+          break;
+        case "no":
+          break;
+      }
+    } catch (err: any) {
+      logLabeledError("dataconnect", `Failed to create a ${choice} app template`);
```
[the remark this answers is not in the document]
maneesht: Can we include this error?

src/init/features/dataconnect/sdk.ts
```diff
@@ -75,18 +75,22 @@ export async function askQuestions(setup: Setup): Promise<void> {
...
+        case "next":
+          await createNextApp(newUniqueId("web-app", listFiles(cwd)));
+          break;
+        case "flutter":
+          await createFlutterApp(newUniqueId("flutter_app", listFiles(cwd)));
+          break;
+        case "no":
+          break;
+      }
+    } catch (err: any) {
```
maneesht: I think this should be type `Error` instead of `any`

fredzqm: In the CLI code base `catch (err: any)` appeared 216 times, `catch (err: Error)` appeared no time. `catch (err: unknown)` appeared 104 times.

Consulted Gemini, it sounds like `unknown` is the recommended approach.

https://g.co/gemini/share/401c43dec1c0

src/init/features/dataconnect/sdk.ts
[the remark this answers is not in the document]
fredzqm: Added a comment to explain it.

See the screenshot above. The error is already piped to stderror. If we include it here, they will be repeated twice.

fredzqm:
@maneesht Can you take another look? Thank you!

src/init/features/dataconnect/sdk.ts
maneesht: We should *never* use `any` in TypeScript. Any existing uses should be changed to `unknown` in a different PR. `unknown` in this context is fine.

maneesht (approved):

PR #9138, opened 2025-09-17
(feat) Crashlytics tools improvements

maxl0rd:
### Description

- Making interval handling more robust
- Adding return messages to avoid undefined tool responses
- Adding BatchGetEvents

### DIFF
--- a/src/crashlytics/events.ts
+++ b/src/crashlytics/events.ts
@@ -1,7 +1,7 @@
 import { logger } from "../logger";
 import { FirebaseError, getError } from "../error";
 import { CRASHLYTICS_API_CLIENT, parseProjectNumber, TIMEOUT } from "./utils";
-import { ListEventsResponse } from "./types";
+import { BatchGetEventsResponse, ListEventsResponse } from "./types";
 import { EventFilter, filterToUrlSearchParams } from "./filters";
 
 /**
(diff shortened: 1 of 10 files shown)


src/crashlytics/events.ts
```diff
@@ -43,3 +43,44 @@ export async function listEvents(
...
+ * @param eventNames the resource names for the desired events.
+ * Format: "projects/{project}/apps/{app_id}/events/{event_id}"
+ * @return A BatchGetEventsResponse including an array of Event.
+ */
+export async function batchGetEvents(
+  appId: string,
+  eventNames: string[],
+): Promise<BatchGetEventsResponse> {
+  const requestProjectNumber = parseProjectNumber(appId);
+  if (eventNames.length > 100) throw new FirebaseError("Too many events in batchGet request");
```
schnecle, line 60: Should we even let this many come through? I feel like we shouldn't pull back more than like 30 tops to keep the context window from blowing up.

src/crashlytics/filters.ts
```diff
@@ -14,8 +14,14 @@ export const IssueIdSchema = z.string().describe("Crashlytics issue id, as hexid
 
 export const EventFilterSchema = z
   .object({
-    intervalStartTime: z.string().optional().describe(`A timestamp in ISO 8601 string format`),
-    intervalEndTime: z.string().optional().describe(`A timestamp in ISO 8601 string format.`),
+    intervalStartTime: z
+      .string()
+      .optional()
+      .describe(`A timestamp in ISO 8601 string format. Defaults to 7 days ago.`),
```
schnecle, line 20: Should we add the default value to zod?

src/crashlytics/events.ts
```diff
@@ -43,3 +43,44 @@ export async function listEvents(
...
+
+/**
+ * Get multiple events by resource name.
+ * Can be used with the `sampleEvent` resource included in topIssues reports.
+ * @param appId Firebase app_id
+ * @param eventNames the resource names for the desired events.
+ * Format: "projects/{project}/apps/{app_id}/events/{event_id}"
+ * @return A BatchGetEventsResponse including an array of Event.
+ */
+export async function batchGetEvents(
```
schnecle, line 55: It looks like we don't have a tool defined for this - do we want to use this now? Should we keep it out of here until we find we have a need for it?

src/crashlytics/events.ts
maxl0rd: I haven't written the tool yet, but I think we could limit it there. These defs map to the api's limits.

src/crashlytics/filters.ts
maxl0rd: We wouldn't want a static timestamp burned into the tool definition.

src/crashlytics/events.ts
maxl0rd: This should be fine to add the library method here, but I want to do more testing before adding the tool.

PR #9149, opened 2025-09-20
feat(mcp): Adds MCP resources and a `read_resources` tool.

mbleigh:
- Adds `/firebase:init` prompt with placeholder guidance.
- Resources are defined in `src/mcp/resources`.
- Prompts and other output can "link" to resources by saying to use the `read_resources` tool with a particular URI.


package.json
```diff
@@ -273,5 +273,6 @@
     "node-fetch": {
       "whatwg-url": "^14.0.0"
     }
-  }
+  },
+  "packageManager": "pnpm@10.13.1+sha256.0f9ed48d808996ae007835fb5c4641cf9a300def2eddc9e957d9bbe4768c5f28"
```
joehan: Revert.

src/mcp/index.ts
```diff
@@ -413,6 +426,39 @@ export class FirebaseMcpServer {
     }
   }
 
+  async mcpListResources(req: ListResourcesRequest): Promise<ListResourcesResult> {
```
joehan:
```suggestion
  async mcpListResources(_: ListResourcesRequest): Promise<ListResourcesResult> {
```

src/mcp/resources/guides/init_firestore.ts
```diff
@@ -0,0 +1,31 @@
+import { resource } from "../../resource";
+
+export const init_firestore = resource(
+  {
+    uri: "firebase://guides/init/firestore",
+    name: "firestore_init_guide",
+    title: "Firestore Init Guide",
+    description: "guides the coding agent through configuring Firestore in the current project",
+  },
+  async (uri, ctx) => {
```
joehan:
```suggestion
  async (uri,_) => {
```

src/mcp/resources/guides/init_rtdb.ts
```diff
@@ -0,0 +1,32 @@
+import { resource } from "../../resource";
+
+export const init_rtdb = resource(
```
joehan, line 3: Do we need to use underscore case here? Seems simpler to keep camel case like the rest of the CLI.

src/mcp/util/detect.ts
```diff
@@ -0,0 +1,27 @@
+import { Config } from "../../config";
+
+export type WorkspacePlatform = "web" | "android" | "ios" | "flutter" | "react-native" | "unity";
+
+export function detectWorkspacePlatform(config: Config): WorkspacePlatform | null {
```
joehan: This is super similar to getPlatformFromFolder in src/dataconnect/appFinder.ts. Can we pick one implementation and stick to that?

src/mcp/prompts/core/index.ts
```diff
@@ -1,3 +1,4 @@
+import { init } from "./init";
 import { deploy } from "./deploy";
 
-export const corePrompts = [deploy];
+export const corePrompts = [deploy, init];
```
joehan: Stick this behind mcpalpha  experiment flag for the moment (and we'll unflag it later this week)? The guide files are pretty underbaked atm, but I'd love to get this PR in so that we can loop in more folks to work on guides.

joehan (commented):
Really like the general idea here! The guides obviously need work, but the dynamic instruction loading is nice.

src/mcp/resources/guides/init_rtdb.ts
mbleigh: I was on the fence - consistent with rest of CLI vs. consistent with the other subdirectories immediately adjacent to this one. I felt like the latter was better, but can change it back if you prefer.

src/mcp/util/detect.ts
mbleigh: Didn't realize both existed, I'll use that (it should probably get hoisted out of `dataconnect` but I'll leave that for now.

src/mcp/prompts/core/index.ts
mbleigh: Done.

src/mcp/resources/guides/init_rtdb.ts
joehan: Don't care too strongly either way, just noticed that it was triggering the linter.

src/mcp/resources/guides/init_ai.ts
```diff
@@ -0,0 +1,22 @@
...
+    description:
+      "guides the coding agent through configuring GenAI capabilities in the current project utilizing Firebase",
+  },
+  async (uri) => {
+    return {
+      contents: [
+        {
+          uri,
+          type: "text",
+          text: `Create ai.ts with import { AI } from "firebase";`,
```
joehan, line 17: Suggestion from shopping this around to folks on the FDC team - WDYT about moving this content into actual md files and loading it into these files? That would make it very easy for PMs/other folks to write, and it would be simpler to reuse content from cursor rules and other files

joehan (approved):
LGTM and mergable as is, but i think we might wanna consider moving to loading from MDs.

src/mcp/resources/guides/init_rtdb.ts
mbleigh: I fixed the linter, let's keep it this way for now.

src/mcp/resources/guides/init_ai.ts
mbleigh: My suggestion is we leave it for now. PMs should be able to manipulate TS files that are primarily just big strings, and it's important to make prompts and resources able to be dynamic. I suspect most prompts will be a combination of a big static hunk of text and some dynamic bits.

If we eventually want to go down this route, I'm going to bring [Dotprompt](https://github.com/google/dotprompt) to do the job properly 😄

PR #9172, opened 2025-09-24
refactor(mcp): change consult from a tool to a prompt

mbleigh:
image


src/mcp/prompts/core/consult.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (!projectId)
+      return [
+        {
+          role: "user" as const,
+          content: {
+            type: "text",
+            text: "Inform the user that this command requires an active project to execute. Use the `firebase_update_environment` tool if the user supplies a project. After setting the project, encourage them to re-run the `firebase:consult` command.",
+          },
+        },
+      ];
```
fredzqm, line 33: I created a `requireGeminiToS` helper to ensure projectId is present, ToS is accepted and enable GiF API if needed.

Consider re-use it, so this tool can automatically enable GiF API is ToS is accepted.

We may want to tweak the error message slightly. It refers to "tool".

https://github.com/firebase/firebase-tools/blob/96eebc81232d8d063e469f9cf6c71ac65c4eb474/src/mcp/errors.ts#L12

src/mcp/prompts/core/consult.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const platform = await getPlatformFromFolder(config.projectDir);
```
fredzqm: Love the idea of providing GiF API context on what apps we have locally.

If we have multiple apps, `getPlatformFromFolder` would return a confusing `Platform.MULTIPLE`. 
To handle the multi-platform case well, I'd recommend using `detectApps` and find unique platforms from it. I meant to deprecate `getPlatformFromFolder`, but it's still used by some exploratory code.

https://github.com/firebase/firebase-tools/blob/96eebc81232d8d063e469f9cf6c71ac65c4eb474/src/dataconnect/appFinder.ts#L42

fredzqm (approved):

visumickey (approved):

src/mcp/prompts/core/consult.ts
mbleigh: Done.

src/mcp/prompts/core/consult.ts
mbleigh: Done.

joehan (approved):
LGTM, just needs linting

PR #9174, opened 2025-09-24
feat(mcp): revamp firebase init with orchestration API integration

TrCaM:
- Integration with orchestration API
- Project/App/AiLogic provisioning
- App config file writing
- Local directory conflict handling

## Conflict Handling
- **Project conflicts**: Prevents overwriting existing `.firebaserc` projects unless `overwrite_project: true`
- **Config file conflicts**: Prevents overwriting existing app config files unless `overwrite_configs: true`
- **Directory resolution**: Automatically finds existing apps or creates unique directory names

## Testing with Gemini CLI

Clone PR with GitHub CLI:
```bash
gh pr checkout 9174
```

Install and link:
```bash
npm install
npm link
```

Update `~/.gemini/settings.json` with:
```json
{
  "mcpServers": {
    "firebase": {
      "command": "npx",
      "args": [
        "-y",
        "firebase-tools",
        "experimental:mcp",
        "--dir",
        "."
      ]
    }
  }
}
```

Start Gemini CLI and try a prompt like:
```
Help me create a project and an app with ai_logic enable with firebase mcp init tool.
```

## Limitations
- Will follow up with prompt update to improve tool calling correctness. Gemini CLI takes several tries to get the input correct
- A workflow involving project creation / app creation / AI Logic enablement take up to 2.5mins to complete. Our team is investigating why the orchestration API is taking that long

### DIFF
--- a/src/mcp/tools/core/init.ts
+++ b/src/mcp/tools/core/init.ts
@@ -24,6 +24,7 @@ import { RC } from "../../../rc";
 import * as fs from "fs-extra";
 import * as path from "path";
 import { requireGeminiToS } from "../../errors";
+import { FirebaseError } from "../../../error";
 
 /**
  * Resolves app context by either finding existing app or planning new directory creation
@@ -147,6 +148,9 @@ export function buildProvisionOptions(
         case "organizations":
           projectOptions.parent = { type: "organization", organizationId: id };
           break;
+        default:
+          // This should be caught by validation, but as a safeguard:
+          throw new Error(`Unsupported parent type: ${type}`);
       }
     }
   }
@@ -450,8 +454,9 @@ export const init = tool(
         // Update context with provisioned project ID for subsequent operations
         projectId = provisionedProjectId;
       } catch (error) {
-        throw new Error(
+        throw new FirebaseError(
           `Provisioning failed: ${error instanceof Error ? error.message : String(error)}`,
+          { original: error instanceof Error ? error : new Error(String(error)), exit: 2 },
         );
       }
     }


src/mcp/tools/core/init.ts
```diff
@@ -4,136 +4,463 @@ import { toContent } from "../../util";
...
       requiresProject: false, // Can start from scratch.
-      requiresAuth: false, // Will throw error if the specific feature needs it.
+      requiresAuth: true, // Required for provisioning and Firebase operations.
     },
   },
-  async ({ features }, { projectId, config, rc }) => {
+  async ({ features, provisioning, project, app }, { projectId, config, rc }) => {
+    validateProvisioningInputs(provisioning, project, app);
+
+    if (provisioning?.enable) {
```
fredzqm, line 414: Haven't looked in-depth yet.

High-level thought is the orchestration API specific logic should be pulled out to a separate package.

Files in [commands/*](https://github.com/firebase/firebase-tools/tree/master/src) are entry point to CLI commands. Those are entry point to MCP tools. We should keep it focused on defining the input argument of the tool and delegate involved logics to other modules.

There is a great chance the orchestration API could be re-used in other part of the CLI repos.

src/init/index.ts
```diff
@@ -96,11 +97,19 @@ const featuresList: Feature[] = [
     askQuestions: features.apptestingAskQuestions,
     actuate: features.apptestingAcutate,
   },
+  {
```
fredzqm, line 100: If it's not ready to launch, we should gate it by an experiment or leave it commented out now.

src/mcp/tools/core/utils.ts
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+
+  return undefined;
+}
+
+/**
+ * Finds existing Android app with the given package name
+ */
+export async function findExistingAndroidApp(
```
fredzqm, line 150: BTW, there is a `detectApps` helper that we added for FDC.

Sounds like we are trying to achieve the same goal here.
If possible, consider re-use.

Feel free to ignore this comment if the use case is significantly different.

https://github.com/firebase/firebase-tools/blob/96eebc81232d8d063e469f9cf6c71ac65c4eb474/src/dataconnect/appFinder.ts#L42

src/management/provision.ts
```diff
@@ -0,0 +1,189 @@
+import { Client } from "../apiv2";
```
fredzqm, line 1: My 2c is there is enough logic to create a `src/management/provision` folder with modular files in it.

Glancing through the PR. We can organize logics into `client.ts`, `setup_apps.ts`, etc.

joehan: +1, especially since we expect to add more features to this in the future.

src/management/provision.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  if (options.project.parent) {
+    request.parent = buildParentString(options.project.parent);
+  }
+
+  if (options.features?.location) {
+    request.location = options.features.location;
+  }
+
+  if (options.requestId) {
```
joehan, line 122: Do we actually need all of these ifs? Isn't 
```
request.requestId = options.requestId
```
effectively equivalent, as it will just set the field to undefined if it is not included in options?

src/management/provision.ts
```diff
@@ -0,0 +1,189 @@
...
+    case "folder":
+      return `folders/${parent.folderId}`;
+    default:
+      throw new Error("Unsupported parent type");
+  }
+}
+
+export function buildProvisionRequest(
+  options: ProvisionFirebaseAppOptions,
+): Record<string, unknown> {
```
joehan, line 108: This return type is questionable - we're constructing the object ourselves so we know the keys and value types. Can you define a return type for this?

src/management/provision.ts
```diff
@@ -0,0 +1,189 @@
...
+ * @param options The provision options including project, app, and feature configurations
+ * @return Promise resolving to the provisioned Firebase app response containing config data and app resource name
+ */
+export async function provisionFirebaseApp(
+  options: ProvisionFirebaseAppOptions,
+): Promise<ProvisionFirebaseAppResponse> {
+  try {
+    const request = buildProvisionRequest(options);
+
+    const response = await apiClient.request<unknown, { name: string }>({
```
joehan, line 165: Same though as above - we build the request ourselves, it is not unknown. Use a stronger type.

src/management/provision.ts
TrCaM: Looks like Joe created the provision folder in the latest commit :). Going to resolve this

src/management/provision.ts
TrCaM: Fixed

PR #9178, opened 2025-09-24
Expose init prompt without mcpalpha experiment

samedson:


src/mcp/prompts/core/index.ts
```diff
@@ -2,9 +2,6 @@ import { init } from "./init";
 import { deploy } from "./deploy";
 import { isEnabled } from "../../../experiments";
```
joehan: Remove this, and then LGTM

joehan (approved):
LGTM after the unused dep is removed.

PR #9179, opened 2025-09-24
Add short-circuiting functionality to the init prompt

samedson:
Screenshot 2025-09-24 at 4 35 12 PM


src/mcp/prompts/core/init.ts
```diff
@@ -18,14 +19,38 @@ export const init = prompt(
     },
   },
   async ({ prompt }, { config, projectId, accountEmail }) => {
+    const shortCircuitUrls: Record<string, string> = {
+      "ai-logic": "firebase://guides/init/ai",
+      backend: "firebase://guides/init/backend",
+    };
```
mbleigh: Let's move this out of the prompt fn definition and have it be something like `const GUIDE_PARAMS = {...}` then check against the constant object in this spot..

src/mcp/prompts/core/init.ts
```diff
@@ -18,14 +19,38 @@ export const init = prompt(
...
-          text: `
+          text: shortCircuitUrl
+            ? makeShortCircuitPrompt(shortCircuitUrl)
+            : makeDefaultPrompt(platform, projectId, accountEmail, config),
+        },
+      },
+    ];
+  },
+);
+
```
mbleigh: I was imagining we would literally just programmatically load the content of the short-circuited-to guide (i.e. import `init_ai` and inline it in place of the default prompt) rather than making the model do a tool call to retrieve.

samedson: Let me try that out

src/mcp/prompts/core/init.ts
samedson: Done

samedson:
This is updated

PR #9183, opened 2025-09-24
chore(mcp): drop "experimental" from mcp command

mbleigh:
Still allows the command to be run with `experimental:` prefix to avoid breaking existing users.


joehan:
Dupe of https://github.com/firebase/firebase-tools/pull/9168 i beleive

PR #9185, opened 2025-09-25
feat(mcp): AI Logic Init Feature (CLI Command and MCP Firebase Init Tool)

TrCaM:
Simplify and temporary adaptation of https://github.com/firebase/firebase-tools/pull/9174, limit to AI Logic enablement. As more products are supported via provisioning API, we will revert this back to #9174 later. 

This PR introduces the AI Logic initialization feature, accessible via both the Firebase CLI command and the MCP Firebase init tool.

Key Features:
- **AI Logic resource provisioning**: Use the new Firebase provisioning API to create 
- **CLI Initialization**: Users can now initialize AI Logic directly using the command: `firebase init ailogic`, or via firebase_init MCP tool
- **Project Handling**: The tool will automatically create a new Firebase project if one doesn't exist, or pick up an existing project from the `.firebaserc` file.
- **App config file generation**: a Firebase app (Android/iOS/Web) will be created and its config file is returned and written into a file.
- **Configuration Overwrite**: If a configuration file already exists, users can force an overwrite by providing the `overwrite_config` input.

Limitations:
- **Extra App Creation**: If the project folder is already integrated with Firebase, an additional app will always be created. However, the AI Logic feature will still be enabled and fully functional.

## Testing with Gemini CLI

To test this feature using the Gemini CLI:

1.  **Clone PR with GitHub CLI**:
    ```bash
    gh pr checkout 9174
    ```

2.  **Install and Link**:
    ```bash
    npm install
    npm link
    ```

3.  **Update `~/.gemini/settings.json`**:
    Add the following configuration:
    ```json
    {
      "mcpServers": {
        "firebase": {
          "command": "npx",
          "args": [
            "-y",
            "firebase-tools",
            "experimental:mcp",
            "--dir",
            "."
          ]
        }
      }
    }
    ```

4.  **Start Gemini CLI and try a prompt**:
    ```
    Help me getting started with Firebase AI Logic in the current directory.
    ```


src/init/features/ailogic/index.spec.ts
```diff
@@ -0,0 +1,312 @@
...
+    let inputStub: sinon.SinonStub;
+    let confirmStub: sinon.SinonStub;
+
+    beforeEach(() => {
+      selectStub = sandbox.stub(prompt, "select");
+      inputStub = sandbox.stub(prompt, "input");
+      confirmStub = sandbox.stub(prompt, "confirm");
+    });
+
+    it("should populate ailogic featureInfo for android", async () => {
```
joehan: Nonblocking: It would be great to cover the invalid input cases here too - there is some regex in this path to assert that the app name is formatted correctly that is not meaningfully covered by this.

src/init/features/ailogic/index.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  setup.featureInfo.ailogic = {
+    appPlatform: platform,
+    appNamespace: appNamespace,
+    overwriteConfig: overwriteConfig,
+  };
+}
+
+/**
+ * AI Logic provisioning: auto-detects/creates project and app, provisions via API
```
joehan: As I read through this more, I am worried that this is pushing beyond the scope of what a feature init should be doing. During init, we first ask you what project you want to use, and then ask what features. With this change, I could get the following confusing behavior:

- I run firebase init, and create projectA
- I select FAIL from the list of products, plus a few others
- The other products get init'd using projectA
- FAIL picks up projectB from the resource, and then tries to create projectB and enable FAIl there.

We might need to rethink init more fundamentally in order to correctly use this new API

src/init/features/ailogic/index.ts
```diff
@@ -0,0 +1,149 @@
...
+      setup.projectId,
+      platform,
+      ailogicInfo.appNamespace,
+    );
+
+    // 4. Provision Firebase app
+    const response = await provisionAiLogicApp(provisionOptions);
+
+    // 5. Extract project ID and update setup
+    const projectId = extractProjectIdFromAppResource(response.appResource);
```
joehan: This step seems potentially very problematic. At this point, we may have used setup.projectId for other products.

src/init/features/ailogic/index.ts
```diff
@@ -0,0 +1,149 @@
...
+    setup.projectId = projectId;
+
+    // 6. Write config file to current directory
+    writeAppConfigFile(configFilePath, response.configData);
+
+    // 7. Update .firebaserc if project was created
+    if (setup.rcfile && setup.projectId) {
```
joehan: This should not be included in AI logic feature, it is already handled in src/init/feature/project.ts

joehan (changes requested):

src/init/features/ailogic/index.ts
TrCaM: Completely removed the ability create new project/app in the implementation. The code now check for existing project/app and only set up and enable Firebase AI Logic for the app

src/init/features/ailogic/index.spec.ts
TrCaM: Greatly simplify the code, this is outdated

src/init/features/ailogic/index.ts
```diff
@@ -0,0 +1,96 @@
...
+export interface AiLogicInfo {
+  appId: string;
+}
+
+/**
+ * Ask questions for AI Logic setup via CLI
+ */
+export async function askQuestions(setup: Setup): Promise<void> {
+  // Ask for Firebase app ID
+  const appId = await input({
```
joehan: This is kind of painful for users to have to type in - would be better to list out the existing apps and let users choose from the list - in psuedocode:

```
const apps = await listApps(projectId);

await select({
  message: "Which app would you like to use?",
  options: apps.map((app) => { return {value: app.name}})
})
```

src/init/features/ailogic/utils.ts
```diff
@@ -0,0 +1,167 @@
...
+    features: {
+      firebaseAiLogicInput: {}, // Enable AI Logic
+    },
+  };
+}
+
+/**
+ * Step 2: Verify project number matches app ID's parsed project number
+ */
+export function validateProjectNumberMatch(
```
joehan, line 69: Once we switch to a select prompt, we won't need this validation logic.

src/init/features/ailogic/utils.ts
```diff
@@ -0,0 +1,167 @@
...
+      `App ${appInfo.appId} belongs to project number ${appInfo.projectNumber} but current project has number ${projectInfo.projectNumber}.`,
+      { exit: 1 },
+    );
+  }
+}
+
+/**
+ * Step 3: Validate that app exists
+ */
+export async function validateAppExists(appInfo: AppInfo): Promise<void> {
```
joehan, line 84: Once we switch to a select prompt, we won't need this validation logic.

src/init/features/ailogic/utils.ts
```diff
@@ -0,0 +1,167 @@
...
+      exit: 1,
+      original: error instanceof Error ? error : new Error(String(error)),
+    });
+  }
+}
+
+/**
+ * Provisions Firebase app using the provisioning API
+ */
+export async function provisionAiLogicApp(
```
joehan: This function seems like an unnecessary level of wrapping to me - it takes the same arguments as provisionFirebaseApp, and has roughly the same error handling behavior. Consider using provisionFirebaseApp directly, and maybe moving this error message into provisionFirebaseApp itself

joehan (commented):
One more rounf of suggtions to simplify this code, but this looks very close

src/commands/init.ts
```diff
@@ -117,6 +117,12 @@ if (isEnabled("apptesting")) {
   });
 }
 
+choices.push({
```
joehan: Just realized that we never did an API review for this command - can we put it behind an experiment flag for now (see src/experiments.ts), and then send out an API proposal

src/init/features/ailogic/utils.ts
TrCaM: Refactored

src/init/features/ailogic/utils.ts
TrCaM: We need it for firebase MCP tool since we don't select the project/app. I moved the existing project/app validation to mcp/core/init logic

src/init/features/ailogic/utils.ts
TrCaM: Answered in another comment

src/init/features/ailogic/index.ts
TrCaM: Done!

src/commands/init.ts
TrCaM: Add an experiment. WIll draft the API review ASAP

PR #9193, opened 2025-09-26
Merge localBuild support into `firebase deploy`

annajowang:
### Description
Process localBuild source deploys configs in `firebase deploy` for apphosting.

- If `localBuild` is indicated in firebase.json, the deploy process will run a local apphosting build and upload the built app into a GCS bucket (instead of the entire source code).
- Code to actually set `localBuild` in the user's firebase.json is not included here.

### Scenarios Tested

firebase deploy --only apphosting
with `apphosting { localBuild:true }` in firebase.json.

The rollout is not expected to work on the backend yet, but a local build should be kicked off and the output should be successfully uploaded into buckets.

`firebase deploy --only apphosting` without localBuild continue to work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,0 @@
-- Deprecated Java versions below 21. Support will be dropped in Firebase CLI v15. Please upgrade to Java version 21 or above to continue using the emulators.
-- Fix Functions MCP log tool to normalize sort order and surface Cloud Logging error details (#9247)
-- Fixed an issue where `firebase init` would require log in even when no project is selected. (#9251)
--- a/README.md
+++ b/README.md
@@ -161,11 +161,17 @@ Detailed doc is [here](https://firebase.google.com/docs/cli/auth).
 
 ### Remote Config Commands
 
-| Command                        | Description                                                                                                |
-| ------------------------------ | ---------------------------------------------------------------------------------------------------------- |
-| **remoteconfig:get**           | Get a Firebase project's Remote Config template.                                                           |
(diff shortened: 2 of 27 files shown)


annajowang:
(All changes aside from some linting and tests have been code-reviewed  in the jojwang-localbuild branch)

src/deploy/apphosting/prepare.ts
```diff
@@ -144,7 +147,33 @@ export default async function (context: Context, options: Options): Promise<void
...
+
+  Object.values(context.backendConfigs).forEach(async (cfg) => {
+    if (!cfg.localBuild) {
+      return;
+    }
+    logLabeledBullet("apphosting", `Starting local build for backend ${cfg.backendId}`);
+    try {
+      const { outputFiles, annotations, buildConfig } = await localBuild(
+        options.projectRoot || "./",
+        "nextjs",
```
Yuangwang, line 159: Is it noted anywhere to the user that only nextjs is supported? What happens if they try to deploy an Angular app

Yuangwang (approved):

src/deploy/apphosting/prepare.ts
annajowang: No, that's also why this feature is not exposed to the users anywhere yet. The original intention was to have this used exclusively by Studio first, so the nextjs hardcode is fine.

The underlaying local build CAN actually support angular, but the parsing of the build output to zip and upload to GCS needs tweaking to support angular.

We will also need the micro-discovery feature to tell the under-lying local build which adapter to run. This is very far from being done. 

So the general steps of this work is:
- get an MVP working that support nextjs for team testing
- expand support beyond nextjs and MVP
- document things when it's ready for users to use directly via the CLI

PR #9200, opened 2025-09-26
[FDC] Support private IP Cloud SQL with VPC

fredzqm:
Required settings
```
  const op = await client.patch<Partial<Instance>, Operation>(
    `projects/${instance.project}/instances/${instance.name}`,
    {
      settings: {
        ipConfiguration: {
          enablePrivatePathForGoogleCloudServices: /** iff has private IP */,
        },
        databaseFlags: dbFlags,
        enableGoogleMlIntegration,
      },
    },
  );
```

### `firebase deploy` with a private IP instance
Screenshot 2025-09-26 at 4 15 50 PM

Screenshot 2025-09-26 at 4 16 16 PM

###  In a GCE VM within the VPC

Screenshot 2025-09-26 at 4 53 24 PM

Screenshot 2025-09-26 at 4 55 07 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,4 +3,4 @@
 - `firebase_update_environment` MCP tool supports accepting Gemini in Firebase Terms of Service.
 - Fixed a bug when `firebase init dataconnect` failed to create a React app when launched from VS Code extension (#9171).
 - Improved the clarity of the `firebase apptesting:execute` command when you have zero or multiple apps.
-- `firebase dataconnect:sql:migrate` now supports private IP Cloud SQL instance within its VPC. (##9200)
+- `firebase dataconnect:sql:migrate` now supports Cloud SQL instances with only private IPs. The command must be run in the same VPC of the instance to work. (##9200)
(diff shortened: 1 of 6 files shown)


joehan:
The error message for failing to connect to the private instance is pretty obscure right now. Is it feasbile to catch the error and give a bit more guidance?

fredzqm:
> The error message for failing to connect to the private instance is pretty obscure right now. Is it feasbile to catch the error and give a bit more guidance?

I added a warning about private IP only instance, hoping that gives folks some leads to investigate.

Yeah, troubleshoot why "I cannot connect CSQL" is actually a really hard problem. We don't know if private IP is the problem or something else. It's hard to know if we are actually within the VPC.

It could be your wifi is down, connectivity Cloud SQL is not stable. Cloud SQL instance is struggling to handle connection.
Surfacing the original error is probably the best we can do, so customers can dig from there.

CHANGELOG.md
```diff
@@ -3,3 +3,4 @@
 - `firebase_update_environment` MCP tool supports accepting Gemini in Firebase Terms of Service.
 - Fixed a bug when `firebase init dataconnect` failed to create a React app when launched from VS Code extension (#9171).
 - Improved the clarity of the `firebase apptesting:execute` command when you have zero or multiple apps.
+- `firebase dataconnect:sql:migrate` now supports private IP Cloud SQL instance within its VPC. (##9200)
```
yuchenshi:
```suggestion
- `firebase dataconnect:sql:migrate` now supports Cloud SQL instances with only private IPs. The command must be run in the same VPC of the instance to work. (##9200)
```

yuchenshi (approved):

PR #9201, opened 2025-09-26
Another round of prompt tweaks

joehan:
### Description
Spent some time testing and tweaking this prompt to:
- Give more explicit advice on setting up the web SDK
- Warn against putting plain text API keys in hosting public files
- Recommend using tools over commands for selecting a project


fredzqm (approved):
Glanced through them. LG

You have my thumbs up!

PR #9204, opened 2025-09-28
[VS Code] Add a code lens to refine / generate FDC operation

fredzqm:
Developer should never need to write Graphql themselves.

Natural language is FDC's true API. 😄 

https://screencast.googleplex.com/cast/NjIzOTgxMjY4MzAzODcyMHxkOWQ0MmMzZi03Mg

For new project without Gemini in Firebase, VS Code opens a modal for ToS confirmation.
Screenshot 2025-09-28 at 12 15 05 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,3 +3,4 @@
 - `firebase_update_environment` MCP tool supports accepting Gemini in Firebase Terms of Service.
 - Fixed a bug when `firebase init dataconnect` failed to create a React app when launched from VS Code extension (#9171).
 - Improved the clarity of the `firebase apptesting:execute` command when you have zero or multiple apps.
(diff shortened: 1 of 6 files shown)


hlshen (approved):
Thanks for adding the GiF tos approval! Only real concern here is that all comments will have this codeLens, but I guess there isn't much we can do about that right now.

fredzqm:
> Thanks for adding the GiF tos approval! Only real concern here is that all comments will have this codeLens, but I guess there isn't much we can do about that right now.

Yeah, good thought. We could do some heuristic to not show this code lens on comments part of a query.

Can consider in future iterations.

I also want to make "run" button work when there are incomplete GQL in other places. Right now, the whole connector must be legit for the button to work.

PR #9227, opened 2025-10-01
clarify backend service use cases and ensure user permission before de…

charlotteliang:
- clarify backend service use cases
- ensure to get user's permission before deploying to hosting site.

### DIFF
--- a/src/mcp/prompts/core/init.ts
+++ b/src/mcp/prompts/core/init.ts
@@ -80,7 +80,7 @@ IMPORTANT: The backend setup guide is for web apps only. If the user requests ba
 
 The following Firebase services are available to be configured. Use the Firebase \`read_resources\` tool to load their instructions for further guidance.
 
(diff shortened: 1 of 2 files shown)


src/mcp/prompts/core/init.ts
```diff
@@ -80,7 +80,7 @@ IMPORTANT: The backend setup guide is for web apps only. If the user requests ba
 
 The following Firebase services are available to be configured. Use the Firebase \`read_resources\` tool to load their instructions for further guidance.
 
-- [Backend Services](firebase://guides/init/backend): Read this resource to setup backend services for the user such as user authentication, database, or hosting.
+- [Backend Services](firebase://guides/init/backend): Read this resource to setup backend services for the user such as setting up a database, a user authentication sign up and login page, and deploying your app.
```
joehan:
```suggestion
- [Backend Services](firebase://guides/init/backend): Read this resource to setup backend services for the user such as a database, a user authentication sign up and login page, or deployments for static web apps.
```

joehan: Grabbed this language from Julia's revisions to `firebase init`

PR #9228, opened 2025-10-01
Fixed an issue where empty firebase.json files would break init

joehan:
### Description
@jpreid7396 caught this one - `firebase init` should work even if firebase.json is an empty file.

### Scenarios Tested
Tested init with an empty file, and with a populated one. Both work correctly.
Tested deploy with an empty file - it still errors out with 'Could not understand what targets to deploy/serve'


bkendall (changes requested):
You could add a test for this in `src/config.spec.ts` :)

PR #9233, opened 2025-10-02
[VS Code] Always show generate code lens

fredzqm:


fredzqm:
@hlshen  Here is a very small incremental improvement.

This can improve discoverability of this code lens even if the developer started without a project.

hlshen (approved):

PR #9236, opened 2025-10-02
Add common appUtils to augment get_environment with app level data

schnecle:
### Description

* Augment app utils to look for app ids in google-services files
* Augment get_environment to output app ids

### Scenarios Tested

* Added unit test coverage for get_environment as well as the new apputils
* Testing e2e locally on several different apps

### Sample Commands

```
gemini
/crashlytics:connect
```

```
gemini
"What's my firebase environment look like?"
```


src/dataconnect/appFinder.ts
```diff
@@ -17,96 +21,58 @@ export function appDescription(a: App): string {
...
+  const output: App = {
+    platform: toDataConnectPlatform(app.platform),
+    directory: app.directory,
   };
-  return glob(`**/${filePattern}`, options);
+
+  if (app.frameworks) {
+    output.frameworks = app.frameworks.map((framework) => toDataConnectFramework(framework));
+  }
+  return output;
```
fredzqm: We should probably refactor the FDC code to use structs defined in the upstream package.

To avoid adding extra tech debt, we should move `appFinder.ts` to package of choice and extend them in place.
Feel free to move the code around and refactor existing callers~

src/appUtils.ts
```diff
@@ -0,0 +1,337 @@
...
+  return `${a.directory} (${a.platform.toLowerCase()})`;
+}
+
+/**
+ * Given a directory, determine the platform type.
+ * @param dirPath The directory to scan.
+ * @return A list of platforms detected.
+ */
+export async function getPlatformsFromFolder(dirPath: string): Promise<Platform[]> {
+  const apps = await detectApps(dirPath);
```
fredzqm, line 51: Can this be simplified?

```
const platforms = apps.map((a) => a.platform);
const uniquePlatforms = [...new Set(allPlatforms)];
```

We can get rid of `Platform.NONE` and represent it using empty list.

src/appUtils.ts
```diff
@@ -0,0 +1,337 @@
...
+  return !relativePath.startsWith(`..`);
+}
+
+async function packageJsonToWebApp(dirPath: string, packageJsonFile: string): Promise<App> {
+  const fullPath = path.join(dirPath, packageJsonFile);
+  const packageJson = JSON.parse((await fs.readFile(fullPath)).toString());
+  return {
+    platform: Platform.WEB,
+    directory: path.dirname(packageJsonFile),
+    frameworks: getFrameworksFromPackageJson(packageJson),
```
fredzqm, line 173: What about web? Do we want to call `detectAppIdsForPlatform` here?

src/appUtils.ts
```diff
@@ -0,0 +1,337 @@
...
+      platform: Platform.IOS,
+      directory: iosDir,
+      appId: app.appId,
+      bundleId: app.bundleId,
+    })),
+  );
+  return iosApps.flat();
+}
+
+async function processAndroidDir(dirPath: string, filePath: string) {
```
fredzqm: Add return type.

dito for every method here.

src/appUtils.ts
```diff
@@ -0,0 +1,337 @@
...
+}
+
+/**
+ * Represents a detected application.
+ */
+export interface App {
+  platform: Platform;
+  directory: string;
+  appId?: string;
+  bundleId?: string;
```
fredzqm, line 36: Can there ever be a case where an app directory found >1 `appId` / `bundleId`?

If the underlying `detectAppIdsForPlatform` could potentially return multiple? Maybe it's safer to include all responses in the `App` struct.

We can log those anomaly at a higher level.

src/dataconnect/appFinder.ts
schnecle: I wanted to leave this in place for now to make sure I didn't break anything. I can follow up with a further refactor. Do you have strong feelings on doing it as part of this initial change?

src/appUtils.ts
schnecle: I left web out because it's actually a little tricky. You can put the web configuration in any part of your code, so in the worst we would need to read through the entire code base. :/ 

If you've got ideas for how to do this more efficiently, I'm all ears!

Technically the same is true of flutter, but there is at least a convention around it.

src/mcp/index.ts
```diff
@@ -237,6 +243,7 @@ export class FirebaseMcpServer {
   }
 
   get availableTools(): ServerTool[] {
+    this.log("error", `active features for prompts: ${this.activeFeatures || "<none>"}`);
```
maxl0rd: Not sure if this is an intentional log line or was temporary debugging.

src/mcp/tools/core/get_environment.spec.ts
```diff
@@ -0,0 +1,367 @@
...
+    return {
+      projectId: projectId || "",
+      host: server,
+      accountEmail: accountEmail ? accountEmail : null,
+      rc,
+      config,
+    };
+  };
+
+  it("should show minimal environment", async () => {
```
maxl0rd, line 60: Is there some way to assert the conditions you need without asserting the complete contents of the output? Any changes here will break a lot of tests.

src/mcp/index.ts
schnecle: Just for debugging! Good catch!

src/mcp/tools/core/get_environment.spec.ts
schnecle: I can use include statements, but that won't necessarily help us with verifying the formatting.

src/appUtils.ts
schnecle: This can and does happen. In this case, I create multiple app instances. They are unique by appId and bundleId but may have the same platform and directory.

src/appUtils.ts
schnecle: Done.

schnecle: Actually, if we remove Platform.NONE, then we run into trouble with refactoring the DataConnect stuff to use these models. Specifically, we leverage that in the sdk initialization here as a load bearing value: https://github.com/firebase/firebase-tools/blob/master/src/init/features/dataconnect/sdk.ts#L114

I like the idea of getting rid of it and using the empty list, but then I worry about messing up the dataconnect stuff. It feels like someone with better knowledge of the dataconnect repo should do that refactor.

src/appUtils.ts
schnecle: Done.

src/appUtils.ts
schnecle: I at least left a comment to that effect in the code.

src/mcp/tools/core/get_environment.spec.ts
schnecle: Made a change so that we have 2 tests - one that is verifying minimal information formatting and one for the full environment, and switched the others to just be include statements. Seems like a reasonable middle ground.

src/appUtils.ts
schnecle: Ok, I've made this change in a different branch to isolate it and we can look at it there.

src/dataconnect/appFinder.ts
fredzqm: Unfortunately, there aren't robust unit tests coverage now.
Leaving parallel data structures doesn't necessarily reduce the risks of breaking existing commands as the underlying implementation is changed.

However, they add additional sources of bugs and complexity during debugging.

Would appreciate splitting a smaller PR to do the code move and refactor. Refactor only PR are generally easier to review and ensure correctness.

schnecle: I've removed these files entirely from this PR.

src/mcp/tools/core/get_environment.ts
```diff
@@ -24,37 +25,53 @@ export const get_environment = tool(
...
-# Available Project Aliases (format: '[alias]: [projectId]')
-
-${dump(rc.projects).trim()}
-
-# Available Accounts:
-
-${dump(getAllAccounts().map((account) => account.user.email)).trim()}
+Detected App IDs: ${detectedAppsMap.size > 0 ? `\n\n${dump(Object.fromEntries(detectedAppsMap)).trim()}\n` : "<NONE>"}
+Available Project Aliases (format: '[alias]: [projectId]'): ${Object.entries(rc.projects).length > 0 ? `\n\n${dump(rc.projects).trim()}\n` : "<NONE>"}
+Available Accounts: ${hasOtherAccounts ? `\n\n${dump(allAccounts).trim()}` : "<NONE>"}
```
joehan: Can we omit this header entirely if there are no other accounts logged in? Its a pretty rare use case in the first place.

src/mcp/tools/core/get_environment.ts
```diff
@@ -24,37 +25,53 @@ export const get_environment = tool(
...
 \`\`\`json
 ${config.readProjectFile("firebase.json")}
 \`\`\``
-    : `\n# Empty Environment
+    : `\nNo firebase.json file was found.
+      
+If this project does not use Firebase tools that require a firebase.json file, no action is necessary.
 
-It looks like the current directory is not initialized as a Firebase project. The user will most likely want to:
+If this project uses Firebase tools that require a firebase.json file, the user will most likely want to:
```
joehan:
```suggestion
If this project uses Firebase services that require a firebase.json file, the user will most likely want to:
```

src/mcp/tools/core/get_environment.ts
```diff
@@ -24,37 +25,53 @@ export const get_environment = tool(
...
+  projectFileExists
+    ? `\nfirebase.json contents:
 
 \`\`\`json
 ${config.readProjectFile("firebase.json")}
 \`\`\``
-    : `\n# Empty Environment
+    : `\nNo firebase.json file was found.
+      
+If this project does not use Firebase tools that require a firebase.json file, no action is necessary.
```
joehan:
```suggestion
If this project does not use Firebase services that require a firebase.json file, no action is necessary.
```

src/appUtils.spec.ts
```diff
@@ -0,0 +1,655 @@
...
+  ],
+  "configuration_version": "1"
+}`;
+
+function cleanUndefinedFields(apps: App[]): App[] {
+  return apps.map((app) => {
+    const leanApp = Object.fromEntries(
+      Object.entries(app).filter((entry) => entry[1] !== undefined),
+    ) as App;
+    console.log(`LEAN APP: ${JSON.stringify(leanApp)}`);
```
joehan: Remove

src/appUtils.spec.ts
```diff
@@ -0,0 +1,655 @@
...
+  extractAppIdentifiersFlutter,
+  extractAppIdentifierIos,
+  extractAppIdentifiersAndroid,
+  Platform,
+  getPlatformsFromFolder,
+  detectApps,
+  App,
+} from "./appUtils";
+
+const FLUTTER_CONFIG = `
```
joehan, line 13: Rather than having these configs as constants and setting up new directories every time, we should just add the test directories to the src/test/fixtures directory

src/appUtils.ts
```diff
@@ -0,0 +1,314 @@
+import * as fs from "fs-extra";
```
joehan, line 1: This file seems to duplicate a lot of what we've built in https://github.com/firebase/firebase-tools/blob/master/src/dataconnect/appFinder.ts - we really ought to unify the 2. I recall @fredzqm was looking at doing something like this

schnecle: I have followup PR to do just that. I had some of that in here, but Fred recommended I take it out. See the conversation here: https://github.com/firebase/firebase-tools/pull/9236#discussion_r2400008522

src/appUtils.spec.ts
schnecle: Spoke about this offline - recording the decision here: 
There are a few too many permutations to make using test fixtures reasonable, but I have switched this to using mockfs to limit file creation.

src/appUtils.spec.ts
schnecle: Done.

src/mcp/tools/core/get_environment.ts
schnecle: Done.

src/mcp/tools/core/get_environment.ts
schnecle: Done.

src/mcp/tools/core/get_environment.ts
schnecle: Done.

PR #9239, opened 2025-10-02
Instrument read_resource with what resource was read

joehan:
### Description
Instrument read_resource with what resource was read.

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -49,13 +49,13 @@ let choices: {
   },
   {
     value: "apphosting",
-    name: "App Hosting: Enable web app deployments with App Hosting",
+    name: "App Hosting: Set up deployments for full-stack web apps (supports server-side rendering)",
     checked: false,
     hidden: false,
   },
   {
     value: "hosting",
-    name: "Hosting: Configure files for Firebase Hosting and (optionally) set up GitHub Action deploys",
+    name: "Hosting: Set up deployments for static web apps",
     checked: false,
   },
   {


src/mcp/tools/core/read_resources.ts
```diff
@@ -36,6 +38,7 @@ export const read_resources = tool(
 
     const out: string[] = [];
     for (const uri of uris) {
+      await trackGA4("mcp_read_resource", { resource_name: uri });
       const resolved = await resolveResource(uri, ctx);
```
mbleigh: Can you move this to a `{track: true}` option on `resolveResource` that is supplied both here and on the `mcpReadResource` method? That way if MCP clients are directly reading the resource via MCP resource calls (vs. doing a tool call) we can see that too.

src/mcp/tools/core/read_resources.ts
```diff
@@ -24,6 +25,7 @@ export const read_resources = tool(
   },
   async ({ uris }, ctx) => {
     if (!uris?.length) {
+      await trackGA4("mcp_read_resource", { resource_name: "listAvaiable" });
```
mbleigh:
```suggestion
      await trackGA4("mcp_read_resource", { resource_name: "__list__" });
```

would prefer something more like this sentinel i think, and track the same thing in the MCP ListResources method.

src/mcp/tools/core/read_resources.ts
joehan: Done.

src/mcp/tools/core/read_resources.ts
```diff
@@ -36,6 +38,7 @@ export const read_resources = tool(
 
     const out: string[] = [];
     for (const uri of uris) {
+      await trackGA4("mcp_read_resource", { resource_name: uri });
```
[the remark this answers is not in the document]
mbleigh: Agree, tho you may need to do `void trackGA4(...)`

src/mcp/resources/index.ts
```diff
@@ -27,6 +28,7 @@ export const resourceTemplates = [docs];
 export async function resolveResource(
   uri: string,
   ctx: McpContext,
+  track: boolean = true,
```
mbleigh, line 31: nit: I personally prefer options objects even if we're not sure we'll ever add more because they make the code more self-documenting, but nbd either way

src/mcp/resources/index.ts
```diff
@@ -39,16 +41,18 @@ export async function resolveResource(
   // check if an exact resource name matches first
   const resource = resources.find((r) => r.mcp.uri === uri);
   if (resource) {
+    track || (await trackGA4("mcp_read_resource", { resource_name: uri }));
```
mbleigh: Shouldn't this be `track &&` not `track ||`? (same below)

Might prefer `if (track) void trackGA4(...)` for readability.

Same below.

mbleigh (approved):

PR #9250, opened 2025-10-02
New MCP tool for running mobile tests (via app distribution).

jrothfeder:
### Description
New MCP tool for running mobile tests. Extracted business logic from appdistribution CLI so that it can be used by both that CLI and this tool.

### Scenarios Tested

Ran the equivalent of 

```
firebase appdistribution:distribute app/build/outputs/apk/debug/app-debug.apk --app <gmp-app-id> --test-devices="model=MediumPhone.arm,version=34,locale=en,orientation=portrait" --test-case-ids crash-app
```

via Gemini cli. This can be initiated by typing `appdistribution` in the prompt, gemini cli will then walk you through providing the data needed to upload a new distribution binary and run tests.

### Sample Commands
initiated by typing `appdistribution` in the prompt, gemini cli will then walk you through providing the data needed to upload a new distribution binary and run tests.


src/mcp/tools/appdistribution/tests.ts
```diff
@@ -0,0 +1,40 @@
...
+export const run_tests = tool(
+  {
+    name: "run_test",
+    description: `Upload an APK and run an existing test against it.`,
+    inputSchema: z.object({
+      appId: ApplicationIdSchema,
+      releaseBinaryFile: z.string().describe("Path to the binary release (APK)."),
+      testDevices: z.string().describe(
+        `Semicolon-separated list of devices to run automated tests on, in the format
+          'model=<model-id>,version=<os-version-id>,locale=<locale>,orientation=<orientation>'. Run 'gcloud firebase test android|ios models list' to see
```
lfkellogg, line 21: Since this is Android-only right now (incl. the fact that "APK" is referenced in the description) I think we can just use "android" instead of "android|ios".

src/mcp/tools/appdistribution/tests.ts
```diff
@@ -0,0 +1,40 @@
...
+  toAppName,
+} from "../../../appdistribution/options-parser-util";
+
+export const run_tests = tool(
+  {
+    name: "run_test",
+    description: `Upload an APK and run an existing test against it.`,
+    inputSchema: z.object({
+      appId: ApplicationIdSchema,
+      releaseBinaryFile: z.string().describe("Path to the binary release (APK)."),
```
lfkellogg, line 18: I assume you tried this out and Gemini CLI was able to find the path to the binary on its own? Is it also smart enough to try and perform a build if needed?

src/mcp/tools/appdistribution/tests.ts
```diff
@@ -0,0 +1,40 @@
...
+      testDevices: z.string().describe(
+        `Semicolon-separated list of devices to run automated tests on, in the format
+          'model=<model-id>,version=<os-version-id>,locale=<locale>,orientation=<orientation>'. Run 'gcloud firebase test android|ios models list' to see
+          available devices.`,
+      ),
+      testCaseIds: z.string().describe(`A comma-separated list of test case IDs.`),
```
lfkellogg, line 24: Since running existing test cases is out of scope for this effort (right?), I figured instead of using test case IDs you would add the ability to run a test by calling CreateReleaseTest with the AiInstructions set: http://google3/google/firebase/appdistro/v1alpha/firebase_app_distribution.proto?l=783-785

That way you would be running a one-off test without have to call CreateTestCase first.

src/appdistribution/client.ts
```diff
@@ -66,7 +66,7 @@ export class AppDistributionClient {
     });
   }
 
-  async updateReleaseNotes(releaseName: string, releaseNotes: string): Promise<void> {
+  async updateReleaseNotes(releaseName: string, releaseNotes?: string): Promise<void> {
```
tagboola, line 69: Why would you ever call `updateReleaseNotes` with `releaseNotes` set to `null`?

src/mcp/tools/appdistribution/tests.ts
```diff
@@ -0,0 +1,40 @@
...
+  {
+    name: "run_test",
+    description: `Upload an APK and run an existing test against it.`,
+    inputSchema: z.object({
+      appId: ApplicationIdSchema,
+      releaseBinaryFile: z.string().describe("Path to the binary release (APK)."),
+      testDevices: z.string().describe(
+        `Semicolon-separated list of devices to run automated tests on, in the format
+          'model=<model-id>,version=<os-version-id>,locale=<locale>,orientation=<orientation>'. Run 'gcloud firebase test android|ios models list' to see
+          available devices.`,
```
lfkellogg, line 22: I would suggest also including some suggested device configurations, so Gemini can easily run a test without having to have the gcloud CLI installed. Something like:

"model=tokay,version=36,locale=en,orientation=portrait;model=e1q,version=34,locale=en,orientation=portrait" (Pixel 9 and Galaxy S24)

Side note: we could consider adding a tool to get the catalog, so we wouldn't have to rely on the gcloud CLI. Here's the API: https://firebase.google.com/docs/test-lab/reference/testing/rest/v1/testEnvironmentCatalog/get?apix_params=%7B%22environmentType%22%3A%22ANDROID%22%7D

src/appdistribution/distribution.ts
```diff
@@ -58,3 +230,59 @@ export class Distribution {
     return this.fileName;
   }
 }
+
+async function awaitTestResults(
+  releaseTests: ReleaseTest[],
+  requests: AppDistributionClient,
+): Promise<void> {
+  const releaseTestNames = new Set(releaseTests.map((rt) => rt.name!));
```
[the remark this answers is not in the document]
tagboola: +1 to handling name being null better

src/appdistribution/options-parser-util.ts
```diff
@@ -8,7 +8,7 @@ import { FieldHints, LoginCredential, TestDevice } from "./types";
  * file and converts the input into an string[].
  * Value takes precedent over file.
  */
-export function parseIntoStringArray(value: string, file: string): string[] {
+export function parseIntoStringArray(value: string, file = ""): string[] {
```
tagboola, line 11: Why not default to null or undefined?

src/appdistribution/options-parser-util.ts
```diff
@@ -70,7 +73,7 @@ export function getAppName(options: any): string {
  * and converts the input into a string[] of test device strings.
  * Value takes precedent over file.
  */
-export function parseTestDevices(value: string, file: string): TestDevice[] {
+export function parseTestDevices(value: string, file = ""): TestDevice[] {
```
tagboola, line 76: Same here

src/mcp/tools/appdistribution/tests.ts
tagboola, line 24: Since we're running on-off tests, I don't think we want to require testCaseIds since that would mean we are supporting running existing tests. If you're just keeping this here for now for simplicity and plan to update it in a future change I'm good with that.

src/mcp/tools/appdistribution/tests.ts
```diff
@@ -0,0 +1,40 @@
...
+} from "../../../appdistribution/options-parser-util";
+
+export const run_tests = tool(
+  {
+    name: "run_test",
+    description: `Upload an APK and run an existing test against it.`,
+    inputSchema: z.object({
+      appId: ApplicationIdSchema,
+      releaseBinaryFile: z.string().describe("Path to the binary release (APK)."),
+      testDevices: z.string().describe(
```
tagboola, line 19: Can we make this optional and provide a default? I don't think we want to require the user to specify a device.

src/appdistribution/client.ts
jrothfeder: I agree but this is how the code already worked.

src/appdistribution/options-parser-util.ts
jrothfeder: Is null or undefined preferable to the empty string in TS?

src/mcp/tools/appdistribution/tests.ts
jrothfeder: Yeah, I was thinking we could support both on-off and existing and this would be a good start. If you're comfortable, I'd like to see if we can get this PR in and then build the on-off functionality separately.

src/mcp/tools/appdistribution/tests.ts
jrothfeder: Makes sense. I'll update the tool to send the following by default: `model=MediumPhone.arm,version=34,locale=en,orientation=portrait`

src/mcp/tools/appdistribution/tests.ts
lfkellogg: I'm not sure it makes sense to give Gemini access to a tool that requires test case IDs unless we also give it the ability to at least list test cases, and preferably to create test cases as well.

src/mcp/tools/appdistribution/tests.ts
jrothfeder: Yep, it finds them (in most cases there are multiple to choose from). I haven't tested the case when there isn't a binary yet.

src/mcp/tools/appdistribution/tests.ts
jrothfeder: I agree that it's technically out of scope, however, it is far simpler to do so I figured we'd support both. 

I was thinking we'd get this functionality in and then add the ability for one-off.

src/appdistribution/client.ts
jrothfeder: The code around parameter processing needs to be refactored to simplify all of this. All I did was move it to a shared-library. Would you be comfortable with addressing the parameter processing in a separate PR?

src/appdistribution/options-parser-util.ts
jrothfeder: yeah, again I think we should revisit all of this. The existing code is a bit overcomplicated with how it handles parameters.

src/mcp/tools/appdistribution/index.ts
```diff
@@ -0,0 +1,4 @@
+import type { ServerTool } from "../../tool";
+import { run_tests } from "./tests";
+
+export const appdistributionTools: ServerTool[] = [run_tests];
```
tagboola, line 4: Can we place these behind the mcpalpha experiment?

src/mcp/tools/appdistribution/tests.ts
jrothfeder: Fair enough.

src/mcp/tools/appdistribution/tests.ts
jrothfeder: I'll add this as the default. 

Getting the catalog is a good idea. Maybe we'll flesh this out a bit more when we wire this up in to the prompt (that @tagboola is working on)?

src/appdistribution/client.ts
tagboola: I think leave this the same i.e. don't make `releaseNotes` nullable and then in the code where you call it, only call it if `releaseNotes` is defined. I feel like that isn't too much of a change and can easily be addressed here.

src/appdistribution/options-parser-util.ts
tagboola: Good question. I guess my Kotlin mind just assumed null was prefered, but maybe empty string is in TS.

src/mcp/tools/appdistribution/tests.ts
tagboola: I think with zod you can set a default value

https://zod.dev/api?id=defaults

src/mcp/tools/appdistribution/tests.ts
tagboola: I'm happy to add it as a TODO, don't know if we'll add it in the short term though.

PR #9251, opened 2025-10-03
[VS Code] Support projectless & loginless emulator experience

fredzqm:
`firebase init` requires `firebase login`.

This is a reasonable requirement for most use cases.
However, if you want to try out the emulator experience without a project, you shouldn't need to login either.

This PR supports `firebase init --project demo-no-project`.

### DIFF
--- a/src/init/features/project.ts
+++ b/src/init/features/project.ts
@@ -15,6 +15,7 @@ import { logger } from "../../logger";
 import * as utils from "../../utils";
 import * as prompt from "../../prompt";
 import { Options } from "../../options";
+import { EmulatorHub } from "../../emulator/hub";
 
 const OPTION_NO_PROJECT = "Don't set up a default project";
 const OPTION_USE_PROJECT = "Use an existing project";
@@ -100,6 +101,10 @@ async function projectChoicePrompt(options: any): Promise<FirebaseProjectMetadat
  */
 export async function doSetup(setup: any, config: any, options: any): Promise<void> {
   setup.project = {};
+  if (options.projectId === EmulatorHub.MISSING_PROJECT_PLACEHOLDER) {
+    logger.info(`Skipping Firebase project given --project=${options.projectId}`);
+    return;
+  }
 
   logger.info();
   logger.info(`First, let's associate this project directory with a Firebase project.`);
--- a/firebase-vscode/src/core/index.ts
+++ b/firebase-vscode/src/core/index.ts
@@ -15,7 +15,7 @@ import { registerWebhooks } from "./webhook";
 import { createE2eMockable } from "../utils/test_hooks";
 import { runTerminalTask } from "../data-connect/terminal";
 import { AnalyticsLogger } from "../analytics";
-import { EmulatorHub } from "../emulator/hub";
+import { EmulatorHub } from "../../../src/emulator/hub";
 
 export async function registerCore(
   broker: ExtensionBrokerImpl,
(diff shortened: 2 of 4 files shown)


fredzqm:
@joehan Hope you are less busy today.

Wonder what's your take on this.

There are many ways (interactive & non-interactive) to skip project setup in Firebase init. Though login is always required.

It's reasonable to expect FDC emulator experience to work e2e without login and project.

Given our prior art to use `demo-no-project`, do you think we can use it to skip both auth and project in init?

src/init/features/project.ts
```diff
@@ -100,6 +102,11 @@ async function projectChoicePrompt(options: any): Promise<FirebaseProjectMetadat
  */
 export async function doSetup(setup: any, config: any, options: any): Promise<void> {
   setup.project = {};
+  if (options.project === EmulatorHub.MISSING_PROJECT_PLACEHOLDER) {
```
joehan: We can just use the isDemoProject util here. That will also keep the codebase a little less entangled

joehan:
> @joehan Hope you are less busy today.
> 
> Wonder what's your take on this.
> 
> There are many ways (interactive & non-interactive) to skip project setup in Firebase init. Though login is always required.
> 
> It's reasonable to expect FDC emulator experience to work e2e without login and project.
> 
> Given our prior art to use `demo-no-project`, do you think we can use it to skip both auth and project in init?

I think this is a totally valid use case! In general, init shouldn't require auth unless you choose a project, and if there is no project/user, it should just write files and skip any provisioning. This current implementation seems quite laser focused on the VSCE use case, but i'd prefer to expand it to cover all flows.

src/init/features/project.ts
```diff
@@ -100,6 +102,11 @@ async function projectChoicePrompt(options: any): Promise<FirebaseProjectMetadat
...
+    logger.info(`Skipping Firebase project setup given --project=${options.project}`);
+    return;
+  }
+  await requireAuth(options);
```
joehan: This seems too early to requireAuth - the user hasn't even selected a project yet. Ideall,y we should do this after they choose 'create new project' or 'use existing', and skip it if they say 'no project'

fredzqm: That is a great point!
Today, we require auth upfront though.
Let me improve it more

fredzqm: Done. Moved it after the prompt.

Tested with personal accounts.

src/init/features/project.spec.ts
```diff
@@ -94,26 +96,6 @@ describe("project", () => {
...
-          await doSetup(setup, emptyConfig, options);
-        } catch (e: any) {
-          err = e;
-        }
-
-        expect(err.message).to.equal("Project ID cannot be empty");
-        expect(prompt.select).to.be.calledOnce;
-        expect(prompt.input).to.be.calledTwice;
-        expect(createFirebaseProjectStub).to.be.not.called;
-      });
```
fredzqm, line 116: Still work. The validation is at prompt layer.

Image

src/init/features/project.spec.ts
```diff
@@ -137,27 +119,6 @@ describe("project", () => {
...
-          await doSetup(setup, emptyConfig, options);
-        } catch (e: any) {
-          err = e;
-        }
-
-        expect(err.message).to.equal("Project ID cannot be empty");
-        expect(prompt.select).to.be.calledOnce;
-        expect(promptAvailableProjectIdStub).to.be.calledOnce;
-        expect(addFirebaseProjectStub).to.be.not.called;
-      });
```
fredzqm, line 160: This still works. I moved the validation into `promptAvailableProjectId`.

CHANGELOG.md
```diff
@@ -1 +1,2 @@
 - Fix Functions MCP log tool to normalize sort order and surface Cloud Logging error details (#9247)
+- `firebase init` only requires `firebase login` when a valid project is passed. It accepts demo projects without login. (#9251)
```
joehan:
```suggestion
- Fixed an issue where `firebase init` would require log in even when no project is selected. (#9251)
```

PR #9256, opened 2025-10-03
[FDC] Ask for region if app description is given

fredzqm:
1. Ask for region if app description is given. Missed this in the past, we create "unlinked" services in us-central1 before.
2. Changed default region to be `us-east4` in template as well.

### DIFF
--- a/src/init/features/dataconnect/index.ts
+++ b/src/init/features/dataconnect/index.ts
@@ -40,6 +40,9 @@ import {
 import { configstore } from "../../../configstore";
 import { trackGA4 } from "../../../track";
 
+// Default GCP region for Data Connect
+const DEFAULT_GCP_REGION = "us-east4";
+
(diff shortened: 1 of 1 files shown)


fredzqm:
@yuchenshi Small fix PR. Mind taking a look? Thank you!

src/init/features/dataconnect/index.ts
```diff
@@ -642,6 +642,17 @@ async function promptForCloudSQL(setup: Setup, info: RequiredInfo): Promise<void
   return;
 }
 
+async function promptForLocation(setup: Setup, info: RequiredInfo): Promise<void> {
+  if (info.locationId === "") {
+    const choices = await locationChoices(setup);
+    info.locationId = await select<string>({
+      message: "What location would you like to use?",
```
yuchenshi:
```suggestion
      message: "What location should the new Cloud SQL instance be in?",
```

fredzqm: Will need to revert this suggestion because we ask for region even when CSQL isn't being created.

PR #9274, opened 2025-10-07
Implement dataconnect:execute command.

yuchenshi:
### Description

This adds a new command to execute Data Connect operations.

### Scenarios Tested

image

See Sample Commands below for more.

### Sample Commands

```sh
firebase dataconnect:execute MyOperation
# REMOVED for now: firebase dataconnect:execute my-connector MyOperation
firebase dataconnect:execute ./my-file.gql
firebase dataconnect:execute ./my-file.gql MyOperation
```

### DIFF
--- a/src/dataconnect/load.ts
+++ b/src/dataconnect/load.ts
@@ -2,6 +2,7 @@ import * as path from "path";
 import * as fs from "fs-extra";
 import * as clc from "colorette";
 import { glob } from "glob";
+
 import { Config } from "../config";
 import { FirebaseError } from "../error";
 import {


src/commands/dataconnect-execute.ts
```diff
@@ -0,0 +1,306 @@
...
+        throw new FirebaseError("Got invalid response body with neither .data or .errors", {
+          context: {
+            body: response.body,
+            response: response,
+          },
+          status: response.status,
+        });
+      }
+
+      // Log the body to stdout to allow pipe processing (even with .errors).
```
yuchenshi, line 165: This doesn't work quite nicely since errors thrown (below, if any) are also logged to stdout (globally handled). Piping to `jq` for example can choke on the extra output

src/commands/dataconnect-execute.ts
```diff
@@ -0,0 +1,306 @@
...
+        const dir = relative(process.cwd(), connector.directory);
+        process.stderr.write(`${clc.cyan(`Executing ${opDisplay} in ${clc.bold(dir)}`)}${EOL}`);
+        query = squashGraphQL(connector.connector.source);
+        if (!query) {
+          throw new FirebaseError(`${dir} contains no GQL files or only empty ones`);
+        }
+      } else {
+        if (fileOrConnector === "-") {
+          stdinUsedFor = "operation source code";
+          if (process.stdin.isTTY) {
```
yuchenshi: As discussed internally, I tried to throw an error if `options.nonInteractive` but that creates a false positive when `--json` is passed, since `--json` right now implies `options.nonInteractive` (handled globally).

I don't think erroring on `--json` matches user intention, so I left that part out.

fredzqm:
Can you limit the scope of this PR to a minimum set of API that has no ongoing debates? Probably will make it easier to make progress and unblock the primary seed data CUJ in quick start.

```
# Yes, please!
firebase dataconnect:execute ./my-file.gql

# Let's defer.
# The tests CreatePlayer example made me nervous. tests appear like the folder name for someone new to FDC.
firebase dataconnect:execute my-connector MyOperation

# Prefer to defer it. Though won't block on it.
# There is less ambiguity between file and operation name, we can detect based on convention.
# Though the alternative `-n MyOperation` is simple, unambitious. It also allows taking multiple files. Multi seed file is a super common use case.
firebase dataconnect:execute MyOperation
firebase dataconnect:execute ./my-file.gql MyOperation
```

yuchenshi:
Removed connector and `--impersonate`. Keeping operation name as a positional argument per offline discussion

src/commands/dataconnect-execute.ts
```diff
@@ -0,0 +1,278 @@
+import * as clc from "colorette";
```
joehan, line 1: Super nit: Prefer to split internal and external imports - ie :
```
import * as clc from "colorette";
import { readFile } from "node:fs/promises";
import { EOL } from "node:os";
//...
import { Command } from "../command";
//...
```

src/commands/dataconnect-execute.ts
```diff
@@ -0,0 +1,278 @@
...
+import { EOL } from "node:os";
+import { relative } from "node:path";
+import { text } from "node:stream/consumers";
+import { logger } from "../logger";
+import { responseToError } from "../responseToError";
+
+let stdinUsedFor: string | undefined = undefined;
+
+export const command = new Command("dataconnect:execute [file] [operationName]")
+  .description("execute a Data Connect query or mutation")
```
joehan: Consider explicitly mentioning that this can be used on the emulator?
```suggestion
  .description("execute a Data Connect query or mutation. If FIREBASE_DATACONNECT_EMULATOR_HOST is set (such as during 'firebase emulator:exec', executes against the emulator instead.")
```

src/commands/dataconnect-execute.ts
```diff
@@ -0,0 +1,278 @@
...
+function parseJsonObject(json: string, subject: string): Record<string, any> {
+  try {
+    const obj = JSON.parse(json || "{}") as unknown;
+    if (typeof obj !== "object" || obj == null)
+      throw new Error(`Provided ${subject} is not an object`);
+    return obj;
+  } catch (e) {
+    throw new FirebaseError(`expected ${subject} to be valid JSON string, got: ${json}`);
+  }
+}
```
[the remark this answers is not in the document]
yuchenshi: Fixed

src/dataconnect/load.ts
```diff
@@ -180,3 +181,26 @@ function toFile(sourceDir: string, fullPath: string): File {
...
+ */
+export function squashGraphQL(source: Source): string {
+  if (!source.files || !source.files.length) {
+    return "";
+  }
+  if (source.files.length === 1) {
+    return source.files[0].content;
+  }
+  let query = "";
+  for (const f of source.files) {
```
joehan, line 198: Highly tempted to do this with functional programming as .filter(...).map(...).reduce(...), but this probably reads cleaner as is.

src/commands/dataconnect-execute.ts
```diff
@@ -0,0 +1,279 @@
...
+        if (!process.stdin.isTTY) {
+          file = "-";
+        } else {
+          throw new FirebaseError(
+            "At least one of the [file] [operationName] arguments is required.",
+          );
+        }
+      }
+      let query: string;
+      if (file === "-") {
```
joehan, line 68: The deeply nested logic here is a bit scary to reason through. If feasible, it would be nice to split into a few helpers to handle the different cases (ie readGQLFromDirectroy, readGQLFromFile ...)

I understand that this complexity is a direct consequence of the decision to make this command accept a bunch of different input formats, I just worry that this code will be fragile and hard to reason about.

yuchenshi: Better now?

PR #9275, opened 2025-10-07
Fix misleading typing for options.json.

yuchenshi:
### Description

`options.json` has always been unset since at least 5 years ago but the typing
in `options.ts` says that it exists. This led to two kinds of mistakes.

1. Using `if (options.json)` in command handlers. This effectively creates a
    dead branch. I've tried my best to remove those -- should be noop.
1. Providing `json: false` when Options are required. It's a not a
    noop strictly-speaking but the tests should still pass since it is not
    actually accessed in commands. (These are mostly found in tests.)

----

Alternatives considered: Setting `options.json` for real seems appealing but
branching on `options.json` is almost always a mistake. Unless in some obscure
use cases (not found in the code base), it is best to let the global handling
deal with `--json` the flag.

Things taken care of by global handling:

1. Suppressing normal output: https://github.com/firebase/firebase-tools/blob/a952fb4067891bfd18c1c0fccecd82dc5dd7a8d5/src/command.ts#L315-L319
1. Serializing the return value to JSON: https://github.com/firebase/firebase-tools/blob/a952fb4067891bfd18c1c0fccecd82dc5dd7a8d5/src/command.ts#L211

Note that the last step wraps the return value in a wrapper object:
`{status: "success", result: ...}`. The dead branches I've removed all
failed to account for this. So if we had made `options.json` a real boolean,
it'd cause quite a few breaking changes unless we fix those one by one.

### Scenarios Tested

Existing unit tests. I've also logged `options.json` in commands and made sure
it is always unset regardless of flags.

### Sample Commands

N/A

### DIFF
--- a/firebase-vscode/src/options.ts
+++ b/firebase-vscode/src/options.ts
@@ -26,7 +26,6 @@ const defaultOptions: Readonly<VsCodeOptions> = {
   projectNumber: "",
   projectRoot: "",
   account: "",
-  json: true,
   nonInteractive: true,
   interactive: false,
   debug: false,
--- a/src/command.ts
+++ b/src/command.ts
@@ -313,6 +313,7 @@ export class Command {
     }
 
     if (getInheritedOption(options, "json")) {
+      options.interactive = false;
       options.nonInteractive = true;
     } else if (!options.isMCP) {
       useConsoleLoggers();
--- a/src/commands/firestore-backups-list.ts
+++ b/src/commands/firestore-backups-list.ts
@@ -23,17 +23,15 @@ export const command = new Command("firestore:backups:list")
     const listBackupsResponse: ListBackupsResponse = await listBackups(options.project, location);
     const backups: Backup[] = listBackupsResponse.backups || [];
 
-    if (options.json) {
-      logger.info(JSON.stringify(listBackupsResponse, undefined, 2));
-    } else {
-      printer.prettyPrintBackups(backups);
-      if (listBackupsResponse.unreachable && listBackupsResponse.unreachable.length > 0) {
-        logWarning(
-          "We were not able to reach the following locations: " +
-            listBackupsResponse.unreachable.join(", "),
-        );
-      }
+    printer.prettyPrintBackups(backups);
+    if (listBackupsResponse.unreachable && listBackupsResponse.unreachable.length > 0) {
+      logWarning(
+        "We were not able to reach the following locations: " +
+          listBackupsResponse.unreachable.join(", "),
+      );
     }
 
+    // TODO: Consider returning listBackupResponse instead for --json. This will
+    // be a breaking change but exposes .unreachable, not just .backups.
     return backups;
   });
--- a/src/requireConfig.spec.ts
+++ b/src/requireConfig.spec.ts
@@ -14,7 +14,6 @@ const options: Options = {
   config: new Config({}),
   filteredTargets: [],
   force: false,
-  json: false,
   nonInteractive: false,
   interactive: false,
   debug: false,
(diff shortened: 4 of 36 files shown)


firebase-vscode/src/options.ts
```diff
@@ -26,7 +26,6 @@ const defaultOptions: Readonly<VsCodeOptions> = {
   projectNumber: "",
   projectRoot: "",
   account: "",
-  json: true,
```
yuchenshi, line 29: The VSCE right now doesn't call Firestore commands. So I believe this should be a noop.

src/command.ts
```diff
@@ -313,6 +313,7 @@ export class Command {
     }
 
     if (getInheritedOption(options, "json")) {
+      options.interactive = false;
```
yuchenshi, line 316: This is a real logic bug but it seems to only affect one command. And that command usually isn't called with `--json`.

https://github.com/firebase/firebase-tools/blob/a952fb4067891bfd18c1c0fccecd82dc5dd7a8d5/src/init/features/aitools/promptUpdater.ts#L89

src/commands/firestore-backups-list.ts
```diff
@@ -23,17 +23,15 @@ export const command = new Command("firestore:backups:list")
     const listBackupsResponse: ListBackupsResponse = await listBackups(options.project, location);
     const backups: Backup[] = listBackupsResponse.backups || [];
 
-    if (options.json) {
-      logger.info(JSON.stringify(listBackupsResponse, undefined, 2));
```
yuchenshi, line 27: This branch isn't consistent with the `return backups`. I believe the intention was to print `listBackupsResponse` when `--json` is used. But fulfilling that intention will cause a breaking change. So I kept it WAI (working as implemented) and left a TODO instead

src/requireConfig.spec.ts
```diff
@@ -14,7 +14,6 @@ const options: Options = {
   config: new Config({}),
   filteredTargets: [],
   force: false,
-  json: false,
   nonInteractive: false,
   interactive: false,
```
yuchenshi, line 18: Most of these mock options have both `nonInteractive` and `interactive` set to false. This is clearly a mistake. Lmk if you'd like me to fix it in a follow-up PR.

src/command.ts
joehan: Not too worried here - prety confident that --json on init would blow up in other ways due to supressing logging + prompting the user

src/requireConfig.spec.ts
joehan: Go for it!

joehan (approved):

joehan:
+ @rwhogg  FYI, for future commands

src/commands/firestore-backups-list.ts
bw-yuhogg: That was indeed the intention. It's possible that I didn't notice this when this was implemented the first time... sorry.

rwhogg (approved):

src/commands/firestore-backups-list.ts
yuchenshi: No problem. The next major (breaking) release of the CLI should be coming up soon and feel free to make it return `listBackupsResponse` by then

src/requireConfig.spec.ts
yuchenshi: https://github.com/firebase/firebase-tools/pull/9324

src/commands/firestore-backups-list.ts
yuchenshi: https://github.com/firebase/firebase-tools/pull/9392

PR #9280, opened 2025-10-08
Use a datastructure to represent test devices rather than a string.

jrothfeder:


src/mcp/tools/apptesting/tests.ts
```diff
@@ -1,40 +1,54 @@
...
-} from "../../../appdistribution/options-parser-util";
+import { toContent } from "../../util";
+import { parseIntoStringArray, toAppName } from "../../../appdistribution/options-parser-util";
+
+const TestDeviceSchema = z
+  .object({
+    model: z.string(),
+    version: z.string(),
+    locale: z.string(),
+    orientation: z.enum(["portrait", "landscape"]),
```
tagboola, line 14: Are any of these fields optional because there are default values? If so you may want to use `.optional()` and/or `.default()` methods

src/mcp/tools/apptesting/tests.ts
```diff
@@ -1,40 +1,54 @@
...
+        },
+      ]),
       testCaseIds: z.string().describe(`A comma-separated list of test case IDs.`),
     }),
   },
   async ({ appId, releaseBinaryFile, testDevices, testCaseIds }) => {
-    if (!appId) return mcpError(`Must specify 'appId' parameter.`);
-    if (!releaseBinaryFile) return mcpError(`Must specify 'releaseBinaryFile' parameter.`);
-    if (!testDevices) return mcpError(`Must specify 'testDevices' parameter.`);
-    if (!testCaseIds) return mcpError(`Must specify 'testCaseIds' parmeter.`);
```
tagboola, line 31: Why remove these errors?

jrothfeder: Github's AI reviewer said they could be removed because the schema's default is "required".

src/mcp/tools/apptesting/tests.ts
jrothfeder: No, these fields aren't optional.

PR #9282, opened 2025-10-08
feat(dataconnect): add confirmation for Gemini schema generation

fredzqm:
Instead of directly asking for an app description to generate a schema with Gemini, this change first asks the user to confirm if they want to use Gemini.

If the user confirms, it then prompts for the app description with a default value of "an app for ${setup.projectId}".

### DIFF
--- a/src/init/features/dataconnect/index.ts
+++ b/src/init/features/dataconnect/index.ts
@@ -118,9 +118,16 @@ export async function askQuestions(setup: Setup): Promise<void> {
           "Learn more about Gemini in Firebase and how it uses your data: https://firebase.google.com/docs/gemini-in-firebase#how-gemini-in-firebase-uses-your-data",
         );
       }
-      info.appDescription = await input({
-        message: `Describe your app to automatically generate a schema with Gemini [Enter to use a template]:`,
+      const wantToGenerate = await confirm({
+        message: "Do you want to generate a schema with Gemini?",
+        default: true,
       });
+      if (wantToGenerate) {
+        info.appDescription = await input({
+          message: `Describe your app to automatically generate a schema with Gemini:`,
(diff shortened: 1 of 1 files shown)


src/init/features/dataconnect/index.ts
```diff
@@ -118,12 +119,19 @@ export async function askQuestions(setup: Setup): Promise<void> {
...
+      const wantToGenerate = await confirm({
+        message: "Do you want to generate schema and example queries with Gemini?",
+        default: false,
       });
-      if (info.appDescription) {
+      if (wantToGenerate) {
         configstore.set("gemini", true);
         await ensureGIFApiTos(setup.projectId);
+        info.appDescription = await input({
+          message: `Describe your app ideas [leave blank to be creative]:`,
```
hlshen: nit: idea*

Also create isn't the clearest. Maybe "Describe your app idea [leave blank for a random one]"?

fredzqm: Thank you!

Chatted with Tyler offline. Changing it to require customer to enter one.

src/init/features/dataconnect/index.ts
```diff
@@ -646,7 +656,7 @@ async function promptForLocation(setup: Setup, info: RequiredInfo): Promise<void
   if (info.locationId === "") {
     const choices = await locationChoices(setup);
     info.locationId = await select<string>({
-      message: "What location should the new Cloud SQL instance be in?",
+      message: "What location would you like to use?",
```
fredzqm, line 664: Had to revert it because we ask this even when CSQL isn't being provisioned.

fredzqm:
@joehan Can you take a look? Thanks!

Adding this confirmation based on Tyler's feedback in Tuesday's bugbash.

PR #9284, opened 2025-10-08
Stop following symlinks when archiving directories

joehan:
### Description
For security reasons, we should stop following symlinks when archiving directories. Considering whether this needs to be a breaking change and part of the next major release.

### DIFF
--- a/src/archiveDirectory.ts
+++ b/src/archiveDirectory.ts
@@ -76,24 +76,19 @@ async function zipDirectory(
 
   let files: fsAsync.ReaddirRecursiveFile[];
   try {
-    files = await fsAsync.readdirRecursive({ path: sourceDirectory, ignore: options.ignore });
+    files = await fsAsync.readdirRecursive({
+      path: sourceDirectory,
+      ignore: options.ignore,
+      ignoreSymlinks: true,
+    });
   } catch (err: any) {
     if (err.code === "ENOENT") {
       throw new FirebaseError(`Could not read directory "${sourceDirectory}"`, { original: err });
     }
     throw err;
   }
-  // For security, filter out all symlinks. This code is a bit obtuse to preserve ordering.
-  const realFiles = (
-    await Promise.all(
-      files.map(async (f) => {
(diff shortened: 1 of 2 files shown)


src/archiveDirectory.ts
```diff
@@ -141,7 +83,17 @@ async function zipDirectory(
     }
     throw err;
   }
-  for (const file of files) {
+  // For security, filter out all symlinks. This code is a bit obtuse to preserve ordering.
```
yuchenshi: BLOCKING

1. Line 79 above should pass in `readdirRecursive: true`
2. Is this code block still needed once Line 79 is modified?

src/fsAsync.ts
```diff
@@ -68,6 +71,13 @@ export async function readdirRecursive(
     .createFilter();
 
   const filter = (t: string): boolean => {
+    if (options.ignoreSymlinks) {
+      const stats = lstatSync(t);
```
yuchenshi: Performance nit: `readdirSync` actually takes `{withFileTypes: true}`. Doing so above makes it return enough info to tell if an entry is a symlink. This can make the code cleaner and also avoid repeated `lstatSync` calls.

yuchenshi: And also, once we do that, Line 37 `statSync` isn't necessary except when the entry is a symlink and `ignoreSymlinks === false`. In that specific case, `statSync` must be used to follow the symlink and it will return info about the target.

src/archiveDirectory.ts
joehan: Derp, done.

src/fsAsync.ts
joehan: Done - tho statSync is still necessary to get the file mode (Dirent doesn't expose it)

yuchenshi:
@joehan btw you may have forgotten to push your latest changes

joehan:
> @joehan btw you may have forgotten to push your latest changes

I most certainly did.

PR #9287, opened 2025-10-08
Update functions template to meet modern dev practices.

taeold:
- Removed the ESLint-on-init prompt; JavaScript and TypeScript setups now always scaffold a Biome config and skip adding lint steps for JS predeploys
- Refreshed both JS/TS function templates to use ESM syntax
- Consolidated template packages: renamed the “nolint” variants to the primary package.json, added type: "module"
- Bumped firebase- admin/firebase-functions to current versions.
-  Modernized the TypeScript tsconfig.json (ES2022 target, ESNext modules, bundler resolution, skip lib checks)

### DIFF
--- a/src/init/features/functions.spec.ts
+++ b/src/init/features/functions.spec.ts
@@ -196,7 +196,7 @@ describe("functions", () => {
               "firebase-debug.*.log",
               "*.local",
             ],
-            predeploy: ['npm --prefix "$RESOURCE_DIR" run lint'],
+            predeploy: [],
           },
         ]);
         expect(askWriteProjectFileStub.getCalls().map((call) => call.args[0])).to.deep.equal([
--- a/templates/init/functions/javascript/index.js
+++ b/templates/init/functions/javascript/index.js
@@ -1,6 +1,6 @@
 import { setGlobalOptions } from "firebase-functions";
 import * as logger from "firebase-functions/logger";
-import { onRequest } from "firebase-functions/v2/https";
+import { onRequest } from "firebase-functions/https";
 
 // Start writing functions:
 //   https://firebase.google.com/docs/functions
--- a/templates/init/functions/typescript/index.ts
+++ b/templates/init/functions/typescript/index.ts
@@ -1,6 +1,6 @@
 import { setGlobalOptions } from "firebase-functions";
 import * as logger from "firebase-functions/logger";
-import { onRequest } from "firebase-functions/v2/https";
+import { onRequest } from "firebase-functions/https";
 
 // Start writing functions:
 //   https://firebase.google.com/docs/functions/typescript


templates/init/functions/typescript/package.json
```diff
@@ -1,25 +1,29 @@
 {
   "name": "functions",
+  "type": "module",
```
jhuleatt, line 3: 🎉

templates/init/functions/typescript/tsconfig.json
```diff
@@ -1,16 +1,17 @@
...
-    "module": "NodeNext",
+    "target": "ES2022",
+    "module": "ESNext",
+    "moduleResolution": "Bundler",
+    "verbatimModuleSyntax": true,
     "esModuleInterop": true,
-    "moduleResolution": "nodenext",
     "noImplicitReturns": true,
     "noUnusedLocals": true,
     "outDir": "lib",
```
jhuleatt, line 10: since we're relying on type stripping, this can be replaced with `"noEmit": true`

I think that means `target` can be removed too

templates/init/functions/typescript/package.json
```diff
@@ -1,25 +1,29 @@
...
+  "private": true,
+  "engines": {
+    "node": "{{RUNTIME}}"
+  },
+  "main": "lib/index.js",
```
jhuleatt, line 8: want to rely on type stripping and just point to index.ts?

jhuleatt: ohh, maybe we can't since this still allows Node 18 and 20

jhuleatt: wait, it always defaults to the latest supported Node version, so we can!

taeold: I think it might be a little safe to stick with js though - type stripping is available on "recent" version of nodejs22 and not all nodejs22

src/init/features/functions/typescript.ts
```diff
@@ -1,58 +1,28 @@
...
-      PACKAGE_NO_LINTING_TEMPLATE.replace(
-        "{{RUNTIME}}",
-        supported.latest("nodejs").replace("nodejs", ""),
-      ),
-    );
-  }
+  cbconfig.predeploy = [
+    'npm --prefix "$RESOURCE_DIR" run lint',
+    'npm --prefix "$RESOURCE_DIR" run build',
+  ];
```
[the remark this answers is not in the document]
taeold: updated the spec.

templates/init/functions/typescript/tsconfig.json
taeold: discussed offline - let's try to not depend on type stripping just yet.

templates/init/functions/javascript/package.json
```diff
@@ -1,23 +1,26 @@
...
   },
-  "engines": {
-    "node": "{{RUNTIME}}"
-  },
-  "main": "index.js",
   "dependencies": {
-    "firebase-admin": "^12.6.0",
-    "firebase-functions": "^6.0.1"
+    "firebase-admin": "^13.5.0",
+    "firebase-functions": "^6.4.0"
```
jhuleatt, line 20: this can be updated to `^7.0.0` now 🎉

PR #9299, opened 2025-10-10
[VS Code] Add more metrics

fredzqm:
Removing unused metric names and added a few more.

Besides those, small tweaks:
- When GCA window is open, mark GiF ToS as accepted give the clear message below that button.
- Removed the "Starting emulators..." info message. User just clicked the button, they should expect this already.


firebase-vscode/src/data-connect/execution/execution.ts
```diff
@@ -130,6 +130,13 @@ export function registerExecution(
     { document, documentPath, position }: OperationLocation,
     instance: InstanceType,
   ) {
+    analyticsLogger.logger.logUsage(
```
fredzqm, line 133: Moved it here, so re-run from histories are also tracked.

PR #9301, opened 2025-10-10
[VS Code] Edit missing variables / auth

fredzqm:
### Description

- Improved the missing variable auto-fix logics.
   - Removed the modal. Execute the operation as it is, but always apply the suggested fixed in the variables panel.
   - Fixed prior bugs with default value population.
- Display variables and auth in the result panel along with result data.
- Improved the re-run a previous query CUJ.
   - When users select an item in history panel, we update both params and result panel to show its value, so developers can easily re-execute the 3rd last query.
   - Removed the concept of lastExecutedOperation.
- Renamed "configuration" panel to "params".

### Scenarios Tested

https://screencast.googleplex.com/cast/NDYyMjE3NjYwMTc2NzkzNnw2YmY1OTg2YS1lNg

### DIFF
--- a/src/command.ts
+++ b/src/command.ts
@@ -224,12 +224,15 @@ export class Command {
             });
           }
           const duration = Math.floor((process.uptime() - start) * 1000);
-          const trackSuccess = trackGA4("command_execution", {
-            command_name: this.name,
-            result: "success",
+          const trackSuccess = trackGA4(
+            "command_execution",
+            {
+              command_name: this.name,
+              result: "success",
+              interactive: getInheritedOption(options, "nonInteractive") ? "false" : "true",
+            },
             duration,
-            interactive: getInheritedOption(options, "nonInteractive") ? "false" : "true",
-          });
+          );
           if (!isEmulator) {
             await withTimeout(5000, trackSuccess);
           } else {
--- a/src/dataconnect/provisionCloudSql.ts
+++ b/src/dataconnect/provisionCloudSql.ts
@@ -24,7 +24,7 @@ export async function setupCloudSql(args: {
   instanceId: string;
   databaseId: string;
   requireGoogleMlIntegration: boolean;
-  source: "init" | "mcp_init" | "deploy";
+  source: "mcp_init" | "init" | "init_sdk" | "deploy";
   dryRun?: boolean;
 }): Promise<void> {
   const { projectId, instanceId, requireGoogleMlIntegration, dryRun } = args;
@@ -37,7 +37,7 @@ export async function setupCloudSql(args: {
     success = true;
   } finally {
     if (!dryRun) {
-      await trackGA4(
+      void trackGA4(
         "dataconnect_cloud_sql",
         {
           source: args.source,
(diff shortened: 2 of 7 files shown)


firebase-vscode/webviews/data-connect/DataConnectExecutionResultsApp.tsx
```diff
@@ -85,11 +128,7 @@ function InternalErrorView({ error }: { error: SerializedError }) {
     <>
       <Label>Error</Label>
       <p>
-        {
-          // Stacktraces usually already include the message, so we only
-          // display the message if there is no stacktrace.
-          error.stack ? <StackView stack={error.stack} /> : error.message
-        }
+        {error.message}
```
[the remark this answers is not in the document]
fredzqm: This is intended to remove clustered error details.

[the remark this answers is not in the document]
fredzqm: ack

fredzqm:
Hi, @hlshen 

Sorry about the massive PR.

I went though the reworked the VS Code execution param and result penel with some incremental fixes (started from the bug missing variable logic).

Not in a rush, but can use your expertise to review it~ Thank you!

firebase-vscode/src/data-connect/execution/execution.ts
```diff
@@ -249,92 +227,52 @@ export function registerExecution(
...
-    } catch (error) {
-      updateAndSelect({
-        state: ExecutionState.ERRORED,
-        results:
-          error instanceof Error
-            ? error
-            : new DataConnectError("Unknown error", error),
-      });
+    if (item.state === ExecutionState.ERRORED) {
+      await paramsService.applyDetectedFixes(ast);
```
hlshen, line 274: Two biggest concerns with the auto generated fields + auto rerun:

1. This is also applying to production. Can we add a confirmation for production mutations specifically?

2. Users will encounter an error on a mutation w/ repeated ID the second time around. Not a huge concern, but users may now expect it to auto-magically work every time.

firebase-vscode/src/data-connect/execution/execution-params.ts
```diff
@@ -0,0 +1,118 @@
...
+          throw new Error(
+            "Unable to parse auth claims as JSON. Check the authentication panel.\n" + e.message,
+          );
+        }
+      default:
+        throw new Error(`Unknown auth params kind: ${auth}`);
+    }
+  }
+
+  async applyDetectedFixes(ast: OperationDefinitionNode): Promise<void> {
```
hlshen, line 81: It'd be great to unit test this file/function specifically, ensuring it doesn't overwrite  user set varisbles.

firebase-vscode/src/data-connect/execution/execution.ts
fredzqm: 1) The confirmation for prod mutation is still there. It's prompted before they got here.

We execute with the prior variables, which will almost certainly return the "$varName is missing" error, then they will see the variables panel with default values populated, ready to update.

2) "repeated ID', are you talking about the AlreadyExists error for double insert? Yeah, this is an intended behavior of SQL database. Consistent in local & prod. Don't think there is much we can do.

firebase-vscode/src/data-connect/execution/execution-params.ts
fredzqm: Great suggestion.

With gemini CLI's help, just added a unit test for it~

hlshen (approved):

PR #9316, opened 2025-10-14
feat: Add isAvailable function for dynamic tool availability

joehan:
### Description
AI generated draft of a better tool discovery system - not sure we want to do this exactly, but it might be a helpful starting point.

### DIFF
--- a/firebase-vscode/src/core/config.ts
+++ b/firebase-vscode/src/core/config.ts
@@ -68,7 +68,7 @@ export async function updateFirebaseRCProject(values: {
   };
 }) {
   let didChange = false;
-  const newRCPath = path.join(currentOptions.value.cwd, ".firebaserc");
+  const newRCPath = path.join(currentOptions.value.cwd!, ".firebaserc");
   const isNewRC = !firebaseRC.value?.tryReadValue;
 
   const rc = firebaseRC.value?.tryReadValue ?? new RC(newRCPath, {});
--- a/firebase-vscode/src/options.ts
+++ b/firebase-vscode/src/options.ts
@@ -27,7 +27,6 @@ const defaultOptions: Readonly<VsCodeOptions> = {
   projectRoot: "",
   account: "",
   nonInteractive: true,
-  interactive: false,
   debug: false,
   rc: new RC(),
   exportOnExit: false,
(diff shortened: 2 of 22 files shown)


src/mcp/prompts/index.ts
```diff
@@ -40,27 +40,43 @@ function namespacePrompts(
...
     activeFeatures = Object.keys(prompts) as ServerFeature[];
   }
   if (!activeFeatures.includes("core")) {
-    activeFeatures = ["core", ...activeFeatures];
+    activeFeatures.unshift("core");
   }
-  for (const feature of activeFeatures) {
-    allPrompts.push(...namespacePrompts(prompts[feature], feature));
+  for (const key of activeFeatures) {
+    promptDefs.push(...promptDefs[key]);
```
joehan: This feels like a typo - promptDefs is an empty array going into this, so this is gonna ...undefined and throw every time. I think its meant to be:

```suggestion
    promptDefs.push(...prompts[key]);
```

joehan: In fact, could we simplify this whole function into something like:

```
return Object.entries(prompts).filter((k, v) => !activeFeatures.length || activeFeatures.includes(k)).map((k, v) => v);
```

src/mcp/prompt.ts
```diff
@@ -16,11 +17,21 @@ export interface ServerPrompt {
...
 
-export function prompt(options: ServerPrompt["mcp"], fn: ServerPrompt["fn"]): ServerPrompt {
+export function prompt(
+  feature: ServerFeature,
+  options: ServerPrompt["mcp"],
+  fn: ServerPrompt["fn"],
+  isAvailable?: (ctx: McpContext) => Promise<boolean>,
+): ServerPrompt {
+  //default to the feature level availability check, but allow override
+  const isAvailableFunc = isAvailable || getDefaultFeatureAvailabilityCheck(feature);
```
joehan, line 30: Should we switch some of the Firestore/Data Connect/etc tools that just call checkFeatureActive to just use a feature level check instead?

schnecle: I actually changed all of the tools that don't have explicit overrides to use the default. Gemini backed some of the changes out on my behalf because it didn't reload the files after I made changes 🤦 This should be good to go now.

src/mcp/prompts/index.ts
schnecle: I'm going to leave this as is for now because we are doing the thing where we add "core" back in if it's not included which mucks with the single line nature of this. We have to first see if we have any active features before we can add it back in and then do the filtering bit. Let me noodle on it a bit more though because it really feels like we should be able to simplify.

maxl0rd (approved):
LGTM

PR #9325, opened 2025-10-15
feat(dataconnect): `dataconnect:sdk:generate` automatically run `firebase init dataconnect` if no gen SDK is setup

fredzqm:
Improve `dataconnect:sdk:generate`'s behavior when the project folder hasn't been setup

- If the Firebase project hasn't been setup or dataconnect feature hasn't been setup, run the same flow as `firebase init dataconnect` (in src/init/features/dataconnect/index.ts)
- If the dataconnect feature is setup, but generated SDK hasn't been setup, run the same flow as `firebase init dataconnect:sdk` (in src/init/features/dataconnect/sdk.ts)

Then continue with the remaining steps~

We can position `firebase dataconnect:sdk:generate --watch` as that single command to setup up everything you need to use data connect in an app.

### In an empty folder

Screenshot 2025-10-15 at 3 11 07 PM

Screenshot 2025-10-15 at 3 09 59 PM

### In a folder with firebase.json, but not dataconnect

Screenshot 2025-10-15 at 3 13 16 PM

### In a folder with firebase.json & dataconnect, but no gen SDK setup
Screenshot 2025-10-15 at 3 14 21 PM

Screenshot 2025-10-15 at 3 15 03 PM

### In a folder with gen SDK setup

Screenshot 2025-10-15 at 3 11 56 PM

### DIFF
--- a/src/init/features/dataconnect/sdk.ts
+++ b/src/init/features/dataconnect/sdk.ts
@@ -105,7 +105,7 @@ export async function chooseApp(): Promise<App[]> {
       `Detected existing apps ${apps.map((a) => appDescription(a)).join(", ")}`,
     );
   } else {
-    logLabeledWarning("dataconnect", "Cannot detect any existing apps in the current directory.");
+    logLabeledWarning("dataconnect", "Cannot detect an existing app in the current directory.");
   }
   // Check for environment variables override.
   const envAppFolder = envOverride(FDC_APP_FOLDER, "");


src/commands/dataconnect-sdk-generate.ts
```diff
@@ -21,42 +26,98 @@ export const command = new Command("dataconnect:sdk:generate")
...
     }
-    if (options.watch) {
-      await Promise.race(serviceInfosWithSDKs.map(generateSDK));
-    } else {
-      for (const s of serviceInfosWithSDKs) {
-        await generateSDK(s);
+
+    let serviceInfosWithSDKs = await loadAllWithSDKs(projectId, config);
+    if (!serviceInfosWithSDKs.length) {
+      if (justRanInit || options.nonInteractive) {
```
yuchenshi, line 59: I think we should continue to ask SDK questions even if `justRanInit`. The command being run is `firebase dataconnect:sdk:generate` -- which means the user is clearly interested in SDKs. Let's give them SDKs.

fredzqm: The only way they can get here is:

`firebase init dataconnect` already asked whether they want to an app template, and they said no. Maybe they want to create their own app, but right now we don't support generating SDKs without a linked app.

This is the second screenshot above. If we run `init sdk` again, it would just fails with a different error.

Here is a behavior w/o this check:
Screenshot 2025-10-16 at 2 56 32 PM

fredzqm:
Improved the app detection logging message based on feedback

yuchenshi (approved):

PR #9333, opened 2025-10-16
Add eval runner for Gemini CLI + Firebase CLI

samedson:
### Description
Allows firebase-tools developers to create test cases against the Gemini CLI.

### Sample Commands

```
cd scripts/agent-evals
npm run test
```

### DIFF
--- a/src/appUtils.ts
+++ b/src/appUtils.ts
@@ -296,7 +296,13 @@ export function extractAppIdentifiersAndroid(fileContent: string): AppIdentifier
   return identifiers;
 }
 
-async function detectFiles(dirPath: string, filePattern: string): Promise<string[]> {
+/**
+ * Detects files matching a pattern within a directory, ignoring common dependency and build folders.
+ * @param dirPath The directory to search in.
+ * @param filePattern The glob pattern for the files to detect (e.g., "*.json").
+ * @return A promise that resolves to an array of file paths relative to `dirPath`.
+ */
+export async function detectFiles(dirPath: string, filePattern: string): Promise<string[]> {
   const options = {
     cwd: dirPath,
     ignore: [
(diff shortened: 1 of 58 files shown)


scripts/agent-evals/runner/setup.ts
```diff
@@ -0,0 +1,24 @@
+import { exec } from "child_process";
+import { promisify } from "util";
+
+const execPromise = promisify(exec);
+
+export async function buildFirebaseCli() {
+  console.log(`Building the Firebase CLI...`);
+  await execPromise("cd ../../ && npm run build");
```
joehan: We've got a script cleanInstall that would be preferable to use for this test - it builds and packages the CLI in the exact way that we would before publishing a version. There's _usually_ no difference, but we've caught a few bugs and broken builds by using it (ie npm run build will work if you directly import a transitive dep, which will break when published)

scripts/agent-evals/runner/setup.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export async function clearUserMcpServers() {
```
joehan, line 19: Would love if there was a cleaner way to do this - right now, this test isn't hermetic since it blows away your local environment if you run it locally.

Not really blocking tho.

scripts/agent-evals/runner/setup.ts
samedson: Awesome - updated to that. I'm probably gonna add an argument to skip install if you're iterating on a feature (eg. using "npm run build:watch").

scripts/agent-evals/runner/setup.ts
samedson: Yeah let me add a bug for this. At first glance there wasn't a good way to get the CLI to ignore your ".gemini" user settings, but will take a deeper look

samedson:
Ready for another look

joehan (approved):

PR #9337, opened 2025-10-17
[FDC] Relax project requirement in sdk:generate & Fixed GA metrics label ambiguity

fredzqm:
- Pulled  `dataconnectSource` to a parent object, so caller can set it.
- Always use `await isBillingEnabled` to avoid mis-leading project_status. It's cached, so no performance concern.
- Relaxed `projectId` requirement in `dataconnect:sdk:generate`. It should be able to run init & setup SDKs without a project.


yuchenshi (commented):
PTAL at the failing tests / type checks

fredzqm:
@yuchenshi Should be fixed now~ Sorry about it

PR #9341, opened 2025-10-18
BREAKING: Rewrite functions:config:export command

taeold:
Target the new defineJsonSecret API as migration target for functions.config() usage. The new API is a simpler migration target for existing functions.config() use cases.

Example flow:

```shell
$ firebase functions:config:export
i  This command retrieves your Runtime Config values (accessed via functions.config()) and exports them as a Secret Manager secret.

i  Fetching your existing functions.config() from danielylee-90... ✔  Fetched your existing functions.config().

i  Configuration to be exported:
⚠  This may contain sensitive data. Do not share this output.

{
   <CONFIG>
}

✔ What would you like to name the new secret for your configuration? RUNTIME_CONFIG

✔  Created new secret version projects/XXX/secrets/RUNTIME_CONFIG/versions/1

i  To complete the migration, update your code:

  // Before:

  exports.myFunction = functions.https.onRequest((req, res) => {
    const apiKey = functions.config().service.key;
    // ...
  });

  // After:
  const { defineJsonSecret } = require('firebase-functions/params');

  const config = defineJsonSecret("RUNTIME_CONFIG");

  exports.myFunction = functions
    .runWith({ secrets: [config] })  // Bind secret here
    .https.onRequest((req, res) => {
      const apiKey = config.value().service.key;
      // ...
    });

i  Note: defineJsonSecret requires firebase-functions v6.6.0 or later. Update your package.json if needed. i  Then deploy your functions:
```

### DIFF
--- a/src/commands/functions-config-export.ts
+++ b/src/commands/functions-config-export.ts
@@ -31,6 +31,8 @@ const SECRET_MANAGER_PERMISSIONS = [
   "secretmanager.versions.add",
 ];
 
+const DEFAULT_SECRET_NAME = "FUNCTIONS_CONFIG_EXPORT";
+
 function maskConfigValues(obj: any): any {
   if (typeof obj === "object" && obj !== null && !Array.isArray(obj)) {
     const masked: Record<string, any> = {};
@@ -43,8 +45,8 @@ function maskConfigValues(obj: any): any {
 }
 
 export const command = new Command("functions:config:export")
-  .description("export functions.config() values as JSON and store it as a single secret in Cloud Secret Manager")
-  .option("--secret <name>", "name of the secret to create (default: RUNTIME_CONFIG)")
+  .description("export environment config as a JSON secret to store in Cloud Secret Manager")
+  .option("--secret <name>", `name of the secret to create (default: ${DEFAULT_SECRET_NAME})`)
   .withForce("use default secret name without prompting")
   .before(requireAuth)
   .before(ensureApi)
@@ -89,15 +91,14 @@ export const command = new Command("functions:config:export")
       console.log("");
     }
 
-    const defaultSecretName = "FUNCTIONS_CONFIG_EXPORT";
     let secretName = options.secret as string;
     if (!secretName) {
       if (options.force) {
-        secretName = defaultSecretName;
+        secretName = DEFAULT_SECRET_NAME;
       } else {
         secretName = await input({
           message: "What would you like to name the new secret for your configuration?",
-          default: defaultSecretName,
+          default: DEFAULT_SECRET_NAME,
           nonInteractive: options.nonInteractive,
         });
       }
@@ -189,7 +190,7 @@ export const command = new Command("functions:config:export")
       logBullet(
         clc.bold("Note: ") +
           "defineJsonSecret requires firebase-functions v6.6.0 or later. " +
-          "Update your package.json if needed.",
+          `Update to a newer version with ${clc.bold("npm i firebase-functions @latest")}${!sdkVersion ? " if needed" : ""}.`,
       );
     }
     logBullet("Then deploy your functions:\n  " + clc.bold("firebase deploy --only functions"));


src/commands/functions-config-export.ts
```diff
@@ -1,151 +1,143 @@
...
+    console.log("");
 
-    if (!confirmed) {
-      throw new FirebaseError("Command aborted!");
+    // Display config in interactive mode
+    if (!options.nonInteractive) {
+      logBullet(clc.bold("Configuration to be exported:"));
+      logWarning("This may contain sensitive data. Do not share this output.");
+      console.log("");
+      console.log(JSON.stringify(configJson, null, 2));
```
jhuleatt: you could consider an `Object.keys` here to just print the names and not the values

src/commands/functions-config-export.ts
```diff
@@ -1,151 +1,143 @@
...
-}
 
-function fromEntries<V>(itr: Iterable<[string, V]>): Record<string, V> {
-  const obj: Record<string, V> = {};
-  for (const [k, v] of itr) {
-    obj[k] = v;
-  }
-  return obj;
-}
+    const defaultSecretName = "RUNTIME_CONFIG";
```
jhuleatt: maybe `FUNCTIONS_DOT_CONFIG_EXPORT`? too long?

src/commands/functions-config-export.ts
```diff
@@ -1,151 +1,143 @@
...
 
-    let pInfos = configExport.getProjectInfos(options);
-    checkReservedAliases(pInfos);
-
+    const secretVersion = await addVersion(projectId, key, secretValue);
+    console.log("");
+
+    logSuccess(`Created new secret version ${toSecretVersionResourceName(secretVersion)}`);
+    console.log("");
+    logBullet(clc.bold("To complete the migration, update your code:"));
```
jhuleatt, line 148: wow, gnarly. could a template string let you do this in a single call instead of having to break it out line by line?

src/commands/functions-config-export.ts
```diff
@@ -1,151 +1,143 @@
...
+    console.log(clc.gray(`      // ...`));
+    console.log(clc.gray(`    });`));
+    console.log("");
     logBullet(
-      "Importing functions configs from projects [" +
-        pInfos.map(({ projectId }) => `${clc.bold(projectId)}`).join(", ") +
-        "]",
+      clc.bold("Note: ") +
+        "defineJsonSecret requires firebase-functions v6.6.0 or later. " +
+        "Update your package.json if needed.",
```
jhuleatt: how hard would it be to detect an old version and either directly edit or suggest `npm i firebase-functions@latest` (and not show this if someone is on a recent version)

src/commands/functions-config-export.ts
taeold: good idea

src/commands/functions-config-export.ts
taeold: yeah let me see if there is a non-flakey way to do this!

src/commands/functions-config-export.ts
taeold: trying out FUNCTIONS_CONFIG_EXPORT

src/commands/functions-config-export.ts
```diff
@@ -1,151 +1,198 @@
...
-  pInfos = pInfos.filter((pInfo) => !pInfo.config);
-  const testPermissions = pInfos.map((pInfo) =>
-    testIamPermissions(pInfo.projectId, REQUIRED_PERMISSIONS),
-  );
-  const results = await Promise.all(testPermissions);
-  for (const [pInfo, result] of zip(pInfos, results)) {
-    if (result.passed) {
-      // We should've been able to fetch the config but couldn't. Ask the user to try export command again.
+export const command = new Command("functions:config:export")
+  .description("export environment config as a JSON secret to store in Cloud Secret Manager")
```
jhuleatt:
```suggestion
  .description("export functions.config() values as JSON and store it as a single secret in Cloud Secret Manager")
```

src/commands/functions-config-export.ts
```diff
@@ -1,151 +1,198 @@
...
+  .option("--secret <name>", "name of the secret to create (default: RUNTIME_CONFIG)")
```
jhuleatt: use the `defaultSecretName` variable here

src/commands/functions-config-export.ts
```diff
@@ -1,151 +1,198 @@
...
-
-    for (const [filename, content] of Object.entries(filesToWrite)) {
-      await options.config.askWriteProjectFile(path.join(configDir, filename), content);
+    if (!sdkVersion || semver.lt(sdkVersion, "6.6.0")) {
+      logBullet(
+        clc.bold("Note: ") +
+          "defineJsonSecret requires firebase-functions v6.6.0 or later. " +
+          "Update your package.json if needed.",
+      );
     }
```
jhuleatt, line 195:
```suggestion
    if (!sdkVersion || semver.lt(sdkVersion, "6.6.0")) {
      logBullet(
        clc.bold("Note: ") +
          "defineJsonSecret requires firebase-functions v6.6.0 or later. " +
          `Update to a newer version with ${clc.bold("npm i firebase-functions@latest")}${!sdkVersion ? " if needed." : ""}.`,
      );
    }
```

adds a bit of complexity, but this way we can be a bit more assertive if we _know_ someone isn't on the right version

jhuleatt (approved):

src/commands/functions-config-export.ts
```diff
@@ -1,151 +1,198 @@
...
 
-async function promptForPrefix(errMsg: string): Promise<string> {
-  logWarning("The following configs keys could not be exported as environment variables:\n");
-  logWarning(errMsg);
-  return await input({
-    default: "CONFIG_",
-    message: "Enter a PREFIX to rename invalid environment variable keys:",
-  });
-}
+    const defaultSecretName = "FUNCTIONS_CONFIG_EXPORT";
```
brittanycho: optional nit: should we move all constants to the top of the file and keep them together?

src/commands/functions-config-export.ts
```diff
@@ -1,151 +1,198 @@
...
+        ") and exports them as a Secret Manager secret.",
+    );
+    console.log("");
+
+    logBullet(`Fetching your existing functions.config() from ${clc.bold(projectId)}...`);
+
+    let configJson: Record<string, unknown>;
+    try {
+      configJson = await functionsConfig.materializeAll(projectId);
+    } catch (err: unknown) {
```
brittanycho, line 70: optional nit: do you think it'd be useful to differentiate between different error statuses here like 404, 403, etc?

brittanycho (approved):

src/commands/functions-config-export.ts
taeold: that's a good call. I think we usually leave those details in the debug logs for now, but I'll think about this when we get user issues.

src/commands/functions-config-export.ts
taeold: i like it.

PR #9352, opened 2025-10-20
Update detectFeatures to use new availability check

schnecle:
Update detectFeatures to use the new availability check

This revealed a couple of race-condition-y issues: 
1. Because Crashlytics requires a project directory before it can check for whether Crashlytics installed, we now need detectProjectDir and detectActiveFeatures to be run serially
2. If we log anything before the server is initialized, it was being dropped on the floor. This now caches pending messages during start up and then flushes them in order when the server is ready.

### Scenarios Tested

* Tested Crashlytics feature detection on an Android App, iOS App, and Flutter App

### Sample Commands

Run gemini cli in a directory that has Firebase Crashlytics installed
`$ gemini`

Then run `/mcp list`
You should see Crashlytics tools and prompts present

### DIFF
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -150,7 +150,7 @@ export class FirebaseMcpServer {
       return {};
     });
 
-    this.detectProjectSetup();
+    void this.detectProjectSetup();
   }
 
   /** Wait until initialization has finished. */
@@ -514,7 +514,7 @@ export class FirebaseMcpServer {
       while (this._pendingMessages.length) {
         const message = this._pendingMessages.shift();
         if (!message) continue;
-        this.log(message.level, message.data);
+        this.server.sendLoggingMessage({ level: message.level, data: message.data });
       }
 
       void this.server.sendLoggingMessage({ level, data });


src/mcp/index.ts
```diff
@@ -499,6 +508,18 @@ export class FirebaseMcpServer {
...
+      while (this._pendingMessages.length) {
+        const message = this._pendingMessages.shift();
+        if (!message) continue;
+        this.log(message.level, message.data);
+      }
+
+      void this.server.sendLoggingMessage({ level, data });
+    } else {
+      this._pendingMessages.push({ level, data });
+    }
```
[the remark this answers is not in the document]
schnecle: I think this is just incorrect - push will append to the end of the array, and shift will pull from the beginning of the array so that messages come in the correct order.

push: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push
shift: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/shift

src/mcp/index.ts
```diff
@@ -148,8 +150,7 @@ export class FirebaseMcpServer {
       return {};
     });
 
-    this.detectProjectRoot();
-    this.detectActiveFeatures();
+    this.detectProjectSetup();
```
joehan:
```suggestion
    void this.detectProjectSetup();
```

src/mcp/util/crashlytics/availability.ts
```diff
@@ -52,14 +63,21 @@ async function iosAppUsesCrashlytics(appPath: string): Promise<boolean> {
   const podfiles = await detectFiles(appPath, "Podfile");
   for (const file of podfiles) {
     const content = await fs.readFile(path.join(appPath, file), "utf8");
-    if (content.includes("Firebase/Crashlytics")) {
+    if (content.includes("Crashlytics")) {
```
visumickey, line 66: Just a small nit pick comment: What happens if someone has commented out the dependency on Crashlytics? We would still show the tools I think - I feel that should be ok. But wanted to highlight that part.

src/mcp/index.ts
maxl0rd: Is this an async call in the constructor?

src/mcp/index.ts
```diff
@@ -148,8 +150,7 @@ export class FirebaseMcpServer {
...
   }
 
   /** Wait until initialization has finished. */
```
maxl0rd, line 156: Return type?

src/mcp/util/crashlytics/availability.ts
schnecle: I think that's preferred to not loading them, but it's definitely a possibility.

src/mcp/index.ts
```diff
@@ -499,6 +508,18 @@ export class FirebaseMcpServer {
...
     }
 
-    if (this._ready) void this.server.sendLoggingMessage({ level, data });
+    if (this._ready) {
+      // once ready, flush all pending messages before sending the next message
+      // this should only happen during startup
+      while (this._pendingMessages.length) {
+        const message = this._pendingMessages.shift();
+        if (!message) continue;
+        this.log(message.level, message.data);
```
maxl0rd: Why recurse instead of calling sendLoggingMessage here?

joehan (approved):

src/mcp/index.ts
schnecle: the robots - it shouldn't :)

src/mcp/index.ts
schnecle: yes, but so were the previous calls - this seemed like an unobtrusive way to get those two calls to be serial but not impact the rest of the logic.

PR #9354, opened 2025-10-21
Add disallowLegacyRuntimeConfig property to skip Runtime Config API during function deploys.

taeold:
Adds per-codebase firebase.json property `disallowLegacyRuntimeConfig` to prevent fetching and using legacy Runtime Config (functions.config()) during deployment. Defaults to false for backward compatibility.

Example firebase.json:
```json
    {
      "functions": {
        "source": "functions",
        "codebase": "default",
        "disallowLegacyRuntimeConfig": true // NEW
      }
    }
```

New functions codebase initialized using `firebase init function`  will have this property set to true. This shouldn't be disruptive since template code uses the v2 API which doesn't allow use of `functions.config()` API anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Add a command `firebase firestore:databases:clone` to clone a Firestore database (#9262).
 - Add JSON format support for Cloud Functions secrets with `--format json` flag and auto-detection from file extensions (#1745)
 - `firebase dataconnect:sdk:generate` will run `init dataconnect:sdk` automatically if no SDKs are configured (#9325).
 - Tighten --only filter resolution for functions deployment to prefer codebase names (#9353)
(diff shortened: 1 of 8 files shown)


src/deploy/functions/prepare.ts
```diff
@@ -486,7 +488,12 @@ export async function loadCodebases(
       "functions",
       `Loading and analyzing source code for codebase ${codebase} to determine what to deploy`,
     );
-    const discoveredBuild = await runtimeDelegate.discoverBuild(runtimeConfig, {
+
+    const codebaseRuntimeConfig = shouldUseRuntimeConfig(codebaseConfig)
```
jhuleatt, line 492: worth logging here? maybe if `true`, log a warning that runtime config is going away? tempting to log if `false`, but that would probably get too noisy and might confuse.

jhuleatt: Will functions.config() just silently fail if someone on v6 has runtime config turned off in their firebase.json but tries to use it in their functions code? the env var the firebase-functions SDK reads for functions.config will just be undefined, right?

taeold: yes it would be equivalent to functions.config() returning an empty object

PR #9375, opened 2025-10-23
feat: Add hosting to firebase_init MCP tool, and use it in the init prompt

joehan:
### Description
Adds hosting to the firebase_init MCP tool. It lets users configure the siteId, public directory, and whether to configure it as a single page app. Omitting web frameworks options for now to keep things simple.

I've also swapped the /firebase:init prompt to use the new tool. This should help with cases where the prompt would try to overwrite the entire firebase.json instead of adding hosting to it.

I started off this PR in Jules - original prompt:
```
I'd like to add hosting to the firebase_init MCP tool (located in src/mcp/tools/core/init.ts). It should let users configure the siteId, public directory, and whether to configure it as. a single page app. It should not ask about web frameworks related options.

Do not change eslint, tsconfig, package.json or npm-shrinkwrap files
```

### DIFF
--- a/src/mcp/resources/guides/init_hosting.ts
+++ b/src/mcp/resources/guides/init_hosting.ts
@@ -19,7 +19,7 @@ export const init_hosting = resource(
           text: `
 ### Configure Firebase Hosting
 Default hosting site for ${ctx.projectId}: ${defaultHostingSite || "Does not exist"}
-If there is not default hosting site configure, ask the user what the site ID should be, and suggest ${ctx.projectId} as a good choice.
+If there is not a default hosting site configured, ask the user what the site ID should be, and suggest ${ctx.projectId} as a good choice.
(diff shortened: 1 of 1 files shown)


src/mcp/resources/guides/init_hosting.ts
```diff
@@ -8,26 +9,30 @@ export const init_hosting = resource(
...
+    const defaultHostingSite = await getDefaultHostingSite(ctx);
     return {
       contents: [
         {
           uri,
           type: "text",
           text: `
 ### Configure Firebase Hosting
+Default hosting site for ${ctx.projectId}: ${defaultHostingSite || "Does not exist"}
+If there is not default hosting site configure, ask the user what the site ID should be, and suggest ${ctx.projectId} as a good choice.
```
samedson: Super duper nit: "configured"

samedson (approved):

PR #9380, opened 2025-10-24
Mock MCP tools during Agent Evals

samedson:
### Description

Add the ability to mock tools for Agent Evals

### DIFF
--- a/.github/workflows/node-test.yml
+++ b/.github/workflows/node-test.yml
@@ -125,6 +125,27 @@ jobs:
   #     - uses: codecov/codecov-action@v3
   #       if: matrix.node-version == '20'
 
+  agent_evals_build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version:
+          - "22"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: npm
+          cache-dependency-path: scripts/agent-evals/package-lock.json
+
+      - run: npm i -g npm@9.5
+      - run: npm ci
+      - run: npm install
+        working-directory: scripts/agent-evals
+      - run: npm run build
+        working-directory: scripts/agent-evals
+
   unit:
     runs-on: ubuntu-latest
     strategy:
--- a/scripts/agent-evals/package-lock.json
+++ b/scripts/agent-evals/package-lock.json
@@ -16,20 +16,394 @@
       },
       "devDependencies": {
         "@types/mocha": "^9.0.0",
-        "@types/node": "^24.9.0",
-        "rxjs": "^7.8.2"
+        "@types/node": "^24.9.0"
+      }
+    },
+    "node_modules/@esbuild/aix-ppc64": {
+      "version": "0.25.11",
+      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.25.11.tgz",
+      "integrity": "sha512-Xt1dOL13m8u0WE8iplx9Ibbm+hFAO0GsU2P34UNoDGvZYkY8ifSiy6Zuc1lYxfG7svWE2fzqCUmFp5HCn51gJg==",
+      "cpu": [
+        "ppc64"
+      ],
+      "optional": true,
+      "os": [
+        "aix"
+      ],
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "node_modules/@esbuild/android-arm": {
+      "version": "0.25.11",
+      "resolved": "https://registry.npmjs.org/@esbuild/android-arm/-/android-arm-0.25.11.tgz",
+      "integrity": "sha512-uoa7dU+Dt3HYsethkJ1k6Z9YdcHjTrSb5NUy66ZfZaSV8hEYGD5ZHbEMXnqLFlbBflLsl89Zke7CAdDJ4JI+Gg==",
+      "cpu": [
+        "arm"
+      ],
+      "optional": true,
+      "os": [
+        "android"
+      ],
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "node_modules/@esbuild/android-arm64": {
+      "version": "0.25.11",
+      "resolved": "https://registry.npmjs.org/@esbuild/android-arm64/-/android-arm64-0.25.11.tgz",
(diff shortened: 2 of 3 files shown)


scripts/agent-evals/src/mock/mocks/next-js-with-project-mock.ts
```diff
@@ -0,0 +1,66 @@
...
+  DEFAULT_FIREBASE_PROJECT_NUMBER,
+  DEFAULT_FIREBASE_USER,
+  DEFAULT_FIREBASE_WEB_APP_ID,
+  DEFAULT_FIREBASE_WEB_APP_NAME,
+  DEFAULT_FIREBASE_WEB_APP_API_KEY,
+} from "../../data/index.js";
+import { toMockContent } from "../tool-mock-utils.js";
+
+export const nextJsWithProjectMock = {
+  firebase_login: toMockContent(`Successfully logged in as ${DEFAULT_FIREBASE_USER}`),
```
samedson, line 13: Long term, I'm not a fan of this. Ideally the outputs of each tool would be a templated string that we could fill in with constants here. Wanted to get this in for now

samedson: I felt that mocking APIs would be tedious but we should just be able to use nock in mock-tools-main.ts if we want to go that route

scripts/agent-evals/src/runner/gemini-cli-runner.ts
```diff
@@ -43,9 +46,12 @@ export class GeminiCliRunner implements AgentTestRunner {
     private readonly testName: string,
     testDir: string,
     runDir: string,
+    toolMocks: ToolMockName[],
   ) {
     // Create a settings file to point the CLI to a local telemetry log
     this.telemetryPath = path.join(testDir, "telemetry.log");
+    const mockPath = path.resolve(path.join(getAgentEvalsRoot(), "lib/mock/mock-tools-main.js"));
+    const firebasePath = execSync("which firebase").toString().trim();
```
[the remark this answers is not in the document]
samedson: Not sure whether this applies since we're running this only in CI

scripts/agent-evals/src/mock/mocks/next-js-with-project-mock.ts
joehan: Makes sense - I like the idea of splitting out a 'populateContent' function in each tool that handles the templating of API responses, and then just using those here.

scripts/agent-evals/src/mock/mock-tools-main.ts
```diff
@@ -0,0 +1,84 @@
...
+  return new Proxy(requiredModule, {
+    get(target, prop, receiver) {
+      if (prop !== "availableTools") {
+        return Reflect.get(target, prop, receiver);
+      }
+
+      logToFile(`Intercepting access to 'availableTools'`);
+
+      const originalAvailableTools = Reflect.get(target, prop, receiver);
+      return (ctx: any, features?: string[]): Promise<any[]> => {
```
joehan: Can ctx be typed as MCPContext or Partial<MCPContext>?

scripts/agent-evals/src/mock/mock-tools-main.ts
```diff
@@ -0,0 +1,84 @@
...
+  if (!pathRelativeToCliRoot.endsWith(MCP_TOOLS_INDEX_PATH)) {
+    return requiredModule;
+  }
+
+  logToFile(
+    `Creating proxy implementation for file: ${pathRelativeToCliRoot} with tool mocks: ${JSON.stringify(Object.keys(mocks))}`,
+  );
+
+  return new Proxy(requiredModule, {
+    get(target, prop, receiver) {
```
joehan, line 37: Can this be async get? That would let us use await inside instead of old school .then()

scripts/agent-evals/src/mock/mock-tools-main.ts
```diff
@@ -0,0 +1,84 @@
...
+          });
+
+          return finalTools;
+        });
+      };
+    },
+  });
+};
+
+const logToFile = (message: string) => {
```
joehan: Make this a function instead of a const - that will silence the 'used before defined' warnings above i think

scripts/agent-evals/src/runner/gemini-cli-runner.ts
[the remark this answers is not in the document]
joehan: Nah, i think GCA is confused here. ChildProcess.spawn would not fix anything - firebasePath would still be the same either way. 

It is correct that if you snuck a malicious firebase path onto a machine, you could get gemini to execute it when it starts up MCP servers. However, if you are able to do that, you could also just inject a malicious path for gemini to achieve the same result.

scripts/agent-evals/src/runner/paths.ts
```diff
@@ -0,0 +1,11 @@
+import path from "path";
+import { fileURLToPath } from "url";
+
+export function getAgentEvalsRoot(): string {
+  const thisFilePath = path.dirname(fileURLToPath(import.meta.url));
+  return path.resolve(path.join(thisFilePath, "..", ".."));
```
joehan, line 6: I don't love that these path.resolves() depend so tightly on exactly where files are in the directory structure (in a way that is more obscure than normal imports), but I can't think of a better approach, so 🤷

scripts/agent-evals/src/mock/mock-tools-main.ts
samedson: I'm purposefully not typing it based on types in the Firebase CLI yet just cause I didn't want to pull all that in, but it might be worthwhile... Added a task b/454924667

scripts/agent-evals/src/runner/paths.ts
samedson: I know... I realized there were a few place that needed these paths, so I put it all in one file in anticipation that this will break. Haven't thought of a better plan yet

scripts/agent-evals/src/runner/gemini-cli-runner.ts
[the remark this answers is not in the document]
samedson: See Joe's comment

scripts/agent-evals/src/mock/mock-tools-main.ts
samedson: Ah good call - I can't async this `get` method but I can `return async...` in here. Updated

PR #9381, opened 2025-10-24
Build Agent Evals for commits

samedson:
### Description

Build, but don't run Agent Evals per-commit. These run too slowly to be run that frequently.

### DIFF
--- a/.github/workflows/node-test.yml
+++ b/.github/workflows/node-test.yml
@@ -130,7 +130,6 @@ jobs:
     strategy:
       matrix:
         node-version:
-          - "20"
           - "22"
     steps:
       - uses: actions/checkout@v4


scripts/agent-evals/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "",
   "type": "module",
   "scripts": {
+    "build": "tsc",
```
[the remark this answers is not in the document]
samedson: Smart

.github/workflows/node-test.yml
```diff
@@ -125,6 +125,28 @@ jobs:
   #     - uses: codecov/codecov-action@v3
   #       if: matrix.node-version == '20'
 
+  agent_evals_build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version:
+          - "20"
```
joehan: Probably unnecessary to run this on 2 versions of node. We don't ship this code, so the cost of breaking it accidentally is low (and its pretty uncommon to break things on just one node version)

samedson: Ah was wondering what the two versions were for

PR #9393, opened 2025-10-27
Added FM Delivery Data API MCP Tool

gsiddh:
Adding MCP tooling for FCM's [Android Data Delivery API](https://firebase.google.com/docs/reference/fcmdata/rest/v1beta1/projects.androidApps.deliveryData/list#http-request). 

This will enable developers to play around with their FCM delivery data in their Agent of choice :)

### DIFF
--- a/src/mcp/tools/messaging/get_delivery_data.ts
+++ b/src/mcp/tools/messaging/get_delivery_data.ts
@@ -1,0 +1,33 @@
+import { z } from "zod";
+import { tool } from "../../tool";
+import { mcpError, toContent } from "../../util";
+import { getAndroidDeliveryData } from "../../../messaging/getDeliveryData";
+
+export const get_fcm_delivery_data = tool(
+  "messaging",
+  {
+    name: "get_fcm_delivery_data",
+    description: "Gets FCM's delivery data",
+    inputSchema: z.object({
+      appId: z.string().describe("appId to fetch data for"),
+      pageSize: z.number().optional().describe("How many results to fetch"),
+      pageToken: z.string().optional().describe("Next page token"),
+    }),
+    annotations: {
+      title: "Fetch FCM Delivery Data",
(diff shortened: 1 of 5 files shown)


src/mcp/tools/messaging/get_delivery_data.ts
```diff
@@ -0,0 +1,33 @@
+import { z } from "zod";
+import { tool } from "../../tool";
+import { mcpError, toContent } from "../../util";
+import { getAndroidDeliveryData } from "../../../messaging/getDeliveryData";
+
+export const get_fcm_delivery_data = tool(
+  "messaging",
+  {
+    name: "get_fcm_delivery_data",
+    description: "Gets FCM's delivery data",
```
fredzqm, line 10: Have you tested it with GCLI?

The MCP description here should make it clear to LLM what this tool can do, when to use it, etc.

A good description should be validated with some sample prompts.

src/mcp/tools/messaging/get_delivery_data.ts
```diff
@@ -0,0 +1,33 @@
...
+    inputSchema: z.object({
+      appId: z.string().describe("appId to fetch data for"),
+      pageSize: z.number().optional().describe("How many results to fetch"),
+      pageToken: z.string().optional().describe("Next page token"),
```
fredzqm, line 14: MCP tools aren't API.

If pageSize & pageToken can make LLM work more smooth, a few example prompts can help illustrate that.

In practice, I see it struggle with detailed API token, but works better with high-level user intent. For example, get recent delivery data in the past 7 days.

Two suggestions:
- `pageSize` should have a good default.
- Remove `pageToken` unless it improves a specific CUJ.
 
If the customer intent it to fetch a lot of delivery data, the MCP tool probably should abstract out the paging portion and return all data in one go.

PR #9423, opened 2025-10-31
Add arm64 dataconnect binary

stephenarosaj:
### Description

* Add `arm64` binary details to `downloadableEmulatorInfo.json`
* Update `downloadableEmulators.ts` to download arm64 binary when a `darwin` `arm64` environment is detected
* Add unit tests to verify correct binary is chosen for certain environments
  * Edit `EmulatorDownloadDetails` to be generated on-demand to facilitate unit testing

### Scenarios Tested

* `linux` environment --> `linux-amd64` binary
* `win32` environment --> `windows-amd64` binary
* `darwin (arm64)` environment --> `macos-arm64` binary
* `darwin (non-arm64)` environment --> `macos-amd64` binary

### DIFF
--- a/scripts/agent-evals/src/runner/gemini-cli-runner.ts
+++ b/scripts/agent-evals/src/runner/gemini-cli-runner.ts
@@ -10,11 +10,12 @@ import {
 } from "./tool-matcher.js";
 import fs from "fs";
 import { throwFailure } from "./logging.js";
-import { getAgentEvalsRoot } from "./paths.js";
+import { getAgentEvalsRoot, RunDirectories } from "./paths.js";
 import { execSync } from "node:child_process";
 import { ToolMockName } from "../mock/tool-mocks.js";
 
 const READY_PROMPT = "Type your message";
+const INSTALL_ID = "238efa5b-efb2-44bd-9dce-9b081532681c";
 
 interface ParsedTelemetryLog {
   attributes?: {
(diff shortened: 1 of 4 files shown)


src/emulator/downloadableEmulators.spec.ts
```diff
@@ -35,6 +38,8 @@ describe("downloadDetails", () => {
     process.env["FIRESTORE_EMULATOR_BINARY_PATH"] = tempEnvVars["firestore"];
     process.env["DATABASE_EMULATOR_BINARY_PATH"] = tempEnvVars["database"];
     process.env["PUBSUB_EMULATOR_BINARY_PATH"] = tempEnvVars["pubsub"];
+    Object.defineProperty(process, "platform", { value: originalProcessPlatform });
```
yuchenshi: Please use sinon to mock fields. Don't do it manually. See the other tests in the repo

stephenarosaj: Done. Updated the existing `process.env` changes to be mocked with sinon too

src/emulator/downloadableEmulators.ts
```diff
@@ -480,7 +494,7 @@ async function _runBinary(
  * @param emulator
  */
 export function getDownloadDetails(emulator: DownloadableEmulators): EmulatorDownloadDetails {
-  const details = DownloadDetails[emulator];
+  const details = generateDownloadDetails()[emulator];
```
yuchenshi: I think we may as well make `generateDownloadDetails` take `emulator` as an argument

stephenarosaj: i changed this and realized now `generateDownloadDetails` and `getDownloadDetails` have the same function signature just with a different name. Everything outside of 2 test cases uses get instead of generate, and based on implementation it seems like get is a second step that's meant to come after generate, so i switched those over to use get and made generate private.  

Done.

yuchenshi (approved):

PR #9425, opened 2025-10-31
Upgrades functions::list command to utilize Cloud Run API

brittanycho:
Upgrades functions::list command to utilize Cloud Run API:

Upgrades by adding new function (listServices) to runV2.ts and defining trigger within endpointFromService function. Then listServices and endpointFromService are newly utilized in functions-list.ts so that this command now uses cloud run api

### DIFF
--- a/src/deploy/functions/backend.spec.ts
+++ b/src/deploy/functions/backend.spec.ts
@@ -39,6 +39,41 @@ describe("Backend", () => {
     status: "ACTIVE",
   };
 
+  const RUN_SERVICE: runv2.Service = {
+    name: "projects/project/locations/region/services/id",
+    labels: {
+      "goog-managed-by": "cloud-functions",
+      "goog-cloudfunctions-runtime": "nodejs16",
+      "firebase-functions-codebase": "default",
+    },
+    annotations: {
+      "cloudfunctions.googleapis.com/function-id": "id",
+      "cloudfunctions.googleapis.com/trigger-type": "HTTP_TRIGGER",
+    },
+    template: {
+      containers: [
+        {
+          name: "worker",
+          image: "image",
+          env: [{ name: "FUNCTION_TARGET", value: "function" }],
+          resources: {
+            limits: {
+              cpu: "1",
+              memory: "256Mi",
+            },
+          },
+        },
+      ],
+      containerConcurrency: 80,
+    },
+    generation: 1,
+    createTime: "2023-01-01T00:00:00Z",
+    updateTime: "2023-01-01T00:00:00Z",
+    creator: "user",
+    lastModifier: "user",
+    etag: "etag",
+  };
+
   describe("Helper functions", () => {
     it("isEmptyBackend", () => {
       expect(backend.isEmptyBackend(backend.empty())).to.be.true;
@@ -127,6 +162,7 @@ describe("Backend", () => {
           ],
           unreachable: ["region"],
         });
+        listServices.onFirstCall().resolves([]);
         const firstBackend = await backend.existingBackend(context);
 
         const secondBackend = await backend.existingBackend(context);
@@ -158,10 +194,9 @@ describe("Backend", () => {
           functions: [],
           unreachable: [],
         });
-        listServices.onFirstCall().resolves([]);
-        listServices.throws(new FirebaseError("HTTP Error: 500, Internal Error", { status: 500 }));
 
         const context = newContext();
+        listServices.throws(new FirebaseError("HTTP Error: 500, Internal Error", { status: 500 }));
         const have = await backend.existingBackend(context);
 
         expect(have).to.deep.equal(backend.empty());
@@ -192,10 +227,30 @@ describe("Backend", () => {
           functions: [],
           unreachable: [],
         });
-        listServices.onFirstCall().resolves([]);
+        listServices.onFirstCall().resolves([RUN_SERVICE]);
         const have = await backend.existingBackend(newContext());
 
-        expect(have).to.deep.equal(backend.empty());
+        const want = backend.of({
+          ...ENDPOINT,
+          platform: "gcfv2",
+          httpsTrigger: {},
+          concurrency: 80,
+          cpu: 1,
+          availableMemoryMb: 256,
+          environmentVariables: {
+            FUNCTION_TARGET: "function",
+          },
+          secretEnvironmentVariables: [],
+          labels: {
+            "goog-managed-by": "cloud-functions",
+            "goog-cloudfunctions-runtime": "nodejs16",
+            "firebase-functions-codebase": "default",
+          },
+          codebase: projectConfig.DEFAULT_CODEBASE,
+          state: "ACTIVE",
+        });
+
+        expect(have).to.deep.equal(want);
       });
 
       it("should deduce features of scheduled functions", async () => {
@@ -277,28 +332,6 @@ describe("Backend", () => {
         expect(logLabeledWarning).to.have.been.called;
       });
 
-      it("should warn if an unused GCFv2 backend is unavailable", async () => {
-        listAllFunctions.onFirstCall().resolves({
-          functions: [],
-          unreachable: [],
-        });
-        listServices.onFirstCall().resolves([]);
-        // Simulate unreachable run/v2 regions by manually setting context
-        const context = newContext();
-
-        await backend.existingBackend(context);
-
-        // Manually set the run region as unreachable to test the warning
-        if (!context.unreachableRegions) {
-          context.unreachableRegions = { gcfV1: [], run: [] };
-        }
-        context.unreachableRegions.run = ["region"];
-
-        await backend.checkAvailability(context, backend.empty());
-
-        expect(logLabeledWarning).to.have.been.called;
-      });
-
       it("should throw if a needed GCFv1 region is unavailable", async () => {
         listAllFunctions.onFirstCall().resolves({
           functions: [],
@@ -310,53 +343,6 @@ describe("Backend", () => {
           FirebaseError,
           /The following Cloud Functions regions are currently unreachable:/,
         );
-      });
-
-      it("should throw if a GCFv2 needed region is unavailable", async () => {
-        listAllFunctions.onFirstCall().resolves({
-          functions: [],
-          unreachable: [],
-        });
-        listServices.onFirstCall().resolves([]);
-        const want: backend.Backend = backend.of({
-          ...ENDPOINT,
-          platform: "gcfv2",
-          httpsTrigger: {},
-        });
-
-        const context = newContext();
-        await backend.existingBackend(context);
-        if (!context.unreachableRegions) {
-          context.unreachableRegions = { gcfV1: [], run: [] };
-        }
-        context.unreachableRegions.run = ["region"];
-
-        await expect(backend.checkAvailability(context, want)).to.eventually.be.rejectedWith(
-          FirebaseError,
-          /The following Cloud Functions V2 regions are currently unreachable:/,
-        );
-      });
-
-      it("Should only warn when deploying GCFv1 and GCFv2 is unavailable.", async () => {
-        listAllFunctions.onFirstCall().resolves({
-          functions: [],
-          unreachable: [],
-        });
-        listServices.resolves([]);
-
-        const want = backend.of({ ...ENDPOINT, httpsTrigger: {} });
-        const context = newContext();
-        await backend.existingBackend(context);
-        if (!context.unreachableRegions) {
-          context.unreachableRegions = { gcfV1: [], run: [] };
-        }
-        context.unreachableRegions.run = ["us-central1"];
-
-        await backend.checkAvailability(context, want);
-
-        expect(listAllFunctions).to.have.been.called;
-        expect(listServices).to.have.been.called;
-        expect(logLabeledWarning).to.have.been.called;
       });
 
       it("Should only warn when deploying GCFv2 and GCFv1 is unavailable.", async () => {
@@ -372,6 +358,48 @@ describe("Backend", () => {
         expect(listAllFunctions).to.have.been.called;
         expect(listServices).to.have.been.called;
         expect(logLabeledWarning).to.have.been.called;
+      });
+
+      it("should warn if an unused Cloud Run region is unavailable", async () => {
+        listAllFunctions.onFirstCall().resolves({
+          functions: [],
+          unreachable: [],
+        });
+        listServices.resolves([]);
+        const context = newContext();
+        context.unreachableRegions = { gcfV1: [], run: ["region"] };
+        // Pre-populate the promise to prevent loadExistingBackend from overwriting unreachableRegions
+        context.existingBackendPromise = Promise.resolve(backend.empty());
+
+        await backend.checkAvailability(context, backend.empty());
+
+        expect(logLabeledWarning).to.have.been.calledWith(
+          "functions",
+          sinon.match(/The following Cloud Run regions are currently unreachable/),
+        );
+      });
+
+      it("should throw if a needed Cloud Run region is unavailable", async () => {
+        listAllFunctions.onFirstCall().resolves({
+          functions: [],
+          unreachable: [],
+        });
+        listServices.resolves([]);
+        const context = newContext();
+        context.unreachableRegions = { gcfV1: [], run: ["region"] };
+        context.existingBackendPromise = Promise.resolve(backend.empty());
+
+        const want = backend.of({
+          ...ENDPOINT,
+          platform: "gcfv2",
+          httpsTrigger: {},
+          region: "region",
+        });
+
+        await expect(backend.checkAvailability(context, want)).to.eventually.be.rejectedWith(
+          FirebaseError,
+          /The following Cloud Run regions are currently unreachable:/,
+        );
       });
     });
   });
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -551,8 +551,12 @@ async function loadExistingBackend(ctx: Context): Promise<Backend> {
       existingBackend.endpoints[endpoint.region][endpoint.id] = endpoint;
     }
   } catch (err: any) {
-    logger.debug(err.message);
-    unreachableRegions.run = ["unknown"]; // Indicate that Run services could not be listed
+    if (err.status === 404 && err.message?.toLowerCase().includes("method not found")) {
+      // customer has preview enabled without allowlist set
+    } else {
+      logger.debug(err.message);
+      unreachableRegions.run = ["unknown"];
+    }
   }
 
   ctx.existingBackend = existingBackend;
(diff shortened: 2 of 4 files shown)


src/commands/functions-list.ts
```diff
@@ -6,26 +6,57 @@
...
+      const services = await listServices(projectId);
+      v2Endpoints = services.map((service) => endpointFromService(service));
+    } catch (err: any) {
+      logger.debug(`Failed to list v2 functions:`, err);
+      logger.warn(
+        `Failed to list v2 functions. Ensure you have the Cloud Run Admin API enabled and the necessary permissions.`,
+      );
+    }
+
+    const endpointsList = [...v1Endpoints, ...v2Endpoints].sort(backend.compareFunctions);
```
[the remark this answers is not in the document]
brittanycho: done

src/commands/functions-list.ts
```diff
@@ -6,26 +6,67 @@ import { requirePermissions } from "../requirePermissions";
...
+    }
+
+    let v2Endpoints: backend.Endpoint[] = [];
+    try {
+      const services = await listServices(projectId);
+      v2Endpoints = services.map((service) => endpointFromService(service));
+    } catch (err: any) {
+      logger.debug(`Failed to list v2 functions:`, err);
+      logger.warn(
+        `Failed to list v2 functions. Ensure you have the Cloud Run Admin API enabled and the necessary permissions.`,
```
taeold: I think this entire logic should live inside `backend.existingBackend(context)`

existingBackend is the abstraction we have that calls GCF v1/v2 API and then transform it into a common representation. We should add support for Run functions there directly (which would benefit this command along with other places where we need to list functions (e.g. during deploy)

src/commands/functions-list.ts
```diff
@@ -6,26 +6,67 @@ import { requirePermissions } from "../requirePermissions";
...
       style: { head: ["yellow"] },
     });
+
     for (const endpoint of endpointsList) {
       const trigger = backend.endpointTriggerType(endpoint);
       const availableMemoryMb = endpoint.availableMemoryMb || "---";
       const entry = [
         endpoint.id,
-        endpoint.platform === "gcfv2" ? "v2" : "v1",
+        endpoint.platform === "gcfv2" ? "v2" : endpoint.platform === "run" ? "run" : "v1",
```
taeold: hmm maybe we can use a small map, like `PLATFORM_TO_DISPLAY_NAME` or something?

type PLATFORM_DISPLAY_NAME = "v1" | "v2" | "run";
const PLATFORM_TO_DISPLAY_NAME: Record<type endpoint.platform, PLATFORM_DISPLAY_NAME>  = {
  gcfv1: "v1",
  gcfv2: "v2",
  run: "run",
};

// Returns 'v2', 'run', or falls back to 'v1'
const version = platformMap[endpoint.platform]

src/gcp/runv2.spec.ts
```diff
@@ -452,4 +455,65 @@ describe("runv2", () => {
       expect(runv2.endpointFromService(service)).to.deep.equal(expectedEndpoint);
     });
   });
+
+  describe("listServices", () => {
```
taeold, line 459: looks like its missing test on whether filtering by label works correctly

src/commands/functions-list.ts
```diff
@@ -6,26 +6,67 @@ import { requirePermissions } from "../requirePermissions";
...
+    } catch (err: any) {
+      logger.debug(`Failed to list v1 functions:`, err);
+      logger.warn(
+        `Failed to list v1 functions. Ensure you have the Cloud Functions API enabled and the necessary permissions.`,
+      );
+    }
+
+    let v2Endpoints: backend.Endpoint[] = [];
+    try {
+      const services = await listServices(projectId);
```
taeold: it looks like this would list all Run services, which is not what we want.

brittanycho: updated

src/gcp/runv2.spec.ts
brittanycho: updated

src/commands/functions-list.ts
brittanycho: updated

src/commands/functions-list.ts
brittanycho: updated

src/gcp/runv2.ts
```diff
@@ -200,6 +200,38 @@ export async function updateService(service: Omit<Service, ServiceOutputFields>)
...
+
+  do {
+    const queryParams: Record<string, string> = {};
+    if (pageToken) {
+      queryParams["pageToken"] = pageToken;
+    }
+    if (filter) {
+      const parts = filter.split("=");
+      queryParams["filter"] = `labels."${parts[0]}"="${parts[1]}"`;
+    }
```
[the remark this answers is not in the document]
brittanycho: updated

brittanycho:
I see the tests are failing but seem unrelated to my current changes? will hopefully pass when re-syncing tomorrow

src/commands/functions-list.ts
```diff
@@ -1,31 +1,58 @@
...
-    for (const endpoint of endpointsList) {
+    }) as any;
+
+    for (const endpoint of endpoints) {
       const trigger = backend.endpointTriggerType(endpoint);
       const availableMemoryMb = endpoint.availableMemoryMb || "---";
       const entry = [
         endpoint.id,
-        endpoint.platform === "gcfv2" ? "v2" : "v1",
+        PLATFORM_TO_DISPLAY_NAME[endpoint.platform] || "v1",
```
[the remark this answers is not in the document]
brittanycho: will leave as is based on prior feedback

src/commands/functions-list.ts
```diff
@@ -1,31 +1,58 @@
...
-    const endpointsList = backend.allEndpoints(existing).sort(backend.compareFunctions);
+
+    let endpoints: backend.Endpoint[] = [];
+    try {
+      const existing = await backend.existingBackend(context);
+      endpoints = backend.allEndpoints(existing);
+    } catch (err: any) {
+      logger.debug(`Failed to list functions:`, err);
+      logger.warn(
+        `Failed to list functions. Ensure you have the Cloud Functions and Cloud Run APIs enabled and the necessary permissions.`,
```
taeold: think it would be helpful to include the err message here

src/deploy/functions/backend.ts
```diff
@@ -559,6 +560,21 @@ async function loadExistingBackend(ctx: Context): Promise<Backend> {
     }
   }
 
+  try {
+    const runServices = await listServices(ctx.projectId);
+    for (const service of runServices) {
+      const endpoint = endpointFromService(service);
+      existingBackend.endpoints[endpoint.region] = existingBackend.endpoints[endpoint.region] || {};
+      existingBackend.endpoints[endpoint.region][endpoint.id] = endpoint;
+    }
```
taeold, line 552: since this API call retrieves superset of v2 + run functions, i suspect that we no longer need the gcfv2 api call above? wdyt?

src/deploy/functions/backend.ts
```diff
@@ -1,5 +1,6 @@
 import * as gcf from "../../gcp/cloudfunctions";
 import * as gcfV2 from "../../gcp/cloudfunctionsv2";
+import { listServices, endpointFromService } from "../../gcp/runv2";
```
taeold: for consistency, i recommend importing this as a namespace, e.g.

```
import * as run from "../../gcp/runv2";
```

and using the namespaced call, e.g.

```
run.listServices(...)
```

CHANGELOG.md
```diff
@@ -1 +1 @@
-- [Added] support for new google-genai plugin during `init genkit` (#8957)
```
taeold: did you mean to remove other entries in the changelog?

src/gcp/runv2.ts
```diff
@@ -487,9 +522,14 @@ export function endpointFromService(service: Omit<Service, ServiceOutputFields>)
...
       id,
-
-    // TODO: trigger types.
-    httpsTrigger: {},
+    ...(service.annotations?.[TRIGGER_TYPE_ANNOTATION] === "HTTP_TRIGGER"
+      ? { httpsTrigger: {} }
+      : {
+          eventTrigger: {
+            eventType: service.annotations?.[TRIGGER_TYPE_ANNOTATION] || "unknown",
+            retry: false,
```
taeold, line 538: why is retry always false here?

src/gcp/runv2.ts
```diff
@@ -200,6 +200,41 @@ export async function updateService(service: Omit<Service, ServiceOutputFields>)
...
+      queryParams["pageToken"] = pageToken;
+    }
+
+    const res = await client.get<{ services?: Service[]; nextPageToken?: string }>(
+      `/projects/${projectId}/locations/-/services`,
+      { queryParams },
+    );
+
+    if (res.status !== 200) {
+      throw new FirebaseError(`Failed to list services: ${res.status} ${JSON.stringify(res.body)}`);
```
taeold: mind looking at if run services API return `unreachable` in the response body? usually this field is present in api calls that is multi-region (used in cases where api is temporarily unavailable in some region)

https://github.com/firebase/firebase-tools/blob/a26f4c757a9cd87e281f8e7da82d33ca2d67d6b6/src/gcp/cloudfunctionsv2.ts#L343

src/commands/functions-list.ts
brittanycho: updated

src/deploy/functions/backend.ts
brittanycho: updated

src/deploy/functions/backend.ts
brittanycho: updated

src/commands/functions-list.ts
```diff
@@ -1,31 +1,58 @@
...
-    const endpointsList = backend.allEndpoints(existing).sort(backend.compareFunctions);
+
+    let endpoints: backend.Endpoint[] = [];
+    try {
+      const existing = await backend.existingBackend(context);
+      endpoints = backend.allEndpoints(existing);
+    } catch (err: any) {
+      logger.debug(`Failed to list functions:`, err);
+      logger.warn(
+        `Failed to list functions: ${err.message}. Ensure you have the Cloud Functions and Cloud Run APIs enabled and the necessary permissions.`,
```
taeold: nit* This error message could be more helpful. For example, can it link to the links to enable the API or check the permission?

I might actually remove it all together and print out the error message for the first cut.

src/gcp/runv2.ts
```diff
@@ -200,6 +200,41 @@ export async function updateService(service: Omit<Service, ServiceOutputFields>)
   return svc;
 }
 
+export async function listServices(projectId: string): Promise<Service[]> {
```
taeold, line 209: can you add a small comment on the top to note that this API only returns services with label "cloud-functions" or "firebase-functions"? only suggesting this b/c we have other firebase products that also use Cloud Run (e.g. firebase app hosting)

src/deploy/functions/backend.ts
```diff
@@ -542,21 +542,20 @@ async function loadExistingBackend(ctx: Context): Promise<Backend> {
...
   } catch (err: any) {
-    if (err.status === 404 && err.message?.toLowerCase().includes("method not found")) {
-      // customer has preview enabled without allowlist set
-    } else {
-      throw err;
-    }
+    utils.logLabeledWarning(
+      "functions",
+      `Failed to list Cloud Run services: ${err.message}. Ensure you have the Cloud Run Admin API enabled and the necessary permissions.`,
+    );
```
taeold: this message should probably be printed by caller, though having debug message printed would be helpful.

src/deploy/functions/backend.ts
```diff
@@ -542,21 +542,20 @@ async function loadExistingBackend(ctx: Context): Promise<Backend> {
...
+    unreachableRegions.run = ["unknown"]; // Indicate that Run services could not be listed
```
taeold: this is a bit surprising. does the Run API not return list of unreacheable regions?

brittanycho: Hmm I couldn't find anything in the services.list Service object or in other Cloud Run documentation through a google search but I could definitely be missing something - I will follow-up and ask in the Cloud Run Space

CHANGELOG.md
brittanycho: No did not mean to, must've accidentally removed when I was merging with the main branch - I'll do another merge to get the changelog in sync with the latest

src/gcp/runv2.ts
brittanycho: iiuc, I don't think the Cloud Run Service object stores info on retry policies so I don't think we'd be able to know the actual retry setting for a trigger? I didn't see any related fields in https://cloud.google.com/run/docs/reference/rest/v2/projects.locations.services#resource:-service

src/gcp/runv2.ts
brittanycho: Followed up offline here: https://chat.google.com/room/AAAA8PjxsIE/mgFct9F0k9I/mgFct9F0k9I?cls=10

src/deploy/functions/backend.ts
brittanycho: Update on above here: https://chat.google.com/room/AAAA8PjxsIE/mgFct9F0k9I/mgFct9F0k9I?cls=10

Discussed further offline with Daniel and decided to move forward with this PR without unreachable regions and will add it back in later when it is fully deployed by cloud run

src/commands/functions-list.ts
brittanycho: updated

src/gcp/runv2.ts
brittanycho: updated

src/deploy/functions/backend.ts
brittanycho: updated

src/commands/functions-list.ts
```diff
@@ -1,38 +1,63 @@
...
-    const endpointsList = backend.allEndpoints(existing).sort(backend.compareFunctions);
+
+    let endpoints: backend.Endpoint[] = [];
+    try {
+      const existing = await backend.existingBackend(context);
+      endpoints = backend.allEndpoints(existing);
+    } catch (err: any) {
+      logger.debug(`Failed to list functions:`, err);
+      logger.warn(err.message);
+    }
```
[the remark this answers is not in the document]
brittanycho: Not going to update above based on: https://github.com/firebase/firebase-tools/pull/9425#discussion_r2519401649

src/gcp/runv2.spec.ts
```diff
@@ -452,4 +455,114 @@ describe("runv2", () => {
...
+      getStub.resolves({ status: 500, body: "Internal Server Error" });
+
+      try {
+        await runv2.listServices(PROJECT_ID);
+        expect.fail("Should have thrown an error");
+      } catch (err: any) {
+        expect(err).to.be.instanceOf(FirebaseError);
+        expect(err.message).to.contain('Failed to list services: 500 "Internal Server Error"');
+      }
+    });
```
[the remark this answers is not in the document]
brittanycho: n/a

src/commands/functions-list.ts
```diff
@@ -1,31 +1,56 @@
...
-    const existing = await backend.existingBackend(context);
-    const endpointsList = backend.allEndpoints(existing).sort(backend.compareFunctions);
+
+    let endpoints: backend.Endpoint[] = [];
+    try {
+      const existing = await backend.existingBackend(context);
+      endpoints = backend.allEndpoints(existing);
+    } catch (err: any) {
+      logger.debug(`Failed to list functions:`, err);
+      logger.warn(err.message);
```
taeold: hmm should we warn or err?

taeold: also, if we are not doing anything useful with the returned error, removing the try/cactch block and letting the CLI just throw the error is an option

src/deploy/functions/backend.spec.ts
```diff
@@ -126,17 +91,20 @@ describe("Backend", () => {
   describe("existing backend", () => {
     let listAllFunctions: sinon.SinonStub;
     let listAllFunctionsV2: sinon.SinonStub;
```
taeold: can we remove this stub as well?

src/deploy/functions/backend.spec.ts
```diff
@@ -126,17 +91,20 @@ describe("Backend", () => {
...
+    let listServices: sinon.SinonStub;
     let logLabeledWarning: sinon.SinonSpy;
 
     beforeEach(() => {
       listAllFunctions = sinon.stub(gcf, "listAllFunctions").rejects("Unexpected call");
       listAllFunctionsV2 = sinon.stub(gcfV2, "listAllFunctions").rejects("Unexpected v2 call");
+      listServices = sinon.stub(runv2, "listServices").resolves([]);
```
taeold: should call .reject() like other stubs

src/deploy/functions/backend.spec.ts
```diff
@@ -163,18 +131,14 @@ describe("Backend", () => {
           ],
           unreachable: ["region"],
         });
-        listAllFunctionsV2.onFirstCall().resolves({
-          functions: [],
-          unreachable: [],
-        });
```
taeold, line 169: instead of deleting this, can we have it return empty run services (goes w/ changing runStub to reject in the before() setup

src/deploy/functions/backend.ts
```diff
@@ -542,21 +543,16 @@ async function loadExistingBackend(ctx: Context): Promise<Backend> {
...
     }
-    unreachableRegions.gcfV2 = gcfV2Results.unreachable;
   } catch (err: any) {
-    if (err.status === 404 && err.message?.toLowerCase().includes("method not found")) {
-      // customer has preview enabled without allowlist set
-    } else {
-      throw err;
-    }
+    logger.debug(err.message);
+    unreachableRegions.run = ["unknown"]; // Indicate that Run services could not be listed
```
taeold: q: why do we ignore errors? e.g. what if 403 is thrown?

src/deploy/functions/backend.spec.ts
```diff
@@ -400,17 +339,22 @@ describe("Backend", () => {
...
+        listServices.onFirstCall().resolves([]);
         const want: backend.Backend = backend.of({
           ...ENDPOINT,
           platform: "gcfv2",
           httpsTrigger: {},
         });
 
-        await expect(backend.checkAvailability(newContext(), want)).to.eventually.be.rejectedWith(
+        // Manually set the run region as unreachable to simulate unavailability
```
taeold:
```suggestion
```

src/deploy/functions/backend.spec.ts
```diff
@@ -400,17 +339,22 @@ describe("Backend", () => {
...
+        const context = newContext();
+        await backend.existingBackend(context);
+        if (!context.unreachableRegions) {
+          context.unreachableRegions = { gcfV1: [], run: [] };
+        }
+        context.unreachableRegions.run = ["region"];
+
+        await expect(backend.checkAvailability(context, want)).to.eventually.be.rejectedWith(
```
taeold: hmm why does this test pass? i thought we don't call out to v2 api endpoint anymore

src/deploy/functions/backend.ts
```diff
@@ -587,7 +583,7 @@ export async function checkAvailability(context: Context, want: Backend): Promis
   const neededUnreachableV1 = context.unreachableRegions?.gcfV1.filter((region) =>
     gcfV1Regions.has(region),
   );
-  const neededUnreachableV2 = context.unreachableRegions?.gcfV2.filter((region) =>
+  const neededUnreachableV2 = context.unreachableRegions?.run.filter((region) =>
     gcfV2Regions.has(region),
```
taeold: should we rename this var or no?

CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+- [Added] support for new google-genai plugin during `init genkit` (#8957)
+- Updated to v2.17.1 of the Data Connect emulator, which fixes an admin SDK bug for operation without argument #9449 (#9454).
+- Upgraded functions::list command to use cloud run api for v2 functions (#9425)
```
taeold: remove and add our own changelist bullet?

taeold (changes requested):

src/commands/functions-list.ts
brittanycho: sounds good, removed try/catch block

src/deploy/functions/backend.spec.ts
brittanycho: updated

src/deploy/functions/backend.spec.ts
brittanycho: updated

CHANGELOG.md
brittanycho: discussed offline how above is related to merging to master branch so will remove and only update changelog at the end

src/deploy/functions/backend.ts
brittanycho: Accidentally removed above, added back in. I think previous case only handled 404 - is this ok to keep as is or should I add 403 error handling as well?

src/deploy/functions/backend.spec.ts
```diff
@@ -31,42 +31,6 @@ describe("Backend", () => {
...
-    serviceConfig: {
-      service: "service",
-      uri: GCF_URL,
-      availableCpu: "1",
-      maxInstanceRequestConcurrency: 80,
-    },
-    url: GCF_URL,
-    state: "ACTIVE",
-    updateTime: new Date(),
-  };
```
taeold, line 68: Can we replace this with equiavalent cloud run api response?

src/deploy/functions/backend.spec.ts
brittanycho: updated to remove all v2 references so now above should only refer to cloud run regions (and will also add a comment that unreachableRegions needs to be updated once Cloud Run team updates their API on this)

src/deploy/functions/backend.ts
brittanycho: updated

src/deploy/functions/backend.spec.ts
brittanycho: updated

brittanycho:
completed above in another PR so closing this one

PR #9426, opened 2025-11-03
feat(mcp): Format the text content representation of Crashlytics events

maxl0rd:
### Description

Introduces a new text content representation for Crashlytics events that is cleaner and more token efficient than the default yaml representation.

1. More scannable/understandable stack traces for both humans and models
2. Filter superfluous threads, log lines and breadcrumbs to reduce token usage

### DIFF
--- a/src/mcp/tools/crashlytics/events.ts
+++ b/src/mcp/tools/crashlytics/events.ts
@@ -44,16 +44,20 @@ function formatFrames(origFrames: Frame[], maxFrames = 20): string[] {
 }
 
 // Formats an event into more legible, token-efficient text content sections
+
 function toText(event: Event): Record<string, string> {
   const result: Record<string, string> = {};
   for (const [key, value] of Object.entries(event)) {
-    // Apply special case text formatting for some sections
     if (key === "logs") {
+      // [2025-01-01T00:00.000:00Z] Log message 1
+      // [2025-01-01T00:00.000:00Z] Log message 2
       const logs: Array<Log> = (value as Array<Log>) || [];
       const slicedLogs = logs.length > 100 ? logs.slice(logs.length - 100) : logs;
       const logLines = slicedLogs.map((log) => `[${log.logTime}] ${log.message}`);
       result["logs"] = logLines.join("\n");
     } else if (key === "breadcrumbs") {
+      // [2025-10-30T06:56:43.147Z] Event_Title1 { key1: value1, key2: value2 }                                                                               │
+      // [2025-10-30T06:56:50.328Z] Event_Title2 { key1: value1, key2: value2 }
       const breadcrumbs = (value as Breadcrumb[]) || [];
       const slicedBreadcrumbs = breadcrumbs.length > 10 ? breadcrumbs.slice(-10) : breadcrumbs;
       const breadcrumbLines = slicedBreadcrumbs.map((b) => {
@@ -65,6 +69,9 @@ function toText(event: Event): Record<string, string> {
       });
       result["breadcrumbs"] = breadcrumbLines.join("\n");
     } else if (key === "threads") {
+      // Thread: Name (crashed)                                                                                                                                                                     │
+      // at java.net.ClassName.methodName (Filename.java:123)
+      // at ...
       let threads = (value as Thread[]) || [];
       if (event.issue?.errorType === ErrorType.FATAL || event.issue?.errorType === ErrorType.ANR) {
         threads = threads.filter((t) => t.crashed || t.blamed);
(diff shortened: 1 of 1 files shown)


src/mcp/tools/crashlytics/events.ts
```diff
@@ -1,21 +1,103 @@
...
+      }
+      line += ")";
+    }
+    return line;
+  });
+  if (frames.length > 20) {
+    formatted.push("... frames omitted ...");
+  }
+  return formatted;
+}
```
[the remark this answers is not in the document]
maxl0rd: I'll allow it.

src/mcp/tools/crashlytics/events.ts
```diff
@@ -1,21 +1,103 @@
...
+        return [header, ...frameStrings].join("\n");
+      });
+      result["errors"] = errorStrings.join("\n\n");
+    } else {
+      result[key] = dump(value, DUMP_OPTIONS);
+    }
   }
-  return sample;
+  return result;
 }
```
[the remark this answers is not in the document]
maxl0rd: Nah, this idea is terrible due to the differing type signature of each formatter.

src/mcp/tools/crashlytics/events.ts
```diff
@@ -1,21 +1,103 @@
...
+      const breadcrumbs = (value as Breadcrumb[]) || [];
+      const slicedBreadcrumbs = breadcrumbs.length > 10 ? breadcrumbs.slice(-10) : breadcrumbs;
+      const breadcrumbLines = slicedBreadcrumbs.map((b) => {
+        const params =
+          b.params && Object.keys(b.params).length > 0
+            ? ` { ${Object.entries(b.params)
+                .map(([k, v]) => `${k}: ${v}`)
+                .join(", ")} }`
+            : "";
+        return `[${b.eventTime}] ${b.title}${params}`;
```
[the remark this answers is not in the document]
maxl0rd: OK

src/mcp/tools/crashlytics/events.ts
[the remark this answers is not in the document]
schnecle: While this specifically is a bad option, I don't hate the idea of having some helper methods that are specific to the section of the proto that it is transforming. It is a little tough to parse what's happening.

maxl0rd:
Added some comments which make following the code much easier

src/mcp/tools/crashlytics/events.ts
```diff
@@ -1,21 +1,117 @@
...
+      const slicedBreadcrumbs = breadcrumbs.length > 10 ? breadcrumbs.slice(-10) : breadcrumbs;
+      const breadcrumbLines = slicedBreadcrumbs.map((b) => {
+        const paramString = Object.entries(b.params)
+          .map(([k, v]) => `${k}: ${v}`)
+          .join(", ");
+        const params = paramString ? ` { ${paramString} }` : "";
+        return `[${b.eventTime}] ${b.title}${params}`;
+      });
+      result["breadcrumbs"] = breadcrumbLines.join("\n");
+    } else if (key === "threads") {
```
visumickey, line 71: Just a food for thought: Do you see a need for limiting the number of threads that are relevant to the crash? The number of threads could be so high that sometime that could be deviant. Eg of some threads that might or might not be super helpful - system framework threads, background threads (when app is foreground) and likes.

src/mcp/tools/crashlytics/events.ts
```diff
@@ -1,21 +1,117 @@
...
+  });
+  if (shouldTruncate) {
+    formatted.push("... frames omitted ...");
+  }
+  return formatted;
+}
+
+// Formats an event into more legible, token-efficient text content sections
+
+function toText(event: Event): Record<string, string> {
```
visumickey, line 48: Do we want to provide a configurable parameter to enable or disable such a formatting? By default, we can should a concise information, but when requesting full detailed information (by an agent), we can skip this data limitation.

src/mcp/tools/crashlytics/events.ts
```diff
@@ -1,21 +1,117 @@
...
-import { mcpError, toContent } from "../../util";
+import { mcpError } from "../../util";
 
-function pruneThreads(sample: Event): Event {
-  if (sample.issue?.errorType === ErrorType.FATAL || sample.issue?.errorType === ErrorType.ANR) {
-    // Remove irrelevant threads from the response to reduce token usage
-    sample.threads = sample.threads?.filter((t) => t.crashed || t.blamed);
+const DUMP_OPTIONS: DumpOptions = { lineWidth: 200 };
+
+function formatFrames(origFrames: Frame[], maxFrames = 20): string[] {
```
raymondlam, line 22: What happens if the blame frame is beyond the maxFrames?

maxl0rd: The blame frame is actually separated out into its own top-level section, so it's always accessible.

src/mcp/tools/crashlytics/events.ts
maxl0rd: Yes, this optimization is already in there. It made a massive difference in shrinking down native crashes.

src/mcp/tools/crashlytics/events.ts
maxl0rd: Let's YAGNI and see if it comes up!

src/mcp/tools/crashlytics/events.ts
raymondlam: Ahh, perfect!

PR #9440, opened 2025-11-07
Support webhooks in Data Connect behind an experiment flag.

rosalyntan:
### Description

This change adds initial webhooks support in Firebase Data Connect CLI behind an experiment flag.

Notable behavioral changes:

- If the experiment flag is enabled, `firebase init --only dataconnect` will create a dataconnect.yaml file with a `schemas` field, instead of `schema` (all changes are backwards compatible with the previous dataconnect.yaml file)
- If secondary schemas are present, the CLI will deploy them in addition to the main schema as part of `firebase deploy --only dataconnect`

Outstanding TODOs:

- Download and update dataconnect.yaml file with existing secondary schemas as part of `firebase init`
- Add a command to init secondary schemas
- Update `dataconnect:services:list` to list secondary schemas as well
- Supporting secondary schemas in MCP tooling

### DIFF
--- a/firebase-vscode/src/data-connect/config.ts
+++ b/firebase-vscode/src/data-connect/config.ts
@@ -1,6 +1,6 @@
 import { isPathInside } from "./file-utils";
 import { DeepReadOnly } from "../metaprogramming";
-import { ConnectorYaml, DataConnectYaml, mainSchemaYaml } from "../dataconnect/types";
+import { ConnectorYaml, DataConnectYaml, mainSchemaYaml } from "../../../src/dataconnect/types";
 import { Result, ResultValue } from "../result";
 import { computed, effect, signal } from "@preact/signals-core";
 import {


src/dataconnect/types.ts
```diff
@@ -211,9 +225,43 @@ export function toDatasource(
...
+        uri: ds.httpGraphql.uri,
+        timeout: ds.httpGraphql.timeout,
+      },
+    };
+  }
   return {};
 }
 
+/** Returns the main schema YAML for a Data Connect YAML */
+export function mainSchemaYaml(dataconnectYaml: DataConnectYaml): SchemaYaml {
```
fredzqm, line 240: non-blocking: Code health thoughts

Looking at how other are being use. It feels like a method on `ServiceInfo`.

Shall we consider make `ServiceInfo` a class instead of interface? We can quickly access mainSchema() &  mainSchemaYaml() from it.

`ServiceInfo` is only constructed in `load.ts`. Would it be modular if move its definition there.

src/dataconnect/client.ts
```diff
@@ -83,11 +83,12 @@ export async function deleteService(serviceName: string): Promise<types.Service>
 
 /** Schema methods */
 
-export async function getSchema(serviceName: string): Promise<types.Schema | undefined> {
+export async function getSchema(
+  serviceName: string,
+  schemaId: string,
```
fredzqm: nit: we could give it a default `= types.SCHEMA_MAIN_ID` to reduce blast radius of this change.

src/init/features/dataconnect/index.ts
```diff
@@ -521,8 +527,20 @@ async function promptForExistingServices(setup: Setup, info: RequiredInfo): Prom
...
-  const schema = await getSchema(serviceName);
-  if (!schema) {
+  let schemas: Schema[] = [];
+  try {
+    schemas = await listSchemas(serviceName, [
+      "schemas.name",
+      "schemas.datasources.postgresql.database",
+      "schemas.datasources.postgresql.cloudSql.instance",
+      "schemas.source.files",
+    ]);
```
fredzqm, line 536: Does this work?

```
[
      "schemas.name",
      "schemas.datasources",
      "schemas.source",
    ]
```

If not, I think `["*"]` would return all fields.

fredzqm (approved):

src/dataconnect/client.ts
rosalyntan: Done.

src/init/features/dataconnect/index.ts
rosalyntan: Done.

src/dataconnect/types.ts
rosalyntan: I think we may want to use `mainSchema` outside of a `ServiceInfo` context (for example, getting the main schema from the result of a `listSchemas` call).

PR #9447, opened 2025-11-10
Partially implement javascript source map upload command

andrewbrook:
### Description

This is a new, partially implemented command for uploading javascript mapping files. The command takes the user-supplied mapping file path, or a path to a directory, and uploads all available *.js.map files to the customer's Google Cloud Storage bucket. If the bucket does not exist yet, it is created.

The next steps for this command are to:
- extract the app version information
- notify Firebase Telemetry of the newly uploaded files

### Scenarios Tested

- Uploading a single mapping file
- Specifying and uploading a nested directory of mapping files
- Creating a new GCS bucket
- Changing the location of the GCS bucket

### Sample Commands

firebase crashlytics:sourcemap:upload path/to/build/mapping.js.map --app <appId>
firebase crashlytics:sourcemap:upload .next --app <appId>

### DIFF
--- a/src/commands/crashlytics-sourcemap-upload.spec.ts
+++ b/src/commands/crashlytics-sourcemap-upload.spec.ts
@@ -88,12 +88,14 @@ describe("crashlytics:sourcemap:upload", () => {
 
   it("should find and upload mapping files in a directory", async () => {
     await command.runner()(DIR_PATH, { app: "test-app" });
-    expect(gcsMock.uploadObject).to.be.calledTwice;
-    expect(gcsMock.uploadObject.firstCall.args[0].file).to.match(
-      /test-app-default-src-test-fixtures-mapping-files-mock_mapping\.js\.map\.zip/,
-    );
-    expect(gcsMock.uploadObject.secondCall.args[0].file).to.match(
-      /test-app-default-src-test-fixtures-mapping-files-subdir-subdir_mock_mapping\.js\.map\.zip/,
-    );
+    const expectedFiles = [
+      "test-app-default-src-test-fixtures-mapping-files-mock_mapping.js.map.zip",
+      "test-app-default-src-test-fixtures-mapping-files-subdir-subdir_mock_mapping.js.map.zip",
+    ];
+    const uploadedFiles = gcsMock.uploadObject
+      .getCalls()
+      .map((call) => call.args[0].file)
+      .sort();
+    expect(uploadedFiles).to.deep.equal(expectedFiles);
   });
 });


tonybaroneee (approved):

src/commands/crashlytics-sourcemap-upload.spec.ts
```diff
@@ -0,0 +1,96 @@
+import * as chai from "chai";
```
joehan, line 1: The file above this (mockdata/mock_mapping.js.map) should probably live in /src/test/fixtures instead.

I think the mockdata directory slipped past review in the past - I'll move the existing files in the over to the fixtures directory in a separate PR.

src/commands/crashlytics-sourcemap-upload.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+  it("should upload a single mapping file", async () => {
+    await command.runner()(FILE_PATH, { app: "test-app" });
+    expect(gcsMock.uploadObject).to.be.calledOnce;
+    expect(gcsMock.uploadObject).to.be.calledWith(sinon.match.any, BUCKET_NAME);
+    expect(gcsMock.uploadObject.firstCall.args[0].file).to.match(
+      /test-app-default-mockdata-mock_mapping\.js\.map\.zip/,
+    );
+  });
+
+  it("should find and upload mapping files in a directory", async () => {
```
joehan, line 89: Do we need to cover the case where a child directory has mapping files?

src/commands/crashlytics-sourcemap-upload.ts
```diff
@@ -0,0 +1,178 @@
...
+      throw new FirebaseError(
+        "provide a valid file path or directory to mapping file(s), e.g. app/build/outputs/app.js.map or app/build/outputs",
+      );
+    }
+    if (fstat.isFile()) {
+      await uploadMap(mappingFiles, bucketName, appVersion, options);
+    } else if (fstat.isDirectory()) {
+      logLabeledBullet("crashlytics", "Looking for mapping files in your directory...");
+      const files = (
+        await readdirRecursive({ path: filePath, ignore: ["node_modules", ".git"] })
```
joehan: I don't think the readdirRecursive function supports this yet, but reading files recursively with no maxDepth scares me (and has recently caused a incident for us). While you are here, could you add a maxDepth option to readdirRecursive, and set it to some reasonable value for this use case?

src/commands/crashlytics-sourcemap-upload.ts
```diff
@@ -0,0 +1,178 @@
...
+      ).filter((f) => f.name.endsWith(".js.map"));
+      for (const file of files) {
+        await uploadMap(file.name, bucketName, appVersion, options);
```
joehan: Nit: consider doing this in parrallel with promise.all:
```
await Promise.all(files.map(f => uploadMap(f.name, bucketName, appVersion, options);
```

src/commands/crashlytics-sourcemap-upload.ts
```diff
@@ -0,0 +1,178 @@
...
+      },
+      bucketName,
+    );
+    logLabeledBullet("crashlytics", `Uploaded to gs://${bucket}/${object}`);
+  } catch (e) {
+    logLabeledWarning("crashlytics", `Failed to upload mapping file ${mappingFile}:\n${e}`);
+  }
+}
+
+async function createArchive(mappingFile: string, options: CommandOptions): Promise<string> {
```
joehan: Nit: createArchive sees like a pretty generic function that could be reused by other parts of the codebase. Consider moving this to a shared file (maybe rename src/archiveDirectory.ts to src/archive.ts and add this there as archiveFile?)

PR #9455, opened 2025-11-13
fix: make runtime and buildConfig optional to prevent crash on parallel deploys

taeold:
This PR makes `runtime` and `buildConfig` optional for cloud functions.

This prevents the `TypeError: Cannot read properties of undefined (reading 'runtime')` crash that occurred during parallel deployments when the API returned a function object without `buildConfig` - which could happen during early part of function creation.


jhuleatt (approved):
giving this one a rubber stamp - are there risks to making these optional?

taeold:
@jhuleatt spent some time to make sure it doesn't impact deploys. smoke deploy worked fine. it is a bit scary change, so I'll monitor GitHub issues after release.

PR #9464, opened 2025-11-14
feat: add attemptDeadlineSeconds support to scheduled functions

taeold:
This PR adds support for `attemptDeadlineSeconds` in scheduled functions, allowing users to configure the deadline for job attempts. This mirrors the recent changes in the Functions SDK.

### Changes
- Updated `ScheduleTrigger` in `backend.ts` and `build.ts`.
- Updated `Job` interface and `jobFromEndpoint` in `cloudscheduler.ts`.
- Updated `v1alpha1.ts` to parse `attemptDeadlineSeconds` from discovery.
- Added unit tests.

### DIFF
--- a/.github/workflows/node-test.yml
+++ b/.github/workflows/node-test.yml
@@ -48,6 +48,7 @@ jobs:
         node-version:
           - "20"
           - "22"
+          - "24"
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v3
@@ -153,6 +154,7 @@ jobs:
         node-version:
           - "20"
           - "22"
+          - "24"
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v3
@@ -301,7 +303,6 @@ jobs:
       matrix:
         node-version:
           - "20"
-          - "22"
 
     steps:
       - uses: actions/checkout@v4
@@ -321,7 +322,6 @@ jobs:
       matrix:
         node-version:
           - "20"
-          - "22"
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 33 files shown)


src/gcp/cloudscheduler.spec.ts
```diff
@@ -283,5 +283,32 @@ describe("cloudscheduler", () => {
         },
       });
     });
+
+    it("should copy attemptDeadlineSeconds for v1 endpoints", async () => {
```
brittanycho: Should we add a test case for v2 as well?

src/deploy/functions/build.ts
```diff
@@ -603,6 +604,11 @@ function discoverTrigger(endpoint: Endpoint, region: string, r: Resolver): backe
     } else if (endpoint.scheduleTrigger.retryConfig === null) {
       bkSchedule.retryConfig = null;
     }
+    if (typeof endpoint.scheduleTrigger.attemptDeadlineSeconds !== "undefined") {
+      bkSchedule.attemptDeadlineSeconds = r.resolveInt(
+        endpoint.scheduleTrigger.attemptDeadlineSeconds,
+      );
+    }
```
brittanycho, line 618: nit: wondering if further validation should be added to check that attemptDeadlineSeconds doesn't exceed or not meet certain max or min values?

src/gcp/cloudscheduler.spec.ts
taeold: oh good find. actually - we should only support this in v2

src/deploy/functions/build.ts
taeold: great call. will add.

PR #9472, opened 2025-11-17
Agent Evals: Copy Firebase CLI configs in dev mode

samedson:
### Description
We don't yet have a setup where it can login automatically. For dev mode, we're going to rely on your current login.


joehan (approved):
Seems reasonable to me

PR #9482, opened 2025-11-18
updates runv2.ts with listServices function and trigger updates

brittanycho:
Adds listServices and also defines trigger within runv2.ts 

Adds listServices to runV2.ts and defines trigger within endpointFromService function. These functions will be utilized  listServices so that functions::list can utilize cloud run api

### DIFF
--- a/src/init/features/dataconnect/index.ts
+++ b/src/init/features/dataconnect/index.ts
@@ -18,7 +18,7 @@ import {
   createService,
   upsertSchema,
 } from "../../../dataconnect/client";
-import { Schema, Service, File, MAIN_SCHEMA_ID, mainSchema } from "../../../dataconnect/types";
+import { Schema, Service, File, MAIN_SCHEMA_ID, isMainSchema } from "../../../dataconnect/types";
 import { parseCloudSQLInstanceName, parseServiceName } from "../../../dataconnect/names";
 import { logger } from "../../../logger";
 import { readTemplateSync } from "../../../templates";
@@ -48,6 +48,7 @@ const DATACONNECT_YAML_TEMPLATE = readTemplateSync("init/dataconnect/dataconnect
 const DATACONNECT_WEBHOOKS_YAML_TEMPLATE = readTemplateSync(
   "init/dataconnect/dataconnect-fdcwebhooks.yaml",
 );
+const SECONDARY_SCHEMA_YAML_TEMPLATE = readTemplateSync("init/dataconnect/secondary_schema.yaml");
 const CONNECTOR_YAML_TEMPLATE = readTemplateSync("init/dataconnect/connector.yaml");
 const SCHEMA_TEMPLATE = readTemplateSync("init/dataconnect/schema.gql");
 const QUERIES_TEMPLATE = readTemplateSync("init/dataconnect/queries.gql");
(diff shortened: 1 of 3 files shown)


src/gcp/runv2.ts
```diff
@@ -200,6 +200,49 @@
...
+        ) {
+          allServices.push(service);
+        }
+      }
+    }
+    pageToken = res.body.nextPageToken;
+  } while (pageToken);
+
+  return allServices;
+}
```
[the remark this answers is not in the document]
brittanycho: afaik I don't think cloud run api supports a filter param based on the following?: https://cloud.google.com/run/docs/reference/rest/v2/projects.locations.services/list

src/gcp/runv2.spec.ts
```diff
@@ -452,4 +455,114 @@
...
+
+      const services = await runv2.listServices(PROJECT_ID);
+
+      expect(services).to.deep.equal([mockServices[0], mockServices[1]]);
+      expect(getStub).to.have.been.calledOnceWithExactly(
+        `/projects/${PROJECT_ID}/locations/-/services`,
+        { queryParams: {} },
+      );
+    });
+  });
```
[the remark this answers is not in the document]
brittanycho: same comment as above - afaik I don't think cloud run api supports a filter param based on the following?: https://cloud.google.com/run/docs/reference/rest/v2/projects.locations.services/list

src/gcp/runv2.ts
```diff
@@ -487,9 +530,14 @@ export function endpointFromService(service: Omit<Service, ServiceOutputFields>)
...
       id,
-
-    // TODO: trigger types.
-    httpsTrigger: {},
+    ...(service.annotations?.[TRIGGER_TYPE_ANNOTATION] === "HTTP_TRIGGER"
+      ? { httpsTrigger: {} }
+      : {
+          eventTrigger: {
+            eventType: service.annotations?.[TRIGGER_TYPE_ANNOTATION] || "unknown",
+            retry: false,
```
taeold, line 540: can you add a TODO comment here on figuring out how we currently set this to false b/c we don't know how to recover the info from Run (vs Functions API)

src/gcp/runv2.ts
```diff
@@ -487,9 +530,14 @@ export function endpointFromService(service: Omit<Service, ServiceOutputFields>)
       service.annotations?.[FUNCTION_TARGET_ANNOTATION] ||
       service.annotations?.[FUNCTION_ID_ANNOTATION] ||
       id,
-
-    // TODO: trigger types.
-    httpsTrigger: {},
+    ...(service.annotations?.[TRIGGER_TYPE_ANNOTATION] === "HTTP_TRIGGER"
```
taeold, line 534: note: There are actually few other triggers types that all kind of look like HTTP function:

- HTTP function
- Callable function
- Auth Blocking function
- Task Queue function
- Scheduled function

I'd be okay leaving this as todo as we figure out how to encode this information to the underlying Run service that is compatible with both V2 functions and "direct to run" functions

taeold (approved):

src/gcp/runv2.ts
brittanycho: added TODO

src/gcp/runv2.ts
brittanycho: added TODO

PR #9491, opened 2025-11-19
working progress of mcp server registry

charlotteliang:
Adding server.json so that we can get the MCP listed on the MCP registry. The registry only allows us to hard code the version of the package to point to, so I also updated our release script to bump that version as we do (credit to Gemini for the shell script).

Following https://modelcontextprotocol.info/tools/registry/publishing/ as I go, this is steps 1-3

##Scenarios Tested
I manually tested out the update-server-json-version shell script, and it works correctly.

### DIFF
--- a/.github/workflows/agent-evals.yaml
+++ b/.github/workflows/agent-evals.yaml
@@ -39,5 +39,26 @@ jobs:
       - run: npm install
         working-directory: scripts/agent-evals
       - name: "Run agent-evals tests"
-        run: npm run test
+        run: |-
+          set -euo pipefail
+
+          # Create temporary directories and clean them up when we're done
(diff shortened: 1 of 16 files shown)


src/mcp/server.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-10-17/server.schema.json",
+  "name": "io.github.firebase/firebase-mcp",
+  "description": "A MCP server that provides AI development tools the ability to work with Firebase in your app.",
```
MichaelDoyle: "A MCP server" is redundant here. Maybe something like (stolen from the CLI extension page):


```suggestion
  "description": "Gives AI development tools Firebase-specific capabilities and expertise.",
```

MichaelDoyle (approved):

PR #9492, opened 2025-11-20
Adds cloud run api call in backend.ts to fetch functions

brittanycho:
Adds cloud run api call in backend.ts to fetch functions

Added cloud run api call (from runv2.ts) in backend.ts in order to get v2 functions from Cloud Run. Gated behind an experiment until we can confirm that 100% of the needed v2 fields are covered by the Cloud Run api

### DIFF
--- a/src/experiments.ts
+++ b/src/experiments.ts
@@ -148,6 +148,11 @@ export const ALL_EXPERIMENTS = experiments({
     default: false,
     public: true,
   },
+  fdcift: {
+    shortDescription: "Enable instrumentless trial for Data Connect",
+    public: false,
+    default: false,
+  },
   apptesting: {
     shortDescription: "Adds experimental App Testing feature",
     public: true,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,0 @@
-- Add support for nodejs24 (beta) runtime for Cloud Functions for Firebase. (#9475)
-- Adds listServices and also defines trigger within runv2.ts (#9482)
-- Updates the `firebase init functions` template to use [`firebase-functions` 7.0.0](https://firebase.google.com/support/releases#cloudfunctions-700-launch) (#9496)
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,12 +1,12 @@
 {
   "name": "firebase-tools",
-  "version": "14.25.1",
+  "version": "14.26.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "firebase-tools",
-      "version": "14.25.1",
+      "version": "14.26.0",
       "license": "MIT",
       "dependencies": {
         "@apphosting/build": "^0.1.6",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "firebase-tools",
-  "version": "14.25.1",
+  "version": "14.26.0",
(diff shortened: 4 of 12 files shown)


src/deploy/functions/backend.ts
```diff
@@ -550,20 +553,39 @@ async function loadExistingBackend(ctx: Context): Promise<Backend> {
...
+      const runServices = await run.listServices(ctx.projectId);
+      for (const service of runServices) {
+        const endpoint = run.endpointFromService(service);
+        existingBackend.endpoints[endpoint.region] =
+          existingBackend.endpoints[endpoint.region] || {};
+        existingBackend.endpoints[endpoint.region][endpoint.id] = endpoint;
+      }
+    } catch (err: any) {
+      if (err.status === 404 && err.message?.toLowerCase().includes("method not found")) {
+        // customer has preview enabled without allowlist set
```
taeold: hmm what does this mean?

(this code might have been from v2 function code, and it's an edge ase that we no longer have to support)

src/deploy/functions/backend.ts
```diff
@@ -550,20 +553,39 @@ async function loadExistingBackend(ctx: Context): Promise<Backend> {
...
+        const endpoint = gcfV2.endpointFromFunction(apiFunction);
+        existingBackend.endpoints[endpoint.region] =
+          existingBackend.endpoints[endpoint.region] || {};
+        existingBackend.endpoints[endpoint.region][endpoint.id] = endpoint;
+      }
+      unreachableRegions.gcfV2 = gcfV2Results.unreachable;
+    } catch (err: any) {
+      if (err.status === 404 && err.message?.toLowerCase().includes("method not found")) {
+        // customer has preview enabled without allowlist set
+      } else {
```
taeold: we can probably clean this up too!

probably from the time when GCFv2 was in private preview (loooong time ago)

src/experiments.ts
```diff
@@ -62,6 +62,9 @@ export const ALL_EXPERIMENTS = experiments({
       "Functions created using the V2 API target Cloud Run Functions (not production ready)",
     public: false,
   },
+  functionsrunapionly: {
+    shortDescription: "Use Cloud Run API to list v2 functions",
```
taeold, line 66: set public: false

src/deploy/functions/backend.spec.ts
```diff
@@ -318,6 +361,53 @@ describe("Backend", () => {
...
+
+        const have = await backend.existingBackend(newContext());
+
+        const wantEndpoint = {
+          ...ENDPOINT,
+          platform: "gcfv2" as const,
+          concurrency: 80,
+          cpu: 1,
+          httpsTrigger: {},
+          availableMemoryMb: 256 as backend.MemoryOptions,
```
taeold: nit this was probably missing as const


```suggestion
          availableMemoryMb: 256 as const,
```

brittanycho: updated

src/experiments.ts
brittanycho: updated

src/deploy/functions/backend.ts
brittanycho: cleaned up

src/deploy/functions/backend.ts
brittanycho: Based on https://github.com/firebase/firebase-tools/pull/9492#discussion_r2548232698, cleaned this up

PR #9493, opened 2025-11-20
Implement a `firebase init dataconnect:resolver` command.

rosalyntan:
### Description

Implement a `firebase init dataconnect:resolver` command behind an experiment flag. This updates `dataconnect.yaml` and creates an empty schema.gql file in a `./schema_<SCHEMA_ID>` directory.

### Sample Commands

Screenshot 2025-12-02 at 11 17 04 AM

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,9 @@
 version: 2
 updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
   - package-ecosystem: "npm"
     directory: "/"
     schedule:


fredzqm:
Some suggestions on the prompt questions.

We can probably infer the cloud run url from secondary schema name and FDC region.

Cloud run URL isn't the easiest for someone to remember.

If they want more customization, they can always edit it manually.


[Don't wait for my review, OOO next week]

rosalyntan:
> Some suggestions on the prompt questions.
> 
> We can probably infer the cloud run url from secondary schema name and FDC region.
> 
> Cloud run URL isn't the easiest for someone to remember.
> 
> If they want more customization, they can always edit it manually.
> 
> [Don't wait for my review, OOO next week]

@fredzqm to clarify, are you suggesting skipping prompting the Cloud Run URL entirely? As currently implemented, a default Cloud Run URL value of `https://${RESOLVER_ID}-${PROJECT_NUMBER}.${LOCATION}.run.app/graphql` is already provided in the prompt (so if the developer just presses "enter" through the flow without providing a value, that's what will be set). I prefer to keep it as is, so developers who already have a function deployed can just paste the URL here, but we also provide an inferred reasonable default.

fredzqm:
Thanks for clarifying! The default params certainly help.

There are two CUJs we are looking at
A) Create a new resolver from scratch.

Folks may want to customize Cloud service name and URL path. I think we can prompt them and provide reasonable defaults.

Service name default could be "<fdc-service-name>-<fdc-schema-name>" or just the secondary name. Will let you make the call.

Path name can default to "/graphql", we can generate scaffolding code using this path.

B) Link an existing Cloud run service as a resolver.

We could list all Cloud run instances in the same region and provide them as options to save them a copy-and-paste.

We did something similar for existing Cloud SQL instances.

The part I feel strongly is we shouldn't ask developers to enter the complete Cloud run URL as it's really easy to pick a wrong URL variant. We can ask for relevant resource identifiers (run service name & path) that they can personalize.

Similarly, we include a Cloud SQL instance resource name in yaml & proto, but only prompted for the instance name.

rosalyntan:
> Thanks for clarifying! The default params certainly help.
> 
> There are two CUJs we are looking at A) Create a new resolver from scratch.
> 
> Folks may want to customize Cloud service name and URL path. I think we can prompt them and provide reasonable defaults.
> 
> Service name default could be "-" or just the secondary name. Will let you make the call.
> 
> Path name can default to "/graphql", we can generate scaffolding code using this path.
> 
> B) Link an existing Cloud run service as a resolver.
> 
> We could list all Cloud run instances in the same region and provide them as options to save them a copy-and-paste.
> 
> We did something similar for existing Cloud SQL instances.
> 
> The part I feel strongly is we shouldn't ask developers to enter the complete Cloud run URL as it's really easy to pick a wrong URL variant. We can ask for relevant resource identifiers (run service name & path) that they can personalize.
> 
> Similarly, we include a Cloud SQL instance resource name in yaml & proto, but only prompted for the instance name.

Done -- updated to remove prompting entirely for the Cloud Run URL, and added a TODO to query for existing Cloud Run instances.

src/init/features/dataconnect/resolver.ts
```diff
@@ -0,0 +1,144 @@
...
+import { Source } from ".";
+
+export interface ResolverRequiredInfo {
+  id: string;
+  uri: string;
+  serviceInfo: ServiceInfo;
+}
+
+export async function askQuestions(setup: Setup, config: Config): Promise<void> {
+  if (!experiments.isEnabled("fdcwebhooks")) {
```
fredzqm: Shall we hide this option if the experiment is off in `src/commands/init.ts`?

To be cautious, I don't remember if that list is printed out in other spot.

src/init/features/dataconnect/resolver.ts
```diff
@@ -0,0 +1,144 @@
...
+      clc.bold(`dataconnect.yaml`) +
+      " later.",
+  );
+
+  setup.featureInfo = setup.featureInfo || {};
+  setup.featureInfo.dataconnectResolver = resolverInfo;
+}
+
+export async function actuate(setup: Setup, config: Config) {
+  if (!experiments.isEnabled("fdcwebhooks")) {
```
fredzqm, line 74: super nit: I think the gate in caller site should be enough.

It's only used by init command & MCP tool.

src/init/features/dataconnect/resolver.ts
```diff
@@ -0,0 +1,144 @@
...
+    void trackGA4(
+      "dataconnect_init",
+      {
+        source,
+        project_status: setup.projectId
+          ? (await isBillingEnabled(setup))
+            ? "blaze"
+            : "spark"
+          : "missing",
+        ...{},
```
fredzqm: this line is no-op~

src/init/features/dataconnect/resolver.ts
rosalyntan: So `dataconnect:resolver` is set to `hidden:true` in `src/commands/init.ts` -- is there somewhere else this should be hidden?

src/init/features/dataconnect/resolver.ts
rosalyntan: Ack -- I figure it doesn't hurt to have this here.

src/init/features/dataconnect/resolver.ts
rosalyntan: Done.

src/init/features/dataconnect/resolver.ts
rosalyntan: Done, clarified offline.

PR #9494, opened 2025-11-20
FDC free trial

tammam-g:
### Description

Adds sql free trial creation and handling

### Scenarios Tested
Create a free trial instance on spark account
image

Try to create a second free trial instance when a free trial instance exists
image

Delete free instance with gcloud and make sure we still can't create instance
image

Free trial in constraint location
image

Local flag for enabling instrumentless billing disabled
image

Free trial created on non whitelisted Blaze plan account (current common case):
image

### Sample Commands
```
firebase experiments:enable fdcift && firebase init
```

### DIFF
--- a/src/deploy/dataconnect/prepare.spec.ts
+++ b/src/deploy/dataconnect/prepare.spec.ts
@@ -21,7 +21,6 @@ describe("dataconnect prepare", () => {
   let getResourceFiltersStub: sinon.SinonStub;
   let diffSchemaStub: sinon.SinonStub;
   let setupCloudSqlStub: sinon.SinonStub;
-  let checkBillingEnabledStub: sinon.SinonStub;
 
   beforeEach(() => {
     sandbox = sinon.createSandbox();
@@ -32,14 +31,14 @@ describe("dataconnect prepare", () => {
     getResourceFiltersStub = sandbox.stub(filters, "getResourceFilters").returns(undefined);
     diffSchemaStub = sandbox.stub(schemaMigration, "diffSchema").resolves();
     setupCloudSqlStub = sandbox.stub(provisionCloudSql, "setupCloudSql").resolves();
-    checkBillingEnabledStub = sandbox.stub(cloudbilling, "checkBillingEnabled").resolves();
     sandbox.stub(projectUtils, "needProjectId").returns("test-project");
     sandbox.stub(utils, "logLabeledBullet");
+    sandbox.stub(cloudbilling, "checkBillingEnabled").resolves();
   });
 
   afterEach(() => {
     sandbox.restore();
-    nock.cleanAll()
+    nock.cleanAll();
   });
 
   it("should do nothing if there are no services", async () => {


src/dataconnect/freeTrial.ts
```diff
@@ -28,19 +27,11 @@ export async function checkFreeTrialInstanceUsed(projectId: string): Promise<boo
...
-      "CloudSQL no cost trial has already been used on this project.",
-    );
-  } else {
-    utils.logLabeledSuccess("dataconnect", "CloudSQL no cost trial available!");
-  }
   return used;
 }
 
 export function upgradeInstructions(projectId: string): string {
-  return `To provision a CloudSQL Postgres instance on the Firebase Data Connect no-cost trial:
```
fredzqm, line 43: Probably want to keep the message around until IFT is fully launched.

src/gcp/cloudsql/cloudsqladmin.ts
```diff
@@ -70,9 +74,21 @@ export async function createInstance(args: {
   freeTrialLabel: DataConnectLabel;
 }): Promise<void> {
   const databaseFlags = [{ name: "cloudsql.iam_authentication", value: "on" }];
+  const billingEnabled = await checkBillingEnabled(args.projectId);
+  const instrumentlessTrialEnabled = envOverride("ENABLE_INSTRUMENTLESS_TRIAL", "") === "true";
```
fredzqm: You can use the CLI experiment flag.

https://github.com/firebase/firebase-tools/blob/644808cd1f8e0f2193a9f7204f84cfb190ec4f6a/src/experiments.ts#L117-L122

nit: `FDC_IFT` It should mention FDC.

src/gcp/cloudsql/cloudsqladmin.ts
```diff
@@ -70,9 +74,21 @@ export async function createInstance(args: {
...
   }
+  if (args.freeTrialLabel === "nt" && !billingEnabled) {
+    throw new FirebaseError(
+      `The Cloud SQL free trial instance has already been used for this project. ${upgradeInstructions(args.projectId)}
+      
+      Add a billing account to your project ${args.projectId} create new Cloud SQL instances.`,
+    );
+  }
+  if (args.freeTrialLabel === "ft" && !billingEnabled && !instrumentlessTrialEnabled) {
+    throw new FirebaseError(`You need an active billing account to create free trial instance`);
```
fredzqm: [blocking] Include the old upgrade message here

"Local flag for enabling instrumentless billing disabled" case above should print the same message as in master branch.

src/init/features/dataconnect/index.ts
```diff
@@ -708,21 +636,26 @@ async function promptForCloudSQL(setup: Setup, info: RequiredInfo): Promise<void
...
       } else {
-        choices.push({ name: "Create a new CloudSQL instance", value: "", location: "" });
+        choices.push({ name: "Create a new CloudSQL instance", value: PAID, location: "" });
       }
       info.cloudSqlInstanceId = await select<string>({
         message: `Which CloudSQL instance would you like to use?`,
         choices,
       });
-      if (info.cloudSqlInstanceId !== "") {
+      if (![FREE, PAID].includes(info.cloudSqlInstanceId)) {
```
fredzqm: We can use "" to represent new instance option, because only one of those options will be added.

Is this assumption changed? Will we ever ask if they want to create FREE or PAID instance?
Would suggest revert for simplification

```
      if (freeTrialAvailable) {
        choices.push({ name: "Create a new free trial instance", value: FREE, location: "" });
      } else {
        choices.push({ name: "Create a new CloudSQL instance", value: PAID, location: "" });
      }
```

fredzqm (changes requested):

src/init/features/dataconnect/index.ts
```diff
@@ -708,10 +722,10 @@ async function promptForCloudSQL(setup: Setup, info: RequiredInfo): Promise<void
     // If we've already chosen a region (ie service already exists), only list instances from that region.
     choices = choices.filter((c) => info.locationId === "" || info.locationId === c.location);
     if (choices.length) {
-      if (!(await checkFreeTrialInstanceUsed(setup.projectId))) {
-        choices.push({ name: "Create a new free trial instance", value: "", location: "" });
+      if (freeTrialAvailable) {
+        choices.push({ name: "Create a new free trial instance", value: FREE, location: "" });
       } else {
-        choices.push({ name: "Create a new CloudSQL instance", value: "", location: "" });
+        choices.push({ name: "Create a new CloudSQL instance", value: PAID, location: "" });
```
fredzqm: Don't forget to remove `FREE` and `PAID` here.

fredzqm (approved):
LGTM

src/gcp/cloudsql/cloudsqladmin.ts
tammam-g: Ok, fixed we don't do anything here but during the prompting we will skip the whole flow here if billing isn't enabled (same behaviour as before)

PR #9501, opened 2025-11-21
feat(functions): avoid blocking on missing v2 source tokens

kdawgwilk:
Add an optional timeout to SourceTokenScraper so callers can bail out when a
source token never returns. Use a 2s cap in v2 create/update paths to keep
builds running in parallel, while still reusing tokens when available. Add a
spec covering the timeout behavior.


kdawgwilk:

PR #9503, opened 2025-11-21
feat(functions): add 1.5GB and 3GB memory options for v2

kdawgwilk:
- Extend memory option types/mappings to include 1536MB and 3072MB with display labels.
- Map new sizes to sensible v2 CPU defaults (1 CPU for 1.5GB, 2 CPU for 3GB) and keep gen1 allowed set unchanged.
- Add validation guard so gen1 rejects unsupported memory values; update coverage to allow new sizes for v2 and assert the gen1 rejection.


kdawgwilk:

PR #9505, opened 2025-11-21
fix: honor functions --except for codebases and IAM checks

kdawgwilk:
- parse/propagate exclude filters alongside --only, skipping excluded codebases/endpoints
- apply include/exclude filtering in prepare, deploy planning, and HTTP IAM checks
- add unit coverage for deployment filters, codebase targeting, and excluded build skipping

Now if you do

```
firebase deploy --except functions:<codebase_a>
```

it will not try and load `codebase_a`.  However if you do

```
firebase deploy --except functions:<codebase_a>:<function_x>
```

It will still load the codebase in case there are other functions in the codebase

### DIFF
--- a/src/deploy/functions/args.ts
+++ b/src/deploy/functions/args.ts
@@ -36,6 +36,7 @@ export interface Payload {
 export interface Context {
   projectId: string;
   filters?: deployHelper.EndpointFilter[];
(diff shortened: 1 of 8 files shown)


kdawgwilk:

PR #9519, opened 2025-11-24
Lazy load commands

joehan:
### Description
Lazy load commands as needed, instead of 'require'ing them all at once at the start of every command

### Scenarios Tested
Manually validated that commands still work, and check that 'firebase' and 'firebase help' still work. Also tested out flags, both that expected ones are allowed and unexpected ones as not:
Screenshot 2025-11-25 at 3 08 44 PM

Also tested out that  nested commands  still work -  ie`firebase target` and `firebase target:apply` 

Added some timer code, and found that previously, we would do about 800ms of loading before every command - now it is  <10ms.
Before this change - every commands would take about 800ms upfront:
Screenshot 2025-11-24 at 2 25 19 PM
After - different commands take differing amounts of time, but range from >10ms (for smaller ones) to ~300ms (for the ones with the largest dep trees)
Screenshot 2025-11-24 at 2 33 59 PM

Finally, I also validated that `require('firebase-tools').getCommand("commandName")` still works. IDK if anyone in the world uses this, but we technically export it (and use it in the barely used `firebase ext` command)

### DIFF
--- a/src/bin/cli.ts
+++ b/src/bin/cli.ts
@@ -107,33 +107,33 @@ export function cli(pkg: any) {
   process.on("uncaughtException", (err) => {
     errorOut(err);
   });
-  
-    // If this is a help command, load all commands so we can display them.
-    const isHelp = !args.length || args[0] === "help" || (args.length === 1 && args[0] === "ext");
-    if (isHelp) {
-      const seen = new Set();
-      const loadAll = (obj: any) => {
-        if (seen.has(obj)) return;
-        seen.add(obj);
-        for (const [key, value] of Object.entries(obj)) {
-          if (isCommandModule(value)) {
-            value.load();
-          } else if (
-            typeof value === "object" &&
-            value !== null &&
-            !Array.isArray(value) &&
-            key !== "cli"
-          ) {
-            loadAll(value);
-          }
+
+  // If this is a help command, load all commands so we can display them.
+  const isHelp = !args.length || args[0] === "help" || (args.length === 1 && args[0] === "ext");
+  if (isHelp) {
+    const seen = new Set();
+    const loadAll = (obj: any) => {
+      if (seen.has(obj)) return;
+      seen.add(obj);
+      for (const [key, value] of Object.entries(obj)) {
+        if (isCommandModule(value)) {
+          value.load();
+        } else if (
+          typeof value === "object" &&
+          value !== null &&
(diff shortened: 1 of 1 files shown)


scripts/client-integration-tests/tests.ts
```diff
@@ -50,7 +50,7 @@ describe("deployHosting", () => {
       cwd: projectDirectory,
       only: "hosting:client-integration-site",
     });
-  }).timeout(10 * 1e3); // Deploying takes several steps.
+  }).timeout(10 * 1e4); // Deploying takes several steps.
```
joehan: This test has been flaky even before this PR, unfortunately. I bumped the timeout here so that I could be sure that this PR isn't breaking usage as a module

bkendall: Consider changing what you've changed though. 1e3 is the milliseconds units, so it's easy to see that this is 10s - now you have to move the extra zero over figure out why it's 1e4...

I also question making it take 100s - if it fails, you don't want it taking 100s, you want it to fail quickly. Maybe making it 20s would be fine, but if it's hanging and unable to finish in 10s, it's unlikely an extra 90 would help (if it's completely hung up)

src/commands/index.ts
```diff
@@ -1,20 +1,36 @@
...
+      return run(...args);
+    }) as CommandRunner;
+
+    // Store the load function on the runner so we can trigger it without running.
+    runner.load = () => {
+      const tStart = process.hrtime.bigint();
+      require(`./${name}`).command.register(client);
+      const tEnd = process.hrtime.bigint();
+      const diff = (tEnd - tStart) / BigInt(1e6);
+      if (diff > 75) {
```
bkendall: Do you need to keep this?

joehan:
> This is a pretty drastic change. Consider doing this w/ the breaking changes, just in case it does go sideways.

Yeah, seems like a good call. I've tested pretty extensively (including usage as a module), but extra caution never hurts.

scripts/client-integration-tests/tests.ts
joehan: Ahhh, i totally misread this timeout as 1 second at first.  Switched to 20s

src/commands/index.ts
joehan: Probably not, cleaned up

PR #9529, opened 2025-11-25
Fix(dataconnect): Allow non-string variables in execute

nohe427:
This PR fixes an issue where GraphQL variables of types other than string were not being correctly passed to the Data Connect backend. This was caused by a type mismatch in the `ExecuteGraphqlRequest` and `ExecuteOperationRequest` interfaces, where the `variables` field was typed as `{ [key: string]: string }` instead of `{ [key: string]: any }`.

This change updates the types to allow any value type for variables, which aligns with the actual usage in `dataconnect-execute.ts` and how the backend handles variables.


nohe427:
This was auto-created from me attempting to automate some of my workflow. Please Ignore

PR #9558, opened 2025-12-02
Improvements to the publish script

joehan:
### Description
Fixing some annoying bugs in the publish script.

- Fix a bug where 'npm view' would return an outdated version for the latest tag and cause artifacts to be created for a previous version.
- Fix a bug where preview releases would get the 'latest' tag instead of the 'canary' tag.

### DIFF
--- a/firebase-vscode/src/data-connect/execution/execution.ts
+++ b/firebase-vscode/src/data-connect/execution/execution.ts
@@ -199,14 +199,6 @@ export function registerExecution(
       }
     }
 
-    // build schema to check for compilation errors
-    // TODO: run schema check on locally modified schema
-    const introspect = await dataConnectService.introspect();
-    if (!introspect.data) {
-      executionError("Please check your compilation errors");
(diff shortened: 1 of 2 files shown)


scripts/publish.sh
```diff
@@ -133,4 +140,8 @@ if [[ $VERSION != "preview" ]]; then
   echo "Publishing release notes..."
   hub release create --file "${RELEASE_NOTES_FILE}" "v${NEW_VERSION}"
   echo "Published release notes."
+else
+  echo "Publishing canary version to npm..."
+  npx clean-publish@5.0.0 --before-script ./scripts/clean-shrinkwrap.sh -- --tag canary
```
bkendall: nit: should the tag be `canary` or `preview`. Firebase tends to be fairly on the "preview" bandwagon

joehan: We've used canary in the past, but preview seems better tbh.

PR #9566, opened 2025-12-03
Update past react 19.2.0

joehan:
### Description
Updating some integration tests past a severe CVE in react. This code doesn't ship to users, so there should be no risk there, but we want to be thorough in cleaning up any vulnerable versions.


scripts/webframeworks-deploy-tests/tests.ts
```diff
@@ -357,7 +357,7 @@ describe("webframeworks", function (this) {
       it("should have working ISR", async () => {
         const response = await fetch(`${NEXTJS_HOST}/pages/isr`);
         expect(response.ok).to.be.true;
-        expect(response.headers.get("cache-control")).to.eql("private");
+        expect(response.headers.get("cache-control")).to.include("s-maxage=10");
```
joehan, line 360: This might be a problem... seems like it actually changed a meaningful assertion here

PR #9570, opened 2025-12-03
Replace hosting with deploy prompt in init command

charlotteliang:
Tested locally that it no longer retrieves hosting guide and directed to app hosting prompt successfully. 
Also tested that it's able to deploy to public URL successfully.


src/mcp/resources/guides/init_backend.ts
```diff
@@ -19,14 +19,16 @@ export const init_backend = resource(
 1. Determine based on what you already know about the user's project or by asking them which of the following services is appropriate.
 2. Use the Firebase \`read_resources\` tool to load the guide to setup the product you choose.
 
-The user will likely need to setup Firestore, Authentication, and Hosting. Read the following guides in order:
+The user will likely need to setup Firestore, Authentication. Read the following guides in order:
  1. [Firestore](firebase://guides/init/firestore): read this to setup Firestore database
  2. [Authentication](firebase://guides/init/auth): read this to setup Firebase Authentication to support multi-user apps
  3. [Firestore Rules](firebase://guides/init/firestore_rules): read this to setup the \`firestore.rules\` file for securing your database
- 4. [Hosting](firebase://guides/init/hosting): read this if the user would like to deploy to Firebase Hosting
```
joehan, line 26: I don't think this is how we want to approach this. Instead of cutting out all references to hosting here, let's update the hosting guide to just hlep configure firebase.json, but not suggest running deploy.

charlotteliang: I thought we will use Deploy prompt instead of hosting guide moving forward, my understanding is that the Deploy command should consolidate both and will determine which product to use based on the app. @jpreid7396

PR #9588, opened 2025-12-05
[VS Code] Fix rendering of generate query code lens

fredzqm:
- Don't show "Generate query" in schema file.
  - Don't show "Generate query" for comments within an operation.
  - Show "Add Data" for @table type only.
  - Show "View Data" for @table and @view type.

Screenshot 2025-12-04 at 9 10 35 PM

### DIFF
--- a/scripts/agent-evals/templates/next-app-hello-world/package-lock.json
+++ b/scripts/agent-evals/templates/next-app-hello-world/package-lock.json
@@ -8,7 +8,7 @@
       "name": "my-app",
       "version": "0.1.0",
       "dependencies": {
-        "next": "16.0.7",
+        "next": "16.0.0",
         "react": "19.2.1",
         "react-dom": "19.2.1"
       },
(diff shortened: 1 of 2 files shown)


firebase-vscode/src/data-connect/code-lens-provider.ts
```diff
@@ -72,84 +72,93 @@ export class OperationCodeLensProvider extends ComputedCodeLensProvider {
...
-    const projectId = this.watch(firebaseRC)?.tryReadValue?.projects.default;
     if (!fdcConfigs) {
       return [];
     }
-
-    const codeLenses: vscode.CodeLens[] = [];
+    const projectId = this.watch(firebaseRC)?.tryReadValue?.projects.default;
 
     const documentText = document.getText();
-    // TODO: replace w/ online-parser to work with malformed documents
```
yuchenshi, line 83: This TODO still makes sense. Mostly so that the codelens do not disappear and appear repeatedly while you're typing. Let's leave this here even if you don't want to address it any time soon

fredzqm: Sure.

Though I did some quick search on online-parser. Didn't actually find anything that we could use easily~

Do you know any online GQL parser library by any chance?

yuchenshi: The GQL language server we already depend on has an online parser, but we haven't got the time to pipe its output to our code lens provider just yet.

PR #9594, opened 2025-12-05
Reorganizing Crashlytics MCP Tools

maxl0rd:
### Description

- Consolidating all of the `crashlytics_get_*` tools into a single tool `crashlytics_get_report`. This will enable the set of supported reports to expand without adding additional tools.
- Moving most of the content from the `crashlytics:connect` prompt to guide resources. This provides better support for the many MCP clients without support for prompts.
- Numerous other improvements in prompt engineering and error handling


src/crashlytics/filters.ts
```diff
@@ -3,13 +3,7 @@ import { FirebaseError } from "../error";
...
 export const ApplicationIdSchema = z
   .string()
-  .describe(
-    "Firebase app id. For an Android application, read the " +
-      "mobilesdk_app_id value specified in the google-services.json file for " +
-      "the current package name. For an iOS Application, read the GOOGLE_APP_ID " +
-      "from GoogleService-Info.plist. If neither is available, ask the user to " +
-      "provide the app id.",
-  );
+  .describe("Firebase App Id. Leave blank if unknown to receive instructions for finding it.");
```
schnecle: Should we instruct it to use read_resource here rather than telling it to leave it blank?

src/mcp/tools/crashlytics/reports.ts
```diff
@@ -1,191 +1,89 @@
...
   "crashlytics",
   {
-    name: "get_top_apple_devices",
-    description: `Counts events and distinct impacted users, grouped by apple *device*.
-      Groups are sorted by event count, in descending order.
-      Only counts events matching the given filters.
-      Only relevant for iOS, iPadOS and MacOS applications.`,
+    name: "get_report",
+    description:
+      `Use this to request numerical reports from Crashlytics. Reports contain aggregated metrics describing the number of crash events and number of impacted end users. The reports are grouped by different dimensions such as issue, version or device.
```
schnecle: Should we make mention of the guide in the tool description?

src/crashlytics/reports.spec.ts
schnecle, line 129: Can we add some tests around the different validation cases we expect with the error responses?

src/mcp/resources/guides/crashlytics_connect.ts
```diff
@@ -0,0 +1,49 @@
+import { resource } from "../../resource";
+
+export const RESOURCE_CONTENT = `
+### Instructions for Working with Firebase Crashlytics Tools
+
+Only ask the user one question at a time. Do not proceed without user instructions. Upon receiving user instructions, refer to the relevant resources for guidance.
```
schnecle: Is this a strong enough statement? I've seen it ignore this pretty readily without emphasis - asking multiple questions at a time being the main thing it tries to do.

src/mcp/resources/guides/crashlytics_connect.ts
```diff
@@ -0,0 +1,49 @@
...
+    
+  2. [Firebase Crashlytics Issues Guide](firebase://guides/crashlytics/issues)
+     This guide details how to work with issues within Crashlytics. The agent should read this guide before prioritizing issues or presenting issue data to the user.
+
+  3. [Investigating Crashlytics Issues](firebase://guides/crashlytics/investigations)
+     This guide provides instructions on investigating the root causes of crashes and exceptions reported in Crashlytics issues.
+
+### Check That You Are Connected
+
+Verify that you can read the app's Crashlytics data by getting the topVersions report. This report will tell you which app versions have the most events.
```
schnecle, line 24: We didn't have this step before - how has this played out in your testing? My one concern here is that it increases token consumption by default without interaction from the user. That happens not infrequently anyway, but curious if it happens more with this instruction.

src/mcp/tools/crashlytics/issues.ts
```diff
@@ -23,9 +26,21 @@ export const get_issue = tool(
...
   },
   async ({ appId, issueId }) => {
-    if (!appId) return mcpError(`Must specify 'appId' parameter.`);
-    if (!issueId) return mcpError(`Must specify 'issueId' parameter.`);
-
+    const result: CallToolResult = { content: [] };
+    if (!appId) {
+      result.isError = true;
+      result.content.push({ type: "text", text: "Must specify 'appId' parameter" });
+      result.content.push({ type: "text", text: forceAppIdGuide });
```
schnecle, line 33: We don't return the issue guide here. At some iteration, I feel like we were doing that. If we were, why did we pull it back?

src/mcp/tools/crashlytics/reports.ts
```diff
@@ -1,191 +1,89 @@
...
-  getReportContent(
-    CrashlyticsReport.TopAndroidDevices,
-    `The crashlytics_get_top_issues tool can report the top issues for the devices in this response.
-    Pass the device.displayName in the filter.deviceDisplayNames field. 
-    The crashlytics_list_events tool can retrieve a list of events for a device in this response.`,
-  ),
-);
+    if (!report) {
+      result.isError = true;
+      result.content.push({ type: "text", text: `Error: ${REPORT_ERROR_CONTENT}` });
```
schnecle, line 59: Same question here - why not reference reading the guide in the error response?

src/crashlytics/filters.ts
maxl0rd: Decided that the lib isn't the place for mcp instructions to live...

src/crashlytics/reports.spec.ts
maxl0rd: Evaluating that is a good idea. But it doesn't make sense to test the tool behavior in this file's unit test.

src/mcp/resources/guides/crashlytics_connect.ts
maxl0rd: I actually think we might want to remove this entirely and let the agent and developer determine the working agreement. What if the agent is meant to be autonomous?

src/mcp/tools/crashlytics/issues.ts
maxl0rd: I removed it because there wasn't a good error case in which to include it.

src/mcp/tools/crashlytics/reports.ts
maxl0rd: I found that this error case never happens so it doesn't really matter what we do here lol.

src/crashlytics/reports.ts
```diff
@@ -62,23 +66,26 @@ export function simplifyReport(report: Report): Report {
...
 
 export async function getReport(
-  report: CrashlyticsReport,
+  reportName: CrashlyticsReport,
   appId: string,
   filter: EventFilter,
   pageSize = DEFAULT_PAGE_SIZE,
 ): Promise<Report> {
+  if (!reportName) {
+    throw new FirebaseError("Invalid Crashlytics report " + reportName);
```
schnecle, line 75: It might be worth reminding it to read the manual here if it messes this up, though I really didn't see that happening.

src/crashlytics/reports.ts
```diff
@@ -62,23 +66,26 @@ export function simplifyReport(report: Report): Report {
 }
 
 export async function getReport(
-  report: CrashlyticsReport,
+  reportName: CrashlyticsReport,
   appId: string,
   filter: EventFilter,
   pageSize = DEFAULT_PAGE_SIZE,
 ): Promise<Report> {
+  if (!reportName) {
```
visumickey, line 74: Is there a default reportName that might be applied??

src/mcp/tools/crashlytics/index.ts
```diff
@@ -2,14 +2,7 @@ import type { ServerTool } from "../../tool";
...
 import { list_events, batch_get_events } from "./events";
-import {
-  get_top_issues,
-  get_top_variants,
-  get_top_versions,
-  get_top_apple_devices,
-  get_top_operating_systems,
-  get_top_android_devices,
-} from "./reports";
+import { get_report } from "./reports";
```
visumickey, line 5: When we remove MCP tools, what happens to the history of the MCP calls made by the agents in the IDE? Do you know if those caches would be invalidated? 

Not a blocker - but was curious what happens in those cases.

visumickey (approved):
Thanks for the changes!

src/mcp/tools/crashlytics/index.ts
maxl0rd: Previous conversation history has the tool responses burned in. Might confuse it if you restart a convo from a few weeks ago, but should be fine for the most part.

src/crashlytics/reports.ts
maxl0rd: Everything I tried always passes a correct report name. Default or empty would be an error.

src/crashlytics/reports.ts
maxl0rd: Yes, that is handled in the tool.

PR #9601, opened 2025-12-09
feat(mcp): fine grained control over tool loading

MichaelDoyle:
### Changes

- `--tools` will override discovery and provide _only_ the listed tools; existing discovery behavior is unchanged
- adds `--help, -h` to the mcp command to explain the available options

### Tools

image

### Help Text

```
michaeldoyle-mac2:firebase-tools michaeldoyle$ firebase mcp --help
Usage: firebase mcp [options]

Description:
  Starts the Model Context Protocol (MCP) server for the Firebase CLI. This server provides a
  standardized way for AI agents and IDEs to interact with your Firebase project.

Tool Discovery & Loading:
  The server automatically determines which tools to expose based on your project context.

  1. Auto-Detection (Default):
     - Scans 'firebase.json' for configured services (e.g., Hosting, Firestore).
     - Checks enabled Google Cloud APIs for the active project.
     - Inspects project files for specific SDKs (e.g., Crashlytics in Android/iOS).

  2. Manual Overrides:
     - Use '--only' to restrict tool discovery to specific feature sets (e.g., core, firestore).
     - Use '--tools' to disable auto-detection entirely and load specific tools by name.

Options:
  --dir <path>              Project root directory (defaults to current working directory).
  --only <features>         Comma-separated list of features to enable (e.g. core, firestore).
                            If specified, auto-detection is disabled for other features.
  --tools <tools>           Comma-separated list of specific tools to enable. Disables
                            auto-detection entirely.
  -h, --help                Show this help message.
```


src/bin/mcp.ts
```diff
@@ -22,19 +22,55 @@ This is a running process of the Firebase MCP server. This command should only b
...
+  --only <features>         Comma-separated list of features to enable (e.g. core, firestore).
+                            If specified, auto-detection is disabled for other features.
+  --tools <tools>           Comma-separated list of specific tools to enable. Disables
+                            auto-detection entirely.
+  --generate-tool-list      Generate markdown documentation for all available tools.
+  --generate-prompt-list    Generate markdown documentation for all available prompts.
+  --generate-resource-list  Generate markdown documentation for all available resources.
+  -h, --help                Show this help message.
+`;
+
```
MichaelDoyle, line 51: @joehan is there a more appropriate / idiomatic way to do this?

I know for CLI we just Commander, which more or less dictates how this is done.

src/bin/mcp.ts
```diff
@@ -53,9 +89,15 @@ export async function mcp(): Promise<void> {
   useFileLogger();
   const activeFeatures = (values.only || "")
     .split(",")
+    .map((f) => f.trim())
```
MichaelDoyle, line 89: Ran into this with the new flag, so it made sense to change here as well.

src/bin/mcp.ts
```diff
@@ -22,19 +22,55 @@ This is a running process of the Firebase MCP server. This command should only b
...
+     - Use '--only' to restrict tool discovery to specific feature sets (e.g., core, firestore).
+     - Use '--tools' to disable auto-detection entirely and load specific tools by name.
+
+Options:
+  --dir <path>              Project root directory (defaults to current working directory).
+  --only <features>         Comma-separated list of features to enable (e.g. core, firestore).
+                            If specified, auto-detection is disabled for other features.
+  --tools <tools>           Comma-separated list of specific tools to enable. Disables
+                            auto-detection entirely.
+  --generate-tool-list      Generate markdown documentation for all available tools.
```
joehan: We can hide these `--generate-*` flags - we've been using them to generate our internal docs, but they probably aren't useful for external users.

src/bin/mcp.ts
joehan: This seems fine for MCP, since we've totally cut commander out of the flow. We could build some more structure around flag handling here so that we don't need to write parsing logic for each - but I don't expect to have a ton more flags, so its probably overkill.

src/mcp/tools/index.ts
```diff
@@ -59,26 +42,66 @@ const tools: Record<ServerFeature, ServerTool[]> = {
...
-      },
-    },
-  }));
+const allToolsMap = new Map(
+  Object.values(tools)
+    .flat()
+    .map((t) => [t.mcp.name, t]),
+);
+
+function getToolsByName(names: string[]): ServerTool[] {
```
joehan, line 52: Behavior Q - right now, it looks like we just silently ignore it when you ask for a nonexistent tool (ie `--tools tool-doesnt-exist`). Would it be better to just error out with a clear message?

MichaelDoyle: Done. BTW, I noticed we don't error out for erroneous "features" (i.e. `--only`). I can do another PR that errors for that if it's the behavior we want.

MichaelDoyle:
@joehan added a few tests. punting on raising errors as the do not seem to be visible in mcp clients. we discussed potentially erroring on tool load, but not until its clear those errors will be visible in GCLI, AGY, etc.

joehan (approved):

PR #9612, opened 2025-12-11
Update functions CLI to support deploying `onGraphRequest` functions.

rosalyntan:
### Description

This adds support to the Firebase Functions CLI to deploy `onGraphRequest` functions. The deployed function will by default be IAM restricted to the Firebase Data Connect P4SA, with any explicitly specified invokers also granted access.

In addition, if `schemaFilePath` is specified in the `onGraphRequest` options, the CLI will pull in the schema GQL source files at `schemaFilePath` and bundle them with the functions deploy payload.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,3 @@
 - Updated TypeScript templates for `ext:dev:init` to fix build failures (#9524)
 - Fixed a bug when `firebase emulators:start` incorrectly deletes discovery file of another emulator process (#9672)
+- Added `onGraphRequest` function support (#9612)


src/deploy/functions/prepare.ts
```diff
@@ -230,13 +230,33 @@ export async function prepare(
...
 
     if (backend.someEndpoint(wantBackend, (e) => e.platform === "gcfv2")) {
-      const packagedSource = await prepareFunctionsUpload(sourceDir, localCfg);
+      const schPathSet = new Set<string>();
+      for (const e of backend.allEndpoints(wantBackend)) {
+        if (
+          backend.isDataConnectGraphqlTriggered(e) &&
+          e.dataConnectGraphqlTrigger.schemaFilePath
+        ) {
+          schPathSet.add(e.dataConnectGraphqlTrigger.schemaFilePath);
```
taeold, line 239: maybe we need to make some validation here that we don't escape out of firebase project dir, e.g.

`schemaFilePath: "../../.ssh/keys"`

src/deploy/functions/prepareFunctionsUpload.ts
```diff
@@ -89,6 +91,16 @@ async function packageSource(
         mode: file.mode,
       });
     }
+    for (const name of additionalSources) {
+      const absPath = path.join(projectDir, name);
+      const mode = fs.statSync(absPath).mode;
```
taeold, line 99: what happens if this file doesn't exist?

src/deploy/functions/build.ts
```diff
@@ -559,6 +576,21 @@ function discoverTrigger(endpoint: Endpoint, region: string, r: Resolver): backe
       httpsTrigger.invoker = endpoint.httpsTrigger.invoker.map(r.resolveString);
     }
     return { httpsTrigger };
+  } else if (isDataConnectGraphqlTriggered(endpoint)) {
+    const dataConnectGraphqlTrigger: backend.DataConnectGraphqlTrigger = {};
+    if (endpoint.dataConnectGraphqlTrigger.invoker === null) {
+      dataConnectGraphqlTrigger.invoker = null;
```
taeold, line 583: this would end up clearing up the invoker setting to "private", including getting rid of access for FDC P4SA 🤔  wondering if that's what we want

src/deploy/functions/build.ts
```diff
@@ -71,6 +71,14 @@ export interface HttpsTrigger {
   invoker?: Array<ServiceAccount | Expression<string>> | null;
 }
 
+export interface DataConnectGraphqlTrigger {
+  // Which service account should be able to trigger this function. No value means that only the
+  // Firebase Data Connect P4SA can trigger this function. For more, see go/cf3-http-access-control
```
taeold: then if the user sets invoker, FDC P4SA will not have permission to trigger the function?

src/deploy/functions/services/dataconnect.ts
```diff
@@ -18,3 +25,27 @@ export function ensureDataConnectTriggerRegion(
...
+/**
+ * Finds the required project level IAM bindings for the Firebase Data Connect service agent
+ * @param projectNumber project identifier
+ */
+export function obtainDataConnectBindings(projectNumber: string): Promise<Array<iam.Binding>> {
+  const dataConnectServiceAgent = `serviceAccount:${getServiceAccount(projectNumber)}`;
+  const cloudRunInvokerBinding = {
+    role: CLOUD_RUN_INVOKER_ROLE,
+    members: [dataConnectServiceAgent],
+  };
```
taeold: this would give FDC access to trigger all functions in a project. Is that really necessary? Maybe we should just mutate the invoker array for that specific function

src/deploy/functions/build.ts
rosalyntan: Updated the comment (and verified that setting a user invoker still keeps the FDC P4SA as an invoker).

src/deploy/functions/prepareFunctionsUpload.ts
rosalyntan: Added error handling.

src/deploy/functions/prepare.ts
rosalyntan: Done.

src/deploy/functions/services/dataconnect.ts
rosalyntan: Done.

src/deploy/functions/build.ts
rosalyntan: I think this is addressed with my changes in fabricator.ts -- what is a test case that would trigger this? I've verified that removing the invoker from the options when redeploying does not remove access for the FDC P4SA.

src/deploy/functions/release/fabricator.ts
```diff
@@ -421,6 +422,14 @@ export class Fabricator {
           .run(() => run.setInvokerCreate(endpoint.project, serviceName, invoker))
           .catch(rethrowAs(endpoint, "set invoker"));
       }
+    } else if (backend.isDataConnectGraphqlTriggered(endpoint)) {
+      const invoker = endpoint.dataConnectGraphqlTrigger.invoker ?? [];
```
taeold, line 427: can we add a comment here like other code blocks, e.g.

      // Like HTTPS triggers, dataConnectGraphqlTrigger have an invoker, but FDC P4SA must always be an invoker.

src/deploy/functions/release/index.ts
```diff
@@ -127,7 +127,9 @@ export async function release(
  * has updated the URI of endpoints after deploy.
  */
 export function printTriggerUrls(results: backend.Backend): void {
-  const httpsFunctions = backend.allEndpoints(results).filter(backend.isHttpsTriggered);
+  const httpsFunctions = backend
+    .allEndpoints(results)
+    .filter(backend.isHttpsTriggered || backend.isDataConnectGraphqlTriggered);
```
taeold: i don't think it's useful to print trigger url for FDC urls, esp. since it's not meant to be a public function.

src/deploy/functions/prepareFunctionsUpload.ts
```diff
@@ -89,6 +91,24 @@ async function packageSource(
...
     }
+    for (const name of additionalSources) {
+      const absPath = path.join(projectDir, name);
+      if (!fs.existsSync(absPath)) {
+        throw new FirebaseError(clc.bold(absPath) + " does not exist.", { exit: 1 });
+      }
+      if (path.relative(projectDir, absPath).includes("..")) {
+        throw new FirebaseError(clc.bold(absPath) + " is outside of the project directory.", {
+          exit: 1,
+        });
```
taeold: nit* there is `utils.ts/resolveWithin` helper

src/deploy/functions/release/fabricator.ts
rosalyntan: Done.

src/deploy/functions/release/index.ts
rosalyntan: We need the URL printed so developers can copy-paste it into their dataconnect.yaml files. However, I realized that this logic wasn't actually working, so updated with a fix. In addition, it seems like only the non-deterministic run.app URL gets returned from the backend, so I manually construct the deterministic URL here (it's a little hacky).

src/deploy/functions/prepareFunctionsUpload.ts
rosalyntan: Done.

PR #9621, opened 2025-12-15
Schnecle/crashlytics resources exploration

schnecle:
### Description

This is playing around with removing more instructions but emphasizing formatting and getting tool calls to encourage resource reading without the introduction of a session id.

### Scenarios Tested

/crashlytics:connect
Manually asking to investigate an individual issue and fetching reports.


maxl0rd:
Absorbed into PR 9594

PR #9622, opened 2025-12-16
add dataconnect sdk app setup validation

aalej:
### Description

Add validation when prompting which apps to add DataConnect SDK. Ideally, the scenario `if (!pickedApps || !pickedApps.length)` will never happen, but in the unlikely chance it does, it would be a good idea to keep it and throw an error

### Scenarios Tested

```
=== Dataconnect Setup
i  dataconnect: ensuring required API firebasedataconnect.googleapis.com is enabled...
i  dataconnect: ensuring required API sqladmin.googleapis.com is enabled...
✔ Your project already has existing services. Which would you like to set up local files for? us-central1/test-dc-2
⚠  dataconnect: CloudSQL no cost trial has already been used on this project.
✔  dataconnect: Detected existing apps web-app-2 (web), web-app (web)
? Which apps do you want to set up Data Connect SDKs in? (Press <space> to select, <a> to toggle all, <i> to invert selection, and <enter> to proceed)
❯◯ web-app-2 (web)
 ◯ web-app (web)
> Please choose at least one app.
```


fredzqm:
Good idea! Thank you!

PR #9626, opened 2025-12-16
New command that allows for a user to run mobile tests that are defined as yaml.

jrothfeder:
### Description

New command that allows for a user to run mobile tests that are defined as yaml. Tests will be read from a test dir and run remotely. Filters can be supplied to specify which tests are run.

### Scenarios Tested

Tested running mobile tests from the command line.

### Sample Commands

```
firebase apptesting:execute --app=1:359214138129:android:423758dfdb494e05f002b0 --test-dir=mytests ./app/build/outputs/apk/alpha/debug/app-alpha-debug.apk
```


src/experiments.ts
```diff
@@ -165,6 +165,10 @@ export const ALL_EXPERIMENTS = experiments({
     shortDescription: "Adds experimental App Testing feature",
     public: true,
   },
+  mata: {
+    shortDescription: "Adds experimental Mobile App Testing Agent (MATA) feature",
```
kjelko: i don't think we want to expose the MATA naming anywhere. Is it not possible to keep this under the apptesting experiment?

jrothfeder: Sure

src/commands/apptesting-mobile-execute.ts
```diff
@@ -0,0 +1,102 @@
...
+const defaultDevices = [
+  {
+    model: "MediumPhone.arm",
+    version: "30",
+    locale: "en_US",
+    orientation: "portrait",
+  },
+];
+
+export const command = new Command("apptesting:mobile-execute <target>")
```
tagboola: Did you consider putting this under the app distribution namespace?

jrothfeder: Yes. We went back and forth on that.

src/commands/apptesting-mobile-execute.ts
```diff
@@ -0,0 +1,115 @@
...
```
kaibolay: Having a dash in the command is out-of-style for the firebase CLI.

So this could be `apptesting:mobile:execute`, but that'd be weird. I think just extending the existing `apptesting:execute` makes much more sense (mobile vs. web can be determined by looking at the app ID).

Or the app distribution namespace...

src/apptesting/parseTestFiles.spec.ts
```diff
@@ -204,4 +214,166 @@ describe("parseTestFiles", () => {
...
   });
+  describe("prerequisite test cases", () => {
+    it("merges the steps from the prerequisite test case", async () => {
+      writeFile(
+        "my_test.yaml",
+        stringify({
+          tests: [
+            {
+              id: "my-first-test",
+              testName: "my first test",
```
kaibolay: Shouldn't this be `displayName`?  Why does this work?

src/apptesting/parseTestFiles.spec.ts
```diff
@@ -204,4 +214,166 @@ describe("parseTestFiles", () => {
...
+              steps: [{ goal: "do something first" }],
```
kaibolay, line 217: I'm confused: In other tests in this file I see the `steps` under `instructions`. Is both supported?

The problem is that we want to be compatible with the format produced by `appdistribution:testcases:export` and the existing web test configuration. I think we need some better test coverage for that.

src/apptesting/parseTestFiles.spec.ts
tagboola: WATA uses `testName`. This initially tripped me up as well, so maybe we need to chat some more about this.

src/apptesting/parseTestFiles.spec.ts
tagboola: So we need to be compatible with the WATA file format _and_ the format produced by `appdistribution:testcases:export`? I wonder if we can just decide on one particular format instead...

jrothfeder: This keeps coming up but there's no requirement that we are compatible with `appdistribution:export`. Being compatible with wata makes sense here and also allows us to reuse the testcase loading code.

src/apptesting/invokeTests.spec.ts
```diff
@@ -88,14 +86,12 @@ describe("invokeTests", () => {
...
-                    {
-                      goal: "retest it",
-                      successCriteria: "a dialog appears",
-                    },
-                  ],
-                },
+                steps: [
+                  {
+                    goal: "retest it",
+                    successCriteria: "a dialog appears",
```
lfkellogg, line 92: We had discussed changing this to `finalScreenAssertion`. Do you still plan to make that change in this CL?

kaibolay: let's do that in a separate PR which changes it here and for import/export at the same time

src/apptesting/parseTestFiles.ts
```diff
@@ -1,76 +1,151 @@
...
 
-function createFilter(pattern?: string) {
-  const regex = pattern ? new RegExp(pattern) : undefined;
-  return (s: string) => !regex || regex.test(s);
-}
-
 export async function parseTestFiles(
   dir: string,
-  targetUri: string,
+  targetUri?: string,
```
kaibolay, line 10: should this be removed?

src/commands/apptesting.ts
```diff
@@ -0,0 +1,115 @@
...
+const defaultDevices = [
+  {
+    model: "MediumPhone.arm",
+    version: "30",
+    locale: "en_US",
+    orientation: "portrait",
+  },
+];
+
+export const command = new Command("apptesting:execute <target>")
```
kaibolay: What is `<target>`?

kaibolay: For `appdistribution:distribute` this is called `<release-binary-file>` which is much more descriptive. I think we should use the same here for consistency.

src/apptesting/parseTestFiles.ts
jrothfeder: targetUri is used by wata so it can be removed if/when we remove wata

src/apptesting/parseTestFiles.spec.ts
kaibolay, line 173: Rename this method?

It's really unfortunate that Google APIs use "name" to mean resource name, but we can't change that. So, with that in mind, "test case name" means "resource name of the test case", which has the format `projects/{project_number}/apps/{app}/testCases/{test_case}`.

So to avoid any confusion, I think we shouldn't use "test case name", but "test case display name" (not just for this method, but throughout the code)

templates/init/apptesting/smoke_test.yaml
```diff
@@ -1,5 +1,5 @@
 tests:
```
kaibolay, line 1: How difficult would it be to make the parser more forgiving, so that this `tests:` would be optional?

src/commands/apptesting.ts
```diff
@@ -0,0 +1,124 @@
...
+import { AppDistributionClient } from "../appdistribution/client";
+import { Distribution, upload } from "../appdistribution/distribution";
+import { AIInstruction, ReleaseTest, TestDevice } from "../appdistribution/types";
+import { getAppName, parseTestDevices } from "../appdistribution/options-parser-util";
+
+// TODO rothbutter add ability to specify devices
+const defaultDevices = [
+  {
+    model: "MediumPhone.arm",
+    version: "30",
```
lfkellogg: Can you default to API level 36? (the current latest)

src/commands/apptesting.ts
```diff
@@ -0,0 +1,124 @@
...
+import * as ora from "ora";
+import { TestCaseInvocation } from "../apptesting/types";
+import { FirebaseError, getError } from "../error";
+import { marked } from "marked";
+import { AppDistributionClient } from "../appdistribution/client";
+import { Distribution, upload } from "../appdistribution/distribution";
+import { AIInstruction, ReleaseTest, TestDevice } from "../appdistribution/types";
+import { getAppName, parseTestDevices } from "../appdistribution/options-parser-util";
+
+// TODO rothbutter add ability to specify devices
```
lfkellogg: I think you can remove this TODO

src/commands/apptesting.ts
```diff
@@ -0,0 +1,124 @@
...
+      clc.bold(`\n${clc.white("===")} Running ${pluralizeTests(testInvocations.length)}`),
+    );
+    logger.info(await marked(`View progress and results in the Firebase Console`));
+  });
+
+function pluralizeTests(numTests: number) {
+  return `${numTests} test${numTests === 1 ? "" : "s"}`;
+}
+
+async function invokeMataTests(
```
lfkellogg: Nit: I'd avoid using "MATA" anywhere since its an internal name. "ATA" would be fine, or maybe "release tests" to match the API.

PR #9631, opened 2025-12-17
middleware/proxy fixes for next.js 15/16

leoortizz:
### Description

This PR updates the Next.js framework adapter to seamlessly support middleware changes introduced in Next.js 16, specifically handling both the new `proxy` convention and the deprecated `middleware` convention.

*   **Next.js 16 Support**: Implements support for reading middleware configuration from `functions-config-manifest.json`, ensuring compatibility with Next.js 16's Node.js runtime middleware (`proxy.ts`/`js`).
*   **Backward Compatibility**: Adds a fallback mechanism to support the deprecated `middleware.ts`/`js` convention by reading from `middleware-manifest.json`.
*   **Dev Mode Detection**: Updates `isUsingMiddleware` to automatically detect `proxy.js` and `proxy.ts` files in development mode, in addition to existing checks.
*   **Type Definitions**: Adds comprehensive TypeScript interfaces for Next.js 16 manifests (`FunctionsConfigManifest`, `MiddlewareManifestV3`, etc.) with documentation linking to the source.
*   **Test Refactoring**: Reorganizes middleware tests by version and environment for better clarity and maintainability, adding specific test cases for deprecated middleware scenarios and proxy detection.

### Scenarios Tested

- Confirmed that middleware matchers are correctly extracted from `functions-config-manifest.json` when using the new `proxy.ts` convention.
- Confirmed that the adapter correctly falls back to `middleware-manifest.json` when using the deprecated `middleware.ts` convention.
- Verified detection of `proxy.js`/`ts` alongside `middleware.js`/`ts` in dev mode(emulator).
- Verified correct middleware detection and matcher extraction based on manifest presence when deployed.

### Sample Commands

`firebase deploy` will now automatically handle Next.js 16 projects using either the new proxy or legacy middleware conventions without user intervention.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,3 @@
 - Fixed an issue where functions deployments would silently fail (#6989)
 - Add support for Next.js 16 middleware (`proxy.ts`/`proxy.js`) (#9631)
+- Updates the default region for new App Hosting backends to us-east4 (#10271)
--- a/src/apphosting/backend.spec.ts
+++ b/src/apphosting/backend.spec.ts
@@ -347,7 +347,7 @@ describe("apphosting setup functions", () => {
 
   describe("promptLocation", () => {
     const supportedLocations = [
-      { name: "us-central1", locationId: "us-central1" },
+      { name: "us-east4", locationId: "us-east4" },
       { name: "us-west1", locationId: "us-west1" },
     ];
 
@@ -358,16 +358,16 @@ describe("apphosting setup functions", () => {
 
     it("returns a location selection", async () => {
       const location = await promptLocation(projectId, /* prompt= */ "");
-      expect(location).to.be.eq("us-central1");
+      expect(location).to.be.eq("us-east4");
     });
 
     it("uses a default location prompt if none is provided", async () => {
       await promptLocation(projectId);
 
       expect(promptStub.select).to.be.calledWith({
-        default: "us-central1",
+        default: "us-east4",
         message: "Please select a location:",
-        choices: ["us-central1", "us-west1"],
+        choices: ["us-east4", "us-west1"],
       });
     });
 
--- a/src/apphosting/constants.ts
+++ b/src/apphosting/constants.ts
@@ -1,3 +1,3 @@
-export const DEFAULT_LOCATION = "us-central1";
+export const DEFAULT_LOCATION = "us-east4";
 export const DEFAULT_DEPLOY_METHOD = "github";
 export const ALLOWED_DEPLOY_METHODS = [{ name: "Deploy using github", value: "github" }];


src/frameworks/next/interfaces.ts
```diff
@@ -83,10 +83,104 @@ export interface ExportMarker {
...
+  | {
+      type: "header" | "cookie" | "query";
+      key: string;
+      value?: string;
+    }
+  | {
+      type: "host";
+      key?: undefined;
+      value: string;
+    };
```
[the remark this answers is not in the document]
leoortizz: the local dependency is subject to change, risking to break the logic that rely on this type.

src/frameworks/next/interfaces.ts
```diff
@@ -83,10 +83,104 @@ export interface ExportMarker {
...
+      regions?: string[] | string;
+      matchers?: Array<{
+        regexp: string;
+        originalSource: string;
+        has?: RouteHas[];
+        missing?: RouteHas[];
+      }>;
+    }
+  >;
+}
```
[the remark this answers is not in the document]
leoortizz: the types here are intentionally copied verbatim from the Next.js codebase to ensure parity

src/frameworks/next/interfaces.ts
```diff
@@ -83,10 +83,104 @@ export interface ExportMarker {
   isNextImageImported: boolean;
 }
 
-export type MiddlewareManifest = MiddlewareManifestV1 | MiddlewareManifestV2FromNext;
+export type MiddlewareManifest =
+  | MiddlewareManifestV1
+  | MiddlewareManifestV2FromNext
+  | MiddlewareManifestV3;
```
annajowang, line 89: Are there reasonable more descriptive names that could be used instead of V1 or V3?

Are these all just different manifest types introduced at different nextjs versions?

For example could MiddlewareManifestNext16 or something work for MiddlewareManifestV3 ?

annajowang: ah nvm, these are literally the versions on the interface itself.

src/frameworks/next/utils.ts
```diff
@@ -195,24 +198,36 @@ export async function hasUnoptimizedImage(sourceDir: string, distDir: string): P
...
 
 /**
- * Whether Next.js middleware is being used
+ * Whether Next.js proxy/middleware is being used
  *
  * @param dir in development must be the project root path, otherwise `distDir`
  * @param isDevMode whether the project is running on dev or production
  */
 export async function isUsingMiddleware(dir: string, isDevMode: boolean): Promise<boolean> {
   if (isDevMode) {
```
annajowang, line 207: Sorry maybe it's more obvious to those who know nextjs really well. But why is checking for middleware and proxy files in devMode enough? Could that be documented?

src/frameworks/next/interfaces.ts
annajowang: Actually, would we some day remove support for very old nextjs versions? In which case would some of these manifest versions would no longer be needed? could those versions be in the docstrings for MiddlewareManifestv2 and v1 ?
I think the "Middleware manifest types for Next.js 16" you added for v3 is very helpful.

src/frameworks/next/interfaces.ts
annajowang, line 3: Was there a specific reason we decided not to copy v2 over like the others?

are we pinning to a specific version of the plugin somewhere? if that version gets wouldn't this break?

src/frameworks/next/utils.ts
leoortizz: IIRC the main reason is that the manifest files were not always available in dev mode, e.g. if it's the first time the Next.js dev server is running. Added a comment here

src/frameworks/next/interfaces.ts
leoortizz: We're not copying it intentionally here to have less code, and to make sure we're matching the Next.js interfaces. It's coming directly from the Next.js source code as it's a dev dependency. It's currently [^14.1.0](https://github.com/firebase/firebase-tools/blob/8cf8a3a573e19f319f58393d7f9c28cf52e6922d/package.json#L250). If the dependency version changes this will indeed break, requiring manual changes. That's how we're making sure it's adhering to the types we want.

One alternative I tried in the past was to have multiple Next.js versions as dependencies(with aliases), enabling us to just import the types rather than rewriting them. However, that caused type errors, breaking the firebase-tools build.

src/frameworks/next/interfaces.ts
leoortizz: > Actually, would we some day remove support for very old nextjs versions?

Deferring to @jamesdaniels, but I don't think we would remove support for old Next.js versions as they just work?

> In which case would some of these manifest versions would no longer be needed?

The manifests won't be needed anymore as soon as we migrate to use the [Adapters API](https://nextjs.org/docs/app/guides/upgrading/version-16#build-adapters-api-alpha). That's also why I chose to copy the V3 types instead of bumping the Next.js dependency, as it would require more changes that would not be necessary when we switch to the Adapters API.

> could those versions be in the docstrings for MiddlewareManifestv2 and v1 ?
I think the "Middleware manifest types for Next.js 16" you added for v3 is very helpful.

The version of manifests does not match the Next.js version necessarily. I updated the comments to say in which versions they were used. I realized the manifest-middleware version 3 changed in Next.js 14.2.0, so I updated the comment. It's just that in Next.js 16 the proxy route matchers are in functions-config-manifest instead.

src/frameworks/next/interfaces.ts
```diff
@@ -83,11 +83,114 @@ export interface ExportMarker {
...
 
+/**
+ * Middleware manifest type used between Next.js 12.2.0- 14.1.4
```
leoortizz:
```suggestion
 * Middleware manifest type used between Next.js 12.2.0 - 14.1.4
```

PR #9671, opened 2025-12-26
Starting publishing releases as drafts

joehan:
### Description
Change our publishing script to create draft releases. The release runner will then manually check that all of the release artifacts are in order, and then publish the release via the GitHub UI.

This will let us safely enable immutable releases.


morganchen12 (approved):

morganchen12:
Please correct the gemini review comment before merging (reference: https://hub.github.com/hub-release.1.html)

PR #9672, opened 2025-12-27
[VS Code] Fix the emulator discovery bugs

fredzqm:
### Issue

When emulator is started quickly, there is a race condition that leaves the hub locator file deleted with active emulator process running.

We can easily reproduce this bug by clicking "run local" button repeatedly quickly. [Reproduce](https://screencast.googleplex.com/cast/NDcyNjE3MjQ0MDA2ODA5NnwzMjRiOTI3Ny1jZg)

1. p1 `firebase emulator:start`
2. p1 write locator file
3. p2 `firebase emulator:start`
4. p2 write locator file
5. p2 exits due to port occupied
6. p2 deletes the locator file

....
p1 continues to run without a locator file, so VS Code don't know about it. 

### Description

Once a process wrote the locator file, subsequent emulator shouldn't override it.
Add `pid` to the locator file to ensure a process can only delete its own locator.

1. Read existing file
2. Check if another process is active based on `locator.pid`.
3. Write the file with `{version, origins, pid}`.
4. Delete the locator file only if the process wrote it.

There is still a possibilities of race condition if between step 1 and 3. The chances are much smaller than before. I looked into safer solution using advisory file lock, but got stuck trying to install `npm i fs-ext`. This fix is an incremental improvement.

[data-connect.entry.tsx‎](https://github.com/firebase/firebase-tools/pull/9672/changes#diff-20809b8bc9e624f8614b21521a8a04c38949ca02ff7895dc98c04b444046be94) is not used

When "run local" is clicked, try to connect to the emulator first before start one.

### Tested

A) Click "run local" repeatedly

Expectation: Only one emulator starts successfully. "run local" is able to discover and hit it.

B) Start an emulator, refresh emulator view to find it, stop the emulator process, then hit "run local".

Expectation: VS Code should realize emulator is no longer running and try to start it again.

C) VS has no project selected. `firebase use some-project`, Hit "run local"

Expectation: It should start an emulator under the project configured in VS Code `demo-no-project` in this case.

### DIFF
--- a/firebase-vscode/CHANGELOG.md
+++ b/firebase-vscode/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## NEXT
 
-- Fix the emulator discovery bug when "run local" starts the emulator repeatedly.
+- Fix the data connect emulator discovery bugs.
 
 ## 2.0.0
 
--- a/firebase-vscode/src/core/emulators.ts
+++ b/firebase-vscode/src/core/emulators.ts
@@ -21,7 +21,10 @@ export class EmulatorsController implements Disposable {
 
     // called by emulator UI
     this.subscriptions.push(
-      broker.on("runStartEmulators", () => {
+      broker.on("runStartEmulators", async () => {
+        if (await this.areEmulatorsRunning()) {
+          return;
+        }
         this.startEmulators();
       }),
     );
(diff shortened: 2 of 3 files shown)


src/emulator/hub.ts
```diff
@@ -194,14 +188,13 @@ export class EmulatorHub extends ExpressBasedEmulator {
 
   async stop(): Promise<void> {
     await super.stop();
-    await this.deleteLocatorFile();
```
fredzqm, line 197: This line was buggy. It blinds deletes the locator file regardless whether it was created by itself.

PR #9694, opened 2026-01-02
Adding support for enabling auth providers via firebase deploy

joehan:
### Description
First draft of adding Auth provisioning support as part of `firebase deploy`

### Scenarios Tested
Tested on a few new and existing projects:
Screenshot 2026-01-05 at 9 02 36 AM
Screenshot 2026-01-05 at 9 14 37 AM
Screenshot 2026-01-05 at 5 44 29 PM
Screenshot 2026-01-05 at 5 44 38 PM
Screenshot 2026-01-05 at 5 45 11 PM

### Known Issues
This currently does not work if you are using Application Default Credentials, as the required API fields are only visible on the CLIs consumer project. Working with Auth + Orchestration API folks to see if we can lift that restriction. It

### DIFF
--- a/.agent/skills/update-pubsub-emulator/SKILL.md
+++ b/.agent/skills/update-pubsub-emulator/SKILL.md
@@ -1,0 +1,62 @@
+---
+name: Update Pub/Sub Emulator
+description: How to update the Pub/Sub emulator
+---
+
+# Update Pub/Sub Emulator
+
+1.  **Update Local Emulator**
+    Run the following command to make sure you have the latest version of the pubsub emulator installed via gcloud:
+    ```bash
+    gcloud components update pubsub-emulator
+    ```
+
+2.  **Locate Emulator Directory**
+    The emulator represents a directory likely located at `<gcloud-install-path>/platform/pubsub-emulator`.
+    You can find the exact path by running the emulator and checking the output, or by inspecting where `gcloud` is installed (e.g. `which gcloud` usually points to a bin directory, and the platform directory is a sibling of `bin`).
+    Verify the version by running the emulator or checking the `VERSION` file if it exists.
+
+3.  **Package the Emulator**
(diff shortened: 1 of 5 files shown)


src/commands/deploy.ts
```diff
@@ -73,6 +74,8 @@ export const TARGET_PERMISSIONS: Record<(typeof VALID_DEPLOY_TARGETS)[number], s
     "firebasedataconnect.schemas.list",
     "firebasedataconnect.schemas.update",
   ],
+  apphosting: ["firebaseapphosting.sites.update"],
```
TrCaM: Doesn't seem related to this PR. Should we move it to a different PR?

src/firebaseConfig.ts
```diff
@@ -319,6 +319,18 @@ export type AppHostingMultiple = AppHostingSingle[];
...
 export type AppHostingConfig = AppHostingSingle | AppHostingMultiple;
 
+export interface AuthConfig {
+  providers?: {
+    anonymous?: boolean;
+    emailPassword?: boolean;
+    googleSignIn?: {
+      oAuthBrandDisplayName?: string;
+      supportEmail?: string;
+      authorizedRedirectUris?: string[];
```
TrCaM, line 329: Looks like we doesn't allow users to provide `authorizedRedirectUris` from the init command for simplicity is it  correct? Users will need to edit firebase.json to modify that after init?

TrCaM (approved):
Just a couple minor questions but overall LGTM! Thanks for driving the whole API approval and implementation!

src/firebaseConfig.ts
joehan: Ya, I intentionally omitted that from init since its a bit more of an advanced feature - if you just want to use the default hosting domian, it will be authorized automatically. You only need to add to this list if you are using custom domains.

PR #9703, opened 2026-01-06
Handle invalid connector errors when deploying secondary schemas

rosalyntan:
### Description

Handle invalid connector errors when deploying secondary schemas the same way we handle them for main schema.

### Sample Commands

Screenshot 2026-01-06 at 5 04 18 PM

### DIFF
--- a/src/dataconnect/schemaMigration.ts
+++ b/src/dataconnect/schemaMigration.ts
@@ -311,7 +311,7 @@ export async function upsertSecondarySchema(args: {
   stats?: DeployStats;
 }): Promise<void> {
   const { options, schema, stats } = args;
-  const serviceName = schema.name.replace(`/schemas/${schema.name.split("/").pop()}`, "");
+  const serviceName = serviceNameFromSchema(schema);
   try {
     await upsertSchema(schema, false);
   } catch (err: any) {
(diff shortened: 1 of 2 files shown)


src/dataconnect/schemaMigration.ts
```diff
@@ -304,6 +304,46 @@ export async function migrateSchema(args: {
...
 }
 
+/** Upsert a secondary schema, handling invalid connector errors. */
+export async function upsertSecondarySchema(args: {
+  options: Options;
+  schema: Schema;
+  stats?: DeployStats;
+}): Promise<void> {
+  const { options, schema, stats } = args;
+  const serviceName = schema.name.replace(`/schemas/${schema.name.split("/").pop()}`, "");
```
yuchenshi: nit: This is a bit awkward to read. If you want to remove the last two segments, just pop twice from the split array to make it more readable. Or use a regex if it helps. Also consider making a helper function called `serviceNameFromSchemaName`

yuchenshi (approved):

src/dataconnect/schemaMigration.ts
rosalyntan: Done.

PR #9705, opened 2026-01-06
Remove tools and prompts that rely on GiF ToS

joehan:
### Description
Remove dataconnect_generate_operation, dataconnect_generate_schema, and /firebase:consult, to fully remove the reliance on GiF ToS.

Alternatively, we could remove just the tools, and add a ToS disclaimer to the description of /firebase:consult


fredzqm:
Sad. Hopefully, we will see them back soon

Issue #9663, opened 2025-12-19
[MCP] Firebase MCP server hangs/crashes during tools/list initialization with large datasets


stevexusc:
# [MCP] Firebase MCP server hangs/crashes during tools/list initialization with large datasets

## Description
The Firebase MCP server (`firebase-tools@latest mcp`) hangs, times out, or crashes with out-of-memory errors when Claude Code (or other MCP clients) attempts to list available tools via the `tools/list` MCP method. This issue occurs specifically when the Firebase project contains a large volume of Crashlytics data, making the MCP integration completely unusable for production applications.

**This happens with BOTH configurations:**
- With `--only crashlytics` flag
- Without any `--only` flag (loading all features)

## Environment
- **firebase-tools version**: `latest` (tested with automatic updates via `npx -y`)
- **MCP client**: Claude Code v2.0.73
- **Node.js version**: v20.18.1
- **OS**: macOS (Darwin 25.2.0)
- **Firebase project**: Large production Android app with extensive Crashlytics history
- **MCP server configs tested**:
  1. `npx -y firebase-tools@latest mcp --only crashlytics`
  2. `npx -y firebase-tools@latest mcp --dir /path/to/project` (no --only flag)

## Steps to Reproduce

1. **Set up a Firebase project with significant Crashlytics data**
   - Production Android app
   - Multiple app variants
   - Months/years of crash data
   - Example: SafetyCulture Android app (1:140348146905:android:0d46d219d040c070)

2. **Configure Firebase MCP server in Claude Code**

   **Config A (with --only flag):**
   ```json
   {
     "mcpServers": {
       "firebase": {
         "type": "stdio",
         "command": "npx",
         "args": ["-y", "firebase-tools@latest", "mcp", "--only", "crashlytics"]
       }
     }
   }
   ```

   **Config B (without --only flag, per documentation):**
   ```json
   {
     "mcpServers": {
       "firebase": {
         "type": "stdio",
         "command": "npx",
         "args": ["-y", "firebase-tools@latest", "mcp", "--dir", "/path/to/project"]
       }
     }
   }
   ```

3. **Authenticate with Firebase**
   ```bash
   firebase login
   ```

4. **Start Claude Code session**
   ```bash
   claude
   ```

5. **MCP server attempts to initialize and list tools**

## Expected Behavior

The MCP server should:
- Successfully initialize
- Return a list of available Crashlytics tools via `tools/list`
[...]

PR #9713, opened 2026-01-08
fix(mcp): Update --only behavior to bypass availability checks

schnecle:
### Description

Updates the behavior of --only and --tools flags to bypass availability checks.

Why make this change?
=================

When attempting to reproduce behavior for https://github.com/firebase/firebase-tools/issues/9663, I noticed that --only was not working for Crashlytics in directories that have non-standard dependency management because it was still executing an availability check. This availability check looks for a dependency on Crashlytics for apps that use gradle, swift package manager, pod, carthage, and flutter's standard pubspec.yaml file.

We direct customers to use the --only flag if they do not fall into one of the supported categories above for Crashlytics MCP.

### Scenarios Tested

* Added unit tests
* Ran the mcpinspector in a directory with no Crashlytics dependency with the --only flag and saw Crashlytics tools loaded
* Ran the mcpinspector in a directory with no Crashlytics dependency without the --only flag and saw Crashlytics tools loaded
* Ran the mcpinspector in a directory with no Crashlytics dependency with the --tools flag with crashlytics_create_note and saw that tool loaded
* Ran Gemini CLI in a directory with no Crashlytics dependency with the --only flag and saw Crashlytics prompts and tools loaded
* Ran Gemini CLI in a directory with a Crashlytics dependency and saw auto-detection still works

### Sample Commands

`gemini`
`npx @modelcontextprotocol/inspector firebase mcp --only crashlytics`

### DIFF
--- a/src/mcp/prompts/index.spec.ts
+++ b/src/mcp/prompts/index.spec.ts
@@ -50,10 +50,9 @@ describe("availablePrompts", () => {
 
   it("should include feature-specific prompts when activeFeatures is provided", async () => {
     const prompts = await availablePrompts(mockContext, ["crashlytics"]);
-    const crashPrompt = prompts.find((p) => p.mcp._meta?.feature === "crashlytics");
 
-    // Should exist because we bypass availability check
-    expect(crashPrompt).to.exist;
+    const features = [...new Set(prompts.map((p) => p.mcp._meta?.feature))];
+    expect(features).to.have.members(["core", "crashlytics"]);
 
     // getDefaultFeatureAvailabilityCheck execution is deferred/lazy-loaded in prompt.ts
     // Since activeFeatures bypasses checking .isAvailable on the prompt, the stub should NOT be called.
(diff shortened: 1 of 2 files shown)


src/mcp/prompts/index.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    const corePrompt = prompts.find((p) => p.mcp._meta?.feature === "core");
+    expect(corePrompt).to.exist;
+  });
+
+  it("should include feature-specific prompts when activeFeatures is provided", async () => {
+    const prompts = await availablePrompts(mockContext, ["crashlytics"]);
+    const crashPrompt = prompts.find((p) => p.mcp._meta?.feature === "crashlytics");
+
+    // Should exist because we bypass availability check
+    expect(crashPrompt).to.exist;
```
raymondlam: nit: it is probably better to assert on the exact list of features we expect from the list of prompts (so that we check that core is included as well)

```
const features = [...new Set(prompts.map((p) => p.mcp._meta?.feature))];
expect(features).to.have.members(['core', 'crashlytics']);
```

schnecle: Done.

PR #9730, opened 2026-01-13
Adding better tracking of AI agents responsible for API calls

joehan:
### Description

In apiv2.ts, we have some STANDARD_HEADERS that we include on all requests. We use these to attribute API usage to the CLI.

Currently, they look like:

  "User-Agent": `FirebaseCLI/${CLI_VERSION}`,
  "X-Client-Version": `FirebaseCLI/${CLI_VERSION}`,

After this change, they will look like 

  "User-Agent": `Firebase${CLI or MCP}$/${CLI_VERSION}/${AgentName or "None"}`,
  "X-Client-Version": `Firebase${CLI or MCP}$/${CLI_VERSION}/${AgentName or "None"}`


src/utils.ts
```diff
@@ -1055,3 +1055,15 @@ export function resolveWithin(base: string, subPath: string, errMsg?: string): s
...
+export function detectAIAgent(): string {
+  if (process.env.ANTIGRAVITY_CLI_ALIAS) return "antigravity";
+  if (process.env.CLAUDECODE) return "claude_code";
+  if (process.env.CLINE_ACTIVE) return "cline";
+  if (process.env.CODEX_SANDBOX) return "codex_cli";
+  if (process.env.CURSOR_AGENT) return "cursor";
+  if (process.env.GEMINI_CLI) return "gemini_cli";
+  if (process.env.OPENCODE) return "open_code";
+  return "unknown";
+}
```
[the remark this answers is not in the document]
joehan: This is not better,  just different, and limits us from looking at other environmental hints.

src/apiv2.ts
```diff
@@ -18,10 +19,15 @@ import * as FormData from "form-data";
 const pkg = require("../package.json");
 const CLI_VERSION: string = pkg.version;
 
+const agent = detectAIAgent();
+const agentStr = agent === "unknown" ? "" : ` agent-name/${agent}`;
+const platform = isFirebaseMcp() ? "FirebaseMCP" : "FirebaseCLI";
+const clientVersion = `${platform}/${CLI_VERSION}${agentStr}`;
```
VinayGuthal, line 25: should be ${platform}/${CLI_VERSION} agent-name/${agentStr}

joehan: It is - see l23:

agentStr = ` agent-name/${agent}`

We just do this slightly awkward construction so that when there is no agent, we send only `${platform}/${CLI_VERSION}` instead of `${platform}/${CLI_VERSION} agent-name/unknown`

VinayGuthal: ahh i see you are right

VinayGuthal: Resolve

PR #9733, opened 2026-01-13
Adding `/firestore:generate_rules` and `/storage:generate_rules` MCP prompts.

joehan:
### Description
Moving these prompts for safe security rule generation over from the snippets repo.
Screenshot 2026-01-13 at 3 49 07 PM
Screenshot 2026-01-13 at 3 48 57 PM

### TODO

- [ ] Do we want to inject any project specific info? We can do that now since these are programmatically defined
- [ ] We're looking at adding agent skills - when we do so, this is probably better suited as a skill so that it can be triggered by agents instead of requiring a manual slash command from the user.

### DIFF
--- a/src/mcp/prompts/firestore/generate_rules.ts
+++ b/src/mcp/prompts/firestore/generate_rules.ts
@@ -1,7 +1,7 @@
 import { prompt } from "../../prompt";
 
 export const generateRules = prompt(
-  "core",
+  "firestore",
   {
     name: "generate_rules",
     description:
@@ -284,7 +284,7 @@ npm test
 
 Provide your response in this structure:
 
-''''
+'''
 ## Analysis Summary
 [Your codebase analysis findings]
 
@@ -332,7 +332,7 @@ rules_test/
   - rules_test/firestore.rules
   - rules_test/tests/firestore.test.js
   - rules_test/README.md
-''''
+'''
 
 **After providing the analysis and summary, create all necessary files:**
 
--- a/src/mcp/prompts/storage/generate_rules.ts
+++ b/src/mcp/prompts/storage/generate_rules.ts
@@ -1,7 +1,7 @@
 import { prompt } from "../../prompt";
 
 export const generateRules = prompt(
-  "core",
+  "storage",
   {
     name: "generate_rules",
     description:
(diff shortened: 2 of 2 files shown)


src/mcp/prompts/firestore/generate_rules.ts
```diff
@@ -0,0 +1,387 @@
+import { prompt } from "../../prompt";
+
+export const generateRules = prompt(
+  "firestore",
+  {
+    name: "generate_rules",
```
rachelsaunders, line 6: Can this pls be:

`generate_security_rules`

"Rules" are an overloaded term in the AI world, and this could be confused for generating "AI rules" for Firestore -- like general best practices for using the product.

src/mcp/prompts/firestore/generate_rules.ts
```diff
@@ -0,0 +1,387 @@
...
+
+## Security Analysis
+[Devil's advocate findings and iterations]
+
+## Validation Results
+[Results from 'firebase_validate_security_rules' tool]
+
+## Generated Files Structure
+
+A complete 'rules_test/' directory will be created as an independent Node.js project:
```
rachelsaunders, line 299: Since this directory is for FIRESTORE specifically, I discussed with the team that this should be `security_rules_test_firestore`.

(leading with `security_rules_test_` so that it sorts with the analogous Storage directory)

src/mcp/prompts/storage/generate_rules.ts
```diff
@@ -0,0 +1,275 @@
...
+
+## Security Analysis
+[Devil's advocate findings and iterations]
+
+## Validation Results
+[Results from 'firebase_validate_security_rules' tool]
+
+## Generated Files Structure
+
+A complete 'storage_rules_test/' directory will be created as an independent Node.js project:
```
rachelsaunders, line 145: To match with the analogous Firestore directory, I discussed with the team that this should be `security_rules_test_storage`.

(leading with `security_rules_test_` so that it sorts with the analogous Firestore directory)

src/mcp/prompts/storage/generate_rules.ts
```diff
@@ -0,0 +1,275 @@
+import { prompt } from "../../prompt";
+
+export const generateRules = prompt(
+  "storage",
+  {
+    name: "generate_rules",
```
rachelsaunders, line 6: Can this pls be:

`generate_security_rules`

"Rules" are an overloaded term in the AI world, and this could be confused for generating "AI rules" for Storage -- like general best practices for using the product.

PR #9734, opened 2026-01-14
[FDC] Remove the `ExtendQueryPatch`

fredzqm:
Removed the extended query patch because we switched FDC emulator to always use non-prepared SQL, so error can be properly surfaced back. The upstream protocol bug still exists, but we got around it.

Improved log format to include the message type name:
```
[-> Execute] E���	�����
[<- CommandComplete] C���
SELECT 0�
[-> Sync] S���
[<- ReadyForQuery] Z���I
[-> Query] Q���%SELECT extname FROM pg_extension�
[<- RowDescriptionMessage] T��� �extname��������@������
[<- DataRow] D�������plpgsql
[<- DataRow] D�������	uuid-ossp
[<- CommandComplete] C���
SELECT 2�
[<- ReadyForQuery] Z���I
```


fredzqm:
Not in a rush. @joehan 

Do you remember what edge cases that necessitated the ExtendQueryPatch before?

Haven't seen any issues during my testing. Wanna double check if the original issues are gone when clean them up.

joehan:
I think the 2 cases that were causing problems were explicit transactions and errors. If its working ok in both cases, I think this is reasonable

src/emulator/dataconnect/pgliteServer.ts
```diff
@@ -75,7 +75,7 @@ export class PostgresServer {
         },
       });
 
-      const extendedQueryPatch: PGliteExtendedQueryPatch = new PGliteExtendedQueryPatch(connection);
+      const extendedQueryPatch: PGlitePatch = new PGlitePatch(connection);
```
joehan: Should we rename this now that it doesn't mess with extended queries anymore? Maybe something like 'protocolLoggerPatch'?

fredzqm: `PGliteLoggerPatch`?

fredzqm:
> I think the 2 cases that were causing problems were explicit transactions and errors. If its working ok in both cases, I think this is reasonable

Sounds good. Those were ones I was trying to fix as well. Tested

PR #9757, opened 2026-01-17
Adding skill evaluation to agent-evals

joehan:
### Description
Starting on adding evals for the agent skills. So far, I'm just focusing on the frontmatter and skill activation.

Changes:
- Added a 'skills' argument that allows us to configure what skills are available to the runner
- Added a 'enableMCP' argument that turns on or off the MCP server in tests (so we can evaluate skills on their own)
- Added `expectSkillActivated', which checks telemtry logs for the 'activate_skill' tool
- Added some simple activation tests for 'firestore_basics' and 'firebase_basics' that check if the skill is activated with a variety of prompts
- Added a 'skill choice' test, that makes all skills available and then checks that the 'right' skill is chosen for a variety of prompts


scripts/agent-evals/src/tests/skills/skill-choice.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+const FIRESTORE_BASICS_PATH = path.resolve(process.cwd(), "../../skills/firestore-basics");
+
+const ALL_SKILLS = [FIREBASE_BASICS_PATH, FIRESTORE_BASICS_PATH];
+
+describe("Skill Choice: Selection Logic", function (this: Mocha.Suite) {
+  this.timeout(90000); // Increased timeout for multiple skill loading
+
+  const testCases: {
+    prompt: string,
+    expectedSkill: string, // The skill that SHOULD be activated
```
MichaelDoyle: Do we ever expect multiple skills to active? Say firebase and firestore basics together? why or why not?

scripts/agent-evals/src/tests/skills/firebase-basics.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+
+      await run.type(tc.prompt);
+
+      if (tc.expectSkillEnabled) {
+        await run.expectSkillActivated("firebase-basics");
+      } else {
+        await run.dont.expectSkillActivated("firebase-basics");
+      }
+    });
+  }
```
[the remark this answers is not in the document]
MichaelDoyle: This looks resolved?

scripts/agent-evals/src/tests/skills/firestore-basics.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+      await run.type(tc.prompt);
+
+      if (tc.expectSkillEnabled) {
+        await run.expectSkillActivated("firestore-basics");
+      } else {
+        await run.dont.expectSkillActivated("firestore-basics");
+      }
+    });
+  }
```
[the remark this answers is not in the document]
MichaelDoyle: This looks resolved?

scripts/agent-evals/src/tests/skills/skill-choice.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+import * as path from "path";
+import { startAgentTest, AgentTestRunner } from "../../runner/index.js";
+import "../../helpers/hooks.js";
+
+const FIREBASE_BASICS_PATH = path.resolve(process.cwd(), "../../skills/firebase-basics");
+const FIRESTORE_BASICS_PATH = path.resolve(process.cwd(), "../../skills/firestore-basics");
+
+const ALL_SKILLS = [FIREBASE_BASICS_PATH, FIRESTORE_BASICS_PATH];
+
+describe("Skill Choice: Selection Logic", function (this: Mocha.Suite) {
```
MichaelDoyle: What does this test that the other test cases do not? If the other cases had a list of "not expected skills" would the coverage be complete?

joehan: Right now, its definitely a bit redundant. I think it will be more useful once we have a bunch of skills, but for now I'm gonna remove it so that we don't bloat these  evals before they are useful.

scripts/agent-evals/src/tests/skills/skill-choice.spec.ts
joehan: Good point - it is be totally reasonable to activate multiple skills and we should expect it. In some cases (ie 'Build me an app that stores user data in Firestore' probably needs auth-basics and firestore-basics)  I'm gonna remove this eval for now, but if/when we add it back, we'll refactor this to take the multiple skills cases into account

PR #9768, opened 2026-01-21
Fix compilation errors with secondary schemas in VSCode.

rosalyntan:
### Description

This PR passes in secondary schema files to the language server, in addition to main schema files, allowing fields defined in secondary schema files to be used in connectors without showing a compilation error in VSCode.

### DIFF
--- a/firebase-vscode/CHANGELOG.md
+++ b/firebase-vscode/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## NEXT
 
 - Render GraphQL error debug details and error code in the execution panel nicely. (#9769)
-- [Fixed] Language server now recognizes multiple schemas. (#9768)
+- Language server now recognizes multiple schemas. (#9768)
 
 ## 2.0.1
 


src/dataconnect/types.ts
```diff
@@ -253,6 +253,14 @@ export function mainSchemaYaml(dataconnectYaml: DataConnectYaml): SchemaYaml {
   return mainSch;
 }
 
+/** Returns the secondary schema YAMLs for a Data Connect YAML */
+export function secondarySchemaYamls(dataconnectYaml: DataConnectYaml): SchemaYaml[] {
+  if (dataconnectYaml.schema) {
```
hlshen, line 258: to clarify, there's a oneof field called "schema" or "schemas"?

rosalyntan: Yes, that's correct -- all new dataconnect.yaml files generated once we enable support for custom resolvers will have the `schemas` field instead of `schema`, but we still support `schema` for backwards compatibility.

PR #9794, opened 2026-01-23
Initialize a new function as part of `init dataconnect:resolver`

rosalyntan:
### Description

In `firebase init dataconnect:resolver`, also initialize a new functions codebase with sample code demonstrating the use of the `onGraphRequest` method.

**Note:** As part of this change, I've also refactored the legacy `doSetup` method for the `firebase init functions` flow into separate `askQuestions` and `actuate` steps, which aligns with other products' `init` flows. 

Future work: If there are existing Cloud Run functions already deployed, allow the user to select a function URL to be associated with their custom resolver schema and bypass initializing a new functions codebase.

### Scenarios Tested

Typescript (with eslint): https://paste.googleplex.com/4671059091652608
Typescript (without eslint): https://paste.googleplex.com/4716839617822720

Javascript: (with eslint): https://paste.googleplex.com/6309290279305216
Javascript (without eslint): https://paste.googleplex.com/5223639467294720

Regression testing (`firebase init functions`): http://paste.googleplex.com/4665756417654784

### DIFF
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -766,7 +766,7 @@ export function compareFunctions(
 
 /**
  * Returns the deterministic Cloud Run URI for a given HTTPS function and project number if available based on the DNS segment length.
- * If the function name is too long to guarantee a deterministic URI, this method returns the non-deterministic URI from the backend instead.
+ * If the function name is too long to have a deterministic URI, this method returns the non-deterministic URI from the backend instead.
  * See https://docs.cloud.google.com/run/docs/triggering/https-request#deterministic for more details.
  */
 export function maybeDeterministicCloudRunUri(httpsFunc: Endpoint, projectNumber: string): string {
@@ -774,8 +774,8 @@ export function maybeDeterministicCloudRunUri(httpsFunc: Endpoint, projectNumber
   const dnsSegment = `${serviceName}-${projectNumber}`;
   // TODO: Add deploy-time validation to prevent service names that would exceed this.
   if (dnsSegment.length > 63) {
-    logger.debug(
-      `Function name ${httpsFunc.id} is too long to guarantee a deterministic Cloud Run URI. Printing the non-deterministic URI instead.`,
+    logger.info(
+      `Function name ${httpsFunc.id} is too long to have a deterministic Cloud Run URI. Printing the non-deterministic URI instead.`,
     );
     return httpsFunc.uri!;
   }


src/deploy/functions/release/index.ts
```diff
@@ -150,7 +150,7 @@ export function printTriggerUrls(results: backend.Backend, projectNumber: string
       logger.info(
         clc.bold("Function URL"),
         `(${getFunctionLabel(httpsFunc)}):`,
-        `https://${httpsFunc.id.toLowerCase()}-${projectNumber}.${httpsFunc.region}.run.app`,
+        `https://${httpsFunc.id.toLowerCase().replaceAll("_", "-")}-${projectNumber}.${httpsFunc.region}.run.app`,
```
fredzqm: nit: It gets a bit long, maybe extract a longer variable to highlight the run URL format.

inlined:
Can we set up a short meeting to discuss what you're doing? I have some thoughts and questions:

1. Is it your desire that the canonical style of resolvers is to have a separate codebase, or is this specifically just a new template you can init with?
2. If the latter, I'm wondering if I should pick back up a project I had on my backburner for 2025 before I left. I was going to generalize the normal functions init so that it could have a dialog asking you what kind of project you would init. We can have "hello world", "AI backend", "DataConnect resolver" etc. The plan of record was to rewrite init genkit to short circuit some of the new logic in that streamlined workflow. Would that work for you too? (I would assume you don't want to wait until I could create such a utility)

src/deploy/functions/release/index.ts
rosalyntan: Done.

src/deploy/functions/release/index.ts
```diff
@@ -143,14 +143,15 @@ export function printTriggerUrls(results: backend.Backend, projectNumber: string
       continue;
     }
     if (backend.isDataConnectGraphqlTriggered(httpsFunc)) {
+      const funcId = httpsFunc.id.toLowerCase().replaceAll("_", "-");
```
inlined: nit: maybe add deterministicCloudRunUrl to backend.ts as a helper here?

src/deploy/functions/release/index.ts
inlined, line 152: Alternatively we can just make this URI the correct URI in fabricator.ts? That might be better than here because (A) it's testable and (B) we will never accidentally use the wrong URL.

rosalyntan: Do you mean by getting the result returned from the backend and choosing the deterministic URI? From looking at the [Cloud Functions API](https://docs.cloud.google.com/functions/docs/reference/rest/v2/projects.locations.functions#Function), it looks like it only returns one URI in the response. I didn't see how to get it to return the deterministic URI instead of the non-deterministic one (and Duckie is also telling me the deterministic URI must be constructed and is not returned from the API), but if I'm missing something happy to make the update!

src/deploy/functions/release/index.ts
rosalyntan: Done.

PR #9798, opened 2026-01-23
Add a Firebase skill for migrating/importing the user's project for local development

falahat:


falahat:
We submitted these improvements into the existing firebase skills rather than an entirely new skill

Issue #9762, opened 2026-01-20
Dataconnect MISSING_JOIN_TABLE_KEY


federicobaiocco:
### [REQUIRED] Environment info

**firebase-tools:** 14.27.0

**Platform:** macOS

### [REQUIRED] Test case

Trying to create a schema like this one:
```

type Item @table(key: ["id"]) {
  id: String! @default(expr: "uuidV4()")
  name: String!
}

type Person @table(key: ["id"]) @unique(fields: ["organization", "kind", "displayName"]) {
  id: String! @default(expr: "uuidV4()")
  organization: String!
  kind: String!
  displayName: String!
}

type ItemPersonRole
  @table(key: ["person", "item", "role"]) {
  person: Person!
  item: Item!
  role: String!
  orderIndex: Int
}
```

Does not work. We get the error:
`MISSING_JOIN_TABLE_KEY: On ItemPersonRole: This table looks like a join table, but it is missing the composite primary key definition.Firebase`

This does not work either
```

type ItemPersonRole
  @table(key: ["id"]) @unique(fields:  ["person", "item", "role"]) {
  id: String! @default(expr: "uuidV4()")
  person: Person!
  item: Item!
  role: String!
  orderIndex: Int
}
```

However, we can do it if we execute SQL script directly in SQL Studio to modify the primary key. 

### [REQUIRED] Steps to reproduce

Create tables defined in the example Schema

### [REQUIRED] Expected behavior

I am able to create ItemPersonRole using 3 columns as primary key or unique constraint. 

### [REQUIRED] Actual behavior

I get the error: `MISSING_JOIN_TABLE_KEY: On ItemPersonRole: This table looks like a join table, but it is missing the composite primary key definition.Firebase`

aalej:
Thanks @federicobaiocco for the detailed report! From what can tell, the `This table looks like a join table, but it is missing the composite primary key` may be a linter warning since I'm able to still run mutations to insert an `itemPersonRole` data. Queries to get `itemPersonRoles` also work.

Let me verify this with our engineering team to see it this is an issue with the linter or if it's something else.

sharing mcve here - https://github.com/aalej/issues-9762

fredzqm:
Thank you @federicobaiocco for reporting this issue!

Feel free to ignore the MISSING_JOIN_TABLE_KEY message. It is a linter warning. Doesn't impact using this schema.

[...]

Issue #9630, opened 2025-12-17
[Data Connect] `@view` with two FK fields triggers `MISSING_JOIN_TABLE_KEY` error with no resolution path


samtsangbiz:
### [REQUIRED] Environment info

**firebase-tools:** 13.x (latest)

**Platform:** macOS

**Data Connect Emulator:** 2.15.1

### [REQUIRED] Test case

Create a `@view` type with two foreign key reference fields:

**Schema (schema.gql):**
```graphql
type User @table {
  id: UUID! @default(expr: "uuidV4()")
  name: String!
}

type Product @table {
  id: UUID! @default(expr: "uuidV4()")
  name: String!
  price: Float
}

type Purchase @table {
  user: User!
  product: Product!
  quantity: Int!
  purchaseDate: Timestamp!
}

# This view triggers the error
type UserProductStats @view(sql: """
  SELECT
    user_id,
    product_id,
    COUNT(*) as purchase_count,
    SUM(quantity) as total_quantity
  FROM purchase
  GROUP BY user_id, product_id
""") {
  user: User
  product: Product
  purchaseCount: Int
  totalQuantity: Int
}
```

### [REQUIRED] Steps to reproduce

1. Create the schema above with two `@table` types and one `@view` type
2. The `@view` contains two FK reference fields (`user: User` and `product: Product`)
3. Run `firebase emulators:start --only dataconnect` or `firebase deploy --only dataconnect`

### [REQUIRED] Expected behavior

The `@view` should compile successfully because:
- Views are read-only SQL projections, not actual database tables
- PostgreSQL views don't have primary keys
- The `@view` is an aggregation view, not a join table

### [REQUIRED] Actual behavior

Data Connect returns the error:
```
MISSING_JOIN_TABLE_KEY: This table looks like a join table, but it is missing the composite primary key definition
```

Attempting to add `key: ["user", "product"]` to the `@view` directive:
```graphql
type UserProductStats @view(sql: "...", key: ["user", "product"]) {
  ...
}
```

Results in:
```
Undefined argument key for directive view
```

**Root Cause:** Data Connect's join table detection scans all types (both `@table` and `@view`) for two FK reference fields. When found, it expects a composite `key` definition. However:
- `@table` supports `key: [...]` parameter ✓
- `@view` does NOT support `key` parameter ✗

This creates an impossible situation where `@view` types with two FK fields cannot be defined.

### Workaround

Use plain scalar fields instead of FK references:

```graphql
[...]

PR #9805, opened 2026-01-25
[VS Code] Surface GQL error at the right diagnosis level

fredzqm:
Use `warningLevel` to determine whether to surface GQL error as Diagnosis.ERROR/WARNING/INFO.

So far two customers have confused `MISSING_JOIN_TABLE_KEY` as blocking error rather than non-blocking warning.

https://github.com/firebase/firebase-tools/issues/9762 https://github.com/firebase/firebase-tools/issues/9630

- FDC compiler surface many useful linter errors, many of them at `LOG_ONLY` level.
- Show INSECURE in VS Code if emulator emit those.


fredzqm:
@rosalyntan Small PR, which you should be familiar

Wanted to surface GQL errors based on warningLevel for while~

PR #9834, opened 2026-01-29
Updating App Hosting skills to recommend deploy from source

joehan:
### Description
Coming back to the app hosting skill to recommend the 'firebase.json' + 'firebase deploy' flow for most users. Also, cleaning up some excess text.

### DIFF
--- a/skills/firebase-app-hosting-basics/SKILL.md
+++ b/skills/firebase-app-hosting-basics/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: firebase-app-hosting-basics
-description: Deploy and manage web apps with Firebase App Hosting. Use this skill when deploying Next.js/Angular apps with backends..
+description: Deploy and manage web apps with Firebase App Hosting. Use this skill when deploying Next.js/Angular apps with backends.
 ---
 
 # App Hosting Basics
@@ -20,16 +20,16 @@ This skill enables the agent to deploy and manage modern, full-stack web applica
 - You need Server-Side Rendering (SSR) or ISR.
 - You want an automated "git push to deploy" workflow with zero configuration.
 
-## Deploying  to App Hosting
+## Deploying to App Hosting
 
-### Deploying via CLI
+### Deploy from Source
 
 This is the recommended flow for most users. 
 1. Configure `firebase.json` with an `apphosting` block.
     ```json
     {
       "apphosting": {
-        "backendId": my-app-id",
+        "backendId": "my-app-id",
         "rootDir": "/",
         "ignore": [
           "node_modules",
@@ -48,6 +48,7 @@ This is the recommended flow for most users.
 ### Automated deployment via GitHub (CI/CD)
 
 Alternatively, set up a backend connected to a GitHub repository for automated deployments "git push" deployments.
+This is only recommended for more advanced users, and is not required to use App Hosting.
 See [CLI Commands](references/cli_commands.md) for more information on how to set this up using CLI commands.
 
 ## Emulation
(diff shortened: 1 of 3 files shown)


skills/firebase-app-hosting-basics/references/cli_commands.md
```diff
@@ -4,28 +4,30 @@ The Firebase CLI provides a comprehensive suite of commands to manage App Hostin
...
 ## Backend Management
 ### `firebase apphosting:backends:create`
-- **Purpose**: Creates a new App Hosting backend.
-- **Usage**: `firebase apphosting:backends:create`
-- **Interactive Prompts**:
-  - Select location (e.g., `us-central1`).
-  - Connect/Select GitHub repository.
-  - Set root directory (if monorepo).
-  - Set live branch (e.g., `main`).
+- **Purpose**: Creates a new App Hosting backend. Use this when setting up automated deployments via GitHub.
```
falahat, line 18: non-blocking, you could add "if you want to deploy from local code use firebase init". Sometimes, I found the agent to think that you could *only* deploy via GitHub triggers and wasn't aware of local source deployment

skills/firebase-app-hosting-basics/SKILL.md
```diff
@@ -21,40 +20,36 @@ This skill enables the agent to deploy and manage modern, full-stack web applica
...
-
-## Resources
-- [CLI Commands](references/cli_commands.md)
-- [Configuration](references/configuration.md)
-- [Emulation](references/emulation.md)
+## Deploying  to App Hosting
+
+### Deploying via CLI
+
+This is the recommended flow for most users.
```
falahat, line 27: is this the recommended flow (to edit the firebase.json file manually) or is it better to have the agent use the CLI / tools like a human would?

In my experience, the agent was able to use the CLI commands (init, deploy) without manually modifying this file. I might be missing something however

skills/firebase-app-hosting-basics/SKILL.md
```diff
@@ -21,40 +20,36 @@ This skill enables the agent to deploy and manage modern, full-stack web applica
...
+          "functions"
+        ]
+      }
+    }
+    ```
+2. Create or edit `apphosting.yaml`- see [Configuration](references/configuration.md) for more information on how to do so.
+3. If the app needs safe access to sensitive keys, use `firebase apphosting:secrets` commands to set and grant access to secrets.
+4. Run `firebase deploy` when you are ready to deploy.
+
+### Automated deployment via GitHub (CI/CD)
```
falahat, line 48: It might be worth adding a section for "How to Deploy from Source" or adding a line that GitHub deployments are not the *only* way. The most common source of frustration was when the agent either:

1. Thought that GitHub connections were required and it used backend:create instead of init
2. Thought that GitHub connections were *required* and didn't know it could deploy from local code

skills/firebase-app-hosting-basics/SKILL.md
joehan: I've found that some agents struggle to use interactive commands like init (particularly for cases when they have non text inputs like selecting a resource from a list). Manually writing this file has _generally_ been smoother and quicker in my experience.

skills/firebase-app-hosting-basics/references/cli_commands.md
joehan: Took a a pass through and emphasized these points more heavily.

skills/firebase-app-hosting-basics/SKILL.md
joehan: Took a a pass through and emphasized this more heavily.

PR #9839, opened 2026-01-29
test: add unit tests for apphosting:backends:create

joehan:
### Description
Working on a new AI workflow for generating tests to improve out coverage. This one seems pretty useful, and is validating some relevant error cases.


src/commands/apphosting-backends-create.ts
```diff
@@ -28,7 +28,7 @@ export const command = new Command("apphosting:backends:create")
     "specify the primary region for the backend. Required with --non-interactive.",
   )
   .option("--root-dir <rootDir>", "specify the root directory for the backend.")
-  .before(requireAuth)
+  .before((options: Options) => requireAuth(options))
```
aalej, line 31: This might not be intended, we do `.before(requireAuth)` on all of our commands

aalej (changes requested):

PR #9843, opened 2026-01-29
Removing webframeworks questions from init hosting

joehan:
### Description
Removing the web frameworks set up flow from `firebase init hosting`, and adding a unit test while we are here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,3 @@
+- Add support for VPC direct connect in GCF 2nd gen (#10033)
+- Added `--only` flag for `emulators:export` (#4033)
+- Added support for custom PostgreSQL schema names in Data Connect. (#9271)
--- a/firebase-vscode/CHANGELOG.md
+++ b/firebase-vscode/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## NEXT
 
+## 2.2.0
+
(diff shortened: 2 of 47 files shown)


jpreid7396:
Awesome - just want to clarify:
What happens when a customer runs `firebase init hosting` in a directory containing a Next.js app (or any web app that uses SSR)? We don't currently have framework detection iiuc, so they will initialize hosting and then the deployment will fail? 

To improve the experience, we could:

Option 1: Keep the web frameworks experiment on, but change the behavior so that if the user answers yes to

```
=== Hosting Setup
✔ Detected an existing Next.js codebase in the current directory, do you want to use
 this?
```

We tell them to run `firebase init apphosting` and terminate the flow

OR

Option 2: Kill the experiment and add a big warning message in the Hosting Setup flow: 

```
=== Hosting Setup
Firebase Hosting only supports static web apps. 
If your app needs a server for server-side rendering, initialize the new Firebase App Hosting
instead with firebase init apphosting
```

Thoughts?

annajowang:
We can do Option (1)  while also getting rid of the webframeworks experiment. So this framework detection can just run every time (confirmed with james daniels, no reason we can't just run it every time).

@leoortizz can take this and make the changes on top's of this PR

annajowang:
*Option (1) recommend app hosting only if nodejs frameworks were detected

Issue #9855, opened 2026-02-01
Deploying Python Cloud Functions fails since v15.5.0


vipshek:
### Environment info

**firebase-tools:** v15.5.0

**Platform:** Ubuntu Docker image

### Steps to reproduce

Run `firebase deploy --only functions` in a project containing Python Functions.

### Expected behavior

Firebase deploy runs and inspects Python source to determine what to deploy, logging lines like this:

```
functions: Loading and analyzing source code for codebase python to determine what to deploy
```

### Actual behavior

Firebase deploy fails, in my case with a variety of strange errors.

In our case we have a script to validate that Python 3.12 is present before running Firebase deploy, and we now see that Python is no longer present in the Docker image at all, preventing the code discovery features

### Root cause

It appears Python was removed from the firebase-cli Docker image in #9767, which seems to be the root cause here. I'll make a PR to add this back, and include a comment for why the Python dep is necessary.

### Other issues

Maybe related: #9853

aalej:
Thanks for the report and opening opening up a PR @vipshek! I'll raise this to our engineering team so they can take a look. 

I don't think that issue #9853 is related, the error from that issue seems to be a Node.js module import(`firebase-functions/v2/https`).

PR #9856, opened 2026-02-01
Add back Python to Firebase Docker image

vipshek:
### Description

Bugfix for #9855 - just adding back Python dependencies to the Docker image to fix deploying Python Firebase functions. Targeted revert of a change introduced in #9767.


joehan:
Closing in favor of #9860 so i could add a changelog and address the GCA comments. Thanks for reporting this issue and proposing a fix!

PR #9860, opened 2026-02-02
Add Python back into our Docker image

### LINKED ISSUE #9855: Deploying Python Cloud Functions fails since v15.5.0

joehan:
### Description
Fixes #9855 , and reorders the command to avoid layers with outdated deps.

### DIFF
--- a/scripts/publish/firebase-docker-image/package-lock.json
+++ b/scripts/publish/firebase-docker-image/package-lock.json
@@ -7869,9 +7869,9 @@
       }
     },
     "node_modules/tar": {
-      "version": "7.5.3",
-      "resolved": "https://registry.npmjs.org/tar/-/tar-7.5.3.tgz",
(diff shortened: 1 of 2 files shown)


aalej (approved):
LGTM!

PR #9881, opened 2026-02-05
Fix quota project for billing api calls

bkendall:
### Description

Adds the quota project header to the billing calls.

### Scenarios Tested

Tested this logic in a command and verified the header was added and the result came back.


joehan (approved):
LGTM, test failure was a flake. Will send a pr to disable that test later

PR #9887, opened 2026-02-06
redirect to init apphosting for webframeworks on init hosting

leoortizz:
When running `firebase init hosting`, run framework detection. If a webframework with a backend is found, trigger `firebase init apphosting` instead.

Based on the following comments from #9843:
- https://github.com/firebase/firebase-tools/pull/9843#issuecomment-3837391280
- https://github.com/firebase/firebase-tools/pull/9843#issuecomment-3855939949
- https://github.com/firebase/firebase-tools/pull/9843#issuecomment-3856095295

### Scenarios Tested

`firebase init hosting` in a webframework project

### Sample Commands

`firebase init hosting`


src/init/features/hosting/index.ts
```diff
@@ -58,6 +62,15 @@ export async function askQuestions(setup: Setup, config: Config, options: Option
     }
   }
 
+  const discoveredFramework = await discoverPromise;
+  if (discoveredFramework && discoveredFramework.mayWantBackend) {
+    const frameworkName =
+      WebFrameworks[discoveredFramework.framework]?.name ?? discoveredFramework.framework;
+    throw new FirebaseError(
+      `Detected a ${frameworkName} codebase. Use ${clc.bold("firebase init apphosting")} instead.`,
```
joehan: Alternatively - why not just print a message then kick off the `init apphosting` flow?


I'd also push this code block higher up in the file before we ask all the questions - it seems annoying to answer a bunch of questions that turn out to be irrelevant

leoortizz: makes sense, thanks. Just pushed this change

src/init/features/hosting/index.ts
```diff
@@ -27,6 +29,19 @@ export interface RequiredInfo {
...
 export async function askQuestions(setup: Setup, config: Config, options: Options): Promise<void> {
+  const discoveredFramework = await discover(config.projectDir, false);
+  if (discoveredFramework && discoveredFramework.mayWantBackend) {
+    const frameworkName = discoveredFramework.framework;
+    logger.info();
+    logger.info(
+      `Detected a ${frameworkName} codebase. Setting up ${clc.bold("App Hosting")} instead.`,
+    );
+    setup.featureInfo ||= {};
+    setup.featureInfo.hosting = { redirectToAppHosting: true };
```
annajowang: app hosting does not support flutter and i believe one of the web frameworks detected is flutter

leoortizz: Thank you for bringing this up. Flutter won't actually trigger this because its `mayWantBackend` is always `false`, i.e. it's always static.

However, frameworks other than Angular and Next.js that (I think) are not supported by App Hosting will have this issue:
- Express.js: always
- Nuxt 2:  always
- Nuxt 3: when SSR is enabled
- Astro: when output is not "static"
- Vite: when app type is not "spa"

WDYT about:
1. only run `init apphosting` automatically for Next.js and Angular projects
2. error out for non supported frameworks with a backend?

annajowang: (1) sounds good
(2) We have some support for all of those: https://firebase.blog/posts/2025/06/app-hosting-frameworks/

but i'm not sure how advanced the support is and at least some of them seem to require some minimal set up from the user.
So yes, agree to exit, also add a message linking to the blog post and then telling user to `init apphosting`

jpreid7396: Hey - thanks for this! I think in all the cases you listed in (2), App Hosting should just work. Can we try testing App Hosting for these apps, and if it works we initialize App Hosting in these scenarios?

src/init/features/hosting/index.ts
jpreid7396: Let's start by exiting with a message that says to run `firebase init apphosting` instead of automatically triggering the App Hosting init flow, so that devs are more aware of what's happening. cc @jhuleatt

src/init/features/hosting/index.ts
```diff
@@ -27,6 +29,19 @@ export interface RequiredInfo {
 // TODO: come up with a better way to type this
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export async function askQuestions(setup: Setup, config: Config, options: Options): Promise<void> {
+  const discoveredFramework = await discover(config.projectDir, false);
+  if (discoveredFramework && discoveredFramework.mayWantBackend) {
+    const frameworkName = discoveredFramework.framework;
+    logger.info();
+    logger.info(
+      `Detected a ${frameworkName} codebase. Setting up ${clc.bold("App Hosting")} instead.`,
+    );
```
jhuleatt: @annajowang @jpreid7396 We really need to get consent before switching to App Hosting. There should be a bit of an explainer and a chance for the developer to explicitly choose to use App Hosting instead. It might end up being an "offer they can't refuse" if we just say "Hosting is for static sites, sorry, we can't deploy your Next.js app", but at least there's a choice. They may have to turn on billing as part of the flow too.

Quick language brainstorm with @egilmorez:

> Detected a Next.js codebase with SSR features. We can't guarantee that this site will work on Firebase Hosting, which is only for static sites. Another product, Firebase **App** Hosting, was designed for SSR web apps. Would you like to use App Hosting instead? Learn more here: https://firebase.google.com/docs/app-hosting/product-comparison#hostings (Y / n)

jpreid7396: Love it! Thank you

src/init/features/hosting/index.ts
annajowang: We still need to split things into two categories

(1) angular, nextjs, nuxt, expressjs, sveltekit
These work OOB. (nuxt/svelte confirmed by Leo)
Also
https://nuxt.com/deploy/firebase#firebase-app-hosting-recommended

I think for these we should just confirm with user to run `firebase init apphosting` using jeff's proposed language in the other comment.

(2) vite, astro
We have an astro adapter but it requires set up. And Leo found that Vite does not work OOB.
We should recommend the user look into app hosting and also link to jeff's blogpost before exiting. With similar language:
```
Detected a {framework} codebase with SSR features. We can't guarantee that this site will work on Firebase Hosting, which is only for static sites. Another product, Firebase App Hosting, was designed for SSR web apps. Learn about App Hosting here: https://firebase.google.com/docs/app-hosting/product-comparison#hostings
Learn how to deploy frameworks with App Hosting here: https://firebase.blog/posts/2025/06/app-hosting-frameworks/
```

src/init/features/hosting/index.ts
```diff
@@ -27,6 +29,28 @@ export interface RequiredInfo {
...
 export async function askQuestions(setup: Setup, config: Config, options: Options): Promise<void> {
+  const discoveredFramework = await discover(config.projectDir, false);
+  if (discoveredFramework && discoveredFramework.mayWantBackend) {
+    const frameworkName =
+      WebFrameworks[discoveredFramework.framework]?.name ?? discoveredFramework.framework;
+    logger.info();
+    const useAppHosting = await confirm({
+      message:
+        `Detected a ${frameworkName} codebase with SSR features. We can't guarantee that ` +
+        `this site will work on Firebase Hosting, which is only for static sites. Another ` +
```
egilmorez: I'm thinking "optimized for static sites" makes more sense when viewed alongside the integrations we document in "Serve dynamic content and host microservices using Firebase Hosting" -- https://firebase.google.com/docs/hosting/serverless-overview.

src/init/features/hosting/index.ts
leoortizz: @annajowang just pushed this change

src/init/features/hosting/index.ts
```diff
@@ -27,6 +29,46 @@ export interface RequiredInfo {
...
+      WebFrameworks[discoveredFramework.framework]?.name ?? discoveredFramework.framework;
+
+    switch (discoveredFramework.framework) {
+      case "next":
+      case "angular":
+      case "nuxt":
+      case "nuxt2":
+      case "express":
+      case "svelekit":
+      case "sveltekit":
```
annajowang, line 45: can we just have "sveltekit"

and fix this typo?
https://github.com/firebase/firebase-tools/blob/e4937afbcac1b450abd837d3004a989d3257413f/src/frameworks/frameworks.ts#L27

or will that somehow have some annoying downstream/ breaking impact

src/init/features/hosting/index.ts
annajowang: with comments below, final state here is that we're:
(1) for certain frameworks that work OOB, we're asking users for consent to run `firebase init apphosting` instead
(2) for other frameworks that don't work / require some setup (astro and vite) we will show a similar message + a link to jeff's blog https://firebase.blog/posts/2025/06/app-hosting-frameworks/ which shows how users can make these work.

src/init/features/hosting/index.ts
leoortizz: yes it might cause issues elsewhere... cc @jamesdaniels to confirm. I remember some metrics used `svelekit`

annajowang: ok, I'll ask around if i have time.

If james is too busy to respond before we want to merge, could you just add a comment in the code that says something like "svelekit should not be removed unless it's confirmed to not be used in metrics or not cause breakages."

src/init/features/hosting/index.ts
```diff
@@ -27,6 +29,46 @@ export interface RequiredInfo {
...
+            `https://firebase.google.com/docs/app-hosting/product-comparison#hostings`,
+          default: true,
+        });
+        if (useAppHosting) {
+          setup.featureInfo ||= {};
+          setup.featureInfo.hosting = { redirectToAppHosting: true };
+          setup.features?.unshift("apphosting");
+          return;
+        }
+        break;
```
annajowang, line 62: If the user chooses "n" we should exit.  We don't want more users using this experiment.

I believe this is currently just continuing with the `firebase init hosting` flow?

src/init/features/hosting/index.ts
annajowang: If this is just about metrics only, i found the metrics. It wouldn't be breaking anything if the typo was fixed the svelekit line would just stop and turn into the sveltekit line.

That said, lets just go with a comment here and I'll find time to confirm there's nothing else.

src/init/features/hosting/index.ts
```diff
@@ -27,6 +29,46 @@ export interface RequiredInfo {
 // TODO: come up with a better way to type this
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export async function askQuestions(setup: Setup, config: Config, options: Options): Promise<void> {
+  const discoveredFramework = await discover(config.projectDir, false);
+  if (discoveredFramework && discoveredFramework.mayWantBackend) {
+    const frameworkName =
+      WebFrameworks[discoveredFramework.framework]?.name ?? discoveredFramework.framework;
+
+    switch (discoveredFramework.framework) {
```
annajowang, line 37: per @jhuleatt 
"How smart is discovery? If someone happens to be doing static-only Next.js (unlikely as that is), can we detect that and allow a Hosting deploy?"

src/init/features/hosting/index.ts
egilmorez: We should keep in mind that this would also prevent users whose apps trip the detection criteria from initializing Hosting (classic) as described in our Hosting onboarding docs (internally staged version): https://firebase.devsite.corp.google.com/docs/hosting/quickstart?db=egilmore#initialize

Toward the end of that section we're saying "You can proceed with App Hosting setup or continue with Hosting (classic) for static-only deploys," so the behavior and docs are in line; but is this fully desired behavior, to stop this subset of Hosting (classic) adopters as well as the CLI experiment?

annajowang: You're right eric, thanks for catching. The goal is to block new Web Framework participants (which I believe the current code is doing and my comment was incorrect).

If we DID just exit here and block hosting onboarding ideally it would be NOOP because ideally, what is detected as a would-be web framework participant would also NOT be a plain (non-web framework) hosting user. 

but that's not true if the detection is imperfect.

So, i see no harm in keeping this code as is. wrong detection or no wrong detection, we're already telling users that the app might not work with hosting. They can continue with plain hosting and see what happens.

So no change here....but then I think we WOULD need to to change the code below. For any other ssr apps (that don't work out of the box with app hosting), to be consistent we should let users proceed with hosting if they wish.

src/init/features/hosting/index.ts
leoortizz: discovery is smart for some frameworks, not for others. We're only running this for frameworks that work with SSR in webframeworks(mayWantBackend = true): https://github.com/firebase/firebase-tools/blob/11a5b28f2595acf7bc0c5b9b29cbe29799576fa5/src/init/features/hosting/index.ts#L33-L34

Smart SSR discover (based on framework config):
- Astro — reads Astro config, output !== "static"
- Nuxt (v3+) — reads Nuxt config, checks ssr value
- Vite — reads Vite config, appType !== "spa"
- Lit — Vite discover + lit dep, appType !== "spa"
- React — Vite plugin vite:react-jsx, appType !== "spa"
- Preact — Vite plugin vite:preact-jsx, appType !== "spa"
- Svelte — Vite plugin vite-plugin-svelte, appType !== "spa"
- SvelteKit — Vite discover + @sveltejs/kit dep, appType !== "spa"

Discover hardcoded as SSR:
- Next — always true. SSR depends on build
- Angular — always true, SSR depends on build
- Express — always true, SSR depends on build
- Nuxt2 — always true, SSR depends on build

Always false:
- Flutter

annajowang: Thank you for this breakdown!

annajowang:
I'm getting the following error:
```
.....                                                                                                                                                       
                                                                                                                                                                                                                         
=== Hosting Setup                                                                                                                                                                                                        
                                                                                                                                                                                                                         
✔ Detected a Next.js codebase with SSR features. We can't guarantee that this site will work on Firebase Hosting, which is optimized for static sites. Another product, Firebase App Hosting, was designed for SSR web   
apps. Would you like to use App Hosting instead? Learn more here: https://firebase.google.com/docs/app-hosting/product-comparison#hostings Yes                                                                           
                                                                                                                                                                                                                         
=== App Hosting Setup                                                                                                                                                                                                    
                                                                                                                                                                                                                         
Error: Unexpected error occurred while testing for IAM service account permissions                                                                                                                                       
                                                                                                                                                                                                                         
```

It's happening even with `firebase init apphosting`. 
but no issue when i switch to the main branch.

leoortizz:
> I'm getting the following error:
> 
> ```
> .....                                                                                                                                                       
>                                                                                                                                                                                                                          
> === Hosting Setup                                                                                                                                                                                                        
>                                                                                                                                                                                                                          
> ✔ Detected a Next.js codebase with SSR features. We can't guarantee that this site will work on Firebase Hosting, which is optimized for static sites. Another product, Firebase App Hosting, was designed for SSR web   
> apps. Would you like to use App Hosting instead? Learn more here: https://firebase.google.com/docs/app-hosting/product-comparison#hostings Yes                                                                           
>                                                                                                                                                                                                                          
> === App Hosting Setup                                                                                                                                                                                                    
>                                                                                                                                                                                                                          
> Error: Unexpected error occurred while testing for IAM service account permissions                                                                                                                                       
>                                                                                                                                                                                                                          
> Having trouble? Try firebase [command] --help
> ```
> 
> It's happening even with `firebase init apphosting`. but no issue when i switch to the main branch.

yes that's happening to me too, it seems like there's a fix in main but not in the base branch `jh-remove-wf`. The base branch would need to be rebased

leoortizz:
> > I'm getting the following error:
> > ```
> > .....                                                                                                                                                       
> >                                                                                                                                                                                                                          
> > === Hosting Setup                                                                                                                                                                                                        
> >                                                                                                                                                                                                                          
> > ✔ Detected a Next.js codebase with SSR features. We can't guarantee that this site will work on Firebase Hosting, which is optimized for static sites. Another product, Firebase App Hosting, was designed for SSR web   
> > apps. Would you like to use App Hosting instead? Learn more here: https://firebase.google.com/docs/app-hosting/product-comparison#hostings Yes                                                                           
> >                                                                                                                                                                                                                          
> > === App Hosting Setup                                                                                                                                                                                                    
> >                                                                                                                                                                                                                          
> > Error: Unexpected error occurred while testing for IAM service account permissions                                                                                                                                       
> >                                                                                                                                                                                                                          
> > Having trouble? Try firebase [command] --help
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It's happening even with `firebase init apphosting`. but no issue when i switch to the main branch.
> 
> yes that's happening to me too, it seems like there's a fix in main but not in the base branch `jh-remove-wf`. The base branch would need to be rebased

@annajowang I merged latest from main in the base branch/here. The error should be gone now

src/init/features/hosting/index.ts
leoortizz: @egilmorez @annajowang Thanks for the feedback. I believe I have implemented the changes mentioned, please let me know if I missed anything

PR #9896, opened 2026-02-09
replaces quota override with retry logic

bkendall:
### Description

This API endpoint isn't enabled in every project, so we shouldn't override it.


aalej (approved):
LGTM! Tested on both projects with Cloud Billing API enabled and one with Cloud Billing API not enabled. 

Also, not hitting the quota errors anymore, did around 20 deployments in succession.(probably also due to the quota increase)

PR #9902, opened 2026-02-10
Updating pubsub emulator to 0.8.27, and add an agent skill for the task

joehan:
### Description
Update pubsub emulator to 0.8.27.

Also, adding an agent skill for this task, to make it easy to automate in the future.


aalej:
Did a quick test for the PubSub emulator, unfortunately, it doesn't seems like the new version fixes the memory leak issue.
image

%CPU compared with v0.8.17

Screenshot 2026-02-10 at 9 26 10 PM

PR #9908, opened 2026-02-10
Fixed an issue where get_documents tool couldnt handle subcollections

### LINKED ISSUE #9698: [MCP] firestore_query_collection does not seem to be working for sub collections
## Summary

`firestore_query_collection` does not seem to be working for sub collections 

## Bug Info 

**Affected Tool(s):** Firebase MCP tools, `firestore_query_collection` - https://github.com/firebase/firebase-tools/blob/main/src/mcp/tools/firestore/query_collection.ts
**MCP Client:** Claude 1.0.2339 (1782e2) 2025-12-16T19:35:52.000Z
**Operating System:** macOS Tahoe 26.2

### Steps to Reproduce

- Set up the MCP server and connect Claude Desktop to it
- Start asking Claude about the Firestore database project, it can access most of the data but not sub collections
- Ask it to query a sub collection e.g. "tell about about all of the transactions for each trip"
- Claude gets an error: Collection id "trips/0NK4xpv3AMbbdUq9ypIc/transactions" is invalid because it contains "/"

More detail from Claude:

-----------------------------

**Key part:** The `collection_path` description says:
> "A collection path (e.g. `collectionName/` or `parentCollection/parentDocument/collectionName`)"

This **explicitly states** that subcollection paths like `parentCollection/parentDocument/collectionName` should work!

**But when I try it, I get:**
> Collection id "trips/0NK4xpv3AMbbdUq9ypIc/transactions" is invalid because it contains "/"

-----------------------------

So transactions is a collection within each trip document.

It would be amazing to build an AI application using Firestore and the MCP tools however this seems like a blocker. Unless I'm missing something?

Here is my schema:

Image

joehan:
### Description
Fixes #9698


src/mcp/tools/firestore/query_collection.ts
```diff
@@ -92,11 +92,35 @@ export const query_collection = tool(
...
+      cleanPath = cleanPath.slice(1);
+    }
+    if (cleanPath.endsWith("/")) {
+      cleanPath = cleanPath.slice(0, -1);
+    }
+
+    const parts = cleanPath.split("/");
+    if (parts.length % 2 === 0) {
+      return mcpError(
+        `Invalid collection path: "${collection_path}". Path must point to a collection, not a document.`,
```
chkuang-g: The MCP error message should be clear about the problem and the path forward. What if something like 

```
Invalid collection path: "${collection_path}". Path must point to a collection, not a document. A path like this contains even numbers of parts, ex. `collectionName/` or `parentCollection/parentDocument/collectionName`.
```

src/mcp/tools/firestore/query_collection.ts
```diff
@@ -92,11 +92,35 @@ export const query_collection = tool(
...
+    }
+
+    // Remove leading and trailing slashes
+    let cleanPath = collection_path.trim();
+    if (cleanPath.startsWith("/")) {
+      cleanPath = cleanPath.slice(1);
+    }
+    if (cleanPath.endsWith("/")) {
+      cleanPath = cleanPath.slice(0, -1);
+    }
```
chkuang-g: If the agent frequently adds leading or tailing `/`, it feels like a tool parameter description problem. Perhaps we should update L22 as well?

chkuang-g (changes requested):
all nits.

PR #9921, opened 2026-02-13
Developer Knowledge OneMCP proxy

chkuang-g:
### Description

- Add OneMCP proxy support to Firebase MCP server. Now 
- Bundle [Developer Knowledge MCP](https://developers.google.com/knowledge/mcp) as part of Firebase MCP server

### Scenarios Tested

- Unit tests with `.spec.ts`
- Use `npx -y @modelcontextprotocol/inspector` to ensure `tools/call` and `tools/list` is working.

### Sample Commands

N/A. It is MCP feature

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,5 @@
+- Update the Firebase Data Connect local toolkit to v3.1.4, which includes the following changes: (#9944)
+  - Add support for `@searchable` on varchar fields.
+  - Fix a runtime error when using aggregations with nested reference fields.
+  - Update the Golang dependency version from v1.24.12 to v1.24.13.
 - Added new Developer Knowledge MCP tools
--- a/src/emulator/downloadableEmulatorInfo.json
+++ b/src/emulator/downloadableEmulatorInfo.json
@@ -54,36 +54,36 @@
   },
   "dataconnect": {
     "darwin": {
-      "version": "3.1.3",
-      "expectedSize": 30753632,
-      "expectedChecksum": "db15fbb3d596b056390c08d065500c63",
-      "expectedChecksumSHA256": "f1b98402e9d8497e88a96de6ac80425aee3a61513eccc1fb76b4e387862625ee",
-      "remoteUrl": "https://storage.googleapis.com/firemat-preview-drop/emulator/dataconnect-emulator-macos-amd64-v3.1.3",
-      "downloadPathRelativeToCacheDir": "dataconnect-emulator-3.1.3"
+      "version": "3.1.4",
+      "expectedSize": 30761824,
+      "expectedChecksum": "199b9bd400ab42c549f19254c668e148",
+      "expectedChecksumSHA256": "6f972b594708978b7d42348007de1d1614e6f43b583167e05bf8f4d4eff99173",
+      "remoteUrl": "https://storage.googleapis.com/firemat-preview-drop/emulator/dataconnect-emulator-macos-amd64-v3.1.4",
+      "downloadPathRelativeToCacheDir": "dataconnect-emulator-3.1.4"
     },
     "darwin_arm64": {
-      "version": "3.1.3",
-      "expectedSize": 30203330,
-      "expectedChecksum": "8ac08faf1b797615d91c58f7ea781af0",
-      "expectedChecksumSHA256": "eb00f5bd3c97c17b617a9ce5f5153d6d9f20bc482e4ac091a6361f9ff3bd8841",
-      "remoteUrl": "https://storage.googleapis.com/firemat-preview-drop/emulator/dataconnect-emulator-macos-arm64-v3.1.3",
-      "downloadPathRelativeToCacheDir": "dataconnect-emulator-3.1.3"
+      "version": "3.1.4",
+      "expectedSize": 30236450,
+      "expectedChecksum": "0f59789d0c62ec4a0135d356187fe542",
+      "expectedChecksumSHA256": "59d590c7b60598be4d86155451abc55fbf0f65801ba11c06900accf491977e69",
+      "remoteUrl": "https://storage.googleapis.com/firemat-preview-drop/emulator/dataconnect-emulator-macos-arm64-v3.1.4",
+      "downloadPathRelativeToCacheDir": "dataconnect-emulator-3.1.4"
     },
     "win32": {
-      "version": "3.1.3",
-      "expectedSize": 31262208,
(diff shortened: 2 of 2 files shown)


src/mcp/onemcp/onemcp_server.ts
```diff
@@ -0,0 +1,81 @@
...
+      ctx.projectId
+        ? {
+          headers: {
+            "x-goog-user-project": ctx.projectId,
+          },
+        }
+        : {},
+    );
+    if (res.body.error) {
+      throw new Error(
```
joehan: This should return an MCP error instead of throwing:
```
return MCPError(res.body.error.message)
```

src/mcp/onemcp/onemcp_server.ts
```diff
@@ -0,0 +1,81 @@
...
+   */
+  async fetchRemoteTools(): Promise<ServerTool[]> {
+    const res = await this.listClient.post<any, any>("/mcp", {
+      method: "tools/list",
+      jsonrpc: "2.0",
+      id: 1,
+    });
+
+    const parsed = ListToolsResultSchema.parse(res.body.result);
+    return (parsed.tools as any[]).map((mcpTool) => ({
```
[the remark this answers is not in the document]
joehan: Agreed

src/mcp/onemcp/onemcp_server.ts
```diff
@@ -0,0 +1,81 @@
+import { CallToolResultSchema, ListToolsResultSchema } from "@modelcontextprotocol/sdk/types.js";
+import { Client } from "../../apiv2";
+import { ServerTool } from "../tool";
+import { McpContext, ServerFeature } from "../types";
+
+/**
+ * OneMcpServer encapsulates the logic for interacting with a remote MCP server.
+ */
+export class OneMcpServer {
```
joehan, line 19: Nit: I think its OneMCP, not OneMcp

src/mcp/onemcp/onemcp_server.ts
```diff
@@ -0,0 +1,81 @@
...
+      urlPrefix: this.serverUrl,
+      auth: true,
+    });
+  }
+
+  /**
+   * Fetches tools from the remote MCP server.
+   */
+  async fetchRemoteTools(): Promise<ServerTool[]> {
+    const res = await this.listClient.post<any, any>("/mcp", {
```
joehan: Probably can pass ctx to this function, and put a call to ensure in here, before post, ie: 

```
await ensure(ctx.projectId, this.serverUrl, "MCP", /**silent=**/true
```

src/mcp/util/availability.ts
```diff
@@ -6,6 +6,7 @@ import { isAppTestingAvailable } from "./apptesting/availability";
 const DEFAULT_AVAILABILITY_CHECKS: Record<ServerFeature, (ctx: McpContext) => Promise<boolean>> = {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   core: async (ctx: McpContext): Promise<boolean> => true,
+  developerknowledge: async (ctx: McpContext): Promise<boolean> => true,
```
joehan:
```suggestion
  developerknowledge: async (): Promise<boolean> => true,
```

TS is funny about function signatures - this fulfills the type

src/apiv2.ts
```diff
@@ -237,9 +237,8 @@ export class Client {
   /**
    * Makes a request as specified by the options.
    * By default, this will:
-   *   - use content-type: application/json
-   *   - assume the HTTP GET method
-   *
+   * - use content-type: application/json
+   * - assume the HTTP GET method
```
joehan: Nit: Can remove this file from the PR since there are no real changes.

src/mcp/onemcp/onemcp_server.ts
chkuang-g: Similar to `McpContext`, just try to keep a consistent camelCase here

src/mcp/onemcp/onemcp_server.ts
chkuang-g: Refactored error handling here

CHANGELOG.md
```diff
@@ -0,0 +1,2 @@
+- Added OneMCP proxy support.
```
joehan: Can remove this first line as it is an implementation detail, we just include user facing changes here.

src/mcp/onemcp/onemcp_server.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * Proxies a tool call to the remote MCP server.
+   */
+  private async proxyRemoteToolCall(
+    toolName: string,
+    args: any,
+    ctx: McpContext,
+  ): Promise<CallToolResult> {
+    await ensure(ctx.projectId, this.serverUrl, this.feature, /* silent=*/ true);
```
joehan, line 74: Nit: we probably should stop making this call once we've seen an 'enabled' result -its pretty rare to disable an API after its been enabled, and the error message from the server if you do so is still pretty clear and actionable. That could be a good improvement within ensure itself, so we can do it in a follow up PR.

src/mcp/onemcp/onemcp_server.ts
```diff
@@ -0,0 +1,104 @@
...
+    try {
+      const res = await this.callClient.post<any, any>(
```
joehan: Optional: Probably could improve this typing - the 1st generic is at least partially known (can leave args as any), and the 2nd generic is probably more accurate as Unknown rather than any.

src/mcp/onemcp/onemcp_server.ts
chkuang-g: Maybe a retry pattern like

1. Make API Call anyway
2. If failed with specific error message, ensure and retry again.

This can make sure it is slow only for the first time.

chkuang-g: Added a TODO

src/mcp/tools/index.spec.ts
```diff
@@ -56,3 +59,72 @@ describe("availableTools", () => {
...
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it("should call fetchRemoteTools on servers in ONEMCP_SERVERS", async () => {
+    const mockTool = { mcp: { name: "remote_tool" } };
+    const fetchStub = sandbox
+      .stub(OneMcpServer.prototype, "fetchRemoteTools")
```
joehan: Please fix

PR #9966, opened 2026-02-24
feat: add Dart runtime delegate for emulator

Lyokone:
Adds Dart as a new runtime for Firebase Functions, gated behind the functionsrunapionly experiment flag.
                                                                                                                                
- Dart runtime delegate: Detects Dart projects via pubspec.yaml, discovers functions from functions.yaml generated by           
build_runner, and manages build_runner watch for hot reload via the delegate's watch() method.                                  
- Emulator support: Dart uses a shared-process model (like Python) — one process serves all functions, routing by request path  
instead of FUNCTION_TARGET env var.                                                                                             
- firebase init functions: Scaffolds a new Dart Functions project.                                                              
- Added dart3 runtime with experimental status and isLanguageRuntime() helper for clean runtime checks.                       

Scenarios Tested

- npm run build / npm run lint / npm test — all pass
- npm run generate:json-schema — regenerated with dart3
- Manual emulator testing: HTTPS triggers route correctly, hot reload works, clean shutdown

Sample Commands
```
firebase experiments:enable functionsrunapionly
firebase init functions  # Select "Dart"
```

### DIFF
--- a/src/firebase_studio/migrate.spec.ts
+++ b/src/firebase_studio/migrate.spec.ts
@@ -242,7 +242,7 @@ describe("migrate", () => {
             "add",
             "firebase/agent-skills",
             "-a",
-            "antigravity",
+            "gemini-cli",
             "--skill",
             "*",
             "-y",
@@ -307,6 +307,11 @@ describe("migrate", () => {
           throw Object.assign(new Error("File not found"), { code: "ENOENT" });
         throw new Error(`Unexpected readFile: ${pStr}`);
       });
+
+      commandStub.withArgs("npx").returns(true);
+      confirmStub
+        .withArgs(sinon.match({ message: sinon.match(/Firebase MCP server/) }))
+        .resolves(true);
 
       await migrate(testRoot);
 
@@ -601,6 +606,68 @@ describe("migrate", () => {
       const mcpConfigPath = path.join(mcpConfigDir, "mcp_config.json");
 
       expect(writeFileStub.calledWith(mcpConfigPath, sinon.match(/"dart"/))).to.be.false;
+    });
+
+    it("should prompt user for Firebase MCP server and configure if they confirm", async () => {
+      commandStub.withArgs("npx").returns(true);
+      confirmStub
+        .withArgs(sinon.match({ message: sinon.match(/Firebase MCP server/) }))
+        .resolves(true);
+
+      await migrate(testRoot);
+
+      const mcpConfigDir = path.join(require("os").homedir(), ".gemini", "antigravity");
+      const mcpConfigPath = path.join(mcpConfigDir, "mcp_config.json");
+      expect(writeFileStub.calledWith(mcpConfigPath, sinon.match(/"firebase":/))).to.be.true;
+    });
+
+    it("should prompt user for Firebase MCP server and NOT configure if they decline", async () => {
+      commandStub.withArgs("npx").returns(true);
+      confirmStub
+        .withArgs(sinon.match({ message: sinon.match(/Firebase MCP server/) }))
+        .resolves(false);
+
+      await migrate(testRoot);
+
+      const mcpConfigDir = path.join(require("os").homedir(), ".gemini", "antigravity");
+      const mcpConfigPath = path.join(mcpConfigDir, "mcp_config.json");
+      expect(writeFileStub.calledWith(mcpConfigPath, sinon.match(/"firebase":/))).to.be.false;
+    });
+
+    it("should skip Firebase MCP server if npx is not available", async () => {
+      commandStub.withArgs("npx").returns(false);
+
+      await migrate(testRoot);
+
+      const mcpConfigDir = path.join(require("os").homedir(), ".gemini", "antigravity");
+      const mcpConfigPath = path.join(mcpConfigDir, "mcp_config.json");
+      expect(writeFileStub.calledWith(mcpConfigPath, sinon.match(/"firebase":/))).to.be.false;
+    });
+
+    it("should prompt user for Dart MCP server and configure if they confirm", async () => {
+      accessStub.withArgs(sinon.match("pubspec.yaml")).resolves();
+      commandStub.withArgs("dart").returns(true);
+      confirmStub.withArgs(sinon.match({ message: sinon.match(/Dart MCP server/) })).resolves(true);
+
+      await migrate(testRoot);
+
+      const mcpConfigDir = path.join(require("os").homedir(), ".gemini", "antigravity");
+      const mcpConfigPath = path.join(mcpConfigDir, "mcp_config.json");
+      expect(writeFileStub.calledWith(mcpConfigPath, sinon.match(/"dart":/))).to.be.true;
+    });
+
+    it("should prompt user for Dart MCP server and NOT configure if they decline", async () => {
+      accessStub.withArgs(sinon.match("pubspec.yaml")).resolves();
+      commandStub.withArgs("dart").returns(true);
+      confirmStub
+        .withArgs(sinon.match({ message: sinon.match(/Dart MCP server/) }))
+        .resolves(false);
+
+      await migrate(testRoot);
+
+      const mcpConfigDir = path.join(require("os").homedir(), ".gemini", "antigravity");
+      const mcpConfigPath = path.join(mcpConfigDir, "mcp_config.json");
+      expect(writeFileStub.calledWith(mcpConfigPath, sinon.match(/"dart":/))).to.be.false;
     });
   });
 
--- a/src/firebase_studio/migrate.ts
+++ b/src/firebase_studio/migrate.ts
@@ -53,12 +53,23 @@ async function setupAntigravityMcpServer(rootPath: string, appType?: AppType): P
     let updated = false;
 
     if (!mcpConfig.mcpServers["firebase"]) {
-      mcpConfig.mcpServers["firebase"] = {
-        command: "npx",
-        args: ["-y", "firebase-tools@latest", "mcp", "--dir", path.resolve(rootPath)],
-      };
-      updated = true;
-      logger.info(`✅ Configured Firebase MCP server in ${mcpConfigPath}`);
+      if (utils.commandExistsSync("npx")) {
+        const confirmFirebase = await prompt.confirm({
+          message: "Would you like to enable the Firebase MCP server for Antigravity?",
+          default: true,
+        });
+
+        if (confirmFirebase) {
+          mcpConfig.mcpServers["firebase"] = {
+            command: "npx",
+            args: ["-y", "firebase-tools@latest", "mcp", "--dir", path.resolve(rootPath)],
+          };
+          updated = true;
+          logger.info(`✅ Configured Firebase MCP server in ${mcpConfigPath}`);
+        }
+      } else {
+        logger.info("ℹ️ npx not found on PATH, skipping Firebase MCP server configuration.");
+      }
     } else {
       logger.info("ℹ️ Firebase MCP server already configured in Antigravity, skipping.");
     }
@@ -66,12 +77,19 @@ async function setupAntigravityMcpServer(rootPath: string, appType?: AppType): P
     if (appType === "FLUTTER") {
       if (utils.commandExistsSync("dart")) {
         if (!mcpConfig.mcpServers["dart"]) {
-          mcpConfig.mcpServers["dart"] = {
-            command: "dart",
-            args: ["mcp-server"],
-          };
-          updated = true;
-          logger.info(`✅ Configured Dart MCP server in ${mcpConfigPath}`);
+          const confirmDart = await prompt.confirm({
+            message: "Would you like to enable the Dart MCP server for Antigravity?",
+            default: true,
+          });
+
+          if (confirmDart) {
+            mcpConfig.mcpServers["dart"] = {
+              command: "dart",
+              args: ["mcp-server"],
+            };
+            updated = true;
+            logger.info(`✅ Configured Dart MCP server in ${mcpConfigPath}`);
+          }
         } else {
           logger.info("ℹ️ Dart MCP server already configured in Antigravity, skipping.");
         }
@@ -280,7 +298,9 @@ async function injectAntigravityContext(
   try {
     const result = spawnSync(
       "npx",
-      ["-y", "skills", "add", "firebase/agent-skills", "-a", "antigravity", "--skill", "*", "-y"],
+      // gemini-CLI is chosen as a workaround for the .agents subfolder (instead of .agent)
+      // which is current for antigravity's location from vercel.
+      ["-y", "skills", "add", "firebase/agent-skills", "-a", "gemini-cli", "--skill", "*", "-y"],
       {
         cwd: rootPath,
         stdio: "ignore",


src/deploy/functions/runtimes/dart/index.ts
```diff
@@ -0,0 +1,316 @@
...
+    await fs.promises.mkdir(binDir, { recursive: true });
+
+    logLabeledBullet("functions", "compiling Dart to linux-x64 executable...");
+
+    const compileProcess = spawn(
+      this.bin,
+      [
+        "compile",
+        "exe",
+        "lib/main.dart",
```
brittanycho: Are we sure we want to keep this entry point hard coded like this for "lib/main.dart"?

src/deploy/functions/runtimes/dart/index.ts
```diff
@@ -0,0 +1,316 @@
...
+
+    // Cross-compile Dart to a Linux x86_64 executable for Cloud Run.
+    // Skip compilation when running in the emulator (the emulator runs
+    // Dart source directly via `dart run`).
+    if (EmulatorRegistry.isRunning(Emulators.FUNCTIONS)) {
+      logger.debug("Skipping Dart compilation in emulator mode.");
+      return;
+    }
+
+    // Requires Dart 3.8+ for --target-os and --target-arch support.
```
brittanycho: Do you think it's worthwhile to enforce this "Requires Dart 3.8+" or check to provide a error message if the user's local Dart SDK is too old?

src/deploy/functions/runtimes/dart/index.ts
```diff
@@ -0,0 +1,316 @@
...
+  if (!supported.isRuntime(runtime)) {
+    throw new FirebaseError(`Runtime ${runtime as string} is not a valid Dart runtime`);
+  }
+  if (!supported.runtimeIsLanguage(runtime, "dart")) {
+    throw new FirebaseError(
+      `Internal error. Trying to construct a dart runtime delegate for runtime ${runtime}`,
+      { exit: 1 },
+    );
+  }
+  return Promise.resolve(new Delegate(context.projectId, context.sourceDir, runtime));
```
brittanycho, line 42: Do you think it makes sense to add a check like in validate() that explicitly verifies the binary exists and optionally checks the version (going back to the other comment I left about requiring Dart 3.8+)?

src/deploy/functions/runtimes/dart/index.ts
Lyokone: I've added a constant for lib/main.dart

src/deploy/functions/runtimes/dart/index.ts
Lyokone: I've added enforcing for this Dart version

templates/init/functions/dart/pubspec.yaml
```diff
@@ -0,0 +1,14 @@
...
+description: Firebase Functions for Dart
+version: 1.0.0
+
+environment:
+  sdk: '>=3.0.0 <4.0.0'
+
+dependencies:
+  # TODO(ehesp): Replace with published package version once available on pub.dev
+  firebase_functions:
+    path: ../
```
brittanycho, line 11: Should we leave this as a relative path or update it?

templates/init/functions/dart/pubspec.yaml
```diff
@@ -0,0 +1,14 @@
+name: functions
+description: Firebase Functions for Dart
+version: 1.0.0
+
+environment:
+  sdk: '>=3.0.0 <4.0.0'
```
brittanycho: Should we update the template constraint to match the runtime requirement of 3.8?

src/deploy/functions/runtimes/dart/index.ts
```diff
@@ -0,0 +1,365 @@
...
+      await fs.promises.access(entryPointPath, fs.constants.R_OK);
+    } catch (err: any) {
+      throw new FirebaseError(
+        `Could not find entry point at ${entryPointPath}. ` +
+          `Firebase Functions for Dart expects your main function in ${this.entryPoint}.`,
+      );
+    }
+  }
+
+  async build(): Promise<void> {
```
brittanycho, line 150: Should a check be implemented in validate() or build() to run `dart pub get` if .dart_tool/package_config.json is missing or stale? I think without this, iiuc users would need to manually run pub get, before the deploy or the deployment will fail

brittanycho:
Left some additional comments, mostly minor nits and otherwise I think generally lgtm

@inlined, @Ehesp - was curious if you had any thoughts or comments here?

templates/init/functions/dart/server.dart
```diff
@@ -0,0 +1,13 @@
...
+
+void main(List<String> args) {
+  fireUp(args, (firebase) {
+
+    firebase.https.onRequest(
+      name: 'helloWorld',
+      options: const HttpsOptions(cors: Cors(['*'])),
+      (request) async {
+        return Response(200, body: 'Hello from Dart Functions!');
+      });
```
natebosch: [optional]


```suggestion
      (request) async => Response(200, body: 'Hello from Dart Functions!'),
     );
```

templates/init/functions/dart/pubspec.yaml
natebosch:
```suggestion
  sdk: ^3.8.0
```

templates/init/functions/dart/pubspec.yaml
```diff
@@ -0,0 +1,14 @@
+name: functions
+description: Firebase Functions for Dart
```
natebosch: In the flutter and Dart templates we fill in a package name  (I think based on the directory name). For this template assuming we need a fixed value I might expand it a little bit for clarity.

```suggestion
name: functions_template
description: An app using Firebase Functions for Dart
```

scripts/agent-evals/templates/crashlytics-flutter/.dart_tool/package_config.json
natebosch, line 1: Are these .dart_tool directory checkins intentional? They are not designed to be source controlled and I'm not sure of all the implications of doing that. The gitignore below correctly includes them.

scripts/agent-evals/templates/crashlytics-flutter/pubspec.lock
natebosch, line 1: This file is also typically not checked in except for leaf dependency app packages.

templates/init/functions/dart/server.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:firebase_functions/firebase_functions.dart';
+
+void main(List<String> args) {
+  fireUp(args, (firebase) {
+
+    firebase.https.onRequest(
+      name: 'helloWorld',
+      options: const HttpsOptions(cors: Cors(['*'])),
```
jhuleatt: can we also set `maxInstances`? maybe globally like JS if that works in Dart?

https://github.com/firebase/firebase-tools/blob/87ec0f78afe1d3803afaac14a0feecec55a8aca9/templates/init/functions/javascript/index.js#L14-L24

src/deploy/functions/runtimes/dart/index.ts
Lyokone: I've added this as well.

templates/init/functions/dart/pubspec.yaml
Lyokone: As soon as it's published we'll be able to add the correct version

src/deploy/functions/runtimes/dart/index.ts
```diff
@@ -0,0 +1,398 @@
...
+  if (!supported.isRuntime(runtime)) {
+    throw new FirebaseError(`Runtime ${runtime as string} is not a valid Dart runtime`);
+  }
+  if (!supported.runtimeIsLanguage(runtime, "dart")) {
+    throw new FirebaseError(
+      `Internal error. Trying to construct a dart runtime delegate for runtime ${runtime}`,
+      { exit: 1 },
+    );
+  }
+  return Promise.resolve(new Delegate(context.projectId, context.sourceDir, runtime));
```
inlined, line 42: You don't need to call Promise.resolve if you're in an async function

src/deploy/functions/runtimes/dart/index.ts
```diff
@@ -0,0 +1,398 @@
...
+        `Could not find a Dart SDK. Make sure the '${this.bin}' command is available on your PATH.`,
+      );
+    }
+
+    // `dart --version` outputs e.g. "Dart SDK version: 3.8.0 (stable) ... on "macos_arm64""
+    const versionOutput = (result.stdout || result.stderr || "").toString();
+    const match = /Dart SDK version:\s*(\d+\.\d+\.\d+)/.exec(versionOutput);
+    if (match) {
+      const installedVersion = match[1];
+      if (installedVersion.localeCompare(MIN_DART_SDK_VERSION, undefined, { numeric: true }) < 0) {
```
inlined, line 86: Fascinating method of doing this. Do you want to make sure you blindly accept all versions of dart instead of using the semver library and doing semver arithmetic?

src/deploy/functions/runtimes/dart/index.ts
```diff
@@ -0,0 +1,398 @@
...
+            `Please upgrade your Dart SDK.`,
+        );
+      }
+    } else {
+      logger.debug(`Could not parse Dart SDK version from: ${versionOutput}`);
+    }
+
+    // Verify pubspec.yaml exists and is readable.
+    const pubspecYamlPath = path.join(this.sourceDir, "pubspec.yaml");
+    try {
```
inlined, line 99: This is the more modern version of what you do in tryCreateDelegate. Why not reconcile to just one?

src/deploy/functions/runtimes/dart/index.ts
```diff
@@ -0,0 +1,398 @@
...
+      await fs.promises.access(pubspecYamlPath, fs.constants.R_OK);
+    } catch (err: any) {
+      throw new FirebaseError(`Failed to read pubspec.yaml at ${pubspecYamlPath}: ${err.message}`);
+    }
+
+    // Verify the entry point file exists.
+    const entryPointPath = path.join(this.sourceDir, this.entryPoint);
+    try {
```
inlined, line 107: If you're going to check for three files using async libraries you might as well do it in parallel

src/deploy/functions/runtimes/dart/index.ts
```diff
@@ -0,0 +1,398 @@
...
+    } catch (err: any) {
+      throw new FirebaseError(
+        `Could not find entry point at ${entryPointPath}. ` +
+          `Firebase Functions for Dart expects your main function in ${this.entryPoint}.`,
+      );
+    }
+
+    // Run `dart pub get` if dependencies have not been resolved yet.
+    const packageConfigPath = path.join(this.sourceDir, ".dart_tool", "package_config.json");
+    try {
```
inlined, line 118: Always prefer to return early and avoid nesting. If this try is effectively an if statement, consider something like:

```
try {
  await access
  return;
} catch () { /* noop */ }

// Install dependencies
```

Also, all of this is probably because exists is deprecated and you don't have a promise version? The reasons for its deprecation are basically irrelevant here (it's encouraged for you to try to do the thing rather than using exists as a dry run).

I might just define exists using promisify and simplify all this code:

```
const [pubspecExists, entryPointExists, packageConfigExists] = await Promise.all([pubspecYamlPath, entryPointPath, packageConfigPath]).map(exists);

if (!pubspecExists) {
  throw error
}

if (!entryPointExists) {
  throw erroR
}

if (packageConfigExists) {
  return;
}

// install pacakges
```

src/deploy/functions/runtimes/dart/index.ts
```diff
@@ -0,0 +1,398 @@
...
+      await fs.promises.access(packageConfigPath, fs.constants.R_OK);
+    } catch {
+      logLabeledBullet("functions", "running dart pub get...");
+      const pubGetProcess = spawn(this.bin, ["pub", "get"], {
```
inlined, line 122: I know this is copying other code but I wonder if there's an opportunity for a utility here that we can use throughout the codebase...

```
import { spawn } from 'node:child_process';
import { createInterface } from 'node:readline';
import { once } from 'node:events';

async function execChild({
  bin: string,
  args: string[],
  cwd?: string,
  logsPrefix?: string,
): Promise<number> {
  const child = spawn('dart', ['pub', 'get'], { stdio: ['ignore', 'pipe', 'pipe'] });
  child.stderr?.pipe(child.stdout!);

  const rl = createInterface({ input: child.stdout!, terminal: false });
  const printOutput = (async () => {
    for await (const line of rl) {
      logger.debug(logsPrefix ? `${logsPrefix} ${line.trim}` ? line.trim()`);
    }
  })();

  const processEnd = Promise.race([
    once(child, 'close'),
    once(child, 'error').then(([err]) => { throw err; })
  ]);

  const [_, exitCode] = await Promise.all([printOutput, processEnd]);

  return exitCode;
}
```

Then you could just say

```
const exitCode = await execChild({ bin: this.bin, args: ["pub", "get"], logsPrefix: "[dart pub get"]});
if (exitCode) { 
  throw new FirebaseError(...);
}
```

here and below say

```
const buildRunnerExit =  await execChild({
  bin: this.bin,
  args: ["run", "build_runner", "build", "--delete-conflicting-outputs"],
  cwd: this.sourceDir,
  loggerPrefix: "[build_runner]"
});


// ...
const compileExit = await execChild({
  bin: this.bin,
  args: ["compile", "exe",  this.entryPoint, "-o", "bin/server", "--target-os=linux", "--target-arch=x64"],
  cwd: this.sourceDir,
  logsPrefix: "[dart compile]",
});
```

src/deploy/functions/runtimes/dart/index.ts
```diff
@@ -0,0 +1,398 @@
...
+
+    logLabeledBullet("functions", "Dart compilation complete.");
+  }
+
+  /**
+   * Start build_runner in watch mode for hot reload.
+   * Returns a cleanup function that stops the build_runner process.
+   * The returned promise resolves once the initial build completes.
+   */
+  async watch(onRebuild?: () => void): Promise<() => Promise<void>> {
```
inlined, line 254: This is amazing

src/emulator/functionsEmulator.ts
```diff
@@ -419,11 +421,18 @@ export class FunctionsEmulator implements EmulatorInstance {
       "Content-Type": "application/json",
       "Content-Length": `${reqBody.length}`,
     };
+
+    // For Dart, include the function name in the path so the server can route
+    // For other runtimes, use / as they use FUNCTION_TARGET env var
+    const isDart = isLanguageRuntime(record.backend.runtime, "dart");
```
inlined, line 427: I know this is a bit of a bigger change, but.... what if we just make this the behavior when platform is "run" and always make dart emit "run"?

We'll almost certainly want to investigate allowing collocation of functions in Run services if/when we move other languages to CRF.

src/emulator/functionsEmulator.ts
```diff
@@ -1676,8 +1771,10 @@ export class FunctionsEmulator implements EmulatorInstance {
     const secretEnvs = await this.resolveSecretEnvs(backend, trigger);
 
     let runtime;
-    if (backend.runtime!.startsWith("python")) {
+    if (isLanguageRuntime(backend.runtime, "python")) {
       runtime = await this.startPython(backend, { ...runtimeEnv, ...secretEnvs });
+    } else if (isLanguageRuntime(backend.runtime, "dart")) {
+      runtime = await this.startDart(backend, { ...runtimeEnv, ...secretEnvs });
     } else {
```
inlined, line 1778: I think we now have the ability to now say

```
} else if (isLangaugeRuntime(backend.runtime, "nodejs")) {
  // ...
}
assertExhaustive(backend.runtime);
```

brittanycho (approved):

PR #9986, opened 2026-02-27
Firebase Functions can handle an Extensions outage

inlined:
To address existing outage

### DIFF
--- a/src/deploy/extensions/prepare.spec.ts
+++ b/src/deploy/extensions/prepare.spec.ts
@@ -41,7 +41,11 @@ describe("Extensions prepare", () => {
 
       const context: Context = {};
       const payload: Payload = {};
-      const options: any = {};
+      const options: any = {
+        config: {
+          src: { functions: { source: "functions" } },
+        },
+      };
       const builds = {};
 
       // This should not throw.
@@ -62,7 +66,10 @@ describe("Extensions prepare", () => {
       const context: Context = {};
       const payload: Payload = {};
       const options: any = {
-        config: { get: () => [] },
+        config: {
+          get: () => [],
+          src: { functions: { source: "functions" } },
+        },
         rc: { getEtags: () => [] },
         dryRun: true,
       };


src/deploy/extensions/prepare.ts
```diff
@@ -167,19 +167,26 @@ export async function prepareDynamicExtensions(
...
-  let haveExtensions = await planner.haveDynamic(projectId);
-  haveExtensions = haveExtensions.filter((e) =>
-    extensionMatchesAnyFilter(e.labels?.codebase, e.instanceId, filters),
-  );
+    haveExtensions = await planner.haveDynamic(projectId);
+    haveExtensions = haveExtensions.filter((e) =>
+      extensionMatchesAnyFilter(e.labels?.codebase, e.instanceId, filters),
+    );
+  } catch (err) {
+    logLabeledError("extensions",
```
pavelgj: the catch block feels a bit too broad to confidently assume that "Extensions is having an outage"... I can imagine permission errors perhaps...
also, wondering if it the option to "skip extentions" should be provided to the user

pavelgj: lol... gemini-code-assist said the exact same thing.

pavelgj (approved):

src/deploy/extensions/prepare.ts
inlined: We can massage the text but >=500 is not correct. The current outage causes a 403

src/deploy/extensions/prepare.ts
```diff
@@ -167,19 +167,26 @@ export async function prepareDynamicExtensions(
...
+      "Firebase Extensions is having an outage. Skipping extensions from functions codebase.",
+    );
+    return;
```
taeold, line 192: This will still try to deploy the extensions no?

taeold: nvm. missted the implication of return statement.

PR #9998, opened 2026-03-02
fix(mcp): write firebase-debug.log to ~/.firebase/ instead of cwd

### LINKED ISSUE #9982: [MCP] firebase-debug.log is created in user's working directory on every MCP server startup

**firebase-tools:** 15.1.0 (via Claude Code Firebase plugin)

**Platform:** macOS (Darwin 25.3.0)

### [REQUIRED] Test case

Install the Firebase plugin for Claude Code (which runs `npx firebase-tools mcp --dir .` as an MCP server). Open any project directory and use any Firebase MCP tool.

1. Install the official Firebase plugin for Claude Code
2. Open a terminal in any project directory (e.g. an iOS project with no Firebase config)
3. Start a Claude Code session — the Firebase MCP server starts automatically
4. Invoke any Firebase MCP tool (e.g. `firebase_get_environment`)
5. Check the working directory

The MCP server should not create `firebase-debug.log` in the user's working directory. When running as an MCP server, debug logs should either:

- Be written to a well-known location (e.g. `~/.firebase/logs/`), or
- Be suppressed by default unless explicitly enabled, or
- Be configurable via an environment variable (e.g. `FIREBASE_DEBUG_LOG_FILE`)

A `firebase-debug.log` file is created in `process.cwd()` every time the MCP server starts. This pollutes the user's project directory with an unrelated file.

The root cause is in `src/logger.ts` — `findAvailableLogFile()` hardcodes the log path to the current working directory:

```ts
const logFilename = path.join(process.cwd(), c);
```

This behavior makes sense for interactive CLI usage, but is problematic for MCP server mode where the server runs in the background and the user did not explicitly invoke a Firebase command.

VishnuVardhanBR:
## Problem

When the Firebase MCP server starts (via `firebase mcp`), it calls `useFileLogger()` unconditionally. `useFileLogger()` calls `findAvailableLogFile()` which writes `firebase-debug.log` to `process.cwd()`. This pollutes the user's project directory with an unrelated file on every MCP server startup — even in projects that have no Firebase configuration at all.

Fixes #9982.

## Root Cause

In `src/bin/mcp.ts`:
```ts
process.env.IS_FIREBASE_MCP = "true";
useFileLogger(); // writes to process.cwd()/firebase-debug.log
```

And in `src/logger.ts`, `findAvailableLogFile()` hardcodes the log path to `process.cwd()`:
```ts
const logFilename = path.join(process.cwd(), c);
```

## Fix

Pass an explicit log file path to `useFileLogger()` when starting in MCP mode. The log is redirected to `~/.firebase/firebase-debug.log`, which:
- Is already used by Firebase tooling for non-project data
- Keeps the user's working directory clean
- Requires no changes to `logger.ts`

```ts
const mcpLogDir = join(homedir(), ".firebase");
mkdirSync(mcpLogDir, { recursive: true });
useFileLogger(join(mcpLogDir, "firebase-debug.log"));
```

## Testing

Manually verified by running `npx firebase-tools mcp` in a non-Firebase directory and confirming:
- No `firebase-debug.log` is created in the working directory
- `~/.firebase/firebase-debug.log` is created with expected debug output

### DIFF
--- a/firebase-vscode/CHANGELOG.md
+++ b/firebase-vscode/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## NEXT
 
+## 2.2.0
+
+- Update internal `firebase-tools` dependency to 15.9.0
+
 ## 2.1.2
 
 - Update internal `firebase-tools` dependency to 15.7.0
--- a/firebase-vscode/package-lock.json
+++ b/firebase-vscode/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "firebase-dataconnect-vscode",
-  "version": "2.1.2",
+  "version": "2.2.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "firebase-dataconnect-vscode",
-      "version": "2.1.2",
+      "version": "2.2.0",
       "dependencies": {
         "@preact/signals-core": "^1.4.0",
         "@preact/signals-react": "1.3.6",
--- a/firebase-vscode/package.json
+++ b/firebase-vscode/package.json
@@ -4,7 +4,7 @@
   "publisher": "GoogleCloudTools",
   "icon": "./resources/firebase_dataconnect_logo.png",
   "description": "Firebase Data Connect for VSCode",
-  "version": "2.1.2",
+  "version": "2.2.0",
   "engines": {
     "vscode": "^1.69.0"
   },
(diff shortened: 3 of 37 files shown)


joehan:
Thanks for the contirbution @VishnuVardhanBR! I've spun this off into #10175 to target a location that is a little more appropriate + clean up the change log, but this was a very helpful change. Closing in favor of that one.

PR #10000, opened 2026-03-02
Update source map upload command options

tonybaroneee:
- Remove `telemetryServerUrl` option
- Change mapping file positional argument from required to optional (defaults to root dir)

Feedback from API design.

### DIFF
--- a/src/commands/crashlytics-sourcemap-upload.ts
+++ b/src/commands/crashlytics-sourcemap-upload.ts
@@ -54,7 +54,7 @@ export const command = new Command("crashlytics:sourcemap:upload [mappingFiles]"
 
     // Find and upload mapping files
     const rootDir = options.projectRoot ?? process.cwd();
-    const filePath = path.relative(rootDir, mappingFiles || "");
+    const filePath = path.relative(rootDir, mappingFiles || ".") || ".";
     let fstat: fs.Stats;
     try {
       fstat = statSync(filePath);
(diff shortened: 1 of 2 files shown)


src/commands/crashlytics-sourcemap-upload.ts
```diff
@@ -40,10 +39,8 @@ export const command = new Command("crashlytics:sourcemap:upload <mappingFiles>"
     "--app-version <appVersion>",
     "the version of your Firebase app (defaults to Git commit hash, if available)",
   )
-  .option("--telemetry-server-url <telemetryServerUrl>", "the url of the telemetry API")
-  .action(async (mappingFiles: string, options: CommandOptions) => {
+  .action(async (mappingFiles: string | undefined, options: CommandOptions) => {
```
andrewbrook, line 42: We should probably filter out node_modules, otherwise searching every subdirectory might be too slow

src/commands/crashlytics-sourcemap-upload.ts
```diff
@@ -57,7 +54,7 @@ export const command = new Command("crashlytics:sourcemap:upload <mappingFiles>"
 
     // Find and upload mapping files
     const rootDir = options.projectRoot ?? process.cwd();
-    const filePath = path.relative(rootDir, mappingFiles);
+    const filePath = path.relative(rootDir, mappingFiles || "");
```
andrewbrook: Can you add a unit test to ensure this starts in the right place?

andrewbrook (approved):

src/commands/crashlytics-sourcemap-upload.ts
tonybaroneee: `node_modules` and `.git` are already being filtered out in the directory-recursive code below this.

src/commands/crashlytics-sourcemap-upload.ts
tonybaroneee: Done.

PR #10003, opened 2026-03-02
GA4 tracking, platform (windows vs linux) and framework.

christhompsongoogle:
GA4 tracking, platform (windows vs linux) and framework.

### DIFF
--- a/src/firebase_studio/migrate.ts
+++ b/src/firebase_studio/migrate.ts
@@ -549,6 +549,6 @@ export async function migrate(
     );
   }
 
-  void track.trackGA4("firebase_studio_migrate", { app_type: appType, result: "success" });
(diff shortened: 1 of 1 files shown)


src/firebase_studio/migrate.ts
```diff
@@ -439,4 +500,6 @@ export async function migrate(
   }
 
   await askToOpenAntigravity(rootPath, appName, options.noStartAgy);
+
+  void track.trackGA4("firebase_studio_migrate", { app_type: appType, result: "success" });
```
joehan: Maybe move this before askToOpenAgy to make sure that it doesn't get cut off by the process ending? Or, to be safe, just await it.

christhompsongoogle: Good idea, that

PR #10012, opened 2026-03-03
Observe the project (ID) flag

christhompsongoogle:
Observe the project (ID) flag

### DIFF
--- a/src/appdistribution/yaml_helper.ts
+++ b/src/appdistribution/yaml_helper.ts
@@ -1,8 +1,9 @@
 import * as jsYaml from "js-yaml";
 import { getErrMsg, FirebaseError } from "../error";
-import { TestCase } from "./types";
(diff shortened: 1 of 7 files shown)


src/commands/studio-export.ts
```diff
@@ -18,5 +18,5 @@ export const command = new Command("studio:export <path>")
     }
     const rootPath = path.resolve(exportPath);
     logger.info(`Exporting Studio apps from ${rootPath} to Antigravity...`);
-    await migrate(rootPath, { noStartAgy: !options.startAgy });
+    await migrate(rootPath, { noStartAgy: !options.startAgy, projectId: options.project as string });
```
joehan: Rather than casting the type here, you can just extend the Options type from options.ts above

christhompsongoogle: Done

joehan (approved):

PR #10017, opened 2026-03-04
Secret upload apphosting

christhompsongoogle:
Apphosting secret upload, this is a follow-on from https://github.com/firebase/firebase-tools/pull/10010


falahat (commented):
LGTM, but feedback is the name uploadSecrets might be too general since it really just handles a specific secret (Gemini API keys) but overall approved

PR #10022, opened 2026-03-04
update fdc emulator to 3.2.1

hlshen:
- Updated the Firebase Data Connect local toolkit to v3.2.1, which includes the following changes: (#10022)
  - Resolved metadata pathing issue for @redact and @check directives


rosalyntan:
Some copy-paste issues: `v3.2.1` should just be `3.2.1` (and there's a `vv3.2.1` that should be `v3.2.1`)

PR #10033, opened 2026-03-06
Add support for VPC direct connect in GCF 2nd gen

inlined:
Companion PR to firebase/firebase-functions#1823.

Successfully tested e2e with 
Screenshot 2026-03-06 at 12 03 02 PM

### DIFF
--- a/src/commands/index.ts
+++ b/src/commands/index.ts
@@ -266,9 +266,9 @@ export function load(client: CLIClient): CLIClient {
   client.target.clear = loadCommand("target-clear");
   client.target.remove = loadCommand("target-remove");
   client.use = loadCommand("use");
+  client.apptesting = {};
+  client.apptesting.execute = loadCommand("apptesting");
   if (experiments.isEnabled("apptesting")) {
-    client.apptesting = {};
-    client.apptesting.execute = loadCommand("apptesting");
     client.apptesting.wata = loadCommand("apptesting-wata");
   }
 


src/deploy/functions/runtimes/discovery/v1alpha1.ts
```diff
@@ -175,10 +180,15 @@ function assertBuildEndpoint(ep: WireEndpoint, id: string): void {
   });
   if (ep.vpc) {
     assertKeyTypes(prefix + ".vpc", ep.vpc, {
-      connector: "string",
+      connector: "string?",
       egressSettings: (setting) => setting === null || build.AllVpcEgressSettings.includes(setting),
+      networkInterfaces: "array?",
     });
-    requireKeys(prefix + ".vpc", ep.vpc, "connector");
+    if (!ep.vpc.connector && !ep.vpc.networkInterfaces) {
```
brittanycho, line 187: Based on the logic here, it looks like its checking if at least one if present but it also looks like it allows both connector and networkInterfaces to be present (and iiuc, then silently prioritize connector and drop networkInterfaces if both are present)? Do we want to allow clients to provide both or should we log a warning or error if they do?

src/gcp/cloudfunctionsv2.ts
```diff
@@ -742,6 +753,16 @@ export function endpointFromFunction(gcfFunction: OutputCloudFunction): backend.
         "egressSettings",
         "vpcConnectorEgressSettings",
       );
+    } else if (gcfFunction.serviceConfig.directVpcNetworkInterface) {
+      endpoint.vpc = { networkInterfaces: gcfFunction.serviceConfig.directVpcNetworkInterface };
+      if (
```
brittanycho: optional nit: wondering if it makes sense to use string replacement or mapping dictionary here to safeguard against any future additions that might not start with "VPC_EGRESS_"?

src/deploy/functions/runtimes/discovery/v1alpha1.ts
inlined: Great check. Throwing an error.

src/gcp/cloudfunctionsv2.ts
inlined: Yeah doesn't hurt. You're definitely going to hold me to a higher standard for validation it seems 😊

PR #10034, opened 2026-03-06
Create a command to export Firebase Functions' IaC

inlined:
Checkpointing a work in progress for an experimental feature to export Functions' IaC. This can be used to eventually support management via Terraform or the Application Design Center.


inlined:
My update changes so much that I might as well create a fresh review.

PR #10047, opened 2026-03-09
Include cache configs for sdk generation

itsrakhil:
This PR adds type definitions and configuration forwarding for the `clientCache` property within `connector.yaml` for Data Connect SDK generation. Developers can now pass caching settings (such as `maxAge`, `maxSize`, and `storage`) down to the underlying `dataconnect` toolkit for the generated JavaScript, Swift, Kotlin, and Dart SDKs.

### DIFF
--- a/src/frameworks/frameworks.ts
+++ b/src/frameworks/frameworks.ts
@@ -24,6 +24,7 @@ export const WebFrameworks: Record<string, Framework> = {
   nuxt2,
   preact,
   svelte,
+  // TODO: replace with "sveltekit" once we confirm it won't cause breakages
   svelekit,
   react,
   vite,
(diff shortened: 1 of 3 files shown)


yuchenshi (approved):

src/dataconnect/types.ts
```diff
@@ -173,29 +173,48 @@ export interface SupportedFrameworks {
...
+   */
+  maxSize?: string;
+  storage?: "persistent" | "memory";
+}
+
 export interface AdminNodeSDK {
   outputDir: string;
   package: string;
   packageJsonDir?: string;
+  clientCache?: CacheSettings;
```
yuchenshi: I don't think this is supported on the admin SDK ar all

yuchenshi (changes requested):

src/dataconnect/types.ts
aashishpatil-g: Correct. Only client SDKs.

aashishpatil-g (approved):

itsrakhil:
Tested manually calling the connector apis and added unit tests.

PR #10048, opened 2026-03-09
Define a common method, getAutoinitEnvVars.

falahat:
### Description

This simplifies some future changes and allows us to reuse getAutoinitEnvVars across our codebase.

### Scenarios Tested

I did the same testing as https://github.com/firebase/firebase-tools/pull/10049 (pr#10049)

I tested a source deploy of a nextjs app which included secrets, env vars, emulator-specific env vars, and testing for the admin sdk (load an image from google storage bucket.)

I deployed the backend (via source deploy) and verified it worked as expected.

I started an emulator and verified my local app and verified it worked as expected. apphosting.emulator.yaml was respected.

I also tested these scenarios to make sure the priority for the autoinit vars works correctly:
I set up a real time database so that the database url env var would be auto-injected.

1. I tested running the emulator with no overrides, and it used the values from prod/default autoinjected values (i.e. for the database url)
2. I ran the emulator, including the storage/database emulator and this correctly injected the FIREBASE_DATABASE_EMULATOR_HOST value, for example.
3. I overrode FIREBASE_CONFIG in my apphosting.yaml and ran the emulator. It correctly used the values from this.

### Sample Commands
```
firebase deploy
```
```
firebase emulators:start
```

### DIFF
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -201,7 +201,7 @@ export class FirebaseMcpServer {
     this.logger.debug("detecting active features of Firebase MCP server...");
     const projectId = (await this.getProjectId()) || "";
     const accountEmail = await this.getAuthenticatedUser();
-    const isBillingEnabled = projectId ? await checkBillingEnabled(projectId) : false;
+    const isBillingEnabled = projectId ? await this.safeCheckBillingEnabled(projectId) : false;
     const ctx = this._createMcpContext(projectId, accountEmail, isBillingEnabled);
     const detected = await Promise.all(
       SERVER_FEATURES.map(async (f) => {
@@ -252,7 +252,7 @@ export class FirebaseMcpServer {
     // We need a project ID and user for the context, but it's ok if they're empty.
     const projectId = (await this.getProjectId()) || "";
     const accountEmail = await this.getAuthenticatedUser();
-    const isBillingEnabled = projectId ? await checkBillingEnabled(projectId) : false;
+    const isBillingEnabled = projectId ? await this.safeCheckBillingEnabled(projectId) : false;
     const ctx = this._createMcpContext(projectId, accountEmail, isBillingEnabled);
     return availableTools(ctx, this.activeFeatures, this.detectedFeatures, this.enabledTools);
   }
@@ -266,7 +266,7 @@ export class FirebaseMcpServer {
     // We need a project ID and user for the context, but it's ok if they're empty.
     const projectId = (await this.getProjectId()) || "";
     const accountEmail = await this.getAuthenticatedUser();
-    const isBillingEnabled = projectId ? await checkBillingEnabled(projectId) : false;
+    const isBillingEnabled = projectId ? await this.safeCheckBillingEnabled(projectId) : false;
     const ctx = this._createMcpContext(projectId, accountEmail, isBillingEnabled);
     return availablePrompts(ctx, this.activeFeatures, this.detectedFeatures);
   }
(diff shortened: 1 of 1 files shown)


src/emulator/apphosting/serve.ts
```diff
@@ -156,14 +157,7 @@ export async function start(options?: StartOptions): Promise<{ hostname: string;
...
   } else {
     const webappConfig = await getBackendAppConfig(options?.projectId, options?.backendId);
-    if (webappConfig) {
-      environmentVariablesToInject["FIREBASE_WEBAPP_CONFIG"] ||= JSON.stringify(webappConfig);
-      environmentVariablesToInject["FIREBASE_CONFIG"] ||= JSON.stringify({
-        databaseURL: webappConfig.databaseURL,
-        storageBucket: webappConfig.storageBucket,
-        projectId: webappConfig.projectId,
-      });
-    }
```
falahat, line 166: The pevious version of this code used ||= (logical or operator) to *conditionally* assign these variables if they were NOT already defined in the `environmentVariablesToInject` map.

falahat: ah this is actually safe because Object.assign(x, {}) is a no-op. If the config is empty or has empty subvalues, Object,assign will leave the original value in environmentVariablesToInject untouched.

src/emulator/apphosting/serve.ts
```diff
@@ -156,14 +157,7 @@ export async function start(options?: StartOptions): Promise<{ hostname: string;
...
+    Object.assign(environmentVariablesToInject, getAutoinitEnvVars(webappConfig));
```
annajowang: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#description

`Object.assign(target, source1)`

"Properties in the target object are overwritten by properties in the sources if they have the same [key](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys). Later sources' properties overwrite earlier ones."
and

"The properties are overwritten by other objects that have the same properties later in the parameters order."


so if i'm reading correctly: I believe to match old behavior we would want:
```
Object.assign({}, getAutoinitEnvVars(webappConfig), environmentVariablesToInject)
```

falahat: I might be misunderstanding, but with the old code, the intent is to write values into environmentVariablesToInject, so we want "environmentVariablesToInject" to be the "target" of the assignment. 

For example, `environmentVariablesToInject["FIREBASE_WEBAPP_CONFIG"] ||= JSON.stringify(webappConfig);` would set the FIREBASE_WEBAPP_CONFIG iff webappconfig is defined.

The "source" of the data (where we read from) should also be `getAutoInitEnvVars(...)`. 

Let me know if i'm misunderstanding the docs / method though

annajowang: My read was

Old code:
`environmentVariablesToInject["FIREBASE_WEBAPP_CONFIG"] ||= JSON.stringify(webappConfig);`
=> if there is no "FIREBASE_WEBAPP_CONFIG" in `environmentVariablesToInject` then set it to `JSON.stringify(webappConfig)`
=> `environmentVariablesToInject` takes precedence

New code:
`Object.assign(environmentVariablesToInject, getAutoinitEnvVars(webappConfig))` 
=>  whatever values from `getAutoinitEnvVars(webappConfig)` will override values in `environmentVariablesToInject` 
=> `environmentVariablesToInject` does NOT take precedence

falahat: ah you're correct, I was forgetting that environmentVariablesToInject was preloaded with FIREBASE_WEBAPP_CONFIG (potentially). I refactored it to be more clear and to inject these variables when we create the environmentVariablesToInject. Now it's more clear the order of priorities:

injected vars get overridden by the emulator env vars which are overridden by apphosting.yaml env vars.

PR #10049, opened 2026-03-09
Extract App Hosting YAML configuration loading into a common method

falahat:
### Description

By extracting this code, it's possible to re-use the logic in other parts of our cli code.

### Scenarios Tested

I tested a source deploy of a nextjs app which included secrets, env vars, emulator-specific env vars, and testing for the admin sdk (load an image from google storage bucket.)

I deployed the backend (via source deploy) and verified it worked as expected.

I started an emulator and verified my local app and verified it worked as expected. apphosting.emulator.yaml was respected.

I did run into some friction (permission issues) before I added the correct secret permissions. I also had to create the storage bucket and run application default credentials locally to allow the emulator app to access it.

### Sample Commands

`firebase deploy`

`firebase emulators:start`

### DIFF
--- a/src/appdistribution/options-parser-util.ts
+++ b/src/appdistribution/options-parser-util.ts
@@ -64,7 +64,16 @@ export function getAppName(options: any): string {
   return toAppName(options.app);
 }
 
-export function toAppName(appId: string) {
+const APP_ID_FORMAT = /^\d+:\d+:(android|ios|web):[a-fA-F0-9]+$/;
+
+function validateAppId(appId: string): void {
+  if (!APP_ID_FORMAT.test(appId)) {
+    throw new FirebaseError(`Invalid Firebase app ID: ${appId}`);
+  }
+}
+
+export function toAppName(appId: string): string {
+  validateAppId(appId);
   return `projects/${appId.split(":")[1]}/apps/${appId}`;
 }
 


src/apphosting/config.ts
```diff
@@ -111,6 +112,47 @@ export function load(yamlPath: string): yaml.Document {
...
+ * Loads in apphosting.yaml, apphosting.emulator.yaml & apphosting.local.yaml as an
+ * overriding union. In order to keep apphosting.emulator.yaml safe to commit,
+ * users cannot change a secret environment variable to plaintext.
+ * apphosting.local.yaml can, however, for reverse compatibility, though its existence
+ * will be downplayed and tooling will not assist in creating or managing it.
+ */
+export async function getAppHostingConfiguration(
+  backendDir: string,
+): Promise<AppHostingYamlConfig> {
+  const appHostingConfigPaths = dynamicDispatch.listAppHostingFilesInPath(backendDir);
```
annajowang, line 137: lint warning:
`
'dynamicDispatch' was used before it was defined
`
but builds and tests succeed?

falahat: Ah good catch. I fixed the linter error and did some thorough manual testing of the emulator and source deploys. I added that to the CL description above.

PR #10055, opened 2026-03-10
Update the error message for missing path to be friendlier

christhompsongoogle:
Update the error message for missing path to be friendlier


src/commands/studio-export.ts
```diff
@@ -6,15 +6,19 @@ import * as path from "path";
 import * as experiments from "../experiments";
 import { FirebaseError } from "../error";
 
-export const command = new Command("studio:export <path>")
+export const command = new Command("studio:export [path]")
   .description(
     "Bootstrap Firebase Studio apps for migration to Antigravity. Run on the unzipped folder from the Firebase Studio download.",
   )
   .option("--no-start-agy", "skip starting the Antigravity IDE after migration")
+  .help("sup")
```
joehan: Remove?

joehan (approved):

PR #10056, opened 2026-03-10
Check if apphosting deployment exists. If not, omit apphosting config

christhompsongoogle:
Check if apphosting deployment exists. If not, omit apphosting configs. 

TODO: what if firebase.json exists AND apphosting exists?


src/firebase_studio/migrate.ts
```diff
@@ -454,6 +473,19 @@ export async function uploadSecrets(
     return;
   }
 
+  try {
+    const backendsData = await apphosting.listBackends(projectId, "-");
```
joehan, line 477: This call is duplicated now between here and createFirebaseConfig. Can we dedupe?

PR #10075, opened 2026-03-11
Adding windows support for studio:export

joehan:
### Description
Windows support for studio:export. Tested on a personal Windows machine and confirmed that the export is successful + Antigravity starts up

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
 - Add support for VPC direct connect in GCF 2nd gen (#10033)
 - Added `--only` flag for `emulators:export` (#4033)
 - Added support for custom PostgreSQL schema names in Data Connect. (#9271)
+- When SSR web app features are detected in the `firebase init hosting` flow, offer to switch to App Hosting (#9887)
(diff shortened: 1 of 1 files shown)


src/mcp/tools/index.spec.ts
```diff
@@ -36,7 +36,7 @@ describe("availableTools", () => {
     const loginTool = tools.find((t) => t.mcp.name === "firebase_login");
 
     expect(loginTool).to.exist;
-  }).timeout(2000);
+  }).timeout(4000);
```
joehan, line 39: Unrelated, but this was annoying me on CI

src/firebase_studio/migrate.ts
```diff
@@ -600,11 +619,7 @@ export async function migrate(
   rootPath: string,
   options: MigrateOptions = { startAntigravity: true },
 ): Promise<void> {
-  if (process.platform === "win32") {
-    throw new FirebaseError("Firebase Studio migration is currently not supported on Windows.", {
-      exit: 1,
-    });
-  }
+  // No-op for platform support check now
```
christhompsongoogle: nit: can omit this comment

PR #10077, opened 2026-03-12
Include clientCache field during init

itsrakhil:
Adds clientCache: {} to the init SDK configurations for Web, Swift, Kotlin, and Dart when initialized via the Firebase CLI. This ensures that client-side caching is enabled by default for all newly created SDK configurations.

### DIFF
--- a/src/deploy/apphosting/prepare.spec.ts
+++ b/src/deploy/apphosting/prepare.spec.ts
@@ -8,9 +8,12 @@ import * as devconnect from "../../gcp/devConnect";
 import * as prompt from "../../prompt";
 import { RC } from "../../rc";
 import { Context } from "./args";
+import { FirebaseError } from "../../error";
 import prepare, { getBackendConfigs } from "./prepare";
 import * as localbuilds from "../../apphosting/localbuilds";
 import * as experiments from "../../experiments";
+import * as getProjectNumber from "../../getProjectNumber";
+import * as resourceManager from "../../gcp/resourceManager";
 
 const BASE_OPTS = {
   cwd: "/",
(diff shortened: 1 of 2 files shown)


fredzqm:
Let's add an experiment "fdcrealtime" [here](https://github.com/firebase/firebase-tools/blob/main/src/experiments.ts#L173).

`firebase init dataconnect:sdk` should populate default value of client_cache (on by default).

itsrakhil:
> Let's add an experiment "fdcrealtime" [here](https://github.com/firebase/firebase-tools/blob/main/src/experiments.ts#L173).
> 
> `firebase init dataconnect:sdk` should populate default value of client_cache (on by default).

Done

src/experiments.ts
```diff
@@ -184,6 +184,11 @@ export const ALL_EXPERIMENTS = experiments({
     default: true,
     public: false,
   },
+  fdcrealtime: {
+    shortDescription: "Enable Firebase Data Connect realtime feature.",
+    default: true,
```
fredzqm: [blocking] Default should be `false`.

When we are ready to launch after bugbash, we can flip it to true.

yuchenshi:
I'm also curious -- what syntax does the created / modified YAML file use? `clientCache: {}` or just`clientCache:`? The latter is valid but confusing

itsrakhil:
> Can you include a screenshot of this running?
> 
> In CLI, manual testing is most effective. Attaching screenshot is the easist way to both show that it works and provide context for future readers.
> 
> `cat` the content of connector.yaml for reference.

Yes, added the manual test screenshot here
https://docs.google.com/document/d/1Ngo5rlxKdHWnMFvFfBarZ-AFq8VzxG2FgcqDWwElNOw/edit?tab=t.0

itsrakhil:
> clientCache: {}

It would be adding `clientCache: {}`

itsrakhil:
Tests are in hung state. Closing as PR #10167 with the changes is merged.

PR #10088, opened 2026-03-13
Flutter support for FBS -> AGY migration

christhompsongoogle:
Flutter support for FBS -> AGY migration

### DIFF
--- a/src/firebase_studio/migrate.ts
+++ b/src/firebase_studio/migrate.ts
@@ -27,7 +27,7 @@ interface McpConfig {
   mcpServers: Record<string, McpServerConfig>;
 }
 
-async function setupAntigravityMcpServer(rootPath: string): Promise<void> {
+async function setupAntigravityMcpServer(rootPath: string, appType?: AppType): Promise<void> {
   const mcpConfigDir = path.join(os.homedir(), ".gemini", "antigravity");
   const mcpConfigPath = path.join(mcpConfigDir, "mcp_config.json");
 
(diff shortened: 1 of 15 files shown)


src/firebase_studio/migrate.ts
```diff
@@ -438,23 +444,37 @@ async function createFirebaseConfigs(
...
 
-async function writeAntigravityConfigs(rootPath: string, framework?: string): Promise<void> {
+async function writeAntigravityConfigs(rootPath: string, framework: AppType): Promise<void> {
   // 5. IDE Configs (VS Code / AGY)
   const vscodeDir = path.join(rootPath, ".vscode");
   await fs.mkdir(vscodeDir, { recursive: true });
 
   // Create tasks.json for pre-launch tasks
-  const tasksJson = {
+  const tasksJson: any = {
```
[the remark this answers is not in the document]
christhompsongoogle: I knew you'd say that

src/firebase_studio/migrate.ts
```diff
@@ -438,23 +447,37 @@ async function createFirebaseConfigs(
...
-        label: "npm-install",
-        type: "shell",
-        command: "npm install",
-        problemMatcher: [],
-      },
-    ],
+    tasks: [],
   };
+
+  if (framework === "FLUTTER") {
```
joehan, line 487: This assumes that flutter is already installed. Should we be setting that up for them?

christhompsongoogle: I think the flutter CLI installation needs to be in the workflow, similar to node.js. Mostly because of the non-deterministic nature (and for flutter: difficulty) of installation.

PR #10107, opened 2026-03-16
Delete mcp.json if it exists since this will only be confusing.

christhompsongoogle:
Delete mcp.json if it exists since this will only be confusing.


joehan:
Looks like the tests are actually broken, please fix.

PR #10119, opened 2026-03-17
Add basic 1st gen functions support

inlined:
Adds the Terraform exporter and implements 1st gen functions (not including "advanced" trigger types).

This is a checkpoint to get feedback and more invasive testing will be done in the future. There's some hacks present in the way we run the builder because a separate PR is going to need to update both firebase-functions and firebase-tools to allow injecting TF variables and declaring dependencies on standard variables (so as to not create a variables.tf that asks for more things than are used)

This is not really sufficient yet because I don't have providers declarations. Will come in part 4 where I actually field test this

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ test_output.log
 scripts/emulator-tests/functions/index.js
 yarn.lock
 .npmrc
+*.tsbuildinfo
 
 .DS_Store
 .idea
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Fixed an issue where functions deployments would silently fail (#6989)
--- a/firebase-vscode/CHANGELOG.md
+++ b/firebase-vscode/CHANGELOG.md
@@ -1,5 +1,13 @@
 ## NEXT
 
+## 2.2.3
+
+- Update internal `firebase-tools` dependency to 15.13.0
+
+## 2.2.2
+
+- Update internal `firebase-tools` dependency to 15.12.0
+
 ## 2.2.1
 
 - Update internal `firebase-tools` dependency to 15.10.0
--- a/firebase-vscode/package-lock.json
+++ b/firebase-vscode/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "firebase-dataconnect-vscode",
-  "version": "2.2.1",
+  "version": "2.2.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "firebase-dataconnect-vscode",
-      "version": "2.2.1",
+      "version": "2.2.3",
       "dependencies": {
         "@preact/signals-core": "^1.4.0",
         "@preact/signals-react": "1.3.6",
--- a/firebase-vscode/package.json
+++ b/firebase-vscode/package.json
@@ -4,7 +4,7 @@
   "publisher": "GoogleCloudTools",
   "icon": "./resources/firebase_dataconnect_logo.png",
   "description": "Firebase Data Connect for VSCode",
-  "version": "2.2.1",
+  "version": "2.2.3",
   "engines": {
     "vscode": "^1.69.0"
   },
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,12 +1,12 @@
 {
   "name": "firebase-tools",
-  "version": "15.12.0",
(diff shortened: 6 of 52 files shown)


src/functions/iac/export.ts
```diff
@@ -53,3 +96,101 @@ export async function getInternalIac(
     "functions.yaml": yaml.dump(build),
   };
 }
+
+/**
+ *
+ */
+export async function getTerraformIac(
+  options: any,
```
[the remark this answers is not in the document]
inlined: STFU. The options type isn't accurate and isn't compatible with its actual use.

src/gcp/cloudfunctions.ts
```diff
@@ -738,3 +741,213 @@ export function functionFromEndpoint(
...
+    attributes["environment_variables"] = endpoint.environmentVariables;
+  }
+  if (endpoint.secretEnvironmentVariables) {
+    attributes["secret_environment_variables"] = endpoint.secretEnvironmentVariables.map(
+      (secret) => {
+        return {
+          key: secret.key,
+          secret: secret.secret,
+          // TODO: Where does this get resolved normally?
+          version: "latest",
```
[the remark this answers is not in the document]
inlined: Intentional. Leaving open for now until I look up how to solve it and what those ramifications are (e.g. making more things async)

src/functions/iac/export.spec.ts
```diff
@@ -87,4 +87,42 @@ describe("export", () => {
...
+      };
+
+      const result = await exportIac.getTerraformIac(options, codebase);
+
+      expect(result["variables.tf"]).to.be.a("string");
+      expect(result["main.tf"]).to.be.a("string");
+
+      const mainTf = result["main.tf"];
+      expect(mainTf).to.include('resource "google_cloudfunctions_function" "my_func"');
+      expect(mainTf).to.include('resource "google_cloudfunctions_function_iam_binding" "my_func"');
```
rainshen49, line 124: Could you assert on the entire string, instead of includes, kind of like minion? This will also allow me to see a full example in case there are any missed edges.

src/gcp/cloudfunctions.ts
```diff
@@ -738,3 +741,207 @@ export function functionFromEndpoint(
...
+/**
+ * Create a Terraform IAM binding block for a Cloud Function invoker.
+ */
+export function invokerTerraform(id: string, endpoint: build.Endpoint): tf.Block | null {
+  let members: string[] = [];
+
+  if (!build.isCallableTriggered(endpoint) && !build.isHttpsTriggered(endpoint)) {
+    return null;
+  }
+  if (build.isHttpsTriggered(endpoint) && endpoint.httpsTrigger.invoker) {
```
brittanycho: Was wondering, should we be concerned about callable functions here? It seems they would fail this second if statement?

src/gcp/cloudfunctions.ts
```diff
@@ -738,3 +741,207 @@ export function functionFromEndpoint(
...
+  if (endpoint.environmentVariables) {
+    attributes["environment_variables"] = endpoint.environmentVariables;
+  }
+  if (endpoint.secretEnvironmentVariables) {
+    attributes["secret_environment_variables"] = endpoint.secretEnvironmentVariables.map(
+      (secret) => {
+        return {
+          key: secret.key,
+          secret: secret.secret,
+          // TODO: Where does this get resolved normally?
```
brittanycho, line 840: Do we need projectId here too? (re:SecretEnvVar in functions/build.ts)?

src/gcp/cloudfunctions.ts
inlined: Yeah I think this is wrong. Will look.

src/gcp/cloudfunctions.ts
inlined: Good catch.

inlined:
@rainshen49 I added a new upstream branch to improve pretty printing. Then I tinkered with this (where I did do some additional pretty printing support to align equals signs) and made tests check full strings now that they're deterministic. I also added an --out <dir> option to the command so I could check against `terraform plan` and after a quick bugfix (event_trigger is a sub-block, not an attribute) the planner is happy.

src/functions/iac/export.spec.ts
inlined: Did a bunch of work to create good looking formatted strings. Take another look.

PR #10134, opened 2026-03-20
Fix storage emulator crash on invalid JWT and add unit tests

kevmoo:
### Description

The emulator crashes when it gets a malformed or missing auth token

### Scenarios Tested

The auth code returns `null` (instead of crashing) when the token is missing or invalid


kevmoo:
The CI failure is unrelated - fixed in https://github.com/firebase/firebase-tools/pull/10141

PR #10140, opened 2026-03-20
Fix storage emulator crash on invalid JWT and add unit tests

brittanycho:
Branched from https://github.com/firebase/firebase-tools/pull/10134 to unblock stalled tests

See further context in PR #10134


brittanycho:
Actually this PR is not necessary, hub + git push should allow the tests to run on the original PR - closing

PR #10141, opened 2026-03-20
fix(storage): catch unhandled promise rejection in rules watcher

kevmoo:
### Description
Resolves a flaky test issue in the storage emulator integration tests ('Storage Rules Manager should load multiple rulesets on start').
The test was crashing asynchronously due to an ENOTDIR error when reading '/dev/null/storage.rules'. This occurred because chokidar would occasionally fire a 'change' event for the mock file, triggering a synchronous `readFile` call inside an unawaited async event listener. The resulting unhandled promise rejection would crash the Node process and cause the test to fail with 'done() called multiple times'.
This commit wraps the `readFile` call in a try...catch to suppress the error and safely log it as a debug warning instead of crashing.

### Scenarios Tested
- npm run test:storage-emulator-integration

### Sample Commands
- none

### DIFF
--- a/src/emulator/storage/rules/manager.ts
+++ b/src/emulator/storage/rules/manager.ts
@@ -99,11 +99,12 @@ class DefaultStorageRulesManager implements StorageRulesManager {
             "Change detected, updating rules for Cloud Storage...",
           );
           await this.loadRuleset();
-        } catch (e: any) {
+        } catch (e) {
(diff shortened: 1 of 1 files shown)


kevmoo:
see https://github.com/firebase/firebase-tools/actions/runs/23358674709/job/67956761803?pr=10134#step:10:1065

PR #10142, opened 2026-03-20
Migrate to use SWC

maneesht:
Replaced tsc with SWC.


maneesht:
Closing as SWC doesn't do type checking.

PR #10144, opened 2026-03-20
Add RPC library for AI Logic triggers

inlined:


src/functions/events/v2.ts
```diff
@@ -59,3 +59,29 @@ export const CONVERTABLE_EVENTS: Partial<Record<Event, Event>> = {
   "google.cloud.firestore.document.v1.written":
     "google.cloud.firestore.document.v1.written.withAuthContext",
 };
+
+export const AI_LOGIC_BEFORE_GENERATGE_CONTENT =
```
brittanycho: nit: looks like a typo is here

src/functions/events/v2.ts
```diff
@@ -59,3 +59,29 @@ export const CONVERTABLE_EVENTS: Partial<Record<Event, Event>> = {
...
+
+export const AI_LOGIC_EVENTS = [
+  AI_LOGIC_BEFORE_GENERATGE_CONTENT,
+  AI_LOGIC_AFTER_GENERATE_CONTENT,
+] as const;
+
+export const isAILogicEventType = (
+  eventType: string,
+): eventType is (typeof AI_LOGIC_EVENTS)[number] => {
+  return !(eventType in AI_LOGIC_EVENTS);
```
brittanycho: nit: wondering if we should use `.includes()` rather than `in` with the array here?: https://stackoverflow.com/questions/48499181/what-is-the-difference-between-in-operator-and-includes-for-javascript-arr#:~:text=Array%20includes%20vs.%20in%20operator:%20The%20JavaScript,index)%20contains%20a%20specific%20*key*%20or%20*property*.

src/functions/events/v2.ts
brittanycho: do we need to add AI_LOGIC_EVENTS here?

brittanycho (approved):
lgtm other than comments below and assuming all tests pass

src/functions/events/v2.ts
inlined: You're absolutely right. I'm used to using "in" for maps, but that's not what this is.

PR #10145, opened 2026-03-21
Organize terraform attributes when serializing

inlined:
Makes output look more natural and creates a deterministic result for tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,2 @@
+- Moved MCP server `firebase-debug.log` to `~/.cache/firebase/firebase-debug.log`. (#9982)
 - Added a prompt to `firebase init` to install Agent Skills for Firebase.
--- a/src/bin/mcp.ts
+++ b/src/bin/mcp.ts
@@ -1,6 +1,8 @@
 #!/usr/bin/env node
 
-import { resolve } from "path";
+import { resolve, join } from "path";
+import { mkdir } from "fs/promises";
+import { homedir } from "os";
 import { parseArgs } from "util";
 import { useFileLogger } from "../logger";
 import { FirebaseMcpServer } from "../mcp/index";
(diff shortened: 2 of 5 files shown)


src/functions/iac/terraform.ts
```diff
@@ -128,10 +128,102 @@ export function serializeValue(value: Value, indentation = 0): string {
   throw new FirebaseError(`Unsupported terraform value type ${typeof value}`, { exit: 1 });
 }
 
+/**
```
rainshen49: This comment seems out of place?

src/functions/iac/terraform.ts
```diff
@@ -128,10 +128,102 @@ export function serializeValue(value: Value, indentation = 0): string {
...
+    if (META_ARGUMENTS.has(k)) {
+      metaGroup.push(serialized);
+    } else if (nonBlockParams.has(k)) {
+      nonBlockGroup.push(serialized);
+    } else {
+      blockGroup.push(serialized);
+    }
+  }
+
+  const nonemptyGroups = [metaGroup, nonBlockGroup, blockGroup].filter((g) => g.length > 0);
```
rainshen49: IIUC, this will put all meta arguments at the front, right? Based on https://developer.hashicorp.com/terraform/language/style#resource-order, not all meta arguments are equal.
1. `count` or `for_each` goes first
2. resource specific flat fields
3. resource specific block fields
4. `lifecycle` meta block
5. `depends_on` meta block

src/functions/iac/terraform.ts
rainshen49, line 101: Optional: If there are nested blocks, IMO it's implied that the field ordering within a block should also follow the style guide of flat fields first then block fields.

inlined: Nested block support comes in the updated version of the next branch. I've added a bit of that code back here (e.g. indentation in blockToString) but I'll finish the ordering changes in the next branch to save some merge headache

PR #10149, opened 2026-03-22
Registration for FAL events

inlined:
Define FAL blocking triggers and wire them up with the hooks necessary for their triggers to be registered.

Since the backend is currently being developed, this is only unit tested and not yet integration tested.

### DIFF
--- a/src/experiments.ts
+++ b/src/experiments.ts
@@ -193,6 +193,11 @@ export const ALL_EXPERIMENTS = experiments({
     default: true,
     public: false,
   },
+  fdcrealtime: {
+    shortDescription: "Enable Firebase Data Connect realtime feature.",
+    default: false,
+    public: false,
+  },
 });
 
 export type ExperimentName = keyof typeof ALL_EXPERIMENTS;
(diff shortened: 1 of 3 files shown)


src/deploy/functions/services/ailogic.ts
```diff
@@ -0,0 +1,95 @@
+import * as backend from "../backend";
+import { FirebaseError } from "../../../error";
+import { Name, Service } from "./index";
+import * as ailogicApi from "../../../gcp/ailogic";
+import { AI_LOGIC_EVENTS_TO_TRIGGER } from "../../../gcp/ailogic";
+
+export const AI_LOGIC_BEFORE_GENERATE_CONTENT =
+  "firebase.vertexai.v1beta.beforeGenerateContent" as const;
+export const AI_LOGIC_AFTER_GENERATE_CONTENT =
+  "firebase.vertexai.v1beta.afterGenerateContent" as const;
```
brittanycho, line 9: optional nit: I think I noticed the same constants in v2.ts - would it make sense to add these kind of const in a centralized file in the future?

brittanycho (approved):

src/deploy/functions/services/ailogic.ts
inlined: IIRC I moved them in this PR to better separate what was logically AI logic, what is part of the GCP API, etc.

inlined: Nope. Wil fix

PR #10154, opened 2026-03-23
feat(mcp): respect FIREBASE_DEBUG_FILE env var in MCP and CLI

kukgini:
## Problem
 
`firebase mcp` (and the CLI) always writes `firebase-debug.log` to the
current working directory, ignoring the `FIREBASE_DEBUG_FILE` environment
variable. Users who set `FIREBASE_DEBUG_FILE=/dev/null` to suppress the
log file are surprised to still find it being created.
 
The root cause: `useFileLogger()` in both `mcp.ts` and `cli.ts` is called
with no argument, so `findAvailableLogFile()` always falls back to
`process.cwd()/firebase-debug.log` regardless of any environment variable.
 
## Fix
 
- Pass `process.env.FIREBASE_DEBUG_FILE` to `useFileLogger()` in both
  `mcp.ts` and `cli.ts`.
- Guard the `unlinkSync` in `cli.ts` against `/dev/null` to avoid a
  permissions error when the log is suppressed.
 
## Usage
 
```sh
# Suppress debug log entirely
FIREBASE_DEBUG_FILE=/dev/null firebase mcp ...
 
# Or redirect to a fixed path
FIREBASE_DEBUG_FILE=/tmp/firebase.log firebase mcp ...


kukgini:
Closing this PR, since the underlying problem has since been addressed upstream:

- #10175 moved MCP debug logs to `~/.cache/firebase/`, so they no longer pollute the working directory.
- #11023 added the `FIREBASE_DEBUG_PATH` environment variable for customizing the debug log location in both the CLI and the MCP server.

Together these cover the use case this PR was aiming at, so there is no need to keep it open. Thanks!

PR #10163, opened 2026-03-23
Support env vars for local/zip deploys

falahat:

### DIFF
--- a/src/apphosting/config.spec.ts
+++ b/src/apphosting/config.spec.ts
@@ -483,6 +483,8 @@ env:
 
       const { build, runtime } = config.splitEnvVars(env);
 
+      expect(build["STR"].value).to.equal("string");
+      expect(build["NUM"].value).to.equal("12345");
       expect(build["BUILD_AND_RUNTIME_NUM"].value).to.equal("67890");
       expect(runtime).to.deep.include({ variable: "STR", value: "string" });
       expect(runtime).to.deep.include({ variable: "NUM", value: "12345" });
--- a/src/apphosting/config.ts
+++ b/src/apphosting/config.ts
@@ -406,7 +406,7 @@ export function splitEnvVars(env: EnvMap): { build: EnvMap; runtime: Env[] } {
       envVal.value = String(envVal.value);
     }
 
-    if (val.availability?.includes("BUILD")) {
+    if (val.availability?.includes("BUILD") || !val.availability) {
       build[key] = envVal;
     }
     if (val.availability?.includes("RUNTIME") || !val.availability) {


src/deploy/apphosting/prepare.ts
```diff
@@ -170,6 +225,7 @@ export default async function (context: Context, options: Options): Promise<void
         buildDir: outputFiles[0],
         buildConfig,
         annotations,
+        env: runtimeEnv[cfg.backendId] || [],
```
falahat: this env is redundant. buildConfig has a buildConfig.env field

src/deploy/apphosting/release.ts
```diff
@@ -50,7 +50,14 @@ export default async function (context: Context, options: Options): Promise<void
...
       location: context.backendLocations[backendId],
       buildInput: {
-        config: context.backendLocalBuilds[backendId]?.buildConfig,
+        // TODO: Confirm this config behavior
+        config: {
+          ...context.backendLocalBuilds[backendId]?.buildConfig,
+          env: [
+            ...(context.backendLocalBuilds[backendId]?.buildConfig?.env || []),
+            ...(context.backendLocalBuilds[backendId]?.env || []),
+          ],
```
falahat: This block is suspect. BuildConfig already has environments, we shouldn't need to override .env here.  Also, why do we need `context.backendLocalBuilds[backendId]?.buildConfig?.env` and `context.backendLocalBuilds[backendId]?.env` both?

src/deploy/apphosting/prepare.ts
```diff
@@ -155,10 +187,32 @@ export default async function (context: Context, options: Options): Promise<void
...
+      try {
+        const webappConfig = (await managementApps.getAppConfig(
+          backend.appId,
+          managementApps.AppPlatform.WEB,
+        )) as WebConfig;
+        const autoinitVars = getAutoinitEnvVars(webappConfig);
+        buildEnv[cfg.backendId] = {
+          ...buildEnv[cfg.backendId],
+          ...Object.fromEntries(
+            Object.entries(autoinitVars).map(([key, value]) => [key, { value }]),
```
falahat: is this the most concise way of injecting the autoInitEnvVars?

falahat:
After my recent changes, the env vars stopped working. I suspect it's because of this:

I removed the "env" field from LocalBuild, because i assumed it was redundant. After all, we already have LocalBuild.BuildConfig.Env. Therefore, I just put the env vars into this field. 

However, I wonder if BuildConfig.Env is only for build-level configs.

PR #10165, opened 2026-03-23
Add P4SA Flow for Zip Deploy / Local Build

falahat:
### Scenarios Tested

Test plan:
* Remove the permission, and do a build without it to confirm it fails
* Then install this PR/git branch and run the deployment and verify that it auto-adds the permission AND the rollout is successful
* Verify that a source deployment is successful and NOT changed by this code
* Verify that this code is NOT triggered if the apphostinglocalbuilds experiment is disabled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Added a prompt to `firebase init` to install Agent Skills for Firebase.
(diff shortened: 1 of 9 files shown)


falahat:
Fails with:

> Unable to verify App Hosting service agent permissions for service-394522926746@gcp-sa-firebaseapphosting.iam.gserviceaccount.com

I need to confirm the name is correct

falahat:
Actually I just needed to show "Google Created Roles" in pantheon and I was able to see the correct user. I also verified that the user role (bucket viewer permission) was added when I ran a deployment!

PR #10167, opened 2026-03-24
Include empty clientCache configs on init

itsrakhil:
Adds `clientCache: {}` to the init SDK configurations for Web, Swift, Kotlin, and Dart when initialized via the Firebase CLI. This ensures that client-side caching is enabled by default for all newly created SDK configurations.


fredzqm:
Wait, looks like I lost owner permission to the repos. Emm

PR #10177, opened 2026-03-25
Revert "Throttle CRF to two concurrent deploys"

asperpharma:
Reverts firebase/firebase-tools#10173


kevmoo:
Did it break something?

inlined:
This is a modification to an unreleased feature. Unless there's an actual bug to report, we are closing the PR.

PR #10180, opened 2026-03-25
Fix race condition whith multiple rewrites to the same service

inlined:
EDIT: Turns out to not be the root cause of a previously linked bug. Still good to do, but not a high priority.

Currently if there are more than one rewrite to the same service then ensureLatestRevisionTagged gets the same service more than once. Adding this deduplication logic avoids simultaneous edits to the same service

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,4 @@
 - Fixed an issue where functions deployments would silently fail (#6989)
+- Add support for Next.js 16 middleware (`proxy.ts`/`proxy.js`) (#9631)
 - Updates the default region for new App Hosting backends to us-east4 (#10271)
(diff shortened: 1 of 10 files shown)


src/hosting/runTags.ts
```diff
@@ -87,25 +87,21 @@ export async function setRewriteTags(
...
-  const services: run.Service[] = await Promise.all(
-    rewrites
-      .map((rewrite) => {
-        if (!("run" in rewrite)) {
-          return null;
-        }
-        if (rewrite.run.tag !== TODO_TAG_NAME) {
-          return null;
-        }
+  // Note: Deduplication is necessary to avoid race conditions when there are mutiple
```
bkendall: This sentence feels unfinished.

PR #10192, opened 2026-03-26
Support Environment Variables for Local Builds

falahat:
### Description

Adds support for Environment Variables (from apphosting.yaml files, for example) and AutoInit Env Vars for Local Builds.
This correctly splits Build vs Runtime env vars. It passes build env vars to the build process and passes runtime env vars to the control plane when creating the build. It injects the SDK AutoInit env vars into both.

### Scenarios Tested

* I created a local build backend with ABIU enabled and the nodejs22 runtime
* I set set the localBuild flag to true in firebase.json:
```
{
  "apphosting": {
    "backendId": "local-build",
    "rootDir": "/",
    "localBuild": true,
    "ignore": [
      "node_modules",
      ".git",
      "firebase-debug.log",
      "firebase-debug.*.log",
      "functions"
    ]
  }
}
```
* I ran `firebase deploy`
* I verified the app was built locally and deployed as expected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,0 @@
-- Added a prompt to `firebase init` to install Agent Skills for Firebase.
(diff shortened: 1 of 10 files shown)


falahat:
I have verified this one last time after our fixes.

PR #10228, opened 2026-03-31
fix: detect next/image usage in client components

leoortizz:
## Description

Thanks to @dxdc for reporting this in #10214.

When every file that imports `next/image` is a `"use client"` component, `isUsingImageOptimization()` fails to detect image usage because:

- **Tier 1 (`export-marker.json`)**: `isNextImageImported` is only set for Pages Router pages — always `false` in App Router-only projects.
- **Tier 2 (`client-reference-manifest`)**: The Image component never crosses a server→client RSC boundary, so it's never registered in the manifest.

This adds a fallback that scans prerendered HTML files under `.next/server/app/` for the `data-nimg="` attribute that `next/image` renders on every `<img` it produces. This attribute has been stable since Next.js 11.1, covering all supported versions (12–16). The check only runs when both existing tiers fail, so it's fully backward compatible.

Also refactors `isUsingNextImageInAppDirectory` into two explicit sub-checks:
- `isUsingNextImageInServerComponent` — scans `client-reference-manifest.js` (existing logic)
- `isUsingNextImageInClientComponent` — scans prerendered HTML for `data-nimg="` (new)

## Scenarios Tested

- App Router project where `next/image` is used in Server Components (Tier 2 detects it)
- App Router project where **all** `next/image` imports are in `"use client"` components (new Tier 3 detects it via `data-nimg="`)
- App Router project with `images.unoptimized: true` — correctly returns `false` even when `data-nimg="` is present
- App Router project with no `next/image` usage — correctly returns `false`
- Pages Router project — existing Tier 1 (`export-marker.json`) still works

Verified against a [Next.js 15.5.14 test app](https://github.com/monogramdesign/firebase-tools-test-apps/tree/main/nextjs/15.5.x) with `"use client"` on all image-importing components.

## Sample Commands

```bash
firebase deploy
```
```bash
npx mocha --require ts-node/register src/frameworks/next/utils.spec.ts
```

### DIFF
--- a/src/frameworks/next/utils.ts
+++ b/src/frameworks/next/utils.ts
@@ -23,6 +23,8 @@ import type {
   AppPathRoutesManifest,
   ActionManifest,
   NextConfigFileName,
+  FunctionsConfigManifest,
+  MiddlewareManifestV3,
 } from "./interfaces";
 import {
   APP_PATH_ROUTES_MANIFEST,
@@ -32,6 +34,7 @@ import {
   WEBPACK_LAYERS,
   CONFIG_FILES,
   ESBUILD_VERSION,
+  FUNCTIONS_CONFIG_MANIFEST,
 } from "./constants";
 import { dirExistsSync, fileExistsSync } from "../../fsutils";
 import { IS_WINDOWS } from "../../utils";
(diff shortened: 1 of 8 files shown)


src/frameworks/next/utils.ts
```diff
@@ -247,9 +245,22 @@ export async function isUsingImageOptimization(
...
 export async function isUsingNextImageInAppDirectory(
+  projectDir: string,
+  distDir: string,
+): Promise<boolean> {
+  if (await isUsingNextImageInServerComponent(projectDir, distDir)) {
+    return true;
+  }
+  return isUsingNextImageInClientComponent(projectDir, distDir);
+}
+
```
annajowang, line 282: non-blocking nit: can this just be replaced by an in-line  
```
isUsingNextImageInServerComponent(..) || isUsingNextImageInClientComponent(...)
```

Issue #6989, opened 2024-04-11
firebase deploy returns an exit code == 0 even if a function failed to be deployed


Philmod:
### [REQUIRED] Environment info

Version: 13.7.1

OS: Ubuntu

### [REQUIRED] Test case

Deployment with a failing function update.

### [REQUIRED] Steps to reproduce

`firebase deploy` returns an exit code of `0` even if there is a failure like: 

```
# firebase deploy
⚠  functions: failed to update function projects/philmod-godds/locations/us-central1/functions/authentication-delete
Failed to update function projects/philmod-godds/locations/us-central1/functions/authentication-delete

# echo $?
0
```

### [REQUIRED] Expected behavior

Returns an exit code != 0 if it failed.

### [REQUIRED] Actual behavior

Return an exit code == 0 even in case of a failure.

Philmod:
> This issue does not seem to follow the issue template. Make sure you provide all the required information.

fixed

aalej:
Hey @Philmod, thanks for filing this issue. I’m currently unable to reproduce the behavior you mentioned. I tried `firebase deploy` using the ff:

functions/index.js
```js
const functions = require("firebase-functions/v1");
const logger = require("firebase-functions/logger");

exports.authDelete = functions.auth.user().onDelete((user) => {
 logger.log("---- user1 ----");
 logger.debug(user);
});
```

functions/.env - setting `PATH` causes the function update to fail
```env
PATH="some/path"
```

And an error is being raised, which returns a non-zero exit code(2).
```shell
i  functions: updating Node.js 18 (1st Gen) function authDelete(us-central1)...
⚠  functions: failed to update function projects/PROJECT_ID/locations/us-central1/functions/authDelete
Failed to update function projects/PROJECT_ID/locations/us-central1/functions/authDelete

Functions deploy had errors with the following functions:
        authDelete(us-central1)
i  functions: cleaning up build files...

Error: There was an error deploying functions

Having trouble? Try again or contact support with contents of firebase-debug.log
USER$ echo $?
2
```

To get a better understanding of what’s causing the failed update to not raise an error, could you provide code snippets of the functions you’re trying to deploy, and additional details on your project’s setup such as env variables being set?

Philmod:
Hey @aalej,

I'm using Google Cloud Build to deploy the firebase functions, using this [docker image](https://github.com/GoogleCloudPlatform/cloud-builders-community/tree/master/firebase).
[...]

PR #10253, opened 2026-04-02
fixing issue 6989

shettyvarun268:
Fixes: #6989 

Deployments can falsely exit with success code 0 even when internal tasks fail. Right before shutting down, the CLI tries to send background analytics. In secure or restrictive CI/CD environments, this analytics ping hangs and times out. That timeout crashes the CLI wrapper's error-handler prematurely, masking the original deployment failure before it can set the non-zero exit code.

Solution: We wrapped the background analytics timeouts in protective try...catch blocks. Now, if the telemetry tracker hangs or fails to reach the portals, the CLI simply ignores the analytics failure and continues with its shutdown routine. This ensures the CLI always returns the authentic deployment failure code to the host system.

Did not add new unit tests for this specific logic because the wrapper lives at the top-level process exit boundary. Writing automated tests here requires faking CLI process shutdowns and restricted networks, which creates brittle mocks. Verified the logic safety using an isolated sandbox playground and confirmed all existing test trees continue to pass.

### DIFF
--- a/firebase-vscode/CHANGELOG.md
+++ b/firebase-vscode/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## NEXT
 
+## 2.2.3
+
+- Update internal `firebase-tools` dependency to 15.13.0
+
 ## 2.2.2
 
 - Update internal `firebase-tools` dependency to 15.12.0
--- a/firebase-vscode/package-lock.json
+++ b/firebase-vscode/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "firebase-dataconnect-vscode",
-  "version": "2.2.2",
+  "version": "2.2.3",
   "lockfileVersion": 3,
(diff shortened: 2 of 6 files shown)


src/command.ts
```diff
@@ -236,28 +236,28 @@ export class Command {
...
+              tracks.push(
                 trackEmulator("command_success", {
                   command_name: this.name,
                   duration,
                 }),
-              ]),
-            );
+              );
+            }
+            await withTimeout(5000, Promise.all(tracks));
```
joehan: Nit: 5s is insanely long to wait for these calls, probably safe to lower this to 1s - same for the other one.

joehan (approved):
LGTM, but please add a CHANGELOG entry saying something like:
```
- Fixed an issue where functions deployments would silently fail (#6989)
```

PR #10258, opened 2026-04-03
feat: add SSE mode support for MCP server

joehan:
Adds support for running the MCP server in SSE (HTTP) mode. This opens up some interesting use cases, but mostly it makes it much easier to test out the MCP apps work that is incoming.

I also opened up the typing to support structured content returned from tools - minor, but makes ti much easier to have apps read tool responses later.

### DIFF
--- a/src/bin/mcp.ts
+++ b/src/bin/mcp.ts
@@ -49,6 +49,8 @@ Options:
                             If specified, auto-detection is disabled for other features.
   --tools <tools>           Comma-separated list of specific tools to enable. Disables
                             auto-detection entirely.
+  --sse                     Start the server in SSE (HTTP) mode instead of default Stdio.
+  --port <port>             The port to listen on when running in SSE mode (defaults to 3000).
   -h, --help                Show this help message.
 `;
 
@@ -58,6 +60,8 @@ export async function mcp(): Promise<void> {
       only: { type: "string", default: "" },
       tools: { type: "string", default: "" },
       dir: { type: "string" },
+      sse: { type: "boolean", default: false },
+      port: { type: "string", default: "3000" },
       "generate-tool-list": { type: "boolean", default: false },
       "generate-prompt-list": { type: "boolean", default: false },
       "generate-resource-list": { type: "boolean", default: false },
@@ -103,6 +107,9 @@ export async function mcp(): Promise<void> {
     enabledTools,
     projectRoot: values.dir ? resolve(values.dir) : undefined,
   });
-  await server.start();
+  await server.start({
+    useSSE: values.sse,
+    port: values.port ? parseInt(values.port, 10) : undefined,
+  });
   if (process.stdin.isTTY) process.stderr.write(STARTUP_MESSAGE);
 }
--- a/src/bin/mcp.ts
+++ b/src/bin/mcp.ts
@@ -49,7 +49,7 @@ Options:
                             If specified, auto-detection is disabled for other features.
   --tools <tools>           Comma-separated list of specific tools to enable. Disables
                             auto-detection entirely.
-  --sse                     Start the server in SSE (HTTP) mode instead of default Stdio.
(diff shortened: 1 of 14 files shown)


src/bin/mcp.ts
```diff
@@ -58,6 +60,8 @@ export async function mcp(): Promise<void> {
       only: { type: "string", default: "" },
       tools: { type: "string", default: "" },
       dir: { type: "string" },
+      sse: { type: "boolean", default: false },
+      port: { type: "string", default: "3000" },
```
chkuang-g, line 64: Does parseArgs support `type: "integer"`? If so, port should be an integer.

src/bin/mcp.ts
```diff
@@ -58,6 +60,8 @@ export async function mcp(): Promise<void> {
       only: { type: "string", default: "" },
       tools: { type: "string", default: "" },
       dir: { type: "string" },
+      sse: { type: "boolean", default: false },
```
chkuang-g: To make the parameter clear, consider:
```
mode: {type: "string", default: "stdio"}
```

Best if parseArg can validate if the string is within the set of `("stdio", "sse")`

src/mcp/types.ts
```diff
@@ -32,4 +32,5 @@ export interface McpContext {
   rc: RC;
   firebaseCliCommand: string;
   isBillingEnabled: boolean;
+  progressToken?: string | number;
```
chkuang-g, line 35: What is `progressToken`? Is it a common concept for MCP server? If not, some comment here can be helpful

src/mcp/util.ts
```diff
@@ -41,7 +41,8 @@ export function toContent(
   const suffix = options?.contentSuffix || "";
   return {
     content: [{ type: "text", text: `${prefix}${text}${suffix}` }],
-  };
+    structuredContent: data,
```
chkuang-g, line 45: I did not see any change in this PR using `structuredContent`. Is this for a different PR?

joehan: Yeah, unused at the moment. It'll be used in follow ups to make it simpler to parse tool responses from within apps.

src/bin/mcp.ts
joehan: It does not, unfortunately. It's a helper function in the 'node:util' package, so we can't add it either

src/bin/mcp.ts
joehan: Good call.

chkuang-g (commented):
LGTM. But I will let the Firebase CLI and Javascript expert to approve this :)

src/mcp/index.ts
```diff
@@ -488,7 +494,71 @@ export class FirebaseMcpServer {
...
   }
 
-  async start(): Promise<void> {
+  async start(options?: { useSSE?: boolean; port?: number }): Promise<void> {
+    if (options?.useSSE) {
+      const app = express();
+
+      app.use(cors());
+
+      const port = options.port || 3000;
```
maneesht, line 503: nit: should we have the 3000 default be a static variable somewhere? It seems like the default is set above, and I want to makes sure we're always consistent

src/mcp/types.ts
maneesht: I have the same exact question

src/mcp/util.ts
```diff
@@ -41,7 +41,8 @@ export function toContent(
...
+  } as CallToolResult & { structuredContent: any };
```
maneesht, line 46: Should this be `any` or `unknown`? Hoping that we can have a better typing here

src/mcp/types.ts
joehan: https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/progress - built in concept to track LROs

maneesht (approved):

PR #10264, opened 2026-04-03
fix(functions): fix permission denied error for Dart deployments on Windows

lahirumaramba:
### Description
When deploying Dart functions from Windows, the compiled binary `bin/server` is uploaded without the POSIX execute permission (`+x`). This causes the application to fail to start on Cloud Run with a "permission denied" error.
This PR adds support for specifying `executablePaths` when packaging functions, and forces the mode to `0o755` for those files in the generated archive. It automatically detects Dart projects by checking for `pubspec.yaml` and marks `bin/server` as executable.
### Scenarios Tested
- Added unit tests in `prepareFunctionsUpload.spec.ts` to verify that `addFilesToArchive` correctly sets permissions for specified paths.
- Verified that all unit tests pass.

### DIFF
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -238,7 +238,7 @@ export async function prepare(
         ? "tar.gz"
         : "zip";
 
-      const isDart = supported.runtimeIsLanguage(wantBuilds[codebase].runtime, "dart");
+      const isDart = supported.runtimeIsLanguage(wantBuilds[codebase].runtime!, "dart");
       const executablePaths = isDart ? ["bin/server"] : [];
 
       const packagedSource = await prepareFunctionsUpload(


src/deploy/functions/prepare.ts
```diff
@@ -237,13 +239,17 @@ export async function prepare(
       const exportType = backend.someEndpoint(wantBackend, (e) => e.platform === "run")
         ? "tar.gz"
         : "zip";
+      
+      const isDart = fs.existsSync(path.join(sourceDir, "pubspec.yaml"));
```
inlined: Can look at an endpoint being dart I think. All endpoints in a codebase have the same runtime

src/deploy/functions/prepare.ts
```diff
@@ -237,13 +239,17 @@ export async function prepare(
...
+      const executablePaths = isDart ? ["bin/server"] : [];
```
inlined, line 242: This might be something to put in deploy/functions/runtimes?

src/deploy/functions/prepareFunctionsUpload.spec.ts
```diff
@@ -93,4 +95,32 @@ describe("prepareFunctionsUpload", () => {
...
+
+  describe("addFilesToArchive", () => {
+    it("should set mode to 0o755 for executable paths", async () => {
+      const archive = {
+        file: sinon.stub(),
+      } as unknown as any;
+
+      const files = [
+        { name: path.join("src", "index.js"), mode: 0o644 },
+        { name: path.join("src", "bin", "server"), mode: 0o644 },
```
inlined, line 108: Curious if using bin/server above and path.join here means you’ll fail windows tests

lahirumaramba: Added a path normalizer, which I think should fix this

src/deploy/functions/prepare.ts
lahirumaramba: Great point! Changed this to use `backend.someEndpoint(...)`, is that what you meant here?

src/deploy/functions/prepare.ts
lahirumaramba: I think we should as the right long term approach. How do you feel about keeping this fix for now and address that in a follow-up?

lahirumaramba:
@gemini-code-assist review

src/deploy/functions/prepare.ts
```diff
@@ -237,13 +237,17 @@ export async function prepare(
       const exportType = backend.someEndpoint(wantBackend, (e) => e.platform === "run")
         ? "tar.gz"
         : "zip";
+
+      const isDart = backend.someEndpoint(wantBackend, (e) => (e.runtime as string) === "dart");
+      const executablePaths = isDart ? ["bin/server"] : [];
```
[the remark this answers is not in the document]
lahirumaramba: @inlined is this true? :)

src/deploy/functions/prepare.ts
inlined: Sure. Create a bug and assign it to ajp to assign out. It'll probably go to brittany

src/deploy/functions/prepare.ts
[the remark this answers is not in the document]
inlined: Well I'll be darned. There actually is specific runtime language support. import deploy/functions/runtimes and call

```
runtimeIsLanguage(wantBuilds[codebase].runtime, "dart")
```

Runtimes are strings, but they're very strongly typed strings powered by/extracted from a support matrix.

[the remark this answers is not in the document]
inlined: IDK why runtime is nullable but you need to either assert that it exists or add a ! at the end to fix the build and commit

brianquinlan:
With this PR, I am able to deploy from Windows.

PR #10265, opened 2026-04-03
feat: add mcpapps experiment flag and helper

joehan:
Adds `mcpapps` experiment flag and applyAppMeta helper function, which will be used later to omit the `_meta.ui.reourceUri` field when the experiment is disabled.

### DIFF
--- a/src/experiments.ts
+++ b/src/experiments.ts
@@ -24,6 +24,12 @@ export const ALL_EXPERIMENTS = experiments({
     shortDescription: "enables the experiments family of commands",
   },
 
+  mcpapps: {
+    shortDescription: "Enables MCP Apps features",
+    fullDescription: "Enables MCP Apps features, including returning UI resource URIs.",
+    public: true,
+  },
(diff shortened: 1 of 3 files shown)


src/experiments.ts
```diff
@@ -179,6 +179,11 @@ export const ALL_EXPERIMENTS = experiments({
     default: false,
     public: true,
   },
+  mcpapps: {
+    shortDescription: "Enables MCP Apps features",
+    fullDescription: "Enables MCP Apps features, including returning UI resource URIs.",
+    public: true,
```
maneesht, line 185: I noticed `mcpalpha` has a default to `false`, yet it wasn't explicit here. Should it be added for the `mcpapps` feature?

maneesht (approved):
One minor nit

src/experiments.ts
joehan: Nah, the default for 'default' is falsy

PR #10293, opened 2026-04-08
Default function regions are decided much later in prepare

inlined:
This doesn't yet actually _change_ the default from us-central1, but gives us a nice place to make all those changes.

us-central1 is GCP's oldest region and is oversubscribed. People use it because it's the default but it ends up having slower cold-starts and might even see more service disruptions. This is a place where we can examine new functions and decide whether there's a better place to put them. E.g.:

1. As a legacy feature, us-central1 functions are allow to listen to a backend from any region (e.g. an asia-east1 storage bucket with a us-central1 function). We can make the default _match the event source_ and dramatically improve performance
2. For resources in the NAM5 multi-region, it may be better to use us-east1 for faster coldstarts
3. For functions on global resources, we might choose to look at the location defined in the user's admin config or just us-east1 again to move to a less subscribed region.

_Nothing in here is intended to prevent a customer from deploying in us-central1 if they know they want to_. This just helps us modify defaults for customers who have shown no preference.


src/deploy/functions/prepare.ts
```diff
@@ -329,6 +330,82 @@ export async function prepare(
...
+  backend.endpoints[region][endpoint.id] = endpoint;
+  delete backend.endpoints[build.REGION_TBD][endpoint.id];
+  if (Object.keys(backend.endpoints[build.REGION_TBD]).length === 0) {
+    delete backend.endpoints[build.REGION_TBD];
+  }
+}
+
+/**
+ * Verifies that we don't have a peculiar edge case where we cannot know what region a default endpoint was in.
+ * This is only possible in insaine edge cases (esp since you can only have multi-region functions for HTTPS and
```
Berlioz: insaine -> insane

src/deploy/functions/prepare.ts
```diff
@@ -329,6 +330,82 @@ export async function prepare(
...
+ * regional AI Logic functions) where a customer HAD specified multiple regions in a function and then deleted
+ * the regions annotation entirely and we don't know which to delete an which to keep.
```
Berlioz: an -> and

Berlioz (approved):

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 - Fixed an issue where functions deployments would silently fail (#6989)
+- Add foundation for being smarter about where to place functions when the region is not specified (#10293)
```
joehan, line 1: Is this a user facing change? A bit hard to tell from the code without diving into the broader codebase

joehan (approved):

CHANGELOG.md
inlined: It will eventually but isn't yet.

PR #10317, opened 2026-04-10
Function rename deletion sequence change

shettyvarun268:
This pull request resolves a deployment downtime issue in the Cloud Functions release process by refactoring the Fabricator class to implement a globally phased deployment strategy. Previously, function creates, updates, and deletes across all regions were executed in parallel. This behavior caused potential downtime during function renames, as an existing function could be deleted before its replacement was successfully created or updated in another region.

To address this, the execution lifecycle in applyPlan has been split into two distinct global phases separated by a wait barrier. In the first phase, all function creates and updates across all changesets are triggered and awaited simultaneously. In the second phase, deletes are executed only if all operations in the first phase completed successfully. If any create or update operation fails, all pending deletes are aborted, leaving the existing functions running in production and preventing accidental downtime.

The original applyChangeset method was preserved but refactored to accept a phase parameter. This allows it to handle either creations and updates or deletions depending on the current phase of execution directed by applyPlan. Additionally, a reusable wrapOperation helper was extracted to eliminate code duplication across operations and ensure uniform timing and error handling, making the orchestration logic easier to read and maintain.


src/deploy/functions/release/fabricator.spec.ts
```diff
@@ -1712,6 +1715,40 @@ describe("Fabricator", () => {
...
+          endpointsToCreate: [],
+          endpointsToUpdate: [],
+          endpointsToDelete: [ep2],
+          endpointsToSkip: [],
+        },
+      };
+
+      let createFinished = false;
+      const createEndpoint = sinon.stub(fab, "createEndpoint").callsFake(async () => {
+        await new Promise((resolve) => setTimeout(resolve, 100));
```
inlined: We shouldn't be sleeping in tests. If you need to add ordering dependencies. you can have this be

firstThingtThatShouldResovle.then(resolve).

Each resolution happens in a different iteration of the runloop. Otherwise sinon has ways to verify ordering.

src/deploy/functions/release/fabricator.ts
```diff
@@ -95,87 +95,134 @@ export class Fabricator {
...
-      .filter((r) => r.status === "rejected")
-      .map((r) => (r as utils.PromiseRejectedResult).reason);
-    if (errs.length) {
-      logger.debug(
-        "Fabricator.applyRegionalChanges returned an unhandled exception. This should never happen",
-        JSON.stringify(errs, null, 2),
-      );
+    const createAndUpdateResultsArray = await utils.allSettled(createAndUpdatePromises);
+
+    createAndUpdateResultsArray.forEach((r, i) => {
```
inlined: forEach is frowned upon. You should use `for of` or `map` (if you're seeing "summary.results.push" on every value that sounds like a "map")

Edit: It actually looks like you have an array of promise state of arrays. `reduce` is your functional friend here. If you didn't have the promise state in the middle you could have just used the `flatten` utility.

src/deploy/functions/release/fabricator.ts
```diff
@@ -95,87 +95,134 @@ export class Fabricator {
...
-      try {
-        await fn();
-        this.logOpSuccess(op, endpoint);
-      } catch (err: any) {
-        result.error = err as Error;
+  async applyChangeset(
+    changes: planner.Changeset,
+    scraperV1: SourceTokenScraper | undefined,
+    scraperV2: SourceTokenScraper | undefined,
+    phase: "createsAndUpdates" | "deletes",
```
inlined: 1. nit: the combination of create (insert) and update is typically called "upsert" in the industry.
2. The idea that this is one method with two completely disjoint behaviors depending on the "phase" parameter suggests that this is actually two methods.

src/deploy/functions/release/fabricator.ts
```diff
@@ -95,87 +95,134 @@ export class Fabricator {
...
-      logger.debug(
-        "Fabricator.applyRegionalChanges returned an unhandled exception. This should never happen",
-        JSON.stringify(errs, null, 2),
-      );
+    const createAndUpdateResultsArray = await utils.allSettled(createAndUpdatePromises);
+
+    createAndUpdateResultsArray.forEach((r, i) => {
+      if (r.status === "fulfilled") {
+        summary.results.push(...r.value);
+      } else {
```
inlined: Nit: prefer to return early rather than have an "else" if possible. Pyramid code is hard to follow

src/deploy/functions/release/fabricator.ts
```diff
@@ -95,87 +95,134 @@ export class Fabricator {
...
+      for (const changes of changesets) {
+        for (const endpoint of changes.endpointsToDelete) {
+          summary.results.push({
+            endpoint,
+            durationMs: 0,
+            error: new reporter.AbortedDeploymentError(endpoint),
+          });
+        }
+      }
+    } else {
```
inlined: Again, early exits are your friends, even if it means creating utility functions.

src/deploy/functions/release/fabricator.ts
```diff
@@ -95,87 +95,134 @@ export class Fabricator {
...
+      const deletePromises = changesets.map(async (changes) => {
+        return this.applyChangeset(changes, undefined, undefined, "deletes");
+      });
+      const deleteResultsArray = await utils.allSettled(deletePromises);
+
+      for (const r of deleteResultsArray) {
+        if (r.status === "fulfilled") {
+          summary.results.push(...r.value);
+        } else {
+          logger.debug(
```
inlined: Why should this never happen here but is explicitly handled above?

src/deploy/functions/release/fabricator.ts
```diff
@@ -95,87 +95,134 @@ export class Fabricator {
...
+      }
+    });
+
+    const hasFailures =
+      summary.results.some((r) => r.error) ||
+      createAndUpdateResultsArray.some((r) => r.status === "rejected");
+
+    if (hasFailures) {
+      utils.logLabeledWarning("functions", "Deploys failed. Skipping deletes.");
+      for (const changes of changesets) {
```
inlined: An example of what reduce would look like here:

```
changesets.reduce((accum, changes) => {
  const aborts = changes.endpointsToDelete.map({
    endpoint,
    durationMs: 0,
    error: new reporter.AbortedDeploymentError(endpoint),
  });
  return [...accum, ...aborts];
}, summary.results);
```

It technically is O(N^2) writes, but at this volume nobody cares. It's more important to not have a bunch of nested code.

src/deploy/functions/release/fabricator.ts
```diff
@@ -95,87 +95,134 @@ export class Fabricator {
...
+        }
+        for (const update of changes.endpointsToUpdate) {
+          summary.results.push({ endpoint: update.endpoint, durationMs: 0, error });
+        }
+      }
+    });
+
+    const hasFailures =
+      summary.results.some((r) => r.error) ||
+      createAndUpdateResultsArray.some((r) => r.status === "rejected");
```
inlined: Isn't this the case redundant with the previous line if you keep the code to error out every function in a whole codebase if applyChangeset fails?

inlined (changes requested):
TL;DR:

1. Break a function that does two disjoint things into two functions
2. Decide whether you handle unhandled exceptions in a function that should never throw or not. Feel free to just log it like I had so in the unlikely event we add this sort of "do something important not in the executor" we just have a debug log when the issues flood in
3. I couldn't use reduce in the old code because reduce doesn't take async functions. If you're going to separate running code and inserting into summary.results, reduce is the functional pattern you're looking for. 
4. Try to early exit with the edge case. If you have an if {} else {} I have to keep reading past the end of the else to see if there's any code left that still applies to the if case.

src/deploy/functions/release/fabricator.ts
```diff
@@ -95,87 +95,144 @@ export class Fabricator {
...
+    const changesets = Object.values(plan);
+
+    // Phase 1: Creates and Updates
+    const scraperV1 = new SourceTokenScraper();
+    const scraperV2 = new SourceTokenScraper();
+    const createAndUpdatePromises = changesets.map((changes) => {
+      return this.applyUpserts(changes, scraperV1, scraperV2);
     });
-    const promiseResults = await utils.allSettled(deployChangesets);
+    const createAndUpdateResultsArray = await utils.allSettled(createAndUpdatePromises);
```
inlined: Nit: just realized that we still have utils.allSettled. Promise.allSettled is standard in Node 12 which is now lower than our minimum version.

src/deploy/functions/release/fabricator.ts
shettyvarun268: Removed. That check was redundant since we already map rejections to errors in summary.results

src/deploy/functions/release/fabricator.ts
shettyvarun268: Adopted! I used the reduce pattern you suggested to create the aborted results for skipped deletes. It's much cleaner.

src/deploy/functions/release/fabricator.ts
shettyvarun268: Fixed. With the method split and reduce in place, the code is much cleaner and avoids inconsistent handling.

src/deploy/functions/release/fabricator.ts
shettyvarun268: Handled this as mentioned above.

src/deploy/functions/release/fabricator.ts
shettyvarun268: Done. I refactored applyPlan to use early exits (returning early if there are failures after Phase 1), which removed the large else block and flattened the code.

src/deploy/functions/release/fabricator.ts
shettyvarun268: Done. I split applyChangeset into two specialized methods (applyUpserts and applyDeletes) to remove the disjoint behavior and simplify the signatures.

src/deploy/functions/release/fabricator.ts
shettyvarun268: Done. I refactored the result processing to use reduce to accumulate the results and handle edge cases without side-effects, resulting in cleaner code.

src/deploy/functions/release/fabricator.spec.ts
shettyvarun268: Done. I refactored this test to use a controlled promise that I resolve manually in the test to enforce the ordering without relying on time/sleep.

src/deploy/functions/release/fabricator.ts
```diff
@@ -95,87 +95,144 @@ export class Fabricator {
...
+            durationMs: 0,
+            error,
+          })),
+        ];
+        return [...acc, ...aborts];
+      },
+      [],
+    );
+
+    summary.results.push(...upsertResults);
```
inlined: summary.results is known to be empty before this, right? Can't you just assign to summary.results above?

src/deploy/functions/release/fabricator.ts
```diff
@@ -95,87 +95,144 @@ export class Fabricator {
...
-      .filter((r) => r.status === "rejected")
-      .map((r) => (r as utils.PromiseRejectedResult).reason);
-    if (errs.length) {
+    // Simplify failure check (remove redundant check on createAndUpdateResultsArray)
+    const hasFailures = summary.results.some((r) => r.error);
+
+    if (hasFailures) {
+      utils.logLabeledWarning("functions", "Deploys failed. Skipping deletes.");
+
+      const aborts = changesets.reduce<reporter.DeployResult[]>((accum, changes) => {
```
inlined: summary.results = changesets.reduce(callback, summary.results)

src/deploy/functions/release/fabricator.ts
```diff
@@ -95,87 +95,144 @@ export class Fabricator {
...
+          endpoint,
+          durationMs: 0,
+          error: new reporter.AbortedDeploymentError(endpoint),
+        }));
+        return [...accum, ...currentAborts];
+      }, []);
+
+      summary.results.push(...aborts);
+      summary.totalTime = timer.stop();
+      return summary; // Early exit!
```
inlined: Comments should explain why you do something if it's surprising, your code should be clean enough to understand what you're doing. It should be clear enough that you're exiting early when reading a return statement.

src/deploy/functions/release/fabricator.ts
```diff
@@ -95,87 +95,144 @@ export class Fabricator {
...
+    }
+
+    // Phase 2: Deletes
+    const deletePromises = changesets.map((changes) => this.applyDeletes(changes));
+    const deleteResultsArray = await utils.allSettled(deletePromises);
```
inlined: Again Promise.allSettled. You can also probably do this in one map

src/deploy/functions/release/fabricator.ts
```diff
@@ -95,87 +95,144 @@ export class Fabricator {
...
+        return [...acc, ...r.value];
+      }
       logger.debug(
-        "Fabricator.applyRegionalChanges returned an unhandled exception. This should never happen",
-        JSON.stringify(errs, null, 2),
+        "Fabricator.applyDeletes returned an unhandled exception. This should never happen",
+        JSON.stringify(r.reason, null, 2),
       );
-    }
+      return acc;
```
inlined, line 154: 1. As an FYI, I personally prefer to handle errors and make the success case the standard case and the error the early exit, but when success is so much less work than errors, it makes sense
2. Please be consistent on whether you're converting the "never going to happen but let's program defensively" errors results or not between the upsert & delete code. I'm actually wondering if we should _not_ have it because we might actually report something wrong (e.g. we did delete 3  out of 5 functions before throwing and we report that an error occurred and nothing was deleted)

inlined (approved):

PR #10332, opened 2026-04-14
feat: add warnings when using experimental triggers in Dart runtime

Lyokone:
### Description

Adding warnings when using experimental triggers in Dart runtime.
Fixes #10305 

### Scenarios Tested

running `emulator:start` and `deploy`

You can see the warning here for firestore triggers.

Screenshot 2026-04-14 at 09 32 43

Screenshot 2026-04-14 at 09 49 03

PR #10338, opened 2026-04-14
feat: add warnings when using experimental triggers in Dart runtime

ajperel:
Addressing comments of #10332


joehan (approved):

src/deploy/functions/prepare.ts
```diff
@@ -541,6 +543,29 @@ export async function loadCodebases(
...
+  }
+
+  const { emulatorOnly, experimental } = classifyNonProductionEndpoints(dartEndpoints);
+  const unsupported = [...emulatorOnly, ...experimental];
+  if (unsupported.length > 0) {
+    logLabeledWarning(
+      "functions",
+      `The following Dart functions use triggers that are not yet supported for production deployment: ${unsupported.map((ep) => ep.id).join(", ")}. ` +
+        "They will be deployed but may not work as expected. " +
+        "See https://github.com/firebase/firebase-functions-dart#status-alpha-v010 for current trigger support.",
```
jhuleatt:
```suggestion
        "See https://github.com/firebase/firebase-functions-dart for current trigger support.",
```

nit: the `status-alpha-v010` anchor link will change every time we change the version, so probably just worth linking to the repo. The support table is at the top of the readme

CHANGELOG.md
```diff
@@ -1,2 +1,3 @@
 - Updated Pub/Sub emulator to version 0.8.30
 - Renamed Data Connect displayed text to SQL Connect (#10270)
+- Improved warning messages when deploying or emulating Dart functions
```
jhuleatt, line 3: nit: I don't think we've mentioned Dart in any other changelogs: https://github.com/firebase/firebase-tools/releases?q=dart&expanded=true

but it's not a big deal to leave this in

jhuleatt (approved):

CHANGELOG.md
ajperel: Fair... we've got the SDK now in our github org so I'm not too concerned with this either. Or just inspecting the code.

PR #10348, opened 2026-04-15
refactor: remove utils.allSettled polyfill in favor of native Promise.allSettled

kyungseopk1m:
## Description

Removes the custom `allSettled` polyfill from `src/utils.ts` and replaces all call sites with the native `Promise.allSettled`.

## Motivation

The polyfill was introduced to support Node.js versions below 12.9.0, where `Promise.allSettled` was not yet available. The source comment made this intent explicit:

```ts
// TODO: delete once min Node version is 12.9.0 or greater
```

The minimum Node.js version requirement is now **>=20.0.0**, as declared in `package.json` (`engines` field) and enforced at runtime in `src/bin/firebase.ts`. This far exceeds the 12.9.0 threshold, making the polyfill dead code.

This is the same pattern as several recent cleanups in the repo (e.g. `cloneDeep` TODO comments also targeting a minimum-version gate).

## Changes

- **Removed** `PromiseFulfilledResult`, `PromiseRejectedResult`, `PromiseResult` type exports and `allSettled` function from `src/utils.ts`
- **Removed** corresponding unit tests for the polyfill from `src/utils.spec.ts`
- **Updated** all 8 call sites across the codebase to use `Promise.allSettled` directly:
  - `src/emulator/functionsEmulator.ts`
  - `src/functions/secrets.ts`
  - `src/commands/ext-dev-deprecate.ts`
  - `src/commands/ext-dev-undeprecate.ts`
  - `src/deploy/functions/validate.ts`
  - `src/deploy/functions/release/fabricator.ts` (×3 call sites)
  - `src/deploy/functions/release/reporter.ts`
- The `PromiseRejectedResult` cast in `fabricator.ts` now resolves against the TypeScript standard lib type rather than the removed local definition — no behavior change

## Testing

- `npx tsc --noEmit` passes with no errors
- Existing tests unaffected; the removed test block covered only the polyfill's own implementation, not any business logic

### DIFF
--- a/src/commands/ext-dev-deprecate.ts
+++ b/src/commands/ext-dev-deprecate.ts
@@ -89,7 +89,7 @@ async function deprecate(
   } else {
     throw new FirebaseError("No extension versions matched the version predicate.");
   }
-  await utils.allSettled(
+  await Promise.allSettled(
     filteredExtensionVersions.map(async (extensionVersion) => {
       await deprecateExtensionVersion(extensionVersion.ref, options.message);
     }),
--- a/src/commands/ext-dev-undeprecate.ts
+++ b/src/commands/ext-dev-undeprecate.ts
@@ -59,7 +59,7 @@ export const command = new Command("ext:dev:undeprecate <extensionRef> <versionP
     } else {
       throw new FirebaseError("No extension versions matched the version predicate.");
     }
-    await utils.allSettled(
+    await Promise.allSettled(
       extensionVersions.map(async (extensionVersion) => {
         await undeprecateExtensionVersion(extensionVersion.ref);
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,3 @@
+- Removed `utils.allSettled` polyfill in favor of native `Promise.allSettled` (requires Node >=20)
```
joehan: No need for a changelog here - this is a non-user facing change.

joehan (approved):

kyungseopk1m:
removed CHANGELOG entry and resolved conflicts with main.

PR #10350, opened 2026-04-15
test: improve coverage for dataconnect names

joehan:
### Description
Add parsing and validation regex checks for Data Connect entities.

### Scenarios Tested
- Extracting groups for service and connectors


src/dataconnect/names.ts
```diff
@@ -80,7 +80,7 @@ export function parseCloudSQLInstanceName(cloudSQLInstanceName: string): cloudSQ
     throw new FirebaseError(`${cloudSQLInstanceName} is not a valid cloudSQL instance name`);
   }
   const toString = () => {
-    return `projects/${projectId}/locations/${location}/services/${instanceId}`;
+    return `projects/${projectId}/locations/${location}/instances/${instanceId}`;
```
aalej, line 83: Just double checking, is this change intended?

note: 

`firebase init dataconnect` and `firebase deploy --only dataconnect` are working fine with the changes

joehan: Yep - this function parses a CSQL instance name, which follows this pattern. AFAICT, it wasn't used anywhere, so the bug was limited, but this fix is correct.

aalej (approved):
LGTM!

PR #10367, opened 2026-04-16
Automatically tag previous image during release

joehan:
### Description
Automating one more chore in the release process - on each release, tag the previous release as no-new-use-public-image. We don't republish new versions of these old images with vulnerabilties addressed. Adding this tag opts the images out of future AutoVM scans and signals to users that they should be moving to latest.

### DIFF
--- a/scripts/publish/cloudbuild.yaml
+++ b/scripts/publish/cloudbuild.yaml
@@ -159,6 +159,27 @@ steps:
           echo "Skipping firepit build for preview version."
         fi
 
+  # Tag the previous version of the docker image
+  - name: "gcr.io/google.com/cloudsdktool/cloud-sdk:slim"
+    entrypoint: "bash"
+    args:
+      - "-c"
+      - |
+        if [ "${_VERSION}" != "preview" ]; then
+          IMAGE_NAME="us-docker.pkg.dev/${_ARTIFACT_REGISTRY_PROJECT}/us/firebase"
+          TAGS=$(gcloud container images list-tags $${IMAGE_NAME} --filter="tags:latest" --format="value(tags)")
+          echo "Tags for latest: $${TAGS}"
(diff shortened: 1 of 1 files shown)


scripts/publish/cloudbuild.yaml
```diff
@@ -159,6 +159,27 @@ steps:
...
+      - "-c"
+      - |
+        if [ "${_VERSION}" != "preview" ]; then
+          IMAGE_NAME="us-docker.pkg.dev/${_ARTIFACT_REGISTRY_PROJECT}/us/firebase"
+          TAGS=$(gcloud container images list-tags $${IMAGE_NAME} --filter="tags:latest" --format="value(tags)")
+          echo "Tags for latest: $${TAGS}"
+          PREVIOUS_VERSION=$(echo $${TAGS} | tr ',' '\n' | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$$' | head -n 1)
+          echo "Detected previous version: $${PREVIOUS_VERSION}"
+          if [ -n "$${PREVIOUS_VERSION}" ]; then
+            gcloud container images add-tag $${IMAGE_NAME}:latest $${IMAGE_NAME}:no-new-use-public-image-$${PREVIOUS_VERSION} --quiet
```
christhompsongoogle, line 175: Discussed: 

Since this runs before the new image, the current image is soon to be the previous image, hence the tagging of IMAGE_NAME and not PREVIOUS_IMAGE_NAME

christhompsongoogle (approved):

scripts/publish/cloudbuild.yaml
joehan: Ahhh - to be clear, image name is always the same (us/firebase by default), we just publish new versions/tags of it.

PR #10374, opened 2026-04-17
fix(functions): isolate SourceTokenScraper per codebase

oudmane:
### Description

**Fixes a regression introduced in v15.15.0** (likely via the `Fabricator` refactor in #10293) where `SourceTokenScraper` instances were moved from a per-changeset scope to a plan-wide scope.

In projects utilizing multiple `codebase` definitions (common in monorepos), each codebase has its own isolated source bundle and ZIP file. Because the `SourceTokenScraper` is stateful and captures optimization tokens from GCF deployment operations, sharing a single scraper across the entire deployment plan caused different codebases to "leak" source tokens to one another.

When **Codebase B** incorrectly uses a source token intended for **Codebase A**, Google Cloud reuses the processed module/container from Codebase A. This results in Codebase B being deployed with the wrong code, leading to runtime failures: `Function '...' is not defined in the provided module`.

This PR restores isolation by maintaining a registry of scrapers keyed by `codebase` within the `Fabricator.applyPlan` method. This ensures that tokens are only shared among functions that actually share the same source (same codebase across different regions), while keeping independent codebases strictly isolated.

### Scenarios Tested

- **Monorepo Batch Deployment**: Tested on a project with 48 independent codebases. Verified that batch deployments (batch size 10) no longer cause functions to be deployed with the wrong source code artifacts.
- **Runtime Validation**: Confirmed that the "Function not defined" error is resolved in the Google Cloud logs for all functions when deployed concurrently.
- **Cross-Region Consistency**: Verified that functions within the *same* codebase (e.g., the same source directory deployed to `us-central1` and `europe-west1`) still correctly share the same scraper instance to maintain existing deployment performance optimizations.

### Sample Commands

Standard deployment command in a monorepo setup:

Example `firebase.json` structure where this fix is critical:
```json
{
  "functions": [
    { "codebase": "app-api", "source": "packages/api/dist" },
    { "codebase": "app-auth", "source": "packages/auth/dist" }
  ]
}
```

### DIFF
--- a/src/deploy/functions/release/fabricator.spec.ts
+++ b/src/deploy/functions/release/fabricator.spec.ts
@@ -1618,6 +1618,46 @@ describe("Fabricator", () => {
       expect(results[0].error).to.be.instanceOf(reporter.DeploymentError);
       expect(results[0].error?.message).to.match(/create function/);
     });
+    it("uses different scrapers for different codebases", async () => {
+      const ep1 = endpoint({ httpsTrigger: {} }, { codebase: "codebaseA" });
+      const ep2 = endpoint({ httpsTrigger: {} }, { codebase: "codebaseB" });
+      const plan: planner.DeploymentPlan = {
+        "us-central1": {
+          endpointsToCreate: [ep1],
+          endpointsToUpdate: [],
+          endpointsToDelete: [],
+          endpointsToSkip: [],
+        },
+        "us-west1": {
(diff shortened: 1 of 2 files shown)


oudmane:
Hey @cabljac, I signed the CLA, do I need to do something else ?

oudmane:
This PR got ignored, but the bug was reported in #10428 and fixed in #10432

PR #10382, opened 2026-04-17
Integrate FAH Local Builds with Universal Maker

falahat:
### Description

This takes the existing local builds solution and uses the Universal Maker binary (which runs all relevant buildpacks) instead of hackily running the apphosting adapter manually.

The Universal Maker is a more well-supported tool and behaves more similar to Cloud Builds so we can have more confidence about framework support and fidelity.

Major changes

* Automatically downloads the universal maker when it's needed, verifying its hash
* Calls the universal maker, correctly handling the .apphosting folder.
* For now, this relies on the apphosting generated bundle.yaml file rather than UM's output.json, but that will be fixed later

### Scenarios Tested

Created a local build with the 
### Sample Commands

```
firebase experiments:enable apphostinglocalbuilds
firebase experiments:enable universalMaker

firebase deploy
```


src/deploy/apphosting/util.ts
```diff
@@ -27,47 +27,93 @@ export async function createLocalBuildTarArchive(
...
+      }
+    }
+    const rootNext = path.join(rootDir, ".next");
+    if (fs.existsSync(rootNext)) {
+      fs.cpSync(rootNext, path.join(tempDir, ".next"), { recursive: true });
+    }
+    const rootNodeModules = path.join(rootDir, "node_modules");
+    if (fs.existsSync(rootNodeModules)) {
+      fs.cpSync(rootNodeModules, path.join(tempDir, "node_modules"), { recursive: true });
+    }
```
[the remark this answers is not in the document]
falahat: for now, this is only meant for nextjs support so that is expected.

src/apphosting/universalMakerDownload.ts
```diff
@@ -0,0 +1,142 @@
...
+    throw new FirebaseError(`Could not find download details for platform: ${platformKey}`);
+  }
+
+  return details;
+}
+
+/**
+ * Checks whether the file at `filepath` has the expected size.
+ */
+function validateSize(filepath: string, expectedSize: number): Promise<void> {
```
joehan: I think this is redundant with validateSize from src/emulators/download.ts - we should move them into a single shared file

src/apphosting/universalMakerDownload.ts
```diff
@@ -0,0 +1,142 @@
...
+            { exit: 1 },
+          ),
+        );
+  });
+}
+
+/**
+ * Checks whether the file at `filepath` has the expected SHA256 checksum.
+ */
+function validateChecksumSHA256(filepath: string, expectedChecksum: string): Promise<void> {
```
joehan: If we make a shared file for utilities, this ought ot go there as well

src/apphosting/localbuilds.ts
```diff
@@ -1,14 +1,181 @@
...
+    if (res.error) {
+      throw res.error;
+    }
+    if (res.status !== 0) {
+      throw new FirebaseError(`Universal Maker failed with exit code ${res.status ?? "unknown"}.`);
+    }
+  } catch (e) {
+    if (e && typeof e === "object" && "code" in e && e.code === "EACCES") {
+      throw new FirebaseError(
+        "Failed to execute the Universal Maker binary due to permission constraints. Please assure you have set chmod +x on your file.",
```
falahat: "Failed to execute the Universal Maker binary due to permission constraints. Please assure you have set chmod +x on your file."

This is too vague. It should be clear that it's the universal maker binary that should have execution permissions. Can we tell the user the filepath?

src/apphosting/localbuilds.ts
```diff
@@ -1,14 +1,181 @@
...
+
+  let umOutput: UniversalMakerOutput;
+  try {
+    umOutput = JSON.parse(outputRaw) as UniversalMakerOutput;
+  } catch (e) {
+    throw new FirebaseError(`Failed to parse build_output.json: ${(e as Error).message}`);
+  }
+
+  let finalRunCommand = `${umOutput.command} ${umOutput.args.join(" ")}`;
+  let finalOutputFiles: string[] = [".apphosting"]; // Fallback
```
falahat: Why do we need this ".apphosting" as a fallback? 

It would be more clear if we just failed if we couldn't find/construct the correct folders.

src/apphosting/localbuilds.ts
```diff
@@ -1,14 +1,181 @@
...
+  args: string[];
+  language: string;
+  runtime: string;
+  envVars?: Record<string, string | number | boolean>;
+}
+
+/**
+ * Runs the Universal Maker binary to build the project.
+ */
+export async function runUniversalMaker(
```
falahat, line 27: the runUniversalMaker method is way too cumbersome and large. We should decompose it into smaller, self-contained helper methods. 

Executing the binary and handling its error results (if it fails) should be a separate method, for example.

Handling all of the output files and configurations from Universal Maker binary can also be split into 1 or more new methods.

src/apphosting/localbuilds.ts
```diff
@@ -62,9 +229,26 @@ export async function localBuild(
     process.env[key] = value;
   }
 
-  let apphostingBuildOutput;
+  let apphostingBuildOutput: AppHostingBuildOutput;
   try {
-    apphostingBuildOutput = await localAppHostingBuild(projectRoot, framework);
+    if (experiments.isEnabled("universalMaker")) {
+      apphostingBuildOutput = await runUniversalMaker(projectRoot, framework);
+      logger.debug(
```
falahat: this debug log can be removed.

src/apphosting/localbuilds.ts
```diff
@@ -82,11 +266,16 @@ export async function localBuild(
...
+      ({ variable, value, availability }) => {
+        const validAvail = availability.filter(
+          (a): a is Availability => a === "BUILD" || a === "RUNTIME",
+        );
+        return {
+          variable,
+          value,
+          availability: validAvail,
+        };
+      },
```
falahat: Confirm that we correctly handle if an env var has invalid availability. Also does it handle lowercase/uppercase correctly? I'd prefer to not have to do this weird filtering in this PR.

src/emulator/download.ts
```diff
@@ -51,8 +50,8 @@ export async function downloadEmulator(name: DownloadableEmulators): Promise<voi
   }
 
   if (!emulator.opts.skipChecksumAndSize) {
-    await validateSize(tmpfile, emulator.opts.expectedSize);
-    await validateChecksum(tmpfile, emulator.opts.expectedChecksum);
+    await downloadUtils.validateSize(tmpfile, emulator.opts.expectedSize);
+    await downloadUtils.validateChecksum(tmpfile, emulator.opts.expectedChecksum, "md5");
```
falahat, line 54: why do we have to explicitly pass "md5" as the hash but didn't before? Confirm that this does NOT change actual behavior.

src/downloadUtils.ts
```diff
@@ -47,3 +48,46 @@ export async function downloadToTmp(remoteUrl: string, auth: boolean = false): P
...
+    return stat.size === expectedSize
+      ? resolve()
+      : reject(
+          new FirebaseError(
+            `download failed, expected ${expectedSize} bytes but got ${stat.size}`,
+            { exit: 1 },
+          ),
+        );
+  });
+}
```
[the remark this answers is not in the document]
falahat: while this is true, I do NOT want to make any non-critical changes. I'd rather keep the existing code/behavior to avoid churn/risk in this PR.

src/emulator/download.ts
falahat: Note: the previous code was hardcoded to md5. Now, we explicitly set it in the function. We also still default to md5, so existing code will NOT change behavior.

src/downloadUtils.ts
```diff
@@ -47,3 +48,46 @@ export async function downloadToTmp(remoteUrl: string, auth: boolean = false): P
 
   return tmpfile.name;
 }
+
+/**
+ * Checks whether the file at `filepath` has the expected size.
+ */
```
falahat, line 54: add if the size is bytes, mb, or what. The current docs are unclear.

src/emulator/download.ts
```diff
@@ -113,7 +112,7 @@ function removeOldFiles(
   for (const file of files) {
     const fullFilePath = path.join(emulator.opts.cacheDir, file);
 
-    if (file.indexOf(emulator.opts.namePrefix) < 0) {
+    if (!file.includes(emulator.opts.namePrefix)) {
```
falahat: I'd rather not make any non-crucial changes.

src/apphosting/universalMakerDownload.ts
```diff
@@ -0,0 +1,103 @@
...
+import * as os from "os";
+import * as path from "path";
+
+import { FirebaseError } from "../error";
+import * as downloadUtils from "../downloadUtils";
+import { logger } from "../logger";
+import * as universalMakerInfo from "./universalMakerInfo.json";
+
+const CACHE_DIR =
+  process.env.FIREBASE_UNIVERSAL_MAKER_PATH ||
```
falahat: I don't think we need an process.env.FIREBASE_UNIVERSAL_MAKER_PATH env var. We should find and remove existing usages of this.

falahat (commented):
Some readability / organization comments. I tested this manually but I re-read the code and think it's quite gross/unreadable currently.

src/apphosting/localbuilds.ts
```diff
@@ -1,14 +1,181 @@
...
+    } catch (e) {
+      // Fall back gracefully if parser fails
+    }
+  }
+
+  return {
+    metadata: {
+      language: umOutput.language,
+      runtime: umOutput.runtime,
+      framework: framework || "nextjs",
```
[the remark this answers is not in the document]
falahat: it's hardcoded for now and that's okay for private preview.

src/apphosting/localbuilds.ts
```diff
@@ -1,14 +1,181 @@
...
+  }
+
+  let umOutput: UniversalMakerOutput;
+  try {
+    umOutput = JSON.parse(outputRaw) as UniversalMakerOutput;
+  } catch (e) {
+    throw new FirebaseError(`Failed to parse build_output.json: ${(e as Error).message}`);
+  }
+
+  let finalRunCommand = `${umOutput.command} ${umOutput.args.join(" ")}`;
```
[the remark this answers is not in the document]
falahat: this is an existing limitation with our system

src/apphosting/localbuilds.ts
```diff
@@ -1,14 +1,181 @@
...
+      ["-application_dir", projectRoot, "-output_dir", projectRoot, "-output_format", "json"],
+      {
+        env: {
+          ...process.env,
+          X_GOOGLE_TARGET_PLATFORM: "fah",
+          FIREBASE_OUTPUT_BUNDLE_DIR: bundleOutput,
+        },
+        stdio: "pipe",
+      },
+    );
```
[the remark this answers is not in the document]
falahat: I'm going to keep the current code to minimize complexity. The current behavior doesn't feel slow/stuck though.

src/downloadUtils.ts
```diff
@@ -47,3 +48,46 @@ export async function downloadToTmp(remoteUrl: string, auth: boolean = false): P
...
+    return stat.size === expectedSize
+      ? resolve()
+      : reject(
+          new FirebaseError(
+            `download failed, expected ${expectedSize} bytes but got ${stat.size}`,
+            { exit: 1 },
+          ),
+        );
+  });
+}
```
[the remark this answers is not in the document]
falahat: don't want to modify this existing method

src/apphosting/localbuilds.ts
```diff
@@ -1,14 +1,210 @@
...
+      fs.renameSync(path.join(bundleOutput, file), dest);
+    }
+    fs.rmdirSync(bundleOutput);
+  }
+
+  let umOutput: UniversalMakerOutput;
+  try {
+    umOutput = JSON.parse(outputRaw) as UniversalMakerOutput;
+  } catch (e) {
+    throw new FirebaseError(`Failed to parse build_output.json: ${(e as Error).message}`);
```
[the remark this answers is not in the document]
falahat: there are error parsing utils, I used that

src/apphosting/localbuilds.ts
```diff
@@ -1,14 +1,206 @@
...
+    } catch (e: unknown) {
+      logger.debug(`Failed to parse bundle.yaml: ${getErrMsg(e)}`);
+    }
+  }
+
+  if (!finalOutputFiles) {
+    throw new FirebaseError(
+      "Failed to resolve build artifacts. Ensure Universal Maker produced a valid bundle.yaml with outputFiles.",
+    );
+  }
```
falahat: can be its own method to make this code readable.

src/apphosting/localbuilds.ts
```diff
@@ -1,14 +1,206 @@
...
+
+      if (bundleData?.runConfig?.runCommand) {
+        finalRunCommand = bundleData.runConfig.runCommand;
+      }
+
+      if (bundleData?.outputFiles?.serverApp?.include) {
+        finalOutputFiles = bundleData.outputFiles.serverApp.include;
+      }
+    } catch (e: unknown) {
+      logger.debug(`Failed to parse bundle.yaml: ${getErrMsg(e)}`);
```
falahat: Also this method itself doesn't need to catch and log errors necessarily. We can move the error handling up to runUniversalMaker so the methods it calls don't have to swallow+log errors.

If we fail to parse bundle.yaml, it's a irrecoverable error, so we shouldn't just swallow this error. We should allow it to bubble up like the other errors so they're handled in runUniversalMaker.

In general, I need to be more clear about the error handling here to know which files return errors.

src/apphosting/localbuilds.ts
```diff
@@ -1,14 +1,206 @@
...
+      framework: framework || "nextjs",
+    },
+    runConfig: {
+      runCommand: finalRunCommand,
+      environmentVariables: Object.entries(umOutput.envVars || {})
+        .filter(([k]) => k !== "FIREBASE_OUTPUT_BUNDLE_DIR")
+        .map(([k, v]) => ({
+          variable: k,
+          value: String(v),
+          availability: ["RUNTIME"],
```
falahat, line 179: i need to verify that the discovered env vars are ONLY the runtime ones.

src/apphosting/universalMakerDownload.ts
```diff
@@ -0,0 +1,101 @@
...
+  } else if (process.platform === "linux") {
+    if (process.arch === "x64") {
+      platformKey = "linux_x64";
+    } else {
+      throw new FirebaseError(
+        "Linux ARM (linux_arm64) is not currently supported for Universal Maker.",
+      );
+    }
+  } else if (process.platform === "win32") {
+    throw new FirebaseError("Windows (win32) is not currently supported for Universal Maker.");
```
Yuangwang, line 42: Do you know when/if these other platforms will be supported?

src/apphosting/universalMakerInfo.json
```diff
@@ -0,0 +1,16 @@
+{
+  "darwin_arm64": {
+    "version": "1.0.0",
+    "expectedSize": 16111618,
+    "expectedChecksumSHA256": "4b77d02a5f80f26d9bd1428f388c293c1fb264995d75b51c7d50fec7c87bcf58",
+    "remoteUrl": "https://artifactregistry.googleapis.com/download/v1/projects/serverless-runtimes-qa/locations/us-central1/repositories/universal-maker/files/darwin-arm64%3A1.0.0%3Auniversal_maker:download?alt=media",
```
Yuangwang, line 6: I'm guessing this is downloading from the serverless-runtimes-qa for now since this is still in preview but do we expect to change this to a prod project in the future?

Yuangwang (approved):
Left some questions but otherwise lgtm

src/apphosting/universalMakerDownload.ts
falahat: oh yes, we discussed it and they're in progress (esp windows)

src/apphosting/universalMakerInfo.json
falahat: I'm not sure but I will follow up

PR #10401, opened 2026-04-21
Cli prompts fixed to respect non interactive flag

tammam-g:
Before file overrides and agent skill prompts would show up even when --non-interactive mode is used. After this change it's possible to call `firebase dataconnect init --non-interactive` and get an initial dataconnect scaffold


src/init/features/dataconnect/index.ts
```diff
@@ -752,16 +769,18 @@ async function promptForCloudSQL(setup: Setup, info: RequiredInfo): Promise<void
...
+          nonInteractive: options.nonInteractive,
         });
       }
     }
   }
 
   if (info.locationId === "") {
-    await promptForLocation(setup, info);
+    await promptForLocation(setup, info, options);
     info.shouldProvisionCSQL = await confirm({
```
fredzqm: Would it be cleaner if we short-circuit the confirm?

```
info.shouldProvisionCSQL = !options.nonInteractive && await confirm({
```

src/init/features/dataconnect/index.ts
```diff
@@ -733,6 +748,8 @@ async function promptForCloudSQL(setup: Setup, info: RequiredInfo): Promise<void
       info.cloudSqlInstanceId = await select<string>({
         message: `Which CloudSQL instance would you like to use?`,
         choices,
+        default: "",
```
fredzqm: In this branch, `if (choices.length) {`, so we already have a CSQL somewhere.

I feel like the most reasonable default behavior is to re-use the existing Cloud SQL instance with a unique database name (already handled below).

So here, let's pass `choices[0]` as default

src/init/features/dataconnect/index.ts
```diff
@@ -682,10 +692,15 @@ async function chooseExistingService(existing: Service[]): Promise<Service | und
     message:
       "Your project already has existing services. Which would you like to set up local files for?",
     choices,
+    nonInteractive: options.nonInteractive,
```
fredzqm, line 697: `nonInteractive=true` would fail if `default` isn't set.

The one-linter setup script from console would set `FDC_SERVICE` to pick a specific FDC service. You can leverage this in your demo env.

If it's not set and it's an non-interactive mode, I feel like it's pretty reasonable to setup the first service in the list.

So pass `choices[0]` here

src/init/features/dataconnect/index.ts
```diff
@@ -154,10 +156,11 @@ export async function askQuestions(setup: Setup): Promise<void> {
             }
             return "Please enter a description for your app idea.";
           },
+          nonInteractive: options.nonInteractive,
```
fredzqm, line 160: Non-interactive mode should never enter the generate schema mode.

Let's revert changes in this branch and add a condition in its entrance condition.

```
if (!info.serviceGql && !options.nonInteractive) {
```

Leave a comment saying in an empty project, `firebase init dataconnect` in non-interactive mode always setup the static movie app template.

fredzqm (changes requested):

src/agentSkills.ts
```diff
@@ -2,11 +2,13 @@ import { spawn, spawnSync } from "child_process";
...
 import * as prompt from "./prompt";
 import { getErrMsg } from "./error";
+import { Options } from "./options";
 
-export async function promptForAgentSkills(): Promise<boolean> {
+export async function promptForAgentSkills(options?: Options): Promise<boolean> {
   return prompt.confirm({
     message: "Would you like to install agent skills for Firebase?",
     default: true,
+    nonInteractive: options?.nonInteractive,
```
fredzqm, line 11: This LGTM, but @joehan should review it as well~

Can you think of any cases that installing agent skills during init is not intentionally? Or cause harm

joehan: This is an interesting one - i think that it is pretty low cost to install skills so i'm ok with it.

src/init/features/dataconnect/index.ts
tammam-g: Done

fredzqm (approved):

src/agentSkills.ts
tammam-g: Actually I changed this to be false in non interactive mode, turns out it could break thinks in some environments, better reduce file system changes in non-interactive mode.

PR #10408, opened 2026-04-22
feat: add warning nudges for cross-region event triggers

shettyvarun268:
This PR implements robust cross-region latency warnings within the Firebase Functions deployment pipeline to nudge developers away from unnecessary transoceanic network hops. When an event-triggered function (such as Firestore or Cloud Storage) is explicitly assigned or defaults to us-central1 while its backing event resource operates in a distinct non-US location, the CLI now issues a user-friendly cautionary warning encouraging explicit collocation with the event trigger's region.

To maintain a clean CLI experience, the PR aggregates multiple offending functions into a single, consolidated warning log to avoid spamming the terminal during large deployments. It correctly exempts US multi-region databases (such as nam5 and nam7) to prevent false-positives, and introduces an opt-out environment flag (FIREBASE_SUPPRESS_REGION_WARNING=true). The flag allows headless CI/CD pipelines to silence repetitive warnings safely, with instructions on how to apply it printed directly inside the warning log.

### DIFF
--- a/src/deploy/functions/triggerRegionHelper.spec.ts
+++ b/src/deploy/functions/triggerRegionHelper.spec.ts
@@ -3,7 +3,13 @@ import * as sinon from "sinon";
 
 import * as backend from "./backend";
 import * as storage from "../../gcp/storage";
+import * as firestore from "../../gcp/firestore";
+import * as database from "../../management/database";
+import * as utils from "../../utils";
 import * as triggerRegionHelper from "./triggerRegionHelper";
+import * as storageService from "./services/storage";
+import * as firestoreService from "./services/firestore";
+import * as databaseService from "./services/database";
 
 const SPEC = {
   region: "us-west1",
@@ -14,14 +20,26 @@ const SPEC = {
 describe("TriggerRegionHelper", () => {
   describe("ensureTriggerRegions", () => {
     let storageStub: sinon.SinonStub;
+    let firestoreStub: sinon.SinonStub;
+    let databaseStub: sinon.SinonStub;
+    let logWarningSpy: sinon.SinonSpy;
 
     beforeEach(() => {
+      storageService.clearCache();
+      firestoreService.clearCache();
+      databaseService.clearCache();
       storageStub = sinon.stub(storage, "getBucket").throws("unexpected call to storage.getBucket");
+      firestoreStub = sinon.stub(firestore, "getDatabase").throws("unexpected call to firestore.getDatabase");
+      databaseStub = sinon.stub(database, "getDatabaseInstanceDetails").throws("unexpected call to database.getDatabaseInstanceDetails");
+      logWarningSpy = sinon.spy(utils, "logLabeledWarning");
     });
 
     afterEach(() => {
       sinon.verifyAndRestore();
+      delete process.env.FIREBASE_SUPPRESS_REGION_WARNING;
     });
+
+    // --- V2 Tests ---
 
     it("should throw an error if we can't find the bucket region", async () => {
       const ep: backend.Endpoint = {
@@ -39,38 +57,6 @@ describe("TriggerRegionHelper", () => {
       await expect(triggerRegionHelper.ensureTriggerRegions(backend.of(ep))).to.be.rejectedWith(
         "Can't find the storage bucket region",
       );
-    });
-
-    it("should skip v1 and callable functions", async () => {
-      const v1EventFn: backend.Endpoint = {
-        id: "v1eventfn",
-        entryPoint: "v1Fn",
-        platform: "gcfv1",
-        eventTrigger: {
-          eventType: "google.storage.object.create",
-          eventFilters: { resource: "projects/_/buckets/my-bucket" },
-          retry: false,
-        },
-        ...SPEC,
-      };
-      const v2CallableFn: backend.Endpoint = {
-        id: "v2callablefn",
-        entryPoint: "v2callablefn",
-        platform: "gcfv2",
-        httpsTrigger: {},
-        ...SPEC,
-      };
-
-      await triggerRegionHelper.ensureTriggerRegions(backend.of(v1EventFn, v2CallableFn));
-
-      const want: backend.EventTrigger = {
-        eventType: "google.storage.object.create",
-        eventFilters: { resource: "projects/_/buckets/my-bucket" },
-        retry: false,
-      };
-
-      expect(v1EventFn.eventTrigger).to.deep.eq(want);
-      expect(v2CallableFn.httpsTrigger).to.deep.eq({});
     });
 
     it("should set trigger region from API", async () => {
@@ -99,7 +85,7 @@ describe("TriggerRegionHelper", () => {
     });
 
     it("should set trigger region from API then reject on invalid function region", async () => {
-      storageStub.resolves({ location: "US" });
+      storageStub.resolves({ location: "EUROPE-WEST1" });
       const wantFn: backend.Endpoint = {
         id: "wantFn",
         entryPoint: "wantFn",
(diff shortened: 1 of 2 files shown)


wandamora (approved):
Lgtm, thank you!

src/deploy/functions/triggerRegionHelper.ts
```diff
@@ -16,4 +16,40 @@ export async function ensureTriggerRegions(want: backend.Backend): Promise<void>
...
+
+  // Warn if an event function defaults to or is assigned to us-central1 but its trigger is out of the US,
+  // to avoid unnecessary cross-region network hops (e.g., transatlantic).
+  if (process.env.FIREBASE_SUPPRESS_REGION_WARNING === "true") {
+    return;
+  }
+
+  const offendingFunctions: string[] = [];
+  for (const ep of backend.allEndpoints(want)) {
+    if (ep.platform === "gcfv1" || !backend.isEventTriggered(ep)) {
```
ajperel: Why are we ignoring v1 functions? Is there a reason those should stay with international hops?

src/deploy/functions/triggerRegionHelper.spec.ts
```diff
@@ -117,5 +125,135 @@ describe("TriggerRegionHelper", () => {
...
+
+      await triggerRegionHelper.ensureTriggerRegions(backend.of(wantFn));
+
+      expect(logLabeledWarningStub).to.not.have.been.called;
+
+      logLabeledWarningStub.restore();
+      delete process.env.FIREBASE_SUPPRESS_REGION_WARNING;
+    });
+
+    it("should not warn when regions match", async () => {
```
ajperel: This is the most basic non-warn case. I don't think we have to be exhaustive, but probably good to test some of the other cases we don't warn about yet like nam5 or even just a different US source.

src/deploy/functions/triggerRegionHelper.ts
```diff
@@ -16,4 +16,40 @@ export async function ensureTriggerRegions(want: backend.Backend): Promise<void>
...
+    const triggerRegion = ep.eventTrigger?.region;
+    if (ep.region !== "us-central1" || !triggerRegion) {
+      continue;
+    }
+    const isUS =
+      triggerRegion === "us" ||
+      triggerRegion === "nam5" ||
+      triggerRegion === "nam7" ||
+      triggerRegion.startsWith("us-");
+    if (isUS) {
```
ajperel: There's only one line left in the loop after this. I think it'd be slightly cleaner to just do

```
if (!isUS) {
  offendingFunctions.push(...
}
```

rather than `continue` in the positive case to only have one line after.

src/deploy/functions/triggerRegionHelper.ts
```diff
@@ -16,4 +16,40 @@ export async function ensureTriggerRegions(want: backend.Backend): Promise<void>
...
+  const offendingFunctions: string[] = [];
+  for (const ep of backend.allEndpoints(want)) {
+    if (ep.platform === "gcfv1" || !backend.isEventTriggered(ep)) {
+      continue;
+    }
+    const triggerRegion = ep.eventTrigger?.region;
+    if (ep.region !== "us-central1" || !triggerRegion) {
+      continue;
+    }
+    const isUS =
```
ajperel: Nit: Could be good to extract this to a helper method so that it's both easy to update what counts as US if we need (add a new multi-region) and the main logic is easier to read.

src/deploy/functions/triggerRegionHelper.ts
```diff
@@ -16,4 +16,40 @@ export async function ensureTriggerRegions(want: backend.Backend): Promise<void>
...
+  if (process.env.FIREBASE_SUPPRESS_REGION_WARNING === "true") {
+    return;
+  }
+
+  const offendingFunctions: string[] = [];
+  for (const ep of backend.allEndpoints(want)) {
+    if (ep.platform === "gcfv1" || !backend.isEventTriggered(ep)) {
+      continue;
+    }
+    const triggerRegion = ep.eventTrigger?.region;
```
ajperel: I don't think you need the `?` given that you check above that `backend.isEventTriggered` so this should be _guaranteed_ to exist at this point right?

src/deploy/functions/triggerRegionHelper.ts
```diff
@@ -16,4 +16,40 @@ export async function ensureTriggerRegions(want: backend.Backend): Promise<void>
...
+    if (ep.region !== "us-central1" || !triggerRegion) {
```
ajperel: For my own learning... when would region be undefined and why is it safe to continue in that case?

ajperel (commented):
Thank you!

src/deploy/functions/triggerRegionHelper.ts
ajperel: We could maybe simplify this even more with some of the checks above to just like:

```
if (
      ep.region === "us-central1" && 
      triggerRegion && 
      !isUSRegion(triggerRegion)
    ) {
    offendingFunctions.push(`- ${ep.id} (us-central1, Trigger: ${triggerRegion})`);
 }
```

Please ignore any terrible formatting / lint issues in that suggestion.

src/deploy/functions/triggerRegionHelper.ts
shettyvarun268: `triggerRegion` is usually `undefined` for global triggers (like Firebase Auth or Alerts). 
Since these triggers aren't anchored to any single regional data center on the map, cross-region networking hops aren't a concern. It's safe to skip because we don't need to warn users about region mismatches for global concepts.

cabljac: This new check only fires when the function is in us-central1, but the underlying problem (unnecessary cross-region hops for event triggers) is sort of region and direction agnostic right?

We could broaden this for any mismatch (keeping the US-multi-region exemption) - or narrow it for *just* the default case?

src/deploy/functions/triggerRegionHelper.spec.ts
```diff
@@ -117,5 +125,135 @@ describe("TriggerRegionHelper", () => {
...
+
+      const logLabeledWarningStub = sinon.stub(utils, "logLabeledWarning");
+
+      await triggerRegionHelper.ensureTriggerRegions(backend.of(wantFn));
+
+      expect(logLabeledWarningStub).to.have.been.calledOnceWith(
+        "functions",
+        `The following functions have triggers in different regions than they are located:\n` +
+          `- wantFn (us-central1, Trigger: europe-west1)\n` +
+          `To avoid unnecessary cross-region network hops, you should explicitly assign these functions to their trigger regions. ` +
```
cabljac: `you should explicitly assign these functions to their trigger regions` reads as if the user hasn't assigned a region, but this warning also fires when they've explicitly chosen us-central1 right?

Maybe something like "Consider assigning these functions to their trigger regions / consider collocating"

cabljac: or we distinguish default vs. explicit assignment upstream and only recommend reassignment in the default case?

src/deploy/functions/triggerRegionHelper.ts
ajperel: Yes. There's a few things going on here with us-central1 being the default and also being special where things work cross-region in ways I've been told they wouldn't for other cross-region pairings. I think for now we don't need to expand it (but we always can later if we see it'll be useful) but it's a good point that if a user has _explicitly_ chosen to be cross-region that's a good signal.

If we could see that the user was using us-central1 as the default and not explicitly chosen and they have a giant international hop I think it could be better to start with just warning in that case.

src/deploy/functions/triggerRegionHelper.spec.ts
shettyvarun268: Good catch on the copy. I will update the message to 'Consider assigning these functions to their trigger regions or collocating them' to handle cases where the user explicitly chose us-central1. Regarding distinguishing default vs. explicit assignment: it turns out the SDK outputs us-central1 in both cases to the manifest, so the CLI cannot distinguish them at this stage. Updating the copy is the safest fix for now.

src/deploy/functions/triggerRegionHelper.ts
shettyvarun268: Noted. Made the required changes.

src/deploy/functions/triggerRegionHelper.ts
shettyvarun268: Got it. Created a helper method to handle this and make it easier to read.

src/deploy/functions/triggerRegionHelper.ts
shettyvarun268: Simplified the code here

src/deploy/functions/triggerRegionHelper.spec.ts
shettyvarun268: Added test cases as needed

src/deploy/functions/triggerRegionHelper.ts
shettyvarun268: I am still debating this. One one hand I agree that the v1 users might get the biggest benefit out of this since they might have deployed in different regions. But at the same time v1 functions would require some additional lookup which would make the deploy process slower for them(especially with multiple functions).

ajperel: Can you quantify this at all? How complicated and slow is the additional lookup? If we're increasing the deployment time from like 1 minute to 1:02 ... I'm fine with that slowdown. If we were going from 0:30 to 1:30 that would be a problem.

shettyvarun268: For small projects with a few functions, we could probably parallelize the lookups keeping the delay under 1 to 2 seconds (so within your limit!).

The problem really comes down to scale and permissions. If a user has 20+ V1 functions pointing at several distinct databases or buckets, making individual API hits to solve those missing region fields on V1 manifests could tack on 10+ seconds of dead weight to the deploy time. Also, doing high-speed sequential hits runs a risk of tagging rate limits or firing IAM check fails for developers on locked down corporate accounts.

src/deploy/functions/triggerRegionHelper.ts
```diff
@@ -1,19 +1,143 @@
...
+    return;
+  }
+
+  const offendingFunctions: string[] = [];
+  for (const ep of backend.allEndpoints(want)) {
+    if (!backend.isEventTriggered(ep)) {
+      continue;
+    }
+
+    let triggerRegion: string | undefined = undefined;
```
ajperel: I don't know TypeScript very well... do you actually need this or can typescript accurately infer the typing in this case?

ajperel (approved):

PR #10411, opened 2026-04-23
fix: preserve existing cloud run invokers for scheduled v2 functions

### LINKED ISSUE #6549: `firebase deploy` overrides members of invoker role on CloudRun service, even if `preserve_external_changes` is set to `True`

**firebase-tools:** 12.5.3

**Platform:** macOS

### [REQUIRED] Test case

```
from firebase_functions.scheduler_fn import on_schedule, ScheduledEvent

@on_schedule(
    schedule="*/15 * * * *",
    region="europe-west3",
    ingress="ALLOW_INTERNAL_ONLY",
    concurrency=1,
    preserve_external_changes=True,
)
def update_prd_from_meeting_transcripts(event: ScheduledEvent) -> None:
    print(event["job_name"], event["schedule_time"])
    return
```

1. Deploy the example function above (scheduled function) with `firebase deploy functions --project $PROJECT_ID`
2. Create a custom service account (we use terraform, but Cloud Console would work, too)
3. Update the created CloudScheduler job to use the custom service account created in 2 (we use terraform, but Cloud Console would work, too)
4. Add a `role/run.invoker` IAM binding to the CloudRun service created in 1., with the custom service account created in 2 being one of the members.
5. Re-deploy the function or deploy an update with `firebase deploy functions --project $PROJECT_ID`

The updated function is deployed and the `role/run.invoker` on the Cloud Run service is either not updated or only updated additively in order to preserve my changes.
As using the default compute service account is discouraged as a security risk this is an important behavior to rely on for mature organisations that have a requirement **not** to use the default compute service account.

The updated function is deployed and the `role/run.invoker` on the Cloud Run service is set to contain only the default compute service account. 
This is due to the following code in firebase cli:
1. Updating the CloudRun permission replaces the members for the invoker role instead of doing an additive update [here](https://github.com/firebase/firebase-tools/blob/f0c247195db10b4fa01adb8f8c9fb93d1a13f9ed/src/gcp/run.ts#L312).
2. When updating a scheduled function, its invokers are hard-coded to the default compute service account [here](https://github.com/firebase/firebase-tools/blob/f0c247195db10b4fa01adb8f8c9fb93d1a13f9ed/src/deploy/functions/release/fabricator.ts#L542).

Since this log contains a lot of sensitive information, I'd like to provide it if required, which I don't believe is the case as there is a 100% reproduction and there is a clear connection to code in the description above.

IzaakGough:
## **Summary**

Fixes https://github.com/firebase/firebase-tools/issues/6549

Scheduled GCFv2 functions currently **overwrite** the `roles/run.invoker` binding on Cloud Run during deploy. This causes externally added invoker members to be removed on create/update, even if the scheduler-required service account is still present.

This change ensures that scheduled GCFv2 deploys **preserve existing Cloud Run invoker members** by **merging** them with the required scheduler invoker, instead of replacing the binding outright.

---

## **What Changed**

- Added an optional **merge mode** to:
  ```ts
  src/gcp/run.ts -> setInvokerUpdate(...)
  ```

- Updated scheduled GCFv2 **create flow** in `src/deploy/functions/release/fabricator.ts`
  to use:
  ```ts
  setInvokerUpdate(..., { mergeExistingMembers: true })
  ```

- Updated scheduled GCFv2 **update flow** in the same file to pass the same merge option.

- Added helper-level test coverage in `src/gcp/run.spec.ts`

- Added deploy-layer test coverage in `src/deploy/functions/release/fabricator.spec.ts` for scheduled create/update call paths.

---

## **Why**

Before this change, scheduled GCFv2 deploys computed a single desired invoker:

- The function’s configured `serviceAccount`, **or**
- The default compute service account

This value was then written back as the full `roles/run.invoker` binding, which **removed any additional invoker members** added outside of Firebase.

### **Reproduction**

1. Deploy a scheduled GCFv2 function
2. Add an additional custom invoker member on the backing Cloud Run service
3. Re-deploy the function (or make a trivial update)
4. Observe that the custom invoker is removed

Further details of this reproduction can be seen [here](https://github.com/firebase/firebase-tools/issues/6549#issuecomment-4296806721)

---

## **Tests**

Added coverage for:

- ```ts
  run.setInvokerUpdate(..., { mergeExistingMembers: true })
  ```
  preserving existing invoker members

- Scheduled `createV2Function` using merge mode with:
  - default service account
  - explicit service account

- Scheduled `updateV2Function` using merge mode

---

## **Important Behavior Note**

This change is currently **unconditional** for scheduled GCFv2 functions. It does **not** consult or derive from the `preserveExternalChanges / preserve_external_changes` option. 

Instead, scheduled-function invoker updates now **always merge existing `roles/run.invoker` members by default**. This is intentional for this fix, but worth calling out explicitly.

---

## **Potential Follow-up**

A future improvement could thread a real `preserveExternalChanges` signal into this path, making the behavior **conditional** rather than always-on for scheduled GCFv2 functions.

### DIFF
--- a/src/deploy/functions/release/fabricator.spec.ts
+++ b/src/deploy/functions/release/fabricator.spec.ts
@@ -820,32 +820,37 @@ describe("Fabricator", () => {
     });
 
     describe("scheduledTrigger", () => {
-      it("sets the default compute service account by default", async () => {
+      it("sets the default compute service account by default and preserves existing invokers", async () => {
         gcfv2.createFunction.resolves({ name: "op", done: false });
         poller.pollOperation.resolves({ serviceConfig: { service: "service" } });
-        run.setInvokerCreate.resolves();
+        run.setInvokerUpdate.resolves();
         const ep = endpoint(
           { scheduleTrigger: { schedule: "every 5 minutes" } },
           { platform: "gcfv2" },
         );
 
         await fab.createV2Function(ep, new scraper.SourceTokenScraper());
-        expect(run.setInvokerCreate).to.have.been.calledWith(ep.project, "service", [
-          await gce.getDefaultServiceAccount(fab.projectNumber),
-        ]);
+        expect(run.setInvokerUpdate).to.have.been.calledWith(
+          ep.project,
+          "service",
+          [await gce.getDefaultServiceAccount(fab.projectNumber)],
+          { mergeExistingMembers: true },
+        );
       });
 
       it("sets the an explicit service account", async () => {
         gcfv2.createFunction.resolves({ name: "op", done: false });
         poller.pollOperation.resolves({ serviceConfig: { service: "service" } });
-        run.setInvokerCreate.resolves();
+        run.setInvokerUpdate.resolves();
         const ep = endpoint(
           { scheduleTrigger: { schedule: "every 5 minutes" } },
           { platform: "gcfv2", serviceAccount: "sa@" },
         );
 
         await fab.createV2Function(ep, new scraper.SourceTokenScraper());
-        expect(run.setInvokerCreate).to.have.been.calledWith(ep.project, "service", ["sa@"]);
+        expect(run.setInvokerUpdate).to.have.been.calledWith(ep.project, "service", ["sa@"], {
+          mergeExistingMembers: true,
+        });
       });
     });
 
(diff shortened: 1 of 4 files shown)


cabljac:
Ah so the main problem is, as the PR description says - you're making it unconditional, whereas the proper fix would make `preserve_external_changes=True` work?

cabljac:
Closing for now, have a new branch with a full fix on incoming.

PR #10414, opened 2026-04-23
Change default resource locations for functions

wandamora:
This updates the default resource location for various functions from us-central1. This should only impact NEW functions, not the location of existing functions. The exhaustive list includes:

- New event functions match the resource of their event source. (Firestore, Cloud Storage, Firebase Realtime Database, Firebase Data Connect)
- If a Firestore bucket is in a multi-region, then nam5 & nam7  -> us-central1, and  eur3 -> europe-west1
- If a Cloud storage bucket is in a multi-region, then
   - “us” -> us-east1
   - “eu” -> “europe-west1”
   - “asia” -> asia-east1
- Auth blocking functions and Global AI Logic blocking functions (beforeGenerateContent, afterGenerateContent) are set to us-east1.
- Remaining global event functions (Pub/Sub, Firebase Auth/GCIP, Firebase Test Lab, Firebase Remote Config, Firebase Alerts) are set to us-east1

### DIFF
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -12,6 +12,7 @@ import * as runtimes from "./runtimes";
 import * as supported from "./runtimes/supported";
 import * as validate from "./validate";
 import * as ensure from "./ensure";
+import * as events from "../../functions/events/v1";
 import { getDatabase } from "./services/firestore";
 import { getBucket } from "./services/storage";
 import { getDatabaseInstanceDetails } from "./services/database";
@@ -42,7 +43,8 @@ import { needProjectId, needProjectNumber } from "../../projectUtils";
 import { logger } from "../../logger";
 import { ensureTriggerRegions } from "./triggerRegionHelper";
 import { ensureServiceAgentRoles, ensureGenkitMonitoringRoles } from "./checkIam";
-import { FirebaseError } from "../../error";
+import { FirebaseError, getErrStack } from "../../error";
+
 import {
   configForCodebase,
   normalizeAndValidate,
@@ -61,6 +63,7 @@ import { DeployOptions } from "..";
 import * as prompt from "../../prompt";
 
 export const EVENTARC_SOURCE_ENV = "EVENTARC_CLOUD_EVENT_SOURCE";
+export const DEFAULT_FUNCTION_REGION = "us-central1";
 
 /**
  * Prepare functions codebases for deploy.
@@ -405,7 +408,7 @@ export async function resolveDefaultRegions(
 
   const endpoints = Object.values(want.endpoints[build.REGION_TBD] || {});
 
-  const resolvedEndpoints = await Promise.all(
+  await Promise.all(
     endpoints.map(async (endpoint) => {
       let resolvedRegion = "us-central1";
 
@@ -418,27 +421,20 @@ export async function resolveDefaultRegions(
       } catch (err: any) {
         logger.debug(
           `Failed to resolve region for endpoint ${endpoint.id}. Defaulting to us-central1.`,
-          err,
+          getErrStack(err),
         );
       }
 
-      return { endpoint, resolvedRegion };
+      moveEndpointToRegion(want, endpoint, resolvedRegion);
     }),
   );
-
-  for (const { endpoint, resolvedRegion } of resolvedEndpoints) {
-    moveEndpointToRegion(want, endpoint, resolvedRegion);
-  }
 }
 
 function resolveRegionForBlockingTrigger(
   endpoint: backend.Endpoint & backend.BlockingTriggered,
 ): string {
   const eventType = endpoint.blockingTrigger.eventType;
-  if (
-    eventType === "providers/cloud.auth/eventTypes/user.beforeCreate" ||
-    eventType === "providers/cloud.auth/eventTypes/user.beforeSignIn"
-  ) {
+  if (eventType === events.BEFORE_CREATE_EVENT || eventType === events.BEFORE_SIGN_IN_EVENT) {
     return "us-east1";
   }
 
@@ -446,7 +442,7 @@ function resolveRegionForBlockingTrigger(
     return "us-east1";
   }
 
-  return "us-central1";
+  return DEFAULT_FUNCTION_REGION;
 }
 
 async function resolveRegionForEventTrigger(
@@ -455,6 +451,7 @@ async function resolveRegionForEventTrigger(
   const eventTrigger = endpoint.eventTrigger;
   const eventType = eventTrigger.eventType;
 
+  // Global functions should be deployed to us-east1.
   if (
     eventType.startsWith("google.cloud.pubsub.") ||
     eventType.startsWith("providers/cloud.auth/eventTypes/") ||
@@ -466,6 +463,11 @@ async function resolveRegionForEventTrigger(
     return "us-east1";
   }
 
+  // Firestore functions should be deployed to the same region as the database.
+  // In multi-region locations, we default to:
+  // * nam5 -> us-central1
+  // * nam7 -> us-central1
+  // * eur3 -> europe-west1
   if (eventType.startsWith("google.cloud.firestore.")) {
     try {
       const databaseId = eventTrigger.eventFilters?.database || "(default)";
@@ -476,10 +478,15 @@ async function resolveRegionForEventTrigger(
       if (locationId === "eur3") return "europe-west1";
       return locationId;
     } catch (err: any) {
-      logger.debug("Failed to resolve Firestore database location", err);
+      logger.debug("Failed to resolve Firestore database location", getErrStack(err));
     }
   }
 
+  // Cloud Storage functions should be deployed to the same region as the bucket.
+  // In multi-region locations, we default to:
+  // * us -> us-east1
+  // * eu -> europe-west1
+  // * asia -> asia-east1
   if (eventType.startsWith("google.cloud.storage.")) {
     try {
       const bucketName = eventTrigger.eventFilters?.bucket;
@@ -493,10 +500,11 @@ async function resolveRegionForEventTrigger(
         return locationId;
       }
     } catch (err: any) {
-      logger.debug("Failed to resolve Cloud Storage bucket location", err);
+      logger.debug("Failed to resolve Cloud Storage bucket location", getErrStack(err));
     }
   }
 
+  // Realtime Database functions should be deployed to the same region as the database.
   if (eventType.startsWith("google.firebase.database.")) {
     if (eventTrigger.region) return eventTrigger.region;
 
@@ -509,15 +517,16 @@ async function resolveRegionForEventTrigger(
         }
       }
     } catch (err: any) {
-      logger.debug("Failed to resolve Realtime Database instance location", err);
+      logger.debug("Failed to resolve Realtime Database instance location", getErrStack(err));
     }
   }
 
+  // DataConnect functions should be deployed to the same region as the database.
   if (eventType.startsWith("google.firebase.dataconnect.")) {
     if (eventTrigger.region) return eventTrigger.region;
   }
 
-  return "us-central1";
+  return DEFAULT_FUNCTION_REGION;
 }
 
 /**
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -409,7 +409,7 @@ export async function resolveDefaultRegions(
 
     if (backend.isBlockingTriggered(endpoint)) {
       // Set region for blocking functions to us-east1. This includes:
-      // - Auth blocking functions
+      // - Auth blocking functions (beforeCreate, beforeSignIn)
       // - Global AI Logic functions
       const eventType = endpoint.blockingTrigger.eventType;
       if (
@@ -677,11 +677,11 @@ export async function loadCodebases(
     if (firebaseJsonRuntime && !supported.isRuntime(firebaseJsonRuntime as string)) {
       throw new FirebaseError(
         `Functions codebase ${codebase} has invalid runtime ` +
-          `${firebaseJsonRuntime} specified in firebase.json. Valid values are: \n` +
-          (Object.keys(supported.RUNTIMES) as supported.Runtime[])
-            .filter((runtime) => !supported.isDecommissioned(runtime))
-            .map((s) => `- ${s}`)
-            .join("\n"),
+        `${firebaseJsonRuntime} specified in firebase.json. Valid values are: \n` +
+        (Object.keys(supported.RUNTIMES) as supported.Runtime[])
+          .filter((runtime) => !supported.isDecommissioned(runtime))
+          .map((s) => `- ${s}`)
+          .join("\n"),
       );
     }
     const runtimeDelegate = await runtimes.getRuntimeDelegate(delegateContext);
@@ -734,8 +734,8 @@ function warnIfDartBackendHasUnsupportedTriggers(want: backend.Backend): void {
     logLabeledWarning(
       "functions",
       `The following Dart functions use triggers that are not yet supported for production deployment: ${unsupported.map((ep) => ep.id).join(", ")}. ` +
-        "They will be deployed but may not work as expected. " +
-        "See https://github.com/firebase/firebase-functions-dart for current trigger support.",
+      "They will be deployed but may not work as expected. " +
+      "See https://github.com/firebase/firebase-functions-dart for current trigger support.",
     );
   }
 }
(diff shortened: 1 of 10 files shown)


src/deploy/functions/prepare.ts
```diff
@@ -391,20 +396,103 @@ export async function resolveDefaultRegions(
...
+        resolvedRegion = "us-east1";
+      } else if (eventType.startsWith("google.cloud.firestore.")) {
+        // Set region for Firestore events based on database location, or to the
+        // nearest single-region location if the database is multi-regional.
+        try {
+          const databaseId = eventTrigger.eventFilters?.database || "(default)";
+          const db = await getDatabase(endpoint.project, databaseId);
+          const locationId = db.locationId.toLowerCase();
+          if (locationId === "nam5" || locationId === "nam7") {
+            resolvedRegion = "us-central1";
```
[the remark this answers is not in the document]
wandamora: Consensus in design doc was to leave these resources in us-central1.

src/deploy/functions/prepare.ts
```diff
@@ -390,22 +394,121 @@ export async function resolveDefaultRegions(
...
+}
+
+async function resolveRegionForEventTrigger(
+  endpoint: backend.Endpoint & backend.EventTriggered,
+): Promise<string> {
+  const eventTrigger = endpoint.eventTrigger;
+  const eventType = eventTrigger.eventType;
+
+  if (
+    eventType.startsWith("google.cloud.pubsub.") ||
```
ajperel, line 455: I think we talked yesterday about maybe this code could be simplified to rather than checking event type just check if the target region is global and if so target us-east1. That could be robust to new global things not accidentally defaulting to us-central1

Do you think that would work? Better or not?

src/deploy/functions/prepare.ts
```diff
@@ -390,22 +394,121 @@ export async function resolveDefaultRegions(
...
+    return "us-east1";
+  }
+
+  if (eventType.startsWith("google.cloud.firestore.")) {
+    try {
+      const databaseId = eventTrigger.eventFilters?.database || "(default)";
+      const db = await getDatabase(endpoint.project, databaseId);
+      const locationId = db.locationId.toLowerCase();
+
+      if (locationId === "nam5" || locationId === "nam7") return "us-central1";
```
ajperel, line 476: We have location resolution logic throughout all this code. It _does_ match the design. But, taking a step back... I think we are _always_ suggesting the a single mapping from (multi-region) to region. It just so happens that a bunch of regions don't matter for each event type.

Do you think it'd be clearer to have something like 
```
getFunctionRegionFromEventSourceRegion() {
    // global case
    // multi-region cases (nam5, etc.)
    // multi-regional cases (us, eu, asia)
    // single region default (return the region)
}
```

Then it's basically "If you're in scope to go near the default we call this one function to do the mapping" and if you're not in scope to move you fall through to the old default of us-central1.

This would only break down if we were resolved differently for different event types, e.g., Firestore triggers in nam5 resolve to a different region than Cloud Storage. I have no idea if that would ever happen?

I think the flip side argument to this is that some of these regions actually only really make sense for specific services and we're sort of conflating all that.

I'm not 100% sold on my suggestion so curious your thoughts.

src/deploy/functions/prepare.ts
```diff
@@ -390,22 +394,121 @@ export async function resolveDefaultRegions(
...
+): string {
+  const eventType = endpoint.blockingTrigger.eventType;
+  if (
+    eventType === "providers/cloud.auth/eventTypes/user.beforeCreate" ||
+    eventType === "providers/cloud.auth/eventTypes/user.beforeSignIn"
+  ) {
+    return "us-east1";
+  }
+
+  if (isGlobalAILogicEndpoint(endpoint)) {
```
ajperel, line 440: Should we be doing anything for non-global AI logic endpoints?

src/deploy/functions/prepare.ts
wandamora: I decided not to go ahead with that approach because not all global functions are guaranteed to have a `eventTrigger.region === "global"` property, specifically Pub/Sub and non-blocking Auth functions.

src/deploy/functions/prepare.ts
wandamora: I think your suggestion could make sense if source mapping was uniform across all functions, but given that some of these require lookups of different sources and also don't have the same fallbacks suggested in the design, I'm more inclined to keep this split by function type where it's different enough.

src/deploy/functions/prepare.ts
wandamora: Thomas' doc doesn't specifically call it out, so naively I'm letting those fallback to us-central1.

src/deploy/functions/prepare.ts
ajperel: Lovely, OK. 

I don't think we have the right info here to do assertExhaustive. I can live with this as is, but if we can think of a way to make sure any _new_ global event types properly get the right default that would be better. But off the top of my head I haven't thought of one yet.

src/deploy/functions/prepare.ts
ajperel: I was assuming those were 'regional' ones and we should pair them with region but I don't know. I think we could talk to the AI Logic folks and ask.

wandamora: So the scenario that will happen is that the regional functions are deployed to where the user defines it (e.g. `exports.myAiFunction = onBeforeGenerateContent({ region: "europe-west1" }, (event) => { ... });`, otherwise the fallback is `us-central1` if no preference. It seems in this way, we are always co-locating the function with the trigger when the user specifies it.

ajperel: Interesting. I guess this is different from say, Firestore, where we're matching an event based on where the Firestore service is running for that user. I guess for AI Logic it's regional _only_ in the case where you care to chose a region.

ajperel (approved):
LGTM. Thanks for talking through a bunch of it with me. Since this is a fairly consequential change I think it'd be good to get a secondary review from a more knowledgeable expert.

src/deploy/functions/services/database.ts
```diff
@@ -1,5 +1,56 @@
...
+const instanceCache = new Map<string, DatabaseInstance>();
+const instancePromiseCache = new Map<string, Promise<DatabaseInstance>>();
+
+/**
+ * Clear the database instance cache. Used for testing.
+ * @internal
+ */
+export function clearCache(): void {
+  instanceCache.clear();
+  instancePromiseCache.clear();
```
taeold: hmm having both instance cache and promise cache feels a bit weird to me? what's the intention with having both caches?

src/deploy/functions/services/database.ts
```diff
@@ -1,5 +1,56 @@
...
+      instanceCache.set(key, details);
+      instancePromiseCache.delete(key);
+      return details;
+    })
+    .catch((error) => {
+      instancePromiseCache.delete(key);
+      throw error;
+    });
+
+  instancePromiseCache.set(key, instancePromise);
```
taeold: I'd usually write this using async/await calls, e.g.

```typescript
  const details = await getDetails(projectId, instanceName);
  instanceCache.set(key, details);
  return details
```

but maybe there is some reason why we want to keep the promise in cache for reasons I'm not quite groking.

src/deploy/functions/services/storage.ts
```diff
@@ -7,6 +7,48 @@ import { regionInLocation } from "../../../gcp/location";
...
+    .getBucket(bucketName)
+    .then((b) => {
+      bucketCache.set(bucketName, b);
+      bucketPromiseCache.delete(bucketName);
+      return b;
+    })
+    .catch((error) => {
+      bucketPromiseCache.delete(bucketName);
+      throw error;
+    });
```
taeold: same comment to what I made in database.ts. Not sure why promise cache is necessary.

src/deploy/functions/prepare.ts
```diff
@@ -390,22 +394,121 @@ export async function resolveDefaultRegions(
...
+    moveEndpointToRegion(want, endpoint, resolvedRegion);
+  }
+}
+
+function resolveRegionForBlockingTrigger(
+  endpoint: backend.Endpoint & backend.BlockingTriggered,
+): string {
+  const eventType = endpoint.blockingTrigger.eventType;
+  if (
+    eventType === "providers/cloud.auth/eventTypes/user.beforeCreate" ||
```
taeold: nit* probably a good idea to use constant like here:

https://github.com/firebase/firebase-tools/blob/f3f8d4ed415d6e0c5df0c58edce58c5da07e4424/src/deploy/functions/services/auth.ts#L89

src/deploy/functions/prepare.ts
```diff
@@ -390,22 +394,121 @@ export async function resolveDefaultRegions(
...
+    eventType === "providers/cloud.auth/eventTypes/user.beforeSignIn"
+  ) {
+    return "us-east1";
+  }
+
```
taeold, line 439: i think this might be missing couple of other auth blocking function types?

src/deploy/functions/prepare.ts
```diff
@@ -390,22 +394,121 @@ export async function resolveDefaultRegions(
...
+  if (isGlobalAILogicEndpoint(endpoint)) {
+    return "us-east1";
+  }
+
+  return "us-central1";
```
taeold: nit* maybe worth pulling us-central1 as a global constant somewhere.

src/deploy/functions/prepare.ts
```diff
@@ -390,22 +394,121 @@ export async function resolveDefaultRegions(
...
+}
+
+async function resolveRegionForEventTrigger(
+  endpoint: backend.Endpoint & backend.EventTriggered,
+): Promise<string> {
+  const eventTrigger = endpoint.eventTrigger;
+  const eventType = eventTrigger.eventType;
+
+  if (
```
taeold, line 454: I'd leave some comment here to explain the rationale as it was left in the prev code left by Thomas

src/deploy/functions/prepare.ts
taeold, line 476: are we not suggesting us-east1 for nam5? is that going to be a follow up?

src/deploy/functions/prepare.ts
```diff
@@ -390,22 +394,121 @@ export async function resolveDefaultRegions(
...
+          return details.location.toLowerCase();
+        }
+      }
+    } catch (err: any) {
+      logger.debug("Failed to resolve Realtime Database instance location", err);
+    }
+  }
+
+  if (eventType.startsWith("google.firebase.dataconnect.")) {
+    if (eventTrigger.region) return eventTrigger.region;
```
taeold, line 525: hmm i might double check why we aren't looking up dataconnect region like how we are checking Firestore or RTDB. IIUC, eventTrigger.region == function.region, so I wouldn't be surprised if this isn't doing anything.

src/deploy/functions/prepare.ts
```diff
@@ -390,22 +394,121 @@ export async function resolveDefaultRegions(
...
+    endpoints.map(async (endpoint) => {
+      let resolvedRegion = "us-central1";
+
+      try {
+        if (backend.isBlockingTriggered(endpoint)) {
+          resolvedRegion = resolveRegionForBlockingTrigger(endpoint);
+        } else if (backend.isEventTriggered(endpoint)) {
+          resolvedRegion = await resolveRegionForEventTrigger(endpoint);
+        }
+      } catch (err: any) {
```
taeold: probably want to log the err too

src/deploy/functions/prepare.ts
```diff
@@ -390,22 +394,121 @@ export async function resolveDefaultRegions(
...
+          err,
+        );
+      }
+
+      return { endpoint, resolvedRegion };
+    }),
+  );
+
+  for (const { endpoint, resolvedRegion } of resolvedEndpoints) {
+    moveEndpointToRegion(want, endpoint, resolvedRegion);
```
taeold, line 428: nit*. you could probably inline this call in the above Promise.all() call to avoid having basically 2 for loops.

src/deploy/functions/prepare.ts
```diff
@@ -390,22 +394,121 @@ export async function resolveDefaultRegions(
...
-    // 3. Functions that have a global resource (e.g. Auth, Test Lab, Pub/Sub, etc.) can be placed in
-    //    a region with higher headroom.
-    // 4. HTTP functions may be defaultable to a different region because it is up to the customer to
-    //    wire up the URL we give them irrespective.
-    // 5. Callable functions could be moved by default, though this will require breaking changes to
-    //    the client SDKs as well.
-    moveEndpointToRegion(want, endpoint, "us-central1");
+  const endpoints = Object.values(want.endpoints[build.REGION_TBD] || {});
+
+  const resolvedEndpoints = await Promise.all(
```
taeold: Just a rule of thumb, don't mind if you ignore this:

I think Promise.all() is kinda wacky and makes error handling a lot harder.

I get why we would want to use Promise.all here - resolving region for an endpoint might end up making an API call to get information about db/gcs bucket, and parallelizing it should give us a speed up here.

Another way of writing this code though could be to

1) prune through all the endpoints, figure out whether we have Firestore/RTDB/GCS triggers

2) Get information about event source

3) Then loop through each endpoint and resolve based on the fetched resource region

I usually opt to write it like this just to avoid Promise.all(). The code we have here probably works, but when it blows up  due to transient networking error, server being down, or some auth issue, it will bubble up in the CLI in a way that looks pretty cryptic to the end user.

src/deploy/functions/prepare.ts
wandamora: I skipped the rest because Thomas' doc specifies only beforeSignIn and beforeCreate in the scope (https://docs.google.com/document/d/13hPRhW1guCCr_Nl7uXToyBwhnAunsi3MiH2oN55QJPs/edit?pli=1&resourcekey=0-LdpDfGSrBJ_XSyhg8wu7Wg&tab=t.z4eba3y8y17l#heading=h.yr6811iiucxn).

Re-reading it now though, and it says "Most blocking functions attach to a global resource. These functions can also be placed in us-east1. This applies to all Auth blocking functions..."

It's hard to know whether he intentionally left out the remaining blocking functions (or why) in the scope or if we should capturing all of them. Do you have any insight on whether it might be okay to also move `beforeSendEmail` and `beforeSendSms`?

src/deploy/functions/services/database.ts
wandamora: It's a way to dedupe concurrent async requests. By caching the promise, we can also capture pending promises and prevent duplicating network calls for the same resource. The goal is to keep deployments speedy. TBH this was a reaction to the gemini code assist bot suggesting we should cache database details to avoid slow deployments, but if you think this is overkill then I'm happy to remove it too.

src/deploy/functions/services/storage.ts
wandamora: Answered in the database.ts comment.

src/deploy/functions/prepare.ts
wandamora: We were asked by the firestore team to keep the default in us-central1 for now.

src/deploy/functions/prepare.ts
wandamora: Still looking into this but I addressed most of the comments.

wandamora: I have little knowledge of Dataconnect, but it seems like its event source is the dataconnect service that is discoverable via the service or connector field.

src/deploy/functions/prepare.ts
taeold: My read is that he didn't leave them out intentionally. I think all the existing auth blocking functions are "global" in the same sense as the other 2 that is specifically handled here.

src/deploy/functions/prepare.ts
taeold: sg

src/deploy/functions/services/database.ts
taeold: ah i see. I think this is fine. If you do take up the suggestion I made in https://github.com/firebase/firebase-tools/pull/10414#discussion_r3158092429, you probably won't need this.

src/deploy/functions/prepare.ts
wandamora: Thanks for the suggestion, I removed the Promise.all() approach.

src/deploy/functions/prepare.ts
wandamora: Added all auth blocking functions.

src/deploy/functions/services/database.ts
wandamora: Removed the promise cache now that I'm not calling these with Promise.all()

PR #10415, opened 2026-04-23
fix: body-parser shim to allow strict parse failures to reach the app

### LINKED ISSUE #10404: Next.js Hosting frameworks SSR returns 500 for malformed application/json while local next start returns 400
## Summary
When deploying a Next.js app through Firebase Hosting framework support, malformed `application/json` requests can return `500 Internal Server Error` from deployed SSR/API routes even when:
- the route handlers themselves are hardened to return `400 Invalid JSON body`
- the same app returns `400` for the same malformed requests under plain local `next start`

In this case, the failure reproduces on both the custom domain and the direct generated SSR Cloud Run service URL, which suggests the issue is in the Firebase frameworks / deployed SSR adapter path rather than app route code or plain local Next runtime behavior.

## Environment
- firebase-tools local baseline: `15.5.1`
- firebase-tools stable lane also tested: `15.15.0`
- generated deploy bundle dependency: `firebase-frameworks: 0.11.8`
- app next version: `14.2.35`
- hosting config uses framework-aware Hosting with `frameworksBackend.region = us-central1`

## Generated SSR wrapper
From `.firebase/dinepoll/functions/server.js`:

```js
const { onRequest } = require('firebase-functions/v2/https');
const server = import('firebase-frameworks');
exports.ssrdinepoll = onRequest({"region":"us-central1"}, (req, res) => server.then(it => it.handle(req, res)));
```

## Repro shape
Affected endpoints in the app:
- `/api/notifications/register`
- `/api/restaurant-leads`
- `/api/notifications/restaurant-reply`

These routes were explicitly hardened to read raw body text and return `400 Invalid JSON body` on JSON parse failure.

### Probe request
- Method: `POST`
- Header: `Content-Type: application/json`
- Body: `{`

### Local plain Next runtime (`next start`)
Host:
- `http://127.0.0.1:4010`

Observed on all three endpoints:
- status `400`
- body contains `Invalid JSON body`

### Deployed Firebase frameworks SSR path
Hosts:
- `https://dinepoll.com`
- `https://ssrdinepoll-dm5q7suora-uc.a.run.app`

Observed on all three endpoints:
- status `500`
- body `Internal Server Error`

## Expected behavior
Malformed `application/json` should return a client error such as `400 Bad Request`, or at minimum allow the Next route handler to surface its guarded invalid-JSON response, instead of turning the request into an upstream `500` before route-level validation runs.

## Why this seems adapter-specific
This behavior was narrowed with three controls:
1. malformed `text/plain` against deployed routes now returns honest `400`s after route hardening
2. malformed `application/json` returns `500`s on both the custom domain and the direct generated SSR service URL
3. the exact same malformed `application/json` requests return `400`s under plain local `next start`

That combination suggests the deployed Firebase frameworks / SSR adapter path is handling malformed `application/json` differently from both the route code and plain local Next runtime.

## Additional deploy-path signal
To test whether this was only an older CLI behavior, the deploy was also retried with a pinned stable Firebase CLI lane (`15.15.0`).

Observed deploy sequence:
- Next build completes
- frameworks function packaging completes
- `.firebase/dinepoll/functions` source upload completes
- hosting file upload completes
- deploy then stalls specifically at:
  - `functions: updating Node.js 22 (2nd Gen) function firebase-frameworks-dinepoll:ssrdinepoll(us-central1)...`

The same stall was also observed from the local `15.5.1` CLI. That means the deploy/update path itself is now suspect around the generated 2nd Gen function update, not just the malformed-JSON runtime behavior after deployment.

## Related issues checked
Potentially related wrapper-path issues, but not this exact malformed-JSON symptom:
- #8249, generated wrapper with `server.then(it => it.handle(req, res))`
- #6620, deployed SSR `500` behavior differing from local/emulator behavior

## Requested outcome
- confirm whether malformed `application/json` is being parsed upstream in the deployed Firebase frameworks SSR path before Next route handlers run
- confirm whether this is expected behavior or a bug/regression in `firebase-frameworks` / `firebase-tools`
- advise whether there is a current workaround for apps that need malformed JSON to fail as `400` instead of `500`
- advise whether the separate stall at `functions: updating ... ssrdinepoll(us-central1)` is a known frameworks deploy issue for 2nd Gen Next.js Hosting backends

leoortizz:
### Description

Fixes #10404. Malformed application/json returns 500 on Hosting framework SSR while the same route returned 400(reached the API handler) locally; the shim only clears `entity.parse.failed` so the framework can handle the request.

- Add patchBodyParser, installBodyParserTolerance, and getBodyParserToleranceShim in utils; document PARSE_FAILED_TYPE against body-parser read.js
- Prepend the shim to generated server.js for CJS and ESM; ESM path adds createRequire(import.meta.url) before the snapshot so require and require.cache match the shim
- Use non-enumerable __bodyParserPassthroughApplied on body-parser exports for idempotency
- Add express/supertest tests for #10404-style JSON failures, other parsers, limits, urlencoded depth, and shim string checks

### Scenarios Tested

**Deployments / runtime (primary)**

- **Hosting + frameworks SSR backend** — Deploy an app that uses framework-aware Hosting with a generated **2nd gen** `server.js` (same shape as today: shim first, then `firebase-functions` / dynamic `firebase-frameworks` import). Confirm **`server.js`** on disk (or in the uploaded bundle) starts with the **body-parser shim**, then the function wiring.
- **Malformed JSON on deployed URLs (#10404)** — From the **custom domain** and from the **direct Cloud Run / SSR service URL**, `POST` the same routes with `Content-Type: application/json` and an invalid body (e.g. truncated `{`). Expect the app to behave like **`next start`**: route-level handling can run (e.g. **400** + app body), not an opaque **500** from failing inside Express `body-parser` before the framework.
- **`type: "module"` vs CommonJS** — One project with **`"type": "module"`** and one without; after deploy, confirm both bundles load (no early crash on import), and the **ESM** path has **`createRequire(import.meta.url)`** before any shim code that uses `require` / `require.cache`.
- **Regression: happy paths** — Valid JSON POSTs, normal SSR pages, and non-JSON content-types still behave as before (no change to success paths).

**Unit tests (supporting)**

- `utils.spec.ts` body-parser tolerance + shim string checks

### Sample Commands

deploy using a build from this PR then test a request:
```bash
curl -i -X POST https://<deployment-url>/api/test -H 'Content-Type: application/json' --data '{'
```

### DIFF
--- a/src/frameworks/index.ts
+++ b/src/frameworks/index.ts
@@ -25,6 +25,7 @@ import {
   findDependency,
   conjoinOptions,
   frameworksCallToAction,
+  getBodyParserToleranceShim,
   getFrameworksBuildTarget,
 } from "./utils";
 import {
@@ -539,25 +540,35 @@ ${
 
       // TODO move to templates
 
+      // Prepends getBodyParserToleranceShim() before framework imports; see
+      // https://github.com/firebase/firebase-tools/issues/10404
+      // ESM needs createRequire first: the shim uses require() and require.cache. Same shim string
+      // for both branches; only import/export wrapping differs.
+      const bodyParserToleranceShim = getBodyParserToleranceShim();
+
       if (packageJson.type === "module") {
         await writeFile(
           join(functionsDist, "server.js"),
-          `import { onRequest } from 'firebase-functions/v2/https';
-  const server = import('firebase-frameworks');
-  export const ${functionId} = onRequest(${JSON.stringify(
-    frameworksBackend || {},
-  )}, (req, res) => server.then(it => it.handle(req, res)));
-  `,
+          `import { createRequire } from 'module';
+const require = createRequire(import.meta.url);
+${bodyParserToleranceShim}
+const { onRequest } = require('firebase-functions/v2/https');
+const server = import('firebase-frameworks');
+export const ${functionId} = onRequest(${JSON.stringify(
+            frameworksBackend || {},
+          )}, (req, res) => server.then(it => it.handle(req, res)));
+`,
         );
       } else {
         await writeFile(
           join(functionsDist, "server.js"),
-          `const { onRequest } = require('firebase-functions/v2/https');
-  const server = import('firebase-frameworks');
-  exports.${functionId} = onRequest(${JSON.stringify(
-    frameworksBackend || {},
-  )}, (req, res) => server.then(it => it.handle(req, res)));
-  `,
+          `${bodyParserToleranceShim}
+const { onRequest } = require('firebase-functions/v2/https');
+const server = import('firebase-frameworks');
+exports.${functionId} = onRequest(${JSON.stringify(
+            frameworksBackend || {},
+          )}, (req, res) => server.then(it => it.handle(req, res)));
+`,
         );
       }
     } else {
--- a/src/frameworks/utils.spec.ts
+++ b/src/frameworks/utils.spec.ts
@@ -1,9 +1,20 @@
 import { expect } from "chai";
 import * as sinon from "sinon";
 import * as fs from "fs";
+import * as express from "express";
+import * as supertest from "supertest";
 import { resolve, join } from "path";
 
-import { warnIfCustomBuildScript, isUrl, getNodeModuleBin, conjoinOptions } from "./utils";
+import {
+  BodyParserExports,
+  conjoinOptions,
+  getBodyParserToleranceShim,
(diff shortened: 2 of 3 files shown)


Yuangwang:
I see this is still causing some issues for the user for now https://github.com/firebase/firebase-tools/issues/10404 please let me know when the fix is confirmed and I will take a closer look then

PR #10417, opened 2026-04-24
Make dartfunctions experiment public in the CLI

### LINKED ISSUE #10396: Make dartfunctions experiment public in CLI
Right now, the dartfunctions experiment has its config set to `public: false`, making it not visible to users in the CLI when running `firebase experiments:list`. This should be flipped now that Dart functions are public to use.

wandamora:
This will make the dartfunctions experiment visible to users in the CLI when they list experiments using `firebase experiments:list`.


shettyvarun268 (approved):
LGTM

PR #10419, opened 2026-04-24
[FDC] Add always required warning level to block deployment

mtr002:


src/dataconnect/build.ts
```diff
@@ -21,7 +21,7 @@ export async function build(
   const buildResult = await DataConnectEmulator.build(args);
   if (buildResult?.errors?.length) {
     buildResult.errors.forEach((e) => {
-      if (e.extensions?.warningLevel) {
+      if (e.extensions?.warningLevel && e.extensions.warningLevel !== "ALWAYS_REQUIRED") {
```
fredzqm: This is just GA stats.

We can track ALWAYS_REQUIRED in a different Google Analytics metric label than `numBuildErrors`

Let's revert the change here.

fredzqm (approved):

src/dataconnect/build.ts
mtr002: Done.

PR #10426, opened 2026-04-28
fix: resolve default regions before matching endpoints and check all function generations under dartfunctions

### LINKED ISSUE #10409: Deploying event functions fails on `v15.15.0`

**firebase-tools:** 15.15.0

**Platform:** macOS

### [REQUIRED] Test case

N/A

1. Deploy an `onValueUpdated` realtime database event cloud function using `firebase-tools: 15.14.0`
2. Update `firebase-tools:` to `15.15.0`
3. Update something in the event function and deploy again
4. See crash

Function is deployed successfully as usual.

Changing from a HTTPS function to an event function is not allowed. Please delete your function and create a new one instead.

wandamora:
fix: resolve default regions before matching endpoints and preserve trigger metadata under dartfunctions

### Description

This changes when we resolve default regions to before grouping endpoints. Before, new endpoints with REGION_TBD set were mismatched against early, existing endpoints that had a region resolved. This also changes how v2 functions are discovered when the `dartfunctions` experiment is enabled. Before, we were skipping invoking the v2 discovery when the experiment was set and only using the Cloud Run API, but we should be handling all generations of functions with the experiment enabled.

### Scenarios Tested

`firebase experiments:enable dartfunctions`
`firebase deploy --project $PROJECT_ID`

```
=== Deploying to '$PROJECT_ID'...

i  deploying functions
i  functions: preparing codebase default for deployment
i  artifactregistry: ensuring required API artifactregistry.googleapis.com is enabled...
i  functions: Loading and analyzing source code for codebase default to determine what to deploy
Serving at port 8943

i  extensions: ensuring required API firebaseextensions.googleapis.com is enabled...
i  functions: packaged /usr/local/google/home/morawand/issue-10409-repro/issues-10409/functions (70.02 KB) for uploading
i  functions: ensuring required API run.googleapis.com is enabled...
i  functions: ensuring required API eventarc.googleapis.com is enabled...
i  functions: ensuring required API pubsub.googleapis.com is enabled...
i  functions: ensuring required API storage.googleapis.com is enabled...
i  functions: generating the service identity for pubsub.googleapis.com...
i  functions: generating the service identity for eventarc.googleapis.com...
✔  functions: functions source uploaded successfully

The following functions are found in your project but do not exist in your local source code:
        helloWorld(us-central1)

If you are renaming a function or changing its region, it is recommended that you create the new function first before deleting the old one to prevent event loss. For more info, visit https://firebase.google.com/docs/functions/manage-functions#modify

✔ Would you like to proceed with deletion? Selecting no will continue the rest of the deployments. No
i  functions: creating Node.js 24 (2nd Gen) function onValueUpdated(us-central1)...
✔  functions[onValueUpdated(us-central1)] Successful create operation.

✔  Deploy complete!

Project Console: https://console.firebase.google.com/project/$PROJECT_ID/overview
```

updated the function, then reran `firebase deploy --project $PROJECT_ID`

```
=== Deploying to '$PROJECT_ID'...

i  deploying functions
i  functions: preparing codebase default for deployment
i  artifactregistry: ensuring required API artifactregistry.googleapis.com is enabled...
i  functions: Loading and analyzing source code for codebase default to determine what to deploy
Serving at port 8873

i  extensions: ensuring required API firebaseextensions.googleapis.com is enabled...
i  functions: packaged /usr/local/google/home/morawand/issue-10409-repro/issues-10409/functions (70.03 KB) for uploading
i  functions: ensuring required API run.googleapis.com is enabled...
i  functions: ensuring required API eventarc.googleapis.com is enabled...
i  functions: ensuring required API pubsub.googleapis.com is enabled...
i  functions: ensuring required API storage.googleapis.com is enabled...
i  functions: generating the service identity for pubsub.googleapis.com...
i  functions: generating the service identity for eventarc.googleapis.com...
✔  functions: functions source uploaded successfully

The following functions are found in your project but do not exist in your local source code:
        helloWorld(us-central1)

If you are renaming a function or changing its region, it is recommended that you create the new function first before deleting the old one to prevent event loss. For more info, visit https://firebase.google.com/docs/functions/manage-functions#modify

✔ Would you like to proceed with deletion? Selecting no will continue the rest of the deployments. No
i  functions: updating Node.js 24 (2nd Gen) function onValueUpdated(us-central1)...
✔  functions[onValueUpdated(us-central1)] Successful update operation.

✔  Deploy complete!

Project Console: https://console.firebase.google.com/project/$PROJECT_ID/overview
```

### DIFF
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -618,7 +618,7 @@ async function loadCloudRunServices(
       }
     }
   } catch (err: any) {
-    logger.debug(err.message);
+    logger.debug(`Error loading Cloud Run services: ${err.message}`);
     unreachableRegions.run = ["unknown"];
   }
 }
--- a/src/deploy/functions/backend.ts
+++ b/src/deploy/functions/backend.ts
@@ -573,19 +573,8 @@ async function loadExistingBackend(ctx: Context): Promise<Backend> {
   }
   unreachableRegions.gcfV1 = gcfV1Results.unreachable;
 
-  if (experiments.isEnabled("functionsrunapionly") || experiments.isEnabled("dartfunctions")) {
-    try {
-      const runServices = await run.listServices(ctx.projectId);
-      for (const service of runServices) {
-        const endpoint = run.endpointFromService(service);
-        existingBackend.endpoints[endpoint.region] =
-          existingBackend.endpoints[endpoint.region] || {};
-        existingBackend.endpoints[endpoint.region][endpoint.id] = endpoint;
-      }
-    } catch (err: any) {
-      logger.debug(err.message);
-      unreachableRegions.run = ["unknown"];
-    }
+  if (experiments.isEnabled("functionsrunapionly")) {
+    await loadCloudRunServices(ctx, existingBackend, unreachableRegions, false);
   } else {
     const gcfV2Results = await gcfV2.listAllFunctions(ctx.projectId);
     for (const apiFunction of gcfV2Results.functions) {
@@ -594,11 +583,44 @@ async function loadExistingBackend(ctx: Context): Promise<Backend> {
       existingBackend.endpoints[endpoint.region][endpoint.id] = endpoint;
     }
     unreachableRegions.gcfV2 = gcfV2Results.unreachable;
+
+    if (experiments.isEnabled("dartfunctions")) {
+      await loadCloudRunServices(ctx, existingBackend, unreachableRegions, true);
+    }
   }
 
   ctx.existingBackend = existingBackend;
   ctx.unreachableRegions = unreachableRegions;
   return ctx.existingBackend;
+}
+
+/**
+ * Loads Cloud Run services into the existing backend.
+ * @param ctx Context from the Command library, used for caching.
+ * @param existingBackend The existing backend to load Cloud Run services into.
+ * @param unreachableRegions Object to track unreachable regions.
+ * @param onlyMissing If true, only loads missing Cloud Run services.
+ */
+async function loadCloudRunServices(
+  ctx: Context,
+  existingBackend: Backend,
+  unreachableRegions: { run: string[] },
+  onlyMissing: boolean,
+): Promise<void> {
+  try {
+    const runServices = await run.listServices(ctx.projectId);
+    for (const service of runServices) {
+      const endpoint = run.endpointFromService(service);
+      if (!onlyMissing || !existingBackend.endpoints[endpoint.region]?.[endpoint.id]) {
+        existingBackend.endpoints[endpoint.region] =
+          existingBackend.endpoints[endpoint.region] || {};
+        existingBackend.endpoints[endpoint.region][endpoint.id] = endpoint;
+      }
+    }
+  } catch (err: any) {
+    logger.debug(err.message);
+    unreachableRegions.run = ["unknown"];
+  }
 }
 
 /**
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -271,9 +271,15 @@ export async function prepare(
   // ===Phase 4. Fill in details and validate endpoints. We run the check for ALL endpoints - we think it's useful for
   // validations to fail even for endpoints that aren't being deployed so any errors are caught early.
   payload.functions = {};
+  // Resolve default regions for backends we want before grouping endpoints by codebase.
+  // This way, endpoints aren't incorrectly grouped together under the REGION_TBD region if the
+  // region is unresolved for multiple codebases.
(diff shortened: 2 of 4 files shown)


src/deploy/functions/prepare.ts
```diff
@@ -271,16 +271,19 @@ export async function prepare(
   // ===Phase 4. Fill in details and validate endpoints. We run the check for ALL endpoints - we think it's useful for
   // validations to fail even for endpoints that aren't being deployed so any errors are caught early.
   payload.functions = {};
+  const existingBackend = await backend.existingBackend(context);
+  for (const [, wantBackend] of Object.entries(wantBackends)) {
+    await resolveDefaultRegions(wantBackend, existingBackend);
```
ajperel: Maybe add a comment on why this is happening here since we've had a bug for it happening just a little later.

ajperel (commented):
Is it possible to add reasonable tests around the fixed functionality?

src/deploy/functions/prepare.ts
shettyvarun268: This might be a rare case, but matching regions across all existing functions could cause issues if someone has multiple codebases. If codebaseA and codebaseB both have a function with the exact same name, a new function in codebaseA with an unset region might accidentally get deployed in codebaseB's region just because the names match.

To keep them isolated, we can filter the existing functions to only match if they belong to the same codebase, or have no codebase label at all.

We can probably handle it like:
```js
const existingBackend = await backend.existingBackend(context);
for (const [codebase, wantBackend] of Object.entries(wantBackends)) {
  const relevantEndpoints = backend.allEndpoints(existingBackend).filter(
    (e) => e.codebase === codebase || e.codebase === undefined
  );
  await resolveDefaultRegions(wantBackend, backend.of(...relevantEndpoints));
}

Comment (wandamora) [reply src/deploy/functions/prepare.ts]: Thanks for the suggestion.
With this change I ran a scenario where I have two codebases with the same function signature, except one has europe-west1 set as the region and the other did not have one specified.

`firebase deploy --only functions:default` (default codebase using an HTTPS function in europe-west, aka a non-default region)
then
`firebase deploy --only functions:copy` (codebase containing no region definition).
image

**Review by ajperel — APPROVED** [review]
Thanks for adding tests

File: src/deploy/functions/backend.ts [hunk 2]
```diff
@@ -594,13 +583,46 @@ async function loadExistingBackend(ctx: Context): Promise<Backend> {
...
+    for (const service of runServices) {
+      const endpoint = run.endpointFromService(service);
+      if (!onlyMissing || !existingBackend.endpoints[endpoint.region]?.[endpoint.id]) {
+        existingBackend.endpoints[endpoint.region] =
+          existingBackend.endpoints[endpoint.region] || {};
+        existingBackend.endpoints[endpoint.region][endpoint.id] = endpoint;
+      }
+    }
+  } catch (err: any) {
+    logger.debug(err.message);
```
joehan: Super nit: Can be helpful to have a message around this so that it is easy to identify where the debug log is coming from - ie:

```
logger.debug(`Error loading Cloud Run services: ${err.message}`)
```

src/deploy/functions/prepare.ts
```diff
@@ -271,16 +271,25 @@ export async function prepare(
...
   // validations to fail even for endpoints that aren't being deployed so any errors are caught early.
   payload.functions = {};
+  // Resolve default regions for backends we want before grouping endpoints by codebase.
+  // This way, endpoints aren't incorrectly grouped together under the REGION_TBD region if the
+  // region is unresolved for multiple codebases.
+  const existingBackend = await backend.existingBackend(context);
+  for (const [codebase, wantBackend] of Object.entries(wantBackends)) {
+    const relevantEndpoints = backend
+      .allEndpoints(existingBackend)
+      .filter((e) => e.codebase === codebase || e.codebase === undefined);
```
joehan, line 281: Quick sanity check on the logic here - i'm not super familiar, so this might just read weird to me:
- We loop through the codebases, for each:
- we get all endpoints, and then filter down to just those with no codebase or the right codebase
- then we resolve the regions

Does this 'double resolve' the region for endpoints that do not have a codebase attached? Does that matter?

joehan (commented):
One question, but mostly lgtm. If my concern is not relevant, feel free to merge without me approving

src/deploy/functions/prepare.ts
wandamora: Only the wantBackend gets resolved, and the endpoints we are filtering down to are the ones we should match against per codebase. So no, there shouldn't be an issue with "double resolving" the want endpoints.

joehan (approved):

PR #10431, opened 2026-04-29
feat: Support creating search indexes for Firestore

rwhogg:
### Description

Adds support for creating Firestore search indexes

### Scenarios Tested

Tested search configs with and without other existing index field types.

### Sample Commands

firebase deploy --only firestore

### DIFF
--- a/src/firestore/api-sort.spec.ts
+++ b/src/firestore/api-sort.spec.ts
@@ -1,5 +1,6 @@
 import { expect } from "chai";
 
+import { Index, Order, QueryScope, TextIndexType, TextMatchType } from "./api-types";
 import { Backup, BackupSchedule, DayOfWeek } from "../gcp/firestore";
 import { durationFromSeconds } from "../gcp/proto";
 import * as sort from "./api-sort";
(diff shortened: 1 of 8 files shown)


src/firestore/api-sort.spec.ts
```diff
@@ -76,3 +77,42 @@ describe("compareApiBackupSchedule", () => {
...
+            textSpec: {
+              indexSpecs: [
+                { indexType: TextIndexType.TOKENIZED, matchType: TextMatchType.MATCH_GLOBALLY },
+              ],
+            },
+          },
+        },
+      ],
+    };
+    const indexWithSearch2: any = {
```
joehan: Use a specific type instead of any - is there a reason we can't just use index here?

Applies throughout

src/firestore/api.ts
```diff
@@ -30,21 +30,21 @@ export class FirestoreApi {
...
+    // this matches the direction of the last non-name non-vector/non-search field in the index.
     let fields = index.fields;
     const suffixOrder: types.Order = FirestoreApi.lastIndexFieldOrder(fields);
     const nameSuffix = { fieldPath: "__name__", order: suffixOrder } as types.IndexField;
 
     const lastField = index.fields?.[index.fields.length - 1];
-    if (lastField.vectorConfig) {
-      // lastField is vector field, refer to the second from last field
-      const vectorField = lastField;
+    if (lastField?.searchConfig || lastField?.vectorConfig) {
```
joehan: Now that we have 2 special types, can both show up in the same index? If so, i think this logic might be broken

src/firestore/api.ts
```diff
@@ -378,6 +378,19 @@ export class FirestoreApi {
         validator.assertType("vectorConfig.dimension", field.vectorConfig.dimension, "number");
         validator.assertHas(field.vectorConfig, "flat");
       }
+
+      if (field.searchConfig) {
+        if (field.searchConfig.textSpec) {
+          validator.assertHas(field.searchConfig.textSpec, "indexSpecs");
+          field.searchConfig.textSpec.indexSpecs.forEach((spec: any) => {
```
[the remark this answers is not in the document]
joehan: I think we could use `unknown` here instead of any - you are validating all the fields

joehan: Nit: Generally, prefer `for (const... of ...)` over `.forEach()` 

go/tsjs-style#iterating-containers

src/firestore/api.ts
bw-yuhogg: As it turns out, this method is only used for STANDARD databases and search indexes only apply to ENTERPRISE.

So most of these changes were just reverted.

joehan (approved):
LGTM - but please add a changelog entry before submitting

PR #10438, opened 2026-04-30
make local builds respect "ignore" files when uploading the built output

falahat:
### Description

We already respect the "ignore" entry in firebase.json for source deploys and we want to do the same for local builds.

One crucial difference. For local builds, we apply this filter AFTER the build, on the built output. We can not prevent files from being included in the build. 

### Scenarios Tested

* Test a local build with an ignore file
* Manually inspect the uploaded content


src/deploy/apphosting/deploy.ts
```diff
@@ -70,50 +71,67 @@ export default async function (context: Context, options: Options): Promise<void
...
+        context.backendStorageUris[cfg.backendId] =
+          `gs://${bucketName}/${path.basename(zippedSourcePath)}`;
+      }),
+    );
+  } finally {
+    // Clean up local build directories
+    const rootDir = options.projectRoot || process.cwd();
+    for (const cfg of Object.values(context.backendConfigs)) {
+      if (cfg.localBuild) {
+        const localBuildDir = path.join(rootDir, "local_build");
```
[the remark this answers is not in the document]
falahat: #wiz_ignore reason

In this case, the text is the user's text on their own machine.

src/deploy/apphosting/prepare.ts
```diff
@@ -377,3 +387,44 @@ async function ensureAppHostingServiceAgentRoles(
...
+ * source files that would be uploaded to Cloud Build.
+ */
+async function prepareLocalBuildScratchDirectory(
+  rootDir: string,
+  localBuildScratchDir: string,
+  cfg: AppHostingSingle,
+): Promise<void> {
+  // Resolve ignores for local builds, including default node_modules ignore
+  const ignore = resolveIgnorePatterns(cfg);
+  ignore.push(path.basename(localBuildScratchDir)); // Always ignore the build directory itself
```
annajowang: it looks like you're creating this in the project rootDir (where the firebase.json lives) (or cwd)

A single firebase.json and a single project directory can be associated with multiple app hosting backends. I believe this was useful for:
- monorepo projects => in which case the cfg.rootDir would be different
- but also for when a single app has both staging/prod app hosting backends associated with it => in which case their cfg.rootDir would be the same

also IIUC the CLI allows for deploying multiple backends at once. (Though I forget if this happens sequentially or async, i'd hope the latter for perf reasons)

tl;dr there might be a chance of another FAH backend's scratch dir getting included.

i think just updating this to ignore all dirs with the scratchdir prefix would cleanly solve this.

src/deploy/apphosting/prepare.ts
```diff
@@ -377,3 +387,44 @@ async function ensureAppHostingServiceAgentRoles(
...
+
+  // Check if local build scratch dir already exists
+  if (fs.existsSync(localBuildScratchDir)) {
+    throw new FirebaseError(
+      `The local build scratch directory '${localBuildScratchDir}' already exists. Please delete it and try again.`,
+    );
```
annajowang, line 412: I only see code to cleanup the scratchDir in release.ts

do we want to clean up scratchDir if the actual build here fails too? or was it not deleted to let the user debug the failure? if so, i think it's worth noting in a comment somewhere (maybe closer to the local build? no strong opinion)

a p2 optimization in that case is we ask the user for a simple "y" confirmation before deleting it for them.

annajowang (approved):
I'm gonna stamp this to reduce roundtrips and because this is experimental.

src/deploy/apphosting/prepare.ts
falahat: Ah, so prepare.ts is the first thing that runs.

If you run local builds but this folder already exists, we fail it so that we dont accidentally clobber user data. It's unlikely, but it's just in case they already had a folder named something like ".local_build...". 

However, if it doesn't exist, we will correctly create/set it up in prepare.ts. For the rest of the deployment, we expect the folder to exist. 

I like the idea of allowing them to confirm it if it's interactive mode. I might land that after this PR separately though

src/deploy/apphosting/prepare.ts
falahat: ah good catch, we might include the other build dirs.

to support parallel builds I actually changed it so the local build folder is unique to each backend. it's `.local_build_{backend_id}`. I'm going to also explicitly skip the other local build folders.

falahat: Done, I added the local build ignore lists by default

PR #10469, opened 2026-05-07
Resolve VPC Connector region if REGION_TBD

### LINKED ISSUE #10467: deploys with 15.17.0 fail without explicit `setGlobalOptions({ region: 'us-central1' })` with vpcConnector
Looks like our CI is using the latest 15.17.0 somehow and deploys are failing today when they worked fine on the previous release. 

It works if including `setGlobalOptions({ region: 'us-central1' })` in the index.ts file but this seems like a regression? 

**firebase-tools:** 15.17.0

**Platform:** Ubuntu (github CI)

### [REQUIRED] Test case
Deploy a function like this:
```
export const jobCreated = onDocumentCreated(
  {
    document: `jobs/{jobId}`,
    vpcConnector: 'connector-name',
    vpcConnectorEgressSettings: 'PRIVATE_RANGES_ONLY',
  },
  onCreateJob
)
```

Try deploying a function defined like that without `setGlobalOptions({ region: 'us-central1' })`

These are existing functions already in `us-central1` - the same region as the VPC connector. 

Function deploys as expected. 

Errors with: `functions: Request to https://cloudfunctions.googleapis.com/v2/projects/%%service-url%% had HTTP Error: 400, Could not update Cloud Run service servicename. spec.template.metadata.annotations: The target region us-central1 for the deployment must be the same as the reg`

wandamora:
This dynamically resolves the VPC connector region to be the same as the function if not defined.

This also includes a small comment fix for the region field in Endpoint.

### Scenarios Tested
Create function with VPC connector
`firebase deploy --only functions` on 15.16.0
Update function and use local firebase-tools build
Update was successful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Set region for VPC connector to match function region if not specified. (#10469)
--- a/src/deploy/functions/build.ts
+++ b/src/deploy/functions/build.ts
@@ -257,8 +257,8 @@ export type Endpoint = Triggered & {
   // Defaults to the compute service account when a function is first created as a GCF gen 2 function.
   serviceAccount?: ServiceAccount | Expression<string> | null;
 
-  // defaults to ["us-central1"], overridable in firebase-tools with
-  //  process.env.FIREBASE_FUNCTIONS_DEFAULT_REGION
+  // Defaults to REGION_TBD. The deployment region is resolved dynamically at deploy-time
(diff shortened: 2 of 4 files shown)


shettyvarun268 (approved):
LGTM!
Might want to re-run the failing checks and update the changelog for this

wandamora:
Closing in favor of #10471

PR #10471, opened 2026-05-07
refactor: resolve function regions during build phase to fix VPC connectors

wandamora:
This moves the resolution logic to be before we resolve the build to a backend to set the defaults in functions, and the correct region can be known when building the VPC connector identfier.

### Scenarios Tested

Create function with VPC connector
`firebase deploy --only functions on 15.16.0`
Update function and use local firebase-tools build
Update was successful.

### DIFF
--- a/src/deploy/functions/prepare.spec.ts
+++ b/src/deploy/functions/prepare.spec.ts
@@ -186,8 +186,6 @@ describe("prepare", () => {
     });
   });
 
-
-
   describe("resolveDefaultRegionsForBuild", () => {
     let sandbox: sinon.SinonSandbox;
     let getDatabaseStub: sinon.SinonStub;
--- a/src/deploy/functions/prepare.spec.ts
+++ b/src/deploy/functions/prepare.spec.ts
@@ -203,7 +203,7 @@ describe("prepare", () => {
       sandbox.restore();
     });
 
-    it("does nothing if no endpoints", async () => {
+    it("does nothing if no endpoints or in REGION_TBD", async () => {
       const want = build.empty();
       const have = backend.empty();
       await prepare.resolveDefaultRegionsForBuild(want, have);
@@ -218,6 +218,7 @@ describe("prepare", () => {
           project: "project",
           runtime: latest("nodejs"),
           httpsTrigger: {},
+          region: [build.REGION_TBD],
         },
       });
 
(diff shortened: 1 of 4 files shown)


shettyvarun268 (approved):
LGTM

shettyvarun268:
Might want to rebase this PR to take in Joe's changes for the test cases failing.

src/deploy/functions/prepare.ts
```diff
@@ -349,83 +350,52 @@ export async function prepare(
...
-        if (matching) {
-          throw new FirebaseError(
-            `Cannot resolve default region for function ${id}. It exists in multiple regions. The region must be specified to continue.`,
+
+      if (matching) {
+        resolvedRegion = matching.region;
+      } else {
+        // Match triggers.
+        try {
+          const fullEndpoint = { ...endpoint, id } as any;
```
ajperel, line 383: This is done because `endpoint` (of type `build.Endpoint`) does not have an `id` field, but `resolveRegionForEventTrigger` expects a `backend.Endpoint` which requires an id. While this works, it bypasses TypeScript's type checks. Consider defining a minimal interface for what `resolveRegionForEventTrigger` actually needs (it seems to only need `project` and `eventTrigger`) or casting to a specific type that fulfills that contract rather than `any`.

src/deploy/functions/prepare.ts
```diff
@@ -349,83 +350,52 @@ export async function prepare(
...
+          if (build.isBlockingTriggered(endpoint)) {
+            resolvedRegion = resolveRegionForBlockingTrigger(fullEndpoint);
+          } else if (build.isEventTriggered(endpoint)) {
+            resolvedRegion = await resolveRegionForEventTrigger(fullEndpoint);
```
ajperel, line 387: If a user has many event-triggered functions, this will execute network calls (like fetching database or bucket locations) sequentially (because of the for loop it's in), which can slow down the prepare phase. Consider gathering the promises and using Promise.all to resolve them in parallel.

src/deploy/functions/prepare.spec.ts
```diff
@@ -253,74 +203,122 @@ describe("prepare", () => {
...
-      await prepare.resolveDefaultRegions(want, have);
+      await prepare.resolveDefaultRegionsForBuild(want, have);
 
-      expect(want.endpoints["us-east1"]?.["id"]).to.exist;
-      expect(want.endpoints["us-east1"]?.["id"].region).to.equal("us-east1");
-      expect(want.endpoints[build.REGION_TBD]).to.not.exist;
+      expect(want.endpoints["id"].region).to.deep.equal(["us-east1"]);
+    });
+
+    it("resolves region to us-east1 and correctly formats VPC connector path with us-east1", async () => {
```
ajperel, line 233: AI suggests also testing this case:

Pre-formatted VPC Connector Paths:
Scenario: An endpoint with a vpc.connector that is already a full resource path (e.g., projects/my-project/locations/us-east1/connectors/my-connector).
Purpose: Verify that the resolution logic does not attempt to prefix it again or corrupt it, even after the region is resolved. This ensures the check connector.includes("/") in build.ts works correctly in this flow.

ajperel (commented):
My AI review prompt had a few more good suggestions we can consider for a follow up PR.

src/deploy/functions/prepare.ts
wandamora: Daniel once advised against using Promise.all() since transient errors could bubble up in the CLI in a way that looks cryptic to the end user.

wandamora: https://github.com/firebase/firebase-tools/pull/10414#discussion_r3158092429 for reference

ajperel: Ok. That's a good counter argument. I think it'd be good for us to see what the cryptic error is and if we could fix it so things can be faster. But I guess let's save that for a day when we need to improve performance.

src/deploy/functions/prepare.ts
wandamora: Follow up: https://github.com/firebase/firebase-tools/pull/10504

src/deploy/functions/prepare.spec.ts
wandamora: Follow up: https://github.com/firebase/firebase-tools/pull/10505

PR #10477, opened 2026-05-08
security: ignore symlinks by default in fsAsync.readdirRecursive

wecbaiyk-blip:
﻿## Summary

`fsAsync.readdirRecursive` is the directory walker used by every firebase-tools deploy that enumerates a source directory. It already supports an `ignoreSymlinks` option, but that option defaults to `false`. Three of the four current callers omit the option entirely, so symlinks in the source tree are followed during App Hosting deploys (`createLocalBuildTarArchive`, `createSourceDeployArchive`) and Cloud Functions deploys (`prepareFunctionsUpload`).

This PR flips the default to `true`. Callers that legitimately need symlink-following can opt in explicitly with `ignoreSymlinks: false`.

## Why

A symlink inside the source tree pointing at `/proc/self/environ`, `~/.config/gcloud/application_default_credentials.json`, `~/.ssh/id_rsa`, or similar will be followed by the recursion and its target''s contents uploaded to whatever destination the deploy targets (Firebase Hosting public bucket, App Hosting build bucket, Cloud Functions GCS staging bucket, etc.). The attack surface is largest in CI workflows that extract attacker-supplied tarballs (e.g. PR-preview deploys) but applies anywhere a deploy operates on attacker-influenced files.

The one caller that already does the right thing (`src/archiveDirectory.ts`) keeps working unchanged because it explicitly passes `ignoreSymlinks: true`.

Caller audit:

| Caller | Passed `ignoreSymlinks: true`? |
|---|---|
| `src/archiveDirectory.ts:79` | yes |
| `src/deploy/apphosting/util.ts:32` (`createLocalBuildTarArchive`) | no |
| `src/deploy/apphosting/util.ts:98` (`createSourceDeployArchive`) | no - error message at line 113 even tells the user to "Remove links and shortcuts and try again", suggesting symlinks were known to be problematic but weren''t blocked |
| `src/deploy/functions/prepareFunctionsUpload.ts:117` | no |

## Changes

- `src/fsAsync.ts`: default `ignoreSymlinks` to `true`; emit a `debug`-level log when a symlink is dropped; add security-rationale JSDoc on the option and the function.
- `src/fsAsync.spec.ts`: replace the existing single symlink test with two - one asserting the new safer default, one asserting that `ignoreSymlinks: false` still includes symlinks for users who explicitly opt in.

No changes to call sites are required for the security fix: the three currently-unsafe callers (`apphosting/util.ts` x2, `prepareFunctionsUpload.ts`) inherit the safer default for free.

## Backward compatibility

The behavior changes only for callers who (a) didn''t pass `ignoreSymlinks` at all and (b) have a source tree that contains symlinks they want followed during deploy. Both conditions together are a narrow population; the typical project tree is unaffected. Users who rely on symlink-following can keep the old behavior by passing `ignoreSymlinks: false` explicitly.

If the maintainers prefer a warn-then-deprecate path instead (default stays `false`, log a warning when a symlink is encountered, flip the default in a future major), happy to adjust.

## Testing

Two new tests in `src/fsAsync.spec.ts`:
- "ignores symlinks by default (security: prevents symlink-following during deploy)" - asserts the new default.
- "includes symlinks only when ignoreSymlinks is explicitly set to false" - asserts the explicit opt-out still works.


wecbaiyk-blip:
@googlebot I signed it!

wecbaiyk-blip:
Rebased onto current `main` and resolved the conflict in `src/fsAsync.ts`.

The add/add conflict was in the `readdirRecursiveHelper(...)` call: `main` added the `supportGitIgnore` / `gitIgnoreStack` fields (and passed `ignoreSymlinks: !!options.ignoreSymlinks`), while this PR introduces the secure default `const ignoreSymlinks = options.ignoreSymlinks !== false`. Resolved by keeping `main`s new fields **and** passing the secure-default `ignoreSymlinks` variable (so symlinks are ignored unless a caller explicitly opts in with `false`). Also dropped an empty `trigger CLA recheck` commit. CLA is green. Ready for re-review.

PR #10493, opened 2026-05-12
fix failing tests if fdcrealtime experiment is disabled locally

aalej:
### Description

Some tests were failing because I had the `fdcrealtime` disabled locally

### Scenarios Tested

`npm run test`


IzaakGough (approved):
Lgtm!

PR #10498, opened 2026-05-12
support nested .gitignore files in source deploys

### LINKED ISSUE #10483: App Hosting deploy .zip includes files specified in .gitignore from subdirectories
When zipping the source to upload to App Hosting, the .gitignore in children directories of the root path are ignored. This is unexpected because it doesn't align with how .gitignore files are used in git and can lead to unintentionally including secrets in the zip file.

The fix would be making [parseGitIgnorePatterns](https://github.com/firebase/firebase-tools/blob/cf9e7e651e6cf796d4446bbd17b6279225f37f9e/src/deploy/apphosting/util.ts#L120) include .gitignore files in children directories.

**firebase-tools:** 15.17.0
**Platform:** macOS

### [REQUIRED] Test case

Directory structure:

./a/.gitignore
./a/.env
./.gitignore
./b.txt

-- Content of ./.gitignore 
b.txt

-- Content of ./a/.gitignore 
.env

firebase deploy --only apphosting

.env is not included in the zip file

.env is included in the zip file

annajowang:
### Description
Enables dynamic cascading discovery and stacked processing of nested `.gitignore` files during recursive directory scanning in `fsAsync.readdirRecursive`. This ensures nested directories respect local `.gitignore` definitions, negation overrides, and scoped anchored rules inside child directories, aligning exactly with Git's standard specifications.

Also performs extensive code cleanups:
- Renamed `isGitIgnore` to `supportGitIgnore` for clear intent.
- Renamed `ignore` to `ignoreStrings` to avoid confusion with ignore engines.
- Renamed `rules` to `ignoreRules` for explicit matching clarity.
- Unified all GitIgnore matching logic into a single stack-based traversal, removing redundant parallel checks and obsolete manual parsers from callers.

### Scenarios Tested
- Discovered and parsed `.gitignore` files dynamically inside nested subdirectories.
- Scoped subdirectory paths and matched anchored rules correctly.
- Evaluated nested negation rules (`!file`) and prioritize child directory rules over parent ignores correctly.
- Handled subdirectories without `.gitignore` files and directory-named `.gitignore` folders gracefully.

### Sample Commands
`npm test`
`npx mocha src/fsAsync.spec.ts`

### DIFF
--- a/src/fsAsync.ts
+++ b/src/fsAsync.ts
@@ -1,15 +1,20 @@
-import { readdirSync, statSync } from "fs-extra";
+import { readdirSync, statSync, readFileSync } from "fs-extra";
 import ignorePkg from "ignore";
 import * as _ from "lodash";
 import * as minimatch from "minimatch";
 import { join, relative } from "path";
+import { logger } from "./logger";
 
 export interface ReaddirRecursiveOpts {
   // The directory to recurse.
   path: string;
-  // Files to ignore.
-  ignore?: string[];
-  isGitIgnore?: boolean;
+  // Array of glob patterns representing files and folders to ignore.
+  // Note: The matching behavior changes depending on the supportGitIgnore option.
+  ignoreStrings?: string[];
+  // When true, ignoreStrings are matched using the GitIgnore spec rules (supporting negations and scope priorities)
+  // and local nested .gitignore files are discovered and stacked automatically.
+  // When false or undefined, ignoreStrings are matched flatly using simple minimatch wildcard globs.
+  supportGitIgnore?: boolean;
   // Files in the ignore array to include.
   include?: string[];
   // Maximum depth to recurse.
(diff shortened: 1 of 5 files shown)


src/fsAsync.ts
```diff
@@ -70,28 +109,50 @@ export async function readdirRecursive(
...
-    if (options.isGitIgnore) {
-      // the git ignore filter will return true if given path should be included,
-      // so we need to negative that return false to avoid filtering it.
-      return !gitIgnoreRules(relative(options.path, t));
+  const filter = (targetPath: string, gitIgnoreStack: GitIgnoreState[] = []): boolean => {
+    if (options.supportGitIgnore) {
+      let ignored = false;
+      // Loop in reverse (from deepest subdirectory to root) so that local subdirectory
+      // rules take precedence and override any parent directory rules.
+      // Loop using a decrementing index to avoid allocating array copies or reversing in the hot path.
```
falahat: wait wouldn't we want the most nested .gitignore to act *last* if we want it to have top priority? That way it would override any existing values?

falahat (approved):

src/fsAsync.ts
annajowang: We terminate early. so as long as an ignore rule applies, we apply it and move on.

PR #10501, opened 2026-05-12
Adding default timeout for dart functions

### LINKED ISSUE #10487: Default timeout for Dart functions is not 60 seconds

**firebase-tools:** 15.17.0

**Platform:** macOS

### [REQUIRED] Test case

```dart
import 'package:firebase_functions/firebase_functions.dart';

void main() {
  runFunctions((firebase) {
    firebase.https.onRequest(name: 'helloWorld', (_) async {
      return Response(200, body: 'Hello from Dart Functions!');
    });
  });
}

```

1. Deploy the sample code to google cloud via `firebase deploy --only functions`
2. See in [Google Cloud Console](https://console.cloud.google.com/run/deploy/us-central1/hello-world) under **Requests** that timeout is 300 seconds

The default timeout should be 60 seconds similar to node.js functions

Image

The timeout is set to 300 seconds

Image

shettyvarun268:
Fixes: #10487 

This PR refactors the handling of default timeouts for functions deployed to the Cloud Run platform (such as Dart functions). To improve maintainability, it introduces a dedicated resolveDefaultTimeout method in the Firebase CLI (src/deploy/functions/prepare.ts) to apply a 60-second default timeout to all functions with platform: "run" if not specified, matching the expectation set by Node.js functions.

Additionally, the PR preserves existing timeouts for already deployed functions. If a function is already live in the cloud, the CLI will keep that timeout if the local code does not specify a new one, rather than overriding it with the new default.

Comprehensive unit tests have been added to prepare.spec.ts and updated in index.spec.ts to verify these behaviors, and all existing tests passed successfully.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,3 @@
 - Updated Pub/Sub emulator to version 0.8.31
 - Resolves undefined regions earlier, during the build to backend resolution phase (#10471)
-- Default timeout for Dart functions is now 60 seconds when not explicitly set (#10506)
+- Default timeout for Dart functions is now 60 seconds when not explicitly set (#10501)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,3 @@
 - Updated Pub/Sub emulator to version 0.8.31
 - Resolves undefined regions earlier, during the build to backend resolution phase (#10471)
+- Default timeout for Dart functions is now 60 seconds when not explicitly set (#10506)
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -302,6 +302,7 @@ export async function prepare(
     inferDetailsFromExisting(wantBackend, haveBackend, codebaseUsesEnvs.includes(codebase));
     await ensureTriggerRegions(wantBackend);
     resolveCpuAndConcurrency(wantBackend);
+    resolveDefaultTimeout(wantBackend);
     validate.endpointsAreValid(wantBackend);
     inferBlockingDetails(wantBackend);
   }
@@ -651,10 +652,15 @@ export function resolveCpuAndConcurrency(want: backend.Backend): void {
     if (!e.concurrency) {
       e.concurrency = e.cpu >= 1 ? backend.DEFAULT_CONCURRENCY : 1;
     }
+  }
+}
 
-    // Cloud Run defaults to 300s timeout if not specified. For functions deployed
-    // directly to Cloud Run (like Dart), we want to enforce the same 60s default
-    // that GCF enforces for consistency.
+/**
+ * Assigns the default timeout to a function if it is deployed to Cloud Run
+ * and no timeout was specified.
+ */
+export function resolveDefaultTimeout(want: backend.Backend): void {
+  for (const e of backend.allEndpoints(want)) {
     if (e.platform === "run" && e.timeoutSeconds === undefined) {
       e.timeoutSeconds = backend.DEFAULT_TIMEOUT_SECONDS;
     }
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -552,6 +552,10 @@ export function inferDetailsFromExisting(
       wantE.cpu = haveE.cpu;
     }
 
+    if (typeof wantE.timeoutSeconds === "undefined" && haveE.timeoutSeconds) {
+      wantE.timeoutSeconds = haveE.timeoutSeconds;
+    }
+
     // N.B. concurrency has different defaults based on CPU. If the customer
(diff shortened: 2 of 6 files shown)


src/deploy/functions/runtimes/dart/index.ts
```diff
@@ -390,6 +390,9 @@ export class Delegate implements runtimes.RuntimeDelegate {
     const isEmulator = envs.FUNCTIONS_EMULATOR === "true";
     if (!isEmulator) {
       for (const ep of Object.values(discovered.endpoints)) {
+        if (ep.timeoutSeconds === undefined) {
+          ep.timeoutSeconds = 60;
```
wandamora: optional nit: maybe this timeout should be a global constant

wandamora (changes requested):

wandamora (approved):
LGTM  % small nit

src/deploy/functions/runtimes/dart/index.ts
ajperel: We should add a comment on _why_ we're doing this, e.g., to be consistent with v1 & v2 timeouts of 60 where the default is 300.

src/deploy/functions/runtimes/dart/index.ts
```diff
@@ -390,6 +390,9 @@ export class Delegate implements runtimes.RuntimeDelegate {
     const isEmulator = envs.FUNCTIONS_EMULATOR === "true";
     if (!isEmulator) {
       for (const ep of Object.values(discovered.endpoints)) {
+        if (ep.timeoutSeconds === undefined) {
```
ajperel: This logic is as written specific to Dart, but if we start deploying Node functions using Cloud Run Functions it'll have the same problem right? Can we make this at a higher level where it applies to all "run" deployed functions.

CHANGELOG.md
```diff
@@ -1,2 +1,3 @@
 - Updated Pub/Sub emulator to version 0.8.31
 - Resolves undefined regions earlier, during the build to backend resolution phase (#10471)
+- Default timeout for Dart functions is now 60 seconds when not explicitly set (#10501)
```
ajperel, line 3: Thank you for remembering to do this!

src/deploy/functions/runtimes/dart/index.ts
shettyvarun268: We can add this condition a bit higher in the chain to ensure that the 60s timeout is applied any time a run function is deployed. Just testing out the new code before pushing it through.

src/deploy/functions/prepare.ts
```diff
@@ -647,6 +651,13 @@ export function resolveCpuAndConcurrency(want: backend.Backend): void {
     if (!e.concurrency) {
       e.concurrency = e.cpu >= 1 ? backend.DEFAULT_CONCURRENCY : 1;
     }
+
+    // Cloud Run defaults to 300s timeout if not specified. For functions deployed
```
ajperel: I'm not sure myself... do you think this comment is better here or on the DEFAULT_TIMEOUT_SECONDS declaration?

src/deploy/functions/prepare.ts
```diff
@@ -647,6 +651,13 @@ export function resolveCpuAndConcurrency(want: backend.Backend): void {
...
+    // directly to Cloud Run (like Dart), we want to enforce the same 60s default
+    // that GCF enforces for consistency.
+    if (e.platform === "run" && e.timeoutSeconds === undefined) {
```
ajperel, line 664: I think this works, but we're doing something pretty unintuitive here because we're shoving the logic into a method named `resolveCpuAndConcurrency` which is *all* this method did before. Someone reading this wouldn't expect it to also adjust a timeout.

We need to either update this method name to something more general that correctly identify what it does OR do this in a separate (new or more appropriate) method like `resolveDefaultTimeout`

ajperel (commented):
Thanks for the quick work. A few more nits.

src/deploy/functions/prepare.ts
shettyvarun268: I think the comment makes sense here because the DEFAULT_TIMEOUT_SECONDS is declared in a different file altogether and its not being used there at all. For someone who might be going through the code, it might make more sense to know why we are doing this at the place of execution rather than a place where this constant is just being declared.

src/deploy/functions/prepare.ts
shettyvarun268: Done. Added a separate method to handle the default time out

src/deploy/functions/prepare.ts
ajperel: OK. Minor and probably fine either way :)

ajperel (approved):

PR #10503, opened 2026-05-12
Set requiredProjectBindings in AI Logic services

Berlioz:
In the prepare phase of function deployment, the CLI calls [ensureServiceAgentRoles](https://github.com/firebase/firebase-tools/blob/0b59f188861ac6e9ccb38c8f44398bacacd44383/src/deploy/functions/checkIam.ts#L239) which checks and upserts all requiredProjectBindings for all Services in the deployed Backend.

Projects containing an AI Logic function must grant the `roles/run.invoker` binding to the AI Logic service account.

### DIFF
--- a/src/deploy/functions/services/ailogic.ts
+++ b/src/deploy/functions/services/ailogic.ts
@@ -2,6 +2,7 @@ import * as backend from "../backend";
 import { FirebaseError, getErrStatus } from "../../../error";
 import { Name, Service } from "./index";
 import * as ailogicApi from "../../../gcp/ailogic";
+import * as iam from "../../../gcp/iam";
 import {
   AI_LOGIC_BEFORE_GENERATE_CONTENT,
   AI_LOGIC_AFTER_GENERATE_CONTENT,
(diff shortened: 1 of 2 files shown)


src/deploy/functions/services/ailogic.ts
```diff
@@ -51,6 +52,22 @@ export class AILogicService implements Service {
   ensureTriggerRegion: (ep: backend.Endpoint & backend.EventTriggered) => Promise<void> = () =>
     Promise.resolve();
 
+  /**
+   * The AI logic proxy server uses a service account to invoke functions.
+   * Setting requiredProjectBindings here causes the ensureServiceAgentRoles
+   * call during prepare phase to upsert the corresponding IAM binding.
+   */
+  async requiredProjectBindings(projectNumber: string): Promise<Array<iam.Binding>> {
```
ajperel, line 60: I realize this is a pretty simple method. Do you think it's worth writing a test for it?

src/deploy/functions/services/ailogic.ts
```diff
@@ -51,6 +52,22 @@ export class AILogicService implements Service {
...
+    return [
+      {
+        role: "roles/run.invoker",
+        members: [
+          `serviceAccount:service-${projectNumber}@gcp-sa-firebasevertexai.iam.gserviceaccount.com`,
```
ajperel, line 65: I assume we have to hard code this and there's no good way to look this up or derive this that's more resilient if the service account name changes in the future? If there is, we should do that.

ajperel (approved):

src/deploy/functions/services/ailogic.ts
Berlioz: Honestly, why not. Done.

src/deploy/functions/services/ailogic.ts
Berlioz: Yeah, hardcoded service account forms are rampant through firebase-tools since it's not like there's an authoritative endpoint you can hit or whatever that lists them.

PR #10504, opened 2026-05-12
refactor: clean up type casting in region resolution

wandamora:
### Description

This PR refactors the region resolution logic in `src/deploy/functions/prepare.ts` to eliminate unneeded object spreading (`{ ...endpoint, id } as any`) and loose type casting. Instead of forcing regional helper functions to accept a monolithic `backend.Endpoint` union type, the helpers have been updated to accept only the specific properties or trigger configurations they actually require.

### Changes

* **`src/deploy/functions/services/ailogic.ts`**:
  * Renamed `isGlobalAILogicEndpoint` to `isGlobalAILogicTrigger` and refactored it to accept a `build.BlockingTrigger` instead of a full `backend.Endpoint`.

* **`src/deploy/functions/prepare.ts`**:
  * Removed the `fullEndpoint` hacks in the trigger matching loop.
  * Updated `resolveRegionForBlockingTrigger` to take a `build.BlockingTrigger`.
  * Updated `resolveRegionForEventTrigger` to explicitly take the `project` string and `build.EventTrigger` object instead of an intersected endpoint type.


src/deploy/functions/prepare.ts
```diff
@@ -399,9 +414,10 @@ export async function resolveDefaultRegionsForBuild(
   }
 }
 
-function resolveRegionForBlockingTrigger(
-  endpoint: backend.Endpoint & backend.BlockingTriggered,
-): string {
+function resolveRegionForBlockingTrigger(endpoint: TriggerResolutionTarget): string {
+  if (!endpoint.blockingTrigger) {
```
ajperel: I don't love this. Like we should only be calling this for blocking triggers and this was enforced by the old code right with a type of `backend.Endpoint & backed.blockingTriggered`

This seems to open us up to subtle bugs in the future where we call the wrong region resolution method and then incorrectly fall back to the default function region.

ajperel: The root problem seems to be that we're trying to construct a single `endpoint` ish object that has everything the method needs. But we don't really have to do that right? I realize the overall thing is working w/ endpoints and backends. But we don't have to _only_ pass those around.

What if these methods just took individual args for _only_ the info the needed.

E.g. this one only needs the `blockingTrigger` which already has a type and the `resolveRegionForEventTrigger` only needs `eventTrigger` and `project`?

I guess the issue with that is nominally we need a _single_ interface if we want to in our next refactoring move these methods onto the service so we just call 

`service.resolveRegionForEventTrigger`

But we could just also provide `project` for blocking triggers even though it's not used?

Thoughts?

wandamora: Switching to individual args makes the code much cleaner.

I think when we eventually refactor this to be part of the service, we can continue to use `backend.Endpoint` as the arg for a `resolveRegion` function, and extract the information from the endpoint we need per service, like the rest of the API calls in the Service interface: https://github.com/firebase/firebase-tools/blob/750a8abd51fae4d0c57875baa7029a4fca2ba40f/src/deploy/functions/services/index.ts#L35

ajperel (approved):
Glad the suggestions were useful. Thanks for cleaning this up :)

PR #10542, opened 2026-05-21
Update index definition reference in README.md

jlara310:
Reconcile this page with [documentation](https://firebase.google.com/docs/reference/firestore/indexes). Update index definition for Native mode support in Enterprise edition.


IzaakGough (changes requested):
Thank you for contributing! 

Can we address the comments from Gemini? These look like reasonable suggestions to me.
One additional mismatch: the field-mode note still only calls out order and arrayConfig, but validateIndex() also accepts vectorConfig and searchConfig. Could that be updated too?

PR #10548, opened 2026-05-22
Clean up Firebase Init Dataconnect flow w/ AgentService

hlshen:
* Remove CloudAICompanion references
* * Remove enablement checks
* * Remove all remaining code
* Move schema deploy as a separate process to generateQuery, now that generateQuery accepts schema files
* Add a new heuristic for getting gql code, in case backticks are missing

https://screencast.googleplex.com/cast/NTM5NjY5NDQ3OTkyOTM0NHxkZDk4ODgwNy1jZQ

Screenshot 2026-05-22 at 4 03 43 PM
Screenshot 2026-05-22 at 4 03 55 PM


firebase-vscode/src/data-connect/execution/execution.ts
```diff
@@ -318,7 +317,7 @@ ${arg.existingQuery ? `\n\nRefine this existing operation:\n${arg.existingQuery}
       const serviceName = await dataConnectService.servicePath(
         arg.document.fileName,
       );
-      if (!(await ensureGIFApiTos(arg.projectId))) {
+      if (!configstore.get("gemini")) {
```
fredzqm: Do we still need this ToS acceptance page?

When will we remove showGiFToSModal?

src/gemini/fdcExperience.ts
```diff
@@ -243,5 +243,14 @@ export function extractCodeBlock(text: string): string {
...
+
+  // Loose parsing if no backticks are present
+  if (text.includes("{")) {
+    // Scenario B: No backticks, but has "{". Assume entire response is GraphQL.
+    return text.trim();
+  }
+
+  // Scenario C: No backticks, no "{". Just text.
+  logger.debug("[Agent Service] Response seems to be plain text, no GraphQL code block found.");
   return text.trim();
```
fredzqm, line 253: nit: `Scenario B` and `Scenario C` both return `text.trim()`. Feels like the same scenario.

So the condition really only impact debug logging. Maybe tweak the branch to make that more obvious. If this case is super rare in practice, we can log at warning level.

```
if (!text.includes("{")) {
   // logging...
}
```

src/init/features/dataconnect/index.ts
```diff
@@ -288,32 +287,64 @@ async function actuateWithInfo(
...
+          generateOperation(PROMPT_GENERATE_SEED_DATA, serviceName, projectId, schemas),
+        ]);
+        return { op, seed, success: true };
+      } catch (err: any) {
+        return { success: false, error: err };
+      }
+    })();
+
+    // Await both to complete
+    const [, opResult] = await Promise.all([deployPromise, opPromise]);
```
fredzqm, line 339: Interesting. Nice improvement. We can parallelize deployment and query generation.

nit: Can you keep combine them into one Spinner?

```
  const [_, operationGql, seedDataGql] = await promiseWithSpinner(
      () =>
        Promise.all([
          deploySchema(),    
          generateOperation(PROMPT_GENERATE_CONNECTOR, serviceName, projectId),
          generateOperation(PROMPT_GENERATE_SEED_DATA, serviceName, projectId),
        ]),
      "Deploy SQL Connect Schemas and Generating Operations...",
    );
```

We shouldn't need the complexities of tracking errors like `genOperationsError = opResult.error;` Any errors thrown from those promise can be surfaced to the top.

firebase-vscode/src/data-connect/execution/execution.ts
hlshen: good catch, removing

src/init/features/dataconnect/index.ts
hlshen: I'm confused by the benefit of this? The promises are still in one spinner on line 339.

Tracking generationError separately just to avoid writing to file in case of failure.

fredzqm: Code Health comment. We can achieve the same algorithm with implicit exceptions thrown.

The local variables & branches make the code hard to follow.
```
  let operationGql = "";
  let seedDataGql = "";
  let genOperationsSuccess = false;
  let genOperationsError: unknown = null;
```

fredzqm (approved):
LGTM. Excited to subset of the legacy cloudaicompanion API!

One code health comment remaining.

fredzqm:
BTW, I was debugging a bug from main branch and filed this.

http://b/516820861 Firebase Data Connect CLI: fb init fails on 'Generating Operations' while Cloud SQL is PENDING_CREATE

Slight related to the PR. It caught a pretty bad bug in the backend returning `PENDING_CREATE` error.

PR #10555, opened 2026-05-26
fix: handle cloud-sql-connector unhandled exceptions on close

### LINKED ISSUE #10549: firebase deploy --only dataconnect fails with "CloudSQLConnectorError: The connector was closed" (transitive @google-cloud/cloud-sql-connector ≥1.7.0)

**firebase-tools:** 15.18.0 (also reproduces on 15.17.0)

**Platform:** macOS (Darwin 25.5.0), Node.js 24.16.0 (also reproduces on 25.8.1)

**Cloud SQL instance:** Postgres 17.9, IAM auth enabled (`cloudsql.iam_authentication=on`), public IP

### [REQUIRED] Test case

Any project with a deployed Data Connect service backed by Cloud SQL Postgres, deploying via:

```sh
firebase deploy --only dataconnect --project <project-id>
```

1. Have an existing Data Connect service with a Cloud SQL Postgres instance (IAM auth on).
2. Run `firebase deploy --only dataconnect --project <project-id>`.
3. Deploy proceeds past connector deployment, reaches the schema-migration step ("Generating SQL schema migrations to be compatible..."), then fails immediately with `Error: An unexpected error has occurred.`

Deploy completes successfully; `firebase-tools` should not surface a connector teardown error as a deploy failure.

`--debug` stack trace:

```
[…] Logged in as <iam-user>
[…] > SELECT pg_get_userbyid(nspowner)
          FROM pg_namespace
          WHERE nspname = 'public';
[…] (query returns in ~13ms)
[…] CloudSQLConnectorError: The connector was closed.
    at CloudSQLInstance.close (.../@google-cloud/cloud-sql-connector/dist/cjs/cloud-sql-instance.js:259:32)
    at Connector.close (.../@google-cloud/cloud-sql-connector/dist/cjs/connector.js:252:32)
    at cleanUpFn (.../firebase-tools/lib/gcp/cloudsql/connect.js:70:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async execute (.../firebase-tools/lib/gcp/cloudsql/connect.js:91:5)
    at async executeSqlCmdsAsIamUser (.../firebase-tools/lib/gcp/cloudsql/connect.js:98:12)
    at async getSchemaMetadata (.../firebase-tools/lib/gcp/cloudsql/permissionsSetup.js:139:31)
    at async setupSchemaIfNecessary (.../firebase-tools/lib/dataconnect/schemaMigration.js:31:28)
    at async migrateSchema (.../firebase-tools/lib/dataconnect/schemaMigration.js:137:5)
    at async default_1 (.../firebase-tools/lib/deploy/dataconnect/release.js:53:9)

```

### Root cause

This is a transitive-dependency regression. `firebase-tools` declares `"@google-cloud/cloud-sql-connector": "^1.3.3"`, which today resolves to **1.11.0**.

In `@google-cloud/cloud-sql-connector >= 1.7.0`, `CloudSQLInstance.close()` calls `socket.destroy(new CloudSQLConnectorError({ code: 'ERRCLOSED', … }))` on every tracked socket. That schedules an `'error'` event via `process.nextTick`. By that time, `cleanUpFn` in `firebase-tools/lib/gcp/cloudsql/connect.js:67-71` has already `await`-ed `pool.end()`, and the `pg` driver has detached its error listeners. When the scheduled `'error'` event fires next tick with no listener, Node escalates it to an uncaught exception, which bubbles up through `executeSqlCmdsAsIamUser` and surfaces as `Error: An unexpected error has occurred.`

Bisected on the connector with `firebase-tools` held constant:

| `@google-cloud/cloud-sql-connector` | result |
|---|---|
| ≤ 1.6.1 | works (`Connector.close()` calls `socket.destroy()` with no argument) |
| ≥ 1.7.0 | broken (introduced `socket.destroy(new CloudSQLConnectorError(...))`) |

Upstream issue already filed and tracking the connector-side fix: https://github.com/GoogleCloudPlatform/cloud-sql-nodejs-connector/issues/576

### Workaround

Pin the transitive connector inside the global `firebase-tools` install:

```sh
cd $(npm root -g)/firebase-tools
npm install --no-save @google-cloud/cloud-sql-connector@1.6.1
```

After this, `firebase deploy --only dataconnect` completes normally.

### Suggested fix in `firebase-tools`

Even after #576 is fixed upstream, `cleanUpFn` should be defensive — a synchronous throw from a teardown step should not be reported as the overall failure of the operation. Two options, either or both:

1. **Pin the connector** until 576 is fixed:
   ```diff
   - "@google-cloud/cloud-sql-connector": "^1.3.3",
   + "@google-cloud/cloud-sql-connector": ">=1.3.3 <1.7.0",
   ```
2. **Wrap cleanup in try/catch** in `lib/gcp/cloudsql/connect.js`:
   ```js
   const cleanUpFn = async () => {
     try { conn.release(); } catch {}
     try { await pool.end(); } catch {}
     try { connector.close(); } catch {}
   };
   ```
   This is correct regardless of the upstream fix — teardown errors should never mask a successful operation's results.

fredzqm:
### Description
This commit resolves an issue where deploying a Data Connect service backed by Cloud SQL Postgres fails with an unexpected unhandled exception (`CloudSQLConnectorError: The connector was closed`).

In `@google-cloud/cloud-sql-connector` version `>= 1.7.0`, the `connector.close()` method destroys underlying sockets by asynchronously emitting an `error` event. Previously, `firebase-tools` closed the `pg.Pool` first, which caused the database client to detach all error listeners from its sockets. Consequently, the later `connector.close()` call caused Node.js to crash due to an unhandled exception on those sockets.

To prevent this:
1. Re-ordered cleanup sequences so that `connector.close()` is executed prior to `pool.end()`.
2. Attached an explicit `'error'` handler to `pg.Pool` to ensure any connection teardown errors on idle clients are safely logged and absorbed.
3. Wrapped cleanup functions inside `try/catch` blocks for added defensiveness.

Fixes #10549

### Scenarios Tested
- Successfully ran all dataconnect and cloudsql unit tests.
- Verified teardown sequence completes gracefully without crash using a local reproduction script.
- Verified fix handles both `connect.ts` deployments and `dataconnect-sql-shell.ts` shell exits.

### Sample Commands
`firebase deploy --only dataconnect --project <project-id>`


joehan (approved):
Even better than pinning - thanks!

PR #10569, opened 2026-05-27
fix: update accountExporter unit test to match current CSV output

joehan:
Fixing unit test that was broken during merge.


dawogfather:
apologies! I missed that test on rerun

PR #10570, opened 2026-05-27
fix mistake in class initialization for ailogic

Berlioz:


ajperel (approved):
LG. I wish I understood better _why_ this fixes something, but I dont' think it will hurt and there's more important things to focus on.

Issue #8469, opened 2025-04-19
Issue with deploying onRequest cloud function as exports['dummystore-bot'] = onRequest(botApp)


aqibnawazdev:
I am trying  to deploy onRequest cloud function with name include hyphen (-) as "dummystore-bot", Instead of dummystoreBot with CLI, After "folder uploaded successful" it give following error  

![Image](https://github.com/user-attachments/assets/e513171a-4286-4029-b717-076954e85334)

Is there any way to do it using JavaScript..?

I also tried gcloud console but still it gives error

PR #10572, opened 2026-05-28
fix: trust discovered entry points for hyphenated functions

IzaakGough:
Related to `firebase/firebase-tools#8469`

Companion SDK PR: 

This PR updates `firebase-tools` to trust the `entryPoint` emitted by `firebase-functions` discovery instead of re-deriving it from function ids.

## Problem

The companion `firebase-functions` change fixes discovery for quoted export names containing hyphens, for example:

```js
const func = onRequest((req, res) => {
  res.send("ok");
});

export { func as "dummystore-bot" };
```

After that SDK fix, the discovered `entryPoint` is correctly preserved as `dummystore-bot`.

However, `firebase-tools` still had two places that assumed hyphens should be converted into dot notation:

- gen2 deploy code set `FUNCTION_TARGET` by rewriting `entryPoint` with `- -> .`
- `functions:shell` reconstructed shell bindings from `trigger.name` instead of using the discovered `trigger.entryPoint`

That behavior is still correct for grouped exports like `grouped.fn`, but it is incorrect for flat quoted export names that literally contain hyphens.

## Change

This PR makes `firebase-tools` use the SDK-provided `entryPoint` as the source of truth.

### Gen2 deploy

`cloudfunctionsv2` now sets:

- `FUNCTION_TARGET = buildConfig.entryPoint`

without rewriting hyphens to dots.

### Functions shell

`functions:shell` now initializes REPL bindings from `trigger.entryPoint` rather than deriving a path from `trigger.name`.

This preserves both behaviors correctly:

- grouped exports continue to bind as nested paths like `grouped.fn`
- quoted export names with hyphens remain flat bindings like `dummystore-bot`

## Tests

This PR adds or updates coverage for:

- gen2 deploy requests preserving a hyphenated `FUNCTION_TARGET`
- shell context initialization using `trigger.entryPoint` for both:
  - flat hyphenated names
  - grouped dotted entry points

## Relationship To `firebase-functions`

This PR is the CLI-side follow-up to the primary SDK fix in `firebase-functions`.

On its own, this change improves correctness whenever `entryPoint` is already accurate. Together with the companion `firebase-functions` PR, it fixes the full quoted-hyphen export flow end to end.

Companion PR: 

## Validation

Validated with:

- targeted `cloudfunctionsv2` and shell-context regression tests
- the companion local `firebase-functions` fix
- a packaged end-to-end flow using:
  - `npm pack` on local `firebase-functions`
  - installing that tarball into local `firebase-tools`
  - `npm pack` on local `firebase-tools`
  - deploying from a throwaway test project using the packed CLI and packed SDK

That deploy successfully recognized and updated a function named `dummystore-bot`.

### DIFF
--- a/src/emulator/auth/operations.ts
+++ b/src/emulator/auth/operations.ts
@@ -1230,9 +1230,6 @@ export function setAccountInfoImpl(
     }
 
     if (reqBody.deleteProvider?.includes(PROVIDER_PASSWORD)) {
-      updates.email = undefined;
-      updates.emailVerified = undefined;
-      updates.emailLinkSignin = undefined;
       updates.passwordHash = undefined;
       updates.salt = undefined;
     }
--- a/src/emulator/auth/setAccountInfo.spec.ts
+++ b/src/emulator/auth/setAccountInfo.spec.ts
@@ -1097,6 +1097,26 @@ describeAuthEmulator("accounts:update", ({ authApi, getClock }) => {
     "google.com",
   );
 
+  it("should retain email after deleting password provider", async () => {
+    const email = "alice@example.com";
+    const { idToken } = await registerUser(authApi(), { email, password: "notasecret" });
+
+    await authApi()
+      .post("/identitytoolkit.googleapis.com/v1/accounts:update")
+      .query({ key: "fake-api-key" })
+      .send({ idToken, deleteProvider: ["password"] })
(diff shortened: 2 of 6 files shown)


IzaakGough:
Closing in favor of https://github.com/firebase/firebase-tools/pull/10573

PR #10584, opened 2026-06-02
chore: remove temporary Dart console warning

demolaf:
The Firebase Console now supports Dart functions, so the temporary warning directing users to the Cloud Console is no longer needed.


Lyokone (approved):

shettyvarun268 (approved):
LGTM.
You might want to add an entry in the change log but overall looks okay!

PR #10586, opened 2026-06-02
refactor: move default region resolution to service interface

wandamora:
### Description

This PR refactors the trigger default region resolution logic during Cloud Functions deployment to make it less brittle, easier to maintain, and more scalable.

Previously, prepare.ts contained complex service-specific logic and switch/if blocks to inspect different event types (Firestore, Cloud Storage, Realtime Database, etc.) and look up their corresponding default regions. This was highly brittle: if a new service or event type is added, developers would have to manually update prepare.ts, which could be easy to miss.

### DIFF
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -12,7 +12,7 @@ import * as runtimes from "./runtimes";
 import * as supported from "./runtimes/supported";
 import * as validate from "./validate";
 import * as ensure from "./ensure";
-import { serviceForEndpoint } from "./services";
+import { serviceForEndpoint, FALLBACK_DEPLOYMENT_REGION } from "./services";
 import {
   functionsOrigin,
   artifactRegistryDomain,
@@ -59,7 +59,6 @@ import { DeployOptions } from "..";
 import * as prompt from "../../prompt";
 
 export const EVENTARC_SOURCE_ENV = "EVENTARC_CLOUD_EVENT_SOURCE";
-export const DEFAULT_FUNCTION_REGION = "us-central1";
 
 /**
  * Prepare functions codebases for deploy.
@@ -356,7 +355,7 @@ export async function resolveDefaultRegionsForBuild(
 ): Promise<void> {
   for (const [id, endpoint] of Object.entries(buildObj.endpoints)) {
     if (!endpoint.region?.length || endpoint.region.includes(build.REGION_TBD)) {
-      let resolvedRegion = DEFAULT_FUNCTION_REGION;
+      let resolvedRegion = FALLBACK_DEPLOYMENT_REGION;
 
       // Match existing endpoints.
       let matching: backend.Endpoint | undefined;
@@ -379,7 +378,7 @@ export async function resolveDefaultRegionsForBuild(
           resolvedRegion = await resolveRegionForTrigger(endpoint);
         } catch (err: any) {
           logger.debug(
-            `Failed to resolve region for endpoint ${id}. Defaulting to ${DEFAULT_FUNCTION_REGION}.`,
+            `Failed to resolve region for endpoint ${id}. Defaulting to ${FALLBACK_DEPLOYMENT_REGION}.`,
             getErrStack(err),
           );
         }
@@ -392,10 +391,7 @@ export async function resolveDefaultRegionsForBuild(
 
 async function resolveRegionForTrigger(endpoint: build.Endpoint): Promise<string> {
   const service = serviceForEndpoint(endpoint);
-  if (service.getDefaultRegion) {
-    return await service.getDefaultRegion(endpoint);
-  }
-  return DEFAULT_FUNCTION_REGION;
+  return await service.getDefaultRegion(endpoint);
 }
 
 /**
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -12,12 +12,7 @@ import * as runtimes from "./runtimes";
 import * as supported from "./runtimes/supported";
 import * as validate from "./validate";
 import * as ensure from "./ensure";
-import * as events from "../../functions/events/v1";
-import { getDatabase } from "./services/firestore";
-import { getBucket } from "./services/storage";
-import { getDatabaseInstanceDetails } from "./services/database";
-import { isGlobalAILogicTrigger } from "./services/ailogic";
-import { parseServiceName, parseConnectorName } from "../../dataconnect/names";
+import { serviceForEndpoint } from "./services";
 import {
   functionsOrigin,
   artifactRegistryDomain,
@@ -381,14 +376,7 @@ export async function resolveDefaultRegionsForBuild(
       } else {
         // Match triggers.
         try {
-          if (build.isBlockingTriggered(endpoint)) {
-            resolvedRegion = resolveRegionForBlockingTrigger(endpoint.blockingTrigger);
-          } else if (build.isEventTriggered(endpoint)) {
-            resolvedRegion = await resolveRegionForEventTrigger(
-              endpoint.project,
-              endpoint.eventTrigger,
-            );
-          }
+          resolvedRegion = await resolveRegionForTrigger(endpoint);
         } catch (err: any) {
           logger.debug(
             `Failed to resolve region for endpoint ${id}. Defaulting to ${DEFAULT_FUNCTION_REGION}.`,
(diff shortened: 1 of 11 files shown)


src/deploy/functions/prepare.ts
```diff
@@ -402,114 +390,11 @@ export async function resolveDefaultRegionsForBuild(
   }
 }
 
-function resolveRegionForBlockingTrigger(blockingTrigger: build.BlockingTrigger): string {
-  const eventType = blockingTrigger.eventType;
-  if ((events.AUTH_BLOCKING_EVENTS as readonly string[]).includes(eventType)) {
-    return "us-east1";
+async function resolveRegionForTrigger(endpoint: build.Endpoint): Promise<string> {
+  const service = serviceForEndpoint(endpoint);
+  if (service.getDefaultRegion) {
```
ajperel: I will do a full review soon but taking a quick glance now.

Do you think we should make this a required method for the Service interface? Then we'd never forget it for new services. But I don't know if we want to hard code the us-central default in a million places. Something to mull over.

src/deploy/functions/services/index.ts
```diff
@@ -63,6 +79,7 @@ const pubSubService: Service = {
   validateTrigger: noop,
   registerTrigger: noop,
   unregisterTrigger: noop,
+  getDefaultRegion: () => Promise.resolve("us-east1"),
```
ajperel: Should we use a constant for this region?

src/deploy/functions/prepare.ts
wandamora: Yeah, you're right. I set the noop service object default to be `us-central1` since that is the general fallback for functions.

src/deploy/functions/services/index.ts
wandamora: Done, added `DEFAULT_GLOBAL_TRIGGER_REGION`.

src/deploy/functions/prepare.ts
wandamora: I added `FALLBACK_DEPLOYMENT_REGION` to services/index.ts to reference `us-central1` as the primary fallback.

src/gcp/location.ts
```diff
@@ -40,6 +40,20 @@ export const DUAL_REGION_MAPPING: Record<string, string> = {
   "us-east1": "nam4",
 };
 
+/** Firestore multi-region mapping (https://cloud.google.com/firestore/docs/locations#multi-region) */
+export const FIRESTORE_MULTI_REGION_MAPPING: Record<string, string> = {
```
ajperel: Hmm. These two new maps feel opposite of the ones above in that they go from multi-region -> specific region. The ones above go from region -> multi-region but they are named identically. 

I feel like that's weird and we should try to do better. I don't really want a large refactor but something like:
* REGION_TO_MULTI_REGION_MAPPING
* MULTI_REGION_TO_REGION|_MAPPING 

could work, and we could just do it for these new methods for now?

would be one way to resolve it. Or we consistently do region -> multi-region and find a way to make it work (but I don't love you at runtime always inverting the map)

src/gcp/location.ts
ajperel, line 64: Again maybe not for this PR, but this is now maybe better regionInCloudStorageLocation since we don't include the multi-regions for other products.

src/deploy/functions/services/index.ts
ajperel, line 183: I notice a bunch of these are in events files can we re-use those constants: https://github.com/firebase/firebase-tools/blob/5bcd59331130ccbf8e6e59173d3907e07d3e713f/src/functions/events/v1.ts#L4

src/deploy/functions/services/index.ts
```diff
@@ -178,19 +209,13 @@ const EVENT_SERVICE_MAPPING: Record<events.Event, Service> = {
   "google.firebase.ailogic.v1.afterGenerate": aiLogicService,
```
ajperel, line 209: Does this mapping need providers/firebase.auth/eventTypes/ ?

src/deploy/functions/services/index.ts
```diff
@@ -1,14 +1,28 @@
 import * as backend from "../backend";
 import * as iam from "../../../gcp/iam";
 import * as events from "../../../functions/events";
+import * as build from "../build";
 import { AuthBlockingService } from "./auth";
```
ajperel, line 5: AI says we have a runtime circular dependency because auth also imports index.js to get `noop` (the others aren't runtime dependencies). And we could fix that by not importing `noop` and duplicating that very simple promise. but .. I guess that predates this PR and has been working so far so maybe OK?

src/deploy/functions/services/dataconnect.ts
```diff
@@ -38,3 +40,27 @@ export function getDataConnectP4SA(projectNumber: string): string {
   }
   return `service-${projectNumber}@${PROD_DATACONNECT_SA_DOMAIN}`;
 }
+
+/**
+ * Get the default region for a DataConnect event trigger.
+ */
+export async function getDefaultRegion(endpoint: build.Endpoint): Promise<string> {
```
ajperel, line 47: Should we add unit tests for `getDefaultRegion` or do you think the slightly higher level tests from our first implementation are sufficient?

src/deploy/functions/services/index.ts
ajperel, line 181: On the exhaustion front AI claims

"Because it is typed as Record<events.Event, Service>, TypeScript's compiler strictly enforces that every member of the events.Event union type must be mapped. If a developer adds a new event type to events.Event but forgets to add it here, compilation will fail. This is an excellent, compiler-enforced design pattern."

So maybe you have that, but I'm not 100% sure if it's hallucinating and didn't double check

src/deploy/functions/prepare.ts
ajperel, line 151: Not in this PR but AI says we have a problem here when CEL expressions are used that we might want to fix in another PR:

During Cloud Functions deployment, the process is executed in sequential phases:

Load Codebases (loadCodebases): The Firebase CLI runs the developer's source code to discover the declared functions and their trigger configurations. This produces the wantBuild object. At this stage, any configuration parameterized using the Firebase Functions SDK (e.g. defineString('BUCKET_NAME')) is exported as an unresolved CEL expression string (e.g. {{ params.BUCKET_NAME }}).
Resolve Default Regions (resolveDefaultRegionsForBuild): The CLI determines regions for endpoints that don't have one explicitly set. This happens before parameters are resolved.
Resolve Backend (build.resolveBackend): The CLI prompts the user or reads .env files to resolve the values of all parameters, replacing the CEL expressions with actual values.
The Problem
Because default region resolution (Phase 2) runs before parameter resolution (Phase 3), any service-specific lookup uses the raw CEL expression:

For a storage trigger parameterized with a bucket, the event filter bucket name will be "{{ params.BUCKET_NAME }}".
The Storage service's getDefaultRegion method will receive this string and attempt to query the Cloud Storage API: storage.getBucket("{{ params.BUCKET_NAME }}").
Since "{{ params.BUCKET_NAME }}" is not a valid bucket name, the GCS API call will always fail.
Although the CLI catches this failure and gracefully falls back to the default region (us-central1), it performs a wasteful network request that is guaranteed to fail on every single deploy.

The Solution
We should check if the resource identifier (bucket name, database ID, database instance name, etc.) is a CEL expression before attempting the API call.

By importing isCelExpression from src/deploy/functions/cel.ts and checking the identifier:

typescript
if (isCelExpression(bucketName)) {
  throw new FirebaseError("Cannot resolve bucket region for a parameterized bucket name");
}
We immediately stop the lookup and trigger the fallback path, avoiding unnecessary API calls and keeping log output cleaner in --debug mode.

ajperel (commented):
A bunch of more minor feedback. Should this get a CHANGELOG entry?

AI Test writing recommendations:
* AI Logic Triggers: Test that resolveDefaultRegionsForBuild resolves us-east1 for global AI Logic triggers but resolves us-central1 for AI Logic triggers with regionalWebhook: true.
* Graceful Fallback on API failure: A test asserting that when getBucket or getDatabase throws an API error, resolveDefaultRegionsForBuild catches it and correctly falls back to us-central1 instead of crashing the deployment.

src/gcp/location.ts
wandamora: Are you suggesting we change the pre-existing mappings to REGION_TO_MULTI_REGION_MAPPING in a later PR, and for the ones I'm adding to MULTI_REGION_TO_REGION_MAPPING now?

What do you think about:
FIRESTORE_MULTI_REGION_TO_REGION_MAPPING and STORAGE_MULTI_REGION_TO_REGION_MAPPING for the new mappings, and update the documentation explaining these are the trigger region to deployment region mappings for functions?

src/deploy/functions/services/index.ts
wandamora: Some but not all. This is room for improvement for a separate PR, though.

src/deploy/functions/services/index.ts
wandamora: Yeah, I'm hesitant to expand the scope of this PR and change code unrelated to the refactor or seems to have had no issues up until this point. The AI is probably making fair claims but I fear wants to overcorrect the code beyond the scope of the PR.

src/deploy/functions/services/index.ts
wandamora: ah sorry, I spoke too soon. these are all covered by constants. Again, I prefer not refactor or update the code outside of the scope of this PR.

src/deploy/functions/services/index.ts
wandamora: In response to this gotcha, https://github.com/firebase/firebase-tools/pull/10586/changes#r3352639212, yes because they are the v1 events listed in src/functions/events/v1.ts.

wandamora:
> A bunch of more minor feedback. Should this get a CHANGELOG entry?

I didn't think so since this doesn't impact users of the CLI, it's only a refactor.

> AI Test writing recommendations:
> 
> * AI Logic Triggers: Test that resolveDefaultRegionsForBuild resolves us-east1 for global AI Logic triggers but resolves us-central1 for AI Logic triggers with regionalWebhook: true.
> * Graceful Fallback on API failure: A test asserting that when getBucket or getDatabase throws an API error, resolveDefaultRegionsForBuild catches it and correctly falls back to us-central1 instead of crashing the deployment.

Done.

src/deploy/functions/services/index.ts
wandamora: Took a look through the list of events in https://github.com/firebase/firebase-tools/blob/5bcd59331130ccbf8e6e59173d3907e07d3e713f/src/functions/events/index.ts#L6 and they are all present.

src/deploy/functions/services/dataconnect.ts
wandamora: I think the high-level tests are sufficient! They're covering all the services we use.

src/gcp/location.ts
ajperel: I figured we could say changing the pre-existing ones is out of scope (even if it'd be clearer) and just clearly name the new ones clearly. This is probably a minor point but it did feel weird. Those names seem fine for now... to my point on decisions earlier this is super easy to change so long but descriptive seems good enough and if we have better ideas it's a quick fix.

src/deploy/functions/services/index.ts
ajperel: I'm more torn on this since I think we introduced it in our sequence of changes, but OK. Maybe at least add to the TODO to de-duplicate as well.

src/deploy/functions/services/index.ts
ajperel: talked through on chat...

These should map to the no-op service so not having them effectively does that. Maybe it'd be a _little_ clearer to explicitly put that but this works.

I was confused by the difference between auth blocking events and just auth events and the authService is really authBlockingService.

ajperel (approved):
I think at this point there's lots we've identified that _could_ be better but mostly related to things that are a bit out of scope. So other than maybe improving the mapping names we can save them all for the future. Jetski authored folow ups welcome though :)

PR #10588, opened 2026-06-02
fix(hosting): route Cloud Run rewrites to local functions emulator

demolaf:
When using Firebase Hosting rewrites with `"run"` format (required for Dart/Cloud Run functions),
the hosting emulator always hit the live Cloud Run API to resolve the service URL — even when the
functions emulator was running locally. This caused an error like:

```
Error looking up URL for Cloud Run service: 404, Resource 'app' does not exist
```

## Fix

Add emulator detection to `cloudRunProxy.ts`, mirroring the existing logic in `functionsProxy.ts`.
When `targets` includes `"functions"` and the functions emulator is running, route directly to the
local emulator URL instead of calling the Cloud Run API.

Related https://github.com/firebase/firebase-functions-dart/pull/200


Lyokone (approved):

shettyvarun268 (approved):
LGTM.
NIT: I think it would be a good idea to update the CHANGELOG.md for this PR as well.

PR #10589, opened 2026-06-02
Adds support for optional params

Berlioz:
Params with optional = true are only prompted for on a "fresh" functions deploy, where the corresponding .env file is empty.

Otherwise, they are automatically filled in with the param's default value, or a zero value based on the param's type. This can bypass validation for param types that support validation regexes.

Optional params resolved this way are not written to .env. (Optional params resolved by prompting on a fresh deploy are).

While we're here, remove the `immutable` param configuration, which is a holdover from cargo culting Extensions params and doesn't really have any meaning in the functions world. Immutability wasn't implemented and it wasn't present in the wire format for a param from the SDK so that config flag was just totally marooned from the outside universe.

### DIFF
--- a/src/deploy/functions/params.ts
+++ b/src/deploy/functions/params.ts
@@ -120,8 +120,10 @@ type ParamBase<T extends string | number | boolean | string[]> = {
   // Default value. If not provided, a param must be supplied.
   default?: T | build.Expression<T>;
 
-  // default: false
-  immutable?: boolean;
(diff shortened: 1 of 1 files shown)


shettyvarun268:
The code looks clean. This will help us with the pain point with extensions where we won't have to prompt the user for every param in the CLI. Should we add some test cases for this? Considering any changes in the near future that might cause issues if not consistent.

PR #10596, opened 2026-06-03
emulator: validate Host header to mitigate DNS-rebinding

adilburaksen:
## Description

The local emulator HTTP servers all use `cors({ origin: true })` and perform no `Host`-header validation. A web page open in a developer's browser can therefore reach the (unauthenticated) emulator APIs via **DNS-rebinding** — rebind an attacker-controlled domain to `127.0.0.1`, after which requests are same-origin so CORS no longer applies — and drive the emulator endpoints (read/modify/delete local emulator data, toggle Cloud Functions triggers, etc.).

## Fix

Add a small `hostValidationMiddleware` (`src/emulator/hostValidation.ts`) that rejects requests whose `Host` header is not a loopback/configured host, wired as the first middleware on every emulator's express app — Hub/UI via `ExpressBasedEmulator`, plus Auth, Storage, Eventarc, Tasks, and Functions.

Enforcement happens **only when the emulator is bound exclusively to loopback addresses** (the rebinding-vulnerable default). If the operator bound to a non-loopback address (`0.0.0.0`, `::`, or a specific host) they have opted into remote access — e.g. the tunnel support from #4227 — so the middleware becomes a no-op and those setups keep working.

The legitimate callers (the `HubClient` node client, the CLI, and the Emulator UI) send loopback `Host` headers and requests without a `Host` header are allowed, so they are unaffected.

## Testing

- Typecheck (`tsc --noEmit`) passes; files are prettier-clean.
- Middleware behavior verified: under a loopback bind it allows `Host: localhost` / `127.0.0.1` / `::1` / `[::1]` / the configured host (and requests with no `Host`), and returns `403` for an untrusted host (e.g. a rebound attacker domain); under a non-loopback bind it is a pass-through.

### DIFF
--- a/src/emulator/ExpressBasedEmulator.ts
+++ b/src/emulator/ExpressBasedEmulator.ts
@@ -6,6 +6,7 @@ import * as utils from "../utils";
 import { Emulators, EmulatorInstance, EmulatorInfo, ListenSpec } from "./types";
 import { createServer } from "node:http";
 import { IPV4_UNSPECIFIED, IPV6_UNSPECIFIED } from "./dns";
+import { hostValidationMiddleware } from "./hostValidation";
 import { ListenOptions } from "node:net";
 
 export interface ExpressBasedEmulatorOptions {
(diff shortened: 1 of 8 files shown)


adilburaksen:
@joehan this is the companion to the Hub cross-origin PR — it validates the Host header on the emulator to mitigate DNS-rebinding against localhost. Happy to fold the two together if you'd rather review them as one.

PR #10608, opened 2026-06-09
fix-windows-python-init

shettyvarun268:
This pull request fixes a bug on Windows where initializing Firebase Functions with Python fails during the dependency installation step, throwing a misleading error that says it cannot find the file activate.bat (ENOENT).

The Problem
During investigation, we found that this error is caused by two different issues. First, if a user does not have Python installed or configured in their system PATH, the step to create the virtual environment fails silently because the Firebase CLI does not check its exit code. The CLI then moves on to execute the activation script, which fails because the virtual environment folder was never actually created.

Second, even if a user has Python installed, the installation command tries to run pip3 directly to upgrade pip. On Windows, the operating system locks the pip3 executable while it is running, which causes the upgrade to fail. In both of these failure cases, a quirk in the command-spawning library on Windows translates the failure into a misleading "file not found" error for the activate.bat file, hiding the real root cause.

The Solution
To fix these issues, we updated the virtual environment creation step to output python errors directly to the terminal and to verify the command's exit code. If the virtual environment fails to build (due to a missing Python installation), the CLI now stops immediately and shows a clear error message.

We also changed the dependency commands to use python -m pip instead of calling pip3 directly. This approach is the recommended standard because it runs the command through the Python interpreter, avoiding the Windows file-locking issue. Finally, we added unit tests to ensure that the Python codebase is initialized with the correct configuration and that failures are caught and reported properly.

### DIFF
--- a/src/init/features/functions/python.ts
+++ b/src/init/features/functions/python.ts
@@ -1,4 +1,5 @@
-import * as spawn from "cross-spawn";
+import { ChildProcess } from "child_process";
+import { wrapSpawn } from "../../spawn";
 
 import { Config } from "../../../config";
 import { getPythonBinary } from "../../../deploy/functions/runtimes/python";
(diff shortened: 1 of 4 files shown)


src/init/features/functions/python.ts
```diff
@@ -1,20 +1,80 @@
...
+    p.on("exit", resolve);
+    p.on("error", () => resolve(-1));
+  });
+}
+
+/**
+ * Helper to spawn a process and create a python virtual environment.
+ */
+async function createVirtualEnv(pythonBinary: string, cwd: string): Promise<void> {
+  const venvProcess = spawn(pythonBinary, ["-m", "venv", "venv"], {
```
ajperel: I think you can be consistent with other code and simplify this abit by using our wrapspawn utility: https://github.com/firebase/firebase-tools/blob/938d93cc91d063a44bb900d1c28d294ab45f34a0/src/init/spawn.ts#L11

ajperel (commented):
I'm very unfamiliar with this code. It seems fine but might not be a bad idea to get a second review from someone more familiar. Left one quick comment. 

Also... a few warnings in the tests that would be good to resolve.

src/init/features/functions/python.ts
shettyvarun268: Thanks for the review!
I'll have someone else take a look as well to get approval.

joehan (approved):

PR #10609, opened 2026-06-09
feat(auth-emulator): Add mock Passkey (WebAuthn) support

macastelaz:
### Description

Add support to the auth emulator for the passkey API surface. Passkey support is currently in "private preview" on a feature branch of the firebase web auth SDK (https://github.com/firebase/firebase-js-sdk/tree/passkey) and we are working towards a beta release which will involve updating the [auth quickstart](https://github.com/firebase/quickstart-js/tree/master/auth) that in turn relies on the auth emulator.

### Scenarios Tested

Using a locally modified quickstart version with passkey support pointing at a locally running auth emulator with these changes (https://screenshot.googleplex.com/3EbEZapL5Ax8o8h.png) I manually verified: 

- Sign in with passkey
- Enrolling a passkey
- Unerolling a passkey
- Getting passkeys enrolled for a user


scripts/gen-auth-api-spec.ts
```diff
@@ -454,6 +458,12 @@ function addEmulatorOperations(openapi3: any): void {
         },
         type: "object",
       },
+      emailPrivacyConfig: {
```
joehan, line 461: This seems new/possibly unrelated to the Passkey support. Just want to double check that it ought to be here.

src/emulator/auth/operations.ts
```diff
@@ -1239,6 +1246,12 @@ export function setAccountInfoImpl(
     if (reqBody.deleteProvider?.includes(PROVIDER_PHONE)) {
       updates.phoneNumber = undefined;
     }
+    if (reqBody.deletePasskey && reqBody.deletePasskey.length > 0) {
```
joehan, line 1249:
```suggestion
    if (reqBody.deletePasskey) {
```

Checking length is unnecessary here, as "" is false-y in JS

src/emulator/auth/operations.ts
```diff
@@ -2276,6 +2289,116 @@ async function mfaSignInFinalize(
...
+      userVerification: "required",
+    },
+  };
+}
+
+function passkeySignInFinalize(
+  state: ProjectState,
+  reqBody: Schemas["GoogleCloudIdentitytoolkitV2FinalizePasskeySignInRequest"],
+):
+  | Promise<Schemas["GoogleCloudIdentitytoolkitV2FinalizePasskeySignInResponse"]>
```
joehan, line 2370: This return type seems wrong to me - I do not see where this would return a Promise.

src/emulator/auth/operations.ts
```diff
@@ -2294,16 +2417,19 @@ function updateConfig(
...
-    if (Object.prototype.hasOwnProperty.call(reqBody.blockingFunctions!.triggers, event)) {
-      assert(
-        Object.values(BlockingFunctionEvents).includes(event as BlockingFunctionEvents),
-        "INVALID_BLOCKING_FUNCTION : ((Event type is invalid.))",
-      );
-      assert(
-        parseAbsoluteUri(reqBody.blockingFunctions!.triggers[event].functionUri!),
-        "INVALID_BLOCKING_FUNCTION : ((Expected an absolute URI with valid scheme and host.))",
-      );
+  if (reqBody.blockingFunctions?.triggers) {
```
joehan, line 2420: Good catch - this code seems like it would have caused runtime exceptions when blockingFunctions is undefined. However, we can avoid adding this if statement by just using a fallback value in the for loop:

`for (const event in (reqBody.blockingFunctions?.triggers ?? [])) {`

src/emulator/auth/operations.ts
```diff
@@ -3367,6 +3493,9 @@ function generateBlockingFunctionJwt(
   return jwtStr;
 }
 
+/**
```
joehan, line 3496: Remove (or better yet, add a real comment describing the function)

src/emulator/auth/operations.ts
```diff
@@ -3367,6 +3493,9 @@ function generateBlockingFunctionJwt(
...
+ *
+ */
 export function parseBlockingFunctionJwt(jwt: string): BlockingFunctionsJwtPayload {
   const decoded = decodeJwt(jwt, { json: true }) as any as BlockingFunctionsJwtPayload;
```
joehan, line 3500: Drive by (optional fix since you didn't write this): `blah as any as MyType` is a code smell. We ought to do something safer like a type guard: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

src/emulator/auth/state.ts
```diff
@@ -932,10 +933,16 @@ interface TemporaryProofRecord {
   // a bit easier. Therefore, there's no need to record createdAt timestamps.
 }
 
+/**
+ *
```
joehan, line 937: Remove (or write a real comment)

src/emulator/auth/state.ts
```diff
@@ -932,10 +933,16 @@ interface TemporaryProofRecord {
...
+ */
 export function encodeRefreshToken(refreshTokenRecord: RefreshTokenRecord): string {
   return Buffer.from(JSON.stringify(refreshTokenRecord), "utf8").toString("base64");
 }
 
+/**
+ *
```
joehan, line 944: Remove (or write a real comment)

joehan (changes requested):

scripts/gen-auth-api-spec.ts
macastelaz: Odd - I'm honestly not sure how this snuck in. Thanks for catching it and flagging it. It has been removed in https://github.com/firebase/firebase-tools/pull/10636

src/emulator/auth/operations.ts
macastelaz: Good catch! Updated

src/emulator/auth/operations.ts
macastelaz: Agreed - shouldn't need to return a Promise

src/emulator/auth/state.ts
macastelaz: Removed

src/emulator/auth/state.ts
macastelaz: Removed

src/emulator/auth/operations.ts
macastelaz: Good call - done

src/emulator/auth/operations.ts
macastelaz: Done

src/emulator/auth/operations.ts
macastelaz: Good call - done!

PR #10614, opened 2026-06-09
[SQL CONNECT] Auto-approve schema migrations on first service deploy

mtr002:
First Deploy:

https://paste.googleplex.com/6148653852000256

Second Deploy:

https://paste.googleplex.com/4971040332709888

### DIFF
--- a/src/dataconnect/schemaMigration.ts
+++ b/src/dataconnect/schemaMigration.ts
@@ -165,7 +165,7 @@ export async function migrateSchema(args: {
 
   const projectId = needProjectId(options);
   const { serviceName, instanceId, instanceName, databaseId, schemaName } = getIdentifiers(schema);
-  await ensureServiceIsConnectedToCloudSql(
+  const isFirstDeploy = await ensureServiceIsConnectedToCloudSql(
     serviceName,
     instanceName,
     databaseId,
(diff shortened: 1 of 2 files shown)


src/dataconnect/schemaMigration.ts
```diff
@@ -539,19 +541,28 @@ async function handleIncompatibleSchemaError(args: {
...
   if (!err) {
     return "none";
   }
   const defaultChoice = validationMode === "STRICT_AFTER_COMPATIBLE" ? "none" : "all";
   displaySchemaChanges(err, validationMode);
+  // On first deploy, auto-approve without prompting since there's nothing to migrate from.
+  if (isFirstDeploy && !validateOnly) {
+    logLabeledBullet(
+      "dataconnect",
+      "First schema deployment detected — automatically applying SQL changes.",
```
fredzqm, line 562: nit: Feels okay to skip this info log. It is already clear because user won't be getting a confirmation but seeing SQL executed.

fredzqm (approved):

fredzqm:
Thank you! Nice polish improvement!

fredzqm (changes requested):
Wait. This is actually risky than it looks like.

If we point a new FDC service against an existing Cloud SQL instance, then the "skip confirmation" heuristic may actually drops some tables dangerously.

The safe signal to look for is whether the current Cloud SQL schema is empty, but we don't know that for sure in the context of CLI. So yeah, maybe defer it for now.

mtr002:
Wait you're right! No way to query Cloud SQL and check if the schema is empty. Closing

PR #10618, opened 2026-06-10
add validation for the app directory prompt

aalej:
### Description

When prompted for the app hosting app directory and you provide a non-existent path the CLI will error out:
```
=== App Hosting Setup
i  This command links your local project to Firebase App Hosting. You will be able to deploy your web app with `firebase deploy` after setup.
✔ Please select an option Link to an existing backend
✔ Which backend would you like to link? BACKEND_ID
i  === Deploy local source setup
✔ Specify your app's root directory relative to your firebase.json directory /non-existent
✔  Wrote configuration info to firebase.json
i  Writing default settings to apphosting.yaml...

Error: Failed to write apphosting.yaml file because app root directory /Users/PATH/apphosting-folder-setup/non-existent does not exist. Please try again with a valid directory.
```

Updated the behavior to validate the path during prompting:
```
=== App Hosting Setup
i  This command links your local project to Firebase App Hosting. You will be able to deploy your web app with `firebase deploy` after setup.
✔ Please select an option Link to an existing backend
✔ Which backend would you like to link? BACKEND_ID
i  === Deploy local source setup
? Specify your app's root directory relative to your firebase.json directory /non-existent
> Directory /Users/PATH/apphosting-folder-setup/non-existent does not exist. Please enter a valid directory.
```

### Scenarios Tested

```
$ firebase init apphosting 

     ######## #### ########  ######## ########     ###     ######  ########
     ##        ##  ##     ## ##       ##     ##  ##   ##  ##       ##
     ######    ##  ########  ######   ########  #########  ######  ######
     ##        ##  ##    ##  ##       ##     ## ##     ##       ## ##
     ##       #### ##     ## ######## ########  ##     ##  ######  ########

You're about to initialize a Firebase project in this directory:

  /Users/PATH/apphosting-folder-setup

Before we get started, keep in mind:

  * You are initializing within an existing Firebase project directory


=== Project Setup

First, let's associate this project directory with a Firebase project.
You can create multiple project aliases by running firebase use --add, 

i  Using project PROJECT_ID (PROJECT_NAME) .

=== App Hosting Setup
i  This command links your local project to Firebase App Hosting. You will be able to deploy your web app with `firebase deploy` after setup.
✔ Please select an option Link to an existing backend
✔ Which backend would you like to link? BACKEND_ID
i  === Deploy local source setup
? Specify your app's root directory relative to your firebase.json directory /non-existent
> Directory /Users/PATH/apphosting-folder-setup/non-existent does not exist. Please enter a valid directory.
```


src/init/features/apphosting.ts
```diff
@@ -113,6 +113,13 @@ export async function doSetup(setup: Setup, config: Config, options: Options): P
   backendConfig.rootDir = await input({
     default: "/",
     message: "Specify your app's root directory relative to your firebase.json directory",
+    validate: async (input: string) => {
+      const absPath = path.join(process.cwd(), input);
+      if (!existsSync(absPath)) {
+        return `Directory ${absPath} does not exist. Please enter a valid directory.`;
+      }
+      return true;
+    },
```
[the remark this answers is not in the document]
aalej: the `absPath` code was copied from over https://github.com/firebase/firebase-tools/blob/073a95321395a3307bc12fe45f23ffd27fd0c40e/src/init/features/apphosting.ts#L122

Although, `aynsc` should be removed since there are no promises or awaits inside the function. Good call out!

src/init/features/apphosting.ts
```diff
@@ -113,6 +113,13 @@ export async function doSetup(setup: Setup, config: Config, options: Options): P
   backendConfig.rootDir = await input({
     default: "/",
     message: "Specify your app's root directory relative to your firebase.json directory",
+    validate: (input: string) => {
+      const absPath = path.join(process.cwd(), input);
```
joehan: This logic assumes that process.cwd() == the firebase directory. However, if you run init from a child of the firebase directory, that will not be true. We've got some utils that resolve the firebase project dir - lets use those instead here.

joehan (changes requested):

src/init/features/apphosting.ts
aalej: Good catch! Running the init command on a sub-directory does break the prompt. We do set a config early on in https://github.com/firebase/firebase-tools/blob/2e3b10dadabb9f2b4ddc0e9f83ce4522d029952c/src/config.ts#L52

switching process.cwd() to config.projectDir

joehan (approved):

PR #10621, opened 2026-06-10
Functions can declare dependencies on additional APIs

inlined:
### Description
Updates `ensureAllRequiredAPIsEnabled` during function deployment to check manifest-required APIs against a built-in allowlist of standard Firebase APIs. For any additional APIs, it prompts the user to enable them if they are currently disabled.

### Scenarios Tested
- CLI unit tests verify prompting logic (presented when additional APIs are missing, not presented otherwise).
- Verified that `--force` confirms and `nonInteractive` without `--force` aborts.


src/deploy/functions/prepare.ts
```diff
@@ -680,14 +680,72 @@ export async function warnIfNewGenkitFunctionIsMissingSecrets(
...
+ * @param options Deploy execution options for prompting configuration.
+ * @param [options.force] Whether to force confirm interactive prompts.
+ * @param [options.nonInteractive] Whether the CLI is running in non-interactive mode.
+ */
 export async function ensureAllRequiredAPIsEnabled(
   projectNumber: string,
   wantBackend: backend.Backend,
+  options: { force?: boolean; nonInteractive?: boolean } = {},
 ): Promise<void> {
+  const STANDARD_APIS = [
```
wandamora: Where does this list come from?

wandamora: OK I see, thanks. On this note, would this be better as a global constant?

inlined: Yeah, I'll do that and add a comment.

wandamora (approved):

PR #10627, opened 2026-06-11
[VS Code] Fix compile error bug from generate operation button

fredzqm:
### Description

Fixes an "compilation error" bug on "Generate Operation" code lens when the user didn't login with `gcloud`.

> Schema compilation failed: failed to build HTTP client for calling Firebase Data Connect. Error: failed to set up an OAuth token source. Error google: could not find default credentials. See https://cloud.google.com/docs/authentication/external/set-up-adc for more information

This compilation check is brittle and resulted in false negative when user never login via `gcloud auth application-default login`, which is never expected of SQL Connect developers.
This compilation check is not necessary because the backend API can trivially return compile errors.

### Scenarios Tested
- Modified the extension code and ran `npm run build` inside `firebase-vscode` to verify successful compilation.
- Verified that all unit/e2e build and type-checking pass without warnings/errors.

### Sample Commands
N/A

### DIFF
--- a/firebase-vscode/CHANGELOG.md
+++ b/firebase-vscode/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## NEXT
 
+- [Fixed] Fix auth issue from Generate Operation button by skipping unnecessary schema compilation check.
 - [Changed] Replace generateSchema and generateQuery backends with the new fdc AgentService API.
 - [Added] Improved error handling and status updates for query generation
 - [Added] Local schema aware query generation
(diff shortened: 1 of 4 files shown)


hlshen (approved):
Sorry! Left this in a branch but never submitted. Thanks for getting to it.

hlshen:
Need to update the unit tests

fredzqm:
> Need to update the unit tests

@google-labs-jules

fredzqm:
Updated the unit tests to remove the verifySchemaCompiles check and successfully ran tests locally. Pushed changes.

PR #10635, opened 2026-06-11
feature/add auth event arc experiment

shettyvarun268:
This PR registers and gates the new experimental autheventarc feature flag within the Firebase CLI to control the deployment of v2 Auth Eventarc triggers (onUserCreated and onUserDeleted). It officially defines these event types in the CLI's v2 events registry and maps them to the default service map to ensure type safety. During deployment validation, the CLI checks for these trigger types and asserts that the autheventarc experiment is enabled, blocking deployments with a friendly error if it is not. This ensures the experimental triggers can be safely integrated into the SDK while preventing accidental deployments until the backend is fully launched. All changes have been verified with type checks and corresponding unit tests.

### DIFF
--- a/src/deploy/functions/services/index.ts
+++ b/src/deploy/functions/services/index.ts
@@ -207,6 +207,8 @@ const EVENT_SERVICE_MAPPING: Record<events.Event, Service> = {
   "google.firebase.dataconnect.connector.v1.mutationExecuted": dataconnectService,
   "google.firebase.ailogic.v1.beforeGenerate": aiLogicService,
   "google.firebase.ailogic.v1.afterGenerate": aiLogicService,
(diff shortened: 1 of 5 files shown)


wandamora (approved):

inlined (approved):
I wouldn't create an experiment for this (especially a public one). Experiments help us avoid clutter (e.g. the internal test utilities), guard against changes, or inform a customer that they're trying beta code.

Historically for adding new event types we've just not added documentation to the API and not added it to index.ts. If you want to add additional rigor to our process that's fine by me, but we shouldn't make the experiment public because it draws attention when we wren't asking for beta testers ATM.

And Gemini is right; setEnabled is there for testing so you can prefer it over mocks.

PR #10641, opened 2026-06-12
ci: fix storage emulator integration tests on Node 24

joehan:
### Description
The storage emulator test was broken on Node 24 due to a puppeteer incompatibility. We'll need to update our version of puppeteer eventually to address this, but for now, we can just stick this on node 22


shettyvarun268 (approved):
LGTM

PR #10646, opened 2026-06-14
fix: correct function name length in validation error message to 63

ree-rishun:
### Description

Fixes two inaccuracies in the function name validation error message in `functionIdsAreValid()` (`src/deploy/functions/validate.ts`). The message disagreed with both the validation regex `/^[a-zA-Z][a-zA-Z0-9_-]{0,62}$/` and the function's own JSDoc comment:

1. **Character limit (62 → 63):** The message said names must "not exceed **62** characters in length", but the regex allows **63** characters (1 leading letter + up to 62 additional characters), and the JSDoc comment also states "not exceed 63 characters in length". As a result, a 63-character function name deploys successfully, yet the message claimed the limit was 62 — misleading users into thinking valid names are invalid.

2. **Missing underscores:** The message listed only "letters, numbers, hyphens", but the regex (`[a-zA-Z0-9_-]`) and the JSDoc comment both allow **underscores**. Underscores were silently omitted from the user-facing message.

This change updates the message so it is consistent with the regex, the JSDoc comment, and the messages already used elsewhere (e.g. `applyPrefix()` in `src/deploy/functions/build.ts`, which says "maximum length of 63 characters").

```diff
- numbers, hyphens, and not exceed 62 characters in length
+ numbers, underscores, hyphens, and not exceed 63 characters in length
```

Note: this validation also rejects names that don't start with a letter or contain other invalid characters. This PR only corrects the wording of the error message and does not change any validation behavior.

### Scenarios Tested
- Ran `npm test` locally; compilation and the existing test suite pass.
- Deployed functions to a separate Firebase project and confirmed in the console that the deploys succeeded for both a 63-character name and a name containing underscores (e.g. `my_func`), verifying that 63 is the true limit (not 62) and that underscores are allowed.
- This is a message-only change; no validation logic was modified.

```console
$ TZ='America/Los_Angeles' npm test

... (truncated) ...

  4506 passing (38s)
  10 pending

... (coverage table truncated) ...
```

### Sample Commands
No command or flag changes. The corrected error surfaces during deploy:

```
Error: 1nnnnnnnnn function name(s) can only contain letters, numbers, underscores, hyphens, and not exceed 63 characters in length
```


IzaakGough (approved):
This looks good to me, nice work! The failing CI check is unrelated to this PR.

ree-rishun:
Hi @IzaakGough , thanks again for the review and approval.

It looks like the CI workflow now requires maintainer approval again after I updated the branch from `main`. I’ll avoid updating the branch further unless needed.

Could you or another maintainer approve/re-run the CI when you have a chance? Please let me know if there’s anything else I should change on my side.

PR #10647, opened 2026-06-15
refactor: replace uuid with native crypto.randomUUID

joehan:
### Description
Replaced the external third-party `uuid` dependency with Node.js native `crypto.randomUUID()` across all source files, as our Node.js >=20 engine constraint natively supports it. Removed `uuid` from primary `package.json` dependencies (while retaining security overrides).

### Scenarios Tested
- Fast unit test suite (`npm run mocha:fast`) successfully passed (4506 tests).
- Verified `npm-shrinkwrap.json` was regenerated clean and correctly synchronizes.

### Sample Commands
- `npm run mocha:fast`


joehan (commented):
Addressed inline review comments to cast data to Record<string, unknown> and synchronized npm 11.9 shrinkwrap.

bkendall (approved):

PR #10665, opened 2026-06-16
Show login:*, target:*, and ext:* subcommands in firebase --help

### LINKED ISSUE #10652: login:* and target:* subcommands missing from `firebase --help` (still functional)
### Description
The `login:*` and `target:*` subcommands are not listed in `firebase --help`, even though they are fully functional. Other namespaced commands (`auth:*`, `apps:*`, `database:*`, etc.) appear correctly. This hurts discoverability — e.g. a user who needs `login:use`/`login:add` to switch accounts after an auth failure cannot find them in help.

### Version
firebase-tools 15.20.0 (also reproduces on 15.x)

### Steps to reproduce
```
firebase --help | grep login:    # → no results
firebase --help | grep target:   # → no results
firebase login:list --help       # → works: "list authorized CLI accounts"
firebase login:list              # → works
```

### Expected
`login:list`, `login:add`, `login:use`, `login:ci`, `target:apply`, `target:clear`, `target:remove` appear in `firebase --help`.

### Actual
Only the bare `login` and `target` commands are listed; their colon-subcommands are omitted.

### Root cause
The help-time loader in `lib/bin/cli.js` (`loadAll`) registers a command via `value.load()` when it `isCommandModule`, but only recurses into children via the `else if (typeof value === "object")` branch. Commands that are *both* a leaf command and a namespace parent (`login`, `target`) are stored as functions with subcommands attached as properties — so the recursion never descends into them and their subcommands are never registered for help. They remain reachable at runtime because the resolver in `lib/index.js` walks the `:`-split path and lazy-loads on demand.

sarathfrancis90:
## Bug

Fixes #10652. The `login:*`, `target:*`, and `ext:*` subcommands don't appear in `firebase --help`, even though they work at runtime (e.g. `firebase login:list` runs fine and `firebase login:list --help` shows usage). Other namespaced commands like `auth:*` and `apps:*` show up correctly.

## Cause

The help-time loader in `src/bin/cli.ts` walked the command tree with an `if/else if`: a value was either loaded as a command *or* recursed into as a namespace, never both. `login`, `target`, and `ext` are command functions that also carry their subcommands as attached properties, so the loader registered the bare command and then skipped the recursion that would have registered the subcommands. They stayed reachable only through the lazy resolver at runtime.

## Fix

I pulled the loader out into `loadAllCommands` and made it load command functions *and* recurse into objects and functions alike, so subcommands attached to a command that is also a namespace parent get registered for help too.

## Testing

- Added `src/bin/cli.spec.ts` covering the both-leaf-and-parent case (fails on `main`, passes with this change), plus the plain-namespace and `cli`/array-skipping cases.
- `firebase --help` now lists `login:add/ci/list/use`, `target:apply/clear/remove`, and the `ext:*` subcommands; bare commands and other namespaces are unchanged.
- `npm run test:compile`, eslint (`--quiet`), prettier, and the `src/bin` + `src/commands` test suites all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed an issue where `login:*`, `target:*`, and `ext:*` subcommands were missing from `firebase --help`.
 - Functions can declare additional API dependencies (#10621)
 - Added mock Passkey (WebAuthn) support to the Auth emulator. (#10636)
 - Fixes `spawn activate.bat ENOENT` error on Windows when initializing Python functions. (#10608)
--- a/src/bin/cli.spec.ts
+++ b/src/bin/cli.spec.ts
@@ -1,0 +1,60 @@
+import { expect } from "chai";
+import * as sinon from "sinon";
+
+import { loadAllCommands } from "./cli";
+
+function makeCommand(): any {
+  const fn: any = function () {
+    return undefined;
+  };
+  fn.load = sinon.stub();
+  return fn;
+}
+
+describe("cli loadAllCommands", () => {
+  it("loads commands nested under a plain namespace object", () => {
+    const authExport = makeCommand();
+    const client: any = {
+      cli: {},
+      auth: { export: authExport },
+    };
+
+    loadAllCommands(client);
+
+    expect(authExport.load.calledOnce).to.be.true;
+  });
+
+  it("loads subcommands attached to a command that is also a namespace parent", () => {
+    // `login`/`target`/`ext` are command functions that also carry their
+    // subcommands as attached properties. All of them must be registered.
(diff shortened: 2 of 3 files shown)


IzaakGough (commented):
Lgtm awesome work! I think we should probably clean the types slightly (see gemini comments).

sarathfrancis90:
Good call — dropped the `any`s. `loadAllCommands` now takes `Record<string, unknown>` and leans on the existing `isCommandModule` guard for narrowing, and the spec uses a proper `MockCommand` interface. Thanks for the review!

IzaakGough (approved):

joehan (approved):

PR #10678, opened 2026-06-17
nonEmpty param input option support

Berlioz:
Also gets rid of that immutable copy-paste mistake which has never done anything.


src/deploy/functions/params.ts
```diff
@@ -619,7 +626,7 @@ async function promptBooleanParam(
     if (param.description) {
       prompt += ` \n(${param.description})`;
     }
-    return promptText<boolean>(prompt, param.input, resolvedDefault, isTruthyInput);
+    return promptText<boolean>(prompt, param.input, resolvedDefault, false, isTruthyInput);
```
inlined: please always leave a comment on any boolean literal ever used as a parameter

src/deploy/functions/params.ts
```diff
@@ -713,6 +726,7 @@ async function promptResourceString(
   prompt: string,
   input: ResourceInput,
   projectId: string,
+  disallowEmpty: boolean | undefined,
```
inlined: why disallowEmpty and nonEmpty? Also, why mandatory but supports undefined?

src/deploy/functions/params.ts
```diff
@@ -120,11 +120,11 @@ type ParamBase<T extends string | number | boolean | string[]> = {
...
   default?: T | build.Expression<T>;
 
-  // default: false
-  immutable?: boolean;
-
   // Defines how the CLI will prompt for the value of the param if it's not in .env files
   input?: ParamInput<T>;
+
+  // Reject the empty string and empty array as valid input.
+  nonEmpty?: boolean;
```
inlined: I wonder how hard it is to make the field only present for the right types. At minimum you can say:

```
type TypeIfExtends<Result, Test, Base> = Test extends Base ? Result : never;

// ...
nonEmpty: TypeIfExtends<boolean, T, string | string[]>
```

src/deploy/functions/params.ts
```diff
@@ -571,6 +571,7 @@ async function promptList(
       prompt,
       param.input,
       resolvedDefault,
+      param.nonEmpty,
```
inlined: why is nonEmpty not part of input?

inlined (commented):
I think I recommend that nonEmpty is a part of input? we can talk offline.

src/deploy/functions/params.ts
ajperel: Yeah (naming aside) this could just be `disallowEmpty: boolean = false` so there's a default value for the case where we have undefined from it not being in functions.yaml right? (same elsewhere).

I have a hunch on why the name change from nonEmpty to disallowEmpty but I'd like to hear Victor's answer :) ... I do think it's a debateable shift.

src/deploy/functions/params.ts
ajperel: If it is (which from what I understand with a bit of research I would support) do we also need to update the API proposal to be clear about that?

Or we enforce nonEmpty at deployment time and not just input prompting time?

I guess a difference between functions and extensions is users can edit params _after_ the input form by hand editing .env files (or could add nonEmpty _after_ a first deployment that writes an empty string).

ajperel (commented):
Chiming in with a few thoughts since I went through this PR trying to understand it. But overall, I think just having Thomas's LG (and not blocking on more review from me) is fine.

src/deploy/functions/params.ts
Berlioz: Ack

src/deploy/functions/params.ts
Berlioz: I am not as confident about reading type system stuff as you are but my instinct is that this is a huge loss of readability to double-enforce a constraint that the SDK can already enforce much more clearly.

src/deploy/functions/params.ts
Berlioz: Update: we talked it out, updated the SDK, and yeah, moved to input fields.

src/deploy/functions/params.ts
Berlioz: I think changing the var name from nonEmpty to disallowEmpty as it moves from the wire format into the function calling chain makes sense: nonEmpty is describing something about the input itself, while disallowEmpty is more a modifier on the behavior of the prompt: "this prompt is nonempty" seems more vague than "this prompt disallows empty".

That's very much a subjective take though and if you two feel like it's an unnecessary or bad naming decision I really don't mind reverting it.

The boolean | undefined thing was just me being dumb and forgetting that typescript had optional parameters for a second. Fixed.

ajperel: I get the move from describing the input to enforcing the constraint (that was what I thought you were doing). I'd say if we want to make that distinction we could have a slightly better throughline with a different name.

nonEmpty -> enforceNonEmpty or something like that.

Berlioz: oh, yeah, that's a good compromise

src/deploy/functions/params.ts
```diff
@@ -580,15 +584,22 @@ async function promptList(
...
       prompt += ` \n(${param.description})`;
     }
-    return promptText<string[]>(prompt, param.input, resolvedDefault, (res: string): string[] => {
-      return res.split(param.delimiter || ",");
-    });
+    return promptText<string[]>(
+      prompt,
+      param.input,
+      resolvedDefault,
+      false, // enforceNonEmpty
```
inlined: Doesn't text input support nonempty?

src/deploy/functions/params.ts
```diff
@@ -757,12 +793,20 @@ async function promptResourceStrings(
           }),
         },
       };
-      return promptSelectMultiple<string>(prompt, forgedInput, undefined, (res: string[]) => res);
+      return promptSelectMultiple<string>(
+        prompt,
+        forgedInput,
+        undefined,
```
inlined: please comment on what parameter a literal is

src/deploy/functions/params.ts
```diff
@@ -757,12 +793,20 @@ async function promptResourceStrings(
...
+        enforceNonEmpty,
+        (res: string[]) => res,
+      );
     default:
       logger.warn(
         `Warning: unknown resource type ${input.resource.type}; defaulting to raw text input...`,
       );
-      return promptText<string[]>(prompt, { text: {} }, undefined, (res: string) => res.split(","));
+      return promptText<string[]>(prompt, { text: {} }, undefined, enforceNonEmpty, (res: string) =>
```
inlined: same. Comment on what undefined is (I know this is preexisting)

src/deploy/functions/params.ts
```diff
@@ -788,16 +833,20 @@ async function promptText<T extends RawParamValue>(
         textInput.text.validationErrorMessage ||
           `Input did not match provided validator ${userRe.toString()}, retrying...`,
       );
-      return promptText<T>(prompt, textInput, resolvedDefault, converter);
+      return promptText<T>(prompt, textInput, resolvedDefault, enforceNonEmpty, converter);
     }
   }
+  if (enforceNonEmpty && res === "") {
+    logger.error(`Input cannot be the empty string, retrying...`);
```
inlined: Can we be a little less nerdy? E.g. "A value is required, retrying..."

src/deploy/functions/params.ts
Berlioz: Sure does. Straight up copy-paste error.

src/deploy/functions/params.ts
Berlioz: Ack

src/deploy/functions/params.ts
Berlioz: Ack

src/deploy/functions/params.ts
Berlioz: Ack

inlined (approved):

PR #10690, opened 2026-06-19
fix(frameworks): re-assert SSR public invoker on deploy

### LINKED ISSUE #10631: Web frameworks: hosting:channel:deploy creates broken preview channels when the SSR function lacks public invoker IAM

**firebase-tools:** 15.19.1

**Platform:** macOS

### [REQUIRED] Test case

A minimal Next.js app deployed through the Hosting web frameworks integration
(`"source"` + `"frameworksBackend"` in `firebase.json`), where the generated
SSR function's Cloud Run service is missing the public invoker IAM binding
(`allUsers` with `roles/run.invoker`).

Note: the app needs at least one page using `next/image` (or some other
reason for a backend, like a dynamic route). A fully static app never gets an
SSR function created, so there is nothing to reproduce against. Image
optimization is the easiest trigger and also what surfaces the failure on
otherwise prerendered pages.

We hit this state in production: the IAM policy on our `ssr<site>` service
was completely empty (just an etag). We don't know exactly how it got that
way. The repro below sidesteps that question by
removing the binding manually, which puts the service in the same state.

```json
// firebase.json
{
  "hosting": {
    "source": ".",
    "frameworksBackend": { "region": "us-central1" }
  }
}
```

1. `npx create-next-app@latest repro` and add a page that renders a
   statically imported image with `next/image`. This is required, otherwise
   no SSR function is created and the deploy is fully static.
2. `firebase init hosting` (web frameworks), then deploy live with
   `firebase deploy --only hosting`. Everything works.
3. Put the service in the broken state:

   ```bash
   gcloud run services remove-iam-policy-binding ssr<site> \
     --member=allUsers --role=roles/run.invoker --region us-central1
   ```

4. Deploy a preview channel:

   ```bash
   firebase hosting:channel:deploy preview --only <site>
   ```

   The command succeeds and prints the channel URL.
5. Open the channel URL and load a page with an image.

The channel deploy should check that the rewrite target is publicly
invokable, and either re-assert the invoker binding (live frameworks deploys
do this when they redeploy the function) or fail loudly. Even a one-line
warning ("the ssr function is not publicly invokable, dynamic routes on this
channel will return 403") would save a lot of debugging time.

The channel deploy reports success, but the channel is broken for all
dynamic content:

- Prerendered pages load fine, since they come from the Hosting CDN.
- Anything that has to invoke the SSR function, including `/_next/image`,
  returns Google's generic 403 page ("Error: Forbidden. Your client does not
  have permission to get URL ...") with nothing in the function logs. The
  request is rejected by IAM before the container runs, so there is no
  invocation to log. For example:

  ```
  https://<site>--preview-<hash>.web.app/_next/image?url=%2F_next%2Fstatic%2Fmedia%2F1.479aff49.jpg&w=1080&q=75
  -> 403 Forbidden
  https://<site>--preview-<hash>.web.app/about
  -> 200 (prerendered, served from CDN)
  ```

- Meanwhile the live channel looks healthy the whole time, because the
  popular `_next/image` URLs are already in the Hosting CDN cache. That makes
  this look like a preview-channel-specific problem and makes it very hard to
  search for.

What finally surfaced it:

```bash
$ gcloud run services get-iam-policy ssr<site> --region us-central1
etag: ACAB
```

That's an empty policy, no allUsers binding at all. Restoring it fixed every
channel immediately, no redeploy needed:

```bash
gcloud run services add-iam-policy-binding ssr<site> \
  --member=allUsers --role=roles/run.invoker --region us-central1
```

Running the channel deploy with `--debug` shows no IAM-related warning or
error anywhere, which is the core of the issue. Happy to attach the full
debug logs if that helps.

leoortizz:
### Description

Fixes #10631.

Framework SSR runs on a Cloud Run function that Hosting invokes anonymously, so its service needs `allUsers` as `run.invoker`. The generated function declares no invoker, so only the create path set it and the update path never re-applied it. When that binding gets removed out-of-band such as by an org policy, later deploys don't restore it, and `hosting:channel:deploy` ships a channel that 403s on all server-rendered content while reporting success.

This adds `run.ensureInvokerPublic(service)`, called for frameworks SSR functions on deploy. It is additive: it appends `allUsers`, keeps any existing invoker members, and is a no-op when `allUsers` is already present, so it never overwrites deliberate IAM changes. It is also non-fatal: if the grant is blocked it logs a warning instead of failing the deploy. The check is scoped to frameworks codebases, so other functions are untouched.

### Scenarios Tested

Ran a Next.js app with next/image and dynamic routes on a live project. I stripped `allUsers` and ran `hosting:channel:deploy`; the binding was restored automatically and server-rendered routes returned 200. I then pre-seeded a custom invoker service account, stripped `allUsers`, and deployed; `allUsers` was added back and the custom account was preserved.

The issue author suspected the preview deploys might be stripping the binding, so I checked: shipped 15.19.1 does not modify the invoker policy on live or channel deploys, which points to the loss being external rather than deploy-caused.

Added unit tests for `ensureInvokerPublic` covering add, additive, and no-op, and for the `updateV2Function` branch covering ensure, warn-on-failure, and skip for non-frameworks functions.

### Sample Commands

No new commands or flags.

    firebase deploy --only hosting
    firebase hosting:channel:deploy preview


inlined (changes requested):
If I recall correctly, this should be authenticated by the Hosting service directly, not wide permissions on Run. Can you sanity check me @bkendall ?

leoortizz:
> If I recall correctly, this should be authenticated by the Hosting service directly, not wide permissions on Run. Can you sanity check me @bkendall ?

Sounds good, let me know how to proceed. As I'm more focused on web frameworks, I'm not fully aware of how CF should be handled in this case. Would changing this to just a warning for now allow us to merge this sooner? The OP mentioned that a warning alone would have helped him identify the issue and fix it much faster. I can do a follow-up if we end up deciding that invokers should be set to public for web frameworks.

CC @bkendall

PR #10698, opened 2026-06-22
Add warning logs for Local Builds + ABIU if using the wrong versions of nodejs

falahat:
### Description

Universal Maker will use the user's own nodejs version installed locally to build the app. It does NOT use package.json#engines nor backend.runtime (the ABIU setting.)

However, for the runtime nodejs version, we DO respect the ABIU setting. That means you can build an app locally with nodejs 22 and then deploy it, using nodejs24 during runtime. In fact, I tested this and it worked. 

This PR adds warning logs to detect if you are using inconsistent nodejs versions. For the most part, we just log warnings so that you are aware if this doesnt work, but we don't block users (unless ABIU is not enabled at all.)

### DIFF
--- a/src/apphosting/localbuilds.ts
+++ b/src/apphosting/localbuilds.ts
@@ -331,8 +331,8 @@ export function validateLocalBuildNodeVersion(backend: Backend, projectRoot: str
         `and your deployed app will run on the backend's configured ABIU runtime (${runtimeValue}).`,
     );
 
-    const targetVersion = `${targetMajor}.0.0`;
-    if (!semver.satisfies(targetVersion, enginesNode)) {
+    const targetRange = `^${targetMajor}.0.0`;
+    if (!semver.intersects(targetRange, enginesNode)) {
       logLabeledWarning(
         "apphosting",
         `The Node.js version range specified in your package.json engines ("${enginesNode}") ` +
--- a/src/apphosting/localbuilds.ts
+++ b/src/apphosting/localbuilds.ts
@@ -332,7 +332,7 @@ export function validateLocalBuildNodeVersion(backend: Backend, projectRoot: str
     );
 
     const targetRange = `^${targetMajor}.0.0`;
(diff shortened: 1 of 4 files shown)


src/apphosting/localbuilds.ts
```diff
@@ -266,3 +267,92 @@ async function toProcessEnv(projectId: string, env: EnvMap): Promise<NodeJS.Proc
...
+  const localMajor = localMajorMatch ? parseInt(localMajorMatch[1], 10) : null;
+
+  if (localMajor !== null && localMajor !== targetMajor) {
+    logLabeledWarning(
+      "apphosting",
+      `Local Node.js version (${localNodeVersion}) does not match your backend's target ABIU runtime major version (Node.js ${targetMajor}). ` +
+        `Because local builds compile code using your host's Node.js version, this mismatch can cause native modules ` +
+        `to be compiled for the wrong version, leading to silent startup crashes in production (running on Node.js ${targetMajor}). ` +
+        `Please switch your local environment to Node.js ${targetMajor} to ensure build-to-run parity.`,
+    );
```
falahat, line 355: less verbose, less specific warning. Just note that this may cause issues.

src/deploy/apphosting/prepare.spec.ts
```diff
@@ -298,6 +305,7 @@ describe("apphosting", () => {
           {
             name: "projects/my-project/locations/us-central1/backends/foo",
             appId: "my-app-id",
+            runtime: { value: "nodejs22" },
```
falahat, line 308: We shouldn't modify the other test cases (aka, don't shove nodejs22 into unrelated tests here.)

falahat: nevermind these are local build tests

Yuangwang (approved):

Yuangwang (approved):

PR #10706, opened 2026-06-24
inject default Angular SSR environment variables for local builds

falahat:
### Description

This adds NG_ALLOWED_HOSTS and NG_ALLOWED_PROXY_HEADERS similar to Firebase App Hosting Source Deploys. This is required for Angular SSR to work with Firebase Deployments. 

### Scenarios Tested

I tested Angular v19 through v22 with local vs source deploys and confirmed that the results were correct and the same (not falling back to CSR for local builds.)

### DIFF
--- a/src/deploy/apphosting/prepare.spec.ts
+++ b/src/deploy/apphosting/prepare.spec.ts
@@ -822,20 +822,20 @@ describe("apphosting", () => {
       readFileSyncStub = sinon.stub(fs, "readFileSync");
     });
 
-    it("should do nothing for non-Angular applications", () => {
+    it("should do nothing for non-Angular applications", async () => {
       const cfg: AppHostingSingle = { backendId: "foo", rootDir: "/", ignore: [] };
       const buildEnv: Record<string, EnvMap> = { foo: {} };
       const runtimeEnv: Record<string, EnvMap> = { foo: {} };
 
       existsStub.returns(false);
 
-      injectAngularEnvVars(cfg, "/app-dir", buildEnv, runtimeEnv);
+      await injectAngularEnvVars(cfg, "/app-dir", "my-project", "us-central1", buildEnv, runtimeEnv);
 
       expect(buildEnv["foo"]).to.be.empty;
       expect(runtimeEnv["foo"]).to.be.empty;
     });
 
-    it("should inject defaults for Angular applications when headers are missing", () => {
+    it("should inject defaults for Angular applications when headers are missing", async () => {
       const cfg: AppHostingSingle = { backendId: "foo", rootDir: "/", ignore: [] };
       const buildEnv: Record<string, EnvMap> = { foo: {} };
       const runtimeEnv: Record<string, EnvMap> = { foo: {} };
@@ -849,19 +849,19 @@ describe("apphosting", () => {
         }),
       );
 
-      injectAngularEnvVars(cfg, "/app-dir", buildEnv, runtimeEnv);
+      await injectAngularEnvVars(cfg, "/app-dir", "my-project", "us-central1", buildEnv, runtimeEnv);
 
       expect(runtimeEnv["foo"]["NG_TRUST_PROXY_HEADERS"]).to.deep.equal({
         value: "x-forwarded-host,x-forwarded-port,x-forwarded-proto,x-forwarded-for",
         availability: ["RUNTIME"],
       });
       expect(runtimeEnv["foo"]["NG_ALLOWED_HOSTS"]).to.deep.equal({
-        value: "*.hosted.app,*.run.app,*.firestack.app,localhost",
+        value: "foo-123456789.us-central1.run.app,foo--my-project.us-central1.hosted.app,foo--my-project.web.app,foo--my-project.firebaseapp.com",
         availability: ["RUNTIME"],
       });
     });
 
-    it("should accept valid NG_TRUST_PROXY_HEADERS overrides", () => {
+    it("should accept valid NG_TRUST_PROXY_HEADERS overrides", async () => {
       const cfg: AppHostingSingle = { backendId: "foo", rootDir: "/", ignore: [] };
       const buildEnv: Record<string, EnvMap> = { foo: {} };
       const runtimeEnv: Record<string, EnvMap> = {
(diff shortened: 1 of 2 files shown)


src/deploy/apphosting/prepare.spec.ts
```diff
@@ -811,4 +812,205 @@ describe("apphosting", () => {
...
+      );
+
+      injectAngularEnvVars(cfg, "/app-dir", buildEnv, runtimeEnv);
+
+      expect(runtimeEnv["foo"]["NG_TRUST_PROXY_HEADERS"]).to.deep.equal({
+        value: "x-forwarded-host,x-forwarded-port,x-forwarded-proto,x-forwarded-for",
+        availability: ["RUNTIME"],
+      });
+      expect(runtimeEnv["foo"]["NG_ALLOWED_HOSTS"]).to.deep.equal({
+        value: "*.hosted.app,*.run.app,*.firestack.app,localhost",
```
falahat: this is too permissive, I need to scope it to the user's project/backend, something like

```
[BACKEND_ID]--[PROJECT_ID].[REGION].[SHARED_DOMAIN],
[BACKEND_ID]--[PROJECT_ID].web.app,
[BACKEND_ID]--[PROJECT_ID].firebaseapp.com
```

src/deploy/apphosting/prepare.ts
```diff
@@ -423,3 +425,114 @@ async function prepareLocalBuildScratchDirectory(
...
+ * Returns true if the directory contains an Angular application.
+ */
+function isAngularApplication(appDir: string): boolean {
+  const packageJsonPath = path.join(appDir, "package.json");
+  if (fs.existsSync(packageJsonPath)) {
+    try {
+      const parsed: unknown = JSON.parse(fs.readFileSync(packageJsonPath, "utf8"));
+      if (parsed && typeof parsed === "object") {
+        const pkg = parsed as PackageJson;
+        const deps = { ...pkg.dependencies, ...pkg.devDependencies };
```
sjjj986: I think this is creating new object in memory and shallow copying each key value dependency and might be inefficient. Maybe its better to check existence directly using `pkg.dependencies?.["@angular/core"]`.

src/deploy/apphosting/prepare.ts
```diff
@@ -423,3 +425,114 @@ async function prepareLocalBuildScratchDirectory(
...
+        if (deps["@angular/core"] || deps["@angular/ssr"] || deps["@angular/platform-server"]) {
+          return true;
+        }
+      }
+    } catch (e: unknown) {
+      logLabeledWarning(
```
sjjj986: In our internal version, we are erroring out when `isAngularApplication` returns an error, so I think we should log error here to match that instead of logging a warning.

src/deploy/apphosting/prepare.ts
```diff
@@ -423,3 +425,114 @@ async function prepareLocalBuildScratchDirectory(
...
+        "apphosting",
+        `Failed to parse package.json in ${appDir}: ${e instanceof Error ? e.message : String(e)}`,
```
sjjj986: Nit: it'd be better to keep our error message as consistent as possible between go and typescript code "error when checking if application is angular: ... "

src/deploy/apphosting/prepare.ts
```diff
@@ -423,3 +425,114 @@ async function prepareLocalBuildScratchDirectory(
...
+
+  const backendId = cfg.backendId;
+  runtimeEnv[backendId] ??= {};
+  buildEnv[backendId] ??= {};
+
+  const allowedProxyHeadersValue =
+    "x-forwarded-host,x-forwarded-port,x-forwarded-proto,x-forwarded-for";
+
+  // 1. Inject NG_TRUST_PROXY_HEADERS
+  // If they have defined RUNTIME env vars for the proxy header though we will log a warning that we're overriding it.
```
sjjj986: Our preparer go code validates if user-defined proxy headers is a subset of the allowed values, and if so respects user configuration. And it only overrides when user input is empty? Should we match this logic here too in CLI.

src/deploy/apphosting/prepare.ts
```diff
@@ -423,3 +425,114 @@ async function prepareLocalBuildScratchDirectory(
...
+      .filter(Boolean) || [];
+
+  const additionalHosts = userHosts.filter((h) => {
+    const lower = h.toLowerCase();
+    if (defaultHostsLower.has(lower)) return false;
+    defaultHostsLower.add(lower);
+    return true;
+  });
+
+  const allowedHostsValue = [...defaultHosts, ...additionalHosts].join(",");
```
sjjj986: This logic here is merging user defined hosts with default hosts no matter what situation, while our Go buildpack does not inject default values if the user has defined `NG_ALLOWED_HOSTS` env var themselves.

src/deploy/apphosting/prepare.ts
```diff
@@ -423,3 +425,114 @@ async function prepareLocalBuildScratchDirectory(
...
+
+  runtimeEnv[backendId]["NG_TRUST_PROXY_HEADERS"] = {
+    value: allowedProxyHeadersValue,
+    availability: ["RUNTIME"],
+  };
+
+  // 2. Inject/Merge NG_ALLOWED_HOSTS
+  const projectNumber = await getProjectNumber({ project: projectId });
+  const sharedDomain = "hosted.app";
+  const defaultHosts = [
```
sjjj986: This is missing the custom domains retrieved dynamically by the control plane during Cloud Build creation?

falahat: ahh good call out. I think realistically, we won't be able to fetch the custom domains ahead of time in the CLI, so I will document this as a feature gap. For future improvements, we can consolidate the angular handling.

src/deploy/apphosting/prepare.ts
falahat: Good call, I'll adjust this to trust the user's setting if they have a set value. Same with the user proxy headers.

src/deploy/apphosting/prepare.ts
falahat: Good call, I'll match this to defer to the user if they have defined somthing already.

sjjj986 (approved):

PR #10713, opened 2026-06-24
tighten security posture of workflows

bkendall:
### Description

Updates workflows for security policy requirements


joehan (approved):

joehan (changes requested):
Hmmmm, this maybe broke the workflows...

joehan:
Aha - https://github.com/firebase/firebase-tools/actions/runs/28136147542 we need to explicitly allowlist these actions. Will do so tomorrow monring

PR #10714, opened 2026-06-24
.env-based secret param binding for Functions

Berlioz:
relnote: none

Pulls secrets from the Functions .env files and propagates them into the Functions Build's SecretEnvVars (and from there, into the resolved Backend, and from there, into the JSON uploaded to Cloud Functions/Cloud Run).

Params support not included in this PR, to reduce the size of the review. This is obviously going to be a terrible user experience before params are supported (hence the experiment flag) but this should be enough to test the actual binding behavior.

Quick guide to file change intent:

environments.ts - Adds an experiment flag which by default prevents any changes from being made to the Build. If the experiment is not enabled, FIREBASE_SECRET_REF_ envs will still be accepted by the parser, but they'll do nothing (and will not appear in process.env)

env.ts - Modifies the key validation logic to allow keys that begin with FIREBASE_SECRET_REF_ even though FIREBASE_ is reserved. Otherwise knows nothing about this feature.

prepare.ts - Partitions the key-value pairs passed from env.ts into the normal entries and the secrets, which have FIREBASE_SECRET_REF_ stripped from the key at this point. If feature flag enabled, calls build.applyEnvSecretOverrides()

build.ts - Implements applyEnvSecretOverrides, which mutates a Build's state to upsert .env-defined secrets into SecretEnvVars before the Build is compiled into a Functions Backend. These secrets contain an `allowVersionPinning` sentinel value if the version is specified, which will be relevant real soon now.

backend.ts - Adds the `allowVersionPinning` to a Backend's version of a SecretEnvVar so it can propagate to...

validate.ts - Bypasses the logic that forces secret versions to always resolve to the current `latest` if `allowVersionPinning` is set.


src/functions/env.ts
```diff
@@ -46,21 +46,29 @@ const RESERVED_KEYS = [
...
   "gms"                      // flags: global, multiline, dotall
 );
 
+// Regexp to capture a valid reference to a Secret.
+const SECRET_REF_RE = new RegExp(
+  "^" + // begin line
+    "^[a-z]([-a-z0-9]*[a-z0-9])?" + // cloud resource ID
+    "(?:@(.+))?" + // optional reference to a specific version
+    "$", // end line
+);
```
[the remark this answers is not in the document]
Berlioz: Just for the record, this regex isn't supposed to capture secret names, it's supposed to capture Secret resource IDs. The double ^ was an issue though.

src/deploy/functions/build.ts
```diff
@@ -9,6 +9,15 @@ import { ExprParseError } from "./cel";
 import { defineSecret } from "firebase-functions/params";
 
 export const REGION_TBD = "REGION_TBD";
+export const SECRET_REF_PREFIX = "FIREBASE_SECRET_REF_";
+// prettier-ignore
+export const SECRET_REF_RE = new RegExp(
+  "^" +                                      // start of string
+  "csm:\/\/" +                               // scheme csm://
+  "([a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)" +    // secret ID is a GCP resource ID
+  "(?:\/([-a-z0-9]+))?" +                    // capture an optional GCP label for version
```
inlined: It is either a number or "latest"

src/deploy/functions/prepare.spec.ts
```diff
@@ -20,6 +20,18 @@ import { ValidatedConfig } from "../../functions/projectConfig";
 import { BEFORE_CREATE_EVENT, BEFORE_SIGN_IN_EVENT } from "../../functions/events/v1";
 import { latest } from "./runtimes/supported";
 
+describe("partition env helper", () => {
+  it("splits a Record into two based on which keys begin with FIREBASE_SECRET_REF", () => {
+    const input = {
+      foo: "bar",
+      FIREBASE_SECRET_REF_baz: "quux",
+    } as Record<string, string>;
+    const { userEnvs: userEnvs, secretRefs: secretRefs } = prepare.partitionUserEnvs(input);
```
inlined, line 31: FYI: functional.ts already has a partition function. It may be worth just making that support partitioning objects not just arrays.

src/deploy/functions/prepare.ts
```diff
@@ -648,6 +654,9 @@ function warnIfDartBackendHasUnsupportedTriggers(want: backend.Backend): void {
 // Genkit almost always requires an API key, so warn if the customer is about to deploy
 // a function and doesn't have one. To avoid repetitive nagging, only warn on the first
 // deploy of the function.
+/**
+ *
+ */
```
inlined, line 816: No empty javadocs

src/experiments.ts
```diff
@@ -214,6 +214,12 @@ export const ALL_EXPERIMENTS = experiments({
     default: false,
     public: true,
   },
+  secretEnvParams: {
```
inlined, line 217: What is the rationale for keeping this experiment guarded? This won't affect any existing behavior because the key was previously disallowed

src/deploy/functions/build.ts
```diff
@@ -734,3 +743,47 @@ export function applyPrefix(build: Build, prefix: string): void {
...
+/**
+ * Applies overrides from the .env file binding Secrets to a different Cloud Secret Manager resource.
+ * Secrets references are of the form csm://secretName/version, referencing a Secret in the same project as the Endpoint.
+ * /version can be omitted and will cause the secret to resolve to whatever the latest version was at time of deploy.
+ *
+ * For each binding imported from the .env file,
+ * 1) TODO: Check if a conflicting SecretParam with the same name exists. If so, override the param so that the prompting flow will look in the right place when deciding whether or not to create a new Secret.
+ * 2) Upsert the binding directly into the Build's SecretEnvVars, which will cause it to be actually available in process.ENV
+ */
+export function applyEnvSecretOverrides(build: Build, envSecrets: Record<string, string>) {
```
inlined: nit: this method sounds scary. Maybe something like detectSecretReferences?

src/deploy/functions/build.ts
```diff
@@ -734,3 +743,47 @@ export function applyPrefix(build: Build, prefix: string): void {
...
+  Object.entries(envSecrets).forEach(([key, secretRef]) => {
```
inlined: Google style guide forbids forEach. Use for ... of

inlined (approved):

src/deploy/functions/prepare.ts
Berlioz: Ack

src/experiments.ts
Berlioz: Uh at first this was written before API review so I had to guard it in case it was merged before formal approval. I actually do like the idea of guarding it until the params implementation is done though, since anyone who tries to use this without param prompting is in for some undefined and unpleasant behavior.

src/deploy/functions/build.ts
Berlioz: Ack

src/deploy/functions/build.ts
Berlioz: Ack

src/deploy/functions/build.ts
Berlioz: Discussed over chat: GCP implemented a new feature where users could create arbitrary aliases for specific versions.

src/deploy/functions/build.spec.ts
```diff
@@ -3,6 +3,59 @@ import * as build from "./build";
...
+
+  it("errors on bad formats", () => {
+    expect(() => {
+      build.parseSecretRef("");
+    }).to.throw(FirebaseError, /Unknown format/);
+    expect(() => {
+      build.parseSecretRef("b@d:characters");
+    }).to.throw(FirebaseError, /Unknown format/);
+    expect(() => {
+      build.parseSecretRef("UPPERCASE");
```
inlined: Why is this bad? Aren't all legacy cases uppercase?

src/deploy/functions/build.spec.ts
```diff
@@ -389,6 +442,175 @@ describe("envWithType", () => {
...
+        projectId: "other-project",
+        secretId: "bar",
+      },
+    };
+    expect(() => build.applyEnvSecretBindings(testBuild, testSecretRefs)).to.throw(
+      /unsupported cross-project secret/,
+    );
+  });
+
+  it("should add the referenced secret to secretEnvironmentVariables if not present", () => {
```
inlined: This is not expected/desired behavior. We require explicit binding to secrets as part of the principle of least privilege

src/deploy/functions/build.spec.ts
inlined: Also, per gemini, [A-Za-z_-]{1,255} is allowed (including not needing to start alpha

src/deploy/functions/build.spec.ts
```diff
@@ -389,6 +442,175 @@ describe("envWithType", () => {
...
+            },
+          ],
+        },
+      },
+      params: [],
+      requiredAPIs: [],
+    };
+    const testSecretRefs: Record<string, build.ParsedSecretRef> = {
+      foo: {
+        projectId: "test-project",
```
inlined, line 675: nit: can we have stable ordering? AT least of the same keys in the same test?

src/deploy/functions/build.ts
```diff
@@ -9,6 +9,26 @@ import { ExprParseError } from "./cel";
 import { defineSecret } from "firebase-functions/params";
 
 export const REGION_TBD = "REGION_TBD";
+export const SECRET_REF_PREFIX = "FIREBASE_SECRET_REF_";
+// prettier-ignore
+const GCP_RESOURCE_ID_PATTERN = "([a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)"
+export const SECRET_REF_SHORT_RE = new RegExp(
```
inlined, line 17: Wow... took me so long to find something canonical but https://docs.cloud.google.com/secret-manager/docs/reference/rest/v1/projects.secrets/create#query-parameters does say that upper case letters are allowed. Also, there's no mention that it must start with an alpha

src/functions/env.ts
```diff
@@ -10,6 +10,7 @@ import { logBullet, logWarning } from "../utils";
 const FUNCTIONS_EMULATOR_DOTENV = ".env.local";
 
 const RESERVED_PREFIXES = ["X_GOOGLE_", "FIREBASE_", "EXT_"];
+const RESERVED_KEY_PREFIXES = ["FIREBASE_SECRET_REF"];
```
inlined: Nit: isn't a trailing _ part of the prefix too? Also, I wonder if there's a better name for this that makes it clearer.

src/functions/env.ts
```diff
@@ -177,12 +178,20 @@ export function validateKey(key: string): void {
         ", and then consist of uppercase ASCII letters, digits, and underscores.",
     );
   }
-  if (RESERVED_PREFIXES.some((prefix) => key.startsWith(prefix))) {
+  if (
+    RESERVED_PREFIXES.some(
```
inlined: I personally don't like really nested if conditionals like this. Maybe we could have a function called "disallowedKey" now that this logic has gotten so big?

src/deploy/functions/prepare.ts
```diff
@@ -916,3 +927,26 @@ export async function ensureAllRequiredAPIsEnabled(
...
+ * but the Cloud Secret Manager resource they reference is loaded through SecretEnvVars.
+ */
+export function partitionUserEnvs(allEnvs: Record<string, string>): {
+  userEnvs: Record<string, string>;
+  secretRefs: Record<string, string>;
+} {
+  const [secretRefs, userEnvs] = Object.entries(allEnvs).reduce(
+    (accumulatedSplit, [k, v]) => {
+      if (k.startsWith(build.SECRET_REF_PREFIX)) {
+        accumulatedSplit[0][k.replace(new RegExp("^" + build.SECRET_REF_PREFIX), "")] = v;
```
inlined: nit: seems like substr would be more straightforward here.

src/deploy/functions/build.ts
```diff
@@ -764,3 +785,101 @@ export function applyPrefix(build: Build, prefix: string): void {
...
+/**
+ * Applies overrides from the .env file binding Secrets to a different Cloud Secret Manager resource.
+ * Secrets references are of the form csm://secretName/version, referencing a Secret in the same project as the Endpoint.
+ * /version can be omitted and will cause the secret to resolve to whatever the latest version was at time of deploy.
+ *
+ * For each binding imported from the .env file,
+ * 1) TODO: Check if a conflicting SecretParam with the same name exists. If so, override the param so that the prompting flow will look in the right place when deciding whether or not to create a new Secret.
+ * 2) Upsert the binding directly into the Build's SecretEnvVars, which will cause it to be actually available in process.ENV
+ */
+export function applyEnvSecretBindings(build: Build, envSecrets: Record<string, ParsedSecretRef>) {
```
inlined: please fix TSLint error; just add that it's void.

src/deploy/functions/prepare.ts
```diff
@@ -269,14 +269,23 @@ export async function prepare(
...
+    const rawUserEnvs = functionsEnv.loadUserEnvs(userEnvOpt);
+    const { userEnvs: userEnvs, secretRefs: secretRefs } = partitionUserEnvs(rawUserEnvs);
     const envs = { ...userEnvs, ...firebaseEnvs };
 
     const relevantEndpoints = backend
       .allEndpoints(existingBackend)
       .filter((e) => e.codebase === codebase || e.codebase === undefined);
     await resolveDefaultRegionsForBuild(wantBuild, backend.of(...relevantEndpoints));
 
+    if (experiments.isEnabled("secretEnvParams")) {
```
inlined: I think I was wrong regarding whether this is good to have as an experiment because it can change existing behavior, but not here. I would assume that the best place to secret-guard is when we prompt for a secret value in resolving the build, but if they already have a codebase that uses secret refs, just move forwards with it. This prevents weird errors where we're collaborating on a codebase and it works on my computer but not yours.

inlined (changes requested):

src/deploy/functions/build.ts

src/deploy/functions/prepare.ts
Berlioz: Yeah, I can see that. Didn't consider the collaboration angle.

src/deploy/functions/prepare.ts
Berlioz: I totally forgot javascript has .replace and .replaceAll

src/functions/env.ts
Berlioz: Ack

src/functions/env.ts
Berlioz: Honestly, everything I thought of sounded clunky. RESERVED_PREFIX_ALLOWLIST, maybe?

src/deploy/functions/build.ts
Berlioz: Good find. The validation predicator on the internal proto also matches that.

src/deploy/functions/build.spec.ts
Berlioz: Ack

src/deploy/functions/build.spec.ts
Berlioz: Talked about it in chat, we're all agreed on this.

src/deploy/functions/build.spec.ts
Berlioz: Something feels...wrong about not requiring GCP stuff to start alpha, but...

src/deploy/functions/build.ts
```diff
@@ -764,3 +787,104 @@ export function applyPrefix(build: Build, prefix: string): void {
...
+
+    for (const endpointName of Object.keys(build.endpoints)) {
+      const endpoint = build.endpoints[endpointName];
+      if (projectId && projectId !== endpoint.project) {
+        throw new FirebaseError(
+          `Secret binding ${key} referenced unsupported cross-project secret in '${projectId}'`,
+        );
+      }
+      let notFound = true;
+      endpoint.secretEnvironmentVariables?.forEach((envVar) => {
```
inlined: style guide says to use for ... of last I checked.

src/deploy/functions/build.ts
```diff
@@ -764,3 +787,104 @@ export function applyPrefix(build: Build, prefix: string): void {
...
+ * Parses any of the supported formats used to refer to a Secret in .env:
+ * API_KEY=<secret-id>
+ * API_KEY=<secret-id>:<version>
+ * API_KEY=projects/<project-id>/secrets/<secret-id>
+ * API_KEY=projects/<project-id>/secrets/<secret-id>:<version>
+ * API_KEY=projects/<project-id>/secrets/<secret-id>/versions/<version>
+ * @return An object populated with project id, secret id, and version, with a field undefined if not provided.
+ */
+export function parseSecretRef(ref: string): ParsedSecretRef {
+  const shortMatch = SECRET_REF_SHORT_RE.exec(ref);
```
inlined, line 876: nit: Can JS REs do named captures? Would that be a better simplification here?

src/deploy/functions/prepare.ts
```diff
@@ -269,14 +269,21 @@ export async function prepare(
...
+    const rawUserEnvs = functionsEnv.loadUserEnvs(userEnvOpt);
+    const { userEnvs: userEnvs, secretRefs: secretRefs } = partitionUserEnvs(rawUserEnvs);
     const envs = { ...userEnvs, ...firebaseEnvs };
 
     const relevantEndpoints = backend
       .allEndpoints(existingBackend)
       .filter((e) => e.codebase === codebase || e.codebase === undefined);
     await resolveDefaultRegionsForBuild(wantBuild, backend.of(...relevantEndpoints));
 
+    const parsedSecretRefs = Object.fromEntries(
```
inlined: This is already in functional.ts as `mapObject`

src/deploy/functions/prepare.spec.ts
inlined: Actually, I stand corrected; functional.ts already has `partitionRecord` which seems like it should be used instead of a custom function.

src/functions/env.ts
```diff
@@ -10,6 +10,7 @@ import { logBullet, logWarning } from "../utils";
 const FUNCTIONS_EMULATOR_DOTENV = ".env.local";
 
 const RESERVED_PREFIXES = ["X_GOOGLE_", "FIREBASE_", "EXT_"];
```
inlined: Maybe it would be good to add KIT_ to this now and see if anyone yells before we start depending on it being a reserved prefix.

inlined (approved):

src/functions/env.ts
Berlioz: Oooh, that's a good idea.

src/deploy/functions/build.ts
Berlioz: I considered it, but the way i have the last capture group for the version jankily split up between multiple lines of the composed regex (so I can deal with the #/@ special cases) would make it kind of awkward to keep consistent here.

src/deploy/functions/build.ts
Berlioz: Ack.

src/deploy/functions/prepare.ts
Berlioz: Ack

src/deploy/functions/prepare.spec.ts
Berlioz: It's _almost_ there but the fact that partitionUserEnvs modifies the keys in the process of partitioning (to strip the special key prefix) makes it kind of annoying to use directly. You'd have to take a second pass with something that basically does a mapObject but mutates keys instead of values. Nothing I tried looked any more readable than just having a custom function, unfortunately.

PR #10717, opened 2026-06-25
fix: disable keep-alive on GoogleAuth transporter to avoid Premature …

### LINKED ISSUE #10716: Auth regression in 15.22.2: ADC / Workload Identity Federation fails with "Failed to authenticate, have you run firebase login?"
**Version**: 15.22.2 (15.22.1 works)
**Platform**: GitHub Actions (ubuntu-latest, Node 24), non-interactive deploy authenticating via Workload Identity Federation.
**What happened**: After **15.22.2** was published (2026-06-24), CI deploys that authenticate via Application Default Credentials began failing with:

`Error: Failed to authenticate, have you run firebase login?`

Pinning back to **15.22.1** resolves it immediately. No change to our workflow, secrets, or WIF config.

**Setup**:

- `google-github-actions/auth@v2` with `create_credentials_file: true` and `export_environment_variables: true`, producing an `external_account` credentials file and exporting `GOOGLE_APPLICATION_CREDENTIALS` (confirmed set in the job env).
- Command: `firebase deploy --only functions,firestore,storage --project <id> --non-interactive`

**Repro**:

- In CI, authenticate via WIF so `GOOGLE_APPLICATION_CREDENTIALS` points at the external_account credentials file.
- `npx firebase-tools@15.22.1 deploy --only functions --non-interactive` → succeeds.
- `npx firebase-tools@15.22.2 deploy --only functions --non-interactive` → Failed to authenticate, have you run firebase login?

joehan:
### Description
Disables HTTP keep-alive on the transporter utilized by `google-auth-library` to fetch and exchange auth tokens (e.g. via WIF or ADC). This works around a Node.js keep-alive socket reuse regression shipped in security updates for Node 22.23.0 and Node 24.17.0 (referenced in nodejs/node#63989) which causes token exchange requests to fail with `Premature close`.

Fixes #10716

### Scenarios Tested
Wrote up a minimal reproduction script `scripts/repro.ts` under Node v24.17.0 using Application Default Credentials:
- Verified it fails on v15.22.1 and v15.22.2 (without the fix) with a `Premature close` GaxiosError.
- Verified it successfully passes after applying the fix, making the Google API call and receiving a standard API response.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Disable 'keep-alive' in `google-auth-library` calls to avoid `Premature close` errors on some Node versions (#10716).
--- a/src/apiv2.ts
+++ b/src/apiv2.ts
@@ -168,7 +168,7 @@ function proxyURIFromEnv(): string | undefined {
 // https://github.com/node-fetch/node-fetch/issues/1767.
 const httpAgentNoKeepAlive = new http.Agent({ keepAlive: false });
 const httpsAgentNoKeepAlive = new https.Agent({ keepAlive: false });
-function noKeepAliveAgent(parsedURL: URL): http.Agent | https.Agent {
+export function noKeepAliveAgent(parsedURL: URL): http.Agent | https.Agent {
   return parsedURL.protocol === "https:" ? httpsAgentNoKeepAlive : httpAgentNoKeepAlive;
 }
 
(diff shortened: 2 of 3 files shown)


ajperel (approved):
Seems pretty straightforward fix to the Node regression. Thanks for quickly fixing CLI!

PR #10719, opened 2026-06-25
add release vscode workflow; fix changelog

hlshen:
Create a release-vscode-ext workflow, used to trigger the Kokoro build, sign, and publish pipeline.

### DIFF
--- a/.github/workflows/release-vscode-ext.yaml
+++ b/.github/workflows/release-vscode-ext.yaml
@@ -6,7 +6,12 @@ on:
 
 jobs:
   release:
-    if: github.event.issue.pull_request && startsWith(github.event.comment.body, '/run-release')
+    if: |
+      github.event.issue.pull_request &&
+      startsWith(github.event.comment.body, '/run-release') &&
+      (github.event.comment.author_association == 'OWNER' ||
+       github.event.comment.author_association == 'MEMBER' ||
+       github.event.comment.author_association == 'COLLABORATOR')
     runs-on: ubuntu-latest
     steps:
       - name: Checkout PR branch
@@ -17,8 +22,10 @@ jobs:
           fetch-depth: 0
 
       - name: Push to trigger branch
+        env:
+          COMMENT_BODY: ${{ github.event.comment.body }}
         run: |
-          if [[ "${{ github.event.comment.body }}" == *"--build-only"* ]]; then
+          if [[ "$COMMENT_BODY" == *"--build-only"* ]]; then
             echo "Build-only mode detected. Creating skip_publish.txt"
             git config user.name "github-actions[bot]"
             git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
--- a/.github/workflows/release-vscode-ext.yaml
+++ b/.github/workflows/release-vscode-ext.yaml
@@ -4,6 +4,11 @@ on:
   issue_comment:
     types: [created]
 
+permissions:
+  contents: write
+  pull-requests: read
+  issues: read
+
 jobs:
   release:
     if: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release-vscode-ext.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Checkout PR branch
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ github.event.repository.full_name }}
+          ref: refs/pull/${{ github.event.issue.number }}/head
+          fetch-depth: 0
+
+      - name: Push to trigger branch
+        run: |
+          if [[ "${{ github.event.comment.body }}" == *"--build-only"* ]]; then
```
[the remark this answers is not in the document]
hlshen: resolved as string instead of inline code

fredzqm (approved):

.github/workflows/release-vscode-ext.yaml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          PR_NUM="${{ github.event.issue.number }}"
+          echo "Fetching SHA for PR $PR_NUM..."
+          SHA=$(gh pr view "$PR_NUM" --json headRefOid -q .headRefOid)
+          echo "PR Head SHA is $SHA"
+
+          if [[ "$COMMENT_BODY" == *"--build-only"* ]]; then
+            TARGET_BRANCH="release-trigger-build-only"
+          else
+            TARGET_BRANCH="release-trigger"
```
joehan: Can we make this `vscoce-release-trigger` and `vscoce-release-trigger-build-only`? Bit misleading otherwise.

joehan (approved):
Seems fine once we change the branch name

.github/workflows/release-vscode-ext.yaml
```diff
@@ -0,0 +1,48 @@
+name: Release VS Code Extension
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  contents: write
```
joehan, line 8: Granting write is a little scary (tho necessary for this appraoch)... did you run this by @morganchen12?

.github/workflows/release-vscode-ext.yaml
```diff
@@ -0,0 +1,48 @@
...
+  pull-requests: read
+  issues: read
+
+jobs:
+  release:
+    if: |
+      github.event.issue.pull_request &&
```
joehan, line 15: This triggers a release before we merge the commit, no?

joehan (changes requested):

.github/workflows/release-vscode-ext.yaml
joehan: Shouldn't this only happen after we merge the PR to main?

hlshen: I'm actually of the opinion that the current process is backwards. With the requirement to always be updated with main, we end up picking up some new changes while waiting for PR approval. As long as it's after the CLI releaes, I think building it first, then merging is OK

.github/workflows/release-vscode-ext.yaml
hlshen: done, good idea.

.github/workflows/release-vscode-ext.yaml
```diff
@@ -0,0 +1,48 @@
+name: Release VS Code Extension
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
```
morganchen12, line 7: This is going to get flagged in the zizmor rollout, instead of putting it in the top-level of the workflow file you should add it to the specific step that requires that permission.

.github/workflows/release-vscode-ext.yaml
```diff
@@ -0,0 +1,48 @@
...
+  issues: read
+
+jobs:
+  release:
+    if: |
+      github.event.issue.pull_request &&
+      startsWith(github.event.comment.body, '/run-release') &&
+      (github.event.comment.author_association == 'OWNER' ||
+       github.event.comment.author_association == 'MEMBER' ||
+       github.event.comment.author_association == 'COLLABORATOR')
```
morganchen12: Remove collaborator unless you want non-org members making the release (not recommended)

.github/workflows/release-vscode-ext.yaml
```diff
@@ -0,0 +1,48 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger Kokoro via API
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COMMENT_BODY: ${{ github.event.comment.body }}
+        run: |
+          PR_NUM="${{ github.event.issue.number }}"
```
morganchen12, line 29: This will also get flagged by zizmor, you should move it to the `env` section above.

morganchen12 (changes requested):

.github/workflows/release-vscode-ext.yaml
morganchen12: Actually it looks like this is ok since there's only one action in this workflow.

.github/workflows/release-vscode-ext.yaml
morganchen12: Seems like this is not true and zizmor doesn't flag issue numbers being templated into the script.

.github/workflows/release-vscode-ext.yaml
hlshen: done

.github/workflows/release-vscode-ext.yaml
morganchen12: I chatted with Harold, this is the minimum perm required for the workflow unfortunately.

hlshen: reviewed with morgan

joehan (approved):

.github/workflows/release-vscode-ext.yaml
```diff
@@ -0,0 +1,47 @@
...
+          echo "Fetching SHA for PR $PR_NUM..."
+          SHA=$(gh pr view "$PR_NUM" --json headRefOid -q .headRefOid)
+          echo "PR Head SHA is $SHA"
+
+          if [[ "$COMMENT_BODY" == *"--build-only"* ]]; then
+            TARGET_BRANCH="vscode-release-trigger-build-only"
+          else
+            TARGET_BRANCH="vscode-release-trigger"
```
morganchen12, line 37: Given these are going to be origins for release artifacts, you should work with Joe to set up branch protections for these branches.

.github/workflows/release-vscode-ext.yaml
```diff
@@ -0,0 +1,47 @@
...
+  pull-requests: read
+  issues: read
+
+jobs:
+  release:
+    if: |
+      github.event.issue.pull_request &&
+      startsWith(github.event.comment.body, '/run-release') &&
+      (github.event.comment.author_association == 'OWNER' ||
+       github.event.comment.author_association == 'MEMBER' ||
```
joehan: Lets also assert that the pr aiuthor is an owner or maintainer as well. I don't think we'd ever run this on a contributor branch, but we should encode that into the rule

.github/workflows/release-vscode-ext.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+permissions:
+  contents: write
+  pull-requests: read
+  issues: read
+
+jobs:
+  release:
+    if: |
+      github.event.issue.pull_request &&
```
morganchen12, line 15: Please also add an author check here.

```
(github.event.issue.author_association == 'OWNER' ||
github.event.issue.author_association == 'MEMBER')
```

.github/workflows/release-vscode-ext.yaml
hlshen: Is anyone on fdc team a maintainer? We wouldn't be able to run the release in that case.

hlshen: spoke offline - now checking if PR creator is owner/member

.github/workflows/release-vscode-ext.yaml
hlshen: checkign if pr author == commentor

morganchen12 (approved):
LGTM. Discussed the Wiz finding with Joe offline and it should be ok.

PR #10728, opened 2026-06-26
feat: support Task lifecycle hooks in functions

wandamora:
### Description

This PR introduces support for Cloud Functions (CF3) lifecycle hooks, allowing developers to run actions after deployment events. This only supports Task Queue hooks. Execution of Callable and HTTP hooks will be implemented in a future CL.

The implementation spans:
1. **Manifest & Discovery Parsing**: Parses `lifecycleHooks` config from the generated functions build manifest/metadata. Support is added to define `task` (Task Queue functions), `call` (Callable functions), and `http` (HTTP functions or custom URLs) hooks.
2. **Release Execution Flow**: During deployment release, the CLI compares the codebase's current status and target backend configurations to determine if a deploy is a fresh install (`afterInstall` hook target) or a subsequent update (`afterUpdate` hook target).
3. **Task Queue Hook Execution & Authentication**: Integrates with Cloud Tasks (`cloudtasks.ts`) to enqueue tasks targeting hook functions. Resolves the service account to attach an `oidcToken` so that invocations are securely authenticated.
4. **Validation & Prefixing**: 
   - Validates that target endpoints exist in the backend, match expected triggers (e.g., `task` hook targeting a task queue function), and are GCF Gen 2 functions. 
   - Supports codebase prefixing for target functions in multi-codebase setups.
5. **Logging**: Logs a Google Cloud Console log viewer link to easily monitor hook function execution.

---

### Scenarios Tested

- Verified afterInstall only deploys for new codebases
- Verified afterUpdate only deploys for any changes in the codebase
- Verified that error is thrown if either hook is declared more than once
- Verified task queues are dispatched correctly
- Verified hooks are removed upon deletion
- Verified hooks in multiple codebases depend on their respective codebases
- Verified log URLs are correctly output on success


src/deploy/functions/release/index.ts
```diff
@@ -114,6 +115,10 @@ export async function release(
   const wantBackend = backend.merge(...Object.values(payload.functions).map((p) => p.wantBackend));
   printTriggerUrls(wantBackend, projectNumber);
 
+  for (const [codebase, { wantBackend: w, haveBackend: h }] of Object.entries(payload.functions)) {
```
inlined, line 118: Printing URLs should probably be at the end users don't have to dig through the output to get them

src/deploy/functions/release/lifecycle.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      expect(executed).to.be.false;
+    });
+
+    it("enqueues task when afterInstall TaskQueue hook is configured on fresh install", async () => {
+      const enqueueStub = sandbox.stub(cloudtasks, "enqueueTask").resolves();
+      const loggerStub = sandbox.stub(logger, "info");
+      const wantBackend = backend.of({
+        id: "installHookTask",
+        project: "myProj",
+        region: "us-central1",
```
inlined: nit: Since us-c1 is used as a default in many places, let's use a different region to make sure we're not accidentally using it.

src/deploy/functions/release/lifecycle.ts
```diff
@@ -0,0 +1,167 @@
...
+    logger.debug(`No lifecycle hook configured for event: ${delta}`);
+    return false;
+  }
+
+  if (delta === "afterUpdate" && plan) {
+    const relevantChangesets = Object.entries(plan)
+      .filter(([key]) => {
+        if (!codebase) return true;
+        if (key.startsWith(`${codebase}-`)) return true;
+        if (codebase === "default" && key.startsWith("-")) return true;
```
inlined, line 53: What does - mean here? This might warrant a "why" comment

src/deploy/functions/release/lifecycle.ts
```diff
@@ -0,0 +1,161 @@
...
+  const queueName = cloudtasks.queueNameForEndpoint(targetEndpoint);
+  const bodyStr = taskHook.body ? JSON.stringify(taskHook.body) : "";
+  const body = bodyStr ? Buffer.from(bodyStr).toString("base64") : undefined;
+
+  const url = targetEndpoint.uri;
+  if (!url) {
+    throw new FirebaseError(`Target endpoint "${taskHook.function}" does not have a trigger URI.`);
+  }
+
+  const projectNumber = await getProjectNumber({ projectId: targetEndpoint.project });
```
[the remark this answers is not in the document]
inlined: Seems worth fixing 😄

src/deploy/functions/release/lifecycle.ts
```diff
@@ -0,0 +1,167 @@
...
+  const sa =
```
inlined, line 117: I don't think this was tested

src/deploy/functions/release/lifecycle.ts
```diff
@@ -0,0 +1,167 @@
...
+    logLabeledSuccess(
+      "functions",
+      `Successfully queued task for lifecycle hook ${taskHook.function} in queue ${queueName}.`,
+    );
+    logLabeledBullet(
+      "functions",
+      `View logs for ${taskHook.function} at: ${getCloudConsoleLogUrl(targetEndpoint)}`,
+    );
+  } catch (err: unknown) {
+    const errorMsg = err instanceof Error ? err.message : String(err);
```
inlined, line 150: If we're swallowing an error it may be worth mentioning why

src/deploy/functions/release/lifecycle.ts
```diff
@@ -0,0 +1,167 @@
...
+      `Failed to enqueue task for lifecycle hook ${taskHook.function}: ${errorMsg}`,
+    );
+  }
+}
+
+function findTargetEndpoint(
+  backendSpec: backend.Backend,
+  targetId: string,
+): backend.Endpoint | undefined {
+  for (const endpoint of backend.allEndpoints(backendSpec)) {
```
inlined, line 162: Was just looking up the backend utility method (findEndpoint) and realized that there's an edge case we're not handling here where there may be multiple functions with the same name in different regions. Though I guess that shouldn't be a problem for task queue functions (HTTP functions can trigger this by specifying multiple regions)

src/deploy/functions/backend.ts
```diff
@@ -425,6 +425,24 @@ export interface RequiredAPI {
   api: string;
 }
 
+export interface LifecycleHook {
```
inlined: I'd personally prefer this be a union of different types so taht you can't have more than one trigger. The parser should also verify that only one is set

src/deploy/functions/validate.spec.ts
```diff
@@ -494,6 +494,123 @@ describe("validate", () => {
...
+        const want = backend.of(nonTaskEp);
+        want.lifecycleHooks = {
+          afterInstall: {
+            task: {
+              function: "nontaskfunc",
+            },
+          },
+        };
+        expect(() => validate.endpointsAreValid(want)).to.throw(
+          /Target endpoint "nontaskfunc" is not a task queue function for lifecycle hook "afterInstall"/,
```
inlined: Maybe the error message should explain that the afterInstall hook expects a task queue function? As is, it sounds more like all lifecycle hooks must be TQs

src/deploy/functions/validate.spec.ts
```diff
@@ -494,6 +494,123 @@ describe("validate", () => {
...
+        });
+        want.lifecycleHooks = {
+          afterInstall: {
+            http: {
+              url: "https://example.com/hook",
+            },
+          },
+        };
+        expect(() => validate.endpointsAreValid(want)).to.throw(
+          /Lifecycle hook action type "http" is not supported in the CLI yet./,
```
inlined, line 610: This also reminds me of why my instinct was to have a type union for hooks. When executing the hooks, handle all three cases and then use `assertExhaustive()` That'll help pop exceptions if you define a new hook type wherever it isn't handled.

src/gcp/cloudtasks.ts
```diff
@@ -135,6 +135,24 @@ export async function deleteQueue(name: string): Promise<void> {
...
 
+export interface Task {
+  httpRequest: {
+    url: string;
+    httpMethod?: string;
+    headers?: Record<string, string>;
+    body?: string;
+    oidcToken?: {
+      serviceAccountEmail: string;
+      audience?: string;
```
inlined, line 146: Set this audience so that the token cannot be forwarded

inlined (approved):
Done with a pass. Nothing foundational needs to be changed and I'm off today, so I'm LGTMing

src/deploy/functions/release/index.ts
wandamora: AFAICT this already does? At least it was for when I was deploying functions in a test project.

src/deploy/functions/release/lifecycle.spec.ts
wandamora: Changed to us-east1

src/deploy/functions/release/lifecycle.ts
wandamora: TBH some of it was a gemini code assist suggestion. It claimed that when in the "default" codebase, the keys can start with "-". Added a justification comment.

src/deploy/functions/release/lifecycle.ts
wandamora: Added more testing.

src/deploy/functions/backend.ts
wandamora: Done

src/deploy/functions/validate.spec.ts
wandamora: Done

src/gcp/cloudtasks.ts
wandamora: Set it in lifecycle.ts.

src/deploy/functions/release/lifecycle.ts
wandamora: Should we address that in a separate CL? This seems like a non-blocking issue for this CL.

wandamora: Looking back I totally blanked on replacing this redundant helper function. I address it here: https://github.com/firebase/firebase-tools/pull/10752

Still won't cover the edge case, but cleans up the code.

src/deploy/functions/release/index.ts
```diff
@@ -114,6 +115,10 @@ export async function release(
...
+    await executeLifecycleHooks(w, h, plan, codebase);
```
ajperel, line 119: Didn't we only want to execute this if there were no deployment errors and otherwise warn that hooks weren't run? It looks like we don't look into deployment failures till after this. (or I just haven't gotten far enough to see I'm wrong or it's in a subsequent PR I haven't reviewed yet in which case, I'm sorry). I think we want to fix that and have a test case that validates hook behavior when not everything successfully deploys.

src/deploy/functions/release/lifecycle.ts
```diff
@@ -0,0 +1,178 @@
...
+import * as planner from "./planner";
+import { FirebaseError } from "../../../error";
+import { logger } from "../../../logger";
+import { logLabeledBullet, logLabeledSuccess, logLabeledWarning } from "../../../utils";
+import * as cloudtasks from "../../../gcp/cloudtasks";
+import * as computeEngine from "../../../gcp/computeEngine";
+import { getProject } from "../../../management/projects";
+import { assertExhaustive } from "../../../functional";
+
+export type LifecycleDelta = "afterInstall" | "afterUpdate";
```
ajperel, line 11: Wow somehow this sent me down a naming hole... sorry but to bring it all up but....

1. Interesting that the type is `LifecycleDelta` but the method is `determineDeploymentDelta` ... shouldn't we have consistent naming between the two?

2. On "Delta" that seems to imply to me a diff but we're really just talking about where we are in the lifecycle of a codebase so maybe lifecycleState? Not sure that we'll have a use case for other states in the future though?

3. On "afterInstall" | "afterUpdate" - I'm suddenly realizing this is maybe _Rxtensions_ specific terminology. In the same console we're going to have lines like
    *  `✔  functions[v2-makeUppercaseV2Test(us-west1)] Successful create operation.` 
    *  `✔  functions[v2-addMessageV2(us-central1)] Successful update operation.`
So should we instead be consistent with Functions and rename these to be `onCreate` and `onUpdate` while we still can?

src/deploy/functions/release/lifecycle.ts
```diff
@@ -0,0 +1,178 @@
...
+        return false;
+      })
+      .map(([, c]) => c);
+    const hasResourceModifications = relevantChangesets.some(
+      (changeset) =>
+        changeset.endpointsToCreate.length > 0 ||
+        changeset.endpointsToUpdate.length > 0 ||
+        changeset.endpointsToDelete.length > 0,
+    );
+    if (!hasResourceModifications) {
```
ajperel, line 63: Just to check... if we _only_ changed `.env` that would still count as a resource modification right?

src/deploy/functions/release/index.ts
ajperel: Should failing to *delete* a function also count as a failure stopping afterUpdate?

src/deploy/functions/release/lifecycle.ts
```diff
@@ -0,0 +1,178 @@
...
+    }
+  }
+
+  if ("task" in hook) {
+    logLabeledBullet(
+      "functions",
+      `Executing ${delta} lifecycle hook targeting: ${hook.task.function}...`,
+    );
+    await executeTaskQueueHook(hook.task, wantBackend);
+    return true;
```
ajperel, line 78: Not sure there's a better option, but 'true' here feels potentially wrong given `executeTaskQueueHook` below swallows errors and just logs:       `Failed to enqueue task for lifecycle hook ${taskHook.function}: ${errorMsg}`, since the jsdoc says "true means task queue executed"

Seems like true means... we attempted to run a hook but may have silently swallowed it and false is "we determined we didn't need to run a hook" do maybe better documentation is the answer.

ajperel (commented):
A few comments and one maybe important naming bikeshed while we still haven't released this all.

src/deploy/functions/release/lifecycle.ts
wandamora: Yes, the hash diff is calculated with https://github.com/firebase/firebase-tools/blob/0038038859143bab2a6ac3bd807318d61d7bc942/src/deploy/functions/cache/hash.ts#L45-L52, which includes environment variables.

src/deploy/functions/release/lifecycle.ts
wandamora: You're right that the delta naming is all over the place here. I think we can align the naming for `LifecycleDelta` and `determineDeploymentDelta` to say `LifecycleEvent` instead. I can modify this along with the "afterInstall" / "afterCreate" renaming we discussed in chat in a new PR.

src/deploy/functions/release/index.ts
wandamora: You're right.. I should move this to after we check for errors. I'll make sure to move this along with adding a test case. I agree that failing to delete a function should also count as a failure that stops afterUpdate.

src/deploy/functions/release/lifecycle.ts
wandamora: This part of the code was refactored in https://github.com/firebase/firebase-tools/pull/10747, so the behavior should be as documented now.

src/deploy/functions/release/lifecycle.ts
ajperel: SG. LifecycleEvent works for me. DeploymentEvent would also be fine?

src/deploy/functions/release/index.ts
ajperel: SG

src/deploy/functions/release/lifecycle.ts
wandamora: Done in https://github.com/firebase/firebase-tools/pull/10767/changes

Changed to DeploymentEvent.

src/deploy/functions/release/index.ts
wandamora: Here's the PR: https://github.com/firebase/firebase-tools/pull/10769

PR #10733, opened 2026-06-29
feat: add IAM and Resource Manager utilities for service accounts (PR 2)

inlined:
### Description
This is PR 2 in the requiresRole PR chain.
It implements low-level Cloud IAM and Resource Manager API utilities:
1. `generateManagedServiceAccountName` for collision-resolved SA names.
2. `computeRolesEtag` for salt-and-hash role etag calculations.
3. `addServiceAccountRoles` and `removeServiceAccountRoles` to mutate project IAM policies.
4. Static cache of standard roles in `knownRoles.json` to prevent API roundtrips.

### Scenarios Tested
- Run unit tests: `npx mocha src/gcp/iam.spec.ts src/gcp/resourceManager.spec.ts`
- Run npm run build & npm run lint

### Sample Commands
N/A


src/gcp/iam.ts
```diff
@@ -274,3 +278,71 @@ export function printManualIamConfig(
...
+/**
+ * Generates a unique managed service account name by appending a random number
+ * and checking against GCP until an unused name is found.
+ */
+export async function generateManagedServiceAccountName(
+  projectId: string,
+  prefix: string,
+): Promise<string> {
+  const maxAttempts = 10;
+  for (let i = 0; i < maxAttempts; i++) {
```
Berlioz, line 309: 10 isn't that many, but is there a chance that without any kind of backoff, or if the user is doing something that might call this function multiple times, they might run into quota issues on the GetServiceAccount API call? Especially for low-reputation projects?

Berlioz (approved):
LGTM, with one possibly irrelevant nit

src/gcp/iam.ts
inlined: The quota is 300 requests per minute, so I assume this is not necessary. I can codify that though.

PR #10740, opened 2026-06-29
fix: prevent premature caching of API enablement status

fredzqm:
Fixes a race condition where the CLI prematurely caches Google Cloud APIs as enabled immediately upon initiating the enablement process, leading to false positives during polling.

---
*PR created automatically by Jules for task [14193484741930371072](https://jules.google.com/task/14193484741930371072) started by @fredzqm*


joehan (commented):
So, we added this caching in order to avoid overloading our shared quota. Does this disable it entirely, or is there a later spot where the chaching gets triggered?

PR #10748, opened 2026-06-30
build: push specific docker image version tags during Cloud Build step

avolkovi:


joehan (approved):
Code looks fine, tho i am confused why we need this

PR #10752, opened 2026-07-01
fix: remove duplicate find endpoint function

wandamora:
### Description
`findTargetEndpoint` is a redundant utility function to discover an endpoint. We could simply reuse `backend.findEndpoint` and use the same predicate to discover based on id.

### Scenarios Tested

`npx mocha src/deploy/functions/release/lifecycle.spec.ts`

### DIFF
--- a/src/deploy/functions/release/lifecycle.ts
+++ b/src/deploy/functions/release/lifecycle.ts
@@ -92,7 +92,7 @@ async function executeTaskQueueHook(
   taskHook: { function: string; body?: Record<string, unknown> },
   wantBackend: backend.Backend,
 ): Promise<void> {
-  const targetEndpoint = findTargetEndpoint(wantBackend, taskHook.function);
+  const targetEndpoint = backend.findEndpoint(wantBackend, (ep) => ep.id === taskHook.function);
   if (!targetEndpoint) {
     throw new FirebaseError(
       `Target endpoint "${taskHook.function}" not found in backend for lifecycle hook.`,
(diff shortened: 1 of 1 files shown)


src/deploy/functions/release/lifecycle.ts
```diff
@@ -92,7 +92,7 @@ async function executeTaskQueueHook(
   taskHook: { function: string; body?: Record<string, unknown> },
   wantBackend: backend.Backend,
 ): Promise<void> {
-  const targetEndpoint = findTargetEndpoint(wantBackend, taskHook.function);
+  const targetEndpoint = backend.findEndpoint(wantBackend, (ep) => ep.id === taskHook.function);
```
ajperel, line 95: As a minor comment not quite related to this change... it's not obvious to me why an endpoint id would be equal to taskhook.function ... I can probably dig in and figure it out, but I wonder if there's some better names that would make this more obviously correct with little context. Not  something to change in this PR though and a minor nit.

ajperel (approved):

src/deploy/functions/release/lifecycle.ts
inlined: endpoint is a generic name for "something you can call" in prep for other types than just functions or multiple APIs (e.g. there's a functionToEndpiont and endpointToFunction in both functions APIs and I assume the Run API). It also It was partially an experiment for making the SDK work with App Hosting, but can instead be used for things like multiple endpoints on a service, though we'll need to be careful because I'm sure the abstraction may have calcified some functions-type assumptions.

inlined (approved):

PR #10760, opened 2026-07-03
fix(hosting): preserve request body when a web frameworks rewrite falls through to a Cloud Function

### LINKED ISSUE #5986: Hosting rewrite post call to function times out

**firebase-tools:** 12.3.1

**Platform:** macOS

All default steps with added a rewrite entry in firebase.json, however I created a repo which demonstrates the problem
https://github.com/jkibble/firebase-sveltekit-broken-rewrite

npx firebase emulators:start
navigate to http://localhost:5000 
click the Get call
click the Post call
 
Post calls succeeds 

Get call succeeds 
Post call times out with 504 gateway error without running any server side code

leoortizz:
### Description

Fixes #5986.

With Web Frameworks enabled, the Hosting emulator forwards every request to the framework's dev server before evaluating rewrites, and that consumes the request body. When the dev server returned a 404 and the request fell through to a `function` rewrite, the emulator forwarded an empty body while still advertising the original `content-length`, so the function call hung until it timed out with a 504. `GET` requests worked; anything with a body failed.

The emulator now keeps a copy of the request body before handing the request to the dev server and replays it if the request falls through to a rewrite. Behavior outside Web Frameworks is unchanged, and this only affects the emulator — production Hosting was never affected.

### Scenarios Tested

- SvelteKit (the original report), Vite, Astro, and Angular apps in the Hosting emulator with a `function` rewrite: `POST`/`PUT`/`DELETE` with JSON, empty, chunked, and ~200KB bodies now reach the function with the body intact (previously a 60s hang → 504). Verified against the reporter's reproduction repo.
- The framework's own API routes (e.g. SvelteKit `+server`, Astro endpoints) still receive request bodies normally.
- Requests to paths owned by neither the framework nor a rewrite still return a fast 404.
- Static pages, query strings, and `GET` rewrites behave as before.
- Unit tests

### Sample Commands

```
firebase emulators:start
```


src/frameworks/utils.ts
```diff
@@ -160,6 +161,18 @@ export function proxyResponse(
   return proxiedRes;
 }
 
+/**
+ * Read a request stream into a Buffer so a consumed request body can be replayed
+ * to a downstream handler.
+ */
+async function bufferRequestBody(req: IncomingMessage): Promise<Buffer> {
```
annajowang: do we limit the payload size at all? could a large enough payload cause an OOM crash? I believe cloud functions limits to 10MB anyway, should we just limit if here as well for a more realistic dev experience?

src/frameworks/utils.ts
```diff
@@ -193,6 +206,27 @@ export function simpleProxy(hostOrRequestHandler: string | RequestHandler) {
...
+      // See https://github.com/firebase/firebase-tools/issues/5986
+      let rawBody: Buffer | undefined;
+      if (!["GET", "HEAD"].includes(method)) {
+        try {
+          rawBody = await bufferRequestBody(originalReq);
+        } catch (err) {
+          // Client aborted / stream errored mid-upload. End gracefully rather
+          // than hanging the response.
+          logger.debug("Error buffering request body:", method, path, err);
+          originalRes.end();
```
annajowang: What would be the originalRes.statusCode be in this case?
should it explicitly be set to something?

src/frameworks/utils.ts
```diff
@@ -193,6 +206,27 @@ export function simpleProxy(hostOrRequestHandler: string | RequestHandler) {
...
+      // Buffer the body before forwarding it to the dev server, which consumes
+      // the stream. If the dev server 404s we cascade (call next) to the next
+      // rewrite — e.g. a Cloud Function — and that handler replays the body via
+      // req.rawBody; otherwise it would forward a drained stream with a stale
+      // content-length and hang until timeout.
+      // See https://github.com/firebase/firebase-tools/issues/5986
+      let rawBody: Buffer | undefined;
+      if (!["GET", "HEAD"].includes(method)) {
+        try {
+          rawBody = await bufferRequestBody(originalReq);
```
annajowang, line 230: IIUC this will slow down every single request....could we "Tee" stream this?

annajowang (approved):

src/frameworks/utils.ts
leoortizz: We didn’t limit it before but it’s a simple change, just implemented it. FWICT Web Frameworks’s CF is gen 2, so that is 32MB: https://docs.cloud.google.com/functions/quotas#resource_limits

src/frameworks/utils.ts
leoortizz: Good call thanks, changed this to return 400 when the response is still writable. I previously tried setting 400 but was having an error since the response wasn't writable anymore, but a simple check fixed it.

src/frameworks/utils.ts
leoortizz: I tried it, but the results are pretty much the same across all frameworks. It'd be a rewrite of how we handle the proxy, so I wasn't sure it's worth the risk for this PR. Framework API routes are ~15% faster on large uploads, but rewrites (this PR's/issue case) behave the same.

Since it's emulator/local only and payloads are capped at 32MB, the gain is just a few ms. Happy to implement it if you'd prefer.

Helbss (approved):

annajowang:
thanks! still lgtm, don't have to invest in "Tee" stream for emulator if not worth it

annajowang (approved):

PR #10769, opened 2026-07-07
fix: run lifecycle hooks only for successful deployments

wandamora:
### Description

This change ensures that lifecycle hooks are only triggered if all function deployments (creates, updates, deletes) in the release plan succeeded.

- **Conditional Execution:** Wrapped the lifecycle hook execution block to only run when there are no deployment errors.
- **Billing Prevention:** Kept `setupArtifactCleanupPolicies` running even if there are deployment errors, ensuring Artifact Registry cleanup policies are still set up for successfully deployed functions to avoid unexpected storage charges.
- **Visibility:** Added a warning logger (`utils.logLabeledWarning`) to explicitly inform developers when lifecycle hooks were configured but skipped due to deployment errors.
- **Abstractions:** Implemented a `hasLifecycleHooks` helper function in `lifecycle.ts` to query whether a backend has configured lifecycle hooks.

### Scenarios Tested

- `npx mocha src/deploy/functions/release/index.spec.ts`
- `npx mocha src/deploy/functions/release/lifecycle.spec.ts`

### DIFF
--- a/src/deploy/functions/release/index.ts
+++ b/src/deploy/functions/release/index.ts
@@ -19,7 +19,7 @@ import { getProjectNumber } from "../../../getProjectNumber";
 import { release as extRelease } from "../../extensions";
 import * as artifacts from "../../../functions/artifacts";
 import { runtimeIsLanguage } from "../runtimes/supported";
-import { executeLifecycleHooks, hasLifecycleHooks } from "./lifecycle";
+import { determineDeploymentEvent, executeLifecycleHooks } from "./lifecycle";
 
 /** Releases new versions of functions and extensions to prod. */
 export async function release(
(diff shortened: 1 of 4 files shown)


inlined (approved):
Great! Love the explicit error messaging

src/deploy/functions/release/index.ts
```diff
@@ -125,8 +132,13 @@ export async function release(
     Object.keys(wantBackend.endpoints),
   );
 
-  const allErrors = summary.results.filter((r) => r.error).map((r) => r.error) as Error[];
   if (allErrors.length) {
+    if (hasLifecycleHooks(wantBackend)) {
+      utils.logLabeledWarning(
+        "functions",
+        "Lifecycle hooks were configured but not executed because one or more function deployments failed.",
```
ajperel: Is anything telling the user what hook failed so they can figure out which kit/codebase lifecycle function they might want to manually run?

wandamora: I updated this to specify the codebase and hook that was skipped.

wandamora: and by hook i mean the "deployment event" that would have been triggered.

src/deploy/functions/release/index.ts
```diff
@@ -125,8 +132,18 @@ export async function release(
...
-  const allErrors = summary.results.filter((r) => r.error).map((r) => r.error) as Error[];
   if (allErrors.length) {
+    for (const [codebase, { wantBackend: w, haveBackend: h }] of Object.entries(
+      payload.functions,
+    )) {
+      const event = determineDeploymentEvent(h);
+      if (w.lifecycleHooks?.[event]) {
+        utils.logLabeledWarning(
+          "functions",
+          `Lifecycle hook (${event}) for codebase "${codebase}" was configured but not executed because one or more function deployments failed.`,
```
ajperel: nit:  I'd do `"${event}"` instead of `(${event})` it feels like you're adding necessary detail same as codebase rather than a paranthetical. But this is very nitty. The content is great.

ajperel (approved):

src/deploy/functions/release/index.ts
wandamora: Done!

PR #10771, opened 2026-07-07
Fix env var name in dataconnect:execute help text.

yuchenshi:
### Description

See CHANGELOG diff.

### Scenarios Tested

See Sample Commands.

### Sample Commands

`firebase dataconnect:execute --help`


joehan (approved):
Lgtm, but maybe we should make this a constant

yuchenshi:
> Lgtm, but maybe we should make this a constant

Done but note that the name of the constant is the source of the the confusion in the first place. I'm not sure this is better.

PR #10778, opened 2026-07-08
feat: add firebase ailogic:providers CLI commands

marb2000:
### Description

Adds the `firebase ailogic:providers:*` command group to manage the Gemini API providers for Firebase AI Logic from the CLI, so developers and CI pipelines can enable, disable, and inspect providers without using the Firebase Console.

- `ailogic:providers:enable <providerType>` — enables the Firebase AI Logic API (`firebasevertexai.googleapis.com`) if needed, then the selected provider's underlying API. `gemini-developer-api` enables `generativelanguage.googleapis.com`; `agent-platform-gemini-api` requires the Blaze (pay-as-you-go) plan and enables `aiplatform.googleapis.com`.
- `ailogic:providers:disable <providerType>` — prompts for confirmation (skippable with `--force`) and turns off the proxy when no providers remain enabled.
- `ailogic:providers:list` — reports which providers are enabled.

Supporting changes:
- A shared interactive enablement flow (`ensureAILogicApiEnabled`) used when the API isn't yet enabled, which checks the `serviceusage.services.enable` permission up front so callers get a clear error before any prompts.
- `firebase help <namespace>` now lists the subcommands under a prefix (e.g. `firebase help ailogic:providers`) for discoverability.
- `ensureApiEnabled.uncacheEnabledAPI` to keep the local API-enablement cache consistent after enabling/disabling an API.

All commands live under the fixed `global` location (no `--location` flag), support `--json` output, and reject client-breaking actions in `--non-interactive` mode unless `--force` is passed. Permissions map to the `firebasevertexai` and `serviceusage` IAM permissions.

This is the first of a series of scoped PRs bringing Firebase AI Logic management (providers, configuration, prompt templates, triggers) to the CLI.

### Scenarios Tested

- `providers:enable gemini-developer-api` on a project with the API disabled — enables the AI Logic API and `generativelanguage.googleapis.com`.
- `providers:enable agent-platform-gemini-api` on a Blaze project succeeds; on a Spark project it fails with an upgrade link and does not change the plan.
- `providers:disable` — confirmation prompt shown; `--force` and `--non-interactive` paths covered; proxy is disabled only when no providers remain.
- `providers:list` — reflects enablement state with and without billing.
- Interactive enablement flow — re-prompts on a Spark billing failure; fails up front when the caller lacks `serviceusage.services.enable`.
- `firebase help ailogic:providers` — lists the namespace subcommands.
- Unit tests added in `src/gcp/ailogic.spec.ts`, `src/gcp/serviceusage.spec.ts`, and `src/commands/help.spec.ts`. `npm test` (lint + compile + mocha) passes.

### Sample Commands

```
firebase ailogic:providers:enable gemini-developer-api
firebase ailogic:providers:list
firebase ailogic:providers:list --json
firebase ailogic:providers:disable gemini-developer-api --force
firebase help ailogic:providers
```

### DIFF
--- a/src/commands/ailogic-providers-disable.ts
+++ b/src/commands/ailogic-providers-disable.ts
@@ -1,0 +1,40 @@
+import { Command } from "../command";
+import { requirePermissions } from "../requirePermissions";
+import { needProjectId } from "../projectUtils";
+import * as ailogic from "../gcp/ailogic";
+import * as clc from "colorette";
+import { logger } from "../logger";
+import { FirebaseError } from "../error";
+import { confirm } from "../prompt";
+
+import { Options } from "../options";
+
+export const command = new Command("ailogic:providers:disable <providerType>")
+  .description("disable a Gemini API provider service")
+  .option("-f, --force", "bypass confirmation prompt")
+  .before(requirePermissions, ["serviceusage.services.disable", "firebasevertexai.config.update"])
+  .action(async (providerType: string, options: Options) => {
+    const projectId = needProjectId(options);
+    if (providerType !== "gemini-developer-api" && providerType !== "agent-platform-gemini-api") {
+      throw new FirebaseError(
+        `Invalid provider type: ${clc.bold(providerType)}. Must be 'gemini-developer-api' or 'agent-platform-gemini-api'.`,
+      );
+    }
+    if (options.nonInteractive && !options.force) {
+      throw new FirebaseError(
+        `Disabling provider ${clc.bold(providerType)} requires confirmation.\n\n` +
+          `To proceed in non-interactive mode, rerun with --force:\n\n` +
+          `  firebase ailogic:providers:disable ${providerType} --force`,
+      );
+    }
+    const confirmed = await confirm({
+      message: `You are about to disable ${clc.bold(providerType)}. This will stop running apps from invoking it. Are you sure?`,
+      force: options.force,
+      nonInteractive: options.nonInteractive,
+    });
+    if (!confirmed) {
+      throw new FirebaseError("Command aborted.", { exit: 1 });
+    }
+    await ailogic.disableProvider(projectId, providerType as ailogic.ProviderType);
+    logger.info(clc.green(`Successfully disabled provider: ${clc.bold(providerType)}`));
+  });
--- a/src/commands/ailogic-providers-disable.ts
+++ b/src/commands/ailogic-providers-disable.ts
@@ -15,26 +15,17 @@ export const command = new Command("ailogic:providers:disable <providerType>")
   .before(requirePermissions, ["serviceusage.services.disable", "firebasevertexai.config.update"])
   .action(async (providerType: string, options: Options) => {
     const projectId = needProjectId(options);
-    if (providerType !== "gemini-developer-api" && providerType !== "agent-platform-gemini-api") {
-      throw new FirebaseError(
-        `Invalid provider type: ${clc.bold(providerType)}. Must be 'gemini-developer-api' or 'agent-platform-gemini-api'.`,
-      );
-    }
-    if (options.nonInteractive && !options.force) {
-      throw new FirebaseError(
-        `Disabling provider ${clc.bold(providerType)} requires confirmation.\n\n` +
-          `To proceed in non-interactive mode, rerun with --force:\n\n` +
-          `  firebase ailogic:providers:disable ${providerType} --force`,
-      );
-    }
+    const provider = ailogic.parseProviderType(providerType);
+    // confirm() aborts (throws) in non-interactive mode unless --force is set, so a
+    // separate non-interactive guard is unnecessary here.
     const confirmed = await confirm({
-      message: `You are about to disable ${clc.bold(providerType)}. This will stop running apps from invoking it. Are you sure?`,
+      message: `You are about to disable ${clc.bold(provider)}. This will stop running apps from invoking it. Are you sure?`,
       force: options.force,
       nonInteractive: options.nonInteractive,
     });
     if (!confirmed) {
       throw new FirebaseError("Command aborted.", { exit: 1 });
     }
-    await ailogic.disableProvider(projectId, providerType as ailogic.ProviderType);
-    logger.info(clc.green(`Successfully disabled provider: ${clc.bold(providerType)}`));
+    await ailogic.disableProvider(projectId, provider);
+    logger.info(clc.green(`Successfully disabled provider: ${clc.bold(provider)}`));
   });
--- a/src/commands/ailogic-providers-enable.ts
+++ b/src/commands/ailogic-providers-enable.ts
@@ -1,0 +1,23 @@
+import { Command } from "../command";
+import { requirePermissions } from "../requirePermissions";
+import { needProjectId } from "../projectUtils";
+import * as ailogic from "../gcp/ailogic";
+import * as clc from "colorette";
+import { logger } from "../logger";
+import { FirebaseError } from "../error";
+
+import { Options } from "../options";
+
+export const command = new Command("ailogic:providers:enable <providerType>")
+  .description("enable a Gemini API provider service")
+  .before(requirePermissions, ["serviceusage.services.enable", "firebasevertexai.config.update"])
+  .action(async (providerType: string, options: Options) => {
+    const projectId = needProjectId(options);
+    if (providerType !== "gemini-developer-api" && providerType !== "agent-platform-gemini-api") {
+      throw new FirebaseError(
+        `Invalid provider type: ${clc.bold(providerType)}. Must be 'gemini-developer-api' or 'agent-platform-gemini-api'.`,
+      );
+    }
+    await ailogic.enableProvider(projectId, providerType as ailogic.ProviderType);
+    logger.info(clc.green(`Successfully enabled provider: ${clc.bold(providerType)}`));
+  });
--- a/src/commands/ailogic-providers-enable.ts
+++ b/src/commands/ailogic-providers-enable.ts
@@ -4,7 +4,6 @@ import { needProjectId } from "../projectUtils";
 import * as ailogic from "../gcp/ailogic";
 import * as clc from "colorette";
 import { logger } from "../logger";
-import { FirebaseError } from "../error";
 
 import { Options } from "../options";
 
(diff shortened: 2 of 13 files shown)


marb2000:
Thanks for the review! Addressed all the feedback:

- **`providers:disable` / `providers:list`**: removed the `ensureAILogicApiEnabled` call so these commands no longer force or prompt for API enablement — `list` now just reports providers as disabled and `disable` is a no-op prompt-free path when the API is already off.
- **`listProviders` billing check**: removed. Since `aiplatform.googleapis.com` can't be enabled without billing, an enabled Vertex API already implies the agent-platform provider is available; this also avoids failures for callers without billing-viewer permission.
- **Unit tests**: consolidated the two `listProviders` list tests into one now that billing state is irrelevant.
- **`help.spec.ts` `any` cast**: replaced. Note the exact suggested cast doesn't compile (`TS2352`) because `actionFn` is a private member of `Command` and the types don't structurally overlap, so I used `as unknown as { actionFn: ... }` with a fully-specified target type — no `any`.

src/commands/ailogic-providers-disable.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { Options } from "../options";
+
+export const command = new Command("ailogic:providers:disable <providerType>")
+  .description("disable a Gemini API provider service")
+  .option("-f, --force", "bypass confirmation prompt")
+  .before(requirePermissions, ["serviceusage.services.disable", "firebasevertexai.config.update"])
+  .action(async (providerType: string, options: Options) => {
+    const projectId = needProjectId(options);
+    if (providerType !== "gemini-developer-api" && providerType !== "agent-platform-gemini-api") {
```
paulb777: It would be more symmetric to use "gemini-developer-api" and "gemini-agent-platform-api"

src/commands/ailogic-providers-disable.ts
```diff
@@ -0,0 +1,40 @@
...
+      throw new FirebaseError(
+        `Invalid provider type: ${clc.bold(providerType)}. Must be 'gemini-developer-api' or 'agent-platform-gemini-api'.`,
+      );
+    }
+    if (options.nonInteractive && !options.force) {
+      throw new FirebaseError(
+        `Disabling provider ${clc.bold(providerType)} requires confirmation.\n\n` +
+          `To proceed in non-interactive mode, rerun with --force:\n\n` +
+          `  firebase ailogic:providers:disable ${providerType} --force`,
```
paulb777: Singular "provider" would be better, since this is an action on a single provider.

src/commands/ailogic-providers-disable.ts
christhompsongoogle: Re-use ailogic.ProviderType here instead of copying

src/commands/ailogic-providers-disable.ts
```diff
@@ -0,0 +1,40 @@
...
+  .option("-f, --force", "bypass confirmation prompt")
+  .before(requirePermissions, ["serviceusage.services.disable", "firebasevertexai.config.update"])
+  .action(async (providerType: string, options: Options) => {
+    const projectId = needProjectId(options);
+    if (providerType !== "gemini-developer-api" && providerType !== "agent-platform-gemini-api") {
+      throw new FirebaseError(
+        `Invalid provider type: ${clc.bold(providerType)}. Must be 'gemini-developer-api' or 'agent-platform-gemini-api'.`,
+      );
+    }
+    if (options.nonInteractive && !options.force) {
```
christhompsongoogle: This seems to disallow non-interactive without also adding  --force. Generally I think one or the other would be acceptable, but if this is particularly dangerous we could do both

src/commands/ailogic-providers-enable.ts
```diff
@@ -0,0 +1,23 @@
...
+import { FirebaseError } from "../error";
+
+import { Options } from "../options";
+
+export const command = new Command("ailogic:providers:enable <providerType>")
+  .description("enable a Gemini API provider service")
+  .before(requirePermissions, ["serviceusage.services.enable", "firebasevertexai.config.update"])
+  .action(async (providerType: string, options: Options) => {
+    const projectId = needProjectId(options);
+    if (providerType !== "gemini-developer-api" && providerType !== "agent-platform-gemini-api") {
```
christhompsongoogle: Re-use ailogic.ProviderType here

src/commands/help.ts
```diff
@@ -14,7 +15,46 @@ export const command = new Command("help [command]")
...
+          break;
+        }
+        const nextKey = Object.keys(current).find((k) => k.toLowerCase() === key.toLowerCase());
+        if (nextKey) {
+          current = current[nextKey];
+        } else {
+          matched = false;
+          break;
+        }
+      }
```
christhompsongoogle, line 40: This block is a little vague in its intent. Needs a comment to describe what the goal is here. The block below this too

src/commands/index.ts
```diff
@@ -224,6 +224,12 @@ export function load(client: CLIClient): CLIClient {
       client.apphosting.rollouts.list = loadCommand("apphosting-rollouts-list");
     }
   }
+  client.ailogic = {};
+  client.ailogic.providers = {};
+  client.ailogic.providers.enable = loadCommand("ailogic-providers-enable");
```
christhompsongoogle: Will need to add an experiment flag here (see "internaltesting" above as an example) until this is API council approved, since the API could change in the near term

src/gcp/ailogic.ts
```diff
@@ -191,10 +201,209 @@ export async function upsertBlockingFunction(endpoint: AILogicEndpoint): Promise
...
+      prefix,
+      true,
+    );
+    if (!isDeveloperEnabled) {
+      await serviceUsage.disableServiceAndPoll(
+        projectId,
+        "firebasevertexai.googleapis.com",
+        prefix,
+      );
+      ensureApiEnabled.uncacheEnabledAPI(projectId, "firebasevertexai.googleapis.com");
```
christhompsongoogle: can we call this inside the disableServiceAndPoll rather than calling it afterwards each time? Similarly to the other 4 callsites

src/gcp/serviceusage.ts
```diff
@@ -70,3 +70,45 @@ export async function generateServiceIdentityAndPoll(
...
+  }
+}
+
+/**
+ * Calls disableService and polls till the operation is complete.
+ */
+export async function disableServiceAndPoll(
+  projectId: string,
+  service: string,
+  prefix: string,
```
christhompsongoogle: rename to logging prefix

src/gcp/ailogic.ts
```diff
@@ -191,10 +201,209 @@ export async function upsertBlockingFunction(endpoint: AILogicEndpoint): Promise
...
+}
+
+/**
+ * Disables a Gemini API provider service.
+ */
+export async function disableProvider(
+  projectId: string,
+  providerType: ProviderType,
+): Promise<void> {
+  const prefix = "ailogic";
```
christhompsongoogle: nit: rename AILOGIC_LOGGING_PREFIX and move to top of the class

src/commands/ailogic-providers-disable.ts
marb2000: Done in dc501fea. Renamed `agent-platform-gemini-api` to `gemini-agent-platform-api` everywhere, so both provider ids share the `gemini-` prefix. It's a CLI-facing value only (never sent to a backend), so the rename is safe.

src/commands/ailogic-providers-disable.ts
marb2000: I kept the namespace plural (`ailogic:providers:*`) to stay consistent with the existing CLI conventions like `appdistribution:groups`, `firestore:databases`, and `database:instances`, and because `providers:list` acts on all of them. Happy to switch if you feel strongly.

src/commands/ailogic-providers-disable.ts
marb2000: Done in dc501fea. Added `ailogic.parseProviderType()` (plus `isProviderType()` and `PROVIDER_TYPES`) in `gcp/ailogic.ts`, and both commands now call it instead of copying the union.

src/commands/ailogic-providers-disable.ts
marb2000: Removed the bespoke guard in dc501fea. `confirm()` already aborts in non-interactive mode unless `--force` is passed, so the behavior is now handled by the standard path with no duplicated check.

src/commands/ailogic-providers-enable.ts
marb2000: Done in dc501fea, same `ailogic.parseProviderType()` helper.

src/commands/help.ts
marb2000: Done in dc501fea. Added comments to both blocks: the first walks the nested client command tree to resolve a namespace (for example `ailogic:providers`), and the second lists every registered command under that prefix.

src/commands/index.ts
marb2000: Done in dc501fea. Added an `ailogic` experiment and gated registration behind `experiments.isEnabled("ailogic")`, modeled on `internaltesting`. Enable it with `FIREBASE_CLI_EXPERIMENTS=ailogic` until the API is council-approved.

src/gcp/ailogic.ts
marb2000: Done in dc501fea. Moved `uncacheEnabledAPI` into `serviceusage.disableServiceAndPoll` and removed the per-callsite calls (ailogic is its only caller). Added a test in `serviceusage.spec.ts` covering it.

src/gcp/serviceusage.ts
marb2000: Done in dc501fea. Renamed the param to `loggingPrefix`.

src/gcp/ailogic.ts
marb2000: Done in dc501fea. Added an exported `AILOGIC_LOGGING_PREFIX` constant at the top of the module and replaced the repeated `"ailogic"` literals with it.

christhompsongoogle (approved):

marb2000:
While working on the config PR I found an inconsistency bug in the provider commands and fixed it here (commit 5187f7a4).

Enabling a provider turns on both the provider's service API and the Firebase AI Logic API (firebasevertexai.googleapis.com), but providers:list derived provider status purely from the provider service APIs. So a project with generativelanguage.googleapis.com enabled but the AI Logic API off (enabled outside the CLI, or a partial enable failure) would report the provider as enabled while the AI Logic commands report AI Logic as disabled on the same project.

Two changes:
- listProviders now also requires the AI Logic API to be enabled, otherwise it reports no providers.
- enableProvider enables the AI Logic API first, so a partial failure can no longer land in that inconsistent state.

Added tests for both.

marb2000:
Closing in favor of #10849, which moves this to a branch on the main repo as @joehan requested. The branch is identical (all review feedback addressed, plus the provider-state consistency fix). @christhompsongoogle FYI. Review approvals unfortunately do not carry over, so a fresh stamp on #10849 would be appreciated.

PR #10779, opened 2026-07-08
fix(security): default service account sentinel and partial deploy validation

inlined:
### Description
This PR resolves the deployment crash when opting out of declarative security and refines unenrollment safety:
1. Replaces the "default" string sentinel with null to revert to the default service account when unenrolling.
2. Prohibits partial codebase deployments (filtered deploys) when enrolling or unenrolling from declarative security to prevent inconsistent states.
3. Enables unenrollment prompts and cleanup for codebase-level filtered deploys.

### Scenarios Tested
- Ran mocha unit tests: npm run mocha:fast

### DIFF
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -102,7 +102,8 @@ export async function discoverSecurityDetails(
 
   if (isPartiallyFiltered && (isEnrolling || isUnenrolling)) {
     throw new FirebaseError(
-      `Cannot deploy a partial codebase when enrolling or unenrolling from declarative security`,
+      "To ensure a whole codebase is migrated cleanly, you may not deploy only part of a " +
+      "codebase when opting into or out of declartive secuirty (starting or no longer using `requireRoles`)"
     );
   }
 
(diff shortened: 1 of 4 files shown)


src/deploy/functions/prepare.ts
```diff
@@ -90,6 +91,21 @@ export async function discoverSecurityDetails(
...
+    filters.some(
+      (f) => (!f.codebase || f.codebase === codebase) && f.idChunks && f.idChunks.length > 0,
+    )
+  );
+  const isEnrolling = !!requiredRoles && !existingManagedSA;
+  const isUnenrolling = !requiredRoles && !!existingManagedSA && !!haveRolesEtag;
+
+  if (isPartiallyFiltered && (isEnrolling || isUnenrolling)) {
+    throw new FirebaseError(
+      `Cannot deploy a partial codebase when enrolling or unenrolling from declarative security`,
```
Berlioz: Should the message explain why, to prevent this from feeling frustratingly arbitrary?

Berlioz (approved):

src/deploy/functions/release/planner.ts
ajperel, line 186: if we care about filters here why don't we also care about filters for enrollment in this instance? I thought you didn't want to do enrollment or unenrollment on partial deploys?

ajperel, line 189: (1) Do we need this check at all? Given you fail for partial deployments in prepare.ts shouldn't we know it's safe to unenroll if we get to this point?

(2) If we do need it (or just want it for redundant safety), don't we need a similar check for _partial_ filters where we look at what codebase we're unenrolling and make sure the filters apply to that one, e.g, isFiltered here should match isPartiallyFiltered with the codebase check in prepare.ts? Otherwise can't a filter in an unrelated code base cause us to not remove the SA for a managed and fully deployed codebase.

ajperel, line 185: Would it be safe to remove roles on a partial deployment?

ajperel (commented):
Leaving the substantive review to Victor, but asking some questions just so I can learn as y'all add all this functionality.

src/deploy/functions/release/planner.ts
inlined: Certainly much more safe. Should we block it all?

inlined: Both are blocked in prepare.ts.

inlined: isFiltered is a check for partial filters (not full codebase filters). That's what the idChunks is.

src/deploy/functions/prepare.ts
inlined: IDK. I don't want an essay; maybe "because it is unsafe?"

ajperel (approved):

src/deploy/functions/release/planner.ts
ajperel: I'm OK with this. Worst case if something is removed on a partial deploy that shouldn't a workaround to fix would be a full deploy. This seems like something we can address if we get feedback about it.

ajperel: yeah, that's why I was confused by any checking

PR #10782, opened 2026-07-09
feat: prompt user before executing lifecycle hooks when recovering from partially failed deployments

wandamora:
### Description

This PR introduces recovery detection for partial deployment failures and interactive user prompting:
- **`isRecoveredFromPartialDeploy` helper**: Evaluates whether a successful deploy is recovering from a previous partial failure by checking if `haveBackend` contains functions with a matching bundle hash (or in a `FAILED` state) while `wantBackend` introduces net new functions (excluding `FAILED` resources in `haveBackend`).
- **Interactive Lifecycle Prompting**: When `isRecoveredFromPartialDeploy` returns `true`, the user is prompted to clarify which lifecycle hook (`afterFirstDeploy`, `afterRedeploy`, or `skip`) should be executed.

### Scenarios Tested
- Unit tests

### DIFF
--- a/src/deploy/functions/prompts.ts
+++ b/src/deploy/functions/prompts.ts
@@ -406,20 +406,19 @@ export async function promptForLifecycleEvent(
 ): Promise<DeploymentEvent | undefined> {
   const hooks = wantBackend.lifecycleHooks || {};
   const choices: { name: string; value: DeploymentEvent | "skip" }[] = [];
-  if (hooks.afterFirstDeploy) {
-    choices.push({ name: "afterFirstDeploy", value: "afterFirstDeploy" });
-  }
-  if (hooks.afterRedeploy) {
-    choices.push({ name: "afterRedeploy", value: "afterRedeploy" });
-  }
-  choices.push({ name: "skip (default)", value: "skip" });
 
-  if (choices.length === 1) {
+  for (const hookName of Object.keys(hooks)) {
+    choices.push({ name: hookName, value: hookName as DeploymentEvent });
+  }
+
+  if (choices.length === 0) {
     return undefined;
   }
 
+  choices.push({ name: "skip (default)", value: "skip" });
+
   const selection = await select<DeploymentEvent | "skip">({
-    message: `We detected a partially failed deployment for codebase "${codebase}". Which lifecycle hook would you like to run?`,
+    message: `We cannot determine whether this deployment is a first deploy or a redeploy for codebase "${codebase}" because it is recovering from a previous partial failure.`,
     choices,
     default: "skip",
     nonInteractive: options?.nonInteractive,
--- a/src/deploy/functions/prompts.ts
+++ b/src/deploy/functions/prompts.ts
@@ -398,6 +398,7 @@ export async function promptForSecurityChanges(
 
 /**
  * Prompts the user to select which lifecycle hook to execute when a partial deployment retry is detected.
+ * If no hooks are defined, then the prompt will be skipped.
  */
 export async function promptForLifecycleEvent(
   codebase: string,
--- a/src/deploy/functions/prompts.ts
+++ b/src/deploy/functions/prompts.ts
@@ -398,7 +398,7 @@ export async function promptForSecurityChanges(
 
 /**
  * Prompts the user to select which lifecycle hook to execute when a partial deployment retry is detected.
- * If no hooks are defined, then the prompt will be skipped.
+ * If no lifecycle hooks are defined in wantBackend, the prompt is skipped and returns undefined.
  */
 export async function promptForLifecycleEvent(
   codebase: string,
--- a/src/deploy/functions/prompts.ts
+++ b/src/deploy/functions/prompts.ts
@@ -2,7 +2,7 @@ import * as clc from "colorette";
 
 import { getFunctionLabel } from "./functionsDeployHelper";
 import { FirebaseError } from "../../error";
-import { confirm, number } from "../../prompt";
+import { confirm, number, select } from "../../prompt";
 import { logger } from "../../logger";
 import * as backend from "./backend";
 import * as pricing from "./pricing";
@@ -11,6 +11,7 @@ import * as artifacts from "../../functions/artifacts";
 import { Options } from "../../options";
 import { EndpointUpdate } from "./release/planner";
 import * as iam from "../../gcp/iam";
+import { DeploymentEvent } from "./release/lifecycle";
 
 export interface ActiveSecurityPlan {
   managedServiceAccount: string;
@@ -393,4 +394,36 @@ export async function promptForSecurityChanges(
       }
     }
   }
+}
+
+/**
+ * Prompts the user to select which lifecycle hook to execute when a partial deployment retry is detected.
+ */
+export async function promptForLifecycleEvent(
+  codebase: string,
+  wantBackend: backend.Backend,
+  options?: Options,
+): Promise<DeploymentEvent | undefined> {
+  const hooks = wantBackend.lifecycleHooks || {};
+  const choices: { name: string; value: DeploymentEvent | "skip" }[] = [];
+  if (hooks.afterFirstDeploy) {
+    choices.push({ name: "afterFirstDeploy", value: "afterFirstDeploy" });
+  }
+  if (hooks.afterRedeploy) {
+    choices.push({ name: "afterRedeploy", value: "afterRedeploy" });
+  }
+  choices.push({ name: "skip (default)", value: "skip" });
+
+  if (choices.length === 1) {
+    return undefined;
+  }
+
+  const selection = await select<DeploymentEvent | "skip">({
+    message: `We detected a partially failed deployment for codebase "${codebase}". Which lifecycle hook would you like to run?`,
+    choices,
+    default: "skip",
+    nonInteractive: options?.nonInteractive,
+  });
+
+  return selection === "skip" ? undefined : selection;
 }
--- a/src/deploy/functions/release/index.ts
+++ b/src/deploy/functions/release/index.ts
@@ -124,13 +124,9 @@ export async function release(
   const wantBackend = backend.merge(...Object.values(payload.functions).map((p) => p.wantBackend));
   printTriggerUrls(wantBackend, projectNumber);
 
-  const allErrors = summary.results.filter((r) => r.error).map((r) => r.error) as Error[];
-  const isPartialFailure = allErrors.length > 0 && allErrors.length < summary.results.length;
-
-  if (allErrors.length === 0 || isPartialFailure) {
-    for (const [codebase, { wantBackend: w, haveBackend: h }] of Object.entries(
-      payload.functions,
-    )) {
+  for (const [codebase, { wantBackend: w, haveBackend: h }] of Object.entries(payload.functions)) {
+    const { errors, isPartialFailure } = getCodebaseDeployStats(codebase, w, h, summary);
+    if (errors.length === 0 || isPartialFailure) {
       await executeLifecycleHooks(w, h, plan, codebase, options, isPartialFailure);
     }
   }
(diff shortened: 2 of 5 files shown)


src/deploy/functions/release/lifecycle.ts
```diff
@@ -32,8 +41,27 @@ export async function executeLifecycleHooks(
...
+  let event: DeploymentEvent | undefined;
+  if (isPartialFailure) {
+    event = await prompts.promptForLifecycleEvent(codebase ?? "default", wantBackend, options);
+    if (!event) {
+      logLabeledBullet(
+        "functions",
+        `Skipping lifecycle hooks for codebase "${codebase ?? "default"}". You can retry a lifecycle hook in isolation by running: firebase functions:lifecycle:run <event> ${codebase ?? "default"}`,
+      );
+      return false;
+    }
```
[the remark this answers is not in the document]
wandamora: I decided to remove the command suggestion since its more intended for failed hooks.

src/deploy/functions/release/lifecycle.spec.ts
```diff
@@ -257,5 +283,100 @@ describe("lifecycle", () => {
...
+            function: "installHookTask",
+          },
+        },
+      };
+      const haveBackend = backend.empty();
+
+      const executed = await executeLifecycleHooks(
+        wantBackend,
+        haveBackend,
+        undefined,
```
inlined: please leave comments when using non-obvious literal values what the parameter is (e.g. undefined, true)

inlined (approved):

src/deploy/functions/prompts.ts
```diff
@@ -394,3 +395,35 @@ export async function promptForSecurityChanges(
     }
   }
 }
+
+/**
+ * Prompts the user to select which lifecycle hook to execute when a partial deployment retry is detected.
```
ajperel, line 400: Could document that if called with a backend that doesn't have lifecycle hooks will default to skip without prompting.

src/deploy/functions/prompts.ts
```diff
@@ -394,3 +395,35 @@ export async function promptForSecurityChanges(
...
+): Promise<DeploymentEvent | undefined> {
+  const hooks = wantBackend.lifecycleHooks || {};
+  const choices: { name: string; value: DeploymentEvent | "skip" }[] = [];
+  if (hooks.afterFirstDeploy) {
+    choices.push({ name: "afterFirstDeploy", value: "afterFirstDeploy" });
+  }
+  if (hooks.afterRedeploy) {
+    choices.push({ name: "afterRedeploy", value: "afterRedeploy" });
+  }
+  choices.push({ name: "skip (default)", value: "skip" });
```
ajperel: I think you could maybe future proof this to adding more hook types with something like


```
  const hooks = wantBackend.lifecycleHooks || {};
  const choices: { name: string; value: DeploymentEvent | "skip" }[] = [];

  for (const hookName of Object.keys(hooks)) {
    choices.push({ name: hookName, value: hookName as DeploymentEvent });
  }

  choices.push({ name: "skip (default)", value: "skip" });
```

src/deploy/functions/release/lifecycle.ts
```diff
@@ -32,8 +41,27 @@ export async function executeLifecycleHooks(
...
   codebase?: string,
+  options?: Options,
+  isPartialFailure?: boolean,
 ): Promise<boolean> {
-  const event = determineDeploymentEvent(haveBackend);
+  if (!hasLifecycleHooks(wantBackend)) {
+    return false;
+  }
+
+  let event: DeploymentEvent | undefined;
```
ajperel, line 91: It is my fault this is called `DeploymentEvent` but we should rename it `LifecycleEvent` ... we already use Lifecycle all over the Functions SDK and you even called the method below `promptForLifecycleEvent` and this file is `lifecycle.ts`

While we are again carrying over something from Extensions... we just haven't talked about lifecycles in any form in Functions that I know of so re-using this term seems fine.

I think this is all internal and can be a separate non-urgent PR though. I can do it since it's my fault once I won't be messing up a PR you have in flight.

src/deploy/functions/release/index.ts
```diff
@@ -136,14 +136,10 @@ export async function release(
...
 
-  const allErrors = summary.results.filter((r) => r.error).map((r) => r.error) as Error[];
-
-  // Only execute lifecycle hooks when all deployments are successful.
-  if (allErrors.length === 0) {
-    for (const [codebase, { wantBackend: w, haveBackend: h }] of Object.entries(
-      payload.functions,
-    )) {
-      await executeLifecycleHooks(w, h, plan, codebase);
+  for (const [codebase, { wantBackend: w, haveBackend: h }] of Object.entries(payload.functions)) {
```
ajperel, line 127: Not for this PR... but I wonder if this should move inside `Fabricator.applyPlan` it feels like we've designed lifecycle events to be part of the deployment plan in spirit if not in practice.

ajperel (commented):
I don't think we need this. It's not _bad_, but the previous behavior was instead of prompting to just say "there was a partial deployment so we didn't run lifecycle hook X, if you want to run this command Y" right?  I feel like that's probably good enough.

The hard part with this is what's safe/correct probably depends on implementation details of the functions that users might not know. But I'm assuming that the on average most correct action sequence would be:

- Time A: Partial deploy 
- Time B (immediately after): Run no lifecycle hooks because partial deployment
- Time C: Next successfully deploy
- Time D (immediately after): Run lifecycle hooks that should have been run for the deployment at Time A

And would optimize for that and save users some prompts. But I'm open to arguments we want both. It's not too hard to change this later (unless removing it is considered a breaking change)

src/deploy/functions/release/lifecycle.ts
wandamora: I also realize I don't stay consistent `LifecycleEvent` and `LifecycleHook` either, sigh. Since the manifest reads it as `LifecycleHook`, maybe we should double down on the latter.

wandamora:
I updated this PR to instead only prompt users if we think they're trying to recover from a previous partial deployment. PTAL, thanks!

src/deploy/functions/release/lifecycle.ts
ajperel: I _think_ that distinction may make sense. Hook is the thing that gets called when the event happens? So events trigger hooks?

wandamora: Yeah, events trigger hooks. I think in conversation we interchange the two names a lot which made me think it should all be the same.

src/deploy/functions/release/lifecycle.ts
```diff
@@ -15,14 +17,59 @@ export type DeploymentEvent = "afterFirstDeploy" | "afterRedeploy";
...
+ */
+export function isRecoveredFromPartialDeploy(
+  wantBackend: backend.Backend,
+  haveBackend: backend.Backend,
+): boolean {
+  const wantEndpoints = backend.allEndpoints(wantBackend);
+  const haveEndpoints = backend.allEndpoints(haveBackend);
+
+  const wantHashes = new Set(wantEndpoints.map((ep) => ep.hash).filter((h): h is string => !!h));
+  if (!wantHashes.size) {
```
ajperel, line 48: I think it's worth writing a comment on why returning false in this case. and maybe updating JSDoc too. I have spent way too much time thinking about this.

I think... in a truly accurate sense... if haveEndpoints > 0 and wantEndpoints = 0 then this _could_ in the strictest sense be a recovery deployment if the previous deployment was trying to delete all functions and that failed. But in that case there is no wantEndpoint hash recorded, and if there were, it obviously wouldn't match the haveEndpoint hash since the whole point is we failed to delete it.

I don't think we can detect redeploy to correctly delete functions right? In which case we should document that and note we an only say concretely it's a recoveryDeployment when it was a failure to _create_ functions.... now I'm going into edge cases in my head where maybe we _can_ detect that.... but really those will never be the _afterFirstDeploy_ case so I don't think they matter for our purposes. Still I think we should document the limitations but not worry about trying to fix any of them.

src/deploy/functions/release/lifecycle.ts
```diff
@@ -15,14 +17,59 @@ export type DeploymentEvent = "afterFirstDeploy" | "afterRedeploy";
...
+  }
+
+  // 1. We know a previous deploy was a partial success if haveBackend includes the same hash
+  // but wantBackend includes different functions.
+  const hasSameHash = haveEndpoints.some((ep) => ep.hash && wantHashes.has(ep.hash));
+  if (!hasSameHash) {
+    return false;
+  }
+
+  // 2. If there are no net new functions, we know this was an update.
```
ajperel: If there are no net few functions and the hashes are the same then the .env file and secrets are identical too. In this case... is anything actually being deployed? I would think in this case deploy does nothing because there's no update to be done. But maybe there's some other state that could be deploying that I'm not thinking of?

It's sort of an aside from this flow (we don't need to prompt user in this case) but... maybe we shouldn't be running lifecycle events at all in that case?

src/deploy/functions/release/lifecycle.ts
```diff
@@ -15,14 +17,59 @@ export type DeploymentEvent = "afterFirstDeploy" | "afterRedeploy";
...
+ */
+export function hasLifecycleHooks(backendSpec: backend.Backend): boolean {
+  return !!(backendSpec.lifecycleHooks && Object.keys(backendSpec.lifecycleHooks).length > 0);
+}
+
+/**
+ * Detects if the codebase is recovering from a previous partially successful deploy
+ * that requires prompting the user to select the appropriate lifecycle event.
+ */
+export function isRecoveredFromPartialDeploy(
```
ajperel: This method name isn't quite correct because we don't know from the wantBackend state if we had a successful deployment of wantBackend. We're only recovered if the second deploy fully succeeded.
 
It's more like

isRecoveryDeployment()

and we've recovered if it's succesful

src/deploy/functions/release/lifecycle.ts
```diff
@@ -15,14 +17,59 @@ export type DeploymentEvent = "afterFirstDeploy" | "afterRedeploy";
...
 
+/**
+ * Checks if the backend specification has any lifecycle hooks configured.
+ */
+export function hasLifecycleHooks(backendSpec: backend.Backend): boolean {
+  return !!(backendSpec.lifecycleHooks && Object.keys(backendSpec.lifecycleHooks).length > 0);
+}
+
+/**
+ * Detects if the codebase is recovering from a previous partially successful deploy
```
ajperel: (1) I'd maybe define how we determine this since there are cases that it breaks down. If you made _any_ code modification after deployment, even a comment or something, it's going to change the hash even though from a logical perspective you could still be "recovering" on your next deploy. 

(2) I'd omit "that requires prompting the user to select the appropriate lifecycle event"... while true it's not really relevant _here_. It's why we wrote this method, but we could use it for other reasons in the future. As long as that's clear where we are _using_ it for that purpose I think we're good.

Combining with my naming comment below (since we don't actually know if this want deployment has or will succeed either...)

"Detects if this is deployment is a redeploy of a partially successful but identical previous deployment. This will be true only if the hashes for both backends are the same but the specified endpoint are different."

src/deploy/functions/release/lifecycle.ts
```diff
@@ -15,14 +17,59 @@ export type DeploymentEvent = "afterFirstDeploy" | "afterRedeploy";
...
+
+  // 1. We know a previous deploy was a partial success if haveBackend includes the same hash
+  // but wantBackend includes different functions.
+  const hasSameHash = haveEndpoints.some((ep) => ep.hash && wantHashes.has(ep.hash));
+  if (!hasSameHash) {
+    return false;
+  }
+
+  // 2. If there are no net new functions, we know this was an update.
+  // If there are net new functions we know this might be a first deploy (it could also be
```
ajperel: This comment is confusing me. I think even in this context of "first deploy" as being literally "nothing is deployed yet" and so there shouldn't be a haveBackend/endpoint at all. I think you can rewrite it to be a bit more understandable.

If we have "have" state and hashes are the same but there will be net new functions deployed, then we know this isn't a first deployment and that the current deployment is trying to redo (with the same spec) a previous deployment that didn't deploy everything successfully, and thus in this case we are *attempting* to recover from a previous deploy rather than a modified codebase (which is ambiguous).

ajperel (approved):
I think the logic is sound and the naming and comments could be improved to reduce confusion and be a bit more accurate.

src/deploy/functions/release/lifecycle.ts
wandamora: Rewrote the inline comment to clarify that matching hashes with net new endpoints indicates an attempted redeploy of an identical specification.

src/deploy/functions/release/lifecycle.ts
wandamora: If there are absolutely no changes, the hook will be skipped. See http://github.com/firebase/firebase-tools/blob/0b637cd2ff4b73362a2c6325c5f9dba01d608e43/src/deploy/functions/release/lifecycle.ts#L45-L61

PR #10783, opened 2026-07-09
Integrate .env secret bindings with param prompting

Berlioz:
This PR centers around changes made to the handleSecret()/now ensureSecret() method invoked during param resolution. Formerly, handleSecret() only made sure that a Cloud Secret with a resource name matching the name of the Secret Param existed with a readable latest version.

Now:
1) It can use an alternative resource name and version identifier from the Secret Param, which has been merged into the Param object earlier in deployment by applyEnvSecretBindings().
2) It emits a string representing the secret binding which should be written to .env files. These strings are propagated up the call chain to prepare.ts, which has them inserted to .env files only if not present using the same mechanisms that currently write other params.
3) It logs an INFO-level notification in the case where the backing Cloud Secret already existed but no matching record in .env files existed, in which case we are basically "importing" a secret that already existed. Open to wording suggestions here, especially if we should specifically advise users to commit to version control.

Intended behavior after merging this PR on deploy:
- Secret Param exists, .env secret exists: applyEnvSecretBindings seamlessly overrides the Secret to look in the place specified by .env, which satisfies ensureSecret().
- Secret Param missing, .env secret exists: no-op; since defineSecret() was never called during discovery, no corresponding record exists in build.SecretEnvVars and applyEnvSecretBindings skips the .env secret.
- Secret Param exists, .env secret missing: ensureSecret() checks to see whether a Secret resource exists at the "default" secret location. If so, it logs a warning and writes that default secret to .envs. Otherwise, it helps the user create a secret and upon success writes it to .env files.

### DIFF
--- a/src/deploy/functions/build.ts
+++ b/src/deploy/functions/build.ts
@@ -819,11 +819,11 @@ export function applyEnvSecretBindings(
     const secretRef = envSecrets[key];
     const { projectId, secretId, version } = secretRef;
 
-    for (const secret of build.params.filter((param) => param.type === "secret")) {
-      if (secret.name === key) {
-        secret.resourceId = secretId;
-        secret.version = version;
-        secret.inLocalEnvironment = true;
+    for (const secretParam of build.params.filter((param) => param.type === "secret")) {
+      if (secretParam.name.toUpperCase() === key) {
+        secretParam.resourceId = secretId;
+        secretParam.version = version;
+        secretParam.inLocalEnvironment = true;
       }
     }
 
--- a/src/deploy/functions/build.ts
+++ b/src/deploy/functions/build.ts
@@ -347,7 +347,7 @@ interface ResolveBackendOpts {
  */
 export async function resolveBackend(
   opts: ResolveBackendOpts,
-): Promise<{ backend: backend.Backend; envs: Record<string, params.ParamValue> }> {
+): Promise<{ backend: backend.Backend; envs: Record<string, params.ParamValue>, newSecretRefs: Record<string, string> }> {
   const paramValues = await params.resolveParams(
     opts.build.params,
     opts.firebaseConfig,
@@ -356,7 +356,7 @@ export async function resolveBackend(
     opts.isEmulator,
   );
 
-  return { backend: toBackend(opts.build, paramValues), envs: paramValues };
+  return { backend: toBackend(opts.build, paramValues), envs: paramValues, newSecretRefs: {} };
 }
 
 /**
@@ -806,6 +806,14 @@ export function applyEnvSecretBindings(build: Build, envSecrets: Record<string, 
     const secretRef = envSecrets[key];
     const { projectId, secretId, version } = secretRef;
     const secretPinsVersion = typeof secretRef.version !== "undefined";
+
+    for (const secret of build.params.filter(param => param.type === "secret")) {
+      if (secret.name.toLowerCase() === key.toLowerCase()) {
+        secret.resourceId = secretId;
+        secret.version = version;
+        secret.inLocalEnvironment = true;
+      }
+    }
 
     for (const endpointName of Object.keys(build.endpoints)) {
       const endpoint = build.endpoints[endpointName];
--- a/src/deploy/functions/build.ts
+++ b/src/deploy/functions/build.ts
@@ -341,6 +341,7 @@ interface ResolveBackendOpts {
   userEnvs: Record<string, string>;
   nonInteractive?: boolean;
   isEmulator?: boolean;
+  force?: boolean;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


src/deploy/functions/build.ts
```diff
@@ -817,6 +819,14 @@ export function applyEnvSecretBindings(
     const secretRef = envSecrets[key];
     const { projectId, secretId, version } = secretRef;
 
+    for (const secret of build.params.filter((param) => param.type === "secret")) {
+      if (secret.name === key) {
+        secret.resourceId = secretId;
+        secret.version = version;
+        secret.inLocalEnvironment = true;
+      }
+    }
```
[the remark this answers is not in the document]
Berlioz: Is that...true? I mean the case thing definitely is true, fixed.

src/deploy/functions/params.ts
```diff
@@ -509,7 +538,21 @@ async function handleSecret(
         secretParam.label || secretParam.name
       } is in illegal state ${metadata.secretVersion.state}`,
     );
+  } else {
+    secretAlreadyExisted = true;
   }
+
+  const secretRefString =
+    typeof version === "undefined" ? "resourceId" : `${resourceId}:${version}`;
```
inlined: I don't think you mean resourceId to be quoted here

src/deploy/functions/params.ts
```diff
@@ -458,42 +477,52 @@ function populateDefaultParams(config: FirebaseConfig): Record<string, ParamValu
...
-          `\tfirebase functions:secrets:set ${secretParam.name}${secretParam.format === "json" ? " --format=json --data-file <file.json>" : ""}`,
+          `\tfirebase functions:secrets:set ${resourceId}${secretParam.format === "json" ? " --format=json --data-file <file.json>" : ""}`,
       );
     }
-    const promptMessage = `This secret will be stored in Cloud Secret Manager (https://cloud.google.com/secret-manager/pricing) as ${
-      secretParam.name
-    }. Enter ${secretParam.format === "json" ? "a JSON value" : "a value"} for ${
+    if (experiments.isEnabled("secretEnvParams")) {
+      // TODO: Move the explanation and link to Cloud Secret Manager in the next prompt here once this makes it out of experimental.
+      resourceId = await input({
```
inlined: 1. Did we want to default to the existing behavior if the secret doesn't already exist (YOLO secret = key)?
2. I thought we wanted to confirm loop before reusing a secret
3. If reusing a secret we shouldn't needt o create & add a version.

src/deploy/functions/params.ts
```diff
@@ -509,7 +538,21 @@ async function handleSecret(
         secretParam.label || secretParam.name
       } is in illegal state ${metadata.secretVersion.state}`,
     );
+  } else {
+    secretAlreadyExisted = true;
```
inlined, line 551: Am I reading diffs wrong? It looks like this is set if the experiment is false?

src/functions/env.ts
```diff
@@ -450,3 +446,26 @@ export function writeResolvedParams(
...
+  for (const secretName of Object.keys(resolvedSecretRefs)) {
+    const uppercaseName = secretName.toUpperCase();
+    const resolvedRef = resolvedSecretRefs[uppercaseName];
+    if (!Object.prototype.hasOwnProperty.call(haveSecretRefs, uppercaseName)) {
+      const reservedKey = "FIREBASE_SECRET_REF_" + uppercaseName;
+      toWrite[reservedKey] = resolvedRef;
+    }
+  }
+
+  writeUserEnvs(toWrite, userEnvOpt);
```
inlined, line 470: is it safe to call writeUserEnvs repeatedly in different places? Does this append? If so, maybe wee should rename it appendUserEnvs

src/deploy/functions/params.ts
Berlioz: Staring at my monitor for a bit and trusting vs code's brace matching, I'm pretty sure this is the fall-through for the big "what's the state of the referenced secret" if block that begins `if (!metadata.secret)`

src/functions/env.ts
Berlioz: It appends yeah. I did want to maintain some sense that this is a filesystem touch when skimming the code though, maybe writeNewUserEnvs()?

src/deploy/functions/params.ts
Berlioz: e:
1) I don't think so--we talked it over in chat and this would break flows where the user had multiple environments deployed in the same project. Added --force support though.
2) I thought so too but I was working out the logic in my head and right now, if the secret already exists and was created manually or with secrets:set or whatever, we would just pull it in without asking, right? This doesn't actually change the current behavior of secrets?
3) Ack. Recursing like this seems like the most elegant way of handling the input-a-reused-secret case.

src/deploy/functions/params.ts
Berlioz: sad trombone noises

inlined (approved):

PR #10785, opened 2026-07-09
update vscode release to build first before upload

hlshen:


.github/workflows/release-vscode-ext.yaml
```diff
@@ -21,30 +21,86 @@ jobs:
...
-          if gh api -X PATCH "/repos/${{ github.repository }}/git/refs/heads/$TARGET_BRANCH" \
+      - name: Checkout repository
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          ref: ${{ env.SHA }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@1e60f620b9541d16bece96c5465dc8ee9832be0b # v4.0.3
+        with:
+          node-version: 21
```
joehan, line 50: Why are we using 21? That is not a supported version anymore (and odd versions are essentially betas). Can you use 24 instead?

.github/workflows/release-vscode-ext.yaml
```diff
@@ -21,30 +21,86 @@ jobs:
...
+
+      - name: Write skip publish flag (if build-only)
+        if: env.BUILD_ONLY == 'true'
+        run: |
+          echo "true" > firebase-vscode/skip_publish.txt
+
+      - name: Authenticate to Google Cloud
+        uses: google-github-actions/auth@71fee32a0bb7e97b4d33d548e7d957010649d8fa # v2.1.3
+        with:
+          credentials_json: ${{ secrets.FIREBASE_DEVELOPER_GCP_SA_KEY }}
```
joehan, line 75: New secret?

.github/workflows/release-vscode-ext.yaml
hlshen: Ha, I just defaulted to not 20. Thanks for the explanation

PR #10789, opened 2026-07-10
Display a helpful error instead of Unexpected Error for key validation

Berlioz:
Because KeyValidationError wasn't instanceof FirebaseError, the catchall error handler in errorOut.ts was just printing "Error: An unexpected error has occurred"


shettyvarun268 (approved):
We might need to take off public from the constructor as per what gemini says. But overall, LGTM

PR #10796, opened 2026-07-14
fix(auth): use 127.0.0.1 for OAuth loopback redirect instead of localhost

### LINKED ISSUE #10750: firebase login: use 127.0.0.1 (not localhost) for OAuth loopback + surface --no-localhost hint on hang
### Which command?

login

### Version info

firebase-tools 15.22.3, Node 20.x, Windows 11 (corp-managed / SSO)

### Platform

Windows

### CLI/emulator log

(design/behaviour report — no crash log; behaviour observed on multiple corp Windows setups)

### Reproduction steps

1. On a corp-managed Windows machine (SSO-gated Google account, hosts file locked down, dual-stack IPv6 enabled), run `firebase login`.
2. Complete Google SSO in browser. The success page never loads; CLI hangs on `Waiting for authentication...`.
3. Root cause: `src/auth.ts` uses `http://localhost:${port}` as the OAuth `redirect_uri` (see `getCallbackUrl`, line ~286), and `server.listen(port, ...)` binds to all interfaces without an explicit host. On dual-stack Windows the browser may resolve `localhost` to `::1` while the loopback listener has the port claimed on `0.0.0.0` — or a corp hosts-file entry has `localhost` pointing somewhere unusable. Either way the callback is silently dropped.

The workaround (`firebase login --no-localhost`) works but is undiscoverable — the docs and CLI output on hang don't hint at it.

### Expected behavior

Two low-risk, additive fixes:

1. **Use `127.0.0.1` for the OAuth loopback**, per [Google's OAuth 2.0 for Native Apps guidance](https://developers.google.com/identity/protocols/oauth2/native-app#request-parameter-redirect_uri) which explicitly recommends `http://127.0.0.1:<port>` over `http://localhost:<port>` to avoid resolver ambiguity. Bind the http server to `127.0.0.1` too, so IPv6-only or hosts-file-locked machines can't miss the callback.
2. **After ~30s of no callback in `loginWithLocalhost`**, print a hint pointing at `firebase login --no-localhost` and `firebase login:ci`. Today the CLI hangs with only `Waiting for authentication...` — corp users often reach for Ctrl-C and start googling.

Files: `src/auth.ts` (`getCallbackUrl`, `loginWithLocalhost`).

Related (superset scope, generic auth failure): #8304. This issue is narrower and code-level.

ATKasem:
## Description

`firebase login` builds the Google OAuth loopback redirect as `http://localhost:${port}` (`getCallbackUrl` in `src/auth.ts`).

On dual-stack hosts, the browser can resolve `localhost` to `::1` while the OAuth redirect never reaches the CLI, and login hangs on `Waiting for authentication...`.

This uses the `127.0.0.1` IP literal for the **Google** redirect URI so it matches [Google's OAuth 2.0 loopback requirement](https://developers.google.com/identity/protocols/oauth2/native-app#redirect-uri_loopback) (IP literal, not `localhost`).

The **GitHub** login flow stays on `localhost`, since GitHub validates `redirect_uri` against its registered OAuth callback URL.

The local callback server still binds with `server.listen(port)` (all interfaces), so both `127.0.0.1` (Google) and `localhost`/`::1` (GitHub) can reach it on dual-stack hosts.

Fixes #10750.

## Scenarios Tested

- Unit tests: `getCallbackUrl(port, "127.0.0.1")` → `http://127.0.0.1:${port}`, default host stays `localhost`, no port yields the OOB URN.
- `npm run mocha -- src/auth.spec.ts` passes.

## Sample Commands

`firebase login`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,3 @@
 - Add `MCP-Protocol-Version`, `Mcp-Method`, and `Mcp-Name` HTTP headers to `OneMcpServer` requests per the MCP 0728 standard release candidate (https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/ and https://modelcontextprotocol.io/seps/2243-http-standardization).
 - Fixes Storage Emulator to support JSON uploads larger than 100KB without hanging or throwing 413 error (#8355)
+- Fixed `firebase login` hanging on some dual-stack hosts by using the `127.0.0.1` IP literal for the OAuth loopback redirect instead of `localhost`. (#10750)
--- a/src/auth.spec.ts
+++ b/src/auth.spec.ts
@@ -5,6 +5,7 @@ import {
   assertAccount,
   getAdditionalAccounts,
   getAllAccounts,
+  getCallbackUrl,
   getGlobalDefaultAccount,
   getProjectDefaultAccount,
   selectAccount,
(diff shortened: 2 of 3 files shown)


ATKasem:
Addressed the Gemini review feedback:

- Google redirect stays on `127.0.0.1` (needed for Google's loopback IP-literal requirement + the dual-stack hang).
- GitHub redirect stays on `localhost` (registered callback host).
- Callback server is back to `server.listen(port)` with no host binding, so `localhost`/`::1` clients can still reach it.

PR #10807, opened 2026-07-16
fix package dependency causing package-lock inconsistencies

hlshen:
fix package dependency causing package-lock inconsistencies, which prevented npm CI from properly running in github actions + kokoro environments


Helbss (changes requested):
• >/2026UpgradeByHelbss(c)\>

yuchenshi (approved):

PR #10810, opened 2026-07-17
fix: consume stream response body via streamToString instead of res.response.text()

### LINKED ISSUE #10809: hosting: upload HTTP errors are masked by "Body is unusable" after stream response wrapping
### Environment

- firebase-tools 15.23.0
- The same code path is present in the current v15.24.0 source
- macOS arm64
- Reproduced with Node.js 24.18.0, 24.14.0, and 22.20.0

### Test case

This minimal local reproduction uses the public Firebase CLI package internals and does not contact Firebase:

```js
const http = require("http");
const { Client } = require("firebase-tools/lib/apiv2.js");

const server = http.createServer((_req, res) => {
  res.writeHead(400, { "content-type": "text/plain" });
  res.end("real upstream error");
});

server.listen(0, "127.0.0.1", async () => {
  try {
    const port = server.address().port;
    const client = new Client({
      urlPrefix: `http://127.0.0.1:${port}`,
      auth: false,
    });

    const result = await client.request({
      method: "GET",
      path: "/upload",
      resolveOnHTTPError: true,
      responseType: "stream",
    });

    try {
      await result.response.text();
    } catch (error) {
      console.log(error.message);
    }

    let body = "";
    for await (const chunk of result.body) body += chunk.toString();
    console.log(body);
  } finally {
    server.close();
  }
});
```

### Steps to reproduce

1. Install firebase-tools 15.23.0.
2. Run the test case on Node.js 24.18.0, 24.14.0, or 22.20.0.
3. Observe the first output line.

### Expected behavior

The Hosting uploader should surface or log the actual non-200 response body:

```
real upstream error
```

### Actual behavior

The uploader attempts to consume the original response after its body has already been wrapped as a Node stream and throws:

```
Body is unusable: Body has already been read
```

The real error remains available through `result.body`, but the uploader never reads it because it calls `res.response.text()`.

### Root cause

For `responseType: "stream"`, `apiv2` wraps the fetch body with `Readable.fromWeb(res.body)`. The Hosting uploader then handles a non-200 status by calling `res.response.text()`, which tries to consume the already-locked web response body a second time.

Current source:
- https://github.com/firebase/firebase-tools/blob/v15.24.0/src/apiv2.ts
- https://github.com/firebase/firebase-tools/blob/v15.24.0/src/deploy/hosting/uploader.ts#L247-L254

This is not Node.js 24-specific: the focused test produces the same result on Node.js 22.20.0.

### Suggested fix

On a non-200 upload response, consume and decode `res.body` (the Node readable returned by `apiv2`) instead of calling `res.response.text()`. Alternatively, defer wrapping the response until after the status is checked.

Related broad error-reporting issue: #1053.

joehan:
Fixes #10809

### Root Cause
When native `fetch` is used with `responseType: "stream"`, `apiv2.ts` converts the Fetch API's `ReadableStream` (`res.body`) into a Node.js `Readable` stream using `Readable.fromWeb(res.body)`. Under Web Streams API semantics, this locks the underlying `ReadableStream`. Subsequent calls to `res.response.text()` on the native `Response` object throw `TypeError: Body is unusable: Body has already been read`.

### Fix
Updated handlers that consume stream responses to read from `res.body` via `streamToString(res.body)` rather than calling `res.response.text()`.

### Verification
- Tested with minimal reproduction script from #10809.
- All unit tests (`npm test`) passed.
- Linter (`npm run lint`) passed cleanly.

### DIFF
--- a/src/commands/database-get.ts
+++ b/src/commands/database-get.ts
@@ -118,7 +118,7 @@ export const command = new Command("database:get <path>")
 
     if (res.status >= 400) {
       // TODO(bkendall): consider moving stream-handling logic to responseToError.
-      const r = await utils.streamToString(res.body);
+      const r = res.body ? await utils.streamToString(res.body) : "";
       let d;
       try {
         d = JSON.parse(r);
--- a/src/commands/database-get.ts
+++ b/src/commands/database-get.ts
@@ -118,7 +118,7 @@ export const command = new Command("database:get <path>")
 
     if (res.status >= 400) {
       // TODO(bkendall): consider moving stream-handling logic to responseToError.
-      const r = await res.response.text();
+      const r = await utils.streamToString(res.body);
       let d;
       try {
         d = JSON.parse(r);
--- a/src/commands/database-get.ts
+++ b/src/commands/database-get.ts
@@ -118,7 +118,7 @@ export const command = new Command("database:get <path>")
 
     if (res.status >= 400) {
       // TODO(bkendall): consider moving stream-handling logic to responseToError.
-      const r = res.body ? await utils.streamToString(res.body) : "";
+      const r = await utils.streamToString(res.body);
       let d;
       try {
         d = JSON.parse(r);
--- a/src/deploy/hosting/uploader.ts
+++ b/src/deploy/hosting/uploader.ts
@@ -246,7 +246,7 @@ export class Uploader {
       logger.debug("[hosting][upload]", this.uploadQueue.stats());
     }
     if (res.status !== 200) {
-      const errorMessage = await streamToString(res.body as NodeJS.ReadableStream);
+      const errorMessage = res.body ? await streamToString(res.body as NodeJS.ReadableStream) : "";
       logger.debug(
         `[hosting][upload] ${this.hashMap[toUpload]} (${toUpload}) HTTP ERROR ${
           res.status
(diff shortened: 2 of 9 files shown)


src/emulator/hubExport.ts
```diff
@@ -337,7 +338,8 @@ export class HubExport {
       resolveOnHTTPError: true,
     });
     if (res.status >= 400) {
-      throw new FirebaseError(`Failed to export storage: ${await res.response.text()}`);
+      const errorMsg = res.body ? await streamToString(res.body as NodeJS.ReadableStream) : "";
```
joehan: Instead of casting, please set the generic type on client.request (as we do already in profiler.ts)

src/deploy/hosting/uploader.ts
```diff
@@ -245,7 +246,7 @@ export class Uploader {
       logger.debug("[hosting][upload]", this.uploadQueue.stats());
     }
     if (res.status !== 200) {
-      const errorMessage = await res.response.text();
+      const errorMessage = res.body ? await streamToString(res.body as NodeJS.ReadableStream) : "";
```
joehan: Instead of casting, please set the generic type on client.request (as we do already in profiler.ts)

src/commands/database-get.ts
```diff
@@ -118,7 +118,7 @@ export const command = new Command("database:get <path>")
 
     if (res.status >= 400) {
       // TODO(bkendall): consider moving stream-handling logic to responseToError.
-      const r = await res.response.text();
+      const r = res.body ? await utils.streamToString(res.body) : "";
```
joehan: Needing to do this at the call site is ugly. Let's move it into the util.streamtoString function

bkendall (approved):

PR #10820, opened 2026-07-20
fix: improve error message when deploying to a gcp project where fire…

### LINKED ISSUE #10379: Functions deployment to Google Cloud Projects fails with unspecific error
## Description

When running deploy of functions to a Google Cloud Project that doesn't have Firebase added you get a very unspecific 403 error. It would be nice to fail with something a bit more user friendly like "Error: Cannot deploy to project <project id> as it doesn't support Firebase. You can add Firebase support to this project in the Firebase Console."

## Repro steps

1. Create a new project in Google Cloud Console
2. Enable billing
3. In a terminal, create a folder and initialize:
`firebase init functions`
4. Try to deploy:
`firebase deploy --only functions --project <project id of cloud project>

Expected result: Deployment succeeds, or a nice error is shown
Actual result: A hard to understand 403 error:

```
=== Deploying to 'test1-493612'...

i  deploying functions
i  functions: preparing codebase default for deployment
i  artifactregistry: ensuring required API artifactregistry.googleapis.com is enabled...
⚠  functions: missing required API cloudbuild.googleapis.com. Enabling now...
⚠  functions: missing required API cloudfunctions.googleapis.com. Enabling now...
✔  artifactregistry: required API artifactregistry.googleapis.com is enabled

Error: Request to https://firebase.googleapis.com/v1beta1/projects/test1-493612/adminSdkConfig had HTTP Error: 404, Requested entity was not found.
```

I propose we at least update the error message to be easier to read, e.g.:

"Error: Cannot deploy to project <project id> as it doesn't support Firebase. You can add Firebase support to this project in the Firebase Console."

IzaakGough:
Fixes #10379

This PR improves the error message produced when a user attempts to deploy to a gcp project which doesn't have firebase added/enabled.

Note on the status code: the issue title and body say 403, but the log the reporter pasted is a 404 from `adminSdkConfig`, so that is the status matched here.

`getFirebaseConfig` also backs `functions:delete`, `functions:export` and the `ext:*` commands, so the message is command-neutral rather than phrased as a deploy failure. It also doesn't assert a cause, since the same endpoint 404s for a typo'd, deleted, or inaccessible project as well as for a Cloud project without Firebase.


src/functionsConfig.ts
```diff
@@ -114,10 +114,22 @@ export function getAppEngineLocation(config: any): string {
...
-  return response.body;
+  try {
+    const response = await apiClient.get<args.FirebaseConfig>(
+      `/v1beta1/projects/${projectId}/adminSdkConfig`,
+    );
+    return response.body;
+  } catch (err: unknown) {
+    if (err instanceof FirebaseError && err.status === 404) {
+      throw new FirebaseError(
+        `Cannot deploy to project ${clc.bold(projectId)} because it doesn't have Firebase enabled. ` +
```
CorieW: AI flagged the following:

**Message says "Cannot deploy" — 3 of 4 callers aren't deploys.**
`functions:delete`, `functions:export`, and `ext:install/configure/update` all hit this. `firebase functions:delete` printing "Cannot deploy" is a problem.

This is worth confirming.

src/functionsConfig.ts
```diff
@@ -114,10 +114,22 @@ export function getAppEngineLocation(config: any): string {
...
+          `Add Firebase to this Google Cloud project in the Firebase console, then try again:\n\n` +
+          `${consoleOrigin()}/project/${projectId}/settings/general`,
+        { original: err, exit: 1, status: 404 },
+      );
```
CorieW: Same endpoint 404s for typo'd, nonexistent, deleted, and invisible projects. Branch asserts one cause as fact, then links a console page for a project that may not exist.

Perhaps something like this instead:

```suggestion
      throw new FirebaseError(
        `Cannot deploy to project ${clc.bold(projectId)}. ` +
          `Make sure the project exists and your account has permission to access it.`,
        { original: err, exit: 1, status: 404 },
      );
```

src/functionsConfig.spec.ts
```diff
@@ -42,3 +46,47 @@ describe("config.parseUnsetArgs", () => {
...
+  after(() => {
+    nock.enableNetConnect();
+  });
+
+  afterEach(() => {
+    expect(nock.isDone()).to.be.true;
+  });
+
+  it("should return the admin SDK config on success", async () => {
+    nock("https://firebase.googleapis.com")
```
CorieW: Spec hardcodes the API origin multiple times.

Fix:
`import { firebaseApiOrigin } from "./api";` -> `nock(firebaseApiOrigin()).`

src/functionsConfig.ts
```diff
@@ -114,10 +114,25 @@ export function getAppEngineLocation(config: any): string {
...
+  try {
+    const response = await apiClient.get<args.FirebaseConfig>(
+      `/v1beta1/projects/${projectId}/adminSdkConfig`,
+    );
+    return response.body;
+  } catch (err: unknown) {
+    // adminSdkConfig 404s for any project this account can't see, so don't
+    // assert a cause. The addfirebase hint covers the Cloud-project case.
+    if (err instanceof FirebaseError && err.status === 404) {
+      throw new FirebaseError(
```
cabljac: Nit: the other multi-paragraph messages in this file sit as constants up top and get referenced at the throw, see :20 and :35 used at :54. Yours is the only long one assembled inline from concatenated fragments with `\n\n` in them, so you can't see the shape the user gets. Short one-liners do stay inline (:220, :228), so the split looks like size.

A helper alongside `getFunctionsConfigDeprecationMessage` on :42 would fit, since you need the project id:

```ts
const projectNotFoundMessage = (projectId: string): string => `Firebase project ${clc.bold(projectId)} was not found.
Make sure the project exists and that your account has access to it.

If ${clc.bold(projectId)} is a Google Cloud project without Firebase, add Firebase to it:

  firebase projects:addfirebase ${projectId}
`;
```

CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- Replaced the raw 404 from `adminSdkConfig` with a message naming the project and pointing at `firebase projects:addfirebase` (#10379)
```
cabljac: Nit: reads from our side rather than the user's. Something like "Improved the error shown when deploying to a Google Cloud project that does not have Firebase enabled (#10379)."

cabljac (approved):
@IzaakGough looks good. I checked the detection since the change hangs off it: `responseToError` sets `status` from the HTTP code (src/responseToError.ts:56-62) and `FirebaseError` defaults `status` to 500 (src/error.ts:13), so nothing else can fall into the 404 branch by accident. Reverted the guard and the 404 test goes red, broadened it and the 500 passthrough goes red, so the tests are doing real work. Two nits inline, neither blocking.

Worth a line in the description: #10379 says "403" in the title and body, but the log the reporter pasted is a 404, which is what you matched. Otherwise people will think you handled the wrong status.

src/functionsConfig.ts
IzaakGough: Moved it to a `projectNotFoundMessage(projectId)` helper next to `getFunctionsConfigDeprecationMessage`, so the message is one template literal and the shape the user sees is visible. Rendered output:

```
Firebase project my-gcp-project was not found.
Make sure the project exists and that your account has access to it.

If my-gcp-project is a Google Cloud project without Firebase, add Firebase to it:

  firebase projects:addfirebase my-gcp-project
```

CHANGELOG.md
IzaakGough: Reworded to your version, thanks.

src/functionsConfig.spec.ts
```diff
@@ -42,3 +47,53 @@ describe("config.parseUnsetArgs", () => {
...
+  before(() => {
+    nock.disableNetConnect();
+  });
+
+  after(() => {
+    nock.enableNetConnect();
+  });
+
+  afterEach(() => {
+    expect(nock.isDone()).to.be.true;
```
ajperel: Gemini suggests you also add `nock.cleanAll();` to ensure that any unused interceptors are cleared between tests, especially if a test fails before reaching assertions.

ajperel (approved):
This looks good to me but it might be worth waiting until next week to get review from @joehan since this will have impacts outside of functions.

I like this since it catches everything across commands nicely and improves the error message. But it does happen a bit into the process after checking a bunch of other stuff. One _could_ imagine we should ensure Firebase is enabled upfront for all (almost all?) commands potentially but that would be a much larger change I expect.

src/functionsConfig.spec.ts
IzaakGough: done in 4eebe5cfb940d29b1c11898fbeb9194588af489e

PR #10822, opened 2026-07-20
Add web support for Crashlytics MCP tools and prompts

tagboola:
### Description

Add support for web apps with Crashlytics MCP tools and prompts.


src/crashlytics/filters.ts
```diff
@@ -28,7 +28,10 @@ export const EventFilterSchema = z
         `Counts events originating from the given app versions. Must be obtained from the *displayName* field in an API response. `,
       ),
     issueId: z.string().optional().describe(`Count events for the given issue`),
-    issueVariantId: z.string().optional().describe(`Count events for the given issue variant`),
+    issueVariantId: z
```
schnecle, line 31: We should align on linters to remove this noise.

src/crashlytics/filters.ts
```diff
@@ -38,7 +41,7 @@ export const EventFilterSchema = z
     issueSignals: z
       .array(z.enum(["SIGNAL_EARLY", "SIGNAL_FRESH", "SIGNAL_REGRESSED", "SIGNAL_REPETITIVE"]))
       .optional()
-      .describe(`Counts events matching the given signals`),
+      .describe(`Counts events matching the given signals. Only supported for mobile apps.`),
```
schnecle, line 44: This looks like the kind of thing that an agent is really going to mess up. Have you tested this one out locally? It might perform better if you say mobile apps and then qualify it like "mobile apps (e.g. iOS and Android)"

src/mcp/util/crashlytics/availability.spec.ts
```diff
@@ -295,6 +295,41 @@ describe("isCrashlyticsAvailable", () => {
...
+            firebase: "^10.0.0",
+          },
+        }),
+      },
+    });
+
+    const result = await isCrashlyticsAvailable(mockContext("/test-dir"));
+
+    expect(result).to.be.false;
+  });
```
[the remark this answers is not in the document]
schnecle: This is only true if we what it is saying below is true.

src/mcp/util/crashlytics/availability.ts
```diff
@@ -84,3 +89,14 @@ async function flutterAppUsesCrashlytics(appPath: string): Promise<boolean> {
...
+async function webAppUsesCrashlytics(appPath: string): Promise<boolean> {
+  const packageJsonFiles = await detectFiles(appPath, "package.json");
+  for (const file of packageJsonFiles) {
+    const content = await fs.readFile(path.join(appPath, file), "utf8");
+    if (content.includes("firebase/crashlytics")) {
+      return true;
+    }
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
schnecle: Is that true? How is the js sdk bundled?

src/mcp/resources/guides/crashlytics_reports.ts
```diff
@@ -46,25 +48,44 @@ Mobile apps have one of the following reports available, depending on the platfo
...
+    display_name: "Web Metrics"
+    usage: |
+      Counts events and distinct impacted users, grouped by error type.
+      Error types are sorted by event count, in descending order.
+
 Report responses contain the following metrics:
 
   - eventsCount: the number of events matching
   - impactedUsers: the number of distinct end users in all the matching events
+  - sessionsCount: the number of distinct sessions in all the matching events. Only availabe for we apps.
```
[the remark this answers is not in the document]
schnecle: Please fix.

src/crashlytics/types.ts
```diff
@@ -204,6 +218,22 @@ export interface OperatingSystem {
   displayName?: string;
 }
 
+/** Web broeser metadata */
```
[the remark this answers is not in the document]
schnecle: Please fix.

schnecle (commented):
A couple small things

src/crashlytics/filters.ts
maxl0rd: It wrapped because he added a second sentence :)

src/crashlytics/filters.ts
maxl0rd: I think this instruction will mostly work. If the api can just discard invalid filters, this is probably fine.

src/crashlytics/types.ts
```diff
@@ -22,6 +24,8 @@ export interface IntervalMetrics {
   eventsCount: number;
   /** The cardinality of distinct users in the set of events. */
   impactedUsersCount: number;
+  /** Web only.The number of distinct sessions in the set of events. */
```
maxl0rd: How about something like: "The number of distinct sessions in the set of events. Only computed for web applications."

src/crashlytics/types.ts
```diff
@@ -22,6 +24,8 @@ export interface IntervalMetrics {
   eventsCount: number;
   /** The cardinality of distinct users in the set of events. */
```
maxl0rd: "The number of distinct users in the set of events."

(Our tech writers hate "cardinality")

src/crashlytics/types.ts
```diff
@@ -30,7 +34,7 @@ export interface Version {
   displayVersion?: string;
   /** One display_version can have many build_version. */
   buildVersion?: string;
-  /** Compound human-readable string containing both display and build versions. */
+  /** Mobile apps only. Compound human-readable string containing both display and build versions. */
```
maxl0rd: "Compound human-readable string containing both display and build versions. Only present for mobile applications."

src/crashlytics/types.ts
```diff
@@ -608,6 +677,14 @@ export interface GetReportRequest {
   pageSize?: number;
   /** A page token, received from a previous calls. */
   pageToken?: string;
+  /**
+   * For interval metrics, in addition to listing the value for the entire time
```
maxl0rd: Consider rewriting this comment to be more direct.

src/crashlytics/filters.ts
schnecle: oh, oh, I just didn't see it on this line - please ignore!

src/mcp/resources/guides/crashlytics_reports.ts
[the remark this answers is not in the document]
tagboola: Done.

src/crashlytics/types.ts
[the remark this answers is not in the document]
tagboola: Done.

src/mcp/util/crashlytics/availability.ts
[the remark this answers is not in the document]
tonybaroneee: So during the EAP, they’ll need to install and import “@firebase/crashlytics” explicitly due to how things are packaged/exposed. This will be detectable in package.json.

After we merge to main and are part of the larger SDK release, then they can just import “firebase/crashlytics” in their code as it will be bundled in the larger “firebase” dependency in package.json. **At this point, there is no longer a guarantee that they have anything with "crashlytics" in their package.json.**

Happy to chat this out more in person if it's still unclear.

src/crashlytics/types.ts
tagboola: Done.

src/crashlytics/types.ts
tagboola: Done.

src/crashlytics/types.ts
tagboola: Done.

src/crashlytics/types.ts
tagboola: Done.

src/mcp/util/crashlytics/availability.ts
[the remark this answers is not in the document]
tagboola: Updated this to check to see if `firebase` is installed _and_ `firebase/crashlytics` is imported.

src/mcp/util/crashlytics/availability.ts
```diff
@@ -92,11 +98,36 @@ async function flutterAppUsesCrashlytics(appPath: string): Promise<boolean> {
...
+    } catch {
+      // Ignore invalid JSON or unreadable package.json
     }
   }
+
+  if (!hasFirebaseDep) {
+    return false;
+  }
+
+  const sourceFiles = await detectFiles(appPath, "*.{js,jsx,ts,tsx,mjs,cjs,html,vue,svelte}");
```
schnecle: This could be very costly to the MCP start up time on a big app. For the other options, we try to do something that works for most folks and then give the workaround that they can force the crashlytics tools to be there even if not detected. Do we have any conventions around where someone would import the crashlytics dep that we could use to cover that 90% case?

tagboola: Agreed-- I'm worried about the performance of this as well. @tonybaroneee is there a particular file where we expect most developers to import crashlytics? 

I just remembered for the app testing agent that we just do a check to see if the app distribution API is enabled on the project to determine if we expose the app testing agent MCP tools. I wonder if is there an API or onboarding step that we can check to determine if they're onboarded for crash for web. That may be easier than trying to pinpoint a specific file that has a crashlytics import.

tonybaroneee: Yeah, maybe let's just call a GET to the telemetry admin API config endpoint to see if they've onboarded. That's a pretty strong signal (even though they still might not have provisioned the SDK in their app yet).

tagboola: Updated this to just check if the telemetry admin API config exists instead of checking to see if the depedency is declared in the package.json file.

src/mcp/util/crashlytics/availability.ts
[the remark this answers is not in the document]
tagboola: Prefer https://github.com/firebase/firebase-tools/pull/10822#discussion_r3639857700 to continue discussing this.

src/mcp/tools/crashlytics/notes.ts
```diff
@@ -8,7 +8,8 @@ export const create_note = tool(
   "crashlytics",
   {
     name: "create_note",
-    description: "Add a note to an issue from crashlytics.",
+    description:
+      "Add a note to an issue from crashlytics. Only supported for mobile (Android and iOS) apps.",
```
joehan, line 12: If an agent tries to call this for a web app, what happens? Is the error clear?

joehan (approved):
Add a changelog entry, but seems fine to me otherwise.

src/mcp/tools/crashlytics/notes.ts
tagboola: Testing in Jetski, the agent wouldn't make the call because its a web app, which is expected.  If the agent were to actually make the call, the backend returns a 404 so it may not be super clear why it failed. If this becomes an issue I think we could update the backend to return a better error.

PR #10823, opened 2026-07-20
extensions deprecation CLI warning

shettyvarun268:
### Description

With the Firebase Extensions managed service scheduled for official decommission on March 31, 2027, this PR introduces the `extdeprecationwarnings` experiment flag (enabled by default) to notify developers across `ext:*` CLI commands without cluttering terminal output or breaking non-interactive CI workflows.

Key features introduced:
1. **Experiment Registration**: Registered `extdeprecationwarnings` in `src/experiments.ts` (default: `true`). Added support for negative opt-out (`FIREBASE_CLI_EXPERIMENTS=-extdeprecationwarnings`).
2. **Categorized Warning Architecture (`src/extensions/warnings.ts`)**:
   - **Category 1 (`ext:install`, `ext:configure`, `ext:update`, `ext:sdk:install`)**: Displays a concise warning before interactive prompts notifying users of the March 31, 2027 decommission.
   - **Category 2 (`ext:list`, `ext:info`)**: Displays a single-line notice as a footer after successful command execution.
   - **Category 3 (`ext:uninstall`, `ext:dev:deprecate`, `ext:export`)**: Silenced (`NO_WARNING_COMMANDS`).
   - **Category 4 (`ext:dev:init`, `ext:dev:upload`, `ext:dev:list`, `ext:dev:usage`, `ext:dev:undeprecate`)**: Displays publisher banner immediately before command execution.
   - **Category 5 (`ext:dev:register`)**: Hard stop throwing `FirebaseError(..., { exit: 1 })`.
3. **Automated Silencing**: Automatically suppresses text warnings when `--json`, `--non-interactive`, `--quiet`, non-TTY stdio, or automated CI environments (`process.env.CI`, `GITHUB_ACTIONS`, `BUILD_ID`, `TF_BUILD`) are present.
4. **CLI Hook**: Intercepts `Command.runner()` in `src/command.ts` to trigger warning hooks before pre-action execution and after completion.

---

### Scenarios Tested

- **Unit Tests**: Added comprehensive Mocha test suites in `src/extensions/warnings.spec.ts` and `src/experiments.spec.ts` verifying Category 1–5 output formatting, hard errors under `--json`, and CI test environment isolation (`12 passing`).
- **Manual Interactive Testing**: Built and tested locally (`./lib/bin/firebase.js`):
  - Verified concise warning appears before `ext:install` interactive prompts.
  - Verified publisher notice banner appears before `ext:dev:upload`.
  - Verified post-execution footer appears after `ext:list`.
  - Verified hard stop failure (`exit 1`) on `ext:dev:register`.
  - Verified clean JSON output when `--json` flag is provided.
  - Verified silencing when running with `FIREBASE_CLI_EXPERIMENTS=-extdeprecationwarnings`.
- **Linting & Formatting**: Verified `npm run format` and `npm run lint:changed-files` pass with 0 errors.

---

### Sample Commands

```bash
# Default behavior (warnings visible)
firebase ext:install

# Test publisher notice banner
firebase ext:dev:upload <ref>

# Test hard exit
firebase ext:dev:register

# Explicit opt-out
FIREBASE_CLI_EXPERIMENTS=-extdeprecationwarnings firebase ext:install

### DIFF
--- a/src/experiments.ts
+++ b/src/experiments.ts
@@ -299,14 +299,8 @@ export function setEnabled(name: ExperimentName, to: boolean | null): void {
 export function enableExperimentsFromCliEnvVariable(): void {
   const experiments = process.env.FIREBASE_CLI_EXPERIMENTS || "";
   for (const experiment of experiments.split(",")) {
-    const trimmed = experiment.trim();
-    if (trimmed.startsWith("-")) {
-      const expName = trimmed.slice(1);
-      if (isValidExperiment(expName)) {
-        setEnabled(expName, false);
-      }
-    } else if (isValidExperiment(trimmed)) {
-      setEnabled(trimmed, true);
+    if (isValidExperiment(experiment)) {
+      setEnabled(experiment, true);
     }
   }
 }
--- a/src/experiments.ts
+++ b/src/experiments.ts
@@ -214,6 +214,11 @@ export const ALL_EXPERIMENTS = experiments({
     default: false,
     public: true,
   },
+  extdeprecationwarnings: {
+    shortDescription: "Show deprecation warnings for Firebase Extensions CLI commands.",
+    default: false,
+    public: true,
+  },
 });
 
 export type ExperimentName = keyof typeof ALL_EXPERIMENTS;
@@ -288,8 +293,14 @@ export function setEnabled(name: ExperimentName, to: boolean | null): void {
 export function enableExperimentsFromCliEnvVariable(): void {
   const experiments = process.env.FIREBASE_CLI_EXPERIMENTS || "";
   for (const experiment of experiments.split(",")) {
-    if (isValidExperiment(experiment)) {
-      setEnabled(experiment, true);
+    const trimmed = experiment.trim();
+    if (trimmed.startsWith("-")) {
+      const expName = trimmed.slice(1);
+      if (isValidExperiment(expName)) {
+        setEnabled(expName, false);
+      }
+    } else if (isValidExperiment(trimmed)) {
+      setEnabled(trimmed, true);
     }
   }
 }
--- a/src/experiments.ts
+++ b/src/experiments.ts
@@ -216,7 +216,7 @@ export const ALL_EXPERIMENTS = experiments({
   },
   extdeprecationwarnings: {
     shortDescription: "Show deprecation warnings for Firebase Extensions CLI commands.",
-    default: false,
+    default: true,
     public: true,
   },
 });
--- a/src/extensions/warnings.ts
+++ b/src/extensions/warnings.ts
@@ -3,7 +3,9 @@ import * as clc from "colorette";
 import { logPrefix } from "./extensionsHelper";
 import { humanReadable } from "../deploy/extensions/deploymentSummary";
 import { InstanceSpec, getExtensionVersion } from "../deploy/extensions/planner";
+import { FirebaseError } from "../error";
 import { logger } from "../logger";
+import { Options } from "../options";
 import * as utils from "../utils";
 
 const toListEntry = (i: InstanceSpec) => {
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: src/extensions/warnings.ts [hunk 1]
```diff
@@ -53,3 +55,109 @@ export function outOfBandChangesWarning(instanceIds: string[], isDynamic: boolea
...
+  if (commandName === "ext:dev:register") {
+    throw new FirebaseError(
+      `ext:dev:register is disabled. Registering new publisher profile IDs is no longer supported.\n` +
+        `The Firebase Extensions service will shut down on March 31, 2027.\n` +
+        `Learn more: ${FAQ_URL}`,
+      { exit: 1 },
+    );
+  }
+
+  if (
```
ajperel: You don't need this check at all right because you'll only show warnings on the code below if you are the commands that show deprecation warnings before the command. I think you can just remove this and once you do that you don't need NO_WARNING_COMMANDS

src/extensions/warnings.ts
```diff
@@ -53,3 +55,109 @@ export function outOfBandChangesWarning(instanceIds: string[], isDynamic: boolea
...
+/**
+ * Displays post-execution deprecation warnings (e.g. single-line footer for ext:list and ext:info).
+ * @param commandName Name of the command being run.
+ * @param options Command execution options object.
+ */
+export function showDeprecationWarningAfter(
+  commandName: string,
+  options: Options | Record<string, unknown>,
+): void {
+  if (isSilenced(options) || !LIST_COMMANDS.has(commandName)) {
```
ajperel: I think the approach in _before_ is better in that I think it's better to only have a positive check for the commands you want to show a warning for than a negative check to exit early if it's _not_ those commands.

I realize our command set isn't going to change. But as a general practice... if we were creating more commands etc. I think we'd be less likely to have subtle bugs this way. For example... do you have this warning for the NO_WARNING_COMMANDS as is?

src/extensions/warnings.ts
```diff
@@ -53,3 +55,109 @@ export function outOfBandChangesWarning(instanceIds: string[], isDynamic: boolea
       `\nIf you proceed with this deployment, those changes will be overwritten.${extra}`,
   );
 }
+
+const FAQ_URL = "https://firebase.google.com/docs/extensions/faq-and-troubleshooting";
+
+const NO_WARNING_COMMANDS = new Set(["ext:uninstall", "ext:dev:deprecate", "ext:export"]);
+const LIST_COMMANDS = new Set(["ext:list", "ext:info"]);
```
ajperel: This breakdown of commands totally works, but puts a burden on your reader and the rest of the code to map command types to how we want them to behave in the warnings. I think it might make more sense to do something like 

WARN_BEFORE_COMMANDS =  // install commands
WARN_STRONGLY_BEFORE_COMMANDS = // publisher commands
WARN_AFTER_COMMANDS = // the list ones

Then it's pretty readable to be like "we're in after, so check the after commands" and we can document the why of each list as we define them

src/experiments.ts
```diff
@@ -294,8 +299,14 @@ export function setEnabled(name: ExperimentName, to: boolean | null): void {
 export function enableExperimentsFromCliEnvVariable(): void {
   const experiments = process.env.FIREBASE_CLI_EXPERIMENTS || "";
   for (const experiment of experiments.split(",")) {
-    if (isValidExperiment(experiment)) {
-      setEnabled(experiment, true);
+    const trimmed = experiment.trim();
+    if (trimmed.startsWith("-")) {
```
ajperel: Is this an API change adding support for turning off experiments with a '-'? Doesn't feel like we should bury that in this PR but do it separately (and would probably need review). I would say take this out.

People will have to run `firebase experiments:disable extdeprecationwarnings`

And follow up with Joe/Thomas on if we want to support broadly an environment variable way of turning off extensions for an invocation.

src/extensions/warnings.ts
```diff
@@ -53,3 +55,109 @@ export function outOfBandChangesWarning(instanceIds: string[], isDynamic: boolea
...
+  "ext:dev:list",
+  "ext:dev:usage",
+  "ext:dev:undeprecate",
+]);
+
+/**
+ * Checks if deprecation warnings should be silenced (e.g. non-interactive, JSON, non-TTY, quiet mode, or CI).
+ * @param options Command execution options object.
+ */
+export function isSilenced(options: Options | Record<string, unknown>): boolean {
```
ajperel, line 85: The fact that we apparently don't have this utility because we show all other warnings in most of these environments makes me question a bit if we really need to do it here. But I don't think it hurts to do it and this code will all go away after the deprecation so.... sure. Why not keep it.

src/extensions/warnings.ts
shettyvarun268: Done! I removed this check entirely along with NO_WARNING_COMMANDS

src/extensions/warnings.ts
shettyvarun268: Great suggestion! I've refactored the command sets to use positive matching (WARN_BEFORE_COMMANDS, WARN_STRONGLY_BEFORE_COMMANDS, and WARN_AFTER_COMMANDS) and completely removed NO_WARNING_COMMANDS. This makes the warning behavior self-documenting and avoids potential bugs if new commands are added in the future.

src/experiments.ts
shettyvarun268: Noted. I've removed the leading hyphen parsing from enableExperimentsFromCliEnvVariable() and reverted src/experiments.ts back to main. Users can turn off the experiment using firebase experiments:disable extdeprecationwarnings.

src/extensions/warnings.ts
shettyvarun268: Right. I got this thought since we did the silencing part for the stock-out warnings as well thinking it can get annoying for users to keep seeing warnings.

src/extensions/warnings.ts
shettyvarun268: I've renamed LIST_COMMANDS to WARN_AFTER_COMMANDS to explicitly define the whitelist of commands that should trigger a post-execution footer. showDeprecationWarningAfter now only runs if the command is positively matched in WARN_AFTER_COMMANDS, ensuring commands like ext:export or ext:uninstall never trigger it.

src/experiments.ts
ajperel: Yeah. To be clear, I think this is a nifty idea. It's likely worth doing the small proposal for it and bringing it back in another PR. I just didn't want to sneak it in with this change.

src/extensions/warnings.ts
```diff
@@ -53,3 +55,114 @@ export function outOfBandChangesWarning(instanceIds: string[], isDynamic: boolea
...
+/**
+ * Displays deprecation warnings or throws hard exit errors before an ext:* command executes.
+ * @param commandName Name of the command being run (e.g. "ext:install").
+ * @param options Command execution options object.
+ */
+export function showDeprecationWarningBefore(
+  commandName: string,
+  options: Options | Record<string, unknown>,
+): void {
+  if (commandName === "ext:dev:register") {
```
ajperel, line 119: Technically it's probably worth having this particular check even outside the experiment because it's never going to work (we've disabled it) and disabling the experiment doesn't change that. But... it's not going to work either way, and this is super rare I assume, so I don't know that it's worth taking the time to change.

ajperel (approved):

inlined (approved):

PR #10826, opened 2026-07-20
fix(apiv2): sanitize domain-scoped project ID in x-goog-user-project header

lfkellogg:
### Description
Strip domain scope prefix from GOOGLE_CLOUD_QUOTA_PROJECT before setting x-goog-user-project header to prevent API rejections when domain-scoped project IDs are used.

### Scenarios Tested
- Verified apiv2 header generation logic.

### Sample Commands
GOOGLE_CLOUD_QUOTA_PROJECT=domain.com:api-project-123 firebase apptesting:execute

TAG=agy
CONV=107aa4d3-03d0-4299-afcc-27e78b24392e


lfkellogg:
I actually don't think this is what the customer was referring to. Closing for now.

PR #10828, opened 2026-07-20
update workflow to use environment secrets

hlshen:
Update vscode release workflow to use APP token instead of default github token.

### DIFF
--- a/.github/workflows/release-vscode-ext.yaml
+++ b/.github/workflows/release-vscode-ext.yaml
@@ -24,10 +24,11 @@ jobs:
     steps:
       - name: Generate Token
         id: generate_token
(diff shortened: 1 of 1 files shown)


.github/workflows/release-vscode-ext.yaml
```diff
@@ -21,9 +22,15 @@ jobs:
        github.event.issue.author_association == 'MEMBER')
     runs-on: ubuntu-latest
     steps:
+      - name: Generate Token
+        id: generate_token
+        uses: actions/create-github-app-token@v1
```
morganchen12:
```
actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
```

morganchen12 (approved):

PR #10834, opened 2026-07-21
fix!: add validation for names of task queue funcs

### LINKED ISSUE #7365: `firebase deploy` fails to delete task function with an underscore in its name

**firebase-tools:** 13.11.4

**Platform:** Ubuntu 23.10 / fish

### [REQUIRED] Test case

```python
from firebase_functions import tasks_fn

@tasks_fn.on_task_dispatched()
def dummy_function(request: tasks_fn.CallableRequest) -> str:
    return "Hello World"
```

1. Initialize functions.
1. Save the above function in `functions/main.py` (the default).
1. Execute `firebase deploy --only functions`. `dummy_function` will be deployed. (cf. https://github.com/firebase/firebase-tools/issues/6324)
1. Rename `dummy_function` to `dummy` in `functions/main.py`.
1. Execute `firebase deploy --only functions`. When queried for deletion of `dummy_function`, answer `y` (yes).

The old function should be deleted.

The old function is not deleted. Instead, an error is shown.

```

The following functions are found in your project but do not exist in your local source code:
	dummy_function(us-central1)

If you are renaming a function or changing its region, it is recommended that you create the new function first before deleting the old one to prevent event loss. For more info, visit https://firebase.google.com/docs/functions/manage-functions#modify

? Would you like to proceed with deletion? Selecting no will continue the rest of the deployments. (y/N) 
```

```
i  functions: deleting Python 3.12 (2nd Gen) function dummy_function(us-central1)...
HTTP Error: 400, Queue ID "dummy_function" can contain only letters ([A-Za-z]), numbers ([0-9]), or hyphens (-). Queue ID must be between 1 and 100 characters.
```

IzaakGough:
Fixes #7365

Cloud Functions names may contain underscores, but Cloud Tasks queue IDs may not, and the queue name is derived from the function id. So a task queue function like `dummy_function` deploys, its queue is never created, and deleting it later calls `updateQueue` with an illegal name and fails with a 400. There is no way to remove the function through the CLI.

Two changes:

- `taskQueueFunctionNamesAreValid` rejects task queue function names that are not legal queue IDs at deploy time, so no new functions get into this state.
- `disableTaskQueue` skips the call when the function id is not a legal queue ID. No queue by that name can exist, so this is a no-op for every validly named function and unblocks deletion for the ones already stuck.

The queue ID rule now lives in `cloudtasks.isValidQueueId` so both sides use the same definition.

## Decision

This is a breaking change, deliberately. `endpointsAreValid` runs over every endpoint in the codebase, not just the filtered set, so anyone with an underscore-named task queue function in source now has their whole deploy fail rather than just that function. Those functions were already broken (their queue never existed), so this surfaces a silent failure rather than creating a new one, but it will be a new error for people who have not noticed yet.

Python is the main population: endpoint ids come from `func.__name__` verbatim, and Python names cannot contain hyphens, so the error tells those users to pick a name with no separator.

## Testing

Unit tests for the predicate, the validation, its wiring into `endpointsAreValid`, and the delete-path skip. Manually verified the deploy-time error.


src/gcp/cloudtasks.spec.ts
```diff
@@ -31,6 +32,26 @@ describe("CloudTasks", () => {
...
+
+    it("rejects underscores", () => {
+      expect(cloudtasks.isValidQueueId("dummy_function")).to.be.false;
+    });
+
+    it("accepts uppercase", () => {
+      expect(cloudtasks.isValidQueueId("MyQueue")).to.be.true;
+    });
+
+    it("rejects empty and over-long ids", () => {
```
CorieW: Nit:
```suggestion
    it("rejects ids that are empty or too long", () => {
```

CorieW:
Since this is a breaking change, is there a convention that this repo follows for breaking changes, such as:

- Renaming PR title to `fix!: add validation for names of task queue funcs`.
- Or something in CHANGELOG.md, like `[BREAKING]`?

src/deploy/functions/validate.spec.ts
```diff
@@ -114,6 +114,44 @@ describe("validate", () => {
...
 
+  describe("taskQueueFunctionNamesAreValid", () => {
+    const ENDPOINT_BASE = {
+      platform: "gcfv2",
+      id: "id",
+      region: "us-east1",
+      project: "project",
+      entryPoint: "id",
+      runtime: "nodejs16",
+    } as const;
```
CorieW:
```suggestion
    const ENDPOINT_BASE: Omit<backend.Endpoint, "httpsTrigger"> = {
      platform: "gcfv2",
      id: "id",
      region: "us-east1",
      project: "project",
      entryPoint: "id",
      runtime: "nodejs16",
    };
```

Remains consistent with https://github.com/firebase/firebase-tools/blob/8b00f81c2a4216a69b7d42a537e398f0c77b9a2c/src/deploy/functions/validate.spec.ts#L732

src/deploy/functions/validate.ts
```diff
@@ -329,6 +331,27 @@ export function functionIdsAreValid(functions: { id: string; platform: string }[
...
+ * https://cloud.google.com/tasks/docs/reference/rest/v2/projects.locations.queues#Queue
+ * @throws { FirebaseError } Task queue function names must be valid Cloud Tasks queue IDs.
+ */
+export function taskQueueFunctionNamesAreValid(endpoints: backend.Endpoint[]): void {
+  const invalidIds = endpoints
+    .filter(backend.isTaskQueueTriggered)
+    .filter((ep) => !cloudtasks.isValidQueueId(ep.id));
+  if (invalidIds.length !== 0) {
+    const msg =
+      `Task queue function name(s) ${invalidIds.map((f) => f.id).join(", ")} cannot be used as ` +
```
CorieW: Nit: `f` and `ep` name the same value three lines apart. `ep` is dominant in the file; `f` is copy-drift from a neighbour whose parameter really is `functions`.

```suggestion
      `Task queue function name(s) ${invalidIds.map((ep) => ep.id).join(", ")} cannot be used as ` +
```

CorieW (approved):
Looks good, mostly nit picking.

PR #10839, opened 2026-07-22
BUG: experimental secret support for non-standard resource IDs was not propagating to deploy prepare

Berlioz:
resolveParams() can now change the resourceID field of a secret param, which means that changes to them must be propagated to the Build's secretEnvVars before Backend compilation to be picked up by Functions/Run.

### DIFF
--- a/src/deploy/functions/build.ts
+++ b/src/deploy/functions/build.ts
@@ -880,17 +880,20 @@ export function matchSecretEnvVars(build: Build): void {
       continue;
     }
     const secretParam = param;
+    if (!secretParam.resourceId) {
+      continue;
+    }
 
-    for (const endpointName of Object.keys(build.endpoints)) {
-      const endpoint = build.endpoints[endpointName];
+    for (const endpoint of Object.values(build.endpoints)) {
       for (const envVar of endpoint.secretEnvironmentVariables ?? []) {
-        if (envVar.key === secretParam.name) {
-          if (secretParam.resourceId && envVar.secret !== secretParam.resourceId) {
-            logger.debug(
-              `Inserted newly created secret into build.SecretEnvVars: ${envVar.key}=${secretParam.resourceId}`,
-            );
-            envVar.secret = secretParam.resourceId;
-          }
+        if (envVar.key.toUpperCase() !== secretParam.name.toUpperCase()) {
+          continue;
+        }
+        if (envVar.secret !== secretParam.resourceId) {
+          logger.debug(
+            `Inserted newly created secret into build.SecretEnvVars: ${envVar.key}=${secretParam.resourceId}`,
+          );
+          envVar.secret = secretParam.resourceId;
         }
       }
     }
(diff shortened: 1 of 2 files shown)


src/deploy/functions/build.ts
```diff
@@ -863,6 +865,38 @@ export function applyEnvSecretBindings(
...
+            logger.debug(
+              `Inserted newly created secret into build.SecretEnvVars: ${envVar.key}=${secretParam.resourceId}`,
+            );
+            envVar.secret = secretParam.resourceId;
+          }
+        }
+      }
+    }
+  }
+}
```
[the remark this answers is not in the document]
Berlioz: but it came through applyEnvSecretBindings, so we know--whatever, fine

shettyvarun268:
Since matchSecretEnvVars is newly exported, we might want to have some unit tests in place. Specifically to test that envVar.secret updates when secretParam.resourceId changes and that case-insensitive matching works properly.

src/deploy/functions/build.ts
```diff
@@ -863,6 +865,41 @@ export function applyEnvSecretBindings(
...
+        }
+        if (envVar.secret !== secretParam.resourceId) {
+          logger.debug(
+            `Inserted newly created secret into build.SecretEnvVars: ${envVar.key}=${secretParam.resourceId}`,
+          );
+          envVar.secret = secretParam.resourceId;
+        }
+      }
+    }
+  }
```
shettyvarun268, line 900: Jetski thinks we can flatten this more:

```javascript
export function matchSecretEnvVars(build: Build): void {
  const secretParamsByKey = new Map<string, SecretParam>();
  for (const param of build.params) {
    if (param.type === "secret" && param.resourceId) {
      secretParamsByKey.set(param.name.toUpperCase(), param);
    }
  }
  if (secretParamsByKey.size === 0) return;

  for (const endpoint of Object.values(build.endpoints)) {
    for (const envVar of endpoint.secretEnvironmentVariables ?? []) {
      const secretParam = secretParamsByKey.get(envVar.key.toUpperCase());
      if (secretParam && envVar.secret !== secretParam.resourceId) {
        logger.debug(
          `Inserted newly created secret into build.SecretEnvVars: ${envVar.key}=${secretParam.resourceId}`,
        );
        envVar.secret = secretParam.resourceId;
      }
    }
  }
}

Berlioz: I'm pretty sure Record is preferred to Map in this repo? 

More seriously, I don't agree with Jetski here and I don't even think this suggestion would compile without some fairly ugly surgery: in the context of this file, SecretParam is build.SecretParam which contains just the wire-level representation of a secret that we get from the SDK. For the types to line up here the Map would have to be declared with the params.SecretParam type, which isn't currently exported.

shettyvarun268: Fair point on both counts! I didn't catch that SecretParam in build.ts was a local type collision with params.ts. Since you have already pushed the test cases, going ahead with the approval

shettyvarun268 (approved):
LGTM

PR #10840, opened 2026-07-22
BUG: functions:secrets:prune was not affecting secrets created after apphosting launch

Berlioz:
see: https://github.com/firebase/firebase-tools/blob/16fb909bffb745d39b13678e573006628ca21589/src/gcp/secretManager.ts#L474


shettyvarun268 (approved):
Thanks for fixing this!
LGTM

PR #10841, opened 2026-07-23
feat(ailogic): add ailogic:config CLI commands

marb2000:
### Description

Adds the `firebase ailogic:config:*` commands to read and modify Firebase AI Logic configuration, gated behind the `ailogic` experiment (same as `ailogic:providers`).

- `ailogic:config:get [path]` reads the config. With no path it prints the whole resource; with a path it prints one value. Providers appear as read-only status derived from API enablement.
- `ailogic:config:set <path> <value>` writes one setting. Developer-facing paths map to the underlying resource fields:
  - `security.auth-only` -> `trafficFilter.firebaseAuthRequired`
  - `security.template-only` -> `trafficFilter.templateOnly`
  - `monitoring.state` -> `telemetryConfig.mode` (ALL/NONE)
  - `monitoring.sample-rate-percentage` -> `telemetryConfig.samplingRate` (an integer percent 1-100 exposed as the API's (0,1] fraction)

Tightening `security.auth-only` or `security.template-only` from false to true prompts for confirmation because existing clients start getting rejected; `confirm()` enforces `--force` in non-interactive mode. Input (path and value) is validated up front, before the API-enablement flow, so bad input fails fast.

**Depends on #10778** (`ailogic:providers`). This branch is rebased on top of that one, so until #10778 merges the diff here also shows the provider commits; once #10778 lands in `main` the diff narrows to the config-only changes automatically. Please review the `ailogic-config-*` and config-related `gcp/ailogic.ts` changes here; the provider files are the already-approved #10778 content.

### Scenarios Tested

- Unit tests: `ailogic-config-get.spec.ts`, `ailogic-config-set.spec.ts`, plus `getConfig`/`updateConfig` in `gcp/ailogic.spec.ts`. Covers unknown path, non-boolean and out-of-range values, false->true tightening confirmation, non-interactive-without-force, value/field mapping, nested-path reads, and fail-fast-before-enablement.
- `npm run build`, `npm run lint`, and `tsc --noEmit` are clean.
- Commands are hidden by default and appear under `FIREBASE_CLI_EXPERIMENTS=ailogic`.

### Sample Commands

```
firebase experiments:enable ailogic
firebase ailogic:config:get
firebase ailogic:config:get security.auth-only
firebase ailogic:config:set security.auth-only true
firebase ailogic:config:set monitoring.sample-rate-percentage 50
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 - Added `firebase ailogic:providers:*` CLI commands to enable, disable, and list Gemini API providers.
+- Added `firebase ailogic:config:*` CLI commands to read and modify AI Logic configuration settings.
 - Add `MCP-Protocol-Version`, `Mcp-Method`, and `Mcp-Name` HTTP headers to `OneMcpServer` requests per the MCP 0728 standard release candidate (https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/ and https://modelcontextprotocol.io/seps/2243-http-standardization).
 - Fixes Storage Emulator to support JSON uploads larger than 100KB without hanging or throwing 413 error (#8355)
 - Add `extdeprecationwarnings` experiment to display phased deprecation notices and guidance across `ext:*` CLI commands.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,4 +2,3 @@
 - Add `MCP-Protocol-Version`, `Mcp-Method`, and `Mcp-Name` HTTP headers to `OneMcpServer` requests per the MCP 0728 standard release candidate (https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/ and https://modelcontextprotocol.io/seps/2243-http-standardization).
 - Fixes Storage Emulator to support JSON uploads larger than 100KB without hanging or throwing 413 error (#8355)
 - Add `extdeprecationwarnings` experiment to display phased deprecation notices and guidance across `ext:*` CLI commands.
-
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,2 @@
-- Fixed an issue in `apps:create` where App Store ID was always prompted for even when unnecessary.
-- Add `functions:lifecycle:list` and `functions:lifecycle:run` commands to view and run
-  lifecycle hooks in isolation.
-- Updated the Firebase SQL Connect local toolkit to v3.4.15, which supports for 1:1 nested mutations. (#10773)
+- Added `firebase ailogic:providers:*` CLI commands to enable, disable, and list Gemini API providers.
+- Updated Pub/Sub emulator to version 0.8.34
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,3 @@
-- Added `firebase ailogic:providers:*` CLI commands to enable, disable, and list Gemini API providers.
-- Added `firebase ailogic:config:*` CLI commands to read and modify AI Logic configuration settings.
 - Add `MCP-Protocol-Version`, `Mcp-Method`, and `Mcp-Name` HTTP headers to `OneMcpServer` requests per the MCP 0728 standard release candidate (https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/ and https://modelcontextprotocol.io/seps/2243-http-standardization).
 - Fixes Storage Emulator to support JSON uploads larger than 100KB without hanging or throwing 413 error (#8355)
 - Add `extdeprecationwarnings` experiment to display phased deprecation notices and guidance across `ext:*` CLI commands.
--- a/src/commands/ailogic-config-get.ts
+++ b/src/commands/ailogic-config-get.ts
@@ -1,0 +1,85 @@
+import { Command } from "../command";
+import { requirePermissions } from "../requirePermissions";
+import { needProjectId } from "../projectUtils";
+import * as ailogic from "../gcp/ailogic";
+import { logger } from "../logger";
+import { FirebaseError } from "../error";
+
+import { Options } from "../options";
+
+export const command = new Command("ailogic:config:get [path]")
+  .description("read AI Logic configuration")
+  .before(requirePermissions, ["firebasevertexai.config.get", "serviceusage.services.get"])
+  .action(async (path: string | undefined, options: Options) => {
+    const projectId = needProjectId(options);
+
+    if (!(await ailogic.isAILogicApiEnabled(projectId))) {
+      logger.info("Firebase AI Logic is not enabled on this project.");
+      return;
+    }
+    const config = await ailogic.getConfig(projectId);
+
+    const authOnly = config.trafficFilter?.firebaseAuthRequired ?? false;
+    const templateOnly = config.trafficFilter?.templateOnly ?? false;
+    const monitoringState = config.telemetryConfig?.mode === "ALL";
+    const sampleRatePercent =
+      config.telemetryConfig?.samplingRate !== undefined
+        ? Math.round(config.telemetryConfig.samplingRate * 100)
+        : 100;
+
+    const enabledProviders = await ailogic.listProviders(projectId);
+    const hasDeveloperApi = enabledProviders.includes("gemini-developer-api");
+    const hasAgentPlatform = enabledProviders.includes("agent-platform-gemini-api");
+
+    const configObj = {
+      providers: {
+        "gemini-developer-api": hasDeveloperApi,
+        "agent-platform-gemini-api": hasAgentPlatform,
+      },
+      security: {
+        "auth-only": authOnly,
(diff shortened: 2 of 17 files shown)


src/commands/index.ts
```diff
@@ -224,6 +224,20 @@ export function load(client: CLIClient): CLIClient {
...
+    client.ailogic = {};
+    client.ailogic.providers = {};
+    client.ailogic.providers.enable = loadCommand("ailogic-providers-enable");
+    client.ailogic.providers.disable = loadCommand("ailogic-providers-disable");
+    client.ailogic.providers.list = loadCommand("ailogic-providers-list");
+  }
+
+  client.ailogic.config = {};
+  client.ailogic.config.get = loadCommand("ailogic-config-get");
+  client.ailogic.config.set = loadCommand("ailogic-config-set");
```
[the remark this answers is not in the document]
marb2000: Good catch, fixed. Moved the `ailogic:config` registration inside the `if (experiments.isEnabled("ailogic"))` block. A rebase had left it outside, which crashed startup when the experiment is off. Verified: with the experiment off, `firebase --help` runs cleanly and the commands are absent; with it on, they register.

src/gcp/ailogic.ts
```diff
@@ -191,10 +211,357 @@ export async function upsertBlockingFunction(endpoint: AILogicEndpoint): Promise
...
+    true,
+  );
+  // aiplatform.googleapis.com cannot be enabled without billing (the Blaze plan),
+  // so an enabled Vertex API already implies the agent-platform provider is available.
+  if (isVertexEnabled) {
+    enabled.push("gemini-agent-platform-api");
+  }
+
+  return enabled;
+}
```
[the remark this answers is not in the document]
marb2000: Done. `listProviders` now runs the two independent enablement checks with `Promise.all`.

src/commands/ailogic-config-set.ts
```diff
@@ -0,0 +1,120 @@
...
+  "monitoring.sample-rate-percentage",
+];
+
+/** Parses a "true"/"false" flag value, throwing a FirebaseError otherwise. */
+function parseBool(pathStr: string, value: string): boolean {
+  if (value !== "true" && value !== "false") {
+    throw new FirebaseError(`Value for ${clc.bold(pathStr)} must be 'true' or 'false'.`);
+  }
+  return value === "true";
+}
```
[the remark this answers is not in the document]
marb2000: Done. `parseBool` now lowercases the value first, so `True`/`FALSE` are accepted. Added a case-insensitive test.

src/commands/ailogic-config-get.ts
```diff
@@ -0,0 +1,84 @@
...
+  ...ailogic.WRITABLE_CONFIG_PATHS.filter((p) => p.startsWith("monitoring.")),
+];
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === "object" && value !== null;
+}
+
+export const command = new Command("ailogic:config:get [path]")
+  .description("read AI Logic configuration")
+  .before(requirePermissions, ["firebasevertexai.config.get", "serviceusage.services.get"])
```
joehan, line 37: Please add detailed help text explaining how to use this command using the `.help()` method. Would be particularly nice to explain what values are allowed for [path]

src/commands/ailogic-config-set.ts
```diff
@@ -0,0 +1,133 @@
...
+      // Unreachable behind assertKnownConfigPath; kept exhaustive so a newly added
+      // writable path cannot silently fall into the wrong branch.
+      throw new FirebaseError(`Unknown configuration path: ${pathStr}`);
+  }
+}
+
+export const command = new Command("ailogic:config:set <path> <value>")
+  .description("set one configuration value")
+  .option("-f, --force", "bypass confirmation prompt")
+  .before(requirePermissions, [
```
joehan, line 116: Please add detailed help text explaining how to use this command using the .help() method. Would be particularly nice to explain what values are allowed for [path]

src/commands/ailogic-providers-disable.ts
```diff
@@ -0,0 +1,31 @@
...
+import { FirebaseError } from "../error";
+import { confirm } from "../prompt";
+
+import { Options } from "../options";
+
+export const command = new Command("ailogic:providers:disable <providerType>")
+  .description("disable a Gemini API provider service")
+  .option("-f, --force", "bypass confirmation prompt")
+  .before(requirePermissions, ["serviceusage.services.disable", "firebasevertexai.config.update"])
+  .action(async (providerType: string, options: Options) => {
```
joehan, line 28: Please add detailed help text explaining how to use this command using the .help() method. Would be particularly nice to explain what values are allowed for providerType

src/commands/ailogic-providers-enable.ts
```diff
@@ -0,0 +1,18 @@
...
+import * as ailogic from "../gcp/ailogic";
+import * as clc from "colorette";
+import { logger } from "../logger";
+
+import { Options } from "../options";
+
+export const command = new Command("ailogic:providers:enable <providerType>")
+  .description("enable a Gemini API provider service")
+  .before(requirePermissions, ["serviceusage.services.enable", "firebasevertexai.config.update"])
+  .action(async (providerType: string, options: Options) => {
```
joehan, line 25: Please add detailed help text explaining how to use this command using the .help() method. Would be particularly nice to explain what values are allowed for <providerType>

CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+- Added `firebase ailogic:providers:*` CLI commands to enable, disable, and list Gemini API providers.
```
joehan: We can omit the changelog while we are developing behind a flag - we'll add Changelogs for all when we flip the falg

joehan (approved):
LGTM overall, with some simple nits

joehan:
One other note - please move these PRs over to a branch on the main repo. If you don't have access, go/firebase-github-access

src/gcp/ailogic.ts
```diff
@@ -191,10 +210,323 @@ export async function upsertBlockingFunction(endpoint: AILogicEndpoint): Promise
   }
 }
 
+/**
```
joehan: Remove empty comment blocks - agents love adding these for some reason

src/gcp/ailogic.ts
```diff
@@ -191,10 +210,323 @@ export async function upsertBlockingFunction(endpoint: AILogicEndpoint): Promise
...
+        validPaths.map((p) => `  ${p}`).join("\n"),
+    );
+  }
+}
+
+/**
+ * Converts the API's telemetry sampling rate, a fraction in (0,1], to the
+ * integer percentage (1-100) the CLI exposes.
+ */
+export function samplingRateToPercent(samplingRate: number): number {
```
joehan: Super overkill to make this a function when it is only used once. Same with the inverse function. Also, no need to export functions that are not used elsewhere.

src/commands/ailogic-config-get.ts
marb2000: Done. Added help text listing all the readable paths (built from the path list so it stays in sync) plus an example.

src/commands/ailogic-config-set.ts
marb2000: Done. Added help text listing each writable path with the values it accepts.

src/commands/ailogic-providers-disable.ts
marb2000: Done. Added help text listing the valid provider types.

src/commands/ailogic-providers-enable.ts
marb2000: Done. Added help text listing the valid provider types, noting the Blaze plan requirement for the agent platform.

CHANGELOG.md
marb2000: Done. Removed the changelog entries; will add them back when the flag flips.

src/gcp/ailogic.ts
marb2000: Done. Removed both empty comment blocks. They had been added to silence a jsdoc lint warning, but it is warn-level only, so removing them is harmless.

src/gcp/ailogic.ts
marb2000: Done. Removed both functions and inlined the conversions at their single call sites.

marb2000:
FYI: I found and fixed a provider-state inconsistency bug on the providers PR (details in https://github.com/firebase/firebase-tools/pull/10778#issuecomment-5087396123). This branch picks the fix up via a merge commit.

marb2000:
Closing in favor of #10850, which moves this to a branch on the main repo as @joehan requested. The branch is identical (all review nits addressed, and it carries the provider-state consistency fix via the providers branch). @christhompsongoogle FYI. Approvals do not carry over, so a fresh stamp on #10850 would be appreciated.

PR #10842, opened 2026-07-23
feat: add firebase.json schema support and validation for function kits

wandamora:
### Description

Adds comprehensive support for Function Kits in `firebase.json` (`kit` stanza as a peer to `codebase`), including configuration parsing, validation, JSON schema updates, deployment handling, and emulator / local serving integration behind the `kits` experiment flag.

Specifically, this PR:
- **Config & Validation (`src/firebaseConfig.ts`, `src/functions/projectConfig.ts`)**:
  - Adds `KitSourcePackage` and `KitFunctionConfig` types, updating the `FunctionConfig` union.
  - Adds `ValidatedKitSingle` type and validation logic in `src/functions/projectConfig.ts` enforcing:
    - Mutual exclusivity between `kit` and `codebase` / `remoteSource` / `prefix`.
    - Required `instances` map mapping instance IDs to configuration subdirectories (`configDir`).
    - Project-wide kit name and kit instance ID uniqueness across kit stanzas.
    - Mutual exclusivity between standard codebase names and kit instance IDs.
    - Format and character validation for kit names and instance IDs.
  - Updates `isLocalConfig` and `requireLocal`, and adds `isKitConfig` type guards.
  - Adds `configForCodebase` helper for resolving configurations by standard codebase name or kit instance ID.
- **Schema & Experiments (`schema/firebase-config.json`, `src/experiments.ts`)**:
  - Updates generated JSON schema (`schema/firebase-config.json`) to include `kit` stanzas with instance mapping.
  - Registers the `kits` experiment flag (`ALL_EXPERIMENTS.kits`).
- **Deployment Support & Kit Instance Prefixing (`src/deploy/functions/`)**:
  - Treats kit instances as codebases during deployment (`functionsDeployHelper.ts` and `prepare.ts`).
  - Sets the prefix for every kit instance backend to `kit-<instance-id>`, automatically applying `kit-<instance-id>-` to all function endpoint IDs during discovery in `prepare.ts` to isolate instances and prevent function name collisions.
  - Updates function selector parsing (`parseFunctionSelector`) and target codebase resolution (`targetCodebases`) to recognize kit instance IDs alongside codebase names.
- **Emulator & Local Serving (`src/emulator/controller.ts`, `src/serve/functions.ts`)**:
  - Adds support for kit backends in the Functions emulator and `serve` commands by instantiating an `EmulatableBackend` for each kit instance with its assigned `configDir` and `kit-<instance-id>` prefix.

### Scenarios Tested

- `firebase deploy --only functions` with a mix of standard `codebase` stanzas and `kit` stanzas
- `firebase deploy --only functions:<instance-id>`
- `firebase deploy --only functions:<instance-id>:<function-name>`


src/functions/projectConfig.ts
```diff
@@ -177,13 +211,44 @@ function assertUniqueSourcePrefixPair(config: ValidatedConfig): void {
...
+    if ("instances" in c && c.instances) {
+      for (const instanceId of Object.keys(c.instances)) {
+        if (instanceIds.has(instanceId)) {
+          throw new FirebaseError(
+            `functions kit instance ID must be unique across all kits, but '${instanceId}' was used more than once.`,
+          );
+        }
+        instanceIds.add(instanceId);
+      }
+    }
```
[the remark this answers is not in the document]
wandamora: I've added checks since then to check if the instance id is less than 40 characters.

src/functions/projectConfig.ts
```diff
@@ -177,13 +214,44 @@ function assertUniqueSourcePrefixPair(config: ValidatedConfig): void {
...
+    if ("instances" in c && c.instances) {
+      for (const instanceId of Object.keys(c.instances)) {
+        if (instanceIds.has(instanceId)) {
+          throw new FirebaseError(
+            `functions kit instance ID must be unique across all kits, but '${instanceId}' was used more than once.`,
+          );
+        }
+        instanceIds.add(instanceId);
+      }
```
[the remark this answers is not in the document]
wandamora: instance ids have different requirements, such as a shorter character limit. this is taken care of.

src/deploy/functions/prepare.ts
shettyvarun268: Not a part of your diff, but looks like for kit configs, configDir isn't on the top level of localCfg. It's nested inside localCfg.instances[codebase].

Right now proto.convertIfPresent looks for localCfg.configDir, so userEnvOpt.configDir ends up undefined. This means .env files inside a kit instance's config folder won't get loaded when running firebase deploy. (Looks like controller.ts and serve/functions.ts handle this for the emulator using localCfg.instances).

wandamora: This was a good catch, thank you :)

schema/firebase-config.json
```diff
@@ -909,6 +912,102 @@
             },
             "type": "object"
         },
+        "KitFunctionConfig": {
+            "additionalProperties": false,
+            "properties": {
+                "codebase": {
+                    "type": "string"
+                },
+                "configDir": {
```
ajperel: Why do we need configDir here and prefix below? Don't we want to generally get rid of those?

src/deploy/functions/functionsDeployHelper.ts
```diff
@@ -59,18 +59,27 @@ export function endpointMatchesFilter(endpoint: backend.Endpoint, filter: Endpoi
...
-      return [{ codebase: fragments[0] }];
+  const target = fragments[0];
+
+  // Check if target matches a known codebase name or kit instance ID
+  const codebaseNames = config.flatMap((c) => {
+    if (isKitConfig(c)) {
+      return c.instances ? Object.keys(c.instances) : [];
     }
-    // It's not a codebase name, assume it is a function id in default codebase
+    return c.codebase ? [c.codebase] : [];
```
ajperel: Do you need to check c.codebase existence?

Either it should be a codebase or kit and you've already covered the kit case right?

If so I think this can just be `return c.codebase`?

src/deploy/functions/functionsDeployHelper.ts
```diff
@@ -158,26 +167,29 @@ export function getFunctionLabel(fn: backend.TargetIds & { codebase?: string }):
  * Returns list of codebases specified in firebase.json filtered by --only filters if present.
  */
 export function targetCodebases(config: ValidatedConfig, filters?: EndpointFilter[]): string[] {
-  const codebasesFromConfig = [...new Set(Object.values(config).map((c) => c.codebase))];
+  const codebasesFromConfig = [
+    ...new Set(
+      config.flatMap((c) => {
```
ajperel: You had this exact flatMap above. Should it be a method?

src/emulator/controller.ts
```diff
@@ -542,22 +541,40 @@ export async function startAll(
...
-      proto.convertIfPresent(backend, localCfg, "configDir", (cd) => path.join(projectDir, cd));
-      emulatableBackends.push(backend);
+
+      if (isKitConfig(localCfg)) {
+        for (const [instanceId, configDir] of Object.entries(localCfg.instances)) {
+          const backend: EmulatableBackend = {
+            functionsDir,
+            runtime,
+            codebase: instanceId,
+            prefix: `kit-${instanceId}`,
```
ajperel: we hardcoded this `kit-` prefix in prepare.ts as well. Do you think it should be a shared constant?

though I guess it's the kit prefix prefix 😂

src/functions/projectConfig.ts
```diff
@@ -61,6 +75,18 @@ export function validateCodebase(codebase: string): void {
   }
 }
 
+/**
+ * Check that the kit name is 40 characters or less and only contains allowed characters.
```
ajperel, line 79: called kit name here but kitId in method name we should be consistent is it a name or an id?  which do we use for codebase?

src/functions/projectConfig.ts
```diff
@@ -177,13 +239,60 @@ function assertUniqueSourcePrefixPair(config: ValidatedConfig): void {
...
+ * Validate each instance ID format and ensure instance IDs are unique across all kit stanzas in the project.
+ */
+export function validateKitInstances(
+  instances: Record<string, string>,
+  allProjectInstanceIds: Set<string>,
+): void {
+  for (const instanceId of Object.keys(instances)) {
+    if (instanceId.length === 0 || instanceId.length > 40 || !/^[a-z0-9_-]+$/.test(instanceId)) {
+      throw new FirebaseError(
+        `Invalid kit instance ID '${instanceId}'. Instance ID must be 40 characters or less and ` +
```
ajperel: do we want to allow instance ids to start with any character? Technically it should be safe since we prefix them with 'kit-'? Like gcf functions I think must start with a letter and must end with a letter or a number.

ajperel (commented):
Very nice! A few comments but all pretty minor.

ajperel (commented):
I don't think this is new to your PR, but Gemini also suggests we validate that configDir resolves to something in the project. 

There's probably not a good reason to allow something like: "inst1": "../../../../some/other/repo/config"

schema/firebase-config.json
wandamora: This comment made me realize that inheriting `FunctionConfigBase` as is were causing these unwanted properties to be in the generated schema. I refactored `firebaseConfig.ts` a bit to separate these out of `KitFunctionConfig`.

src/functions/projectConfig.ts
wandamora: If we really want to follow the `codebase` pattern, it calls it the "name" and doesn't include that in the function name. I renamed this to `validateKit` to better copy `validateCodebase` above.

src/functions/projectConfig.ts
wandamora: I'm inclined to say it's not necessary since we add a the `kit-` prefix. I think I'm not too worried about the ending of an instance name either, since that is also treated like a prefix to the function name.

src/emulator/controller.ts
wandamora: True, I made this into a shared constant :)

src/deploy/functions/functionsDeployHelper.ts
wandamora: Good observation, I changed this to just return codebase into the new helper function.

src/deploy/functions/functionsDeployHelper.ts
wandamora: Created `getCodebasesFromConfig`.

src/functions/projectConfig.ts
ajperel: I'm tempted to at least say the shouldn't being or end with '-' just so we don't end up with a confusing state like 'kit--myInstance--function' not that 'kit-\_myInstance\_-function' is that great either but at least a little clearer. But I could be convinced to not be fussed about this.

wandamora: I added the additional check since it's trivial.

src/functions/projectConfig.ts
ajperel: Nice fix for method name.... agree it's "codebase" name... ironically our documentation seems to just talk about the codebase property and avoid saying name or id altogether 😂  ... I'm good sticking with name for consistency.

src/experiments.ts
```diff
@@ -90,6 +90,11 @@ export const ALL_EXPERIMENTS = experiments({
     public: true,
     default: false,
   },
+  kits: {
+    shortDescription: "Enable Functions Kits.",
+    fullDescription: "Adds support for Function Kits schema and deployment features.",
+    public: true,
```
ajperel: I don't know if we need this to be public yet?

ajperel (approved):
Looks good and if there are subtle issues .... it's behind an experiment and we'll keep iterating.

src/experiments.ts
wandamora: Oh this should definitely be private. Overlooked jetski change 😅

PR #10845, opened 2026-07-24
[FSQL] Prevent client socket errors from escalating to pglite server crashes

mtr002:
### Description

This PR fixes Issue#10821

Cancelling an in-flight Data Connect request makes the Go backend Postgres driver send a CancelRequest frame on a new TCP connection. pg-gateway throws Unexpected initial message, which pgliteServer.ts currently re-emits as a server error—causing dataconnectEmulator.ts to call cleanShutdown() and stop all emulators.

This PR decouples client socket errors from server-level errors in pgliteServer.ts. Per-socket errors are now logged and destroyed so single connection drops don't crash the emulator stack. Real server errors (EADDRINUSE) are still handled normally.

### Scenarios Tested

* Sent CancelRequest packet to port while emulators were running; verified stack stayed alive and served subsequent GraphQL queries.
* Verified reproduction crash on npx firebase-tools@latest vs survival on local fix.
* Ran full emulator test suite and npm run lint:changed-files.

### DIFF
--- a/src/emulator/dataconnect/pgliteServer.ts
+++ b/src/emulator/dataconnect/pgliteServer.ts
@@ -79,8 +79,14 @@ export class PostgresServer {
       socket.on("end", () => {
         logger.debug("Postgres client disconnected");
       });
+      // Log and destroy individual client socket errors instead of escalating them
+      // to server-level errors.
       socket.on("error", (err) => {
-        server.emit("error", err);
+        logger.debug(
+          "Postgres client socket error (if this happened during request cancellation, it is expected and harmless):",
+          err,
+        );
+        socket.destroy();
       });
     });
     this.server = server;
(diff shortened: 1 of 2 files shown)


src/emulator/dataconnect/pgliteServer.ts
```diff
@@ -79,8 +79,11 @@ export class PostgresServer {
       socket.on("end", () => {
         logger.debug("Postgres client disconnected");
       });
+      // Log and destroy individual client socket errors instead of escalating them
+      // to server-level errors.
       socket.on("error", (err) => {
-        server.emit("error", err);
+        logger.debug("Postgres client socket error (expected during request cancellation):", err);
```
yuchenshi: I think this is fine but the logs can be a bit misleading if it happens outside of cancellation. Maybe add as a newline suffix e.g. "If this happened during request cancellation, it's expected and harmless"

yuchenshi (approved):

yuchenshi:
Please add an CHANGELOG entry

src/emulator/dataconnect/pgliteServer.ts
mtr002: Good point. Done

PR #10851, opened 2026-07-27
feat: Add Eventarc deployment for Dart Cloud Run functions

Lyokone:
## Summary
Adds production deployment support for Eventarc-backed Dart functions running directly on Cloud Run.
- Discover and reconcile Firebase-managed Dart Cloud Run services alongside GCF functions
- Create, update, replace, roll back, and delete Eventarc triggers
- Support Firestore CloudEvents with protobuf payloads
- Configure Eventarc receiver and Cloud Run invoker IAM roles
- Preserve service identity and deployment metadata across updates
- Validate unsupported event types and retry policies before deployment
- Handle legacy Dart CloudEvent services and partial deployment failures


Ortes:
We independently hit all three defects this PR fixes, and can confirm the approach works against a real project.

**Reproduced on `firebase-tools` 15.30.1**, deploying a Dart codebase (`platform === "run"`):

1. `setTrigger()` has no event-trigger branch, so no Eventarc trigger is created — and unlike the schedule/task-queue/blocking cases it doesn't throw, so the deploy reports success.
2. `createRunFunction()` grants no `run.invoker` to the delivery service account.
3. `serviceFromEndpoint()` doesn't persist the trigger type, so the second deploy fails with `Changing from an HTTPS function to a background triggered function is not allowed` and aborts the whole codebase.

**This PR's branch, deployed to a staging project**, produced a working Firestore trigger: create/update/delete all delivered with wildcard params and the protobuf payload decoded, the second deploy passed, and removing the function cleaned up both the service and the trigger with no orphan Pub/Sub topic. Storing `eventTrigger` in the existing `firebase-functions-metadata` annotation is what makes the read-back work, and it resolves the `// TODO: Figure out how to encode all trigger types...` in `runv2.ts` without adding a new annotation — the GCF `trigger-type` one being read-only.

**The trailing-hyphen issue `gemini-code-assist` raised is real**, not theoretical. A 63+ char id sliced at 54 ending in `-` gives `firebase-ondocumentwritten-dart-trigger-probe-probeid--5e1cac9a`. Trimming trailing hyphens before appending the hash fixes it.

**Rebasing onto current `main` is less clean than it looks.** Only two textual conflicts, but three things then fail to compile: `#11047` dropped `DEFAULT_RETRY_CODES` and replaced `RunOptions.retryCodes` with `retryPredicates`, and the auto-merge silently produces duplicate `runServiceId` keys in `runv2.spec.ts` where `#10916` already added the property.

We're running a rebased build with those fixes and the hyphen fix, tests green. Happy to open it as a PR against this branch if that's useful, @Lyokone.

PR #10853, opened 2026-07-27
feat(ailogic): add ailogic:templates CLI commands

marb2000:
### Description

Adds Firebase AI Logic server prompt template management to the CLI, gated behind the `ailogic` experiment (same as `ailogic:providers` and `ailogic:config`). Builds on the merged #10849 (providers) and #10850 (config).

**Deploying templates goes through `firebase deploy`**, like every other declared-state product (per review feedback, replacing an earlier standalone `ailogic:templates:deploy` command):

```jsonc
// firebase.json
"ailogic": { "templates": "prompts" }
```

- `firebase deploy --only ailogic` (or `--only ailogic:templates`; unknown filters are an error) reconciles the prompts directory with the project: new `.prompt` files are created, changed files are updated, unchanged files are skipped, and templates whose files were removed are deleted after confirmation (`--force` in CI). An empty directory is refused as "delete everything". `--dry-run` shows the plan without writing.
- The directory is read recursively; a nested file's relative path flattens into its id with `/` becoming `.` (`agents/support.prompt` → `agents.support`, since template ids are single URL segments). Two files mapping to one id is a validation error naming both.
- The target registers like any other but asserts the `ailogic` experiment in prepare (mirroring the `webframeworks` precedent), so nothing is reachable while the API is pre-approval.

**Management commands** for cloud-side inspection and one-off mutations:

- `ailogic:templates:list` — id, display name, lock state, content preview.
- `ailogic:templates:get <templateId>` — prints one template in full.
- `ailogic:templates:delete <templateId>` — deletes after confirmation (`--force` to skip).
- `ailogic:templates:lock` / `unlock <templateId>` — a lock blocks updates and deletes, including by deploy; `--force` never overrides a lock.

Design notes (concurrency behavior verified against the live v1beta API, not assumed):

- Deploy planning is a pure function (`planTemplateDeploy` in `src/ailogic/templates.ts`): local files plus remote state in, creates/updates/unchanged/deletes/lockedViolations out (mirrors the functions release planner). Prepare does all reads, validation, planning, and the prune confirmation; release does all writes, so nothing mutates until every target's prepare has passed.
- Lock and unlock use the dedicated `:modifyLock` RPC; `locked` is output-only on the resource, and the server enforces locks on writes (`FAILED_PRECONDITION`), so the CLI-side checks are a fail-fast nicety on top of real server enforcement. A locked template with *unchanged* content does not block a deploy (nothing would be written).
- Writes carry etag preconditions: `templates:delete` passes the etag from its pre-confirmation read, and deploy passes each template's listed etag on updates and deletes. A template modified between planning and apply fails with a clear "re-run" message (409) instead of being silently overwritten or deleted.
- Deploy updates are masked to `templateString` (an unmasked PATCH replaces the whole resource), so a `displayName` set outside the CLI survives deploys; `displayName` defaults to the template id on creates only.

### Scenarios Tested

- Unit tests: `src/ailogic/templates.spec.ts` (validator, recursive directory reader against real temp dirs, planner), `src/deploy/ailogic/ailogic.spec.ts` (experiment gate, filters, full prepare+release flows, conflict and 404 handling), and one spec per management command. 1714 tests pass repo-wide; `npm run build`, lint, and `tsc --noEmit` are clean.
- Live end-to-end against a real project: experiment gate error, `--dry-run` planning, creates including nested dotted ids, unchanged idempotent re-deploy ("already up to date"), `--only ailogic:templates` sub-target, deletion of removed files with `--force`, a locked template blocking a real deploy during planning, and a `displayName` set outside the CLI surviving a content deploy.
- Runtime smoke test: CLI startup with the experiment off (commands absent, `--only ailogic` fails with the experiment error) and on (commands registered, deploy target active).

### Sample Commands

```
firebase experiments:enable ailogic

firebase deploy --only ailogic
firebase deploy --only ailogic:templates --force
firebase deploy --only hosting,functions,ailogic
firebase deploy --only ailogic --dry-run

firebase ailogic:templates:list
firebase ailogic:templates:get welcome
firebase ailogic:templates:lock welcome
firebase ailogic:templates:unlock welcome
firebase ailogic:templates:delete welcome --force
```


marb2000:
Heads up on a fix pushed in aacf0a53: while verifying the API's concurrency semantics against a live project, I found that 'locked' is output-only on PromptTemplate, so the original lock/unlock implementation (PATCH with updateMask=locked) was rejected by the server with 400 INVALID_ARGUMENT. They now use the dedicated POST :modifyLock RPC.

Same commit adds etag preconditions, since the API turned out to fully enforce them: templates:delete passes the etag from its pre-confirmation read, and deploy passes each template's listed etag on updates and prunes. A template modified between planning and apply now fails with a clear re-run message (409) instead of being silently overwritten.

All six commands were exercised end to end against a real project, including lock enforcement on deploy/delete (the server rejects writes to locked templates with FAILED_PRECONDITION, so the CLI-side checks are a fail-fast nicety on top of real server enforcement).

src/commands/ailogic-templates-list.ts
```diff
@@ -0,0 +1,59 @@
...
+      const templateString = t.templateString ?? "";
+      const preview =
+        templateString.length > PREVIEW_LENGTH
+          ? templateString.substring(0, PREVIEW_LENGTH - ELLIPSIS.length) + ELLIPSIS
+          : templateString;
+      table.push([
+        clc.bold(ailogic.templateIdFromName(t.name)),
+        t.displayName || "",
+        t.locked ? "Yes" : "No",
+        preview.replace(/\n/g, " "),
```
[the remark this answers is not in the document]
marb2000: Good catch, fixed. The preview now strips \r?\n, and there is a test asserting no carriage return survives into the table output.

src/ailogic/templates.ts
```diff
@@ -0,0 +1,147 @@
...
+  const close = FRONTMATTER_CLOSE.exec(rest);
+  if (!close) {
+    return "Frontmatter block is not closed (missing terminating '---').";
+  }
+  const frontmatter = rest.slice(0, close.index);
+  try {
+    const parsed = yaml.load(frontmatter);
+    if (parsed !== undefined && parsed !== null && typeof parsed !== "object") {
+      return "Frontmatter must be a YAML mapping.";
+    }
```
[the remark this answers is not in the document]
marb2000: Done. validatePromptFile now rejects arrays with an Array.isArray check, and the frontmatter test covers a YAML sequence.

src/gcp/ailogic.ts
```diff
@@ -198,3 +214,433 @@ export async function deleteBlockingFunction(endpoint: AILogicEndpoint): Promise
...
+  let pageToken: string | undefined;
+  const templates: Template[] = [];
+
+  do {
+    const queryParams: Record<string, string> = pageToken ? { pageToken } : {};
+    const res = await client.get<ListTemplatesResponse>(`${parent}/templates`, { queryParams });
+    if (res.body.templates) {
+      templates.push(...res.body.templates);
+    }
+    pageToken = res.body.nextPageToken;
```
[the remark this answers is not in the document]
marb2000: Done, switched to optional chaining on the list response body.

joehan:
Seems like all the previous changes are showing up in the PR as well - can you go submit those and sync this to main, and that should get us to a cleaner diff to review?

marb2000:
Done: #10849 and #10850 are submitted and this branch is synced to main, so the diff is now the template-only changes (16 files, pure additions). Ready for review.

src/commands/ailogic-templates-deploy.ts
```diff
@@ -0,0 +1,197 @@
...
+import { confirm } from "../prompt";
+
+import { Options } from "../options";
+
+interface DeployOptions extends Options {
+  dir?: string;
+  prune?: boolean;
+}
+
+export const command = new Command("ailogic:templates:deploy")
```
joehan: High level feedback on the shape of this command:

I'm wondering if this should actually be properly part of `firebase deploy`. For most of our products, we use the single `firebase deploy` command so that our users can treat all the resources in their app as a single deployable unit. For the odd cases where they want to deploy just a single piece at a time, we have the --only flag (ie`firebase deploy --only firestore'). The deploy command has the same semantics as this one - it diifs what you have locally with what is deployed, and determines that API calls to make to reconcile them. 

Is there a clear reason that we should treat ALF as a special case?

src/commands/ailogic-templates-list.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { Options } from "../options";
+
+const PREVIEW_LENGTH = 60;
+const ELLIPSIS = "...";
+
+export const command = new Command("ailogic:templates:list")
+  .description("list deployed templates")
+  .help(
+    `lists every deployed server prompt template with its id, display name, lock state, and a preview of its content. Use ailogic:templates:get <templateId> to print a template in full.`,
```
joehan: Nit - wrap command text in backticks 
```suggestion
    `lists every deployed server prompt template with its id, display name, lock state, and a preview of its content. Use `ailogic:templates:get <templateId>` to print a template in full.`,
```
Applies throughout

src/commands/ailogic-templates-lock.ts
```diff
@@ -0,0 +1,36 @@
...
+import * as ailogic from "../gcp/ailogic";
+import * as clc from "colorette";
+import * as utils from "../utils";
+
+import { Options } from "../options";
+
+export const command = new Command("ailogic:templates:lock <templateId>")
+  .description("lock a template")
+  .help(
+    `locks one deployed server prompt template. A locked template cannot be updated or deleted (including by ailogic:templates:deploy) until it is unlocked with ailogic:templates:unlock.
```
joehan:
```suggestion
    `locks one deployed server prompt template. A locked template cannot be updated or deleted (including by `ailogic:templates:deplo`y) until it is unlocked with `ailogic:templates:unlock`.
```

src/commands/ailogic-templates-lock.ts
```diff
@@ -0,0 +1,36 @@
...
+
+<templateId> is the template's id, as shown in ailogic:templates:list.
```
joehan:
```suggestion
<templateId> is the template's id, as shown in `ailogic:templates:list`.
```

src/commands/ailogic-templates-lock.ts
```diff
@@ -0,0 +1,36 @@
...
+
+For example:
+
+  firebase ailogic:templates:lock my-template`,
```
joehan:
```suggestion
  `firebase ailogic:templates:lock my-template`,
```

src/commands/ailogic-templates-deploy.ts
marb2000: I thought about it and you're right, you've convinced me. The semantics here are exactly deploy semantics: diff the declared local state against what is live, and reconcile. No reason to treat AI Logic as a special case.

So I'm moving this into the deploy command, with all the consequences:

- firebase.json gets a new section: `"ailogic": { "templates": "prompts" }`
- `firebase deploy --only ailogic` deploys the templates. `--only ailogic:templates` also works, so triggers can join later the same way firestore:rules and firestore:indexes did
- `ailogic:templates:deploy` goes away. The other template commands (list, get, delete, lock, unlock) stay
- behavior stays the same under the hood: validation before any API call, confirmation before pruning, locked templates block the deploy, unchanged files are skipped
- subfolders inside prompts/ work: `agents/support.prompt` becomes the id `agents.support`, and two files ending up with the same id is an error
- still behind the ailogic experiment

I'll also fix the backticks in the help text. New code coming shortly.

marb2000: Done and pushed in b11dda0e8. `firebase deploy --only ailogic` (and `--only ailogic:templates`) is live in this PR exactly as described above: firebase.json section, recursive prompts dir with dotted ids, all the guard behavior carried over, and `ailogic:templates:deploy` is gone. Verified end to end against a live project, including --dry-run and a locked template blocking a real deploy during planning. The PR description is updated to match.

src/commands/ailogic-templates-list.ts
marb2000: Done, backticks added throughout the help texts.

src/commands/ailogic-templates-lock.ts
marb2000: Done. Also updated the wording since ailogic:templates:deploy no longer exists: it now says 'including by `firebase deploy`'. (Fixed the stray backtick placement from the suggestion while applying.)

src/commands/ailogic-templates-lock.ts
marb2000: Done.

src/commands/ailogic-templates-lock.ts
marb2000: Done.

joehan (changes requested):
Overall, this is looking great & code is clean, thank you for moving this over into firebase deploy.

The one remaining issue I see is that  AFAICT, we are not hiding the 'deploy AI Logic templates' logic behind the experiment flag anymore. The simplest way to handle this is probably to add a .before() check to the command in src/commands/deploy that checks if AI logic is in the filtered targets. If it is and the expeirement is not enabled, we should just throw an error right there saying something like 'AI Logic deployment is experimental. Please run 'firebase experiments:enable ailogic' to use this feature'

marb2000:
The gate was still there, just not where you looked: the first line of the ailogic target's prepare() was `experiments.assertEnabled("ailogic", ...)`, so `firebase deploy --only ailogic` with the experiment off already failed with the enable instructions. I tested that path against a live project.

That said, you're right that the command level is a better place for it. Targets prepare in order and ailogic went last, so the error fired after every other target already did its prepare work. Moved it in b9a8f750d: the deploy command's target-filtering hook now asserts the experiment whenever ailogic is in the filtered targets, so it fails immediately, before the permissions check and before any target runs. Branch is also synced with main.

joehan (approved):

PR #10856, opened 2026-07-27
feat: automatically enable non-interactive mode for AI agents

joehan:
This PR automatically sets options.nonInteractive = true when an AI agent is detected in the environment. This makes the CLI more usable for agents by failing fast instead of hanging on prompts.

## Evaluation Results

We ran the complete 51-case evaluation suite in google3 (`firebase_tools_evals`) to compare the baseline (`origin/main`) against this branch:

| Metric | Baseline (`main`) | Branch (`enable-noninteractive`) | Change |
| :--- | :--- | :--- | :--- |
| **Pass Rate** | 51.0% (26/51) | **64.7% (33/51)** | **+13.7%** (+7 cases) |
| **Input Tokens** | 7,489.8k | 10,413.4k | +39.0% |
| **Output Tokens**| 259.8k | 420.6k | +61.9% |
| **Total Tokens** | 7,749.6k | 10,834.0k | **+39.8%** |
| **Total Turns** | 1,553 | 2,197 | +41.5% |
| **Tool Calls** | 684 | 838 | +22.5% |

### Note on Token Usage
Automatically enabling non-interactive mode increases token usage and turns because it prevents the agent from hanging on prompts. Instead of waiting and timing out, the agent receives immediate feedback and can use its remaining turns to actively troubleshoot and find alternative non-interactive flags (e.g. appending `--project` or `--non-interactive` explicitly). This active recovery is what drove the **13.7% increase in pass rate**.


src/prompt.ts
```diff
@@ -23,7 +23,7 @@ export interface BasicOptions<T> {
 export function guard<T>(
   opts: BasicOptions<T>,
 ): { shouldReturn: true; value: T } | { shouldReturn: false; value: undefined } {
-  if (!opts.nonInteractive) {
+  if (!opts.nonInteractive && process.env.NON_INTERACTIVE !== "true") {
```
christhompsongoogle: comparing to "true" is safe - but a little ugly. Can we infer the truth-y value just by checking the variable is set? I suppose one could set it to "false" but is that valid, vs `unset NON_INTERACTIVE`

christhompsongoogle: should we check if these disagree? I think based on the change in command.ts they should always match but in theory this function could be called with a mismatch. If they're expected to match always by design we should assert that here

src/dataconnect/types.ts
```diff
@@ -249,16 +249,17 @@ export function toDatasource(
...
-    return {
-      postgresql: {
-        database: ds.postgresql.database,
-        schema: ds.postgresql.schema,
-        cloudSql: {
-          instance: `projects/${projectId}/locations/${locationId}/instances/${ds.postgresql.cloudSql.instanceId}`,
-        },
-        schemaValidation: ds.postgresql.schemaValidation,
-      },
+    const postgresql: PostgreSql = {
```
christhompsongoogle: unrelated? could move to second PR

christhompsongoogle (approved):

src/dataconnect/types.ts
joehan: Oh wtf, this definitely was not intended to be part of the PR. removing

src/prompt.ts
joehan: Honestly, probably should just remove this env var functionality, it was added for testing mostly.

joehan: We need a global non-interactive safeguard because many commands (e.g. `init`) invoke prompt wrappers without explicitly passing the `nonInteractive` option. Without a global flag, these prompts try to go interactive and hang in non-interactive environments.

However, using `process.env` is indeed not ideal and prone to test pollution. I have refactored this to use a module-level global variable in `src/prompt.ts` (managed via exported `setNonInteractive` helpers) to avoid polluting the environment.

src/dataconnect/types.ts
joehan: I have reverted the changes in `src/dataconnect/types.ts` as they were unintended for this PR.

PR #10864, opened 2026-07-29
Detect the Android app module by basename in apps:init

### LINKED ISSUE #10863: apps:init: Android module detected by first path segment instead of basename, producing app/app output path
### Which CLI command are you experiencing an issue with?

`apps:init` (Android)

### Description

`findIntelligentPathForAndroid` in `src/management/apps.ts` decides "am I already inside the Android app module?" by looking at the **first** path segment instead of the last:

https://github.com/firebase/firebase-tools/blob/master/src/management/apps.ts#L794-L797

```ts
const paths = appDir.split("/");
// For when app/build.gradle is found
if (paths[0] === "app") {
  return appDir;
}
```

The docstring right above it says the intent is the opposite of what the code does:

```
android/build.gradle // if it's this, choose app
android/app/build.gradle // if it's this, choose current dir.
```

For any realistic `appDir` the first segment is not `app` — it is `android` for `android/app`, and `""` for an absolute path like `/home/me/proj/android/app`. So the early return is effectively dead, and correctness falls through to the `src` heuristic below it. When the module has no `src` directory, the function appends `app` a second time and returns a path that does not exist:

| appDir | expected | actual |
| --- | --- | --- |
| `android/app` (contains `src/`) | `android/app` | `android/app` (via the `src` check, not the intended branch) |
| `android/app` (no `src/`, e.g. `libs/` + `build.gradle`) | `android/app` | `android/app/app` |

`getSdkOutputPath` then writes `google-services.json` into that non-existent nested directory.

There is a second, smaller problem in the same line: `split("/")` is separator-dependent, so it can never match on Windows, where `appDir` arrives with `\` after `path.join`.

### Steps to reproduce

Adding this case to the existing `getAndroidPlatform` table in `src/management/apps.spec.ts` reproduces it:

```ts
{
  desc: "Inside app folder without a src directory",
  folderName: "android/app",
  folderItems: { libs: {}, "build.gradle": "" },
  output: "android/app",
},
```

```
1 failing
  Inside app folder without a src directory:
  AssertionError: expected 'android\app\app' to equal 'android/app'
```

### Expected behavior

An `appDir` whose basename is `app` should be treated as the app module and returned as-is, regardless of path separator or how many segments precede it.

### Additional context

Using `path.basename(appDir) === "app"` fixes both the wrong-segment logic and the separator dependence.

Separately, while confirming this on Windows I found that three assertions in `src/management/apps.spec.ts` hardcode forward slashes while the implementation builds paths with `path.join`, so `npm run mocha -- src/management/apps.spec.ts` is red on a Windows checkout (`expected 'android\app' to equal 'android/app'`, and the same for `ios\abc`). Unit tests only run on `ubuntu-latest`/`macos-latest` in `node-test.yml`, so CI doesn't surface it. I'll include that in the PR since it's the same line-of-code area.

### Environment

- firebase-tools: `15.24.0` (current `master`, commit `0636759`)
- Node: v22.18.0
- OS: Windows 11

Zuhef:
## Summary

`findIntelligentPathForAndroid` decides whether the user is already inside the Android app module by testing the **first** segment of the path, when the intent — stated in its own docstring — is the **last** one:

```ts
const paths = appDir.split("/");
// For when app/build.gradle is found
if (paths[0] === "app") {
  return appDir;
}
```

For any realistic input the first segment is not `app`: it's `android` for `android/app`, and `""` for an absolute path like `/home/me/proj/android/app`. The branch therefore never runs, and correctness falls through to the `src` heuristic below it.

## Root cause

Because the early return never fires, an app module that has no `src` directory reaches the fallback that appends `app` a second time, and `getSdkOutputPath` writes `google-services.json` into a directory that does not exist:

| `appDir` | expected | before |
| --- | --- | --- |
| `android/app` (has `src/`) | `android/app` | `android/app` — but via the `src` check, not the intended branch |
| `android/app` (no `src/`) | `android/app` | `android/app/app` |

`split("/")` is also separator-dependent, so the check can never match on Windows, where `appDir` arrives with `\` from `path.join`.

## Fix

Use `path.basename(appDir) === "app"`, which expresses the documented intent directly and is correct for relative paths, absolute paths, trailing separators, and both path separators. Behaviour for the existing cases is unchanged — `android/app` with a `src/` directory still returns `android/app`, just through the branch that was meant to handle it.

## Testing

Added the regression case to the existing `getAndroidPlatform` table. It fails before this change:

```
1 failing
  Inside app folder without a src directory:
  AssertionError: expected 'android\app\app' to equal 'android/app'
```

and passes after it.

```console
$ npx mocha "src/management/apps.spec.ts"
46 passing

$ npx mocha "src/management/*.spec.ts"
97 passing

$ npm run test:compile
(clean)
```

`npx eslint` and `npx prettier --check` are clean on both changed files.

### One extra thing, same lines of code

Three assertions in `apps.spec.ts` hardcoded forward slashes while the implementation builds those values with `path.join`, so `apps.spec.ts` was red on a Windows checkout before this PR — `42 passing, 3 failing`:

```
AssertionError: expected 'android\app' to equal 'android/app'
AssertionError: expected 'ios\abc' to equal 'ios/abc'
```

`node-test.yml` runs unit tests on `ubuntu-latest` and `macos-latest` only (Windows is integration-only), so CI never showed this. I switched those three expectations to `path.join(...)`, which is a no-op on Linux and macOS and makes the suite pass on Windows. The two cases where the function returns `appDir` verbatim keep their literal values, since that string is echoed back unchanged. Happy to split this into its own PR if you'd rather keep the fix isolated.

Fixes #10863


Zuhef:
@googlebot I signed it!

maneesht (approved):

joehan:
Thanks for the contirbution!

PR #10867, opened 2026-07-30
Use operation-specific wording in App Check debug token prompts

Zuhef:
## Description

`getOrPromptAppId` in `src/appcheck/prompts.ts` hardcodes the app-selection prompt:

```ts
const selectedApp = await selectAppInteractively(apps, AppPlatform.ANY, {
  message: "Select the app to register a debug token for:",
});
```

All three App Check debug token commands call it through the `getOrPromptProjectAndAppId` alias — `appcheck:debugtokens:create`, `appcheck:debugtokens:list` and `appcheck:debugtokens:delete`. So when a project has more than one app and the user has not passed `--app`, both of the non-create commands describe the wrong operation:

```
$ firebase appcheck:debugtokens:delete <id>
? Select the app to register a debug token for: (Use arrow keys)
```

That is worst in `delete`, where the prompt is the step that chooses *which app a token gets destroyed in*, and it tells the user it is about to register one.

## Fix

Give `getOrPromptAppId` an optional `message` parameter that defaults to the existing string, and have `list` and `delete` pass their own wording:

* `list` — "Select the app to list debug tokens for:"
* `delete` — "Select the app to delete a debug token from:"

`create` is untouched and its prompt is byte-identical, since it relies on the default.

## Testing

`src/appcheck/prompts.ts` had no spec, so this adds `src/appcheck/prompts.spec.ts` covering the helper's branches:

* uses `options.app` without listing apps
* throws when the project has no apps
* selects the only app without prompting
* throws for multiple apps in non-interactive mode
* narrows the choices to locally detected apps
* prompts with the register wording by default
* prompts with the caller supplied wording

```console
$ npx mocha "src/appcheck/*.spec.ts" "src/commands/appcheck-*.spec.ts"
19 passing
```

That is `12 passing` on an unmodified checkout, so all 7 additions are new coverage. Reverting only the three source files fails the regression test:

```
✖ prompts with the caller supplied wording
  AssertionError: expected { Object (message) } to deeply equal { Object (message) }
```

`npm run test:compile` is clean. `npx eslint` on the four changed files reports no findings other than the CRLF `prettier/prettier` noise my Windows checkout produces in every file (I fixed the one genuine `jsdoc/tag-lines` warning my new doc comment introduced).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Fixed `appcheck:debugtokens:list` and `appcheck:debugtokens:delete` prompting "Select the app to register a debug token for" when asking which app to operate on.
--- a/src/appcheck/prompts.spec.ts
+++ b/src/appcheck/prompts.spec.ts
@@ -1,0 +1,100 @@
+import { expect } from "chai";
+import * as sinon from "sinon";
+
+import { getOrPromptAppId, getOrPromptProjectAndAppId } from "./prompts";
+import { AppCheckDebugOptions } from "./types";
+import * as apps from "../management/apps";
+import * as appUtils from "../appUtils";
+import * as projectUtils from "../projectUtils";
+import { FirebaseError } from "../error";
+
+describe("appcheck prompts", () => {
+  const projectId = "my-project";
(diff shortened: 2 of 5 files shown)


joehan (approved):
LGTM - will get this out shortly

joehan:
Thanks for the contirbution!

PR #10872, opened 2026-07-30
fix: set vscode version explicitly in tests

tonybaroneee:


firebase-vscode/src/test/runTest.ts
```diff
@@ -16,6 +16,7 @@ async function main() {
     // Download VS Code, unzip it and run the integration test
     const tmpUserData = path.join(os.tmpdir(), `vsc-ud-${Math.random().toString(36).substring(2, 7)}`);
     await runTests({
+      version: "1.96.4",
```
yuchenshi: Please extract a constant from https://github.com/firebase/firebase-tools/blob/e4f28d5083949ecfc37fef1b84380a14d1f2a589/firebase-vscode/src/test/default_wdio.conf.ts#L12

yuchenshi (changes requested):

firebase-vscode/src/test/runTest.ts
tonybaroneee: Done.

yuchenshi (approved):
If the tests pass then we're good. Thanks for the change!

tonybaroneee:
> If the tests pass then we're good. Thanks for the change!

No sweat! Yep, waiting on the result before merging. Thanks Yuchen.

joehan (approved):

Issue #10866, opened 2026-07-30
Functions emulator calls Secret Manager for secret params since 15.25.0 (isEmulator/force args swapped at resolveParams call site)


Bradley-McCallion:
### [REQUIRED] Environment info

**firebase-tools:** 15.25.0 (regression from 15.24.0; `master` still affected as of writing)

**Platform:** Ubuntu 24.04 (GitHub Actions `ubuntu-latest`), Node 24.18.1

### [REQUIRED] Test case

A functions codebase that declares a secret param, run against a demo project under the emulator:

```python
# src/transactions/_common/config.py
from firebase_functions import params
TRANSACTIONS_APPLE_PRIVATE_KEY = params.SecretParam("TRANSACTIONS_APPLE_PRIVATE_KEY")
```

```jsonc
// .firebaserc
{ "projects": { "testing": "demo-abcd" } }
```

`src/transactions/.secret.local` exists and contains a value for `TRANSACTIONS_APPLE_PRIVATE_KEY`.

We hit this with `runtime: python313` and `firebase-functions==0.5.0`, but the affected code is runtime-agnostic (`src/deploy/functions/build.ts`), so `defineSecret()` in a Node codebase should be affected identically.

### [REQUIRED] Steps to reproduce

```
firebase emulators:exec --project testing "<any command>"
```

### [REQUIRED] Expected behavior

The Functions emulator resolves secret params locally from `.secret.local` and never calls Cloud Secret Manager. This is the behaviour guaranteed by the `isEmulator` guard in `resolveParams`, added for #5520 / #7401:

```ts
// The functions emulator will handle secrets
if (!isEmulator) {
  for (const param of needSecret) { ... }
}
```

This worked in 15.24.0 and every prior version we've used.

### [REQUIRED] Actual behavior

The emulator makes an authenticated Secret Manager call against the demo project during trigger discovery and the codebase fails to load, so no triggers are registered:

```
i  emulators: Detected demo project ID "demo-abcd", emulated services will use a demo configuration
i  functions: Watching ".../src/transactions" for Cloud Functions...
   * Serving Flask app 'serving'
   127.0.0.1 - - "GET /__/functions.yaml HTTP/1.1" 200 -
   127.0.0.1 - - "GET /__/quitquitquit HTTP/1.1" 200 -
⬢  functions: Failed to load function definition from source: FirebaseError: Request to
[...]

PR #10875, opened 2026-07-30
fix argument transposition in resolveParams call

Berlioz:
Quick, embarassing, fix for https://github.com/firebase/firebase-tools/issues/10866.

Honestly the call to resolveParams has gotten out of hand to the point where using an options struct (like the ResolveBackendOpts struct that gets passed in and basically does nothing except be destructured anyway) would make the code cleaner and easier to avoid a recurrence of this typo. But let's get the bug fix pushed out before worrying about that.


joehan (approved):
LGTM after the GCA fix

ajperel (approved):

PR #10878, opened 2026-07-31
feat: add firepitOnly and dockerOnly release modes, support version overrides, and increase timeout

joehan:
### Description
Adds support for running selective release stages and specific versions when publishing artifacts:
- Introduces `firepitOnly` and `dockerOnly` modes to build and publish only those specific artifacts.
- Introduces a `--version-number` flag for `artifactsOnly`, `firepitOnly`, and `dockerOnly` runs to override the latest NPM version lookup.
- Increases the Cloud Build execution timeout to 60 minutes (`3600s`) to prevent jobs from timing out under load.

### Scenarios Tested
- Submitting dockerOnly one-off Cloud Build jobs.
- Local syntax and parameter parsing validation.

### DIFF
--- a/scripts/publish/cloudbuild.yaml
+++ b/scripts/publish/cloudbuild.yaml
@@ -95,6 +95,21 @@ steps:
     entrypoint: "bash"
     args: ["-c", "mkdir -vp ~/.config && cp -v hub ~/.config/hub"]
 
+  # Determine version number for artifactsOnly, firepitOnly, or dockerOnly modes.
+  - name: "gcr.io/$PROJECT_ID/package-builder"
+    entrypoint: "bash"
+    args:
+      - "-c"
+      - |
+        if [[ "${_VERSION}" == "artifactsOnly" || "${_VERSION}" == "firepitOnly" || "${_VERSION}" == "dockerOnly" ]]; then
+          if [[ -n "${_VERSION_NUMBER}" ]]; then
+            echo "${_VERSION_NUMBER}" > /workspace/version_number.txt
+          else
+            npm view firebase-tools version > /workspace/version_number.txt
+          fi
+          echo "Using version number: $(cat /workspace/version_number.txt)"
+        fi
+
   # Publish the package.
   - name: "gcr.io/$PROJECT_ID/package-builder"
     dir: "${_REPOSITORY_NAME}"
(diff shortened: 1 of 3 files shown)


scripts/publish/cloudbuild.yaml
```diff
@@ -95,17 +95,47 @@ steps:
...
+
+          # Update the package-lock.json for the docker image to use the target version
+          DOCKER_DIR="./firebase-docker-image"
+          if [[ ! -d "$DOCKER_DIR" ]]; then
+            DOCKER_DIR="./scripts/publish/firebase-docker-image"
+          fi
+          if [[ ! -d "$DOCKER_DIR" ]]; then
+            DOCKER_DIR="./firebase-tools/scripts/publish/firebase-docker-image"
+          fi
+          npm --prefix "$DOCKER_DIR" install firebase-tools@$VERSION_NUM
```
joehan: 🔴 **[Runtime Safety] Potential version drift in Docker build**\n\n*Rationale*: `npm install firebase-tools@$VERSION_NUM` (without `--save-exact`) will typically write a caret dependency (e.g., `^12.0.0`) to `package.json`. Since the `Dockerfile` runs `npm update`, it might upgrade `firebase-tools` to a newer version if one exists on npm, defeating the version override.\n\n*Suggested Fix*:\nUse `--save-exact` (or `-E`) to pin the version in `package.json` before building the docker image:\n```yaml\n          npm --prefix "$DOCKER_DIR" install --save-exact firebase-tools@$VERSION_NUM\n```

scripts/publish/run.sh
```diff
@@ -2,15 +2,44 @@
...
+VERSION_NUMBER=""
+
+if [[ $VERSION == "preview" ]]; then
+  BRANCH=$1
+  shift || true
+fi
+
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    --version-number)
```
joehan: 🔴 **[Script Robustness] Flag consumption bug**\n\n*Rationale*: If `--version-number` is passed without a value, but followed by another flag (e.g., `--version-number --another-flag`), the script will consume `--another-flag` as the version number because it only checks if `$2` is empty, not if it looks like a flag.\n\n*Suggested Fix*:\nValidate that the value does not start with `-`:\n```bash\n      if [[ -z "$2" || "$2" == -* ]]; then\n        echo "Error: --version-number requires a version argument."\n        printusage\n        exit 1\n      fi\n```

scripts/publish/run.sh
```diff
@@ -2,15 +2,44 @@
...
+shift || true
+
+BRANCH=""
+VERSION_NUMBER=""
+
+if [[ $VERSION == "preview" ]]; then
+  BRANCH=$1
+  shift || true
+fi
+
```
joehan: 🔴 **[CLI/UX] Missing validation for --version-number**\n\n*Rationale*: The `--version-number` option is only intended for `artifactsOnly`, `firepitOnly`, and `dockerOnly` modes (as documented in `printusage`), but there is no code enforcing this restriction. Running `run.sh patch --version-number 1.0.0` will silently ignore the override in Cloud Build.\n\n*Suggested Fix*:\nAdd validation after the argument parsing loop:\n```bash\nif [[ -n "$VERSION_NUMBER" ]]; then\n  if [[ ! ($VERSION == "artifactsOnly" || $VERSION == "firepitOnly" || $VERSION == "dockerOnly") ]]; then\n    echo "Error: --version-number is only allowed for 'artifactsOnly', 'firepitOnly', or 'dockerOnly' modes."\n    printusage\n    exit 1\n  fi\nfi\n```

PR #10880, opened 2026-07-31
fix: initialize auth in requirePermissions before projectId check

### LINKED ISSUE #10879: Firebase CLI does not use Google application default credentials consistently
Firebase is not respecting the Google App default creds that were set by gcloud and wants to authenticate me separately for some commands.

Not sure if bug or a feature, I'll let you decide, @joehan 

----

I logged in with gCloud's application default creds

Image

Ran `firebase firestore:locations` (a very boring cli command that should return me something), fails.

Image

Interestingly enough, `firebase projects:list` worked

Image

Tried to login, and it wanted to auth me independently

Image

-- 

## Info

```
➜  Desktop uname -a
Darwin richieforeman-mac.roam.internal 25.5.0 Darwin Kernel Version 25.5.0: Tue Jun  9 22:28:34 PDT 2026; root:xnu-12377.121.10~1/RELEASE_ARM64_T6041 arm64
```

```
➜  Desktop firebase --version
15.25.1
```

```
➜  Desktop gcloud --version
Google Cloud SDK 576.0.0
bq 2.1.34
bundled-python3-unix 3.14.5
core 2026.07.10
enterprise-certificate-proxy 0.3.18
gcloud-crc32c 1.0.0
gsutil 5.37
```

joehan:
### Description
Fixes an issue where running commands with Application Default Credentials (ADC) without a selected/configured project causes a confusing `Unable to refresh auth: not yet authenticated` error.

When a command that performs client-side IAM checks via `requirePermissions` is run without an active project, `requirePermissions` returned early before calling `requireAuth`. This left the local auth state (e.g. `lastOptions` and `accessToken`) uninitialized. When the command's action then proceeded to make API calls (which fallback to the active auth token), the auth client encountered an uninitialized auth state, resulting in a rejected promise with the `Unable to refresh auth` message.

This PR fixes it by calling `requireAuth(options)` at the beginning of `requirePermissions`, ensuring that the auth client and local tokens are always initialized before any early return.

Fixes #10879

### Scenarios Tested
1. Tested running `firestore:locations` under ADC without an active project (expected behavior: proper GCP API/project error instead of auth failure).
2. Tested running `firestore:locations` under ADC with a valid project (expected behavior: lists locations successfully).
3. Ran full Mocha test suite (`npm run mocha:fast`) to ensure no regressions in auth hooks.

### Sample Commands
- Running without a project (before fix vs after fix):
  `firebase firestore:locations --debug`
  - **Before Fix:**
    ```
    Error: Unable to refresh auth: not yet authenticated.
    ```
  - **After Fix:**
    ```
    Error: Request to https://firestore.googleapis.com/v1/projects/null/locations had HTTP Error: 403, Permission denied on resource project null.
    ```

- Running with a project under ADC:
  `firebase firestore:locations --project adam-test-c4a4d --debug`
  - **Before / After Fix:**
    Lists locations successfully.


aalej (approved):
LGTM! Although, just to note, we do call `requireAuth` before `requirePermissions` in most commands which would explain why we never really encountered this issue on other commands. For example `functions:secrets:get` - https://github.com/firebase/firebase-tools/blob/af93f9ec55b9636a721468083cc57c574c12930e/src/commands/functions-secrets-get.ts#L7-L12

we don't do this pattern on `firestore:locations`
https://github.com/firebase/firebase-tools/blob/af93f9ec55b9636a721468083cc57c574c12930e/src/commands/firestore-locations.ts#L10-L14

Since we moved `requireAuth` on the very top of `requirePermissions`, we might be able to remove the `before(requireAuth)` on other commands

note: an alternative fix might be to follow the existing pattern and add `requireAuth` before `requirePermissions`

PR #10881, opened 2026-07-31
fix: validate active project in firestore:locations command

joehan:
### Description
Updates the `firestore:locations` command action to retrieve the active project ID via `needProjectId(options)`.

Previously, running `firestore:locations` without an active project would pass `options.project` (which is undefined/null) directly to `api.locations(options.project)`. This resulted in calling the GCP Firestore API with `/projects/null/locations` or `/projects/undefined/locations`.

By using `needProjectId(options)` at the top of the command action, we ensure that a clean user-facing validation error ("No currently active project...") is thrown immediately if no project is active, matching standard CLI patterns.

### Scenarios Tested
1. Tested running `firestore:locations` without an active project (expected behavior: clean "No currently active project..." error message instead of failing on backend API call).
2. Tested running `firestore:locations` with a valid active project (expected behavior: successfully fetches and lists the firestore locations).

### Sample Commands
- Running without a project:
  `firebase firestore:locations`
  - **Before:** Fails on backend call with 403 Permission Denied on project null.
  - **After:**
    ```
    Error: No currently active project.
    To run this command, you need to specify a project. You have two options:
    - Run this command with --project <alias_or_project_id>.
    - Set an active project by running firebase use --add, then rerun this command.
    To list all the Firebase projects to which you have access, run firebase projects:list.
    To learn about active projects for the CLI, visit https://firebase.google.com/docs/cli#project_aliases
    ```


aalej (approved):
LGTM!

note: failing test seems to be flaky. re-running

PR #10883, opened 2026-08-03
fix(storage emulator): don't report request failures as "document not found" in firestore.get()

jon-comley:
### Description

`fetchFirestoreDocument()` in `src/emulator/storage/rules/runtime.ts` the path
backing `firestore.get()` / `firestore.exists()` inside Storage rules collapses
every possible failure into "the document does not exist":

```ts
try {
  const doc = await client.get(pathname);
  const { name, fields } = doc.body as { name: string; fields: string };
  const result = { name, fields };
  return { result, status: DataLoadStatus.OK, warnings: [], errors: [] };
} catch (e) {
  // Don't care what the error is, just return not_found
  return { status: DataLoadStatus.NOT_FOUND, warnings: [], errors: [] };
}
```

A confirmed HTTP 404 and a request that never reached a healthy server are
reported identically, with no retry and nothing logged. So a connection refusal,
a timeout, or a malformed response all cause a rule like:

```
allow write: if firestore.get(/databases/(default)/documents/jobs/$(jobId)).data.ownerId == request.auth.uid;
```

to evaluate as though the document is absent, and deny the write indistinguishable,
from the caller's side, from a genuine permissions failure.

Because the `try` also wraps the response destructuring, a malformed payload
(e.g. a null `body`) throws a `TypeError` that lands in the same catch and is
likewise reported as "not found", hiding a schema or programming error.

### Fix

- Only a **confirmed HTTP 404** is treated as a real "not found", returned
  immediately so the common "does this exist yet" path is unchanged and stays fast.
- Any other failure is retried a small, bounded number of times before giving up,
  on the grounds that it reflects the request not completing rather than an answer
  about the document.
- Response parsing moves **outside** the `try`, so a malformed payload surfaces as
  the error it is instead of being retried as a network problem.
- Errors are narrowed with a `hasHttpStatus` type guard rather than `any`.

### Scope and honesty about evidence

I hit a run of failures where Storage rules using `firestore.get()` behaved as
though the referenced document was missing under `emulators:exec`, which is what
led me here. I have **not** been able to reproduce that failure on demand since
I tried a dropped page cache, cold JARs, and heavy CPU contention, and the
unpatched code passed every time. I also never captured the swallowed error
itself, so I can't tell you which error it was.

I've therefore deliberately avoided asserting a specific root cause in this PR.
What I'm proposing stands on the code as written: conflating "the request failed"
with "the document doesn't exist" produces incorrect rule evaluations, and the
current code cannot distinguish them. If maintainers consider the existing
behaviour intentional, I'm happy to close this.

### Testing

Unit tests added in `src/emulator/storage/rules/runtime.spec.ts`:

- successful first attempt (no retries)
- confirmed 404 returns immediately **without** retrying
- transient connection failure retried, then succeeds
- malformed response body returns `NOT_FOUND` after exactly one call (no retries)
- persistent non-404 failure gives up and returns `NOT_FOUND`

`npm run test:compile` and `npm run lint:quiet` pass; the `runtime.spec.ts` suite
(11 tests) passes locally. I have not run the integration suites, which need
emulators and credentials I don't have configured.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,3 @@
 - Added support for a two-phase non-interactive login flow. Initiate this by running `firebase login --non-interactive`, navigate to the printed link to get an authorization code, and complete the login by running `firebase login <auth_code>`.
 - Fixed `apps:init` writing `google-services.json` to an `app/app` path when the Android module has no `src` directory, by detecting the module from the directory basename instead of the first path segment (#10863).
+- Fixed `firestore.get()`/`firestore.exists()` in Storage rules reporting a failed request as "document not found", which could cause rules to deny access on the strength of a transport failure. Only a confirmed 404 is now treated as a real answer; other failures are retried.
--- a/src/emulator/storage/rules/runtime.spec.ts
+++ b/src/emulator/storage/rules/runtime.spec.ts
@@ -1,6 +1,8 @@
 import { expect } from "chai";
-import { createAuthExpressionValue, StorageRulesRuntime } from "./runtime";
-import { RulesetOperationMethod, RuntimeActionResponse } from "./types";
+import * as sinon from "sinon";
+import { createAuthExpressionValue, fetchFirestoreDocument, StorageRulesRuntime } from "./runtime";
+import { DataLoadStatus, RulesetOperationMethod, RuntimeActionResponse } from "./types";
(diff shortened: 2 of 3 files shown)


jon-comley:
@googlebot I signed it!

jon-comley:
@googlebot I signed it!

PR #10884, opened 2026-08-03
fix: tolerate 404 when deleting scheduled function's schedule

### LINKED ISSUE #4795: Function deploy fails to delete scheduled function without Cloud Scheduler job

IzaakGough:
Fixes #4795

When deleting a scheduled function, `firebase deploy` failed the whole deployment if the Cloud Scheduler job or its Pub/Sub topic was already gone (404). Deleting a resource that doesn't exist is the desired end state, so these deletes now tolerate a 404 and continue; non-404 errors still fail as before.

This unblocks users who hit a partially-deleted schedule and could not deploy again without manual cleanup.

### Testing

- Manually checked that fix works on real deploy
- Automated tests pass


cabljac (approved):
lgtm

wandamora (approved):
LGTM for google review

PR #10887, opened 2026-08-04
feat(appcheck): add appcheck:services commands for enforcement

marb2000:
### Description

Adds App Check enforcement to the CLI, next to the debug token commands from #10801. Today the CLI can create a debug token but cannot turn on the enforcement that token is for, so App Check still needs the console.

Three commands, behind the new `appcheckadmin` experiment. `appcheck:debugtokens:*` is unchanged and stays generally available.

```
firebase appcheck:services:list
firebase appcheck:services:get <service>
firebase appcheck:services:set <service> <mode> [--replay-protection <mode>] [--force]
```

```
$ firebase appcheck:services:list
┌─────────────┬────────────────────────────┬─────────────┬───────────────────┬─────────────────────────────┐
│ Service     │ Service API                │ Enforcement │ Replay Protection │ Updated                     │
├─────────────┼────────────────────────────┼─────────────┼───────────────────┼─────────────────────────────┤
│ ailogic     │ Firebase AI Logic          │ Unenforced  │ Off               │ 2026-08-04T04:16:03.452719Z │
│ auth        │ Firebase Authentication    │ Unenforced  │ Off               │ 2025-12-15T21:01:43.672511Z │
│ database    │ Firebase Realtime Database │ Off         │ Off               │ never                       │
│ firestore   │ Cloud Firestore            │ Unenforced  │ Off               │ 2026-08-04T01:06:31.086831Z │
│ storage     │ Cloud Storage for Firebase │ Unenforced  │ Off               │ 2026-03-11T21:41:26.948831Z │
└─────────────┴────────────────────────────┴─────────────┴───────────────────┴─────────────────────────────┘
```

**Firebase AI Logic.** Starting November 2, 2026, Firebase automatically enforces App Check for all Gemini API requests via AI Logic, and it cannot be un-enforced. The CLI warns when a project is not ready, before someone relaxes it, and after the write, using the wording from the customer notice. AI Logic also skips the usual unenforced, watch metrics, enforced rollout: it stays enforced, with a debug token during development and a real attestation provider before shipping.

Attestation providers and an apps overview come in a follow up PR, to keep this one small.

### Design notes

- **Behind an experiment**, since the surface is not API council approved. Debug tokens already shipped, so they stay outside the flag.
- **Short names in, product names out.** `ailogic`, `auth`, `firestore`, `database`, `storage`, `dataconnect`. Service ids stay internal: they are an API detail, and App Check knows AI Logic as `firebaseml.googleapis.com`, which is a different product's name. A full id still works for the Google services that have no short name.
- **We ask only before the change that breaks clients.** Enforcing can lock out users on old app versions. Relaxing does not, except for AI Logic, where relaxing is the risky move.
- **A service with no enforcement reads as `Off`.** The API omits `enforcementMode` when it is OFF, so an explicitly disabled service and one nobody configured are the same response. `never` in the Updated column marks an untouched one.
- **No `deploy --only appcheck`.** Enforcement is a live operational setting, not repo state.

Three API details worth knowing, all confirmed against a live project: `firebasevertexai.googleapis.com` and `cloudfunctions.googleapis.com` are not App Check services; replay protection works on AI Logic but returns a bare 400 on Firestore, Storage, Auth and SQL Connect, so the CLI explains it; and the IAM permissions are `firebaseappcheck.services.get` and `.services.update`.

### Scenarios Tested

- 48 new unit tests, full suite passes (4891).
- Live runs on `ai-logic-demos`: list, get and set, a real enforce and revert on a service with no traffic, the unknown service error, the non interactive abort, and the replay protection rejection, which writes nothing. A read only check on `book-watch-ac89e`, where AI Logic is enforced, for the un-enforce warning. Both projects were left as they were found.
- Experiment off: the commands do not exist and `firebase help appcheck` shows only debugtokens. Experiment on: services appears.

### Sample Commands

```
firebase experiments:enable appcheckadmin

firebase appcheck:services:list
firebase appcheck:services:get ailogic
firebase appcheck:services:set firestore unenforced
firebase appcheck:services:set firestore enforced
```

### DIFF
--- a/src/appcheck/services.ts
+++ b/src/appcheck/services.ts
@@ -61,7 +61,7 @@ export function displayNameForServiceId(serviceId: string): string {
  * no alias. Listed here only so a full id passes validation instead of being
  * rejected by the CLI before the API can answer.
  */
-const OTHER_SUPPORTED_SERVICE_IDS = new Set([
+const OTHER_SUPPORTED_SERVICE_IDS: ReadonlySet<string> = new Set([
   "maps-backend.googleapis.com",
   "places.googleapis.com",
   "oauth2.googleapis.com",
@@ -74,7 +74,7 @@ const OTHER_SUPPORTED_SERVICE_IDS = new Set([
  * AI Logic is a paid resource that abusers target, so it should never sit
  * unprotected.
  */
-const AUTO_ENFORCED_SERVICE_IDS = new Set(["firebaseml.googleapis.com"]);
+const AUTO_ENFORCED_SERVICE_IDS: ReadonlySet<string> = new Set(["firebaseml.googleapis.com"]);
 
 /**
  * The day App Check enforcement becomes mandatory for Firebase AI Logic.
--- a/src/appcheck/services.ts
+++ b/src/appcheck/services.ts
@@ -1,0 +1,258 @@
+import { FirebaseError } from "../error";
+import { EnforcementMode, Service } from "./types";
+
+/**
+ * Short names for the services App Check can protect, mapped to the service ids
+ * the API uses.
+ *
+ * Nobody can guess these ids, so the aliases are what we document and what the
+ * error messages list. A full service id is still accepted, which is also how
+ * the services the CLI has no alias for (Google Maps, Google Identity for iOS)
+ * stay reachable.
+ *
+ * `ailogic` maps to `firebaseml.googleapis.com`, which looks wrong because
+ * Firebase ML is a different product and AI Logic talks to
+ * `firebasevertexai.googleapis.com`. It is right: an App Check service id names
+ * a service to App Check and is not the API host, as the API reference says
+ * itself. `firebasevertexai.googleapis.com` is rejected with "Service not
+ * supported", and `firebaseml.googleapis.com` is what the API reference labels
+ * Firebase AI Logic and what Firebase auto enforces on AI Logic projects.
+ */
+export const SERVICE_ALIAS_TO_ID: Readonly<Record<string, string>> = {
+  auth: "identitytoolkit.googleapis.com",
+  firestore: "firestore.googleapis.com",
+  database: "firebasedatabase.googleapis.com",
+  storage: "firebasestorage.googleapis.com",
+  ailogic: "firebaseml.googleapis.com",
+  dataconnect: "firebasedataconnect.googleapis.com",
+};
+
+const SERVICE_ID_TO_ALIAS: Record<string, string> = Object.fromEntries(
+  Object.entries(SERVICE_ALIAS_TO_ID).map(([alias, id]) => [id, alias]),
+);
+
+/**
+ * Product names as the App Check documentation writes them.
+ *
+ * These are what we show. The service ids are an implementation detail of the
+ * API, and a couple of them are actively confusing: AI Logic is identified as
+ * `firebaseml.googleapis.com`, which is a different product's name. Developers
+ * should read "Firebase AI Logic" and type `ailogic`, and never have to care.
+ */
+const SERVICE_DISPLAY_NAMES: Readonly<Record<string, string>> = {
+  "identitytoolkit.googleapis.com": "Firebase Authentication",
+  "firestore.googleapis.com": "Cloud Firestore",
+  "firebasedatabase.googleapis.com": "Firebase Realtime Database",
+  "firebasestorage.googleapis.com": "Cloud Storage for Firebase",
+  "firebaseml.googleapis.com": "Firebase AI Logic",
+  "firebasedataconnect.googleapis.com": "Firebase SQL Connect",
+  "maps-backend.googleapis.com": "Maps JavaScript API",
+  "places.googleapis.com": "Places API (New)",
+  "oauth2.googleapis.com": "Google Identity for iOS",
+};
+
+/** The product name for a service id, or the id when we have no name for it. */
+export function displayNameForServiceId(serviceId: string): string {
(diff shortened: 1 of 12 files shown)


src/commands/appcheck-apps-list.ts
```diff
@@ -0,0 +1,99 @@
...
+        configsByProvider: new Map(ATTESTATION_PROVIDERS.map((p, i) => [p, configLists[i]])),
+      };
+    }, "Reading apps and App Check providers");
+
+    const rows: AppRow[] = apps.map((app) => {
+      const providers = providersForPlatform(app.platform).filter((provider) => {
+        const config = configsByProvider
+          .get(provider)
+          ?.find((c) => appIdFromConfigName(c.name) === app.appId);
+        return config ? isConfigured(provider, config) === true : false;
```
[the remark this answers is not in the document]
marb2000: Good catch, fixed. Here false and null both mean the same thing for this list, so there is nothing to tell apart. Used Boolean() to match the style in the rest of the module, and added a line explaining why a provider we cannot confirm is left out.

marb2000:
Hi @joehan, whenever you have a moment, could you take a look at this one?

It finishes the App Check surface that Austin started with the debug tokens: enforcement per service, the attestation providers for each app, and a small apps overview. It reuses his module and conventions rather than adding anything next to them.

Everything new is behind an `appcheckadmin` experiment, so it stays hidden until the surface goes through API council. The debug token commands are untouched.

No rush, but there is one reason to land it sooner rather than later: from November 2 App Check is enforced for AI Logic and cannot be turned off, and right now the CLI cannot even tell a developer where their project stands. The PR adds warnings for that, using the same wording as the notice we sent customers.

Thanks!

marb2000:
Good call, split it. This PR is now just `appcheck:services` (enforcement), down from 20 files to 12 and from 2141 lines to 970.

The attestation providers and `apps:list` are ready on a branch and will come as a second PR once this one lands, so you do not have to review them on top of each other.

I kept this half first because it is the one the November 2 AI Logic deadline needs.

src/appcheck/types.ts
```diff
@@ -16,3 +16,29 @@ export interface AppCheckDebugOptions extends Options {
...
+ */
+export interface Service {
+  name: string;
+  enforcementMode?: EnforcementMode;
+  replayProtection?: EnforcementMode;
+  updateTime?: string;
+  etag?: string;
+}
+
+export interface ListServicesResponse {
```
christhompsongoogle: Curious if services is expected to be empty or actually missing? This object has all optional fields. Prefer services to be a non-null list instead of a nullable AND possibly empty list

src/commands/appcheck-services-set.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+  };
+}
+
+describe("appcheck:services:set", () => {
+  let getServiceStub: sinon.SinonStub;
+  let updateServiceStub: sinon.SinonStub;
+  let confirmStub: sinon.SinonStub;
+
+  beforeEach(() => {
+    (command as unknown as { befores: unknown[] }).befores = []; // bypass auth hook
```
christhompsongoogle: better to find a known type here than `as unknown as T`

src/appcheck/services.ts
```diff
@@ -0,0 +1,258 @@
...
+export function displayNameForServiceId(serviceId: string): string {
+  return SERVICE_DISPLAY_NAMES[serviceId] ?? serviceId;
+}
+
+/**
+ * Service ids the API accepts but that are not Firebase products, so they get
+ * no alias. Listed here only so a full id passes validation instead of being
+ * rejected by the CLI before the API can answer.
+ */
+const OTHER_SUPPORTED_SERVICE_IDS = new Set([
```
christhompsongoogle: nit: ReadonlySet preferred for immutability here

src/appcheck/services.ts
```diff
@@ -0,0 +1,258 @@
...
+]);
+
+/**
+ * Firebase turns App Check on for these services by default, so relaxing them
+ * is the risky direction and the CLI confirms before doing it.
+ *
+ * AI Logic is a paid resource that abusers target, so it should never sit
+ * unprotected.
+ */
+const AUTO_ENFORCED_SERVICE_IDS = new Set(["firebaseml.googleapis.com"]);
```
christhompsongoogle: nit: ReadonlySet preferred for immutability here

christhompsongoogle (approved):

src/appcheck/types.ts
marb2000: Missing, not empty: the API leaves `services` out when a project has no service configured, the same way it omits `enforcementMode` when the value is OFF. So the optional field is only the shape of the raw response.

Agreed the exposed contract should not be nullable, and `listServices` already returns a plain `Service[]`. Moved the interface into api.ts as a private wire type so it is not part of the shared types at all, with a comment saying why the field is optional.

src/commands/appcheck-services-set.spec.ts
marb2000: Fixed. `befores` is private on Command, which is why the specs in this repo reach it through a cast, but `command["befores"] = []` typechecks and drops the cast entirely. Nice, thanks.

src/appcheck/services.ts
marb2000: Done.

src/appcheck/services.ts
marb2000: Done.

PR #10889, opened 2026-08-04
fix: v2 to v1 downgrade failing with CPU error

### LINKED ISSUE #5461: [Bug - GCF] Error deploying a v1 auth blocking function

**firebase-tools: 11.21.0**

**Platform: MacOS**

### [REQUIRED] Test case

Deps:
```
    "firebase-admin": "^11.3.0",
    "firebase-functions": "^4.2.0",
```
Try deploying the following function:

```ts
import { user } from "firebase-functions/v1/auth";
import { db } from "./setup/db";
import { firestore } from "firebase-admin";
import { logger } from "firebase-functions/v1";
import {
  AuthEventContext,
  AuthUserRecord,
} from "firebase-functions/lib/common/providers/identity";

export const syncUserToDb = user().beforeCreate(
  async (user: AuthUserRecord, context: AuthEventContext) => {
    const uid = user.uid;

    // Sync to user table.
    await db.doc(`users/${uid}`).create({
      uid,
      displayName: defaultName,
      email: user.email,
      creationTime: firestore.FieldValue.serverTimestamp(),
      lastSignInTime: firestore.FieldValue.serverTimestamp(),
    });

    logger.info(`Added user ${uid} to table`);
  }
);
```

Try deploying the above function.

Should be able to deploy

Got the following error:
```
Error: Cannot set CPU on the functions syncusertodb because they are GCF gen 1
```

Think it comes from here:
https://github.com/firebase/firebase-tools/blob/ad99ec653a6e8486f28ef97bcc9471a83dc0268b/src/deploy/functions/validate.ts#L14

IzaakGough:
Fixes #5461

Redeploying an existing v2 function as v1 failed with a confusing `Cannot set CPU on ... because they are GCF gen 1` error. `inferDetailsFromExisting` copied `cpu` from the deployed v2 function onto the v1 endpoint, masking the real reason: v2 cannot be downgraded to v1.

The downgrade is now rejected in prepare's phase 2.5, before the source is prepared and before anything is inherited from the existing function, so the CPU and timeout errors are unreachable on this path. The check is shared with the release planner via `generationDowngradeMessage`, and the message tells the user to `firebase functions:delete` first.

This also closes a gap: a Cloud Run service redeployed as gcfv1 previously fell through `checkForIllegalUpdate` with no error at all.

Verified with unit tests on `noGenerationDowngrades`, `generationDowngradeMessage`, and `checkForIllegalUpdate`, each mutation-checked. `prepare()` has no integration harness, so the phase 2.5 call site itself is not covered by a test.


src/deploy/functions/validate.ts
```diff
@@ -91,6 +96,9 @@ export function endpointsAreValid(
   validateLifecycleHooks(wantBackend, existingBackend);
   const endpoints = backend.allEndpoints(wantBackend);
   functionIdsAreValid(endpoints);
+  if (existingBackend) {
+    noGenerationDowngrades(wantBackend, existingBackend);
```
cabljac: Worth pinning this position with a test. The order matters: timeout is still copied from the existing function regardless of platform (prepare.ts:651-653), so a v2 function with a long timeout redeployed as v1 reaches this function carrying that timeout. Today that's harmless because your check throws first, but move this call below `validateTimeoutConfig` on :102 and we're back to showing "3600s (limit: 540s)" instead of the real reason.

Nothing catches that at the moment. The two downgrade tests would still pass with this call moved to the end, because their `want` is clean so no other check fires either way.

The test needs the timeout on `want` rather than `have`, since `endpointsAreValid` doesn't run `inferDetailsFromExisting` and only walks the want backend:

```ts
it("reports the downgrade rather than the inherited timeout", () => {
  // Mirrors the state after inferDetailsFromExisting, where want has picked up the v2 timeout.
  const want = backend.of({ ...ENDPOINT_BASE, platform: "gcfv1", timeoutSeconds: 3600 });
  const have = backend.of({ ...ENDPOINT_BASE, platform: "gcfv2", timeoutSeconds: 3600 });

  expect(() => validate.endpointsAreValid(want, have)).to.throw(
    /cannot be downgraded from GCFv2 to GCFv1/,
  );
});
```

I checked that one both ways: it passes with the call where you have it, and fails on the 540s message with it moved to the end.

Good habit generally: if a fix works by putting one check in front of another, write a test that fails when they swap. Otherwise the next person to tidy that function undoes it and nothing goes red.

src/deploy/functions/release/planner.ts
```diff
@@ -382,10 +383,9 @@ export function checkForIllegalUpdate(want: backend.Endpoint, have: backend.Endp
       )}] Changing from ${haveType} function to ${wantType} function is not allowed. Please delete your function and create a new one instead.`,
     );
   }
-  if (want.platform === "gcfv1" && have.platform === "gcfv2") {
-    throw new FirebaseError(
-      `[${getFunctionLabel(want)}] Functions cannot be downgraded from GCFv2 to GCFv1`,
-    );
+  const downgrade = generationDowngradeMessage(want, have);
```
cabljac, line 393: Nit: this changes the planner as well. Before, a Cloud Run service redeployed as gcfv1 got no error at all here. The trigger type check above passes because `triggerType` looks at the trigger shape rather than the platform, so a run service with an https trigger and a gcfv1 https function both come out as "an HTTPS", and `checkForV2Upgrade` on :402 only covers the gcfv1 to gcfv2 direction. It fell straight through.

I think throwing is right, letting it through silently would be worse, but it's a change beyond the CPU fix in the title. Worth covering in the description rewrite above.

cabljac (commented):
@IzaakGough nice find on the root cause. The cpu inheritance masking the real error is exactly it, and moving the check ahead of the source upload is the right call rather than letting people wait on an upload that was always going to fail. Reading it against main, the shape looks right to me. One test to add and one nit inline.

Separately, the description doesn't quite match the code. It says all three of `cpu`, `availableMemoryMb` and `timeoutSeconds` are "now only inherited when the platform is unchanged", but the guard you added is on `cpu` alone, and it's `wantE.platform !== "gcfv1"` rather than a platform-equality check. Memory and timeout still cross platforms as before.

I suspect that's deliberate, since both are valid on either generation and the paths that survive the new check all have compatible ranges, in which case only the wording needs fixing. But if you did intend to guard all three, worth saying why you settled on cpu only.

src/deploy/functions/functionsDeployHelper.ts
```diff
@@ -171,6 +171,23 @@ export function getFunctionLabel(fn: backend.TargetIds & { codebase?: string }):
...
+ */
+export function generationDowngradeMessage(
+  want: backend.Endpoint,
+  have: backend.Endpoint,
+): string | undefined {
+  if (want.platform !== "gcfv1" || have.platform === "gcfv1") {
+    return undefined;
+  }
+  const from = have.platform === "gcfv2" ? "GCFv2" : "Cloud Run";
+  return `[${getFunctionLabel(want)}] Functions cannot be downgraded from ${from} to GCFv1`;
```
cabljac: Worth telling the user what to do next: "Please delete the existing function (`firebase functions:delete <id>`) and redeploy to recreate it as GCFv1." Users hitting #5461 usually want the downgrade.

src/deploy/functions/validate.ts
```diff
@@ -132,6 +140,23 @@ export function endpointsAreValid(
...
+ */
+function noGenerationDowngrades(
+  wantBackend: backend.Backend,
+  existingBackend: backend.Backend,
+): void {
+  for (const want of backend.allEndpoints(wantBackend)) {
+    const have = existingBackend.endpoints[want.region]?.[want.id];
+    const msg = have && generationDowngradeMessage(want, have);
+    if (msg) {
+      throw new FirebaseError(msg);
```
cabljac: The sibling validators here collect all offenders and throw once, but this throws on the first - is that intended?

cabljac (approved):
Approving, nice work @IzaakGough. Verified on a fresh checkout: mutation-testing the check order fails exactly your ordering test, so it genuinely pins the fix. Suite green (983).

Two asks:
1. CHANGELOG entry, e.g. `- [Fixed] Report the GCFv2-to-GCFv1 downgrade error during validation instead of a misleading CPU error (#5461).` (unless changelogs are generated in this repo, i forget)
2. One planner.spec test for run -> gcfv1 matching `/cannot be downgraded from Cloud Run to GCFv1/` - it's new wording nothing exercises.

src/deploy/functions/validate.ts
IzaakGough: Not intended, good catch. Changed it to collect and throw once, like `validateTimeoutConfig` does below.

It matters here because this check runs before the source upload. Someone moving a whole codebase back to v1 would otherwise fix one function, redeploy, wait for discovery, and hit the next one.

The per-endpoint message already carries its own label and `functions:delete` command, so the messages just join with a newline and no wording changes. Added a test with two downgraded endpoints (one from GCFv2, one from Cloud Run) that asserts both appear; it fails on the old first-throw version.

Left the planner's `checkForIllegalUpdate` as is, since it's called per endpoint and has nothing to aggregate. `endpointsAreValid` is also still first-check-wins overall, which is what keeps the downgrade error ahead of the 540s timeout error.

src/deploy/functions/release/planner.spec.ts
```diff
@@ -739,6 +739,15 @@ describe("planner", () => {
 
       expect(() => planner.checkForIllegalUpdate(want, have)).to.throw();
     });
+
+    it("should throw if a user redeploys a Cloud Run service as v1", () => {
+      const want: backend.Endpoint = { ...func("id", "region"), platform: "gcfv1" };
+      const have: backend.Endpoint = { ...func("id", "region"), platform: "run" };
```
ajperel: while I expect this works I'd use gcfv2 not "run" since we don't have run platform node functions yet. Though when we do the behavior should be the same.

CHANGELOG.md
```diff
@@ -1,2 +1,5 @@
+- [Added] Add -f, --force option to `firebase ext:migrate`.
```
ajperel: I think you have some extra changes in here.

src/deploy/functions/prepare.ts
```diff
@@ -649,7 +649,10 @@ export function inferDetailsFromExisting(
       wantE.availableMemoryMb = haveE.availableMemoryMb;
     }
 
-    if (typeof wantE.cpu === "undefined" && haveE.cpu) {
+    // cpu does not exist on gcfv1. Inheriting it from an existing gcfv2 function onto
+    // a gcfv1 endpoint fails CPU validation and masks the accurate "cannot be
+    // downgraded" error.
+    if (typeof wantE.cpu === "undefined" && haveE.cpu && wantE.platform !== "gcfv1") {
```
ajperel: This feels a little hacky, to fall through for later errors. But other things are weird falling through too. I'd rather we universally try to catch errors with downgrading earlier in the process and not require hacks here.

What if instead we make the check for downgrading happen at phase 2.5 ~line 420 above with noGenerationDowngrades and fail fast?

And then we also don't need that whole test of this passthrough either.

IzaakGough: Agreed this is a bit hacky. I like your suggestion here of moving the downgrading check up. Added in b2122eb4b565fd19cf4eddee8bfe887ef5686ea1

ajperel (approved):
Looks great. Thanks for the changes.

PR #10891, opened 2026-08-05
feat(appcheck): add providers and apps:list commands

marb2000:
### Description

Second half of the App Check surface, after #10887 added enforcement. Providers configure how one app proves it is real, and `apps:list` shows which apps are ready before you turn enforcement on.

Four commands, behind the same `appcheckadmin` experiment.

```
firebase appcheck:providers:list [--app <appId>]
firebase appcheck:providers:get <provider> [--app <appId>]
firebase appcheck:providers:set <provider> [--app <appId>] [provider flags]
firebase appcheck:apps:list
```

Only the providers that fit the app platform are shown:

```
$ firebase appcheck:providers:list --app 1:861083271982:ios:d0aec6d9bd634dddeeb0bf
App: 1:861083271982:ios:d0aec6d9bd634dddeeb0bf (IOS)
┌──────────────────────┬────────────┬───────────┬────────────────┐
│ Provider             │ Configured │ Token TTL │ Settings       │
├──────────────────────┼────────────┼───────────┼────────────────┤
│ app-attest           │ n/a        │ 1h        │                │
│ device-check         │ No         │ 1h        │                │
│ recaptcha-enterprise │ No         │ 1h        │ min score: 0.5 │
└──────────────────────┴────────────┴───────────┴────────────────┘
```

Flags per provider, with secrets accepting `@file` so a private key does not end up in the shell history:

```
firebase appcheck:providers:set device-check --app <iOS app> --key-id ABCD1234EF --private-key @AuthKey.p8
firebase appcheck:providers:set recaptcha-enterprise --app <web app> --site-key 6Lc_... --min-score 0.7
firebase appcheck:providers:set play-integrity --app <Android app> --min-device-integrity strong
```

### Design notes

- **`n/a` where the API cannot answer.** A provider config always returns 200 with defaults, even when nobody configured it. For device-check, recaptcha-v3 and recaptcha-enterprise there is a key or secret to check, so "Configured" is real. App Attest and Play Integrity have neither, so they report `n/a` instead of a guess.
- **Only the flags the user passed go into the update mask**, so setting a site key never resets the token lifetime.
- **One apps listing per command.** `getOrPromptApp` resolves `--app` and the platform together, and validates the app against the project, so a wrong app id fails with a clear message instead of a later API error.
- **`apps:list` uses batchGet**, one call per provider for every app, so a project with ten apps costs six calls instead of fifty. It has no debug token column: counting tokens needs one more call per app, and `appcheck:debugtokens:list --app` already answers that for one app.
- reCAPTCHA works on iOS and Android as well as web now, through the provider the mobile SDKs added in June 2026. It uses the same reCAPTCHA Enterprise settings, so `recaptcha-enterprise` covers it, and the help text says so.

### Scenarios Tested

- 25 new unit tests, full suite passes (4916). They cover the platform rules, the update mask builder for every provider, token TTL and score boundaries, `@file` secrets, and the providers whose state cannot be known.
- Live runs on `ai-logic-demos`: `providers:list` on iOS, Android and Web apps, `providers:get`, a real `providers:set` write and revert on the token lifetime with the other settings surviving, every input error, and `apps:list`.
- Experiment off: the commands do not exist and `firebase help appcheck` shows only debugtokens. Experiment on: providers and apps appear.

### Sample Commands

```
firebase experiments:enable appcheckadmin

firebase appcheck:providers:list --app 1:1234567890:ios:abc123
firebase appcheck:providers:get recaptcha-enterprise --app 1:1234567890:web:abc123
firebase appcheck:providers:set recaptcha-enterprise --app 1:1234567890:web:abc123 --site-key 6Lc_abc --min-score 0.7
firebase appcheck:apps:list
```


src/appcheck/providers.ts
```diff
@@ -0,0 +1,333 @@
...
+function resolveSecretFlag(value: string): string {
+  if (!value.startsWith("@")) {
+    return value;
+  }
+  const path = value.slice(1);
+  if (!fs.existsSync(path)) {
+    throw new FirebaseError(`File not found: ${path}`);
+  }
+  return fs.readFileSync(path, "utf8").trim();
+}
```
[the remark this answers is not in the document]
marb2000: Done, and I went a bit further. I dropped the `existsSync` check as well. It was a second look at the file that still let a directory through, and the `try/catch` covers both cases on its own.

```
Error: Could not read /no/such/key.p8: ENOENT: no such file or directory, open '/no/such/key.p8'
Error: Could not read /tmp: EISDIR: illegal operation on a directory, read
```

src/commands/appcheck-providers-set.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (result.siteSecretSet) {
+      logger.info(`   Site secret: set`);
+    }
+    const minScore = result.riskAnalysis?.minValidScore ?? result.minValidScore;
+    if (minScore !== undefined) {
+      logger.info(`   Min score:   ${minScore}`);
+    }
+    if (result.deviceIntegrity?.minDeviceRecognitionLevel) {
+      logger.info(`   Min device:  ${result.deviceIntegrity.minDeviceRecognitionLevel}`);
+    }
```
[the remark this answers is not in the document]
marb2000: Good catch, fixed, but not quite the way suggested. The `!== undefined` guard would not have worked: the API leaves these fields out instead of sending `false`, so turning a setting off would still have printed nothing. Real responses:

```
--require-licensed      -> "accountDetails": { "requireLicensed": true }
--no-require-licensed   -> "accountDetails": {}
```

A missing field means off, so I print both lines on every play-integrity write instead of guarding on them.

Looking into this also turned up a worse bug. `--require-licensed` and `--allow-unrecognized-version` had no negative form, so once a user turned one on there was no way to turn it off. Both have a `--no-` form now, which gives the three states the update mask needs: absent means leave it alone, true, and false.

src/appcheck/providers.ts
```diff
@@ -0,0 +1,336 @@
...
+interface ProviderMeta {
+  /** The per app config sub resource, for example "appAttestConfig". */
+  configResource: string;
+  /** The app platforms this provider can attest. */
+  platforms: AppCheckPlatform[];
+  /** Human readable name for tables and messages. */
+  label: string;
+}
+
+/**
```
joehan: Nit: This comment is probably excessive detail and can be removed. At most, maybe just put an inline comment next to recaptcha-enterprise stating when mobile platform suppor was added

src/appcheck/providers.ts
```diff
@@ -0,0 +1,336 @@
...
+}
+
+const TTL_MIN_SECONDS = 30 * 60;
+const TTL_MAX_SECONDS = 7 * 24 * 60 * 60;
+
+/**
+ * Turns `30m`, `2h`, `1d` or `3600s` into the seconds string the API wants.
+ * The API accepts 30 minutes to 7 days.
+ */
+function parseTokenTtl(ttl: string): string {
```
joehan, line 116: We have some very similar utils in src/firestore/backupUtils.ts that parse duration  strings into number of seconds. Can we dedupe this code into one single utility that is shared? 

Feel free to do this in a follow up PR.

src/appcheck/providerUpdate.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+import { buildProviderUpdate } from "./providers";
+import { FirebaseError } from "../error";
+import { ProviderFlags } from "./types";
+
+function options(extra: ProviderFlags = {}): ProviderFlags {
+  return extra;
+}
+
+describe("appcheck buildProviderUpdate", () => {
```
joehan: No real reason to put this in a separate file - combine with provider.spec.ts

src/appcheck/providers.ts
```diff
@@ -0,0 +1,336 @@
+import * as fs from "fs";
+
+import { FirebaseError, getErrMsg, getError } from "../error";
+import { ProviderConfig, ProviderFlags, ProviderType } from "./types";
+
+/** The app platforms App Check can attest. */
+type AppCheckPlatform = "IOS" | "ANDROID" | "WEB";
```
joehan: Seems like these should live in src/appcheck/types, right?

joehan (approved):
Some minor stuff I'd like to address, and one request that can be done now or in a follow up. Otherwise, this lgtm!

src/commands/appcheck-prompts.ts
christhompsongoogle, line 1: nit: this is really more coupled with the command interface and not the service or logic, suggest renaming to appcheck-prompts.ts and moving under commands

src/appcheck/providers.ts
```diff
@@ -0,0 +1,336 @@
...
+    case "recaptcha-v3":
+      return Boolean(config.siteSecretSet);
+    case "recaptcha-enterprise":
+      return Boolean(config.siteKey);
+    default:
+      return null;
+  }
+}
+
+const TTL_MIN_SECONDS = 30 * 60;
```
christhompsongoogle: nit: constants at top of file

christhompsongoogle (approved):

src/appcheck/providers.ts
marb2000: Done. Cut down to two lines next to `recaptcha-enterprise`.

src/appcheck/providerUpdate.spec.ts
marb2000: Done, folded into providers.spec.ts.

src/appcheck/providers.ts
marb2000: Yes, moved both to src/appcheck/types.ts.

src/commands/appcheck-prompts.ts
marb2000: Done. It is `src/commands/appcheck-prompts.ts` now. You are right that it builds the command interface rather than the App Check logic.

src/appcheck/providers.ts
marb2000: Done.

src/appcheck/providers.ts
marb2000: Taking you up on the follow up PR, if that is alright.

The two parsers accept different units. `calculateRetention` takes `hdmw`, mine takes `smhd`. So a shared parser either widens what `firestore backups --retention` accepts (it would start taking `30s`) or narrows what a token TTL accepts. Both are behaviour changes in shipped commands, and the firestore one deserves its own review rather than riding along here.

I will send a small PR that puts one duration parser in a shared spot and moves both callers onto it, with each keeping its own range check and error message.

marb2000: Hi Joe, thanks for the pointer. I looked into it and I have a concern I want to raise before I do it.

There are actually three copies of this, not two. Hosting has one as well:

- `src/hosting/expireUtils.ts` accepts `m h d` and returns milliseconds, max 30d
- `src/firestore/backupUtils.ts` accepts `m h d w` and returns seconds
- my `parseTokenTtl` accepts `s m h d` and returns a `"3600s"` string, 30m to 7d

The shared part is small. Match a number and a letter, then multiply. The problem is everything around it is different on purpose. Backups allow weeks. Token TTLs allow seconds. Hosting wants milliseconds because its API does.

So if the shared parser has one fixed list of units, all three commands change. `firestore backups --retention 30s` would start being accepted, and `hosting:channel:create --expires 2w` would too. Neither works today.

That is my worry. Those are behavior changes in Hosting and Firestore, and I do not own those products. I am not comfortable changing what their flags accept as a side effect of an App Check cleanup, especially without their owners looking at it.

There is a safe version. The parser takes the allowed units as an argument, so each command keeps exactly what it accepts today and each keeps its own error message. Then it is a pure refactor with nothing visible to users, and the existing hosting and firestore tests stay untouched as the proof.

Happy to send that as a small follow up PR if you think it is worth it. I would just want someone from Hosting and Firestore on the review. Or if you would rather leave the three as they are, that is fine with me too. What do you prefer?

PR #10893, opened 2026-08-05
fix(functions): don't orphan the Python discovery server

### LINKED ISSUE #10847: Killed deploy orphans the Python discovery server (serving.py); orphans wedge and later deploys hang forever on connect ETIMEDOUT
### Environment
- firebase-tools: 15.9.1 (repo-pinned devDependency, run from `node_modules/.bin/firebase`)
- Platform: macOS 15 (Apple Silicon), self-hosted GitHub Actions runners
- Project: multi-codebase functions — one nodejs codebase + two Python codebases (`firebase-functions` Python SDK)
- Command: `firebase deploy --only functions --force --project <id>` in CI, with a job-level 60-minute timeout

### Problem
When a `firebase deploy` run is killed externally mid-flight (CI job cancellation or timeout), the Python discovery admin server the CLI spawned (`firebase_functions/private/serving.py`, launched with `ADMIN_PORT=<port>`) is left running. The orphan keeps its admin port bound but eventually stops accepting connections.

Once one or more such orphans exist on the host, **subsequent deploys hang indefinitely** during "Loading and analyzing source code for codebase &lt;python-codebase&gt;": every localhost call to the admin port — including `/__/quitquitquit` — fails with `connect ETIMEDOUT 127.0.0.1:<port>` (timeout, not refused: a bound-but-not-accepting listener), and the CLI never errors out. In our CI this meant each affected deploy ran until the 60-minute job timeout killed it — which orphaned *its* discovery server too. Over 5 days ~35 orphaned `serving.py` processes accumulated on one runner host, at which point every deploy landing on it hung.

Debug log excerpt from a hung deploy (nodejs codebase discovery succeeded in ~1s just before this):

```
[debug] Running admin server with args: ["python3.12","<venv>/site-packages/firebase_functions/private/serving.py"] and env: {..., "ADMIN_PORT":"8120"} in <project>/firebase/functions/python-ytmusicapi
[info]  * Serving Flask app 'serving'
[debug] Failed to call quitquitquit. This often means the server failed to start request to http://127.0.0.1:8120/__/quitquitquit failed, reason: connect ETIMEDOUT 127.0.0.1:8120
[error]  * Running on http://127.0.0.1:8120
```
(no further log lines; the process sat there until the CI job timeout ~59 minutes later)

### Verification
Killing the orphans (`pkill -f "firebase_functions/private/serving.py"` as the runner user) immediately restored ~2-minute deploys **on the same host** — reproduced across two hung runs and one clean rerun.

### Expected behavior
1. The CLI should hard-kill the discovery child process (SIGKILL by pid / process group) when `/__/quitquitquit` is unreachable, and on its own exit paths — an HTTP-only shutdown can never clean up a wedged server.
2. Discovery should fail fast with a clear error when the admin port cannot be reached within a bounded time, instead of hanging the whole deploy indefinitely.

### Workaround
A CI step before `firebase deploy` that pkills stale `serving.py` processes.

IzaakGough:
Fixes #10847 

The Python discovery admin server (`serving.py`) could survive a deploy: `serveAdmin`'s shutdown logic cleared its force-kill timer immediately after setting it, so the `SIGKILL` fallback never ran, and it waited on `exit`/`error` with no timeout. It also only signalled the shell that `runWithVirtualEnv` spawns, not the Python process underneath. An orphaned server kept the discovery port bound, so later deploys hung on `connect ETIMEDOUT`.

Shutdown now escalates properly: `/__/quitquitquit` with a timeout, then a force-kill of the child's process group (`taskkill /T /F` on Windows), with an overall cap so a wedged server can never hang the deploy. The child is spawned detached so the whole group can be killed, and is tracked so `SIGINT`/`SIGTERM`/`SIGHUP` and process exit clean it up if the CLI itself goes away, which is what left servers behind on CI cancellation.


src/functions/python.ts
```diff
@@ -43,3 +43,118 @@ export function runWithVirtualEnv(
...
+  for (const signal of CLEANUP_SIGNALS) {
+    const handler = (): void => {
+      killAllTrackedChildren();
+      // Attaching a signal listener suppresses Node's default "terminate on
+      // signal" behaviour, so restore it: drop our listeners and re-raise, but
+      // only if nobody else (e.g. the emulator's own shutdown handler) is still
+      // listening and expecting to drive the exit itself.
+      removeCleanupHandlers();
+      if (process.listenerCount(signal) === 0) {
+        process.kill(process.pid, signal);
```
cabljac: From reading libuv, `process.kill(pid, "SIGHUP")` throws ENOSYS on Windows (only SIGTERM/SIGKILL/SIGINT/SIGQUIT are supported there), which would crash the handler on console close. @CorieW could you confirm on a Windows box? If it holds, wrap the re-raise in try/catch with a `process.exit(128 + n)` fallback.

src/functions/python.ts
```diff
@@ -43,3 +43,118 @@ export function runWithVirtualEnv(
...
+  if (IS_WINDOWS) {
+    // taskkill /T walks the process tree by parent pid, so it doesn't rely on
+    // the process group trick used below.
+    cp.spawnSync("taskkill", ["/pid", pid.toString(), "/T", "/F"]);
+    return;
+  }
+  try {
+    // A negative pid signals the whole process group rather than just `pid`.
+    process.kill(-pid, "SIGKILL");
+  } catch (e) {
```
cabljac: Worth a `logger.debug` on the dropped taskkill result and the swallowed EPERM - without it the next Windows orphan report is undiagnosable.

src/deploy/functions/runtimes/python/index.spec.ts
```diff
@@ -34,4 +39,167 @@ describe("PythonDelegate", () => {
...
+        signalCode: null,
+        stdout: Object.assign(new EventEmitter(), { destroy: destroyStdoutStub }),
+        stderr: Object.assign(new EventEmitter(), { destroy: destroyStderrStub }),
+        unref: unrefStub,
+      });
+      runWithVirtualEnvStub = sandbox.stub(pythonUtils, "runWithVirtualEnv").returns(child);
+      killProcessTreeStub = sandbox.stub(pythonUtils, "killProcessTree");
+      // Tracking installs real process-level signal handlers; not under test here.
+      sandbox.stub(pythonUtils, "trackVirtualEnvChild");
+      sandbox.stub(pythonUtils, "untrackVirtualEnvChild");
```
cabljac: These stubs are never asserted. Please add: tracked on spawn, untracked on clean exit, still tracked after the give-up path (that last one guards against recreating the orphan).

cabljac (approved):
Approving @IzaakGough, thorough fix. Verified all three root causes from #10847 locally, including reproducing the orphan and confirming the group kill reaps it. 21/21 tests pass.

For the record, two things I checked that are correct as written: resolving `exited` on "error" can't untrack a live child (an exitCode guard would fail your spawn-failure test), and the force-kill timer can't hit a reaped pid.


Nits inline. Also worth filing a follow-up: emulator `startPython` and `modulesDir()` aren't tracked, so they can still orphan the same way.

src/functions/python.ts
IzaakGough: Good catch, thanks. I could not exercise this on a Windows box, so I have not confirmed the ENOSYS throw directly, but the fallback is right either way and costs nothing on platforms that can re-raise.

fd60868 wraps the re-raise in `reRaise()`: on a throw it logs and exits with `128 + os.constants.signals[signal]`, the code a shell would report. Covered by a test that stubs the throw and asserts the exit code.

src/functions/python.ts
IzaakGough: Agreed. 48d25aa logs both: the taskkill exit status and stderr when it does not return 0, and the POSIX kill error, which is normally ESRCH but would be EPERM in the case that actually leaves an orphan. Test covers the EPERM path.

src/deploy/functions/runtimes/python/index.spec.ts
IzaakGough: Added in 57ad4ef, all three: tracked on spawn, untracked once it exits cleanly, and still tracked after the give-up path. I checked each one fails if the behaviour is removed, including untracking the survivor unconditionally, which is the orphan-recreating case you flagged.

IzaakGough: Filed as #11069. I reproduced it against the real emulator rather than just the shell-wrapper mechanism: after a clean shutdown the `functions-framework` process reparents to PID 1 and is still serving HTTP 200 on its port. Scoped to `startPython`, since `modulesDir()` is a short-lived `python -c` that is awaited to exit and binds nothing.

src/deploy/functions/runtimes/python/index.ts
```diff
@@ -161,30 +181,85 @@ export class Delegate implements runtimes.RuntimeDelegate {
...
+      if (exitedCleanly) {
+        untrackVirtualEnvChild(childProcess);
+      } else {
+        // A detached child and its pipes both hold the CLI's event loop open, so
+        // leaving it alive would hang the deploy at process exit instead of here.
+        // Releasing them is also what lets the 'exit' handler run at all, and the
+        // child stays tracked so that handler gets one last attempt at it.
+        childProcess.stdout?.destroy();
+        childProcess.stderr?.destroy();
+        childProcess.unref();
```
ajperel, line 253: 🟡 **Nit: Destroy `stdout`/`stderr` streams unconditionally in `finally`**

*Rationale*: In Node.js `child_process`, the `'exit'` event fires as soon as the direct child (`sh` / `cmd.exe`) terminates (`waitpid`), even if stdio pipes remain open (unlike `'close'`, which waits for all stdio FDs to close). If user code in `main.py` spawns a background subprocess during discovery that inherits `stdout`/`stderr` FDs and outlives `serving.py` after a clean `/__/quitquitquit`, `exited` will resolve (`exitedCleanly === true`), skipping the `else` block. The active `.on("data")` listeners on `childProcess.stdout` and `childProcess.stderr` could then still hold the CLI's event loop open at exit.

Moving `stdout?.destroy()`, `stderr?.destroy()`, and `unref()` into `finally` ensures the parent's pipe handles are released in both paths.

*Suggested Fix*:
```ts
    } finally {
      clearTimeout(forceKill);
      clearTimeout(giveUp);
      childProcess.stdout?.destroy();
      childProcess.stderr?.destroy();
      childProcess.unref();
    }
```

ajperel (commented):
> ⚠️ **Disclaimer:** This draft review was generated by an experimental AI review agent. Please verify all findings before acting on them.

## Code Review Summary: `firebase/firebase-tools` PR #10893

### 🟢 Strengths & LGTM Aspects
- Cleanly addresses all three root causes of orphaned Python discovery servers (`serving.py`): process-group killing via `detached: !IS_WINDOWS` + `killProcessTree(-pid)` / `taskkill /T /F`, lazy process-level signal/exit tracking (`trackVirtualEnvChild` / `untrackVirtualEnvChild`), and bounded shutdown timeouts (`QUITQUITQUIT_TIMEOUT_MS`, `FORCE_KILL_DELAY_MS`, `SHUTDOWN_TIMEOUT_MS`).
- Safe handling of `pid <= 0`, pre-shutdown exits (`once('exit')` attached at spawn time), and Windows `ENOSYS` fallback on `reRaise`.
- Comprehensive unit test suite (24/24 passing) with proper Sinon sandbox and signal listener teardown.

### 🟡 Overview of Findings
- One non-blocking nit in `src/deploy/functions/runtimes/python/index.ts` to destroy `stdout`/`stderr` pipes unconditionally in `finally`.

ajperel (commented):
> ⚠️ **Disclaimer:** This draft review was generated by an experimental AI review agent. Please verify all findings before acting on them.

## Code Review Summary: `firebase/firebase-tools` PR #10893

### 🟢 Strengths & LGTM Aspects
- Cleanly addresses all three root causes of orphaned Python discovery servers (`serving.py`): process-group killing via `detached: !IS_WINDOWS` + `killProcessTree(-pid)` / `taskkill /T /F`, lazy process-level signal/exit tracking (`trackVirtualEnvChild` / `untrackVirtualEnvChild`), and bounded shutdown timeouts (`QUITQUITQUIT_TIMEOUT_MS`, `FORCE_KILL_DELAY_MS`, `SHUTDOWN_TIMEOUT_MS`).
- Safe handling of `pid <= 0`, pre-shutdown exits (`once('exit')` attached at spawn time), and Windows `ENOSYS` fallback on `reRaise`.
- Comprehensive unit test suite (24/24 passing) with proper Sinon sandbox and signal listener teardown.

### 🟡 Overview of Findings
- One non-blocking nit in `src/deploy/functions/runtimes/python/index.ts` to destroy `stdout`/`stderr` pipes unconditionally in `finally`.

joehan (approved):
The change looks good to me, tho this pr is extremely heavily commented to the point that it hinders comprehension imo. I've noticed that a lot lately from Claude - would ask that you have the agent trim down the comments a bit before merging

ajperel (approved):
Between Joe and my AI's confidence.... this seems like a lot but good. Please do address his feedback first.

src/deploy/functions/runtimes/python/index.ts
IzaakGough: Applied in c772bc0806a0645f7c066bfd211d3e7190d031d4

PR #10895, opened 2026-08-06
experimental ext:export support for exporting to functions .env format

Berlioz:


src/extensions/export.ts
```diff
@@ -84,3 +89,59 @@ function displaySpecs(specs: DeploymentInstanceSpec[]): void {
...
+  // Also pull in ALLOWED_EVENTS and EVENTARC_CHANNEL
+  if (typeof instance.config.allowedEventTypes !== "undefined") {
+    envs["EXT_SELECTED_EVENTS"] = instance.config.allowedEventTypes.toString();
+  }
+  if (typeof instance.config.eventarcChannel !== "undefined") {
+    envs["EVENTARC_CHANNEL"] = instance.config.eventarcChannel;
+  }
+
+  return envs;
+}
```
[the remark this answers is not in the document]
Berlioz: What? Am I being gaslit by the bot here? It manifestly does

src/extensions/export.ts
```diff
@@ -3,6 +3,8 @@ import { humanReadable } from "../deploy/extensions/deploymentSummary";
 import { logger } from "../logger";
 import { parseSecretVersionResourceName, toSecretVersionResourceName } from "../gcp/secretManager";
 import { getActiveSecrets } from "./secretsUtils";
+import { ExtensionInstance } from "./types";
```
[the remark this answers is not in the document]
Berlioz: what's that even mean? the linter is very not happy with this suggestion

src/extensions/export.ts
```diff
@@ -84,3 +89,59 @@ function displaySpecs(specs: DeploymentInstanceSpec[]): void {
...
+      envs[specParam.param] = liveParams[specParam.param];
+    } else {
+      envs[specParam.param] = specParam.default ?? "";
+    }
+  });
+
+  // System params aren't necessarily defined in the spec, but we do respect any defaults
+  Object.entries(liveSystemParams).forEach(([sysParamName, sysParamValue]) => {
+    const renamed = sysParamName
+      .replace("firebaseextensions.v1beta.function/", "FIREBASE_SYSTEM_")
```
ajperel: I need to do more thorough review, which I will do tomorrow, but I started by just running this export with an instance of the firestore-to-bigquery extension and it had an error because that extension has a v2 function (one of the few).

Gemini says this and like 132 need to be updated to:

`.replace(/^firebaseextensions\.v1beta\.(v2)?function\//, "FIREBASE_SYSTEM_")`

src/commands/ext-export.ts
```diff
@@ -17,79 +18,159 @@ import { Options } from "../options";
...
   .withForce()
   .action(async (options: Options) => {
-    const projectId = needProjectId(options);
-    const projectNumber = await getProjectNumber(options);
-    // Look up the instances that already exist,
-    // set any secrets to latest version,
-    // and strip project IDs from the param values.
-    // Note that this does not, nor should it include instances defined via SDK.
-    const have = await Promise.all(await planner.have(projectId));
+    if (experiments.isEnabled("internaltesting") && options.mode === "functions") {
```
ajperel: Should this be under it's own experiment? internal testing sounds fairly generic.

Or just grouped with kits CLI under "kits" experiment?

src/commands/ext-export.ts
```diff
@@ -17,79 +18,159 @@ import { Options } from "../options";
...
+      // Functions handler:
+      // - writes to <instanceId>/.env-<projectId>
+      // - does not parametrize project number and ID (e.g "12345678" instead of "{param:PROJECT_NUMBER}")
+      // - explicitly sets unspecified user params to the empty string instead of leaving them out (and causing a prompt on first deploy)
+      // - coerces system param naming format to be valid .env keys (e.g FIREBASE_SYSTEM_MEMORY=256 instead of firebaseextensions.v1beta.function/memory=256)
```
ajperel, line 50: Am I correct in understanding everything with FIREBASE_SYSTEM_* is something that should turn into a `setGlobalOptions()` option? If so few things here:

* Users aren't supposed to set reserved ENV variables though right and we reserve everything under FIREBASE_* so this is a little weird.
* Without some coordination nothing is going to pick this up and do anything with it as we try to preserve behavior. 

We need too coordinate what you do here with our migration instructions and how function kits and ext:migrate and kits:install work.

I think the first piece is we all agree on the ENV variable names that will show up both in the export AND in the `index.js` of a kit. We must have them align so that kits automatically pick these up.

Then we need to make sure the parameters that are set are actually uncommented in index.js since by default we were going to comment out everything but location. Some things I think we could do here:

1) When `kits:install` is called by `ext:migrate` with this data, we know in advance that we want to set these. Thus `install` should make sure those lines aren't commented out.

2) When `kits:install` is called _before_ export when someone is doing a more manual migration then `index.js` will be in a bad state. We _could_ ask users to fix this or we could potentially add a 3rd mode for kits or maybe an additional flag --kit-instance. If you do 

`ext:export --mode functions --instance firestore-bigquery-export --kit-instance foo` then we can modify the behavior in two important ways:
* We can know to put the `.env.<project-id>` file in the kit config directory.
* We could modify the existing index.js to set the right system params.


I'm open to other options, but we do have to at a minimum coordinate on parameter names between exports and kits.

ajperel: It would be ugly in some ways but a way to simplify this would be that we _don't_ use parameters by default for global options in kits except for location since it _always_ must be specified.

For all other system params we look at `process.env.<SYSTEM VARIABLE NAME>` if it exists we use it in our `setGlobalOptions()` call and if it doesn't we don't. 

Thus we don't need any changes to `index.js` to handle the specific options that exist. We could proceed all this with a general comment that's like "if you want to force users to specify these options you can define them as parameters like is done with location".

src/commands/ext-export.ts
```diff
@@ -17,79 +18,159 @@ import { Options } from "../options";
...
+      // - writes references to secrets in the Functions format (e.g FIREBASE_SECRET_REF_API_KEY=foo:latest instead of API_KEY=projects/${param:PROJECT_NUMBER}/secrets/API_KEY/versions/latest)
+      // - makes DeploymentInstanceSpec.eventarcChannel and allowedEventTypes available as FIREBASE_EVENTARC_CHANNEL and EXT_SELECTED_EVENTS
```
ajperel, line 52: These just keeps them working as they did in Extensions if extesions code is migrated directly to kits right?

And in the future we can suggest people follow Thomas's suggestions to just write callbacks instead?

src/commands/ext-export.ts
```diff
@@ -17,79 +18,159 @@ import { Options } from "../options";
...
-    const projectId = needProjectId(options);
-    const projectNumber = await getProjectNumber(options);
-    // Look up the instances that already exist,
-    // set any secrets to latest version,
-    // and strip project IDs from the param values.
-    // Note that this does not, nor should it include instances defined via SDK.
-    const have = await Promise.all(await planner.have(projectId));
+    if (experiments.isEnabled("internaltesting") && options.mode === "functions") {
+      // Functions handler:
+      // - writes to <instanceId>/.env-<projectId>
```
ajperel, line 47: I think the two most common cases are:

(1) Thomas calls the guts of this in ext:migrate and we never write to file
(2) We want to write to a known location that isn't <instanceId>/.env-projectid   but that's based on kit configDir.

And as a rarer (3) --- saving .env values for future use but not migrating to a kit now. 

I wonder if we should handle this differently:
* I assume we can't use stout because we have other log lines so something like `firebase ext:export <flags> > <path/to/file>` won't work?
* Have a flag that just specifies the directory to write .env file in?
* Have a flag that specifies the kit instance and we use that to look up `configDir`?

In the worst case I think instructions woudl be something like:

1. Run `firebase ext:export`
2. `mv <instance>/.env-<projectId> <path/to/configDir>`
3. `rm -r <instance>`

src/commands/ext-export.ts
ajperel, line 52: And these will continue to just work for Extensions with these names right?

And if we want in the future we can suggest people stop this pattern and just do callbacks but it's not required for migration?

src/commands/ext-export.ts
ajperel, line 50: Ok we need to coordinate with @wandamora and @inlined here. IIUC all the FIREBASE_SYSTEM_* variables need to be set using `setGlobalOptions` in index.js and this won't happen automatically without coordination on our part.

At the very least we need to agree on the names and use the same ones in `index.js` generated by `kits:install`

But more than that... the plan of record today was to _only_ have the location parameter set in `index.js` and users could hand modify `index.js` if they want to set more. But I don't want to require hand modification during our migration flow. I think we have a few options we can discuss.

### As designed today
We thought we'd have all system parameters as... actually defined function params. If we do that we can only define them _if_ they are set. We were going to have all unused ones commented out. If we keep this plan we need some way to make sure the ones used by a given extension instance are uncommented and set appropriate.

1. `ext:migrate` can pass them to `kits:install` and install could make this change.
2. when doing a multiple command migration either the user would have to do this by hand (bad) or we could modify `ext:export` to know when it's replacing a particular kit instance and both write to the right `configDir` and update `index.js`
3. We make users hand modify `index.js` after exporting. :(
4. Something good y'all suggest???I really don't love

Unfortunately I don't love 2 or 3 but want us to have that as a well supported path where we don't use `ext:migrate`. Especially for anyone who is forking an extension a publisher didn't replace.

### Alternative
We could instead change `index.js` template to support this more gracefully.

* Only Location becomes a param because we _always_ want it explicitly chosen.
* For all other global options we have a reserved ENV variable we check. If present at runtime in .env we include it in setGlobalOptions, otherwise we do not. Then we don't need any uncommenting in `index.js` just the existing in `process.env` is sufficient.
* We could add a comment that if users want to _force_ all kit instances to specify one of these variables they can define a param like is done for location.

src/commands/ext-export.ts
```diff
@@ -17,79 +18,159 @@ import { Options } from "../options";
...
-      }
-      return {
-        instanceId: spec.instanceId,
-        ref: spec.ref,
-        params: buildBindingOptionsWithBaseValue(paramCopy),
-      };
-    });
+async function fnHandler(options: Options) {
+  if (!options.instance) {
+    logger.info(`ext:export must be scoped to a specific instance when exporting to Functions`);
```
ajperel: Do you think we should suggest running ext:list to see possible instances?

ajperel: Also maybe be a bit more specific on how shoudl we note the `--instance <instance id>` flag explicitly?

src/extensions/export.ts
Berlioz: gah, yeah that make sense, didn't know v2 was on the cards

src/commands/ext-export.ts
```diff
@@ -17,79 +18,159 @@ import { Options } from "../options";
...
-      if (spec.allowedEventTypes) {
-        paramCopy.ALLOWED_EVENT_TYPES = spec.allowedEventTypes.join(",");
-      }
-      return {
-        instanceId: spec.instanceId,
-        ref: spec.ref,
-        params: buildBindingOptionsWithBaseValue(paramCopy),
-      };
-    });
+async function fnHandler(options: Options) {
```
ajperel: Just to check.... will this handle both console installed extensions and CLI installed extensions?

src/commands/ext-export.ts
```diff
@@ -17,79 +18,159 @@ import { Options } from "../options";
...
+  const projectId = needProjectId(options);
+  const instances = await listInstances(projectId);
+  if (instances.length < 1) {
+    logger.info(`No extension instances installed on ${projectId}, so there is nothing to export.`);
+    return;
+  }
 
-    saveEtags(options.rc, projectId, have);
+  let found = false;
+  instances.forEach((instance) => {
```
ajperel: Can you simplify this a bit and maybe reduce nesting by doing something like:

```
const instance = instances.find((i) => {
    const instanceId = last(i.name.split("/")) ?? "";
    return instanceId === options.instance;
  });

if (!instance) {
  logger.info(`No extensions instances found matching instance ID ${options.instance}`);
  // exit
}

// else all the stuff you do?
```

src/extensions/export.ts
```diff
@@ -44,6 +46,9 @@ export async function setSecretParamsToLatest(
   return { ...spec, params: newParams };
 }
 
+/**
```
ajperel, line 49: Why add empty comment?

src/extensions/export.ts
```diff
@@ -84,3 +89,59 @@ function displaySpecs(specs: DeploymentInstanceSpec[]): void {
...
+ * Translates a currently deployed Extension instance into a Functions environment.
+ * This includes setting any default params not set in the deployed instance to their
+ * default value, writing any system params under the reserved FIREBASE_SYSTEM_ prefix,
+ * and writing any secret references under the reserved FIREBASE_SECRET_REF_ prefix.
+ */
+export function functionsEnvFromInstance(instance: ExtensionInstance): Record<string, string> {
+  const liveParams = instance.config?.params || {};
+  const liveSystemParams = instance.config?.systemParams || {};
+  const specParams = instance.config?.source?.spec?.params || {};
```
ajperel, line 102: (1) Gemini I think correctly finds for me that your fallback should be `[]` and not `{}` if you're going to call forEach on it?

(2) When will this stuff not exist and do we need to do more to fetch it in that case? It'd be bad if this ever silently writes a bad `.env` file because it couldn't find something?  Would erroring be better?

ajperel (commented):
Ok. Overall pretty good. I think the big things are:

(1) Handling v2 system params that I commented on yesterday
(2) We need to coordinate on export vs. kits `index.js` so everything "just works" after a migration.
(3) Simiarly we should talk more about where / controlling where this env file is written.

But a few other comments throughout

src/extensions/export.ts
Berlioz: uh the linter did when asked to --fix, oops

src/commands/ext-export.ts
Berlioz: Either would do, I think. Internaltesting is the "standard" choice when committing code before api review, but I think any non-listed experiement would be fine.

inlined: I can see an argument for kits in general as one experiment and extensions migration as another, but I agree that "internaltesting" is not the right one. That experiment isn't for "experimentation" IIRC, it's for exposing commands/features that we use to test our product.

src/commands/ext-export.ts
inlined: Note that if we _do_ preserve the existing behavior we're going to have to do some additional work too. Extensions upserts the FIREBASE_EVENTARC_CHANNEL

src/commands/ext-export.ts
inlined: I think system parameters may need some additional thinking. I'm not actually sure they should be set to env vars in a normal export. The point of system env vars is that they're not bound to anything in the customer's code, which is why I had planned to use setGlobalOptions. In fact, in Extensions they're not even available at runtime!

I really really really really don't think we should get in the business of "modifying" JS. Creating a file is easy. Modifying it means AST parsing.

I think I see a couple of paths in the road:

1. The library version of this creates two objects: env and system params. We maybe write both files? Or maybe we just print "The following system parameters were set, but are not part of your functions' environment". We can even indicate how to set them with a setGlobalOptions snippet. ext:migrate uses this to make sure the initial code works as advertised though
2. We make FIREBASE_SYSTEM_* another pinhole in our reserved keywords. It's a generally undocumented feature that we know how to use those env vars even though they're not valid params?
3. I give in and we have another config file. Something like container.<project>.yaml that can specify system parameters outside of code.

src/commands/ext-export.ts
```diff
@@ -17,79 +18,159 @@ import { Options } from "../options";
...
+      // - makes DeploymentInstanceSpec.eventarcChannel and allowedEventTypes available as FIREBASE_EVENTARC_CHANNEL and EXT_SELECTED_EVENTS
+      await fnHandler(options);
+    } else {
+      // Extensions handler:
+      // - does none of that
+      await extHandler(options);
+    }
+  });
 
+async function extHandler(options: Options) {
```
inlined: This is a pretty easy lint error to fix and generally I agree with it.

src/commands/ext-export.ts
inlined: Also .forEach is against style rules IIRC.

src/commands/ext-export.ts
```diff
@@ -17,79 +18,159 @@ import { Options } from "../options";
...
+  let found = false;
+  instances.forEach((instance) => {
+    const instanceId = last(instance.name.split("/")) ?? "";
+    if (instanceId !== options.instance) {
+      return;
+    }
+
+    found = true;
+    const convertedEnv = functionsEnvFromInstance(instance);
+    for (const key of Object.keys(convertedEnv)) {
```
inlined: Is this just for debug purposes?

src/extensions/export.ts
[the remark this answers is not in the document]
inlined: I assume it thinks you're supposed to use this somewhere, but I'm not sure where.

src/extensions/export.ts
```diff
@@ -84,3 +89,59 @@ function displaySpecs(specs: DeploymentInstanceSpec[]): void {
...
+      envs[renamed] = liveParams[specParam.param];
+    } else if (specParam.param in liveParams) {
+      envs[specParam.param] = liveParams[specParam.param];
+    } else {
+      envs[specParam.param] = specParam.default ?? "";
+    }
+  });
+
+  // System params aren't necessarily defined in the spec, but we do respect any defaults
+  Object.entries(liveSystemParams).forEach(([sysParamName, sysParamValue]) => {
```
inlined: Please rewrite these more canonically as:

```
for (const [sysParamName, sysParamValue] of Object.entries(liveSystemParams)) {
// ...
}
```

There's no need to have a closure or not use built-ins. Going back to language built-ins gives you access to continue and break, doesn't change the meaning of return, and is easily extended to async iteration.

src/extensions/export.ts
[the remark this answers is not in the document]
inlined: 1. I think this is why it's telling you to add the import above, because it wants you to use that type here.
2. What do you mean "it manifestly does". It does return the fields that it's saying are missing? Might be worth pinging an Extensions person to see if API docs are wrong or something.

src/commands/ext-export.ts
```diff
@@ -17,79 +18,159 @@ import { Options } from "../options";
...
-      manifestSpecs,
-      existingConfig,
-      {
-        nonInteractive: options.nonInteractive,
-        force: options.force,
-      },
-      true /** allowOverwrite */,
-    );
+  const projectId = needProjectId(options);
+  const instances = await listInstances(projectId);
```
inlined: Stupid Q, but why are we listing instances and then filtering for the one with this ID instead of just using a getInstance?

inlined: Yes, I see that there is not a getIsntance method, but API wrappers are trivial

inlined (commented):
Need to think much deeper about eventarc and system params unfortunately.

inlined:
After talking with Andy he explained an idea that makes sense to me. Don't make system params special, lean on the fact that we're gong to make process.env available at discovery. So a migrated template can have something like:

```
const region = defineString("REGION");

setGlobalOptions({
  region,

  // These may have been set from the original extension. Feel free
  // to replace these values with literals or parameters. Because these were originally
  // advanced parameters in Extensions which did not require much focus they are
  // added here in a way that works if set but does not flood users with prompts
  memory: process.env.ENV_EXPORT_MEMORY ?? null,
  concurrency: process.env.ENV_EXPORT_CONCURRENCY ?? null,
  // ...
  });
  ```

File: src/commands/ext-export.ts (hunk 16 above)
Comment (Berlioz) [reply src/commands/ext-export.ts]: There actually is a getInstance method already, which makes this even simpler.

File: src/commands/ext-export.ts (hunk 14 above)
Comment (Berlioz) [reply src/commands/ext-export.ts]: The extensions version of this command writes the config to stdout, so I figured we may as well be as consistent as possible with that version. And hey, it does help with debugging.

File: src/extensions/export.ts (hunk 12 above)
Comment (Berlioz) [reply src/extensions/export.ts:102]: (1) These consts are all of type Record<string, string>, so {} here is correct. I think forEach isn't idiomatic on records, as Thomas flagged later, which is probably what confused Gemini.

(2) I mean, these are coming from the results of a getInstance API call. If the API call errors out, we never make it here.

File: src/commands/ext-export.ts (hunk 10 above)
Comment (Berlioz) [reply src/commands/ext-export.ts]: I feel like someday I will learn this lesson for good, but today is apparently not that lesson.

File: src/commands/ext-export.ts (hunk 9 above)
Comment (Berlioz) [reply src/commands/ext-export.ts]: Yeah, just tested.

File: src/commands/ext-export.ts (hunk 8 above)
Comment (Berlioz) [reply src/commands/ext-export.ts]: Ack

File: src/extensions/export.ts (hunk 1 above)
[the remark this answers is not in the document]
Comment (Berlioz) [reply src/extensions/export.ts:147]: Yeah, the fields are absolutely in the API response. I think Gemini is confused because the Extensions export path uses the `planner.have` helper to retrieve ExtensionInstances via the same API method, but that helper explicitly strips out the extension `spec` since it isn't relevant to the question of "what do I currently have."

This new kits export path doesn't use that helper, so the spec is present.

File: src/extensions/export.ts (hunk 15 above)
Comment (Berlioz) [reply src/extensions/export.ts]: Ack

**Review by ajperel — COMMENTED** [review]
running this for `firestore-bigquery-export` I get:

`EXT_MIGRATED_SYSTEM_VPCCONNECTOREGRESSSETTINGS=VPC_CONNECTOR_EGRESS_SETTINGS_UNSPECIFIED`

In the output which feels like it should be omitted?

(edit from Berlioz, who can't figure out how to reply to this in the UI: I think this is actually okay? It just looks terrible because this is an enum in the backing proto so the zero value can't just be zero. we can special-case this later if usability turns out to be an issue)

**Comment by ajperel** [comment]
Also in my test with `firestore-bigquery-export` I get

`EXT_MIGRATED_SYSTEM_LOCATION=us-west1`

Didn't we agree this one we should special case to `DEFAULT_FUNCTION_REGION=<region>` to match what we'll parameterize in kits?

File: src/commands/ext-export.ts [hunk 17]
```diff
@@ -17,79 +18,173 @@ import { Options } from "../options";
...
-    // set any secrets to latest version,
-    // and strip project IDs from the param values.
-    // Note that this does not, nor should it include instances defined via SDK.
-    const have = await Promise.all(await planner.have(projectId));
+    if (experiments.isEnabled("extMigrationFeatures") && options.mode === "functions") {
+      // Functions handler:
+      // - writes to <instanceId>/.env-<projectId>
+      // - does not parametrize project number and ID (e.g "12345678" instead of "{param:PROJECT_NUMBER}")
+      // - explicitly sets unspecified user params to the empty string instead of leaving them out (and causing a prompt on first deploy)
+      // - coerces system param naming format to be valid .env keys (e.g FIREBASE_SYSTEM_MEMORY=256 instead of firebaseextensions.v1beta.function/memory=256)
```
ajperel, line 50: update now that it's not FIREBASE_SYSTEM

src/commands/ext-export.ts
ajperel: We should still talk about the right UX here.

src/extensions/export.ts
ajperel: I'd either revert, or better yet add a basic description

src/extensions/export.ts
ajperel: Ok

ajperel (approved):
Approving but please fix location and if needed that VPC_CONNECTOR thing before landing. Changing the UX to write to different files locations I think can be a follow up.

PR #10900, opened 2026-08-07
feat(functions): add functions:kits:install command

wandamora:
### Description

Adds the `firebase functions:kits:install` command (gated behind the `kits` experiment) to allow developers to install and configure reusable Cloud Function kits in their Firebase projects.

Note: this is the first iteration of the command that just handles the first instance in a single project.

**Key Changes:**
- **Command Registration & Experiment Gating:** Registers `functions:kits:install` conditionally when the `kits` experiment is enabled.
- **Interactive & Flag Input:** Accepts a `--npm_package <package>` flag, and interactively prompts for the kit and initial instance ID.
- **Package Parsing & Validation:**
  - Parses scoped and unscoped package specifiers with optional versions/tags.
  - Sanitizes package names into valid kit identifiers.
  - Validates kit ID uniqueness and ensures instance IDs do not collide with existing codebase names or other kit instances in `firebase.json`.
- **Security & Third-Party Safeguards:**
  - Differentiates first-party (`@firebase-functions-kits/*`) and third-party packages.
  - Inspects packages for `npm-shrinkwrap.json` via `npm pack --dry-run --json` to warn users when dependencies are unlocked.
  - Prompts for explicit user confirmation before installing third-party kits.
  - Installs third-party packages with `--ignore-scripts` for safety.
- **Wrapper Project Scaffolding:**
  - Scaffolds a wrapper directory under `function-kits/<kit-id>/` containing `package.json`, `tsconfig.json`, `.gitignore`, and `src/index.ts` from templates.
  - Generates `index.ts` using `templates/init/functions/typescript/index-kit.ts`, re-exporting the kit package with customizable global function parameters.
  - Automatically runs `npm install` and `npm run build`.
- **Configuration Updates:** Adds the kit configuration block and instance mapping into `firebase.json`.
- **Testing:** Comprehensive unit test suite in `src/commands/functions-kits-install.spec.ts`.

### Scenarios Tested

- Experiment gating: Verified error is thrown when `kits` experiment is disabled.
- Precondition checks: Verified error when run outside a Firebase project directory (`firebase.json` missing).
- Package specifier parsing for scoped/unscoped packages with and without versions.
- Package name sanitization to valid kit identifiers (length truncation, character filtering).
- Third-party package detection and warnings for packages without `npm-shrinkwrap.json`.
- User confirmation flow cancellation / acceptance for third-party kits.
- Validated that third-party kits run `npm install --ignore-scripts`.
- Kit installation end-to-end scaffolding (`package.json`, `tsconfig.json`, `index.ts`, `firebase.json`).
- Conflict validations:
  - Duplicate `functions.kit` ID in `firebase.json`.
  - Collision between kit `instanceId` and existing function `codebase` name.

### Sample Commands

```bash
# Enable the kits experiment
firebase experiments:enable kits

# Install a package CLI
firebase functions:kits:install --npm_package @invertase/example-firebase-kit-hello-world@0.1.0
```

### DIFF
--- a/src/commands/functions-kits-install.spec.ts
+++ b/src/commands/functions-kits-install.spec.ts
@@ -5,7 +5,8 @@ import * as fs from "fs-extra";
 
 import {
   command,
-  parsePackageSpecifier,
+  parseNpmPackageSpecifier,
+  validateNpmPackageName,
   sanitizePackageNameToKitName,
   isThirdPartyPackage,
   checkPackageHasShrinkwrap,
@@ -40,32 +41,107 @@ describe("functions:kits:install", () => {
     sinon.restore();
   });
 
-  describe("parsePackageSpecifier", () => {
+  describe("validateNpmPackageName", () => {
+    it("should accept valid unscoped package names", () => {
+      expect(() => validateNpmPackageName("my-kit")).to.not.throw();
+      expect(() => validateNpmPackageName("firestore-export")).to.not.throw();
+      expect(() => validateNpmPackageName("kit_123.v1")).to.not.throw();
+    });
+
+    it("should accept valid scoped package names with exactly one slash", () => {
+      expect(() =>
+        validateNpmPackageName("@firebase-functions-kits/firestore-bigquery-export"),
+      ).to.not.throw();
+      expect(() => validateNpmPackageName("@invertase/example-kit")).to.not.throw();
+    });
+
+    it("should reject package names with multiple slashes", () => {
+      expect(() => validateNpmPackageName("@scope/pkg/extra")).to.throw(
+        FirebaseError,
+        /Invalid NPM package name/,
+      );
+      expect(() => validateNpmPackageName("foo/bar/baz")).to.throw(
+        FirebaseError,
+        /Invalid NPM package name/,
+      );
+    });
+
+    it("should reject unscoped package names containing slashes", () => {
+      expect(() => validateNpmPackageName("foo/bar")).to.throw(
+        FirebaseError,
+        /Invalid NPM package name/,
+      );
+    });
+
+    it("should reject empty or malformed package names", () => {
+      expect(() => validateNpmPackageName("")).to.throw(FirebaseError, /Invalid NPM package name/);
+      expect(() => validateNpmPackageName("@scope")).to.throw(
+        FirebaseError,
+        /Invalid NPM package name/,
+      );
+      expect(() => validateNpmPackageName("a".repeat(215))).to.throw(
+        FirebaseError,
+        /Invalid NPM package name/,
+      );
+    });
+  });
+
+  describe("parseNpmPackageSpecifier", () => {
     it("should parse scoped package with version", () => {
-      const res = parsePackageSpecifier("@firebase-functions-kits/firestore-bigquery-export@1.0.0");
+      const res = parseNpmPackageSpecifier(
+        "@firebase-functions-kits/firestore-bigquery-export@1.0.0",
+      );
       expect(res).to.deep.equal({
         packageName: "@firebase-functions-kits/firestore-bigquery-export",
         version: "1.0.0",
       });
     });
 
+    it("should parse scoped package with release candidate version", () => {
+      const res = parseNpmPackageSpecifier(
+        "@firebase-functions-kits/firestore-bigquery-export@1.0.0-rc.1",
+      );
+      expect(res).to.deep.equal({
+        packageName: "@firebase-functions-kits/firestore-bigquery-export",
+        version: "1.0.0-rc.1",
+      });
+    });
+
+    it("should parse scoped package with tag", () => {
+      const res = parseNpmPackageSpecifier(
+        "@firebase-functions-kits/firestore-bigquery-export@latest",
+      );
+      expect(res).to.deep.equal({
+        packageName: "@firebase-functions-kits/firestore-bigquery-export",
+        version: "latest",
+      });
+    });
+
     it("should parse scoped package without version", () => {
-      const res = parsePackageSpecifier("@firebase-functions-kits/firestore-bigquery-export");
+      const res = parseNpmPackageSpecifier("@firebase-functions-kits/firestore-bigquery-export");
       expect(res).to.deep.equal({
         packageName: "@firebase-functions-kits/firestore-bigquery-export",
       });
     });
 
     it("should parse non-scoped package with version", () => {
-      const res = parsePackageSpecifier("my-kit@^2.0.0");
+      const res = parseNpmPackageSpecifier("my-kit@^2.0.0");
       expect(res).to.deep.equal({
         packageName: "my-kit",
         version: "^2.0.0",
       });
     });
 
+    it("should parse non-scoped package with tag", () => {
+      const res = parseNpmPackageSpecifier("my-kit@next");
+      expect(res).to.deep.equal({
+        packageName: "my-kit",
+        version: "next",
+      });
+    });
+
     it("should parse non-scoped package without version", () => {
-      const res = parsePackageSpecifier("my-kit");
+      const res = parseNpmPackageSpecifier("my-kit");
       expect(res).to.deep.equal({
         packageName: "my-kit",
       });
@@ -94,10 +170,12 @@ describe("functions:kits:install", () => {
   describe("isThirdPartyPackage", () => {
     it("should return false for packages under @firebase-functions-kits scope", () => {
       expect(isThirdPartyPackage("@firebase-functions-kits/firestore-bigquery-export")).to.be.false;
-      expect(isThirdPartyPackage("@firebase-functions-kits")).to.be.false;
     });
 
     it("should return true for packages outside @firebase-functions-kits scope", () => {
+      expect(isThirdPartyPackage("firebase-functions-kits")).to.be.true;
+      expect(isThirdPartyPackage("@firebase-function-kits-fake/foo")).to.be.true;
+      expect(isThirdPartyPackage("@firebase-functions-kits-fake/foo")).to.be.true;
       expect(isThirdPartyPackage("@other-scope/my-kit")).to.be.true;
       expect(isThirdPartyPackage("third-party-kit")).to.be.true;
     });
@@ -173,6 +251,25 @@ describe("functions:kits:install", () => {
         FirebaseError,
         /set the --npm_package option to a valid NPM package and try again\./,
       );
+    });
+
+    it("should throw an error if --npm_package has an invalid package name", async () => {
+      const mockConfig = {
+        projectDir: "/mock/project",
+        src: {
+          functions: [],
+        },
+        path: (p: string) => path.join("/mock/project", p),
+      } as unknown as Config;
+
+      await expect(
+        command.runner()({
+          npm_package: "@scope/pkg/extra@1.0.0",
+          cwd: "/mock/project",
+          config: mockConfig,
+          nonInteractive: true,
+        }),
+      ).to.be.rejectedWith(FirebaseError, /Invalid NPM package name/);
     });
 
     it("should successfully install a first-party kit into firebase.json", async () => {
@@ -380,6 +477,85 @@ describe("functions:kits:install", () => {
           nonInteractive: true,
         }),
       ).to.be.rejectedWith(FirebaseError, /must be mutually exclusive/);
+    });
+
+    it("should reject instance ID that collides with another kit instance ID", async () => {
+      const mockConfig = {
+        projectDir: "/mock/project",
+        src: {
+          functions: [
+            {
+              kit: "other-kit",
+              source: "function-kits/other-kit",
+              instances: {
+                "my-kit": "function-kits/other-kit/config-my-kit",
+              },
+            },
+          ],
+        },
+        path: (p: string) => path.join("/mock/project", p),
+        writeProjectFile: sinon.stub(),
+      } as unknown as Config;
+
+      await expect(
+        command.runner()({
+          npm_package: "@firebase-functions-kits/my-kit",
+          cwd: "/mock/project",
+          config: mockConfig,
+          nonInteractive: true,
+        }),
+      ).to.be.rejectedWith(
+        FirebaseError,
+        /functions kit instance ID must be unique across all kits, but 'my-kit' was used more than once/,
+      );
+    });
+
+    it("should prompt confirmation when a first-party kit lacks npm-shrinkwrap.json", async () => {
+      spawnWithOutputStub.resolves(JSON.stringify([{ files: [{ path: "package.json" }] }]));
+      const confirmStub = sinon.stub(prompt, "confirm").resolves(true);
+
+      const mockConfig = {
+        projectDir: "/mock/project",
+        src: { functions: [] },
+        path: (p: string) => path.join("/mock/project", p),
+        writeProjectFile: sinon.stub(),
+        askWriteProjectFile: sinon.stub().resolves(),
+      } as unknown as Config;
+
+      await command.runner()({
+        npm_package: "@firebase-functions-kits/my-kit",
+        cwd: "/mock/project",
+        config: mockConfig,
+        nonInteractive: true,
+      });
+
+      expect(confirmStub).to.have.been.calledOnceWith({
+        message:
+          "Are you sure you want to install @firebase-functions-kits/my-kit without locked dependencies?",
+        default: false,
+        nonInteractive: true,
+      });
+    });
+
+    it("should cancel installation if user declines confirmation for missing shrinkwrap", async () => {
+      spawnWithOutputStub.resolves(JSON.stringify([{ files: [{ path: "package.json" }] }]));
+      sinon.stub(prompt, "confirm").resolves(false);
+
+      const mockConfig = {
+        projectDir: "/mock/project",
+        src: { functions: [] },
+        path: (p: string) => path.join("/mock/project", p),
+        writeProjectFile: sinon.stub(),
+      } as unknown as Config;
+
+      await expect(
+        command.runner()({
+          npm_package: "@firebase-functions-kits/my-kit",
+          cwd: "/mock/project",
+          config: mockConfig,
+          nonInteractive: true,
+        }),
+      ).to.be.rejectedWith(FirebaseError, "Installation cancelled.");
     });
   });
 });
--- a/src/commands/functions-kits-install.spec.ts
+++ b/src/commands/functions-kits-install.spec.ts
@@ -1,0 +1,330 @@
+import { expect } from "chai";
+import * as sinon from "sinon";
+import * as path from "path";
+import * as fs from "fs-extra";
+
+import {
+  command,
+  parsePackageSpecifier,
+  sanitizePackageNameToKitName,
+  isThirdPartyPackage,
+  checkPackageHasShrinkwrap,
+} from "./functions-kits-install";
+import * as experiments from "../experiments";
+import * as initSpawn from "../init/spawn";
+import { Config } from "../config";
+import { FirebaseError } from "../error";
+
+describe("functions:kits:install", () => {
+  let assertEnabledStub: sinon.SinonStub;
+  let wrapSpawnStub: sinon.SinonStub;
+  let spawnWithOutputStub: sinon.SinonStub;
+
+  beforeEach(() => {
+    (command as unknown as { befores: unknown[] }).befores = [];
+    sinon.stub(command, "prepare").resolves();
+    assertEnabledStub = sinon.stub(experiments, "assertEnabled");
+    wrapSpawnStub = sinon.stub(initSpawn, "wrapSpawn").resolves();
+    spawnWithOutputStub = sinon
+      .stub(initSpawn, "spawnWithOutput")
+      .resolves(JSON.stringify([{ hasShrinkwrap: true }]));
+    sinon.stub(fs, "ensureDir").resolves();
+    sinon.stub(fs, "pathExists").resolves(false);
+    sinon.stub(fs, "readJson").resolves({});
+    sinon.stub(fs, "writeJson").resolves();
+    sinon.stub(fs, "writeFile").resolves();
+  });
+
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  describe("parsePackageSpecifier", () => {
+    it("should parse scoped package with version", () => {
+      const res = parsePackageSpecifier("@firebase-functions-kits/firestore-bigquery-export@1.0.0");
+      expect(res).to.deep.equal({
+        packageName: "@firebase-functions-kits/firestore-bigquery-export",
+        version: "1.0.0",
+      });
+    });
+
+    it("should parse scoped package without version", () => {
+      const res = parsePackageSpecifier("@firebase-functions-kits/firestore-bigquery-export");
+      expect(res).to.deep.equal({
+        packageName: "@firebase-functions-kits/firestore-bigquery-export",
+      });
+    });
+
+    it("should parse non-scoped package with version", () => {
+      const res = parsePackageSpecifier("my-kit@^2.0.0");
+      expect(res).to.deep.equal({
+        packageName: "my-kit",
+        version: "^2.0.0",
+      });
+    });
+
+    it("should parse non-scoped package without version", () => {
+      const res = parsePackageSpecifier("my-kit");
+      expect(res).to.deep.equal({
+        packageName: "my-kit",
+      });
+    });
+  });
+
+  describe("sanitizePackageNameToKitName", () => {
+    it("should extract kit name from scoped package name", () => {
+      expect(
+        sanitizePackageNameToKitName("@firebase-functions-kits/firestore-bigquery-export"),
+      ).to.equal("firestore-bigquery-export");
+      expect(sanitizePackageNameToKitName("@foo/bar")).to.equal("bar");
+    });
+
+    it("should sanitize non-scoped package name", () => {
+      expect(sanitizePackageNameToKitName("my-kit")).to.equal("my-kit");
+      expect(sanitizePackageNameToKitName("My_Kit!")).to.equal("my_kit");
+    });
+
+    it("should truncate long names to 40 characters", () => {
+      const longName = "@scope/" + "a".repeat(50);
+      expect(sanitizePackageNameToKitName(longName)).to.equal("a".repeat(40));
+    });
+  });
+
+  describe("isThirdPartyPackage", () => {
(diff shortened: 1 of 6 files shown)


src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,292 @@
...
+      message: "What would you like to name your first instance of this kit?",
+      default: kitId,
+      nonInteractive: options.nonInteractive,
+    });
+
+    const sourcePath = path.join("function-kits", kitId, "src");
+    const configDirPath = path.join("function-kits", kitId, `config-${instanceId}`);
+
+    let existingFunctions: ValidatedConfig | [] = [];
+    const configFunctions = options.config.src?.functions;
```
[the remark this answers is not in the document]
wandamora: src is not expected to be undefined or null, so the ? is redundant. removed.

src/commands/functions-kits-install.spec.ts
```diff
@@ -0,0 +1,385 @@
...
+      expect(res).to.deep.equal({
+        packageName: "@firebase-functions-kits/firestore-bigquery-export",
+      });
+    });
+
+    it("should parse non-scoped package with version", () => {
+      const res = parsePackageSpecifier("my-kit@^2.0.0");
+      expect(res).to.deep.equal({
+        packageName: "my-kit",
+        version: "^2.0.0",
```
ajperel, line 154: Probably also good to test release candidate versions since we'll use them

src/commands/functions-kits-install.spec.ts
```diff
@@ -0,0 +1,385 @@
...
+  });
+
+  describe("isThirdPartyPackage", () => {
+    it("should return false for packages under @firebase-functions-kits scope", () => {
+      expect(isThirdPartyPackage("@firebase-functions-kits/firestore-bigquery-export")).to.be.false;
+      expect(isThirdPartyPackage("@firebase-functions-kits")).to.be.false;
+    });
+
+    it("should return true for packages outside @firebase-functions-kits scope", () => {
+      expect(isThirdPartyPackage("@other-scope/my-kit")).to.be.true;
```
ajperel, line 201: I'd test "@firebase-function-kits-fake/foo" to also be 3rd party. Gotta stop the people trying to be malicious.

src/commands/functions-kits-install.spec.ts
```diff
@@ -0,0 +1,385 @@
...
+        command.runner()({
+          npm_package: "@firebase-functions-kits/firestore-bigquery-export",
+          cwd: "/mock/project",
+          config: mockConfig,
+          nonInteractive: true,
+        }),
+      ).to.be.rejectedWith(FirebaseError, /functions.kit must be unique/);
+    });
+
+    it("should reject instance ID that collides with codebase name", async () => {
```
ajperel: Even now we should also test and reject instance ID that collides with another since you could have two different kits that decide to use the same instance id if the user picks silly ones.

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+
+const PACKAGE_NO_LINTING_TEMPLATE = readTemplateSync(
+  "init/functions/typescript/package.nolint.json",
+);
+const TSCONFIG_TEMPLATE = readTemplateSync("init/functions/typescript/tsconfig.json");
+const GITIGNORE_TEMPLATE = readTemplateSync("init/functions/typescript/_gitignore");
+const INDEX_KIT_TEMPLATE = readTemplateSync("init/functions/typescript/index-kit.ts");
+
+export interface FunctionsKitsInstallOptions extends Options {
+  npm_package?: string;
```
ajperel, line 36: As I see this.... I am again debating between this and

--package
(future) --package_manager

Pros:
* We only ever need 2 flags (I hope)
* Maybe easier to re-use logic in the future in cases where like... yarn packages and npm packages are the same since they are both npm behind the scenes

Cons:
* Users not using npm have to always specify two flags instead of one.

I am still very torn. Curious what you and @inlined think after getting into it more.

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+  }
+  return { packageName: rawPkg };
+}
+
+/**
+ * Sanitizes an npm package name into a valid kit identifier.
+ * e.g., "@firebase-functions-kits/firestore-bigquery-export" -> "firestore-bigquery-export"
+ */
+export function sanitizePackageNameToKitName(packageName: string): string {
+  const parts = packageName.split("/");
```
ajperel, line 94: I think this is true but you've verified that package names can only have one "/" and only if scoped right?

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+  const nameWithoutScope = parts[parts.length - 1] || packageName;
+  const sanitized = nameWithoutScope
+    .toLowerCase()
+    .replace(/^@/, "")
```
ajperel: technically you shouldn't need this since the next line will also remove it right?

src/commands/functions-kits-install.spec.ts
ajperel: or tags like @latest

technically npm distinguished between versions and tags but maybe we can just call it all version and get away with that?

Thoughts @inlined ?

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+  return (sanitized || "kit").slice(0, 40);
+}
+
+/**
+ * Checks if a package name is third-party (outside the @firebase-functions-kits scope).
+ */
+export function isThirdPartyPackage(packageName: string): boolean {
+  return (
+    !packageName.startsWith("@firebase-functions-kits/") &&
+    packageName !== "@firebase-functions-kits"
```
ajperel: I don't think this is a valid package name?

`@firebase-function-kits` implies a _scope_ which must be followed by a package I think?

`firebase-function-kits` could be a package (I don't know if npm lets scope names and package names conflict)

but either way we don't plan on using `firebase-function-kits` the plain package so for now if we haven't put it explicitly in our control I'd say we should consider it 3P.

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+      }
+    }
+  } catch (err: unknown) {
+    logger.debug(`Failed to inspect package shrinkwrap for ${rawPkgName}: ${getErrMsg(err)}`);
+  }
+  return false;
+}
+
+export const command = new Command("functions:kits:install")
+  .description("install a Cloud Function kit into your project")
```
ajperel: I think we're trying to avoid strong branding maybe this should be s/Cloud Function kit/function kit/? 

do you agree @inlined ?

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+  .option("--npm_package <package>", "NPM package name or specifier for the function kit")
```
ajperel: for the function kit -> to install as a function kit?

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+    if (!options.config) {
+      throw new FirebaseError("Not in a Firebase project directory (firebase.json not found).");
+    }
+
+    const rawPkgName = options.npm_package;
+    if (!rawPkgName) {
+      throw new FirebaseError("set the --npm_package option to a valid NPM package and try again.");
+    }
+
+    const { packageName, version } = parsePackageSpecifier(rawPkgName);
```
ajperel: Seeing this in context... maybe we should rename it `parseNpmPackageSpecifier` to avoid later renames since this whole thing is designed to support other package types eventually.

But this is very easy to do later so feel free to ignore. I also debated only support --package for now and in that context... this seems also fine :P

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+      logger.warn(
+        clc.yellow(
+          `Warning: Package ${clc.bold(packageName)} is a third-party kit (outside the @firebase-functions-kits scope).`,
+        ),
+      );
+      const hasShrinkwrap = await self.checkPackageHasShrinkwrap(rawPkgName);
+      if (!hasShrinkwrap) {
+        logger.warn(
+          clc.yellow(
+            `Warning: Package ${clc.bold(packageName)} does not have an npm-shrinkwrap.json file. npm-shrinkwrap guarantees that you deploy the same version of dependencies that the publisher tested against. Since this kit does not have an npm-shrinkwrap, it is possible that deploys or updates may introduce bugs or vulnerabilities in newer dependency versions that the publisher did not test agianst.`,
```
ajperel: @inlined - Do you think we need to be more explicit about supply chain attacks or is "vulnerabilities" good enough?

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+          ),
+        );
+      }
+      const confirmInstallation = await confirm({
```
ajperel, line 179: We (Google) _should_ always include the shrink wrap file, but in the case where we messed up and didn't... I think it'd be good to have these as independent checks not checking for shrink wrap only if it's a 3rd party package.

And confirm installation if _either_ one or both are an issue.

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+    });
+    validateKit(kitId);
+
+    const instanceId = await input({
+      message: "What would you like to name your first instance of this kit?",
+      default: kitId,
+      nonInteractive: options.nonInteractive,
+    });
+
+    const sourcePath = path.join("function-kits", kitId);
```
ajperel: might be good to make "function-kits" a constant so we never mess it up?

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    const existingCodebases = new Set<string>();
+    const existingInstanceIds = new Set<string>();
+    for (const c of existingFunctions) {
+      if ("codebase" in c && c.codebase) {
+        existingCodebases.add(c.codebase);
+      }
+      if ("instances" in c && c.instances) {
+        validateKitInstances(c.instances, existingInstanceIds);
```
ajperel: this is relying on the fact that `validateKitInstances` modifies `existingInstanceIds` right, especially with what you have below? 

This feels pretty unintuitive. And all existing functions should be valid so is this just a longabout way of constructing existingInstanceIds?

Also it's odd that this function `instances` takes a record of instance to configdir but doesn't care about configdir at all.  Feels like maybe it should just take the collection of instance ids? 

The existing usage calls this with an empty set and then uses the output of existingInstanceIds....

I feel like these should both be cleaned up to create sets of instance ids upfront and have validateKitInstances just do the validation and throw if there is an error and not mutate inputs.

then below use just need something like

`validateKitInstances([instanceId], existingInstanceIds)`

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+      try {
+        existingFunctions = normalizeAndValidate(configFunctions);
+      } catch (err: unknown) {
+        throw new FirebaseError(`Invalid existing functions configuration: ${getErrMsg(err)}`);
+      }
+    }
+
+    for (const c of existingFunctions) {
+      if (isKitConfig(c) && c.kit === kitId) {
+        throw new FirebaseError(
```
ajperel: What happens when these errors are thrown? Does it just cause the command to exit with a failure?

I _hope_ users get prompted to choose a new ID but if so I don't understand the magic making that happen yet.

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+    const absSourcePath = options.config.path(sourcePath);
+    const absConfigDirPath = options.config.path(configDirPath);
+
+    await fs.ensureDir(absSourcePath);
+    await fs.ensureDir(absConfigDirPath);
+
+    const relPackageJsonPath = path.join(sourcePath, "package.json");
+    const absPackageJsonPath = options.config.path(relPackageJsonPath);
+    let pkgJson: {
+      name?: string;
```
ajperel, line 290: Some of these _are_ required yes? At least name. I assume version?

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+      const latestNodeVersion = supported.latest("nodejs").replace("nodejs", "");
+      const subbedTemplate = PACKAGE_NO_LINTING_TEMPLATE.replace("{{RUNTIME}}", latestNodeVersion);
+      try {
+        pkgJson = JSON.parse(subbedTemplate) as typeof pkgJson;
+      } catch (err: unknown) {
+        throw new FirebaseError("Failed to parse package.nolint.json template: " + getErrMsg(err));
+      }
+    }
+
+    pkgJson.name = `${kitId}-wrapper`;
```
ajperel, line 316: I think it's worth a comment explaining what you're doing and why.

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+    );
+
+    const relIndexTsPath = path.join(sourcePath, "src", "index.ts");
+    const absIndexTsPath = options.config.path(relIndexTsPath);
+    if (!(await fs.pathExists(absIndexTsPath))) {
+      const indexContent = INDEX_KIT_TEMPLATE.replace("{{PACKAGE_NAME}}", packageName);
+      await options.config.askWriteProjectFile(relIndexTsPath, indexContent);
+    }
+
+    const installArgs = isThirdParty ? ["install", "--ignore-scripts"] : ["install"];
```
ajperel, line 337: I wonder if we should just always ignore scripts. Yes, we could technically trust ours more... but when would we use them?

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,291 @@
...
+    try {
+      await wrapSpawn("npm", ["run", "build"], absSourcePath);
+    } catch (err: unknown) {
+      throw new FirebaseError(`TypeScript build failed: ${getErrMsg(err)}`);
+    }
+
+    const newKitConfig: KitFunctionConfig = {
+      kit: kitId,
+      sourcePackage: {
+        id: packageName,
```
ajperel, line 355: This is 100% what we put in the design doc... but as I brush up while reviewing...
none of the package systems talk about ids. They talk about package / distribution names. I wonder if we should also name this `name`. But.... maybe not a big deal? If we did want to do it maybe easier in a separate follow up change? Thoughts.

package systems also talk about the full spec (including version, etc.) but we're not storing that and I don't think we should since it'll get out of date.

templates/init/functions/typescript/index-kit.ts
```diff
@@ -0,0 +1,101 @@
+import { setGlobalOptions } from "firebase-functions";
+import {
+//   defineBoolean,
```
ajperel: I have feedback on Victor's PR that maybe changes how we do this completely in which case this comment is not needed.

This is fine but I also was wondering if just to save people effort uncommenting if we should do something like 

import * as params from "firebase-functions/params"

and then like params.defineString()

Only to save us this long list of maybe not used things?

templates/init/functions/typescript/index-kit.ts
```diff
@@ -0,0 +1,101 @@
...
+//   defineList,
+//   defineSecret,
+  defineString,
+} from "firebase-functions/params";
+
+export const regionParam = defineString("FUNCTION_KIT_REGION", {
+  description: "Region where functions should be deployed.",
+});
+
+// To configure more https://firebase.google.com/docs/reference/functions/2nd-gen/node/firebase-functions.globaloptions
```
ajperel: I think we could make this a more user friendly comment if we keep it guide users a bit more.

something like

To require setting a global option for each instance of this kit uncomment the option parameter definition and the line configuring it below. Learn more about these options at: https://firebase.google.com/docs/reference/functions/2nd-gen/node/firebase-functions.globaloptions

or if we go the process.env route a different but equally helpful comment.

ajperel (commented):
Thanks for getting this out!

A bunch of minor comments and questions and a few things we should iron out through discussion.

src/commands/functions-kits-install.ts
wandamora: I don't feel that strongly either way TBH. I preferred `package` and `package_manager` to pave the way to support other package types in the future, but I also see the argument that we're implying the utility is larger than it is.

src/commands/functions-kits-install.ts
wandamora: That's a fair nit, let's keep it ID here and change it in a follow-up.

src/commands/functions-kits-install.ts
wandamora: The command exits, I didn't include the re-prompting in this PR (mainly because I forgot), but I added the reprompting now.

src/commands/functions-kits-install.ts
wandamora: For private wrapper packages, npm only requires name (which we explicitly assign as ${kitId}-wrapper), while version is optional and omitted in private scaffolding. The interface fields are marked optionalvbecause pkgJson is initialized as {} before loading from disk or template, and reading from disk cannot guarantee every field is present.

src/commands/functions-kits-install.spec.ts
wandamora: Added more tests.

src/commands/functions-kits-install.spec.ts
wandamora: Done.

src/commands/functions-kits-install.ts
wandamora: I added a `validateNpmPackageName` method to verify the package name is valid to begin with, and call it before this method.

src/commands/functions-kits-install.ts
wandamora: Done.

src/commands/functions-kits-install.ts
wandamora: Done.

src/commands/functions-kits-install.spec.ts
wandamora: Done.

src/commands/functions-kits-install.ts
wandamora: Good catch, removed.

src/commands/functions-kits-install.ts
wandamora: Update this check to just check if it begins with `@firebase-functions-kits/`.

src/commands/functions-kits-install.ts
wandamora: Updated.

src/commands/functions-kits-install.ts
wandamora: This was an easy change, so I applied it.

src/commands/functions-kits-install.ts
wandamora: Done.

templates/init/functions/typescript/index-kit.ts
wandamora: Updated.

templates/init/functions/typescript/index-kit.ts
wandamora: Updated.

src/commands/functions-kits-install.ts
ajperel: I meant more that we know that's true of valid package names, but I'm pretty sure it is.

src/commands/functions-kits-install.spec.ts
```diff
@@ -0,0 +1,640 @@
...
+
+  describe("isThirdPartyPackage", () => {
+    it("should return false for packages under @firebase-functions-kits scope", () => {
+      expect(isThirdPartyPackage("@firebase-functions-kits/firestore-bigquery-export")).to.be.false;
+    });
+
+    it("should return true for packages outside @firebase-functions-kits scope", () => {
+      expect(isThirdPartyPackage("firebase-functions-kits")).to.be.true;
+      expect(isThirdPartyPackage("@firebase-function-kits-fake/foo")).to.be.true;
+      expect(isThirdPartyPackage("@firebase-functions-kits-fake/foo")).to.be.true;
```
ajperel, line 200: duplicate line

src/commands/functions-kits-install.spec.ts
```diff
@@ -0,0 +1,640 @@
...
+
+      expect(confirmStub).to.have.been.calledOnceWith({
+        message:
+          "Are you sure you want to install @firebase-functions-kits/my-kit without locked dependencies?",
+        default: false,
+        nonInteractive: true,
+      });
+    });
+
+    it("should cancel installation if user declines confirmation for missing shrinkwrap", async () => {
```
ajperel: should we test decline of 3P code too?

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,361 @@
...
+
+/**
+ * Sanitizes an npm package name into a valid kit identifier.
+ * e.g., "@firebase-functions-kits/firestore-bigquery-export" -> "firestore-bigquery-export"
+ */
+export function sanitizePackageNameToKitName(packageName: string): string {
+  const parts = packageName.split("/");
+  const nameWithoutScope = parts[parts.length - 1] || packageName;
+  const sanitized = nameWithoutScope.toLowerCase().replace(/[^a-z0-9_-]/g, "");
+  return (sanitized || "kit").slice(0, 40);
```
ajperel: I can't imagine the `|| "kit"` case will ever happen in practice but I guess it doesn't hurt?

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,361 @@
...
+    if (!hasShrinkwrap) {
+      logger.warn(
+        clc.yellow(
+          `Warning: Package ${clc.bold(packageName)} does not have an npm-shrinkwrap.json file. npm-shrinkwrap guarantees that you deploy the same version of dependencies that the publisher tested against. Since this kit does not have an npm-shrinkwrap, it is possible that deploys or updates may introduce bugs or vulnerabilities in newer dependency versions that the publisher did not test against.`,
+        ),
+      );
+    }
+
+    if (isThirdParty || !hasShrinkwrap) {
+      const confirmMessage = isThirdParty
```
ajperel: We should handle all 3 cases with different messages this handles no shrink wrap and no third-party (with an awkward check) but if they are _both_ true we should  be like 

are you sure you want to install the third-party kit FOO without locked dependencies?

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,361 @@
...
+      throw new FirebaseError(
+        `functions.kit must be unique but '${kitId}' was used more than once.`,
+      );
+    }
+
+    const instanceCollisions = new Set([...existingInstanceIds, ...existingCodebases]);
+    const defaultInstanceId = generateUniqueId(kitId, instanceCollisions);
+
+    const instanceId = await input({
+      message: "What would you like to name your first instance of this kit?",
```
ajperel: This is fine I guess for this PR but you'll soon have to change it as you can re-run to add instances.

Maybe just make it generic now "What would you like to name this instance of this kit?" Or maybe even "this instance of <kit name>"

src/commands/functions-kits-install.ts
```diff
@@ -0,0 +1,361 @@
...
+      default: defaultInstanceId,
+      nonInteractive: options.nonInteractive,
+      validate: (val: string) => {
```
ajperel, line 250: for my own learning is this some magic that re-runs till it validates?

src/functions/projectConfig.ts
```diff
@@ -275,7 +278,10 @@ function assertUniqueKitInstancesAndCodebases(config: ValidatedConfig): void {
       codebases.add(c.codebase);
     }
     if ("instances" in c && c.instances) {
-      validateKitInstances(c.instances, instanceIds);
+      validateKitInstances(Object.keys(c.instances), instanceIds);
+      for (const id of Object.keys(c.instances)) {
```
ajperel: Ugh... I see why you put this _in_ validateKitInstances.

I could be convinced for validateKitInstances to also update `instanceIds` as long as you properly document and name it. Something like `validateAndAddKitInstances` or even something where the joint is a return value

`instanceIds = validateAndAddKitInstances(..., instanceIds)`

Maybe in that case it's a new Set and not a mutation that's also returned but this is in the weeds, fine with any of it.

templates/init/functions/typescript/index-kit.ts
```diff
@@ -0,0 +1,94 @@
+import { setGlobalOptions } from "firebase-functions";
+import * as params from "firebase-functions/params";
+
+export const regionParam = params.defineString("FUNCTION_KIT_REGION", {
```
ajperel: I think we agreed today on `FUNCTION_DEFAULT_REGION`

ajperel (approved):
Thanks for the fast iteration.

src/commands/functions-kits-install.ts
wandamora: Yes, exactly :)

templates/init/functions/typescript/index-kit.ts
wandamora: Updated the whole template based on our chat thread with Thomas.

src/commands/functions-kits-install.ts
wandamora: Changed to "What would you like to name this instance?"

src/commands/functions-kits-install.ts
wandamora: Done.

src/commands/functions-kits-install.ts
wandamora: Eh, I removed it. Some jetski-ness that I don't think is necessary at a second glance.

src/commands/functions-kits-install.spec.ts
wandamora: Done.

src/commands/functions-kits-install.spec.ts
wandamora: Removed.

src/functions/projectConfig.ts
wandamora: This called for better refactoring. I split validateKitInstances into two helper functions: validateAndAddKitInstances and validateKitInstanceID, since we just need the latter for the function-kits-install script.

src/commands/functions-kits-install.ts
wandamora: I'm not sure, but we can leave it as is. We can always change it if we find a reason to.

PR #10906, opened 2026-08-08
feat: inject FIREBASE_KIT_INSTANCE_ID environment variable for function kits

ajperel:
### Description
This change ensures that when deploying or emulating functions that belong to a Kit codebase, an environment variable `FIREBASE_KIT_INSTANCE_ID` is automatically injected into the process containing exactly the kit instance string identifier.

This mirrors how variables like `FIREBASE_CONFIG` and `GCLOUD_PROJECT` are orchestrated through `loadFirebaseEnvs()` and prevents the developer from manually mapping these contexts. The variable remains omitted if the user is deploying a standard codebase.

### Scenarios Tested
- Deployed multiple kits which contain functions that log and return their instance ids showing correct behavior.
- Deployed a non-kit codebase with the kits experiment on and off
- Run all tests

### Sample Commands
- `npm run test`
- `firebase deploy --only functions`

### DIFF
--- a/src/deploy/functions/prepare.spec.ts
+++ b/src/deploy/functions/prepare.spec.ts
@@ -149,6 +149,54 @@ describe("prepare", () => {
       }
     });
 
+    it("should provide FIREBASE_KIT_INSTANCE_ID to discovery envs only for kit instances", async () => {
+      experiments.setEnabled("kits", true);
+      try {
+        const config: ValidatedConfig = [
+          {
+            kit: "my-kit",
+            sourcePackage: { id: "@firebase-functions-kits/my-kit" },
+            source: "source",
+            instances: {
+              "inst-alpha": "config/inst-alpha",
+              "inst-beta": "config/inst-beta",
+            },
+            runtime: "nodejs22",
+          },
+          { source: "source-default", codebase: "default", runtime: "nodejs22" },
+        ];
+        const options = {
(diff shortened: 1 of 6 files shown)


src/emulator/controller.ts
wandamora, line 1: You may also want to edit https://github.com/firebase/firebase-tools/blob/a943b9553ed20e8a3390b302e0439d69ebc440d9/src/serve/functions.ts#L38-L49 to include the `FIREBASE_KIT_INSTANCE_ID` variable as well. It is also related to the emulator.

src/deploy/functions/prepare.spec.ts
wandamora, line 1: nit: add another test case for kits with multiple instances, and verify each discovery has the correct FIREBASE_KIT_INSTANCE_ID populated.

wandamora (approved):
LGTM % small changes.

inlined (approved):
LGTM after Wanda's great catch.

src/emulator/controller.ts
ajperel: Done. Thank you for catching this!

src/deploy/functions/prepare.spec.ts
ajperel: Just made it one expanded test case but done. Also a good suggestion.

Issue #5888, opened 2023-05-24
Failed to load function definition from source: FirebaseError: User code failed to load. Cannot determine  backend specification


sunwupark:
### [REQUIRED] Environment info
**firebase-tools:**
firebase-tools: 12.1.0
"firebase-admin": "^11.8.0",
"firebase-functions": "^4.3.1"
"firebase-functions-test": "^3.1.0"

![image](https://github.com/firebase/firebase-tools/assets/52268188/1dcde0e5-a667-4362-8dd8-a77f147306be)

![image](https://github.com/firebase/firebase-tools/assets/52268188/2e93df81-59ce-4800-82c5-8105cab8daf0)

**Platform: Windows wsl (Ubuntu)
Node: v18.16.0
Python: v3.10.6
openjdk: v"11.0.19"

### [REQUIRED] Test case

i have followed the exact steps in
https://firebase.google.com/docs/functions/get-started?gen=2nd&hl=ko#about_this_tutorial

### [REQUIRED] Expected behavior
No error should have occured 

### [REQUIRED] Actual behavior

**When I run firebase emulators:start**

i  emulators: Starting emulators: functions, firestore
⚠  functions: The following emulators are not running, calls to these services from the Functions emulator will affect production: auth, database, hosting, pubsub, storage
i  firestore: Firestore Emulator logging to firestore-debug.log
✔  firestore: Firestore Emulator UI websocket is running on 
9150.
i  ui: Emulator UI logging to ui-debug.log
i  functions: Watching "/mnt/c/Users/sunwu/Downloads/Tryagain/pythontry/functions" for Cloud Functions...
✔  functions: Using node@18 from host.
⬢  functions: Failed to load function definition from source: FirebaseError: User code failed to load. Cannot determine 
backend specification

┌─────────────────────────────────────────────────────────────┐
│ ✔  All emulators ready! It is now safe to connect your app. │
│ i  View Emulator UI at http://127.0.0.1:4000/
  │
└─────────────────────────────────────────────────────────────┘

┌───────────┬────────────────┬─────────────────────────────────┐
│ Emulator  │ Host:Port      │ View in Emulator UI
   │
├───────────┼────────────────┼─────────────────────────────────┤
│ Functions │ 127.0.0.1:5001 │ http://127.0.0.1:4000/functions │
├───────────┼────────────────┼─────────────────────────────────┤
│ Firestore │ 127.0.0.1:8080 │ http://127.0.0.1:4000/firestore │
[...]

PR #10911, opened 2026-08-10
fix(functions): report discovery crashes instead of blaming the timeout

### LINKED ISSUE #7775: Issue with Global `googleapis` Import Causing Firebase Functions to Fail Locally with "Cannot Determine Backend Specification" Error
#### Problem:
I’m encountering an issue when integrating the `googleapis` library in my Firebase Cloud Function, specifically with the Google Drive API. After adding the global `googleapis` import, both the **local environment** and **deployment** fail.

When I try to run the Firebase Emulator locally after adding the following line:
```javascript
const { google } = require('googleapis');
```

I receive the following error message:

```
⚠  functions: Please note that there will be breaking changes when you upgrade.
✔  functions: Using node@18 from host.
i  functions: Loaded environment variables from .env.
Serving at port 8739

shutdown requested via /__/quitquitquit

⬢  functions: Failed to load function definition from source: FirebaseError: User code failed to load. Cannot determine backend specification

┌─────────────────────────────────────────────────────────────┐
│ ✔  All emulators ready! It is now safe to connect your app. │
│ i  View Emulator UI at http://127.0.0.1:4000/               │
└─────────────────────────────────────────────────────────────┘

```
Steps to Reproduce:
1. Set up a Firebase Cloud Function.
2. Add the following code to globally import and interact with the Google Drive API:

```javascript
const { google } = require('googleapis');  // Global import
const path = require('path');

exports.myFunction = functions.https.onRequest(async (req, res) => {
  try {
    // Authenticate using GoogleAuth
    const auth = new google.auth.GoogleAuth({
      keyFile: path.join(__dirname, 'credentials.json'), // Ensure the correct path
      scopes: ['https://www.googleapis.com/auth/drive'],
    });

    // Initialize Google Drive API
    const drive = google.drive({ version: 'v3', auth });

    // Example: Listing files in Google Drive
    const response = await drive.files.list({
      pageSize: 10,
      fields: 'files(id, name)',
    });

    // Send the result back to the client
    res.status(200).send(response.data.files);
  } catch (error) {
    console.error('Error:', error);
    res.status(500).send('Error in Google Drive API');
  }
});
```

3. Run the Firebase Emulator locally:

```
```

4. After adding the global import const { google } = require('googleapis');, the deployment or local environment fails with the error:

```
Failed to load function definition from source: FirebaseError: User code failed to load. Cannot determine backend specification
```

### Environment

Node.js version: 18
Firebase Functions SDK version: 4.3.1
googleapis version: 144.0.0
Operating System: WSL Ubuntu

### Attempted Solutions

- Removing the global googleapis import resolves the issue, but it’s needed for integrating with the Google Drive API.

IzaakGough:
Fixes #7775 (and #5888)

## What was broken

`User code failed to load. Cannot determine backend specification` blamed a timeout for everything. A discovery server that crashed while loading kept returning `ECONNREFUSED` until the budget ran out, so the real error never surfaced, and one that failed to spawn looked identical. Separately, `FUNCTIONS_DISCOVERY_TIMEOUT` is documented in seconds but the error printed milliseconds and named neither.

## What changed

`discovery.watchDiscoveryProcess` rejects on `close`, so a crash reports its exit code and the tail of its stderr, and a failed spawn reports its Error rather than the negative errno it closes with. Shutdown has no reject path, so it cannot replace the error discovery failed with. `FUNCTIONS_DISCOVERY_TIMEOUT` accepts `s` and `ms` suffixes and is parsed once rather than per emulator worker. Polls back off 100ms.

## Verification

Unit tests for the watcher, `serveAdmin`, and the env var, with both new behaviours confirmed by mutation. Event ordering was measured against `cross-spawn` rather than assumed: ENOENT and EACCES give `error` then `close`, never `exit`. Slow-loading and crashing codebases were run against a real emulator before the switch to `close`; since then, unit tests only.

## For the reviewer

The fix is Node only. The reasoning for leaving Python to #10893 is in the changelog thread.


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
 - Added `appcheck:providers:list`, `appcheck:providers:get` and `appcheck:providers:set` to configure App Check attestation providers for an app.
 - Added `appcheck:apps:list` to show every app with its configured App Check providers.
 - Added web app support for Crashlytics MCP tools and prompts.>>>>>>> main
+- Fixed function discovery reporting a timeout when the discovery server had actually crashed, hiding the underlying error (#7775).
```
cabljac: Worth scoping the entry to Node: python still calls `detectFromPort` without `serverExited`, so a crashing Python discovery server is still reported as a timeout. Either extend the fix or say "for Node.js functions" here.

src/deploy/functions/runtimes/node/index.ts
```diff
@@ -255,28 +267,75 @@ export class Delegate {
...
+    let stderr = "";
+    childProcess.stderr?.on("data", (chunk: Buffer) => {
+      stderr += chunk.toString();
+    });
+
+    // Armed only while discovery is in flight. Once we have asked the server to
+    // quit, an exit is expected rather than a crash.
+    let armed = true;
+    const serverExited = new Promise<never>((_resolve, reject) => {
+      childProcess.once("exit", (code, signal) => {
```
cabljac: Listen on `close` instead of `exit`: it fires for both real exits and spawn failures (ENOENT emits `error`+`close`, never `exit`), and per the Node docs it's the event that guarantees the stderr pipe has finished. For the spawn-failure case build the message from the `error` event's Error, and keep it wrapped in FirebaseError - a raw Error reaching the CLI is reported as "An unexpected error has occurred." (errorOut.ts:13).

src/deploy/functions/runtimes/node/index.ts
```diff
@@ -255,28 +267,75 @@ export class Delegate {
...
-    return Promise.resolve(async () => {
-      const p = new Promise<void>((resolve, reject) => {
-        childProcess.once("exit", resolve);
-        childProcess.once("error", reject);
+    // Attached at spawn time rather than in kill(): neither event replays, so a
+    // server that died before shutdown ran would leave a listener that can never
+    // fire and a kill() that never resolves.
+    let exited = false;
+    const exit = new Promise<void>((resolve, reject) => {
+      childProcess.once("exit", () => {
```
cabljac: Attaching listeners at spawn time is right (the old code could hang forever here). But rejecting on `error` makes `kill()` throw inside `discoverBuild`'s finally, replacing whatever discovery threw. Resolve on `error` like #10893 does for python, and let `serverExited` be the one channel that reports what went wrong. Rule of thumb: cleanup should never out-rank the error it's cleaning up after.

src/deploy/functions/runtimes/node/index.ts
```diff
@@ -255,28 +267,75 @@ export class Delegate {
...
+        resolve();
       });
+      childProcess.once("error", reject);
+    });
+    exit.catch(() => {
+      // kill() is the only intended consumer; it may never be called.
+    });
 
+    let stderr = "";
+    childProcess.stderr?.on("data", (chunk: Buffer) => {
```
cabljac: +1 to gemini's size cap. Also `spawnFunctionsProcess` already logs every stderr chunk at line 243, so this re-prints output the user just saw - keeping just the tail (~8KB) covers both.

src/deploy/functions/runtimes/discovery/index.ts
```diff
@@ -12,8 +12,66 @@ import { FirebaseError } from "../../../../error";
...
+    return 0;
+  }
+
+  const value = Number(match[1]);
+  const unit = (match[2] || "s").toLowerCase();
+  if (unit === "ms") {
+    return value;
+  }
+
+  if (value >= SUSPICIOUS_SECONDS) {
```
cabljac, line 70: This fires per emulator worker (functionsRuntimeWorker.ts:269 calls it per function), so a user with a big value set sees the warning repeatedly. Memoize the parse: compute it once on first call, store the result in a module-level variable, and return that on every later call. The env var can't change mid-process, so parsing (and warning) once is enough.

src/deploy/functions/runtimes/discovery/index.ts
```diff
@@ -12,8 +12,66 @@ import { FirebaseError } from "../../../../error";
...
+ * off-by-1000 mistakes.
+ */
 export function getFunctionDiscoveryTimeout(): number {
-  return +(process.env[TIMEOUT_OVERRIDE_ENV_VAR] || 0) * 1000; /* ms */
+  const raw = process.env[TIMEOUT_OVERRIDE_ENV_VAR]?.trim();
+  if (!raw) {
+    return 0;
+  }
+
+  const match = /^(\d+(?:\.\d+)?)\s*(ms|s)?$/i.exec(raw);
```
cabljac, line 56: Negative values used to fire the timeout instantly (`setTimeout` with a negative delay); the fallback is a real improvement - please add a test so it doesn't regress. The rejected exotic forms (`1e3`, `.5`, hex) are fine to drop but belong in the changelog entry.

src/deploy/functions/runtimes/discovery/index.spec.ts
```diff
@@ -120,4 +121,88 @@ describe("detectFromPort", () => {
     const parsed = await discovery.detectFromPort(8080, "project", "nodejs16", 0, 500);
     expect(parsed).to.deep.equal(BUILD);
   });
+
+  it("explains how to extend the timeout when it expires", async () => {
```
cabljac, line 145: These tests inherit ambient env: `FUNCTIONS_DISCOVERY_TIMEOUT=3 npx mocha src/deploy/functions/runtimes/discovery/index.spec.ts` fails the timeout test. The `getFunctionDiscoveryTimeout` block below already does save/delete/restore - hoist it to file level.

cabljac (changes requested):
Thanks @IzaakGough, this looks like the right general design (racing the child's exit against the poll). Requesting changes for some things:

1. CHANGELOG.md line 3 has a committed conflict marker (`>>>>>>> main`). Rebase onto main and resolve.
2. gemini's high-priority comment is right: a failed spawn emits `error` then `close`, never `exit` (I confirmed with a spawn script), so discovery still rides out the full timeout. Rather than adding an `error` listener, suggest listening on `close` - details inline.
3. `kill()` can reject over the real error - the user then sees "An unexpected error has occurred." Details inline.
4. The new lifecycle code in node/index.ts has no tests. #10893's python spec shows the pattern (stub spawn, drive an EventEmitter as the child); please mirror it here.

Smaller items inline. One general note: worth a final `git diff main` read before pushing - a committed conflict marker means the merge was committed unresolved.

src/deploy/functions/runtimes/node/index.ts
```diff
@@ -255,28 +267,75 @@ export class Delegate {
...
+        const details = stderr.trim();
+        reject(
+          new FirebaseError(
+            `User code failed to load. Cannot determine backend specification. ` +
+              `The functions process exited with ${how} before it could be analyzed.` +
+              (details ? `\n\n${details}` : ""),
+          ),
+        );
+      });
+    });
```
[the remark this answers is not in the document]
IzaakGough: Fixed in 0e481b9d2, though by listening on `close` rather than adding an `error` listener. A failed spawn emits `error` and then `close`, and never `exit`. Probing `cross-spawn` directly gives `error(ENOENT) -> close(code=-2)` for a missing binary and `error(EACCES) -> close(code=-13)` for a non-executable one, against `exit(code=7) -> close(code=7)` for a normal crash. `close` also guarantees stderr has finished arriving, so one listener covers both cases.

The `error` event is still handled, but only to hold onto the Error: `close` reports the errno as a negative exit code, so without it the message would say "exited with code -2" instead of naming the spawn failure.

src/deploy/functions/runtimes/discovery/index.ts
```diff
@@ -12,8 +12,66 @@ import { FirebaseError } from "../../../../error";
...
+// CPU with the very process we are waiting on, which on constrained machines is
+// enough to cause the timeout it is meant to detect.
+const RETRY_DELAY_MS = 100;
+
+// A bare value at or above this is almost certainly milliseconds. The variable is
+// documented in seconds, so 30000 means 8.3 hours, which silently disables the
+// timeout rather than extending it.
+const SUSPICIOUS_SECONDS = 600;
+
+const sleep = (ms: number): Promise<void> => new Promise((resolve) => setTimeout(resolve, ms));
```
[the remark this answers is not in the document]
IzaakGough: Done in 0e481b9d2, now importing `sleep` from `src/utils`.

src/deploy/functions/runtimes/node/index.ts
```diff
@@ -255,28 +267,75 @@ export class Delegate {
...
+      childProcess.once("error", reject);
+    });
+    exit.catch(() => {
+      // kill() is the only intended consumer; it may never be called.
+    });
 
+    let stderr = "";
+    childProcess.stderr?.on("data", (chunk: Buffer) => {
+      stderr += chunk.toString();
+    });
```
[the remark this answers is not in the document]
IzaakGough: Done in 0e481b9d2, capped at 8KB. Kept the tail rather than the head, since `spawnFunctionsProcess` already streams every chunk to the user and the end of the output is the part that identifies the failure.

CHANGELOG.md
IzaakGough: Scoped the entry to Node.js in 0e481b9d2.

Python needs the same wiring, but #10893 rewrites `serveAdmin` and `shutdownAdmin` end to end, so doing it here would leave both PRs editing the same function. The lifecycle handling now lives in `discovery.watchDiscoveryProcess`, shared rather than Node-specific, so Python becomes a few lines on top of #10893 once one of them lands.

src/deploy/functions/runtimes/node/index.ts
IzaakGough: Done in 0e481b9d2. The lifecycle moved into `discovery.watchDiscoveryProcess`, which watches `close`, wraps the failure in a `FirebaseError`, and builds the message from the `error` event's Error when there was one.

Confirmed the event order with `cross-spawn` before changing it: ENOENT gives `error(ENOENT) -> close(code=-2)`, EACCES gives `error(EACCES) -> close(code=-13)`, and a normal crash gives `exit(code=7) -> close(code=7)`.

src/deploy/functions/runtimes/node/index.ts
IzaakGough: Done in 0e481b9d2. `watched.ended` resolves on whichever of `exit` or `close` comes first and has no reject path at all, so `kill()` cannot throw over the real error. The `error` event now only records the Error for `serverExited` to report.

Covered by "shuts down without throwing after a spawn failure" in `node/index.spec.ts`. It fails if `ended` is given a reject path again.

On the `exit`-or-`close` choice for `ended`: `close` can lag `exit` indefinitely when a surviving grandchild still holds the stdio pipes, and shutdown should not wait on that. `serverExited` still uses `close` alone, where the stderr guarantee matters.

src/deploy/functions/runtimes/node/index.ts
IzaakGough: Done in 0e481b9d2, kept as the last 8KB for exactly that reason.

src/deploy/functions/runtimes/discovery/index.ts
IzaakGough: Done in 0e481b9d2. The parse is cached, so the warning is emitted once no matter how many workers ask. Keyed by the raw env value rather than a single module variable, which costs nothing and keeps the tests able to vary it.

src/deploy/functions/runtimes/discovery/index.ts
IzaakGough: Both done in 0e481b9d2. Added "ignores a negative value rather than timing out instantly", and the changelog entry now names the rejected forms:

> Values it cannot parse, including negative numbers and forms such as `1e3`, `.5` and `0x10`, are ignored with a warning so the default timeout applies.

src/deploy/functions/runtimes/discovery/index.spec.ts
IzaakGough: Done in 0e481b9d2, hoisted to file level so the whole spec runs against a cleared variable. `FUNCTIONS_DISCOVERY_TIMEOUT=3 npx mocha src/deploy/functions/runtimes/discovery/index.spec.ts` now passes.

IzaakGough:
Thanks for the review. All of it is addressed in 0e481b9d2, with the details in the threads. Covering the two numbered points that were not inline:

**1, the conflict marker.** Merged `origin/main` in and resolved `CHANGELOG.md` properly. The branch was 19 commits behind, and the two stale entries the bad merge had resurrected (the App Check and Crashlytics ones) shipped in 15.27.0, so they are gone again.

**4, tests for the lifecycle code.** The lifecycle moved into `discovery.watchDiscoveryProcess`, shared rather than Node-specific, so it is tested directly using the EventEmitter-as-child pattern: crash exit code with stderr, kill signal, spawn failure, the 8KB stderr cap, disarm, and the healthy case where nothing settles. `node/index.spec.ts` then covers `serveAdmin` wiring it up, including that `kill()` resolves after a spawn failure.

Both new behaviours were checked by mutation: reverting `close` back to `exit` fails four of them, and giving `ended` a reject path again fails the `kill()` one.

PR #10913, opened 2026-08-10
feat(extensions): add extensions replacement registry and automated g…

shettyvarun268:
Because Firebase Extensions is shutting down on March 31, 2027, developers need migration paths to official replacement packages. This PR builds the foundation to track those replacements by adding a central file (src/extensions/replacements.json) containing 113 verified extensions and their current replacement statuses. It also provides a helper (src/extensions/replacementRegistry.ts) so the CLI and Console can easily check whether an extension has a replacement and format a clear warning message.

To keep this list up to date without manual editing, this PR also introduces an automated scraper tool. When executed, the script reaches out to the GitHub repositories of all 68 extensions, reads their README files to check if the publisher has posted a replacement tag, and automatically updates replacements.json whenever new replacement packages are announced.

Runtime CLI warning banners  will be hooked up in a follow-up PR.

Runbook:
Run Unit Tests:

```bash
npx mocha -r ts-node/register src/extensions/replacementRegistry.spec.ts scripts/extensions-scraper/index.spec.ts
```
Expect: 11 tests passing in ~20ms.

Run Live GitHub Scraper:

```bash
npm run scrape:extensions
```
Expect: Scans all 113 repositories, detects merged README tags, and writes updates directly into src/extensions/replacements.json.

Check In Updates: Whenever partner publishers merge new replacement notice tags, running the scraper command above will update replacements.json, which can then be committed and pushed.

### DIFF
--- a/scripts/extensions-scraper/index.spec.ts
+++ b/scripts/extensions-scraper/index.spec.ts
@@ -1,0 +1,84 @@
+import { expect } from "chai";
+import {
+  extractReplacementFromReadme,
+  getRepoUrlForExtension,
+  processExtensionReadmes,
+  toRawGithubUrl,
+} from "./index";
+
+describe("extensions-scraper", () => {
+  describe("extractReplacementFromReadme", () => {
+    it("should extract package name from 1P extension tag", () => {
+      const readme = `
+<!-- FIREBASE_EXTENSION_REPLACEMENT: extension="firebase/firestore-send-email" package="@firebase/firestore-send-email" -->
+> Deprecation notice: Migrate to package.
+      `;
+      const pkg = extractReplacementFromReadme(readme);
+      expect(pkg).to.equal("@firebase/firestore-send-email");
+    });
+
+    it("should extract package name from 2P partner tag with single quotes", () => {
+      const readme = `
+<!-- FIREBASE_EXTENSION_REPLACEMENT: extension='stripe/firestore-stripe-payments' package='@stripe/firestore-stripe-payments' -->
+      `;
+      const pkg = extractReplacementFromReadme(readme);
+      expect(pkg).to.equal("@stripe/firestore-stripe-payments");
+    });
+  });
+
+  describe("toRawGithubUrl", () => {
+    it("should convert github.com tree URLs to raw.githubusercontent.com", () => {
+      const webUrl = "https://github.com/firebase/extensions/tree/main/firestore-send-email/README.md";
+      expect(toRawGithubUrl(webUrl)).to.equal(
+        "https://raw.githubusercontent.com/firebase/extensions/main/firestore-send-email/README.md",
+      );
+    });
+
+    it("should convert github.com blob URLs to raw.githubusercontent.com", () => {
+      const webUrl = "https://github.com/firebase/firestore-bundle-builder/blob/master/README.md";
+      expect(toRawGithubUrl(webUrl)).to.equal(
+        "https://raw.githubusercontent.com/firebase/firestore-bundle-builder/master/README.md",
+      );
+    });
+  });
+
+  describe("getRepoUrlForExtension", () => {
+    it("should resolve 1P firebase extensions to github.com repo tree", () => {
+      const url = getRepoUrlForExtension("firebase/firestore-send-email");
+      expect(url).to.equal(
+        "https://github.com/firebase/extensions/tree/main/firestore-send-email/README.md",
+      );
+    });
+
+    it("should resolve 2P partner extensions using explicit extensionRepositoryUrl", () => {
+      const url = getRepoUrlForExtension("stripe/firestore-stripe-payments", {
+        extensionRepositoryUrl:
+          "https://github.com/stripe/stripe-firebase-extensions/tree/main/firestore-stripe-payments/README.md",
+      });
+      expect(url).to.equal(
+        "https://github.com/stripe/stripe-firebase-extensions/tree/main/firestore-stripe-payments/README.md",
+      );
+    });
+  });
+
+  describe("processExtensionReadmes", () => {
+    it("should update registry when valid replacement tags are found", () => {
+      const readmes = {
+        "firebase/firestore-send-email":
+          '<!-- FIREBASE_EXTENSION_REPLACEMENT: package="@firebase/firestore-send-email" -->',
+      };
+      const initialRegistry = {
+        replacements: {
+          "firebase/firestore-send-email": { status: "PENDING_PUBLISHER" },
+        },
+      };
+
+      const { updatedRegistry, results } = processExtensionReadmes(readmes, initialRegistry);
+      expect(results[0].status).to.equal("REPLACEMENT_AVAILABLE");
+      expect(updatedRegistry.replacements["firebase/firestore-send-email"]).to.deep.equal({
+        status: "REPLACEMENT_AVAILABLE",
+        npmPackage: "@firebase/firestore-send-email",
+      });
+    });
+  });
+});
--- a/scripts/extensions-scraper/index.spec.ts
+++ b/scripts/extensions-scraper/index.spec.ts
@@ -1,11 +1,11 @@
 import { expect } from "chai";
+import { ReplacementRegistrySchema } from "../../src/extensions/replacementRegistry";
 import {
   extractReplacementFromReadme,
   getRepoUrlForExtension,
   processExtensionReadmes,
   toRawGithubUrl,
 } from "./index";
-import { ReplacementRegistrySchema } from "../../src/extensions/replacementRegistry";
 
 describe("extensions-scraper", () => {
   describe("extractReplacementFromReadme", () => {
@@ -25,6 +25,32 @@ describe("extensions-scraper", () => {
       const pkg = extractReplacementFromReadme(readme);
       expect(pkg).to.equal("@stripe/firestore-stripe-payments");
     });
+
+    it("should return undefined when replacement tag is missing the package attribute", () => {
+      const readme = `
+<!-- FIREBASE_EXTENSION_REPLACEMENT: extension="firebase/firestore-send-email" target="npm" -->
+## Migration Guide
+Check out the replacement guide below.
+      `;
+      const pkg = extractReplacementFromReadme(readme);
+      expect(pkg).to.be.undefined;
+    });
+
+    it("should return undefined for unrelated HTML comments in README", () => {
+      const readme = `
+<!-- TODO: update configuration instructions before v2 release -->
+<!-- markdownlint-disable MD013 -->
+# Firestore Extension
+Standard documentation and configuration details.
+      `;
+      const pkg = extractReplacementFromReadme(readme);
+      expect(pkg).to.be.undefined;
+    });
+
+    it("should return undefined for empty or whitespace content", () => {
+      expect(extractReplacementFromReadme("")).to.be.undefined;
+      expect(extractReplacementFromReadme("   \n\t  ")).to.be.undefined;
+    });
   });
 
   describe("toRawGithubUrl", () => {
@@ -42,23 +68,29 @@ describe("extensions-scraper", () => {
         "https://raw.githubusercontent.com/firebase/firestore-bundle-builder/master/README.md",
       );
     });
-  });
 
-  describe("getRepoUrlForExtension", () => {
-    it("should resolve 1P firebase extensions to github.com repo tree", () => {
-      const url = getRepoUrlForExtension("firebase/firestore-send-email");
-      expect(url).to.equal(
-        "https://github.com/firebase/extensions/tree/main/firestore-send-email/README.md",
+    it("should throw error for non-GitHub hosts", () => {
+      expect(() => toRawGithubUrl("https://gitlab.com/owner/repo/README.md")).to.throw(
+        "Unsupported host",
       );
     });
 
-    it("should resolve 2P partner extensions using explicit extensionRepositoryUrl", () => {
-      const url = getRepoUrlForExtension("stripe/firestore-stripe-payments", {
+    it("should throw error for malformed URLs", () => {
+      expect(() => toRawGithubUrl("not-a-valid-url")).to.throw(
+        "Failed to convert to raw GitHub URL",
+      );
+    });
+  });
+
+  describe("getRepoUrlForExtension", () => {
+    it("should return the mandatory extensionRepositoryUrl from the entry", () => {
+      const url = getRepoUrlForExtension({
+        status: "PENDING_PUBLISHER",
         extensionRepositoryUrl:
-          "https://github.com/stripe/stripe-firebase-extensions/tree/main/firestore-stripe-payments/README.md",
+          "https://github.com/firebase/extensions/tree/main/firestore-send-email/README.md",
       });
       expect(url).to.equal(
-        "https://github.com/stripe/stripe-firebase-extensions/tree/main/firestore-stripe-payments/README.md",
+        "https://github.com/firebase/extensions/tree/main/firestore-send-email/README.md",
       );
     });
   });
@@ -67,7 +99,7 @@ describe("extensions-scraper", () => {
     it("should update registry when valid replacement tags are found", () => {
       const readmes = {
         "firebase/firestore-send-email":
-          '<!-- FIREBASE_EXTENSION_REPLACEMENT: package="@firebase/firestore-send-email" -->',
+          '<!-- FIREBASE_EXTENSION_REPLACEMENT: package="@firebase-function-kits/firestore-send-email" -->',
       };
       const initialRegistry: ReplacementRegistrySchema = {
         replacements: {
@@ -83,7 +115,7 @@ describe("extensions-scraper", () => {
       expect(results[0].status).to.equal("REPLACEMENT_AVAILABLE");
       expect(updatedRegistry.replacements["firebase/firestore-send-email"]).to.deep.equal({
         status: "REPLACEMENT_AVAILABLE",
-        npmPackage: "@firebase/firestore-send-email",
+        npmPackage: "@firebase-function-kits/firestore-send-email",
         extensionRepositoryUrl:
           "https://github.com/firebase/extensions/tree/main/firestore-send-email/README.md",
       });
--- a/scripts/extensions-scraper/index.spec.ts
+++ b/scripts/extensions-scraper/index.spec.ts
@@ -5,6 +5,7 @@ import {
   processExtensionReadmes,
   toRawGithubUrl,
 } from "./index";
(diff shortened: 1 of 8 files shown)


scripts/extensions-scraper/index.ts
```diff
@@ -0,0 +1,112 @@
...
+  detectedPackage?: string;
+  status: "REPLACEMENT_AVAILABLE" | "PENDING_PUBLISHER";
+}
+
+/**
+ * Extracts replacement npm package name from extension README markdown content.
+ * Checks for machine tag:
+ *   <!-- FIREBASE_EXTENSION_REPLACEMENT: extension="publisher/extension-name" package="@scope/package-name" -->
+ * or single-quoted:
+ *   <!-- FIREBASE_EXTENSION_REPLACEMENT: package='@scope/package-name' -->
```
ajperel, line 18: is the extension optional or did you not mean to omit it?

scripts/extensions-scraper/index.ts
```diff
@@ -0,0 +1,112 @@
...
+  if (publisher === "firebase") {
+    return `https://github.com/firebase/extensions/tree/main/${extensionId}/README.md`;
+  }
+  return `https://github.com/${publisher}/${extensionId}/tree/main/README.md`;
+}
+
+/**
+ * Scans a list of extension README files or content strings and updates the registry object.
+ */
+export function processExtensionReadmes(
```
ajperel, line 86: IS this function ever used? you also have usages `extractReplacementFromReadme` elsewhere that seem to do the actual processing.

src/extensions/replacementRegistry.ts
```diff
@@ -0,0 +1,59 @@
+import * as replacementsData from "./replacements.json";
+
+export type ReplacementStatus =
+  | "REPLACEMENT_AVAILABLE"
+  | "CONFIRMED_NO_REPLACEMENT"
+  | "PENDING_PUBLISHER";
+
+export interface ReplacementInfo {
```
ajperel: You could type this a bit better to make sure we have npm packages when we should doing:

```
export type ReplacementInfo =
  | { status: "REPLACEMENT_AVAILABLE"; npmPackage: string; extensionRepositoryUrl?: string }
  | { status: "CONFIRMED_NO_REPLACEMENT"; npmPackage?: never; extensionRepositoryUrl?: string }
  | { status: "PENDING_PUBLISHER"; npmPackage?: never; extensionRepositoryUrl?: string };
```

And actually shouldn't `extensionRepositoryUrl` be required? We can't really do anything without it right?

scripts/extensions-scraper/run-live-scan.ts
```diff
@@ -0,0 +1,95 @@
...
+
+function fetchUrlContent(url: string, maxRedirects = 5): Promise<string> {
+  return new Promise((resolve) => {
+    if (maxRedirects < 0) {
+      resolve("");
+      return;
+    }
+    try {
+      https
+        .get(url, (res) => {
```
ajperel: Just to be safe this should probably have a timeout like

const req = https.get(url, { timeout: 10000 }, (res)

scripts/extensions-scraper/run-live-scan.ts
```diff
@@ -0,0 +1,95 @@
...
+      console.log(`  Web URL:    ${webUrl} (No tag present)\n`);
+    }
+  }
+
+  if (detectedCount > 0) {
+    fs.writeFileSync(replacementsPath, JSON.stringify(registry, null, 2) + "\n");
+    console.log(`[Scraper] Saved updated registry to ${replacementsPath}\n`);
+  }
+
+  console.log("=======================================================");
```
ajperel, line 127: I didn't see much if any explicit error handling. This script should log if it fails to scan a README for any extension. Then we can know and either (1) fix the issue or at least (2) manually check those repos.

scripts/extensions-scraper/index.ts
shettyvarun268: Yes, extension is optional here. The scraper already knows which extension it's inspecting based on the repository/folder it's scanning, so package is the only strictly required attribute. Having extension there is just optional context for someone reading the markdown. I've updated the docstring in index.ts to make this clearer.

scripts/extensions-scraper/index.ts
shettyvarun268: Good catch! Refactored run-live-scan.tsto pipe the fetched READMEs throughprocessExtensionReadmes directly.

src/extensions/replacementRegistry.ts
shettyvarun268: Updated to a discriminated union and made extensionRepositoryUrl required on all entries.

scripts/extensions-scraper/run-live-scan.ts
shettyvarun268: Added a 10s timeout with cleanup handler on https.get.

scripts/extensions-scraper/run-live-scan.ts
shettyvarun268: Added explicit error tracking and a failed-scan summary at the end of the scan.

scripts/extensions-scraper/index.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  processExtensionReadmes,
+  toRawGithubUrl,
+} from "./index";
+import { ReplacementRegistrySchema } from "../../src/extensions/replacementRegistry";
+
+describe("extensions-scraper", () => {
+  describe("extractReplacementFromReadme", () => {
+    it("should extract package name from 1P extension tag", () => {
+      const readme = `
+<!-- FIREBASE_EXTENSION_REPLACEMENT: extension="firebase/firestore-send-email" package="@firebase/firestore-send-email" -->
```
ajperel, line 14: Not a big deal but you could add some other content to a readme or have some negative cases that _Shouldn't_ work like

Hi I'm a publisher who can't follow instructions!



Please find my kit on npm! You can do it!

But honestly maybe overkill for all this.

scripts/extensions-scraper/index.ts
```diff
@@ -0,0 +1,130 @@
...
+    const parsed = new URL(url);
+    if (parsed.hostname === "raw.githubusercontent.com") {
+      return url;
+    }
+    if (parsed.hostname === "github.com") {
+      const pathname = parsed.pathname.replace(/\/tree\//, "/").replace(/\/blob\//, "/");
+      return `https://raw.githubusercontent.com${pathname}`;
+    }
+  } catch {
+    // Return unchanged if not a valid URL
```
ajperel: I don't expect us to ever hit this since we control all the links but... shouldn't this be one of the error cases you log.

scripts/extensions-scraper/index.ts
```diff
@@ -0,0 +1,130 @@
...
+  return url;
+}
+
+/**
+ * Resolves the repository URL for an extension.
+ * - If entry has explicit "extensionRepositoryUrl", returns that.
+ * - For 1P extensions (firebase/*), defaults to firebase/extensions main branch.
+ * - For 2P partner extensions (publisher/*), constructs standard GitHub URL.
+ */
+export function getRepoUrlForExtension(
```
ajperel: We made this mandatory right? So you shouldn't really need this you can always use entry.extensionRepositoryUrl

Also... I don't think we can guarantee this woudl be right:  

return `https://github.com/${publisher}/${extensionId}/tree/main/README.md`;  

I don't think we enforce anything like that so requiring the repo URL was a good thing. Don't overcomplicate it!

scripts/extensions-scraper/index.ts
ajperel: lol.... you also could have just deleted it if the other approach worked. I didn't have a problem with how it was implemented, just the dead code.

src/extensions/replacements.json
```diff
@@ -0,0 +1,289 @@
...
+      "status": "PENDING_PUBLISHER",
+      "extensionRepositoryUrl": "https://github.com/cloudflare/turnstile-firebase-app-check-provider/tree/main/README.md"
+    },
+    "moralis/moralis-streams": {
+      "status": "PENDING_PUBLISHER",
+      "extensionRepositoryUrl": "https://github.com/MoralisWeb3/youtube-tutorials/tree/main/StreamsFrontend"
+    },
+    "iproov/auth-iproov": {
+      "status": "PENDING_PUBLISHER",
+      "extensionRepositoryUrl": "https://extensions.dev/extensions/iproov/auth-iproov"
```
ajperel: These are obviously wrong. We don't host extensions on extensions.dev. Same below

src/extensions/replacements.json
```diff
@@ -0,0 +1,289 @@
...
+    },
+    "firebase/storage-resize-images": {
+      "status": "REPLACEMENT_AVAILABLE",
+      "npmPackage": "@firebase-function-kits/storage-resize-images",
+      "extensionRepositoryUrl": "https://github.com/firebase/extensions/tree/main/storage-resize-images/README.md"
+    },
+    "firebase/firestore-bigquery-export": {
+      "status": "REPLACEMENT_AVAILABLE",
+      "npmPackage": "@firebase-function-kits/firestore-bigquery-export",
+      "extensionRepositoryUrl": "https://github.com/firebase/extensions/tree/main/firestore-bigquery-export/README.md"
```
ajperel: I notice the scrapper doesn't pick up the replacements for this extension. This is because Invertase made the change in the branch: https://github.com/firebase/extensions/tree/kits/firestore-bigquery-export

BUT ... this code is correct... we want them to make it on the main branch. But... they shouldn't yet. It's a bit of a catch 22. I think it could be OK to scrape the test branches now to validate everything and then scrape the correct main branch at launch time. Though in practice as long as they have the right data we'd survive.

src/extensions/replacementRegistry.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (!extensionRef) {
+    return undefined;
+  }
+  return registry.replacements?.[extensionRef];
+}
+
+/**
+ * Formats a clean deprecation notice string for CLI display.
+ */
+export function getDeprecationWarningMessage(extensionRef: string): string | undefined {
```
ajperel, line 45: This wasn't needed yet, but I guess fine to add now.

src/extensions/replacements.json
ajperel: These URLS are obviously wrong. we don't host extensions source code on extensions.dev nor is this a README.md.

Please don't just trust AI and validate _every_ URL. By actually opening it. It's toil, but we only have to do it once.

ajperel (commented):
Gemini ran your scrapper script for me and says you have some broken URLs to fix:

purchasely/purchasely-in-app-purchases
URL: https://github.com/purchasely/purchasely-firebase-extension/tree/main/README.md
Error: HTTP 404 (The branch name might be different or repository renamed).
moralis/moralis-streams

URL: https://github.com/MoralisWeb3/youtube-tutorials/tree/main/StreamsFrontend
Error: HTTP 404 (Points to a directory. The raw URL converter expects a file. It should likely be updated to point to StreamsFrontend/README.md).

Also it says:

"[Scraper] Scanning 68 extensions from GitHub..." 

But aren't there ~100 extensions? What about the rest?

scripts/extensions-scraper/index.ts
inlined: Agreed. Always handle or log an error. If it should never happen, maybe we should just let the script explode so we don't start missing coverage.

scripts/extensions-scraper/index.ts
inlined: Agree that this should just be accessing what's in the extension spec.

OTOH, I tried following some of those links today from extensions.dev and got a 404 😱 
Google was able to help me find the moved locations. IDK what we should do about this. Maybe run the script once, collect the 404s, Google for their replacement, and add to a hard coded list?

scripts/extensions-scraper/run-live-scan.ts
```diff
@@ -0,0 +1,143 @@
...
+  ReplacementRegistrySchema,
+} from "../../src/extensions/replacementRegistry";
+
+interface FetchResult {
+  content: string;
+  statusCode?: number;
+  error?: string;
+}
+
+function fetchUrlContent(url: string, maxRedirects = 5): Promise<FetchResult> {
```
inlined: Can't this all just be replaced with `fetch(url, { follow: 5 });` ? I don't see what we're benefitting doing an IO pipe technique

src/extensions/replacements.json
inlined: This is actually a really interesting point. I wonder if we need to support branches? But we'd never know because the source links are exact commit hashes, so I guess we have to special case these?

src/extensions/replacements.json
inlined: You can do this by going to the details page on extensions.dev and clicking to view the source. YOU WILL GET 404s and will have to google to find some.

scripts/extensions-scraper/run-live-scan.ts
shettyvarun268: Done, swapped out the stream logic with native fetch()

src/extensions/replacements.json
shettyvarun268: Added a fallback in the script to check the kits branch if main doesn't have the tag yet, so we can test Invertase's branch now before they merge to main. Once the branches have moved into main, we can get rid of this in a followup. We can let this exist now just for testing and our sanity checks

scripts/extensions-scraper/index.ts
shettyvarun268: Done, simplified this to directly use entry.extensionRepositoryUrl, and verified all the target URLs are live.

scripts/extensions-scraper/index.ts
shettyvarun268: Done, made it throw an explicit error on invalid/non-GitHub URLs so it doesn't fail silently.

scripts/extensions-scraper/index.spec.ts
shettyvarun268: Done, added negative test cases for malformed tags, missing attributes, and unrelated comments.

src/extensions/replacementRegistry.ts
shettyvarun268: Noted

src/extensions/replacements.json
shettyvarun268: Noted. Updated the list and verified that all the URLs are correct now.

shettyvarun268: Corrected the list and added all the extensions on marketplace with verified URLs

scripts/extensions-scraper/index.ts
```diff
@@ -0,0 +1,129 @@
...
+        detectedPackage,
+        status: "REPLACEMENT_AVAILABLE",
+      });
+    } else if (
+      existingEntry &&
+      existingEntry.status === "REPLACEMENT_AVAILABLE" &&
+      existingEntry.npmPackage
+    ) {
+      // Preserve existing verified / pre-seeded replacement
+      results.push({
```
ajperel, line 97: Ohh. I think this is good in that if a publisher decides they want to pick a new package name this will pick it up and replace. But it's interesting in that if they _stop_ advertising a replacement we will continue with the existing one.

I honestly don't know what to do in this case. Like... they published a replacement but now don't want it? I think I'm fine with this behavior for now, but a case could be made that we should revert to "publisher pending" if they removed it from their README? But I guess we _need_ this case for now for the hand written confirmed no replacement cases.

scripts/extensions-scraper/run-live-scan.ts
```diff
@@ -0,0 +1,140 @@
...
+        reason: `Invalid URL format: ${String(err)}`,
+      });
+      console.log(`[✗ ERROR] ${extRef}`);
+      console.log(`  Reason:  Invalid URL (${webUrl})\n`);
+      continue;
+    }
+
+    const fetchResult = await fetchUrlContent(rawUrl);
+    let discoveredPackage = extractReplacementFromReadme(fetchResult.text);
+    let sourceBranch = "main";
```
ajperel: Shouldn't this come from the URL. It doesn't have to be main.

scripts/extensions-scraper/run-live-scan.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    // Support pre-merge testing against 'kits' branch if 'main' doesn't have the tag yet
```
ajperel: I don't think we want this. We don't want to report to users about a replacement that's not officially available yet. And officially available should mean in the official README of the extension not some dev branch.

I acknowledge that for the purposes of testing this script we wanted to scrape a version that reports a replacement before we've updated our official readmes. But I think a cleaner way to do that is to temporarily just change the URLs in replacements.json and fix it before we officially launch this.

scripts/extensions-scraper/run-live-scan.ts
```diff
@@ -0,0 +1,140 @@
...
+    if (discoveredPackage) {
+      detectedCount++;
+      registry.replacements[extRef] = {
+        ...registry.replacements[extRef],
+        status: "REPLACEMENT_AVAILABLE",
+        npmPackage: discoveredPackage,
+      };
+      console.log(`[✓ DETECTED] ${extRef}`);
+      console.log(`  Package: ${discoveredPackage}`);
+      console.log(`  Branch:  ${sourceBranch}`);
```
ajperel: Yeah, I'd probably just remove the sourceBranch variable and this log since it should be in the URL right? And otherwise we're parsing it just to be clear here. This seems to be optimizing for the kits vs main branch that I don't think we should be doing in the official script as mentioned above.

shettyvarun268: Done, removed `sourceBranch` and that log line.

scripts/extensions-scraper/run-live-scan.ts
shettyvarun268: Agreed. Removed the branch fallback logic completely. As suggested, I've pointed the 7 Firebase extension URLs to the `kits` branch directly in `replacements.json` for testing, and we can switch them to `main` once merged.

scripts/extensions-scraper/run-live-scan.ts
shettyvarun268: Good catch. Removed `sourceBranch` and relying entirely on the URL specified in `replacements.json`.

scripts/extensions-scraper/index.ts
shettyvarun268: Makes sense. For now this protects any verified / manual entries from being accidentally overwritten if a network fetch fails, but if a publisher ever asks to revert an extension back to pending, we can easily do a clean reset on that entry.

ajperel (approved):
Thanks for all the iterating.

PR #10919, opened 2026-08-11
fix: prevent gen1 node functions from deploying in kits

ajperel:
### Description

We only support gen2 node functions in kits, so we want to fail as fast as possible if a user attempts to deploy one that contains gen1 functions.

This explicitly restricts kit codebase deployments to gen2 functions by introducing an early AST inspection step in `src/deploy/functions/prepare.ts`. If any gen1 functions are discovered within a kit prior to resolving the backend, a terminal `FirebaseError` is immediately thrown to safely halt the deployment.

### Scenarios Tested

- Regular codebases (`codebase: "default"`) containing gen1 endpoints pass successfully.
- Kit configured codebases containing exclusively gen2 endpoints pass successfully.
- Kit configured codebases containing a gen1 endpoint throw the specific `FirebaseError`.
- Automated test coverage added for all three evaluation scenarios in `prepare.spec.ts`.

### Sample Commands

TAG=agy
CONV=fbade4e0-0d6f-426c-ae14-4f505375f008

### DIFF
--- a/src/deploy/functions/prepare.spec.ts
+++ b/src/deploy/functions/prepare.spec.ts
@@ -714,6 +714,62 @@ describe("prepare", () => {
     });
   });
 
+  describe("checkKitForGen1", () => {
+    it("should do nothing for regular codebases with gen1 functions", () => {
+      const localCfg = { source: "src", codebase: "default" };
+      const wantBuild = build.of({
+        test: {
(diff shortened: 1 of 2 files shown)


wandamora (approved):
LGTM after addressing the GCA comments.

PR #10920, opened 2026-08-11
feat(functions): Add kit migration template and support both net new install and migration case

wandamora:
### Description

This adds a new template (`index-kit-migration.ts`) to support the Extension -> kit migration case, where users will be leveraging exported environment variables from the `ext:export` command to set the global options in their kit instance. A new flag, `--template`, is added to the install command to configure which template to use. `installation` is the default using the standard `index-kit.ts` template, and `migration` will use the `index-kit-migration.ts` template.

### Scenarios Tested
 
* Installing a package with `--template migration` adds the right template and it compiles
* Installing a package without the template command works the same as before.

### Sample Commands

* `firebase functions:kits:install --project <project> --template migration`
* `firebase functions:kits:install --project <project>`
* `npm run build`

### DIFF
--- a/src/commands/functions-kits-install.ts
+++ b/src/commands/functions-kits-install.ts
@@ -33,6 +33,11 @@ const INDEX_KIT_MIGRATION_TEMPLATE = readTemplateSync(
   "init/functions/typescript/index-kit-migration.ts",
 );
 
+const TEMPLATES = {
+  installation: INDEX_KIT_TEMPLATE,
+  migration: INDEX_KIT_MIGRATION_TEMPLATE,
+};
+
 const FUNCTION_KITS_DIR = "function-kits";
 
 export interface FunctionsKitsInstallOptions extends Options {
@@ -151,7 +156,7 @@ export const command = new Command("functions:kits:install")
     }
 
     const templateType = options.template || "installation";
-    if (templateType !== "installation" && templateType !== "migration") {
+    if (!(templateType in TEMPLATES)) {
       throw new FirebaseError(
         `Invalid template '${templateType}'. Template must be 'installation' or 'migration'.`,
       );
--- a/src/commands/functions-kits-install.ts
+++ b/src/commands/functions-kits-install.ts
@@ -29,11 +29,15 @@ const PACKAGE_NO_LINTING_TEMPLATE = readTemplateSync(
 const TSCONFIG_TEMPLATE = readTemplateSync("init/functions/typescript/tsconfig.json");
 const GITIGNORE_TEMPLATE = readTemplateSync("init/functions/typescript/_gitignore");
 const INDEX_KIT_TEMPLATE = readTemplateSync("init/functions/typescript/index-kit.ts");
+const INDEX_KIT_MIGRATION_TEMPLATE = readTemplateSync(
+  "init/functions/typescript/index-kit-migration.ts",
+);
 
 const FUNCTION_KITS_DIR = "function-kits";
 
 export interface FunctionsKitsInstallOptions extends Options {
   npm_package?: string;
+  template?: string;
 }
 
 /**
@@ -134,11 +138,23 @@ export async function checkPackageHasShrinkwrap(rawPkgName: string): Promise<boo
 export const command = new Command("functions:kits:install")
   .description("install a function kit into your project")
   .option("--npm_package <package>", "NPM package name or specifier to install as a function kit")
+  .option(
+    "--template [installation|migration]",
+    "template to use for the kit index file",
+    "installation",
+  )
   .action(async (options: FunctionsKitsInstallOptions): Promise<void> => {
     experiments.assertEnabled("kits", "install a function kit");
 
     if (!options.config) {
       throw new FirebaseError("Not in a Firebase project directory (firebase.json not found).");
+    }
+
+    const templateType = options.template || "installation";
+    if (templateType !== "installation" && templateType !== "migration") {
+      throw new FirebaseError(
+        `Invalid template '${templateType}'. Template must be 'installation' or 'migration'.`,
+      );
     }
 
     const rawPkgName = options.npm_package;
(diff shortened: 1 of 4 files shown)


src/commands/functions-kits-install.ts
```diff
@@ -134,13 +138,25 @@ export async function checkPackageHasShrinkwrap(rawPkgName: string): Promise<boo
 export const command = new Command("functions:kits:install")
   .description("install a function kit into your project")
   .option("--npm_package <package>", "NPM package name or specifier to install as a function kit")
+  .option(
+    "--template [installation|migration]",
```
ajperel, line 147: I could be easily convinced this is fine but I wonder if we want to use a slightly more specific name like `ext-migration` just in case we ever move anything else to kits?  But this is very bikesheddy. Let's land the PR and we can change name in a follow up if we want.

src/commands/functions-kits-install.ts
```diff
@@ -134,13 +138,25 @@ export async function checkPackageHasShrinkwrap(rawPkgName: string): Promise<boo
...
+    "installation",
+  )
   .action(async (options: FunctionsKitsInstallOptions): Promise<void> => {
     experiments.assertEnabled("kits", "install a function kit");
 
     if (!options.config) {
       throw new FirebaseError("Not in a Firebase project directory (firebase.json not found).");
     }
 
+    const templateType = options.template || "installation";
```
ajperel, line 158: you use "installation" and "migration" a fair number of times... maybe best to add constants?

My gemini agreed

const TEMPLATES = {
  installation: INDEX_KIT_TEMPLATE,
  migration: INDEX_KIT_MIGRATION_TEMPLATE,
};

if (!(templateType in TEMPLATES))

templates/init/functions/typescript/index-kit-migration.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Extension to its equivalent function kit. It will set the global options for
+ * the function kit to equal what is configured for the Extension instance.
+ * Configuration from the Extension instance to migrate can be exported using
+ * the `firebase ext:export --mode functions --instance <ext-instance-id>`
+ * command.
+ */
+
+import { setGlobalOptions } from "firebase-functions";
+import { MemoryOption, VpcEgressSetting, IngressSetting } from "firebase-functions/v2/options";
+import * as params from "firebase-functions/params";
```
ajperel: I suggested this when we're going to have a ton of commented out defineBoolean, etc. IT's fine but I think it'd be OK now that we're only using and intending to use defineString to specifically only import that.

templates/init/functions/typescript/index-kit-migration.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const regionParam = params.defineString("FUNCTION_DEFAULT_REGION", {
+  description: "Region where functions should be deployed.",
```
ajperel: Maybe a bit better:

Global default region where functions should be deployed. Can be overriden per-function.

templates/init/functions/typescript/index-kit-migration.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Configuration from the Extension instance to migrate can be exported using
+ * the `firebase ext:export --mode functions --instance <ext-instance-id>`
+ * command.
+ */
+
+import { setGlobalOptions } from "firebase-functions";
+import { MemoryOption, VpcEgressSetting, IngressSetting } from "firebase-functions/v2/options";
+import * as params from "firebase-functions/params";
+
+export const regionParam = params.defineString("FUNCTION_DEFAULT_REGION", {
```
ajperel: I'd copy the comment here from the other template.

templates/init/functions/typescript/index-kit-migration.ts
```diff
@@ -0,0 +1,45 @@
...
+  description: "Region where functions should be deployed.",
+});
+
+// This allows you to set default options that apply to all functions in this
+// kit. Learn more about these options and additional configurations at:
+// https://firebase.google.com/docs/reference/functions/2nd-gen/node/firebase-functions.globaloptions
+setGlobalOptions({
+  region: regionParam,
+  memory: (process.env.EXT_MIGRATED_SYSTEM_MEMORY as MemoryOption) || undefined,
```
ajperel: I'd probably use `??` over `||` so that it doesn't coerce a falsy value like empty string.

ajperel: Can you double check that this should be undefined and not null. @inlined had mentioned using null I think but I don't know if setGlobalOptions treats the same. We want the behavior like these aren't set at all.

templates/init/functions/typescript/index-kit-migration.ts
```diff
@@ -0,0 +1,45 @@
...
+  vpcConnector: process.env.EXT_MIGRATED_SYSTEM_VPCCONNECTOR || undefined,
+  maxInstances: process.env.EXT_MIGRATED_SYSTEM_MAXINSTANCES
+    ? Number(process.env.EXT_MIGRATED_SYSTEM_MAXINSTANCES)
+    : undefined,
+  minInstances: process.env.EXT_MIGRATED_SYSTEM_MININSTANCES
+    ? Number(process.env.EXT_MIGRATED_SYSTEM_MININSTANCES)
+    : undefined,
+  ingressSettings: (process.env.EXT_MIGRATED_SYSTEM_INGRESSSETTINGS as IngressSetting) || undefined,
+  labels: process.env.EXT_MIGRATED_SYSTEM_LABELS
+    ? (JSON.parse(process.env.EXT_MIGRATED_SYSTEM_LABELS) as Record<string, string>)
```
ajperel, line 44: Let's land this, but please follow up on if this parse is actually the right thing to do or not. We're not confident on how labels will be represented.

ajperel (approved):
Approved with some nits

templates/init/functions/typescript/index-kit-migration.ts
wandamora: null isn't an accepted value for these params, therefore I went with undefined. These values should be safely ignored when set to undefined. 

According to Jetski:

**When the environment variable is NOT present / unset**: Setting the field to undefined means "no global option specified". `firebase-functions` will ignore it and fall back to:
1. Any function-level configuration defined inside the kit, or
2. Cloud Functions / Cloud Run platform defaults (e.g., standard memory, standard timeout, default VPC/ingress settings).

templates/init/functions/typescript/index-kit-migration.ts
wandamora: Ack. Keeping this open.

templates/init/functions/typescript/index-kit-migration.ts
wandamora: Changed this for both templates.

src/commands/functions-kits-install.ts
wandamora: I'll keep this as `migration` since I don't foresee that being the case (famous last words).

PR #10922, opened 2026-08-11
refactor(kits): rename sourcePackage id to name

wandamora:
### Description

Simply renames the `id` field in `sourcePackage` to `name` for function kits because it is the more common convention to describe the package name.

### Scenarios Tested

* Unit tests


ajperel (approved):
Did a quick manual test too.

PR #10926, opened 2026-08-12
fix(functions): use dart build cli for Dart function bundles

### LINKED ISSUE #10591: [cloud functions for dart] Allow functions in a folder different than "bin" to workaround build hook issues
See: (https://github.com/firebase/firebase-functions-dart/issues/201)](https://github.com/firebase/firebase-functions-dart/issues/201)

Sorry for the cross posting here but I think the issue belongs to here...

### Version info

Dart SDK: dart 3.12
firebase_functions: 0.6.0
firebase cli: 15.19

### Test case

Simply add a dependency (even as a transitive dependency) that requires build_hooks (such as sqlite3)

### Steps to reproduce

- Generate a project using `firebase init functions`
- Add sqlite3 as a dependency
- Call firebase deploy: `firebase deploy --only functions`

### Expected behavior

It should deploy

### Actual behavior

```
i  functions: compiling Dart to linux-x64 executable...

Error: Dart compilation failed with exit code 255. Make sure your Dart project compiles successfully with: dart compile exe bin/server.dart --target-os=linux --target-arch=x64
```
And indeed:
```
$ dart compile exe bin/server.dart --target-os=linux --target-arch=x64
'dart compile' does not support build hooks, use 'dart build' instead.
Packages with build hooks: sqlite3.
```

Looking at https://github.com/dart-lang/sdk/blob/80f770eb5752322a320b003b45dc79b9e14a5fba/pkg/dartdev/lib/src/commands/compile.dart#L742-L759
We cannot bypass this check and force dart compile to work as the "bin" folder is a hardcoded required.

## Solutions

- Allow configuring the dart entry point (pubspec.yaml?)
- Look in multiple locations, not just the bin folder
- Use `dart build` instead of `dart compile`

demolaf:
Closes https://github.com/firebase/firebase-tools/issues/10591

Following the release of Dart [3.13.0](https://github.com/dart-lang/sdk/commit/da6595cd6bb5d4c0a185d759a025e879ff06e631), we can now cross-compile with `dart build cli` (previously only `dart compile exe` supported `--target-os`/`--target-arch`), which unblocks using build hooks for Dart functions.

`dart compile exe` produces a single ahead-of-time compiled executable and skips Dart's native build hooks, so functions depending on packages with native assets don't build correctly. Switching to `dart build cli` runs those hooks and produces a proper bundle instead of a bare executable.

This updates the Dart runtime delegate to invoke `dart build cli` with the same target/os/arch flags, and updates `prepare.ts` to look for the resulting server executable at its new bundle path (`build/cli/linux_x64/bundle/bin/server`) instead of `bin/server`.

Related to https://github.com/firebase/firebase-functions-dart/issues/201

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Adds --immediate flag to ext:uninstall (#10921)
 - Updated dataconnect emulator version to v3.4.18.
 - Configured OneMCP server tools to require a Firebase project by default, with options to opt-out specific tools (such as Developer Knowledge document search).
-- Fixed a bug where deploying Dart functions with dependencies that use native build hooks (e.g. `sqlite3`) failed to compile.
+- Fixed a bug where deploying Dart functions with dependencies that use native build hooks (e.g. `sqlite3`) failed to compile. Note: Dart functions now require Dart SDK 3.13.0 or later.
 - Fixed a bug where deploying functions with the `dartfunctions` experiment enabled could incorrectly prompt to delete existing GCF v2 functions.
 - Added `outputSchema` support for local MCP tools.
 - Skip functions lifecycle hooks during partial (filtered) deployments, and print instructions for running them manually.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,3 +6,13 @@
 - Load existing `.env` files and passes environment variables to functions discovery in `runtimeDelegate`.
 - Adds --immediate flag to ext:uninstall (#10921)
 - Updated dataconnect emulator version to v3.4.18.
+- Configured OneMCP server tools to require a Firebase project by default, with options to opt-out specific tools (such as Developer Knowledge document search).
+- Fixed a bug where deploying Dart functions with dependencies that use native build hooks (e.g. `sqlite3`) failed to compile.
+- Fixed a bug where deploying functions with the `dartfunctions` experiment enabled could incorrectly prompt to delete existing GCF v2 functions.
+- Added `outputSchema` support for local MCP tools.
+- Skip functions lifecycle hooks during partial (filtered) deployments, and print instructions for running them manually.
+- Added `appcheck:providers:list`, `appcheck:providers:get` and `appcheck:providers:set` to configure App Check attestation providers for an app.
+- Added `appcheck:apps:list` to show every app with its configured App Check providers.
+- Added web app support for Crashlytics MCP tools and prompts.
+- Added support for forwarding custom HTTP headers (`Mcp-Param-*`) to remote MCP tools when defined in tool parameter input schemas (`x-mcp-header`), per [SEP-2243](https://modelcontextprotocol.io/seps/2243-http-standardization).
+- Improved function parameter prompting clarity for multi-codebase deploys (#10897)
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -32,6 +32,7 @@ import {
 } from "./functionsDeployHelper";
 import { logLabeledBullet, logLabeledWarning } from "../../utils";
 import { isDartEndpoint, classifyNonProductionEndpoints } from "./runtimes/dart/triggerSupport";
+import { DART_BUNDLE_EXECUTABLE_PATH } from "./runtimes/dart";
 import { getFunctionsConfig, prepareFunctionsUpload } from "./prepareFunctionsUpload";
 import { promptForFailurePolicies, promptForMinInstances } from "./prompts";
 import { needProjectId, needProjectNumber } from "../../projectUtils";
@@ -445,8 +446,7 @@ export async function prepare(
         ? "tar.gz"
         : "zip";
 
-      const isDart = supported.runtimeIsLanguage(wantBuilds[codebase].runtime, "dart");
-      const executablePaths = isDart ? ["build/cli/linux_x64/bundle/bin/server"] : [];
+      const executablePaths = getExecutablePaths(wantBuilds[codebase].runtime);
 
       const packagedSource = await prepareFunctionsUpload(
         options.config.projectDir,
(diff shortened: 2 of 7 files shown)


Lyokone (approved):

src/deploy/functions/runtimes/dart/index.ts
wandamora: This needs to be bumped to 3.13.0 to use `dart build cli` right? Update the doc comment too.

src/deploy/functions/prepare.ts
```diff
@@ -431,7 +431,7 @@ export async function prepare(
         : "zip";
 
       const isDart = supported.runtimeIsLanguage(wantBuilds[codebase].runtime, "dart");
-      const executablePaths = isDart ? ["bin/server"] : [];
+      const executablePaths = isDart ? ["build/cli/linux_x64/bundle/bin/server"] : [];
```
wandamora: Can you make "build/cli/linux_x64/bundle/bin/server" a constant? Perhaps `DART_BUNDLE_EXECUTABLE_PATH`?

wandamora (commented):
It seems more changes are required for this to work.

1. https://github.com/firebase/firebase-tools/blob/91745b08a02c98af4e9b0347059dd29b01f62dee/src/init/features/functions/dart.ts#L27 Because `build` is in config.ignore, readdirRecursive (https://github.com/firebase/firebase-tools/blob/91745b08a02c98af4e9b0347059dd29b01f62dee/src/deploy/functions/prepareFunctionsUpload.ts#L109-L117) skips the entire `build/` directory. The uploaded archive will omit `build/cli/linux_x64/bundle/bin/server` and its companion native asset libraries (`build/cli/linux_x64/bundle/lib/`), causing Cloud Run deployments to fail at runtime. We would need to change the default ignore config for dart and also ensure that `build/cli/**` files are discoverable before calling `prepareFunctionUpload`: https://github.com/firebase/firebase-tools/blob/91745b08a02c98af4e9b0347059dd29b01f62dee/src/deploy/functions/prepare.ts#L433-L436

To avoid archive bloat, we could change the ignored files to allowlist the bundle:
```
"ignore": [
  ".dart_tool",
  "build/**",
  "!build/cli/linux_x64/bundle/**"
]
```

2. `bin/server` should be updated to `build/` in the .gitignore template: https://github.com/firebase/firebase-tools/blob/91745b08a02c98af4e9b0347059dd29b01f62dee/templates/init/functions/dart/_gitignore#L2

wandamora:
Also, adding some test cases could be useful:

1. In `src/deploy/functions/runtimes/dart/index.spec.ts`:
   - `validate()`: Verify that an installed Dart SDK < 3.13.0 rejects with a FirebaseError instructing the user to upgrade.
   - `build()`: Verify that spawn is called with arguments `["build", "cli", "--target", "bin/server.dart", "--target-os", "linux", "--target-arch", "x64"]`.
   - `build()` failure: Verify that non-zero exit codes throw a `FirebaseError` containing the updated `dart build cli --target ...` remediation command.
   - `build()` emulator check: Verify that `build()` is skipped when `EmulatorRegistry.isRunning(Emulators.FUNCTIONS)` is true.

2. In `src/deploy/functions/prepare.spec.ts`:
   - Verify that `prepare()` sets `executablePaths` to `["build/cli/linux_x64/bundle/bin/server"]` when `wantBuilds[codebase].runtime` is Dart.

src/deploy/functions/runtimes/dart/index.ts
demolaf: Yes, thanks for catching that. Also bumped in `firebase_functions` pubspec.yaml.

src/deploy/functions/prepare.ts
demolaf: I'll do that

src/deploy/functions/runtimes/dart/index.ts
demolaf: Interestingly, I did a `npm link` to this branch and I deployed successfully and successfully fired a request to the function, so that makes me wonder if `functions.ignore` applied correctly.

src/deploy/functions/prepare.ts
```diff
@@ -430,8 +431,7 @@ export async function prepare(
         ? "tar.gz"
         : "zip";
 
-      const isDart = supported.runtimeIsLanguage(wantBuilds[codebase].runtime, "dart");
-      const executablePaths = isDart ? ["bin/server"] : [];
+      const executablePaths = getExecutablePaths(wantBuilds[codebase].runtime);
```
wandamora, line 451: So, for new projects using `firebase init`, we don't need to worry about excluding the `build/` files now. However for existing users, they'll continue to have the `build/` directory ignored (the template won't change for them. We may need to add some work around in prepare.ts, such as:

```
let uploadCfg = localCfg;
if (isDart && uploadCfg.ignore?.includes("build")) {
  uploadCfg = {
    ...uploadCfg,
    ignore: uploadCfg.ignore.filter((i) => i !== "build" && i !== "build/"),
  };
}

const packagedSource = await prepareFunctionsUpload...
```

with a comment explaining why.

demolaf: I've added `stripStaleDartBuildIgnore` to auto-strip the stale `build` ignore entry in `firebase.json`

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 - Configured OneMCP server tools to require a Firebase project by default, with options to opt-out specific tools (such as Developer Knowledge document search).
+- Fixed a bug where deploying Dart functions with dependencies that use native build hooks (e.g. `sqlite3`) failed to compile.
```
wandamora: I would briefly mention that we require 3.13.0 now. Could add "Note: Dart functions now require Dart SDK 3.13.0 or later."

wandamora (approved):
Overall,

demolaf:
A follow-up PR will be opened to make this a non-breaking change by auto-detecting the Dart language version and choosing the compilation path based on that.

PR #10928, opened 2026-08-12
fix(functions): cleanup declarative security service account when codebase functions are deleted

inlined:
Fixes an issue where deleting all functions in a codebase that uses declarative security (requireRoles) failed or left the managed service account (firebase-fn-...) orphaned in GCP IAM.

Specifically:
- In discoverSecurityDetails: When wantBackend has 0 endpoints, skip IAM permission checks, role queries, and managed SA generation. Return existing security metadata if an existing SA was deployed, or {} if both want and have backends are empty.
- In createDeploymentPlan: Require hasWantEndpoints to be true before treating declarative security as active. If wantBackend has 0 endpoints and haveBackend has an existing managed SA, set serviceAccountToDelete = existingManagedSA so the service account is cleanly deleted. If both wantBackend and haveBackend are empty, no service account is created or deleted.

- Deleting all functions from a codebase while keeping requireRoles in global options: verified that serviceAccountToDelete is set and the managed service account is deleted.
- Deploying an empty codebase with requireRoles specified (0 functions in want and have): verified that no service account is created and no IAM permission errors are thrown.
- Deploying functions with requireRoles: verified managed service account creation and role assignment works normally.
- Partial function deployments (--only functions:func1): verified managed service account is retained.

- npx mocha src/deploy/functions/release/planner.spec.ts src/deploy/functions/prepare.spec.ts
- firebase deploy --only functions

### DIFF
--- a/src/deploy/functions/prepare.ts
+++ b/src/deploy/functions/prepare.ts
@@ -65,11 +65,17 @@ import * as iam from "../../gcp/iam";
 import * as resourcemanager from "../../gcp/resourceManager";
 
 export const EVENTARC_SOURCE_ENV = "EVENTARC_CLOUD_EVENT_SOURCE";
+export const DECLARATIVE_SECURITY_ETAG_LABEL = "firebase-declarative-security-etag";
 
 /**
  * Discovers and coordinates declarative security details for a codebase.
- * Mutates want Backend to populate managed service account and etag labels.
- * Returns existing discovered roles, or undefined if no security changes needed.
+ * Mutates `want` Backend to populate managed service account and etag labels.
+ *
+ * Returns security metadata based on deployment state:
+ * - **Enrolling / Active**: Returns `{ haveRoles, haveRolesEtag, existingManagedSA, managedSA, newEtag }`.
+ * - **Unenrolling (Opting Out)**: Returns `{ existingManagedSA, haveRolesEtag }` so planner can schedule role revocation.
+ * - **Deleting All Functions**: Returns `{ existingManagedSA, ...(haveRolesEtag ? { haveRolesEtag } : {}) }` so planner can delete the SA.
+ * - **Inactive**: Returns `{}` when declarative security is not used in `want` or `have`.
  */
 export async function discoverSecurityDetails(
   codebase: string,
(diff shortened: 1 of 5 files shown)


src/deploy/functions/prepare.ts
```diff
@@ -89,15 +89,13 @@ export async function discoverSecurityDetails(
...
-      ? firstHave.serviceAccount
-      : undefined;
-  }
+  const haveEndpoints = backend.allEndpoints(have);
+  const existingManagedSA = haveEndpoints.find(
+    (e) => typeof e.serviceAccount === "string" && e.serviceAccount.startsWith("firebase-fn-"),
+  )?.serviceAccount;
+  const haveRolesEtag = haveEndpoints.find(
+    (e) => e.labels?.["firebase-declarative-security-etag"],
+  )?.labels?.["firebase-declarative-security-etag"];
```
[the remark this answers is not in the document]
inlined: done

src/deploy/functions/release/planner.ts
```diff
@@ -181,7 +181,9 @@ export async function createDeploymentPlan(args: PlanArgs): Promise<CodebasePlan
     !deleteAll
   );
 
-  if (requiredRoles) {
+  const hasWantEndpoints = backend.allEndpoints(wantBackend).length > 0;
```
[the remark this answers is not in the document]
inlined: done

src/deploy/functions/prepare.ts
```diff
@@ -115,6 +113,16 @@ export async function discoverSecurityDetails(
...
 
+  if (backend.allEndpoints(want).length === 0) {
+    if (existingManagedSA && haveRolesEtag) {
+      return {
+        haveRolesEtag,
+        existingManagedSA,
+      };
+    }
+    return {};
+  }
```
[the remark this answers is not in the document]
inlined: done

src/deploy/functions/prepare.ts
```diff
@@ -89,15 +89,15 @@ export async function discoverSecurityDetails(
...
-      : undefined;
-  }
+  const existingManagedSA =
+    backend.findEndpoint(
+      have,
+      (e) => typeof e.serviceAccount === "string" && e.serviceAccount.startsWith("firebase-fn-"),
+    )?.serviceAccount ?? undefined;
+  const haveRolesEtag = backend.findEndpoint(
+    have,
+    (e) => !!e.labels?.["firebase-declarative-security-etag"],
```
ajperel: I wonder if this magic string should be pulled out into a constant.

src/deploy/functions/prepare.ts
ajperel: As I try to understand all this... I think this comment could be improved. Are you only returning existing roles if they should be deleted? It's a little hard to know what will be populated in what circumstances.

ajperel (approved):
LGTM though definitely we could make it easier to follow all the logic and edge cases with some comments, etc.

src/deploy/functions/prepare.ts
inlined: done

src/deploy/functions/prepare.ts
inlined: done

